Issue #962, opened 2019-04-29
Replace communication prefs with email prefs link


ryanfeeley:
As discussed with legal, we can remove the communication preferences section, and instead offer a link to email preferences at the bottom of fxa.
![email-preferences](https://user-images.githubusercontent.com/68213/56906871-1d7b9d80-6a71-11e9-9fc7-0f247b12689b.png)

Icon link:
https://design.firefox.com/icons/viewer/#open%20in

Position 20px to the right of the link, vertically aligned to the center of the text as depicted here:
<img width="240" alt="image" src="https://user-images.githubusercontent.com/68213/56907140-a85c9800-6a71-11e9-873a-0a55e1151696.png">

dannycoates:
@ryanfeeley I noticed you already changed the title of "Communication preferences" to "Email preferences". Afaict we can only link directly to the newsletter prefs on `https://www.mozilla.org/en-US/newsletter/existing/...` if the user has previously opted in (and has a record in basket). What should we do when they haven't? The current behavior has a "Subscribe" or "Manage" button depending on their state.

shane-tomlinson:
I had a chat with @pmac about doing this. The TL;DR, is we can't really do this for the reason @dannycoates described. There is nowhere we can link to if the user hasn't already subscribed to a newsletter.

pmac:
@shane-tomlinson I'm doing some work related to this soon that will cause a toggle to show in the email pref center to opt out of accounts related email (separate from the newsletter). It's possible I can do some other work to allow what you need... we'd have to figure out a way to allow this to happen, but it is something we could potentially do. This is all an issue because we're in a strange state where the FxA data isn't in the same place as the rest of our Salesforce stuff, but that should be resolved later this year as well. In the interim we can discuss some solutions if you'd like.

shane-tomlinson:
> In the interim we can discuss some solutions if you'd like.

@pmac sent an invite for Monday the 13th!

shane-tomlinson:
ref https://docs.google.com/document/d/1kl8HQtjdUy0EcT0lid1idx4Q0k0l0_9erxHuP_VStKg

shane-tomlinson:
Some more context from [the basket bug](https://github.com/mozmeao/basket/issues/167#issuecomment-493567809):

@vbudhram is going to take over this work. Copying some context from the email:

[...]

PR #1243, opened 2019-05-23
feat(trailhead): update comms link to go directly to basket

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/962

This patch updates the FxA communication panel to link directly to basket. It also appends the user's email address so that basket can look the user up. It uses the `https://www-dev.allizom.org/newsletter/existing/?email=a%40a.com` url.

This url is configurable for different environments ref https://github.com/mozilla/fxa/blob/baea0b5a7a67e21cfec40b73736c3dde803ee170/packages/fxa-content-server/server/lib/configuration.js#L369.

- [x] Add icon to panel
- [x] Add tests
- [x] Determine the email preferences url that will be linked to

Screen Shot 2019-05-23 at 4 21 35 PM

cc @pmac


shane-tomlinson:
### Desktop

Screenshot 2019-05-24 at 12 30 58

### Mobile
Screenshot 2019-05-24 at 12 30 52

### RTL
Screenshot 2019-05-24 at 12 30 40

@mozilla/fxa-devs - r?

We'll still need to confirm the URL with @pmac, but we can update that using configuration easily enough.

@pmac - one final question, with this link, will users of all locales be able to opt-in to newsletters? Right now we only show the pane for users [of these locales](https://github.com/mozilla/fxa/blob/04df8cd9417108f6fc03dbed398857e7b6740c62/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js#L14)

shane-tomlinson:
Note to @jrgm and @jbuck - once this lands, we'll no longer need the basket proxy. 🎆🎆🎆🎆🎆🎆🎆🎆

pmac:
> We'll still need to confirm the URL with @pmac, but we can update that using configuration easily enough.

I got the Oauth flow working last night against stable 🎉 So the Oauth dance is viable for us and should work. There are still questions flying around on whether we can legally use it since it requires auth, but if we do go with the Oauth flow then the links will be in the form `https://basket.mozilla.org/fxa/?email=user%40example.com`.
 
> @pmac - one final question, with this link, will users of all locales be able to opt-in to newsletters? Right now we only show the pane for users [of these locales](https://github.com/mozilla/fxa/blob/04df8cd9417108f6fc03dbed398857e7b6740c62/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js#L14)

That is a tricky question. Technically all locales can opt-in, but the email preferences page is only translated into roughly that same list of locales, and the newsletters are translated and sent often in even fewer. But a lot of people might choose an available language that isn't their primary choice, so I hesitate to suggest limiting access...

pmac:
I've gotten confirmation that we're going with the Oauth thing. So the URL for email preferences is what I said above:

```
https://basket.mozilla.org/fxa/?email=user%40example.com
```

I have this deployed to basket-stage and working against FxA stable. e.g. `https://basket.allizom.org/fxa/?email=user%40example.com`. The `email` parameter in the URL is optional but provides a better UX.

shane-tomlinson:
@mozilla/fxa-devs - r?

clouserw:
@dannycoates can you review this so we can get it in train 138?

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -46,24 +46,29 @@ const conf = module.exports = convict({
...
     },
     api_url: {
       default: 'http://127.0.0.1:10140',
-      doc: 'Url for the Basket API server',
+      doc: 'DEPRECATED - Url for the Basket API server',
+      format: String
+    },
+    email_preferences_url: {
+      default: 'https://basket.allizom.org/fxa/',
+      doc: 'DEPRECATED - Url for the Basket proxy server',
```
pmac: @shane-tomlinson this doc string isn't right. And the default URL is for basket staging. This is known right? The prod URL will be different? Perhaps the default should be basket prod?

shane-tomlinson: You are right, and this was intentional. We kinda/sorta go back and forth about whether to put production values as default, or development values. We are settling on development values. I'll open a PR against the cloudops-deployment repo with the correct URL for prod.

PR #1256, opened 2019-05-24
chore(l10n): Add the "Email preferences" string for l10n extraction

shane-tomlinson:
Extraction from PR #1243 so it can be extracted today.

@dannycoates - r?


shane-tomlinson:
Closing, this string is [already being extracted](https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/af/LC_MESSAGES/client.po#L1561), this PR is not needed.

Issue #380, opened 2019-02-21
Refactor common validation schemata


philbooth:
In mozilla/fxa-auth-server#2929, validation was added to the auth server that duplicates equivalent rules in the content server. We should probably extract it from both places to this repo, along with any other common validation logic.

Admittedly I haven't thought through all of the intricacies of this yet. Like, should `joi` be a peer dependency to this repo or a fully-fledged dependency in its own right? The former is probably easier when it comes to updating deps in normal development, but could maybe get into a pickle when the versions get updated independently?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-598)
┆Issue Number: FXA-598

PR #1261, opened 2019-05-26
feat: initial Typescript based fxa event broker

bbangert:
Implements the message receiving portion of the event broker in a new TypeScript based Node app.

I removed the Rust version, though it may be worth revisiting at a later point when we get some Google SDK libraries that are more ergonomic (work is in progress in other groups to address this). Handling it in Node was the most timely way to proceed, and I went with TypeScript to match the new Typescript client-side work.

This is using a few things that have been discussed in various FxA meetings and input/thoughts on the results of using them here are appreciated:
- Prettier for consistent code formatting
- tslint on base recommended level (the stricter ones play less nicely without fully type declared JS libs)
- no grunt, and no *.js scripts where possible, just npm run scripts
- sqs-consumer rather than our custom sqs receiver code duplicated elsewhere
- Moved to @hapi/joi as joi was deprecated/moved.

I otherwise stuck with existing conventions for the other Node microservices we use, along with the same mock/test/validation/etc libs.

**Note**: Issues must be filed for the remaining work before this PR should be merged:
- ~~Add tests for the database~~ #1266 
- ~~Add .circleci/Dockerfile and integrate with the rest of the FxA build/release process~~ #1267
- ~~Implement GCP Pub/Sub storing of messages~~ #1268
- ~~Implement RP proxy from GCP Pub/Sub HTTPS endpoint~~ #1269


bbangert:
I think your rationale for removing it makes sense, the tab completion thing hadn't occurred to me. I'll remove it.

packages/fxa-event-broker/lib/db/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { FirestoreDatastore } from './firestore';
+import { InMemoryDatastore } from './in-memory';
+
+interface Datastore {
+  /**
+   * Stores that a login has occured for user `uid` on service `clientId`.
+   *
+   * The Datastore should treat this as an imdempotent request.
```
vladikoff: typo `idempotent`?

packages/fxa-event-broker/.gitignore
```diff
@@ -1,3 +1,21 @@
...
+/test/load/bin
+/test/load/lib*
+/test/load/include
+*~
+*.pyc
+.nyc_output
+npm-debug.log
+server.pot
+.mail_output
+fxa-content-server-l10n/
```
vladikoff: nit: hm, not sure we need all this here?

packages/fxa-event-broker/.prettierrc
```diff
@@ -0,0 +1 @@
+{ "useTabs": false, "printWidth": 80, "tabWidth": 2, "singleQuote": true, "trailingComma": "none", "jsxBracketSameLine": false, "parser": "typescript" }
```
vladikoff: is it possible to format this (with new lines for prettier rules?)?

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { SQS } from 'aws-sdk';
+import { Consumer } from 'sqs-consumer';
+import * as Joi from 'typesafe-joi';
+
+import { Logger } from 'mozlog';
```
vladikoff: nit: `mozlog` is a module liek the rest above, should the new line be after it?

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.app.start();
+  }
+
+  public stop() {
+    this.app.stop();
+  }
+
+  private async handleMessage(sqsMessage: SQS.Message) {
+    const body = JSON.parse(sqsMessage.Body || '{}');
+    const message = Joi.attempt(JSON.parse(body.Message), BASE_MESSAGE_SCHEMA);
```
vladikoff: 👍 good to see validation here

packages/fxa-event-broker/types/mozlog/index.d.ts
```diff
@@ -0,0 +1,18 @@
+declare module 'mozlog' {
```
vladikoff: I wonder if we can add this to the mozlog npm package?

release.sh
```diff
@@ -293,7 +293,6 @@ packages/fxa-content-server
 packages/fxa-customs-server
 packages/fxa-email-event-proxy
 packages/fxa-email-service
-packages/fxa-event-broker
```
vladikoff: don't we want to include this as part of the release process or?

packages/fxa-event-broker/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "fxa-event-broker",
+  "version": "1.137.1",
+  "description": "Firefox Accounts Event Broker",
+  "scripts": {
+    "build": "./node_modules/typescript/bin/tsc",
+    "tslint": "./node_modules/tslint/bin/tslint -p .",
+    "test": "./node_modules/mocha/bin/mocha -r ts-node/register test/**/*.spec.ts",
+    "start-processor": "node ./dist/bin/worker.js"
```
vladikoff: can we just add a `start` script here that runs a `build` and calls start-processor?

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+      case SUBSCRIPTION_UPDATE_EVENT: {
+        const subMessage = Joi.attempt(message, SUBSCRIPTION_UPDATE_SCHEMA);
+        const clientIds = await this.db.fetchClientIds(subMessage.uid);
+        // TODO: Queue a subscription event for each clientId for delivery.
+        return;
+      }
+    }
+    // Anything that isn't a message we want
+    this.logger.debug('unwantedMessage', { message });
```
vladikoff: should this remain `debug` or be a `warn` or `info` so we can catch an unwanted message in prod?

packages/fxa-event-broker/types/mozlog/index.d.ts
bbangert: That'd definitely be preferable.

packages/fxa-event-broker/package.json
bbangert: Sure.

release.sh
bbangert: We will, yes, think this was caught during a prior commit that was squashed.

_scripts/install_all.sh
```diff
@@ -28,8 +28,6 @@ cargo build --bin fxa_email_send;
 ../../_scripts/clone-authdb.sh
 cd ..
 
-cd fxa-event-broker; cargo build; cd ..
```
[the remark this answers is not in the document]
bbangert: Caught by a prior squashed commit, will fix that here.

packages/fxa-event-broker/config/index.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as Convict from 'convict';
+import * as Fs from 'fs';
+import * as Path from 'path';
+
+const conf = Convict({
+  env: {
+    default: 'prod',
+    doc: 'The current node.js environment',
+    env: 'NODE_ENV',
+    format: ['dev', 'test', 'stage', 'prod']
```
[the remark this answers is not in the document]
bbangert: Indeed, I'm not sure what the conclusion of that was. Was it to use prod or production?

packages/fxa-event-broker/.gitignore
bbangert: I definitely copied it from one of the other projects.

packages/fxa-event-broker/.prettierrc
bbangert: Yup!

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Yup.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,110 @@
...
+    switch (message.event) {
+      case LOGIN_EVENT: {
+        const loginMessage = Joi.attempt(message, LOGIN_SCHEMA);
+        await this.db.storeLogin(loginMessage.uid, loginMessage.clientId);
+        return;
+      }
+      case SUBSCRIPTION_UPDATE_EVENT: {
+        const subMessage = Joi.attempt(message, SUBSCRIPTION_UPDATE_SCHEMA);
+        const clientIds = await this.db.fetchClientIds(subMessage.uid);
+        // TODO: Queue a subscription event for each clientId for delivery.
```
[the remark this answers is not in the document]
bbangert: Yup, was trying to keep this PR from getting larger. So doing he clientId fetch + delivery to each GCP pub topic is a separate issue.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Since we are listening for all events, and only processing the login/subscription ones (this will hang off the main service notification SNS topic), there will be lots of messages we don't want. I'll check with ops to see if they'd prefer to try and configure this SQS listener to filter it down to just the two message event types we want. Later on we will want all of them though, for event logging.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,117 @@
...
+    uid: joi.string().required()
+  })
+  .unknown(true)
+  .required();
+
+const LOGIN_SCHEMA = joi
+  .object()
+  .keys({
+    clientId: joi.string().required(),
+    deviceCount: joi
```
shane-tomlinson: Do we care how strict these schema are? There is a bunch of logic on the auth/content servers for validation that might be able to be shared here. I have a goal of sharing some validation logic in fxa-shared ([@philbooth does too](https://github.com/mozilla/fxa/issues/380)) across the repos.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.app.start();
+  }
+
+  public stop() {
+    this.app.stop();
+  }
+
+  private async handleMessage(sqsMessage: SQS.Message) {
+    const body = JSON.parse(sqsMessage.Body || '{}');
+    const message = joi.attempt(JSON.parse(body.Message), BASE_MESSAGE_SCHEMA);
```
shane-tomlinson: What happens if this attempt fails?

packages/fxa-event-broker/config/index.ts
```diff
@@ -0,0 +1,85 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as Convict from 'convict';
```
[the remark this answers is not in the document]
bbangert: Yep, that's how importing the whole module into a single variable is done in Typescript.

packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Datastore } from './index';
+
+class FirestoreDatastore implements Datastore {
+  private db: Firestore;
+
+  constructor(config: {}) {
+    this.db = new Firestore(config);
+  }
+
+  public async storeLogin(uid: string, clientId: string) {
```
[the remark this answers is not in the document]
bbangert: Nope, it has the same effect with the async keyword here and `: void` is implicit with the interface defining it.

packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (doc.exists) {
+      const data: any = doc.data();
+      if (data.oauth_clients && data.oauth_clients[clientId]) {
+        // Record is already in the database
+        return;
+      }
+    }
+    await document.set(
+      { oauth_clients: { [clientId]: true } },
+      { merge: true }
```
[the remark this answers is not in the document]
bbangert: It does from all the docs I can find on it. I'll find out for sure on my next PR with the db tests. ;)

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: It throws an error. One of the tests explicitly verifies this.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: That would definitely be preferred. Ideally the shared one can use the typesafe-joi, as it helpfully adds proper types to the return value based on the schema.

bbangert: I wasn't too picky about how strict they are since I'm not storing them and such at the moment.

packages/fxa-event-broker/bin/worker.ts
```diff
@@ -0,0 +1,28 @@
...
+import {
+  createDatastore,
+  FirestoreDatastore,
+  InMemoryDatastore
+} from '../lib/db';
+import { ServiceNotificationProcessor } from '../lib/notificationProcessor';
+
+const logger = mozlog(Config.get('log'))('notificationProcessor');
+const db =
+  Config.get('firestore.projectId') === ''
```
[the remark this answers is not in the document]
bbangert: Ah, updated.

packages/fxa-event-broker/lib/db/in-memory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Datastore } from './index';
+
+class InMemoryDatastore implements Datastore {
+  private users: { [key: string]: { [key: string]: boolean } };
+
+  constructor(config: {}) {
+    this.users = {};
+  }
+
+  public async storeLogin(uid: string, clientId: string) {
```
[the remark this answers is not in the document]
bbangert: The annotation isn't needed in TypeScript since the interface defines it. All the tooling works as expected as well. The Typescript interface doesn't support specifying it as `async` which is why the Promise return type is used, while the implementation may use async still. Most of the examples I've found don't bother with annotating implementations since the interface and typechecker validate it already.

PR #1262, opened 2019-05-28
feat(clients): List all attached clients in the devices-and-apps view.

### LINKED ISSUE #1553: Correctly show Fenix device records in the "devices and apps" view
This is a continuation of https://github.com/mozilla/fxa/issues/466 specifically for doing the corresponding front-end changes.

Over in https://github.com/mozilla/fxa/pull/1552 we're adding a new endpoint `/account/attached_clients` which will produce the full list of everything connected to the user's account. This issue is for updating the content-server's "devices and apps" view to use it, rather than the current logic where we make several requests for different subsets of the list and the try to merge them together on the front-end.

rfk:
Fixes #1553. Depends on https://github.com/mozilla/fxa/pull/1552.

This uses the new `/account/attached_clients` endpoint from #1552 to power the "devices and apps" view on the settings page, replacing previous logic that pulled in results from `/sessions` and `/client-tokens`.  

~~As a nod to backwards-compatibility, we have some fallback code that will continue using `/sessions` and `/client-tokens` if the new route is not available.~~ (Edit: since the backend APIs are shipping in train-140, I've removed the fallback code and now assume that they are always available).


packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorized-clients/list.js
```diff
@@ -0,0 +1,138 @@
...
+  },
+  response: {
+    schema: Joi.array().items(Joi.object({
+      client_id: validators.clientId,
+      refresh_token_id: validators.token.optional(),
+      client_name: Joi.string().required(),
+      creation_time: Joi.number().min(0).required(),
+      creation_time_formatted: Joi.string().optional().allow(''),
+      last_access_time: Joi.number().min(0).required().allow(null),
+      last_access_time_formatted: Joi.string().optional().allow(''),
```
rfk: Note to self: the front-end doesn't call this endpoint, so no need to do this formatting; it'll be done by auth-server.

packages/fxa-content-server/app/scripts/models/attached-clients.js
```diff
@@ -9,61 +9,43 @@
...
-    }
-
-    if (clientTypes.sessions) {
-      fetchItems.push(user.fetchAccountSessions(account));
-    }
+  modelId (attrs) {
+    // Make a synthetic `id` attribute because attached clients don't have
+    // a single stable unique identifier. It's only for our internal use.
+    return `${attrs.clientId || ''}:${attrs.deviceId || ''}:${attrs.refreshTokenId || ''}:${attrs.sessionTokenId || ''}`;
+  },
```
rfk: This is kinda gross, happy to take suggestions for an alternative way of doing it.

packages/fxa-content-server/app/scripts/models/attached-client.js
```diff
@@ -0,0 +1,50 @@
...
+
+  destroy () {
+    // Both a sessionToken and deviceId are needed to destroy a device. An account `has a` device, therefore account
+    // destroys the device.
+    //
+    // XXX TODO: the above comment was copied from the `Device model and doesn't seem quite right to me. It suggests to
+    // me that the account will destroy the device in response to observing a 'destroy' notification on the model, but
+    // AFAICT that's not what happens in practice. It's true that the account destroys the device, but it does so
+    // *before* this notification is triggered, and it triggers this notification as a side-effect. So I'm not entirely
+    // sure what this is about...
```
rfk: @shane-tomlinson I will probably need your help understanding this part.

packages/fxa-content-server/app/scripts/models/user.js
```diff
@@ -568,99 +568,29 @@ var User = Backbone.Model.extend({
...
-  /**
-   * Fetch the devices for the given account, populated the passed in
-   * Devices collection.
-   *
-   * @param {Object} account - account for which device list is requested
-   * @param {Object} devices - Devices collection used to store list.
-   * @returns {Promise} resolves when the action completes
-   */
-  fetchAccountDevices (account, devices) {
-    return account.fetchDevices(devices);
```
rfk: By the way, these were all being called in a confusing manner, `account.fetchDevices` returns the list rather than populating it into the passed argument.

packages/fxa-auth-server/lib/routes/attached-clients.js
```diff
@@ -0,0 +1,264 @@
...
+        const uid = sessionToken.uid;
+
+        if (request.payload.deviceId) {
+          // If we got a `deviceId`, then deleting that should also delete the sessionToken
+          // and should return the refreshToken. It's fishy if the values returned here don't
+          // match ones we were given in the request.
+          // XXX TODO: needs to trigger notifications, like `/account/device/destroy`.
+          const device = await db.deleteDevice(uid, request.payload.deviceId);
+          if (device && device.refreshTokenId) {
+            await oauthdb.revokeRefreshTokenById(device.refreshTokenId);
```
rfk: From mtg: this will also need to wipe access tokens, to prevent the client from showing up in the "devices and apps" list.

packages/fxa-auth-server/lib/routes/utils/clients.js
```diff
@@ -0,0 +1,74 @@
...
+              stateCode: location.stateCode,
+            };
+          } else {
+            const territories = require(`cldr-localenames-full/main/${language}/territories.json`);
+            client.location = {
+              country: territories.main[language].localeDisplayNames.territories[location.countryCode]
+            };
+          }
+        } catch (err) {
+          log.warn('attached-clients.formatLocation.warning', {
```
vladikoff: I wonder if keep an eye on these logs at all :shrugs:

packages/fxa-content-server/app/scripts/models/attached-client.js
```diff
@@ -0,0 +1,50 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Device information
+ */
+
+import Backbone from 'backbone';
+
+var AttachedClient = Backbone.Model.extend({
```
vladikoff: I Think i like that this simplifies the types of clients we show 👍

packages/fxa-auth-server/docs/api.md
```diff
@@ -1357,6 +1359,67 @@ by the following errors
   Feature not enabled
 
 
+
+#### GET /account/attached_clients
```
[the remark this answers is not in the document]
rfk: Yeah, the oauth one needs to be a `POST` so it can accept the assertion in the request body.

packages/fxa-auth-server/lib/devices.js
```diff
@@ -117,7 +116,7 @@ module.exports = (log, db, push) => {
           // so try to synthesize one if necessary.
           let deviceName = result.name;
           if (! deviceName) {
-            deviceName = synthesizeName(deviceInfo);
+            deviceName = credentials.client && credentials.client.name || synthesizeName(deviceInfo);
```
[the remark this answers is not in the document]
rfk: I'm trying to do more "fill in defaults on read" than "fill in defaults on write", so that e.g. if we update the oauth client name it will affect all corresponding device records. (The code that sets the name on write is still in place though, because I couldn't make it work nicely in all cases).

packages/fxa-auth-server/lib/routes/defaults.js
```diff
@@ -34,7 +34,7 @@ module.exports = (log, db) => {
     function getVersion() {
       return new Promise((resolve, reject) => {
         // ignore errors and default to 'unknown' if not found
-        const gitDir = path.resolve(__dirname, '..', '..', '.git');
+        const gitDir = path.resolve(__dirname, '..', '..', '..', '..', '.git');
```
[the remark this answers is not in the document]
rfk: I was getting test failures without this change, due to the monorepo. I'll submit it as a separate PR.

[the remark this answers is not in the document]
rfk: ref https://github.com/mozilla/fxa/pull/1372

rfk:
FYI I have split off the server-side chunks of this into https://github.com/mozilla/fxa/pull/1551 and https://github.com/mozilla/fxa/pull/1552 because I think they're ready for review, and will keep this PR in draft for the content-server changes. I've edited the issue description accordingly.

rfk:
This seems to be working well for me in local testing, and tests are passing locally.  Taking "draft" off of it, but I may tweak a few things if the tests turn out to not be passing in CI...

rfk:
If you want to see how this looks in practice, I made a little demo video: https://mozilla.slack.com/archives/CEDLQM1CJ/p1561973345000700

rfk:
@6a68 hi! Shane mentioned that you were interested in learning more about the different parts of FxA, and that this could be a good PR to work through together to get more familiar with some content-server stuff. Would you like to take a look? I'd be happy to meet to go over it synchronously if we can find a time that works.

6a68:
@rfk Sounds great! I'll run through it today and add comments/questions to this PR, and we can get together next week to chat. Would you like to find a time? I think you're more meeting-constrained than I am, these days. I'm happy to stay until 6 my time if that helps.

6a68:
This is a fairly intricate web of related bugs/PRs, not quite sure I know what questions to ask just yet, but I'll continue reading on Monday :+1:

rfk:
@vladikoff if you have a chance if would be great to get your final 👍 on this as well, given how tightly you've been involved in this work.

packages/fxa-content-server/app/tests/spec/models/user.js
```diff
@@ -7,16 +7,13 @@ import { assert } from 'chai';
 import AuthErrors from 'lib/auth-errors';
 import Constants from 'lib/constants';
 import helpers from '../../lib/helpers';
-import Device from 'models/device';
-import AttachedClients from 'models/attached-clients';
```
shane-tomlinson: Glad to see the simplification here.

packages/fxa-content-server/app/tests/spec/models/user.js
```diff
@@ -1083,100 +1024,51 @@ describe('models/user', function() {
...
-      sinon.stub(account, 'destroyOAuthApp').callsFake(function() {
-        return Promise.resolve();
-      });
-
-      oAuthApp = new OAuthApp({
-        id: 'oauth-1',
-        name: 'oauthy',
+      it('removes the account from local storage', () => {
+        assert.isTrue(user.isSignedInAccount.calledWith(account));
+        assert.isTrue(user.removeAccount.calledOnce);
```
shane-tomlinson: You can combine this and the next assertion for a `calledOnceWith`

packages/fxa-content-server/app/tests/spec/models/attached-clients.js
```diff
@@ -158,57 +166,26 @@ describe('models/attached-clients', function() {
...
-          assert.equal(attachedClients.length, 2);
-          assert.equal(attachedClients.at(0).get('clientType'), 'device');
-          assert.equal(attachedClients.at(1).get('clientType'), 'oAuthApp');
-        });
-    });
-
-    it('fetches just devices', function() {
-      return attachedClients.fetchClients({ devices: true }, user).then(() => {
-        assert.equal(attachedClients.length, 1);
+    it('fetches the list of attached clients clients', function() {
```
shane-tomlinson: nit: `clients clients`

packages/fxa-content-server/app/tests/spec/models/attached-client.js
```diff
@@ -0,0 +1,107 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { assert } from 'chai';
+import AttachedClient from 'models/attached-client';
+import sinon from 'sinon';
+
+describe('models/attached-client', function() {
+  var client;
```
shane-tomlinson: Can you use `let` instead of `var` here?

packages/fxa-content-server/app/tests/spec/models/attached-client.js
```diff
@@ -0,0 +1,107 @@
...
+
+  describe('when given details for a device', function() {
```
shane-tomlinson: Since this is a new file, might as well fat arrow all the `function ()`s.

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -2177,238 +2173,155 @@ describe('models/account', function() {
...
         assert.equal(result.length, 3);
-        assert.equal(result[0].clientType, 'device');
         assert.equal(result[0].name, 'alpha');
-        assert.equal(result[0].type, 'desktop');
         assert.equal(result[0].deviceType, 'desktop');
-        assert.ok(result[0].isDevice);
-        assert.notOk(result[0].isWebSession);
 
-        assert.equal(result[1].clientType, 'webSession');
         assert.equal(result[1].name, 'session');
```
shane-tomlinson: Does anything else need to be asserted for result[1] for completeness? Do any conversions take place that should be checked?

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -2177,238 +2173,155 @@ describe('models/account', function() {
...
-      });
-
-      return account.destroyOAuthApp(device).then(assert.fail, () => {
-        assert.isTrue(account._fetchProfileOAuthToken.called, 'fetch called');
-        assert.isTrue(
-          account._oAuthClient.destroyOAuthApp.calledTwice,
-          'called twice'
-        );
-      });
+    it('tells the backend to destroy the client with correct ids', function() {
```
shane-tomlinson: Should there be a test about destroying a refresh token w/o an access token (maybe they all expired?), or is that implied here?

packages/fxa-content-server/app/scripts/views/settings/clients.js
```diff
@@ -286,31 +281,23 @@ const View = FormView.extend(
...
-            this.logEventOnce(itemsLog + '.two');
-          } else {
-            this.logEventOnce(itemsLog + '.many');
-          }
+      return this._attachedClients.fetchClients(this.user).then(() => {
+        // log the number of items
+        const numOfClients = this._attachedClients.length;
+        const itemsLog = 'settings.clients.items';
+
+        if (numOfClients === 0) {
```
shane-tomlinson: Are we making use of these events anywhere? I can't find any reference to them on the backend or fxa-shared, the only thing I see is a reference to [`settings.clients.disconnect.submit`](https://github.com/mozilla/fxa/blob/ae34698dee9370a411e88ab5e5361c965309ece6/packages/fxa-content-server/server/lib/amplitude.js#L175). This might be a remnant from DataDog days, and if so, might as well remove the events while you are digging around in here.

rfk:
> it seems like the fxa-js-client changes should have landed prior to this, unless you
> have figured out the magic incantation to make that not necessary. 
 
I definitely didn't figure out anything clever there; rebasing now..

packages/fxa-content-server/app/scripts/views/settings/clients.js
rfk: I went ahead and removed them.

rfk:
Thanks @shane-tomlinson, rebased and nits addressed.

shane-tomlinson:
@rfk, before merging, can you place a size estimate on this as defined in https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=CLOUDSERVICES&title=Sprint+Planning?

PR #1263, opened 2019-05-28
chore(deps): Update to the latest Backbone

shane-tomlinson:
A couple of tests needed to be updated
because:

* notifier-mixin was testing deep internals of how once was hooked up.
* Backbone.Model.destroy now calls `stopListening` itself, no need to do it again.

This is in support of adding Typescript. Backbone updated the way Views initialize that work better with ES2015 classes, which work better with Typescript than View.extend.

Not attached to an issue


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #1271, opened 2019-05-28
fix(scripts): use version sort to get last tag for train bump

### LINKED ISSUE #1270: Release script fails after tags are made on old trains
Currently the release script uses `git describe --tags --abbrev=0` to grab the last tag. This assumes the most recent tag chronologically is the one we want to bump from. If it isn't, the script fails. That's happening right now because `136.6` is more recent than any tags on `train-137`.

Maybe it's sufficient to just make this clearer in the comment at the top of the script, so that people always know to cut subsequent tags on anysubsequent branches. Or maybe we need to grab the last tag alphanumerically instead.

philbooth:
Fixes #1270.

Currently the release script fails when tagging `138.0` because the most recent tag chronologically is `136.6`. This changes it so that path uses git's version-sort ordering instead.

@mozilla/fxa-devs r?

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -76,7 +76,11 @@ if [ "$STATUS" != "" ]; then
 fi
 
 # 3. Find the last tag.
-LAST_TAG=`git describe --tags --abbrev=0`
+if [ "$BUILD_TYPE" = "Train" ]; then
+  LAST_TAG=`git tag -l --sort=version:refname | tail -1`
+else
+  LAST_TAG=`git describe --tags --abbrev=0`
+fi
 
 # 4. Check there have been some commits since the last tag.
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -76,7 +76,11 @@ if [ "$STATUS" != "" ]; then
 fi
 
 # 3. Find the last tag.
-LAST_TAG=`git describe --tags --abbrev=0`
+if [ "$BUILD_TYPE" = "Train" ]; then
+  LAST_TAG=`git tag -l --sort=version:refname | tail -1`
```
vladikoff:
```
git tag -l --sort=version:refname | tail -1
v1.137.4
```
👍

PR #1274, opened 2019-05-28
feat(script): script for reading security events from db

### LINKED ISSUE #1248: Complete the Firefox Account dashboard security script
For phase one of the GSoC project, the deliverable is to have a script that can query our database and pretty print out all the security events, session token data, and device information.

Please reference the [feature document](https://docs.google.com/document/d/1LNfU9wZx6PdkUxAkR95CWPWQtN7VYAxU253X69oN5ZI/edit#) for more details.

@farhan787 is currently on this issue. I'll tentatively target train-139, since that lines up with the original schedule for this feature.

farhan787:
fixes #1248


packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+    const devices = await readDevices(db)
+    securityEventsObject.devices = devices
+
+    const sessions = await readSessionTokens(db)
+    securityEventsObject.sessions = sessions
+
+    const events = await readSecurityEvents(db)
+    securityEventsObject.securityEvents = events
+
+    console.log(securityEventsObject) //eslint-disable-line no-console
```
shane-tomlinson: You are creating a log object above and using console.log here, let's use the same logger in both.

packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+
+    db.close()
+  })
+}
+
+function readDevices(db) {
```
shane-tomlinson: You could async/await these methods too for clarity.

packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+  let { uid } = user
+  uid = Buffer.from(uid, 'hex')
+
+  return db
+    .accountDevices(uid)
+    .then(devices => {
```
shane-tomlinson: This `.then` doesn't really do anything, tokens will be returned by the promise already.

packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+    })
+}
+
+function readSessionTokens(db) {
+  let { uid } = user
+  uid = Buffer.from(uid, 'hex')
+
+  return db
+    .sessions(uid)
+    .then(sessions => {
```
shane-tomlinson: This `.then` doesn't really do anything, sessions will be returned by the promise already.

packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+      process.exit()
+    })
+}
+
+function readSecurityEvents(db) {
+  const { uid, ipAddr } = user
+  const binaryUid = Buffer.from(uid, 'hex')
+
+  return db
+    .securityEvents({ id: binaryUid, ipAddr })
```
shane-tomlinson: Requiring an IP address is a good approach to get started, in the near future we are going to want *all* security events for a user, regardless of IP address. That will involve creating a new stored procedure that [acts similar to fetchSecurityEvents_1 but does not require `inIpAddr`](https://github.com/mozilla/fxa/blob/bd8ff27e8d9ec32b55bc43dfb8e179e752990bba/packages/fxa-auth-db-mysql/lib/db/schema/patch-032-033.sql#L56). 

Not related to anything you've done, I'm a bit surprised the current stored procedure has the [limit 50](https://github.com/mozilla/fxa/blob/bd8ff27e8d9ec32b55bc43dfb8e179e752990bba/packages/fxa-auth-db-mysql/lib/db/schema/patch-032-033.sql#L74) without providing a start index.

packages/fxa-auth-db-mysql/scripts/readSecurityEvents.js
```diff
@@ -0,0 +1,96 @@
...
+const config = require('../config')
+
+const securityEventsObject = {}
+
+// This input of user's data can be modified according to the usecase
+const user = {
+  // email: 'farhan@gmail.com',
+  // uid: '2A4A31EBAC9E46D9AD483B92789EC934',
+  uid: process.argv[2],
+  ipAddr: '127.0.0.1'
```
shane-tomlinson: Ah, now I see why this is needed, the [fetchSecurityEvents_1](https://github.com/mozilla/fxa/blob/bd8ff27e8d9ec32b55bc43dfb8e179e752990bba/packages/fxa-auth-db-mysql/lib/db/schema/patch-032-033.sql#L74) stored procedure requires it.

farhan787:
@shane-tomlinson I've also change `var` to `let` in mysql.js for methods.

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -662,23 +662,23 @@ module.exports = function (log, error) {
 
   // Delete : accountResetTokens
   // Where  : tokenId = $1
-  var DELETE_ACCOUNT_RESET_TOKEN = 'CALL deleteAccountResetToken_1(?)'
+  const DELETE_ACCOUNT_RESET_TOKEN = 'CALL deleteAccountResetToken_1(?)'
```
shane-tomlinson: You could refactor these `var` => `const` changes to make this PR slightly smaller.

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1364,6 +1364,16 @@ module.exports = function (log, error) {
       )
   }
 
+  const READ_SECURITY_EVENTS = 'CALL readSecurityEvents_1(?)'
+  MySql.prototype.getSecurityEvents = function (uid) {
+    return this.read(READ_SECURITY_EVENTS, [uid])
+      .then(
```
shane-tomlinson: You could even collapse this to `.then(result => result[0])`

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1364,6 +1364,16 @@ module.exports = function (log, error) {
       )
   }
 
+  const READ_SECURITY_EVENTS = 'CALL readSecurityEvents_1(?)'
```
shane-tomlinson: The distinction between `READ_SECURITY_EVENTS` and `FETCH_SECURITY_EVENTS` or `readSecurityEvents` and `fetchSecurityEvents` isn't really clear. We could rename `fetchSecurityEvents_1` for clarity, but it might be easier to call `readSecurityEvents_1` to something like `fetchSecurityEventsByUid_1`. Then at least the two are obviously related, but the new one says what it's fetching by.

packages/fxa-auth-db-mysql/lib/db/schema/patch-099-100.sql
```diff
@@ -0,0 +1,22 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('99');
+
+CREATE PROCEDURE `readSecurityEvents_1` (
```
shane-tomlinson: See the note above about renaming this to `fetchSecurityEventsByUid_1` for clarity and to easily distinguish it from the existing `fetchSecurityEvents`

packages/fxa-auth-db-mysql/lib/db/schema/patch-099-100.sql
```diff
@@ -0,0 +1,22 @@
...
+        n.name,
+        e.verified,
+        e.createdAt
+    FROM
+        securityEvents e
+    LEFT JOIN securityEventNames n
+        ON e.nameId = n.id
+    WHERE
+        e.uid = inUid
+    ORDER BY e.createdAt DESC;
```
shane-tomlinson: I suspect this list can end up being pretty big, I wonder whether we should limit by 50 like the other version until we better understand the behavior of the overall feature.

packages/fxa-auth-db-mysql/lib/db/schema/patch-100-099.sql
```diff
@@ -0,0 +1,5 @@
+-- SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+-- DROP PROCEDURE `readSecurityEvents_1`;
```
shane-tomlinson: 👍

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,93 @@
...
+      log.error(err.stack || err.message || err)
+      process.exit()
+    }
+  })
+}
+
+async function readDevices(db) {
+  try {
+    let { uid } = user
+    uid = Buffer.from(uid, 'hex')
```
shane-tomlinson: `Buffer.from(uid, 'hex')` is repeated 3 times, maybe just set user.uid to that version everywhere.

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,93 @@
...
+  stat: console.log,
+  info: console.log
+} //eslint-disable-line no-console
+const DB = require('../lib/db/mysql')(log, require('../db-server').errors)
+const config = require('../config')
+
+const securityEventsObject = {}
+
+// This input of user's data can be modified according to the usecase
+const user = {
```
shane-tomlinson: For the future what might be nice for the script is to be able to fetch by email OR uid. For the future, don't worry about it now.

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1364,6 +1364,16 @@ module.exports = function (log, error) {
...
+  MySql.prototype.getSecurityEvents = function (uid) {
```
shane-tomlinson: Going along with the above theme, the difference between `securityEvents` and `getSecurityEvents` isn't clear, let's rename this to `securityEventsByUid`.

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1364,6 +1364,16 @@ module.exports = function (log, error) {
...
+    return this.read(READ_SECURITY_EVENTS, [uid])
```
shane-tomlinson: We'll need a corresponding method [for the in-memory database](https://github.com/mozilla/fxa/blob/9ca5c4057cde5da1e2866cb9515e88bb18e5fb2b/packages/fxa-auth-db-mysql/lib/db/mem.js#L1070) too. The in-memory database is what we normally use while developing and allows us to rapidly test the API w/o requiring MySQL.

packages/fxa-auth-db-mysql/lib/db/mysql.js
farhan787: I followed the style of other methods, I'll do it your way.

packages/fxa-auth-db-mysql/lib/db/schema/patch-099-100.sql
farhan787: But it will fetch for only 1 user's uid, so how can the list get big ?

packages/fxa-auth-db-mysql/lib/db/mysql.js
farhan787: > You could even collapse this to `.then(result => result[0])`

I think it's better to use ES5 function syntax rather than ES6 arrow functions like other methods.

packages/fxa-auth-db-mysql/lib/db/schema/patch-099-100.sql
shane-tomlinson: I think so, let's start there.

packages/fxa-auth-db-mysql/lib/db/mysql.js
shane-tomlinson: There are some methods below that use fat arrows. I'm all for updating/improving code as we go along.

farhan787:
@shane-tomlinson I've made the changes, could you please check out the new changes.

farhan787:
@shane-tomlinson I've added the tests, would you mind reviewing it ?

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,85 @@ module.exports = function (config, DB) {
...
+            assert.isBelow(results[2].createdAt, results[1].createdAt)
+          })
+      })
+
+      it('should get event after session verified', () => {
+        return db.verifyTokens(session1.tokenVerificationId, { uid })
+          .then(() => db.securityEventsByUid(uid))
+          .then((results) => {
+            assert.lengthOf(results, 3)
+            assert.isTrue(!! results[0].verified)
```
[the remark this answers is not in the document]
shane-tomlinson: The answer is yes, they are needed because `verified` is a number rather than a boolean.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,85 @@ module.exports = function (config, DB) {
...
+        return db.createSecurityEvent({
+          uid: uid,
+          ipAddr: addr,
+          name: name,
+          tokenId: session
+        })
+      }
+
+      beforeEach(() => {
+        session1 = makeMockSessionToken(accountData.uid)
```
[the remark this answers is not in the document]
farhan787: Nice catch @shane-tomlinson

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,85 @@ module.exports = function (config, DB) {
...
+          db.createSessionToken(session3.tokenId, session3)
+        ])
+        // Don't paralleize these, the order of them matters
+        // because they record timestamps in the db.
+          .then(() => insert(uid, addr, evA, session2.tokenId).delay(10))
+          .then(() => insert(uid, addr, evB, session1.tokenId).delay(10))
+          .then(() => insert(uid, addr, evC).delay(10))
+      })
+
+      it('should get security event', () => {
```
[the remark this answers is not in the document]
farhan787: IMHO, I think going with the module style is good for now.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,85 @@ module.exports = function (config, DB) {
...
+            assert.equal(results[1].name, evB, 'correct event name')
+            assert.equal(!! results[1].verified, false, 'second session is not verified yet')
+            assert.isBelow(results[1].createdAt, results[0].createdAt)
+            assert.equal(results[2].name, evA, 'correct event name')
+            assert.equal(!! results[2].verified, true, 'first session is already verified')
+            assert.isBelow(results[2].createdAt, results[1].createdAt)
+          })
+      })
+
+      it('should get event after session verified', () => {
```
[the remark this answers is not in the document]
farhan787: It's returning the events after the session token is verified like in the same test for db.securityEvents test suite above.

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,24 @@ module.exports = function (log, error) {
...
+
+    return P.resolve(events.filter(function (ev) {
+      return ev.uid.toString('hex') === key
+    }).map(function (ev) {
+      return {
+        name: ev.name,
+        createdAt: ev.createdAt,
+        verified: ev.verified
+      }
+    }).reverse())
```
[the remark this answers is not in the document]
farhan787: Done

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,103 @@ module.exports = function (config, DB) {
...
+        const anotherAccountData = createAccount()
+        db.createAccount(anotherAccountData.uid, anotherAccountData)
+
+        session1 = makeMockSessionToken(accountData.uid)
+        session2 = makeMockSessionToken(accountData.uid)
+        // Make session verified
+        delete session2.tokenVerificationId
+
+        session3 = makeMockSessionToken(accountData.uid)
+        anotherAccountSession = makeMockSessionToken(anotherAccountData.uid)
```
shane-tomlinson: Perfect!

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,33 @@ module.exports = function (log, error) {
...
+    }
+    var key = uid.toString('hex')
+    var events = securityEvents[key] || []
+    var securityEventsLimit = 50
+
+    var filteredSecurityEvents = events.filter(function (ev) {
+      return ev.uid.toString('hex') === key
+    }).reverse()
+
+    if (filteredSecurityEvents.length > securityEventsLimit) {
```
shane-tomlinson: This will update the `length` property of filteredSecurityEvents but won't actually limit the array size to 50. 

An alternative is to use 2 filters to ensure there are only 50 items in the list. Look up the the function signature of events.filter, the callback takes multiple arguments, the 2nd is an index.

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,38 @@ module.exports = function (log, error) {
...
+      return P.reject(error.notFound())
+    }
+    var key = uid.toString('hex')
+    var events = securityEvents[key] || []
+    var securityEventsLimit = 50
+
+    function filterEvents (events, eventsLimit, key) {
+      var filteredEvents = []
+
+      for (var i = 0; i < events.length; i++) {
```
farhan787: @shane-tomlinson I've used here vanilla for loop and not filter method with the index concept because of this reason

```js
// Suppose if events is a list of numbers and I want 2 events which are greater than 4, 
// then this index concept will only return me an empty array though [5, 6] should be returned
// and there's no method that can stop filtering events when it has 50 reached events 
// that's why I made this custom function.

var events = [1, 2, 3, 4, 5, 6, 7, 8]
var eventsLimit = 2
events.filter((ev, index) => {
    if(index < eventsLimit)
        return ev > 4
})
```

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,38 @@ module.exports = function (log, error) {
...
+        if (filteredEvents.length === eventsLimit) break
```
shane-tomlinson: Nit: put the `break` on the next line, use `{}`.

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,38 @@ module.exports = function (log, error) {
...
+        if (events[i].uid.toString('hex') === key)
```
shane-tomlinson: nit: use `{}` to surround line 1103.

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+// Usage: node read-security-events.js uid
+
+const log = { trace: console.log, error: console.error, stat: console.log, info: console.log } //eslint-disable-line no-console
+const DB = require('../lib/db/mysql')(log, require('../db-server').errors)
+const config = require('../config')
+
+const securityEventsObject = {}
+const scriptInputUid = process.argv[2]
```
shane-tomlinson: Can you log an error message and exit if the uid is not specified? Otherwise this is displayed:

```bash
➜  scripts git:(farhan-gsoc) node read-security-events.js                                                                                                                                  git:(farhan-gsoc|)
buffer.js:207
    throw new ERR_INVALID_ARG_TYPE(
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined
    at Function.from (buffer.js:207:11)
    at Object.<anonymous> (/Users/shanetomlinson/development/fxa/packages/fxa-auth-db-mysql/scripts/read-security-events.js:23:15)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)
```

packages/fxa-auth-db-mysql/lib/db/mem.js
shane-tomlinson: That's fine, nit on the `var`, use `let` or `const` instead.

packages/fxa-auth-db-mysql/scripts/read-security-events.js
farhan787: @shane-tomlinson done

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function readEvents() {
+  DB.connect(config).then(async db => {
+    try {
+      const devices = await readDevices(db)
+      securityEventsObject.devices = devices
+
+      const sessions = await readSessionTokens(db)
+      securityEventsObject.sessions = sessions
```
shane-tomlinson: I think both devices and sessions can contain buffers, for example in sessions both tokenId and uid are buffers. Can you convert those to hex before printing those to the console? Otherwise it looks like a bunch of gobledygook:

Screenshot 2019-06-10 at 17 06 03


In addition, can you convert dates from an integer to a human readable format? Perhaps show both. Current output below:

Screenshot 2019-06-10 at 17 06 30

farhan787: Done

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,102 @@ module.exports = function (config, DB) {
...
+          uid: uid,
+          ipAddr: addr,
+          name: name,
+          tokenId: session
+        })
+      }
+
+      beforeEach( () => {
+        const anotherAccountData = createAccount()
+        db.createAccount(anotherAccountData.uid, anotherAccountData)
```
vbudhram: `db.createAccount` returns a promise, to avoid race conditions do something like https://github.com/mozilla/fxa/blob/ebd32d7eccd804d8a9ef359f04b15ac9509811f1/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js#L1816.

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,41 @@ module.exports = function (log, error) {
     }).reverse())
   }
 
+  Memory.prototype.securityEventsByUid = function (uid) {
+    if (! uid) {
+      return P.reject(error.notFound())
+    }
+    const key = uid.toString('hex')
```
vbudhram: I would just reassigned uid buffer to uid string here. It makes it a little bit easier to follow the code than using `key`.

```
uid = uid.toString('hex')
```

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -1086,6 +1086,41 @@ module.exports = function (log, error) {
...
+    const events = securityEvents[key] || []
+    const securityEventsLimit = 50
```
vbudhram: Lets pull this to the top of the file similar to https://github.com/mozilla/fxa/blob/ebd32d7eccd804d8a9ef359f04b15ac9509811f1/packages/fxa-auth-db-mysql/lib/db/mem.js#L57.

I think we can make this limit 100.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -1474,6 +1474,102 @@ module.exports = function (config, DB) {
...
+          })
+      })
+
+      it('should give no securityEvent with new unknown uid', () => {
+        const unknownUid = newUuid()
+        return db.securityEventsByUid(unknownUid)
+          .then( results => assert.lengthOf(results, 0))
+      })
+
+      it('should delete events when account is deleted', () => {
```
vbudhram: We should be testing `should delete events`. I would modify this to call `db.deleteSecurityEventsByUid`. It looks like that implementation is missing for the memory database.

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,164 @@
...
+  try {
+    const { uid } = user
+    const devices = await db.accountDevices(uid)
+
+    if (devices.length) {
+      for (const d of devices) {
+        d.id = d.id.toString('hex')
+        d.uid = d.uid.toString('hex')
+        d.sessionTokenId = d.sessionTokenId.toString('hex')
+        d.refreshTokenId = d.refreshTokenId.toString('hex')
```
vbudhram: d.refreshTokenId is not always guaranteed to exist. I don't think you need to include it.

packages/fxa-auth-db-mysql/scripts/read-security-events.js
```diff
@@ -0,0 +1,164 @@
...
+      const devices = await readDevices(db)
+      securityEventsObject.devices = devices
+
+      const sessions = await readSessionTokens(db)
+      securityEventsObject.sessions = sessions
+
+      const events = await readSecurityEvents(db)
+      securityEventsObject.securityEvents = events
+
+      log.info(securityEventsObject)
```
vbudhram: While this does work, I think we can print this out a little better. Perhaps in a table or something a little easier to read.

farhan787:
@vbudhram I resolved the merge conflicts!!

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -2537,6 +2537,102 @@ module.exports = function(config, DB) {
...
+      it('should give no securityEvent with new unknown uid', () => {
+        const unknownUid = newUuid()
+        return db.securityEventsByUid(unknownUid)
+          .then(results => assert.lengthOf(results, 0))
+      })
+
+      it('should get no events when events are deleted', () => {
+        return db.deleteSecurityEventsByUid(accountData.uid)
+          .then(() => db.securityEventsByUid(uid))
+          .then(res => assert.lengthOf(res, 0))
```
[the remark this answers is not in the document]
farhan787: but `securityEventsByUid(uid)` returns a list of events, so shouldn't we assert length of that list ?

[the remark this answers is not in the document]
vbudhram: Oh I see, you can keep the assert for length, but also assert the response from `deleteSecurityEventsByUid`.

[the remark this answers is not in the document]
farhan787: Done :-)

Issue #1160, opened 2019-05-20
Add configuration to payments index.html


ianb:
This is a followup to #1133 and specifically https://github.com/mozilla/fxa/pull/1133#discussion_r284920190

We should inject some configuration into the HTML page so it's loadable by [config-loader](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/config-loader.js):

```html
<meta name="fxa-content-server/config" content="URI-encoded JSON">
```

ianb:
Fixed (mostly) in #1275. #1280 will complete the work.

PR #1275, opened 2019-05-28
feat(payments): add configuration to payments index.html

ianb:
This injects some whitelisted config values into <meta name=fxa-content-server/config> (and a stubbed feature flag tag). The misnomer is to allow a shared library to load that config.

The config is both injected into the development server (which is proxied to) and the production server (which is based on a static file)

Also this adds routes to the production server, which should serve the standard index.html on all routes (e.g., /subscriptions).

Some followups that would be called for:
* Move `config-loader.js` into shared
* Actually read the config!
* Add feature flags
* Add ETag to the production index.html serving

### DIFF
--- a/packages/fxa-payments-server/public/index.html
+++ b/packages/fxa-payments-server/public/index.html
@@ -10,6 +10,8 @@
     <meta name=robots content=noindex,nofollow>
     <meta name=viewport content="width=device-width,initial-scale=1,maximum-scale=2,user-scalable=yes">
     <meta name=apple-itunes-app content="app-id=989804926, affiliate-data=ct=smartbanner-fxa">
+    <meta name="fxa-content-server/config" content="__SERVER_CONFIG__">
+    <meta name="fxa-feature-flags" content="__FEATURE_FLAGS__">
 
     <script src="https://js.stripe.com/v3/"></script>
 </head>
--- a/packages/fxa-payments-server/server/lib/server.js
+++ b/packages/fxa-payments-server/server/lib/server.js
@@ -6,6 +6,7 @@
 
 module.exports = () => {
   const path = require('path');
+  const fs = require('fs');
 
   // setup version first for the rest of the modules
   const logger = require('./logging/log')('server.main');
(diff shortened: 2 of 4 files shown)


packages/fxa-payments-server/public/index.html
```diff
@@ -10,6 +10,8 @@
     <meta name=robots content=noindex,nofollow>
     <meta name=viewport content="width=device-width,initial-scale=1,maximum-scale=2,user-scalable=yes">
     <meta name=apple-itunes-app content="app-id=989804926, affiliate-data=ct=smartbanner-fxa">
+    <meta name="fxa-content-server/config" content="__SERVER_CONFIG__">
+    <meta name="fxa-feature-flags" content="__FEATURE_FLAGS__">
```
ianb: I picked arbitrary strings, I don't know if there's any special pattern. I didn't think we should use Mustache here because it's a rendered page, not a template.

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -22,6 +23,21 @@ module.exports = () => {
 
   const app = express();
 
+  // Each of these config values (e.g., 'servers.content') will be exposed as the given
+  // variable to the client/browser (via fxa-content-server/config)
+  const CLIENT_CONFIG_MAP = {
+    contentUrl: 'servers.content',
+    oAuthUrl: 'servers.oauth',
+    profileUrl: 'servers.profile',
+  };
```
ianb: I copied the names from https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/routes/get-index.js#L38-L55 – but probably it would be more sensible to just create a partial copy of the object in a naive way, like:

```js
const CLIENT_CONFIG = {
  servers: {
    content: config.get('servers.content'),
    oauth: config.get('servers.oauth'),
    profile: config.get('servers.profile'),
  }
};
```

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -22,6 +23,21 @@ module.exports = () => {
...
+
+  // This is a list of all the paths that should resolve to index.html:
+  const INDEX_ROUTES = [
+    '/',
+    '/subscriptions',
+    '/products/:productId',
+  ];
```
ianb: This needs to stay in sync with React Routes, but I didn't do anything to accomplish that beyond comments

lmorchard: There might be a way to do this programmatically, but heck if I know how off the top of my head

lmorchard:
Oh, but for follow ups - do you want to leave #1160 open? Or file as separate follow-up issues? Should make sure we don't lose track of them

ianb:
I've made followups, the only critical one being #1280 (which I put in the next train). I'm guessing we'll backlog the others.

packages/fxa-payments-server/server/lib/server.js
ianb: I guess a shared file they both load that lists the routes, but I think that cure is worse than the disease

packages/fxa-payments-server/server/lib/server.js
ianb: I might do this in #1280

Issue #1267, opened 2019-05-28
[fxa-event-broker] Add deployment integration


bbangert:
Add .circleci/Dockerfile, CI integration, release updates.

bbangert:
Also should include Sentry for error logging.

bbangert:
Closed by #1334.

PR #1287, opened 2019-05-29
feat(event-broker): add database tests

### LINKED ISSUE #1266: [fxa-event-broker] Database tests
Add tests for the database functionality.

bbangert:
Adds local google cloud pubsub/firestore emulator and rigs tests into
it.

Issue #1267

Closes #1266

### DIFF
--- a/_scripts/check.sh
+++ b/_scripts/check.sh
@@ -46,3 +46,15 @@ elif [ "$os" = "Linux" ]; then
       echo "sudo service docker restart"
     fi
 fi
+
+if [[ $(which gcloud) && $(gcloud --version) ]]; then
+  gcloud=y
+else
+  gcloud=n
+fi
+
+if [ "$gcloud" = "n" ]; then
+  echo "install gcloud to continue installation"
+  echo "https://cloud.google.com/sdk/"
+  exit 1
+fi
--- a/mysql_servers.json
+++ b/mysql_servers.json
@@ -221,6 +221,41 @@
         "LOGGING_FORMAT": "pretty",
         "NODE_ENV": "development"
       }
+    },
+    {
+      "name": "event-broker",
+      "script": "npm",
+      "args": [
+        "run",
+        "start-dev"
+      ],
+      "cwd": "packages/fxa-event-broker",
+      "max_restarts": "1",
+      "env": {
+        "NODE_ENV": "development",
+        "WORKER_HOST": "0.0.0.0"
+      },
+      "min_uptime": "2m"
+    },
+    {
+      "name": "google-pubsub-emulator PORT 8005",
+      "script": "gcloud",
+      "interpreter": "bash",
+      "args": [
+        "beta", "emulators", "pubsub", "start", "--host-port=localhost:8005"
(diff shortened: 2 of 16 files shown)


bbangert:
The `package.json` changed so much mainly from adding `@firebase/testing` which dropped in 246 packages. Annoying, but :man_shrugging:

mysql_servers.json
```diff
@@ -185,6 +185,41 @@
       "max_restarts": "1",
       "min_uptime": "2m"
     },
+    {
```
philbooth: I think that, as a general rule, we usually prefer to append new services to the end of this JSON file, so that the `pm2` ordinal shortcuts for everything else stay the same and people's muscle memory isn't interrupted. But this is probably close enough to the end not to matter, I doubt many people have internalised the numbers for the items that follow these.

Also, in this case it leads to the ordinals for other things being different depending on whether `servers.json` or `mysql_servers.json` is used. I think we try to avoid that too.

bbangert: Ah, makes sense. Should I add this to `servers.json` as well? I had totally forgotten about that one, do people regularly use that over `mysql_servers.json`?

_scripts/check.sh
```diff
@@ -46,3 +46,15 @@ elif [ "$os" = "Linux" ]; then
...
 fi
+
+if [[ $(which gcloud) && $(gcloud --version) ]]; then
+  gcloud=y
+else
+  gcloud=n
+fi
+
+if [ "$gcloud" = "n" ]; then
+  echo "install gcloud to continue installation"
```
vladikoff: you could also automate the installation for osx and / or darwin based on instructions in https://cloud.google.com/sdk/docs/quickstart-linux and https://cloud.google.com/sdk/docs/quickstart-macos

mysql_servers.json
philbooth: > Should I add this to `servers.json` as well? I had totally forgotten about that one, do people regularly use that over `mysql_servers.json`?

I still use `servers.json` a lot of the time, but only from habit formed before we had the MySQL one. Maybe we should get rid of it, but as long as it still exists it's worth keeping them both updated I guess?

_scripts/check.sh
bbangert: I had considered that, but I don't have an OSX box to test it with, and their d/l links are pegged to versions which move fairly quickly so then we'd have to keep updating it.

packages/fxa-event-broker/bin/workerDev.ts
```diff
@@ -0,0 +1,88 @@
...
+delete firestoreConfig.enabled;
+
+let db = firestoreEnabled
+  ? createDatastore(FirestoreDatastore, firestoreConfig)
+  : createDatastore(InMemoryDatastore, {});
+
+async function main() {
+  AWS.config.update({
+    accessKeyId: 'fake',
+    ['endpoint' as any]: 'localhost:4100',
```
bbangert: Typescript 'escape hatch' to reference properties that aren't public.

PR #1296, opened 2019-05-29
feat(signup): Verify signups with a TOTP-like code 

### LINKED ISSUE #457: Feat: Sign up verification codes instead of verification links
The current signup verification flow requires a user to exit the FxA flow and open a verification link from their email.

This causes several problems:

* Users often open verification links in a different tab/browser, causing an explosion of paths we have to worry about. @ryanfeeley and I had a spreadsheet tracking the combinatoric explosion. It's awful. 
* Users sometimes never get back to the app they were trying to sign up for because the links opens in a browser.
* Mobile devices may kill apps that go into the background.

For Lockbox, we ask users that are *signing in* to copy a short "confirmation code" from their email into a screen on FxA to ensure that users make it back to FxA to sign in. Perhaps @irrationalagent has some details on signin performance of codes vs links that can help guide prioritization.

@rfk wrote some comments in an email that I'm copying here:

> For the current email-codes experiment on login, we maintain a separate "email code" column in the db with the shorter code. This is nice and simple but it has a couple of drawbacks.  Chiefly, we haven't implemented the ability to regenerate this code once it expires, so any attempts to re-send the verification email will fall back to sending a link rather than a code.
>
> Taking the same approach for verifying sign-up would mean adding new "emailShortCode" and "emailShortCodeExpiry" columns to the 'accounts' and/or 'emails' table, and probably figuring out a way to safely regenerate these when they expire.
>
> However, we already know another technique for turning a high-entropy long-lived secret into a short time-limited code: TOTP.
>
> If we treated the current email link code as a TOTP master secret, we could easily generate and verify short email confirmation codes in a way that doesn't require new db columns and doesn't require any extra logic to handle regeneration after expiry.

@rfk has also expressed an interest in getting this for Fenix.

cc @davismtl, @ryanfeeley

shane-tomlinson:
Super early draft from last month. Opening it in its sorry state because I couldn't find the patch once and don't want to lose it again.

fixes #457

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -111,6 +111,7 @@ function createServer(db) {
       return db.verifyEmail(req.params.id, req.params.emailCode)
     })
   )
+  api.post('/account/:id/verify', withIdAndBody(db.verifyAccount))
   api.post('/account/:id/locale', withIdAndBody(db.updateLocale))
   api.get('/account/:id/sessions', withIdAndBody(db.sessions))
 
--- a/packages/fxa-auth-db-mysql/lib/db/mem.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mem.js
@@ -403,6 +403,7 @@ module.exports = function (log, error) {
     if (tokenCount === 0) {
       return P.reject(error.notFound())
     }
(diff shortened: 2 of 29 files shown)


shane-tomlinson:
Closing my version, @vbudhram has taken over this work and he has opened his work in #2071 and #2177

PR #1304, opened 2019-05-29
fix: install gcloud for circleci

bbangert:
I'll admit I don't entirely understand why 'install-content-server' actually installs everything in the FxA repo, but it does, so here we are.

### DIFF
--- a/.circleci/install-content-server.sh
+++ b/.circleci/install-content-server.sh
@@ -3,6 +3,10 @@
 DIR=$(dirname "$0")
 
 if grep -e "fxa-content-server" -e 'all' $DIR/../packages/test.list; then
+  CLOUD_SDK_REPO="cloud-sdk-$(grep VERSION_CODENAME /etc/os-release | cut -d '=' -f 2)"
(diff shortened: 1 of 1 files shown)


.circleci/install-content-server.sh
```diff
@@ -3,6 +3,10 @@
 DIR=$(dirname "$0")
 
 if grep -e "fxa-content-server" -e 'all' $DIR/../packages/test.list; then
+  CLOUD_SDK_REPO="cloud-sdk-$(grep VERSION_CODENAME /etc/os-release | cut -d '=' -f 2)"
+  echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+  curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+  sudo apt-get update -y && sudo apt-get install google-cloud-sdk -y
```
vladikoff: I wonder if `sudo apt-get update ` will slow things down on the install even more :\

bbangert: It seemed to run pretty quick on circleci when I was watching it, took maybe 2 seconds. Granted that's in the ideal case where the network is behaving itself.

vladikoff: thanks for checking!

PR #1311, opened 2019-05-30
fix(scripts): stop the release script getting hung up on colour codes

philbooth:
@shane-tomlinson can you give this a try locally to see if it runs to completion for you? Works for me locally, so 🤞.


philbooth:
> sorry to drag you into this

Don't be silly, I wrote the thing! More than happy to help debug/fix it!

PR #1300, opened 2019-05-29
fix(tests): switch to a better QR reader for pairing tests

vladikoff:
Ref https://github.com/mozilla/fxa/issues/1295

Issue #1295, opened 2019-05-29
The failing circle tests are really annoying


shane-tomlinson:
At least 50% of my Circle test runs fail for what appear to be random and unrelated reasons. Sometimes a session can't be created, sometimes the CSS doesn't seem to be built, sometimes the pairing tests fail because the QR code reader can't find enough points.

The pairing tests are by far the most common failures, perhaps 25% of the time.

Whatever the reasons, it's really annoying.

cc @vladikoff for the pairing tests.

philbooth:
See also: #1188

vladikoff:
Was Firefox updated at all?

vladikoff:
WRT to pairing tests, replacing the QR scanning library should do the trick, or if we find a way to make scanning more consistent

vladikoff:
>  replacing the QR scanning library should do the trick

Big bet on this: https://github.com/mozilla/fxa/pull/1300

clouserw:
I think this has been addressed...  If not, please open specific issues.  Thanks.

PR #1312, opened 2019-05-30
fix(tests): switch to a better QR reader for pairing tests

shane-tomlinson:
Ref https://github.com/mozilla/fxa/issues/1295

Originally opened by @vladikoff as https://github.com/mozilla/fxa/pull/1300, opening against train-138 to help those tests.


shane-tomlinson:
@philbooth - r?

PR #1313, opened 2019-05-30
feat(build): Add typescript to the content server

shane-tomlinson:
This converts verification-reasons and verification-methods
to typescript. These modules are super simple and can be
converted to enums, a basic data type.

This PR is to prove that adding typescript to the build
works.


shane-tomlinson:
Changing this to ready for review! 

@mozilla/fxa-devs - r?

In particular, @lmorchard, @bbangert, and @ianb might have some feedback from their recent work with Typescript.

packages/fxa-content-server/webpack.config.js
```diff
@@ -45,11 +45,12 @@ const webpackConfig = {
   },
 
   resolve: {
-    extensions: ['.js'],
+    extensions: ['.ts', '.js'],
     modules: [
       path.resolve(__dirname, 'app/scripts'),
       path.resolve(__dirname, 'app/scripts/templates'),
       path.resolve(__dirname, 'app'),
+      path.resolve(__dirname, '.tscompiles/scripts'),
```
lmorchard: Typo here? `s/tscompiles/tscompiled/`

packages/fxa-content-server/app/scripts/lib/verification-methods.ts
```diff
@@ -8,9 +8,11 @@
  * In the future SMS and TOTP could also be included.
  */
 
-export default {
-  EMAIL: 'email',
-  EMAIL_2FA: 'email-2fa',
-  EMAIL_CAPTCHA: 'email-captcha',
-  TOTP_2FA: 'totp-2fa'
+enum VerificationMethods {
```
bbangert: The Programming Typescript book recommends using const enum's to opt into a stricter form of enums that doesn't allow key lookup that may be unsafe.

packages/fxa-content-server/webpack.config.js
shane-tomlinson: Huh, turns out this isn't needed at all!

packages/fxa-content-server/app/scripts/lib/verification-methods.ts
shane-tomlinson: Updated.

shane-tomlinson: Oh, we have to do key lookup, I reverted that change.

bbangert:
LGTM!

PR #1334, opened 2019-06-01
feat(event-broker): add GCP pub/sub storing of subscription changes

### LINKED ISSUE #1268: [fxa-event-broker] GCP Pub/Sub Storing
Implement:
- Fetch OAuth clients/capabilities from fxa-auth-server. Automatically refresh this list periodically.
- Fetch RP clientId:webhook URL mapping from firestore
- Use Capability list with subscribe service notification to determine what pub/sub topics to store the new message in.
- Store the message in the appropriate topics.

bbangert:
Also adds:
* Caching abstraction for automatically refreshed data.
* Cached services for webhook urls and the client capabilities.
* Dockerfile for CI integration.

Closes #1268, #1267

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,6 +79,9 @@ jobs:
             - run:
                 name: Start redis
                 command: docker pull redis && docker run -d --name redis-server -p 6379:6379 redis
+            - run:
+                name: Start gcloud firestore
+                command: docker pull jdlk7/firestore-emulator && docker run -d --name gcloud-firestore -p 8006:9090 jdlk7/firestore-emulator
 
       - run: ../../.circleci/build-test-deploy.sh << parameters.test >>
 
@@ -293,6 +296,12 @@ workflows:
           requires:
             - install
       - build-module:
+          name: fxa-event-broker
+          module: fxa-event-broker
+          db: true
+          requires:
+            - install
+      - build-module:
           name: fxa-oauth-console
           module: fxa-oauth-console
           requires:
@@ -408,6 +417,16 @@ workflows:
               ignore: /.*/
           name: fxa-email-service
           module: fxa-email-service
+          requires:
+            - install
+      - deploy-module:
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              ignore: /.*/
+          name: fxa-event-broker
+          module: fxa-event-broker
           requires:
             - install
       - deploy-module:
--- a/_scripts/firestore.sh
+++ b/_scripts/firestore.sh
@@ -7,20 +7,16 @@
 # the script running.
 
 function on_sigint() {
-  echo "MySQL shutting down."
-  docker stop mydb
+  echo "Firestore Emulator shutting down."
+  docker stop firestore
   exit 0
 }
 
 trap on_sigint INT
 
 # Create pushbox db on start (because pushbox doesn't create it)
-docker run --rm --name=mydb \
-  --network fxa-net \
-  -e MYSQL_ALLOW_EMPTY_PASSWORD=true \
-  -e MYSQL_ROOT_HOST=% \
-  -e MYSQL_DATABASE=pushbox \
-  -p 3306:3306 \
-  mysql/mysql-server:5.6 &
+docker run --rm --name=firestore \
+  -p 8006:9090 \
+  jdlk7/firestore-emulator &
 
 while :; do read; done
--- a/_scripts/pubsub.sh
+++ b/_scripts/pubsub.sh
@@ -7,20 +7,16 @@
 # the script running.
 
 function on_sigint() {
-  echo "MySQL shutting down."
-  docker stop mydb
+  echo "PubSub Emulator shutting down."
+  docker stop pubsub
   exit 0
 }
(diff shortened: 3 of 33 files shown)


packages/fxa-event-broker/.prettierrc
```diff
@@ -1,6 +1,6 @@
 {
   "useTabs": false,
-  "printWidth": 80,
+  "printWidth": 100,
```
bbangert: 80 was a bit narrow, and most of the FxA code seemed to hew to 100, so I upped it for TypeScript here.

packages/fxa-event-broker/lib/selfUpdatingService/index.ts
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
bbangert: This service should probably be in fxa-shared as I could see the pattern being useful elsewhere. There's a different version of this I've seen in some Python caching libraries that implements some logic that only does the fetch on-demand (vs background interval like here), which prevents concurrent fetches and anyone after the refresh trigger gets the prior cached value. An abstraction for that pattern would also be handy to have in fxa-shared.

packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -20,29 +22,36 @@ class FirestoreDatastore implements Datastore {
...
   }
 
   public async fetchClientIds(uid: string): Promise<string[]> {
     const document = this.db.doc(`users/${uid}`);
     const doc = await document.get();
     if (doc.exists) {
-      const data: any = doc.data();
-      if (data.oauth_clients) {
+      const data = doc.data();
+      if (data && data.oauth_clients) {
```
bbangert: Google API's have types already, so casting to any lost important type data, such as doc.data() possibly returning undefined, which is now tested for.

.circleci/config.yml
```diff
@@ -182,6 +185,17 @@ jobs:
           at: ~/
       - run: ../../.circleci/test-js-client.sh
 
+  fxa-event-broker:
```
dannycoates: i dont think you need this job. `build-module` does this

_scripts/firestore.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -ex
+
+# This docker image doesn't react to SIGINTs (Ctrl+C) which is used by
+# pm2 to kill processes.
+# We go around this by defining a SIGINT handler, running docker in the
+# background (but still logging on the screen), and running a read to keep
+# the script running.
+
+function on_singint() {
```
philbooth: Typo in the function name here, although a song about ctrl+c sounds pretty cool too. 😄

bbangert: That's what I get for copy/pasting from mysql.sh. I'll fix them all.

_scripts/pubsub.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -ex
+
+# This docker image doesn't react to SIGINTs (Ctrl+C) which is used by
+# pm2 to kill processes.
+# We go around this by defining a SIGINT handler, running docker in the
+# background (but still logging on the screen), and running a read to keep
+# the script running.
+
+function on_singint() {
```
philbooth: 🎤 🎶

packages/fxa-event-broker/lib/result.ts
```diff
@@ -0,0 +1,12 @@
+/* Pattern captured from blog post: https://spin.atomicobject.com/2018/02/23/error-either-typescript/
+ */
+
+export type Result<T> = T | Error;
```
philbooth: Nice, if I squint my eyes a bit I can pretend we're using Rust.

bbangert: If you really want to pretend its Rust, there's even https://github.com/bcherny/tsoption ;)

packages/fxa-event-broker/lib/selfUpdatingService/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // Verify we can do the initial fetch
+    const result = await this.updateFunction();
+    if (isError(result)) {
+      this.logger.error('clientCapabilityService', { err: result });
+      throw result;
+    }
+    this.data = result;
+    this.timer = setInterval(async () => await this.updateService(), this.refreshInterval);
```
philbooth: I'm just splitting hairs here so no need to change anything, but if I was implementing this I'd probably call `setTimeout` inside `updateService` instead of using `setInterval` here.

It would require an extra bit of state for sane cancellation, but would prevent multiple concurrent handlers from being spawned in the event that `this.updateFunction` is badly behaved and doesn't fulfil it's promise for some reason (i.e. potentially leaky).

bbangert: I was contemplating adding a bit more around the updateService call here to at least log exceptions in the updateFunction so that this would always wrap up correctly. I'll add that in a separate PR.

Issue #1254, opened 2019-05-24
Mobile view for device list needs improvement


ryanfeeley:
1. At this breakpoint content should be as wide as the right side of the *Done* button
1. Disconnect button should be wider and/or have smaller font (there may be another issue for this)
![IMG_9960](https://user-images.githubusercontent.com/68213/58347943-2d4a8f80-7e2d-11e9-92f5-897ae30a0a86.PNG)

PR #1293, opened 2019-05-29
fix(trailhead): Fix the vertical alignment on sync selection area

shane-tomlinson:
Add a 15px padding to the sync selection area, including both the h2
and the checkbox elements.

fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/663

### Desktop

Screenshot 2019-05-29 at 11 35 47

### Mobile
Screenshot 2019-05-29 at 11 35 53

@mozilla/fxa-devs - r?

PR #1335, opened 2019-06-02
fix(ui): adjust button size and client name in devices list

karansapolia:
Adjust button size to accomodate button title better
and increase content width in mobile view.

fixes: #1254

@ryanfeeley review please

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -467,7 +467,10 @@ section.modal-panel {
     min-width: 100px;
     position: absolute;
     top: 0;
-    width: 20%;
+
+    @include respond-to('big') {
+      width: 20%;
+    }
 
     html[dir='ltr'] & {
       right: 0;


shane-tomlinson:
Thanks for this @karansapolia - this looks great on mobile viewports but on desktop the buttons are far too wide:

Screenshot 2019-06-07 at 13 32 01

I suspect that removing the "width: 20%" has caused this. Perhaps this could be fixed by wrapping the removed `width: 20%` with a `respond-to('big')`:

```sass
@include respond-to('big') {
  width: 20%;
}
```

karansapolia:
> Thanks for this @karansapolia - this looks great on mobile viewports but on desktop the buttons are far too wide:

Ah, right! I didn't check it for desktop. My bad. I'll be meticulous now onwards. 

> When doing UI changes like this, it can be really helpful to attach screen shots of the result in both desktop and mobile mode.

Point duly noted. Such a mistake won't happen again.
I'll update with the required changes in a while. Thanks for the review.

shane-tomlinson:
> Point duly noted. Such a mistake won't happen again.

Here are a couple of examples from PRs I've done recently:

* https://github.com/mozilla/fxa/pull/1294
* https://github.com/mozilla/fxa/pull/1293

Mostly I suggest this because I find that it helps my PR get reviewed and merged a lot faster than if I don't.

karansapolia:
> Perhaps this could be fixed by wrapping the removed `width: 20%` with a `respond-to('big')`:
> 
> ```sass
> @include respond-to('big') {
>   width: 20%;
> }
> ```

Done as requested. 

**Desktop**
Screenshot from 2019-07-07 20-19-11

**Mobile**
![Screenshot from 2019-07-07 20-19-43](https://user-images.githubusercontent.com/10175625/60770092-ebcc2600-a0f4-11e9-9725-f6919494eac1.png

karansapolia:
Hey @shane-tomlinson @ryanfeeley, any word on this?

packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -396,14 +396,6 @@ section.modal-panel {
   .client-name {
     line-height: 1.2;
 
-    html[dir='ltr'] & {
```
shane-tomlinson: This is needed because really really long device names end up being covered by the disconnect button.

Screenshot 2019-07-22 at 15 35 08

karansapolia: Changes made as requested.

karansapolia:
Any idea why the build is failing @shane-tomlinson?

PR #1337, opened 2019-06-03
feat(test): Run *all* tests (including OAuth) on Circle

### LINKED ISSUE #672: Add OAuth tests to circle-ci run
Now that the tests are run in parallel and fxa-ci supports OAuth, we should be able to add the OAuth tests to circle.

shane-tomlinson:
Combine the test list from functional_circle and functional_oauth
into functional and run them all on Circle.

fixes #672


shane-tomlinson:
I'm fully expecting these tests to fail since we have never run OAuth tests on Circle.

shane-tomlinson:
@mozilla/fxa-devs - r?

shane-tomlinson:
> a 33minute run is pretty slow, but 🤷‍♂

It is, we could try increasing parallelization again, and then we might be able to save time by [doing a better job at caching](https://github.com/mozilla/fxa/issues/1338).

shane-tomlinson:
Plus, we should really remove tests where there is a high chance of duplication. For example, all the non-oauth, non-sync, non-settings tests, they could probably be jettisoned.

clouserw:
Reviewing tests for duplication and any other improvements could/should be on that modernizing fxa document which we'll talk about at the all hands

PR #1349, opened 2019-06-03
fix(emails): translate the "Confirm email" view action

### LINKED ISSUE #1339: Verification reminder email action text is not translated
We added email actions to the verification reminders, but the text for them lacks the surrounding `{{t ...}}` delimiters used for l10n.

philbooth:
Fixes #1339.

When I added email actions to the verification reminders, I forgot to translate the text.

@mozilla/fxa-devs r?


packages/fxa-auth-server/lib/senders/partials/verification_reminder_first.html
```diff
@@ -55,7 +55,7 @@ <h1 style="font-family: sans-serif; font-size: 21px; line-height: 29px; font-wei
   "description": "Reminder: Finish Creating Your Account",
```
shane-tomlinson: Does the description need to be translated too?

shane-tomlinson:
LGTM, do you want this against train-138 instead of master? Also, do the descriptions need to be translated?

packages/fxa-auth-server/lib/senders/partials/verification_reminder_first.html
philbooth: No.

philbooth:
> do you want this against train-138 instead of master?

No.

> do the descriptions need to be translated?

No.

PR #1380, opened 2019-06-04
feat(metrics): emit view, engage and submit events for CWTS

philbooth:
Related to #938.

This hooks up view, engage, submit and back events for CWTS. It's not a complete fix for the linked issue because I'm going to add a user property for the selected Sync engines in a follow-up PR.

Sample flows showing the new events.

* Flow events:
  
  ```
  {"event":"flow.signup.view","flow_id":"50fd6c525cb33e8f47e46ae513cf0457af8be04e7f89a91b8301d363d444e615","flow_time":2142,"hostname":"philbooth-5ejhd3.local","locale":"en","op":"flowEvent","pid":78120,"time":"2019-06-04T14:17:19.127Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:65.0) Gecko/20100101 Firefox/65.0","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
  {"event":"flow.signup.engage","flow_id":"50fd6c525cb33e8f47e46ae513cf0457af8be04e7f89a91b8301d363d444e615","flow_time":2994,"hostname":"philbooth-5ejhd3.local","locale":"en","op":"flowEvent","pid":78120,"time":"2019-06-04T14:17:19.979Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:65.0) Gecko/20100101 Firefox/65.0","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
  {"event":"flow.signup.submit","flow_id":"50fd6c525cb33e8f47e46ae513cf0457af8be04e7f89a91b8301d363d444e615","flow_time":10102,"hostname":"philbooth-5ejhd3.local","locale":"en","op":"flowEvent","pid":78120,"time":"2019-06-04T14:17:27.087Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:65.0) Gecko/20100101 Firefox/65.0","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
[...]

Issue #938, opened 2019-04-25
Track CWTS screen view in Amplitude


davismtl:
When a user sees the choose what to sync screen during a registration, we should log a view and the choices submitted.

**Add**
`fxa_reg - cwts_view`
`fxa_reg - cwts_submit`

I added them to the taxonomy doc with proposed properties:
https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4/edit#gid=879531055

I think the main _user property_ that would be different is `cwts_reg` for `fxa_reg - cwts_submit`. I would like to log an array of the engines left enabled when they submitted.

I'm filing this in response to a Slack conversation around confirmation codes between @ryanfeeley and @shane-tomlinson 

CC @irrationalagent -> Make alternative proposal to this taxonomy and event names.

davismtl:
@irrationalagent do you think that this would help better understand the drop off in our funnel? Relevant to Trailhead?

irrationalagent:
I'll look to see if we log in this in redshift

irrationalagent:
Looks like we have [events for views but not for engage](https://sql.telemetry.mozilla.org/queries/62867/source#161319). If these are important to have in amplitude, I could probably look into it and put up a PR.

ryanfeeley:
@irrationalagent It's great to know how users interact with this screen because verifying the email submits the form as is (or perhaps submits it with all checked).

philbooth:
> If these are important to have in amplitude, I could probably look into it and put up a PR.

Fwiw I think there's three parts to this, if you do decide to tackle it:

- [ ] Import and use `FlowEventsMixin` in `app/scripts/views/choose_what_to_sync.js`. That should be sufficient to generate the engage and submit events on the client side.

- [ ] Optionally add `data-flow-event` attributes to `<a>` tags in `app/scripts/templates/choose_what_to_sync.mustache` if you want track clicks on them.

- [ ] Update the definition of `VIEW_ENGAGE_SUBMIT_EVENT_GROUPS` in `server/lib/amplitude.js`.

irrationalagent:
thanks as always for the guidance @philbooth if I can clear my plate in the next few day's I'll assign this issue to myself.

philbooth:
Put my face on this and opened some PRs because it was a small chunk of work and I'm trying not to leave anything big unfinished before going on PTO.

irrationalagent:
[...]

PR #1382, opened 2019-06-04
feat(metrics): add optional sync_engines user property to amplitude

philbooth:
Related to #938. Depends on #1380.

Adds a new user property to Amplitude, `sync_engines`. Ignored unless it's populated, because of the way it defaults to the empty array in the content server (as it's a user property, we don't want it to be clobbered by the default value). Opened as draft because it includes commits from #1380, which hasn't landed yet.


philbooth:
Ffs, forgot to make it draft and GitHub doesn't seem to let you change that. Closing.

PR #1343, opened 2019-06-03
chore(deps): update shrinkwrap in the fxa-content-server

vladikoff:
Testing a shrinkwrap update, I suspect there could be breaking changes in the test dependencies

PR #1383, opened 2019-06-04
feat(metrics): pass selected sync engines to amplitude

### LINKED ISSUE #938: Track CWTS screen view in Amplitude

philbooth:
Fixes #938. Depends on #1384.

Passes the selected Sync engines from the CWTS view through to `POST /metrics`, from which point they're propagated to Amplitude automatically. Opened as draft because it includes commits from #1380 and #1384, which haven't landed yet, and it needs shrinkwrap regenerating after I publish a new release of `fxa-shared`.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
+++ b/packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
@@ -92,7 +92,10 @@ const View = FormView.extend({
     });
 
     return this.user.setAccount(account)
-      .then(this.onSubmitComplete);
+      .then(account => {
+        this.notifier.trigger('set-sync-engines', offeredSyncEngines);
+        return this.onSubmitComplete(account);
+      });
   },
 
   /**
--- a/packages/fxa-content-server/npm-shrinkwrap.json
+++ b/packages/fxa-content-server/npm-shrinkwrap.json
@@ -5672,9 +5672,9 @@
       }
     },
     "fxa-shared": {
-      "version": "1.0.24",
-      "resolved": "https://registry.npmjs.org/fxa-shared/-/fxa-shared-1.0.24.tgz",
-      "integrity": "sha512-xrAB4sIRhwWcZoxWjwrRoDCDCEo7TzKxlGpkG1CKQDCpbO7uT5ddWEiDYfbOxfe17621fy/VWm8n+8TUOFJddg==",
+      "version": "1.0.26",
+      "resolved": "https://registry.npmjs.org/fxa-shared/-/fxa-shared-1.0.26.tgz",
+      "integrity": "sha512-JlLmvoQ8UZ5OasFro9JUTsfabATlDUS/411EZxznvQJtmBCZQ8URNcpJAadsUlu2cAA8i0DO/LcPztZSoCaxBQ==",
       "requires": {
         "accept-language": "2.0.17",
         "ajv": "6.10.0",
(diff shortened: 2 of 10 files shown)


packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
```diff
@@ -92,7 +92,10 @@ const View = FormView.extend({
     });
 
     return this.user.setAccount(account)
-      .then(this.onSubmitComplete);
+      .then(() => {
+        this.notifier.trigger('set-sync-engines', offeredSyncEngines.map(e => e.text));
```
shane-tomlinson: Do you need the .text here? `_getOfferedEngineIds` returns the element IDs rather than the text.

packages/fxa-content-server/server/lib/validation.js
```diff
@@ -24,6 +24,7 @@ const PATTERNS = {
   FORM_TYPE: /^(email|other)$/,
   MIGRATION: /^(sync11|amo|none)$/,
   SERVICE: /^(sync|content-server|none|[0-9a-f]{16})$/,
+  SYNC_ENGINE: /^[\w\s.:-]+$/,
```
shane-tomlinson: Can there be whitespace in the sync engine name?

packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
philbooth: Ah thanks, didn't realise. Hadn't actually got round to testing this one yet, was waiting for the other two to land. 😊

packages/fxa-content-server/server/lib/validation.js
philbooth: I was erring on the side of caution because of the `.text` thing above. But I can get rid now it's using ids.

philbooth:
@mozilla/fxa-devs r?

packages/fxa-content-server/npm-shrinkwrap.json
```diff
@@ -5672,9 +5672,9 @@
       }
     },
     "fxa-shared": {
-      "version": "1.0.24",
-      "resolved": "https://registry.npmjs.org/fxa-shared/-/fxa-shared-1.0.24.tgz",
-      "integrity": "sha512-xrAB4sIRhwWcZoxWjwrRoDCDCEo7TzKxlGpkG1CKQDCpbO7uT5ddWEiDYfbOxfe17621fy/VWm8n+8TUOFJddg==",
+      "version": "1.0.26",
+      "resolved": "https://registry.npmjs.org/fxa-shared/-/fxa-shared-1.0.26.tgz",
+      "integrity": "sha512-JlLmvoQ8UZ5OasFro9JUTsfabATlDUS/411EZxznvQJtmBCZQ8URNcpJAadsUlu2cAA8i0DO/LcPztZSoCaxBQ==",
```
philbooth: I cheated here and didn't do a full shrinkwrap because #1343.

PR #1388, opened 2019-06-04
feat(payments): fix #1280, consume configuration in payments frontend

### LINKED ISSUE #1280: [payments] Consume meta config data
This is a followup to #1160 and #1275:

That PR adds a `<meta>` tag to the page, that is in the format that can be consume by [config-loader.js](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/config-loader.js)

We should:

- [ ] Move config-loader.js into some shared location usable by both servers
- [ ] Maybe remove underscore from that library (since it's used for something trivial)
- [ ] Include it in the payments server
- [ ] Use the config data in the payments server
- [ ] Maybe rename `fxa-content-server/config` to a more neutral `fxa-config`

ianb:
This also changes the config naming on the frontend to match the server configuration.

I started this out trying to do the "right thing" and move `packages/fxa-content-server/app/scripts/lib/config-loader.js` to fxa-shared, but that became complex and awkward, and the actual code is only 25 lines of real functionality. So I switched back to the simple thing and just reimplemented the functionality.

Also a bunch of automatic whitespace trimming happened to the files I touched.

### DIFF
--- a/packages/fxa-payments-server/server/lib/server.js
+++ b/packages/fxa-payments-server/server/lib/server.js
@@ -26,10 +26,24 @@ module.exports = () => {
 
   // Each of these config values (e.g., 'servers.content') will be exposed as the given
   // variable to the client/browser (via fxa-content-server/config)
-  const CLIENT_CONFIG_MAP = {
-    contentUrl: 'servers.content',
-    oAuthUrl: 'servers.oauth',
-    profileUrl: 'servers.profile',
+  const CLIENT_CONFIG = {
+    servers: {
+      auth: {
+        url: config.get('servers.auth.url'),
+      },
+      content: {
+        url: config.get('servers.content.url'),
+      },
+      oauth: {
+        url: config.get('servers.oauth.url'),
+      },
+      profile: {
+        url: config.get('servers.profile.url'),
+      },
+    },
+    stripe: {
+      apiKey: config.get('stripe.apiKey'),
+    },
   };
 
   // This is a list of all the paths that should resolve to index.html:
@@ -88,17 +102,10 @@ module.exports = () => {
     return result;
   }
 
-  function getClientConfig() {
-    // See also packages/fxa-content-server/server/lib/routes/get-index.js
-    const clientConfig = {};
-    for (const exportVariable in CLIENT_CONFIG_MAP) {
-      clientConfig[exportVariable] = config.get(CLIENT_CONFIG_MAP[exportVariable]);
-    }
-    return clientConfig;
-  }
-
   const STATIC_DIRECTORY =
     path.join(__dirname, '..', '..', config.get('staticResources.directory'));
+
+  const STATIC_INDEX_HTML = fs.readFileSync(path.join(STATIC_DIRECTORY, 'index.html'), {encoding: 'UTF-8'});
 
   const proxyUrl = config.get('proxyStaticResourcesFrom');
   if (proxyUrl) {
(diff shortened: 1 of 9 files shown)


packages/fxa-payments-server/server/lib/server.js
```diff
@@ -26,10 +26,21 @@ module.exports = () => {
...
+      oauth: {
+        url: config.get('servers.oauth.url'),
+      },
+      profile: {
+        url: config.get('servers.profile.url'),
+      },
+    },
+    stripe: {
+      apiKey: config.get('stripe.apiKey'),
+    },
```
ianb: I could have even done: `servers: config.get('servers')` but that seemed a tad to implicit to me. But maybe it should be that, I don't know...

packages/fxa-payments-server/src/index.tsx
```diff
@@ -58,7 +58,7 @@ async function getVerifiedAccessToken({
 
   try {
     const result = await fetch(
-      `${config.OAUTH_API_ROOT}/verify`,
+      `${config.servers.oauth.url}/v1/verify`,
```
ianb: I ended up leaving the version numbers (`/v1`) out of the server config (it doesn't mean anything to ops), and then putting it right where the URLs are used. In a sense this fetch is coded to use the v1 protocol, so it seems somewhat correct to me that it's encoded here.

ianb:
@lmorchard I'm not actually sure how `config.ts` should act in a test environment, thoughts?

packages/fxa-payments-server/src/index.tsx
lmorchard: Yeah, I think this makes sense.

packages/fxa-payments-server/src/lib/config.ts
```diff
@@ -0,0 +1,53 @@
...
+  Object.assign(config, decodeConfig(configEl.getAttribute('content')));
+  const featureEl = document.head.querySelector('meta[name="fxa-feature-flags"]');
+  if (!featureEl) {
+    throw new Error('<meta name="fxa-feature-flags"> is missing');
+  }
+  config.featureFlags = decodeConfig(featureEl.getAttribute('content'));
+  config.lang = document.documentElement.lang;
+}
+
+init();
```
lmorchard: Might do with a little control inversion, here.

Instead of calling `init()` here, I'd export it and call it in [the main `index.tsx`](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/index.tsx#L24) just before the first `<App>` render.

Also a good idea to accept `document` as a parameter in `init()`. The call in `index.tsx` can pass the real one, but we could easily pass in a mock `document` in unit tests of `config.ts`.

Then, I'd add & export another function for use in tests & stories for setting mock config. Something like:
```javascript
export function setConfig(newConfig) {
  return Object.assign(config, newConfig);
}
```

packages/fxa-payments-server/server/lib/server.js
lmorchard: I think this is fine. We could play around with it more, if someday we accumulate a dozen more settings or something.

ianb:
@lmorchard : I'll squash the commits before landing, but I thought it would be easier to review the change: 5c77fd62426efd0f248e24ae22b629c9d30f4790

I made initialization from a meta tag explicit, which basically solves the issue without even addressing any testing. I don't love the empty default config that is exposed until you load something in, nor do I love the fact that the object must always be updated in-place (since the object itself gets imported), but I don't _hate_ either of those things.

lmorchard:
LGTM, r+!

Issue #941, opened 2019-04-26
Standardise a mechanism for cross-package dependencies


philbooth:
I thought we already had an issue for this, but couldn't find one. Close as a dup if I missed it.

Currently there are two mechanisms in play for cross-repo dependencies:

* The historical approach of adding them as dependencies at the npm level, in `package.json`.

* The [ad hoc approach](https://github.com/mozilla/fxa/pull/796) we rushed through for the auth db, where a dependency is copied into consuming packages by scripting.

The disadvantage of using npm is that it enforces a multi-step approach, where changes must be published before they can be used in-tree. It's also node-specific.

The disadvantage of the ad hoc approach is that it requires scripting to run for the dependency to be present. I think I caught most of the cases in that PR but still, it's kind of messy and wastes disk space (which would be more of a concern if we were to adopt it as a standard mechanism for all the deps).

In an ideal world, if we get #694 done we might find that the repo size including dependencies is small enough that we could use a single Docker image for all packages that includes the entire repo. Then we could just `../` up the tree willy-nilly. That seems like it would be the simplest approach, but is weighed down by the fact that #694 is not a priority and will likely hang around for a few months yet.

Maybe we could introduce a generalised version of the copy approach in our scripting as an interim measure, e.g. where we look for certain dependencies in `package.json` then do the copying for all of those in the same places that we did it for the auth db stuff. That would mean adding a `package.json` to the email service, which is written in Rust, but I don't have a problem with that.

Plus there's probably other possibilities I haven't considered here, @mozilla/fxa-devs feel free to weigh in with suggestions or whatever.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-827)
┆Issue Number: FXA-827

lmorchard:
Can [lerna](https://github.com/lerna/lerna) help us out more here? I haven't played with it a bunch yet. I see lots of mentions in the docs about linking inter-dependent packages from the project itself, but scratching my head at whether it helps keep external dependencies in sync between local packages.

philbooth:
@mozilla/fxa-devs, this keeps coming up, I think we should try to come to a decision one way or the other fairly soon.

bbangert:
[...]

PR #796, opened 2019-04-10
fix(tests): pick up latest auth db in auth server tests

philbooth:
Fixes #793.

This is what I had in mind to hook the auth server tests up with the auth db from this repo. When Circle is back I'll restart the build to see how it gets on.

PR #1392, opened 2019-06-05
feat(deps): Use the fxa-js-client from the repo

shane-tomlinson:


philbooth:
@shane-tomlinson, see the discussion in #941, I don't think this is going to work unless you explicitly copy it into the Docker image, is it?

We hacked together an approach like that for the auth db fwiw, see #796 in case you want to do something similar. But ideally (long-term) I think we'd work towards a single common Docker image for all the node packages, which permits direct `../` dependencies.

shane-tomlinson:
Shoot, I had the fxa-js-client initially installed as a `file:..` and then overwrote that.

shane-tomlinson:
> Shoot, I had the fxa-js-client initially installed as a file:.. and then overwrote that.

No, wait, this was intentional, that's why I added the `npm install ../fxa-js-client` in Dockerfile-build, to see if Docker would copy that in.

shane-tomlinson:
Closing this yet again.

PR #1546, opened 2019-06-07
chore(csp): Change CSP logging format to mozlog 2.0

jbuck:
This still needs some test changes before merging, but one step closer to getting off heka!

### DIFF
--- a/packages/fxa-content-server/server/lib/routes.js
+++ b/packages/fxa-content-server/server/lib/routes.js
@@ -63,6 +63,7 @@ module.exports = function(config, i18n) {
   if (config.get('csp.enabled')) {
     routes.push(
       require('./routes/post-csp')({
+        op: 'server.csp',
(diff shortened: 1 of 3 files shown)


jbuck:
I am not going to lie, I am incredibly surprised that this passed tests. Should, uh, we merge it?

shane-tomlinson:
> I am not going to lie, I am incredibly surprised that this passed tests. Should, uh, we merge it?

Can you verify there is a test for this?

jbuck:
https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/tests/server/routes/post-csp.js

shane-tomlinson:
Huh, they *are* failing locally:

Screenshot 2019-06-10 at 17 25 46

That seems to imply server tests are *not* run on Circle. Opening an issue.

PR #1549, opened 2019-06-09
feat(event-broker): add proxy for RPs and integration test utils

### LINKED ISSUE #1269: [fxa-event-broker] Relying Party proxy from Pub/Sub
- Add an API endpoint that GCP Pub/Sub will hit using HTTPS
- Verify Pub/Sub JWT signature
- Sign new message with FxA JWT and send to RP (webhook URL will be on the message)

bbangert:
This took a bit longer than expected, as configuring the various integration utilities to get a fully working setup was a bit of a challenge as well as the bug in the Firestore emulator preventing it from working correctly (It has been fixed in the latest Firestore, but not released yet).

The JWT signing is using the [SET](https://tools.ietf.org/html/rfc8417) format per @rfk's suggestion, but its still entirely likely I goofed on something there so that should be checked quite carefully.

This PR does not include the `.well-known` mappings (#1583) for the JWT public key, which will be in a later PR.

Closes #1269


packages/fxa-event-broker/config/index.ts
```diff
@@ -90,11 +90,58 @@ const conf = convict({
...
+  openid: {
+    issuer: {
+      default: '',
+      doc: 'OpenID Issuer',
+      env: 'OPENID_ISSUER',
+      format: String
+    },
+    key: {
+      default: {},
+      doc: 'Private JWK to sign id_tokens',
```
rfk: nit: you're not generating `id_token`s, you're generating `SET`s.

packages/fxa-event-broker/lib/grant.ts
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rfk: File naming nit: this doesn't have anything to do with "grants" the way the oauth-server means them, this file would probably be more clearly named as "jwts.ts" or "setokens.js" or similar; (naming it "set.js" would unfortunately be confusing for readers who don't know that's suppose to be an acronym...)

(The oauth-server could really use a glossary, but that's a separate issue...)

packages/fxa-event-broker/lib/grant.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Generate a Security Event Token for delivery to Relying Parties.
+   *
+   * See: https://tools.ietf.org/html/rfc8417
+   *
+   * @param grant
+   */
+  public generateSET(grant: securityEvent): Promise<string> {
```
rfk: as above, "grant" doesn't seem like the right name in this context; "claims" would be the generic name for "the things I want to put inside a JWT" but "event" would probably work fine in this more specific context.

packages/fxa-event-broker/lib/grant.ts
```diff
@@ -0,0 +1,79 @@
...
+    const now = Math.floor(Date.now() / 1000);
+    const claims = {
+      aud: grant.clientId,
+      events: grant.events,
+      iat: now,
```
rfk: I think you should get `iat` for free from the underlying JWT library; if you don't, that's a bug in jwtool and we should add it there.

packages/fxa-event-broker/lib/grant.ts
```diff
@@ -0,0 +1,79 @@
...
+      iss: this.issuer,
```
rfk: you should also get `iss` for free because you configured it when creating the key above.

packages/fxa-event-broker/lib/jwts.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * Generate a Subscription Security Event Token.
+   *
+   * @param subEvent
+   */
+  public generateSubscriptionSET(subEvent: subscriptionEvent): Promise<string> {
+    return this.generateSET({
+      clientId: subEvent.clientId,
+      events: {
+        [SUBSCRIPTION_STATE_EVENT_ID]: {
```
lmorchard: Haven't thoroughly read the SET spec yet, but skimming it and this seems like it makes sense as a carrier for this info

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -118,17 +126,19 @@ class ServiceNotificationProcessor {
...
-    const message = joi.attempt(JSON.parse(body.Message), BASE_MESSAGE_SCHEMA);
+    const message = joi.attempt(body, BASE_MESSAGE_SCHEMA);
     switch (message.event) {
       case LOGIN_EVENT: {
         const loginMessage = joi.attempt(message, LOGIN_SCHEMA);
         await this.db.storeLogin(loginMessage.uid, loginMessage.clientId);
+        this.logger.debug('sqs.loginEvent', loginMessage);
         return;
       }
       case SUBSCRIPTION_UPDATE_EVENT: {
```
lmorchard: I'm super fuzzy on SQS goings-on in FxA right now - is this `SUBSCRIPTION_UPDATE_EVENT` something yet to be implemented on the `fxa-auth-server` side? I seem to remember we do a [`push.notifyProfileUpdated(uid, devices);` and `log.notifyAttachedServices('profileDataChanged', request, { uid, email });`](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/subscriptions.js#L160) when subscriptions are created and deleted. But, I'm not sure if that does the trick here.

packages/fxa-event-broker/test/lib/api/proxy-controller.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+import { assert, SinonSpy } from 'sinon';
+import * as sinon from 'sinon';
+import { stubInterface } from 'ts-sinon';
+
+import Config from '../../../config';
+import { SUBSCRIPTION_STATE_EVENT_ID } from '../../../lib/jwts';
+import * as proxyserver from '../../../lib/proxy-server';
+import { ClientWebhookService } from '../../../lib/selfUpdatingService/clientWebhookService';
+
+const TEST_KEY = {
```
lmorchard: I'm also fuzzy on JWT keys & signing, but the tests pass and they look clever so :+1:

packages/fxa-event-broker/lib/api/proxy-controller.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (Array.isArray(value)) {
+        for (const val of value) {
+          resp = resp.header(name, val);
+        }
+      } else {
+        resp = resp.header(name, value);
+      }
+    });
+
+    return resp;
```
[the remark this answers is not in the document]
bbangert: We don't, but Google will log it.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: There's a separate event emitted for subscription state changes that @philbooth landed awhile back.

PR #1550, opened 2019-06-10
feat(oauth): Add a config option to disable particular OAuth clients.

rfk:
If we discover something busted with the Account or Sync integration in Fenix, we may find ourselves wanting to forcibly disable it - particularly if some unexpected client misbehaviour means it is destabilizing our service for other clients.

Here's a blunt instrument which would allow us to do so.  By adding the Fenix OAuth clent_id to a server-side config option, we could prevent new users from signing in to Fenix and prevent already-signed-in users from renewing their access tokens. That's sad for those users, but it may be better than a service outage for everyone.

By returning a 503, clients *should* treat this as a transient server-side error, so clients that already have a refresh_token will stay signed in and will be able to sync again once we re-enable the client.

@shane-tomlinson @vladikoff @jbuck what do you think, worth having something like this in our pocket just in case?

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/error.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/error.js
@@ -295,4 +295,29 @@ AppError.invalidGrantType = function invalidGrantType() {
   });
 };
 
+AppError.disabledClient = function disabledClient(clientId) {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: 122,
+    message: 'This client has been temporarily disabled'
+  }, { clientId });
+};
+
+AppError.serviceUnavailable = function serviceUnavailable(retryAfter) {
+  if (! retryAfter) {
+    retryAfter = 30;
+  }
+  return new AppError({
+    code: 503,
+    error: 'Service Unavailable',
+    errno: 201,
+    message: 'Service unavailable'
+  }, {
+    retryAfter: retryAfter
+  }, {
+    'retry-after': retryAfter
+  });
+};
+
 module.exports = AppError;
--- a/packages/fxa-auth-server/lib/oauthdb/utils.js
+++ b/packages/fxa-auth-server/lib/oauthdb/utils.js
@@ -58,6 +58,10 @@ module.exports = {
       return error.staleAuthAt(err.authAt);
     case 120:
       return error.insufficientACRValues(err.foundValue);
+    case 122:
+      return error.disabledClientId(err.clientId);
+    case 201:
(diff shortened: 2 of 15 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/error.js
```diff
@@ -295,4 +295,20 @@ AppError.invalidGrantType = function invalidGrantType() {
...
 };
 
+AppError.serviceUnavailable = function serviceUnavailable(retryAfter) {
+  if (! retryAfter) {
+    retryAfter = 30;
+  }
+  return new AppError({
+    code: 503,
+    error: 'Service Unavailable',
+    errno: 201,
```
shane-tomlinson: We'll have to add this to the [list of front-end OAuth server errors](https://github.com/mozilla/fxa/blob/dc06472fd5278f179065a75779fdfe4baa4efa39/packages/fxa-content-server/app/scripts/lib/oauth-errors.js#L100).

packages/fxa-auth-server/lib/oauthdb/utils.js
```diff
@@ -58,6 +58,8 @@ module.exports = {
       return error.staleAuthAt(err.authAt);
     case 120:
       return error.insufficientACRValues(err.foundValue);
+    case 201:
+      return error.serviceUnavailable(err.retryAfter);
```
shane-tomlinson: Strange that `serviceUnavailable` returns [SERVER_BUSY](https://github.com/mozilla/fxa/blob/dc06472fd5278f179065a75779fdfe4baa4efa39/packages/fxa-auth-server/lib/error.js#L497), that's what we show [on the frontend too](https://github.com/mozilla/fxa/blob/dc06472fd5278f179065a75779fdfe4baa4efa39/packages/fxa-content-server/app/scripts/lib/auth-errors.js#L255). I wonder how we can do better.

rfk:
Good point; I think ideally we'd return a special flag in the OAuth client metadata, and have the front-end show a specific error message. I'll see if there's a small change that would allow that, or move it to a followup.

rfk:
@shane-tomlinson I updated this to return a specific errno, and added an error message on the content-server.

I also updated it to return a "disabled" flag in the oauth client info, in the hope that the frontend could detect this and show a helpful error message to the user straight away, rather than waiting for them to try to login and then failing with the backend server error. Unfortunately I couldn't figure out how to plumb that in nicely on the front-end; thoughts?

rfk:
I'm going to re-open this as a point-release for train-139 branch.

PR #1551, opened 2019-06-10
feat(clients): Add oauth-server endpoint for listing all authorized clients

rfk:
This new endpoint allows listing each individual `refresh_token` as a separate instance of an attached client, which will be useful for displaying them in the "devices and apps" view.  It is an extraction of part of https://github.com/mozilla/fxa/pull/1262 into a smaller standalone PR, since this part is ready for review.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/docs/api.md
+++ b/packages/fxa-auth-server/fxa-oauth-server/docs/api.md
@@ -74,8 +74,10 @@ The currently-defined error responses are:
   - [POST /v1/developer/activate][developer-activate]
 - [POST /v1/verify][verify]
 - [POST /v1/key-data][key-data]
-- [GET /v1/client-tokens][client-tokens]
-- [DELETE /v1/client-tokens/:id][client-tokens-delete]
+- [POST /v1/authorized-clients][authorized-clients]
+- [POST /v1/authorized-clients/destroy][authorized-clients-destroy]
+- (**DEPRECATED**) [GET /v1/client-tokens][client-tokens]
+- (**DEPRECATED**) [DELETE /v1/client-tokens/:id][client-tokens-delete]
 
 ### GET /v1/client/:id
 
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/fxa-oauth-server/docs/api.md
```diff
@@ -74,8 +74,10 @@ The currently-defined error responses are:
   - [POST /v1/developer/activate][developer-activate]
 - [POST /v1/verify][verify]
 - [POST /v1/key-data][key-data]
-- [GET /v1/client-tokens][client-tokens]
-- [DELETE /v1/client-tokens/:id][client-tokens-delete]
+- [POST /v1/authorized-clients][authorized-clients]
+- [POST /v1/authorized-clients/destroy][authorized-clients-destroy]
+- (**DEPRECATED**) [GET /v1/client-tokens][client-tokens]
+- (**DEPRECATED**) [DELETE /v1/client-tokens/:id][client-tokens-delete]
```
bbangert: Out of curiosity, what is the versioning in the API being used for? Are there plans for a /v2/, or are we mainly claiming by versioning that we won't change individual route parameters/responses?

packages/fxa-auth-server/fxa-oauth-server/lib/db/memory.js
```diff
@@ -309,6 +309,63 @@ MemoryStore.prototype = {
...
+   * Get all access tokens for a given user.
+   * @param {String} uid User ID as hex
+   * @returns {Promise}
+   */
+  getAccessTokensByUid: async function getAccessTokensByUid(uid) {
+    if (! uid) {
+      return P.reject(new Error('Uid is required'));
+    }
+    const accessTokens = [];
+    Object.keys(this.tokens).forEach(id => {
```
bbangert: Why not `Object.entries` since you want the token as well?

packages/fxa-auth-server/fxa-oauth-server/lib/db/memory.js
```diff
@@ -309,6 +309,63 @@ MemoryStore.prototype = {
...
+  },
+
+  /**
+   * Get all refresh tokens for a given user.
+   * @param {String} uid User ID as hex
+   * @returns {Promise}
+   */
+  getRefreshTokensByUid: async function getRefreshTokensByUid(uid) {
+    if (! uid) {
+      return P.reject(new Error('Uid is required'));
```
bbangert: nit: This feels like a bit of a syntactic mismatch to return a `Promise.reject` instead of throw'ing in an async function.

packages/fxa-auth-server/fxa-oauth-server/docs/api.md
rfk: I believe it's "just in case" versioning if we decide we want to make a wildly different API, but we don't have concrete plans for a `v2`. (And it's amazing how much time can otherwise be spent on bikeshedding versioning schemes for hypothetical future API changes).

packages/fxa-auth-server/fxa-oauth-server/lib/db/memory.js
rfk: Huh, TIL about `Object.entries` and `Object.values. Thanks!

rfk:
re-pushed with nits addressed.

PR #1552, opened 2019-06-10
Add a route for listing all attached clients.

### LINKED ISSUE #466: Add a route specifically for generating the "devices and apps" list on the settings page
The "devices and apps" list on the settings page currently pulls in data from two different sources:

* The auth-server's `/sessions` endpoint, which includes all device records that have a `sessionToken`.
* The oauth-server's `/client-tokens` endpoint, which lists each connected OAuth client, but which doesn't distinguish between multiple instances of the same client (ref https://github.com/mozilla/fxa-auth-server/issues/2928)

And it does a bit of a complex sort-and-merge to produce the final list of all the things connected to the user's account.

As we work to [allow OAuth clients to access the devices API](https://github.com/mozilla/fxa-auth-server/issues/2547), managing that view is only going to get more complex.  I suggest we simplify it by creating an auth-server endpoint specifically for this purpose.

Strawman: add a new route `GET /v1/account/connected_clients`, authenticated with a `sessionToken`, that lists all of the things connected to the user's account.  Each entry returned by this endpoint would correspond precisely to an entry in the "devices and apps" list:
    * One entry for each device record (Firefox, Fenix, Reference Browser)
    * One entry for each OAuth refresh token that isn't owned by a device record (Monitor, Send, Lockbox)
    * One entry for each OAuth client_id that has live access tokens that don't tie back to a refresh token (Pocket, I'm looking at you...)
    * One entry for each sessionToken that isn't owned by a device record (web sessions)

We *could* synthesize an equivalent list on the client side by requesting the individual lists and merging them, but it seems like it would be both more responsive and more reliable to do this work on the server.  

The new route would aim to **replace** the current `/sessions` and `/client-tokens` routes, which IIUC are only called from web content for the purposes of rendering the "devices and apps" list.  Hopefully this would address earlier concerns about proliferating new routes and new maintenance burdens as we do the oauth-devices work.

@shane-tomlinson @vladikoff thoughts?

rfk:
This is an extraction of part of https://github.com/mozilla/fxa/pull/1262, and depends on https://github.com/mozilla/fxa/pull/1551. Review https://github.com/mozilla/fxa/pull/1551 first because this PR includes that commit.

This adds a new route `/account/attached_clients` that can be used to get the canonical list of all the things connected to the user's account. It combines web sessions, oauth apps, and device records into a single coherent list which should correspond directly to what we want to show the user in their "devices and apps" view.

Fixes #466.

### DIFF
--- a/packages/fxa-auth-server/lib/db.js
+++ b/packages/fxa-auth-server/lib/db.js
@@ -761,7 +761,7 @@ module.exports = (
                       return true;
                     }
 
-                    if (sessionTokenId && device.sessionToken === sessionTokenId ||
+                    if (sessionTokenId && device.sessionTokenId === sessionTokenId ||
                       deviceInfo.refreshTokenId && device.refreshTokenId === deviceInfo.refreshTokenId) {
                       conflictingDeviceId = device.id;
                     }
@@ -815,7 +815,7 @@ module.exports = (
             .then(devices => {
               let conflictingDeviceId;
               devices.some(device => {
-                if (device.sessionToken === sessionTokenId) {
+                if (device.sessionTokenId === sessionTokenId) {
                   conflictingDeviceId = device.id;
                   return true;
                 }
@@ -1469,12 +1469,13 @@ module.exports = (
     const mergedInfo = Object.assign({}, device, token);
     return {
       id: mergedInfo.id,
-      sessionToken: mergedInfo.sessionTokenId,
+      sessionTokenId: mergedInfo.sessionTokenId,
       refreshTokenId: mergedInfo.refreshTokenId,
       lastAccessTime: lastAccessTimeEnabled ? mergedInfo.lastAccessTime : null,
       location: mergedInfo.location,
       name: mergedInfo.name,
       type: mergedInfo.type,
+      createdAt: mergedInfo.createdAt,
       pushCallback: mergedInfo.callbackURL,
       pushPublicKey: mergedInfo.callbackPublicKey,
       pushAuthKey: mergedInfo.callbackAuthKey,
--- a/packages/fxa-auth-server/lib/routes/attached-clients.js
+++ b/packages/fxa-auth-server/lib/routes/attached-clients.js
@@ -1,0 +1,253 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const isA = require('joi');
+const validators = require('./validators');
+const error = require('../error');
+
+const HEX_STRING = validators.HEX_STRING;
+const DEVICES_SCHEMA = require('../devices').schema;
+
+module.exports = (log, db, oauthdb, devices, clientUtils) => {
+
+  return [
+    {
+      method: 'GET',
+      path: '/account/attached_clients',
+      options: {
+        auth: {
+          strategy : 'sessionToken'
+        },
+        response: {
+          schema: isA.array().items(isA.object({
+            clientId: isA.string().regex(HEX_STRING).allow(null).required(),
+            deviceId: DEVICES_SCHEMA.id.allow(null).required(),
+            sessionTokenId: isA.string().regex(HEX_STRING).allow(null).required(),
+            refreshTokenId: isA.string().regex(HEX_STRING).allow(null).required(),
+            isCurrentSession: isA.boolean().required(),
+            deviceType: DEVICES_SCHEMA.type.allow(null).required(),
+            name: DEVICES_SCHEMA.nameResponse.allow('').allow(null).required(),
+            createdTime: isA.number().min(0).required().allow(null),
+            createdTimeFormatted: isA.string().optional().allow(''),
+            lastAccessTime: isA.number().min(0).required().allow(null),
+            lastAccessTimeFormatted: isA.string().optional().allow(''),
+            approximateLastAccessTime: isA.number().min(0).optional(),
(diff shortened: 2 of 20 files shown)


packages/fxa-auth-server/lib/db.js
```diff
@@ -761,7 +761,7 @@ module.exports = (
                       return true;
                     }
 
-                    if (sessionTokenId && device.sessionToken === sessionTokenId ||
+                    if (sessionTokenId && device.sessionTokenId === sessionTokenId ||
```
bbangert: Was this to address a prior bug of some sort?

rfk: Oh, I meant to leave a comment about this - the field returned from the db is `sessionTokenId` but this was quietly changing it to `sessionToken` which lead to some frustrating test failure debugging. I changed to be `sessionTokenId` everywhere for consistency. If only we had some sort of technology that could prevent this kind of confusion in the first place... ;-)

bbangert:
@rfk Looks like this needs a rebase now.

packages/fxa-auth-server/lib/routes/attached-clients.js
```diff
@@ -0,0 +1,254 @@
...
+          isCurrentSession: false,
+          deviceType: null,
+          name: null,
+          createdTime: Infinity,
+          lastAccessTime: 0,
+          scope: null,
+          location: null,
+          userAgent: '',
+          os: null,
+          //scopes: null,
```
vladikoff: is this left for later?

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -126,7 +58,7 @@ module.exports = (log, db, config, customs, push, pushbox, devices, oauthdb) =>
       // But these need to be non-falsey, using default fallbacks if necessary
       id: (device && device.id) || credentials.deviceId,
       name: (device && device.name) || credentials.deviceName || devices.synthesizeName(credentials),
-      type: (device && device.type) || credentials.deviceType || 'desktop',
+      type: (device && device.type) || credentials.deviceType || (credentials.client || device.refreshTokenId ? 'mobile' : 'desktop'),
```
vladikoff: this should fix https://github.com/mozilla/application-services/issues/1163

packages/fxa-auth-server/test/client/api.js
```diff
@@ -650,6 +650,20 @@ module.exports = config => {
     );
   };
 
+  ClientApi.prototype.attachedClientDestroy = function (sessionTokenHex, clientData) {
+    return tokens.SessionToken.fromHex(sessionTokenHex)
+      .then(
+        (token) => {
+          return this.doRequest(
+            'POST',
+            `${this.baseURL  }/account/attached_client/destroy`,
```
vladikoff: `${this.baseURL  }` what happened here?

packages/fxa-auth-server/test/client/api.js
```diff
@@ -720,6 +734,19 @@ module.exports = config => {
       );
   };
 
+  ClientApi.prototype.attachedClients = function (sessionTokenHex) {
+    return tokens.SessionToken.fromHex(sessionTokenHex)
+      .then(
+        (token) => {
+          return this.doRequest(
+            'GET',
+            `${this.baseURL  }/account/attached_clients`,
```
vladikoff: `${this.baseURL  }`

packages/fxa-auth-server/test/client/api.js
rfk: Huh, `api.js` seems to be full of these, for reasons I don't understand. I assume I copy-pasted it from elsewhere.  Will fix mine and then address the others in a followup. Nice catch!

packages/fxa-auth-server/lib/routes/attached-clients.js
rfk: whoops, accidental dupe of a not-commented-out field.

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -475,29 +407,40 @@ module.exports = (log, db, config, customs, push, pushbox, devices, oauthdb) =>
...
             return deviceArray.map(device => {
-              return Object.assign({
+              const formattedDevice = {
                 id: device.id,
-                isCurrentDevice: !! ((credentials.id && credentials.id === device.sessionToken) ||
+                isCurrentDevice: !! ((credentials.id && credentials.id === device.sessionTokenId) ||
                   (credentials.refreshTokenId && credentials.refreshTokenId === device.refreshTokenId)),
-                location: marshallLocation(device.location, request),
+                lastAccessTime: device.lastAccessTime,
+                location: device.location,
```
philbooth: @rfk, I believe this change (the removal of `marshallLocation`) is the cause of:

https://sentry.prod.mozaws.net/operations/auth-prod/issues/6027903/?environment=prod

The response schema doesn't include `countryCode` in `location` and `marshallLocation` was responsible for stripping it out. What was the reason for getting rid of it here?

philbooth: Ah right, just seen noticed it moved down into `lib/routes/utils/clients.js`, but now it falls back to leaving the unadulterated `location` property behind instead of an empty object. I can fix it now, you can ignore the above sorry.

philbooth: Ref: https://github.com/mozilla/fxa/pull/1699

rfk: Ugh, sorry :-(

> What was the reason for getting rid of it here?

There's no particular unifying vision behind changing this specifically, it just got swept up in a broader refactor to share code between these routes.

PR #1571, opened 2019-06-12
[WIP] refactor(display-name): rendered the component in react instead of backbone

hritvi:
Worked on rendering the display-name component in react.
Still, need to work on these things:
- [ ] Writing tests
- [ ] Localize the component[The component in only being rendered in English, translator still needs to be added]

Some screenshots:
Screenshot from 2019-06-12 10-16-02

Screenshot from 2019-06-12 10-16-18

Screenshot from 2019-06-12 10-16-53

Screenshot from 2019-06-12 10-17-15


packages/fxa-content-server/webpack.config.js
```diff
@@ -132,7 +132,13 @@ const webpackConfig = {
             loader: 'babel-loader',
             options: {
               cacheDirectory: true,
-              presets: ['@babel/preset-env'],
+              presets: [
+                [
+                  '@babel/preset-react', {
```
lmorchard: Looks like this was easier to get included in the build process than I feared. Hooray!

packages/fxa-content-server/app/scripts/views/settings/display_name.jsx
```diff
@@ -0,0 +1,203 @@
...
+import SettingsPanelMixin from '../mixins/settings-panel-mixin';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Translator from '../../lib/translator';
+
+const translator = new Translator();
+
+const t = msg => translator.get(msg);
+
+function DisplayName(props) {
```
lmorchard: Consider adding `export` to all these React components. While we do `export` the `View` at the end of the module, being able to access each individual component separately might make it easier to test things in smaller pieces.

packages/fxa-content-server/app/scripts/views/settings/display_name.jsx
```diff
@@ -0,0 +1,203 @@
...
+  );
+}
+function DisplayNameComponent() {
+  return (
+    <header className="settings-unit-summary">
+      <h2 className="settings-unit-title">{t('Display name')}</h2>
+    </header>
+  );
+}
+function ChangeorAddButtonComponent(props) {
```
lmorchard: For example, if this component were `export`ed, you could write a small test for `ChangeorAddButtonComponent` in isolation that exercises what it does with and without `displayName` defined.

You might also be able to separate React tests from the rest of the existing tests. I have a suspicion that might make things simpler.

hritvi:
Have opened a new PR for it: #1611. Closing this.

Issue #1561, opened 2019-06-11
[fxa-payments-server] fail to install


vladikoff:
Does anyone else get this when running `npm i` on fxa?

```
+ cd fxa-payments-server
+ npm ci
npm WARN prepare removing existing node_modules/ before installation
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! 
npm ERR! Missing: @types/classnames@^2.2.8
```

rfk:
Yep, I hit this, doing `npm install` to regenerate fixed it for me in ac7d124bb74b6a1f4e6d3678805150bb45c15da6 but no idea if that's the right way to fix it...

PR #1577, opened 2019-06-13
Feature Flags for First Fenix Flight

rfk:
Opening this as a point-release to train-139, in the hope of having it in production before Fenix launch.

We have a new relier coming online with unknown but potentially significant volume, and it's going to be exercising our API in new and interesting ways. This adds some feature-flags that we can use to engage a graceful degradation of service if client behaviour causes unexpected problems:

* `OAUTH_DISABLE_NEW_CONNECTIONS_FOR_CLIENTS` on fxa-auth-server -- lets us refuse new signins or signups from a particular OAuth client, while leaving existing users connected.
  * This will be useful if we feel like current server load is sustainable but we don't want to increase it any further.
* `OAUTH_CLIENTS_DISABLED` on fxa-oauth-server -- lets us refuse to generate any tokens for particular oauth clients.
  * This will prevent new client instances from connecting, and additionally prevent connected instances from issuing new tokens. It's a big hammer!
  * Due to a client-side limitation, enabling this for Fenix will [sign users out](https://github.com/mozilla-mobile/android-components/issues/3347) even though we don't technically need to.
* ` OAUTH_DEVICE_COMMANDS_ENABLED` on fxa-auth-server -- lets us prevent OAuth clients from using the device-commands infrastructure (aka send-tab).
  * This lets the clients continue to do basic device registration and show up in the "devices and apps" list, but prevents them from using send-tab.
* `OAUTH_DEVICE_ACCESS_ENABLED` on fxa-auth-server -- lets us deny OAuth clients access to the entire devices API.
  * They will still be able to sync successfully (unless the broader disablement options above are also enabled)

Much of the diff here is tests, because I did a bit of cleanup as I went. It works better if you view without whitespace changes.

Using a local stack, I have manually tested that each of these have the desired effect on Fenix client behaviour when enabled via the instructions in the [Fenix Rollout Gameplan](https://docs.google.com/document/d/1doYLD7p6DnxAPLb-KYWSKUsi85M4JWbhJgnG6HOIstU/edit?ts=5d014545#heading=h.iyydstq02om).

### DIFF
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -662,6 +662,28 @@ const conf = convict({
       default: {},
       env: 'OAUTH_CLIENT_IDS'
     },
+    // A safety switch for disabling new signins/signups from particular clients,
+    // as a hedge against unexpected client behaviour.
+    disableNewConnectionsForClients: {
+      doc: 'Comma-separated list of oauth client ids for which new connections should be temporarily refused',
+      env: 'OAUTH_DISABLE_NEW_CONNECTIONS_FOR_CLIENTS',
+      format: Array,
+      default: []
+    },
+    // Some safety switches to disable oauth-based device API operations,
+    // in case problems with the client logic cause server overload.
+    deviceAccessEnabled: {
+      doc: 'Is oauth-based access to the devices API allowed at all?',
+      format: Boolean,
+      env: 'OAUTH_DEVICE_ACCESS_ENABLED',
+      default: true
+    },
+    deviceCommandsEnabled: {
+      doc: 'Are oauth-based devices allowed to use  device commands?',
+      format: Boolean,
+      env: 'OAUTH_DEVICE_COMMANDS_ENABLED',
+      default: true
+    },
     clientInfoCacheTTL: {
       doc: 'TTL for OAuth client details (in milliseconds)',
       format: 'duration',
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/error.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/error.js
@@ -295,4 +295,16 @@ AppError.invalidGrantType = function invalidGrantType() {
   });
 };
 
+// N.B. `errno: 201` is traditionally our generic "service unavailable" error,
+// so let's reserve it for that purpose here as well.
+
+AppError.disabledClient = function disabledClient(clientId) {
+  return new AppError({
+    code: 503,
+    error: 'Client Disabled',
+    errno: 202,
+    message: 'This client has been temporarily disabled'
+  }, { clientId });
+};
+
 module.exports = AppError;
--- a/packages/fxa-auth-server/fxa-oauth-server/test/api.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/test/api.js
@@ -3375,7 +3375,7 @@ describe('/v1', function() {
 
       it('should list canGrant=1 clients that have refresh tokens', function () {
         return db.registerClient({
-          name: 'test/api/client-tokens/z-list-can-grant',
+          name: 'test/api/client-tokens/aaa-list-can-grant',
           id: client2Id,
           hashedSecret: encrypt.hash(unique.secret()),
           redirectUri: 'https://example.domain',
@@ -3410,10 +3410,10 @@ describe('/v1', function() {
           .then(function (res) {
             var result = res.result;
             assert.equal(result.length, 2);
-            assert.equal(result[0].id, client1Id.toString('hex'));
-            assert.deepEqual(result[0].scope, ['clients:write', 'profile']);
-            assert.equal(result[1].id, client2Id.toString('hex'));
-            assert.deepEqual(result[1].scope, ['scopeFromRefreshToken']);
+            assert.equal(result[0].id, client2Id.toString('hex'));
+            assert.deepEqual(result[0].scope, ['scopeFromRefreshToken']);
+            assert.equal(result[1].id, client1Id.toString('hex'));
+            assert.deepEqual(result[1].scope, ['clients:write', 'profile']);
             assertSecurityHeaders(res);
           });
       });
--- a/packages/fxa-auth-server/fxa-oauth-server/test/lib/mocks.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/test/lib/mocks.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const path = require('path');
+const proxyquire = require('proxyquire');
 
 module.exports = {
   require: requireDependencies,
@@ -41,29 +42,12 @@ function requireDependency(dependency, modulePath, basePath) {
     return dependency.ctor;
   }
 
-  var localPath = dependency.path;
-
-  if (localPath[0] === '.') {
-    // attempt to find something like ./patch.js from mysql/index.js
-    localPath = path.relative(
-      basePath,
-      path.resolve(basePath, modulePath, localPath)
-    );
+  if (dependency.path[0] !== '.') {
+    return proxyquire(dependency.path, {});
   }
-
-  try {
-    return require(localPath);
-  } catch (e) {
-    // if above require fails, attempt to find the module as a sibling of `modulePath`
-    // this takes care of cases like: ./token.js being a sibling of the given module
-    localPath = path.relative(
-      basePath,
-      path.resolve(basePath, '..', modulePath, '..', dependency.path)
-    );
-
-    return require(localPath);
-  }
-
+  const moduleUnderTest = require.resolve(modulePath, { paths: [basePath] });
+  const dependencyPath = require.resolve(dependency.path, { paths: [path.dirname(moduleUnderTest)] });
+  return proxyquire(dependencyPath, {});
 }
 
 function mockLog(logger, cb) {
--- a/packages/fxa-auth-server/fxa-oauth-server/test/routes/authorization.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/test/routes/authorization.js
@@ -6,8 +6,15 @@
 
 const { assert } = require('chai');
 const Joi = require('joi');
-const route = require('../../lib/routes/authorization');
-const validation = route.validate.payload;
+const sinon = require('sinon');
+const proxyquire = require('proxyquire');
+const mocks = require('../lib/mocks');
+const realConfig = require('../../lib/config');
+
+const routeModulePath = '../../lib/routes/authorization';
+var dependencies = mocks.require([
+  { path: '../config' },
+], routeModulePath, __dirname);
 
 const CLIENT_ID = '98e6508e88680e1b';
 // jscs:disable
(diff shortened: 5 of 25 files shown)


packages/fxa-auth-server/fxa-oauth-server/test/lib/mocks.js
```diff
@@ -41,29 +42,12 @@ function requireDependency(dependency, modulePath, basePath) {
...
-    localPath = path.relative(
-      basePath,
-      path.resolve(basePath, '..', modulePath, '..', dependency.path)
-    );
-
-    return require(localPath);
-  }
-
+  const moduleUnderTest = require.resolve(modulePath, { paths: [basePath] });
+  const dependencyPath = require.resolve(dependency.path, { paths: [path.dirname(moduleUnderTest)] });
```
rfk: Turns out node already has methods for doing this, who knew?

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -66,35 +66,40 @@ function hexString (bytes) {
...
-  const mockLog = mocks.mockLog();
-  const accountRoutes = makeRoutes({
-    config: config,
-    devices: mockDevices,
-    log: mockLog
+  let config, mockRequest, devicesData, mockDevices, mockLog, accountRoutes, route;
+
+  beforeEach(() => {
+    config = {};
+    mockRequest = mocks.mockRequest({
```
rfk: Shared mutable state between these test was upsetting me.

rfk:
The tests seem to be failing due to https://github.com/mozilla/fxa/issues/1561 :-(

packages/fxa-content-server/app/scripts/lib/oauth-errors.js
```diff
@@ -97,6 +97,14 @@ var ERRORS = {
     errno: 121,
     message: 'Invalid grant_type'
   },
+  SERVER_UNAVAILABLE: {
+    errno: 201,
+    message: t('System unavailable, try again soon')
+  },
+  DISABLED_CLIENT_ID: {
+    errno: 202,
+    message: t('Client has been temporarily disabled.')
```
rfk: This is a new string, which I guess it not a good idea in a point-release. Perhaps we can have it say "System unavailable, try again soon" for now?

packages/fxa-payments-server/package-lock.json
```diff
@@ -2905,6 +2905,12 @@
         "@types/node": "*"
       }
     },
+    "@types/classnames": {
+      "version": "2.2.8",
+      "resolved": "https://registry.npmjs.org/@types/classnames/-/classnames-2.2.8.tgz",
+      "integrity": "sha512-3UrLzPnz8u+MMXuJTF++389IfLSQUbl5F3ry9WCxva0BKG5H/oo5NuPRXk+HrpPU1+5pVHSWhnVWRzIaFQ7QuQ==",
+      "dev": true
+    },
```
rfk: I regenerated this to fix https://github.com/mozilla/fxa/issues/1561, not sure if I did it right...

rfk:
I've manually QA'd that each of these environment variables work as expected in a local FxA stack, and here's the experience when I try to sign in with the Fenix OAuth client disabled:

Screenshot_2019-06-13-17-21-47


Not pretty, but highly functional, and IMHO a less frustrating user experience than having sync silently fail to work reliably after you're signed in.

rfk:
From local testing, the `OAUTH_DEVICE_*_ENABLED` prefs do not produce any user-visible impact - Fenix logs the unexpected responses to logcat and goes about its business.

Setting the `OAUTH_DISABLED_CLIENTS` pref will unfortunately cause any already-signed-in users to be disconnected from sync, because Fenix treats any failure to use its refresh_token as fatal. That's unfortunate, but I don't see a way around it on the server.

packages/fxa-auth-server/fxa-oauth-server/test/routes/authorization.js
```diff
@@ -6,115 +6,175 @@
...
         assertion: BASE64URL_STRING,
-        state: 'foo',
-        code_challenge: 'foo',
-        code_challenge_method: PKCE_CODE_CHALLENGE_METHOD,
-      }, 'code_challenge', 'length must be 43 characters long');
+      }, 'state');
     });
 
-    it('works with response_type code (non-default)', () => {
+    it('fails with no state', () => {
```
vbudhram: Should this be accepts state?

packages/fxa-auth-server/fxa-oauth-server/test/routes/authorization.js
```diff
@@ -6,115 +6,175 @@
...
+        Joi.assert({
+          client_id: CLIENT_ID,
+          assertion: BASE64URL_STRING,
+          state: 'foo',
+          code_challenge: PKCE_CODE_CHALLENGE,
+          code_challenge_method: PKCE_CODE_CHALLENGE_METHOD,
+        }, validation);
+      });
+
+      it('accepts code_challenge and code_challenge_method', () => {
```
vbudhram: Should be fails for invalid `code_challenge_method`?

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -471,10 +483,16 @@ module.exports = (log, db, config, customs, push, pushbox, devices, oauthdb) =>
         }
       },
       handler: async function (request) {
-        log.begin('Account.devices', request);
-
+        //log.begin('Account.devices', request);
```
vbudhram: Missed uncomment

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -848,6 +853,26 @@ describe('/account/devices/invoke_command', () => {
...
   });
+
+  it('supports feature-flag for oauth devices', async () => {
+    const mockPushbox = mocks.mockPushbox();
+    const route = getRoute(makeRoutes({
+      config: { oauthDeviceCommandsEnabled: false },
+      customs: mockCustoms,
+      log: mockLog,
+      pushbox: mockPushbox,
+    }), '/account/device/commands');
```
vbudhram: Should this be `/account/devices/invoke_command`?

packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -258,7 +258,7 @@ var ERRORS = {
     errno: 202,
     message: t('Feature not enabled')
   },
-  BACKEND_SERVICE_FAILURE: {
+  DISABLED_CLIENT_ID: {
```
vbudhram: Is it intended to remove the backend service failure error? Also disabled client is 204 on auth-server [errors](https://github.com/mozilla/fxa/blob/9360599f81f2b7841c4ebcff18ee593ee6feb90c/packages/fxa-auth-server/lib/error.js#L98).

packages/fxa-content-server/app/scripts/lib/oauth-errors.js
vbudhram: > Perhaps we can have it say "System unavailable, try again soon" for now?

I'm good with that and follow up with better message.

rfk:
> Generally speaking, if oauth client is disabled, should we disable account creation as well?
> Other than that, this LGTM

Hmm....I suppose we could add a similar check in the `/account/create` endpoint on the auth-server, but something about it feels wrong to me. I'll take a look.

packages/fxa-content-server/app/scripts/lib/auth-errors.js
rfk: This must have been some sort of copy-paste fail, nice catch.

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
rfk: nice catch!

rfk:
Thanks @vbudhram! After our slack conversation and a meeting with stakeholders, I've added a separate `OAUTH_CLIENTS_SIGNIN_DISABLED` flag on the auth-server.  This one disables new signins/signups from the specific OAuth clients, but allows existing connections to continue working uninterrupted.  Enabling it also helps with the bad UX you point out in the comment above.

If this looks good to you, could you please go ahead and tag a point release and update [Bug 1558541](https://bugzilla.mozilla.org/show_bug.cgi?id=1558541)? I'll try to check in on it tomorrow but have some family engagements that might interfere with my ability to do so.

rfk:
> Enabling it also helps with the bad UX you point out in the comment above, by disabling signup before we get to the email confirmation polling page.

Here's what I get locally on account creation attempt (the "continue to Fenix" part is due to my local config, production will still be "continue to Firefox Sync" IIRC):

Screenshot_2019-06-14-14-13-27

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -254,12 +254,18 @@ module.exports = (log, db, config, customs, push, pushbox, devices, oauthdb) =>
...
-        const uid = sessionToken.uid;
-        const deviceId = sessionToken.deviceId;
+        const credentials = request.auth.credentials;
+        const uid = credentials.uid;
+        const deviceId = credentials.deviceId;
         const query = request.query || {};
         const {index, limit} = query;
 
+        if (config.oauthDeviceCommandsEnabled === false) {
+          if (credentials.refreshTokenId) {
```
shane-tomlinson: How come the nested if here? I find it slightly more confusing than just using an &&. Also, shouldn't the refreshToken auth strategy take care of this?

packages/fxa-auth-server/lib/scheme-refresh-token.js
```diff
@@ -15,10 +15,14 @@ const ScopeSet = require('fxa-shared').oauth.scopes;
 // so we limit to the scope below as a safety mechanism
 const ALLOWED_REFRESH_TOKEN_SCHEME_SCOPES = ScopeSet.fromArray(['https://identity.mozilla.com/apps/oldsync']);
 
-module.exports = function schemeRefreshTokenScheme(db, oauthdb) {
+module.exports = function schemeRefreshTokenScheme(config, db, oauthdb) {
   return function schemeRefreshToken(server, options) {
     return {
       async authenticate (request, h) {
+        if (config.oauthDeviceAccessEnabled === false) {
```
shane-tomlinson: mmm.... I understand why you added this here, to keep things DRY, but it makes an assumption that the only thing the refreshToken scheme will ever be used for is for OAuth device access. Fine for now, not sure we should make that a long term policy though.

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -294,9 +300,15 @@ module.exports = (log, db, config, customs, push, pushbox, devices, oauthdb) =>
...
         const {target, command, payload} = request.payload;
         let {ttl} = request.payload;
-        const sessionToken = request.auth.credentials;
-        const uid = sessionToken.uid;
-        const sender = sessionToken.deviceId;
+        const credentials = request.auth.credentials;
+        const uid = credentials.uid;
+        const sender = credentials.deviceId;
+
+        if (config.oauthDeviceCommandsEnabled === false) {
```
shane-tomlinson: Shouldn't the refreshToken scheme take care of this already?

packages/fxa-auth-server/config/index.js
```diff
@@ -583,6 +583,28 @@ const conf = convict({
     env: 'DEVICE_NOTIFICATIONS_ENABLED',
     default: true
   },
+  // A safety switch for disabling new signins/signins from particular clients,
```
shane-tomlinson: `signins/signins` should probably be `signups/signins`

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -636,6 +621,26 @@ describe('/account/device/commands', () => {
...
+    mockRequest.auth.credentials.refreshTokenId = 'aaabbbccc';
+
+    try {
+      await route.handler(mockRequest);
+      assert.fail('should have thrown');
+    } catch (err) {
+      assert.equal(err.output.statusCode, 503);
```
shane-tomlinson: What is the client behavior for any of these routes if a 503 is returned? Will they give up, immediately try again, or wait some time to try again?

packages/fxa-payments-server/package-lock.json
shane-tomlinson: It looks like that issue was fixed by https://github.com/mozilla/fxa/pull/1578, do you still need this?

rfk:
> It looks like that issue was fixed by #1578, do you still need this?

That fix went to master, and it seemed a lot bigger than the one in this PR, so I didn't want to risk it on a point-release for train-139.

> I can't find @philbooth's feature flagging work being pulled in here, it seems like a valid use,
> to be able to react quickly to load.

I'll admit, I had just totally spaced on the fact that this was now available! I'll investigate, but I'm not confident I'll have the bandwidth to move this PR to that solution in the short term.

rfk:
Oh, looks like the feature-flag infra hasn't been plumbed in to the auth-server yet? Ref https://github.com/mozilla/fxa-shared/blob/master/feature-flags/README.md#how-do-i-use-a-feature-flag-in-the-auth-server

packages/fxa-auth-server/config/index.js
```diff
@@ -583,6 +583,28 @@ const conf = convict({
...
+  // as a hedge against unexpected client behaviour.
+  oauthClientsSigninDisabled: {
```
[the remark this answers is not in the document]
rfk: > Is there a meaningful difference between this and disabledClients on the OAuth server? 

The intention is for this one to prevent "new connections" from the specified clients while continuing to operate normally for users who have already connected them; the oauth-server one refuses service to all instances of the specified clients, even if they have already connected to the user's account.

I don't think we ever settled on a really good name for the combination of signup/signin, but I'm happy to take suggestions. I think it needs to be named differently *somehow* to distinguish its slightly later touch from the similar flag on the oauth-server.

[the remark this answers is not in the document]
rfk: What do you think about `oauth.disableNewClientConnections`?

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -3410,10 +3410,10 @@ describe('/v1', function() {
           .then(function (res) {
             var result = res.result;
             assert.equal(result.length, 2);
-            assert.equal(result[0].id, client1Id.toString('hex'));
-            assert.deepEqual(result[0].scope, ['clients:write', 'profile']);
-            assert.equal(result[1].id, client2Id.toString('hex'));
-            assert.deepEqual(result[1].scope, ['scopeFromRefreshToken']);
+            assert.equal(result[0].id, client2Id.toString('hex'));
```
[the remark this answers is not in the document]
rfk: Sorry this deserved a note - there was a timing issue with this test that was causing it to fail locally for me but usually pass in CI, because the output gets sorted by timestamp and then name. Changing the name above forced it into a consistent order regardless of timestamp, and that order happens to be the reverse of what was there previously.

packages/fxa-auth-server/fxa-oauth-server/lib/error.js
```diff
@@ -295,4 +295,29 @@ AppError.invalidGrantType = function invalidGrantType() {
   });
 };
 
+AppError.serviceUnavailable = function serviceUnavailable(retryAfter) {
```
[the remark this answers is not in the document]
rfk: I left it in there mostly to reserve errno `201` for "service unavailable" like it is elsewhere, but I guess it's pretty easy to bring this code back later if we need it.

packages/fxa-auth-server/lib/scheme-refresh-token.js
rfk: Aye. I'll take a followup to remove it on master so we don't live with it forever, ref https://github.com/mozilla/fxa/issues/1585

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
rfk: > shouldn't the refreshToken auth strategy take care of this?

No, because this one is specific to the use of device-commands rather than wholesale disablement of the entire devices API.

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
rfk: > What is the client behavior for any of these routes if a 503 is returned?
> Will they give up, immediately try again, or wait some time to try again?

In my local testing, it logs the error and just tries again on its next scheduled or triggered device action; there is no cascading retry loop or anything too scary like that. We *should* add more thorough backoff handling in the client to guard against retry loops explicitly.

philbooth:
> Oh, looks like the feature-flag infra hasn't been plumbed in to the auth-server yet?

Yep, sorry, that's still outstanding.

rfk:
@shane-tomlinson thanks for the feedback, I've pushed an update with your comments addressed.

Issue #1306, opened 2019-05-29
Make Trailhead Design Responsive for Mobile Firefox


davismtl:
The design was only designed for desktop. Let's revise with UX the flow to see what needs to change for mobile and deploy it there.

**Due date: July 9th**

I started a product one-pager here:
https://docs.google.com/document/d/1wzZCkwsLyPjqmOE-0feoIzwKwIkIPDjf4clMrZjLhYI/edit#

davismtl:
For more context, the overall scope of the mobile trailhead work will be documented here:
https://docs.google.com/document/d/16WFA5r8lsptI6rXJYWdvz2JwzmuDggVaDeE1yCwv-G0/edit?ts=5cf14b71#heading=h.bfb4xhoekuwy

As of right now, that document is a bit outdated but will be refreshed over the next week.

We will add a section about changes happening to FxA flows to support it.

PR #1580, opened 2019-06-13
feat(trailhead): update links and trailhead verification page

vbudhram:
Connects with https://github.com/mozilla/fxa/issues/1306

Updates the account confirmed page (when viewed on mobile) to show a `Start browsing` button. It also updates our exisiting start browsing links to goto `https://www.mozilla.org/firefox/accounts`.

Screen Shot 2019-06-13 at 3 52 24 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/ready.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/ready.mustache
@@ -10,6 +10,11 @@
 
     {{#isSync}}
       <p class="account-ready-service">{{{escapedReadyToSyncText}}}</p>
+        {{#isTrailhead}}
+            <div class="button-row">
+                <button class="primary-button btn-start-browsing">{{#t}}Start browsing{{/t}}</button>
+            </div>
+        {{/isTrailhead}}
     {{/isSync}}
 
     {{^isSync}}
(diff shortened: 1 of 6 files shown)


sv-sdeiac:
Hello,
On mobile, there are some changes to about:config in order to trigger the new UI? From the Desktop view, we have the changes settings and we can confirm the new UI. Also, on iOS we don't have about:config to set the configurations.

vbudhram:
@sv-sdeiac The easiest way to test this is actually opening the account verification links in the Firefox iOS browser. You should see the new UI above.

To connect Firefox iOS to a stage environment is slightly different than on desktop and android. If you goto settings then tap `Version XX.X` button a bunch of times.

IMG_CF18AC0B9CA6-1

I'm not really sure how many but at least 5 times? If you do it enough you will see new debug options in the settings menu. If you scroll to the top, under the Firefox Account section, there is a toggle button to use the stage environment.

IMG_E5D3B9423841-1

Enable the toggle and login to the stage account. Once you do this, you can send tabs, bookmarks etc to a stage desktop Firefox browser.

Let me know if you have any issues!

sv-sdeiac:
Hello, @vbudhram. The staging server is set on iOS but we can't trigger the new UI. For that reason, I said that we can't set the specific configurations like on Desktop. Sorry if I make confusions.

SorinaFlorean:
We can't trigger the new UI on mobile (nor iOS or Android) even if the stage is properly set up. We tried two different scenarios with no success:
- create an account and open the link in Firefox browser - the same device
- create an account on desktop, open the verification link on mobile
We looked through [documentation](https://github.com/mozilla/fxa/issues/1306#issuecomment-506757160) and we didn't find any extra setting to add in about:config. Could be some flag? 

Note: the stage settings are:
Screenshot_20

SorinaFlorean:
After more investigation, we found that we need to add **style=trailhead** in order to trigger the UI. Verified as fixed on stage - Train 140.

PR #1581, opened 2019-06-14
fix(settings): Use the same padding for settings panel body as the header

### LINKED ISSUE #1555: settings panel padding should match the title.
The panel title is 20px in (I think), the rest of the panel looks like it's 10.

Screenshot 2019-06-10 at 16 39 35

shane-tomlinson:
Use a left/right padding of 32px in "big" mode, 16px in "small"

fixes #1555

### Desktop

Screenshot 2019-06-14 at 14 40 52

### Mobile

Screenshot 2019-06-14 at 14 41 00

@hritvi - r? (your first FxA review!)

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -321,7 +321,13 @@ body.settings #stage .settings {
 .settings-unit-details {
   clear: both;
   display: none;
-  padding: 0 16px;
+
+  @include respond-to('big') {
+    padding: 0 32px;
+  }
+  @include respond-to('small') {
+    padding: 0 16px;
+  }
 
   p {
     clear: both;


packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -321,7 +321,12 @@ body.settings #stage .settings {
 .settings-unit-details {
   clear: both;
   display: none;
-  padding: 0 16px;
+  @include respond-to('big') {
```
hritvi: r+ for the code. But I guess you have to add a newline before `@include respond-to('big') {` so that the code goes with the rest.

shane-tomlinson:
@mozilla/fxa-devs - r?

packages/fxa-content-server/app/styles/modules/_settings.scss
shane-tomlinson: Done, thanks @hritvi!

sv-sdeiac:
Verified as fixed on Staging server - [Train 140].
Environments:
- Windows 10;
- Nokia 6 (Android 7.1.1);
- LG G7 fit (Android 8.1).

PR #1593, opened 2019-06-20
chore(style): added prettier to fxa-shared

dannycoates:
Here's the first package using `prettier`. I'll be creating a new PR for each package and then a final PR to move the settings up to the top level and adding a precommit hook to automatically format the files of each commit.

`eslint-config-prettier` is used to get eslint to play nice with the prettier formatted code, so from now on eslint will only be checking for correctness rules and leaving prettier to handle the style rules.


lmorchard:
FWIW, I think the CI failure is due to issue #1594

packages/fxa-shared/feature-flags/index.js
```diff
@@ -28,22 +28,25 @@ module.exports = initialise;
...
+function initialise(config, log, defaults) {
   const { interval, redis: redisConfig } = config;
 
-  if (! (interval > 0 && interval < Infinity)) {
+  if (!(interval > 0 && interval < Infinity)) {
     throw new TypeError('Invalid interval');
   }
 
-  if (! log) {
+  if (!log) {
```
philbooth: I'll be sorry to see these go fwiw. I was converted to whitespace-after-`!` when I joined Mozilla and have since been of the opinion that it genuinely improves readability. Is it configurable, or a necessary casualty of war?

packages/fxa-shared/README.md
```diff
@@ -62,7 +61,7 @@ implied by another scope:
 
 Or you can find out
 what values in the set
-are *not* implied by another scope:
+are _not_ implied by another scope:
```
philbooth: Interesting that it's going to modify all our docs too. Are we all agreed there's value in that? Is turning it off for `.md` files an option (and would we want to consider it if it was)?

packages/fxa-shared/feature-flags/index.js
shane-tomlinson: > or a necessary casualty of war?

[Prettier's options list](https://prettier.io/docs/en/options.html) is intentionally very limited, this doesn't seem to be one of the options.

packages/fxa-shared/README.md
shane-tomlinson: I was a bit surprised by this too though have come around to it.

PR #1601, opened 2019-06-24
fix(scripts): remove non-posix loop redirection from release script

### LINKED ISSUE #1590: Release script requires bash
The hash-bang line in `release.sh` specifies `/bin/sh`, but it seems to use some bash-specific features:

```
rfk@pomelo:fxa(train-139)$ ./release.sh patch
./release.sh: 207: ./release.sh: Syntax error: redirection unexpected
rfk@pomelo:fxa(train-139)$ 
rfk@pomelo:fxa(train-139)$ bash ./release.sh patch
[train-139 0422e18b2] Release 1.139.1
[...succeeds...]
```

philbooth:
Maybe fixes #1590, but I don't have a failing machine to try it on.

There's also a seondary fix I noticed when trying to test this out on a clean machine, pulling from `private/master` failed because I hadn't done a `git fetch private master`.


shane-tomlinson:
@philbooth - do you want someone on Windows to test this?

rfk:
FWIW this seems to allow `./release.sh` to do the right thing for me under WSL, 👍  thanks @philbooth!

philbooth:
@mozilla/fxa-devs r?

PR #1606, opened 2019-06-24
chore(style): added prettier to most of the packages

dannycoates:


dannycoates:
I couldn't enable `prettier` on html for auth-server because it breaks l10n :/

dannycoates:
Next PR will move the prettier deps up to root level and add a precommit hook

packages/fxa-auth-server/CONTRIBUTING.md
```diff
@@ -7,66 +7,73 @@ mailing list or through issues here on GitHub.
...
-    $ cd fxa-auth-server
-    $ git remote add user1 git@github.com:user1/fxa-auth-server.git
-```
+   ```sh
+   $ git clone https://github.com/mozilla/fxa-auth-server.git
+   $ cd fxa-auth-server
+   $ git remote add user1 git@github.com:user1/fxa-auth-server.git
+   ```
+
+````
```
philbooth: This is weird, it seems to have got confused by the mismatched indentation of the backticks below and decided it needs to insert 4 of them here instead. Presumably if we indent the bottom one it will do the right thing here.

philbooth: Might be worth grepping for 4 backticks together before committing this I guess, just in case there's any more.

dannycoates: there's a few but i don't want to redo all the commits to fix them in the right ones so i'll fix it at the end

vbudhram:
@dannycoates Oh oops, I though you were doing a PR for each project. If CI is green, I think this has my 👍🏽

packages/fxa-auth-server/CONTRIBUTING.md
dannycoates: i just noticed these were already borked before formatting ;)

PR #1607, opened 2019-06-24
feat(event-broker): use firestore collection prefix

### LINKED ISSUE #1557: [fxa-event-broker] Allow configurable Firestore collection prefix
Firestore has a namespace per project rather than a 'database'. Since FxA has a single project per deploy and other FxA services might use Firefox, each service should allow a prefix for use by ops.

bbangert:
Closes #1557

### DIFF
--- a/packages/fxa-event-broker/lib/db/firestore.ts
+++ b/packages/fxa-event-broker/lib/db/firestore.ts
@@ -6,10 +6,16 @@ import { Firestore, Settings } from '@google-cloud/firestore';
 import { ClientWebhooks } from '../selfUpdatingService/clientWebhookService';
 import { Datastore } from './index';
 
+type FirestoreDbSettings = Settings & {
+  prefix: string;
+};
+
 class FirestoreDatastore implements Datastore {
   private db: Firestore;
+  private prefix: string;
 
-  constructor(config: Settings, firestore?: Firestore) {
+  constructor(config: FirestoreDbSettings, firestore?: Firestore) {
+    this.prefix = config.prefix;
     if (firestore) {
       this.db = firestore;
     } else {
@@ -22,7 +28,7 @@ class FirestoreDatastore implements Datastore {
   }
 
   public async storeLogin(uid: string, clientId: string) {
-    const document = this.db.doc(`users/${uid}`);
+    const document = this.db.doc(`${this.prefix}users/${uid}`);
     const doc = await document.get();
     if (doc.exists) {
       const data = doc.data();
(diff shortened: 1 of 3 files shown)


packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -8,8 +8,13 @@ import { Datastore } from './index';
 
 class FirestoreDatastore implements Datastore {
   private db: Firestore;
+  private prefix: string;
 
   constructor(config: Settings, firestore?: Firestore) {
+    this.prefix = config.prefix || 'fxa-eb-';
+    if (config.prefix) {
+      delete config.prefix;
```
vladikoff: what's the reasoning for deleting this?

packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -8,8 +8,13 @@ import { Datastore } from './index';
 
 class FirestoreDatastore implements Datastore {
   private db: Firestore;
+  private prefix: string;
 
   constructor(config: Settings, firestore?: Firestore) {
+    this.prefix = config.prefix || 'fxa-eb-';
```
vladikoff: there's an already default in the config, do you need the fallback `fxa-eb`?

packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -22,7 +27,7 @@ class FirestoreDatastore implements Datastore {
   }
 
   public async storeLogin(uid: string, clientId: string) {
-    const document = this.db.doc(`users/${uid}`);
+    const document = this.db.doc(this.prefix + `users/${uid}`);
```
vladikoff: would it be cleaner to use `${this.prefix}.....`?

packages/fxa-event-broker/lib/db/firestore.ts
bbangert: The Firestore init does not need it, even though it technically allows extra keys.

packages/fxa-event-broker/lib/db/firestore.ts
bbangert: Because in TypeScript the key here is not guaranteed to exist. I'll change the Settings type to avoid it.

packages/fxa-event-broker/lib/db/firestore.ts
bbangert: Sure, why not. 😄

PR #933, opened 2019-04-25
feat(oauth): JWT access tokens.

shane-tomlinson:
Super early stage experiment to see how much effort it would be to
convert access tokens to JWTs as defined in https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00.

Loads of caveats.

JWT isn't signed, still requires DB backing, the code is a mess. You know, normal things.

Hey @rfk, this is ridiculously early stage, but I'm opening it as an exploration because we might need something like this soonish, and I want to see what's possible. It's so early it's not really even ready for concrete feedback, but would appreciate your thoughts anyways.

issue #935

PR #1612, opened 2019-06-25
feat(oauth): JWT access tokens.

### LINKED ISSUE #935: Return JWTs for access tokens
There is this new [draft spec to return JWT formatted access tokens](https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00). 

JWT access tokens have a bunch of nice properties such as:

* The can be signed by us and can be validated by RPs without requiring a hit on our DB.
* They contain a bunch of metadata that can be used directly by the RP without involving us.
* In theory nothing would need to be stored in the DB, though revocation becomes more difficult.
* We can add arbitrary data in the access token w/o needing to add new fields in the DB.
* We could make usage more secure by combining them with [Proof-of-Possession Key Semantics for JSON Web Tokens (JWTs)](https://tools.ietf.org/html/rfc7800), though we might be able to do something similar w/o JWTs using the [OAuth 2.0 Proof-of-Possession: Authorization Server to Client Key Distribution](https://tools.ietf.org/html/draft-ietf-oauth-pop-key-distribution-07) spec.

Another downside to this if RPs do not need to hit one of our endpoints to verify an access token, [counting MAU](https://github.com/mozilla/fxa-auth-server/issues/2957) becomes more difficult. 

Perhaps both the counting and revocation problems can be worked around by making access tokens be short lived (on the order of minutes), forcing RPs to use a refresh token to get new access tokens.

shane-tomlinson:
Super early stage experiment to see how much effort it would be to
convert access tokens to JWTs.

Loads of caveats.

Still requires DB backing, the code is a mess. You know, normal things.

Reopened #933 with a "feature." branch name.

fixes #935


shane-tomlinson:
@rfk - could I get your feedback on this PR? It has turned out better than I expected, though I know there is room for some improvement. I didn't want to go too far down the rabbit hole of trying to make everything ready for "JWT everywhere", but just enough to get started. I plan to fast follow with PPIDs, but want to get this part landed before the train cut.

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
```diff
@@ -176,13 +177,31 @@
       "privacyUri": "",
       "trusted": true,
       "publicClient": false
+    },
+    {
+      "id": "a8c528140153d1c6",
```
shane-tomlinson: Can probably remove Secure Proxy from config and only add it to the config needed on stomlinson.dev.lcip.org.

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
```diff
@@ -176,13 +177,31 @@
...
   ],
   "localRedirects": true,
   "logging": {
     "level": "ALL",
     "fmt": "pretty"
   },
+  "jwtAccessTokens": {
+    "enabled": true,
+    "enabledClientIds": [
+      "98e6508e88680e1a",
```
shane-tomlinson: Comment these are for fxa-settings, 123done, and secure proxy. Though remove secure proxy and add that to stomlinson.dev.lcip.org config.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+  // https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00#section-2.2
+  const claims = {
+    // The IETF spec for `aud` refers to https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+    // > REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the
+    // > OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain
+    // > identifiers for other audiences. In the general case, the aud value is an array of
+    // > case sensitive strings. In the common special case when there is one audience, the
+    // > aud value MAY be a single case sensitive string.
+
+    // TODO: accept a "resource" parameter to endpoints where access tokens are returned.
```
shane-tomlinson: File a follow on for this, leave a comment about why it's just the clientId and not a URL.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Get the token ID of the JWT access token.
+ *
+ * @throws `invalidToken` error if access token is invalid.
+ */
+exports.tokenId = async function tokenId(accessToken) {
+  // The access token ID is stored in the jti field of
+  // a JWT access token.
+  const jwt = await exports.verify(accessToken);
+  if (! jwt.jti) {
```
shane-tomlinson: Run prettier, I don't think there should be a space after `!`

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -0,0 +1,75 @@
...
+if (oldPrivKey && Object.keys(oldPrivKey).length) {
+  PRIVATE_KEYS_RAW.push(oldPrivKey);
+}
+
+const SIGNING_KEY = JwTool.JWK.fromObject(config.get('openid.key'), {
+  iss: ISSUER,
+});
+
+const PUBLIC_KEYS_RAW = new Map();
+const VERIFICATION_PUBLIC_KEYS = new Map();
```
shane-tomlinson: Move `VERIFICATION_` as a suffix, or use a _JWK suffix instead for clarity.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+    header = JWTool.unverify(accessToken).header;
+  } catch (e) {
+    throw AppError.invalidToken();
+  }
+  const signingKey = keys.publicJWK(header.kid);
+  // signingKey is undefined if header does not contain `kid`
+  if (!signingKey) {
+    throw AppError.invalidToken();
+  }
+  const payload = signingKey.verifySync(accessToken);
```
rfk: We should `await signingKey.verify()` here rather than using `verifySync`, since we're already in `async` function.

However, I see that this just [calls `verifySync` under the hood anyway](https://github.com/mozilla/fxa-jwtool/blob/d9f822c9e5899bb3f9bea709b0bab102970b0151/index.js#L132).

There's some amount of operational risk here if we end up doing a lot of RSA operations inline in the event loop, rather than dispatching them to background threads (like browserid-verifier does using [node-compute-cluster](https://github.com/lloyd/node-compute-cluster)). It's probably fine for launch scale, and is solvable in a pinch by buying more servers, but something for us to think about for the future.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
rfk: To clarify, these are "live" `PublicJwk` objects rather than just the JSON fields?

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = {
+  /**
+   * Get a public JWK by `kid`. Call `.verify`
+   * w/ JWT to verify.
+   *
+   * @param {String} kid of JWK to get
+   */
+  publicJWK(kid) {
+    return VERIFICATION_PUBLIC_KEYS.get(kid);
```
rfk: I'd be very tempted to have this throw rather than return `undefined` on error, just to make extra sure that the caller has to deal with that error path.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+  if (!payload) {
+    throw AppError.invalidToken();
+  }
+  return payload;
```
rfk: There are additional things we should check here, to set a good example for other consumers of these tokens.  Things that might affect security include:

* The `typ` header must be present and exactly equal `"at+jwt"` (might need to fix fxa-jwtool to return this value)

Things that would be good hygiene but *shouldn't* ever happen in practice (because we know that it's signed with one of our keys, and we'd never sign invalid data) include:

* The `iss` claim should be present and exactly match our issuer value.
* Other expected claims should be present and validate against expected schema.

I suggest validating the claims against an explicit schema, like we do for browserid assertions [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/fxa-oauth-server/lib/assertion.js#L34).

packages/fxa-auth-server/fxa-oauth-server/lib/token.js
```diff
@@ -9,61 +9,94 @@ const config = require('./config');
...
-      // on access tokens and such old tokens won't be allowed.
-      logger.warn('token.verify.expiring_soon', {
-        user: token.userId.toString('hex'),
-        client_id: token.clientId.toString('hex'),
-        scope: token.scope,
-        created_at: token.createdAt,
-        expires_at: token.expiresAt,
-      });
+async function getAccessTokenId(accessToken) {
+  if (/\./.test(accessToken)) {
```
rfk: nit: I think this deserves a comment about why the test works for detecting JWT access tokens.

packages/fxa-auth-server/fxa-oauth-server/lib/token.js
```diff
@@ -9,61 +9,94 @@ const config = require('./config');
...
+  const tokenId = await getAccessTokenId(accessToken);
+  return encrypt.hash(tokenId);
+};
+
+/**
+ * Verify an accessToken.
+ */
+exports.verify = async function verify(accessToken) {
+  // If a JWT accessToken is used, getAccessTokenId verifies
+  // the form of the access token. If the token signature
```
rfk: I don't understand "the form of" here; would it be easier to just say that it verifies the token and throws if there's anything wrong with it?

packages/fxa-auth-server/fxa-oauth-server/lib/validators.js
```diff
@@ -68,3 +67,8 @@ exports.jwe = Joi.string()
   .regex(
     /^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/
   );
+
+exports.accessToken = Joi.alternatives().try([
+  exports.token,
+  exports.assertion,
```
rfk: nit: this will allow BrowserID assertions to validate as valid access tokens. Probably harmless, because they will fail JWT verification, but it may be worth adding an explicit test to that effect.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -917,6 +919,28 @@ describe('/v1', function() {
...
+              client_id: jwtClient.id,
+              response_type: 'token',
+            }),
+          });
+
+          const defaultExpiresIn = config.get('expiration.accessToken') / 1000;
+          assert.equal(res.statusCode, 200);
+          assertSecurityHeaders(res);
+          assert(res.result.access_token);
+          assert.include(res.result.access_token, '.');
```
rfk: As above, naive readers would probably appreciate a comment that this dot uniquely indicates a JWT.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -4657,4 +4681,91 @@ describe('/v1', function() {
...
+        payload: {
+          token: tokenResult.result.access_token,
+          token_type_hint: 'access_token',
+        },
+      });
+
+      assert.equal(introspectInvalidTokenResult.statusCode, 200);
+      assert.isFalse(introspectInvalidTokenResult.result.active);
+    });
+  });
```
rfk: 👍  great!

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
rfk: Why fxa-settings, is it just so we can test it out ourselves concurrently with the proxy team's work?

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+  // a JWT access token.
+  const jwt = await exports.verify(accessToken);
+  if (! jwt.jti) {
+    throw AppError.invalidToken();
+  }
+  return jwt.jti;
+};
+
+/**
+ * Verify the form of a JWT access token.
```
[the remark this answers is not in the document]
shane-tomlinson: It verifies its signature as well, but does not verify, e.g., that the token is not expired. I'll update the comment to say that, and why, right now the canonical source of data to verify whether the token is still valid and not expired is the DB.

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
shane-tomlinson: > Why fxa-settings, is it just so we can test it out ourselves concurrently with the proxy team's work?

Exactly. And because settings sends access tokens to the auth server whereas the others work directly with the oauth server. Sending access tokens to the auth server showed that I hadn't updated the validator logic there to handle JWT access tokens.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -10,7 +10,7 @@ const AppError = require('./error');
 const db = require('./db');
 const util = require('./util');
 const ScopeSet = require('fxa-shared').oauth.scopes;
-const JwTool = require('fxa-jwtool');
+const JWTToken = require('./jwt-token');
```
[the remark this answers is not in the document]
shane-tomlinson: I might rename it to JWTAccessToken for clarity

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -132,9 +138,10 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
 // the specified grant has been sufficiently vetted by calling code.
 module.exports.generateTokens = async function generateTokens(grant) {
   // We always generate an access_token.
-  const access = await db.generateAccessToken(grant);
+  const access = await exports.generateAccessToken(grant);
+
   const result = {
-    access_token: access.token.toString('hex'),
+    access_token: access.jwt_token || access.token.toString('hex'),
```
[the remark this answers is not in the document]
shane-tomlinson: I did that initially, the problem is below we have to de-hex encode it to a buffer so that we can create the hash for looking up the token in the DB. This way ended up being a bunch cleaner.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // TODO: accept a "resource" parameter to endpoints where access tokens are returned.
+    // Convert "aud" to an array
+    aud: clientId,
+    client_id: clientId,
+    exp: Math.floor(accessToken.expiresAt / 1000),
+    iat: Math.floor(Date.now() / 1000),
+    iss: keys.ISSUER,
+    jti: hex(accessToken.token),
+    // scope is not in the spec but added so that the service provider can
```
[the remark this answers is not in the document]
shane-tomlinson: Great, another thing to note for feedback on the draft.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt-token.js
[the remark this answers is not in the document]
rfk: I am surprised to find that `fxa-jwtool` doesn't do the expired check, but indeed it does not.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
shane-tomlinson: Yes

packages/fxa-auth-server/fxa-oauth-server/lib/validators.js
shane-tomlinson: Done

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
```diff
@@ -141,7 +141,8 @@
       "imageUri": "",
       "redirectUri": "urn:ietf:wg:oauth:2.0:oob",
       "trusted": true,
-      "canGrant": true
+      "canGrant": true,
+      "publicClient": true
```
vladikoff: make sure to change this in fxa-dev?

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {Object} claims
+ * @returns {Promise<JWT>}
+ */
+exports.sign = function sign(claims) {
+  return jwt.sign(
+    {
+      // force an issuer to be set for direct calls to .sign,
+      // it can be overridden in the passed in claims.
+      iss: ISSUER,
```
rfk: It's not a big deal, but I'd be tempted to move the setting (and checking) of `iss` up into `jwt.js`.  The issuer string is tightly bound to the signing keys so it would be conceptually cleaner to have them all specified at the same level of the stack here.  Anyone who wanted to set a custom issue would almost certainly want to also specify a different signing key.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {String} accessToken
+ * @throws `invalidToken` error if access token is invalid.
+ * @returns {Promise<Object>}
+ */
+exports.verify = async function verify(accessToken) {
+  let payload;
+  try {
+    payload = await jwt.verify(accessToken, {
+      issuer: ISSUER,
```
rfk: Ditto here re: customizing issuer but using the same signing keys.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -182,5 +187,20 @@ function generateIdToken(grant, access) {
     claims.acr = 'AAL' + grant.aal;
   }
 
-  return ID_TOKEN_KEY.sign(claims);
+  return sign(claims);
```
rfk: (Commenting on this line, but it's about the calculation of `at_hash` a few lines above).

I'm pretty sure this calculation of `at_hash` is wrong, and will be even wronger when generating JWT access tokens :-(

Per https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken

> Its value is the base64url encoding of the left-most half of the hash of the octets of the ASCII representation
> of the access_token value, where the hash algorithm used is the hash algorithm used in the alg Header
> Parameter of the ID Token's JOSE Header. For instance, if the alg is RS256, hash the access_token value with
> SHA-256, then take the left-most 128 bits and base64url encode them. The at_hash value is a case sensitive
> string.

Since we expect clients to (1) validate this hash locally and (2) treat access tokens as opaque strings, the only sensible value to hash here is the ascii string `access_token` value as seen by the client.  For DB tokens that's the hex string, and for JWTs that's the full JWT.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -0,0 +1,70 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: This is going to be a merge conflict nightmare with https://github.com/mozilla/fxa/pull/1668, but I volunteer to take on the rebase of that PR once this one has landed. Just a heads-up.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -0,0 +1,70 @@
...
+    alg: 'RS256',
+    kid: key.kid,
+    'fxa-createdAt': key['fxa-createdAt'],
+    use: 'sig',
+    n: key.n,
+    e: key.e,
+  };
+}
+
+const PRIVATE_KEYS_RAW = [config.get('openid.key')];
```
rfk: nit: for uninitiated readers, it might be clearer to use `_JSON` rather than `_RAW` as a suffix on these; it would be easy to think that `_RAW` indicated e.g. a buffer rather than an object.

rfk: Oh, or `_OBJ` like you use in some below.

packages/fxa-auth-server/fxa-oauth-server/lib/token.js
```diff
@@ -9,61 +9,96 @@ const config = require('./config');
...
-        client_id: token.clientId.toString('hex'),
-        scope: token.scope,
-        created_at: token.createdAt,
-        expires_at: token.expiresAt,
-      });
+async function getAccessTokenId(accessToken) {
+  // JWT access tokens contain at least 1 period whereas
+  // legacy access tokens are only a
+  // combination of hex characters.
+  if (/\./.test(accessToken)) {
```
rfk: you could also consider using the new `validators.jwt` here rather than a custom check; I don't feel strongly either way

packages/fxa-auth-server/fxa-oauth-server/lib/token.js
```diff
@@ -9,61 +9,96 @@ const config = require('./config');
...
+  // If a JWT accessToken is used, getAccessTokenId verifies
+  // the access token. If the token is invalid in any way,
+  // (including expired), an `invalidToken` error will be thrown.
+  const accessTokenId = await getAccessTokenId(accessToken);
+
+  // JWT access tokens are still database backed, continue
+  // to use the database as the canonical source of info
+  // until we fully migrate to JWTs.
+  const token = await db.getAccessToken(
+    await exports.getTokenId(accessTokenId)
```
rfk: Looks like `exports.getTokenId` calls `getAccessTokenId` internally, so it's not clear why you call it explicitly above. Is it just so that we can get the verification behaviour explicitly rather than implicity?

I'd consider inlining `getAccessTokenId` into `getTokenId`, if only to reduce confusion over which of two very-similarly-named functions is responsible for what.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -4657,4 +4684,126 @@ describe('/v1', function() {
...
+
+  describe('Access w/ BrowserID assertion', () => {
+    it('succeeds', async () => {
```
rfk: I'm not sure "succeeds" gives the right impression here :-P

Maybe "is cleanly rejected" or similar?

packages/fxa-auth-server/fxa-oauth-server/test/jwt.js
```diff
@@ -0,0 +1,166 @@
...
+          }
+        );
+
+        const verifiedPayload = await JWT.verify(jwt, {
+          typ: 'custom JWT',
+        });
+        assert.strictEqual(verifiedPayload.foo, 'bar');
+      });
+    });
+  });
```
rfk: Excellent suite of error-case tests 👍

packages/fxa-auth-server/package.json
```diff
@@ -77,6 +77,7 @@
     "node-uap": "git+https://github.com/vladikoff/node-uap.git#9cdd16247",
     "nodemailer": "2.7.2",
     "otplib": "8.0.1",
+    "pem-jwk": "^2.0.0",
```
rfk: Just noting my own though process, that we're already using this dep in fxa-auth-server, which is nice.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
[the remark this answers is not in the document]
shane-tomlinson: > In fact in the case of issuing a JWT access token, it should be the hash of the whole JWT, not just the jti.

That was another thing I wanted to bring up with the spec author, whether to hash the whole token or whether the at_hash should just be the jti. The whole thing makes more sense to me.

packages/fxa-auth-server/fxa-oauth-server/lib/validators.js
```diff
@@ -68,3 +67,13 @@ exports.jwe = Joi.string()
   .regex(
     /^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/
   );
+
+exports.jwt = Joi.string()
+  .max(1024)
+  // JWT format: 'header.payload.signature'
+  .regex(/^([a-zA-Z0-9\-_]+)\.([a-zA-Z0-9\-_]+)\.([a-zA-Z0-9\-_]+)$/);
```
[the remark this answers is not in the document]
shane-tomlinson: Prettier takes care of the formatting now, my guess is because the whole thing fits within 100 or so characters whereas the above doesn't.

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
shane-tomlinson: Good catch, will do.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -182,5 +187,20 @@ function generateIdToken(grant, access) {
...
+exports.generateAccessToken = function generateAccessToken(grant) {
+  const clientId = hex(grant.clientId);
+
+  if (
+    !JWT_ACCESS_TOKENS_ENABLED ||
+    !JWT_ACCESS_TOKENS_CLIENT_IDS.has(clientId)
+  ) {
+    // return the old style access token if JWT access tokens are
+    // not globally enabled or if not enabled for the given clientId.
+    return db.generateAccessToken(grant);
```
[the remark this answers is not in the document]
shane-tomlinson: In reality, this is a great simplification. 👍 Done.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const JWTool = require('jsonwebtoken');
```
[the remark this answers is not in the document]
shane-tomlinson: Just renamed it to jsonwebtool. 👍

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
shane-tomlinson: https://github.com/mozilla/fxa-dev/pull/469

packages/fxa-auth-server/fxa-oauth-server/lib/token.js
shane-tomlinson: You are right, I hadn't seen that while developing. Will update.

PR #1613, opened 2019-06-25
Add prettier precommit hook

dannycoates:
This adds a precommit hook to auto-format changes and top level `npm run format`


dannycoates:
It seems like these CONTRIBUTING.md files should go away in favor of the top level one, but should be a separate issue. This fix is just to make @philbooth happy ;)

PR #1614, opened 2019-06-25
fix(ci): run content-server server tests

dannycoates:
let's see what happens


shane-tomlinson:
@dannycoates, I think once we fix https://github.com/mozilla/fxa/issues/1689, these tests should pass again and we can merge this.

PR #1619, opened 2019-06-26
feat(oauth): Reject new authorizations for disabled OAuth clients.

rfk:
The `$OAUTH_DISABLE_NEW_CONNECTIONS_FOR_CLIENTS` added in https://github.com/mozilla/fxa/pull/1577 does not cover the pairing flow, which allows clients to get connected from an already-connected client.  This PR adds additional checks of that flag so that it also covers the pairing flow.

Not sure if it's worth a point-release, but wanted to put it up for completeness.

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/oauth.js
+++ b/packages/fxa-auth-server/test/local/routes/oauth.js
@@ -26,10 +26,14 @@ const MOCK_TOKEN_RESPONSE = {
 };
 
 describe('/oauth/ routes', () => {
-  let mockOAuthDB, mockLog, sessionToken;
+  let mockOAuthDB, mockLog, mockConfig, sessionToken;
 
   async function loadAndCallRoute(path, request) {
-    const routes = require('../../../lib/routes/oauth')(mockLog, {}, mockOAuthDB);
+    const routes = require('../../../lib/routes/oauth')(
+      mockLog,
+      mockConfig,
+      mockOAuthDB
+    );
     const route = await getRoute(routes, path);
     if (route.options.validate.payload) {
       request.payload = await Joi.validate(request.payload, route.options.validate.payload);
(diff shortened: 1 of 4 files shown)


rfk:
(If we think we might want to backport this to a previous train, let's not merge to master beforehand)

packages/fxa-auth-server/test/local/routes/oauth.js
```diff
@@ -121,6 +124,27 @@ describe('/oauth/ routes', () => {
...
+        payload: {
+          client_id: MOCK_CLIENT_ID,
+          scope: MOCK_SCOPES,
+        },
+      });
+      try {
+        await loadAndCallRoute('/account/scoped-key-data', mockRequest);
+        assert.fail('should have thrown');
+      } catch (err) {
+        assert.equal(err.output.statusCode, 503);
```
bbangert: I'm used to seeing 503's representing server-side errors where a gateway failed, so the request should be retried as-is. Is this request going to work as-is later on its own?

rfk: Yes, once we remove the temporarily-disabled client from the list.

bbangert: Do we include a `Retry-After` header automatically?

rfk: I was about to say "yes we do" but it looks like this is not implemented on the `disabledClientId` error type; it is on some others such as `featureNotEnabled`: https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/error.js#L522

bbangert: I'm good either way, though for consistency it would probably be good to include it like the other 503's.

rfk: I went ahead and added it.

jrgm:
I'm +1 on doing a v1.139.1 release for this and having it in place along with my mis-config fix of puppet before the pending release of Fenix.

rfk:
Updated to target train-139 branch.

PR #1629, opened 2019-06-26
chore(deps): update to latest fxa dev launcher

vbudhram:
Pulls in the latest fxa dev launcher so that it can work in FF67+.


vladikoff:
👍

PR #1631, opened 2019-06-26
fix(style): Fixes #1258 - remove back link from CWTS on email first flow

### LINKED ISSUE #1258: Remove Back link from CWTS on email first flow
The back link's purpose is to allow users to change their email if they made a mistake during the first step of registration.

This link is only useful for legacy flows but it is useless for email first flows since users who want to change their email will do it on the registration step following the email first flow, not on the CWTS screen. 

(this only became obvious while redesigning for trailhead)

LZoog:
First PR! 🎉 Thanks to @dannycoates for walking me through the process. :D

---

Removes the back link and container div from CWTS email first flow.

Fixes #1258

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/choose_what_to_sync.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/choose_what_to_sync.mustache
@@ -46,9 +46,6 @@
           <button id="submit-btn" type="submit" tabindex="1" autofocus>{{#t}}Save selections{{/t}}</button>
         {{/isTrailhead}}
         </div>
-      <div class="links">
-        <a href="#" id="back" data-flow-event="back">{{#t}}Back{{/t}}</a>
-      </div>
     </form>
   </section>
 </div>


dannycoates:
It looks like we do need to fix this test
Screen Shot 2019-06-26 at 3 45 40 PM

which originates here: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/tests/spec/views/choose_what_to_sync.js#L137-L138

LZoog:
@dannycoates awesome, I'll take care of that today or tomorrow after completing all onboarding tasks.

Before merging, I also want to double check with whomever I need to if it's best practice to go ahead and remove the `button-row` class from the div containing the button, as it adds a `margin-bottom` property which is not visually apparent now that the element below it (the back button) has been removed. I know it's likely not important in this particular case but I'd like to familiarize myself with as many conventions as possible. :)

dannycoates:
@LZoog take a screenshot and post it here then @ rfeeley to have a look

LZoog:
Closed in favor of https://github.com/mozilla/fxa/pull/1649. My commit here was not yet GPG signed.

PR #1633, opened 2019-06-27
feat(clients): Add methods for handling attached clients in fxa-js-client

rfk:
The tests in https://github.com/mozilla/fxa/pull/1262/ are failing because it adds new methods to fxa-js-client, but fxa-content-server uses the latest published version rather than the version from the repo. I guess I need to gut a new fxa-js-client release before I can land that PR?  Here's just the fxa-js-client changes as a separate PR in order to facilitate that.

### DIFF
--- a/packages/fxa-js-client/tests/lib/attachedClients.js
+++ b/packages/fxa-js-client/tests/lib/attachedClients.js
@@ -1,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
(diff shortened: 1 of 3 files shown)


packages/fxa-js-client/tests/lib/attachedClients.js
```diff
@@ -0,0 +1,98 @@
...
+        sinon.spy(xhr.prototype, 'open');
+        sinon.spy(xhr.prototype, 'send');
+      });
+
+      afterEach(function() {
+        xhr.prototype.open.restore();
+        xhr.prototype.send.restore();
+      });
+
+      test('#attachedClients', function() {
```
rfk: These were copy-pastad from the existing tests for `/sessions`.

PR #1638, opened 2019-06-27
chore(version) bump fxa-js-client version to 1.0.14

rfk:
What is says on the tin.


vbudhram:
Thanks for publishing!

PR #1642, opened 2019-06-28
fix(errors): add missing subscription errors to content server

### LINKED ISSUE #1602: [content server] missing errors in app/scripts/lib/auth-errors.js
```js
SUBSCRIPTION_ALREADY_CANCELLED: 180,
REJECTED_CUSTOMER_UPDATE: 181,
```

```js
AppError.subscriptionAlreadyCancelled = () => {
  return new AppError({
    code: 400,
    error: 'Bad Request',
    errno: ERRNO.SUBSCRIPTION_ALREADY_CANCELLED,
    message: 'Subscription has already been cancelled'
  });
};

AppError.rejectedCustomerUpdate = (message) => {
  return new AppError({
    code: 400,
    error: 'Bad Request',
    errno: ERRNO.REJECTED_CUSTOMER_UPDATE,
    message
  });
};
```

philbooth:
Fixes #1602.

Note the error returned from the auth server for `REJECTED_CUSTOMER_UPDATE` returns a more specific message that comes from SubHub. Obviously we need to settle on something static for l10n though, so I opted for something suitably generic:

```
Update was rejected, please try again
```

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/auth-errors.js
+++ b/packages/fxa-content-server/app/scripts/lib/auth-errors.js
@@ -146,7 +146,6 @@ var ERRORS = {
     errno: 135,
     message: t('Unable to deliver email'),
   },
-  // Secondary Email errors
   EMAIL_EXISTS: {
     errno: 136,
     message: t('This email was already verified by another user'),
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -261,6 +261,14 @@ var ERRORS = {
...
   },
+  SUBSCRIPTION_ALREADY_CANCELLED: {
+    errno: 180,
+    message: t('Subscription has already been cancelled'),
+  },
+  REJECTED_CUSTOMER_UPDATE: {
+    errno: 181,
+    message: t('Update was rejected, please try again'),
+  },
   // Secondary Email errors end
```
vbudhram: Not related to issue, but this comment seems like it could be removed.

PR #1649, opened 2019-07-01
fix(style): fixes #1258 - remove back link from CWTS on email first flow

### LINKED ISSUE #1258: Remove Back link from CWTS on email first flow

LZoog:
First PR to be merged🎉
Closed https://github.com/mozilla/fxa/pull/1631 in favor of this one due to my commit not yet being GPG verified.

Fixes #1258 - Removes the back link and container div from CWTS email first flow.

With this commit, I've also updated the test that now passes. However, I ran the tests locally and am seeing 4 failed tests in `fxa-content-server` unrelated to my change here, 3 centered around avatars/profile images and 1 around `oauth query parameter`. I'm curious if these will pass in CircleCI and if so I'd like to determine why they're failing locally for me.

I've also removed the class `button-row` on the previous adjacent div due to `padding-bottom` the class added, here's a screenshot requested from the closed PR:
image

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/choose_what_to_sync.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/choose_what_to_sync.mustache
@@ -38,7 +38,7 @@
           </div>
         {{/engines}}
       </div>
-      <div class="button-row">
+      <div>
         {{#isTrailhead}}
           <button id="submit-btn" type="submit" tabindex="1" autofocus>{{#t}}Continue{{/t}}</button>
         {{/isTrailhead}}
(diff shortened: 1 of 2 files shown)


clouserw:
@vbudhram are you able to review this since Danny is out?

LZoog:
Logged my unit tests failures in an issue here: https://github.com/mozilla/fxa/issues/1660

PR #1653, opened 2019-07-01
feat(auth-server): add support form backend

### LINKED ISSUE #798: [Support] Create support form backend
#797 describes the form, but the support form has to submit somewhere.

The backend will receive the form and submit it directly to ZenDesk

Question: where (in what server) should this backend be hosted? Note that this requires private configuration (API keys), but no state (the submission should happen immediately, and any errors reported back to the user so they can attempt to resubmit).

See also [technical description on ZenDesk API](https://docs.google.com/document/d/18EIf2tPT9FSN-4tnUg_qNcDXugYymhVox1956z7VfTQ/edit#heading=h.k91114u9ssqo) and [exact specifications of how the API call should happen](https://docs.google.com/document/d/19pw1rLH33tKWYtGmlGaNTbOwNKU1VZdXHA23YKExADc/edit#heading=h.5pvslv2bxwik).

bbangert:
Closes #798


packages/fxa-auth-server/lib/routes/support.js
```diff
@@ -0,0 +1,96 @@
...
+          schema: isA.object().keys({
+            success: isA.bool().required(),
+            ticket: isA.number().optional(),
+            error: isA.string().optional(),
+          }),
+        },
+      },
+      handler: async function(request) {
+        log.begin('support.ticket', request);
+        const { uid, email } = await handleAuth(request.auth, true);
```
vbudhram: There doesn't appear to be an instance where `fetchEmail` is false in handleAuth (maybe it will in future?). If not, you could update this to always fetch the account.

packages/fxa-auth-server/lib/zendesk-client.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = config => {
+  const zendeskClient = zendesk.createClient({
+    username: config.zendesk.username,
+    token: config.zendesk.token,
+    remoteUri: `https://${config.zendesk.subdomain}.zendesk.com/api/v2`,
+    disableGlobalState: true,
+  });
+
+  // Promisify our callback functions, note util.promisify can't be used
```
vbudhram: We haven't fully moved away from bluebird promisify (being used here https://github.com/mozilla/fxa/blob/b22d8cdeb47cf4dcd52aa90fc8e21f4eedd24725/packages/fxa-auth-server/lib/cache.js#L10), note sure if we wanted to or not, but you could try using that promisify function.

bbangert: @rfk suggested that as well, it didn't seem to work nicely on the zendesk library though.

packages/fxa-auth-server/lib/routes/support.js
bbangert: I copied this method from the `subscriptions.js`, ideally it would be refactored out somewhere that both of these could use it. The current code architecture isn't very amenable to this though.

philbooth: Could it be pushed into the auth strategy? Or if not, just imported from some common helper module?

packages/fxa-auth-server/lib/zendesk-client.js
philbooth: I'm mildly uneasy about this.

The `promisify` implementation here returns native promises but elsewhere we use bluebird promises. Although they share a common API and can be mixed safely, bluebird promises also support other methods that will be missing if someone accidentally tries to invoke them on one of these. Given that we use some of those methods elsewhere in the code, it seems like a lurking footgun if we return native promises from some places and bluebird promises from others. So at the very least, I think you should return bluebird promises from these wrappers.

But I also think it's preferable to use bluebird's `promisify` over re-implementing our own, if we can. Looking at the code, I presume it was that step of aggregating `req` and `result` to an object, because promises can only resolve to a single value. Bluebird has a `multiArgs` boolean arg for that. It aggregates them to an array instead of an object, but they can still be unpacked to named variables on the way out with array destructuring, e.g.:

```js
const Promise = require('bluebird');

// ...

zendeskClient.createRequest = Promise.promisify(zendeskClient.requests.create, {
  multiArgs: true,
  context: zendeskClient.requests,
});

// ...

const [ , result ] = await zendeskClient.createRequest({ /* ... */ });
```

And if we really wanted to eliminate repetition, we could derive the calls to `promisify` from static data, e.g.:

```js
const wrappedMethods = [
  { name: 'createRequest', context: 'requests', target: 'create' },
  { name: 'showUser', context: 'users', target: 'show' },
  { name: 'updateUser', context: 'users', target: 'update' },
];
wrappedMethods.forEach(method => {
  const { name, context, target } = method;
  zendeskClient[name] = Promise.promisify(zendeskClient[context][target], {
    multiArgs: true,
    context: zendeskClient[context],
  });
});
```

Last thing, I think we prefer `camelCase` names in most JS code. Mixing `snake_case` in some places makes things less uniform across the board, which means APIs become less predictable and intuitive when you're working with them. Neither one is better than the other, I just think there's value in consistency (even though we're not 100% consistent in all places, I realise).

bbangert: I tried both uses of bluebird promisify that you mentioned. Neither works, and results in errors about undefined like this:
```
TypeError: Cannot read property 'requester_id' of undefined
```

philbooth: > ```
> TypeError: Cannot read property 'requester_id' of undefined
> ```

Beats me, let's do it your way! :smile:

PR #1112, opened 2019-05-15
feat(subscriptions): support deferred cancellation of subscriptions

philbooth:
Related to #981.

Changes the behaviour of `DELETE /oauth/subscriptions/active/:subscriptionId` so that it marks a subscription as cancelled rather than deletes it from the db. If the subscription can't be found or has already been cancelled, the db method will fail.

Doesn't include any remote tests because #1110.

I think this ticks off the first two checkboxes in the description for #981, but it doesn't make any attempt to tackle the second two.

@lmorchard r?

PR #1655, opened 2019-07-02
feat(subscriptions): implement reactivation of cancelled subscriptions

### LINKED ISSUE #1273: Implement reactivation of cancelled subscription before end of current term
Follow up to Issue #981 - we need a button to reactivate a cancelled subscription that has not yet ended:

Capture

Blocked by this issue in subhub: https://github.com/mozilla/subhub/issues/68

philbooth:
Fixes #1273.

This is the backend for subscription reactivation. ~~Opening as a draft, I haven't looked at any of the front-end stuff yet so I don't know where the new button is supposed to live.~~ It seems like the front-end was already in place, so I just hooked into it by cargo-culting what I saw elsewhere in `fxa-payments-server`.

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
+++ b/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
@@ -4180,7 +4180,7 @@ module.exports = function(config, DB) {
     });
 
     describe('account subscriptions', () => {
-      const subscriptionIds = [...Array(25)].map(() => 'sub-' + Math.random());
+      const subscriptionIds = [...Array(26)].map(() => 'sub-' + Math.random());
       let account;
 
       beforeEach(async () => {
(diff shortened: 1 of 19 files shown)


philbooth:
@mozilla/fxa-devs I think this is ready for review now, r?

packages/fxa-auth-db-mysql/docs/API.md
```diff
@@ -2543,20 +2544,18 @@ Content-Length: 2
 ### Example
 
 ```
-
```
philbooth: Why am I messing with the whitespace here? The extra blank lines inside the pre-formatted blocks caused me to misread which block of text they applied to further down and add extra backticks unnecessarily. I think it's clearer this way.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -4470,6 +4470,100 @@ module.exports = function(config, DB) {
...
+          account.uid,
+          subscriptionIds[25],
+        );
+
+        let failed = false;
+        try {
+          await db.reactivateAccountSubscription(
+            account.uid,
+            subscriptionIds[25],
+          );
```
bbangert: I've seen other tests where we `assert.fail()` in a try if we expect it to error. This is the first time I've seen this pattern of a failed bool that is asserted later. Is there a preference in the code-base that we should aim for?

packages/fxa-auth-db-mysql/docs/API.md
bbangert: Will prettier undo this later, or re-add it?

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
philbooth: I disagree with the use of `assert.fail()` inside `try` blocks because the assert error gets caught by the `catch`, so it relies on secondary assertions there to actually fail the test. That makes it harder to diagnose the true causes of failure and it's even conceivable that a failure might accidentally be swallowed if assertions inside the `catch` don't differentiate it from their expected error type.

@lmorchard previously made the same point to me in https://github.com/mozilla/fxa/pull/1112#discussion_r284338002 and I changed my approach in that PR. But my opinion has hardened since then and I now think using `assert.fail()` like that is objectively wrong.

packages/fxa-auth-db-mysql/docs/API.md
philbooth: It doesn't seem to, for me.

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -528,4 +528,124 @@ describe('subscriptions', () => {
...
+          ...requestOptions,
+          method: 'POST',
+          payload: { subscriptionId: SUBSCRIPTION_ID_1 },
+        }
+      );
+
+      assert.equal(customs.check.callCount, 1);
+
+      assert.equal(db.reactivateAccountSubscription.callCount, 1);
+      let args = db.reactivateAccountSubscription.args[0];
```
vbudhram: Should we assert `subhub.reactivateSubscription` was called?

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -4470,6 +4470,100 @@ module.exports = function(config, DB) {
...
+            Date.now()
+          );
+        } catch (err) {
+          failed = true;
+          assert.equal(err.errno, 116);
+        }
+        assert(failed);
+      });
+
+      it('should fail to reactivate a non-cancelled subscription', async () => {
```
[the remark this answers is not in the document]
philbooth: Yep. I used "non-cancelled" to differentiate it from "reactivated" below, I think the two warrant separate test cases even though the underlying state in the db is the same (that's an implementation detail).

packages/fxa-auth-server/test/local/routes/subscriptions.js
philbooth: Oh, we should yeah. I merged this before I saw your comment, sorry.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
bbangert: Sounds good to me. Can we codify this as a best practice in a code style doc? I'm not sure an ADR makes sense, but ideally things like this should be in our own code style document.

packages/fxa-auth-server/test/local/routes/subscriptions.js
vbudhram: No worries, forgot to hit the submit review button 🤦🏽‍♂️

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
bbangert: Also, is there a reason we can't use `https://www.chaijs.com/plugins/chai-as-promised/` with `assert.isRejected` or `isFulfilled`?

philbooth: Oh, probably not, other than lack of awareness. Let's add it next opportunity.

PR #1656, opened 2019-07-02
subhub integration tweaks for auth and payments servers (revised)

ianb:
This is a fix for the tests in #1592, plus a small pm2 fix. I had to recreate this since the other PR is in Les's repo.

Description on the previous PR:

- add example local subhub server config to auth-server

- stub out subscription reactivate button (pending Issue #1273)

- handle & display errors in calling Stripe createToken API

- change OAuth client for payment pages to publicClient: true

- properly handle 402 errors on subscribe & payment update

- include more payment error info in payment form errors

- add cancel_at_period_end in customer data for cancelled subscriptions

- rename nickname field to plan_name in customer subscriptions data

- tweaks to subhub stubAPI to better reflect subscription cancellation

- default result for fetch states in Redux set to null

- misc tweaks to address lint warnings


packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -180,22 +180,20 @@ module.exports = function(log, config) {
           orig_system: ORIG_SYSTEM,
         });
       } catch (err) {
-        if (err.statusCode === 400) {
+        if (err.statusCode === 400 || err.statusCode === 402 || err.statusCode === 404) {
```
philbooth: Restating https://github.com/mozilla/fxa/pull/1592#discussion_r296603832, isn't this `if` test redundant now with the tests below? The only side effect would be to call `log.error` in every case, which seems like an improvement to me. Can we delete this outer one?

packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -236,13 +234,13 @@ module.exports = function(log, config) {
       try {
         return await api.updateCustomer(uid, { pmt_token });
       } catch (err) {
-        if (err.statusCode === 400 || err.statusCode === 404) {
+        if (err.statusCode === 400 || err.statusCode === 402 || err.statusCode === 404) {
```
philbooth: Same, same.

packages/fxa-auth-server/lib/subhub/client.js
ianb: I don't entirely understand what's going on here, but the error message seems to get logged twice nearly identically when I make that change. So... it seems logically sensible to simplify, but I guess I'll leave it be.

packages/fxa-payments-server/src/routes/Product/index.stories.tsx
```diff
@@ -28,6 +26,7 @@ function init() {
           {
             current_period_end: (Date.now() + 86400) / 1000,
             current_period_start: (Date.now() - 86400) / 1000,
+            cancel_at_period_end: false,
             ended_at: null,
```
bbangert: This needs to be `end_at`.

packages/fxa-payments-server/src/store/types.tsx
```diff
@@ -42,6 +42,7 @@ export interface Subscription {
 }
 
 export interface CustomerSubscription {
+  cancel_at_period_end: boolean;
   current_period_end: number;
   current_period_start: number;
   ended_at: string | null,
```
bbangert: Also `end_at`.

PR #1659, opened 2019-07-02
fix: update subhub validation call to match API

### LINKED ISSUE #1658: [fxa-auth-server] Ensure fxa-auth-server schemas for subhub calls are current
There is some drift on the validation of SubHub calls in fxa-auth-server, these should be resolved and current.

bbangert:
Closes #1658

### DIFF
--- a/packages/fxa-auth-server/lib/routes/validators.js
+++ b/packages/fxa-auth-server/lib/routes/validators.js
@@ -285,6 +285,7 @@ module.exports.activeSubscriptionValidator = isA.object({
 
 // This is subhub's perspective on an active subscription
 module.exports.subscriptionsSubscriptionValidator = isA.object({
+  cancel_at_period_end: isA.bool().required(),
   current_period_end: isA
     .date()
     .timestamp('unix')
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -293,11 +294,10 @@ module.exports.subscriptionsSubscriptionValidator = isA.object({
     .date()
     .timestamp('unix')
     .required(),
-  ended_at: isA.alternatives(
-    isA.date().timestamp('unix'),
-    isA.any().allow(null)
-  ),
-  nickname: isA.string().required(),
+  end_at: isA.alternatives(isA.date().timestamp('unix'), isA.any().allow(null)),
```
rfk: Do we really want to allow `any` here, which IIUC just matches anything?

```
> end_at = isA.alternatives(isA.date().timestamp('unix'), isA.any().allow(null))
> end_at.validate({hello: "world"}).value 
{ hello: 'world' }
>
```

I *think* `isA.any().valid(null)` is what you want here instead; it's not any more intuitive but it seems to work in testing, allowing *only* `null` as the valid value.

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -293,11 +294,10 @@ module.exports.subscriptionsSubscriptionValidator = isA.object({
...
+  failure_code: isA.string().optional(),
+  failure_message: isA.string().optional(),
+  plan_name: isA.string().required(),
```
rfk: Are we able to restrict the characters allowed in these strings, even if only to `routes.validators.DISPLAY_SAFE_UNICODE` or similar? I've seen too many bugs allowed through by the generic `.string()` validator...

bbangert: Someday we can, yes.

PR #1661, opened 2019-07-03
feat(routes): securityEvents GET and DELETE added with uid

farhan787:
Working on `securityEvents` routes for the Firefox Security Dashboard with @vbudhram

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -218,6 +218,18 @@ function createServer(db) {
 
   api.get('/securityEvents/:id/ip/:ipAddr', withParams(db.securityEvents));
   api.post('/securityEvents', withBodyAndQuery(db.createSecurityEvent));
+  api.get(
+    '/securityEvents/:id',
+    op(req => {
+      return db.securityEventsByUid(req.params.id);
+    })
+  );
+  api.del(
+    '/securityEvents/:id',
+    op(req => {
+      return db.deleteSecurityEventsByUid(req.params.id);
+    })
+  );
 
   api.get('/emailBounces/:id', withIdAndBody(db.fetchEmailBounces));
   api.post('/emailBounces', withBodyAndQuery(db.createEmailBounce));
--- a/packages/fxa-auth-server/lib/db.js
+++ b/packages/fxa-auth-server/lib/db.js
@@ -1156,6 +1156,30 @@ module.exports = (config, log, Token, UnblockCode = null) => {
     return this.pool.get(SAFE_URLS.securityEvents, { ipAddr, uid });
   };
 
+  SAFE_URLS.securityEventsByUid = new SafeUrl(
+    '/securityEvents/:uid',
+    'db.securityEventsByUid'
+  );
+  DB.prototype.securityEventsByUid = function(params) {
+    log.trace('DB.securityEventsByUid', {
+      params: params,
+    });
+    const { uid } = params;
+    return this.pool.get(SAFE_URLS.securityEventsByUid, { uid });
+  };
+
+  SAFE_URLS.deleteSecurityEvents = new SafeUrl(
+    '/securityEvents/:uid',
+    'db.deleteSecurityEventsByUid'
+  );
+  DB.prototype.deleteSecurityEvents = function(params) {
+    log.trace('DB.deleteSecurityEvents', {
+      params: params,
+    });
+    const { uid } = params;
+    return this.pool.del(SAFE_URLS.deleteSecurityEvents, { uid });
+  };
+
   SAFE_URLS.createUnblockCode = new SafeUrl(
     '/account/:uid/unblock/:unblock',
     'db.createUnblockCode'
--- a/packages/fxa-auth-server/lib/routes/security-events.js
+++ b/packages/fxa-auth-server/lib/routes/security-events.js
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+module.exports = (log, db, config) => {
+  return [
+    {
+      method: 'GET',
+      path: '/securityEvents',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function(request) {
+        log.begin('SecurityEventsGet', request);
+        const { uid } = request.auth.credentials;
+
+        const events = await db.securityEventsByUid({ uid });
+        return events;
+      },
+    },
+    {
+      method: 'DELETE',
+      path: '/securityEvents',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function(request) {
+        log.begin('SecurityEventsDelete', request);
+        const { uid } = request.auth.credentials;
+
+        await db.deleteSecurityEvents({ uid });
+        return {};
+      },
+    },
+  ];
+};
--- a/packages/fxa-auth-server/test/client/api.js
+++ b/packages/fxa-auth-server/test/client/api.js
@@ -772,6 +772,18 @@ module.exports = config => {
     });
   };
 
+  ClientApi.prototype.securityEvents = function(sessionTokenHex) {
+    return tokens.SessionToken.fromHex(sessionTokenHex).then(token => {
+      return this.doRequest('GET', `${this.baseURL}/securityEvents`, token);
+    });
+  };
+
+  ClientApi.prototype.deleteSecurityEvents = function(sessionTokenHex) {
+    return tokens.SessionToken.fromHex(sessionTokenHex).then(token => {
+      return this.doRequest('DELETE', `${this.baseURL}/securityEvents`, token);
+    });
+  };
+
   ClientApi.prototype.accountProfile = function(sessionTokenHex, headers) {
     const o = sessionTokenHex
       ? tokens.SessionToken.fromHex(sessionTokenHex)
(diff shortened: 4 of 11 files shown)


packages/fxa-auth-server/test/client/index.js
```diff
@@ -33,7 +34,15 @@ module.exports = config => {
 
   Client.Api = ClientApi;
 
-  Client.prototype.setupCredentials = function(email, password) {
+  Client.prototype.setupCredentials = function(email, password, eventType) {
```
vbudhram: You shouldn't have to add this to the `setupCredentials` function.

packages/fxa-auth-server/lib/routes/security-events.js
```diff
@@ -0,0 +1,31 @@
...
+'use strict';
+
+const error = require('../error');
+const isA = require('joi');
+
+module.exports = (log, db, config) => {
+  return [
+    {
+      method: 'GET',
+      path: '/securityEvents/:id',
```
vbudhram: You will need to put this behind a `sessionToken` authentication strategy, https://github.com/mozilla/fxa/blob/b22d8cdeb47cf4dcd52aa90fc8e21f4eedd24725/packages/fxa-auth-server/lib/routes/emails.js#L36. When you do that, you will have access to the the `uid` from the [credentials object](https://github.com/mozilla/fxa/blob/b22d8cdeb47cf4dcd52aa90fc8e21f4eedd24725/packages/fxa-auth-server/lib/routes/emails.js#L60).

You will also be able to remove the `/:id` part from the path.

packages/fxa-auth-server/test/client/api.js
```diff
@@ -158,6 +158,7 @@ module.exports = config => {
         redirectTo: options.redirectTo || undefined,
         resume: options.resume || undefined,
         device: options.device || undefined,
+        securityEvents: options.securityEvents || [],
```
vbudhram: This can be removed too.

packages/fxa-auth-server/test/remote/security_events.js
```diff
@@ -0,0 +1,53 @@
...
+    config.signinConfirmation.skipForNewAccounts.enabled = false;
+    return TestServer.start(config).then(s => {
+      server = s;
+    });
+  });
+
+  it('returns securityEvents on creating and login into an acount', () => {
+    const email = server.uniqueEmail();
+    const password = 'abcdef';
+    return Client.create(config.publicUrl, email, password)
```
vbudhram: You will need to expose some helper functions on the client api.js that uses a sessionToken

https://github.com/mozilla/fxa/blob/e3d1bdd4c31a487a9ef22a03a604eabf4d930819/packages/fxa-auth-server/test/client/index.js#L503

vbudhram: Also see https://github.com/mozilla/fxa/blob/e3d1bdd4c31a487a9ef22a03a604eabf4d930819/packages/fxa-auth-server/test/client/api.js#L770, for where api calls the auth-server.

packages/fxa-auth-db-mysql/db-server/index.js
```diff
@@ -218,6 +218,18 @@ function createServer(db) {
 
   api.get('/securityEvents/:id/ip/:ipAddr', withParams(db.securityEvents));
   api.post('/securityEvents', withBodyAndQuery(db.createSecurityEvent));
+  api.get(
+    '/securityEvents/:id',
+    op(function(req) {
```
vbudhram: I think I missed this in other PR, but lets use the new `=>` syntax where possible.

packages/fxa-auth-server/lib/routes/security-events.js
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+module.exports = (log, db, config) => {
+  return [
+    {
+      method: 'GET',
```
vbudhram: Lets add the corresponding `DEL` route for security events plus tests etc.

packages/fxa-auth-server/test/remote/db_tests.js
```diff
@@ -1288,6 +1288,40 @@ describe('remote db', function() {
...
+        name: 'account.create',
+        uid: account.uid,
+      })
+      .then(() => {
+        return db.deleteSecurityEventsByUid({
+          uid: account.uid,
+        });
+      })
+      .then(events => {
+        assert.deepEqual(events, {});
```
vbudhram: After this assert, try to get the security events, it should return 0.

packages/fxa-auth-server/test/remote/security_events.js
```diff
@@ -0,0 +1,153 @@
...
+        assert.isBelow(events[0].createdAt, new Date().getTime());
+        assert.equal(events[0].verified, false);
+
+        assert.equal(events[1].name, 'account.create');
+        assert.isBelow(events[1].createdAt, new Date().getTime());
+        assert.equal(events[1].verified, true);
+      });
+  });
+
+  it('returns security events after password change, with unverified session', () => {
```
vbudhram: This test can be removed. I thought we were storing password change security events, but it doesn't appear so.

packages/fxa-auth-server/test/remote/security_events.js
```diff
@@ -0,0 +1,153 @@
...
+        assert.isBelow(events[0].createdAt, new Date().getTime());
+        assert.equal(events[0].verified, false);
+
+        assert.equal(events[1].name, 'account.create');
+        assert.isBelow(events[1].createdAt, new Date().getTime());
+        assert.equal(events[1].verified, true);
+      });
+  });
+
+  it('returns security events after password change, with verified session', () => {
```
vbudhram: We can remove this too. Instead try testing the password reset flow, ref https://github.com/mozilla/fxa/blob/b22d8cdeb47cf4dcd52aa90fc8e21f4eedd24725/packages/fxa-auth-server/test/remote/account_reset_tests.js#L157

packages/fxa-auth-server/lib/routes/security-events.js
farhan787: Sure

packages/fxa-auth-server/test/remote/db_tests.js
farhan787: Nice catch @vbudhram

packages/fxa-auth-server/lib/routes/security-events.js
```diff
@@ -0,0 +1,42 @@
...
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function(request) {
+        log.begin('SecurityEvents', request);
+        const { uid } = request.auth.credentials;
+
+        const emptyEvents = await db.deleteSecurityEventsByUid({ uid });
```
vbudhram: I don't think the intermediate variable buys us anything here, you can do something like

```
await db.deleteSecurityEventsByUid({ uid });
return {};
```

packages/fxa-auth-server/lib/routes/security-events.js
```diff
@@ -0,0 +1,42 @@
...
+    {
+      method: 'GET',
+      path: '/securityEvents',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function(request) {
+        log.begin('SecurityEvents', request);
```
vbudhram: To help differentiate the routes in the logs, lets use `SecurityEventsGet` and `SecurityEventsDelete`. We did something similar for recovery keys.

packages/fxa-auth-server/lib/db.js
```diff
@@ -1156,6 +1156,30 @@ module.exports = (config, log, Token, UnblockCode = null) => {
...
+  );
+  DB.prototype.securityEventsByUid = function(params) {
+    log.trace('DB.securityEventsByUid', {
+      params: params,
+    });
+    const { uid } = params;
+    return this.pool.get(SAFE_URLS.securityEventsByUid, { uid });
+  };
+
+  SAFE_URLS.deleteSecurityEventsByUid = new SafeUrl(
```
vbudhram: On reflection, the `deleteSecurityEventsByUid` in can be improved by simply calling it `deleteSecurityEvents`. The functions in here assume you are passing either `uid` or `id` when they are called. 

However, we don't have to the change the `securityEventsByUid` since we already have `securityEvents` function.

packages/fxa-auth-server/test/client/index.js
```diff
@@ -498,6 +498,13 @@ module.exports = config => {
     });
   };
 
+  Client.prototype.securityEvents = function() {
+    const o = this.sessionToken ? P.resolve(null) : this.login();
+    return o.then(() => {
+      return this.api.securityEvents(this.sessionToken);
+    });
+  };
+
```
vbudhram: You should create the `deleteSecurityEvents` client method here.

packages/fxa-auth-server/test/remote/security_events.js
```diff
@@ -0,0 +1,112 @@
...
+        assert.isBelow(events[0].createdAt, new Date().getTime());
+        assert.equal(events[0].verified, false);
+
+        assert.equal(events[1].name, 'account.create');
+        assert.isBelow(events[1].createdAt, new Date().getTime());
+        assert.equal(events[1].verified, true);
+      });
+  });
+
+  it('returns security events after account reset w/o keys, with sessionToken', () => {
```
vbudhram: You will need to add some remote tests for delete security events in here.

farhan787:
> @farhan787 Thanks for updates, there are a few more tests that can be added and I think we should update the function `deleteSecurityEventsByUid` -> `deleteSecurityEvents` where applicable.

Ok done !

farhan787:
@vbudhram made the changes.

farhan787:
> @farhan787 This LGTM 👍, thanks you! Can you please squash commits when ready, then we can merge.

Done :-)

PR #1663, opened 2019-07-03
fix(tests): add missing assertion for subhub.reactivateSubscription

philbooth:
Spillover from https://github.com/mozilla/fxa/pull/1655#discussion_r300032800.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/subscriptions.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions.js
@@ -531,19 +531,22 @@ describe('subscriptions', () => {
 
   describe('POST /oauth/subscriptions/reactivate', () => {
     it('should reactivate cancelled subscriptions', async () => {
-      const res = await runTest(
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -531,19 +531,22 @@ describe('subscriptions', () => {
...
-          ...requestOptions,
-          method: 'POST',
-          payload: { subscriptionId: SUBSCRIPTION_ID_1 },
-        }
-      );
+      const res = await runTest('/oauth/subscriptions/reactivate', {
+        ...requestOptions,
+        method: 'POST',
+        payload: { subscriptionId: SUBSCRIPTION_ID_1 },
+      });
```
philbooth: Turns out the post-commit hook wasn't working when I did #1655, I only just fixed it now hence the fixing up here.

PR #1664, opened 2019-07-03
fix(subs): Add display_name field to subhub create subscription API

### LINKED ISSUE #1597: Include display_name in subhub create subscription payload
It has been determined that subsystems need a user's name to process data creation correctly.  This issue is to request that the profile display_name is included in the body of the  POST /customer/{uid}/subscriptions request.  There is currently an pending PR to accept this value in Subhub .

6a68:
Fixes #1597.

-----

original WIP description:

* Not yet able to get tests running locally; pushing to CI to see if the
tests run there.

* Looks like the payments-server already gets the full profile from the
profile-server, so it's just a matter of passing the display name field
from payments-server to auth-server, and then from auth-server to
subhub.

### DIFF
--- a/packages/fxa-auth-server/lib/subhub/client.js
+++ b/packages/fxa-auth-server/lib/subhub/client.js
@@ -122,6 +122,7 @@ module.exports = function(log, config) {
           plan_id: isA.string().required(),
           email: isA.string().required(),
           orig_system: isA.string().required(),
+          display_name: isA.string().required(),
         },
         response: isA.alternatives(
           validators.subscriptionsSubscriptionListValidator,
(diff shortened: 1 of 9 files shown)


packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -207,6 +210,7 @@ module.exports = function(log, config) {
           if (err.statusCode === 400 || err.statusCode === 402) {
             throw error.rejectedSubscriptionPaymentToken(err.message, err);
           }
+          // TODO: update with whatever the error response is for missing display_name.
```
philbooth: I think you'll need to tweak the condition above as this is one possible meaning of the `400` response.

Ref:
https://github.com/mozilla/subhub/blob/e530fcb67339c8b41035ec2a67b06d23c4ef97b2/subhub/subhub_api.yaml#L133-L136

philbooth: (you can see elsewhere we resort to interrogation of `err.message`, I guess this will need more of the same sadly)

packages/fxa-auth-server/test/remote/subscription_tests.js
```diff
@@ -154,14 +155,16 @@ describe('remote subscriptions:', function() {
       assert.deepEqual(result, []);
     });
 
+    // TODO: not sure if any changes need to be made here
```
philbooth: Yeah, looks like `display_name` is not returned, so there's nothing to assert here:

https://github.com/mozilla/subhub/blob/e530fcb67339c8b41035ec2a67b06d23c4ef97b2/subhub/subhub_api.yaml#L438-L479

packages/fxa-auth-server/lib/subhub/client.js
philbooth: Actually ignore me, given that it's `required` in the API here, I guess you don't need that condition after all?

6a68: Ah, cool! Thanks, I'll remove the TODO

6a68:
@philbooth OK, got the tests green! Mind taking a last look?

Issue #1637, opened 2019-06-27
[Payments] Regression in Subscription Manage button


ianb:
I believe c4b1af14bf91b09960f7fe0f1c3e595f12cfcb0c (by @shane-tomlinson) caused a regression in the Settings subscription Manage button:

<img width="624" alt="image" src="https://user-images.githubusercontent.com/44390/60301711-b4d96180-98f7-11e9-8666-fc9a7e3b00c1.png">

After the commit the button no longer works, with the error message `Error: Invalid parameter: %(param)s`. After some investigation I believe this is coming from `createOAuthToken()` and is due to an error about a missing parameter `client_secret`.

lmorchard:
This *should* be fixed by PR #1592. The `client_secret` thing is a bit of a red herring - what we really need is for the subscription management client [to become a public client](https://github.com/mozilla/fxa/pull/1592/files#diff-bca803ff6b323df3b3407f94d63c23e7R178). Those don't use client secrets at all.

(Still on PTO, but saw this in my inbox and knew the answer)

Issue #926, opened 2019-04-24
Create stable URL that will redirect to the payments page


ianb:
To actually use the payments page you have to come through accounts.firefox.com and get redirected with a token. We want to link directly to payments pages, which will require a page on accounts.firefox.com which redirects immediately (with a token) to the payments URL. This page should probably display a simple loading spinner while it attempts to redirect (redirection has to happen in the browser, as the account information is in the browser and not available to the server).

PRD requirements included in this:

* A 1.2 | FXA should allow users to sign in or sign up via a Firefox Account
* A 1.3 | If users are signing in, FXA should direct them to a payments UI

shane-tomlinson:
@ianb is there a link to a diagram of how this should work? I'm having a hard time imagining the flow.

lmorchard:
> is there a link to a diagram

There has been one going around, though I don't think I have good links to it. [A version was shared in Slack](https://mozilla.slack.com/archives/CG0AJ6E77/p1555362941064100) and appeared in [a private Google doc from a security review](https://docs.google.com/document/d/17PUUQyCseT39seTkK7CPlXVt07p9MDZoypJnSV-TVZs/edit?ts=5cb4a53e#heading=h.8hq3qvrfp4h4) with the flow sketched out. @bbangert might have a better link to an image.

bbangert:
Yes, here's the mermaid code for it. I wish there was an easier way to share/codify these, I'm still looking for a better solution.
```
sequenceDiagram
participant UA as User-Agent<br>(Browser)
participant Svc as Service Offering
participant FxAC as FxA Content Svr<br>(accounts.firefox.com)
participant FxAP as FxA Content Svr<br>(payment.firefox.com)
participant FxA as FxA Auth<br>(api.accounts.f.c)
participant SB as SubHub
participant Stripe as Stripe

UA->>Svc: Click "buy" button
Svc->>UA: 302 accounts.firefox.com
UA->>FxAC: Enter sign-up logic
Note over FxAC: Varies depending on<br>user status, existing<br>account, etc.
FxAC->>FxAC: make short lived oauth token
Note over UA: User is sent to<br>dedicated site that<br>serves payment form,<br>isolated from FxA
FxAC->>UA: 302 payment.f.c {plan_id, oauth_token}
UA->>FxAP: Get stripe payment form {oauth_token}
Note over FxAP: is it useful to validate<br>the oauth token here?
FxAP->>SB: GET list_all_plans
[...]

PR #1665, opened 2019-07-03
[WIP] Subscriptions redirect (updated for master)

ianb:
This is an attempt to recreate #1310 against master. From the previous PR:

- add `/subscriptions/*` path to server and client-side router

- SubscriptionsRedirectView that issues an access token and redirects to
  a path on fxa-payments-server

Issue #926

**Note:** Right now this does not appear to actually work locally, maybe it reintroduces #1637, or maybe it's another issue.


packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -0,0 +1,29 @@
...
+
+import FormView from './form';
+import Template from 'templates/subscriptions_redirect.mustache';
+import PaymentServer from '../lib/payment-server';
+
+class SubscriptionsProductRedirectView extends FormView {
+  mustAuth = true;
+  template = Template;
+
+  initialize (options) {
```
bbangert: I thought prettier would remove the space here, or does that not run on content-server yet?

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -2,25 +2,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-'use strict';
+"use strict";
```
bbangert: I thought our prettier spec would have these all stay single quote?

ianb: I'm a little confused what happened with the style changes too. I worry all the rebasing and conflict merging might have left things a little screwy. I should probably just re-prettify.

ianb: I'm not clear what is being applied, but running prettier makes an even bigger diff than this, so best to hold on that.

ianb:
I've decided to manually recreate #1310 on top of master.

PR #1667, opened 2019-07-04
fix(jwks): Set explicit cache-control header on jwks_uri

rfk:
We intend for RPs to fetch and cache the contents of this URL, so we should really give them some instructions on how to do so. This brings https://oauth.accounts.firefox.com/v1/jwks in line with the existing cache-control settings on https://api.accounts.firefox.com/.well-known/browserid. @shane-tomlinson r?


rfk:
We could also consider making `max-age` here substantially larger than 10 seconds, which seems very low to me, but it's what we use on the auth-server.

shane-tomlinson:
Opened #1686, merged because @rfk is out for the week.

PR #1668, opened 2019-07-04
Add scripts and docs for rotating the signing key

rfk:
In recent conversation, @shane-tomlinson mentioned that we didn't have very good docs written down for how to rotate the `id_token` (and soon `access_token`) signing key. I started to write one and then realized that our current facilities for actually doing the rotation were inadequate, and could allow a brief window where RPs with a cached copy of our JWK Set would get spurious validation failures.

So, here we have:

* A doc describing the intended mechanics of key rotation, and why we need two "staging" periods in the process - one for introducing the new key, and one for phasing out the old key.
* An extra config value to be used for staging the new signing key *before* we actually start using it to sign things.
* A set of scripts for automating the stages of this process.

I don't know if we'd use the scripts verbatim in production, but would be happy to modify them as needed so they're useful, if that's how we want to proceed here.

Rendered explainer doc [here](https://github.com/mozilla/fxa/blob/scripted-key-rotation/packages/fxa-auth-server/fxa-oauth-server/docs/signing-key-management.md).

(I'm also not thrilled that we're doing a bunch of hand-massaging of the components of the keys like this, and I wonder if the node JWX ecosystem has evolved sufficiently that we could use an existing library. But that's a bigger topic, and this PR doesn't cause us to do anything that we weren't doing already).

@shane-tomlinson @dannycoates @bbangert @jbuck @jrgm thoughts?

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/docs/signing-key-management.md
+++ b/packages/fxa-auth-server/fxa-oauth-server/docs/signing-key-management.md
@@ -1,0 +1,58 @@
+# Management of JWT Signing Keys
+
+This server uses signed [JWTs](https://jwt.io/) to represent various kinds of security token,
+including:
+
+- `id_token`s as defined by [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
+- `access_token`s as defined by [JWT Profile for OAuth 2.0 Access
+  Tokens](https://datatracker.ietf.org/doc/draft-bertocci-oauth-access-token-jwt/)
+- Security Event Tokens as defined by [RFC8417](https://tools.ietf.org/html/rfc8417)
+
+RPs are expected to verify the signatures on these JWTs by fetching our public keys via the
+[OpenID Connect Discovery Protocol](https://openid.net/specs/openid-connect-discovery-1_0.html),
+which involves:
+
+1. Fetching our main OpenID metadata document at https://accounts.firefox.com/.well-known/openid-configuration
+2. Extracting the contained `"jwks_uri"` entry
+3. Fetching a [JWK Set](https://tools.ietf.org/html/rfc7517) document from that URI
+4. Respecting standard HTTP cache-control headers on those resources
+
+We thus have a coordination problem between this server and its RPs when it comes to changing our
+JWT signing keys. Let `T_cache` be the max age in the cache-control headers of our metadata documents,
+and `T_tokens` be the maximum lifetime of any token issued by this service. Then:
+
+- We must advertise a new signing key in our JWK Set for at least `T_cache` seconds before we start
+  using it to sign tokens. If we don't, RPs with a cached copy of our JWK Set may not know to trust
+  the new key, resulting in spurious verification failures and/or fragile logic to refresh the cached
+  copy on-demand when verification fails.
+- We must continue to advertise an old signing key in our JWK Set for at least `T_tokens` seconds
+  after we stop using it to sign tokens. If we don't, RPs who refresh their cache of our JWK Set may
+  treat tokens signed with that key as invalid before they expire, resulting in spurious verification
+  failures.
+
+To help manage this, the server config has slots for three different keys, one required and the others
+optional:
+
+- `openid.key`: The private key that is actively being used to sign tokens. This key is required.
+- `openid.newKey`: A new private key that we intend to start using to sign tokens in the future.
+  This key is optional, but can be specified in order to start advertising it before use.
+- `openid.oldKey`: The _public_ component of a key we previously used to sign tokens. This key is
+  optional, but can be specified in order to continue advertising it while tokens bearing its signature
+  may be active.
+
+The procedure for rotating the active signing key is:
+
+- Create the new signing key and configure it as `openid.newKey`, while leaving the existing `openid.key` intact.
+  - This can be performed using `../scripts/prepare-new-signing-key.js`.
+- Deploy to all server instances, then wait for at least `T_cache` seconds plus some margin for clock error.
+- Take the public component of `openid.key` and configure it as `openid.oldKey`, then move `openid.newKey`
+  to `openid.key`.
+  - This can be performed using `../scripts/activate-new-signing-key.js`.
+- Deploy to all server instances, then wait for at least `T_tokens` seconds plus some margin for clock error.
+- Remove `openid.oldKey` from the configuration.
+  - This can be performed using `../scripts/retire-old-signing-key.js`.
+- Deploy to all server instances, then celebrate a job well done.
+
+This scheme keeps things fairly simple, but it also means that each key rotation event must be at least
+`T_cache + T_tokens` seconds apart. If this proves to be a problem in practice, we could reduce the limit
+to `T_cache` by keeping a _list_ of old signing keys rather than a single key.
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
@@ -240,27 +240,50 @@ const conf = convict({
     },
   },
   openid: {
+    // See ../docs/signing-key-management.md to read more about what these different keys are for.
     keyFile: {
-      doc: 'Path to Private key JWK to sign id_tokens',
+      doc: 'Path to private key JWK to sign various kinds of JWT tokens',
       format: String,
       default: '',
       env: 'FXA_OPENID_KEYFILE',
     },
+    newKeyFile: {
+      doc:
+        'Path to private key JWK that will be used to sign JWTs in the future',
+      format: String,
+      default: '',
+      env: 'FXA_OPENID_NEWKEYFILE',
+    },
     oldKeyFile: {
-      doc: 'Path to previous key that was used to sign id_tokens',
+      doc: 'Path to public key JWK that was used to sign JWTs in the past',
       format: String,
       default: '',
       env: 'FXA_OPENID_OLDKEYFILE',
     },
+    // These config values can be used to set JWK JSON objects directly as keys, rather than
+    // loading them from a file. They will be overwritten by the contents of the corresponding
+    // key file if configured.
     key: {
-      doc: 'Private JWK to sign id_tokens',
+      doc: 'Private key JWK to sign various kinds of JWT tokens',
       default: {},
       env: 'FXA_OPENID_KEY',
     },
+    newKey: {
+      doc: 'Private key JWK that will be used to sign JWTs in the future',
+      default: {},
+      env: 'FXA_OPENID_NEWKEY',
+    },
     oldKey: {
-      doc: 'The previous public key that was used to sign id_tokens',
+      doc: 'Public key JWK that was used to sign JWTs in the past',
       default: {},
       env: 'FXA_OPENID_OLDKEY',
+    },
+    unsafelyAllowMissingActiveKey: {
+      doc:
+        'Do not error out if there is no active key; should only be used when initializing keys',
+      format: Boolean,
+      default: false,
+      env: 'FXA_OPENID_UNSAFELY_ALLOW_MISSING_ACTIVE_KEY',
     },
     issuer: {
       // this should match `issuer` in the 'OpenID Provider Metadata' document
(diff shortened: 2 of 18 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -0,0 +1,44 @@
...
+    //    kid: "2017-03-16-ebe69008de771d62cd1cadf9faa6daae"
+    //    "fxa-createdAt": 1489716000,
+    //  }
+    const now = new Date();
+    const pubKeyHash = crypto
+      .createHash('sha256')
+      .update(kp.public)
+      .digest('hex')
+      .slice(0, 32);
+    const privKey = JwTool.JWK.fromPEM(kp.private, {
```
shane-tomlinson: If you want to stop using fxa-jwtool, you can use `pem2jwk` from the `pem-jwk` module.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/jwks.js
```diff
@@ -25,15 +27,22 @@ const KEYS = (function() {
...
-  var keys = [pub(priv)];
-  var old = config.get('openid.oldKey');
-  if (Object.keys(old).length) {
-    keys.push(pub(old));
+  const keys = [pub(privKey)];
+  if (newPrivKey) {
+    keys.push(pub(newPrivKey));
+  }
+  if (oldPubKey) {
+    // Still use `pub()` here, just in case...
```
shane-tomlinson: 👍

packages/fxa-auth-server/fxa-oauth-server/lib/routes/jwks.js
```diff
@@ -25,15 +27,22 @@ const KEYS = (function() {
...
+    keys.push(pub(oldPubKey));
   }
   return { keys: keys };
 })();
 
 module.exports = {
+  cache: {
```
shane-tomlinson: You added this in another PR, need a rebase?

packages/fxa-auth-server/fxa-oauth-server/scripts/activate-new-signing-key.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const oldKey = keys.extractPublicKey(key);
+  fs.writeFileSync(oldKeyPath, JSON.stringify(oldKey, undefined, 2));
+  console.log('OldKey saved:', keyPath); //eslint-disable-line no-console
+
+  fs.writeFileSync(keyPath, JSON.stringify(newKey, undefined, 2));
+  console.log('Key saved:', keyPath); //eslint-disable-line no-console
+
+  fs.writeFileSync(newKeyPath, JSON.stringify({}, undefined, 2));
+  console.log('NewKey wiped:', keyPath); //eslint-disable-line no-console
```
shane-tomlinson: It might be helpful to print a message about the servers needing to be restart.

packages/fxa-auth-server/fxa-oauth-server/scripts/gen_keys.js
```diff
@@ -12,69 +12,52 @@
  Will create these files
 
  ./config/key.json
+ ./config/newKey.json
  ./config/oldKey.json
 
+ This script is used for initializing keys in a dev environment.
  If these files already exist, this script will show an error message
- and exit. You must remove both keys if you want to generate a new
+ and exit. You must remove all keys if you want to generate a new
```
shane-tomlinson: This seems counter-intuitive.

packages/fxa-auth-server/fxa-oauth-server/scripts/prepare-new-signing-key.js
```diff
@@ -0,0 +1,47 @@
...
+  if (
+    fs.existsSync(newKeyPath) &&
+    Object.keys(require(newKeyPath)).length > 0
+  ) {
+    assert(false, 'new key alreayd exists; perhaps you meant to activate it?');
+  }
+
+  const privKey = keys.generatePrivateKey();
+  fs.writeFileSync(newKeyPath, JSON.stringify(privKey, undefined, 2));
+  console.log('NewKey saved:', keyPath); //eslint-disable-line no-console
```
shane-tomlinson: Worth mentioning the servers need to be restart for the keys to take effect?

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
+
+const keyPath = path.resolve(__dirname, '../config/key.json');
+const oldKeyPath = path.resolve(__dirname, '../config/oldKey.json');
+
+function main(cb) {
```
shane-tomlinson: Worth adding tests for these new scripts to ensure they don't bit rot? I know @philbooth has added script tests in the past for just that reason.

packages/fxa-auth-server/fxa-oauth-server/docs/signing-key-management.md
```diff
@@ -0,0 +1,55 @@
+# Management of JWT Signing Keys
+
+This server uses signed [JWTs](https://jwt.io/) to represent various kinds of security token,
+including:
+
+- `id_token`s as defined by [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
+- `access_token`s as defined by [JWT Profile for OAuth 2.0 Access
```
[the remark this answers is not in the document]
bbangert: I think that would be good to include, especially as it is intended to be the primary event distribution scheme.

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -349,34 +362,62 @@ var options = {
...
+  assert.equal(newKey.kty, 'RSA', 'openid.newKey.kty must be RSA');
+  assert(newKey.kid, 'openid.newKey.kid is required');
+  assert.notEqual(
+    key.kid,
+    newKey.kid,
+    'openid.key.kid must differ from openid.newKey.id'
+  );
+  assert(newKey.n, 'openid.newKey.n is required');
+  assert(newKey.e, 'openid.newKey.e is required');
+  assert(newKey.d, 'openid.newkey.d is required');
```
bbangert: We can't run the keys through Joi schemas?

rfk: Good idea, added.

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -396,34 +418,52 @@ var options = {
...
-  assert.equal(oldKey.kty, 'RSA', 'openid.oldKey.kty must be RSA');
-  assert(oldKey.kid, 'openid.oldKey.kid is required');
-  assert.notEqual(
-    key.kid,
-    oldKey.kid,
-    'openid.key.kid must differ from oldKey'
-  );
-  assert(oldKey.n, 'openid.oldKey.n is required');
-  assert(oldKey.e, 'openid.oldKey.e is required');
-  assert(!oldKey.d, 'openid.oldKey.d is forbidden');
```
rfk: I moved a bunch of this validation logic into `keys.js`, along with a Joi schema for doing it.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/introspect.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
       tokenId = await getTokenId(req.payload.token);
     } catch (err) {
       return {
-        active: false
+        active: false,
```
rfk: An extra change thanks to `npm run format`.

packages/fxa-auth-server/fxa-oauth-server/scripts/gen_keys.js
```diff
@@ -12,75 +12,79 @@
...
+process.env.FXA_OPENID_UNSAFELY_ALLOW_MISSING_ACTIVE_KEY = true;
+
 const fs = require('fs');
 const assert = require('assert');
-const crypto = require('crypto');
-const generateRSAKeypair = require('keypair');
-const JwTool = require('fxa-jwtool');
-
-const keyPath = './fxa-oauth-server/config/key.json';
-const oldKeyPath = './fxa-oauth-server/config/oldKey.json';
```
rfk: So, I didn't like the way this was hard-coding paths, and duplicating values for our config files. I've changed it to look up the appropriate file paths from config, so that it'll run anywhere instead of only working for our dev configuration.

Supporting this is the reason that I've added a fun new `FXA_OPENID_UNSAFELY_ALLOW_MISSING_ACTIVE_KEY` config option. I think the code came up cleaner overall.

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
rfk: Yep! Added.

packages/fxa-auth-server/fxa-oauth-server/test/key-management-scripts.js
```diff
@@ -0,0 +1,100 @@
...
+const crypto = require('crypto');
+const rimraf = require('rimraf');
+
+describe('the signing-key management scripts', () => {
+  let runScript;
+  let workDir, keyFile, newKeyFile, oldKeyFile;
+
+  beforeEach(() => {
+    const uniqName = crypto.randomBytes(8).toString('hex');
+    workDir = path.join(os.tmpdir(), `fxa-oauth-server-tests-${uniqName}`);
```
rfk: I couldn't be bothered trying to decide on which tiny utility library to bring in for making tempdirs, this'll do the job.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -15,37 +141,58 @@ function pub(key) {
...
+  const kp = crypto.generateKeyPairSync('rsa', {
+    modulusLength: 256 * 8,
+    publicKeyEncoding: PEM_ENCODING,
+    privateKeyEncoding: PEM_ENCODING,
+  });
+  // We tag our keys with their creation time, and a unique key id
+  // based on a hash of the public key and the timestamp.  The result
+  // comes out like:
+  //  {
+  //    kid: "20170316-ebe69008"
```
rfk: @shane-tomlinson I went ahead and shorted the kids here in service of #1797, thoughts?

packages/fxa-auth-server/package.json
```diff
@@ -66,7 +66,6 @@
     "jed": "0.5.4",
     "joi": "14.0.0",
     "jsonwebtoken": "^8.4.0",
-    "keypair": "1.0.1",
```
rfk: Node has `crypto.generateKeyPair` natively these days, we don't need to pull in an extra dep for it.

rfk:
Thanks @bbangert @shane-tomlinson, updated based on your feedback.

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
```diff
@@ -0,0 +1,41 @@
...
+
+ */
+
+//eslint-disable no-console
+
+const fs = require('fs');
+const config = require('../lib/config');
+
+// We don't use this, but we load it to check key config.
+require('../lib/keys');
```
bbangert: I'll admit I'm not a fan of relying on import side-effects. Why not have a key verification function that can be called?

This could be done as a follow-up PR though, I don't want to hold this one up anymore.

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
```diff
@@ -0,0 +1,41 @@
...
+  cb = cb || (() => {});
+
+  const oldKeyPath = config.get('openid.oldKeyFile');
+  fs.writeFileSync(oldKeyPath, JSON.stringify({}, undefined, 2));
+  console.log('OldKey wiped:', oldKeyPath);
+
+  console.log(
+    'Please restart the server to stop advertising the old signing key'
+  );
+  return cb();
```
bbangert: nit: `return cb ? cb() : {}` feels a little more intuitive than defaulting it above. But that's prolly just me. 😉

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
rfk: Thanks, yeah, I know what you mean. The competing goal here is to make it very very hard to run the server in a broken configuration, but that might be better achieved by having the server process call a special `validateKeys()` function before starting up.

I'll timebox a brief attempt at such a refactor, as a replacement for the weird "no really I want you to start up with an incomplete config" environment var that's been added here. But I won't block on it if it turns into a rabbit-hole.

rfk: > But I won't block on it if it turns into a rabbit-hole.

I poked at this but ran out of steam pretty quick, let's leave it to a future followup...

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
shane-tomlinson: Looks perfect to me!

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -240,28 +240,51 @@ const conf = convict({
...
+      doc: 'Path to private key JWK to sign various kinds of JWT tokens',
       format: String,
       default: '',
       env: 'FXA_OPENID_KEYFILE',
     },
+    newKeyFile: {
+      doc:
+        'Path to private key JWK that will be used to sign JWTs in the future',
+      format: String,
+      default: '',
```
shane-tomlinson: With the default of `''`, I can't run `NODE_ENV=dev node ./prepare-new-signing-key.js`

```
scripts git:(scripted-key-rotation) NODE_ENV=dev node ./prepare-new-signing-key.js                                                                                            git:(scripted-key-rotation|)
assert.js:339
    throw err;
    ^

AssertionError [ERR_ASSERTION]: openid.newKeyFile not specified
    at main (/Users/shanetomlinson/development/fxa/packages/fxa-auth-server/fxa-oauth-server/scripts/prepare-new-signing-key.js:28:5)
    at Object.<anonymous> (/Users/shanetomlinson/development/fxa/packages/fxa-auth-server/fxa-oauth-server/scripts/prepare-new-signing-key.js:51:3)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:829:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

Maybe in dev.json, set a name for oldKeyFile and newKeyFile so these can be run locally w/o needing to edit anything.

packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
```diff
@@ -15,37 +141,58 @@ function pub(key) {
...
+    .slice(0, 8);
+  const privKey = Object.assign(pem2jwk(kp.privateKey), {
+    kid:
+      now
+        .toISOString()
+        .slice(0, 10)
+        .replace(/-/g, '') +
+      '-' +
+      pubKeyFingerprint,
+    alg: 'RS256',
```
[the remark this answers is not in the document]
rfk: Sounds good, let's do that when we come to implement the different signing key type.

packages/fxa-auth-server/fxa-oauth-server/test/key-management-scripts.js
```diff
@@ -0,0 +1,100 @@
...
+    const activatedKid = JSON.parse(fs.readFileSync(keyFile)).kid;
+    assert.equal(activatedKid, newKid);
+
+    // Which should have moved the previous key to old-key.
+    const retiringKid = JSON.parse(fs.readFileSync(oldKeyFile)).kid;
+    assert.equal(retiringKid, kid);
+
+    // From where we can retire it completely.
+    runScript('retire-old-signing-key.js');
+    assert.equal(fs.readdirSync(workDir).length, 3);
```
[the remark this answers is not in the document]
rfk: It's 3 because of the thing you note below where it writes `{}` into the file.

packages/fxa-auth-server/fxa-oauth-server/scripts/retire-old-signing-key.js
```diff
@@ -0,0 +1,41 @@
...
+const config = require('../lib/config');
+
+// We don't use this, but we load it to check key config.
+require('../lib/keys');
+
+function main(cb) {
+  cb = cb || (() => {});
+
+  const oldKeyPath = config.get('openid.oldKeyFile');
+  fs.writeFileSync(oldKeyPath, JSON.stringify({}, undefined, 2));
```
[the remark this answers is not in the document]
rfk: I was doing this because config-loading would freak out if the file was missing, but would happily load an empty object from the file as a signal that there's no key here. With the refactor of how keys are loaded, that's probably no longer necessary, and just unlinking the file would be better.

rfk:
> I don't know if we'd use the scripts verbatim in production, but would be happy to modify them as
> needed so they're useful, if that's how we want to proceed here.

I had a chat with @jrgm and @jbuck about this, and it sounds like we have a good path for writing a little wrapper around these scripts that will let them operate on production values. We can pursue that in a follow-up issue.

PR #1669, opened 2019-07-04
chore(subs): remove `|| []` from call to db.fetchAccountSubscriptions

philbooth:
Not connected to an issue, I noticed this while looking at someone else.

There was an unnecessary and misleading `|| []` after a call to `db.fetchAccountSubscriptions`, so I just removed it and added some test coverage to assert that the db does indeed return the empty array in the default case.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
+++ b/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
@@ -2538,101 +2538,134 @@ module.exports = function(config, DB) {
     });
 
     describe('db.securityEventsByUid', () => {
-      let session1, session2, session3, anotherAccountSession, uid, anotherUid
-      const evA = 'account.login', evB = 'account.create', evC = 'account.reset'
-      const addr = '127.0.0.1'
+      let session1, session2, session3, anotherAccountSession, uid, anotherUid;
+      const evA = 'account.login',
+        evB = 'account.create',
+        evC = 'account.reset';
+      const addr = '127.0.0.1';
 
       function insert(uid, addr, name, session) {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -2538,101 +2538,134 @@ module.exports = function(config, DB) {
...
 
     describe('db.securityEventsByUid', () => {
-      let session1, session2, session3, anotherAccountSession, uid, anotherUid
-      const evA = 'account.login', evB = 'account.create', evC = 'account.reset'
-      const addr = '127.0.0.1'
+      let session1, session2, session3, anotherAccountSession, uid, anotherUid;
+      const evA = 'account.login',
+        evB = 'account.create',
+        evC = 'account.reset';
+      const addr = '127.0.0.1';
```
philbooth: Prettier is complicating the diff here. The substantive change is further down...

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -4254,6 +4287,12 @@ module.exports = function(config, DB) {
         }
       });
 
+      it('should fetch an empty array when there are no subscriptions', async () => {
+        const result = await db.fetchAccountSubscriptions(account.uid);
+        assert.isArray(result);
+        assert.lengthOf(result, 0);
+      });
```
philbooth: ...here you go.

philbooth:
Thanks @vbudhram, although shouldn't you be enjoying the holiday rather than reviewing code?! :smile:

Issue #1071, opened 2019-05-10
Handle Account Deletion for Subscribing Users


johngruen:
* FXA should warn users that they will lose access to any subscriptions they currently use
* FXA should immediately cancel payment as of the current billing cycle for the user
* FXA should tell Stripe to remove all payment information

Account deletion wires can be found here: 

https://mozilla.invisionapp.com/share/K2RBDY3VU9X

lmorchard:
FWIW, I think we have subscription cancellation happening in the backend as part of user deletion. But, the UI for notifying the user of lost access and the new confirmation checkboxes are *not* done yet.

lmorchard:
We discussed this in triage and worked out that literally depending on Relying Parties (#710) could be harder than necessary. (This is what we figured we'd need to do earlier.)

We *do* already track subscription plans for the user. So rather than listing the Relying Parties, we can list just the subscriptions. Those should describe what the user will lose on deletion.

lmorchard:
Ref: https://github.com/mozilla/subhub/issues/61

Need a way to completely delete the customer in subhub while deleting the FxA user

philbooth:
> Need a way to completely delete the customer in subhub while deleting the FxA user

Ref: https://github.com/mozilla/subhub/commit/a675893ea607f08d106b468614c5c44257c70e2d

philbooth:
> Account deletion wires can be found here: 
> 
> https://mozilla.invisionapp.com/share/K2RBDY3VU9X

Maybe I'm holding it wrong, but I don't see anything for account deletion there. I see stuff for cancelling individual subscriptions though. What am I missing?

philbooth:
#1671 is open for a small, backend part of this. Waiting on wireframes for the front-end part though, so taking my face off and adding `needs:ux`.

philbooth:
Correct wireframe link:

https://mozilla.invisionapp.com/share/2MRH9FQHQDT#/screens/358433257_Cover_Page_JC

philbooth:
@LZoog, if you're still hunting for issues to work on, the outstanding portion of this might be an interesting one to take up. A back-end piece landed in #1671, but the UI still needs adding to the settings screen in the content server and there is some API wiring needed to fetch the list of subscriptions too.

Let me know if you're interested and I can walk you through it / answer questions.

LZoog:
Thanks @philbooth, I just scheduled a meeting for us tomorrow to go over this. I'm reassigning this issue to me.

[...]

PR #1671, opened 2019-07-04
feat(subs): delete customer from subhub when destroying account

philbooth:
Related to #1071.

Invokes the new `delete_customer` endpoint added to subhub in https://github.com/mozilla/subhub/commit/a675893ea607f08d106b468614c5c44257c70e2d, so we don't have to iterate through subscriptions and cancel them one-by-one. I haven't done anything about the front-end part of the linked issue yet because I'm not sure what it's supposed to look like.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1429,18 +1429,7 @@ module.exports = (
         const { uid } = emailRecord;
 
         if (config.subscriptions && config.subscriptions.enabled) {
-          // TODO: We should probably delete the upstream customer for
-          // subscriptions, but no such API exists in subhub yet.
-          // https://github.com/mozilla/subhub/issues/61
-
-          // Cancel all subscriptions before deletion, if any exist
-          // Subscription records will be deleted from DB as part of account
-          // deletion, but we have to trigger cancellation in payment systems
-          const subscriptions = await db.fetchAccountSubscriptions(uid);
-          for (const subscription of subscriptions) {
-            const { subscriptionId } = subscription;
-            await subhub.cancelSubscription(uid, subscriptionId);
-          }
+          await subhub.deleteCustomer(uid);
         }
 
         // We fetch the devices to notify before deleteAccount()
--- a/packages/fxa-auth-server/lib/subhub/client.js
+++ b/packages/fxa-auth-server/lib/subhub/client.js
@@ -42,6 +42,7 @@ module.exports = function(log, config) {
       'createSubscription',
       'getCustomer',
       'updateCustomer',
+      'deleteCustomer',
       'cancelSubscription',
       'reactivateSubscription',
     ].reduce(
@@ -107,6 +108,17 @@ module.exports = function(log, config) {
           validators.subscriptionsCustomerValidator,
           ErrorValidator
         ),
+      },
+    },
+
+    deleteCustomer: {
+      path: '/v1/customer/:uid',
+      method: 'DELETE',
+      validate: {
+        params: {
+          uid: isA.string().required(),
+        },
+        response: isA.alternatives(MessageValidator, ErrorValidator),
       },
     },
 
@@ -287,6 +299,20 @@ module.exports = function(log, config) {
             throw error.rejectedCustomerUpdate(err.message, err);
           }
         }
+        throw err;
+      }
+    },
+
+    async deleteCustomer(uid) {
+      try {
+        return await api.deleteCustomer(uid);
+      } catch (err) {
+        log.error('subhub.deleteCustomer.failed', { uid, err });
+
+        if (err.statusCode === 404) {
+          throw error.unknownCustomer(uid);
+        }
+
         throw err;
       }
     },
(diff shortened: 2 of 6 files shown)


packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -193,7 +205,11 @@ module.exports = function(log, config) {
           orig_system: ORIG_SYSTEM,
         });
       } catch (err) {
-        if (err.statusCode === 400 || err.statusCode === 402 || err.statusCode === 404) {
+        if (
+          err.statusCode === 400 ||
+          err.statusCode === 402 ||
+          err.statusCode === 404
+        ) {
```
philbooth: Prettier doing its thing again.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1423,21 +1423,7 @@ module.exports = (
...
-                    // deletion, but we have to trigger cancellation in payment systems
-                    const uid = emailRecord.uid;
-                    const subscriptions = await db.fetchAccountSubscriptions(
-                      uid
-                    );
-                    for (const subscription of subscriptions) {
-                      const { subscriptionId } = subscription;
-                      await subhub.cancelSubscription(uid, subscriptionId);
-                    }
+                    await subhub.deleteCustomer(emailRecord.uid);
```
6a68: Still getting up to speed on this repo, so forgive my naive questions--just trying to understand how this all fits together.

I'm trying to understand what happens if this line [throws](https://github.com/mozilla/fxa/pull/1671/files#diff-cc3fa280c8b38abc492139b546e1f9daR313) (say, due to a 4xx or 5xx error with subhub). I think this inner async function would return a rejected Promise, but this chain doesn't have a rejection handler, and I don't see how a rejected Promise would be handled by the next outer function, either. Is there implicit error handling somewhere? 

I noticed a sort-of related [test](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/test/local/routes/subscriptions.js#L507) that exercises the `subhub.cancelSubscription()` failed case, but it doesn't look like it was updated here. Would it be good to add new tests for `subhub.deleteCustomer`, including failing because of returned 4xx or 5xx errors, or are those tests somewhere else, and I haven't found them yet, or is that just not needed?

I wonder if it might be easier to reason about this handler if the whole thing were converted to await style, instead of chaining / nesting promises with awaits sprinkled in--or maybe I'll get used to it eventually :-)

6a68:
Re-tagging fxa-devs for a review, because I don't think I have enough context to approve changes yet

packages/fxa-auth-server/lib/routes/account.js
philbooth: > I wonder if it might be easier to reason about this handler if the whole thing were converted to await style, instead of chaining / nesting promises with awaits sprinkled in--or maybe I'll get used to it eventually :-)

No, you're right, it is weird and I will refactor it. The mix of styles is just from where code has grown organically and collectively we've been too lazy (or rather, it's not been prioritised) to make a wholesale switch to async/await. As you well know, some of the FxA code is pretty old. 😄

> Is there implicit error handling somewhere?

The extra error handling you're looking for is in `lib/server.js` (which is probably quite a useful module to spend a bit of time getting to know):

https://github.com/mozilla/fxa/blob/95ae176c546bdc33abe18d0f97b9dba4b98d79ef/packages/fxa-auth-server/lib/server.js#L282-L288

That in turn calls `AppError.translate`, which enters the fallback condition for returning a `500` error:

https://github.com/mozilla/fxa/blob/95ae176c546bdc33abe18d0f97b9dba4b98d79ef/packages/fxa-auth-server/lib/error.js#L196-L204

> Would it be good to add new tests for `subhub.deleteCustomer`, including failing because of returned 4xx or 5xx errors, or are those tests somewhere else, and I haven't found them yet, or is that just not needed?

Good point, I'll add a test.

packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -279,5 +299,19 @@ module.exports = function(log, config) {
         throw err;
       }
     },
+
+    async deleteCustomer(uid) {
+      try {
+        return await api.deleteCustomer(uid);
+      } catch (err) {
+        log.error('subhub.deleteCustomer.1', { uid, err });
```
bbangert: `.1`? Just wondering if this is another pattern I'm not familiar with.

packages/fxa-auth-server/lib/routes/account.js
6a68: ohhhh, I see. Thanks for filling in some of the details for me!

packages/fxa-auth-server/lib/subhub/client.js
6a68: I was wondering the same thing TBH

philbooth: I confess, I was just following the convention from the rest of the module here, although I've seen it elsewhere in the auth server too. Typically when writing fresh modules from scratch, I prefer to use more descriptive error names. I'll do that here too instead, thanks for the nudge.

Issue #1673, opened 2019-07-05
Is prettier configured correctly for the payments server?


philbooth:
I made some fairly innocuous changes to add Sentry to the payments server client-side, but when the prettier hook kicked in it also replaced a bunch of single-quotes with double-quotes, which I wasn't expecting:

https://github.com/mozilla/fxa/compare/pb/1299-payments-client-sentry?expand=1

Everywhere else still seems to use single quotes.

Do we need to do something to configure prettier properly for the `fxa-payments-server` package? /cc @dannycoates

philbooth:
> Do we need to do something to configure prettier properly for the `fxa-payments-server` package? /cc @dannycoates

Never mind, fixed in #1676.

philbooth:
Gah, I was wrong! I just rebased my other branch and ran the hook again, it's still flipping the quotes. Re-opening this, sorry.

dannycoates:
I avoided making prettier PRs to the subscription packages because I didn't know how much work in progress it would disrupt. I'm happy to make a formatting PR.

philbooth:
Oh, fair point about the work in progress I guess. I've got into the habit of using `--no-verify` in that directory now, so I'm fine if we close this.

bbangert:
@dannycoates event-broker and payments-server should already be using prettier so there shouldn't be any changes (in theory).

LZoog:
I came across this while working on #2053.

It seems to me like this should be reopened?

6a68:
Reopening to discuss in triage. (@LZoog feel free to reopen bugs :+1:)

LZoog:
@6a68 thanks! I had this issue open to bring up at the end of triage, I'll just reopen from now on and we can close otherwise. :)

lmorchard:
Actually, looks like we never configured prettier for payments-server. 

Brace for incoming enormous PR...

PR #1674, opened 2019-07-05
feat(payments): add sentry to client-side

### LINKED ISSUE #1299: Hook up payments server client-side with Sentry
Related to #1081. I should probably have done it as part of #1298 where the back-end got hooked up, but posted this follow-up issue instead.

philbooth:
Fixes #1299.

Adds sentry to the client-side of the payments server. Uses all the default options, which seem like they should be fine:

https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/options.ts

Only a draft for now because `prettier` has made some unexpected changes on commit. Ref: #1673

### DIFF
--- a/packages/fxa-payments-server/server/lib/server.js
+++ b/packages/fxa-payments-server/server/lib/server.js
@@ -27,6 +27,7 @@ module.exports = () => {
   // Each of these config values (e.g., 'servers.content') will be exposed as the given
   // variable to the client/browser (via fxa-content-server/config)
   const CLIENT_CONFIG = {
+    sentryDsn: config.get('sentryDsn'),
     servers: {
       auth: {
         url: config.get('servers.auth.url'),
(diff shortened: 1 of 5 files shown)


philbooth:
Un-drafting this. I spent a bit of time trying to sort out `prettier` (ref: #1673), but I still can't make it work. In the meantime I re-committed this using `--no-verify` to prevent the quotes from being flipped.

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -27,31 +27,28 @@ module.exports = () => {
   // Each of these config values (e.g., 'servers.content') will be exposed as the given
   // variable to the client/browser (via fxa-content-server/config)
   const CLIENT_CONFIG = {
+    sentryDsn: config.get('sentryDsn'),
     servers: {
       auth: {
-        url: config.get('servers.auth.url'),
+        url: config.get('servers.auth.url')
```
philbooth: Sigh. I used `eslint . --fix` to put the quotes back how they were and in doing so it's changed all this other stuff too. One sec: I'll checkout, re-apply then force push this with just my changes.

PR #1679, opened 2019-07-05
feat(oauth): Add pairwise pseudonymous identifiers in access/id tokens

### LINKED ISSUE #1143: Support OIDC pairwise pseudonymous identifiers
[Pairwise pseudonymous identifiers](https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes) "provides a different sub value to each Client, so as not to enable Clients to correlate the End-User's activities without permission."

> When pairwise Subject Identifiers are used, the OpenID Provider MUST calculate a unique sub (subject) value for each Sector Identifier. The Subject Identifier value MUST NOT be reversible by any party other than the OpenID Provider. 

The OIDC spec mentions a bunch about client dynamic registration and sector_identifier_uri. I think for speed of implementation we can start with a configurable list of client_ids that automatically get PPIDs. Anywhere we return a uid, return the PPID instead: 

* [id_tokens](https://github.com/mozilla/fxa/blob/361d46051162797fd4f06bca4bcd64064983274f/packages/fxa-auth-server/fxa-oauth-server/lib/token.js#L55)
* [/verify](https://github.com/mozilla/fxa/pull/933/files#diff-3e9b7e97bb4e3d3c952bdf62329c9820R187)
* [/introspect](https://github.com/mozilla/fxa/pull/933/files#diff-3e9b7e97bb4e3d3c952bdf62329c9820R187)
* [jwt access tokens](https://github.com/mozilla/fxa/pull/933/files#diff-3e9b7e97bb4e3d3c952bdf62329c9820R187)

PPIDs are meant to protect a user's privacy while providing the RP with a stable identifier. An RP that requests PPIDs should have access to no other PPI - we return emails in all of the same endpoints listed above. If an RP uses PPIDs, we should not return the email in the response.

I wonder if we should ban RPs that use PPIDs from requesting the `profile` scope.

A product we are going to ship is going to require PPIDs with a slight twist. I'll open the slight twist in another issue.

Follow on to https://github.com/mozilla/fxa/issues/935

cc @martinthomson, @dannycoates 

I set the size to 8 because there is still some uncertainty as to what all is involved here. @dannycoates if that seems incorrect, feel free to update accordingly.

### LINKED ISSUE #1144: Support rotating pairwise pseudonymous identifiers
This is follow on from #1143. 

[Pairwise pseudonymous identifiers](https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes) "provides a different sub value to each Client, so as not to enable Clients to correlate the End-User's activities without permission."

This takes that concept a step further to support _rotating_ PPIDs based on an RP provided `counter` that allows the RP to control the rotation.

@martinthomson spec'd out an approach:

```
The FxA server could mint identifiers consistently.  Given a secret that was created for this specific purpose and an update interval, the identifier could be produced with:

    ctxt = floor(now() / update_interval) “.” account
    sub = HKDF-Expand(secret, ctxt, L)

Any injective mapping from interval and account used as input to a pseudo random function would be sufficient. Alternatively, the secret could be rotated on the interval.

If we wanted to allow separate identifiers for private browsing (or containers), then we would need to maintain a counter for each interval. That would be added to the context:

    pctxt = floor(now() / update_interval) “.” account “.” counter++

With the counter being reset each time the update interval rolls over.
```

This is similar to the suggested methods from the OIDC spec:

```
    The Sector Identifier can be concatenated with a local account ID and a salt value that is kept secret by the Provider. The concatenated string is then hashed using an appropriate algorithm.

    Calculate sub = SHA-256 ( sector_identifier || local_account_id || salt ).

    The Sector Identifier can be concatenated with a local account ID and a salt value that is kept secret by the Provider. The concatenated string is then encrypted using an appropriate algorithm.

    Calculate sub = AES-128 ( sector_identifier || local_account_id || salt ).

    The Issuer creates a Globally Unique Identifier (GUID) for the pair of Sector Identifier and local account ID and stores this value.
```

In any of those we could || counter as well.

cc @martinthomson, @dannycoates

shane-tomlinson:
Also enables PPID rotation based on both time and a client sent seed

fixes #1143
fixes #1144

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/config/dev.json
+++ b/packages/fxa-auth-server/fxa-oauth-server/config/dev.json
@@ -188,13 +188,21 @@
     "enabled": true,
     "enabledClientIds": [
       "98e6508e88680e1a",
-      "dcdb5ae7add825d2"
+      "dcdb5ae7add825d2",
+      "325b4083e32fe8e7"
     ]
   },
   "openid": {
     "issuer": "http://127.0.0.1:3030",
     "keyFile": "../config/key.json",
     "key": {}
+  },
+  "ppid": {
+    "enabled": true,
+    "enabledClientIds": ["325b4083e32fe8e7"],
+    "rotatingClientIds": ["325b4083e32fe8e7"],
+    "rotationPeriodMS": 30000,
+    "salt": "a new ppid salt"
   },
   "allowHttpRedirects": true,
   "authServerSecrets": ["megaz0rd"],
--- a/packages/fxa-auth-server/fxa-oauth-server/config/test.json
+++ b/packages/fxa-auth-server/fxa-oauth-server/config/test.json
@@ -1,4 +1,8 @@
 {
+  "browserid": {
+    "issuer": "127.0.0.1:9000",
+    "verificationUrl": "http://127.0.0.1:5050/v2"
+  },
   "clientManagement": {
     "enabled": true
   },
@@ -64,7 +68,7 @@
     {
       "id": "38a6b9b3a65a1872",
       "hashedSecret": "289a885946ee316844d9ffd0d725ee714901548a1e6507f1a40fb3c2ae0c99f1",
-      "name": "Public Client PKCE with no allowedScoeps",
+      "name": "Public Client PKCE with no allowedScopes",
       "imageUri": "https://mozorg.cdn.mozilla.net/media/img/firefox/new/header-firefox.png",
       "redirectUri": "https://example.domain/return?foo=bar",
       "trusted": true,
@@ -91,6 +95,26 @@
       "trusted": true,
       "canGrant": true,
       "publicClient": false
+    },
+    {
+      "id": "38a6b9b3a65a1874",
+      "hashedSecret": "289a885946ee316844d9ffd0d725ee714901548a1e6507f1a40fb3c2ae0c99f1",
+      "name": "PPID JWT Client",
+      "imageUri": "https://jwt.domain/logo",
+      "redirectUri": "https://jwt.domain/redirect",
+      "trusted": true,
+      "canGrant": true,
+      "publicClient": false
+    },
+    {
+      "id": "38a6b9b3a65a1875",
+      "hashedSecret": "289a885946ee316844d9ffd0d725ee714901548a1e6507f1a40fb3c2ae0c99f1",
+      "name": "Rotating PPID JWT Client",
+      "imageUri": "https://jwt.domain/logo",
+      "redirectUri": "https://jwt.domain/redirect",
+      "trusted": true,
+      "canGrant": false,
+      "publicClient": false
     }
   ],
   "logging": {
@@ -100,14 +124,24 @@
   "jwtAccessTokens": {
     "enabled": true,
     "enabledClientIds": [
-      "38a6b9b3a65a1873"
+      "38a6b9b3a65a1873",
+      "38a6b9b3a65a1874",
+      "38a6b9b3a65a1875"
     ]
   },
   "openid": {
-    "keyFile": "../config/key.json",
-    "oldKeyFile": "../config/oldKey.json",
+    "issuer": "http://127.0.0.1:3030",
     "key": {},
-    "oldKey": {}
+    "keyFile": "../config/key.json",
+    "oldKey": {},
+    "oldKeyFile": "../config/oldKey.json"
+  },
+  "ppid": {
+    "enabled": true,
+    "enabledClientIds": ["38a6b9b3a65a1874", "38a6b9b3a65a1875"],
+    "rotatingClientIds": ["38a6b9b3a65a1875"],
+    "rotationPeriodMS": 100,
+    "salt": "a new ppid salt"
   },
   "allowHttpRedirects": true,
   "authServerSecrets": ["meg@z0rd", "v0ltr0n"],
--- a/packages/fxa-auth-server/fxa-oauth-server/docs/api.md
+++ b/packages/fxa-auth-server/fxa-oauth-server/docs/api.md
@@ -393,6 +393,11 @@ The following types of grant are possible:
 
 #### Request Parameters
 
+- `ppid_seed`: (optional) Seed used in `sub` claim generation of
+  JWT access tokens/ID tokens for clients with [Pseudonymous Pairwise
+  Identifiers (PPID)](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md)
+  enabled. Used to forcibly rotate the `sub` claim. Must be an integer in the range 0-1024.
+  Defaults to 0.
 - `ttl`: (optional) Seconds that the access_token should be valid.
   If unspecified this will default to the maximum value allowed by the
   server, which is a configurable option but would typically be measured
--- a/packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
+++ b/packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
@@ -1,0 +1,68 @@
+# Pairwise Pseudonymous Identifiers
+
+A Pairwise Pseudonymous Identifier (PPID) is defined in the [OpenID Connect Spec][#oidc-spec] as:
+
+> Identifier that identifies the Entity to a Relying Party that cannot be correlated with the Entity's PPID at another Relying Party.
+
+Put another way, PPIDs provide a single user with distinct userids across distinct Relying Parties (RPs), enhancing user privacy by preventing cross-site correlation based on userid.
+
+By default, Firefox accounts provides the same userid to all Mozilla-internal RPs to facilitate cross service correlation. As Mozilla expands its product offering to include white labeled 3rd party services, we want to uphold Mozilla's principles regarding user privacy by providing external services with only the minimal data necessary.
+
+PPIDs allow Mozilla to enforce our privacy stance by providing each PPID enabled RP with a distinct userid for a given user. For example, a user that signs into 2 PPID enabled services would have 3 distinct userids:
+
+- FxA userid, only seen by Mozilla properties
+- PPID for service 1, known by Mozilla and service 1
+- PPID for service 2, known by Mozilla and service 2
+
+PPIDs prevents service 1 and service 2 from correlating user data with each other based on userid alone. _PPIDs only prevent correlation based on userid_, transient information such as IP address, shared 3rd party cookies, and user-agent information can still be used by RPs to correlate users. This also assumes that RPs are unable to fetch user profile information such as email address and display name that could be used as additional data points.
+
+PPIDs are returned as the `sub` claim within an [OIDC ID Token][#oidc-id-token] or [OAuth JWT Access Token][#oauth-jwt-access-token].
+
+## Further enhancing privacy through PPID rotation
+
+User privacy can be further enhanced by preventing long term profiles from being built for a given user through PPID rotation. Firefox accounts supports two methods of PPID rotation, RP initiated, and server initiated.
+
+### RP initiated PPID rotation
+
+Any PPID enabled RP that receives JWT access tokens or OIDC ID tokens can initiate PPID rotation by specifying an additional parameter, `ppid_seed`, when requesting tokens from `/token` endpoints on the OAuth server. `ppid_seed` must be an integer between 0 and 1024, and can be rotated at any time.
+
+### Server initiated PPID rotation
+
+By default, Firefox accounts does not enforce sub rotation, but for sensitive RPs where long term user profiles are undesirable, Firefox accounts can enforce periodic rotation without depending on the RP to pass in a `ppid_seed`. The default period is 6 hours.
+
+## Ensuring 3rd parties are unable to access user information
+
+An OAuth access token is a bearer token that can be presented by anyone to an OAuth service provider to access a protected resource. If a JWT access token contains a PPID `sub` claim that is meant to protect a user's privacy, but is granted the `profile` scope allows the service provider to present the same access token to the FxA profile server and learn the user's true identity. As such, tokens _MUST NOT_ be requested with the `profile` scope or any of its implicants such as `profile:uid` or `profile:email`.
+
+## PPIDs and logging
+
+PPIDs are not used by internal Mozilla properties and are not logged.
+
+## PPID generation
+
+PPIDs are generated using [HKDF][#hkdf] with the following inputs:
+
+- **km**: `client_id`.`userid_hex`.`ppid_seed||0`.`serverEnforcedRotationInput`
+- **info**: `oidc ppid sub`
+- **salt**: A secret configured by Ops
+- **len**: 16 bytes, which leads to a 32 hex character result
+
+### Server enforced rotation input
+
+The server enforced rotation input defaults to the string `0`. If server enforced rotation is enabled for the client, it is calculated as follows:
+
+```
+serverEnforcedRotationInput = Math.floor(Date.now() / ROTATION_PERIOD)
+```
+
+If an RP requests two tokens in short succession, their `sub` claims could be different if the first token was generated previous to the rotation epoch and the second after.
+
+## The future
+
+Firefox accounts does not currently support [sector identifiers][#oidc-sector-identifier]. Sector identifiers allow multiple client_ids to receive the same PPID, this would be useful if the RP has applications on multiple platforms, e.g., an app on Android, iOS, and an addon in Firefox.
+
+[#oidc-spec]: https://openid.net/specs/oidc-core-1_0.html#Terminology
+[#oidc-id-token]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+[#oidc-sector-identifier]: https://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg
+[#oauth-jwt-access-token]: https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00
+[#hkdf]: https://tools.ietf.org/html/rfc5869
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
@@ -276,6 +276,39 @@ const conf = convict({
       env: 'FXA_OPENID_TTL',
     },
   },
+  ppid: {
+    enabled: {
+      doc: 'Whether pairwise pseudonymous identifiers (PPIDs) are enabled',
+      format: Boolean,
+      default: true,
+      env: 'PPID_ENABLED',
+    },
+    enabledClientIds: {
+      doc: 'client_ids that receive PPIDs',
+      format: Array,
+      default: [],
+      env: 'PPID_CLIENT_IDS',
+    },
+    rotatingClientIds: {
+      doc:
+        'client_ids that receive automatically rotating PPIDs based on server time',
+      format: Array,
+      default: [],
+      env: 'PPID_ROTATING_CLIENT_IDS',
+    },
+    rotationPeriodMS: {
+      doc: 'salt used in HKDF for PPIDs, converted to milliseconds',
+      format: 'duration',
+      default: '6 hours',
+      env: 'PPID_ROTATION_PERIOD',
+    },
+    salt: {
+      doc: 'salt used in HKDF for PPIDs',
+      format: String,
+      default: 'YOU MUST CHANGE ME',
+      env: 'PPID_SALT',
+    },
+  },
   publicUrl: {
     format: 'url',
     default: 'http://127.0.0.1:9010',
@@ -391,6 +424,10 @@ if (Object.keys(oldKey).length) {
   assert(oldKey.n, 'openid.oldKey.n is required');
   assert(oldKey.e, 'openid.oldKey.e is required');
   assert(!oldKey.d, 'openid.oldKey.d is forbidden');
+}
+
+if (conf.get('ppid.enabled')) {
+  assert.notEqual(conf.get('ppid.salt'), 'YOU MUST CHANGE ME');
 }
 
 module.exports = conf;
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
@@ -16,6 +16,7 @@ const amplitude = require('./metrics/amplitude')(
   logger,
   config.getProperties()
 );
+const sub = require('./jwt_sub');
 
 const ACR_VALUE_AAL2 = 'AAL2';
 const ACCESS_TYPE_OFFLINE = 'offline';
@@ -168,10 +169,10 @@ module.exports.generateTokens = async function generateTokens(grant) {
   return result;
 };
 
-function generateIdToken(grant, access) {
+async function generateIdToken(grant, access) {
   var now = Math.floor(Date.now() / 1000);
   var claims = {
-    sub: hex(grant.userId),
+    sub: await sub(grant.userId, grant.clientId, grant.ppidSeed),
     aud: hex(grant.clientId),
     //iss set in jwt.sign
     iat: now,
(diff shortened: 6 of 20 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
```diff
@@ -0,0 +1,44 @@
...
+  userIdBuf,
+  clientIdBuf,
+  clientSeed = 0
+) {
+  const userIdHex = hex(userIdBuf);
+  const clientIdHex = hex(clientIdBuf);
+
+  // TODO - constrain clientSeed.
+
+  if (PPID_ENABLED && PPID_CLIENT_IDS.has(clientIdHex)) {
```
shane-tomlinson: @martinthomson, could I get your feedback on whether is reasonable/matches your expectations?

martinthomson: Is the idea here that you would configure PPID as a global setting and that it could be requested on a request-by-request basis?

Because I would have thought that if PPID was enabled, you would want an error if there was no client-provided ID.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
```diff
@@ -0,0 +1,44 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const hex = require('buf').to.hex;
+const hkdf = require('../../lib/crypto/hkdf');
+const config = require('./config');
+
+const PPID_ENABLED = config.get('ppid.enabled');
+const PPID_CLIENT_IDS = new Set(config.get('ppid.enabledClientIds'));
+const PPID_ROTATING_CLIENT_IDS = new Set(config.get('ppid.rotatingClientIds'));
```
martinthomson: Do you even use this?

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
```diff
@@ -0,0 +1,44 @@
...
+
+module.exports = async function generateSub(
+  userIdBuf,
+  clientIdBuf,
+  clientSeed = 0
+) {
+  const userIdHex = hex(userIdBuf);
+  const clientIdHex = hex(clientIdBuf);
+
+  // TODO - constrain clientSeed.
```
martinthomson: At a minimum, please scrub both for a period '.' or the function below won't be injective.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
```diff
@@ -0,0 +1,44 @@
...
+
+  if (PPID_ENABLED && PPID_CLIENT_IDS.has(clientIdHex)) {
+    let timeBasedContext = 0;
+    if (PPID_ROTATING_CLIENT_IDS.has(clientIdHex)) {
+      timeBasedContext = Math.floor(Date.now() / PPID_ROTATION_PERIOD_MS);
+    }
+    return hex(
+      await hkdf(
+        `${clientIdHex}.${userIdHex}.${clientSeed}.${timeBasedContext}`,
```
martinthomson: So you have two inputs here from the client: one is the seed (`clientSeed`) and the other is the id (`clientIdHex`).  If the intent is that one is a unique ID for the device (or an index into the devices on the account) and the other is given to each device to choose, then this is probably right.

Documenting all of that clearly in the description of the configuration would help.  I'm not following the description you have.

I think that both can be (small) integers, unless you have other constraints on the format of device identifiers that you can rely on.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
```diff
@@ -141,6 +141,11 @@ const PAYLOAD_SCHEMA = Joi.object({
     is: GRANT_FXA_ASSERTION,
     otherwise: Joi.forbidden(),
   }),
+
+  ppid_seed: Joi.number()
+    .greater(-1)
+    .less(17)
```
martinthomson: 0..16 is an unusual range.  Do you have any analysis to support that?

shane-tomlinson: Nope, just a number pulled out of thin air.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
shane-tomlinson: https://github.com/mozilla/fxa/pull/1679/files#diff-8e82b39c55246a507d7c2a3fe5dc263aR30

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
shane-tomlinson: > Is the idea here that you would configure PPID as a global setting and that it could be requested on a request-by-request basis?

The idea is that PPIDs are enabled on both a global and per client basis. Per-client only matters if enabled on a global basis. We do this sort of gating often in FxA to quickly disable a feature globally if it is found to cause some sort of problem.

> Because I would have thought that if PPID was enabled, you would want an error if there was no client-provided ID.

The only way to get to this function is via the /token endpoint, which requires a [client_id to be passed](https://github.com/mozilla/fxa/pull/1679/files#diff-19a8f6d558692592349f678e6cc65b38R64) in either the Authorization header or POST request body.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
shane-tomlinson: > So you have two inputs here from the client: one is the seed (clientSeed) and the other is the id (clientIdHex). If the intent is that one is a unique ID for the device (or an index into the devices on the account) and the other is given to each device to choose, then this is probably right.

That's right, the idea is that both the addon and FxA can enforce key rotation. The clientId and userId will be the same for a given user all Firefox instances where they have enabled the proxy. The addon could send distinct ppid_seeds for, e.g., private browsing mode and "normal" browsing mode. This would cause distinct subs to be generated for each context. Alternatively, if the secure proxy team wants users to appear distinct on each device, they could send random ppid_seeds on each device and enforce a key change whenever they see fit by changing the seed.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js
shane-tomlinson: Note to self, update the /authorization endpoint w/ ppid_seed and ensure the client_id is required when generating access tokens.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -4771,6 +4816,380 @@ describe('/v1', function() {
...
+          refresh_token: tokenResult.result.refresh_token,
+        },
+      });
+      assert.equal(sameSeedRefreshTokenResult.statusCode, 200);
+      assert.isNull(
+        validators.jwt.validate(sameSeedRefreshTokenResult.result.access_token).error
+      );
+      const sameSeedJWT = decodeJWT(sameSeedRefreshTokenResult.result.access_token);
+
+      assert.strictEqual(sameSeedJWT.claims.sub, tokenJWT.claims.sub);
```
shane-tomlinson: I'm not really happy with this test because the sub could rotate due to server time anyways. I'm unsure of a good way to test this w/o depending on really long delays, and even then it's possible for server clock based rotation.

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
```diff
@@ -188,14 +188,20 @@
...
   },
   "openid": {
     "issuer": "http://127.0.0.1:3030",
     "keyFile": "../config/key.json",
     "key": {}
   },
+  "ppid": {
+    "enabled": true,
+    "enabledClientIds": ["ea3ca969f8c6bb0d"],
+    "rotatingClientIds": ["ea3ca969f8c6bb0d"]
```
shane-tomlinson: I'm not really happy with this either, this was only done to make the [oauth_tests.js](https://github.com/mozilla/fxa/blob/b91441baa744e27e9e1e89049f61f8f8a503ca55/packages/fxa-auth-server/test/remote/oauth_tests.js#L263) module pass, I would have asssumed the test OAuth server would have started using test.json, but it's started using dev.json. When I force the NODE_ENV environment variable to test for the server, then I caused other errors. Still may be something that's straight forward to figure out.

shane-tomlinson:
Moving this to "ready for review". @rfk and @martinthomson, could I get your feedback on this? Should I rope in anyone else?

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -276,6 +276,38 @@ const conf = convict({
...
+      default: true,
+      env: 'PPID_ENABLED',
+    },
+    enabledClientIds: {
+      doc: 'client_ids that receive PPIDs',
+      format: Array,
+      default: [],
+      env: 'PPID_CLIENT_IDS',
+    },
+    rotatingClientIds: {
```
shane-tomlinson: For flexibility, perhaps combine `rotatingClientIds` and `rotationPeriodMS` such that the rotation period for each client is configurable and have the default if a client is not in the list be `Infinity` (no rotation).

```js
clientRotationPeriodMS: {
  '38a6b9b3a65a1873': 21600000
}
```

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -133,9 +134,9 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
 //
 // This function does *not* perform any authentication or validation, assuming that
 // the specified grant has been sufficiently vetted by calling code.
-module.exports.generateTokens = async function generateTokens(grant) {
+module.exports.generateTokens = async function generateTokens(grant, ppidSeed) {
   // We always generate an access_token.
-  const access = await exports.generateAccessToken(grant);
+  const access = await exports.generateAccessToken(grant, ppidSeed);
```
shane-tomlinson: Should we do a check here that for PPID enabled clients, that the grant should not contain the `profile` scope or any of its implicants? Otherwise the service provider or RP could present the access token to the profile server to get the user's profile info which contains the real uid and email address.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -107,6 +110,16 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
     }
   }
 
+  // clients with server enforced PPID rotation are not allowed to
+  // request profile information to prevent the RP or service provider
+  // from learning the user's true identity.
+  if (
```
shane-tomlinson: I just realized this won't work for the secure proxy team. Removing. The secure proxy addon still needs to request profile information so they can display user information in the addon UI. To prevent the secure proxy provider from learning the user's true identity, they are requesting two access tokens, one with the "profile" scope, the other with a scope meant to access the proxy.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -942,6 +953,45 @@ describe('/v1', function() {
...
+          assert.equal(rotatedResult.statusCode, 200);
+          assert(rotatedResult.result.access_token);
+          assert.isNull(
+            validators.jwt.validate(rotatedResult.result.access_token).error
+          );
+
+          const notRotatedJWT = decodeJWT(notRotatedResult.result.access_token);
+          const rotatedJWT = decodeJWT(rotatedResult.result.access_token);
+
+          assert.notEqual(notRotatedJWT.claims.sub, rotatedJWT.claims.sub);
```
rfk: I would consider checking the known value of `notRotatedJWT.claims.sub` here, since you know it. (Otherwise maybe you got one of your booleans the wrong way round and it's accidentally `notRotatedJWT` that we rotated)

packages/fxa-auth-server/fxa-oauth-server/test/api.js
rfk: You could consider retrying once if the subs don't match, sleeping until `Date.now()` passes a rotation epoch.  Or perhaps explore mocking out `Date.now`, I don't know if we're ever done that in this repo yet...

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -189,17 +190,20 @@ function generateIdToken(grant, access) {
   return jwt.sign(claims);
 }
 
-exports.generateAccessToken = async function generateAccessToken(grant) {
+exports.generateAccessToken = async function generateAccessToken(
+  grant,
+  ppidSeed
+) {
```
rfk: Hmm...how will this interact with the desire from the webextension to fetch profile info for local display?

IIUC the webextension needs to be able to:

* Use `/token` to generate an access token with `profile` scope and the raw `uid`, which it can pass to the profile server.
* Use `/token` to generate a different access token with `secure-proxy` scope and the PPID, which it can pass to the third-party service.

But the current code will always generate tokens with the PPID, which won't be properly understood by profile-server.

I wonder if we actually do need to pass in the intended Resource Server here, both so that we can put a more specific value in the `aud` field and so that we know how to calculate the appropriate `sub`.

Alternately, you could enable PPIDs based on *scope* rather than client_id, so that anyone requesting a token with `secure-proxy` scope gets the PPID.

packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
```diff
@@ -0,0 +1,68 @@
...
+
+## The future
+
+Firefox accounts does not currently support [sector identifiers][#oidc-sector-identifier]. Sector identifiers allow multiple client_ids to receive the same PPID, this would be useful if the RP has applications on multiple platforms, e.g., an app on Android, iOS, and an addon in Firefox.
+
+[#oidc-spec]: https://openid.net/specs/oidc-core-1_0.html#Terminology
+[#oidc-id-token]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+[#oidc-sector-identifier]: https://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg
+[#oauth-jwt-access-token]: https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00
+[#hkdf]: https://tools.ietf.org/html/rfc5869
```
rfk: Thank you for this doc! I'm sorry to leave so many nits, it is a very good doc overall :-)

packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
```diff
@@ -0,0 +1,68 @@
...
+
+PPIDs are returned as the `sub` claim within an [OIDC ID Token][#oidc-id-token] or [OAuth JWT Access Token][#oauth-jwt-access-token].
+
+## Further enhancing privacy through PPID rotation
+
+User privacy can be further enhanced to prevent a long term profile from being built for a given user through PPID rotation. Firefox accounts supports two methods of PPID rotation, RP initiated, and server initiated.
+
+### RP initiated PPID rotation
+
+Any RP that receives JWT access tokens or OIDC ID tokens can initiate PPID rotation by specifying an additional parameter, `ppid_seed`, when requesting tokens from either the `/authorization` or `/token` endpoints on the OAuth server. `ppid_seed` must be an integer between 0 and 1024, and can be rotated at any time.
```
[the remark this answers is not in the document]
rfk: I'd like to see us add a section to this doc titles "abuse prevention" or similar, which would explain a little bit about why this value is limited to a small range. Maybe we can jam on such a thing a bit in person at our upcoming 1:1?

packages/fxa-auth-server/fxa-oauth-server/test/api.js
shane-tomlinson: Mocking out Date.now would be alright for unit tests, for integration tests that seems infeasible.

packages/fxa-auth-server/lib/oauthdb/utils.js
```diff
@@ -57,6 +57,10 @@ module.exports = {
         return error.staleAuthAt(err.authAt);
       case 120:
         return error.insufficientACRValues(err.foundValue);
+      case 121:
```
shane-tomlinson: Here's what I was talking about @rfk in #1784!

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
rfk: Ah, interesting! There's a lot hinging on this "`jti` maps to the raw token data in the db" thing...

packages/fxa-auth-server/fxa-oauth-server/config/test.json
```diff
@@ -100,17 +135,26 @@
...
+    "oldKeyFile": "../config/oldKey.json"
+  },
+  "ppid": {
+    "enabled": true,
+    "enabledClientIds": ["38a6b9b3a65a1874", "38a6b9b3a65a1875"],
+    "rotatingClientIds": ["38a6b9b3a65a1875"],
+    "rotationPeriodMS": 100
   },
   "allowHttpRedirects": true,
-  "authServerSecrets": ["meg@z0rd", "v0ltr0n"],
```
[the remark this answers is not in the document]
rfk: I don't feel strongly about either keeping or changing it, FWIW.

packages/fxa-auth-server/fxa-oauth-server/docs/api.md
```diff
@@ -330,16 +330,18 @@ back to the client. This code will be traded for a token at the
...
+- `acr_values`: (optional) A string-separated list of acr values that the token should have a claim for. Specifying `AAL2` will require the token to have an authentication assurance level >= 2 which corresponds to requiring 2FA.
 - `code_challenge_method`: Required if using [PKCE](pkce.md). Must be `S256`, no other value is accepted.
 - `code_challenge`: Required if using [PKCE](pkce.md). A minimum length of 43 characters and a maximum length of 128 characters string, encoded as `BASE64URL`.
-- `keys_jwe`: Optional. A JWE bundle to be returned to the client when it redeems the authorization code.
-- `acr_values`: Optional. A string-separated list of acr values that the token should have a claim for. Specifying `AAL2` will require the token to have an authentication assurance level >= 2 which corresponds to requiring 2FA.
+- `keys_jwe`: (optional) A JWE bundle to be returned to the client when it redeems the authorization code.
+- `ppid_seed`: (optional) Seed used in `sub` claim generation of JWT access tokens/ID tokens for clients with Pseudonymous Pairwise Identifiers (PPID) enabled.
```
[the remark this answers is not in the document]
shane-tomlinson: I haven't removed this yet, but have realized I put `ppid_seed` in the `GET /authorization` section instead of `POST /authorization`. Moving for now to think about whether removing support is the right thing to do.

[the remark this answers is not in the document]
shane-tomlinson: Removed support, it makes no sense to add it to the POST /authorization endpoint if we don't surface a way for RPs to provide a ppid_seed on the GET /authorization endpoint and have it propagated via the content server.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
shane-tomlinson: I wasn't able to assert what you suggested here because PPIDs were enabled, meaning my test had some terrible names. Because of that, I reworked the test and improved testing did point out I had a bug - I was feeding the wrong length to hkdf. 👍

packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
```diff
@@ -0,0 +1,68 @@
+# Pairwise Pseudonymous Identifiers
+
+A Pairwise Pseudonymous Identifier (PPID) is defined in the [OpenID Connect Spec][#oidc-spec] as:
+
+> Identifier that identifies the Entity to a Relying Party that cannot be correlated with the Entity's PPID at another Relying Party.
+
+Put another way, PPIDs provide a single user with distinct userids across distinct RPs, enhancing user privacy by preventing cross-site correlation based on userid.
+
+By default, Firefox accounts provides the same userid to all Mozilla-internal RPs to facilitate cross correlation between Mozilla services. However, as Mozilla expands its product offering to include white labeled 3rd party services, we want to ensure interactions with 3rd parties maintain Mozilla's principles regarding user privacy by providing 3rd parties with the minimal data necessary to successfully run and maintain their service.
```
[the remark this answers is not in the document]
shane-tomlinson: Trimmed it down, thanks for pointing that out.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
shane-tomlinson: Yeah, there is. It seemed like the quickest way to get JWT access tokens into prod while ensuring we don't regress on any current system behaviors.

packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
```diff
@@ -0,0 +1,68 @@
...
+PPIDs are not used by internal Mozilla properties and are not logged.
+
+## PPID generation
+
+PPIDs are generated using [HKDF][#hkdf] with the following inputs:
+
+- **km**: `client_id`.`userid_hex`.`ppid_seed||0`.`serverEnforcedRotationInput`
+- **info**: `oidc ppid sub`
+- **salt**: A secret configured by Ops
+- **len**: 8 bytes, which leads to a 16 hex character result
```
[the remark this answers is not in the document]
shane-tomlinson: You are right, and this was caused by me incorrectly assuming `unique.id` in configuration meant a userid. This is the bug that I pointed out that was found when I changed the tests at your requests. Instead of pulling the length from config, I now [use the length of the userid buffer](https://github.com/mozilla/fxa/blob/ed5d0305dc19343eff283a12798005e1bf7ebd3f/packages/fxa-auth-server/fxa-oauth-server/lib/jwt_sub.js#L50) that is passed in. Updating the docs to reflect that.

packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md
```diff
@@ -0,0 +1,68 @@
...
+
+PPIDs are returned as the `sub` claim within an [OIDC ID Token][#oidc-id-token] or [OAuth JWT Access Token][#oauth-jwt-access-token].
+
+## Further enhancing privacy through PPID rotation
+
+User privacy can be further enhanced by preventing long term profiles from being built for a given user through PPID rotation. Firefox accounts supports two methods of PPID rotation, RP initiated, and server initiated.
+
+### RP initiated PPID rotation
+
+Any PPID enabled RP that receives JWT access tokens or OIDC ID tokens can initiate PPID rotation by specifying an additional parameter, `ppid_seed`, when requesting tokens from `/token` endpoints on the OAuth server. `ppid_seed` must be an integer between 0 and 1024, and can be rotated at any time.
```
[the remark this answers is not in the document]
shane-tomlinson: [The latter currently](https://github.com/mozilla/fxa/blob/ed5d0305dc19343eff283a12798005e1bf7ebd3f/packages/fxa-auth-server/fxa-oauth-server/test/api.js#L4906). I could change it where if ppid_seed is passed in, then do the rotation, but that seemed like a possible footgun at this early stage.

packages/fxa-auth-server/fxa-oauth-server/test/jwt_sub.js
```diff
@@ -0,0 +1,181 @@
...
+          assert.lengthOf(result2, USER_ID_HEX.length);
+
+          assert.strictEqual(result1, result2);
+        } catch (e) {
+          return reject(e);
+        }
+        resolve();
+      };
+
+      setTimeout(callback, 10);
```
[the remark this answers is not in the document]
shane-tomlinson: Oh, I always forget about bluebird.

shane-tomlinson:
@rfk - I'm merging this w/o the doc updates, I'll open a follow on issue and PR with those, as well as the other items we discussed today in our 1:1.

Issue #1678, opened 2019-07-05
OIDC id_token at_hash incorrectly calculated


shane-tomlinson:
As pointed out in https://github.com/mozilla/fxa/pull/1612#discussion_r300210750, an OID ID token's [`at_hash`](https://github.com/mozilla/fxa/blob/7193cf829742652a99c884622806f08eda9dec9f/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js#L179) is incorrectly calculated. We are passing in the raw access token buffer [and converting it to ascii](https://github.com/mozilla/fxa/blob/7193cf829742652a99c884622806f08eda9dec9f/packages/fxa-auth-server/fxa-oauth-server/lib/util.js#L40), then calculating the hash. We should be calculating the hash based on the access token as seen by the user, meaning in hex form.

PR #1681, opened 2019-07-05
fix(openid): fix the openid at_hash calculation

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/1678

This simplifies the `at_hash` calculation for the access token by calculating the hash based on what the client sees.

When I originally read the spec, I suspect the ASCII part threw me off

> Its value is the base64url encoding of the left-most half of the hash of the octets of the ASCII representation of the access_token value

@shane-tomlinson @rfk Does this match what you were thinking?

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
@@ -157,7 +157,7 @@ module.exports.generateTokens = async function generateTokens(grant) {
   }
   // Maybe also generate an idToken?
   if (grant.scope && grant.scope.contains(SCOPE_OPENID)) {
-    result.id_token = await generateIdToken(grant, access);
+    result.id_token = await generateIdToken(grant, result.access_token);
   }
 
   amplitude('token.created', {
@@ -168,7 +168,7 @@ module.exports.generateTokens = async function generateTokens(grant) {
   return result;
 };
 
-function generateIdToken(grant, access) {
+function generateIdToken(grant, accessToken) {
   var now = Math.floor(Date.now() / 1000);
   var claims = {
     sub: hex(grant.userId),
@@ -176,7 +176,7 @@ function generateIdToken(grant, access) {
     //iss set in jwt.sign
     iat: now,
     exp: now + ID_TOKEN_EXPIRATION,
-    at_hash: util.generateTokenHash(access.token),
+    at_hash: util.generateTokenHash(accessToken),
   };
   if (grant.amr) {
     claims.amr = grant.amr;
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -2266,9 +2266,7 @@ describe('/v1', function() {
           assert.equal(claims.acr, ACR);
           assert.equal(claims['fxa-aal'], AAL);
 
-          const at_hash = util.generateTokenHash(
-            Buffer.from(res.result.access_token, 'hex')
-          );
+          const at_hash = util.generateTokenHash(res.result.access_token);
```
rfk: 👍  this change in the test makes it obvious that we're doing the right thing IMHO :-)

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -176,7 +176,7 @@ function generateIdToken(grant, access) {
     //iss set in jwt.sign
     iat: now,
     exp: now + ID_TOKEN_EXPIRATION,
-    at_hash: util.generateTokenHash(access.token),
+    at_hash: util.generateTokenHash(hex(access.token)),
```
rfk: I think you'll want to use ` access.jwt_token || access.token.toString('hex')` here to be consistent with how `result.access_token` is calculated above. Alternately, we could pass in `result.access_token` as an explicit argument to this function, so that the "this is the final form of the token as seen by the client" logic is encapsulated in a single place.

shane-tomlinson: While we are at it, maybe we have an opportunity to make the interface consistent so we don't need to do `access.jwt_token`, maybe we can unify the interface just pass in `access.token` and have the right thing happen everywhere? No need to call `hex`, no need to do the `||`?

vbudhram: Sold, updated 👍🏽

vbudhram:
No rush on this, made the recommended updates.

packages/fxa-auth-server/fxa-oauth-server/lib/util.js
```diff
@@ -29,15 +29,15 @@ const base64URLEncode = function base64URLEncode(buf) {
  * Generates a hash of the access token based on
  * http://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken
  *
- * This value is the hash of the ascii value of the access token, then the base64url
+ * This hash of the access token, then the base64url
```
rfk: nit: grammar here; maybe you mean "This is the hash of" or similar?

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -157,7 +157,7 @@ module.exports.generateTokens = async function generateTokens(grant) {
   }
   // Maybe also generate an idToken?
   if (grant.scope && grant.scope.contains(SCOPE_OPENID)) {
-    result.id_token = await generateIdToken(grant, access);
+    result.id_token = await generateIdToken(grant, result.access_token);
```
shane-tomlinson: Beautiful!

PR #1682, opened 2019-07-05
feat(payments): redirect to payment pages from fxa-content-server

ianb:
I tried to rebase #1310 and get it working, but somehow stuff kept getting in the way. In the end I just manually reapplied the changes from e19a727904ee41d2a2a63995434ccd010eea7921 (with a small change related to another commit that changed the signature of `createOAuthCode`)

- add /subscriptions/* path to server and client-side router

- SubscriptionsRedirectView that issues an access token and redirects to
  a path on fxa-payments-server

Issue #926

The test link for this locally is: http://127.0.0.1:3030/subscriptions/products/allDoneProProduct?plan=allDoneProWeekly


packages/fxa-content-server/app/scripts/views/settings/subscription_redirect.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+import SubscriptionView from './subscription';
+
+const View = SubscriptionView.extend({
```
shane-tomlinson: You could probably use class syntax here.

packages/fxa-content-server/server/lib/routes/get-frontend.js
```diff
@@ -73,6 +73,7 @@ module.exports = function() {
     'sms/sent',
     'sms/sent/why',
     'sms/why',
+    'subscriptions/products/[\\w_]+',
```
shane-tomlinson: Can you add a test to a sample product in both [the server side unit tests](https://github.com/mozilla/fxa/blob/39f7dd570e74aa6815912e50a48b92c85a6a6cb2/packages/fxa-content-server/tests/server/routes.js#L66) and [the functional tests](https://github.com/mozilla/fxa/blob/39f7dd570e74aa6815912e50a48b92c85a6a6cb2/packages/fxa-content-server/tests/functional/pages.js#L57).

shane-tomlinson: More on this, if I load up http://127.0.0.1:3030/subscriptions/producs/124124 (just a random plan ID), I am redirected to the payments server and then it's an infinite loop of "Plan not found". I click the X to close, and it refreshes to the same thing.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -0,0 +1,33 @@
...
+    if (options && options.config && options.config.subscriptions) {
+      this._subscriptionsConfig = options.config.subscriptions;
+    }
+  }
+
+  afterRender() {
+    const searchQuery = this.window.location.search;
+    const productId = this._currentPage.split('/').pop();
+    const redirectPath = `products/${productId}${searchQuery}`;
+    return PaymentServer.navigateToPaymentServer(
```
shane-tomlinson: Do you need to propagate any metrics parameters such as `flowBeginTime` or `flowId`?

packages/fxa-content-server/app/scripts/views/settings/subscription_redirect.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+import SubscriptionView from './subscription';
```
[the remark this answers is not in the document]
lmorchard: Hmm, actually yeah - I think this might be an older version of `subscriptions_product_redirect` that should have been deleted.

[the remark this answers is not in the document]
ianb: This is probably my mistake when I was moving code around.

packages/fxa-content-server/server/lib/routes/get-frontend.js
ianb: I couldn't add a functional test, or at least in that file, as it seems like all those tests specifically expect to land on a page with an element that matches `#stage header` – but this page lands on the payments server where that doesn't exist.

I opened #1730 as a followup about the error message. It's not specifically related to this PR, it just happens the plan-not-found error message uses `location.reload()` on the close button, which isn't sensible here.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
ianb: We don't have any metrics in the payment server at this time, and that is specifically not part of the MVP scope. I didn't see an issue for metrics, so I #1732 and a specific issue #1733 about passing through these parameters.

ianb:
@shane-tomlinson can you look at this again?

In addition to followups in the comments, I also opened #1723 as a followup (no redirect after signup, which I think will be a high priority).

packages/fxa-content-server/server/lib/routes/get-frontend.js
lmorchard: Yeah, this is basically a 404 condition - there's nowhere for the user to go if they land on a URL with an unknown product or plan ID.

And there isn't really a place for a sample product here, I don't think: The content server doesn't do any look-ups against the auth server for valid products in this implementation, it just redirects over to the payment server and lets that side handle the rest.

lmorchard: Oh, and I don't think there's a lot of validation we can do here - a product ID is an opaque `[a-zA-Z0-9_]` string from Stripe without any particular structure we can assert.

shane-tomlinson: Can we at assert that? I'll show you the Amplitude logs sometime, people really do shove all sorts of junk in to try to break us.

shane-tomlinson: Maybe at a minimum we should not allow `\\w_`, rather just the hex chars + `_`? Otherwise this seems fine, no additional validation needed, invalid strings should 404.

ianb:
@shane-tomlinson Oops, I was misreading the CircleCI failure as a build failure, and thought it was intermittent, I only now realized the test failures often end up on a second tab. That's fixed now, plus a little doc addition.

PR #1685, opened 2019-07-08
refactor(api): prefer async/await in POST /account/destroy route

philbooth:
It was pointed out in https://github.com/mozilla/fxa/pull/1671#discussion_r300750165 that the route handler for `POST /account/destroy` is an ugly mix of `.then` chains and `async`/`await`. This refactors it to use `async`/`await` across the board.

Opened as a separate PR because it makes a mess of the diff, so I'll rebase #1671 after this lands.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1370,113 +1370,98 @@ module.exports = (
       },
       handler: async function accountDestroy(request) {
         log.begin('Account.destroy', request);
-        const form = request.payload;
-        const authPW = form.authPW;
 
-        return customs
-          .check(request, form.email, 'accountDestroy')
-          .then(db.accountRecord.bind(db, form.email))
-          .then(emailRecord => {
-            return totpUtils.hasTotpToken(emailRecord).then(hasToken => {
-              const sessionToken = request.auth && request.auth.credentials;
+        const { authPW, email: emailAddress } = request.payload;
 
-              // Someone tried to delete an account with TOTP but did not specify a session.
-              // This shouldn't happen in practice, but just in case we throw unverified session.
-              if (!sessionToken && hasToken) {
-                throw error.unverifiedSession();
-              }
+        await customs.check(request, emailAddress, 'accountDestroy');
 
-              // If TOTP is enabled, ensure that the session has the correct assurance level before
-              // deleting account.
-              if (
-                sessionToken &&
-                hasToken &&
-                (sessionToken.tokenVerificationId ||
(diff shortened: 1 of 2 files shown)


philbooth:
[Clickable `?w=1` for you](https://github.com/mozilla/fxa/pull/1685/files).

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -2703,9 +2703,8 @@ describe('/account/destroy', () => {
         'db.emailRecord was called once'
       );
       let args = mockDB.accountRecord.args[0];
-      assert.equal(args.length, 2, 'db.emailRecord was passed two arguments');
-      assert.equal(args[0], email, 'first argument was email address');
-      assert.equal(args[1], true, 'second argument was customs.check result');
+      assert.equal(args.length, 1);
+      assert.equal(args[0], email);
```
philbooth: Side-effect from not passing the bound copy of `db.emailRecord` directly to `then`. The second argument is/was ignored.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1370,113 +1370,94 @@ module.exports = (
...
-              // deleting account.
-              if (
-                sessionToken &&
-                hasToken &&
-                (sessionToken.tokenVerificationId ||
-                  sessionToken.authenticatorAssuranceLevel <= 1)
-              ) {
-                throw error.unverifiedSession();
-              }
+        await customs.check(request, emailAddress, 'accountDestroy');
```
shane-tomlinson: What happens if this throws now that this is outside of the promise chain/try handler? It seems like the behavior will remain the same, the errors were re-thrown previously.

philbooth: All rejected promises returned by route handlers get wrapped by Hapi and then hit this condition in `lib/server.js`, where they go though `AppError.translate` before being serialised to the response:

https://github.com/mozilla/fxa/blob/94fe8d5917ed6c92b87a623936d38e75af4dffc6/packages/fxa-auth-server/lib/server.js#L282-L288

So throwing is fine here.

philbooth:
Whoops, turns out I broke the remote tests. Closing while I look into it.

lmorchard:
> Whoops, turns out I broke the remote tests. Closing while I look into it.

FWIW, this is why I ended up mixing async/await with promise code there. I broke everything while trying to convert to async/await and then fell back to "screw it, I'll just write the new thing as async/await" 😅

philbooth:
Fixed the async/await problem, I just needed to return something from the handler. 🤦‍♂

Issue #1690, opened 2019-07-08
Use displayName from credit card form


6a68:
In #1597, the payments server starts including the user's display name when creating a subscription on subhub.

It's not clear this will work correctly when the display name is an empty string.

Let's add some tests where the display name is an empty string, and be sure the auth server sends an empty string if there's no display name. If stuff breaks, we'll need to update the production code.

bbangert:
This should actually be the name from the credit card form as it goes to subhub for CC verification.

lmorchard:
Peeking at this and have a few hints on where to go with this:

* The credit card form is handled in [the `PaymentForm` component](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L62) and the billing name is accessible here.

* On successful Stripe payment token creation, [`PaymentForm` calls the `onPayment` handler passed in from props](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L66) - but it only supplies the token response from Stripe. We should probably tweak this from `.then(onPayment)` to something like `.then((tokenResponse: stripe.TokenResponse) => onPayment(tokenResponse, validator.getValues()))` so that the parent component can access form fields.

* The parent component for creating a subscription [defines its `onPayment` here](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Product/index.tsx#L100). If the previously suggested tweak is made in `PaymentForm`, then we can add take a second param here in `onPayment` to get `name` from the submitted form fields and use it as [`displayName` in the `createSubscription` call](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Product/index.tsx#L105).

6a68:
Thanks for the pointers, @lmorchard!

To your last question:

> Not sure if `updatePayment` API needs display name, though

I don't think so; it [looks like](https://github.com/mozilla/subhub/blob/cd5e0e7a227ddf2c79dad6bb87126765601e8f59/subhub/api/payments.py#L269) the subhub server gets the Customer object directly from stripe in the update payment case.

PR #1692, opened 2019-07-08
Revert "fix(subs): Add display_name field to subhub create subscription API"

6a68:
Reverts mozilla/fxa#1664.

Turns out the display name from the credit card input form, not the user's fxa display name, is what's needed. Following up in #1690.


6a68:
Closing. I'll just revert this when I fix the bug properly.

Issue #1307, opened 2019-05-29
Make trailhead flows RTL friendly and compatible


davismtl:
Shane called out that there will likely be work required to support RTL with the trailhead flows.

vbudhram:
@shane-tomlinson I went through the trailhead flows in RTL and they seem to look correct (minus the translations). Was there something else you wanted to check with it?

<img width="546" alt="Screen Shot 2019-06-25 at 1 58 23 PM" src="https://user-images.githubusercontent.com/1295288/60121666-989bc000-9751-11e9-8f47-5c8abf5a9cb8.png">
<img width="533" alt="Screen Shot 2019-06-25 at 1 55 55 PM" src="https://user-images.githubusercontent.com/1295288/60121667-989bc000-9751-11e9-8675-5094632bec66.png">
<img width="587" alt="Screen Shot 2019-06-25 at 1 53 53 PM" src="https://user-images.githubusercontent.com/1295288/60121668-989bc000-9751-11e9-932e-676361e292b1.png">
<img width="547" alt="Screen Shot 2019-06-25 at 1 53 39 PM" src="https://user-images.githubusercontent.com/1295288/60121669-989bc000-9751-11e9-91f2-b600b2c06c02.png">
<img width="789" alt="Screen Shot 2019-06-25 at 1 53 18 PM" src="https://user-images.githubusercontent.com/1295288/60121671-989bc000-9751-11e9-9ced-16477dd8964a.png">
<img width="873" alt="Screen Shot 2019-06-25 at 1 52 57 PM" src="https://user-images.githubusercontent.com/1295288/60121672-99345680-9751-11e9-8569-42a0afd3391c.png">
<img width="570" alt="Screen Shot 2019-06-25 at 1 51 03 PM" src="https://user-images.githubusercontent.com/1295288/60121674-99345680-9751-11e9-9be7-526b5d8a7e4f.png">

shane-tomlinson:
@vbudhram thanks for checking, can you also check mobile? How should we handle the 3 dot progression bar, should that be reversed in RTL too?

vbudhram:
@shane-tomlinson I made the screenshots in https://github.com/mozilla/fxa/pull/1703#issue-295496049 from mobile, which appear to look good. Here is confirmation screen on mobile

[...]

PR #1703, opened 2019-07-08
fix(css): update trailhead progress bar to support rtl

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/1307

This PR updates our trailhead progress bar to support RTL. I opted to create and modify our existing images to put the checkmarks in the right spot. There might be a better way to do this so open to suggestions!

Screen Shot 2019-07-08 at 4 53 24 PM
Screen Shot 2019-07-08 at 4 53 32 PM
Screen Shot 2019-07-08 at 4 53 39 PM

This was moved to Train 142, so no rush on reviews.

### DIFF
--- a/packages/fxa-content-server/app/images/trailhead/step-2-rtl.svg
+++ b/packages/fxa-content-server/app/images/trailhead/step-2-rtl.svg
@@ -1,0 +1,15 @@
+<svg width="141" height="18" xmlns="http://www.w3.org/2000/svg">
+    <g transform="translate(3 3)" fill="none" fill-rule="evenodd">
+        <circle stroke="#E0E0E6" stroke-width="3" cx="6" cy="6" r="7.5"/>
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@vbudhram could you use a CSS transform on the original image to [rotate the image 180 degrees](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)?

vbudhram:
@shane-tomlinson I tried a couple variations but they didn't look right

`transform: scaleX(-1);`

Screen Shot 2019-07-09 at 1 52 32 PM

`transform: rotate(180deg);`

Screen Shot 2019-07-09 at 1 54 51 PM

I might be able to apply the transformation but I would still have to create a new image.

shane-tomlinson:
Ah yeah, makes sense. Didn't think of that. r+

PR #1704, opened 2019-07-08
fix(subs): SubHub display_name should be CC name, not FxA name 

6a68:
OK, think I've got this setup correctly, and tests pass.

I'm not yet sure how much of the previous change (#1664) should be reverted. I think we need the change in the subhub API wrapper, but not elsewhere in the auth server?

~~Can't get payments tests running locally; leaning on CI again~~

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/index.tsx
@@ -97,12 +97,12 @@ export const Product = ({
     selectedPlan = productPlans[0];
   }
 
-  const onPayment = useCallback((tokenResponse: stripe.TokenResponse) => {
+  const onPayment = useCallback((tokenResponse: stripe.TokenResponse, name: string) => {
     if (tokenResponse && tokenResponse.token) {
       createSubscription(accessToken, {
         paymentToken: tokenResponse.token.id,
         planId: selectedPlan.plan_id,
-        displayName: profile.result ? profile.result.displayName : '',
+        displayName: name,
       });  
     } else {
       // This shouldn't happen with a successful createToken() call, but let's
--- a/packages/fxa-payments-server/README.md
+++ b/packages/fxa-payments-server/README.md
@@ -8,6 +8,10 @@ This project uses [Storybook](https://storybook.js.org/) to show each screen wit
 
 You can view the Storybook built from the most recent master at http://mozilla.github.io/fxa/fxa-payments-server/
 
+## Installation notes
+
+On Mac OS, `npm run test` may trigger an `EMFILE` error. In this case, to get tests running, you may need to `brew install watchman`. (If the watchman postinstall step fails, follow the instructions [here](https://stackoverflow.com/a/41320226) to change `/usr/local` ownership from root to your user account.)
+
 ## License
 
 MPL-2.0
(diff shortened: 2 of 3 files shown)


packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -97,20 +97,20 @@ export const Product = ({
...
+        displayName: name,
       });  
     } else {
       // This shouldn't happen with a successful createToken() call, but let's
       // display an error in case it does.
       const error: any = { message: 'No token response received from Stripe' };
       setCreateTokenError(error);
     }
-  }, [ accessToken, selectedPlan, createSubscription, setCreateTokenError ]);
+  }, [ accessToken, name, selectedPlan, createSubscription, setCreateTokenError ]);
```
lmorchard: You shouldn't need to add `name` to the list of closure-captured dependencies here, since it gets passed in as a param to the callback being memoized by the hook.

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -97,20 +97,20 @@ export const Product = ({
...
   }
 
-  const onPayment = useCallback((tokenResponse: stripe.TokenResponse) => {
+  const onPayment = useCallback((tokenResponse: stripe.TokenResponse, name: string) => {
     if (tokenResponse && tokenResponse.token) {
       createSubscription(accessToken, {
         paymentToken: tokenResponse.token.id,
         planId: selectedPlan.plan_id,
-        displayName: profile.result ? profile.result.displayName : '',
+        displayName: name,
```
lmorchard: For bonus points, if you name the param on line 100 as `displayName`, this line could just be `displayName,`

packages/fxa-payments-server/src/routes/Product/index.tsx
6a68: Thanks, yeah, didn't initially realize this was a list of implied globals

packages/fxa-payments-server/src/routes/Product/index.tsx
6a68: There's another displayName used elsewhere in payments-server, so I thought this might be clearer, but idk.

6a68:
Oh, just remembered there was a second spot where this function was called, which will now I think not supply the correct number of arguments to `onProfile`.

packages/fxa-payments-server/src/routes/Product/index.tsx
lmorchard: Yeah, not worth changing if it makes anything less clear :+1:

lmorchard:
The other customer for `onPayment` from `PaymentForm` is [`PaymentUpdateForm` over in the subscriptions management route](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx#L40). It *might* be okay to just ignore the second parameter there, since we're not using it.

6a68:
@lmorchard OK, maybe I could mark the second 'name' parameter as optional, and maybe also add some tests for the PaymentUpdateForm? How's that sound?

6a68:
Oh, I need to revert some of the auth-server changes before this lands. Let me add that...

PR #1705, opened 2019-07-08
chore(package-lock): Add 'optional' key to packages, automatically inserted by 'npm install' in npm 6.6+

LZoog:
While `npm install`ing with npm 6.6+, this `optional` key is added to the `package-lock` for "dependencies ONLY of the top level module or a transitive dependency of one".

See more info here in the first answer here: https://stackoverflow.com/questions/54411377/package-lock-json-file-package-with-optional-true

**Before merge, I'd like to confirm the version update of `fxa-dev-launcher` is correct, as I did not expect that to change.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3670,7 +3670,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
-          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8="
+          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -4213,7 +4232,7 @@
       }
     },
     "fxa-dev-launcher": {
-      "version": "github:vladikoff/fxa-dev-launcher#f5a01841fb6f3ca3fabf711219ed1cf94c8d63e6",
+      "version": "github:vladikoff/fxa-dev-launcher#f50f8111aa5470d637d1db96050556b5f1b12b2e",
```
lmorchard: Hmm, only thing here that made me raise my eyebrows was the change in commit hash here. But, [we don't pin this to a particular commit in `package.json`](https://github.com/mozilla/fxa/blob/master/package.json#L29), and we probably want [that change](https://github.com/vladikoff/fxa-dev-launcher/commit/f5a01841fb6f3ca3fabf711219ed1cf94c8d63e6) anyway...

LZoog: @lmorchard ah yeah, I had in my PR that I'd like to confirm that was correct since I wasn't expecting that to show. 🤔 From your last link it seems we do want that, though.

PR #1192, opened 2019-05-22
Revert "chore(package): add sjcl postinstall step to fxa-js-client"

vladikoff:
Reverts mozilla/fxa#977

PR #1707, opened 2019-07-09
fix(scripts): run grunt sjcl after npm ci for fxa-js-client

philbooth:
Previously I tried to fix this by adding `grunt sjcl` as a post-install step in #977, but that caused npm install to fail for the js client package and was backed out in #1192.

So, this is take 2, do it explicitly in scripting instead.

@mozilla/fxa-devs r?

### DIFF
--- a/_scripts/install_all.sh
+++ b/_scripts/install_all.sh
@@ -30,7 +30,7 @@ cd ..
 
 cd browserid-verifier; npm ci; cd ..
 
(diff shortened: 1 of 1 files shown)


philbooth:
> Didn't we do this at one time and had to remove it for some reason?

That is literally sentence #1 of the PR description :smile:

shane-tomlinson:
> That is literally sentence #1 of the PR description 😄

Aye.

PR #1710, opened 2019-07-09
refactor(change_password): Rendered Change Password using React

shane-tomlinson:


packages/fxa-content-server/app/scripts/views/base.js
```diff
@@ -381,9 +390,6 @@ var BaseView = Backbone.View.extend({
         err => {
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
             this.logError(AuthErrors.toError('SESSION_EXPIRED'));
-            // The redirectTo in .navigate() is lost if there's later navigations, so by saving it here
-            // we can get it later (e.g., in case of a signUp):
-            this.relier.set('redirectTo', this.window.location.href);
```
shane-tomlinson: Does this need to be removed?

packages/fxa-content-server/app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -40,14 +40,16 @@ export default {
...
   },
 
   hidePanel() {
     this._closePanelReturnToSettings();
   },
 
   isPanelOpen() {
-    return this.$('.settings-unit').hasClass('open');
+    return this.model.get('isPanelOpen');
+    // return this.$('.settings-unit').hasClass('open');
```
shane-tomlinson: Remove this please.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+import AuthErrors from '../../lib/auth-errors';
+import BackMixin from '../mixins/back-mixin';
+import Cocktail from 'cocktail';
+import FormView from '../form';
+import ExperimentMixin from '../mixins/experiment-mixin';
+import PasswordMixin from '../mixins/password-mixin';
+import PasswordStrengthMixin from '../mixins/password-strength-mixin';
+import ServiceMixin from '../mixins/service-mixin';
+import SettingsPanelMixin from '../mixins/settings-panel-mixin';
+// import Template from 'templates/settings/change_password.mustache';
```
shane-tomlinson: Remove this please

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+// eslint-disable-next-line
```
shane-tomlinson: It looks like React is used below, you might be able to remove this line.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+
+  this.state = {
+   hasFocused: false,
+   mail: props.account.get('email') || '',
+   newPass: '',
+   newVPass: '',
+   oldPass: '',
+  };
+ }
+ setOldPassword = event => {
```
shane-tomlinson: nit: space above this please.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+   });
+ };
+
+ setNewVPassword = event => {
+  this.setState(
+   {
+    newVPass: event.target.value,
+   });
+ };
+ showValidationErrorsEnd() {
```
shane-tomlinson: nit: space above this line please.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+  if (this.state.newPass && this.validateFormField('#new_password') && this.state.newVPass && this.state.newPass !== this.state.newVPass) {
```
shane-tomlinson: nit: can you break this line up?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
shane-tomlinson: nit: can you rename this just `showValidationErrors`?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+  }
+ }
+ componentDidUpdate() {
+  if (!this.state.hasFocused && this.props.isPanelOpen) {
+   this._input.focus();
+   this.setState({ hasFocused: true });
+  }
+  this.showValidationErrorsEnd();
+ }
+ componentWillUnmount() {
```
shane-tomlinson: What is this return value used for?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+ setFocus = () => {
+  this.setState(
+   {
+    newPass: '',
+    newVPass: '',
+    oldPass: '',
+    hasFocused: false,
+   },
+   () => {
+    this.componentDidUpdate();
```
shane-tomlinson: Will `componentDidUpdate` be called automatically due to the `setState`?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+   }
+  );
+ };
+
+ validateFormField(selector) {
+  try {
+   $(selector).validate();
```
shane-tomlinson: Can you leave a note here about `.validate()` calling the [jQuery plugin to do the validation](https://github.com/hritvi/fxa/blob/ae34698dee9370a411e88ab5e5361c965309ece6/packages/fxa-content-server/app/scripts/views/elements/jquery-plugin.js#L46) and that if the element is invalid an error will be thrown.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+ }
+ handleSubmit = event => {
+  event.preventDefault();
+  const areInputsValid =
+   this.validateFormField('#old_password') &&
+   this.validateFormField('#new_password') &&
+   this.validateFormField('#new_vpassword');
+
+  if (!areInputsValid) {
+   return false;
```
shane-tomlinson: To prevent the password from opening if an element is invalid, maybe try doing an `event.stopPropagation` before the `return false`.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+    () => {
+     this.setState({
+      newPass: '',
+      newVPass: '',
+      oldPass: '',
+      hasFocused: false,
+     });
+    },
+    err => {
+     console.log('error occured 1');
```
shane-tomlinson: Remove this console.log please.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+     if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+      return this.props.showValidationError('#old_password', err);
+     } else if (AuthErrors.is(err, 'PASSWORDS_MUST_BE_DIFFERENT')) {
+      return this.props.showValidationError('#new_password', err);
+     }
+     console.log('error occured');
```
shane-tomlinson: This one too.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+          Without it, the password manager saves the old_password as the username. */}
+     <input type="email" defaultValue={this.state.mail} className="hidden" />
+     <div className="input-row password-row">
+      <input
+       type="password"
+       className="password"
+       id="old_password"
+       placeholder={t('Old password')}
+       onChange={this.setOldPassword}
+       //  required pattern=".{8,}"
```
shane-tomlinson: Can you try putting this line back please?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+   this._input.focus();
+   this.setState({ hasFocused: true });
+  }
+  this.showValidationErrorsEnd();
+ }
+ componentWillUnmount() {
+  return this.state.newPass === this.state.newVPass;
+ }
+
+ setFocus = () => {
```
shane-tomlinson: Maybe rename this `resetInputValuesAndFocus` or similar to give a better indication of what it does.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+  if (!areInputsValid) {
+   return false;
+  }
+  if (this.state.newPass !== this.state.newVPass) {
+   const err = AuthErrors.toError('PASSWORDS_DO_NOT_MATCH');
+   this.props.showValidationError('#new_vpassword', err);
+  } else {
+   this.props.submit(this.state.oldPass, this.state.newPass).then(
+    () => {
+     this.setState({
```
shane-tomlinson: You might be able to call `this.setFocus()` here instead of duplicating the code.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+  return this.getSignedInAccount();
+ }
+
+ initialize(...args) {
+  super.initialize(...args);
+  this.listenTo(this.model, 'change', this.renderReactComponent);
+ }
+
+ renderReactComponent() {
+  return Promise.resolve(translator.fetch()).then(() => {
```
shane-tomlinson: You might be able to do `return translator.fetch().then(() => {`

packages/fxa-content-server/app/tests/spec/views/settings/change_password.jsx
```diff
@@ -0,0 +1,161 @@
...
+      });
+      sinon.stub(account, 'isSignedIn').callsFake(function () {
+        return Promise.resolve(true);
+      });
+
+      sinon.stub(view, 'getSignedInAccount').callsFake(function () {
+        return account;
+      });
+
+      return view.render()
```
shane-tomlinson: You might be able to get away with just rendering the react component and not the view here.

packages/fxa-content-server/app/tests/spec/views/settings/change_password.jsx
```diff
@@ -0,0 +1,161 @@
...
+    describe('render', function () {
+      it('renders properly', function () {
+        const oldPassword = document.querySelectorAll('#old_password');
+        const inputMail = document.querySelectorAll('input[type=email]');
+        assert.lengthOf(oldPassword, 1);
+        assert.lengthOf($('.reset-password'), 1);
+        assert.lengthOf($('#new_password'), 1);
+        assert.lengthOf($('#new_vpassword'), 1);
+        assert.lengthOf($('.input-help'), 1);
+        assert.isTrue($('.input-help-forgot-pw').length === 1);
```
shane-tomlinson: Can you use `assert.lengthOf` here instead of `isTrue`?

shane-tomlinson:
@hritvi - can you sign and re-push this commit too?

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
hritvi: Ya it would be called due to `setState` but because of that the `hasFocused` property is being set to True. So we are calling the `componentDidUpdate` to restore its value to false.

packages/fxa-content-server/app/scripts/models/password_strength/password_strength_balloon.js
```diff
@@ -50,6 +50,7 @@ export default class PasswordStrengthBalloonModel extends Model {
     super(attrsWithDefaults, config);
 
     this.on('change:password', () => this.set('hasUserTakenAction', true));
+    // this.on('change:password .cancel', () => this.set('hasUserTakenAction', false));
```
shane-tomlinson: Let's take this out too.

packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
```diff
@@ -0,0 +1,319 @@
...
+ /* It causes form submission multiple times so to avoid
+  * that we have overwritten the defination of onFormSubmit */
+ onFormSubmit() {}
+
+ submit(oldPassword, newPassword) {
+  console.log('callling submit');
+  var account = this.getSignedInAccount();
+
+  this.hideError();
+  console.log(oldPassword, '  ', newPassword);
```
shane-tomlinson: Remove the `console.log`s from this function please.

hritvi:
@shane-tomlinson I have updated the PR.

Issue #1415, opened 2017-11-16
Retire fxa-basket-proxy, let basket handle everything itself [bz1417298]


bug-mirror:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1417298

As of the deploy of train-100 and the latest update of basket, we should have everything we need in order to retire fxa-basket-proxy and let basket handle all that functionality natively.  Basket has grown an SQS event handler, and its API now accepts FxA OAuth token directly.

This bug is for figuring out and executing on a plan to make that tha transition a reality.

IIUC, we will need to:

* Update fxa-content-server config, so that it talks directly to basket's API via OAuth, rather than to our basket proxy.  In theory this involved just changing some URLs.
* Pick a cutover date for fxa-basket-proxy to stop handling SQS events, and for basket to start.  The config on our side is in [1] but this'll require some careful QA to ensure that we don't get e.g. any weird timezone effects.
* Get the above deployed to production, wait for the cutover date to pass, and assure ourselves that things are working properly.
* Tear down fxa-basket-proxy, remove its SQS queues, archive the repo, celebrate :-)

Shane, can you think of any other things we would need to do here on the FxA side?

Let's try this all our in stage once FxA train-100 is out the door.

[1] https://github.com/mozilla/fxa-basket-proxy/pull/48

shane-tomlinson:
As of now, there should be no more need for the basket proxy.

shane-tomlinson:
From @jrgm in slack:

> @stomlinson there are a tiny number of /lookup-user requests occurring still

I'm assuming these are users who have had settings open for a while. Maybe wait a train and see if that number has died down even further.

shane-tomlinson:
Things left to do:

* [ ] Remove the basket proxy servers from fxa-dev
* [ ] Remove the basket proxy servers from prod/stage
* [ ] Remove the basket proxy server directory from packages

PR #1713, opened 2019-07-09
chore(deps): Remove basket proxy from the server list.

shane-tomlinson:
The basket proxy is no longer needed, Basket now does a normal OAuth flow instead.

issue #1415


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #1714, opened 2019-07-09
fix(scripts): fix splitting of commit messages in release script

### LINKED ISSUE #1712: Release script writes borked changelogs
Looks like #1601 caused bustage that doesn't materialise for patch releases, so I only just noticed it. This is blocking the tag for train 141.

philbooth:
Fixes #1712.

Sets `IFS` so the commit history is split on newline character rather than any whitespace. It's not portable so probably won't work on WSL and maybe other places too. But it's good enough to unblock the release for now.

@mozilla/fxa-devs r?

/cc @vbudhram

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 set -e
+IFS=$'\n'
 
 # This script will automatically update the changelogs and bump the version
 # strings for each of the "main" FxA packages in this tree. It assumes some


rfk:
> It's not portable so probably won't work on WSL 

Oh *that's* why there was weird stuff happening when I tried to cut a 141 point-release 😆 
FWIW explictlying using `bash ./release.sh patch` made it work for me under WSL.

philbooth:
> Oh *that's* why there was weird stuff happening when I tried to cut a 141 point-release

Yeah, sorry @rfk, it was remiss of me not to broadcast that more widely. Let's just change the shebang to bash and have done with it.

Issue #1078, opened 2019-05-10
Subscription management button should only appear when user has at least one active subscription


lmorchard:
The [subscription managment button](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/views/settings/subscription.js) will always be shown, once we enable the subscriptions feature. This button should be hidden unless the user has at least one subscription active.

Some comments in the wireframe about this:

https://mozilla.invisionapp.com/share/K2RBDY3VU9X#/screens/355610597/comments/114028740

clouserw:
Do we want to hide this button even if they used to have a subscription and it has lapsed (like, if they had a payment problem)?  /cc @johngruen

ianb:
Note there's a language check for the button currently, and that check should be removed (it's up to product pages to filter their audience appropriately).

clouserw:
If a subscription has lapsed, don't show the button.  We can make that more graceful after the MVP

Issue #797, opened 2019-04-10
[Support] Create support form


ianb:
We want a support form, which will ultimately be hosted on `https://accounts.firefox.com/premium-support` (we want to be clear this is not a general FxA or Firefox support forum).

The page should require authentication, and should require an active subscription (Q: are we sure about the active subscription?). External pages may link directly to the support page.

Input fields TBD.

shane-tomlinson:
Will this page be a panel under /settings or it's own separate page? 

If under settings, let's make the URL consistent with other settings panels and use `https://accounts.firefox.com/settings/premium-support`.

clouserw:
Support wireframe is on page 11 here: https://mozilla.invisionapp.com/share/K2RBDY3VU9X#/screens/356896706

lmorchard:
We just touched on this in a review of PR #961: Should this support form actually live in fxa-payments-server rather than fxa-content-server?

shane-tomlinson:
If this is within the content server, I can pick this up once I finish #1098.

lmorchard:
Re: living on the content server - just to summarize [my comment](https://github.com/mozilla/fxa/issues/1100#issuecomment-492712305) from #1100 here: 

Probably the only wrinkle that suggests the support form should live on payments-server is that it's a) linked to from subscription management and b) the form returns to subscription management after submission.

juliastarkov:
Ready for production support mockups can be viewed here (screens 6 - 10): https://mozilla.invisionapp.com/share/5BS1I7KANWP#/363599381_Support

juliastarkov:
Ready for production Zeplin mockups can be viewed here: https://zpl.io/2pMEYRN

chenba:
@juliastarkov I haven't used Zeplin before.  Do I need access?  If so, invite bchen@mozilla.com or Zeplin username bchenistaken.

chenba:
Re: content server vs. payment server

The code on the content server simply navigate the user to the payment server for subscription management.  (Because of #1637, I didn't realize this until I look at the code for the subscriptions panel code in content server.)

Given that, and what the mockups are showing, I _think_ this form should be on the payment server?  It might be confusing to users if we navigate them to the payment server for subscription management, and bounce them back for the support form, and send them back to the payment server...?

In other words, I just understood and regurgitated @lmorchard's comment https://github.com/mozilla/fxa/issues/797#issuecomment-492713468

chenba:
More questions (for... @johngruen?):

[...]

PR #1715, opened 2019-07-09
feat(fxa-client): add active subscriptions and support ticket endpoints

chenba:
fxa-content-server needs to know if a user has any active subscriptions
and to able to submit a support ticket.

Issues #797 and #1078

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-js-client/client/FxAccountClient.js
+++ b/packages/fxa-js-client/client/FxAccountClient.js
@@ -2472,6 +2472,69 @@ define([
   };
 
   /**
+   * Get a user's list of active subscriptions.
+   *
+   * @param {String} token A token from the OAuth server.
+   * @returns {Promise} A promise that will be fulfilled with a list of active
+   * subscriptions.
+   */
+  FxAccountClient.prototype.getActiveSubscriptions = function(token) {
+    var self = this;
+
+    return Promise.resolve().then(function() {
+      required(token, 'token');
+      const requestOptions = {
+        headers: {
+          Authorization: `Bearer ${token}`,
+        },
+      };
+      return self.request.send(
+        '/oauth/subscriptions/active',
+        'GET',
+        null,
+        null,
+        requestOptions
+      );
+    });
+  };
+
+  /**
+   * Submit a support ticket.
+   *
+   * @param {String} authorizationHeader A token from the OAuth server.
+   * @param {Object} [supportTicket={}]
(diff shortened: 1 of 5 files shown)


packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -2471,6 +2471,75 @@ define([
...
+  /**
+   * Get a user's list of active subscriptions.
+   *
+   * @param {String} authorizationHeader
+   * The authorization header for the request.
+   * It should be in the form of `Bearer ${SOME_TOKEN}`.
+   * @returns {Promise} A promise that will be fulfilled with a list of active
+   * subscriptions.
+   */
+  FxAccountClient.prototype.getActiveSubscriptions = function(
```
philbooth: Splendid, this is also useful for #1071.

packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -2471,6 +2471,75 @@ define([
       });
   };
 
+  /**
+   * Get a user's list of active subscriptions.
+   *
+   * @param {String} authorizationHeader
+   * The authorization header for the request.
+   * It should be in the form of `Bearer ${SOME_TOKEN}`.
```
philbooth: I think it would be a nicer API if the `Bearer ${refreshToken}` concatenation happened inside this lib. Then callers would only have to worry about passing `refreshToken` into these methods, so less work for them and it's also consistent with how the other token types are passed in.

chenba: Made the suggested change.

packages/fxa-js-client/tests/lib/subscriptions.js
```diff
@@ -0,0 +1,102 @@
...
+          .then(assert.notOk, function(error) {
+            assert.include(error.message, 'Missing supportTicket');
+          });
+      });
+      test('#createSupportTicket', function() {
+        return accountHelper
+          .newVerifiedAccount()
+          .then(function(account) {
+            return respond(
+              client.createSupportTicket('Bearer redpandas', {
```
philbooth: Not that it makes a difference, but maybe just for clarity of someone reading these tests remove the `Bearer ` prefix here?

chenba: D'oh!  I meant to remove it.  Thanks for catching that.

PR #1725, opened 2019-07-10
Merge v1.141.1 to master

rfk:
This commit was originally cherry-picked from master, so there shouldn't be anything in the merge apart from the changelog update...


rfk:
Humph, sorry, the diff here looks gigantic but I promise this is already in master thanks to #1262...

PR #1728, opened 2019-07-10
Add payment error messages

### LINKED ISSUE #1589: [fxa-payments-server] Implement more specific & localizable error messages for payment pages
In this commit, we added error dialog messages on payment failure: https://github.com/mozilla/fxa/commit/93eddbaac81250ca6f778dfaae139e7452d35155

The messages are somewhat generic and/or display en-US text messages conveyed from Stripe. This may present problems for l10n down the road.

We have this spreadsheet of error message copy (which I believe is approved by legal):

https://docs.google.com/spreadsheets/d/1Kl2v8hJTntQvtGUTJcvWpSZQLIVvoYgnRboT2Lziu_w/edit?disco=AAAADAewemI&usp_dm=false&ts=5cdc2bc0#gid=0

We can implement some of these, based on the error code type relayed via subhub. There may need to be additional work to ensure that error type is carried through from subhub calls.

Some may not be feasible to implement without more detailed info asked for in this subhub issue: https://github.com/mozilla/subhub/issues/98

However, as noted in that issue, the more detailed information might be a security concern

meandavejustice:
- fixes #1589

error-dialog

### DIFF
--- a/packages/fxa-payments-server/src/lib/errors.ts
+++ b/packages/fxa-payments-server/src/lib/errors.ts
@@ -1,5 +1,5 @@
 // ref: fxa-auth-server/lib/error.js
-export const AuthServerErrno = {
+const AuthServerErrno = {
   UNKNOWN_SUBSCRIPTION_CUSTOMER: 176,
   UNKNOWN_SUBSCRIPTION: 177,
   UNKNOWN_SUBSCRIPTION_PLAN: 178,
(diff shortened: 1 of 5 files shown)


packages/fxa-payments-server/src/lib/errors.ts
```diff
@@ -1,9 +1,109 @@
 // ref: fxa-auth-server/lib/error.js
-export const AuthServerErrno = {
```
lmorchard: I usually prefer leaving the exports in-line, rather than listing them all at the end. 

That said... there are a lot of intermediate variables here that *aren't* exported. Listing them at the end seems to keep them from getting lost in the rest of the file. So, :+1:

lmorchard:
Oh, but it looks like you do need to [set up GPG commit signing](https://micropipes.com/blog/2016/08/31/signing-your-commits-on-github-with-a-gpg-key/) here since we turned on enforcement of that :/

lmorchard:
Oh yeah, would probably be nice to add some stories to maybe both the Product & Subscription routes to display some of these new errors. But we could spin that off as a follow-up

meandavejustice:
I've added some basic tests for `errors.ts`, updated the commit message to follow our guidelines, and enabled gpg signing but I'm having trouble remember the password for my gpg key 😭. I'm poking around cus I'm pretty sure I wrote it down somewhere this time.

@lmorchard 
> Oh yeah, would probably be nice to add some stories to maybe both the Product & Subscription routes to display & demo some of these new errors. But we could spin that off as a follow-up

I haven't figured out a good way to trigger this error, while developing I was manually adding some characters to the api key to invalidate it. 🤔

lmorchard:
> I haven't figured out a good way to trigger this error, while developing I was manually adding some characters to the api key to invalidate it. 

Oh, actually, for storybook purposes, it might be enough to set up some React views with dialogs just to show how the messages render & appear in the layout. Doesn't have to be a part of this PR though

PR #1762, opened 2019-07-12
fix(events): compile typescript dist in docker build image

jrgm:
Might want to refine this later so we don't have all devDependencies in the docker image, and I'll leave it to @bbangert to choose the right pattern, but this will work for now.

r? - @bbangert


jrgm:
> This looks fine, but it will still recompile it on start as the npm start command does a build.

Ah. I configured to start with `node ./dist/bin/worker.js`.  Seems better to me to not depend on npm to recompile at startup when scaling, and instead run pre-compiled code in the image filesystem.

Issue #1696, opened 2019-07-08
Improve security on payment server access token generation using intermediate codes


lmorchard:
For issue #881, we implemented auth token generation in content server to enable API calls in payment server pages. This works, but is not ideal for many reasons.

An incremental improvement to the security situation would be to supply the payment server web app with a code it can use to fetch an access token. That way, we can enforce various constraints on the transition from content server to payment server - e.g. same IP, etc.

/cc @shane-tomlinson for more erudite details on this.

6a68:
@shane-tomlinson Let's find a time to chat about the right way to implement this. Could this be a spot to use a JWT?

shane-tomlinson:
> Could this be a spot to use a JWT?

I don't think so, JWTs are currently only used for access tokens. Here I'd love to use the full [prompt=none flow](https://github.com/mozilla/fxa/issues/640), but that's not ready. An intermediate step that's a bit better than what we have now is to redirect to the payments page with an authorization code instead of an access token. At least authorization codes are single use.

shane-tomlinson:
> Could this be a spot to use a JWT?

Though, there is [this IETF spec for using JWTs as authorization grants](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-jwt-bearer#section-2.1).

lmorchard:
A question that occurred to me over in #1694: Short of skipping straight to a full OAuth RP implementation (#1695), is there a way we can implement this issue such that the payment pages can refresh the token?

shane-tomlinson:
> A question that occurred to me over in #1694: Short of skipping straight to a full OAuth RP implementation (#1695), is there a way we can implement this issue such that the payment pages can refresh the token?

When generating the authorization code, request "access_type=offline". When trading the code for an access token at `/token`, a refresh token will also be returned. See https://github.com/mozilla/fxa/blob/9e76653ab3bc62e35ebb2487b925ebb9b4cf01cf/packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-authorization-code.js#L21. The refresh token can then be used to get new access tokens: https://github.com/mozilla/fxa/blob/9e76653ab3bc62e35ebb2487b925ebb9b4cf01cf/packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-refresh-token.js#L16

lmorchard:
Ah, okay - I was thinking we wouldn't be using something bespoke and not a more normal authorization code. That probably makes a lot of things more sane.

[...]

Issue #1694, opened 2019-07-08
Handle payment server token expiration more gracefully


lmorchard:
Follow-up to #881

Payment server gets a token with a short expiration. That means the user could more likely be in the middle of something when it expires. 

Currently, we sometimes bounce back to the settings page on the content server when this happens. Other times, we just report errors. This should be smoothed out into something with better UX - automatic token refreshes? something else?

lmorchard:
This might be blocked by #1696 if we want a resolution here to be that the payment app silently requests to refresh its short-lived token and retries the operation. (i.e. The user attempts to submit payment to subscribe, but they took longer than the token lifetime and so the request fails unless we get another token)

In fact, we might want some product / UX thinking on this one, too...

lmorchard:
Adding needs:product and needs:ux labels. Here's a rephrasing of the problem: 

When the user arrives at either a product payment page or the subscription management page, they're sent in with an access token. That access token allows the payment pages to make API calls. The API calls accomplish whatever the user's there to do - e.g. pay for a new subscription, update their current card, cancel subscriptions, etc.

For security reasons, that access token has a short life time. After that life time expires, it's no longer usable to make API calls. In terms of UX & product, that means if the user takes longer than the access token life time to complete a task, that task will fail with an error.

So, what can we do about this in terms of UX & product? Depending on how #1696 works out, maybe we can just request a new access token? Otherwise, we might need some UX finesse to walk the user through starting over with their task, which seems less than happymaking

lmorchard:
Got a positive answer on refreshing the token in #1696 - so I think this de-escalates from needing any product or UX decisions. I think we can just push refreshing the token into the invisible background and not have to surface it as a concern to the user.

lmorchard:
I think the main takeaway from triage today was to bump access token expiration time up to 20-30 minutes until we have moved over to being a proper OAuth RP with refresh token?

Would it make any sense at all to set up a `setTimeout` to periodically ping `/v1/verify` on auth-server to detect when the token has expired, then display an error message and redirect back to FxA settings? This might be an abrupt end to a payment flow, though.

6a68:
Thanks for clarifying.

Issue #640, opened 2018-04-25
Support OIDC prompt=none parameter?


rfk:
This came up in a recent meeting with @gdestuynder about the IAM integration.

The OpenID Connect spec defines some special behavior when the user loads the oauth flow with `?prompt=none`; from [1]:

```
prompt
    OPTIONAL. Space delimited, case sensitive list of ASCII string values that specifies whether the
    Authorization Server prompts the End-User for reauthentication and consent. The defined values are:

        none
            The Authorization Server MUST NOT display any authentication or consent user interface pages.
            An error is returned if an End-User is not already authenticated or the Client does not have
            pre-configured consent for the requested Claims or does not fulfill other conditions for 
            processing the request. The error code will typically be login_required, interaction_required, or
            another code defined in Section 3.1.2.6. This can be used as a method to check for existing
            authentication and/or consent.
```

OIDC clients (such as the Auth0 stack used for IAM integration) use this as a way to check for freshness of a user's login session - they make a background request to the OAuth authorization endpoint with `?prompt=none`, refresh their auth state if it succeeds, and sign the user out if it fails.  It basically allows the relier to ask "if the user did an OAuth flow right now, would they have to log in?"

The IAM integration stack tries to use `prompt=none` in this manner when talking to FxA, but we don't handle it in a spec compliant way; instead we silently ignore it.

One wrinkle: as discussed in https://github.com/mozilla/fxa-content-server/issues/5916, we will *always* require some user interaction when signing in to an OAuth relier, so the correct way for us to handle `prompt=none` would be to immediately redirect back to the relier with an error response.  That might actually make it simpler to implement on our side :-)

@gdestuynder is working on a fix for this in Auth0 config, to prevent it from trying to use this unsupported parameter when talking to FxA.  But that's probably going to be at least a little gross, and it's a feature that's generally useful, so we should consider whether it's worth adding on our side for a better experience.

[1] http://openid.net/specs/openid-connect-core-1_0.html

shane-tomlinson:
Ref https://github.com/mozilla/fxa/issues/1979 to support RP initiated logout.

gdestuynder:
@shane-tomlinson is this a feature we can/should be using with SSO?

The comment in https://github.com/mozilla/fxa/commit/15d0cbf30764323913ace0b575671feb0a3b4f51 is a little scary :)

[...]

PR #1772, opened 2019-07-12
 feat(payments): Use PKCE to access the payments page. 

shane-tomlinson:
This builds on #1682 

This is a bit strange, but here we go.

We aren't doing a traditional OAuth flow here where we start at the RP (payments page)
to set up `state` and PKCE parameters. To do so cleanly would require support for
`prompt=none`, which we do not yet have. See #640, PR #1150.

Instead, we secure the flow and prevent CSRF by relying on cookies.
We set up state and the PKCE parameters on the content server, then send
the params to the content-server backend to store in an HTTP only cookie on
accounts.firefox.com scoped to /payments-pkce.

Still on the content server, we then generate an OAuth code using the
generated state/PKCE params, and redirect to the payments page.

The payments page gets the OAuth code/state out of the URL.
It then fetches the state/PKCE params from the content-server
by making a GET request to /payments-pkce. The content-server
ensures the request comes from the payments domain, reads
the cookie, and then returns the PKCE/state info.

The payments server now has all the stuff it needs to trade
the OAuth code for an access token. It makes the call to the
OAuth server's /token endpoint with the code from the URL,
and the state and PKCE parameters returned from the backend.

A refresh token is also returned enabling fetches of new
access tokens if the current one expires.

issue #1696
issue #1694

cc @6a68 - this is what I was thinking. The 2nd commit is the important one. It's not actually all that complex, just a bit strange.


packages/fxa-payments-server/src/index.tsx
```diff
@@ -69,46 +68,95 @@ const parseParams = (params: string): ParsedParams => params
...
+  try {
+    const pkceResult = await fetch(
+      `${config.servers.content.url}/payments-pkce`,
+      {
+        credentials: 'include',
+        headers: { 'Content-Type': 'application/json' },
+        method: 'get',
+      }
+    )
+    if (pkceResult.status !== 200) {
```
6a68: If this is a standard `fetch`, do we also need to verify that `pkceResult.ok` is true?

packages/fxa-content-server/app/scripts/lib/payment-server.js
```diff
@@ -0,0 +1,120 @@
...
+
+function createRandomString(length) {
+  if (length <= 0) {
+    return '';
+  }
+  var _state = '';
+  var possible =
+    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~';
+  for (var i = 0; i < length; i++) {
+    _state += possible.charAt(Math.floor(Math.random() * possible.length));
```
6a68: Is `Math.random` random enough?

packages/fxa-content-server/server/lib/routes/post-payments-pkce.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    process(req, res) {
+      const cookieData = JSON.stringify({
+        // eslint-disable-next-line camelcase
+        code_verifier: req.body.code_verifier,
+        state: req.body.state,
+      });
+      console.log('cookieData', cookieData);
+      res.cookie('_pkce', cookieData, {
+        httpOnly: true,
```
6a68: The security checklist (see #1128) suggests also setting the `secure` flag on cookies

PR #1773, opened 2019-07-12
feat(subscriptions): add active subscriptions and support ticket endpoint access

chenba:
`fxa-js-client` work was done in #1715.

Issues #797, #1078, #1071


chenba:
cc @lmorchard @LZoog

PR #1776, opened 2019-07-14
Added securityEvents route to js-client

farhan787:
@vbudhram initial commit for adding this route, since the `auth-server` PR is not merged yet I created a new branch from that one and checkout the last commit.

### DIFF
--- a/packages/fxa-js-client/client/FxAccountClient.js
+++ b/packages/fxa-js-client/client/FxAccountClient.js
@@ -1633,7 +1633,7 @@ define([
    * @param {String} sessionToken sessionToken obtained from signIn
    * @return {Promise} A promise that will be fulfilled with JSON `xhr.responseText` of the request
    */
-  FxAccountClient.prototype.sessions = function(sessionToken) {
+  FxAccountClient.prototype.securityEvents = function(sessionToken) {
     var request = this.request;
 
     return Promise.resolve()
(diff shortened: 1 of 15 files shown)


packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -1624,6 +1624,46 @@ define([
...
       });
   };
+  /**
+   * Get a list of user's security events
+   *
+   * @method securityEvents
+   * @param {String} sessionToken sessionToken obtained from signIn
+   * @return {Promise} A promise that will be fulfilled with JSON `xhr.responseText` of the request
+   */
+  FxAccountClient.prototype.sessions = function(sessionToken) {
```
vbudhram: This needs to be `securityEvents`

packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -1624,6 +1624,46 @@ define([
...
+      });
+  };
+
+  /**
+   * Delete user's security events
+   *
+   * @method securityEvents
+   * @param {String} sessionToken sessionToken obtained from signIn
+   */
+  FxAccountClient.prototype.sessions = function(sessionToken) {
```
vbudhram: This should be `deleteSecurityEvents`.

packages/fxa-js-client/tests/lib/securityEvents.js
```diff
@@ -0,0 +1,68 @@
...
+        return accountHelper
+          .newVerifiedAccount()
+          .then(account => {
+            return respond(
+              client.securityEvents(account.signIn.sessionToken),
+              RequestMocks.securityEvents
+            );
+          })
+          .then(res => {
+            assert.ok(res, 'got response');
```
vbudhram: You should be able to assert the events from the auth-server not just getting a response.

vbudhram: For both routes, you need to check that the correct values were set on the request, ex path name. See this example, https://github.com/mozilla/fxa/blob/b7adf5901737703935acd15fe1c6f1a5b0d571b3/packages/fxa-js-client/tests/lib/recoveryCodes.js#L87.

packages/fxa-js-client/tests/lib/securityEvents.js
```diff
@@ -10,58 +10,78 @@ define([
...
+            xhrOpen.args[0][1],
+            '/securityEvents',
+            'path is correct'
+          );
+          assert.ok(res, 'got response');
+          assert.equal(res.length, 2);
+
+          assert.equal(res[0].name, 'account.create');
+          assert.equal(res[0].verified, true);
+          // isBelow is not a function, how to test createdAt field
```
farhan787: @vbudhram what can I use here instead of `isBelow` ?

packages/fxa-js-client/tests/lib/securityEvents.js
```diff
@@ -10,58 +10,78 @@ define([
...
+          // assert.isBelow(res[0].createdAt, new Date().getTime());
```
farhan787: @vbudhram what can I use here instead of isBelow ?

packages/fxa-js-client/tests/lib/securityEvents.js
farhan787: Thanks for correcting me !!

packages/fxa-js-client/tests/lib/securityEvents.js
vbudhram: Yea, I would use `isBelow` here.

farhan787: `isBelow` isn't a function inside `assert` and can't use `equal` there.

vbudhram:
Hey @farhan787, https://github.com/mozilla/fxa/pull/1661 has merged, you should be able to rebase this PR.

farhan787:
Continue with [this PR](https://github.com/mozilla/fxa/pull/1848)

PR #1780, opened 2019-07-15
Create link for maildev local server

meandavejustice:


vbudhram:
We could probably tag this as a `chore(docs): ` with commit message.

meandavejustice:
@vbudhram Updated commit message 👍

PR #1785, opened 2019-07-15
fix(oauth): Ignore unknown refresh_token_id when destroying attached client.

### LINKED ISSUE #1784: Map oauth-server errno 122 to an auth-server error
Ref https://sentry.prod.mozaws.net/operations/auth-prod/issues/6052963/events/31641096/

The auth-server's new `oauthdb.revokeAuthorizedClient` method calls `/v1/authorized-clients/destroy` on the oauth-server, which can return an errno 122 `Unknown Token` if you try to destroy a refresh token that doesn't exist.  We need to map that back to an auth-server error code as part of the oauthdb wrapper.

rfk:
Fixes #1784.

If we get an "unknown token" error when trying to destroy an authorized client record, it almost certainly indicates some sort of race in deleting the refresh token. There is nothing useful we can tell the user, and the account is in the state they wanted anyway, so just ignore it.

(In practice, I think we'll see these when we've got a device record in the auth-server db, but the corresponding refresh token has been deleted from the oauth-server db).

### DIFF
--- a/packages/fxa-auth-server/lib/error.js
+++ b/packages/fxa-auth-server/lib/error.js
@@ -91,6 +91,7 @@ const ERRNO = {
   REJECTED_SUBSCRIPTION_PAYMENT_TOKEN: 179,
   SUBSCRIPTION_ALREADY_CANCELLED: 180,
   REJECTED_CUSTOMER_UPDATE: 181,
+  REFRESH_TOKEN_UNKNOWN: 182,
 
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/error.js
```diff
@@ -1189,6 +1190,15 @@ AppError.insufficientACRValues = foundValue => {
   );
 };
 
+AppError.unknownRefreshToken = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.REFRESH_TOKEN_UNKNOWN,
+    message: 'Unknown refresh token',
```
rfk: I haven't added this to the errno list in fxa-content-server, because we swallow it rather than propagating it out to the user.

philbooth:
@rfk, just checking, do you want to rebase this PR against `train-141` for a point release to save having to cherry-pick?

rfk:
Hrm, good point - to be honest I hadn't considered a point-release because the volume of these errors didn't seem too high in prod, but it's probably worthwhile. Thanks.

Issue #1774, opened 2019-07-12
[fxa-event-broker] Finish productizing


bbangert:
A few things that cropped up during deployment we should resolve:
- [x] Create a `__lbheartbeat__` route.
- [x] Implement `process.on` handlers as in `fxa-auth-server` and elsewhere for graceful restarts, logging of unhandled rejections ,etc.
- [x] Make a `simulate_webhook_call` script for use by ops to verify a RP has setup their webhookUrl handler correctly.
- [x] Handle 404's less noisily at debug level.
- [x] Log any malformed messages and ack to avoid clogging the queue.

jrgm:
I'll add another item to this list:

- Handle 404 requests less noisily at debug level: Right now, a request for an unknown route produces a stack trace. (The internet has all kinds of random traffic (mostly looking for `..\..\..\..\system32.exe` and such ;-) ) and a single 404 log line is sufficient). [Yes, I have debug level on right now; that will change before going live, but thought it might be useful while getting this all up].
```
Debug: handler, error
Error: Not Found
at internals.notFound (/app/node_modules/@hapi/hapi/lib/core.js:623:16)
at module.exports.internals.Manager.execute (/app/node_modules/@hapi/hapi/lib/toolkit.js:41:33)
at Object.internals.handler (/app/node_modules/@hapi/hapi/lib/handler.js:46:48)
at exports.execute (/app/node_modules/@hapi/hapi/lib/handler.js:31:36)
at Request._lifecycle (/app/node_modules/@hapi/hapi/lib/request.js:312:68)
at process._tickCallback (internal/process/next_tick.js:68:7)
```

jrgm:
And another one. 

If there something malformed in the queue that can never be processed that message is not removed from the queue, and is handled again and again (until, I presume, the queue retention ages it out of the queue).

Sorry, I lost the stack for what was malformed, but it was something where I manually inserted into the queue a message like `{"event":"login"}` and it didn't like that it was missing required parameters.

jrgm:
> Create a __heartbeat__ route.

And a `/__lbheartbeat__` route, which returns 200.

PR #1787, opened 2019-07-16
feat(event-broker): clean shutdown, allow extra pubsub vars, lbheartbeat

bbangert:
Issue #1774

### DIFF
--- a/packages/fxa-event-broker/bin/worker.ts
+++ b/packages/fxa-event-broker/bin/worker.ts
@@ -51,7 +51,28 @@ async function main() {
     logger,
     webhookService
   );
-  await server.start();
+  try {
+    await server.start();
+  } catch (err) {
+    logger.error('startup', { err });
+    process.exit(1);
+  }
+
+  process.on('uncaughtException', err => {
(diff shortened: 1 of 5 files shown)


packages/fxa-event-broker/bin/worker.ts
```diff
@@ -51,7 +51,26 @@ async function main() {
...
-  await server.start();
+  try {
+    await server.start();
+  } catch (err) {
+    logger.error('startup', { err });
+    process.exit(1);
+  }
+
+  process.on('uncaughtException', err => {
+    process.exit(8);
```
philbooth: Does this one not warrant a `logger.error` too?

packages/fxa-event-broker/bin/worker.ts
```diff
@@ -51,7 +51,26 @@ async function main() {
...
+  });
+  process.on('unhandledRejection', (reason, promise) => {
+    logger.error('unhandledRejection', { error: reason });
```
philbooth: And this one a `process.exit`?

Issue #1779, opened 2019-07-15
[auth server] Improve code re-use in email methods


philbooth:
I'm adding some new email-related stuff for #927 and the level of boilerplate copypasta in those modules is pretty annoying. Didn't want to do it in that PR but I think we can replace most of it with a couple of parameterised functions.

vbudhram:
FWIW, I'm 👍 for this. There is a good amount of boilerplate stuff that needs to be done to add a new template. 

Additionally, I think parts of `sender.js` and `email.js` could also be combined so that we can avoid doing nested calls with same function names like https://github.com/mozilla/fxa/blob/b22d8cdeb47cf4dcd52aa90fc8e21f4eedd24725/packages/fxa-auth-server/lib/senders/index.js#L133. We could have `email.js` be responsible for just sending emails.

That refactor could be hairy though...

philbooth:
> ...we can avoid doing nested calls with same function names like...

I intend to go further than that fwiw. Not only do we not need a nested layer of calls, we don't need the outer layer either. It's completely unnecessary copypasta. We just need one function `send` that works the same for every template, with any variations in behaviour handled by arguments and/or config. We can delete about 2,000 lines of production code and a load of test code too.

philbooth:
And the duplication in the templates can all be got rid of too, if we refactor the common bits to partials.

philbooth:
And the weird mixing of both nunjucks and handlebars for templates is unnecessary too, we only need one of them.

shane-tomlinson:
> And the weird mixing of both nunjucks and handlebars for templates is unnecessary too, we only need one of them.

IIRC this was to precompile nunjucks templates into handlebars templates, but I forget why that was necessary. @vladikoff, do you remember?

philbooth:
> IIRC this was to precompile nunjucks templates into handlebars templates, but I forget why that was necessary.

It's used for block substitution, which I've replaced with handlebars partials syntax in #1791.

PR #912, opened 2019-04-23
fix(emails): reverse engineer partials for verification reminders

philbooth:
I didn't realise the email templates are generated from partials, so I created the verification reminder templates by hand. This change creates partials for them, so that they can be updated with `grunt templates`. I then ran `grunt templates` to check my work, which explains the changed `<title>` elements.

@mozilla/fxa-devs r?

PR #1791, opened 2019-07-16
feat(email): send new subscribers a product download link

### LINKED ISSUE #927: New email to link users back to a service download page immediately after they've subscribed
[Requirements here](https://docs.google.com/document/d/1dmo9XtEecFl93yVi4qA_EPtVJhWKCmXPQv5bFQHU_0o/edit#heading=h.gskx4txlgqyw)

philbooth:
Fixes #927.

When we're notified of a new subscription, send that user an email containing a product download link.

The designs for the subscription emails are completely different to the normal FxA emails, so I've implemented this as a separate pipeline with its own set of templates and so on. That also represented an opportunity to make a step towards #1779, so you'll notice some differences in the new pipeline:

* It doesn't use nunjucks, only handlebars. That allows us to drop a dependency and stops the templates from being a confusing mix of two different template syntaxes.
* The templates are refactored to use partials for common elements like buttons, icons and metadata. That reduces the amount of duplication in our templates.
* The layouts and partials are applied dynamically without serialising the end result to disk in the source tree. That should eliminate the kind of confusion that led to #912.
* There is a common `render` function that works for all templates. That's the kind of re-use I had in mind when I opened #1779.

Opened as a draft because I've only implemented the structural parts so far, I haven't styled the emails or written any tests yet. I figured the above discussion might be interesting enough that people want to see this alternative approach early doors, so they can tell me why I'm a fool and it won't work. 😄

### DIFF
--- a/packages/fxa-auth-server/bin/subhub_messaging.js
+++ b/packages/fxa-auth-server/bin/subhub_messaging.js
@@ -5,25 +5,51 @@
 
 'use strict';
 
+const LIB_DIR = '../lib';
+
 // This MUST be the first require in the program.
 // Only `require()` the newrelic module if explicity enabled.
 // If required, modules will be instrumented.
-require('../lib/newrelic')();
+require(`${LIB_DIR}/newrelic`)();
 
 const config = require('../config').getProperties();
-const log = require('../lib/log')(config.log.level, 'subhub-messaging');
-const Token = require('../lib/tokens')(log, config);
-const SQSReceiver = require('../lib/sqs')(log);
-// FIXME: import a different module:
-const subhubUpdates = require('../lib/subhub/updates')(log, config);
+const log = require(`${LIB_DIR}/log`)(config.log.level, 'subhub-messaging');
+const Promise = require(`${LIB_DIR}/promise`);
+const Token = require(`${LIB_DIR}/tokens`)(log, config);
+const SQSReceiver = require(`${LIB_DIR}/sqs`)(log);
+const subhubUpdates = require(`${LIB_DIR}/subhub/updates`)(log, config);
 
-const DB = require('../lib/db')(config, log, Token);
+run();
 
-const subhubUpdatesQueue = new SQSReceiver(
-  config.subhubServerMessaging.region,
-  [config.subhubServerMessaging.subhubUpdatesQueueUrl]
-);
+async function run() {
+  try {
+    const subhubUpdatesQueue = new SQSReceiver(
+      config.subhubServerMessaging.region,
+      [config.subhubServerMessaging.subhubUpdatesQueueUrl]
+    );
 
-DB.connect(config[config.db.backend]).then(db => {
-  subhubUpdates(subhubUpdatesQueue, db);
-});
+    const [db, translator] = await Promise.all([
+      require(`${LIB_DIR}/lib/db`)(config, log, Token).connect(
+        config[config.db.backend]
+      ),
+      require(`${LIB_DIR}/senders/translator`)(
+        config.i18n.supportedLanguages,
+        config.i18n.defaultLanguage
+      ),
+    ]);
+
+    const { email: mailer } = await require(`${LIB_DIR}/senders`)(
+      log,
+      config,
+      require(`${LIB_DIR}/error`),
+      require(`${LIB_DIR}/bounces`)(config, db),
+      translator,
+      require(`${LIB_DIR}/oauthdb`)(log, config)
+    );
+
+    subhubUpdates(subhubUpdatesQueue, db, mailer);
+  } catch (err) {
+    log.error('bin.subhub.error', { err });
+    process.exit(1);
+  }
+}
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -290,6 +290,24 @@ const conf = convict({
       default:
         'https://support.mozilla.org/kb/password-manager-remember-delete-change-and-import#w_viewing-and-deleting-passwords',
     },
+    subscriptionDownloadUrl: {
+      default: 'https://accounts.firefox.com/subscriptions/download',
+      doc: 'Subscription download URL',
+      env: 'SUBSCRIPTION_DOWNLOAD_URL',
+      format: String,
+    },
+    subscriptionSettingsUrl: {
+      default: 'https://accounts.firefox.com/subscriptions',
+      doc: 'Subscription settings URL',
+      env: 'SUBSCRIPTION_SETTINGS_URL',
+      format: String,
+    },
+    subscriptionTermsUrl: {
+      default: 'https://accounts.firefox.com/legal/subscription_terms',
+      doc: 'Subscription terms and cancellation policy URL',
+      env: 'SUBSCRIPTION_TERMS_URL',
+      format: String,
+    },
     sesConfigurationSet: {
       doc:
         'AWS SES Configuration Set for SES Event Publishing. If defined, ' +
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -14,7 +14,10 @@ const safeUserAgent = require('../userAgent/safe');
 const Sandbox = require('sandbox');
 const url = require('url');
 
-const TEMPLATE_VERSIONS = require('./templates/_versions.json');
+const TEMPLATE_VERSIONS = {
+  ...require('./templates/_versions.json'),
+  ...require('./subscription-templates/_versions.json'),
+};
 
 const DEFAULT_LOCALE = 'en';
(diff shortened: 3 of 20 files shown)


lmorchard:
Just to keep the wip going - I pushed up a commit with email HTML copypasta'd from salesforce. We can continue to iterate as necessary

philbooth:
This branch now has tests and the output markup has been tested [in Litmus](https://litmus.com/checklist/emails/public/a540006) too.

There are some outstanding questions before it's ready for code review however:

1. The Firefox logo that we got from the marketing email looks different to the one in our designs. Do we need a new logo? Who can provide that?

   Marketing's:

   Screenshot of marketing's Firefox logo

   Our's:

   Screenshot of our Firefox logo

2. We don't yet have a product logo for Secure Proxy. That markup is conditional in the template, so for now it just displays no logo rather than any kind of placeholder. Can we land it without that logo or do we need to block on it?

3. There are 4, or possibly 5, URLs I need to plug into config.

   The 4 definites are the links from the footer:

   Screenshot of the footer links

   Where should these go to?

   The possible 5th is the support link. Right now it links to the same SUMO article we use for the rest of FxA. But that seems wrong in this context:

   https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account

4. It looks like the designs use some fancy font, whereas the marketing emails just use plain old `font-family: Helvetica, Geneva, Tahoma, Verdana, sans-serif`. Where can I find the fancy font that I should be using? (ideally a URL to a CDN)

5. What are the expectations around responsive layout? I didn't see that mentioned anywhere. Currently the panel is hard-coded to 550 pixels wide. See the mobile screenshots [in Litmus](https://litmus.com/checklist/emails/public/a540006) for where that looks funny.

6. One of the tests [in Litmus](https://litmus.com/checklist/emails/public/a540006) is Outlook (dark mode). The Firefox logo suffers from poor contrast there because of the black text on dark background. Do we care enough to do something about that?

6. Front-end stuff is pretty much the opposite of whatever my forté is. Padding, margins, line-heights etc are probably all a bit out. If anyone more front-endy wants to come in here and clean up a bit, be my guest.

Current rendered state, with the above questions unresolved:

Screenshot of the download subscription email

lmorchard:
> There are 4, or possibly 5, URLs I need to plug into config

Both "Cancel subscription" and "Update billing information" should link to `/subscriptions` on content-server. (Not sure if that should also include some metrics params or suchlike)

I'm guessing "Terms" and "Privacy" don't exist and are related to issue #1986

> Front-end stuff is pretty much the opposite of whatever my forté is. Padding, margins, line-heights etc are probably all a bit out. If anyone more front-endy wants to come in here and clean up a bit, be my guest.

Same with me - I was hoping to just copy the Salesforce template for consistency and piggyback off their efforts. Will be weird if this email looks different from all the other billing emails.

philbooth:
> There are 4, or possibly 5, URLs I need to plug into config.

Dammit, I forgot one. Like, the most important one of them all, the actual product download URL. Is that going to be a generic URL that takes a `product_id` query param, or will there be bespoke URLs for every different product?

philbooth:
> Dammit, I forgot one. Like, the most important one of them all, the actual product download URL.

To make progress on this PR in time for l10n, I've taken an punt and gone with the following for now:

https://accounts.firefix.com/subscriptions/download?product_id=wibble

It's set in config so is easy to change as soon as we have an actual URL to work with. I figure that's fine as a follow-up (I'll open issues for all the follow-ups if this PR lands).

philbooth:
Calling this ready for review so that there's plenty of time to land the strings for l10n.

There will be follow-ups, which I'll open issues for if this lands.

Current list of follow-up issues:

* Currently the download link points to https://accounts.firefix.com/subscriptions/download?product_id=wibble (set in config). We either need to replace that with the correct download link, or add another redirect route to the content server to go to the appropriate place.

* We need a production asset for the Secure Proxy icon. Currently we're not displaying any icon instead.

* We probably need a different SUMO link for subscription stuff. Currently it links to the main FxA SUMO article.

* Maybe we need a new Firefox logo. The one I'm using is the same as the one in the marketing emails, but the designs for this email were slightly different (see above).

@mozilla/fxa-devs r?

packages/fxa-auth-server/bin/subhub_messaging.js
```diff
@@ -5,25 +5,51 @@
...
-const subhubUpdates = require('../lib/subhub/updates')(log, config);
-
-const DB = require('../lib/db')(config, log, Token);
-
-const subhubUpdatesQueue = new SQSReceiver(
-  config.subhubServerMessaging.region,
-  [config.subhubServerMessaging.subhubUpdatesQueueUrl]
-);
-
-DB.connect(config[config.db.backend]).then(db => {
```
philbooth: I moved this stuff into an `async` function so that we could `await`. Apologies for the diff noise, the substantive change is to initialise the mailer and pass it to subhub.

packages/fxa-auth-server/config/index.js
```diff
@@ -290,6 +290,24 @@ const conf = convict({
...
+      format: String,
+    },
+    subscriptionSettingsUrl: {
+      default: 'https://accounts.firefox.com/subscriptions',
+      doc: 'Subscription settings URL',
+      env: 'SUBSCRIPTION_SETTINGS_URL',
+      format: String,
+    },
+    subscriptionTermsUrl: {
+      default: 'https://accounts.firefox.com/legal/subscription_terms',
```
philbooth: Ref: #1986

packages/fxa-auth-server/config/index.js
```diff
@@ -290,6 +290,24 @@ const conf = convict({
       default:
         'https://support.mozilla.org/kb/password-manager-remember-delete-change-and-import#w_viewing-and-deleting-passwords',
     },
+    subscriptionDownloadUrl: {
+      default: 'https://accounts.firefox.com/subscriptions/download',
```
philbooth: Needs follow-up.

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -164,20 +175,26 @@ module.exports = function(log, config, oauthdb) {
...
     this.initiatePasswordChangeUrl = mailerConfig.initiatePasswordChangeUrl;
     this.initiatePasswordResetUrl = mailerConfig.initiatePasswordResetUrl;
     this.iosUrl = mailerConfig.iosUrl;
     this.iosAdjustUrl = mailerConfig.iosAdjustUrl;
     this.mailer = sender || nodemailer.createTransport(options);
-    this.emailService = sender || require('./email_service')(config);
     this.passwordManagerInfoUrl = mailerConfig.passwordManagerInfoUrl;
     this.passwordResetUrl = mailerConfig.passwordResetUrl;
+    this.prependVerificationSubdomain =
+      mailerConfig.prependVerificationSubdomain;
```
philbooth: Re-ordering to alphabetical.

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -293,14 +308,27 @@ module.exports = function(log, config, oauthdb) {
   Mailer.prototype.localize = function(message) {
     const translator = this.translator(message.acceptLanguage);
 
-    const localized = this.templates[message.template](
-      extend(
+    let localized;
+    if (message.layout === 'subscription') {
```
philbooth: For now there are two different templating systems in play, but when I fix #1779 this condition will disappear.

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1716,6 +1744,46 @@ module.exports = function(log, config, oauthdb) {
...
+      query,
+      template
+    );
+    const headers = {
+      'X-Link': links.link,
+    };
+    // TODO: product, subject, action and icon must vary per subscription for phase 2
+    const product = gettext('Secure Proxy');
+    const subject = gettext('Welcome to Secure Proxy!');
+    const action = gettext('Download Secure Proxy');
```
philbooth: I figured it was fine to hard-code these for now. When we have more products though, I guess we'll need some kind of lookup operation matching product ids to strings, which sounds like a bit of a nightmare tbh.

philbooth:
Actually, just tried the l10n script and it failed. Digging into it now.

philbooth:
> Actually, just tried the l10n script and it failed. Digging into it now.

All good now, I just needed to pass `ecmaVersion: 2018` in to jsxgettext.

lmorchard:
Filed some followups - let me know if I missed any or didn't get them right:

* https://github.com/mozilla/fxa/issues/2025
* https://github.com/mozilla/fxa/issues/2026
* https://github.com/mozilla/fxa/issues/2027

philbooth:
> Filed some followups

Perfect, thanks. Also filed #2030 fwiw.

Issue #1187, opened 2019-05-21
Rename `productName` to `productId` in `accountSubscriptions` table


lmorchard:
In PR #1174, I kind of tangled up renaming a column (good idea) and adding a column (not quite needed). Decided to just close that PR and run with the existing column in other PRs. As follow-up work, it might be good to circle back to actually just rename `productName` to `productId`

Mainly pushing this off as a follow-up because there's a lot of paperwork between fxa-auth-db-mysql and everything downstream from there that needs doing to reflect the change.

lmorchard:
FWIW, this is the PR that I withdrew and filed this issue instead:

https://github.com/mozilla/fxa/pull/1174

In that PR, I renamed `productId` to `productName` and then added a new `productId`. Ended up being a +176 patch (not 300-400 like I overexaggerated in the last meeting)

Might be worth just dusting off that branch and seeing about shepherding it to a merge.

philbooth:
Sticking my face on this, I have a branch ready to PR when we start the train 145 cycle...

lmorchard:
FWIW, there's going to be some spots in fxa-payments-server that use productName too. I tried to [stick todo comments by them](https://github.com/mozilla/fxa/search?q=productName+todo&unscoped_q=productName+todo), but it's very likely I missed a bunch

philbooth:
Yep, I picked up the ones in the auth server (I think that's what you meant?) and the event broker in my branch. PR coming shortly. The db change itself will be in two parts, so that the migration can run without breaking the existing stored procedures.

philbooth:
Oh wait, no, you did mean the payments server sorry, just grepped. Okay, I'll pick those up tomorrow, thanks.

Issue #970, opened 2019-04-30
Display a message in settings UI when a user’s payment method is out of date


lmorchard:
Things that probably need gluing together for this:

- [ ] Event handling for [`invoice.payment_action_required`](https://stripe.com/docs/billing/webhooks#action-required) webhook from Stripe or similar from subhub
- [ ] DB table or column to record that payment method is out of date
- [ ] auth server API additions to report the payment out of date status
- [ ] content server UI to display the payment out of date status
- [ ] clear the out of date status at the when updating payment method in our UI?

lmorchard:
Just noticed this point in the PRD for which we had no corresponding issue. Filing and estimating high for the unknowns in event handling.

johngruen:
@clouserw confirming that we should do this, but we can backload it in the end of the milestone. @juliastarkov can you add this screen to your wires. 

It should be a persistent warning in settings that says something like "Your payment method is expired. Please update your credit card to continue using %{service}" with a link to the payments update page.

juliastarkov:
@johngruen Yes I will add that.

juliastarkov:
@johngruen @lmorchard Here is the persistent notice screen: https://mozilla.invisionapp.com/share/K2RBDY3VU9X#/362078565_Firefox_Accounts_In_Browser_-Expired_CC-

(I feel like this should be sufficient for every screen, but please let me know if you want something more explicit in the Billing Information section on the Subscriptions page)

clouserw:
Ben says it's possible (or should be) that we can query subhub from the auth server when a user hits their settings page

bbangert:
Linking to mozilla/subhub#45, as it will track any subhub work needed for this.

bbangert:
Per the subhub issue, I don't think we should be storing this in our system. The SubHub API should indicate when we query for the users subscription that it needs updating.

lmorchard:
There's both `/customer/{uid}/subscriptions` and `/customer/{uid}` routes in the API.

The latter (`/customer/{uid}`) is probably a good place for it, since we query that to get current payment info for display anyway.

lmorchard:
Taking a look at recent subhub updates: https://github.com/mozilla/subhub/commit/51ffbf88bd39f260d8a47de5ab31a704e14d2e42

[...]

PR #1801, opened 2019-07-16
feat(fxa-content-server): fixes #1071 - new delete account view to display apps & subscriptions

### LINKED ISSUE #1071: Handle Account Deletion for Subscribing Users

LZoog:
fixes #1071 

Notes:
I checked with Julia about disabled button active states when addressing a hover/active state change request from John and we agreed it made sense to not give visual feedback when clicking a disabled button so that change was made in `button-row.scss` instead of on the delete account button specifically.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
@@ -1,8 +1,11 @@
-<div id="delete-account" class="settings-unit">
+<div id="delete-account" class="settings-unit {{#isPanelOpen}}open{{/isPanelOpen}}">
   <div class="settings-unit-stub">
     <header class="settings-unit-summary">
       <h2 class="settings-unit-title">{{#t}}Delete account{{/t}}</h2>
     </header>
+    <button class="settings-button secondary-button settings-unit-loading" disabled>
+      <div class="spinner spinner-settings-fetch"></div>
+    </button>
     <button class="settings-button secondary-button settings-unit-toggle"
       data-href="settings/delete_account">{{#t}}Delete…{{/t}}</button>
   </div>
@@ -11,7 +14,57 @@
     <div class="error"></div>
 
     <form novalidate>
-      <p>{{#t}}If you use this account for Mozilla apps and services like AMO, Pocket, and Screenshots you should first log in and remove any saved personal information. Once you delete your account here you may lose access to it everywhere.{{/t}}</p>
+      <div class="delete-account-product-container {{#productListError}}hide{{/productListError}}">
+        <p>{{#t}}You've connected your Firefox account to Mozilla products that keep you secure and productive on the web:{{/t}}</p>
+
+        <ul class="delete-account-product-list {{#hasTwoColumnProductList}}two-col{{/hasTwoColumnProductList}}">
+          {{#subscriptions}}
+            {{#plan_name}}
+              <li class="delete-account-product-subscription" title="{{plan_name}}">
+                {{plan_name}}
+              </li>
+            {{/plan_name}}
+          {{/subscriptions}}
+
+          {{#clients}}
+            {{#isWebSession}}
+              {{#title}}
+                <li class="delete-account-product-client" title="{{title}}">
+                  {{title}}
+                </li>
+              {{/title}}
+            {{/isWebSession}}
+
+            {{#isOAuthApp}}
+              {{#name}}
+                <li class="delete-account-product-client" title="{{title}}">
+                  {{name}}
+                </li>
+              {{/name}}
+            {{/isOAuthApp}}
+
+          {{/clients}}
+        </ul>
+      </div>
+
+      <p>{{#t}}Please acknowledge that by deleting your account:{{/t}}</p>
+
+      <ul class="delete-account-checkbox-list">
+        <li class='delete-account-checkbox-list-item'>
+          <input id="delete-account-subscriptions" type="checkbox" value="delete-account-subscriptions" class="delete-account-checkbox" required />
+          <label for="delete-account-subscriptions">{{#t}}Any subscriptions you have will be cancelled{{/t}}</label>
+        </li>
+
+        <li class='delete-account-checkbox-list-item'>
+          <input id="delete-account-saved-info" type="checkbox" value="delete-account-saved-info" class="delete-account-checkbox" required />
+          <label for="delete-account-saved-info">{{#t}}You may lose saved information and features within Mozilla products{{/t}}</label>
+        </li>
+
+        <li class='delete-account-checkbox-list-item'>
+          <input id="delete-account-reactivate" type="checkbox" value="delete-account-reactivate" class="delete-account-checkbox" required />
+          <label for="delete-account-reactivate">{{#t}}Reactivating with this email may not restore your saved information{{/t}}</label>
+        </li>
+      </ul>
 
       <div class="input-row password-row">
         <!-- add the autocomplete=off attribute - the user is deleting the account, they most likely don't want the password saved. -->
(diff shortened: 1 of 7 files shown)


lmorchard:
> Is there a WIP label? I don't see one.

There's no WIP label, but GitHub did introduce a "draft" status for PRs that we started using. Not sure if you can go from "open" back to "draft" though

packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
```diff
@@ -11,7 +14,50 @@
...
     <form novalidate>
-      <p>{{#t}}If you use this account for Mozilla apps and services like AMO, Pocket, and Screenshots you should first log in and remove any saved personal information. Once you delete your account here you may lose access to it everywhere.{{/t}}</p>
+      <p>{{#t}}Your Firefox Account gives you access to products and services that help keep your web experience secure, private, and productive.{{/t}}</p>
+
+      <p>{{#t}}You currently use:{{/t}}</p>
+
+        <ul class="account-summary-list">
+          {{#subscriptions}}
+            <li>
+              {{productName}}
```
lmorchard: Oof, [I wonder if this bit of tech debt is already starting to bite us](https://github.com/mozilla/fxa/issues/1187)? 😓  Despite the name "name", this is actually an ID and not a human-readable description.

Now that I belatedly think of it, the resource that `account.fetchActiveSubscriptions()` requests doesn't actually include human-readable descriptions. That particular resource is used [to build subscription capability lists for profile data](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/utils/subscriptions.js#L25).

On the payments-server pages, we use [the customer data from subhub](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/subscriptions.js#L209) which lists subscriptions that [*do* have human readable names](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/validators.js#L303). We might need a revisit of PR #1715 and PR #1773 to `GET /oauth/subscriptions/customer`

packages/fxa-content-server/app/scripts/views/settings/delete_account.js
```diff
@@ -10,21 +10,92 @@ import ServiceMixin from '../mixins/settings-panel-mixin';
...
+      this._fetchActiveSubscriptions(),
+    ]).then(() => this.render());
+  },
+
+  _fetchActiveSubscriptions() {
+    const account = this.getSignedInAccount();
+    const start = Date.now();
+    return account.fetchActiveSubscriptions().then(subscriptions => {
+      // do I need to this up anywhere else or does this create `timing.subscriptions`?
+      this.logFlowEvent(`timing.subscriptions.fetch.${Date.now() - start}`);
```
philbooth: To answer your question, nope it will just work how you have it. After it's deployed we can update [the charts](https://sql.telemetry.mozilla.org/queries/49429/source#133062) to pull in the new value.

packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
```diff
@@ -11,7 +14,50 @@
...
+            </li>
+          {{/subscriptions}}
+
+          {{#clients}}
```
shane-tomlinson: Are any changes needed here to match [@rfk's PR to list all attached clients in the devices and apps view](https://github.com/mozilla/fxa/pull/1262)? From my reading here it looks like this already conforms, 👍, but want to make sure.

LZoog: 👍 Thanks for calling this out, I made sure to pull from master yesterday but I know I copied from/referenced that clients view late last week (when I may have been up-to-date anyway) so it doesn't hurt to double check.

packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
philbooth: There is overlap here with what I'm doing for #970 too, I now realise. I'm adding a new endpoint that returns, among other things, the result from `subhub.listSubscriptions` (so that we can check for `failure_code` properties and display an error message). Seems like the same data would be useful here (for different reasons).

philbooth: Oh maybe not, `listSubscriptions` returns `plan_name` but not `product_name`. I guess that's different?

LZoog: Ah, yeah, I definitely assumed that was the human-readable _name_. 😅 It sounds like we might need to address #1187 + add a new field if what Phil is working on/`plan_name` is not what we need?

lmorchard: > It sounds like we might need to address #1187 + add a new field

Addressing #1187 won't really help here much. And I don't think we *want* to add a human-readable label to the FxA database: we'd only ever use it where subhub APIs are also available - and they already have the labels.

> Oh maybe not, listSubscriptions returns plan_name but not product_name. I guess that's different?

Using `listSubscriptions` is probably what we want. The `plan_name` describes the user's actual subscription plan - e.g. "123Done Pro Plus Monthly". A `product_name` would describe what the plan grants - e.g. "123Done with custom emoji enabled".

If we really, really need `product_name` we would need to either a) request that subhub make a change to include `product_name` or b) do what we do in payments-server by also fetching `listPlans` and looking up the `product_name` for the respective `plan_name` there.

LZoog: Then perhaps I just need to check with UX that displaying `plan_name` is along the line of what they were envisioning... They seem flexible with this section.

@philbooth, is that something you're working on now?

philbooth: Yeah, let split it out to a separate PR to get it through review quicker...

philbooth: Sorry for the delay, took me longer to write some passing tests than I expected (always the way). Here you go:

https://github.com/mozilla/fxa/pull/1824

packages/fxa-content-server/app/scripts/lib/fxa-client.js
```diff
@@ -1284,6 +1284,15 @@ FxaClientWrapper.prototype = {
...
   getActiveSubscriptions: createClientDelegate('getActiveSubscriptions'),
 
+  /**
+   * Get a list of settings data with a session token.
+   *
+   * @param {String} token A token from the OAuth server.
+   * @returns {Promise} A promise that will be fulfilled with a list of settings
+   * data.
+   */
+  account: createClientDelegate('account'),
```
philbooth: @LZoog, not sure how close this PR is to landing but just fyi, I've copied some of this stuff into the branch that @meandavejustice started for #970. Whoever lands second might get some merge conflicts, but they shouldn't be anything too complicated!

lmorchard:
I just went ahead and butted in with this commit - https://github.com/mozilla/fxa/pull/1801/commits/f941e77b2a6831080dcf67a6c44ec284587c55f8

Seemed easier than a gist this time, and I *think* it'll get some of those commented tests passing. Pretty much all just small tweaks.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -1000,6 +1000,15 @@ const Account = Backbone.Model.extend(
...
     },
 
+    /**
+     * Fetch the account's list of settings data.
+     *
+     * @returns {Promise} - resolves with a list of settings objects.
+     */
+    getSettingsData() {
+      return this._fxaClient.account(this.get('sessionToken'));
+    },
```
philbooth: Hey @LZoog, if you rebase this branch against latest `master`, you should be able to get rid of this method and call the `settingsData` method that was added in #1846.

PR #1806, opened 2019-07-17
fix(scoped-keys): Allow clients to provide `keys_jwk` when not requesting scoped keys.

### LINKED ISSUE #1792: Disable "No-key bearing scopes" requested check
Ref: https://github.com/mozilla/application-services/issues/1400#issuecomment-511253606

rfk:
Based on experience implementing consuming code, it's much more convenient to be able to always pass in `keys_jwk` than it is to require the client code to decide up-front whether any of its requested scopes will have keys.

This removes some "belt and braces" check that were enforcing clients the provide `keys_jwk` if and only if requesting a key-bearing scope, and instead just ignores `keys_jwk` when not requesting any key-bearing scopes.

Fixes #1792.

@vladikoff does this fit with what you were thinking?

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/reliers/oauth.js
+++ b/packages/fxa-content-server/app/scripts/models/reliers/oauth.js
@@ -105,6 +105,7 @@ var OAuthRelier = Relier.extend({
     this._config = options.config;
     this._oAuthClient = options.oAuthClient;
     this._session = options.session;
+    this._wantsScopeThatHasKeys = false;
   },
 
   fetch() {
@@ -317,7 +318,8 @@ var OAuthRelier = Relier.extend({
     return !!(
       this._config &&
       this._config.scopedKeysEnabled &&
-      this.has('keysJwk')
+      this.has('keysJwk') &&
+      this._wantsScopeThatHasKeys
     );
   },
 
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/models/reliers/oauth.js
```diff
@@ -317,16 +318,16 @@ var OAuthRelier = Relier.extend({
     return !!(
       this._config &&
       this._config.scopedKeysEnabled &&
-      this.has('keysJwk')
+      this.has('keysJwk') &&
+      this._wantsScopeThatHasKeys
```
rfk: This extra check prevents us from allocating a `keyFetchToken` if the relier doesn't actually want any keys.

packages/fxa-content-server/app/scripts/models/reliers/oauth.js
```diff
@@ -340,31 +341,25 @@ var OAuthRelier = Relier.extend({
...
     }
 
-    const validation = this._config.scopedKeysValidation || {};
-    let foundRedirectScopeMatch = false;
-
-    // Requesting keys, but not specifying a scope?  That's unexpected.
+    // If they're not requesting any scopes, we're not going to given them any keys.
     if (!this.get('scope')) {
-      throw new Error('Invalid scope parameter');
+      return false;
```
vladikoff: 👍

rfk:
> Did you want to uplift it to an earlier train instead of master, 

I don't think it's worth adding risk to train-141, which is already getting some point-releases in support of time-sensitive projects.

PR #1815, opened 2019-07-17
chore(release): Merge train 141 into master

shane-tomlinson:
Hey, what's the merge conflict here all about?


shane-tomlinson:
Replacing with #1820.

PR #1823, opened 2019-07-17
feat(support-panel): support live user queries

### LINKED ISSUE #704: [Support] Display live user queries
In #700 we only suggest a stub application, but of course this should be hooked up to the real database.

bbangert:
This makes some minor changes to the existing support panel stub such as:
- Converting to TypeScript
- Changing to Hapi for more framework consistency

It now does live queries to a fxa-auth-db-mysql instance directly, and when deployed will require an instance of it configured with a read-only mysql user.

Closes #704


bbangert:
Rebase and try again I guess...

Issue #1808, opened 2019-07-17
[content server] Reduce the number of auth server hits from the settings screen


philbooth:
When you load the settings screen it hits the auth server a whole heap of times, mostly requesting the same resource over and over. We could make a huge reduction in auth server traffic and improve client side performance a bit if we optimised that.

Most of the calls are to `GET /recovery_email/status`, but there's also 2 to `GET /totp/exists` and 1 to `GET /recovery_emails`. Chatting to @shane-tomlinson just now, he said it's probably because each of the settings child views makes it's own independent requests to the auth server.

Instead, we should have one gestalt-status endpoint that returns all of the necessary data for every child view, and request that once only from the parent view. It can then pass data into whichever child views want it and everything should work the same, except now we've helped avert climate change and everyone has extra bandwidth. Peace breaks out in the middle east.

<img width="865" alt="Screenshot of the FxA settings screen repeatedly hitting the auth server for the same resource" src="https://user-images.githubusercontent.com/64367/61378345-ba8deb80-a894-11e9-8f74-08c2ab2178f9.png" />

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-209)
┆Issue Number: FXA-209

philbooth:
> but there's also 2 to `GET /totp/exists`

Actually no, there's 1 to `GET /totp/exists` and 2 to `POST /totp/exists`. The posts seem especially weird because I don't have totp set up on this account and the request data is an empty object. Presumably they can just disappear completely.

philbooth:
I'm thinking of making the new gestalt endpoint just plain `GET /account`. That differentiates it better from the various `/status` endpoint we have, as something that is not meant for polling. And hopefully hints a bit that other properties may be added to it readily whenever needs arise, it's not limited to whatever set we start with.

vbudhram:
@philbooth The bulk of those calls are caused by panel gating, https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/app/scripts/views/mixins/upgrade-session-mixin.js#L129. Currently, the email, account recovery and 2FA panels will show an `Upgrade session via email` screen before you can actually use those features.

Ideally, we should cache the result in the user or account model and have that panel check that value.

data-sync-user:
➤ Wil Clouser commented:

Done with new settings

PR #1824, opened 2019-07-17
feat(api): add a GET /account endpoint

philbooth:
Related to #970, #1801 and #1808. Not a fix for any issue.

For #970, we need a way to interrogate the `failure_code` property from the settings view. We can't add it to any of the existing endpoints that view makes requests to because they are our high-traffic, polling endpoints and we don't want to add more overhead for a minority case. And we don't really want to add yet another request to the backend for this one specific purpose, that's exactly what #1808 is trying to avoid.

So this change adds a new endpoint with a generic name, `GET /account`. The long-term plan is that it will return all the stuff needed by the settings view and we'll only have to make one back-end request instead of the 10 or so that we currently make. But for now all I'm doing is returning the subscriptions array to make the new UI for #970 possible.

I've opened it as a separate PR because @LZoog also needs access to the `plan_name` property for #1801, and I don't want to block her any longer while I finish off #970.

There's also a secondary fix here; the behaviour of the stubbed `subhub.listSubscriptions` method was broken because it assumed the presence of a `uid` property in the subscription data. There isn't any, so I hacked it with some extra storage.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1453,6 +1453,48 @@ module.exports = (
         return {};
       },
     },
+    {
+      method: 'GET',
+      path: '/account',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+        response: {
+          schema: {
+            // This endpoint is evolving, it's not just for subscriptions.
+            // Ultimately we want it to become a one-stop shop for all of
+            // the account data needed by the settings screen, so that we
+            // can drastically reduce how many requests are made to the
+            // backend. Discussion in:
+            //
+            // https://github.com/mozilla/fxa/issues/1808
+            subscriptions: isA
+              .array()
+              .items(validators.subscriptionsSubscriptionValidator),
+          },
+        },
+      },
+      handler: async function(request) {
+        log.begin('Account.get', request);
+
+        const { uid } = request.auth.credentials;
+
+        let subscriptions;
+
+        if (config.subscriptions.enabled) {
+          try {
+            subscriptions = await subhub.listSubscriptions(uid);
+          } catch (err) {
+            if (err.errno !== error.ERRNO.UNKNOWN_SUBSCRIPTION_CUSTOMER) {
+              throw err;
+            }
+          }
+        }
+
+        return { subscriptions };
+      },
+    },
   ];
 
   if (config.isProduction) {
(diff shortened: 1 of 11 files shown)


rfk:
> The long-term plan is that it will return all the stuff needed by the settings view and
> we'll only have to make one back-end request instead of the 10 or so that we currently make. 

But...but...microservices?

Seriously though, 👍

lmorchard:
> But...but...microservices?

This is how you get ants^WGraphQL

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1453,6 +1453,48 @@ module.exports = (
         return {};
       },
     },
+    {
+      method: 'GET',
+      path: '/account',
```
vbudhram: @philbooth Sorry if I missed the reasoning, but why wouldn't we want to hook onto the [/account/profile](https://github.com/mozilla/fxa/blob/14314844e4c8b74d3f151a39809fe04c4009d518/packages/fxa-auth-server/lib/routes/account.js#L932) endpoint? Is it because that route is a high traffic route and used by profile server? 

Only mentioning because it appears to return subscription data already and is sessionToken authenticated.

philbooth: > why wouldn't we want to hook onto the /account/profile endpoint?

Fwiw I didn't consider it at the time because it's not one of the endpoints currently requested by the settings screen and I didn't want to weigh down other callers with stuff they don't need.

Looking at it now though, the subscriptions data returned from it is different to what I need, it's product ids from the auth db. I specifically want the `failure_code` property from subhub for #970 and we also want the `plan_name` property from subhub for #1801. Neither of those is in the auth db (by design), so we'd need to change its behaviour to make the extra request to subhub and also the behaviour of its consumer(s) so they don't expect strings.

I'm minded not to interfere with the other consumer like that and stick to using a fresh endpoint, but also happy to make the change if you think that's wrong. Let me know!

vbudhram: > but also happy to make the change if you think that's wrong

Nope, your reasoning sounds good. We have a couple endpoints that do account status related things and it wouldn't mind deprecate some of them in favor of this route.

Side note, looks like you will need to update the docs for route.

philbooth: > need to update the docs for route

Ah dammit. I remember when those docs used to update themselves, sigh.

PR #1828, opened 2019-07-17
fix: default ppid enabled to false

bbangert:
This change disables ppid by default since it won't run in dev/stage environments without further changes.


shane-tomlinson:
> This change disables ppid by default since it won't run in dev/stage environments without further changes.

Sorry about that, I overlooked latest. I think should re-enable this to set a salt in fxa-dev so that we can test on those boxes. local dev config has a salt set, as does stage/prod.

bbangert:
If it defaults to being properly configured to run on start, then it seems fine to have enabled true.

shane-tomlinson:
> If it defaults to being properly configured to run on start, then it seems fine to have enabled true.

Yeah, this is a better idea. Leave it disabled here, then on fxa-dev enable it and set a salt.

PR #1841, opened 2019-07-18
feat(oauth): Support the `resource` parameter when fetching JWT access tokens.

### LINKED ISSUE #1817: Support "Resource Indicators for OAuth 2.0" draft spec
Follow on from #1679 

See https://datatracker.ietf.org/doc/draft-ietf-oauth-resource-indicators/?include_text=1

The way PPIDs work right now is based on client_id. For a PPID enabled client an access token created to present to the profile server will contain the PPID sub. 

This doesn't currently cause any problems because all access tokens are database backed. We get the original access token from the jti field of jwt access token, then look up the access token in the database, and eventually get the user's real profile info. One of the long term goals of JWT access tokens is to stop writing to/reading from the database every time a token is created or verified. If tokens meant for the profile server contain a PPID sub, then we won't be able to look up a user's profile info.

Resource indicators solve this by allowing an RP to request multiple access tokens, one for each resource server. We can then customize the access token based on which resource server the token goes to, including [shaving info from the JWT](https://github.com/mozilla/fxa/issues/1797) and only using PPIDs for resource servers that need them.

If the RP specifies which `resource` an access token is meant for, we can include that information in the `aud` claim, giving the resource server an additional data point they can use in verification.

Ref #1814

shane-tomlinson:
Note, this only works with JWT access tokens and ID tokens. Whenever a `resource` is
specified, the `aud` claim of JWT access tokens or id tokens changes from a string
to an array of strings containing the client_id in position 0 and the `resource`
in position 1.

This PR does not yet return `aud` from either `/verify` or `/introspect` since
the field is not currently returned (see #1872).

Finally, this does not check the `resource` parameter against a pre-defined list
that is authorized for the given client_id.

fixes #1817

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
@@ -14,20 +14,18 @@ const HEADER_TYP = 'at+JWT';
  */
 exports.create = async function generateJWTAccessToken(accessToken, grant) {
   const clientId = hex(grant.clientId);
+  // The IETF spec for `aud` refers to https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+  // > REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the
+  // > OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain
+  // > identifiers for other audiences. In the general case, the aud value is an array of
+  // > case sensitive strings. In the common special case when there is one audience, the
+  // > aud value MAY be a single case sensitive string.
+  const audience = grant.resource ? [clientId, grant.resource] : clientId;
 
   // Claims list from:
   // https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00#section-2.2
   const claims = {
-    // The IETF spec for `aud` refers to https://openid.net/specs/openid-connect-core-1_0.html#IDToken
-    // > REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the
-    // > OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain
-    // > identifiers for other audiences. In the general case, the aud value is an array of
-    // > case sensitive strings. In the common special case when there is one audience, the
-    // > aud value MAY be a single case sensitive string.
-
-    // We only specify the clientId because we do not yet accept
-    // a `resource` parameter which we could use as the audience.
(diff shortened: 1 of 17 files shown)


shane-tomlinson:
@rfk - could I get your feedback on this? There are some caveats noted in the description, but I feel like this is coming together reasonably. 

For next steps, I'd like to propagate the `aud` that is returned in the JWT access token in `/introspect` and `/verify`, though I'm not sure I want to go through the pallaver of doing this for non-JWT access tokens since this would require a DB migration.

I should probably check `resource` against an authorized list for the given client_id as the spec recommends.

After that, the ppid logic can be updated to be based on resource rather than client_id.

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
```diff
@@ -14,20 +14,18 @@ const HEADER_TYP = 'at+JWT';
  */
 exports.create = async function generateJWTAccessToken(accessToken, grant) {
   const clientId = hex(grant.clientId);
+  // The IETF spec for `aud` refers to https://openid.net/specs/openid-connect-core-1_0.html#IDToken
+  // > REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the
+  // > OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain
+  // > identifiers for other audiences. In the general case, the aud value is an array of
+  // > case sensitive strings. In the common special case when there is one audience, the
+  // > aud value MAY be a single case sensitive string.
+  const audience = grant.resource ? [clientId, grant.resource] : clientId;
```
rfk: It's my continuing opinion that applying the `"It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value"` requirement from `id_token`s to `access_token`s is at best unnecessary, and at worst a harmful confusion of responsibilities that could lead to subtle security problems. But I can see how a strict read of both specs would lead to it being a requirement, so I guess I should take it up with the spec...

shane-tomlinson: I agree, it seems a bit strange. A couple of alternatives if no `resource` is specified:

* Don't send `aud` since it'll be the same as `client_id`.
* Send an empty array

shane-tomlinson:
@rfk I'm going to call this ready, I can address the caveats noted in the description if you feel they are necessary, otherwise I'm happy opening follow-ons since nobody is using these yet.

Issue #1723, opened 2019-07-09
[payments] Product redirect should complete after signup/confirmation


ianb:
Note this only applies after #1310 lands:

If you visit a product redirect page, like http://127.0.0.1:3030/subscriptions/products/123doneProProduct and sign up, you end up at the settings page. You should end up at the product page you intended to go to.

shane-tomlinson:
@ianb is the intention here to end up on the RP with OAuth credentials (an OAuth code/token)?

lmorchard:
> @ianb is the intention here to end up on the RP with OAuth credentials (an OAuth code/token)?

No, I think this issue is a part of the payment process for issue #1364. 

If the user enteres FxA via a subscription product marketing page, we want them to eventually end up on the product payment page in fxa-payments-server (ala issue #1682) after they complete whatever needs doing in the signup / signin flow.

At the end of that, we redirect to an RP product page, but without OAuth credentials - i.e. it's just a plain redirect. If the RP requires creds on the web from there, the user will probably have to sign in again - but that's not a problem we're solving yet.

lmorchard:
Adding to Train 142, because this is basically a sub-issue of #1364

ianb:
The redirect URL might be a simple change for the signin case (it doesn't work, but it's a small fix), but for signup with email verification this might be more complicated, as we want to redirect the user after they paste in the confirmation email. Currently the signup redirect uses the OAuth database to figure out where you want to ultimately go, and we need it to simply redirect to this internal URL (`/subscriptions/products/:productId`)

lmorchard:
FWIW, I just attempted visiting this product redirect URL from a signed-out state:

https://latest.dev.lcip.org/subscriptions/products/prod_Ex9Z1q5yVydhyk

I signed in and FxA redirected me to the payments server as desired. So, that part seems to already work here.

PR #1843, opened 2019-07-18
feat(support form): Add subscription support form

### LINKED ISSUE #797: [Support] Create support form

chenba:
The plumbing work was done in PR #1715 and #1773. 

Fixes #797

@mozilla/fxa-devs r?


packages/fxa-auth-server/lib/routes/support.js
```diff
@@ -45,7 +45,10 @@ module.exports = (log, db, config, customs, zendeskClient) => {
         validate: {
           payload: isA.object().keys({
             topic: isA.string().required(),
-            subject: isA.string().optional(),
+            subject: isA
+              .string()
+              .allow('')
```
chenba: This, or otherwise the client need to remove the `subject` property altogether from the payload when it is an empty string.

chenba:
A functional test is failing in CI because https://screencap.co.uk/images/1b7785ff3fdd9f429812908fb128aa84d730a9a1.png even though the `pm2` list shows that payments server is online.  😕

meandavejustice:
> A functional test is failing in CI because https://screencap.co.uk/images/1b7785ff3fdd9f429812908fb128aa84d730a9a1.png even though the pm2 list shows that payments server is online. 😕

I'm assuming this is because the auth token hasn't been exchanged yet when the test tries to hit `:3031`, this should be fixed in https://github.com/mozilla/fxa/issues/1723

meandavejustice:
Not sure if I'm running into a caching issue but I'm not able to view the support form when running locally, it just links to 404 at `support.accounts.firefox.com`

chenba:
> 
> 
> Not sure if I'm running into a caching issue but I'm not able to view the support form when running locally, it just links to 404 at `support.accounts.firefox.com`

@meandavejustice How are you getting to the form?  If you have a subscription, clicking on the "Manage..." button on the Settings page should correctly take you to the form.  I do remember there's a place with that URL, but can't recall where at the moment.

chenba:
> 
> 
> Not sure if I'm running into a caching issue but I'm not able to view the support form when running locally, it just links to 404 at `support.accounts.firefox.com`

@meandavejustice Sorry, I've been testing by going to the form directly so much that I forgot about that link.  I'll fix it today.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -68,13 +65,16 @@ export const Subscriptions = ({
...
+    config,
     locationReload,
     navigateToUrl,
+    queryParams
   } = useContext(AppContext);
 
   const [showPaymentSuccessAlert, setShowPaymentSuccessAlert] = useState(true);
   const clearSuccessAlert = useCallback(() => setShowPaymentSuccessAlert(false),
                                               [setShowPaymentSuccessAlert]);
+  const SUPPORT_FORM_URL = `${config.servers.content.url}/support`;
```
chenba: @meandavejustice This should now take you to the form.  (An active subscription is still required.)

packages/fxa-payments-server/src/components/DialogMessage.tsx
```diff
@@ -2,10 +2,10 @@ import React, { ReactNode } from 'react';
 import classNames from 'classnames';
 import { useClickOutsideEffect } from '../lib/hooks';
 
-import './DialogMessage.scss';
-
 import Portal from './Portal';
 
+import './DialogMessage.scss';
+
```
chenba: This fixes a `mini-css-extrac-plugin` warning that's treated as an error when the `CI` env var is set.  (See https://github.com/facebook/create-react-app/issues/5372)

packages/fxa-auth-server/lib/routes/support.js
meandavejustice: Have you spoken to julia about removing the empty subject as an option? It seems preferable to use the "general" subject for these cases

chenba: It's a little confusing there's a topic and a subject.  This is the subject text input, which is optional per the mock-ups.  I think you might be thinking of the topic, which is from the drop down.  That isn't optional.

Issue #1826, opened 2019-07-17
Create selenium test for fxa + subhub happy path


6a68:
Write a selenium test that creates a user and signs up for a subscription.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-218)
┆Issue Number: FXA-218

data-sync-user:
➤ Wil Clouser commented:

wontfix subhub

PR #1846, opened 2019-07-19
feat(settings): add ui for expired credit card alert

### LINKED ISSUE #970: Display a message in settings UI when a user’s payment method is out of date

philbooth:
Fixes #970.

Joint effort by me and @meandavejustice.

Displays a banner on the settings screen if a user has a subscription with an expired card. Wires in to the auth server work from #1824.

Although all the pieces are tested individually, I haven't actually done a proper end-to-end test because I'm not sure how. If someone else understands how to make that happen, let me know.

And /cc @LZoog, this conflicts with your WIP PR #1801. Note the changes to the `settingsData` method where it caches its own result to protect us from the situation discussed in #1808. That should mean we can each call it from wherever we need without needing to co-ordinate or whatever.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/constants.js
+++ b/packages/fxa-content-server/app/scripts/lib/constants.js
@@ -149,5 +149,8 @@ module.exports = {
   TWO_STEP_AUTHENTICATION_ACR: 'AAL2',
 
   STYLE_TRAILHEAD: 'trailhead',
+
+  // https://stripe.com/docs/error-codes#expired-card
+  CC_EXPIRED: 'expired_card',
 };
 /*eslint-enable sorting/sort-object-props*/
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -368,6 +368,35 @@ const Account = Backbone.Model.extend(
     },
 
     /**
+     * Fetches account details from GET /account, for use by the settings views.
+     * Caches its own result, because it's not intended for that endpoint to be
+     * called repeatedly like some of the polling methods are. If you're really,
+     * really sure you need to, pass the `force` option to force a clean request.
+     *
+     * @param {Object} [options]
+     *   @param {Boolean} [options.force=false] - Ignore any cached results
+     *
+     * @returns {Promise} - Resolves to the result of `GET /account`, as defined in
+     *                      `packages/fxa-auth-server/lib/routes/account.js`.
+     */
+    settingsData(options = {}) {
+      return Promise.resolve().then(() => {
+        if (this._settingsData && !options.force) {
+          return this._settingsData;
+        }
+
+        const sessionToken = this.get('sessionToken');
+        if (!sessionToken) {
+          throw AuthErrors.toError('INVALID_TOKEN');
+        }
+
+        return this._fxaClient
+          .account(sessionToken)
+          .then(result => (this._settingsData = result));
+      });
+    },
+
+    /**
      * This function simply returns the session status of the user. It differs
      * from `sessionStatus` function above because it is not used to determine
      * which view to take a user after the login. This function also does not
--- a/packages/fxa-content-server/app/scripts/templates/settings.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings.mustache
@@ -1,7 +1,7 @@
 <div id="fxa-settings-header-wrapper">
   {{#ccExpired}}
     <div class="cc-expired-alert" role="alert">
-      {{#unsafeTranslate}}Your credit card has expired. Please update it <a href="#" class="alert-link">here</a>.{{/unsafeTranslate}}
+      {{#unsafeTranslate}}Your credit card has expired. Please update it <a %(escapedCcExpiredLinkAttrs)s>here</a>.{{/unsafeTranslate}}
     </div>
   {{/ccExpired}}
   <header id="fxa-settings-header">
--- a/packages/fxa-content-server/app/scripts/templates/settings.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings.mustache
@@ -1,4 +1,9 @@
 <div id="fxa-settings-header-wrapper">
+  {{#ccExpired}}
+    <div class="cc-expired-alert" role="alert">
+      {{#unsafeTranslate}}Your credit card has expired. Please update it <a href="#" class="alert-link">here</a>.{{/unsafeTranslate}}
+    </div>
+  {{/ccExpired}}
   <header id="fxa-settings-header">
     <h1 id="fxa-manage-account"><span class="fxa-account-title">{{#t}}Firefox Accounts{{/t}}</span></h1>
     {{#showSignOut}}
--- a/packages/fxa-content-server/app/scripts/templates/settings.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings.mustache
@@ -13,6 +13,7 @@
 
   <div class="settings-success-wrapper">
     <div class="success settings-success"></div>
+    <div class="error settings-error {{#error}}visible{{/error}}">{{error}}</div>
   </div>
 </div>
 
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -107,6 +107,7 @@ const View = BaseView.extend({
     const account = this.getSignedInAccount();
 
     context.set({
+      // ccExpired: true,
       showSignOut: !account.isFromSync(),
       unsafeHeaderHTML: this._getHeaderHTML(account),
     });
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -147,13 +147,17 @@ const View = BaseView.extend({
       account.fetchProfile(),
       this.user.setAccount(account),
       account.settingsData(),
-    ]).then(([, , data]) => {
-      if (data && Array.isArray(data.subscriptions)) {
-        this._ccExpired = data.subscriptions.some(
-          s => s.failure_code === Constants.CC_EXPIRED
-        );
-      }
-    });
+    ])
+      .then(([, , data]) => {
+        if (data && Array.isArray(data.subscriptions)) {
+          this._ccExpired = data.subscriptions.some(
+            s => s.failure_code === Constants.CC_EXPIRED
+          );
+        }
+      })
+      .catch(err => {
+        this.model.set('error', err);
+      });
   },
 
   _onAccountUpdate(account) {
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -19,6 +19,7 @@ import ClientDisconnectView from './settings/client_disconnect';
 import ClientsView from './settings/clients';
 import Cocktail from 'cocktail';
 import CommunicationPreferencesView from './settings/communication_preferences';
+import Constants from '../lib/constants';
 import DeleteAccountView from './settings/delete_account';
 import DisplayNameView from './settings/display_name';
 import Duration from 'duration';
@@ -107,7 +108,8 @@ const View = BaseView.extend({
     const account = this.getSignedInAccount();
 
     context.set({
-      // ccExpired: true,
+      ccExpired: !!this._ccExpired,
+      escapedCcExpiredLinkAttrs: `href="/subscriptions" class="alert-link"`,
       showSignOut: !account.isFromSync(),
       unsafeHeaderHTML: this._getHeaderHTML(account),
     });
@@ -141,7 +143,17 @@ const View = BaseView.extend({
   beforeRender() {
     const account = this.getSignedInAccount();
 
-    return account.fetchProfile().then(() => this.user.setAccount(account));
+    return Promise.all([
+      account.fetchProfile(),
+      this.user.setAccount(account),
+      account.settingsData(),
+    ]).then(([, , data]) => {
+      if (data && Array.isArray(data.subscriptions)) {
+        this._ccExpired = data.subscriptions.some(
+          s => s.failure_code === Constants.CC_EXPIRED
+        );
+      }
+    });
   },
 
   _onAccountUpdate(account) {
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -50,11 +50,11 @@ body.settings #stage .settings {
   color: white;
   padding: 5px;
   text-align: center;
-}
 
-.cc-expired-alert a {
-  text-decoration: underline;
-  color: white;
+  a {
+    text-decoration: underline;
+    color: white;
+  }
 }
 
 // settings header controls
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -42,6 +42,21 @@ body.settings #stage .settings {
   }
 }
 
+.cc-expired-alert {
+  position: absolute;
+  top: 0;
+  width: 100%;
+  background: #d70022;
+  color: white;
+  padding: 5px;
+  text-align: center;
+}
+
+.cc-expired-alert a {
+  text-decoration: underline;
+  color: white;
+}
+
 // settings header controls
 #fxa-settings-header-wrapper {
   align-items: center;
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -136,6 +136,7 @@ body.settings #stage .settings {
     z-index: 2;
   }
 
+  .settings-error,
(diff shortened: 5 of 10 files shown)


packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -367,6 +367,34 @@ const Account = Backbone.Model.extend(
       });
     },
 
+    /**
+     * Fetches account details from GET /account, for use by the settings views.
+     * Caches its own result, because it's not intended for that endpoint to be
+     * called repeatedly like some of the polling methods are. If you're really,
+     * really sure you need to, pass the `force` option to force a clean request.
```
6a68: If the user's card expires, how would caching invalidation happen?

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -78,6 +79,11 @@ const View = BaseView.extend({
     this._subscriptionsManagementLanguages =
       options.subscriptionsManagementLanguages;
 
+    if (options.config && options.config.subscriptions) {
+      this._subscriptionManagementUrl =
+        options.config.subscriptions.managementUrl;
```
philbooth: I just read about #1850 in my bugmail, so it seems likely this doesn't do what I think it does.

@lmorchard, do you know what this URL is for, if it's not for #1850?

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -142,7 +150,17 @@ const View = BaseView.extend({
   beforeRender() {
     const account = this.getSignedInAccount();
 
-    return account.fetchProfile().then(() => this.user.setAccount(account));
+    return account
+      .fetchProfile()
+      .then(() => this.user.setAccount(account))
+      .then(() => account.settingsData())
```
6a68: It seems like you'll need to pass the force=true parameter here, otherwise you may not detect when something's gone wrong with the subscription or credit card. Or, if force=true might hammer on the endpoint because of React re-rendering, maybe do the `settingsData` caching at this view layer, so that a hard page refresh would re-fetch the data?

packages/fxa-content-server/app/scripts/lib/constants.js
```diff
@@ -149,5 +149,8 @@ module.exports = {
   TWO_STEP_AUTHENTICATION_ACR: 'AAL2',
 
   STYLE_TRAILHEAD: 'trailhead',
+
+  // https://stripe.com/docs/error-codes#expired-card
+  CC_EXPIRED: 'expired_card',
```
6a68: I'm not totally sure this will be the error string, although Les pointed me at [this line](https://github.com/mozilla/subhub/blob/master/subhub/api/payments.py#L216) in subhub, which suggests the stripe error codes are indeed passed through. We should confirm this with the subhub team. I filed https://github.com/mozilla/subhub/issues/148 to ask for some documentation.

6a68:
@philbooth One other comment - I'm not sure how far you'd want to go with an end-to-end test, but I think it should be possible to test for the card error by adding a mock error response to the auth server's mock subhub API (or setting up a sinon spy to return a 'subscription/card ok' response the first time it's called, and a 'card expired' response the second time). That would be an 'internal' end-to-end auth+content server test. I'm not sure we do those, and I'm not sure where we would put it, but it could be useful :-)

Another, more heavyweight approach: I'm working on an end-to-end test (#1826) that stands up a dockerized local copy of subhub, and creates a subscription using a stripe test key (until subhub gets the stripe-mock option in place). Once I get the happy path working (and PR landed), it would make sense to start adding tests for these kinds of edge cases, maybe by using the stripe API to directly update the card to a test expired card, then reload the page and look for the error UI.

packages/fxa-content-server/app/scripts/models/account.js
philbooth: It doesn't. I'm operating on the basis that:

1. Users don't tend to have long-lived sessions clicking around the settings view in FxA.

2. The message is informative rather than some kind of mandatory error, so it doesn't matter in the cases where it doesn't immediately pop up on someone's view the moment a card expires.

3. We don't really want to hit the auth server for this endpoint every time e.g. navigation between settings panels occurs.

And of course, it's not a cache in any real sense, a page refresh or a new session is all that's required to force the call to happen again.

But any or all of those assumption might be wrong.

packages/fxa-content-server/app/scripts/views/settings.js
philbooth: > ...so that a hard page refresh would re-fetch the data?

Oh wait, maybe I've missed something then. So the account model is persisted in local storage or somewhere? I didn't even realise.

philbooth:
Closing this while I tweak that caching behaviour, I didn't realise about the account model thing.

philbooth:
Actually no, I think we're fine on the storage thing, I don't see it anywhere locally. Re-opening, although we still need to sort out the end-to-end tests question.

Screenshot of sessionStorage and localStorage for a local FxA instance

packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -42,6 +42,21 @@ body.settings #stage .settings {
...
+  background: #d70022;
+  color: white;
+  padding: 5px;
+  text-align: center;
+}
+
+.cc-expired-alert a {
+  text-decoration: underline;
+  color: white;
+}
```
LZoog: You can nest this selector.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -367,6 +367,34 @@ const Account = Backbone.Model.extend(
...
+     *
+     * @param {Object} [options]
+     *   @param {Boolean} [options.force] - Ignore any cached results
+     *
+     * @returns {Promise}
+     */
+    settingsData(options = {}) {
```
shane-tomlinson: In the @returns documentation, can you leave a blurb about which fields the result contains? The name tells me *where* the data is used, but not what's in it.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -367,6 +367,34 @@ const Account = Backbone.Model.extend(
...
+      return Promise.resolve().then(() => {
+        if (this._settingsData && !options.force) {
```
shane-tomlinson: It seems like each of the fields in `this._settingsData` could be stored in the [account's attributes](https://github.com/mozilla/fxa/blob/c2e696500db9715eca8a42660c60c5a016335050/packages/fxa-content-server/app/scripts/models/account.js#L58) rather than stored as an instance variable? That would allow you to do things like `account.get('subscriptions')`. I can make a case either way, though storing model data in the model attributes is closer to how the models are meant to be used.

packages/fxa-content-server/app/scripts/templates/settings.mustache
```diff
@@ -1,4 +1,9 @@
 <div id="fxa-settings-header-wrapper">
+  {{#ccExpired}}
+    <div class="cc-expired-alert" role="alert">
+      {{#unsafeTranslate}}Your credit card has expired. Please update it <a %(ccExpiredLinkAttrs)s>here</a>.{{/unsafeTranslate}}
```
shane-tomlinson: I'm really surprised [ccExpiredLinkAttrs isn't being flagged as an invalid variable name](https://github.com/mozilla/fxa/blob/c2e696500db9715eca8a42660c60c5a016335050/packages/fxa-content-server/app/scripts/views/base.js#L503). Variable names within `unsafeTranslate` are supposed to [require an `escaped` prefix](https://github.com/mozilla/fxa/blob/c2e696500db9715eca8a42660c60c5a016335050/packages/fxa-content-server/app/scripts/lib/strings.js#L12) to remind the developer to properly escape them (comment left below about that).

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -367,6 +367,34 @@ const Account = Backbone.Model.extend(
...
     },
 
+    /**
+     * Fetches account details from GET /account, for use by the settings views.
+     * Caches its own result, because it's not intended for that endpoint to be
+     * called repeatedly like some of the polling methods are. If you're really,
+     * really sure you need to, pass the `force` option to force a clean request.
+     *
+     * @param {Object} [options]
+     *   @param {Boolean} [options.force] - Ignore any cached results
```
shane-tomlinson: nit: Change to `[options.force=false]` to indicate the default value.

packages/fxa-content-server/app/styles/modules/_settings.scss
shane-tomlinson: Yeah, might read a bit more easily if you put the `a {...}` block above nested in the `.cc-expired-alert` block.

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -2683,6 +2683,66 @@ describe('models/account', function() {
...
+    });
+
+    it('called fxaClient.account', () => {
+      assert.equal(fxaClient.account.callCount, 1);
+      const args = fxaClient.account.args[0];
+      assert.lengthOf(args, 1);
+      assert.equal(args[0], 'wibble');
+    });
+
+    describe('second call to settingsData:', () => {
```
shane-tomlinson: Should you have a test for `force: true`?

packages/fxa-content-server/app/tests/spec/views/settings.js
```diff
@@ -131,8 +136,90 @@ describe('views/settings', function() {
     assert.equal(args[1], 'wibble');
   });
 
+  describe('setInitialContext:', () => {
```
shane-tomlinson: Ahha, this is why the invalid variable name wasn't caught, the full render is never done with it in the unit tests.

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -109,6 +115,8 @@ const View = BaseView.extend({
     const account = this.getSignedInAccount();
 
     context.set({
+      ccExpired: !!this._ccExpired,
```
shane-tomlinson: If you manually set this to `true`, you'll see the problem I referred to above, `ccExpiredLinkAttrs` is an invalid variable name within `unsafeTranslate`

Screenshot 2019-07-23 at 10 23 37

packages/fxa-content-server/app/scripts/views/settings.js
shane-tomlinson: Is an error message displayed if `settingsData` rejects with an error? I think it'll end up [redirecting to the 500 page with no real user information](https://github.com/mozilla/fxa/blob/c2e696500db9715eca8a42660c60c5a016335050/packages/fxa-content-server/app/scripts/views/app.js#L150) which is probably not the desired behavior. Maybe catch errors and call `this.displayError`. Of course, for that to actually print anything, you need an element with the `.error` class in the settings template where the message will be printed.

It looks like `account.fetchProfile` would have suffered from this problem for years. How have we gotten away with this for so long, or am I missing something?

shane-tomlinson: Calling `fetchProfile` and `settingsData` serially will delay render, especially on slower devices. Can you do a `Promise.all` and make both requests at once?

packages/fxa-content-server/app/scripts/models/account.js
philbooth: Just making sure I understand the suggestion. Are you saying do that and still have this call return the aggregated result, or are you saying do that and then have callers call `account.get` for whichever properties it is they're interested in?

shane-tomlinson: I was saying the 2nd, though on reflection that doesn't seem like a good idea. Keep it like this.

packages/fxa-content-server/app/scripts/views/settings.js
philbooth: Yep, and actually, I just noticed that `user.setAccount` calls `account.fetch`, which calls `account.sessionStatus`, which is also a remote request. So I'll `Promise.all` all three...

packages/fxa-content-server/app/scripts/templates/settings.mustache
philbooth: I'll add another test case. Looks like I just need to replace my mock context with a backbone model and then call `renderTemplate` to trigger it...

packages/fxa-content-server/app/scripts/models/account.js
philbooth: The trap I wanted to avoid was defining something in a comment here that later becomes a lie when we return more stuff from `GET /account`. It seems unlikely that we (or at least I) will remember to come and update this comment when we do that. So I've hedged and said that it returns all the data from that endpoint, and linked to the module where the endpoint is defined.

Good enough?

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -109,6 +117,8 @@ const View = BaseView.extend({
...
+      escapedCcExpiredLinkAttrs: `href="${this._subscriptionManagementUrl}" class="alert-link"`,
```
lmorchard: This payments-server URL doesn't currently work as a plain link - it needs an access token to enable auth-server API access.

We just landed PR #1899 for issue #1850 to support linking to subscription management. The link URL here would need to be `{content-server base URL}/subscriptions/`. That will generate an access token and then redirect to subscription management on payments-server.

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -142,7 +152,21 @@ const View = BaseView.extend({
   beforeRender() {
     const account = this.getSignedInAccount();
 
-    return account.fetchProfile().then(() => this.user.setAccount(account));
+    return Promise.all([
+      account.fetchProfile(),
+      this.user.setAccount(account),
+      account.settingsData(),
+    ])
+      .then(([, , data]) => {
```
lmorchard: This looks like it fetches the profile, sets an account, and fetches settings data in parallel. The original line fetched the profile and then set the account in series. Is the change to parallel fetch intentional? (i.e. if fetched profile data is a dependency of `setAccount`, we may end up with a race condition here)

packages/fxa-content-server/app/scripts/views/settings.js
philbooth: Great, thanks. This is what I was asking about in https://github.com/mozilla/fxa/pull/1846#discussion_r305427026, I had a hunch what I'd done might be wrong.

packages/fxa-content-server/app/scripts/views/settings.js
philbooth: They're all independent operations. The decision to run them concurrently comes from https://github.com/mozilla/fxa/pull/1846#discussion_r306221340.

lmorchard: If it's a problem, maybe this should be something more like:
```javascript
    return Promise.all([
      account.fetchProfile().then(() => this.user.setAccount(account)),
      account.settingsData(),
    ])
      .then(([, data]) => {
```

philbooth: There's no need for `fetchProfile` and `setAccount` to be run serially, unless I'm missing something? They're completely independent.

lmorchard: Yeah, I'm fuzzy on whether there's a real problem here (I didn't see one)

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -107,6 +108,8 @@ const View = BaseView.extend({
     const account = this.getSignedInAccount();
 
     context.set({
+      ccExpired: !!this._ccExpired,
+      escapedCcExpiredLinkAttrs: `href="/subscriptions" class="alert-link"`,
```
lmorchard: I think this needs an explicit trailing `/` - so `/subscriptions/` here. It's a 404 without

packages/fxa-content-server/app/scripts/views/settings.js
lmorchard: Oh yeah, sorry - missed this comment until now :)

packages/fxa-content-server/app/scripts/views/settings.js
philbooth: Thanks.

Tangential, but that seems different to the rest of our routes fwiw. `/signin`, `/signup`, `/settings`, `/sms` etc all work fine. Conscious decision?

lmorchard: Hmm, not a conscious decision, now that you mention it. Maybe the better fix here is to let this link stay as-is and tweak the server.

lmorchard: In fact: https://github.com/mozilla/fxa/pull/1926

PR #1848, opened 2019-07-19
feat(routes): added securityEvents to js-client

farhan787:
@vbudhram I've closed previous PR because that was branch was child of my remote auth-server's branch, please continue with this one.

### DIFF
--- a/packages/fxa-js-client/client/FxAccountClient.js
+++ b/packages/fxa-js-client/client/FxAccountClient.js
@@ -1648,6 +1648,48 @@ define([
   };
 
   /**
+   * Get a list of user's security events
+   *
+   * @method securityEvents
+   * @param {String} sessionToken sessionToken obtained from signIn
+   * @return {Promise} A promise that will be fulfilled with JSON `xhr.responseText` of the request
+   */
+  FxAccountClient.prototype.securityEvents = function(sessionToken) {
+    let request = this.request;
+
+    return Promise.resolve()
+      .then(() => {
+        required(sessionToken, 'sessionToken');
+
+        return hawkCredentials(sessionToken, 'sessionToken', HKDF_SIZE);
+      })
+      .then(creds => {
+        return request.send('/securityEvents', 'GET', creds);
+      });
+  };
+
+  /**
+   * Delete user's security events
+   *
+   * @method deleteSecurityEvents
+   * @param {String} sessionToken sessionToken obtained from signIn
+   * @return {Promise} A promise that will be fulfilled with JSON `xhr.responseText` of the request
+   */
+  FxAccountClient.prototype.deleteSecurityEvents = function(sessionToken) {
+    let request = this.request;
+
+    return Promise.resolve()
+      .then(() => {
+        required(sessionToken, 'sessionToken');
+
+        return hawkCredentials(sessionToken, 'sessionToken', HKDF_SIZE);
+      })
+      .then(creds => {
+        return request.send('/securityEvents', 'DELETE', creds);
+      });
+  };
+
+  /**
    * Get a list of user's attached clients
    *
    * @method attachedClients
--- a/packages/fxa-js-client/tests/lib/securityEvents.js
+++ b/packages/fxa-js-client/tests/lib/securityEvents.js
@@ -1,0 +1,100 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'intern!tdd',
+  'intern/chai!assert',
+  'tests/addons/environment',
+  'tests/addons/sinon',
+], function(tdd, assert, Environment, sinon) {
+  with (tdd) {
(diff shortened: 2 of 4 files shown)


packages/fxa-js-client/tests/lib/securityEvents.js
```diff
@@ -0,0 +1,86 @@
...
+        ).then(res => {
+          assert.equal(xhrOpen.args[0][0], 'DELETE', 'method is correct');
+          assert.include(
+            xhrOpen.args[0][1],
+            '/securityEvents',
+            'path is correct'
+          );
+          assert.ok(res, 'got response');
+          assert.deepEqual(res, {});
+        }, assert.notOk);
```
vbudhram: After the delete here, try to fetch the security events, it should be 0.

packages/fxa-js-client/tests/mocks/request.js
```diff
@@ -223,6 +227,21 @@ define(['client/lib/errors', 'tests/lib/push-constants'], function(
       status: 200,
       body: '{"subscriptions":[{"foo":"bar"}]}',
     },
+    securityEvents: {
```
farhan787: @vbudhram see this, IMHO we can't call again securityEvents route after deleting securityEvents because this mock request will return this `securityEvents` object.

packages/fxa-js-client/tests/lib/securityEvents.js
farhan787: @vbudhram [see this](https://github.com/mozilla/fxa/pull/1848/files#r305920652)

packages/fxa-js-client/tests/mocks/request.js
vbudhram: @farhan787 you can create a new mock result and use that. Ex. `securityEventsEmptyResponse`

farhan787: Ok gotcha !

packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -1647,6 +1647,47 @@ define([
...
+      .then(creds => {
+        return request.send('/securityEvents', 'GET', creds);
+      });
+  };
+
+  /**
+   * Delete user's security events
+   *
+   * @method deleteSecurityEvents
+   * @param {String} sessionToken sessionToken obtained from signIn
```
vbudhram: nit: This also returns a promise when completed.

packages/fxa-js-client/tests/mocks/request.js
```diff
@@ -223,6 +227,25 @@ define(['client/lib/errors', 'tests/lib/push-constants'], function(
...
+        {
+          name: 'account.login',
+          verified: true,
+          createdAt: new Date().getTime(),
+        },
+      ]),
+    },
+    securityEventsEmptyResponse: {
+      status: 200,
+      body: '[]',
```
vbudhram: This returns response `{}` instead of an array, ref https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/security-events.js#L38.

packages/fxa-js-client/tests/mocks/request.js
```diff
@@ -26,6 +26,10 @@ define(['client/lib/errors', 'tests/lib/push-constants'], function(
       headers: {},
       body: '{}',
     },
+    deleteSecurityEvents: {
+      status: 200,
+      body: '{}',
```
vbudhram: Just noticed this, but it is actually an empty array `[]` instead of empty response. Can you update the test and mock.

farhan787: No problem but it was an empty array [before](https://github.com/mozilla/fxa/pull/1848#discussion_r306831951)

farhan787: @vbudhram done.

packages/fxa-js-client/tests/mocks/request.js
```diff
@@ -223,6 +227,25 @@ define(['client/lib/errors', 'tests/lib/push-constants'], function(
...
+          createdAt: new Date().getTime(),
+        },
+        {
+          name: 'account.login',
+          verified: true,
+          createdAt: new Date().getTime(),
+        },
+      ]),
+    },
+    securityEventsEmptyResponse: {
```
vbudhram: @farhan787 Sorry I meant `securityEventsEmptyResponse` should return `[]`.

packages/fxa-js-client/tests/mocks/request.js
```diff
@@ -26,6 +26,10 @@ define(['client/lib/errors', 'tests/lib/push-constants'], function(
       headers: {},
       body: '{}',
     },
+    deleteSecurityEvents: {
+      status: 200,
+      body: '[]',
```
vbudhram: We can revert this back to `{}`.

packages/fxa-js-client/tests/mocks/request.js
farhan787: @vbudhram I was also thinking that `securityEventsEmptyResponse` should return [].

PR #1855, opened 2019-07-19
feat: reduce build time for fxa-email-service

bbangert:
This is now fine to review. The fxa-email-service build needs a careful review to verify that its structure matches and works correctly given the Dockerfile build is a bit different.

`install-content-server` is substantially faster as all the packages are installed concurrently using the npm concurrently package. This reduces `npm install` in the root dir from 8:45 to 2:24. The `test-content-server` is now the longest step, but reducing that should drastically reduce overall build time now.

There's also GNU `parallel`, but since concurrently was trivial to include as a root dependency, I added it there for use in the post-install build script.


.circleci/config.yml
```diff
@@ -237,12 +239,41 @@ jobs:
...
+          cp target/release/fxa_email_send bin
+          cp target/release/fxa_email_queues bin
+          cargo clean
       - store_artifacts:
           path: fxa-auth-db-mysql.log
       - setup_remote_docker
       - run: |
-          ../../.circleci/build.sh fxa-email-service
-          ../../.circleci/deploy.sh fxa-email-service
+          ../../.circleci/tag.sh fxa-email-service
```
philbooth: How come both this and `fxa-email-service-tag` do the release build and tag stuff? Does it need to be here, or would it be sufficient for this job to stop after `cargo audit`?

.circleci/tag.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash -e
```
philbooth: It looks like this script has a lot of overlap with `.circleci/deploy.sh`. Is it possible to eliminate the duplication, or perhaps even to eliminate one of the scripts with params or whatever?

.circleci/config.yml
bbangert: Mainly to match prior behavior where it was build and deployed in both places. I'm not sure why it was setup for that either.

.circleci/tag.sh
bbangert: There is, but I didn't see a clear way to pull apart the steps, so I made a new script that merged the desired steps and checked for a new Dockerfile-tag that needs to function differently.

.circleci/config.yml
philbooth: Nvm, I just remembered why. It's so that `feature.xxx` and `dockerpush.yyy` branches pushed to dockerhub.

bbangert:
@philbooth I have built this locally and verified the container contents match with the files in the same places.

PR #1860, opened 2019-07-19
fix(package): allow ../ references to fxa-shared and fxa-js-client

philbooth:
This is my attempt to replace the `fxa-js-client` and `fxa-shared` npm modules in the content server with in-tree `../` references. The objective is to stop us having to independently version bump and then npm publish those modules, as it slows down development when working on changes that span packages.

I've followed the same approach that @jbuck took for `fxa-payments-server`, explicitly copying the extra directories into the Docker image alongside `/app`. It passes the tests in CI but isn't working when I deploy to `fxa-dev`, I still need to figure out what's up with that.

Opening as a draft so I don't lose track of it and can come back to it when we have more time. I was hoping to get it landed this week but I probably already spent more time on it than I should have.


packages/fxa-auth-server/lib/metrics/amplitude.js
```diff
@@ -192,6 +193,7 @@ module.exports = (log, config) => {
               emailService: data.email_service,
               emailTypes: EMAIL_TYPES,
               service: getService(request, data, metricsContext),
+              version: VERSION,
```
philbooth: These changes to the auth server weren't strictly necessary yet, because it still uses the old `fxa-shared` version from npm. But they have no effect on it either, and mean it will automatically continue to work if/when we upgrade the auth server to import `../fxa-shared` too.

packages/fxa-content-server/Dockerfile-build
```diff
@@ -4,23 +4,34 @@ RUN npm install -g npm@6 && rm -rf ~app/.npm /tmp/*
...
+
 WORKDIR /app
 
 # S3 bucket in Cloud Services prod IAM
 ADD https://s3.amazonaws.com/dumb-init-dist/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
 RUN chmod +x /usr/local/bin/dumb-init
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
 
-RUN apk add --no-cache git
+RUN apk add --no-cache git make perl
```
philbooth: These extra deps were necessary for `grunt sjcl` to succeed.

packages/fxa-content-server/app/scripts/lib/constants.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const popularEmailDomains = require('./popular-email-domains.json');
+const popularEmailDomains = require('../../../../fxa-shared/email/popularDomains.json');
```
philbooth: `popular-email-domains.json` was a symlink. We could have kept it around but it seemed unnecessary now we can just pull in the actual file directly.

packages/fxa-content-server/webpack.config.js
```diff
@@ -170,17 +173,17 @@ const webpackConfig = {
...
-                ],
-                '@babel/preset-env', 
-                '@babel/preset-typescript'
+                ['@babel/preset-react', {}],
+                '@babel/preset-env',
+                '@babel/preset-typescript',
+              ],
+              plugins: [
+                '@babel/syntax-dynamic-import',
+                '@babel/plugin-proposal-class-properties',
```
philbooth: Not me, `prettier`.

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -4,13 +4,6 @@
 
 'use strict';
 
-let APP_VERSION;
-try {
-  APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(
-    require('../../../package.json').version
```
philbooth: This `require` statement used to assume we were being invoked from inside the `node_modules` directory. Now that's not the case, we needed another way to access the version number of the consumer. Hence passing it in as data below, and the equivalent changes in `fxa-content-server` and `fxa-auth-server`.

philbooth:
This is the error in `fxa-dev`:

Screenshot of a content server error saying `t.default is not a constructor`

Presumably it's coming from this line:

https://github.com/mozilla/fxa/blob/3bccd57a8712ae9cd6e68d6d076b4feb19a8bcfb/packages/fxa-content-server/app/scripts/lib/fxa-client.js#L148

This was the config I used for `fxa-dev`:

https://github.com/mozilla/fxa-dev/compare/deps?expand=1

LZoog:
I took a look at this PR because I'm having to move some stuff into `fxa-shared` for #1923.

I can't help but think that there has got to be a better way to declare the path to the other packages. I guess there's not some global entry point per package that's always hit (even in tests) where we could set a symlink for an entire package or set the path as a const somewhere, removing at least some redundancy? Relative paths that start with `../../../../` give me a headache, but it does look like we do this frequently at least in tests.

philbooth:
I have AWS creds again now, so I'll try to get this finished this week.

philbooth:
Closing this, going to bring it back just with `fxa-geodb` and `fxa-shared`. Including `fxa-js-client` will be easier after #2039 lands.

PR #1865, opened 2019-07-22
Fix flaky OAuth test

### LINKED ISSUE #1837: Intermittent oauth server test failure:  `POST /authorized-clients should seperately list different refresh tokens from the same client`
Maybe 1 in 3 or 1 in 4 times I run the oauth server tests, this happens from api.js:

```
 1) /v1
       /authorized-clients
         POST /authorized-clients
           should seperately list different refresh tokens from the same client:

      AssertionError: expected [ 'other', 'scope' ] to deeply equal [ 'profile' ]
      + expected - actual

       [
      -  "other"
      -  "scope"
      +  "profile"
       ]
```

rfk:
Fixes #1837, in which a test would sometimes pass if two tokens had the same timestamp (causing them to be sorted by scope values) and would sometimes fail if they did not. The scope values now produce the same sort order as the timestamps.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorized-clients/list.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorized-clients/list.js
@@ -127,10 +127,10 @@ module.exports = {
         return -1;
       }
       // To help provide a deterministic result order to simplify testing, also sort of scope values.
-      if (a.scope < b.scope) {
+      if (a.scope > b.scope) {
         return 1;
       }
-      if (a.scope > b.scope) {
+      if (a.scope < b.scope) {
         return -1;
       }
       return 0;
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorized-clients/list.js
```diff
@@ -127,10 +127,10 @@ module.exports = {
         return -1;
       }
       // To help provide a deterministic result order to simplify testing, also sort of scope values.
-      if (a.scope < b.scope) {
```
rfk: These were accidentally sorting in reverse order, which seems needlessly confusing.

philbooth:
Test failure is `cargo audit` in the email service, having a look at it now...

philbooth:
> Test failure is `cargo audit` in the email service, having a look at it now...

*...eventually gets to @rfk's comment in the `crossbeam` issue thread, realises he's way behind...*

philbooth:
> Test failure is `cargo audit` in the email service, having a look at it now...

https://github.com/mozilla/fxa/pull/1866

rfk:
(rebased in the hope of unsticking CI)

PR #1866, opened 2019-07-22
chore(ci): ignore RUSTSEC-2019-0011

philbooth:
Currently builds are failing in CI because of a rustsec advisory. We can ignore it because the offending function is not called by the parent dependency in our tree, `crossbeam-epoch`. Ref: https://github.com/crossbeam-rs/crossbeam/issues/401#issuecomment-513679197

Note there is also a preliminary commit here where I updated rust and all the deps too. I started down that path on autopilot, before looking at the failing advisory, and didn't realise it wouldn't fix the issue. Figured we may as well keep the updates around anyway. There were some knock-on changes required in our codebase, which I'll comment on inline.

@mozilla/fxa-devs r?


packages/fxa-email-service/src/db/core/mod.rs
```diff
@@ -33,18 +36,20 @@ use crate::{settings::Settings, types::error::AppResult};
 /// as a safeguard against leaking PII.
 #[derive(Debug)]
 pub struct Client {
-    client: RedisClient,
+    client: Mutex<RedisClient>,
```
philbooth: Methods on `redis::Client` are now `&mut self`, which means we either need to bubble `&mut self` up through our own callstack, or add interior mutability. I opted for the latter as a lower-impact change.

`Cell` was not an option though, because this struct is passed around as Rocket-managed state so it has to be threadsafe. There's no benefit to `RwLock` over `Mutex` in this case because all the usage requires `mut`. Ergo `Mutex`.

philbooth: Actually, I changed my mind while typing that comment. Bubbling the mutability seems simpler than using a mutex, I'll have a look at it...

packages/fxa-email-service/src/lib.rs
```diff
@@ -27,7 +27,6 @@
 #![feature(decl_macro)]
 #![feature(proc_macro_hygiene)]
 #![feature(plugin)]
-#![feature(type_alias_enum_variants)]
```
philbooth: One more experimental feature ticked off the list #winning

packages/fxa-email-service/src/queues/mod.rs
```diff
@@ -36,7 +36,7 @@ pub struct Queues {
 }
 
 /// An incoming bounce/complaint queue.
-pub trait Incoming: Debug + Sync {
+pub trait Incoming: Debug + Send + Sync {
```
philbooth: I'm not sure which update caused it, but these need to be both `Send` and `Sync` now. Previously we compiled ok with just `Sync`.

packages/fxa-email-service/src/db/core/mod.rs
philbooth: > Bubbling the mutability seems simpler than using a mutex, I'll have a look at it...

It doesn't buy us anything actually, because this is all stuff that gets passed around as `rocket::State<...>`. So it would be `rocket::State<Mutex<...>>` instead and we'd just be pushing the problem up to a higher level.

philbooth:
> perhaps someone with more rust experience should have a look too

Landing it now to fix CI. I'm happy to make changes in follow-ups if someone with more Rust experience wants me to.

Issue #1862, opened 2019-07-22
[support-panel] valid user without totp enabled shows as `Unable to fetch user`


jrgm:
This is because `/totp/<uid>` returns 404 is no totp token is available and https://github.com/mozilla/fxa/blob/master/packages/fxa-support-panel/lib/api.ts#L95-L98 treats that as an error and bails out of any other response.

So, expect 404 where the API is returning normal response (most users do not have totp tokens).

philbooth:
My bad @bbangert, I should have spotted both this and #1861 in code review, sorry!

bbangert:
Closed by #1873.

PR #1873, opened 2019-07-22
fix(support-panel): handle totp returning a 404, and use enabled column

### LINKED ISSUE #1861: [support-panel] user with totp enabled shows as not enabled
I think this line here - https://github.com/mozilla/fxa/blob/master/packages/fxa-support-panel/lib/api.ts#L109

The docs say it's `enable`, but the code just returns to table row, which is `enabled`. That also matches what I see on the wire. - https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-db-mysql/lib/db/mysql.js#L1433

bbangert:
Closes #1861, #1862


jrgm:
lgtm

PR #1879, opened 2019-07-22
feat(payments): extend payment server session from 15 min to 30 min

### LINKED ISSUE #1694: Handle payment server token expiration more gracefully

6a68:
Fixes #1694.

I think I've hit all the relevant configuration bits in the code, including the tests. Didn't see anything in the payments-server that seemed to need to be updated.

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/lib/payment-server.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/payment-server.js
@@ -19,7 +19,7 @@ describe('lib/payment-server-redirect', () => {
       subscriptions: {
         managementClientId: 'MOCK_CLIENT_ID',
         managementScopes: 'MOCK_SCOPES',
-        managementTokenTTL: 900,
+        managementTokenTTL: 1800,
         managementUrl: 'http://example.com',
       },
     };
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -625,7 +625,7 @@ const conf = (module.exports = convict({
       format: String,
     },
     managementTokenTTL: {
-      default: 900,
+      default: 1800,
       doc:
         'OAuth token time-to-live (in seconds) for subscriptions management pages',
       env: 'SUBSCRIPTIONS_MANAGEMENT_TOKEN_TTL',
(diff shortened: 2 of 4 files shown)


packages/fxa-content-server/app/tests/spec/lib/payment-server.js
```diff
@@ -19,7 +19,7 @@ describe('lib/payment-server-redirect', () => {
       subscriptions: {
         managementClientId: 'MOCK_CLIENT_ID',
         managementScopes: 'MOCK_SCOPES',
-        managementTokenTTL: 900,
+        managementTokenTTL: 1800,
```
lmorchard: You probably don't *need* to update this in all the tests, since expiration is just a mock value and isn't *really* exercised time-wise.

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -625,7 +625,7 @@ const conf = (module.exports = convict({
       format: String,
     },
     managementTokenTTL: {
-      default: 900,
+      default: 1800,
```
lmorchard: We *might* want to double check with ops to be sure that there is no `SUBSCRIPTIONS_MANAGEMENT_TOKEN_TTL` configured in any environment and that this default is used. (I'm pretty sure we're fine here - I don't think this got configured unless we asked for it to be done.)

6a68: Just verified with @jrgm in Slack that the `SUBSCRIPTIONS_MANAGEMENT_TOKEN_TTL` isn't set via env var, ops just takes the defaults. Good catch, though :+1:

jrgm: Answered 6a68 in slack channel, but no stage/prod do not configure an override of the default SUBSCRIPTIONS_MANAGEMENT_TOKEN_TTL

6a68:
Rebased against current master. Hopefully the tests will now pass

PR #1882, opened 2019-07-22
fix(payments): update styles on payment success alert

### LINKED ISSUE #1719: Notification bar visual design
Currently, it looks like this:
Screen Shot 2019-07-09 at 2 24 17 PM

Please change it to this: 
Screen Shot 2019-07-09 at 2 25 10 PM

(see code: https://zpl.io/2ZDwB4X)

meandavejustice:
- fixes #1719

success-alert

https://app.zeplin.io/project/5d13e5c151e1e056b3f68fa7/screen/5cf013bf6af3e11bb4550a13


packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -79,6 +79,11 @@ export const Subscriptions = ({
     }
   }, [ fetchSubscriptionsRouteResources, accessToken ]);
 
+  let showPaymentSuccessAlert = true;
```
lmorchard: Should use a [`useState` hook](https://reactjs.org/docs/hooks-reference.html#usestate) to track this state, and a [`useCallback` hook](https://reactjs.org/docs/hooks-reference.html#usecallback) to wrap setting the state with `dismissAlert`. Otherwise, this will get set to `true` and the callback will be redefined every every time the component renders.

PR #1898, opened 2019-07-22
chore(deps): Restore `keypair` dependency

rfk:
I attempted to remove this dependency in #1668 in favour of node's new built-in functions that do the same thing, but clearly failed to find all the uses of it, because CI is now busted. Restoring the dep, with apologies for not catching this before merge.


rfk:
@vladikoff any chance you can understand what's happening with the `NoSuchDriver` content-server errors in https://circleci.com/gh/mozilla/fxa/48044#queue-placeholder/containers/5 ?

rfk:
I've retried the job a few times, but I just cannot get a clean run of content-server tests in CI. @mozilla/fxa-devs anyone able to advice how to push through this situation?

PR #1900, opened 2019-07-23
chore(config): Add FireTV to the validation list for "oldsync" scope.

rfk:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1566346

### DIFF
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -498,6 +498,7 @@ const conf = (module.exports = convict({
             'https://lockbox.firefox.com/fxa/android-redirect.html',
             'https://accounts.firefox.com/oauth/success/a2270f727f45f648', // Fenix
             'https://accounts.firefox.com/oauth/success/3c49430b43dfba77', // Reference browser
+            'https://accounts.firefox.com/oauth/success/85da77264642d6a1', // Firefox for FireTV
             'urn:ietf:wg:oauth:2.0:oob:pair-auth-webchannel',
           ],
         },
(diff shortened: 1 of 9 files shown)


packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -498,6 +498,7 @@ const conf = (module.exports = convict({
             'https://lockbox.firefox.com/fxa/android-redirect.html',
             'https://accounts.firefox.com/oauth/success/a2270f727f45f648', // Fenix
             'https://accounts.firefox.com/oauth/success/3c49430b43dfba77', // Reference browser
+            'https://accounts.firefox.com/oauth/success/85da77264642d6a1', // Firefox for FireTV
```
shane-tomlinson: Having *this* config here and all other config in cloudops-deployment is a recipe for confusion and disaster! What do we need to do to get this out of here and into cloudops-deployment instead?

shane-tomlinson:
@rfk - should we target this as a point release for train-141 so @mcomella can test?

packages/fxa-content-server/server/lib/configuration.js
shane-tomlinson: Opened https://github.com/mozilla/fxa/issues/1903

rfk:
> should we target this as a point release for train-141 

Sure; I assume it will then also need to be forward-ported to 142, and then master. I can wrangle that today.

rfk:
I've changed this to target train-141, but had to pick up https://github.com/mozilla/fxa/pull/1866 in order to get CI (hopefully...) passing.

rfk:
CI now seems to be failing on the language change from https://github.com/mozilla/fxa/pull/1870; I cherry-picked the fix from that PR but it doesn't seem to help, and I don't have any more bandwidth to keep poking at this. @shane-tomlinson how should I unstick it?

rfk:
Woohoo, CI passing! @shane-tomlinson please merge and cut this if it all good to you, otherwise I'll revisit in the morning.

Issue #453, opened 2019-03-27
Accept OAuth client credentials as Basic Auth header in /oauth/token


rfk:
Followon from https://github.com/mozilla/fxa-auth-server/pull/2983

The OAuth "token endpoint" needs to be able to accept the OAuth client id and secret as either as  request body parameter, or as a HTTP basic auth header.  We have at least one integration in the wild that currently does it that way so we'll have to add support for that before they can start using the auth-server version of this route.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-644)
┆Issue Number: FXA-644

PR #1901, opened 2019-07-23
chore(config): Advertize auth-server /oauth/token endpoint in OIDC config

rfk:
We intend for the auth-server endpoint to replace the oauth-server version, so might as well start telling clients to use it.

@shane-tomlinson any reason we wouldn't want to do this?

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/get-openid-configuration.js
+++ b/packages/fxa-content-server/server/lib/routes/get-openid-configuration.js
@@ -9,7 +9,7 @@ const config = require('../configuration');
 const authorizationEndpoint = config.get('public_url') + '/authorization';
 const issuer = config.get('public_url');
 const jwksEndpoint = config.get('oauth_url') + '/v1/jwks';
-const tokenEndpoint = config.get('oauth_url') + '/v1/token';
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
> @shane-tomlinson any reason we wouldn't want to do this?

Yes, see [this note on the auth server's /oauth/token route](
https://github.com/mozilla/fxa/blob/c2e696500db9715eca8a42660c60c5a016335050/packages/fxa-auth-server/lib/routes/oauth.js#L104:L105).

> XXX TODO: To be able to fully replace the /token route from oauth-server,
> this route must also be able to accept 'client_secret' as Basic Auth in header.

shane-tomlinson:
Ref https://github.com/mozilla/fxa/issues/453 about accepting the `client_secret` in the Basic Auth header.

6a68:
Rebased to pull in circle-CI fixes

6a68:
I think it might be good to close out this vintage bug, and reopen after #453 gets prioritized and fixed. @rfk thoughts?

rfk:
Boom.

6a68:
🍻

PR #1916, opened 2019-07-23
feat(subscriptions): show subscriptions management Settings subpanel …

### LINKED ISSUE #1078: Subscription management button should only appear when user has at least one active subscription

chenba:
…only when there are active subscriptions

This patch also removes everything related to language check for
subscriptions management.  See https://github.com/mozilla/fxa/issues/1078#issuecomment-509296642

Fixes #1078

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/subscription.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/subscription.js
@@ -11,7 +11,6 @@ import Template from 'templates/settings/subscription.mustache';
 import PaymentServer from '../../lib/payment-server';
 
 const View = FormView.extend({
-  template: Template,
   className: 'subscription',
   viewName: 'settings.subscription',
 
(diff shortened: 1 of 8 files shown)


packages/fxa-content-server/app/scripts/views/settings/subscription.js
```diff
@@ -26,6 +25,36 @@ const View = FormView.extend({
...
+    });
+  },
+
+  /**
+   * Check to see if the account has any active subscriptions.
+   *
+   * @returns {Promise} resolves to a bool.
+   */
+  _hasActiveSubscriptions() {
+    return new Promise((resolve, reject) => {
```
lmorchard: I don't think you need to wrap the Promise code here. Unless I'm missing something, this should be enough:
```javascript
_hasActiveSubscriptions() {
  return this.getSignedInAccount()
    .fetchActiveSubscriptions()
    .then(subscriptions => subscriptions.length > 0);
}
```

lmorchard: Also, it might be a good idea someday to use the new account data API resource. But, we can do that as a followup and not block this PR.

chenba: Updated the promise code.  Thanks!

I was not aware of the new account data API.  Where should I look?

chenba: Filed https://github.com/mozilla/fxa/issues/1921

PR #1918, opened 2019-07-24
WIP add support for ES256 JWT signatures.

rfk:
@shane-tomlinson I did a little spike on integrating `ES256` JWT signatures, here's my very very rough code. Connects to #1769 

The hardest part here is that `pem-jwk` only supports RSA keys. I've switched to using `node-jose` for translating between PEM and JWK here, but the APIs it exposes for that are Promise-based, so that resulted in some flow-on changes and some "XXX TODO" notes. There may be a better way to achieve what we need here.

I *think* this works; at least the tokens it produces can be verified successfully using https://jwt.io. Definitely a lot of cleanup required before we could use this for real, but hopefully it will help unblock you.  I don't think I'll have any more time to work on it this week, but you're very welcome to pick it up or scavenge it for parts.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/keys.js
@@ -4,75 +4,138 @@
 
 const assert = require('assert');
 const config = require('./config');
-const { jwk2pem, pem2jwk } = require('pem-jwk');
 const crypto = require('crypto');
+const jose = require('node-jose');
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Thanks @rfk, I'll try to get this across the line once I finish prompt=none

shane-tomlinson:
Closing this (but not deleting the branch), until we prioritize using ES256 for JWT signatures.

PR #1919, opened 2019-07-24
chore(docs): auto-generate the api docs again

### LINKED ISSUE #1845: [auth server] Either reinstate the api docs script, or delete the api docs
In https://github.com/mozilla/fxa/pull/1824#discussion_r304956725, it was pointed out that I forgot to document a new endpoint. Because I'm lazy, I decided to fix it by sneakily bringing back the [docs script](https://github.com/mozilla/fxa-auth-server/pull/2988) just to generate them for that one new endpoint.

The resulting diff highlighted other cases where the docs have not been updated recently, or had gone stale for other reasons; for new endpoints, new validation schemata, new properties in response bodies, new error codes, changes to error codes, assorted refactorings etc and so on.

Of course, as ever, there were also some things that the script got wrong, which was the reason we deleted it in the first place. But the exercise has raised a couple of questions for me:

* Which is worse; documentation that is up to date but has some omissions where an automated script didn't behave correctly, or documentation that hasn't been updated at all and is just plain wrong in some cases?

* If nobody notices the docs are wrong for 3 months, is there any point in having docs? Why are we spending dev time maintaining (or pretending to maintain) something that nobody uses?

Fwiw I'm in favour of deleting the API docs completely, and dropping the pretence that they're somehow accurate or that we actively maintain them. I don't think anybody refers to them, ever, and they're a total waste of time. By definition, they'll always be less accurate than just reading the source code, which is what I think most people do anyway.

In case anyone's interested in the diff, it's here:

https://github.com/mozilla/fxa/compare/pb/get-account-docs?expand=1#diff-b73c5082111d34ac941c5abd568d519f

End of rant! :smile:

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-223)
┆Issue Number: FXA-223

philbooth:
Fixes #1845 (maybe).

Seeing as my mission to delete the auth server API docs failed, I've opted to try out plan B and reinstate the generator script.

I know, I know, I was the one who deleted it in the first place because it kept breaking. But I really think keeping broken documentation around is worse, if people are going to refer to it and use e.g. the wrong values for error codes (which is one of the things the docs are wrong about right now). Docs that are outright wrong like that are worse than useless, they're actively harmful.

So anyway, this brings back the script, but doesn't install it as a pre-commit hook, so it won't be as intrusive when it fails as it was before. People referring to the documentation are free to run it on an ad hoc basis to make sure things are up to date.

Note there are some conflicts with `prettier`, but adding `docs/api.md` to `.prettierignore` didn't seem to affect that. I made some changes to the script to try and harmonise the two, but you'll still see a bigger diff locally when you run it than you do after `prettier` has had its way.

If you want to check anything here against the script in its previous form, you can refer to the PR where it got removed: mozilla/fxa-auth-server#2988

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/.prettierignore
+++ b/packages/fxa-auth-server/.prettierignore
@@ -11,6 +11,7 @@ LICENSE
 *.handlebars
 *.html
 Dockerfile*
+docs/api.md
 test/load/*
 test/scripts/test_output/*
 fxa-auth-db-mysql/*
--- a/packages/fxa-auth-server/docs/api.md
+++ b/packages/fxa-auth-server/docs/api.md
@@ -1,5 +1,9 @@
 # Firefox Accounts authentication server API
 
+This document is automatically generated.
+If you are editing it,
+read [this section](#this-document) first.
+
 <!--begin-abstract-->
 
 This document provides protocol-level details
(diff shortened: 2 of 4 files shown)


bbangert:
Would it be more reliable to add a few extra bits to each hapi route so we could have full Swagger API docs generated?

e.g. https://github.com/glennjones/hapi-swagger

philbooth:
> Would it be more reliable to add a few extra bits to each hapi route so we could have full Swagger API docs generated?

That's cool with me, as long as we're allowed to delete these docs that are actively wrong as part of that.

bbangert:
Agreed, we should only have the generated docs as a result.

PR #1924, opened 2019-07-24
fix(123done): configure stage paymentURL

jrgm:
r? - @lmorchard 

That product code came from Ben.

### DIFF
--- a/packages/123done/static/js/123done.js
+++ b/packages/123done/static/js/123done.js
@@ -14,16 +14,18 @@ $(document).ready(function() {
   let paymentURL;
   switch (window.location.host) {
     case '123done-latest.dev.lcip.org':
-      paymentURL = 'https://latest.dev.lcip.org/subscriptions/products/prod_Ex9Z1q5yVydhyk';
+      paymentURL =
+        'https://latest.dev.lcip.org/subscriptions/products/prod_Ex9Z1q5yVydhyk';
       break;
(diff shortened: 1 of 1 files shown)


jrgm:
Hrm. Did prettify make those other changes?

packages/123done/static/js/123done.js
```diff
@@ -14,16 +14,18 @@ $(document).ready(function() {
...
   switch (window.location.host) {
     case '123done-latest.dev.lcip.org':
-      paymentURL = 'https://latest.dev.lcip.org/subscriptions/products/prod_Ex9Z1q5yVydhyk';
+      paymentURL =
+        'https://latest.dev.lcip.org/subscriptions/products/prod_Ex9Z1q5yVydhyk';
       break;
     case '123done-stage.dev.lcip.org':
-      paymentURL = 'TBD';
+      paymentURL =
+        'https://accounts.stage.mozaws.net/subscriptions/products/prod_FUUNYnlDso7FeB';
```
lmorchard: This URL looks like the right pattern, though the product ID leads to a "no such product" error. Guessing we still need to switch to stage subhub & stripe where this product ID is available

jrgm:
> assuming we still have config to tweak for subhub & stripe

Yes, but that **should** be the final url.

lmorchard:
> Hrm. Did prettify make those other changes?

Also, yeah, probably prettify. Good changes to make though

PR #1927, opened 2019-07-24
fix(support-panel): add gitignore for dist/common files

### LINKED ISSUE #1869: Support panel `dist` directory shows up as not checked in
After I did a full repo update and did a `git status`, I saw this:

> Untracked files:
  (use "git add <file>..." to include in what will be committed)
>
>	packages/fxa-support-panel/dist/

Screenshot 2019-07-22 at 12 24 46

bbangert:
Closes #1869

### DIFF
--- a/packages/fxa-support-panel/.gitignore
+++ b/packages/fxa-support-panel/.gitignore
@@ -1,0 +1,6 @@
+/coverage.html
+/node_modules
+/dist
+*.gpg
+*.log
+npm-debug.log


packages/fxa-support-panel/.gitignore
```diff
@@ -0,0 +1,6 @@
+/coverage.html
+/node_modules
+/dist
+*.gpg
+*.log
+npm-debug.log
```
vladikoff: do you need npm-debug.log, if you already have *.log?

bbangert: Probably not, but that is an artifact from the other packages that was copypasta'd in here too it seems. We should prolly revisit these in all the sub-packages.

vladikoff: that's what i thought!

PR #1929, opened 2019-07-24
fix(support-panel): log totp error, return appropriate errors

### LINKED ISSUE #1863: [support-panel] Log errors as log level `error`
This line https://github.com/mozilla/fxa/blob/master/packages/fxa-support-panel/lib/api.ts#L96 should really be `this.logger.erro(...)`.

Although given https://github.com/mozilla/fxa/issues/1862 this would get noisy if we're treating lack of totp tokens as fatal errors.

bbangert:
Closes #1863

### DIFF
--- a/packages/fxa-support-panel/test/lib/api.spec.ts
+++ b/packages/fxa-support-panel/test/lib/api.spec.ts
@@ -123,15 +123,25 @@ describe('Support Controller', () => {
     cassert.equal(result.statusCode, 200);
   });
 
-  it('gracefully handles 404s', async () => {
+  it('gracefully handles 404s/500', async () => {
     const defaults = createDefaults();
     defaults.account.status = 404;
     mockCalls(defaults);
-    const result = await server.inject({
+    let result = await server.inject({
       method: 'GET',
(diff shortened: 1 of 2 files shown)


packages/fxa-support-panel/test/lib/api.spec.ts
```diff
@@ -123,15 +123,25 @@ describe('Support Controller', () => {
...
+  it('gracefully handles 404s/500', async () => {
     const defaults = createDefaults();
     defaults.account.status = 404;
     mockCalls(defaults);
-    const result = await server.inject({
+    let result = await server.inject({
+      method: 'GET',
+      url: `/?uid=${uid}`
+    });
+    cassert.equal(result.statusCode, 500);
```
chenba: I think this should be `404`.

bbangert: I had considered that originally, but if its unable to fetch URL's that should work with expected status codes, there's an error occurring. It logs an error, and it seems reasonable to indicate a service error with a 500 as well.

PR #1935, opened 2019-07-24
feat(signup): Fix #1723, redirect to original page after signup

### LINKED ISSUE #1723: [payments] Product redirect should complete after signup/confirmation

ianb:
This adds a redirectTo property to the relier, which is already used to add a redirectTo query parameter to verification links, though it wasn't previously being read after signup.

Out of caution, we ensure the redirect can only be to a location local to the content-server (relying party signups go through a different path).

Additionally the confirmation page will redirect based on the in-memory value of `relier.get('redirectTo')`

Re: tests: I wasn't able to figure out how to test the confirmation page redirect. The testing process (which doesn't exactly match the manual process) seems to load a new page, losing the value of redirectTo. There doesn't seem to be any special testing logic about the confirmation page, so I'm not sure why it navigates differently in testing.

We may want to revert part of this patch once payments is a relying party; or maybe this is functionality worth preserving, I'm not sure.

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/auth-errors.js
+++ b/packages/fxa-content-server/app/scripts/lib/auth-errors.js
@@ -578,6 +578,10 @@ var ERRORS = {
     message: t('Accounts with two-step authentication do not support pairing at this time')
   },
    */
+  INVALID_REDIRECT_TO: {
+    errno: 1062,
+    message: t('Invalid redirect'),
+  },
 };
 /*eslint-enable sorting/sort-object-props*/
 
--- a/packages/fxa-content-server/app/scripts/views/base.js
+++ b/packages/fxa-content-server/app/scripts/views/base.js
@@ -381,6 +381,9 @@ var BaseView = Backbone.View.extend({
         err => {
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
             this.logError(AuthErrors.toError('SESSION_EXPIRED'));
+            // The redirectTo in .navigate() is lost if there's later navigations, so by saving it here
+            // we can get it later (e.g., in case of a signUp):
+            this.relier.set('redirectTo', this.window.location.href);
             this.navigate(this._reAuthPage(), {
               redirectTo: this.currentPage,
             });
--- a/packages/fxa-content-server/app/scripts/views/behaviors/navigate-or-redirect.js
+++ b/packages/fxa-content-server/app/scripts/views/behaviors/navigate-or-redirect.js
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A behavior that navigates to the relier's redirectTo value if present, or navigates to a fixed endpoint
+ */
+
+import _ from 'underscore';
+import AuthErrors from '../../lib/auth-errors';
+import NavigationBehavior from './navigate.js';
+
+const NavigateOrRedirectBehavior = function(endpoint, options = {}) {
+  const behavior = function(view, account) {
+    let redirectTo = view.relier.get('redirectTo');
+    if (redirectTo) {
+      redirectTo = new URL(redirectTo, location.href);
+      if (redirectTo.origin !== location.origin) {
+        const err = AuthErrors.toError('INVALID_REDIRECT_TO');
+        this.logError(err);
+        throw err;
+      }
+      redirectTo = redirectTo.pathname + redirectTo.search;
+    }
+    const redirectEndpoint = redirectTo || endpoint;
+    const navigateBehavior = new NavigationBehavior(redirectEndpoint, options);
+    return navigateBehavior(view, account);
+  };
+
+  // used for testing
+  _.assign(behavior, options, {
+    endpoint: endpoint,
+    halt: true,
+    type: 'navigateOrRedirect',
+  });
+
+  return behavior;
+};
+
+export default NavigateOrRedirectBehavior;
--- a/packages/fxa-content-server/tests/functional/sign_up.js
+++ b/packages/fxa-content-server/tests/functional/sign_up.js
@@ -41,12 +41,15 @@ const {
   testUrlInclude,
   type,
   visibleByQSA,
+  waitForUrl,
 } = FunctionalHelpers;
 
 var SIGNUP_ENTRYPOINT = 'entrypoint=' + encodeURIComponent('fxa:signup');
 var SYNC_CONTEXT_ANDROID = 'context=fx_fennec_v1';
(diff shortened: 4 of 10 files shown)


packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -578,6 +578,10 @@ var ERRORS = {
     message: t('Accounts with two-step authentication do not support pairing at this time')
   },
    */
+  INVALID_REDIRECT_TO: {
+    errno: 1062,
+    message: t('Invalid redirect'),
+  },
```
ianb: I feel like there's probably more to be done here to add a new error...

packages/fxa-content-server/app/scripts/views/base.js
```diff
@@ -381,6 +381,9 @@ var BaseView = Backbone.View.extend({
         err => {
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
             this.logError(AuthErrors.toError('SESSION_EXPIRED'));
+            // The redirectTo in .navigate() is lost if there's later navigations, so by saving it here
+            // we can get it later (e.g., in case of a signUp):
+            this.relier.set('redirectTo', window.location.href);
```
ianb: `window.location.href` doesn't seem quite right here, but I'm not sure. `this.currentPage` is another candidate (they'd almost be equivalent, except `this.currentPage` doesn't start with a `/` so it's not really a proper relative URL, but obviously that could be adjusted).

packages/fxa-content-server/app/scripts/views/behaviors/navigate-or-redirect.js
```diff
@@ -0,0 +1,40 @@
...
+  const behavior = function(view, account) {
+    let redirectTo = view.relier.get('redirectTo');
+    if (redirectTo) {
+      redirectTo = new URL(redirectTo, location.href);
+      if (redirectTo.origin !== location.origin) {
+        const err = AuthErrors.toError('INVALID_REDIRECT_TO');
+        this.logError(err);
+        throw err;
+      }
+      redirectTo = redirectTo.pathname + redirectTo.search;
```
ianb: Open redirects seem a little scary, which is why I put this extra logic here (and below) to at least make sure the redirect can't leave the server.

I guess they should probably be a single function somewhere, I'll try to fix that tomorrow.

packages/fxa-content-server/tests/functional/sign_up.js
```diff
@@ -480,6 +483,23 @@ registerSuite('signup', {
...
+      return (
+        this.remote
+          .then(openPage(PRODUCT_URL, selectors.SIGNUP.HEADER))
+          .then(fillOutSignUp(email, PASSWORD))
+          .then(testAtConfirmScreen(email))
+          // Note the way getVerificationLink works, it doesn't get the link from the email but instead only gets the
+          // verification code. We have to ask it to add redirectTo=... to recreate the entire URL
+          .then(
+            openVerificationLinkInSameTab(email, 0, {
+              query: { redirectTo: PRODUCT_URL },
```
ianb: What makes me uncomfortable here is that other tests *do not* set redirectTo, and won't see any change in behavior due to this patch (because the verification link works slightly differently in tests). So I'm not super-comfortable that this isn't regressing anything.

packages/fxa-content-server/app/scripts/views/behaviors/navigate-or-redirect.js
ianb: Huh... but where? `app/scripts/lib/validate.js` ? I can't decide.

lmorchard: `validate.js` looks like as good a place as any, IMO

packages/fxa-content-server/app/scripts/views/base.js
lmorchard: `this.currentPage` is probably a good option, if only because [it looks like it does a bit of normalization and strips off leading & trailing slashes](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/router.js#L442)?

philbooth:
> But, is there a sane opportunity somewhere in here to maybe tack on a query parameter indicating whether the redirect resulted from an existing user sign-in or a new user sign-up?

There is precedent for this fwiw. Iirc, we use a `reason=signin` query param in some places already.

packages/fxa-content-server/app/scripts/views/base.js
shane-tomlinson: `this.currentPage` only has the pathname, it doesn't contain any of the origin. Is the origin necessary here?

packages/fxa-content-server/tests/functional/sign_up.js
shane-tomlinson: It looks like `redirectTo` is used when signing into an account to try to get to panels in the settings page such as account recovery, setting up 2FA, or seeing a list of emails. In BigQuery, I ran this query:

```sql
SELECT * FROM `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content_20190726` where jsonPayload.fields.path like '%redirectTo%' LIMIT 100
```

It seems like more explicit tests could be added by opening the signin/signup page with a `redirectTo` of some internal page, maybe the setup 2FA page or something. When the user completes signing in/up, see if the user ends up where we expect them to be.

packages/fxa-content-server/app/scripts/views/base.js
philbooth: > It also seems like we'd need to propagate the current `flowId` and `flowBeginTime` so the flow metrics behave as expected. @philbooth should be able to provide guidance there.

They're both already included in the resume token, is that being used for the redirect you're talking about here? Should just work if so.

packages/fxa-content-server/app/scripts/views/behaviors/navigate-or-redirect.js
ianb: Well... I ended up taking out one of the instances of redirection (instead always using behaviors) so I guess since there's only once instance reuse is a moot point, right? 🤷‍♂

ianb:
I removed some of the logic and instead used behaviors, for a slight simplification (and to be clearer about when a redirect is allowed). We'll see how the tests go. I think there's a couple comments here on the PR that I have to look at more closely, but pushing what I have now.

6a68:
Rebased to pick up circle-ci fixes

ianb:
I was going to wait for an r+ from @shane-tomlinson, but I think he's out tomorrow, and so I think it would be best to just get it landed. I think it's all OK...?

PR #1936, opened 2019-07-24
feat(event-broker): extract region from service notification url

### LINKED ISSUE #1763: event-broker: make the SQS region a config option and use it when constructing `new SQS()`
Still bootstrapping, and I have a workaround, but probably should be considered for fix.

Starting up in stage I get errors for `notificationProcessor.consumerError` due to `"err":"SQSError: SQS receive message failed: Missing region in config"`.

Sorry, my typescript isn't great, but we pass in a `new SQS()`, ya? 

In https://github.com/bbc/sqs-consumer/blob/master/src/consumer.ts#L120-L122, the Consumer takes the injected object;  otherwise, it would create a default SQS, but make sure to set a region.

A simple workaround is to set `AWS_REGION="us-east-1"` in the container environment. But, perhaps, `region` should be a config parameter of the event-broker so that it's more explicit in the code (as opposed to a magic environment variable when deployed).

bbangert:
Closes #1763

### DIFF
--- a/packages/fxa-event-broker/bin/worker.ts
+++ b/packages/fxa-event-broker/bin/worker.ts
@@ -34,11 +34,23 @@ async function main() {
     db
   );
   const pubsub = new PubSub();
+
+  // Extract region for SQS object
+  const serviceNotificationQueueUrl = Config.get('serviceNotificationQueueUrl');
+  const matchResult = serviceNotificationQueueUrl.match(/(?:.*\/sqs\.)([^\.]+)/i);
+  if (!matchResult || matchResult.length !== 2) {
+    logger.error('invalidServiceUrl', {
+      message: 'Cant find region in service url',
+      serviceNotificationQueueUrl
+    });
+    process.exit(8);
+  }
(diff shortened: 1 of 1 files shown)


packages/fxa-event-broker/bin/worker.ts
```diff
@@ -34,11 +34,23 @@ async function main() {
     db
   );
   const pubsub = new PubSub();
+
+  // Extract region for SQS object
+  const serviceNotificationQueueUrl = Config.get('serviceNotificationQueueUrl');
+  const matchResult = serviceNotificationQueueUrl.match(/(?:.*\/sqs\.)([^\.]+)/i);
```
lmorchard: Might be nice to wrap this in a little utility function and unit test it, but meh 🤷‍♂

bbangert: Good point, mainly cause its regex and it definitely took me some fiddling in a web console before it worked right. I'll file an issue to add that because it does feel precarious to let it hang out there like this.

PR #1952, opened 2019-07-25
fix(copy): Added password and went to Title Case

### LINKED ISSUE #1951: Rename "Logins" datatype to "Logins and Passwords"
[Bug 1568979](https://bugzilla.mozilla.org/show_bug.cgi?id=1568979) requests that we do this in the native desktop Sync Preferences, so I figure we should make the same change in ChooseWhatToSync on the web for consistency.

ryanfeeley:
On Choose What to Sync page I appended " and Passwords" to "Logins" and went to Title Case for the whole list.

Fixes #1951.


6a68:
Related to https://bugzilla.mozilla.org/show_bug.cgi?id=1568979

PR #1969, opened 2019-07-26
fix(fxa-content-server): Fix horizontal bar caused by settings header by changing width unit

### LINKED ISSUE #1968: [content-server] Settings header causing horizontal scrollbar
There is a horizontal scrollbar on the settings page. The fix took me about a minute to figure out so while this is "unimportant," I'm going to quickly follow up with a PR. 

image

LZoog:
fixes #1968

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -62,7 +62,7 @@ body.settings #stage .settings {
   align-items: center;
   align-self: center;
   display: flex;
-  width: 100vw;
+  width: 100%;
 
   @include respond-to('big') {
     height: 96px;


6a68:
I've been poking at these failures on circle-CI for a while now. I tried sshing into the circle-ci VM, setting the intern to verbose logging, then re-running tests, and found a pretty consistent content process crash happening in Firefox: https://gist.github.com/6a68/eadb87dad6f9a56e0ea1b30062cc9c2c . This is what's actually happening behind the scenes when the `NoSuchDriver` failure is shown in the circle-ci output.

I've tried running just the failing reset_password tests from the circle VM, and the error doesn't occur, so there seems to be something about this particular point in the suite's execution that's causing the crash. 🤔 

Just to see what will happen, I've opened #1974 to see if the same CI failures will occur if I revert the commit in this branch, effectively resetting to the same state as master.

shane-tomlinson:
This may be an OOM error, we had OOM problems when running on Teamcity and fixed it by forcing Firefox to clear its memory on occasion. See https://github.com/mozilla/fxa/blob/07c791df2da70b90e52bb0a413c539736cfdea78/packages/fxa-content-server/tests/functional/lib/helpers.js#L2191:LL2204

LZoog:
Thanks for looking into this. I'd actually ran Circle CI twice on this PR and I saw failures on different tests in the content server. I'm rerunning them now to see if we hit it again this morning.

@shane-tomlinson, if that was the case, would you expect hooking in a `.then(cleanMemory())` in functional tests `beforeEach` to resolve the weird functional test failures? If this fails again, I'll try sticking it where I see the failure, it just seems so sporadic that we'd likely need more testing...

LZoog:
It looks like they succeeded this time. If these issues keep up this week, let's keep Shane's suggestion in mind.

PR #1971, opened 2019-07-26
fix(fxa-content-server): dump scopedKeys.validation at startup

### LINKED ISSUE #1970: [content] dump `scopedKeys.validation` at startup
For config audit.

jrgm:
Fixes #1970 

r? - @vladikoff


jrgm:
A miracle. Tests passed after retrying, but lots of skips, and failed 3 times previously. But taking my window and merging this now.

PR #1990, opened 2019-07-29
chore(test): Update to Firefox 67 and intern@4.4.3

shane-tomlinson:
Let's see if this works!


packages/fxa-content-server/package.json
```diff
@@ -155,7 +155,7 @@
         "grunt-todo": "0.5.0",
         "htmlparser2": "3.9.0",
         "install": "0.12.1",
-        "intern": "4.3.1",
+        "intern": "^4.4.3",
         "jsqr": "1.2.0",
         "leadfoot": "1.7.4",
```
vladikoff: might need to bump this too here

chenba:
🤞

packages/fxa-content-server/package.json
chenba: That is the latest leadfoot on npm.  `intern` itself uses `@theintern/leadfoot`, which is much newer.

vladikoff: should switch to `@theintern/leadfoot` here then

.circleci/test-content-server.sh
```diff
@@ -49,7 +49,7 @@ if grep -e "$MODULE" -e 'all' $DIR/../packages/test.list; then
   # ensure email-service is ready
   check 127.0.0.1:8001/__heartbeat__
   cd packages/fxa-content-server
-  mozdownload --version 58.0 --destination firefox.tar.bz2
+  mozdownload --version 67.0 --destination firefox.tar.bz2
```
6a68: @shane-tomlinson Is there some reason we wouldn't use `latest` here, instead of a specific version number?

shane-tomlinson: The biggest reason is that sometimes digdug/theintern do not work with the latest and greatest.

Issue #1734, opened 2019-07-10
[Firefox For Fire TV] Meta Bug


Baron-Severin:
This issue tracks usability issues discovered when logging in on Fire TV devices.

davismtl:
Making email first login flow work well on Fire TV. Input with a TV remote is very different from mouse, keyboard and touch device.

davismtl:
@dannycoates from legal:

In regards to this quote from you:
“the TV interface is sort of like tab navigation but with more “stops” since there’s no cursor. With the TOS between the input and button the user must navigate over those links before getting to the button, and they must do that twice when creating an account with email-first. The TV team doesn’t like that (i agree). Moving them below the button would make the common flow more streamlined. In the interim I’ve set the links as tabindex=-1 which skips over them, but now those links are unreachable on TV”

- TOS and PN can be on the 2nd screen only. Not needed on first screen since we don't save that email unless they complete step 2.
- TOS and PN can go IMMEDIATELY after the registration button on screen 2. (anything with a checkbox needs to remain above the button)

davismtl:
Oops, call me a Zenhub n00b. I put the TOS Privacy Notice issue as a blocker. Somehow I'm unable to remove the blocker.

davismtl:
PR to make changes to this will happen today.

There was a request for us to hide registration. How should we approach this?
@ryanfeeley

abodea:
@athomasmoz I just created a new issue **`The back button from Terms of Service doesn't have any functionality and cannot be selected`** see more details here([2221](https://github.com/mozilla/fxa/issues/2221))

davismtl:
- A bit behind right now
- Work done to get to make it work on tv
- Will be circling back to this after the removal of style=trailhead

davismtl:
@dannycoates continuing work on this this week.
@davismtl to update Ashley

davismtl:
Shipped by FireTV yesterday.

Only one small UX bug for us to resolve on our end.

PR #1991, opened 2019-07-29
Fix Fire TV issues

dannycoates:
Fixes all issues of #1734

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/password-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/password-mixin.js
@@ -41,7 +41,6 @@ export default {
    */
   _updateShowPasswordLabel(passwordEls) {
     const $targetEl = this.$(passwordEls);
-    this._addShowPasswordLabel($targetEl);
 
     if ($targetEl.val().length === 0) {
       $targetEl.addClass('empty');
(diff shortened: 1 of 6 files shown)


packages/fxa-content-server/app/scripts/views/mixins/password-mixin.js
```diff
@@ -41,7 +41,6 @@ export default {
    */
   _updateShowPasswordLabel(passwordEls) {
     const $targetEl = this.$(passwordEls);
-    this._addShowPasswordLabel($targetEl);
```
dannycoates: the issue here was the lazy creation of the label which changed the dom while the on-screen-keyboard had focus causing it to lose it's place on the input element. I switched to eager creation but set the style as 'empty'

dannycoates:
now to fix the tests

packages/fxa-content-server/app/tests/spec/views/sign_up.js
```diff
@@ -1141,9 +1141,8 @@ describe('views/sign_up', function() {
       // wait for tooltip
       return p.delay(50).then(() => {
         assert.equal($('.tooltip-suggest').text(), 'Did you mean gmail.com?✕');
-        // there are exactly 3 elements with tabindex in the page (show
-        // password button has not been added to the page).
-        assert.equal($('[tabindex]').length, 3);
```
dannycoates: fragile tests like this make me 🤬

i'd attempt making it better but i don't even know what it's actually testing

PR #1993, opened 2019-07-29
chore(ci): Remove CI config from within packages subdir.

### LINKED ISSUE #1992: .circleci subdirectories left over in packages
Over in https://github.com/mozilla/fxa/pull/1990 I tried to update Firefox from 58.0 to 67.0. Turns out I updated the wrong file, I tried to update packages/fxa-content-server/.circleci/config.yml. That file isn't used, I should have been updating .circleci/test-content-server.sh instead.

Let's remove any .circleci dirs in the packages that aren't used so this doesn't happen again.

shane-tomlinson:
All CircleCI config is now defined in the top level .circleci directory.

fixes #1992


lmorchard:
Capture

🎉

PR #1995, opened 2019-07-29
fix(tests): stop using `before` in support form functional tests

### LINKED ISSUE #1985: support form functional test uses `before` method meaning it's *always* run
`before` in the Intern is a bit of a special flower. `before` methods are always run, whether any tests within the suite are run or not.

[The support tests](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/tests/functional/support.js#L38) make use of `before`, causing these methods to always be run.

For example, if I run `npm run test-functional -- --grep="totp"`, those methods are still called. Each of those suites only have one test, changing `before` to `beforeEach` would mitigate this.

chenba:
Fixes #1985


6a68:
Rebased to pick up Les's fix for the Circle-CI issues (#1996)

PR #1999, opened 2019-07-30
chore(ci): Use Firefox 68 on CircleCI

shane-tomlinson:
Let's see if Firefox 68 works on Circle. It'd be nice to test on the latest Firefox.


shane-tomlinson:
Closing, 1 failing test on 68.

shane-tomlinson:
The failure was caused by #2001, re-opening

PR #2005, opened 2019-07-30
refactor(tests): tidy up the email tests

philbooth:
I've been working with the auth server's email tests today and spotted a couple of things that could be cleaned up a little bit:

* As a hangover from the bad old days (pre-chai), we were using `assert.ok` with a custom `includes` predicate to check the existence of expected substrings. That works, but it means the error messages are really unhelpful when tests fail: `expected false to be truthy`

  By switching to chai's `assert.include` and `assert.notInclude` instead, now we get more useful errors that actually print the strings in question.

* The same `includes` predicate was also being used with arrays to categorise the different email types. Changing those to sets instead means we don't have to iterate through each array every time we check membership. It's not a huge saving in terms of time (a second or two on my machine), but it means we can remove that custom predicate entirely.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/test/local/senders/email.js
+++ b/packages/fxa-auth-server/test/local/senders/email.js
@@ -47,7 +47,7 @@ const messageTypes = [
   'postRemoveAccountRecoveryEmail',
 ];
 
-const typesContainSupportLinks = [
+const typesContainSupportLinks = new Set([
   'lowRecoveryCodesEmail',
   'newDeviceLoginEmail',
   'passwordChangedEmail',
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -123,13 +124,14 @@ const typesContainLocationData = [
   'postConsumeRecoveryCodeEmail',
   'postNewRecoveryCodesEmail',
   'passwordResetAccountRecoveryEmail',
-  'postRemoveTwoStepAuthenticationEmail',
```
philbooth: A few of the sets had dup items, this one...

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -84,31 +84,32 @@ const typesContainPasswordChangeLinks = [
   'postChangePrimaryEmail',
   'postRemoveTwoStepAuthenticationEmail',
   'postVerifySecondaryEmail',
-  'postVerifySecondaryEmail',
```
philbooth: ...and this one...

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -138,17 +140,15 @@ const typesContainManageSettingsLinks = [
   'postRemoveTwoStepAuthenticationEmail',
   'postNewRecoveryCodesEmail',
   'postConsumeRecoveryCodeEmail',
-  'postRemoveTwoStepAuthenticationEmail',
```
philbooth: ...and this one.

PR #2008, opened 2019-07-30
feat(payments): add breadcrumb navigation to contact and sub list

### LINKED ISSUE #1693: [fxa-payments-server, fxa-contents-server] Add Breadcrumbs
In the mockups (e.g. https://mozilla.invisionapp.com/share/5BS1I7KANWP#/screens/363599381), the subscriptions management page on the payments server and the support form on the content server have breadcrumbs for navigation.  No breadcrumbs exist at the moment AFAICT.

Related to https://github.com/mozilla/fxa/issues/926

meandavejustice:
- fixes #1693

Still need to find a way to detect which environment we are running in, so that in the breadcrumbs that are in the support form page (content  server) can link to the subscriptions list page (payments server) and vice versa.

contact-breadcrumbs
sub-list

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/support.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/support.mustache
@@ -3,6 +3,12 @@
     <h1 id="fxa-manage-account"><span class="fxa-account-title">{{#t}}Firefox Accounts{{/t}}</span></h1>
   </header>
 
+  <ol class="breadcrumbs">
+    <li><a href="/settings">Account Home</a></li>
+    <li><a href="/subscriptions">Subscriptions</a></li>
+    <li><a href="/support">Contact Us</a></li>
+  </ol>
+
   <div class="settings-success-wrapper">
     <div class="success settings-success"></div>
   </div>
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -62,6 +62,7 @@ body.settings #stage .settings {
   align-items: center;
   align-self: center;
   display: flex;
+  flex-direction: column;
   width: 100%;
 
   @include respond-to('big') {
@@ -779,5 +780,28 @@ section.modal-panel {
   html[dir='rtl'] & {
     background-position: center left;
     padding-left: 25px;
+  }
+}
+
+.breadcrumbs {
+  width: 100%;
+  margin-top: 0px;
+  margin-left: 100px;
+
+  li {
+    list-style: none;
+    float: left;
+    margin: 0px;
+    padding: 0px;
+
+    &:before {
+      content: "\003E";
+      padding: 0px 10px;
+    }
+
+    &:first-child:before {
+      content: "";
+      padding: 0px;
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


lmorchard:
Adding fxa-devs in case someone else wants to take a look too

packages/fxa-content-server/app/scripts/templates/support.mustache
```diff
@@ -3,6 +3,12 @@
     <h1 id="fxa-manage-account"><span class="fxa-account-title">{{#t}}Firefox Accounts{{/t}}</span></h1>
   </header>
 
+  <ol class="breadcrumbs">
+    <li><a href="/settings">Account Home</a></li>
+    <li><a href="//127.0.0.1:3031/subscriptions">Subscriptions</a></li>
```
lmorchard: This URL should just be `/subscriptions`

packages/fxa-payments-server/src/components/AppLayout/index.tsx
```diff
@@ -47,6 +49,13 @@ export const SettingsLayout = ({
     return () => document.body.classList.remove('settings');
   }, [ children ]);
 
+  let breadcrumbs = (
+      <Breadcrumbs>
+        <li><a href="//127.0.0.1:3030/settings">Account Home</a></li>
```
lmorchard: This is where you'll want to use the content server URL from config via AppContext

packages/fxa-payments-server/src/components/Breadcrumbs/index.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Breadcrumbs } from './Breadcrumbs';
```
lmorchard: I think here, you just want `./index` - this is what's failing in CI

packages/fxa-payments-server/src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children: any,
+  className?: string,
+};
+
+export const Breadcrumbs = ({
+  children,
+  className = 'breadcrumbs',
+}: BreadcrumbsProps) => {
+  return (
+      <ol className={className}>{children}</ol>
```
lmorchard: If this is all that the breadcrumbs component does, you could just leave it out and use `<ol>` directly in AppLayout. Alternatively, you could move the links and the code to access `AppContext` into here, too

packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -781,3 +782,26 @@ section.modal-panel {
...
+      content: "\003E";
+      padding: 0px 10px;
+    }
+
+    &:first-child:before {
+      content: "";
+      padding: 0px;
+    }
+  }
+}
```
chenba: The Github syntax highlighting is suggesting that you don't need `px` after the `0`.  😄

chenba: Or not.  I just don't understand colors apparently.

PR #2010, opened 2019-07-30
fix(security): update HSTS to 31536000

chenba:
Update HSTS to the value asked on the Security Checklist.

For consistency, I'm updating all the HSTS max-age values in the monorepo.

Part of #1128

@mozilla/fxa-devs r>

### DIFF
--- a/packages/browserid-verifier/lib/server.js
+++ b/packages/browserid-verifier/lib/server.js
@@ -53,7 +53,7 @@ app.use(function(req, res, next) {
   res.setHeader('X-XSS-Protection', '1; mode=block');
   res.setHeader('X-Content-Type-Options', 'nosniff');
   res.setHeader('X-Frame-Options', 'DENY');
-  res.setHeader('Strict-Transport-Security', 'max-age=15552000');
+  res.setHeader('Strict-Transport-Security', 'max-age=31536000');
   res.setHeader(
     'Content-Security-Policy',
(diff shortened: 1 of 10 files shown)


chenba:
We can force it to expire by setting `0`.  But like you said, I don't see why we'd want to do that.

shane-tomlinson:
I'm not sure why we weren't at 1 year, and it seems like we should add subscriptions.firefox.com to the HSTS preload list. We aren't sending the [`preload` parameter](https://infosec.mozilla.org/guidelines/web_security#http-strict-transport-security) in the header either, strange.

@jrgm - does anything need to be added to the nginx config to update hsts values?

shane-tomlinson:
>  We aren't sending the preload parameter in the header either, strange.

accounts.firefox.com and its subdomains are pre-loaded, [@jrgm and @fmarier did the grunt work long ago](https://bugzilla.mozilla.org/show_bug.cgi?id=958313).

shane-tomlinson:
Ref https://src.chromium.org/viewvc/chrome/trunk/src/net/http/transport_security_state_static.json?pathrev=290628#l818

shane-tomlinson:
https://github.com/mozilla/fxa-content-server/issues/208 on why we set to 180 days, [I note that value was tentative](https://github.com/mozilla/fxa-content-server/issues/208#issuecomment-32282671).

Issue #741, opened 2019-04-04
[Support Integration] Security Checklist


jvehent:
Risk Management
---------------
* [x] The service must have performed a Rapid Risk Assessment and have a Risk Record bug
* [x] [todo: julien] The service must be registered via a [New Service issue](https://github.com/mozilla-services/foxsec/issues/new?template=NewService.md&labels=New%20Service&assignee=psiinon&title=New%20Service:%20)

Infrastructure
--------------

* [x] Access and application logs must be archived for a minimum of 90 days
* [x] Use [Modern](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) or [Intermediate](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility) TLS
* [x] Set HSTS to 31536000 (1 year)
  * `strict-transport-security: max-age=31536000`
  * [x] [todo: julien] If the service is not hosted under `services.mozilla.com`, it must be manually added to [Firefox's preloaded pins](https://dxr.mozilla.org/mozilla-central/source/security/manager/tools/PreloadedHPKPins.json#184). This only applies to production services, not short-lived experiments.
* [x] Correctly set client IP
  * [x] Confirm client ip is in the proper location in [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For), modifying what is sent from the client if needed. AWS and GCP's load balancers will do this automatically.
  * [x] Make sure the web server and the application get the true client IP by configuring trusted IP's within Nginx or Apache
    * Nginx: [ngx_http_realip_module](https://nginx.org/en/docs/http/ngx_http_realip_module.html)
    * Apache: [mod_remoteip](https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html)
  * [x] If you have a service-oriented architecture, you must always be able to find the IP of the client that sent the initial request. We recommend passing along the `X-Forwarded-For` to all back-end services.
* If service has an admin panels, it must:
  * [x] only be available behind Mozilla VPN (which provides MFA)
  * [x] require Auth0 authentication

Development
-----------
* [x] Ensure your code repository is configured and located appropriately:
[...]

PR #2015, opened 2019-07-30
chore(security): add HSTS to Support Panel

chenba:
Part of #741


packages/fxa-support-panel/lib/server.ts
```diff
@@ -21,7 +21,17 @@ export function init(config: ServerConfig, logger: Logger): hapi.Server {
...
+    port: config.listen.port,
+    routes: {
+      security: {
+        hsts: {
+          maxAge: 31536000,
+          includeSubDomains: true,
+          preload: false,
+        },
+        xss: true,
+      },
```
chenba: Some of these options are here because they are required by the TypeScript interface of that config type.

PR #2016, opened 2019-07-30
Renovate updates

6a68:
This PR adds support-server to renovate's list of managed package.json files, and also updates the [`rangeStrategy`](https://renovatebot.com/docs/configuration-options/#rangestrategy) to `replace` old ranges with new ones when a package gets a major version bump.

I'm curious what factors led to the current FxA approach of using lockfiles and specifying ranges. Renovate provides a [lengthy article](https://renovatebot.com/docs/dependency-pinning/) on pros and cons of dependency pinning approaches, but I'm not sure what would be best for FxA.

Feedback welcome!

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,13 +1,15 @@
 {
+  "automerge": false,
+  "branchPrefix": "renovate/",
   "extends": [
     "config:base"
   ],
-  "prConcurrentLimit": 2,
-  "schedule": "every weekend",
-  "automerge": false,
-  "branchPrefix": "renovate/",
   "includePaths": [
     "packages/fxa-payments-server/package.json",
-    "packages/fxa-payments-server/Dockerfile"
-  ]
+    "packages/fxa-payments-server/Dockerfile",
+    "packages/fxa-support-panel/package.json"
+  ],
+  "rangeStrategy": "replace",
+  "prConcurrentLimit": 2,
+  "schedule": "every weekend"
 }


renovate.json
```diff
@@ -1,13 +1,15 @@
...
-  "prConcurrentLimit": 2,
-  "schedule": "every weekend",
-  "automerge": false,
-  "branchPrefix": "renovate/",
   "includePaths": [
     "packages/fxa-payments-server/package.json",
-    "packages/fxa-payments-server/Dockerfile"
-  ]
+    "packages/fxa-payments-server/Dockerfile",
+    "packages/fxa-support-server/package.json"
```
chenba: Unless I'm very behind and/or out of the loop, there's no `fxa-support-server`.  Did you mean `fxa-support-panel`?

philbooth:
Can we keep the commit messages in our uniform fix/chore/whatever format? I noticed there's a few stragglers creeping in to the tree of late, it just makes `git log` a bit easier to work with if everything is formatted the same and your eye can skip straight to the interesting stuff.

renovate.json
6a68: doh, yes, search/replace fail

6a68:
I'm wondering if an ADR would be appropriate for documenting whether we pin dependencies / devDependencies, or use ranges, and how we want to set up a process to update pinned / range dependencies. If you think this would be useful, comment or add a :+1: to this comment

shane-tomlinson:
> I'm wondering if an ADR would be appropriate for documenting whether we pin dependencies / devDependencies, or use ranges, and how we want to set up a process to update pinned / range dependencies. 

We decided in a retro that carets were fine.

6a68:
gah, hit the green button and realized the formatting is off in the merge commit.

PR #2035, opened 2019-08-01
Working on Security Dashboard

farhan787:
@vbudhram our final phase, working on `content-server` to make the security dashboard.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/security_events.js
+++ b/packages/fxa-content-server/app/scripts/views/security_events.js
@@ -1,0 +1,65 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import 'modal';
+import BaseView from './base';
+import SecurityEvent from '../../scripts/models/security-events';
+import Template from 'templates/security_events.mustache';
+
+let account;
+
+const View = BaseView.extend({
+  template: Template,
+  className: 'security-events',
+  viewName: 'security_events',
+
+  mustVerify: true,
+
+  events: {
+    'click #delete-events': '_deleteSecurityEvents',
+  },
+
+  beforeRender() {
+    account = this.getSignedInAccount();
+    if (!account) {
+      this.navigate('/signin');
+    }
+
+    return this._fetchSecurityEvents();
+  },
+
+  setInitialContext(context) {
+    context.set({
+      securityEvents: this._securityEvents,
+    });
+  },
+
+  _fetchSecurityEvents() {
+    return account.securityEvents().then(events => {
+      this._securityEvents = events.map(event => {
+        event.createdAt = formatDate(new Date(event.createdAt));
+
+        return new SecurityEvent(event).toJSON();
+      });
+    });
+  },
+
+  _deleteSecurityEvents() {
+    return account.deleteSecurityEvents().then(() => {
+      this._securityEvents = [];
+      return this.render();
+    });
+  },
+});
+
+function formatDate(dateObj) {
+  const date = dateObj.toDateString();
+  let time = dateObj.toTimeString();
+  const indexOfFirstSpace = time.indexOf(' ');
+  time = time.substring(0, indexOfFirstSpace);
+
+  return `${date} ${time}`;
+}
+
+export default View;
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -90,6 +90,7 @@ const View = BaseView.extend({
     context.set({
       ccExpired: !!this._ccExpired,
       escapedCcExpiredLinkAttrs: `href="/subscriptions" class="alert-link"`,
+      securityEventsVisible: this.displaySecurityEvents(),
       showSignOut: !account.isFromSync(),
       unsafeHeaderHTML: this._getHeaderHTML(account),
     });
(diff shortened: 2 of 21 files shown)


farhan787:
@vbudhram I haven't added test for `views/security-events.js` because that's not complete yet and we definitely will have to modify that in upcoming work.

farhan787:
@vbudhram content-server test is failing because of [this issue](https://github.com/mozilla/fxa/issues/2061), I've [fixed it](https://github.com/mozilla/fxa/pull/2062).

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -3022,4 +3022,56 @@ describe('models/account', function() {
...
+        return Promise.resolve(events);
+      });
+
+      return account.securityEvents().then(res => {
+        assert.isTrue(fxaClient.securityEvents.calledOnce);
+        assert.isTrue(fxaClient.securityEvents.calledWith(SESSION_TOKEN));
+
+        assert.equal(res.length, 2);
+
+        assert.equal(res[0].name, 'account.login');
```
vbudhram: Any reason why you couldn't do compare the entire res and events objects?

packages/fxa-content-server/tests/functional/security_events.js
```diff
@@ -0,0 +1,42 @@
...
+  openPage,
+  testElementExists,
+} = FunctionalHelpers;
+
+const config = intern._config;
+const SECURITY_EVENTS_URL = `${config.fxaContentRoot}security_events?security_events=true`;
+const PASSWORD = '12345678';
+let email;
+
+registerSuite('security_events', {
```
vbudhram: At a high level, we should be testing the following

* No link when no `security_events` query param
* Link when `security_events` query param
* Testing that user can log in and event is show
* Testing that user can reset password and event is show
* Testing that user can delete events (it will fail since the js-client package needs to be republished)

packages/fxa-content-server/app/scripts/views/security_events.js
```diff
@@ -0,0 +1,73 @@
...
+    account = this.getSignedInAccount();
+    if (!account) {
+      this.navigate('/signin');
+    }
+
+    return this._fetchSecurityEvents();
+  },
+
+  initialize(options = {}) {
+    if (options.securityEvents) {
```
vbudhram: I'm not really sure if this function is needed? The before render will fetch the events.

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -292,6 +297,18 @@ const View = BaseView.extend({
     }, this.SUCCESS_MESSAGE_DELAY_MS);
     return BaseView.prototype.unsafeDisplaySuccess.apply(this, arguments);
   },
+
+  showSecurityEvents() {
+    if (this.window.location.href.includes('security_events=true')) {
```
vbudhram: While this does work, it is better to set the broker capabilities similar to https://github.com/mozilla/fxa/blob/33342a9dca24818fbed0b8ac1cf296e8d80998c4/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js#L52.

farhan787:
@vbudhram PR is updated.

packages/fxa-content-server/app/tests/spec/views/settings.js
```diff
@@ -258,6 +262,7 @@ describe('views/settings', function() {
   describe('with uid', function() {
     beforeEach(function() {
       sinon.stub(account, 'settingsData').callsFake(() => Promise.resolve({}));
+      broker.setCapability('securityEvents', false);
```
vbudhram: Is this still needed? The capability should be false by default.

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -255,6 +259,20 @@ const View = BaseView.extend({
...
   },
 
+  /**
+   * Should the security events be displayed ?
+   *
+   * @returns {Boolean}
+   * @private
+   */
+  _areSecurityEventsVisible() {
+    if (this.window.location.href.includes('security_events=true')) {
```
vbudhram: Lets move this check into the broker fetch function, https://github.com/mozilla/fxa/blob/52c86a8d17d17e6362bac4b5a3e0f1b85954015b/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js#L150,

From there you can do something like
```
this.setCapability('showSecurityEvents', !! this.getSearchParam('security_events'));
```

This will remove the need to do `this.window.location.href.includes`. From the settings page, you will be able to directly check the broker capabilities and decided to show the link.

farhan787:
@vbudhram your review.

packages/fxa-content-server/tests/functional/security_events.js
```diff
@@ -0,0 +1,143 @@
...
+const SECURITY_EVENTS_URL = `${config.fxaContentRoot}security_events`;
+const PASSWORD = '12345678';
+let email;
+
+registerSuite('security_events', {
+  beforeEach: function() {
+    email = TestHelpers.createEmail();
+
+    return this.remote
+      .then(createUser(email, PASSWORD))
```
vbudhram: This should set the `preverified` flag to true to create a verified account, ref https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/tests/functional/sign_in_blocked.js#L39.

packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -90,6 +91,7 @@ const View = BaseView.extend({
     context.set({
       ccExpired: !!this._ccExpired,
       escapedCcExpiredLinkAttrs: `href="/subscriptions" class="alert-link"`,
+      securityEventsVisible: this._securityEventsVisible,
```
vbudhram: You can call `securityEventsVisible: this.displaySecurityEvents();` and remove the `this._securityEventsVisible` propery.

PR #2017, opened 2019-07-31
fix(emails): sentence-case and simplify the email subjects

philbooth:
Fixes #1684.

~~This is waiting on a couple of outstanding UX decisions so it's not ready for review yet.~~

But fwiw, it updates the email subjects to use sentence casing and simplifies a number of them. 

More discussion in the linked issue and [the spreadsheet](https://docs.google.com/document/d/1LRf1SiFg3zxJbDVFbilCUVDKjKcRt4_cnT_epzKm684/edit).

PR #2038, opened 2019-08-01
fix(email): remove codes from email subjects

philbooth:
Unfortunately this only came up #2017 after landed, but apparently secops advise that we shouldn't include any kind of authorisation code in the subject line of emails. So this removes the two cases that got added earlier and, at @ryanfeeley's suggestion, includes the client name instead.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -805,12 +805,15 @@ module.exports = function(log, config, oauthdb) {
       'X-Report-SignIn-Link': links.reportSignInLink,
     };
 
+    const clientName = safeUserAgent.name(message.uaBrowser);
+
     return this.send(
       Object.assign({}, message, {
         headers,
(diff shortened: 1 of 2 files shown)


philbooth:
(also, stating the obvious, but string change alert, we need to land this before they're cut tomorrow)

philbooth:
> are there any tests that need to be updated for this?

There should be but there aren't. Our email tests are pretty hit and miss, the newer emails tend to be more comprehensively covered than the older ones. I'll see if I can come up with a relatively painless way to make sure there are assertions for all the email subjects before merging, busy making dinner at the moment though.

lmorchard:
> I'll see if I can come up with a relatively painless way to make sure there are assertions for all the email subjects before merging, busy making dinner at the moment though.

Yeah, I'd say if it looks at all to be a big job, we can just punt to a follow up issue and merge as-is

PR #2040, opened 2019-08-01
fix(js-client): remove minified version of the client

vladikoff:
The client is minified by the content server or some other consumer.
The current version is broken anyway due to incompat of ES6 and Uglify

Fixes https://github.com/mozilla/fxa/commit/83d0cde8d2970ff82a29db8d3203fbfa383d0f25#r34537908


vladikoff:
> Are you planning on bumping the version?

Nope, I don't need a new version right now

Issue #2022, opened 2019-07-31
[FFTV] The "Show password" button cannot be accessed 


athomasmoz:
### Summary

"Show password" option does not work on Firefox for Fire TV. Looking to FxA team to prioritize based on usage of the feature

### Steps to reproduce

1. From FFTV home screen
2. Select button in the top right corner labled "CC"
    -> Browser will navigate to sign in
3. Hit enter to begin text input
    -> Soft keyboard will slide up
4. Enter email
5. Select submit
6. Enter a password.
7. Try to select and tap on the "Show password" button.

### Expected behavior

The password is revealed.

### Actual behavior

Button cannot be used with DPAD remote

### FFTV build information

* Build from master at https://github.com/mozilla-mobile/firefox-tv
  * Use debug flavor (should be default)
  * Fxa is currently behind a feature flag. To enable:
    * Install application
    * Stop application (adb shell am force-stop org.mozilla.tv.firefox.debug)
    * Start application with Fxa flag enabled (adb shell am start -n org.mozilla.tv.firefox.debug/org.mozilla.tv.firefox.MainActivity -a android.intent.action.MAIN --esa qaActiveExperiments SendTab-2511:B)
    * We have documentation on flagged code [here](https://github.com/mozilla-mobile/firefox-tv/wiki/QA)
* Should be run on a Fire TV device
  * The app mostly works on emulators, but some of these edge cases behave differently between Android and Amazon WebViews
* Please reach out to the FFTV team at #firefox-for-fire-tv on Slack for any help building the project

dannycoates:
I suggest wontfix because the on-screen keyboard has it's own 'show password' and special casing the ui for tv is undesirable

dannycoates:
even though i said wontfix, i went and fixed it anyway in #2041

Issue #1470, opened 2019-05-24
Password reset requirements blocks form on android [bz1550744]


bug-mirror:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1550744

User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0
Firefox for Android

Steps to reproduce:

I experienced this bug in android firefox and android chrome.  I clicked the link in my firefox sync password reset email.  This took me to a page that has a form with two text fields for a new password and a confirmation of that password. I put my password in the first field.  As I typed it, a box appeared below the field.  The box informed me of password length and complexity requirements.  The box also covers the second text entry field and the submit button.  There is no way to dismiss it.  It does not disappear no matter where I press on the screen. I cannot enter the confirming password in the second field, nor hit submit.

Actual results:

Nothing happened.  The box sat there, I couldn't move forward, and so I gave up on sync.  My primary need for sync was between mobile and my laptop, and I was giving it a tryout after years of ignoring it.  If android isn't a supported platform, I'll figure something else out, so it's not a big deal.

Expected results:

I don't know what the preferred UI path should be.  Whatever hints you care to give me about password requirements (and I do appreciate getting the hints), they should not interrupt the workflow to the degree they do.

Perhaps, when I hit the "enter" key on my android keyboard, I could be taken to the next field.  That's not sufficient (because some users won't know to do that) but it would help.

At a minimum, when I press anywhere outside the text field, the hints box should disappear.

PR #2041, opened 2019-08-01
fix(tv): fixed some ui glitches on Fire TV

### LINKED ISSUE #2013: Can't sign up on FFTV due to focus issue
### Steps to Reproduce

1. Open FFTV pointed at https://latest.dev.lcip.org/ (here's an APK https://drive.google.com/file/d/1IIdMoVjbNtWKmE1H76XKjRRAMfZu5EsL/view?usp=sharing )
2. Click "Receive Tabs" in toolbar or go to https://www.mozilla.org/en-CA/firefox/accounts/
3. Click sign up
4. Enter Email
5. Click Continue

### Expected Behaviour

You are taken to "Create a Firefox Account" page and "Password" field is focused

### Actual Behaviour

You are taken to "Create a Firefox Account" page, but no field is focused. 

Using the remote arrows, you can reach “terms of service”, “create account” or “mistyped email?” You can't reach the "password" field because the focus gets stuck on "how old are you"

7-30-fxa-flow

dannycoates:
This temporarily disables the password strength on small screens until a better design can be implemented.

@ryanfeeley I'm happy to re-enable and change the strength balloon once we have a solution. This is just a quick fix to unblock Fire TV.

fixes #2013

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/coppa-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/coppa-mixin.js
@@ -121,6 +121,8 @@ export default function(config = {}) {
           keyCode === KeyCodes.TAB ||
           keyCode === KeyCodes.LEFT_ARROW ||
           keyCode === KeyCodes.RIGHT_ARROW ||
+          keyCode === KeyCodes.UP_ARROW ||
+          keyCode === KeyCodes.DOWN_ARROW ||
           keyCode === KeyCodes.ENTER ||
           (keyCode >= KeyCodes.NUM_0 && keyCode <= KeyCodes.NUM_9) ||
           (keyCode >= KeyCodes.NUMPAD_0 && keyCode <= KeyCodes.NUMPAD_9)
--- a/packages/fxa-content-server/app/scripts/views/mixins/password-strength-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/password-strength-mixin.js
@@ -32,11 +32,6 @@ export default function(config = {}) {
           // problems in the reset-password screen when using the recovery key
           return;
         }
-        if (window.matchMedia('(max-width: 960px)').matches) {
-          // Temporarily disable balloon on small screens.
-          // Part of https://github.com/mozilla/fxa/issues/2013
-          return;
-        }
 
         const passwordModel = (this.passwordModel = this._createPasswordStrengthBalloonModel());
         // wait a short time after the last invalid event to log the invalid reason.
--- a/packages/fxa-content-server/app/scripts/views/mixins/password-strength-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/password-strength-mixin.js
@@ -32,6 +32,11 @@ export default function(config = {}) {
           // problems in the reset-password screen when using the recovery key
           return;
         }
+        if (window.matchMedia('(max-width: 960px)').matches) {
+          // Temporarily disable balloon on small screens.
+          // Part of https://github.com/mozilla/fxa/issues/2013
+          return;
+        }
 
         const passwordModel = (this.passwordModel = this._createPasswordStrengthBalloonModel());
         // wait a short time after the last invalid event to log the invalid reason.
--- a/packages/fxa-content-server/app/scripts/views/password_strength/password_with_strength_balloon.js
+++ b/packages/fxa-content-server/app/scripts/views/password_strength/password_with_strength_balloon.js
@@ -70,9 +70,10 @@ const PasswordWithStrengthBalloonView = FormView.extend({
       // which causes the check to occur a 2nd time.
       return !!this.model.validate({ password });
     }
-
-    // There is no prefilled password, yes, show the balloon.
-    return true;
+    // when css `respond-to('balloonSmall') wait until the password
+    // isn't blank to show in order to allow the user to see
+    // the form unobstructed
+    return window.matchMedia('(min-width: 960px)').matches;
   },
 
   createBalloon() {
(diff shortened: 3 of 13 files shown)


packages/fxa-content-server/app/scripts/views/mixins/coppa-mixin.js
```diff
@@ -121,6 +121,8 @@ export default function(config = {}) {
           keyCode === KeyCodes.TAB ||
           keyCode === KeyCodes.LEFT_ARROW ||
           keyCode === KeyCodes.RIGHT_ARROW ||
+          keyCode === KeyCodes.UP_ARROW ||
+          keyCode === KeyCodes.DOWN_ARROW ||
```
dannycoates: missing these caused the coppa to be a black hole for the remote control

packages/fxa-content-server/app/styles/_breakpoints.scss
```diff
@@ -5,9 +5,9 @@ $breakpoints: (
     '(max-width: 520px), (max-height: 640px), (orientation: landscape) and (min-width: 641px) and (max-height: 640px)',
   big: '(min-width: 521px) and (min-height: 641px)',
   balloonSmall: '(max-width: 960px)',
-  balloonBig: '(min-width: 960px)',
+  balloonBig: '(min-width: 961px)',
```
dannycoates: not explicitly necessary while the balloon is disabled, but when the screen was exactly 960px (as the TV is) both small and big selectors were triggered

packages/fxa-content-server/app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -32,6 +32,11 @@ export default function(config = {}) {
           // problems in the reset-password screen when using the recovery key
           return;
         }
+        if (window.matchMedia('(max-width: 960px)').matches) {
```
shane-tomlinson: Eh? I think it will make @ryanfeeley sad to disable the balloon universally on mobile devices. The balloon is supposed to go away once the password is valid, is that broken? See also https://github.com/mozilla/fxa/issues/1470.

dannycoates: It's broken on TV because it prevents you from completing the form at all. I'm happy to fix it but my priority is to unblock them and follow up with a working balloon

shane-tomlinson: Can we UA sniff? Gross, but doesn't kill the feature for everyone.

dannycoates: Seems broken for everyone to me with #1470

ryanfeeley: This UI needs some TLC. The repeat password message should disappear when the basic requirements are met (for the second field that means the password is repeated). Original design: https://mozilla.invisionapp.com/share/5UM3J3TXWMR#/316338169_Artboard

shane-tomlinson: > Seems broken for everyone to me with #1470

In the bugzilla bug I asked the user if the password bubble goes away if the requirements are met. Yes, yes they do.

> I just tried it again, and it does indeed go away when the requirements are met. I still maintain the interface is surprising. Covering form elements with a little box like that so you cannot click on them is not a common way to communicate that elements are not available or enabled. Perhaps showing those elements but graying them out would be a more effective way to communicate the message. However, if you don't consider this a bug, I suppose it's not a bug. Thanks for taking a look.

dannycoates: There isn't anything very unique about the ua string on TV.
`Mozilla/5.0 (Linux; Android 7.1.2) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Focus/4.2-LAT1 Chrome/59.0.3071.125 Mobile Safari/537.36` (`Focus` is the closest thing worth filtering on I suppose)

I don't see anything wrong with disabling it on mobile until it's fixed. It's obviously causing enough confusion for someone to file a bug, even if it technically works.

dannycoates: If we merge this after the train I may even have the bubble fixed before the next train. Really, I just want this on `latest`

shane-tomlinson: > If we merge this after the train I may even have the bubble fixed before the next train. Really, I just want this on latest

You've convinced me! r+

dannycoates:
@shane-tomlinson instead of doing a new PR I just added a new commit here that fixes the password balloon.

fixes #2051 
fixes #2022

packages/fxa-content-server/app/scripts/views/mixins/coppa-mixin.js
shane-tomlinson: The best part about this change is that up/down arrow on Fx Desktop now cause the value to increment/decrement as you'd expect in a number input! 💯

packages/fxa-content-server/app/scripts/views/password_strength/password_with_strength_balloon.js
```diff
@@ -70,9 +70,10 @@ const PasswordWithStrengthBalloonView = FormView.extend({
       // which causes the check to occur a 2nd time.
       return !!this.model.validate({ password });
     }
-
-    // There is no prefilled password, yes, show the balloon.
-    return true;
+    // when css `respond-to('balloonSmall') wait until the password
```
shane-tomlinson: 👍 This is a great approach.

PR #2043, opened 2019-08-01
Add oauth_webchannel_v1 for Fenix and GeckoView apps

### LINKED ISSUE #1980: [fxa-content-server] Support OAuth WebChannels for Fenix
As part of adding Webchannels to Fenix, we need to add support for an OAuth Fenix flow that has CWTS and supports account events

Ref: https://github.com/mozilla/fxa/pull/1913

vladikoff:
Fixes #1980
Ref: mozilla-mobile/android-components#3881


packages/fxa-content-server/app/scripts/lib/app-start.js
```diff
@@ -637,6 +639,15 @@ Start.prototype = {
       Constants.DEVICE_PAIRING_AUTHORITY_REDIRECT_URI
     );
   },
+  /**
+   * Is the user initiating a device pairing flow as
+   * the auth device?
```
rfk: nit: comment copy-pasta from `isDevicePairingAsAuthority` above.

packages/fxa-content-server/app/scripts/lib/channels/receivers/web-channel.js
```diff
@@ -31,8 +31,12 @@ _.extend(WebChannelReceiver.prototype, Backbone.Events, {
...
 
   receiveMessage(event) {
-    const detail = event.detail;
-
+    let detail;
+    try {
+      detail = JSON.parse(event.detail);
+    } catch (e) {
+      detail = event.detail;
+    }
```
rfk: I'm curious why this change was necessary, is it a limitation of the webextension webchannel receiver?

I'd also be tempted to do a `typeof event.details` here to ensure you don't try to `JSON.parse` something that's not a string.

vbudhram:
@vladikoff is there a train you want to land this in and estimate of work?

packages/fxa-content-server/docs/relier-communication-protocols/fx-webchannel.md
```diff
@@ -83,6 +83,25 @@ Sent when a user successfully authenticates with Firefox Accounts and sync can b
...
+
+Sent when a user successfully authenticates via OAuth.
+
+##### data
+
+```
+{
+  "code": "02f3cfea84ac4c143662b38d6c7f0c82c6f91eb041befc7cecda446b1b4887c1",
+  "state": "vHao1p6OizzwReCkQMSpZA",
+  "redirect": "http://127.0.0.1:3030/oauth/success/a2270f727f45f648?code=02f3cfea84ac4c143662b38d6c7f0c82c6f91eb041befc7cecda446b1b4887c1&state=vHao1p6OizzwReCkQMSpZA",
```
shane-tomlinson: Is this right, or would it be an urn like `urn:ietf:wg:oauth:2.0:oob:oauth-redirect-webchannel`?

packages/fxa-content-server/app/scripts/lib/channels/receivers/web-channel.js
vladikoff: WebExtensions in Geckoview/ Fenix cannot pass js objects, only strings. Otherwise it displays `[protected]`, `[inaccessible]` or something

packages/fxa-content-server/docs/relier-communication-protocols/fx-webchannel.md
vladikoff: ah yes, we can do that!

packages/fxa-content-server/app/scripts/lib/channels/receivers/web-channel.js
rfk: Hrm, I also recall an ancient bug about sending only strings on Desktop as well, which has had to get added to a special allowlist in order to bypass.

shane-tomlinson: It's when *sending* an event to the browser that detail has to be stringified: 

* https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/channels/senders/web-channel.js#L117
* https://bugzilla.mozilla.org/show_bug.cgi?id=1275616
* https://bugzilla.mozilla.org/show_bug.cgi?id=1238128

PR #2049, opened 2019-08-02
chore(base): added a function to render the react component

hritvi:
@shane-tomlinson can you please review this!

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/base.js
+++ b/packages/fxa-content-server/app/scripts/views/base.js
@@ -275,14 +275,23 @@ var BaseView = Backbone.View.extend({
         // view is rendered or else stale data may
         // be returned.
         this._context = null;
-        this.$el.html(this.renderTemplate(this.template.bind(this)));
-        this.trigger('rendered');
+        return Promise.resolve()
+          .then(() => {
+            if (this.renderReactComponent) {
+              return this.renderReactComponent();
+            } else {
+              this.$el.html(this.renderTemplate(this.template.bind(this)));
+            }
+          })
+          .then(() => {
+            this.trigger('rendered');
 
-        // Track whether status messages were made visible via the template.
-        this._isErrorVisible = this.$('.error').hasClass('visible');
-        this._isSuccessVisible = this.$('.success').hasClass('visible');
+            // Track whether status messages were made visible via the template.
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/views/base.js
```diff
@@ -381,9 +390,6 @@ var BaseView = Backbone.View.extend({
         err => {
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
             this.logError(AuthErrors.toError('SESSION_EXPIRED'));
-            // The redirectTo in .navigate() is lost if there's later navigations, so by saving it here
-            // we can get it later (e.g., in case of a signUp):
-            this.relier.set('redirectTo', this.window.location.href);
```
shane-tomlinson: Can you put this back?

packages/fxa-content-server/app/tests/spec/views/base.js
```diff
@@ -180,6 +180,11 @@ describe('views/base', function() {
   });
 
   describe('render', function() {
+    it('renders react component correctly', () => {
+      if (view.renderReactComponent) {
```
shane-tomlinson: What needs to happen here is to create a method on the view called `renderReactComponent`, call `render()`, and then ensure the method is called. Right now no `renderReactComponent` method is on the view, so the test on line 185 isn't happening.

Have a look around at the uses of `sinon.spy`, that should give you the functionality that you need.

hritvi:
@shane-tomlinson I have updated the PR. Can you please see this now?

packages/fxa-content-server/app/tests/spec/views/base.js
```diff
@@ -180,6 +180,13 @@ describe('views/base', function() {
   });
 
   describe('render', function() {
+    it('renders react component correctly', () => {
+      var renderReactComponent = sinon.spy();
+      view.render(renderReactComponent);
+      view.render().then(function() {
```
shane-tomlinson: can you add a `return` before this so that the promise is returned too, otherwise we don't know if the assertion fails.

hritvi:
@shane-tomlinson updated the PR!

shane-tomlinson:
@hritvi - can you sign and re-push the commit so that we can merge it?

PR #2050, opened 2019-08-02
fix(request): adding payload to deleteSecurityEvents and fix js-client tests

farhan787:
cc @vbudhram


vladikoff:
@farhan787 you need to rebase this PR

farhan787:
>  the order should be reversed. The auth-server actually returns the newest events first.

Thanks @vbudhram for correcting me !

PR #2053, opened 2019-08-02
chore(fxa-payments-server): fixes #1923 - add CSP to the payments server

### LINKED ISSUE #1923: Add CSP to the payments-server
Seems a little involved, spinning off as a separate sub-issue from #1128.

We might be able to reuse the CSP-related middleware from the fxa-content-server package.

LZoog:
fixes #1923 

@shane-tomlinson and I met and agreed the CSP violation reporting should go to the content server rather than duplicate code or only moving this piece related to CSP into `fxa-shared`. The merge of this depends on getting the fix for #2176.

NOTE: TLDR below; went with option 3 and created https://github.com/mozilla/fxa/issues/2073

---
---

I believe I've got this mostly figured out, I would just like some input.

Files mentioned in this comment:
payment server config: `packages/fxa-payments-server/server/config/index.js`
content server config:`packages/fxa-content-server/server/lib/configuration.js`
shared file: `packages/fxa-shared/csp/blocking.js` (can I rename this to `middleware`?)

In the file shared between the content server and payment server mentioned above, we grab some data from the server config file. The payment server config is set up pretty differently than the content server config, and arguably much better - for example, the content server uses the `fxaccount_url` key for the same data that the payment server has set for `servers.auth.url`. The payment server schema seems clearer to me and better thought-out.

It seems to me the best thing to do is to get these configs to match up better, but that's going to require a lot of testing and time for the content server, and I don't really want to change the payment server config setup to one I feel is not as clear.

Even if this worked in the payment server config, it would be a temporary solution, but it doesn't work without adding `fxaccount_url` to the payments schema so it's 👎:
```
if (! conf.has('fxaccount_url')) {
  conf.set('fxaccount_url', conf.get('servers.auth.url'));
}
```

When retrieving the data, I can't use an `||` in the const declaration because it throws an error so I could do this, but it still is not ideal:
```
let authServer;
  try {
    authServer = getOrigin(config.get('fxaccount_url'))
  } catch(e) {
  authServer = getOrigin(config.get('servers.auth.url'));
}
```
However, I will need to do something like this anyway for a few vars due to needing to declare Stripe stuff in this file: https://stripe.com/docs/security#content-security-policy

It seems our options are:
1. Spend time getting the config files matching (there's also a `config` vs `configuration` file name difference)
2. Use try/catch like how I have it above, and create a new issue for the first option and to refactor this later
3. Create separate CSP middleware, despite a lot of reuse across them, and probably still create an issue for the first option and to refactor later

What's the preferred direction? I'd guess it'd be between 2/3.

Also, looks like we don't have a `prettier` file in `payments-server` - after committing, I have conflicts with eslint. I can open a separate PR to get that fixed and pull it in here.
Edit: relates to https://github.com/mozilla/fxa/issues/1673

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -15,11 +15,35 @@ const conf = convict({
     env: 'CLIENT_ADDRESS_DEPTH',
     format: Number,
   },
+  csp: {
+    enabled: {
+      default: true,
+      doc: 'Send "Content-Security-Policy" header',
+      env: 'CSP_ENABLED',
+    },
+    /*eslint-disable sorting/sort-object-props*/
+    reportUri: {
+      default: 'https://accounts.firefox.com/_/csp-violation',
+      doc: 'Location of "report-uri" for full, blocking CSP rules',
+      env: 'CSP_REPORT_URI',
+    },
+    reportOnlyEnabled: {
+      default: false,
+      doc: 'Send "Content-Security-Policy-Report-Only" header',
+      env: 'CSP_REPORT_ONLY_ENABLED',
+    },
+    reportOnlyUri: {
+      default: 'https://accounts.firefox.com/_/csp-violation-report-only',
+      doc: 'Location of "report-uri" for report-only CSP rules',
+      env: 'CSP_REPORT_ONLY_URI',
+    },
+    /*eslint-enable sorting/sort-object-props*/
+  },
   env: {
     default: 'production',
     doc: 'The current node.js environment',
     env: 'NODE_ENV',
-    format: ['development', 'production'],
+    format: ['development', 'production', 'test'],
   },
   hstsMaxAge: {
     default: 31536000, // a year
@@ -130,6 +154,14 @@ const conf = convict({
         format: 'url',
       },
     },
+    profileImages: {
+      url: {
+        default: 'http://127.0.0.1:1112',
+        doc: 'The url of the Firefox Account Profile Image Server',
+        env: 'FXA_PROFILE_IMAGES_URL',
+        format: 'url',
+      },
+    },
   },
   staticResources: {
     directory: {
@@ -157,6 +189,24 @@ const conf = convict({
       doc: 'API key for Stripe',
       env: 'STRIPE_API_KEY',
       format: String,
+    },
+    apiUrl: {
+      default: 'https://api.stripe.com',
+      doc: 'The Stripe API url',
+      env: 'STRIPE_API_URL',
+      format: 'url',
+    },
+    hooksUrl: {
+      default: 'https://hooks.stripe.com',
+      doc: 'The Stripe hooks url',
+      env: 'STRIPE_HOOKS_URL',
+      format: 'url',
+    },
+    scriptUrl: {
+      default: 'https://js.stripe.com',
+      doc: 'The Stripe script url',
+      env: 'STRIPE_SCRIPT_URL',
+      format: 'url',
     },
   },
 });
--- a/packages/fxa-payments-server/server/lib/csp/blocking.js
+++ b/packages/fxa-payments-server/server/lib/csp/blocking.js
@@ -3,9 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Middleware to take care of CSP. CSP headers are not sent unless config
-// option 'csp.enabled' is set (default true in development), with a special
-// exception for the /tests/index.html path, which are the frontend unit
-// tests.
+// option 'csp.enabled' is set (default true).
 
 const url = require('url');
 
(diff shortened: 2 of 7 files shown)


lmorchard:
You might want to give this a run through Prettier - looks like some files changed with respect to quotes that look like inadvertent noise in the diff.

packages/fxa-content-server/server/bin/fxa-content-server.js
```diff
@@ -59,8 +59,10 @@ const fourOhFour = require('../lib/404');
 const serverErrorHandler = require('../lib/500');
 const localizedRender = require('../lib/localized-render');
 const csp = require('../lib/csp');
-const cspRulesBlocking = require('../lib/csp/blocking')(config);
-const cspRulesReportOnly = require('../lib/csp/report-only')(config);
+const cspRulesBlocking = require('../../../fxa-shared/csp/blocking')(config);
```
lmorchard: I don't think you can easily move things to another package like `fxa-shared`, currently. [You'll probably run into being unable to use `../` to refer to files outside the current package directory when we run things in Docker](https://github.com/mozilla/fxa/issues/941#issuecomment-505511291). Might run locally, but will hit snags in testing and building for deployment.

LZoog: Sorry, I should have mentioned Phil's PR: https://github.com/mozilla/fxa/pull/1860

I should be able to do this once that is merged, right?

LZoog:
> You might want to give this a run through Prettier - looks like some files changed with respect to quotes that look like inadvertent noise in the diff.

Yeah, sorry my PR comment is so long, I mentioned that in there at the end😅

packages/fxa-content-server/server/bin/fxa-content-server.js
6a68: Sorry @LZoog, looks like I steered you in the direction of #1860 without understanding the limitations of that approach. @philbooth, which approach to accessing fxa-shared would be best for this PR?

LZoog: Huh well, I read through it and thought myself that it would be possible to do this with that landed! I commented out a section in `fxa-shared/index.js` that sets it up the current/old way - I can go that route if this won't work.

lmorchard: Looks like #1860 won't solve this problem generally. That PR partly solves it for fxa-content-server, by jumping up a directory and [explicitly copying fxa-shared into the container with fxa-content-server](https://github.com/mozilla/fxa/pull/1860/files#diff-bdd3d89f8c2822d921c755a374232952R30). You'd have to replicate those tweaks here for it to work for fxa-payment-server

chenba:
I don't understand this change: https://github.com/mozilla/fxa/pull/2053/files#diff-32f5486059cc5b0266f2a9c04ba6d4a2 😀

packages/fxa-content-server/server/bin/fxa-content-server.js
LZoog: Ah, I see. That makes sense.
Since we'd said on Slack that option #3 is the best way to go anyway, I'll remove the shared stuff and create the separate middleware. I won't need to reference the shared server at all.

LZoog:
> I don't understand this change: https://github.com/mozilla/fxa/pull/2053/files#diff-32f5486059cc5b0266f2a9c04ba6d4a2 😀

The commented out stuff? Yeah, I'll be removing it, I thought it could be a way to get the config files matching so I could share some files but I'm going with option #3.

I'll make those changes today and let you guys know when I've got this working/when I start on the tests. :)

chenba:
> 
> 
> > I don't understand this change: https://github.com/mozilla/fxa/pull/2053/files#diff-32f5486059cc5b0266f2a9c04ba6d4a2 😀
> 
> The commented out stuff? 

That anchor tag doesn't go to the right spot on click. :(  I meant the empty `packages/fxa-payments-server/server/lib/csp.js`, which was why I couldn't leave an in place comment.

LZoog:
> > > I don't understand this change: https://github.com/mozilla/fxa/pull/2053/files#diff-32f5486059cc5b0266f2a9c04ba6d4a2 😀
> > 
> > 
> > The commented out stuff?
> 
> That anchor tag doesn't go to the right spot on click. :( I meant the empty `packages/fxa-payments-server/server/lib/csp.js`, which was why I couldn't leave an in place comment.

Ah gotcha, I was wondering if I was referring to the right thing! :P I created that since we have `fxa-content-server/server/lib/csp.js`, I _think_ I'll need to duplicate it since I'm essentially copying what's in the content server but now that I have a direction I'll have this cleaned up a lot by EoD and give you a better answer after I push and you take another look if it's still confusing. :)

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -116,7 +143,7 @@ module.exports = () => {
           userReq /*, userRes*/
         ) {
           const contentType = proxyRes.headers['content-type'];
-          if (!contentType || !contentType.startsWith('text/html')) {
+          if (! contentType || ! contentType.startsWith('text/html')) {
```
LZoog: This was some linter complaint, although I believe we've gotten a fix in now so I could change this back🤷‍♀️

LZoog:
@shane-tomlinson will you look this over when you get a chance?

packages/fxa-content-server/server/lib/csp.js
```diff
@@ -20,6 +20,7 @@ module.exports = function(config) {
   const cspMiddleware = helmet.contentSecurityPolicy(config.rules);
 
   return htmlOnly((req, res, next) => {
+    cspMiddleware(req, res, next);
```
chenba: We don't need this because it's on line 25 below.

packages/fxa-payments-server/server/config/index.js
```diff
@@ -15,11 +15,41 @@ const conf = convict({
...
+      doc: 'Send "Content-Security-Policy-Report-Only" header',
+      env: 'CSP_REPORT_ONLY_ENABLED',
+    },
+    reportOnlyUri: {
+      default: 'http://accounts.firefox.com/_/csp-violation-report-only',
+      doc: 'Location of "report-uri" for report-only CSP rules',
+      env: 'CSP_REPORT_ONLY_URI',
+    },
+    /*eslint-enable sorting/sort-object-props*/
+  },
```
chenba: `https` for the URLs?

packages/fxa-payments-server/server/lib/csp/blocking.js
```diff
@@ -0,0 +1,106 @@
...
+        // to break the site for users who already use a Gravatar as
+        // their profile image.
+        GRAVATAR,
+        PROFILE_IMAGES_SERVER,
+      ]),
+      mediaSrc: [BLOB],
+      objectSrc: [NONE],
+      reportUri: config.get('csp.reportUri'),
+      scriptSrc: addCdnRuleIfRequired([SELF, STRIPE_SCRIPT_URL]),
+      styleSrc: addCdnRuleIfRequired([SELF, UNSAFE_INLINE]),
```
chenba: The security checklist (https://github.com/mozilla/fxa/issues/1128) says no unsafe-line for style-src.  What's the reason for this exception?

lmorchard: I think mainly because the components from Stripe use unsafe-inline styles and we can't get around that.

chenba: Oh right!  👍

packages/fxa-payments-server/server/lib/server.js
shane-tomlinson: Huh, yeah, I thought the style rules were consolidated.

packages/fxa-payments-server/server/lib/csp/blocking.js
```diff
@@ -0,0 +1,106 @@
...
+      imgSrc: addCdnRuleIfRequired([
+        SELF,
+        DATA,
+        // Gravatar support was removed in #4927, but we don't want
+        // to break the site for users who already use a Gravatar as
+        // their profile image.
+        GRAVATAR,
+        PROFILE_IMAGES_SERVER,
+      ]),
+      mediaSrc: [BLOB],
```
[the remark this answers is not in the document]
LZoog: Cool, good catch, this is something I was going to ask about and it slipped by.

packages/fxa-payments-server/server/lib/csp/blocking.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  const rules = {
+    directives: {
+      connectSrc: [
+        SELF,
+        AUTH_SERVER,
+        OAUTH_SERVER,
+        PROFILE_SERVER,
+        PAIRING_SERVER_WEBSOCKET,
```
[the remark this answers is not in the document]
LZoog: @shane-tomlinson I believe we need this for local dev. If I remove it, I get this:

image

LZoog:
@shane-tomlinson I actually took DATA out because I couldn't find where we were using it, but without it I do hit:

image

I'm adding it back and will push in just a sec, I believe I have addressed all other comments (thanks for the reviews! 😄). I really hope Circle CI plays nicely though, I had to rerun it 3 times...

packages/fxa-payments-server/server/lib/csp/blocking.js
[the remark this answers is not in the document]
shane-tomlinson: Oh, that's the hot-reloading websocket rather than the pairing server. Can you rename the variable here and ensure that it's only enabled for dev?

packages/fxa-payments-server/server/lib/csp/blocking.js
```diff
@@ -0,0 +1,106 @@
...
+        STRIPE_API_URL,
+      ],
+      defaultSrc: [SELF],
+      fontSrc: addCdnRuleIfRequired([SELF]),
+      frameSrc: [STRIPE_SCRIPT_URL, STRIPE_HOOKS_URL],
+      imgSrc: addCdnRuleIfRequired([
+        SELF,
+        DATA,
```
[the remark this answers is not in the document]
shane-tomlinson: Yup, the spinner. Resolving.

Issue #999, opened 2019-05-01
Login Error on Monitor


davismtl:
![image.png](https://images.zenhubusercontent.com/5ca2448cb4170956617cd475/74c3f129-11bd-464f-adcc-df8d02dde5bb)

Conditions:
- Nightly 68.0a1 (2019-04-29) (64-bit)
- Mac
- Signed in to sync in the browser

Steps to reproduce:
- Go to monitor.firefox.com
- Click Sign in top right of page
- On account.firefox.com (your email should already be displayed), click Sign In

Result:
- Unexpected Error (See screenshot)

Expected:
- Successful login

Console log:
![image.png](https://images.zenhubusercontent.com/5ca2448cb4170956617cd475/74001581-50ed-47a2-bb5a-2f86a8746195)

davismtl:
Filed from Slack conversation:
https://mozilla.slack.com/archives/C93K0MBM5/p1556570277037900

Betsy Mikel also experienced this.

I'm unable to reproduce on Beta.

davismtl:
I also confirm that I am properly signed into my browser and that I can access my preferences at accounts.firefox.com

rfk:
Michael Feldman also experienced this, so I imagine it's fairly widespread.  I wasn't able to reproduce with a fresh signin to sync, but it could be specific to particular settings in the browser.

rfk:
From network logs, this is a 401 on `/certificate/sign`, which explains the initial "invalid token".

rfk:
Response body: `{“code”:401,“errno”:110,“error”:“Unauthorized”,“message”:“Invalid authentication token: Missing authentication”,“info”:“https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format“}`

rfk:
The error string "Missing authentication" is very unusual here - it come from the hapi auth machinery, and AFAICT it means that hapi did not find an appropriate authentication strategy to use for the request. Which suggests maybe a corrupted authorization header rather than a problem with the sessionToken.

rfk:
I wasn't able to reproduce this yesterday, but tried it again today and I'm getting the same error (and same tracebacks). One odd thing I noticed is that my state in `localStorage` seems to have a `sessionTokenContext` but no `sessionToken`:

![Capture](https://user-images.githubusercontent.com/34695/57106258-3daca480-6d70-11e9-87a8-90e20b1be020.PNG)

[...]

PR #2056, opened 2019-08-02
feat(signin): Reduce the need to enter a password

### LINKED ISSUE #1371: RP flows asking me for my password even when I'm signed in to other services and the browser
Normally when I am already signed in to the browser and visit one of our services, it recognized me and I just need to click one button to sign in.

This doesn't seem to be happening anymore. I first thought it was a problem with the new flow on /whatsnew which promotes Monitor when you're signed in to the browser:
https://www.mozilla.org/en-US/firefox/67.0.1/whatsnew/

After further investigation, Send and AMO also asked me for my password. So I signed into them to see if it would help the Monitor flow. It did not seem to help improve the flow of those services. They all still ask for my password, one by one.

shane-tomlinson:
Do not ask for a password for users signing into a non-key
requesting RP if they already have an FxA session of any sort.

I removed a lot of tests that now seem duplicate, since there is
no longer a distinction between a "Sync" signin and an "OAuth" signin
when it comes to whether to display the password field.

I also added onto a bunch of the "cached session" tests
to actually click the "submit" button to ensure the user
is redirected where they are supposed to be. The idea
is to follow on with trying expired cached credentials
targeting the problems uncovered in #999.

fixes #1371

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/cached-credentials-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/cached-credentials-mixin.js
@@ -32,6 +32,11 @@ export default {
    * @returns {Boolean}
    */
   isPasswordNeededForAccount(account) {
+    // If the account doesn't have a sessionToken, we'll need a password
+    if (!account.get('sessionToken')) {
+      return true;
+    }
+
     // If the account doesn't yet have an email address, we'll need a password too.
     if (!account.get('email')) {
       return true;
@@ -40,13 +45,6 @@ export default {
     // If the relier wants keys, then the user must authenticate and the password must be requested.
     // This includes sync, which must skip the login chooser at all cost
     if (this.relier.wantsKeys()) {
-      return true;
-    }
-
-    // We need to ask the user again for their password unless the credentials came from Sync.
-    // Otherwise they aren't able to "fully" log out. Only Sync has a clear path to disconnect/log out
-    // your account that invalidates your sessionToken.
-    if (!this.user.isSyncAccount(account)) {
       return true;
     }
 
--- a/packages/fxa-content-server/app/scripts/views/sign_in.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_in.js
@@ -52,11 +52,8 @@ const View = FormView.extend({
     // should be re-rendered with the default avatar.
     const account = this.getAccount();
     this.listenTo(account, 'change:accessToken', () => {
-      // if no access token and password is not visible we need to show the password field.
-      if (
-        !account.has('accessToken') &&
-        this.$(PASSWORD_SELECTOR).is(':hidden')
-      ) {
+      // if no access token we need to show the password field.
+      if (!account.has('accessToken')) {
         this.model.set('chooserAskForPassword', true);
         return this.render().then(() => this.setDefaultPlaceholderAvatar());
       }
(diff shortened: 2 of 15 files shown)


packages/fxa-content-server/app/scripts/views/mixins/cached-credentials-mixin.js
```diff
@@ -33,7 +33,7 @@ export default {
    */
   isPasswordNeededForAccount(account) {
     // If the account doesn't yet have an email address, we'll need a password too.
```
6a68: Would be good to update this comment along with the conditional on line 36.

packages/fxa-content-server/app/scripts/views/sign_in.js
```diff
@@ -53,10 +53,7 @@ const View = FormView.extend({
     const account = this.getAccount();
     this.listenTo(account, 'change:accessToken', () => {
       // if no access token and password is not visible we need to show the password field.
```
6a68: I think this comment is also now out of date with the changed line below.

packages/fxa-content-server/app/tests/spec/views/sign_in.js
```diff
@@ -177,10 +177,11 @@ describe('views/sign_in', () => {
...
 
       it('re-renders, keeps the original email, forces user to enter password', () => {
+        console.log('w eare here');
         assert.equal(view.render.callCount, 2);
-        assert.equal(view.$('.prefillEmail').text(), 'a@a.com');
-        assert.equal(view.$('input[type=email]').val(), 'a@a.com');
-        assert.lengthOf(view.$('input[type=password]'), 1);
+        //assert.equal(view.$('.prefillEmail').text(), 'a@a.com');
+        //assert.equal(view.$('input[type=email]').val(), 'a@a.com');
+        //assert.lengthOf(view.$('input[type=password]'), 1);
```
6a68: I guess this is the WIP part ^_^

packages/fxa-content-server/app/tests/spec/views/mixins/cached-credentials-mixin.js
```diff
@@ -62,22 +62,20 @@ describe('views/mixins/cached-credentials-mixin', () => {
...
+      sinon.stub(relier, 'wantsKeys').callsFake(() => false);
       sinon.stub(user, 'isSyncAccount').callsFake(() => true);
       model.unset('chooserAskForPassword');
-      sinon
-        .stub(view, 'getPrefillEmail')
-        .callsFake(() => 'testuser@testuser.com');
+      sinon.stub(view, 'getPrefillEmail').callsFake(() => '');
 
       assert.isTrue(view.isPasswordNeededForAccount(account));
     });
```
6a68: I don't think there's a test to verify that `it('does not ask for a password if the account has email and sessionToken')`, so I'd suggest adding one.

packages/fxa-content-server/app/tests/spec/views/sign_in.js
shane-tomlinson: Yeah, I'm going to lean on functional tests to ensure this behavior is as expected.

packages/fxa-content-server/app/tests/spec/views/mixins/cached-credentials-mixin.js
shane-tomlinson: Nice catch and it pointed out that the tests weren't always testing what we expected!

shane-tomlinson: i.e., tests weren't failing when I expected them to fail.

shane-tomlinson:
@mozilla/fxa-devs, and @6a68 - I'm calling this ready. r?

vbudhram:
Heh well almost, looks like a couple test failures.

PR #2057, opened 2019-08-02
feature.signup codes

vbudhram:
Just for testing, building https://github.com/mozilla/fxa/pull/1296 as a feature branch.


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -31,6 +32,14 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push,
   const totpUtils = require('./utils/totp')(log, config, db);
   const skipConfirmationForEmailAddresses = config.signinConfirmation.skipForEmailAddresses;
 
+  // Default options for TOTP
+  otplib.authenticator.options = {
+    digits: 8,
+    encoding: 'hex',
+    step: 10 * 60, // 10 minutes in seconds
+    window: 1 // consider 1 previous window valid
+  };
```
rfk: Will this interfere with `lib/routes/totp.js`, which also configures global options on the `otplib` module?

vbudhram: > Will this interfere with lib/routes/totp.js

Yea it most likely will. I plan on making a lot of changes to the branch, currently it is for getting a basic user testing environment up.

PR #2076, opened 2019-08-05
fix(CI): check payments server before tests in CI

### LINKED ISSUE #1947: Check Payments Server's Health in test-content-server.sh
A webpack plugin was throwing a warning (log below) that was treated as an error in CI because the `CI` env var was set.  This was blocking the payments server from starting.  It went unnoticed because nothing was making requests to the payments server in CI at the time.

The issue that generated the warning was fixed in #1843.  But we should `check` the payments server's health in [`test-content-server.sh`](https://github.com/mozilla/fxa/blob/master/.circleci/test-content-server.sh#L8).

```
chunk 0 [mini-css-extract-plugin]
Conflicting order between:
 * css ./node_modules/css-loader/dist/cjs.js??ref--6-oneOf-5-1!./node_modules/postcss-loader/src??postcss!./node_modules/sass-loader/lib/loader.js??ref--6-oneOf-5-3!./src/components/DialogMessage.scss
 * css ./node_modules/css-loader/dist/cjs.js??ref--6-oneOf-5-1!./node_modules/postcss-loader/src??postcss!./node_modules/sass-loader/lib/loader.js??ref--6-oneOf-5-3!./src/components/Portal.scss
```

chenba:
Fixes #1947

### DIFF
--- a/.circleci/test-content-server.sh
+++ b/.circleci/test-content-server.sh
@@ -63,6 +63,7 @@ if grep -e "$MODULE" -e 'all' $DIR/../packages/test.list; then
     mozinstall firefox.tar.bz2
     # ensure email-service is ready
     check 127.0.0.1:8001/__heartbeat__
+    check 127.0.0.1:3031/__lbheartbeat__
     test_suite circle
   fi
 else


lmorchard:
Hmm, payments server should have its own set of tests, separate from test-content-server. Is this the right spot for it?

6a68:
oh, @lmorchard has a good point. maybe there's a better spot for that check?

chenba:
A few of the functional tests in `test-content-server` rely on the payments server being up.  That's the main reason for this.

lmorchard:
> A few of the functional tests in test-content-server rely on the payments server being up.

Ah, okay, then the cross-over is not so weird.

PR #2079, opened 2019-08-05
refactor(subscriptions): use account's settings data to check for active subscriptions

### LINKED ISSUE #1921: [content-server] Update Active Subscriptions Check for Subscription Management/Support to Use Account.settingsData
PR #1846 is adding `settingsData` to the Account model, which contains a list of subscriptions.  We could use that when checking for active subscriptions before displaying the Subscriptions subpanel or the support form.

chenba:
In the Support Form and the Subscriptions subpanel in Settings, we check
to see if the account has any active subscriptions.  Previously we were
fetching the subscriptions list directly from the active subscriptions
endpoint.  However, since a list of subscriptions is available in the
account model's settings data, we can make use of it, and save a few
requests.

Fixes #1921

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -1015,6 +1015,19 @@ const Account = Backbone.Model.extend(
     },
 
     /**
+     * Check to see if the account has any subscriptions.
+     *
+     * @returns {Promise} resolves to a bool.
+     */
+    hasSubscriptions() {
+      return this.settingsData().then(
+        settingsData =>
+          Array.isArray(settingsData.subscriptions) &&
+          settingsData.subscriptions.length > 0
+      );
+    },
+
+    /**
      * Fetch the account's list of active subscriptions.
      *
      * @returns {Promise} - resolves with a list of subscription objects.
--- a/packages/fxa-content-server/app/scripts/views/settings/subscription.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/subscription.js
@@ -13,6 +13,7 @@ import PaymentServer from '../../lib/payment-server';
 const View = FormView.extend({
   className: 'subscription',
   viewName: 'settings.subscription',
+  template: Template,
 
   events: {
     'click button': 'submit',
@@ -25,27 +26,8 @@ const View = FormView.extend({
     }
   },
 
-  template() {
-    return '';
-  },
-
   afterRender() {
-    return this._hasActiveSubscriptions().then(hasSubs => {
-      if (hasSubs) {
-        this.$el.html(this.renderTemplate(Template));
-      }
-    });
-  },
-
-  /**
-   * Check to see if the account has any active subscriptions.
-   *
-   * @returns {Promise} resolves to a bool.
-   */
-  _hasActiveSubscriptions() {
-    return this.getSignedInAccount()
-      .fetchActiveSubscriptions()
-      .then(subscriptions => subscriptions.length > 0);
+    return this.getSignedInAccount().hasSubscriptions();
   },
 
   submit() {
--- a/packages/fxa-content-server/app/tests/spec/models/account.js
+++ b/packages/fxa-content-server/app/tests/spec/models/account.js
@@ -2972,9 +2972,32 @@ describe('models/account', function() {
           return Promise.resolve(subs);
         });
 
-      account.fetchActiveSubscriptions().then(resp => {
+      return account.fetchActiveSubscriptions().then(resp => {
         assert.isTrue(subsStub.calledWith(token));
         assert.deepEqual(resp, subs);
+      });
+    });
+  });
+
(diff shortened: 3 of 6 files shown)


packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -2972,13 +2972,45 @@ describe('models/account', function() {
           return Promise.resolve(subs);
         });
 
-      account.fetchActiveSubscriptions().then(resp => {
+      return account.fetchActiveSubscriptions().then(resp => {
```
chenba: This change is not directly related to the issue/PR.  But this test was not running because it was missing the `return`.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -1014,6 +1014,19 @@ const Account = Backbone.Model.extend(
       return this._fxaClient.attachedClients(this.get('sessionToken'));
     },
 
+    /**
+     * Check to see if the account has any active subscriptions.
+     *
+     * @returns {Promise} resolves to a bool.
+     */
+    hasActiveSubscriptions() {
```
lmorchard: Nice little refactor, moving this to a utility method :+1:

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -1014,6 +1014,19 @@ const Account = Backbone.Model.extend(
...
+      return this.settingsData().then(
+        settingsData =>
+          Array.isArray(settingsData.subscriptions) &&
+          settingsData.subscriptions.some(s => s.status === 'active')
```
lmorchard: I don't think we want this filter on the list - just a `settingsData.subscriptions.length > 0` should do.

I *think* [the `status` will equal `past_due` or `incomplete`](https://stripe.com/docs/api/subscriptions/object#subscription_object-status) in the case where the user's card has expired or been declined on an automatic payment. That's when [we display a warning on the settings page](https://github.com/mozilla/fxa/issues/970) and the user will need to click a link in the warning or the button to manage subscriptions and update their billing method. 

Technically, their subscription will be *not* active at that point, but I think we'll still want to offer access to the management pages.

packages/fxa-content-server/app/tests/spec/models/account.js
shane-tomlinson: great catch!

packages/fxa-content-server/app/scripts/models/account.js
shane-tomlinson: > Technically, their subscription will be not active at that point, but I think we'll still want to offer access to the management pages.

This seems reasonable so that users will be able to re-activate their subscription or update payment methods.

packages/fxa-content-server/app/scripts/views/settings/subscription.js
```diff
@@ -30,22 +30,13 @@ const View = FormView.extend({
...
-   * @returns {Promise} resolves to a bool.
-   */
-  _hasActiveSubscriptions() {
     return this.getSignedInAccount()
-      .fetchActiveSubscriptions()
-      .then(subscriptions => subscriptions.length > 0);
+      .hasActiveSubscriptions()
+      .then(hasSubs => {
+        if (hasSubs) {
+          this.$el.html(this.renderTemplate(Template));
```
shane-tomlinson: If there are no subscriptions, does this panel appear empty? What you can do is instead of calling render here and having template() return an empty string is use the fact that if `afterRender` returns false, the panel shouldn't render:

```js

template: Template,
...
afterRender() {
  return this.getSignedInAccount()
  .hasActiveSubscriptions()
  .then(hasSubs => {
    // view is not rendered if hasSubs is falsey.
    return !! hasSubs;
  });
}
```

I think that should work within settings panels too, it definitely works in normal views.

chenba:
Pushed feedback based changes.  I've also renamed the method to `hasSubscriptions` to reflect that we don't check the `status` property for 'active'.  (Sounds a little awkward?  `isSubscriber`?)

packages/fxa-content-server/app/scripts/views/settings/subscription.js
```diff
@@ -25,27 +26,8 @@ const View = FormView.extend({
...
-  /**
-   * Check to see if the account has any active subscriptions.
-   *
-   * @returns {Promise} resolves to a bool.
-   */
-  _hasActiveSubscriptions() {
-    return this.getSignedInAccount()
-      .fetchActiveSubscriptions()
-      .then(subscriptions => subscriptions.length > 0);
+    return this.getSignedInAccount().hasSubscriptions();
```
chenba: The rendering related changes here is from @shane-tomlinson [pointing out that we can control view rendering in `afterRender`](https://github.com/mozilla/fxa/pull/2079#discussion_r311106508).

PR #2085, opened 2019-08-06
feat: add firestore type-checking

bbangert:
This uses a port of the browser-based typesafe-firestore that works with the Node version to provide strong typing of Firestore documents.


chenba:
A couple people brought up commit message formatting inconsistency recently... the commit message is missing a `scope`.

Issue #2087, opened 2019-08-06
Improve maintainability of auth server tests


philbooth:
Some of the tests in the auth server suffer from some concrete problems that make them quite painful to maintain. They're long files with repetitive boilerplate, which can make it hard to keep track of what's covered and what isn't. And when changes to the production code require knock-on changes to the tests, those knock-on changes are sometimes more difficult and take longer than the change to the production code.

This has led to gaps in our coverage, either where we haven't realised that things aren't covered or there wasn't time to cover them properly.

I'll update this description with specific examples as I work on it throughout the quarter.

- [ ] `test/local/routes/*.js`: Spotty coverage, clunky setup, brittle tests, hard to diagnose failures.
- [x] `test/local/senders/email.js`: Big ball of mud / maintenance nightmare. Bugs exist in the auth server because of it. Fixed in #2095 and #2148.
- [ ] `test/remote/*.js`: Multiple cases/assertions per test, long promise chains, hard to diagnose failures.

philbooth:
I'm not going to have time for this before I leave, so let's close it to stop it stinking up your backlog. If you guys have your own thoughts about improving the tests in the future, issues can be opened at that point.

PR #2095, opened 2019-08-06
 refactor(tests): eliminate repetitive boilerplate from local email tests 

philbooth:
Related to #2087. Tangentially related to #1779.

There are a number of problems with the auth server's email tests:

1. There's a lot of them, so the module is long, but that's exacerbated by all the repetitive boilerplate that's copy/pasted between test cases. It's hard to zero in on the substantive parts of each test, because of all the flannel surrounding the part you actually want to look at. The signal:noise ratio is all wrong.

2. Various assertions are strewn across the module, with no uniformity governing where they're located. That makes it hard to find things but worse, it makes it hard to audit what's covered and what isn't.

   Case in point, `postRemoveSecondaryEmail` is not covered by tests at all, even though assertions have explicitly been written for it. That's because it's not included in the `messageTypes` array that governs the behaviour of the outermost loop. The assertions are in a different part of the file so the discrepancy isn't obvious, I only happened to notice because I was pulling the file apart line-by-line while refactoring.

3. Some assertions are imprecise, only testing small substrings without any of the surrounding context. That's harmless in most cases but is a hiding place for bugs in some.

   For instance, right now there are no `utm_campaign` params on links in the `unblockCodeEmail` template. That's because of a bad key in the `templateNameToCampaign` map, but the existing tests don't catch it because they don't test whole URLs.

4. Some assertions are tautological, using implementation functions from the object under test to generate expected results to assert against. By definition, those assertions can never catch bugs in those functions, because they ipso facto suffer from the same bug.

To address these problems, I wanted to refactor the tests so that:

* All the assertions for a particular email live in one place.
* The assertions are concise enough that they can be eyeballed to audit test coverage.
* Boilerplate is not copied between test cases.
* It should be straightforward and obvious how to make the required test changes when changing email templates.

The approach I took was to move the assertions to static data, then iterate over them in the body of a single, generic, declarative test case. The downside of this is that any failures will always be reported from the same test, but I tried to mitigate that by including the email type and the property being tested in the assertion message. It seems to work okay, here's an example failure:

```
AssertionError: verifyLoginEmail: X-Link: expected 'http://confirm.127.0.0.1:3030/complete_signin?code=abc123&uid=uid&service=sync&utm_medium=email&utm_campaign=fx-new-signin&utm_content=fx-confirm-signin' to equal 'http://confirm.127.0.0.1:3030/complete_signin?code=abc123&service=sync&utm_medium=email&utm_campaign=fx-new-signin&utm_content=fx-confirm-signin'
+ expected - actual

-http://confirm.127.0.0.1:3030/complete_signin?code=abc123&uid=uid&service=sync&utm_medium=email&utm_campaign=fx-new-signin&utm_content=fx-confirm-signin
+http://confirm.127.0.0.1:3030/complete_signin?code=abc123&service=sync&utm_medium=email&utm_campaign=fx-new-signin&utm_content=fx-confirm-signin
```

The diff is long and possibly contentious, so I'll add some inline commentary too.

Opening as a draft because I have test-fatigue now and want to look at this with fresh eyes tomorrow before wasting anyone's time on a proper code review. There's probably some tidying up to do.


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -72,7 +72,7 @@ module.exports = function(log, config, oauthdb) {
     postAddAccountRecoveryEmail: 'account-recovery-generated',
     postRemoveAccountRecoveryEmail: 'account-recovery-removed',
     recoveryEmail: 'forgot-password',
-    unblockCode: 'new-unblock',
+    unblockCodeEmail: 'new-unblock',
```
philbooth: Bonus fix, we weren't setting `utm_campaign` because of this.

packages/fxa-auth-server/lib/senders/templates/post_verify_trailhead.txt
```diff
@@ -1,6 +1,6 @@
 {{t "Welcome to the Firefox family" }}
 
-{{t "You're signed in and ready to start exploring safely and securely." }}
+{{{t "You're signed in and ready to start exploring safely and securely." }}}
```
philbooth: Bonus fix, the `'` was being encoded as `&#x27;` even though this is supposed to be a plaintext email.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -16,216 +16,790 @@ const sinon = require('sinon');
...
+  type: 'secondary',
+  uaBrowser: 'Firefox',
+  uaBrowserVersion: '70.0a1',
+  uaOS: 'Windows',
+  uaOSVersion: '10',
+  uid: 'uid',
+  unblockCode: 'AS6334PK',
+};
+
+const COMMON_TESTS = new Map([
```
philbooth: These are the assertions that hold true for every email type.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -16,216 +16,790 @@ const sinon = require('sinon');
...
-  'postChangePrimaryEmail',
-  'postRemoveTwoStepAuthenticationEmail',
-  'postVerifySecondaryEmail',
-  'postConsumeRecoveryCodeEmail',
-  'postNewRecoveryCodesEmail',
-  'passwordResetAccountRecoveryEmail',
-  'postAddAccountRecoveryEmail',
-  'postRemoveAccountRecoveryEmail',
+// prettier-ignore
+const TESTS = new Map([
```
philbooth: And these are the assertions that are email-specific.

`// prettier-ignore` because I expressly want it to be easy to find a particular key by eye, but that's harder to do when everything is spread out over multiple lines and you have to scroll much further. I'm not bothered by the long line lengths because it's just data and as long as you can find the key that interests you, it's easy to scroll right at the point of interest. If it was code, I'd absolutely have a problem with the line lengths.

philbooth:
Forgot to mention, the other upside to these changes is the time taken to run this test module is reduced from ~18 seconds to ~6 seconds on my machine. It's because we're not paying the cost of instantiating the mailer a bajillion times in ` beforeEach`and then calling `mailer.stop` a bajillion times in `afterEach`.

We should be able to take that further in a follow-up and bring the whole caboodle in under a second.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -16,216 +16,790 @@ const sinon = require('sinon');
...
+    ['subject', { test: 'equal', expected: 'Verify Primary Email' }],
+    ['headers', new Map([
+      ['X-Link', { test: 'equal', expected: `${config.smtp.verifyPrimaryEmailUrl}?code=${MESSAGE.code}&uid=${MESSAGE.uid}&type=primary&primary_email_verified=${encodeURIComponent(MESSAGE.email)}&service=${MESSAGE.service}&utm_medium=email&utm_campaign=fx-welcome-primary&utm_content=fx-activate` }],
+      ['X-SES-MESSAGE-TAGS', { test: 'equal', expected: sesMessageTagsHeaderValue('verifyPrimaryEmail') }],
+      ['X-Template-Name', { test: 'equal', expected: 'verifyPrimaryEmail' }],
+      ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.verifyPrimaryEmail }],
+      ['X-Verify-Code', { test: 'equal', expected: MESSAGE.code }],
+    ])],
+    ['html', [
+      { test: 'include', expected: `href="${config.smtp.initiatePasswordChangeUrl}?email=${encodeURIComponent(MESSAGE.email)}&utm_medium=email&utm_campaign=fx-welcome-pri
```
philbooth: And there's a spot for trailhead-specific assertions too, although presumably these can all be unified once that special-casing is taken out.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -239,802 +813,171 @@ describe('lib/senders/email:', () => {
...
-    } else {
-      expectedTemplateName = type;
-    }
-
-    if (SUBJECT_ASSERTIONS.has(type)) {
-      it(`set the correct subject for ${type}`, () => {
-        mailer.mailer.sendMail = stubSendMail(emailConfig => {
-          assert.equal(emailConfig.subject, SUBJECT_ASSERTIONS.get(type));
+  it('declarative tests', async () => {
+    for (const [type, test] of TESTS) {
```
philbooth: This is the single, declarative test case that exercises all of the above assertion data.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -239,802 +813,171 @@ describe('lib/senders/email:', () => {
...
-            `${location.city}, ${location.country}`
-          );
-          assert.include(
-            emailConfig.text,
-            `${location.city}, ${location.country}`
-          );
-        });
-        return mailer[type](message);
-      });
+  it('verifyEmail handles no service', () => {
```
philbooth: From this point on the tests are all conventional. In another follow-up I plan to extract the tests for `selectEmailService` to a dedicated module as there are so many of them, it should help make things a bit more manageable.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -16,216 +16,810 @@ const sinon = require('sinon');
...
+        { test: 'equal', expected: config.smtp.sesConfigurationSet },
+      ],
+      ['X-Uid', { test: 'equal', expected: MESSAGE.uid }],
+    ]),
+  ],
+  [
+    'text',
+    [
+      // Ensure no HTML character entities appear in plaintext emails, &amp; etc
+      { test: 'notMatch', expected: /(?:&#x?[0-9a-f]+;)|(?:&[a-z]+;)/i },
```
philbooth: I thought this was quite a nice new assertion, which we never had before.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -2590,6 +2427,95 @@ if (config.get('redis.email.enabled')) {
...
+function sesMessageTagsHeaderValue(templateName, serviceName) {
+  return `messageType=fxa-${templateName}, app=fxa, service=${serviceName ||
+    'fxa-auth-server'}`;
+}
+
+function configHref(key, campaign, content, ...params) {
+  return `href="${configUrl(key, campaign, content, ...params)}"`;
+}
+
+function configUrl(key, campaign, content, ...params) {
```
philbooth: This is the new helper I added to reduce the URL boilerplate.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -16,216 +16,810 @@ const sinon = require('sinon');
...
+const TESTS = new Map([
+  ['downloadSubscriptionEmail', new Map([
+    ['subject', { test: 'equal', expected: 'Welcome to Secure Proxy!' }],
+    ['headers', new Map([
+      ['X-SES-MESSAGE-TAGS', { test: 'equal', expected: sesMessageTagsHeaderValue('downloadSubscription') }],
+      ['X-Template-Name', { test: 'equal', expected: 'downloadSubscription' }],
+      ['X-Template-Version', { test: 'equal', expected: SUBSCRIPTION_TEMPLATE_VERSIONS.downloadSubscription }],
+    ])],
+    ['html', [
+      { test: 'include', expected: configHref('privacyUrl', 'new-subscription', 'privacy') },
```
philbooth: And this is an example of the helper being invoked. If it obscures the guts of the assertion too much we can revert that commit, no worries.

philbooth:
Thanks @chenba!

PR #2121, opened 2019-08-08
feat(recovery): Clear recovery keys when resetting account

### LINKED ISSUE #2106: Clear recovery keys when resetting the account
From https://bugzilla.mozilla.org/show_bug.cgi?id=1569356#c30:

It looks like we delete recovery keys on the server if they are successfully consumed, but we leave them on the account if you reset your password *without* using a recovery key. That means a scenario like this is possible:

1.  Create account, with initial encryption key `k1`. Set up recovery key `rk1`.
2.  Start syncing, causing tokenserver to record `x-c-s-history=[hash(k1)]`.
3.  Reset your password without using the recovery key. Get a new encryption key `k2`.
4.  Sync again, causing tokenserver to record `x-c-s-history=[hash(k2), hash(k1)]`.
5.  Reset your password again, this time using recovery key `rk1` to revert your encryption key to its previous value `k1`.
6. Try to sync, sending `x-c-s=hash(k1)` to tokenserver.
7. Tokenserver sees that you are using an old key and refuses to process the request, on the assumption that you need to refresh your login state and get the most recent keys.  (This is a safety measure to prevent clients writing data with the wrong encryption key).

This scenario violates an assumption made by downstream consumers, that your encryption key should never revert to a previous value.  I don't think we've documented that guarantee explicitly anywhere, but we should, because it's an important part of helping consumers implement robust handling of encrypted data.

I think the right thing to do here is to `DELETE FROM recoveryKeys WHERE uid = uidArg` as part of the `resetAccount` stored procedure.

vbudhram:
Fixes #2106

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/mem.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mem.js
@@ -959,6 +959,7 @@ module.exports = function(log, error) {
       deleteByUid(uid, passwordChangeTokens);
       deleteByUid(uid, passwordForgotTokens);
       deleteByUid(uid, unverifiedTokens);
+      deleteByUid(uid, recoveryKeys);
 
       account.verifyHash = data.verifyHash;
       account.authSalt = data.authSalt;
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -959,6 +959,7 @@ module.exports = function(log, error) {
       deleteByUid(uid, passwordChangeTokens);
       deleteByUid(uid, passwordForgotTokens);
       deleteByUid(uid, unverifiedTokens);
+      deleteByUid(uid, recoveryKeys);
```
vbudhram: This is the only important change in this file, everything else is style changes.

philbooth:
@vbudhram, should this be targeted at the `train-143` branch for a point release?

vbudhram:
> maybe we should target the PR at train-143 and tag a point release?

I was on the fence about it, but I guess there isn't a harm. I'll push a point release 👍🏽

vbudhram:
Closing in favor of https://github.com/mozilla/fxa/pull/2123

PR #2122, opened 2019-08-08
fix(signin): Better invalid cached credential handling

### LINKED ISSUE #999: Login Error on Monitor

shane-tomlinson:
If a cached credential is stored and used on the signin
page, check whether the sessionToken is valid before rendering.
If the sessionToken is invalid, re-render and ask for a password.

If the sessionToken expires while the user is on the page,
also re-render, asking the user for their password.

This slightly modifies cached signin behavior. If a user
clicks "use a different account", we no longer remove the
stored sessionToken from localStorage, allowing Sync
accounts to remain the default account displayed on
the signin page for the next service, with a side
goal of minimizing the number of new sessions that
are created.

Part of this refactor is to re-render whenever an accessToken
or a sessionToken is removed from an account. Because
both can happen at the same time, I added a new method
"rerender" that is debounced so it's only processed once
within a 50ms timeframe if `rerender` is called multiple times.

fixes #999

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sign_in_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sign_in_password.mustache
@@ -32,7 +32,7 @@
       {{/isPasswordNeeded}}
 
       <div class="button-row">
-        <button id="submit-btn" type="submit">{{#t}}Sign in{{/t}}</button>
+        <button id="submit-btn" {{^isPasswordNeeded}}class="use-logged-in"{{/isPasswordNeeded}} type="submit">{{#t}}Sign in{{/t}}</button>
       </div>
 
       <div class="links">
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -158,11 +158,14 @@ const Account = Backbone.Model.extend(
     },
 
     discardSessionToken() {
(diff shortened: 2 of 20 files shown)


packages/fxa-content-server/app/scripts/templates/sign_in_password.mustache
```diff
@@ -32,7 +32,7 @@
       {{/isPasswordNeeded}}
 
       <div class="button-row">
-        <button id="submit-btn" type="submit">{{#t}}Sign in{{/t}}</button>
+        <button id="submit-btn" {{^isPasswordNeeded}}class="use-logged-in"{{/isPasswordNeeded}} type="submit">{{#t}}Sign in{{/t}}</button>
```
shane-tomlinson: This change was just so I could check for the expected button from the functional tests.

shane-tomlinson:
I'm calling this ready for review! @mozilla/fxa-devs - r?

PR #2039, opened 2019-08-01
refactor(js-client): convert fxa-js-client and its tests to commonjs

vladikoff:
* changes client structure
* moves to mocha

Fixes #1857

PR #2126, opened 2019-08-08
feat(deps): use ../ paths to fxa-shared and fxa-geodb

philbooth:
Related to #1860.

Tweaks the docker images for content, auth and oauth, so they pull `fxa-shared` and `fxa-geodb` in to the container directly. That means we no longer have to version and `npm publish` those packages separately. A similar approach should work for the js client too, I'm just waiting for #2039 to land first because the combination of webpack and AMD proved tricky to get right.

Note there was an additional fix I had to make in the oauth server, because of a dependency change when I updated shrinkwrap. As of version 8.5, [`jsonwebtoken.verify` supports a `complete` option](https://github.com/auth0/node-jsonwebtoken/pull/522) to return the broader response object instead of just the payload.

We were already passing `complete: true` to that function for some reason, which had no effect in previous versions. But with the dependency update, the behaviour changed and broke our assumptions. It was a surprising one to debug!

But I digress. This branch is deployed to a dev box here, for anyone who wants to take it for a spin:

https://deps.dev.lcip.org/

@mozilla/fxa-devs r?


packages/fxa-auth-server/fxa-oauth-server/lib/jwt.js
```diff
@@ -52,7 +52,6 @@ exports.verify = async function verify(jwt, options = {}) {
 
   return jwtverify(jwt, getKey, {
     algorithms: options.algorithms || [SIGNING_ALG],
-    complete: true,
```
philbooth: This was the suprising knock-on dependency change I had to make after updating shrinkwrap.

philbooth: Ref: https://github.com/auth0/node-jsonwebtoken/pull/522

packages/fxa-auth-server/Dockerfile-build
```diff
@@ -6,21 +6,35 @@ RUN apk add --no-cache git make gcc g++ linux-headers openssl python
...
+COPY fxa-auth-server/scripts/gen_keys.js scripts/gen_keys.js
+COPY fxa-auth-server/scripts/gen_vapid_keys.js scripts/gen_vapid_keys.js
+COPY fxa-auth-server/fxa-oauth-server/scripts/gen_keys.js fxa-oauth-server/scripts/gen_keys.js
 
 RUN npm ci --production && rm -rf ~app/.npm /tmp/*
 
-COPY . /app
+COPY fxa-auth-server /app
+
+COPY ["fxa-geodb", "../fxa-geodb/"]
```
shane-tomlinson: My Docker foo is very non-existent, why `[]` here and above no? Does that mean copy a directory?

packages/fxa-content-server/app/scripts/lib/popular-email-domains.json
```diff
@@ -1 +0,0 @@
-../../../node_modules/fxa-shared/email/popularDomains.json
```
shane-tomlinson: Very nice!

packages/fxa-auth-server/fxa-oauth-server/lib/jwt.js
shane-tomlinson: Huh... Was I using a newer version of jsonwebtoken while developing? This does indeed seem like the correct behavior now.

packages/fxa-auth-server/Dockerfile-build
philbooth: They're interchangeable afaik, the discrepancy is jsut because I copied this `COPY` from elsewhere and hadn't noticed the inconsistency.

philbooth: (we seem to use both styles)

PR #2127, opened 2019-08-09
feat(payments): Allow `npm test` to test files under 'server/'

LZoog:
Testing in `server/` is needed for my CSP changes in #2053.

We currently use `react-scripts` to run jest, but this will not allow for testing outside of the `src/` directory or multiple jest config files.

The options were:
1. Eject `create-react-app` to enable more customization
2. Write `server/` tests in a `__tests__` dir, inconsistent with other tests in this package
2. Leave jest config in `package.json` or in the root-level directory which would run tests in the `src/` directory, and create a separate `npm run test-server` command for `server/` tests which would run from a config living in `server/`
4. **Use `jest` to run all tests instead of `react-scripts` while keeping all tests running from one `npm test` command via the `--project` option and separate `server/jest.config.js` + `src/jest.config.js` files (this PR)**

In `tsconfig.json`, I could not get these tests to run without changing `"jsx": "preserve"` to `    "jsx": "react"`. Otherwise, I was getting errors in every test with `render(<...` (due to `<`) , and I tried many babel options with no success. (I may not need `"test-setup.ts"` in the includes there, will explore more tomorrow.)

It _appears_ this change forces a kind of stricter type checking in tests (correct me if I'm wrong). I could not get `src/routes/Subscriptions/index.test.tsx` to run without pulling in `config` as, otherwise, the schema for it was incorrect. I'm now running into the same issue in `Tooltip.test.tsx` from line 25: `getScreenInfo?: () => ScreenInfo | null;`
With the error:
```
 Type 'ScreenInfo | null' is not assignable to type 'ScreenInfo'.
            Type 'null' is not assignable to type 'ScreenInfo'.
```

I want to make sure I'm not going down a rabbit hole here.

~*_Also_*, I wanted opinions on using CSS modules instead of inline styles. I made one change to show what this will look like. We will need to get away from inline styles anyway, [unless we want to allow inline CSS styles in our CSP](https://github.com/mozilla/fxa/pull/2053/files#diff-135ecd18e06446a5495f649def76826cR47).~ Created a separate issue for discussion of this: https://github.com/mozilla/fxa/issues/2139

### DIFF
--- a/packages/fxa-payments-server/src/jest.config.js
+++ b/packages/fxa-payments-server/src/jest.config.js
@@ -1,0 +1,20 @@
+module.exports = {
+  preset: 'ts-jest',
+  moduleNameMapper: {
+    '\\.(css|scss)$': '<rootDir>/jest-style-mock.js',
+  },
+  collectCoverageFrom: [
+    '**/*.{js,jsx,ts,tsx}',
+    '!**/*.stories.*',
+    '!**/types.tsx',
+    '!**/*.d.ts',
+  ],
+  coverageThreshold: {
+    global: {
+      branches: 59,
+      functions: 49,
+      lines: 61,
+      statements: 60
+    },
+  },
+};
--- a/packages/fxa-payments-server/.eslintrc
+++ b/packages/fxa-payments-server/.eslintrc
@@ -4,6 +4,7 @@ plugins:
 extends:
   - plugin:fxa/client
   - prettier
+  - plugin:jest/recommended
 
 # Required for `import( )` webpack feature
 parser: babel-eslint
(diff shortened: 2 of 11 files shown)


packages/fxa-payments-server/src/jest.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+  preset: 'ts-jest',
+  moduleNameMapper: {
+    '\\.(css|scss)$': '<rootDir>/../node_modules/jest-css-modules',
```
LZoog: The alternative to this is to create a mock file containing an empty object for CSS/SCSS files to hit.

It does set us up for this, though: https://github.com/mozilla/fxa/issues/2139

lmorchard:
I did just notice one issue: `src/index.tsx` and `src/App.tsx` are not getting included in testing, for some reason. Noticed the test coverage went **WAY** up and squinted my eyes at the results. :) I can look into fixing that in my next test coverage PR though

PR #2143, opened 2019-08-09
fix(support-panel): add CSP frame-ancestors

### LINKED ISSUE #2142: [support-panel] needs x-frame-options: allow-from <some config> to work within zd domain
So, I updated the stage support panel to latest, which pulled in HSTS, a couple of logic fixes handling db response, and stronger query validation.

Now when Marty loads the panel, it's blank. Console says: `The resource at "<URL>" was blocked because content blocking is enabled`.

In adding HSTS, a set of default security related headers was also enabled, and I think `x-frame-options` needs to do an `allow-from` for a configurable domain name (or names).

chenba:
Hapi's route security configurations by default set 'DENY' for
X-Frame-Options.  The 'frame-ancestors' CSP directive obsoletes the
X-Frame-Options header.

Note that the module/Hapi plugin used to generate the CSP header also
produces default values for other directives; those are _not_ configured
in this patch.

Fixes #2142

### DIFF
--- a/packages/fxa-support-panel/config/index.ts
+++ b/packages/fxa-support-panel/config/index.ts
@@ -66,6 +66,16 @@ const conf = convict({
         format: ['default_fxa', 'dev_fxa', 'default', 'dev', 'short', 'tiny']
       }
     }
+  },
+  security: {
+    csp: {
+      frameAncestors: {
+        default: 'none',
+        doc:
+          'https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors',
+        env: 'CSP_FRAME_ANCESTORS'
+      }
+    }
   }
 });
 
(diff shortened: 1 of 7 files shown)


packages/fxa-support-panel/config/index.ts
```diff
@@ -66,6 +66,16 @@ const conf = convict({
...
       }
     }
+  },
+  security: {
+    csp: {
+      frameAncestors: {
+        default: 'none',
+        doc:
+          'https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors',
+        env: 'CSP_FRAME_ANCESTORS'
```
chenba: @jrgm Here's the new env var.

jrgm: Thanks @chenba 

@marty331 can you let me know what the origin domains should be in dev, stage and prod?

jrgm: > @marty331 can you let me know what the origin domains should be in dev, stage and prod?

Or if more then one, or a wildcard pattern, what that is.

marty331: @jrgm for Zendesk right?

jrgm: @marty331 - yes, for the zendesk domain(s) that frame the support-panel

marty331: It should always be `https://mozilladev.zendesk.com`

PR #2148, opened 2019-08-12
refactor(email): pull selectEmailServices out to its own module

philbooth:
Related to #2087.

The main change here is to pull the tests out for `selectEmailServices` so they're in a separate module, because there are so many of them and it makes the email tests module easier to work with. Overall, test coverage remains the same.

I also mopped up something left over from #2095, where I forgot to re-use the `setup` function I created in the email tests module.

@mozilla/fxa-devs r?


philbooth:
Tests pass on my machine but not in CI. Closing while I dig into it.

philbooth:
Just needed to set a longer timeout on the real redis tests...

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -1279,799 +1296,12 @@ describe('lib/senders/email:', () => {
       });
     });
   });
-
-  describe('single email address:', () => {
```
vbudhram: Digging the re-organization, 2k+ test files are hard follow.

Issue #535, opened 2019-03-25
Only serve compiled ES5 JS to user agents that need it


shane-tomlinson:
Right now we serve the same Babel compiled ES5 JS to all user-agents regardless of whether the user-agent supports new JS syntax or not. This has a couple of downsides:

1. The result is a much larger bundle, by ~ 100kb

<img width="830" alt="Screenshot 2019-03-25 at 12 55 21" src="https://user-images.githubusercontent.com/848085/54926534-64bdbe00-4f08-11e9-8958-de8a2f94a735.png">
<img width="839" alt="Screenshot 2019-03-25 at 12 56 44" src="https://user-images.githubusercontent.com/848085/54926535-64bdbe00-4f08-11e9-95fc-4c4c478d3f9a.png">

2. Stepping through code is more tedious because we have to rely on source maps, and sometimes the source maps don't map 1:1 to the actual code.

I could see this panning out like so:

1. Compile 2 JS bundles, one with "modern" syntax, one with ES5. 
  1a. Have to update [webpack.config.js](https://github.com/mozilla/fxa-content-server/blob/da94f530c8b5bb7477e207944ea0c6623af2f710/webpack.config.js) create two bundles, one with [babel-preset-es2015](https://github.com/mozilla/fxa-content-server/blob/da94f530c8b5bb7477e207944ea0c6623af2f710/webpack.config.js#L167) and one without.
2. Use [<script type="module">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) to serve "modern syntax" JS code, and `<script type="javascript" nomodule>` to serve ES5 code.
3. Win.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-696)
┆Issue Number: FXA-696

6a68:
Note that fxa-content-server issue #7094 has been migrated to https://github.com/mozilla/fxa/issues/532

PR #2154, opened 2019-08-12
refactor(account): Persist valid accounts to localStorage automatically on change

shane-tomlinson:
There are a lot of `user.setAccount`s sprinkled around the code,
it's easy to forget one whenever an account is updated and
for the account stored in localStorage to be stale.

This automatically persists an account to localStorage
as long as the account has both an email and uid.

Extraction from #2122

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
+++ b/packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
@@ -82,18 +82,18 @@ const View = FormView.extend(
     },
 
     submit() {
-      const account = this.getAccount();
-      const declinedSyncEngines = this._getDeclinedEngineIds();
-      const offeredSyncEngines = this._getOfferedEngineIds();
+      return Promise.resolve().then(() => {
+        const account = this.getAccount();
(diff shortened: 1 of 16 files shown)


packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
```diff
@@ -82,18 +82,18 @@ const View = FormView.extend(
     },
 
     submit() {
-      const account = this.getAccount();
-      const declinedSyncEngines = this._getDeclinedEngineIds();
-      const offeredSyncEngines = this._getOfferedEngineIds();
+      return Promise.resolve().then(() => {
```
philbooth: I notice quite a few of these in the diff. Would they be more aesthetic as `async` functions instead? (we can use those on the front-end, right?)

shane-tomlinson: If we did https://github.com/mozilla/fxa/issues/535, then I'd say "yeah, let's do it" since most users would be on UAs that support async/await.

PR #2158, opened 2019-08-12
refactor(signin): Pull account verify into a helper function for reuse

vbudhram:
Connects with #457 

Pulling this into it's own commit to make the review process easier. This PR adds signup utils for verify account function. No tests were added/removed since no underlying feature functionally didn't changed.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -21,7 +21,8 @@ module.exports = (
   config,
   customs,
   push,
-  verificationReminders
+  verificationReminders,
+  signupUtils
 ) => {
   const REMINDER_PATTERN = new RegExp(
     `^(?:${verificationReminders.keys.join('|')})$`
@@ -360,16 +361,7 @@ module.exports = (
       handler: async function(request) {
         log.begin('Account.RecoveryEmailVerify', request);
 
-        const {
-          code,
-          marketingOptIn,
-          newsletters,
-          reminder,
-          service,
-          style,
-          type,
-          uid,
-        } = request.payload;
+        const { code, service, type, uid } = request.payload;
 
         // verify_code because we don't know what type this is yet, but
         // we want to record right away before anything could fail, so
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/routes/utils/signup.js
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const P = require('../../promise');
+
+module.exports = (log, db, mailer, push, verificationReminders) => {
```
philbooth: If this is important enough to pull out to a separate module, should it also have some dedicated tests that exercise it directly? I'm wary of how thorough the test coverage is when it's one-layer removed.

And it makes auditing test coverage harder too, because we have to go looking in a completely different place to see if things have tests. And if want to make changes to this module later, it's much harder to test them by writing/changing endpoint tests than it is direct ones.

(kind of related to my testing OKR / #2087)

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -529,69 +521,11 @@ module.exports = (
...
-                      }
-
-                      return mailer.sendPostVerifyEmail(
-                        [],
-                        account,
-                        mailOptions
-                      );
-                    }
-                  });
+                return signupUtils.verifyAccount(
```
philbooth: The more I see of these various "utils" functions, the more I think they should be methods on proper domain objects in the `lib/` directory. Like this could be `account.verify`, then there could be an `account.signin` or whatever too...

packages/fxa-auth-server/lib/routes/utils/signup.js
vbudhram: Lets make a dent on that issue! I'm good with adding a few tests for this. We can probably expand on it once we update our test coverage tool and see where the gaps are.

vbudhram:
@philbooth Thanks for review! I've made test changes you recommended but will hold off on the proper domain objects parts.

Welcome to take re-look at changes, otherwise I'll merge when green.

PR #2161, opened 2019-08-12
feat(event-broker): add delete event propagation

### LINKED ISSUE #2110: [fxa-event-broker] add delete events
The event broker should propagate account delete events to RPs that a user has logged into.

bbangert:
Also refactors event handling to be a bit better structured for
additional event types.

Closes #2110

### DIFF
--- a/packages/fxa-event-broker/README.md
+++ b/packages/fxa-event-broker/README.md
@@ -1,5 +1,81 @@
 # fxa-event-broker
 
+The Firefox Accounts Event Broker (fxa-event-broker) is a webhook event delivery
+system for events in Firefox Accounts. These events are currently consumed from an
+internal priviledged queue that FxA services publish to, screened to determine what
+Relying Parties (RPs) should receive them, then distributed via webhooks using the
+[Security Event Token (SET)][set] format.
+
+This event broker also stores event metadata, tracking which RPs a user has logged
+into to screen event delivery to relevant RPs.
+
+## Event Format
+
+A relying party will get webhook calls for events. These events are encoded in
+SETs with the following formats. See the [SET RFC][set] for definitions and other
+examples.
+
+### Subscription State Change
+
+Sent when a user's subscription state has changed to RPs that provide the changed
+subscription capability.
+
+**NOTE**: There are strict requirements about subscription state change handling
+based on the `changeTime` as documented below.
+
+- Event Identifier
+  - `https://schemas.accounts.firefox.com/event/subscription-state-change`
+- Event Payload
+  - [Subscription Event Identifier]
+    - capabilities
+      - List of subscription capabilities
+    - isActive
+      - Boolean indicating if the subscription should be considered active or not
+        for the subscription capabilities provided.
+    - changeTime
+      - Time in seconds when the state change occured in the payment system.
+        This value MUST be tracked by the receiving system, and events with a
+        changeTime older than the last tracked time MUST be discarded.
+
+### Example Subscription State Change Event
+
+    {
+     "iss": "https://accounts.firefox.com/",
+     "sub": "FXA_USER_ID",
+     "aud": "REMOTE_SYSTEM",
+     "iat": 1565720808,
+     "jti": "e19ed6c5-4816-4171-aa43-56ffe80dbda1",
+     "events": {
+       "https://schemas.accounts.firefox.com/event/subscription-state-change": {
+           "capabilities": ["capability_1", "capability_2"],
+           "isActive": true,
+           "changeTime": 1565721242227
+       }
+     }
+
+### Delete User
+
+Sent when a user has been deleted from Firefox Accounts. RPs MUST delete all user
+records for the given user when receiving this event.
+
+- Event Identifier
+  - `https://schemas.accounts.firefox.com/event/delete-user`
+- Event Payload
+  - [Delete Event Identifier]
+    - `{}`
+
+### Example Delete Event
+
+    {
+     "iss": "https://accounts.firefox.com/",
+     "sub": "FXA_USER_ID",
+     "aud": "REMOTE_SYSTEM",
+     "iat": 1565720810,
+     "jti": "1b3d623a-300a-4ab8-9241-855c35586809",
+     "events": {
+       "https://schemas.accounts.firefox.com/event/delete-user": {}
+     }
+
 ## Deployment
 
 ### Testing a Webhook URL
@@ -16,12 +92,10 @@ Where `CAPABILITIES` is a comma-seperated string of capabilities to include.
 
 #### Usage
 
-    ```bash
-
     $ npm run build
     $ export LOG_FORMAT=pretty
     $ node dist/bin/simulate-webhook-call.js a9238ba https://example.com/webhook capability_1
     fxa-event-broker.simulateWebhookCall.INFO: webhookCall {"statusCode":200,"body":"ok\n"}
     $
 
-    ```
+[set]: https://tools.ietf.org/html/rfc8417
--- a/packages/fxa-event-broker/bin/simulate-webhook-call.ts
+++ b/packages/fxa-event-broker/bin/simulate-webhook-call.ts
@@ -36,6 +36,7 @@ async function main() {
   const uid = chance.hash();
   const jwtPayload = await jwt.generateSubscriptionSET({
     capabilities: args.capabilities.split(','),
+    changeTime: Math.trunc(Date.now() / 1000),
     clientId: args.clientId,
     isActive: true,
     uid
--- a/packages/fxa-event-broker/lib/api/proxy-controller.ts
+++ b/packages/fxa-event-broker/lib/api/proxy-controller.ts
@@ -8,6 +8,7 @@ import * as requests from 'request-promise-native';
 
 import { JWT } from '../jwts';
 import { ClientWebhookService } from '../selfUpdatingService/clientWebhookService';
+import { DELETE_EVENT, SUBSCRIPTION_UPDATE_EVENT } from '../serviceNotifications';
 import { proxyPayload } from './proxy-validator';
 
 export default class ProxyController {
@@ -43,12 +44,7 @@ export default class ProxyController {
       return h.response('Invalid message').code(400);
     }
 
-    const jwtPayload = await this.jwt.generateSubscriptionSET({
-      capabilities: message.capabilities,
-      clientId,
-      isActive: message.isActive,
-      uid: message.uid
-    });
+    const jwtPayload = await this.generateSET(message, clientId);
 
     const options: requests.OptionsWithUri = {
       headers: { Authorization: 'Bearer ' + jwtPayload },
@@ -87,5 +83,27 @@ export default class ProxyController {
     });
 
     return resp;
+  }
+
+  private async generateSET(message: any, clientId: string) {
+    switch (message.event) {
+      case SUBSCRIPTION_UPDATE_EVENT: {
+        return await this.jwt.generateSubscriptionSET({
+          capabilities: message.capabilities,
+          changeTime: message.changeTime,
+          clientId,
+          isActive: message.isActive,
+          uid: message.uid
+        });
+      }
+      case DELETE_EVENT: {
+        return await this.jwt.generateDeleteSET({
+          clientId,
+          uid: message.uid
+        });
+      }
+      default:
+        throw Error(`Invalid event: ${message.event}`);
+    }
   }
 }
--- a/packages/fxa-event-broker/lib/jwts.ts
+++ b/packages/fxa-event-broker/lib/jwts.ts
@@ -3,10 +3,18 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import * as jwtool from 'fxa-jwtool';
+import * as uuid from 'uuid';
 
 // SET Event identifiers
 export const SUBSCRIPTION_STATE_EVENT_ID =
   'https://schemas.accounts.firefox.com/event/subscription-state-change';
+
+export const DELETE_USER_EVENT_ID = 'https://schemas.accounts.firefox.com/event/delete-user';
+
+type deleteEvent = {
+  clientId: string;
+  uid: string;
+};
 
 type securityEvent = {
   uid: string;
@@ -23,6 +31,7 @@ type subscriptionEvent = {
   clientId: string;
   capabilities: string[];
   isActive: boolean;
+  changeTime: number;
 };
 
 type JWTConfig = {
@@ -47,10 +56,11 @@ export class JWT {
    * @param event
    */
   public generateSET(event: securityEvent): Promise<string> {
-    const now = Math.floor(Date.now() / 1000);
     const claims = {
       aud: event.clientId,
       events: event.events,
+      iat: Date.now() / 1000,
+      jti: uuid.v4(),
       sub: event.uid
     };
     return this.tokenKey.sign(claims);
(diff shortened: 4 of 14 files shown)


packages/fxa-event-broker/lib/serviceNotifications.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+function multiSchemaAttempt(
+  schemaTable: SchemaTable,
+  rawMessage: object
+): ServiceNotification | undefined {
+  const message = joi.attempt(rawMessage, BASE_MESSAGE_SCHEMA);
+  if (schemaTable[message.event]) {
+    return joi.attempt(message, schemaTable[message.event]);
+  } else {
+    return undefined;
```
6a68: Why are you explicitly returning undefined?

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,92 @@ class ServiceNotificationProcessor {
...
+        const topicName = this.topicPrefix + clientId;
+        const rpMessage = Object.assign(
+          {},
+          {
+            ...baseMessage,
+            capabilities
+          }
+        );
+
+        // TODO: Failures to publish due to missing queue should be Sentry reported.
```
6a68: This refactoring might be a good moment to file a followup issue and add the issue number to this TODO.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,92 @@ class ServiceNotificationProcessor {
...
     const body = JSON.parse(sqsMessage.Body || '{}');
-    let message: baseMessageSchema;
-    try {
-      message = joi.attempt(body, BASE_MESSAGE_SCHEMA);
-    } catch (err) {
-      this.logger.error('badBaseMessage', { err });
+    const message = ServiceNotification.from(this.logger, body);
+    if (!message) {
+      // Anything that isn't a message we want
+      this.logger.debug('unwantedMessage', { message });
```
6a68: I get that this is just refactoring existing code, but here it seems `message` will be `undefined` if a match isn't found, so we'll be logging `{ undefined }`. Is that the intent, or should `body` be logged instead? (I noticed that, in `handleLoginEvent`, when an unwanted message is logged, the actual message content is included.)

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,92 @@ class ServiceNotificationProcessor {
...
-        return;
+        await this.handleSubscriptionEvent(message);
+        break;
+      }
+      case DELETE_EVENT: {
+        await this.handleDeleteEvent(message);
+        break;
       }
+      default:
+        unhandledEventType(message);
```
6a68: If I understand correctly, this line means that unhandled messages of known type will throw, instead of slip by silently?

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Yup, a new issue was opened today to add Sentry reporting.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: That creates an exhaustive check, such that the code will not compile if this is reachable.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Ah, nice catch! That is indeed a bug, I should've logged the body, not the undefined message.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,92 @@ class ServiceNotificationProcessor {
...
+    const baseMessage = {
+      capabilities: [],
+      event: message.event,
+      isActive: message.isActive,
+      uid: message.uid
+    };
+
+    Object.entries(notifyClientIds)
+      .filter(([clientId]) => clientIds.includes(clientId))
+      .forEach(async ([clientId, capabilities]) => {
```
shane-tomlinson: Does `async` work within `.forEach`? I suppose nothing after the loop depends on it finishing.

packages/fxa-event-broker/lib/serviceNotifications.ts
bbangert: Forgot it wasn't needed. 😄

packages/fxa-event-broker/lib/jwts.ts
```diff
@@ -74,4 +80,21 @@ export class JWT {
...
+   * Generate a Delete Security Event Token.
+   *
+   * @param  delEvent
+   */
+  public generateDeleteSET(delEvent: deleteEvent): Promise<string> {
+    return this.generateSET({
+      clientId: delEvent.clientId,
+      events: {
+        [DELETE_EVENT_ID]: {
+          delete: true
```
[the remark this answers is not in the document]
bbangert: Yup, it would, there's an example of that here actually, I'll switch to it.
https://tools.ietf.org/html/rfc8417#section-2.1.2

packages/fxa-event-broker/lib/jwts.ts
```diff
@@ -8,6 +8,13 @@ import * as jwtool from 'fxa-jwtool';
 export const SUBSCRIPTION_STATE_EVENT_ID =
   'https://schemas.accounts.firefox.com/event/subscription-state-change';
 
+export const DELETE_EVENT_ID = 'https://schemas.accounts.firefox.com/event/delete';
```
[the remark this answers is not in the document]
bbangert: Indeed, but it shouldn't be a guessing game so I'll add user to this.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,92 @@ class ServiceNotificationProcessor {
...
+      const messageId = await this.pubsub
+        .topic(topicName)
+        .publishJSON({ event: message.event, uid: message.uid });
+      this.logger.debug('publishedMessage', { topicName, messageId });
+    });
+  }
+
+  private async handleLoginEvent(message: loginSchema) {
+    // Sync and some logins don't emit a clientId, so we have nothing to track
+    if (!message.clientId) {
```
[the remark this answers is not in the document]
bbangert: That would definitely be preferable, the Service Notifications message format page never mentioned that this happens either.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: It does not. I've switched this around, I assumed async was handled more than it was. 😄

packages/fxa-event-broker/lib/jwts.ts
```diff
@@ -47,10 +55,11 @@ export class JWT {
    * @param event
    */
   public generateSET(event: securityEvent): Promise<string> {
-    const now = Math.floor(Date.now() / 1000);
     const claims = {
       aud: event.clientId,
       events: event.events,
+      iat: Date.now() / 1_000,
```
6a68: `1_000` looks like a typo

packages/fxa-event-broker/lib/serviceNotifications.ts
```diff
@@ -0,0 +1,124 @@
...
+    return joi.attempt(message, schemaTable[message.event]);
+  }
+}
+
+export const ServiceNotification = {
+  from(logger: Logger, rawMessage: object): ServiceNotification {
+    try {
+      const validMessage = multiSchemaAttempt(eventSchemas, rawMessage);
+      if (!validMessage) {
+        return undefined;
```
6a68: no need to explicitly `return undefined` (unless this is TS magic at work)

packages/fxa-event-broker/lib/serviceNotifications.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+      logger.debug('from.sqsMessage', { msg: validMessage });
+      return validMessage;
+    } catch (err) {
+      logger.error('from.sqsMessage', { message: 'Invalid message', err });
+      return undefined;
```
6a68: `return undefined` again

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,93 @@ class ServiceNotificationProcessor {
     this.webhookService.stop();
   }
 
+  private async handleDeleteEvent(message: deleteSchema) {
+    const clientIds = await this.db.fetchClientIds(message.uid);
+    clientIds.forEach(async clientId => {
```
6a68: Looking again at this function, the async/foreach thing pops up here too. I think a simple for loop would allow the async calls to be managed correctly with minimal rewriting.

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -130,85 +87,93 @@ class ServiceNotificationProcessor {
...
+      const topicName = this.topicPrefix + clientId;
+      const messageId = await this.pubsub
+        .topic(topicName)
+        .publishJSON({ event: message.event, uid: message.uid });
```
6a68: What happens if the message isn't sent successfully?

packages/fxa-event-broker/lib/jwts.ts
bbangert: I wanted this number to be in seconds, Date.now() is milliseconds so I divided by 1000. The underscore is a Typescript numeric separator. I suppose we could discuss how many zero's warrants using them, I'm pretty used to them in Rust as well.

packages/fxa-event-broker/lib/serviceNotifications.ts
bbangert: This is being used for Typescripts type refinement, to prove under the block that validMessage is no longer undefined. However I can drop a few lines of code by inverting the condition and letting the implicit undefined return value occur.

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Then a processing_error occurs, which logs the error and the message will not be acknowledged to AWS SQS. It will continue retrying.

packages/fxa-event-broker/lib/jwts.ts
6a68: I think it's probably better to just use `1000`, that's used a lot throughout the codebase

packages/fxa-event-broker/lib/notificationProcessor.ts
bbangert: Ah yes, fixed that to use a for..of loop.

packages/fxa-event-broker/lib/api/proxy-controller.ts
```diff
@@ -88,4 +84,25 @@ export default class ProxyController {
 
     return resp;
   }
+
+  private async generateSET(message: any, clientId: string) {
+    switch (message.event) {
+      case SUBSCRIPTION_UPDATE_EVENT: {
+        return await this.jwt.generateSubscriptionSET({
+          capabilities: message.capabilities,
```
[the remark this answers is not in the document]
bbangert: It will be shortly! ;)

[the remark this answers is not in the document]
bbangert: Docs added!

packages/fxa-event-broker/lib/serviceNotifications.ts
bbangert: Fixed! :)

packages/fxa-event-broker/lib/notificationProcessor.ts
```diff
@@ -78,6 +34,7 @@ class ServiceNotificationProcessor {
   private readonly capabilityService: ClientCapabilityService;
   private readonly webhookService: ClientWebhookService;
   private readonly pubsub: PubSub;
+  private readonly topicPrefix = 'rpQueue-';
```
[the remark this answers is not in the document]
bbangert: That's the eventual goal, but I didn't want to plumb it out in this PR.

packages/fxa-event-broker/README.md
```diff
@@ -1,5 +1,81 @@
...
+
+## Event Format
+
+A relying party will get webhook calls for events. These events are encoded in
+SETs with the following formats. See the [SET RFC][set] for definitions and other
+examples.
+
+### Subscription State Change
+
+Sent when a users subscription state has changed to RPs that provide the changed
```
chenba: proofread nit: "user's"

packages/fxa-event-broker/bin/simulate-webhook-call.ts
```diff
@@ -36,6 +36,7 @@ async function main() {
   const uid = chance.hash();
   const jwtPayload = await jwt.generateSubscriptionSET({
     capabilities: args.capabilities.split(','),
+    changeTime: Math.trunc(Date.now() / 1000),
```
6a68: The docs above say `changeTime` is time in milliseconds, but here you divide by 1000, converting to seconds.

packages/fxa-event-broker/README.md
```diff
@@ -1,5 +1,81 @@
...
+  - [Subscription Event Identifier]
+    - capabilities
+      - List of subscription capabilities
+    - isActive
+      - Boolean indicating if the subscription should be considered active or not
+        for the subscription capabilities provided.
+    - changeTime
+      - Time in milliseconds when the state change occured in the payment system.
+        This value MUST be tracked by the receiving system, and events with a
+        changeTime older than the last tracked time MUST be discarded.
```
6a68: If messages can arrive out of order, should we future-proof this event system by making a timestamp mandatory on all event types? (Fine to defer to a follow-up issue)

packages/fxa-event-broker/lib/serviceNotifications.ts
```diff
@@ -0,0 +1,126 @@
...
+  .object()
+  .keys({
+    event: joi
+      .string()
+      .valid(SUBSCRIPTION_UPDATE_EVENT)
+      .required(),
+    eventCreatedAt: joi
+      .number()
+      .integer()
+      .required(),
```
6a68: How about using [joi's timestamp validator](https://github.com/hapijs/joi/blob/v9.0.0-3/API.md#datetimestamptype) here, to ensure we only accept millisecond-based timestamps? `Joi.date().timestamp('javascript')`

packages/fxa-event-broker/bin/simulate-webhook-call.ts
bbangert: The seconds value is correct, changing the docs.

packages/fxa-event-broker/README.md
bbangert: There is a timestamp, but the timestamp is the generation of the security event token, which does not correspond to the originating event generation this reflects way back in Stripe. The Stripe event time is what appears here as `changeTime` which is the value that needs to be used to determine if a message should be ignored. Since these message may be not only out of order, but also encounter delays it didn't make sense to use that value as the SET's `iat`, as that might be confusing to debug and seeing tokens you were just sent with times 20+ seconds in the past. Thus adding the value here separately purely for this use-case.

packages/fxa-event-broker/lib/serviceNotifications.ts
bbangert: The number here is actually in seconds (its Stripe's event timestamp). I'm actually hoping to get rid of all these joi schemas we have floating about and use JSON-Schema, but that's coming up. 😄

PR #2163, opened 2019-08-12
fix(payments): Fix payments compilation

### LINKED ISSUE #2162: payments server not starting on circle
See https://circleci.com/gh/mozilla/fxa/60907#tests/containers/2

shane-tomlinson:
* Remove the unused `State`

fixes #2162

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-payments-server/src/store/utils.tsx
+++ b/packages/fxa-payments-server/src/store/utils.tsx
@@ -1,5 +1,5 @@
 import { ActionType as PromiseActionType } from 'redux-promise-middleware';
-import { Action, State, FetchState } from './types';
+import { Action, FetchState } from './types';
 
 type MappedObject = { [propName: string]: any };
 export const mapToObject = (


packages/fxa-payments-server/tsconfig.json
```diff
@@ -13,7 +17,9 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "preserve"
```
lmorchard: The new test setup that runs both frontend and server-side tests needs this `"jsx": "react"` setting, if I recall.

lmorchard: Oh, I see, but other bits from create-react-app (i.e. `react-scripts build`) want to switch it back :/ Might need to do some more headscratching on this

shane-tomlinson: I'll change it back.

PR #2173, opened 2019-08-13
chore(subhub): cache the plans list in Redis

### LINKED ISSUE #2075: [fxa-payments-server] Cache plans response
Stripe has API limits that it'd be good to stay under. One of the calls that is likely to be hit the most is the SubHub plans API call to get a list of current plans. This call is hit for every single request to see the payment page, which could be quite a bit even if users don't proceed further.

Ideally we'd cache this for a few minutes with memcached (as we cache other things in FxA).

chenba:
Fixes #2075

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -63,6 +63,7 @@ function run(config) {
   let database = null;
   let customs = null;
   let oauthdb = null;
+  let subhub = null;
 
   function logStatInfo() {
     log.stat(server.stat());
@@ -83,6 +84,7 @@ function run(config) {
         database = db;
         const bounces = require('../lib/bounces')(config, db);
         oauthdb = require('../lib/oauthdb')(log, config);
+        subhub = require('../lib/subhub/client')(log, config);
 
         return require('../lib/senders')(
           log,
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/bin/key_server.js
```diff
@@ -105,7 +107,8 @@ function run(config) {
             Password,
             config,
             customs,
-            zendeskClient
+            zendeskClient,
+            subhub
```
chenba: Moved `subhub` out here so we can call its `close()` when shutting down the server.

packages/fxa-auth-server/lib/subhub/client.js
```diff
@@ -178,8 +187,41 @@ module.exports = function(log, config) {
...
+        promises.push(redis.close());
+      }
+      return await P.all(promises);
+    },
+
     async listPlans() {
-      return api.listPlans();
+      const cacheKey = 'listPlans';
+      const ttlSeconds = config.subhub.plansCacheTtlSeconds;
+      const cacheIsEnabled = ttlSeconds && redis;
```
philbooth: No biggie, but I feel like these should all be defined earlier during initialisation, instead of being method-local. Elsewhere we tend to pull properties off `config` early as a defensive measure, to eliminate any kind of spooky-action-at-a-distance.

PR #2175, opened 2019-08-13
chore(style): remove jank from payment server load

johngruen:
This patch fixes a loading jank on the payment server with the following fixes:

1. Brings minimal critical-path styles into a `<style>` to prevent the loading spinner from rendering incorrectly at the top of the document.
2. Updates Index page html to better match that of the content server.


packages/fxa-payments-server/public/index.html
```diff
@@ -19,23 +19,57 @@
...
+
+        to {
+          transform: rotate(360deg);
+        }
+      }
+    </style>
   </head>
 
   <body>
-    <div id="root">
```
lmorchard: `id="root"` is where the React app renders, once it's loaded. The contents here are just kind of pre-loading stuff until React comes in and replaces it all. 

Moving the loading spinner above this in the DOM means the app is never visible. There's also a bunch of repeated elements here (static-footer, stage, etc) because React renders them in root.

Capture

johngruen:
Thanks Les...yeah, i'm going to revisit this without moving around stuff in the index so much...I'm not adding inline styles at all here. is internal CSS also verboten b/c of CSP?

lmorchard:
> is internal CSS also verboten b/c of CSP?

Yeah, I think it's supposed to be, but we are kind of punting on it because Stripe components play with both if I recall

johngruen:
it looks like you can set a nonce in CSP rules and in a <style> tag that would make it acceptable...i have no idea how hard that would be...documentation is a little way down in [this section](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Unsafe_inline_styles) of the CSP MDN page

shane-tomlinson:
> Thanks Les...yeah, i'm going to revisit this without moving around stuff in the index so much...I'm not adding inline styles at all here. is internal CSS also verboten b/c of CSP

On the payments server `unsafe-inline` is used because the Stripe component uses inline style attributes and we can't control what they do. If that's the case, you *should* be fine to use a normal style tag here. Hopefully that makes using a nonce/sha moot:

> it looks like you can set a nonce in CSP rules and in a <style> tag that would make it acceptable...i have no idea how hard that would be...documentation is a little way down in this section of the CSP MDN page

johngruen:
@lmorchard i kind of want to revisit this in 148... it looks like @shane-tomlinson says that using a style tag is fine.

johngruen:
actually I think i can do this by just getting rid of the pre-payload spinner

packages/fxa-payments-server/src/components/LoadingSpinner.tsx
```diff
@@ -6,7 +6,7 @@ const spinnerImage =
 
 export const LoadingSpinner = () => (
   <div id="loading-spinner" data-testid="loading-spinner">
-    <img alt="Loading" src={spinnerImage} />
```
johngruen: This is kinda controversial i guess, but I don't think this image sticks around long enough for the alt text to have semantic importance (it could easily be a CSS background), and sometimes the word "Loading" flashes briefly on the screen on refresh

LZoog: For this to be technically valid HTML, we should at least set it to an empty string (`alt=''`).

johngruen: Kk, I can fix that

PR #2177, opened 2019-08-13
feat(codes): Add routes and emails to verify account by otp code

vbudhram:
Connects with #457 

- [x] Add routes (/session/verify_code, /session/resend_code)
- [x] Update route (/account/create
- [x] Add email templates
- [x] Add local tests
- [x] Add remote tests

I think this PR is ready for review. I've updated this based on the feature doc at https://docs.google.com/document/d/1i003YoL4cchn8fOpA9yb0ntnD-rHhq4zO-WMyX6Y_IE/edit#. A few notes

-  `/session/verify_code` not backwards compatible with email links (added too much complexity)
- When session become verified, they use the `email-2fa` level since that is what we currently use for signin codes
- Sign-up codes expired between 10-20minutes

Rendered email:
Screen Shot 2019-08-14 at 2 45 42 PM

### DIFF
--- a/packages/fxa-auth-server/lib/error.js
+++ b/packages/fxa-auth-server/lib/error.js
@@ -92,6 +92,7 @@ const ERRNO = {
   SUBSCRIPTION_ALREADY_CANCELLED: 180,
   REJECTED_CUSTOMER_UPDATE: 181,
   REFRESH_TOKEN_UNKNOWN: 182,
+  INVALID_EXPIRED_OTP_CODE: 183,
 
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
@@ -1196,6 +1197,15 @@ AppError.unknownRefreshToken = () => {
     error: 'Bad Request',
     errno: ERRNO.REFRESH_TOKEN_UNKNOWN,
     message: 'Unknown refresh token',
+  });
+};
+
+AppError.invalidOrExpiredOtpCode = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.INVALID_EXPIRED_OTP_CODE,
+    message: 'Invalid or expired OTP code',
   });
 };
 
--- a/packages/fxa-auth-server/lib/senders/index.js
+++ b/packages/fxa-auth-server/lib/senders/index.js
@@ -115,6 +115,21 @@ module.exports = (log, config, error, bounces, translator, oauthdb, sender) => {
           );
         });
       },
+      sendVerifyShortCode: function(emails, account, opts) {
+        // This function differs from `sendVerifyCode` since it sends a code that
+        // is expected to be entered into a input field rather than a link to verify
+        // the account. It is expected to be much shorter than the `emailCode`.
+        const primaryEmail = account.email;
+        return getSafeMailer(primaryEmail).then(mailer => {
+          return mailer.verifyShortCodeEmail(
+            Object.assign({}, opts, {
+              acceptLanguage: opts.acceptLanguage || defaultLanguage,
+              email: primaryEmail,
+              uid: account.uid,
+            })
+          );
+        });
+      },
       sendVerifyLoginEmail: function(emails, account, opts) {
         return getSafeMailerWithEmails(emails).then(result => {
           const mailer = result.ungatedMailer;
--- a/packages/fxa-auth-server/test/local/routes/account.js
+++ b/packages/fxa-auth-server/test/local/routes/account.js
@@ -6,7 +6,7 @@
 
 const sinon = require('sinon');
 
-const { assert } = require('chai');
+const assert = { ...sinon.assert, ...require('chai').assert };
 const mocks = require('../../mocks');
 const getRoute = require('../../routes_helpers').getRoute;
 const proxyquire = require('proxyquire');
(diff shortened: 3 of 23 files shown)


packages/fxa-auth-server/lib/senders/index.js
```diff
@@ -115,6 +115,21 @@ module.exports = (log, config, error, bounces, translator, oauthdb, sender) => {
           );
         });
       },
+      sendVerifyShortCode: function(emails, account, opts) {
+        // This function defers from `sendVerifyCode` since it sends a code that
+        // is expected to be entered into a input field rather than a link to verify
+        // the account.
```
philbooth: With this comment in mind, and to keep the nomenclature consistent at every level, should we call the partial/templates `verify_short_code.*` and the method above `verifyShortCodeEmail`?

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -6,7 +6,7 @@
 
 const sinon = require('sinon');
 
-const { assert } = require('chai');
+const assert = { ...sinon.assert, ...require('chai').assert };
```
philbooth: Oh nice. As you're doing this more than once, should we pull it into in a little `test/assert.js` module just to eliminate the repetition?

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -947,6 +948,37 @@ describe('/account/create', () => {
...
+            args[2].location,
+            mockRequest.app.geo.location,
+            'location set'
+          );
+        });
+      });
+    });
+  });
+
+  it('should send sign-up code when `');
```
philbooth: ❓

packages/fxa-auth-server/test/local/routes/session.js
```diff
@@ -56,6 +79,24 @@ function hexString(bytes) {
...
+
+async function assertFailure(asyncFn, assertFn) {
+  let err;
+  try {
+    await asyncFn;
+  } catch (e) {
+    err = e;
+  }
+  assertFn(err);
+}
```
philbooth: And if you do decide to create `test/assert.js`, maybe you could put this in there too, I noticed it's defined again further down.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -761,6 +761,38 @@ const TESTS = new Map([
...
+    ['headers', new Map([
+      ['X-SES-MESSAGE-TAGS', { test: 'equal', expected: sesMessageTagsHeaderValue('verifyEmail') }],
+      ['X-Template-Name', { test: 'equal', expected: 'verifyCodeEmail' }],
+      ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.verifyCodeEmail }],
+      ['X-Verify-Short-Code', { test: 'equal', expected: MESSAGE.code }],
+    ])],
+    ['html', [
+      { test: 'include', expected: configHref('privacyUrl', 'welcome', 'privacy') },
+      { test: 'include', expected: configHref('supportUrl', 'welcome', 'support') },
+      { test: 'include', expected: 'Is this you signing up?' },
```
philbooth: Fwiw, I saw most value in these kinds of assertion for ensuring data has been substituted in the right position (and with the right value obviously). Also if content is inside conditional blocks. So this assertion I'm not sure is adding a great deal...

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -761,6 +761,38 @@ const TESTS = new Map([
...
+      { test: 'notInclude', expected: 'utm_source=email' },
+    ]],
+    ['text', [
+      { test: 'include', expected: `Mozilla Privacy Policy ${configUrl('privacyUrl', 'welcome', 'privacy')}` },
+      { test: 'include', expected: `For more information, please visit ${configUrl('supportUrl', 'welcome', 'support')}` },
+      { test: 'include', expected: 'Is this you signing up?' },
+      { test: 'include', expected: `IP address: ${MESSAGE.ip}` },
+      { test: 'include', expected: `${MESSAGE.location.city}, ${MESSAGE.location.stateCode}, ${MESSAGE.location.country} (estimated)` },
+      { test: 'include', expected: 'If yes, use this verification code:' },
+      { test: 'include', expected: MESSAGE.code },
```
philbooth: ...whereas these two I think would be improved by combining them so that we're asserting the correct position of `MESSAGE.code`:

```js
{ test: 'include', expected: `If yes, use this verification code:\n${MESSAGE.code}` },
```

(doesn't work for the `html` assertions sadly, but just doing it here gives us some extra coverage)

packages/fxa-auth-server/lib/error.js
```diff
@@ -1199,6 +1200,15 @@ AppError.unknownRefreshToken = () => {
   });
 };
 
+AppError.invalidOrExpiredOtpCode = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.INVALID_EXPIRED_OTP_CODE,
+    message: 'Invalid or expired otp code',
```
philbooth: Should we uppercase `OTP` in this error message?

packages/fxa-auth-server/lib/senders/index.js
philbooth: Also typo, just noticed: `s/defers/differs/`

vbudhram:
@philbooth Thanks for review!

Issue #1797, opened 2019-07-16
Reduce the size of the JWT access token


shane-tomlinson:
Right now JWT access tokens come in at a hefty 800+ bytes. That's pretty hefty, especially if they are transferred often. A stated target is closer to 500 bytes.

Breaking down a token from prod:

* ~ 70 bytes are used for the header
* ~ 380 bytes are for the body
* ~ 340 bytes are for the signature

The header contains: `{\"alg\":\"RS256\",\"typ\":\"at+JWT\",\"kid\":\"2016-01-19-1\"}`

The body contains the following fields:
* `aud` - audience, right now the client id
* `client_id` - the client id
* `exp` - the time the token expires
* `iat` - the time the token was created
* `iss` - issuer, the domain of the server that created the token
* `jti` - a token ID, used as a key into the database for the original access token
* `scope` - scopes the token is granted. Can be quite lengthy
* `sub` - the userid for the user, 16 characters long

We don't *really* want to drop required fields, the fields that are optional are `iat` and `jti`. We can't drop `jti` just yet because this is the access token value that is used to access the token in the database. We could drop `iat`.

`scope` can be reduced by using short scope names, breaking [the rule we set for ourselves](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/fxa-oauth-server/docs/scopes.md#short-name-scope-values).

The biggest bang for the buck might be changing the [signing key to `Ed25519`](https://github.com/mozilla/fxa/issues/1794).

rfk:
We could shorten our `kid` to shave a few more bytes, at the cost of some readability.

Issue #1794, opened 2019-07-16
Use Ed25519 to sign id tokens and JWT access tokens


shane-tomlinson:
RSA generates large signatures, Ed25519 significantly smaller and will save some bytes.

rfk:
Just poking at the possibilities here, it looks like `Ed25519` support is available natively from node v12, based on the history notes on https://nodejs.org/api/crypto.html. Using it for JWTs doesn't yet appear to be supported in either [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken/issues/115) or [node-jose](https://github.com/cisco/node-jose/issues/219).

Personally, I think we'd get most of the bang for a lot less buck if we move to `alg=ES256` in the short term and investigate `alg=Ed25519` longer-term.

From https://jwt.io, their example `RS256` JWT is 471 bytes long:

```
eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.POstGetfAytaZS82wHcjoTyoqhMyxXiWdR7Nn7A29DNSl0EiXLdwJ6xC6AfgZWF1bOsS_TuYI3OG85AmiExREkrS6tDfTQ2B3WXlrr-wp5AokiRbz3_oB4OxG-W9KcEEbDRcZc0nH3L7LzYptiy1PtAylQGxHTWZXtGz4ht0bAecBgmpdgXMguEIcoqPJ1n3pIWk_dUZegpqx0Lka21H6XxUTxiy8OcaarA8zdnPUnV6AmNP3ecFawIFYdvJB_cm-GvpCSbr8G8y_Mllj8f4x9nBH8pQux89_6gUY618iYv7tuPWBFfEbLxtF2pZS6YC1aSfLQxeNe8djT9YjpvRZA
```

While the same data using `ES256` is only 215 bytes long:

```
[...]

PR #2179, opened 2019-08-13
[WIP] chore(node): Use node 12 for the auth/oauth servers.

shane-tomlinson:
@linuxwolf suggests we migrate from node-jose to
https://github.com/panva/jose for our JW* needs
since it is more fully supported.

Moving to this library would also allow us to sign
JWTs using Ed25519 as well as make it simpler to
migrate to ES256. However, the lirary requires
node 12.

This is an attempt at using node 12 on the
OAuth and Auth servers.

issue #1797
issue #1794

### DIFF
--- a/packages/fxa-auth-server/test/local/redis.js
+++ b/packages/fxa-auth-server/test/local/redis.js
@@ -32,7 +32,7 @@ describe('lib/redis:', () => {
     };
     fxaShared = sinon.spy(() => redis);
     wrapper = proxyquire(`${LIB_DIR}/redis`, {
-      '../../../fxa-shared/redis': fxaShared,
+      '../../fxa-shared/redis': fxaShared,
     })(config, log);
   });
 
--- a/packages/fxa-customs-server/package.json
+++ b/packages/fxa-customs-server/package.json
@@ -20,7 +20,7 @@
     "format": "prettier '**' --write"
   },
   "dependencies": {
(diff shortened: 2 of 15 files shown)


jbuck:
I should really switch these off alpine...

packages/fxa-auth-server/test/local/redis.js
```diff
@@ -32,7 +32,7 @@ describe('lib/redis:', () => {
     };
     fxaShared = sinon.spy(() => redis);
     wrapper = proxyquire(`${LIB_DIR}/redis`, {
-      '../../../fxa-shared/redis': fxaShared,
+      '../../fxa-shared/redis': fxaShared,
```
shane-tomlinson: I'm not really sure why this change was necessary, I haven't looked to see if node 12 resolves modules differently to node 10.

packages/fxa-customs-server/package.json
```diff
@@ -20,7 +20,7 @@
     "format": "prettier '**' --write"
   },
   "dependencies": {
-    "@google-cloud/pubsub": "0.24.1",
+    "@google-cloud/pubsub": "^0.31.0",
```
shane-tomlinson: The previous version of pubsub was incompatible with node 12 and this update fixed that.

shane-tomlinson:
Closing this until I have more time.

PR #2185, opened 2019-08-13
fix(fxa-content-server): fix #1907 - change server.csp log schema

### LINKED ISSUE #1907: Change server.csp log schema
Follow-up on #1546. Right now a CSP log looks like:

```
{
	"EnvVersion": "2.0",
	"Fields": {
		"agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36",
		"blocked": "data",
		"referrer": "https://mail.google.com/mail/u/1/",
		"sample": "",
		"source": "",
		"time": "2019-07-23T12:00:00.000Z",
		"violated": "font-src"
	},
	"Logger": "fxa-content-server",
	"Pid": 6,
	"Severity": 6,
	"Timestamp": 1563884830047000000,
	"Type": ".server.csp"
}
```

The two things that need to change are:

1. `Fields.time` being a string, when BigQuery expects it to be an integer (from the amplitudeEvent logs)
2. `Type` starting with a `.`

LZoog:
fix for #1907 

As discussed with @jbuck, `Fields.time` is removed with this PR because BigQuery expects an integer and cannot be changed, but if change this to an integer we'd just be getting duplicate data since we're already capturing `Timestamp`.

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/post-csp.js
+++ b/packages/fxa-content-server/server/lib/routes/post-csp.js
@@ -9,7 +9,7 @@
 'use strict';
 const _ = require('lodash');
 const joi = require('joi');
-const logger = require('../logging/log')('');
+const logger = require('../logging/log')();
 const url = require('url');
 const validation = require('../validation');
 
@@ -80,7 +80,6 @@ module.exports = function(options = {}) {
         referrer: stripPIIFromUrl(report['referrer']),
         sample: report['script-sample'],
         source: stripPIIFromUrl(report['source-file']),
-        time: today.toISOString(),
         violated: report['violated-directive'],
       };
 


packages/fxa-content-server/server/lib/routes/post-csp.js
```diff
@@ -9,7 +9,7 @@
 'use strict';
 const _ = require('lodash');
 const joi = require('joi');
-const logger = require('../logging/log')('');
+const logger = require('../logging/log')('server.csp');
```
LZoog: @jbuck I looked everywhere for a stray `.` to fix `Type` without avail but based on how we're invoking this in other route files and because `mozlogger` does a `Type: nameParts.join('.')` this seems right, but let me know if I'm missing something.

shane-tomlinson: Huh, my comment about removing `'server.csp'` doesn't seem to have made it through. 

With the `server.csp`, the logs show a `Type` of `server.csp.server.csp`. 

> {"Timestamp":1565779877155000000,"Logger":"fxa-content-server","Type":"server.csp.server.csp","Severity":6,"Pid":17239,"EnvVersion":"2.0","Fields":{"agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:70.0) Gecko/20100101 Firefox/70.0","blocked":"inline","column":1861,"line":1,"referrer":"","source":"moz-extension://675522c6-75b1-8247-a598-823fb86b1213/build/inject.js","violated":"script-src"}}

If you remove the entire `'server.csp'` from here, it shows as we'd expect, with a Type of `server.csp`:

> {"Timestamp":1565779979969000000,"Logger":"fxa-content-server","Type":"server.csp","Severity":6,"Pid":17313,"EnvVersion":"2.0","Fields":{"agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:70.0) Gecko/20100101 Firefox/70.0","blocked":"inline","column":1,"line":28,"referrer":"","source":"http://127.0.0.1:3030/signup","violated":"style-src"}}

LZoog: > Can you remove the `'server.csp'` argument when creating the logger, otherwise the type shows up as `server.csp.server.csp`

Thanks for checking on that Shane! I wasn't sure if I had that correct which is why I'd called it out to @jbuck. I wasn't sure how to test this since I wasn't seeing the full CSP report in the logs. Per our Slack messages, changing `logging.fmt` from "pretty" to "heka" enabled me to see the output and verify my new push is correct. I'll add this to my notes on what we could better explain on the front-page docs (along with different types of content server testing).

PR #2188, opened 2019-08-13
fix(css): fix #2125 - Remove horizontal scrollbars 

### LINKED ISSUE #2125: Header wrapper CSS creates horizontal scroll
I noticed that FxA settings was wiggling left-to-right and found the problem:
Screen Shot 2019-08-08 at 4 43 45 PM

By disabling the CSS, it is apparent that the header wrapper width is overflowing the page width. 
Screen Shot 2019-08-08 at 4 43 18 PM

LZoog:
fix for #2125 

I think what Ryan pointed out was actually fixed in #1969, but I did find at other breakpoints, `100vw` was still causing the scrollbar.

I also found that in the payments server, the breadcrumbs were causing a horizontal scrollbar due to use of `margin` instead of `padding` there. I also very minorly cleaned up some related CSS.

### DIFF
--- a/packages/fxa-content-server/app/styles/_layout.scss
+++ b/packages/fxa-content-server/app/styles/_layout.scss
@@ -49,7 +49,7 @@
     max-width: 640px;
     padding: 32px 0 0;
     transition: box-shadow 250ms cubic-bezier(0.07, 0.95, 0, 1);
-    width: 100vw;
+    width: 100%;
   }
 
   @include respond-to('small') {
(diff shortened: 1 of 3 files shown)


6a68:
@LZoog looks like this failed on a sass-lint error, otherwise it's good to land:

```
src/components/AppLayout/index.scss
  34:3  error  Expected `margin-top`, found `padding-left`  property-sort-order
  35:3  error  Expected `padding-left`, found `margin-top`  property-sort-order

LZoog:
@6a68 thanks for the callout, it looks like I didn't have my scss-lint setup and I hadn't yet realized it. Just installed `rvm` and all the works needed for the `scss-lint` package in VSC, made the fix, and pushed.

PR #2189, opened 2019-08-14
fix(tests): add missing coverage for the verification reminder script

### LINKED ISSUE #2184: Write tests for the verification reminders script
Ref: #2183

The library code for verification reminders has test coverage but the script itself doesn't. We need to add some to stop stuff like #2182 happening again in the future.

philbooth:
Fixes #2184.

Not the most intricate test I've ever written, but it would have caught the problem yesterday. Just runs the script and makes sure it doesn't fail. Not perfect, but better than what we have now, which is nothing.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/test/scripts/verification-reminders.js
+++ b/packages/fxa-auth-server/test/scripts/verification-reminders.js
@@ -1,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const ROOT_DIR = '../..';
+
+const cp = require('child_process');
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/test/scripts/verification-reminders.js
```diff
@@ -0,0 +1,21 @@
...
+
+'use strict';
+
+const ROOT_DIR = '../..';
+
+const cp = require('child_process');
+const Promise = require(`${ROOT_DIR}/lib/promise`);
+const path = require('path');
+
+cp.execAsync = Promise.promisify(cp.exec);
```
shane-tomlinson: Could you just use [execSync](https://nodejs.org/docs/latest-v10.x/api/child_process.html#child_process_child_process_execsync_command_options), depending on this behavior:

> If the process times out or has a non-zero exit code, this method will throw. The Error object will contain the entire result from child_process.spawnSync().

Issue #434, opened 2018-09-18
Enable the queues process in prod


philbooth:
Blocks mozilla/fxa-email-service#168. Depends on mozilla/fxa-email-service#83. Depends on mozilla/fxa-email-service#162. Depends on mozilla/fxa-auth-server#2633.

When we're ready, let's turn on the queues process in prod.

* [x] write docs
* [ ] deploy

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-629)
┆Issue Number: FXA-629

PR #2192, opened 2019-08-14
chore(email): remove auth server's bounce/delivery handlers

philbooth:
Related to #434.

We're turning on the email service's bounce/complaint/delivery handling, which means we can delete the equivalent code in the auth server once it's up and running.

There remains some queue logic to handle notifications from the email service, because the auth server is still responsible for emitting flow and amplitude events.

Opened as a draft for now because it's fine for the two implementations to run side-by-side for a while. Effectively the two will compete for events, but there won't be any duplicate bounce records created or whatever.


philbooth:
See mozilla-services/cloudops-deployment#3358 for docker-related deployment monkeying.

meandavejustice:
@shane-tomlinson Should we reassign this?

PR #2193, opened 2019-08-14
fix(signin): Prefill email signing into Sync after OAuth

### LINKED ISSUE #1949: Stop asking for email address if an FxA session exists
I think we should be able to reduce this bump in the road, to provide a more seamless experience when traversing FxA relying-services.

**STEPS TO REPRODUCE**
1. Install Notes extension and **Sign in to Sync** with a new account
1. Verify the account and confirm you have a session by visiting https://accounts.firefox.com
1. In FxA toolbar menu select **Turn on Sync…**

**EXPECTED RESULTS**
* Existing session is shown, user merely has to re-enter password (for services that require keys)

**ACTUAL RESULTS**
* User has to re-enter their email address
Screen Shot 2019-07-25 at 1 50 20 PM

shane-tomlinson:
When signing into Sync, instead of using *only* the browser's
view of the world and using a memory store, merge the browser's
view of the world with the accounts stored in localStorage.

If there is an account in localStorage with the same uid,
then merge the accounts.

This PR slightly changes the behavior of which account is
chosen as the "cached account". An account that is signed
into Sync and has an active session always takes precendence
over an OAuth signin. This was to simplify and unify logic
over how to choose the account.

fixes #1949

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/app-start.js
+++ b/packages/fxa-content-server/app/scripts/lib/app-start.js
@@ -354,7 +354,7 @@ Start.prototype = {
 
   initializeUser() {
     if (!this._user) {
-      const user = (this._user = new User({
+      this._user = new User({
         fxaClient: this._fxaClient,
         metrics: this._metrics,
         notifier: this._notifier,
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
Calling this ready for review, @mozilla/fxa-devs - r?

Issue #1832, opened 2019-07-17
(Updated) Mobile Screens


juliastarkov:
NOTE: These are not pixel perfect and should be used only as a reference point. They are not in Zeplin but if anyone needs pixel perfect Zeplin mobile screens don't hesitate to reach out and I will create them.

Please default to FxA styles where it makes more sense (i.e. button widths, padding, etc.)

Onboarding:
![Onboarding (Mobile)](https://user-images.githubusercontent.com/26334275/61419014-54fc2800-a8b1-11e9-8ae9-e49a8981a159.png)

Subscription Page:
![Subscriptions Page (Mobile)](https://user-images.githubusercontent.com/26334275/61419034-62b1ad80-a8b1-11e9-975b-c6b9ce094cb9.png)

Billing:
![Billing (Mobile - filled out)](https://user-images.githubusercontent.com/26334275/61419081-8674f380-a8b1-11e9-81a1-09f01f672d89.png)

Support:
![Support (Mobile - Filled out)](https://user-images.githubusercontent.com/26334275/61419083-8aa11100-a8b1-11e9-82f2-b3af4f78d225.png)

Subscription Termination:
![Termination (Mobile)](https://user-images.githubusercontent.com/26334275/61419094-9096f200-a8b1-11e9-8428-9b039a6e4803.png)

Notifications:
![Notification (Mobile)](https://user-images.githubusercontent.com/26334275/61419151-cd62e900-a8b1-11e9-9edb-a478c5337943.png)

clouserw:
Are there specific changes here?

juliastarkov:
@clouserw Sorry, can you clarify what you mean?

clouserw:
Thanks Julia.  I was just confused -- we talked in the ux/dev meeting.  

This is a bug to implement and/or verify the screens we have work on small screens.

meandavejustice:
Here are the screens I've addressed so far. I think all I have left is the support form on the content server. It was being hidden previously, and I'm seeing something weird with the inputs when mobile is detected. I should get to the bottom of that tomorrow. I will also need to test ipads etc..

![phase-1-mobile-subscription-cancel-expanded](https://user-images.githubusercontent.com/1844554/62915307-eef4aa80-bd61-11e9-981a-770c673378ee.png)
![phase-1-mobile-subscription-payment-update-form](https://user-images.githubusercontent.com/1844554/62915308-eef4aa80-bd61-11e9-868b-e16a5ec42aa8.png)
[...]

PR #2198, opened 2019-08-14
fix(support form): allow mobile browsers with chosen-js

chenba:
Chosen, the library used for the support form drop down, does not
support mobile browsers.  I forked Chosen, removed the mobile useragent
detection, built the project (it's written in coffeescript), and
committed the dist files into a branch
(https://github.com/chenba/chosen/commit/d3a62f4fd5f4550f2af0b6ac8cd5cb62dd1edcfa).

The support form does not rely on any of Chosen's styles and does not
use many of its features, so this approach should not be an issue.

Part of #1832

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/npm-shrinkwrap.json
+++ b/packages/fxa-content-server/npm-shrinkwrap.json
@@ -2892,9 +2892,8 @@
             }
         },
         "chosen-js": {
-            "version": "1.8.7",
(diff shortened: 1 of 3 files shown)


meandavejustice:
bizarre choice for chosenjs to not support mobile. Ideally I think we should use a different library, if we need to have our own third party version of the select box. We should create an issue to investigate this in the future.

PR #2201, opened 2019-08-14
feat: add json-schema/schemaver ADR

bbangert:
This is a proposal to mandate that JSON messages used in FxA have JSON-Schema's created in a defined location, where tooling will generate documentation for them.

JSON-Schema validators like [AJV](https://ajv.js.org/) should be used to validate messages before sending and validated by consumers in a similar manner.

### DIFF
--- a/docs/adr/0006-json-schemas-for-messaging.md
+++ b/docs/adr/0006-json-schemas-for-messaging.md
@@ -1,0 +1,99 @@
+# Utilizing JSON-Schemas, SemVer, and Tooling for JSON Messaging
+
+- Deciders: Ben Bangert, Jared Hirsch, Ryan Kelly, Shane Tomlinson
+- Date: 2019-08-14
+
+## Context and Problem Statement
+
+Firefox Accounts sends and receives messages in JSON utilizing queues. This ADR addresses this
+use-case only and does not apply to JSON used in RESTful APIs.
+
+Tracking and minimizing drift between the accuracy of message format documentation and
+the underlying implementation is an error-prone process. In addition, JSON messages are
+frequently produced and consumed from multiple systems, and in multiple langauges. Ensuring
+validity of these messages frequently requires implementing validation logic in each
+language that consumes these JSON messages to ensure correctness for message handling.
+
+Adding a field to a message format is non-trivial, as every consumer must also update
+its own validation logic. Currently there are a variety of places where messages are consumed
+and produced, most producers do not validate the message they send as the consumer code is not
+always in the same service where the validation code typically resides.
+
+There is currently no versioning for our message formats, which also creates difficulties in
+knowing when validation logic must change and when existing flexibility is sufficient.
+
+## Decision Drivers
+
+- Accurate documentation via JSON-Schema doc generation
+- Validation logic generated from JSON-Schema to reduce manually written/maintained code
+- [Semantic Versioning][schemaver] to effectively communicate breaking vs. non-breaking message format changes
+
+## Considered Options
+
+1. Keep existing approach
+2. Use protobuf to format/validate/document message formats
+3. Use JSON-Schemas with SemVer for doc/validation of existing JSON message formats
+
+## Decision Outcome
+
+Chosen option: "option 3", because our existing messages can be represented without any
+migration needed and external consumers already work with JSON.
+
+### [option 1] Keep existing approach
+
+Currently our documentation drifts if a message format is tweaked and the docs are not updated. Ideally
+our code review process would catch this but sometimes code subleties are not caught that alter the
+message format. Validation changes can also fail to appear in the documentation, even though what is
+considered a valid message key/value should be documented.
+
+We have to occasionally duplicate `joi` schemas between repos, which should likely be moved to `fxa-shared`
+to benefit re-use, but this does not help non-Node projects consume our messages.
+
+- Pros
+  - No additional work is needed that we don't already do.
+- Cons
+  - Our documentation frequently drifts.
+  - It's additional work to maintain accurate `joi` schemas spread throughout the code-base _or_
+  - It's additional work to consolidate our `joi` schemas in `fxa-shared`.
+  - `joi` schemas can't be used by consumers not written in JS.
+
+### [option 2] Use protobuf
+
+Using protobuf would allow us to validate and document valid message formats that would also consume
+less space and be more performant to serialize/deserialize. Validation is built-in as invalid messages
+will not deserialize, and documenation can be generated based on the protobuf spec for a message.
+
+Protobuf tooling is widely available for most langauges.
+
+- Pros
+  - High throughput messaging could benefit from higher efficiency of serializing/deserializing of messages
+  - Documentation stays in sync with message spec
+  - Validation is built-in to the message serialization/deserialization process
+- Cons
+  - Messages are in binary and not easily introspectable
+  - Existing message consumers will need to be updated to handle protobuf
+  - Migration procedure will need to be created/implemented to shift from existing JSON format
+
+### [option 3] Use JSON-Schemas
+
+JSON-Schemas can define existing FxA JSON messages used in our queue system. There are libraries in
+most languages to consume these schemas for validation purposes, and libraries exist to generate
+documenation from a JSON-Schema.
+
+[SchemaVer] can be introduced to existing FxA JSON messages as consumers ignore unknown keys, while
+consumers using the JSON-Schema can verify they can handle new message formats or if they're incompatible.
+Some of this logic is manual, as [SchemaVer] is not an official part of the JSON-Schema specification. There
+is a recommendation of [SchemaVer] in [Best Web Practices for Data](https://www.w3.org/TR/dwbp/#dataVersioning).
+
+- Pros
+  - No changes needed for existing consumers
+  - Validation can be done based on the JSON-Schema without custom code
+  - Documentation can be generated to stay in sync with message spec
+  - JSON is human readable without further processing.
+  - JSON is widely used on the web and our increasing use of JWTs also result in JSON.
+  - JSON-Schema is widely supported in a variety of langauges.
+- Cons
+  - [SchemaVer] is a draft that isn't officially used by the JSON-Schema committees, but no other versioning
+    scheme for JSON-Schema appear to exist as of this writing.
+
+[schemaver]: https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/


docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,92 @@
...
+documenation from a JSON-Schema.
+
+[SchemaVer] can be introduced to existing FxA JSON messages as consumers ignore unknown keys, while
+consumers using the JSON-Schema can verify they can handle new message formats or if they're incompatible.
+Some of this logic is manual, as [SchemaVer] is not an official part of the JSON-Schema specification.
+
+- Pros
+  - No changes needed for existing consumers
+  - Validation can be done based on the JSON-Schema without custom code
+  - Documentation can be generated to stay in sync with message spec
```
rfk: Do you have an example of documentation generated from a JSON-Schema? I don't find the raw schema definitions to be particularly for casual readers, which makes me wonder if we'd end up maintaining parallel "narrative" docs in addition to the schema.

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,92 @@
...
+### [option 2] Use protobuf
+
+Using protobuf would allow us to validate and document valid message formats that would also consume
+less space and be more performant to serialize/deserialize. Validation is built-in as invalid messages
+will not deserialize, and documenation can be generated based on the protobuf spec for a message.
+
+Protobuf tooling is widely available for most langauges.
+
+- Pros
+  - High throughput messaging could benefit from higher efficiency of serializing/deserializing of messages
```
rfk: Do we have places where this is currently a bottleneck, which might push us into biting the bullet here and using a more efficient approach?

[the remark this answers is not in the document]
bbangert: For the event broker, its not that important, our message throughput isn't terribly high. We're maybe a few per second vs. the hundreds or thousands per second that would make this a bigger priority.

docs/adr/0006-json-schemas-for-messaging.md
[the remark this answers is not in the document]
bbangert: Leaving out the second one, cause we're using Security Event Tokens rather than JSON for RPs..... except... the Security Event Token format is also JSON, so heh. :)

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,92 @@
...
+[SchemaVer] can be introduced to existing FxA JSON messages as consumers ignore unknown keys, while
+consumers using the JSON-Schema can verify they can handle new message formats or if they're incompatible.
+Some of this logic is manual, as [SchemaVer] is not an official part of the JSON-Schema specification.
+
+- Pros
+  - No changes needed for existing consumers
+  - Validation can be done based on the JSON-Schema without custom code
+  - Documentation can be generated to stay in sync with message spec
+- Cons
+  - [SchemaVer] is a new draft that isn't officially used by the JSON-Schema comittees
```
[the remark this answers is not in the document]
bbangert: Yes, I looked into this, and it does seem to be getting used in a [variety ](http://signalk.org/specification/1.0.0/doc/versioning.html) of [places](http://docs.mediachain.io/publishing/schema-recommendations#using-json-schema) or linked to from them.

SchemaVer is also referenced [as a best web practice for data versioning](https://www.w3.org/TR/dwbp/#dataVersioning).

docs/adr/0006-json-schemas-for-messaging.md
bbangert: Yes, [Adobe built one](https://github.com/adobe/jsonschema2md) that produces[ output like this](https://github.com/adobe/jsonschema2md/blob/master/examples/docs/example.schema.md).

docs/adr/0006-json-schemas-for-messaging.md
bbangert: Replied to the same thing for @6a68

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,92 @@
...
+always in the same service where the validation code typically resides.
+
+There is currently no versioning for our message formats, which also creates difficulties in
+knowing when validation logic must change and when existing flexibility is sufficient.
+
+## Decision Drivers
+
+- Accurate documentation via JSON-Schema doc generation
+- Validation logic generated from JSON-Schema to reduce manually written/maintained code
+- [Semantic Versioning][schemaver] to effectively communicate breaking vs. non-breaking message format changes
```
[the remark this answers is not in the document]
bbangert: I had considered that, but SchemaVer seems like the only real option for these. I included more references to it in a response to @rfk below.

docs/adr/0006-json-schemas-for-messaging.md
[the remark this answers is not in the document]
bbangert: I've added references to this in the doc (the best web practice and lack of any alternatives).

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,98 @@
...
+message format. Validation changes can also fail to appear in the documentation, even though what is
+considered a valid message key/value should be documented.
+
+We have to occasionally duplicate `joi` schemas between repos, which should likely be moved to `fxa-shared`
+to benefit re-use, but this does not help non-Node projects consume our messages.
+
+- Pros
+  - No additional work is needed that we don't already do.
+- Cons
+  - Our documentation occasionally drifts.
```
shane-tomlinson: s/occasionally/frequently/ 🤦‍♂

docs/adr/0006-json-schemas-for-messaging.md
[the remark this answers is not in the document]
shane-tomlinson: There are messages *to* the event broker, and there are messages *from* the event broker. It might be a bit much to ask of service providers to set up protobuf to consume message coming *from* the event broker.

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,98 @@
...
+Protobuf tooling is widely available for most langauges.
+
+- Pros
+  - High throughput messaging could benefit from higher efficiency of serializing/deserializing of messages
+  - Documentation stays in sync with message spec
+  - Validation is built-in to the message serialization/deserialization process
+- Cons
+  - Messages are in binary and not easily introspectable
+  - Existing message consumers will need to be updated to handle protobuf
+  - Migration procedure will need to be created/implemented to shift from existin JSON format
```
shane-tomlinson: typo: `existin` is missing its `g`

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Pros
+  - No changes needed for existing consumers
+  - Validation can be done based on the JSON-Schema without custom code
+  - Documentation can be generated to stay in sync with message spec
+  - JSON is human readable without further processing.
+  - JSON is widely used on the web and our increasing use of JWTs also result in JSON.
+  - JSON-Schema is widely supported in a variety of langauges.
+- Cons
+  - [SchemaVer] is a draft that isn't officially used by the JSON-Schema comittees, but no other versioning
```
shane-tomlinson: typo: `comittees` => `committees`

docs/adr/0006-json-schemas-for-messaging.md
shane-tomlinson: This was one of my concerns as well. I wonder if there is a way to import the schema doc into a more narrative doc.

I haven't played with JSONSchema enough to know, is it possible to generate full API docs like we have at https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/api.md?

docs/adr/0006-json-schemas-for-messaging.md
```diff
@@ -0,0 +1,98 @@
...
+Some of this logic is manual, as [SchemaVer] is not an official part of the JSON-Schema specification. There
+is a recommendation of [SchemaVer] in [Best Web Practices for Data](https://www.w3.org/TR/dwbp/#dataVersioning).
+
+- Pros
+  - No changes needed for existing consumers
+  - Validation can be done based on the JSON-Schema without custom code
+  - Documentation can be generated to stay in sync with message spec
+  - JSON is human readable without further processing.
+  - JSON is widely used on the web and our increasing use of JWTs also result in JSON.
+  - JSON-Schema is widely supported in a variety of langauges.
```
shane-tomlinson: Are there good libraries to consume JSON schema in both Hapi and Express? Without support, migrating to JSON Schema might just result in "yet another set of docs that are out of date". I see https://www.npmjs.com/package/express-jsonschema and https://github.com/simon-p-r/hapi-json-schema, https://github.com/bloglovin/timplan. I have no idea of their quality and at least the Hapi libraries do not seem to have been updated in quite a while.

docs/adr/0006-json-schemas-for-messaging.md
bbangert: The docs there are API docs, which is more like what Swagger would generate. For message formats we'd want to generate something closer to:
https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/service_notifications.md

Using the above output from the Adobe tool, we could generate a page for each message schema, then still have a narrative doc page that links to each generated message page.

docs/adr/0006-json-schemas-for-messaging.md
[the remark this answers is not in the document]
bbangert: Agreed, I definitely think JSON-Schema is the way to go.

docs/adr/0006-json-schemas-for-messaging.md
bbangert: This ADR is intended strictly for message formats, not API payloads (though perhaps we should represent the JWT payloads with JSON-Schema). For API's I think we should use gRPC and/or Swagger, which do this kind of stuff.

shane-tomlinson: OK, in my mind I broadened the scope of where this applies.

bbangert: I'll prolly come up with a Swagger/gRPC ADR next. 😉

docs/adr/0006-json-schemas-for-messaging.md
6a68: I've worked a bit with JSON Schema when working on WebExtension APIs. There are some really nice things about it. You can do things like declare a type and refer back to it using `$id`, and collect shared types or examples in a single spot, then point to them across schema files using `$ref`. Here are some relevant docs on managing large / complex schemas: https://json-schema.org/understanding-json-schema/structuring.html

> Do you have an example of documentation generated from a JSON-Schema?

Here is the schema I wrote for the Telemetry WebExtension API: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/schemas/telemetry.json

And here are the docs auto-generated from the schema: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/webextension-api.html

>  I haven't played with JSONSchema enough to know, is it possible to generate full API docs like we have at https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/api.md?

There's a JSON Hyper-Schema standard as well, which sounds like what you're after. Check out this blog post from Cloudflare, which goes into detail on how they auto-generate their API docs from Hyper-Schema: https://blog.cloudflare.com/cloudflares-json-powered-documentation-generator/

Also see the list of tools for Schema and Hyper-Schema: http://json-schema.org/implementations.html

docs/adr/0006-json-schemas-for-messaging.md
6a68: > Are there good libraries to consume JSON schema in both Hapi and Express?

I'm not sure about hapi or express connectors, but AJV is a JSON Schema validator that's widely used.

Moving from a JS-based schema (expressed in Joi) to a JSON Schema will reduce what we can do (JSON Schema isn't a full-fledged programming language), but I'm not sure how complex our Joi validation currently is. There are a couple of libraries that convert back and forth between Joi and JSON Schema, not sure about their quality, but if there's interest, we could play with auto-converting some of the stuff, then manually extracting out duplicated types using `$id` and `$ref`.

See also the list of tools here: http://json-schema.org/implementations.html

> For API's I think we should use gRPC and/or Swagger, which do this kind of stuff.

Swagger is a really good option to consider; I think the tooling is better, but I'm not sure how the tooling we'd actually use compares, and I'm not sure what else we'd get vs JSON (Hyper-)Schema. I imagine the security folks, and maybe QA, would be able to use more off-the-shelf testing tools if we opted for Swagger. Given that the FxA APIs are RESTful, gRPC and its protobuf transport seems like less of a clear fit. All that said, I think it'd be good to dig into the use cases carefully (in a separate ADR).

PR #2207, opened 2019-08-15
chore(support): Document support for Firefox 60+

shane-tomlinson:
Current ESR is based on 68.0. ESR - 1 is based on Firefox 60.

@mozilla/fxa-devs - r?


shane-tomlinson:
For reference, see the diagram here: https://www.mozilla.org/en-US/firefox/organizations/

PR #2214, opened 2019-08-15
refactor(tests): rename assertFailure to assert.failsAsync

philbooth:
Hey @vbudhram, I had a slightly different idea about that new `assert` module in your PR, so opening this to see what you think. Feel free to close if you don't like it.

The changes are:

* Moved it up to the `test/` directory so that it sits alongside `mocks` etc.

* Moved the new assertion function onto the `assert` object, so it can be invoked like other assertions as `assert.thing()`.

* Renamed it to `failsAsync` so that the async nature is clear in the name and the name is in the present tense like other assertions.

* Pass in expected error properties as an object rather than a callback with nested assertions. Just because I figure we'll always want to assert properties of the `err` object, and this is less typing and less indirection.

So usage changes from this:

```js
await assertFailure(runTest(route, request), err => {
  assert.equal(err.output.statusCode, 400);
  assert.equal(err.errno, 183, 'invalid or expired code errno');
});
```

To this:

```js
await assert.failsAsync(runTest(route, request), {
  errno: 183,
  'output.statusCode': 400,
});
```

Thoughts?


philbooth:
(oh, and the expected properties are optional, so it still fails if they're not passed in)

PR #2215, opened 2019-08-15
fix(avatar): Remove the WebRTC polyfill.

### LINKED ISSUE #2213: Remove the WebRTC adapter
The WebRTC adapter does some tests that cause some of our co-workers some pain. We added the WebRTC adapter to normalize WebRTC connections across various browsers to enable setting a profile using the camera.

Time has moved on, [our supported browser list has too](https://github.com/mozilla/fxa/blob/master/docs/supported-browsers.md), the browsers we officially support should all have decent WebRTC support.

We could query whether the browser has WebRTC support by checking the [methods we need are available](https://github.com/mozilla/fxa/blob/0c0c25e7220dc112ddbc5b8361b452e9ba91d353/packages/fxa-content-server/app/scripts/views/settings/avatar_camera.js#L70:L73), and if not, do not display the option to use the camera.

Ref https://github.com/mozilla/secure-proxy/issues/428

shane-tomlinson:
The WebRTC adapter does some tests that cause some of our co-workers some pain.
We added the WebRTC adapter to normalize WebRTC connections across various
browsers to enable setting a profile using the camera.

Time has moved on, our supported browser list has too, the browsers we
officially support should all have decent WebRTC support.

fixes #2213

### DIFF
--- a/packages/fxa-content-server/webpack.config.js
+++ b/packages/fxa-content-server/webpack.config.js
@@ -17,7 +17,6 @@ const webpackConfig = {
     appDependencies: [
       'lib/jquery',
       'backbone',
-      'canvasToBlob',
       'chosen-js',
       'cocktail-lib',
       'duration',
@@ -31,7 +30,6 @@ const webpackConfig = {
       'ua-parser-js',
       'uuid',
       'vat',
-      'webrtc',
       'styles/main.scss',
     ],
     head: './head/boot.js',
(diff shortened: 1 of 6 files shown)


packages/fxa-content-server/webpack.config.js
```diff
@@ -46,7 +45,7 @@ const webpackConfig = {
   },
 
   resolve: {
-    extensions: ['.ts', '.js','.jsx'],
+    extensions: ['.ts', '.js', '.jsx'],
```
vladikoff: extra space

shane-tomlinson: That was added automatically, looks like it was missing before.

vladikoff: woops, right! 👁

Issue #2019, opened 2019-07-31
Duplicate browser sessions appear in devices and apps list when site data is cleared


LZoog:
This may be a the same issue seen in https://github.com/mozilla/fxa/issues/1802, but is at least related.

When I clear site data and sign back in to my account, my new web session appears and my previous persists.

<img width="746" alt="image" src="https://user-images.githubusercontent.com/13018240/62226738-45381580-b380-11e9-8088-8f413c03f88f.png">

This also affects the new `delete_account` view:
<img width="734" alt="image" src="https://user-images.githubusercontent.com/13018240/62226813-63057a80-b380-11e9-904d-5a004b70b1b3.png">

rfk:
It's semi-related; "clear site data" just discards your `sessionToken` locally but doesn't do anything to delete it from the server, so the server has no way of knowing it has become inactive.

LZoog:
Do these inactive server-side session tokens eventually get purged?

rfk:
Yes, I believe web sessions expire after 4 weeks.

LZoog:
👍 So, shortening that wouldn't be the solution. What if we consolidated web session clients with the same name ("Firefox 70" and "Firefox 68" here) and displayed them with the most recent time accessed to kind of filter the list and get rid of the duplication on the client-side? Could that potentially hide anything that should be shown?

rfk:
> Could that potentially hide anything that should be shown?

Depends what you're trying to achieve I guess. Consolidating seems reasonable for the "delete account" view in my opinion, but I'm not well-versed in what that view is trying to achieve. For example, do we really need to show Web Sessions there at all?

LZoog:
Yeah, someone else had commented to me that it was odd, but UX did request the browser be shown. I offered hiding them as a solution for `delete_account` view yesterday but since it seemed to also affect devices & apps, I created this issue. The intent of `delete_account` is to just show an "account summary" of all active Mozilla products, including browsers, apps, and subscriptions.

cc: @johngruen @janiceahn for visibility on this discussion.

rfk:
> UX did request the browser be shown

Having absolutely zero context on these discussions, I wonder if what they want here are browsers in the "browsers connected to sync" sense, moreso than every place where you're signed in on the web.

> all active Mozilla products, including browsers, apps, and subscriptions.

[...]

PR #2216, opened 2019-08-15
fix(fxa-content-server): delete_account fix for #2019 - client-side filter to only render unique browsers

LZoog:
Partial fix for #2019, as this does not address duplication in "Devices & apps," should we open a separate issue for that? From the discussion in #2019, I'm not sure what the server-side solution would be or if that is something we would want to address.

image
image

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/delete_account.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/delete_account.js
@@ -33,27 +33,18 @@ var View = FormView.extend({
       });
     }
     this._activeSubscriptions = [];
+    this._uniqueBrowserNames = [];
     this._hasTwoColumnProductList = false;
     this._productListError = false;
   },
 
-  _formatTitle(items) {
-    return items.map(item => {
-      item.title = item.name;
-      if (item.clientType === CLIENT_TYPE_WEB_SESSION && item.userAgent) {
-        item.title = item.userAgent;
-      }
-      return item;
-    });
-  },
-
   setInitialContext(context) {
-    const clients = this._attachedClients.toJSON();
     context.set({
       email: this.getSignedInAccount().get('email'),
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/views/settings/delete_account.js
```diff
@@ -33,27 +33,18 @@ var View = FormView.extend({
...
 
-  _formatTitle(items) {
-    return items.map(item => {
-      item.title = item.name;
-      if (item.clientType === CLIENT_TYPE_WEB_SESSION && item.userAgent) {
-        item.title = item.userAgent;
-      }
-      return item;
-    });
-  },
```
LZoog: We no longer need to do this since I'm filtering web session titles in `_setuniqueBrowserSessions()` and I updated the template to just use `name` for `clients`.

LZoog:
@dannycoates agreed, that's a better name. Will make the update, thanks for the review!

janiceahn:
@LZoog This looks good to me, in that I'm seeing "Firefox browser" as umbrella term for any/all web browsers/sessions.

juliastarkov:
@LZoog Looks good to me

packages/fxa-content-server/app/scripts/views/settings/delete_account.js
```diff
@@ -101,10 +93,31 @@ var View = FormView.extend({
...
+      ...new Set(
+        this._attachedClients
+          .toJSON()
+          .filter(
+            client =>
+              client.clientType &&
+              client.clientType === CLIENT_TYPE_WEB_SESSION &&
+              client.userAgent
+          )
+          .map(({ userAgent }) => userAgent.replace(/[0-9]+/g, 'browser'))
```
l-hedgehog: Should this "browser" string be localizable?

PR #2218, opened 2019-08-16
chore(docs): add changelogs for subscription packages

### LINKED ISSUE #1963: Add CHANGELOG.md to fxa-event-broker, fxa-payments-server, fxa-support-panel
The release script updates a changelog file if it finds one. Do we want one for the subplat packages?

philbooth:
Fixes #1963.

Adds changelogs for the event broker, the payments server and the support panel. Also adds them for `fxa-shared` and `fxa-geodb`, although I was less certain whether that'd be useful. But it can't do any harm, and now they're no longer npm published I didn't see any reason to version them independently.

~~Additionally, `fxa-support-panel` had a different config to everything else and it was causing problems. Because `"parser": "typescript"` was explcitly specified in `.prettierrc`, it fell over on line 1, character 1 of the changelog. I've given it the same config as all the other packages, presumably that's fine.~~

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-event-broker/CHANGELOG.md
+++ b/packages/fxa-event-broker/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Change history
+
+## 1.143.4
+
+Prehistoric.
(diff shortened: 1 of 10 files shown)


philbooth:
Oh, this broke the build. Closing while I figure it out.

philbooth:
Fix for the build is in #2219, I'll rebase this after that lands.

philbooth:
Rebase was a resounding success, so r?

PR #2219, opened 2019-08-16
chore(package): give prettier the same config as other packages

philbooth:
I noticed we weren't running the prettier hook against `.ts` files, so added that. I also took the opportunity to run a repo wide scan in case any of the typescript needed a big ol' reformat, but it only picked up changes to `.json` files.

Includes the same `fxa-support-panel` prettier config change I added in #2218 because I needed it to run the scan, but I'll `git rebase --skip` after whichever lands first.

@mozilla/fxa-devs r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,10 @@
       "prettier --write",
       "git add"
     ],
+    "*.ts": [
+      "prettier --write",
+      "git add"
+    ],
     "*.css": [
       "prettier --write",
       "git add"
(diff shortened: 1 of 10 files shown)


philbooth:
This also broke the build (because it cherry-picked the same commit). Closing while I etc yada and so on.

philbooth:
Turns out I hadn't run `prettier` everywhere when I updated the config, so the linter was failing. This build should pass though. r?

PR #2223, opened 2019-08-16
feat(install): Check for node 10 on install

### LINKED ISSUE #1964: Idea: Check better node/npm version checking
#1599 is just one example of several where contributors have a difficult time running the FxA stack because of incompatible node/npm versions. Our [package.json](https://github.com/mozilla/fxa/blob/master/package.json#L41:L44) defines support as node >= 10 even though we *only* support node 10, and users of node 8 seem to either not be warned or ignore the warning. Can we do something in [postinstall.sh](https://github.com/mozilla/fxa/blob/master/_scripts/postinstall.sh) to ensure the correct node/npm versions?

shane-tomlinson:
If a version of node other than 10 is installed, error out.

fixes #1964

@mozilla/fxa-devs - r?

### DIFF
--- a/_scripts/check.sh
+++ b/_scripts/check.sh
@@ -2,6 +2,13 @@
 
 set -ex
 
+node_version="$(node -v | cut -f 1 -d '.' | cut -f 2 -d 'v')"
+if [ "$node_version" -ne "10" ]; then
+  echo "intall node 10 to continue installation"
+  echo "http://nodejs.org/"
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
> What was the thing you mentioned about node 12 in Slack? Are you going to land that soon?

Probably not today, the profile server tests are blowing up trying to upload an image.

PR #2224, opened 2019-08-16
fix(metrics): Emit an `fxa_login - complete` when signing in with cached credentials

### LINKED ISSUE #2186: Metrics for one-click signins / signins w/ cached credentials
Related to https://github.com/mozilla/fxa/issues/2009

Ideally we would

1. emit an amplitude `fxa_login - complete` event in these cases.
2. add a property that distinguishes these logins from cases that require a password. For example a boolean like `password_required` and make that an amplitude *event* property. 
3. (edit) ensure that the service the user is logging into is attached to these events.

cc @shane-tomlinson @davismtl

shane-tomlinson:
Emit the event whenever an `oauth.signin.signin.complete` or a `signin.signin.complete`
event is emit.

Looks like this:

> amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_login - complete","time":1565954783099,"user_id":"054d30e20521427c8dcf79378087153e","device_id":"88245ead2ab64fd7a84c5a22808c6c10","session_id":1565954773431,"app_version":"143","language":"es-ES","os_name":"Mac OS X","os_version":"10.14","event_properties":{"service":"321done","oauth_client_id":"325b4083e32fe8e7"},"user_properties":{"flow_id":"489130346c1cb8776d61a239c384c251100a936155b7fda5c0408b87bf861814","ua_browser":"Firefox","ua_version":"70.0","$append":{"fxa_services_used":"321done"}}}

Note, I'm doing this from the content server rather than the OAuth server out of expedience and nothing else. @rfk's suggestion of emitting the event whenever a code is traded for a token is valid, but requires much more work.

This does not take care of the `password_required` event property. I'll open a subsequent PR for that, possibly after returning from PTO.

issue #2186

### DIFF
--- a/packages/fxa-content-server/server/lib/amplitude.js
+++ b/packages/fxa-content-server/server/lib/amplitude.js
@@ -67,6 +67,10 @@ const EVENTS = {
     group: GROUPS.settings,
     event: 'logout',
   },
+  'signin.success': {
+    group: GROUPS.login,
+    event: 'complete',
+  },
 };
 
 const VIEW_ENGAGE_SUBMIT_EVENT_GROUPS = {
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -384,7 +384,10 @@ const conf = (module.exports = convict({
     format: String,
   },
   oauth_client_id_map: {
-    default: {},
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
@philbooth - mind doing a review of this?

Issue #440, opened 2018-08-11
Migrate away from using fxa-auth-db's database & schema


jbuck:
There's two parts to this:

1) From an operational POV, once we're only using fxa-email-service in production we can spin up a new database rather than reusing the fxa-auth-db's database.
2) It'd be good to add database migrations to this repo at that point in time

Requires https://github.com/mozilla/fxa-auth-server/issues/2631

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-635)
┆Issue Number: FXA-635

Issue #437, opened 2018-08-11
Running as a separate service


jbuck:
Depends on #166. Depends on #167. Depends on #168.

The service should be able to be run as a separate service from fxa-auth

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-632)
┆Issue Number: FXA-632

PR #2226, opened 2019-08-16
refactor(email): remove bounce checks from auth server

### LINKED ISSUE #2191: [auth server] remove old bounce-checking code
The auth server still contains code for checking bounce records before sending email. We do those checks in the email service too, so I think we can remove the auth code now. There's also a flag for it in config, so we could disable it softly before a hard delete.

Ref: `lib/bounces.js`, calling code is in `lib/senders/index.js`

philbooth:
Fixes #2191.

The auth server still checks bounce records before sending email, but the email service checks them too. That's bad because it means we're making unnecessary reads from the database. It's also a blocker for #437 and #440, because we can't switch to the email service's own db while the auth server still reads bounce records from the auth db. And it's semi-related to some of the simplifications I intend to make for #1779.

So this PR does two things:

1. It removes the old bounce-checking code from the auth server.

2. It fixes the email service so it replicates the old bounce-checking logic. We hadn't noticed the discrepancies there because the auth server checks were masking them.

@mozilla/fxa-devs r?


packages/fxa-email-service/src/api/send/mod.rs
```diff
@@ -77,29 +77,20 @@ impl FromDataSimple for Email {
 #[post("/send", format = "json", data = "<email>")]
 pub fn handler(
     email: AppResult<Email>,
-    bounces: State<DeliveryProblems<DbClient>>,
+    delivery_problems: State<DeliveryProblems<DbClient>>,
```
philbooth: Evidently this got missed when I did the big `bounce => delivery_problem` rename last year. Mopping it up now for consistency.

packages/fxa-email-service/src/api/send/mod.rs
```diff
@@ -77,29 +77,20 @@ impl FromDataSimple for Email {
...
-            bounces.check(&address)?;
-            refs.push(address.as_ref());
-        }
-        refs
-    } else {
-        Vec::new()
-    };
+    let empty_vec = Vec::new();
+    let (to, cc) =
+        delivery_problems.check_all(&email.to, email.cc.as_ref().unwrap_or(&empty_vec))?;
```
philbooth: `empty_vec` is a bit gross I know, but it can't be created inline in the function call because lifetime errors (`to` and `cc` are refs).

packages/fxa-email-service/src/db/delivery_problems/mod.rs
```diff
@@ -56,6 +56,44 @@ where
...
+                addresses.push(address.as_ref());
+            }
+            addresses
+        });
+
+        let to = if let Err(error) = self.check(&to) {
+            if cc.len() == 0 {
+                return Err(error);
+            } else {
+                cc.pop().unwrap()
```
philbooth: One tiny, tiny difference to the auth server's logic here, which I don't believe to be significant: I `pop` because it's more efficient than `remove`, but the auth server uses `shift`.

Just means a different secondary email gets used if there's more than one of them, which is fine. As George Orwell famously didn't say, "all secondary emails are equal, but some are more equal than others".

packages/fxa-email-service/src/db/delivery_problems/test.rs
```diff
@@ -132,14 +132,19 @@ fn check_soft_bounce() {
...
-    fn get_bounces(&self, _address: &EmailAddress) -> AppResult<Vec<LegacyDeliveryProblem>> {
+    fn get_bounces(&self, address: &EmailAddress) -> AppResult<Vec<LegacyDeliveryProblem>> {
         let now = now_as_milliseconds();
-        Ok(vec![LegacyDeliveryProblem {
-            address: "foo@example.com".parse().unwrap(),
-            problem_type: ProblemType::SoftBounce,
-            problem_subtype: ProblemSubtype::Undetermined,
-            created_at: now - DAY + SECOND * 2,
-        }])
+        let result = if address.as_ref() == "foo@example.com" {
```
philbooth: I'm leaning on these mocks below, so tweaked them to be more selective about the errors they return.

philbooth:
Of course, I broke the build didn't I. Closing while I sort that out.

philbooth:
Of course, I broke the build didn't I. Closing while I sort that out.

philbooth:
Build is fixed now, so r?

(not sure if anyone is around to review today, no rush if not!)

Issue #2229, opened 2019-08-16
Enable audit-filter for npm dependencies for all packages in monorepo


6a68:
Related to #2228 and inspired by the checklist item in #1128, let's make sure transitive dependencies are up-to-date across the whole monorepo. Copying in the checklist item from #1128 for guidance, and adding checkboxes for each package within the monorepo:

Packages with `npm audit --json` integrated into testing via a `lint:deps` npm task:

- [x] 123done
- [x] browserid-verifier
- [ ] fxa-auth-db-mysql
- [x] fxa-auth-server
- [x] fxa-basket-proxy
- [ ] fxa-content-server
- [x] fxa-customs-server
- [x] fxa-email-event-proxy
- [x] fxa-email-service
- [x] fxa-event-broker
- [x] fxa-geodb
- [x] fxa-js-client
- [x] fxa-payments-server (#1896)
- [x] fxa-profile-server
- [x] fxa-shared
- [x] fxa-support-panel (#741)

Security guidance for reference:

* [ ] enable security scanning of 3rd-party libraries and dependencies
  * For node.js, use [`npm audit`](https://docs.npmjs.com/cli/audit) with [audit-filter](https://github.com/mozilla-services/audit-filter) to review and handle exceptions (see example in [speech-proxy](https://github.com/mozilla/speech-proxy/pull/77/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2))
  * For Python, enable pyup security updates:
    * Add a pyup config to your repo (example config: https://github.com/mozilla-services/antenna/blob/master/.pyup.yml)
    * Enable branch protection for master and other development branches. Make sure the approved-mozilla-pyup-configuration team *CANNOT* push to those branches.
    * From the "add a team" dropdown for your repo /settings page
      * Add the "Approved Mozilla PyUp Configuration" team for your github org (e.g. for [mozilla](https://github.com/orgs/mozilla/teams/approved-mozilla-pyup-configuration) and [mozilla-services](https://github.com/orgs/mozilla-services/teams/approved-mozilla-pyup-configuration))
      * Grant it write permission so it can make pull requests
    * notify secops@mozilla.com to enable the integration in pyup

PR #2234, opened 2019-08-16
chore(deps): add npm audit to most packages in monorepo

6a68:
Refs #2229.

Encountering weird test failures with fxa-auth-db-mysql and fxa-content-server, so those packages aren't included here.

-----

Let's see if _this_ works. Had to force-update a few npm audit failures, and left a few unhandled. Probably those will fail, and I'll need to add .npmrc files.


philbooth:
> just use package-lock everywhere

I vote for this option.

PR #2241, opened 2019-08-16
Revert "chore(fxa-payments-server): fixes #1923 - add CSP to the payments server"

### LINKED ISSUE #1923: Add CSP to the payments-server

clouserw:
Reverts mozilla/fxa#2053  

This is breaking the payments flow and we need it running for QA


chenba:
Fixed what we really wanted to fix in #2243

PR #2247, opened 2019-08-19
fix(email): fix broken copy in download subscription email

philbooth:
Just noticed two mistakes I made in the download subscription email we added in #1791:

Screenshot of the broken download subscription email copy

* Somehow the product name got lost from the template data, so it said "If you haven't already downloaded undefined".

* There was a typo where it said "let's gets started" instead of "let's get started".

Fixes both, and the catching test is added that should have been there in the first place.

@mozilla/fxa-devs r?


philbooth:
I've also just pushed a secondary commit to turn off translations for these emails. It was pointed out in the recent l10n brouhaha that we didn't need to localise the download emails because launch will be limited to the USA. And this change is too late for the string cut too, so the translations wouldn't work in train 144 regardless.

PR #2260, opened 2019-08-20
refactor(db): rename productName to productId

### LINKED ISSUE #2252: [auth server] TypeError: Cannot read property 'refreshTokenId' of null
https://sentry.prod.mozaws.net/operations/auth-prod/issues/6156330/

https://github.com/mozilla/fxa/blob/046a4be333385eaa05f9b8600435d785fd15683c/packages/fxa-auth-server/lib/routes/devices-and-sessions.js#L81

philbooth:
Fixes #2252.

Fixes the Sentry error that came up yesterday, where `device` defaults to `null` in `buildDeviceResponse`. Defaulting it to an empty object instead fixes it, and added a regression test too.

@mozilla/fxa-devs, this is the last remaining issue before we can tag train 144, so if someone is free to review this today that would be splendid! 😄

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -340,7 +340,7 @@ function createServer(db) {
       db.createAccountSubscription(
         req.params.id,
         req.params.subscriptionId,
-        req.body.productName,
+        req.body.productId,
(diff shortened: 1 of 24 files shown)


philbooth:
Whoops, this is completely the wrong branch! One sec... :blush:

PR #2263, opened 2019-08-20
feat(codes): Add the ux for signup codes

vbudhram:
This is the final part for #457. It adds all the UX that is needed to support verification by signup code.

- [ ] Add feature
- [ ] Add experimentation logic (for sync only)
- [ ] Add supporting unit tests
- [ ] Add support functional tests
- [ ] Document where applicable


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -446,6 +447,10 @@ module.exports = (
             response.keyFetchToken = keyFetchToken.data;
           }
 
+          if (verificationMethod) {
+            response.verificationMethod = verificationMethod;
+          }
```
philbooth: Is there a test somewhere we should be updating to cover this?

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
```diff
@@ -0,0 +1,99 @@
...
+    ) {
+      return false;
+    }
+
+    const { featureFlags } = subject;
+
+    if (subject.clientId) {
+      let client = this.ROLLOUT_CLIENTS[subject.clientId];
+      if (featureFlags && featureFlags.signupCodeClients) {
+        client = featureFlags.signupCodeClients[subject.clientId];
```
philbooth: Great! If/when we want to tweak the rollout rates we can do that through Redis, you'll be the guinea pig!

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -771,6 +772,42 @@ const Account = Backbone.Model.extend(
...
+        this.unset('newsletters');
+        options.newsletters = newsletters;
+      }
+
+      return this._fxaClient
+        .sessionVerifyCode(this.get('sessionToken'), code, options)
+        .then(() => {
+          if (newsletters) {
+            this._notifier.trigger('flow.initialize');
+            this._notifier.trigger('flow.event', {
```
philbooth: This feels like the wrong place to emit these from, but then I checked the source and we already do the same in `verifySignUp`. So I guess it's fine?

Previously I think we tried to emit all the flow events from the view layer though, just because semantically that seems like where they should live. And if multiple views need to emit the same event, we use a mixin to eliminate duplicate code (and there's already a view mixin for `flow.initialize`). Is there some reason why that approach won't work here? (sorry I don't have all the context)

philbooth:
> Lgtm! :+1:

Although...

```
× firefox on linux 4.15.0-1043-aws - support form without valid session - go to support form, redirects to signin (20.774s)
    NoSuchElement: [POST http://localhost:4444/wd/hub/session/cdfa3f94-c35f-4eb3-8b62-9b6a49cf4812/element / {"using":"css selector","value":"#fxa-signin-header"}] Unable to locate element: #fxa-signin-header
    For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/no_such_element.html
    Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
    System info: host: '29221647750d', ip: '192.168.16.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.15.0-1043-aws', java.version: '11.0.4'
    Driver info: driver.version: unknown
      at Command.<anonymous> @ tests/functional/lib/helpers.js:191:22
      at Command.<anonymous> @ tests/functional/lib/helpers.js:61:10
      at Command.<anonymous> @ tests/functional/lib/helpers.js:75:18
      at Command.<anonymous> @ tests/functional/lib/helpers.js:69:10
      at Command.parent.get.setFindTimeout.then.then @ tests/functional/lib/helpers.js:1376:30
      at Command.<anonymous> @ tests/functional/lib/helpers.js:1374:8
      at Command.<anonymous> @ tests/functional/lib/helpers.js:61:10
      at Command.<anonymous> @ tests/functional/lib/helpers.js:1380:8
      at Command.<anonymous> @ tests/functional/lib/helpers.js:61:10
      at Command.<anonymous> @ tests/functional/lib/helpers.js:69:10
      at Test.go to support form, redirects to signin [as test] @ tests/functional/support.js:31:26
      @ src/lib/Test.ts:263:51
```

Issue #1857, opened 2019-07-19
[fxa-js-client] convert js-client unit tests to mocha


vladikoff:
We should probably convert the current AMD Intern unit tests to mocha to match the rest of the repos. 

(and later on even move the js-client directly into fxa-content-server, because it's not used anywhere else)

shane-tomlinson:
> We should probably convert the current AMD Intern unit tests to mocha to match the rest of the repos.

👍 Love it.

shane-tomlinson:
intern-geezer is now causing security alerts because of old dependencies. :(

PR #2268, opened 2019-08-20
fix(deps): Remove remaining amd loader parts

vbudhram:
Connects with https://github.com/mozilla/fxa/issues/1857.

Currently, js-client 1.0.18 fails when being loaded from node. I think this is the correct fix, thoughts @vladikoff ?


vladikoff:
@vbudhram Yesss!

PR #2271, opened 2019-08-20
chore(ts): convert fxa-shared/l10n/localizeTimestamp to typescript

dannycoates:


.circleci/test-content-server.sh
```diff
@@ -38,6 +38,12 @@ if grep -e "$MODULE" -e 'all' $DIR/../packages/test.list; then
 
   cd ../fxa-shared
   npm ci
+  npm run build
+
+  cd ../fxa-auth-server
+  node scripts/gen_keys.js
+  node scripts/gen_vapid_keys.js
+  node fxa-oauth-server/scripts/gen_keys.js
```
dannycoates: I'm puzzled why this became necessary

packages/fxa-shared/types/accept-language/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'accept-language';
```
dannycoates: `accept-language` v3.x is typescript. perhaps we should try upgrading

PR #2284, opened 2019-08-21
refactor(db): rename productName to productId

### LINKED ISSUE #1187: Rename `productName` to `productId` in `accountSubscriptions` table

philbooth:
Fixes #1187.

The `productName` thing has spread like wildfire. This PR includes changes to the auth db, auth server, event broker, payments server and support panel. It would have been a simpler change if we'd put a stop to it right away!

This is the first installment of a two-parter. We can't just run a migration that renames the column and adds new stored procedures, because deployed code running against the old procedures still expects to find the old column name. Instead we add a new column and new stored procedures in this migration, then in the next train we can remove the old junk in another migration. In the intervening period we read from both columns using `COALESCE`.

Because this PR creeps into some corners of the code I'm unfamiliar with, it might good if the responsible individuals for each particular area take a closer look at those bits. I'll give shoutouts in the inline comments. It's possible that I didn't need to change all of these places, but I was concerned about inconsistent nomenclature confusing us in the future.

Tests pass though so hopefully that's a good start (content server failures are unrelated).

@lmorchard, just fyi, this PR contains merge conflicts with your #2278, both the migration number and the versioning of the `createAccountSubscription` stored procedure.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/docs/service_notifications.md
+++ b/packages/fxa-auth-server/docs/service_notifications.md
@@ -216,5 +216,5 @@ so other services probably shouldn't use this.
 - `uid`: User id.
 - `subscriptionId`: Subscription id.
 - `isActive`: Boolean indicating whether the subscription is active.
-- `productName`: Product name.
+- `productId`: Product id.
 - `productCapabilities`: Array of product capabilities.
--- a/packages/fxa-auth-server/lib/subhub/updates.js
+++ b/packages/fxa-auth-server/lib/subhub/updates.js
@@ -90,7 +90,7 @@ module.exports = function(log, config) {
               eventCreatedAt: message.eventCreatedAt,
               subscriptionId: message.subscriptionId,
               isActive: message.active,
-              productName: message.productName,
+              productId: message.productName,
               productCapabilities:
                 config.subscriptions.productCapabilities[message.productName] ||
                 [],
--- a/packages/fxa-event-broker/lib/serviceNotifications.ts
+++ b/packages/fxa-event-broker/lib/serviceNotifications.ts
@@ -55,7 +55,10 @@ const SUBSCRIPTION_UPDATE_SCHEMA = joi
       .array()
       .items(joi.string())
       .required(),
-    productName: joi.string().required(),
+    productId: joi.string().optional(),
+    // TODO: productName is the legacy name for productId, remove it
+    //       in due course then make productId required again.
+    productName: joi.string().optional(),
     subscriptionId: joi.string().required(),
     uid: joi.string().required()
   })
--- a/packages/fxa-support-panel/lib/api.ts
+++ b/packages/fxa-support-panel/lib/api.ts
@@ -55,7 +55,7 @@ export interface DevicesResponse extends Array<Device> {}
 interface Subscription {
   uid: string;
   subscriptionId: string;
-  productName: string;
+  productId: string;
   createdAt: number;
 }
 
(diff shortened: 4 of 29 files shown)


packages/fxa-auth-server/docs/service_notifications.md
```diff
@@ -216,5 +216,5 @@ so other services probably shouldn't use this.
 - `uid`: User id.
 - `subscriptionId`: Subscription id.
 - `isActive`: Boolean indicating whether the subscription is active.
-- `productName`: Product name.
+- `productId`: Product id.
```
philbooth: @bbangert, note I decided to propagate this change through the service notification boundary and into the event broker too. My thinking was just for consistency, although of course I don't have a full picture of what's involved for downstream consumers so if you think that's a step too far, I'm happy to roll it back.

packages/fxa-auth-server/lib/subhub/updates.js
```diff
@@ -90,7 +90,7 @@ module.exports = function(log, config) {
               eventCreatedAt: message.eventCreatedAt,
               subscriptionId: message.subscriptionId,
               isActive: message.active,
-              productName: message.productName,
+              productId: message.productName,
```
philbooth: (...so this passes on to the event broker...)

packages/fxa-event-broker/lib/serviceNotifications.ts
```diff
@@ -55,7 +55,10 @@ const SUBSCRIPTION_UPDATE_SCHEMA = joi
       .array()
       .items(joi.string())
       .required(),
-    productName: joi.string().required(),
+    productId: joi.string().optional(),
+    // TODO: productName is the legacy name for productId, remove it
+    //       in due course then make productId required again.
+    productName: joi.string().optional(),
```
philbooth: (...and necessitates these knock-on changes)

@bbangert, I figured the correct first step in the event broker was to accept both variants as properties, then remove `productName` and make `productId` required later.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -456,7 +456,7 @@ const MOCK_ACTIVE_SUBSCRIPTIONS = [
   {
     uid: 'a90fef48240b49b2b6a33d333aee9b13',
     subscriptionId: 'sub0.28964929339372136',
-    productName: '123doneProProduct',
+    productId: '123doneProProduct',
```
philbooth: @lmorchard, in the payments server I just did a blanket `s/productName/productId/` and nothing seemed to fail, so...?

packages/fxa-support-panel/lib/api.ts
```diff
@@ -55,7 +55,7 @@ export interface DevicesResponse extends Array<Device> {}
 interface Subscription {
   uid: string;
   subscriptionId: string;
-  productName: string;
+  productId: string;
```
philbooth: One change in the support panel too, /cc @chenba.

philbooth:
> (content server failures are unrelated)

Ref: #2285

packages/fxa-event-broker/lib/serviceNotifications.ts
bbangert: At least one of these is necessary, I believe there's a joi OneOf thing.

bbangert: Hmm, guess the typesafe-joi doesn't work with `object.or` right now, so this is fine.

PR #2286, opened 2019-08-21
Async/await workshop PR

6a68:
## Promises and async/await: an asynchronous adventure

We're going to convert some auth-server routes from promises to async. The hapi route handlers are already async functions, so hopefully this should be pretty straightforward 🤞.

## Preliminaries

Bluebird (our promise library) is not compatible with async/await: https://github.com/petkaantonov/bluebird/issues/1434

However, the auth route handlers are already async functions, so we might not need to worry about this for today. At least, the tests passed when I made some changes 😬

## Converting Promise code to async/await code

The `await` keyword unwraps a Promise, converting a resolved promise into its value, and a rejected promise into an error.

The `async` keyword wraps a function in a Promise: return value becomes a resolved promise, thrown error becomes a rejected promise.

Remember: you can mix async and Promise code, because async functions return Promises. You don't have to convert a whole chain of functions at once.

### Some nice things about async/await

* Removes promise nesting due to conditional responses to intermediate values

* Allows sync and async calls to be handled by the same error handling code

* Escape from the 'pyramid of doom'

* TODO other nice stuff

### Some gotchas:

* Things get a little tricky with error handling `return foo()` (returns rejected promise) vs `return await foo()` (throws). Simpler to understand if you separate out assigning the awaited value and returning the value.

* If you want to parallelize, don't await serially. Instead, assign the promises and `await Promise.all(slowThings)`.

More at MDN: https://mdn.io/asyncawait

### A crude replacement algorithm

When you see a promise chain, turn the function outside the then/catch chain into an async function.

If you have a `.catch` callback, move the `.then` and `.catch` callbacks into a try-catch block.

Then, await the promise instead of chaining to get to the next step.

If a value is passed from promise to promise, assign its awaited value.

If the value of the whole promise chain is returned, `return foo().then(bar)`, split this into assignment and returning the awaited value: `const result = await foo().then(bar); return result`.

## Auth server routes: how deeply nested?

We have a few that are 8+ levels deep:

```
 ]$ echo "8 levels deep: " && ack "                .then"
8 levels deep: 
emails.js
687:                .then(() => {

recovery-key.js
159:                .then(() => db.accountRecord(email))
160:                .then(result => (uid = result.uid));

utils/signin.js
145:                .then(code => {
153:                .then(() => {
165:                    .then(() => {

sign.js
109:                .then(result => {

password.js
68:                .then(match => {
82:                .then(wrapKb => {
90:                    .then(keyFetchToken => {
95:                        .then(passwordChangeToken => {
251:                .then(() => {

account.js
1212:                .then(accountData => (account = accountData));
1238:                .then(result => (wrapKb = result));
1315:                  .then(result => {
```

## Let's walk through one together

I've started hacking on password.js to open this PR. We can make some more changes together.

## Do the thing

Grab a file and put your name next to it, let's get to hackin.

Open PRs against this `async-conversion` branch, not against master :+1:

- [ ] account.js
- [ ] attached-clients.js
- [ ] defaults.js
- [ ] devices-and-sessions.js
- [ ] emails.js
- [ ] foo.js
- [ ] idp.js
- [ ] index.js
- [ ] oauth.js
- [ ] password.js **@6a68 working on this one**
- [ ] recovery-codes.js
- [ ] recovery-key.js
- [ ] security-events.js
- [ ] session.js
- [ ] sign.js
- [ ] signin-codes.js
- [ ] sms.js
- [ ] subscriptions.js
- [ ] support.js
- [ ] token-codes.js
- [ ] totp.js
- [ ] unblock-codes.js
- [ ] util.js
- [ ] utils/clients.js
- [ ] utils/email.js
- [ ] utils/oauth.js
- [ ] utils/request_helper.js
- [ ] utils/signin.js
- [ ] utils/signup.js
- [ ] utils/subscriptions.js
- [ ] utils/totp.js
- [ ] validators.js


6a68:
Phil suggested we just land these refactorings as they are completed, since the changes are going to bit rot otherwise. I've copied the checklist into #2305 for future work.

I reviewed the changes other than password.js, could use a review for that one (and any others people have time to glance at).

packages/fxa-auth-server/lib/routes/password.js
```diff
@@ -169,92 +153,74 @@ module.exports = function(
...
-            });
+          const hex = await random.hex(32); // TODO why is this async?
+          authSalt = hex;
+          password = new Password(authPW, authSalt, verifierVersion);
+          await db.deletePasswordChangeToken(passwordChangeToken);
+          const hash = await password.verifyHash();
+          verifyHash = hash;
+          const wrapWrapKb = await password.wrap(wrapKb);
+          // Reset account, delete all sessions and tokens
+          const result = db.resetAccount(passwordChangeToken, {
```
bbangert: Needs await.

packages/fxa-auth-server/lib/routes/password.js
```diff
@@ -169,92 +153,74 @@ module.exports = function(
...
+            verifyHash: verifyHash,
+            authSalt: authSalt,
+            wrapWrapKb: wrapWrapKb,
+            verifierVersion: password.version,
+          });
+          request.emitMetricsEvent('account.changedPassword', {
```
bbangert: Await

packages/fxa-auth-server/lib/routes/password.js
```diff
@@ -53,71 +52,56 @@ module.exports = function(
...
+          );
+          const match = await signinUtils.checkPassword(
+            emailRecord,
+            password,
+            request.app.clientAddress
+          );
+          if (!match) {
+            throw error.incorrectPassword(emailRecord.email, form.email);
+          }
+          const password = new Password(
```
bbangert: password was already declared above as const.

packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -117,69 +101,52 @@ module.exports = (log, db, mailer, customs, config) => {
...
+            timeZone: geoData.timeZone,
+            uaBrowser: request.app.ua.browser,
+            uaBrowserVersion: request.app.ua.browserVersion,
+            uaOS: request.app.ua.os,
+            uaOSVersion: request.app.ua.osVersion,
+            uaDeviceType: request.app.ua.deviceType,
+            uid,
+          };
+
+          await mailer.sendPostRemoveTwoStepAuthNotification(
```
bbangert: Would be good to have @rfk verify this should be waited on here, as previously the promise was not returned to be waited on.

packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -201,42 +168,29 @@ module.exports = (log, db, mailer, customs, config) => {
...
-            );
+        try {
+          const token = await db.totpToken(sessionToken.uid);
+
+          // If the token is not verified, lets delete it and report that
+          // it doesn't exist. This will help prevent some edge
+          // cases where the user started creating a token but never completed.
+          if (!token.verified) {
+            await db.deleteTotpToken(sessionToken.uid);
+          }
```
bbangert: Lost the else statement here, exists should only be true if the token is verified.

packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -271,141 +225,110 @@ module.exports = (log, db, mailer, customs, config) => {
...
-        // If this is a new registration, replace and generate recovery codes
-        function replaceRecoveryCodes() {
-          if (isValidCode && !tokenVerified) {
-            return db
-              .replaceRecoveryCodes(uid, RECOVERY_CODE_COUNT)
-              .then(result => (recoveryCodes = result));
-          }
+        if (isValidCode) {
+          log.info('totp.verified', { uid });
+          request.emitMetricsEvent('totpToken.verified', { uid });
```
bbangert: Should we await on these or not? I think in another file they were waited on.

packages/fxa-auth-server/lib/routes/util.js
```diff
@@ -15,14 +15,13 @@ module.exports = (log, config, redirectDomain) => {
...
-          },
-          err => {
-            throw err;
-          }
-        );
+        let bytes;
+        try {
+          bytes = await random(32);
+          return { data: bytes.toString('hex') };
+        } catch (err) {
```
bbangert: If we don't catch this, won't it be thrown?

packages/fxa-auth-server/lib/routes/totp.js
6a68: Yes, seems like it is universally awaited elsewhere :+1:

packages/fxa-auth-server/lib/routes/totp.js
rfk: Hrm; to be honest I don't recall. Most of the places where we deliberately don't await a promise have a comment about why. I guess the only reason not to wait here would be to avoid a mail-sending error from "failing" the entire request, making it look like we failed to update the account when we really did. If we're worried about that, I think it would be better to explicitly await and then catch-and-ignore the error.

packages/fxa-auth-server/lib/routes/util.js
rfk: +1, can just remove the `try`-`catch` here.

PR #2306, opened 2019-08-22
chore(async): convert promises to async/await for auth server emails route

chenba:


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -201,95 +199,90 @@ module.exports = (
...
-            if (!sendEmail) {
-              return;
-            }
+        await customs.check(
+          request,
+          sessionToken.email,
+          'recoveryEmailResendCode'
+        );
+        await setVerifyCode();
+        setVerifyFunction();
```
bbangert: await?

bbangert: Ah, nevermind, I found setVerifyFunction which isn't async, odd that it was in a then before.

PR #2307, opened 2019-08-22
chore(deps): remove newrelic step one

dannycoates:


6a68:
rebased to pick up the security fix that just hit master

PR #2315, opened 2019-08-25
chore(js-client): simplify release process

### LINKED ISSUE #980: Simplify fxa-js-client release process
Running `grunt release` in the js client directory still tries to build and deploy the docs, which fails and is redundant anyway because we have the docs being built and deployed automagically in Circle now. We should take that stuff out.

mirefly:
Fixes #980


philbooth:
Fwiw, I think we have a plan to bundle this directly from the content server docker image instead of publishing it through npm, in which case all the bump-related stuff can go too. In the meantime we've just been bumping it manually for now.

So not sure if this is worth landing or not, let's see what everyone else thinks.

philbooth:
> So not sure if this is worth landing or not...

That came out wrong. It's definitely worth landing *something* here, but it may be that something is to remove all of the release/bump related stuff, instead of trying to make it work with the monorepo.

mirefly:
Thanks @philbooth . I updated it.

PR #2316, opened 2019-08-25
fix(content-server): advertise introspection and revocation endpoints

### LINKED ISSUE #1844: Advertize introspection endpoint in authorization server metadata
We currently advertise some endpoints via https://accounts.firefox.com/.well-known/openid-configuration, but not all.  Over in https://github.com/mozilla/application-services/pull/1429 we're adding some code that uses the "token introspection" endpoint, so it would be useful to advertise at least that one.

There's a spec for doing so in https://tools.ietf.org/html/rfc8414.

### LINKED ISSUE #2267: Advertize token-destroy endpoint for oauth clients
OAuth reliers aren't able to determine the token-destroy endpoint automatically from https://accounts.firefox.com/.well-known/openid-configuration. If we'd like them to destroy tokens on signout, we should tell them where to do so.

mirefly:
Fixes #2267 
Fixes #1844

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/get-openid-configuration.js
+++ b/packages/fxa-content-server/server/lib/routes/get-openid-configuration.js
@@ -7,16 +7,20 @@
 const config = require('../configuration');
 
 const authorizationEndpoint = config.get('public_url') + '/authorization';
+const introspectionEndpoint = config.get('oauth_url') + '/v1/introspect';
 const issuer = config.get('public_url');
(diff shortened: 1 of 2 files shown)


clouserw:
@mirefly are you still working on this?  Thanks!

mirefly:
@clouserw I am waiting for the reply https://github.com/mozilla/fxa/pull/2316#discussion_r317809323. @vladikoff have another look?

clouserw:
thank you!

Issue #963, opened 2019-04-29
Consider allowing get-metrics-flow to log other events besides form views, and to take a uid.


irrationalagent:
right now we allow reliers to hit our metrics endpoint for the purposes of logging a top-of-funnel `flow.enter-email.view` flow event (which gets translated to `fxa_email_first - view` for amplitude)

https://github.com/mozilla/fxa/blob/cf9ccbb2150f12b6173be8e787023ef0f0ddb236/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L31

we may want to expand this so that other events can be logged via this mechanism. examples:

1. a relier has a "sign up for fxa" button on their site, but no email form. if we want to track top-of-funnel for these cases, `flow.enter-email.view` isn't really the right event. however I believe we are logging this in the case of some singin buttons now e.g. send and monitor (we would also need to make a `button.view` event or similar.)

2. monitor also has the following use case. they want clicks on monitor emails (that are only sent to FxA subscribers) to count towards FxA MAU. one way to do this would be to (1) allow `get-metrics-flow` to accept FxA uids and (2) allow it to log an email click event. monitor would then append the user's FxA uid as a query parameter in their email link, and when the user arrives at the monitor site, ping our metrics endpoint with the uid attached. we then log an email click event. 

CC @sandysage @davismtl

vbudhram:
This is to help our reliers, but could be worked on post trailhead.

hoosteeno:
We just discovered that we are unable to accurately measure the top of funnel on www.mozilla.org Whatsnew pages, including Trailhead (67.0.1) and 68, because we haven't instrumented the Monitor sign up button that signed-in-to-sync users get on that page. 

![Screen Shot 2019-08-05 at 5 09 36 PM](https://user-images.githubusercontent.com/94519/62500465-d275d280-b7a3-11e9-96b2-778968d9b4b5.png)

We didn't instrument it in part because there isn't a form_type to associate with this button. It's not an email-first form; it's a different thing. 

On the Fx69 Whatsnew page, we'll show Monitor buttons to all users. If we can't measure the top of the funnel, we won't be able to do good analyses of the performance of that page in acquiring new FxA and activating existing FxA; we won't be able to differentiate the two. And that will carry over into 70. We should fix this.

In preparation for Fx69 and Skyline, we need a new form type to associate with Monitor buttons like the one in the attached screenshot. cc @davismtl @amychurchwell

davismtl:
[...]

PR #2318, opened 2019-08-26
feat(metrics): allow get-metrics-flow to take form_type button

irrationalagent:
addresses https://github.com/mozilla/fxa/issues/963

This patch should do 2 things (not sure it actually does, yet):

1. allows get-metrics-flow endpoint to accept a new form_type = button
2. when 1 happens, log new flow and amplitude events (fxa_button - view for amplitude and flow.button.view)

edit: so the meager tests i wrote seem to pass so i think it works? not sure how i would otherwise verify that the events are logged properly here, do i need to mock a fetch request to the endpoint somehow?


philbooth:
Thanks @irrationalagent!

PR #2322, opened 2019-08-27
fix(payments): update payment views for mobile

### LINKED ISSUE #1832: (Updated) Mobile Screens

meandavejustice:
- fixes #1832

This is first pass at mobile screens in payment flow/support form. @janiceahn Feel free to ping me for changes if you don't wanna post them all here

phase-1-mobile-subscription-cancel-expanded
phase-1-mobile-subscription-payment-update-form
phase-1-mobile-subscription-home
phase-1-mobile-screens
phase-1-mobile-support

cc/ @janiceahn for UX review

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useContext } from 'react';
 import {
   injectStripe,
   CardNumberElement,
@@ -21,12 +21,17 @@ import {
   useValidatorState,
 } from '../../lib/validator';
 import { formatCurrencyInCents } from '../../lib/formats';
+import { AppContext } from '../../lib/AppContext';
 
 import './index.scss';
 import { Plan } from '../../store/types';
 
+export const SMALL_DEVICE_RULE = "(max-width: 480px)";
+export const SMALL_DEVICE_LINE_HEIGHT = "37px";
+export const DEFAULT_LINE_HEIGHT = "45px";
+
 // ref: https://stripe.com/docs/stripe-js/reference#the-elements-object
-const STRIPE_ELEMENT_STYLES = {
+let stripeElementStyles = {
   base: {
     //TODO: Figure out what this really should be - I just copied it from computed styles because CSS can't apply through the iframe
     fontFamily:
@@ -36,6 +41,18 @@ const STRIPE_ELEMENT_STYLES = {
     lineHeight: '45px',
   },
 };
+
+type StripeElementStyles = {
+  fontFamily: string,
+  fontSize: string,
+  fontWeight: string,
+  lineHeight: string,
+};
+
+export function checkMedia(matched: boolean, stripeElementStyles: { base: StripeElementStyles }) {
+  let lh = matched? SMALL_DEVICE_LINE_HEIGHT : DEFAULT_LINE_HEIGHT;
+  return Object.assign(stripeElementStyles, {base: {lineHeight: lh}});
+}
 
 // Define a minimal type for what we use from the Stripe API, which makes
 // things easier to mock.
@@ -73,6 +90,8 @@ export const PaymentForm = ({
     middleware: validatorMiddlewareReducer,
   });
 
+  const { matchMedia } = useContext(AppContext);
+
   const onSubmit = useCallback(
     ev => {
       ev.preventDefault();
@@ -91,6 +110,8 @@ export const PaymentForm = ({
     },
     [validator, onPayment, onPaymentError, stripe]
   );
+
+  stripeElementStyles = checkMedia(matchMedia(SMALL_DEVICE_RULE), stripeElementStyles);
 
   return (
     <Form
@@ -122,7 +143,7 @@ export const PaymentForm = ({
           component={CardNumberElement}
           name="creditCardNumber"
           label="Card number"
-          style={STRIPE_ELEMENT_STYLES}
+          style={stripeElementStyles}
           className="input-row input-row--xl"
           required
         />
@@ -131,7 +152,7 @@ export const PaymentForm = ({
           component={CardExpiryElement}
           name="expDate"
           label="Exp. date"
-          style={STRIPE_ELEMENT_STYLES}
+          style={stripeElementStyles}
           required
         />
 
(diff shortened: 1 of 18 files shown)


meandavejustice:
contact-dropdown-1
contact-dropdown-2
update-billing
cancel-sub

juliastarkov:
@meandavejustice These look great, thanks for implementing the changes.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -21,12 +21,13 @@ import {
...
 } from '../../lib/validator';
 import { formatCurrencyInCents } from '../../lib/formats';
+import { AppContext } from '../../lib/AppContext';
 
 import './index.scss';
 import { Plan } from '../../store/types';
 
 // ref: https://stripe.com/docs/stripe-js/reference#the-elements-object
-const STRIPE_ELEMENT_STYLES = {
+let stripeElementStyles = {
```
lmorchard: I don't think this needs to be switched from const to let - the variable itself isn't reassigned, just properties of the object.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -37,6 +38,15 @@ const STRIPE_ELEMENT_STYLES = {
   },
 };
 
+/* istanbul ignore next */
```
lmorchard: This looks like something that *should* be covered in testing, not ignored

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -92,6 +104,9 @@ export const PaymentForm = ({
     [validator, onPayment, onPaymentError, stripe]
   );
 
+  /* istanbul ignore next */
```
lmorchard: This should get covered in testing too, if only to assert that it's called with this string. Might also be good to move this string into a constant at the top of the file.

packages/fxa-payments-server/src/lib/AppContext.tsx
```diff
@@ -20,6 +21,7 @@ export const defaultAppContext = {
   config,
   getScreenInfo: () => new ScreenInfo(),
   locationReload: noopFunction,
+  matchMedia: window.matchMedia || function() { return { matches: false }},
```
lmorchard: Don't refer to `window` here, just keep that isolated to `src/index.tsx` in the wrapper function. This default shouldn't get used in practice - both the app and tests should be providing a value, so `noopFunction` should work here.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -260,13 +264,15 @@ describe('routes/Product', () => {
...
       queryByText,
+      matchMedia,
       navigateToUrl,
       apiMocks,
     } = await commonSubmitSetup(createToken);
 
     fireEvent.click(getByTestId('submit'));
 
     await findByText('Your subscription is ready');
+    expect(matchMedia).toBeCalled();
```
lmorchard: This should probably be something like `expect(matchMedia).toBeCalledWith("(max-width: 480px)");` - or better yet, a constant imported from the component. 

Should also have a couple cases to mock `matchMedia` with both `true` & `false` results to exercise the `lineHeight` change in `checkMedia`

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -182,6 +182,18 @@ export const defaultAppContextValue = (): AppContextType => ({
   accessToken: 'at_12345',
   config,
   queryParams: {},
+  matchMedia: jest.fn().mockImplementation(query => {
+    return {
+      matches: false,
+      media: query,
```
lmorchard: Does this mock need to have all these properties? Looks like ` jest.fn(() => { return { matches: false } })` is what's used everywhere else

packages/fxa-payments-server/src/lib/AppContext.tsx
meandavejustice: noopFunction causes typescript to fail in this case :thinking:

packages/fxa-payments-server/src/lib/test-utils.tsx
meandavejustice: These are necessary to prevent failures in typescript checks

packages/fxa-payments-server/src/lib/AppContext.tsx
lmorchard: Hmm, also looks like `checkMedia` doesn't need the whole `MediaQueryList` object. Could just return `matches` from the wrapper in `src/index.tsx` and not have to cover the whole `MediaQueryList` object in types & mocks.

packages/fxa-payments-server/src/index.tsx
```diff
@@ -59,6 +60,10 @@ function locationReload() {
   window.location.reload();
 }
 
+function matchMedia(query: string) {
+  return window.matchMedia(query);
```
lmorchard: If this was `return window.matchMedia(query).matches` then the type of this function could just be `(query: string) => boolean` and not drag in all of `MediaQueryList`

packages/fxa-payments-server/src/lib/test-utils.tsx
lmorchard: Yeah, seems like that could be fixed by making the type simpler from the wrapper function in `src/index.tsx`

packages/fxa-payments-server/src/index.tsx
meandavejustice: that would greatly simplify things, thank you

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -37,6 +42,11 @@ const STRIPE_ELEMENT_STYLES = {
   },
 };
 
+export function checkMedia(matched: boolean, stripeElementStyles: object) {
```
lmorchard: Changes look good to me! :+1:

meandavejustice: Thanks!

PR #2324, opened 2019-08-27
refactor(api): prefer async/await in emails routes

chenba:
I'm really leaning on the tests to tell me that I haven't borked anything.  Preferably someone who's familiar with the emails routes will review this.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -66,9 +66,10 @@ module.exports = (
           });
         }
 
-        return cleanUpIfAccountInvalid().then(createResponse);
+        await cleanUpIfAccountInvalid();
+        return createResponse();
 
-        function cleanUpIfAccountInvalid() {
+        async function cleanUpIfAccountInvalid() {
           const now = new Date().getTime();
           const staleTime = now - config.emailStatusPollingTimeout;
 
@@ -88,20 +89,17 @@ module.exports = (
             // can never be verified, so the best we can do is
             // to delete them so the browser will stop polling.
             if (!validators.isValidEmailAddress(sessionToken.email)) {
-              return db.deleteAccount(sessionToken).then(() => {
-                log.info('accountDeleted.invalidEmailAddress', {
-                  ...sessionToken,
-                });
-                // Act as though we deleted the account asynchronously
-                // and caused the sessionToken to become invalid.
-                throw error.invalidToken(
-                  'This account was invalid and has been deleted'
-                );
+              await db.deleteAccount(sessionToken);
+              log.info('accountDeleted.invalidEmailAddress', {
+                ...sessionToken,
               });
+              // Act as though we deleted the account asynchronously
+              // and caused the sessionToken to become invalid.
+              throw error.invalidToken(
+                'This account was invalid and has been deleted'
+              );
             }
           }
-
-          return P.resolve();
         }
 
         function createResponse() {
@@ -186,7 +184,6 @@ module.exports = (
         let verifyFunction;
         let event;
         let emails = [];
-        let sendEmail = true;
 
         // Return immediately if this session or token is already verified. Only exception
         // is if the email param has been specified, which means that this is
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -196,100 +196,96 @@ module.exports = (
...
+          uaOSVersion: sessionToken.uaOSVersion,
+          uaDeviceType: sessionToken.uaDeviceType,
+          uid: sessionToken.uid,
+          style,
+        };
 
-              // This user is attempting to verify a secondary email that doesn't belong to the account.
-              if (emails.length === 0) {
-                throw error.cannotResendEmailCodeToUnownedEmail();
+        await verifyFunction(emails, sessionToken, mailerOpts);
```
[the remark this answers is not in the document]
chenba: `setVerifyFunction` (https://github.com/mozilla/fxa/pull/2324/files#diff-a8b1e870a711340cf536638ff5ec0730R291) and `verifyFunction(...)` (https://github.com/mozilla/fxa/pull/2324/files#diff-a8b1e870a711340cf536638ff5ec0730L235) are in the old code.  `setVerifyFunction` assigns a value to `event` from its outer scope in addition to `verifyFunction`.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -196,100 +196,96 @@ module.exports = (
...
-
-              // Check to see if this account has a verified TOTP token. If so, then it should
-              // not be allowed to bypass TOTP requirement by sending a sign-in confirmation email.
-              return db.totpToken(sessionToken.uid).then(
-                result => {
-                  if (result && result.verified && result.enabled) {
-                    sendEmail = false;
-                    return;
-                  }
-                  code = sessionToken.tokenVerificationId;
```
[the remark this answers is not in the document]
chenba: This line, line 269 above, and line 283 below are identical.  I kept the first assignment from line 269.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -196,100 +193,94 @@ module.exports = (
...
+          uid: sessionToken.uid,
+          style,
+        };
+
+        await verifyFunction(emails, sessionToken, mailerOpts);
+        await request.emitMetricsEvent(`email.${event}.resent`);
+        return emptyPromise;
+
+        // Returns a boolean to indicate whether to send email.
+        async function setVerifyCode() {
```
shane-tomlinson: Well done on incremental improvement, this stuff is dense. The myriad ways `code` and `emails` can be set is *really* hard to follow. I tried refactoring this function locally so that it returned the code/emails instead of depending on spooky action at a distance, and failed. I'm wondering whether your improvements can be further simplified, preventing us from shooting ourselves in the foot because of complexity.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -367,157 +358,143 @@ module.exports = (
...
+          let matchedEmail;
+          const emails = await db.accountEmails(uid);
+          const isEmailVerification = emails.some(email => {
+            if (email.emailCode && code === email.emailCode) {
+              matchedEmail = email;
+              log.info('account.verifyEmail.secondary.started', {
+                uid,
+                code,
               });
+              return true;
```
shane-tomlinson: This return value isn't actually used, is it needed?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -367,157 +358,143 @@ module.exports = (
...
+
+        // Any matching code verifies the account
+        await signupUtils.verifyAccount(request, account, request.payload);
+
+        return {};
+
+        async function verifySecondaryEmail(account) {
+          let matchedEmail;
+          const emails = await db.accountEmails(uid);
+          const isEmailVerification = emails.some(email => {
```
shane-tomlinson: What happens if someone programatically calls this endpoint with a primary email and specifies a `type` of `secondary`? Do we handle that correctly? And what about the reverse, they call the endpoint with a secondary email and specify a type of `primary`?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -586,108 +558,92 @@ module.exports = (
...
+            timeZone: geoData.timeZone,
+            uaBrowser: sessionToken.uaBrowser,
+            uaBrowserVersion: sessionToken.uaBrowserVersion,
+            uaOS: sessionToken.uaOS,
+            uaOSVersion: sessionToken.uaOSVersion,
+            uid,
           });
+        } catch (err) {
+          log.error('mailer.sendVerifySecondaryEmail', { err: err });
+          await db.deleteEmail(emailData.uid, emailData.normalizedEmail);
```
shane-tomlinson: What happens if db.deleteEmail fails here? 500 error?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -586,108 +558,92 @@ module.exports = (
...
+                  ...secondaryEmailRecord,
                 });
-            });
+                return;
+              } else {
+                throw error.unverifiedPrimaryEmailNewlyCreated();
+              }
+            }
+
+            // Only delete secondary email if it is unverified and does not belong
```
shane-tomlinson: Does this mean we remove a secondary email address from another account if it's not verified?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -712,59 +668,43 @@ module.exports = (
...
+            item.normalizedEmail === email.toLowerCase() && item.isVerified
+          );
+        });
+
+        // Don't bother sending a notification if removing an email that was never verified
+        if (!emailIsVerified) {
+          return {};
         }
+
+        // Notify only primary email and all *other* verified secondary emails about the
```
shane-tomlinson: This comment is a bit confusing, would it be more clear as `Notify any verified email address associated with the account of the deletion`

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -789,58 +729,41 @@ module.exports = (
...
-              return mailer.sendPostChangePrimaryEmail(
-                account.emails,
-                account,
-                {
-                  acceptLanguage: request.app.acceptLanguage,
-                  uid,
-                }
-              );
-            });
+        if (!secondaryEmail.isPrimary) {
```
shane-tomlinson: What happens if `secondaryEmail.isPrimary` is already set? Do you know how that'd happen? Should we only do any of the notification/mailing if `isPrimary` is false, otherwise we'd be notifying for an address that's already marked primary.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -196,100 +193,94 @@ module.exports = (
...
 
-              // This user is attempting to verify a secondary email that doesn't belong to the account.
-              if (emails.length === 0) {
-                throw error.cannotResendEmailCodeToUnownedEmail();
+              if (result && result.verified && result.enabled) {
+                return false;
               }
-
-              // Don't resend code for already verified emails
-              if (emailVerified) {
```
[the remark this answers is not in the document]
bbangert: The only place that has touched this variable is in a block that returns before this line can run, I stared at it for a few minutes as well before grokking it. I'm guessing it was leftover from some change before?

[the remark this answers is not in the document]
chenba: Not sure if was intentional, but it was certainly a mistake by me.  `emailVerified` was assigned as a side effect of building `emails`, which is either empty or has only one element.  I'll make some changes so that we'll return `foundEmail.isVerified` here.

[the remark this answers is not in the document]
chenba: Err I meant `!foundEmail.isVerified` since we don't want to send email if the address has been verified.

packages/fxa-auth-server/lib/routes/emails.js
chenba: That's the return value for the `emails.some()` callback.

packages/fxa-auth-server/lib/routes/emails.js
chenba: The short answer is I don't know.  This was the `then` handler from https://github.com/mozilla/fxa/pull/2324/files#diff-a8b1e870a711340cf536638ff5ec0730L381.  

For this patch I really did try to just collapse the promise chains and use async/await.  I can look into those scenarios, but I don't think those type of changes (if any needed) should go into this patch.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -196,100 +193,94 @@ module.exports = (
...
-
-            return verifyFunction(emails, sessionToken, mailerOpts).then(() =>
-              request.emitMetricsEvent(`email.${event}.resent`)
-            );
-          })
-          .then(() => {
-            return {};
-          });
+            return true;
+          } else if (sessionToken.tokenVerificationId) {
```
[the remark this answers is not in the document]
vbudhram: Created https://github.com/mozilla/fxa/issues/3127 to track this. Curently, no clear documentation on what this value is. You might be able to gather from poking around the code but we should make it easier.

packages/fxa-auth-server/lib/routes/emails.js
chenba: That appears to be the case.  Since the next line is throwing an error, it's also a 500 if `db.deleteEmail` succeeds.

packages/fxa-auth-server/lib/routes/emails.js
chenba: Whoa, it certainly looks that way.  😟

packages/fxa-auth-server/lib/routes/emails.js
chenba: Yes, good catch.  Thanks!

packages/fxa-auth-server/lib/routes/emails.js
chenba: (Note that this is in prod: https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/emails.js#L636)

chenba: There's a test for it: https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/test/remote/recovery_email_emails.js#L91-L144.  So I suppose that's correct then!

chenba:
@bbangert @shane-tomlinson I've made some changes based on your feedback: https://github.com/mozilla/fxa/pull/2324/commits/04f7801167c274da78461a2b689978ddd87d8f0f.  I kept it in a separate commit so it's easier to review.

packages/fxa-auth-server/lib/routes/emails.js
chenba: Ok, so the endpoint doesn't accept an email, and the only accepted value for type is 'secondary' (not in validation but in the `if`).  I think the question is: what happens if someone with a code for an account's primary email set `type` to 'secondary' in the payload.  In that case, the email will be verified, but the user will received a different post-verify email, and https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/utils/signup.js#L16 get bypassed.

PR #2333, opened 2019-08-28
feat(metrics): add subscription events and new top funnel event

### LINKED ISSUE #988: Add support for a new top of funnel event
Event Name = `fxa_subscribe - view`
Event Properties = `product_id`
User Properties = none

@irrationalagent  on Slack says "right now we log those types of events by having the relying party ping this endpoint https://github.com/mozilla/fxa/blob/627380f39ceda34c77cd9c5f65c1ab6f6051581f/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js and then we log an event on our side in response. but we only log one type of event currently, so we will need to add support for the fxa_subscribe - view event there"

See the [product requirements](https://docs.google.com/document/d/1d39yRiYt4rhahtMae04Pwcx5vldNQJeF1grz_KO8qJw/edit#)

bbangert:
Modeled on some bits of mozilla/fxa#2318 to add a new top level funnel event as well as the other events required for the subscription platform.

Closes #988


packages/fxa-content-server/server/lib/routes/get-metrics-flow.js
```diff
@@ -136,6 +141,25 @@ module.exports = function(config) {
         metricsData,
         req
       );
+    } else if (metricsData.form_type === FORM_TYPE_SUBSCRIBE) {
```
philbooth: 3 lookups against the same variable is probably the point at which I'd change this block to a `switch` or, even better, just have a map of `form_type` to `type` properties so that we can ditch the condition entirely and emit the events declaratively (if you see what I mean).

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -19,6 +19,12 @@ const GROUPS = {
   registration: 'fxa_reg',
   settings: 'fxa_pref',
   sms: 'fxa_sms',
+  sub: 'fxa_subscribe',
+  subCancel: 'fxa_subscribe_cancel',
+  subManage: 'fxa_subscribe_manage',
+  subPayManage: 'fxa_pay_manage',
+  subPaySetup: 'fxa_pay_setup',
+  subSupport: 'fxa_subscribe_support',
```
philbooth: What are all these? Don't seem to be used, unless I'm missing something...

packages/fxa-content-server/server/lib/routes/get-metrics-flow.js
```diff
@@ -98,31 +109,13 @@ module.exports = function(config) {
...
-          flowTime: flowBeginTime,
-          time: flowBeginTime,
-          type: ENTER_EMAIL_FLOW_EVENT_NAME,
-        },
-        metricsData,
-        req
-      );
-    } else if (metricsData.form_type === FORM_TYPE_BUTTON) {
+    const metricTypes = FORM_TYPES[metricsData.form_type];
+    if (metricTypes) {
```
philbooth: Exactly what I had in mind, thanks. :+1:

PR #2355, opened 2019-08-28
feat(support form): add subscription plan dropdown

### LINKED ISSUE #2279: Add dropdown to Support form
Please add a dropdown to the top of the support form. It will only have two options ("Secure Proxy" and "Other"). Set the default to Secure Proxy. 

_Purpose: (1) We will need this drop down when we have more than one subscription, so this will scale faster, and (2) viewers of the support demo noted the current UI doesn't mention Secure Proxy, this solves for that issue without adding unnecessary copy._ 

Support Form (Filled out)

chenba:
Fixes #2279 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/support.js
+++ b/packages/fxa-content-server/app/scripts/views/support.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import $ from 'jquery';
+import _ from 'underscore';
 import allowOnlyOneSubmit from './decorators/allow_only_one_submit';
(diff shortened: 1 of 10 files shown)


bbangert:
Rebase to master should fix the email-service failure.

packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -137,6 +159,11 @@ const SupportView = BaseView.extend({
     });
   },
 
+  navigateOnCancel(e) {
+    e.preventDefault();
```
chenba: The cancel button never worked correctly(?!) because we didn't do this.  A functional test is added in the PR for this also.

PR #2361, opened 2019-08-29
fix(codes): Updates to sign-up codes from user testing

vbudhram:


vbudhram:
Closing this down for now, I'll reopen when it is in a better state.

PR #2369, opened 2019-08-30
Will use the system path instead of the local path

Lourens-Rich:


vladikoff:
Why do you need this? 

We provide the bin from the local node_modules dir: https://github.com/mozilla/fxa/blob/master/_scripts/install_packages.sh#L39

PR #2357, opened 2019-08-29
Merge Train 144

vbudhram:

PR #2377, opened 2019-09-02
Merge train-144 into master

philbooth:
Replaces #2357; the other PR needed to pick up the audit fix made by @6a68 last week.

Merging from the `train-144-merge` branch because I already had it left over from #2328, with that fix. So I just merged `train-144` into that and hey presto.

@mozilla/fxa-devs r?


philbooth:
Note, we probably want to merge this before tagging train 145, if we don't want to get ourselves in a pickle with version string and changelog merge conflicts.

Issue #2305, opened 2019-08-22
[Meta] Async/await conversion of auth server routes and libs


jaredhirsch:
Copying over the list of routes from #2286. Add your name next to a checkbox item if you start refactoring it.

- [x] account.js **@chenba**, #3655
- [x] attached-clients.js (already using async/await)
- [x] defaults.js **@shane-tomlinson** 
- [x] devices-and-sessions.js **@philbooth**, #2301
- [x] emails.js **@chenba** https://github.com/mozilla/fxa/pull/2324
- [X] idp.js (made no promises)
- [X] index.js (made no promises)
- [x] oauth.js (already using async/await)
- [ ] password.js **@6a68 working on this one**
- [x] recovery-codes.js **@philbooth**, #2387
- [x] recovery-key.js **@philbooth**, #2386
- [x] security-events.js (already using async/await)
- [x] session.js **@philbooth**, #2384
- [x] sign.js @dannycoates
- [x] signin-codes.js **@philbooth**, #2383
- [x] sms.js **@philbooth**, #2382
- [x] subscriptions.js (already using async/await)
- [x] support.js (already using async/await)
- [x] token-codes.js **@philbooth**, #2381
- [x] totp.js
- [x] unblock-codes.js PR #2292
- [x] util.js
- [x] utils/clients.js (made no promises)
- [x] utils/email.js  (made no promises)
- [x] utils/oauth.js (already using async/await)
- [x] utils/request_helper.js  (made no promises)
- [x] utils/signin.js **shane-tomlinson**
- [x] utils/signup.js  (@bbangert)
- [x] utils/subscriptions.js (already using async/await)
- [x] utils/otp.js  (@bbangert)
- [x] validators.js  (made no promises)

Some non-routes libs that we can also investigate and convert:

(Note, the number at the end is the number of instances of `.then` in the file)

- [x] metrics/amplitude.js (3) (@shane-tomlinson)
- [x] metrics/events.js (8) (@shane-tomlinson)
- [x] metrics/context.js (3) (@bbangert)
- [x] crypto/random.js (3)
- [x] crypto/password.js (5)
- [x] scheme-refresh-token.js (5) (@bbangert)
- [x] customs.js (5) (@bbangert)
- [x] subhub/client.js (1)
- [x] features.js (1)
- [x] cache.js (4) (@bbangert)
- [x] server.js (6) (@bbangert)
- [x] db.js (51) (@bbangert)
- [x] pushbox.js (2)
[...]

PR #2382, opened 2019-09-03
refactor(api): prefer async/await in sms routes

philbooth:
Related to #2305.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/sms.js
+++ b/packages/fxa-auth-server/lib/routes/sms.js
@@ -49,79 +49,51 @@ module.exports = (log, db, config, customs, sms) => {
         request.validateMetricsContext();
 
         const sessionToken = request.auth.credentials;
-        const phoneNumber = request.payload.phoneNumber;
-        const templateName = TEMPLATE_NAMES.get(request.payload.messageId);
+        const { messageId, phoneNumber } = request.payload;
+        const templateName = TEMPLATE_NAMES.get(messageId);
         const acceptLanguage = request.app.acceptLanguage;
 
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/sms.js
```diff
@@ -49,79 +49,51 @@ module.exports = (log, db, config, customs, sms) => {
...
+          signinCode = await db.createSigninCode(
+            sessionToken.uid,
+            metricsContext.flow_id
           );
         }
 
-        function logSuccess() {
-          return request.emitMetricsEvent(`sms.${templateName}.sent`);
-        }
+        await sms.send(phoneNumber, templateName, acceptLanguage, signinCode);
```
shane-tomlinson: What happens if `request.app.features.has('signinCodes')` returns `false`, there won't be a signin code to send.

philbooth: We have two links in config, one to use when there is a `signinCode` and one when there isn't it. So the SMS gets sent in both cases:

https://github.com/mozilla/fxa/blob/c27c02379581873b02d61255123f36805ee0ec1f/packages/fxa-auth-server/lib/senders/sms.js#L159-L166

shane-tomlinson: Thanks for that info. Do we ever use that non-signinCode SMS anymore? Should a follow on be opened to remove it if not?

philbooth: Great question! I'll have to do some research to find out the answer though.

PR #2384, opened 2019-09-03
refactor(api): prefer async/await in session routes

philbooth:
Related to #2305.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/session.js
+++ b/packages/fxa-auth-server/lib/routes/session.js
@@ -55,37 +55,28 @@ module.exports = function(
       },
       handler: async function(request) {
         log.begin('Session.destroy', request);
+
         let sessionToken = request.auth.credentials;
-        const uid = request.auth.credentials.uid;
+        const { uid } = sessionToken;
 
-        return P.resolve()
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -325,81 +285,75 @@ module.exports = function(
...
-              origSessionToken.tokenVerificationCode.length / 2
-            );
-            newVerificationState.tokenVerificationCodeExpiresAt = 0;
-          }
-          return P.props(newVerificationState);
+        // Copy verification state of the token, but generate
+        // independent verification codes.
+        let newVerificationState = {};
+        if (origSessionToken.tokenVerificationId) {
+          newVerificationState.tokenVerificationId = random.hex(
```
bbangert: I assume these return promises? That's why there's `P.props` below? If so, why not await them here and avoid the `P.props` entirely?

philbooth: Yep. Just for efficiency I think, so they're awaited concurrently.

PR #2386, opened 2019-09-03
refactor(api): prefer async/await in recovery-key routes

philbooth:
Related to #2305.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/recovery-key.js
+++ b/packages/fxa-auth-server/lib/routes/recovery-key.js
@@ -27,57 +27,44 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
       handler: async function(request) {
         log.begin('createRecoveryKey', request);
 
-        const uid = request.auth.credentials.uid;
         const sessionToken = request.auth.credentials;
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/recovery-key.js
```diff
@@ -133,37 +112,32 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
...
+            throw errors.missingRequestParameter('email');
+          }
+
+          await customs.check(request, email, 'recoveryKeyExists');
+          ({ uid } = await db.accountRecord(email));
+        }
+
+        // When checking from `/settings` a sessionToken is required and the
+        // request is not rate limited.
+        return db.recoveryKeyExists(uid);
```
bbangert: Doesn't this need to be awaited?

shane-tomlinson: If `db.recoveryKeyExists` returns a promise, nope. Only need the await if we depend on the synchronous-like behavior it provides afterwards.

PR #2332, opened 2019-08-28
[HOLD] fix(experiments): Add 'de-DE' to newsletter safelist

6a68:
Some German language users were unable to access the email settings page with the 'de-DE' language enabled as their primary language in about:preferences. Previously, 'de' was allowed, but 'de-DE' was explicitly disallowed.

Fixes #2217.

-----

Notes for reviewer: 

I'm not sure of the backstory here, but the test I modified indicates that, originally, `de` was allowed, while `de-DE` was explicitly not allowed (along with any other region-specific language subtags like `de-AT`).

I'm not sure what the wider consequences will be of widening the list of allowed languages to include `de-DE`. This might affect experiment targeting. Hopefully someone with experience using the FxA experiment framework can review.

Note also that users with French language subtags like `fr-FR` set as the first language in their about:preferences list of languages will not be able to manage their email preferences, since `fr` but not `fr-FR` is in the allowlist; this presumably holds true for other languages where we only specify the two-character base language in the `communication-prefs.js` file. TBH I wonder if this issue has come up before, but can't find anything in the bug list, or in that file's commit history. 🤷‍♂ 

-----

Here's the preferences screen in a few languages, showing this change actually fixes #2217. Note the presence or absence of the 'email newsletter' link at the bottom of the screen:

American English (en-US):

Screen Shot 2019-08-27 at 4 49 43 PM

er, German German (de-DE):

Screen Shot 2019-08-27 at 4 49 55 PM

Austrian German (de-AT):

Screen Shot 2019-08-27 at 4 50 45 PM

PR #2394, opened 2019-09-03
fix(experiments): experiment bucketing should liberally match on region-free languages

### LINKED ISSUE #2217: Email preferences not showing up in German
It was reported to me by marketing that email preferences are not showing up in German.

English:
image

German:
image

I'm not quite sure why. I have not had the chance to verify this but the screenshots convinced me.

6a68:
Some German language users were unable to access the email settings page
with the 'de-DE' language enabled. Previously, 'de' was allowed, but
'de-DE' was explicitly disallowed. This commit changes the matching
behavior to work as follows:

* Loosely match languages specified without a region: if 'de' is
given, then 'de', 'de-DE', and 'de-AT' will all match.

* Strictly match languages specified with a region: if 'zh-tw'
is given, then 'zh' and 'zh-cn' will not match.

Fixes #2217, see also #2332 for context.

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
+++ b/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
@@ -10,22 +10,24 @@
 
 const BaseGroupingRule = require('./base');
 
-const AVAILABLE_LANGUAGES = [
-  'de',
-  'en',
-  'en-[a-z]{2}',
-  'es',
-  'es-[a-z]{2}',
-  'fr',
-  'hu',
-  'id',
-  'pl',
-  'pt-br',
-  'ru',
-  'zh-tw',
-];
+// If only the language, not the region, is specified, then all regions
+// will be considered a match. For instance, 'de' will match 'de', 'de-de',
+// 'de-at', etc. #2217
+const AVAILABLE_LANGUAGES = ['de', 'en', 'es', 'fr', 'hu', 'id', 'pl', 'ru'];
 
-const AVAILABLE_LANGUAGES_REGEX = arrayToRegex(AVAILABLE_LANGUAGES);
+// If the region is specified, other regions will not match. For instance,
+// 'pt-br' will not match 'pt' or 'pt-pt'. #2217
+const AVAILABLE_REGIONS = ['pt-br', 'zh-tw'];
+
+const AVAILABLE_LANGUAGES_REGEX = generateRegex(
+  AVAILABLE_LANGUAGES,
+  AVAILABLE_REGIONS
+);
+
+function generateRegex(langs, regions) {
+  const combined = langs.map(x => `${x}|${x}-[a-z]{2}`).concat(regions);
+  return arrayToRegex(combined);
+}
 
 function normalizeLanguage(lang) {
   return lang.toLowerCase().replace(/_/g, '-');
--- a/packages/fxa-content-server/app/tests/spec/lib/experiments/grouping-rules/communication-prefs.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/experiments/grouping-rules/communication-prefs.js
@@ -18,18 +18,25 @@ describe('lib/experiments/grouping-rules/communication-prefs', () => {
 
   [
     'de',
+    'de-AT',
+    'de-DE',
     'en',
+    'en-CA',
     'en-US',
     'en-GB',
     'es',
     'es-ES',
     'es-MX',
     'fr',
+    'fr-CA',
+    'fr-CH',
+    'fr-FR',
     'hu',
     'id',
     'pl',
-    'pt-br',
+    'pt-BR',
     'ru',
+    'ru-MO',
     'zh-TW',
   ].forEach(lang => {
     it(`choose returns true for ${lang}`, () => {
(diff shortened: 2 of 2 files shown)


packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
```diff
@@ -10,22 +10,24 @@
...
-  'id',
-  'pl',
-  'pt-br',
-  'ru',
-  'zh-tw',
-];
+// If only the language, not the region, is specified, then all regions
+// will be considered a match. For instance, 'de' will match 'de', 'de-de',
+// 'de-at', etc. #2217
+const AVAILABLE_LANGUAGES = ['de', 'en', 'es', 'fr', 'hu', 'id', 'pl', 'ru'];
```
6a68: Prettier made this formatting change, not me :-P

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
```diff
@@ -10,22 +10,24 @@
...
-  'de',
-  'en',
-  'en-[a-z]{2}',
-  'es',
-  'es-[a-z]{2}',
-  'fr',
-  'hu',
-  'id',
-  'pl',
-  'pt-br',
```
shane-tomlinson: This one may be problematic because Brazillian Portuguese is significantly different to Portuguese from Portugal.

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
```diff
@@ -10,22 +10,24 @@
...
-  'ru',
-  'zh-tw',
```
shane-tomlinson: Same here, assuming `zh` should apply to `zh-cn` is probably *not* the right thing to do as the `zh-tw` means Taiwan only, and `zh-cn` means China. The politics of sending newsletters tailored to Taiwan to Chinese users could be politically messy. Plus @l-hedgehog runs a separate Chinese stack where newsletters are specifically disabled. See https://github.com/mozilla/fxa/blob/7228646ae64fb7ddce9071e567c94b1041382bcd/packages/fxa-content-server/server/lib/configuration.js#L367 and https://github.com/mozilla/fxa/commit/5c42ab2b2a6012d60ff77e56ab7d69df52f2b7d4.

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
```diff
@@ -10,22 +10,24 @@
...
-];
+// If only the language, not the region, is specified, then all regions
+// will be considered a match. For instance, 'de' will match 'de', 'de-de',
+// 'de-at', etc. #2217
+const AVAILABLE_LANGUAGES = ['de', 'en', 'es', 'fr', 'hu', 'id', 'pl', 'ru'];
 
-const AVAILABLE_LANGUAGES_REGEX = arrayToRegex(AVAILABLE_LANGUAGES);
+// If the region is specified, other regions will not match. For instance,
+// 'pt-br' will not match 'pt' or 'pt-pt'. #2217
+const AVAILABLE_REGIONS = ['pt-br', 'zh-tw'];
```
6a68: @shane-tomlinson Note that this change strictly matches languages where the region is specified, like 'pt-br' and 'zh-tw'. These will not match on the base language or other regional languages. The tests have been updated to verify this works as intended.

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
6a68: Yup, note that this is the deleted line, not the new line.

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
6a68: Yup, the new code will preserve the distinction and strictly match on `zh-tw`.

packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/communication-prefs.js
```diff
@@ -10,22 +10,24 @@
...
+
+const AVAILABLE_LANGUAGES_REGEX = generateRegex(
+  AVAILABLE_LANGUAGES,
+  AVAILABLE_REGIONS
+);
+
+function generateRegex(langs, regions) {
+  const combined = langs.map(x => `${x}|${x}-[a-z]{2}`).concat(regions);
```
shane-tomlinson: 👍

6a68:
@davismtl Hey, so I think this version of the #2217 fix is what the marketing folks had in mind:

* for languages like `de`, also match on any specified subregion, like `de-DE` or `de-AT`
* for languages with a region like `zh-tw`, match only on that subregion

This change will enlarge the user group targeted by language-specific marketing experiments, though I don't know by how much.

Do you think we're good to land this change?

6a68:
@clouserw This is ready to land once we get signoff from Alex (hopefully won't bitrot while I'm out)

davismtl:
Based on the Slack conversations, you and Shane seem to have aligned with Lauren so I'll give my thumbs up based on that.

6a68:
Thanks @davismtl! We can always tweak this later if the change has unintended consequences for targeting :+1:

PR #2395, opened 2019-09-04
feat(event-broker): add config flag to allow capability check to fail

### LINKED ISSUE #2363: [fxa-event-broker] Gracefully run without access to subscriptions information
Right now if the auth-server doesn't provide subscriptions, the event broker refuses to start. This makes it difficult to deploy in production when the entire subscription platform is not deployed first. The event broker should be able to run without this information, it just won't propagate subscription state changes. This should be a flag to allow this run-state.

bbangert:
Closes #2363

### DIFF
--- a/packages/fxa-event-broker/config/index.ts
+++ b/packages/fxa-event-broker/config/index.ts
@@ -27,6 +27,12 @@ const conf = convict({
       doc: 'Refresh interval in seconds for fetching capability updates',
       env: 'CAPABILITY_REFRESH_INTERVAL',
       format: Number
+    },
+    requireCapabilities: {
+      default: true,
+      doc: 'Whether to fail to start if capabilties are not available from the auth server',
+      env: 'CAPABILITY_FAIL_ON_START',
+      format: Boolean
     }
   },
   clientWebhooks: {
--- a/packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
+++ b/packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
@@ -15,6 +15,7 @@ interface ServiceConfig {
   authToken: string;
   clientUrl: string;
   refreshInterval: number;
+  requireCapabilities?: boolean;
 }
 
 /**
@@ -35,7 +36,9 @@ class ClientCapabilityService extends SelfUpdatingService<ClientCapabilities> {
   private readonly clientUrl: string;
 
   constructor(logger: Logger, config: ServiceConfig) {
-    super(logger, config.refreshInterval * 1000, {});
+    const requireCapabilities =
+      config.requireCapabilities !== undefined ? config.requireCapabilities : true;
+    super(logger, config.refreshInterval * 1000, {}, requireCapabilities);
     this.authToken = config.authToken;
     this.clientUrl = config.clientUrl;
   }
--- a/packages/fxa-event-broker/lib/selfUpdatingService/index.ts
+++ b/packages/fxa-event-broker/lib/selfUpdatingService/index.ts
@@ -11,12 +11,19 @@ abstract class SelfUpdatingService<T> {
   private data: T;
   private timer: NodeJS.Timeout | null;
   private readonly refreshInterval: number;
+  private failOnStart: boolean;
 
-  constructor(logger: Logger, refreshInterval: number, defaultValue: T) {
+  constructor(
+    logger: Logger,
+    refreshInterval: number,
+    defaultValue: T,
+    failOnStart: boolean = true
+  ) {
     this.logger = logger;
     this.refreshInterval = refreshInterval;
     this.data = defaultValue;
     this.timer = null;
+    this.failOnStart = failOnStart;
   }
 
   /**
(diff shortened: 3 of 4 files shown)


packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
```diff
@@ -35,7 +36,9 @@ class ClientCapabilityService extends SelfUpdatingService<ClientCapabilities> {
   private readonly clientUrl: string;
 
   constructor(logger: Logger, config: ServiceConfig) {
-    super(logger, config.refreshInterval * 1000, {});
+    const requireCapabilities =
```
shane-tomlinson: Huh, I thought this would cause an error since super isn't called first.

packages/fxa-event-broker/lib/selfUpdatingService/index.ts
```diff
@@ -34,9 +41,12 @@ abstract class SelfUpdatingService<T> {
     const result = await this.updateFunction();
     if (isError(result)) {
       this.logger.error('clientCapabilityService', { err: result });
-      throw result;
+      if (this.failOnStart) {
```
shane-tomlinson: Should the error in the updateService interval propagate an error too?

packages/fxa-event-broker/test/lib/selfUpdatingService/clientCapabilityService.spec.ts
```diff
@@ -53,21 +54,54 @@ describe('Client Capability Service', () => {
...
+  it('fails on start if required to get capabilities', async () => {
+    nock('http://localhost')
+      .get('/v1/oauth/subscriptions/clients')
+      .reply(404, 'Not Found');
+    let error = null;
+    await service.start().catch(err => {
+      error = err;
+    });
+    cassert.isNotNull(error);
+    cassert.equal(((error as unknown) as Error).toString(), 'StatusCodeError: 404 - "Not Found"');
```
shane-tomlinson: What does `((error as unknown) as Error)` do?

packages/fxa-event-broker/config/index.ts
```diff
@@ -27,6 +27,12 @@ const conf = convict({
       doc: 'Refresh interval in seconds for fetching capability updates',
       env: 'CAPABILITY_REFRESH_INTERVAL',
       format: Number
+    },
+    requireCapabilities: {
+      default: true,
+      doc: 'Whether to fail to start if capabiltiies are not available from the auth server',
```
philbooth: Two small nits:

* Typo, "capabilities".

* I misread it the first couple of times as "whether to start if capabilities are not available..." , which is obviously the opposite of what's true. My eyes skipped over the all important "to fail" bit. Would it be clearer if it was just "whether to fail if capabilities are not available..." instead? (or possibly just poor reading skills on my part)

packages/fxa-event-broker/config/index.ts
```diff
@@ -27,6 +27,12 @@ const conf = convict({
...
+      env: 'CAPABILITY_START_ON_FAIL',
```
philbooth: And actually, I think the name of this env var may have contributed to my misreading too, it thought it meant the negation of what it actually is. Would "...FAIL_ON_START" be clearer than "...START_ON_FAIL"?

packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
```diff
@@ -35,7 +36,9 @@ class ClientCapabilityService extends SelfUpdatingService<ClientCapabilities> {
...
+      config.requireCapabilities !== undefined ? config.requireCapabilities : true;
```
philbooth: Would it ever actually be `undefined`? I thought the whole point of defaults in convict was that it takes care of giving you the default value in the fallback case, there's no need to implement that code yourself.

philbooth:
Whoops, started my review before I saw the other one come in, sorry!

packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
shane-tomlinson: I wondered the same to be honest, and then immediately forgot to comment on it.

packages/fxa-event-broker/config/index.ts
shane-tomlinson: Another one where I wondered the same.

packages/fxa-event-broker/lib/selfUpdatingService/index.ts
bbangert: That only logs in any case, so logging that it still can't fetch seems fine.

packages/fxa-event-broker/test/lib/selfUpdatingService/clientCapabilityService.spec.ts
bbangert: I couldn't cast it directly to Error since it was first infer'd as a null, so I had to cast to unknown first.

packages/fxa-event-broker/config/index.ts
bbangert: Ah, thanks! Fixed the typo and re-worded the env var.

packages/fxa-event-broker/lib/selfUpdatingService/clientCapabilityService.ts
bbangert: The code can't know the entire calling chain. It is possible to create this object with an undefined requireCapabilities per its ServiceConfig interface, so this has to be checked for type safety.

PR #2410, opened 2019-09-05
fix(oauth): Make refresh_token grants default to using original scopes.

### LINKED ISSUE #1867: scopes are not returned when using a refresh token to obtain an access token and not specified in the params
I noticed this a while back and forgot about it until [fixing this mistaken test](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/test/remote/oauth_tests.js#L196). If `scope` is not specified in the params, the [OAuth 2.0 spec](https://tools.ietf.org/html/rfc6749#section-6) says this should be treated as equal to the scope originally granted by the resource owner:

>  scope
         OPTIONAL.  The scope of the access request as described by
         Section 3.3.  The requested scope MUST NOT include any scope
         not originally granted by the resource owner, and if omitted is
         treated as equal to the scope originally granted by the
         resource owner.

### LINKED ISSUE #2408: Return original scopes by default in refresh_token grant
According to https://tools.ietf.org/html/rfc6749#section-6 when a client uses `grant_type=refresh_token` and does not provide the `scope` parameter, we should grant the same set of scopes as was originally requested:

```
   scope
         OPTIONAL.  The scope of the access request as described by
         Section 3.3.  The requested scope MUST NOT include any scope
         not originally granted by the resource owner, and if omitted is
         treated as equal to the scope originally granted by the
         resource owner.
```

We currently appear treat a missing scope as equivalent to the empty scope, which is not correct in this case.

rfk:
Previously, clients that did not provide an explicit `scope` parameter with their refresh_token request would generate an access_token bearing the empty scope. This changes it to the behaviour required by RFC6749 of defaulting to the originally-requested scopes.

Fixes #2408 
fixes #1867

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
@@ -50,7 +50,7 @@ module.exports = {
         .uri({
           scheme: ALLOWED_SCHEMES,
         }),
-      scope: validators.scope,
+      scope: validators.scope.required(),
       response_type: Joi.string()
         .valid(RESPONSE_TYPE_CODE, RESPONSE_TYPE_TOKEN)
         .default(RESPONSE_TYPE_CODE),
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/routes/key_data.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/routes/key_data.js
@@ -27,7 +27,7 @@ module.exports = {
     payload: {
       client_id: validators.clientId,
       assertion: validators.assertion.required(),
-      scope: validators.scope,
+      scope: validators.scope.required(),
     },
   },
   response: {
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
@@ -105,10 +105,16 @@ const PAYLOAD_SCHEMA = Joi.object({
     .default(MAX_TTL_S)
     .optional(),
 
-  scope: validators.scope.when('grant_type', {
-    is: Joi.string().valid(GRANT_REFRESH_TOKEN, GRANT_FXA_ASSERTION),
-    otherwise: Joi.forbidden(),
-  }),
+  scope: Joi.alternatives()
+    .when('grant_type', {
+      is: GRANT_REFRESH_TOKEN,
+      then: validators.scope.optional(),
+    })
+    .when('grant_type', {
+      is: GRANT_FXA_ASSERTION,
+      then: validators.scope.required(),
+      otherwise: Joi.forbidden(),
+    }),
 
   access_type: Joi.string()
     .valid(ACCESS_TYPE_OFFLINE, ACCESS_TYPE_ONLINE)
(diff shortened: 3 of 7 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
```diff
@@ -105,10 +106,16 @@ const PAYLOAD_SCHEMA = Joi.object({
...
-    otherwise: Joi.forbidden(),
-  }),
+  scope: Joi.alternatives()
+    .when('grant_type', {
+      is: GRANT_REFRESH_TOKEN,
+      then: validators.scope.optional(),
+    })
+    .when('grant_type', {
+      is: GRANT_FXA_ASSERTION,
+      then: validators.scope.default(ScopeSet.fromArray([])),
```
rfk: I think it would be reasonable to make `scope` required when doing an fxa-assertion grant, but didn't want to risk breaking in some edge-case where we're not doing that currently.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
```diff
@@ -50,7 +51,7 @@ module.exports = {
         .uri({
           scheme: ALLOWED_SCHEMES,
         }),
-      scope: validators.scope,
+      scope: validators.scope.default(ScopeSet.fromArray([])),
```
rfk: I think it would be reasonable to make `scope` required when granting an authorization code, but didn't want to risk breaking in some edge-case where we're not doing that currently.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/key_data.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
     payload: {
       client_id: validators.clientId,
       assertion: validators.assertion.required(),
-      scope: validators.scope,
+      scope: validators.scope.required(),
```
rfk: This is already marked `required` in the corresponding auth-server route so I figure it's safe to make it required here as well.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
shane-tomlinson: We require a scope of min length 1 on the front end, so my feeling is it should be safe to make required here too - see https://github.com/mozilla/fxa/blob/7228646ae64fb7ddce9071e567c94b1041382bcd/packages/fxa-content-server/app/scripts/models/reliers/oauth.js#L54:L56

packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
shane-tomlinson: Similar to the above, I think it should be fine to require scopes for assertion grants since we require a scope of min length 1 on the front end.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -2036,6 +2036,32 @@ describe('/v1', function() {
...
+                  client_secret: secret,
+                  grant_type: 'refresh_token',
+                  refresh_token: res.result.refresh_token,
+                },
+              });
+            })
+            .then(function(res) {
+              assert.equal(res.statusCode, 200);
+              assertSecurityHeaders(res);
+              assert.equal(res.result.scope, 'foo bar:baz');
```
shane-tomlinson: 👍

rfk:
I had to tweak a few tests to make `scope` required, but I think it all worked out OK.

PR #2419, opened 2019-09-05
Train 145 into master

vladikoff:


vladikoff:
File so we don't forget, might need the point release commits in here as well

packages/fxa-auth-server/CHANGELOG.md
```diff
@@ -1,34 +1,38 @@
...
+No changes.
+
 ## 1.145.0
 
 ### New features
 
-* support form: add subscription plan dropdown (bf89c002d)
-* codes: Add the ux for signup codes (37929e13a)
-* fortress: add "fortress" relying party (2ffe1334d)
+- support form: add subscription plan dropdown (bf89c002d)
```
shane-tomlinson: All these changes were automatic and are consistent with the rest of the changelog.

vladikoff:
@shane-tomlinson r+

bbangert:
@vladikoff looks like you forgot to sign a commit in this.

shane-tomlinson:
> @vladikoff looks like you forgot to sign a commit in this.

Aye, I didn't catch that since the original PR merged to train-145 instead of master.

Issue #2338, opened 2019-08-28
Create point release for signup codes experiment in Train 145


vbudhram:
From slack, the boomerang signup codes experiment will run on 40% of sync registrations. 20% for control group (uses links) and 20% for treatment group (uses codes). This will be a point release on train 145 so we can rollback if necessary.

cc @irrationalagent

clouserw:
@vbudhram what's the status of this?

vbudhram:
This has landed, closing!

PR #2424, opened 2019-09-05
chore(codes): Enable signup code sync experiment and enable for proxy

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/2338

Enrolls 40% of sync users (20% control, 20% treatment), and enables for all secure proxy clients.

cc @irrationalagent


vbudhram:
@mozilla/fxa-devs r?

PR #2426, opened 2019-09-05
refactor(api): prefer async/await in lib/metrics/events.js

shane-tomlinson:
issue #2305

@mozilla/fxa-devs - r?


packages/fxa-auth-server/lib/metrics/events.js
```diff
@@ -75,56 +87,41 @@ module.exports = (log, config) => {
...
-        return P.resolve();
+        return;
       }
 
       const request = this;
       let isFlowCompleteSignal = false;
 
-      return P.resolve()
-        .then(() => {
-          if (ACTIVITY_EVENTS.has(event)) {
```
dannycoates: i wonder why this was in a separate `then`? *doesn't matter*

shane-tomlinson: I wondered the same, maybe it could have thrown?

Issue #2367, opened 2019-08-29
Update support form handling to include product id


bbangert:
Due to consolidating our Zendesk installs, we need to include additional information on the tickets we create. Specifically we need to include the product id as a custom field. The Zendesk custom field id is `360022027772` that should be used for the product name.

clouserw:
We should use the product ID instead of the product name here

chenba:
I see that @bbangert changed the title from 'product id' to 'product name', but then @clouserw's comment asks for product id.  😕

chenba:
We allow users to select "Other" on the subscription drop down of the support form.  What should be the value for this new custom field when "Other" is selected?

bbangert:
This should be product id, not name. Sorry bout the confusion.

chenba:
Bumped the estimate up because the product id isn't part of the subscriptions data from the account model.

Issue #994, opened 2019-05-01
Add new events and properties for Subscription Support funnel


clouserw:
- [ ] Add `product_id, plan_id` event properties to a new `fxa_subscribe_support - view` event
- [ ] Add `product_id, plan_id` event properties to a new `fxa_subscribe_support - engage` event
- [ ] Add `product_id, plan_id` event properties to a new `fxa_subscribe_support - submit` event
- [ ] Add `product_id, plan_id` event properties to a new `fxa_subscribe_support - success` event
- [ ] Add `reason, product_id, plan_id` event properties to a new `fxa_subscribe_support - fail` event
- [ ] Add `product_id, plan_id` event properties to a new `fxa_subscribe_support - complete` event

See additional details in the [product requirements](https://docs.google.com/document/d/1d39yRiYt4rhahtMae04Pwcx5vldNQJeF1grz_KO8qJw/edit#)

See a [breakdown of events and properties](https://docs.google.com/spreadsheets/d/1ttgV60YeurrMhmdlCZ5WNBJXa-PcnRmrGgOEow71E1U/edit#gid=0)

PR #2429, opened 2019-09-05
chore(fxa-js-client): add method to get subscription plans

chenba:
Part of #2367 and #994 

@mozilla/fxa-devs r?


chenba:
Gah!  I missed a slash on the path in the request args.  😢

PR #2431, opened 2019-09-06
refactor(api): prefer async/await in lib/senders/email.js

shane-tomlinson:
issue #2305


shane-tomlinson:
Closing until I have time to re-visit.

PR #2433, opened 2019-09-06
feat(support form): include product name in support ticket

### LINKED ISSUE #2367: Update support form handling to include product id

chenba:
Fixes #2367

This PR replaces PR #2409 and move the product name look up to the client side as a side effect of needing the product id on the client side for metrics.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/support.js
+++ b/packages/fxa-content-server/app/scripts/views/support.js
@@ -63,7 +63,15 @@ const SupportView = BaseView.extend({
     return account.getSubscriptions().then(subscriptions => {
       if (subscriptions.length) {
         this.model.set('subscriptions', subscriptions);
-        return account.fetchProfile();
+
+        const getPlansPromise = account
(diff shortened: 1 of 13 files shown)


packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -112,8 +120,13 @@ const SupportView = BaseView.extend({
   onFormChange(e) {
     e.stopPropagation();
 
+    const plan = this.planEl.val();
+    const subhubPlan = this.findPlan(plan);
```
chenba: @bbangert The subhub plan with `product_id` and `plan_id` based on the user's selection.  It is `undefined` if the user selects "Other", however.

PR #2442, opened 2019-09-08
feat(oauth): support the session token scope for user agent clients

vladikoff:
Fixes https://github.com/mozilla/application-services/issues/1669
Requires https://github.com/mozilla/application-services/pull/1759
Ref: https://docs.google.com/document/d/1Q7xinEA3uXnUVobT5yaU7gXHcKK2Zj49nBrkUwHZVbg/edit?ts=5d5f0ffc#heading=h.bnyo94zc3qqz
Ref: https://github.com/mozilla-mobile/fenix/pull/5203

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/config/dev.json
+++ b/packages/fxa-auth-server/fxa-oauth-server/config/dev.json
@@ -134,7 +134,7 @@
       "redirectUri": "http://127.0.0.1:3030/oauth/success/3c49430b43dfba77",
       "canGrant": true,
       "trusted": true,
-      "allowedScopes": "https://identity.mozilla.com/apps/oldsync",
+      "allowedScopes": "https://identity.mozilla.com/apps/oldsync https://identity.mozilla.com/tokens/session",
       "publicClient": true
     },
     {
@@ -145,7 +145,7 @@
       "redirectUri": "http://127.0.0.1:3030/oauth/success/a2270f727f45f648",
       "canGrant": true,
       "trusted": true,
-      "allowedScopes": "https://identity.mozilla.com/apps/oldsync",
+      "allowedScopes": "https://identity.mozilla.com/apps/oldsync https://identity.mozilla.com/tokens/session",
       "publicClient": true
     },
     {
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
@@ -22,6 +22,7 @@ const ACR_VALUE_AAL2 = 'AAL2';
 const ACCESS_TYPE_OFFLINE = 'offline';
 
 const SCOPE_OPENID = ScopeSet.fromArray(['openid']);
+const { OAUTH_SCOPE_SESSION_TOKEN } = require('../../lib/constants');
 
 const ID_TOKEN_EXPIRATION = Math.floor(config.get('openid.ttl') / 1000);
 
@@ -77,24 +78,30 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
     }
   }
 
-  // For key-bearing scopes, is the client allowed to request them?
-  // We probably want to clean this logic up in the future, but for now,
-  // all trusted clients are allowed to request all non-key-bearing scopes.
+  // For custom scopes (besides the UNTRUSTED_CLIENT_ALLOWED_SCOPES list), is the client allowed to request them?
+  let requiresVerifiedToken = false;
   const scopeConfig = {};
-  const keyBearingScopes = ScopeSet.fromArray([]);
+  const customScopes = ScopeSet.fromArray([]);
   for (const scope of requestedGrant.scope.getScopeValues()) {
     const s = (scopeConfig[scope] = await db.getScope(scope));
-    if (s && s.hasScopedKeys) {
-      keyBearingScopes.add(scope);
+    if (s) {
+      if (s.hasScopedKeys) {
+        // scoped keys require verification, see comment below.
+        requiresVerifiedToken = true;
+      }
+      customScopes.add(scope);
     }
   }
-  if (!keyBearingScopes.isEmpty()) {
-    const invalidScopes = keyBearingScopes.difference(
+  if (!customScopes.isEmpty()) {
+    const invalidScopes = customScopes.difference(
       ScopeSet.fromString(client.allowedScopes || '')
     );
     if (!invalidScopes.isEmpty()) {
       throw AppError.invalidScopes(invalidScopes.getScopeValues());
     }
+  }
+
+  if (requiresVerifiedToken && !verifiedClaims['fxa-tokenVerified']) {
     // Any request for a key-bearing scope should be using a verified token,
     // so we can also double-check that here as a defense-in-depth measure.
     //
@@ -103,9 +110,7 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
     // verified by email before 2FA was enabled on the account. Such sessions must
     // be able to access sync even after 2FA is enabled, hence checking `verified`
     // rather than the `aal`-related properties here.
-    if (!verifiedClaims['fxa-tokenVerified']) {
-      throw AppError.invalidAssertion();
-    }
+    throw AppError.invalidAssertion();
   }
 
   // If we grow our per-client config, there are more things we could check here:
@@ -119,6 +124,7 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
     email: verifiedClaims['fxa-verifiedEmail'],
     scope: requestedGrant.scope,
     scopeConfig,
+    sessionTokenId: verifiedClaims['fxa-sessionTokenId'],
     offline: requestedGrant.access_type === ACCESS_TYPE_OFFLINE,
     authAt: verifiedClaims['fxa-lastAuthAt'],
     amr: verifiedClaims['fxa-amr'],
@@ -159,6 +165,11 @@ module.exports.generateTokens = async function generateTokens(grant) {
   // Maybe also generate an idToken?
   if (grant.scope && grant.scope.contains(SCOPE_OPENID)) {
     result.id_token = await generateIdToken(grant, result.access_token);
+  }
+
+  if (grant.scope && grant.scope.contains(OAUTH_SCOPE_SESSION_TOKEN)) {
+    result.session_token_id =
+      grant.sessionTokenId && grant.sessionTokenId.toString('hex');
   }
 
   amplitude('token.created', {
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
@@ -178,6 +178,8 @@ async function generateAuthorizationCode(client, payload, grant) {
     Object.assign(grant, {
       codeChallengeMethod: payload.code_challenge_method,
       codeChallenge: payload.code_challenge,
+      sessionTokenId:
+        grant.sessionTokenId && Buffer.from(grant.sessionTokenId, 'hex'),
     })
   );
   code = hex(code);
--- a/packages/fxa-auth-server/lib/constants.js
+++ b/packages/fxa-auth-server/lib/constants.js
@@ -2,8 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// The expected patch level of the database.
-// Update this if you add a new patch, and don't forget to update
-// the documentation for the current schema in ../schema.sql.
-
-module.exports.level = 28;
+module.exports = {
+  OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
+  OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
+};
--- a/packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-authorization-code.js
+++ b/packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-authorization-code.js
@@ -33,6 +33,7 @@ module.exports = config => {
         access_token: validators.accessToken.required(),
         refresh_token: validators.refreshToken.optional(),
         id_token: validators.assertion.optional(),
+        session_token_id: validators.sessionTokenId.optional(),
         scope: validators.scope.required(),
         token_type: Joi.string()
           .valid('bearer')
--- a/packages/fxa-auth-server/lib/routes/oauth.js
+++ b/packages/fxa-auth-server/lib/routes/oauth.js
@@ -16,9 +16,12 @@
  */
 
 const Joi = require('joi');
+const validators = require('../routes/validators');
 
 const error = require('../error');
 const oauthRouteUtils = require('./utils/oauth');
+const { OAUTH_SCOPE_SESSION_TOKEN } = require('../constants');
+const ScopeSet = require('../../../fxa-shared').oauth.scopes;
 
 module.exports = (log, config, oauthdb, db, mailer, devices) => {
   const OAUTH_DISABLE_NEW_CONNECTIONS_FOR_CLIENTS = new Set(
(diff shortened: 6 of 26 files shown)


packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -99,7 +101,19 @@ module.exports = (log, config) => {
...
+        const scopeSet = ScopeSet.fromString(codeResponse.scope);
+        if (scopeSet.contains(SESSION_TOKEN_OAUTH_SCOPE)) {
+          // at this point the auth code request has
+          // validated that the session scope is allowed for this client
+          codeResponse.session_token = await db.createSessionToken({
+            uid: codeResponse.uid,
+          });
+        }
+
+        return codeResponse;
```
vladikoff: `session_token` here is 

> {
  "data": "a5f118d0ecbbeaa2aaf45d1c5bb3d9234d50ada7aa8a93f1e1a84e8735bb4935",
  "id": "090d066d5efdb01c10e222b7abbcb7bc12d4769ba90ff4fa38aac8ffd210f89a",
  "authKey": "00a2ca8b8ec18be614873f61d5330724c284953672908cfb08b3bf9636215a0c",
  "bundleKey": "5f877ab616474f119db6be98502af6289b7ccda5f5a74852437849fab9d46977",
  "uid": "26f9b9198abd401c817e8b752eaaac4f",
  "lifetime": 2419200000,
  "createdAt": 1567967475605,
  "email": null,
  "emailCode": null,
  "emailVerified": false,
  "authAt": 0,
  "locale": null,
  "mustVerify": false,
  "tokenVerificationId": null,
  "tokenVerified": true,
  "tokenVerificationCode": null,
  "tokenVerificationCodeExpiresAt": null,
  "verificationMethod": null,
  "verifiedAt": null
}

atm

packages/fxa-auth-server/lib/constants.js
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = {
+  SESSION_TOKEN_OAUTH_SCOPE: 'https://identity.mozilla.com/tokens/session',
```
rfk: You could also consider lifting the "oldsync" scope into this file, it's currently repeated in a few places as a literal string.

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -93,6 +93,8 @@ module.exports.clientId = module.exports.hexString.length(16);
 module.exports.clientSecret = module.exports.hexString;
 module.exports.refreshToken = module.exports.hexString.length(64);
 module.exports.authorizationCode = module.exports.hexString.length(64);
+module.exports.uid = module.exports.hexString.length(32);
```
rfk: Wow, I'm surprised to see we didn't have a validator for this already.

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -99,7 +101,19 @@ module.exports = (log, config) => {
 
     async grantTokensFromAuthorizationCode(oauthParams) {
       try {
-        return await api.grantTokensFromAuthorizationCode(oauthParams);
+        const codeResponse = await api.grantTokensFromAuthorizationCode(
+          oauthParams
+        );
+        const scopeSet = ScopeSet.fromString(codeResponse.scope);
```
rfk: nit: I think this deserves a bit more of a comment about what it's doing and why, for context for future readers. Something about how sessionTokens live in the auth-server db so we have to create them here after the oauth-server has validated the request.

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -99,7 +101,19 @@ module.exports = (log, config) => {
...
+        if (scopeSet.contains(SESSION_TOKEN_OAUTH_SCOPE)) {
+          // at this point the auth code request has
+          // validated that the session scope is allowed for this client
+          codeResponse.session_token = await db.createSessionToken({
+            uid: codeResponse.uid,
+          });
```
rfk: This call to `db.createSessionToken` is, I think, the hardest part. The object you pass in here can have [a tremendous number of optional properties](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/tokens/session_token.js#L23) which affect the security of the system.

In general we need to make sure that the sessionToken created here is not any more powerful than the one that was used to create the authorization code. The [`/session/duplicate` route](https://github.com/mozilla/fxa/blob/dfc7a258402840e35139fd918c7bc93bc39ca9b7/packages/fxa-auth-server/lib/routes/session.js#L272) may be a good reference point for the sorts of things we need to consider here from a security perspective. In fact, "return a duplicate of the sessionToken that granted to code" is probably a good way to frame the requirements here.

Some thoughts:

* You're creating a *verified* sessionToken here. Is that appropriate? I think yes, but we'll need to ensure that only verified tokens can be used to create codes with `tokens/session` scope, and I don't think that's true right now.
* If we do create verified tokens, to what should we set the `verificationMethod` and `verifiedAt` properties? Should they be copied from the originating sessionToken?
* What should the `createdAt` and `authAt` be on this token? It's important that the last-auth-at time of the new token does not exceed the last-auth-at time of the token from which it was generated, otherwise an attacker could use this to get around security measures based checking whether the user recently entered their password.

Ultimately, I think you will need to somehow know the originating sessionToken in order to correctly create a new one.

(As a side note, the fact that `db.createSessionToken` lets you call the API without explicit answers to these questions seems like a bit of a API footgun, but that's a separate issue...)

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -60,6 +60,12 @@ const conf = convict({
     env: 'OAUTH_CLIENTS',
     default: [],
   },
+  sessionTokenClients: {
+    doc:
+      'List of OAuth clients that are allowed to use the https://identity.mozilla.com/tokens/session scope',
+    env: 'SESSION_TOKEN_OAUTH_CLIENTS',
+    default: [],
+  },
```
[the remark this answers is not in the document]
rfk: (From comments below, it looks like the existing "allowedScopes" mechanism only currently works for key-bearing scopes; it may be worth expanding that mechanism rather than adding a parallel one here).

rfk:
One other thing to consider: if the caller also requested `apps/oldsync` scope and a refresh token, then we will end up creating a placeholder device record for that refresh token. We should ensure that the sessionToken also gets attached to that same device record.

packages/123done/config-local.json
```diff
@@ -3,7 +3,7 @@
   "client_secret": "b93ef8a8f3e553a430d7e5b904c6132b2722633af9f03128029201d24a97f2a8",
   "redirect_uri": "http://127.0.0.1:8080/api/oauth",
   "issuer_uri": "http://127.0.0.1:3030",
-  "scopes": "profile openid",
```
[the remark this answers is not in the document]
vladikoff: Yeap good point!

vladikoff:
@rfk updated

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -99,7 +105,56 @@ module.exports = (log, config) => {
...
+        );
+        const scopeSet = ScopeSet.fromString(codeResponse.scope);
+
+        if (scopeSet.contains(OAUTH_SCOPE_SESSION_TOKEN)) {
+          const origSessionToken = await db.sessionToken(
+            codeResponse.session_token
+          );
+          // Copy verification state of the token, but generate
+          // independent verification codes.
+          let newVerificationState = {};
```
vladikoff: @rfk this is from the /duplicate route, we can possibly extract this into a helper

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -161,6 +166,11 @@ module.exports.generateTokens = async function generateTokens(grant) {
     result.id_token = await generateIdToken(grant, result.access_token);
   }
 
+  if (grant.scope && grant.scope.contains(OAUTH_SCOPE_SESSION_TOKEN)) {
+    result.session_token =
```
rfk: Nit: You're not returning a fully-functional sessionToken here, you're only returning the id. I suggest naming this field `session_token_id` or `session_id` to avoid confusion.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/authorization.js
```diff
@@ -51,6 +51,7 @@ module.exports = {
           scheme: ALLOWED_SCHEMES,
         }),
       scope: validators.scope.required(),
+      session_token: Joi.string().optional(),
```
rfk: Same here re: `session_token_id` naming.

packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-authorization-code.js
```diff
@@ -33,6 +33,7 @@ module.exports = config => {
         access_token: validators.accessToken.required(),
         refresh_token: validators.refreshToken.optional(),
         id_token: validators.assertion.optional(),
+        session_token: Joi.string().optional(),
```
rfk: IIUC, this is the point where you want to return the full session token rather than just its id, so I think changing the names of the others will help make it more obvious that something special is happening here.

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -19,9 +19,13 @@
...
 
 const createBackendServiceAPI = require('../backendService');
+const { OAUTH_SCOPE_SESSION_TOKEN } = require('../constants');
 const { mapOAuthError, makeAssertionJWT } = require('./utils');
+const ScopeSet = require('../../../fxa-shared').oauth.scopes;
+const P = require('../promise');
+const random = require('../crypto/random');
 
-module.exports = (log, config) => {
+module.exports = (log, config, db) => {
```
rfk: TBH I'm not super thrilled that we have to pass in a reference to `db` here, and I wonder if the graph of dependencies would be cleaner if we moved the "create a sessiontoken" logic into the route, in a similar place to where we currently do the "create a placeholder device revice" part.

But I don't feel strongly about it and would not block landing just for that.

packages/fxa-auth-server/lib/oauthdb/index.js
rfk: I wonder if a `duplicate` method on the `SessionToken` class would suffice for this?  I am so used to things in this codebase not being "proper" classes that I forget that our tokens do actually have that level of abstraction already, and are smart objects rather than just bags of properties.

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -99,7 +105,56 @@ module.exports = (log, config) => {
...
+            newVerificationState.tokenVerificationId = random.hex(
+              origSessionToken.tokenVerificationId.length / 2
+            );
+          }
+
+          if (origSessionToken.tokenVerificationCode) {
+            // Using expiresAt=0 here prevents the new token from being verified via email code.
+            // That's OK, because we don't send them a new email with the new verification code
+            // unless they explicitly ask us to resend it, and resend only handles email links
+            // rather than email codes.
```
rfk: Lol, yeah, I figure if you have to copy-paste an explanatory comment like this, it's probably a good sign that you should be extracting a shared helper function 😆

packages/fxa-auth-server/lib/routes/utils/oauth.js
```diff
@@ -48,6 +47,13 @@ module.exports = {
         .toString('hex');
     }
 
+    if (!credentials.id && grant.session_token) {
+      // if the caller also requested apps/oldsync scope and a refresh token, then
+      // we will end up creating a placeholder device record for that refresh token.
+      // We should ensure that the sessionToken also gets attached to that same device record.
+      credentials.id = grant.session_token;
```
rfk: Ugh, this code is correct, but it's also super non-obvious that the `id` field of this object should contain the sessionToken id. I guess this is a side-effect of passing around raw Hapi credential objects in order places. Maybe worth a comment for uninitiated readers?

packages/fxa-auth-server/fxa-oauth-server/config/dev.json
```diff
@@ -145,7 +145,7 @@
       "redirectUri": "http://127.0.0.1:3030/oauth/success/a2270f727f45f648",
       "canGrant": true,
       "trusted": true,
-      "allowedScopes": "https://identity.mozilla.com/apps/oldsync",
+      "allowedScopes": "https://identity.mozilla.com/apps/oldsync https://identity.mozilla.com/tokens/session",
```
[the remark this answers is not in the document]
vladikoff: it is not

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -77,24 +78,30 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
...
     const s = (scopeConfig[scope] = await db.getScope(scope));
-    if (s && s.hasScopedKeys) {
-      keyBearingScopes.add(scope);
+    if (s) {
+      if (s.hasScopedKeys) {
+        // scoped keys require verification, see comment below.
+        requiresVerifiedToken = true;
+      }
+      customScopes.add(scope);
     }
```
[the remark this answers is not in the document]
vladikoff: nope, we have scopes that have "keyBearing" set to 0

packages/fxa-auth-server/lib/oauthdb/create-authorization-code.js
```diff
@@ -18,6 +18,8 @@ module.exports = config => {
           .default('code'),
         client_id: validators.clientId.required(),
         assertion: validators.assertion.required(),
+        // attach a session token id to the generated code
+        session_token: Joi.string().optional(),
```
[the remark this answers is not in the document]
vladikoff: > session_token_id

👍 I had it like that before too :)

packages/fxa-auth-server/lib/oauthdb/grant-tokens-from-authorization-code.js
vladikoff: we should chat about this in the 1:1

packages/fxa-auth-server/lib/oauthdb/index.js
```diff
@@ -90,6 +90,8 @@ module.exports = (log, config) => {
 
     async createAuthorizationCode(sessionToken, oauthParams) {
       oauthParams.assertion = await makeAssertionJWT(config, sessionToken);
+      oauthParams.session_token = sessionToken.id;
```
vladikoff: this should be `id`

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -77,24 +78,30 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
...
   }
-  if (!keyBearingScopes.isEmpty()) {
-    const invalidScopes = keyBearingScopes.difference(
+  if (!customScopes.isEmpty()) {
```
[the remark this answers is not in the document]
vladikoff: Filed https://github.com/mozilla/fxa/issues/2478

vladikoff:
> The approach seems reasonable, I'd like to see this concept taken one step further.

Filed follow ups in https://github.com/mozilla/fxa/issues/2480 and https://github.com/mozilla/fxa/issues/2479

packages/fxa-auth-server/lib/routes/oauth.js
```diff
@@ -152,6 +160,42 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+        const scopeSet = ScopeSet.fromString(grant.scope);
+
+        if (scopeSet.contains(OAUTH_SCOPE_SESSION_TOKEN)) {
+          // the OAUTH_SCOPE_SESSION_TOKEN allows the server to create a new session token.
+          // the sessionTokens live in the auth-server db, we create them here after the oauth-server has validated the request.
+          let origSessionToken;
+          try {
+            origSessionToken = await db.sessionToken(grant.session_token_id);
+          } catch (e) {
+            throw error.invalidToken();
```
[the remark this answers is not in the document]
vladikoff: https://github.com/mozilla/fxa/issues/2499

packages/fxa-auth-server/lib/routes/oauth.js
```diff
@@ -152,6 +160,42 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
         }
 
+        const scopeSet = ScopeSet.fromString(grant.scope);
+
+        if (scopeSet.contains(OAUTH_SCOPE_SESSION_TOKEN)) {
+          // the OAUTH_SCOPE_SESSION_TOKEN allows the server to create a new session token.
+          // the sessionTokens live in the auth-server db, we create them here after the oauth-server has validated the request.
+          let origSessionToken;
+          try {
+            origSessionToken = await db.sessionToken(grant.session_token_id);
```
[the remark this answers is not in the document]
vladikoff: https://github.com/mozilla/fxa/issues/2500

PR #2445, opened 2019-09-09
feat(dataflow): handle new action messages from dataflow pipeline

### LINKED ISSUE #2116: [customs] Handle new action events from Dataflow fraud detection pipeline
Related to #2012.

Secops are going to start publishing actionable events for the customs server to consume, e.g. to block a given IP address. We should handle them but we should also put that handling behind config on a per-event basis, so we can turn things off if there are any teething problems.

philbooth:
Fixes #2116.

This is not a perfect implementation, but hopefully it's a reasonable first stab. It seeks to be a compromise between the vision set out in the [proposal](https://docs.google.com/document/d/12PhwR8k8cuesjhmgIjYrYTZwSAD-7M_oNecOZ5zuZSE) and the current reality of the customs server codebase. Essentially I've tried to implement everything in the least disruptive way possible, because I'm conscious that there's an overhaul planned at some point.

I'll add inline commentary to call out some of the more contentious details, but the headlines are:

* New properties for `suspect` and `disable` are added to the email and IP records.

* `disable` is stored in memcached like everything else, even though it's intended to be a long-term ban that's wholly unsuited to memcached's eviction policy. There's some discussion around this point in the proposal doc and a follow-up PR will be needed to implement the long-term resolution.

* Contrary to discussion in the proposal, `block` re-uses the existing limit from config instead of an independent one, just to keep the code a bit cleaner. It makes the email and IP records (and their callers) messier if we have two different types of blocks, with independent limits and methods for setting/checking.

* The existing dataflow module has pretty much been gutted because there's no relationship between the old message format and the new one. To that end, the diff is kind of unhelpful here and it might be easier to review the entire module instead of looking at the diff.

* A `reportOnly` property is added to config, initially defaulting to `true` so we can make sure we're comfortable with it in prod before flipping switches to make it actually do stuff.

Opened as a draft because I haven't done any end-to-end testing against real messages yet. And also because we don't want to land it until we're ready.

Still worth reviewing though, I think the code should be in a good state and there's unit test coverage in place.

@mozilla/fxa-devs r?

/cc @ameihm0912.

### DIFF
--- a/packages/fxa-customs-server/lib/config/config.js
+++ b/packages/fxa-customs-server/lib/config/config.js
@@ -43,10 +43,24 @@ module.exports = function(fs, path, url, convict) {
     },
     limits: {
       blockIntervalSeconds: {
-        doc: 'Duration of a manual ban',
+        doc: 'Duration of a manual `block` ban',
         default: 60 * 60 * 24,
         format: 'nat',
         env: 'BLOCK_INTERVAL_SECONDS',
+      },
+      suspectInterval: {
+        doc:
+          'Duration of a `suspect` flag, instigated by the Dataflow pipeline',
+        default: '1 day',
+        env: 'SUSPECT_INTERVAL',
+        format: 'duration',
+      },
+      disableInterval: {
+        doc:
+          'Duration of a long-term `disable` flag, instigated by the Dataflow pipeline',
+        default: '1 year',
+        env: 'DISABLE_INTERVAL',
+        format: 'duration',
       },
       rateLimitIntervalSeconds: {
         doc: 'Duration of automatic throttling',
@@ -352,6 +366,13 @@ module.exports = function(fs, path, url, convict) {
         format: Boolean,
         default: false,
         env: 'DATAFLOW_ENABLED',
+      },
+      reportOnly: {
+        doc:
+          'Treat all suggested actions from the Dataflow pipeline as `report`',
+        default: true,
+        env: 'DATAFLOW_REPORT_ONLY',
+        format: Boolean,
       },
       gcpPubSub: {
         projectId: {
--- a/packages/fxa-customs-server/lib/config/config.js
+++ b/packages/fxa-customs-server/lib/config/config.js
@@ -374,6 +374,13 @@ module.exports = function(fs, path, url, convict) {
         env: 'DATAFLOW_REPORT_ONLY',
         format: Boolean,
       },
+      ignoreOlderThan: {
+        doc:
+          'Ignore messages older than this value. Or set to `0` to never ignore old messages',
+        default: '1 day',
+        env: 'DATAFLOW_IGNORE_OLDER_THAN',
+        format: 'duration',
+      },
       gcpPubSub: {
         projectId: {
           doc: 'GCP PubSub project ID',
--- a/packages/fxa-customs-server/lib/dataflow.js
+++ b/packages/fxa-customs-server/lib/dataflow.js
@@ -26,7 +26,7 @@ module.exports = (config, log, fetchRecords, setRecords) => {
   }
 
   const { projectId, subscriptionName } = config.dataflow.gcpPubSub;
-  const { reportOnly } = config.dataflow;
+  const { ignoreOlderThan, reportOnly } = config.dataflow;
 
   if (!projectId) {
     throw new Error(
@@ -61,18 +61,12 @@ module.exports = (config, log, fetchRecords, setRecords) => {
       const action = parseMessage(message);
       const type = TYPES.get(action.indicator_type);
 
+      let level, op;
+
       if (reportOnly || action.suggested_action === 'report') {
-        log.warn({
-          op: 'dataflow.report',
-          [type]: action.indicator,
-          severity: action.severity,
-          confidence: action.confidence,
-          heuristic: action.heuristic,
-          heuristic_description: action.heuristic_description,
-          reason: action.reason,
-          suggested_action: action.suggested_action,
-        });
-      } else {
+        level = 'info';
+        op = 'dataflow.message.report';
+      } else if (isFresh(action.timestamp)) {
         const records = await fetchRecords({
           [type]: action.indicator,
         });
@@ -80,14 +74,26 @@ module.exports = (config, log, fetchRecords, setRecords) => {
         records[type][action.suggested_action]();
 
         await setRecords(records);
+
+        level = 'info';
+        op = 'dataflow.message.success';
+      } else {
+        level = 'warn';
+        op = 'dataflow.message.ignore';
       }
 
       message.ack();
 
-      log.info({
-        op: 'dataflow.message.success',
+      log[level]({
+        op,
         id: message.id,
+        timestamp: action.timestamp,
         [type]: action.indicator,
+        severity: action.severity,
+        confidence: action.confidence,
+        heuristic: action.heuristic,
+        heuristic_description: action.heuristic_description,
+        reason: action.reason,
         suggested_action: action.suggested_action,
         reportOnly,
       });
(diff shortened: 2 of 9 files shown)


packages/fxa-customs-server/lib/email_record.js
```diff
@@ -18,7 +18,9 @@ module.exports = function(limits, now) {
   EmailRecord.parse = function(object) {
     var rec = new EmailRecord();
     object = object || {};
-    rec.bk = object.bk; // timestamp when the account was banned
+    rec.bk = object.bk; // timestamp when the account was blocked
+    rec.su = object.su; // timestamp when the account was suspected
+    rec.di = object.di; // timestamp when the account was disabled
```
philbooth: Storing the `disable` flag in memcached is problematic. As specced, it's meant to be a permanent ban but for this first cut I'm just sticking it in memcached with everything else. The memcached eviction policy means we need to do something more permanent than that.

Ideally that means using a reliable data store, but I'm conscious there's been talk of getting rid of or re-implementing the customs server. So I wasn't sure we wanted to make the effort to add Redis or MySQL now, if we're going to replace the whole thing eventually anyway.

I did consider a workaround of emitting some kind of heartbeat of all the `disable` states to the logs at regular intervals, to allow those messages to be replayed if evictions are detected. But that's complicated by the fact our records are all keyed by email/IP address, meaning we'd have to iterate every single record to build up the heartbeat state. We could fix *that* by maintaining an inverted index just for the disables, but then we're back to the question of where to store the inverted index. If we put it in memcached it can be evicted, and if we add Redis or whatever then we may as well just skip the whole heartbeat idea and store the disables in Redis.

Suggestions welcome.

packages/fxa-customs-server/test/local/email_record_tests.js
```diff
@@ -22,7 +24,7 @@ test('shouldBlock works', function(t) {
   var er = simpleEmailRecord();
 
   t.equal(er.shouldBlock(), false, 'record has never been blocked');
-  er.rl = 499;
+  er.rl = 500;
```
philbooth: This was just to tighten up the test case a bit, rather than asserting either side of the boundary assert on the boundary itself.

packages/fxa-customs-server/lib/config/config.js
```diff
@@ -43,11 +43,25 @@ module.exports = function(fs, path, url, convict) {
     },
     limits: {
       blockIntervalSeconds: {
-        doc: 'Duration of a manual ban',
+        doc: 'Duration of a manual `block` ban',
         default: 60 * 60 * 24,
         format: 'nat',
         env: 'BLOCK_INTERVAL_SECONDS',
       },
+      suspectInterval: {
```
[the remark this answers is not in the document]
philbooth: In two minds about this suggestion. I can see the reasoning, but `blockIntervalSeconds` is a funny case where it is used by both the dataflow stuff and the non-dataflow, so it would seem wrong to nest it with the other two but also kind of wrong if they're nested together without it. So I'm leaning towards keeping them at this level, if that's cool.

philbooth:
Based on discussion with @ameihm0912 yesterday, I've pushed a second commit with an extra bit of config, `ignoreOlderThan`. Default value is 1 day and hopefully it honours the principle of least astonishment; it's there to prevent stale messages from being acted on if they show up in the queue.

packages/fxa-customs-server/lib/email_record.js
shane-tomlinson: If `su` or `di` are not available, are there any downstream repercussions? I see you check whether `this.su` or `this.di` have values below before comparing against them so I assume all is fine?

packages/fxa-customs-server/test/local/ip_record_tests.js
```diff
@@ -30,15 +32,45 @@ test('shouldBlock works', function(t) {
...
+  t.equal(ir.shouldBlock(), false);
+
   t.end();
 });
 
-test('block works', function(t) {
-  var ir = simpleIpRecord();
+test('suspect', t => {
+  const record = simpleIpRecord();
+  t.equal(record.isSuspected(), false);
```
shane-tomlinson: Oh no, a non-mocha test runner!

packages/fxa-customs-server/lib/server.js
```diff
@@ -212,7 +212,7 @@ module.exports = function createServer(config, log) {
       ipEmailRecord,
       smsRecord,
     }) {
-      if (ipRecord.isBlocked()) {
+      if (ipRecord.isBlocked() || ipRecord.isDisabled()) {
```
shane-tomlinson: Ahha, only IPs can be disabled, not accounts.

packages/fxa-customs-server/test/local/dataflow_tests.js
```diff
@@ -6,883 +6,678 @@ const proxyquire = require('proxyquire');
...
     });
 
-    test('unknown category logs an error', () => {
-      fetchSpy = sinon.spy();
-
-      checkForUnexpectedBlock({
-        customs_category: 'category',
-        accountid: 'testuser@testusre.com',
-        sourceaddress: '127.0.0.1',
+    test('old message', async () => {
```
shane-tomlinson: Does 'old message' mean it was emit before the epoch value from config?

packages/fxa-customs-server/lib/config/config.js
```diff
@@ -43,11 +43,25 @@ module.exports = function(fs, path, url, convict) {
...
+        default: '1 day',
+        env: 'SUSPECT_INTERVAL',
+        format: 'duration',
+      },
+      disableInterval: {
+        doc:
+          'Duration of a long-term `disable` flag, instigated by the Dataflow pipeline',
+        default: '1 year',
+        env: 'DISABLE_INTERVAL',
+        format: 'duration',
```
shane-tomlinson: Are durations automatically converted to MS? Should the `nat`s be converted to use `duration` (not for this PR of course)?

packages/fxa-customs-server/lib/config/config.js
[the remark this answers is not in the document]
vbudhram: Thats cool with me, I didn't realize that the `blockIntervalSeconds` was used in both.

packages/fxa-customs-server/lib/email_record.js
philbooth: > I see you check whether `this.su` or `this.di` have values below before comparing against them so I assume all is fine?

Correct.

packages/fxa-customs-server/lib/dataflow.js
```diff
@@ -38,203 +40,139 @@ module.exports = function(config, log, fetchRecords) {
...
+        op: 'dataflow.message.error',
+        id: message.id,
+        data: message.data,
+        error: error.message,
       });
-    }
 
-    if (block) {
-      _checkForUnexpectedBlock(block);
+      message.nack();
```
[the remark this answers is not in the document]
philbooth: Yep, that's my expectation.

It's the same approach we take [in the Amplitude pipeline](https://github.com/mozilla/fxa-amplitude-send/blob/a328aed55534afbd32a545c4a449657c5f99a3ac/pubsub/index.js#L138) fwiw, to push failed messages back onto the queue in case the failures are transient. Note that above there is also a condition to `ack` old messages with no action, so things won't get pushed back on to the queue indefinitely.

packages/fxa-customs-server/test/local/dataflow_tests.js
philbooth: Before `Date.now() - config.dataflow.ignoreOlderThan` specifically, yep.

packages/fxa-customs-server/lib/config/config.js
philbooth: > Are durations automatically converted to MS?

Yep, see:

https://github.com/mozilla/node-convict/blob/7483615640c4cfc0c9466744ec85a86a6460edcb/test/format-tests.js#L64-L67

And then:

https://github.com/mozilla/node-convict/blob/7483615640c4cfc0c9466744ec85a86a6460edcb/test/format-tests.js#L180-L182

> Should the `nat`s be converted to use `duration` (not for this PR of course)?

I considered the same, but am somewhat wary of making broad changes in this repo. But in theory, yep, they should be durations too.

philbooth:
> Probably worth opening an issue to discuss whether any remediation path should be available to users who are locked out.

I have an issue open for the technical implementation of that in #2471, not sure if that's what you had in mind? It doesn't really cover how the users would go about raising it with us, although I guess bugzilla would be the normal channel.

shane-tomlinson:
> I have an issue open for the technical implementation of that in #2471, not sure if that's what you had in mind? 

Perfect!

Issue #1107, opened 2019-05-15
[Train 137]Dotted square presented on the view of changing picture 


sv-sdeiac:
Environment:
Builds: Stage [Train 137];
Devices: Android, Windows 10 ;

Prerequisites:
You are logged on a valid account.

Steps to reproduce:
1. Go to Manage account;
2. Tap on 'Change' button from the Account picture section;

Expected results:
The change picture pop-up menu is displayed without any issues.

Actual results:
The change picture pop-up menu is displayed but a dotted square is presented on view.

Notes:
- Not reproducible on production;
- Video
![20190515_150116](https://user-images.githubusercontent.com/40264165/57775762-6fd1e500-7726-11e9-85b6-cc5531b6d079.gif)

ryanfeeley:
@clouserw Bump… 
![v3ruz45nrt4x](https://user-images.githubusercontent.com/68213/60831777-5f6b4300-a188-11e9-841a-a2969aa84968.jpg)

vbudhram:
from mtg:

We have some ideas on how this can be fixed. Bumping to next train until then.

PR #2457, opened 2019-09-10
fix(css): Fix change avatar spinner not loading correctly

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/1107

Image was not loading in stage because the relative path was not pointing to correct location. Fixed by loading the image from css.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
@@ -1,11 +1,8 @@
 <div id="avatar-change">
   <section class="modal-panel">
     <div class="main-avatar">
-      <div id="loading-avatar-spinner">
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
```diff
@@ -1,11 +1,8 @@
 <div id="avatar-change">
-  <section class="modal-panel">
+    <section class="modal-panel">
```
philbooth: This indentation change looks wrong? The `section` on line 2 should be in between the indentation level of the `div` elements on lines 1 and 3?

vbudhram: Ugh, not sure how changed, maybe code editor? I'll clean this up.

Issue #511, opened 2018-03-22
Add more automation/documentation around email template creation


vbudhram:
The steps and process to add a new email template to the auth-server has grown. We should add some automation and documentation for future us. These are a few off top of my head

* Create new template in `/lib/senders`, run `grunt template`
* Create a text version (file with `.txt`) of template
* Add template to [template index](https://github.com/mozilla/fxa-auth-server/blob/master/lib/senders/templates/index.js#L62)
* Add email logic in `/lib/senders/email.js`
* Add template name to [UTM mapping](https://github.com/mozilla/fxa-auth-server/blob/01d37ee77647d923f7310d3e1acc551314f4cf61/lib/senders/email.js#L30)
* Expose mailer function in `/lib/senders/index`, function must have these 3 params, `function (emails, account, opts)`
* Update local email [test cases](https://github.com/mozilla/fxa-auth-server/blob/01d37ee77647d923f7310d3e1acc551314f4cf61/test/local/senders/email.js#L26)
* Add to metric [templates](https://github.com/mozilla/fxa-auth-server/blob/c68105343e38e2c60bf4eb03aba86c48584efc3f/lib/metrics/amplitude.js#L79)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-684)
┆Issue Number: FXA-684

Cronus1007:
@vbudhram  If you are currently  not working on this issue shall I take over this issue?

Cronus1007:
@clouserw Shall I take over this issue.

PR #2459, opened 2019-09-10
fix(experiment): Fix signup code experiment so it reports to amplitude

### LINKED ISSUE #2458: Signup code metrics are not being reported in amplitude
We should have been receiving metrics for the Boomerang signup code experiment in Amplitude, but we are not. Forcing the experiment will give you the expected flow, however we do not report metrics when the flow is force.

Tracking bug to investigate why no metrics being reported.

cc @irrationalagent @davismtl

vbudhram:
Fixes #2458 

After a little digging, I found that the experiment was not included in the manual experiments. Once it was added there the experiment data was logged. I also updated the experiment docs to mention this.

Targeting as a point release on train 145

Screen Shot 2019-09-10 at 4 21 03 PM

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/amplitude.js
+++ b/packages/fxa-auth-server/lib/metrics/amplitude.js
@@ -38,6 +38,7 @@ const EMAIL_TYPES = {
   recoveryEmail: 'reset_password',
   unblockCode: 'unblock',
   verifyEmail: 'registration',
+  verifyShortCodeEmail: 'registration',
   verifyLoginEmail: 'login',
   verifyLoginCodeEmail: 'login',
   verifyPrimaryEmail: 'verify',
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/metrics/amplitude.js
```diff
@@ -38,6 +38,7 @@ const EMAIL_TYPES = {
   recoveryEmail: 'reset_password',
   unblockCode: 'unblock',
   verifyEmail: 'registration',
+  verifyShortCodeEmail: 'registration',
```
vbudhram: This was needed for amplitude to log the email metrics, it was in the documentation I just missed it 🤦🏽‍♂️

rfk: I wonder if we can prevent this in future by having a check when the auth-server starts up, that checks whether all the known emails have a corresponding entry in this list.

vbudhram: I agree, having a check would have helped catch this 👍🏽, we could probably/add extend this in this [issue](https://github.com/mozilla/fxa/issues/511).

PR #2474, opened 2019-09-11
feat(metrics): add metrics around SubHub calls

chenba:
This PR is for "appropriating the flow events" (in the words of Phil) to record performance metrics around auth server's integration with SubHub.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -76,7 +76,9 @@ module.exports = (log, db, config, customs, push, oauthdb, subhub) => {
       handler: async function(request) {
         log.begin('subscriptions.listPlans', request);
         await handleAuth(request.auth);
-        return subhub.listPlans();
+        const metricsContext = await request.gatherMetricsContext({});
+        const plans = subhub.listPlans(metricsContext);
+        return plans;
       },
     },
     {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -76,7 +76,9 @@ module.exports = (log, db, config, customs, push, oauthdb, subhub) => {
       handler: async function(request) {
         log.begin('subscriptions.listPlans', request);
         await handleAuth(request.auth);
-        return subhub.listPlans();
+        const metricsContext = await request.gatherMetricsContext({});
```
philbooth: One thing to be aware of is that the metrics context has an explicit, short lifetime that is a maximum of 2 hours or until the flow is completed. A "flow" in this case being a sign-in, sign-up or password-reset.

So in those cases, your empty object will be returned unadulterated and you won't have a `flowId`, which means your flow events won't get imported to Redshift. If that's a problem, it's probably fine if we extend the 2 hours to something larger. IIRC that was mostly a heuristic we could use to ditch obviously bogus events, so there should be leeway.

philbooth: Although tweaking the limit wouldn't help for completed flows of course, which are going to be all of them here right. Maybe we should start generating non-flow flow-ids for backend events that occur outside of the recognised flows, he wondered out loud...

philbooth: ...or maybe, and I think this is a saner suggestion, we should treat a signed-in user landing on settings as the start of a new flow, create a flow id then in the usual manner, and then use it to emit events corresponding with whatever the user does.

chenba: 😖 The lack of a flow id is definitely a problem if another RP is granted the subscription management scope to use the subscription routes.

philbooth: Honestly, I'd be tempted to emit some kind of new `performanceEvent` type here, include whatever data we have on it, then add tables and whatever scripting we need to surface those events separately in Redshift. Then later, as a follow-up, we can move across all the perf events that I piggy-backed on the flow metrics as a quick hack too.

chenba:
Closing this in favor of #2594.

PR #2481, opened 2019-09-11
feat(metrics): add support form metrics

### LINKED ISSUE #994: Add new events and properties for Subscription Support funnel

bbangert:
Closes #994

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/lib/metrics.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/metrics.js
@@ -46,6 +46,8 @@ describe('lib/metrics', function() {
         isSampledUser: true,
         lang: 'db_LB',
         notifier,
+        plan_id: 'plid',
+        product_id: 'pid',
         screenHeight: 1200,
         screenWidth: 1600,
         service: 'sync',
@@ -76,12 +78,13 @@ describe('lib/metrics', function() {
   });
 
   it('has the expected notifications', () => {
-    assert.lengthOf(Object.keys(metrics.notifications), 7);
+    assert.lengthOf(Object.keys(metrics.notifications), 8);
 
     assert.isTrue('flow.initialize' in metrics.notifications);
     assert.isTrue('flow.event' in metrics.notifications);
     assert.isTrue('set-email-domain' in metrics.notifications);
     assert.isTrue('set-sync-engines' in metrics.notifications);
+    assert.isTrue('set-plan-and-product-id' in metrics.notifications);
     assert.isTrue('set-uid' in metrics.notifications);
     assert.isTrue('clear-uid' in metrics.notifications);
     assert.isTrue('once!view-shown' in metrics.notifications);
@@ -315,6 +318,10 @@ describe('lib/metrics', function() {
         xhr: xhr,
       });
       notifier.trigger('set-uid', 'mock uid');
+      notifier.trigger('set-plan-and-product-id', {
+        planId: 'plid',
+        productId: 'pid',
+      });
     });
 
     afterEach(function() {
@@ -369,7 +376,7 @@ describe('lib/metrics', function() {
             var data = JSON.parse(
               windowMock.navigator.sendBeacon.getCall(0).args[1]
             );
-            assert.lengthOf(Object.keys(data), 33);
+            assert.lengthOf(Object.keys(data), 35);
             assert.equal(data.broker, 'none');
             assert.equal(data.context, Constants.CONTENT_SERVER_CONTEXT);
             assert.match(data.deviceId, /^[0-9a-f]{32}$/);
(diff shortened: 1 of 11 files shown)


bbangert:
@philbooth I look forward to hearing about places I missed something. 😉

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -235,6 +235,7 @@ module.exports = {
         pruneUnsetValues({
           service: serviceName,
           oauth_client_id: clientId,
+          plan_id: data.plan_id,
           product_id: data.product_id,
```
philbooth: Can you update one of the test cases to include `plan_id` and `product_id`?

packages/fxa-content-server/server/lib/routes/post-metrics.js
```diff
@@ -123,6 +123,8 @@ const BODY_SCHEMA = {
     })
     .optional(),
   numStoredAccounts: OFFSET_TYPE.min(0).optional(),
+  plan_id: STRING_TYPE.optional(),
+  product_id: STRING_TYPE.optional(),
```
philbooth: And can you add coverage for these to `tests/server/routes/post-metrics.js`?

packages/fxa-content-server/server/lib/amplitude.js
```diff
@@ -82,6 +82,7 @@ const VIEW_ENGAGE_SUBMIT_EVENT_GROUPS = {
   signup: GROUPS.registration,
   sms: GROUPS.connectDevice,
   subscribe: GROUPS.sub,
+  support: GROUPS.subSupport,
```
philbooth: Can you add test cases for the new view, engage and submit events to `tests/server/amplitude.js`? Ideally that module should have a test case for every event that we emit, although I think that practice may have slipped a bit recently.

packages/fxa-content-server/app/tests/spec/views/support.js
```diff
@@ -210,6 +274,8 @@ describe('views/support', function() {
             supportTicket
           );
           assert.isTrue(view.navigateToSubscriptionsManagement.calledOnce);
+          assert.isTrue(view.logFlowEvent.calledOnce);
+          assert.isTrue(view.logFlowEvent.calledWith('success', 'support'));
```
philbooth: Fwiw, I sometimes find these `isTrue` assertions unhelpful for debugging test failures because they just give a message like "expected false to be true". Where possible I think it produces more useful error messages if you can assert the actual value, so `assert.equal(view.logFlowEvent.callCount, 1)` in this case. Or in some places we use the assertions module from sinon itself, and those also produce helpful failures.

Not saying you have to do that here, just wanted to point it out in case it wasn't something you'd considered.

packages/fxa-content-server/app/tests/spec/lib/metrics.js
```diff
@@ -76,12 +76,13 @@ describe('lib/metrics', function() {
...
 
   it('has the expected notifications', () => {
-    assert.lengthOf(Object.keys(metrics.notifications), 7);
+    assert.lengthOf(Object.keys(metrics.notifications), 8);
 
     assert.isTrue('flow.initialize' in metrics.notifications);
     assert.isTrue('flow.event' in metrics.notifications);
     assert.isTrue('set-email-domain' in metrics.notifications);
     assert.isTrue('set-sync-engines' in metrics.notifications);
+    assert.isTrue('set-plan-and-product-id' in metrics.notifications);
```
philbooth: Looks like we also need a test to assert that this actually sets the ids correctly?

packages/fxa-content-server/server/lib/routes/post-metrics.js
bbangert: I don't see where these attributes are tested for in the `post-metrics.js` test, so adding them to an existing test is a bit of a head-scratcher for me.

philbooth: I was just thinking e.g. add them to these:

https://github.com/mozilla/fxa/blob/28cc9035ded8c6deaa9c9cea0affac3b60503623/packages/fxa-content-server/tests/server/routes/post-metrics.js#L92-L103

Just something to assert they make it through validation is all I had in mind.

packages/fxa-content-server/server/lib/amplitude.js
bbangert: Yup, added.

packages/fxa-content-server/app/tests/spec/lib/metrics.js
bbangert: Yup, added.

PR #2483, opened 2019-09-12
fix(payments): respond to /__version__

jrgm:
r? - @lmorchard

Adding standard `/__version__` route so it's easy to check the deployed version information.

I'm not sure about the change to `Dockerfile-test`, but it was repeating what had been done in the base :build image for the most part.

There is a benefit to docker image size of not installing `devDependencies` in the production image, and adding them into the `:test` image. But we can look at maybe doing that later.

### DIFF
--- a/packages/fxa-payments-server/package.json
+++ b/packages/fxa-payments-server/package.json
@@ -76,6 +76,7 @@
     "prettier": "^1.18.2",
     "redux-devtools-extension": "^2.13.8",
     "sass-lint": "^1.13.1",
+    "supertest": "4.0.2",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.1.0",
     "typescript": "^3.4.5",
--- a/packages/fxa-payments-server/server/lib/server.test.js
+++ b/packages/fxa-payments-server/server/lib/server.test.js
@@ -1,3 +1,53 @@
-it('works', () => {
-  expect(true).toEqual(true);
+const request = require('supertest');
+const server = require('./server')();
+const app = server.app;
+const pkg = require('../../package.json');
+
+function expectValueNotToBeUnknown(value) {
+  expect(value).toBeTruthy();
+  expect(value).not.toBe('unknown');
+}
+
+describe('Test simple server routes', () => {
+  test('__version__ should have correct structure', done => {
+    request(app)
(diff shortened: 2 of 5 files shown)


packages/fxa-payments-server/package.json
```diff
@@ -76,6 +76,7 @@
     "prettier": "^1.18.2",
     "redux-devtools-extension": "^2.13.8",
     "sass-lint": "^1.13.1",
+    "supertest": "4.0.2",
```
jrgm: Oh, actually, meant to say that this kinda picks the testing tool for server routes on this module, and I'm the last person who should be picking that. I can remove all the test related changes from this PR if something else would be more appropriate.

lmorchard: Yeah, we should probably rework the testing here... we already have progress on using jest for this stuff (e.g. [csp tests](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/server/lib/csp.test.js))

lmorchard: Oh, but maybe this works within jest... haven't had a chance to actually pull this down and try it yet

lmorchard: Actually: Never mind me - this *does* work within jest. For some reason, I thought this was a new test framework and not a tool that works within the framework. Probably a fine choice by virtue of 1) it works and 2) we didn't have any server route tests like this already.

packages/fxa-payments-server/server/lib/server.test.js
```diff
@@ -1,3 +1,53 @@
...
+const app = server.app;
+const pkg = require('../../package.json');
+
+function expectValueNotToBeUnknown(value) {
+  expect(value).toBeTruthy();
+  expect(value).not.toBe('unknown');
+}
+
+describe('Test simple server routes', () => {
+  test('__version__ should have correct structure', done => {
```
lmorchard: Only nit would be that we use `it()` elsewhere rather than `test()` - but it's not a showstopper.

packages/fxa-payments-server/package.json
jrgm: Yeah, still just. Just adding supertest for the route calls.

PR #2484, opened 2019-09-12
feat(cwts): allow skipping the CWTS screen

vladikoff:
cc @grigoryk @rfk feedback?


packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
```diff
@@ -52,6 +52,14 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
    * @private
    */
   onFxaStatus(response = {}) {
+    const skipCWTS =
+      response.capabilities && response.capabilities.skip_choose_what_to_sync;
```
vladikoff: the other version of this could be `    const cwtsStatus = response.capabilities && response.capabilities.choose_what_to_sync || true;` , default to `true`?

rfk:
Seems OK to me...capabilities are in general pretty ad-hoc so I'm not really able to suggest any clever alternatives here.

PR #2530, opened 2019-09-16
fix(vuln): updated hapi to fix https://npmjs.com/advisories/1165

dannycoates:


dannycoates:
welp `Error: Cannot find module '@hapi/address/lib/domain'` 🙄

PR #2531, opened 2019-09-17
fix(email): link to support form in subscription emails

### LINKED ISSUE #2030: Should subscription emails get a different SUMO link?
In #1791 we added a new email layout for subscription stuff. It incudes a support link that points at the main FxA SUMO article:

https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account

That doesn't seem like the right page to point to in the context of subscriptions. Do we have, or should we ask for a SUMO article specifically about subscriptions?

philbooth:
Fixes #2030.

@mozilla/fxa-devs r?


philbooth:
I also took the liberty of pushing a little secondary refactoring. Previously I made the other subscription URLs absolute in config, because I hadn't spotted the `baseUri` stuff at the bottom of the file. I just fixed them to hang off that instead.

I can split it out to a separate PR if it's a problem. I'll file a follow-up in `cloudups-deployment` if it lands.

packages/fxa-auth-server/lib/senders/subscription-templates/downloadSubscription.html
```diff
@@ -21,7 +21,7 @@ <h1 style="font-size: 1.1em; line-height: 1.5;">
 <tr style="page-break-before: always">
   <td align="left" valign="top" style="padding: 30px 30px 15px;">
     <p style="font-size: 0.9em; line-height: 1.8;">
-      Questions about your subscription? Our <a {{{supportLinkAttributes}}}>support team</a> is here to help you.
+      Questions about your subscription? Our <a href="{{{subscriptionSupportUrl}}}">support team</a> is here to help you.
```
shane-tomlinson: I'm presuming the strings are only going to be translated once we ship outside of the US?

philbooth: Affirmative

PR #2534, opened 2019-09-17
feat(login): let users opt-out of sync when signing in to Firefox

### LINKED ISSUE #2396: Let users opt-out of sync when signing in to Firefox
This issue is for the FxA-side details of [Bug 1577690](https://bugzilla.mozilla.org/show_bug.cgi?id=1577690).

As we expand the suite of products and services powered by FxA, we want users to be able to "Join Firefox" by signing in from about:welcome without forcing them to enable Sync. In conversation with @ryanfeeley we designed an approach that gives users this choice as part of the choose-what-to-sync screen.

### Current Experience

In the current approach, the user loads https://accounts.firefox.com/?service=sync&context=fx_desktop_v3 to sign in to Sync and they get the following after entering their credentials:

New user | Existing user 
--- | ---
 | (no CWTS screen)

Notably:

* New users *have* to make sync datatype selections in order to continue the flow. If they close the tab or choose "back" then they won't be signed in to Firefox.
* Existing users don't see anything at this step, they get signed in and start syncing as soon as their credentials are accepted.

We should keep this experience for browser touchpoints where the user is deliberately signing in to sync, such as the sync preferences page. (We should also [show choose-what-to-sync for all new sync users](https://bugzilla.mozilla.org/show_bug.cgi?id=1572344) rather than only for new FxA users, but that's an orthogonal issue).

### New Experience

I propose making the `service=sync` part of that URL optional, and having about:welcome omit it, because the user is signing in to Firefox but not intending to use any particular service.

In this case we'll offer the user to set up sync but give them the option to decline.  So they would load https://accounts.firefox.com/?context=fx_desktop_v3  and after entering credentials we would instead show the following:

New user | Existing user 
--- | ---
 | 

Notable changes:

* Existing users now see a kind of "choose whether to sync" screen.
* All users now have a "Don't sync this device" option which, if clicked, will continue the signin flow without enabling sync.
* There is no "back" option from this screen. I need to confirm with @ryanfeeley but I take this to mean that the user is successfully signed in to the browser at this point, and closing the tab or otherwise abandoning the flow is equivalent to choosing "don't sync this device".

vladikoff:
Fixes #2396
Blocks https://github.com/mozilla/application-services/issues/1690

# Testing this change

You can use this build of Firefox https://queue.taskcluster.net/v1/task/Z5Bu_T3oTrqcyUEKoml0Ug/runs/0/artifacts/public/build/target.dmg 

Open:

## http://127.0.0.1:3030/signin?&context=fx_desktop_v3&entrypoint=fxa_app_menu

(no sync parameter)

⚠️ Make sure to fix the autoconfig uri, it's currently setup to use https://nosync.dev.lcip.org/signup

Patch: https://treeherder.mozilla.org/#/jobs?repo=try&revision=7e412555d8d027f8e0c8b0b4ad5488a0e3e4f3b7

Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1577690


vladikoff:
@shane-tomlinson @vbudhram I hope you have time to look at this for Firefox 71 / FxA 147 👐

packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -273,7 +274,26 @@ const FxSyncChannelAuthenticationBroker = FxSyncAuthenticationBroker.extend(
      * @private
      */
     _getLoginData(account) {
-      const loginData = account.pick(ALLOWED_LOGIN_FIELDS);
+      const loginData = account.pick(ALLOWED_LOGIN_FIELDS) || {};
+      const isMultiService = this.relier && this.relier.get('multiService');
+      if (isMultiService) {
```
vladikoff: wrap this into a new function

packages/fxa-content-server/app/scripts/models/reliers/browser.js
```diff
@@ -20,6 +20,10 @@ const QUERY_PARAMETER_SCHEMA = {
   // context is not available when verifying.
   context: Vat.string().min(1),
   country: Vat.string().valid(...AllowedCountries),
+  // user is signing into the browser and declined to Sync
+  doNotSync: Vat.boolean(),
```
vladikoff: i don't think this should be part of QUERY_PARAMETER_SCHEMA , hmmm,..

packages/fxa-content-server/app/scripts/models/reliers/browser.js
```diff
@@ -34,6 +38,9 @@ const QUERY_PARAMETER_SCHEMA = {
 export default Relier.extend({
   defaults: _.extend({}, Relier.prototype.defaults, {
     action: undefined,
+    name: 'browser',
```
vladikoff: add a test for the name

packages/fxa-content-server/app/scripts/models/reliers/browser.js
```diff
@@ -50,6 +57,9 @@ export default Relier.extend({
 
       if (this.get('service')) {
         this.set('serviceName', t('Firefox Sync'));
+      } else {
+        // if no service provided, then we are just signing into the browser
+        this.set('serviceName', t('Firefox'));
```
vladikoff: add a test for the non service browser login

packages/fxa-content-server/app/scripts/models/reliers/base.js
```diff
@@ -12,6 +12,7 @@ import Backbone from 'backbone';
 var Relier = Backbone.Model.extend({
   defaults: {
     context: null,
+    name: 'base',
```
vladikoff: add a test for base

packages/fxa-content-server/app/scripts/templates/choose_what_to_sync.mustache
```diff
@@ -17,9 +17,14 @@
           </div>
         {{/engines}}
       </div>
-      <div>
-        <button id="submit-btn" type="submit" tabindex="1" autofocus>{{#t}}Continue{{/t}}</button>
+      <div class="button-row">
+        <button id="submit-btn" type="submit" tabindex="1" autofocus>{{#t}}Enable Sync{{/t}}</button>
+      </div>
+      {{#showDoNotSyncButton}}
```
vladikoff: need a test for `showDoNotSyncButton`

packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
```diff
@@ -59,9 +64,13 @@ const View = FormView.extend(
           // See #5554
           .then(() => this.broker.persistVerificationData(account))
           .then(() => {
-            this.waitForSessionVerification(account, () =>
-              this.validateAndSubmit()
-            );
+            if (!this.allowToDisableSync) {
```
vladikoff: add a test for no `waitForSessionVerification` case

packages/fxa-content-server/app/scripts/views/choose_what_to_sync.js
```diff
@@ -27,6 +28,10 @@ const View = FormView.extend(
       // a continuation function is passed in that should be called
       // when submit has completed.
       this.onSubmitComplete = this.model.get('onSubmitComplete');
+
+      // we only want to show the do not sync button for multi-service browser,
+      // that are not logging into sync and that came from a sign up flow.
+      this.allowToDisableSync = this.model.get('allowToDisableSync') || false;
```
vladikoff: remove the `false`

packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
```diff
@@ -104,6 +104,22 @@ export default {
           });
         }
 
+        if (
```
vladikoff: isolate into its own function

vladikoff:
> @vladikoff Did a first pass at this, but will do another. Is there a custom build of FF that I can use to test this?



Thank you!
Yeap see the `Testing this change` in the first comment of this PR.

packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -52,6 +52,12 @@ export default BaseAuthenticationBroker.extend({
    */
   onFxaStatus(response = {}) {
     const syncEngines = this.get('chooseWhatToSyncWebV1Engines');
+    const multiService =
+      response.capabilities && response.capabilities.multiService;
+    this.relier.set('multiService', multiService);
+    if (multiService) {
+      this.relier.set('service', response.clientId);
```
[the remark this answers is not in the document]
vladikoff: I will add some docs for it., but to explain here: 

> we return the OAuth client id for Desktop Firefox [in order to replace some uses of service=sync on the backend](https://github.com/mozilla/fxa/issues/2396#issuecomment-530662772)

shane-tomlinson:
When signing in, I am presented with the screen that asks whether I want to Sync. If I click the blue button to sync the device, CWTS displays and then I am redirected to the SMS form immediately:

skip-cwts

If I click the browser's "back" button from there, then "this.onFormSubmitComplete is not a function" shows up in the console.

Screenshot 2019-09-27 at 15 29 33

vbudhram:
I'm also getting 
Screen Shot 2019-09-27 at 10 32 32 AM

Which I think makes sense since there is no return function here?

https://github.com/mozilla/fxa/blob/f4b3caf368e757fd9b71d89e2614aa12558e2906/packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync-channel.js#L306

shane-tomlinson:
When signing up, clicking on "Don't sync this device" has no effect:

dont-sync-this-device-no-effect

vbudhram:
Hey @vladikoff, nice work! I was able to do some more testing after applying the return fix. Here some notes:

* No sync specific emails, was the intention to do https://github.com/mozilla/fxa/issues/2396#issuecomment-530662772 in a follow up or are we nuking sync specific messaging?
* I received a `Sync Enabled...will begin syncing momentarily` push notification, when I opted to not to enable sync. Would that be fixed as part of markh patch?
* When doing the sign-in and enable syncing, the CWTS page flashes and logs me in

screenflash

* Sign-in confirmation links works as expected! The user will get the `Confirm sign-in` email, and clicking it will verify the account

vladikoff:
Sorry the loginData issue has now been fixed, I accidentally refactored it out...

vladikoff:
>  I received a Sync Enabled...will begin syncing momentarily push notification, when I opted to not to enable sync. Would that be fixed as part of markh patch?

Yeap different, patch, filed in 
https://github.com/mozilla/application-services/issues/1893

vladikoff:
> No sync specific emails

yea! ff now sends client id

vladikoff:
@shane-tomlinson @vbudhram please try now!

packages/fxa-content-server/app/scripts/views/would_you_like_to_sync.js
```diff
@@ -0,0 +1,62 @@
...
+import BackMixin from './mixins/back-mixin';
+import Cocktail from 'cocktail';
+import DoNotSyncMixin from './mixins/do-not-sync-mixin';
+import FlowEventsMixin from './mixins/flow-events-mixin';
+import FormView from './form';
+import Template from 'templates/would-you-like-to-sync.mustache';
+
+const SCREEN_CLASS = 'screen-would-you-like-to-sync';
+
+const View = FormView.extend(
```
shane-tomlinson: Is the account already created and email sent when this view is displayed? If so, does this view need to poll similar to CWTS in case the user verifies, or are we punting on that b/c of boomerang?

packages/fxa-content-server/app/tests/spec/views/choose_what_to_sync.js
```diff
@@ -128,30 +129,62 @@ describe('views/choose_what_to_sync', () => {
...
+          sinon.spy(view, 'waitForSessionVerification');
+          return view.afterVisible();
+        })
+        .then(() => {
+          assert.isTrue(view.waitForSessionVerification.calledOnce);
+        });
+    });
 
-        assert.isTrue($('body').hasClass(View.SCREEN_CLASS));
+    it('does not poll', () => {
```
shane-tomlinson: When does this happen?

packages/fxa-content-server/app/scripts/views/would_you_like_to_sync.js
vladikoff: boomerang + I think in most cases when there's a choice of SYNC vs NOT-Sync, we want to wait for the user...

@ryanfeeley might confirm

packages/fxa-content-server/app/tests/spec/views/choose_what_to_sync.js
vladikoff: same scenario as https://github.com/mozilla/fxa/pull/2534#discussion_r329639774 , I spoke to @ryanfeeley and for now with email verification we do not poll when there's a choice of not syncing

vladikoff: same scenario as #2534 (comment) , I spoke to @ryanfeeley and for now with email verification we do not poll when there's a choice of not syncing

Issue #2511, opened 2019-09-13
Document how emails are sent


clouserw:
Some emails are sent from FxA...

* How did we choose which emails are sent from FxA and which are sent from another service?
* Which services do we route the emails through?
* What happens when they bounce?
* How are the templates localized?  How do we know which locale to send to people?

philbooth:
Have a document outline for this but realise it's going to be a lot easier to fill out after #1779 is done, because that gets rid of a load of stuff I won't have to waste time explaining. So I might try and make time to do that this week.

philbooth:
🧠💩:

https://docs.google.com/document/d/1SHu64k27fA1SdxSPcLQ3-b9i5hHvJZRXlb5XIf7RCIA/edit

philbooth:
Tentatively closing this based on the above linked doc, but re-open if there's more to do.

clouserw:
http://localhost:3000/ecosystem-platform/docs/fxa-engineering/fxa-email

PR #2535, opened 2019-09-17
refactor(email): reduce email-sending boilerplate

### LINKED ISSUE #1779: [auth server] Improve code re-use in email methods

philbooth:
Fixes #1779.

I've had a long-standing desire to clean up the email-sending code in the auth server and eliminate some of the repetition from it. But now I have to document that stuff for #2511 and I don't want to waste my time documenting bad code that we plan to refactor. And also my departure is looming and this is a loose end I don't want to leave hanging around. So I decided to see how far I could get with it today and what the outcome would look like.

As you can see from the headline, there's a net reduction of about 5.5k lines of code. ~~It's not all working yet, I still have to wade through some outstanding test failures and properly eyeball all of the rendered emails to make sure nothing is broken. But I'm encouraged with the direction. Opening as a draft to put it in the public domain in case people are interested.~~

Briefly the changes are:

* The remaining nunjucks stuff has been removed. All of the templates are now 100% handlebars.

* The templates have all been migrated to the layouts/partials structure I implemented for #1791 and the `subscription-templates` directory from that PR has been merged into the one-and-only `templates` directory.

* The extra layer of indirection for each function in `lib/senders/index.js` has mostly been replaced by a single function that works in all cases, except for `sendVerifySecondaryEmail` and `sendPostVerifySecondaryEmail`, which still have bespoke wrappers (for now).

* There's been some collateral clean-up of names to facilitate the reduction in boilerplate, which has hopefully made the mailer API more consistent to consume.

* Lots of repeated common markup in the templates has been extracted to new partials.

* There were a number of cases of broken markup that I spotted while creating those new partials, so we have well-formed markup now.

~~I'll come back tomorrow morning with fresh eyes to fix the stuff that's broken. I'm already aware of one merge conflict where templates were changed in `master` but have been reverted here. I'll be sure to remove those tomorrow, hopefully there won't be too many more merge conflicts after that.~~


packages/fxa-auth-server/lib/senders/templates/_versions.json
```diff
@@ -1,32 +1,33 @@
...
-  "passwordResetEmail": 3,
-  "passwordResetRequiredEmail": 3,
-  "postChangePrimaryEmail": 3,
-  "postRemoveSecondaryEmail": 3,
-  "postVerifyEmail": 3,
-  "postVerifySecondaryEmail": 3,
-  "postVerifyTrailheadEmail": 3,
-  "recoveryEmail": 3,
+  "downloadSubscription": 1,
+  "lowRecoveryCodes": 5,
```
philbooth: The template names in metrics now match the template names in the codebase, which improves greppability and debuggability when investigating things that show up in metrics.

Of course, it comes at the cost of a one-time discontinuity in the data, but we've been through those bumps before and they quickly fade with time. I think the naming consistency is a win worth paying that cost for.

I also bumped the version numbers, although that wasn't strictly necessary.

packages/fxa-auth-server/lib/senders/templates/downloadSubscription.html
```diff
@@ -16,7 +16,7 @@ <h1 style="font-size: 1.1em; line-height: 1.5;">
   </td>
 </tr>
 
-{{> button}}
+{{> subscriptionButton}}
```
philbooth: The subscription emails have a different button style.

packages/fxa-auth-server/lib/senders/templates/index.js
```diff
@@ -4,93 +4,115 @@
 
 'use strict';
 
+const LIB_DIR = '../..';
```
philbooth: This is just the same `index.js` that was previously in the `subscription-templates` directory. But because there was an old `index.js` here before, it's showing as a diff.

packages/fxa-auth-server/lib/senders/templates/index.js
```diff
@@ -4,93 +4,115 @@
...
+        data,
+      });
+      throw error.unexpectedError();
+    }
+
+    const layout = layouts.get(layoutName) || {};
+    let html, text;
+
+    if (template.html) {
+      html = renderWithOptionalLayout(template.html, layout.html, data);
```
philbooth: This `renderWithOptionalLayout` is a slight change to the old `subscription-templates` behaviour however. It's here because the SMS template doesn't have a layout, and the old code used to bailout if no layout was specified. Now it just renders the content template directly instead.

packages/fxa-auth-server/lib/senders/templates/layouts/fxa.txt
```diff
@@ -0,0 +1,9 @@
+{{{body}}}
+
+Mozilla. 331 E Evelyn Ave, Mountain View, CA 94041
+
+{{t "Mozilla Privacy Policy" }}
+{{{privacyUrl}}}
```
philbooth: In the process of moving this to the layout, I also opted to introduce some newline characters before links. Most of our text emails were an impenetrable wall of text previously, it's like no-one on the team had ever really looked at them.

Just by putting URLs on their own line, it makes them a bit more readable without causing any string changes for localisers. But bear it in mind if you're reviewing the changes locally, if you diff the before and after rendered text emails, you'll see the newlines show up.

packages/fxa-auth-server/lib/senders/templates/partials/automatedEmailChangePassword.html
```diff
@@ -0,0 +1,8 @@
+<tr style="page-break-before: always">
+  <td border="0" cellpadding="0" cellspacing="0" height="100%" width="100%">
+    <p class="secondary" style="font-family: sans-serif; font-weight: normal; margin: 0 0 12px 0; text-align: center; color: #737373; font-size: 11px; line-height: 18px; width: 310px !important; word-wrap:break-word">
```
philbooth: The old templates often had a weird `<br />` before this `<p>`. It turns out the `<br />` was there to "fix" problems caused by unclosed tags earlier in the document. Now that we have well-formed markup, I was able to get rid of it.

packages/fxa-auth-server/lib/senders/templates/passwordResetRequired.txt
```diff
@@ -3,13 +3,14 @@
 {{{t "To prevent further unauthorized access to your Firefox Account, we’ve reset your password and are notifying you as a precaution." }}}
 
 {{{t "What should you do?" }}}
-{{{t "First, change your Firefox Account password."}}} {{{ resetLink }}}
 
-{{{t "Second, if Firefox stores passwords to websites, you should change those passwords as well. Attackers may have access to those passwords, so you should check those accounts for suspicious activity."}}} {{{t "Learn how to see what passwords Firefox is storing for you."}}} {{{passwordManagerInfoUrl}}}
+{{{t "First, change your Firefox Account password."}}}
+{{{ resetLink }}}
```
philbooth: Again, newlines introduced for readabilty around URLs.

packages/fxa-auth-server/lib/senders/templates/password_reset_account_recovery.html
```diff
@@ -1,87 +0,0 @@
...
-        <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">{{t "You have successfully reset your password using a recovery key from the following device:" }}</p>
-        <p style="font-family:sans-serif; font-size: 13px; line-height: 20px; font-weight: normal; margin: 0 0 24px 0px; text-align: center; color: #4a4a4f;">
-    {{#if primaryEmail }}{{ primaryEmail }}<br/>{{/if}}
-    {{#if device }}{{ device }}<br/>{{/if}}
-    {{#if location }}{{ location }}<br/>{{/if}}
-    {{#if ip }}{{t "IP address: %(ip)s" }}<br/>{{/if}}
-    {{#if timestamp }}{{ timestamp }}<br/>{{/if}}
-</p>
-
-        <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "You will need to generate a new recovery key."}}
```
philbooth: Here's one of the examples of bad markup. This `<p>` never gets closed...

packages/fxa-auth-server/lib/senders/templates/password_reset_account_recovery.html
```diff
@@ -1,87 +0,0 @@
...
-                    <![endif]-->
-                </td>
-            </tr>
-        </table>
-    </td>
-</tr>
-<!--Button Area-->
-<tr style="page-break-before: always">
-    <td border="0" cellpadding="0" cellspacing="0" height="100%" width="100%">
-        <br/>
```
philbooth: ...and this `<br />` is added as an apparent consequence.

packages/fxa-auth-server/lib/senders/templates/verify_login.html
```diff
@@ -1,86 +0,0 @@
...
-    {{/if}}
-  </td>
-</tr>
-
-
-<!--Header Area-->
-<tr style="page-break-before: always">
-  <td valign="top">
-    <h1 style="font-family: sans-serif; font-size: 21px; line-height: 29px; font-weight: normal; margin: 0 0 11px 0; text-align: center;">{{t "New sign-in to %(clientName)s" }}</h1>
-    <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">{{t "For added security, please confirm this sign-in from the following device:"}}<p/>
```
philbooth: There were a few examples like this too, where the author obviously intended to add a closing `</p>` but typoed it as `<p/>` by mistake, creating an extra, nested empty element instead of closing the parent one.

packages/fxa-auth-server/scripts/template-version-bump.js
```diff
@@ -42,11 +36,7 @@ const stagedTemplates = cp
   })
   .map(templatePath => templatePath.split('/')[5])
   .filter(fileName => !IGNORE.has(fileName))
-  .map(fileName =>
-    templates.generateTemplateName(
```
philbooth: `generateTemplateName` is unnecessary now because the templates have the same name as the strings in metrics.

packages/fxa-auth-server/test/local/senders/index.js
```diff
@@ -243,12 +243,12 @@ describe('lib/senders/index', () => {
 
             const args = email._ungatedMailer.postVerifyEmail.getCall(0).args;
             assert.equal(args[0].email, EMAIL, 'email correctly set');
-            assert.equal(args[0].ccEmails, undefined, 'no cc emails set');
+            assert.lengthOf(args[0].ccEmails, 1);
```
philbooth: This is an artifact of using that common reducer function, but it's a moot point. Newly verified accounts will only have one email address so `ccEmails` will never be set in practice.

philbooth:
Also, in order to verify the changes, there are 3 things you can do:

1. Just in local dev, you can verify that I haven't broken integration with the local mail helper. Check that the links still show up in `./pm2 logs 0`.

2. Also locally, you can run `node scripts/write-emails-to-disk all` from the auth server package directory. That dumps all of the rendered emails to a directory called `.mail_output` and you can then eyeball them in a browser to make sure there are no differences to the previous output, other than the vertical spacing fixes I mentioned in the inline comments.

3. I've deployed this branch to https://pb-email.dev.lcip.org/. Go and sign up, sign in, reset password, whatever. Make sure the emails are sent and that they look correct in your mail client (fwiw I just pushed a fix for a styling bug I found this way, it will take a few minutes for that to get deployed on the box).

packages/fxa-auth-server/test/local/senders/index.js
vbudhram: You can probably pass in a more correct `EMAILS` value here to test `ccEmails` is 0 or undefined.

packages/fxa-auth-server/lib/senders/index.js
```diff
@@ -9,292 +9,90 @@ const createSms = require('./sms');
...
-    },
-    sendRecoveryCode(emails, account, options) {
-      const { to, cc } = splitEmails(emails);
-
-      return mailer.recoveryEmail({
-        ...options,
-        acceptLanguage: options.acceptLanguage || defaultLanguage,
-        ccEmails: cc,
-        email: to,
-        emailToHashWith: account.email,
```
vbudhram: Nice! By removing this level, less likely to miss passing some of the email options.

Issue #2478, opened 2019-09-11
[fxa-oauth-server] We should be stricter with OAuth scopes


vladikoff:
Improvement can be made to make the scope list stricter. We should cache allowed the scope list and validate against it.
See : https://github.com/mozilla/fxa/pull/2442#discussion_r323148483

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-345)
┆Issue Number: FXA-345

shane-tomlinson:
Adding the context from the original comment:

> If an RP with allowedScopes defined requests a short scope that isn't in the allowedScopes list, then that scope will be allowed. Likewise, if a requested long scope name is not in the DB, and not in allowedScopes, we still allow it. It seems to me that we should be strict and only allow scopes that are in the list whether those scopes are in the DB or not. Which leads to the question of why do we do it this way?

PR #2538, opened 2019-09-18
feat(jwt-access-tokens): Add subscription info to JWT access tokens

### LINKED ISSUE #1595: Add subscription capabilities to JWT access token
Depends on #935 

RPs can query a user's subscription capabilities in the profile server / userinfo endpoint. However, it would also be useful for some RPs for this information to appear as a claim in a JWT format access token. That way, it's supplied on login without the need for an API request.

As a hint for how capabilities are derived, [here is where it gets done in the `/account/profile` route on  fxa-auth-server](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/account.js#L867).

The algorithm for generating the list is extracted into [a `determineClientVisibleSubscriptionCapabilities` utility function](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/utils/subscriptions.js#L14).

shane-tomlinson:
Make a backendService request from the oauth server to the auth
server to get the user's subscription info since the oauth server
does not have that state.

fixes #1595

Note, using a backendService request won't be necessary once @dannycoates's work to merge the auth/oauth servers is complete because the oauth server code will just be able to invoke the /account/profile route handler directly and have it "just work". 

Also note that since /account/profile is a GET request, I had to use query params instead of POST params. It's ugly. When I tried to request only the "profile:subscriptions" scope, safeUrl said that was an unsafe query parameter. If I just try for "profile", no problem. I tried encodeURIComponent on `profile:subscriptions`, but it still complained.

Left to do:
* [x] - Get feedback whether this approach is even remotely reasonable.
* [x] - Determine whether we want subscriptions appended as scopes, or as their own claim. I can make a case for either. I went with a separate claim for now while I work through this.
* [x] - Only allow query params on the /account/profile route if using the `profileSecret` auth strategy.
* [x] - tests

@lmorchard, @bbangert, @rfk, @dannycoates - f?

### DIFF
--- a/docs/adr/0007-subscription-claim-jwt-access-token.md
+++ b/docs/adr/0007-subscription-claim-jwt-access-token.md
@@ -1,0 +1,82 @@
+# Placing subscription info in the `fxa-subscriptions` claim of JWT access tokens
+
+- Status: proposed
+- Deciders: Shane Tomlinson
+- Date: 2019-09-30
+
+Technical Story: https://github.com/mozilla/fxa/issues/1595
+
+## Context and Problem Statement
+
+A mechanism is needed to inform service providers (SP) that a user has paid for a given subscription. A SP could fetch the user's profile information, however, we prevent some 3rd party SPs from doing so, and instead give them all the information they need in a JWT format access token. Adding subscription information in JWT access tokens will give SPs the information they need to verify users have paid for a subscription.
+
+## Decision Drivers
+
+- Security - Users should not be able to get access to subscriptions they have not paid for.
+- Extensibility - Adding subscription info should not inhibit future extensions to JWT access tokens.
+- Standards - The [JWT access token draft spec][#jwt-draft-spec] format should be followed as closely as possible.
+
+## Considered Options
+
+1. Add subscriptions to the `scopes` claim
+2. Add subscriptions as its own claim, `fxa-subscriptions`
+
+## Decision Outcome
+
+Chosen option: Adding subscription info into its own claim was chosen because FxA's lax scope checking means bad acting users could grant themselves access to subscriptions they have not paid for. See [this bug regarding FxA's lax scope checking][#lax-scope-checking].
+
+### Positive Consequences
+
+- The JWT access token contains all the information an SP needs to verify the user has paid for a subscription.
+- Users are unable to grant themselves access to subscriptions they have not paid for.
+
+### Negative Consequences
+
+- SPs must now check two claims from the JWT to ensure a user is able to access a protected resource.
+- An additional claim is added that is not defined in [the JWT access token draft spec][#jwt-draft-spec].
+
+## Pros and Cons of the Options
+
+### Add subscriptions to the `scopes` claim
+
+An access token for a user that has paid for `subscription1` would have `subscription1` in it's scope claim.
+
+e.g.,
+
+```json
+{
+  "jti": "cafecafe",
+  "sub": "deadbeef",
+  "scope": "profile:read subscription1",
+  ...
+}
+```
+
+- Good, because subscription information is in `scope` which is most likely the expected claim in which an SP would look for this info.
+- Good, because no new claims are added over what's defined in [the JWT access token draft spec][#jwt-draft-spec].
+- Bad, because [FxA's lax scope checking][#lax-scope-checking] means users would grant themselves access to subscriptions they have not paid for.
+
+### Add subscriptions to the `fxa-subscriptions` claim
+
+An access token for a user that has paid for `subscription1` would have `subscription1` in it's scope claim.
+
+e.g.,
+
+```json
+{
+  "jti": "cafecafe",
+  "sub": "deadbeef",
+  "scope": "profile:read",
+  "fxa-subscriptions": "subscription1"
+  ...
+}
+```
+
+- Good, because users are unable to grant themselves access to subscriptions they have not paid for.
+- Good, because subscription information is isolated from other possible future JWT extensions.
+- Bad, because `fxa-subscriptions` is a non-standard claim and SP developers may not expect to look there.
+
+## Links
+
+[#jwt-draft-spec]: https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00
+[#lax-scope-checking]: https://github.com/mozilla/fxa/issues/2478
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const createBackendServiceAPI = require('../../lib/backendService');
+const Joi = require('joi');
+const { signJWT } = require('../../lib/serverJWT');
+const AppError = require('./error');
+
+module.exports = (log, config) => {
+  const AuthServerAPI = createBackendServiceAPI(log, config, 'auth', {
+    getUserProfile: {
+      path: '/v1/account/profile',
+      method: 'GET',
+      validate: {
+        headers: {
+          authorization: Joi.string().required(),
+        },
+        response: {
+          email: Joi.string().optional(),
+          locale: Joi.string()
+            .optional()
+            .allow(null),
+          authenticationMethods: Joi.array()
+            .items(Joi.string().required())
+            .optional(),
+          authenticatorAssuranceLevel: Joi.number().min(0),
+          subscriptions: Joi.array()
+            .items(Joi.string().required())
+            .optional(),
+          profileChangedAt: Joi.number().min(0),
+        },
+      },
+    },
+  });
+
+  const api = new AuthServerAPI(config.auth.url, config.auth.poolee);
+
+  return {
+    api,
+
+    close() {
+      api.close();
+    },
+
+    async getUserProfile({ client_id, scope, uid }) {
+      const claims = {
+        client_id,
+        scope,
+        sub: uid,
+      };
+      const jwt = await signJWT(
+        claims,
+        config.auth.url,
+        config.publicUrl,
+        config.auth.jwtSecretKey
+      );
+      try {
+        return await api.getUserProfile({ authorization: `OAuthJWT ${jwt}` });
+      } catch (error) {
+        throw this.mapAuthError(error);
+      }
+    },
+
+    mapAuthError(error) {
+      // If it's already an instance of our internal error type,
+      // then just return it as-is.
+      if (error instanceof AppError) {
+        return error;
+      }
+      if (!error.errno) {
+        // If there's no `errno`, it must be some sort of internal implementation error.
+        // Let it bubble up and be caught by the top-level unexpected-error-handling logic.
+        throw error;
+      }
+
+      switch (error.errno) {
+        case 110: {
+          return AppError.invalidToken();
+        }
+        case 998: {
+          let key;
+          try {
+            key = Object.keys(error.output.payload.data.value)[0];
+          } catch (e) {
+            // ignore, no key found
+          }
+          return AppError.invalidRequestParameter(key);
+        }
+        default: {
+          log.warn('auth_server.mapAuthError', {
+            err: error,
+            errno: error.errno,
+            warning: 'unmapped auth-server errno',
+          });
+          return AppError.unexpectedError();
+        }
+      }
+    },
+  };
+};
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
@@ -30,6 +30,32 @@ const conf = convict({
     format: Boolean,
     default: false,
   },
+  auth: {
+    poolee: {
+      timeout: {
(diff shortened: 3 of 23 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1026,13 +1037,13 @@ module.exports = (
...
         ) {
-          const capabilities = await determineClientVisibleSubscriptionCapabilities(
+          const capabilities = (await determineClientVisibleSubscriptionCapabilities(
             config,
             auth,
             db,
             uid,
             client_id
-          );
+          )) || ['my-sub'];
```
shane-tomlinson: I'll remove 'my-sub' once I get further along.

packages/fxa-auth-server/lib/server.js
```diff
@@ -382,6 +382,19 @@ async function create(log, error, config, routes, db, oauthdb, translator) {
   }));
   server.auth.strategy('subscriptionsSecret', 'subscriptionsSecret');
 
+  server.auth.scheme('profileSecret', () => ({
+    async authenticate(request, h) {
+      if (
+        constantTimeCompare(request.headers.authorization, SUBSCRIPTIONS_SECRET)
```
shane-tomlinson: This is identical to `subscriptionsSecret` above, maybe just use that?

packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
```diff
@@ -0,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const createBackendServiceAPI = require('../../lib/backendService');
+const Joi = require('joi');
+
+module.exports = (log, config) => {
+  const AuthServerAPI = createBackendServiceAPI(log, config, 'auth', {
+    getUserProfile: {
```
shane-tomlinson: Since all we really want is the subscription info, maybe scope this down to `getUserSubscriptions` and do not allow a scope, and only allow `subscriptions` to be returned.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -995,6 +1002,10 @@ module.exports = (
           uid = auth.credentials.uid;
           scope = { contains: () => true };
           client_id = null;
+        } else if (auth.strategy === 'profileSecret') {
```
shane-tomlinson: Replace `profileSecret` with `subscriptionSecret` and with that, hard code the `profile:subscriptions` scope.

philbooth:
> safeUrl said that was an unsafe query parameter

Just want to point out it's totally fine to change the behaviour of `safeUrl`. I just used a simple regex in mozilla/fxa-auth-serve#2368, there was no explicit intention to disallow `:`. Newline characters were the specific thing we were protecting against at the time.

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -222,5 +223,19 @@ exports.generateAccessToken = async function generateAccessToken(grant) {
     return accessToken;
   }
 
+  // This is awful. The auth server is the canonical source of truth
```
rfk: I agree with your assessment, but I haven't been able to come up with a better suggestion, especially since we need to profile info to update to reflect the current state of the user's account when using the refreshToken grant.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -966,7 +966,14 @@ module.exports = (
...
       options: {
         auth: {
-          strategies: ['sessionToken', 'oauthToken'],
+          strategies: ['sessionToken', 'oauthToken', 'profileSecret'],
+        },
+        validate: {
+          query: {
+            client_id: isA.string().optional(),
+            scope: isA.string().optional(),
+            uid: isA.string().optional(),
```
rfk: If you didn't want to put these in the query params, you could bundle them up into a JWT and sign that with the shared secret, then have the `profileSecret` auth strategy extract and validate them for you. We already have the pieces needed to do that, because it's basically how auth-server talks to oauth-server right now.

shane-tomlinson: I went down that route, but stopped because the auth server passes the JWT in [POST body params](https://github.com/mozilla/fxa/blob/a142c2528ea83dfd2852e0e5fa9cd79e74193904/packages/fxa-auth-server/lib/oauthdb/index.js#L117) and since this is a GET request, we'd still have to pass the JWT as a query param. That, or I'd have to modify backendService.js to accept a header as part of the request instead of when [creating the Service](https://github.com/mozilla/fxa/blob/a142c2528ea83dfd2852e0e5fa9cd79e74193904/packages/fxa-auth-server/lib/backendService.js#L86). Seemed like a bunch of extra hoops to go through to clean up code we want to soon tear out.

shane-tomlinson:
> Broadly though, I'm still not clear on exactly why we're sending this data in this way. 

It's early in the WIP process and I wanted to explore. This approach didn't seem quite right, so I updated today to return the subscriptions on the scopes.

shane-tomlinson:
> Just want to point out it's totally fine to change the behaviour of safeUrl

Oh, I had always assumed this was an external utility! Thanks for the tip, updated!

packages/fxa-auth-server/lib/routes/account.js
shane-tomlinson: So, I was just exploring with modifying backendService.js and that's not as bad as I expected, but then realized the way I implemented jwt creation and validation makes assumptions about which key to use, and then over on the auth server, we'd have to use the oauth server's jwt validation logic. It's a bit of a mess really.

shane-tomlinson:
Converting this to READY!

packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
```diff
@@ -0,0 +1,76 @@
...
+  };
+};
+
+function mapAuthError(error) {
+  switch (error.errno) {
+    case 110: {
+      return AppError.invalidToken();
+    }
+    default: {
+      return new AppError();
```
vbudhram: Should we be passing in the `error` object here or rethrow it?

packages/fxa-auth-server/lib/server.js
```diff
@@ -13,7 +13,8 @@ const path = require('path');
 const url = require('url');
 const userAgent = require('./userAgent');
 const schemeRefreshToken = require('./scheme-refresh-token');
-
+const schemeServerJWT = require('./scheme-serverJWT');
+const { verifyJWT } = require('./serverJWT');
```
vbudhram: I don't see this used in here?

packages/fxa-auth-server/lib/scheme-serverJWT.js
```diff
@@ -0,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const { verifyJWT } = require('./serverJWT');
```
vbudhram: Since there are two files prefixed with `scheme` we could pull them into a specific folder.

packages/fxa-auth-server/test/local/serverJWT.js
```diff
@@ -0,0 +1,117 @@
...
+        assert.equal(args[1], 'old');
+        assert.deepEqual(args[2], {
+          algorithms: ['HS256'],
+          audience: 'foo',
+          issuer: 'bar',
+        });
+      });
+    });
+
+    describe('no key found', () => {});
```
vbudhram: Missing tests?

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
```diff
@@ -35,6 +35,10 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
     sub: await sub(grant.userId, grant.clientId, grant.ppidSeed),
   };
 
+  if (grant['fxa-subscriptions']) {
+    claims.scope = `${claims.scope} ${grant['fxa-subscriptions']}`;
```
lmorchard: Okay, so I tweaked my local configs a bit, subscribed to 123done Pro locally. I can see subscription capabilities appearing in the `scope` of the JWT access token:
```
"scope": "profile 123donePro",
```
Seems kind of weird that these are mingled, IMO: subscription capabilities aren't *exactly* like OAuth scopes. RPs can't request a token with them, we expose them using the intersection of user subscriptions and RP capabilities.

But, I'm also vague on my OAuth-fu - so if this makes pragmatic & standards-wise sense, it's not a thing I'd fuss too much over.

shane-tomlinson: I'm going to go back to adding a new claim for subscriptions because [scope checking is very lax](https://github.com/mozilla/fxa/issues/2478). A malicious user who figures out what the value of the subscription is could load up FxA to sign into the proxy, edit the URL to add the subscription name to the scope list, sign in, and the subscription value would end up in the scope list in the JWT sent to the service provider. If we make scope checking more restrictive, including taking into account subscription values, we can re-visit this. I just tried this using local 123done and it works. :(

Thanks for pushing back here @lmorchard.

docs/adr/0007-subscription-claim-jwt-access-token.md
```diff
@@ -0,0 +1,82 @@
+# Placing subscription info in the `fxa-subscriptions` claim of JWT access tokens
```
lmorchard: Yay for ADRs! 🎉

packages/fxa-auth-server/fxa-oauth-server/lib/grant.js
```diff
@@ -222,5 +223,24 @@ exports.generateAccessToken = async function generateAccessToken(grant) {
...
+    const { subscriptions } = await authServer.getUserProfile({
+      client_id: hex(grant.clientId),
+      scope: 'profile:subscriptions',
+      uid: hex(grant.userId),
+    });
+    // To avoid mutating the input grant, create a
+    // copy and add the new property there.
+    grant = {
+      ...grant,
+      'fxa-subscriptions': subscriptions,
```
lmorchard: :+1:

packages/fxa-auth-server/fxa-oauth-server/lib/jwt_access_token.js
```diff
@@ -35,6 +35,16 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
...
   };
 
+  // Note, a new claim is used rather than scopes because
+  // FxA's scope checking somewhat blindly accepts user input,
+  // meaning a malicious user could reload FxA after editing the URL
+  // to contain subscription name in the scope list and the subscription
+  // would end up in the user's scope list whether they actually
+  // paid for it or not. See https://github.com/mozilla/fxa/issues/2478
+  if (grant['fxa-subscriptions']) {
+    claims['fxa-subscriptions'] = grant['fxa-subscriptions'].join(' ');
```
lmorchard: 👍

docs/adr/0007-subscription-claim-jwt-access-token.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+- Good, because subscription information is in `scope` which is most likely the expected claim in which an SP would look for this info.
+- Good, because no new claims are added over what's defined in [the JWT access token draft spec][#jwt-draft-spec].
+- Bad, because [FxA's lax scope checking][#lax-scope-checking] means users would grant themselves access to subscriptions they have not paid for.
+
+### Add subscriptions to the `fxa-subscriptions` claim
+
+An access token for a user that has paid for `subscription1` would have `subscription1` in it's scope claim.
```
chenba: Nit: `s/it's/its/`

packages/fxa-auth-server/fxa-oauth-server/test/jwt_access_token.js
```diff
@@ -77,6 +77,18 @@ describe('lib/jwt_access_token', () => {
...
       ]);
     });
+
+    it('should propagate `fxa-subscriptions`', async () => {
+      requestedGrant['fxa-subscriptions'] = ['subscription1', 'subscription2'];
+      await JWTAccessToken.create(mockAccessToken, requestedGrant);
+      const signedClaims = mockJWT.sign.args[0][0];
+      assert.lengthOf(Object.keys(signedClaims), 8);
+
+      assert.deepEqual(
```
shane-tomlinson: This should probably be an `equal` or `strictEqual`.

packages/fxa-auth-server/fxa-oauth-server/lib/config.js
```diff
@@ -30,6 +30,32 @@ const conf = convict({
...
+    },
+    jwtSecretKey: {
+      default: 'megaz0rd',
+      doc: 'Shared secret for signing oauth-to-auth server JWT assertions',
+      env: 'AUTH_SERVER_SHARED_SECRET',
+      format: String,
+    },
+    url: {
+      default: 'http://127.0.0.1:9000',
+      format: 'url',
```
shane-tomlinson: Ah man, I forgot to add an ENV and document for this.

PR #2541, opened 2019-09-19
fix(payments): only show zip input error onBlur

### LINKED ISSUE #2096: Zip code error jumping the gun
Every time I click into the zip code on the billing form and add a number it immediately gives me an error. Can we remove this and have the error only show up if the user clicks outside of the form element?

Screen Shot 2019-08-06 at 11 21 40 AM

meandavejustice:
- fixes #2096

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -26,9 +26,9 @@ import { AppContext } from '../../lib/AppContext';
 import './index.scss';
 import { Plan } from '../../store/types';
 
-export const SMALL_DEVICE_RULE = "(max-width: 480px)";
-export const SMALL_DEVICE_LINE_HEIGHT = "37px";
-export const DEFAULT_LINE_HEIGHT = "45px";
+export const SMALL_DEVICE_RULE = '(max-width: 480px)';
+export const SMALL_DEVICE_LINE_HEIGHT = '37px';
+export const DEFAULT_LINE_HEIGHT = '45px';
 
 // ref: https://stripe.com/docs/stripe-js/reference#the-elements-object
 let stripeElementStyles = {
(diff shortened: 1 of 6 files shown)


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -111,7 +116,7 @@ export const PaymentForm = ({
     [validator, onPayment, onPaymentError, stripe]
   );
 
-  stripeElementStyles = checkMedia(matchMedia(SMALL_DEVICE_RULE), stripeElementStyles);
+  stripeElementStyles = checkMedia({ matched: matchMedia(SMALL_DEVICE_RULE), stripeElementStyles });
```
lmorchard: This looks like a change not related to the zip field

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -49,7 +49,11 @@ type StripeElementStyles = {
   lineHeight: string,
 };
 
-export function checkMedia(matched: boolean, stripeElementStyles: { base: StripeElementStyles }) {
+export function checkMedia({ matched, stripeElementStyles }: {
```
lmorchard: This change looks unrelated too

packages/fxa-payments-server/src/components/fields.tsx
```diff
@@ -130,6 +133,8 @@ export const Input = (props: InputProps) => {
     [name, validator, onValidate, required]
   );
 
+  const onChangeWrapper = (ev: any) => onChange(ev);
```
lmorchard: Doesn't look like this wrapper does anything? Should also be defined using `React.useCallback` so as not to be re-created on every component render.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
meandavejustice: This was a typescript checker change

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
meandavejustice: same as ⬆️ (typescript checker)

packages/fxa-payments-server/src/components/fields.tsx
meandavejustice: This is to keep the typescript checker happy. It doesn't like assigning a function with two arguments to `onChange`

meandavejustice:
@lmorchard Thanks for your review, it somehow slipped through my radar. I can pull the stripeElementStyles changes out into a separate pr, add max length, try and figure out the not calling onBlur on selecting checkbox. And the other changes I think we should punt(I can file follow up issues), as I've spent a ton of time on such a minor change already.

meandavejustice:
@lmorchard Pulled out the stripeElementStyles refactoring stuff out and made the `maxlength` updates. Should be pretty clean and straight forward now.

lmorchard:
Yeah, I think this is looking pretty good now. Still see an issue where if the zip code field is valid but not blurred away from, the submit button never enables. But I think that might be an edge case to fix if it ends up being a real problem

meandavejustice:
> Still see an issue where if the zip code field is valid but not blurred away from, the submit button never enables
@lmorchard What are the steps to reproduce? I wanna file a followup issue and I haven't seen that myself

lmorchard:
>What are the steps to reproduce? I wanna file a followup issue and I haven't seen that myself

Basically, this thing [I mentioned earlier](https://github.com/mozilla/fxa/pull/2541#discussion_r328781869):

> If I click the confirmation checkbox first, then fill out all the form fields but never blur away from zip code - the value doesn't get validated and the button never gets enabled. Maybe this is more an edge case? Will probably end up filed as an issue though

PR #2547, opened 2019-09-19
fix(oauth): fix the status check for the oauth webchannel broker

### LINKED ISSUE #2539: Support visiting the settings page with `context=oauth_webchannel_v1`
The "manage account" menu item in Fenix will link to the settings page like so:

  https://latest.dev.lcip.org/settings?email=foo@bar.net&service=sync&context=oauth_webchannel_v1

Visiting this URL *should* use an `fxa_status` webchannel message to sync login state with the browser, and then display the user's settings page.

When @grigoryk tried this out just now, the `fxa_status` message went through OK, but it then directed him to the signin page.  Let's make sure we're treating the new `context=oauth_webchannel_v1` the same as `context=fxa_desktop_v1` in this case and letting it load the settings page.

vladikoff:
Fixes #2539

@vbudhram r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
@@ -25,7 +25,6 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
 
   defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
     chooseWhatToSyncWebV1: true,
-    fxaStatus: true,
     openWebmailButtonVisible: false,
   }),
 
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/tests/functional/oauth_webchannel.js
```diff
@@ -123,5 +125,30 @@ registerSuite('oauth webchannel', {
...
         .then(testIsBrowserNotified('fxaccounts:oauth_login'));
     },
+    settings: function() {
+      const SETTINGS_PAGE_URL = `${config.fxaContentRoot}settings?context=oauth_webchannel_v1`;
+
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
+        .then(result => {
+          return this.remote.then(
+            openPage(SETTINGS_PAGE_URL, null, {
```
vladikoff: hm this test only passes sometimes...

packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
```diff
@@ -38,11 +37,13 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
     this._channel = options.channel;
     this._scopedKeys = ScopedKeys;
     this._metrics = options.metrics;
+    options.fxaStatus = true;
 
     proto.initialize.call(this, options);
-    this.request(this.getCommand('FXA_STATUS'), {
```
vladikoff: we don't need this request. `.fetch` already does it for us.

PR #2549, opened 2019-09-19
fix(event-broker): response to /__version__ route

jrgm:
Return version information on `/__version__`

Like `curl 'https://oauth.accounts.firefox.com/__version__'`:
```
{
  "version": "1.145.4",
  "commit": "04f47a3c381ca922dec07498e92c29bb5b5eba13",
  "source": "https://github.com/mozilla/fxa"
}
```

### DIFF
--- a/packages/fxa-event-broker/lib/api/proxy-controller.ts
+++ b/packages/fxa-event-broker/lib/api/proxy-controller.ts
@@ -11,6 +11,7 @@ import { JWT } from '../jwts';
 import { ClientWebhookService } from '../selfUpdatingService/clientWebhookService';
 import { DELETE_EVENT, SUBSCRIPTION_UPDATE_EVENT } from '../serviceNotifications';
 import { proxyPayload } from './proxy-validator';
+import { version } from './version';
 
 export default class ProxyController {
   constructor(
@@ -22,6 +23,14 @@ export default class ProxyController {
 
   public async heartbeat(request: hapi.Request, h: hapi.ResponseToolkit) {
     return h.response({}).code(200);
+  }
+
+  public async version(request: hapi.Request, h: hapi.ResponseToolkit) {
+    const body = JSON.stringify(version);
+    return h
+      .response(body)
+      .type('application/json')
+      .code(200);
   }
 
   public async proxyDelivery(request: hapi.Request, h: hapi.ResponseToolkit) {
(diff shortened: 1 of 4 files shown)


packages/fxa-event-broker/lib/version.ts
```diff
@@ -0,0 +1,71 @@
...
+      commit,
+      source,
+      version: packageInfo.version
+    };
+  }
+
+  return {};
+}
+
+export const version = getVersionInfo();
```
jrgm: Typescript corrections welcome ;-)

bbangert: This file should probably be under `api/` for all the API routes.

bbangert: I think the only thing I'd change is define more concretely what the object is composed of. I don't entirely understand why its not certain where the `package.json` is. Because otherwise this seems like it should always return an object with a `commit` , `source`, and `version` property, no?

jrgm: I moved verison.ts from ./lib to ./lib/api

jrgm: > I don't entirely understand why its not certain where the package.json is

So the tests use `ts-node` and to  the typescript, so the relative positioning is `../../package.json` (or `../../version.json`).

```
`-- package.json
`-- version.json
`-- lib/
    `-- api/
        `-- version.ts
```

When the compiled javascript is run, the relative positioning is `../../../package.json`.

```
`-- package.json
`-- version.json
`-- dist/
    `-- lib/
        `-- api/
            `-- version.js
```

Having this "maybe it's here, but maybe it's there" check is definitely ugly. Perhaps there is a more idiomatic Typescript way to do this. (I suppose if the typescript code was all under `./src` the relative depths would be the same, but rearranging this entire module for the sake of this change wouldn't be worth it.)

philbooth: Could we create a symlink in `dist/` to make the relative path consistent in both cases?

jrgm: A symlink could be a fix for the `package.json` case, although `version.json` doesn't exist in a local build and spat into the docker image as part of the docker build.

jrgm: > Because otherwise this seems like it should always return an object with a commit , source, and version property, no?

I pushed up a change to simplify and define a return type for `getVersion()`, if that looks right.

bbangert: I think ideally the source should move into a `src/` dir, which will make the relative path locations consistent. We can do that later though.

PR #2552, opened 2019-09-20
fix(oauth): Ignore single-use scopes in the refresh_token grant.

### LINKED ISSUE #2548: Don't generate id_tokens in a refresh token grant
OIDC `id_tokens` are supposed to be a one-time communication of claims about a user, so IMHO no good can come from returning a new `id_token` when `grant_type=refresh_token`.  The OIDC spec explicitly permits that the response "may not contain an `id_token`" in this case, ref https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens.

Ref https://github.com/mozilla/fxa/pull/2497#discussion_r324959206

rfk:
Some OAuth scopes (such as "openid") are used to mediate a one-time
exchange of claims between the RP and FxA. While I don't believe any
harm can come from allowing these scopes to be used multiple times
via the refresh_token flow, I don't think any good can possibly come
from it either.

So, let's ignore such single-scope scopes in any refresh_token grants.

Fixes #2548.

### DIFF
--- a/packages/fxa-auth-server/test/remote/oauth_tests.js
+++ b/packages/fxa-auth-server/test/remote/oauth_tests.js
@@ -199,9 +199,8 @@ describe('/oauth/ routes', function() {
       refresh_token: res.refresh_token,
     });
     assert.ok(res2.access_token);
-    // NOTE - this assertion is commented out because of
-    // https://github.com/mozilla/fxa/issues/1867
-    //assert.equal(res2.scope, SCOPE);
+    assert.notOk(res2.id_token);
+    assert.equal(res2.scope, OAUTH_SCOPE_OLD_SYNC);
     assert.ok(res2.expires_in);
     assert.ok(res2.token_type);
     assert.notEqual(res.access_token, res2.access_token);
@@ -277,7 +276,6 @@ describe('/oauth/ routes', function() {
     assert.ok(tokenRes.auth_at);
     assert.ok(tokenRes.expires_in);
     assert.ok(tokenRes.token_type);
-
     const tokenJWT = decodeJWT(tokenRes.access_token);
     assert.ok(tokenJWT.claims.sub);
     assert.strictEqual(tokenJWT.claims.aud, JWT_ACCESS_TOKEN_CLIENT_ID);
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/test/remote/oauth_tests.js
```diff
@@ -191,6 +191,7 @@ describe('/oauth/ routes', function() {
       refresh_token: res.refresh_token,
     });
     assert.ok(res2.access_token);
+    assert.notOk(res2.id_token);
     // NOTE - this assertion is commented out because of
     // https://github.com/mozilla/fxa/issues/1867
     //assert.equal(res2.scope, SCOPE);
```
dannycoates: what can we say about `res2.scope` here, empty?

rfk: thanks, updated; it should contain just the "oldsync" scope, which is what will be on the issued token.

shane-tomlinson:
> I don't think any good can possibly come from it either.

Can you say more about this? The spec says the response "might not contain an id_token", however, it explicitly allow it and defines the rules that apply to the id_token when fetched using a refresh_token in [section 12.2](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse).

rfk:
I'll admit, this is mostly my opinion on the importance of understanding whether you want "authentication" or "authorization".  In my mental model of the system, refresh tokens are a long-lived authorization grant that says the holder is allowed to do specific things on the user's account, while id_tokens are a short-lived assertion about who the user is, used to prove that identity to an RP.

As an RP, if you're deliberately trying to get a fresh id_token, I reckon you're probably doing something that would be better served by some other mechanism (like an access_token, or doing a `prompt=none` flow). It seems like a footgun, and the presence of a bunch of rules in the spec about how to handle this case seems like an attempt to limit its damage.

This is a strong opinion weakly held though.

I guess one possible legitimate use would be: if you want to re-confirm user presence using the `prompt=none` flow, you could request a fresh `id_token` to include in the `id_token_hint` parameter, and it would save you the trouble of having to store the original `id_token` for that purpose.

(I think we should definitely ignore the session-token scope, even if we allow openid through).

shane-tomlinson:
> I'll admit, this is mostly my opinion on the importance of understanding whether you want "authentication" or "authorization". In my mental model of the system, refresh tokens are a long-lived authorization grant that says the holder is allowed to do specific things on the user's account, while id_tokens are a short-lived assertion about who the user is, used to prove that identity to an RP.

This is a good point, I'll offer a slight tweak to the argument. I've been looking at it as the openid scope authorizes the RP to learn about the user's authentication state on an ongoing basis, similar to prompt=none, but without user interaction.

I don't really have a concrete use case for this capability at the moment, perhaps it could be useful on the "RP knows who is signed in" experiment to make the solution more robust, but until we know we need that, I'm fine with being more restrictive.

PR #2553, opened 2019-09-20
feat(browserid): Assume all BrowserID client are `service=sync` for metrics purpoes

### LINKED ISSUE #2542: Can we safely assume that all `/certificate/sign` requests are coming from sync?
Now that it's possible to directly grant OAuth tokens using a sessionToken, I believe that the only clients calling our `/certificate/sign` endpoint in the wild will be browsers connected to sync.  Is that true?

According to [this amplitude chart](https://analytics.amplitude.com/mozilla-corp/chart/t6a0g50), the vast majority of `cert_signed` events come through with no `service` property.  I take these to be certificate-sign requests from Desktop which does not send an explicit `service=sync`.  There is a much smaller volume of `cert_signed` events that have `service=sync` set explicitly.  And there is an extremely low volume (single-digit occurrences per day) of `cert_signed` events where `service` is some OAuth client.

I'm not sure how to explain the background rate of non-sync certificate signings. Are they maybe users who are somehow still using an older version of the content-server code that uses older OAuth endpoint?  I wonder if we should error out in these cases because we really don't expect that to be happening anymore.

If it is in fact true that the only users of this endpoint are for sync, I'd like to update the [`account.signed` event](https://github.com/mozilla/fxa/blob/098fe0780059c44faa09f7b5f868624b36130a09/packages/fxa-auth-server/lib/routes/sign.js#L178) to include an explicit `service=sync` in all cases. This will help us with distinguishing users who have disabled sync in their browser after [the great decoupling](https://bugzilla.mozilla.org/show_bug.cgi?id=1570565).

rfk:
This is essentially the change I want to land pending discussion in https://github.com/mozilla/fxa/issues/2542, to help us identify `service=sync` in our server-metrics as we work on decoupling fxa and sync in Firefox Desktop.

Fixes #2542.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/sign.js
+++ b/packages/fxa-auth-server/lib/routes/sign.js
@@ -22,7 +22,7 @@ module.exports = (log, signer, db, domain, devices) => {
         },
         validate: {
           query: {
-            service: validators.service,
+            service: validators.service.valid('sync').optional(),
           },
           payload: {
             publicKey: isA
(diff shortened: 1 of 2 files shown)


rfk:
> Do we want the certs signed to get assertions for profile requests on Fx for iOS and Fennec
> to be considered Sync?

I think this will be OK for metrics purposes, since if those devices are fetching your profile, they're definitely also syncing.

PR #2558, opened 2019-09-20
fix(copy): Changed OTP code to verification code

ryanfeeley:
Using the term verification code to be more user friendly than acronym

### DIFF
--- a/packages/fxa-auth-server/lib/error.js
+++ b/packages/fxa-auth-server/lib/error.js
@@ -1205,7 +1205,7 @@ AppError.invalidOrExpiredOtpCode = () => {
     code: 400,
     error: 'Bad Request',
     errno: ERRNO.INVALID_EXPIRED_OTP_CODE,
-    message: 'Invalid or expired OTP code',
+    message: 'Invalid or expired verification code',
   });
 };
 


packages/fxa-auth-server/lib/error.js
```diff
@@ -1205,7 +1205,7 @@ AppError.invalidOrExpiredOtpCode = () => {
     code: 400,
     error: 'Bad Request',
     errno: ERRNO.INVALID_EXPIRED_OTP_CODE,
-    message: 'Invalid or expired OTP code',
+    message: 'Invalid or expired verification code',
```
vbudhram: This *should* have been handled in the content-server https://github.com/mozilla/fxa/blob/033507d45a9d4eea394006e2f1bc9377ae9ed004/packages/fxa-content-server/app/scripts/lib/auth-errors.js#L271, but it wasn't 😳. I'll investigate a little more why that isn't that case.

6a68: @vbudhram The PR queue is getting a bit out of hand, so I've filed https://github.com/mozilla/fxa/issues/2701 to follow up on your question.

shane-tomlinson: Changes to the auth-server error module does not affect what is displayed in the content server. For that to happen, we have to ensure the errno returned here has an entry in [app/scripts/lib/auth-errors.js](https://github.com/mozilla/fxa/blob/483f18b15b830b11f010a89b8e88c08031f40daa/packages/fxa-content-server/app/scripts/lib/auth-errors.js#L36). This translation layer ensures error strings are translated for the user's locale.

Issue #2540, opened 2019-09-18
Add a /oauth/destroy route


eoger:
This would allow us to remove the OAuth-server [client code](https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccountsOAuthGrantClient.jsm) entirely from Firefox.

rfk:
+1; this might also be a good opportunity to ensure compatibility with the [token revocation spec](https://tools.ietf.org/html/rfc7009) without breaking b/w compat with older clients.

shane-tomlinson:
Is this the auth server's equivalent of the [/v1/destroy route on the OAuth server](https://github.com/mozilla/fxa/blob/a142c2528ea83dfd2852e0e5fa9cd79e74193904/packages/fxa-auth-server/fxa-oauth-server/lib/routes/destroy.js)?

eoger:
> Is this the auth server's equivalent of the [/v1/destroy route on the OAuth server](https://github.com/mozilla/fxa/blob/a142c2528ea83dfd2852e0e5fa9cd79e74193904/packages/fxa-auth-server/fxa-oauth-server/lib/routes/destroy.js)?

Yup!

PR #2566, opened 2019-09-21
feat(oauth): Accept and validate client credentials when destroying a token.

rfk:
The OAuth /destroy endpoint currently does not require clients to authenticate
themselves, assuming that anyone holding a token should have the power to
destroy it. By contrast, the standard OAuth token revocation endpoint (RFC7009)
mandates that  clients authenticate themselves in order to use it.

This change makes our current /destroy endpoint *optionally* authenticate clients.
You don't have to provide credentials in order to call it, but if you *do* then
those credentials will be checked.  That's sufficient to let us build a new
endpoint that wraps our existing one and complies with RFC7009.

Connects to #2540.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/client.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/client.js
@@ -1,0 +1,147 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const crypto = require('crypto');
+const buf = require('buf').hex;
+const Joi = require('joi');
+
+const AppError = require('./error');
+const validators = require('./validators');
+const db = require('./db');
+const encrypt = require('./encrypt');
+const logger = require('./logging')('client');
+
+// Client credentials can be provided in either the Authorization header
+// or the request body, but not both.
+// These are some re-useable validators to assert that requirement.
+module.exports.clientAuthValidators = {
+  headers: Joi.object({
+    authorization: Joi.string()
+      .regex(validators.BASIC_AUTH_HEADER)
+      .optional(),
+  }).options({ allowUnknown: true }),
+
+  // The use of `$headers` here is Joi syntax for a "context reference"
+  // as described at https://hapi.dev/family/joi/?v=16.1.4#refkey-options.
+  // Hapi provides the headers as part of the context when validating request payload,
+  // as noted in https://github.com/hapijs/hapi/blob/master/API.md#-routeoptionsresponseschema.
+  clientId: validators.clientId.when('$headers.authorization', {
+    is: Joi.string().required(),
+    then: Joi.forbidden(),
+  }),
+
+  clientSecret: validators.clientSecret.when('$headers.authorization', {
+    is: Joi.string().required(),
+    then: Joi.forbidden(),
+  }),
+};
+
+/**
+ * Extract and normalize client credentials from a request.
+ * Clients may provide credentials in either the Authorization header
+ * or the request body, but not both.
+ *
+ * @param {Object} headers the headers from the request
+ * @param {Object} params the payload from the request
+ * @returns {Object} credentials
+ *   @param {String} client_id
+ *   @param {String} [client_secret]
+ */
+module.exports.getClientCredentials = function getClientCredentials(
+  headers,
+  params
+) {
+  const creds = {
+    client_id: params.client_id,
+    client_secret: params.client_secret,
+  };
+
+  // Clients are allowed to provide credentials in either
+  // the Authorization header or request body, but not both.
+  if (headers.authorization) {
+    const authzMatch = validators.BASIC_AUTH_HEADER.exec(headers.authorization);
+    const err = new AppError.invalidRequestParameter('authorization');
+    if (!authzMatch || creds.client_id || creds.client_secret) {
+      throw err;
+    }
+    const [clientId, clientSecret, ...rest] = Buffer.from(
+      authzMatch[1],
+      'base64'
+    )
+      .toString()
+      .split(':');
+    if (rest.length !== 0) {
+      throw err;
+    }
+    creds.client_id = Joi.attempt(clientId, validators.clientId, err);
+    creds.client_secret = Joi.attempt(
+      clientSecret,
+      validators.clientSecret,
+      err
+    );
+  }
+
+  return creds;
+};
+
+/**
+ * Authenticate a request made by an OAuth client, using credentials from
+ * either the Authorization header or request body parameters.
+ *
+ * @param {Object} headers the headers from the request
+ * @param {Object} params the payload from the request
+ * @returns {Promise} resolves with info about the client, or
+ *                    rejects if invalid credentials were provided
+ */
+module.exports.authenticateClient = async function authenticateClient(
+  headers,
+  params
+) {
+  const creds = exports.getClientCredentials(headers, params);
+
+  const client = await getClientById(creds.client_id);
+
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/client.js
```diff
@@ -0,0 +1,107 @@
...
+  }),
+};
+
+// Authenticate a request made by an OAuth client, using credentials from
+// either the Authorization header or request body parameters.
+//
+// If successful, returns an object with info about the client.
+// If unsuccessful, throws an error.
+//
+module.exports.authenticateClient = async function authenticateClient(
```
rfk: Previously we only needed this logic in `/token`, but it's now moved to a shared helper.

packages/fxa-auth-server/fxa-oauth-server/test/routes/token.js
```diff
@@ -21,7 +21,8 @@ const CLIENT_SECRET =
 const CLIENT_ID = '98e6508e88680e1b';
 const CODE = 'df6dcfe7bf6b54a65db5742cbcdce5c0a84a5da81a0bb6bdf5fc793eef041fc6';
 const PKCE_CODE_VERIFIER = 'au3dqDz2dOB0_vSikXCUf4S8Gc-37dL-F7sGxtxpR3R';
-const DISABLED_CLIENT_ID = 'd15ab1edd15ab1ed';
+const DISABLED_CLIENT_ID = '38a6b9b3a65a1871';
+const NON_DISABLED_CLIENT_ID = '98e6508e88680e1a';
```
rfk: Thanks to the refactoring of the `authenticateClient` function, we now have to use real client ids in this test.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/destroy.js
```diff
@@ -2,21 +2,28 @@
...
 /*jshint camelcase: false*/
 
 module.exports = {
   validate: {
+    headers: clientAuthValidators.headers,
     payload: Joi.object()
       .keys({
-        client_secret: Joi.string().allow(''),
+        client_id: clientAuthValidators.clientId.optional(),
+        // For historical reasons, we accept and ignore a client_secret if one
```
shane-tomlinson: Do we know whether this still happens and if so, which services are sending a client_secret without a client_id?

packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
```diff
@@ -185,89 +171,25 @@ module.exports = {
...
-        throw err;
-      }
-      params.client_id = Joi.attempt(creds[0], validators.clientId, err);
-      params.client_secret = Joi.attempt(
-        creds[1],
-        validators.clientSecret,
-        err
-      );
-    }
+    const client = await authenticateClient(req.headers, params);
```
shane-tomlinson: This is *much* cleaner. 👍

packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
```diff
@@ -368,7 +290,7 @@ async function validateRefreshTokenGrant(client, params) {
   // Does it belong to this client?
   if (!crypto.timingSafeEqual(tokObj.clientId, client.id)) {
     logger.debug('refresh_token.mismatch', {
-      client: params.client_id,
+      client: hex(client.id),
```
shane-tomlinson: This this just to work from the canonical source of truth rather than the params?

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -3588,14 +3588,14 @@ describe('/v1', function() {
         });
     });
 
-    it('should accept client_secret', function() {
+    it('should accept and ignore client_secret without client_id, for historical reasons', function() {
```
shane-tomlinson: Can you link to the PR with the explanation here?

packages/fxa-auth-server/fxa-oauth-server/test/api.js
```diff
@@ -3620,6 +3621,75 @@ describe('/v1', function() {
...
+          client_id: clientId,
+          client_secret: badSecret,
+          token: tok.access_token,
+        },
+      });
+      assert.equal(res.statusCode, 400);
+      assert.equal(res.result.errno, 102);
+    });
+
+    it('should succeed if valid client credentials are provided in request body', async () => {
```
shane-tomlinson: Is there a test that it should fail if both a valid `authorization` header and client credentials are sent?

packages/fxa-auth-server/fxa-oauth-server/test/routes/token.js
```diff
@@ -230,28 +228,33 @@ describe('/token POST', function() {
     });
 
     it('allows clients that have not been disabled via config', async () => {
+      request.payload.client_id = NON_DISABLED_CLIENT_ID;
```
shane-tomlinson: Does the description of this test match what is being tested? It seems more like "it fails if grant_type does not match passed in grant" or something similar.

packages/fxa-auth-server/fxa-oauth-server/test/routes/token.js
```diff
@@ -230,28 +228,33 @@ describe('/token POST', function() {
...
     it('allows code grants for clients that have been disabled via config', async () => {
       request.payload.client_id = DISABLED_CLIENT_ID;
       request.payload.grant_type = 'authorization_code';
+      request.payload.code = CODE;
       try {
         await route.handler(request);
         assert.fail('should have errored');
       } catch (err) {
-        assert.equal(err.errno, 101); // Unknown client.
+        assert.equal(err.errno, 105); // Unknown code.
```
shane-tomlinson: From the description I would have expected this test to pass. I know what you are trying to test, but the description and what's being tested don't really match.

packages/fxa-auth-server/fxa-oauth-server/lib/client.js
```diff
@@ -0,0 +1,130 @@
...
+// or the request body, but not both.
+// These are some re-useable validators to assert that requirement.
+module.exports.clientAuthValidators = {
+  headers: Joi.object({
+    authorization: Joi.string()
+      .regex(validators.BASIC_AUTH_HEADER)
+      .optional(),
+  }).options({ allowUnknown: true }),
+
+  clientId: validators.clientId.when('$headers.authorization', {
```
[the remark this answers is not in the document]
rfk: This is a "context ref" to some context provided automagically by Hapi; https://hapi.dev/family/joi/?v=16.1.4#refkey-options has some commentary on the `$` syntax.  I'll add a note in the comments.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/destroy.js
rfk: I do not, and I suspect none are. I'll add a log line while I'm in here so we can evaluate.

packages/fxa-auth-server/fxa-oauth-server/lib/routes/token.js
rfk: Yes; previously we mutated `params` in-place when extracting the credentials, which felt a bit off after moving it into a separate function.

packages/fxa-auth-server/fxa-oauth-server/test/api.js
rfk: Good idea, added.

packages/fxa-auth-server/fxa-oauth-server/test/routes/token.js
rfk: It does, but it's testing it in a kind of weird way in order to minimize setup; I'll add some explanation.

PR #2567, opened 2019-09-21
Add RFC7009-compatible token revocation endpoint.

### LINKED ISSUE #2540: Add a /oauth/destroy route

rfk:
Fixes #2540. Depends on #2566.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth.js
+++ b/packages/fxa-auth-server/lib/routes/oauth.js
@@ -17,6 +17,10 @@
 
 const Joi = require('joi');
 const validators = require('../routes/validators');
+const {
+  clientAuthValidators,
+  getClientCredentials,
+} = require('../../fxa-oauth-server/lib/client');
 
 const error = require('../error');
 const oauthRouteUtils = require('./utils/oauth');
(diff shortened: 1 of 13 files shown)


rfk:
@eoger I need to add some tests for this, but could you please take a look and see if it'll meet your needs for #2540?  The API is a little bit differently shaped than the one on the oauth-server but I'm hoping it's a simple transition for the client code.  You will need to pass `token` and `client_id` in the request body.

eoger:
This will work for me, thanks @rfk!
You mentioned in the other PR it is an authenticated endpoint, does this mean it is hawk-authenticated with the session token?

rfk:
> You mentioned in the other PR it is an authenticated endpoint, does this mean it is
> hawk-authenticated with the session token?

No. Since Firefox is a public client, "authenticated" in this case means only that you need to provide your `client_id` to tell us who you are. It's more relevant for private clients like Monitor who would also need to provide their `client_secret`.

rfk:
Alrighty, I think this is ready.  Please note that it includes the commit from https://github.com/mozilla/fxa/pull/2566 and so should be reviewed after that has merged.

rfk:
I should add, this probably conflicts with https://github.com/mozilla/fxa/pull/2497, but from what I've seen in that PR it'll hopefully be a fairly mechanical merge.  @dannycoates is the merging of https://github.com/mozilla/fxa/pull/2497 imminent enough that I should wait and rebase this on top of it?

packages/fxa-auth-server/lib/routes/oauth.js
```diff
@@ -215,6 +219,56 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+        },
+        response: {},
+      },
+      handler: async function(request) {
+        // This endpoint implements the API for token revocation from RFC7009,
+        // which says that if we can't find the token using the provided token_type_hint
+        // then we MUST search other possible types of token as well. So really
+        // token_type_hint just tells us what *order* to try different token types in.
+        let methodsToTry = ['revokeAccessToken', 'revokeRefreshToken'];
+        if (request.payload.token_type_hint === 'refresh_token') {
```
shane-tomlinson: If the passed token is a JWT, does it make sense to just skip checking for a refreshToken? Of course that makes the assumption that refresh tokens will never be JWTs...

packages/fxa-auth-server/test/local/oauthdb/revoke-refresh-token.js
```diff
@@ -21,6 +22,11 @@ const mockOAuthServer = nock(mockConfig.oauth.url).defaultReplyHeaders({
   'Content-Type': 'application/json',
 });
 
+const CLIENT_ID = '0123456789ABCDEF';
+const CLIENT_SECRET = '00001111222233334444555566667777';
+const REFRESH_TOKEN =
+  'DEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF';
```
shane-tomlinson: No actual cows were hurt in this test. 🐮

packages/fxa-auth-server/test/remote/oauth_tests.js
```diff
@@ -312,4 +312,52 @@ describe('/oauth/ routes', function() {
...
+
+    res = await client.grantOAuthTokens({
+      client_id: PUBLIC_CLIENT_ID,
+      code: res.code,
+      code_verifier: MOCK_CODE_VERIFIER,
+    });
+    assert.ok(res.access_token);
+    assert.ok(res.refresh_token);
+
+    await client.revokeOAuthToken({
```
shane-tomlinson: You should be able to call [`/introspect`](https://github.com/mozilla/fxa/blob/86daadeb8bba29f994291f7829f9d59952166701/packages/fxa-auth-server/fxa-oauth-server/lib/routes/introspect.js#L22) with this access token to see if has actually been invalidated.

packages/fxa-auth-server/test/remote/oauth_tests.js
```diff
@@ -312,4 +312,52 @@ describe('/oauth/ routes', function() {
...
+      client_id: PUBLIC_CLIENT_ID,
+      token: res.access_token,
+    });
```
shane-tomlinson: Is it possible to add a test for a JWT access token? Needed?

packages/fxa-auth-server/lib/routes/oauth.js
rfk: I think this is reasonable, and we can make it more generic by checking against the validators rather than hard-coding a check for whether it's a JWT.

packages/fxa-auth-server/test/remote/oauth_tests.js
rfk: Added, and in fact this uncovered an unhandled error case - if you pass in an invalid JWT access token, we would try to destroy it as a refreshToken and would fail internal route validation.

rfk:
I think I've addressed all the other concerns here, but I'm not sure what to do about the CORS issue; AFAICT I haven't done anything differently in this route versus the other OAuth routes in that regard.

rfk:
>  but I'm not sure what to do about the CORS issue

Aha, got it! The new header validation was causing it to strip all headers besides `authorization`, so the CORS logic was not able to echo the `Origin` header in the response.  Adding an explicit `stripUnknown: false` seems to have fixed the issue.

dannycoates:
>  it'll hopefully be a fairly mechanical merge.

the code is but the tests not so much 🍅

PR #2570, opened 2019-09-23
feat(keys): Explicitly track timestamp of last key rotation.

### LINKED ISSUE #490: Explicitly track the key-rotation timestamp for kB
When dealing with scoped keys [1] we want to be able to form a timestamped key identifier that looks like this:

   `"<timestamp when the key changed>-<hash derived from the key material>"`

Unfortunately we don't currently track the timestamp at which a user's `kB` changed.  We track a `veriferSetAt` timestamp and echo it out to reliers as the `fxa-generation` number.  But this timestamp can increase both on both password change (which preserves kB) and password reset (which might changes kB, or might not if a recovery key is used).

Let's track a  separate `keysChangedAt` timestamp that is only incremented when the user's kB actually changes, so we can use that to build key ids in the future.  We'll need to:

* [ ] make a db migration to store an additional column on the `accounts` table
* [ ] add logic for updating this column when `kB` changes, and preserving it when it doesnt
* [ ] add a `"fxa-keysChangedAt"` timestamp to the metadata in our identity assertions, alongside `"fxa-generation"`.
* [ ] update fxa-oauth-server to use the new `"fxa-keysChangedAt"` field when it informs fxa-content-server of the key rotation timestamp, falling back to `"fxa-generation"` if it's not present.

[1] https://docs.google.com/document/d/1IvQJFEBFz0PnL4uVlIvt8fBS_IPwSK-avK0BRIHucxQ/edit#

rfk:
Up until now we have been using `verifierSetAt` for the timestamp component of OAuth scoped-key key ids. This works but means that we generate a new kid whenever the user changes their password, even if that did not involve resetting the master encryption key `kB`.

With this change, we track a separate timestamp that is only bumped when kB actually changes, which should produce more stable key ids while preserving the ability to quickly compare them to determine which is the most recent.

Fixes #490.

Key ideas:

* From previous PRs, the DB contains a `keysChangedAt` field on the `accounts` table, and we report this value out to consumers in OAuth key ids and in BrowserID assertions. Since nothing has been writing to that field yet, it has been defaulting to the value of `verifierSetAt`.
* `db.resetAccount` now takes a boolean `keysHaveChanged` argument to indicate whether or not the value of `kB` has changed; we can't tell this by inspecting the other arguments, because encryption.
* The value of `keysHaveChanged` causes the db to either bump the `keysChangedAt` timestamp, or set it explicitly to its previous value.
* Thus, it is now either the value that we were previously reporting before this change went live, or it is a more recent value corresponding to an actual key-change event.

Safe deployment of this change depends on prep work done in previous commits. We need to be sure that sync clients all see a consistent key-rotation timestamp regardless of whether they are using BrowserID or OAuth, regardless of whether they hit a server running the old code or the new code, and in the face of a rollback. This is assured as follows:

* As of #3060, both BrowserID and OAuth clients will report the value from the `keysChangedAt` column if it exists, or fallback to using `verifierSetAt` if it is NULL. So when the new code starts writing to that column, the old code will use it. If we rollback the old code will keep using values written into that column.
* As of #2571, users who reset their password on a server running the old code will have `keysChangedAt` reset to `NULL`. In this case both new and old codebases will fall back to using  `verifierSetAt` as described above.

I'm happy with the code here, but I'm leaving it as "draft" until:

* [x] We have deployed https://github.com/mozilla/fxa/pull/3060 to production.
* [x] We have [deployed an update to tokenserver](https://bugzilla.mozilla.org/show_bug.cgi?id=1589565) to stop it from assuming that OAuth key ids contain the generation number.
* [x] We have re-targetted this PR against `master` rather than against the branch for https://github.com/mozilla/fxa/pull/3060

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/test/backend/remote.js
+++ b/packages/fxa-auth-db-mysql/db-server/test/backend/remote.js
@@ -141,176 +141,163 @@ module.exports = function(cfg, makeServer) {
       );
     });
 
-    it('add account, add email, get secondary email, get emails, delete email', () => {
-      var user = fake.newUserDataHex();
-      var secondEmailRecord = user.email;
-      var thirdEmailRecord;
+    it('add account, add email, get secondary email, get emails, delete email', async () => {
+      const user = fake.newUserDataHex();
+      const secondEmailRecord = user.email;
 
-      return client
-        .putThen('/account/' + user.accountId, user.account)
-        .then(function(r) {
-          respOkEmpty(r);
-          return client.postThen(
-            '/account/' + user.accountId + '/emails',
-            user.email
-          );
-        })
-        .then(function(r) {
-          respOkEmpty(r);
-          return client.getThen('/account/' + user.accountId + '/emails');
-        })
-        .then(function(r) {
-          respOk(r);
+      let r = await client.putThen('/account/' + user.accountId, user.account);
+      respOkEmpty(r);
 
-          var result = r.obj;
-          assert.lengthOf(result, 2);
+      r = await client.postThen(
+        '/account/' + user.accountId + '/emails',
+        user.email
+      );
+      respOkEmpty(r);
 
-          // Verify first email is email from accounts table
-          assert.equal(
-            result[0].email,
-            user.account.email,
-            'matches account email'
-          );
-          assert.equal(
-            !!result[0].isPrimary,
-            true,
-            'isPrimary is true on account email'
-          );
-          assert.equal(
-            !!result[0].isVerified,
-            !!user.account.emailVerified,
-            'matches account emailVerified'
-          );
+      r = await client.getThen('/account/' + user.accountId + '/emails');
+      respOk(r);
+      let result = r.obj;
+      assert.lengthOf(result, 2);
 
-          // Verify second email is from emails table
-          assert.equal(
-            result[1].email,
-            secondEmailRecord.email,
-            'matches secondEmail email'
-          );
-          assert.equal(
-            !!result[1].isPrimary,
-            false,
-            'isPrimary is false on secondEmail email'
-          );
-          assert.equal(
-            !!result[1].isVerified,
-            false,
-            'matches secondEmail isVerified'
-          );
+      // Verify first email is the primary email
+      assert.equal(
+        result[0].email,
+        user.account.email,
+        'matches account email'
+      );
+      assert.equal(
+        !!result[0].isPrimary,
+        true,
+        'isPrimary is true on account email'
+      );
+      assert.equal(
+        !!result[0].isVerified,
+        !!user.account.emailVerified,
+        'matches account emailVerified'
+      );
 
-          var emailCodeHex = secondEmailRecord.emailCode.toString('hex');
-          return client.postThen(
-            '/account/' + user.accountId + '/verifyEmail/' + emailCodeHex
-          );
-        })
-        .then(function(r) {
-          respOkEmpty(r);
-          return client.getThen('/account/' + user.accountId + '/emails');
-        })
-        .then(function(r) {
-          respOk(r);
+      // Verify second email is the secondary email
+      assert.equal(
+        result[1].email,
+        secondEmailRecord.email,
+        'matches secondEmail email'
+      );
+      assert.equal(
+        !!result[1].isPrimary,
+        false,
+        'isPrimary is false on secondEmail email'
+      );
+      assert.equal(
+        !!result[1].isVerified,
+        false,
+        'matches secondEmail isVerified'
+      );
 
-          var result = r.obj;
-          assert.lengthOf(result, 2);
-          assert.equal(
-            result[1].email,
-            secondEmailRecord.email,
-            'matches secondEmail email'
-          );
-          assert.equal(
-            !!result[1].isPrimary,
-            false,
-            'isPrimary is false on secondEmail email'
-          );
-          assert.equal(
-            !!result[1].isVerified,
-            true,
-            'matches secondEmail isVerified'
-          );
+      var emailCodeHex = secondEmailRecord.emailCode.toString('hex');
+      r = await client.postThen(
+        '/account/' + user.accountId + '/verifyEmail/' + emailCodeHex
+      );
+      respOkEmpty(r);
 
-          thirdEmailRecord = fake.newUserDataHex().email;
-          return client.postThen(
-            '/account/' + user.accountId + '/emails',
-            thirdEmailRecord
-          );
-        })
-        .then(function(r) {
-          respOkEmpty(r);
-          return client.getThen('/account/' + user.accountId + '/emails');
-        })
-        .then(function(r) {
-          respOk(r);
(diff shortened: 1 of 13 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1173,70 +1173,49 @@ module.exports = (
...
-                  uid: account.uid,
-                  generation: account.verifierSetAt,
-                }),
-                customs.reset(account.email),
-              ]);
-            });
-        }
-
-        function setupKb() {
+        async function resetAccountData() {
```
rfk: I took the liberty of async-awaitify this while I was here, because it really needed it.

packages/fxa-auth-db-mysql/lib/db/mem.js
```diff
@@ -97,7 +97,7 @@ module.exports = function(log, error) {
       return P.reject(error.duplicate());
     }
 
-    accounts[uid.toString('hex')] = data;
+    accounts[uid.toString('hex')] = extend({}, data);
```
rfk: Previously, we'd just store whatever object the calling code gave us, meaning that it could change unexpectedly.

packages/fxa-auth-server/test/remote/account_reset_tests.js
```diff
@@ -21,239 +22,194 @@ describe('remote account reset', function() {
     });
   });
 
-  it('account reset w/o sessionToken', () => {
+  it('account reset w/o sessionToken', async () => {
```
rfk: I async-awaitified these tests while I was in here, but I think in retrospect it generated a bit much churn. Sorry for the added noise to this PR, I refrained from doing any of the other test files.

packages/fxa-auth-server/test/remote/account_reset_tests.js
```diff
@@ -21,239 +22,194 @@ describe('remote account reset', function() {
...
+    }
+  });
 
-        return client.forgotPassword();
-      })
-      .then(() => server.mailbox.waitForCode(email))
-      .then(code => {
-        // Stash original reset code then attempt to use it after another reset
-        originalCode = code;
+  it('account reset updates keysChangedAt', async () => {
```
rfk: This is the only new test in this file, the others were just refactored to async/await.

packages/fxa-auth-server/lib/routes/account.js
6a68: :100: huge improvement

6a68: :100: huge improvement

rfk:
> I suspect we may need to have the corresponding sync tokenserver change in production before
> deploying this, but I'll dig into that more tomorrow.

I don't believe we need to full tokenserver change in place, but we do need to have [this smaller change](https://github.com/mozilla/fxa/pull/2570) deployed to production, otherwise we risk locking out OAuth sync clients.  I've filed [Bug 1589565](https://bugzilla.mozilla.org/show_bug.cgi?id=1589565) to track its deployment.

rfk:
> I don't believe we need to full tokenserver change in place, but we do need to have this smaller
> change deployed to production, otherwise we risk locking out OAuth sync clients. I've filed Bug
> 1589565 to track its deployment.

As noted in #3060, I've pulled the bulk of this change out into a separate PR that I believe is safe to deploy without waiting on the above tokenserver change. Having that in production separately will also help smooth out the deploy of the remaining changes, for reasons which I will outline in an updated issue description above.

rfk:
> We have re-targetted this PR against master rather than against the branch for #3060

This is done, including a rebase against master with #3060 merged.

packages/fxa-auth-db-mysql/db-server/test/backend/remote.js
```diff
@@ -141,176 +141,163 @@ module.exports = function(cfg, makeServer) {
       );
     });
 
-    it('add account, add email, get secondary email, get emails, delete email', () => {
-      var user = fake.newUserDataHex();
-      var secondEmailRecord = user.email;
-      var thirdEmailRecord;
+    it('add account, add email, get secondary email, get emails, delete email', async () => {
```
rfk: Huh, I thought I left a comment about this test update, but it doesn't seem to be here.

So, this test started failing for me in CI, and I tracked it down to an issue with determinism - this PR seems to somehow perturb the order in which secondary emails are returned by the db, and this test was depending on them being in a specific order.  The only functional change is that I've added a check to put the results in a consistent order, but in the process of debugging this test I asyncified it, hence the churny diff.

packages/fxa-auth-db-mysql/db-server/test/backend/remote.js
```diff
@@ -141,176 +141,163 @@ module.exports = function(cfg, makeServer) {
...
+        thirdEmailRecord
+      );
+      respOkEmpty(r);
+
+      r = await client.getThen('/account/' + user.accountId + '/emails');
+      respOk(r);
+
+      result = r.obj;
+      assert.lengthOf(result, 3);
+      // Secondary emails are not returned in a deterministic order; normalize.
```
rfk: This is the important part in getting the test to pass reliably.

shane-tomlinson:
@rfk - is there more you want to do here before changing this to ready for review?

rfk:
> is there more you want to do here before changing this to ready for review?

Nope, I'm just keeping it in draft so we don't merge it until the deploys mentioned in the issue description are complete and stable. Please feel free to review in the meantime :-)

rfk:
@vbudhram while this is still draft because it's waiting on some deployments, I think the code here is ready for review - if you have a chance to take a look so it can be "ready to roll" once the deployments are complete, that would be very awesome :-)

clouserw:
I'm going to kick this to the next milestone.  Is this ready to land in the next 2 weeks?  Thanks

rfk:
> Is this ready to land in the next 2 weeks?

I'm told the tokenserver deploy in https://bugzilla.mozilla.org/show_bug.cgi?id=1589565 is going out today, and all other operational blockers have been resolved. Pending an r+ here it's ready to rock and roll.

rfk:
I'm hitting "ready for review" because the code here really is ready and has been for some time, but please don't merge until we confirm that the tokenserver deploy is stable in production.

chelsea-lewis:
@rfk @clouserw Do we know if it's possible to release this work with 150, or in a dot release following? (I'm not seeing the dot release as an opportunity due to auth server work, but may be wrong on what I'm seeing here).

The Durable Sync work is blocked until this is released to Prod, so any moves to get this to production sooner would be beneficial to their timelines.

rfk:
I think a 150 dot-release would be reasonable FWIW, and am happy to do the necessary git-wrangling (pending the deployment in https://bugzilla.mozilla.org/show_bug.cgi?id=1589565).

rfk:
> (pending the deployment in https://bugzilla.mozilla.org/show_bug.cgi?id=1589565).

This is done now, but I'm going to give it 24-hours just in case of any need to rollback.

PR #2571, opened 2019-09-23
fix(db): Reset `keysChangedAt` to NULL if we don't know its correct value

rfk:
I started working on #490 to track the timestamp at which the user's keys were last changed, and I realized that there could be a subtle issue if we deployed that change and then had to roll it back.  Details in comments in the code below, but I think if we get this small change deployed first, then rolling out the bigger change for #490 will be safer.

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-104-105.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-104-105.sql
@@ -1,0 +1,58 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('104');
+
+CREATE PROCEDURE `resetAccount_14` (
+  IN `uidArg` BINARY(16),
+  IN `verifyHashArg` BINARY(32),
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-db-mysql/lib/db/schema/patch-104-105.sql
```diff
@@ -0,0 +1,58 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('104');
+
+CREATE PROCEDURE `resetAccount_14` (
```
philbooth: Is it deliberate that you didn't update the call in `lib/db/mysql.js`?

philbooth: Nvm, just noticed you didn't actually ask for review yet, sorry!

rfk: it is not; thanks!

rfk:
Thanks @philbooth, I think this should be ready for r? now if you're happy to take a look.

Issue #2581, opened 2019-09-23
Add an /__lbheartbeat__ endpoint


clouserw:
This should be the same as /__heartbeat__ but not hit any external services.  The LBs will use this to check if the service is responding properly.  /cc @jrgm

PR #2588, opened 2019-09-23
fix(oauth): /__lbheartbeat__ route; no connect to database

jrgm:
`/__lbheartbeat__` already existed, running the same `db.ping()` as `/__heartbeat__`. This just changes it to return `{}`; previous route test for `/__lbheartbeat__` kept as is.

r? @mozilla/fxa-devs


jrgm:
Fixes #2581

PR #2592, opened 2019-09-24
fix(dataflow): ensure old-format messages don't clog the queue

philbooth:
Related to #2445.

The customs dataflow integration re-uses a pubsub queue that used to contain messages in a different format. As written, the error-handling code would keep pushing those old messages back onto the queue. This change fixes it by calling `message.ack()` instead of `message.nack()` on any old messages that can't be parsed.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-customs-server/lib/dataflow.js
+++ b/packages/fxa-customs-server/lib/dataflow.js
@@ -98,14 +98,31 @@ module.exports = (config, log, fetchRecords, setRecords) => {
         reportOnly,
       });
     } catch (error) {
+      let op;
+
+      if (isFresh(message.publishTime)) {
+        op = 'dataflow.message.error';
+
+        // Call `message.nack` to put the message back on the queue in case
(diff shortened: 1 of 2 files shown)


packages/fxa-customs-server/lib/dataflow.js
```diff
@@ -98,14 +98,31 @@ module.exports = (config, log, fetchRecords, setRecords) => {
...
+        op = 'dataflow.message.error';
+
+        // Call `message.nack` to put the message back on the queue in case
+        // it's a transient error. If the error persists, eventually the
+        // subsequent condition will be entered instead.
+        message.nack();
+      } else {
+        op = 'dataflow.message.error-old';
+
+        // The queue previously carried messages in a different format.
```
shane-tomlinson: What's the rollout plan, are we going to set `ignoreOlderThan` to some low number to begin with and gradually increase it as messages in the old format don't overwhelm the queue? Any idea on how many of these messages we could expect?

Issue #997, opened 2019-05-01
Ensure we're monitoring backend performance data for new Subscriptions service


clouserw:
**2019-09-30 EDIT (@chenba)** We've settled on using `statsd` (with the `hot-shots` npm module) to send stats and the InfluxDB stack to collect and query our data.

The following PRs add `statsd` timing calls and relevant configs:

- [x] https://github.com/mozilla/fxa/pull/2594
- [x] https://github.com/mozilla/fxa/pull/2631
- [x] https://github.com/mozilla/fxa/pull/2616
- [x] https://github.com/mozilla/fxa/pull/2652

TODOs before we can start sending stats from FxA:

- [ ] [Deploy Telegraf to collect metrics](https://github.com/mozilla/fxa/issues/2665)
[...]

PR #2594, opened 2019-09-24
feat(metrics): collect perf metrics with StatsD on auth server's HTTP API clients

chenba:
Part of #997 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -11,9 +11,17 @@ require('../lib/newrelic')();
 
 const jwtool = require('fxa-jwtool');
 const P = require('../lib/promise');
+const StatsD = require('hot-shots');
 
 function run(config) {
   const log = require('../lib/log')(config.log);
+  const statsd = new StatsD({
+    ...config.statsd,
+    errorHandler: err => {
+      // eslint-disable-next-line no-use-before-define
+      log.error('statsd.error', err);
+    },
+  });
   const getGeoData = require('../lib/geodb')(log);
   // Force the geo to load and run at startup, not waiting for it to run on
   // some route later.
@@ -54,7 +62,7 @@ function run(config) {
       : null,
   };
 
-  const Customs = require('../lib/customs')(log, error);
+  const Customs = require('../lib/customs')(log, error, statsd);
 
   const Server = require('../lib/server');
   let server = null;
@@ -82,8 +90,8 @@ function run(config) {
     .spread(
       (db, translator) => {
         database = db;
-        oauthdb = require('../lib/oauthdb')(log, config);
-        subhub = require('../lib/subhub/client')(log, config);
+        oauthdb = require('../lib/oauthdb')(log, config, statsd);
+        subhub = require('../lib/subhub/client')(log, config, statsd);
 
         return require('../lib/senders')(
           log,
@@ -106,7 +114,8 @@ function run(config) {
             config,
             customs,
             zendeskClient,
-            subhub
+            subhub,
+            statsd
           );
 
           statsInterval = setInterval(logStatInfo, 15000);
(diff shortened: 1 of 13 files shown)


packages/fxa-auth-server/config/index.js
```diff
@@ -742,6 +742,26 @@ const conf = convict({
       env: 'FLOW_ID_EXPIRY',
     },
   },
+  statsd: {
+    host: {
+      default: 'localhost',
```
lmorchard: It doesn't look like there's a statsd server running at port 8125 in `servers.json` for pm2. Should we add one?

packages/fxa-auth-server/lib/backendService.js
```diff
@@ -217,6 +224,12 @@ module.exports = function createBackendServiceAPI(
         payload,
         this._headers
       );
+
+      // The statsD dependency is optional
+      if (statsd) {
+        statsd.timing(`${serviceName}.${methodName}`, Date.now() - startTime);
```
lmorchard: Haven't run this code yet, but does this fail gracefully when the statsd server can't be reached?

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -11,9 +11,11 @@ require('../lib/newrelic')();
 
 const jwtool = require('fxa-jwtool');
 const P = require('../lib/promise');
+const StatsD = require('hot-shots');
 
 function run(config) {
   const log = require('../lib/log')(config.log);
+  const statsd = new StatsD(config.statsd);
```
lmorchard: Peeking at the docs, there appears to be [an `errorHandler` that we can set up here](https://www.npmjs.com/package/hot-shots#errors) and trap all errors. Maybe we want to do that here, rather than have the errors bubble up from inside service code?

chenba: Yeah.  When I was tailing the log, I did not notice any errors when there's nothing listening on the port.  (It's UDP so that kind of make sense.)  But it does throw a DNS error if it cannot resolve the host from the config.

packages/fxa-auth-server/config/index.js
chenba: `nc -lvu 127.0.0.1 8125` was the extent of me verifying the output. :D  I don't know if a server would add much value since we don't do anything with those stats locally.  Unless you want something more than just printing out the stats or even an actual telegraf server to ingest the data?

packages/fxa-auth-server/lib/backendService.js
chenba: Yeah, I didn't see anything in the log.

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -11,9 +11,17 @@ require('../lib/newrelic')();
...
 
 function run(config) {
   const log = require('../lib/log')(config.log);
+  const statsd = new StatsD({
+    ...config.statsd,
+    errorHandler: err => {
+      // eslint-disable-next-line no-use-before-define
+      log.error('statsd.error', err);
+    },
+  });
```
chenba: @lmorchard Added this to trap and log errors from statsd.

In a future patch the order the `log` and `statsd` above will be swapped so I can pass `statsd` into the `log` constructor.  That's the reason for the eslint comment I accidentally left in there. 😬

packages/fxa-auth-server/config/index.js
lmorchard: Yeah, probably not worth over-doing. I think failing gracefully and handling any potential errors are the more valuable things here.

packages/fxa-auth-server/lib/backendService.js
l-hedgehog: IIUC, this won't happen if an error is thrown from the `sendRequest` call above?

chenba: Yes.  I'll file a bug and create a patch.  Thanks!

PR #2597, opened 2019-09-24
fix(build): npm audit fix

dannycoates:


6a68:
This breakage is affecting the other PR I'm looking at, so I can add a commit to fix these issues

6a68:
For the fxa-customs-server changes, I was considering upgrading the stale affected packages, but we're several versions behind on `csv-parse` and a version behind on `restify`, and many of the unit tests are failing, so I don't trust them to verify the upgrade doesn't break anything. Opened #2599 to deal with the unit tests, and #2600 to upgrade those deps.

6a68:
Also filed #2601 for the hapi 18 upgrade needed to correctly resolve https://npmjs.com/advisories/1168.

Issue #1079, opened 2019-05-10
Improve TypeScript coverage of Redux store in fxa-payments-server


lmorchard:
I kind of fudged the TypeScript annotations in the Redux store code for fxa-payments-server with very broad types. It could be reworked to be much more specific - which would enable much better auto-complete in IDEs and validate & lint action names and parameters in code.

lmorchard:
FWIW, [I started tinkering](https://github.com/lmorchard/fxa/commit/6892067e6de88643b656a951ff61990555861e46) with reworking the Redux store toward resolving this issue. But, putting this work on pause for other priorities.

lmorchard:
To elaborate on an aspect that could be improved here:

Actions in the Redux store are [defined using `createActions()` from the redux-actions module](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/store/index.tsx#L72). That's [a utility](https://redux-actions.js.org/api/createaction#createactions) which does a lot of magical work to generate Redux action creators while reducing boilerplate. The problem, though, is that this magic doesn't play well with TypeScript and obscures types.

As a practical example - if you edit a React component that imports `actions` from the `store/index.tsx` module, there's no auto-complete of action names in VSCode. That's because the list of available actions are obscured on the way through `createActions` and I have not found a good set of type definitions that fixes this.

So, one solution is to create a new `actions.ts` module under `store` and split each of these actions out into a function explicitly exported from that module.

This might introduce the dreaded boilerplate, which could be okay as a cost of making the Redux store more self-documenting when used in other places in the app. Would be nice if we could keep the repeated code to a minimum while improving type coverage, though.

lmorchard:
There are three big niceties from `createAction` in `redux-actions`. 

* Giving every action creator function a `.toString()` method, so we don't need a separate list of `ACTION_TYPE` constants
* Building [Flux action objects](https://github.com/redux-utilities/flux-standard-action) with minimal boilerplate.
* Generating identity function action creators with just a list of names.

Consider this code:
```
const actions = createActions(
  { foo: param => `foo ${param}` },
  "bar"
);
```
This is semi-equivalent to:
```
const actions = {};

actions.foo = param => ({
  type: 'foo',
  payload: `foo ${param}`,
});

actions.foo.toString = () => 'foo';

actions.bar = param => ({
  type: 'bar',
  payload: param,
});

actions.bar.toString = () => 'bar';
```
[...]

PR #2598, opened 2019-09-24
 refactor(payments) fix #1591,#1079,#2745: Better organize Redux modules and API calls, remove cruft, fix console warnings

### LINKED ISSUE #1591: [fxa-payments-server] Refactor API calls in redux store into an API client class
This is a nice-to-have and not at all necessary for any launches.

Currently, most API calls made from the payments server app are done directly from Redux actions. Might be nicer to refactor these out into a separate module providing an API client class. Some suggested steps:

* Move [API calls](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/store/index.tsx#L78) into a new class `APIClient` defined in `src/lib/apiClient.js`
* Create an instance of `APIClient` in `src/index.tsx` with `config` and `accessToken`
* Pass the `APIClient` instance instead of `accessToken` via `src/lib/AppContext.tsx`
* Switch to using `APIClient` rather than `accessToken` in Redux actions

LZoog:
~draft~ fix for #1591, #1079, and #2745 

~@lmorchard, before I continue on typing `apiContext`, removing comments, running tests (I can sign up/cancel/resignup for subscriptions though) I wanted to make sure this is what you had in mind here. It feels kind odd passing around this object around so much since it's a singleton.~ This gets passed around like `config` does now.


packages/fxa-payments-server/src/App.tsx
```diff
@@ -55,7 +51,7 @@ export const App = ({
   return (
     <AppContext.Provider value={appContextValue}>
       <AppErrorBoundary>
-        <StripeProvider apiKey={config.stripe.apiKey}>
+        <StripeProvider apiKey={apiClient.config.stripe.apiKey}>
```
lmorchard: Hmm, I don't think we should move `config` into the new `apiClient`. There are things using `config` that shouldn't ever make API calls, and I think `apiClient` should mainly be used by just the Redux store right now.

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -0,0 +1,94 @@
+import { Config } from './config';
```
lmorchard: This might be a weird idea, but seeing how this is shaping up as an instance that we pass around everywhere... What if we made the `apiClient` module a more proper singleton - not unlike how [the `config` module](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/lib/config.ts) works now?

That is, rather than export a class to be instantiated given `config` and `accessToken` - make `config` and `accessToken` into module-local variables and export `updateConfig` & `updateAccessToken` functions along with all the API client functions. Then, we wouldn't need to pass an `apiClient` instance around, just import methods from the module

packages/fxa-payments-server/src/store/index.tsx
```diff
@@ -71,64 +66,43 @@ export const selectors: Selectors = {
...
-    fetchSubscriptions: (accessToken: string) =>
-      apiGet(
-        accessToken,
-        `${config.servers.auth.url}/v1/oauth/subscriptions/active`
-      ),
-    fetchToken: (accessToken: string) =>
-      apiPost(accessToken, `${config.servers.oauth.url}/v1/introspect`, {
-        token: accessToken,
+    fetchProfile: (apiClient: any) =>
+      apiClient.get(`${apiClient.config.servers.profile.url}/v1/profile`),
```
lmorchard: I'd also say all these paths should probably be refactored into the `apiClient` module too. Maybe export more descriptive functions like `getProfile()` & `getPlans()` to use here and we can manage these URLs in the `apiClient` if/when they change later.

lmorchard: And I've realized that this will strip these functions down further to something like `fetchProfile: () => apiClient.getProfile()` - but simplifying the action creators will probably also make weaning away from redux-actions easier?

packages/fxa-payments-server/src/lib/apiClient.ts
LZoog: I like that much better👍, the way I've got it now was definitely starting to feel off.

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -0,0 +1,159 @@
...
+      const { code, errno, error, message, statusCode } = this.body;
+      Object.assign(this, { code, errno, error, message, statusCode });
+    }
+  }
+}
+
+let accessToken = '';
+let config = defaultConfig();
+
+export function updateAPIClientConfig(configFromMeta: any) {
```
lmorchard: I think this is looking a lot better. Main two things are a) would be nice to cover this new extracted module in a unit test and b) I wonder if it might make sense to get even more specific with the config here - i.e. maybe have one function that accepts an accessToken and just config.servers since that's all we use here.

Kind of splitting hairs with the config thing, so I wouldn't prioritize or block on that

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -179,7 +179,6 @@ export const elementChangeResponse = ({
 });
```
lmorchard: I think I know why you're seeing connection refused errors - we need to set up mock config for the new singleton API client in test-utils. Otherwise, the API client will try to make real network requests instead of hitting the mock handlers in tests. These changes get most of the tests passing for me:
```diff
diff --git a/packages/fxa-payments-server/src/lib/test-utils.tsx b/packages/fxa-payments-server/src/lib/test-utils.tsx
index 68c2676c5..0d2ef5f8d 100644
--- a/packages/fxa-payments-server/src/lib/test-utils.tsx
+++ b/packages/fxa-payments-server/src/lib/test-utils.tsx
@@ -2,6 +2,7 @@ import React, { useContext, ReactNode } from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
 import { AppContext, AppContextType } from '../../src/lib/AppContext';
 import { config, updateConfig } from '../../src/lib/config';
+import { updateAPIClientToken, updateAPIClientConfig } from '../../src/lib/apiClient';
 import ScreenInfo from '../../src/lib/screen-info';
 import { ReactStripeElements } from 'react-stripe-elements';
 import nock from 'nock';
@@ -17,6 +18,8 @@ declare global {
   }
 }
 
+export const MOCK_ACCESS_TOKEN = '8675309jenny';
+
 export const wait = (delay: number) =>
   new Promise(
     resolve => setTimeout(resolve, delay)
@@ -79,6 +82,8 @@ export function mockOptionsResponses(baseUrl: string) {
 
 export const setupMockConfig = (config?: typeof mockConfig) => {
   updateConfig(config || mockConfig);
+  updateAPIClientConfig(config || mockConfig);
+  updateAPIClientToken(MOCK_ACCESS_TOKEN);
 };
 
 // Minimal mock for react-stripe-elements that lets us trigger onChange
```

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -38,8 +40,8 @@ export type ProductProps = {
   createSubscriptionStatus: CreateSubscriptionFetchState;
   plansByProductId: (id: string) => Array<Plan>;
   createSubscription: Function;
-  resetCreateSubscription: () => void;
-  resetCreateSubscriptionError: () => void;
+  resetCreateSubscription: any;
```
lmorchard: FWIW, using an `any` type is okay during development - but we should work to get more specific with types.

One thing you could do here is use `typeof resetCreateSubscription` to pull in the exact type of the action creator expected here, now that we can actually see that type. This will probably cause issues that need fixing elsewhere that rely on a more relaxed type, though.

packages/fxa-payments-server/src/store/selectors.ts
```diff
@@ -0,0 +1,41 @@
...
+export const reactivateSubscriptionStatus: Selector = state =>
+  state.api.reactivateSubscription;
+
+export const updatePaymentStatus: Selector = state => state.api.updatePayment;
+
+export const plansByProductId: Selector = state => (
+  productId: string
+): Array<Plan> => {
+  const fetchedPlans = plans(state).result || [];
+  return fetchedPlans.filter((plan: Plan) => plan.product_id === productId);
```
LZoog: This was:
```
    return productId
      ? fetchedPlans.filter((plan: Plan) => plan.product_id === productId)
      : fetchedPlans;
```
... but I took out the check for `productId` because `productId` is required (line 41 in `Product/index.tsx`) so we shouldn't need the ternary.

packages/fxa-payments-server/src/store/utils.tsx
```diff
@@ -13,91 +13,6 @@ export const mapToObject = (
...
-export const apiGet = (...args: [string, string, object?]) =>
-  apiFetch('GET', ...args);
-
-export const apiDelete = (...args: [string, string, object?]) =>
-  apiFetch('DELETE', ...args);
-
-export const apiPost = (accessToken: string, path: string, body: object) =>
-  apiFetch('POST', accessToken, path, { body: JSON.stringify(body) });
-
 export const setStatic = (newState: object) => (state: object): object => ({
```
LZoog: Jest has our branch test coverage @ 50% for `setFromPayLoad` (line 24) due to the return of `payload || defval` for `[name]`. I can write a test for this, but I don't see that we're using `setAsPayload`, `setFromPayload`, or `setFromPayLoadFn` anywhere. Can I remove these functions & their related tests?

packages/fxa-payments-server/src/components/LoadingOverlay.tsx
```diff
@@ -19,8 +15,3 @@ export const LoadingOverlay = ({ isLoading }: LoadingOverlayProps) =>
       <img alt="Loading" src={spinnerImage} />
     </div>
   );
-
-/* istanbul ignore next - skip testing redux connect() */
-export default connect((state: State) => ({
-  isLoading: selectors.isLoading(state),
-}))(LoadingOverlay);
```
LZoog: I removed this from our selectors, as it wasn't being hit:

```
isLoading = state => Object.values(state.api).some(v => v && !!v.loading);
```

It was showing as uncovered in tests, as we only ever directly pass in `isLoading={true}` or `isLoading={false}` such as the following in `Subscriptions/index.tsx`:

```
  if (
    customer.loading ||
    subscriptions.loading ||
    profile.loading ||
    plans.loading
  ) {
    return <LoadingOverlay isLoading={true} />;
  }
```

With our checks as they are, I don't think this needs to be connected to the store.

packages/fxa-payments-server/src/store/utils.tsx
lmorchard: Yeah, if this is dead code now, feel free to remove it! I think it might have been used in the react-actions code that you refactored away

packages/fxa-payments-server/src/components/LoadingOverlay.tsx
lmorchard: :+1: I think this might be a leftover from when we displayed a general loading indicator without checking the individual API loading states

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -0,0 +1,159 @@
...
+
+let accessToken = '';
+let config: Config = defaultConfig();
+
+export function updateAPIClientConfig(configFromMeta: any) {
+  config = configFromMeta;
+}
+
+export function updateAPIClientToken(token: string) {
+  accessToken = token;
```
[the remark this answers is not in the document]
lmorchard: I think it's fine to leave a few things untested here and kick the can down the road. Overall, test coverage looks slightly improved in this PR:

Capture

In comparison, master branch looks like this right now:

Capture

packages/fxa-payments-server/src/store/actions.ts
```diff
@@ -0,0 +1,90 @@
...
+  payload: apiUpdatePayment(paymentToken),
+});
+updatePayment.toString = () => 'updatePayment';
+
+export const resetCreateSubscription = (payload?: Payload) =>
+  makeActionObject('resetCreateSubscription', payload);
+resetCreateSubscription.toString = () => 'resetCreateSubscription';
+
+export const resetCancelSubscription = (payload: Payload) =>
+  makeActionObject('resetCancelSubscription', payload);
```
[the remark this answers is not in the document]
lmorchard: Yeah, this is a part where things are still a little convoluted. The `resetCancelSubscription` used in the `Subscriptions` route is actually something like `p => store.dispatch(resetCancelSubscription(p))` thanks to [the redux connect() utility call](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Subscriptions/index.tsx#L343). 

The type annotation to describe that gets kind of complicated, but all that the onDismiss handler for the dialog cares about is calling a function, any function. So the `onDismiss as () => void` type kind of hand waves that away. 

Hmm, not sure if that explanation helps. But... I wonder if the reason these lines show up as untested is that we just don't have a test that dismisses the dialog after subscription cancellation?

PR #2604, opened 2019-09-25
feat(db): Try using redis to store access tokens.

shane-tomlinson:
Kind of sort of works with caveats. The basics work, but complex queries
that join tokens w/ other tables do not fully work.


6a68:
This might be the path out of the current outage. Looking now, might try to figure out ways around the join-related breakage

6a68:
Looking through the [queries](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js#L129-L263), I think we can get around needing to do joins by including the client name and client ID into the token records in redis. We'll also need to have a secondary index on the user ID, assuming the primary key is the token ID.

shane-tomlinson:
I'm closing this until we've had a chance to decide on next steps and decide Redis is the way to go.

Issue #2608, opened 2019-09-25
Randomize the access token lifetime


6a68:
To reduce 'thundering herd' effects due to mass token expiration, we could introduce randomness into the access token lifetime.

PR #2610, opened 2019-09-25
fix(oauth): Add some random expiration time to access tokens

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/2608

Started working on this with Jared. It will add some configurable random amount of expiration time to tokens.

@shane-tomlinson @vladikoff Thoughts on the approach?

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
@@ -19,6 +19,9 @@ const unique = require('../../unique');
 const patch = require('./patch');
 
 const MAX_TTL = config.get('expiration.accessToken');
+const EXPIRES_AT_TTL_PADDING = config.get(
+  'expiration.accessTokenExpiresAtPadding'
+);
 const REQUIRED_SQL_MODES = ['STRICT_ALL_TABLES', 'NO_ENGINE_SUBSTITUTION'];
 const REQUIRED_CHARSET = 'UTF8MB4_UNICODE_CI';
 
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
```diff
@@ -469,15 +472,17 @@ MysqlStore.prototype = {
     return this._write(QUERY_CODE_DELETE, [hash]);
   },
   generateAccessToken: function generateAccessToken(vals) {
+    // In certain scenarios to help avoid thundering herd, we add some padding to the expiration time of the token.
+    vals.ttl = (vals.ttl * 1000 || MAX_TTL) + EXPIRES_AT_TTL_PADDING;
```
6a68: Should the second term in the sum be something like 

`(Math.floor(Math.random() * EXPIRES_AT_TTL_PADDING))`?

vbudhram: Missed the most important part 🙇🏽‍♂️, fixed!

packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
```diff
@@ -469,15 +472,19 @@ MysqlStore.prototype = {
     return this._write(QUERY_CODE_DELETE, [hash]);
   },
   generateAccessToken: function generateAccessToken(vals) {
+    // In certain scenarios to help avoid thundering herd, we add some padding to the expiration time of the token.
+    vals.ttl =
```
lmorchard: Maybe not a real problem and just a nit, but modifying incoming params in a function always looks like future side-effect trouble to me. Might be better to assign to a new const var and use that below.

6a68:
Looks like this failed on the `cargo audit` errors addressed by #2622. Rebased to pick up that fix.

lmorchard:
Might've rebased against the wrong commit? Looks like this is 164 commits now 😕

6a68:
yeah, I thought I force-pushed from the right branch back into this one, but I still see a ton of extra commits hanging out. I can clean this up in a bit.

6a68:
Looks like I've now fixed the bad rebase. Now this is train-145, plus the CI fix for the email server, and what I think is the latest version of the actual fix commit. Still might need to pull in the nsp fixes to get this to pass, we'll see.

vbudhram:
Since this won't land today and train-146 gets out on Monday, it makes sense to target this against that train, closing here.

clouserw:
@vbudhram you closed this here -- was there a PR somewhere else against a different train?

PR #2616, opened 2019-09-25
feat(metrics): add statsd perf metrics for SNS/SQS in auth server

chenba:
Part of #997 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/senders/sms.js
+++ b/packages/fxa-auth-server/lib/senders/sms.js
@@ -24,7 +24,7 @@ class MockCloudwatch {
   }
 }
 
-module.exports = (log, translator, templates, config) => {
+module.exports = (log, translator, templates, config, statsd) => {
   const cloudwatch = initService(config, Cloudwatch, MockCloudwatch);
   const sns = initService(config, Sns, MockSns);
 
@@ -65,11 +65,15 @@ module.exports = (log, translator, templates, config) => {
           },
           PhoneNumber: phoneNumber,
         };
+        const startTime = Date.now();
 
         return sns
           .publish(params)
           .promise()
           .then(result => {
+            if (statsd) {
+              statsd.timing('sms.send', Date.now() - startTime);
+            }
             log.info('sms.send.success', {
               templateName,
               acceptLanguage,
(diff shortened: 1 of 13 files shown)


packages/fxa-auth-server/test/local/sqs.js
```diff
@@ -0,0 +1,58 @@
...
+    statsd = { timing: sinon.stub() };
+    SQSReceiver = require(`${ROOT_DIR}/lib/sqs`)(log, statsd);
+    testQueue = new SQSReceiver('testo', [
+      'https://sqs.testo.meows.xyz/fxa/quux',
+    ]);
+    const receiveStub = sinon.stub();
+    receiveStub.onFirstCall().callsFake((qParams, cb) => {
+      cb(null, { Messages: [JSON.stringify({ Body: 'SYN' })] });
+    });
+    receiveStub.returns(null);
```
chenba: The callback is [recursive](https://github.com/mozilla/fxa/blob/28d25cfd6c26d9232c1e36310b7b29efc8763cac/packages/fxa-auth-server/lib/sqs.js#L87).  This is to prevent the recursion.

chenba:
@lmorchard Since you reviewed the previous two statsd PRs, could you review this as well? 😀

packages/fxa-auth-server/lib/senders/sms.js
```diff
@@ -65,18 +65,25 @@ module.exports = (log, translator, templates, config) => {
...
           PhoneNumber: phoneNumber,
         };
+        const startTime = Date.now();
 
         return sns
           .publish(params)
           .promise()
           .then(result => {
+            if (statsd) {
+              statsd.timing('sms.send', Date.now() - startTime);
```
bbangert: Might be useful to split these. `sms.send.success` and `sms.send.error` as the log messages do.

chenba: D'oh!  Just merged.  I'll created a follow-up PR.

chenba: Addressed in #2767

PR #2624, opened 2019-09-26
feat(db): Improve redis integration for access token management

6a68:
Hopefully moves #2604 forward a bit.

* Add a client ID -> token ID index, to enable client deletes.

* Denormalize data, copying some useful client fields into the redis
  token (publicClient, clientName, clientCanGrant), so that we don't
  need to do joins with the redis data.

* Rely on redis to auto-expire tokens, instead of manually running a
  periodic purge. Not sure how this will work for non-expiring pocket
  tokens.

* Many open TODOs called out in the code, which is also untested.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
@@ -49,6 +49,7 @@ function MysqlStore(options) {
     });
   });
 
+  // TODO: other configs needed here?
   this.redis = redisFactory(
     {
       enabled: true,
@@ -60,6 +61,7 @@ function MysqlStore(options) {
 // Apply patches up to the current patch level.
 // This will also create the DB if it is missing.
 
+// TODO: add migrations / patcher for redis (#2617)
 function updateDbSchema(patcher) {
   logger.verbose('updateDbSchema', patcher.options);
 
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
```diff
@@ -438,7 +441,16 @@ MysqlStore.prototype = {
...
-  removeClient: function removeClient(id) {
+  // TODO: beware, this is a very slow series of blocking deletes. We should
+  //       upgrade to a redis library that supports the UNLINK command, which
+  //       would instead do the deletes of all client tokens in a non-blocking
+  //       background thread. #2623
+  removeClient: async function removeClient(id) {
+    const clientId = id.toString('hex');
+    const clientTokenIds = await this.redis.smembers(clientId);
+    for (const tokenId of clientTokenIds) {
+      await this.removeAccessToken(tokenId);
```
dannycoates: i think an unlink would be just as slow as a del on a single key. it's only when you pass multiple keys in the same call that unlink is faster

6a68: :+1: Yeah, my concern with unlink is for deleting clients, where we'd need to delete every token stored under the client ID

shane-tomlinson:
I've closed https://github.com/mozilla/fxa/pull/2604 which was the original Redis PR, at least until we decide that's the path we are going to take.

PR #2631, opened 2019-09-26
chore(metrics): add sampling rate config for statsd

chenba:
In light of what Shane said (on Slack) about how statsd affected performance previously, default to off and a low sampling rate.

@mozilla/fxa-devs r?


lmorchard:
Test failure looks maybe unrelated, will try a re-run

PR #2655, opened 2019-09-30
feat(metrics): add product_id and plan_id to more amplitude events

### LINKED ISSUE #989: Add event properties to existing Account Creation funnel
These events already exist, but the properties are new.

- [ ] Add `product_id` property to the `fxa_subscribe - view` event
- [x] Add `product_id, plan_id` properties to the `fxa_reg - view` event
- [x] Add `product_id, plan_id` properties to the `fxa_reg - engage` event
- [x] Add `product_id, plan_id` properties to the `fxa_reg - submit` event
- [ ] Add `email_type=registration, email_template=downloadSubscription` properties to the `fxa_email - sent` event
- [ ] Add `email_type=registration, email_template=??` properties to the `fxa_email - click` event
- [x] Add `product_id, plan_id` properties to the `fxa_reg - email_confirmed` event
- [ ] Add `product_id, plan_id` properties to the `fxa_reg - complete` event

See the [product requirements](https://docs.google.com/document/d/1d39yRiYt4rhahtMae04Pwcx5vldNQJeF1grz_KO8qJw/edit#)

philbooth:
Fixes #989 (maybe).

Hey @bbangert, this is what I had in mind to add `planId` and `productId` to our other metrics, based on what we discussed in Slack the other week. Opened as draft because I need some guidance to test it properly and also to tighten up the parameter validation a bit. I'll call that out in the inline comments.

The approach I took refactors some of what you wrote before, I moved some stuff into a new subscription model that follows the same pattern laid out by the flow model. There are enough similarities between the two that I thought it made sense to keep them consistent.


packages/fxa-content-server/app/scripts/lib/metrics.js
```diff
@@ -53,8 +54,8 @@ const ALLOWED_FIELDS = [
   'migration',
   'navigationTiming',
   'numStoredAccounts',
-  'plan_id',
-  'product_id',
+  'planId',
+  'productId',
```
philbooth: Most of the properties in here a camel-case, so I renamed these for consistency.

packages/fxa-content-server/app/scripts/lib/metrics.js
```diff
@@ -383,8 +401,10 @@ _.extend(Metrics.prototype, Backbone.Events, {
       marketing: flattenHashIntoArrayOfObjects(this._marketingImpressions),
       migration: this._migration,
       numStoredAccounts: this._numStoredAccounts,
-      plan_id: this._planId,
-      product_id: this._productId,
+      // planId and productId are optional so we can physically remove
+      // them from the payload instead of sending NOT_REPORTED_VALUE
+      planId: this._subscriptionModel.get('planId') || undefined,
+      productId: this._subscriptionModel.get('productId') || undefined,
```
philbooth: `|| undefined` is to overwrite `null`, because `undefined` won't be serialised so saves a little bit of space in the payload.

packages/fxa-content-server/app/scripts/models/subscription.js
```diff
@@ -0,0 +1,50 @@
...
+      this.set('productId', options.productId);
+      return;
+    }
+
+    this.window = options.window || window;
+    const params = Url.searchParams(this.window.location.search);
+
+    if (params.plan_id) {
+      this.set('planId', params.plan_id);
+      this.set('productId', this.window.location.pathname.split('/').pop());
```
philbooth: This is hackish and based on our Slack conversation. `planId` is fine, but obviously I'm making a pretty big assumption that I can always just pop `productId` from the path like that. Is there some validation/matching I can do to make it more reliable? Do product ids conform to a standard format at all?

philbooth: > Do product ids conform to a standard format at all?

Or, more sensibly, are they always underneath `product/` in the path?

shane-tomlinson:
@bbangert - what more needs to be done here to merge this?

bbangert:
@shane-tomlinson it needs to treat the plan_id as optional. Right now I believe the code requires it.

packages/fxa-content-server/app/scripts/models/subscription.ts
```diff
@@ -0,0 +1,102 @@
...
+      const planId = this.getSearchParam('plan_id');
+      if (planId) {
+        this.set('planId', planId);
+      }
+    }
+  }
+
+  _getProductIdFromPath() {
+    const pathname = this.getPathname();
+    const result = SUBSCRIBE_PRODUCT_PATHNAME_REGEXP.exec(pathname);
```
shane-tomlinson: @bbangert I decided to be a bit more strict here and search for `/subscriptions/products/prod_(rest_of_id)`. Does the product ID include the `prod_` prefix?

packages/fxa-content-server/app/scripts/models/subscription.ts
```diff
@@ -0,0 +1,102 @@
...
+  _setSubscriptionInfoFromUrl() {
+    const productId = this._getProductIdFromPath();
+    if (productId) {
+      // If the user is browsing directly to /subscriptions/prod_*
+      this.set('productId', productId);
+
+      // only get the planId from the query params if the user is at
+      // a product path. It's possible for the plan_id to not be
+      // specified, in which case the default plan will be used.
+      const planId = this.getSearchParam('plan_id');
```
shane-tomlinson: @bbangert IIUC the comments you made to Phil, the planId is optional AND should only be fetched from the URL if the pathname begins with `/subscriptions/products/prod_*`

packages/fxa-content-server/app/scripts/models/subscription.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import Backbone from 'backbone';
+import Cocktail from '../lib/cocktail';
+import ResumeTokenMixin from './mixins/resume-token';
+import UrlMixin from './mixins/url';
+
+const SUBSCRIBE_PRODUCT_PATHNAME_REGEXP = /^\/subscriptions\/products\/(prod_\w+)/;
+
+// Neither UrlMixin nor ResumeTokenMixin are classes or interfaces
+interface IUrlMixin {
```
shane-tomlinson: @bbangert what's a good way to define interfaces for JS files like these that don't export classes but are still mixins? My typescript foo is not great here.

packages/fxa-content-server/app/scripts/models/subscription.ts
bbangert: The query params should always be checked for a plan_id, we might require them in the future based off some integration issues so far.

packages/fxa-content-server/app/scripts/models/subscription.ts
bbangert: It should always include that, yes.

packages/fxa-content-server/app/scripts/models/subscription.ts
shane-tomlinson: UrlMixin and RefreshTokenMixin are mixins, but not Typescript type mixins based on classes. Typescript needs to be made aware their interface is available on the consuming class or else it complains about methods not being available.

packages/fxa-content-server/app/scripts/models/subscription.ts
shane-tomlinson: "should always be checked" meaning should *always* be checked, even in a verification flow where that query param won't be there and is only found in the RefreshToken, or always checked as part of a signup/signin flow pre-email verification as well as the subscriptions flow?

shane-tomlinson:
@mozilla/fxa-devs - r?

You can see the product_id and plan_id show up in these screen caps:

## login screen event
Screenshot 2019-10-11 at 11 53 31

## signup screen event
Screenshot 2019-10-11 at 11 54 07

## signup verification event - the params make it through the resume token!
Screenshot 2019-10-11 at 11 54 28

shane-tomlinson:
@6a68, @chenba - I'm going to be out this afternoon, if you think this is ready to merge, can you merge it before the train cut?

shane-tomlinson:
Note, this PR handles these reqs from #989:

* Add product_id, plan_id properties to the fxa_reg - view event
* Add product_id, plan_id properties to the fxa_reg - engage event
* Add product_id, plan_id properties to the fxa_reg - submit event
* Add product_id, plan_id properties to the fxa_reg - email_confirmed event

But it does *not* handle these:

* Add product_id property to the fxa_subscribe - view event
* Add email_type=registration, email_template=?? properties to the fxa_email - sent event
* Add email_type=registration, email_template=?? properties to the fxa_email - click event
* Add product_id, plan_id properties to the fxa_reg - complete

1. I'm not sure where to add the first one.
2. The 2nd seems like it should already be done here: https://github.com/mozilla/fxa/blob/c9ee419a6e8ee1a2b5f4b4fb859e575bf2b6360b/packages/fxa-auth-server/lib/email/utils/helpers.js#L86
3. The third is [emit from the content server](https://github.com/mozilla/fxa/blob/c9ee419a6e8ee1a2b5f4b4fb859e575bf2b6360b/packages/fxa-content-server/server/lib/amplitude.js#L190), so I'm not sure how to get the email_template info there in a maintainable way unless the link contains a query param indicating which template is used.
4. The forth is emit from the auth server, we could send the product_id/plan_id from the resume token whenever verifying so that it's correctly logged.

packages/fxa-content-server/app/scripts/models/subscription.ts
bbangert: I think you can just make an interface that extends the other multiple classes, then assign it.

e.g. https://www.typescriptlang.org/docs/handbook/mixins.html

packages/fxa-content-server/app/scripts/models/subscription.ts
bbangert: We are using `plan`, not `plan_id`. Sorry for the oversight previously on this, I didn't notice the change.

bbangert: I'm not entirely clear about the context in your question. Anytime we would check for the product in the URL, we should check for the plan in the query.

packages/fxa-content-server/app/scripts/models/subscription.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (resumeToken) {
+      // this.resumeTokenFields is not set until the constructor
+      // completes, and this method can be called from within
+      // the constructor. Set the value
+      this.populateFromStringifiedResumeToken(resumeToken);
+    }
+  }
+}
+
+interface SubscriptionModel extends IUrlMixin, IResumeTokenMixin {}
```
bbangert: You should be able to extend ResumeTokenMixin and UrlMixin here according to that Typescript doc on mixins.

packages/fxa-content-server/app/scripts/models/subscription.ts
shane-tomlinson: I tried that but the reference mixins are class based, ours are object based and then typescript blows up.

bbangert: Doh. I think doing this is as good as it gets then.

shane-tomlinson:
> @shane-tomlinson Tested this locally and everything seems to be working except it doesn't log plan_id, maybe related to ben's comment?

{"op":"amplitudeEvent","event_type":"fxa_login - engage","time":1571087465630,"user_id":"26d445cc0b50448db828e3180422e043","device_id":"de114419d87546708fcab2d5726f53a3","session_id":1571087462941,"app_version":"147","language":"en","os_name":"Mac OS X","os_version":"10.14","event_properties":{"product_id":"123doneProProduct"},"user_properties":{"flow_id":"57a29baa35ee6c317358b7e80531bdf13c208442b7a1b9c809d8c27145a36cb3","ua_browser":"Firefox","ua_version":"70.0"}}
{"op":"amplitudeEvent","event_type":"fxa_login - submit","time":1571087475422,"user_id":"26d445cc0b50448db828e3180422e043","device_id":"de114419d87546708fcab2d5726f53a3","session_id":1571087462941,"app_version":"147","language":"en","os_name":"Mac OS X","os_version":"10.14","event_properties":{"product_id":"123doneProProduct"},"user_properties":{"flow_id":"57a29baa35ee6c317358b7e80531bdf13c208442b7a1b9c809d8c27145a36cb3","ua_browser":"Firefox","ua_version":"70.0"}}


Are you adding a `plan=<xxx>` query param? When I load the content server w/ `&plan=plan_123`, I see this in the logs:

Screenshot 2019-10-15 at 13 32 27

Issue #2072, opened 2019-08-05
[content server] Update Subscription Platform Terms of Service page


chenba:
**2019-09-30 EDIT**: This issue is for updating the ToS and PN footer links on the payment page.  They should link to https://www.mozilla.org/privacy/firefox-private-network and https://www.mozilla.org/about/legal/terms/firefox-private-network

**2019-09-26 EDIT**: This was a follow-up to an issue for adding a ToS page for the Subscription Platform.  However, that is no longer the case.  We just need to add a couple Firefox Private Network specific links.  Since the ToS and Privacy Notice content actually live in https://github.com/mozilla/legal-docs/, there is nothing to do in this repo.

This is a follow-up to #1986, to be completed once we have the copy in https://github.com/mozilla/legal-docs.

Specifically we need to at least
- [ ] update `fxa-content-server/server/lib/legal-templates.js` (look for a `TODO`) to search for the correct template file
- [ ] update grunt tasks to generate the template files

clouserw:
Blocked on legal content

clouserw:
Sorry Barry, there is a change of plans here.  We'll no longer host this ourselves and instead link out to an external URL.  I'll put the external URL here once we have it...

clouserw:
Unassigning myself as we are no longer blocked.  URLs below

PRIVACY_POLICY_URL = "https://www.mozilla.org/privacy/firefox-private-network"
TERMS_AND_CONDITIONS_URL = "https://www.mozilla.org/about/legal/terms/firefox-private-network"

chenba:
Just to confirm: we are keeping the newly created page, but instead of legal copy, we put two links on the page?  Also, the new page itself is not linked from anywhere (at least within the content server).

clouserw:
We should not have the new page.  We should link directly to the URLs.

chenba:
I just noticed that within the Terms of Service and Privacy Notice pages on the content server there are links out to terms and notices for individual products.  And the FPN ToS is already linked from https://accounts.firefox.com/legal/terms.

For this issue, I think we just need add the FPN Privacy Notice to the section in https://github.com/mozilla/legal-docs/blob/master/firefox_privacy_notice/en-US.md#firefox-accounts--join-firefox?  (In addition to reverting PR #1362 in this repo.)

@clouserw does that sound correct?

chenba:
Filed https://github.com/mozilla/legal-docs/issues/1363.

clouserw:
[...]

PR #2037, opened 2019-08-01
chore(legal): add a page for subplat terms of service

chenba:
Add /legal/subscription_terms to the content server.  It serves the
regular ToS until the actual content is available.  (See TODO in
fxa-content-server/server/lib/legal-templates.js.)

Part of #1986

@mozilla/fxa-devs r?

PR #2658, opened 2019-09-30
chore(legal): revert PR #2037; no SubPlat ToS

chenba:
This reverts PR #2037.  We are no longer adding a Subscription Platform ToS page.

Part of #2072


packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -586,10 +586,13 @@ var ERRORS = {
     errno: 1062,
     message: t('Invalid redirect'),
   },
+  /*
+  Removed for https://github.com/mozilla/fxa/issues/2072
```
6a68: Nit: might as well stick with the syntax on the item above (line 579), `Removed in https://github.com/mozilla/fxa/pull/2658`

chenba: PR was not created at the time of writing; but can do now.  👍

PR #2404, opened 2019-09-04
remove trailhead specific style

dannycoates:
This addresses the points in https://github.com/mozilla/fxa/issues/1755#issuecomment-523645220 and design in https://mozilla.invisionapp.com/share/5UM3J3TXWMR#/screens (minus the key icon)

PR #2663, opened 2019-09-30
fix(content-server): fix #2399 - add autofocus to password field during sign up

### LINKED ISSUE #2399: [iOS] [Train 145] "Repeat password" field is highlighted
Environment:
iOS, stage - Train 145

Steps to reproduce:
1. Go to "Sign in to Sync".
2. Enter the email and tap on "Continue"

Expected result:
"Password" field is highlighted.

Actual result: 
"Repeat password" field is highlighted.

Video: https://drive.google.com/file/d/1nQvs-KYURiRYd7D6MN9qezQX5AZGkewe/view?usp=sharing

Note: not reproducible on Android/Win10.

LZoog:
fix for #2399 

@dannycoates if possible I'd like to get the r+ from at least you because this seemed deceptively simple and you've modified this code a lot recently. (My screenshot differs from the video QA provided because of your PR here which removed the trailhead specific style: #2404)

This fix works for me in XCode:
image

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
@@ -21,7 +21,7 @@
       <a href="/" class="use-different">{{#t}}Change email{{/t}}</a>
 
       <div class="input-row password-row">
-        <input id="password" type="password" class="password check-password tooltip-below" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required data-form-prefill="password" data-synchronize-show="true" />
+        <input id="password" type="password" class="password check-password tooltip-below" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required autofocus data-form-prefill="password" data-synchronize-show="true" />
         <div class="helper-balloon" tabindex="-1"></div>
       </div>
 


dannycoates:
We actually don't want autofocus at all on mobile so that you can see the form unobstructed on first load. I can tell from the video that the bug is from before my changes in train 146 so it's possible this is already fixed.

LZoog:
@dannycoates awesome, thanks, something told me this fix might be off. When testing in XCode, I didn't see the autofocus in "repeat password." I'll confirm this is still an issue with QA.

shane-tomlinson:
This also fixes https://github.com/mozilla/fxa/issues/2705

shane-tomlinson:
> We actually don't want autofocus at all on mobile so that you can see the form unobstructed on first load. I can tell from the video that the bug is from before my changes in train 146 so it's possible this is already fixed.

In views/base.js, the [focusAutofocusElement](https://github.com/mozilla/fxa/blob/3f3321edcb41c9725bb0b44bf2d3daa2dff68b57/packages/fxa-content-server/app/scripts/views/base.js#L953) only tries to autofocus on desktop and skips for devices with touch event support, so this seems like this fix should work. And it'd fix #2705.

LZoog:
@janiceahn @juliemccracken @ryanfeeley 

We've got #2399 open from QA to focus on the first password field in mobile, and #2705 open for focus in desktop.

Is the expected behavior still to focus in desktop and _not_ focus on mobile? I can close this PR + issue #2399 if that's the case.

janiceahn:
@LZoog Yes, confirming what Danny and Shane have said—no autofocus in mobile. does this answer your question?

LZoog:
Closing this PR as mobile is behaving as expected.

PR #2667, opened 2019-09-30
fix(codes): Fix resend code for the sign-in code feature

vbudhram:
I noticed this while creating the point release for the experiment for sign in codes. Previously, when the user attempted to resend the sign-in code, it would send the incorrect email template. However the code was correct and that was why the tests passed.

This fixes the issue and adds some more tests that would have correctly caught this. Targeting as a point release for train-146, since we are opening up the experiment for more users.


vbudhram:
I've pull a commit from master to help resolve the npm advisories.

PR #2674, opened 2019-10-01
fix(authorization): Fix timezone issues when trading a code for a token

### LINKED ISSUE #2626: OAuth authorization codes for 123done expire immediately when using MySQL backend
STR

1. Start the servers using `./pm2 start mysql_server.json`
2. Go to `http://127.0.0.1:8080`, sign in.
3. When returning to 123done, an error is displayed: 

> name	"JWTVerificationError"
> message	"malformed"

shane-tomlinson:
Times written to the OAuth database are in UTC time. Times read out
do not have timezone information and are assumed to be in local time.

In London for this time of the year, that means we are 1 hour ahead
of UTC, and all dates read from the DB are one hour in the past.

Specify a `timezone` when configuring the MySQL connection, default
to `UTC` which is the timezone of the MySQL database.

fixes #2626

@jrgm or @jbuck - will this cause any problems for the production servers? The MySQL docker containers seem to be running in UTC, but I don't want to make that assumption for prod.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/config.js
@@ -239,6 +239,15 @@ const conf = convict({
       default: 10,
       env: 'MYSQL_CONNECTION_LIMIT',
     },
+    timezone: {
+      default: 'Z',
+      doc:
+        'The timezone configured on the MySQL server. This is used to type cast server date/time' +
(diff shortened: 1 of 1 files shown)


jrgm:
Fxa RDS db's are set to UTC, timezone unaware, so the database data has always been this way. Is there a particular need to set this?

shane-tomlinson:
> Fxa RDS db's are set to UTC, timezone unaware, so the database data has always been this way. Is there a particular need to set this?

Yup, full context over in #2626

PR #2681, opened 2019-10-01
fix(config): Add an environment variable name for the auth server url

shane-tomlinson:
follow on for #2538

@vbudhram - r?


vbudhram:
Thanks @shane-tomlinson!

PR #2690, opened 2019-10-01
task(repo): add security.md to repo

meandavejustice:


6a68:
@meandavejustice Hey, what's the background on this change? Is this something secops wants to add to all repos?

SECURITY.md
```diff
@@ -0,0 +1,3 @@
+# Security Policy
+
+For information on vulnerabilities, security advisories, our bug bounty program, and how to report a sensitive vulnerability to go https://www.mozilla.org/en-US/security/
```
hwine: The "en-US" should be removed, so the viewer sees it in their desired locale.

meandavejustice:
@6a68  it was suggested in github UI, I haven't seen any requests

6a68:
Cool, I think our process in the past has been to direct people to file confidential bugs on bugzilla. Not sure if that's changed. @jvehent where should we be directing sec bug filers for FxA?

meandavejustice:
@6a68 All of the info for sec bug filing, vulnerabilities, the bounty program etc is listed at `/security`. I figured we would just link there, so we don't end up with stale information if it updates.

hwine:
fwiw, mozilla-services has an org wide default [security.md](https://github.com/mozilla-services/.github/blob/master/SECURITY.md), which might be directly applicable to repos deployed by mozilla-services.

shane-tomlinson:
@meandavejustice it seems reasonable to use the security.md @hwine linked to, for consistency across repos as much as anything.

Issue #807, opened 2019-04-11
Replace tap with mocha in the customs server


shane-tomlinson:
The customs server uses the tap test runner whereas our other repos use mocha. We should standardize on mocha.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-804)
┆Issue Number: FXA-804

vbudhram:
Another reason to do this 😬, https://github.com/mozilla/fxa/pull/2694

data-sync-user:
➤ Wil Clouser commented:

Thank you for filing this issue. This is not currently on our roadmap, and in an effort to focus our work, we are closing old issues that we are unlikely to be closed in the future. Thanks again.

PR #2694, opened 2019-10-01
chore(tests): Update tap to run customs server tests serially

### LINKED ISSUE #2691: Customs server tests fail locally and on Cirlce CI
First noticed this today while making train-147 release, the local custom server tests were failing. Oddly enough, Circle CI also reports these tests as passing, https://app.circleci.com/jobs/github/mozilla/fxa/84607.

A while back, we had to run custom-server on a much older version of node, so maybe that might be part of it?

vbudhram:
Starting in Tap 13 (custom server test runner), it defaults to running tests in parallel (number based on CPU cores). This currently breaks custom server tests because we reuse IP addresses for different tests, and they are stored in memcache.

Per [documentation](https://node-tap.org/docs/api/parallel-tests/#enablingdisabling-parallelism-in-the-test-runner), we can disable it by adding a top level file called `tap-parallel-not-ok`. Eventually we should make the tests run in parallel or better yet replace tap.

Fixes #2691


shane-tomlinson:
Ref #807 to replace tap with mocha

PR #2709, opened 2019-10-02
fix(dev): clean up pm2 scripts and docs

vladikoff:


mysql_servers.json
```diff
@@ -1,21 +1,5 @@
 {
   "apps": [
-    {
```
vladikoff: this is for muscle memory of getting to the mail helper by doing `./pm2 logs 0`

mysql_servers.json
```diff
@@ -235,6 +219,23 @@
...
+    },
+
+    {
+      "name": "MySQL server PORT 3306",
+      "script": "_scripts/mysql.sh",
+      "max_restarts": "1",
+      "min_uptime": "2m"
+    },
+    {
+      "name": "auth-server db mysql PORT 8000",
```
vbudhram: Mind moving this one below mailer helper, for my muscle memory of `./pm2 logs 1`?

vladikoff: ah yes thanks

PR #2710, opened 2019-10-02
fix(email): revert email subjects to sentence casing

### LINKED ISSUE #1684: Inconsistent casing of subject lines in emails
Most of the FxA emails have the subject line in sentence-casing. But while reviewing #1675, I noticed that "Finish Creating Your Account", "Reminder: Finish Creating Your Account" and "Your Firefox Account is Confirmed" have title-casing.

It looks like it's how they were specified in the designs for #1073 / #1087, but that seems weirdly inconsistent to me. Is it a bug or a feature?

philbooth:
Fixes #1684.

I think there was some confusion about who was going to fix this that ended up with nobody doing it. So this updates all the subjects to match the final decisions from the Google Doc in the linked issue.

Note there are a couple of discrepancies because we still have code for non-trailhead emails in here. I'm going to remove that in a separate PR, just in case I'm wrong in thinking it can be removed now.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -425,7 +425,7 @@ module.exports = function(log, config, oauthdb) {
 
     let templateName = 'verify';
     const metricsTemplateName = templateName;
-    let subject = gettext('Verify Your Account');
+    let subject = gettext('Verify your account');
     let action = gettext('Activate now');
     const query = {
       uid: message.uid,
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -508,7 +508,7 @@ module.exports = function(log, config, oauthdb) {
     const templateName = 'verifyShortCode';
     const metricsTemplateName = 'verify';
     const code = message.code;
-    const subject = gettext('Verification code: %(code)s');
+    const subject = gettext('Account authorization code');
```
vbudhram: After user testing they wanted to keep this, I think we can sentence casing the original?

Objections @ryanfeeley ?

philbooth: Ah, right-o, I was just going from the Google Doc.

philbooth: This should be back how you want it now, @vbudhram.

PR #2719, opened 2019-10-02
fix(config): fixed missing oauth config types

dannycoates:
fxa-dev is broken because the schema lost some convict format types


dannycoates:
test failure unrelated

PR #2733, opened 2019-10-03
chore(deps): update babel

dannycoates:


dannycoates:
IDK why #2313 is busted, but adding profile-server to the test-content-server script shouldn't be necessary. I think this one updates all the same deps.

PR #2734, opened 2019-10-03
feat(metrics): add endpoint for emitting amplitude metrics

### LINKED ISSUE #2425: [fxa-payments-server] Create metrics reporting capability in payments
The payments-server has no metrics reporting infrastructure in it. Before any of the metrics issues can be done that add new events in the payment-server, some infrastructure to actually log the metrics must be built.

In content-server, the events are emitted to the `post-metrics` end-point, which logs them out for the log collector to submit to amplitude. Porting the code into the payment-server from content-server is likely to be more trouble than its worth, so the payments-server should have some similar setup with an end-point that the client-side can emit to, that logs out the event for collection.

philbooth:
Fixes #2425.

This is a quickly hacked-together implementation of some back-end logic that accepts Amplitude metrics from the payments server front-end. It adds a `POST /metrics` route that does some rudimentary validation, calls the common data-wrangling code from `fxa-shared`, then logs the events. Opened as draft because it's not complete yet and I may not have time to complete it before I leave.

Because we're not building on top of pre-existing metrics layers like we had to for the auth and content servers, the code here can be a bit simpler than they can. In particular, we don't have to build up complex maps for `EVENTS` and `FUZZY_EVENTS`, instead we can just pass in a single fuzzy event mapping that works in all cases. On the front-end we'd then submit events in the format `amplitude.${GROUP}.${EVENT}`, where `${GROUP}` is one of the keys defined [in `fxa-shared`](https://github.com/mozilla/fxa/blob/526f5d3b6d5743affceb7047da60c39839429dc9/packages/fxa-shared/metrics/amplitude.js#L11-L28) and `${EVENT}` is the part of the event name you want to appear after the hypen (have a look at our other event names if that's not clear).

Things that still need doing:

- [x] Pick a validation lib. I know there's been some ~~Hapi~~ Joi vs JSON schema discussion in the past, and didn't want to assume anything on that front.
- [x] Implement the non-existent `./user-agent` dependency I've referred to. Worst-case it can be copy/pasted from the content server, but maybe it should be refactored into `fxa-shared`.
- [ ] Submit some events from the front-end. I recommend using the [beacon API](https://www.w3.org/TR/beacon/) to actually send the data.
- [ ] Write tests.
- [ ] After it lands, arrange with @jbuck to include the payments server logs in the Amplitude pipeline.

Also, I wrote it in JS because the other back-end code all appears to be JS, but maybe it should have been TS instead?

I'm not gone yet, so still might have time do some more here, but wanted to get it up now in case I don't.


packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const { GROUPS, initialize } = require('../../../fxa-shared/metrics/amplitude');
+const log = require('./logging/log')();
+const ua = require('./user-agent');
```
philbooth: This doesn't exist yet.

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,84 @@
...
+const { version: VERSION } = require('../../package.json');
+
+const FUZZY_EVENTS = new Map([
+  [
+    // Emit events from the front-end as `amplitude.${GROUP}.${EVENT}`
+    /^amplitude\.([\w-]+)\.([\w-]+)$/,
+    {
+      group: group => GROUPS[group],
```
philbooth: https://github.com/mozilla/fxa/blob/526f5d3b6d5743affceb7047da60c39839429dc9/packages/fxa-shared/metrics/amplitude.js#L11-L28

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,84 @@
...
+    ...mapLocation(data.location),
+    ...data,
+  });
+
+  if (amplitudeEvent) {
+    log.info('amplitudeEvent', amplitudeEvent);
+  }
+};
+
+function mapBrowser(userAgent) {
```
philbooth: All of the code from here down is copy/pasted from the content server so could be refactored to `fxa-shared` if you like.

packages/fxa-payments-server/server/lib/routes/post-metrics.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const { data, events } = request.body;
+    events.forEach(event => amplitude(event, request, data));
+    response.status(200).end();
+  },
+};
+
+function validate(body) {
+  // TODO: replace this with JSON schema or hapi or whatever
```
philbooth: When you implement proper validation, I also recommend including some kind of server secret. This endpoint WILL end up being hit by port scans and receive all kinds of illegitimate traffic. The metrics wlil be complete bunk if you don't prevent those getting through. (not to mention security concerns too of course)

chenba:
👍 Thanks for the draft PR, @philbooth 

This is maybe a n00b question, but is there a reason we cannot send a metrics request from a payments page to the content server?

philbooth:
> is there a reason we cannot send a metrics request from a payments page to the content server?

No reason why we can't, but having it's own endpoint feels conceptually a little bit simpler and more semantically correct to me. And it's removes the possibility of a footgun that could happen if they share an endpoint.

For metrics to be accepted by the content server endpoint, they need to have a valid `flowId`/`flowBeginTime` pair. There's a little crypto step for validating them, so the payments server back-end would need to grow the ability to generate that pair and pass it to the front-end. And we'd also need to open it up the content server endpoint to CORS from the payments server.

Those are both trivial enough things to work round, but they still leave us with a potential footgun of clashing event names. The way the Amplitude stuff works right now is you give it a list of event name mappings and then it transforms a bunch of source events to Amplitude events as it encounters them. If we're sending events from both front-ends to the content server, it leaves open the possibility that payments server code could one day accidentally start emitting an event name that matches a mapping defined by the content server. Then we'd be generating bogus events in Amplitude without necessarily realising it, so our metrics would be wrong.

If we keep the two separate, that can't happen.

6a68:
Thanks for opening this, @philbooth! I'll be taking this over, and I really appreciate the start at an implementation, and _especially_ the checklist of open questions :beers:

6a68:
Hoping to integrate payments as an RP instead of adding custom code. Closing this for now, I'll reopen if the standard RP approach doesn't work out.

6a68:
Reopening: the RP stuff isn't going to happen by Friday, so I'll plan to finish this one up for now :+1:

6a68:
I've made a few changes today:

- added joi as the validation library (hoping to be able to reuse the validators from the content-server's post-metrics route, once I better understand how the payments-server payload differs)
- extracted the `user-agent.js` lib from content-server into fxa-shared, updated the tests, and connected it to payments-server
- extracted some copy-pasted functions from content-server amplitude.js into fxa-shared, and updated the content-server and payments-server amplitude.js files

Next, I'll figure out validation with Shane and Barry tomorrow, try to connect this to Barry's PR to submit events from the front-end, add tests, and start to coordinate with jbuck around ops work on the backend.

chenba:
> connect this to Barry's PR to submit events from the front-end

My PR is only submitting performance timing events, so it won't exercise this PR.

6a68:
Rebased and added a few little changes

packages/fxa-payments-server/server/lib/routes/post-metrics.js
```diff
@@ -0,0 +1,53 @@
...
+  validate: {
+    body: BODY_SCHEMA,
+  },
+  preProcess: function(req, res, next) {
+    // convert text/plain types to JSON for validation.
+    if (/^text\/plain/.test(req.get('content-type'))) {
+      try {
+        req.body = JSON.parse(req.body);
+      } catch (error) {
+        logger.error(error);
```
6a68: TODO: not sure if I need to require logger here, or if the logger middleware takes care of it

packages/fxa-payments-server/Dockerfile
```diff
@@ -9,19 +9,45 @@ RUN npm ci
...
 COPY ["fxa-content-server", "../fxa-content-server/"]
 WORKDIR /fxa-content-server
 RUN npm ci
 WORKDIR /app
 ENV PUBLIC_URL /
 ENV INLINE_RUNTIME_CHUNK false
 RUN npm run build
-
-FROM node:12-stretch-slim
+FROM node:12-stretch
```
6a68: I'm not sure if this is OK, or if we should stick with node:12-stretch-slim and manually install git.

packages/fxa-payments-server/Dockerfile
```diff
@@ -9,19 +9,45 @@ RUN npm ci
...
+# Also need to copy over the tsconfig.json, or the linter will fail
+COPY --chown=app:app ["fxa-shared/tsconfig.json", "../fxa-shared/tsconfig.json"]
+# And the linter also needs the nsprc...
+COPY --chown=app:app ["fxa-shared/.nsprc", "../fxa-shared/.nsprc"]
+RUN mkdir /fxa-shared/node_modules
+WORKDIR /fxa-shared
+USER root
+RUN chown -R app:app /fxa-shared
+USER app
+RUN npm ci
```
6a68: Ugh, these Dockerfile additions are the fruit of a long struggle. Suggestions welcome.

packages/fxa-payments-server/Dockerfile
shane-tomlinson: This is a question for @jrgm and @jbuck.

packages/fxa-payments-server/Dockerfile
```diff
@@ -9,19 +9,45 @@ RUN npm ci
...
 COPY --chown=app:app [ "fxa-payments-server/", "./" ]
+
+# Borrowing this stanza from fxa-content-server/Dockerfile-build
+# Except, to avoid installing redis, we're just copying over the two files
+# that are explicitly needed.
+COPY --chown=app:app ["fxa-shared/metrics/user-agent.js", "../fxa-shared/metrics/user-agent.js"]
+COPY --chown=app:app ["fxa-shared/metrics/amplitude.js", "../fxa-shared/metrics/amplitude.js"]
+COPY --chown=app:app ["fxa-shared/package.json", "../fxa-shared/package.json"]
+COPY --chown=app:app ["fxa-shared/package-lock.json", "../fxa-shared/package-lock.json"]
+# Copy over one test, so npm run test won't barf
```
shane-tomlinson: Does this mean we'd need to copy over individual test files every time we add a new one?

packages/fxa-payments-server/Dockerfile
```diff
@@ -9,19 +9,45 @@ RUN npm ci
...
+COPY --chown=app:app ["fxa-shared/test/metrics/amplitude.js", "../fxa-shared/test/metrics/amplitude.js"]
+# Also need to copy over the tsconfig.json, or the linter will fail
```
shane-tomlinson: Why is the linter being run on fxa-shared and not just the payments code?

packages/fxa-payments-server/server/config/index.js
```diff
@@ -9,12 +9,28 @@ const path = require('path');
 const convict = require('convict');
 
 const conf = convict({
+  amplitude: {
+    disabled: {
```
shane-tomlinson: I'm a bit surprised that the content server uses `disabled` already. We generally use `enabled` within config rather than `disabled` because negating negatives is hard - `disabled: false` just takes too much thought.

packages/fxa-payments-server/Dockerfile
```diff
@@ -9,19 +9,45 @@ RUN npm ci
...
+FROM node:12-stretch
 RUN groupadd --gid 10001 app  && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app
 USER app
 WORKDIR /app
 COPY --chown=app:app --from=node-builder /app .
 COPY --chown=app:app [ "fxa-payments-server/", "./" ]
+
+# Borrowing this stanza from fxa-content-server/Dockerfile-build
+# Except, to avoid installing redis, we're just copying over the two files
```
shane-tomlinson: I'm a bit confused by this. Are you trying to avoid installing the redis db, or the redis npm module? Copying over just a few files feels like a future footgun, I can imagine someone shouting in frustration "but why aren't my files there?! The payments server already uses fxa-shared, and the files aren't there!"

Also, is this comment out of date? It looks like 7 files are copied rather than 2.

packages/fxa-payments-server/Dockerfile
jrgm: the difference is node:12-stretch is a 1GB image, -slim  is 1/6th that size. Mattered for time to provision a new ec2 instance for scaling in our AWS config, but may not matter in kubernetes on gcp.
also, the general principle of only installing what is required.

packages/fxa-payments-server/server/lib/amplitude.test.js
```diff
@@ -0,0 +1,26 @@
...
+const assert = require('chai').assert;
+
+const amplitudeConfig = {
+  disabled: false,
+};
+
+const testEvents = ['amplitude.sub.relyingPartyView'];
+// TODO: mock out logger to examine the fired log things.
+
+// TODO: test the snakeCase conversions: amplitude.sub.relyingPartyView -> amplitude.fxa_subscribe.relying_party_view
```
shane-tomlinson: Are you worried about these TODOs before merging?

packages/fxa-payments-server/server/lib/amplitude.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+describe('amplitude', () => {
+  describe('transforms and logs valid data for all planned ping formats', () => {
+    // TODO enumerate all planned pings, ensure they validate.
+    it('needs a placeholder or will fail', () => {
+      assert(true);
```
shane-tomlinson: This test isn't actually doing anything.

packages/fxa-payments-server/server/lib/routes/post-metrics.js
```diff
@@ -0,0 +1,63 @@
...
+
+module.exports = {
+  method: 'post',
+  path: '/metrics',
+  validate: {
+    body: BODY_SCHEMA,
+  },
+  preProcess: function(req, res, next) {
+    // convert text/plain types to JSON for validation.
+    if (/^text\/plain/.test(req.get('content-type'))) {
```
shane-tomlinson: IIRC it was only old Chrome that sent text/plain, this might no longer be needed.

packages/fxa-payments-server/server/lib/routes/post-metrics.js
shane-tomlinson: The content-server requires a logger here: https://github.com/mozilla/fxa/blob/56e95b4e68c13399ec988e61f74fa7dcc241ff6f/packages/fxa-content-server/server/lib/routes/post-metrics.js#L11

packages/fxa-shared/package.json
```diff
@@ -54,6 +54,7 @@
     "bluebird": "^3.5.5",
     "generic-pool": "^3.7.1",
     "moment": "^2.24.0",
+    "node-uap": "git://github.com/vladikoff/node-uap.git#9cdd16247",
```
shane-tomlinson: I wonder why we are still using this version, @vladikoff, insights?

packages/fxa-payments-server/Dockerfile
6a68: I agree, this feels hacky. I did this to avoid introducing redis into the build, which apparently is required for fxa-shared. I can try just adding redis and git to the node:12-stretch-slim image, that seems more reliable long-term.

packages/fxa-payments-server/Dockerfile
6a68: `npm ci` for fxa-shared runs the linter :-\

packages/fxa-payments-server/server/config/index.js
6a68: :+1: I'll flip the value here and file a bug to invert the value on the content server, too.

packages/fxa-payments-server/server/lib/amplitude.test.js
6a68: Yes, I'm working on the tests now

packages/fxa-payments-server/server/lib/amplitude.test.js
6a68: Yeah, mocha barfs if there's an empty test, that was a placeholder to use while getting the build step green.

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,58 @@
...
+    // Emit events from the front-end as `amplitude.${GROUP}.${EVENT}`
+    /^amplitude\.([\w-]+)\.([\w-]+)$/,
+    {
+      group: group => GROUPS[group],
+      event: (group, event) => toSnakeCase(event),
+    },
+  ],
+]);
+
+const transform = initialize({}, {}, FUZZY_EVENTS);
```
lmorchard: Rather than call `initialize` here, you might do some dependency injection to make testing easier.

That is, call this `initialize` in post-metrics.js and pass the `transform` result into the function exported below when it's called in the route. Then, you can create a mock `transform` in testing without needing to try mocking the whole module.

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,58 @@
...
+const {
+  GROUPS,
+  initialize,
+  mapBrowser,
+  mapFormFactor,
+  mapLocation,
+  mapOs,
+  mapUserAgentProperties,
+  toSnakeCase,
+} = require('../../../fxa-shared/metrics/amplitude.js');
```
lmorchard: Don't need the `.js` after `amplitude` here.

packages/fxa-payments-server/server/lib/amplitude.test.js
```diff
@@ -0,0 +1,26 @@
+const amplitude = require('./amplitude.js');
```
lmorchard: Don't need the `.js` after `amplitude` here.

6a68: The typescript module loader complained when the .js was omitted :-\

packages/fxa-payments-server/server/lib/amplitude.js
6a68: Great idea! I'll give that a try

chenba:
Looks good so far. 👍

packages/fxa-payments-server/server/lib/amplitude.test.js
6a68: Ah, it was Jest that couldn't find the modules without the .js extension. Just ran into that error again on CI.

packages/fxa-payments-server/Dockerfile
```diff
@@ -1,27 +1,37 @@
-FROM node:12-stretch AS node-builder
+FROM node:12-stretch-slim AS node-builder
 RUN groupadd --gid 10001 app  && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app
 RUN mkdir /fxa-content-server && chown -R app:app /fxa-content-server
+RUN apt-get -y update && apt-get -y install git && apt-get clean && rm -rf /var/lib/apt/lists
```
jbuck: I think if you revert the first line of this docker file back to `node:12-stretch` you don't need to install `git` because it already comes with git:

```
$ docker run --rm node:12-stretch git --version
git version 2.11.0
```

packages/fxa-payments-server/Dockerfile
```diff
@@ -1,27 +1,59 @@
...
 COPY ["fxa-payments-server/.storybook", ".storybook/"]
 COPY ["fxa-payments-server/public", "public/"]
 COPY ["fxa-payments-server/src", "src/"]
+
+#COPY --chown=app:app ["fxa-shared", "../fxa-shared/"]
+#WORKDIR /fxa-shared
+#USER root
+#RUN chown -R app:app /fxa-shared
+#USER app
+#RUN npm ci
```
chenba: Delete these line if they aren't needed no mo'.

6a68: Cool, I'm going to add a bug number too, as it's a fairly hacky looking  Dockerfile

6a68:
Filed #2938 to work on the tests, and #2937 to improve the Dockerfile.

packages/fxa-payments-server/Dockerfile
6a68: Thanks! Yeah, I mistakenly changed the first one from stretch-12 to stretch-12-slim, when I was supposed to be changing the second one :-P

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -0,0 +1,68 @@
...
+    // Emit events from the front-end as `amplitude.${GROUP}.${EVENT}`
+    /^amplitude\.([\w-]+)\.([\w-]+)$/,
+    {
+      group: group => GROUPS[group],
+      event: (group, event) => toSnakeCase(event),
+    },
+  ],
+]);
+
+// Ease testing by allowing mocks to be directly injected.
```
lmorchard: I don't think we should have a special carve-out for testing here, or module-global vars for `transform` & `log`. Looks like they're only used in the function exported by the module. Can't we pass `transform` and `log` in as parameters from the code using this exported function - i.e. whether "real" or testing code?

lmorchard: Oh, are tests for this code deferred for a follow-up? Can't find another module using this stuff https://github.com/mozilla/fxa/issues/2938

6a68: Yeah, tests are deferred for the moment, but I felt weird about adding that injection spot. I'll remove it.

6a68:
I see /metrics log lines in the payments-server in my local machine, so I think this is working. Closing. Thanks all for your help and reviews 🍻

PR #2737, opened 2019-10-03
fix(bug): ensure the jwt audience is set to the oauth-server url

dannycoates:
The jwt audience should be the oauth-server. Since both auth-server and oauth-server set `publicUrl` differently we can't use it as the audience directly. This change picks the right one based on the context the config was loaded in.


dannycoates:
> any way we can add a test?

working through that now :)

PR #2747, opened 2019-10-04
fix(email): remove old trailhead logic and promote trailhead templates

philbooth:
The trailhead-specific logic was removed from the content server some time ago, but it was never removed from the emails. From what I can tell, I think this means we reverted to pre-trailhead templates at that point. This change copies the trailhead templates over the old pre-trailhead ones, and removes all of the trailhead conditional logic.

@vbudhram, I think you might be the best person to review this?


philbooth:
I'll need one of you guys to merge this, sorry. I already removed myself from the GH org.

PR #2752, opened 2019-10-04
fxa-dev special case

### LINKED ISSUE #2746: oauth /v1/token broken on fxa-dev
fxa-dev runs auth-server under `/auth` i.e. `https://latest.dev.lcip.org/auth` but the underlying pool used by https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js#L37 only supports bare base urls i.e. `https://latest.dev.lcip.org` so the profile request ends up going to `https://latest.dev.lcip.org/v1/account/profile` and 404s, eventually ending with a `JWTVerificationError` in the browser.

Local dev and prod should be unaffected because they have simple base urls. Fixing this for fxa-dev will require annoying special case code changes, until oauth->auth is complete.

dannycoates:
fixes #2746 (I hope)

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/auth_server.js
@@ -10,7 +10,9 @@ const AppError = require('./error');
 module.exports = (log, config) => {
   const AuthServerAPI = createBackendServiceAPI(log, config, 'auth', {
     getUserProfile: {
-      path: '/v1/account/profile',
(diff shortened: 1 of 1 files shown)


dannycoates:
fyi this entire file will be gone soon so I don't care how silly this hack is

vladikoff:
@dannycoates hm I don't think fix worked, investigating... I still get 404s for /profile route on a new fxa-dev box

PR #2757, opened 2019-10-04
oauth->auth phase 2

dannycoates:
## Recap

In phase 2 the goal is to have all oauth web traffic going to the auth-server by pointing `oauth.accounts.firefox.com` at it. The OG oauth-server will still be deployed and running for a couple reasons:
- to serve requests while dns propagates
- as a fail-safe if we need to rollback
- to handle `fxa-notifier-aws` SQS events (for now)

## Local Setup

You'll need to update your `oauth_url` in `fxa-content-server/server/config/local.json` to `http://127.0.0.1:9000`

If you use `fxa-dev-launcher` you'll need to apply [this patch](https://gist.github.com/dannycoates/634be06bb7bbc13015702a9864f57cc0) to change the oauth server path.

## Changes

Most of the oauth-server code has been copied verbatim (aside from `require` or config path changes) into `lib/oauth`. I've called out non-trivial changes in review comments. Tests have been copied into `test/oauth` and added to `npm test`.

All oauth-server config has been added to the auth-server under `oauthServer` maintaining their original environment variable names.

I've tried to do minimal refactoring. There will be plenty of things we can make cleaner but this diff is already huge.

## Exclusions

I've intentionally excluded the administrative `/clients` and `/developer` routes because they aren't used in production and imho should live somewhere else anyway. Before phase 3 we ought to have a plan for what (if anything) we should do with those APIs.

The original `events.js` that handles SQS events is replaced by direct calls to the oauth db.

## Reviewing

I've attempted to break up the changes into smaller (relatively speaking) commits that are each test complete (100% passing). I recommend reviewing each commit individually in sequence. Since the bulk of the 16k+ additions are copies, I've left review comments where that isn't the case, but if you're curious or want to check for yourself, diffing the files from `fxa-oauth-server/lib` with `lib/oauth` will show what changed. For example you can do this in VSCode by selecting the two files in the explorer and right-click selecting 'Compare Selected' but that can get tedious. Other tools can diff whole folders.


packages/fxa-auth-server/bin/key_server.js
```diff
@@ -26,6 +26,7 @@ function run(config) {
   }
 
   const log = require('../lib/log')({ ...config.log, statsd });
+  require('../lib/oauth/logging')(log);
```
dannycoates: oauth-server uses a different logging style and initialization strategy. This welds the two together until it's convenient to refactor.

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -163,6 +164,7 @@ function run(config) {
     )
     .then(() => {
       return {
+        server,
```
dannycoates: Added for the test harness since there's no separate oauth-server wrapper anymore.

packages/fxa-auth-server/config/index.js
```diff
@@ -762,6 +762,381 @@ const conf = convict({
       },
     },
   },
+  oauthServer: {
```
dannycoates: all of oauth-server config is added here

packages/fxa-auth-server/fxa-oauth-server/lib/config/index.js
```diff
@@ -20,23 +20,14 @@ const envConfig = path.join(
   `${conf.get('env')}.json`
 );
 
-// This is sneaky and gross, but temporary.
-if (process.mainModule.filename.includes('key_server')) {
```
dannycoates: see, temporary ;)

packages/fxa-auth-server/lib/error.js
```diff
@@ -200,6 +201,9 @@ AppError.translate = function(request, response) {
   if (response instanceof AppError) {
     return response;
   }
+  if (OauthError.isOauthRoute(request && request.route.path)) {
+    return OauthError.translate(response);
+  }
```
dannycoates: a little shim for converting oauth errors. Unfortunately we've got overlapping `errno`s with different meanings between auth and oauth

packages/fxa-auth-server/lib/oauth/events.js
```diff
@@ -0,0 +1,70 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const config = require('../../config').getProperties();
+const db = require('./db');
+const P = require('../promise');
+const logger = require('./logging')('events');
+const Sink = require('fxa-notifier-aws').Sink;
+const HEX_STRING = require('./validators').HEX_STRING;
+
+let fxaEvents;
```
dannycoates: The file is here, but isn't wired up to anything yet.

packages/fxa-auth-server/lib/oauth/grant.js
```diff
@@ -0,0 +1,258 @@
...
+      hex(grant.userId),
+      grant.clientId
+    );
+    // To avoid mutating the input grant, create a
+    // copy and add the new property there.
+    grant = {
+      ...grant,
+      'fxa-subscriptions': capabilities,
+    };
+  }
```
dannycoates: In oauth-server this makes a request to the auth-server. Instead of making an request to itself I refactored it to get the data directly. One bit of hackery here is that `authdb` from up on line 55-58 that gets set by `lib/routes/index.js`. There's probably a nicer way to do it but it'd require a bigger change.

packages/fxa-auth-server/lib/oauth/logging/index.js
```diff
@@ -0,0 +1,29 @@
...
+    // probably a test
+    mozlog = {
+      logger: require('mozlog')({ app: 'unknown', level: 'critical' })(
+        nameOrLog
+      ),
+    };
+    return mozlog.logger;
+  }
+  mozlog = nameOrLog;
+};
```
dannycoates: this got shimmed to hell because of how different oauth-server's logging strategy is from auth-server. `/bin/key_server.js` sets the `mozlog` here before any other modules that depend on it get `require`d. gross but works.

packages/fxa-auth-server/lib/routes/index.js
```diff
@@ -48,6 +48,8 @@ module.exports = function(
   // The routing modules themselves.
   const defaults = require('./defaults')(log, db);
   const idp = require('./idp')(log, serverPublicKeys);
+  const oauthServer = require('../oauth/routes');
+  require('../oauth/grant').setDB(db);
```
dannycoates: as mentioned on `grant.js` this is where the auth db gets set

packages/fxa-auth-server/lib/routes/index.js
```diff
@@ -189,7 +191,7 @@ module.exports = function(
   defaults.forEach(r => {
     r.path = basePath + r.path;
   });
-  const allRoutes = defaults.concat(idp, v1Routes);
+  const allRoutes = defaults.concat(idp, v1Routes, oauthServer.routes);
```
dannycoates: where the rubber meets the road

packages/fxa-auth-server/lib/server.js
```diff
@@ -393,6 +394,9 @@ async function create(log, error, config, routes, db, oauthdb, translator) {
   );
   server.auth.strategy('oauthServerJWT', 'fxa-oauthServerJWT');
```
dannycoates: `oauthServerJWT` ought to be obsolete now and removed in phase 3, or now maybe...

packages/fxa-auth-server/lib/server.js
```diff
@@ -393,6 +394,9 @@ async function create(log, error, config, routes, db, oauthdb, translator) {
...
 
+  server.auth.scheme(authBearer.AUTH_SCHEME, authBearer.strategy);
+  server.auth.strategy(authBearer.AUTH_STRATEGY, authBearer.AUTH_SCHEME);
```
dannycoates: used by the oauth routes

packages/fxa-auth-server/test/lib/server.js
```diff
@@ -0,0 +1,66 @@
...
+
+  wrap.api = api;
+  wrap.config = config;
+  wrap.close = close;
+  return wrap;
+}
+
+module.exports.start = async function() {
+  const { server, close } = await createServer(testConfig);
+  return wrapServer(server, close);
```
dannycoates: this whole file is a bit of a hack to keep the oauth api tests working with minimal change

packages/fxa-auth-server/test/lib/util.js
```diff
@@ -36,7 +37,22 @@ function decodeJWT(b64) {
   };
 }
 
+function assertSecurityHeaders(res, expect = {}) {
```
dannycoates: copied from `fxa-oauth-server/test/lib/util.js` minus the bit about cache-control

packages/fxa-auth-server/test/test_server.js
```diff
@@ -124,3 +116,26 @@ TestServer.prototype.uniqueUnicodeEmail = function() {
...
+        handler: async function(request, h) {
+          const data = JSON.parse(Buffer.from(request.payload.token, 'hex'));
+          return h.response(data).code(data.code || 200);
+        },
+      },
+    ]);
+  } catch (e) {
+    console.error(e);
+  }
+}
```
dannycoates: woof. this was gnarly. this replaces what `oauth_helper.js` did. can't say I agree with it but I'll be damned if I'm gonna rewrite all the api tests to change it.

dannycoates:
This isn't totally done but since it's so much I'd like to open for review early. Remaining tasks are to port over newer oauth-server changes.

test/curl.sh
```diff
@@ -76,7 +76,7 @@ check 127.0.0.1:9000
 check 127.0.0.1:9001 404
 
 # oauth
-check 127.0.0.1:9010
+# check 127.0.0.1:9010
```
shane-tomlinson: Maybe just remove these lines?

packages/fxa-auth-server/test/test_server.js
shane-tomlinson: Can you leave a note about why you patch the verify route to avoid that context being lost?

packages/fxa-auth-server/test/remote/attached_clients_tests.js
```diff
@@ -94,11 +91,13 @@ describe('attached clients listing', function() {
     );
     allClients = await client.attachedClients();
     assert.equal(allClients.length, 2);
-    assert.equal(allClients[0].sessionTokenId, mySessionTokenId);
-    assert.equal(allClients[0].deviceId, device.id);
-    assert.equal(allClients[1].refreshTokenId, refreshTokenId);
-    assert.equal(allClients[1].deviceId, device2.id);
-    assert.equal(allClients[1].name, 'test device');
+    const one = allClients.findIndex(c => c.name === 'test device');
```
shane-tomlinson: wat?

packages/fxa-auth-server/lib/server.js
shane-tomlinson: Seems fine to defer to phase 3, though could cause less churn in cloudops deployment if the environment variables used to support this are removed.

packages/fxa-auth-server/lib/oauth/events.js
shane-tomlinson: Oh, it should be possible to skip the whole SQS queue to update the OAuth DB now, shouldn't it? Whenever a user is removed, call the OAuth DB's removeUser, and whenever a password reset or change happens, call removePublicAndCanGrantTokens.

packages/fxa-auth-server/lib/oauth/grant.js
```diff
@@ -226,23 +238,20 @@ exports.generateAccessToken = async function generateAccessToken(grant) {
...
-    //   scope: 'profile:subscriptions',
-    //   uid: hex(grant.userId),
-    // });
-    // // To avoid mutating the input grant, create a
-    // // copy and add the new property there.
-    // grant = {
-    //   ...grant,
-    //   'fxa-subscriptions': subscriptions,
-    // };
+    const capabilities = await determineClientVisibleSubscriptionCapabilities(
```
shane-tomlinson: YES!

packages/fxa-auth-server/scripts/activate-new-signing-key.js
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env node
```
shane-tomlinson: @jrgm and @jbuck will need to update their signing key scripts to point to the new location.

packages/fxa-auth-server/test/lib/server.js
shane-tomlinson: Can you leave a comment about what this is doing and how it's used? It's obviously an API wrapper, but apart from that I'm not totally sure what's going on.

packages/fxa-auth-server/lib/oauth/logging/index.js
shane-tomlinson: In a subsequent PR, would it be possible to unify the strategies?

packages/fxa-auth-server/test/remote/attached_clients_tests.js
dannycoates: I was having this test fail intermittently because the order of `allClients` was reversed. idk why exactly

packages/fxa-auth-server/scripts/activate-new-signing-key.js
dannycoates: It's also in still in the old location for now.

packages/fxa-auth-server/lib/oauth/logging/index.js
dannycoates: That's absolutely the plan

packages/fxa-auth-server/lib/oauth/events.js
dannycoates: Thanks! Yeah, I forgot that auth-server was the source of these events. I'll see if making that change is practical in this phase

dannycoates:
>  I had to update local.json to point to the new OAuth server

Thanks. I updated the PR notes

dannycoates:
Ok, y'all, modulo the comments @shane-tomlinson asked for this is ready for a full review

packages/fxa-auth-server/config/key.json
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
dannycoates: Thanks for bringing this up. I'm gonna leave the names for now but plan on changing in phase 3. It also reminds me to fixup the gen_keys scripts...

PR #2770, opened 2019-10-07
feat(docs): Add a JWT access token doc.

shane-tomlinson:
I'm not sure if this is the right place for this
or whether this should be an ADR.


lmorchard:
Might be worth an ADR if we can add other alternatives considered & rejected in favor of this solution. Otherwise, it might be weird to shoehorn into the ADR form. Probably fine as just a standalone doc, IMO

PR #2782, opened 2019-10-07
Enable 100% signin codes for sync users

vbudhram:
Targeted as a point release for Train 147.


packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -90,10 +90,7 @@ module.exports = class TokenCodeGroupingRule extends BaseGroupingRule {
       }
 
       if (this.bernoulliTrial(syncRolloutRate, subject.uniqueUserId)) {
-        return this.uniformChoice(
-          ['treatment-code', 'treatment-link'],
-          subject.uniqueUserId
-        );
+        return this.uniformChoice(['treatment-code'], subject.uniqueUserId);
```
shane-tomlinson: If the Sync rollout rate is 1 and there is only one possible group, you could get rid of lines 92-94 and just do a `return 'treatment-code';`

vbudhram:
This is failing on the missing privacy links, merging.

PR #2794, opened 2019-10-08
fix(image): fix #2668 - Load Firefox browser logo correctly on iOS 13

### LINKED ISSUE #2668: Firefox Browser logo is displayed incorrectly on iOS
See [this report on reddit](https://www.reddit.com/r/firefox/comments/dbiq38/spotted_a_different_new_firefox_icon_in_firefox/): The Firefox browser logo looks incorrect on iOS.

The glitched file appears to be [graphic-logo-firefox.svg](https://github.com/mozilla/fxa/blob/f562748de2e1a35b75b020a5e8076c70fc59c5bd/packages/fxa-content-server/app/images/trailhead/graphic-logo-firefox.svg). Loading it in iOS 13.1 (17A844) Safari displays as follows:

IMG_2106

After removing the `fx` and `fy` attributes from `<radialGradient id="radial5"`, it displays correctly:

IMG_2107

LZoog:
Fixes #2668, and ran through `svgo` for size optimization.

This bug is happening on at least Safari and Firefox on **iOS 13 only**. This was not an issue on iOS 12.x.

It does appear the best solution I could find was provided in the Reddit thread linked in the issue, which was to remove the `fx` and `fy` attributes on what was named `radial5`. iOS seems to have an issue with this `radialGradient` that I could not identify (the properties matched other `radialGradient` properties) - if you take the gradient completely out, it renders the same as what's shown in the bug.

@ryanfeeley I was very careful to compare a before and after of this change, and do not see a visual difference.
image

This is confirmed working on iOS 13 Safari & Firefox, and Mac OS Firefox. I would like to confirm this works on Android before merging.

---

Another note - in that Reddit thread, this image was linked: https://cdn.sso.mozilla.com/nlx/02cf955/images/firefox.svg

This needs to be updated as well. Removing `fx` and `fy` from `radialGradient id="g"` fixes the issue there, but I did notice a visual difference between this SVG and the one above. as there is a small section that is not rounded in the fox in the lower left-hand corner. As @meandavejustice mentioned yesterday, it would be nice to fix this problem wherever we use this logo... Should I file a bug in Bugzilla for this iOS issue with the image inconsistency mentioned, or is there another repo I can file this issue?


shane-tomlinson:
The test failure is because of the failing links on the privacy page, feel free to merge.

shane-tomlinson:
> Should I file a bug in Bugzilla for this iOS issue with the image inconsistency mentioned, or is there another repo I can file this issue?

@ryanfeeley - who is best to talk to about the original artifact? Horlander?

ryanfeeley:
Yes @shorlander can help on the logo front. And seeing as it's changing, now is a good time to loop him in.

ryanfeeley:
@shane-tomlinson @LZoog We can now use the new logo for the Firefox client (different than the master brand that we use on accounts website). All versions are available here. https://drive.google.com/drive/folders/1yryPGL0dwyPrtqGL1oZxctXiEsjtOICq

Issue #996, opened 2019-05-01
Ensure we're monitoring client-side performance data for new Subscriptions pages


clouserw:
We can use re:dash for this since that is where our other performance data is going.  I just want to make sure the new pages are instrumented and we can see any performance concerns.

philbooth:
Specifically, the implementer should tweak this bit of code in the content server, so that the new views get their own performance category instead of being lumped in with `other`:

https://github.com/mozilla/fxa/blob/dcf8db2529513409bfa767840ef56b281f9f1b18/packages/fxa-content-server/server/lib/flow-event.js#L94

The `other` category has pretty bad performance due to the `/settings` view, which we know to be slow. It would make sense to analyse these numbers without being dragged down by that.

philbooth:
> Specifically, the implementer should tweak this bit of code in the content server...

Ignore me, just remembered that these aren't going to be hosted in the content server.

chenba:
I can think of three new pages: support form, payments forms, and subscription management.

Support form is on the content server, and the others are on the payments server.

chenba:
Since https://github.com/mozilla/fxa/pull/2481 added the flow event mixin to the support form view, performance metrics should just show up in Redshift once the Support Form is in prod.  Although I think they'd go into the `flow.performance.other*` types.

chenba:
hahah never you literally answered my question above already!!  sorry!

chenba:
https://github.com/mozilla/fxa/pull/2574 removed the auth/other categories. 🎉

PR #2801, opened 2019-10-08
[HOLD] feat(metrics): add flow perf metrics to payments server pages

chenba:
Part of #996


packages/fxa-payments-server/package.json
```diff
@@ -14,7 +14,7 @@
     "build": "react-scripts build",
     "eject": "react-scripts eject",
     "test": "npm-run-all test:*",
-    "test:frontend": "react-scripts test --coverage --verbose",
+    "test:frontend": "react-scripts test --coverage --verbose --transformIgnorePatterns 'node_modules/(?!speed-trap)/'",
```
chenba: Jest wants this particular module transpiled.  Moreover, adding that option into `jest.config.js` did not have any effect.  😞

packages/fxa-payments-server/src/lib/flow-metrics.ts
```diff
@@ -0,0 +1,80 @@
...
+    utm_campaign: NOT_REPORTED_VALUE,
+    utm_content: NOT_REPORTED_VALUE,
+    utm_medium: NOT_REPORTED_VALUE,
+    utm_source: NOT_REPORTED_VALUE,
+    utm_term: NOT_REPORTED_VALUE,
+  };
+
+  // This is not an Action insofar that it has no bearing on the app state.
+  window.navigator.sendBeacon(
+    `${config.servers.content.url}/metrics`,
```
6a68: Looks like this PR uses the content server to submit metrics? I'm hoping to come up with an alternative in https://github.com/mozilla/fxa/issues/2425, not sure if this should land in the meantime. @shane-tomlinson thoughts?

shane-tomlinson: I suppose, what is the long term plan and is this approach on the path to that plan? I have wondered if reporting to the content server is a reasonable solution too, @philbooth pointed out we'd have to worry about naming collisions. What will these metrics look like when sent to re:dash and amplitude, will they be under their own namespace, or do we just depend on the service being 'payments-server'?

If sending to the content server, the [POST /metrics endpoint](https://github.com/mozilla/fxa/blob/1cb0d0a4e8104b700a78a414fcc2ecf96e0fd6fc/packages/fxa-content-server/server/lib/routes/post-metrics.js#L163) will need CORS enabled. Right now we only expect requests from the content server, which is on the same domain. To avoid displaying the full list of domains that are allowed to use CORS, we'd want to do something like [GET /metrics-flow](https://github.com/mozilla/fxa/blob/1cb0d0a4e8104b700a78a414fcc2ecf96e0fd6fc/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L33), where the origin is compared against an allow list.

All that said, it would be worthwhile to explore what is needed to comple Phil's PR that adds a [POST /metrics endpoint](https://github.com/mozilla/fxa/pull/2734/files) to the payments server. There would be more Ops work to ensure those events make their way to Amplitude, but might be a reasonable tradeoff to simplify the mental model of how metrics work.

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -155,7 +182,7 @@ module.exports = () => {
             return proxyResData;
           }
           const body = proxyResData.toString('utf8');
-          return injectHtmlConfig(body, CLIENT_CONFIG, {});
+          return injectHtmlConfig(body, CLIENT_CONFIG, {}, flowMetricsData());
```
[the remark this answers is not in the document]
shane-tomlinson: Also, have you considered just calling the content server's [GET /metrics-flow](https://github.com/mozilla/fxa/blob/1cb0d0a4e8104b700a78a414fcc2ecf96e0fd6fc/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L83) endpoint like other services do? We could add a new [form_type](https://github.com/mozilla/fxa/blob/1cb0d0a4e8104b700a78a414fcc2ecf96e0fd6fc/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L66) and doing so would allow us to avoid the [proliferation of the HMAC key](https://github.com/mozilla/fxa/pull/2801/files#diff-36e2db7c1336ac4eec3760f309b6f90fR118).

packages/fxa-payments-server/package.json
chenba: 👍

packages/fxa-payments-server/src/lib/flow-metrics.ts
chenba: This PR is strictly for flow performance metrics.  That is, `flow.performance.*` events written to `stderr`, then piped through https://github.com/mozilla/fxa-activity-metrics/, and resulting in https://sql.telemetry.mozilla.org/dashboard/fxa-performance.

That being said, there is nothing preventing us from collecting them in payments server.  I simply want to avoid duplication.

As for log pipeline, I suppose that isn't difficult per se?  I recently learned that the content server and payments server are on different cloud services.  But perhaps the logs are shipped to a central place first?  Anyways, it's probably pretty easy for Ops, but it's not nothing.

Namespace remains a (minor) concern as long as we collect these metrics into the same DB.  I've prefix'd `payments` to the metric type name, i.e. all the type names from this patch will begin with `flow.performance.payments`.

I thought CORS wasn't an issue since we were already allowing payments server to post CSP reports to content server.  But it doesn't sound like that's the case.  I shall make changes.

6a68: @chenba I've sent you an invite to a payments metrics meeting Shane and I had scheduled for Thursday. We can figure out a plan then :+1:

packages/fxa-payments-server/package.json
```diff
@@ -14,8 +14,8 @@
     "build": "react-scripts build",
     "eject": "react-scripts eject",
     "test": "npm-run-all test:*",
-    "test:frontend": "react-scripts test --coverage --verbose",
-    "test:server": "jest --coverage --verbose --config server/jest.config.js",
+    "test:frontend": "react-scripts test --coverage --verbose --transformIgnorePatterns 'node_modules/(?!speed-trap)/'",
```
shane-tomlinson: I have published speed-trap 0.0.10 so you shouldn't have to special case this anymore.

packages/fxa-payments-server/package.json
```diff
@@ -15,7 +15,7 @@
     "eject": "react-scripts eject",
     "test": "npm-run-all test:*",
     "test:frontend": "react-scripts test --coverage --verbose",
-    "test:server": "jest --coverage --verbose --config server/jest.config.js",
+    "test:server": "FLOW_ID_KEY=testo jest --coverage --verbose --config server/jest.config.js",
```
lmorchard: This looks like something that should get added to mocks in server tests. Shouldn't need to configure the environment for jest

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -85,6 +91,14 @@ export const Product = ({
     error: false,
   });
 
+  // This effect is to simulate a componentDidMount.
+  useEffect(() => {
+    if (config.metrics.flow.enabled) {
+      FlowMetrics.logLoadedEvent('product', config, getScreenInfo);
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
lmorchard: What deps are getting flagged by eslint to require the disabled rule? Looks like `config`, `FlowMetrics`, and `getScreenInfo` - those should be okay to list here. Should mean that this effect gets run on initial render or whenever one of those deps changes (which should basically be never)

packages/fxa-payments-server/package.json
chenba: @lmorchard Thanks for looking at this PR.  It was created with the intention of sending the metrics directly to content server.  Now that's no longer the case; we'll build on the server code that's been added for Amplitude metrics.  And I think the request from the frontend will also be different and simpler.

tl;dr I'll add a [HOLD] to the PR title now.

chenba:
This is on hold at the moment.  It will be changed to send the metrics to the payments server, with a differently shaped payload.

chenba:
Closing this in favor of https://github.com/mozilla/fxa/pull/3071.

Issue #2650, opened 2019-09-27
subscriptions.firefox.com emits duplicate headers


clouserw:
From #1128:

> HSTS is in place but observatory complains about it because, well, the site returns 3 HSTS headers... one would be sufficient :)

Let's drop the extra headers

6a68:
@jbuck Hey, I think something in the cloudops config might be appending additional HSTS headers. I'm seeing 2 `vary: Accept-Encoding` and 3 `strict-transport-security` headers when I curl subscriptions.firefox.com, but I've tested the application code, and I think it only sets one of each. Could you poke around at the openresty config to see if HSTS is getting appended even if the header already exists?

6a68:
Ah yeah, looking at the nginx-configmap.yaml for the payments project, nginx indeed sets one of the 3 headers:

`add_header Strict-Transport-Security "max-age=31536000" always;`

Not sure if I should drop the link in a public bug or not 🤷‍♂️, continuing to dig for the other dupe.

6a68:
Found one instance of `vary: Accept-Encoding`, looking in the same nginx config file, I see the `gzip_vary` config value is set.

Now digging in the app code for the same header.

6a68:
I think the other `vary: Accept-Encoding` and the other HSTS might be getting copied over by the `express-http-proxy`, but need to dig further to be sure it's actually used in production.

jrgm:
I added https://github.com/mozilla-services/cloudops-infra/pull/1465 to address a second (of 3) HSTS headers being sent in our kubernetes production config. I've also added a PR https://github.com/mozilla/fxa/pull/2802 for payments-server to not send HSTS if disabled. With those I can configure cloudops-infra to not send the third header when HSTS_ENABLED is set to false.

jrgm:
Okay, the above two PRs, plus this one (https://github.com/mozilla-services/cloudops-infra/pull/1467) have merged. The kubernetes related changes are in production now, but to get the app to stop sending STS we'd need either a point release to v1.147, or some tag that includes the change in #2802.

clouserw:
The tag we make today will include those changes

PR #2802, opened 2019-10-09
fix(payments): add config to disable HSTS to avoid multiple headers in production

jrgm:
Because in production, under kubernetes we run iprep-nginx in front of the nodejs app server, and it unilaterally adds its own HSTS headers, the resulting response has >1 `Strict-Transport-Security` headers (#2650). If this merges, I'll configure production to `HSTS_ENABLED=false`, with the net result that only one HSTS header will be sent (although without the includeSubmains assertion).

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -45,6 +45,12 @@ const conf = convict({
     env: 'NODE_ENV',
     format: ['development', 'production', 'test'],
   },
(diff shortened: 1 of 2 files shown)


jrgm:
> is a corresponding PR needed to cloudops-deployment to disable HSTS sending in stage/prod?

So, this is is more like "when running under kubernetes, always send HSTS from nginx and disable the app+helmet send of HSTS".

PR #2803, opened 2019-10-09
use custom docker image for content-server ci tests

dannycoates:
Move some install tasks up into a docker image and moved some tasks around to save a few minutes of CI time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -331,6 +331,11 @@ workflows:
           module: fxa-support-panel
           requires:
             - install
+      - build-module:
+          name: fxa-circleci
+          module: fxa-circleci
+          requires:
+            - install
       - fxa-content-server:
           requires:
             - install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   nodejs:
     docker:
-      - image: circleci/node:10
+      - image: dannycoates/fxa-circleci
 commands:
   npm-install:
     parameters:
@@ -11,16 +11,7 @@ commands:
         type: string
         default: "dependencies"
     steps:
-      - restore_cache:
-          keys:
-            - v2-<< parameters.module >>-{{ checksum "package.json" }}
-            # fallback to using the latest cache if no exact match is found
-            - v2-<< parameters.module >>-
       - run: npm ci
-      - save_cache:
-          paths:
-            - node_modules
-          key: v2-<< parameters.module >>-{{ checksum "package.json" }}
 
 jobs:
   install:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -3,24 +3,15 @@ version: 2.1
 executors:
   nodejs:
     docker:
-      - image: circleci/node:10
+      - image: dannycoates/fxa-circleci
```
dannycoates: This image should move somewhere else, but it's been useful for testing, and probably ok in the short term. IDK the best place for it. It's just a `Dockerfile`. It _could_ go in the fxa repo, but then would it have a ci step... 🤔

.circleci/test-content-server.sh
```diff
@@ -33,46 +33,25 @@ function test_suite() {
...
 
-  sudo apt-get install -y python-setuptools python-dev build-essential graphicsmagick &> /dev/null
-  sudo easy_install pip &> /dev/null
-  sudo pip install mozdownload mozinstall &> /dev/null
-
-  if [[ ! $(which rustup) ]]; then
-    curl https://sh.rustup.rs -sSf | sh -s -- -y
-    PATH=$PATH:$HOME/.cargo/bin
-  fi
-
```
dannycoates: this is prepackaged in the docker image

.circleci/config.yml
```diff
@@ -3,24 +3,15 @@ version: 2.1
...
-      - restore_cache:
-          keys:
-            - v2-<< parameters.module >>-{{ checksum "package.json" }}
-            # fallback to using the latest cache if no exact match is found
-            - v2-<< parameters.module >>-
       - run: npm ci
-      - save_cache:
-          paths:
-            - node_modules
-          key: v2-<< parameters.module >>-{{ checksum "package.json" }}
```
dannycoates: these caches weren't really doing anything

.circleci/config.yml
```diff
@@ -115,25 +106,11 @@ jobs:
...
-    steps:
-      - attach_workspace:
-          at: ~/.
-
-      - run: ../../.circleci/install-content-server.sh
-
-      - persist_to_workspace:
-          root: ~/.
-          paths:
-            - project/*
```
dannycoates: turns out to be faster to just do this in the test job (multiple times) 🤷‍♂️

.circleci/config.yml
```diff
@@ -149,42 +126,23 @@ jobs:
...
-      - image: redis
-      - image: memcached
-      - image: pafortin/goaws
-      - image: circleci/mysql:5.7
-      - image: mozilla/fxa-email-service
-        environment:
-          NODE_ENV: dev
-          FXA_EMAIL_ENV: dev
-          FXA_EMAIL_LOG_LEVEL: debug
-          RUST_BACKTRACE: 1
```
dannycoates: build doesn't need these

circleci_servers.json
```diff
@@ -0,0 +1,165 @@
+{
```
dannycoates: leaves out the docker services, redis, mysql, etc.

.circleci/config.yml
bbangert: A separate fxa repo with an auto-updated docker hub build?

.circleci/config.yml
bbangert: When I was watching one of the runs, I noticed their network is so slow, that it can be faster to do some work from scratch rather than wait for large cache stores to move between wherever they're moving between.

bbangert:
Actually, after thinking about it a bit more, CircleCI will auto-build an image for one of our repo's as well that is usable for this.

dannycoates:
> CircleCI will auto-build an image for one of our repo's as well that is usable for this.

got a reference for this? I'm not finding anything

bbangert:
@jrgm I know we the current CI builds docker images we use, how would we setup a new repo to have the containers available publicly?

dannycoates:
> I know Circle CI builds docker images we use, how would we setup a new repo to have the containers available publicly?

I went ahead and added it to this repo under `packages/fxa-circleci` and have ci build/deploy it like any other. Once the first one is up on docker hub we can switch the image name in config.yml.

jrgm:
Credentials have been added.

PR #2816, opened 2019-10-09
chore(style): add icon to subscription success page

johngruen:


packages/fxa-payments-server/src/images/fpn.svg
```diff
@@ -0,0 +1,30 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="96" height="96" viewBox="0 0 96 96">
```
LZoog: I would love for the team to start getting into the habit of running SVGs through something called `svgo` which compresses/optimizes SVGs for us, similar to sending a large PNG through tinyPNG. There's a command line tool you can install with `npm i -g svgo` and then you can run `svgo /path/to/file`, or you can go to https://jakearchibald.github.io/svgomg/ and open the SVG there. :) I've compressed a few in our repo and it's reduced the file size up to 50-80%.

Side note: we _could_ explore automating this, but every now and then (like 1 - 3% of the time?) this will strip out a mask or something that's needed for it to display properly so I always double check it looks right before committing.

Your PR looks good either way though. To reiterate what I said in Slack / for future reference, we'll probably end up doing something like "className={`graphic-${productName}`}" (or whatever the variable for the name is here) on the div when we have multiple products to make this dynamic.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -28,6 +28,7 @@ export const SubscriptionRedirect = ({
   return (
     <div className="subscription-ready">
       <h2>Your subscription is ready</h2>
+      <div className="graphic graphic-fpn"></div>
```
LZoog: Also, if you're going to go ahead and compress that SVG, you can take out the `graphic` class here too since we're currently only styling `graphic-fpn`.

lmorchard:
Looks like some of the CI failures here are unrelated. But a few look like CSS linting issues:
```
src/index.scss
  135:33  warning  Quotes around URLs are required  url-quotes

src/routes/Product/SubscriptionRedirect/index.scss
  10:29  error  Strings must use single quotes  quotes

src/routes/Subscriptions/index.scss
  73:31  warning  Quotes around URLs are required  url-quotes
```

packages/fxa-payments-server/src/images/fpn.svg
johngruen: I think the intent is to eventually just have a generic "check" icon instead of doing something per product, but i'll svgo the svg!

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
johngruen: `.graphic` is inherited from the content server styles

PR #2817, opened 2019-10-09
fix(sync): send an empty sync object for sign-in flows 

vladikoff:
@vbudhram r?

cc @mhammond

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync-channel.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync-channel.js
@@ -300,6 +300,10 @@ const FxSyncChannelAuthenticationBroker = FxSyncAuthenticationBroker.extend(
           offeredEngines: loginData.offeredSyncEngines,
           declinedEngines: loginData.declinedSyncEngines,
         };
+      } else if (this.relier.get('service') === 'sync') {
+        // if service is already forced to be Sync, then for sign-in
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -300,6 +300,10 @@ const FxSyncChannelAuthenticationBroker = FxSyncAuthenticationBroker.extend(
           offeredEngines: loginData.offeredSyncEngines,
           declinedEngines: loginData.declinedSyncEngines,
         };
+      } else if (this.relier.get('service') === 'sync') {
+        // if service is already forced to be Sync, then for sign-in
+        // attempts we tell Desktop to just enable sync
+        loginData.services.sync = {};
```
vladikoff: This will make sure desktop enables Sync after a sign in

vbudhram: Tested this locally with Nightly and it doesn't seem to enable sync after reconnecting?

Screen Shot 2019-10-09 at 8 32 02 PM

vladikoff: Yeah requires a desktop patch, cc @mhammond

mhammond:
See [bug 1587691](https://bugzilla.mozilla.org/show_bug.cgi?id=1587691)

PR #2823, opened 2019-10-10
feat: add queue prefix config and fix login validation

### LINKED ISSUE #2819: [fxa-event-broker] Fix validation of service key
In event-broker, login messages are required to contain a service key. This key doesn't always exist as some services (sync) don't emit one.

bbangert:
Because:

* The queue prefix is not configurable and FxA reuses client ID's in
  the dev/stage environment.
* Login messages don't always include a service property.

This commit:

* Allows the queue prefix to be configured.
* Makes the service property optional for login messages.

Closes #2819


jrgm:
@vbudhram - I've made the requisite config change

PR #2824, opened 2019-10-10
fix(payments): redirect to settings if customer has no subscriptions

### LINKED ISSUE #2372: [payments server] Subscription Management Page Should Redirect to Settings If the User Doesn't Have Any Subscriptions
While the Settings page does not display the button for the subscription management page when a user does not have any subscriptions, if the user know the URL for the subscription management page, they can still land there.

mbStavola:
fixes #2372
- remove outdated comment
- add redirect to settings
- update tests

### DIFF
--- a/packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
@@ -72,7 +72,7 @@ export type SubscriptionsProps = {
   plans: PlansFetchState;
   customer: CustomerFetchState;
   subscriptions: SubscriptionsFetchState;
-  customerSubscriptions: Array<CustomerSubscription>;
+  customerSubscriptions: Array<CustomerSubscription> | null;
   fetchSubscriptionsRouteResources: Function;
   cancelSubscription: Function;
   cancelSubscriptionStatus: CancelSubscriptionFetchState;
@@ -213,9 +213,20 @@ export const Subscriptions = ({
     );
   }
 
+  // If the customer has no subscriptions, redirect to the settings page
+  if (
+    (customerSubscriptions && customerSubscriptions.length === 0) ||
+    (customer.error &&
+      customer.error.errno === AuthServerErrno.UNKNOWN_SUBSCRIPTION_CUSTOMER)
+  ) {
+    const SETTINGS_URL = `${config.servers.content.url}/settings`;
+    navigateToUrl(SETTINGS_URL);
+    return <LoadingOverlay isLoading={true} />;
+  }
+
   return (
     <div className="subscription-management" onClick={onAnyClick}>
-      {cancelSubscriptionStatus.result !== null && (
+      {customerSubscriptions && cancelSubscriptionStatus.result !== null && (
         <CancellationDialogMessage
           {...{
             subscription: cancelSubscriptionStatus.result,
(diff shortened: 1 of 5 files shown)


packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -158,6 +158,17 @@ export const Subscriptions = ({
     );
   }
 
+  // If the customer has no subscriptions, redirect to the settings page
+  if (
+    customerSubscriptions.length === 0 ||
+    (customer.error &&
+      customer.error.errno === AuthServerErrno.UNKNOWN_SUBSCRIPTION_CUSTOMER)
```
chenba: This condition will never be evaluated since it's already up in the `if` on line 146.  I think we can remove that `if` block above.

chenba:
I'm seeing some sort of timing issue where _sometimes_ I'd be redirect back to the settings page even when I have an active subscription.  I don't think it's caused by this patch.  But let's not merging this PR until we figure that out.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
mbStavola: The other `if` block checks if there's an error, but _not_ an `UNKNOWN_SUBSCRIPTION_CUSTOMER` error in order to show an error message. 

From what I understand, that specific error code is another way to say the user has no subscriptions-- maybe a bit redundant if `customerSubscriptions` will always be 0 in that case. 

I can remove it if you want.

chenba: (Sorry for the late response; been busy.)

You're right, I misread that comparison in the `if` above.  Let's keep this as it is then.  Thanks again!

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -158,6 +158,17 @@ export const Subscriptions = ({
     );
   }
 
+  // If the customer has no subscriptions, redirect to the settings page
+  if (
+    customerSubscriptions.length === 0 ||
```
chenba: Ah so, the default state of `customerSubscriptions` is an empty array (https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/store/selectors.ts#L33).  That is, when Subscriptions is initially mounted, this is _always_ true.  That's why I get redirected even when I have a subscription.

I think we could change the default state to `null`.  We do need to update any code that expects `customerSubscriptions` to be an array.

mbStavola: Good find! Makes total sense.

I can update the type for `customerSubscriptions` to be `Array<CustomerSubscription> | null`, but I do worry about knock on effects for that sort of change, especially since it seems like `--strictNullChecks` isn't turned on for this project.

I'm curious about [this block](https://github.com/mozilla/fxa/pull/2824/files#diff-f2c1cf4bcabdeb9f1594a4f419931657R108-R113). I'd have figured that on the initial mount the `loading` state for these would be `true` and we wouldn't reach the redirect block. Does `loading` not get set to true until after the component has rendered the first time? If that's not the case it seems a bit strange.

Would it be valuable to look down that path?

chenba: The `loading` states won't be true until `fetchSubscriptionsRouteResources` is called (https://github.com/mozilla/fxa/pull/2824/files#diff-f2c1cf4bcabdeb9f1594a4f419931657R100).  So there's a race condition between that and the `if` here.

Allowing `null` make sense to me.  Before data is fetched, the state is really unknown, not "no subscriptions".  `customerSubscriptions` is only used a few place, it should not be a lot of work to update those lines with a null check.

mbStavola: Okay cool, looks like that was it! After switching to null, I don't see the improper redirects anymore when I have a subscription to 123done.

chenba:
@mbStavola Hey sorry, I forgot to ask you to squash everything into one commit.  Once that's done, we'll merge.  Thanks!

mbStavola:
Squashed and rebased-- sorry it took a few days to get to!

PR #2827, opened 2019-10-10
chore(email-first): Remove legacy signin/signup flows.

shane-tomlinson:


shane-tomlinson:
Closing this, breaking it out into smaller PRs

PR #2835, opened 2019-10-10
feat(oauth): allow Firefox Monitor to use prompt=none features

vladikoff:
Done to build out features mentioned in https://docs.google.com/presentation/d/1BdGjT3U74ZwwFUy5z-UjBAKQJYQsb9zHDe7TJ80-gnQ/edit#slide=id.g62842b55a5_0_89 

@shane-tomlinson r?

### DIFF
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -409,7 +409,11 @@ const conf = (module.exports = convict({
       },
       enabled_client_ids: {
         // 123done enabled for functional tests, 321done is not.
-        default: ['dcdb5ae7add825d2', '7f368c6886429f19'],
+        default: [
+          'dcdb5ae7add825d2',
+          '7f368c6886429f19',
+          '802d56ef2a9af9fa', // Firefox Monitor
+        ],
         doc: 'client_ids for which prompt=none is enabled',
         env: 'OAUTH_PROMPT_NONE_ENABLED_CLIENT_IDS',
         format: Array,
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -409,7 +409,11 @@ const conf = (module.exports = convict({
       },
       enabled_client_ids: {
         // 123done enabled for functional tests, 321done is not.
-        default: ['dcdb5ae7add825d2', '7f368c6886429f19'],
+        default: [
+          'dcdb5ae7add825d2',
+          '7f368c6886429f19',
+          '5882386c6d801776', // Firefox Desktop
```
vbudhram: Should this be done as a cloudops deployment?

vladikoff: they don't use that ENV var yet (I couldn't find it)

cc @jrgm

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -409,7 +409,11 @@ const conf = (module.exports = convict({
       },
       enabled_client_ids: {
         // 123done enabled for functional tests, 321done is not.
-        default: ['dcdb5ae7add825d2', '7f368c6886429f19'],
+        default: [
+          'dcdb5ae7add825d2',
+          '7f368c6886429f19',
+          '802d56ef2a9af9fa', // Firefox Monitor
```
shane-tomlinson: Yeah, can we get this in cloudops-deployment instead?

PR #2843, opened 2019-10-10
chore(npm): changed 'npm start' to run 'pm2 start mysql_servers.json'

dannycoates:
Moved `fxa-dev-launcher` to `npm run start-firefox` and made `npm start` run `pm2 start mysql_servers.json`

Using `npm start` again on an already running cluster will restart all the services.

Also extended the pm2 `kill_timeout` for docker processes to make restarts behave better.

### DIFF
--- a/mysql_servers.json
+++ b/mysql_servers.json
@@ -1,6 +1,67 @@
 {
   "apps": [
     {
+      "name": "MySQL server PORT 3306",
+      "script": "_scripts/mysql.sh",
+      "max_restarts": "1",
+      "min_uptime": "2m",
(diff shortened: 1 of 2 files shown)


mysql_servers.json
```diff
@@ -1,5 +1,66 @@
 {
   "apps": [
+    {
+      "name": "MySQL server PORT 3306",
```
lmorchard: Did you mean to re-order all the services?

dannycoates: yes, since some of the docker process take longer to start I moved them up

PR #2844, opened 2019-10-10
feat(metrics): Propagate flow_* params from content to payments server

shane-tomlinson:
This fetches the flow_begin_time and flow_id out of a view's metrics and propagates them whenever redirecting to the payments server as the flow_begin_time and flow_id query params.

There were some indentation fixes, so this might be easier to review with ?w=1

issue #2425
issue #2801

@6a68, @chenba - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/payment-server.js
+++ b/packages/fxa-content-server/app/scripts/lib/payment-server.js
@@ -4,12 +4,22 @@
 
 'use strict';
 
+import Url from './url';
+
 const PaymentServer = {
+  /**
+   * Navigate to the payments server after flushing metrics.
+   *
+   * @param {Object} view
+   * @param {Object} subscriptionsConfig
+   * @param {String} redirectPath
+   * @param {Object} [queryParams={}]
+   */
   navigateToPaymentServer(
     view,
     subscriptionsConfig,
     redirectPath,
-    queryParams
+    queryParams = {}
   ) {
     const {
       managementClientId,
@@ -17,15 +27,13 @@ const PaymentServer = {
       managementTokenTTL,
       managementUrl,
     } = subscriptionsConfig;
-    const queryString =
-      typeof queryParams === 'object' &&
-      Object.keys(queryParams)
-        .filter(k => queryParams[k] !== null && queryParams[k] !== '')
-        .map(
-          k => `${encodeURIComponent(k)}=${encodeURIComponent(queryParams[k])}`
-        )
-        .join('&');
-    const qs = queryString.length ? `?${queryString}` : '';
+
+    const {
+      deviceId,
+      flowBeginTime,
+      flowId,
+    } = view.metrics.getFlowEventMetadata();
+
     return view
       .getSignedInAccount()
       .createOAuthToken(managementClientId, {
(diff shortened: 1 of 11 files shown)


packages/fxa-content-server/app/scripts/lib/payment-server.js
```diff
@@ -4,38 +4,54 @@
...
-    queryParams
+    queryParams = {}
   ) {
     const {
       managementClientId,
       managementScopes,
       managementTokenTTL,
       managementUrl,
     } = subscriptionsConfig;
-    const queryString =
```
shane-tomlinson: The Url module does this, so I removed the duplicate logic here.

packages/fxa-content-server/app/scripts/lib/url.js
```diff
@@ -114,7 +114,11 @@ export default {
     const params = [];
     for (const paramName in obj) {
       const paramValue = obj[paramName];
-      if (typeof paramValue !== 'undefined' && paramValue !== null) {
+      if (
+        typeof paramValue !== 'undefined' &&
+        paramValue !== null &&
+        paramValue !== ''
```
shane-tomlinson: I did have to update this to filter out empty param values though.

packages/fxa-content-server/app/tests/spec/lib/payment-server.js
```diff
@@ -32,12 +32,18 @@ describe('lib/payment-server-redirect', () => {
...
+          flowId: 'foo',
+        })),
+      },
     };
   });
 
   it('redirects as expected', () => {
     const REDIRECT_PATH = 'example/path';
-    PaymentServer.navigateToPaymentServer(
+    return PaymentServer.navigateToPaymentServer(
```
shane-tomlinson: These tests weren't returning the promises and were actually failing, so had to do a bit of tangential repair.

packages/fxa-content-server/app/tests/spec/views/subscriptions_management_redirect.js
```diff
@@ -2,78 +2,75 @@
...
-
-     it('calls PaymentServer.navigateToPaymentServer as expected', () => {
-       assert.deepEqual(PaymentServer.navigateToPaymentServer.args, [
-         [view, config.subscriptions, `subscriptions`],
-       ]);
-     });
-   });
- });
+  describe('render', () => {
+    it('renders correctly, initializes flow evnets, navigates to payments server', () => {
```
6a68: "evnets" -> "events"

packages/fxa-content-server/app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -22,6 +22,13 @@ export default {
     submit: '_submitFlowEventsForm',
   },
 
+  /**
+   * Initialize the flow events
```
6a68: Not sure this is a useful comment, I'd probably omit it

packages/fxa-content-server/app/tests/spec/lib/payment-server.js
lmorchard: (whoops!)

packages/fxa-content-server/app/scripts/lib/payment-server.js
```diff
@@ -4,38 +4,54 @@
...
-      Object.keys(queryParams)
-        .filter(k => queryParams[k] !== null && queryParams[k] !== '')
-        .map(
-          k => `${encodeURIComponent(k)}=${encodeURIComponent(queryParams[k])}`
-        )
-        .join('&');
-    const qs = queryString.length ? `?${queryString}` : '';
+
+    const { flowBeginTime, flowId } = view.metrics.getFlowEventMetadata();
+
```
chenba: FWIW I see the in teh logs that `device_id` is used for an amplitude event, but not the flowBeginTime.

shane-tomlinson: I'll add that.

packages/fxa-content-server/app/scripts/lib/payment-server.js
```diff
@@ -4,38 +4,59 @@
...
-          accessToken.get('token')
-        )}`;
+        const queryString = Url.objToSearchString({
+          // device_id, flow_begin_time, and flow_id need to be propagated to
+          // the payments server so that the user funnel can be traced from the RP,
+          // through the content server, and to the payments server, appearing as
+          // the same user throughout.
+          device_id: deviceId,
+          flow_begin_time: flowBeginTime,
+          flow_id: flowId,
```
chenba: Can these be in camelCase?  I don't know the reason for the snake_case here.

chenba: Oh I just noticed that it's been merged.  No prob. :D

Issue #2829, opened 2019-10-10
[Intermittent] Validation code turns in invalid after inserting for the first time an incorrect code


vasilicatamas:
### Build:
Prod [Train 1.147.3]

### Affected platforms:
Windows 10 x64

### Steps to reproduce
1. Launch Firefox and create Firefox account.
2. At “Enter verification code” step enter the 6-digit code received via email and add one empty space at the beginning and another one at the end of the code
3. Click “Verify” button
4. Erase the empty spaces
5. Click again “Verify” button

### Expected result
The correct 6-digit code is valid and the account is confirmed.

### Actual result
- The correct 6-digit code is seen as “Expired or invalid code” and the account is not validated.
- No new "Verification code" is received.
- The validation time was not exceeded (as you can see in the attached video the email was received 3 minutes ago)

### Additional details
- This is an intermittent issue.
- I am attaching a video.

![code-error](https://user-images.githubusercontent.com/18114210/66555295-16e06e80-eb57-11e9-955c-695343d6136c.gif)

vbudhram:
@vasilicatamas Thanks for report, currently that input field only validates for digits, all other characters are considered invalid. I have a PR in https://github.com/mozilla/fxa/issues/2829 that strips away space characters so that this shouldn't be an issue.

Issue #2427, opened 2019-09-05
Deprecate routes to tokenCode verification


vbudhram:
Since https://github.com/mozilla/fxa/pull/2398, will update the type of code that is generated on sign-in confirmations when using a code. We no longer need the manually generated code but can use the otp based one. Specifically, we should do the following

- [x] Depreciate https://github.com/mozilla/fxa/blob/9a265905fd292178999f6b32b422401d1234f9bb/packages/fxa-auth-server/lib/routes/token-codes.js#L5
- [x] Remove https://github.com/mozilla/fxa/blob/5d709f96d371d51c781d40567cead4f15a70688c/packages/fxa-auth-db-mysql/db-server/index.js#L179
- [ ] Add migration to remove columns `unverifiedToken`.`tokenVerificationCodeHashArg` and `unverifiedTokens`.`tokenVerificationCodeExpiresAt`
- [x] Remove js-client endpoints
- [x] Remove associated tests

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-328)
┆Issue Number: FXA-328

vbudhram:
from mtg: vbudhram to work this since the new stuff will land. This can be worked in train 147+ since the new sign-in codes need to run for at least a train.

data-sync-user:
➤ Wil Clouser commented:

This is done except for removing columns which we don't need to do.

PR #2845, opened 2019-10-10
refactor(code): Rename token code input to otp code input and cleanup logic

vbudhram:
The `token code` name will be going away in https://github.com/mozilla/fxa/issues/2427, lets cleanup relevant parts on the content-server.

As part of cleanup logic, this was tweaked to strip away spaces before validating the input. Keeping this as a draft until I can verify that the keypad is present when signing up in FxiOS.

Fixes https://github.com/mozilla/fxa/issues/2829

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
@@ -20,7 +20,8 @@
         <p class="verification-email-message">{{#unsafeTranslate}}Please enter the verification code that was sent to %(escapedEmail)s within 5 minutes.{{/unsafeTranslate}}</p>
         <form novalidate>
             <div class="input-row">
-                <input type="number" pattern="\d*" class="tooltip-below token-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
+                <!-- Using `type="text" inputmode="numeric"` shows the numericpad on mobile and strips out whitespace on desktop. -->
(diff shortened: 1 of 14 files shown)


packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
```diff
@@ -20,7 +20,7 @@
         <p class="verification-email-message">{{#unsafeTranslate}}Please enter the verification code that was sent to %(escapedEmail)s within 5 minutes.{{/unsafeTranslate}}</p>
         <form novalidate>
             <div class="input-row">
-                <input type="number" pattern="\d*" class="tooltip-below token-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
+                <input type="text" inputmode="numeric" pattern="\d[ ]*" class="tooltip-below otp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
```
vbudhram: Using `type="text" inputmode="numeric"` allows the numberpad to be displayed on mobile devices. This also removes spaces from input.

bbangert: I find these comments really useful, is there no way to include them in the markup (as comments) or in the commit message (so git commit blame might enlighten someone)?

I know one approach is to dig up the commit, then find the PR, then find this comment, but that seems like a lot more work.

vbudhram: Will add as markup.

PR #2866, opened 2019-10-11
fix(css): Fix authorize sign-in screen size on mobile

### LINKED ISSUE #2859: Firefox iOS: "Authorize this sign-in" viewport broken
## Description and  STR
Try to sign into FxA with an unverified account.
 When you get to the "Authorize this sign-in" dialog, background then foreground the app.
Result: The dialog appears zoomed in upon foregrounding and can't be zoomed out.
Happens on iPhone and iPad.

https://user-images.githubusercontent.com/1866605/66643043-9ee17980-ec26-11e9-83be-1eb21dce158f.png

### Environment

iOS 13, Firefox iOS Beta. Not sure if it happens in non beta.

vbudhram:
Fixes #2859 

This sets an explicit 300px width for small screens. Small screens don't have the card panel that the big screens do so I think this should be ok to do. Also, big screen width is set to 480px so this only applies for the screen sizes in between.

Oct-11-2019 15-54-50

### DIFF
--- a/packages/fxa-content-server/app/styles/_layout.scss
+++ b/packages/fxa-content-server/app/styles/_layout.scss
@@ -8,6 +8,10 @@
     @include respond-to('big') {
       width: 480px;
     }
+
+    @include respond-to('small') {
+      width: 300px;
+    }
   }
 
   &.panel {
@@ -29,9 +33,7 @@
 
   @include respond-to('small') {
     margin: 0 auto;
-    max-width: 640px;
     min-height: 300px !important;
-    min-width: 300px;
     padding: 76px 20px 20px 20px;
     position: relative;
     width: 94%;


vbudhram:
@mozilla/fxa-devs Any objections to doing this? r?

LZoog:
What does `max-width: 100%` look like? That might give less of a high left/right space look on phones bigger than the iPhone 5, assuming we have `padding` on that container.

LZoog:
Although, it does look like we have an explicit width set for `@include respond-to('big') {` so this is fine too IMO.

vbudhram:
`width: 100%` does the same as original bug 🤔

Screen Shot 2019-10-11 at 4 30 57 PM

PR #2874, opened 2019-10-13
fix(oauth): Use audience when creating JWTs for auth server

jbuck:


jbuck:
I think this is a follow-up fix to #2737.

PR #2875, opened 2019-10-13
fix(oauth): Fix getUserProfile tests

jbuck:


jbuck:
Follow-up fix for #2874 - didn't catch this in the CircleCI tests

PR #2879, opened 2019-10-14
chore(styles): make price and billing cycle bold on payments page

### LINKED ISSUE #2878: Sub Plat Payment Page — bold the amount being charged in payment confirmation check
In usability test, participants read the first few words beside the checkbox "I authorize Mozilla..." and skim the rest.

Of seven participants, one feels that she did not proactively authorize the charge, even though she checked the box.

Quick fix: bold just the amount being charged ("$4.99") within that copy.

johngruen:
Fixes #2878

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -213,18 +213,14 @@ export const PaymentForm = ({
       </FieldGroup>
 
       {confirm && plan && (
-        <Checkbox
-          data-testid="confirm"
-          name="confirm"
-          required
-          label={`
-          I authorize Mozilla, maker of Firefox products, to charge my
-          payment method $${formatCurrencyInCents(plan.amount)} per ${
-            plan.interval
-          }, according to payment
-          terms, until I cancel my subscription.
-        `}
-        />
+        <Checkbox data-testid="confirm" name="confirm" required>
+          I authorize Mozilla, maker of Firefox products, to charge my payment
+          method{' '}
+          <strong>
+            ${`${formatCurrencyInCents(plan.amount)} per ${plan.interval}`}
+          </strong>
+          , according to payment terms, until I cancel my subscription.
+        </Checkbox>
       )}
 
       {onCancel ? (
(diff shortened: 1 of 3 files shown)


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -190,18 +193,15 @@ export const PaymentForm = ({
...
-          required
-          label={`
-          I authorize Mozilla, maker of Firefox products, to charge my
-          payment method $${formatCurrencyInCents(plan.amount)} per ${
-            plan.interval
-          }, according to payment
-          terms, until I cancel my subscription.
-        `}
-        />
+        <Checkbox data-testid="confirm" name="confirm" required>
```
johngruen: It looks like Prettier did some tidying in this file, but this is the actual change

johngruen:
image

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -244,7 +248,10 @@ export const PaymentForm = ({
 
       <div className="legal-blurb">
         <p>Mozilla uses Stripe for secure payment processing.</p>
-        <p>View the <a href="https://stripe.com/privacy">Stripe privacy policy</a>.</p>
+        <p>
+          View the{' '}
```
LZoog: Is `{' '}` because you'd like more space before the link?

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -190,18 +193,15 @@ export const PaymentForm = ({
...
-          }, according to payment
-          terms, until I cancel my subscription.
-        `}
-        />
+        <Checkbox data-testid="confirm" name="confirm" required>
+          I authorize Mozilla, maker of Firefox products, to charge my payment
+          method{' '}
+          <strong>
+            ${`${formatCurrencyInCents(plan.amount)}`} per
+            {` ${plan.interval}`}
```
LZoog: `${formatCurrencyInCents(plan.amount)} per {plan.interval}` should give the same output.

If we're not in a rush to get this merged it might be nice to have a test for this to make sure it properly renders `children`. You could create a mock test in `fields.test.tsx` and essentially duplicate this test but with the new way that you're using it.
```
describe('Checkbox', () => {
  it('renders its own label', () => {
```

Could call it something like `renders its own label with children`. I can jump in here and do this for you too if you're busy the next couple of days.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
johngruen: i kinda think prettier added that...it's not necessary

johngruen:
@LZoog added tests and fixed the `{' '}` thing

PR #2888, opened 2019-10-14
fix(codes): Enabled signup code experiment for FPN website

vbudhram:
cc @johngruen

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
+++ b/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
@@ -27,6 +27,11 @@ const ROLLOUT_CLIENTS = {
     name: 'Lockbox Extension iOS',
     rolloutRate: 0.0,
   },
+  e6eb0d1e856335fc: {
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
```diff
@@ -27,6 +27,11 @@ const ROLLOUT_CLIENTS = {
     name: 'Lockbox Extension iOS',
     rolloutRate: 0.0,
   },
+  e6eb0d1e856335fc: {
```
shane-tomlinson: Not for this PR of course, but would this feature be a good candidate for redis based feature flagging?

shane-tomlinson: This is the client_id of the VPN, do you also want this feature for the proxy? The proxies new client_id will be 565585c1745a144d.

vbudhram: Landing this as is but @johngruen thoughts?

PR #2904, opened 2019-10-15
fix(errors): Scrub `x-forwarded-for` from response data on a 500 error.

shane-tomlinson:
fixes mozilla/fxa-private#66

@mozilla/fxa-devs


dannycoates:
why is this 137 commits behind master?

Issue #2869, opened 2019-10-11
Modify "Disconnect from Sync" modal copy


LZoog:
While looking at #636 I noticed that, for screen readers, we need an `aria-labelledby` attribute that corresponds to text explaining what the radio selection is for. I was going to go a hidden header route, but @janiceahn requested we:

- Remove the "?" from "Disconnect from Sync?"
- Add copy "You are disconnecting [device name] from Sync because this device is:" at the end of current copy which will give screenreaders context (and it's also best practice for screenreader experience to be as similar to the non-screenreader experiences as possible)

<img width="542" alt="image" src="https://user-images.githubusercontent.com/13018240/66684180-128f8080-ec3f-11e9-893d-23614b46fed3.png">

Self-assigning, will wrap this into my fix for #636 .

LZoog:
Note: changed title due to updated UX request.

PR #724, opened 2019-04-03
feat(modals): Keyboard focus capture for modals

karansapolia:
Implement keyboard focus capture for modals in
content-server. Whenever a modal is opened,
track tab and shift+tab key combinations to keep
keyboard focus within modals and to cycle through
focusable modal elements until esc is pressed
and modal closes.

fixes: #597
fixes: #636

continuation of https://github.com/mozilla/fxa-content-server/pull/7102

@shane-tomlinson @vbudhram @lmorchard, would like to have your thoughts on this.

Issue #636, opened 2018-05-03
Disconnect from Sync screen does not receive keyboard focus when opened


shane-tomlinson:
STR

1. Create an account, attach to Sync.
2. Open /settings
3. Open Devices & apps
4. Click on the "Disconnect" button next to a device.
5. Try to tab around using the keyboard. The page underneath scrolls.

This makes it difficult for screen reader users to use the functionality.

![disconnect-device-focus](https://user-images.githubusercontent.com/848085/39572182-2173fe5e-4ec6-11e8-9088-7f11d246426b.gif)

LZoog:
Patch: #2911

PR #2911, opened 2019-10-15
fix(modal): Make modal keyboard accessible

### LINKED ISSUE #597: Make profile picture selection keyboard accessible
As tabbing seems to select items in the background, it doesn't look like I can select any of the avatar options except `esc` to close inaccessible-avatar

LZoog:
Fixes #597, #636, #2869

@shane-tomlinson I'd like you to take a look at this when you can since you'd reviewed #724, I have a question around testing... Should this be taken out of unit tests and placed in avatar & client disconnection selenium tests? Or how would you test these changes?

Also, I created #2869 while working on these issues, but I'm actually having trouble getting into the modal with a screenreader cursor to benefit from the context created there with `aria-labelledby`, but I asked UX if they'd like me to leave that header in there anyway and they confirmed it should be left. I'll take a better look at this accessibility-wise when/after we do the settings redesign. Ideally, we use `aria-hidden` and other `aria` properties, but this issue fix is to make the modal _keyboard_ accessible and not necessarily screenreader accessible, which will be on my radar for settings redesign.

image
Text spacing was given the r+ from Julia and Janice. Screenshot shows what "focus" looks like when tabbing around (which is the same as hovering).

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/client_disconnect.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/client_disconnect.mustache
@@ -1,7 +1,7 @@
 <div id="client-disconnect">
   <section class="modal-panel">
     {{^hasDisconnected}}
-      <h2>{{#t}}Disconnect from Sync?{{/t}}</h2>
+      <h2>{{#t}}Disconnect from Sync{{/t}}</h2>
       <div class="error"></div>
 
       <p class="intro">
@@ -9,22 +9,24 @@
       </p>
 
       <form novalidate>
+        <div role="radiogroup" aria-labelledby="radio-header">
+          <h3 id="radio-header" class="radio-header">{{#t}}You are disconnecting %(deviceName)s from Sync because this device is:{{/t}}</h3>
           <p class="disconnect-reasons fxa-radio">
             <label>
               <input type="radio" name="disconnect-reasons" value="suspicious"/>
-              {{#t}}Suspicious device{{/t}}
+              {{#t}}Suspicious{{/t}}
             </label>
             <label>
               <input type="radio" name="disconnect-reasons" value="lost"/>
-              {{#t}}Lost or stolen device{{/t}}
+              {{#t}}Lost or stolen{{/t}}
             </label>
             <label>
               <input type="radio" name="disconnect-reasons" value="old"/>
-              {{#t}}Old or replaced device{{/t}}
+              {{#t}}Old or replaced{{/t}}
             </label>
             <label>
               <input type="radio" name="disconnect-reasons" value="duplicate"/>
-              {{#t}}Duplicate device{{/t}}
+              {{#t}}Duplicate{{/t}}
             </label>
             <label>
               <input type="radio" name="disconnect-reasons" value="no" checked="checked"/>
--- a/packages/fxa-content-server/app/scripts/views/mixins/modal-panel-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/modal-panel-mixin.js
@@ -10,13 +10,16 @@
  */
 
 import $ from 'jquery';
+import KeyCodes from '../../lib/key-codes';
 
 export default {
   isModal: true,
-
   notifications: {
     navigate: 'closePanel',
     'navigate-back': 'closePanel',
+  },
+  events: {
+    keydown: 'keyDownListener',
   },
 
   /**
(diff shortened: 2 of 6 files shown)


packages/fxa-content-server/app/scripts/views/mixins/modal-panel-mixin.js
```diff
@@ -33,8 +36,66 @@ export default {
...
+      } else {
+        getNextSelectable(document.activeElement).focus();
+      }
+      // the form submits on 'enter' keypress by default, but if focus is on a radio button,
+      // we want to select it instead
+    } else if (
+      event.keyCode === KeyCodes.ENTER &&
+      document.activeElement.getAttribute('type') === 'radio'
+    ) {
+      event.preventDefault();
```
shane-tomlinson: The behavior here seems slightly strange to me when testing it, I suppose I assumed tab would select the radio button. How hard would it be to submit the form if the radio button is already selected, and select if not?

packages/fxa-content-server/app/scripts/templates/settings/client_disconnect.mustache
```diff
@@ -1,14 +1,16 @@
...
+      <h2>{{#t}}Disconnect from Sync{{/t}}</h2>
       <div class="error"></div>
 
       <p class="intro">
       {{#t}}Your browsing data will remain on %(deviceName)s, but it will no longer sync with your account.{{/t}}
       </p>
 
       <form novalidate>
+        <div role="radiogroup" aria-labelledby="radio-header">
+          <h3 id="radio-header" class="radio-header">{{#t}}You are disconnecting %(deviceName)s from Sync because this device is:{{/t}}</h3>
```
shane-tomlinson: The wording "this device is" makes the radio button labels seem off, .e.g.:

"this device is:" "Suspicious device"

The "device" on the end of "suspicious" seems somewhat superfluous.

packages/fxa-content-server/app/scripts/views/mixins/modal-panel-mixin.js
```diff
@@ -33,8 +36,66 @@ export default {
...
+      const previousIndex = focusableElements.indexOf(element) - 1;
+      // return last element if the previous element doesn't exist
+      return focusableElements[previousIndex]
+        ? focusableElements[previousIndex]
+        : focusableElements[focusableElements.length - 1];
+    }
+
+    //avoid IME composition keydown events
+    //ref: https://developer.mozilla.org/docs/Web/Events/keydown#Notes
+    if (event.isComposing || event.keyCode === 229) return;
```
shane-tomlinson: Can you add `229` to [key-codes.js](https://github.com/mozilla/fxa/blob/0bbac2ba3666749309d9bfc03448fc38a1b21331/packages/fxa-content-server/app/scripts/lib/key-codes.js#L4) and use a named constant here?

Also, can you put the `return` on its own line and add some block`{}`?

I thought we had a lint rule about that.

shane-tomlinson:
Some more thoughts, I think @ryanfeeley wants to get rid of the modal dialogs all together because of the UI issues around them, I wonder if its worthwhile holding off to find out how much effort this warrants? If removing the modals is more than a few months away, getting this merged with a couple of basic tests is probably still worthwhile.

packages/fxa-content-server/app/scripts/templates/settings/client_disconnect.mustache
LZoog: @janiceahn @janiceahn Agreed, I didn't notice this. Would you like to tweak this copy?

packages/fxa-content-server/app/scripts/views/mixins/modal-panel-mixin.js
LZoog: Done. Haha, developer preference. I don't even use semicolons on personal projects -le gasp-

LZoog:
@shane-tomlinson I addressed all of your comments besides the one I'll need UX input on for copy phrasing. Your camera close issue stemmed from me changing the `a[href]` selector to `#file` and `#camera` when I was working with that test template but I think these are going to have to be seleinum tests so I can revert those and leave the selector as `a[href]`.

The form now submits when you hit "enter" on a radio button already selected - this is a good suggestion IMO, just from using screenreaders and such before I personally wouldn't expect a tab selection to select for you because users should be able to tab to a submit button if they want to. It's a can of worms like you said though, someone else could argue for some other functionality. 🤷‍♀️ Regardless I think what you suggested is a good option.

I'm going to have to put getting tests in on hold until I get through a couple of train 149 priorities first. * old man yells at cloud emoji *

LZoog:
@shane-tomlinson 

Since it doesn't look like [the simulated events (here Tab) does not trigger browser functionality](https://github.com/theintern/leadfoot/issues/98#issuecomment-264970531), specifically advance focus to next element in tab order, this doesn't seem to work as a Selenium test, and making `test_template.mustache` pop open a modal to test seems overly complicated since we are redesigning settings anyway. I left a single test to check for the active element after the tab is open... I know this doesn't have super great code coverage, but I think this should be fine as-is, considering the upcoming redesign. 🤷‍♀️

PR #2919, opened 2019-10-15
fix(images): add viewBox to fpn svg

johngruen:
@LZoog


johngruen:
Didn't file a bug for this, but the icon was not showing up on staging and we think the lack of viewbox is the issue

PR #2922, opened 2019-10-15
fix(l10n): We sentence cased the email titles, so tests started failing

shane-tomlinson:
Update the tests.


shane-tomlinson:
Other l10n related tests fail now, closing this since the 147 tag should still build.

PR #2924, opened 2019-10-15
chore(release): Merge train-147 into train-148

shane-tomlinson:


packages/fxa-auth-db-mysql/CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ No changes.
 
 ### New features
 
-* add vscode tasks for running tests and debugger (dac5e8b98)
+- add vscode tasks for running tests and debugger (dac5e8b98)
```
shane-tomlinson: It seems like `-` is normally used, so there are a bunch of these conversions from `*` to `-`. Not sure where the `*` is coming from.

PR #2928, opened 2019-10-15
fix(emails): Fix write-emails-to-disk script

### LINKED ISSUE #2964: Add auth-server write-emails-to-disk script to CI runs
write-emails-to-disk is a script used as a development tool for email templates on auth-server. We made a change to email machinery that broke the script and didn't notice. Maybe we should run it in CI so that we notice in the future?

vbudhram:
Found in https://github.com/mozilla/fxa/pull/2921. This updates the script to pass mock statd object.

Fixes #2964

### DIFF
--- a/packages/fxa-auth-server/scripts/write-emails-to-disk.js
+++ b/packages/fxa-auth-server/scripts/write-emails-to-disk.js
@@ -3,28 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * Write emails to disk. Output is written to ./.mail_output/<email type>.html
+ * Write all emails to disk. Output is written to ./.mail_output/<email type>.html
+ * and ./.mail_output/<email type>.txt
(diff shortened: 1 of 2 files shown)


lmorchard:
Hmm, if this script broke and we didn't notice until now, maybe we should run it as part of CI runs?

shane-tomlinson:
> Hmm, if this script broke and we didn't notice until now, maybe we should run it as part of CI runs?

We started adding tests for scripts like these, maybe we never added one for this? If not, we should because this script has broken a few times now.

lmorchard:
Not intending to block this PR at all, I created a follow-up issue: https://github.com/mozilla/fxa/issues/2964

vbudhram:
@shane-tomlinson @lmorchard I've updated this PR with the following

* Removed generating specific email templates from command line (I always generate all templates and made testing a bit easier)
* Added a tests to check that template was created

PR #2930, opened 2019-10-15
fix(oauth): fix session token ids for pairing and session token scope

vladikoff:


packages/fxa-auth-server/lib/routes/oauth.js
```diff
@@ -197,6 +197,7 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
           const newSessionToken = await db.createSessionToken(
             sessionTokenOptions
           );
+          grant.session_token_id = newSessionToken.id;
```
vladikoff: should probably rename this or something, otherwise the id is modified on the fly

shane-tomlinson: Is `session_token_id` used in oauthRouteUtils.newTokenNotification? Otherwise this is confusing because the value is deleted [down here](https://github.com/mozilla/fxa/pull/2930/files#diff-ba2df0b1af69d2687f589b68e4d42930R218).

Also, some places in the grant make use of session_token_id, some [sessionTokenId](https://github.com/mozilla/fxa/blob/bc1a15c98cae7932c5aadf00c42bdc068cd5c31f/packages/fxa-auth-server/fxa-oauth-server/lib/grant.js#L173). Within grants, we should probably choose one variant and stick with it to avoid the discrepancy biting us.

vladikoff: > Is session_token_id used in oauthRouteUtils.newTokenNotification? Otherwise this is confusing because the value is deleted down here.

yea exactly

vladikoff:
@rfk any input on this?

packages/fxa-auth-server/lib/routes/oauth.js
```diff
@@ -197,11 +197,15 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
           const newSessionToken = await db.createSessionToken(
             sessionTokenOptions
           );
+          // the new session token information is later
+          // used later in 'newTokenNotification' to attach it to device records
```
rfk: nit: "later used later" is awkward; I'd phrase this along the lines of:

```
// The session token info is used by `newTokenNotification` when creating a device record,
// so make sure it reflects the newly-created session token.
```

Assuming I correctly understand what's happening here, that is :-)

vbudhram:
@vladikoff was this related to https://sentry.prod.mozaws.net/operations/auth-prod/issues/6443091/events/34086382/?

PR #2931, opened 2019-10-15
Train 148

clouserw:


packages/fxa-auth-db-mysql/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.148.2
+
+### Other changes
+
+* release: Merge branch 'train-147' into train-148-merge-147 (66e170d45)
```
vbudhram: These merge branches shouldn't really be added to the changelog. We can resolve that in another PR.

packages/fxa-auth-db-mysql/CHANGELOG.md
```diff
@@ -6,7 +12,11 @@ No changes.
 
 ### New features
 
-* add vscode tasks for running tests and debugger (dac5e8b98)
+- add vscode tasks for running tests and debugger (dac5e8b98)
```
vbudhram: Also not sure why `-` and `*` differences. Were you using an older script @shane-tomlinson? Either way we can clean this up on master after tagging.

PR #2895, opened 2019-10-15
fix(auth-server): rewrite subhub msg processor

bbangert:
Because:

* The existing subhub processor had multiple states it failed to
  account for.
* Profile cache locations were not updated.

This commit:

* Clears the profile cache and notifies that the profile was updated.
* Handles cancelled subscription state.
* Refactors subhub processor for less closure indents.

Closes #2851

PR #2941, opened 2019-10-15
fix(auth-server): check proper code location

bbangert:
Issue #2895


bbangert:
Indeed, honestly, TypeScript would've caught this....

PR #2947, opened 2019-10-16
restore workdir in payments dockerfile build

jrgm:
Hey Jared,

The payments server was failing on start in kubernetes dev because it was looking for bin/fxa-payments-server.js relative to /fxa-shared and failing. (I suppose the startup command could use an  absolute path, but still seems better to just restore the WORKDIR).

But then a second issue on the build: lint was failing. So added the second commit. I dunno if that's the idiomatic way to ignore an unused variable from a destructure, so let me know if there's a better way.

r? @6a68

### DIFF
--- a/packages/fxa-payments-server/server/lib/amplitude.js
+++ b/packages/fxa-payments-server/server/lib/amplitude.js
@@ -11,12 +11,10 @@ const {
   mapFormFactor,
   mapLocation,
   mapOs,
-  mapUserAgentProperties,
   toSnakeCase,
 } = require('../../../fxa-shared/metrics/amplitude.js');
 const config = require('../config');
 const amplitude = config.get('amplitude');
-const joi = require('joi');
 const log = require('./logging/log')();
 const ua = require('../../../fxa-shared/metrics/user-agent');
 const { version: VERSION } = require('../../package.json');
--- a/packages/fxa-payments-server/Dockerfile
+++ b/packages/fxa-payments-server/Dockerfile
@@ -50,4 +50,5 @@ RUN chown -R app:app /fxa-shared
 USER app
 RUN npm ci
 
+WORKDIR /app
 CMD [ "/usr/local/bin/node", "server/bin/fxa-payments-server.js" ]


jrgm:
p.s., -dev is currently running this branch docker image. Next commit to master that passes will use :latest and start failing again. Although why was the building failing with this eslint error. I dunno.

jrgm:
>  Although why was the building failing with this eslint error. I dunno.

```
+ grep eslint ../../.circleci/../packages/fxa-payments-server/Gruntfile.js
grep: ../../.circleci/../packages/fxa-payments-server/Gruntfile.js: No such file or directory
+ grep '"eslint"' ../../.circleci/../packages/fxa-payments-server/package.json
    "eslint": "^6.2.1",
+ docker run --net=host fxa-payments-server:test npm run lint

> fxa-shared@1.148.2 lint /fxa-shared
> npm-run-all --parallel lint:*
```
So it was linting /fxa-shared, which was okay. Okay, so restoring the WORKDIR is the right fix for other reasons.

packages/fxa-payments-server/server/lib/amplitude.js
```diff
@@ -11,12 +11,11 @@ const {
   mapFormFactor,
   mapLocation,
   mapOs,
-  mapUserAgentProperties,
+  mapUserAgentProperties, //eslint-disable-line no-unused-vars
```
6a68: Sorry about this; I don't understand why the linter didn't run when I had this in the PR.

The better fix is to just delete `mapUserAgentProperties` from this list; if it isn't used, we don't need it.

jrgm: Okay, I'll delete that line.

I noted the reason why the linter passed above in https://github.com/mozilla/fxa/pull/2947#issuecomment-542511065. I was linting `/fxa-shared`, not `/app`

jrgm: Updated the branch

jrgm:
Thanks!

PR #2948, opened 2019-10-16
fix(payments): Get post-metrics route working and add tests

6a68:
I assumed the `validate` and `preProcess` route properties were supported on payments-server, as they are on content-server, but noticed `/metrics` was 404ing locally, which led me to this fix. I've inlined the validate and preprocess code inside the route handler.

For some reason, the " __version__ should have correct structure" route test is failing locally for me, but doesn't seem to fail on CI. ¯\_(ツ)_/¯

### DIFF
--- a/packages/fxa-payments-server/package-lock.json
+++ b/packages/fxa-payments-server/package-lock.json
@@ -4672,6 +4672,15 @@
       "resolved": "https://registry.npmjs.org/caseless/-/caseless-0.12.0.tgz",
       "integrity": "sha1-G2gcIf+EAzyCZUMJBolCDRhxUdw="
     },
+    "celebrate": {
+      "version": "10.0.1",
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
@6a68 - I updated this PR so that fxa-payments-server route definitions can now use `preProcess` and `validate` like the content server, I did so by copying the content server code. There is now a lot of duplication between the two services, and I'm working on another PR to push that into fxa-shared, but didn't want to add more to this PR than necessary.

The reason the __version__ test was failing is that the route definition came after the catchall handler used in dev mode here: https://github.com/mozilla/fxa/blob/3347940a117f8a784964d6629ea011923acb6e14/packages/fxa-payments-server/server/lib/server.js#L190

I'm going to r+ this, and if you agree, let's merge this and work on a follow up to extract the common code!

6a68:
I'm the only one seeing this webpack error; Les and Barry aren't. Merging.

PR #2951, opened 2019-10-16
refactor(routes): Extract Express routing helpers into fxa-shared

shane-tomlinson:
follow on from #2948

@6a68 - The third commit is the refactor, the first 2 commits are from #2948. Think this would be worthwhile?


6a68:
Yeah, I would like to avoid this degree of duplication across the monorepo. Seems likely that bug fixes on one side wouldn't 100% make their way to the other side.

6a68:
Not sure what's up with the test failures; landed the first two commits, I'll debug this in a bit.

Issue #2846, opened 2019-10-10
Avatar on Subscriptions Management page has a hover and active style


chenba:
## Description

Avatar on Subscriptions Management page has a hover and active style.

### Steps to reproduce

1. Subscribe to a product 
1. Go to Subscriptions Management page
1. Hover and click the avatar image

### Expected result

No visual effects.

### Actual result

Visual effect.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-429)
┆Issue Number: FXA-429

PR #2953, opened 2019-10-16
fix(styles): match subscription styles to fxa settings

johngruen:
it's been bugging me that the subscription settings are slightly off from the FxA settings in a few ways...this patch fixes that:

* Makes settings paragraphs 1.5 line-height
* Makes settings paragraph color conform to fxa settings
* makes display name and email in header conform to display pattern on fxa settings
* makes button text conform to sizing on fxa settings
* better aligns CC information
* makes contact support a primary action
* removes `:hover` interaction from avatar

Before
image

After
image


johngruen:
Fixes #2846

vbudhram:
@johngruen Are trying to get this on `train-148`? If so, it might be easier to open this PR against that branch, that way you won't have to cherry-pick.

Issue #2345, opened 2019-08-28
Ensure source maps show up in Sentry


clouserw:
Lots of the errors in sentry are in our JS bundles.  Let's add sourcemaps so we can decipher them.

Example: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/6177303/

meandavejustice:
Sentry sourcemaps are working on content server in my branch, I've run into a bit of trouble with configuration of the sourcemaps on the payment server, I think this may have to do with use using `react-scripts`, I had to bring in a separate package in order to override the default webpack config.

I've switched off of this until I finish the other issues in the current milestone, but will hopefully get back to it before the end of the week.

PR #2959, opened 2019-10-16
task(sentry): add sourcemaps to sentry for content server

meandavejustice:
- refs #2345

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/sentry.js
+++ b/packages/fxa-content-server/app/scripts/lib/sentry.js
@@ -4,7 +4,7 @@
 
 import _ from 'underscore';
 import Logger from './logger';
-import Raven from 'raven';
+import * as Sentry from '@sentry/browser';
 import Url from './url';
 
 var ALLOWED_QUERY_PARAMETERS = [
@@ -20,13 +20,6 @@ var ALLOWED_QUERY_PARAMETERS = [
   'setting',
   'style',
 ];
-
-// last error that Sentry sent
-// stored in the global state due to
-// callback options and multiple Sentry interfaces
-var SAME_ERROR_LIMIT = 10;
-var SAME_ERROR_ATTEMPTS = 0;
-var PREVIOUS_ERROR = null;
 
 /**
  * function that gets called before data gets sent to error metrics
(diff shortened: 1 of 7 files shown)


packages/fxa-content-server/app/scripts/lib/sentry.js
```diff
@@ -1,233 +0,0 @@
...
-    });
-
-    var extraContext = {
-      tags: tags,
-    };
-
-    if (this._release) {
-      // add release version if available
-      extraContext.release = this._release;
-    }
```
chenba: I don't understand why this module is being completely removed.  Wouldn't we lose the extra context at the very least?

meandavejustice: We are now using the direct sentry package instead of a fork of the outdated package. We no longer send the metrics to a custom endpoint proxy. We should make sure that we don't leave anything behind in the new version.

meandavejustice: @chenba After looking into this a bit more, I think we should go ahead with this conversion to get us off the the legacy raven package, and file a follow up issue to add any extra info after we see what info we are getting in sentry.

chenba: This file was a wrapper.  I don't know much about the Sentry API and how it has changed from the old one to new one, so I can't speak to much work it is to re-wrap the new one.  But it doesn't seem possible that, by default, the new Sentry package will handle everything in here.  For example, there's a hard coded whitelist of query params (https://github.com/mozilla/fxa/pull/2959/files/31407a79ef0f1a0fc9318607457a0b242207c393#diff-af20f0c892e15cdf0f5aa74531981dcbL10).  Without that or any additional configuration, I don't see how Sentry would know what query params are allow to be collected.

bbangert: I added this file when revamping the Sentry support in auth-server to do filtering of request data:
https://github.com/mozilla/fxa/pull/3128/files#diff-a44a71b1f6c14d7564bb44a782497ff4

I assume some aspects of this can be applied here.

meandavejustice:
closing for https://github.com/mozilla/fxa/pull/3388

PR #2960, opened 2019-10-16
fix(auth-server): compare message timestamp in milliseconds

### LINKED ISSUE #2956: [fxa-auth-server]: Fix inconsistent timestamp checks in sqs receiver
The SQS receiver does timestamp comparisons, but the timestamp from the database is in milliseconds, while the event createdAt is in seconds.

bbangert:
Fixes #2956

### DIFF
--- a/packages/fxa-auth-server/lib/subhub/updates.js
+++ b/packages/fxa-auth-server/lib/subhub/updates.js
@@ -99,14 +99,18 @@ class SubHubMessageProcessor {
       }
     }
 
+    // The createdAt in our db always has milliseconds, so we compare and store
+    // this value in milliseconds as this Stripe value is in seconds.
+    const messageCreatedAt = message.eventCreatedAt * 1000;
+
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/subhub/updates.js
```diff
@@ -99,14 +99,18 @@ class SubHubMessageProcessor {
       }
     }
 
+    // The createdAt in our db always has milliseconds, so we compare and store
+    // this value in milliseconds as this Stripe value is in seconds.
+    const messageCreatedAt = message.eventCreatedAt + 1000;
```
lmorchard: Did you mean *plus* 1000 or *times* 1000 here?

bbangert: I definitely meant *times*.

6a68: ouch.  reviewer fail

PR #2966, opened 2019-10-16
chore(tests): tests

vbudhram:
Master should build correctly, right?


vbudhram:
From slack: Need to update docker mysql image, closing here.

PR #2967, opened 2019-10-16
fix(emails): Send service when verifying account with code

vbudhram:
This fixes an issue where we don't send a post verification email after the user verifies their account using a code.

I targeted against train-148 since it is a regression, but it could go on 149.

cc @davismtl @ryanfeeley  @irrationalagent


vbudhram:
@mozilla/fxa-devs r?

PR #2970, opened 2019-10-16
chore(codes): Enabled signup codes for all users

vbudhram:
From meeting: We should enable signup codes for all reliers.

I chose to keep the experiment logic since that involved the least amount of test changes. Once skyline is over, we can clean this up further.

Targetted against train-148, but can be on another.

cc @davismtl @clouserw

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
+++ b/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/signup-code.js
@@ -5,48 +5,11 @@
 'use strict';
 
 const BaseGroupingRule = require('./base');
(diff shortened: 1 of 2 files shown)


vladikoff:
yay!

shane-tomlinson:
> I chose to keep the experiment logic since that involved the least amount of test changes. Once skyline is over, we can clean this up further.

Oh man, that's gonna be as big as the email-first test conversion! May the force be with you.

shane-tomlinson:
There are some lint failures, can you fix those and see whether the rest of the tests pass?

vbudhram:
@mozilla/fxa-devs r?

Issue #2936, opened 2019-10-15
Inconsistent line-height within subscriptions screens vs modals


LZoog:
We should make sure the `line-height` is consistent. We should either increase it where it is lower or decrease it where it is set higher.

Low line-height:
<img width="700" alt="image" src="https://user-images.githubusercontent.com/13018240/66870830-e722d380-ef67-11e9-884c-71378ee9a7da.png">
<img width="678" alt="image" src="https://user-images.githubusercontent.com/13018240/66870845-f0ac3b80-ef67-11e9-9a38-158c14b99fd2.png">
<img width="542" alt="image" src="https://user-images.githubusercontent.com/13018240/66871223-d161de00-ef68-11e9-95f1-d40d97105913.png">

Modal with higher appearing line-height:
<img width="700" alt="image" src="https://user-images.githubusercontent.com/13018240/66870931-23eeca80-ef68-11e9-90cb-06358fbe911d.png">
<img width="700" alt="image" src="https://user-images.githubusercontent.com/13018240/66871447-4cc38f80-ef69-11e9-9373-80978e044bd6.png">

We seem to have `line-height: 185%` (the same as `1.85em`) set on modals. I just checked a Zeplin board and that does seem accurate for those. @ UX, would you like the non-modal copy also set to `1.85` or what would you prefer? We tend to use `1.5` in settings copy.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-445)
┆Issue Number: FXA-445

johngruen:
I’m not in love with the line heights in the modals...maybe we can do 1.5 everywhere?

johngruen:
@johngruen i think i can close this, will confirm

PR #2972, opened 2019-10-16
fix(styles): align support styles with the rest of fxa

johngruen:


johngruen:
Fixes #2936

johngruen:
@chenba wanted to make sure the styles in the support page better matched settings in general

packages/fxa-content-server/app/scripts/templates/support.mustache
```diff
@@ -25,17 +25,10 @@
...
-            <header class="settings-unit-summary">
-              <h2 class="settings-unit-title">Subscriptions</h2>
-            </header>
-          </div>
-        </div>
         <div class="settings-unit">
           <div class="support-form">
             <header>
-              <h3>{{#t}}Contact Us{{/t}}</h3>
+              <h2>{{#t}}Contact Us{{/t}}</h2>
```
LZoog: I think this should remain an `h3` as it makes sense for it to fall under "Subscriptions". (It looks like we have an empty `h2` rendered if there's no display name, I may not have latest though)

image

packages/fxa-content-server/app/styles/modules/_chosen.scss
```diff
@@ -71,9 +71,9 @@
 
   .chosen-single {
     div {
-      border-color: #686869 transparent transparent transparent;
+      border-color:  $grey-30 transparent transparent transparent;
```
LZoog: 🙌 for variable usage

packages/fxa-content-server/app/scripts/templates/support.mustache
johngruen: sure!

johngruen:
Made @LZoog's requested change and merged

Issue #2976, opened 2019-10-16
Investigate amplitude email metrics from train-147


vbudhram:
Starting in train-147 our amplitude email metrics no longer get set. Most likely cause was https://github.com/mozilla/fxa/pull/2535, where we changed the template names. However, I would have expected new template names to get populated.

Ref: https://analytics.amplitude.com/mozilla-corp/chart/new/rvcb9rq

cc @davismtl @irrationalagent

vbudhram:
@irrationalagent It appears that the email template names changed, https://github.com/mozilla/fxa/pull/2535#discussion_r326268219. Do we need to do anything in amplitude to pick this change up?

vbudhram:
Blehhhh looks like this needs to be updated with new template names, https://github.com/mozilla/fxa/blob/77a2c2821af087d5e5ccdabeff7061d440986e10/packages/fxa-auth-server/lib/metrics/amplitude.js#L19.

PR #2978, opened 2019-10-16
fix(metrics): Fix email metrics not reporting to amplitude

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/2976

Creating as draft since I need to figure out why test cases didn't catch it.

If we do another point on train-148 it would be nice to get on it.

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/amplitude.js
+++ b/packages/fxa-auth-server/lib/metrics/amplitude.js
@@ -17,30 +17,33 @@ const { version: VERSION } = require('../../package.json');
 
 // Maps template name to email type
 const EMAIL_TYPES = {
-  lowRecoveryCodesEmail: '2fa',
-  newDeviceLoginEmail: 'login',
-  passwordChangedEmail: 'change_password',
-  passwordResetEmail: 'reset_password',
-  passwordResetAccountRecoveryEmail: 'account_recovery',
-  passwordResetRequiredEmail: 'reset_password',
-  postChangePrimaryEmail: 'change_email',
-  postRemoveSecondaryEmail: 'secondary_email',
-  postVerifyEmail: 'registration',
-  postVerifySecondaryEmail: 'secondary_email',
-  postAddTwoStepAuthenticationEmail: '2fa',
-  postRemoveTwoStepAuthenticationEmail: '2fa',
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/test/local/metrics/amplitude.js
```diff
@@ -565,751 +565,83 @@ describe('metrics/amplitude', () => {
...
-
-    describe('email.verifySecondaryEmail.sent', () => {
-      beforeEach(() => {
-        return amplitude(
-          'email.verifySecondaryEmail.sent',
-          mocks.mockRequest({})
-        );
-      });
+    describe('email templates', () => {
+      const templates = require('../../../lib/senders/templates/_versions');
```
vbudhram: Updated this test all check email templates vs email versions, which then tests the bounce/sent/delivered metrics. This gives us some assurance that we won't miss new templates for being reported to amplitude.

packages/fxa-auth-server/lib/metrics/amplitude.js
```diff
@@ -17,30 +17,33 @@ const { version: VERSION } = require('../../package.json');
...
-  postVerifyEmail: 'registration',
-  postVerifySecondaryEmail: 'secondary_email',
-  postAddTwoStepAuthenticationEmail: '2fa',
-  postRemoveTwoStepAuthenticationEmail: '2fa',
-  postAddAccountRecoveryEmail: 'account_recovery',
-  postRemoveAccountRecoveryEmail: 'account_recovery',
-  postConsumeRecoveryCodeEmail: '2fa',
-  postNewRecoveryCodesEmail: '2fa',
-  recoveryEmail: 'reset_password',
+  downloadSubscription: 'subscription_download',
```
vbudhram: After adding tests, noticed that `downloadSubscription`, `verificationReminderFirst`, `verificationReminderSecond` were not being reported to amplitude, so adding here fixes that issue.

vbudhram:
@mozilla/fxa-devs calling this ready for review!

vbudhram:
> Was it a missing key?

Yup it was this exactly. Particularly the last step in https://github.com/mozilla/fxa/issues/511. This doesn't solve the problem 100%, however email version numbers gets created automatically with new templates, so hopefully it will allow us to catch it with this test.

PR #2992, opened 2019-10-17
chore(tests): add tests for Amplitude event actions in Subscriptions

### LINKED ISSUE #2940: [payments-server] Add Tests for `view` and `engage` Amplitude Event Callbacks in Subscriptions
PR #2852 added four actions/callbacks for Amplitude events in Subscriptions.  Since the tests use a connected Subscriptions component, we cannot simply pass in mocks for these callbacks.  Having `jest` mock the entire `actions` module break existing tests, which is why these tests have been deferred to this follow-up issue.

The actions:
 - `manageSubscriptionsMounted`
 - `manageSubscriptionsEngaged`
 - `cancelSubscriptionMounted`
 - `cancelSubscriptionEngaged`

chenba:
Fixes #2940 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-payments-server/src/routes/Subscriptions/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Subscriptions/index.test.tsx
@@ -4,6 +4,24 @@ import '@testing-library/jest-dom/extend-expect';
 import nock from 'nock';
 import waitForExpect from 'wait-for-expect';
 
+// mock before the connected Subscriptions is imported below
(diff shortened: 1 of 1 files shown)


chenba:
> 
> 
> Seemed like this might still be trying to send out amplitude events...

Good catch.  There are two reason why no events are sent:

1. `sendBeacon` is undefined in the execution context of these tests.
1. No flow event query params.

The first isn't really under our control though.

Issue #2704, opened 2019-10-02
Metrics: Secondary Email Setup


davismtl:
Currently, we only know how many people setup a secondary email based on the emails we send out.

We should also track an event when it is enabled and log a binary user property in Amplitude so we can cohort user retention more easily down the road.

This will also help support this feature work:
https://github.com/mozilla/fxa/issues/2595

Feature doc:
https://docs.google.com/document/d/1Wz0itS76pNS24zeNS5MEqqjQdwxHuNyfq3PTj8C688s/edit#

PR #2997, opened 2019-10-17
feat(metrics): Add amplitude user properties for security features

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/2704
 
This PR adds user properties to our amplitude events on the content-server. The new user properties correspond to whether or not the user has enabled secondary email, recovery key and two step authentication.

```
amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_pref - view","time":1571344015406,"user_id":"26d445cc0b50448db828e3180422e043","device_id":"eaea82064cc54ec3aee3985bcef58cbd","session_id":1571344014615,"app_version":"148","language":"en","os_name":"Mac OS X","os_version":"10.14","event_properties":{},"user_properties":{"flow_id":"e94d846b7ebdc5f07d23277cf8608d5b25f169e271abed962856b7180ebb4581","ua_browser":"Firefox","ua_version":"70.0","utm_campaign":"fx-welcome-secondary","utm_content":"fx-activate","utm_medium":"email","$append":{"two_step_authentication":false,"account_recovery":false,"emails":true}}}
```

Prefs
- two_step_authentication
  - True or false whether user has 2FA enabled
- account_recovery
  - True or false whether user has account recovery enabled
- emails
  - True or false whether user has secondary emails enabled

- [x] Add tests

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/account_recovery/account_recovery.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/account_recovery/account_recovery.js
@@ -47,10 +47,12 @@ const View = BaseView.extend({
 
   beforeRender() {
     const account = this.getSignedInAccount();
-    return this.setupSessionGateIfRequired().then(() => {
-      return account.checkRecoveryKeyExists().then(status => {
-        this.model.set('hasRecoveryKey', status.exists);
-      });
+    return this.setupSessionGateIfRequired().then(isEnabled => {
+      if (isEnabled) {
+        return account.checkRecoveryKeyExists().then(status => {
+          this.model.set('hasRecoveryKey', status.exists);
+        });
+      }
     });
   },
 
@@ -63,6 +65,8 @@ const View = BaseView.extend({
       hasRecoveryKey: !!hasRecoveryKey,
       isPanelOpen: this.isPanelOpen(),
     });
+
+    this.metrics.logUserPreferences(this.className, !!hasRecoveryKey);
   },
 
   refresh: showProgressIndicator(
--- a/packages/fxa-content-server/app/tests/spec/lib/metrics.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/metrics.js
@@ -425,7 +425,7 @@ describe('lib/metrics', function() {
             var data = JSON.parse(
               windowMock.navigator.sendBeacon.getCall(0).args[1]
             );
-            assert.lengthOf(Object.keys(data), 35);
+            assert.lengthOf(Object.keys(data), 36);
             assert.equal(data.broker, 'none');
             assert.equal(data.context, Constants.CONTENT_SERVER_CONTEXT);
             assert.match(data.deviceId, /^[0-9a-f]{32}$/);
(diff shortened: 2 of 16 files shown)


vbudhram:
@mozilla/fxa-devs this is ready for review!

vbudhram:
@davismtl @irrationalagent do we have a document for keeping track of new amplitude properties?

bbangert:
Just out of curiosity, why not include this descriptive PR message as the commit message? It'd be immediately accessible in the commit log if someone is trying to understand why this changed.

packages/fxa-content-server/server/lib/routes/post-metrics.js
```diff
@@ -152,6 +152,10 @@ const BODY_SCHEMA = {
   uniqueUserId: STRING_TYPE.regex(UNIQUE_USER_ID_PATTERN)
     .allow('none')
     .required(),
+  userPreferences: joi
```
shane-tomlinson: Can you do that now? We keep finding people throwing arbitrary garbage into the payloads to see what they can get away with.

vbudhram: Whoops, meant to do that in this PR, missed my own TODO 🙇🏽‍♂️

vbudhram:
Thanks @shane-tomlinson, I've updated this to add the extra validation.

> why not include this descriptive PR message as the commit message?

@bbangert I wonder if adding the more detailed commit message would be useful in the merge commit? It is a little easier to add it through github UI than through the code editor.

packages/fxa-content-server/app/scripts/views/settings/account_recovery/account_recovery.js
```diff
@@ -63,6 +65,8 @@ const View = BaseView.extend({
       hasRecoveryKey: !!hasRecoveryKey,
       isPanelOpen: this.isPanelOpen(),
     });
+
+    this.metrics.logUserPreferences(this.className, !!hasRecoveryKey);
```
chenba: mega nit: I suppose the `!!` here is redundant.

packages/fxa-content-server/app/tests/spec/lib/metrics.js
```diff
@@ -1051,4 +1052,49 @@ describe('lib/metrics', function() {
...
+          return metrics.flush();
+        })
+        .then(() => {
+          // a second flush!
+          assert.deepEqual(
+            metrics._send.args[1][0].userPreferences,
+            userPrefMetrics
+          );
+
+          return metrics.flush();
```
chenba: Why is this flush() necessary?

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -343,6 +349,22 @@ function mapExperiments(data) {
   }
 }
 
+function mapUserPreferences(data) {
+  const { userPreferences } = data;
+
+  // Don't send user preferences metric if there are none!
+  if (!userPreferences || Object.keys(userPreferences) === 0) {
```
chenba: `Object.keys(userPreferences).length`?

packages/fxa-content-server/app/tests/spec/lib/metrics.js
vbudhram: Looks like I was following the convention in tests above, however this doesn't really need the second metrics flush, consider it removed!

PR #3008, opened 2019-10-18
feat(email-first): Email first by default brokers that have it enabled.

shane-tomlinson:
This leans on the experiment code to enable email first by default
for all brokers with support. This approach was taken because
the number of test updates that are needed is astronomical and
this is a step in the right direction without causing too much
churn.

A step towards #2827


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3015, opened 2019-10-18
feat(email-first): Disable legacy signin/signup flows in fx_desktop_v3

### LINKED ISSUE #595: Use email-first as default for fx_desktop_v3
blocked by #6528
blocked by #6529 
blocked by #6534

shane-tomlinson:
Add a new broker capability: 'emailFirstForced', if set to `true`, then legacy signin/signup
pages are disabled for that broker. Used in fx_desktop_v3, and loads of tests updated.

Partial extraction from #2827
fixes #595

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/force_auth.js
+++ b/packages/fxa-content-server/app/scripts/views/force_auth.js
@@ -139,6 +139,11 @@ var View = SignInView.extend({
 
     return this.navigate('signup', {
       error: AuthErrors.toError('DELETED_ACCOUNT'),
+      // account is for the email-first signup flow, without an account
+      // the signup page redirects to index
+      account,
+      // forceEmail is a hint to the signup page that the user should not
+      // be given the option to change their address.
       forceEmail: account.get('email'),
     });
   },
--- a/packages/fxa-content-server/tests/functional/settings_secondary_emails.js
+++ b/packages/fxa-content-server/tests/functional/settings_secondary_emails.js
@@ -25,6 +25,7 @@ const {
   closeCurrentWindow,
   createUser,
   fillOutResetPassword,
+  fillOutEmailFirstSignIn,
   fillOutSignIn,
   fillOutSignUp,
   getUnblockInfo,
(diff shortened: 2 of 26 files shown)


shane-tomlinson:
Builds on #3008

shane-tomlinson:
Ready for review! @mozilla/fxa-devs - r?

shane-tomlinson:
@vbudhram - You said I could tag you in this when ready. Ready!

packages/fxa-content-server/app/scripts/views/force_auth.js
```diff
@@ -139,6 +139,11 @@ var View = SignInView.extend({
 
     return this.navigate('signup', {
       error: AuthErrors.toError('DELETED_ACCOUNT'),
+      // account is for the email-first signup flow, without an account
+      // the signup page redirects to index
+      account,
+      // forceEmail is a hint to the signup page that the user should not
```
vbudhram: I am probably missing something but in what cases would we want to force emails for a user signup?

packages/fxa-content-server/tests/functional/settings_secondary_emails.js
```diff
@@ -51,9 +52,6 @@ registerSuite('settings secondary emails', {
     return this.remote.then(clearBrowserState({ force: true }));
   },
 
-  afterEach: function() {
```
vbudhram: 👍

packages/fxa-content-server/tests/functional/sign_in_totp.js
```diff
@@ -265,11 +261,9 @@ registerSuite('TOTP - unverified session', {
   beforeEach: function() {
     email = TestHelpers.createEmail('sync{id}');
 
-    return this.remote.then(createUser(email, PASSWORD, { preVerified: true }));
-  },
-
-  afterEach: function() {
-    return this.remote.then(clearBrowserState({ force: true }));
```
vbudhram: I can't remember why I have `clearBrowserState` in the `afterEach` and `beforeEach`, maybe being extra cautious 🤷🏽‍♂️

packages/fxa-content-server/app/scripts/views/force_auth.js
shane-tomlinson: If a user comes from Firefox's settings page, /settings is opened with a uid/email. If the account for that email no longer exists, we send the user to force_auth to sign up again.

PR #3018, opened 2019-10-18
fix(log): Add err to redisConflict

vbudhram:
Adds the redis error to our custom error. Couldn't get conflict to trigger locally but hopefully there is something useful in there.


vbudhram:
Closing this since it can be replicated locally sometime.

PR #3021, opened 2019-10-18
fix(styles): clean up payment page UI a bit

### LINKED ISSUE #3013: UX nits on the payments page
making this more generic so i don't file a bunch of bugs:
* Start typing in the name input and the border turns red...it should only turn red on submit if a user does not fill it in
* the Stripe fields turn the text color red on invalid, but the non-stripe fields don't...we should be consistent here
* we should kill all of the empty input validation on the payment page b/c the whole form is blocked from submission

johngruen:
Fixes #3013

Hey Les and Dave, wanted to get your eyes on this PR since it removes some of the form validation on the payments page. Namely, it gets rid of onChange validation on inputs that seemed to make error tool tips a bit over eager. It also gets rid of empty field validation since you can't submit the form without completing fields and the tool-tips seemed really loud to me in the case of empty fields.

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -52,9 +52,16 @@ type StripeElementStyles = {
   lineHeight: string;
 };
 
+type StripeElementDirectStyles = {
+  color: string;
+};
+
 export function checkMedia(
   matched: boolean,
-  stripeElementStyles: { base: StripeElementStyles }
+  stripeElementStyles: {
+    base: StripeElementStyles;
+    invalid: StripeElementDirectStyles;
+  }
 ) {
   let lh = matched ? SMALL_DEVICE_LINE_HEIGHT : DEFAULT_LINE_HEIGHT;
   return Object.assign(stripeElementStyles, { base: { lineHeight: lh } });
@@ -282,7 +289,8 @@ export const PaymentForm = ({
       <div className="legal-blurb">
         <p>Mozilla uses Stripe for secure payment processing.</p>
         <p>
-          View the <a href="https://stripe.com/privacy">Stripe privacy policy</a>.
+          View the{' '}
+          <a href="https://stripe.com/privacy">Stripe privacy policy</a>.
         </p>
       </div>
     </Form>
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -40,6 +40,9 @@ let stripeElementStyles = {
     fontWeight: '500',
     lineHeight: '48px',
   },
+  invalid: {
+    color: '#0c0c0d',
+  },
 };
 
 type StripeElementStyles = {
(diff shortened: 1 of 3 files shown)


packages/fxa-payments-server/src/components/fields.tsx
```diff
@@ -82,12 +82,12 @@ export const Field = ({
     <div className={className}>
       <label>
         {label && <span className="label-text">{label}</span>}
-        {children}
         {tooltip && tooltipParentRef && validator.getError(name) && (
           <Tooltip parentRef={tooltipParentRef}>
             {validator.getError(name)}
           </Tooltip>
         )}
+        {children}
```
johngruen: flipping this order here actually buys us some sibling selector CSS from the payment server (`aside + input`) without changing the page layout

packages/fxa-payments-server/src/components/fields.tsx
```diff
@@ -116,7 +116,7 @@ export const Input = (props: InputProps) => {
   const onChange = useCallback(
     ev => {
       const { value } = ev.target;
-      validator.updateField({ name, value });
+      validator.updateField({ name, value, valid: true });
```
johngruen: flush invalidation when a user starts typing

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -157,9 +160,6 @@ export const PaymentForm = ({
         spellCheck={false}
         onValidate={value => {
           let error = null;
-          if (value !== null && !value) {
-            error = 'Please enter your name';
```
lmorchard: Removing this means that there is *no* validation error at all on an empty name field - I can submit without entering a name.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -203,7 +203,7 @@ export const PaymentForm = ({
             let error = null;
             value = ('' + value).substr(0, 5);
             if (!value) {
-              error = 'Zip code is required';
```
lmorchard: Same here - without validating on empty zip field, the submit button goes active as long as only the CC-related fields are filled and the checkbox is checked

packages/fxa-payments-server/src/components/fields.tsx
lmorchard: This also allows the zip code field to appear valid when filled with letters

lmorchard:
We can separate the display of error tooltips from marking fields valid, if that helps. But it looks like these changes practically disable validation altogether.

johngruen:
Yeah, that makes sense, i hadn't looked what was making the submit button turn enabled, and my changes seemed to be working so i rolled with it....Now i can see that if you clear the fields the submit button remains enabled, and yeah, all your comments make sense. Why don't i pick out the validation parts of this from the rest and file a separate issue for that.

meandavejustice:
@johngruen Les opened up a pr at #3027 that I just approved with the updates to the validation stuff. You can rebase with that and just change around this commit to only have the styles and I'll make sure it gets reviewed swiftly :)

johngruen:
Okay @lmorchard @meandavejustice i fixed this to just be the CSS tweaks.

Issue #2705, opened 2019-10-02
email-first: Password field not focused in desktop in signup flow


shane-tomlinson:
![no-focus-on-signup-password-screen](https://user-images.githubusercontent.com/848085/66020766-1e14d600-e4e0-11e9-86c8-3ece5f824505.gif)

My understanding is the password field *should* be focused on desktop, and *not* focused on mobile. At least that's how @ryanfeeley and I discussed the behavior in the past.

irrationalagent:
@vbudhram suggests that this may be a cause of a drop in our registration rate as of train 146. I tend to agree - it lines up in time, and we would only expect it to affect desktop, which seems to be the case: https://analytics.amplitude.com/mozilla-corp/chart/jozv8fk

shane-tomlinson:
fixed by #3025, closing.

PR #3025, opened 2019-10-18
fix(autofocus): Autofocus the sign-up page password input field

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/2705 and taken from https://github.com/mozilla/fxa/pull/2663 (Thanks @LZoog!)

Seems like worthwhile fix to get on train-148 since it is dropping registration rates.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
@@ -21,7 +21,7 @@
       <a href="/" class="use-different">{{#t}}Change email{{/t}}</a>
 
       <div class="input-row password-row">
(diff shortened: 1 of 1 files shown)


LZoog:
If I recall, this fix makes the field focus on mobile, which we _don't want_. 🤯

LZoog:
Taken from #2663:
(@dannycoates)
> We actually don't want autofocus at all on mobile so that you can see the form unobstructed on first load. I can tell from the video that the bug is from before my changes in train 146 so it's possible this is already fixed.

(@shane-tomlinson)
In views/base.js, the [focusAutofocusElement](https://github.com/mozilla/fxa/blob/3f3321edcb41c9725bb0b44bf2d3daa2dff68b57/packages/fxa-content-server/app/scripts/views/base.js#L953) only tries to autofocus on desktop and skips for devices with touch event support, so this seems like this fix should work. And it'd fix #2705.

So there might be a bug in hitting this `base.js` fn?

PR #3035, opened 2019-10-21
feat(email-first): Add the sync-suggestion UI to email-first flow

### LINKED ISSUE #3034: Email first does not have "Sync suggestion" UI
If you visit the "/signup" page directly by visiting https://accounts.firefox.com/signup, there is this UI:

Screenshot 2019-10-21 at 15 53 36

The email-first flow doesn't have that. We should add it.

shane-tomlinson:
fixes #3034


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3036, opened 2019-10-21
feat(payments): Add reactivate subscription confirmation modals

### LINKED ISSUE #3041: Show price on resubscribe
CF https://bugzilla.mozilla.org/show_bug.cgi?id=1589427

LZoog:
Fixes #3041

image

image


packages/fxa-payments-server/src/routes/Subscriptions/SubscriptionItem.tsx
```diff
@@ -267,26 +281,52 @@ const ReactivateSubscriptionPanel = ({
...
-        <div>
-          <p>You cancelled your subscription on {cancelledAtDate}.</p>
+    <>
+      {reactivateConfirmationRevealed && (
+        <DialogMessage onDismiss={hideReactivateConfirmation}>
+          <img alt="FPN" src={fpnImage} />
+          <h4 data-testid="resubscribe-message-title">
+            Want to keep using {plan.plan_name}?
+          </h4>
+          {/* TO DO: display card type, IE 'to the Visa card ending...' */}
```
LZoog: New issue for this: https://github.com/mozilla/fxa/issues/3037

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -2,14 +2,11 @@ import React, { useEffect, useContext } from 'react';
 import { Plan } from '../../../store/types';
 import { AppContext } from '../../../lib/AppContext';
 
+import fpnImage from '../../../images/fpn';
```
LZoog: I decided to import this since as of this PR, we're using this image in 3 places.

I created a separate issue to address product image scalability and making this less painful: https://github.com/mozilla/fxa/issues/3039

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
```diff
@@ -70,7 +70,31 @@ function init() {
...
           ...cancelledProps,
-          reactivateSubscription: linkTo('routes/Subscriptions', 'subscribed'),
+          reactivateSubscription: linkTo(
+            'routes/Subscriptions',
+            'reactivation confirmation'
+          ),
+        }}
+      />
+    ))
+    .add('reactivation confirmation', () => (
```
LZoog: @lmorchard - This displays `reactivate` and the post-success reactivation confirmation screen. I feel we should also add the interim modal before the user subscribes, but I'm not sure what the best approach is to modify a prop like I'm doing here (similar to `revealCancel`, which is not in Storybook either), and it does pop up when you click "Resubscribe" within the "reactivate" screen. Thoughts? 🤔

I'm not sure what the best way to kind of simulate a mouse click here (or simply update the state)

lmorchard: Just starting to get around to looking at this PR, but a general thing for storybook: Stories shouldn't need to modify props unless absolutely necessary.

But what you *can* do is set up several different stories with prop values for various states, then use [the links add-on](https://github.com/storybookjs/storybook/tree/next/addons/links#getting-started) to link between the stories on click to simulate UX flow.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -257,6 +258,27 @@ export const Subscriptions = ({
         </DialogMessage>
       )}
 
+      {console.log(reactivateSubscriptionStatus)}
```
LZoog: Whoops, missed console log, will remove this with any other fixes.

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
lmorchard: > I feel we should also add the interim modal before the user subscribes

I'm not sure what you mean here... Do you mean you want to add a modal after submitting payment on the product payment page?

> but I'm not sure what the best approach is to modify a prop like I'm doing here (similar to revealCancel, which is not in Storybook either)

The revealCancel state is [internal to the `CancelSubscriptionPanel` component](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Subscriptions/SubscriptionItem.tsx#L142), not a prop supplied from outside. So, storybook can't rig that up, because it only works through props.

> Thoughts?

I don't think I quite understand what you're trying to do

lmorchard: > I feel we should also add the interim modal before the user subscribes

Unless you mean before the user *un*subscribes... Because walking through this, I'm a little confused why we reveal a checkbox to confirm cancellation - but we display a modal dialog on the resubscribe side. Seems inconsistent to me that they don't use the same confirmation mechanism

LZoog: The first modal in my screenshots is rigged up the same way `revealCancel` is - an internal state (sorry for the confusion) triggered by the "Resubscribe" button click. You can't see it in Storybook without clicking the "Resubscribe" button. I wondered if there was a way I could get it to trigger so in Storybook we could have "reactivate," "reactivation modal," and "reactivate confirmed," so both modals could be shown there, but if Storybook only works through props, this may not be possible.

Hmm yeah, I agree it's inconsistent. 🤔

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
lmorchard: Yeah, looks like this needs to be removed

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
lmorchard: Oh, durr, I see what you mean now! Yeah, Storybook can't manipulate internal state, just props passed in externally. You *could* try moving that reactivation modal state up into the parent component, rather than using state internal to the subscription item, but that sounds a bit convoluted :/

johngruen:
@LZoog i am seeing the incorrect string on 123done locally

image

PR #3043, opened 2019-10-21
feat: use email instead of 'Anonymous' for support

### LINKED ISSUE #2912: Support tickets are filed from Anonymous User
Support tickets in Zendesk are showing up as from Anonymous User.  Email is populating correctly, but name is not.  

https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/support.js#L79

bbangert:
Because:

* FxA does not store the users actual user name, it is only
  submitted to Stripe for storage.
* The 'name' in FxA is user configured and many users have not
  entered a name.
* Sending emails to the user calling them 'Anonymous User' is not a
  great experience.
* Displaying a long list of 'Anonymous User' in the Zendesk UX is
  not helpful for skimming.

This commit:

* Uses the email address for the display name.

Closes #2912

### DIFF
--- a/packages/fxa-auth-server/lib/routes/support.js
+++ b/packages/fxa-auth-server/lib/routes/support.js
@@ -76,7 +76,7 @@ module.exports = (log, db, config, customs, zendeskClient) => {
           subject,
           requester: {
             email,
-            name: 'Anonymous User',
+            name: email,
           },
         };
         zendeskReq[config.zendesk.productNameFieldId] =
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/support.js
```diff
@@ -76,7 +76,7 @@ module.exports = (log, db, config, customs, zendeskClient) => {
           subject,
           requester: {
             email,
-            name: 'Anonymous User',
+            name: email,
```
vbudhram: Will this break anything if the user changed their email?

bbangert: We have an existing ticket to push email changes to Zendesk that is lower priority, which would affect this first. But no, this won't break anything.

PR #3046, opened 2019-10-21
fix(ios): Do not autofocus elements on iOS

### LINKED ISSUE #3040: E-mail first flow - mobile autofocuses on first password input
Expected behavior is that autofocus is present on desktop but _not_ mobile.

We've had a lot of issues open around this. See #3025 for a quick fix we got in to add autofocus to desktop that also added focus to mobile.

shane-tomlinson:
iOS autofocuses elements with the [autofocus] attribute
itself, without relying on our `focusAutofocusElement` method.

In `afterRender`, if it's a mobile device, remove
the `autofocus` attribute before writing to the DOM
to prevent iOS from autofocusing elements.

fixes #3040

@vbudhram or @LZoog - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/base.js
+++ b/packages/fxa-content-server/app/scripts/views/base.js
@@ -292,6 +292,18 @@ var BaseView = Backbone.View.extend({
             this._isErrorVisible = this.$('.error').hasClass('visible');
             this._isSuccessVisible = this.$('.success').hasClass('visible');
 
+            // make a huge assumption and say if the device has touch, it's a mobile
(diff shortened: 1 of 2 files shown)


LZoog:
Note: per a Slack discussion, we may want to hold on merging this as it may have inadvertently caused an improved user experience.

@davismtl and/or @ryanfeeley should confirm this should be merged.

shane-tomlinson:
> Per Slack, let's hold off on merging. It seems to have been a beneficial change even if it is less appealing visually. The outcome may be that we make it more visually appealing.

I'm going to close this PR, it can be brought back to life if we decide we want this behavior.

PR #3048, opened 2019-10-21
fix(links): stop converting links to text on Firefox iOS

### LINKED ISSUE #3047: Firefox iOS: Enable SUMO/help links
#  Description

On Firefox iOS, we now open a secondary webview to enable showing help links: https://github.com/mozilla-mobile/firefox-ios/pull/5639

If the class`visible-url` is removed from help links, they will open in an overlay webview as they already have `target=_blank`. The target attribute must not be removed of course, we need them to open in a new window. 

### Steps to reproduce

Currently, help links are displayed in full and are not tappable.

### Expected result

Tapping help links triggers target=_blank behaviour and opens them in an overlay webview. Pressing back on the overlay webview closes it and returns the user to the FxA webview.

vladikoff:
Fixes #3047

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
@@ -550,10 +550,6 @@ const BaseAuthenticationBroker = Backbone.Model.extend({
      */
     browserTransitionsAfterEmailVerification: true,
     /**
-     * Should external links be converted to text?
-     */
-    convertExternalLinksToText: false,
-    /**
      * Should the legacy signin/signup pages be disabled?
      */
     disableLegacySigninSignup: false,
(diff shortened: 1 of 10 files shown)


garvankeeley:
You are too fast! Should we try to align this somewhat with the release of v21 (~Dec 3rd)? That is when Firefox iOS will be able to handle these links.

vladikoff:
@garvankeeley ok we should be able to determine which train will deploy after Dec 3

vladikoff:
@clouserw could we schedule this to be part of a `> Dec 3` train?

bbangert:
Just a note that a rebase is now needed for this.

clouserw:
@vladikoff please rebase this and then we're ready to land it.  Thank you

vladikoff:
Rebased! We should merge this for 151 at some point, maybe once 150 won't be needing anymore point releases

PR #3053, opened 2019-10-21
fix(payments): Append time property to amplitude events

### LINKED ISSUE #3050: Payments metrics missing `time` key

6a68:
Fixes #3050.

@jbuck discovered this morning that the `time` property was missing from the payments server events, which is required. This PR pulls over the bit of code from the content server's ([stderr-metrics-collector](https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/server/lib/metrics-collector-stderr.js#L16-L21)) that appends the missing property to the event object.

I'm working on tests in another PR, and trying to figure  out if  we should just reuse the stderr-metrics-collector for both payments and content server, but wanted to get this out quickly and with a minimal patch, in case we want to add it to a point release this week.

### DIFF
--- a/packages/fxa-payments-server/server/lib/amplitude.test.js
+++ b/packages/fxa-payments-server/server/lib/amplitude.test.js
@@ -1,0 +1,108 @@
+const amplitude = require('./amplitude');
+const log = require('./logging/log')();
+jest.spyOn(log, 'info').mockImplementation(() => {});
+
+jest.mock('../config', () => ({
+  ...jest.requireActual('../config'),
+  get: key => {
+    switch (key) {
+      case 'amplitude':
+        return { enabled: true };
+    }
+  },
+}));
+
+const mocks = {
+  event: {
+    offset: 150,
+    type: 'amplitude.subPaySetup.view',
+  },
+  invalidEventType: {
+    offset: 150,
+    type: 'foo.bar.baz',
+  },
+  data: {
+    deviceId: '0123456789abcdef0123456789abcdef',
+    flowBeginTime: 1570000000000,
+    flowId: '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef',
+    flushTime: 9002,
+    perfStartTime: 9001,
+    requestReceivedTime: 11000,
+    view: 'product',
+  },
+  request: {
+    headers: {
+      'user-agent':
+        'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:72.0) Gecko/20100101 Firefox/72.0',
+    },
+  },
+  requestReceivedTime: 1570000001000,
+};
+
+const expectedOutput = {
+  app_version: '148.8',
+  device_id: '0123456789abcdef0123456789abcdef',
+  event_properties: {},
+  event_type: 'fxa_pay_setup - view',
+  op: 'amplitudeEvent',
+  os_name: 'Mac OS X',
+  os_version: '10.14',
+  session_id: 1570000000000,
+  user_properties: {
+    flow_id: '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef',
+    ua_browser: 'Firefox',
+    ua_version: '72.0',
+  },
+};
+
+describe('lib/amplitude', () => {
+  beforeEach(() => {
+    log.info.mockClear();
+  });
+  it('logs a correctly formatted message', done => {
+    amplitude(
+      mocks.event,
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
> I'm working on tests in another PR, and trying to figure out if we should just reuse the stderr-metrics-collector for both payments and content server

This does make me re-wonder whether we should have just reported payments events to the content server. I suppose that ship has sailed though, and hopefully in the end reporting metrics to each server is cleaner.

packages/fxa-payments-server/server/lib/routes/post-metrics.js
```diff
@@ -34,6 +34,13 @@ const BODY_SCHEMA = {
     .optional(),
 };
 
+function addTime(metricsEvent) {
+  // round the date to the nearest hour.
+  const today = new Date();
+  today.setMinutes(0, 0, 0);
+  metricsEvent.time = today.toISOString();
```
jbuck: This needs to be changed to unix epoch in ms to match the Amplitude API:

> The timestamp of the event in milliseconds since epoch. It will be set to the upload time by default. IE 1396381378123

6a68: Whoa, that's a good catch. Any idea why doesn't the content-server have to do the same?

Looking here: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/metrics-collector-stderr.js#L20

chenba: That line is for client metrics.  (Activity metrics and perf metrics also use ISO strings.)  Amplitude event.time is from: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/flow-event.js#L89

6a68:
Updated to do the minimum to get the correctly calculated time onto the amplitude event. There's so much refactoring to do between content-server, auth-server, and payments-server, it seemed better to fix this little bug separately.

6a68:
Actually, I think this should be rebased on top of Barry's PR that adds the flow timing to the payments server. Time calcs won't work without it.

6a68:
This will need to be rebased to pick up any final changes to Barry's PR before landing

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -442,3 +443,18 @@ function mapLocation(location) {
     };
   }
 }
+
+function estimateTime(times) {
+  const skew = times.received - times.sent;
+  return times.start + times.offset + skew;
+}
```
chenba: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/flow-event.js#L70 can make use of this now.

packages/fxa-payments-server/server/lib/amplitude.test.js
```diff
@@ -0,0 +1,108 @@
...
+  request: {
+    headers: {
+      'user-agent':
+        'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:72.0) Gecko/20100101 Firefox/72.0',
+    },
+  },
+  requestReceivedTime: 1570000001000,
+};
+
+const expectedOutput = {
```
jbuck: There's no `time` key added to the `expectedOutput` - Is this working as expected?

chenba: Good catch.  I think jest's `toMatchObject` is not particularly strict: the test is passing because the given object has all the properties of the expected object.  I'll see if there's a better assertion method we can use.

chenba: @jbuck I think you are correct that this patch didn't fix missing `time` property bug. ☹️

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -442,3 +443,18 @@ function mapLocation(location) {
...
+
+function mapTime(data, receivedTime) {
```
chenba: The call site passes in `data`, which does not have the `offset`.   `offset` is a property in each event.

packages/fxa-payments-server/server/lib/amplitude.test.js
chenba: I've reopened the issue: https://github.com/mozilla/fxa/issues/3050#issuecomment-547120175.  Thanks, @jbuck!

Issue #2508, opened 2019-09-13
[EPIC] Improve technical documentation


clouserw:
This Epic is specifically for improving our technical documentation.  I'm going to file issues for some areas so that we can assign them and budget time for them.

Documentation should live in https://github.com/mozilla/ecosystem-platform and we'll get that auto-deploying.

PRD: https://docs.google.com/document/d/1Fx0_iYtjEeV1L8uH4YlHNSqzW6uRPIag3Ve2ick9Hyg/edit?ts=5db86d15

┆Issue is synchronized with this [Jira Epic](https://jira.mozilla.com/browse/FXA-352)
┆Issue Number: FXA-352

clouserw:
I have a super rough draft outline of how this could mesh together: https://docs.google.com/document/d/14P94X3RnVb1SyatOp2H_NOPlj2jIvxrTDsusSOxK2zQ/edit

davismtl:
Added PRD to epic description.

davismtl:
Relying party docs have landed.

We're about half way there but here is a great demo of the progress that has been made:

https://drive.google.com/file/d/1j_8GXuu7Svo7eqwxF6Rfwj_iRhf_4Aw3/view

Will carryover into Q1.

davismtl:
Wil will look through these. We need to see if we want to revise in this meeting moving forward.

PR #3060, opened 2019-10-22
feat(keys): Report keysChangedAt in assertions and key ids.

rfk:
This PR extracts part of https://github.com/mozilla/fxa/pull/2570 into a standalone commit that we can safely merge and deploy without waiting for any tokenserver changes.

Connects to #490.

#### Background

The "durable sync" project wants to bucket storage of server-side sync data based on the id of the key used to encrypt it, including a timestamp component so that it can easily determine which data
is newest. Our OAuth scoped key ids include a leading timestamp designed for just this purpose.

Up until now we have been using `verifierSetAt` for that timestamp. This works but means that we generate a new kid whenever the user changes their password, even if that did not involve resetting the master encryption key `kB`. This means that durable sync will give users a new storage bucket, in  response to a password change, discarding the user's previous sync data even though their key had
not changed.

Instead, we need to explicitly track the timestamp at which key material actually changed, and report that in both OAuth key ids and in BrowserID assertions so that all Sync clients see a consistent
timestamp.  We already have a `keysChangedAt` database column from a previous migration, but haven't been making any use of it, and it currently defaults to the value of `verifierSetAt`.

#### This Change

With this commit, we start using `keysChangedAt` for the timestamp component of OAuth scoped key ids, and start reporting it in BrowserID assertions. For now the reported value will always be
the same as `verifierSetAt`, which will let us get the new logic into production without changing what timestamps are observed by the sync service.

A follow-up commit will be required to actually start tracking `keysChangedAt` separately from `verifierSetAt`. I'd like to get this smaller change into production first, to help ensure a smooth deployment. That follow-up commit commit also has to wait on a Tokenserver deployment, and will be the remainder of https://github.com/mozilla/fxa/pull/2570.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/assertion.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/assertion.js
@@ -32,6 +32,9 @@ const logger = require('./logging')('assertion');
 const { verifyJWT } = require('../../lib/serverJWT');
 
 const HEX_STRING = /^[0-9a-f]+$/;
+
+// FxA sends several custom claims, ref
+// https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/api.md#post-certificatesign
 const CLAIMS_SCHEMA = Joi.object({
   uid: Joi.string()
     .length(32)
@@ -63,6 +66,10 @@ const CLAIMS_SCHEMA = Joi.object({
     .max(3)
     .optional(),
   'fxa-profileChangedAt': Joi.number()
+    .integer()
+    .min(0)
+    .optional(),
+  'fxa-keysChangedAt': Joi.number()
     .integer()
     .min(0)
     .optional(),
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1173,70 +1173,43 @@ module.exports = (
             });
         }
 
-        function resetAccountData() {
-          let authSalt;
-          return random
-            .hex(32)
-            .then(hex => {
-              authSalt = hex;
-              password = new Password(authPW, authSalt, config.verifierVersion);
-              return password.verifyHash();
-            })
-            .then(verifyHashData => {
-              verifyHash = verifyHashData;
-
-              return setupKb();
-            })
-            .then(() => {
-              return db.resetAccount(accountResetToken, {
-                authSalt,
-                verifyHash,
-                wrapWrapKb,
-                verifierVersion: password.version,
-              });
-            })
-            .then(() => {
-              // Delete all passwordChangeTokens, passwordForgotTokens and
-              // accountResetTokens associated with this uid
-              return db.resetAccountTokens(accountResetToken.uid);
-            })
-            .then(() => {
-              // Notify the devices that the account has changed.
-              request.app.devices.then(devices =>
-                push.notifyPasswordReset(accountResetToken.uid, devices)
-              );
-
-              return db
-                .account(accountResetToken.uid)
-                .then(accountData => (account = accountData));
-            })
-            .then(() => {
-              return P.all([
-                request.emitMetricsEvent('account.reset', {
-                  uid: account.uid,
-                }),
-                log.notifyAttachedServices('reset', request, {
-                  uid: account.uid,
-                  generation: account.verifierSetAt,
-                }),
-                customs.reset(account.email),
-              ]);
-            });
-        }
-
-        function setupKb() {
+        async function resetAccountData() {
+          const authSalt = await random.hex(32);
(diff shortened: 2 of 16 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1173,70 +1173,44 @@ module.exports = (
...
-                  uid: account.uid,
-                  generation: account.verifierSetAt,
-                }),
-                customs.reset(account.email),
-              ]);
-            });
-        }
-
-        function setupKb() {
+        async function resetAccountData() {
```
rfk: Per https://github.com/mozilla/fxa/pull/2570#discussion_r326949419, I asyncified this function while I was in here. The logic should be unchanged, but will change slightly in the promised follow-up commit.

packages/fxa-auth-server/test/remote/account_reset_tests.js
```diff
@@ -21,239 +22,194 @@ describe('remote account reset', function() {
     });
   });
 
-  it('account reset w/o sessionToken', () => {
+  it('account reset w/o sessionToken', async () => {
```
rfk: Per https://github.com/mozilla/fxa/pull/2570#discussion_r327004747, I asyncified a bunch of the tests in this file, which was a nice cleanup but generated too much code churn that distracts from the review, so I stopped short of doing any more.  The important part of this change is the new `account reset updates keysChangedAt` test at the end of the file.

packages/fxa-auth-server/fxa-oauth-server/lib/assertion.js
```diff
@@ -66,6 +66,10 @@ const CLAIMS_SCHEMA = Joi.object({
     .integer()
     .min(0)
     .optional(),
+  'fxa-keysChangedAt': Joi.number()
```
vbudhram: More of a documentation nit (or add as part of https://github.com/mozilla/fxa/issues/2508), but I'm not sure we have a location to document our claims schema. Closest I found was https://github.com/mozilla/fxa/blob/eddea169bb45547776ab81779541edc1365686a9/packages/fxa-auth-server/docs/api.md#L3263. Most of the values are self explanatory but some are not.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1173,70 +1173,44 @@ module.exports = (
...
-            });
+            wrapWrapKb = await random.hex(32);
+            wrapKb = await password.unwrap(wrapWrapKb);
           }
+          // db.resetAccount() deletes all the devices saved in the account,
+          // so grab the list to notify before we call it.
+          const devicesToNotify = await request.app.devices;
+          // Reset the account, and delete any other outstanding passwordChangeTokens,
+          // passwordForgotTokens or accountResetTokens associated with that account.
+          await db.resetAccount(accountResetToken, {
```
vbudhram: I think the comment got outdated, might be better to remove or update to `account related tokens`.

packages/fxa-auth-server/test/remote/account_signin_verification_tests.js
```diff
@@ -497,6 +497,14 @@ describe('remote account signin verification', function() {
             1000 * 60 * 60,
           'lastAuthAt is plausible'
         );
+        assert.ok(
```
vbudhram: Thanks for adding this!

packages/fxa-auth-server/test/remote/recovery_key_tests.js
```diff
@@ -135,48 +136,61 @@ describe('remote recovery keys', function() {
...
-        assert.equal(
-          emailData.headers['x-template-name'],
-          'passwordResetAccountRecovery'
-        );
-        return client.keys();
-      })
-      .then(res => {
-        assert.equal(res.kA, keys.kA, 'kA are equal returned');
-        assert.equal(res.kB, keys.kB, 'kB are equal returned');
+  it('should change password and keep kB', async () => {
```
vbudhram: nit: `should reset password`

rfk:
> Are there any other endpoints that might potential need to return this?

I don't think so - I basically grepped for anything that was reporting `fxa-generation` and updated it to also report `fxa-keysChangedAt`.

packages/fxa-auth-server/fxa-oauth-server/lib/assertion.js
rfk: Thanks, I added some docs for other fields (like `fxa-tokenVerifier`) while I was in here.

rfk:
Merging with nits addressed.

dannycoates:
Y'all, srsly...

dannycoates:
TODO port this fcker to #2757

Issue #3017, opened 2019-10-18
It takes two "Revoke" clicks to actually revoke device access


rlr:
## Description

When I go to the FxA settings page to revoke access, it takes two clicks of the Revoke button to actually revoke.

### Steps to reproduce

- Sign into a site/app that uses FxA. https://monitor.firefox.com/ for example
- Go to https://accounts.firefox.com/settings
- Expand "Devices & apps"
- Click "Revoke" button for the site. "Firefox Monitor Last active a few seconds ago", for example.
- Click the "Refresh status" button below. The site revoked reappears.
- Click "Revoke button again". Now it's gone for good.

Screen recording: https://cl.ly/e8a6daf57bb6

### Expected result

It should take one click to revoke a site.

### Actual result

It takes two clicks

### Environment

I've  consistently seen this on stage and prod.

rlr:
I *think* I only see this when I've access the site recently. If I was a gambler, I'd bet one revoke click is revoking the access_token and the next click is revoking the refresh_token.

clouserw:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1589463

rfk:
Thanks for the report!

> If I was a gambler, I'd bet one revoke click is revoking the access_token and the next click is
> revoking the refresh_token.

I suspect it's the opposite - the first click is revoking the refresh_token but leaving active access_tokens in the db. The second click destroys the access tokens and removes the client from the list entirely.

Addressing https://github.com/mozilla/fxa/issues/1249 would fix this issue, but involves a migration on the oauth `tokens` table (so that we can know which refresh_token was used to generate a particular access_token), and I get the impression we're a bit hesitant about doing those at the moment...

juliemccracken:
Made this a blocker after discussion in this morning's SubPlat meeting. Let's see what needs to be done about this.

shane-tomlinson:
Caused by #2480 and #1249

davismtl:
Are #2584 and #1802 related?

rfk:
> Are #2584 and #1802 related?

I don't believe so.

shane-tomlinson:
Basic idea outlined in https://github.com/mozilla/fxa/issues/1249#issuecomment-544720161. I'm happy to do this tomorrow, unless someone beats me to it.

Issue #1249, opened 2019-05-23
Destroying a `refresh_token` should destroy all `access_tokens` that were created from it


rfk:
Ref https://github.com/mozilla/blurts-server/issues/992#issuecomment-495366822

If an RP destroys their refresh token, they might still have some active access tokens, and hence might still show up in the "devices and apps" list despite considering themselves "signed out".

It seems reasonable to me that destroying a `refresh_token` should also destroy any `access_tokens` that were created from it.

Implementing this would be complicated by two things:
* We don't currently link access tokens back to their originating refresh token, if any
* We're trying to reduce the lifetime of access tokens, and if the lifetime gets sufficiently small, perhaps there's no benefit to explicitly revoking them at all.

shane-tomlinson:
I thought this was mentioned in the spec too.

> We're trying to reduce the lifetime of access tokens, and if the lifetime gets sufficiently small, perhaps there's no benefit to explicitly revoking them at all.

I doubt we'll drop the default below 24 hours any time soon, so revoking will still be beneficial. There is already [confusion around what revoking OAuth credentials means](https://github.com/mozilla/fxa/issues/1203), anything we can do to minimize surprises is beneficial.

groovecoder:
FWIW I've got a PR to fix the issue for Monitor by revoking both access tokens when someone removes Firefox Monitor. https://github.com/mozilla/blurts-server/pull/999 So, no big pressure from our side.

rfk:
Hrm, I'll probably need to figure something out for this as part of #1262 because otherwise disconnecting a refresh-token client will not work correctly.

shane-tomlinson:
Now that access token pruning takes place and the tokens table, adding new columns to the table shouldn't be too difficult. We could of course add a new table with the bindings too.

rfk:
Would it be unreasonable for destroying a refresh token to destroy *all* access tokens for the associated `client_id`? That risks destroying tokens held by other instances of the same client, but they should be prepared to handle those tokens being invalidated anyway and just generate some new ones.

shane-tomlinson:
> Would it be unreasonable for destroying a refresh token to destroy all access tokens for the associated client_id?

I'm not sure about this, it feels sloppy and like it pushes work onto RPs that we should be doing ourselves.

shane-tomlinson:
> I'm not sure about this, it feels sloppy and like it pushes work onto RPs that we should be doing ourselves.

[...]

PR #3062, opened 2019-10-22
refactor(db): Remove the memory DB in the OAuth server.

shane-tomlinson:


shane-tomlinson:
We've talked about it and lazy me pulled the trigger. I was making an update to the MySQL DB today for #1249, #3017 and didn't want to update the memory DB.

@dannycoates - whatcha think?

dannycoates:
TODO port to #2757

PR #3063, opened 2019-10-22
fix(oauth): Revoke access tokens when revoking a refresh token

### LINKED ISSUE #1249: Destroying a `refresh_token` should destroy all `access_tokens` that were created from it

### LINKED ISSUE #3017: It takes two "Revoke" clicks to actually revoke device access

shane-tomlinson:
All access tokens for a client_id, uid pair are revoked whenever
revoking a refresh token to avoid phantom entries in the
user's devices & apps list.

Before this PR, the problem was revoking a refresh token would
only revoke the refresh token and none of the access tokens created
from it. This led to a user revoking a refresh token, but then
the access tokens created from it showing up in the user's devices
& apps list, and needing to revoke those separately.

Because we don't associate access tokens and refresh tokens, we can't
delete just the access token associated with the refresh token. Instead,
this PR uses a blunt instrument and revokes *all* the access tokens
for the user/client_id. If the user has multiple active refresh tokens,
then the clients will use those refresh tokens to get new access
tokens for the ones that were revoked prematurely.

fixes #1249
fixes #3017 

@vladikoff, @vbudhram, @rfk - r?

### DIFF
--- a/packages/123done/oauth.js
+++ b/packages/123done/oauth.js
@@ -48,6 +48,7 @@ function verifyIdToken(oauthConfig, token) {
 
 function setupOAuthFlow(req, action, options = {}, cb) {
   var params = {
+    access_type: 'offline',
     client_id: config.client_id,
     pkce_client_id: config.pkce_client_id,
     redirect_uri: config.redirect_uri,
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
@@ -593,8 +593,16 @@ MysqlStore.prototype = {
 
   /**
    * Delete a specific refresh token, for some clientId and uid.
-   * We don't actually need to know the clientId or uid in order to delete a refresh token,
-   * but since they're available we use them a an additional check.
+   * Also deletes *all* access tokens for the clientId and uid combination,
+   * otherwise the refresh token is deleted but none of the access tokens
+   * created from that refresh token are, leaving ghost access tokens
+   * to appear in the users devices & apps list. See:
+   *
+   * https://github.com/mozilla/fxa/issues/1249
+   * https://github.com/mozilla/fxa/issues/3017
+   *
+   * If a user has multiple refresh tokens for a given client_id, clients
+   * will use their active refresh tokens to get new access tokens.
    *
    * @param {String} refreshTokenid Refresh Token ID as Hex
    * @param {String} clientId Client ID as Hex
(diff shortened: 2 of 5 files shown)


packages/123done/oauth.js
```diff
@@ -48,6 +48,7 @@ function verifyIdToken(oauthConfig, token) {
 
 function setupOAuthFlow(req, action, options = {}, cb) {
   var params = {
+    access_type: 'offline',
```
shane-tomlinson: I added this so that 123done would receive refresh tokens and we could add a functional test to the content server to ensure this is fixed.

packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
```diff
@@ -606,12 +614,22 @@ MysqlStore.prototype = {
...
+      DELETE_ACTIVE_TOKENS_BY_CLIENT_AND_UID,
+      [buf(clientId), buf(uid)]
+    );
+
+    const deleteRefreshTokens = this._write(
+      DELETE_REFRESH_TOKEN_WITH_CLIENT_AND_UID,
+      [buf(refreshTokenId), buf(clientId), buf(uid)]
+    );
+
+    const [, refreshTokenRes] = await P.all([
```
vbudhram: Should this be done in parallel? Is there a scenario where deleting a refresh tokens could fail and we don't want to delete access tokens?

Since this is parallel it is more performant though.

packages/fxa-auth-server/fxa-oauth-server/test/db/index.js
```diff
@@ -623,4 +623,59 @@ describe('db', function() {
...
+      );
+      assert.isTrue(wasRevoked);
+
+      assert.notOk(await db.getRefreshToken(refreshTokenIdHash));
+
+      // all access tokens for the client_id, uid should be revoked as well
+      // as the refresh token. Clients that do the right thing will use
+      // active refresh tokens to get new access tokens for the ones that
+      // have been revoked. Deleting all the access tokens is to prevent
+      // ghost access tokesn from appearing in the user's devices & apps list.
```
vbudhram: nit: `tokesn` -> `tokens`

packages/fxa-auth-server/fxa-oauth-server/lib/db/mysql/index.js
shane-tomlinson: Good suggestion. This is a pretty low traffic route, so doing those serially should be fine.

dannycoates:
TODO port to #2757

PR #3070, opened 2019-10-22
refactor(db): Remove the memory DB in the Auth server

shane-tomlinson:
The original idea of the memory DB was so developers
didn't need to install MySQL on their local machines.

Now that we run MySQL in a container, running MySQL
is no longer the burden it once was. This old memory
DB is just crufty code that has to be maintained and
often slows us down.

Goes along with #3062 where the memory DB was removed
for the OAuth server.


dannycoates:
https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-db-mysql/index.js#L9

shane-tomlinson:
I dug in a bit further here today and was finally able to get the tests running, however each test module starts and stops its own test servers, eventually MySQL complains about having too many open connections. Will dig in more as I get time.

shane-tomlinson:
Closing until I can get back to this.

PR #3071, opened 2019-10-22
feat(metrics): add flow perf metrics to payments server pages

### LINKED ISSUE #996: Ensure we're monitoring client-side performance data for new Subscriptions pages

chenba:
Fixes #996

@mozilla/fxa-devs r?


packages/fxa-content-server/server/lib/flow-event.js
```diff
@@ -10,6 +10,7 @@ const flowMetrics = require('./flow-metrics');
 const geolocate = require('./geo-locate');
 const os = require('os');
 
+const VALIDATION_PATTERNS = require('./validation').PATTERNS;
```
6a68: :+1: nice! great to see the validation getting extracted to a central place

packages/fxa-payments-server/server/lib/flow-performance.js
```diff
@@ -0,0 +1,125 @@
+const os = require('os');
+const log = require('./logging/log')();
+
+const MAX_DATA_LENGTH = 100;
+const VERSION = 1;
+const HOSTNAME = os.hostname();
+const PERFORMANCE_TIMINGS = [
```
6a68: Functions and constants duplicated from content server's `lib/flow-event.js`, like `PERFORMANCE_TIMINGS`, `limitLength`, and `isValidTime`, should instead be extracted to fxa-shared. This will ensure we don't fix a bug in one place, but not the other.

packages/fxa-payments-server/server/lib/flow-performance.js
```diff
@@ -0,0 +1,125 @@
...
+        },
+        data,
+        request
+      );
+    }
+  });
+};
+
+function writePerfEvent(event, data, request) {
+  const eventData = {
```
6a68: This event format is very subtly different from the format logged by `logFlowEvent` in content-server `lib/flow-event.js`. It appears that `country` and `region` are present there, but omitted here. Might be better to use one event format for both spots, unless there's a really good reason to omit those properties here, and extract the duplicated code out to fxa-shared.

chenba: I think I left those out because they are not imported into Redshift.  I will add them for parity.

packages/fxa-payments-server/src/lib/config.test.ts
```diff
@@ -81,6 +94,24 @@ describe('readConfigFromMeta', () => {
     readConfigFromMeta(baseHeadQuerySelector);
     expect(config.featureFlags).toEqual(mockFeatureFlags);
   });
+
+  it('does not throws an error if meta[name="fxa-perf-start-time"] is empty', () => {
```
6a68: "does not throw"

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -121,6 +124,9 @@ describe('routes/Product', () => {
     const displayName = useDisplayName ? 'Foo Barson' : undefined;
     const apiMocks = initApiMocks(displayName);
     const { findByText, queryByText, queryByTestId } = render(<Subject />);
+    if (window.onload) {
+      window.onload(new Event('testo'));
```
6a68: `onload` expects a function

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -81,13 +82,20 @@ export const Product = ({
...
-  const { queryParams, locationReload } = useContext(AppContext);
+  const { config, locationReload, queryParams } = useContext(AppContext);
 
   const {
     plan: planId = '',
     activated: accountActivated = false,
   } = queryParams;
 
+  // There is no way to do this with a React Hook. We need the
+  // `navgationTiming.domComplete` value to calculate the "client" perf metric.
```
6a68: typo: `navigationTiming`

packages/fxa-payments-server/src/routes/Subscriptions/index.test.tsx
```diff
@@ -145,10 +143,15 @@ describe('routes/Subscriptions', () => {
     const { findByTestId, queryAllByTestId, queryByTestId } = render(
       <Subject />
     );
+    if (window.onload) {
+      window.onload(new Event('testo'));
```
6a68: same thing, need a function inside `onload`

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -113,6 +115,13 @@ export const Subscriptions = ({
 }: SubscriptionsProps) => {
   const { config, locationReload, navigateToUrl } = useContext(AppContext);
 
+  // There is no way to do this with a React Hook. We need the
+  // `navgationTiming.domComplete` value to calculate the "client" perf metric.
```
6a68: `navigationTiming` typo again

packages/fxa-payments-server/server/lib/flow-performance.js
```diff
@@ -0,0 +1,109 @@
...
+    !flowId ||
+    !navigationTiming
+  ) {
+    return;
+  }
+
+  // Since the flow originates in the content-server, the flow begin time should
+  // be before the perf start time.
+  if (perfStartTime < flowBeginTime) {
+    return;
```
6a68: It seems like we could silently lose perf metrics data here. Should we log something in this case?

packages/fxa-payments-server/Dockerfile
```diff
@@ -27,17 +27,27 @@ WORKDIR /app
...
+# Except, to avoid installing redis, we're just copying over the files
 # that are explicitly needed.
 COPY --chown=app:app ["fxa-shared/metrics/user-agent.js", "../fxa-shared/metrics/user-agent.js"]
 COPY --chown=app:app ["fxa-shared/metrics/amplitude.js", "../fxa-shared/metrics/amplitude.js"]
+COPY --chown=app:app ["fxa-shared/metrics/flow-performance.js", "../fxa-shared/metrics/flow-performance.js"]
 COPY --chown=app:app ["fxa-shared/express/index.js", "../fxa-shared/express/index.js"]
 COPY --chown=app:app ["fxa-shared/express/routing.js", "../fxa-shared/express/routing.js"]
 COPY --chown=app:app ["fxa-shared/express/cors.js", "../fxa-shared/express/cors.js"]
+COPY --chown=app:app ["fxa-shared/express/remote-address.js", "../fxa-shared/express/remote-address.js"]
+COPY --chown=app:app ["fxa-shared/express/remote-address.js", "../fxa-shared/express/geo-locate.js"]
```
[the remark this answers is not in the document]
chenba: 🤦‍♂

packages/fxa-payments-server/server/lib/flow-performance.js
chenba: Since we don't validate the flow params, `flowBeginTime` is ultimately untrustworthy.  And, it is not used in calculating the perf stats.  I think I could just remove this.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
chenba: This is calling the `onload` handler, not registering a handler to it.  I'll update it to a `dispatchEvent`.

PR #3076, opened 2019-10-22
fix(metrics): Update app_version to send complete train version number

### LINKED ISSUE #3028: Add Dot Release Train Versions To Amplitude Metrics
In Amplitude, we only track the under the `Version` property the high-level train number.
Train 145, 146, 147, etc...

Can we have that value reflect the dot releases?

It was really complicated in the past prior to being a mono-repo but I'm hoping it is more feasible now.

Why? It will make it easier to track the impact of fixes and patches that were shipped after a train.

vbudhram:
Fixes #3028 

~Marking as draft since I fully expect some test bustage.~

Example amplitude log
```
amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_email_first - engage","time":1571938016089,"device_id":"0222132c0469472496d5592b80a230f1","session_id":1571938014298,"app_version":"148.8","language":"en","os_name":"Mac OS X","os_version":"10.14","event_properties":{"service":"sync"},"user_properties":{"entrypoint":"fxa_discoverability_native","flow_id":"3d567c9a9d914d935ef2babbffccb91735be2a2c065cdb28084e72f28745c41d","ua_browser":"Firefox","ua_version":"70.0","$append":{"fxa_services_used":"sync","experiments":["email_first_treatment"]}}}
```

### DIFF
--- a/packages/fxa-content-server/tests/server/amplitude.js
+++ b/packages/fxa-content-server/tests/server/amplitude.js
@@ -30,7 +30,7 @@ const amplitude = proxyquire(path.resolve('server/lib/amplitude'), {
   },
   './logging/log': () => logger,
 });
-const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(pkg.version)[1];
+const APP_VERSION = /([0-9]+)\.([0-9])$/.exec(pkg.version)[0];
 
 registerSuite('amplitude', {
   beforeEach: function() {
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/tests/server/amplitude.js
```diff
@@ -30,7 +30,7 @@ const amplitude = proxyquire(path.resolve('server/lib/amplitude'), {
   },
   './logging/log': () => logger,
 });
-const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(pkg.version)[1];
+const APP_VERSION = pkg.version;
```
6a68: This will change the app version from, say, `148`, to `1.148.3`.

I think we might want to continue to strip out the initial `1.`?

vbudhram: I was indifferent to the `1.` so updated to string it out 👍🏽

vbudhram:
This is ready for review!

PR #3111, opened 2019-10-23
feat(payments): more explicit auto-focus for name field

### LINKED ISSUE #3098: Autofocus Name field on Payments Page
Fewer clicks = Faster typing = faster paying

lmorchard:
fixes #3098

### DIFF
--- a/packages/fxa-payments-server/src/components/fields.tsx
+++ b/packages/fxa-payments-server/src/components/fields.tsx
@@ -45,6 +45,7 @@ type FieldProps = {
   className?: string;
   maxLength?: number;
   minLength?: number;
+  autoFocus?: boolean;
 };
 
 type FieldHOCProps = {
@@ -60,6 +61,7 @@ export const Field = ({
   initialValue = null,
   tooltip = true,
   required = false,
+  autoFocus = false,
   label,
   className = 'input-row',
   children,
@@ -74,6 +76,18 @@ export const Field = ({
     [name, required, fieldType]
   );
   /* eslint-enable react-hooks/exhaustive-deps */
+
+  const focusElement =
+    autoFocus && tooltipParentRef && tooltipParentRef.current;
+  useEffect(() => {
+    if (focusElement && focusElement.focus) {
+      // TODO: figure out how to get rid of this setTimeout because it's
+      // a race condition - but, something seems to steal focus otherwise
+      setTimeout(() => {
+        focusElement.focus();
+      }, 200);
+    }
+  }, [focusElement]);
 
   return (
     <div className={className}>
@@ -124,6 +138,7 @@ export const Input = (props: InputProps) => {
     initialValue,
     onValidate = defaultInputValidator,
     tooltip,
+    autoFocus,
     required = false,
     className,
     ...childProps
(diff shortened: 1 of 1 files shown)


packages/fxa-payments-server/src/components/fields.tsx
```diff
@@ -75,6 +77,14 @@ export const Field = ({
   );
   /* eslint-enable react-hooks/exhaustive-deps */
 
+  useEffect(() => {
+    if (autoFocus && tooltipParentRef && tooltipParentRef.current) {
+      setTimeout(() => {
```
lmorchard: So, I'm actually kind of unsure about this PR: This seems to intermittently not work on the payment page. 😞 Might need a second opinion on that page - I think something else on there is stealing focus, but I can't tell what. This `setTimeout` is a brute force attempt to steal back focus after a vague wait.

LZoog: Yeah, this is odd, the `body` doesn't want to lose focus. The `useLayoutEffect` hook has no affect and the examples I've found online suggest to do it the way you are here. I'll see if I can keep digging since this does feel pretty hacky😅

lmorchard: Yeah, I don't like what I've done here at all. 😓

johngruen:
@LZoog we maybe think u have a concept here. Please feel free to un-assign if that's not true?

packages/fxa-payments-server/src/components/fields.tsx
LZoog: Well, I tried conditionally rendering `autoFocus` based on `childProps` in `input` rather than fields, then playing with the hook and hook dependencies, and looking up a ton of stuff on SO with no dice... I've set `setTimeOut` to `0` and thrown a bunch of logs in product, payment form, and fields, to see if I could catch something weird with render timing, and even when `setTimeout` logs at the very end, `0` consistently does not work so it's very difficult to identify what is causing the `body` to regain focus.

Have we not had React dev tools working the entire time on payments? It tells me it's not a React application. That could come in handy for debugging something like this for highlighting rerenders. 🤔 I also can't connect to dev tools React on staging:

image

If we can get React dev tools working to help see what's happening (maybe I'm missing something on getting it working?) this might be worth revisiting before merging... Otherwise, we might want to ship it with a "TO DO" comment written above. 😩

packages/fxa-payments-server/src/components/fields.tsx
```diff
@@ -75,6 +77,18 @@ export const Field = ({
   );
   /* eslint-enable react-hooks/exhaustive-deps */
 
+  const focusElement =
```
lmorchard: Gave this a try, thinking that maybe re-running the effect when the available element ref changes would help. Doesn't seem to have done, though :/

lmorchard:
@LZoog You might want to take one more peek at this. I added the TODO comment and slightly tweaked the event hook to fire when there's an element reference change. Though that latter thing didn't seem to really solve the conundrum. 

FWIW, I got React Dev Tools working locally by turning off CSP, but that didn't quite seem to clue me into anything either

lmorchard:
Dang it. CI failure on [a security advisory](https://www.npmjs.com/advisories/1300)

lmorchard:
Think I'll merge this for now, we can come back at it some other time. I do wonder if the stripe elements are doing some odd things with focus, since they're each a separate iframe

PR #3128, opened 2019-10-25
feat(auth-server): capture req data to Sentry

### LINKED ISSUE #3110: [fxa-auth-server] Sentry: record release and additional call context for backend service calls
We don't record the release information for fxa-auth-server so we can't easily determine if an issue was resolved in a new release. We should record that as well as more context for the backend service calls so we can easily see in Sentry which service was the problem.

bbangert:
Because:

* Default Sentry did not annotate stack traces with useful data
  such as the release information, request that was in flight when the
  traceback occurred, etc.
* Had no way to leave Sentry breadcrumbs for the current request
  scope.
* Nested the getVersion within a function making it inaccessible to
  other callers.

This commit:

* Add's Hapi request data to the Sentry report.
* Add's a Sentry.Scope object to the Hapi request object for use by
  other code to add bread-crumbs, tags, etc.
* Refactors the getVersion to its own function so that other calling
  code (such as Sentry.init) can report the release data.

Closes #3110

### DIFF
--- a/packages/fxa-auth-server/lib/error.js
+++ b/packages/fxa-auth-server/lib/error.js
@@ -1209,7 +1209,22 @@ AppError.invalidOrExpiredOtpCode = () => {
   });
 };
 
-AppError.backendServiceFailure = (service, operation) => {
+AppError.backendServiceFailure = (service, operation, extra) => {
+  if (extra) {
+    return new AppError(
+      {
+        code: 500,
+        error: 'Internal Server Error',
+        errno: ERRNO.BACKEND_SERVICE_FAILURE,
+        message: 'A backend service request failed.',
+      },
+      {
+        service,
+        operation,
+        ...extra,
+      }
+    );
+  }
   return new AppError(
     {
       code: 500,
--- a/packages/fxa-auth-server/lib/sentry.js
+++ b/packages/fxa-auth-server/lib/sentry.js
@@ -1,0 +1,138 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const Hoek = require('@hapi/hoek');
+const Sentry = require('@sentry/node');
+
+const getVersion = require('./version').getVersion;
+
+// Matches uid, session, oauth and other common tokens which we would
+// prefer not to include in Sentry reports.
+const TOKENREGEX = /[a-fA-F0-9]{32,}/gi;
+const FILTERED = '[Filtered]';
+const URIENCODEDFILTERED = encodeURIComponent(FILTERED);
+
+/**
+ * Filters all of an objects string properties to remove tokens.
+ *
+ * @param {Object} obj Object to filter values on
+ */
+function filterObject(obj) {
+  if (typeof obj === 'object') {
+    for (const [key, value] of Object.entries(obj)) {
+      if (typeof value === 'string') {
+        obj[key] = value.replace(TOKENREGEX, FILTERED);
+      }
+    }
+  }
+  return obj;
+}
+
+/**
+ * Filter a sentry event for PII in addition to the default filters.
+ *
+ * Current replacements:
+ *   - A 32-char hex string that typically is a FxA user-id.
+ *
+ * Data Removed:
+ *   - Request body.
+ *
+ * @param {Sentry.Event} event
+ */
+function filterSentryEvent(event, hint) {
+  if (event.breadcrumbs) {
+    for (const bc of event.breadcrumbs) {
+      if (bc.message) {
+        bc.message = bc.message.replace(TOKENREGEX, FILTERED);
+      }
+      if (bc.data) {
+        bc.data = filterObject(bc.data);
+      }
+    }
+  }
+  if (event.request) {
+    if (event.request.url) {
+      event.request.url = event.request.url.replace(TOKENREGEX, FILTERED);
+    }
+    if (event.request.query_string) {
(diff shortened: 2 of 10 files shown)


packages/fxa-auth-server/lib/sentry.js
```diff
@@ -0,0 +1,106 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const Hoek = require('@hapi/hoek');
+const Sentry = require('@sentry/node');
+
+const getVersion = require('./version').getVersion;
+
+const UIDREGEX = /[a-fA-F0-9]{32}/gi;
```
rfk: This will also match various other kinds of token, e.g. session token ids and oauth tokens and what-not. Those things probably won't be showing up in any of the places you're filtering though, and if they do, well it's probably the right thing to filter those as well!

bbangert: Renaming to reflect that with a comment, any other places in the request data that such a token might be?

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -0,0 +1,109 @@
...
+'use strict';
+
+const Hoek = require('@hapi/hoek');
+const Sentry = require('@sentry/node');
+
+const getVersion = require('./version').getVersion;
+
+// Matches uid, session, oauth and other common tokens which we would
+// prefer not to include in Sentry reports.
+const TOKENREGEX = /[a-fA-F0-9]{32}/gi;
```
rfk: You could consider making this consume 32-or-more hex chars, so that longer tokens don't wind up looking like `[Filtered][Filtered]`.

packages/fxa-auth-server/lib/sentry.js
rfk: I think you covered them - we're already not logging the obvious things like request body and auth headers, right?

packages/fxa-auth-server/lib/sentry.js
bbangert: Sold! 😄

packages/fxa-auth-server/lib/sentry.js
bbangert: Auth header, no. Body might be captured. 🤔

bbangert: Updated to remove request data entirely.

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+  }
+  if (event.request) {
+    if (event.request.url) {
+      event.request.url = event.request.url.replace(TOKENREGEX, FILTERED);
+    }
+    const headers = event.request.headers;
+    if (headers) {
+      if (headers.origin) {
```
vbudhram: These the only headers? Could we go through all headers and replace fields that match?

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -0,0 +1,116 @@
...
+ * Current replacements:
+ *   - A 32-char hex string that typically is a FxA user-id.
+ *
+ * Data Removed:
+ *   - Request body.
+ *
+ * @param {Sentry.Event} event
+ */
+function filterSentryEvent(event) {
+  if (event.breadcrumbs) {
```
vbudhram: This didn't filter breadcrumbs for me. Looks like the object is structured like this

Screen Shot 2019-10-25 at 12 18 08 PM

Ref: https://sentry.prod.mozaws.net/operations/fxa-ben-test/issues/6539388/

packages/fxa-auth-server/lib/sentry.js
bbangert: Sure, added that.

packages/fxa-auth-server/lib/sentry.js
bbangert: I pushed another change that hopefully will grab these, not sure why I couldn't iterate through the event to re-assign them.

vbudhram: Hmm doesn't look like the breadcrumbs are being [filtered](https://sentry.prod.mozaws.net/operations/fxa-ben-test/issues/6539808/?query=is:unresolved). When testing locally, I couldn't trigger a breakpoint in `beforeBreadcrumb` but could in `filterSentryEvent`. Is there another config that would need to be enabled?

packages/fxa-auth-server/lib/error.js
```diff
@@ -1224,6 +1224,22 @@ AppError.backendServiceFailure = (service, operation) => {
   );
 };
 
+AppError.backendServiceFailure = (service, operation, extra) => {
```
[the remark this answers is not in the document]
bbangert: Doh, I failed my function overload. Folded in correctly now I hope.

packages/fxa-auth-server/lib/sentry.js
bbangert: Ok, *now* it should be filtering them. I configured it incorrectly it seems.

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * Data Removed:
+ *   - Request body.
+ *
+ * @param {Sentry.Event} event
+ */
+function filterSentryEvent(event, hint) {
+  if (event.breadcrumbs) {
+    for (const bc of event.breadcrumbs) {
+      bc.message = bc.message.replace(TOKENREGEX, FILTERED);
```
vbudhram: Locally, `message` is undefined for me, https://sentry.prod.mozaws.net/operations/fxa-ben-test/issues/6540571/?query=is:unresolved

PR #3129, opened 2019-10-25
fix(logging): Log excluded amplitude events under a different name.

### LINKED ISSUE #3160: Restore logging of oauth activity events to BigQuery
#2996 had the desired effect of avoiding sending an overwhelming deluge of oauth activity event into amplitude, but it had the side-effect of making those events unavailable for analysis in other tools such as BigQuery. In particular, they're no longer available for [export to CDP](https://bugzilla.mozilla.org/show_bug.cgi?id=1573359).

Let's restore them.

I put up a trial PR for this in #3129, but @jbuck said it may not be necessary, we perhaps we can instead revert #2996 and move the amplitude filtering into infrastructure.

rfk:
#2996 had the desired effect of avoiding sending an overwhelming deluge of oauth activity event into amplitude, but it had the side-effect of making those events unavailable for analysis in other tools such as BigQuery. In particular, they're no longer available for [export to CDP](https://bugzilla.mozilla.org/show_bug.cgi?id=1573359).

Would it be OK if we continue to log them, but under a different name so that they don't get sent off to amplitude?

Fixes #3160.

---

Trying out @bbangert's suggested format for commit messages:

Because:

* It's useful to analyse oauth activity events in BigQuery even
  if they're not available in amplitude.
* We need to provide an accurate summary of what services a user
  has been active on for export the Customer Data Platform.
* But we don't want to overwhelm amplitude with oauth activity events.

This patch:

* Restores logging of oauth activity events for all clients, even
  the really heavy ones like Firefox Desktop.
* But writes them out under a different event name, so they show up
  in raw log analysis but don't get exported to amplitude.

### DIFF
--- a/packages/fxa-auth-server/fxa-oauth-server/lib/metrics/amplitude.js
+++ b/packages/fxa-auth-server/fxa-oauth-server/lib/metrics/amplitude.js
@@ -30,10 +30,8 @@ const EVENTS = {
 
 const FUZZY_EVENTS = new Map([]);
 
-function sane(event) {
-  if (!event) {
-    return false;
-  }
+function shouldExclude(event) {
+  // Exclude certain high-volume clients from logging events into amplitude.
   const props = event.event_properties;
   const excluded =
     (props.service === 'fennec-stage' &&
@@ -42,7 +40,7 @@ function sane(event) {
       props.oauth_client_id === '5882386c6d801776') ||
     (props.service === 'firefox-ios' &&
       props.oauth_client_id === '1b1a3e44c54fbb58');
-  return !excluded;
+  return excluded;
 }
 
 module.exports = (log, config) => {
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/fxa-oauth-server/lib/metrics/amplitude.js
```diff
@@ -79,8 +77,12 @@ module.exports = (log, config) => {
       eventData
     );
 
-    if (sane(amplitudeEvent)) {
-      log.info('amplitudeEvent', amplitudeEvent);
+    if (amplitudeEvent) {
+      if (shouldExclude(amplitudeEvent)) {
+        log.info('excludedAmplitudeEvent', amplitudeEvent);
```
jbuck: I think we'll need to modify the `op` key on the object here too :/

Or we need to switch the Amplitude processing over to GCP entirely, where we can filter out events again

dannycoates:
waiting on #2757

bbangert:
#2757 has merged. @jbuck are changes still needed?

rfk:
From slack comments (reflected in https://github.com/mozilla/fxa/issues/3160) I think @jbuck had an alternate plan for this that would let us do a plain revert #2996 and then discard the events at a later stage of the processing pipeline, but I don't have any more details...

juliemccracken:
We'll deploy Amplitude processing over the GCP now that we're done with Skyline.  This can happen today. As a result, we won't need rfkelly's fix.

rfk:
I'm going to close this then, but leave #3160 open.

PR #3140, opened 2019-10-25
Release 1.149.0

clouserw:


AUTHORS
```diff
@@ -26,12 +26,12 @@ ckarlof
 Dan Callahan
 Danny Amey
 Danny Coates
-Dave Justice
```
vladikoff: I think this AUTHORS reorder happens because someone has an older version of git on their machine,

@clouserw and @vbudhram could you sync up on Git Versions? :)

PR #3143, opened 2019-10-27
feat(auth-server): add typescript checking capability

bbangert:
Because:

* Javascript is prone to a variety of type errors.

This commit:

* Allows TypeScript checking of Javascript.

### DIFF
--- a/packages/fxa-auth-server/test/client/index.js
+++ b/packages/fxa-auth-server/test/client/index.js
@@ -253,7 +253,7 @@ module.exports = config => {
             c.authAt = data.authAt;
             c.verified = data.verified;
             c.verificationReason = data.verificationReason;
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/test/client/index.js
```diff
@@ -253,7 +253,7 @@ module.exports = config => {
             c.authAt = data.authAt;
             c.verified = data.verified;
             c.verificationReason = data.verificationReason;
-            c.verificationMetho = data.verificationMethod;
```
bbangert: The type checker found this amongst lots of probably erroneous bits. This seemed pretty obviously a mistake.

vbudhram: Oof, guess we weren't really checking this.

PR #3154, opened 2019-10-28
fix(browserid): Let clients specify service=${client_id} in /certificate/sign

rfk:
Due to some issues with processing the high volume of OAuth token-creation
events, we may need to have desktop Firefox use BrowserID assertions to
create its OAuth tokens, meaning it may want to call /certificate/sign even
if the `sync` service is not enabled.

This commit allows values other than `sync` in the `service` query param
for /certificate/sign, so that we won't get bogus metrics from browsers
that are calling this endpoint but are not connected to sync. For more
context see https://bugzilla.mozilla.org/show_bug.cgi?id=1591312.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/sign.js
+++ b/packages/fxa-auth-server/lib/routes/sign.js
@@ -22,7 +22,7 @@ module.exports = (log, signer, db, domain, devices) => {
         },
         validate: {
           query: {
-            service: validators.service.valid('sync').optional(),
+            service: validators.service.optional(),
           },
           payload: {
             publicKey: isA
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/sign.js
```diff
@@ -54,8 +54,9 @@ module.exports = (log, signer, db, domain, devices) => {
         const sessionToken = request.auth.credentials;
         const publicKey = request.payload.publicKey;
         const duration = request.payload.duration;
-        // This is a legacy endpoint that's only used by clients connected to sync,
-        // so always set `service=sync` for metrics logging purposes.
+        // This is a legacy endpoint that's typically only used by
+        // clients connected to sync, so assume `service=sync` for
+        // metrics logging purposes unless we're told otherwise.
         if (!request.query.service) {
```
vbudhram: I think you might be able to do `default('sync')` in joi and remove the if block here.

vbudhram: Granted it is easy to see what is happening with the comment and the if block so your call.

rfk: Thanks, yeah, I mostly just left it as-is to keep the change as small as possible.

rfk:
@vbudhram I won't merge this to avoid stepping on your toes for the point release, but I think it's ready.

PR #3155, opened 2019-10-28
Train 149 cherry pick master

vbudhram:
From today's meeting these are the commits that we wanted to get into the point release. 

Ref: https://docs.google.com/document/d/1dIkxIyp-6mieJ3oYG7f-jHX_1Vj_9fAmnyXxuKxGOMg/edit#

Cirlce is having a bad day and not running tests, so I'll hold off on tagging until we get ✅


chenba:
Unfortunately we just found a bug with 07b01d9.  It doesn't hurt anything if we keep it in this release, but it won't fix #3053. 🙁

vbudhram:
@chenba No worries thanks for heads up, I'll still go ahead and tag this since there a couple commits to get in.

PR #3161, opened 2019-10-29
chore(import): added fxa-dev-launcher to monorepo

dannycoates:


dannycoates:
> should we also add to https://github.com/mozilla/fxa/blob/master/_scripts/install_packages.sh#L20?

as of #3156 it is implicitly

PR #3167, opened 2019-10-29
fix(customs): Fix ip_blocklist .netstat file parse error.

### LINKED ISSUE #3158: (customs): The customs-server in v1.149.0 does not start.
The customs-server in v1.149.0 does not start. It logs only the below lines on the way out (last okay line, then `"op":"customs.bin.error","err":{"cause":{}` thank you very much, and docker recording that the container has exited.
`
Oct 28 23:26:01 ip-172-31-81-183.ec2.internal docker-fxa-customs[14825]: {"name":"customs-server","hostname":"ip-172-31-81-183.ec2.internal","pid":1,"level":30,"op":"requestChecks.validate.changed","key":"flowIdRequiredOnLogin","currentFlowIdRequiredOnLogin":true,"futureFlowIdRequiredOnLogin":false,"msg":"","time":"2019-10-28T23:26:01.378Z","v":0}
Oct 28 23:26:01 ip-172-31-81-183.ec2.internal docker-fxa-customs[14825]: {"name":"customs-server","hostname":"ip-172-31-81-183.ec2.internal","pid":1,"level":50,"op":"customs.bin.error","err":{"cause":{},"isOperational":true},"msg":"","time":"2019-10-28T23:26:01.383Z","v":0}
Oct 28 23:26:01 ip-172-31-81-183.ec2.internal systemd[1]: docker-fxa-customs.service: main process exited, code=exited, status=1/FAILURE
`
This in turn prevents the fxa-auth-server from working.

The only "change" since v1.148.8 is an upgrade of `csv-parse` from "1.0.4" to "4.4.6" (which it uses to read IP block lists from disk).

In a remarkable coincidence, if I do `npm test` locally, it fails a number of csv tests. 

But under circle-ci it doesn't appear to run any tests, presumably since no code change (except code did change with the dependency upgrade of course).

shane-tomlinson:
We updated csv-parse from 1.0.4 to 4.4.6, and the major
version change included changes to how csv-parse treats
lines that start with a `#`, no longer are they considered
comments by default.

Instead of using csv-parse, this ditches the library
all together because we only used the first column of
each row anyways. Instead of relying on csv-parse,
just trim the line, then filter out empty or comment
lines.

fixes #3158

Note, locally, I still have a ton of remote test failures, just not in the local/ip_blocklist_tests.js module.

I imagine we'll have to uplift this to 149. Or I can target 149 and then merge to master.

@jrgm, @mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-customs-server/lib/ip_blocklist.js
+++ b/packages/fxa-customs-server/lib/ip_blocklist.js
@@ -13,7 +13,6 @@
 var Promise = require('bluebird');
 var readFile = Promise.promisify(require('fs').readFile);
 var statFile = Promise.promisify(require('fs').stat);
-var parse = Promise.promisify(require('csv-parse'));
 var path = require('path');
 var ip = require('ip');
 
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
> Note, locally, I still have a ton of remote test failures, just not in the local/ip_blocklist_tests.js module.

This is because I had the customs server running when I typed `npm run test`. All good now.

PR #3181, opened 2019-10-29
feat(auth-server): retain backendService errors

### LINKED ISSUE #3138: [fxa-auth-server] Chain errors to avoid losing stack information
In fxa-auth-server (and possibly other packages), we catch and rethrow errors, typically as AppError for use in the request cycle control-flow. These are used to exit with specific request information such as a status code and message set, but when reported by Sentry have lost the original stack trace.

Using a package like [verror](https://www.npmjs.com/package/verror), we could re-plumb these error re-throws to include the `cause` each time an error is thrown and retain the full error stack. This would help us know the call stack that resulted in the error and more easily annotate it with useful debug information.

bbangert:
Because:

* backendService.js dropped the original error when throwing an
  AppError.

This commit:

* Retains the original error when throwing backend service or
  validation errors in backendService.js.

Closes #3138

### DIFF
--- a/packages/fxa-auth-server/lib/backendService.js
+++ b/packages/fxa-auth-server/lib/backendService.js
@@ -147,7 +147,11 @@ module.exports = function createBackendServiceAPI(
             value,
           });
           reject(
-            error.internalValidationError(fullMethodName, { location, value })
+            error.internalValidationError(
+              fullMethodName,
+              { location, value },
+              err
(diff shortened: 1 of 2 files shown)


bbangert:
Refactoring the rest of the file broke the support form page for reasons I don't understand so I backed that part out and restricted this to the most important bit.

Issue #3170, opened 2019-10-29
Move content-server experiment base logic into fxa-shared


vbudhram:
While working on https://github.com/mozilla/fxa/issues/2595, I didn't want to duplicate the experiment grouping logic found in the [content-server](https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/base.js#L35) to the auth-server.

I think it makes sense to pull this base class into fxa-shared and reused in both services. However, I still think the experiment rules should still live in the respective services.

Issue #2595, opened 2019-09-24
(Phase 1) Recovery Keys and Secondary Email During Registration


davismtl:
We’ve always had a steady volume of users complaining or reaching out to us via Bugzilla, SUMO, Twitter and Reddit due to being locked out of their accounts.

There are three reasons for this:
- They don’t have access to their email anymore;
- They forgot their password;
- They lost their device for TOTP.

This feature proposes to tackle the first two. The impact of the first reason is that users can’t get into their accounts. The impact of the second reason is that users risk losing their Sync data if they do not have more than 1 device actively Syncing.

As a first phase, we will send users an email immediately after registration to encourage them to setup a recovery key and a secondary email. This email will link to the settings. As a phase 2, we will link people to a screen that can be used from various places like our login flow and account toolbar menu in the browser.

Feature Doc: https://docs.google.com/document/d/1Wz0itS76pNS24zeNS5MEqqjQdwxHuNyfq3PTj8C688s/edit#

┆Issue is synchronized with this [Jira Epic](https://jira.mozilla.com/browse/FXA-370)
┆Issue Number: FXA-370

davismtl:
The feature doc is now ready for design.

davismtl:
Ryan added proposed email templated to the feature. Jeff will help with copy.

We've realized that there may be two ways of doing this:
- 1 email with 2 calls to action
- 2 emails with 1 call to action

Let's see what amount of scope that would add?

davismtl:
from mtg:
Work has started. Email templates have been created. Should land in 150.
Updating experiment logic so that auth server can use it.

davismtl:
R+
It's a point release for train 151.
The experiment is setup and there will be 3 treatments with different email variations.

davismtl:
Just working on enabling it and measuring it.

Config change for it to go out. No deployment needed.

irrationalagent:
Rolled out on 2019-12-16 at about 7:50 PM PDT

[Started a dashboard](https://analytics.amplitude.com/mozilla-corp/dashboard/wwsbahr)

data-sync-user:
➤ Alex Davis commented:

Need to log a conclusion. [~adavis@mozilla.com]

PR #3183, opened 2019-10-30
feat(experiments): Add experiments support in fxa-shared

vbudhram:
Connect with https://github.com/mozilla/fxa/issues/3170

Unfortunately, I was not able to 100% move the [base experiment](https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/base.js#L7) in content-server over to our fxa-shared project. Please see the linked issue for the different things I tried. 

To spare my sanity I opted to update the experiment file to typecript and pull over the same tests from content-server.

This should be enough for me to build out the experiment parts needed in the auth-server and finish https://github.com/mozilla/fxa/issues/2595.


packages/fxa-shared/experiments/base.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Return a 32 bit hash of `key`
+   *
+   * @param {String} key
+   * @returns {Number}
+   */
+  public hash(key: string): number {
+    // md5 returns 32 hex bytes, we want 32 bits. 32bits = 8 hex bytes.
+    const hash = md5(`${this.name || ''}:${key}`).substr(0, 8);
```
chenba: nit: since the `name` member has a default of empty string, `|| ''` isn't necessary here.

packages/fxa-shared/test/experiments/base.js
```diff
@@ -0,0 +1,258 @@
...
+    it('must be overridden', () => {
+      assert.throws(function() {
+        experiment.choose();
+      }, 'choose must be overridden');
+    });
+
+    it('throws if deprecated', () => {
+      assert.throws(() => {
+        experiment.name = 'oldExperiment';
+        experiment.deprecated = true;
```
chenba: `name` and `deprecated` are private members.  Can they be assigned this way?

packages/fxa-shared/test/experiments/base.js
```diff
@@ -0,0 +1,258 @@
...
+      }
+
+      choose(subject) {
+        if (subject.experimentChooser.choose(subject) !== this.name) {
+          return false;
+        }
+
+        const GROUPS = ['control', 'treatment'];
+        return this.uniformChoice(GROUPS, subject.uuid);
+      }
```
chenba: nit: from looking at this impl, `choose` either returns false or a string?  Could add that the method declaration in the base class.

packages/fxa-shared/test/experiments/base.js
vbudhram: I've updated them to be public. Typescript didn't complain but don't think this will hurt?

packages/fxa-shared/test/experiments/base.js
vbudhram: TIL that typescript has [union types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types). Allows function to return boolean or string.

vbudhram:
@chenba Thanks for taking a look at this, I've made the updates your recommended.

PR #3191, opened 2019-10-30
fix(tests): baaaaAAAAA 🤯

dannycoates:
running auth-server tests locally broked

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -154,7 +154,7 @@ function run(config) {
             }
           }
 
-          init();
+          return init();
         });
       },
       err => {
--- a/packages/fxa-auth-server/lib/routes/utils/oauth.js
+++ b/packages/fxa-auth-server/lib/routes/utils/oauth.js
@@ -4,7 +4,7 @@
 
 'use strict';
 
-const encrypt = require('../../../fxa-oauth-server/lib/encrypt');
+const encrypt = require('../../../lib/oauth/encrypt');
 const { OAUTH_SCOPE_OLD_SYNC } = require('../../constants');
 const ScopeSet = require('../../../../fxa-shared').oauth.scopes;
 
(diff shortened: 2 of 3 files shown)


bbangert:
It would be nice if this indicated when it got broken and why, reference the other commit?

dannycoates:
> It would be nice if this indicated when it got broken and why, reference the other commit?

go for it ;) ... I think it was more than one thing

dannycoates:
Yes, I was irritated when I wrote this. I found that the auth-server tests were broken doing some other work on a different branch and found it reproducible on master so I went about fixing it. This isn't a broken feature so I didn't particularly care about spending more time tracking down the cause, but the `test-local.sh` is probably from #3156 and I have no idea where the `key_server.js` one is from. I don't think spending more time on this will enrich anyone's lives.

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -154,7 +154,7 @@ function run(config) {
             }
           }
 
-          init();
```
bbangert: Sorry, I'll rephrase. Git blame shows the last change to this line from 4 months ago during prettier. If this line must include `return`, how did we miss that for 4 months? What wasn't working correctly? Is there a test we can include that will prevent repeat breakage?

bbangert: Ok, I located where this came from, which I'm sure @vladikoff can appreciate 😉 :
https://github.com/mozilla/fxa/commit/98ad6719b643637c1a60add100e718393178bfbf#diff-1c5b95cf27fd1faec187d6374b32bdabL123

If the `return` is necessary, I'm a bit flummoxed how we didn't notice that for over a year.

dannycoates: It's possible this wasn't an issue before because there wasn't anything pressuring the race condition. Now the tests initialize faster because they're doing less by not starting up an oauth-server. The error originated here https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/test/lib/server.js#L63 because `server` was `null`. It is weird that I only saw this today since it should have shown up in earlier oauth->auth phase 2 changes. Maybe my machine likes the cooler weather.

bbangert: This is a great 'why' to include in the commit!

PR #3195, opened 2019-10-30
fix(cwts): no longer present a CWTS option for sign-in flows

### LINKED ISSUE #3083: CWTS incorrectly shown during sync-optional signin flow
Running latest nightly, I have the following:

#### STRs

1) Visit about:welcome and enter the email address of an existing Firefox Account to "Join Firefox".
2) Click "continue" and observe that you're taken to the signin page:

3) Sign in, and observe that you're asked whether to sync this device to your account:

4) Click "Sync this device".

#### Expected result

Per the [UX flowchart spec](https://www.lucidchart.com/documents/edit/6aec3e89-aa60-4aef-ba55-68d5ac918fbd/socX9~xb8QR7?beaconFlowId=8506F5E2F3EC274F), step (4) should be the final step in the flow - the device should be signed in and start syncing and the user should see the "connect another device" screen.

#### Actual result

The user is asked to "choose what to sync", with all checkboxes checked:

This is inconsistent with signing in via other sync flows, where existing accounts do not see the CWTS screen.

Short-term, we should not show the CWTS screen here. It's inconsistent and it doesn't respect any existing data choices the user may have made.

Longer-term, we need to figure out how to show CWTS only to users who have not previously made any data choices.

vladikoff:
Fixes #3083

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
@@ -111,6 +111,9 @@ export default {
           // force_auth attempts do not a choice for Sync
           return this.navigate('would_you_like_to_sync', {
             account: account,
+            // ask the user if they want to sync, but don't ask to specify data choices via CWTS
+            // see https://github.com/mozilla/fxa/issues/3083 for details
+            skipCWTS: true,
             // propagate the onSubmitComplete to choose_what_to_sync screen if needed
             onSubmitComplete: this.onSignInSuccess.bind(this),
           });
--- a/packages/fxa-content-server/app/scripts/views/would_you_like_to_sync.js
+++ b/packages/fxa-content-server/app/scripts/views/would_you_like_to_sync.js
@@ -23,6 +23,7 @@ const View = FormView.extend(
       // to keep the view from knowing too much about the state machine,
       // a continuation function is passed in that should be called
       // when submit has completed.
+      this.skipCWTS = this.model.get('skipCWTS');
       this.onSubmitComplete = this.model.get('onSubmitComplete');
     },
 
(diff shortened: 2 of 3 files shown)


packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
```diff
@@ -111,6 +111,9 @@ export default {
           // force_auth attempts do not a choice for Sync
           return this.navigate('would_you_like_to_sync', {
             account: account,
+            // ask the user if they want to sync, but don't ask to specify data choices via CWTS
+            // see https://github.com/mozilla/fxa/issues/3083 for details
+            skipCWTS: true,
```
rfk: Under what circumstances would we ever set this to `false`?

packages/fxa-content-server/app/scripts/views/would_you_like_to_sync.js
```diff
@@ -40,6 +41,11 @@ const View = FormView.extend(
...
       return Promise.resolve().then(() => {
         const account = this.getAccount();
+        if (this.skipCWTS) {
+          // don't ask to specify data choices via CWTS
+          // see https://github.com/mozilla/fxa/issues/3083 for details
+          return this.onSubmitComplete(account);
+        }
 
         // we replace the current view to avoid various problems with the back button
         return this.replaceCurrentPage('choose_what_to_sync', {
```
rfk: In other words, I don't think we ever want to go from the WYLTS screen to the CWTS screen, so I don't understand why we need a branch here that forwards to CWTS.

vladikoff: good point! updated!

PR #3209, opened 2019-10-31
feat(email-first): Email-first everywhere!

### LINKED ISSUE #3204: Use email-first as default for "web" broker
The "web" broker is for users that sign into FxA when browsing directly to https://accounts.firefox.com, and is used a lot in our functional tests.

blocks #639

shane-tomlinson:
fixes #3204

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sign_in_unblock.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_in_unblock.js
@@ -52,16 +52,17 @@ const View = FormView.extend({
     );
   },
 
-  onSignInError(account, password, err) {
-    if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+  onSignInError(account, password, error) {
+    if (AuthErrors.is(error, 'INCORRECT_PASSWORD')) {
       // The user must go enter the correct password this time.
       this.navigate(this._getAuthPage(), {
+        account: this.getAccount(),
         email: account.get('email'),
-        error: err,
+        error,
       });
     } else {
       // re-throw, it'll be displayed at a lower level.
-      throw err;
+      throw error;
     }
   },
 
--- a/packages/fxa-content-server/tests/functional/change_password.js
+++ b/packages/fxa-content-server/tests/functional/change_password.js
@@ -10,7 +10,7 @@ const FunctionalHelpers = require('./lib/helpers');
 const selectors = require('./lib/selectors');
 
 const config = intern._config;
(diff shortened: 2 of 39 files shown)


packages/fxa-content-server/tests/functional.js
```diff
@@ -22,7 +22,6 @@ module.exports = [
   'tests/functional/cookies_disabled.js',
   'tests/functional/delete_account.js',
   'tests/functional/email_opt_in.js',
-  'tests/functional/email_service.js',
```
vladikoff: @shane-tomlinson did you want to delete this? this is unrelated to email_first ?

shane-tomlinson: It is unrelated, but we are using the email-service everywhere now so it seems like a bit of a superfluous test.

packages/fxa-content-server/app/scripts/views/sign_in_unblock.js
```diff
@@ -52,16 +52,17 @@ const View = FormView.extend({
     );
   },
 
-  onSignInError(account, password, err) {
-    if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+  onSignInError(account, password, error) {
+    if (AuthErrors.is(error, 'INCORRECT_PASSWORD')) {
       // The user must go enter the correct password this time.
       this.navigate(this._getAuthPage(), {
+        account: this.getAccount(),
```
shane-tomlinson: This was needed to get signin unblock to go to the /signin page where the user could enter their password rather than to `/`

packages/fxa-content-server/tests/functional.js
shane-tomlinson: Brought the test back.

shane-tomlinson:
I'm calling this ready for review, @mozilla/fxa-devs - r?

The main change here is to make the "web" broker (the fallback when user browses directly to the site) use email-first. That caused a *ton* of test upheaval.

packages/fxa-content-server/tests/functional/change_password.js
```diff
@@ -57,9 +57,6 @@ registerSuite('change_password', {
     email = TestHelpers.createEmail();
   },
 
-  afterEach: function() {
```
vbudhram: It isn't obvious to me why this was removed?

shane-tomlinson: Because we do it in the beforeEach, and I think everywhere should do it in beforeEach. Doing it in afterEach and then beforeEach in the subsequent tests just slows things down.

PR #3215, opened 2019-10-31
fix(payments): correctly calculate and append time to amplitude events

### LINKED ISSUE #3050: Payments metrics missing `time` key

6a68:
The bug was looking for event.time, not data.time, in pruneUnsetValues,
but I added and cleaned up some tests and mocks along the way, which
seems worth keeping.

Fixes #3050.

(Looks like I broke something in fxa-content-server with the fix to fxa-payment-server, I'll look at that in a bit)


6a68:
eh...gonna try something simpler

PR #3217, opened 2019-10-31
fix(payments): correctly append event time to amplitude events

### LINKED ISSUE #3050: Payments metrics missing `time` key

6a68:
Deferring any attempt at refactoring until the payments metrics are
actually working.

Fixes #3050.

### DIFF
--- a/packages/fxa-payments-server/server/lib/amplitude.js
+++ b/packages/fxa-payments-server/server/lib/amplitude.js
@@ -11,7 +11,6 @@ const {
   mapFormFactor,
   mapLocation,
   mapOs,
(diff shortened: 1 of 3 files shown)


chenba:
> 
> 
> LGTM. what was broken about `...mapTime`?

The first arg passed into it didn't have `offset`.

6a68:
The issue with `mapTime` was that the auth and content routes actually do the event.time calcs in the route handler, before the call to `amplitude`, so when I got `mapTime` properly working, I discovered I'd broken the auth and content server tests. Seemed like the simplest fix would be a better first step.

Issue #3151, opened 2019-10-28
Change card_declined error message to use card_error string


johngruen:
This is the generic declined message and it shows when a user mis-enters their card expiry. We should use the `card_error` message which is more descriptive.

Related to #3106

lmorchard:
I'm having trouble reproducing this one... can we get a screenshot of the exact error that's happening, and what error should appear?

LZoog:
I _think_ I see what you're wanting here, but agreed we need a screenshot of some sort - when handling Stripe `card_error` errors at the `createToken` level we're displaying the message implemented in [this PR](3136). We have a `createSubscription` piece that does show a "Card declined" header with the message received from Stripe (admittedly I'm not entirely sure when we hit this instead of `createToken` errors without digging further) - is this what you're wanting us to change? We have "Subscription failed" if it fails at this level but is _not_ a `card_declined` error. I'm not sure what the history is around handling that the way we are there.

See [the code here](https://github.com/mozilla/fxa/blob/5546fb97b310c4d8dee8e166c0cd5a5ac9613bce/packages/fxa-payments-server/src/routes/Product/index.tsx#L272). This might be what you're referencing?

LZoog:
@johngruen I know we chatted really briefly about this on Slack today - after looking into this a bit more I'd still like you to post a screenshot if possible.

@lmorchard, correct me if I'm wrong, but it looks like we hit a `createSubscription` error if the Stripe payment goes through but the subscription fails to submit on our end (with the plan name and such), but it doesn't make sense for that to be applicable here. If Stripe rejects the payment, which it should if the expiration is incorrect, it should be caught in `createTokenError.error`, in which case, this issue should already have been taken care of.

BTW John, are you using your own card on prod? I can't find a Stripe test card that works locally or on stage to reject for this reason... 🤔

johngruen:
I am using my own card on Prod, yes...here's a screenshot of the error. Not sure if this is helpful...my concern is that this error is being displayed when i mis-enter my card's expiration date.

<img width="740" alt="Screen Shot 2019-10-28 at 7 23 31 PM" src="https://user-images.githubusercontent.com/3323249/68040520-c5974b00-fcce-11e9-80c9-42138c6f70f4.png">

lmorchard:
Taking a quick look at [where this dialog message is coming from](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/routes/Product/index.tsx#L27).

[...]

PR #3230, opened 2019-11-04
chore(payments/copy): Update subscription failed error message to display generic error

LZoog:
fix for #3151

@johngruen please confirm we also want to update the header to be more generic.


LZoog:
John approved the header message for now, merging.

PR #3231, opened 2019-11-04
refactor(build): fixed default `npm install` run

dannycoates:


_scripts/install_packages.sh
```diff
@@ -6,9 +6,9 @@ ulimit -S -n 2048 || echo "Setting ulimit failed"
 
 if [ "${CI}" = "true" ]; then
   # it seems the filesystem on circleci can't handle full concurrency
-  npx lerna exec --concurrency 2 -- npm ci
+  npx lerna exec --concurrency 6 -- npm ci
```
dannycoates: 2 is so slow... 8 was too high. maybe 6?

_scripts/install_packages.sh
```diff
@@ -6,9 +6,9 @@ ulimit -S -n 2048 || echo "Setting ulimit failed"
...
 else
-  npm run npm-ci-all
+  npx lerna bootstrap --ignore fxa-amplitude-send
```
dannycoates: `node-parquet` from fxa-amplitude-send adds a whole host of native dependencies that should either get added to `check.sh` or refactored away by using a different parquet module. since f-a-s isn't a core dev module I'm ignoring it here in the meantime to keep `npm install` working for local dev.

dannycoates:
`node-parquet` has slowed down the CI build by several minutes. Maybe we should exclude `fxa-amplitude-send` from there too.

Issue #3090, opened 2019-10-23
Support Submit Error / Incorrect state reported to user


johngruen:
From Andrew Morales...reported in production

User sees an error, but also receives an email indicating that support request is received.

> When I went to submit my Support Ticket I received an error letting me know that my request was not successful. I tried twice. Got the same error twice. When I closed my window and went to my GMail inbox I see that my support tickets were created even though I was given an error message. 

<img width="704" alt="180-9ffb08b9570e970af7b6291d06f42192_Screen_Shot_2019-10-22_at_11 00 24_AM" src="https://user-images.githubusercontent.com/3323249/67388076-4e5c0b80-f598-11e9-994c-1e29abeb2676.png">

johngruen:
Ben to look in logs

bbangert:
I should note, I can see why this is happening. Due to how Zendesk works, we must first create a ticket, which creates the user record if needed. Then we have to fetch and update the user to add the fxa user id to the Zendesk user record. Since the support request was created, the first call to make the ticket did work, but one of the followup calls to add the fxa user id must have failed.

To attempt to remedy this, I will add a retry to the request, and then capture the error to Sentry (it's currently not reported at all).

clouserw:
We discussed this in our general meeting this morning and confirmed that this is **not** a blocker.

bbangert:
Retries have been added, and more error reporting to track if it happens again.

PR #3247, opened 2019-11-05
feat(auth-server): add zendesk API retries

bbangert:
Because:

* Zendesk calls could have intermittent network failures.
* Failing to update a created Zendesk user to include the FxA uid
  makes it harder to link to the FxA user.

This commit:

* Retries Zendesk calls to reduce errors from network errors.

Issue #3090

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/support.js
+++ b/packages/fxa-auth-server/test/local/routes/support.js
@@ -234,6 +234,36 @@ describe('support', () => {
         nock.isDone();
       });
 
+      it('should handle retrying an update user call', async () => {
+        config.subscriptions.enabled = true;
+        nock(`https://${SUBDOMAIN}.zendesk.com`)
+          .post('/api/v2/requests.json')
+          .reply(201, MOCK_CREATE_REPLY);
+        nock(`https://${SUBDOMAIN}.zendesk.com`)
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/test/local/routes/support.js
```diff
@@ -234,6 +234,36 @@ describe('support', () => {
...
+          .post('/api/v2/requests.json')
+          .reply(201, MOCK_CREATE_REPLY);
+        nock(`https://${SUBDOMAIN}.zendesk.com`)
+          .get(`/api/v2/users/${REQUESTER_ID}.json`)
+          .reply(500)
+          .get(`/api/v2/users/${REQUESTER_ID}.json`)
+          .reply(200, MOCK_NEW_SHOW_REPLY);
+        nock(`https://${SUBDOMAIN}.zendesk.com`)
+          .put(`/api/v2/users/${REQUESTER_ID}.json`)
+          .reply(200, MOCK_UPDATE_REPLY);
```
chenba: We need a failed PUT before this in order to test the retry at https://github.com/mozilla/fxa/pull/3247/files#diff-c01ed3e5dd5108a3df8ed8fc6bed2383R110-R114

bbangert: Yes, I only added one failure to test, mainly as the blocks were otherwise identical.

PR #3257, opened 2019-11-06
feat(signup-code): CWTS on the signup password page experiment

### LINKED ISSUE #3080: Should Confirming Email with Codes Come Before CWTS?
If you drop out of the funnel on the CWTS page, you cannot confirm your email. ~Roughly 15% of user drop off at CWTS~ so they will not be able to confirm their emails.

Edit: when adding a chart, I noticed that I could not replicate this value in charts.

Edit 2: Feature doc link:
https://docs.google.com/document/d/1vl9qtckQvepWvS_l1I82etozrUq27q5Fmn7EHjbbeEw/edit#

┆Issue is synchronized with this [Jira Epic](https://jira.mozilla.com/browse/FXA-478)
┆Issue Number: FXA-478

shane-tomlinson:
Some 7-8% of users never submit CWTS and at this point
a verification code would have been sent to them. Without
somewhere to enter the verification code, we sometimes
lose users.

This moves CWTS onto the signup password page and only
sends the email once the user has submit the form,
while they are looking at the "enter your code" page.

feature doc:

https://docs.google.com/document/d/1vl9qtckQvepWvS_l1I82etozrUq27q5Fmn7EHjbbeEw

fixes #3080


shane-tomlinson:
This is currently running on https://stomlinson.dev.lcip.org.

Using [fxa-dev-launcher](https://github.com/vladikoff/fxa-dev-launcher) open it like this:

```
$fxa-dev-launcher> FXA_ENV=stomlinson npm start
```

Then sign up via any of your favorite touchpoints with a @mozilla.com address.

@vladikoff, @vbudhram - Could I get your feedback on this? I'm not sure whether to add loads of unit tests for this or not, or whether to add that as a v2 if we decide to keep the experiment. I can see adding unit tests to the new experiment grouping rule. I did add functional tests. Most of the code in the new mixin is pulled directly from choose_what_to_sync.js.

shane-tomlinson:
cc @davismtl, @ryanfeeley

shane-tomlinson:
Added unit tests, marking this ready for review!

@vladikoff, @vbudhram, @ryanfeeley, and @mozilla/fxa-devs - r?

shane-tomlinson:
I deselected a bunch of sync engines, and you can see that both the experiment name and the remaining selected sync engines are logged in both the fxa_reg - submit fxa_reg - signup_code_submit events:

Screenshot 2019-11-07 at 13 06 04


Screenshot 2019-11-07 at 13 04 05

vbudhram:
Since this is a fairly big change, we should probably update deploy docs and have QA test the flows explicitly.

vasilicatamas:
Listed the testing results in #3080.

PR #3265, opened 2019-11-07
fix(payments): include time fields in amplitude events

### LINKED ISSUE #3050: Payments metrics missing `time` key

6a68:
Note: opening as a draft because this still needs tests, but I want to make sure the overall approach seems correct.

Commit message continues:
-----

Specifically, include the `perfStartTime`, `startTime`, and `flushTime`
in each amplitude event emitted by the front-end.

More details:

The content-server amplitude events are also flow events, meaning
the `perfStartTime` and other times needed to reconstruct the event time
on the server are already available on the event object. The arithmetic
used to calculate `event.time` was copied over to the payments server
post-metrics route handler code without verifying that the needed data
was available on the amplitude events emitted by the front-end.

Unlike content server, the payments server amplitude events are
generated by a redux middleware that observes certain actions dispatched
while the user interacts with the payments pages, which is separate from
how payments flow events are generated and processed. This means that the
necessary time values need to be explicitly added to each amplitude event
before the event is sent down to the server.

I verified this patch fixed the issue by attaching the node debugger to
a local payments server process, but you could also look in the payments
server logs for an amplitude event logged with a `time` property. One
such event (the engage event) is emitted as soon as a character is typed
into the subscription form.

Hopefully, this really fixes #3050.

### DIFF
--- a/packages/fxa-payments-server/src/store/amplitude-middleware.ts
+++ b/packages/fxa-payments-server/src/store/amplitude-middleware.ts
@@ -1,6 +1,7 @@
 import { AnyAction, Dispatch, Middleware, MiddlewareAPI } from 'redux';
 import { logAmplitudeEvent } from '../lib/flow-event';
 import * as Sentry from '@sentry/browser';
+import { config } from '../lib/config';
 
 const eventGroupNames = {
   createSubscription: 'subPaySetup',
@@ -41,6 +42,7 @@ export const AmplitudeMiddleware: Middleware = (
         logAmplitudeEvent(
           eventGroupNames.manageSubscriptions,
           eventTypeNames.view,
+          config.perfStartTime,
           {}
         );
         break;
@@ -48,6 +50,7 @@ export const AmplitudeMiddleware: Middleware = (
         logAmplitudeEvent(
           eventGroupNames.manageSubscriptions,
           eventTypeNames.engage,
+          config.perfStartTime,
           {}
         );
         break;
@@ -56,6 +59,7 @@ export const AmplitudeMiddleware: Middleware = (
         logAmplitudeEvent(
           eventGroupNames.createSubscription,
           eventTypeNames.view,
+          config.perfStartTime,
           getPlanPropsFromAction(action)
         );
         break;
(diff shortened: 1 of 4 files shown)


packages/fxa-payments-server/src/store/amplitude-middleware.ts
```diff
@@ -100,79 +109,95 @@ export const AmplitudeMiddleware: Middleware = (
...
-          ...getFailureReasonFromAction(action),
-        });
+        logAmplitudeEvent(
+          eventGroupNames.updatePayment,
+          eventTypeNames.fail,
+          config.perfStartTime,
+          {
+            ...getPlanPropsFromAction(action),
+            ...getFailureReasonFromAction(action),
+          }
```
6a68: idk why prettier does this formatting sometimes but not always 🤷‍♂️

chenba:
That make sense.  Thanks @6a68 for digging into it.

Looking at this, it seems kinda silly that we re-calculate the time of the event given that we have the value at the time of the event [0]!  We send an `offset` because that's what content server did.  But what would break if we send the event time in addition or instead?  (I'm mostly mumbling to myself here... we don't need to change the patch or anything.)

[0] which was what I think the original patch was expecting?

6a68:
Really good point about the use of `offset` 🤔 ...I hope we can make time to simplify all this metrics code soon

chenba:
```
fxa-payments-server.INFO: amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_subscribe_cancel - complete","time":1573160831540,"device_id":"2bd2fc04efa340e8955cdf9cb30a7b88","session_id":"1573160811376","app_version":"149.4","os_name":"Windows","os_version":"10","event_properties":{"plan_id":"123doneProMonthly","product_id":"123doneProProduct"},"user_properties":{"flow_id":"2cdacfc6d57f7092e23f87fcdbf286045f96eb446732294d289560d72bb65a39","ua_browser":"Firefox","ua_version":"72.0"}}
```
Verified locally.

Issue #3135, opened 2019-10-25
Card validation errors should use existing fxa pattern, not modal


johngruen:
Right now, card validation errors show up in a modal. This is unhelpful as it obscures the CC form and makes it hard for users to spot the error in their input while seeing the instructional text of the error. 

We should ditch the error modal and use the `.error` pattern utilized by other parts of FxA.

This pattern should apply in subscription settings as well

![image](https://user-images.githubusercontent.com/3323249/67584907-e865b480-f74e-11e9-8628-f3cb6267ed74.png)

johngruen:
This implies that we ditch headers for error messages as well

LZoog:
Patch: #3266

PR #3266, opened 2019-11-07
feat(payments): Create/animate 'ErrorMessage' component to display card-specific errors instead of a modal

### LINKED ISSUE #3240: Incorrectly Entered Expiration date returns CVC error
### STR

* Input a correct CC, CVC, and Exipration
* Misenter your card's expiration date. You can do this by entering a valid future month, year combination that is not the same as your card's.

### Expected
* Stripe returns`invalid_expiry_month` or `invalid_expiry_year`

### Actual
* Stripe returns the following
```
{
  "error": {
    "code": "incorrect_cvc",
    "doc_url": "https://stripe.com/docs/error-codes/incorrect-cvc",
    "message": "Your card's security code is incorrect.",
    "param": "cvc",
    "type": "card_error"
  }
}
```

LZoog:
fix for #3135 - confirmed by John that this should, for now, only cover `createToken.error`, `updatePaymentStatus.error`, and `createSubscriptionStatus.error.code === 'card_declined'`

bonus, also fixes #3240 

The animation looks a little choppier in this gif than it does on the screen. I tried to get the animation close to what we're doing for FxA errors but we're using jQuery's `slideDown` / `slideUp` there and it'd be nice to get us away from jQ. @johngruen or @ryanfeeley we can fine-tooth comb this animation in a follow up if you think it's necessary.

We may also want a new issue to go over when error messaging should break onto new lines like in the mockup John provided in #3135.

error


packages/fxa-payments-server/src/components/ErrorMessage.scss
```diff
@@ -0,0 +1,26 @@
...
+    text-align: center;
+  }
+
+  + form {
+    margin-top: 12px;
+  }
+}
+
+.animate-fade-enter {
+  animation: fade-down-tt $short-transition;
```
LZoog: When I go through our CSS this quarter/in Q1, one thing I'll do is make our animation names a little more generic. We shouldn't need `-tt` here for "tooltip" for example, as now that I'm using it here, it feels like an awkward name.

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -140,6 +142,12 @@ export const Product = ({
     [setCreateTokenError]
   );
 
+  // clear any error rendered with `ErrorMessage` on form change
+  const onChangeErrorDismiss = useCallback(() => {
+    setCreateTokenError({ type: '', error: false });
+    resetCreateSubscriptionError();
+  }, [setCreateTokenError, resetCreateSubscriptionError]);
```
LZoog: I was thinking I could add a check here for each error message (if createTokenError, setCreateTokenError) but that didn't seem quite right, and now that I'm looking at this again, maybe it's better to pass in a new boolean prop like `displayErrorMessage` into `PaymentForm` that when true, clears all the possible errors. Otherwise this gets triggered on every `onChange` event in the form.

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -249,10 +234,48 @@ export const Product = ({
       <h3 className="billing-title">
         <span>Billing Information</span>
       </h3>
+
+      <TransitionGroup component={null}>
+        {createTokenError.error && (
+          <CSSTransition classNames="animate-fade" timeout={150}>
```
LZoog: This seems gross to me because I'd like to not have to specify `animate-fade` with `timeout={150}` every time, but if we want to control the conditional in _this_ file and use an animation on `ErrorMessage`, we need `CSSTransition` in the same place as the conditional and `TransitionGroup` around the whole thing. (At least, this was the only way I could get it working.)

`TransitionGroup` sticks this all in a `div` by default, `component={null}` just says "don't add an extra container component".

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -249,10 +234,48 @@ export const Product = ({
...
+            ) : (
+              <DialogMessage
+                className="dialog-error"
+                onDismiss={resetCreateSubscriptionError}
+              >
+                <h4 data-testid="error-subscription-failed">
+                  Subscription failed
+                </h4>
+                <p>{createSubscriptionStatus.error.message}</p>
+              </DialogMessage>
```
LZoog: I removed this from being its own component (`CreateSubscriptionErrorDialog`) because `onDismiss` is required for `DialogMessage` but not for `ErrorMessage`. TypeScript complained when I made it optional since it could be undefined.

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
```diff
@@ -140,7 +140,6 @@ function init() {
                 'https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/api.md#response-format',
             }),
           },
-          resetUpdatePayment: linkTo('routes/Subscriptions', 'subscribed'),
```
LZoog: This one is a little odd - I removed this because without it, you don't see the error at all because you're immediately redirected. Ideally it's triggered on form change, however, even if that happened, since `updateRevealed` is handled internally, I think you'd have to click on "Change" anyway to see the error was gone, so this seems OK to me.

johngruen:
Thank you @LZoog , this looks solid!

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -140,6 +141,23 @@ export const Product = ({
...
+      createSubscriptionStatus.error.code === 'card_declined'
+    ) {
+      resetCreateSubscriptionError();
+    }
+  }, [
+    createTokenError,
+    createSubscriptionStatus,
+    setCreateTokenError,
+    resetCreateSubscriptionError,
+  ]);
```
LZoog: @lmorchard I originally had this callback like so:

```
  const onChangeErrorDismiss = useCallback(() => {
    setCreateTokenError({ type: '', error: false });
    resetCreateSubscriptionError();
  }, [setCreateTokenError, resetCreateSubscriptionError]);
```

... but it fired on every `onChange` event from the form. This is probably a micro-optimization since React is really fast - you think the change I made is worth keeping? It only resets the error when it needs to now, but it's kind of less scalable.

(PR is ready for review)

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -154,7 +156,10 @@ export const PaymentForm = ({
       validator={validator}
       onSubmit={onSubmit}
       className="payment"
-      onChange={engage}
+      onChange={() => {
```
lmorchard: We should try to avoid defining inline functions for event handlers like this. Better way might be to rename [the `engage()` callback](https://github.com/mozilla/fxa/blob/34b9bcccb5ecbc9fbc155bde2ae779962e466e3c/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L120) to a more generic `onChange` and move the `onChangeErrorDismiss()` call there.

(I've also just realized all the `onValidate` handlers are inline functions, which is poor form but I/we can refactor at some future point.)

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -346,16 +346,13 @@ describe('routes/Product', () => {
...
       getByTestId,
       findByTestId,
       queryByTestId,
-      debug,
     } = await commonSubmitSetup(createToken);
     fireEvent.click(getByTestId('submit'));
     await findByTestId('error-payment-submission');
-    fireEvent.click(getByTestId('dialog-dismiss'));
-    expect(queryByTestId('error-payment-submission')).not.toBeInTheDocument();
+    expect(queryByTestId('error-payment-submission')).toBeInTheDocument();
```
lmorchard: I think this `expect()` ends up redundant, since the preceeding `await findByTestId()` will fail the test after a little while if the element isn't found.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -412,8 +409,35 @@ describe('routes/Product', () => {
...
 
-    await findByTestId('error-card-declined');
+    await findByTestId('error-card-rejected');
     expect(queryByText(message)).toBeInTheDocument();
+
+    // hide on form change
+    fireEvent.change(getByTestId('zip'), { target: { value: '12345' } });
+    // wait for the animation to finish
+    setTimeout(() => {
+      expect(getByTestId('error-payment-submission')).not.toBeInTheDocument();
```
lmorchard: Instead of a `setTimeout`, this is probably a good spot to use `waitForExpect`. For example: https://github.com/mozilla/fxa/blob/34b9bcccb5ecbc9fbc155bde2ae779962e466e3c/packages/fxa-payments-server/src/routes/Subscriptions/index.test.tsx#L330
```javascript
    await waitForExpect(() =>
      expect(cancelButton).not.toHaveAttribute('disabled')
    );
```

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -412,8 +409,35 @@ describe('routes/Product', () => {
...
+    } = await commonCreateSubscriptionFailSetup('incorrect_cvc', message);
+    fireEvent.click(getByTestId('submit'));
+
+    await findByTestId('error-card-rejected');
+    expect(queryByText(message)).toBeInTheDocument();
+
+    // hide on form change
+    fireEvent.change(getByTestId('zip'), { target: { value: '12345' } });
+    // wait for the animation to finish
+    setTimeout(() => {
```
lmorchard: Also a good spot for `waitForExpect` I think

packages/fxa-payments-server/src/routes/Product/index.tsx
lmorchard: Y'know, without benchmarking, I couldn't say which is more worth doing... though I'd say since you already did the work of making this callback more discriminating about what it calls, I'd just leave it this way for now?

packages/fxa-payments-server/src/routes/Product/index.tsx
lmorchard: Could we move the repeated `TransitionGroup` and `CSSTransition` boilerplate into the new `ErrorMessage` component - but maybe add an `visible` prop for the conditional (i.e. `visible={createTokenError.error}` and use `{visible && ( <CSSTransition...` in `ErrorMessage`? Seems like that would render as nothing if the only thing returned is an empty `TransitionGroup`

packages/fxa-payments-server/src/routes/Product/index.tsx
lmorchard: Should the `<DialogMessage>` be a child of the `<TransitionGroup>` here? We could extract the conditional out into a variable like `isCVCError` up next to `inProgress`. Then, use `isCVCError && <ErrorMessage>` and `!isCVCError && <DialogMessage>`

```javascript
const isCVCError =
  createSubscriptionStatus.error !== null &&
  ( createSubscriptionStatus.error.code === 'card_declined' ||
    createSubscriptionStatus.error.code === 'incorrect_cvc');
```

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -249,10 +245,34 @@ export const Product = ({
       <h3 className="billing-title">
         <span>Billing Information</span>
       </h3>
+
+      <ErrorMessage isVisible={!!createTokenError.error}>
+        {createTokenError.error && (
```
LZoog: There is a little bit of redundancy around the checks with this approach or we get complaints from React, but it's so much better than it was. 👌

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -125,6 +125,22 @@ export const PaymentUpdateForm = ({
         </div>
       ) : (
         <>
+          <ErrorMessage isVisible={!!createTokenError.error}>
```
LZoog: I like `isVisible` being a Boolean value since it's more specific, but I know there's a gotcha around using `!!` if this is ever an empty string (it's not in this case) and some folks just don't like it in general. I liked it over `new Boolean(createTokenError.error)` or passing in the entire object + taking in an `any` type for `isVisible` 🤷‍♀️

PR #3275, opened 2019-11-07
fix(password): Add inline confirm password validation

### LINKED ISSUE #631: Validate password & vpassword inline instead of on form submit
Blocks https://github.com/mozilla/fxa-content-server/issues/6083

I suspect this could be handled on #vpassword blur. 

Other changes:
* Highlight both fields
* Attach the tooltip to the first field.

screen shot 2018-05-21 at 14 43 53

vbudhram:
Fixes #631 

inline_password

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sign_up_password.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_up_password.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { assign } from 'underscore';
+import { assign, debounce } from 'underscore';
 import AuthErrors from '../lib/auth-errors';
 import CWTSOnSignupPasswordExperimentMixin from './mixins/cwts-on-signup-password-experiment-mixin';
 import Cocktail from 'cocktail';
@@ -22,6 +22,10 @@ import Template from 'templates/sign_up_password.mustache';
 
 const t = msg => msg;
 
+const PASSWORD_INPUT_SELECTOR = '#password';
+const VPASSWORD_INPUT_SELECTOR = '#vpassword';
+const DELAY_BEFORE_PASSWORD_CHECK_MS = 1500;
+
 const proto = FormView.prototype;
 const SignUpPasswordView = FormView.extend({
   template: Template,
@@ -32,6 +36,7 @@ const SignUpPasswordView = FormView.extend({
 
   events: assign({}, FormView.prototype.events, {
     'click .use-different': preventDefaultThen('useDifferentAccount'),
+    'keyup #vpassword': '_onConfirmPasswordKeyUp',
   }),
 
   useDifferentAccount() {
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -30,6 +33,8 @@ const SignUpPasswordView = FormView.extend({
 
   events: assign({}, FormView.prototype.events, {
     'click .use-different': preventDefaultThen('useDifferentAccount'),
+    'keyup #vpassword': '_onConfirmPasswordKeyUp',
+    'blur #vpassword': '_onConfirmPasswordBlur',
```
vbudhram: I added check on blur because if user tabs to age input and the delay hasn't check it, it will automatically show the tooltip error. It seemed to be a little smoother.

packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -20,6 +20,9 @@ import Template from 'templates/sign_up_password.mustache';
 
 const t = msg => msg;
 
+const PASSWORD_INPUT_SELECTOR = '#vpassword';
+const DELAY_BEFORE_PASSWORD_CHECK_MS = 2000;
```
vbudhram: @ryanfeeley 2 seconds seem reasonable to me while manually testing. Any objections?

vbudhram:
@mozilla/fxa-devs I think this is ready! r?

packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -104,6 +120,20 @@ const SignUpPasswordView = FormView.extend({
...
     return this._getPassword() === this._getVPassword();
   },
+
+  _onConfirmPasswordKeyUp() {
+    this.checkPasswordsMatchDebounce();
+  },
+
+  _onConfirmPasswordBlur() {
+    this._checkPasswordsMatch();
+  },
```
chenba: Does this mean the check is performed twice when I tab out of the confirm password field?  Once on the blur, and then again `DELAY_BEFORE_PASSWORD_CHECK_MS` after the last debounce call?

vbudhram: Hmm yea, it does. Let me add a few checks to make this work a little smoother.

vbudhram:
@chenba made the updates!

vbudhram:
@davismtl is this something you would like to get on train 150?

packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -104,6 +123,27 @@ const SignUpPasswordView = FormView.extend({
   _doPasswordsMatch() {
     return this._getPassword() === this._getVPassword();
   },
+
+  _onConfirmPasswordKeyUp() {
+    if (!passwordCheckLock) {
+      passwordCheckLock = true;
```
chenba: This changes the previous debounce behavior.  IIRC previously `debounce` is called on every keyup.  With this extra state, `debounce` is called on the first keyup, then again on the first keyup after `DELAY_BEFORE_PASSWORD_CHECK_MS`.  If the blur handler executes during the delay, it won't perform the check since `passwordCheckLock` will be true.

Is this intentional?

vbudhram: Yea this was intentional because otherwise you would see a slight flicker on the tooltip when called twice.

vbudhram:
So on further investigating and testing, I am opting to remove the password match check on `blur`. This means that when a user tabs over to the age field, it will rely on the debounced password check rather than performing the check automatically. 

In practice it works wells and I don't see a huge difference. The second password field will get focused as part of the debounced function.

Keeping the original behavior did not play well with the first password field (you couldn't focus the input to change it). I think we can come up with a more general solution that will work for both password fields as part of the work @LZoog is doing.

cc @davismtl @mozilla/fxa-devs r?

PR #3278, opened 2019-11-08
Use lua scripts for redis transactions

dannycoates:
More transactional operations are coming with the move of oauth tokens
into redis that the existing `update` function is not well equipped to
handle. By moving transactional operations into lua we can also simplify
our connection handling because we'll no longer require a pool for serial
message isolation. Converting the existing uses of `update` to lua helps
prove out the concept.

Moving the custom serialization of the session tokens into lua allows
the server logic to be agnostic about the storage layer details.


packages/fxa-auth-server/lib/luaScripts/getSessionTokens.lua
```diff
@@ -0,0 +1,52 @@
+local REDIS_SESSION_TOKEN_PROPERTIES = {
```
shane-tomlinson: Can lua scripts be unit tested somehow? Are they tested via integration tests?

packages/fxa-auth-server/lib/luaScripts/getSessionTokens.lua
```diff
@@ -0,0 +1,52 @@
...
+}
+local REDIS_SESSION_TOKEN_LOCATION_PROPERTIES = {
+  'city',
+  'state',
+  'stateCode',
+  'country',
+  'countryCode'
+}
+
+local function unpackArray(keys, values)
```
shane-tomlinson: I'm not normally one to ask to document "how" things work, but I'm a redis-lua n00b and have no idea which is what and what is that. Some brief documentation about what's happening here would help other newcomers too. Are unpack and unpackArray just deserializing the result that's stored in redis? If so, how come it's not possible to use cjson.decode(value) directly?

packages/fxa-auth-server/lib/luaScripts/touchSessionToken.lua
```diff
@@ -0,0 +1,49 @@
...
+  'countryCode'
+}
+
+local function packToken(token)
+  local result = {}
+  for i, name in ipairs(REDIS_SESSION_TOKEN_PROPERTIES) do
+    if name == 'location' and type(token[name]) == 'table' then
+      local location = token[name]
+      local r = {}
+      for j, n in ipairs(REDIS_SESSION_TOKEN_LOCATION_PROPERTIES) do
```
shane-tomlinson: I wrote a big long comment here about why and then went back up to [this comment](https://github.com/mozilla/fxa/pull/3278/files?w=1#diff-3758ff68a278bb5d275d4003ab6831dcL12) which has answers to all the questions I had.

packages/fxa-auth-server/lib/luaScripts/pruneSessionTokens.lua
```diff
@@ -0,0 +1,27 @@
...
+
+local function decode(value)
+  if not value then
+    return {}
+  end
+  return cjson.decode(value)
+end
+
+local tokens = decode(redis.call('get', KEYS[1]))
+local tokenIds = decode(ARGV[1])
```
shane-tomlinson: Reading through the [redis lua docs](https://redis.io/commands/eval#atomicity-of-scripts), I'm confused by the difference between KEYS and ARGV.

packages/fxa-auth-server/lib/redis.js
```diff
@@ -4,37 +4,106 @@
...
+    return this.redis.touchSessionToken(uid, JSON.stringify(token));
   }
 
-  return Object.entries(redis).reduce((object, [key, value]) => {
-    if (typeof value === 'function') {
-      object[key] = async (...args) => {
-        try {
-          return await value(...args);
-        } catch (err) {
-          if (err.message === 'redis.watch.conflict') {
```
shane-tomlinson: Wahey, no more conflicts! I assume this is because of [redis' guarantee of the atomicity of scripts](https://redis.io/commands/eval#atomicity-of-scripts) and this snippet: `Redis guarantees that a script is executed in an atomic way: no other script or Redis command will be executed while a script is being executed.`

packages/fxa-shared/redis/index.js
```diff
@@ -1,78 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
dannycoates: removing fxa-shared/redis is just me seeing what breaks. prob won't remove it, but update it to the impl I'm testing in auth-server

packages/fxa-auth-server/lib/luaScripts/getSessionTokens.lua
dannycoates: we can test them with mocha via our redis.js against a real redis server

packages/fxa-auth-server/lib/luaScripts/pruneSessionTokens.lua
dannycoates: I had to research this too. When using a sharded redis cluster KEYS gets used for routing and to ensure all KEYS present are on the same node (otherwise it's an error). ARGV get no special consideration. If we ever decide to use a sharded cluster we'll have to make sure our keys play nicely with the [key distribution model](https://redis.io/topics/cluster-spec#keys-distribution-model) (they don't currently, but could be made to)

packages/fxa-auth-server/lib/redis.js
dannycoates: precisely

packages/fxa-auth-server/lib/db.js
```diff
@@ -322,26 +291,19 @@ module.exports = (config, log, Token, UnblockCode = null) => {
...
     }
 
     return P.all(promises).spread(
       (mysqlSessionTokens, redisSessionTokens = {}) => {
+        // console.error(uid, redisSessionTokens)
         // for each db session token, if there is a matching redis token
         // overwrite the properties of the db token with the redis token values
-        const lastAccessTimeEnabled =
-          isRedisOk && features.isLastAccessTimeEnabledForUser(uid);
+        const lastAccessTimeEnabled = features.isLastAccessTimeEnabledForUser(
```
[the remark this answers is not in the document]
dannycoates: good question. I don't know yet, but figuring it out is still on my todo list

packages/fxa-auth-server/lib/luaScripts/getSessionTokens.lua
dannycoates: We ought to be able to prove backward compatibility with test cases

packages/fxa-auth-server/lib/luaScripts/touchSessionToken.lua
dannycoates: Yeah, I'll port that comment and ref the og PR

dannycoates:
#3385

cc @shane-tomlinson

packages/fxa-auth-server/lib/db.js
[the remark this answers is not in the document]
dannycoates: I've verified that when redis is down we return an empty tokens object which once we get down to `mergeDeviceInfoFromRedis` is ok because it doesn't override anything

packages/fxa-auth-server/test/local/db.js
```diff
@@ -412,885 +424,60 @@ describe('redis enabled, token-pruning enabled:', () => {
...
+  it('should call redis.pruneSessionTokens in db.createSessionToken', () => {
     return db.createSessionToken({ uid: 'wibble' }).then(() => {
-      assert.equal(redis.update.callCount, 1);
-      assert.equal(redis.update.args[0].length, 2);
-      assert.equal(redis.update.args[0][0], 'wibble');
-      assert.equal(typeof redis.update.args[0][1], 'function');
-    });
-  });
-
-  it('db.devices handles old-format and new-format token objects from redis', () => {
```
dannycoates: these old-format tests get replace by the single redis test on getSessionTokens

dannycoates:
I'm not _quite_ sure of the best way to handle the "redis down" scenario (or how important it is). The old `safeRedisGet` left something to be desired (it's dependent on the behavior of the lower level redis lib and not very clear) but I didn't want to remove it without a replacement. I went with a `Promise.race` on the lua functions, but I left the regular redis calls as is, which tests show to be equivalent in our current usage. I'm fuzzy on what, if any, strategy we need for the general case. I imagine we can address that in a future PR.

packages/fxa-auth-server/lib/verification-reminders.js
```diff
@@ -165,23 +165,28 @@ module.exports = (log, config) => {
...
-              .reduce(async (reminders, item, index) => {
-                if (index % 2 === 0) {
-                  const uid = item;
-                  let metadata = await redis.get(`${METADATA_KEY}:${uid}`);
-                  if (metadata) {
-                    const [flowId, flowBeginTime] = JSON.parse(metadata);
-                    metadata = { flowId, flowBeginTime };
-                    if (keyIndex === keys.length - 1) {
-                      await redis.del(`${METADATA_KEY}:${uid}`);
+              .then(async items => {
```
dannycoates: switching to native promises means a slight refactor; i used a `for` loop

packages/fxa-auth-server/lib/luaScripts/pruneSessionTokens_1.lua
```diff
@@ -0,0 +1,29 @@
+-- https://github.com/mozilla/fxa-auth-server/pull/2254 for packing scheme info
```
chenba: Packing scheme doesn't seem relevant to this script.

packages/fxa-auth-server/lib/luaScripts/readme.md
```diff
@@ -0,0 +1,22 @@
+# Redis Lua Scripts
+
+Our lua scripts are for transactions and operations more complex than individual Redis commands.
+
+## Script Naming Convention
+
+- camelCase
+- `${functionName}_${number of KEYS parameters}.lua`
```
chenba: 💯

packages/fxa-auth-server/lib/redis.js
```diff
@@ -4,37 +4,125 @@
...
 
-module.exports = (config, log) => {
-  const redis = require('../../fxa-shared/redis')(config, log);
-  if (!redis) {
-    return;
+const scriptNames = readdirSync(resolve(__dirname, 'luaScripts'), {
+  withFileTypes: true,
+})
+  .filter(de => de.isFile() && extname(de.name) === '.lua')
+  .map(de => basename(de.name, '.lua'));
```
chenba: What does `de` stand for?  I can't figure it out.

dannycoates: [Directory Entry](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_class_fs_dirent) but if you need to ask it's prob a bad shorthand

packages/fxa-auth-server/lib/luaScripts/pruneSessionTokens_1.lua
dannycoates: true

packages/fxa-auth-server/lib/verification-reminders.js
chenba: `items.reduce` should work right?  I don't understand why a `for...of` loop is necessary.

packages/fxa-auth-server/test/local/redis.js
```diff
@@ -1,143 +1,184 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
chenba: 😆

packages/fxa-auth-server/test/local/redis.js
```diff
@@ -1,143 +1,184 @@
...
 
-    it('rejected with 409 conflict', () => {
-      assert.isUndefined(result);
-      assert.isObject(err);
-      assert.equal(err.errno, error.ERRNO.REDIS_CONFLICT);
-      assert.equal(err.message, 'Redis WATCH detected a conflicting update');
-      assert.equal(err.output.statusCode, 409);
-      assert.equal(err.output.payload.error, 'Conflict');
+    it('does nothing for unknown uids', async () => {
+      await redis.pruneSessionTokens('x');
```
chenba: This test doesn't have any assertions.

packages/fxa-auth-server/test/local/redis.js
```diff
@@ -1,143 +1,184 @@
...
+      const rawData = await redis.get(uid);
+      assert.isNull(rawData);
+    });
+  });
+});
+
+describe('Redis down', () => {
+  describe('touchSessionToken', () => {
+    it('returns without error', async () => {
+      await downRedis.touchSessionToken(uid, {});
```
chenba: No assertion?

chenba:
Are we going to remove `fxa-shared/redis`?

dannycoates:
> Are we going to remove `fxa-shared/redis`?

not yet, but it's probably safe to remove the `update` function now. I'll see in a follow-up

packages/fxa-auth-server/lib/verification-reminders.js
dannycoates: Array.reduce isn't async afaik. I seem to remember trying that

chenba: Ah, ok.  Got it.  👍

dannycoates:
note to self: rebase

packages/fxa-auth-server/test/local/redis.js
```diff
@@ -4,140 +4,187 @@
...
     });
+  });
 
-    it('called the underlying redis method', () => {
-      assert.equal(redis.foo.callCount, 1);
+  describe('pruneSessionTokens', () => {
+    it('throws a timeout error', async () => {
+      try {
+        await downRedis.pruneSessionTokens(uid);
+      } catch (e) {
```
vbudhram: Could we use something like https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/test/assert.js#L15 here?

packages/fxa-auth-server/lib/luaScripts/getSessionTokens_1.lua
```diff
@@ -0,0 +1,58 @@
...
+local function unpackSessions(value)
+  if not value then
+    return {}
+  end
+  local result = {}
+  local tokens = cjson.decode(value)
+  for id, values in pairs(tokens) do
+    if type(values) == 'table' and values[1] then
+      -- it's an array
+      local token = unpackArray(REDIS_SESSION_TOKEN_PROPERTIES, values)
```
vbudhram: By moving to lua scripts would we no longer have to `truncatePackedArray`?

dannycoates: good question! By virtue of `JSON.stringify` keys with `undefined` values get eliminated. Here, missing keys will be replaced by `null` if they are leading and excluded if they're trailing. For example
```
...
"location": {"state":"FL"},
...
```
gets encoded to
```
... [null,"WA"] ...
```
because city was undefined, along with the rest

But we also ought to remove `"uaFormFactor":null` so I've updated the stringify in redis.js to:
```js
// remove keys with null values
const json = JSON.stringify(token, (k,v) =>(v == null ? undefined : v))
```

packages/fxa-auth-server/test/local/redis.js
dannycoates: i suppose. for now i prefer this since it's one less level of indirection

PR #3279, opened 2019-11-08
feat(email-first): Add tab synchronization in email first flow

### LINKED ISSUE #3269: Email first: /, /signin, /signup do not stay in sync if user signs in in other tab
If a user is at any of `/`, `/signin`, or `/signup`, and the user signs in in another tab, the original tab should redirect to /settings  like the legacy `/signin` and `/signup` pages.

blocks #639

shane-tomlinson:
There was no tab synchronization on the /, /signin, and /signup
pages in the email-first flow.

This adds synchronization, if the user is on any of /, /signin,
/signup then signs in using another tab, redirect the user to /settings.

fixes #3269


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3281, opened 2019-11-08
fix(test): Ensure settings_common tests are run

### LINKED ISSUE #3280: settings_common functional tests are not run
The functional tests in the content server's /tests/functional/settings_common.js are not run because the [suites are incorrectly registered](https://github.com/mozilla/fxa/blob/4cf9f7cfbf81bf7c46ed54461ad741c8337a36f7/packages/fxa-content-server/tests/functional/settings_common.js#L157). The suites register the `<suitename>.test` rather than just `<suitename>`.

And above, the tests are added to the suite directly rather than to the `<suite>.tests` property.

shane-tomlinson:
The functional tests in /tests/functional/settings_common.js were
not being run because the suites were being added to the test
runner incorrectly.

This ensures the tests run, as well as pass (they didn't once
enabled), and converts inline selectors to use the selectors
module.

fixes #3280

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_common.js
+++ b/packages/fxa-content-server/tests/functional/settings_common.js
@@ -5,28 +5,29 @@
 'use strict';
 
 const { registerSuite } = intern.getInterface('object');
-const nodeXMLHttpRequest = require('xmlhttprequest');
-const FxaClient = require('fxa-js-client');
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3282, opened 2019-11-08
fix(email-first): Redirect to correct settings page after authentication

### LINKED ISSUE #3270: email-first: user is not redirected to correct settings page after signing in
STR

1. In a fresh profile, browse directly to https://accounts.firefox.com/settings/avatar/change
2. Sign in with an existing account
3. We are redirected to /settings instead of /settings/avatar/change

blocks #639

shane-tomlinson:
If a user tried to go to, e.g., /settings/avatar/change
without having a session, they were asked to sign in.
After sign-in, they were redirected back to /settings
rather than /settings/avatar/change because redirectTo
was not correctly propagated. Instead of propagating
the value in the view->view model, rely on the value
stored in the relier.

This also changes the default behavior of an unauthenticdated
user - instead of sending the user to `/signin`,
send the user to `/`

builds on #3281 

fixes #3270

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/index.js
+++ b/packages/fxa-content-server/app/scripts/views/index.js
@@ -217,7 +217,9 @@ class IndexView extends FormView {
           // ensure it's email is set.
           account.set('email', email);
         }
-        this.navigate(nextEndpoint, { account });
+        this.navigate(nextEndpoint, {
+          account,
+        });
       });
   }
 }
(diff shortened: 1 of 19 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

packages/fxa-content-server/app/scripts/views/index.js
```diff
@@ -217,7 +217,9 @@ class IndexView extends FormView {
           // ensure it's email is set.
           account.set('email', email);
         }
-        this.navigate(nextEndpoint, { account });
+        this.navigate(nextEndpoint, {
+          account,
+        });
```
chenba: Prettier works in mysterious ways.

PR #3284, opened 2019-11-08
feat(email-first): delete-account now completes at /

shane-tomlinson:
extraction from #3209


packages/fxa-content-server/app/scripts/views/settings/delete_account.js
```diff
@@ -161,8 +161,17 @@ var View = FormView.extend({
...
+        // to redirect a user to signup/signin. Clear them to avoid
+        // sending the user directly to the signup page.
+        this.relier.unset('email');
+        this.relier.unset('uid');
+        // force email-first until it's the default flow everywhere.
+        this.relier.set('action', 'email');
+
         this.navigate(
-          'signup',
+          '/',
```
vladikoff: :+1:

PR #3287, opened 2019-11-08
feat(sign-up): validate email domain with DNS

### LINKED ISSUE #3196: Do an MX Lookup on Domain of Email Used for Registration
At my past job, we did MX lookups on the domain of the email address being used for registration to make sure people didn't have typos and to make sure we wouldn't send emails that would bounce. This improved our registrations because it highlighted some types of potential typos that are less visible to a regex.

Obviously if the domain is Google, Yahoo, msn, hotmail, gmx, outlook, etc... we don't need to do a lookup.

How would we surface there's a problem? We may not have time to do a lookup before the user hits submit on an email first flow. If we can do it there, great but we could potentially surface it at the next step (registration form). We have a link to change the email for registration.

Our metrics show that the majority of emails that are bouncing are sent to a domain that is not one of the above. This is where the MX lookup would help.

See chart:
https://analytics.amplitude.com/mozilla-corp/chart/new/osniv8u

The volume in the above chart shows that nearly 5% of our registration submits will result in a bounced email sent to a domain with potentially no MX record. (I don't think we distinguish soft from hard bounces in our metrics)

Edit:
[Feature document here](https://docs.google.com/document/d/1qKy8NFOasmCvwniqhso8ndFjMecI595Z7dW6tccUjI4/edit#)

chenba:
This patch adds an experiment in which some sign-up email addresses'
domain names will be validated on the server side with DNS lookups.
Doing so might help reduce email bounces from typos.

Fixes #3196

@mozilla/fxa-devs r?


packages/fxa-content-server/app/scripts/lib/top-email-domains.js
```diff
@@ -0,0 +1,195 @@
+// Email domains that we are not going to validate.
```
chenba: This file will be augmented with top domains from our data.  #TODO

6a68:
What's the right way to test this feature locally? I'm not sure how to force my browser session into one experiment group or the other.

chenba:
@6a68 Try going to http://127.0.0.1:3030/?action=email&forceExperiment=emailMxValidation&forceExperimentGroup=treatment

packages/fxa-content-server/app/scripts/lib/mailcheck.js
```diff
@@ -35,6 +35,53 @@ const SECOND_LEVEL_DOMAINS = [
   'outlook',
   'live',
   'yahoo',
+  'icloud',
```
shane-tomlinson: Will adding these domains to mailcheck skew the A/B test since all users will receive earlier feedback as to whether a domain might be invalid or not?

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -0,0 +1,116 @@
...
+    );
+    tooltip.render();
+    view.logEvent('emailDomainValidation.fallback');
+  };
+
+  return new Promise((resolve, reject) => {
+    const emailAddress = $element.val();
+    const [, domain] = emailAddress.split('@');
+
+    if (TopEmailDomains.has(domain)) {
```
shane-tomlinson: Can you leave a comment here about if the domain is part of a known good list, then skip the network check?

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -0,0 +1,116 @@
...
+      return resolve();
+    }
+
+    // User could repeatedly submit the form
+    if ($element.data('previousDomain') === domain) {
```
shane-tomlinson: How come this data is stored on the element rather than just locally? Is this shared state with mailcheck?

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -0,0 +1,62 @@
...
+    if (NotFoundErrorCodes.includes(err.code)) {
+      return false;
+    }
+    throw err;
+  }
+};
+
+module.exports = function() {
+  return {
+    method: 'post',
```
shane-tomlinson: Semantically, a POST seems strange here since we aren't actually creating/updating any records. How come POST instead of a GET? Is it to keep the domain name from being logged? If so, can you leave a comment about why?

packages/fxa-content-server/app/scripts/views/index.js
```diff
@@ -214,12 +222,24 @@ class IndexView extends FormView {
...
           account = this.user.getAccountByEmail(email);
           // the returned account could be the default,
-          // ensure it's email is set.
+          // ensure its email is set.
           account.set('email', email);
+          this.navigate(nextEndpoint, { account });
+        } else {
+          // The email address does not belong to a current user. Validate its
+          // domain name.
+          if (this.isInExperimentGroup('emailMxValidation', 'treatment')) {
```
shane-tomlinson: For this experiment to be reported correctly, you'll need to do a bit more setup.

First, you'll need to add `emailMxValidation` to the list [lib/experiment.js](https://github.com/chenba/fxa/blob/6e5714881980035daa086bdd2dddd9674f67654b/packages/fxa-content-server/app/scripts/lib/experiment.js#L23).

And here, you need to call [createExperiment](https://github.com/chenba/fxa/blob/6e5714881980035daa086bdd2dddd9674f67654b/packages/fxa-content-server/app/scripts/lib/experiment.js#L170) with the experiment group.

You can do something like 

```js
const group = this.getExperimentGroup('emailMxValidation');
this.createExperiment('emailMxValidation', group);
if (group === 'treatment') {
...
}
```

What will the ultimate measure of success be, whether the user has verified their email?

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -0,0 +1,116 @@
...
+    view.logEvent('emailDomainValidation.triggered');
+
+    resolveDomain(domain)
+      .then(resp => {
+        const { result } = resp;
+        $element.data('previousDomain', domain);
+        $element.data('previousDomainResult', result);
+
+        if (result === 'MX') {
+          view.logEvent('emailDomainValidation.success');
```
shane-tomlinson: Do any of these events need to be reported to Amplitude? If so, you'll need to work with @irrationalagent on the event names/properties, and add some translation [to this list of events in server/lib/amplitude.js](https://github.com/chenba/fxa/blob/6e5714881980035daa086bdd2dddd9674f67654b/packages/fxa-content-server/server/lib/amplitude.js#L40).

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -0,0 +1,170 @@
...
+    $el.data('previousDomainResult', 'A');
+    sinon.spy($el, 'data');
+    return checkEmailDomain($el, view).then(() => {
+      assert.isTrue($el.data.calledWith('previousDomain'));
+      assert.isTrue($el.data.calledWith('previousDomainResult'));
+    });
+  });
+
+  it('should resolve on a successful MX record lookup', () => {
+    sinon.stub(xhr, 'ajax').returns(Promise.resolve({ result: 'MX' }));
```
shane-tomlinson: TIL about `sinon.stub().returns` and `sinon.stub().resolves`, which looks like you could use the latter here?

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -0,0 +1,62 @@
...
+          .string()
+          .hostname()
+          .required(),
+      },
+    },
+    process: async function(req, res, next) {
+      const { domain } = req.body;
+
+      try {
+        if (await tryResolveWith(resolver.resolveMx.bind(resolver))(domain)) {
```
shane-tomlinson: A series of questions about this:

* Are results cached at all? 
* Does this rely on the OS resolver which caches requests, or does it make a fresh request to the upstream DNS server every time? I couldn't figure that out from the node API documentation.
* Do we have any idea how many of these requests will be made assuming common domains are all part of the front-end allow lists that bypass this check? 
* Do requests time out eventually?
* Does this make 2 requests to the DNS resolver if there is no MX record? Can we make a single request to [`resolve`](https://nodejs.org/docs/latest-v10.x/api/dns.html#dns_dnspromises_resolve_hostname_rrtype) and rely upon the array of returned results instead?

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
chenba: True, `checkEmailDomain` scope variables should work here.  Thanks!

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
chenba: @davismtl Do we need to report email domain validation results to Amplitude?

packages/fxa-content-server/app/scripts/lib/mailcheck.js
chenba: I think so.  I can add these domains in another PR once the experiment's over.

packages/fxa-content-server/app/scripts/views/index.js
chenba: Thanks!

Yep, ultimate measure of success is sign up completion.

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
chenba: Huh, it did not occur to me whatsoever, but looks like I could.  I'm not sure what we'd gain from it though?

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
chenba: I viewed it as a RPC instead of a RESTful API, so I did not place a lot of emphasis on the HTTP method.  Sending payloads in the body feels "cleaner".  Obviously it was the wrong choice; if it confused you then it will confuse others.  I will change it to GET. 👍

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
chenba: > Are results cached at all?

From what I gathered in my quick read on the docs of the pertinent Node functions while I was working on this, no caching.

> Does this rely on the OS resolver which caches requests, or does it make a fresh request to the upstream DNS server every time? I couldn't figure that out from the node API documentation.

The Node docs do make its readers hunt around a bit, but the info is there, and the `resolve*` functions "always use the network to perform DNS queries".  `dns.lookup` is the one function that relies on OS facilities.

> Do we have any idea how many of these requests will be made assuming common domains are all part of the front-end allow lists that bypass this check?

Maybe @davismtl have the stats to give an estimate here?  In percentage terms, it is low, but I don't know what it might be in number of requests.

> Do requests time out eventually?

Yes.  There's a timeout error code.  There's no option to set a timeout from the calling code.

> Does this make 2 requests to the DNS resolver if there is no MX record? Can we make a single request to resolve and rely upon the array of returned results instead?

Yes.  And now I cannot recall the line of reasoning that led me to doing it that way.  I'll change it.  👍

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
davismtl: @shane-tomlinson asks a good question.

I think we could (should?) track failed and successful domain validation on an email first form submit event property. We should look into this. CC @irrationalagent 

I would however not block on this because we can infer a lot from our volume of bounced emails. I would expect that with this enhancement that we will see bounced emails drop.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
davismtl: > Maybe @davismtl have the stats to give an estimate here? In percentage terms, it is low, but I don't know what it might be in number of requests.

Without giving too many numbers about our registrations, here some numbers that might be useful:
- 80% of registrations are with one of the top 20 domains which should be checked on the front-end. So we will check 20% of our registrations at most.
- 5% of our registration emails bounce and this is what we want to reduce.

chenba: @shane-tomlinson while working on the changes to use a ANY query, I found legitimate reasons to stick with (potentially) two lookups.

From testing the changes, I noticed that Cloudflare (1.1.1.1) does not support ANY, while Google (8.8.8.8) does.  Looks like Cloudflare stopped supporting it in 2015: https://blog.cloudflare.com/deprecating-dns-any-meta-query-type/.  Moreover, https://tools.ietf.org/html/rfc8482 effectivelly officially deprecates ANY.

chenba:
@shane-tomlinson Made changes in https://github.com/mozilla/fxa/pull/3287/commits/272abef31307b83ac7526a175e20d1de9f59acba.  I kept it as a separate commit for ease of re-review.  I will squash once the PR's ready to be merged.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
shane-tomlinson: @jrgm, @jbuck - who are our upstream DNS providers?

shane-tomlinson: > 80% of registrations are with one of the top 20 domains which should be checked on the front-end. So we will check 20% of our registrations at most.

20% of registrations seems alright to me.

packages/fxa-content-server/server/lib/amplitude.js
```diff
@@ -101,6 +101,30 @@ const EVENTS = {
     group: GROUPS.rp,
     event: 'engage',
   },
+  'flow.email-domain-validation.engage': {
+    group: GROUPS.registration,
+    event: 'domain_validation_engage',
```
shane-tomlinson: @irrationalagent, @davismtl - do we want these reported as individual events, or as event properties on the submit event?

packages/fxa-content-server/server/lib/amplitude.js
```diff
@@ -101,6 +101,30 @@ const EVENTS = {
     group: GROUPS.rp,
     event: 'engage',
   },
+  'flow.email-domain-validation.engage': {
```
shane-tomlinson: @irrationalagent, @davismtl do we want these reported as individual events, or as event properties on the submit event?

shane-tomlinson:
> @chenba I asked Alex and Leif whether the amplitude events should be individual events or event properties on the existing fxa_reg - submit event.

AND my questions are already answered here: https://github.com/mozilla/fxa/issues/3196#issuecomment-556473355

packages/fxa-content-server/server/lib/amplitude.js
shane-tomlinson: From https://github.com/mozilla/fxa/issues/3196#issuecomment-556473355, should this be `triggered` rather than `engage`?

packages/fxa-content-server/server/lib/amplitude.js
```diff
@@ -101,6 +101,30 @@ const EVENTS = {
...
+    group: GROUPS.registration,
+    event: 'domain_validation_engage',
+  },
+  'flow.email-domain-validation.skip': {
+    group: GROUPS.registration,
+    event: 'domain_validation_skip',
```
shane-tomlinson: `skip` => `skipped`

packages/fxa-content-server/server/lib/amplitude.js
```diff
@@ -101,6 +101,30 @@ const EVENTS = {
...
+  },
+  'flow.email-domain-validation.success': {
+    group: GROUPS.registration,
+    event: 'domain_validation_success',
```
shane-tomlinson: It looks like `domain_validation_success`, `domain_validation_block` and `domain_validation_warn` should be changed to `domain_validation_result` with an event property `validation_result: (block|warn|success)` to help minimize the number of distinct event types. 

To do so, I *think* we'd have to:

1. Merge these into a single regexp based entry in [FUZZY_EVENTS](https://github.com/mozilla/fxa/blob/e9dbcf5dcb151f4d0ecbaa80db0a20a8bc4f82fd/packages/fxa-content-server/server/lib/amplitude.js#L120) like so:
```
[
  /^flow\.email\-domain\-validation\.(block|warn|success)$/, {
    group: GROUPS.registration,
    event: 'domain_validation_result'
 }
]
```

2. Add an [event property map function to fxa-shared](https://github.com/mozilla/fxa/blob/e9dbcf5dcb151f4d0ecbaa80db0a20a8bc4f82fd/packages/fxa-shared/metrics/amplitude.js#L51):

```js
function mapDomainValidationType(eventType, eventCategory, eventTarget) {
  // this function is called for all fxa_reg event types, only add the event properties
  // for the results pertaining to domain_validation_result.
  if (eventType === 'domain_validation_result') {
    return { validation_result: eventCategory };
  }
}
```

Really, I wish the event property mapper could be defined on the FUZZY_EVENT entry above, not sure how difficult it would be to add that, but it seems like that sort of logic should be as close as possible to the original entry, at least if it's not shared across multiple event types.

chenba: @shane-tomlinson This commit was before @irrationalagent's comment.  Thanks for reviewing, and especially the tip on "fuzzy events".

chenba:
Push a couple of squashed commits, hopefully making re-reviewing easier.

 - d201a31: changes based on Shane's original review
 - be3dda2: Amplitude events

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
jbuck: > @jrgm, @jbuck - who are our upstream DNS providers?

We don't configure anything special here, so whatever AWS uses and whatever GCP uses. We can run tests on instances if you want confirm that `ANY` requests will work here, but perhaps it is better to make the two requests given that it's deprecated

shane-tomlinson: 2 requests it is then!

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -50,25 +51,33 @@ const resolveDomain = domain =>
   });
 
 const checkEmailDomain = ($element, view) => {
+  // The two functions below have their tooltip displaying code wrapped in $()
```
shane-tomlinson: Gotta love obscure problems like this. Nice debugging. Same exact reason we had to do this here: https://github.com/mozilla/fxa/blob/f98235ed4a226e00df985fb0692947557487a2d8/packages/fxa-content-server/app/scripts/views/base.js#L973

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -189,32 +192,34 @@ describe('lib/email-domain-validator', () => {
...
-        assert.isTrue(
-          view.logFlowEvent.calledWith('email-domain-validation.triggered')
-        );
-        assert.isTrue(
-          view.logFlowEvent.calledWith('email-domain-validation.block')
-        );
-
-        AuthErrors.toInvalidEmailDomainError.restore();
-        xhr.ajax.restore();
+        p.delay(50).then(() => {
```
shane-tomlinson: Since this arrow function uses a `{`, You'll need to add `return` here or else any errors in this test will not be caught correctly.

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -103,6 +103,19 @@ function mapDisconnectReason(eventType, eventCategory) {
...
 
+function mapDomainValidationResult(
+  eventType,
+  eventCategory,
+  eventTarget,
+  data
+) {
+  // This function is called for all fxa_reg event types, only add the event
+  // properties for the results pertaining to domain_validation_result.
+  if (eventType === 'domain_validation_result' && eventCategory) {
```
shane-tomlinson: @irrationalagent - if an event, e.g., `domain_validation_result` is emit 3 times with different values for the same event property, will that show up as 3 different events?

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
chenba: Good catch.  I'll remove the `{`.  Thanks!

shane-tomlinson:
The failing tests look to be the same as what @vbudhram is trying to fix over in https://github.com/mozilla/fxa/pull/3479

packages/fxa-shared/metrics/amplitude.js
irrationalagent: Sorry I missed this. Yes, it will be three separate events, one event with each event property value.

vasilicatamas:
@chenba or @shane-tomlinson could you please provide some reliable steps in order to verify this patch?

vasilicatamas:
> @chenba or @shane-tomlinson could you please provide some reliable steps in order to verify this patch?

This question is no longer valid considering that #3196 has been already verified by the QA in stage Train 1.151.0

chenba:
@vasilicatamas Sorry, I was on PTO.  Next time I'll include some steps on the issue.

PR #3289, opened 2019-11-08
[event-broker]: esmoduleinterop, sentry release, and docker optimization

### LINKED ISSUE #3285: [fxa-event-broker]: update hotshot lib to use telegraf format
telegraf uses a different wire format for some statsd style metrics. This is likely why we are not getting any timing information in our system correctly from the event broker.

bbangert:
A few commits made this a bit larger than expected. For convenience this is the full commit log for these changes:

commit 50b4899890fafc5d3e5e3eaee93c090627c5e650 (HEAD -> feat/issue-3285, origin/feat/issue-3285)
Author: Ben Bangert <ben@groovie.org>
Date:   Fri Nov 8 14:26:06 2019 -0800

    feat(event-broker): telegraf opts & sentry info
    
    Because:
    
    * Sentry was not including the release version.
    * hot-shots defults to statsd, and telegraf has some options we
      may want to use at some point.
    
    This commit:
    
    * Includes the Sentry release version.
    * Uses telegraf format for hot-shots metrics reporting.
    
    Closes #3285

commit 538829ec1cfc2998e19636f43fda4d320a0ff73a
Author: Ben Bangert <ben@groovie.org>
Date:   Fri Nov 8 14:25:25 2019 -0800

    refactor(event-broker): use esModuleInterop
    
    Because:
    
    * esModuleInterop is default for new projects as its been generally
      recommended for use.
    
    This commit:
    
    * Turns on esModuleInterop and updates imports as needed.

commit ae2999d72266e8fa7b2ec394da52da16ed4b3898
Author: Ben Bangert <ben@groovie.org>
Date:   Fri Nov 8 14:09:38 2019 -0800

    feat(event-broker): optimize docker build
    
    Because:
    
    * We prefer using slim docker images instead of alpine in production.
    * Prior docker builds were including Typescript build data instead of
      just the built Javascript.
    
    This commit:
    
    * Reduces the compressed size of the Docker build from 139MB to 86 MB.
    * Uses a separate build step to avoid including the Typescript sources.

commit 94de1045884b6c5598c630a1457c6b52e04055b7
Author: Ben Bangert <ben@groovie.org>
Date:   Fri Nov 8 14:08:23 2019 -0800

    feat(event-broker): update dependencies
    
    Because:
    
    * Several dependencies were out of date with a sec. advisory.
    
    This commit:
    
    * Uses the latest versions of the google/sentry libs.


bbangert:
I believe we have some things using node12 already in prod. The dockerfile update was partially based on the latest guardian-vpn dockerfile that @jbuck showed me.

packages/fxa-event-broker/lib/version.ts
```diff
@@ -0,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ * Return version info based on package.json, the git sha, and source repo
+ *
```
jrgm: Not a full review, moreso a drive-by comment, but did you mean to remove the ./lib/app/version.ts ~copy of this, and fixup the require in ./lib/api/proxy-controller.ts to use this new version.ts?

packages/fxa-event-broker/config/index.ts
```diff
@@ -120,6 +120,12 @@ const conf = convict({
       doc: 'Metric prefix for statsD',
       env: 'METRIC_PREFIX',
       format: String
+    },
+    telegraf: {
+      default: true,
```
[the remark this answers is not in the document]
bbangert: It will be if/when we use telegraf features, yes. Otherwise there's no effect for now.

packages/fxa-event-broker/lib/version.ts
bbangert: Good point, yes, the original was supposed to be removed in this change!

PR #3263, opened 2019-11-06
3082 eslint consolidation

meandavejustice:
* fixes #3082
* .eslintrc at root is extended in packages where necessary
* eslint-plugin-fxa is still used in a couple of places
* eslint-plugin-fxa and eslint-config-prettier moved to root dependency list
* grunt removed from fxa-geodb, it was only being used to
    run eslint

eslint config changes are in one commit, while lint fixes are in the second commit

PR #3290, opened 2019-11-08
task(monorepo): eslint consolidation

### LINKED ISSUE #3082: Consolidate eslint configs into eslint-config-fxa
We have a lot of different eslint configs throughout our packages. We also have an old [eslint-config](https://www.npmjs.com/package/eslint-config-fxa). We should consolidate and update the npm package.

meandavejustice:
- fixes #3082
- remove grunt when can be replaced with only eslint
- update to new eslint config using prettier rules
- fix linting errors


shane-tomlinson:
@meandavejustice does this replace #3263? If so, can you close that one?

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -124,7 +124,7 @@ const TESTS = new Map([
       { test: 'include', expected: configUrl('subscriptionTermsUrl', 'new-subscription', 'subscription-terms') },
       { test: 'include', expected: configUrl('subscriptionSupportUrl', 'new-subscription', 'subscription-support') },
       { test: 'include', expected: `Welcome to ${subscriptionProductEmail}!` },
-      { test: 'include', expected: `If you haven\'t already downloaded ${subscriptionProductEmail}, let\'s get started using all the features included in your subscription:` },
+      { test: 'include', expected: `If you haven't already downloaded ${subscriptionProductEmail}, let's get started using all the features included in your subscription:` },
```
6a68: Have you verified that removing the `\` chars doesn't mess up the rendering of the email template? I'm not sure how to test rendering this email, and this change makes me a bit nervous.

packages/fxa-content-server/app/tests/spec/views/subscriptions_management_redirect.js
```diff
@@ -43,7 +43,7 @@ describe('views/subscriptions_management_redirect', function() {
 
     view = new View({
       config,
-      currentPage: `subscriptions`,
+      currentPage: 'subscriptions',
```
6a68: Seems weird to me that we're replacing a lot of template strings with regular strings, but I guess that's what our rules dictate? 🤷‍♂

packages/fxa-content-server/server/.eslintrc
```diff
@@ -1,8 +1,8 @@
...
-  id-blacklist: 0
-  no-var: 2
-  strict: 0
+{
+  "extends": "../.eslintrc",
+  "rules": {
+    "id-blacklist": 0,
+    "no-var": 2,
+    "strict": 0
+  }
```
6a68: Is there a reason the .eslintrc here is pretty different from the others? It's fine if we need to special case content-server, just want to check.

packages/fxa-customs-server/package.json
```diff
@@ -16,11 +16,12 @@
     "start": "node bin/customs_server.js",
     "lint": "npm-run-all --parallel lint:*",
     "lint:deps": "npm audit --json | audit-filter --nsp-config=.nsprc --audit=-",
+    "lint:eslint": "eslint .",
     "test": "scripts/test-local.sh",
     "format": "prettier '**' --write"
   },
   "dependencies": {
-    "@google-cloud/pubsub": "^0.31.1",
+    "@google-cloud/pubsub": "1.1.5",
```
6a68: Hm, a major version bump. Is this supposed to be part of this PR?

packages/fxa-payments-server/server/lib/routes/post-metrics.test.js
```diff
@@ -31,27 +31,37 @@ const invalidBody = {
...
+  test('POST valid input should return 200', () => {
+    return new Promise(done => {
+      request(app)
+        .post('/metrics')
+        .send(validBody)
+        .set('Accept', 'application/json')
+        .expect(200)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
```
6a68: I don't have a lot of experience with async Jest tests, but based on reading [the docs](https://jestjs.io/docs/en/asynchronous), it seems that resolving the promise with an error string will not fail the test. This should be `reject(err)`, where line 36 would change to `return new Promise((resolve, reject) => {`.

packages/fxa-payments-server/server/lib/routes/post-metrics.test.js
```diff
@@ -31,27 +31,37 @@ const invalidBody = {
...
+  test('POST invalid input should return 400', () => {
+    return new Promise(done => {
+      request(app)
+        .post('/metrics')
+        .send(invalidBody)
+        .set('Accept', 'application/json')
+        .expect(400)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
```
6a68: Same issue here, I think we want to `reject` the promise if `err` is truthy.

packages/fxa-profile-server/package.json
```diff
@@ -40,9 +41,9 @@
   "devDependencies": {
     "audit-filter": "^0.5.0",
     "commander": "2.9.0",
-    "eslint-config-prettier": "^5.0.0",
-    "eslint-plugin-fxa": "git://github.com/mozilla/eslint-plugin-fxa.git#41504c9dd30e8b52900c15b524946aa0428aef95",
-    "grunt": "^1.0.4",
+    "eslint": "6.6.0",
+    "eslint-plugin-fxa": "2.0.1",
```
6a68: 🎉

packages/fxa-auth-server/test/local/senders/email.js
meandavejustice: I have not, these are test emails. Just removed because of the [no-useless-escape](https://eslint.org/docs/rules/no-useless-escape) rule.

packages/fxa-content-server/app/tests/spec/views/subscriptions_management_redirect.js
meandavejustice: yeah, rules ask for single quotes around strings that aren't actually templates (don't have variables). Guessing this is because they are just converted to regular strings anyway

packages/fxa-content-server/server/.eslintrc
meandavejustice: These were a couple of rules with huge breakages in the server code. I felt it was too invasive of a change for this pr. Though, I don't know that it would even be worth addressing in the future

packages/fxa-customs-server/package.json
meandavejustice: This was an npm advisory, updating this and filed https://github.com/mozilla/fxa/issues/3376

PR #3295, opened 2019-11-10
feat(auth-server): add more backend error data

bbangert:
Because:

* EndpointError has no stack trace, and no information about what
  request was inflight, to what service, with what exact response.

This commit:

* Determines if the error object has EndpointError attributes and if
  so extracts additional data about the underlying backend request
  call, method, path, and full response error data.

Here's what an EndpointError with this PR includes in Sentry:
image


packages/fxa-auth-server/lib/sentry.js
```diff
@@ -88,6 +88,9 @@ async function configureSentry(server, config) {
       beforeSend(event, hint) {
         return filterSentryEvent(event, hint);
       },
+      integrations: [
+        new Sentry.Integrations.LinkedErrors({ key: 'jse_cause' }),
```
bbangert: I hadn't noticed before that the default Sentry LinkedErrors integration looks for `cause`. This updates it to look for the proper cause key for `verror`.

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -131,6 +129,33 @@ async function configureSentry(server, config) {
             return sentryEvent;
           });
           scope.setExtra('exception', exception);
+          const cause = verror.cause(err);
```
shane-tomlinson: Hmm, I find it somewhat distressing that the unit tests for the sentry module are somewhat non-existent: https://github.com/mozilla/fxa/blob/d8b7629291ec6e0c3846c6c8954521cffd145884/packages/fxa-auth-server/test/local/sentry.js#L7

bbangert:
@shane-tomlinson okey, a test has been added that verifies the error and nested error is extracted ok.

packages/fxa-auth-server/test/local/sentry.js
```diff
@@ -5,30 +5,81 @@
...
+      statusCode: 500,
+    };
+    const sentryCaptureSpy = sinon.stub(Sentry, 'captureException');
+    const scopeContextSpy = sinon.fake();
+    const scopeSpy = {
+      addEventProcessor: sinon.fake(),
+      setContext: scopeContextSpy,
+      setExtra: sinon.fake(),
+    };
+    sinon.replace(Sentry, 'withScope', fn => fn(scopeSpy));
```
shane-tomlinson: sinon's fakes, TIL! https://sinonjs.org/releases/latest/fakes/

PR #3311, opened 2019-11-11
refactor(event-broker): update deps, TS 3.7

bbangert:
Because:

* Multiple dependencies were out of date.
* TS 3.7 has some handy optional chaining.

This commit:

* Updates the out of date dependencies.
* Uses TS 3.7 optional chaining were applicable.


packages/fxa-event-broker/lib/db/firestore.ts
```diff
@@ -43,7 +43,7 @@ class FirestoreDatastore implements Datastore {
     const doc = await document.get();
     if (doc.exists) {
       const data = doc.data();
-      if (data && data.oauth_clients && data.oauth_clients[clientId]) {
```
shane-tomlinson: This is so much nicer.

PR #3314, opened 2019-11-11
fix(metrics): Restore high volume oauth events from amplitude logging

jbuck:


jbuck:
This should be added to train-150 as a point release once it's been merged

clouserw:
@meandavejustice sounds like this will need to be on a point release to stage

dannycoates:
This should have also updated https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/oauth/metrics/amplitude.js

PR #3319, opened 2019-11-12
fix(email): Ensure the action buttons are translated

### LINKED ISSUE #3258: Several strings are not localized
While using FxA in French, I’ve noticed a few strings that are not localized.

Top screen notification:

Capture d’écran 2019-11-06 à 14 18 47

CTA button in multiple emails, after creating an account and when adding a secondary email then making it primary:

Screenshot_2019-11-06 (496) Boîte de réception theochevalier pm me ProtonMail(1)
Screenshot_2019-11-06 (496) Boîte de réception theochevalier pm me ProtonMail(2)
Screenshot_2019-11-06 (497) Boîte de réception theochevalier pm me ProtonMail
Screenshot_2019-11-06 (497) Boîte de réception theochevalier pm me ProtonMail(1)
Screenshot_2019-11-06 Rechercher theochevalier pm me ProtonMail

shane-tomlinson:
The action button text was not being translated. The
action button text was wrapped in `gettext`, but `gettext`
on its own is only for string extraction. To translate,
the text needs to be run through `translator.gettext`
rather than just gettext.

I opted to do this in the `localize` method so
that translation could be centralized.

fixes #3258

@mozilla/fxa-devs - r? Should we try to get this onto train-150?

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -282,14 +282,19 @@ module.exports = function(log, config, oauthdb) {
   Mailer.prototype.localize = function(message) {
     const translator = this.translator(message.acceptLanguage);
 
+    const templateValues = {
+      ...message.templateValues,
+      action:
+        message.templateValues.action &&
+        translator.gettext(message.templateValues.action),
+      language: translator.language,
+      translator,
+    };
+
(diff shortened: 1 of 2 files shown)


davismtl:
OMG yes please. I would have never guessed that this was not localized.

PR #3320, opened 2019-11-12
chore(email-first): Remove legacy signin/signup code

shane-tomlinson:
Builds on #3209


shane-tomlinson:
Closing this, going to further break it down.

PR #3329, opened 2019-11-12
fix(style): Update payments product avatar/logo style to prevent overlap

### LINKED ISSUE #3325: (payments) Hoisted avatar off-center from FF logo
This is regression from #3268 I did not see we needed to account for.

This is kind of done in a hacky way but we can refactor during redesign next quarter, for now I'm going to grab this and just adjust the `margin-top` that's already set.

image

LZoog:
fixes #3325 

Not a perfect solution, but a little less hacky since we're no longer overlapping the image. We can make these consistent with the settings redesign next quarter. If we need to update this padding again, the avatar in payments will still look good since it's now a variable.

image

image

### DIFF
--- a/packages/fxa-content-server/app/styles/_layout.scss
+++ b/packages/fxa-content-server/app/styles/_layout.scss
@@ -28,13 +28,13 @@
     border-radius: $big-border-radius;
     box-shadow: $card-box-high;
     margin: 0 auto;
-    padding: 70px 40px 40px 40px;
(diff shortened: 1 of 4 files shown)


LZoog:
@vbudhram thanks for the review, I changed my comment to reference a new issue I created and will merge when tests pass. 👍

Although, I believe the current settings redesign epic is for design and not engineering. I believe we're getting a new one in Q1?

johngruen:
@clouserw we should uplift this...the underlying error should not hit prod

PR #3350, opened 2019-11-13
chore(email-first): Remove legacy signin/signup views.

shane-tomlinson:
The legacy signin/signup views were still in the repo even
though they were not directly accessed.

This removes the legacy signin/signup views and their tests.
To make this possible, much of the sign_in view needed to be
merged with the force_auth view.

extraction from #3320


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3351, opened 2019-11-13
feat(metrics): Allow RPs to submit a generic 'engage' event to amplitude

### LINKED ISSUE #2743: send an “engagement” metrics ping from Monitor
We would like to send an “engagement” ping from Monitor if our visitor came from one of our breach alert email links.

We’d like to be able to attribute this valuable engagement that our subscribers have with Monitor, even if they don’t sign in to the site.

We could definitely include (Monitor) uids in the email links sending them back to Monitor, and then look up their FxA uid to include in the ping.

cc @irrationalagent 

_edit: related to https://github.com/mozilla/blurts-server/issues/1191_

6a68:
### Commit message / issue overview

With this change, relying parties can submit an 'engage' ping to the
    get-metrics-flow endpoint as a way to record that a particular FxA user
    has interacted in their service in some meaningful way. Only one event is
    possible per RP in this approach, and the meaning of the 'engage' event
    has to be defined out of band.
    
This approach is a bit of a hack, since the 'engage' event for Monitor
    isn't really a flow event, but a ping fired by the Monitor _server_ when a
    user visits the site after clicking a breach alert email link. Also, the
    raw FxA uid is included in the request; longer term, we want to use the
    anonymized uid equivalent being developed as part of ecosystem telemetry
    work.
    
Fixes #2743; see the bug for additional discussion. See also the data
    collection review bug, https://bugzil.la/1594388.

### Testing this patch

1. Disable the [CORS check](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L44) in the `get-metrics-flow` endpoint (for some reason, I wasn't able to get the content server to pick up changes to the local config file's allowed CORS origins list):

```diff
// line 44, fxa-content-server/server/lib/routes/get-metrics-flow.js
     origin: function corsOrigin(origin, callback) {
-      if (ALLOWED_CORS_ORIGINS.includes(origin)) {
+      if (true || ALLOWED_CORS_ORIGINS.includes(origin)) {
```
2. Fire up ye olde server: `npm install && npm start`
3. Tail the content server  logs: `./pm2 logs "content-server PORT 3030"`
4. Open the following URL in a browser: `http://127.0.0.1:3030/metrics-flow?event_type=engage&uid=surprisingly-a-valid-fx-accounts-uid&service=dcdb5ae7add825d2`
- The `service` param value is the local server oauth client ID for 123done
- I'm not able to submit the event via curl, hitting a geodb error. Hopefully this won't happen on the dev server.
5. In the content-server logs, you should see an amplitude event, something like:

fxa-content-server.INFO: amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_rp - engage","time":1573621753075,"user_id":"surprisingly-a-valid-fx-accounts-uid","device_id":"61b2f81d93d74b6cbeffd9b45ff51b04","app_version":"150.0","os_name":"Mac OS X","os_version":"10.14","event_properties":{"service":"123done","oauth_client_id":"dcdb5ae7add825d2"},"user_properties":{"flow_id":"f8f05dcfa5556b366315a8c99f5d3b062100641b347037b384cc1e617a2c22f6","ua_browser":"Firefox","ua_version":"71.0","$append":{"fxa_services_used":"123done"}}}

Pretty-printed, here's the event format. Does this look correct?

```
{
  "op":"amplitudeEvent",
  "event_type":"fxa_rp - engage",
  "time":1573621753075,
  "user_id":"surprisingly-a-valid-fx-accounts-uid",
  "device_id":"61b2f81d93d74b6cbeffd9b45ff51b04",
  "app_version":"150.0",
  "os_name":"Mac OS X",
  "os_version":"10.14",
  "event_properties":{
    "service":"123done",
    "oauth_client_id":"dcdb5ae7add825d2"
  },
  "user_properties": {
    “flow_id":"f8f05dcfa5556b366315a8c99f5d3b062100641b347037b384cc1e617a2c22f6",
    "ua_browser":"Firefox",
    "ua_version":"71.0",
    "$append":{
      "fxa_services_used":"123done"
    }
  }
}
```

### Open questions

- I’m just firing one "fxa_rp - engage" amplitude ping. I'm not firing a corresponding `beginEvent` ping, because this isn't really a "flow", it's a request sent from the Monitor server to the FxA server. 
  - Do I need to fire the `beginEvent` amplitude ping and flow ping, then the engage ping?
  - Do I need to also fire the flow ping alongside the amplitude ping?
- Technically, any RP listed in the `oauth_client_id_map` could use this to submit an 'engage' ping  without us doing any extra work. Is this good or bad? Do we want to use a separate list of allowed clients to decide who can submit this 'engage' ping?
-  Where are the right spots to document this addition to the get-metrics-flow endpoint?


irrationalagent:
> ### Open questions
> 
>     * I’m just firing one "fxa_rp - engage" amplitude ping. I'm not firing a corresponding `beginEvent` ping, because this isn't really a "flow", it's a request sent from the Monitor server to the FxA server.
>       
>       * Do I need to fire the `beginEvent` amplitude ping and flow ping, then the engage ping?

There's no amplitude event that corresponds to `flow.begin`, so iiuc that should not be a prerequisite to logging any amplitude event.

>       * Do I need to also fire the flow ping alongside the amplitude ping?

If we don't fire it, then (obviously) we won't have the data in the flow events (i.e. redshift and FxA Activity Events in redash). I think this is OK - the consumers of this event won't be accessing it through redash/redshift - they will use amplitude. However, it is kind of awkward that we trigger the amplitude event based on the flow event type `flow.rp.engage` even though we don't actually log that flow event. But, that's the way the metrics work at this point so I guess it is what it is. 

>     * Technically, any RP listed in the `oauth_client_id_map` could use this to submit an 'engage' ping  without us doing any extra work. Is this good or bad? Do we want to use a separate list of allowed clients to decide who can submit this 'engage' ping?

I think its good. We want (known) others to be able to use this, in theory. A separate whitelist seems like extra overhead, and if another RP wants to use this I'm guessing that they will come to us first. 

>     * Where are the right spots to document this addition to the get-metrics-flow endpoint?

Good question, [this doc](https://github.com/mozilla/ecosystem-platform/blob/master/docs/relying-parties/metrics-for-relying-parties.md) maybe?

PR #3359, opened 2019-11-13
fix(tests): Add Fairfield to known cities

vbudhram:
Not associated with an issue, but noticed geodb reports a new city. This updates tests to check for it.


vladikoff:
ah yes Fairfield

PR #3362, opened 2019-11-13
Train 150 | v1.150.2

meandavejustice:


packages/fxa-auth-server/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.150.2
+
+### Bug fixes
+
+* tests: Add Fairfield to known cities (61daf8709)
```
shane-tomlinson: Love those wandering IP addresses.

PR #3375, opened 2019-11-15
chore(payments-server): set payments-server to listen on 0.0.0.0

chenba:
This patch updates payments server to listen on 0.0.0.0 (all
interfaces).  This would allow connections from external machines.

Note that 0.0.0.0 is the default on node and express.  It also what
content server uses.

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -110,7 +110,7 @@ const conf = convict({
   },
   listen: {
     host: {
-      default: '127.0.0.1',
+      default: '0.0.0.0',
       doc: 'The ip address the server should bind',
       env: 'IP_ADDRESS',
       format: 'ipaddress',


lmorchard:
We've got some specific CSP headers and such that depend on 127.0.0.1:3031 exactly - is there a scenario this enables that `IP_ADDRESS` env var doesn't help?

chenba:
> We've got some specific CSP headers and such that depend on 127.0.0.1:3031 exactly

This patch shouldn't break anything relying on 127.0.0.1:3031.

>  is there a scenario this enables that IP_ADDRESS env var doesn't help?

Setting IP_ADDRESS would definitely work.  But I also don't see why the default couldn't be 0.0.0.0.

The scenario is this:
 - machine A is running the services
 - machine B is the dev's main machine, with Firefox
 - the dev runs `socat TCP-LISTEN:3031,fork TCP:[SOME IP ADDR on A]:3031` (and for other services that are accessed from the browser)
 - the dev visits the payments UI in the browser on B with 127.0.0.1:3031/SOME_VALID_PATH

The payments server won't respond as it is only listening on local loopback.

Issue #2999, opened 2019-10-18
Error: Redis WATCH detected a conflicting update


clouserw:
https://sentry.prod.mozaws.net/operations/auth-prod/issues/6489019/

```
Error: Redis WATCH detected a conflicting update
  File "/app/lib/error.js", line 992, in Function.AppError.redisConflict
    return new AppError({
  File "/app/lib/redis.js", line 25, in Object.object.(anonymous function) [as update]
    throw error.redisConflict();
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-460)
┆Issue Number: FXA-460

clouserw:
@dannycoates we're seeing this occur in low levels since the push.  Any thoughts?

rfk:
>   File "/app/lib/redis.js", line 25, in Object.object.(anonymous function) [as update]

AFAICT this only caller of `redis.update()` is in `./lib/db.js`, where we manage some session-token data in redis.  In theory we should only see this conflict if a user tries to perform two concurrent actions that modify sessionToken metadata, e.g. by syncing two different devices at the same instant.

I don't have a theory about what could have changed in the latest release to cause this to spike, and I'm lightly suspicious that it may indicate some concurrency bug in the server code, where the server itself is trying to do concurrent modifications as part of handling a single request.

clouserw:
@vbudhram is adjusting the logging to log a better stack

dannycoates:
I've reproduced this locally. The browser is making two calls to `/certificate/sign` in quick succession. Now I'm trying to figure out what changed to trigger the conflict.

dannycoates:
I also reproduced with v1.147.5

dannycoates:
ima chalk this up to updating our sentry dependency so now fewer errors are getting dropped

rfk:
Could be fallout from recent refactoring in the browser code, I'll cross-link this in the sync decouple channel.

rfk:
> I've reproduced this locally. The browser is making two calls to /certificate/sign in quick succession. 

I haven't been able to reproduce this in latest nightly, and I wonder if we actually perturbed it *away* with the latest refactor to how FxA works in the browser.  @dannycoates are you still able to reproduce?

dannycoates:
fixed by #3278

PR #3385, opened 2019-11-17
feat(redis): ADR proposing redis lua scripts

dannycoates:


docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Negative Consequences
+
+- TBD
+
+## Pros and Cons of the Options
+
+### Continue expanding on fxa-shared/redis
+
+`fxa-shared/redis` uses a connection pool and utilizes the bluebird module disposer feature for transaction isolation and connection management. It currently depends on an older redis module that hasn't been updated in two years. With this option we'll likely switch to ioredis but maintain our connection pool. We'll also need to add new functionality to handle new transactions that the existing `update` function can't handle. We'll need to design a retry mechanism to handle write conflicts because the existing code simply ignores these, which is ok for its current use case, but insufficient for some new cases. Transactional cases will require more code and error handling using this method.
```
vbudhram: From https://github.com/mozilla/fxa/issues/2999, when an error is thrown in redis they don't bubble up the stacktrace nicely. Do lau scripts provide additional debug information or are they easier to debug?

dannycoates: Good question. I know that lua scripts are no _worse_ for debugging. While working on the prototype I was able to see stack traces down to the js function that made the call to redis, like `pruneSessionTokens` in the logs. The error _message_ from lua is there too, but not a stacktrace.

For debugging individual lua scripts during dev [redis-cli --ldb](https://redis.io/topics/ldb) worked well.

6a68: I actually wonder if eliminating the indirection of the connection pool and promisified wrapper might fix stack traces.

docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,69 @@
...
+  - Net fewer lines of code
+  - Synchronous transactions
+  - Simpler transaction code
+  - Eliminates write conflicts
+  - More performant
+  - Uses native Promises
+- Cons
+  - Lua isn't js. It's an additional language to deal with.
+  - Sharded clustering requires more care because all keys used by a script must reside on the same node. _We don't currently use sharding._
+
```
6a68: Another 'con': a slow Lua script could hurt performance. I doubt we'll do anything too involved, but good to mention.

docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,69 @@
...
+  - WATCH/MULTI/EXEC redis commands are cumbersome in an async language like js, requiring a connection pool for transaction isolation and multiple redis round trips, adding complexity and maintenance burden.
+  - Currently depends on an outdated redis node module
+  - Depends on bluebird specific features
+  - The `update` function isn't well suited for new transactional operations
+  - New code will need to be added for other transactions
+  - Write conflicts will still occur and handling them will get more complex
+
+### Use lua scripts
+
+Lua scripts offer some benefits over WATCH/MULTI. They execute atomically, can do any number of reads/writes, and accept arbitrary arguments. What would take multiple round trips to read, modify, and write using js and standard redis commands can be executed in a single EVAL call. With this option we'll eliminate the connection pool and our wrapper will require less code.
```
6a68: I'd also add that there's a fully-featured Lua debugger included with Redis, and it has good docs, https://redis.io/topics/ldb.

docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,69 @@
+# Redis Lua Scripts
+
+- Status: draft
+- Deciders: dannycoates + TBD
+- Date: 2019-11-15
+
+## Context and Problem Statement
+
+We're planning on expanding our use of Redis to replace some or all of our MySQL database for OAuth. Our `fxa-shared/redis` module also requires maintenance as part of our modernization effort. The expansion will include use cases that require new indexing, join-like, and transactional operations.
```
bbangert: It would be good to include a more thorough background for what problems moving the data from MySQL to Redis will help solve and how we got to the decision to use Redis more. There's a lot of implied contextual knowledge that is skipped here, which would be great to capture. e.g. that our `fxa-shared/redis` module is currently capturing application logic that would be better handled in Redis.

docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Decision Drivers
+
+- Need for robust redis transactions
+- Increased use of redis for oauth
+- Updating module dependencies
```
bbangert: Indicating what's wrong with the existing fxa-shared redis module would be good to add to the Context.

docs/adr/0008-redis-lua-scripts.md
bbangert: Probably also want to cite that there's some question about how we version/upgrade Lua scripts and the code that interacts with them. Like with stored procedures, change can be tricky.

chenba:
👍

docs/adr/0008-redis-lua-scripts.md
dannycoates: That MySQL to Redis probably deserves its own ADR. This one is specifically about switching to lua scripts for transactional ops. 
... 
_wait, did i just create more work for myself_🤦‍♂️

docs/adr/0008-redis-lua-scripts.md
dannycoates: I'll note it but a lua script change needs no additional considerations over a js change would. There's no prestaging, function name changing, or other stored proc shenanigans.

docs/adr/0008-redis-lua-scripts.md
bbangert: Ah, okey, good point. Then maybe just background on what application logic or issues we have now that the Lua scripts will resolve.

docs/adr/0008-redis-lua-scripts.md
bbangert: Oh, I was under the impression that the Lua script needed to be loaded in there, then it could be executed later. If the whole script is sent over on each call, then yea, that wouldn't apply.

docs/adr/0008-redis-lua-scripts.md
```diff
@@ -0,0 +1,75 @@
+# Redis Lua Scripts
+
+- Deciders: dannycoates, et. al
```
bbangert: This becomes somewhat circular, but anyone reviewing this should probably be added as a decider.

docs/adr/0008-redis-lua-scripts.md
bbangert: There's `verror` in use now as well which should help with the stack traces.

Issue #2411, opened 2019-09-05
oauth-server doesn't send CORS headers to requests with no `Origin` header


rfk:
:pauljt and I discovered this while testing the proxy addon. In latest Nightly, CORS requests from a webextension do not include an `Origin` header. This breaks attempts by webextensions to talk to the oauth-server API, which gives a 404 to the CORS preflight request:

```
>>> # This works OK
>>> requests.request("OPTIONS", "https://oauth.accounts.firefox.com/v1/token", headers={
  "Origin": "https://example.com",
  "Access-Control-Request-Method": "POST"
})
...
<Response [200]>
>>>
>>> # This gives an error 
>>> requests.request("OPTIONS", "https://oauth.accounts.firefox.com/v1/token", headers={
    "Access-Control-Request-Method": "POST"
})
...
<Response [404]>                                                                                                                                                    >>>
```

I don't know enough about CORS to know whether this behaviour is expected or desired.

rfk:
This appears to be deliberate behaviour from our web framework:

  https://github.com/hapijs/hapi/blob/10d30d0b3c6b7a497a85f1be938cf7157f90566f/lib/cors.js#L115

rfk:
We currently specify `cors: true` in the Hapi server definition. This will accept CORS requests from any origin, but try to send an `Access-Control-Allow-Origin` response header reflecting the incoming value from `Origin`, which obviously won't work if there is no `Origin` header.

IIUC from the code and docs, we could instead specify `cors: { origin: 'ignore' }` to accept requests from any origin without sending an `Access-Control-Allow-Origin` header. This should let the requests being sent from current Nightly through successfully.

However, I don't want to rush into that sort of a change without understanding why the current behaviour has changed in Nightly, and what other implications there might be of setting this option.

rfk:
The culprit seems to be https://bugzilla.mozilla.org/show_bug.cgi?id=1405971 in which `Origin` headers were stripped from webextension requests to avoid fingerprinting.

rfk:
Reading around a bit, I think setting `cors: { origin: 'ignore' }` will be an acceptable solution here.

It will cause us to return `Access-Control-Allow-Origin: *` in the response, telling the browser that any origin is allowed to make requests to this endpoint. Which is true.

That said, I expect we'll be far from the only site that expects CORS pre-flight requests to contain an `Origin` header, so I don't know if the current behaviour in nightly will stick.

clouserw:
[...]

PR #3386, opened 2019-11-18
feat(cors): Ignore `Origin` header if all origins are allowed.

rfk:
Because:

* Upcoming changes to Firefox may result in webextensions sending
  CORS pre-flight requests that do not have an `Origin` header.

This commit:

* Uses Hapi's special 'ignore' setting when our config says to allow
  any origin, which is more forgiving of missing `Origin` headers
  than the default configuration.

Connects to #2411


dannycoates:
fyi auth-server in prod _doesn't_ use `*` for `corsOrigin`. #2411 is about oauth only and not other auth-server routes, right? I'll be PRing a patch to use `origin: 'ignore'` for oauth routes shortly.

rfk:
> oauth only and not other auth-server routes, right?

Correct.

PR #3387, opened 2019-11-18
chore(email-first): Remove email-first related experiment code

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #3388, opened 2019-11-18
task(sentry): add sourcemaps to sentry for content server

meandavejustice:
- refs #2345


packages/fxa-content-server/app/scripts/lib/sentry.js
```diff
@@ -4,7 +4,7 @@
 
 import _ from 'underscore';
 import Logger from './logger';
-import Raven from 'raven';
+import * as Sentry from '@sentry/browser';
```
shane-tomlinson: Not possible to skip the `* as Sentry` here and just use `Sentry`?

packages/fxa-content-server/app/scripts/lib/sentry.js
```diff
@@ -76,31 +69,6 @@ function beforeSend(data) {
...
- */
-function shouldSendCallback(data) {
-  if (data && data.message) {
-    var sameError = data.message === PREVIOUS_ERROR;
-    if (sameError) {
-      SAME_ERROR_ATTEMPTS++;
-    } else {
-      SAME_ERROR_ATTEMPTS = 0;
-    }
-    if (SAME_ERROR_LIMIT === SAME_ERROR_ATTEMPTS) {
```
shane-tomlinson: No need for this sort of limiting anymore?

packages/fxa-content-server/package.json
```diff
@@ -38,6 +38,7 @@
         "@babel/preset-env": "7.6.2",
         "@babel/preset-react": "^7.0.0",
         "@babel/preset-typescript": "^7.3.3",
+        "@sentry/browser": "5.6.3",
```
shane-tomlinson: Do [`raven` and `raven-js`](https://github.com/mozilla/fxa/pull/3388/files#diff-57ab2cfbf986be5762b650f526ffb546R112) need to be removed?

packages/fxa-content-server/app/scripts/lib/sentry.js
meandavejustice: This was an option specific to raven, removed on line 184

packages/fxa-content-server/app/scripts/lib/sentry.js
meandavejustice: This is because of the way they have packaged `@sentry/browser`. There isn't a function named `Sentry` only a bunch of methods. So we need to do `* as Sentry` to get the Object to call the methods from. It's also recommended in their docs

packages/fxa-content-server/package.json
meandavejustice: raven-js does need to be removed, I also need to remove it in the webpack config. I didn't touch the places that are using raven directly as this issue was only to get sourcemaps working on the client code. I'd like to completely remove it in the future, but it will be a tad more refactoring, so I'd like to get this in so we can do our sentry triage this wednesday, and follow up with the rest of raven removal in a follow up bug.

I also have a couple references to raven-js in a couple of tests that should be removed before this is merged. Thanks for catching this :)

packages/fxa-content-server/app/tests/spec/lib/channels/receivers/web-channel.js
```diff
@@ -178,22 +177,20 @@ describe('lib/channels/receivers/web-channel', () => {
 
     beforeEach(() => {
       sandbox = sinon.sandbox.create();
-      sandbox.spy(Raven, 'captureMessage');
```
meandavejustice: Removed instead of replacing with the `@sentry/browser` package here because we aren't able to spy on es6 modules as they are read only. I think the way to test this properly would be to spy on `/app/scripts/lib/sentry.js` once it is used in `web-channel.js`. Filed https://github.com/mozilla/fxa/issues/3419

packages/fxa-content-server/app/scripts/lib/sentry.js
```diff
@@ -200,34 +148,18 @@ SentryMetrics.prototype = {
 
     this._exceptionTags.forEach(function(tagName) {
       if (tagName in err) {
-        tags[tagName] = err[tagName];
+        Sentry.configureScope(function(scope) {
```
bbangert: Does this work? I thought the exception had to be reported within this block.

meandavejustice:
@bbangert would you mind taking a look at the test removals in this pr, only removed things I had to in order to get it passing

packages/fxa-content-server/app/scripts/lib/channels/receivers/web-channel.js
```diff
@@ -88,7 +87,7 @@ _.extend(WebChannelReceiver.prototype, Backbone.Events, {
    */
   _reportError(error) {
     this._logger.error('WebChannel error:', error.message);
-    Raven.captureMessage('WebChannel error: ' + error.message, {
+    Sentry.captureMessage('WebChannel error: ' + error.message, {
```
bbangert: It'd be great to know why this wasn't caught as a full stack trace, but I understand this is mainly an update.

packages/fxa-content-server/app/scripts/lib/sentry.js
bbangert: It's the same thing with Sentry in TypeScript as well, it always needs the `*` import.

packages/fxa-content-server/app/scripts/lib/sentry.js
```diff
@@ -135,99 +103,60 @@ function cleanUpQueryParam(url = '') {
...
-      if (tagName in err) {
-        tags[tagName] = err[tagName];
-      }
+    Sentry.withScope(function(scope) {
+      this._exceptionTags.forEach(function(tagName) {
+        if (tagName in err) {
+          scope.setTag(tagName, err[tagName]);
+        }
+      });
+      Sentry.captureException(err);
```
bbangert: 👍

shane-tomlinson:
Removed in https://github.com/mozilla/fxa/pull/3521, bug to re-enable it in https://github.com/mozilla/fxa/issues/3534.

PR #3393, opened 2019-11-18
fix(metrics): Use the correct uid validator in get-metrics-flow

### LINKED ISSUE #2743: send an “engagement” metrics ping from Monitor

6a68:
The first implementation used the wrong validator: unique user ID
instead of user ID. 

Fixes #2743.

### Testing this patch

This patch should allow any registered RP to submit an "engage" event including a valid uid and a registered oauth client ID as 'service' param. The 'Origin' header needs to be set to one of the registered RP domains, too. This should work locally:

`]$ curl -H "Origin: http://127.0.0.1:8001" "http://127.0.0.1:3030/metrics-flow?event_type=engage&service=dcdb5ae7add825d2&uid=b01dfacec01dfacebeeffacebeef1ace"`

You should see the following 2 log lines in the content-server logs after submitting this curl ping:

```
fxa-content-server.INFO: amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_rp - engage","time":1574101610963,"user_id":"b01dfacec01dfacebeeffacebeef1ace","device_id":"cd4241c0c44e4feba49191c252883f0f","app_version":"150.1","event_properties":{"service":"123done","oauth_client_id":"dcdb5ae7add825d2"},"user_properties":{"flow_id":"fc24eb26e2e44b779a8486388c5d1800a8704cd2e3dbf8583abd259eee7f315c","ua_browser":"curl","ua_version":"7.54.0","$append":{"fxa_services_used":"123done"}}}
fxa-content-server.server.requests.INFO: route GET /metrics-flow?event_type=engage&service=dcdb5ae7add825d2&uid=b01dfacec01dfacebeeffacebeef1ace 1.252 200
```

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js
+++ b/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js
@@ -27,10 +27,10 @@ const {
   FORM_TYPE: FORM_TYPE_PATTERN,
   PRODUCT_ID: PRODUCT_ID_PATTERN,
   SERVICE: SERVICE_PATTERN,
-  UNIQUE_USER_ID: UID_PATTERN,
 } = validation.PATTERNS;
 
 const {
+  HEX32: FXA_UID_TYPE,
   STRING: STRING_TYPE,
   UTM: UTM_TYPE,
   UTM_CAMPAIGN: UTM_CAMPAIGN_TYPE,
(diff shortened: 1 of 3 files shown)


6a68:
Thanks for the reviews! Sure, I can target train-150.

PR #3397, opened 2019-11-18
chore(deps): Remove stale nsp exceptions from .nsprc files

6a68:
Just a quick bit of cleanup; many of the .nsprc exceptions have been fixed by updating dependencies. The fxa-geodb fix involved transitive dependencies, which is why its `package-lock.json` is included.

Testing this PR: `npm run lint:deps` from top-level in the FxA repo should complete without errors.


packages/fxa-auth-server/.nsprc
```diff
@@ -1,9 +1,7 @@
 {
+  "comment_766": "766 is sandbox breakout in sandbox.",
```
chenba: 😆

PR #3408, opened 2019-11-19
fix(circleci): use default size resources for content-server tests

dannycoates:
testing out the run time...


dannycoates:
content-server times are still in the 14m range with medium (today at least)

PR #3413, opened 2019-11-19
feat(auth-server): add tiered subs with subhub

### LINKED ISSUE #3273: Implement new subscription upgrade APIs in auth-server
Subhub will make API changes to support subscription upgrades. We need to update auth-server to expose those APIs to payments-server. This includes tests and whatever local API mocking we need in lieu of a solution for #1082

bbangert:
Because:
    
* We want to support plan upgrades for subscriptions.
* SubHub API now has tiered subscription API's.
    
This commit:
    
* Extends the upgrade API to use the new SubHub API.
    
Closes #3273

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -286,9 +286,6 @@ const createRoutes = (
       },
       handler: async function(request) {
         log.begin('subscriptions.updateSubscription', request);
-        if (!stripeHelper) {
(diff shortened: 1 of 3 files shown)


bbangert:
Note that this is currently based on #3343, and I will rebase this after it is merged.

bbangert:
This is actually missing an important bit..... it doesn't update the accountSubscriptions! Adding that now.

bbangert:
@lmorchard updated with better error handling, properly updating the database, etc. re-requesting review as it changed a bit.

Issue #3414, opened 2019-11-19
did a graphicsmagick update break profile-server?


dannycoates:
working with 1.3.31-r0 -- https://circleci.com/gh/mozilla/fxa/115808
broken with 1.3.32-r0 -- https://circleci.com/gh/mozilla/fxa/116318

PR #3417, opened 2019-11-19
fix(profile-server): revert graphicsmagick to 1.3.31

dannycoates:
Unit tests are failing with 1.3.32

maybe this mitigates #3414


jrgm:
Coincidentally, before putting profile server v1.150.4 live, this was failing to process images, so it didn't go live for train-150

PR #3436, opened 2019-11-20
feat(signup): fix #3197 - Autocomplete common email domains during signup for better UX

### LINKED ISSUE #3197: Autocomplete Domain of Email During Registration
80% of registrations are with an email from one of the following providers (I shorten the list):
- gmail
- outlook
- hotmail
- yahoo
- qq
- web.de
- aol.com
- mail.ru
- icloud
- gmx.de

Let's make registration easier for users (especially on mobile) and propose to auto-complete the email domain.

The LinkedIn mobile app does this pretty well however when I tried to replicate it for a screenshot, it seemed to have a UX bug. Ultimately, using my locale, it proposed in the drop-down the most likely domains for my email. We may not need to localize but we could propose to auto-complete as people type of provide options in a drop-down.

Screenshot 2019-10-30 at 17 57 05

Note: Given we don't track emails in our metrics (and just most common domains... because privacy of course), I can only identify 80% of domains users.

src: https://analytics.amplitude.com/mozilla-corp/chart/new/70x6ky6

LZoog:
fix for #3197

Edit: expected functionality below is still applicable, but see [this comment](https://github.com/mozilla/fxa/pull/3436#issuecomment-557394065) for an update on tested devices/caveats.

---
Tested on Firefox, Chrome, Safari, iOS Firefox & Safari. `datalist` works on Edge according to CanIUse but it'd be nice if someone could see if anything is actually broken there.

Expected functionality:
- autocomplete works, in that if the browser has an e-mail used to sign up in other forms, those suggestions still show.
- when a user types '@' and one character after, domain drop-down autocompletion kicks in. This will use native styles per browser.
image

- if the user backspaces this letter and the ending character is "@", all of the choices show in the drop-down. This functionality actually made browser cross-compatibility easier since I was able to do everything inside a `keyup` event.
- if the user backspaces the "@", the drop-down disappears
- if the user changes the text before the "@", the drop-down shows on one more character or press of the "down" arrow
- autocomplete and `datalist` play nicely and show both options with a horizontal rule between them
- when the native keyboard is used in iOS, the options will show in the keyboard and a down caret will show, giving users the option to scroll through the list if they choose to
image

**one caveat:**: If the user types enough in iOS for the suggestions to kick in, and they choose one, the caret disappears. However, if the user has the suggestions show and then they backspace to before the "@" sign (so if they go from "hello@g" to "hello"), the caret still shows and nothing happens when you click on it. I have tried removing this and haven't found the work around yet. I even tried a bunch of things with CSS, but it doesn't appear to be style-able in iOS.

@6a68 I looked into your suggestion around duplicate emails from what the browser natively suggests with `autocomplete` and from what I've found, that native dropdown is not in the DOM and we can't easily grab values from it. :/

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/partial/email-autocomplete-domains.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/partial/email-autocomplete-domains.mustache
@@ -1,0 +1,7 @@
+<input name="email" type="email" class="email tooltip-below" autofocus placeholder="{{#t}}Email{{/t}}"/>
+<datalist id="autocomplete-domain">
+    {{#options}}
+        <option />
+    {{/options}}
+</datalist>
+<div id="focus-hack" tabindex="-1" aria-hidden="true"></div>
--- a/packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
@@ -1,0 +1,117 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A mixin to autocomplete common domains from `datalist option` `value`s.
+ */
+import Template from 'templates/partial/email-autocomplete-domains.mustache';
+import UserAgentMixin from '../../lib/user-agent-mixin';
+
+export const EMAIL_SELECTOR = 'input[type=email]';
+export const DATALIST_OPTIONS_SELECTOR = '#autocomplete-domain option';
+export const FOCUS_HACK_SELECTOR = '#focus-hack';
+
+export const DOMAINS = [
+  'gmail.com',
+  'outlook.com',
+  'hotmail.com',
+  'yahoo.com',
+  'qq.com',
+  'web.de',
+  'aol.com',
+  'mail.ru',
+  'icloud.com',
+  'gmx.de',
+  't-online.de',
+  'orange.fr',
+  'yandex.ru',
+  'yahoo.fr',
+  'live.com',
+  '163.com',
+  'msn.com',
+  'comcast.net',
+  'hotmail.co.uk',
+  'hotmail.fr',
+];
+
+export default {
+  events: {
+    [`keyup ${EMAIL_SELECTOR}`]: '_toggleDomainAutocomplete',
+  },
+  dependsOn: [UserAgentMixin],
+
+  setInitialContext(context) {
+    const unsafeEmailAutocompleteDomainsHTML = this.renderTemplate(Template, {
+      options: DOMAINS,
+    });
+    context.set({
+      unsafeEmailAutocompleteDomainsHTML,
+    });
+  },
+
+  afterRender() {
+    this.isAndroid = this.getUserAgent().isAndroid();
+    this.isChrome = this.getUserAgent().isChrome();
+    this.isEdge = this.getUserAgent().isEdge();
+    this.input = this.$(EMAIL_SELECTOR);
+    this.datalistOptions = this.$(DATALIST_OPTIONS_SELECTOR);
+    this.focusHack = this.$(FOCUS_HACK_SELECTOR);
+    this.previousUsername = null;
+  },
+
+  /**
+   * A catch-most-browsers hack when datalist options are modified.
+   * Chrome otherwise shows all options when '@' is typed,
+   * Safari immediately goes to the dropdown preventing users
+   * from typing further, and FF will not update the dropdown on `keyup`
+   * and has a known bug with dynamic datalists
+   * https://bugzilla.mozilla.org/show_bug.cgi?id=1474137
+   */
+  _focusHack() {
+    this.focusHack.focus();
+    this.input.focus();
+  },
+
+  _toggleDomainAutocomplete() {
+    const inputValue = this.input.val();
+    const [username] = inputValue.split('@');
+
+    if (
+      inputValue.length >= 2 &&
+      inputValue.includes('@') &&
+      (username !== this.previousUsername || !this.input.attr('list'))
+    ) {
+      // Chrome, Android browsers,a nd Edge show all options when '@' is typed.
+      // Non-Android Chrome is covered in `_focusHack` but if the device is
+      // Android we must wait until there is one additional character.
+      if ((!this.isAndroid && !this.isEdge) || !inputValue.endsWith('@')) {
+        // Trigger _before_ datalist option mods here because Chrome
+        // looks janky if text entered pulls up the autocomplete list, but
+        // NOT in Chrome Android because it _causes_ jankiness
+        if (!this.isAndroid && this.isChrome) {
+          this._focusHack();
+        }
+        // Dynamically add/remove `list` attr because it makes conditionals
+        // easier and Chrome otherwise shows all options on focus
+        this.input.attr('list', 'autocomplete-domain');
+        this.datalistOptions.each((index, option) => {
+          option.setAttribute('value', `${username}@${DOMAINS[index]}`);
+        });
+
+        // Causes jankiness in Chrome Android, needed in other cases
+        if (!this.isAndroid || (this.isAndroid && !this.isChrome)) {
+          this._focusHack();
+        }
+      }
+    } else if (!inputValue.includes('@') && this.input.attr('list')) {
+      this.input.removeAttr('list');
+      this.datalistOptions.each((_, option) => {
+        option.removeAttribute('value');
+      });
+      this._focusHack();
+    }
+
+    this.previousUsername = username;
+  },
+};
--- a/packages/fxa-content-server/app/scripts/lib/user-agent.js
+++ b/packages/fxa-content-server/app/scripts/lib/user-agent.js
@@ -46,12 +46,21 @@ const UserAgent = function(userAgent) {
     },
 
     /**
+     * Check if the browser is Chrome
+     *
+     * @returns {Boolean}
+     */
+    isChrome() {
+      return this.browser.name === 'Chrome';
+    },
+
+    /**
      * Check if the browser is Chrome for Android
      *
      * @returns {Boolean}
      */
     isChromeAndroid() {
-      return this.browser.name === 'Chrome' && this.os.name === 'Android';
+      return this.isChrome() && this.isAndroid();
     },
 
     /**
--- a/packages/fxa-content-server/app/scripts/templates/index.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/index.mustache
@@ -18,7 +18,7 @@
 
     <form novalidate>
       <div class="input-row">
-        <input name="email" type="email" class="email tooltip-below" autofocus placeholder="{{#t}}Email{{/t}}"/>
+        {{{ unsafeEmailAutocompleteDomainsHTML }}}
       </div>
 
       <div class="button-row">
--- a/packages/fxa-content-server/app/scripts/views/index.js
+++ b/packages/fxa-content-server/app/scripts/views/index.js
@@ -21,6 +21,7 @@ import mailcheck from '../lib/mailcheck';
 import ServiceMixin from './mixins/service-mixin';
 import SignedInNotificationMixin from './mixins/signed-in-notification-mixin';
 import SyncSuggestionMixin from './mixins/sync-suggestion-mixin';
+import EmailAutocompleteDomainsMixin from './mixins/email-autocomplete-domains-mixin';
 import Template from 'templates/index.mustache';
 
 const EMAIL_SELECTOR = 'input[type=email]';
@@ -222,6 +223,7 @@ Cocktail.mixin(
   IndexView,
   CachedCredentialsMixin,
   CoppaMixin({}),
+  EmailAutocompleteDomainsMixin,
   TokenCodeExperimentMixin,
   FlowBeginMixin,
   FormPrefillMixin,
--- a/packages/fxa-content-server/app/styles/modules/_input-row.scss
+++ b/packages/fxa-content-server/app/styles/modules/_input-row.scss
@@ -9,6 +9,7 @@
 
   input {
     transition: border-color $short-transition, box-shadow $short-transition;
+    padding: 0 $input-left-right-padding;
 
     &::placeholder {
       color: $input-placeholder-color !important;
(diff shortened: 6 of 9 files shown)


packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,101 @@
...
+  'orange.fr',
+  'yandex.ru',
+  'yahoo.fr',
+  'live.com',
+  '163.com',
+  'msn.com',
+  'comcast.net',
+  'hotmail.co.uk',
+  'hotmail.fr',
+];
```
LZoog: @chenba I know you mentioned in #3197 we should stay synced with what's in `mailcheck.js` but this list seems a little more distinct than what we've got in there. I could move these into that file. 🤷‍♀️

Also, @davismtl, you may want to verify these are what you're wanting.

davismtl: That list seems accurate. Same 20 I had in mind.

We can look into expanding it in the future but it's an excellent start.

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,101 @@
...
+
+    if (
+      inputValue.length >= 2 &&
+      inputValue.includes('@') &&
+      ((this.previousUsername !== null && username !== this.previousUsername) ||
+        !this.input.attr('list'))
+    ) {
+      // also trigger _before_ datalist option mods here because Chrome
+      // looks janky if text entered pulls up the autocomplete list
+      this._focusHack();
```
LZoog: I considered checking for user agent while working on this PR, but this line is the only place I'd need the check (for Chrome) since we actually need `_focusHack()` for other browsers in the other two spots that method is called. 🤷‍♀️

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
LZoog: 👍 The order shown here is also going to be the order the list pulls up in the dropdown for desktop & mobile so I checked that Amplitude link you sent before and went in that order. I know I showed this in the demo yesterday but top 3 are shown in the keyboard in mobile if the user types and then backspaces to the "@" sign.

image

LZoog:
🤨 I can't seem to locally reproduce the functional sign up test failures CI is complaining about, they pass for me. Hopefully another rerun will get them passing.

shane-tomlinson:
> 🤨 I can't seem to locally reproduce the functional sign up test failures CI is complaining about, they pass for me. Hopefully another rerun will get them passing.

I can get this one locally:

> × firefox on linux 4.15.0-1043-aws - signup - signup with invalid email address (26.913s)
    Error: ElementNotVisible - input[type=email] + .tooltip
      at Command.<anonymous> @ tests/functional/lib/helpers.js:177:15
      at Command.<anonymous> @ tests/functional/lib/helpers.js:175:6
      at Command.<anonymous> @ tests/functional/lib/helpers.js:61:10
      at Command.<anonymous> @ tests/functional/lib/helpers.js:75:18
      at Command.<anonymous> @ tests/functional/lib/helpers.js:69:10
      at Test.signup with invalid email address [as test] @ tests/functional/sign_up.js:274:12
      @ src/lib/Test.ts:263:51

It looks like the tooltip location has changed in the DOM so that the tooltip is no longer directly adjacent, but rather is a sibling, so the `+` [in selectors.ENTER_EMAIL.TOOLTIP](https://github.com/mozilla/fxa/blob/e9dbcf5dcb151f4d0ecbaa80db0a20a8bc4f82fd/packages/fxa-content-server/tests/functional/lib/selectors.js#L182) needs changed to `~`.

packages/fxa-content-server/app/scripts/templates/partial/email-autocomplete-domains.mustache
```diff
@@ -0,0 +1,7 @@
+<input name="email" type="email" class="email tooltip-below" autofocus placeholder="{{#t}}Email{{/t}}"/>
+<datalist id="autocomplete-domain">
+    {{#options}}
+        <option />
+    {{/options}}
+    </datalist>
```
shane-tomlinson: nit, can you unindent the closing /datalist line?

LZoog:
> Does this and the work that @chenba is doing need to be isolated from each other? Is this going to be A/B tested? If so, we'll have to hook up some testing infrastructure.

I don't believe we're A/B testing this or setting any metrics up for it.

LZoog:
On an old Android OS (5) with updated FF & Chrome, this functions _fairly_ well. The entire list populates on "@" character enter and there is no scrollable list, but that's probably because of the old OS since `datalist` displays based on it. In FF, the suggestions don't disappear when the user backspaces before "@". I wonder if this is fixed with an updated OS. Neither of these minor things seem like blockers to get this in Train 151 either way.

(I'm trying to upgrade to Android 9 now)

davismtl:
While we may not A/B test this. I'd like for us to see if there is anything we should be measuring. Check with @irrationalagent . He worked with Barry yesterday to identify metrics for the mx lookup. I suspect there may be some similarities.

irrationalagent:
Is the main risk here that users have multiple options in the drop down, "slip up" and select the wrong autocomplete option, don't notice, and then have the email go to the wrong address? I doubt that this would offset any gains caused by the increased convenience, but i guess its possible in theory?

These are the ways I see this potentially affecting our metrics (at the minimum, there may be more):
1. If it makes filling in the email field more convenient, the submission rate of the email first form should go up.
2. If users make net more mistakes by selecting the wrong autocomplete option (and don't correct it), our email confirmation rate could go down. 

iiuc, the mx lookup work will also be affecting similar metrics. absent an a/b test, if both features would go out on the same train, then it would make it hard to disentangle their effects.

thoughts? @davismtl

LZoog:
Is the MX lookup going in Train 151?

I finally got setup to test on Android with an updated OS, and am working on a small Android-specific tweak now. This should be ready for merge tomorrow morning pending an r+ from the team, so let me know if we should hold.

davismtl:
Thanks @irrationalagent 
I suspect there is nothing further to measure.

> Is the MX lookup going in Train 151?
Correct. They are both going out with next train.

LZoog:
I did a whole bunch-o testing in Android today and my work needed a little bit of refactoring for a consistent experience. I've been plentiful in commenting because I think otherwise some of what I've got would just be a headscratcher. Some of the `if` statements are also a little more explicit than it perhaps I need to be but I think it makes it a little more readable.

This works well in at least:
- desktop FF, Safari, & Chrome
- iOS (latest) FF, Safari, & Chrome
- Android OS9 FF & Chrome

I'm probably going to download Edge for VB tomorrow morning to verify nothing is broken there, but I would like to get this merged in tomorrow morning for Train 151.

@mozilla/fxa-devs r?

Possible future micro-improvement:
In iOS, when the user types the “@“ sign with a character after, a drop-down caret appears. When the user clicks on the caret it shows a scrollable list of options. However, if the user backspaces and the “@“ sign is not present, 1) the caret does not disappear (and does nothing when tapped), and 2) predictive in-keyboard text remains what it was until the user adds the “@“ sign with one more character again.

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * A catch-most-browsers hack when datalist options are modified.
+   * Chrome otherwise shows all options when '@' is typed,
+   * Safari immediately goes to the dropdown preventing users
+   * from typing further, and FF will not update the dropdown on `keyup`
+   * and has a known bug with dynamic datalists
+   * https://bugzilla.mozilla.org/show_bug.cgi?id=1474137
+   */
+  _focusHack() {
+    this.focusHack.focus();
```
shane-tomlinson: So focus an element after the datalist then focus the input element again. Will screen readers be happy with this?

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,124 @@
...
+      ((this.previousUsername !== null && username !== this.previousUsername) ||
+        !this.input.attr('list'))
+    ) {
+      const { isAndroid, isChrome } = this.getContext();
+      // Chrome and Android browsers show all options when '@' is typed.
+      // Non-Android Chrome is covered in `_focusHack` but if the device is
+      // Android we must wait until there is one additional character.
+      if (
+        !isAndroid ||
+        (isAndroid && inputValue.charAt(inputValue.length - 1) !== '@')
```
shane-tomlinson: You don't really need the `isAndroid &&` here, because if the first condition isn't met, `isAndroid` is guaranteed to be truthy. It also looks like `inputValue.charAt(inputValue.length - 1) !== '@'` could be simplified by using [String.prototype.endsWith](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith).

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,124 @@
...
+  '163.com',
+  'msn.com',
+  'comcast.net',
+  'hotmail.co.uk',
+  'hotmail.fr',
+];
+
+export default {
+  events: {
+    [`keyup ${EMAIL_SELECTOR}`]: '_toggleDomainAutocomplete',
```
shane-tomlinson: Have you tested this with Firefox's username/password fill feature to make sure they interact well together? With the normal form fill, I see a little bit of jankiness when typing/erasing the `@`, but nothing to block on:

flashy-email-dropdown

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,124 @@
...
+  setInitialContext(context) {
+    const isAndroid = this.getUserAgent().isAndroid();
+    const isChrome = this.getUserAgent().isChrome();
+
+    const unsafeEmailAutocompleteDomainsHTML = this.renderTemplate(Template, {
+      options: DOMAINS,
+    });
+    context.set({
+      unsafeEmailAutocompleteDomainsHTML,
+      isAndroid,
```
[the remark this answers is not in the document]
LZoog: That would mean I would need to recalculate the userAgent on every single `keyup`. I'm sure it's not expensive but I thought it'd be a small micro-optimization. Maybe in `afterRender` instead if doing it in `setContext` is strange?

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
LZoog: I actually tested this in VoiceOver on Mac on FireFox to make sure it didn't cause any mishaps and it doesn't seem to complain. Granted I didn't test VO on other browsers. 😕

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
```diff
@@ -0,0 +1,124 @@
...
+  },
+
+  _toggleDomainAutocomplete() {
+    const inputValue = this.input.val();
+    const username = inputValue.substr(0, inputValue.indexOf('@'));
+
+    if (
+      inputValue.length >= 2 &&
+      inputValue.includes('@') &&
+      ((this.previousUsername !== null && username !== this.previousUsername) ||
```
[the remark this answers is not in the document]
LZoog: Hmm, I think I may be able to take it out. I'm pretty sure I needed it when I was first sussing this out for a weird case when the user enters an email, clicks "change email," and the field is already populated. I'll double check everything is good without it.

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
LZoog: Just when you think you've tested a ton of edge cases😂 It looks like when there's an autocomplete list along with the `datalist` set and you play with backspacing/reenter the "@" sign, there's that jankiness. I'm seeing it on Chrome too, it's got to be around the `focusHack` firing when it probably shouldn't be.

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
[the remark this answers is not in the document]
shane-tomlinson: Yeah that'd make more sense. context is stuff that's used in the templates. Not sure how the name came about, but we've stuck with it.

packages/fxa-content-server/app/scripts/views/mixins/email-autocomplete-domains-mixin.js
shane-tomlinson: This isn't a blocker, I wouldn't spend too much time on it.

LZoog: @shane-tomlinson I'm going to ask QA to vet this thing and file an issue with all the findings. At the very least, we should have a list of all of any awkwardness/edge cases like this and determine what we can fix or what should be changed. I can't find a quick solution for this one in particular because the autocomplete drop-down list is not in the DOM.

Other changes have been pushed.

LZoog: (I am, however, going to test in Edge real quick to make sure nothing is broken)

LZoog:
Pushed up something else after checking for Edge compatibility and checked that nothing regressed in FF or Android devices. I think I have all my bases covered. 😅

chenba:
> Is the MX lookup going in Train 151?

Uncertain at the moment since teh PR haven't received a 👍 from Shane yet.

LZoog:
Going to :shipit: since Shane gave an r+ here.

To the QA team:
I have tested this in:
- desktop FF, Safari, & Chrome
- iOS (latest) FF, Safari, & Chrome
- Android OS9 FF & Chrome
- Edge on Windows 10
- VoiceOver for desktop Firefox on Mac

Please create an issue for any jankiness you find - edge cases, accessibility issues, etc. We may not fix all of them, but it will be great to have an issue with a list of anything weird.
Known issues:
- In iOS, when the user types the “@“ sign with a character after, a drop-down caret appears. When the user clicks on the caret it shows a scrollable list of options. However, if the user backspaces and the “@“ sign is not present, 1) the caret does not disappear (and does nothing when tapped), and 2) predictive in-keyboard text remains what it was until the user adds the “@“ sign with one more character again.
- when there's an autocomplete list along with the `datalist` set and you play with backspacing/reenter the "@" sign, there's a little bit of jankiness in at least desktop Chrome & FF
flashy-email-dropdown

PR #3444, opened 2019-11-21
fix(error): keep zip field from erroring while user inputs

### LINKED ISSUE #3168: Over eager invalid handling on zip code input
Focus the zip code and start typing, when a single character is entered, the box is marked `invalid`. The box should only look to see if it's invalid on blur

johngruen:
Fixes #3168

@lmorchard this seems like a simple way to get around the edge case you identified in your comment on that bug.

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -274,15 +274,17 @@ const validateName: OnValidateFunction = (value, focused) => {
 };
 
 const validateZip: OnValidateFunction = (value, focused) => {
-  let valid = true;
+  let valid = undefined;
   let error = null;
   value = ('' + value).substr(0, 5);
   if (!value) {
     valid = false;
     error = 'Zip code is required';
-  } else if (value.length !== 5) {
+  } else if (value.length !== 5 && !focused) {
     valid = false;
     error = 'Zip code is too short';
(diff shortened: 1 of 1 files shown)


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -274,15 +274,17 @@ const validateName: OnValidateFunction = (value, focused) => {
 };
 
 const validateZip: OnValidateFunction = (value, focused) => {
-  let valid = true;
+  let valid = null;
```
lmorchard: Looks like a type validation thing failed CI here - this should probably be `let valid = undefined;`

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -274,15 +274,17 @@ const validateName: OnValidateFunction = (value, focused) => {
...
   let error = null;
   value = ('' + value).substr(0, 5);
   if (!value) {
     valid = false;
     error = 'Zip code is required';
-  } else if (value.length !== 5) {
+  } else if (value.length !== 5 && !focused) {
     valid = false;
     error = 'Zip code is too short';
+  } else if (value.length == 5) {
```
lmorchard: :+1: good thinking! I don't think this exact thing had occurred to me!

PR #3448, opened 2019-11-21
fix(email): update subscription TOS url to be correct

### LINKED ISSUE #3301: Subscription Email PN link is incorrect
It should point to
https://www.mozilla.org/en-US/privacy/firefox-private-network/

johngruen:
Fixes #3301


johngruen:
oh yeah, good catch...i'll do that and merge

lmorchard:
Just a note: This doesn't necessarily change the URL in prod or stage - this PR just changes the local development setting and the config default. The actual URL is a server setting that we'll need to get ops help to change.

PR #3455, opened 2019-11-22
OAuth CORS

dannycoates:
oauth-server used to use '\*' in prod while auth-server
is limited to a couple important domains. In the
oauth->auth merge this fact got overlooked, breaking oauth
for some reliers, like Send when we flipped DNS.
With #2411 we want to expand '*' to 'ignore' anyway.

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/routes/index.js
+++ b/packages/fxa-auth-server/lib/oauth/routes/index.js
@@ -113,6 +113,7 @@ exports.clients = [
 ];
 
 routes.forEach(function(route) {
+  route.config.cors = { origin: 'ignore' };
   var method = route.method.toUpperCase();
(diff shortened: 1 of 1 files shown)


dannycoates:
We need this before we can flip oauth DNS to the auth-server

packages/fxa-auth-server/lib/oauth/routes/index.js
```diff
@@ -113,6 +113,7 @@ exports.clients = [
 ];
 
 routes.forEach(function(route) {
+  route.config.cors = { origin: 'ignore' };
```
rfk: nit: maybe for a followup, but it might be worth checking for any existing `cors` declaration on the route so that we don't mysteriously overwrite someone's future changes to these routes.

PR #3458, opened 2019-11-22
fix(totp): Set TOTP options when instantiated instead of in route

### LINKED ISSUE #3277: Authorization code expired about 30 seconds after being sent
Just hit this in Firefox 70. I'm certain it's been less than a minute between when the code was sent and when I've tried using it.

image

vbudhram:
This hopefully fixes #3277. 

I believe there was a race condition between the two TOTP configurations used. The one used for signup/signin is configured for 10mins with 2 time windows and the 2FA version is configured for 30 seconds and 2 time windows.

I suspect that when the respective routes get created it might pull from the global configuration. At the very least this change does not break anything. I also updated the otplib to version 11.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/totp.js
+++ b/packages/fxa-auth-server/lib/routes/totp.js
@@ -15,13 +15,6 @@ const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema;
 module.exports = (log, db, mailer, customs, config) => {
   const otpUtils = require('../../lib/routes/utils/otp')(log, config, db);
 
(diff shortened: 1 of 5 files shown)


vbudhram:
@mozilla/fxa-devs calling this ready!

shane-tomlinson:
Does this need to be uplifted to train-150?

vbudhram:
> Does this need to be uplifted to train-150?

The issue seemed rare to me? But I don't see the harm to uplift if we haven't tag 151 yet.

Issue #3427, opened 2019-11-20
Can't log in after changing primary email


ryanfeeley:
## Description

Though Sync UI updated with the new primary email, when I installed the Firefox Notes extension, and tried syncing Notes, I was unable to log in at all.

### Steps to reproduce

1. Sign in to Firefox/Sync in desktop
1. Add secondary email
1. Make secondary email the primary email
1. Confirm new email address in Firefox/Sync UI (toolbar, prefs, etc)
1. Install Firefox Notes and select **Sign in to Sync**

### Expected result
* New primary email is shown so user doesn't have to retype email, password accepted

### Actual result
* Can't log in
![not-alright-alright-alright](https://user-images.githubusercontent.com/68213/69268186-a9375180-0b9c-11ea-987b-670850da761e.png)

### Environment

Nightly Mac

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-559)
┆Issue Number: FXA-559

vladikoff:
We should clear the locally cached email after emails have been switched

davismtl:
Based on Slack conversation, this is happening with most services.
This is pretty major if it's the case.

vbudhram:
> Can't log in

@ryanfeeley If you click `Use a different account` you are still not able to login? Does the new attempt use the old email?

ryanfeeley:
@vbudhram You are right, clicking use a different account does work, but the bug is pretty severe because no one would think to do that (as the right account is being shown above).

Luckily, if my change modications are correct, retention is still a lot higher for those that change their primary email than those that don’t. They are also different user types too probably.

https://analytics.amplitude.com/mozilla-corp/chart/7y22vf0

davismtl:
For context, here is the volume of users per month adding a secondary email:
https://analytics.amplitude.com/mozilla-corp/chart/new/tk2unio

I suspect this volume will go up with the work @vbudhram is doing this quarter.

clouserw:
Being able to log in with the secondary email is a better user experience, but we're unsure if there are security concerns.  @jvehent are there any concerns with this change?

davismtl:
Talked w/ @vbudhram 
Another simple solution would be to include a link in the error message to the email-first form.
We also suspect that his solution to login with secondary email would be convenient but would not resolve the entire problem so he'll file another bug for what we would want to do longer term. We expect this problem to surface more often as we push users to add a secondary email and to audit their primary emails.

arroway:
[...]

PR #3460, opened 2019-11-22
fix(emails): Allow users to login with a secondary email

vbudhram:
Lets see what breaks 🤔

Connects #3427

### DIFF
--- a/packages/fxa-auth-server/test/remote/recovery_email_emails.js
+++ b/packages/fxa-auth-server/test/remote/recovery_email_emails.js
@@ -821,7 +821,7 @@ describe('remote emails', function() {
     });
   });
 
-  describe("shouldn't be able to login with secondary email", () => {
+  describe('should be able to login with secondary email', () => {
     let secondEmail;
     beforeEach(() => {
       secondEmail = server.uniqueEmail();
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
Oh this is amazing if it works!

packages/fxa-auth-server/test/remote/recovery_email_emails.js
```diff
@@ -851,14 +851,28 @@ describe('remote emails', function() {
...
-    it('fails to login', () => {
+    it('successful login', () => {
       return Client.login(config.publicUrl, secondEmail, password, {})
         .then(() => {
           assert.fail(new Error('should not have been able to login'));
         })
         .catch(err => {
+          // When logging in with a secondary email the auth-server returns the correct email
+          // to generate the authPW (based on OnePW protocol). This is the email that should be
+          // used to login.
```
rfk: IIRC, one reason we haven't done this in the past is that it's a small unauthenticated info leak - if I know your secondary email address, I could do a fake login attempt and have the FxA server tell me your primary address. Have we / do we need to clear this info leak with security/privacy review?

dannycoates:
This is a pretty significant change in behavior. I think we should be careful here and consider other options.

vbudhram:
After speaking with the team and security folks, we are going not going to change this. Currently, not a lot of users experience this issue and therefore it might not have a big impact.

Ref: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/627683/?query=is:unresolved

PR #3466, opened 2019-11-22
feat(cwts): 100% of sync users see CWTS on signup page

shane-tomlinson:
Does not affect the sync optional flow.

@mozilla/fxa-devs - r? We'll want to cherry-pick this into train-150 and 151 too.


packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/cwts-on-signup-password.js
```diff
@@ -10,7 +10,7 @@ const EXPERIMENT_NAME = 'signupPasswordCWTS';
 
 const GROUPS = ['control', 'treatment'];
 
-const ROLLOUT_RATE = 0.2;
+const ROLLOUT_RATE = 1;
```
shane-tomlinson: This works because of [this branch here](https://github.com/mozilla/fxa/pull/3466/files#diff-d6caf564bc7e3f6c366d27554185d1f4R42).

PR #3479, opened 2019-11-25
fix(totp): Use the totp options instead of global options

vbudhram:
Not linked to an issue, but we have had rising rates of TOTP tests failures. This fixes it...hopefully.

cc @shane-tomlinson

### DIFF
--- a/packages/fxa-auth-server/lib/routes/totp.js
+++ b/packages/fxa-auth-server/lib/routes/totp.js
@@ -246,10 +246,18 @@ module.exports = (log, db, mailer, customs, config) => {
         const sharedSecret = token.sharedSecret;
         const tokenVerified = token.verified;
 
+        // Default options for TOTP
+        const otpOptions = {
+          encoding: 'hex',
+          step: config.step,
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -246,10 +246,18 @@ module.exports = (log, db, mailer, customs, config) => {
...
+          encoding: 'hex',
+          step: config.step,
+          window: config.window,
+        };
+
         const authenticator = new otplib.authenticator.Authenticator();
         authenticator.options = Object.assign(
           {},
           otplib.authenticator.options,
+          otpOptions,
```
vbudhram: This explicitly sets the TOTP options just in case the defaults got changed. Not really sure how they would have though.

chenba:
@vbudhram There were some other failures in CI.  I'm re-running the build in case they are just flaky.  🤞

vbudhram:
Bleh, not flaky fails locally. Let me push fix.

PR #3492, opened 2019-11-26
fix(totp): Use a shared helper to check otp codes

vbudhram:
I think I *finally* got the right fix for https://github.com/mozilla/fxa/issues/3277.

In the `/session/verify_code` route, I was using checking otp codes by generating the expected code. This fails sometimes because codes can be valid for multiple time windows and we only get that check when using the `.check` function on the otplib. 

I've extracted this into a shared helper function it make things a little cleaner.

I'm targeting train-151 because I feel confident this is the right fix.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/session.js
+++ b/packages/fxa-auth-server/lib/routes/session.js
@@ -361,9 +361,9 @@ module.exports = function(
         const account = await db.account(sessionToken.uid);
         const secret = account.primaryEmail.emailCode;
 
-        const expectedCode = otpUtils.generateOtpCode(secret, otpOptions);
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -361,9 +361,9 @@ module.exports = function(
         const account = await db.account(sessionToken.uid);
         const secret = account.primaryEmail.emailCode;
 
-        const expectedCode = otpUtils.generateOtpCode(secret, otpOptions);
+        const isValidCode = otpUtils.verifyOtpCode(code, secret, otpOptions);
 
-        if (expectedCode !== code) {
```
vbudhram: 🤦🏽‍♂️

PR #3497, opened 2019-11-26
chore(tests): de-duplicate some sign in/up functional tests

### LINKED ISSUE #1542: Remove functional tests where functionality being tested is tested elsewhere.
As a result of #1337 and the addition of OAuth tests, Circle runs now take longer. We could improve this situation by removing suites such as [sign_in](https://github.com/mirefly/fxa/blob/f9adb55443f29503ad9a0b46d824988f4529e3ff/packages/fxa-content-server/tests/functional/sign_in.js#L35) and [sign_up](https://github.com/mirefly/fxa/blob/f9adb55443f29503ad9a0b46d824988f4529e3ff/packages/fxa-content-server/tests/functional/sign_up.js#L7) whose functionality is largely duplicated elsewhere. Sure, maybe we have to migrate some tests so we don't lose what they test, but there is a lot of duplication.

chenba:
These tests are considered duplicates because:
 - Some are just plain duplicates
 - Some are sub-tests of other tests
 - Some test the same functionalities but with different starting points
   (content server vs. oauth relying party)

Fixes #1542

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/tests/functional/sign_up.js
+++ b/packages/fxa-content-server/tests/functional/sign_up.js
@@ -104,52 +104,6 @@ registerSuite('signup', {
         .then(testAtConfirmScreen(email));
     },
 
-    'signup, verify same browser': function() {
-      return this.remote
-        .then(openPage(ENTER_EMAIL_URL, selectors.ENTER_EMAIL.HEADER))
-        .then(visibleByQSA(selectors.ENTER_EMAIL.LINK_SUGGEST_SYNC))
-        .then(fillOutEmailFirstSignUp(email, PASSWORD))
-        .then(testAtConfirmScreen(email))
-        .then(openVerificationLinkInNewTab(email, 0))
-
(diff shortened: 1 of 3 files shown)


chenba:
This probably won't make much of a dent in our overall CI functional test run time.  😢

packages/fxa-content-server/tests/functional/sign_up.js
```diff
@@ -180,51 +134,6 @@ registerSuite('signup', {
...
-          .then(openVerificationLinkInDifferentBrowser(email))
-
-          // The original tab should transition to the settings page w/ success
-          // message.
-          .then(testElementExists(selectors.SETTINGS.HEADER))
-          .then(testSuccessWasShown())
-      );
-    },
-
-    "signup, verify different browser - from new browser's P.O.V.": function() {
```
shane-tomlinson: Ha! I removed these same tests elsewhere too!

PR #3500, opened 2019-11-27
fix(cwts): Ensure CWTS tests in sync optional flow test reality.

shane-tomlinson:
The tests were using Firefox 57 even though functionality
is only available starting with Firefox 71, and did not test
which screen the user should be at on the end of the flow.

Use Firefox 71 as the user agent string, check the final
screen.

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/do-not-sync-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/do-not-sync-mixin.js
@@ -7,14 +7,14 @@
  *
  * @mixin DoNotSync mixin
  */
-import SigninMixin from './signin-mixin';
 import FlowEventsMixin from './flow-events-mixin';
+import preventDefaultThen from '../decorators/prevent_default_then';
 
 export default {
-  dependsOn: [SigninMixin, FlowEventsMixin],
+  dependsOn: [FlowEventsMixin],
 
   events: {
-    'click #do-not-sync-device': 'doNotSync',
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/views/mixins/do-not-sync-mixin.js
```diff
@@ -7,14 +7,14 @@
  *
  * @mixin DoNotSync mixin
  */
-import SigninMixin from './signin-mixin';
```
shane-tomlinson: SigninMixin didn't seem to be used for anything.

packages/fxa-content-server/app/scripts/views/mixins/do-not-sync-mixin.js
```diff
@@ -7,14 +7,14 @@
...
 import FlowEventsMixin from './flow-events-mixin';
+import preventDefaultThen from '../decorators/prevent_default_then';
 
 export default {
-  dependsOn: [SigninMixin, FlowEventsMixin],
+  dependsOn: [FlowEventsMixin],
 
   events: {
-    'click #do-not-sync-device': 'doNotSync',
+    'click #do-not-sync-device': preventDefaultThen('doNotSync'),
```
shane-tomlinson: `preventDefaultThen` is needed to prevent the browser from pulling the user back to /choose_what_to_sync after the form submit on the /confirm_signup_code screen. Strange.

PR #3508, opened 2019-11-27
feat(service-mixin): fix #2054 - Display 'Account Settings' if `serviceName` is not provided by relier

### LINKED ISSUE #2054: Include Account Settings subheading when signing in to Settings
So that users have the clearest sense of their destination, we should include the service name in subheading. For login, registration, force auth and password reset, we should include *Account Settings* when service equals none (I think it's none…?).

**Current**
Screen Shot 2019-08-02 at 11 46 01 AM

**Proposed**
Screen Shot 2019-08-02 at 11 46 15 AM

LZoog:
fix for #2054 

Creating a new context key (`noRelierServiceDefault`) and checking for it when `serviceName` is not given in the template seemed like the safest approach for this (since "Account Settings" isn't really a _service_, it felt messy to mutate the context if `serviceName` didn't exist to set it as such). It displays the RP `serviceName` over “Account Settings” when one is provided. Let me know if you guys have a better name.

Some screenshots can be seen in the issue.

Since we aren’t checking for `{{#serviceName}}` in `permissions.mustache`, I didn’t update it, as it doesn’t seem this would be shown unless an actual RP is requesting permissions.

~@ryanfeeley one last question for you. In `ready.mustache`, or when the account is verified/ready, do we want the text to read “Your account is ready!” (current) if they were logging into settings, or “You are now ready to use Account Settings”? That one seemed questionable to me, but it could be argued if they've seen "Account Settings" through the flow that it'd be consistent to see it there too.~
Edit: Ryan and I pinged on Slack and he gave the thumbs up for `ready.js`.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/ready.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/ready.mustache
@@ -30,7 +30,9 @@
           {{/emailVerified}}
 
           {{^emailVerified}}
-              <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
+            {{#noRelierServiceDefault}}
+              <p class="account-ready-generic">{{#t}}You are now ready to use %(noRelierServiceDefault)s{{/t}}</p>
+            {{/noRelierServiceDefault}}
           {{/emailVerified}}
       {{/serviceName}}
     {{/isSync}}
--- a/packages/fxa-content-server/app/scripts/templates/reset_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/reset_password.mustache
@@ -1,12 +1,15 @@
 <div id="main-content" class="card">
   <header>
     <h1 id="fxa-reset-password-header">
+      {{#t}}Reset password{{/t}}
+        <!-- L10N: For languages structured like English, this phrase can read "to continue to %(serviceName)s" -->
       {{#serviceName}}
-        <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(serviceName)s" -->
-        {{#t}}Reset password{{/t}} <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
+        <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
       {{/serviceName}}
       {{^serviceName}}
-        {{#t}}Reset password{{/t}}
+        {{#noRelierServiceDefault}}
+          <span class="service">{{#t}}Continue to %(noRelierServiceDefault)s{{/t}}</span>
+        {{/noRelierServiceDefault}}
       {{/serviceName}}
     </h1>
   </header>
--- a/packages/fxa-content-server/app/scripts/templates/account_recovery_confirm_key.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/account_recovery_confirm_key.mustache
@@ -21,12 +21,15 @@
     {{^isLinkExpired}}
         <header>
             <h1 id="fxa-recovery-key-confirm">
+                {{#t}}Reset password with recovery key{{/t}}
+                <!-- L10N: For languages structured like English, this phrase can read "to continue to %(serviceName)s" -->
                 {{#serviceName}}
-                    <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(serviceName)s" -->
-                    {{#t}}Reset password with recovery key{{/t}} <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
+                    <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
                 {{/serviceName}}
                 {{^serviceName}}
-                    {{#t}}Reset password with recovery key{{/t}}
+                    {{#noRelierServiceDefault}}
+                        <span class="service">{{#t}}Continue to %(noRelierServiceDefault)s{{/t}}</span>
+                    {{/noRelierServiceDefault}}
                 {{/serviceName}}
             </h1>
         </header>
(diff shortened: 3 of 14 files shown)


packages/fxa-content-server/app/scripts/templates/ready.mustache
```diff
@@ -30,7 +30,9 @@
           {{/emailVerified}}
 
           {{^emailVerified}}
-              <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
+            {{#noRelierServiceDefault}}
+              <p class="account-ready-generic">{{#t}}You are now ready to use %(noRelierServiceDefault)s{{/t}}</p>
+            {{/noRelierServiceDefault}}
```
LZoog: I pinged Ryan about this change in Slack and he gave it the thumbs up.

TBH we probably do not need to check for `noRelierServiceDefault` if every one of these views is using the `service-mixin.js` We could just assume in our mustache files that `noRelierServiceDefault` is going to be truthy if `serviceName` is falsey but I'm not sure that's obvious unless you pull up the mixin. 🤷‍♀️ I can easily pull the checks out if that's preferred, though.

packages/fxa-content-server/app/scripts/templates/reset_password.mustache
```diff
@@ -1,12 +1,15 @@
 <div id="main-content" class="card">
   <header>
     <h1 id="fxa-reset-password-header">
+      {{#t}}Reset password{{/t}}
```
shane-tomlinson: I really wish I'd have extracted this header stuff into a partial template so that this would only need to be updated in one place. :(

packages/fxa-auth-server/config/dev.json
```diff
@@ -129,7 +129,7 @@
           "product_metadata": {
             "webIconURL": "http://placekitten.com/512/512?image=11",
             "emailIconURL": "http://placekitten.com/512/512?image=5",
-            "downloadURL": "http://getfirefox.com",
+            "downloadURL": "http://127.0.0.1:8080/"
```
LZoog: Ack, I'd pulled this from master and it snuck its way into this commit, I'll see what I can do about it...

packages/fxa-content-server/app/scripts/templates/reset_password.mustache
LZoog: Yeah, I noticed some redundancy as I was working through this. Perhaps it can be tackled with the settings redesign.

LZoog:
@shane-tomlinson do you think I should remove my checks for `{{#noRelierServiceDefault}}` since that should always be truthy or is this g2g? It seemed like the safest approach (no 500 errors if we use these templates without the service-mixin or something weird happens) but eh, after seeing we don't check for `serviceName` before displaying in `permissions.mustache`,  it also seems a bit superfluous.

shane-tomlinson:
> Creating a new context key (noRelierServiceDefault) and checking for it when serviceName is not given in the template seemed like the safest approach for this (since "Account Settings" isn't really a service, it felt messy to mutate the context if serviceName didn't exist to set it as such). It displays the RP serviceName over “Account Settings” when one is provided. Let me know if you guys have a better name.

It might feel icky to modify context that way, but what feels totally acceptable is to set a default `serviceName` within the base relier. [OAuth](https://github.com/mozilla/fxa/blob/0eab8046c79053534710ea41c9eadbe67ce6b4aa/packages/fxa-content-server/app/scripts/models/reliers/oauth.js#L28) and [Sync](https://github.com/mozilla/fxa/blob/0eab8046c79053534710ea41c9eadbe67ce6b4aa/packages/fxa-content-server/app/scripts/models/reliers/browser.js#L60:L65) reliers both set a serviceName, and the only thing the default *should* be used for is when neither of those are used. If we went this route, we'd want to translate the [serviceName here in the service-mixin](https://github.com/mozilla/fxa/blob/0eab8046c79053534710ea41c9eadbe67ce6b4aa/packages/fxa-content-server/app/scripts/views/mixins/service-mixin.js#L17) before printing it to the DOM so users of other locales see the correct value.

There may be some edge cases around verifying with links, e.g., signin with an unverified account or signup verification reminders. Both signup and signin by default now use codes in most places, but there are a couple of stragglers, it'd be worthwhile testing these.

This *feels* simpler to me, but I haven't dug in too much to know!

shane-tomlinson:
> It might feel icky to modify context that way, but what feels totally acceptable is to set a default serviceName within the base relier.

```patch
diff --git a/packages/fxa-content-server/app/scripts/models/reliers/relier.js b/packages/fxa-content-server/app/scripts/models/reliers/relier.js
index 5e4d7e4bb..3f3b7650f 100644
--- a/packages/fxa-content-server/app/scripts/models/reliers/relier.js
+++ b/packages/fxa-content-server/app/scripts/models/reliers/relier.js
@@ -19,6 +19,8 @@ import ResumeTokenMixin from '../mixins/resume-token';
 import UrlMixin from '../mixins/url';
 import Vat from '../../lib/vat';
 
+const t = (msg) => msg;
+
 const RELIER_FIELDS_IN_RESUME_TOKEN = [
   'entrypoint',
   'entrypointExperiment',
@@ -93,6 +95,7 @@ var Relier = BaseRelier.extend({
     migration: null,
     resetPasswordConfirm: true,
     setting: null,
+    serviceName: t('Account Settings'),
     style: null,
     uid: null,
     utmCampaign: null,
```

This *seems* to work, but I haven't tested the verification link flows. Ah yeah, password reset always uses links too.

LZoog:
@shane-tomlinson yep, it felt like I was missing something, that's the way to go. Closing this in favor of #3569

PR #3532, opened 2019-11-29
feat(signin-code): Add bounce handling to /signin_token_code

### LINKED ISSUE #3528: /signin_token_code does not handle bounces
Similar to #3405, /signin_token_code does not handle email bounces.

shane-tomlinson:
If an email bounces when sending a token code, we never knew and
the user was not informed. This sends the user to /signin_bounced.

fixes #3528

@vbudhram - r?

I wasn't sure if there needed to be any additional logic on the back-end.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
@@ -11,6 +11,7 @@ import Constants from '../lib/constants';
 import FormView from './form';
 import Template from 'templates/sign_in_token_code.mustache';
 import ResendMixin from './mixins/resend-mixin';
+import SessionVerificationPollMixin from './mixins/session-verification-poll-mixin';
 
 const CODE_INPUT_SELECTOR = 'input.otp-code';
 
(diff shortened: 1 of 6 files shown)


packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
```diff
@@ -29,6 +30,12 @@ const View = FormView.extend({
     }
   },
 
+  afterVisible() {
+    this.waitForSessionVerification(this.getAccount(), () => {
```
vbudhram: I would probably add a quick comment about why we need to poll for verification here.

vbudhram: Conversely, we could add a new helper function `pollForEmailBounce` or something similar, but meh.

packages/fxa-content-server/tests/functional/sign_in_token_code.js
```diff
@@ -103,6 +105,22 @@ registerSuite('signin token code', {
...
+    'verified - treatment-code - bounce': function() {
+      experimentParams.query.forceExperiment = 'tokenCode';
+      experimentParams.query.forceExperimentGroup = 'treatment-code';
+
+      return this.remote
+        .then(openFxaFromRp('enter-email', experimentParams))
+        .then(fillOutEmailFirstSignIn(email, PASSWORD))
+        .then(testElementExists(selectors.SIGNIN_TOKEN_CODE.HEADER))
+        .then(getStoredAccountByEmail(email))
+        .then(destroySessionForEmail(email))
```
vbudhram: While looking at https://github.com/mozilla/fxa/pull/3533/files, why would this have to use `destroySessionForEmail` verses `accountDestroy`?

shane-tomlinson: A sign in verification email doesn't normally destroy the account, that only happens on signup verifications.

vasilicatamas:
Some reliable STR would be very helpful, if the QA attention is still needed.

PR #3533, opened 2019-11-29
feat(signup-code): Add bounce handling to /confirm_signup_code

### LINKED ISSUE #3405: confirm_signup_code does not handle email bounces
I noticed this today while experimenting with updating the functional tests to use email-codes everywhere.

In the link flow, it's handled [in this mixin](https://github.com/mozilla/fxa/blob/5814416ef26faaa1151cf0abb4d9ffd7ab7add8a/packages/fxa-content-server/app/scripts/views/mixins/session-verification-poll-mixin.js#L83), I wonder if we could hook that up in the confirm_signup_code view too.

cc @vbudhram

shane-tomlinson:
/confirm_signup_code didn't have bounce handling. This changes that.

fixes #3405

@vbudhram - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/confirm_signup_code.js
+++ b/packages/fxa-content-server/app/scripts/views/confirm_signup_code.js
@@ -4,12 +4,13 @@
 
 import _ from 'underscore';
 import AuthErrors from '../lib/auth-errors';
-import Cocktail from 'cocktail';
+import Cocktail from '../lib/cocktail';
 import FlowEventsMixin from './mixins/flow-events-mixin';
 import FormView from './form';
 import ServiceMixin from './mixins/service-mixin';
 import Template from 'templates/confirm_signup_code.mustache';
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/views/confirm_signup_code.js
```diff
@@ -26,9 +27,12 @@ class ConfirmSignupCodeView extends FormView {
     return proto.afterVisible
       .call(this)
       .then(() => this.broker.persistVerificationData(account))
-      .then(() =>
-        this.invokeBrokerMethod('beforeSignUpConfirmationPoll', account)
-      );
+      .then(() => {
+        this.waitForSessionVerification(this.getAccount(), () => {
```
vbudhram: Same comment from other PR, it isn't clear that the `waitForSessionVerification` handles email bounces.

vasilicatamas:
Some reliable STR would be very helpful, if the QA attention is still needed.

PR #3536, opened 2019-11-29
feat(signin-code): Use signin codes everywhere signin links are used

shane-tomlinson:
signin codes are the one true way to verify signins! This removes signin links from wherever they are used, and removes a lot of no longer needed tests while we are at it.

Not attached to an issue


shane-tomlinson:
@vbudhram and @mozilla/fxa-devs - r?

packages/fxa-content-server/app/scripts/lib/experiment.js
```diff
@@ -25,7 +25,6 @@ const MANUAL_EXPERIMENTS = {
   // no special experiment is created.
   sendSms: BaseExperiment,
   signupPasswordCWTS: BaseExperiment,
-  tokenCode: BaseExperiment,
   totp: BaseExperiment,
```
vbudhram: Looks like we don't have a totp experiment. Probably safe to remove!

packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
```diff
@@ -52,17 +51,7 @@ export default {
         // Check to see if this user is in the token code experiment,
```
vbudhram: Can probably remove mention to token code experiment here.

packages/fxa-content-server/tests/functional/fx_ios_v1_sign_in.js
```diff
@@ -53,7 +53,7 @@ const setupTest = thenify(function(options = {}) {
   const successSelector = options.blocked
     ? selectors.SIGNIN_UNBLOCK.HEADER
     : options.preVerified
-    ? selectors.CONFIRM_SIGNIN.HEADER
```
vbudhram: I wonder how much value `signin token code` language has now since we have the `confirm signin` views available.

packages/fxa-content-server/app/scripts/lib/experiment.js
shane-tomlinson: Great catch!

vasilicatamas:
I confirm that sign in confirmation codes are used for each signin attempt (Windows 10 64-bit, Android 9. iOS 13.4.1). Removing the needs:qa label.

PR #3537, opened 2019-11-29
refactor(test): Remove Firefox Desktop 57 and below in the tests

shane-tomlinson:
FxA officially only supports Fx desktop v60 and above, yet we still
have lots of code that checks the behavior of Fx 57 and below.

The major difference between Fx 57 and Fx 58 is the removal of
about:accounts. In Fx 57 and below, after verifying a login, about:accounts
was expected to take over and FxA was not supposed to do a screen
transition.

Starting with Fx 58, about:accounts no longer existed and FxA
needed to show a "you have finished the flow" screen, which allowed
us to display /connect_another_device.

Since the only major change was this transition, which if does
occur on Fx 57 and below isn't the biggest deal, I removed
support for preventing the transition and just transition
everywhere. This simplifies broker logic, it simplifies test
logic, and means we have fewer things to think about.

Also, since about:accounts was removed in Fx 58, I removed the
logic for "isAboutAccounts" within the code.

Not attached ot an issue.


packages/fxa-content-server/tests/functional/bounced_email.js
```diff
@@ -14,7 +14,7 @@ let deliveredEmail;
 const PASSWORD = 'password12345678';
 const ENTER_EMAIL_URL = `${
   intern._config.fxaContentRoot
-}?context=fx_desktop_v3&service=sync&automatedBrowser=true&forceAboutAccounts=true&forceUA=${encodeURIComponent(
+}?context=fx_desktop_v3&service=sync&automatedBrowser=true&forceUA=${encodeURIComponent(
   uaStrings.desktop_firefox_58
```
vladikoff: should we update this UA?

shane-tomlinson: I'm on the fence, if we update the ua string to fx 71 everywhere, we have to hook up WebChannel responses for fxaccounts:fxa_status, but maybe that's a good idea anyways, hook up a default response if one isn't specified.

I'll do that in a separate PR. 👍

shane-tomlinson:
@vladikoff and @mozilla/fxa-devs - r?

packages/fxa-content-server/tests/functional/bounced_email.js
shane-tomlinson: I have a branch to add default handlers, but have too many PRs open already. I'll open the PR once some of these land.

shane-tomlinson:
@vladikoff is on PTO, so I've gotta ask someone else to review this. @mozilla/fxa-devs - Any takers? This should further reduce the number of functional tests we have and make things a bit more maintainable!

PR #3543, opened 2019-12-02
feat(content-server): Append '(Web Session)' to clients without a deviceId

### LINKED ISSUE #2505: Append Web Session or device information to client names
**CURRENT**
image

**PROPOSED**
image

LZoog:
fixes #2505 

I'm not completely certain this is the proper solution for this issue. @ryanfeeley this changes "Web Session, [browser name]" references to "[browser name] (Web Session)" and then adds (Web Session) to devices that do _not_ have a `deviceId`.

From @vladikoff's test data provided in the issue (plus one browser session I added) here's an example of how some names render:

```
Firefox 70 (Web Session)
Firefox on Fox’s iPad
vladikoff4’s Nightly on Vlads-MBP
vladikoff’s Nightly on v
Firefox Lockwise
Firefox Send (Web Session)
Firefox Private Network (Web Session)
Firefox Sync (Web Session)
Firefox Monitor (Web Session)
Pocket (Web Session)
```

Is there a way to test Notes, Add-Ons, etc. locally, as Ryan provided in the issue? Lockwise seems like an odd one to me, since we capture the `deviceId` but don't display it with `name`.

I'm going to open this as a PR and not a draft since this may be all we need, but feedback is welcome.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/clients.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/clients.js
@@ -124,20 +124,19 @@ const View = FormView.extend(
       return _.map(items, item => {
         item.title = item.name;
 
-        if (item.scope) {
-          item.title += ' - ' + item.scope;
-        }
-
         if (item.lastAccessTimeFormatted) {
           if (item.clientType === Constants.CLIENT_TYPE_WEB_SESSION) {
             if (item.userAgent) {
-              item.title = this.translate(t('Web Session, %(userAgent)s'), {
+              item.title = this.translate(t('%(userAgent)s (Web Session)'), {
                 userAgent: item.userAgent,
               });
             } else {
               item.title = t('Web Session');
             }
             this._setLastAccessTimeFormatted(item, LAST_ACTIVITY_FORMATS.web);
+          } else if (item.deviceId === null) {
+            item.name = `${item.name} (Web Session)`;
+            item.title = item.name;
           } else if (item.clientType === Constants.CLIENT_TYPE_DEVICE) {
             this._setLastAccessTimeFormatted(
               item,
@@ -153,6 +152,10 @@ const View = FormView.extend(
             // unknown lastAccessTimeFormatted or not possible to format.
             item.lastAccessTimeFormatted = '';
           }
+        }
+
+        if (item.scope) {
+          item.title = `${item.title} - ${item.scope}`;
         }
 
         return item;
--- a/packages/fxa-content-server/app/tests/spec/views/settings/clients.js
+++ b/packages/fxa-content-server/app/tests/spec/views/settings/clients.js
@@ -754,6 +754,15 @@ describe('views/settings/clients', () => {
             isDevice: true,
             name: 'device-2',
           },
+          {
+            clientType: 'device',
+            deviceId: null,
+            deviceType: 'mobile',
+            isCurrentSession: false,
+            isDevice: true,
+            name: 'device-3',
+            lastAccessTimeFormatted: '20 minutes ago',
+          },
         ]);
 
         assert.equal(formatted[0].title, '123Done - profile');
@@ -763,7 +772,7 @@ describe('views/settings/clients', () => {
         );
         assert.equal(formatted[1].title, 'Pocket - profile,profile:write');
         assert.equal(formatted[2].title, 'Add-ons');
-        assert.equal(formatted[3].title, 'Web Session, Firefox 40');
+        assert.equal(formatted[3].title, 'Firefox 40 (Web Session)');
         assert.equal(formatted[4].title, 'Web Session');
         assert.equal(formatted[5].title, 'device-1');
         assert.equal(
@@ -775,6 +784,7 @@ describe('views/settings/clients', () => {
           formatted[6].lastAccessTimeFormatted,
           'Last seen time unknown'
         );
+        assert.equal(formatted[7].title, 'device-3 (Web Session)');
       });
     });
   });


rfk:
Thanks for pushing on this @LZoog!

>  Lockwise seems like an odd one to me, since we capture the deviceId but don't display it
> with name

IIUC the issue here is that the Lockwise client app doesn't set a custom device name, and so the name on its device record falls back to the default value of "Firefox Lockwise". We could consider filing an upstream bug to have it set a more descriptive name like "Firefox Lockwise on some-phone-model".

> Firefox Private Network (Web Session)
> Firefox Sync (Web Session)

These ones don't feel right to me, since FPN and Sync are not web-based properties. But I guess we don't currently track enough information to show anything better here, as you note [here](https://github.com/mozilla/fxa/issues/2505#issuecomment-553676858).

We could do better here if we kept track of which sessionToken was used to issue each OAuth refresh_token, which I could have sworn we had a bug on file about but now cannot find it.

LZoog:
@ryanfeeley, tagging you here for visibility.

> IIUC the issue here is that the Lockwise client app doesn't set a custom device name, and so the name on its device record falls back to the default value of "Firefox Lockwise". We could consider filing an upstream bug to have it set a more descriptive name like "Firefox Lockwise on some-phone-model".

This seems like it's the right move, but I also wonder how "Add Ons," "Notes," etc. returns. Vlad's JSON of his clients was extremely helpful, but I'm not sure how to test for every case here. There's probably others. 😕

> These ones don't feel right to me, since FPN and Sync are not web-based properties.

Right, I agree. I'm not sure what we would display for "Sync"... Maybe that's a special snowflake (meh) that doesn't get `(Web Session)` or a `deviceId`. TBH though it feels funny to me to even give these other products like Monitor or Send a `(Web Session)` at the end of their name. I'm not sure it gives a whole lot of value to users - like, what else could it be if someone didn't access something via "Web Session"? It gives connected browsers some context, but as a user, I don't know why I'd see "Firefox 70 (Web Session)" vs "Firefox on Fox’s iPad."

> We could do better here if we kept track of which sessionToken was used to issue each OAuth refresh_token, which I could have sworn we had a bug on file about but now cannot find it.

Definitely sounds like a good chunk of work. 😛 Is this something we need to do anyway?

rfk:
> though it feels funny to me to even give these other products like Monitor or Send a (Web Session) 

TBH, I personally think appending "Web Session" in almost all of these cases makes the page *worse*. As a user, f I see a `[Revoke]` button next to a thing that says "Web Session", I'm going to expect that clicking the button with terminate my web session with that service. It currently won't, with the possible exception of Send. So rather than leaving me confused, the "web session" text has helped me form an incorrect mental model of what's happening on this page.

But that's only my personal engineering-heavy take; I yield to @ryanfeeley on these matters 😁 

I think the real value here is linking the services back to a device description where appropriate (so I can e.g. distinguish which "Firefox Private Network" is which) but that requires some appropximation of this thing:

> > We could do better here if we kept track of which sessionToken was used to issue each OAuth
> > refresh_token, which I could have sworn we had a bug on file about but now cannot find it.
>
> Definitely sounds like a good chunk of work. 😛 Is this something we need to do anyway?

Yes, I think we will need to do this eventually, for several reasons (the most pressing is likely to be doing a better job of making "Sign out" behave has the user expects in all cases).

packages/fxa-content-server/app/scripts/views/settings/clients.js
```diff
@@ -155,6 +154,10 @@ const View = FormView.extend(
           }
         }
 
+        if (item.scope) {
+          item.title = `${item.title} - ${item.scope}`;
```
shane-tomlinson: There was at least one complaint about adding the scopes to the title, but I can't for the life of me find it.

LZoog:
> TBH, I personally think appending "Web Session" in almost all of these cases makes the page worse. [...] But that's only my personal engineering-heavy take

Yeah, this is how I feel. Can't tell if it's my tech bias or not so we'll leave this up to Ryan F.

> I think the real value here is linking the services back to a device description where appropriate (so I can e.g. distinguish which "Firefox Private Network" is which)

It sounds like either way we should create an issue for keeping track of which sessionToken was used to issue each OAuth refresh_token, but depending on @ryanfeeley's opinion here, maybe the real benefit/issue would instead be to capture devices used to access certain clients. Do we have a list of all possible clients/products anywhere? If we had a list, Ryan could tell us what is ideal where and we can break down what work needs to be done. I'm not sure what's appropriate to display for Sync and others, or if we want to use the "Web Session" text for Send etc.

packages/fxa-content-server/app/scripts/views/settings/clients.js
LZoog: That honestly makes sense, I did think this was a little weird (I didn't add this, I just moved it below the if/elses because I alter the title `if item.deviceId === null`):

image

We don't appear to be using `scope` anywhere else. Maybe we don't need it?

ryanfeeley:
Longer term, I would like us this list to attach services to specific devices.

For example, if we know that my current laptop is using Notes, Protection Report, and Sync, I would prefer to have them shown under one "Details…" button (mockup below).

image

These name changes are a half-measure to get us in that direction. But does it?

LZoog:
> These name changes are a half-measure to get us in that direction. But does it?

Hrmmm, I would say that it doesn't. It sounds like we need a new issue to capture attached devices on services where we don't already do so, and likely get to where you want this to end up in a follow-up issue (unless it should be done in one swoop).

Maybe it'd be good to design that screen out with the settings redesign in mind since I think we're going to get away from the ellipsis button and I don't want someone to do the work on it only to have to change it a month or two later.

I'm leaning towards closing this PR and associated issue as a "won't fix" and filing a new issue with where you're wanting this to end up (and with a `needs:ux` label), unless there are any specific products you think should definitely have "Web Session" text appended. What are your thoughts @ryanfeeley?

ryanfeeley:
@LZoog OK, let's save touching this area for the redesign.

LZoog:
I'm closing this PR and have linked several comments to #3577. Let's continue this discussion over there.

PR #3548, opened 2019-12-02
fix(tooltip): Update/remove inline password tooltip when password changes

### LINKED ISSUE #3486: Password fields are not validated inline with “Show password” option enabled
**Notes:**
This issue is a follow-up bug for #631

**Build:**
Stage [Train 1.151.0]

**Affected platforms:**
Windows 10 64-bit

**Steps to reproduce:**
1. Launch Firefox and click on the FxA toolbar menu button.
2. Type a new email address which does not have associated a Firefox Account..
3. Type an invalid email in the first password field.
4. Type a valid email in the second password field.
5. Enable the “Show password” option before the password error to be displayed.
6. Return to the first password field and insert the same password as the one typed in the second field.

**Expected results:**
“Passwords do not match” error is no longer displayed

**Actual Results:**
“Passwords do not match” error does not disappear even if the passwords are the same.

**Additional notes:**
- This issue was encountered on Firefox 70.0.1.
- I am attaching a video.

passwordsvalidation

### LINKED ISSUE #3489: (sign-up) Repeat password balloon no longer shows
I believe this is regression from https://github.com/mozilla/fxa/pull/3406. The balloon shows in commit 4a2420f21 before this was merged but no longer in e098929f4.

@ryanfeeley @davismtl do we want to rethink this now that we have the inline validation? When does it disappear?

With no inline validation but with balloon (sorry for cutoff):
pw-balloon

With inline validation and no balloon:
pw-balloon-noshow

### LINKED ISSUE #3504: Password Verification Tooltip Incorrectly Displayed
There's a timing(?) issue somewhere that's causing the not matching password tooltip to be displayed, even when the passwords are matching.

image

vbudhram:
Fixes #3504 
Fixes #3489
Fixes #3486

This PR updates the inline password validation rules to fix the above issues.

@LZoog you mentioned working on #3489 but I think I found an easy fix while working those other issues. I can flag you for review after adding tests here.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sign_up_password.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_up_password.js
@@ -36,7 +36,6 @@ const SignUpPasswordView = FormView.extend({
 
   events: assign({}, FormView.prototype.events, {
     'click .use-different': preventDefaultThen('useDifferentAccount'),
-    'keyup #vpassword': '_onConfirmPasswordKeyUp',
   }),
 
   useDifferentAccount() {
@@ -86,6 +85,12 @@ const SignUpPasswordView = FormView.extend({
     return proto.isValidEnd.call(this);
   },
 
+  onFormChange() {
+    this.checkPasswordsMatchDebounce();
+
+    return proto.onFormChange.call(this);
+  },
+
   showValidationErrorsEnd() {
     if (!this._doPasswordsMatch()) {
       this.showValidationError(
@@ -93,6 +98,9 @@ const SignUpPasswordView = FormView.extend({
         AuthErrors.toError('PASSWORDS_DO_NOT_MATCH'),
         false
       );
+    } else {
+      // Called to clear validation tooltips.
+      this.$(VPASSWORD_INPUT_SELECTOR).change();
     }
   },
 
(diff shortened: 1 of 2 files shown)


LZoog:
👍 Sweet, thanks for taking care of these!

packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -78,12 +77,10 @@ const SignUpPasswordView = FormView.extend({
     );
   },
 
-  isValidEnd() {
-    if (!this._doPasswordsMatch()) {
-      return false;
-    }
+  onFormChange() {
```
vbudhram: This was the fix for https://github.com/mozilla/fxa/issues/3489, for some reason having the `keyup #vpassword` event caused the password info bubble to disappear. Instead, I hooked into the existing events that get trigger when form element changed.

packages/fxa-content-server/app/scripts/views/sign_up_password.js
```diff
@@ -93,6 +90,9 @@ const SignUpPasswordView = FormView.extend({
         AuthErrors.toError('PASSWORDS_DO_NOT_MATCH'),
         false
       );
+    } else {
+      // Called to clear validation tooltips.
+      this.$(VPASSWORD_INPUT_SELECTOR).change();
```
vbudhram: This fixes other two issues. If the tooltip is on an element whose value does not change, the the validation rules to remove it were not triggered. Simple fix was to fake an element change.

PR #3550, opened 2019-12-02
task(fxa-content-server): tidy up sentry configuration in content server

meandavejustice:
Just tidying up some unused settings in config for content server


shane-tomlinson:
@meandavejustice - this needs a rebase

Issue #3563, opened 2019-12-03
Consolidate sentry utility wrapper in fxa-shared


meandavejustice:
We are using the sentry utility wrapper in payments server and content server. We should pull it out into `fxa-shared`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-59)
┆Issue Number: FXA-59

data-sync-user:
➤ Vijay Budhram commented:

This is part of the NextJS shared libraries now.

PR #3564, opened 2019-12-03
task(payments-server): update sentry for sourcemaps on payments server

### LINKED ISSUE #2345: Ensure source maps show up in Sentry

meandavejustice:
- fixes #2345

Took `src/lib/url.js` `src/lib/logger.js` and `src/lib/sentry.js` from content server. Filed https://github.com/mozilla/fxa/issues/3563 as a follow up.


packages/fxa-payments-server/src/lib/sentry.js
```diff
@@ -0,0 +1,162 @@
...
+function cleanUpQueryParam(url = '') {
+  var startOfParams = url.indexOf('?');
+  var newUrl = url;
+  var params;
+
+  if (startOfParams === -1) {
+    return newUrl;
+  }
+
+  params = Url.searchParams(url.substring(startOfParams + 1));
```
lmorchard: It looks like here and a few lines below are the only spots where we use the new `Url` module - and that module is the only use of the new `underscore` dependency.

Could we instead use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) from web APIs here? I think [there's a working polyfill](https://github.com/jerrybendy/url-search-params-polyfill) if we absolutely have to support browsers [older than Firefox 29](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility)

Alternatively, maybe tweak the Url module to *not* require underscore? Mainly it would be nice not to add it if we're not using it elsewhere on payments-server

packages/fxa-payments-server/src/lib/config.test.ts
```diff
@@ -155,7 +155,10 @@ const expectedMergedConfig = {
       jenny: '8675309',
     },
   },
-  sentryDsn: '',
+  sentry: {
+    dsn: '',
+    url: '',
```
lmorchard: One of [the test failures](https://circleci.com/gh/mozilla/fxa/124114?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) looks like it's coming from hereabouts: This [has a default value in `defaultConfig`](https://github.com/mozilla/fxa/blob/436f929ee8ddc0cddbdf65348d75e32adcaa56a6/packages/fxa-payments-server/src/lib/config.ts#L52), so it's not going to be an empty string here in `expectedMergedConfig `

packages/fxa-payments-server/src/lib/sentry.js
meandavejustice: I agree, I'd like to avoid adding dependencies also. However, I filed https://github.com/mozilla/fxa/issues/3563 to consolidate these utils into fxa-shared. What do you think about addressing these changes in one place when we deal with that issue?

meandavejustice: Actually, now that the tests pass we still aren't meeting the coverage threshold because of the new code, maybe it does make sense to take care of this now instead of lowering the threshold. 
```
Jest: "global" coverage threshold for statements (88%) not met: 83.51%
Jest: "global" coverage threshold for branches (94%) not met: 80.99%
Jest: "global" coverage threshold for lines (88%) not met: 83.46%
```

lmorchard: > maybe it does make sense to take care of this now instead of lowering the threshold.

Yeah, the thresholds are there to force us to write tests to keep the coverage quality up. Better not to cheat there 😞

Issue #1531, opened 2019-06-06
"Sync Connected" notification should say “Firefox Preview” for Fenix, not "Firefox Sync"


vladikoff:
Authentication screens should refer to “continue to Firefox Sync” and the device name in device manager should be “Firefox Preview” (with android mobile icon). Currently they are the same.

cc @ryanfeeley

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-172)
┆Priority: Medium
┆Issue Number: FXA-172

rfk:
In the 'device manager', we will soon start using the name provided by the client rather than the default OAuth client name.  I think the trickier case will be the "now syncing with..." notificaiton described in https://github.com/mozilla/fxa/issues/1397

ryanfeeley:
Further evidence that this is needed came to light during Sync Fest.
![IMG_0097](https://user-images.githubusercontent.com/68213/59198009-b61d3700-8b60-11e9-905e-ca0e556fcde9.PNG)

vladikoff:
<img src=https://user-images.githubusercontent.com/128755/59200903-42caf380-8b67-11e9-93c6-fb6dc5553b58.png width=350 />

<img src=https://user-images.githubusercontent.com/128755/59200923-4bbbc500-8b67-11e9-8b6c-fa598c25ec1f.png width=350 />

rfk:
> Authentication screens should refer to “continue to Firefox Sync” and the device name in device
> manager should be “Firefox Preview”

Do we have a plan for how we might achieve this?

One option would be to change the logic on the signup screen, to look at the scopes being requested and say "continue to Firefox Sync" for things that are requesting the "oldsync" scope. Feels hacky, but maybe is a step towards a more OAuth-style "give X permission to access Y on your account?" style experience.

vladikoff:
> One option would be to change the logic on the signup screen, to look at the scopes being requested and say "continue to Firefox Sync" for things that are requesting the "oldsync" scope.

philbooth:
@vladikoff I can do this, is there a specific user agent string I can look for so we special case the device name here?

juliemccracken:
@philbooth, I'm checking with the Fenix team to find out how important this is to them. Their go/no go meeting is today for tomorrow's release.

vladikoff:
> special case the device name here?

I hope we can special case by Client ID maybe.. I don't think we can use the user agent

philbooth:
[...]

PR #3569, opened 2019-12-04
feat(content-server): Set relier 'serviceName' default to 'Account Settings'

### LINKED ISSUE #2054: Include Account Settings subheading when signing in to Settings

LZoog:
fixes #2054

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/ready.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/ready.mustache
@@ -16,23 +16,23 @@
     {{/isSync}}
 
     {{^isSync}}
-      {{#serviceName}}
+      {{#isFromRelyingParty}}
         <p class="account-ready-service">{{#t}}You are now ready to use %(serviceName)s{{/t}}</p>
-          {{#showContinueButton}}
-              <div class="button-row">
-                  <button class="primary-button btn-continue">{{#t}}Continue{{/t}}</button>
-              </div>
-          {{/showContinueButton}}
-      {{/serviceName}}
-      {{^serviceName}}
-          {{#emailVerified}}
-              <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
-          {{/emailVerified}}
+        {{#showContinueButton}}
+          <div class="button-row">
+            <button class="primary-button btn-continue">{{#t}}Continue{{/t}}</button>
+          </div>
+        {{/showContinueButton}}
+      {{/isFromRelyingParty}}
+      {{^isFromRelyingParty}}
+        {{#emailVerified}}
+          <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
+        {{/emailVerified}}
 
-          {{^emailVerified}}
-              <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
-          {{/emailVerified}}
-      {{/serviceName}}
+        {{^emailVerified}}
+          <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
+        {{/emailVerified}}
+      {{/isFromRelyingParty}}
     {{/isSync}}
 
       {{#isPasswordReset}}
--- a/packages/fxa-content-server/app/scripts/views/mixins/service-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/service-mixin.js
@@ -14,7 +14,7 @@ export default {
   },
 
   setInitialContext(context) {
-    context.set(this.relier.pick('service', 'serviceName'));
+    context.set(this.relier.pick('service', this.translate('serviceName')));
   },
 
   transformLinks() {
--- a/packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js
+++ b/packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js
@@ -6,6 +6,7 @@ import { assert } from 'chai';
 import Backbone from 'backbone';
 import BaseView from 'views/base';
 import Cocktail from 'cocktail';
+import Constants from 'lib/constants';
 import NullBroker from 'models/auth_brokers/base';
 import OAuthRelier from 'models/reliers/oauth';
 import ServiceMixin from 'views/mixins/service-mixin';
@@ -62,6 +63,15 @@ describe('views/mixins/service-mixin', () => {
       view.setInitialContext(context);
       assert.equal(context.get('service'), 'sync');
       assert.equal(context.get('serviceName'), 'Firefox Sync');
+    });
+
+    it('sets `serviceName` default from the relier', () => {
+      const context = new Backbone.Model({});
+      view.setInitialContext(context);
+      assert.equal(
+        context.get('serviceName'),
+        Constants.RELIER_DEFAULT_SERVICE_NAME
+      );
     });
   });
 
--- a/packages/fxa-content-server/app/tests/spec/views/ready.js
+++ b/packages/fxa-content-server/app/tests/spec/views/ready.js
@@ -15,6 +15,11 @@ import sinon from 'sinon';
 import BrowserRelier from 'models/reliers/browser';
 import View from 'views/ready';
 import WindowMock from '../../mocks/window';
+import {
+  SIGNIN_COMPLETE,
+  SIGNUP_COMPLETE,
+  RESET_PASSWORD_COMPLETE,
+} from '../../../../tests/functional/lib/selectors';
 
 describe('views/ready', function() {
   let broker;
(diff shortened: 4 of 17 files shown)


packages/fxa-content-server/app/scripts/templates/ready.mustache
```diff
@@ -16,23 +16,17 @@
     {{/isSync}}
 
     {{^isSync}}
-      {{#serviceName}}
+      {{#emailVerified}}
```
LZoog: @shane-tomlinson I took out the `serviceName` checks in other template files since we don't need to check for truthiness anymore and reversed the logic here to make sure we display `emailVerified` if it's truthy and display the `serviceName` section here otherwise.

packages/fxa-content-server/app/tests/spec/views/ready.js
```diff
@@ -98,25 +103,24 @@ describe('views/ready', function() {
...
-        var html = view.$('section').text();
-        assert.include(html, 'Firefox Sync');
+        assert.equal(
+          view.$(SIGNIN_COMPLETE.SERVICE_NAME).text(),
+          'You are now ready to use Account Settings'
+        );
       });
     });
 
-    // regression test for #1216
```
LZoog: Doesn't look like we need this test anymore since a `serviceName` will always be provided so I wrote a different test to check for its default value.

packages/fxa-content-server/app/scripts/templates/ready.mustache
shane-tomlinson: Something about this change seems off, because the button on line 26 should only show to OAuth RPs if the account is verified.

shane-tomlinson: After looking at ready.js (and being very confused), `escapedEmailReadyText` is only displayed when a user has to verify their primary/secondary email for making changes to the account settings. 

Now we need to disentangle the rest of the branches. I really don't like mustache sometimes.

Maybe something like this?
```mustache
{{#escapedEmailReadyText}}
  <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
{{/escapedEmailReadyText}}
{{^escapedEmailReadyText}}
  {{#emailVerified}}
    <p class="account-ready-service">{{#t}}You are now ready to use %(serviceName)s{{/t}}</p>
    {{#showContinueButton}}
      <div class="button-row">
        <button class="primary-button btn-continue">{{#t}}Continue{{/t}}</button>
      </div>
    {{/showContinueButton}}
  {{/emailVerified}}
  {{^emailVerified}}
    <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
  {{/emailVerified}}
{{/escapedEmailReadyText}}
```

packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js
```diff
@@ -63,6 +63,12 @@ describe('views/mixins/service-mixin', () => {
       assert.equal(context.get('service'), 'sync');
       assert.equal(context.get('serviceName'), 'Firefox Sync');
     });
+
+    it('sets `serviceName` default from the relier', () => {
+      const context = new Backbone.Model({});
+      view.setInitialContext(context);
+      assert.equal(context.get('serviceName'), 'Account Settings');
```
shane-tomlinson: Ahha! I forgot, we have to ensure the service name gets translated, so in the service-mixin when setting the [context's serviceName](https://github.com/mozilla/fxa/blob/3d0a657355337c27c847bc511fa4560658e17302/packages/fxa-content-server/app/scripts/views/mixins/service-mixin.js#L17), run it through `this.translate`

packages/fxa-content-server/app/scripts/templates/ready.mustache
LZoog: Sorry in advance for the length of this comment but let me tell you how I got from your suggestion to what I just pushed.

Did you try running unit tests with your suggestion? I don't think we can use escaped HTML to check for truthiness. I battled too long with this today:
```
TypeError: string is undefinedinterpolate
```
Seen at: http://127.0.0.1:3030/tests/index.html?grep=views%2Fready

I tried temporarily changing `escapedEmailReadyText` to a string to make sure that check is what's causing that error, and one test fails with a different error. I also don't see where we check for escaped text like this in other mustache files (I searched the project for `{{{` and we always use a different variable check when conditionally rendering escaped text).

This of course doesn't solve the problem, but I'd asked Ryan about that "Your account is ready!" text a few days ago, and he said he wants it replaced with "You are now ready to use Account Settings". I changed what you suggested to this:

```
{{#escapedEmailReadyText}}
  <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
{{/escapedEmailReadyText}}
{{^escapedEmailReadyText}}
  <p class="account-ready-service">{{#t}}You are now ready to use %(serviceName)s{{/t}}</p>
  {{#emailVerified}}
    {{#showContinueButton}}
      <div class="button-row">
        <button class="primary-button btn-continue">{{#t}}Continue{{/t}}</button>
      </div>
    {{/showContinueButton}}
  {{/emailVerified}}
{{/escapedEmailReadyText}}
```

... I then tried creating `showEscapedEmailReadyText` as a boolean of `escapedEmailReadyText`, but it looks like `escapedEmailReadyText` is always truthy anyway. 🤦‍♀️ Since we can't use handlebars `{{#if}}`, I needed a way to check for if there was an RP or not, which was previously done with a check for `serviceName`.

That's how I ended up here. I stuck `'Account Settings'` as a string in `Constants.js` so we can update that text in one place and our `if` check in `ready.js` for `isFromRelyingParty` will still work, but I left the tests as-is (except for the `service-mixin` test) since I feel those should probably be explicit. I reverted back to what we had in `ready.mustache` before modifications, and changed `serviceName` checks to `isFromRelyingParty`, and updated "Your account is ready!" to "You are now ready to use Account Settings". We lose a little reuse here since lines 20 and 33 are the same, but, I think that's fine.

packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js
LZoog: I made this change (haven't pushed due to my other comment), and I knew about `t('text')` for L10n but was unaware I needed `this.translate` as well. I was going to ask if we have this in the docs but [I found it](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-localization)! 😏

I see "`t` is a signal to the l10n extraction script that the embedded string is meant for translation". Is `translate` (looks like it comes from `base.js`) just to protect from unescaped HTML?

shane-tomlinson: `t` as a function in JS doesn't actually do any translation, it's just the signal to extract the string. `{{#t}}` and `{{#unsafeTranslate}}` within the mustache templates translates, but everything else needs translated manually. I should add that to the content server gotchas.

packages/fxa-content-server/app/scripts/templates/ready.mustache
shane-tomlinson: > This of course doesn't solve the problem, but I'd asked Ryan about that "Your account is ready!" text a few days ago, and he said he wants it replaced with "You are now ready to use Account Settings". I changed what you suggested to this:

Unfortunately, a user that opens a verification reminder link in a browser other than the  original browser they signed up in will not be able to be redirected to settings because they won't have a session. Is that handled by one of the other cases? If so, then all good.

LZoog: @shane-tomlinson I'm not sure the user would have had the option to redirect to Account Settings anyway because they wouldn't be shown the "Continue" button. 🤔

I went ahead and reverted that change back to the generic "Your account is ready!" copy... but now I'm wondering if it would actually be OK to display it without the redirect button. @ryanfeeley this goes back to my Slack convo with you about this "ready" page. If we should display a "Continue" button with "You are now ready to use Account Settings," on this screen, I can create another issue for that so we can get this in which will show "Your account is ready!", or, if I understand Shane's comment and all of the different conditionals here, I think we can display "You are now ready to use Account Settings" without it.

LZoog: Per @ryanfeeley, he would prefer "You are now ready to use Account Settings" without the redirect button instead of "Your account is ready!"

I'll change this back later today. I can open a new issue around displaying/getting the continue button to work.

packages/fxa-content-server/app/scripts/views/mixins/service-mixin.js
```diff
@@ -14,7 +14,7 @@ export default {
   },
 
   setInitialContext(context) {
-    context.set(this.relier.pick('service', 'serviceName'));
+    context.set(this.relier.pick('service', this.translate('serviceName')));
```
rfk: Hi @LZoog, I'm poking around in this code to see whether I can fix https://github.com/mozilla/fxa/issues/1531, and I'm confused by the use of `this.translate` here. My understanding of `this.translate` is that this would try to find a localized version of the literal string "serviceName", then fall back to returning "serviceName" if no translation is found. I don't think we have a localization for this string, am I missing something?

packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js
LZoog: Hey @rfk, I'm sorry, I totally missed the notification that you commented on this a few days ago.

Were you able to figure out what you needed? I'm still a little fuzzy on our localization process and asked Shane a couple of questions here about it. It does seem like `this.translate('serviceName')` may be incorrect, does the following seem right?

```
const relierContext = this.relier.pick('service', 'serviceName');
relierContext.serviceName = this.translate(relierContext.serviceName);
context.set(relierContext);
```

PR #3570, opened 2019-12-04
refactor(experiments): Simplify how experiments and metrics are created

shane-tomlinson:
Previous to this, if a user was part of an experiment, we had to
manually call `this.createExperiment` within a view to ensure the
metrics are reported.

I added a new method, `this.chooseExperimentGroup` to the ExperimentInterface
that automatically calls `this.createExperiment` for users that are
part of the experiment.

All the places where `getExperimentGroup` then `createExperiment` were
called have been replaced with a new call to `chooseExperimentGroup`,
simplifying setup and reducing the number of tests that are needed.

Not attached to an issue, noticed this while writing docs on how
to add a new experiment:

https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-content-server-ab-testing

@chenba - you recently went through this process, maybe you found setting up the experiment more cumbersome than need be? Mind having a look?

@vladikoff and @vbudhram - you both have set up many experiments and probably have an opinion!

@mozilla/fxa-devs - r?


shane-tomlinson:
To prove to myself experiments were still being reported, I forced @chenba's EmailMX test using:

http://127.0.0.1:3030/?context=fx_desktop_v3&service=sync&forceExperiment=emailMxValidation&forceExperimentGroup=treatment

You can see the experiment is reported in the Amplitude event:

Screenshot 2019-12-04 at 13 20 55

packages/fxa-content-server/app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -57,37 +57,21 @@ const ExperimentMixin = {
...
-   *
-   * @param {String} experimentName - name of experiment to create.
-   * @param {String} groupType - which group the user is in.
-   * @returns {Object} experiment object, if created.
-   */
-  createExperiment(...args) {
-    if (this.experiments) {
-      // force the flow model to be initialized so that
-      // the experiment is logged.
-      this.notifier.trigger('flow.initialize');
```
shane-tomlinson: I pushed this into the ExperimentInterface which enabled me to replace this with a delegate method below.

packages/fxa-content-server/app/scripts/views/mixins/token-code-experiment-mixin.js
```diff
@@ -1,67 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson: This file was no longer used and should have been removed with another PR that was merged earlier in the week.

packages/fxa-content-server/app/scripts/lib/experiment.js
```diff
@@ -150,12 +150,27 @@ _.extend(
...
+    /**
+     * Choose the experiment group for the user is in for `experimentName`.
+     * If the user is part of an experiment, ensures the experiment and group
+     * are reported to Amplitude.
+     *
+     * @param {String} experimentName
+     * @param {Object} [additionalInfo] additional info to pass to the experiment grouping rule.
+     * @returns {false|String}
+     */
+    chooseExperimentGroup(experimentName, additionalInfo) {
```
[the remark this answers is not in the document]
shane-tomlinson: What about `getAndReportExperimentGroup`?

chenba:
👍

Issue #3531, opened 2019-11-29
Convert the secondary email verification flow from link to code based verification


shane-tomlinson:
Similar to #2815, this is to convert the add a secondary email flow from link to code based verification.

blocks #2814

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-593)
┆Issue Number: FXA-593

vbudhram:
Assigning myself since this is needed for https://github.com/mozilla/fxa/issues/2653

vbudhram:
All the parts for this standalone view have landed, closing.

Issue #2653, opened 2019-09-27
(Phase 2) Recovery Keys and Secondary Email at Registration


davismtl:
We’ve always had a steady volume of users complaining or reaching out to us via Bugzilla, SUMO, Twitter and Reddit due to being locked out of their accounts.

There are three reasons for this:
- They don’t have access to their email anymore;
- They forgot their password;
- They lost their device for TOTP.

This feature proposes to tackle the first two. The impact of the first reason is that users can’t get into their accounts. The impact of the second reason is that users risk losing their Sync data if they do not have more than 1 device actively Syncing.

In this second phase, we will also send users an email immediately after registration to encourage them to setup a recovery key and a secondary email. This email will *NOT* link to the settings like phase 1. For phase 2, we will link people to a screen that can be used from various places like our login flow and account toolbar menu in the browser.

Feature document: 
https://docs.google.com/document/d/1fJA_wfrgVF4cuQVzoWqQh22kiejDHbXJrkzTNuVldII/edit#

┆Issue is synchronized with this [Jira Epic](https://jira.mozilla.com/browse/FXA-383)
┆Issue Number: FXA-383

davismtl:
Secondary email after login:
https://mozilla.invisionapp.com/share/5JUK0JFFN3M#/390502550_Secondary_Email_Recommendation

clouserw:
Please add the issues to this epic.  thanks

davismtl:
from mtg:
This is now in progress. Just need to link out to bugs now.

davismtl:
This will be going out in January.

vbudhram:
* Code has landed to verify secondary email via a code.
* Started working on the UX for secondary email after registration

davismtl:
Should we enable this as an experiment for adding a secondary emails in the flow?
@vbudhram @irrationalagent @davismtl to sync up on this.

data-sync-user:
➤ Alex Davis commented:

[~vbudhram@mozilla.com] , Ryan Feeley and I have been updating the feature document to unblock you.

What is left?

  * Ryan is reviewing the copy because we don't like it.
  * He's making small changes to the design to account for two ways people can see the standalone recovery key flow
    * linked directly (e.g. email, eventually maybe toolbar menu)
    * in an auth flow like login or registration)

  * Ryan should have these done today or tomorrow at latest.

What has changed?

  * The copy will try to outline the benefit and strategic differentiation about how we handle data to underline the importance of a recovery key
[...]

PR #3574, opened 2019-12-04
feat(emails): Add server side support to verify secondary email by code

vbudhram:
Connects https://github.com/mozilla/fxa/issues/3531

This is needed as part of https://github.com/mozilla/fxa/issues/2653. To keeps things manageable, I will split this into adding server-side support and UX screens in another PR.

This PR adds the new email templates and routes to enabled this feature. Some key design decisions:

* Verify secondary email by code is now in its own route `recovery_email/secondary/verify_code`, instead of overloading the `recovery_email/verify_code`. The later route does way to many things and it makes sense to extract this out.
* Resend secondary email by code is in its own route as well for same reasons above.
* The create secondary email route `POST /recovery_email`, now accepts an optional param, `verificationMethod`, if content-server specifies `email-otp` then an otp code is generated using the `emailCode` as the otp secret. This follows the same pattern used in sign-in and sign-up codes.

### DIFF
--- a/packages/fxa-auth-server/docs/api.md
+++ b/packages/fxa-auth-server/docs/api.md
@@ -55,6 +55,8 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     - [POST /recovery_email (:lock: sessionToken)](#post-recovery_email)
     - [POST /recovery_email/destroy (:lock: sessionToken)](#post-recovery_emaildestroy)
     - [POST /recovery_email/set_primary (:lock: sessionToken)](#post-recovery_emailset_primary)
+    - [POST /recovery_email/secondary/verify_code (:lock: sessionToken)](#post-recovery_emailseconaryverify_code)
+    - [POST /recovery_email/secondary/resend_code (:lock: sessionToken)](#post-recovery_emailseconaryresend_code)
   - [Oauth](#oauth)
     - [GET /oauth/client/{client_id}](#get-oauthclientclient_id)
     - [POST /account/scoped-key-data (:lock: sessionToken)](#post-accountscoped-key-data)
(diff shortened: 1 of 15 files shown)


vbudhram:
@mozilla/fxa-devs Calling this ready for review! There is no UX that uses the new routes..yet, so I think this should be relatively safe to merge when ready.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -534,6 +537,7 @@ module.exports = (
         validate: {
           payload: {
             email: validators.email().required(),
+            verificationMethod: validators.verificationMethod,
```
shane-tomlinson: Is this going to be A/B tested, or is it possible to just flip to 'email-otp' and be done with it?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -576,22 +580,47 @@ module.exports = (
         await db.createEmail(uid, emailData);
 
         const geoData = request.app.geo;
+        const code = otpUtils.generateOtpCode(hex, otpOptions);
```
shane-tomlinson: `code` is only used by the `email-otp` branch, should we just calculate it within that branch and void spinning CPU cycles for the default?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -758,6 +787,155 @@ module.exports = (
...
+    {
+      method: 'POST',
+      path: '/recovery_email/secondary/resend_code',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+        validate: {
+          payload: {
+            email: validators.email().required(),
```
shane-tomlinson: Same as the question above, why pass in an email here, is it to support multiple secondaries? If we only support one secondary, couldn't we by default send the email there and avoid relying on user generated input?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -758,6 +787,155 @@ module.exports = (
...
+
+        const sessionToken = request.auth.credentials;
+        const ip = request.app.clientAddress;
+        const geoData = request.app.geo;
+        const { email } = request.payload;
+
+        await customs.check(
+          request,
+          sessionToken.email,
+          'recoveryEmailSecondaryResendCode'
```
shane-tomlinson: Are new customs checks hooked up automatically?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -758,6 +787,155 @@ module.exports = (
...
+      method: 'POST',
+      path: '/recovery_email/secondary/verify_code',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+        validate: {
+          payload: {
+            email: validators.email().required(),
+            code: validators.DIGITS,
```
shane-tomlinson: Should `code` be restricted to a known length?

packages/fxa-auth-server/docs/api.md
```diff
@@ -2196,6 +2198,78 @@ by the following errors
...
+
+<!--begin-route-post-recovery_emailsecondaryresend_code-->
+
+This endpoint resend the otp verification to verify the secondary email.
+
+<!--end-route-post-recovery_emailsecondaryresend_code-->
+
+##### Request body
+
+- `email`: _validators.email.required_
```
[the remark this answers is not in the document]
vbudhram: Correct, currently we support multiple secondary emails but only allow the UX to add one. Passing the email along here will allow us to ensure we are verifying against the correct email.

packages/fxa-auth-server/lib/routes/emails.js
vbudhram: Unfortunately we have to support both methods of verification for a little bit. Secondary emails are created from the settings page using a link to verify, and in https://github.com/mozilla/fxa/issues/2653 they can be created after login/registration using a code.

Hopefully, if phase 2 is successful, we can update the UX in settings to verify using a code and then remove `verificationMethod`

packages/fxa-auth-server/lib/routes/emails.js
vbudhram: Nope! I'll update this

PR #3579, opened 2019-12-05
refactor(test): Hook up default WebChannel responses if none specified

shane-tomlinson:
This simplifies developer lives by hooking up default WebChannel
responses in applicable environments if none are specified.


shane-tomlinson:
@mozilla/fxa-devs - r?

meandavejustice:
Nice to see all of the `webChannelResponses` stuff getting removed

PR #3595, opened 2019-12-06
feat(signin): Convert signin w/ unverified account to codes

### LINKED ISSUE #3459: Signing in with an unverified account uses links instead of codes
STR

1. Sign up with a new email address. Do not verify.
2. Sign in with the email from step 1.
3. A link is sent instead of a signup code.

shane-tomlinson:
Only handles signin that requires password with unverified account,
does not handle cached signin with an unverified account or visiting
settings with an unverified account.

The extra caveat is this does not remove auth-server support for sending 
link based emails, that needs to come in a subsequent train. See https://github.com/mozilla/fxa/blob/2606f6b844c1efb10412652724f6f6a52814c32a/packages/fxa-auth-server/lib/routes/utils/signin.js#L472

fixes #3459

### DIFF
--- a/packages/fxa-auth-server/lib/routes/utils/signin.js
+++ b/packages/fxa-auth-server/lib/routes/utils/signin.js
@@ -313,12 +313,14 @@ module.exports = (log, config, customs, db, mailer) => {
       }
 
       function sendVerifyAccountEmail() {
+        if (verificationMethod === 'email-otp') {
+          return sendVerifyLoginCodeEmail();
+        }
         // If the session doesn't require verification,
         // fall back to the account-level email code for the link.
         const emailCode =
           sessionToken.tokenVerificationId ||
           accountRecord.primaryEmail.emailCode;
-
         return mailer
           .sendVerifyEmail([], accountRecord, {
             code: emailCode,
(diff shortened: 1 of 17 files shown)


packages/fxa-content-server/tests/functional/settings.js
```diff
@@ -246,30 +246,6 @@ registerSuite('settings', {
   },
 });
 
-registerSuite('settings unverified', {
```
shane-tomlinson: This test is duplicated in settings_common.js

shane-tomlinson:
@vbudhram - r?

packages/fxa-auth-server/lib/routes/utils/signin.js
```diff
@@ -467,9 +469,19 @@ module.exports = (log, config, customs, db, mailer) => {
...
     getSessionVerificationStatus(sessionToken, verificationMethod) {
       if (!sessionToken.emailVerified) {
+        // for unverified accounts, only 'email', and 'email-otp' are valid.
+        // email-otp is the end goal, but a transition train is needed.
+        // Set the default to 'email' to handle train->train upgrades. If
+        // a user of content-server X loads their JS and then auth-server X+1
+        // is deployed, the content server of train X is not able to handle
+        // the 'email-otp' result and still send the user to the /confirm screen
+        // that expect verification links.
+        if (verificationMethod !== 'email-otp') {
```
vbudhram: Makes sense 👍🏽

packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
```diff
@@ -130,6 +130,7 @@ export default {
 
         if (
           AuthErrors.is(err, 'TOTP_REQUIRED') ||
+          AuthErrors.is(err, 'INSUFFICIENT_ACR_VALUES') ||
```
vbudhram: Did you get this while testing 2FA?

vasilicatamas:
Removing the needs:qa label based on the #3459 verification (issue solved by this pull request)

PR #3602, opened 2019-12-06
chore(ci): divide functional tests into groups

### LINKED ISSUE #3515: Why did we need xlarge on circleci again?
follow-up to #3512

chenba:
We switched to using xlarge resource class in CircleCI for content
server tests due to suspected OOM errors.  This patch splits the tests
into groups, thus forcing Firefox to restart and release its memory
between each group.

Fixes #3515 (for now at least)

@mozilla/fxa-devs r?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ jobs:
       - run: ../../.circleci/deploy.sh << parameters.module >>
 
   fxa-content-server:
-    resource_class: xlarge
+    resource_class: medium+
     parallelism: 6
     docker:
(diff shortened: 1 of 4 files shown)


chenba:
Hmm we don't run the email service though, by looking at the `PM2` output?

shane-tomlinson:
This is fantastic @chenba - I wonder if this could be used to better balance the number of tests in each of the individual runners on Circle. Right now the number of *suites* is divided evenly among the runners, whereas with an extension of this it seems like it should be possible to divide the number of tests evenly.

PR #3607, opened 2019-12-09
fix(fenix): Do not use Sync Optional with Fenix

### LINKED ISSUE #3553: Do not use the sync-optional flow with Fenix
Ref https://github.com/mozilla/application-services/issues/2283, but filing it here for tracking because the fix will need to be in FxA.

Users who sign in to Fenix currently get the "sync optional" flow. They shouldn't, because Fenix doesn't support it, and IIUC clicking the "do not sync this device" button will not have any effect in Fenix.

The sync-optional flow should only be offered to `context=fx_desktop_v3` clients, and only if they report `multiService: true` in their FxA webchannel handshake

shane-tomlinson:
Add a new capability `syncOptional` to auth_brokers, only
enable this capability in fx_desktop_v3 *if* the `multiService`
capability is returned from the browser.

fixes #3553

@vladikoff - r?

If approved, this needs to be uplifted to train-151

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
@@ -72,6 +72,7 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
       });
       return this.set('chooseWhatToSyncWebV1Engines', syncEngines);
     }
+    return proto.onFxaStatus.call(this, response);
   },
 
   createChannel() {
--- a/packages/fxa-content-server/app/scripts/views/mixins/signup-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/signup-mixin.js
@@ -61,8 +61,8 @@ export default {
             !this.isCWTSOnSignupPasswordEnabled())
         ) {
           return this.navigate('choose_what_to_sync', {
-            account: account,
-            allowToDisableSync: this.relier.get('service') !== 'sync',
+            account,
+            allowToDisableSync: this.broker.hasCapability('syncOptional'),
             pollVerification: this.relier.get('service') === 'sync',
             // choose_what_to_sync screen will call onSubmitComplete
             // with an updated account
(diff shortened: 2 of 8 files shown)


packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
```diff
@@ -72,6 +72,7 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
       });
       return this.set('chooseWhatToSyncWebV1Engines', syncEngines);
     }
+    return proto.onFxaStatus.call(this, response);
```
shane-tomlinson: This is a tangential change, I noticed none of the `onFxaStatus` were calling the super class method, which it should so the base class can set the [supportsPairing](https://github.com/mozilla/fxa/blob/88f6119cc75ae75bd26f87cf5e42dc71e95411d4/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js#L69) capability.

packages/fxa-content-server/app/scripts/views/mixins/signup-mixin.js
```diff
@@ -61,8 +61,8 @@ export default {
             !this.isCWTSOnSignupPasswordEnabled())
         ) {
           return this.navigate('choose_what_to_sync', {
-            account: account,
-            allowToDisableSync: this.relier.get('service') !== 'sync',
+            account,
```
vladikoff: 👍

packages/fxa-content-server/tests/functional/sync_v3_sign_up.js
```diff
@@ -53,6 +53,9 @@ registerSuite('Firefox Desktop Sync v3 signup', {
                 },
                 'fxaccounts:fxa_status': {
                   signedInUser: null,
+                  capabilities: {
+                    multiService: true,
```
vladikoff: thanks

vasilicatamas:
I noticed that CWTS section is no longer displayed in registration form on Fenix (Nighty) for a Mozilla account. Tested on both Prod[1.151.5] and Stage[1.152.0] environments. 

@shane-tomlinson is this an intended behavior?  

Screenshot_20191212-152349

vladikoff:
@vasilicatamas what about Fenix Release?

6a68:
Removing needs:uplift label as this shipped in 151.5: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/CHANGELOG.md#11515

vasilicatamas:
> @vasilicatamas what about Fenix Release?

CWTS section is successfully displayed in registration form on Fenix Nightly (200728) and Fenix Beta 79.0.0-beta9. Removing the needs:qa label.

PR #3609, opened 2019-12-09
fix(fenix): Do not use Sync Optional with Fenix

### LINKED ISSUE #3553: Do not use the sync-optional flow with Fenix

shane-tomlinson:
Add a new capability `syncOptional` to auth_brokers, only
enable this capability in fx_desktop_v3 *if* the `multiService`
capability is returned from the browser.

fixes #3553

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
@@ -591,6 +591,10 @@ const BaseAuthenticationBroker = Backbone.Model.extend({
      */
     supportsPairing: false,
     /**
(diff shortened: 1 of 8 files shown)


shane-tomlinson:
This is #3607 opened up against train-151 so we can uplift it to prod.

PR #3613, opened 2019-12-09
fix(buttons): Add `Not now` button back to sms page

### LINKED ISSUE #3597: We removed "start browsing" on CAD page
This issue was filed because the user thought that CAD was required:
https://github.com/mozilla/fxa/issues/3594

I was surprised because we have a start browsing button on that page. We even have an issue open to measure usage of the "start browsing" link:
https://github.com/mozilla/fxa/issues/3022

I just went through the flow and it seems like we've removed the link:
image

Let's restore the "Start browsing" link. 

Note: Martin Lapatka also brought this to my attention this week. He reported that he knew folks who thought that this was required.

CC @ryanfeeley

vbudhram:
Fixes #3597 

From https://github.com/mozilla/fxa/issues/3597#issuecomment-563440152, I opted to use `Not now`. Clicking the link will take you to the FxA product page in the same window.

Screen Shot 2019-12-09 at 5 31 26 PM

### DIFF
--- a/packages/fxa-content-server/tests/functional/send_sms.js
+++ b/packages/fxa-content-server/tests/functional/send_sms.js
@@ -192,6 +192,12 @@ const suite = {
         .then(testElementExists(selectors.SMS_SEND.HEADER));
     },
 
+    'not now': function() {
+      return this.remote
+        .then(openPage(SEND_SMS_URL, selectors.SMS_SEND.HEADER))
+        .then(testElementExists(selectors.SMS_SEND.LINK_NOT_NOW));
+    },
+
     'empty phone number': function() {
       return this.remote
         .then(openPage(SEND_SMS_URL, selectors.SMS_SEND.HEADER))
--- a/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
@@ -21,6 +21,7 @@
     </div>
     <div class="links">
       <a href="/sms/why" data-flow-event="link.why" class="left">{{#t}}Why sign in on another device?{{/t}}</a>
+      <a href='https://www.mozilla.org/firefox/accounts' class="btn-not-now" data-flow-event="link.not_now">{{#t}}Not now{{/t}}</a>
     </div>
     <aside class="child-view"></aside>
 </div>
--- a/packages/fxa-content-server/tests/functional/lib/selectors.js
+++ b/packages/fxa-content-server/tests/functional/lib/selectors.js
@@ -442,6 +442,7 @@ module.exports = {
     LINK_MARKETING: '.marketing-link',
     LINK_MARKETING_ANDROID: '.marketing-link-android',
     LINK_MARKETING_IOS: '.marketing-link-ios',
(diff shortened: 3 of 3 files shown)


packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -31,11 +33,19 @@ class SmsSendView extends FormView {
...
+    'click .btn-not-now': preventDefaultThen('gotoProductPage'),
+  });
+
   getAccount() {
     // TODO - remove the `|| ...` when done with development
     return this.model.get('account') || this.user.getSignedInAccount();
   }
 
+  gotoProductPage() {
+    this.window.location.href = SmsSendView.FXA_PRODUCT_PAGE_URL;
```
shane-tomlinson: Can you just set the href of the anchor element? The [ExternalLinksMixin](https://github.com/mozilla/fxa/blob/3bbffd84da7139af2b6d1593141a1171f7c528ab/packages/fxa-content-server/app/scripts/views/mixins/external-links-mixin.js) is added to BaseView, it'll take care of flushing metrics for you and you won't need this extra code. Direct assignment of `window.location.href` like this doesn't take care of flushing metrics.

packages/fxa-content-server/app/tests/spec/views/sms_send.js
```diff
@@ -158,6 +158,12 @@ describe('views/sms_send', () => {
         assert.lengthOf(view.$('.success'), 0);
       });
     });
+
+    it('renders `not now` links', () => {
+      return view.render().then(() => {
+        assert.lengthOf(view.$('.btn-not-now'), 1);
```
vbudhram: @shane-tomlinson if I do a `click()` on this element, it will redirect current window to the link. However, that will make all the next mocha tests fail since the page moved.

Is there a better way to test this? Maybe this test isn't needed?

shane-tomlinson: I'd just migrate this whole thing to the functional tests. Or stub out `navigateAway` before clicking.

shane-tomlinson: Maybe check the link is displayed in the functional tests but don't click it. We've had recent issues about tests failing because of external dependencies and minimizing those would be worthwhile.

packages/fxa-content-server/tests/functional/send_sms.js
```diff
@@ -192,6 +193,13 @@ const suite = {
         .then(testElementExists(selectors.SMS_SEND.HEADER));
     },
 
+    'not now': function() {
+      return this.remote
+        .then(openPage(SEND_SMS_URL, selectors.SMS_SEND.HEADER))
+        .then(click(selectors.SMS_SEND.LINK_NOT_NOW))
+        .then(testUrlInclude('https://www.mozilla.org/'));
```
vbudhram: Navigating to the site 301s to the `en-US` version, it made sense to just check first part of url.

packages/fxa-content-server/app/tests/spec/views/sms_send.js
vbudhram: Done!

Issue #3327, opened 2019-11-12
Unpublish fxa-shared from npm


6a68:
We've been using `fxa-shared` via relative paths in the monorepo instead of via the npm package for a while now (see https://github.com/mozilla/fxa/pull/2126).

Is there any reason `fxa-shared` should still be up on npm? If not, let's take it down.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-362)
┆Issue Number: FXA-362

dannycoates:
The work here is in updating any of our packages here to use the local fxa-shared and any dockerfile changes required.

meandavejustice:
once #3614 is merged and makes it through trains we can unpublish

meandavejustice:
package was marked `deprecated`

PR #3614, opened 2019-12-09
chore(fxa-profile-server): install fxa-shared directly in docker build

### LINKED ISSUE #3326: modify fxa-profile-server to install fxa-shared via relative path instead of npm package
follow method used here https://github.com/mozilla/fxa/pull/2126

we should unpublish the fxa-shared package once we no longer depend on it anywhere else

meandavejustice:
- fixes #3326


meandavejustice:
@shane-tomlinson We sure can! As soon as this makes it through the trains. We got #3327 to track it 👍

PR #3615, opened 2019-12-09
feat(event-broker): use proper hapi/sqs filtering

### LINKED ISSUE #3422: Filter PII from event broker Sentry reports
Let's do that

bbangert:
Because:

* The stock Sentry plugin has limitations on what details it filters
  from the error details.
* We've customized Hapi error reporting Sentry for auth-server and have
  not used those elsewhere.

This commit:

* Brings over the Hapi Sentry configuration used in auth-server that
  filters out additional PII such as uid's.
* Add's a new SQS Sentry configuration that capatures additional Sentry
  details for SQS processing failures.

Closes #3422

### DIFF
--- a/packages/fxa-auth-server/lib/sentry.js
+++ b/packages/fxa-auth-server/lib/sentry.js
@@ -73,7 +73,7 @@ function filterSentryEvent(event, hint) {
     }
   }
   if (event.tags && event.tags.url) {
-    event.tags.url = event.request.url.replace(TOKENREGEX, FILTERED);
+    event.tags.url = event.tags.url.replace(TOKENREGEX, FILTERED);
   }
   return event;
 }
--- a/packages/fxa-event-broker/lib/sentry.ts
+++ b/packages/fxa-event-broker/lib/sentry.ts
@@ -1,0 +1,180 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Hapi from '@hapi/hapi';
+import Hoek from '@hapi/hoek';
+import * as Sentry from '@sentry/node';
+import { SQS } from 'aws-sdk';
+import { Consumer } from 'sqs-consumer';
+import { PassThrough } from 'stream';
+
+// Matches uid, session, oauth and other common tokens which we would
+// prefer not to include in Sentry reports.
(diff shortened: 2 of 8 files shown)


bbangert:
A side-note, a bunch of this should likely make its way into fxa-shared as we need it in multiple projects. I didn't want to conflate that PR with this though.

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -73,7 +73,7 @@ function filterSentryEvent(event, hint) {
     }
   }
   if (event.tags && event.tags.url) {
-    event.tags.url = event.request.url.replace(TOKENREGEX, FILTERED);
+    event.tags.url = event.tags.url.replace(TOKENREGEX, FILTERED);
```
bbangert: I noticed during copy/paste that I clearly messed up in the auth-server here. 😞

meandavejustice:
> A side-note, a bunch of this should likely make its way into fxa-shared as we need it in multiple projects. I didn't want to conflate that PR with this though.

I have an issue for consolidating client side sentry filtering from `fxa-content-server` and `fxa-payment-server` into `fxa-shared` #3563, This could possibly piggyback on that issue

packages/fxa-auth-server/lib/sentry.js
shane-tomlinson: Will this be causing any auth-server errors right now?

packages/fxa-event-broker/lib/sentry.ts
```diff
@@ -0,0 +1,178 @@
...
+import Hapi from '@hapi/hapi';
+import Hoek from '@hapi/hoek';
+import * as Sentry from '@sentry/node';
+import { SQS } from 'aws-sdk';
+import { Consumer } from 'sqs-consumer';
+import { PassThrough } from 'stream';
+
+// Matches uid, session, oauth and other common tokens which we would
+// prefer not to include in Sentry reports.
+const TOKENREGEX = /[a-fA-F0-9]{32,}/gi;
```
shane-tomlinson: Do any email addresses make it to the event broker that we should also filter?

packages/fxa-auth-server/lib/sentry.js
bbangert: I don't think so, because the request always has a url on it. And I don't think we frequently attach tags with a url.

packages/fxa-event-broker/lib/sentry.ts
bbangert: They do! I added an email regex (yes, I know).

PR #3621, opened 2019-12-10
fix(metrics): Ensure metrics are flushed before redirecting to external links

### LINKED ISSUE #3619: Redirecting to external sites does not always flush metrics first
Noticed this while reviewing https://github.com/mozilla/fxa/pull/3613. There are a bunch of places where we set `window.location.href` or `navigator.location.href` to a URL without first flushing metrics, meaning there is a good chance we are losing metrics for these users.

We should almost never do a direct assignment on `window.location.href` or `navigator.location.href`, rather from within views we should call [this.navigateAway](https://github.com/mozilla/fxa/blob/3bbffd84da7139af2b6d1593141a1171f7c528ab/packages/fxa-content-server/app/scripts/views/base.js#L924) which ensures metrics are flushed before redirecting, or just use an href attribute of an anchor element which will do the same via the [ExternalLinksMixin](https://github.com/mozilla/fxa/blob/3bbffd84da7139af2b6d1593141a1171f7c528ab/packages/fxa-content-server/app/scripts/views/mixins/external-links-mixin.js) which is mixed in to all views.

shane-tomlinson:
Metrics were not always flushed before redirecting to external sites,
e.g., the product page on /*_confirmed and even sometimes when
redirecting back to an OAuth RP.

This ensures metrics are always flushed before setting the
href.

fixes #3619

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-redirect.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-redirect.js
@@ -99,8 +99,8 @@ export default BaseAuthenticationBroker.extend({
    * @returns {Promise}
    */
   sendOAuthResultToRelier(result) {
-    return Promise.resolve().then(() => {
-      var extraParams = {};
+    return this._metrics.flush().then(() => {
+      const extraParams = {};
       if (result.error) {
         extraParams.error = result.error;
       }
@@ -349,18 +349,18 @@ export default BaseAuthenticationBroker.extend({
   finishOAuthFlow(account, additionalResultData) {
     this.session.clear('oauth');
 
-    return Promise.resolve().then(() => {
-      // There are no ill side effects if the Original Tab Marker is
-      // cleared in the a tab other than the original. Always clear it just
-      // to make sure the bases are covered.
-      this.clearOriginalTabMarker();
-      return this.getOAuthResult(account).then(result => {
-        result = _.extend(result, additionalResultData);
-
-        return this._metrics.flush().then(() => {
-          return this.sendOAuthResultToRelier(result, account);
-        });
-      });
+    // There are no ill side effects if the Original Tab Marker is
+    // cleared in the a tab other than the original. Always clear it just
+    // to make sure the bases are covered.
+    this.clearOriginalTabMarker();
+    return this.getOAuthResult(account).then(result => {
+      return this.sendOAuthResultToRelier(
+        {
+          ...result,
+          ...additionalResultData,
+        },
+        account
+      );
     });
   },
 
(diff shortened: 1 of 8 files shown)


packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-redirect.js
```diff
@@ -99,8 +99,8 @@ export default BaseAuthenticationBroker.extend({
    * @returns {Promise}
    */
   sendOAuthResultToRelier(result) {
-    return Promise.resolve().then(() => {
-      var extraParams = {};
+    return this._metrics.flush().then(() => {
```
shane-tomlinson: This method was being called outside of `finishOAuthFlow` where metrics were originally flushed. Now flushing here to ensure all metrics are sent before redirecting.

packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-redirect.js
```diff
@@ -349,18 +349,18 @@ export default BaseAuthenticationBroker.extend({
...
-        return this._metrics.flush().then(() => {
-          return this.sendOAuthResultToRelier(result, account);
-        });
-      });
+    // There are no ill side effects if the Original Tab Marker is
+    // cleared in the a tab other than the original. Always clear it just
+    // to make sure the bases are covered.
+    this.clearOriginalTabMarker();
+    return this.getOAuthResult(account).then(result => {
+      return this.sendOAuthResultToRelier(
```
shane-tomlinson: `getOAuthResult` already returns a promise, not sure why the outer portion was wrapped in a promise.

packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
```diff
@@ -112,8 +112,9 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
     if (state) {
       result.state = state;
     }
-
-    return this.send(this.getCommand('OAUTH_LOGIN'), result);
+    return this._metrics.flush().then(() => {
```
shane-tomlinson: Similar to the oauth-redirect case, metrics weren't being flushed before the login message was being sent to the browser. If the browser closes the WebView before the metrics are flushed in the onload handler, then we'd lose those metrics. This ensures they are flushed before notifying the browser.

packages/fxa-content-server/app/scripts/views/mixins/external-links-mixin.js
```diff
@@ -37,7 +36,7 @@ export default {
 
     event.preventDefault();
 
-    return this._flushMetricsThenRedirect(event.currentTarget.href);
+    this.navigateAway(event.currentTarget.href);
```
shane-tomlinson: This is just a simplification, `_flushMetricsThenRedirect` had the same effect as `navigateAway`, no reason to have two methods that do the same thing.

PR #3622, opened 2019-12-10
chore(metrics): remove flow perf metrics from Payments

### LINKED ISSUE #3589: [Payments Server] Disable (and Remove?) Performance Metrics
Payments server is currently emitting performance metric events like those in Content server.  However, there are at least two problems:

 - They are not processed and saved to Redshift because the big lift that is #3072 
 - The calculated "network" time values are incorrect due to static content caching on Payments preventing a correct server side start time being included in every page

Since we are not getting any value out of these extra requests and log events, we should disable them.

chenba:
This patch removes flow based performance metrics on Payments server.  The
data from the logs are not present in Redshift due to the lack of a log
pipeline to ingest and process the events.  Furthermore, there was a bug
with the timing values.

This will be followed up with a switch to statsd, so that performance
metrics will be separated from flow events.

Fixes #3589 

@mozilla/fxa-devs r?


chenba:
@jbuck would like your blessing on this since I think you wanted Payments to emit the events even though we aren't processing them.

packages/fxa-payments-server/server/lib/server.test.js
```diff
@@ -2,53 +2,56 @@ const request = require('supertest');
...
-      .then(response => {
-        expect(response.statusCode).toStrictEqual(200);
-        expect(response.headers['content-type']).toStrictEqual(
-          'text/plain; charset=utf-8'
-        );
-        expect(response.text).toStrictEqual('Ok');
-
-        done();
-      });
+  test('__lbheartbeat__ should return as expected', () => {
```
meandavejustice: Why were these switched to create a promise instead of calling `done` provided by the testing suite? (just curious)

meandavejustice:
Looks good from a code quality perspective. I'll wait for jbuck to take a look before giving it the 👍

packages/fxa-payments-server/server/lib/server.test.js
chenba: Because eslint called by the pre-commit hook caught them.  They were not errors when linting from within Payments.

chenba: ref: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-test-callback.md

jbuck:
Yup, this is fine to remove

chenba:
@meandavejustice can I get a review 👍 ?  Thanks!

PR #3634, opened 2019-12-11
Set `service` tag for metrics events based on fxa_status response.

### LINKED ISSUE #3611: Not Tracking Service for form views of Sync decouple flows
See login form views on about:welcome: 
https://analytics.amplitude.com/mozilla-corp/chart/r7iiqzt

See registration form views on about:welcome: 
https://analytics.amplitude.com/mozilla-corp/chart/ln7e496 

We are not tracking a service property value for the Sync decouple flow form views. (reg and login)

CC @irrationalagent

rfk:
Because:

* Metrics events for the sync-optional flow currently have a missing
  `service` attribute.
* The sync-optional flow does not learn the client_id of the
  connecting application until it receives the `fxa_status` message.

This commit:

* Explicitly sets the `service` attribute on the metrics object
  when handling the `fxa_status` message response.

Fixes #3611.

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/metrics.js
+++ b/packages/fxa-content-server/app/scripts/lib/metrics.js
@@ -585,6 +585,17 @@ _.extend(Metrics.prototype, Backbone.Events, {
   },
 
   /**
+   * Set the `service` parameter to use for all future metrics.
+   * This is useful in cases where don't learn the appropriate
+   * service value until after the app has been initialized.
+   *
+   * @param {String} [service] The service identifier
+   */
+  setService(service) {
+    this._service = service || NOT_REPORTED_VALUE;
+  },
+
+  /**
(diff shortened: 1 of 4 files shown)


rfk:
There may be some more clever way to do this without having the broker know it has to update `service` in multiple places, but this is the simplest fix I could come up with.

rfk:
Mostly I just really wanted to ask @shane-tomlinson for one last review 😁

clouserw:
@LZoog Let's pull this in to Train 152 after it lands on master.  Thanks

clouserw:
@LZoog Let's pull this in to Train 152 after it lands on master.  Thanks

clouserw:
@LZoog Let's pull this in to Train 152 after it lands on master.  Thanks

clouserw:
@LZoog Let's pull this in to Train 152 after it lands on master.  Thanks

Issue #3213, opened 2019-10-31
50% of Firefox iOS Registrations don't see CWTS.


davismtl:
According to this chart, 50% of Firefox for iOS users don't see CWTS when registering to Sync:
https://analytics.amplitude.com/mozilla-corp/chart/new/prwip9u

This is substantially lower than all other OS that average about 15% drop off. (which is also alarming but less so)

When looking at registration rates, Firefox for iOS is not doing bad so this isn't preventing people from completing their registration. (good news)
https://analytics.amplitude.com/mozilla-corp/chart/new/at6k5ew

This leads me to believe there are two things potentially causing this anomoly:
- Users aren't seeing CWTS in some Firefox for iOS registration flows (I'm testing but haven't identified any yet)
- We are somehow not tracking the views.

In either case, we have a bug somewhere.

davismtl:
It does not seem to be a specific entrypoint:
https://analytics.amplitude.com/mozilla-corp/chart/new/24s5ggz

shane-tomlinson:
I wonder if we are sending the `login` notification to the browser before the metrics are fully flushed.

juliemccracken:
Alex: Let's think about how we might improve reliability of our metrics for the next year! There might be an Epic that we should open up for work like this.

PR #3636, opened 2019-12-11
fix(metrics): Flush metrics any time messages are sent to the browser

shane-tomlinson:
While writing docs on how to ensure metrics are flushed before
FxA closes, I realized we could probably automate this. Any
time we send a notice to the browser, flush metrics first.

issue #3213

@mozilla/fxa-devs - r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/mixins/channel.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/mixins/channel.js
@@ -28,6 +28,10 @@ function ensureActionReturnsPromise(action) {
 }
 
 const ChannelMixin = {
+  initialize(options = {}) {
+    this._metrics = options.metrics;
+  },
+
   /**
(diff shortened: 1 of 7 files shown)


packages/fxa-content-server/app/scripts/models/auth_brokers/mixins/channel.js
```diff
@@ -73,7 +77,7 @@ const ChannelMixin = {
     const channel = this.getChannel();
     const send = ensureActionReturnsPromise(channel.send.bind(channel));
 
-    return send(message, data);
+    return this._metrics.flush().then(() => send(message, data));
```
vbudhram: nit: We could update comment block to mention that metrics are flushed upon sending.

shane-tomlinson: Good suggestion, will do.

PR #3644, opened 2019-12-11
chore(refactor): Remove token code routes, js endpoints and docs

vbudhram:
Connects to #2427 

Removes most references to the older `token code` route. This was the first version of the code which was supposed to replace confirmation links. It was a randomly generated code with an expiration of 20mins.

This code has been replaced with a time based code that is used for verifying a login and session.

Let's see what breaks!


vbudhram:
Calling this ready!

Issue #3456, opened 2019-11-22
Modify SubscriptionRedirect to display user survey.


marty331:
This is a change request to modify the SubscriptionRedirect flow to include a user survey  and a link to download the newly purchased product.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-572)
┆Issue Number: FXA-572

clouserw:
@johngruen for adding details here

marty331:
![Subscription Ready, Survey](https://user-images.githubusercontent.com/663316/70001520-2199fd00-1523-11ea-8408-c3fce55da58a.png)
Adding mockup per @lmorchard

lmorchard:
We definitely need more details here.

Seems like a small thing to add a survey. But, also has potential to grow in scope - i.e. as we consider adding auth-server routes or a standalone service to guard a surveygizmo API secret. Also, considering that this is at the end of a process where a user has paid us money, it seems a sensitive spot for potential disruption.

Some general answers / requirements we need - if only for review, QA, and general care & feeding of the feature going forward:

* What's the deadline? (should it fit in one of our sprints?)
* How long will this survey live?
* Do we need to instrument this for any metrics?
* Will all users see this survey every time they are sent to download any product?
* What are the failure states and how should we message them? (i.e. what happens when loading the questions or submitting answers fails or is rate limited?)
* Have we done something like this elsewhere at Mozilla as an example? (On other projects, I've seen us *link* to an external survey. But have we embedded like this before? If so, how was that done & how did it work out?)
* Can we bend on the UX & styling if it simplifies integration? (e.g. use an embedded iframe survey hosted by surveygizmo vs implementing our own survey component & API proxies with an ops-managed API secret)
* If we consider embed via iframe
  * Seems secure, but are there concerns worth a review? (probably no worse than the iframe'd Stripe components we already use)
* If we consider including a 3rd party JS lib from surveygizmo on payments-server
  * does it need security review? (payments-server is intended in part as a kind of DMZ for 3rd-party Stripe JS - we're sensitive to adding more external resources to FxA in general)
* If we consider building our own web API proxy to guard the surveygizmo API key
  * How sensitive is the key? Would leaking it grant access to PII or altering surveys?
  * How is the key rotated and who is involved? (e.g. ops, whoever owns the surveygizmo account, anyone else?)
  * Should the key be rotated regularly?
* What are the constraints of the surveygizmo API and our account with them?
[...]

PR #3653, opened 2019-12-12
Implement subscribe survey

marty331:
Should resolve issue #3456

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.scss
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.scss
@@ -1,14 +1,20 @@
 .product-payment {
   .subscription-ready {
+    display: grid;
+    grid-template-rows: repeat(6, auto);
+
+    .subscription-message {
+      margin: -16px 0 5px;
+      padding-bottom: 20px;
+    }
 
     h2 {
-      color: #20123b;
       font-size: 24px;
       font-weight: 500;
+      margin: 0 0 8px;
     }
 
     p {
-      color: #0c0c0d;
       font-size: 16px;
       margin: 20px 100px;
     }
@@ -16,5 +22,23 @@
     a {
       font-size: 16px;
     }
+
+    .exp-message {
+      font-size: 16px;
+    }
+
+    .subscription-ready h2 {
+      margin: 0 0 8px;
+    }
+  }
+}
+
+.survey-frame {
+
+  iframe {
+    border: 0;
+    height: 520px;
+    overflow: hidden;
+    width: 550px;
   }
 }
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
@@ -16,7 +16,7 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
   const { product_id, product_name } = plan;
   const { webIconURL, downloadURL } = metadataFromPlan(plan);
   const {
-    config: { productRedirectURLs },
+    config: { productRedirectURLs, survey },
     navigateToUrl,
   } = useContext(AppContext);
 
@@ -24,26 +24,40 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
     downloadURL || productRedirectURLs[product_id] || defaultProductRedirectURL;
 
   useEffect(() => {
-    navigateToUrl(redirectUrl);
-  }, [navigateToUrl, redirectUrl]);
+    const handleIframeTask = (e: any) => {
+      if (e.data === 'submitted survey') {
+        setTimeout(() => {
+          navigateToUrl(redirectUrl);
+        }, 250);
+      }
+    };
+    window.addEventListener('message', handleIframeTask);
+    return () => window.removeEventListener('message', handleIframeTask);
+  }, [redirectUrl]);
 
   return (
     <div className="product-payment" data-testid="subscription-redirect">
       <div className="subscription-ready">
-        <h2>Your subscription is ready</h2>
-        <img
-          alt={product_name}
-          src={webIconURL || fpnImage}
-          width="96"
-          height="96"
-        />
-        <p>
-          Hang on for a moment while we send you to the{' '}
-          <span className="plan-name">{product_name}</span> download page.
-        </p>
-        <a href={redirectUrl}>
-          Click here if you're not automatically redirected
-        </a>
+        <div className="subscription-message">
+          <h2>Your subscription is ready</h2>
+          <div className="exp-message">
+            Please take a moment to tell us about your experience.
+          </div>
+        </div>
+        <hr />
+        <div className="survey-frame">
+          <iframe
+            sandbox="allow-scripts allow-forms"
+            scrolling="no"
+            src={
+              'http://www.surveygizmo.com/s3/5294819/VPN-Subscription?__no_style=true&env=' +
+              survey
+            }
+          ></iframe>
+        </div>
+        <div>
+          <a href={redirectUrl}>No thanks, just take me to my product.</a>
+        </div>
       </div>
     </div>
   );
--- a/packages/fxa-payments-server/server/config/development.json
+++ b/packages/fxa-payments-server/server/config/development.json
@@ -1,5 +1,6 @@
 {
   "csp": {
     "extraImgSrc": [ "https://placekitten.com" ]
-  }
+  },
+  "survey": "test_env"
 }
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -272,6 +272,12 @@ const conf = convict({
       format: 'url',
     },
   },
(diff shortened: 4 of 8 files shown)


packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -24,26 +24,36 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
...
 
   useEffect(() => {
-    navigateToUrl(redirectUrl);
-  }, [navigateToUrl, redirectUrl]);
+    window.addEventListener('message', handleIframeTask);
+  });
+
+  const handleIframeTask = (e: any) => {
+    if (e.data === 'submitted survey') {
+      console.log('survey: ' + e.data);
```
lmorchard: Should snip out this `console.log` before merge

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -24,26 +24,36 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
...
+      navigateToUrl(redirectUrl);
+    }
+  };
 
   return (
     <div className="product-payment" data-testid="subscription-redirect">
       <div className="subscription-ready">
-        <h2>Your subscription is ready</h2>
-        <img
```
lmorchard: This is [causing some test failures](https://circleci.com/gh/mozilla/fxa/127687?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link), now that the product icon & name are gone from the view.

Need to update / tweak these tests:

```
FAIL src/routes/Product/index.test.tsx
  routes/Product
    with product icon defined
      ✕ handles a successful payment submission as expected (113ms)
      ✕ redirects to product page if user is already subscribed (44ms)
    with default product icon
      ✕ handles a successful payment submission as expected (256ms)
      ✕ redirects to product page if user is already subscribed (43ms)

FAIL src/routes/Product/SubscriptionRedirect/index.test.tsx
  ✕ performs a redirect to the expected URL for local product (13ms)
  ✕ performs a redirect to the default URL for unknown product (15ms)
```

Removing icon & name seems consistent with the mock in #3456, but I wonder if we'll want to add them back later or conditionally?

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -24,26 +24,36 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
...
+          <h2>Your subscription is ready</h2>
+        </div>
+        <div className="formBreakTop breakBuffer"></div>
+        <div className="expMessage">
+          Please take a moment to tell us about your experience.
+        </div>
+        <div className="survey-frame">
+          <iframe
+            sandbox="allow-scripts allow-forms"
+            src="http://www.surveygizmo.com/s3/5294819/VPN-Subscription?__no_style=true"
```
lmorchard: This survey URL seems like something we'll want to make configurable someday, maybe? Maybe not worth worrying about right now, though.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -24,26 +24,36 @@ export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
     downloadURL || productRedirectURLs[product_id] || defaultProductRedirectURL;
 
   useEffect(() => {
-    navigateToUrl(redirectUrl);
-  }, [navigateToUrl, redirectUrl]);
+    window.addEventListener('message', handleIframeTask);
+  });
```
lmorchard: I think this `useEffect` could use a few tweaks. Something like this:
```
  useEffect(() => {
    const handleIframeTask = (e: any) => {
      if (e.data === 'submitted survey') {
        navigateToUrl(redirectUrl);
      }
    };
    window.addEventListener('message', handleIframeTask);
    return () => window.removeEventListener('message', handleIframeTask);
  }, [redirectUrl]);
```
This returns a cleanup function to remove the event listener on component un-mount, as well as limiting creation of the `handleIframeTask` function to only when `redirectUrl` changes. Would also be nice to have a more specific parameter type for `(e: any)` - but I'd have to dig around to figure out what that should be.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.scss
```diff
@@ -1,20 +1,61 @@
...
+        .formBreakTop {
+          border: solid 1px #d5d4d5;
+          height: 1px;
+          margin-bottom: 1.8em;
+          margin-left: -3em;
+          width: 115%;
+        }
+
+        .expMessage {
+          font-size: 16px !important;
```
LZoog: Why do we need `!important` here, can we override by being more specific with a selector?

Also, my SCSS linter complains with: "SelectorFormat: Selector `expMessage` should be written in lowercase with hyphens", can we change these to lower case dashes?

And, is there any way we can use `px` values here instead of `em`s? Since `em` depends on the parent element, we might get some weirdness here if styles on the parent element change. `px` gives us more control.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
marty331: I think it is safe to assume that we would want to add them back at some point, but for now they won't be there.  How do you want me to handle this?  Keep the code and comment it out or remove it?

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
marty331: I'm happy to do that if you give me guidance on how to do configurable variables in this app.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
marty331: OK I made that change, will be in my next commit.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.scss
marty331: I made these changes and also changed all the other class names that were camel cased.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
marty331: done.

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
marty331: done.

johngruen:
One other technical change...can we add a query param that specifies if the environment is prod so like `?prod=true`

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
johngruen: Let's do this as a follow up...I can file a ticket

bbangert:
Just a note before this gets merged, the commit should be written using the [fxa guidelines for commit messages](https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#git-commit-guidelines) and the commit should be GPG signed.

lmorchard:
> can we add a query param that specifies if the environment is prod so like ?prod=true

@johngruen Add the query param where? And only for prod? Or also stage / dev?

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
lmorchard: We shouldn't leave code commented out - we can always get it back from previous commits. Might as well remove the bits that are failing. (i.e. not the entire tests, just the parts that depend on product name & icon being available)

marty331: OK I removed the tests that were causing error.

marty331:
> Just a note before this gets merged, the commit should be written using the [fxa guidelines for commit messages](https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#git-commit-guidelines) and the commit should be GPG signed.

OK I followed the guide and GPG signed my commit.

LZoog:
LGTM, but will defer to @lmorchard for r+ since he had a lot more change requests than I did

johngruen:
@lmorchard @marty331 so a little more clarity about the query parameter. I would like it the survey link to include a param indicating if it is on prod or not so that we can filter out production test results from any other survey responses from staging, dev or local environments. 

To @lmorchard point, it might be better to just have an `env=${env}` param instead of just a param for prod. That way we can verify all the survey filtering is working on dev and stage.

marty331:
@johngruen @lmorchard please provide instructions on how to implement environment variables in this code base and who they should be delivered to for loading.

PR #3655, opened 2019-12-12
refactor(api): prefer async/await in lib/routes/account.js

chenba:
Part of #2305 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -140,51 +140,48 @@ module.exports = (
           planId,
         } = await request.app.metricsContext;
 
-        return customs
-          .check(request, email, 'accountCreate')
-          .then(deleteAccountIfUnverified)
-          .then(setMetricsFlowCompleteSignal)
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -140,51 +140,48 @@ module.exports = (
...
-                if (secondaryEmailRecord.isVerified) {
-                  throw error.verifiedSecondaryEmailAlreadyExists();
-                }
-
-                return db.deleteEmail(
-                  secondaryEmailRecord.uid,
-                  secondaryEmailRecord.email
-                );
+        await customs.check(request, email, 'accountCreate');
+        await deleteAccountIfUnverified();
```
lmorchard: Kind of wish we'd un-wrap & inline all these functions along the way to await-ifying & simplifying things.

But, on the other hand, I guess this kind of serves as a table-of-contents for the functionality. Also makes the conversion safer and smaller.

So, I'm mainly thinking out loud and this is not even a nit.

PR #3657, opened 2019-12-12
chore(node): updated node to v12

### LINKED ISSUE #957: Support Node 12 to prepare for its LTS
[Node 12 is now the "current" version](https://nodejs.org/en/blog/release/v12.0.0/). It is [scheduled to go into LTS on Oct 22nd](https://nodejs.org/en/about/releases/).

We could start ensuring our servers support both v10 and v12 to speed along the upgrade once 12 goes to LTS.

dannycoates:
closes #957


packages/fxa-auth-server/test/local/routes/utils/clients.js
```diff
@@ -126,19 +126,6 @@ describe('clientUtils.formatLocation', () => {
     const args = log.warn.args[0];
     assert.lengthOf(args, 2);
     assert.equal(args[0], 'attached-clients.formatLocation.warning');
-    assert.deepEqual(args[1], {
-      err:
-        "Cannot find module 'cldr-localenames-full/main/wibble/territories.json'",
```
dannycoates: error string changed in 12 to include a partial stack trace. I don't really care what this log line says 🤷‍♂️

PR #3658, opened 2019-12-12
chore(refactor): Add await to `request.app.devices` where needed

vbudhram:
From https://github.com/mozilla/fxa/pull/3639#discussion_r357219474, looks like the `request.app.devices` was not properly being awaited on in the emails routes. This fixes that and removes the bluebird dependency in db.js


dannycoates:
do we need the `await` on line 136 anymore?

vbudhram:
> do we need the await on line 136 anymore?

Nope, I'll push update

PR #3662, opened 2019-12-12
refactor(auth-server): asyncify lib/tokens

bbangert:
Issue #2305

### DIFF
--- a/packages/fxa-auth-server/lib/tokens/key_fetch_token.js
+++ b/packages/fxa-auth-server/lib/tokens/key_fetch_token.js
@@ -5,7 +5,6 @@
 'use strict';
 
 const inherits = require('util').inherits;
-const P = require('../promise');
 
 module.exports = function(log, Token) {
   function KeyFetchToken(keys, details) {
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/tokens/key_fetch_token.js
```diff
@@ -21,21 +20,17 @@ module.exports = function(log, Token) {
...
+    return token;
   };
 
-  KeyFetchToken.fromId = function(id, details) {
+  KeyFetchToken.fromId = async function(id, details) {
     log.trace('KeyFetchToken.fromId');
-    return P.resolve(
-      new KeyFetchToken({ id, authKey: details.authKey }, details)
-    );
+    return new KeyFetchToken({ id, authKey: details.authKey }, details);
```
chenba: Unnecessary `async` since I don't see an `await`.

bbangert: This function previously returned a Promise, so I had to make the function `async` so that this would match the API. Am I missing something?

chenba: No, you are not missing anything.  There are probably still `.then`s chaining off of it.

r+

Issue #2350, opened 2019-08-28
Create Recovery Code check screen/mixin


6a68:
Maybe we'll have fewer lockouts if we force users to enter a recovery code just after they view their recovery codes.

This would be something to add to both the settings-based 2FA setup flow and the login-based 2FA setup flow.

Refs #2024

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-310)
┆Priority: Medium
┆Issue Number: FXA-310

PR #3665, opened 2019-12-12
refactor(auth-server): asyncify utils/otp and utils/signup

bbangert:
Issue #2350

### DIFF
--- a/packages/fxa-auth-server/lib/routes/utils/otp.js
+++ b/packages/fxa-auth-server/lib/routes/utils/otp.js
@@ -16,21 +16,18 @@ module.exports = (log, config, db) => {
      * @param account
      * @returns boolean
      */
-    hasTotpToken(account) {
+    async hasTotpToken(account) {
       const { uid } = account;
-      return db.totpToken(uid).then(
-        result => {
-          if (result && result.verified && result.enabled) {
-            return true;
-          }
-        },
-        err => {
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/utils/otp.js
```diff
@@ -16,21 +16,18 @@ module.exports = (log, config, db) => {
      * @param account
      * @returns boolean
      */
-    hasTotpToken(account) {
+    async hasTotpToken(account) {
       const { uid } = account;
-      return db.totpToken(uid).then(
-        result => {
-          if (result && result.verified && result.enabled) {
-            return true;
```
bbangert: Just noticed that the way this is written means it could return undefined if the result here doesn't satisfy this condition.

packages/fxa-auth-server/lib/routes/utils/otp.js
```diff
@@ -16,21 +16,18 @@ module.exports = (log, config, db) => {
...
+      try {
+        result = await db.totpToken(uid);
+      } catch (err) {
+        if (err.errno === errors.ERRNO.TOTP_TOKEN_NOT_FOUND) {
+          return false;
         }
-      );
+        throw err;
+      }
+      return result && result.verified && result.enabled;
```
dannycoates: for a definite boolean result it should be `!!(result && result.verified && result.enabled)` but "truthy" has been working so far so up to you if you'd like to change it

PR #3667, opened 2019-12-13
chore(test): Default Fx UA string to Fx 60

shane-tomlinson:
Firefox 60 is our minimum officially supported version, this updates the
default UA string from Firefox 40 to 60.

Not attached to an issue


shane-tomlinson:
Woah, so many failure, this will require https://github.com/mozilla/fxa/pull/3579 to have any hope of passing at all.

shane-tomlinson:
This breaks far too many tests and I am not going to fix them all. Closing.

PR #3668, opened 2019-12-13
feat(signin): Cached signin w/ unverified account uses codes now.

### LINKED ISSUE #3649: Convert cached signin with an unverified account from links to codes
If a user signs in with a cached unverified account, we send them to the legacy "confirm signup" screen.  We should send them to the "enter a code" screen instead.

Ref #3648

shane-tomlinson:
Before this signing in to an OAuth RP with a cached unverified account
would send the user to /confirm, which would have resent verification
links.

This converts that flow to use codes instead.

fixes #3649


shane-tomlinson:
@vbudhram - r?

PR #3676, opened 2019-12-13
Train 152.1 point release -> master

LZoog:


LZoog:
Hmm, content server CI tests have failed 3 times with "intall node 10 to continue installation" (yes, with that typo) but they pass in `fxa-private`. Going to trigger one more time, I don't think the recently merged node 12 upgrade is related...

chenba:
@dannycoates looks like the node version in fxa-circleci is causing some headache for @LZoog 

Easiest fix is probably just cherry-pick the node upgrade commit(s) in?

PR #3678, opened 2019-12-13
chore(sentry): wrap dns query errors by error code

### LINKED ISSUE #3572: [content-server] Group Email Domain DNS Record Validation Server Errors by Error Code
There are 22 potential types of errors in the email domain validation route that would throw a 500 and get logged by Sentry.  We should catch and re-throw these errors so that they are grouped by their error codes.  We know at least one of these errors, `SERVFAIL`, includes the queried domain in its message.  This unncessarily creates many unique errors in Sentry.

The list of errors are list on [Node's docs for `dns`](https://nodejs.org/docs/latest-v10.x/api/dns.html#dns_error_codes).  The `validate-email-domain` route is currently handling `dns.NODATA` and `dns.NOTFOUND`.  The error code is simply the `code` property in the error object.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-55)
┆Issue Number: FXA-55

chenba:
This patch wraps errors from DNS queries in the email domain MX
record validation route by their error codes.  The main goal of this
exercise is to reduce the number of unique errors reported to Sentry.

Fixes #3572

@mozilla/fxa-devs  r?

### DIFF
--- a/packages/fxa-content-server/package.json
+++ b/packages/fxa-content-server/package.json
@@ -123,6 +123,7 @@
         "ua-parser-js": "git://github.com/vladikoff/ua-parser-js.git#fxa-version",
         "uglifyjs-webpack-plugin": "1.2.7",
         "underscore": "1.8.3",
+        "verror": "1.10.0",
         "webpack": "^4.41.2",
         "webpack-cli": "^3.3.10"
     },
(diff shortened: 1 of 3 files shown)


chenba:
@meandavejustice what's the status on updating Sentry on content server?

chenba:
Filed #3726 for the `@sentry/node` upgrade.

PR #3711, opened 2019-12-17
feat(support): add location to support ticket

### LINKED ISSUE #2675: [Support] Track support form request originating region
For use by support agents, we should send in a field to Zendesk that contains the users originating IP. This can be used later depending on the support request to verify the user is accessing our service from the same location they typically do.

This may require our Support team to setup a new field to store this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-388)
┆Priority: Medium
┆Issue Number: FXA-388

chenba:
This patch adds the IP based location of the support form request into
the support ticket.  The location info might be helpful to the support
staff.

Fixes #2675 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/routes/support.js
+++ b/packages/fxa-auth-server/lib/routes/support.js
@@ -66,11 +66,18 @@ module.exports = (log, db, config, customs, zendeskClient) => {
       handler: async function(request) {
         log.begin('support.ticket', request);
         const { uid, email } = await handleAuth(request.auth, true);
+        const { location } = request.app.geo;
         await customs.check(request, email, 'supportRequest');
         let subject = `${request.payload.topic} for ${request.payload.plan}`;
         if (request.payload.subject) {
           subject = subject.concat(': ', request.payload.subject);
         }
+        const {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/support.js
```diff
@@ -79,6 +80,7 @@ module.exports = (log, db, config, customs, zendeskClient) => {
             email,
             name: email,
           },
+          location,
```
bbangert: The Zendesk API accepts this and stores it somewhere user visible?

chenba: There's discussion regarding the new Zendesk field over at https://github.com/mozilla/fxa/issues/2675#issuecomment-566730022.

chenba: @jackiemunroe helped with the custom fields on the Zendesk side, and changes have been made: https://github.com/mozilla/fxa/pull/3711/files#diff-c01ed3e5dd5108a3df8ed8fc6bed2383R89-R94

Issue #3704, opened 2019-12-17
convert each component and route in payment server to use fluent l10n


meandavejustice:
from `src/components`:
- [x] AlertBar
- [x] DialogMessage
- [x] FetchErrorDialogMessage
- [x] LoadingSpinner
- [x] Tooltip
- [x] LoadingOverlay
- [x] PaymentForm
- [x] PaymentLegalBlurb
- [x] Portal
- [ ] fields
- [x] AppLayout
- [x] ErrorMessage
- [x] CloseIcon

from `/src/routes/Product`:
- [x] PlanDetails
- [x] SubscriptionCreate
- [ ] SubscriptionUpgrade
- [x] index
- [x] ProfileBanner
- [x] SubscriptionRedirect

from `/src/routes/Subscriptions`:
- [x] PaymentUpdateForm
- [x] Reactivate
- [x] SubscriptionItem
- [x] index

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-11)
┆Issue Number: FXA-11

LZoog:
Hey @meandavejustice, I know we closed this 3 days ago, but I think we're still lacking translations in/for at least:
- `PaymentUpdateForm`
- `SubscriptionItem`
- `routes/Subscriptions/index`
- `AccountActivatedBanner`
- `Product/SubscriptionCreate/index`
- 3x `aria-label`s (in `components/DialogMessage`, `routes/Product/SubscriptionUpgrade`, and `routes/Subscriptions`

LZoog:
Also, I wonder if work completed for this mean we can also close #1089 or are the date formats and "edgecases" (per Les and Wil's comments) keeping that one open?

meandavejustice:
@LZoog Thank you, reopening to check these out

PR #3720, opened 2019-12-18
chore(payments-server): add fluent to payments server

### LINKED ISSUE #3610: Add fluent dependency to payments server
Let's start getting L10n going

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-37)
┆Priority: Medium
┆Issue Number: FXA-37

meandavejustice:
fixes #3610

This is initial commit of this effort. So far only things handled
by fluent are error codes and the document title.

The rest will be addressed in separate commits per component
in issue #3704


packages/fxa-payments-server/package.json
```diff
@@ -62,17 +64,28 @@
...
+    "async-wait-until": "^1.2.4",
     "audit-filter": "^0.5.0",
     "babel-eslint": "^10.0.3",
     "babel-loader": "^8.0.6",
     "browserslist": "^4.7.2",
     "caniuse-lite": "^1.0.30001010",
     "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
+    "chai-enzyme": "^1.0.0-beta.1",
+    "enzyme": "3.7.0",
```
lmorchard: I don't think we want to add enzyme & chai - we're already using jest & react-testing-library as a framework here

meandavejustice: I agree, pulled the l10n and l10.test.js files from `lockwise-addon`. Wanted to ensure it all worked before converting to jest/react-testing-library

meandavejustice:
@lmorchard got rid of the extra deps

packages/fxa-payments-server/src/App.tsx
```diff
@@ -55,39 +58,48 @@ export const App = ({
...
-            </Router>
-          </ReduxProvider>
-        </StripeProvider>
-      </AppErrorBoundary>
+      <AppLocalizationProvider
+        userLocales={navigator.languages}
+        bundles={['main']}
+      >
+        <Localized id="document" attrs={{ title: true }}>
+          <DocumentTitle title="fIREFOx aCCOUNTs pAYMENTs">
```
LZoog: Was `fIREFOx aCCOUNTs pAYMENTs` to double check this was working? If I saw this while working I'd question its purpose.

Can we remove the `<title>` piece from `public/index.html`?

Also, are there plans for docs? 😁

meandavejustice: `fIREFOx aCCOUNTs pAYMENTs` style is what we used on lockwise. It's so that we can easily see if a translation is broken in testing before it goes live.

I left the tile in `public/index.html` so that we have something while the initial page load happens.

I should definitely add docs. Though some things may change as I updated all of the other components, so I should probably add after all those are done

clouserw: Can you expand on the capitalization style to test l10n?  I'd expect that to be done with a pseudolocale and not something that a user might see.

meandavejustice: From what I remember for the decision in lockwise, we want to always be able to catch anything that wouldn't translate before it gets to a user. Something noticeable like this, ensures in the QA process that it shouldn't make it to the user. The reasoning was basically, if non translated text makes it to the user, its wrong either way. Since most of us will be testing in english it'll be easier for these to slip past us

clouserw: If a translation changes, or there's a syntax error when a localizer checks the file in, it falls back to this string, right?  Shouldn't this string be a failsafe?

meandavejustice: Reverted to english text for these strings. We can use pseudolocalization to track translations in the future. Filed https://github.com/mozilla/fxa/issues/3756

packages/fxa-payments-server/public/locales/fr-FR/main.ftl
```diff
@@ -0,0 +1,6 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+document =
+  .title = Enregistrements Lockbox
```
clouserw: Why is fr-FR included but no other locales?

Even though this string will be overwritten by pontoon, it seems out of place in this patch.

Which brings up another point -- we probably need to ping the pontoon folks to add this new server.

meandavejustice: I included this just for testing, thought i would be easier to review. I'll remove the patch for now

packages/fxa-payments-server/src/lib/l10n.js
```diff
@@ -0,0 +1,115 @@
...
+      const cx = new FluentBundle(locale);
+      for (const i of mergedBundle[locale]) {
+        cx.addMessages(i);
+      }
+      yield cx;
+    }
+  };
+}
+
+export default class AppLocalizationProvider extends Component {
```
LZoog: Do you prefer `AppLocalizationProvider` to live in this file? It seems like it could be moved into its own file, we could use TS types instead of `prop-types`, and we could potentially use a hook instead of `componentDidMount`. Would this be worth filing a follow-up issue for or was it done this way for a particular reason?

(Great work on this BTW!)

packages/fxa-payments-server/src/lib/l10n.test.js
```diff
@@ -0,0 +1,138 @@
...
+import fetchMock from 'fetch-mock';
+import AppLocalizationProvider from './l10n';
+
+
+describe('<AppLocalizationProvider/>', () => {
+  const locales = ['en-US', 'es-ES'];
+  const bundles = ['greetings', 'farewells'];
+  function waitUntilTranslated() {
+    return waitUntil(() => {
+      return AppLocalizationProvider.prototype.render.callCount === 2;
```
LZoog: Super minor, but can we get rid of `{ return` and simply `=> AppLocalizationProvider...`?

packages/fxa-payments-server/src/lib/l10n.js
meandavejustice: I happy to move this into a separate file and use TS types. Filed https://github.com/mozilla/fxa/issues/3762

packages/fxa-payments-server/src/lib/l10n.test.js
meandavejustice: fixed

PR #3724, opened 2019-12-18
feat(content-server): Improve password advice functionality

### LINKED ISSUE #607: Improve styling of password advice
I started my own PR here, but didn’t find an elegant way to fix this, and didn't want to just hack it in. The password advice list items should only be 300 font weight, with 21px margin below the first section (as opposed to 7px). Reads more easily, don't you think @shane-tomlinson ?

# Current
screen shot 2018-08-13 at 1 46 34 pm

# Proposed
screen shot 2018-08-13 at 1 47 47 pm

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-743)
┆Priority: Medium
┆Issue Number: FXA-743

LZoog:
fixes #607 

It was determined the issue could be closed out with the following features:
- Update copy + key image for “repeat password": taken care of in #3235
- Make “repeat password” popup occur in mobile: I'm not sure if this was worked on in another issue, but I'm seeing the popup now. I tested in Safari emulator and XCode iOS. 
- Persist first password tooltip until onblur: this PR✅ 
- Return focus to last focused password field after 👁 is toggled on: this PR✅

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
+++ b/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
@@ -18,10 +18,6 @@ import OneVisibleOfTypeMixin from '../mixins/one-visible-of-type-mixin';
 import Template from '../../templates/partial/password-strength-balloon.mustache';
 import XSS from '../../lib/xss';
 
-// Allow the balloon to stay visible for a bit so that
-// the user can see all the criteria were met.
-const DELAY_BEFORE_HIDE_MS = 750;
-
 const DELAY_BEFORE_HIDE_BALLOON_EL_MS = 500;
 
 const PASSWORD_STRENGTH_BALLOON_SELECTOR = '.password-strength-balloon';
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
```diff
@@ -33,9 +29,8 @@ class PasswordStrengthBalloonView extends BaseView {
   template = Template;
 
   initialize(config = {}) {
-    this.delayBeforeHideMS = config.delayBeforeHideMS || DELAY_BEFORE_HIDE_MS;
-
     this.listenTo(this.model, 'change:password', this.update);
+    this.listenTo(this.model, 'change:inputFocus', this.shouldHide);
```
LZoog: I needed a way to consume an event in `password_strength_balloon.js` that's set in `password_with_strength_balloon.js` because I haven't seen where we set `events` in a class and I thought it made for better separation of concerns anyway. If setting this on `this.model` isn't a good approach I'm happy to change it in favor of a better suggestion.

packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
```diff
@@ -18,10 +18,6 @@ import OneVisibleOfTypeMixin from '../mixins/one-visible-of-type-mixin';
 import Template from '../../templates/partial/password-strength-balloon.mustache';
 import XSS from '../../lib/xss';
 
-// Allow the balloon to stay visible for a bit so that
-// the user can see all the criteria were met.
-const DELAY_BEFORE_HIDE_MS = 750;
```
LZoog: We no longer need this and related code since the balloon is now hidden when the user changes focus instead of waiting 750ms.

LZoog:
Hey @vbudhram, since you've worked on this screen can I bother you for a review here?

PR #3729, opened 2019-12-19
chore(metrics): up the default auth server StatsD sampling rate to 1

### LINKED ISSUE #3696: Up the StatsD Sampling Rate in Auth Server
This is a config change, which can be done by updating teh default value in `config/index.js` or through an env var.

The issue here is mostly a request for comment from Ops.

When StatsD was proposed for the auth-server, Shane expressed concern because of some performance issues from back in the Datadog days.  As a result I conservatively set the default sampling rate to 0.1.

Now that we've been collecting stats for a while and have not experienced any related performance issues, I think we could increase the sampling rate.

@jbuck @jrgm What's your take on this?  How high can we go?

cc @clouserw

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-13)
┆Priority: Medium
┆Issue Number: FXA-13

chenba:
Fixes #3696

### DIFF
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -1162,7 +1162,7 @@ const conf = convict({
     sampleRate: {
       doc: 'Sampling rate for StatsD',
       format: Number,
-      default: 0.1,
+      default: 1,
       env: 'STATSD_SAMPLE_RATE',
     },
     maxBufferSize: {


clouserw:
Will @jbuck need to change an environment variable or is there not one set?

chenba:
I think we've been using the default, so this PR changes that.  But could set an environment variable (instead or in addition).

clouserw:
Ok, this is fine.  I just want @jbuck to know what it's called in case things *do* blow up :+1:

jbuck:
I added a PR so we can turn the dial as needed: https://github.com/mozilla-services/cloudops-deployment/pull/3589

PR #3731, opened 2019-12-19
chore(build): [blocked by #3828] Revert allowing Node 10 in check.sh, update docs

LZoog:
Since the CI image appears to pull from `master`, we had to allow Node 10 as an option for the train 152.1 point release which was still on Node 10.

This PR should be merged before the Node 12 upgrade is placed on a release, _after_ any other point releases.

I left these as separate commits for the history, I can squash them if preferred. 🤷‍♀️

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If you get an `error` status for any of the servers please verify that you insta
 
 > Required developer dependencies:
 > [Git](http://git-scm.com/book/en/v2/Getting-Started-Installing-Git),
-> [node.js **10+** with npm 6](http://nodejs.org/),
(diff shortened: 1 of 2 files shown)


LZoog:
We may actually want hold on the merge a little longer, we had an issue with the profile-server related to Node 12. See https://github.com/mozilla/fxa/pull/3755/commits/0a4371212303dfabf8b32da8bddb3f517ddf7c4d

meandavejustice:
@LZoog This may be good to go now that #3828 is fixed

LZoog:
@meandavejustice thanks for the ping, switched from draft to ready for review. This should be good to merge.

PR #3748, opened 2019-12-21
feat(totp): Require recovery code confirmation before enabling TOTP

dannycoates:


packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -170,39 +191,35 @@ module.exports = (log, db, mailer, customs, config) => {
...
           throw errors.unverifiedSession();
         }
 
         try {
           const token = await db.totpToken(sessionToken.uid);
-
-          // If the token is not verified, lets delete it and report that
-          // it doesn't exist. This will help prevent some edge
-          // cases where the user started creating a token but never completed.
-          if (!token.verified) {
```
vbudhram: Nice work on removing this edge case!

packages/fxa-content-server/app/scripts/lib/crypto/totp.js
```diff
@@ -0,0 +1,51 @@
...
+
+function trimOrPad(num, digits) {
+  const str = num.toString().substr(-digits);
+  if (str.length === digits) {
+    return str;
+  }
+  return new Array(digits - str.length + 1).join('0') + str;
+}
+
+export function getCode(secret, digits = 6, now = Date.now()) {
```
vbudhram: This kind of makes me wonder if we should replace [otplib](https://github.com/yeojz/otplib) in auth-server with this implementation. 

Either way, this is 💯

packages/fxa-content-server/app/scripts/lib/crypto/totp.js
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import base32Decode from 'base32-decode';
```
vbudhram: It is probably worthwhile to add some documentation for future readers why this exist. You could also mention the [configuration](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/config/index.js#L1442) in auth-server.

packages/fxa-content-server/app/scripts/views/settings/recovery_codes.js
```diff
@@ -39,7 +43,50 @@ const View = FormView.extend({
...
+      return getCode(totpSecret)
+        .then(code => {
+          const account = this.getSignedInAccount();
+          return account.verifyTotpCode(code, this.relier.get('service'));
+        })
+        .then(() =>
+          this.displaySuccess(t('Two-step authentication enabled'), {})
+        )
+        .then(done);
+      // todo catch
```
vbudhram: Is this still a todo?

packages/fxa-content-server/app/scripts/views/settings/two_step_authentication.js
```diff
@@ -63,10 +64,11 @@ const View = FormView.extend({
     return code.replace(/(\w{4})/g, '$1 ');
   },
 
-  _showRecoveryCodes(recoveryCodes) {
-    this.model.set('recoveryCodes', recoveryCodes);
+  _showRecoveryCodes(recoveryCodes, totpSecret) {
+    // this.model.set('recoveryCodes', recoveryCodes);
```
vbudhram: Extra comment

PR #3749, opened 2019-12-22
App layout l10n

meandavejustice:
should be merged after #3720

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -2,6 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+## branding
+project-brand = Firefox Accounts
+
 document =
   .title = Firefox Accounts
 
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/src/App.tsx
```diff
@@ -55,39 +58,48 @@ export const App = ({
...
-            </Router>
-          </ReduxProvider>
-        </StripeProvider>
-      </AppErrorBoundary>
+      <AppLocalizationProvider
+        userLocales={navigator.languages}
+        bundles={['main']}
+      >
+        <Localized id="document" attrs={{ title: true }}>
+          <DocumentTitle title="fIREFOx aCCOUNTs pAYMENTs">
```
vbudhram: Is this here so that you can ensure that fluent overrides the value?

packages/fxa-payments-server/src/lib/l10n.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+async function fetchMessages(baseDir, locale, bundle) {
+  try {
+    const response = await fetch(`${baseDir}/${locale}/${bundle}.ftl`);
+    const messages = await response.text();
+
+    return messages;
+  } catch (e) {
+    // We couldn't fetch any strings; just return nothing and fluent will fall
```
vbudhram: 👍🏽

packages/fxa-payments-server/src/App.tsx
meandavejustice: Correct, this is the method we used through out https://github.com/mozilla-lockwise/lockwise-addon

meandavejustice: Removed this formatting

meandavejustice:
@vbudhram Not seeing the "middleware is not a function" error locally

meandavejustice:
rebased to squash commits

PR #3758, opened 2019-12-25
fix(css): improve line-height of input placeholders in webkit browsers

### LINKED ISSUE #1743: Placeholder text on iOS too high
Not sure why, but the field placeholder text on iOS is too (damned) high. Ideally it can align with the text without displacing anything on other platforms.
placeholder

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-193)
┆Priority: Medium
┆Issue Number: FXA-193

jodyheavener:
Closes #1743

This is my attempt at fixing the janky line-height on input fields in iOS. Because the selector being modified targets Webkit browsers we also have to take desktop Safari into consideration.

One thing I noticed adjacent to this issue was that the placeholder psuedo selector had a font-size of 17px, where as far as I can see the value type for these fields are font-size 16px. Removing this helped with consistency, but I am also unaware of why it was added so any insight here is valuable.

Some previews:

iOS

Safari

You'll notice that the "after" for Safari is still off by a pixel or so, but the iOS fix looks a noticeably better. My thinking is that iOS should take priority here as we'll encounter more instances of this jankiness in Firefox iOS (which is just iOS Safari, whereas someone who has an FxA would probably use true Firefox on desktop - speculation on my part).

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_input-row.scss
+++ b/packages/fxa-content-server/app/styles/modules/_input-row.scss
@@ -13,7 +13,6 @@
 
     &::placeholder {
       color: $input-placeholder-color !important;
-      font-size: 17px;
       font-weight: $font-weight-body-20;
       opacity: 1;
     }
@@ -21,7 +20,7 @@
     &::-webkit-input-placeholder {
       // Fixes a problem in Safari/Fx for iOS where placeholder
       // text is not vertically centered and becomes cut-off.
-      line-height: 1 !important;
+      line-height: 1.3 !important;
     }
 
     // disable arrows on HTML5 number inputs


jodyheavener:
> Regarding the 17px placeholder text vs 16px input - this was very likely a mistake.

Ah, ok cool. Thanks for the confirmation on that, and the review :)

LZoog:
Hey @jodyheavener, I just realized that your commits don't appear to be GPG signed. Sorry you've put in several PRs, we have this as a requirement for security reasons. See the "Sending Pull Requests" section [here](https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md). It looks like you can sign old commits manually though so hopefully it's not too much trouble.

Let me know if you run into any issues.

jodyheavener:
@LZoog Hi! Sorry about that. I went and tried to sign the one commit here using `git commit -S --amend`, but the push failed with a [strange error](https://gist.github.com/jodyheavener/fd973921a79eae92d70e85def786893d). Any ideas? Googling the issue makes me think it's an NPM thing?

LZoog:
@jodyheavener No prob, it's easily missed! You can look at the log file it generated for more details, but every now and then I see a similar error and deleting `node_modules`/rerunning `npm i` usually fixes the error. You can add a `--no-verify` flag to get around it temporarily if it persists, which is not ideal but the changes in your PRs I've reviewed so far should be fine without the pre-push hook stuff.

PR #3761, opened 2019-12-26
refactor(settings): change actions in 2fa, recovery, and avatar modals from anchors to buttons

### LINKED ISSUE #624: Use buttons instead of links
From https://github.com/mozilla/fxa-content-server/pull/6239/files#r192063057 and https://github.com/mozilla/fxa-bugzilla-mirror/issues/504, there some places where we use links that do not navigate to anywhere. In these scenarios, we should probably be using a button.

Ideally, we want the buttons to be styled like this
screen shot 2018-06-05 at 11 49 35 am

where there is an image and the text is centered below it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-755)
┆Priority: Medium
┆Issue Number: FXA-755

jodyheavener:
Closes #624

Hey all. I decided to take a crack at this. The issue was pretty straightforward, but I'm going to break my changes down a bit.

- First and foremost, I've updated the actions in the Two-step authentication, Account recovery, and Account picture modals from `<a>`s to `<button>`s, with one exception: the action to toggle the camera takes you to a new modal and URL, so I felt it appropriate to leave it as an anchor,
- I made sure to use `type="button"` on them all so as to avoid accidental page submissions.
- The previous Two-step authentication and Account recovery actions had an underline effect on hover, whereas the Account picture modal ones had a color shift on hover. I decided to make these consistent and apply the color shift to the former ones, removing their hover underline.
- The Two-step authentication and Account recovery actions were lacking some padding that make tab-focus a little tight looking. I updated that.

Would love to hear your thoughts.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/account_recovery/recovery_key.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/account_recovery/recovery_key.mustache
@@ -13,27 +13,27 @@
 
         {{^isIos}}
             <div class="button-row save-options">
-                <a href="#" class="save-option download-option">
+                <button class="save-option download-option">
                     <div class="graphic graphic-download-option"></div>
                     <p class="name">{{#t}}Download{{/t}}</p>
-                </a>
-                <a href="#" class="save-option copy-option">
+                </button>
+                <button class="save-option copy-option">
                     <div class="graphic graphic-copy-option"></div>
                     <p class="name">{{#t}}Copy{{/t}}</p>
-                </a>
-                <a href="#" class="save-option print-option">
+                </button>
+                <button class="save-option print-option">
                     <div class="graphic graphic-print-option"></div>
                     <p class="name">{{#t}}Print{{/t}}</p>
-                </a>
+                </button>
             </div>
         {{/isIos}}
 
         {{#isIos}}
             <div class="button-row">
-                <a href="#" class="save-option copy-option">
+                <button class="save-option copy-option">
                     <div class="graphic graphic-copy-option"></div>
                     <p class="name">{{#t}}Copy{{/t}}</p>
-                </a>
+                </button>
             </div>
         {{/isIos}}
 
--- a/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
@@ -9,10 +9,12 @@
 
     <nav id="avatar-options">
       <input type="file" id="imageLoader" name="imageLoader" accept="image/png, image/jpeg"/>
-      <a href="#" id="file">{{#t}}Upload{{/t}}</a>
-      <a href="/settings/avatar/camera" id="camera">{{#t}}Camera{{/t}}</a>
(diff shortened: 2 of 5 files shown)


jodyheavener:
Whoops, upon further inspection this needs a little more finessing.

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -13,27 +13,27 @@
 
         {{^isIos}}
             <div class="button-row save-options">
-                <a href="#" class="save-option download-option">
+                <button type="button" href="#" class="save-option download-option">
```
LZoog: We should be able to remove `href="#"` on all of these now.

Also, I don't believe we need `type="button"` unless these guys are in a `form` since users can't submit if they're not in one. Some of what's been changed are in forms and some are not, but that's not a big deal.

packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
```diff
@@ -9,10 +9,12 @@
 
     <nav id="avatar-options">
       <input type="file" id="imageLoader" name="imageLoader" accept="image/png, image/jpeg"/>
-      <a href="#" id="file">{{#t}}Upload{{/t}}</a>
+      <button type="button" id="file">{{#t}}Upload{{/t}}</button>
+
       <a href="/settings/avatar/camera" id="camera">{{#t}}Camera{{/t}}</a>
```
LZoog: We can add `role="button"` here to mimic this being read as a type of "button" for screenreaders.

packages/fxa-content-server/app/styles/modules/_settings-totp.scss
```diff
@@ -150,8 +150,24 @@
 }
 
 .save-option {
+  background-color: transparent;
+  color: $link-color-default;
+  display: inline-block;
+  font-weight: 400;
+  height: initial;
+  margin: -10px 0;
+  padding: 10px 0;
```
LZoog: Instead of adding a negative margin here, let's reduce some padding elsewhere. If we reduce `.recovery-key-container` margin-bottom to `15px` instead of `25px`, and then change the margin of`.save-options` to `15px 0 10px`, we'll get the same affect.

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/recovery_key.mustache
jodyheavener: Good find on the hrefs. Missed those.

Yes, agree - we don't need `type="button"` on buttons not inside a form. I noticed the buttons inside the 2FA modal are inside a form so I added them, and then kinda went to town. I suppose they couldn't hurt, but I've gone ahead and removed the excess ones.

LZoog:
Thanks @jodyheavener for getting on top of those changes! It's appreciated. :) This PR looks good to me. Could you rebase against master and squash your commits into one? I'll approve/merge after that!

jodyheavener:
Rebased and squashed - once green should be good to go :)

Thanks for looking at all these PRs today @LZoog!

PR #3769, opened 2019-12-30
chore(tests): Remove vesitigial reference to `serviceUri` in tests.

rfk:
Because:

* The relier field `serviceUri` was [removed many years ago](https://github.com/mozilla/fxa-content-server/commit/b361aba35df3215105e8d57f108db3f64007e886).
* I happened to be grepping around and noticed that one test still had a reference to that field in its setup logic.

This commit:

* Deletes the test setup code that initializes this vestigial field.


vladikoff:
Please rebase, should fix the test failure

PR #3770, opened 2019-12-30
fix(oauth): Force serviceName="Firefox Sync" when connecting to sync.

### LINKED ISSUE #1531: "Sync Connected" notification should say “Firefox Preview” for Fenix, not "Firefox Sync"

rfk:
Because:

* We want the signin UX to say "continue to Firefox Sync" for any
  client that's requesting access to sync, even OAuth clients.
* We currently ensure this by setting the OAuth client name for
  Fenix to be "Firefox Sync", which produces confusing UX in other
  other parts of the product (such as notifications popping up to
  say that "Firefox Sync has connected to Firefox Sync").

This commit:

* Makes the OAuth Relier model responsible for enforcing this special-case
  behaviour, so that it's localized to just the signin UX.

Connects to #1531.

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -633,8 +633,7 @@ section.modal-panel {
 
   &[data-name='Fenix'],
   &[data-name='Android Components Reference Browser'],
-  // It's a hack, but the only OAuth client named "Firefox Sync" is on Android.
-  &[data-name='Firefox Sync'] {
+  &[data-name='Firefox Preview'] {
     background-image: url('/images/icon-device-phone-android.svg');
   }
 
--- a/packages/fxa-content-server/app/scripts/lib/constants.js
+++ b/packages/fxa-content-server/app/scripts/lib/constants.js
@@ -68,10 +68,12 @@ module.exports = {
     'profile:email',
     'profile:uid',
   ],
+  OAUTH_OLDSYNC_SCOPE: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_WEBCHANNEL_REDIRECT:
     'urn:ietf:wg:oauth:2.0:oob:oauth-redirect-webchannel',
 
   RELIER_DEFAULT_SERVICE_NAME: 'Account Settings',
+  RELIER_SYNC_SERVICE_NAME: 'Firefox Sync',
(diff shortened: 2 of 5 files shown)


packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -633,8 +633,7 @@ section.modal-panel {
 
   &[data-name='Fenix'],
   &[data-name='Android Components Reference Browser'],
-  // It's a hack, but the only OAuth client named "Firefox Sync" is on Android.
-  &[data-name='Firefox Sync'] {
+  &[data-name='Firefox Preview'] {
```
rfk: FWIW, I don't think this style block will be correctly applied. Fenix sets a custom device name after signing in,  meaning that the name field ends up as something unpredictable like "Ryan's Nightly on Samsung Galaxy S8" rather than the configured OAuth client name. But that's a separate issue.

vladikoff:
Please rebase, should fix the test failure

rfk:
👍 rebased

rfk:
Huh, I thought I had posted a comment here with instructions on testing locally, but either github ate it or I forgot to click "comment".

You can test it locally by running the local FxA servers and either requesting or not requesting the "oldsync" scope.  For me, visiting [this url](http://127.0.0.1:3030/oauth/?client_id=a2270f727f45f648&scope=profile) shows "Continue to Fenix" while [this url](http://127.0.0.1:3030/oauth/?client_id=a2270f727f45f648&scope=profile%20https%3A%2F%2Fidentity.mozilla.com%2Fapps%2Foldsync) shows "Continue to Firefox Sync".

rfk:
I'm also going to tag @ryanfeeley for r? here, specifically to confirm that this bit of UX interpretation is correct:

> We want the signin UX to say "continue to Firefox Sync" for any
> client that's requesting access to sync, even OAuth clients.

6a68:
I'm having some trouble with my local install, but don't want to hold this back. Marking as R+

Issue #3771, opened 2019-12-30
build fail: fxa-email-service failing with ECONNREFUSED 127.0.0.1:3306


vladikoff:
https://circleci.com/gh/mozilla/fxa/133002?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

causes issues with PRs like here https://github.com/mozilla/fxa/pull/3769

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-853)
┆Priority: Medium
┆Issue Number: FXA-853

PR #3772, opened 2019-12-30
fix(build): set MySQL 5.7.27 for the email service testing build

vladikoff:
Might fix https://github.com/mozilla/fxa/issues/3771


vladikoff:
Seems to have fixed #3771 !

Issue #3773, opened 2019-12-30
Auth-server tests failing because of fxa-geodb


vbudhram:
From https://app.circleci.com/jobs/github/mozilla/fxa/133100, we get

```
  1) geodb
       returns location data when enabled:
     Error: ENOENT: no such file or directory, open '/fxa-geodb/db/cities-db.mmdb'
```

That code hasn't changed in a while, so might be maxmind geodb related.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-854)
┆Priority: Medium
┆Issue Number: FXA-854

vbudhram:
Per https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

`Starting December 30, 2019, we will be requiring users of our GeoLite2 databases to register for a MaxMind account and obtain a license key in order to download GeoLite2 databases.`

clouserw:
In production we have a full license, but we need something in dev.

immanuelfodor:
Not only in dev, there is also a geodb error in Firefox Send which uses this project as a dependency, and a self-hosted Send won't start without removing it: https://github.com/mozilla/send/issues/1430

PR #3774, opened 2019-12-30
fix(tests): Temporary use fxa hosted cities db

vbudhram:
While we are waiting for a resolution to https://github.com/mozilla/fxa/issues/3773, we will use an fxa hosted cities db.

### DIFF
--- a/packages/fxa-geodb/test/fxa-geodb.js
+++ b/packages/fxa-geodb/test/fxa-geodb.js
@@ -57,8 +57,8 @@ describe('fxa-geodb', () => {
     assert.deepEqual(
       location.latLong,
       {
-        latitude: 37.3893,
-        longitude: -122.0783,
+        latitude: 37.3897,
+        longitude: -122.0832,
       },
       'LatLong returned correctly'
     );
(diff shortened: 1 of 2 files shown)


vbudhram:
Tests are passing again! r?

dannycoates:
58mb :( is there an easy non-git way to do this so it won't be stuck in the git tree forever?

vladikoff:
What happened to the db?

vladikoff:
I recommend we fetch it from elsewhere, like S3

vbudhram:
@vladikoff https://github.com/mozilla/fxa/issues/3773#issuecomment-569737044 :(

@dannycoates I've removed from git and added to S3 👍🏽

packages/fxa-geodb/test/fxa-geodb.js
```diff
@@ -57,8 +57,8 @@ describe('fxa-geodb', () => {
     assert.deepEqual(
       location.latLong,
       {
-        latitude: 37.3893,
-        longitude: -122.0783,
+        latitude: 37.3897,
+        longitude: -122.0832,
```
vbudhram: Latest version had a slight change 🤷🏽‍♂️

vladikoff:
@vbudhram could you please update the commit message?

Issue #3718, opened 2019-12-18
Error: Non-Error exception captured with keys: joi, meta


clouserw:
https://sentry.prod.mozaws.net/operations/content-express-server-prod/issues/6602579/

```
Error: Non-Error exception captured with keys: joi, meta
  File "/app/node_modules/express/lib/router/index.js", line 335, in Function.process_params
    return done();
  File "/app/node_modules/express/lib/router/index.js", line 275, in next
    self.process_params(layer, paramcalled, req, res, function (err) {
  File "/app/node_modules/body-parser/lib/read.js", line 130, in null.<anonymous>
    next()
  File "/app/node_modules/raw-body/index.js", line 224, in invokeCallback
    callback.apply(null, args)
  File "/app/node_modules/raw-body/index.js", line 213, in done
    invokeCallback()
...
(5 additional frame(s) were not displayed)

Non-Error exception captured with keys: joi, meta
```

That output above isn't super helpful, but the Sentry page has more detail.  On our /metrics URL where we accept top-level funnel metrics we accept an `entrypoint` parameter, but if it has a `#` in it we throw this validation error.  We should just strip off a `#` (and anything following it) before running the validation.

vladikoff:
dup of https://github.com/mozilla/fxa/issues/3703

irrationalagent:
Any idea what the entrypoint values that contain `#` are? Are we confident that they are coming from a known source and should not be dropped?

vladikoff:
@irrationalagent : `entrypoint: awesomescreen-signup#, `

irrationalagent:
OK. that looks like fennec. I'll confirm with them that they are doing this but it seems like we are ok to start stripping them out. Maybe we should work on advertising our validation restrictions to other teams rather than having to adapt or change our validation system? Or maybe not (not an engineer)

clouserw:
We should, and if someone can point me to where the validation occurs, I'm happy to add it to the docs :)

6a68:
Hmm. The validation occurs [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-shared/express/routing.js#L58-L62), but it really seems like the `#` should be parsed as a fragment, not as a query string. I wonder if this is a bug in `celebrate` or if we're using its API incorrectly.

I suspect the metrics validation code will soon move to JSON Schema to match what Data Engineering uses. Adding this to a metrics grab bag epic related to the pipeline changes.

[...]

PR #3777, opened 2019-12-30
chore(fxa-shared): update celebrate chasing validation issue

meandavejustice:
- refs #3718


packages/fxa-shared/package.json
```diff
@@ -58,7 +58,7 @@
     "accept-language": "^2.0.17",
     "ajv": "^6.10.2",
     "bluebird": "^3.5.5",
-    "celebrate": "^10.0.1",
+    "celebrate": "11.0.0",
```
dannycoates: did you mean `11` or `^11`?

meandavejustice: 11. I don't trust the carots in version numbers, I like to convert to this style whenever I update versions

6a68: @meandavejustice Ah, I asked about this [a while ago](https://github.com/mozilla/fxa/pull/2016#issuecomment-517180157), and was told that the FxA team had decided to use carets in a past retro. If you'd like to change how the versioning is tracked in package.json files, maybe open a pull request with an ADR that outlines the pros and cons?

meandavejustice: I'll just move it back for now, I haven't run into any issues on fxa with it yet, just previously with testpilot and lockwise.

6a68: Cool, sounds good. Not sure what's up with the test failures; trying another run.

PR #3779, opened 2019-12-30
chore(payments-server): add l10n for payment legal blurb

meandavejustice:
- refs #3704

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
@@ -1,21 +1,28 @@
 import React from 'react';
+import { Localized } from 'fluent-react';
 
 import './index.scss';
 
 export const PaymentLegalBlurb = () => (
   <div className="payment-legal-blurb">
-    <p>Mozilla uses Stripe for secure payment processing.</p>
-    <p>
-      View the{' '}
-      <a
-        href="https://stripe.com/privacy"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        Stripe privacy policy
-      </a>
-      .
-    </p>
+    <Localized id="payment-legal-copy">
+      <p>Mozilla uses Stripe for secure payment processing.</p>
+    </Localized>
+
+    <Localized
+      id="payment-legal-link"
+      a={
+        <a
+          href="https://stripe.com/privacy"
+          target="_blank"
+          rel="noopener noreferrer"
+        ></a>
+      }
+    >
+      <p>
+        View the <a>Stripe privacy policy</a>.
+      </p>
+    </Localized>
   </div>
 );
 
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
```diff
@@ -1,21 +1,28 @@
...
-        href="https://stripe.com/privacy"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        Stripe privacy policy
-      </a>
-      .
-    </p>
+    <Localized id="payment-legal-copy">
+      <p>Mozilla uses Stripe for secure payment processing.</p>
```
LZoog: Is it possible to set `payment-legal-copy` and `payment-legal-link` to variables here like we do in `errors.ts` (`const BASIC_ERROR = 'basic-error-message'`) so we don't have duplicate copy?

meandavejustice: Not sure I understand, those two id's have different copy

LZoog: Sorry I missed your comment. I meant, can we set these strings in one spot so we don't have "Mozilla uses Stripe for secure payment processing." in `index.tsx` _and_ `main.ftl`?  I'm not entirely sure how Fluent works, but in `errors.ts` we've got:

```
const BASIC_ERROR = 'basic-error-message';
```

Is it possible to set `const PAYMENT_LEGAL_COPY = 'payment-legal-copy'` and then set `<p>${PAYMENT_LEGAL_COPY}</p>` here?

LZoog: Or is there some other way of doing that? Having every piece of copy in two spots seems 😢

meandavejustice: I see, in this case  the copy in `index.tsx` would be used as a fall back if the translation doesn't work. So if we had 

```
const PAYMENT_LEGAL_COPY = 'payment-legal-copy';
return (
<p>${PAYMENT_LEGAL_COPY}</p>
);
```
it would return: `<p>payment-legal-copy</p>`, instead of the translated text.

You could also say that the text inside of `index.tsx` is optional, and that it's only there to help understanding the code easier for the team

LZoog: Makes sense. It'd be nice if it'd pull in the default string provided if it failed for some reason.

 I feel at some point, we're going to need to update some copy and we'll only do it in one place and miss the other. Is there any way to set our strings in another place and import them into our files + Fluent (does it _have_ to be a `.ftl` file)? Or maybe this is fine for now and we can [automate `.ftl` file creation](https://github.com/udacity/fluent-react-utils/tree/master/packages/extraction) (that package or another) in the future?

meandavejustice: The syntax for fluent strings is different than react/vanilla js when it comes to variables. So it could end up being a lot more complex / time consuming than we'll be willing to devote to it. But I agree, that would be optimal

LZoog: https://bedrock.readthedocs.io/en/latest/l10n.html#the-ftl-helper-function

This could be useful👀

PR #3789, opened 2019-12-31
chore(payments-server): remove unused App.scss file

meandavejustice:


lmorchard:
Hmm, looks like this broke the Storybook build, but that breakage didn't cause CI to fail. Hopefully this fixes it https://github.com/mozilla/fxa/pull/3818

PR #3791, opened 2019-12-31
chore(payments-server): l10n for plan details

meandavejustice:
- refs #3704


packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -32,3 +32,7 @@ settings-subscriptions = Subscriptions
 ## legal footer
 terms = Terms of Service
 privacy = Privacy Notice
+
+
```
LZoog: For consistency we could get a comment here ## plan details

meandavejustice: totally, thanks for catching that

PR #3792, opened 2019-12-31
fix(password-balloon): Hide balloon when user enters invalid password, then valid password, then changes focus

### LINKED ISSUE #3750: If user types password in sign-up and quickly tabs away when matching criteria, strength balloon persists
I believe this is an edge-case I didn't catch in #3724.

If a user types their password quickly enough and tabs before all 3 checks are shown, the password strength balloon doesn't disappear.

pw

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-845)
┆Priority: Medium
┆Issue Number: FXA-845

LZoog:
fixes #3750 

This is a simple solution for a slightly little tricky problem since we're chasing the debounce initialized in `password_with_strength_balloon`, set from options or a default. I don't like `setTimeOut`s where we don't need them, and we can't conditionally run `this.render()` in the `update` function because then there's still a red border around the input if the previous password was invalid.

Note: When making this recording, I noticed the strength balloon doesn't fade out as slowly as it does outside of this edgecase. We're using a `setTimeout` for the transition because (per a comment) "`transitionend` is not fired for an unknown reason", referring to the jQ method `.hide`. It looks like for some reason, neither `transitionend` or `setTimeout` work in this edgecase to fade out the balloon. It works normally otherwise. I can create a follow-up issue but it'll definitely be a P5 issue thrown into the backlog.

pw

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
+++ b/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
@@ -55,14 +55,18 @@ class PasswordStrengthBalloonView extends BaseView {
     if (!this.model.validationError && !this.model.get('inputFocused')) {
       return this.hide();
     }
+    // OneVisibleOfTypeMixin uses 'show' to destroy any other
+    // tooltip-like views.
+    return this.show();
   }
 
   afterRender() {
-    if (this.model.validationError) {
-      // OneVisibleOfTypeMixin uses 'show' to destroy any other
-      // tooltip-like views.
-      this.show();
-    }
+    // We want to conditionally hide after rendering because
+    // users can type an invalid password, and then type a valid
+    // password and quickly change focus to beat the debounce
+    // on the password change event. This allows the model to
+    // update but hides the balloon in that scenario. Ref #3750
+    this.shouldHide();
   }
 
   update() {


chenba:
Couldn't we combine everything into one function that either hide or show based on the conditions in `shouldHide`?

LZoog:
@chenba On my first pass at this it felt like `shouldHide` should conditionally run `hide` and return true/false for the `else` needed in `afterRender`, but having a side-effect smelled bad. Yet I agree what I've got now seems like it could be DRY'd up. Reverting my changes and adding the `else` in `shouldHide` would run `.show` unnecessarily when the user shifts focus and hasn't met password criteria since the balloon wouldn't be hidden, but that might be a little cleaner. I'll check now.

chenba:
> ... adding the else in `shouldHide` would run .show unnecessarily...

That's true.  The small performance hit here is worth it to me.  But you can make the call. :D

LZoog:
@chenba that did end up feeling cleaner so I pushed those changes. I'm pretty sure the balloon not fading out in this scenario is due to not rerendering after running `.show`, so `afterRender` might be an awkward place for showing/hiding... now that we're hiding the balloon when the user switches focus and there's no error, I just noticed this case where because of the delay on setting an error, if the user types and focuses away before the balloon shows an error, this happens, and also doesn't fade in. It doesn't break anything and is also edgecase-y, but isn't ideal.

I'm going to see if I can fix this relatively quickly, otherwise I'll create a new issue.

pw

LZoog:
Closing this in favor of #3795

Issue #3575, opened 2019-12-04
Improve edge cases for inline password validation


vbudhram:
From https://github.com/mozilla/fxa/pull/3548, there are a couple edge cases where the tooltip is not correctly displayed.

Ref

<img width="444" alt="Screen Shot 2019-12-04 at 1 33 07 PM" src="https://user-images.githubusercontent.com/1295288/70170419-a52a2a00-169a-11ea-9568-a5093d148a54.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-53)
┆Issue Number: FXA-53

LZoog:
Note: first edgecase is fixed by #3795

byrw:
May I work on the second edge case?

The solution that best worked for me, is just verifying if the `#new_password` input is valid before verifying whether passwords do match. This solution is implemented in the same way as the function `showValidationErrors()`, in the file change_passwords.jsx.

vasilicatamas:
This issue is verified as fixed on FxA Stage [Train 1.187.2] under Windows 10 64-bit. "Passwords do not match" error is no longer displayed if there is a validation error. 

![Screenshot_48](https://user-images.githubusercontent.com/18114210/93070154-2f220d00-f687-11ea-918e-fb6fa9a2221f.png)

PR #3795, opened 2020-01-02
fix(password-balloon): Improve password strength balloon edgecases

### LINKED ISSUE #3465: Password Strength Bubble Not Displayed for a Short, Edited Password
#### STP
1. Enter a valid password so that password strength bubble goes away
1. Hold down backspace until the password is fewer than eight character
1. Quickly press tab to move to confirm password input

#### Expected

Password strength bubble automatically re-appears as the password is too short

#### Actual

Password strength bubble does not show when the user tab over to the confirm password field.  It does re-appear when the user try to submit the password.

image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-575)
┆Priority: Medium
┆Issue Number: FXA-575

### LINKED ISSUE #3750: If user types password in sign-up and quickly tabs away when matching criteria, strength balloon persists

LZoog:
Fixes #3750 
Fixes #3465
Partial fix for #3575 (fixes the first issue mentioned)

Also:
- Fixes edgecase shown at the bottom of #3792
- The balloon now updates on keyboard paste. 🎉 While the balloon does not update on mouse paste, it _does_ update when you paste and then tab away (not new).

pw

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
+++ b/packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
@@ -30,7 +30,7 @@ class PasswordStrengthBalloonView extends BaseView {
 
   initialize(config = {}) {
     this.listenTo(this.model, 'change:password', this.update);
-    this.listenTo(this.model, 'change:inputFocused', this.shouldHide);
+    this.listenTo(this.model, 'change:inputFocused', this.hideOrShow);
   }
 
   setInitialContext(context) {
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/views/password_strength/password_strength_balloon.js
```diff
@@ -55,14 +55,18 @@ class PasswordStrengthBalloonView extends BaseView {
     if (!this.model.validationError && !this.model.get('inputFocused')) {
       return this.hide();
     }
+    // OneVisibleOfTypeMixin uses 'show' to destroy any other
+    // tooltip-like views.
+    return this.show();
```
chenba: We probably should update the function name now that it shows and hides.

packages/fxa-content-server/app/scripts/views/password_strength/password_with_strength_balloon.js
```diff
@@ -24,8 +24,7 @@ const PasswordWithStrengthBalloonView = FormView.extend({
   events: {
     change: 'updateModelAfterDelay',
     focus: 'focusHandler',
-    keypress: 'updateModelAfterDelay',
-    blur: 'blurHandler',
+    keydown: 'updateModelAfterDelay',
```
chenba: Doesn't a keydown also trigger a change event, which is handled by `updateModelAfterDelay` a couple lines up?

LZoog: Ooo @chenba good call out - it doesn't look like we need `change` here at all. `keypress` also fired `updateModelAfterDelay` but this may have been missed due to the debounce. I took out `change` and readded `blur`, and passed the `event.type` to `updateModel` instead of using jQ to check if the el was out of focus. I also made a comment clearer, added a test for `inputFocused` on the model, and updated `shouldHide` to `hideOrShow`.

I checked all the edgecases and tried to break the thing and it's looking good to me. Clicking on password reveal all works as expected too.

PR #3805, opened 2020-01-03
feat(ready-view): Display serviceName if user is signed in

### LINKED ISSUE #3623: Display `serviceName` in "ready" view if user is authenticated
Follow-up to #3569 

@shane-tomlinson I'm going to spell this out to make sure I've got this right.

`ready.mustache` needs some more finessing. UX would prefer "You are now ready to use Account Settings" to display instead of "Your account is ready!," but we need a condition to check if the user is signed in. Otherwise, a user can hit this screen because of an email verification reminder, and "You are now ready to use Account Settings" doesn't make sense because they're not signed in.

It looks like we want to change this:
```
{{^isFromRelyingParty}}
  {{#emailVerified}}
    <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
  {{/emailVerified}}
  {{^emailVerified}}
     <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
  {{/emailVerified}}
{{/isFromRelyingParty}}
```

(`escapedEmailReadyText` should display "'You are now ready to make changes to your Firefox Account.")

to:
```
{{^isFromRelyingParty}}
  {{#emailVerified}}
    <p class="account-ready-service">{{{escapedEmailReadyText}}}</p>
  {{/emailVerified}}
  {{^emailVerified}}
    {{#isAuthenticated}}
         <p class="account-ready-service">{{#t}}You are now ready to use %(serviceName)s{{/t}}</p>
    {{/isAuthenticated}}
    {{^isAuthenticated}}
      <p class="account-ready-generic">{{#t}}Your account is ready!{{/t}}</p>
    {{/isAuthenticated}}
  {{/emailVerified}}
{{/isFromRelyingParty}}
```

This screen is getting a little convoluted. The `escapedEmailReadyText` differentiation seems strange to me - I don't think I'd know as a user what the difference is between "You are now ready to make changes to your Firefox Account" and "Your account is ready!".

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-34)
┆Priority: Medium
┆Issue Number: FXA-34

LZoog:
fixes #3623 

Per Ryan F.'s request, the ready view should show `You are now ready to use ${serviceName}` (defaults to "Account Settings") instead of `Your account is ready!`. In #3569, Shane brought up that we still need to use the latter when a user is not signed in. This PR takes both into consideration.

I added the "Continue to my account" button per Shane's suggestion in #3623, which I agree makes sense here. I also wrapped the `margin-top` marketing style in `&:not(:empty)` because otherwise, there's 20px of extra margin below the text/button.

I double checked these via `/primary_email_verified`.

Hitting this page with `primary_email_verified=true` param:

Hitting this page without this param (not sure if it’s actually possible to hit “primary email” without it, just wanted to verify the copy displayed correctly):

And hitting this page while not signed in:

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/views/ready.js
+++ b/packages/fxa-content-server/app/tests/spec/views/ready.js
@@ -14,6 +14,7 @@ import Session from 'lib/session';
 import sinon from 'sinon';
 import BrowserRelier from 'models/reliers/browser';
 import View from 'views/ready';
+import User from 'models/user';
 import WindowMock from '../../mocks/window';
 import {
   SIGNIN_COMPLETE,
@@ -30,6 +31,7 @@ describe('views/ready', function() {
   let relier;
   let view;
   let windowMock;
+  let user;
 
   function createDeps() {
     windowMock = new WindowMock();
@@ -48,6 +50,7 @@ describe('views/ready', function() {
     fxaClient = new FxaClient();
     model = new Backbone.Model({});
     notifier = new Notifier();
+    user = new User();
 
     metrics = {
       flush: sinon.spy(Promise.resolve),
@@ -63,6 +66,7 @@ describe('views/ready', function() {
       lang: lang,
       metrics: metrics,
       model,
+      user,
       notifier: notifier,
       relier: relier,
       type: type,
@@ -70,6 +74,7 @@ describe('views/ready', function() {
       window: windowMock,
     });
 
+    sinon.stub(user, 'isSignedInAccount').callsFake(() => true);
     sinon.stub(view, 'getSignedInAccount').callsFake(() => {});
   }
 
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/tests/spec/views/ready.js
```diff
@@ -63,13 +66,15 @@ describe('views/ready', function() {
...
       model,
+      user,
       notifier: notifier,
       relier: relier,
       type: type,
       viewName: 'ready',
       window: windowMock,
     });
 
+    sinon.stub(user, 'isSignedInAccount').callsFake(() => true);
```
chenba: `sinon.stub(user, 'isSignedInAccount').returns(true);` also works.

packages/fxa-content-server/app/tests/spec/views/ready.js
```diff
@@ -234,6 +239,28 @@ describe('views/ready', function() {
...
+
+      return view.render().then(function() {
+        assert.equal(
+          view.$('.account-ready-generic').text(),
+          'Your account is ready!'
+        );
+      });
+    });
+  });
+
```
chenba: No existing tests for `emailVerified`?  Wanna add some?  😁

LZoog: @chenba Sure thing. I'll create a follow-up PR this morning so this one can get in this train - we also aren't testing that `secondaryEmailVerified` spits out the correct copy and we probably should be.

PR #3849, opened 2020-01-08
chore(payments-server): move AppLocalizationProvider into own file

### LINKED ISSUE #3762: move AppLocalizationProvider into separate file and use TS types over proptypes
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-849)
┆Issue Number: FXA-849

meandavejustice:
and replace proptypes with typescript types
- fixes #3762


packages/fxa-payments-server/src/lib/AppLocalizationProvider.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+interface State {
+  baseDir: string,
+  userLocales: Array<string>;
+  bundles: Array<string>;
+  messages: FluentBundle;
+};
+
+type Props {
```
chenba: Type aliases need an assignment operator.

packages/fxa-payments-server/src/App.tsx
```diff
@@ -58,7 +58,7 @@ export const App = ({
   return (
     <AppContext.Provider value={appContextValue}>
       <AppLocalizationProvider
-        userLocales={navigator.languages}
+        userLocales={[...navigator.languages]}
```
chenba: Why do we need to spread `navigator.languages` here?

packages/fxa-payments-server/src/lib/AppLocalizationProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+    return messages;
+  } catch (e) {
+    // We couldn't fetch any strings; just return nothing and fluent will fall
+    // back to the default locale if needed.
+    return '';
+  }
+}
+
+function fetchAllMessages(baseDir: string, locale: string, bundles: Array<string>) {
+  return Promise.all(bundles.map((i) => fetchMessages(baseDir, locale, i)));
```
chenba: nit: `i` seems inconsistent when it's next to other more descriptive var names.

packages/fxa-payments-server/src/App.tsx
meandavejustice: Its a typescript thing, navigator.languages is an array of readonly strings, we need to copy the array to assign it to userLocales.

chenba: Ah, gotcha.  I wonder if that's still needed if we declare `userLocales` to be readonly also.

meandavejustice: Declaring as `ReadonlyArray<string>` fixed it. 👍

PR #3900, opened 2020-01-14
refactor(auth-server): async and setup stripeHelper

bbangert:
Because:

* Multiple upcoming PR's will need the stripeHelper.
* key_server main was a yucky promise chain.

This commit:

* Makes stripeHelper available in routing.
* Uses async/await to clean-up the main function in key_server.

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -9,11 +9,11 @@
 // If required, modules will be instrumented.
 require('../lib/newrelic')();
 
+const error = require('../lib/error');
 const jwtool = require('fxa-jwtool');
-const P = require('../lib/promise');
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/bin/key_server.js
```diff
@@ -69,130 +70,120 @@ function run(config) {
...
-              });
-            }
+  return {
+    server,
+    log: log,
+    close() {
+      return new P(resolve => {
+        log.info({ op: 'shutdown' });
+        clearInterval(statsInterval);
+        server.stop().then(() => {
```
dannycoates: why not `await` the remaining `then`s in this file?

bbangert: Awwww.... cause I didn't see that one, doh!

PR #3901, opened 2020-01-14
chore(tests): Add tests for ready.js covering 'escapedEmailReadyText' for primary and secondary email verifications

LZoog:
I changed `expectedHeaders` to `expectedHeadersForSyncViews` because I had some trouble when trying to stick a verification reason type in this object since `TEMPLATE_INFO.PRIMARY_EMAIL_VERIFIED.readyToSyncText` (and `SECONDARY_EMAIL_VERIFIED.readyToSyncText`) aren't defined in `ready.js` and this name would have given me a hint.

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/views/ready.js
+++ b/packages/fxa-content-server/app/tests/spec/views/ready.js
@@ -88,7 +88,7 @@ describe('views/ready', function() {
       view.destroy();
     });
 
-    const expectedHeaders = {
+    const expectedHeadersForSyncViews = {
       FORCE_AUTH: '#fxa-force-auth-complete-header',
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/tests/spec/views/ready.js
```diff
@@ -119,6 +119,36 @@ describe('views/ready', function() {
       });
     });
 
+    it('renders `escapedEmailReadyText` as expected with `primary_email_verified` view type and param', () => {
+      windowMock.location.search = '?primary_email_verified=some@email.com';
+      sinon.stub(relier, 'isSync').returns(false);
```
chenba: It might be good to run `relier.isSync.restore()` at the end of the test, in case someday someone in some other tests expects `relier.isSync` to have its original behavior.

Issue #3855, opened 2020-01-09
Create standalone views for adding account recovery


vbudhram:
As part of https://github.com/mozilla/fxa/issues/2653 we need to develop a non-settings flow for enabling account recovery.

Ref designs at https://mozilla.invisionapp.com/share/Q3UK63V4PKA#/screens/390543494_Recovery_Key_Recommendation

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-902)
┆Issue Number: FXA-902

Issue #3854, opened 2020-01-09
Enable the Phase 2 experiment for increasing account recovery and secondary emails


vbudhram:
Tracking issue to make sure this gets completed

Connects with https://github.com/mozilla/fxa/issues/2653

- [ ] Add experiment logic
- [x] Confirm rollout rate
- [x] Add missing tests
- [ ] Confirm metrics

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-901)
┆Issue Number: FXA-901

PR #3903, opened 2020-01-14
feat(recovery): Initial account recovery standalone screens

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/3855

- [x] Add ux views
- [x] Add unit tests
- [x] Add functional tests

This PR adds standalone screens for creating a recovery key.

* Adds UX based on https://mozilla.invisionapp.com/share/Q3UK63V4PKA#/screens/390543494_Recovery_Key_Recommendation
* Did not add the `confirm key` screen because that will require some additional backend work, plan is to work that in a follow up issue since this PR is pretty big already
* To test, login and open http://127.0.0.1:3030/post_verify/account_recovery/add_recovery_key


packages/fxa-content-server/app/scripts/views/post_verify/account_recovery/confirm_password.js
```diff
@@ -0,0 +1,79 @@
...
+        this.navigate(
+          '/post_verify/account_recovery/save_recovery_key',
+          recoveryKey
+        );
+      })
+      .catch(err => {
+        if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+          return this.showValidationError(this.$(PASSWORD_SELECTOR), err);
+        }
+        throw err;
```
LZoog: Would `throw err` in a `.catch` cause an unhandled promise rejection warning?

packages/fxa-content-server/app/scripts/templates/post_verify/account_recovery/confirm_password.mustache
```diff
@@ -0,0 +1,35 @@
...
+
+            <div class="button-row">
+                <button id="submit-btn" type="submit">{{#t}}Continue{{/t}}</button>
+            </div>
+            <div class="links">
+                <a id="maybe-later-btn" data-flow-event="link.maybe-later">{{#t}}Maybe later{{/t}}</a>
+            </div>
+        </form>
+    </section>
+</div>
```
LZoog: We may want a follow-up issue to add an "eye" toggle to this input, but probably `needs:ux`.

image

packages/fxa-content-server/app/scripts/views/post_verify/account_recovery/confirm_password.js
vbudhram: Not in this case, the error is caught and handled at a lower level.

packages/fxa-content-server/app/scripts/templates/post_verify/account_recovery/confirm_password.mustache
vbudhram: Yea we should do that, although I thought this would have happened automatically since this was a password field 🤔

vbudhram:
> Also, just curious, what users are we going to show this to post-verify and will that be setup in the mentioned follow-up PR with the confirm key screen functionality?

The `post-verify` name is a bit misleading now that I think about it. It was originally intended to be shown to users after a login/signin, but really they are standalone pages that could be linked from anywhere. In https://github.com/mozilla/fxa/issues/3936, only the missing screens will be added. In https://github.com/mozilla/fxa/issues/3854, feature will get enabled.

PR #3905, opened 2020-01-14
refactor key_server.js

dannycoates:
building on #3900 and tidying up some more. no new functionality


packages/fxa-auth-server/bin/key_server.js
```diff
@@ -4,95 +4,52 @@
 
 'use strict';
 
-// This MUST be the first require in the program.
-// Only `require()` the newrelic module if explicity enabled.
-// If required, modules will be instrumented.
-require('../lib/newrelic')();
```
vbudhram: 👍

PR #3908, opened 2020-01-14
fix(devices): maintain the device id during the token exchange

vladikoff:
Fixes https://github.com/mozilla/application-services/issues/2473

@rfk This fixes the CLI use case.

Draft PR until I add tests.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/utils/oauth.js
+++ b/packages/fxa-auth-server/lib/routes/utils/oauth.js
@@ -57,9 +57,12 @@ module.exports = {
     credentials.tokenVerified = true;
     credentials.client = await oauthdb.getClientInfo(clientId);
 
-    // The following upsert gets no `deviceInfo`.
-    // However, `credentials.client` lets it generate a default name for the device.
-    await devices.upsert(request, credentials, {});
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/utils/oauth.js
```diff
@@ -57,9 +57,11 @@ module.exports = {
     credentials.tokenVerified = true;
     credentials.client = await oauthdb.getClientInfo(clientId);
 
-    // The following upsert gets no `deviceInfo`.
     // However, `credentials.client` lets it generate a default name for the device.
-    await devices.upsert(request, credentials, {});
+    await devices.upsert(request, credentials, {
```
rfk: Nit: The "However" in the comment above is now confusing, I think we should move it down with the other comment and say something like:

// Connect the new refreshToken to the existing device record, or create a new placeholder if there isn't one.
// We don't provide any other device info, but `credentials.client` will let it generate a default name for the device.

vladikoff: 👍

vladikoff:
@rfk added a Test, and testing via the emulator! 

Landing on `master` for train-155

Issue #3809, opened 2020-01-06
Implement direct Stripe API calls


bbangert:
For all calls that previously used SubHub, direct Stripe API calls should be used. These should be implemented in fxa-shared and utilized from fxa-auth-server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-866)
┆Issue Number: FXA-866

Issue #3808, opened 2020-01-06
Use cached Stripe customer data as source of subscription status


lmorchard:
We currently use a local DB table in auth-server that tracks subscriptions per user. This information is redundant when considering the user's customer record in Stripe as the source of truth. So, let's use that customer record directly instead - and remove the local DB table altogether.

Also, let's cache the data locally to reduce hits to the Stripe API. This data is used by every RP that fetches a user's profile to check subscriptions, so that will be a hot spot for access.

Cache lifetime & invalidation policy is a concern. Stale cache means that a recently purchased or expired subscription is not reflected in the system. So, we should explicitly expire a cached customer record whenever we know the data has changed - i.e. when a user subscribes or cancels, or when Stripe sends a webhook event that changes a subscription.

Another handy point to invalidate the cache could be when the user authenticates to use the subscription management pages. That would help ensure that the data is fresh as the user begins to engage in subscription changes.

It might also help to have an explicit cache invalidation tool or button for a support agent to force-refresh a user's customer record in the cache.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-865)
┆Issue Number: FXA-865

bbangert:
@lmorchard notes that we will store the entire Stripe customer object with nested fields for subscriptions in Redis.

Issue #3418, opened 2019-11-19
Implement support for subscription capabilities from Stripe product/plan metadata


lmorchard:
Along with other [product/plan metadata we're adding](https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=COPS&title=SP+Tiered+Product+Support#SPTieredProductSupport-MetadataAgreements) for #3243 Upgradable Subscriptions, we asked for `capability:{clientId}` metadata keys to be added. These will list comma-separated product capabilities that we can use to determine subscription capabilities in a user's profile when requested by an RP.

This metadata can replace or supplement [the `productCapabilities` and `clientCapabilities` server config settings](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/config/index.js#L674) required by the [`determineClientVisibleSubscriptionCapabilities`](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/utils/subscriptions.js#L14) function that calculates capabilities for user profiles & JWT tokens. 

This issue is about implementing support for that new metadata.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-555)
┆Issue Number: FXA-555

lmorchard:
I didn't put this into the #3243 epic because it's not strictly a blocker for upgrading subscriptions. More of a nice-to-have.

Issue #886, opened 2019-04-18
Implement better failure handling for subscription creation and cancellation?


lmorchard:
Related to #706 

@shane-tomlinson [asks](https://github.com/mozilla/fxa/pull/819#discussion_r276180011):

> Is it possible for the subscription to be created on subhub but fail when being created in our DB? If so, how do we handle the SubHub subscription? Should it be deleted? Since these two calls are not atomic, it seems like we'll need to handle failures.

> Follow on: Similar to the comment below, the operations to update subhub and our local db are not atomic, meaning it's possible for the call to db.deleteAccountSubscription to fail and subhub.cancelSubscription to succeed. We'll want to think about how to handle those cases. Do we have to re-subscribe the user to subhub if the db.deleteAccountSubscription fails? Do we re-try? Do we alert loudly?

Maybe we can fall back to async events from subhub (#713) to ensure subscription status is updated? Creating this issue to call out the concern and decide what might need doing one way or another.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-817)
┆Issue Number: FXA-817

rfk:
One option could be to have a boolean "dirty bit" on the subscription records in our db, and update the db in two phases.  So adding a subscription would look like:

* Write the subscription into the db, flagged as dirty=true
* Write the subscription into subhub
* Update the subscription in the db to set dirty=false

Deleting a subscription would go in the opposite order:

* Update the subscription in the db to set dirty=true
* Delete the subscription in subhub
* Delete the subscription in the db

When the user comes to exercise their subscriptions, you would check the dirty bit and, if true, do some reconciliation with their account state in subhub to see what the correct state is.

There would be nontrivial complexity involved here, not sure if it would be worth it in practice. But wanted to throw it out there as a suggestion.

rfk:
(Another option is to just assume it will happen rarely, and clean it up by hand when users complain...which could in fact be the lowest-cost option overall)

clouserw:
From our team discussion:
- Any communication with subhub should have the capability to resend/retry
- We expect Subhub to reply immediately (synchronously) with a success message

If we don't get a success message, then: 
- If 5xx: Data is resent immediately
- If 4xx: Data is not resent.  Log errors, I guess.
- If anything else: Data is resent immediately

Maximum number of retries: Default 3, but it should be a configuration variable.  This should include a backoff option for subsequent retries.  Subhub hasn't implemented a backoff header at this point.

[...]

PR #3917, opened 2020-01-15
feat(auth-server): direct stripe API usage

bbangert:
Because:

* Sub Plat Architecture 2.0 has FxA using Stripe direcly.
* Remaining Arch 2.0 issues all require the same base API functionality
  created here.

This commit:

* Implements direct Stripe API calls for the subscription routes.
* Skips the unfinished tests for the direct integration as the other
  issues are blocked and unblocking them is a priority.

Issue #3809

Co-authored-by: Jackie Munroe <jackiemunroe@users.noreply.github.com>


packages/fxa-auth-server/lib/error.js
```diff
@@ -1228,6 +1238,14 @@ AppError.subscriptionAlreadyChanged = () => {
   });
 };
 
+AppError.subscriptionAlreadyExists = () => {
+  return new AppError({
+    code: 409,
+    error: 'Already subscribed',
+    message: 'User already subscribed.',
```
chenba: No `errno` for this guy?

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -10,6 +10,597 @@ const ScopeSet = require('../../../fxa-shared').oauth.scopes;
...
+    config,
+    customs,
+    push,
+    mailer,
+    profile,
+    payments
+  );
+
+  // FIXME: All of these need to be wrapped in Stripe error handling
+  // FIXME: Many of these stripe calls need retries with careful thought about
```
lmorchard: Might be a good to consider under the umbrella of the long-lived issue #886

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -154,4 +155,59 @@ describe('StripeHelper', () => {
...
+      assert.isObject(thrown);
+      assert.equal(thrown.message, 'A backend service request failed.');
+      assert.equal(
+        thrown.cause().message,
+        'Customer for email: test@example.com in Stripe has mismatched uid'
+      );
+    });
+  });
+
+  describe('findPlayById', () => {
```
lmorchard: `s/Play/Plan/` but I might have made this change in my forthcoming branch for #3808 & #3418 since I add tests here

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -534,43 +539,709 @@ describe('subscriptions', () => {
...
+
+      assert.isTrue(failed);
+    });
+  });
+});
+
+/**
+ * Direct Stripe integration tests
+ */
+describe.skip('subscriptions (using direct stripe access)', () => {
```
lmorchard: Noted in the PR description that these are skipped... just noting again.

packages/fxa-auth-server/lib/error.js
bbangert: Ah, it should have one. Adding.

packages/fxa-payments-server/server/config/index.js
```diff
@@ -272,7 +272,7 @@ const conf = convict({
   },
   stripe: {
     apiKey: {
-      default: 'pk_test_FL2cOisOukoCQUZsrochvTlk00ff4IakfE',
+      default: 'pk_test_UrTy6gLWGrc0aOHtbZZz0UBs005mSI4AS2',
```
lmorchard: IIRC, the change here is to stop using my personal Stripe key (😬) and switch to what our dev deployment uses.

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -10,6 +10,597 @@ const ScopeSet = require('../../../fxa-shared').oauth.scopes;
...
+        response: {
+          schema: isA.array().items(
+            isA.object().keys({
+              clientId: isA.string(),
+              capabilities: isA.array().items(isA.string()),
+            })
+          ),
+        },
+      },
+      handler: request => directStripeRoutes.getClients(request),
```
lmorchard: Curious what folks think of this pattern for getting handlers out of this hapi data structure - and if we might want to eventually extend it into reorgs of other routes?

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -95,7 +95,9 @@ class StripeHelper {
     this.plansCacheIsEnabled = this.cacheTtlSeconds && redis;
 
     this.stripe = new stripe(config.subscriptions.stripeApiKey, {
+      apiVersion: '2019-12-03',
       maxNetworkRetries: 3,
+      typescript: undefined,
```
chenba: I cannot find what values are allowed here on their README.  I guess `undefined` is essentially the same as `false`?

bbangert: Yes, though oddly the way they setup the typing, if you include this object at all, then it *must* include the `typescript` flag... which can only be true. This is just an artifact of the weird typing they did.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -136,6 +138,34 @@ class StripeHelper {
...
+  async fetchCustomer(uid, email, expand) {
+    const customerResponse = await this.stripe.customers
+      .list({ email, expand })
+      .autoPagingToArray({ limit: 20 });
+    if (customerResponse.length === 0) {
+      return;
+    }
+    const customer = customerResponse[0];
+
+    if (customer.metadata.userid !== uid) {
```
chenba: I suppose in the rare case where the caller passes in `undefined` for the uid and there's no userid in the metadata, this guard will be bypassed.

packages/fxa-auth-server/lib/payments/stripe.js
bbangert: I filed an issue with the stripe lib to not require this value.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -245,6 +290,51 @@ class StripeHelper {
...
+        plan_name: sub.plan.nickname,
+        plan_id: sub.plan.id,
+        status: sub.status,
+        subscription_id: sub.id,
+        failure_code,
+        failure_message,
+      });
+    }
+    return subs;
+  }
```
chenba: Style preference comment: I prefer `return subscriptions.data.map(sub => { ... })`

packages/fxa-auth-server/lib/payments/stripe.js
bbangert: It's my hope that we will start moving to TypeScript sooner rather than later, and this does have JSDoc typing on it for use with typechecking in JS.

packages/fxa-auth-server/lib/payments/stripe.js
bbangert: This code is doing an `await` in it though, which didn't seem to play nice with map afaik.

chenba: Oh right!  This is the second time today I tripped on awaits in loops in reviews. 😛

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -10,6 +10,597 @@ const ScopeSet = require('../../../fxa-shared').oauth.scopes;
...
+  const { user: uid } = auth.credentials;
+  let email;
+  if (!fetchEmail) {
+    ({ email } = auth.credentials);
+  } else {
+    const account = await db.account(uid);
+    ({ email } = account.primaryEmail);
+  }
+  return { uid, email };
+}
```
chenba: As a caller, I wouldn't know what to do with the `fetchEmail` flag.  I'd get an email address if I set it to true, but I may or may not get an email address if it's `false`?  Or I'd always get an email address but the source of it is based on the flag?

bbangert: This was retained as-is, its just being moved. If you include it, then the email is fetched from the db, vs. extracted from the credentials. An email will always be included, its just how its retrieved. I'm not entirely clear on why it works like this as the function is from a previous PR.

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -10,6 +10,597 @@ const ScopeSet = require('../../../fxa-shared').oauth.scopes;
...
+   *
+   * Note that the invoice *must have the `payment_intent` expanded*
+   * or this function will fail.
+   *
+   * @param {Invoice} invoice
+   * @returns {boolean}
+   */
+  paidInvoice(invoice) {
+    if (typeof invoice.payment_intent === 'string') {
+      throw error.backendServiceFailure('stripe', 'PI fetch');
```
chenba: Maybe expand `PI` in the message.

packages/fxa-auth-server/lib/routes/subscriptions.js
lmorchard: FWIW, this is [an existing util function that got moved](https://github.com/mozilla/fxa/pull/3917/files/0f875831223bfdc39bd92bb6c010511c3cf27007#diff-eedc9e44d94c523680530e9c145c9496L41). I think I wrote the original but I can't quite remember why it works the way it does.

I want to say I found a problem where `auth.credentials` almost never contains `email`, and so `fetchEmail` is true for those routes that absolutely need an email address. We probably could / should remove the bit that checks `auth.credentials`

packages/fxa-auth-server/lib/routes/subscriptions.js
bbangert: Ah, this is a missed type-cast, I'll fix that.

packages/fxa-auth-server/lib/routes/subscriptions.js
lmorchard: More context, since it bothers me not remembering why it works this way... But I think the context is that this is an issue from back in May that I vaguely meant to get back to eventually 😬 

https://github.com/mozilla/fxa/pull/1175#pullrequestreview-240147690

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -10,6 +10,597 @@ const ScopeSet = require('../../../fxa-shared').oauth.scopes;
...
+    const devices = await request.app.devices;
+    await this.push.notifyProfileUpdated(uid, devices);
+    this.log.notifyAttachedServices('profileDataChanged', request, {
+      uid,
+      email,
+    });
+    await this.profile.deleteCache(uid);
+
+    return { subscriptionId };
+  }
```
chenba: Scrolling through these *Subscription functions, it looks like there's quite a bit of common code.  I wonder if we could write a function to create these functions.  Not sure if it's worthwhile though.

bbangert: Once production goes to direct stripe, we're going to remove the other version entirely.

lmorchard: *Some* of this will be rolled up into a common function in my branch that's a follow-on to this PR. Namely, the bits that we have to do when a customer / profile changes:

https://github.com/mozilla/fxa/compare/feat/issue-3809...lmorchard:3808-customer-caching#diff-eedc9e44d94c523680530e9c145c9496R62

PR #3919, opened 2020-01-16
feat(auth-server): update Stripe on email change

### LINKED ISSUE #3813: Update Stripe on primary e-mail changes
Like #1118, if a user has a subscription, we need to update Stripe to reflect the current primary email address. Since users do not frequently change their primary e-mail address, rather than check our cached subscription data, we should query Stripe to see if a customer matching the previous primary email exists, and update that user.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-869)
┆Issue Number: FXA-869

bbangert:
Because:

* We need to keep Stripe up to date on email changes.

This commit:

* Update the customer record in Stripe when the email changes.

Closes #3813

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -47,6 +47,31 @@ async function updateZendeskPrimaryEmail(
   });
 }
 
+/**
+ * Update the primary email in Stripe
+ *
+ * @param {import('../payments/stripe').StripeHelper} stripeHelper
+ * @param {string} uid
+ * @param {string} currentPrimaryEmail
+ * @param {string} newPrimaryEmail
+ * @returns {Promise<void | import('stripe').Stripe.Customer>}
+ */
+async function updateStripeEmail(
+  stripeHelper,
+  uid,
+  currentPrimaryEmail,
+  newPrimaryEmail
+) {
+  const customer = await stripeHelper.fetchCustomer(uid, currentPrimaryEmail);
+  if (!customer) {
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -47,6 +47,29 @@ async function updateZendeskPrimaryEmail(
   });
 }
 
+/**
+ *
+ * @param {import('../payments/stripe')} stripeHelper
+ * @param {string} uid
+ * @param {string} currentPrimaryEmail
+ * @param {string} newPrimaryEmail
```
chenba: Maybe add return type?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -47,6 +47,29 @@ async function updateZendeskPrimaryEmail(
   });
 }
 
+/**
+ *
```
chenba: Remove line or add description?

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -813,26 +837,38 @@ module.exports = (
...
                 newEmail: secondaryEmail.normalizedEmail,
+                system: source,
               });
               Sentry.captureException(err);
             });
-          });
+          };
+
+          // Fire off intentionally without waiting for all the network requests
+          // required to update Zendesk/STripe. Capture enough to manually update
```
chenba: s/STripe/Stripe/ 😄

PR #3932, opened 2020-01-17
feat(avatar spinner): Replace avatar spinner with 'signin' spinner across our views

### LINKED ISSUE #3148: Replace avatar spinner with our best version
We previously had a [fantastic spinner](https://github.com/mozilla/fxa-content-server/pull/3323) made by a contributor that we should use everywhere the profile image is loading. In the recent spinner fix, the small spinner was scaled up, and looks a bit silly.

Please apply to everywhere the avatar has a loading state (avatar modal, sign-in screens, etc.)

CURRENT
Screen Shot 2019-10-28 at 12 21 28 PM

PROPOSED
64e246fe-a2b2-11e5-99ed-aef7a36d6d8a

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-496)
┆Issue Number: FXA-496

LZoog:
fixes #3148

This PR:
* Shows the avatar spinner previously only implemented in the ‘signin' view in every view the avatar is loaded
* In views where there is a `border`/`box-shadow`, adds these styles only after the spinner completes via `spinner-completed`
* Changes 'image-holder' element from `p` tag to`div`
* Cleans up `displayAccountProfileImage`, clears the setTimeout, makes what we're doing with the change-avatar link more obvious
* Tweaks the avatar spinner load times and other CSS, approved by UX
* Adds missing tests and cleans up/modernizes some other tests

I added a 500ms delay for the screenshots so we can see the spinner when it's being fairly slow. It also looks smoother than this, Licecap looks laggy.

profilepic1
profilepic2

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -175,7 +175,7 @@ const View = BaseView.extend({
 
   _setupAvatarChangeLinks() {
     if (this.supportsAvatarUpload()) {
-      this.$('.avatar-wrapper > *').wrap(
+      this.$('.change-avatar-inner').wrap(
         '<a href="/settings/avatar/change" class="change-avatar"></a>'
       );
     } else {
--- a/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/avatar_change.mustache
@@ -1,8 +1,7 @@
 <div id="avatar-change">
   <section class="modal-panel">
     <div class="main-avatar">
-      <div id="loading-avatar-spinner"></div>
-      <p id="image-holder" class="avatar-wrapper"></p>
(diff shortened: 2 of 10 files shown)


packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -175,7 +175,11 @@ const View = BaseView.extend({
 
   _setupAvatarChangeLinks() {
     if (this.supportsAvatarUpload()) {
-      this.$('.avatar-wrapper > *').wrap(
+      const avatarImage = this.$('.avatar-wrapper > img');
+      const avatarElement = avatarImage.length
+        ? avatarImage
+        : this.$('.avatar-wrapper > span');
+      avatarElement.wrap(
```
LZoog: Since we have the spinner displayed in the settings view now, this check for what element to wrap became necessary. `> *` was proving a little buggy, as `.change-avatar` was in some cases appending twice.

`.avatar-wrapper > span` is for when the default image is showing.

PR #3937, opened 2020-01-21
fixed validation string for recovery code confirmation

### LINKED ISSUE #3878: Incorrect error message for empty “Recovery code” field
**Build:**
Stage [Train 1.154.0]

**Affected platforms:**
Windows 10
Mac Os X 10.14.6
Android Q
iOS 13.1.2

**Affected versions:**
Firefox 72.0.1
Fennec 68.4.0
Firefox for iOS 21.0

**Steps to reproduce:**
1. Launch Firefox and login using a Firefox Account.
2. Navigate to Manage account page.
3. Click on "Enable" button from 2FA section.
4. Enter the code received after the QR code was scanned and tap on "Ready".
5. At “Save recovery codes” click on "Confirm" button without entering any code.

**Expected results:**
An error message is displayed saying "Recovery code required".

 **Actual Results:**
"Invalid recovery code" error message is displayed.

Screenshot_4

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-920)
┆Issue Number: FXA-920

dannycoates:
fixes #3878

Screen Shot 2020-01-21 at 11 12 17 AM

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/recovery_codes.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/recovery_codes.js
@@ -79,9 +79,11 @@ const View = FormView.extend({
     const input = this.getElementValue('input.recovery-code');
     const codes = this.model.get('recoveryCodes');
     if (!codes.includes(input.toLowerCase())) {
+      const e =
+        input.length > 0 ? 'INVALID_RECOVERY_CODE' : 'RECOVERY_CODE_REQUIRED';
       return this.showValidationError(
         this.$('input.recovery-code'),
-        AuthErrors.toError('INVALID_RECOVERY_CODE')
+        AuthErrors.toError(e)
       );
     }
 


packages/fxa-content-server/app/scripts/views/settings/recovery_codes.js
```diff
@@ -79,9 +79,11 @@ const View = FormView.extend({
     const input = this.getElementValue('input.recovery-code');
     const codes = this.model.get('recoveryCodes');
     if (!codes.includes(input.toLowerCase())) {
+      const e =
+        input.length > 0 ? 'INVALID_RECOVERY_CODE' : 'RECOVERY_CODE_REQUIRED';
```
vbudhram: Would you be able to check the validation of the input directly? Kind of like https://github.com/mozilla/fxa/blob/d667541a9e2568e331fb348f0997974a201895ac/packages/fxa-content-server/app/scripts/views/form.js#L275.

Since the input has the class `recovery-code` I would think rules in https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/app/scripts/views/elements/recovery-code-input.js#L26 would still apply?

dannycoates: thanks, i saw that but decided against it because it'd split the checking into two places. The static validation in `validate.js` isn't enough. If anything I'd prefer to remove the element validation instead

Issue #3895, opened 2020-01-13
Remove hover and down state for disabled buttons


ryanfeeley:
I noticed that our disabled button style offers a hover and down state. As they don't do anything, we should stop doing this.

Instead we can change the cursor on hover to 🚫 with `cursor: not-allowed;`

![secondary-hover](https://user-images.githubusercontent.com/68213/72292485-7ac77700-361f-11ea-9b25-a801b907a6ad.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-932)
┆Issue Number: FXA-932

farhan787:
@ryanfeeley may I fix this ?

LZoog:
@farhan787 go ahead! :)

farhan787:
Thanks @LZoog

LZoog:
Hey @farhan787, we just merged your PR but it's only a partial fix. There's still a hover state on the disabled button in "secondary email" when you add another email, it darkens the background a tad. We'll need `not(.disabled)` on our hover CSS or something similar. Would you like to take a stab at that? 🙂

![disabled-hover](https://user-images.githubusercontent.com/13018240/72844906-19c52200-3c63-11ea-9541-90c53f212a4a.gif)

farhan787:
@LZoog sure, fixing it.

LZoog:
Patches: #3949, #3946

farhan787:
@clouserw is there still anything needed to fix this issue ?

LZoog:
@farhan787 There's no work remaining, it looks like you weren't assigned before your PRs were merged and we tend to leave ourselves assigned when an issue is closed so I think you were just assigned late. Feel to remove yourself if you prefer. :)

farhan787:
> Feel to remove yourself if you prefer. :)

No, it's ok to get the assigned tag

clouserw:
Thanks @farhan787 !

PR #3949, opened 2020-01-21
fix(style): background darkness fixed on hovering disable button

farhan787:
Hover fix issue #3895

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_button-row.scss
+++ b/packages/fxa-content-server/app/styles/modules/_button-row.scss
@@ -73,6 +73,7 @@ button {
 
   &:disabled,
   &.disabled {
+    background: $button-background-color-default;
     cursor: not-allowed;
     opacity: 0.4;
   }


packages/fxa-content-server/app/styles/modules/_button-row.scss
```diff
@@ -73,6 +73,7 @@ button {
 
   &:disabled,
   &.disabled {
+    background: darken($button-background-color-default-hover, 5);
```
LZoog: I believe `$button-background-color-default` is the same color as this. Can we use that variable instead for consistency?

farhan787: @LZoog yeah, that makes more sense for consistency. Fixed that.

dannycoates:
@farhan787 can you possibly sign this commit?

farhan787:
> @farhan787 can you possibly sign this commit?

oh yeah sure, I forgot to do that...Done.

PR #3959, opened 2020-01-22
fix(service): Show service name when loading reset password with recovery key

### LINKED ISSUE #3842: ServiceName not displayed in “Reset password with recovery key” page
**Build:**
Stage [Train 1.154.0]

**Affected versions**
Firefox 72.0

**Affected platforms**
Windows 10 64-bit

**Steps to reproduce**
1. Go to Sign In Page and tap on "Forget password?"
2. Insert the email and tap on “Begin Reset”.
3. Go to email and tap on "Create new password" button.

**Expected Results**
“Reset password with recovery key” page is correctly displayed.

**Actual Results**
Service Name appears as “%(serviceName)s” 

Screenshot_6

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-894)
┆Issue Number: FXA-894

vbudhram:
Fixes #3842 

@LZoog Took a look at this and it turned out to be a one liner 😬. We needed to import the `ServiceNameMixin` into the view. I'm not really sure when this bug was introduced or regressed.

 Also added some more tests to hopefully catch it next time. r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/account_recovery_confirm_key.js
+++ b/packages/fxa-content-server/app/scripts/views/account_recovery_confirm_key.js
@@ -9,6 +9,7 @@ import FormView from './form';
 import FlowEventsMixin from './mixins/flow-events-mixin';
 import preventDefaultThen from './decorators/prevent_default_then';
 import PasswordResetMixin from './mixins/password-reset-mixin';
+import ServiceNameMixin from './mixins/service-mixin';
 import ResendMixin from './mixins/resend-mixin';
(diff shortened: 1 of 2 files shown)


LZoog:
Oh hey, thanks for looking at this! I think that test needs an `it`, I want to double check it renders `Account Settings` default anyway so I'll check in your branch and make the edit and make sure you approve. 👍

vbudhram:
That sounds good @LZoog, I'll merge this and can expand tests for the default state. Flag for review when ready.

PR #3963, opened 2020-01-23
refactor(support-panel): get subscriptions from auth-server

### LINKED ISSUE #3794: fxa-support-panel enhancement
## Description

Due to the adoption of Architecture 2 being adopted changes need to be made to support the delivered information in Zendesk.  The current API call that provides customer subscription information will cease to exist, therefore this information will need to be returned in the fxa-support-panel.

Please add the following to the fxa-support-panel iFrame:
List of Subscriptions for user:
  Subscription Name
  Subscription Status
  Last Payment Date
  Next Payment Date

Below is a screenshot of the current view for reference:
fxa-support-panel-current

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-857)
┆Issue Number: FXA-857

chenba:
This patch removes the support panel's dependency on SubHub by fetching
a FxA user's subscriptions from the auth server instead.  The requests
are authenticated by a shared secret between the support panel and the
auth server.

Fixes #3794 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -42,7 +42,7 @@ async function run(config) {
   try {
     database = await DB.connect(config[config.db.backend]);
   } catch (err) {
-    log.error({ op: 'DB.connect', err: { message: err.message } });
+    log.error('DB.connect', { err: { message: err.message } });
     process.exit(1);
   }
 
@@ -112,13 +112,11 @@ async function run(config) {
 
   try {
     await server.start();
-    log.info({
-      op: 'server.start.1',
+    log.info('server.start.1', {
       msg: `running on ${server.info.uri}`,
     });
   } catch (err) {
-    log.error({
-      op: 'server.start.1',
+    log.error('server.start.1', {
       msg: 'failed startup with error',
       err: { message: err.message },
     });
@@ -128,7 +126,7 @@ async function run(config) {
     server,
     log: log,
     async close() {
-      log.info({ op: 'shutdown' });
+      log.info('shutdown');
       await server.stop();
       await customs.close();
       oauthdb.close();
@@ -138,10 +136,7 @@ async function run(config) {
         senders.email.stop();
       } catch (e) {
         // XXX: simplesmtp module may quit early and set socket to `false`, stopping it may fail
-        log.warn({
-          op: 'shutdown',
-          message: 'Mailer client already disconnected',
-        });
+        log.warn('shutdown', { message: 'Mailer client already disconnected' });
       }
       await database.close();
     },
@@ -153,14 +148,11 @@ async function main() {
   try {
     const server = await run(config.getProperties());
     process.on('uncaughtException', err => {
-      server.log.fatal(err);
+      server.log.fatal('uncaughtException', err);
       process.exit(8);
     });
     process.on('unhandledRejection', (reason, promise) => {
-      server.log.fatal({
-        op: 'promise.unhandledRejection',
-        error: reason,
-      });
+      server.log.fatal('promise.unhandledRejection', { error: reason });
     });
     const shutdown = async () => {
       await server.close();
@@ -170,7 +162,7 @@ async function main() {
     server.log.on('error', shutdown);
 
     if (config.get('env') !== 'prod') {
-      server.log.info(config.toString(), 'starting config');
+      server.log.info('startConfig', { config: config.toString() });
     }
   } catch (err) {
     console.error(err); // eslint-disable-line no-console
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -1604,6 +1604,15 @@ const conf = convict({
       },
     },
   },
+  supportPanel: {
+    secretBearerToken: {
+      default: 'YOU MUST CHANGE ME',
+      doc:
+        'Shared secret to access certain endpoints.  Please only use for GET.  No state mutation allowed!',
+      env: 'SUPPORT_PANEL_AUTH_SECRET_BEARER_TOKEN',
+      format: 'String',
+    },
+  },
 });
 
 // handle configuration files.  you can specify a CSV list of configuration
@@ -1735,6 +1744,7 @@ if (conf.get('isProduction')) {
     'oauth.jwtSecretKeys',
     'oauth.secretKey',
     'profileServer.secretBearerToken',
+    'supportPanel.secretBearerToken',
   ];
   for (const key of SECRET_SETTINGS) {
     if (conf.get(key) === conf.default(key)) {
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -7,12 +7,14 @@ const subhub = require('../subhub/client');
 
 const stripe = require('stripe').Stripe;
 
-/** @typedef {import('stripe').Stripe.Customer} Customer */
-/** @typedef {import('stripe').Stripe.Product} Product */
-/** @typedef {import('stripe').Stripe.Plan} Plan */
-/** @typedef {import('stripe').Stripe.Subscription} Subscription */
 /** @typedef {import('stripe').Stripe.ApiList<Subscription>} Subscriptions */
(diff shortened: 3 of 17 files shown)


packages/fxa-auth-server/lib/routes/utils/subscriptions.js
```diff
@@ -72,7 +72,7 @@ const SubscriptionUtils = (module.exports = {
 async function fetchSubscribedProductsFromStripe(uid, stripeHelper, email) {
   const customer = await stripeHelper.customer(uid, email);
   if (!customer || !customer.subscriptions.data) {
-    return;
+    return [];
```
chenba: The `undefined` returned from here could overwrite the default set at https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/utils/subscriptions.js#L38, causing errors when later code call array functions on it (`includes`, `reduce`, whathaveyou).

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -112,13 +112,11 @@ async function run(config) {
 
   try {
     await server.start();
-    log.info({
-      op: 'server.start.1',
+    log.info('server.start.1', {
```
lmorchard: Should other log calls be updated in this file?

Looks like most others in this file follow the old pattern (e.g [like here during shutdown](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/bin/key_server.js#L140)) and that doesn't seem to match [the mozlog module docs](https://www.npmjs.com/package/mozlog#usage).

packages/fxa-auth-server/config/index.js
```diff
@@ -1604,6 +1604,15 @@ const conf = convict({
       },
     },
   },
+  supportPanel: {
+    secretBearerToken: {
+      default: 'YOU MUST CHANGE ME',
```
lmorchard: Since this *must* have a setting, it should probably be added to [the list that aborts startup](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/config/index.js#L1739) if the setting is left as default. 

Also, we should add a note to the next train's deploy doc that this config change needs doing.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -217,6 +219,17 @@ class StripeHelper {
...
   }
 
+  /**
+   * Fetch a list of subscriptions for a customer from Stripe.
+   *
+   * @param {string} customerId
+   * @param {SubscriptionListParams} options
+   * @returns {Promise<Subscriptions>}
+   */
+  async fetchAllSubscriptionsForCustomer(customerId, options = {}) {
```
lmorchard: Do we need this method? We have access to the cached customer record in [`customer()`](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/lib/payments/stripe.js#L193) and non-cached fetch via [`fetchCustomer()`](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/lib/payments/stripe.js#L165) - both of these [include subscriptions](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/lib/payments/stripe.js#L215). I'd say ideally use the cached customer data, but you can use the direct fetch if the latest data is absolutely necessary. 

(I might also suggest that if stale data is a concern during support requests, we should call `deleteCachedCustomer()` somewhere in that process to refresh the cache)

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -440,6 +442,65 @@ class DirectStripeRoutes {
 
     return {};
   }
+
+  /**
+   * Get a list of subscriptions with a FxA UID and email address.
```
lmorchard: We have existing API routes for [`listActive`](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/lib/routes/subscriptions.js#L345) and [`getCustomer`](https://github.com/mozilla/fxa/blob/9589fc7906810c085ce755852f42804476dca8e5/packages/fxa-auth-server/lib/routes/subscriptions.js#L371) - do neither of these work for this purpose? 

The `listActive` route is kind of "legacy" since it was the view of the FxA DB table, made obsolete by moving to direct Stripe access but still used on payments-server. The customer data should list subscriptions, though

lmorchard: Alternatively, if we really *do* need this new API route - I'd say maybe drop the use of `subscriptionsToResponse` on `stripeHelper` and return the Stripe data directly. I think `subscriptionsToResponse` is a kind of legacy adapter to make data look like subhub responses, but we can move away from that. (@bbangert could probably speak to that better, e.g. maybe we need to do some sanitizing for privacy / security on this data as it passes by)

packages/fxa-auth-server/bin/key_server.js
chenba: Good catch.  I thought these two calls were from recent changes, but they were probably just formatting changes?  I'll update the other log calls.

packages/fxa-auth-server/lib/payments/stripe.js
chenba: This is called _only_ when the subscriptions list in customer has the `has_more` set to `true`.  I'm handling that for completeness sake, but practically no customer will have that many subscriptions any time soon.  (I hope I'm very wrong on that though. 😁 )

packages/fxa-auth-server/lib/routes/subscriptions.js
chenba: D'oh.... `fetchCustomer` isn't what I needed here; I'll change it to `customer`.  Thanks!

As for the new route, I landed on that (after some struggling/indecisiveness) because:
 - Hapi doesn't offer a way to do multiple auth strategies if both of the strategies use the same header scheme (the `authorization` header in this case); we would need to write a single strategy that check both and auth on the first valid token.
 - Since the requests don't look the same, we'd need to move the `uid` and `email` query param validation somewhere, instead of declaring them with the route.

If there's a clean way to have a single route in Hapi, I'd love to know.

packages/fxa-auth-server/lib/payments/stripe.js
lmorchard: > This is called only when the subscriptions list in customer has the has_more set to true. 

Hmm, if we're missing subscriptions in `fetchCustomer`, we need to fix that. That call *should* include *all* subscriptions because the code that determines enabled capabilities counts on that. (i.e. if any subscriptions are missing, RPs will not get the data in the profile that enables the service)

If *that's* the reason for this additional method, it sounds like a bug that needs to be fixed instead.

chenba: > Hmm, if we're missing subscriptions in `fetchCustomer`, we need to fix that. That call _should_ include _all_ subscriptions...

Ok, let's make that change then.

chenba: Augmented `fetchCustomer`: https://github.com/mozilla/fxa/pull/3963/files#diff-26644f246b5f22f6205561269d539ce7R182-R190

packages/fxa-auth-server/lib/routes/auth-schemes/shared-secret.js
```diff
@@ -0,0 +1,24 @@
...
+
+const constantTimeCompare = (subject, object) => {
+  const size = Buffer.byteLength(object);
+  return (
+    crypto.timingSafeEqual(Buffer.alloc(size, subject), Buffer.from(object)) &&
+    subject.length === object.length
+  );
+};
+
+exports.strategy = secret => (server, options) => ({
```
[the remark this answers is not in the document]
chenba: There was: https://github.com/mozilla/fxa/pull/3963/files#diff-58b4017d78df28736c1720cf87347012L350-L360

I moved that code in here for reuse.

PR #3964, opened 2020-01-23
 test(account-recovery): Modify 'render view' test for default serviceName

LZoog:
Follow-up to #3959

Straightforward changes:

* Change `serviceNameMixin` references to `serviceMixin` to match all other imports of this file
* Import the `relier` instead of the `base` relier in the test file (this was the cause of some "wut" feelings earlier today, heh 🤦‍♀️)
* Update test to check for the default `serviceName` from the relier which covers the original issue. Setting other services on the relier are covered in other tests, it seemed superfluous to have two tests here for the relier.

@vbudhram tagging you for an r+ since you wrote the test.

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/views/account_recovery_confirm_key.js
+++ b/packages/fxa-content-server/app/tests/spec/views/account_recovery_confirm_key.js
@@ -8,7 +8,7 @@ import Broker from 'models/auth_brokers/base';
 import AuthErrors from 'lib/auth-errors';
 import Metrics from 'lib/metrics';
 import Notifier from 'lib/channels/notifier';
-import Relier from 'models/reliers/base';
+import Relier from 'models/reliers/relier';
 import SentryMetrics from 'lib/sentry';
 import sinon from 'sinon';
 import TestHelpers from '../../lib/helpers';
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/tests/spec/views/account_recovery_confirm_key.js
```diff
@@ -84,12 +81,12 @@ describe('views/account_recovery_confirm_key', () => {
     });
   });
 
-  it('renders view', () => {
+  it('renders the view with the default serviceName from the relier', () => {
```
vbudhram: Looks like this test already exists in [service-mixin](https://github.com/mozilla/fxa/blob/2ddb356924e344e661a841439f349d57cff8e6ef/packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js#L68), guess it wouldn't hurt.

LZoog: 👍 It does, but we also cover setting a non-default `serviceName` on the relier in [a few](https://github.com/mozilla/fxa/blob/2ddb356924e344e661a841439f349d57cff8e6ef/packages/fxa-content-server/app/tests/spec/views/mixins/service-mixin.js#L65) [other](https://github.com/mozilla/fxa/blob/2ddb356924e344e661a841439f349d57cff8e6ef/packages/fxa-content-server/app/tests/spec/views/reset_password.js#L103) [views](https://github.com/mozilla/fxa/blob/2ddb356924e344e661a841439f349d57cff8e6ef/packages/fxa-content-server/app/tests/spec/views/sign_in_password.js#L87). "Account Settings" is what I see when hitting that screen through the e-mail link and checking for it in a test is what caused me to realize the wrong relier file was being imported to capture the default. Probably more important that we have a test here at all than default vs non-default though just to make sure the fixed issue doesn't reoccur. 🤷‍♀️

PR #3971, opened 2020-01-24
feat(support-panel): display geo info

### LINKED ISSUE #2677: [Support] Display IP/Geo-region of prior service access for a user
For verification by support agents, our support-panel should display the services a user has accessed, and their IP/geo-region of access based on the oauth tokens issued.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-389)
┆Issue Number: FXA-389

chenba:
This patch displays all the available geo info from a user's session
token meta data in Support Panel.  A new route is added to auth server
to support this feature.

Fixes #2677 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -32,6 +32,11 @@ async function run(config) {
     stripeHelper = createStripeHelper(log, config);
   }
 
+  const redis = require('../lib/redis')(
+    { ...config.redis, ...config.redis.sessionTokens },
+    log
+  );
+
   const DB = require('../lib/db')(
     config,
     log,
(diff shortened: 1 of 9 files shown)


chenba:
Note that this PR contains some identical changes that are also in #3963.  This is because they both involve the support panel fetching some data from auth server.

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -32,6 +32,11 @@ async function run(config) {
     stripeHelper = createStripeHelper(log, config);
   }
 
+  const redis = require('../lib/redis')(
```
lmorchard: Hmm, we also require & instantiate a redis client in `lib/payments/stripe.js`... I wonder if this would be worth a follow up issue to find where all we use redis and have them share this top-level instance?

chenba: They use different configs, the most important of which is probably the key prefix string.  I do agree that we should sort that out so they can share an instance.

PR #3972, opened 2020-01-24
fix(codes): Verify session first then send push notifications

### LINKED ISSUE #3970: Finalize account state before sending "account verified" push notification
This is a fun one, and a little hard to believe, but it's the best explanation I have for what we're seeing in [Bug 1609166](https://bugzilla.mozilla.org/show_bug.cgi?id=1609166).

When verifying a new account via email codes, we call the `/session/verify_code` endpoint, which does the following in a non-transactional manner:

1. Check that the submitted code is valid.
2. Update the db to mark the account as verified.
3. Send a push notification to tell the browser the account is verified.
4. Update the db to mark the session as verified.

If the server is slow enough and the user's network is fast enough, it's possible that their browser will receive the push notification and poll for the status of the account after step 3, but before step 4. If this happens then the browser will see an incorrect transient account state with `emailVerified=true` but `sessionVerified=false`.

Due to a related client bug, the browser will stop polling for account status updates when it receives the push notification from step 3, so it will never discover the correct account state after step 4 finishes.

We're going to fix the client behaviour to keep polling in this case, but I'd argue this is actually a server bug, and it shouldn't be telling the client any transient account states.

The classically-correct fix here would be to ensure that all db changes for a single request occur within the scope of a single transaction, but that's very difficult in our current architecture. Perhaps instead we could arrange for the push notification to only go out after all db updates have been finalized, to reduce the chance of clients reading partial states in practice.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-974)
┆Issue Number: FXA-974

vbudhram:
Fixes #3970 

Simple fix based on rfk recommendation of reordering the order that the verification events and notifications happen. Make sure that the session and account are verified and then send push notifications to connected devices.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/session.js
+++ b/packages/fxa-auth-server/lib/routes/session.js
@@ -354,12 +354,14 @@ module.exports = function(
         const options = request.payload;
         const sessionToken = request.auth.credentials;
         const { code } = options;
+        const { uid } = sessionToken;
+        const devices = await request.app.devices;
 
         request.emitMetricsEvent('session.verify_code');
 
         // Check to see if the otp code passed matches the expected value from
         // using the account's' `emailCode` as the secret in the otp code generation.
-        const account = await db.account(sessionToken.uid);
+        const account = await db.account(uid);
         const secret = account.primaryEmail.emailCode;
 
         const isValidCode = otpUtils.verifyOtpCode(code, secret, otpOptions);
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -373,16 +374,19 @@ module.exports = function(
         if (!account.primaryEmail.isVerified) {
           await signupUtils.verifyAccount(request, account, options);
```
rfk: IIRC there is another push notification sent somewhere inside `verifyAccount` here, which could also trigger this issue.

packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -368,20 +370,22 @@ module.exports = function(
...
-          await push.notifyAccountUpdated(uid, devices, 'accountConfirm');
         }
 
-        // If a valid code was sent, this verifies the session using the `email-2fa` method.
-        // The assurance level will be ["pwd", "email"] or level 1.
-        await db.verifyTokensWithMethod(sessionToken.id, 'email-2fa');
+        // To avoid any potential race conditions, the very last thing done is send
+        // a push notifications to connected devices that the account and
+        // session were verified.
+        await push.notifyAccountUpdated(uid, devices, 'accountConfirm');
```
rfk: With the latest update, I think this could stay in the `else` branch - you don't want to end up sending *both* the `accountVerify` notification and the `accountConfirm` notification in the case where `!account.primaryEmail.isVerified` is true.

packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -368,21 +370,19 @@ module.exports = function(
           throw error.invalidOrExpiredOtpCode();
         }
 
+        // If a valid code was sent, this verifies the session using the `email-2fa` method.
+        // The assurance level will be ["pwd", "email"] or level 1.
```
rfk: For future readers, it may be worth adding a note here to say that we do this first before verifying the account, because we want it to hit the DB before we send any push notifications.

vbudhram:
@rfk Thanks for review, I've made the changes 👍

Issue #3994, opened 2020-01-30
Ensure test coverage metrics across packages are visible and useful


lmorchard:
Seems like we're not sure how to see test coverage metrics for many packages in FxA. Look into this, ensure the metrics are visible and useful for understanding current coverage

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-994)
┆Issue Number: FXA-994

PR #4016, opened 2020-01-31
feat(coverage): Add coveralls coverage

vbudhram:
Adds test coverage reporting to coveralls for

- fxa-auth-server
- fxa-auth-db-mysql
- fxa-profile-server

You can view current coverage at https://coveralls.io/github/mozilla/fxa. The reports are also uploaded as test artifacts in circleci.

Connects to https://github.com/mozilla/fxa/issues/3994

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,9 @@ jobs:
       db:
         type: boolean
         default: false
+      upload_coverage:
+        type: boolean
+        default: false
     working_directory: ~/project/packages/<< parameters.module >>
     steps:
       - attach_workspace:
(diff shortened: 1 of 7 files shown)


bbangert:
Looks like a bunch of these don't seem to have the /app/reports expected so the build step fails.

vbudhram:
@bbangert The problem I am encountering is coverage files are created in the docker container but circleci can't upload those files artifacts since it doesn't have access to inside of them. I've seen a few articles about copying files from container to circleci but they didn't quite work.

I might punt this for now until I can get to take a deeper look into it.

dannycoates:
@vbudhram db-mysql works https://143971-36831081-gh.circle-artifacts.com/0/reports/lcov-report/index.html but not all the other packages create `/app/reports`

vbudhram:
@lmorchard I'm going to merge this in but not mark the issue fixed. This initial version adds the basic structure for getting code coverage reported and more visible.

PR #4033, opened 2020-02-03
fix(viewport): Update viewport metatag to include minimum-scale, and make value consistent across all instances

### LINKED ISSUE #3887: add minimum-scale=1.0 and initial-scale=1.0 to the FxA signup page [bz1599582]
From https://bugzilla.mozilla.org/show_bug.cgi?id=1599582

Created attachment 9111737
Screenshot_20191126-091227_Firefox.jpg

Steps to reproduce: 
Install Firefox, sign out of FxA (if signed in), clear cache.
Open the following deeplink URL with Firefox set as your default browser: firefox://fxa-signin?entrypoint=fx-push-notification%26utm_source=leanplum-push-notification-test%26utm_campaign=android-push-notification-test

Expected behavior: FxA signup page resolves normally.
Actual behavior: FxA signup page resolves but is very small (screenshot attached).

Tested using: 
Device: Galaxy Note 10+
Android OS Version: 9

Note that the deeplink schema was tested and validated in this ticket:
https://bugzilla.mozilla.org/show_bug.cgi?id=1522704

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-927)
┆Issue Number: FXA-927

LZoog:
fixes #3887 

I could not reproduce this in an emulator, only on my old physical Android device (and note: FF also does not have to be the default browser).

`minimum-zoom=1` should* disable zooming _out_ on Android, but from an a11y standpoint, this is acceptable because we still allow users to zoom _in_ up to 200%. I don't quite understand why a deep link would seemingly ignore our already-present `width=device-width,initial-scale=1` metatag, but this appears to fix the problem for me on my physical device.

*I'm not sure why, but I was not able to zoom out on the sign up page _before_ the change, and iOS appears to ignore the scaling tags. In other words, on my devices, our pages act identically on iOS and Android before and after this change but this appears to fix the problem described in the bug.

I went ahead and made our `viewport` metatag consistent across all of our pages.

### DIFF
--- a/packages/123done/static/about.html
+++ b/packages/123done/static/about.html
@@ -5,7 +5,7 @@
     <title>123done - your tasks, simplified</title>
     <meta type="description" content="" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes" />
     <link type="text/css" rel="stylesheet" href="/en/drsugiyama/fonts.css" />
     <link rel="stylesheet" href="css/styles.css" type="text/css" />
     <link rel="stylesheet" href="css/bootstrap.css" type="text/css" />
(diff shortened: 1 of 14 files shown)


LZoog:
👍 Thanks Vijay, agreed.

PR #4057, opened 2020-02-05
feat(auth-server): add hapi route metric timings

### LINKED ISSUE #4056: Add request timing metrics for auth-server
There's a hapi-statsd plugin that should work fine for our needs.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1039)
┆Issue Number: FXA-1039

bbangert:
Because:

* We'd like to know what are the slow/fast routes in our application.

This commit:

* Add a prePostResponse metric emitter modeled on hapi-statsd that works
  with our hotshots based metric reporter.

Closes #4056

### DIFF
--- a/packages/fxa-auth-server/bin/key_server.js
+++ b/packages/fxa-auth-server/bin/key_server.js
@@ -113,7 +113,8 @@ async function run(config) {
     routes,
     database,
     oauthdb,
-    translator
+    translator,
+    statsd
   );
 
   try {
--- a/packages/fxa-auth-server/lib/server.js
+++ b/packages/fxa-auth-server/lib/server.js
@@ -53,7 +53,16 @@ function logEndpointErrors(response, log) {
   }
 }
 
-async function create(log, error, config, routes, db, oauthdb, translator) {
+async function create(
+  log,
+  error,
+  config,
+  routes,
+  db,
+  oauthdb,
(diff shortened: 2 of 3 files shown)


bbangert:
I was looking for where to add a test, but we don't have tests for `server.js`, so I did verify locally with `nc` that we are emitting stats:

```
fxa-auth-server.v1_verify.POST.200:1|c
fxa-auth-server.v1_verify.POST.200:6|ms
fxa-auth-server.v1_recoveryKey_exists.POST.200:1|c
fxa-auth-server.v1_recoveryKey_exists.POST.200:12|ms
fxa-auth-server.v1_recovery_email_status.GET.200:1|c
fxa-auth-server.v1_recovery_email_status.GET.200:6|ms
```

Is what the statsd output looks like.

sciurus:
A couple suggestions

The metric name should describe what is being measured, and you should use tags for the dimensions that people want to slice and dice the data on. So instead of a metric named `fxa-auth-server.v1_verify.POST.200` you'd have one named `fxa-auth-server.response` with tags for route, http method, and status code.

You don't need to submit a counter when you're already submitting a timer. stats will keep track of the number of times you submit the timer and record that as the `count` alongside the other summary stats it records for it.

PR #4062, opened 2020-02-06
feat(logging): Log details of the grant when generating OAuth tokens.

rfk:
Because:

* We need to understand the current rate of OAuth token creation in
  order to understand the potential operational impact of new clients
  such as Fenix.
* The current web logs don't provide enough information to determine
  the volume of tokens created through each different grant type.

This commit:

* Logs an explicit event when creating OAuth tokens, with the appropriate
  details such as grant type and scope.

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/grant.js
+++ b/packages/fxa-auth-server/lib/oauth/grant.js
@@ -162,6 +162,14 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
 // This function does *not* perform any authentication or validation, assuming that
 // the specified grant has been sufficiently vetted by calling code.
 module.exports.generateTokens = async function generateTokens(grant) {
+  logger.info('oauth.generateTokens', {
+    grantType: grant.grantType,
+    keys: !!grant.keysJwe,
+    scope: grant.scope.getScopeValues(),
+    service: hex(grant.clientId),
+    resource: grant.resource,
+  });
+
   // We always generate an access_token.
   const access = await exports.generateAccessToken(grant);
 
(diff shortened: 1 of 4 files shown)


rfk:
@6a68 I picked you for review here because it's logging/metrics related, but feel free to delegate :-)

bbangert:
We do have a statsd object in auth-server now that can emit metrics (optionally with tags). I also just added a route metric today that has counts of specific routes/methods and their response time. Do we want to be logging these, or would statsd style metrics be sufficient?

rfk:
> Do we want to be logging these, or would statsd style metrics be sufficient?

Good question. Logging these puts them in bigquery next to the other data I've been poking at for this effort, which is probably simpler *for me*. I'd like to be able to analyze counts by grant type, client id, etc. But I'm open to the suggestion that statsd is a more appropriate place for this.

bbangert:
The main thing I'm considering is that if we want to see these on a dashboard, statsd will put them in influxdb where we can put them on grafana. We can then watch them by grant type/client id over time. I assume the cardinality of the grant type + client id is relatively low?

I would also assume it should be fine to include the actual client id, rather than hex it.

rfk:
> I would also assume it should be fine to include the actual client id, rather than hex it.

Hexing it is definitely an artifact of the log lines going out as JSON (although I'd argue that a lot of this code would be simpler if we treated the hex string as the canonical value, since that's what is always seen by clients).

rfk:
> grant type + client id is relatively low?

There are only 3 possible grant types, but each OAuth client could potentially use more than one grant type over its lifetime.

rfk:
From the meeting today: we'd like to get this PR as-is into the point-release of FxA that's being tagging imminently, and we can follow up on statsd metrics for this separately.

PR #3957, opened 2020-01-22
feat(keys): Add ability to enable/disable recovery key

vbudhram:
Fixes #3936 

This PR adds the DB support and auth-server support for enabling a recovery key. The enable flag is specified when creating the recovery key.

Some notes

* Added `enabled` column on recoveryKey table
* New db method `updateRecoveryKey_1` to update column value
* New route `/recoveryKey/verify` can enable a recovery key

Issue #3936, opened 2020-01-21
Add ability to confirm recovery key before enabling it


vbudhram:
From UX in https://mozilla.invisionapp.com/share/Q3UK63V4PKA#/screens/390543494_Recovery_Key_Recommendation, the recovery key needs to be confirmed before enabling.

* Add database migration to add column `enabled`
* Expose endpoint to enabled recovery key

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-959)
┆Issue Number: FXA-959

PR #4066, opened 2020-02-06
feat(codes): Add auth-server verify recovery key route

vbudhram:
Builds off https://github.com/mozilla/fxa/pull/3957 which should be reviewed first.
Connects to #3936 

This PR adds the supporting auth-server routes to verify a recovery key before enabling it. Marking as draft until I can clean the code up a bit.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/recovery-key.js
+++ b/packages/fxa-auth-server/lib/routes/recovery-key.js
@@ -21,6 +21,7 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
           payload: {
             recoveryKeyId: validators.recoveryKeyId,
             recoveryData: validators.recoveryData,
+            enabled: isA.boolean().default(true),
           },
         },
       },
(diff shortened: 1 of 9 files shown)


packages/fxa-auth-server/lib/routes/recovery-key.js
```diff
@@ -34,35 +35,73 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
...
+        if (sessionToken.tokenVerificationId) {
+          throw errors.unverifiedSession();
+        }
+
+        const { uid } = sessionToken;
+        const { recoveryKeyId } = request.payload;
+
+        // Attempt to retrieve a recovery key, if it exists and is not already enabled,
+        // then we enable it.
+        const recoveryKeyData = await db.getRecoveryKey(uid, recoveryKeyId);
```
rfk: A quick drive-by to say, it looks like this route would let you test whether a given `recoveryKeyId` is valid, so should probably be subject to the same rate-limiting as other routes like `GET /recoveryKey/{recoveryKeyId}`.

vbudhram: Ah yea that makes sense, I will make those changes. In meantime I am going to merge these changes into https://github.com/mozilla/fxa/pull/3957.

PR #4068, opened 2020-02-06
chore(mem): Remove auth server db memory database

### LINKED ISSUE #745: [fxa-auth-db-mysql] Drop the memory db
Post-docker there's no utility in the memory db. Deleting would mean we don't have to implement everything twice and remove the possibility of behavioural discrepancies between the different db backends.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-792)
┆Issue Number: FXA-792

vbudhram:
Removed the auth-server db memory db, lets see what breaks!

Thanks for the help @dannycoates! I've squashed commits and think this is ready for review.

Fixes #745


packages/fxa-auth-db-mysql/index.js
```diff
@@ -6,10 +6,12 @@ var config = require('./config');
 var dbServer = require('./db-server');
 var error = dbServer.errors;
 var logger = require('./lib/logging')('bin.server');
-var DB = require('./lib/db/mem')(logger, error);
+var DB = require('./lib/db/mysql')(logger, error);
 
 module.exports = function() {
   return DB.connect(config).then(function(db) {
-    return dbServer.createServer(db);
+    const server = dbServer.createServer(db);
```
vbudhram: This was the key why mysql kept having too many connections, the database was not being closed when the test finished.

packages/fxa-auth-server/test/remote/verifier_upgrade_tests.js
```diff
@@ -24,84 +23,55 @@ const DB = require('../../lib/db')(
...
-        })
-        .then(() => {
-          try {
-            db_server.close();
-          } catch (e) {
-            // This connection may already be dead if a real mysql server is
-            // already bound to :8000.
-          }
-        });
+  it('upgrading verifierVersion upgrades the account on password change', async () => {
```
vbudhram: I had to update this test case to correctly close db connection, while in there I updated to latest syntax.

packages/fxa-auth-server/test/test_server.js
```diff
@@ -84,23 +84,34 @@ TestServer.prototype.start = function() {
...
+    await maybeServer.stop();
+    maybeServer = undefined;
+  }
+
+  if (currentServer) {
+    await currentServer.stop();
+    currentServer = undefined;
+  }
+
+  if (currentDBServer) {
```
vbudhram: Added some extra checks to make sure the db gets closed.

packages/fxa-auth-server/scripts/test-local.sh
```diff
@@ -7,7 +7,7 @@ if [ -z "$CORS_ORIGIN" ]; then export CORS_ORIGIN="http://foo,http://bar"; fi;
 
 set -u
 
-DEFAULT_ARGS="-R dot --recursive --timeout 5000 --exit"
+DEFAULT_ARGS="--recursive --timeout 5000 --retries 2 --exit"
```
dannycoates: why retries?

vbudhram: Removed, this was a debugging leftover.

PR #4069, opened 2020-02-06
fix(auth-server): handle incomplete subscription

### LINKED ISSUE #4019: Users can have access to the VPN even if they paid with a card not issues in the US
#### Affected versions:
- All Firefox

#### Affected Platforms:
- Ubuntu 19.04 (x64)
- Windows 10 (x64)

#### Preconditions
- Have a FxA account without a subscriptions

#### Steps to reproduce
1. Open the [VPN page](https://stage.guardian.nonprod.cloudops.mozgcp.net/vpn) and 
2. Click on the Subscribe Now button and complete the subscription process using a credit card number outside of the US. (eg. 6200-0000-0000-0005)
3. Go back to the [VPN page](https://stage.guardian.nonprod.cloudops.mozgcp.net/vpn) and click on the subscribe now button.

#### Expected result:
- The payment page is displayed. 

#### Actual result:
- You are redirected to the "Your subscription is ready" page. (this is an ongoing loop until you click on the "No thanks, just take me to my product." hyperlink because you are signed out after reloading the page).
- After clicking the "No thanks, just take me to my product." hyperlink you have access to download the VPN even if the card used was outside of US.

GIF

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1012)
┆Issue Number: FXA-1012

bbangert:
Because:

* Incomplete subscriptions have never had a valid payment so they
  should be ignored by the client.
* Paying for a subscription when a prior incomplete is present throws
  an error instead of returns, which isn't what the Stripe docs
  indicate.

This commit:

* Screens incomplete subscriptions so the client is not made aware of
  them.
* Properly handles fetching the payment intent, updating it, and
  capaturing errors thrown if another declined card is used.

Fixes #4019

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -133,13 +133,21 @@ class DirectStripeRoutes {
     if (subscription && subscription.latest_invoice) {
       let invoice = /** @type {Invoice} */ (subscription.latest_invoice);
       if (invoice.status === 'open') {
-        const payment_intent =
-          /** @type {PaymentIntent} */ (invoice.payment_intent);
+        const payment_intent = await this.stripeHelper.stripe.paymentIntents.retrieve(
+          /** @type {string} */ (invoice.payment_intent)
+        );
         if (payment_intent.status === 'requires_payment_method') {
           // Re-run the payment
-          invoice = await this.stripeHelper.stripe.invoices.pay(invoice.id, {
-            expand: ['payment_intent'],
-          });
+          try {
+            invoice = await this.stripeHelper.stripe.invoices.pay(invoice.id, {
+              expand: ['payment_intent'],
+            });
+          } catch (err) {
+            if (err.code === 'card_declined') {
+              throw error.paymentFailed();
+            }
+            throw err;
+          }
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -368,13 +376,15 @@ class DirectStripeRoutes {
...
           plan: { product: productId },
         } = subscription;
-        activeSubscriptions.push({
-          uid,
-          subscriptionId,
-          productId,
-          createdAt: created * 1000,
-          cancelledAt: canceled_at ? canceled_at * 1000 : null,
-        });
+        if (['trialing', 'active', 'past_due'].includes(subscription.status)) {
```
jackiemunroe: Just double-checking: do we allow access to products when subscriptions are in a `past_due` state?

jackiemunroe:
Failing test in auth service:


```
remote subscriptions:
       config.subscriptions.enabled = true and direct stripe access:
         with a subscription
           should return active subscriptions:
      AssertionError: expected [] to have a length of 1 but got 0
      + expected - actual
      -0
      +1
at Context.<anonymous> (test/remote/subscription_tests.js:226:16)

PR #4070, opened 2020-02-06
fix(auth-server): screen profile stripe subs

bbangert:
Because:

* Direct stripe integration wasn't screening valid subsfrom the profile
  server results.
* Profile server could erroroneously indicate a user has a subscription
  valid when they do not.

This commit:

* Properly screens the subscriptions for the profile server.

Fixes FXA-1045

### DIFF
--- a/packages/fxa-auth-server/lib/routes/utils/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/utils/subscriptions.js
@@ -69,6 +69,13 @@ const SubscriptionUtils = (module.exports = {
   },
 });
 
+/**
+ * Fetch subscribed products from Stripe and format appropriately
+ *
+ * @param {string} uid
+ * @param {import('../../payments/stripe').StripeHelper} stripeHelper
+ * @param {string} email
+ */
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/utils/subscriptions.js
```diff
@@ -78,9 +85,10 @@ async function fetchSubscribedProductsFromStripe(uid, stripeHelper, email) {
   const subscribedProducts = [
     // All accounts get this psuedo-product
     PRODUCT_REGISTERED,
-    ...customer.subscriptions.data.map(
-      ({ plan: { product: productId } }) => productId
-    ),
+    ...customer.subscriptions.data
+      // TODO: Centralize subscription filtering logic for active subs
```
lmorchard: This might be worth an `activeSubscriptionsForCustomer` method in StripeHelper someday, but 🤷‍♂ this works for now

bbangert: Yea, mainly if Stripe add's more, as its pretty clear-cut at the moment.

PR #4076, opened 2020-02-07
chore(deps): update babel monorepo

lmorchard:
This is an attempt to fix #4020 from renovate, which might not work.

After wrestling with it off & on for a few days, I finally realized that renovate does not like being corrected and will force-push to clobber any changes to its PRs. So, I'm going to try this PR in a different branch and see if it passes tests


lmorchard:
Oof, keeps failing on mocha tests in CI that don't fail locally. Will close now and come back to this later

PR #4081, opened 2020-02-07
fix(auth-server): use tags for paths/method

bbangert:
Because:

* We can't bracket on metric name, only tags.

This commit:

* Uses tags for path/method/status code so that we can select all
  and then filter results as desired by path/method.

### DIFF
--- a/packages/fxa-auth-server/lib/server.js
+++ b/packages/fxa-auth-server/lib/server.js
@@ -411,7 +411,6 @@ function defineLazyGetter(object, key, getter) {
  *    https://github.com/mac-/hapi-statsd/blob/master/lib/hapi-statsd.js
  */
 function metricFactory(statsdClient) {
-  const template = '{path}.{method}.{statusCode}';
   const pathSeparator = '_';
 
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/server.js
```diff
@@ -439,14 +438,11 @@ function metricFactory(statsdClient) {
...
-      .replace('{method}', request.method.toUpperCase())
-      .replace('{statusCode}', statusCode);
-
-    statName = statName.indexOf('.') === 0 ? statName.substr(1) : statName;
-    statsdClient.increment(statName);
-    statsdClient.timing(statName, Date.now() - startDate.getTime());
+    statsdClient.timing('url_request', Date.now() - startDate.getTime(), 1, {
+      path: normalizePath(path),
+      method: request.method.toUpperCase(),
+      statusCode,
```
sciurus:
```suggestion
      status_code: statusCode,
```

bbangert: Is this a influxdb specific thing?

PR #4083, opened 2020-02-08
test(subscription routes) - add tests for direct stripe integration

### LINKED ISSUE #3809: Implement direct Stripe API calls

### LINKED ISSUE #4073: Finish Architecture 2.0 tests
Let's finish the tests that came out of FXA-866

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1046)
┆Issue Number: FXA-1046

jackiemunroe:
- add tests for direct stripe integration for subscription routes
- refactor changes to make testing easier

issue #3809
issue #4073


packages/fxa-auth-server/test/mocks.js
```diff
@@ -779,3 +780,13 @@ function mockVerificationReminders(data = {}) {
     ),
   };
 }
+
+function mockStripeHelper(methods) {
+  //const StripeHelper = require('../lib/payments/stripe').StripeHelper;
+  //const log = module.exports.mockLog();
+  //const config = {
+  //  subhub: {plansCacheTtlSeconds: 1000}
+  //};
```
LZoog: Missed comments?

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -157,6 +159,44 @@ class StripeHelper {
...
+   *
+   * @param {string} uid
+   * @param {string} email
+   * @param {string} displayName
+   * @param {string} paymentToken
+   *
+   * @returns {Promise<Customer>}
+   */
+  async createCustomer(uid, email, displayName, paymentToken) {
+    const customer = await this.stripe.customers.create({
```
lmorchard: Tiny nit: not doing any exception handling here, so could just return the result of the `create` call directly without an `await`. Same with `updateCustomerPaymentMethod` below. (Doesn't hurt anything to leave as-is, though.)

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -203,20 +139,132 @@ class DirectStripeRoutes {
...
+        paymentToken
+      );
+    } catch (err) {
+      if (err.type === 'StripeCardError') {
+        throw error.rejectedSubscriptionPaymentToken(err.message, err);
+      }
+      throw err;
+    }
+
+    const subscription = await this.stripeHelper.createSubscription(
```
lmorchard: Might be good to move the `createSubscription` call into the try/catch along with `createCustomer`? I still have a strong suspicion that this `StripeCardError` will come up from either of these spots

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -203,20 +139,132 @@ class DirectStripeRoutes {
...
+      const invoice = /** @type {Invoice} */ (subscription.latest_invoice);
+      if (invoice.status === 'open') {
+        await this.handleOpenInvoice(invoice);
+        return subscription;
+      } else if (invoice.status === 'paid') {
+        throw error.subscriptionAlreadyExists();
+      }
+    }
+
+    subscription = await this.stripeHelper.createSubscription(
```
lmorchard: Same thing here for catching StripeCardErrors. Then again, @bbangert had mentioned wrapping the entire route in an error handler, so that might be a smarter approach.

packages/fxa-auth-server/test/mocks.js
jackiemunroe: Oh, all of those are me working on tests mocking. This is a piecemeal pr, trying to break things down into smaller hunks 😓

packages/fxa-auth-server/lib/payments/stripe.js
jackiemunroe: I was thinking about doing that. The reason it is as it currently appears: I wanted to be able to set a breakpoint at the return so I could check the value.

packages/fxa-auth-server/lib/routes/subscriptions.js
jackiemunroe: I was in debate over this very thing: to put it all in the `try/catch` or not

PR #4100, opened 2020-02-11
fix(show-password): Fix a11y issue in Android Talkback and fix general mobile bug

### LINKED ISSUE #3991: [a11y] Show password is read, but cannot be accessed when creating an account
## Steps to reproduce
1. Turn Talkback or Voice Assistant on.
2. Go to Settings - Turn on Sync and create a new account.
3. Fill in a valid email account, a valid password and then repeat the password.
4. Tap on the "Show password" icon.

### Expected behavior
The "Show password" icon is read and double-tapping on it is accessed and the user can see the password filled.

### Actual behavior
The "Show password" icon is read even if it isn't highlighted, but double-tapping on it does nothing. The password is not displayed, the user cannot see it.

### Device information
* Android device: Samsung Galaxy Tab S3 (Android 8)
* Fenix version: Beta 3.1.0-beta 3

Untitled

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-988)
┆Issue Number: FXA-988

LZoog:
fixes #3991 

TLDR; Android Talkback requires visible content in input labels for them to be interacted with. This PR moves the background-image into the label (via `img`) and fixes a mobile bug double toggling (separate commits).

---

This was a doozy to debug and find the right solution for - 

For Android TalkBack (in Firefox, not Chrome) to allow interaction with input labels, the label apparently must have content. In all other scenarios that I tested, simply having the `text-indent` moves the text “Show Password” off screen left the label in place, but when TalkBack is enabled, the label is moved off-screen as well.

Instead of a text indent, I tried various things such as making the text transparent, adding non-breaking whitespaces and padding, visibility and other font-size tricks, and none of these hacks worked.

Our options are to 1) expand the width of the input itself (currently `1px` wide to hide visually), lying it perfectly underneath the label at all breakpoints/ltr vs rtl languages + adding another listener/function to handle this specific case, or 2) ensure there is visible content in the label, matching the height/width of the toggle area.

I opted for the latter. I modified the SVG to take up the entire width/height, finessing the `viewBox` until it matched the previous “eye” size in mobile, and have `5px` padding around the image at the larger breakpoint to make the “eye” smaller, matching the style we had previously. Also added `pointer-events: none` so the user clicks through the image onto the label.

Tested across the board in desktop FF, Chrome, iOS, Android with/without Talkback. Mac VO appears to behave the same way but there doesn’t seem to be an option to test VO in Xcode. I’ll ask QA to verify this behaves as expected once it’s in.

*Also*, I removed the `touchstart` event. I noticed that when you attempt to quickly toggle in at least iOS, the “show password” toggle wasn’t working. You can see this on prod right now. This appears to be because `click` is fired after `touchstart` so it was quickly toggling on and then off.

I tried to match the eyes as best as I could to current:

image


packages/fxa-content-server/app/tests/spec/views/mixins/password-mixin.js
```diff
@@ -126,17 +126,6 @@ describe('views/mixins/password-mixin', function() {
         assert.equal(view.$('#password').attr('type'), 'password');
       });
 
-      it('works with touch events', () => {
```
LZoog: Mouse events are simulated in mobile devices but `touchstart` is fired first: https://developer.mozilla.org/en-US/docs/Web/API/Touch_events

This was causing the toggle to fire twice, I'm only see the password reveal for about 1/2 a second on prod. I could have also called `stopPropogation` if it's a `touchstart` event, that just feels unnecessary. `touchstart` should be used if there's functionality (like swiping) that can't be done in desktop.

PR #4108, opened 2020-02-12
fix(auth-server): handle stripe webhooks correctly

### LINKED ISSUE #4040: Stripe webhook needs to determine when subscription state has changed
Currently the stripe webhook invalidates our cache. However, we need to notify the event-broker via a service notification when a users subscription status changes from active to inactive and vice versa. To do this our webhook will need to examine the event to see what is changing so it can take the appropriate action.

We may need to handle the subscription events separately from customer.updated to accomplish this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1027)
┆Issue Number: FXA-1027

bbangert:
Because:

* `customer.updated` does not include subscription information as we
  previously assumed.
* We were not sending out the SQS subscription state change event.

This commit:

* Handles `customer.subscription.(created/updated/deleted)` so that we
  properly emit subscription state changes and refresh the cached
  customer at the appropriate time.

Closes #4040

Co-authored-by: Lauren Zugai <lauren@zugai.com>

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -16,6 +16,7 @@ const SUBSCRIPTIONS_MANAGEMENT_SCOPE =
 
 /** @typedef {import('hapi').Request} Request */
 /** @typedef {import('stripe').Stripe.Customer} Customer */
+/** @typedef {import('stripe').Stripe.Event} Event */
 /** @typedef {import('stripe').Stripe.Subscription} Subscription */
 /** @typedef {import('stripe').Stripe.Invoice} Invoice */
 /** @typedef {import('stripe').Stripe.PaymentIntent} PaymentIntent */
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -323,7 +323,7 @@ class StripeHelper {
    */
   async refreshCachedCustomer(uid, email) {
     try {
-      return await this.customer(uid, email, true);
```
lmorchard: Removing the `await` here means that the try/catch never catches an exception. The unresolved failure bubbles to the caller. If that's intentional, we should remove the try/catch

bbangert: Woops! Fixed.

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -471,32 +472,163 @@ class DirectStripeRoutes {
...
+    }
+    await this.stripeHelper.refreshCachedCustomer(uid, email);
+    await this.profile.deleteCache(uid);
+    await this.sendSubscriptionStatusToSqs(
+      request,
+      uid,
+      event,
+      { id: sub.id, productId: /** @type {string} */ (sub.plan.product) },
+      true
+    );
```
lmorchard: Nit: Might be worth extracting this chunk as a method shared between the handlers? Seems like the same code for each, except for the logic in `handleSubscriptionUpdatedEvent` which could be an early return like here in this handler

bbangert: Sort of, the order is different so that some work can be avoided in the updated event handler, which means only the deleted also would benefit from that. But that might still be cleaner.

PR #4114, opened 2020-02-12
feat(auth-server): add stripe metric timings

### LINKED ISSUE #3862: Metric timings for stripe API calls
We currently have metrics on subhub timings, our direct stripe API calls should have similar metric timings so we can break-down latency issues on our requests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-908)
┆Issue Number: FXA-908

bbangert:
Because:

* We want to know metrics on how long stripe calls take.

This commit:

* Add's a response handler to stripe to send out metric timings.

Closes #3862

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -96,8 +96,9 @@ class StripeHelper {
    *
    * @param {object} log
    * @param {object} config
+   * @param {import('hot-shots').StatsD | undefined} statsd
    */
-  constructor(log, config) {
+  constructor(log, config, statsd) {
     this.log = log;
     this.cacheTtlSeconds =
       config.subhub.plansCacheTtlSeconds ||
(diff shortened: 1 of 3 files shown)


bbangert:
It's less than ideal to have no tests for this small fragment, and I did try constructing one. Except that it results in such an artificial situation that its not really testing that it works.

I did run this locally to verify:
```
fxa-auth-server.stripe_request:193|ms
fxa-auth-server.stripe_request:165|ms
fxa-auth-server.stripe_request:95|ms
```

Stripe types also help ensure that the response property access is valid.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -119,9 +120,15 @@ class StripeHelper {
     this.stripe = new stripe(config.subscriptions.stripeApiKey, {
       apiVersion: '2019-12-03',
       maxNetworkRetries: 3,
-      typescript: undefined,
+      typescript: true,
```
vbudhram: Is this change related?

bbangert: Oops, removing.

bbangert:
Updated the commit for the right issue.

PR #4141, opened 2020-02-13
fix(auth-server): use cached customer only

bbangert:
Because:

* Routes that hit Stripe should be minimized to when we know the user
  is a customer.

This commit:

* Uses only the Redis copy of stripe data in more routes that we could
  accidentally hit in settings pages for all FxA users.


bbangert:
They're still pending for this portion.

PR #4158, opened 2020-02-14
fix(auth-server): don't delete the stripe customer

bbangert:
Because:

* SubHub will handle deleting the customer and cleanup.

This commit:

* Updates the metadata attribute subhub is watching for.


bbangert:
This route has tests that are coming in @jackiemunroe 's upcoming test PR's.

bbangert:
Ack, my bad, that would not be in those tests as this is in the account route. I'll file another issue to add a test to this to ensure I don't forget it.

PR #4168, opened 2020-02-15
test(fxa stripeHelper): add tests

jackiemunroe:
increased code coverage for payments/stripe.js
added Arch 2.0 code coverage and test runner to launch.json

issue #4073

### DIFF
--- a/packages/fxa-auth-server/.vscode/launch.json
+++ b/packages/fxa-auth-server/.vscode/launch.json
@@ -82,6 +82,26 @@
       },
       "preLaunchTask": "Stop PM2 Auth Server",
       "postDebugTask": "Start PM2 Auth Server"
-    }
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Arch 2.0 with Coverage",
+      "program": "${workspaceFolder}/node_modules/nyc/bin/nyc.js",
+      "args": [
+        "--reporter=lcov",
+        "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+        "--timeout",
+        "999999",
+        "--colors",
+        "${workspaceFolder}/test/local/payments/stripe.js",
+        "${workspaceFolder}/test/local/routes/subscriptions.js"
+      ],
+      "env": { "NODE_ENV": "dev" },
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "runtimeVersion": "12.14.0",
+      "protocol": "inspector"
+    },
   ]
 }
--- a/packages/fxa-auth-server/test/local/payments/stripe.js
+++ b/packages/fxa-auth-server/test/local/payments/stripe.js
@@ -8,6 +8,7 @@ const sinon = require('sinon');
 const { assert } = require('chai');
 const { mockLog } = require('../../mocks');
 const error = require('../../../lib/error');
+const stripeError = require('stripe').Stripe.errors;
 
 let mockRedis;
 const proxyquire = require('proxyquire').noPreserveCache();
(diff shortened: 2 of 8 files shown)


packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -243,6 +389,57 @@ describe('StripeHelper', () => {
...
       );
     });
+
+    it('returns void if no customers are found', async () => {
+      const listResponse = {
+        autoPagingToArray: sinon.fake.resolves([]),
+      };
+      sandbox.stub(stripeHelper.stripe.customers, 'list').returns(listResponse);
+
+      assert.isEmpty(stripeHelper.fetchCustomer('user123', 'test@example.com'));
```
bbangert: `await` the `fetchCustomer` result?

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -375,6 +584,60 @@ describe('StripeHelper', () => {
...
+        assert.isNotEmpty(result);
+        assert(
+          log.error.calledOnceWith(
+            'subhub.cachedResult.getCachedResponse.failed'
+          )
+        );
+      });
+
+      it('logs error with undefined refreshFunction (cacheOnly = true)', async () => {
+        mockRedis.get.restore();
```
bbangert: These could go into a `beforeEach` for the parent `describe`.

packages/fxa-auth-server/.vscode/launch.json
```diff
@@ -82,6 +82,26 @@
       },
       "preLaunchTask": "Stop PM2 Auth Server",
       "postDebugTask": "Start PM2 Auth Server"
-    }
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Arch 2.0 with Coverage",
```
bbangert: Nice!

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -123,53 +127,123 @@ describe('StripeHelper', () => {
...
+      it('thows an unknownSubscriptionPlan error', async () => {
+        return stripeHelper
+          .verifyPlanUpgradeForSubscription('plan_F4G9jB3x5i6Dpj', 'plan_bad')
+          .then(
+            () => Promise.reject(new Error('Method expected to reject')),
+            err => {
+              assert.equal(err.errno, error.ERRNO.UNKNOWN_SUBSCRIPTION_PLAN);
+            }
+          );
+      });
```
bbangert: Was this approach easier to reason about then doing something like:
```javascript
try {
    await stripeHelper.verifyPlanUpgradeForSubscription('plan_F4G9jB3x5i6Dpj', 'plan_bad');
    assert.fail('Method expected to reject');
} catch (err) {
    assert.equal(err.errno, error.ERRNO.UNKNOWN_SUBSCRIPTION_PLAN);
}
```

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -243,6 +389,57 @@ describe('StripeHelper', () => {
...
+
+        const result = await stripeHelper.fetchCustomer(
+          'testuid',
+          customer.email
+        );
+
+        assert.isFalse(result.subscriptions.has_more);
+        assert(
+          result.subscriptions.data.length === 2,
+          'both subscriptions should be loaded'
```
bbangert: `assert.strictEqual` is usually more useful in these cases since it'll output the actual value, while the `assert` message is less useful as it doesn't output the actual value.

packages/fxa-auth-server/test/local/payments/stripe.js
jackiemunroe: I just saw it done both ways in the code base, and it just made more sense to me because the structure clearly shows both of the possible outcomes.

jackiemunroe:
@bbangert I made updates based on your comments, would you mind reviewing when you get a chance? :)

PR #4172, opened 2020-02-15
Update .nvmrc to use Node v12

jodyheavener:
Smol change.

If you try to `npm install` with a v10 of Node it can't proceed, per `_scripts/check.sh`:

```bash
➜  fxa git:(master) npm install

> fxa@2.0.0 postinstall /Documents/Repositories/fxa
> _scripts/install_all.sh

+ '[' '' '!=' true ']'
+ _scripts/install_packages.sh
+ _scripts/check.sh
++ node -v
++ cut -f 1 -d .
++ cut -f 2 -d v
+ node_version=10
+ '[' 10 -ne 12 ']'
+ echo 'install node 12 to continue installation'
install node 12 to continue installation
+ echo http://nodejs.org/
http://nodejs.org/
+ exit 1
```


LZoog:
Thanks @jodyheavener!

PR #4180, opened 2020-02-18
fix(metrics): add Amplitude events for /pair on content server

### LINKED ISSUE #4103: Add a `pairing` option for the `connect_device_flow` amplitude user property
Right now when a user visits firefox.com/pair they see this view:
image

Which triggers the amplitude event `fxa_connect_device - view` with event property `connect_device_flow` = `sms`. The event property value is incorrect, we should implement a new value `pairing` so that (1) we can know how many users visit the pairing page and (2) we stop mixing the pairing metrics with the SMS CAD metrics. 

CC @davismtl this is now preventing us from measuring pairing properly.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1069)
┆Issue Number: FXA-1069

chenba:
This patch adds a couple missing Amplitude events to /pair on the
content server.  The events are 'fxa_connect_device - view' and
'fxa_connect_device - submit' with the event property
'connect_device_flow' set to 'pairing'.

Fixes #4103 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/pair/index.js
+++ b/packages/fxa-content-server/app/scripts/views/pair/index.js
@@ -4,6 +4,7 @@
 
 import FormView from '../form';
 import Cocktail from 'cocktail';
+import FlowEventsMixin from '../mixins/flow-events-mixin';
 import Template from '../../templates/pair/index.mustache';
 import UserAgentMixin from '../../lib/user-agent-mixin';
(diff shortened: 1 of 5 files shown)


chenba:
> 
> 
> r+, nit: can we also have a "success" event for Amplitude? or would submit be enough?

I don't see a "success" for the 'sms' case in Amplitude.  If and when @irrationalagent or @davismtl want something different, we can make it so then.

PR #4188, opened 2020-02-19
fix(webchannel): ignore native Gecko WebChannel errors for the OAuth WebChannel

### LINKED ISSUE #3116: Include `context` param in `fxa_status` webchannel message
When the browser wants to link out to FxA management pages on the web, it currently includes `?service=sync&context=fx_desktop_v3` query parameters, like so:

  https://accounts.firefox.com/settings?service=sync&context=fx_desktop_v3

The presence of these query parameters causes accounts.firefox.com to send an `fxa_status` webchannel message to the browser to interrogate its login state, and the browser can choose whether to reply with the credentials of the signed-in user.

Now that it's possible to sign in to the browser without enabling sync, it doesn't really make sense to include `service=sync` in those links, and its presence there risks muddying FxA server-side metrics. We'd like to change these links to just be:

  https://accounts.firefox.com/settings?context=fx_desktop_v3

For this to work, we'll need FxA to report the value of `context` in its `fxa_status` handshake message to the browser, so that the browser can decide whether or not to respond to the message without having to check for `service=sync`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-489)
┆Issue Number: FXA-489

vladikoff:
@rfk r?

Fixes #3116

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/lib/channels/web.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/channels/web.js
@@ -204,6 +204,25 @@ describe('lib/channels/web', () => {
       });
     });
 
+    describe('OAuth WebChannel', () => {
+      it('does not reject outstanding requests for the OAuth WebChannel', () => {
+        channel = new WebChannel('MyChannel');
+        channel.initialize({
+          window: windowMock,
+          isOAuthWebChannel: true,
+        });
+        sandbox
+          .stub(channel, 'rejectAllOutstandingRequests')
+          .callsFake(() => {});
+
+        const webChannelError = {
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/tests/spec/lib/channels/web.js
```diff
@@ -204,6 +204,25 @@ describe('lib/channels/web', () => {
       });
     });
 
+    describe('OAuth WebChannel', () => {
+      it('does not rejects outstanding requests for the OAuth WebChannel', () => {
```
rfk: nit: "rejects" -> "reject"

PR #4210, opened 2020-02-20
test(auth-server): add missing direct stripe test

### LINKED ISSUE #4160: Add direct stripe tests for account route in auth-server
Ensure that POST /account/destroy route verifies the direct stripe integration, which should update the customer's metadata.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1113)
┆Issue Number: FXA-1113

bbangert:
Because:

* My prior commit didn't update tests for the direct stripe account
  delete route.

This commit:

* Add's tests for the direct stripe account delete route.

Closes #4160


bbangert:
Note, I updated this test to use direct stripe only and skip'd the two prior subhub related ones since we're now on architecture 2.0. The rest of the subhub bits can be removed in a later PR.

PR #4211, opened 2020-02-20
task(payments-server): Add subscription details component

### LINKED ISSUE #4142: create subscription details component
This will be used throughout the subscription flow

* Need to make sure it works for confirmation and payment pages
* add tests

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1101)
┆Issue Number: FXA-1101
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12500/details-box-component.png">details-box-component.png</a>

meandavejustice:
- fixes #4142

TODO:
- load product image
- l10n
- add product descriptions (still a little discussion going on in slack about this)


packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <li>something abt the product</li>
+            <li>something abt the product</li>
+            {/* unordered list generated
+            over a loop with each description details
+          */}
+          </ul>
+        </section>
+      ) : null}
+      {showExpandButton ? (
+        <section className="footer">
```
LZoog: If we're going to use `<header>`, we should use `<footer>`.

Normally though, there's content inbetween `<header>` and `<footer>` and there isn't any on initial load... Looking at the wires, I agree (`.plan-details-component`) should be its own `section`. We could remove the `<header>` grouping ([this article](http://html5doctor.com/avoiding-common-html5-mistakes/) goes into details on section/header overuse) _but_ I could see how "show/hide details" could be part of the `<footer>` for that section. I feel more strongly about `.plan-details-list` just being a regular `div` though.

Looking at the wires I think `Set up your subscription` can remain the `h2` for the page, but man it's a little awkward moving this entire section out and to the right in desktop (and hiding the show/hide buttons). The product name (`.plan-details-product`) should probably be an h3 and "Product details" an h4 (which is how you have it), and since this is self-contained in its own `section`, "Billing information" can remain an h3.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+              <p
+                className="btn down-arrow"
+                onClick={() => setDetailsState(false)}
+              >
+                Show details
+              </p>
+            </Localized>
+          ) : (
+            <Localized id="plan-details-hide-button" span={up}>
+              <p className="btn up-arrow" onClick={() => setDetailsState(true)}>
```
LZoog: These hide/shows should be `button`s since they trigger some sort of functionality. Also, the `aria-expanded` value should be on these buttons, not on the content.

packages/fxa-payments-server/src/images/products/fpn.svg
```diff
@@ -0,0 +1,37 @@
...
+<linearGradient id="paint3_linear" x1="6.87718" y1="22.8427" x2="24.7785" y2="4.94144" gradientUnits="userSpaceOnUse">
+<stop offset="0.371" stop-color="#FFF36E" stop-opacity="0"/>
+<stop offset="0.426" stop-color="#FFF36E" stop-opacity="0.039"/>
+<stop offset="0.519" stop-color="#FFF36E" stop-opacity="0.142"/>
+<stop offset="0.639" stop-color="#FFF36E" stop-opacity="0.312"/>
+<stop offset="0.779" stop-color="#FFF36E" stop-opacity="0.543"/>
+<stop offset="0.917" stop-color="#FFF36E" stop-opacity="0.8"/>
+</linearGradient>
+</defs>
+</svg>
```
LZoog: We can run this through `svgo` to clean it up a bit, but not a big deal.

packages/fxa-payments-server/src/components/PlanDetails/index.scss
```diff
@@ -0,0 +1,116 @@
...
+      margin-bottom: 16px;
+
+      ::marker {
+        font-size: 12px;
+      }
+    }
+  }
+
+  .btn {
+    background: none !important;
```
LZoog: I know we do this in the auth-server email CSS, but we should _really_ avoid using `!important`. If we need to override this later we'll have to override selector specificity _and_ add `!important`. Can we achieve the same thing with selector specificity?

packages/fxa-payments-server/src/components/PlanDetails/index.scss
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  header img {
+    display: block;
+    height: 64px;
+    width: 64px;
+  }
+
+  .plan-details-list {
+    border-bottom: 1px solid#e7e7e7;
```
LZoog: Missing a space here as well as line 37.

Can we set some of these colors to variables?

packages/fxa-payments-server/src/components/PlanDetails/index.scss
```diff
@@ -0,0 +1,116 @@
...
+
+  .plan-details-heading-wrap {
+    display: flex;
+    flex-direction: column;
+    padding-left: 10px;
+    text-align: left;
+
+    .plan-details-product {
+      /* Grey/Grey 50 */
+      color: rgba(12, 12, 13, 0.8);
```
LZoog: Do we need to change the opacity of some of these text elements? Can we just use a hex color of a lighter grey (easier to picture the color & probably reuse with variables)?

packages/fxa-payments-server/src/components/PlanDetails/index.scss
vbudhram: We do it in the content-server as well and if it has to be there, a comment explaining why would be good.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
meandavejustice: I went with `.footer` here because of the global css on the page for the `<footer>` element was making it hard to reset. I'm happy to remove header/footer references

packages/fxa-payments-server/src/components/PlanDetails/index.scss
meandavejustice: Yeah I just copied from content-server. From what I understand we only need to use `!important` here when we are resetting the default `<button>` styles. I can try to find a workaround, if I can't find one quick I'll just leave a comment

packages/fxa-payments-server/src/components/PlanDetails/index.scss
meandavejustice: We don't seem to be using variables anywhere on payment server. We definitely should be though. I feel like Danny had a bug for this in the maintenance board

packages/fxa-payments-server/src/components/PlanDetails/index.scss
meandavejustice: These colors were all copied from figma, I can check with anthony

packages/fxa-payments-server/src/components/PlanDetails/index.scss
LZoog: 👍 We're importing the variables from the content server in at least a couple places, but, as I've said before I'm going to revamp our SCSS with the settings redesign and reviewing/setting variables is part of that. I'd already planned on going through payments as part of that to standardize us across the board so I'd say let's not spend too much time on that but import them where it makes sense.

meandavejustice:
closing for https://github.com/mozilla/fxa/pull/4259

PR #4221, opened 2020-02-20
Release 1.159.0

dannycoates:


AUTHORS
```diff
@@ -103,8 +103,8 @@ Nicholas Mandel
 Nick Alexander
 Nick Chapman
 Omkar Yadav
-petercpg@mail.moztw.org
```
vladikoff: do these happen due to different version of git?

PR #4239, opened 2020-02-21
test(auth-server): Coverage for DirectStripeRoutes updateSubscription

### LINKED ISSUE #4229: Code coverage for Direct Stripe Route - Update Subscription
fxa-auth-server/lib/routes/subscriptions.js

Function: DirectStripeRoutes.updateSubscription

Only one line is missing coverage, and that is for the `if (!subscription)` case. Add to the existing cases for this method to handle this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1175)
┆Issue Number: FXA-1175

jackiemunroe:
- added coverage for the specified function
- updated vscode script for Arch 2.0 coverage

fixes #4229

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/subscriptions.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions.js
@@ -1754,7 +1754,7 @@ describe('DirectStripeRoutes', () => {
     };
 
     log = mocks.mockLog();
-
+    customs = mocks.mockCustoms();
     profile = mocks.mockProfile({
       deleteCache: sinon.spy(async uid => ({})),
     });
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/.vscode/launch.json
```diff
@@ -95,7 +95,8 @@
         "999999",
         "--colors",
         "${workspaceFolder}/test/local/payments/stripe.js",
-        "${workspaceFolder}/test/local/routes/subscriptions.js"
+        "${workspaceFolder}/test/local/routes/subscriptions.js",
+        "--exit"
```
bbangert: I assume at some point this task will be removed as its mainly useful in the specific case someone is adding test coverage just for this one portion of the auth-server vs. its entirety? Or should we have one task per logical unit of this? Or would it make sense to duplicate the task that runs the tests for the current file so that it can run coverage on the current file as well?

jackiemunroe: I was thinking about updating it to run the current file.

jackiemunroe:
@bbangert commit message updated :)

bbangert:
Thanks! Though.... looks like this was more sensitive to the ordering than I realized. It needs a rebase.

PR #4248, opened 2020-02-21
test(auth-server): Coverage for Delete Subscription

### LINKED ISSUE #4225: Code coverage for Direct Stripe Route - Delete Subscription
fxa-auth-server/lib/routes/subscriptions.js

Function: DirectStripeRoutes.deleteSubscription

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1172)
┆Issue Number: FXA-1172

jackiemunroe:
- refactored deleteSubscription, so as to not call stripe library directly
- added coverage for deleteSubscription in subscriptions.js DirectStripeRoutes
- added coverage for new cancelSubscriptionForCustomer in stripe.js StripeHelper

fixes #4225

### DIFF
--- a/packages/fxa-auth-server/test/local/payments/stripe.js
+++ b/packages/fxa-auth-server/test/local/payments/stripe.js
@@ -266,6 +266,54 @@ describe('StripeHelper', () => {
     });
   });
 
+  describe('cancelSubscriptionForCustomer', () => {
+    let stripeSubscriptionsUpdateStub;
+
+    beforeEach(() => {
+      stripeSubscriptionsUpdateStub = sandbox
+        .stub(stripeHelper.stripe.subscriptions, 'update')
+        .resolves();
+    });
+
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -266,6 +266,54 @@ describe('StripeHelper', () => {
...
+      it('throws an error', async () => {
+        sandbox.stub(stripeHelper, 'subscriptionForCustomer').resolves();
+        return stripeHelper
+          .cancelSubscriptionForCustomer(
+            '123',
+            'test@example.com',
+            subscription2.id
+          )
+          .then(
+            () => Promise.reject(new Error('Method expected to reject')),
```
lmorchard: You *could* do a try/catch here instead of `.then()`, but not a huge deal

PR #4249, opened 2020-02-21
task(payment-server): add new header component for new flow

### LINKED ISSUE #4145: Update header for subscription payment redesign
This is used at all stages of subscription payment

Looks the same at all sizes, just logo and avatar

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1102)
┆Issue Number: FXA-1102
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12501/header-component.png">header-component.png</a>

meandavejustice:
- fixes #4145


packages/fxa-payments-server/src/components/Header/index.scss
```diff
@@ -0,0 +1,49 @@
+@import '../../../../fxa-content-server/app/styles/breakpoints';
+
+#fxa-settings-header {
```
LZoog: Do you have a SCSS linter installed? I have `scss-lint` installed from the VSCode extensions list (perhaps there's a better one), most of these squiggles are due to indentation but here's what I'm seeing:

image

Another thing I'm planning to address in the content server with the settings redesign is fixing up all of our warnings. One major player in this is styling with IDs, it's best practice to style with classes or `.class-name [element]` and the SCSS linter is complaining about this. However... we could end up turning this SCSS rule off if the team strongly prefers to style with IDs, I haven't asked anyone yet what their preference is. So it'd be better practice to turn this into a class, but you could also make the argument that we haven't set a hard rule for this yet and I'm nitpicking too hard here when I'm going to revisit a lot of our CSS in a few months, heh.

At the very least, we can pull `#fxa-manage-account` out of `#fxa-settings-header` nesting. When styling with an ID you should be able to do that at the top-level since nothing else will be named `#fxa-settings-header` on the page to cause conflicting styles and thus this level of specificity.

packages/fxa-payments-server/src/components/Header/index.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import MockApp from '../../../.storybook/components/MockApp';
+import Header from './index';
+
+const userProfile = {
+  avatar: './avatar.svg',
```
LZoog: This path seems to be incorrect, I'm seeing the alt text in storybook.

packages/fxa-payments-server/src/components/Header/index.scss
meandavejustice: I also don't like styling with ids, I copied this style from another location in the repo. I'll update and fix linter warnings. I should be using the same linter configs as you, I'll see if it got disabled for some reason cus I haven't been seeing sass lint warnings for a while

packages/fxa-payments-server/src/components/Header/index.scss
```diff
@@ -0,0 +1,55 @@
...
+  padding: 0 32px;
+  position: fixed;
+  top: 0;
+  white-space: nowrap;
+  width: 100%;
+  z-index: 2;
+
+  @include max-width('mobileLandscape') {
+    height: 64px;
+  }
```
LZoog: Can we set the default height to `64px` and use `@include min-width('mobileLandscape') { height: 81px` for extra mobile friendliness?

PR #4250, opened 2020-02-21
fix(auth-server): stripe refresh on primary change

### LINKED ISSUE #4247: Primary email address change doesn't refresh Redis customer copy
Changing the primary email address properly updates Stripe, but the Redis copy is not refreshed using the new email address. This causes the Redis lookup to fail and we fail to see the users subscription.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1189)
┆Issue Number: FXA-1189

bbangert:
Because:

* We were failing to refresh our cached customer record for Stripe
  subscriptions on primary email change.
* Our cached copy is keyed by email + uid, so the email change meant
  we lost our local reference.

This commit:

* Refreshes the Redis record for a customer to the new email address.

Fixes #4247

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -82,6 +82,7 @@ module.exports = (
   verificationReminders,
   signupUtils,
   zendeskClient,
+  /** @type import('../payments/stripe').StripeHelper */
   stripeHelper
 ) => {
   const REMINDER_PATTERN = new RegExp(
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -864,12 +865,25 @@ module.exports = (
...
+            // Wait here to update stripe and our local cache to avoid loss of
+            // valid subscription status.
+            try {
+              await updateStripeEmail(
+                stripeHelper,
+                uid,
+                primaryEmail,
+                secondaryEmail.normalizedEmail
+              );
+              await stripeHelper.refreshCachedCustomer(
```
lmorchard: Might be better to just include this as part of the `updateStripeEmail` function? Is there any time we'd use `updateStripeEmail` *without* needing to `refreshCachedCustomer`?

bbangert: Probably not.

bbangert: Should note that I kept them separate mainly cause they do different things.

PR #4251, opened 2020-02-21
test(auth-server): Coverage for reactivateSubscription

### LINKED ISSUE #4230: Code coverage for Direct Stripe Route - Reactivate Subscription
fxa-auth-server/lib/routes/subscriptions.js

Function: DirectStripeRoutes.reactivateSubscription

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1174)
┆Issue Number: FXA-1174

jackiemunroe:
- refactored reactivateSubscription, so as to not call stripe library directly
- added coverage for reactivateSubscription in subscriptions.js DirectStripeRoutes
- added coverage for new reactivateSubscriptionForCustomer in stripe.js StripeHelper
- added placeholders for upcoming tests to reduce likelihood of merge conflicts

fixes #4230

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -491,7 +491,7 @@ class StripeHelper {
   }
 
   /**
-   * Cancel a givel subscription for a customer
+   * Cancel a given subscription for a customer
    * If the subscription does not belong to the customer, throw an error
    *
    * @param {string} uid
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -513,6 +513,50 @@ class StripeHelper {
     });
   }
 
+  /**
+   * Reactivate a givel subscription for a customer
```
bbangert: `givel`?

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -513,6 +513,50 @@ class StripeHelper {
...
+    }
+    const subscription = await this.stripe.subscriptions.update(
+      subscriptionId,
+      {
+        cancel_at_period_end: false,
+      }
+    );
+
+    if (!['active', 'trialing'].includes(subscription.status)) {
+      const err = new Error(
```
bbangert: What conditions cause this? Is a user likely to keep clicking to try and re-activate it and wonder what's wrong?

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -2095,6 +2019,113 @@ describe('DirectStripeRoutes', () => {
...
+        }
+      );
+    });
+  });
+
+  describe('listActive', () => {});
+
+  describe('getCustomer', () => {});
+
+  describe('sendSubscriptionStatusToSqs', () => {});
```
bbangert: These are coming later?

jackiemunroe: I added placeholders for upcoming tests to reduce likelihood of merge conflicts

packages/fxa-auth-server/lib/payments/stripe.js
jackiemunroe: This was in the existing code. Actually, this check should happen before the update call to verify that the subscription should be allowed to be updated. Doing the check after-the-fact only shows that the subscription wasn't eligible for a state change.

packages/fxa-auth-server/lib/payments/stripe.js
jackiemunroe: Typo.

packages/fxa-auth-server/lib/payments/stripe.js
jackiemunroe: I've updated to reflect this opinion. Let me know your thoughts.

bbangert: Did you push your update? The typo above is still present as well.

jackiemunroe: I did! It looks like the typo occurred in more than one place 😓 I will fix the place that I missed

jackiemunroe: Ready to go @bbangert

jackiemunroe: The change is that now it verifies that the subscription is in an eligible state prior to re-activation, rather than checking afterwards.

PR #4192, opened 2020-02-19
feat(recovery): Add confirm recovery key UX

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/3854

This adds the final UX parts for confirming your recovery key screens and instruments the views. However, the two main components missing in this PR are

- Updated email templates to enable the experiment
- Emitting a metric when user consumes a recovery code https://github.com/mozilla/fxa/issues/4217

Since it took a little bit longer to write the tests, all parts of this will not get landed in this train.

PR #4258, opened 2020-02-24
feat(recovery): Add confirm recovery key js client calls

vbudhram:
Extracted from https://github.com/mozilla/fxa/pull/4192 so that we can land the js client functions first and get the tests passing.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/recovery-key.js
+++ b/packages/fxa-auth-server/lib/routes/recovery-key.js
@@ -37,7 +37,26 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
         const { uid } = sessionToken;
         const { recoveryKeyId, recoveryData, enabled } = request.payload;
 
-        await db.createRecoveryKey(uid, recoveryKeyId, recoveryData, enabled);
+        // Users that already have an enabled recovery key can not
+        // create a second recovery key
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/routes/recovery-key.js
```diff
@@ -37,7 +37,26 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
         const { uid } = sessionToken;
         const { recoveryKeyId, recoveryData, enabled } = request.payload;
 
-        await db.createRecoveryKey(uid, recoveryKeyId, recoveryData, enabled);
+        // Users that already have an enabled recovery key can not
+        // create a second recovery key
+        try {
+          await db.createRecoveryKey(uid, recoveryKeyId, recoveryData, enabled);
+        } catch (err) {
+          if (err.errno !== errors.ERRNO.RECOVERY_KEY_EXISTS) {
```
6a68: This use of try-catch for flow control seems like an antipattern. Is there an alternative we could use instead? For instance, if `db.createRecoveryKey` was changed to `db.upsertRecoveryKey`, where the step of deleting any half-baked key could be abstracted into the database, the code would be simpler: "if recovery key exists, bail, else, upsert recovery key".

PR #4266, opened 2020-02-25
feat(oauth): store access tokens in redis

dannycoates:
The intent here is to store oauth access tokens in redis instead of mysql in order to move load off of mysql in anticipation of increased load from Fenix. We also have a legacy use case that [tokens for Pocket are kept indefinitely](https://bugzilla.mozilla.org/show_bug.cgi?id=1547902) that we need to keep supporting. Instead of putting new Pocket tokens in redis and having to support them in two places I opted to keep them going into mysql while vanilla oauth access tokens only go into redis where they can expire with a set TTL.

### DIFF
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -206,7 +206,9 @@
     },
     "clientIdToServiceNames": {
       "dcdb5ae7add825d2": "123done",
-      "98e6508e88680e1a": "fxa-settings"
+      "98e6508e88680e1a": "fxa-settings",
+      "7377719276ad44ee": "pocket-mobile",
+      "749818d3f2e7857f": "pocket-web"
     },
     "clients": [
       {
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -890,6 +890,9 @@ const conf = convict({
       },
     },
     clientIdToServiceNames: {
+      // This is used by oauth/db/index.js to identify pocket client ids so that it
+      // can store them separately in mysql.
+      // It's also used for amplitude stats
       doc:
         'Mappings from client id to service name: { "id1": "name-1", "id2": "name-2" }',
       default: {},
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -337,6 +337,24 @@ const conf = convict({
       format: 'port',
       doc: 'Port for Redis server',
     },
+    accessTokens: {
+      host: {
+        default: '127.0.0.1',
+        env: 'ACCESS_TOKEN_REDIS_HOST',
+        format: String,
+      },
+      port: {
+        default: 6379,
+        env: 'ACCESS_TOKEN_REDIS_PORT',
+        format: 'port',
+      },
+      prefix: {
+        default: 'at:',
+        env: 'ACCESS_TOKEN_REDIS_KEY_PREFIX',
+        format: String,
+        doc: 'Key prefix for access tokens in Redis',
+      },
+    },
     sessionTokens: {
       enabled: {
         default: true,
--- a/packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
+++ b/packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
@@ -1,14 +1,18 @@
 local userId = KEYS[1]
-local tokenIds = redis.call('smembers', userId)
-local result = {}
-if tokenIds then
-  for _, tokenId in ipairs(tokenIds) do
-    local t = redis.call('get', tokenId)
-    if t then
-      table.insert(result, t)
+local ids = redis.call('smembers', userId)
+local results = {}
+if #ids > 0 then
+  local toDel = {}
+  local values = redis.call('mget', unpack(ids))
+  for i, value in ipairs(values) do
+    if value then
+      table.insert(results, value)
     else
-      redis.call('srem', userId, tokenId);
+      table.insert(toDel, ids[i])
     end
   end
+  if #toDel > 0 then
+    redis.call('srem', userId, unpack(toDel));
+  end
 end
-return result
+return results
--- a/packages/fxa-auth-server/lib/luaScripts/removeAccessToken_1.lua
+++ b/packages/fxa-auth-server/lib/luaScripts/removeAccessToken_1.lua
@@ -1,8 +1,6 @@
-local tokenId = KEYS[1]
-local _, _, prefix = tokenId:find('(%a+:)')
-local value = redis.call('get', tokenId)
-if value then
-  local token = cjson.decode(value)
-  redis.call('srem', prefix..token.userId, tokenId)
-  return redis.call('del', tokenId)
+local id = KEYS[1]
+local exists = redis.call('exists', id)
+if exists then
+  redis.call('del', id)
 end
+return exists
--- a/packages/fxa-auth-server/lib/luaScripts/removeAccessTokensForPublicClients_1.lua
+++ b/packages/fxa-auth-server/lib/luaScripts/removeAccessTokensForPublicClients_1.lua
@@ -1,18 +1,20 @@
 local uid = KEYS[1]
 local ids = redis.call('smembers', uid)
 local toDel = {}
-for _, id in ipairs(ids) do
-  local v = redis.call('get', id)
-  if v then
-    local t = cjson.decode(v)
-    if t.canGrant or t.publicClient then
-      table.insert(toDel, id)
+if #ids > 0 then
+  local values = redis.call('mget', unpack(ids))
+  for i, value in ipairs(values) do
+    if value then
+      local token = cjson.decode(value)
+      if token.canGrant or token.publicClient then
+        table.insert(toDel, ids[i])
+      end
+    else
+      table.insert(toDel, ids[i])
     end
-  else
-    table.insert(toDel, id)
   end
-end
-if #toDel > 0 then
-  redis.call('srem', uid, unpack(toDel))
-  redis.call('unlink', unpack(toDel))
+  if #toDel > 0 then
+    redis.call('srem', uid, unpack(toDel))
+    redis.call('unlink', unpack(toDel))
+  end
 end
--- a/packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
+++ b/packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
@@ -8,4 +8,4 @@ if ttl then
   table.insert(args, ttl)
 end
 redis.call('set', tokenId, unpack(args))
-return redis.call('sadd', userId, tokenId)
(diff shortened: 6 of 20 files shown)


packages/fxa-auth-server/config/dev.json
```diff
@@ -206,7 +206,9 @@
     },
     "clientIdToServiceNames": {
       "dcdb5ae7add825d2": "123done",
-      "98e6508e88680e1a": "fxa-settings"
+      "98e6508e88680e1a": "fxa-settings",
+      "7377719276ad44ee": "pocket-mobile",
+      "749818d3f2e7857f": "pocket-web"
```
dannycoates: In order to "quarantine" pocket access tokens in mysql I opted to use this existing bit of config. These are already assigned in prod here: https://github.com/mozilla-services/cloudops-deployment/blob/8adf05f820c05fe6b29b315e557df9afe3baaeba/projects/fxa/puppet/yaml/app/fxa.prod.yaml

packages/fxa-auth-server/config/index.js
```diff
@@ -337,6 +337,24 @@ const conf = convict({
       format: 'port',
       doc: 'Port for Redis server',
     },
+    accessTokens: {
+      host: {
+        default: '127.0.0.1',
+        env: 'ACCESS_TOKEN_REDIS_HOST',
+        format: String,
+      },
```
dannycoates: these should live on their own instance in prod, but defaults to the shared one locally

packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
```diff
@@ -1,14 +1,18 @@
...
-if tokenIds then
-  for _, tokenId in ipairs(tokenIds) do
-    local t = redis.call('get', tokenId)
-    if t then
-      table.insert(result, t)
+local ids = redis.call('smembers', userId)
+local results = {}
+if #ids > 0 then
+  local toDel = {}
+  local values = redis.call('mget', unpack(ids))
```
dannycoates: I switched this to an `mget` to reduce the count of redis calls

packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
```diff
@@ -1,14 +1,18 @@
...
+  for i, value in ipairs(values) do
+    if value then
+      table.insert(results, value)
     else
-      redis.call('srem', userId, tokenId);
+      table.insert(toDel, ids[i])
     end
   end
+  if #toDel > 0 then
+    redis.call('srem', userId, unpack(toDel));
```
dannycoates: switched to a single `srem` instead of doing them in the loop

packages/fxa-auth-server/lib/luaScripts/removeAccessToken_1.lua
```diff
@@ -1,8 +1,6 @@
-local tokenId = KEYS[1]
-local _, _, prefix = tokenId:find('(%a+:)')
```
dannycoates: this prefix logic was fragile and since `getAccessTokens` prunes missing ids from the index it's better to just let the index get updated there.

packages/fxa-auth-server/lib/luaScripts/removeAccessTokensForPublicClients_1.lua
```diff
@@ -1,18 +1,20 @@
...
 local ids = redis.call('smembers', uid)
 local toDel = {}
-for _, id in ipairs(ids) do
-  local v = redis.call('get', id)
-  if v then
-    local t = cjson.decode(v)
-    if t.canGrant or t.publicClient then
-      table.insert(toDel, id)
+if #ids > 0 then
+  local values = redis.call('mget', unpack(ids))
```
dannycoates: again `mget` and single `srem` for "efficiency"

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
```diff
@@ -8,4 +8,4 @@ if ttl then
   table.insert(args, ttl)
 end
 redis.call('set', tokenId, unpack(args))
-return redis.call('sadd', userId, tokenId)
```
dannycoates: the result of `sadd` isn't used by the caller

packages/fxa-auth-server/lib/oauth/db/accessToken.js
```diff
@@ -40,7 +40,7 @@ class AccessToken {
     /** @type {Buffer} */
     this.token = token || unique.token();
     /** @type {Date} */
-    this.createdAt = createdAt || new Date();
+    this.createdAt = createdAt || new Date(new Date().setMilliseconds(0));
```
dannycoates: Dates in mysql were only down to the second. Some tests assumed this and it's easier and more consistent to leave it as such.

packages/fxa-auth-server/lib/oauth/db/accessToken.js
```diff
@@ -63,6 +63,14 @@ class AccessToken {
     return this.expiresAt.getTime() - Date.now();
   }
 
+  get lastAccessTime() {
+    return this.createdAt;
+  }
+
+  get id() {
+    return this.clientId;
+  }
```
dannycoates: aliases to keep some parts of the code happy with mixed `AccessToken` and mysql results

packages/fxa-auth-server/test/oauth/api.js
```diff
@@ -2849,38 +2852,7 @@ describe('/v1', function() {
...
           assert.equal(res.statusCode, 400);
           assertSecurityHeaders(res);
-          assert.equal(res.result.errno, 115);
-        })
-        .finally(function() {
-          config.set('oauthServer.expiration.accessTokenExpiryEpoch', epoch);
-        });
-    });
-
-    it('should accept expired tokens from before the epoch', function() {
```
dannycoates: There's a shift in behavior here since redis tokens are actually deleted when they expire so there's no "token exists but is expired" state (except for pocket tokens which are still in mysql)

I think it makes sense to return a 108 "Invalid token" here.

packages/fxa-auth-server/test/oauth/db/purgeExpiredTokens.js
```diff
@@ -39,13 +39,15 @@ function makeTests(name, purgeMethod) {
       });
     }
 
-    // Inserts 2000 access tokens with the following breakdown
-    // ClientIdA - 500 expired, 500 valid
-    // ClientIdB - 500 expired, 500 valid
```
dannycoates: These tests are slow with 2000 and I think 200 is enough exercise for this

packages/fxa-auth-server/test/oauth/db/purgeExpiredTokens.js
```diff
@@ -39,13 +39,15 @@ function makeTests(name, purgeMethod) {
...
+    // ClientIdA - 50 expired, 50 valid
+    // ClientIdB - 50 expired, 50 valid
     before('setup clients', function() {
       email = 'asdf@asdf.com';
-      clientIdA = randomString(8);
-      clientIdB = randomString(8);
+      // These are Pocket client ids.
+      // Only Pocket clients are stored in mysql.
+      clientIdA = '7377719276ad44ee';
+      clientIdB = '749818d3f2e7857f';
```
dannycoates: to store in mysql we need to use clientIds from the config for pocket

packages/fxa-auth-server/lib/oauth/grant.js
```diff
@@ -138,9 +138,11 @@ module.exports.validateRequestedGrant = async function validateRequestedGrant(
   //   * Is this client allowed to request ACCESS_TYPE_OFFLINE?
   //   * Is this client allowed to request all the non-key-bearing scopes?
   //   * Do we expect this client to be using OIDC?
-
   return {
     clientId: client.id,
+    name: client.name,
+    canGrant: client.canGrant,
+    publicClient: client.publicClient,
```
dannycoates: To save an extra lookup query the access tokens now store this inline

packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
```diff
@@ -1,14 +1,18 @@
...
-local tokenIds = redis.call('smembers', userId)
-local result = {}
-if tokenIds then
-  for _, tokenId in ipairs(tokenIds) do
-    local t = redis.call('get', tokenId)
-    if t then
-      table.insert(result, t)
+local ids = redis.call('smembers', userId)
+local results = {}
+if #ids > 0 then
```
vbudhram: To check my understanding, this returns all access tokens for a userId. If there is a token stored in the set returned but doesn't have a record in redis, it gets deleted from the token set?

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
```diff
@@ -540,7 +525,7 @@ MysqlStore.prototype = {
    * @param {String} uid User ID as hex
    * @returns {Promise}
    */
-  getAccessTokensByUid: async function getAccessTokensByUid(uid) {
+  _getAccessTokensByUid: async function _getAccessTokensByUid(uid) {
```
vbudhram: Should this be `async function getAccessTokensByUid`?

packages/fxa-auth-server/lib/redis.js
```diff
@@ -111,6 +111,9 @@ class FxaRedis {
    * @param {AccessToken} token
    */
   setAccessToken(token) {
+    if (token.ttl < 1) {
```
vbudhram: It is unclear to me why this was added? Is this to avoid storing pocket tokens?

packages/fxa-auth-server/lib/luaScripts/getAccessTokens_1.lua
dannycoates: correct

packages/fxa-auth-server/lib/redis.js
dannycoates: it's a guard against setting and invalid ttl. redis would throw an error with a negative ttl. it shouldn't happen in prod, maybe we should log it...

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
dannycoates: i don't think so. i renamed all the functions that `db/index.js` implements so that the part that copies the functions from here doesn't overwrite them.

clouserw:
> @dannycoates Thanks for leaving some notes, I think some might be useful as comments. Overall, I had a few questions but nothing blocking. Nice work 👍🏽

I was thinking that same thing.  I'm a fan of comments in the code. :)

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
dannycoates: oic... i wasn't consistent in renaming both sides on all the functions. i'll fix that

Issue #4207, opened 2020-02-19
[admin panel A1.0] Implement the graphQL resolvers for showing a user


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1142)
┆Issue Number: FXA-1142

Issue #4206, opened 2020-02-19
[admin panel A1.1] Implement graphQL mutation for removing an email block


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1143)
┆Issue Number: FXA-1143

PR #4269, opened 2020-02-25
feat(admin-server): add db and resolvers

### LINKED ISSUE #4202: [admin panel A1.0/A1.1] Create the ORM objects for the graphQL schema, update config
We need to create the ORM objects needed for the graphQL schema and then update the config with user setting.

Note: We need to join the email table with the accounts table.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1141)
┆Issue Number: FXA-1141

bbangert:
Because:

* User admin needs to talk to the db and resolve the graphql
  objects.

This commit:

* Adds the db models to talk to the necessary tables in the database.
* Adds the resolvers to lookup the data and return it.

Fixes #4202, #4206, #4207

### DIFF
--- a/packages/fxa-admin-server/src/lib/db/models/emails.ts
+++ b/packages/fxa-admin-server/src/lib/db/models/emails.ts
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Model } from 'objection';
+
+import { intBoolTransformer, uuidTransformer } from '../transformers';
+
+const binaryFields = ['uid', 'emailCode'];
+const booleanFields = ['isVerified', 'isPrimary'];
+
+export class Emails extends Model {
+  public static tableName = 'emails';
+
+  public normalizedEmail!: string;
+  public email!: string;
+  public uid!: string;
+  public isVerified!: boolean;
+  public isPrimary!: boolean;
+  public emailCode!: string;
+  public verifiedAt?: number;
+  public createdAt!: number;
+
+  public $parseDatabaseJson(json: any) {
+    json = super.$parseDatabaseJson(json);
+    for (const field of binaryFields) {
+      if (json[field]) {
+        json[field] = uuidTransformer.from(json[field]);
+      }
+    }
+    for (const field of booleanFields) {
+      if (json[field]) {
+        json[field] = intBoolTransformer.from(json[field]);
+      }
+    }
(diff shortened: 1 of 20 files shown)


packages/fxa-admin-server/src/lib/db/models/emails.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (json[field]) {
+        json[field] = intBoolTransformer.from(json[field]);
+      }
+    }
+    return json;
+  }
+
+  public $formatDatabaseJson(json: any) {
+    json = super.$formatDatabaseJson(json);
+    for (const field of binaryFields) {
```
lmorchard: Could be worth tossing this binary & boolean conversion handling into a shared superclass, since this stuff happens for both `accounts` & `emails` models. Then again, it's only two models, so might be a premature optimization

packages/fxa-admin-server/src/test/lib/db/models/accounts.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE `accounts` (
```
lmorchard: I don't suppose there's any way to sync this with the schema from `fxa-auth-db-mysql` is there? Probably not, since the schema there consists only of the 100+ patches to apply. Would be nice to test against the "real" thing though.

packages/fxa-admin-server/src/lib/db/transformers.ts
```diff
@@ -0,0 +1,25 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/;
+
+function unbuffer(val: string | Buffer) {
+  return Buffer.isBuffer(val) ? val.toString('hex') : val;
+}
+
+function buffer(value: string | Buffer) {
+  // Don't convert things with no value, but we still want
```
lmorchard: Might be worth a small test or two to exercise these error conditions

packages/fxa-admin-server/src/lib/db/models/emails.ts
bbangert: Yea, I figured as this grows to tackle more of the db, a shared base class that lets one define uuid/binary/intBool field conversions would be handy.

packages/fxa-admin-server/src/test/lib/db/models/accounts.sql
bbangert: Not that I could see, this schema doesn't actually exist in one place as-is, I had to extract the DDL statements to create the current table directly from MySQL.

PR #4270, opened 2020-02-25
feat(metrics): collect navigation timing on Payments with StatsD

### LINKED ISSUE #3596: Collect Page Performance Metrics with StatsD on Payments Server
Content server uses flow events to collect performance metrics (https://sql.telemetry.mozilla.org/dashboard/fxa-performance).  But we ran into some [speed hump](https://github.com/mozilla/fxa/issues/3072)s when trying to replicate that in Payments.

We should consider using StatsD instead.  Let's try it with Payments first, then move Content over as well if we like the result.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-43)
┆Issue Number: FXA-43

chenba:
This patch uses sendBeacon and Navigation Timing Level 2
(https://www.w3.org/TR/navigation-timing-2/) to report performance
timing from Payments pages.  The metrics are collected with StatsD.

Fixes #3596 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-payments-server/server/lib/routes/navigation-timing.js
+++ b/packages/fxa-payments-server/server/lib/routes/navigation-timing.js
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const joi = require('joi');
+const URL = require('url').URL;
+
+const TIMESTAMP_MS = joi
+  .number()
+  .integer()
+  .required();
+// Validate the subset of PerformanceNavgationTiming properties used here.
+// (Ref: https://www.w3.org/TR/navigation-timing-2/#dom-performancenavigationtiming)
+const navigationTimingSchema = joi.object().keys({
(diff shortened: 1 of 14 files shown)


chenba:
There are a couple tests in this PR that are not related to the feature.  They've been added because there's a coverage threshold to reach in order to pass CI.  It should be obvious which tests are the unrelated ones.

packages/fxa-payments-server/server/lib/routes/navigation-timing.js
```diff
@@ -0,0 +1,76 @@
...
+    .string()
+    .uri()
+    .required(),
+  redirectStart: TIMESTAMP_MS,
+  requestStart: TIMESTAMP_MS.min(joi.ref('domainLookupStart')),
+  responseEnd: TIMESTAMP_MS.min(joi.ref('responseStart')),
+  responseStart: TIMESTAMP_MS.min(joi.ref('requestStart')),
+});
+
+module.exports = (geolocate, UAParser, statsd) => ({
```
dannycoates: how come `UAParser` is injected vs `require`ed here?

chenba: It makes testing a little easier, kind of.  Although `jest.mock` is pretty easy.  So, no real good reason. 😅

PR #4275, opened 2020-02-25
test(auth-server): coverage for updatePayment

### LINKED ISSUE #4224: Code coverage for Direct Stripe Route - Update Payment
fxa-auth-server/lib/routes/subscriptions.js

Function: DirectStripeRoutes.updatePayment

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1173)
┆Issue Number: FXA-1173

jackiemunroe:
- added code coverage for DirectStripeRoutes.updatePayment

fixes #4224

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/subscriptions.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions.js
@@ -2320,7 +2320,92 @@ describe('DirectStripeRoutes', () => {
     });
   });
 
-  describe('updatePayment', () => {});
+  describe('updatePayment', () => {
+    let request;
+    const paymentToken = 'tok_visa';
+
+    beforeEach(() => {
+      request = { ...VALID_REQUEST, payload: { paymentToken } };
+
+      directStripeRoutesInstance.stripeHelper.updateCustomerPaymentMethod.returns();
+    });
+
+    describe('when the customer exists', () => {
+      it('updates the payment method', async () => {
+        directStripeRoutesInstance.stripeHelper.fetchCustomer.returns(
+          customerFixture
+        );
+
+        await directStripeRoutesInstance.updatePayment(request);
+
+        assert.isTrue(
+          directStripeRoutesInstance.log.begin.calledOnceWith(
+            'subscriptions.updatePayment',
+            request
+          )
+        );
+
+        assert.isTrue(
+          directStripeRoutesInstance.customs.check.calledOnceWith(
+            request,
+            TEST_EMAIL,
+            'updatePayment'
+          )
+        );
+
+        assert.isTrue(
+          directStripeRoutesInstance.stripeHelper.updateCustomerPaymentMethod.calledOnceWith(
+            customerFixture.id,
+            paymentToken
+          )
+        );
+
+        assert.isTrue(
+          directStripeRoutesInstance.log.info.calledOnceWith(
+            'subscriptions.updatePayment.success',
+            { uid: UID }
+          )
+        );
+      });
+    });
+
+    describe('when the customer does not exist', () => {
+      it('throws an error', async () => {
+        directStripeRoutesInstance.stripeHelper.fetchCustomer.returns();
+
+        try {
+          await directStripeRoutesInstance.updatePayment(request);
+          assert.fail('Method epected to reject');
+        } catch (err) {
+          assert.equal(err.errno, error.ERRNO.BACKEND_SERVICE_FAILURE);
+          assert.equal(err.message, 'A backend service request failed.');
+          assert.equal(err.output.payload['service'], 'stripe');
+          assert.equal(err.output.payload['operation'], 'updatePayment');
+
+          assert.isTrue(
+            directStripeRoutesInstance.log.begin.calledOnceWith(
+              'subscriptions.updatePayment',
+              request
+            )
+          );
+
+          assert.isTrue(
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -2319,7 +2319,92 @@ describe('DirectStripeRoutes', () => {
...
   });
 
-  describe('updatePayment', () => {});
+  describe('updatePayment', () => {
+    let request;
+    const paymentToken = 'tok_visa';
+
+    beforeEach(() => {
+      request = Object.create(VALID_REQUEST);
+      request.payload = { paymentToken };
```
chenba: Any reason you favor `Object.create` over something like `request = {...VALID_REQUEST, payload: {paymentToken}};`?

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -2319,7 +2319,92 @@ describe('DirectStripeRoutes', () => {
...
+          )
+        );
+      });
+    });
+
+    describe('when the customer does not exist', () => {
+      it('throws an error', () => {
+        directStripeRoutesInstance.stripeHelper.fetchCustomer.returns();
+
+        return directStripeRoutesInstance.updatePayment(request).then(
```
chenba: Mixing `async/await` in the previous test and promises here is kind of weird.  Can you also using `async/await` here, and `try/catch` the error you want to test against?

chenba:
nit: subject of commit message doesn't conform to https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#subject

packages/fxa-auth-server/test/local/routes/subscriptions.js
jackiemunroe: No reason in particular, it was just where my mind went. Updating.

packages/fxa-auth-server/test/local/routes/subscriptions.js
jackiemunroe: I can do that. I saw errors tested different ways in the code base when it came to promises - `try/catch` and `.then(()=>{}, err=>{})`. It didn't seem that we had a preferred methodology for error testing.

jackiemunroe:
> nit: subject of commit message doesn't conform to https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#subject

Updated commit subject

chenba:
> > nit: subject of commit message doesn't conform to https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#subject
> 
> Updated commit subject

Oh man, sorry I wasn't specific, so now I'm gonna sound super extra nit...  my comment was mainly about the phrasing of the commit message.  Although now that I've re-read the first bullet point on the contributing doc, it's a bit unclear if it's asking for the same thing, which is: start with a verb.

With that, the commit message would read "[implicit subject of patch/commit] does this thing".

We've been fairly but not completely consistent with that.

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -2319,7 +2319,91 @@ describe('DirectStripeRoutes', () => {
     });
   });
 
-  describe('updatePayment', () => {});
+  describe('updatePayment', () => {
+    let request;
+    const paymentToken = 'tok_visa';
+
+    beforeEach(() => {
+      request = request = { ...VALID_REQUEST, payload: { paymentToken } };
```
chenba: Double assignment?

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -2319,7 +2319,91 @@ describe('DirectStripeRoutes', () => {
...
+
+      directStripeRoutesInstance.stripeHelper.updateCustomerPaymentMethod.returns();
+    });
+
+    describe('when the customer exists', () => {
+      it('updates the payment method', async () => {
+        const customer = Object.create(customerFixture);
```
chenba: I think there's no need to new up an object here.  I don't see a reason why passing `customerFixture` on the next line wouldn't work.

packages/fxa-auth-server/test/local/routes/subscriptions.js
jackiemunroe: aha good catch... that's embarrassing ':D

packages/fxa-auth-server/test/local/routes/subscriptions.js
jackiemunroe: I can update. Thanks :)

jackiemunroe:
> > > nit: subject of commit message doesn't conform to https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#subject
> > 
> > 
> > Updated commit subject
> 
> Oh man, sorry I wasn't specific, so now I'm gonna sound super extra nit... my comment was mainly about the phrasing of the commit message. Although now that I've re-read the first bullet point on the contributing doc, it's a bit unclear if it's asking for the same thing, which is: start with a verb.
> 
> With that, the commit message would read "[implicit subject of patch/commit] does this thing".
> 
> We've been fairly but not completely consistent with that.

Ah. I see. Yeah, I read it as "if you use a verb in the subject, then it should use the imperative, present tense"

PR #4277, opened 2020-02-25
task(payments-server): add PaymentConfirmation component

### LINKED ISSUE #4143: Create confirmation info view for subscription confirmation
* add/ensure we can get invoice number from api
* add view
* add tests

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1100)
┆Issue Number: FXA-1100
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12499/confirm-info-view.png">confirm-info-view.png</a>

meandavejustice:
- fixes #4143

Opening early so any eager 👀 can take a look. This is the payment confirmation component, once this is merged I'll be able to paste together all of our new components to create the PaymentConfirmation view and test with multiple products.

There are a couple tasks left here:
- fetch invoice number and pass down to component (this will require and update to auth-server stripe route)
- fetch credit card type and display correct image
- Add breakpoints for mobile views
- make sure we have proper test coverage
- run svgs through optimizer


LZoog:
Looking good so far, we probably want our button hover/active styles on the "click to download" link.

Also seeing a `border-bottom` here on `.footer`:
image

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -262,6 +262,8 @@ class StripeHelper {
     }
     const customer = customerResponse[0];
 
+    this.log.error(`CUSTOMER.RESPONSE:::::`, customer);
```
LZoog: Possible missed log.

packages/fxa-auth-server/lib/subhub/stubAPI.js
```diff
@@ -18,9 +18,10 @@ module.exports.buildStubAPI = function buildStubAPI(log, config) {
 
   const customer = {
     payment_type: 'credit',
-    last4: '8675',
+    last4: '7777',
```
LZoog: I confirmed with Ben `/lib/subhub/` code is essentially dead, we shouldn't need this. 😎

packages/fxa-auth-server/test/local/payments/fixtures/customer1.json
```diff
@@ -53,6 +54,7 @@
         "current_period_end": 1568573770,
         "cancel_at_period_end": false,
         "ended_at": null,
+        "latest_invoice": "628031D-0002",
```
LZoog: Hmm from the [Stripe docs](https://stripe.com/docs/api/subscriptions/object#subscription_object-latest_invoice), it looks like `latest_invoice` only comes from a subscription object, not a customer object.

image

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -110,6 +110,8 @@ export const Product = ({
     );
   }
 
+  console.log('CUSTOMER PRODUCT ROUTE', customer);
```
LZoog: Console log here

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -454,13 +454,16 @@ class DirectStripeRoutes {
       // Currently assume a single source, and we can only access these attributes
       // on cards.
       const src = customer.sources.data[0];
+      // const latestInvoice = customer.subscriptions.data[0]['latest_invoice'];
```
LZoog: Is this WIP?

packages/fxa-auth-server/test/local/payments/fixtures/customer1.json
meandavejustice: I think github had collapsed, but this is in a subscription object, inside of the customer object

meandavejustice:
@LZoog Got the merge situation cleaned up

meandavejustice:
I'll take care of removing the bottom and top box shadows when I add it actually into the content. I think I'll probably leave the box shadow and hide it with header/footers

packages/fxa-payments-server/src/components/PaymentConfirmation/index.stories.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import MockApp from '../../../.storybook/components/MockApp';
+import PaymentConfirmation from './index';
+import { Customer, Profile, Plan } from '../../store/types';
+
+const userProfile: Profile = {
+  avatar: 'http://placekitten.com/256/256',
+  displayName: 'Foxy77',
```
LZoog: Should these fields be populating?

image

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const PaymentConfirmation = ({
+  customer,
+  profile,
+  selectedPlan,
+  className = 'default',
+}: PaymentConfirmationProps) => {
+  const { amount, interval } = selectedPlan;
+  const { displayName, email } = profile;
+  const { brand, last4, subscriptions } = customer;
+  const invoiceNumber = subscriptions[0].latest_invoice;
```
LZoog: If a user has multiple subscriptions, will this one always be the first in the array or should we do something like find the matching `plan_name`?

(what you have could be right, just thinking out-loud here without doing extra digging)

packages/fxa-payments-server/src/components/PaymentConfirmation/index.stories.tsx
meandavejustice: This is a known issue where variables in Localized elements don't populate in our storybook

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
meandavejustice: based on the way it was explained to me from @jackiemunroe, it should be the first in the array. But i'd like to verify

packages/fxa-payments-server/src/components/PaymentConfirmation/index.scss
```diff
@@ -0,0 +1,122 @@
...
+    background: url('./images/discover.svg') no-repeat;
+  }
+
+  .bottom-row {
+    color: $grey-50;
+    display: flex;
+    justify-content: space-between;
+
+    p:last-child {
+      color: $grey-40;
```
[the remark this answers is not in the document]
meandavejustice: cc/ @johngruen ^^^^^^^ 👀

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -738,6 +738,11 @@ class StripeHelper {
       };
 
       const planName = `${productName} ${intervalDict[sub.plan.interval]}`;
+      let latestInvoice = sub.latest_invoice;
+
+      if (latestInvoice && typeof latestInvoice !== 'string') {
+        latestInvoice = latestInvoice.number;
```
jackiemunroe: We probably would want to be using `latestInvoice.id` when we have an invoice object. While there is an `latestInvoice.number`, it is not the same thing that we would see if `latestInvoice` is not expanded. So, for consistency, there are two options:
1. use `id` field, when not expanded
2. if the invoice is not expanded, fetch it, and always use `number`

packages/fxa-payments-server/src/components/PaymentConfirmation/index.scss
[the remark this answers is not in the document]
johngruen: yeah make it darker...istherelike a grey-50?

packages/fxa-auth-server/test/local/payments/fixtures/customer1.json
jackiemunroe: The latest invoice is an invoice id, it will have a pattern similar to `in_1FObmFKb9q6OnNsLTMm1ERZE`

jackiemunroe: There are invoice samples in the fixtures folder.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
jackiemunroe: @meandavejustice This is currently true, but it in the future it will change if we allow multiple subscriptions. I would make a note to update this in the future to search the subscriptions for the desired plan.

packages/fxa-auth-server/lib/payments/stripe.js
LZoog: 👍 TY Jackie, I'm going to link your comment in #4311.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
lmorchard: IIRC, we technically support multiple subscriptions right now. We just don't have multiple products to subscribe to. And the multiple subscriptions are all independently paid for, which is not ideal and a thing we want to work on.

PR #4280, opened 2020-02-26
feat(admin-server): add call logging

### LINKED ISSUE #4245: Logging of mutations and views
As the admin server is a sensitive system, we should log use of it at info level. Mutations and queires should have the active user logged along with the query/mutation.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1187)
┆Issue Number: FXA-1187

bbangert:
Because:

* We want an audit trail of any actions performed.

This commit:

* Add's query/mutation logging including the auth user header that
  will be passed in and the action performed.

Closes #4245


packages/fxa-admin-server/src/test/lib/resolvers/account-resolver.spec.ts
```diff
@@ -29,32 +30,45 @@ const EMAIL_BOUNCE_1 = randomEmailBounce(USER_1.email);
...
+    schema = await buildSchema({
+      resolvers: [AccountResolver, EmailBounceResolver]
+    });
   });
 
   after(async () => {
     await knex.destroy();
   });
 
+  beforeEach(async () => {
```
vbudhram: nit: might be easier to spot the `beforeEach` if above the `after`

packages/fxa-admin-server/src/test/lib/resolvers/account-resolver.spec.ts
```diff
@@ -29,32 +30,45 @@ const EMAIL_BOUNCE_1 = randomEmailBounce(USER_1.email);
 const USER_2 = randomAccount();
 
 describe('accountResolver', () => {
+  // Used to get the shape of the return type
```
vbudhram: Comment/code is a little unclear to me (probably since not very familiar with typescript), but I take it this tells compiler that `contextDummy` is a boolean value?

packages/fxa-admin-server/src/test/lib/resolvers/account-resolver.spec.ts
```diff
@@ -29,32 +30,45 @@ const EMAIL_BOUNCE_1 = randomEmailBounce(USER_1.email);
...
+  const contextDummy = (false as true) && mockContext();
```
chenba: Does `let context: ReturnType<mockContext>;` work?

bbangert: Nope, I tried that one first before googling for this solution. It's because mockContext is a value, not a type..... and in typing that, I realized what *does* work!
```let context: ReturnType<typeof mockContext>;```

packages/fxa-admin-server/src/test/lib/resolvers/account-resolver.spec.ts
bbangert: Yea, found a better solution though.

PR #4285, opened 2020-02-26
fix(delete-account): Include subscription status filtering for 'trialing' and 'past_due'

### LINKED ISSUE #4166: (fxa-content-server) Update 'delete_account' subscription filtering to include 'trialing' and 'past_due'
In the delete_account view, we filter subscriptions in `_fetchActiveSubscriptions` as well as `_getNumberOfProducts` for those that match `.status === 'active'`. We should also include those with a status of 'trialing' and 'past_due'.

We could also update the endpoint we hit here to `subscriptions/active` but it's probably not worth the effort as we would still need to filter out subscriptions with a status of 'canceled'.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1116)
┆Issue Number: FXA-1116

LZoog:
fixes #4166

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/delete_account.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/delete_account.js
@@ -86,8 +86,8 @@ var View = FormView.extend({
     const start = Date.now();
     return account.settingsData().then(({ subscriptions = [] } = {}) => {
       this.logFlowEvent(`timing.settings.fetch.${Date.now() - start}`);
-      this._activeSubscriptions = subscriptions.filter(
-        subscription => subscription.status === 'active'
+      this._activeSubscriptions = subscriptions.filter(subscription =>
+        ['trialing', 'active', 'past_due'].includes(subscription.status)
       );
     });
   },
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/views/settings/delete_account.js
```diff
@@ -121,19 +121,14 @@ var View = FormView.extend({
...
       if (client.isOAuthApp === true) {
         numberOfProducts++;
       }
     }
-    // eslint-disable-next-line no-unused-vars
-    for (const sub of this._activeSubscriptions) {
-      if (sub.plan_id && sub.status === 'active') {
-        numberOfProducts++;
-      }
-    }
```
LZoog: We call `_getNumberOfProducts` after we filter subscriptions in `_fetchActiveSubscriptions` and set `_activeSubscriptions` so we should be able to set the number on the `.length` of the array instead of filtering through again.

packages/fxa-content-server/app/tests/spec/views/settings/delete_account.js
```diff
@@ -26,7 +26,7 @@ describe('views/settings/delete_account', function() {
   const UID = '123';
   const password = 'password';
   let account;
-  let activeSubscriptions;
+  let subscriptions;
```
LZoog: I changed this var name as `subscriptions` is more accurate since it's the mock return for  `account.settingsData()`, before we've filtered for active subscriptions.

packages/fxa-content-server/app/tests/spec/views/settings/delete_account.js
```diff
@@ -398,20 +406,8 @@ describe('views/settings/delete_account', function() {
         );
       });
 
-      it('renders only `active: true` subscriptions', () => {
-        assert.lengthOf(view.$('.delete-account-product-subscription'), 1);
-      });
-
-      it('renders subscription title attributes', () => {
```
LZoog: We had a couple of duplicate tests here that I removed, we're still testing subscription title attributes on line 377 and testing for active subscriptions on line 373.

PR #4304, opened 2020-02-27
Store refresh_token last-used-at timestamp in redis.

### LINKED ISSUE #4146: Mitigate db write load of maintaining refresh_token "last used at"
The OAuth refresh_tokens database table contains a "last used at" column giving the timestamp at which the token was last used to generate on access token. We use this information as part of the user's "devices and apps" list to let them know which connected clients have been recently active.

There's currently [some logic](https://github.com/mozilla/fxa/blob/e6f111cbf8cb41f25103eb34503ab83468465088/packages/fxa-auth-server/lib/oauth/routes/token.js#L333) to guard against writing that column to the db too frequently, because writes can be expensive at scale. But that'll only get us so far, and the broader we make the definition of "too frequently" the less useful this information becomes to the user.

We've already been through this once in the past, with sessionTokens. For sessionTokens we wound up [storing the frequently-changing metadata in redis](https://github.com/mozilla/fxa/blob/e6f111cbf8cb41f25103eb34503ab83468465088/packages/fxa-auth-server/lib/db.js#L290) rather than writing it through to the db, on the logic that it's better to have information that's usually accurate but with a small risk of disappearing entirely, than to have information that's never particularly accurate.

I propose that we do the same for refresh tokens, storing last-used-at in redis in the same way that we currently do for sessionTokens, to mitigate potential scaling issues for when we get a big influx of new refresh-token-using OAuth clients with the Fennec->Fenix migration.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1104)
┆Issue Number: FXA-1104

rfk:
This PR will move storage of the `lastUsedAt` field on OAuth refresh tokens into redis, where it can be written to liberally in memory without crushing disk-based storage.

It is clearly not functional yet, but I'm pushing it for early visibility and to checkin with @dannycoates on some factoring as it relates to the access-tokens-in-redis PR.

Fixes #4146.

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/db/index.js
+++ b/packages/fxa-auth-server/lib/oauth/db/index.js
@@ -10,8 +10,13 @@ const config = require('../../../config');
 const encrypt = require('../encrypt');
 const logger = require('../logging')('db');
 const mysql = require('./mysql');
-const redis = require('../../redis');
+const redis = require('./redis');
 const AccessToken = require('./accessToken');
+const RefreshTokenMetadata = require('./refreshTokenMetadata');
+
+const REFRESH_LAST_USED_AT_UPDATE_AFTER_MS = config.get(
+  'oauthServer.refreshToken.updateAfter'
+);
 
 function getPocketIds(idNameMap) {
   return Object.entries(idNameMap)
@@ -30,14 +35,9 @@ class OauthDB {
       await preClients();
       await scopes();
     });
-    this.redis = redis(
-      {
-        ...config.get('redis.accessTokens'),
-        enabled: true,
-        maxttl: config.get('oauthServer.expiration.accessToken'),
-      },
-      logger
-    );
+
+    this.redis = redis();
+
     Object.keys(mysql.prototype).forEach(key => {
       const self = this;
       this[key] = async function() {
@@ -46,6 +46,7 @@ class OauthDB {
       };
     });
   }
+
   disconnect() {}
 
   async generateAccessToken(vals) {
@@ -83,11 +84,11 @@ class OauthDB {
     return db._getAccessToken(id);
   }
 
-  async removeAccessToken(id) {
-    const done = await this.redis.removeAccessToken(id);
+  async removeAccessToken(token) {
+    const done = await this.redis.removeAccessToken(token.tokenId);
     if (!done) {
       const db = await this.mysql;
-      return db._removeAccessToken(id);
+      return db._removeAccessToken(token.tokenId);
     }
   }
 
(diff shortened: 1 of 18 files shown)


packages/fxa-auth-server/lib/oauth/db/index.js
```diff
@@ -31,10 +31,15 @@ class OauthDB {
...
-    this.redis = redis(
+    this.redisAccessTokens = redis(
       { ...config.get('redis.accessTokens'), enabled: true },
       logger
     );
+    this.redisRefreshTokens = redis(
+      { ...config.get('redis.refreshTokens') },
+      logger
+    );
+
```
rfk: @dannycoates seeing this next to your work on storing the access tokens in redis, does it still make sense to store the refresh-tokens stuff in a different redis instance (the one with the sessiontokens) or would it be cleaner to re-use the redis it's already got from your PR?  I recall we may have wanted to keep the access-tokens one separate for operational reasons.

dannycoates: Maybe it makes sense to have the line mirror the auth/oauth mysql db split? I'm not thinking of any ways having this metadata alongside access tokens would be a security issue.

rfk: > Maybe it makes sense to have the line mirror the auth/oauth mysql db split? 

Not 100% sure what you mean here; is it "there's an auth mysql + auth redis, and there's an oauth mysql + oauth redis"? That makes sense to me FWIW.

dannycoates: That’s what I meant 😀

rfk: So this turns out to be kind of awkward because of the way the redis abstractions sets a key prefix as though it's a conection-level settings, but I think I found a way to encapsulate that in an acceptable fashion. We'll see how it shakes out...

rfk:
I had to shave a few yaks, but I think this is ready for review. @dannycoates mind taking a look?

packages/fxa-auth-server/lib/oauth/db/index.js
```diff
@@ -80,11 +85,11 @@ class OauthDB {
     return db._getAccessToken(id);
   }
 
-  async removeAccessToken(id) {
-    const done = await this.redis.removeAccessToken(id);
+  async removeAccessToken(token) {
```
rfk: `removeRefreshToken` needs to know both the `uid` and `tokenId`, and the `/destroy` endpoint generically calls either `removeAccessToken` or `removeRefreshToken` depending on the type of token. To keep that simple, I opted ti make both of them take the full token object rather than just the token id.

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
```diff
@@ -261,7 +266,7 @@ const QUERY_LIST_ACCESS_TOKENS_BY_UID =
   'FROM tokens LEFT OUTER JOIN clients ON clients.id = tokens.clientId ' +
   'WHERE tokens.userId=?';
 const QUERY_LIST_REFRESH_TOKENS_BY_UID =
-  'SELECT refreshTokens.token AS refreshTokenId, refreshTokens.clientId, refreshTokens.createdAt, refreshTokens.lastUsedAt, ' +
+  'SELECT refreshTokens.token AS tokenId, refreshTokens.clientId, refreshTokens.createdAt, refreshTokens.lastUsedAt, ' +
```
rfk: Here and elsewhere, I've updating naming to call this the `tokenId` consistently, in line with the change to name it consistently when working with access tokens.

dannycoates:
This looks good @rfk but I want to hold off the merge until we get a bit more experience running with redis access tokens in prod.

rfk:
Thanks @dannycoates, sounds good; I don't think we have too much of a rush on this any more.

bbangert:
@dannycoates It's been a month, it seems like our redis token use is ok now. To avoid forgetting about this perhaps we should file a spike to look into when we are ready to merge this so we can track it?

bbangert:
@rfk Looks like a rebase is needed.

rfk:
👍 rebased

rfk:
Term, test failure here seems to be some unrelated redis thing which perhaps my change has perturbed out of working, will dig in.

rfk:
For future reference there was a test failure in the verification reminders code here:

  https://github.com/mozilla/fxa/blob/35af35275896a085f569b84622476b1e1908b22d/packages/fxa-auth-server/test/local/verification-reminders.js#L227

It seems to have gone away on re-running the job, so I wonder if that test is non-deterministic and it should e.g. compare the result as a set rather than a list.

bbangert:
Filed FXA-1668 spike to track review @dannycoates recommended which will unblock this PR.

clouserw:
We'd like to land this once we can confirm that the redis instance is stable.  This morning we changed the token limit from 100 to 50, so, this still has some moving parts but we're keeping an eye on it.

dannycoates:
I want to dust this off soon and make a few changes

clouserw:
@dannycoates what's the status of this?

clouserw:
Rough plan:  Danny will rebase and this can land next week

rfk:
A friendly reminder that we'll thank ourselves for having this in production before the Fenix migration begins in earnest. Let me know if there's anything I can do here to help it along.

dannycoates:
moved to #5803

PR #4317, opened 2020-02-28
Don't log out of sync when clearing the signed in account on accounts.firefox.com

### LINKED ISSUE #4293: Navigating to FxA with incorrect UID param signs user out of sync
## STR
1. Have two Firefox accounts (Account A & Account B)
1. Sign into Sync with Account A
1. Grab the uid from Account B (you can do this by signing in in a different browser and opening the network panel and looking at the profile)
1. In the Firefox where you are signed into Sync with Account A go to `accounts.firefox.com/settings?uid=\{uid-for-account-b}

## Expected
User is prompted to sign into Account B

## Actual
1. User is signed out of Sync and Account A and then prompted to sign into Account A.
1. Signing back into Account A here does not restart sync

### Notes
* passing in UID and EMAIL params for Account B resolves this problem.

* passing in a nonsense param `uid=hello` doesn't cause this problem. The UID has to at least have valid formatting

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1210)
┆Issue Number: FXA-1210
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12634/Screen+Shot+2020-02-27+at+9.34.28+PM.png">Screen Shot 2020-02-27 at 9.34.28 PM.png</a>

dannycoates:
fixes #4293 

cc @johngruen @rfk

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/user.js
+++ b/packages/fxa-content-server/app/scripts/models/user.js
@@ -296,15 +296,6 @@ var User = Backbone.Model.extend({
     return this.initAccount({});
   },
 
-  // Used to clear the current account, but keeps the account details
-  clearSignedInAccount() {
-    var uid = this.getSignedInAccount().get('uid');
-    this.clearSignedInAccountUid();
-    this._notifier.triggerRemote(this._notifier.COMMANDS.SIGNED_OUT, {
-      uid: uid,
-    });
-  },
-
   removeAllAccounts() {
     this.clearSignedInAccountUid();
     this._storage.remove('accounts');
@@ -322,7 +313,7 @@ var User = Backbone.Model.extend({
     var account = this.initAccount(accountData);
 
     if (this.isSignedInAccount(account)) {
-      this.clearSignedInAccount();
+      this.clearSignedInAccountUid();
     }
 
     var accounts = this._accounts();
@@ -649,6 +640,7 @@ var User = Backbone.Model.extend({
 
   /**
    * Disconnect an attached client from the given account.
+   * When the client is the current device also sign out of Sync.
    *
    * @param {Object} account - account with the attached client
    * @param {Object} client - AttachedClient model to disconnect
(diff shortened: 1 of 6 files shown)


dannycoates:
@davismtl there's a few changes in behavior with this fix:

1. If you're logged into sync with account A, you won't be logged out when a link directs you to settings for account B. Upon logging into account B, you'll still be logged into sync with account A and you're tab will be logged into account B. If you go to settings from the account menu, you'll see the settings for account A. From what I understand this is the desired effect, i.e. logged into the browser with account A and in a web context in a tab with account B.

2. The only way you'll get logged out of sync is through the browser ui. No web context actions will log you out of sync.

3. This is an edge case but it has a functional test. With two tabs open to settings, logging out will redirect both tabs to a login page with the email preset to the one you just logged out of. Previously it redirected to the email+password form.

Are those changes ok?

chelsea-lewis:
@davismtl Can you comment on this before end of sprint? [Per @dannycoates request on Slack]

davismtl:
> @davismtl there's a few changes in behavior with this fix:
> 
>     1. If you're logged into sync with account A, you won't be logged out when a link directs you to settings for account B. Upon logging into account B, you'll still be logged into sync with account A and you're tab will be logged into account B. If you go to settings from the account menu, you'll see the settings for account A. From what I understand this is the desired effect, i.e. logged into the browser with account A and in a web context in a tab with account B.

This sounds great.

>     2. The only way you'll get logged out of sync is through the browser ui. No web context actions will log you out of sync.

There is 1 exception. In the device management, you could disconnect the local device
image


>     3. This is an edge case but it has a functional test. With two tabs open to settings, logging out will redirect both tabs to a login page with the email preset to the one you just logged out of. Previously it redirected to the email+password form.
> 
> 
> Are those changes ok?

This sounds good to me.

dannycoates:
I've added the case where disconnecting the current device _should_ log you out of the browser

PR #4333, opened 2020-02-28
feat(admin-server): add sentry reporting

### LINKED ISSUE #4242: Sentry reporting for admin server
Add Sentry config option and reporting of errors to admin server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1186)
┆Issue Number: FXA-1186

bbangert:
Because:

* We have no sentry reporting of internal errors.

This commit:

* Captures internal resolver errors to Sentry.

Closes #4242

### DIFF
--- a/packages/fxa-admin-server/src/lib/version.ts
+++ b/packages/fxa-admin-server/src/lib/version.ts
@@ -1,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ * Return version info based on package.json, the git sha, and source repo
+ *
+ * Try to statically determine commitHash and sourceRepo at startup.
+ *
+ * If commitHash cannot be found from `version.json` (i.e., this is not
+ * production or stage), then an attempt will be made to determine commitHash
+ * and sourceRepo dynamically from `git`. If it cannot be found with `git`,
+ * just show 'unknown' for commitHash and sourceRepo.
+ *
+ * This module may be called as ./dist/lib/version.js, or when testing,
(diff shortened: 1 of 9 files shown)


packages/fxa-admin-server/src/test/lib/sentry.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      undefined,
+      undefined,
+      ['resolver', 'field'],
+      originalError
+    );
+    reportGraphQLError(false, logger, err);
+    const errResult = mockCaptureException.args[0][0];
+    assert.equal(errResult.message, 'boom');
+    const logSpy = (logger.error as unknown) as SinonSpy;
+    assert.isTrue(logSpy.calledOnceWith('graphql', { path: 'resolver.field', error: 'boom' }));
```
chenba: Could additionally assert that mockScope.setContext have been called with the expected arguments.

packages/fxa-admin-server/src/lib/version.ts
```diff
@@ -0,0 +1,73 @@
...
+    readJson(path.resolve(__dirname, '..', '..', '..', 'package.json'));
+
+  return {
+    commit,
+    source,
+    version: packageInfo?.version
+  };
+}
+
+export const version = getVersionInfo();
```
chenba: Is this file from somewhere?  I don't see any tests for, but if it's working code from somewhere 🤷‍♂

bbangert: It is, its from event-broker which is from auth-server. Ideally it should be in fxa-shared.

packages/fxa-admin-server/src/test/lib/sentry.spec.ts
bbangert: Added that.

PR #4334, opened 2020-02-28
fix(style): fix rtl styles on Firefox Family Servcies

### LINKED ISSUE #4263: [RTL] Wrong padding for the firefox family services section on the sign up page [bz1617725]
From https://bugzilla.mozilla.org/show_bug.cgi?id=1617725

Created attachment 9128602
Before

See before/after screenshots.

Relevant code to add:
```css
html[dir="rtl"] .sign-up > .firefox-family-services {
  margin: 90px 36px 0 0;
}

html[dir="rtl"] .firefox-family-services > ul > .firefox-service .logo.logo-lockwise {
  background-position-x: 5px;
}

html[dir="rtl"] .firefox-family-services > ul > .firefox-service .logo.logo-monitor {
  background-position-x: 12px;
}

html[dir="rtl"] .firefox-family-services > ul > .firefox-service .logo.logo-send {
  background-position-x: 8px;
}

html[dir="rtl"] .firefox-family-services > ul > .firefox-service {
  padding: 0 0 20px 10px;
}
```

Code to remove, seems to be a leftover:
```css
html[dir="rtl"] .firefox-family-services > h2 {
  padding-right: 10px;
}
```

As a side bonus, add this rule:
```css
html[dir="rtl"] .input-row input:not([type="number"]) {
  direction: ltr;
}
```
... and remove the `direction: ltr` from the `html[dir="rtl"] .input-row input` rule (this is so the question mark of the "How old are you?" field won't appear in reverse).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1197)
┆Issue Number: FXA-1197

chenba:
This patch fixes some rtl styles on the Firefox family services list on
the sign in/up pages.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1617725

Fixes #4263 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_marketing.scss
+++ b/packages/fxa-content-server/app/styles/modules/_marketing.scss
@@ -36,12 +36,6 @@
     font-size: 18px;
     margin-top: 0;
 
-    @include respond-to('big') {
-      html[dir='rtl'] & {
-        padding-right: 10px;
-      }
-    }
-
     @include respond-to('small') {
       font-size: 16px;
 
@@ -50,6 +44,7 @@
       }
 
       html[dir='rtl'] & {
+        padding-right: 10px;
         text-align: right;
       }
     }
@@ -72,6 +67,10 @@
 
       @include respond-to('big') {
         padding: 0 10px 20px 0;
+
+        html[dir='rtl'] & {
+          padding: 0 0 20px 10px;
+        }
       }
 
       html[dir='ltr'] & {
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/styles/modules/_marketing.scss
```diff
@@ -50,6 +44,7 @@
       }
 
       html[dir='rtl'] & {
+        padding-right: 10px;
```
ItielMaN: @chenba Hmm, why this change? I can't see any `padding-left` for ltr...

chenba: This is so the `h2` would align with the list items below; the `li`s have a bottom and right padding of 10px.

ItielMaN: > 
> 
> This is so the `h2` would align with the list items below; the `li`s have a bottom and right padding of 10px.

@chenba Ah I see. Visually this looks fine, but this is not on par with LTR because the list should be aligned with `main-content`, right now there's a 10px padding before as you said.
The correct fix imo would be to remove this rule, and add this block to the `.firefox-service` rule:
```css
html[dir='rtl'] & {
  padding: 0 0 10px 10px;
}
```
Should I file a new bug? If so, should it be on BMO or here?

chenba: Sure.  File it on Github.  Thanks!

PR #4338, opened 2020-02-28
fix(lint): Fix customs server lint issues

vbudhram:
Not connected to an issue but fixes our lint errors in customs server. Most of the changes here are due to prettier formatting the code.

### DIFF
--- a/packages/fxa-customs-server/.eslintrc
+++ b/packages/fxa-customs-server/.eslintrc
@@ -7,6 +7,7 @@
     }
   },
   "rules": {
-    "strict": "off"
+    "strict": "off",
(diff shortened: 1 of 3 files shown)


packages/fxa-customs-server/.eslintrc
```diff
@@ -7,6 +7,7 @@
     }
   },
   "rules": {
-    "strict": "off"
+    "strict": "off",
+    "handle-callback-err": "off"
```
vbudhram: Instead of adding a bunch of 

```
if (err) {
  console.log(err);
}
```

statements, I opted to ignore the warning for only this service.

PR #4341, opened 2020-02-28
fix(signup): change auth code input type to number

### LINKED ISSUE #4281: "Enter auth code" page on Android shows alpha keyboard
When connecting an acct on Fenix, if you bring up the on-screen keyboard on the "enter auth code we just sent you" page, a full alpha keyboard is shown. It should be possible for the HTML to arrange for a numeric-only keyboard to be shown instead.

(I admit that I haven't looked into this very much - it's certainly possible this is actually a bug in Fenix, or I may have missed something else - apologies in advance if that's true)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1202)
┆Issue Number: FXA-1202

chenba:
This patch changes the input type for the sign up/in auth code from text
to number.  This is mainly due to
https://bugzilla.mozilla.org/show_bug.cgi?id=1205133.

Fixes #4281 

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
@@ -15,8 +15,8 @@
         <p class="verification-email-message">{{#unsafeTranslate}}Please enter the verification code that was sent to %(escapedEmail)s within 5 minutes.{{/unsafeTranslate}}</p>
         <form novalidate>
             <div class="input-row">
-                <!-- Using `type="text" inputmode="numeric"` shows the numericpad on mobile and strips out whitespace on desktop. -->
-                <input type="text" inputmode="numeric" pattern="\d[ ]*" class="tooltip-below otp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
+                <!-- Not using `type="text" inputmode="numeric"` because of https://bugzilla.mozilla.org/show_bug.cgi?id=1205133 -->
+                <input type="number" inputmode="numeric" pattern="\d[ ]*" class="tooltip-below otp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
             </div>
             <div class="button-row">
                 <button id="submit-btn" type="submit">{{#t}}Verify{{/t}}</button>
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/templates/confirm_signup_code.mustache
```diff
@@ -15,8 +15,8 @@
         <p class="verification-email-message">{{#unsafeTranslate}}Please enter the verification code that was sent to %(escapedEmail)s within 5 minutes.{{/unsafeTranslate}}</p>
         <form novalidate>
             <div class="input-row">
-                <!-- Using `type="text" inputmode="numeric"` shows the numericpad on mobile and strips out whitespace on desktop. -->
-                <input type="text" inputmode="numeric" pattern="\d[ ]*" class="tooltip-below otp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
+                <!-- Not using `type="text" inputmode="numeric"` because of https://bugzilla.mozilla.org/show_bug.cgi?id=1205133 -->
+                <input type="number" inputmode="numeric" pattern="\d[ ]*" class="tooltip-below otp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
```
vbudhram: With this change empty /whitespace padded codes come up like this

Screen Shot 2020-02-28 at 4 20 22 PM

With original version it would appear as a tooltip on input field.

vbudhram: @ryanfeeley Do we need to support stripping out the whitespace for the input field? Removing that requirement might make this bug a little easier to fix?

chenba: Yeah, the pattern attribute isn't supported on number inputs. 😞   I'm fine with WONTFIX this also, since the real problem is https://bugzilla.mozilla.org/show_bug.cgi?id=1205133

chenba:
Closing this without merging until we can find a better solution or WONTFIX.

chenba:
Updated to fix the validation issue (which was not the problem I thought it was).  Reopened as #4400.

Issue #2601, opened 2019-09-25
Update server to Hapi 18


6a68:
Currently adding .nsprc exceptions to fxa-auth-server and fxa-profile-server to work around a vulnerability in `subtext`, which is fixed, but in the re-namespaced package, `@hapi/subtext`. Picking up the fix requires us to upgrade to hapi 18, which introduced the naming changes.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-373)
┆Issue Number: FXA-373

shane-tomlinson:
cc @vladikoff didn't you put some effort into converting both of these to hapi 18?

chenba:
Hapi 19 is out now.  We also need to switch to @hapi/hawk.

PR #4342, opened 2020-02-28
chore(auth): updated auth-server to latest hapi

dannycoates:
also couldn't help but rm the old redis lib


dannycoates:
closes #2601

Issue #3756, opened 2019-12-24
use pseudolocalization to test translations on payments server


meandavejustice:
need to get myself up to date with https://github.com/projectfluent/fluent.js/issues/83

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-848)
┆Issue Number: FXA-848

data-sync-user:
➤ Wil Clouser commented:

If we do this it should be across all of FxA.  Closing this narrowly scoped issue.

PR #4397, opened 2020-03-03
Fpn mobile updates

meandavejustice:


meandavejustice:
Many of the recent changes from jackie were removed in this pr, leaving a reference to the commit so we can find it later

https://github.com/mozilla/fxa/commit/dbbe8a26ef597c5b4903e6b6ee3c9d833283441e

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -167,22 +173,58 @@ export const Product = ({
...
+            resetCreateSubscription,
+            validatorInitialState,
+          }}
+        />
+        <PlanDetails
+          {...{
+            profile: profile.result,
+            selectedPlan,
+            isMobile,
+            showExpandButton: isMobile,
```
LZoog: I think this is fine to leave as-is, but I realize seeing this PR in its entirety that you could set `showExpandButton` in `PlanDetails` based off of `isMobile` since they're the same value instead of passing in both values every time we need to render `PlanDetails`.

packages/fxa-payments-server/src/routes/Product/index.scss
```diff
@@ -1,79 +1,13 @@
-@import '../../../../fxa-content-server/app/styles/breakpoints';
-@import '../../../../fxa-content-server/app/styles/variables';
+.mobile-subscription-create-heading {
```
LZoog: I think this could just be `.mobile-subscription-heading` without the word `create`, but ¯\_(ツ)_/¯

packages/fxa-payments-server/src/routes/Product/index.tsx
meandavejustice: I though about changing that also, but I think it's more clear having them separate, especially in tests and stories

LZoog: We could still leave them separate but set the variable in `PlanDetails` instead of setting it before composing `PlanDetails` every time. Will leave that up to you though.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+      </div>
+
+      <div className="footer" data-testid="footer">
+        <Localized id="payment-confirmation-download-button">
+          <a
+            data-testid="download-link"
+            className="button download-link"
+            href={productUrl}
+          >
+            click to download
```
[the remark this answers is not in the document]
meandavejustice: I fixed this to match the `payment-confirmation-download-button` id. I say we leave these fallbacks for now and address them in this issue https://github.com/mozilla/fxa/issues/3756

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useState } from 'react';
+import { Localized } from 'fluent-react';
+import { formatCurrencyInCents } from '../../lib/formats';
+import { metadataFromPlan } from '../../store/utils';
+
+// this is a fallback incase webIconURL is undefined,
+// will check with jgruen on whether we have a different
```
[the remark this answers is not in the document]
meandavejustice: asked in slack, we can fix as a follow up if they don't get back today

PR #4399, opened 2020-03-03
fix(auth-server): update default node version for hapi upgrade

meandavejustice:


dannycoates:
We ought to update them all to 12

PR #4404, opened 2020-03-04
feat(admin-panel): Add base fxa-admin-panel package (client) shell

### LINKED ISSUE #4201: [admin panel A1.0] Setup client shell with React app
This will likely be done via `create-react-app`. We'll want to have any other initial setup in place as well which can likely be copied from the `fxa-payments-server` initial commits and some pieces from `fxa-support-panel`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1138)
┆Issue Number: FXA-1138

LZoog:
* Setup React application via `create-react-app`, clear some cruft
* Add small Express server with HSTS + html-middleware/noblocking headers
* Proxy express server to the port where the webpack-dev-server runs (8092)
* Sets up Typescript for src/ and server/
* Setup package.json with initial npm commands including build
* Add prettier, eslint, Jest config for client, Docker

fixes #4201

~I will create a follow-up issue for `server/` tests and adding version/heartbeat~ Follow-up here: #4410


packages/fxa-admin-panel/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "lint:eslint": "eslint .",
+    "build": "npm-run-all build:client build:server",
+    "build:client": "react-scripts build",
+    "build:server": "tsc -p server/tsconfig.json",
+    "test": "react-scripts test --coverage --verbose",
+    "format": "prettier '**' --write",
+    "eject": "react-scripts eject"
+  },
+  "dependencies": {
+    "@types/helmet": "0.0.45",
```
bbangert: nit: types usually only need to be dev dependencies

lmorchard:
Probably want to add this to the pm2 config like payments-server:

https://github.com/mozilla/fxa/blob/master/mysql_servers.json#L168

LZoog:
> Would like to see this added to the pm2 servers before merge, but wouldn't absolutely block on it. Looks mostly like the same pattern as payments-server, except for how gnarly that package's Dockerfile has gotten

@lmorchard And es6 modules in admin-panel! ;) 

Is this something we _want_ to run automatically with `pm2`? This is kind of separate from the FxA stack, most of the time we probably don't need it. 🤔

bbangert:
The admin-server is also not started with pm2 at the moment, for a similar reason.

LZoog:
I moved that `@types` dep to dev deps, good catch @bbangert. 

I made another issue for this pm2 conversation: #4428

PR #4405, opened 2020-03-04
fix(mysql): Force MySQL connections to always use UTC timezone.

rfk:
Because:

* Some of our security checks (such as checking whether a token
  has expired) compare a timestamp generated in JS with a timestamp
  generated in MySQL.
* MySQL returns timestamps in whatever it believes to be the local
  timezone by default.
* We try to be robust to suboptimally-configured MySQL databases,
  especially when it could lead to security issues.

This commit:

* Explicitly tells MySQL to return timestamps in UTC regardless
  of the system- or database-level timezone setting.

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/mysql.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mysql.js
@@ -1192,34 +1192,31 @@ module.exports = function(log, error) {
         // Enforce sane defaults on every new connection.
         // These *should* be set by the database by default, but it's nice
         // to have an additional layer of protection here.
-        connection.query('SELECT @@sql_mode AS mode;', (err, rows) => {
-          if (err) {
-            return reject(err);
-          }
-
-          const currentModes = rows[0]['mode'].split(',');
-          this.requiredModes.forEach(requiredMode => {
-            if (currentModes.indexOf(requiredMode) === -1) {
-              currentModes.push(requiredMode);
-            }
-          });
-
-          const newMode = currentModes.join(',');
-          connection.query(`SET SESSION sql_mode = '${newMode}';`, err => {
-            if (err) {
-              return reject(err);
-            }
-
-            connection.query('SET NAMES utf8mb4 COLLATE utf8mb4_bin;', err => {
-              if (err) {
-                return reject(err);
+        const query = P.promisify(connection.query, { context: connection });
+        return resolve(
+          (async () => {
+            // Always communicate timestamps in UTC.
+            await query("SET time_zone = '+00:00'");
+            // Always use full 4-byte UTF-8 for communicating unicode.
+            await query('SET NAMES utf8mb4 COLLATE utf8mb4_bin;');
+            // Always have certain modes active. The complexity here is to
+            // preserve any extra modes active by default on the server.
+            // We also try to preserve the order of the existing mode flags,
+            // just in case the order has some obscure effect we don't know about.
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1192,34 +1192,29 @@ module.exports = function(log, error) {
...
-            if (err) {
-              return reject(err);
-            }
-
-            connection.query('SET NAMES utf8mb4 COLLATE utf8mb4_bin;', err => {
-              if (err) {
-                return reject(err);
+        const query = P.promisify(connection.query, { context: connection });
+        return resolve(
+          (async () => {
```
rfk: There was no way I was going to add another layer to this pyramid of doom; happy to take suggestions on a nicer way to async-await-ify this code without this turning into a whole-class refactor.

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
```diff
@@ -978,34 +978,37 @@ MysqlStore.prototype = {
...
-            if (modes.indexOf(requiredMode) === -1) {
-              modes.push(requiredMode);
-              needToSetMode = true;
+        const query = P.promisify(conn.query, { context: conn });
+        return resolve(
+          (async () => {
+            // Always communicate timestamps in UTC.
+            await query("SET time_zone = '+00:00'");
+            // Always use full 4-byte UTF-8 for communicating unicode.
+            await query('SET NAMES utf8mb4 COLLATE utf8mb4_unicode_ci;');
```
rfk: The oauth-server was previously not doing an explicit `SET NAMES` during connection setup, but both auth-server and profile-server were. I've added it here for consistency.

Well...I say "consistency" but it actually uses a different collation order than the other two, for what I assume are really fascinating historical reasons. See `REQUIRED_CHARSET` at the top of the file.

packages/fxa-profile-server/lib/db/mysql/index.js
```diff
@@ -240,38 +240,33 @@ MysqlStore.prototype = {
...
-              conn._fxa_initialized = true;
-              return resolve(conn);
-            });
-          });
-        });
+            // Avoid repeating all that work for existing connections.
+            conn._fxa_initialized = true;
+            return conn;
+          })()
+        );
```
rfk: Love to copy-paste this multiple times, but I wasn't prepared to attempt a bigger refactoring of the MySQL code into fxa-shared at this stage.

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -1192,34 +1192,29 @@ module.exports = function(log, error) {
...
+              if (currentModes.indexOf(requiredMode) === -1) {
+                currentModes.push(requiredMode);
               }
-
-              connection._fxa_initialized = true;
-              resolve(connection);
-            });
-          });
-        });
+            }
```
chenba: IIUC, I suppose this could be written as `currentModes = [...new Set([...currentModes, ...this.requiredModes])]`

rfk: Thanks @chenba!

I believe the current approach deliberately tries to preserve the order of the mode flags although I don't know why and it's probably just from not quite trusting MySQL not to have some nonsense ordering-related problem. I'll add a comment to this effect in the source.

PR #4408, opened 2020-03-04
fix(oauth): Do not store raw access tokens in Redis.

rfk:
The MySQL access-token storage deliberately does not write the raw access token to disk, and instead only writes a hash. This allows us to look up a given token but prevents a db compromise from immediately leaking all tokens in useable form.

The new Redis access-token storage currently includes the raw token in its JSON serialization. We should never need this value again in practice, so I suggest we avoid storing it for all the same reasons we don't store it in MySQL. This commit changes the serialization logic to persist `tokenId` instead, while continuing to support data with the old field that may already be in redis.

Relatedly, the value returned when looking up an access token could differ depending on whether it came from Redis (in which case the `token` field would be the raw token) or from MySQL (in which case the `token` field would be the hashed token id). I've included a bit of a refactor of how the `AccessToken` class is used, so that these differences are encapsulated in the DB layer and calling code will only ever see instances of the `AccessToken` class. (This is partly motivated by making the above storage change, and partly by my own confusion on this point while working on #4304).

I've also included a bonus commit that removes some leftover mentions of the `./fxa-oauth-server` directory, because my old habit of running `npm run test-oauth` gave an error.

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/db/accessToken.js
+++ b/packages/fxa-auth-server/lib/oauth/db/accessToken.js
@@ -10,6 +10,7 @@ const ScopeSet = require('../../../../fxa-shared').oauth.scopes;
 
 class AccessToken {
   constructor(
+    tokenId,
     clientId,
     name,
     canGrant,
@@ -17,12 +18,13 @@ class AccessToken {
     userId,
     email,
     scope,
-    token,
     createdAt,
     profileChangedAt,
     expiresAt,
-    ttl
+    token
   ) {
+    /** @type {Buffer} */
+    this.tokenId = tokenId;
     /** @type {Buffer} */
     this.clientId = clientId;
     /** @type {string} */
@@ -37,17 +39,14 @@ class AccessToken {
     this.email = email;
     /** @type {ScopeSet} */
     this.scope = scope;
-    /** @type {Buffer} */
-    this.token = token || unique.token();
-    /** @type {Date} truncated to the second to match mysql */
-    this.createdAt = createdAt || new Date(new Date().setMilliseconds(0));
-    /** @type {number} */
-    this.profileChangedAt = profileChangedAt || 0;
     /** @type {Date} */
-    this.expiresAt =
-      expiresAt || new Date(Date.now() + (ttl * 1000 || MAX_TTL));
-    /** @type {Buffer} */
-    this.tokenId = encrypt.hash(this.token);
+    this.createdAt = createdAt;
+    /** @type {number} */
+    this.profileChangedAt = profileChangedAt;
+    /** @type {Date} */
+    this.expiresAt = expiresAt;
+    /** @type {Buffer} this won't be proided unless we're creating a brand new token */
+    this.token = token || null;
     this.type = 'bearer';
   }
 
@@ -67,6 +66,8 @@ class AccessToken {
     return this.createdAt;
   }
 
+  // This is not the tokenId, it's the clientId, and it's only here
+  // for compatbility with the data format expected by `aggregateActiveClients`.
   get id() {
     return this.clientId;
   }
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/lib/oauth/db/mysql/index.js
```diff
@@ -181,7 +182,16 @@ const QUERY_ACCESS_TOKEN_INSERT =
...
+// Access token storage in redis annotates each token with client metadata,
+// so we do the same when reading from MySQL, for consistency.
+const QUERY_ACCESS_TOKEN_FIND =
+  'SELECT tokens.token, tokens.clientId, tokens.createdAt, ' +
+  '  tokens.userId, tokens.email, tokens.scope, ' +
+  '  tokens.createdAt, tokens.expiresAt, tokens.profileChangedAt, ' +
+  '  clients.name as clientName, clients.canGrant AS clientCanGrant, ' +
+  '  clients.publicClient ' +
+  'FROM tokens LEFT OUTER JOIN clients ON clients.id = tokens.clientId ' +
+  'WHERE tokens.token=?';
```
rfk: This is a more expensive query, but we'll only do it for a small number of tokens, so IMHO the increase in internal consistency is worth it. I won't be upset if you push back though, and we could instead default the client-related fields to `null`.

packages/fxa-auth-server/lib/oauth/db/accessToken.js
```diff
@@ -87,31 +89,92 @@ class AccessToken {
...
+
+  /**
+   * @param {any} row
+   * @returns {AccessToken}
+   */
+  static fromMySQL(row) {
+    return new AccessToken(
+      // Somewhat confusingly, the `token` column in MySQL is the hashed value
+      // that we refer to as `tokenId`.
+      row.token,
```
dannycoates: what if the query used `tokens.token AS tokenId` instead?

packages/fxa-auth-server/lib/oauth/db/mysql/index.js
dannycoates: LGTM. clients is a tiny table

packages/fxa-auth-server/lib/oauth/db/accessToken.js
```diff
@@ -87,31 +89,92 @@ class AccessToken {
...
+      publicClient,
+      userId,
+      email,
+      scope,
+      // Truncated createdAt to the second to match mysql
+      new Date(new Date().setMilliseconds(0)),
+      profileChangedAt || 0,
+      expiresAt || new Date(Date.now() + (ttl * 1000 || MAX_TTL)),
+      // This is the one and only time the caller can get at the unhashed token.
+      token
```
dannycoates: Not that it matters here and now, but I half wonder, in a more strongly typed future if we should have a `FreshAccessToken` subtype with the token while the base class doesn't.

rfk:
> I'm a little embarrassed I didn't catch the intent of not storing the token in the db. 

Heh, well, it's not like it's written down anywhere! :-/

I'll take an action to document this inline in the MySQL schema definition as well, before merging this PR.

packages/fxa-auth-server/lib/oauth/db/accessToken.js
rfk: I was inches away from returning a two-tuple of `(raw token bytes, access token object)` from this function instead, for similar reasons.

lmorchard:
> > I'm a little embarrassed I didn't catch the intent of not storing the token in the db.

>Heh, well, it's not like it's written down anywhere! :-/

*/me swings by on a vine*

This would make for [a good ADR](https://github.com/mozilla/fxa/tree/master/docs/adr)!

packages/fxa-auth-server/lib/oauth/db/accessToken.js
rfk: 👍 I'm going to this, and use it as an opportunity to document the *why* as well.

PR #4412, opened 2020-03-04
fix(settings): Don't show signout when oauth clients that request oldsync scope

### LINKED ISSUE #4284: [iOS, FxA Rust] FxA settings web view now showing a 'sign out' button
Suddenly seeing a "Sign Out" button, which I am not sure is intentional, as it doesn't exist in release Firefox iOS. I suspect there was a past UX decision to not show this deliberately.

image

Expected Behaviour: "Sign Out" button is hidden on Firefox iOS

x-ref to https://github.com/mozilla-mobile/firefox-ios/issues/6178

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1205)
┆Issue Number: FXA-1205

vbudhram:
Fixes #4284 

I went back and forth on a decent way to do this, but settled on the simple approach verses doing a larger refactor. This PR checks to see if the old sync scope is requested for the oauth client, if so then the `Sign out` button will not appear in the settings view. The expectation is that the client will have its own sign out button.

I considered checking the entrypoint, for ios settings it is `?entrypoint=ios_settings_manage` but that seemed more likely to change so stuck with checking scope.

Note that I wasn't actually able to test this locally because it requires application services library to build built and run against local servers. However, I feel good about the tests so it should work as expected.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/settings.js
@@ -93,7 +93,12 @@ const View = BaseView.extend({
       ccExpired: !!this._ccExpired,
       escapedCcExpiredLinkAttrs: 'href="/subscriptions" class="alert-link"',
       securityEventsVisible: this.displaySecurityEvents(),
-      showSignOut: !account.isFromSync(),
+      showSignOut:
+        // Note: For sync clients and oauth based sync clients, they have their
+        // own UX for signing a user out, typically within their own app.
+        // In these scenarios we don't show the sign out button.
+        !account.isFromSync() &&
+        !(this.relier.isOAuth() && this.relier.containsOldSyncScope()),
       unsafeHeaderHTML: this._getHeaderHTML(account),
     });
   },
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -93,7 +93,9 @@ const View = BaseView.extend({
       ccExpired: !!this._ccExpired,
       escapedCcExpiredLinkAttrs: 'href="/subscriptions" class="alert-link"',
       securityEventsVisible: this.displaySecurityEvents(),
-      showSignOut: !account.isFromSync(),
+      showSignOut:
+        !account.isFromSync() &&
+        !(this.relier.isOAuth() && this.relier.containsOldSyncScope()),
```
dannycoates: can you leave a code comment on why this was added? The new logic is a bit cryptic to me as is.

vbudhram:
@dannycoates Thanks!

garvankeeley:
When can we test this on the client-side (Firefox iOS), the button is still appearing ATM.

vbudhram:
@garvankeeley Turns out the fix here didn't really work, I submitted a simpler PR at https://github.com/mozilla/fxa/pull/4510 that should fully do the trick until we can get a better solution.

Issue #4343, opened 2020-02-28
Local services


6a68:
Problem: New contributors who might have redis, memcached, and/or mysql running in the background won't be able to successfully start the monorepo, but also won't get a clear error message when they `npm start`.

Simple fix: we could address this by mentioning it in the docs.

Fancier fix: we could add some cross-platform checks to ensure those services aren't running, and log a helpful error if they are.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1243)
┆Issue Number: FXA-1243

data-sync-user:
➤ Wil Clouser commented:

Let's hardcode some ports in the startup script

PR #4419, opened 2020-03-04
Check for local services on startup

jodyheavener:
Closes https://github.com/mozilla/fxa/issues/4343

1. Updates README setup instructions to note that you shouldn't be running any services yourself.
2. Modifies root `npm start` to perform a pre-check on any ports our servers use. Doesn't abort, but warns you if anything is already running on those ports.

Screen Shot 2020-03-04 at 4 04 44 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ The Firefox Accounts (fxa) monorepo
    npm start
    ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,6 +22,8 @@ The Firefox Accounts (fxa) monorepo
    npm start
    ```
 
+Note this starts up all required services, including Redis, MySQL, and Memcached. You should not be running them yourself, or occupying any of the [server ports](https://github.com/mozilla/fxa/blob/master/mysql_servers.json). Doing so may result in errors.
```
vbudhram: > You should not be running them yourself

nit: I would probably say something like it is recommended that you don't run these services yourself.

Issue #4390, opened 2020-03-03
Limit email header size to less than 998 characters


rfk:
When sending a password reset email, we embed a lot of contextual information into the password-reset link, and we include that full URL of the link in the `X-Link` email header as an aid for debugging/testing.  This URL can be arbitrarily long, e.g. if you try resetting your password from [this signin page](https://accounts.firefox.com/reset_password?utm_campaign=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&entrypoint=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&utm_medium=utm_campaign=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&entrypoint_variantion=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&utm_source=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&utm_term=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) it will generate a password-reset link that's well over 1000 characters long.

That's a problem for the `X-Link` header. [RFC5322 says that](https://tools.ietf.org/html/rfc5322#section-2.1.1):

```
   There are two limits that this specification places on the number of
   characters in a line.  Each line of characters MUST be no more than
   998 characters, and SHOULD be no more than 78 characters, excluding
   the CRLF.

   The 998 character limit is due to limitations in many implementations
   that send, receive, or store IMF messages which simply cannot handle
   more than 998 characters on a line.  Receiving implementations would
   do well to handle an arbitrarily large number of characters in a line
   for robustness sake.  However, there are so many implementations that
   (in compliance with the transport requirements of [RFC5321]) do not
   accept messages containing more than 1000 characters including the CR
   and LF per line, it is important for implementations not to create
   such messages.
```

It is possible for emails to contain long header values by [folding them on whitespace](https://tools.ietf.org/html/rfc5322#section-2.2.3) to get around the line-length limit, but in our case, the header does not contain any whitespace at which to fold.

Many email servers can handle lines longer than 1000 characters just fine (I was able to test this with my fastmail address and it had no problems with the gigantic header) but there are almost certainly some that can't. I suspect this is the cause of deliverability problems in e.g. Bug 1618831.

[...]

PR #4420, opened 2020-03-04
Enforce a max header size of 998 characters in emails

jodyheavener:
Closes https://github.com/mozilla/fxa/issues/4390

Enforces a maximum header size of 998 characters per [RFC 5322](https://tools.ietf.org/html/rfc5322#section-2.1.1) recommendation. This is a hopeful solution to a failed mail delivery noted in [Bugzilla 1618831](https://bugzilla.mozilla.org/show_bug.cgi?id=1618831).

Note this is 998 characters total (key + value).

### DIFF
--- a/packages/fxa-email-service/src/types/headers/mod.rs
+++ b/packages/fxa-email-service/src/types/headers/mod.rs
@@ -28,6 +28,10 @@ use hyperx::{
     header::{parsing::from_one_raw_str, Formatter, Raw as RawHeader},
 };
 
+// No header's total length should be greater than 998 characters
+// See: https://tools.ietf.org/html/rfc5322#section-2.1.1
+const HEADER_MAX_LENGTH: usize = 998;
+
 macro_rules! custom_header {
     ($struct_name:ident, $header_name:expr) => {
         #[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


packages/fxa-email-service/src/types/headers/mod.rs
```diff
@@ -28,6 +28,8 @@ use hyperx::{
     header::{parsing::from_one_raw_str, Formatter, Raw as RawHeader},
 };
 
+const HEADER_MAX_LENGTH: usize = 998;
```
vbudhram: I think a comment here would be helpful for future us on why this exists.

packages/fxa-email-service/src/types/headers/test.rs
```diff
@@ -138,3 +140,17 @@ fn verify_code() {
     assert_eq!(header.to_string(), "wibble");
     assert_eq!(VerifyCode::header_name(), "X-Verify-Code");
 }
+
+#[test]
+fn any_header_length() {
+    let mut rng = rand::thread_rng();
+
+    let header = Link::new(std::iter::repeat("X").take(rng.gen_range(0, 2000)).collect::<String>().to_owned());
```
dannycoates: idk that this needs to be a random length string, just one over the max.

PR #4422, opened 2020-03-04
fix(auth-server): pass all redis set args

### LINKED ISSUE #4413: Change redis.set call to carry all arguments
The `redis.set` call in our redis file in auth-server only carries the key and value to set. The actual redis set command takes 4 values though, which we assumed would be carried through from our use in `stripe.js`. To fix this, we should update the set call in our `redis.js` to carry the additional two args through.

Closing this bug will require the deploy notes to be updated to note that these two keys (set from stripe.js) should be deleted **after** deployment, so that they can be re-populated with a TTL.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1289)
┆Issue Number: FXA-1289

bbangert:
Because:

* We originally didn't use additional redis.set arguments, our set
  pass-through function dropped anything other than key/value.

This commit:

* Retains all redis.set arguments supported by the ioredis library we
  use.

Fixes #4413

### DIFF
--- a/packages/fxa-auth-server/lib/redis.js
+++ b/packages/fxa-auth-server/lib/redis.js
@@ -175,8 +175,8 @@ class FxaRedis {
     return this.redis.get(key);
   }
 
-  set(key, val) {
-    return this.redis.set(key, val);
+  set(key, val, ...args) {
+    return this.redis.set(key, val, ...args);
   }
   zadd(key, ...args) {
     return this.redis.zadd(key, ...args);
--- a/packages/fxa-auth-server/package-lock.json
+++ b/packages/fxa-auth-server/package-lock.json
@@ -897,6 +897,15 @@
         "@types/shot": "*"
       }
     },
+    "@types/ioredis": {
+      "version": "4.14.8",
+      "resolved": "https://registry.npmjs.org/@types/ioredis/-/ioredis-4.14.8.tgz",
+      "integrity": "sha512-ww38Psb0jj/8h9UYjRuT9RYF46E5O3j+03XSEBx6OQvCIkgPV52kF8DshVLh8DMSZs3fEjeoEzmyM3KP9n53tQ==",
+      "dev": true,
+      "requires": {
+        "@types/node": "*"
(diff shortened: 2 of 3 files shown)


packages/fxa-auth-server/lib/redis.js
```diff
@@ -175,8 +175,8 @@ class FxaRedis {
     return this.redis.get(key);
   }
 
-  set(key, val) {
-    return this.redis.set(key, val);
+  set(key, val, expiryMode, time, setMode) {
+    return this.redis.set(key, val, expiryMode, time, setMode);
```
dannycoates: I'm not sure this will work when not all the args are passed. In the node console i get this error with `set('y','z')`:
```
ReplyError: ERR syntax error
    at parseError (/Users/dcoates/code/fxa/packages/fxa-auth-server/node_modules/ioredis/node_modules/redis-parser/lib/parser.js:179:12)
    at parseType (/Users/dcoates/code/fxa/packages/fxa-auth-server/node_modules/ioredis/node_modules/redis-parser/lib/parser.js:302:14) {
  command: { name: 'set', args: [ 'y', 'z', '', '', '' ] }
```

lmorchard: Hmm, yeah, might want some default values in there

bbangert: Ah yes, added defaults.

dannycoates:
```js
this.redis.set(...Array.from(arguments))
```

ought to work

dannycoates: or maybe just `...arguments`... i always forget how `arguments` works

bbangert: Well, `...arguments` makes the tests happy.

bbangert: I'm going to use the same format as the zadd below, and capture them as ...args for passthrough.

PR #4423, opened 2020-03-04
3929 subscription confirmation flow

meandavejustice:
Trying to open this up early to catch anything crazy, hope to have this in pixel perfect state at all mobile sizes by later this evening so we can make the merge at same time as train

- adding payment form components together into Product/index.tsx view
- adding cpayment onfirmation view components

TODO:
- ~~breakpoints for mobile~~
- ~~sub heading color in subscription redirect component~~
- new color variables in css (I think it makes more sense to address this separately when we update all of our color variables, thoughts?) @LZoog 
- ~~css cleanup (making sure we delete old rules in redirect component sass file~~
- get unit tests passing (2 tests aren't passing, need help from les on tracking down the issue, waiting to hear back friday morning)
- ~~make sure storybook is all rendering correctly~~


packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -1,87 +1,62 @@
...
-export const SubscriptionRedirect = ({ plan }: SubscriptionRedirectProps) => {
+export const SubscriptionRedirect = ({
+  customer,
+  plan,
+  profile,
+}: SubscriptionRedirectProps) => {
   const { product_id } = plan;
   const { downloadURL } = metadataFromPlan(plan);
   const {
-    config: {
```
LZoog: I think @johngruen had mentioned he'd like an option to easily turn the survey on/off instead of stripping out this code. Maybe it can become its own component that we conditionally render?

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -74,20 +81,20 @@ export const PaymentConfirmation = ({
...
+          <Localized id="payment-confirmation-cc-preview" $last4={last4}>
             <p className={`c-card ${brand.toLowerCase()}`}></p>
           </Localized>
         </div>
       </div>
 
       <div className="footer">
         <Localized id="payment-confirmation-download-button">
-          <a>click to download</a>
+          <a href={redirectUrl}>click to download</a>
```
LZoog: Are we going to be automatically redirecting users to this URL? If not, maybe we call it `downloadUrl`?

packages/fxa-payments-server/src/index.scss
```diff
@@ -34,7 +34,129 @@ h3.billing-title {
...
+.main-content {
+  display: flex;
+  padding-top: 0;
+
+  .input-row {
+    // Stripe element class names
+    // https://stripe.com/docs/stripe-js/reference#the-element-container
+
+    .StripeElement {
+      @include input-element();
```
LZoog: Huh, I thought we had trouble styling the form due to the iframe. Check out what we're [doing here](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L276), at the very least we should update `SMALL_DEVICE_RULE` there. Does this mean we could potentially remove `mkStripeElementStyles`? Would need to do more digging, I'll leave this thought here though since this is a draft PR🤔

packages/fxa-payments-server/src/index.scss
```diff
@@ -34,7 +34,129 @@ h3.billing-title {
...
+      }
+    }
+  }
+
+  .input-row-group {
+    align-items: flex-end;
+    display: flex;
+    flex-direction: row;
+
+    @include respond-to('simpleSmall') {
```
LZoog: Old mixin! Same as line 145 (unless this will be addressed in that cleanup issue we made a few days ago)

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
```diff
@@ -1,87 +1,62 @@
...
 import { metadataFromPlan } from '../../../store/utils';
+import { SelectorReturns } from '../../../store/selectors';
+
 import './index.scss';
+import PlanDetails from '../../../components/PlanDetails';
+import PaymentConfirmation from '../../../components/PaymentConfirmation';
+import Header from '../../../components/Header';
+
+const isMobile = window.matchMedia(
+  '(max-width: 520px), (orientation: landscape) and (max-width: 640px)'
```
LZoog: Will this also be updated with the new breakpoints?

We'll also need a debounced window resize event listener somewhere... A quick Google search [gave me this page](https://hackernoon.com/simplifying-responsive-layouts-with-react-hooks-19db73893a7a), I didn't read it in-depth but it seems like it could be a good direction?

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
meandavejustice: Is this in place of the confirmation view?

packages/fxa-payments-server/src/index.scss
meandavejustice: I'd like to avoid refactors like this, maybe we can file a follow up issue. This was just moved from the parent id element to a class

LZoog: Yeah, I wasn't thinking you'd address that in this PR, but might be worth looking into as a new issue.

packages/fxa-payments-server/src/index.scss
meandavejustice: I believe these breakpoint replacements are outside the scope of these updates. This change converting and id to a classname. I think we should only add the new breakpoint style and update the old breakpoints as tech debt issues. Since rewriting all of this as mobile first can turn into a huge effort

LZoog: Some of the old breakpoints are mobile-first but yeah, not all of them. 😓 I had messaged with @johngruen a few weeks ago about using those new breakpoints in the payments server for this project since there's a ton of CSS changes and he said it sounded like a good idea but if that feels like a high LoE / out of scope then we can do it in a follow-up. We'll have to comb through payments CSS again anyway when we set a standard for styling conventions (I keep waiting on seeing the settings redesign visual designs before proposing something). I think it's probably a good idea to make it consistent either way, if you've used `@include mi-width` anywhere we probably want to change it to `respond-to` and an older breakpoint.

meandavejustice: created follow up issue in jira

packages/fxa-payments-server/src/routes/Product/SubscriptionRedirect/index.tsx
LZoog: @meandavejustice Right, but I'm not sure if it would be displayed alongside it or somewhere else in the future. We can create a follow-up issue to turn this into a component for future reuse when we want to turn it back "on" with a link to the commit where it's removed and do it later, I think that's fine.

meandavejustice: https://jira.mozilla.com/browse/FXA-1320

packages/fxa-payments-server/src/components/PlanDetails/index.scss
```diff
@@ -1,14 +1,23 @@
...
   width: 342px;
 
+  @include min-width('tablet') {
+    margin: 80px 24px;
+  }
+
+  .plan-details-component-card {
+    background: $color-white;
+    border-radius: 8px;
+    box-shadow: $card-box-high;
```
LZoog: I don't believe this has a box-shadow from the comps.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -27,67 +27,85 @@ export const PlanDetails = ({
...
+              </Localized>
+            )}
+          </div>
+        ) : null}
       </div>
-      {!detailsHidden ? (
-        <div className="plan-details-list" data-testid="list">
-          <Localized id="plan-details-header">
-            <h4>Product details</h4>
+      {!showExpandButton ? (
```
LZoog: I think there might be some logic mixed around here, I'm seeing "Show details" on desktop and no "show details" on mobile, and the `Total: $/month` is showing in mobile.

packages/fxa-payments-server/src/index.scss
```diff
@@ -34,7 +34,134 @@ h3.billing-title {
   }
 }
 
-#main-content {
+// starting to replace the main-content id with this class
```
LZoog: We can probably remove this comment

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.scss
```diff
@@ -1,10 +1,30 @@
...
+  max-width: 600px;
+  min-width: 375px;
+  padding: 28px 16px 60px;
+
+  @include min-width('tablet') {
+    padding: 28px 48px 48px;
+  }
+
+  @include min-width('desktop') {
+    margin: 30px 24px;
```
LZoog: You see how ~ beautiful ~ it is going from mobile to tablet to desktop like this? 😇😂

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.scss
```diff
@@ -1,10 +1,30 @@
 @import '../../../../../fxa-content-server/app/styles/variables';
+@import '../../../../../fxa-content-server/app/styles/breakpoints';
 
 .product-payment {
+  background: #fff;
```
LZoog: I agree we can wait on adding new photon colors to `_variables`, but we can still add ones like `$color-white` or wherever else we have colors set that are already defined (even if just to make the SCSS linter be quiet).

packages/fxa-payments-server/src/routes/Product/SubscriptionSuccess/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  profile,
+}: SubscriptionSuccessProps) => {
+  const { product_id } = plan;
+  const { downloadURL } = metadataFromPlan(plan);
+  const {
+    config: { productRedirectURLs },
+    matchMedia,
+  } = useContext(AppContext);
+
+  const isMobile = matchMedia('(min-width: 768px)');
```
LZoog: We still need a resize event here to update the state when going from one to the other. I left a link here that may be helpful: https://github.com/mozilla/fxa/pull/4423#discussion_r388035310

packages/fxa-payments-server/src/components/PaymentForm/index.scss
```diff
@@ -52,7 +52,9 @@ form.payment {
   }
 
   .terms {
+    display: flex;
     font-size: 15px;
+    justify-content: center;
     padding-top: 56px;
```
LZoog: Clicking “Terms of service” clickable area is very large due to using `padding`. You can use `margin` instead, and make sure it’s not clickable to the left/right of the link (it probably needs to be inline or inline-flex; you may need a span or a container)

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
meandavejustice: This was an issue with the isMobile logic, fixed up and about to push

packages/fxa-payments-server/src/lib/hooks.tsx
```diff
@@ -67,3 +69,21 @@ export function useClickOutsideEffect<T>(onClickOutside: Function) {
 
   return insideEl;
 }
+
+export function useMatchMedia(mediaQuery: string, matchMedia: Function) {
```
LZoog: A test for this would be awesome if possible. Works really well. 😁

packages/fxa-payments-server/src/components/PlanDetails/index.scss
```diff
@@ -1,14 +1,23 @@
...
-  background: $color-white;
-  border-radius: 8px;
-  box-shadow: $card-box-high;
   font-size: 15px;
-  margin: 0 auto;
-  padding: 20px 20px 0;
+  margin: 65px auto 0;
   width: 342px;
 
+  @include min-width('tablet') {
```
LZoog: This is more of a BTW comment I could have added in another review: mixing the old breakpoints (`respond-to`) with the newer breakpoints from `min-width` might produce strange results, ideally we use one set or the other set across the board IMO, but if things are looking good at major breakpoints then this seems OK 👍

Issue #4369, opened 2020-03-02
”Account verified successfully” incorrect message when finishing/canceling to generate a recovery key


vasilicatamas:
**Note:**
This is a follow-up up for #2653

**Build:**
Stage [Train 1.160.0]

**Affected platforms:**
Windows 10

**Steps to reproduce:**
1. Create and confirm a new Firefox account.
2. Go to the email inbox and open the "Get a recovery key" email.
3. Follow the account creation flow.
4. Click on “Start browsing” at the “Your sync data is protected” confirmation screen. 

**Expected results:**
The user is redirected to Manage Account page and the following confirmation message is displayed “Account recovery enabled”.

**Actual Results:**
The user is redirected to Manage account page and the ”Account verified successfully” message is displayed.

**Additional notes:**
This issue also reproduces when canceling the process by clicking on “Maybe later” buttons displayed in “Protect your data with a recovery key” and “Confirm password” screens.

![incorrectmessage](https://user-images.githubusercontent.com/18114210/75678091-7de0fa00-5c95-11ea-86ad-92a8dec981f5.gif)

![incorrectmessage1](https://user-images.githubusercontent.com/18114210/75678101-82a5ae00-5c95-11ea-9a91-1e28867c96f1.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1265)
┆Issue Number: FXA-1265

vasilicatamas:
This issue is verified as fixed on FxA Prod [1.162.1] under Windows 10 64-bit and Android 9. "Account recovery enabled" successful message is displayed after finishing the recovery key flow and no message is displayed when canceling the setup.

![Screenshot_4](https://user-images.githubusercontent.com/18114210/77087029-3d121080-6a0b-11ea-9615-c858387d3184.png)

PR #4427, opened 2020-03-05
fix(account-recovery): Only display recovery confirmation if you're coming from the confirmation page

jodyheavener:
Closes https://github.com/mozilla/fxa/issues/4369

- Only display "Account recovery enabled" message if you're coming from the final step of the post-verify recovery key setup flow.
- Otherwise, don't display any success message, including from "Maybe later" clicks, unless one is provided via `options.success`.

Additionally, beef up tests with

- Assert that the success and error messages were shown with a specific message
- Refute that success and error messages were shown at all

---

@vbudhram this is one of my first stabs at this stack, so I am 100% cool with if this isn't the right approach, or if my assumptions were off.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/behaviors/settings.js
+++ b/packages/fxa-content-server/app/scripts/views/behaviors/settings.js
@@ -26,7 +26,9 @@ export default function(defaultBehavior, options = {}) {
         let success = t('Account verified successfully');
         let endpoint = 'settings';
 
-        if (options.success) {
+        // Strict comparison because undefined == null, but undefined !== null, and
+        // we want to be able to intentionally unset this value via options.success.
+        if (options.success !== null) {
           success = options.success;
         }
 
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js
@@ -308,6 +308,30 @@ const BaseAuthenticationBroker = Backbone.Model.extend({
   },
 
   /**
+   * Called after the post-verify recovery setup is complete.
+   *
+   * @param {Object} account
+   * @return {Promise}
+   */
+  afterCompleteAddPostVerifyRecovery(account) {
+    return this.unpersistVerificationData(account).then(() =>
+      this.getBehavior('afterCompleteAddPostVerifyRecovery')
+    );
+  },
+
+  /**
+   * Called after the post-verify recovery setup is aborted, via "Maybe later" links.
+   *
+   * @param {Object} account
+   * @return {Promise}
+   */
+  afterAbortAddPostVerifyRecovery(account) {
+    return this.unpersistVerificationData(account).then(() =>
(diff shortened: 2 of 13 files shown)


packages/fxa-content-server/app/scripts/views/behaviors/settings.js
```diff
@@ -23,8 +23,12 @@ export default function(defaultBehavior, options = {}) {
   const behavior = function(view, account) {
     return account.isSignedIn().then(isSignedIn => {
       if (isSignedIn) {
-        let success = t('Account verified successfully');
         let endpoint = 'settings';
+        let success;
+
+        if (view.type === 'recovery_key') {
```
vbudhram: Ideally, we would be able to pass in the `options.success` from when we navigate to the page with the [invokeBrokerMethod](https://github.com/mozilla/fxa/blob/78cbb8cd5bee464262a965b64213a5f2d7cf1a1e/packages/fxa-content-server/app/scripts/views/post_verify/verified.js#L57). However, we would have to add a few of them and that might be more complexity than it is worth 🤷🏽‍♂️

packages/fxa-content-server/app/scripts/views/behaviors/settings.js
```diff
@@ -23,8 +23,12 @@ export default function(defaultBehavior, options = {}) {
   const behavior = function(view, account) {
     return account.isSignedIn().then(isSignedIn => {
       if (isSignedIn) {
-        let success = t('Account verified successfully');
         let endpoint = 'settings';
+        let success;
```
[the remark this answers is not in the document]
jodyheavener: Using this as a default will yield this message when you click "Maybe later" in the initial or password confirmation steps unless we set up an additional check to unset the message. That's easy enough to do in the fashion similar to how I currently have it set up, should I proceed with that?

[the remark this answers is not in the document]
vbudhram: Ah I see. Thinking about this a bit more, I think we should add the updated broker methods [here](https://github.com/mozilla/fxa/blob/3173773c00c9d777b9c82d76be10f120e79af7b2/packages/fxa-content-server/app/scripts/models/auth_brokers/web.js#L23) similar to how password reset gets a new message.

packages/fxa-content-server/app/scripts/views/behaviors/settings.js
```diff
@@ -26,7 +26,7 @@ export default function(defaultBehavior, options = {}) {
         let success = t('Account verified successfully');
         let endpoint = 'settings';
 
-        if (options.success) {
+        if (options.success !== null) {
```
jodyheavener: Strict comparison because `undefined == null`, but `undefined !== null`, and we want to be able to intentionally unset this value via `options.success`.

jodyheavener: Strict comparison because `undefined == null`, but `undefined !== null`, and we want to be able to intentionally unset this value via `options.success`.

PR #4438, opened 2020-03-05
bug(auth-server): add payment_intent check in handling open invoices

### LINKED ISSUE #4437: Add Check for Payment Intent loading
We are experiencing a lot of errors, when users are attempting to pay an open invoice. A potential cause for unwarrented errors in this flow is that the payment_intent object is not loaded, which would result in an error state.

Acceptance Criteria:
Add check for payment_intent being loaded when attempting to handle an open invoice.
If payment_intent is not loaded, fetch it from Stripe

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1304)
┆Issue Number: FXA-1304

jackiemunroe:
- add function to ensure that the payment intent is loaded prior to evaulating the status

because:
- users have been reporting errors when attempting to signup for a subscription when their initial payment failed. after verifying that the stripe payment_intent objects appear to be in correct "payment_method_required" state, a potentialcause is that the payment_intent object is not actually loaded. this adds a StripeHelper function to ensure we are working with an object and not a string.

fixes #4437

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -721,6 +721,19 @@ class StripeHelper {
   }
 
   /**
+   * Retrieve a PaymentIntent from an invoice
+   *
+   * @param {Invoice} invoice
+   * @returns {Promise<PaymentIntent>}
+   */
+  async fetchPaymentIntentFromInvoice(invoice) {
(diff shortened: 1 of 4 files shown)


6a68:
Huh, auth-server tests failed with a redis timeout. Retriggered.

PR #4441, opened 2020-03-05
Add example launch configuration for VSCode

eoger:
I haven't been hacking on fxa for a while, the good news is that the initial setup got way easier, however the VScode debugging setup could get more love :)


eoger:
Yup didn't realize there was one .vscode dir per project, closing!

PR #4445, opened 2020-03-06
feat: add pm2 debug servers and add'l debug scripts

bbangert:
Because:

* It can be useful to debug multiple services at once.

This commit:

* Adds a pm2 config that starts 4 of the servers using start-dev-debug
  with unique inspect ports so that they can be debugged all at the
  same time.
* Updates the package.json files for these services so they don't choose
  the same port.
* Documents the ports used for debugging.
* Add's an attach debug for VS Code that lets one choose which process
  to debug.


mysql_servers.json
```diff
@@ -190,7 +190,8 @@
     },
     {
       "name": "event-broker",
-      "script": "../../_scripts/start_after_mysql.sh",
```
bbangert: I had originally copied this from another start, but it doesn't need to wait for mysql so this isn't needed.

packages/fxa-content-server/package.json
```diff
@@ -1,168 +1,168 @@
...
+    "upng-js": "2.1.0",
+    "webpack-dev-middleware": "^3.7.2",
+    "xmlhttprequest": "git://github.com/zaach/node-XMLHttpRequest.git#onerror",
+    "yargs": "^15.0.2"
+  },
+  "engines": {
+    "node": ">=10",
+    "npm": ">=6.4.1"
+  },
+  "readmeFilename": "README.md"
```
bbangert: Prettier decided it was time to fix the spacing. 🤷‍♂

mysql_servers.json
dannycoates: But it does depend on auth-server which waits for mysql. It was crashing for me when it starts before auth.

bbangert: Ah, good point! That's why I had that there.

bbangert: Reverted.

PR #4447, opened 2020-03-06
feat(experiments): Remove CWTS experiment

### LINKED ISSUE #4159: CWTS on Registration still Running As A/B Test?
On November 19th, we rolled out the A/B test that put CWTS in the registration form. On November 22nd, we chose to already roll it out to 100%. Shane took care of that roll-out and confirmed it. As a result, we annotate our charts to mark the date.

Yesterday, [~rfeeley@mozilla.com] ended up in the old flow via the about:welcome page where CWTS was displayed after the registration form.

According to Vijay, this is still running as an A/B test at 10%/10%. 

When I look at metrics, the experiment seems to have been ended at the above mentioned date with the exception of treatment and the control of the control still being visible (and pretty evenly split):
[https://analytics.amplitude.com/mozilla-corp/chart/bs4sg9z/edit/z5rq0jq](https://analytics.amplitude.com/mozilla-corp/chart/bs4sg9z/edit/z5rq0jq)

It's unclear to me what is happening here.

Also, as of Dec 16th, there seems to be a change in the metrics where the control of the control seems to be outperforming the treatment. CC [~loines@mozilla.com]

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1112)
┆Issue Number: FXA-1112

vbudhram:
Fixes #4159 

This enables CWTS on signup page for all users that are coming from sync or users coming from browsers that support the `multiService` property.


packages/fxa-content-server/app/scripts/views/mixins/cwts-on-signup-password.js
```diff
@@ -101,29 +93,31 @@ export default {
...
-      }
+    const offeredSyncEngines = this._getOfferedEngineIds();
+    const declinedSyncEngines = this._getDeclinedEngineIds();
+    const enabledSyncEngines = offeredSyncEngines.filter(
+      e => declinedSyncEngines.indexOf(e) === -1
+    );
+
+    // Tell multi-service browser integrations whether to send
+    // the offered and declined sync engines.
+    if (declinedSyncEngines.length === offeredSyncEngines.length) {
```
jodyheavener: Pretty verbose put perhaps a more clear way of indicating "all offered sync engines were declined": `offeredSyncEngines.every(e => declinedSyncEngines.includes(e))` - by no means a blocker though.

vbudhram: The other way was a bit clearer to me heh, I'll keep for now.

Issue #4403, opened 2020-03-03
Regular select_email_services.js test failures due to sandbox timeout


rfk:
Whenever I run the auth-server tests locally, I always get test failures in `select_email_services.js`; see the bottom of my comment for the full listing, in case they help someone else searching for this issue. @dannycoates mentioned that he sometimes sees these failing sporadically in CI as well.

I think I finally tracked it down to an issue with the sandboxed execution of email-matching regexes, which we do here:

https://github.com/mozilla/fxa/blob/62c41589f3c7efaf412ca6e56ec9946cbdf7d370/packages/fxa-auth-server/lib/senders/select_email_services.js#L196-L199

Every call to `sandbox.run` spawns a new node process in order to execute the regex. I'm on windows, so spawning new processes is expensive and this basically always times out for me. I find it plausible that this would occasionally timeout in CI as well.

I tried to work around this by increasing the timeout, but it turns out the `timeout` argument we're providing above [gets](https://github.com/gf3/sandbox/pull/53) [ignored](https://github.com/gf3/sandbox/issues/46). If I manually change the default timeout in the sandbox library's code, I can get the tests passing reliably (albeit slowly).

I can continue just ignoring these local failures, but spawning a new process every time we want to pick an email sender seems pretty expensive to me. I'll also note that the [sandbox library](https://github.com/gf3/sandbox) hasn't been updated in 6 years, which isn't in itself a bad thing. It might be a concern if we started using it for processing more untrusted input besides regexes though, especially given [long-standing sandbox escape issues](https://github.com/gf3/sandbox/issues/50).

Can we do something cheaper here to sanitize the config loaded from redis? I believe the thing we're trying to guard against is a catastrophic-backgracking regex DoS if an untrusted regex somehow makes its way into the live config in redis, and spawning a whole new process for each regex test seems like a very large general-purpose hammer for this very specific nail.

---

```
rfk@pomelo:fxa-auth-server(refresh-token-last-used-at-in-redis *)$ VERIFIER_VERSION=0 ./scripts/test-local.sh  ./test/local/senders/select_email_services.js 
keys file already exists

  selectEmailServices:
    redis.get returns sendgrid percentage-only match:
      ✓ selectEmailServices returns the correct data
    redis.get returns sendgrid percentage-only mismatch:
      ✓ selectEmailServices returns the correct data
      redis.get returns sendgrid regex-only match:
        1) selectEmailServices returns the correct data
      redis.get returns sendgrid regex-only mismatch:
[...]

PR #4448, opened 2020-03-06
fix(config): Use a safe regex library rather than sandboxing builtins.

rfk:
I may have gotten a little..uh..nerd-sniped this afternoon, but I took at look at alternatives for https://github.com/mozilla/fxa/issues/4403 and I thought to myself: *"if only we could use the regex crate from rust, which is specifically designed to work with untrusted inputs!"*.

Less than an hour of playing around with `wasm-pack` later, I seem to have a minimal PoC that works as intended:

  https://github.com/rfk/sandboxed-regexp

I don't think we should land this as-is, but with a bit of polish, I think it could actually be a safer and faster approach than what we're currently doing with sandboxed execution.

What do you think, worth pursuing further, or just-too-wacky-to-use-in-practice?

---

Because:

* We want to safely work with regexes from potentially-untrusted sources.
* The native JS RegExp class is vulnerable to DoS under such circumstances.
* Our existing sandboxing solution uses a library with known escape issues.
* It's fun to play with the rust+wasm toolchain.

This commit:

* Replaces our bespoke evaluate-regexp-in-a-sandbox solution with
  a wasm-compiled version of the Rust `regex` crate, which is
  specifically designed for working with untrusted regexes.


packages/fxa-auth-server/test/local/senders/select_email_services.js
```diff
@@ -719,173 +653,6 @@ describe('selectEmailServices:', () => {
...
-      },
-    ]);
-
-    assert.equal(safeRegex.callCount, 1);
-    const args = safeRegex.args[0];
-    assert.equal(args.length, 1);
-    assert.equal(args[0], '^((((.*)*)*)*)*@example\\.com$');
-  });
-});
-
```
rfk: Look at all the tests you don't need when your regex library doesn't have any such thing as an "unsafe regex"!

dannycoates:
Pretty effing sweet @rfk 🤘

rfk:
> Maybe we should move it to the Mozilla org, so you aren't permanently on the hook for solo maintenance.

Yep, very happy to do so, just didn't want to make speculative `mozilla` repo too early in the game 👍 

I poked at this a little more today. I was worried that the overhead of making a wasm module and its more expensive regexp compilation procedure would be too much, so I made a tiny bad comparison with the old "spawn a process in a sandbox" approach using a very simple regex. It seems consistently 60 to 100 times on my machines, which seems promising :-)

@dannycoates @6a68 if you're interested, please try pulling latest master from https://github.com/rfk/sandboxed-regex and running `make bench`. I'd be curious to see what the numbers look like on someone else's machine, just for a gutcheck.

6a68:
Typo: the repo is https://github.com/rfk/sandboxed-regexp. Took me a minute to find it.

Note that I had to manually `npm install sandbox` to get `make bench` to run. I got about 70x faster results from `make bench` across 3 runs.

I could see Glean possibly using the rust-wasm approach for maintaining a single codebase that runs in web content, so this might be a good first step to explore how wasm-rust works out in production.

rfk:
To my surprise, I think this is actually ready for review.

First, you'll want to take a peek at the [sandboxed-regexp](https://github.com/rfk/sandboxed-regexp) repo. There's not a lot of code in there but it's worth reviewing to re-assure yourself that what it's doing makes sense, and considering the [list of caveats](https://github.com/rfk/sandboxed-regexp#being-careful-with-it). If it seems like a good path forward, we can move it under the mozilla org and publish a new version with more co-maintainers.

With that available from npm, the diff here is mostly just a cleanup! You will definitely want to view it with whitespace changes hidden, otherwise git makes an absolute dog's breakfast of the diff.

packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -94,115 +163,60 @@ module.exports = (log, config, mailer, emailService) => {
...
+        liveConfig = lastSeenLiveConfigParsed;
+      } else {
+        try {
+          lastSeenLiveConfig = liveConfig;
+          liveConfig = LiveConfig.parse(liveConfig);
+          lastSeenLiveConfigParsed = liveConfig;
+        } catch (err) {
+          lastSeenLiveConfig = lastSeenLiveConfigParsed = null;
+          log.error('emailConfig.parse.error', { err: err.message });
+        }
```
rfk, line 176: The new sandboxing approach seems uniformly faster than the old one in our tests, so there's technically no need for this extra caching logic here. But it just seems like an easy win for a value that doesn't change very often and can be safely cached in memory.

(Also, creating many `SandboxedRegExp` instances at once will allocate lots of memory, which the garbage-collector will probably stay on top of, but still...we can easily save it some work here).

packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -94,115 +163,60 @@ module.exports = (log, config, mailer, emailService) => {
...
       }
     }
 
-    const services = await emailAddresses.reduce(
-      async (promise, emailAddress) => {
```
rfk, line 105: Not sad to see the end of this very clever, but terrifying nested-promise-reducer-loop. Being able to test the regexes synchronously is a nice side benefit of this change.

bbangert:
@rfk Looks like there's some test failures here.

packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -17,6 +16,98 @@ const SERVICES = {
...
+// Where a percentage and a regex are both present, an email address must
+// satisfy both criteria to count as a match. Where an email address matches
+// sendgrid and ses, sendgrid wins. Where an email address matches socketlabs
+// and ses, socketlabs wins. Where an email address matches sendgrid and
+// socketlabs, sendgrid wins.
+//
+// Regular expressions are executed using the `SandboxedRegExp` class in order
+// to protect againt DoS if a malicious regex happened to make its way into the
+// live config; as a result the regex syntax may differ from that of standard
+// javascript, but you shouldn't be putting complex regex in here anyway.
```
dannycoates, line 45: I'd like a link to https://docs.rs/regex/1.3.6/regex/#syntax or even https://github.com/rfk/sandboxed-regexp#being-careful-with-it here

packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -17,6 +16,98 @@ const SERVICES = {
...
+
+class LiveConfigForSender {
+  constructor(props) {
+    const { percentage, regex } = props;
+    if (percentage && percentage >= 0 && percentage < 100) {
+      this.percentage = percentage;
+    } else {
+      this.percentage = null;
```
dannycoates, line 53: nit: from a types perspective it'd be nice if `this.percentage` was just `number` not `number | null` (in a distant typescript future)

rfk: makes sense, I'll default it to `0` rather than `null`.

rfk:
> If it seems like a good path forward, we can move it under the mozilla org and publish a new
> version with more co-maintainers.

I'm going to go ahead and take a look at transferring this repo.

rfk:
I'd also like to find a way to test this new implementation against our current live config, just in case. @jrgm are you able to suggest an easy way to do that, which doesn't involve sending me the current config values (which I assume contain PII)?

I could for example prep a stand-alone script that uses the new sandboxing code, for someone from ops to try out manually.

jrgm:
We actually haven't used this for any particularly tortuous regex patterns in email routing so you can just pick your own test patterns (a couple of emails, now retired from use).

packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -17,6 +16,101 @@ const SERVICES = {
...
+//     regex: "^.+@example\.com$"
+//   },
+//   socketlabs: {
+//     percentage: 100,
+//     regex: "^.+@example\.org$"
+//   },
+//   ses: {
+//     percentage: 10,
+//     regex: ".*",
+//   }
```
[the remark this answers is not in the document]
rfk: It does have a fair few tests already to cover the various cases, and I'm happy to not have had to change any of them as part of this PR 😄

rfk:
Thanks for the feedback all, I've rebased this and will go ahead and merge if it's green.

It sounds the risk of this going sideways in prod will be fairly low, but a couple of things we should watch out for:

* If the new `SandboxedRegexp` class starts throwing errors, I don't see any "catch errors and default to a fallback" logic, so they'll probably show up as 500 errors on email sending routes, which I'm sure we'll notice quickly i prod.
* If the new logic turns out not to process things correctly in production, it'll show up as a change in the relative volume of emails being sent by different backends. I'm not sure how reliably or quickly we'd detect such an issue in production, but it'll be worth keeping an eye on when this goes live.

Issue #1215, opened 2019-05-22
Improve the spacing of the App and Play Store badges


ryanfeeley:
UX team request to cease the hard right/left alignment of our App and Play Store badges on all screens that contain them.

CURRENT
![Screen Shot 2019-05-22 at 3 15 34 PM](https://user-images.githubusercontent.com/68213/58202762-64426900-7ca6-11e9-960f-56c37aa6df29.png)

PROPOSED (centered, 1/2 height of badge apart, 22px)
![Screen Shot 2019-05-22 at 3 20 45 PM](https://user-images.githubusercontent.com/68213/58202773-699fb380-7ca6-11e9-98d7-7d3134874109.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-99)
┆Issue Number: FXA-99

kunalmohan:
Hi! I would like to work on this issue. I would need some guidance, like the url for the above webpage.

PrachiSarda:
Hey! Can I take up this issue?

6a68:
@ryanfeeley Hey, I'm curious why UX would want to do this. Doesn't moving buttons closer together increase the likelihood of a user clicking the wrong button?

vbudhram:
Hey @kunalmohan Are you still interested in this issue? If not @PrachiSarda you are welcome to take it.

If you open the page https://accounts.firefox.com/connect_another_device in Chrome or any non-Firefox browser, you will see the screen above. The css for the elements are located [here](https://github.com/mozilla/fxa/blob/c4ffbda3ff9ddc90d43516a03e624ae23e5739fb/packages/fxa-content-server/app/styles/modules/_marketing.scss#L1).

Hope that helps!

kunalmohan:
I am not working on this issue. @PrachiSarda feel free to take this up.

SaumyaSinghal:
I'm fairly new to this but I'd love trying to fix it. Can I take this up?

victorphoenix3:
Are you working on this @SaumyaSinghal ? Can I work on this?

clouserw:
@jodyheavener should this issue be closed?

jodyheavener:
Yes, thanks.

PR #4454, opened 2020-03-06
refactor(sms page): cosmetic/copy updates and validation functionality for SMS page

### LINKED ISSUE #4431: SMS Clarity: Simple copy + style updates
* Remove FF logo from the SMS page
* Add new, shorter legal copy
* Group store badges closer together (see GH PR #4170, might be some overlap)
* Shorter link copy

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1297)
┆Issue Number: FXA-1297

### LINKED ISSUE #4433: SMS Clarity: "Send" button disable/enable based on crude client-side verification check
The "send" button should be disabled by default.

We accept many characters including dashes, parentheses, pluses, and spaces and then strip them and check server-side if the number is valid. We display an "invalid phone number" tooltip accordingly. Full number validation on keyup would be complex as we'd likely need to send a POST request anyway to invalidate certain numbers so on the client-side, we can at least block A-Z characters and make sure there is a minimum of 8 numbers entered before enabling the button.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1301)
┆Issue Number: FXA-1301

jodyheavener:
Closes #4431, Closes #4433

Various updates as part of FXA-878

- Validates phone number as you type, conditionally enabling submit button (see code comment)
- Disable Firefox logo in container
- Update SMS disclaimer and multi-device explanation copy
- Groups app download links closer together (I think this also addresses #1215?)

Screen Shot 2020-03-06 at 11 48 28 AM

Note the disabled button is not gray, per the guidelines, but faded blue as this was what already exists in our design patterns/CSS. Happy to change to the gray version if desired.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sms_send.js
+++ b/packages/fxa-content-server/app/scripts/views/sms_send.js
@@ -7,6 +7,7 @@
  * to a mobile device via SMS.
  */
 
+import { assign } from 'underscore';
 import AuthErrors from '../lib/auth-errors';
 import Cocktail from 'cocktail';
 import CountryTelephoneInfo from '../lib/country-telephone-info';
@@ -32,12 +33,11 @@ class SmsSendView extends FormView {
   mustAuth = true;
   template = Template;
 
-  getAccount() {
-    // TODO - remove the `|| ...` when done with development
-    return this.model.get('account') || this.user.getSignedInAccount();
-  }
-
   initialize(options) {
+    this.events = assign(this.events, {
+      'input .phone-number': '_validatePhoneNumber',
+    });
+
     this._attachedClients = options.attachedClients;
     if (!this._attachedClients) {
       this._attachedClients = new AttachedClients([], {
@@ -45,6 +45,45 @@ class SmsSendView extends FormView {
       });
     }
     this._userHasAttachedMobileDevice = false;
+  }
+
+  afterRender() {
+    // Make this a jQuery object so we can use the validation helpers
+    this.phoneField = this.el.querySelector('.phone-number');
+    this.$phoneField = this.$(this.phoneField);
+    this.submitButton = this.el.querySelector('.sms-send');
+
+    // It is possible to land on this page with the phone number field
+    // pre-filled, so let's validate on page load as well if that's the case.
+    if (this.formPrefill.get('phoneNumber')) {
+      this._validatePhoneNumber();
+    }
+  }
+
+  getAccount() {
+    // TODO - remove the `|| ...` when done with development
+    return this.model.get('account') || this.user.getSignedInAccount();
+  }
+
+  _validatePhoneNumber() {
+    try {
+      // `disableValidation` is an arbitrary property so we
+      // can disable this validation for testing purposes.
+      // See: tests/functional/send_sms.js
+      if (!this.phoneField.disableValidation) {
+        this.$phoneField.validate();
+      }
+
+      // We want the disabled style of secondary-button (grey),
+      // but the active state of primary-button (blue).
+      this.submitButton.classList.add('primary-button');
+      this.submitButton.classList.remove('secondary-button');
+      this.submitButton.disabled = false;
+    } catch (e) {
+      this.submitButton.classList.add('secondary-button');
+      this.submitButton.classList.remove('primary-button');
+      this.submitButton.disabled = true;
+    }
   }
 
   setInitialContext(context) {
--- a/packages/fxa-content-server/app/styles/modules/_branding.scss
+++ b/packages/fxa-content-server/app/styles/modules/_branding.scss
@@ -36,6 +36,10 @@
   }
 }
 
+#main-content.send-sms::before {
+  content: initial;
+}
+
 #about-mozilla {
   @include respond-to('big') {
     background: image-url('mozilla.svg') no-repeat center center;
@@ -175,7 +179,7 @@
 .sms-disclaimer {
   font-size: 12px;
   line-height: 1.5;
-  margin: 20px 0;
+  margin: 20px 0 22px;
   padding: 0 20px;
 }
 
--- a/packages/fxa-content-server/tests/functional/send_sms.js
+++ b/packages/fxa-content-server/tests/functional/send_sms.js
@@ -42,6 +42,7 @@ const {
   fillOutEmailFirstSignUp,
   getSms,
   getSmsSigninCode,
+  testElementDisabled,
   noSuchElement,
   openPage,
   switchToWindow,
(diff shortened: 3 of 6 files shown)


packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -31,11 +31,28 @@ class SmsSendView extends FormView {
   mustAuth = true;
   template = Template;
 
+  afterRender() {
+    // Make this a jQuery object so we can use the validation helpers
+    this.phoneField = this.$el.find('.phone-number');
```
jodyheavener: Turns out we already have a [jQuery-based validation method](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/views/elements/tel-input.js#L25-L38) that can be utilized. I am happy to break off of this if desired.

packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -31,11 +31,28 @@ class SmsSendView extends FormView {
...
+    this.submitButton = this.el.querySelector('.sms-send');
+
+    this.phoneField.on('input', this._validatePhoneNumber.bind(this));
```
jodyheavener: I spent a while trying to get a Backbone-style event set up, with no luck. Am I missing something funky with our Backbone setup?

```javascript
import { assign } from 'underscore';

events = assign(this.events, {
  'input .phone-number': '_validatePhoneNumber'
})
```

packages/fxa-content-server/app/scripts/views/sms_send.js
LZoog: Awesome find! I tested a bunch of different stuff in here and it seems to work very well. Alex will be happy. :) 

We probably want to run `_validatePhoneNumber` when the user redirects back to `sms` as well, otherwise when you click "Wrong number?" from `sms/sent`, the button is disabled despite being prefilled with the number just sent. I think we probably also want a disabled grey style based on the feature doc.

Also, will you squash your first 4 commits?

packages/fxa-content-server/tests/functional/send_sms.js
```diff
@@ -202,56 +203,28 @@ const suite = {
...
         .then(openPage(SEND_SMS_URL, selectors.SMS_SEND.HEADER))
         .then(type(selectors.SMS_SEND.PHONE_NUMBER, '2134567a890'))
-        .then(click(selectors.SMS_SEND.SUBMIT))
-        .then(testElementExists(selectors.SMS_SEND.PHONE_NUMBER_TOOLTIP))
-        .then(
-          testElementTextInclude(
-            selectors.SMS_SEND.PHONE_NUMBER_TOOLTIP,
-            'invalid'
-          )
-        );
```
LZoog: I think we want to leave at least one of these tests in here, there's additional validations server-side:

image

jodyheavener: Good call. I've added an override to disable input-based validation so we can test both bad and no input.

packages/fxa-content-server/app/scripts/views/sms_send.js
jodyheavener: Done and done.

packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -31,11 +32,38 @@ class SmsSendView extends FormView {
...
+    });
+  }
+
+  afterRender() {
+    // Make this a jQuery object so we can use the validation helpers
+    this.phoneField = this.el.querySelector('.phone-number');
+    this.$phoneField = this.$(this.phoneField);
+    this.submitButton = this.el.querySelector('.sms-send');
+
+    this._validatePhoneNumber();
```
LZoog: Tiny nit: can we add a small comment here with a reason why we need to run this on `afterRender`? Or maybe do a check for `if (this.formPrefill.get('phoneNumber')) {` (I think that's right) for one less render + kind of self-documenting code?

packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -31,11 +32,38 @@ class SmsSendView extends FormView {
...
+  }
+
   getAccount() {
     // TODO - remove the `|| ...` when done with development
     return this.model.get('account') || this.user.getSignedInAccount();
   }
 
+  _validatePhoneNumber() {
+    try {
+      if (!this.phoneField.disableValidation) {
```
LZoog: Can we put `// for testing only` or something of the sort if that's all we're using this for?

packages/fxa-content-server/app/styles/modules/_branding.scss
```diff
@@ -36,6 +36,10 @@
   }
 }
 
+#main-content.send-sms::before {
+  content: initial;
+}
+
```
LZoog: You may have already checked with Alex/Ryan, but does he want the button to be this disabled blue instead of the grey shown in the Feature Doc? (I think the blue makes the most sense, just double checking this is what design wants)

image

image

packages/fxa-content-server/tests/functional/send_sms.js
```diff
@@ -202,48 +216,49 @@ const suite = {
...
         .then(type(selectors.SMS_SEND.PHONE_NUMBER, '2134567'))
-        .then(click(selectors.SMS_SEND.SUBMIT))
-        .then(testElementExists(selectors.SMS_SEND.PHONE_NUMBER_TOOLTIP))
-        .then(
-          testElementTextInclude(
-            selectors.SMS_SEND.PHONE_NUMBER_TOOLTIP,
-            'invalid'
-          )
-        );
+        .then(testElementDisabled(selectors.SMS_SEND.SUBMIT));
```
LZoog: I still think we should have at least one test with a full number that validates on the client-side and hits the `invalid phone number` server tooltip error (like `12231231234`) but these changes look good 👍

packages/fxa-content-server/app/styles/modules/_custom-rows.scss
```diff
@@ -18,6 +18,11 @@
     }
   }
 
+  .spaced {
+    margin-left: 10px;
+    margin-right: 10px;
```
LZoog: I think we can add these styles to `.marketing-link` without a new class, do you think we'll use this maybe in the future?

packages/fxa-content-server/tests/functional/send_sms.js
jodyheavener: I might be misunderstanding, but that's tested [here](https://github.com/mozilla/fxa/pull/4454/files#diff-fe73cbe23907b9f163e8e8c80dd49cf9R257)? `12231231234` is a valid phone number, so should pass, but we test against a mixed-character type value.

packages/fxa-content-server/app/styles/modules/_custom-rows.scss
jodyheavener: That works for me!

packages/fxa-content-server/tests/functional/send_sms.js
LZoog: Hmm, it passes the client-side validation and the button becomes enabled, but the server still returns this error. I think the jQ plugin is still a crude check.

image

jodyheavener: Updated to test this failure!

PR #4457, opened 2020-03-06
Refactor our CircleCI workflows

dannycoates:
In order to reduce our CI resource usage we've combined some jobs and changed what runs for each workflow.

Most packages have fast test runs. We've combined these under one job. Packages with slower runs use their own job so they can be run in parallel. Also, we've added branch filters to the rarely changing, but resource intensive jobs.

We now have three workflows that run at different times:
    
1. test_pull_request: runs only on PR branches and only runs tests.
2. deploy_master: runs only on master and only builds and deploys docker images
3. test_and_deploy_tag: runs only on tags and tests, builds, and deploys

 ## Review notes

- `config.yml` is the biggie.
- I made a few small changes to some packages, but only to get tests passing.


.circleci/README.md
```diff
@@ -0,0 +1,56 @@
+# CircleCI
```
vbudhram: Thanks for documenting this!

packages/fxa-profile-server/test/api.js
```diff
@@ -875,8 +875,8 @@ describe('/avatar', function() {
 
   describe('upload', function() {
     it('should upload a new avatar', function() {
-      this.slow(2000);
-      this.timeout(3000);
+      this.slow(5000);
```
vbudhram: Did the smaller instance size cause this to be slower?

packages/fxa-event-broker/package.json
```diff
@@ -8,10 +8,10 @@
     "audit": "npm audit --json | audit-filter --nsp-config=.nsprc --audit=-",
     "lint:tslint": "./node_modules/tslint/bin/tslint -p .",
     "watch": "tsc -w",
-    "test": "FIRESTORE_EMULATOR_HOST=localhost:8006 FIRESTORE_PROJECT_ID=fx-event-broker ./node_modules/mocha/bin/mocha -r ts-node/register test/**/*.spec.ts test/**/**/*.spec.ts",
+    "test": "mocha -r ts-node/register test/**/*.spec.ts test/**/**/*.spec.ts",
     "start-prod": "npm run build && node ./dist/bin/worker.js",
-    "start": "npm run build && NODE_ENV=development PUBSUB_EMULATOR_HOST=localhost:8005 FIRESTORE_EMULATOR_HOST=localhost:8006 node ./dist/bin/workerDev.js",
-    "start-dev-debug": "NODE_ENV=development PUBSUB_EMULATOR_HOST=localhost:8005 FIRESTORE_EMULATOR_HOST=localhost:8006 nodemon --legacy-watch --watch ./dist --inspect=0.0.0.0:9180 --nolazy ./dist/bin/workerDev.js"
+    "start": "npm run build && NODE_ENV=development node ./dist/bin/workerDev.js",
+    "start-dev-debug": "NODE_ENV=development nodemon --legacy-watch --watch ./dist --inspect=0.0.0.0:9180 --nolazy ./dist/bin/workerDev.js"
```
bbangert: I was under the impression those env vars need to be set or it won't use the emulators. That isn't the case after all?

packages/fxa-profile-server/test/api.js
dannycoates: yes. it was right on the edge and failed sometimes on a `small`

packages/fxa-event-broker/package.json
dannycoates: It looks like the ports are hard coded in the tests. I didn't want to dig too deep.

dannycoates:
Thanks, y'all. I'm going to rebase and try enabling custom-server and event-broker

lmorchard:
FWIW, whenever this is done & merged, remember to update the ecosystem docs 😅 

https://github.com/mozilla/ecosystem-platform/blob/master/docs/fxa-engineering/fxa-tests-circleci.md

.circleci/README.md
lmorchard: Actually, could this mostly replace the ecosystem site doc? Lots of per-job detail that that seems to go away now

dannycoates: Thanks, Les! I'll update that next. Maybe just to point here

packages/fxa-event-broker/package.json
dannycoates: i put the env vars back

PR #4463, opened 2020-03-08
fix(avatar): update avatar image cropper to work effectively at all window sizes

### LINKED ISSUE #674: Avatar crop UI is uncentered on small screens/mobile
child of https://github.com/mozilla/fxa-content-server/issues/6261

(Originally reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1246138 for the specific case of the first-run flow)

STR: 
- As a logged-in user, visit https://accounts.firefox.com/settings, edit your avatar, and upload a new picture in order to access the /settings/avatar/crop screen.
- Make your window really short.

The resulting cropper UI is off-center like so:
screen shot 2016-02-08 at 13 06 40

If you make the window bigger a different set of styles kick in and it looks fine:
screen shot 2016-02-08 at 13 06 31

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-782)
┆Issue Number: FXA-782

jodyheavener:
Closes #674

This should address the avatar image cropper at smaller window sizes.
This was actually a bit of a doozy, since the cropper sizing was hard coded. The remedies I've used are:

- Update all the cropper component styles to be a little more fluid.
- Remove the ability to specify explicit vertical and horizontal gutters, instead relying on the calculated width of the wrapper versus the size of the cropping box.
- Position accordingly on page load and on window resize (debounced).

[Here's a demo video](https://imgur.com/v2qU6AK) of the cropping container at various window sizes.

Screen Shot 2020-03-07 at 9 36 04 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/cropper.js
+++ b/packages/fxa-content-server/app/scripts/lib/cropper.js
@@ -6,7 +6,6 @@ import _ from 'underscore';
 
 var DEFAULT_DISPLAY_LENGTH = 240;
 var DEFAULT_EXPORT_LENGTH = 480;
-var DEFAULT_GUTTER = 40;
 
 /*
  * options: {
@@ -14,8 +13,6 @@ var DEFAULT_GUTTER = 40;
  *  src: The image source. Data URIs are okay.
  *  width: The image width. Required if src is set.
  *  height: The image height. Required if src is set.
- *  horizontalGutter: The amount of space between the crop zone and the sides of the wrapper
- *  verticalGutter: The amount of space between the crop zone and the top/bottom of the wrapper
  *  displayLength: The length of the crop square during cropping
  *  exportLength: The length of the final cropped image
  *  onRotate: Function to call back when the rotation button is clicked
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/lib/cropper.js
```diff
@@ -44,19 +39,37 @@ function Cropper(options) {
...
+    const maxLength = this.exportLength * 2;
     if (img.width > maxLength || img.height > maxLength) {
-      var scale = maxLength / Math.max(img.width, img.height);
+      const scale = maxLength / Math.max(img.width, img.height);
       img = this.resize(img, scale);
     }
     this.setImageSrc(img.src, img.width, img.height);
   }
+
+  setTimeout(() => {
```
jodyheavener: I spent more time than I'd like to admit trying to debug an issue where the cropper is checking for element rect values before the DOM was finished laying out elements, so this is the result -- forcing the cropper to perform its checks after initial layout and execution is finished.

packages/fxa-content-server/app/scripts/lib/cropper.js
```diff
@@ -44,19 +39,37 @@ function Cropper(options) {
   if (!options.container) {
     throw new Error('A container element is required');
   }
+
   this._setupElements(options.container);
 
+  window.addEventListener(
```
LZoog: This is looking _so much better_, we do need to remove the event listener though, otherwise when you close the cropper and resize settings it's still present. :( 

See my comment here: https://github.com/mozilla/fxa/pull/3763#discussion_r364926864

packages/fxa-content-server/app/scripts/lib/cropper.js
```diff
@@ -44,19 +39,37 @@ function Cropper(options) {
...
+    'resize',
```
LZoog: This is looking _so much better_, we do need to remove the event listener though, otherwise when you close the cropper and resize settings it's still present. :( 

See my comment here: https://github.com/mozilla/fxa/pull/3763#discussion_r364926864

jodyheavener: @LZoog I've gone ahead and destroyed the event listener when you close the cropper.

packages/fxa-content-server/app/scripts/lib/cropper.js
```diff
@@ -33,30 +30,45 @@ function Cropper(options) {
...
+    const maxLength = this.exportLength * 2;
     if (img.width > maxLength || img.height > maxLength) {
-      var scale = maxLength / Math.max(img.width, img.height);
+      const scale = maxLength / Math.max(img.width, img.height);
       img = this.resize(img, scale);
     }
     this.setImageSrc(img.src, img.width, img.height);
   }
+
+  setTimeout(() => {
```
LZoog: I tried a few things here too without much luck. Can we add a comment above for justification?

packages/fxa-content-server/app/scripts/lib/cropper.js
```diff
@@ -110,18 +122,43 @@ Cropper.prototype._setupElements = function(container) {
     $sliderEl.val(this.scale);
     this.onZoomIn();
   });
+};
+
+Cropper.prototype.destroy = function(container) {
```
LZoog: I think we can leave an empty parameter here, otherwise I get a `'container' is declared but its value is never read.`

packages/fxa-content-server/app/scripts/views/settings/avatar_crop.js
```diff
@@ -131,6 +126,14 @@ const View = FormView.extend({
...
     this.logViewEvent('zoom.range');
   },
+
+  _returnToAvatarChange() {
+    this.cropper.destroy();
+  },
+
+  _returnToSettings() {
+    this.cropper.destroy();
+  },
```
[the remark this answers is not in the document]
LZoog: @jodyheavener this is my bad, I think this should be the last change before this sucker finally gets merged. `destroy()` above didn't work because I _meant_ to use `remove()`. 🤦‍♀️

You should be able to remove `_returnToAvatarChange()`, `_returnToSettings()`, and `this.cropper.destroy();` on line 96 and do this instead:

```
  remove() {
    this.cropper.destroy();
  }
```

I just tested it and it looks like it's working for me.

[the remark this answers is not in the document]
jodyheavener: You are correct. I've updated accordingly!

Issue #4432, opened 2020-03-05
SMS Clarity: Don't show CAD or SMS on mobile login/registration


data-sync-user:
These screens should not be shown after login or registration for mobile devices.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1302)
┆Issue Number: FXA-1302
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12653/Screen+Shot+2020-03-06+at+4.22.23+PM.png">Screen Shot 2020-03-06 at 4.22.23 PM.png</a>

data-sync-user:
➤ Lauren Zugai commented:

Q for [~adavis@mozilla.com]: should they be redirected to the "ready" screen or maybe "verified"?

data-sync-user:
➤ Alex Davis commented:

Q for [~adavis@mozilla.com]: should they be redirected to the "ready" screen or maybe "verified"?

Do you have a screenshot of the ready or verified screens? I want to be sure we're talking about the same ones?

Our recent discussions have highlighted that we don't always have alignment on names (e.g. connect another device vs SMS).

data-sync-user:
➤ Alex Davis commented:

Context: we marked this as low priority for now because mobile teams have been working around this "bug". They redirect users away quickly before we have time to ask them to connect a mobile device (when they are already on mobile).

Resolving this fixes a long-time outstanding bug though should not have a tremendously noticeable impact.

data-sync-user:
➤ Jody Heavener commented:

I'm taking a stab at this as we're about to wrap up a few of the other issues in this epic.

[~adavis@mozilla.com] can you elaborate on where users should be redirected to? You mentioned "They redirect users away quickly before we have time to ask them to connect a mobile device", where's that? Also, since a user can land on this screen in a variety of ways (post registration, post Sync, post-verify secondary email decline), should we have a few places they could be redirect to?

I'll note that if you hit the `/connect_another_device` path on a mobile device that is ineligible to receive SMS you are shown the "complete set-up" screen (see attached). If you click "Not now" on the SMS screen you are taken [here](https://www.mozilla.org/en-CA/firefox/accounts/ ( https://www.mozilla.org/en-CA/firefox/accounts/ )). The latter seems like a good contender?

data-sync-user:
➤ Jody Heavener commented:

!Screen Shot 2020-03-06 at 4.22.23 PM.png!

data-sync-user:
[...]

Issue #3951, opened 2020-01-21
'.button-link' color contrast issue, lack of hover state, should probably be a link in some cases


LZoog:
I noticed while working on #3843 that:
- The color we set for `.button-link` does not pass ADA color contrast requirements. It's set to `$color-blue` or `#0a84ff`, which fails for "normal" text (bolded text [below 14pt/18.66px](https://webaim.org/resources/contrastchecker/)). Our normal link color is `#0060df` which does pass. They should probably be this color.
- We don't have a hover state for these. They should underline and darken with a `150ms` transition like our normal links do.
- Thirdly, let's double check all instances of `.button-link` should actually be links and not buttons. At the very least, they should be styled differently. When you hover over one of these, since the button takes up the entire width, you can click in whitespace left on the left/right-hand side and interact with it.

Here's one example, this should probably be a link as it just redirects to `/signin`.
<img src="https://user-images.githubusercontent.com/13018240/72850997-0ff5eb80-3c70-11ea-8399-f351241fb1f7.png" width=400>

previous page with just a link:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/13018240/72851456-4a13bd00-3c71-11ea-94e6-678bc00600b0.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-967)
┆Issue Number: FXA-967

vasilicatamas:
I confirm that all "Remember password? Sign in", "Don't have a recovery key" and "Continue to my account" buttons appears as links, have hover state and #0060df color.  

@LZoog I think "Cancel" button displayed in first modal from generate recovery key flow should also follow these rules - filed #6166 

![Screenshot_6](https://user-images.githubusercontent.com/18114210/89668694-e12b2400-d8e6-11ea-8135-44ed1b996983.png)
![Screenshot_7](https://user-images.githubusercontent.com/18114210/89668698-e1c3ba80-d8e6-11ea-8007-3b7ce91ae9b7.png)
![Screenshot_8](https://user-images.githubusercontent.com/18114210/89668699-e25c5100-d8e6-11ea-97ea-f2b8eb935254.png)
![Screenshot_9](https://user-images.githubusercontent.com/18114210/89668702-e25c5100-d8e6-11ea-978f-c74722cf44a0.png)

Based on the above mentioned I am marking this bug as verified.

PR #4465, opened 2020-03-09
feat(sms): Conditionally render heading text based on mobile deviceType presence

### LINKED ISSUE #4435: SMS Clarity: Add header, conditionally render text
The header should read "Would you like to sync your phone?”, but if a mobile device is attached to user’s account, display instead “Still adding devices?”

We'll need to make a call to the endpoint of user's devices and apps and filter through the returned data to check.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1299)
┆Issue Number: FXA-1299

LZoog:
Also adds aria-label and role to a graphic for for ADA purposes.

fixes #4435 

---

We don't have to worry about potential lag between a device just verifying and then fetching from `attachedClients` because #4432 should redirect users that register with a mobile device.

I expect to have merge conflicts if #4454 is merged first.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
@@ -3,7 +3,15 @@
       <div class="error"></div>
     </header>
     <section class="send-sms">
-      <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+      <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Connect another device{{/t}}"></div>
+      <h1>
+        {{^userHasAttachedMobileDevice}}
+          {{#t}}Would you like to sync your phone?{{/t}}
+        {{/userHasAttachedMobileDevice}}
+        {{#userHasAttachedMobileDevice}}
+          {{#t}}Still adding devices?{{/t}}
+        {{/userHasAttachedMobileDevice}}
+      </h1>
       <p>
         {{#t}}Send Firefox directly to your smartphone and sign in to complete set-up{{/t}}
       </p>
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/templates/sms_send.mustache
```diff
@@ -3,7 +3,15 @@
       <div class="error"></div>
     </header>
     <section class="send-sms">
-      <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+      <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Connect another device{{/t}}"></div>
```
jodyheavener: I'm curious if this needs to be a `div` at all, and not just an `img`. If the goal is to make it not interactive we could apply `user-select: none;`. A little out of scope so feel free to disregard.

packages/fxa-content-server/app/scripts/templates/sms_send.mustache
```diff
@@ -3,7 +3,15 @@
...
+      <h1>
+        {{^userHasAttachedMobileDevice}}
+          {{#t}}Would you like to sync your phone?{{/t}}
```
jodyheavener: I know this is in the SMS doc, so feel free to disregard, but I wonder if this should be "device" instead of "phone", considering other verbiage uses "devices" (and after all, you can send to your tablet).

packages/fxa-content-server/app/scripts/templates/sms_send.mustache
LZoog: Yeah you're right, this should just be an image. That reminds me of #3951 where we have some links as buttons. It'd require some refactoring since we grab the image class server-side and apply it here as a BG-image, I know some folks prefer BG images to `img` but if we do it that way, cases like this should have a `role="img"` and an `aria-label` for alt text and we're doing this in other places without the role/aria-label. Either way, I'll create a follow up issue with an `a11y` label for that.

packages/fxa-content-server/app/scripts/templates/sms_send.mustache
LZoog: I think they _want_ that differentiation here. If the user doesn't have a `deviceType` of `mobile` on their account, we want to ask if they'd like to sync their phone. If they do have a phone attached, we ask if they want to keep syncing devices.

jodyheavener: Got it! Thanks for the clarification.

PR #4481, opened 2020-03-10
refactor(oauth): Delete unused /client-tokens routes

### LINKED ISSUE #4407: Remove the unused OAuth `/client-tokens` API
The OAuth server exposes some API endpoints under ` /client-tokens` that can be used for listing and deleting the OAuth clients attached to a user's account. We *used* to use these as part of the devices-and-apps list, but that has since switched over to using the new `/attached_clients` endpoint. Ripgrepping the code and inspecting request logs in bigquery, I don't think anything is using the `/client-tokens` endpoints any more.

We should remove them, which AFAICT would include:

* Deleting the `fetchOAuthApps` and `deleteOAuthApp` methods on the content-server's `oauth-client.js`.
* Removing the code for these routes in auth-server's `./lib/oauth/routes/client-tokens` and the plumbing that hooks them up.
* Removing the `getActiveClientsByUid` method from the auth-server's OAuth DB object, since it is only used by these routes.
* Removing auth-server's `lib/oauth/db/helpers.js` since it only contains a single function and that function is only used by `getActiveClientsByUid` above.
* Removing the `id` property from auth-server's `lib/oauth/db/accessToken.js`, since it only exists so these objects can be passed to the helper above.

Not a bad potential cleanup overall!

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1286)
┆Issue Number: FXA-1286

imskr:
Fixes #4407

@rfk


imskr:
@rfk I have made changes as requested

imskr:
Thanks for helping me.

Issue #3981, opened 2020-01-30
Longer Duration subscription refunds


data-sync-user:
As a subscriber with a longer-duration subscription, so that I can change my mind about my purchase, I would like to be able to receive a full refund if I elect to in the first 30 days of my subscription

┆Issue is synchronized with this [Jira Story](https://jira.mozilla.com/browse/FXA-982)
┆Issue Number: FXA-982

Tasks:

- [x] Update Formatted Subscription Response [Jira Task: FXA-1334](https://jira.mozilla.com/browse/FXA-1334) - Pull Request: #4489
- [ ] Update Support Panel [Jira Task: FXA-1316](https://jira.mozilla.com/browse/FXA-1316) - Pull Request: #4502
- [X] Update Zendesk Widget [Jira Task: FXA-1317](https://jira.mozilla.com/browse/FXA-1317) - Pull Request: https://github.com/mozilla/omniscient/pull/11
- [x] Deploy Zendesk Widget [Jira Task: FXA-1318](https://jira.mozilla.com/browse/FXA-1318)

data-sync-user:
➤ John Gruen commented:

main thing here is that support agents can see when subscription started in the zendesk dashboard...

PR #4489, opened 2020-03-10
feat(auth-server): add created to subscription response

jackiemunroe:
- add created timestamp to the subscription response

Because:
- as a part of allowing longer duration subscriptions, we need to provide support agents with the date the subscription began for the purpose of time-sensitive refunds

issue #3981

### DIFF
--- a/packages/fxa-auth-server/lib/routes/validators.js
+++ b/packages/fxa-auth-server/lib/routes/validators.js
@@ -302,6 +302,7 @@ module.exports.activeSubscriptionValidator = isA.object({
 
 // This is subhub's perspective on an active subscription
 module.exports.subscriptionsSubscriptionValidator = isA.object({
+  created: isA.number().required(),
   current_period_end: isA.number().required(),
   current_period_start: isA.number().required(),
   cancel_at_period_end: isA.boolean().required(),
--- a/packages/fxa-auth-server/test/remote/subscription_tests.js
+++ b/packages/fxa-auth-server/test/remote/subscription_tests.js
@@ -179,13 +179,14 @@ describe('remote subscriptions:', function() {
 
     describe('with a subscription', () => {
       const subscriptionId = 'sub_12345';
+      const date = Date.now();
       beforeEach(() => {
         mockStripeHelper.customer = async (uid, email) => ({
           subscriptions: {
             data: [
               {
                 id: subscriptionId,
-                created: Date.now() / 1000,
+                created: date,
                 cancelled_at: null,
                 plan: {
                   product: PRODUCT_ID,
@@ -199,8 +200,9 @@ describe('remote subscriptions:', function() {
           {
             subscription_id: subscriptionId,
             plan_id: PLAN_ID,
-            current_period_end: Date.now() / 1000,
-            current_period_start: Date.now() / 1000,
+            created: date,
+            current_period_end: date,
+            current_period_start: date,
             cancel_at_period_end: false,
             end_at: null,
             plan_name: 'foo',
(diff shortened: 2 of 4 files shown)


packages/fxa-auth-server/test/remote/subscription_tests.js
```diff
@@ -199,8 +200,9 @@ describe('remote subscriptions:', function() {
           {
             subscription_id: subscriptionId,
             plan_id: PLAN_ID,
-            current_period_end: Date.now() / 1000,
-            current_period_start: Date.now() / 1000,
+            created: date,
+            current_period_end: date,
+            current_period_start: date,
```
dannycoates: why are these changing from `Date.now() / 1000` to `Date.now() * 1000`?

packages/fxa-auth-server/test/remote/subscription_tests.js
```diff
@@ -179,13 +179,14 @@ describe('remote subscriptions:', function() {
 
     describe('with a subscription', () => {
       const subscriptionId = 'sub_12345';
+      const date = Date.now() * 1000;
```
6a68: Did you mean `Date.now() / 1000`?

packages/fxa-auth-server/test/remote/subscription_tests.js
```diff
@@ -179,13 +179,14 @@ describe('remote subscriptions:', function() {
...
       beforeEach(() => {
         mockStripeHelper.customer = async (uid, email) => ({
           subscriptions: {
             data: [
               {
                 id: subscriptionId,
-                created: Date.now() / 1000,
+                created: date / 1000,
```
6a68: Did you mean `date`?

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -302,6 +302,7 @@ module.exports.activeSubscriptionValidator = isA.object({
 
 // This is subhub's perspective on an active subscription
 module.exports.subscriptionsSubscriptionValidator = isA.object({
+  created: isA.number().required(),
```
6a68: Would it be worthwhile to tighten this to some kind of timestamp value? Or at least a number of digits (so it's clear this is a millisecond vs second timestamp)?

packages/fxa-auth-server/test/remote/subscription_tests.js
jackiemunroe: yep

packages/fxa-auth-server/test/remote/subscription_tests.js
jackiemunroe: Well, the value for this particular part doesn't really matter -  it just is ensuring that the dates are going to have the same result. I was having occasional issues when debugging with the results of `Date.now()` not being the same, so I changed that bit to be easier.

The `Date.now() / 1000` to `Date.now() * 1000` is actually a result of issues I was having with the `listActive` test, I am addressing those via @6a68 's comments.

packages/fxa-auth-server/test/remote/subscription_tests.js
jackiemunroe: Updated

packages/fxa-auth-server/lib/routes/validators.js
jackiemunroe: can do

jackiemunroe: should I update all of the other timestamp instances in that response validator?

jackiemunroe: So, for this response, all of the timestamps are seconds since the Unix epoch. What do you think would be a good validation requirement for this, since technically the size will grow?

6a68: I think it'd be fine to just improve this one line, and just aim to improve the other validators the next time they are edited.

Joi has some [timestamp validation built-ins](https://github.com/hapijs/joi/blob/v9.0.0-3/API.md#datetimestamptype) that could help. I'm now a little confused about whether this is a JS or unix timestamp, but you could do `Joi.date().timestamp()` for millisecond timestamps, or `Joi.date().timestamp('unix')` for second timestamps.

packages/fxa-auth-server/test/remote/subscription_tests.js
6a68: Ah, ok. I wasn't aware of the distinction between the `created` and `createdAt` values

Issue #4436, opened 2020-03-05
SMS Clarity: Add “This Firefox is connected” green banner


data-sync-user:
This banner should only be displayed at the end of a successful login or registration and shouldn’t be displayed through a direct link to the page.
Note: this text already exists in `connect_another_device.mustache`, we should investigate the conditional around that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1298)
┆Issue Number: FXA-1298
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12661/Screen+Shot+2020-03-10+at+9.01.10+AM.png">Screen Shot 2020-03-10 at 9.01.10 AM.png</a>

data-sync-user:
➤ Jody Heavener commented:

[~lzugai@mozilla.com] [~adavis@mozilla.com] Looking into this, connect_another_device.mustache ( https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/templates/connect_another_device.mustache ) isn't the SMS page, it's the "Sign in to this Firefox to complete set-up" page, which would explain why it's not showing up on the SMS page. The Connect Another Device logic is what sends you to the SMS page, so I believe it would be simple enough to pass all the params over and copy this logic over to that page.

data-sync-user:
➤ Lauren Zugai commented:

[~jheavener@mozilla.com] does "This Firefox is connected" ever actually show up in connect_another_device? Do we show that to users? I think [~adavis@mozilla.com] had said to remove it if it was somewhere else but I'd need his clarification there.

Is the logic on `connect_another_device` equivalent to "show on successful login or registration" only? That would be great if so, if we can copy logic over then : partyparrot :

data-sync-user:
➤ Jody Heavener commented:

Yeah, "This Firefox is connected" is shown on CAD after successful Sync signin (see screenshot). I imagine we can shoot this logic over to the SMS page.

data-sync-user:
➤ Jody Heavener commented:

[~adavis@mozilla.com] I've got this set up to display the "This Firefox is connected" banner on the SMS page after you successfully register or sign in via Sync. It can't be seen if you hit the page directly, or even if you refresh the page. Is this the desired flow?

PR #4493, opened 2020-03-10
fix(sms page): display "This Firefox is connected" after sign in or registration via sync occurs

jodyheavener:
Closes https://github.com/mozilla/fxa/issues/4436

We currently display the success banner "This Firefox is connected" on the Connect Another Device screen after a successful sign in or registration via Sync, but there are cases where the user can be shown the SMS screen (skipping CAD altogether), so we want to show that same confirmation message here.

This should only appear in the following scenario:

1. You have just signed in (returning or new registration)
2. You are in a state that is eligible to receive SMS messages

You should not be seeing this message if you hit any variation of `/sms` directly. Even reloading the page when you first see the banner should not show it again.

Screen Shot 2020-03-10 at 9 52 16 AM

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sms_send.js
+++ b/packages/fxa-content-server/app/scripts/views/sms_send.js
@@ -113,6 +113,7 @@ class SmsSendView extends FormView {
 
     context.set({
       userHasAttachedMobileDevice: this._userHasAttachedMobileDevice,
+      isSignedIn: this._isSignedIn(),
       country,
       escapedLearnMoreAttributes,
       graphicId,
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/views/sms_send.js
```diff
@@ -142,6 +143,29 @@ class SmsSendView extends FormView {
     );
   }
 
+  /**
+   * Check if the current user is already signed in.
+   *
+   * @returns {Boolean}
+   * @private
+   */
+  _isSignedIn() {
```
jodyheavener: This method is present in [ConnectAnotherDeviceView](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/views/connect_another_device.js#L179-L194) so I wanted to also bring over the comment context.

packages/fxa-content-server/app/tests/spec/views/sms_send.js
```diff
@@ -310,6 +310,31 @@ describe('views/sms_send', () => {
...
+      assert.isTrue(view._isSignedIn());
+    });
+
+    it('returns `false` otherwise', () => {
+      sinon.stub(user, 'isSignedInAccount').callsFake(() => false);
+      broker.set('isVerificationSameBrowser', false);
+
+      assert.isFalse(view._isSignedIn());
+    });
+  });
```
LZoog: Can we add a test for `showSuccessMessage`? It would be nice to have one in place for when the user should see the banner and when it should be hidden. (probably better as a selenium test but a unit test would be OK)

packages/fxa-content-server/app/styles/modules/_branding.scss
```diff
@@ -36,8 +36,20 @@
   }
 }
 
-#main-content.send-sms::before {
-  content: initial;
+// Note this is non-standard styling for #main-content. It would
+// generally have a top padding of 70px desktop / 82px mobile
```
LZoog: Small nit: these values can change pretty easily and the comment would then be out-of-date, something like "additional" or "more" padding would be a little more evergreen. (Or even just "Less padding is needed when the Firefox logo is hidden" is fine)

packages/fxa-content-server/app/tests/spec/views/sms_send.js
jodyheavener: `showSuccessMessage` is a simple boolean on the model, so I've updated this suite to test the various scenarios in which the banner should be displayed. As discussed in Slack, I was getting a little lost trying to write this as a Selenium test :(

PR #4495, opened 2020-03-10
chore(cleanup): remove obsolete docker files

dannycoates:
PR #4457 made Dockerfile-test obsolete and Dockerfile-build can be renamed to just Dockerfile


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       SKIP_DOCKER: true
     steps:
       - base-install
-      - run: ./packages/<< parameters.package >>/scripts/test-ci.sh
+      - run: ./.circleci/test-package.sh << parameters.package >>
```
dannycoates: I snuck this in. Going directly to `test-ci.sh` skips the check for whether this package needs to be tested. This change fixes that.

.circleci/build.sh
```diff
@@ -18,28 +18,19 @@ if grep -e "$MODULE" -e 'all' $DIR/../packages/test.list > /dev/null; then
   mkdir -p config
   cp ../version.json config
 
+  ODDBALLS=("fxa-auth-server" "fxa-content-server" "fxa-profile-server" "fxa-payments-server")
```
chenba: 💯

PR #4501, opened 2020-03-10
ci(payments): added payments-server to ci tests

dannycoates:
tests hang when run in `test-many`, let's try it's own job.


dannycoates:
It might be cheaper overall to put it in test-many with medium size. let's see

dannycoates:
Napkin math says it's pretty close to the same, but a medium improves times (and prob reliability) for test-many overall

PR #4513, opened 2020-03-11
fix(customs): dump out config at startup

jrgm:
This is so I have a way to check that I have the values I expect when this is set by a comma-separated environment variable.

r? - @clouserw

### DIFF
--- a/packages/fxa-customs-server/lib/settings/requestChecks.js
+++ b/packages/fxa-customs-server/lib/settings/requestChecks.js
@@ -21,9 +21,17 @@ module.exports = (config, Settings, log) => {
       if (!this.flowIdExemptUserAgentREs) {
         this.flowIdExemptUserAgentCompiledREs = [];
       } else {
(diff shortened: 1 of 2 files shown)


packages/fxa-customs-server/lib/settings/requestChecks.js
```diff
@@ -21,9 +21,17 @@ module.exports = (config, Settings, log) => {
...
-        this.flowIdExemptUserAgentCompiledREs = this.flowIdExemptUserAgentREs.map(function(re) {
-          return new RegExp(re);
-        });
+        this.flowIdExemptUserAgentCompiledREs = this.flowIdExemptUserAgentREs.map(
+          function(re) {
+            // Log the regex at startup to check that we have correct values.
+            const regex = new RegExp(re);
+            log.info({
+              op: 'requestChecks.setAll.regex',
+              regex: String(regex),
```
jrgm: This string coercion is to compensate for the logging code in the module being dumb.

PR #4520, opened 2020-03-12
feat(admin-panel): Build out initial UI with nav

### LINKED ISSUE #4204: [admin panel A1.0] Build out initial UI with basic styles and links (depending on wires)
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1139)
┆Issue Number: FXA-1139

LZoog:
This commit sets up routes and introduces a lot of new elements and styles for the beginning of the admin panel email block clearing feature. Components were created with reuse in mind for other packages but styling conventions (including class names, nesting, placement, and use of CSS variables) and components may change as the panel is built out. This commit does not introduce any state just yet.

Tests will be added with the first functional feature.

fixes #4204

image


LZoog:
I'll fix up the test failure in the morning, I need to fix how I have the routing set up.

To reiterate what I said in Slack: I'm still experimenting with CSS vars, I think I like placing them all in `_variables` scoped like they are but it may change or I might even switch to SCSS variables if I miss SCSS features.

packages/fxa-admin-panel/src/components/EmailBlocks/EmailBlock/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <section className="email-block">
+      <ul>
+        <li className="flex justify-space-between">
+          <h3>{account.email}</h3>
+          <span className={`${accountIsVerified ? 'verify' : ''}`}>
+            {accountIsVerified ? 'verified' : 'not verified'}
+          </span>
+        </li>
+        <li className="flex justify-space-between">
```
jodyheavener: I'm curious about the use of utility classes here and concern classes elsewhere. I'm fine with either approach, mostly just curious if we should consider expanding in one direction or the other, _or_ perhaps have a utility library to serve as the foundation of everything and then build on top with individual concerns. Thoughts?

jodyheavener: Oh, I would also be satisifed with "this is just a foundation, don't pay too much attention to it yet" 😁

LZoog: Ideally we use the same methods/namings/approach across the payments server, content server, and admin panel, but what I have in this commit doesn't really yet have things laid out in a solid "preferred" way because I haven't seen any visual designs for the settings redesign yet (the blocker for the CSS architecture ADR I keep mentioning). I went with classes here that made sense for the admin panel and stuck some utility classes on components I'm hoping can be used in the settings redesign. The utility classes _may_ get pulled into their own file and expanded on for that redesign.

We do want to limit dependencies - I personally don't think we're going to need a CSS library for FxA, or if we do, we can make a tiny framework just for us, perhaps with these flex classes as a base for it. This admin panel project is going to end up staying pretty small, besides what I have here we're likely going to only have a few other components.

jodyheavener: Makes sense! Thanks.

PR #4521, opened 2020-03-12
Ensure correct `graphicId` HTML element / labels

### LINKED ISSUE #4473: Ensure correct `graphicId` HTML element / labels
See https://github.com/mozilla/fxa/pull/4465#discussion_r389880425

Places in which we set a graphic through a class on a `div` with a `background-image` need some `a11y` love.

These should preferably be `img` tags for screen readers, however, our image file names receive a hash via grunt on stage/prod build so a simple path to the icon will break in stage. A `data:image/svg+xml;utf8,<svg...` src could be the way to go, however, this may be tricky since we use `getGraphicsId` to check for which class name to set on the div.

At the very least, we can instead place `role="img"` and `aria-label="alt text"` on the divs which will read as images. Reference the link above to see how this was done in one file. Closing this ticket should explore if we can use `img` but if not, proper roles/aria-labels on all divs that receive a `graphicId` will suffice.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1328)
┆Issue Number: FXA-1328

imskr:
added `role` and `aria-label` 

closes #4473

@jodyheavener

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/connect_another_device.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/connect_another_device.mustache
@@ -14,7 +14,7 @@
     {{/isSignedIn}}
   {{/showSuccessMessage}}
   <section>
-    <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Complete your set-up{{/t}}"></div>
 
     {{#canSignIn}}
       <p>
--- a/packages/fxa-content-server/app/scripts/templates/pair/auth_complete.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/pair/auth_complete.mustache
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="{{#t}}Successfully connected{{/t}}"></div>
 
     <h2>{{#t}}You are now syncing with:{{/t}}</h2>
 
--- a/packages/fxa-content-server/app/scripts/templates/pair/index.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/pair/index.mustache
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}"></div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="{{#t}}Scan to connect another device{{/t}}"></div>
     <p>{{#t}}Open the code scanner on your Firefox app and show code when ready.{{/t}}</p>
 
     <form novalidate>
--- a/packages/fxa-content-server/app/scripts/templates/pair/success.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/pair/success.mustache
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}">{{#t}}Device connected{{/t}}</div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="{{#t}}Device successfully paired{{/t}}"></div>
 
     <p>{{#t}}Pairing was successful.{{/t}}</p>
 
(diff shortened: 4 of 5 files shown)


packages/fxa-content-server/app/scripts/templates/connect_another_device.mustache
```diff
@@ -14,7 +14,7 @@
     {{/isSignedIn}}
   {{/showSuccessMessage}}
   <section>
-    <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Successfully authenticated{{/t}}">{{#t}}Success{{/t}}</div>
```
jodyheavener: I think we can remove the text content of these elements (e.g. `{{#t}}Success{{/t}}`). This was likely an older method of aiding screen readers with visibly hidden text, but now that we're moving to a better approach we can probably remove them.

imskr:
@jodyheavener Thanks for your input. I've made the changes as you suggested.

imskr:
okay, I'll be pushing the changes now.

imskr:
@jodyheavener Please review now! Thanks.

packages/fxa-content-server/app/scripts/templates/pair/auth_complete.mustache
```diff
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="{{#t}}Success{{/t}}"></div>
```
jodyheavener: `{{#t}}Successfully connected{{/t}}`

packages/fxa-content-server/app/scripts/templates/pair/index.mustache
```diff
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}"></div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="Connect another device"></div>
```
jodyheavener: `{{#t}}Scan to connect another device{{/t}}` (also, note the addition of the translation tag)

packages/fxa-content-server/app/scripts/templates/pair/success.mustache
```diff
@@ -6,7 +6,7 @@
   <section>
     <div class="error"></div>
 
-    <div class="graphic {{ graphicId }}">{{#t}}Device connected{{/t}}</div>
+    <div class="graphic {{ graphicId }}" role="img" aria-label="{{#t}}Device connected{{/t}}"></div>
```
jodyheavener: `{{#t}}Device successfully paired{{/t}}`

packages/fxa-content-server/app/scripts/templates/sms_sent.mustache
```diff
@@ -6,7 +6,7 @@
     <div class="success"></div>
     <div class="error"></div>
 
-    <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Success{{/t}}"></div>
```
jodyheavener: `{{#t}}We've sent you a text message{{/t}}`

packages/fxa-content-server/app/scripts/templates/connect_another_device.mustache
```diff
@@ -14,7 +14,7 @@
     {{/isSignedIn}}
   {{/showSuccessMessage}}
   <section>
-    <div class="graphic {{graphicId}}">{{#t}}Success{{/t}}</div>
+    <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Successfully authenticated{{/t}}"></div>
```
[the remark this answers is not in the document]
imskr: It is good. I will make the changes

imskr:
@jodyheavener Done

imskr:
@jodyheavener yes, I will do it.

imskr:
It's updated now @jodyheavener

jodyheavener:
@imskr Thank you for taking the time to contribute 🙂

PR #4532, opened 2020-03-12
feat(account deletion): add checkbox to deletion confirming that your developer extensions and themes will be deleted

### LINKED ISSUE #4467: Add warning for AMO Developers during FxA deletion
We are currently working with other teams to ensure that they are listening to FxA deletion events to ensure that account deletions propagate to our relying parties.

AMO requests that we display a message to inform developers that their published add-ons will be deleted.

Attached is a screenshot if what our deletion screen looks like for context.

My proposal is that we add another checkbox that says something like:

* "Any extensions that you published to addons.mozilla.org will be deleted"

In the short-medium term, I don't believe we need to detect if they are a developer. We can display this message to 100% of users even though it will not concern most.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1321)
┆Issue Number: FXA-1321
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12757/image+%282%29.png">image (2).png</a>

jodyheavener:
Closes #4467

No test changes as it's already covered [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/tests/spec/views/settings/delete_account.js#L531) and [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/tests/functional/delete_account.js).

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
@@ -61,6 +61,11 @@
         </li>
 
         <li class='delete-account-checkbox-list-item'>
+          <input id="delete-account-amo" type="checkbox" value="delete-account-amo" class="delete-account-checkbox" required />
+          <label for="delete-account-amo">{{#t}}Any extensions and themes that you published to addons.mozilla.org will be deleted{{/t}}</label>
+        </li>
+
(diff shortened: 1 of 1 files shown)


clouserw:
AMO isn't ready to land this yet and has some work ahead of them before it can be done.  Please reopen this when it's time.  Thank you.

PR #4535, opened 2020-03-12
fix(ci): changed diff flag on fxa-circleci build-ci.sh

dannycoates:
My local version of `diff` is 2.8.1 while the one run
in circleci/node:12 is 3.5 and they apparently treat
-b differently. Let's try -w. But there's still the
underlying question of why line 12 in the script
is changing the line endings from LF to CRLF in the
piped file?


dannycoates:
caught by my own trap 😂

at least i know it works now

PR #4536, opened 2020-03-12
fix(ci): changed diff method of fxa-circleci build-ci.sh

dannycoates:
My local version of `diff` is 2.8.1 while the one run
in circleci/node:12 is 3.5 and they apparently treat
-b differently. Let's try -w. But there's still the
underlying question of why line 12 in the script
is changing the line endings from LF to CRLF in the
piped file?

Edit:
after ^ didn't work...

The `cat | diff` method worked for me locally but doesn't
on CircleCI. Instead copy the Dockerfile out of the latest
image and then diff against the local one.

### DIFF
--- a/packages/fxa-circleci/scripts/build-ci.sh
+++ b/packages/fxa-circleci/scripts/build-ci.sh
@@ -1,17 +1,20 @@
-#!/bin/bash -ex
+#!/bin/bash -e
 
 # In order to save time this script pulls the latest docker image and compares
 # it to the current Dockerfile. If there are no changes the build is skipped.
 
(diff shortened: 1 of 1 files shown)


dannycoates:
This is weird...

```sh
+ docker run --rm -it mozilla/fxa-circleci:latest cat /Dockerfile
+ diff -w Dockerfile -
1c1
< FROM circleci/node:12-stretch-browsers
---
> ^@^@FROM circleci/node:12-stretch-browsers
```

bbangert:
That's very weird.

PR #4552, opened 2020-03-13
fix(access-tokens): prune the access token index on set

dannycoates:
The access token index can grow too big since the only
chance to prune until now was when the full list was
requested, which doesn't happen often.

This change adds a limit config and prunes when a new
token is added if the index is over the limit. It makes
no effort to check expiry so in the worst case up to 10%
of the limit active tokens will be deindexed, but still active.

To limit the size of the initial pruning no more than limit
tokens will be deindexed per call.

r? @jrgm

### DIFF
--- a/packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
+++ b/packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
@@ -1,11 +1,32 @@
 local userId = KEYS[1]
 local tokenId = KEYS[2]
 local token = ARGV[1]
-local ttl = ARGV[2]
+local limit = tonumber(ARGV[2])
+local ttl = tonumber(ARGV[3])
+local maxttl = tonumber(ARGV[4])
 local args = {token}
+local toAdd = {tokenId}
 if ttl then
   table.insert(args, 'PX')
   table.insert(args, ttl)
 end
+local count = redis.call('scard', userId)
+if count > limit then
+  local popped = redis.call('spop', userId, math.min((count - limit) + math.floor(limit / 10), limit))
+  if #popped < limit then
+    -- only a little over. check for expired tokens
+    local values = redis.call('mget', unpack(popped))
+    for i, value in ipairs(values) do
+      if value then
+        -- token has not expired
+        table.insert(toAdd, popped[i])
+      end
+    end
+  else
+    -- a lot over. just delete them
+    redis.call('unlink', unpack(popped))
+  end
+end
 redis.call('set', tokenId, unpack(args))
-redis.call('sadd', userId, tokenId)
+redis.call('sadd', userId, unpack(toAdd))
+redis.call('expire', userId, maxttl)
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -354,6 +354,12 @@ const conf = convict({
         format: String,
         doc: 'Key prefix for access tokens in Redis',
       },
+      accessTokenLimit: {
+        default: 100,
+        env: 'ACCESS_TOKEN_REDIS_LIMIT',
+        format: Number,
+        doc: 'Maximum number of access tokens per account at any one time',
+      },
     },
     sessionTokens: {
       enabled: {
--- a/packages/fxa-auth-server/lib/oauth/db/index.js
+++ b/packages/fxa-auth-server/lib/oauth/db/index.js
@@ -31,7 +31,11 @@ class OauthDB {
       await scopes();
     });
     this.redis = redis(
-      { ...config.get('redis.accessTokens'), enabled: true },
+      {
+        ...config.get('redis.accessTokens'),
+        enabled: true,
+        maxttl: config.get('oauthServer.expiration.accessToken'),
+      },
       logger
     );
     Object.keys(mysql.prototype).forEach(key => {
--- a/packages/fxa-auth-server/lib/redis.js
+++ b/packages/fxa-auth-server/lib/redis.js
@@ -33,6 +33,8 @@ function rejectInMs(ms, err = new Error('redis timeout')) {
 class FxaRedis {
   constructor(config, log) {
     config.keyPrefix = config.prefix;
+    this.accessTokenLimit = config.accessTokenLimit;
+    this.maxttl = config.maxttl;
     this.log = log;
     this.redis = new Redis(config);
     this.timeoutMs = config.timeoutMs || 1000;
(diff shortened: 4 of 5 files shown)


packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
```diff
@@ -1,11 +1,16 @@
...
+local ttl = ARGV[3]
 local args = {token}
 if ttl then
   table.insert(args, 'PX')
   table.insert(args, ttl)
 end
+local count = redis.call('scard', userId)
+if count > limit then
+  redis.call('spop', userId, math.min((count - limit) + math.floor(limit / 10), limit))
+end
```
jrgm: I wasn't following your comment "It makes no effort to check expiry so in the worst case up to 10%
of the limit active tokens will be deindexed, but still active."

But, as an example, if limit = 100, and count = 180, then Math.min(180 - 100 + 10, 100) = 90 tokens, of 50% are removed. So how does the 10% number fit with these?

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
```diff
@@ -1,11 +1,16 @@
 local userId = KEYS[1]
 local tokenId = KEYS[2]
 local token = ARGV[1]
-local ttl = ARGV[2]
+local limit = tonumber(ARGV[2])
+local ttl = ARGV[3]
```
jrgm: If you're adding an argument to ARGV, you want to name this `setAccessToken_3.lua`, ya?

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
```diff
@@ -1,11 +1,17 @@
...
   table.insert(args, 'PX')
   table.insert(args, ttl)
 end
+local count = redis.call('scard', userId)
+if count > limit then
+  redis.call('spop', userId, math.min((count - limit) + math.floor(limit / 10), limit))
+end
 redis.call('set', tokenId, unpack(args))
 redis.call('sadd', userId, tokenId)
+redis.call('expire', userId, math.max(ttl, 259200))
```
jrgm: Can you add a comment to call out what the magic number is?

(The intent is to eventually lower this TTL, and if so, max 3 days is still okay. But if we went in the other direction for some unforeseen reason, this would be wrong. Maybe pass in from the config?)

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
```diff
@@ -1,11 +1,17 @@
 local userId = KEYS[1]
 local tokenId = KEYS[2]
 local token = ARGV[1]
-local ttl = ARGV[2]
+local limit = tonumber(ARGV[2])
+local ttl = tonumber(ARGV[3])
```
jrgm: Hmm, I made a comment on this but it got lost. 

If adding args, this filename should become `_3.lua`, ya?

jrgm: Narrator's voice: It did not get lost.

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
dannycoates: 10% in the steady state case, where count never gets over 101. in the initial overflow case it'll remove up to `limit` so yes, 90 in your case.

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
dannycoates: no, only `KEYS` requires the name change

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
dannycoates: yeah, i was trying to avoid piping that config variable through. I'll add a note

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
jrgm: Ah, ok.

packages/fxa-auth-server/lib/luaScripts/setAccessToken_2.lua
jrgm: Duh.

dannycoates:
@rfk deindex or delete? When someone hits the limit on access tokens we're going to prune the index, but I'm trying to decide if we delete those tokens or just deindex and let them expire on their own.

The advantage of deindexing is that those tokens are still valid and can be used until they naturally expire, so less impact on clients. The disadvantage is that now those tokens can't be explicitly deleted by userId, which is virtually never, but important if our access tokens are long lived. If they're short, like we'd like to get to, it shouldn't matter. I guess the question is, do we consider our current token lifetime short or long?

rfk:
> but I'm trying to decide if we delete those tokens or just deindex and let them expire on their own.

I want to say "deindex but not delete" for operational reasons, but I think that has a small risk of unacceptable user-visible behaviour.

> The disadvantage is that now those tokens can't be explicitly deleted by userId, which is virtually
> never, but important if our access tokens are long lived.

Thinking out loud...

I don't we ever delete a specific access token by uid, but rather we delete all access tokens for a given uid and clientid. I'm wondering if there's something clever we could do here to get the best of both worlds. For example you could have the deletion operation set a "all tokens for uid X and clientID Y created before time T are now invalid" flag in redis...but then you'd have to check that flag on every token read. Maybe more complicated than it's worth.

(This is basically the `fxa-generation` scheme we use for managing this problem in tokenserver with BrowserID certificates).

> If they're short, like we'd like to get to, it shouldn't matter.
> I guess the question is, do we consider our current token lifetime short or long?

So I guess the failure case here is:

* A user takes some action to disconnect a client from their account (password change/reset, explicit disconnect).
* We try to delete all access tokens that might be held by that client, so that it gets disconnected immediately.
* With this change we might miss some access tokens that are not in the index, and the client will continue to be successfully connected to the user's account until those tokens expire.

We've had similar problems with BrowserID certificates in the past, where users notice (and quite rightly complain) that their browser remained connected for several hours after they took some action that should have disconnected it from their account.

We'd have a similarly-shaped problem if our oauth tokens were JWTs, as well.

We know that users notice weird behaviour around expire-vs-delete with BrowserID certificates. OAuth token lifetime is currently longer than BrowserID certificates. So I think that unfortunately, our current token lifetime needs to be considered "long" rather than "short".

dannycoates:
merged into master with #4569

PR #4553, opened 2020-03-13
chore(ci): improved deploy log output

dannycoates:
slightly improved the output of the deploy step in ci

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -1,45 +1,43 @@
 #!/bin/bash -e
 
 MODULE=$1
-DIR=$(dirname "$0")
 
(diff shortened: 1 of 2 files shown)


.circleci/deploy.sh
```diff
@@ -1,45 +1,43 @@
 #!/bin/bash -e
 
 MODULE=$1
-DIR=$(dirname "$0")
-
-if grep -e "$MODULE" -e 'all' "$DIR/../packages/test.list" > /dev/null; then
```
dannycoates: no need to grep `test.list` here anymore. we're going through all the packages anyway

PR #4557, opened 2020-03-14
chore(ci): improved ci build step output

dannycoates:
The build step ci output was really noisy with docker logs
so this writes that output into log files in the build
artifacts instead.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -18,15 +18,17 @@ if grep -e "$MODULE" -e 'all' "$DIR/../packages/test.list" > /dev/null; then
   mkdir -p config
   cp ../version.json config
 
+  mkdir -p ../../artifacts
+
   ODDBALLS=("fxa-auth-server" "fxa-content-server" "fxa-profile-server" "fxa-payments-server")
 
   if [[ -x scripts/build-ci.sh ]]; then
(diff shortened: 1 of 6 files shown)


.circleci/build.sh
```diff
@@ -18,15 +18,17 @@ if grep -e "$MODULE" -e 'all' "$DIR/../packages/test.list" > /dev/null; then
...
   if [[ -x scripts/build-ci.sh ]]; then
-    ./scripts/build-ci.sh
+    time ./scripts/build-ci.sh
   elif [[ "${ODDBALLS[*]}" =~ ${MODULE} ]]; then
     cd ..
-    docker build -f "${MODULE}/Dockerfile" -t "${MODULE}:build" .
+    time docker build --progress=plain -f "${MODULE}/Dockerfile" -t "${MODULE}:build" . > "../artifacts/${MODULE}.log"
   elif [[ -r Dockerfile ]]; then
-    docker build -f Dockerfile -t "${MODULE}:build" .
+    time docker build --progress=plain -t "${MODULE}:build" . > "../../artifacts/${MODULE}.log"
```
bbangert: Ah nice! Too bad there's no way to propagate this somewhere so we can compare over time. Maybe later?

PR #4560, opened 2020-03-16
feat(subscriptions): implement idempotency when creating a subscriptions

### LINKED ISSUE #4516: Implement idempotency key in subscription & payment update flows?
In PR #4476 / Issue #4386 / FXA-1275 the payment form was given logic to guard against multiple submit clicks using a nonce. This nonce is intended to also be used as an [idempotency key](https://stripe.com/docs/api/idempotent_requests?lang=node) for the Stripe API request.

Using the nonce as an idempotency key will require changes:
* support it as a parameter in auth-server APIs
* accept & send it as a parameter from payment-server API client
* send the nonce as idempotency key from payment form
* upgrade the nonce to use [uuidv4](https://github.com/uuidjs/uuid)

Alternatively, we could decide that the idempotency key thing is unnecessary or overkill. In that case, we should:

* drop the nonce generation logic
* switch to a simple `isSubmitting` boolean flag in the payment form rather than matching nonce changes

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1350)
┆Issue Number: FXA-1350

jodyheavener:
Closes #4516

This centers around [Stripe's support for idempotency keys](https://stripe.com/docs/api/idempotent_requests) when performing POST requests against its API, and requires some changes in a few spots:

- Update the payment form to submit a nonce as the idempotency key (not used for creating tokens, though)
- Handle it in Payments Server, update `ClientApi.createSubscription` to support it
- Receive the key on the auth server and pass it to Stripe with each "create" request, this includes customer and subscription creation

Additionally, this is backed by replacing our previous nonce token generation from `${Date.now()}-${Math.random()}` to proper UUID v4 for _max_ entropy.

---

Note - I see the Payments Server uses Storybook. I don't think I've touched much of it, so I can understand if I'm missing some required changes there. 🤔

### DIFF
--- a/packages/fxa-payments-server/src/lib/hooks.tsx
+++ b/packages/fxa-payments-server/src/lib/hooks.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useState, useEffect, useRef, ChangeEvent } from 'react';
+import { v4 as uuidv4 } from 'uuid';
 
 export function useCallbackOnce(cb: Function, deps: any[]) {
   const called = useRef(false);
@@ -70,9 +71,7 @@ export function useClickOutsideEffect<T>(onClickOutside: Function) {
   return insideEl;
 }
 
-export function useNonce(
-  generateNonce = () => `${Date.now()}-${Math.random()}`
-): [string, () => void] {
+export function useNonce(generateNonce = uuidv4): [string, () => void] {
   const [nonce, setNonce] = useState(generateNonce());
   const refreshNonce = () => setNonce(generateNonce());
   return [nonce, refreshNonce];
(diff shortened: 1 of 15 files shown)


packages/fxa-payments-server/src/lib/hooks.tsx
```diff
@@ -70,9 +71,7 @@ export function useClickOutsideEffect<T>(onClickOutside: Function) {
   return insideEl;
 }
 
-export function useNonce(
-  generateNonce = () => `${Date.now()}-${Math.random()}`
-): [string, () => void] {
+export function useNonce(generateNonce = () => uuidv4()): [string, () => void] {
```
lmorchard: Nit: this probably could just be simply
```
export function useNonce(generateNonce = uuidv4): [string, () => void] {
```
But, no biggie.

PR #4564, opened 2020-03-16
chore(metrics): update calculation of auth server route perf metrics

### LINKED ISSUE #4429: [auth server] Ensure Flow Performance and StatsD Metrics Report the Same Values
Because people are looking at [https://earthangel-b40313e5.influxcloud.net/d/J81nRFfWz/auth-server](https://earthangel-b40313e5.influxcloud.net/d/J81nRFfWz/auth-server) instead of the bottom of https://sql.telemetry.mozilla.org/dashboard/fxa-performance

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1296)
┆Issue Number: FXA-1296

chenba:
This patch makes two updates to the auth server's route perf metrics:
 - Hapi's `request.info.responded` timestamp is now used to calculate
   the duration
 - The identical value is reported to Redshift and InfluxDB

Fixes #4429 (FXA-1296)

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/events.js
+++ b/packages/fxa-auth-server/lib/metrics/events.js
@@ -224,7 +224,7 @@ module.exports = (log, config) => {
     return log.flowEvent(
       Object.assign({}, data, {
         event: `route.performance.${path}`,
-        flow_time: Date.now() - request.info.received,
+        flow_time: request.info.completed - request.info.received,
       })
     );
   }
--- a/packages/fxa-auth-server/lib/server.js
+++ b/packages/fxa-auth-server/lib/server.js
@@ -265,8 +265,6 @@ async function create(
     return h.continue;
   });
 
-  const metricReporter = metricFactory(statsd);
-
   server.ext('onPreResponse', (request, h) => {
     let response = request.response;
     if (response.isBoom) {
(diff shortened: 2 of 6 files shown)


packages/fxa-auth-server/lib/server.js
```diff
@@ -439,11 +440,16 @@ function metricFactory(statsdClient) {
...
     }
 
-    statsdClient.timing('url_request', Date.now() - startDate.getTime(), 1, {
-      path: normalizePath(path),
-      method: request.method.toUpperCase(),
-      statusCode,
-    });
+    statsdClient.timing(
+      'url_request',
+      request.info.responded - request.info.received,
```
6a68: Could this ever be invoked for an error response? As of [Hapi 18](https://github.com/hapijs/hapi/issues/3871), `request.info.responded` is only set for success responses, and set to 0 for error responses, while `request.info.completed` is always set to a timestamp that reflects the processing time. I think we're okay here but not totally sure.

packages/fxa-auth-server/lib/metrics/events.js
```diff
@@ -224,7 +224,7 @@ module.exports = (log, config) => {
     return log.flowEvent(
       Object.assign({}, data, {
         event: `route.performance.${path}`,
-        flow_time: Date.now() - request.info.received,
+        flow_time: request.info.responded - request.info.received,
```
6a68: In this case, I think the use of `request.info.responded` is fine, since this is only invoked if the status is 2xx (line 166). Might be worth documenting that this function expects success responses, in case it's reused to log error info in the future?

packages/fxa-auth-server/lib/server.js
chenba: I think "when response transmitted successfully" means that the client didn't close the socket before all the of the response was sent, not successful HTTP status.  But since `request.info.completed` is what we really need anyway, I'll switch to that.

Thanks for the great catch!

PR #4565, opened 2020-03-16
refactor(env): implement ability to supply your secret and publishable Stripe keys in a `local.env` file

### LINKED ISSUE #4483: Allow for a gitignored file to contain local Stripe API keys
We currently need to modify the auth server config and payments server config (or development.json) files to insert our Stripe keys. We have to consciously not commit our changes to these files.

It would be nice to have a root-level .gitignored .env file (or a js file like `locals.js`) to pull in and populate these vars so development can become a little easier.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1333)
┆Issue Number: FXA-1333

jodyheavener:
Closes #4483

**Problem**: developers need to manually modify Auth and Payment server files with their own Stripe keys, which results in changes to tracked files in git. It has resulted in accidental commits of these secrets, and of course don't want to be committing secrets! It's also just a pain to be updating these because of how spread out they are.

~**Solution**: add the ability to create a file called `local.env` at the root level of this repo, where you can supply your Stripe publishable and secret keys, so they can be supplied to the servers that require them.~

**Solution:** in places where we use convict we're already checking and loading in a file that is set at the `CONFIG_FILES` env var, so I've gone ahead and updated the pm2 server script to automatically check for a specific JSON file for two of our servers, so we can load in Stripe credentials. I've also documented this in the root and individual package READMEs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ public-key.json
 secret-key.json
 secret*
 *.gpg
+secrets.json
 
 # Build and temp files
 **/build
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ The Firefox Accounts (fxa) monorepo
 [Getting Started](#getting-started)\
 [Contributing](#contributing)\
 [Dependencies](#dependencies)\
+[Secrets](#secrets)\
 [Firefox Custom Profile](#firefox-custom-profile)\
 [Functional Tests](#functional-tests)\
 [Node debugging](#node-debugging)\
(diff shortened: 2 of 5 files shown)


dannycoates:
a rebase on master ought to fix the content-server tests

.gitignore
```diff
@@ -8,6 +8,7 @@
 
 # Secrets
 .env
```
LZoog: One thing of note, doesn't have to be fixed here but might should be a new issue - we use a `.env` file in the `fxa-payments-server` server as well as the `fxa-admin-panel` package. It's for local dev but we might not want this at the root-level, I had to force add it when making `-panel`. This is how we're proxying an express server to the webpack dev server.

jodyheavener: Just want to make sure I understand - your suggestion is to not use a single .env at the root level, and instead use multiple at the package level to avoid issues with server config?

dannycoates: I agree that, at least for now, the `.env` files should be per service since it matches how we're configuring them now. In the future I'd like to consider using a single config for all services.

jodyheavener: Makes sense. Reminder that we're not tied to .env - this could be a JSON/JS file if we prefer. I believe .env assumes you could pass variables in when starting the server, which is not the case here, it's simply just parsing the text content of the env file.

jodyheavener:
@dannycoates oh hell ya I'm doing that. Stand by.

PR #4568, opened 2020-03-17
fix(recovery key): update post-verify recovery key styling to not look like a modal on mobile

### LINKED ISSUE #4547: “Save recovery key” page is displayed as modal on Android
**Build:**
Stage [Train 1.162.0]

**Affected platforms:**
Android 9

**Steps to reproduce:**
1. Create and confirm a new Firefox account.
2. Click on "Get a recovery key" button from the email.
3. Click on "Get a recovery key" button.
4. Insert the account password and click “Continue” button.
5. Observe the “Save recovery key” page.

**Expected results:**
“Save recovery key” step is displayed as an integrated page.

**Actual Results:**
All the steps from “Get a recovery key” flow appear as integrated pages except the “Save recovery key” which is displayed as a modal.

mdoal

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1376)
┆Issue Number: FXA-1376

jodyheavener:
Closes #4547

This issue was introduced by me in 3332874bce6bac3679b49b522cebcf1094609288.

Instead of modifying the style of the success banner to work like the variety that lives in the modal, I've updated the Javascript that gets called when a copy-print-download button is clicked to target our success banner.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/post_verify/account_recovery/save_recovery_key.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/post_verify/account_recovery/save_recovery_key.mustache
@@ -1,4 +1,4 @@
-<div id="main-content" class="card save-recovery-key modal">
+<div id="main-content" class="card save-recovery-key">
     <header>
         <h1 id="fxa-save-recovery-key-header">
             {{#serviceName}}
@@ -11,8 +11,8 @@
         </h1>
     </header>
 
-    <section class="modal-panel">
-        <p class="modal-success bottom-spaced"></p>
+    <section>
+        <p class="success"></p>
 
         <p class="description">Store it in a safe place on your computer or a printed copy that is easily accessed by only you.</p>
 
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/views/post_verify/account_recovery/save_recovery_key.js
```diff
@@ -52,6 +52,18 @@ class SaveRecoveryKey extends FormView {
     this.print(PrintTemplate({ recoveryKey: this.recoveryKey }));
   }
 
+  // This method typically targets a similar success banner
+  // whose styling and visibility depends on it being contained
+  // inside a modal. Since we don't want that here the more
+  // straightforward approach is to just override this method
+  // to target our other styled success banner.
+  _displaySuccess(msg) {
```
vbudhram: I think we can pull this logic into https://github.com/mozilla/fxa/blob/8701348cdd79dbdc9879b2b4a55a23a135a32bc1/packages/fxa-content-server/app/scripts/views/mixins/save-options-mixin.js#L85 so that it lives in one place. Depending if modal class exists it would apply the correct styles.

Issue #3987, opened 2020-01-30
Show users correct subscription cycle information on payments pages


data-sync-user:
As a user who has elected to buy a longer duration subscription, I would like to see the cost, and billing cycle accurately reflected on the payment page for my product, so that I can verify my subscription choice.

Tasks:

- [x] Update Plan Response - [Jira SubTask FXA-1359](https://jira.mozilla.com/browse/FXA-1359) - PR https://github.com/mozilla/fxa/pull/4554

- [x] Update Payments Page Pricing Header - [Jira SubTask FXA-1360](https://jira.mozilla.com/browse/FXA-1360) - PR https://github.com/mozilla/fxa/pull/4573

- [x] Update Legal Checkbox Text - [Jira SubTask FXA-1361](https://jira.mozilla.com/browse/FXA-1361) - PR https://github.com/mozilla/fxa/pull/4615

- [x] Update Settings Page - [Jira SubTask FXA-1362](https://jira.mozilla.com/browse/FXA-1362) - PR https://github.com/mozilla/fxa/pull/4637

- [x] Update Tiers Page Pricing - [Jira SubTask FXA-1363](https://jira.mozilla.com/browse/FXA-1363) - https://github.com/mozilla/fxa/pull/4638

- [x] Update Tiers Page Legal - [Jira SubTask FXA-1364](https://jira.mozilla.com/browse/FXA-1364) - PR https://github.com/mozilla/fxa/pull/4618

**Details**
Payments page main pricing header

    Monthly: Example Product billed $X.XX monthly
    Six Months: Example Product billed $X.XX every 6 months
    Yearly: Example Product billed $X.XX yearly
    3 Years: Example Product billed $X.XX every 3 years

Legal Checkbox text

    Monthly: to charge my payment method $X.XX monthly
    Six Months: to charge my payment method $X.XX every 6 months
    Yearly: to charge my payment method $X.XX yearly
    3 Years: to charge my payment method $X.XX 3 years

Settings Page

    Monthly: You are billed $X.XX monthly for Example Product.
    Six Months: You are billed $X.XX every 6 months for Example Product.
    Yearly: You are billed $X.XX yearly for Example Product.
    3 Years: You are billed $X.XX every 3 years for Example Product.

Tiers Page Pricing

    Monthly: $X.XX monthly
    Six Months: $X.XX every 6 months
    Yearly: $X.XX yearly
    3 Years: $X.XX every 3 years

Tiers page legal

[...]

PR #4573, opened 2020-03-17
feat(payments): update messaging used for payments page main pricing header

jackiemunroe:
- updated messaging to allow for non-single interval plan details by using the interval_count data

issue #3987


packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -38,7 +38,26 @@ privacy = Privacy Notice
...
+}
+
+month-based-plan-details-amount = { $intervalCount ->
+  [one] { $productName } billed ${ $amount } monthly
+  *[other] { $productName } billed ${ $amount } every {$intervalCount} months
+}
+
+year-based-plan-details-amount = { $intervalCount ->
+  [one] { $productName } billed ${ $amount } yearly
+  *[other] { $productName } billed ${ $amount } every {$intervalCount} years
```
jodyheavener: This is probably pretty unlikely, but do we want to conditionally pluralize the unit in case `$intervalCount` is ever `1`? e.g. `every 1 months` could be awkward looking.

jackiemunroe: If `$intervalCount` is 1, then it will use the `[one]` case and be: `... billed $5.00 yearly`. You can see the verification in the tests. There is a test for equal to one and greater than 1 for each of the translations added.

jackiemunroe: https://github.com/mozilla/fxa/pull/4573/files#diff-8466bb64c869abe0b93953b4ce769fe3R246

jodyheavener: Ahh, got it. I misunderstood how this was set up. Carry on!

PR #4578, opened 2020-03-18
chore(docs): added dependency update log

dannycoates:
Here's an idea for consideration. We talked about making it the release owner's duty to set up the next train with a round of dependency updates, time boxed to ~1h. Here I'm proposing part of that PR should be an update to this document. The intent is to give future updaters an idea of what was last done, when, and any notes on things left behind.

### DIFF
--- a/docs/dependency-log.md
+++ b/docs/dependency-log.md
@@ -1,0 +1,81 @@
+# Dependency Update Log
+
+This is a living document. It summarizes the results of the last round of updates for each package. When you update it please change the **Last Update** section at the top and replace the dated sections for each package you updated. The dated sections should include the results of `npm outdated --long` from _after_ you've updated, to show which packages are still behind and any notes for future updaters.
+
+## Last Update
+
+- Date: 2020-03-18
+- Packages: fxa-shared
+- Updater: dannycoates
+
+# Packages
+
+## root
+
+### 2020-03-12
+
+updated all
+
+## 123done
+
+## browserid-verifier
+
+## fortress
+
+## fxa-admin-panel
+
+## fxa-admin-server
+
+## fxa-auth-db-mysql
+
+## fxa-auth-server
+
+## fxa-content-server
+
+## fxa-customs-server
+
+## fxa-dev-launcher
+
+## fxa-email-event-proxy
+
+## fxa-event-broker
+
+## fxa-geodb
+
+## fxa-js-client
+
+## fxa-metrics-processor
+
+## fxa-payments-server
+
+## fxa-profile-server
+
+## fxa-shared
+
+### 2020-03-18
+
+exceptions:
+
+```
+Package          Current  Wanted  Latest  Location    Package Type  Homepage
+accept-language   2.0.17  2.0.17  3.0.18  fxa-shared  dependencies  https://github.com/tinganho/node-accept-language
+celebrate         10.1.0  10.1.0  12.0.1  fxa-shared  dependencies  https://github.com/arb/celebrate#readme
+node-uap           0.0.3     git     git  fxa-shared  dependencies  https://github.com/fedot/node-uap
+redis              2.8.0   2.8.0   3.0.2  fxa-shared  dependencies  https://github.com/NodeRedis/node_redis
+```
+
+- accept-language
+  - we're on an ancient version
+  - no changelog so it's unclear if there's any breaking changes
+  - no known vulnerabilities so leaving as is
+- celebrate
+  - requires joi@17
+  - probably needs schema changes
+- node-uap
+  - we're on a git fork
+  - we should evaluate other options
+- redis
+  - did not attempt update
+  - not updating because our larger redis strategy may change in the near future
+
+## fxa-support-panel


lmorchard:
Not sure I like this as an added job for release. On one hand, I thought we were working toward semi-automatic releases without owners. On the other hand, I'm not sure I like adding another job atop tracking a release.

I'd rather have us file issues for dependency upgrades. Maybe time box the process of identifying what needs upgrades & filing the issues. Then, triage & plan them as part of the normal sprint process.

Otherwise, it's a hobby that a rotating solo someone has to carve extra time out for - and that never quite seems to go well. Some upgrades are fast, some take code changes, some take a good while of head scratching.

Might be nice if we had a regular report from something like [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates) as a source for issues

dannycoates:
A rotation doesn't need to be a permanent solution but we need some form of process now. It shouldn't just fall on whoever it bugs the most. Release owner is convenient because it's already a rotation, after tagging is a good time to do it, and is less of a diversion for them than anyone else at that time.

> I'd rather have us file issues for dependency upgrades. Maybe time box the process of identifying what needs upgrades & filing the issues. Then, triage & plan them as part of the normal sprint process.

That's what this is intended to be, with the addition that some updates get done as part of the process and we keep track of what/when things were last done. The update notes can be links to the filed issues.

6a68:
Counter-proposal:

* Dependency updates can easily take longer than an hour if there are weird CI failures.
* Release management is already an unpleasant and time-consuming solo chore.
* Sharing unpleasant chores among 2 people can make it nicer than going it alone.

What if we rotate dependency update duties separately from release manager, and those two can work together? Pick the highest-importance dependencies to update, open one PR per week/sprint, make sure it lands with tests passing before the release is tagged.

dannycoates:
It could be a separate duty from release owner.

I did both this last week as a trial run and it wasn't too bad. The timing actually worked out pretty well. I ran `./release.sh` and pushed the tag. While that was building in CI I ran `npm outdated` on the root of the repo. It probably has the smallest dep tree of the lot, so I thought I might get to another one too, but because I was curious I went through the dep list and searched for where each was used. It turned out there were some we could remove along with some disused scripts, so I spent some extra time pruning. After that I checked changelogs of major version updates and incremented those. By the time CI had finished deploying the tag I was about ready to submit a PR. Since I was waiting for the tag to finish it felt like doing this was a good use of time compared to trying to work on another issue and have to context switch.

I updated fxa-shared last night, because boredom/anxiety, and didn't get quite as far in an hour, there's still some deps on old versions, but I still made a decent dent. So progress will vary, but hopefully by doing it regularly there will be less to do over time.

I think if we only file issues they'll always be the first thing to be deferred and never be a priority unless it's a vulnerability. By allocating regular time for it we'll stay up to date and when high priority updates do happen they shouldn't be as painful.

clouserw:
I think we should be cautions about abandoning bots altogether.  They provide notification that a library has been upgraded, an easy way to comment on it (github comments), and, in a perfect case, an easy upgrade, although that doesn't always work.  Moving to a manual system feels a little bit like a step backwards (although I acknowledge I'm not the one in there dealing with PRs so this is an outside view).

If the problem is that PRs sit forever, let's make reviewing PRs a higher priority.  Les's point about identifying and filing issues, and triaging them as part of the normal sprint process makes a lot of sense to me.

We can leverage the release owner rotation even if it's not the release owner doing it.  For example, the previous week's release owner can be the "dependency owner" (or even the previous *two* people so there is load sharing).

I love the "turned out there were some we could remove" dependencies.  That seems worth putting some time in to.

I think having a bot running at least keeps it in the forefront of our minds and helps us identify where to start.  For example, a bot can open 2 PRs a week and the dependency owners are responsible to look at them.  They can either land them or investigate further, but should comment in the PR with any notes/progress.  We could also timebox dependency issues to, say, half-day per sprint.

vbudhram:
> I did both this last week as a trial run and it wasn't too bad. The timing actually worked out pretty well.

Now that most of the team has experience being a release owner, I think the overall time it takes will go down. We should be able to do release and some dependency updates within the timebox.

Here are some quick thoughts

My pros
* Keeps our dependencies updated
* CI build times have gotten faster so we don't have to wait as long to see dependency update failures
* I like [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) vs using a bot

My cons
* Don't like the idea of having another document to update manually in release
  * Maybe a bot/script can do this like we do with the changelog
* I rarely if ever review the dependabot PRs
  * Mostly because they don't really seem urgent

Overall, I think this is a good start and we can improve the process as we go along.

dannycoates:
wrt the doc itself:

The only reason to have a doc like this is if it helps us, our future selves, remember context. I don't like extra paper work. I really don't like having to remember or figure out *why* something is using a three year old version instead of the latest release every 👏 single 👏 time.

To me it's for the same reason we try to have good commit messages. In fact, if we put this info in the commit messages, and had a script that'd print out the commit history comments for every package-lock, then I'd be totally happy not having this doc.

chenba:
I'm okay with a rotation.  I like having the current or previous release owner handling dependency duties mainly just so that we'll have one fewer thing to track.

PR #4579, opened 2020-03-18
fix(experiment): update sendSmsHeader frontend experiment to include previous control option

jodyheavener:
This addresses a misunderstanding in the experiment where we neglected to use a "control". More can be found in [this Slack discussion](https://mozilla.slack.com/archives/C4D36CAJW/p1584452512384100).

Here are the three versions now:

`control`:

Screen Shot 2020-03-18 at 1 40 34 PM

`syncPhone`:

Screen Shot 2020-03-18 at 1 40 44 PM

`syncBrowser`:

Screen Shot 2020-03-18 at 1 40 55 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sms_send.mustache
@@ -9,21 +9,26 @@
     </header>
     <section class="send-sms">
       <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Connect another device{{/t}}"></div>
-      <h1 data-id="prompt-header">
-        {{^userHasAttachedMobileDevice}}
-          {{#isInExperimentGroupSyncBrowser}}
-            {{#t}}Sync this browser with your phone{{/t}}
-          {{/isInExperimentGroupSyncBrowser}}
-          {{^isInExperimentGroupSyncBrowser}}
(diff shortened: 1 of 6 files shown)


packages/fxa-content-server/app/tests/spec/views/sms_send.js
```diff
@@ -56,6 +58,9 @@ describe('views/sms_send', () => {
     model = new Backbone.Model({ account, showSuccessMessage: true });
     notifier = new Notifier();
     relier = new Relier({ service: 'sync' });
+    experimentGroupingRules = {
+      choose: () => 'syncPhone',
```
jodyheavener: 🤔 this makes tests pass locally, but not on CI

packages/fxa-content-server/app/scripts/templates/sms_send.mustache
```diff
@@ -9,21 +9,28 @@
     </header>
     <section class="send-sms">
       <div class="graphic {{graphicId}}" role="img" aria-label="{{#t}}Connect another device{{/t}}"></div>
+
       <h1 data-id="prompt-header">
```
LZoog: Since we're no longer _always_ displaying an `h1`, we'll want to move it inside the conditionals so a blank `h1` isn't rendered.

PR #4580, opened 2020-03-18
fix(redis): access token index ttl is in milliseconds, so use pexpire

dannycoates:


dannycoates:
this got merged in #4581

Issue #3430, opened 2019-11-20
Migrate content-server to react


dannycoates:
it's the future.

todo: convert to epic once we have more clarity

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-561)
┆Issue Number: FXA-561

lmorchard:
FWIW, this might be worth writing up as an ADR!

Also might be handy putting some thought into why React and not something else [like web components](https://briangrinstead.com/blog/firefox-webcomponents/), Svelte, Vue, etc. Obviously, I'm a fan of React and picked it for payments-server, but would be good to show our work in an ADR. (Or, maybe somehow even be surprised that React isn't the right choice?)

shane-tomlinson:
Oooh, when I think about a total conversion away from Backbone, I am filled with fear and dread. Putting elements on the screen isn't the actual problem with a full conversion, it's all the background infrastructure to support:

* A/B tests
* metrics
* integrations with browsers
* input validation
* query param validation
* screen->screen state machine depending on the integration being used
* account management

There was a *lot* of blood, sweat, and tears shed to get where we are.

I think the only reasonable way to do this sort of work is to split it into chunks. settings first since it doesn't have to deal with different integrations as much and screen to screen transitions are consistent across integrations. Or maybe the password reset flow since its so simple. Update the views while keeping backbone models to do some things like manage integrations and accounts.

Granted, the flow *is* getting much simpler with the replacement of links with codes and email-first only. Once all that work is complete and the tests are updated, AND if we actively deprecate support for old browsers (perhaps based on the browser version returned in the fxaccounts:fxa_status WebChannel message) the screen->screen state machine should be *significantly* simpler and a large scale conversion of the signin/signup flows will be significantly easier.

lmorchard:
> There was a lot of blood, sweat, and tears shed to get where we are.
> I think the only reasonable way to do this sort of work is to split it into chunks. 

Yeah, I think this is the right way to think about it. We should only re-implement with React if & where & when it simplifies things - and *definitely* with care to preserve the wisdom encoded in the system so far.

data-sync-user:
➤ Jared Hirsch commented:

We're still talking about this, but not in this bug. Closing

PR #4590, opened 2020-03-19
docs(adrs): (1) Transition FxA from Backbone to React, (2) Create a new React app for settings redesign

### LINKED ISSUE #3741: [ADR needed] Redesign kickoff infrastructure: React app?
There's been some discussion around whether or not the settings redesign could make sense to propel us into Reactland, which has been on everyone's minds for a while. This strongly relates to #3430, migrating the `fxa-content-server` to React. It's possible an ADR could be shared amongst these issues.

The outcome of this decision could also be impacted by Phase 1 / Phase 2 scope & priorities which is TBD at the time of writing, but I'd like to capture some thoughts from folks before the holidays. As the wires ([desktop](https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/398004229) / [mobile](https://mozilla.invisionapp.com/share/7FUIT2QWQRE#/screens)) stand today, it seems likely that "Plans and Billing" will not be included in Phase 1 unless we want a link-out to the application.

I wonder if this could be the ideal solution:
* Discuss and determine LoE to integrate payments into settings, considering
  * per @bbangert, at least "Update your Credit Card" requires a separate domain, but also potentially billing history - how can this be integrated nicely?
  * other engineering priorities and timelines - could this infrastructure work be done in Phase 1?
  * We probably need to identify underlying security concerns. We have an [ADR written for the payments server separation](https://github.com/mozilla/fxa/blob/master/docs/adr/0001-isolating-payment-content-with-third-party-widgets-from-general-account-management.md) that would be good to review. How's this affect CSP, as in payments we currently allow inline CSS styles due to iframes?
* Move forward with the settings updates, excluding payments (and other issues/features determined to be beyond Phase 1) in a React app

While the payments -> settings integration is a separate issue, it probably makes _less_ sense to move settings to a React app with this redesign if payments remains its own separate application.

@dannycoates has also suggested a "phase 0" that would restyle and rearrange our views to reflect something similar to the wires but keep things functionally identical. (Maybe the drawers could be all "open".) I will bring this up in consideration of conversations next week as well.

---

This list is no where complete but can get us started. Whatever we choose, shuffling around views = our tests are not going to be fun to get passing.

Pros:
- Gives us a better opportunity to better organize our queries hitting the settings page, which I think we have several issues open for
- A separate application could make rolling out to some users easier (A/B testing)
- We won't have to create or retrofit the new header and navigation in both Backbone + React, we can just write it in React
- Allows users to have a much more seamless, cohesive Accounts experience
- Propels us into a modern framework
- Storybook?

Cons:
- Time for initial setup & rollout
- Setup for payments restructuring

Approaches (which will need greater detail & planning):
- Create a new React app in `packages/fxa-content-server/fxa-settings` (I think `fxa` is pretty superfluous here, but🤷‍♀️), duplicate files as-needed, and begin [embedding React into the Backbone Views](https://reactjs.org/docs/integrating-with-other-libraries.html)
- There are likely additional approaches TBD as discussion ensues and I spend more time thinking about this.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-841)
┆Issue Number: FXA-841

LZoog:
Gives #3430 an ADR with "Transition FxA from Backbone to React"
Closes #3741 with "Create a new React app for settings redesign"

Ready for review/commentary. ~I still need to add links and will do one more read for proofreading.~

### DIFF
--- a/docs/adr/0010-transition-fxa-from-backbone-to-react.md
+++ b/docs/adr/0010-transition-fxa-from-backbone-to-react.md
@@ -1,0 +1,161 @@
+# Transition FxA from Backbone to React
+
+- Deciders: Lauren Zugai, Les Orchard, Danny Coates
+- Date: 2020-03-19
+
+## Context and Problem Statement
+
+When Firefox Accounts was created some 6-7 years ago, the front-end Javascript framework and HTML templating options were limited significantly to those available at the time. This ADR details the pros and cons of various FE libraries and frameworks, and why it is desirable for FxA to move to a different technology. It does not cover _how_ we will begin the conversion which can be [read about here](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md).
+
+While Backbone and Mustache have served their purposes for FxA thus far, other modern libraries and frameworks not only offer more appealing architecture and developer conveniences, but also offer better data predictability, potential performance boosts, and a strong community of online support.
+
+## Decision Drivers
+
+- The desire for a more modern FE architecture and a more efficient UI
+- As Backbone declines in popularity, the desire to move to a library or framework actively being improved with great community support is prevalent
+- A modern framework will likely feel more familiar to new engineers to the team as well as open source contributors
+
+## Considered Options
+
+- Option A - Continue to use Backbone
+- Option B - Move to client-side React
+- Option C - Move to isomorphic React
+- Option D - Move to Angular
+- Option E - Move to VueJS
+- Option F - Other considerations (Web Components)
+
+## Decision Outcome
+
+Chosen option: "Option B - Move to client-side React", because
+
+- The team desires the benefits we will gain from moving to a more modern library (excluding option A)
+- The team has experience with (and likes) React
+- We're already using it in other parts of the ecosystem, i.e. the payments server
+- FxA does not need server-side rendering (option C)
+- Angular (option D) is heavy, opinionated with a steep learning curve, and offers at least some features we won't need
+- Web components and libraries around them (option F) are quite new, yet a distant potential architecture shift from React to web components appears to be far less dramatic than from Backbone to React
+
+In the end, it is recognized that Vue _may_ have been a good option for FxA, but React was chosen over Vue due to:
+
+- existing FxA engineering team knowledge
+- the FxA payments server is already using React
+  - React has already been reviewed by our security team and "passed" the trial run when it was novel tech in our stack
+- As one of our desires is to best protect against such a major refactor proposal in the future, we have to allow package age, online support, popularity and backing, and stability to be factors in our decision.
+
+## Pros and Cons of the Options
+
+### Option A - Continue to use Backbone
+
+- Description
+  - Don't introduce a new framework and keep using Backbone and Mustache.
+- Pros
+  - No tech debt or large refactoring effort would be necessary right now.
+- Cons
+  - As Backbone's popularity continues to wane, we will likely be presented with challenges in the future around support, documentation, features, and developer knowledge or experience with the framework.
+  - We'd be giving a hard pass on all of the benefits we would gain from another library or framework - i.e. modern approaches, better data clarity and typing, performance optimizations, etc.
+
+Note: also see pros/cons listed on Backbone in general [in this ADR](https://github.com/mozilla/fxa/blob/master/docs/adr/0002-use-react-redux-and-typescript-for-subscription-management-pages.md)
+
+### Options B - E - Common Pros & Cons
+
+These pros and cons are shared for the rest of the choices. The pros and cons listed for all subsequent options will be in comparison to _each other_ and not against option A.
+
+- Description
+  - Move away from Backbone/Mustache in the content server in favor of another option.
+- Pros
+  - Future-proof the FE stack for the foreseeable future: modern libraries and frameworks have significant backings and resources placed to work on them (i.e. Angular by Google, React by Facebook) and are continuing to gain community and feature support.
+  - Benefit from the modern architectural approaches and features other technologies offer.
+  - Increase the chance of new engineers to the team as well as open source contributors being familiar with the stack.
+  - While refactoring, our tests can be assessed for need and clarity. A fresh pair of eyes on these can point out superfluous tests or those we lack.
+- Neutral
+  - Modern libraries and frameworks (React, Angular, VueJS) all offer some flavor of virtual DOM for rerenders and DOM manipulations, and offer performance boosts and optimizations out of the box that Backbone does not offer. To varying degree, we would expect to see a performance boost, but there may be cases in which a granular DOM update would be faster without a virtual DOM.
+- Cons
+  - We will gain tech debt regarding a large refactor effort that may be ongoing for some time.
+  - A lot of work was put into FxA and moving away from Backbone won't be a simple task. There is a lot of infrastructure we must ensure support for and respect the integrity of.
+  - A transition from Backbone to any of the options won't be 1:1 - that is to say, Backbone and Mustache closely resemble the common "MVC" pattern, where Backbone models tied to collections represent the "M", Backbone views act as the "C", and our Mustache templates act as the "V" layer. Angular and Vue appear to follow patterns more closely described as "MVVM", and while React can broadly be seen as the "V" in "MVC", React architecture typically follows the Flux design pattern. We will need to carefully translate pieces of Backbone models and controllers into state, helper functions, or Typescript interfaces where applicable.
+  - Whatever new library or framework we choose will require a learning curve for at least part of the team. It may slow the ability to make front-end changes at first.
+  - Since the timeline of a full transition is unknown, team members will have to learn or become familiar with both Backbone and the library or framework we transition to.
+
+### Option B - Move to client-side React
+
+- Description:
+  - Use client-side React.
+- Pros
+  - React is backed by a well-known company (Facebook) with a lot of dedicated resources working on the library and it's popular, has a lot of community support, and has ample documentation.
+  - The `fxa-payments-server` is already using React. We can begin thinking about reusing components across the FxA monorepo.
+  - React promotes simplicity, clarity, and predictability. We know that data exists at the top and it flows down to children components, and we know what the data is with prop and type checking. The component then takes the data and subjects it to lifecycle methods that execute in a predictable manner every time. Backbone, on the other hand, creates a JSON object that is exposed to mustache files, but we don't always know what this data is or what its types are. Backbone also oftentimes influences views via updates through a model, multiple models, and/or controllers. It's not as clear when a view or its children will be rerendered or a DOM element updated.
+  - React is fast. It bundles state updates together and figures out the fewest operations it can get away with, giving us optimization wins out of the box.
+  - React promotes declarative programming over imperative. In Backbone, we interact with the DOM directly via jQuery, _telling_ the browser exactly what to do (imperative) but with React, we _describe_ the UI with components and let React and the browser figure out how to implement it (declarative). This can be much less tedious for developers and allows React to handle state changes optimally.
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the [banana, monkey, jungle problem](http://rcardin.github.io/design/programming/oop/fp/2018/07/27/the-secret-life-of-objects-part-2.html#the-banana-monkey-jungle-problem)) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
+- Cons
+  - React changes fairly often. The definition of "knowing React" will continue to grow as new concepts and best practices, like React Hooks or deprecated lifecycle methods, are created or recommended.
+  - There have been previous issues with the licensing around React. Facebook switched several of its open-source projects, including React, [from a BSD+Patents license to an MIT license in 2017](https://www.freecodecamp.org/news/facebook-just-changed-the-license-on-react-heres-a-2-minute-explanation-why-5878478913b2/) after community pressure.
+  - While Facebook has held its dominance in the social media realm for more than 10 years, we cannot assume it will always be the leader in the industry and understand it may wane in popularity in the future. However, some risk around potential decreasing popularity is mitigated as Airbnb, Uber, Pinterest, Twitter, Reddit, Netflix, and other major companies also use React.
+
+### Option C - Move to isomorphic React
+
+#### Description
+
+Server-side rendering (SSR) is a technique for rendering a client-side single page app (SPA) on the server and then sending a fully rendered page to the client. It is one _part_ of an "Isomorphic" or "Universal" React application which are terms used synonymously.
+
+How a client-side React app typically operates:
+Browser requests page -> server sends mostly empty HTML with script tag(s) -> browser downloads page and requests scripts -> React takes over and fetches required data -> React receives and processes data and _displays content_
+
+How a SSR React app typically operates:
+Browser requests page -> server loads React in memory, fetches required data, renders React app, sends HTML down to the browser -> browser downloads page and _displays content_ -> Client React takes over
+
+- Pros over option B
+  - Less process burden on the client-side and improved First Meaningful Paint over client-only React gives a perceived performance boost.
+  - Offers much better SEO possibilities.\*
+    - \*This is one of the primary benefits SSR offers, yet is not applicable to FxA because our pages are hidden from search engines for security reasons.
+- Cons over option B
+  - Requires duplicate Redux stores and manually keeping sync in state with the server and client, or using a framework like NextJS.
+  - Adds another layer of complexity and a sharper learning curve to the tech stack.
+  - It may give a performance boost on slow devices since the server does more work, but it doesn’t actually solve a heavy initial network request problem.
+  - There is a very brief period between the initial render and React rehydrating in which if a user very quickly interacts with something, it may do nothing.
+
+### Option D - Move to Angular
+
+- Description
+  - Use AngularJS.
+- Pros
+  - Angular is backed by a well-known company (Google) with a lot of dedicated resources working on the framework and it's popular, has a lot of community support, and has good documentation.
+  - It provides some features we likely would use in FxA without another package dependency. Certain things may just "work" automagically.
+- Cons
+  - Angular has a steep learning curve.
+  - Since Angular is a full-on "framework," it results in a heavier bundle and can result in a slower application compared to React or Vue.
+  - It provides many features we likely wouldn't use in FxA.
+  - Angular is opinionated and expects things to be done in specific ways. While this could make it easier for a developer already familiar with Angular to ramp up on the project, this could set limitations on how we want to architect our application.
+  - Significant changes have historically been introduced to the framework.
+
+### Option E - Move to VueJS
+
+- Pros:
+  - Vue is used proudly by the community of developers that support it. The enthusiasm for open-source and attitude around making something "for developers" aligns with the Mozilla mission.
+  - Vue is flexible - it allows app structure as seen fit and supports many different HTML templating languages.
+  - Likely has the smallest learning curve.
+- Neutral:
+  - Allows for mutable state.
+- Cons:
+  - Vue is the newest library of the bunch, has a much smaller community than other options, and does not have a large company backing. It may not be as battle-tested as other options as a result or have as much documentation support when we run into specific issues.
+
+### Option F - Other considerations (Web Components)
+
+Web components were also considered. However, these are not a library, but are instead "a set of web platform APIs that allow you to create new custom, reusable, encapsulated HTML tags to use in web pages and web apps." They offer strong encapsulation and can be used in any JS framework or library that works with HTML, but don't meet the needs of FxA by themselves. See a more detailed description [with examples here](https://css-tricks.com/an-introduction-to-web-components/).
+
+Using web components with the [Polymer Project](https://www.polymer-project.org/) that allows for functional templates like React but without a virtual DOM could be worth keeping an eye on as the technology matures.
+
+## Links
+
+- [Create a new React app for Settings Redesign](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md)
+- Other Backbone pros/cons: [subscription management stack ADR](https://github.com/mozilla/fxa/blob/master/docs/adr/0002-use-react-redux-and-typescript-for-subscription-management-pages.md)
+- [Facebook switches licensing](https://www.freecodecamp.org/news/facebook-just-changed-the-license-on-react-heres-a-2-minute-explanation-why-5878478913b2/)
+- [Banana, monkey, jungle problem](http://rcardin.github.io/design/programming/oop/fp/2018/07/27/the-secret-life-of-objects-part-2.html#the-banana-monkey-jungle-problem)
+- [React](https://github.com/facebook/react)
+- [NextJS](https://github.com/zeit/next.js/)
+- [AngularJS](https://github.com/angular/angular.js/)
+- [VueJS](https://github.com/vuejs/vue)
+- [Web Components](https://css-tricks.com/an-introduction-to-web-components/)
+- [lit-html](https://lit-html.polymer-project.org/guide)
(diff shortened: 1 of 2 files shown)


docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
+# Transition FxA from Backbone to React
+
+- Deciders: Lauren Zugai, Les Orchard, Danny Coates
+- Date: 2020-03-19
+
+## Context and Problem Statement
+
+When Firefox Accounts was created some 6-7 years ago, the front-end Javascript framework and HTML templating options were limited significantly to those available at the time. This ADR details the pros and cons of various FE libraries and frameworks, and why it is desirable for FxA to move to different technology. It does not cover _how_ we will begin the conversion which can be [read about here](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md).
```
chenba: "...move to [a] different technology..."?

chenba: 404 on that link.

chenba: Hahaha ignore that 404 comment. 😆

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - React promotes declarative programming over imperative. In Backbone, we interact with the DOM directly via jQuery, _telling_ the browser exactly what to do (imperative) but with React, we _describe_ the UI with components and let React and the browser figure out how to implement it (declarative). This can be much less tedious for developers and allows React to handle state changes optimally.
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the banana, monkey, jungle problem) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
+- Cons
+  - React changes fairly often. The definition of "knowing React" will continue to grow as new concepts and best practices, like React Hooks or depreciated lifecycle methods, are created or recommended.
+
+### Option C - Move to isomorphic React
+
+- Description
```
chenba: Formatting issue.  "Description" is the first word of the paragraph.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - Adds another layer of complexity and a sharper learning curve to the tech stack.
+  - It may give a performance boost on slow devices since the server does more work, but it doesn’t actually solve a heavy initial network request problem.
+  - There is a very brief period between the initial render and React rehydrating in which if a user very quickly interacts with something, it may do nothing.
+
+### Option D - Move to Angular
+
+- Description
+  - Use AngularJS.
+- Pros
+  - Angular is backed by a well-known company (Google) with a lot of dedicated resources working on the library and it's popular, has a lot of community support, and has good documentation.
```
chenba: nit: s/library/framework/

docs/adr/0011-create-new-react-app-for-settings-redesign.md
```diff
@@ -0,0 +1,76 @@
...
+
+- Option A - Use the existing JS framework (Backbone) and templating library (Mustache), and make all changes in these files
+- Option B - Take the "100/0" approach by creating all new components with React, but modify existing Backbone/Mustache files
+- Option C - Take the "ground up" approach and create a new settings React application
+
+## Decision Outcome
+
+Chosen option: "Option C - Take the 'ground up' approach and create a new settings React application", because
+
+- The overall goal of the Settings Redesign Project is to offer more flexibility to expand Settings functionality. WeNew features have already been planned. It could be argued that heavy refactoring later (which would be needed for option A or B) would take just as much time as recreating this page with React now.
```
chenba: ~We~New ?

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - A transition from Backbone to any of the options won't be 1:1 - that is to say, Backbone and Mustache closely resemble the common "MVC" pattern, where Backbone models tied to collections represent the "M", Backbone views act as the "C", and our Mustache templates act as the "V" layer. Angular and Vue appear to follow patterns more closely described as "MVVM", and while React can broadly be seen as the "V" in "MVC", React architecture typically follows the Flux design pattern. We will need to carefully translate pieces of Backbone models and controllers into state, helper functions, or Typescript interfaces where applicable.
+  - Whatever new library or framework we choose will require a learning curve for at least part of the team. It may slow the ability to make front-end changes at first.
+
+### Option B - Move to client-side React
+
+- Description:
+  - Use client-side React.
+- Pros
+  - React is backed by a well-known company (Facebook) with a lot of dedicated resources working on the library and it's popular, has a lot of community support, and has ample documentation.
```
clouserw: are there any concerns when/as Facebook's popularity decreases?

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - React promotes declarative programming over imperative. In Backbone, we interact with the DOM directly via jQuery, _telling_ the browser exactly what to do (imperative) but with React, we _describe_ the UI with components and let React and the browser figure out how to implement it (declarative). This can be much less tedious for developers and allows React to handle state changes optimally.
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the banana, monkey, jungle problem) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
+- Cons
+  - React changes fairly often. The definition of "knowing React" will continue to grow as new concepts and best practices, like React Hooks or depreciated lifecycle methods, are created or recommended.
+
```
clouserw: Is React [significantly] worse regarding dependencies?  I thought I saw that in one of our analyses

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+## Links
+
+- [Create a new React app for Settings Redesign](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md)
+- Other Backbone pros/cons: [subscription management stack ADR](https://github.com/mozilla/fxa/blob/master/docs/adr/0002-use-react-redux-and-typescript-for-subscription-management-pages.md)
+- [React](https://github.com/facebook/react)
+- [NextJS](https://github.com/zeit/next.js/)
+- [AngularJS](https://github.com/angular/angular.js/)
+- [VueJS](https://github.com/vuejs/vue)
+- [Web Components](https://css-tricks.com/an-introduction-to-web-components/)
```
clouserw: These links aren't being used in the doc ^

docs/adr/0011-create-new-react-app-for-settings-redesign.md
```diff
@@ -0,0 +1,76 @@
...
+  - This approach will take much longer to rollout than previous options.
+
+## Links
+
+- [Settings Redesign Project PRD](https://docs.google.com/document/d/18zu7JCYIsUp8tUMJqb2uErNlzL9f6CQvefLy9HFZ4UY/edit?pli=1#heading=h.cf57dt1i8634)
+- [Backbone to React ADR](https://github.com/mozilla/fxa/blob/master/docs/adr/0010-transition-fxa-from-backbone-to-react.md)
+- [React](https://github.com/facebook/react)
+- [Backbone](https://backbonejs.org/)
+- [Mustache](https://github.com/mustache/mustache.github.com)
+- [NestedReact](https://github.com/VoliJS/NestedReact/blob/master/docs/05_Migration_from_Backbone.md)
```
clouserw: These links are also not being used

docs/adr/0010-transition-fxa-from-backbone-to-react.md
lmorchard: IMO, React *tooling* involves a lot of dependencies, but that's largely code we don't deploy. (i.e. devDependencies vs dependencies)

docs/adr/0011-create-new-react-app-for-settings-redesign.md
```diff
@@ -0,0 +1,76 @@
...
+- Option B - Take the "100/0" approach by creating all new components with React, but modify existing Backbone/Mustache files
+- Option C - Take the "ground up" approach and create a new settings React application
+
+## Decision Outcome
+
+Chosen option: "Option C - Take the 'ground up' approach and create a new settings React application", because
+
+- The overall goal of the Settings Redesign Project is to offer more flexibility to expand Settings functionality. WeNew features have already been planned. It could be argued that heavy refactoring later (which would be needed for option A or B) would take just as much time as recreating this page with React now.
+- This approach allows us an opportunity to set up Storybook for the content server and review our tests, CSS, and a11y implementations as we implement the UI changes.
+- This approach also simplifies our GH workflow and A/B testing the entire redesign.
```
lmorchard: Would be nice to consider ways to keep both the existing Backbone and future React settings apps working in parallel. Beyond A/B tests,, we could invite alpha users to explicitly use either / both and see how things shape up as we implement.

docs/adr/0011-create-new-react-app-for-settings-redesign.md
```diff
@@ -0,0 +1,76 @@
...
+  - Settings doesn't have to manage different integrations like the signin/signup flows do. It seems like a reasonable place to begin our conversion regardless of a redesign.
+  - Allows for a fresh start and audit of our HTML semantics, CSS architecture, and accessibility practices as components are being built.
+  - Allows for test auditing and scrutiny. We can potentially remove superfluous tests or add those we lack.
+  - Offers an opportunity to set up and use Storybook in FxA. If we do a view-by-view build up, we can deploy changes somewhere for UX and other folks to see progress and request tweaks as features are built out.
+  - We can potentially use this as an opportunity to review and extract model and API code from the existing settings app into shared modules.
+  - Management in GitHub is easy, as changes can go directly into `master` and no feature flagging or branches held in isolation from `master` will be necessary. A/B testing will also likely be easier.
+- Cons
+  - A "ground up remake" is generally viewed as taboo. Changing such a large chunk of the front-end codebase doesn't come without its own set of risks.
```
lmorchard: Yeah, I think we can keep this in mind and mitigate by avoiding a "big bang" surprise replacement. Try to implement and deploy and launch smaller pieces of MVP functionality along the way while keeping the Backbone settings live. Might get tedious at points, but it can be a safety net against losing wisdom earned in the original system

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
+# Transition FxA from Backbone to React
+
+- Deciders: Lauren Zugai, Les Orchard, Danny Coates
+- Date: 2020-03-19
+
+## Context and Problem Statement
+
+When Firefox Accounts was created some 6-7 years ago, the front-end Javascript framework and HTML templating options were limited significantly to those available at the time. This ADR details the pros and cons of various FE libraries and frameworks, and why it is desirable for FxA to move to different technology. It does not cover _how_ we will begin the conversion which can be [read about here](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md).
+
+While Backbone and Mustache have served their purposes for FxA thus far, other modern libraries and frameworks not only offer more appealing architecture and developer conveniences, but also offer better data predictability, potential performance boosts, and a strong community of online support.
```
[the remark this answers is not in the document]
LZoog: Good point. I think it's worth mentioning here that auth server uses mustache.

Might be good as a new issue with a `needs:discussion` label. It could be that it's fine TBH, but I've definitely found the _totally_ logic-less nature of mustache in the content server a little cumbersome. Probably depends on how annoying or limiting it is there. 😅

docs/adr/0010-transition-fxa-from-backbone-to-react.md
LZoog: @clouserw That's a good consideration. I think React's current very high level of popularity can help mitigate some risk around that, but it's hard to predict what will happen (-cough- MySpace). I'll add a comment around this somewhere.

@lmorchard I actually have web components in the "Other considerations" option. Are you talking about using web components with a web component library like [stencilJS](https://stenciljs.com/)? I'm having a hard time picturing how we could get there otherwise, they seem too... simplistic?

docs/adr/0010-transition-fxa-from-backbone-to-react.md
LZoog: I'll look into this further, but my initial thoughts mirror Les's.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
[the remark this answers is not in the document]
chenba: > ... I wonder if there's another template system we should use someday ...

The popularity argument can be made in mustache's favor here.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the banana, monkey, jungle problem) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
+- Cons
+  - React changes fairly often. The definition of "knowing React" will continue to grow as new concepts and best practices, like React Hooks or depreciated lifecycle methods, are created or recommended.
+
+### Option C - Move to isomorphic React
+
+- Description
+  Server-side rendering (SSR) is a technique for rendering a client-side single page app (SPA) on the server and then sending a fully rendered page to the client. It is one _part_ of an "Isomorphic" or "Universal" React application which are terms used synonymously.
```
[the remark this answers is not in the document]
LZoog: Yeah, the layer of complexity it adds for us is not worth what we'd get out of it IMO. 

Ben and I had floated around the idea of introducing GQL with the redesign but there's already so many moving parts that it wasn't talked about at length. 🤷‍♀️

docs/adr/0010-transition-fxa-from-backbone-to-react.md
LZoog: Other ADRs have "additional info" links at the bottom for links that aren't used in the doc. I'm tempted to leave these as-is, but I could remove them or add them where mentioned within the ADR.

docs/adr/0011-create-new-react-app-for-settings-redesign.md
LZoog: Thanks @chenba for catching all my typos 😂

docs/adr/0011-create-new-react-app-for-settings-redesign.md
LZoog: Ooo, I like this idea.

This might be good to discuss in a chat on best approach, I figured I could add a route (I haven't thought of the name yet, `settings_v2`?) and serve the React app from that for dev at least.

I sent this comment to Alex, who responded with:

> Yeah. But we would need two things for it to be really worth it:
> * A way to capture feedback
> * Instrument metrics like crazy on the new one vs the old one.

Which are good points.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
lmorchard: More something like Web Components plus [lit-html](https://lit-html.polymer-project.org/guide) / LitElement, which lets you do functional templates like React but doesn't use a virtual DOM. I don't think we should adopt it now, but it could be worth watching

docs/adr/0010-transition-fxa-from-backbone-to-react.md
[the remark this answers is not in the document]
jodyheavener: Mustache is probably fine for emails, but in case you're looking for a template language that is similar in syntax to Mustache, but slightly smarter (not by a lot, but far less annoying than Mustache) and more actively maintained, might I suggest [Liquid](https://github.com/docs/liquid).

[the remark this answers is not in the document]
jodyheavener: Of course if you wanted to use React wholesale, [this looks neat](https://github.com/chromakode/react-html-email).

docs/adr/0011-create-new-react-app-for-settings-redesign.md
jodyheavener: Super on board with slow rollout of React pages vs existing.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - React promotes declarative programming over imperative. In Backbone, we interact with the DOM directly via jQuery, _telling_ the browser exactly what to do (imperative) but with React, we _describe_ the UI with components and let React and the browser figure out how to implement it (declarative). This can be much less tedious for developers and allows React to handle state changes optimally.
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the banana, monkey, jungle problem) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
+- Cons
+  - React changes fairly often. The definition of "knowing React" will continue to grow as new concepts and best practices, like React Hooks or depreciated lifecycle methods, are created or recommended.
```
vbudhram: In practice how big have you seen these changes to be? Would we have to do an ADR to potentially upgrade to a new React version because of breaking changes etc?

docs/adr/0011-create-new-react-app-for-settings-redesign.md
vbudhram: > Instrument metrics like crazy on the new one vs the old one.

Luckily the settings page is not fully instrumented with metrics and has way less then on login/signup flows. 

It might allow us to take a fresh look at how we could do metrics in settings.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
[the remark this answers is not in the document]
LZoog: I am actually more bothered around the horrendous nature of our CSS in emails than templating language. We use `!important` everywhere.

I've never thought about Liquid outside of Shopify, lol. I was thinking EJS could also be a good one if we need logic, but since I have worked in emails sparingly, I'm not sure if it's worth the effort. Mustache seems to be sufficient for our needs.

I created a follow-up issue here where we can further the discussion: https://github.com/mozilla/fxa/issues/4627

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+## Context and Problem Statement
+
+When Firefox Accounts was created some 6-7 years ago, the front-end Javascript framework and HTML templating options were limited significantly to those available at the time. This ADR details the pros and cons of various FE libraries and frameworks, and why it is desirable for FxA to move to different technology. It does not cover _how_ we will begin the conversion which can be [read about here](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md).
+
+While Backbone and Mustache have served their purposes for FxA thus far, other modern libraries and frameworks not only offer more appealing architecture and developer conveniences, but also offer better data predictability, potential performance boosts, and a strong community of online support.
+
+## Decision Drivers
+
+- The desire for a more modern FE architecture and a more efficient UI
+- The desire to move away from "old tech" in favor of a library or framework actively being improved with great community support
```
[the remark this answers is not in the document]
LZoog: I agree, I've modified it to read "As Backbone declines in popularity, the desire to move to a library or framework actively being improved with great community support is prevalent." It's less hand-wavy that way with "old tech."

docs/adr/0010-transition-fxa-from-backbone-to-react.md
```diff
@@ -0,0 +1,151 @@
...
+  - React is fast. It bundles state updates together and figures out the fewest operations it can get away with, giving us optimization wins out of the box.
+  - React promotes declarative programming over imperative. In Backbone, we interact with the DOM directly via jQuery, _telling_ the browser exactly what to do (imperative) but with React, we _describe_ the UI with components and let React and the browser figure out how to implement it (declarative). This can be much less tedious for developers and allows React to handle state changes optimally.
+  - React promotes the idea of composition over inheritance. In Backbone, we extend views for reusability, which can cause difficulty in seeing or realizing what events and other methods or data you are inheriting (the banana, monkey, jungle problem) and overwriting. The intention of components in React is more clear and data is more deliberately passed around.
+  - React offers many integration options.
+  - React has already been audited by our security team.
```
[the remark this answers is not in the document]
LZoog: This is a good question, maybe @clouserw can answer. I'll send you a link in Slack to the GDoc our security team sent me about React.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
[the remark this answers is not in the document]
LZoog: Oooh, nice catch.

LZoog: I don't believe so - the largest change I've seen and have read about has been around the React 16 upgrade introducing hooks, which introduced "no breaking changes" (see [this section of the React docs](https://reactjs.org/docs/faq-versioning.html) for more info). When the last team I worked with did this upgrade, we did have to make a couple of minor changes since we were using React a little unconventionally, but it wasn't too much effort. They continue to support "older" ways of doing things, but we could potentially have some console warnings to clean up with upgrades.

docs/adr/0010-transition-fxa-from-backbone-to-react.md
LZoog: If we check out [React on NPM](https://www.npmjs.com/package/react) and [Backbone on NPM](https://www.npmjs.com/package/backbone), there's 3 dependencies vs 1. [`create-react-app`](https://www.npmjs.com/package/create-react-app) lists 11, but that appears to be because of development tools. There's an interesting conversation in [this thread](https://news.ycombinator.com/item?id=19195148) about it. Of course dependencies have sub-dependencies, but I wouldn't say dependencies in production are significantly worse by any means.

LZoog:
@lmorchard @jodyheavener @vbudhram @chenba I will squash my commits before merging but I wanted my changes to be view-able so you didn't have to look at this entire thing again, if anyone wants to do a second pass: https://github.com/mozilla/fxa/pull/4590/commits/f7437ec1c9af699157b50ed9c768de384385e2d6

I'm waiting for an r+ from @dannycoates before merging. 👍

docs/adr/0011-create-new-react-app-for-settings-redesign.md
```diff
@@ -0,0 +1,79 @@
...
+- Description
+  - Create a new React application for the Settings page and approach it with an incremental view-by-view build up.
+- Pros
+  - Propels us into our new framework of choice and reduces future refactoring efforts.
+  - Settings doesn't have to manage different integrations like the signin/signup flows do. It seems like a reasonable place to begin our conversion regardless of a redesign.
+  - Allows for a fresh start and audit of our HTML semantics, CSS architecture, and accessibility practices as components are being built.
+  - Allows for test auditing and scrutiny. We can potentially remove superfluous tests or add those we lack.
+  - May allow for a fresh take on how we do metrics in Settings.
+  - Offers an opportunity to set up and use Storybook in FxA. If we do a view-by-view build up, we can deploy changes somewhere for UX and other folks to see progress and request tweaks as features are built out.
+  - We can potentially use this as an opportunity to review and extract model and API code from the existing settings app into shared modules.
```
rfk: Just driving by here to say, I'm excited to see this work happening, and I'm curious how the boundary between settings page and the login flow will work out in practice. They currently both depend on the same set of models and the same data stored in `localStorage`, but I'd love to see a world where they're more cleanly separated. It might even help us make easier progress on things like using third-party JS in the settings page. 👍 to pursuing Option C here.

PR #4598, opened 2020-03-20
chore(styles): update sub cancel warning UI in settings

johngruen:
Fixes FXA-1389

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
@@ -14,7 +14,7 @@
     <div class="error"></div>
     {{#subscriptions.0}}
       <p class="subscription-link">
-        {{#unsafeTranslate}}Looking to cancel a paid subscription? Go to <a href="/subscriptions">Manage Subscriptions</a>.{{/unsafeTranslate}}
+        {{#unsafeTranslate}}Looking to cancel a paid subscription? Go to <a href="/subscriptions">Subscriptions &amp; Payments</a>.{{/unsafeTranslate}}
       </p>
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
```diff
@@ -14,7 +14,7 @@
     <div class="error"></div>
     {{#subscriptions.0}}
       <p class="subscription-link">
-        {{#unsafeTranslate}}Looking to cancel a paid subscription? Go to <a href="/subscriptions">Manage Subscriptions</a>.{{/unsafeTranslate}}
+        {{#unsafeTranslate}}Looking to cancel a paid subscription? Go to <a href="/subscriptions">Subscriptions & Payments</a>.{{/unsafeTranslate}}
```
chenba: I do believe that needs to be an `&amp;` instead of plain ol' `&`

johngruen: on it, thanks

packages/fxa-content-server/app/scripts/templates/settings/subscription.mustache
```diff
@@ -1,7 +1,7 @@
 <div id="manage-subscription" class="settings-unit">
   <div class="settings-unit-stub">
     <header class="settings-unit-summary">
-      <h2 class="settings-unit-title">{{#t}}Subscriptions & Payments{{/t}}</h2>
+      <h2 class="settings-unit-title">{{#t}}Subscriptions &amp; Payments{{/t}}</h2>
```
johngruen: noticed we didn't have the entity here either

PR #4601, opened 2020-03-20
fix(metrics): restrict content-server nav timing metric names

### LINKED ISSUE #4599: [content server] Restrict Navigation Timing Metric Names to Defined Set
From look at stage data, content server is logging some other performance metrics as navigation timings.  Restrict them to the ones defined in flow-performance.js in fxa-shared.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1391)
┆Issue Number: FXA-1391

chenba:
This patch restricts the navigation timing statsd metric names to those
defined in fxa-shared/metrics/flow-performance.js.

Fixes #4599 (FXA-1391)

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-content-server/server/lib/flow-event.js
+++ b/packages/fxa-content-server/server/lib/flow-event.js
@@ -206,12 +206,18 @@ function logFlowEvent(event, data, request) {
 }
 
 function logStatsdPerfEvent(eventData) {
+  // See https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/metrics-events.md#success-event-names
   if (eventData.event.startsWith('flow.performance.')) {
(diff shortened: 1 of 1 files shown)


chenba:
It's a bit baffling, but on stage I'm seeing metrics such as `fxa-content.nt.signup` and `fxa-content.nt.force-auth`.  I have not able to find where they are coming from; there are only two places in the content-server code base where are `flow.performance.*` events are emitted (https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/flow-event.js#L80 and https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/flow-event.js#L131).

PR #4604, opened 2020-03-20
chore(metrics): reduce nav timing metric tags in Payments

### LINKED ISSUE #4594: [payments server] Remove All But the Path Tag from InfluxDB Metric Tags
This will help to reduce the number of metrics in influxdb.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1387)
┆Issue Number: FXA-1387

chenba:
This patch removes the country, OS, and user-agent name tags from statsd
metric.  It helps to lower the amount of metrics in InfluxDB.

Fixes #4594 (FXA-1387)

@mozilla/fxa-devs r?


chenba:
@vbudhram There was a discussion on Slack around tags.  Because of the multiplicative effect of tags, removing them should reduce the number of metrics quite a bit over time.  (The discussion was about the content server metrics, and payments doesn't get a lot of traffic yet, but I figured I'd just remove them.)

PR #4610, opened 2020-03-23
feat(admin panel): add server tests setup, heartbeat and version routes, and csp

### LINKED ISSUE #4410: [admin panel A1.0] Add server tests, version, heartbeat to admin-panel package
Reference the payments server.

The config could also be set with expected types. This can/should likely be done before the security checklist (#FXA-1221).

Also, switching from Express to Hapi at this point could be worth looking into at this point, as I believe we've decided that should be our de facto standard.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1288)
┆Issue Number: FXA-1288

jodyheavener:
Closes #4410

In this PR:

- Add routes for Version and Heartbeat
- Set up server testing command, config, base tests
- Set up a base CSP

Some notes:

- A lot of this was copy-pasted from fxa-payments-server, but it also required converting from CommonJS to ES6 modules and setting up minimal Typescript support. My TS sucks, so I'm 100% open to any feedback.
- For CSP, I removed a number of directive values that were used in the payments server as I don't expect that we'll need them for the admin panel. I also made some assumptions, like I assumed we'll want `'unsafe-inline'` for styles. At any rate I'll definitely want a second set of eyes on it.


packages/fxa-admin-panel/server/lib/csp/blocking.ts
```diff
@@ -0,0 +1,73 @@
...
+  const PUBLIC_URL = config.get('listen.publicUrl');
+  const HOT_RELOAD_WEBSOCKET = PUBLIC_URL.replace(/^http/, 'ws');
+  const EXTRA_IMG_SRC = config.get('csp.extraImgSrc');
+
+  //
+  // Double quoted values
+  //
+  const NONE = "'none'";
+  const SELF = "'self'";
+  const UNSAFE_INLINE = "'unsafe-inline'";
```
LZoog: I know we chatted about this in Slack, but were you able to get this working without `unsafe-inline` styles?

packages/fxa-admin-panel/server/config/index.ts
```diff
@@ -32,6 +32,33 @@ const conf = convict({
...
   },
+  csp: {
+    enabled: {
+      default: true,
+      doc: 'Send "Content-Security-Policy" header',
+      env: 'CSP_ENABLED',
+    },
+    reportUri: {
+      default: 'https://accounts.firefox.com/_/csp-violation',
+      doc: 'Location of "report-uri" for full, blocking CSP rules',
```
LZoog: Hmm so when this was setup in payments we decided to just forward CSP violations to accounts.firefox.com, but I'm not sure if we want to do that with the admin panel. We probably want to remove `https://accounts.firefox.com` from lines 42 and 52 (check out the content server config) but someone else may need to chime in here. 🤔

packages/fxa-admin-panel/server/config/index.ts
```diff
@@ -32,6 +32,33 @@ const conf = convict({
...
+      default: false,
+      doc: 'Send "Content-Security-Policy-Report-Only" header',
+      env: 'CSP_REPORT_ONLY_ENABLED',
+    },
+    reportOnlyUri: {
+      default: 'https://accounts.firefox.com/_/csp-violation-report-only',
+      doc: 'Location of "report-uri" for report-only CSP rules',
+      env: 'CSP_REPORT_ONLY_URI',
+    },
+    extraImgSrc: {
```
LZoog: I believe this was added in payments to account for a CDN where we're storing some product images, but we shouldn't need it here since all the images will be hosted in the app.

packages/fxa-admin-panel/server/lib/server.ts
```diff
@@ -2,30 +2,58 @@
...
+  bodyParser.text({
+    type: 'text/plain',
+  }),
+  bodyParser.json({
+    type: ['json', '*/json', 'application/csp-report'],
+  })
 );
 
+if (config.get('csp.enabled')) {
+  // @ts-ignore
```
LZoog: Most of these `@ts-ignore`s can probably be addressed somehow, does `as express.RequestHandler` fail? Might want a TODO around addressing those if they're being a pain.

packages/fxa-admin-panel/server/lib/csp/blocking.ts
jodyheavener: I was! What I also found out, though, was that I was not able to get it working without `unsafe-inline` on scripts. Any idea why that would be?

packages/fxa-admin-panel/server/config/index.ts
jodyheavener: So, it turns out a couple things are relevent here. jbuck helped me figure it out.

Further to our discussion around whether or not reports are actually handled, it turns out they are indeed. The paths `/_/csp-violation` and `/_/csp-violation-report-only` are handled at the nginx level.

Then, in production the vars for `CSP_REPORT_URI` and `CSP_REPORT_ONLY_URI` are set so the values here don't matter too much. I was advised that setting them to the above paths are fine in this case.

packages/fxa-admin-panel/server/lib/version.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  let stdout = UNKNOWN;
+  const gitDir = path.resolve(__dirname, '..', '..', '.git');
+  const configPath = path.join(gitDir, 'config');
+  const cmd = 'git config --get remote.origin.url';
+
+  try {
+    // @ts-ignore
+    stdout = cp.execSync(cmd, cmd, { env: { GIT_CONFIG: configPath } });
```
jodyheavener: 🤔 TS is yelling at me that `execSync` should only receive up to two arguments, and sure enough [the documentation](https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_execsync_command_options) reflects that.

I thought maybe this was a typo, and should be `cp.execSync(cmd, { env: { GIT_CONFIG: configPath } });`, but the current setup is definitely what's needed; it returns invalid results otherwise.

packages/fxa-admin-panel/server/lib/csp/blocking.ts
LZoog: Ahh I just assumed this was for `styles`, oops. We definitely don't want it on `scripts`. I'll have to check into your branch and poke at it because I'm not sure why it'd complain.

packages/fxa-admin-panel/package.json
```diff
@@ -54,16 +56,18 @@
     "@types/express": "^4.17.2",
     "@types/helmet": "0.0.45",
     "@types/jest": "^25.1.3",
-    "@types/jsdom": "^16.1.0",
+    "@types/jsdom": "^12.2.4",
```
jodyheavener: This was apparently a necessary change (for now), as I was not able to `npm run build:server`. [See this thread](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/26338#issuecomment-589523765).

packages/fxa-admin-panel/server/lib/server.test.ts
```diff
@@ -0,0 +1,56 @@
...
+            'application/json; charset=utf-8'
+          );
+          const body = response.body;
+          expect(Object.keys(body).sort()).toStrictEqual([
+            'commit',
+            'source',
+            'version',
+          ]);
+          expectValueNotToBeUnknown(body.commit);
+          // expectValueNotToBeUnknown(body.source);
```
LZoog: Should this be uncommented?

packages/fxa-admin-panel/server/lib/server.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  expect(value).toBeTruthy();
+  expect(value).not.toBe('unknown');
+}
+
+describe('Test simple server routes', () => {
+  test('__version__ should have correct structure', () => {
+    return new Promise(done => {
+      request(app)
+        .get('/__version__')
+        .then(response => {
```
LZoog: `async`/`await` would be nice but this is fine👍

packages/fxa-admin-panel/server/lib/csp/blocking.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * blockingCspMiddleware is where to declare rules that will cause a resource
+ * to be blocked if it runs afowl of a rule.
+ */
+export default function cspBlocking(config: { [key: string]: any }) {
+  const ADMIN_SERVER = getOrigin(config.get('servers.admin.url'));
+  const CDN_URL = config.get('staticResources.url');
+  const DATA = 'data:';
```
LZoog: I don't think we need `data:` unless we end up base64'ing SVGs (meh).

packages/fxa-admin-panel/server/lib/server.test.ts
jodyheavener: Whoops this was from testing execAsync. Fixing.

PR #4618, opened 2020-03-23
feat(payment-server): update legal checkbox text on tiers page

jackiemunroe:
- updated legal checkbox text on subscription upgrade confirmation page to handle various billing intervals and interval counts

issue #3987

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.test.tsx
@@ -1,8 +1,13 @@
 import React from 'react';
-import { render, cleanup, act, fireEvent } from '@testing-library/react';
+import TestRenderer from 'react-test-renderer';
+import { render, cleanup, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
+import fs from 'fs';
+import path from 'path';
+import { FluentBundle } from 'fluent';
 
 import { APIError } from '../../../lib/apiClient';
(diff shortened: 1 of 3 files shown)


packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.test.tsx
```diff
@@ -111,6 +124,344 @@ describe('routes/Product/SubscriptionUpgrade', () => {
...
+          const expected =
+            'I authorize Mozilla, maker of Firefox products, to charge my payment method <strong>$5.00 daily</strong>, according to payment terms, until I cancel my subscription.';
+
+          const msg = bundle.getMessage(msgID);
+          const actual = bundle.format(msg, {
+            ...args,
+            intervalCount: 1,
+          });
+
+          expect(actual.replace(/(\u2068|\u2069)/gu, '')).toEqual(expected);
```
vbudhram: This might just be how fluent works but why do these values need to be replaced?  Does `bundle.format` have options to omit it?

jackiemunroe: From all of my testing, I couldn't find a way to omit these "hidden" characters. They kept getting added around each of the variables that are displayed.

jackiemunroe:
> @jackiemunroe LGTM. Can these branches be merged into master, or are they dependent on things in feature branch?

They are not dependent on the feature branch. However, after speaking with @clouserw we decided to use the feature branch to ensure that all of the language changes go out together and to make it easier for someone from the l10n team to review the `.ftl` updates.

PR #4622, opened 2020-03-24
chore(config): Default to a useable Fenix OAuth config in local dev

rfk:
Because:

* The fxa-auth-server dev config defaults Fenix and Firefox-iOS to
  using OAuth redirect URIs on https://127.0.0.1:3030.
* The fxa-content-server dev config defaults to checking OAuth scoped
  key access against production OAuth redirect URIs.
* The resulting mis-match means you can't sign in to Fenix using
  a local dev server in the default config.

This commit:

* Changes the default local config for fxa-content-server to check
  OAuth scoped key access against the development OAuth redirect URIs.

### DIFF
--- a/packages/fxa-content-server/server/config/local.json-dist
+++ b/packages/fxa-content-server/server/config/local.json-dist
@@ -18,7 +18,17 @@
     "level": "debug"
   },
   "scopedKeys": {
-    "enabled": true
+    "enabled": true,
+    "validation": {
+      "https://identity.mozilla.com/apps/oldsync": {
+        "redirectUris": [
+         "http://127.0.0.1:3030/oauth/success/a2270f727f45f648",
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/server/config/local.json-dist
```diff
@@ -18,7 +18,17 @@
     "level": "debug"
   },
   "scopedKeys": {
-    "enabled": true
+    "enabled": true,
+    "validation": {
+      "https://identity.mozilla.com/apps/oldsync": {
+        "redirectUris": [
+         "http://127.0.0.1:3030/oauth/success/a2270f727f45f648",
+         "https://127.0.0.1:3030/oauth/success/1b1a3e44c54fbb58",
```
vladikoff: @rfk should this be https?

rfk: Whoops, good catch.

PR #4624, opened 2020-03-24
fix(settings): wrap overflowing settings group headers so they don't break layout

jodyheavener:
Addresses https://github.com/mozilla-mobile/firefox-ios/issues/6130

In English none of our settings titles are long enough to bump up against the right side button. Lucky us. However in localized versions of the settings panel these titles can get quite long. Currently long strings push the content of the element out, effectively breaking the settings layout.

Screen Shot 2020-03-23 at 3 12 27 PM

This fix allows the text to wrap to a new line. To accommodate this I've changed the explicit height of the settings row to a minimum height, and added vertical padding equivalent to the space between the right-side button and the top/bottom of the container.

Screen Shot 2020-03-23 at 3 12 07 PM

This is more unlikely, but works on desktop as well:

Screen Shot 2020-03-23 at 3 11 03 PM

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings.scss
@@ -282,13 +282,13 @@ body.settings #stage .settings {
 .settings-unit-stub {
   align-items: center;
   display: flex;
-  height: 72px;
+  min-height: 72px;
 
   @include respond-to('big') {
-    padding: 0 32px;
+    padding: 16px 32px;
   }
   @include respond-to('small') {
-    padding: 0 16px;
+    padding: 18px 16px;
   }
 
   .settings-unit-summary {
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/styles/modules/_settings.scss
```diff
@@ -331,15 +331,8 @@ body.settings #stage .settings {
     margin: 5px 0 0 0;
   }
 
-  .settings-unit-title,
-  .settings-unit-subtitle {
-    text-overflow: ellipsis;
```
LZoog: We may want to double check with @ryanfeeley if he's back (or ask Alex/Chelsea who we should ask). I think the idea was that we should have an ellipsis like in this screenshot (stolen from the linked issue) if the text is too long, but your solution might be preferable (looks good to me!). I'll approve but pending UX/Product approval. 👍
image

jodyheavener: Sounds good. FWIW my personal preference is to reflow, otherwise you're arbitrarily limiting certain locales. @ryanfeeley what do you think?

davismtl: We do normally fallback to ellipsis but when looking at that example on mobile, it seems like it's less than ideal.

I like your suggestion. In the absence of Ryan, let's just move forward with your suggestion.

PR #4626, opened 2020-03-24
Fix: fxa-content-server mocked Statsd is undefined

### LINKED ISSUE #4608: fxa-content-server mocked Statsd is undefined.
fxa-content-server mocked Statsd is undefined.

https://github.com/mozilla/fxa/blob/00e4ecd27059a80f07d90d8ad4535209e1bf6557/packages/fxa-content-server/server/lib/statsd.js#L11-L17

and will cause error in
https://github.com/mozilla/fxa/blob/00e4ecd27059a80f07d90d8ad4535209e1bf6557/packages/fxa-content-server/server/lib/flow-event.js#L215

It should be
```javascript
 const noopStatsd = () => { 
   const mock = {}; 
   Object.keys(StatsD.prototype).forEach(x => { 
     if (typeof StatsD.prototype[x] === 'function') { 
       mock[x] = NOOP; 
     } 
    return mock;
   }); 
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1394)
┆Issue Number: FXA-1394

imskr:
Fixes #4608

### DIFF
--- a/packages/fxa-content-server/server/lib/statsd.js
+++ b/packages/fxa-content-server/server/lib/statsd.js
@@ -14,6 +14,7 @@ const noopStatsd = () => {
     if (typeof StatsD.prototype[x] === 'function') {
       mock[x] = NOOP;
     }
+    return mock;
   });
 };
 


bbangert:
Dupe of #4616?

imskr:
oh I didn't knew there is PR for this! @bbangert

PR #4637, opened 2020-03-24
feat(payment-server): update settings page billing description

jackiemunroe:
- updated subscription billing description on settings page to handle various billing intervals and interval counts

issue #3987

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/PlanDetails/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/PlanDetails/index.test.tsx
@@ -280,7 +280,7 @@ describe('PlanDetails Component', () => {
           intervalCount: 1,
         });
 
-        expect(actual == expected);
+        expect(actual.replace(/(\u2068|\u2069)/gu, '')).toEqual(expected);
       });
 
       it('returns the correct string for an interval count greater than 1', () => {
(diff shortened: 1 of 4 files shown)


packages/fxa-payments-server/src/routes/Product/PlanDetails/index.test.tsx
```diff
@@ -376,7 +376,7 @@ describe('PlanDetails Component', () => {
           intervalCount: 6,
         });
 
-        expect(actual == expected);
+        expect(actual.replace(/(\u2068|\u2069)/gu, '')).toEqual(expected);
```
jodyheavener: What characters are u2068 and u2069? I feel like this assertion might benefit from a helper function with how frequently it's used and how it's somewhat specific.

jackiemunroe: Ah yes, so this was a discovery I made while trying to perform this test: those characters are "First Strong Isolate" and "Pop Directional Isolate". When a message is rendered with fluent, those characters (which are hidden) surround each variable. We don't normally have tests for our fluent files, but this was added to verify the syntax handling for plurality. The previous check for equality didn't actually validate anything.

jackiemunroe: It is not a helper function currently, because doing so in one branch would have blocked the other work. So instead, it has been duplicated on a per-test suite basis in each of the parts with the intent of refactoring once it's all been merged into the feature branch.

jodyheavener: Cool, thank you for the clarification!

PR #4640, opened 2020-03-24
refactor(assets): remove content server unused images and svgs

### LINKED ISSUE #834: Remove deprecated images from content server
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-808)
┆Issue Number: FXA-808

jodyheavener:
Closes #834


packages/fxa-content-server/app/images/icon-show-password.svg
```diff
@@ -1 +0,0 @@
-<svg data-name="Layer 2" xmlns="http://www.w3.org/2000/svg" width="55" height="46" viewBox="-16 -13 47 38"><path d="M16 5.7A8.32 8.32 0 0 0 8 0 8.32 8.32 0 0 0 .05 5.7a.86.86 0 0 0 0 .59A8.31 8.31 0 0 0 8 12a8.32 8.32 0 0 0 8-5.7.88.88 0 0 0 0-.6zM8.5 3A1.5 1.5 0 1 1 7 4.5 1.5 1.5 0 0 1 8.5 3zM8 10a6.33 6.33 0 0 1-5.94-4 6.08 6.08 0 0 1 2.5-3A4 4 0 0 0 4 5a4 4 0 0 0 8 0 4 4 0 0 0-.56-2 6 6 0 0 1 2.5 3A6.33 6.33 0 0 1 8 10z"/></svg>
```
LZoog: This one is tricky because we actually DO use it, we just [base64'd/inlined](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/templates/partial/show-password.mustache#L2) it.

We probably need an issue made for discussion on the best way to handle our SVGs. We add a hash in the content server so we rarely `<img src="`, but we need to sometimes, and inlining SVGs not only reduces HTTP requests but there's no need to bust any cache on them. It might be ideal to automate something like this:
```
<img src='data:image/svg+xml;utf8,<svg ... > ... </svg>'>
```

... so that way we're still inlining, but but we're not base64'ing (keeping the integrity of the SVG intact and reducing conversions). Regardless I'd say leave this one in here.

packages/fxa-content-server/app/images/send_tab_android.svg
```diff
@@ -1 +0,0 @@
...venodd"><path d="M338.826.275a2.82 2.82 0 00-.44-.12l-.4-.14a6.54 6.54 0 00-.88 0 6.26 6.26 0 00-.88 0l-.42.14a3.6 3.6 0 00-.44.12 3.42 3.42 0 00-.46.26l-.32.16a3.74 3.74 0 00-.68.58l-13.58 13.56a4.525 4.525 0 006.4 6.4l5.86-5.86v21.84a4.52 4.52 0 009.04 0v-21.88l5.84 5.86a4.525 4.525 0 106.4-6.4l-13.56-13.56a3.86 3.86 0 00-.7-.58l-.3-.16z" fill="url(#a)" fill-rule="nonzero" style="animation:bob .5s infinite;transform-origin:94% 20%" transform="scale(-1)"/><path d="M0 57h360v56H0z" fill="#f0f0f4" fill-rule="nonzero"/><path d="M336 81c1.1 0 2-.9 2-2s-.9-2-2-2-2 .9-2 2 .9 2 2 2zm0 2c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zm0 6c-1.1 0-2 .9-2 2s.9 2 2 2 2-.9 2-2-.9-2-2-2zM291.222 95h-9.444A3.778 3.778 0 01278 91.222v-9.444A3.778 3.778 0 01281.778 78h9.444A3.778 3.778 0 01295 81.778v9.444A3.778 3.778 0 01291.222 95zm-9.365-15A1.857 1.857 0 00280 81.857v9.286c0 1.026.831 1.857 1.857 1.857h9.286A1.857 1.857 0 00293 91.143v-9.286A1.857 1.857 0 00291.143 80zm.143-3a2 2 0 012-2h7.987a6 6 0 016 6v.016l-.024 7.987a2 2 0 01-2.006 1.994l.03-9.986V81a4 4 0 00-4-4z" fill="#0c0c0d" fill-opacity=".6"/><rect fill="#e0e0e6" fill-rule="nonzero" height="40" rx="8" width="248" x="8" y="65"/></g></svg>
```
LZoog: I'm not sure if we ever ended up using these (I don't see them when searching, anyway), but we probably also want to leave the `send_tab` SVGs. I animated these 5 or 6 months for us for a "let's teach the user how to use send tab" project. Maybe check with @davismtl. 🤔

jodyheavener: Skipped to be safe. They're not harming anything.

packages/fxa-content-server/app/images/icon-show-password.svg
jodyheavener: Big fan of inline SVG. You also have much more flexibility when styling individual <g>roups or <path>s.

Issue #828, opened 2019-04-15
Remove Reference to Chromeless param in fxa query-params docs


johngruen:
On line 81 of packages/fxa-content-server-docs-query-params.md

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-807)
┆Issue Number: FXA-807

shane-tomlinson:
Ah yes, I forgot to remove that with the removal of the iframe.

championshuttler:
Can i take this up?

johngruen:
@championshuttler sure!

smishra2903:
If this issue is still open, can I work on this? 
Also, i am beginning to work on issues can I get some help on how to start with this?

vladikoff:
@smishra2903 Hey, yeap!

See https://github.com/mozilla/fxa/search?q=Chromeless&unscoped_q=Chromeless and in  packages/fxa-content-server-docs-query-params.md , there still might be references to it. Send a pull request removing that

apsychogirl:
Is this issue open to work on?

vladikoff:
@apsychogirl Yeap!

apsychogirl:
@vladikoff I was not able to find the instance of Chromeless in the fxa query-params doc. Can you please help me in fixing this issue?

vladikoff:
@apsychogirl hm this might be it: https://github.com/mozilla/fxa/search?q=chromeless&unscoped_q=chromeless&type=Code

clouserw:
fixed in https://github.com/mozilla/fxa/pull/5000

PR #4647, opened 2020-03-25
Remove Reference to Chromeless param

apsychogirl:
Fixes - #828 
>Removed the instance of "Chromeless" from [packages/fxa-content-server/app/tests/spec/models/reliers/relier.js] .

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/models/reliers/relier.js
+++ b/packages/fxa-content-server/app/tests/spec/models/reliers/relier.js
@@ -110,7 +110,7 @@ describe('models/reliers/relier', function() {
     });
   });
 
-  ['trailhead-1', 'chromeless'].forEach(value => {
+  ['trailhead-1'].forEach(value => {
     testInvalidQueryParam('style', value);
   });
 


packages/fxa-content-server/app/tests/spec/models/reliers/relier.js
```diff
@@ -110,10 +110,6 @@ describe('models/reliers/relier', function() {
     });
   });
 
-  ['trailhead-1', 'chromeless'].forEach(value => {
```
vladikoff: @apsychogirl could you please just remove the `chromeless`, let's keep the trailhead-1 while we still support the `style` query param.

apsychogirl: > @apsychogirl could you please just remove the `chromeless`, let's keep the trailhead-1 while we still support the `style` query param.

Should I also remove the `value` query param which was earlier taking the value for `Chromeless` in line 114?

vbudhram:
Hey @apsychogirl, thanks for PR! Before we can merge, can you squash commits into one and update message to follow https://github.com/mozilla/fxa/blob/master/CONTRIBUTING.md#git-commit-guidelines.

vbudhram:
@apsychogirl Looks like you will also need to setup GPG signing, ref https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification#gpg-commit-signature-verification.

clouserw:
@apsychogirl are you able to set up GPG signing for your commits?  If not we can commit this for you.  Thanks for the patch!

apsychogirl:
My laptop just broke... won't be able to sign the commit :(

clouserw:
The source branch for this was removed/unknown so I can't push here.  I opened a new PR with your patch in it @ https://github.com/mozilla/fxa/pull/5000

PR #4648, opened 2020-03-25
[Longer Duration Subscriptions] Show users correct subscription cycle information on payments pages - Feature Branch

### LINKED ISSUE #3987: Show users correct subscription cycle information on payments pages

jackiemunroe:
closes #3987


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -208,18 +234,17 @@ export const PaymentForm = ({
 
       {confirm && plan && (
         <Localized
-          id="payment-confirm"
-          $interval={plan.interval}
+          id={`payment-confirm-${plan.interval}`}
+          $intervalCount={plan.interval_count}
           $amount={formatCurrencyInCents(plan.amount)}
```
stasm: I looked at `formatCurrencyInCents` in `lib/formats.ts`, and I'd like to flag it as an issue from the localization POV.

```ts
export function formatCurrencyInCents(amount: number) {
  return (amount / 100.0).toFixed(2);
}
```

The function converts the number to a string using English rules of number formatting, i.e. using the dot `.` as the decimal separation. This won't work for many languages out there.

The right approach is to use [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/NumberFormat) which can format numbers using language specific rules. However, rather than use it here (which would also require access to the current `locale`), you can rely on Fluent doing the right thing for you :)

Fluent auto-formats all numbers passed into translations, but it uses `Intl.NumberFormat` default options which won't work for currencies. Options can be customized by developers in the source code, and then also overriden by localizers in the translations.

#### Custom options set in the source code

These effectively become the new defaults for the one particular number their set for and will be used in all translations, unless explicitly overriden by the localizer. Rather than pass `$amount={formatCurrencyInCents(plan.amount)}` you can instantiate a `FluentNumber`:

```ts
$amount={new FluentNumber(plan.amount, {
    style: "currency",
    currency: "USD",
    currencyDisplay: "symbol",
})}
```

With the above definition of `$amount`, you don't need the `$` symbol in front of the amount in the translation:

```properties
## subscription upgrade plan details
plan-price-day = {$intervalCount ->
    [one] ${$amount} daily
   *[other] ${$amount} every {$intervalCount} days
}
```

↑ becomes ↓

```properties
## subscription upgrade plan details
plan-price-day = {$intervalCount ->
    [one] {$amount} daily
   *[other] {$amount} every {$intervalCount} days
}
```

#### Per-locale option overrides

Localizers can then customize number formatting options if they think it would produce a better translations. For instance, they could choose to change `currencyDisplay` from `symbol` (`5,01 $` in Polish, note the comma) to `code` to better emphasize the fact that this is a foreign currency (`5,01 USD` in Polish).

```properties
## subscription upgrade plan details
plan-price-day = {$intervalCount ->
    [one] {$amount} daily
   *[other] {NUMBER($amount, currencyDisplay: "code")} every {$intervalCount} days
}
```

#### The Caveat

There's one caveat with this approach, which you should be aware of. Currently, all options can be overriden by localizer, including the `currency` itself! This obviously shouldn't be possible. I've meant to fix this for a long time, but there never was a use-case which would make it important, until right now. I'll be happy to fix it in `@fluent/bundle`.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -9,6 +9,9 @@ import {
 import ScreenInfo from '../../src/lib/screen-info';
 import { ReactStripeElements } from 'react-stripe-elements';
 import nock from 'nock';
+import fs from 'fs';
+import path from 'path';
+import { FluentBundle } from 'fluent';
```
stasm: @jackiemunroe I see that you haven't upgraded to `@fluent/bundle` 0.15 nor `@fluent/react` 0.11 in this PR. Did you run into any trouble? I might be able to help :) Or, is it better to upgrade in a separate PR?

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
jackiemunroe: @stasm That sounds great. I am wondering if it might be a good idea to create a separate issue for the purpose of currency display? The reason for this is that right now, the front-end is assuming `USD` in all cases (This is because we only offer subscriptions to US-based customers), however, we can update the API to provide the currency alongside the amount, which would allow for a more future-proof solution, combined with the update to `@fluent/bundle`.
So, what I would like to propose: create a new branch that has an updated version of `@fluent/bundle` and our subscription API, so that we can use localized currency with confidence.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -385,4 +495,36 @@ export const MOCK_CUSTOMER_AFTER_SUBSCRIPTION = {
...
+  args: any
+): string {
+  const msg = bundle.getMessage(msgId);
+  if (msg === undefined) {
+    throw Error(`unable to locate fluent message with id: '${msgId}'`);
+  }
+  const actual = bundle.format(msg, { ...args });
+
+  // remove 'hidden' characters that surround arguements in resulting string
+  return actual.replace(/(\u2068|\u2069)/gu, '');
```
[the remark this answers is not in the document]
jackiemunroe: Excellent!

packages/fxa-payments-server/src/lib/test-utils.tsx
jackiemunroe: @stasm Since there was a decent amount of changes here, I thought it would be better to upgrade fluent in isolation.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
stasm: I'm all for the iterative approach, and I'll be happy to help with the upgrade to `@fluent/bundle` PR.

One thing to keep in mind: depending on when you expose these strings to localizers, you'll need to change their identifiers if you make a change to the `$amout` variable. This way Pontoon will let localizers know that they need to update their translations.

It might be a good idea to remove that `$` symbol right now and for example prepend it in `formatCurrencyToCents`, temporarily.

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -66,7 +85,27 @@ payment-cvc =
   .label = CVC
 payment-zip =
   .label = ZIP code
-payment-confirm = I authorize Mozilla, maker of Firefox products, to charge my payment method <strong>${ $amount } per { $interval }</strong>, according to payment terms, until I cancel my subscription.
+  
+payment-confirm-day = { $intervalCount -> 
+  [one] I authorize Mozilla, maker of Firefox products, to charge my payment method <strong>${$amount} daily</strong>, according to payment terms, until I cancel my subscription.
```
peiying2: Mozilla and Firefox are brand names, should be treated as variables: 
{ -brand-name-mozilla } 
{ -brand-name-firefox }

If these are the only brands used in the overall FxA files, you can define them at the beginning of the file. If there are more to come, you can have a separate brands.ftl  that other files can reference to.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
jackiemunroe: This temporary change has been made

packages/fxa-payments-server/public/locales/en-US/main.ftl
jackiemunroe: Each of these updates were made in separate branches as to not block one another, so I did not see that the formats ended up identical for those translations. As to why they are separate: It is possible that at one point in time they were not the same.

Given they are in two different flows, I would like to get @johngruen 's input on if we can merge them.

peiying2:
The file needs to include more context for the variables. They can be added at the beginning of the cluster of strings starting with # or grouped together at the beginning of the file with ##.
```
# Variables:
    #  $intervalCount (Number) - The interval of payments, in days.
    #  $productName (String) - The name of the subscribed product.
    #  $amount (Number) - The amount billed, in USD. It will be formatted as currency.
    day-based-plan-details-amount = { $intervalCount ->
        [one] { $productName } billed { $amount } daily
       *[other] { $productName } billed { $amount } every {$intervalCount} days
    }

peiying2:
Looks good to me! Approved from l10n side.

PR #4657, opened 2020-03-25
feat(metrics):  validate Amplitude events with a JSON schema

chenba:
This patch adds a JSON schema to fxa-shared and uses it to validate the
Amplitude events in the auth, content, payments servers.

At the moment when an Amplitude event fails validation against this
initial schema, it is still emitted.  This will be changed in a future
commit once the degree of confidence on the schema is higher.

@mozilla/fxa-devs r?

### DIFF
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -46,6 +46,14 @@ const conf = convict({
       env: 'LOG_FORMAT',
     },
   },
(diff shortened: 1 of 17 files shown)


chenba:
While I'm confident about the JSON schema, I do not want to get caught off guard by some "invalid" JOSN blobs in production.  I want to run this in production for at least a week and update the schema if any error surfaces.

PR #4676, opened 2020-03-27
fix(settings): replace Subscriptions & Payments section title encoded "&amp;" with "&"

### LINKED ISSUE #4675: Subscriptions & Payments incorrectly displayed in FxA settings page
#### Affected versions:
- All Firefox

#### Affected Platforms:
- Windows 10 (64-bit)
- macOS 10.14.6

#### Steps to reproduce
1. Go to the [Full-Device VPN](https://stage.guardian.nonprod.cloudops.mozgcp.net/vpn) page.
2. Create a subscription.
3. Go to the [Firefox Accounts](https://accounts.stage.mozaws.net/settings) settings page.
4. Look at the **Subscriptions & Payments** section.

#### Expected result:
- **Subscriptions & Payments** is displayed.

#### Actual result:
- `Subscriptions &amp; Payments` is displayed.

image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1428)
┆Issue Number: FXA-1428

jodyheavener:
Closes #4675

Looks like this was introduced [here](https://github.com/mozilla/fxa/commit/09be98757fc6aef1606d024a89f05ae4c08c5627#diff-6c4fc621920a9a78dfd234a6e3e4a125) and our translated strings don't like encoded characters.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/subscription.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/subscription.mustache
@@ -1,7 +1,7 @@
 <div id="manage-subscription" class="settings-unit">
   <div class="settings-unit-stub">
     <header class="settings-unit-summary">
-      <h2 class="settings-unit-title">{{#t}}Subscriptions &amp; Payments{{/t}}</h2>
+      <h2 class="settings-unit-title">{{#t}}Subscriptions & Payments{{/t}}</h2>
     </header>
     <button class="settings-button primary-button settings-unit-toggle">
       <span class="change-button">{{#t}}Manage…{{/t}}</span>


LZoog:
Does the other one in `delete_account` introduced by that PR need to be changed as well?

We definitely use character encoding in other places like `&nbsp;`. 🤔

LZoog:
@jodyheavener I believe the `&amp;`s need to be in `{{unsafeTranslate}}` tags instead of `{{t}}`. Thanks for jumping on this, it's pretty gross.

jodyheavener:
@LZoog yeah the other instance is inside an `unsafeTranslate`, but it looks like all these settings titles are using the regular `t` function, so unless there's an overwhelming reason to use unsafeTranslate I'm thinking we should stay consistent.

PR #4678, opened 2020-03-27
task(auth-server): add app links to download email

### LINKED ISSUE #3872: Mobile Enhancements for FPN
This epic tracks work requirements for Subscription Platform enhancements to better support Android and iOS clients for FPN, subscription flows on mobile devices, and support flows on mobile devices.

## Why are we doing this?
This work is intended to have the following benefits

1. To support the near-term release of FPN on Android and iOS
1. To improve the UI of the subscription flow based on early user feedback (adding product value, visual payment method cues)
1. Laying the ground for more complex subscriptions and for localization by making the UI more flexible.

## Success Criteria
1. Users successfully subscribing to FPN via mobile applications
1. We've been running at about a 48-50% conversion rate on our current payment UI. By adding product values and payment method visuals, we hope to improve this number.

┆Issue is synchronized with this [Jira Epic](https://jira.mozilla.com/browse/FXA-914)
┆Issue Number: FXA-914

meandavejustice:
- fixes #3872

Opening this up to get some feedback before I get it pixel perfect

cc/ @johngruen I'm wondering about the box shadow border, are we adding this to all emails or just this one?

cc/ @jackiemunroe This patch is assuming app store and play store links will be stored in Stripe, does this seem correct to you? (comment in diff)

This is what our email looks like with the changes in this patch
image

Here is the design from the jira issue.
email-wanted

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -193,6 +193,8 @@ class DirectStripeRoutes {
       productName: selectedPlan.product_name,
       planEmailIconURL: planMetadata.emailIconURL,
       planDownloadURL: planMetadata.downloadURL,
+      appStoreLink: planMetadata.appStoreLink,
+      playStoreLink: planMetadata.playStoreLink,
     });
     this.log.info('subscriptions.createSubscription.success', {
       uid,
--- a/packages/fxa-auth-server/lib/routes/validators.js
+++ b/packages/fxa-auth-server/lib/routes/validators.js
@@ -335,6 +335,8 @@ module.exports.subscriptionProductMetadataValidator = isA
     iconURL: isA.string().optional(),
     upgradeCTA: isA.string().optional(),
     downloadURL: isA.string().optional(),
+    appStoreLink: isA.string().optional(),
+    playStoreLink: isA.string().optional(),
   })
   .unknown(true);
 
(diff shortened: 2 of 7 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -193,6 +193,8 @@ class DirectStripeRoutes {
       productName: selectedPlan.product_name,
       planEmailIconURL: planMetadata.emailIconURL,
       planDownloadURL: planMetadata.downloadURL,
+      appStoreLink: planMetadata.appStoreLink,
+      playStoreLink: planMetadata.playStoreLink,
```
jackiemunroe: These are fine to grab from the metadata, and it makes sense to store it within metadata. However, have these new metadata properties been documented anywhere?
Also, I think that it would make sense for it to be stored on the `product` because, I would assume, that regardless of the plan being purchased (i.e. monthly, yearly, etc), the download link would be the same for each offering.

meandavejustice: I will move to product and document

johngruen:
@meandavejustice the styles as you have them are fine!

packages/fxa-auth-server/lib/routes/subscriptions.js
lmorchard: FWIW, I don't think the metadata needs to be moved here - [`planMetadata`](https://github.com/mozilla/fxa/blob/310d55d0bbf6cfa1e96ddfbfddd726e1d29bb9c0/packages/fxa-auth-server/lib/routes/subscriptions.js#L160) is a misnomer. It's derived from *both* plan & product metadata, [with the plan able to override the product](https://github.com/mozilla/fxa/blob/310d55d0bbf6cfa1e96ddfbfddd726e1d29bb9c0/packages/fxa-auth-server/lib/routes/utils/subscriptions.js#L12).

Issue #4546, opened 2020-03-13
Multiple secondary emails can be added following the same flow at the same time from different devices


vasilicatamas:
**Build:**
Stage [Train 1.162.0]

**Affected platforms:**
Windows 10 x64
Android 9

**Steps to reproduce:**
1. Create and confirm a new Firefox account.
2. Login using the new Firefox account on 2 devices.
3. On Device 1, go to the email inbox and open the "Set up recovery email" email.
4. On Device 2, go to the email inbox and open the "Set up recovery email" email.
5. On Device 2, add a secondary email and click the “Continue” button.
6. On Device 1, add a different secondary email and click the “Continue” button.
7. Confirm both secondary emails.
8. Check in Firefox Account Settings page the emails listed under “Secondary email” section.

**Expected results:**
Only one email can be assigned as a secondary email.

**Actual Results:**
Multiple emails can be set as secondary email following the “Set up recovery email” flow.

![2secondaryemails](https://user-images.githubusercontent.com/18114210/76628764-eeb1cd00-6545-11ea-951a-f3d07e2ddab0.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1375)
┆Issue Number: FXA-1375
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12807/Screen+Shot+2020-03-23+at+5.49.25+PM.png">Screen Shot 2020-03-23 at 5.49.25 PM.png</a>

imskr:
I would like to fix this error if no one is assigned. @vasilicatamas

data-sync-user:
➤ Wil Clouser commented:

[~adavis@mozilla.com] is supporting multiple secondary emails a bug?

data-sync-user:
➤ Alex Davis commented:

It is not a bug. We can keep this one open.
This should probably be an epic to some extent.

Note that our backend supports having multiple emails but we just don't support it on the front end.

data-sync-user:
➤ Jody Heavener commented:

[~adavis@mozilla.com] does this need to go to UX first, or can I use best judgement here? I'm interested in updating the frontend to address this.

data-sync-user:
➤ Alex Davis commented:

[~jheavener@mozilla.com] I'm pretty flexible but it will also depend on what you have in mind.

Is your goal to surface the ability to add multiple emails in fxa preferences to align with this edge case flow? Or is it to prevent people from encountering the flow described above?

[...]

PR #4684, opened 2020-03-27
feat(settings): update settings UI to permit adding multiple secondary emails

jodyheavener:
Closes https://github.com/mozilla/fxa/issues/4546

You'll need to read into the comments of the above issue to get the full details, but the gist is that we actually already support having multiple secondary emails per account on the back end, and currently with a special combination of actions you can end up with multiple emails visible on your account settings. So we're embracing it.

This update introduces proper UI to add multiple secondary emails to your account.

✅ @davismtl reviewed a demo of this flow

It's pretty straightforward, displaying a new "Add email" button after you've added your first secondary email:

Screen Shot 2020-03-27 at 1 45 57 PM

It presents the same screen that is displayed when you first add a secondary email:

Screen Shot 2020-03-27 at 1 48 41 PM

The "Add email" button is displayed up until you get to 10 additional emails (verified or not). To be safe, though, I've added a server check on the API to enforce this, and if you were to somehow get to the add email screen the response would set you straight:

Screen Shot 2020-03-27 at 1 43 16 PM

Now, you get a list of secondary emails, with the same capabilities (verify, make primary, remove). No fancy new styling as this is mostly a stop gap until we get to settings redesign:

Screen Shot 2020-03-27 at 1 46 47 PM

Everything looks good on mobile as well:

Screen Shot 2020-03-27 at 1 47 56 PM

Some other notes:

- Currently, when you are in the Add Email state and you click "Cancel" it closes the panel. This has changed slightly in that if you got to the Add Email state by clicking the new "Add email" button, clicking "Cancel" will take you back to the emails list.
- Currently, if you remove your secondary email the it closes the panel. This has changed slightly in that if you remove a secondary email, but it is not your only secondary email, the panel will not close. However, if you are removing your last secondary email it will close as usual.
- The frontend and auth server both enforce a limit of 10 emails. We can change this if desired.


packages/fxa-auth-server/lib/error.js
```diff
@@ -785,6 +786,15 @@ AppError.unverifiedPrimaryEmailNewlyCreated = () => {
   });
 };
 
+AppError.maxSecondaryEmailsReached = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.MAX_SECONDARY_EMAILS_REACHED,
+    message: 'You cannot have more than 10 secondary emails',
```
vbudhram: From the comments, I didn't see where the limit of 10 came from. That seems a bit high to me. I'm in favor of keeping this number fairly low, probably 3 or less. This could easily be increased after we gather stats about usage.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -622,6 +623,11 @@ module.exports = (
 
         await customs.check(request, primaryEmail, 'createEmail');
 
+        const account = await db.account(uid);
+        if (account.emails.length >= MAX_SECONDARY_EMAILS) {
```
vbudhram: I would add a quick note here that the max means emails records (unverified + verified).

packages/fxa-auth-server/lib/error.js
jodyheavener: Fair enough. 10 was an arbitrary choice. @davismtl you cool with 3?

jodyheavener:
Ack, I still need to sort out these new test failures. Sorry.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -622,6 +623,16 @@ module.exports = (
 
         await customs.check(request, primaryEmail, 'createEmail');
 
+        const account = await db.account(uid);
+        const secondaryEmails = account.emails.filter(
```
vbudhram: 👍

packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -157,6 +157,10 @@ var ERRORS = {
     errno: 140,
     message: t('Account already exists'),
   },
+  MAX_SECONDARY_EMAILS_REACHED: {
+    errno: 188,
+    message: t('You cannot have more than 3 secondary emails'),
```
vbudhram: I would probably say something like `You have reached the maximum allowed secondary emails`, to keep it generic.

packages/fxa-content-server/app/scripts/views/settings/emails.js
```diff
@@ -20,7 +20,11 @@ const t = msg => msg;
 
 const EMAIL_INPUT_SELECTOR = 'input.new-email';
 const EMAIL_REFRESH_SELECTOR = 'button.settings-button.email-refresh';
+const EMAIL_GROUP_SELECTOR = 'li.email-address';
 const EMAIL_REFRESH_DELAYMS = 350;
+// This is compared against all secondary email
+// records, both verified and unverified
+export const MAX_SECONDARY_EMAILS = 3;
```
vbudhram: You can pull this into our constants file, https://github.com/mozilla/fxa/blob/34b98ab954a86e288676c083d3d6984df2781b27/packages/fxa-content-server/app/scripts/lib/constants.js#L157.

vbudhram:
Looks like there was a change email functional test that failed :/

PR #4692, opened 2020-03-30
refactor(profile): use more async await in the profile server

vladikoff:
@vbudhram r?


vladikoff:
Let's bring this in later on..

PR #4694, opened 2020-03-30
task(subscriptions): remove references to plan.name and plan.nickname

### LINKED ISSUE #4693: Remove Usages of Plan Nickname and Plan Name
Remove all references and usages of Plan.nickname and the generated Plan.name in favor of using the Product.name + Plan.interval + Plan.interval_count data.

Acceptance Criteria:

* References to plan.nickname and plan.name will be removed
* Anything currently dependent upon using this data will be updated to use the product.name instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1441)
┆Issue Number: FXA-1441

jackiemunroe:
Because
- with the introduction of longer duration subscriptions, the current plan_name string is no longer applicable. Instead of using this custom formatted name, we will use the product_name along with plan billing data (interval + interval_count)

fixes #4693

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -30,7 +30,6 @@ const stripe = require('stripe').Stripe;
 /**
  * @typedef AbbrevPlan
  * @property {string} plan_id
- * @property {string} plan_name
  * @property {Product['metadata']} plan_metadata
  * @property {string} product_id
  * @property {string} product_name
@@ -472,11 +471,8 @@ class StripeHelper {
         continue;
       }
 
-      const plan_name = await this.formatPlanDisplayName(item);
-
       plans.push({
         plan_id: item.id,
-        plan_name,
         plan_metadata: item.metadata,
         product_id: item.product.id,
         product_name: item.product.name,
@@ -648,24 +644,6 @@ class StripeHelper {
   }
 
   /**
-   *
-   * Returns the product name of a given product.
-   *
-   * @param {String | Product} product
-   * @returns {Promise<String>} product name
-   */
-  async getProductName(product) {
-    if (typeof product === 'string') {
-      const fetchedProducts = await this.allProducts();
-      const matchingProduct = fetchedProducts.find(
-        fetchedProduct => fetchedProduct.product_id === product
-      );
-      return matchingProduct ? matchingProduct.product_name : '';
-    }
-    return product.name;
-  }
-
-  /**
    * Attempt to pay invoice by invoice id
    * Throws payment failed error on failure
    *
@@ -809,7 +787,10 @@ class StripeHelper {
         failure_message = /** @type {Charge} */ (charge).failure_message;
       }
 
-      const planName = await this.formatPlanDisplayName(sub.plan);
+      const product = await this.expandResource(sub.plan.product, 'Product');
+
+      const product_id = product.id;
+      const product_name = product.name;
 
       // FIXME: Note that the plan is only set if the subscription contains a single
       // plan. Multiple product support will require changes here to fetch all
@@ -820,8 +801,9 @@ class StripeHelper {
         current_period_start: sub.current_period_start,
         cancel_at_period_end: sub.cancel_at_period_end,
         end_at: sub.ended_at,
-        plan_name: planName,
         plan_id: sub.plan.id,
+        product_name,
+        product_id,
         status: sub.status,
         subscription_id: sub.id,
         failure_code,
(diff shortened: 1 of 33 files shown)


bbangert:
@jackiemunroe Looks like this needs a rebase.

packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -79,14 +79,7 @@ const SupportView = BaseView.extend({
...
 
-        const getPlansPromise = account
-          .fetchSubscriptionPlans()
-          .then(plans => {
-            this.model.set('plans', plans);
-          })
-          .catch(err => console.log(err));
-
-        return Promise.all([account.fetchProfile(), getPlansPromise]);
+        return Promise.all([account.fetchProfile()]);
```
bbangert, line 82: This doesn't need to be a Promise.all anymore.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -809,7 +805,18 @@ class StripeHelper {
         failure_message = /** @type {Charge} */ (charge).failure_message;
       }
 
-      const planName = await this.formatPlanDisplayName(sub.plan);
+      let product_id;
+      const product = sub.plan.product;
+
+      if (typeof product === 'string') {
```
lmorchard, line 811: I recently added [an `expandResource` utility function](https://github.com/mozilla/fxa/blob/ff24a44895b582f0039e60ef02f42de0d3118958/packages/fxa-auth-server/lib/payments/stripe.js#L949) for Stripe fetches that does essentially this. Also, [an `expandAbbrevProductForPlan` utility](https://github.com/mozilla/fxa/blob/ff24a44895b582f0039e60ef02f42de0d3118958/packages/fxa-auth-server/lib/payments/stripe.js#L962) to get Product details out of cache (i.e. id, name, and metadata). One of those might be handy here

packages/fxa-auth-server/test/remote/subscription_tests.js
```diff
@@ -235,19 +230,13 @@ describe('remote subscriptions:', function() {
...
-        let result = await client.getActiveSubscriptions(tokens[2]);
+        const result = await client.getActiveSubscriptions(tokens[2]);
         assert.isArray(result);
         assert.lengthOf(result, 1);
         assert.equal(result[0].createdAt, date * 1000);
         assert.equal(result[0].productId, PRODUCT_ID);
         assert.equal(result[0].uid, client.uid);
         assert.isNull(result[0].cancelledAt);
-
-        result = await client.account();
```
lmorchard, line 246: Is there a separate test for this API call somewhere? Wondering why it's going away

packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -79,14 +79,7 @@ const SupportView = BaseView.extend({
...
 
-        const getPlansPromise = account
-          .fetchSubscriptionPlans()
-          .then(plans => {
-            this.model.set('plans', plans);
-          })
-          .catch(err => console.log(err));
-
-        return Promise.all([account.fetchProfile(), getPlansPromise]);
+        return Promise.resolve(account.fetchProfile());
```
lmorchard, line 82: Could probably just `return account.fetchProfile` directly here, since it returns a promise.

packages/fxa-auth-server/test/remote/subscription_tests.js
jackiemunroe: I can add that back. I had removed it because it was covered elsewhere and was inconsistently failing on my local.

packages/fxa-content-server/app/scripts/views/support.js
jackiemunroe: Can do

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -809,7 +787,10 @@ class StripeHelper {
         failure_message = /** @type {Charge} */ (charge).failure_message;
       }
 
-      const planName = await this.formatPlanDisplayName(sub.plan);
+      const product = await this.expandResource(sub.plan.product, 'Product');
```
jodyheavener, line 790: Should this be wrapped in try/catch in case the resource fails to be found for any reason?

packages/fxa-content-server/app/scripts/templates/support.mustache
```diff
@@ -33,13 +33,13 @@
             <form id="fxa-support" novalidate>
               <div class="support-field">
                 <div class="form-label">
-                  <label for="plan" id="plan-label">{{#t}}What are you contacting us about?{{/t}}</label>
+                  <label for="product" id="product-label">{{#t}}What are you contacting us about?{{/t}}</label>
```
jodyheavener, line 36: Nit: I don't think I see anywhere this ID is being used, might be able to ditch

packages/fxa-auth-server/lib/payments/stripe.js
jackiemunroe: If this resource fails to be found, we actually cannot display the information to the user - since we rely on the Product name. So, I think that this should result in an error. If we were to simply exclude Subscriptions from the response that we were unable to load, then we would be in a situation where the GUI is misleading the customer. Also, this situation would only occur in the event that Stripe is down, in which case, the user would be unable to manage their subscriptions at all.

Issue #488, opened 2018-10-24
Finish removing the "email" field in oauth /v1/verify response


rfk:
Migrated from https://github.com/mozilla/fxa-oauth-server/issues/493

The `/v1/verify` request currently returns an "email" field, giving the primary email address of the account owning the token.  It's marked as deprecated in our docs:

  https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#response-9

The implementation of this feature will be broken when we ship the ability to change your primary email.  We have no plans to fix it, so we should push ahead with completely removing this functionality from oauth-server.  This will involve:

* Auditing our server logs to see if anyone is still requesting this field; it will show up as "email.requested" warnings logs from [1].
  * If it is still being used by some reliers, getting them to stop!
* Removing support for returning the email address in the `/verify` route route
* Remove logic which caches the email address from the `/authorization` route
* Drop the `email` column from `codes` and `tokens` and `refreshTokens` tables

Given that it involves a DB migration and potential risk of breaking relier flows that are using the deprecated method, let's put this off until after train-100.  But I wanted to make sure we had this work on file.

[1] https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/verify.js#L29

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-665)
┆Issue Number: FXA-665

vladikoff:
continued in https://github.com/mozilla/fxa/issues/4706

PR #4705, opened 2020-03-30
refactor(oauth): finish removing the "email" field in oauth /v1/verify response

vladikoff:
Issue #488

One follow up to this would be to:

* Drop the email column from codes and tokens and refreshTokens tables


vladikoff:
@rfk r? did I miss any remaining changes for this?

PR #4707, opened 2020-03-31
fix(profile): add missing await in cache delete

vladikoff:
@vbudhram r?


vbudhram:
Does this need to land in current train?

vladikoff:
> Does this need to land in current train?

I don't think so, seems minor!

PR #4708, opened 2020-03-31
feat(stats): Add statsd block rate reporting

### LINKED ISSUE #4653: Add statsd reporting to customs-server
We used to be able to view the customs-server metrics in Kibana, but that went away. Since most of FxA servers already use statsd, we should update the customs-server to use that and create a minimum dashboard.

We should be able to track

* Block rates
* All endpoint performance

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1412)
┆Issue Number: FXA-1412

vbudhram:
Fixes #4653 

This PR adds statsd reporting for customs server block rates. I opted to add this into the auth-server library that interacts with the customs-server because it is already configured and required less changes.

Running `nc -klvu 127.0.0.1 8125`, reports values as
```
fxa-auth-server.customs.request.check:1|c|#action:accountLogin,block:true,unblock:true,suspect:false,blockReason:other
fxa-auth-server.customs.request.check:1|c|#action:sendUnblockCode,block:true,unblock:true,suspect:false,blockReason:other
```

### DIFF
--- a/packages/fxa-auth-server/lib/customs.js
+++ b/packages/fxa-auth-server/lib/customs.js
@@ -11,20 +11,20 @@ const localizeTimestamp = require('../../fxa-shared').l10n.localizeTimestamp({
   supportedLanguages: config.get('i18n').supportedLanguages,
   defaultLanguage: config.get('i18n').defaultLanguage,
 });
+const serviceName = 'customs';
 
 module.exports = function(log, error, statsd) {
   const CustomsAPI = createBackendServiceAPI(
     log,
     config,
-    'customs',
+    serviceName,
     {
       check: {
         path: '/check',
         method: 'POST',
         validate: {
           payload: {
-            email: Joi.string()
-              .required(),
+            email: Joi.string().required(),
             ip: Joi.string().required(),
             action: Joi.string().required(),
             headers: Joi.object().optional(),
@@ -81,8 +81,7 @@ module.exports = function(log, error, statsd) {
         method: 'POST',
         validate: {
           payload: {
-            email: Joi.string()
-              .required(),
+            email: Joi.string().required(),
             ip: Joi.string().required(),
             errno: Joi.number().required(),
           },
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/customs.js
```diff
@@ -151,9 +149,35 @@ module.exports = function(log, error, statsd) {
...
+
+      if (options.suspect !== undefined) {
+        tags['suspect'] = suspect;
+      }
+
+      if (options.blockReason !== undefined) {
+        tags['blockReason'] = blockReason;
+      }
+
+      statsd.increment([serviceName, name].join('.'), tags);
```
vbudhram: @chenba Any objections to doing statsd reporting like this or anything I could optimize?

chenba: That LGTM.  As long as the potential set of values for the tags are pretty limited.  I believe they are, but I don't know the customs server well enough to say.

An aside: not sure which is better, array `join` or a template literal.

packages/fxa-auth-server/test/local/customs.js
```diff
@@ -6,42 +6,62 @@
...
-let customsNoUrl;
-let customsWithUrl;
-let customsInvalidUrl;
-
 const customsServer = nock(CUSTOMS_URL_REAL).defaultReplyHeaders({
   'Content-Type': 'application/json',
 });
 
 describe('Customs', () => {
+  let customsNoUrl;
```
vbudhram, line 24: Did some refactoring here to make tests a bit more independent.

packages/fxa-auth-server/test/local/customs.js
```diff
@@ -6,42 +6,62 @@
...
+    customsModule = require(`${ROOT_DIR}/lib/customs.js`)(log, error, statsd);
+
+    request = newRequest();
+    ip = request.app.clientAddress;
+    email = newEmail();
+    action = newAction();
+  });
+
+  afterEach(() => {
+    sandbox.restore();
```
chenba, line 57: A new sandbox is created in `beforeEach` so I don't think this is necessary.

Issue #3208, opened 2019-10-31
Payments cannot connect to React Dev Tool


LZoog:
We should be able to use React Dev Tools in the payments server for easier development and debugging - it shows a view of the component tree, state, component props, and highlights rerenders. I see this message when clicking on the extension:

<img width="428" alt="image" src="https://user-images.githubusercontent.com/13018240/67957863-3dc51a00-fbc4-11e9-8145-b12c749222d9.png">

Looks like we may need to add it as a project dependency to work in our case (perhaps because we work with an iFrame), see "Usage with React DOM": https://github.com/facebook/react/tree/master/packages/react-devtools

Let's avoid ejecting to customize Webpack (or for anything else) if at all possible.

lmorchard:
Starting to dig into this a bit, and I think it's our CSP restrictions that are blocking [the `window.__REACT_DEVTOOLS_GLOBAL_HOOK__` property](https://github.com/facebook/react/blob/master/packages/react-devtools-extensions/src/injectGlobalHook.js) that the React Dev Tools are looking for:
![image](https://user-images.githubusercontent.com/21687/68159293-5c9a1800-ff06-11e9-8bf5-1f44b4cd55cc.png)

lmorchard:
Yeah, when [I disable CSP altogether on my local server](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/server/config/index.js#L36), I get React Dev Tools working:

![image](https://user-images.githubusercontent.com/21687/68159760-3fb21480-ff07-11e9-8d44-63e32aa97e33.png)

Looks like [this is probably a Firefox-specific bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1267027#c48) where our CSP blocks the script injected by an extension.

PR #4720, opened 2020-04-01
fix(admin panel): make HMR work on port 8091, loosen up CSP restrictions when in development

jodyheavener:
With this change you would not longer need to develop on port 8092; everything can work on 8091.

_(It's worth noting that you can still work from 8092 if you need/want to; everything should still be working properly over there. This change just facilitates the ability to combine both the React app with the Express server **while in development**.)_

- Permit both unsafe-inline styles and the development websocket address source for scripts in CSP, only in development.
- Modify `react-scripts start` to always look for the websocket on port 8092, instead of the current page's port. This, plus the CSP change, permits you to run everything (the express server and HMR) all on port 8091 when in development.

Additionally, I've removed `rel="stylesheet"` from static HTML page's style tag to address mime type sniffing console error.

### DIFF
--- a/packages/fxa-admin-panel/public/index.html
+++ b/packages/fxa-admin-panel/public/index.html
@@ -11,7 +11,7 @@
       name="viewport"
       content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes"
     />
-    <link rel="stylesheet" href="styles/main.css" />
+    <link href="styles/main.css" />
   </head>
   <body>
     <noscript>The FxA Admin Panel requires JavaScript.</noscript>
--- a/packages/fxa-admin-panel/server/lib/csp/blocking.ts
+++ b/packages/fxa-admin-panel/server/lib/csp/blocking.ts
@@ -18,6 +18,7 @@ function getOrigin(link: string) {
  */
 export default function cspBlocking(config: { [key: string]: any }) {
   const ADMIN_SERVER = getOrigin(config.get('servers.admin.url'));
+  const HMR_WS_SERVER = getOrigin(config.get('servers.hmr_websocket.url'));
   const CDN_URL = config.get('staticResources.url');
   const PUBLIC_URL = config.get('listen.publicUrl');
   const HOT_RELOAD_WEBSOCKET = PUBLIC_URL.replace(/^http/, 'ws');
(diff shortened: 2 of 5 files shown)


packages/fxa-admin-panel/server/lib/csp/blocking.ts
```diff
@@ -35,16 +37,25 @@ export default function cspBlocking(config: { [key: string]: any }) {
...
+  // at a different port in development so you can access 8091 with HMR
+  if (process.env.NODE_ENV === 'development') {
+    connectSrc.push(HMR_WS_SERVER);
+    styleSrc.push(UNSAFE_INLINE);
+  }
+
   const rules = {
     directives: {
-      connectSrc: [SELF, ADMIN_SERVER],
+      connectSrc: connectSrc,
```
bbangert, line 51: Can't this just be `connectSrc,`?

LZoog:
This is awesome, thanks so much🙌

A couple of comments:
* This appears to refresh the page automatically on `:8091` with changes, but doesn't actually get HMR working entirely. HMR keeps the state in tact. I [pushed a couple line changes to my fork](https://github.com/mozilla/fxa/compare/jh/admin-panel-fixes...LZoog:jh/admin-panel-fixes-state?expand=1), check that link - for HMR to be true we want to be able to update the state by clicking on that button and then changing something else on the page and for the state to stay changed.
* Something else that would make this PR even _better_ would be to allow React dev tools during development. See [this issue](https://github.com/mozilla/fxa/issues/3208). However, that's a little out of scope.

If you'd like, you could open a separate PR with the CSP updates (plus that MIME type fix), and take a stab at HMR in a follow up (not uber high priority, but if that's fixed in admin panel, we can use the same method in payments as well). R+ for CSP changes. :)

packages/fxa-admin-panel/public/index.html
```diff
@@ -11,7 +11,7 @@
       name="viewport"
       content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes"
     />
-    <link rel="stylesheet" href="styles/main.css" />
```
lmorchard, line 14: Do we not need this `rel` attribute?

jodyheavener: As far as I can tell, no, most browsers should download and handle it just fine. I removed it because were encountering [this Express issue](https://stackoverflow.com/q/44657829/717633). I'm definitely open to alternatives, this was just a quick and dirty solution I found.

PR #4730, opened 2020-04-02
fix(admin panel): permit unsafe-inline styles when in development

jodyheavener:
This permits you to use port 8091 in development without CSP violations.

Also remove extraneous stylesheet reference.

In place of #4720

### DIFF
--- a/packages/fxa-admin-panel/public/index.html
+++ b/packages/fxa-admin-panel/public/index.html
@@ -11,7 +11,6 @@
       name="viewport"
       content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes"
     />
-    <link rel="stylesheet" href="styles/main.css" />
   </head>
   <body>
     <noscript>The FxA Admin Panel requires JavaScript.</noscript>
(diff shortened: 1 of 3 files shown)


packages/fxa-admin-panel/public/index.html
```diff
@@ -11,7 +11,6 @@
       name="viewport"
       content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes"
     />
-    <link rel="stylesheet" href="styles/main.css" />
```
jodyheavener, line 14: @LZoog It occurred to me to see what this stylesheet was doing, and.. I don't think it does anything? In development it's not used because all the styles are rendered in inline style tags, and in a production build the chunked version gets appended to the head. Do you know if we actually need this?

packages/fxa-admin-panel/server/lib/csp/blocking.ts
```diff
@@ -35,6 +36,13 @@ export default function cspBlocking(config: { [key: string]: any }) {
     return target;
   }
 
+  let styleSrc = [SELF];
```
chenba, line 39: nit: could narrow the type of `styleSrc` to `("'self'" | "'unsafe-inline'")[]` so it's not simply `string[]`.

packages/fxa-admin-panel/public/index.html
LZoog: Huh... I think you're right. We don't have it in payments. TIL

PR #4735, opened 2020-04-02
WIP: typescript fxa-auth-client

dannycoates:
a drop-in replacement for fxa-js-client


dannycoates:
I think #4753 is the better approach

PR #4736, opened 2020-04-02
refactor(emails): move all email normalization and equality checks to helper functions

### LINKED ISSUE #4644: Abstract "normalize an email address" and "check email addresses for equality" into helper functions
The auth-server code currently normalizes email address by lowercasing them. The most obvious place is [here](https://github.com/mozilla/fxa/blob/2c85d557255de034226e530c34b0c041c366d0ec/packages/fxa-auth-server/lib/db.js#L85), where we calculate the (immutable) value of the `normalizedEmail` column when creating a new account. But this "normalization == lowercase" assumption is spread out in several places in the codebase:

```
rfk@pomelo:fxa-auth-server(master *)$ rg -i "email.tolowercase" lib/
lib/db.js
85:    data.normalizedEmail = data.email.toLowerCase();

lib/routes/account.js
754:          const lowerCaseEmail = account.primaryEmail.normalizedEmail.toLowerCase();

lib/routes/password.js
457:              accountRecord.primaryEmail.normalizedEmail !== email.toLowerCase()

lib/routes/emails.js
311:              userEmail => userEmail.normalizedEmail === email.toLowerCase()
617:          normalizedEmail: email.toLowerCase(),
633:        if (sessionToken.email.toLowerCase() === email.toLowerCase()) {
765:        await db.deleteEmail(uid, email.toLowerCase());
771:            item.normalizedEmail === email.toLowerCase() && item.isVerified
782:          if (item.normalizedEmail !== email.toLowerCase()) {
943:          userEmail => userEmail.normalizedEmail === email.toLowerCase()
1016:          userEmail => userEmail.normalizedEmail === email.toLowerCase()

lib/routes/utils/signin.js
85:        originalLoginEmail.toLowerCase() !==
```

We accept unicode email addresses, and normalizing unicode is hard. See https://github.com/mozilla/fxa/issues/4643 for one example of how it can bite us, but there are many.

I think it would be prudent to factor this out into some sort of `normalizeEmailAddress` helper function, both in preparation for a day when we might discover we need to do something more nuanced, and to encourage our future selves to remember that it's not necessarily as simple as it might appear.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1405)
┆Issue Number: FXA-1405

jodyheavener:
Closes #4644

~I ended up using `String.localeCompare` to check email equality, based on the responses to this [Stack Overflow question](https://stackoverflow.com/q/2140627/717633) ([this response](https://stackoverflow.com/a/51005583/717633) is particularly helpful).~

[For now](https://github.com/mozilla/fxa/pull/4736#discussion_r402647434) we're going to use the regular `toLowerCase` comparison.

In case you're curious, I put together a comparison of the three common methods for string comparison here: https://codepen.io/jodyheavener/pen/rNVbvjg

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -11,6 +11,10 @@ const isA = require('@hapi/joi');
 const random = require('../crypto/random');
 const Sentry = require('@sentry/node');
 const validators = require('./validators');
+const {
+  emailsMatch,
+  normalizeEmail,
+} = require('../../../fxa-shared/email/helpers');
 
 const HEX_STRING = validators.HEX_STRING;
 const MAX_SECONDARY_EMAILS = 3;
@@ -308,8 +312,8 @@ module.exports = (
           if (email) {
             // If an email address is specified in payload, this is a request to verify
             // a secondary email. This should return the corresponding email code for verification.
-            const foundEmail = emailData.find(
-              userEmail => userEmail.normalizedEmail === email.toLowerCase()
+            const foundEmail = emailData.find(userEmail =>
+              emailsMatch(userEmail.normalizedEmail, email)
             );
 
             // This user is attempting to verify a secondary email that doesn't belong to the account.
@@ -615,7 +619,7 @@ module.exports = (
         const { email, verificationMethod } = request.payload;
         const emailData = {
           email: email,
-          normalizedEmail: email.toLowerCase(),
+          normalizedEmail: normalizeEmail(email),
           isVerified: false,
           isPrimary: false,
           uid: uid,
@@ -641,7 +645,7 @@ module.exports = (
           throw error.unverifiedSession();
         }
 
-        if (sessionToken.email.toLowerCase() === email.toLowerCase()) {
+        if (emailsMatch(sessionToken.email, email)) {
           throw error.yourPrimaryEmailExists();
         }
 
@@ -773,14 +777,12 @@ module.exports = (
           throw error.unverifiedSession();
         }
 
-        await db.deleteEmail(uid, email.toLowerCase());
+        await db.deleteEmail(uid, normalizeEmail(email));
         await db.resetAccountTokens(uid);
 
         // Find the email object that corresponds to the email being deleted
         const emailIsVerified = account.emails.find(item => {
-          return (
-            item.normalizedEmail === email.toLowerCase() && item.isVerified
-          );
+          return emailsMatch(item.normalizedEmail, email) && item.isVerified;
         });
 
         // Don't bother sending a notification if removing an email that was never verified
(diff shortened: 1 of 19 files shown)


jodyheavener:
Lol okay I definitely need to comb through everything a bit more.

packages/fxa-shared/email/helpers.js
```diff
@@ -0,0 +1,18 @@
...
+
+function emailsMatch(firstEmail, secondEmail) {
+  return (
+    firstEmail.localeCompare(secondEmail, undefined, {
+      // We use 'accent' instead of 'base' here to ensure a character with
+      // an accent is not considered the same as that character without one
+      sensitivity: 'accent',
+    }) === 0
+  );
+}
```
rfk, line 13: FWIW, I would suggest getting the tests passing and landing this without changing the way the comparison is done, and moving to `localeCompare` in a followup PR. I wouldn't be surprised to find some subtle change in behaviour somewhere in the system that flows from the change here. (Although I don't think we have too many tests with unicode emails, so perhaps such bustage wouldn't show up in tests anyway).

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -790,7 +792,7 @@ module.exports = (
 
         // Notify any verified email address associated with the account of the deletion.
         const emails = account.emails.filter(item => {
-          if (item.normalizedEmail !== email.toLowerCase()) {
+          if (emailsMatch(item.normalizedEmail, email)) {
```
rfk, line 795: This has turned a `!==` into an equals.

packages/fxa-auth-server/lib/routes/password.js
```diff
@@ -454,7 +455,7 @@ module.exports = function(
           .then(db.accountRecord.bind(db, email))
           .then(accountRecord => {
             if (
-              accountRecord.primaryEmail.normalizedEmail !== email.toLowerCase()
+              emailsMatch(accountRecord.primaryEmail.normalizedEmail, email)
```
rfk, line 458: Also `!==` here.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -644,7 +645,7 @@ const Account = Backbone.Model.extend(
           // originally used for login.
           if (
             options.originalLoginEmail &&
-            email.toLowerCase() !== options.originalLoginEmail.toLowerCase()
+            emailsMatch(email, options.originalLoginEmail)
```
rfk, line 648: And here :-)

jodyheavener: Yiiiiiikes. Guess the repetitiveness got to me. Thanks for this!

packages/fxa-shared/email/helpers.js
jodyheavener: That's fair - Unless any pushback, I'll leave this in here, but commented out so we retain the knowledge for the future.

rfk: Heh, I ordinarily wouldn't be...but personally I'm OK with leaving this in the code since it's such a small snippet; please consider also leaving a link to this discussion so future-us can easily rediscover the context.

jodyheavener:
Thanks for reviewing @rfk :)

PR #4740, opened 2020-04-02
fix(profile-server): updated graphicsmagick version in Dockerfile

dannycoates:


jrgm:
Actually, that version number is in two places in that Dockerfile.

PR #4750, opened 2020-04-03
feat(recovery keys): added upsertRecoveryKey db method

### LINKED ISSUE #4268: Consider using using upsertRecoveryKey database method
From https://github.com/mozilla/fxa/pull/4258/files/672e8c4e618646925678de6c2114c1d22f0d7726#diff-a4f2c3f852c9bcb0ac86a24d33a38b39

It might be a little bit cleaner if this was abstracted to the database and let it do the necessary cleanup for unconfirmed recovery keys.

farhan787:
fixes #4268

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
@@ -1,0 +1,63 @@
+-- This migration will help us to abstract the half baked recovery keys of fxa accounts. In this migration,
+-- we're adding a new `upsertRecoveryKey` procedure that will handle the half baked recovery keys by using
+-- the following criterias.
+
+-- 1. If a user has no recovery key then create a new recovery key.
+-- 2. If a user has unverified recovery key then delete that recovery key and create a new unverified
+--    recovery key.
+-- 3. If a user has verified recovery key then throw an exception
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('108');
+
+CREATE PROCEDURE `upsertRecoveryKey_1` (
+  IN `uidArg` BINARY(16),
+  IN `recoveryKeyIdHashArg` BINARY(32),
+  IN `recoveryDataArg` TEXT,
+  IN `createdAtArg` BIGINT UNSIGNED,
+  IN `enabledArg` BOOLEAN
+)
+BEGIN
+
+  DECLARE EXIT HANDLER FOR SQLEXCEPTION
+  BEGIN
+    RESIGNAL;
+  END;
+
+  -- Signal error if no user found
+  SET @accountCount = 0;
+  SELECT COUNT(*) INTO @accountCount FROM `accounts` WHERE `uid` = `uidArg`;
+  IF @accountCount = 0 THEN
+    SIGNAL SQLSTATE '45000' SET MYSQL_ERRNO = 1643, MESSAGE_TEXT = 'Can not create recovery key for unknown user.';
+  END IF;
+
+  -- Signal error if verified recovery key already exists
+  SET @verifiedRecoveryKeyCount = 0;
+  SELECT COUNT(*) INTO @verifiedRecoveryKeyCount FROM `recoveryKeys` WHERE `uid` = `uidArg` AND verifiedAt IS NOT NULL;
+  IF @verifiedRecoveryKeyCount = 1 THEN
+    SIGNAL SQLSTATE '45000' SET MYSQL_ERRNO = 1062, MESSAGE_TEXT = 'Verified recovery key already exists for this user.';
+  END IF;
+
+  DELETE FROM recoveryKeys WHERE uid = uidArg AND verifiedAt IS NULL;
+
+  INSERT INTO recoveryKeys (
+    uid,
+    recoveryKeyIdHash,
+    recoveryData,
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
```diff
@@ -0,0 +1,97 @@
...
+-- 1. If a user has no recovery key then create a new recovery key.
+-- 2. If a user has unverified recovery key then delete that recovery key and create a new unverified
+--    recovery key.
+-- 3. If a user has verified recovery key then throw an exception
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('108');
+
+-- Delete these below 3 lines
```
vbudhram, line 14: We typically don't have drops for procedures that were created in same migration.

packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
```diff
@@ -0,0 +1,97 @@
...
+  IN `enabledArg` BOOLEAN
+)
+BEGIN
+
+  DECLARE EXIT HANDLER FOR SQLEXCEPTION
+  BEGIN
+    RESIGNAL;
+  END;
+
+  SET @recoveryKeyCount = NULL;
```
vbudhram, line 78: I think another way we can look at this that might make the logic easier to follow is this:

1. Does user have a verified recovery key? If so throw error
2. Delete all unverified recovery keys
3. Create recovery key

packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
```diff
@@ -0,0 +1,97 @@
...
+  SET @isRecoveryKeyVerified = NULL;
+
+  SELECT COUNT(*) INTO @recoveryKeyCount FROM `recoveryKeys` WHERE `uid` = `uidArg`;
+  SELECT verifiedAt INTO @isRecoveryKeyVerified FROM `recoveryKeys` WHERE `uid` = `uidArg`;
+
+  IF @recoveryKeyCount = 0 THEN
+    CALL createRecoveryKey_5(`uidArg`, `recoveryKeyIdHashArg`, `recoveryDataArg`, `createdAtArg`, `enabledArg`);
```
vbudhram, line 85: Instead of calling into other stored procedures in here, I would just copy the insert part.

packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
```diff
@@ -0,0 +1,97 @@
...
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('108');
+
+-- Delete these below 3 lines
+DROP PROCEDURE IF EXISTS createRecoveryKey_5;
+DROP PROCEDURE IF EXISTS getRecoveryKey_5;
+DROP PROCEDURE IF EXISTS upsertRecoveryKey_1;
+
+CREATE PROCEDURE `createRecoveryKey_5` (
```
vbudhram, line 19: This could be removed since it is the same as `createRecoveryKey_4`

packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
farhan787: I forgot to remove that, that was for my testing purpose.

farhan787:
@vbudhram made the changes.

packages/fxa-auth-db-mysql/lib/db/schema/patch-108-109.sql
```diff
@@ -0,0 +1,63 @@
...
+  END IF;
+
+  -- Signal error if verified recovery key already exists
+  SET @verifiedRecoveryKeyCount = 0;
+  SELECT COUNT(*) INTO @verifiedRecoveryKeyCount FROM `recoveryKeys` WHERE `uid` = `uidArg` AND verifiedAt IS NOT NULL;
+  IF @verifiedRecoveryKeyCount = 1 THEN
+    SIGNAL SQLSTATE '45000' SET MYSQL_ERRNO = 1062, MESSAGE_TEXT = 'Verified recovery key already exists for this user.';
+  END IF;
+
+  DELETE FROM recoveryKeys WHERE uid = uidArg AND verifiedAt IS NULL;
```
vbudhram, line 42: @farhan787 Thanks for making the updates. I have reached out to our OPs team to see what values in the database look like for the `enabled` column. I suspect it will be easier to select keys based on that columns value (instead of `verifiedAt`).

farhan787: Are we using indexing for the `enabled` column ?

vbudhram: The index is on the `uid` column which we will still need to select on in addition to the `enabled` column.

farhan787: `enabled` column would not guarantee if a `recoveryKey` is verified or not, I've seen in my local db that `enabled` was true but `verifiedAt` was still storing `NULL`, but I don't know about production if there's some difference there.

PR #4760, opened 2020-04-03
feat(metrics): log raw Amplitude events in auth server

### LINKED ISSUE #4697: fxa-auth-server: Create raw amplitude event output
This should augment existing logging, and include necessary server details that can't be reconstructed later (geo data, timestamp if the client's is wrong, user-agent info, device id, etc).

It should be done where the amplitude logging event is handled before the transform/modification calls are done. Ideally by replacing the existing amplitude logging call with one that calls the existing logger, and the new raw event logger. This output should be controlled by a boolean server config flag that defaults to off.

A raw event in this case is the original event the log function was called with, without any additional transforms done as the metrics code does in the server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1444)
┆Issue Number: FXA-1444

chenba:
This patch logs "raw" Amplitude events in auth server to stdout along
with enough context for another service to transform the raw events into
Amplitude events identical to those emitted by fxa-amplitude-send.

Fixes #4697 (FXA-1444)

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/amplitude.js
+++ b/packages/fxa-auth-server/lib/metrics/amplitude.js
@@ -144,11 +144,16 @@ module.exports = (log, config) => {
 
   return receiveEvent;
 
-  async function receiveEvent(event, request, data = {}, metricsContext = {}) {
-    if (!event || !request) {
+  async function receiveEvent(
+    eventType,
+    request,
+    data = {},
+    metricsContext = {}
+  ) {
+    if (!eventType || !request) {
       log.error('amplitude.badArgument', {
         err: 'Bad argument',
-        event,
+        event: eventType,
         hasRequest: !!request,
       });
       return;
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/oauth/metrics/amplitude.js
```diff
@@ -59,13 +66,18 @@ module.exports = (log, config) => {
...
-        type: event,
-        time: data.time || Date.now(),
-      },
-      eventData
-    );
+    if (config.amplitude.rawEvents) {
+      const rawEvent = {
+        event,
+        context: {
+          eventSource: 'oauth',
```
chenba, line 73: Different name here because of the different event transform.

packages/fxa-auth-server/lib/metrics/amplitude.js
```diff
@@ -164,64 +169,122 @@ module.exports = (log, config) => {
...
+        context: {
+          ...picked,
+          eventSource: 'auth',
+          version: VERSION,
+          deviceId,
+          devices,
+          emailDomain: data.email_domain,
+          emailSender: data.email_sender,
+          emailService: data.email_service,
+          emailTypes: EMAIL_TYPES,
```
chenba, line 251: feelsbadman about copying this big dictionary.

PR #4763, opened 2020-04-06
WebCrypto based auth-client

dannycoates:
This replaces `fxa-js-client` in `fxa-content-server`. It uses native WebCrypto instead of sjcl. Since it's quite a bit smaller (in bytes) I moved it to the main bundle instead of a dynamic import. I tested support for older browsers that need shims in Edge 16 but we also need to verify on Android as well.

There's mocha tests for the new crypto functions. The api is only exercised from the functional tests, but they do cover it pretty well.

There's a little more cleanup to do in auth-server tests before we can remove fxa-js-client completely.


dannycoates:
I think we should still have a webcrypto polyfill, dynamically imported when needed, like we did on Send.

dannycoates:
_note to self_: squash and finish commit message

packages/fxa-content-server/app/scripts/lib/fxa-client.js
```diff
@@ -139,10 +136,10 @@ FxaClientWrapper.prototype = {
     if (this._client) {
       return Promise.resolve(this._client);
     }
-
-    return importFxaClient().then(FxaClient => {
-      const client = new FxaClient.default(this._authServerUrl);
-      this._client = wrapClientToNormalizeErrors(client);
+    return AuthClient.checkWebCrypto().then(() => {
```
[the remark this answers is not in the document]
dannycoates: I refactored this check into `AuthClient.create`. It seems like a better place for it.

packages/fxa-content-server/app/scripts/lib/auth/crypto.ts
```diff
@@ -0,0 +1,179 @@
...
+    quickStretchedRaw,
+    'HKDF',
+    false,
+    ['deriveBits']
+  );
+  const authPW = await crypto.subtle.deriveBits(
+    {
+      name: 'HKDF',
+      salt: new Uint8Array(0),
+      // @ts-ignore
```
dannycoates, line 40: The builtin typescript definition for HKDF is wrong. We need to ignore the `info` field for now

packages/fxa-content-server/app/scripts/lib/auth/crypto.ts
```diff
@@ -0,0 +1,179 @@
...
+    await crypto.subtle.digest(
+      'SHA-256',
+      crypto.getRandomValues(new Uint8Array(16))
+    );
+    return true;
+  } catch (err) {
+    try {
+      console.warn('loading webcrypto shim');
+      // prettier-ignore
+      // @ts-ignore
```
dannycoates, line 169: ignore the fact that `window` doesn't already have an `asmCrypto` property

bbangert: The other way to do this is to cast to any:
```
(window as any).asmCrypt =....
```

PR #4764, opened 2020-04-06
fix(build): change fxa-js-client resolution in content-server

dannycoates:
Due to fxa-js-client being a package.json dependency in content-server
and lerna/renovate trying to be "helpful" there was inconsistency
with how the package was resolved based on the environment. In Docker
and CI it loaded from npm but in local dev via an npm link to the
monorepo source. This change just stops using npm resolution at all.


packages/fxa-js-client/package.json
```diff
@@ -5,13 +5,13 @@
   "author": "Mozilla",
   "license": "MPL-2.0",
   "scripts": {
+    "postinstall": "grunt sjcl",
```
vladikoff, line 8: nit: maybe add `publish: false`, because after this changed is added the release cannot be published anymore (it will fail to install via npm)

6a68:
🎉

PR #4773, opened 2020-04-06
refactor(payments): remove ttl for customer caching

### LINKED ISSUE #4651: Create a refresh process for our cached stripe customers
Currently we cache the customer record from Stripe into Redis which we use to verify a user has a subscription. We include a TTL on this, but then we have no process for regularly updating this, which means when the TTL expires we will lose the cache copy and erroneously indicate the user has no subscription.

We **could** use a Stripe invoice event to refresh customers and their TTL when they are charged, but our TTL is 3 months, and we will soon be offering yearly subscriptions so that won't handle the longer duration period. Alternatively, we could drop the use of the TTL, and instead purge records when someone cancels/deletes.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1409)
┆Issue Number: FXA-1409

jackiemunroe:
- remove ttl for cusomer caching
- ensure customer cache entry removal on account deletion

fixes #4651

### DIFF
--- a/packages/fxa-auth-server/config/index.js
+++ b/packages/fxa-auth-server/config/index.js
@@ -716,12 +716,6 @@ const conf = convict({
       format: String,
       doc: 'A shared secret to authenticate Stripe webhook requests',
     },
-    stripeCustomerCacheTtlSeconds: {
-      doc: 'The number of seconds to cache a Stripe customer record',
-      format: 'int',
-      default: 3 * 31 * 24 * 60 * 60, // ~3 months
-      env: 'STRIPE_CUSTOMER_CACHE_TTL_SECONDS',
-    },
     transactionalEmails: {
       // See also: https://jira.mozilla.com/browse/FXA-1148
       enabled: {
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -71,54 +71,6 @@ function validateProductUpgrade(oldMetadata, newMetadata) {
   return oldOrder < newOrder;
 }
 
-/**
- * Get a cached result at a cache key and regenerated it with `refreshFunction`
- * if its expired.
- *
- * TODO fix logging messages: as it stands there is no name for the async functions
- * being passed in, and type safety prevents accessing the caller, so right now it
- * has been removed from the log messages
- *
- * @template T
- * @param {*} redis
- * @param {string} cacheKey
- * @param {number} cacheTtl
- * @param {() => Promise<T> | undefined} refreshFunction
- * @param {boolean} forceRefresh
- * @returns {Promise<T | undefined>} possibly cached result
- */
-async function cachedResult(
-  log,
-  redis,
-  cacheKey,
-  cacheTtl,
-  refreshFunction,
-  forceRefresh = false
-) {
-  if (cacheTtl && !forceRefresh) {
-    try {
-      const json = await redis.get(cacheKey);
-      if (json) {
-        return JSON.parse(json);
-      }
-    } catch (err) {
-      log.error(`subhub.cachedResult.getCachedResponse.failed`, { err });
-    }
-  }
-
-  if (!refreshFunction) {
-    return;
-  }
-
-  const result = await refreshFunction();
-  if (cacheTtl) {
-    redis.set(cacheKey, JSON.stringify(result), 'EX', cacheTtl).catch(err => {
-      log.error(`subhub.cachedResult.setCacheResponse.failed`, { err });
-    });
-  }
-  return result;
-}
-
 class StripeHelper {
   /**
    * Create a Stripe Helper with built-in caching.
@@ -132,8 +84,6 @@ class StripeHelper {
     this.cacheTtlSeconds =
       config.subhub.plansCacheTtlSeconds ||
       config.subscriptions.cacheTtlSeconds;
-    this.customerCacheTtlSeconds =
-      config.subscriptions.stripeCustomerCacheTtlSeconds;
     this.webhookSecret = config.subscriptions.stripeWebhookSecret;
     const redis =
       this.cacheTtlSeconds &&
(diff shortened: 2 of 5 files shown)


packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -160,6 +110,39 @@ class StripeHelper {
...
 
+  /**
+   * Get a cached result at a cache key and regenerated it with `refreshFunction`
+   * if its expired.
+   *
+   * @param {string} cacheKey
+   * @param {() => Promise<T>} refreshFunction
+   * @returns {Promise<T | undefined>} possibly cached result
+   */
+  async cachedResult(cacheKey, refreshFunction) {
```
chenba, line 121: To be consistent with the other function names in here, this should start with verb.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -145,6 +144,71 @@ describe('StripeHelper', () => {
...
+      let actualResult = await stripeHelper.cachedResult(
+        cacheKey,
+        refreshFunction
+      );
+      assert(mockRedis.get.calledOnce);
+      assert(mockRedis.set.calledOnce);
+      assert(refreshFunction.calledOnce);
+      assert.deepEqual(actualResult, expectedResult);
+
+      refreshFunction.resetHistory();
```
chenba, line 168: A little confusing here that `resetHistory` is called on `refreshFunction` but not `mockRedis.get` or `mockRedis.set`.  `assert(refreshFunction.calledOnce)` would tell us that it wasn't called again without the reset.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -145,6 +144,71 @@ describe('StripeHelper', () => {
...
+          refreshFunction
+        );
+
+        assert(
+          log.error.calledOnceWith('stripeHelper.cachedResult.redis.set.failed')
+        );
+        assert.deepEqual(actualResult, expectedResult);
+      });
+    });
+  });
```
chenba, line 210: I don't see any tests verifying that redis `get` and `set` were called with the correct arguments.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -927,32 +991,48 @@ describe('StripeHelper', () => {
...
+          assert(log.error.notCalled);
+          assert.isUndefined(await mockRedis.get(customerKey));
+        });
+
+        it('does not fail if key does not exist', async () => {
+          const key = stripeHelper.customerCacheKey('test-test-test', email);
+          assert.isUndefined(await mockRedis.get(key));
+
+          await stripeHelper.removeCustomerFromCache('test-test-test', email);
+          assert(mockRedis.del.calledOnce);
```
chenba, line 1010: Could add an assertion on the argument passed to `mockRedis.del`.

packages/fxa-auth-server/config/index.js
```diff
@@ -717,6 +717,7 @@ const conf = convict({
       doc: 'A shared secret to authenticate Stripe webhook requests',
     },
     stripeCustomerCacheTtlSeconds: {
+      // Deprecated
```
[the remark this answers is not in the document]
jackiemunroe: Sounds good. I wasn't sure if it was okay to remove it 😅

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -145,6 +144,85 @@ describe('StripeHelper', () => {
...
+    });
+
+    it('follows read-through caching logic', async () => {
+      // On first call: cache miss, call refresh function, cache the result
+      let actualResult = await stripeHelper.getCachedResult(
+        cacheKey,
+        refreshFunction
+      );
+      assert(mockRedis.get.calledOnceWith, { cacheKey });
+      assert(mockRedis.set.calledOnceWith, { cacheKey });
```
chenba, line 164: Doesn't the `set` call take more than one argument?  At the very least, the key and the value?

jackiemunroe: yes, it does take the key and the value. I can add the value to the argument check

chenba: ooh hmm, I think the reason that it was passing is because `assert(mockRedis.set.calledOnceWith, { cacheKey });` isn't actually calling `mockRedis.set.calledOnceWith`.

packages/fxa-auth-server/config/index.js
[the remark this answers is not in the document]
lmorchard: I think it's okay to remove, but also with following up by adding a note to the deploy doc to remove the config on the server for the train where this will ship out

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -160,6 +110,39 @@ class StripeHelper {
...
+      this.log.error(`stripeHelper.getCachedResult.redis.get.failed`, { err });
+    }
+
+    const result = await refreshFunction();
+    try {
+      await this.redis.set(
+        cacheKey,
+        JSON.stringify(result),
+        'EX',
+        this.cacheTtlSeconds
```
lmorchard, line 137: Do we *never* want to have different cache times for different data? I guess we could add the parameter back in if/when we end up doing that again

jackiemunroe: As it stands, there is only one ttl defined, and I think that it's best to reduce the overhead of passing additional parameters when unnecessary. I actually would like to restructure how we are interacting with Stripe objects and caching entirely. I am working on a proposal for that.

jackiemunroe:
@chenba I've made some updates/fixes to the tests. Can you take another look?

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -145,6 +144,91 @@ describe('StripeHelper', () => {
...
+        sandbox.stub(mockRedis, 'get').rejects(Error);
+
+        const actualResult = await stripeHelper.getCachedResult(
+          cacheKey,
+          refreshFunction
+        );
+
+        assert.isTrue(mockRedis.get.calledOnceWith(cacheKey));
+        assert.isTrue(
+          mockRedis.set.calledOnceWith(cacheKey, JSON.stringify(expectedResult))
```
chenba, line 197: :( I'm not as familiar with the Sinon API as I thought.  `calledOnceWith` only partially matches the args you passed it.  I think we should use `calledOnceWithExactly`.

PR #4774, opened 2020-04-07
feat(metrics): add metrics to amplitude events

### LINKED ISSUE #4656: Augment amplitude logging with statsd metrics
So that we know what the overall throughput of transforms/logs/drops looks like from our services that log amplitude metrics (auth-server, content-server, payments). They should all be augmented with statsd metrics to record metrics regarding:

* Messages filtered/dropped by the transform process
* Messages logged

This is to help us know the overall change in logging we will experience when we begin logging out the raw event for later transform/filter/drop in fxa-metrics workers.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1414)
┆Issue Number: FXA-1414

bbangert:
Because:

* We want to know how many amplitude events are dropped by the
  transform step.
* Creating ad-hoc non-uniform singletons is inconsistent and difficult
  to effectively test.

This commit:

* Add's metrics to amplitude events in content-server, auth-server, and
  payments-server.
* Uses typedi for Dependency Injection for a consistent mechanism to set
  singleton type objects and retrieve them.

Closes #4656

### DIFF
--- a/packages/fxa-auth-server/test/local/metrics/amplitude.js
+++ b/packages/fxa-auth-server/test/local/metrics/amplitude.js
@@ -6,6 +6,10 @@
 
 const { assert } = require('chai');
 const { version } = require('../../../package.json');
+const { StatsD } = require('hot-shots');
+const { Container } = require('typedi');
+const sinon = require('sinon');
+
 const amplitudeModule = require('../../../lib/metrics/amplitude');
 const mocks = require('../../mocks');
 const mockAmplitudeConfig = {
@@ -117,6 +121,8 @@ describe('metrics/amplitude', () => {
 
     describe('raw events enabled', () => {
       it('logged a raw event', async () => {
+        const statsd = { increment: sinon.spy() };
+        Container.set(StatsD, statsd);
         mockAmplitudeConfig.rawEvents = true;
         const now = Date.now();
         await amplitude(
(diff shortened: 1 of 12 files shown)


packages/fxa-auth-server/test/local/metrics/amplitude.js
```diff
@@ -225,12 +231,16 @@ describe('metrics/amplitude', () => {
           event: expectedEvent,
           context: expectedContext,
         });
+        sinon.assert.calledTwice(incrementSpy);
```
dannycoates, line 234: There's a mixed approach to referencing the increment spy in this file. This one looks like the odd one out.

bbangert: That's true, a bunch of the others are using `assert.isTrue` with sinon, which will tell you nothing of utility if it fails except it was false. Using the sinon one gets the chai assert behavior and if it fails, it tells you useful information like the count difference, or what arg it was called with that has a mismatch. They should probably all be changed to using the more useful assert style with sinon, but that's a bit outside the scope of this PR.

How to use chai+sinon effectively to get useful feedback when asserts fail should probably be in a style guide for our code-base. Below on line 317 is an `assert.ok` on a boolean expression. If it fails, we have nothing of use. Instead it should use something like chai's `assert.isAbove` to verify one value is above another. It would then indicate the values in a failure.

dannycoates: I agree... but I was just referring to `incrementSpy` vs `statsd.increment`

bbangert: Oh, in this case the container set was happening in the same function, that's why I did it this way. In the other tests the container is setup in a `beforeEach`, so it has to be retrieved by getting it again.

bbangert: I can make it look consistent though. :)

PR #4798, opened 2020-04-08
feat(admin panel): add apollo support and basic UI for querying accounts

### LINKED ISSUE #4205: [admin panel A1.0] Add `apollo-client` for React
This will allow us to send and process graphQL queries/mutations.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1144)
┆Issue Number: FXA-1144

### LINKED ISSUE #4208: [admin panel A1.0] Add input to UI for email query, send/process response
We want to create an input on the UI that on submission, sends a query to the GraphQL endpoint with the email address in question, then receives and displays the response.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1145)
┆Issue Number: FXA-1145

### LINKED ISSUE #4776: Webpack content embeds 127.0.0.1
From our ops meeting:

> Should have filed this as a bug, but the webpack content embeds http://127.0.0.1:3030 in the content. Could I have that with just relative URL?

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1499)
┆Issue Number: FXA-1499

jodyheavener:
Closes #4205 
Closes #4208
Closes #4776

In this PR:

- Server Config
- Adds apollo-client, graphQL query and handling of response
- Favicon, style improvements
- Test out querying
- Adds script for new email bounce


jodyheavener:
Looking into that weird test failure, but overall ready for review.

packages/fxa-admin-panel/server/lib/server.ts
```diff
@@ -71,13 +81,52 @@ app.get('/__version__', (_, res) =>
   res.type('application/json').send(JSON.stringify(version))
 );
 
+function injectMetaContent(html: any, metaContent: any = {}) {
```
chenba, line 84: These don't have to be `any`s.  We do know some type info on them from how they are used.  `html` has a `replace` function and `metaContent` has keys.  At the very least they should be types that satisfy those requirements.

packages/fxa-admin-panel/server/lib/server.ts
```diff
@@ -71,13 +81,52 @@ app.get('/__version__', (_, res) =>
...
+    result = result.replace(
+      k,
+      encodeURIComponent(JSON.stringify(metaContent[k]))
+    );
+  });
+
+  return result;
+}
+
+function injectHtmlConfig(html: any, config: any) {
```
chenba, line 97: No `any`s unless you truly don't know their types.

packages/fxa-admin-panel/server/lib/server.ts
```diff
@@ -71,13 +81,52 @@ app.get('/__version__', (_, res) =>
...
   const proxy = require('express-http-proxy');
-  app.use('/', proxy(proxyUrl));
+  app.use(
+    '/',
+    proxy(proxyUrl, {
+      userResDecorator: function(
+        proxyRes: any,
+        proxyResData: any,
+        userReq: any /*, userRes*/
+      ) {
```
chenba, line 116: `any`s but function body assume these to have certain properties.

packages/fxa-admin-panel/server/lib/server.ts
chenba: Looking at the calls, I think you could just let TS to infer the types here and at `injectMetaContent`?

packages/fxa-admin-server/src/test/lib/db/models/helpers.ts
```diff
@@ -45,12 +45,19 @@ export function randomEmailBounce(email: string) {
...
   normalizedEmail: string;
 }
 
+export interface BounceIsh {
+  bounceSubType: number;
+  bounceType: number;
+  createdAt: number;
+  email: string;
+}
+
```
chenba, line 60: Could `Pick` or `Omit` be helpful here to make the relationships between these `*ish` types and the "real" types easier to understand?  E.g. `BounceIsh = Pick<EmailBounce, 'bounceSubType' | 'bounceType' | 'createAt' | 'email'>`.  The types of the properties will also stay n*sync.

packages/fxa-admin-panel/src/lib/config.ts
```diff
@@ -0,0 +1,84 @@
+// This configuration is a subset of the configuration declared in server/config/index.ts
+// Which config is copied over is defined in server/lib/server.ts
```
chenba, line 2: Sounds like `Pick` (https://www.typescriptlang.org/docs/handbook/utility-types.html#picktk) can be used here.

packages/fxa-admin-panel/src/lib/config.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (!obj[key]) {
+        obj[key] = {};
+      }
+      merge(obj[key], value);
+    }
+  }
+  return obj;
+}
+
+export function updateConfig(newData: any) {
```
chenba, line 82: Can `newData` and `merge`'s `data` at least be an object type since `Object.entries` is called on it?

packages/fxa-admin-panel/src/lib/config.ts
jodyheavener: @chenba Hmm I'm having trouble with specifically this change (my other switches to Pick worked fine).

New setup looks like:

```js
import { default as ServerConfig } from '../../server/config';

export type Config = Pick<ServerConfig, 'env' | 'servers'>;
```

But TS is not liking this: `'ServerConfig' refers to a value, but is being used as a type here.`

Any idea how I can address this? I would have thought the types on `ServerConfig` would come through.

chenba: If the type is exported then import the type.  Otherwise, I think `Pick<typeof ServerConfig, ...` should work.

jodyheavener: Ah ok that did it, but that seems to have surfaced a new thing. Although the note indicates it's a subset of the noted config file, I don't think I can `Pick` from it because ServerConfig is an instance of Convict that relies on get and set methods to retrieve properties, while this new config is just a function that returns a standard object. Unless I'm missing something 🤔

chenba: Oh ok.  My comment was based on that code comment, so I didn't know it was a value.

packages/fxa-admin-panel/server/lib/server.ts
```diff
@@ -71,13 +81,55 @@ app.get('/__version__', (_, res) =>
...
   const proxy = require('express-http-proxy');
-  app.use('/', proxy(proxyUrl));
+  app.use(
+    '/',
+    proxy(proxyUrl, {
+      userResDecorator: function(
+        proxyRes: { headers: { [x: string]: any } },
+        proxyResData: { toString: (arg0: string) => any },
+        userReq: { url: string } /*, userRes*/
+      ) {
```
chenba, line 119: I don't think we need to declare our own types here for `proxyRes` and `userReq`; they are Request and Response from express.

Shouldn't `toString` always return a string?

packages/fxa-admin-panel/src/components/EmailBlocks/index.tsx
```diff
@@ -2,37 +2,100 @@
...
+              data,
+            }}
+          />
+        </>
+      ) : null}
     </div>
   );
 };
 
+const AccountSearchResult = ({ loading, error, data }: any) => {
```
chenba, line 91: Huh!  What happens when you destruct a null or undefined?  I know any really means any, but even then I feel like TS should say something here...

packages/fxa-admin-server/scripts/email-bounce.ts
```diff
@@ -0,0 +1,40 @@
...
+  let account: AccountIsh;
+
+  if (argv.email) {
+    bounce = randomEmailBounce(argv.email as string);
+  } else {
+    account = randomAccount();
+    bounce = randomEmailBounce(account.email);
+    const email = randomEmail(account);
+
+    await (Account as any).query().insertGraph({ ...account, emails: [email] });
```
chenba, line 30: Why `as any` and not some other more appropriate type here?

Issue #4649, opened 2020-03-25
Link to Adjust's Privacy Policy is a 404


data-sync-user:
On https://accounts.firefox.com/legal/privacy we have this content:

{quote}On iOS and Android: Firefox by default sends mobile campaign data to Adjust, our analytics vendor, which has its own [privacy policy](https://www.adjust.com/privacy_policy/). Mobile campaign data includes a Google advertising ID, IP address, timestamp, country, language/locale, operating system, and app version. [...]{quote}

The privacy policy link goes to https://www.adjust.com/privacy_policy/ which is a 404.  It looks like the correct link would be https://www.adjust.com/terms/privacy-policy/

Should we adjust (pun intended!) our link or is it possible to reword this to remove the link altogether?

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1407)
┆Issue Number: FXA-1407

apsychogirl:
Hi, Can I take this up?

clouserw:
Hi @apsychogirl .  I assigned this to myself because it's a little more complex than it appears (the fix isn't even in this repository).  If you're looking for issues, check out the Help Wanted tag.  Cheers

apsychogirl:
> Hi @apsychogirl . I assigned this to myself because it's a little more complex than it appears (the fix isn't even in this repository). If you're looking for issues, check out the Help Wanted tag. Cheers

Okay:)

data-sync-user:
➤ Chelsea Lewis commented:

[~wclouser@mozilla.com] reviewed with Legal today, we just need to update the link to https://www.adjust.com/terms/privacy-policy/ instead of privacy_policy.

data-sync-user:
➤ Wil Clouser commented:

Pull request is open at https://github.com/mozilla/legal-docs/pull/1424

data-sync-user:
➤ Wil Clouser commented:

The pull request landed.  Due to https://jira.mozilla.com/browse/FXA-1023 we need to manually update our legal docs hash, which I've opened in https://github.com/mozilla/fxa/pull/4809

data-sync-user:
➤ Wil Clouser commented:

Fixed in https://github.com/mozilla/fxa/pull/4809

Issue #4031, opened 2020-02-03
Legal docs are out of date again


data-sync-user:
[https://accounts.firefox.com/__version__](https://accounts.firefox.com/__version__) says the legal docs are `c99707983486f30fb4d1cbe10ea9d221fed28da1` which is a commit from last October.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1023)
┆Issue Number: FXA-1023

data-sync-user:
➤ Wil Clouser commented:

The CircleCI logs might have some hints about why this isn't pulling a newer version.

jrgm:
I had a look at circle logs, but it didn't show me a problem. But I pulled the image, and poked around. 

I think the issue may be this:
- package.json wants master - https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/package.json#L95
- but package-lock.json pins to #c99707... - https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/package-lock.json#L8319-L8322

Also, from within the container of mozilla/fxa-content-server:latest, I get: 
```
~ $ npm ls --depth 0 2>&1 | grep legal
+-- UNMET DEPENDENCY legal-docs@git://github.com/mozilla/legal-docs.git#c99707983486f30fb4d1cbe10ea9d221fed28da1
npm ERR! missing: legal-docs@git://github.com/mozilla/legal-docs.git#master, required by fxa-content-server@1.155.0
~ $ 
```
The module though does exist in the container under /app/node_modules.

data-sync-user:
➤ John Morrison commented:

I did the bump to get it back to master of this content repo in https://github.com/mozilla/fxa/pull/4034 ( https://github.com/mozilla/fxa/pull/4034 )

But needs a way to keep that version being bumped regularly to really fix this.

data-sync-user:
➤ Wil Clouser commented:

Thanks for John for fixing this in the short term, I'm kicking it out of the sprint.  The legal docs don't change very often and his explanation makes sense.

data-sync-user:
➤ Lauren Zugai commented:

This was resolved in a recent PR.

PR #4809, opened 2020-04-09
fix(docs): Update legal docs hash.  More detail in #4649

clouserw:


clouserw:
This brings us up to HEAD of https://github.com/mozilla/legal-docs/commits/master .  We're having to do this manually because of https://github.com/mozilla/fxa/issues/4031

Issue #975, opened 2019-05-01
Support OIDC "nonce" parameter


shane-tomlinson:
The OpenID Connect spec [defines a `nonce` query parameter](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) and response in the ID token:

> String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. If present in the ID Token, Clients MUST verify that the nonce Claim Value is equal to the value of the nonce parameter sent in the Authentication Request. If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request. Authorization Servers SHOULD perform no other processing on nonce values used. The nonce value is a case sensitive string. 

I'm not yet sure what all is involved in making this happen. It might be that we need to store and propagate the nonce with any grants and refresh tokens so that any id_tokens that are generated contain the nonce.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-815)
┆Issue Number: FXA-815

data-sync-user:
➤ Wil Clouser commented:

Thank you for filing this issue. This is not currently on our roadmap, and in an effort to focus our work, we are closing old issues that we are unlikely to be closed in the future. Thanks again.

PR #4818, opened 2020-04-09
Add `id_token_hint` support to `prompt=none` flow

6a68:
Adds a server-side id token verifier and threads the token verification
step through fxa-js-client and the account and oauth relier models.

### DIFF
--- a/packages/fxa-auth-server/docs/oauth/api.md
+++ b/packages/fxa-auth-server/docs/oauth/api.md
@@ -303,8 +303,7 @@ content-server page.
   - `signin` triggers the signin flow. (will become deprecated and replaced by `email`)
   - `signup` triggers the signup flow. (will become deprecated and replaced by `email`)
   - `force_auth` requires the user to sign in using the address specified in `email`.
-- `email`: Optional if `action` is `email`, `signup` or `signin`. Required if `action`
-  is `force_auth` or `prompt=none`.
+- `email`: Optional if `action` is `email`, `signup`, `signin`. Optional, but deprecated, if `prompt` is `none` (use `login_hint` instead in this case). Required if `action` is `force_auth`.
   - if `action` is `email`, the email address will be used to determine whether to display the signup or signin form, but the user is free to change it.
   - If `action` is `signup` or `signin`, the email address will be pre-filled into the account form, but the user is free to change it.
   - If `action` is `signin`, the literal string `blank` will force the user to enter an email address and the last signed in email address will be ignored.
@@ -312,6 +311,7 @@ content-server page.
     signed in email address will be used as the default.
   - If `action` is `force_auth`, the user is unable to modify the email
     address and is unable to sign up if the address is not registered.
+- `id_token_hint`: An OpenID ID Token for the user being logged in silently via `prompt=none`. If this parameter and `login_hint` are both included in a `prompt=none` request, this parameter will be used, and the `login_hint` will be ignored. See the [prompt=none doc][prompt-none] for more info.
 - `login_hint`: An alias to `email`
 - `prompt`: Specifies whether the Authorization Server prompts the End-User for reauthentication and consent.
   - `consent`: The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client.
--- a/packages/fxa-auth-server/lib/oauth/jwt_id_token.js
+++ b/packages/fxa-auth-server/lib/oauth/jwt_id_token.js
@@ -1,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const AppError = require('./error');
+const jwt = require('./jwt');
+const logger = require('./logging')('jwt_id_token');
+
+/**
+ * Verify an OIDC ID Token, following the flow in 3.1.3.7 of the spec:
+ * https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+ *
+ * This token verification flow is normally used by relying parties to validate
+ * tokens provided by the OpenID Provider. In this case, we (the OpenID
+ * Provider) use it to validate an ID Token passed in by a relying party, as an
+ * id_login_hint. Some steps in the flow don't apply to our use case, and are
+ * skipped. All steps in the flow are indicated in comments, to hopefully
+ * ease future maintenance.
+ *
+ * @param {String} idToken
+ * @param {String} clientId
+ * @throws `invalidToken` error if ID Token is invalid.
+ * @returns {Promise<Object>} resolves with claims in Token
+ */
+exports.verify = async function verify(idToken, clientId) {
+  // Step 1 is skipped, as we don't support JWE tokens.
+  //
+  // 1. If the ID Token is encrypted, decrypt it using the keys and algorithms
+  //    that the Client specified during Registration that the OP was to use to
+  //    encrypt the ID Token. If encryption was negotiated with the OP at
+  //    Registration time and the ID Token is not encrypted, the RP SHOULD
+  //    reject it.
+
+  // Step 2 is handled by the `jwt.verify()` function.
+  //
+  // 2. The Issuer Identifier for the OpenID Provider MUST exactly match the
+  //    value of the `iss` (issuer) Claim.
+  let claims;
+  try {
+    claims = await jwt.verify(idToken);
+  } catch (err) {
+    logger.debug('verify.error.jwtverify', err);
+    throw AppError.invalidToken();
+  }
+
+  // For Step 3, we just need to verify the audience claim matches the
+  // clientId. FxA controls the list of relying parties, so there's no chance
+  // an untrusted audience could be present.
+  //
+  // 3. The Client MUST validate that the aud (audience) Claim contains its
+  //    client_id value registered at the Issuer identified by the iss (issuer)
+  //    Claim as an audience. The aud (audience) Claim MAY contain an array
+  //    with more than one element. The ID Token MUST be rejected if the ID
+  //    Token does not list the Client as a valid audience, or if it contains
+  //    additional audiences not trusted by the Client.
+  if (typeof claims.aud === 'string' && claims.aud !== clientId) {
+    logger.debug('verify.error.aud', { aud: claims.aud, clientId });
+    throw AppError.invalidToken();
+  } else if (!claims.aud.includes(clientId)) {
+    logger.debug('verify.error.aud', { aud: claims.aud, clientId });
+    throw AppError.invalidToken();
+  }
+
+  // Steps 4 and 5 are skipped, because FxA doesn't support the Authorized
+  // Party (azp) Claim.
+  //
+  // 4. If the ID Token contains multiple audiences, the Client SHOULD verify
+  //    that an 'azp' field is present.
+  // 5. If an 'azp' Claim is present, the Client SHOULD verify that its
+  //    client_id is the Claim Value.
+
+  // Step 6 is skipped, because it doesn't apply to our use case.
+  //
+  // 6. If the ID Token is received via direct communication between the Client
+  //    and the Token Endpoint (which it is in this flow), the TLS server
+  //    validation MAY be used to validate the issuer in place of checking the
+  //    token signature. The Client MUST validate the signature of all other ID
+  //    Tokens according to JWS using the algorithm specified in the JWT 'alg'
+  //    Header Parameter. The Client MUST use the keys provided by the Issuer.
+
+  // Step 7 is handled by the `jwt.verify()` function call above.
+  //
+  // 7. The 'alg' value SHOULD be the default of RS256 or the algorithm sent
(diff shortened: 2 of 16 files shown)


packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+ * @throws `invalidToken` error if ID Token is invalid.
+ * @returns {Promise<Object>} resolves with claims in Token
+ */
+exports.verify = async function verify(idToken, clientId) {
+  // Steps 1 and 2 of the verification flow are already handled by jwt.js:
+  //
+  // 1. if the ID Token is encrypted, decrypt it using the keys and algorithms
+  //    negotiated at Registration time between OpenID Provider and Client.
+  //    If encryption was negotiated but the token is not encrypted, RP SHOULD
+  //    reject it.
```
rfk, line 25: We don't support encrypted JWTs, it may be worth adding a note to that effect in here so that future readers don't go looking for that code.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  //    'iss' Claim.
+  let claims;
+  try {
+    claims = await jwt.verify(idToken, { typ: HEADER_TYP });
+  } catch (err) {
+    throw AppError.invalidToken();
+  }
+
+  // TODO: what is the canonical list of trusted audiences? Do we ever issue
+  // ID Tokens with multiple audiences?
```
rfk, line 36: I don't think it's meaningful to speak of "trusted audiences" in this context, the set of possible audiences for an id_token is exactly the set of RPs. The check you have here looks fine.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  //    that an 'azp' field is present.
+  // 5. If an 'azp' Claim is present, the Client SHOULD verify that its
+  //    client_id is the Claim Value.
+
+  // Skipping this optional step:
+  //
+  // 6. Optionally, TLS server validation can be used if the token was received
+  // directly by the Client from the Token Endpoint.
+
+  // TODO: is this always RS256? AFAICT yes...
```
rfk, line 61: Yes, but you can read this from `./keys.js:SIGNING_ALG` if you don't want to hard-code it.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  //
+  // 7. The 'alg' value should be the default of 'RS256' or the value of
+  //    'id_token_signed_response_alg' set at Registration.
+  if (claims.alg !== 'RS256') {
+    throw AppError.invalidToken();
+  }
```
rfk, line 67: FWIW, `jwt.verify` will have performed this check already, but I understand why you want to do it here for completeness. You could perhaps make it more explicit by passing in an `algorithms` option to the call to `jwt.verify` above.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  // 8. If the JWT `alg` Header Parameter uses a MAC based algorithm such as
+  //    `HS256`, `HS384`, or `HS512`, the octets of the UTF-8 representation of
+  //    the `client_secret` corresponding to the `client_id` contained in the
+  //    `aud` (audience) Claim are used as the key to validate the signature.
+  //    For MAC based algorithms, the behavior is unspecified if the `aud` is
+  //    multi-valued or if an `azp` value is present that is different than
+  //    the `aud` value.
+
+  // 9. The current time MUST be before the time represented by the `exp` Claim.
+  //    TODO: should we allow for clock skew?
```
rfk, line 81: No need to allow for skew here, because the timestamp came from our own clock.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  //     the current time, limiting the amount of time that nonces need to be
+  //     stored to prevent attacks. The acceptable range is Client specific.
+
+  // TODO: AFAICT we don't use the `nonce` claim in our ID Tokens, skipping:
+  //
+  // 11. If a nonce value was sent in the Authentication Request, a `nonce`
+  //     Claim MUST be present and its value checked to verify that it is the
+  //     same value as the one that was sent in the Authentication Request.
+  //     The Client SHOULD check the nonce value for replay attacks. The
+  //     precise method for detecting replay attacks is Client specific.
```
rfk, line 98: Ref https://github.com/mozilla/fxa/issues/975 for support of the `nonce` parameter.

In addition to us not currently using this, I don't think we could meaningfully check `nonce` here anyway even if we did. We don't know or care what specific login flow the `id_token` came from, we're not interested in preventing replay attacks at this point, and I don't think we could learn an appropriate value for `nonce` through a different channel than how we received the JWT anyway.

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Verify an OIDC ID Token, following the flow in 3.1.3.7 of the spec:
+ * https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+ *
+ * @param {String} idToken
+ * @param {String} clientId
+ * @throws `invalidToken` error if ID Token is invalid.
+ * @returns {Promise<Object>} resolves with claims in Token
+ */
```
rfk, line 18: To help set expectations for future readers, I think it would be worth saying a few words at the top of this function to explain that it does all the checks that an RP would ideally perform locally; that is, that it's operating *from the perspective on an RP* which is a slightly unusual thing for code in this repo to be doing.

(I know that's what you're doing and it took me a little while for that to really "click" when reading the code here, so it's probably good context for other readers).

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -0,0 +1,115 @@
...
+  // 12. If the `acr` Claim was requested, Client SHOULD check the asserted
+  //     Claim Value is appropriate. The meaning and processing of `acr` Claim
+  //     Values is out of scope for the OIDC spec.
+
+  // TODO: AFAICT we don't use the `auth_time` or `max_age` at all, skipping:
+  //
+  // 13. If the `auth_time` Claim was requested, either specifically or by
+  //     using the `max_age` parameter, the Client SHOULD check the `auth_time`
+  //     Claim Value and request re-authentication if it determines too much
+  //     time has elapsed since the last End-User authentication.
```
rfk, line 112: Aye, a lot of these are specifically about the RP checking that the login flow it requested was actually the one that took place, and we don't need to care about those here.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -79,6 +81,36 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+        },
+        validate: {
+          headers: clientAuthValidators.headers,
+          payload: {
+            client_id: clientAuthValidators.clientId,
+            id_token: validators.idToken,
+          },
+        },
+        response: {
+          // TODO: sharpen this
```
rfk, line 101: 😆

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -79,6 +81,36 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+          schema: Joi.object().unknown(true),
+        },
+      },
+      handler: async function(request) {
+        const claims = await JWTIdToken.verify(
+          request.payload.id_token,
+          request.payload.client_id
+        );
+        // TODO: do we need to separately check that the clientId has 'openid' permissions?
```
rfk, line 110: My gut reaction is that this would not be a valuable check, FWIW.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -79,6 +81,36 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+    // bothering with the oauth-server indirection.
+    {
+      method: 'POST',
+      path: '/oauth/id-token-verify',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+        validate: {
+          headers: clientAuthValidators.headers,
```
rfk, line 94: This adds support for parsing `client_id` and `client_secret` out of the `Authorization` header, but I don't think you want that here. IIUC we don't intend for clients to call this endpoint directly anyway, and you're not doing anything that the caller couldn't do for itself. So, no need for secrets here.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -1644,6 +1644,20 @@ const Account = Backbone.Model.extend(
...
+     *
+     * @param {String} idToken - the ID Token
+     * @param {String} clientId - the client ID, used to verify the 'aud' claim
+     * @throws `invalidToken` error if Token is invalid.
+     * @returns {Promise<Object>} resolves with claims in the Token
+     */
+    verifyIdToken(idToken, clientId) {
+      const sessionToken = this.get('sessionToken');
+
+      return this._fxaClient.verifyIdToken(idToken, clientId, sessionToken);
```
rfk, line 1659: You got a `sessionToken` on the end of your args here that doesn't feel like it should be necessary (as noted above, this endpoint doesn't do anything secret so you shouldn't need any secrets in order to access it).

6a68:
Thanks for taking a look, @rfk! Glad to know I didn't miss any big chunks 🍻

packages/fxa-auth-server/lib/routes/oauth/index.js
6a68: Ah, I was thinking requiring a sessionToken might make it a little harder for an attacker to experiment with our validator, but happy to remove it :+1:

rfk: Hrm, interesting. I don't think it would add any additional security, but having this be authenticated by a sessionToken might help us just keep the whole thing on the rails a little more effectively - e.g. it could make it harder to accidentally accept an `id_token` when the user has an invalid sessionToken, or make it harder to accidentally mix up an `id_token` for user `A` with a sessionToken for user `B`. I would be happy to see that explored if it makes sense to you.

(You still wouldn't want the `clientAuthValidators` header here, but rather a sessionToken auth stragegy)

packages/fxa-content-server/app/scripts/lib/oauth-errors.js
```diff
@@ -177,6 +177,11 @@ var ERRORS = {
     message: t('Unverified user or session'),
     response_error_code: 'interaction_required',
   },
+  PROMPT_NONE_INVALID_ID_TOKEN_HINT: {
+    errno: 1013,
+    message: t('Invalid id_token_hint'),
+    response_error_code: 'invalid_request',
```
6a68, line 183: I'm not sure if this message should be translated. Also not sure if this should be an `interaction_required` error, rather than an `invalid_request`.

vladikoff: hm `invalid_request` seems fine 👁️

packages/fxa-auth-server/docs/oauth/api.md
```diff
@@ -303,15 +303,15 @@ content-server page.
...
   - if `action` is `email`, the email address will be used to determine whether to display the signup or signin form, but the user is free to change it.
   - If `action` is `signup` or `signin`, the email address will be pre-filled into the account form, but the user is free to change it.
   - If `action` is `signin`, the literal string `blank` will force the user to enter an email address and the last signed in email address will be ignored.
   - If `action` is `signin` and no email address is specified, the last
     signed in email address will be used as the default.
   - If `action` is `force_auth`, the user is unable to modify the email
     address and is unable to sign up if the address is not registered.
+- `id_token_hint`: An OpenID ID Token for the user being logged in silently via `prompt=none`. This is the preferred parameter to use with `prompt=none` for identifying the user, though passing the user's email directly, via the `login_hint` (or deprecated `email`) parameters, is also supported. See the [prompt=none doc][prompt-none] for more info.
```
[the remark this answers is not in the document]
6a68: ah yeah, that phrasing seems weird. I'll fix

packages/fxa-auth-server/test/local/routes/oauth.js
```diff
@@ -108,6 +110,33 @@ describe('/oauth/ routes', () => {
...
+
+      const mockRequest = mocks.mockRequest({
+        payload: {
+          client_id: MOCK_CLIENT_ID,
+          id_token: MOCK_JWT,
+        },
+      });
+
+      await loadAndCallRoute('/oauth/id-token-verify', mockRequest);
+      assert.calledOnce(mockVerify);
```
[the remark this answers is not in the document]
6a68: good catch, I'll round them off

PR #4831, opened 2020-04-09
chore(audit): only audit on push once per day

jodyheavener:
Totally fine if this idea gets rejected, but it was quick so I wanted to get everyone's take on it.

I am consistently running into issues with the NPM audit service, and am frequently having to push with `--no-verify`. This PR changes the pre-push hook to run a script that checks if you've audited within the last 24 hours. If you have, skip it, and audit again when 24 hours has passed.

To be clear, it doesn't solve the problem with auditing, buuuut it reduces the amount of times you could encounter the issue. Plus, with the number of people pushing to the repo there are absolutely going to be checks less than 24 hours apart.

Thoughts?

### DIFF
--- a/_scripts/audit.sh
+++ b/_scripts/audit.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -e
+
+FILE=.last-audit
+NOW=$(date +%s)
+
+function run_audit() {
+  npm run audit
+  echo $NOW > $FILE
+}
+
+if [ -f "$FILE" ]; then
+  THRESHOLD=86400 # 24 hours
+  EXISTING=$(< $FILE)
+  DIFFERENCE=$(($NOW - $EXISTING))
+
+  if [ "$DIFFERENCE" -ge "$THRESHOLD" ]; then
+    run_audit
+  else
+    echo "Audited within the last 24 hours, skipping"
+  fi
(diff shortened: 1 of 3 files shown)


chenba:
> I am consistently running into issues ...

Only occasionally for me, but I'm down.

dannycoates:
tbh i think once a day is enough

_scripts/audit.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
```
dannycoates, line 1: nit: remove `x` option

_scripts/audit.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ -f "$FILE" ]; then
+  THRESHOLD=86400 # 24 hours
+  EXISTING=$(< $FILE)
+  DIFFERENCE=$(($NOW - $EXISTING))
+
+  if [ "$DIFFERENCE" -ge "$THRESHOLD" ]; then
+    run_audit
+  else
+    echo "Skipping audit..."
```
dannycoates, line 19: nit: something about why it was skipped. maybe "Thetan levels nominal. Come back tomorrow for another audit" j/k

PR #4832, opened 2020-04-09
test(functional): add missing scenarios to the delete account and display name test

ashrivastava-qa:
I made the following changes:
- added scenario for "delete account with **incorrect** password" in the delete_account.js, selectors.js file
- made small changes to the "delete account with **correct password** and **click cancel** testcases" in delete_account.js file
- added scenario for 'changing display name' in the settings.js file

@vbudhram - Please review.

### DIFF
--- a/packages/fxa-content-server/tests/functional/delete_account.js
+++ b/packages/fxa-content-server/tests/functional/delete_account.js
@@ -20,7 +20,6 @@ const {
   createEmail,
   createUser,
   type,
-  fillOutDeleteAccount,
   fillOutEmailFirstSignIn,
   testElementTextInclude,
   openPage,
@@ -40,7 +39,7 @@ registerSuite('delete_account', {
   },
 
   tests: {
-    'sign in, delete account with incorrect password': function() {
+    'sign in, delete account': function() {
       return (
         this.remote
           .then(openPage(ENTER_EMAIL_URL, selectors.ENTER_EMAIL.HEADER))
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/tests/functional/delete_account.js
```diff
@@ -51,10 +54,41 @@ registerSuite('delete_account', {
               selectors.SETTINGS_DELETE_ACCOUNT.DETAILS
             )
           )
-          .then(fillOutDeleteAccount(PASSWORD))
+          .findAllByCssSelector(selectors.SETTINGS_DELETE_ACCOUNT.CHECKBOXES)
+          .then(checkboxes => checkboxes.map(checkbox => checkbox.click()))
+          .end()
+          .then(
+            type(selectors.SETTINGS_DELETE_ACCOUNT.INPUT_PASSWORD, 'PASSWORD')
```
vbudhram, line 61: I would probably call the `'PASSWORD'` something like `invalid password` or something similar.

packages/fxa-content-server/tests/functional/delete_account.js
```diff
@@ -51,10 +54,41 @@ registerSuite('delete_account', {
...
+          .then(click(selectors.SETTINGS_DELETE_ACCOUNT.SUBMIT))
+          .then(
+            visibleByQSA(
+              selectors.SETTINGS_DELETE_ACCOUNT.TOOLTIP_INCORRECT_PASSWORD
+            )
+          )
+      );
+    },
+
+    'sign in, delete account with correct password': function() {
```
vbudhram, line 72: I think you can combine the two `sign in, delete account`. In this test case, before you use the correct password, you could use the invalid password.

rbillings:
Suggest the title should include "delete account" so it's easy to know which test changed.

packages/fxa-content-server/tests/functional/delete_account.js
ashrivastava-qa: I made the suggested change.

packages/fxa-content-server/tests/functional/delete_account.js
ashrivastava-qa: I made the suggested change.

PR #4833, opened 2020-04-09
bug(auth-server): fix bad parameter being passed to stripeHelper.expandResource

jackiemunroe:
- fix bad parameter
- add check to expandResource to prevent misuse in the future.


dannycoates:
Can you PR this commit from a mozilla branch instead of your fork so that the stripe tests won't be skipped in CI?

PR #4835, opened 2020-04-09
bug(auth-server): fix bad parameter being passed to stripeHelper.expandResource

jackiemunroe:
- fix bad parameter
- add check to expandResource to prevent misuse in the future.


vbudhram:
@jackiemunroe Looks like the test still fails `support form with an active subscription - go to support form, submits the form`. Did it pass for you locally?

jackiemunroe:
@vbudhram i just pushed a hopeful fix for that. It looks like there is a test that wasn't updated as a part of other work I did.

PR #4848, opened 2020-04-10
chore(styles): convert payments UI to grid and tidy

johngruen:


packages/fxa-payments-server/src/index.scss
```diff
@@ -41,13 +41,24 @@ h3.billing-title {
 }
 
 .main-content {
-  display: flex;
-  flex-direction: column-reverse;
-  padding-top: $header-height;
+  display: grid;
```
LZoog, line 44: Did you also want to change other flex items like `plan-details`? (I noticed with FPN, we probably need a gutter between the title and price)

I'm kind of indifferent on having grid/flex mixed right now because we haven't established a preferred pattern (but we will soon!) so I think flex + grid here is OK in the meantime, but _maybe_ it should wait? We'll either need to refactor this back to flexbox or finish refactoring the rest of payments to grid. What visual differences did this make for us, what issues does this PR fix with the layout?

image

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.scss
```diff
@@ -7,18 +7,18 @@
...
   @include min-width('tablet') {
-    min-height: calc(100vh - #{$header-height-tablet});
+    border-radius: 0;
+    border-top: 0;
+    min-height: 100%;
   }
 
   @include min-width('desktop') {
-    margin: -1px 24px 0;
+    margin: -24px 0;
```
LZoog, line 21: We should avoid using negative margin like this. Instead, the `padding-top` value should be adjusted here.

image

packages/fxa-payments-server/src/components/PaymentConfirmation/index.scss
```diff
@@ -3,18 +3,20 @@
...
+  min-height: 100%;
   padding: 20px 16px 0;
+  width: 100%;
+
+  @include min-width('tablet') {
+    border-top: 0;
+  }
+
+  @include min-width('desktop') {
+    margin: -24px 0;
```
LZoog, line 18: Same comment here as below, let's adjust the padding instead of moving the entire section "up" with negative margin.

packages/fxa-payments-server/src/index.scss
johngruen: so this PR does a few things but the biggest are:
1. It makes the payment success page and payment page behave the same at the same browser widths. previously the layout of the success page only showed up on mobile
2. changes the way the page looks on mobile views so that the details card is sized correctly
3. removes unnesc. border radii

LZoog:
Just pointing this out/might be intentional, it may not be ideal to have a forced width on the left column in tablet. Here's a screen just before it switches to desktop. Wouldn't we want it to fill the extra space on the left instead of being fixed?

image

Also we may want to file an issue for this lack of space that I'd pointed out (or fix it here):
image

PR #4864, opened 2020-04-10
refactor(npm-scripts): refactored npm scripts for better ergonomics

dannycoates:
1. renamed _scripts/*.sh to use kebab-case
2. linted _scripts with shellcheck
3. added _dev directory and moved pm2 json files there
4. added nps and package-scripts.js
  - this allows us to do nice things like: `npm start debug firefox`
  - I'd like to expand this further but it's a start
5. refactored the install scripts to make `npm install` nicer


_dev/pm2/services-debug.json
```diff
@@ -165,19 +112,20 @@
       "min_uptime": "2m"
     },
     {
-      "name": "payments server PORT 3031",
+      "name": "payments-debug",
```
bbangert, line 115: Having inconsistent names when running in debug mode vs. not means that the VSCode scripts are going to have difficulty, for example the auth-server VSCode task runs a command like so:
``` 
     "command": "pm2 stop 'auth-server key server PORT 9000'"
```

dannycoates: pm2 gets confused when you have two different `apps` with the same name. maybe we just ditch the non-debug version

.vscode/launch.json
```diff
@@ -5,12 +5,36 @@
...
+      "cwd": "${workspaceFolder}",
+      "port": 9180
+    },
+    {
+      "name": "Attach to payments",
+      "type": "node",
+      "request": "attach",
+      "skipFiles": ["<node_internals>/**"],
+      "cwd": "${workspaceFolder}",
+      "port": 9170
```
bbangert, line 37: Nice!

PR #4866, opened 2020-04-10
feat(db): modified procedures to set verifiedAt field when email gets verified

### LINKED ISSUE #422: Drop `emails`.`verifiedAt` column?
This was noticed while doing some maintenance on the database. This column is not used anywhere and I vote we drop it.

farhan787:
fixes #422

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
+++ b/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
@@ -51,6 +51,7 @@ function createEmail(data) {
     uid: data.uid,
     emailCode: data.emailCode || crypto.randomBytes(16),
     isVerified: data.isVerified || false,
+    verifiedAt: data.verifiedAt || null,
     isPrimary: false,
     createdAt: Date.now(),
   };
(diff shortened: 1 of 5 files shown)


farhan787:
@vbudhram could you please review it ?

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -862,7 +862,7 @@ module.exports = function(log, error) {
   // Update : accounts, emails
   // Set    : emailVerified = true if email is in accounts table or isVerified = true if on email table
   // Where  : uid = $1, emailCode = $2
-  var VERIFY_EMAIL = 'CALL verifyEmail_8(?, ?)';
+  var VERIFY_EMAIL = 'CALL verifyEmail_9(?, ?)';
```
vbudhram, line 865: We will need to add a few asserts for this, the spots where we return the email are in the `accountEmails`  and `getSecondaryEmail` procedure. I would find relevant test cases that use these procedures and assert that they return the expected `verifiedAt` values.

packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js
```diff
@@ -3539,6 +3575,7 @@ module.exports = function(config, DB) {
               'correct emailCode'
             );
             assert.isTrue(!!res[1].isVerified);
+            assert.isAbove(res[1].verifiedAt, res[1].createdAt);
```
farhan787, line 3578: @vbudhram how're we getting two emails [here](https://github.com/mozilla/fxa/blob/43cb00c79dfac383fdb3958e5b8b05576042c0c7/packages/fxa-auth-db-mysql/db-server/test/backend/db_tests.js#L3528) in the response when we're just creating only 1 email, the first email is having the `isVerified` set as `true` but the `verifiedAt` field is still `null`. I tried to find out why we've 2 emails in the response and only that can answer this that why `verifiedAt` is still null, may be we've to modify some other db procedure also that is setting the `isVerified` true.

farhan787:
@vbudhram I think it's ready for your final review.

PR #4896, opened 2020-04-13
Refactor pm2

dannycoates:
The basic idea is that pm2 should be the standard way to run local dev, even for debugging, that the pm2 config for each service should be located within it's own package directory, and that standard npm commands can start, restart, stop it.

Eventually I'd like to be able to control any service from the base directory with npm commands, but for now there's a base level `npm start` and `stop` as before along with `npm restart services` to restart the services but leave the databases running. Individual services can be controlled with pm2 directly or with npm from their package directory. I shortened the service names in pm2 to make them easier to control: `pm2 logs auth` for example instead of using ids.

Some of the services start with debugging enabled and have watch enabled to restart on code changes. I removed other methods of source watching in the process.

Services that need to wait on other things before starting now handle that in the `prestart` npm script.

For vscode I updated the `.vscode/launch.json` to add attach configs for each service that runs with debugging. I would like to make an extension for better pm2/vscode integration but I think this basic support works ok (?)


dannycoates:
Here's what you get with an `npm start`:
Screen Shot 2020-04-13 at 9 25 40 AM

Incidentally the process tree is much flatter now. One benefit is that the `mem` column is actually accurate, and also `pm2 pid`,`sendSignal`, etc. work as expected, which can be useful for writing scripts.

Screen Shot 2020-04-13 at 9 27 26 AM

package.json
```diff
@@ -27,6 +25,7 @@
   "author": "Mozilla (https://mozilla.org/)",
   "license": "MPL-2.0",
   "dependencies": {
+    "@dannycoates/nps": "^5.9.12",
```
vbudhram, line 28: Were you going to try and submit a PR for your changes to nps?

dannycoates: https://github.com/sezna/nps/pull/220

dannycoates:
>     * I'm different to pulling pm2 into each directory

I agree. lerna bootstrap is supposed to hoist it up so that it's not actually in every package's node_modules taking up space. The other option is to require pm2 to be installed globally, but I hate requiring that, even though I have it that way on mine.

dannycoates:
I'd also like to move config from `dev.json` files into the `env` of pm2.config.js to more closely match how prod sets config

dannycoates:
This one supersedes #4864

lmorchard:
Do we have anywhere that the port numbers are listed? Doing a `./pm2 ps` to look them up is probably a daily thing for me

chenba:
👍

dannycoates:
> Do we have anywhere else that the port numbers are listed?

@lmorchard how about `npm run ports`?

output:

```sh
> fxa@2.0.0 ports /Users/dcoates/code/fxa
> pm2 jlist | json -d'	' -a -c 'this.pm2_env.env.PORT' pm2_env.env.PORT name

6379	redis
8080	123done
10139	321done
5050	browserid
9292	fortress
8091	admin
8092	admin-react
8090	admin-server
8000	auth-db
9000	auth
3030	content
3031	payments
3032	payments-react
1111	profile
7100	support
```

packages/fxa-admin-server/package.json
```diff
@@ -61,11 +61,12 @@
     "mocha": "^7.1.0",
     "nodemon": "^2.0.2",
     "npm-run-all": "^4.1.5",
+    "pm2": "^4.2.3",
     "prettier": "^1.19.1",
     "proxyquire": "^2.1.3",
     "sinon": "^9.0.1",
     "supertest": "^4.0.2",
-    "ts-node-dev": "^1.0.0-pre.44",
+    "ts-node": "^8.8.2",
```
bbangert, line 69: A quick note on why I was using `ts-node-dev` here, from the webpage:

> Tweaked version of node-dev that uses ts-node under the hood.

> It restarts target node process when any of required files changes (as standard node-dev) but shares Typescript compilation process between restarts. This significantly increases speed of restarting comparing to node-dev -r ts-node/register ..., nodemon -x ts-node ... variations because there is no need to instantiate ts-node compilation each time.

This helps **quite a bit** when working on the code for quick restarts.

package.json
sezna: This has been merged now.

Issue #4817, opened 2020-04-09
[setup] Share React components, pull in `LinkExternal`


data-sync-user:
We want to be able to share React components across our various front-ends. `fxa-shared/react` could be the place to do it, but an `fxa-react-shared` specifically for React components may be cleaner.

`create-react-app` seems to have trouble importing outside of the `src/` directory. Can we use a symlink? This relates to an [ADR around CRA ](https://github.com/mozilla/fxa/pull/4897)where `rescripts` is linked. There is a proof of concept on how we can share components.

Let's start by placing `LinkExternal` from the `fxa-admin-panel` into the shared package and ensure we can import it from `fxa-admin-panel`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1538)
┆Issue Number: FXA-1538

data-sync-user:
➤ Jody Heavener commented:

[~lzugai@mozilla.com] because each source for external components needs to be explicitly added, and in order to have a single source of truth, how would you feel about me setting up Rescripts in fxa-admin-panel (very straightfoward now), moving LinkExternal to fxa-components, and import them into both projects?

data-sync-user:
➤ Jody Heavener commented:

too late I'm doing it

PR #4897, opened 2020-04-13
docs(adrs): proposal for "React Toolchain for Settings Redesign"

### LINKED ISSUE #4811: [ADRs] New React App + Settings Redesign: CRA vs "ground up" ADR
Can we use CRA for the settings redesign or will we end up needing to eject anyway to use `fxa-shared` or to configure webpack? It would be ideal to share components between settings, payments, and even admin-panel. Importing components outside of the `src/` directory proves difficult and we can't have a global SCSS var file (which may not matter anyway if we use CSS variables).

Right now, we don't have HMR in payments and rely on Storybook for these changes. Are we taking the ideal approach?

The ADR to close this issue should look at pros and cons of CRA vs manual setup to document why or why we wouldn't want to use CRA. Also, see https://github.com/harrysolovay/rescripts. Is the solution CRA + rescripts?

This issue helps inform a couple under the "initial setup" story.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1529)
┆Issue Number: FXA-1529

jodyheavener:
Closes #4811

📖 Preview [here](https://github.com/mozilla/fxa/blob/jh/4811-adr-settings-react-toolchain/docs/adr/0013-react-toolchain-for-settings-redesign.md) and review [here](https://github.com/mozilla/fxa/pull/4897/files).

### DIFF
--- a/docs/adr/0013-react-toolchain-for-settings-redesign.md
+++ b/docs/adr/0013-react-toolchain-for-settings-redesign.md
@@ -1,0 +1,118 @@
+# React Toolchain for Settings Redesign
+
+- Status: accepted
+- Deciders: Jody Heavener, Lauren Zugai, Les Orchard
+- Date: 2020-04-13
+
+## Context and Problem Statement
+
+Firefox Accounts is introducing a new visual design for Settings (the interface that exists at the address `https://accounts.firefox.com/settings`). The changes involved are significant enough such that it [has been decided](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md) that an entirely new React App will be built more or less independent of the existing Settings web app. With this we can beta test and roll out the new design with minimal disruption.
+
+As there are a multitude of ways a React App can be set up and configured, this decision has introduced the question, "What [toolchain](https://en.wikipedia.org/wiki/Toolchain) should we set up a new React App with?". This ADR serves to answer that question by going over the various approaches we can take to set up a new React App and how it might integrate with our existing FxA ecosystem.
+
+It's important to note that we are not deciding the languages, libraries, or other tools that we will use within the React development environment, but rather, the system that will be used to develop the React App with.
+
+Finally, we are not making decisions about the App's production server here. We'll be using [fxa-content-server](https://github.com/mozilla/fxa/tree/master/packages/fxa-content-server/server)'s Express server for the production build.
+
+## Decision Drivers
+
+- **Familiarity** - unless there is an overwhelming reason, we don't really want to introduce a completely foreign build process for everyone to learn.
+- **Configurability** - if it is less configurable does it meet our needs; if it is more configurable is the time spent worth it?
+- **Maintainability** - how much work do we set up for ourselves if we're maintaining all upgrades and configuration?
+
+Additionally, our team has a desire to share React components and other files across projects. Both the Payments and Admin Panel web apps use React, and we would benefit from being able to develop components that could be imported into any of these React Apps.
+
+## Considered Options
+
+- Option A - Completely Custom Approach
+- Option B - Create React App
+  - Option B2 - Create React App and immediately eject
+  - Option B3 - Create React App and use Rescripts
+- Option C - Neutrino with React Preset
+- Option D - Other Considerations
+
+## Decision Outcome
+
+Chosen option: "B3 - Create React App and use Rescripts", because:
+
+- Our team has sufficient knowledge in working with CRA over the other options.
+- We don't need to re-invent the wheel. Our needs for the Settings project are not so great that it would be worth spending time creating a ground-up toolchain. To some extent this option allows us to do what we do best without needing to worry about the specifics of the development workflow.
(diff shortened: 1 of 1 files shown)


docs/adr/0013-react-toolchain-for-settings-redesign.md
```diff
@@ -0,0 +1,118 @@
...
+
+As there are a multitude of ways a React App can be set up and configured, this decision has introduced the question, "What [toolchain](https://en.wikipedia.org/wiki/Toolchain) should we set up a new React App with?". This ADR serves to answer that question by going over the various approaches we can take to set up a new React App and how it might integrate with our existing FxA ecosystem.
+
+It's important to note that we are not deciding the languages, libraries, or other tools that we will use within the React development environment, but rather, the system that will be used to develop the React App with.
+
+Finally, we are not making decisions about the App's production server here. We'll be using [fxa-content-server](https://github.com/mozilla/fxa/tree/master/packages/fxa-content-server/server)'s Hapi server for the production build.
```
chenba, line 15: content uses express not Hapi.

jodyheavener: whoops good find

docs/adr/0013-react-toolchain-for-settings-redesign.md
```diff
@@ -0,0 +1,118 @@
...
+- Option C - Neutrino with React Preset
+- Option D - Other Considerations
+
+## Decision Outcome
+
+Chosen option: "B3 - Create React App and use Rescripts", because:
+
+- Our team has sufficient knowledge in working with CRA over the other options.
+- We don't need to re-invent the wheel. Our needs for the Settings project are not so great that it would be worth spending time creating a ground-up toolchain. To some extent this option allows us to do what we do best without needing to worry about the specifics of the development workflow.
+- While not completely configurable compared to other options, the addition of Rescripts can provide us with greater flexibility where we might need it (and can possibly help us with the external imports restriction CRA has).
```
LZoog, line 40: I think it would be ideal to determine if rescripts will actually allow us to import outside of `src/`. We have an issue for "share components" in #4817, but that really plays into the decision here. A quick search lead me to https://github.com/oklas/react-app-rewire-alias, I haven't poked at that much but... Maybe the solution is (especially if we're going to use CSS vars over SCSS vars) CRA + that `rewire-alias` tool, which seems to allow for multiple `src` dirs? Or maybe rescripts allows this functionality and has better support than `rewire-alias`?

chenba: rescripts has a function specifically for removing webpack plugins: https://github.com/harrysolovay/rescripts#removewebpackpluginconstructorname-config  Looks simple enough.

docs/adr/0013-react-toolchain-for-settings-redesign.md
```diff
@@ -0,0 +1,118 @@
+# React Toolchain for Settings Redesign
+
+- Status: proposed
+- Deciders: Lauren Zugai, Jody Heavener
```
LZoog, line 4: I'd say put your name first since you wrote this guy, and probably add @lmorchard since I originally was like meh maybe we don't need CRA! And Les brought me back to like actually we do want it for maintainability and such.

docs/adr/0013-react-toolchain-for-settings-redesign.md
LZoog: That LGTM. @jodyheavener I also came across https://github.com/gsoft-inc/craco, looks similar to rescripts with possibly better support?

jodyheavener: @LZoog @chenba I've built out a solution that addresses and demonstrates this; will update the ADR to reflect.

docs/adr/0013-react-toolchain-for-settings-redesign.md
```diff
@@ -0,0 +1,118 @@
...
+- Option C - Neutrino with React Preset
+- Option D - Other Considerations
+
+## Decision Outcome
+
+Chosen option: "B3 - Create React App and use Rescripts", because:
+
+- Our team has sufficient knowledge in working with CRA over the other options.
+- We don't need to re-invent the wheel. Our needs for the Settings project are not so great that it would be worth spending time creating a ground-up toolchain. To some extent this option allows us to do what we do best without needing to worry about the specifics of the development workflow.
+- Rescripts allows us to layer additional configuration overtop of the existing configuration without jeopardizing the integrity or trading away any of the powers of CRA. Specifically, we can use Rescripts to modify CRA to allow external React components to be imported, as demonstrated in [this repo](https://github.com/jodyheavener/rescripts-react-app) (technical notes [here](https://github.com/jodyheavener/rescripts-react-app/blob/master/.rescriptsrc.js)). This capability could be implemented in our other React Apps.
```
LZoog, line 40: 🙌 Thanks for making that proof of concept. While we may end up modifying it, it's great to see something in action working to set us up to share React components in our CRAs across our monorepo with this decision.

Issue #4889, opened 2020-04-13
Contacting support returns an error


SoftVision-LorandJanos:
#### Affected versions:
- All Firefox

#### Affected Platforms:
- All platforms

#### Preconditions
- Have an active subscription.

#### Steps to reproduce
1. Go to the "Manage subscription" page.
2. Click on the "Contact support" button and complete the fields with valid data.
3. Click on the Send button.

#### Expected result:
- A message is displayed, telling the user that the ticket was sent to the support page.

#### Actual result:
- The "Hmm, we're having trouble with our system. We're working on fixing it for you and apologize for the inconvenience. Please try again later." error message is displayed.

![image](https://user-images.githubusercontent.com/48829350/79108806-c6ed9900-7d7f-11ea-886d-d29908f8953d.png)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1596)
┆Issue Number: FXA-1596

chenba:
The plan name was removed in #4693 but the support ticket still requires a plan (https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/support.js#L48).

data-sync-user:
➤ Wil Clouser commented:

https://github.com/mozilla/fxa/pull/4904 ( https://github.com/mozilla/fxa/pull/4904 ) should fix this

SoftVision-LorandJanos:
This issue is no longer reproducible on [stage website](https://stage.guardian.nonprod.cloudops.mozgcp.net/vpn) site, using the latest Firefox Release. Tested on iOS 13.3.1 (iPhone 6S).

PR #4904, opened 2020-04-13
Fxa 1419 fluent updates

### LINKED ISSUE #4668: Payment flow - Localization Updates (result of l10n/Fluent review)
After discussions with the l10n and Fluent teams, we need to make the following updates for our payment flow translations:

1. Update `fluent` packages in payments
2. Update our currency formatting to use Fluent, which will take into account the currency and the locale for display
3. Update our date display to use Fluent

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1419)
┆Issue Number: FXA-1419

jackiemunroe:
- updated fluent packages to the latest
- implemented localizable currency
- implemented localizable dates

fixes #4668

- fixed errors that resulted from rebase onto Mobile Updates
- fixed bug that would cause the subscriptionToResponse to provide an invoice ID instead of an invoice Number
- removed reference to plan name and replaced with product name


packages/fxa-payments-server/src/components/PaymentConfirmation/index.test.tsx
```diff
@@ -65,4 +76,224 @@ describe('PaymentConfirmation', () => {
...
+        );
+      }
+
+      describe('When plan has day interval', () => {
+        const expectedMsgId = dayBasedId;
+
+        it('Handles an interval count of 1', () => {
+          const plan_id = 'plan_daily';
+          const plan = findMockPlan(plan_id);
+          const expectedMsg = `$5.00 daily`;
```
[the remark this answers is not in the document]
jackiemunroe: Yeah, I was playing around with passing in a var there 😅  Will update.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.test.tsx
```diff
@@ -65,4 +76,224 @@ describe('PaymentConfirmation', () => {
...
+        );
+        const testInstance = testRenderer.root;
+
+        const planPriceComponent = testInstance.findByProps({
+          id: expectedMsgId,
+        });
+        const expectedAmount = getLocalizedCurrency(plan.amount, plan.currency);
+
+        expect(planPriceComponent.props.$amount).toStrictEqual(expectedAmount);
+        expect(planPriceComponent.props.$amount).toStrictEqual(expectedAmount);
```
chenba, line 117: I think I'm seeing doubles here.

jackiemunroe: Yep. I will fix that.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -16,14 +19,31 @@ type PaymentConfirmationProps = {
...
+    case 'week':
+      if (intervalCount === 1) return `${amount} weekly`;
+      return `${amount} every ${intervalCount} weeks`;
+    case 'month':
+      if (intervalCount === 1) return `${amount} monthly`;
+      return `${amount} every ${intervalCount} months`;
+    case 'year':
+      if (intervalCount === 1) return `${amount} yearly`;
+      return `${amount} every ${intervalCount} years`;
+  }
```
chenba, line 36: Would be nice to restrict `interval` to the type `'day'|'week'|'month'|'year'`.  Otherwise the caller can pass in any other string and get an `undefined`.  Declaring a return type of `string` would also help.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -18,18 +21,47 @@ type PlanDetailsProps = {
...
+    case 'week':
+      if (intervalCount === 1) return `${amount} weekly`;
+      return `${amount} every ${intervalCount} weeks`;
+    case 'month':
+      if (intervalCount === 1) return `${amount} monthly`;
+      return `${amount} every ${intervalCount} months`;
+    case 'year':
+      if (intervalCount === 1) return `${amount} yearly`;
+      return `${amount} every ${intervalCount} years`;
+  }
```
chenba, line 38: This look identical to the function earlier.  Can we define it once somewhere?

packages/fxa-payments-server/src/lib/formats.ts
```diff
@@ -1,11 +1,108 @@
...
+ * @param currency
+ */
+export function getLocalizedCurrencyString(
+  amountInCents: number,
+  currency: string
+): string {
+  const decimal = amountInCents / 100;
+  const options = { ...baseCurrencyOptions, currency };
+
+  return new Intl.NumberFormat('en-US', options).format(decimal);
```
chenba, line 41: Is `en-US` suppose to be hard coded here?

packages/fxa-payments-server/src/lib/formats.ts
```diff
@@ -1,11 +1,108 @@
...
+export function getLocalizedDateString(
+  unixSeconds: number,
+  numericDate: boolean = false
+): string {
+  const milliseconds = unixSeconds * 1000;
+  var date = new Date(milliseconds);
+
+  const options = numericDate ? numericDateOptions : defaultDateOptions;
+
+  return new Intl.DateTimeFormat('en-US', options).format(date);
```
chenba, line 107: Same question on the `en-US` here.

packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.test.tsx
```diff
@@ -158,8 +158,10 @@ describe('routes/Product/SubscriptionUpgrade', () => {
         );
         const testInstance = testRenderer.root;
         const legalCheckbox = testInstance.findByProps({ id: expectedMsgId });
+        const expectedAmount = getLocalizedCurrency(plan.amount, plan.currency);
 
-        expect(legalCheckbox.props.$amount).toBe(expectedAmount);
+        expect(legalCheckbox.props.$amount).toStrictEqual(expectedAmount);
+        expect(legalCheckbox.props.$amount).toStrictEqual(expectedAmount);
```
chenba, line 164: I think I'm seeing doubles.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -519,12 +520,12 @@ export function getLocalizedMessage(
   msgId: string,
   args: any
 ): string {
-  const msg = bundle.getMessage(msgId);
-  if (msg === undefined) {
+  let localizedMessage = bundle.getMessage(msgId);
+  if (localizedMessage == null || localizedMessage.value === null) {
```
chenba, line 524: Was the `==` here intentional?

packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.test.tsx
```diff
@@ -372,21 +375,18 @@ describe('PlanDetail', () => {
...
       const testRenderer = TestRenderer.create(<PlanDetail {...props} />);
       const testInstance = testRenderer.root;
       const planPriceComponent = testInstance.findByProps({
         id: expectedMsgId,
       });
+      const expectedAmount = getLocalizedCurrency(plan.amount, plan.currency);
 
-      expect(planPriceComponent.props.$amount).toBe(expectedAmount);
+      expect(planPriceComponent.props.$amount).toStrictEqual(expectedAmount);
+      expect(planPriceComponent.props.$amount).toStrictEqual(expectedAmount);
```
chenba, line 389: Doubles

chenba:
I find "annually" more natural than "yearly" even though yearly is correct. 🤷‍♂️

packages/fxa-payments-server/src/lib/formats.ts
jackiemunroe: The logic here: this is the default string that would only be displayed if Fluent fails to translate, currently all of our default strings are en-US. If we use the user's local here (which we totally can), we would run into an issue where the fallback is a mixed language/format. The user would be seeing all of our other default strings (en-US) with a currency that is formatted according to a different local's standard, which could cause confusion.

jackiemunroe:
> I find "annually" more natural than "yearly" even though yearly is correct. 🤷‍♂️
This was according to requirements, but I do agree "annually" sounds better.
@johngruen what are your thoughts?

packages/fxa-payments-server/src/lib/test-utils.tsx
jackiemunroe: Yes, because `getMessage` can return `undefined`. Do you think that this should be updated to `localizedMessage === undefined || localizedMessage === null || localizedMessage.value === null` ?

chenba: Ah.  Upon checking, I see that `getMessage` doesn't return a null, so we could skip the null check.

packages/fxa-auth-server/lib/routes/support.js
```diff
@@ -68,7 +67,7 @@ module.exports = (log, db, config, customs, zendeskClient) => {
         const { uid, email } = await handleAuth(request.auth, true);
         const { location } = request.app.geo;
         await customs.check(request, email, 'supportRequest');
-        let subject = `${request.payload.plan}`;
+        let subject = `${request.payload.productName}`;
```
chenba, line 70: Why is this code in this PR?  Doesn't seem to be fluent related.

packages/fxa-payments-server/src/lib/test-utils.tsx
jackiemunroe: updated

packages/fxa-auth-server/lib/routes/support.js
jackiemunroe: This was updated because `plan` had been removed from the payload, and this had not been corrected.

chenba: Sure, but this PR is for Fluent changes, and that code fixes https://github.com/mozilla/fxa/issues/4889.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
jackiemunroe: updated

packages/fxa-auth-server/lib/routes/support.js
jackiemunroe: Yes, as stated in #4889, this PR fixes that issue.

jackiemunroe: It was fixed here, but can be pulled out. I was afraid that the tests would fail upon trying to open a PR if that had not been fixed. There were errors that resulted from the plan -> product name usage that were not caught because the tests do not run if the PR includes a branch from a forked repo. I had thought that the tests might fail in this PR, since I decided to move my work to `mozilla/fxa`, so I went ahead and fixed it.

chenba: Huh.  I guess then I don't understand why tests would fail now without this change when they didn't fail in PR #4694.  

A separate smaller PR is nicer because then the changelog(s) would make more sense (although I'm not even sure anyone actually read them), and the unit of change can be cherry-picked or reverted easily.

(I already r+ so feel free to merge when you thing the PR is ready. 😀)

vbudhram:
These changes were reviewed and pull directly to the uplift branch. Please ref comment https://github.com/mozilla/fxa/pull/4935#issue-403416398. Closing here.

PR #4905, opened 2020-04-13
fix(email): adjust punctuation by request of L10n team

clouserw:


clouserw:
Yeah, we should.  Tagging @rbillings @vasilicatamas @bsurd

PR #4907, opened 2020-04-13
feat(admin panel): send mutation event to graphql server in order to unblock email

### LINKED ISSUE #4209: [admin panel A1.1] Send mutation request to graphQL to unblock an email, handle response
Send a mutation request to the graphQL endpoint with the selected email address marked to unblock and handle the response. We will want a confirmation prompt.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1146)
┆Issue Number: FXA-1146

jodyheavener:
Closes #4209 

- Send request to GQL server to clear an account's email bounces / updated UI to reflect this action
- Updates/documents email bounce generation script

### DIFF
--- a/packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
+++ b/packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
@@ -5,110 +5,151 @@
 import React from 'react';
 import Chance from 'chance';
 import { render, fireEvent, act } from '@testing-library/react';
-import { MockedProvider, MockedResponse } from '@apollo/react-testing';
+import { MockedProvider, MockedResponse, wait } from '@apollo/react-testing';
 import '@testing-library/jest-dom/extend-expect';
+import { CLEAR_BOUNCES_BY_EMAIL } from './Account/index';
 import { GET_ACCOUNT_BY_EMAIL, EmailBlocks } from './index';
 
 const chance = new Chance();
+let testEmail: string;
+let deleteBouncesMutationCalled = false;
 
-const exampleEmail = chance.email();
-
-function exampleEmailBounce() {
+function exampleBounce(email: string) {
   return {
-    email: exampleEmail,
+    email,
     createdAt: chance.timestamp(),
     bounceType: 'Permanent',
     bounceSubType: 'General',
   };
 }
 
-const exampleAccount: MockedResponse = {
-  request: {
-    query: GET_ACCOUNT_BY_EMAIL,
-    variables: {
-      email: exampleEmail,
-    },
-  },
-  result: {
-    data: {
-      accountByEmail: {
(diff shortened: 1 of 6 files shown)


packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
```diff
@@ -5,110 +5,151 @@
...
 
 it('renders without imploding', () => {
   const renderResult = render(<EmailBlocks />);
   const getByTestId = renderResult.getByTestId;
 
-  expect(getByTestId('form')).toBeInTheDocument();
+  expect(getByTestId('search-form')).toBeInTheDocument();
 });
 
-test.skip('displays the loading state on submit', async () => {
```
jodyheavener, line 51: I'm removing this test because I don't know that we can reliably test this state.

The docs indicate that `MockedProvider` uses a Promise to return the results, so as long as you don't resolve the Promise the loading state can be visible. But we can't get to the loading state without first clicking the search button, which requires the use of `await act(async cb)` because it sets state under the hood, and doing this means resolving the Promise. However, in this MockedProvider setup we provide no mocks, so this will yield an error anyway.

packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
```diff
@@ -5,110 +5,151 @@
...
+let testEmail: string;
+let deleteBouncesMutationCalled = false;
 
-const exampleEmail = chance.email();
-
-function exampleEmailBounce() {
+function exampleBounce(email: string) {
   return {
-    email: exampleEmail,
+    email: email,
```
LZoog, line 19: Can just be `email,`

packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
```diff
@@ -5,110 +5,151 @@
...
-};
+  };
+}
+
+function exampleBounceMutationResponse(email: string): MockedResponse {
+  return {
+    request: {
+      query: CLEAR_BOUNCES_BY_EMAIL,
+      variables: {
+        email: email,
```
LZoog, line 52: `email,`

packages/fxa-admin-panel/src/components/EmailBlocks/index.test.tsx
```diff
@@ -5,110 +5,151 @@
...
+      return {
+        data: {
+          clearEmailBounce: true,
+        },
+      };
+    },
+  };
+}
+
+beforeAll(function() {
```
LZoog, line 66: This doesn't matter of course but for consistency we could use an arrow function here and line 70🤷‍♀️

PR #4938, opened 2020-04-14
refactor(config): replace 127.0.0.1 with localhost

dannycoates:
why, after all these years? 

because there's a few annoying cases where the loopback
makes dev harder. When you try to test a flow from a not
local machine, like a tv, mobile device, or a saucelabs
proxy session. With localhost it's easy enough to forward
but loopback not so much. And now that we're in a monorepo, 
the change shouldn't be a nightmare.

enough is enough

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ The Firefox Accounts (fxa) monorepo
 
 Note this starts up all required services, including Redis, MySQL, and Memcached. It is recommended that you don't run these services yourself, or occupy any of the [server ports](https://github.com/mozilla/fxa/blob/master/mysql_servers.json). Doing so may result in errors.
 
-4. Visit [127.0.0.1:3030](http://127.0.0.1:3030/).
+4. Visit [localhost:3030](http://localhost:3030/).
 
 Use the [PM2 tool](https://github.com/Unitech/PM2#main-features) to stop and start the servers, and read server logs.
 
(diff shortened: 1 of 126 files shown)


dannycoates:
next step will be top level config to set it to whatever you want, but this first.

jodyheavener:
bless you

packages/fxa-auth-server/CHANGELOG.md
```diff
@@ -1797,7 +1797,7 @@ Merge v1.126.3 into train-127
 - **config:** Add environment config options ([14a9b4a](https://github.com/mozilla/fxa-auth-server/commit/14a9b4a))
 - **config:** expose clients config as OAUTH_CLIENTS ([04ebf6f](https://github.com/mozilla/fxa-auth-server/commit/04ebf6f))
 - **config:** expose more environment variables for config ([7a1dd19](https://github.com/mozilla/fxa-auth-server/commit/7a1dd19))
-- **config:** For dev, the openid issuer is http://127.0.0.1:3030 (#583) r=@vladikoff ([38e1d73](https://github.com/mozilla/fxa-auth-server/commit/38e1d73)), closes [#583](https://github.com/mozilla/fxa-auth-server/issues/583) [mozilla/fxa-content-server#6362](https://github.com/mozilla/fxa-content-server/issues/6362)
+- **config:** For dev, the openid issuer is http://localhost:3030 (#583) r=@vladikoff ([38e1d73](https://github.com/mozilla/fxa-auth-server/commit/38e1d73)), closes [#583](https://github.com/mozilla/fxa-auth-server/issues/583) [mozilla/fxa-content-server#6362](https://github.com/mozilla/fxa-content-server/issues/6362)
```
vladikoff, line 1800: probably revert the changelog changes?

packages/fxa-customs-server/scripts/block-ip.js
```diff
@@ -9,7 +9,7 @@ Block an ip address for a number of seconds.
 
 Example:
 
-$ ./block-ip.js localhost:11211 127.0.0.1 600
+$ ./block-ip.js localhost:11211 localhost 600
```
vladikoff, line 12: hm would this here still need to be an ip address?

PR #4953, opened 2020-04-15
test(functional): add testcase for old and blank recovery key

ashrivastava-qa:
Added testcase for:
- entering blank recovery key and verifying with the error message
- trying to use old recovery key to reset password (expect error) and then trying with the new one(expect sign in) 

@vbudhram - Please review.

### DIFF
--- a/packages/fxa-content-server/tests/functional/recovery_key.js
+++ b/packages/fxa-content-server/tests/functional/recovery_key.js
@@ -92,7 +92,7 @@ registerSuite('Recovery key', {
           .then(click(selectors.RECOVERY_KEY.CONFIRM_REVOKE_OK))
           .then(testElementExists(selectors.RECOVERY_KEY.STATUS_DISABLED))
 
-          //create a new recovery key
+          // create a new recovery key
           .then(click(selectors.RECOVERY_KEY.GENERATE_KEY_BUTTON))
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/tests/functional/recovery_key.js
```diff
@@ -81,15 +81,67 @@ registerSuite('Recovery key', {
...
+          .then(
+            testElementExists(selectors.RECOVERY_KEY.CONFIRM_REVOKE_DESCRIPTION)
+          )
+          .then(click(selectors.RECOVERY_KEY.CONFIRM_REVOKE_OK))
+          .then(testElementExists(selectors.RECOVERY_KEY.STATUS_DISABLED))
 
-        .then(click(selectors.RECOVERY_KEY.CONFIRM_REVOKE_OK))
-        .then(testElementExists(selectors.RECOVERY_KEY.STATUS_DISABLED))
-        .end();
+          //create a new recovery key
```
vbudhram, line 95: Some comments have spaces before the text and some dont. I think it looks better with a space before the text.

ashrivastava-qa: made the changes. Thank you! @vbudhram

Issue #4919, opened 2020-04-14
Deprecate `marketingOptIn` in favor of passing array of `newsletters`


data-sync-user:
To help support Skyline we added the ability for a user to opt into multiple newsletters. During sign up and email registration an array of [newsletters]([https://github.com/mozilla/fxa/blob/36e21515d8a742e4439d92252420e0c626bce615/packages/fxa-auth-server/lib/routes/emails.js#L401)](https://github.com/mozilla/fxa/blob/36e21515d8a742e4439d92252420e0c626bce615/packages/fxa-auth-server/lib/routes/emails.js#L401)) were passed instead of the boolean [marketingOptIn]([https://github.com/mozilla/fxa/blob/36e21515d8a742e4439d92252420e0c626bce615/packages/fxa-auth-server/lib/routes/emails.js#L400)](https://github.com/mozilla/fxa/blob/36e21515d8a742e4439d92252420e0c626bce615/packages/fxa-auth-server/lib/routes/emails.js#L400))

The code to use newsletters array has been live for a while and we can depreciate using the marketingOptIn param. If we are re-enabling newletter during sign-up, it would be nice to have the cleanup done.

We should remove marketingOptIn from content-server UX and the auth-server routes. There are some unknowns around removing this param from our resume token and the impact of that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1622)
┆Issue Number: FXA-1622

PR #4961, opened 2020-04-16
chore(cleanup): Remove the `marketingOptIn` param

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/4919

This PR removes the last remains of the `marketingOptIn` param. 

Confirmed with marketing/basket teams that they have moved away from the `marketingOptIn` param and this is safe to merge.


vladikoff:
I am excited for this!

PR #4985, opened 2020-04-16
fix(content-server): fixed build-production after #4763

dannycoates:
Since #4763 introduced async/await the uglifyjs minimizer
fails. The built in webpack minimizer (Terser) does handle
it and performs about equally well.

The other issue is whether we transpile async/await for
older browsers, hence the addition of babel-loader to .ts
files. <strike>I opted for the safer option of doing the transpile
for now. We can defer that decision until later.</strike>

Unless we want to add babel/polyfill we should go with native async/await. According to https://caniuse.com/#feat=async-functions all our supported browsers have it.


chenba:
message from functional test failure screencaps: "regeneratorRuntime is not defined" 🤷

dannycoates:
The current test failure seems to be related to https://github.com/mozilla/fxa/pull/4818

PR #4996, opened 2020-04-17
feat(settings): hello new settings

dannycoates:
Literally just `npx create-react-app --template typescript` and adding to the rest of the build.

<strike>I don't think we need to proxy anything yet, and when we do it'll be to the GQL server, right?</strike> -- calls to GQL server will be to a new host, so not proxied. I can't think of any reason right now to proxy to content-server.

On local dev `localhost:3030/beta/settings` proxies to the settings-react dev server and the prod build copies the react assets into content and host them when the env `NEW_SETTINGS=true` is set.

@jodyheavener @LZoog


dannycoates:
Push to this branch or merge to master whenever

Issue #2088, opened 2019-08-06
sasslint not run on content server checkins


shane-tomlinson:
I just ran `grunt lint` in the content server and saw a bunch of errors from the sasslint task:

```bash
➜  fxa-content-server git:(master) ✗ grunt sasslint                                                                 git:(master|…
Running "sasslint:sass" (sasslint) task

app/styles/fontello.scss
  54:1  warning  Space expected between blocks  empty-line-between-blocks
  57:1  warning  Space expected between blocks  empty-line-between-blocks
  60:1  warning  Space expected between blocks  empty-line-between-blocks

app/styles/modules/_chosen.scss
  123:5  error  Expected `clip`, found `clip-path`                                        property-sort-order
  124:5  error  Expected `clip-path`, found `clip`                                        property-sort-order
  167:5  error  Expected `border`, found `border-bottom-left-radius`                      property-sort-order
  168:5  error  Expected `border-bottom-left-radius`, found `border-bottom-right-radius`  property-sort-order
  169:5  error  Expected `border-bottom-right-radius`, found `border`                     property-sort-order

app/styles/modules/_settings.scss
   46:3   error    Expected `background`, found `position`                          property-sort-order
   47:3   error    Expected `color`, found `top`                                    property-sort-order
   48:3   error    Expected `padding`, found `width`                                property-sort-order
   49:3   error    Expected `position`, found `background`                          property-sort-order
   50:3   error    Expected `text-align`, found `color`                             property-sort-order
   50:10  warning  Color 'white' should be written in its hexadecimal form #ffffff  no-color-keywords
   51:3   error    Expected `top`, found `padding`                                  property-sort-order
   52:3   error    Expected `width`, found `text-align`                             property-sort-order
   55:5   error    Expected `color`, found `text-decoration`                        property-sort-order
   56:5   error    Expected `text-decoration`, found `color`                        property-sort-order
   56:12  warning  Color 'white' should be written in its hexadecimal form #ffffff  no-color-keywords
  347:5   error    Expected `line-height`, found `margin`                           property-sort-order
  348:5   error    Expected `margin`, found `line-height`                           property-sort-order
  397:5   error    Expected `line-height`, found `padding-left`                     property-sort-order
  399:5   error    Expected `padding-left`, found `line-height`                     property-sort-order
  404:5   error    Expected `margin`, found `padding-left`                          property-sort-order
  405:5   error    Expected `padding-left`, found `margin`                          property-sort-order
[...]

PR #4998, opened 2020-04-17
style(fxa-content-server): fix lint errors

byrw:
issue #2088

### DIFF
--- a/packages/fxa-content-server/app/styles/_breakpoints.scss
+++ b/packages/fxa-content-server/app/styles/_breakpoints.scss
@@ -40,11 +40,11 @@ $breakpoints: (
 @mixin min-width($breakpoint) {
   @include breakpoint-error-check($breakpoint) {
     @if $breakpoint == 'tablet' {
-      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) and (min-height: (#{map-get($breakpoints, 'mobileLandscape') + 1px})) {
+      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) and (min-height: (#{map-get($breakpoints, 'mobileLandscape') + 1px})) { // sass-lint:disable-line zero-unit
         @content;
       }
     } @else {
-      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) {
+      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) { // sass-lint:disable-line zero-unit
         @content;
       }
     }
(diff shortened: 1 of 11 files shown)


packages/fxa-content-server/app/styles/_breakpoints.scss
```diff
@@ -40,11 +40,11 @@ $breakpoints: (
 @mixin min-width($breakpoint) {
   @include breakpoint-error-check($breakpoint) {
     @if $breakpoint == 'tablet' {
-      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) and (min-height: (#{map-get($breakpoints, 'mobileLandscape') + 1px})) {
+      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0}) and (min-height: (#{map-get($breakpoints, 'mobileLandscape') + 1px})) {
         @content;
       }
     } @else {
-      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0px}) {
+      @media (min-width: #{map-get($breakpoints, $breakpoint) + 0}) {
```
LZoog, line 47: Hey @byrw, thanks for your PR! It actually looks great besides this line, and a similar edit on line 43. Typically we can drop the `px` on `0` CSS values, but this is a weird case where we coerce the value into `px` here so it's required for these mixins. If the linter is complaining, a comment above the line to ignore the next line would be great.

byrw: Thank you @LZoog.

packages/fxa-content-server/app/styles/_breakpoints.scss
```diff
@@ -40,10 +40,12 @@ $breakpoints: (
 @mixin min-width($breakpoint) {
   @include breakpoint-error-check($breakpoint) {
     @if $breakpoint == 'tablet' {
+      // This must be 0px, ignore the lint warning.
```
LZoog, line 43: @byrw sorry for being a couple of days late on seeing your update here. I should have clarified and asked you to add `// sass-lint:disable zero-unit` to these lines to actually prevent the warning from showing up when running `grunt lint`, but I just tried this quickly myself and I think due to line 43's length, it's not picking up the comment. If you have time to investigate this that would be great, otherwise I'll go ahead merge this early next week since a number of other warnings are taken care of with your PR. :)

byrw: @LZoog thank you for clarifying my misunderstanding.

About the `// sass-lint:disable zero-unit`. According to the source code
https://github.com/aszwemin/sass-lint/blob/19df7ba0a58d38c1a16913a757a274f01e1f089f/lib/ruleToggler.js#L180
`// sass-lint:disable zero-unit`, does not work probably because the sass-lint use the column from where the comment started, as a starting point to disable the rule. 
Example:
```
p {
  margin-left: 0px; // sass-lint:disable zero-unit
  margin-right: 0px;
}
```
The disable will only work on margin-right, and ignore margin-left.

PR #5000, opened 2020-04-17
refactor(content-server): Remove Reference to Chromeless param

clouserw:


clouserw:
Thanks again for the patch @apsychogirl

Issue #4819, opened 2020-04-09
[setup] Setup React shell, ts-config, serve from `/beta/settings`


data-sync-user:
Ideally we use the server in `fxa-content-server` to avoid a lot of configuration work (plus more once it leaves beta).

Add a route to the `fxa-content-server` to serve the build from a new `fxa-settings` React app while in development. The new app source can live adjacent to (so outside of) `fxa-content-server` in its own package.

Get the application running in development and building to `/beta/settings` from the content-server and any additional setup around this. Setting up AppLayout etc. will be done in a follow up issue.

Thanks to [~dcoates@mozilla.com] for the image:  !image-2020-04-09-11-42-34-193.png|thumbnail!

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1539)
┆Issue Number: FXA-1539
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/12954/image-2020-04-09-11-42-34-193.png">image-2020-04-09-11-42-34-193.png</a>

data-sync-user:
➤ Wil Clouser commented:

Getting this resolved will unblock a lot of other issues.

data-sync-user:
➤ Les Orchard commented:

Oof, hadn't seen this issue before and thought the new settings would be hosted on its own domain.

That shifts things a bit for FXA-1559 and the cross-domain assumptions in this ADR PR ( https://github.com/mozilla/fxa/pull/4931 ) - i.e. we don't need to worry about cross-domain issues because a frontend hosted at accounts.firefox.com can access the session token directly

So, if the frontend for Settings Redesign gets hosted on content-server at /beta/settings - at what domain does the GraphQL server live? Assuming that's part of FXA-1531

data-sync-user:
➤ Lauren Zugai commented:

Note: per this ADR ( https://github.com/mozilla/fxa/pull/4897 ), we can use `create-react-app`. I estimated this at a 3 1) because of what this sets up for us and 2) we have examples of CRA setups in payments and the `fxa-admin-panel`, so the "unknowns" are around serving the build from the content server, but [~dcoates@mozilla.com] if there are any gnarly configs we need, this could be a 5.

data-sync-user:
➤ Les Orchard commented:

[...]

PR #5001, opened 2020-04-17
feat(settings): add rescripts, allow external imports

jodyheavener:
Sort of an extension to #4819, but also a precursor for #4817 

This PR allows `fxa-settings` to import React components beyond its `./src` directory. This is not allowed in a standard CRA setup, so we are introducing [Rescripts](https://github.com/harrysolovay/rescripts) per [ADR 0013](https://github.com/mozilla/fxa/blob/master/docs/adr/0013-react-toolchain-for-settings-redesign.md#decision-outcome) to allow us to override CRA's Webpack configuration.

I've documented what I'm doing in `.rescriptsrc.js` -- it might _look_ scary, but it's actually pretty straightforward.

### DIFF
--- a/packages/fxa-settings/.rescriptsrc.js
+++ b/packages/fxa-settings/.rescriptsrc.js
@@ -1,0 +1,59 @@
+const { resolve } = require("path");
+
+const additionalJSImportPaths = [resolve("../fxa-components")];
+
+const permitAdditionalJSImports = config => {
+  // Update ModuleScopePlugin's appSrcs to allow our new directory
+  config.resolve.plugins.forEach(plugin => {
+    if (plugin.constructor && plugin.constructor.name === "ModuleScopePlugin") {
+      plugin.appSrcs.push(...additionalJSImportPaths);
+    }
+  });
+
+  // We need to target a loader that handles compiling the JS/TS/JSX/TSX files, which exists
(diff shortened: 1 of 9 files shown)


jodyheavener:
cool tests bro

dannycoates:
I don't think fxa-shared is the right place for components. IMO it should be in it's own thing `fxa-components` maybe? I can be convinced otherwise, but shared already needs some housekeeping. This might be worth a separate thread, but I think it makes sense to at least have shared code meant exclusively for the backend in it's own package.

packages/fxa-settings/.rescriptsrc.js
```diff
@@ -0,0 +1,59 @@
...
+  //           },
+  //         ]
+  //       }
+  //     ]
+  //   }
+  // }
+  if (
+    config.module.rules[2] &&
+    config.module.rules[2].hasOwnProperty("oneOf") &&
+    config.module.rules[2].oneOf[1].test.toString() ===
```
dannycoates, line 42: This is fine. If path churn here becomes a problem maybe [jsonpath](https://www.npmjs.com/package/jsonpath) could help.

jodyheavener:
@LZoog just want to make sure you get eyes on this, since we talked about fxa-shared previously (it's going in fxa-components now). I also added an additional layer of checking in Rescripts since you last looked at that code in my demo.

PR #5009, opened 2020-04-17
feat(metrics-processor): add's queue worker

### LINKED ISSUE #4698: fxa-metrics-processor: Create pubsub worker, message processor class
Create the pubsub queue worker and config, with basic test that it can operate. It should use the GCP Pubsub polling API, with batch size configurable.

Should result in a project that can be compiled, run, and will drain a given configured queue and output it to a given queue destination. This should copy the src and config/bin/lib layouts of TS projects like the event-broker or admin-server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1446)
┆Issue Number: FXA-1446

bbangert:
Because:

* Metrics processor should be run on raw event queue.

This commit:

* Add's the queue worker with base metric processor class that
  consumes using polling from the PubSub raw event queue and
  outputs processed messages to the amplitude topic. Messages
  that failed to process are added to the dead letter topic.

Closes #4698

Co-authored-by: Barry Chen <chenba@gmail.com>

### DIFF
--- a/packages/fxa-metrics-processor/src/bin/queueWorker.ts
+++ b/packages/fxa-metrics-processor/src/bin/queueWorker.ts
@@ -1,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import 'reflect-metadata';
+
+import { Container } from 'typedi';
+import Sentry from '@sentry/node';
+
+import { version } from '../lib/version';
+import Config from '../config';
+import { QueueProcessor } from '../lib/pubsub';
+import mozlog from 'mozlog';
+
+const logger = mozlog(Config.get('log'))('queueWorker');
+const config = Config.getProperties();
+
+function main() {
+  Sentry.init({
+    dsn: config.sentryDsn,
+    release: version.version,
+    integrations: [new Sentry.Integrations.LinkedErrors()]
+  });
+  Sentry.configureScope(scope => {
+    scope.setTag('process', 'metrics_queue_worker');
+  });
+  const queueProcessor = Container.get(QueueProcessor);
+  process.on('beforeExit', async () => {
+    await queueProcessor.stop();
+    await Sentry.close();
+    logger.info('exitComplete', {});
+  });
+  return queueProcessor.start();
+}
+
+main().catch(err => {
+  logger.error('runtime', err);
+});
--- a/packages/fxa-metrics-processor/src/config.ts
+++ b/packages/fxa-metrics-processor/src/config.ts
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import convict from 'convict';
+import fs from 'fs';
+import path from 'path';
+
+const conf = convict({
+  env: {
+    default: 'production',
+    doc: 'The current node.js environment',
+    env: 'NODE_ENV',
+    format: ['development', 'test', 'stage', 'production']
+  },
+  logging: {
+    app: { default: 'fxa-user-admin-server' },
+    fmt: {
+      default: 'heka',
+      env: 'LOGGING_FORMAT',
+      format: ['heka', 'pretty']
+    },
+    level: {
+      default: 'info',
+      env: 'LOG_LEVEL'
+    },
+    routes: {
+      enabled: {
+        default: true,
+        doc: 'Enable route logging. Set to false to trim CI logs.',
+        env: 'ENABLE_ROUTE_LOGGING'
+      },
+      format: {
+        default: 'default_fxa',
+        format: ['default_fxa', 'dev_fxa', 'default', 'dev', 'short', 'tiny']
+      }
+    }
+  },
+  queues: {
+    pubsubProjectId: {
+      default: 'default-project',
+      doc: 'GCP Pub/Sub Project Id the queues are under',
+      env: 'PUBSUB_PROJECT_ID',
+      format: 'String'
+    },
+    rawEvents: {
+      default: 'fxa-metrics-raw-events',
+      doc: 'GCP Pub/Sub Queue for FxA Raw events',
+      env: 'PUBSUB_RAW_EVENT_QUEUE',
+      format: 'String'
+    },
+    deadLetterTopic: {
+      default: 'fxa-metrics-dead-letter',
+      doc: 'GCP Pub/Sub topic for FxA metric event failures',
+      env: 'PUBSUB_DEADLETTER_QUEUE',
+      format: 'String'
+    },
+    amplitudeTopic: {
+      default: 'fxa-metrics-amplitude',
+      doc: 'GCP Pub/Sub topic for transformed Amplitude events',
+      env: 'PUBSUB_AMPLITUDE_QUEUE',
+      format: 'String'
+    },
+    maxEventsPerBatch: {
+      default: 500,
+      doc: 'GCP Pub/Sub events to pull per batch',
+      env: 'PUBSUB_MAX_EVENT_BATCH',
+      format: 'Number'
+    }
+  },
+  sentryDsn: {
+    default: '',
+    doc: 'Sentry DSN for error and log reporting',
+    env: 'SENTRY_DSN',
+    format: 'String'
+  }
+});
+
+// handle configuration files.  you can specify a CSV list of configuration
+// files to process, which will be overlayed in order, in the CONFIG_FILES
+// environment variable.
+
+// Need to move two dirs up as we're in the compiled directory now
+const configDir = path.dirname(path.dirname(__dirname));
+let envConfig = path.join(configDir, 'config', `${conf.get('env')}.json`);
+envConfig = `${envConfig},${process.env.CONFIG_FILES || ''}`;
+const files = envConfig.split(',').filter(fs.existsSync);
+conf.loadFile(files);
+conf.validate({ allowed: 'strict' });
+
+const Config = conf;
+
+export default Config;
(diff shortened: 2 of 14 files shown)


packages/fxa-metrics-processor/src/config.ts
```diff
@@ -0,0 +1,93 @@
...
+      format: ['heka', 'pretty']
+    },
+    level: {
+      default: 'info',
+      env: 'LOG_LEVEL'
+    },
+    routes: {
+      enabled: {
+        default: true,
+        doc: 'Enable route logging. Set to false to trimming CI logs.',
```
6a68, line 30: 'to trim'

packages/fxa-metrics-processor/src/config.ts
```diff
@@ -0,0 +1,93 @@
...
+  queues: {
+    pubsubProjectId: {
+      default: 'default-project',
+      doc: 'GCP Pub/Sub Project Id the queues are under',
+      env: 'PUBSUB_PROJECT_ID',
+      format: 'String'
+    },
+    rawEvents: {
+      default: 'fxa-metrics-raw-events',
+      doc: 'GCP Pub/Sub Queue for FxA Raw events',
```
6a68, line 48: Would be good to be consistent about the use of "topic" vs. "queue" in the comment strings and the constant names.

packages/fxa-metrics-processor/src/config.ts
```diff
@@ -0,0 +1,93 @@
...
+    format: 'String'
+  }
+});
+
+// handle configuration files.  you can specify a CSV list of configuration
+// files to process, which will be overlayed in order, in the CONFIG_FILES
+// environment variable.
+
+// Need to move two dirs up as we're in the compiled directory now
+const configDir = path.dirname(path.dirname(__dirname));
```
6a68, line 84: Comment: I have mixed feelings about this bit at the bottom, which seems to get copied around a lot. I don't know if there's an effective way to extract it into fxa-shared or upstream it into the convict config library. Not necessarily any action to take here, but would be nice to DRY this up.

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+    this.subClient = new v1.SubscriberClient();
+    this.pubSubClient = new PubSub({ projectId: queueConfig.pubsubProjectId });
+    this.formattedRawEventSubscription = this.subClient.subscriptionPath(
+      queueConfig.pubsubProjectId,
+      queueConfig.rawEvents
+    );
+  }
+
+  /**
+   * Retrieve a given batch of messages from the raw event queue.
```
6a68, line 37: I see an `await` not wrapped in a try-catch. Is it possible for this function to throw errors? What happens if the client connection fails? Would be good to document expected behavior in the doc block.

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    this.sentry.withScope(scope => {
+      scope.setContext('message', {
+        id: message.id,
+        deadLetterId: messageId,
+        publishTime: message.publishTime,
+        data: message.data.toString(),
+        attributes: JSON.stringify(message.attributes)
+      });
+      this.sentry.captureException(err);
```
6a68, line 73: If an error message is sent to the DLQ, but sentry submission fails, how will we reconcile the accounting later?

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+   *   2. Handle messageProcessor errors by sending them to a dead letter topic
+   *      and report them to Sentry.
+   *   3. Acknowledge the batch of messages.
+   */
+  private async processBatch() {
+    const messages = await this.pullMessageBatch();
+    const ackIds: string[] = [];
+    for (const message of messages) {
+      try {
+        const newMessage = this.messageProcessor.processMessage(message);
```
6a68, line 93: I am a little surprised that message processing is synchronous, but that simplifies life, so 🎉

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+    let handled = true;
+    try {
+      messageId = await this.pubSubClient.topic(queueConfig.deadLetterTopic).publish(message.data);
+    } catch (publishErr) {
+      // Replace our error and link to the original one
+      publishErr.cause = err;
+      err = publishErr;
+      handled = false;
+    }
+    this.sentry.withScope(scope => {
```
6a68, line 65: If `handled = false` at this point, should we submit to sentry?

packages/fxa-metrics-processor/src/bin/queueWorker.ts
```diff
@@ -0,0 +1,37 @@
...
+  Sentry.init({
+    dsn: config.sentryDsn,
+    release: version.version,
+    integrations: [new Sentry.Integrations.LinkedErrors()]
+  });
+  Sentry.configureScope(scope => {
+    scope.setTag('process', 'metrics_queue_worker');
+  });
+  const queueProcessor = Container.get(QueueProcessor);
+  process.on('exit', async () => {
```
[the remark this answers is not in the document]
6a68: Not sure, but I think we may also want to explicitly drain the sentry client before shutdown: https://docs.sentry.io/error-reporting/configuration/draining/?platform=javascript

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+    };
+    const [response] = await this.subClient.pull(request);
+    return response.receivedMessages;
+  }
+
+  /**
+   * Handle an error from the message transformation.
+   *
+   * @param message Message that caused the error
+   * @param err Error processing the message
```
6a68, line 52: Would be good to document that this function does not throw, just returns a success/failure boolean.

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+    const ackRequest = {
+      subscription: this.formattedRawEventSubscription,
+      ackIds
+    };
+    await this.subClient.acknowledge(ackRequest);
+  }
+
+  public async start() {
+    while (!this.shouldStop) {
+      await this.processBatch();
```
6a68, line 116: Should this be wrapped in a try-catch, so we can log errors encountered when trying to pull batches?

packages/fxa-metrics-processor/src/lib/pubsub.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (handled) {
+          ackIds.push(message.ackId);
+        }
+      }
+    }
+    const ackRequest = {
+      subscription: this.formattedRawEventSubscription,
+      ackIds
+    };
+    await this.subClient.acknowledge(ackRequest);
```
6a68, line 111: Just to be sure I'm understanding correctly, if the client ack fails at this point, some client processing timer will eventually expire, and the messages will be retried by another worker, is that right?

packages/fxa-metrics-processor/src/test/lib/pubsub.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      it('works', async () => {
+        const emitter = new EventEmitter();
+        queueProcessor.processBatch = () => once(emitter, 'next');
+        assert.equal(queueProcessor.shouldStop, false);
+        // Don't wait for the promise
+        queueProcessor.start();
+
+        // Iterate the loop and yield event loop so it runs
+        emitter.emit('next');
+        await sleep(0);
```
6a68, line 195: Setting the timeout like this is fine, though sinon does provide a [mock clock](https://sinonjs.org/releases/v1.17.6/fake-timers/) that can be used to advance time in a more controlled way.

packages/fxa-metrics-processor/src/config.ts
```diff
@@ -0,0 +1,93 @@
...
+import convict from 'convict';
+import fs from 'fs';
+import path from 'path';
+
+const conf = convict({
+  env: {
+    default: 'production',
+    doc: 'The current node.js environment',
+    env: 'NODE_ENV',
+    format: ['development', 'test', 'stage', 'production']
```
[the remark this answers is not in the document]
bbangert: I believe we settled on the longer name.

packages/fxa-metrics-processor/src/config.ts
bbangert: Well, technically instead of queue, it could be called subscription. Topic is separate though, as a topic can end up publishing to multiple subscriptions. Queue in this case is being used as a 'subscriber', while for publishing in PubSub we do not publish directly to a subscription, but a topic. See: https://cloud.google.com/pubsub/docs/overview#publisher-subscriber-relationships

packages/fxa-metrics-processor/src/config.ts
bbangert: Indeed, and at some point it could be more hassle to pull it in everywhere than copy/paste it.

packages/fxa-metrics-processor/src/lib/pubsub.ts
bbangert: Yup.

packages/fxa-metrics-processor/src/test/lib/pubsub.spec.ts
bbangert: This isn't being used to timeout, it's being used to yield the event loop so the listener waiting on the emitter event can be run. The mock clock does not yield the event loop, its synchronous.

packages/fxa-metrics-processor/src/lib/pubsub.ts
bbangert: We will submit it to Sentry, note on line 62 that it changes the err reference to publishErr, with the original chained on.

packages/fxa-metrics-processor/src/lib/pubsub.ts
bbangert: The return boolean is already documented as the type signature. Or do you mean something else?

packages/fxa-metrics-processor/src/lib/pubsub.ts
bbangert: I was under the impression that Sentry would report this and ops would notice it bail since this would cause the program to exit. Sentry will capture the otherwise unhandled exception.

PR #5032, opened 2020-04-20
fix(test): Attempt to fix denormalization signin error

vbudhram:
Hopefully fixes the `sign in with incorrect email case before normalization fix, on second attempt canonical form is used` test. This passed 7/7 times locally.

### DIFF
--- a/packages/fxa-content-server/tests/functional/sign_in_cached.js
+++ b/packages/fxa-content-server/tests/functional/sign_in_cached.js
@@ -87,21 +87,11 @@ registerSuite('cached signin', {
           // done on the settings page, wait for the signin page to load,
           // denormalize, then refresh. See #4711
           .then(denormalizeStoredEmail(email))
+
           .refresh()
-          // This test is quite flakey without a pause here.
-          // Running locally I see a flash of the lowercase email
-          // before it changes to upper. I don't know where that
-          // first render is happening. - (dcoates)
-          .sleep(500)
+
           .then(testElementExists(selectors.SIGNIN_PASSWORD.HEADER))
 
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/tests/functional/sign_in_cached.js
```diff
@@ -87,21 +87,11 @@ registerSuite('cached signin', {
...
-          // first render is happening. - (dcoates)
-          .sleep(500)
+
           .then(testElementExists(selectors.SIGNIN_PASSWORD.HEADER))
 
-          // email is not yet normalized :(
-          .then(
-            testElementTextEquals(
-              selectors.SIGNIN_PASSWORD.EMAIL_NOT_EDITABLE,
-              email.toUpperCase()
```
dannycoates, line 102: Yes, removing this assert makes it pass, but I assume it was there on purpose. Should we change our expectation of what this test is checking? I find it pretty confusing anyway but it seems to me that if the inputs are the same, and it ends up at the settings page, this intermediate ui state being one case or another isn't very important. So, I'm in favor of removing this assert, but I just want to be sure I'm not missing anything important.

vbudhram: > it seems to me that if the inputs are the same, and it ends up at the settings page, this intermediate ui state being one case or another isn't very important

Yea, I agree with you on that. From searching through the commits I could not really find what the original intent for this test was. Maybe older versions of FF, localstorage performance was not as good and it really did show the denormalized email. Going to merge this since the intermediate ui assert does not really seem to provide value.

PR #5040, opened 2020-04-21
More shared component stuff

jodyheavener:
Just helps pave the way for better shared React components integration.

- Sets up Rescripts in fxa-payments-server - so it can make use of shared components
- Adds very basic Storybook setup in fxa-components so we can start setting up visual components
- Adds fxa-components to CI test-many


jodyheavener:
@lmorchard I set up the bare bones for Storybook in fxa-components, but I know very little else about it. Is there a way to auto-import components rather than defining them manually in story files?

lmorchard:
> Is there a way to auto-import components rather than defining them manually in story files?

No, stories are like tests - you have to manually rig them up with expected scenarios through appropriately composed React props.

packages/fxa-components/package.json
```diff
@@ -18,13 +18,20 @@
     "node": ">=12"
   },
   "scripts": {
-    "test": "jest"
+    "test": "jest",
+    "storybook": "start-storybook -p 6006",
```
LZoog, line 22: We _may_ want to have different ports for different storybook instances, especially as we're moving components around into `fxa-components` and want to check them in payments/settings, but I'm also fine with leaving them all at 6006 until someone says it's an inconvenience 🤷‍♀️

jodyheavener:
Hmm. I wonder how useful Storybook will be here (specifically for individual components). @LZoog whatchu think?

LZoog:
> Hmm. I wonder how useful Storybook will be here (specifically for individual components). @LZoog whatchu think?

I think @meandavejustice wanted to push for storybook in `fxa-components` so I'll definitely wait for his input here, it sounded like it could be nice to show all of the different states for a shared component in a `fxa-components` Storybook... However, @lmorchard made a good point though, "there are some payments-specific mock app things set up for more integrated exercise of routes (we'd probably want similar things in settings refresh" so honestly, I'm fine with having it or not having it here, so long as the state is shown in the application where it's used. It could be extra overhead to add to storybook in two places, but I can see the argument for having it here too.

lmorchard:
> I wonder how useful Storybook will be here (specifically for individual components).

Ideally, that's where Storybook is *most* useful. Set up components in isolation with stories that run them through all their paces. Edit the code & styles, see changes via hot module reload. Then, compose them into more complicated views and repeat. The value should be that you can iterate on things without needing the entire stack running from frontend down through backend data APIs

meandavejustice:
@jodyheavener @LZoog Looks like @lmorchard beat me to it. I was gunna have the exact response as him. 👍

PR #5046, opened 2020-04-21
feat(recovery keys): added upsertRecoveryKey db method

### LINKED ISSUE #4268: Consider using using upsertRecoveryKey database method

farhan787:
fixes #4268

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-109-110.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-109-110.sql
@@ -1,0 +1,63 @@
+-- This migration will help us to abstract the half baked recovery keys of fxa accounts. In this migration,
+-- we're adding a new `upsertRecoveryKey` procedure that will handle the half baked recovery keys by using
+-- the following criterias.
+
+-- 1. If a user has no recovery key then create a new recovery key.
+-- 2. If a user has unverified recovery key then delete that recovery key and create a new unverified
+--    recovery key.
+-- 3. If a user has verified recovery key then throw an exception
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('109');
+
+CREATE PROCEDURE `upsertRecoveryKey_1` (
+  IN `uidArg` BINARY(16),
+  IN `recoveryKeyIdHashArg` BINARY(32),
+  IN `recoveryDataArg` TEXT,
+  IN `createdAtArg` BIGINT UNSIGNED,
+  IN `enabledArg` BOOLEAN
+)
+BEGIN
+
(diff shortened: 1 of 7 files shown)


farhan787:
@vbudhram I've updated this PR as the PR responsible for updating procedures for setting`verifiedAt` column in `email` table has been merged. IMHO I think it's ready for your final review, please take a look at it.

farhan787:
@vbudhram [this test (should fail to create multiple keys)](https://github.com/mozilla/fxa/blob/9a52a7efdbe5bd447b0b10d6ce4ab6636e8de899/packages/fxa-auth-server/test/remote/recovery_key_tests.js#L209) is failing in `auth-server` because the `recoveryKey` isn't verified there and as we've replaced the `createRecoveryKey` method with `upsertRecoveryKey` method which creates a new unverified recovery key if a verified recoveryKey doesn't exist. 

To fix this we can do 2 things, 
- Either verify the recoveryKey in `auth-server` test which I think might not be right as we'd be using upsertRecoveryKey in `auth-db-mysql` and treating it as `createRecoveryKey` in `auth-server`. 
- Or have both `createRecoveryKey` and `upsertRecoveryKey` methods for client to call.

Waiting for your instructions.

packages/fxa-auth-db-mysql/lib/db/schema/patch-109-110.sql
```diff
@@ -0,0 +1,63 @@
...
+  -- Signal error if no user found
+  SET @accountCount = 0;
+  SELECT COUNT(*) INTO @accountCount FROM `accounts` WHERE `uid` = `uidArg`;
+  IF @accountCount = 0 THEN
+    SIGNAL SQLSTATE '45000' SET MYSQL_ERRNO = 1643, MESSAGE_TEXT = 'Can not create recovery key for unknown user.';
+  END IF;
+
+  -- Signal error if verified recovery key already exists
+  SET @verifiedRecoveryKeyCount = 0;
+  SELECT COUNT(*) INTO @verifiedRecoveryKeyCount FROM `recoveryKeys` WHERE `uid` = `uidArg` AND verifiedAt IS NOT NULL;
```
vbudhram, line 37: I've double checked with our OPs team and it should be good to use select on the `enabled` column verses `verifiedAt`.

farhan787:
Blocked because of this, as my local dev environment is broken https://github.com/mozilla/fxa/issues/5216

clouserw:
Please rebase/reopen when ready for merging

PR #5049, opened 2020-04-21
docs(contributing): Add some context about FxA to contributing doc

clouserw:
One of the things I've been thinking about for the past few weeks is how FxA should participate (or not) in the coding contributions side of open source.  It's no question that it's time consuming to work with contributors to get FxA up and running and can be frustrating when they end up disappearing after a few days of work.  On the other hand, many of us got our starts committing small patches to projects after being helped by the project maintainers.

I think this patch sets some context for the project that will help set realistic expectations.  I think the actual throughput of contributions (and time spent) should be controlled by whether or not we are tagging `help wanted` and `good first issue` on our issues.

If we get into a crunch period where we don't have much time, or, for example with the Settings project, we're making big architectural changes that will obsolete many old/small bugs, then we should remove the labels from those issues and stop advertising for code contributions.

If we find some good small bugs and someone has time to mentor them, then we should add the label.  In short, the lever to move for how much time is spent with mentoring good first issues is to label them appropriately.

Comments welcome!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,27 @@
-# Contribution Guidelines for the Firefox Accounts Content Server
+# Contribution Guidelines for Firefox Accounts
 
-Anyone is welcome to help with Firefox Accounts. Feel free to get in touch with other community members on Matrix, the
-mailing list or through issues here on GitHub.
+Anyone is welcome to help with Firefox Accounts, but it's important to
+recognize that Firefox Accounts is a complex tool written by Mozilla and used
+solely within Mozilla or by parties tightly coupled to Mozilla.  The team's
+goal is focused on serving Mozilla's requirements -- not being easy to set up
+or easily usable in other scenarios.
+
+In order to scale effectively, we may use technologies which have significant
+barriers to entry (eg. cost or expertise).
+
+If you're filing bugs, working on documentation, or helping reproduce a
+problem, thank you for your help and please read below for guidelines.
+
+If you want to write code, and you are not an experienced programmer you will
+probably be more successful [looking for other projects at
+Mozilla](https://whatcanidoformozilla.org) to contribute to.  The Firefox
+Accounts team is happy to support open source contributions but we have limited
+time to assist in getting the FxA codebase up and running on other platforms.
+
+We use the standard `help wanted` and `good first issue` labels on GitHub to
+help identify bugs for contributors to work on.
+
+To get in touch with us and other community members:
 
 - Matrix: [#fxa:mozilla.org](https://chat.mozilla.org/#/room/#fxa:mozilla.org)
 - Mailing list: <https://mail.mozilla.org/listinfo/dev-fxacct>
@@ -18,29 +38,9 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
 You can [file issues on GitHub](https://github.com/mozilla/fxa/issues/new). Please try to include as much information as you can and under what conditions
 you saw the issue.
 
-## Bug Triage
+## Development Process
 
-We triage bugs twice a week: in the Monday FxA Weekly meeting (after agenda items), and in the Friday Triage meeting (see the [Firefox Accounts Public Calendar](https://calendar.google.com/calendar/embed?src=mozilla.com_urbkla6jvphpk1t8adi5c12kic%40group.calendar.google.com) for times and details). If you are a contributor and would like to attend triage, reach out and let us know.
-
-The FxA core team uses ZenHub to measure velocity and organize work into Epics.
-
-### Estimation and Point Values
-
-Points are assigned to GitHub issues (ideally before starting work 😉) so that we can track our velocity over time, which aids in planning. Note that you'll need to be signed in to ZenHub to view and set estimates.
-
-We don't assign points to pull requests, unless it's a pull request without a corresponding issue.
-
-The goal of estimation is for us to assess the issue in terms of its relative complexity, effort, and doubt. When applying an estimate, we should consider all the steps in getting the particular work to a completed (ready for production) state for our consumer. The should include effort required for code reviews, security reviews, testing, integration and build/deploy.
-
-| Points  | Relative Size | Description                                                                |
-| ------- | ------------- | -------------------------------------------------------------------------- |
-| 1       | xs            | This is a trivial change with clearly defined parameters.                  |
-| 2       | s             | This is a small change with clearly defined parameters.                    |
-| 3       | m             | This is a small change, but there's some uncertainty.                      |
-| 5       | l             | Moderately complex, will require some effort but clearly defined.          |
-| 8       | xl            | Moderately complex, medium effort, some uncertainty.                       |
-| 13      | xxl           | Complex, large effort, well defined (these should be broken down further)  |
-| 21 (20) | ∞             | Meta issue or We do not have clear scope. (This issue must be broken down) |
+[Read about our development process](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-dev-process)
 
 ## Sending Pull Requests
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,43 +35,23 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
...
-
-### Estimation and Point Values
+## Development Process
 
-Points are assigned to GitHub issues (ideally before starting work 😉) so that we can track our velocity over time, which aids in planning. Note that you'll need to be signed in to ZenHub to view and set estimates.
-
-We don't assign points to pull requests, unless it's a pull request without a corresponding issue.
-
-The goal of estimation is for us to assess the issue in terms of its relative complexity, effort, and doubt. When applying an estimate, we should consider all the steps in getting the particular work to a completed (ready for production) state for our consumer. The should include effort required for code reviews, security reviews, testing, integration and build/deploy.
-
```
clouserw, line 34: I moved the points into our ecosystem docs.  There is more in this doc that could go there also, but I didn't want to confuse this PR more.

CONTRIBUTING.md
```diff
@@ -1,7 +1,24 @@
...
 
-Anyone is welcome to help with Firefox Accounts. Feel free to get in touch with other community members on Matrix, the
-mailing list or through issues here on GitHub.
+Anyone is welcome to help with Firefox Accounts, but it's important to
+recognize that Firefox Accounts is a complex tool written by Mozilla and used
+solely within Mozilla or by parties tightly coupled to Mozilla.  The teams goal
+is focused on serving Mozilla's requirements -- not being easy to set up or
+easily usable in other scenarios.
+
+In order to scale effectively, we may use technolgies which have significant
```
LZoog, line 9: ~technolgies~ technologies

CONTRIBUTING.md
```diff
@@ -1,7 +1,24 @@
...
+barriers to entry (eg. cost or expertise).
+
+If you're filing bugs, working on documentation, or helping reproduce a
+problem, thank you for your help and please read below for guidelines.
+
+If you want to write code, and you are not an experienced programmer you will
+probably be more successful [looking for other projects at
+Mozilla](https://whatcanidoformozilla.org) to contribute to.  The Firefox
+Accounts team is happy to support open source contributions but we have limited
+time to assist in getting the FxA codebase up and running on other platforms.
```
LZoog, line 19: Did we want to add something about `help wanted` or `good first issue` labels here?

Overall, I think this is a good addition to this doc. I do think if folks can get the stack running, the ones we label as `good first issue` are pretty straightforward. I think another hurdle for contributors trying to grab anything that's not labeled as such is we oftentimes file issues with the expectation that one of us will grab it - there's an assumed level of knowledge/familiarity, and a contributor might need additional context to understand what some issues are even about.

I also don't want to discourage anyone from trying, though. Maybe the way you've phrased it here is the right way to convey this.

CONTRIBUTING.md
```diff
@@ -1,7 +1,24 @@
 # Contribution Guidelines for the Firefox Accounts Content Server
```
chenba, line 1: Is this really still content server specific?

CONTRIBUTING.md
```diff
@@ -1,7 +1,24 @@
...
 
-Anyone is welcome to help with Firefox Accounts. Feel free to get in touch with other community members on Matrix, the
-mailing list or through issues here on GitHub.
+Anyone is welcome to help with Firefox Accounts, but it's important to
+recognize that Firefox Accounts is a complex tool written by Mozilla and used
+solely within Mozilla or by parties tightly coupled to Mozilla.  The teams goal
```
chenba, line 5: "team's"

chenba:
👍 on clarifying the purpose of FxA and the barriers to entry.

CONTRIBUTING.md
```diff
@@ -18,43 +35,23 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
...
+[Read about our development process](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-dev-process)
 
 ## Sending Pull Requests
 
 Patches should be submitted as pull requests (PR).
 
 Before submitting a PR:
 
-- Your code must run and pass all the automated tests before you submit your PR for review. "Work in progress" pull requests are allowed to be submitted, but should be clearly labeled as such and should not be merged until all tests pass and the code has been reviewed.
+- Your code must run and pass all the automated tests before you submit your PR for review. "Work in progress" pull requests are allowed to be submitted, but should be marked `draft` and should not be merged until all tests pass and the code has been reviewed.
```
jodyheavener, line 48: Two nits:

- So we don't confuse this with labeling, maybe "opened as a Draft" instead?
- Branch protection prevents merging until CI is green and an approving review has been created, so it might be a little moot to indicate here.

CONTRIBUTING.md
```diff
@@ -18,43 +35,23 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
...
 - Your patch should include new tests that cover your changes. It is your and your reviewer's responsibility to ensure your patch includes adequate tests.
 - Your patch must be [GPG signed](https://help.github.com/articles/managing-commit-signature-verification) to ensure the commits come from a trusted source.
 
 When submitting a PR:
 
-- You agree to license your code under the project's open source license ([MPL 2.0](/LICENSE)).
+- You agree to license your code under the project's [license](/LICENSE).
 - Base your branch off the current `master`.
 - Add both your code and new tests if relevant.
 - Run `grunt lint` and `npm test` to make sure your code passes linting and tests.
```
jodyheavener, line 57: This is probably a relic of content server, but I think we could change `grunt lint` to `npm run lint` or something like that (and note that it needs to be run at the per-package level).

CONTRIBUTING.md
```diff
@@ -18,43 +35,23 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
...
 Patches should be submitted as pull requests (PR).
 
 Before submitting a PR:
 
-- Your code must run and pass all the automated tests before you submit your PR for review. "Work in progress" pull requests are allowed to be submitted, but should be clearly labeled as such and should not be merged until all tests pass and the code has been reviewed.
+- Your code must run and pass all the automated tests before you submit your PR for review. "Work in progress" pull requests are allowed to be submitted, but should be marked `draft` and should not be merged until all tests pass and the code has been reviewed.
 - Your patch should include new tests that cover your changes. It is your and your reviewer's responsibility to ensure your patch includes adequate tests.
 - Your patch must be [GPG signed](https://help.github.com/articles/managing-commit-signature-verification) to ensure the commits come from a trusted source.
 
 When submitting a PR:
```
vbudhram, line 52: One thing I think contributors sometimes don't do is flag @ fxa-devs, or some developer for review. This might happen on initial PR or when they fix followup issues. These PRs might sit around longer than usual.

Using the mark as draft/ ready to review would help reduce this.

CONTRIBUTING.md
6a68: +1, I do think it's important to restrict outside contributors to some specific good-first-bug or similar label. We want to avoid putting contributors under sprint-related time pressure, and also avoid staff spending a lot of time getting a contributor up to speed, only to have them disappear. An effective way to avoid both problems is to point contributors at lower-priority bugs with a low learning curve.

CONTRIBUTING.md
```diff
@@ -18,43 +35,23 @@ You must agree to abide by the [Mozilla Community Participation Guidelines](http
...
 
-- You agree to license your code under the project's open source license ([MPL 2.0](/LICENSE)).
+- You agree to license your code under the project's [license](/LICENSE).
```
6a68, line 54: Might be good to put a slightly finer point on this: "You agree and have the legal right to license your code..."

This can avoid the awkward situation where people have made contributions without their employer's express permission.

CONTRIBUTING.md
clouserw: I'll change #1.  I'm going to leave #2 as it helps set expectations for newcomers.

CONTRIBUTING.md
clouserw: We're also going to start looking for PRs for anything over a couple days old in our weekly triage which should help catch any stragglers. w

CONTRIBUTING.md
clouserw: thanks, fixed

CONTRIBUTING.md
clouserw: I added a sentence.  Thanks

CONTRIBUTING.md
clouserw: thanks, fixed

CONTRIBUTING.md
clouserw: :heavy_check_mark:

CONTRIBUTING.md
clouserw: Yeah, we can adjust the doc more (in particular, I'm thinking more should live in the ecosystem docs) but for this patch I just want to focus on the contributor guidelines.  Would take a PR to change that though ;)

CONTRIBUTING.md
clouserw: thanks, I added it

clouserw:
Thanks for all the comments

PR #5051, opened 2020-04-22
feat(admin-server): set hsts header on all requests

### LINKED ISSUE #5012: [fxa-admin-server] add HSTS header to response headers
the fxa-admin-server should add a strict-transport-security header with values identical to fxa-admin-panel

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1689)
┆Issue Number: FXA-1689

jodyheavener:
Closes #5012

### DIFF
--- a/packages/fxa-admin-server/src/config.ts
+++ b/packages/fxa-admin-server/src/config.ts
@@ -79,6 +79,18 @@ const conf = convict({
     doc: 'Sentry DSN for error and log reporting',
     env: 'SENTRY_DSN',
     format: 'String'
+  },
+  hstsEnabled: {
+    default: true,
+    doc: 'Send a Strict-Transport-Security header',
+    env: 'HSTS_ENABLED',
+    format: Boolean
+  },
+  hstsMaxAge: {
+    default: 31536000, // a year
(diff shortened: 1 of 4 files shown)


packages/fxa-admin-server/src/config.ts
```diff
@@ -79,6 +79,18 @@ const conf = convict({
...
+    default: true,
+    doc: 'Send a Strict-Transport-Security header',
+    env: 'HSTS_ENABLED',
+    format: Boolean
+  },
+  hstsMaxAge: {
+    default: 31536000, // a year
+    doc: 'Max age of the STS directive in seconds',
+    // Note: This format is a number because the value needs to be in seconds
+    format: Number
```
LZoog, line 93: I think we need this in `fxa-admin-panel` as well. I have the config set up for it but didn't add the middleware.

LZoog: Or... maybe not? I see the issue only lists the admin-server.

jodyheavener:
That's done [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-admin-panel/server/lib/server.ts#L67-L76) :D

PR #5056, opened 2020-04-22
refactor(auth server): typescript lib/oauth/logging

### LINKED ISSUE #4880: [auth-server] Add Typescript to files in /lib/oauth/logging
This is part of a larger effort to [adopt Typescript in the Auth Server](https://jira.mozilla.com/browse/FXA-1500). It can be broken down into smaller issues if desired. This needs to be completed before [FXA-1589 (lib/oauth)](https://jira.mozilla.com/browse/FXA-1589) can be closed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1586)
┆Issue Number: FXA-1586

jodyheavener:
Closes #4880 

_Part of a larger effort to [adopt TypeScript in the Auth Server](https://jira.mozilla.com/browse/FXA-1500)._

---

I'm trying to take a slow, educational approach to TypeScript, so please feel free to tear this apart. 😄


jodyheavener:
Ah, I think I need to update mocha to test with ts-node

jodyheavener:
Hmm.

jodyheavener:
Alright I'm marking this for ready, but with some hesitation. I think switching the Mocha tests to ts-node is creating some wicked timeouts, and this has also introduced a weird behaviour in a verification email content server test that I have not been able to nail down just yet. Thoughts welcome.

packages/fxa-auth-server/lib/oauth/logging/index.ts
```diff
@@ -2,28 +2,38 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-let mozlog = null;
+// TODO: once the mozlog module is typed we need
+// to update this type to not use `any`, and also
+// update the retun type of the function.
```
chenba, line 7: I suppose you can `packages/fxa-admin-panel/types/mozlog/index.d.ts` for now?

packages/fxa-auth-server/lib/oauth/logging/summary.ts
```diff
@@ -0,0 +1,96 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Request, UserCredentials } from 'hapi';
+import logging from './';
+
+const logger = logging('summary');
+
+interface SummaryRequest extends Request {
+  app: {
```
chenba, line 11: Looks like `Request` already has an `app` property on it.  Might be an issue if some code expects `SummaryRequest.app` to work like `Request.app`.

packages/fxa-auth-server/lib/oauth/logging/summary.ts
```diff
@@ -0,0 +1,96 @@
...
+  request: SummaryRequest,
+  response: SummaryResponse
+): void {
+  if (request.method === 'options') {
+    return;
+  }
+
+  const payload: {
+    [key: string]: any;
+  } = (request.payload as object) || {};
```
chenba, line 67: I guess it's always an object since the old code's been working?

packages/fxa-content-server/tests/functional/verification_reminders.js
```diff
@@ -46,7 +46,7 @@ registerSuite('verification reminders', {
         // The first email is a code, we need to wait for verification
         // reminders which contain a link.
         .then(sendVerificationReminders())
-        .then(getVerificationLink(email, 1))
```
chenba, line 49: Whaaa? Why are these affected?  😕

chenba: Let's not merge until we figure out what's going on here.

jodyheavener: Yeah, I'm pretty confused about those. I won't merge until I know why that's happening.

packages/fxa-auth-server/lib/oauth/logging/summary.ts
jodyheavener: Hm, this might be a spot where I'm taking the wrong approach -- I'm extending `Request.app` to add `remoteAddressChain` because it doesn't appear to exist on it by default. `Request.app` doesn't appear to have any reachable properties from what I can see.

packages/fxa-auth-server/lib/oauth/logging/index.ts
```diff
@@ -2,28 +2,32 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-let mozlog = null;
+type MozLog = { logger: Logger };
+let mozlog: null | MozLog;
 
-module.exports = function(nameOrLog) {
+module.exports = function(nameOrLog: string | MozLog): Logger | never | void {
```
chenba, line 8: Why is this `never` necessary?

packages/fxa-auth-server/lib/oauth/logging/summary.ts
chenba: Haha `ApplicationState` is just an empty interface.  I think we are good here.  Also, `SummaryRequest` is only used in one spot anyway.

packages/fxa-auth-server/lib/oauth/logging/index.ts
jodyheavener: It's _really_ necessary, but because the function could potentially throw an exception it has the possibility to `never` return anything.

packages/fxa-content-server/tests/functional/verification_reminders.js
jodyheavener: I figured out what the issue was btw, just sorting out some type declaration issues now. Should be updated soon-ish.

packages/fxa-auth-server/lib/oauth/logging/index.ts
chenba: Sure it doesn't hurt anything.  But it also doesn't do anything since `never` is a subtype of every type, in a union it gets absorbed into the other types.

jodyheavener: Got it! This is precisely what I'm looking for in these TS reviews 😄

jodyheavener:
@chenba I made a few more changes, think I could get another pass when you have a minute? Not urgent.

jodyheavener:
@dannycoates hm good point. I was looking at `scripts/build.sh` to see if I could suss out how that would work, but my Docker chops aren't quite there. If you've got time this sprint think I could work with you to set this up? Happy to pair even.

packages/fxa-auth-server/tsconfig.json
```diff
@@ -20,7 +20,8 @@
     "noFallthroughCasesInSwitch": true,
 
     "types": ["node"],
-    "typeRoots": ["./types", "node_modules/@types"]
+    "typeRoots": ["./types", "node_modules/@types"],
+    "noImplicitAny": false
```
chenba, line 24: 😢

chenba:
> 
> 
> @chenba I made a few more changes, think I could get another pass when you have a minute? Not urgent.

👍   But would prefer to not have implicit `any`s.

packages/fxa-auth-server/tsconfig.json
jodyheavener: Good point about this. I added it because my tests were failing when I tried importing a type def from `mozlog`, which I have custom typed in `types/mozlog/index.d.ts`. I can't find the definition. 

However when I switched away from `import { Logger } from 'mozlog'` to `const { Logger } = require('mozlog')`, this issue stopped.

Any idea why?

jodyheavener:
Thank you @chenba @dannycoates for your help with this.

packages/fxa-auth-server/tsconfig.json
chenba: Sorry, didn't notice this message until today.  Without looking into it, I don't know why that import wouldn't work.

PR #5083, opened 2020-04-23
feat(metrics): create $identify Amplitude payload from raw event

### LINKED ISSUE #4979: fxa-metrics-processor: create $identify Amplitude Event from Raw Events
Currently, services would include something in the "user_properties" of an "amplitudeEvent" log entry to signal `fxa-amplitude-send` to extract and create an `$identify` Amplitude event.  For `fxa-metrics-processor` we need to create the `$identify` event directly from the "raw" events.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1671)
┆Issue Number: FXA-1671

chenba:
This patch adds a function to produce the payload for Amplitude's
identify API from other FxA services' "raw" events.

It also provides a function that will accept a raw event and its context
and return the payloads for Amplitude's http, and if applicable,
identify APIs.

Fixes #4979 (FXA-1671)

### DIFF
--- a/packages/fxa-metrics-processor/src/lib/amplitude/index.ts
+++ b/packages/fxa-metrics-processor/src/lib/amplitude/index.ts
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import {
+  Services,
+  PlainEvents,
+  FuzzyEvents,
+  RawEvent,
+  EventContext,
+  AmplitudeHttpEvent,
+} from './transforms/types';
+import { createServiceNameAndClientIdMapper } from './transforms/common';
+import { createAmplitudeEventTransformer } from './transforms/amplitude-event';
+import {
+  createAmplitudeIdentityEvent,
+  AmplitudeIdentifyEvent,
+} from './transforms/amplitude-identify';
+
+export type AmplitudeRequestPayloads = {
+  http: AmplitudeHttpEvent;
+  identify?: AmplitudeIdentifyEvent;
+};
+
+/**
+ * Create a transform function from the given mappings.  A transform function
+ * should be created per service (specifically for each of the "eventSource"
+ * values in RawEvent).
+ *
+ * The transform function returns an object keyed by the Amplitude API for
+ * which the value is intended.  There are two possible keys: "http" and
+ * "identify".
+ *
+ * The "identify" property is optional.  However, if it does exist, then the
+ * API request for it MUST be ahead of the request for the http API.
+ *
+ * @param services An object of client-id:service-name mappings
+ * @param plainEvents  An object of name:definition event mappings, where
+ *                     each definition value is itself an object with `group`
+ *                     and `event` string properties.
+ *
+ * @param fuzzyEvents A map of regex:definition event mappings. Each regex
+ *                    key may include up to two capturing groups. The first
+ *                    group is used as the event "category" and the second is
+ *                    used as the event "target". Again each definition value
+ *                    is an object containing `group` and `event` properties
+ *                    but here `group` can be a string or a function. If it's
+ *                    a function, it will be passed the matched category
+ *                    as its argument and should return the group string.
+ */
+export function createAmplitudePayloadsTransfromer(
+  services: Services,
+  plainEvents: PlainEvents,
+  fuzzyEvent: FuzzyEvents
+) {
+  const servicePropsMapper = createServiceNameAndClientIdMapper(services);
+  const amplitudeHttpEventTransformer = createAmplitudeEventTransformer(
+    plainEvents,
+    fuzzyEvent,
+    servicePropsMapper
+  );
+
+  return (event: RawEvent, context: EventContext): AmplitudeRequestPayloads | null => {
+    const httpApiPayload = amplitudeHttpEventTransformer(event, context);
+
+    if (!httpApiPayload) {
+      return null;
+    }
+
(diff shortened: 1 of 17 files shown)


packages/fxa-metrics-processor/src/lib/amplitude/transforms/common.ts
```diff
@@ -40,7 +40,7 @@ export function mapLocation(context: EventContext): LocationProperties | {} {
     if (('country' in location && location.country) || ('state' in location && location.state)) {
       return {
         country: location?.country,
-        region: location?.state
+        region: location?.state,
```
chenba, line 43: File with only formatting changes. My bad.  😬

packages/fxa-metrics-processor/src/lib/amplitude/transforms/event-properties.ts
```diff
@@ -32,15 +32,15 @@ const EVENT_PROPERTIES: {
   [GROUPS.subPayManage]: NOP,
   [GROUPS.subPaySetup]: NOP,
   [GROUPS.subPayUpgrade]: NOP,
-  [GROUPS.subSupport]: NOP
+  [GROUPS.subSupport]: NOP,
```
chenba, line 35: File with formatting changes only.

packages/fxa-metrics-processor/src/lib/amplitude/transforms/types.ts
```diff
@@ -110,6 +109,7 @@ export type AmplitudeHttpEvent = {
   region?: string;
   os_name?: string;
   os_version?: string;
+  device_model?: string;
```
chenba, line 112: Missed this property in the transforms work.  😦

packages/fxa-metrics-processor/src/lib/amplitude/transforms/user-properties.ts
```diff
@@ -118,14 +118,6 @@ function mapNewsletterState(evt: Event, context: EventContext): NewsLetterStateP
...
   }
 
-  if (!newsletterState) {
-    const { marketingOptIn } = context;
-
-    if (marketingOptIn === true || marketingOptIn === false) {
-      newsletterState = marketingOptIn ? 'subscribed' : 'unsubscribed';
-    }
-  }
-
```
chenba, line 128: All `marketingOptIn` related code now removed because https://github.com/mozilla/fxa/pull/4961

packages/fxa-metrics-processor/src/test/lib/amplitude/transforms/common.spec.ts
```diff
@@ -12,13 +12,13 @@ import {
   mapLocation,
   tee,
   sha256Hmac,
```
chenba, line 14: Everything below (if you are viewing the PR in a browser...not VS Code) is either formatting changes or `marketingOptIn` code that's been removed.

packages/fxa-metrics-processor/src/lib/amplitude/index.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * The "identify" property is optional.  However, if it does exist, then the
+ * API request for it MUST be ahead of the request for the http API.
+ *
+ * @param services An object of client-id:service-name mappings
+ * @param plainEvents  An object of name:definition event mappings, where
+ *                     each definition value is itself an object with `group`
+ *                     and `event` string properties.
+ *
+ * @param fuzzyEvents A map of regex:definition event mappings. Each regex
```
vbudhram, line 42: Not related to this PR (since it was already here), but fuzzy events have always been a bit fuzzy to me. After reading the description, I still have a hard time understanding what this property does. 

We could probably link to some docs to provide context on the amplitude taxonomy and what different terms mean.

packages/fxa-metrics-processor/src/lib/amplitude/transforms/amplitude-identify.ts
```diff
@@ -0,0 +1,100 @@
...
+const IDENTIFY_VERBS = ['$set', '$setOnce', '$add', '$append', '$prepend', '$unset'] as const;
+type IdentifyVerb = typeof IDENTIFY_VERBS[number];
+export type AmplitudeIdentifyEvent = {
+  device_id?: AmplitudeHttpEvent['device_id'];
+  event_type: '$identify';
+  user_id: AmplitudeHttpEvent['user_id'];
+  user_properties: IdentifyEventUserProperties;
+};
+type IdentifyEventUserProperties = {
+  [key in Exclude<IdentifyVerb, '$append'>]?: { [key: string]: {} };
```
vbudhram, line 20: Typescript magic 🔥

packages/fxa-metrics-processor/src/lib/amplitude/transforms/amplitude-identify.ts
```diff
@@ -0,0 +1,100 @@
...
+          ...userPreferences,
+        },
+      },
+    };
+  }
+
+  return null;
+}
+
+function mapServicesUsed(servicePropMapper: ServiceNameAndClientIdMapper, context: EventContext) {
```
vbudhram, line 64: General typescript question, function above specifies null and AmplitudeIdentifyEvent as a return type and others don't specify anything even though they do return null. Is there a standard or best practice for this?

chenba: Off hand I don't have a link or something to share for this.  fxa-metrics-processor has a pretty strict `tsconfig.json`, which among other things, does not allow implicit `any`s.  So as far as tsc doesn't complain about the inferred types, I think it's ok.

The other reasons I didn't type them are that they are not exported, they are very simple types, and iirc I wouldn't have used them anywhere else aside as the return types.

PR #5111, opened 2020-04-24
fix(sentry): Fix sentry reporting in customs server

### LINKED ISSUE #5110: Sentry does not appear to be working in customs server
This was found after seeing elevated sentry [errors](https://sentry.prod.mozaws.net/operations/auth-prod/issues/6697641/?query=is:unresolved) in auth-server on port 7000, which is the customs server. In sentry, the customs server project does not show any errors.

I'm not sure when exactly this error started since sentry hasn't had entries in a while, but we should fix the reporting.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1766)
┆Issue Number: FXA-1766

vbudhram:
Fixes #5110

### DIFF
--- a/packages/fxa-customs-server/lib/sentry.js
+++ b/packages/fxa-customs-server/lib/sentry.js
@@ -1,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const Hoek = require('@hapi/hoek');
+const Sentry = require('@sentry/node');
+const version = require('../package.json').version;
+
+async function configureSentry(server, config, log) {
+  const sentryDsn = config.sentryDsn;
+  if (sentryDsn) {
+    Sentry.init({
+      dsn: sentryDsn,
+      release: version,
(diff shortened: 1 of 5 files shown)


packages/fxa-customs-server/lib/server.js
```diff
@@ -105,6 +105,14 @@ module.exports = async function createServer(config, log) {
     host: config.listen.host,
   });
 
+  api.events.on({ name: 'request', channels: 'error' }, (request, event) => {
```
bbangert, line 108: It would be good to add the request context here. Sentry has a few utilities to extract it, I believe I used some of those in the auth-server to extract details from the request object.

packages/fxa-customs-server/lib/sentry.js
```diff
@@ -0,0 +1,73 @@
...
+      if (err && err.stack) {
+        try {
+          exception = err.stack.split('\n')[0];
+        } catch (e) {
+          // ignore bad stack frames
+        }
+      }
+      Sentry.withScope(scope => {
+        scope.addEventProcessor(_sentryEvent => {
+          const sentryEvent = Sentry.Handlers.parseRequest(
```
bbangert, line 54: This was the primary thing I had meant should have been included, didn't mean you had to copy the whole thing over! Sorry.

PR #5116, opened 2020-04-24
More Release 168.1 Uplift

chenba:


chenba:
I'm cherry-picking a single commit that was approved in https://github.com/mozilla/fxa/pull/5114 so I think we can skip the review here.

Issue #5017, opened 2020-04-20
[metrics] Setup Sentry


data-sync-user:
Sentry errors bubble up to `AppErrorBoundary`. We should be able to reference the payments-server setup for this.

Can this become a shared component within `fxa-components`?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1693)
┆Issue Number: FXA-1693

jodyheavener:
Reference https://github.com/mozilla/fxa/pull/5129 for earlier AppErrorBoundary and Sentry entrypoint setup.

PR #5129, opened 2020-04-27
feat(settings): set up some foundational configs and components

### LINKED ISSUE #4821: [setup] Initial index.html, AppLayout, and AppErrorBoundary stubout
Reference `fxa-admin-panel` and `fxa-payments-server` and get these setup in `fxa-settings` similarly.

This is probably a good spot to cleanup any leftover cruft from `create-react-app` (like images/favicon).

This could also be paired with https://jira.mozilla.com/browse/FXA-1541.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1540)
┆Issue Number: FXA-1540

### LINKED ISSUE #4826: [setup] Use some form of `normalize.css`, autoprefixer, svg-url-loader, CSS linting
Make sure we've got some form of reset included in our CSS (normalize.css, `modern-normalize`, `sanitize.css` etc.) to ensure browsers render elements consistently.

Also make sure our CSS compiles with browser prefixes. We may need to setup the `autoprefixer` package.

Also set up `svg-url-loader` for our SVGs.

Also setup CSS/SCSS linting. Lets have this run on checkins/commits. (Should we have this fail at the CI level?) Related: https://github.com/mozilla/fxa/issues/2088

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1545)
┆Issue Number: FXA-1545

jodyheavener:
Closes #4821
Closes #4826

This PR introduces a world of foundational Settings stuff, including:

- Clean up our base `index.html`
- Add a Firefox favicon
- Add `modern-normalize` for CSS normalization
- Tweak eslint config, add sass-lint and its config, add prettier config, add `lint` scripts to package.json
- Set up extremely basic `App` and `AppLayout`
- Set up `AppErrorBoundary` and `AppErrorDialog`. Sentry is **not** set up yet, though.
- Made @fxa-shared available

Some notes:

- To address #4826, CRA comes with autoprefixer (see browsersList in package.json), and automatically handles SVG usage to the extent we need it (refer to the Readme).
- I went with [modern-normalize](https://github.com/sindresorhus/modern-normalize) over [normalize.css](https://github.com/necolas/normalize.css) because it's slightly more lean, targeted at newer browsers, and appears to be more actively maintained (even though it's based _off_ of normalize.css).
- I found the meta tag `apple-itunes-app` from our existing Content Server, figured it will eventually be needed.
- `AppErrorDialog` does not localize the error text, that comes later.
- Everything is intentionally unstyled.


packages/fxa-settings/src/components/SVGExample/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import FFLogo from './ff-logo.svg';
+
+export const SVGExample = () => (
+  <div>
+    <FFLogo data-testid="logo" />
+  </div>
```
LZoog, line 11: This could be a beautiful one liner:
```
export const SVGExample = () => <FFLogo data-testid="logo" />;
```

Will we remove this example when we have a "real" SVG to use in the project?

packages/fxa-settings/src/index.scss
```diff
@@ -0,0 +1,7 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import 'modern-normalize';
```
LZoog, line 5: 🎊

packages/fxa-settings/src/index.tsx
```diff
@@ -1,11 +1,34 @@
...
+import React from 'react';
+import { render } from 'react-dom';
+import './index.scss';
+import App from './components/App';
+import config from './lib/config';
+import sentryMetrics from './lib/sentry';
+
+async function init() {
+  if (config.sentry.dsn) {
+    sentryMetrics.configure(config.sentry.dsn, config.version);
```
LZoog, line 14: Does this also fix #5017 then? Part of that one was checking on if `AppErrorBoundary` could/should be a shared component (hence the est of 3), I know we said for now let's not do that, but I also partially told you that because I figured we could look into it when we actually set up Sentry, which I didn't know you'd be doing in this PR. 😂 Is it a headache to pass in the instance / is it better separated or do you think we should comment on #5017 and say it's done except for looking further into sharing?

packages/fxa-settings/src/lib/logger.js
```diff
@@ -0,0 +1,78 @@
...
+        this._window.console.error(errorMessage);
+      } else {
+        // Use regular console.error
+        this._window.console.error.apply(this._window.console, arguments);
+      }
+    }
+  },
+};
+
+export default Logger;
```
LZoog, line 78: This one might be a good option for sharing? Or at the very least, more typescript-y.

packages/fxa-settings/src/lib/logger.test.js
```diff
@@ -0,0 +1,162 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { assert } from 'chai';
+import sinon from 'sinon';
+import Logger from './logger';
+
+describe('lib/logger', function() {
+  var logger;
```
LZoog, line 10: Is there any reason these are `var`s instead of `let`s that you know of? (I know it was probably just a copy and paste)

packages/fxa-settings/src/lib/sentry.js
```diff
@@ -0,0 +1,162 @@
...
+ * function that gets called before data gets sent to error metrics
+ *
+ * @param {Object} data
+ *  Error object data
+ * @returns {Object} data
+ *  Modified error object data
+ * @private
+ */
+function beforeSend(data) {
+  if (data && data.request) {
```
LZoog, line 32: If possible this file should also be TS... `data?.request` and other places in this file would be awesome

packages/fxa-settings/src/lib/sentry.test.js
```diff
@@ -0,0 +1,194 @@
...
+      assert.equal(
+        resultData.fingerprint[0],
+        'errno100',
+        'correct fingerprint'
+      );
+      assert.equal(resultData.level, 'info', 'correct known error level');
+    });
+
+    it('properly erases sensitive information from url', function() {
+      var url = 'https://accounts.firefox.com/complete_reset_password';
```
LZoog, line 64: More `var`s🤔

packages/fxa-settings/src/components/SVGExample/index.tsx
jodyheavener: We could. I'm fine with either. Doesn't hurt to test that the SVG loader is working.

packages/fxa-settings/.rescriptsrc.js
```diff
@@ -1,11 +1,35 @@
...
+function addSvgLoader(config) {
+  config.module.rules.push({
+    test: /\.svg$/,
+    include: resolve('./src'),
+    use: [
+      {
+        loader: 'babel-loader',
+      },
+      {
+        loader: 'react-svg-loader',
```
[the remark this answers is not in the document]
jodyheavener: Hm. No, this wouldn't work for a `background-image` URL, as this loads the actual SVG, and not any sort of encoded version or URL. I can switch to that style of loader, but at some point I'd like to be able to load actual SVG files in (and not just a URL).

Regarding, aria-label, this approach allows you to supply custom props, so you could set an aria-label, but I don't think that should be a concern of the loader, and is more just a thing we need to remember to do. Maybe we can lint it.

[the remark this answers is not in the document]
LZoog: > is more just a thing we need to remember to do

Yeah, this is what I hope to avoid, because we will forget. 😂 Images without alt tags will complain in the console but these won't so it's a bit trickier. I'm just wondering if there's a way to require it through TS or yeah maybe some linting option. I'm still wondering if `svg-url-loader` would be the better option though since it would also inline our `background-image`s (and we will have at least a few) and we could use `img` and not forget `alt`. What's the benefit of importing them as React components if we can set `width`/`height` properties in CSS on an image? FWIW I like the idea of `react-svg-loader` better but wondering if the other would serve our purposes a bit better?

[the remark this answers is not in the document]
jodyheavener: > What's the benefit of importing them as React components if we can set width/height properties in CSS on an image?

When you use an SVG as an image src you lose the ability to manipulate the SVG itself, via JS or CSS. For example, if we wanted to set the fill or stroke of a button icon on hover, or maybe we want apply an animation to one of the groups in an SVG, or maybe we'd like to target an SVG node with JavaScript to dynamically update it's path.

_That being said..._ I don't envision a lot of that happening with this project. Maybe the camera icon for the cropper, but that's probably it. I'll change it to a URL/encoded loader, and we can address stuff as it comes up.

packages/fxa-settings/public/index.html
```diff
@@ -1,19 +1,23 @@
...
+
+<head>
+  <meta charset="utf-8" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <title>Firefox Accounts</title>
+  <meta name="description" content="" />
+  <meta name="referrer" content="origin" />
+  <meta name="robots" content="noindex,nofollow" />
+  <meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes" />
+  <meta name="fxa-config" content="__SERVER_CONFIG__" />
```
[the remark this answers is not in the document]
jodyheavener: I'll remove for now

packages/fxa-settings/.rescriptsrc.js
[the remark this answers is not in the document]
jodyheavener: Good news: CRA handles **all of the SVG loading**. I'll reflect this in the Readme with examples :)

packages/fxa-settings/src/lib/logger.js
jodyheavener: Typescript'd and shared!

packages/fxa-settings/src/index.tsx
jodyheavener: Check out my latest push - I have sharing set up I _think_.

packages/fxa-settings/src/components/AppLayout/index.tsx
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import './index.scss';
+
+type AppLayoutProps = {
+  children: any;
```
LZoog, line 9: I think this can be `React.ReactNode` instead of `any`.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,192 @@
...
+ *  Error object data
+ * @returns {Object} data
+ *  Modified error object data
+ * @private
+ */
+function beforeSend(data: Sentry.Event): Sentry.Event {
+  if (data && data.request) {
+    if (data.request.url) {
+      data.request.url = cleanUpQueryParam(data.request.url);
+    }
```
LZoog, line 52: I think this can just be in one `if` statement: `if (data?.request?.url) {`

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (data.tags) {
+      // if this is a known errno, then use grouping with fingerprints
+      // Docs: https://docs.sentry.io/hosted/learn/rollups/#fallback-grouping
+      if (data.tags.errno) {
```
LZoog, line 57: `if (data.tags?.errno)`

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,192 @@
...
+        data.fingerprint = ['errno' + data.tags.errno];
+        // if it is a known error change the error level to info.
+        data.level = Sentry.Severity.Info;
+      }
+    }
+
+    if (data.exception && data.exception.values) {
```
LZoog, line 64: Same here on line 64 and 66 on the `?` check.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,192 @@
...
+          value.stacktrace.frames.forEach((frame: { abs_path?: string }) => {
+            if (frame.abs_path) {
+              frame.abs_path = cleanUpQueryParam(frame.abs_path); // eslint-disable-line camelcase
+            }
+          });
+        }
+      });
+    }
+
+    if (data.request.headers && data.request.headers.Referer) {
```
LZoog, line 76: `data.request.headers?.Referer`

packages/fxa-shared/lib/logger.ts
```diff
@@ -0,0 +1,86 @@
...
+ * interpolation before logging message.
+ *
+ */
+
+interface Logger {
+  _window: Window;
+  info: (...args: any) => void;
+  trace: (...args: any) => void;
+  warn: (...args: any) => void;
+  error: (...args: any) => void;
```
LZoog, line 16: This feels like it could be something other than `any`... 🤔

packages/fxa-settings/src/lib/config.ts
```diff
@@ -0,0 +1,27 @@
...
+  env: 'development',
+  featureFlags: {},
+  lang: '',
+  sentry: {
+    url: 'https://sentry.prod.mozaws.net',
+    dsn: '',
+  },
+};
+
+export default config;
```
LZoog, line 27: You could remove this file as well since we're punting that config issue out.

packages/fxa-settings/README.md
```diff
@@ -11,7 +11,26 @@ You can import React components into this project. This is currently restricted
...
+const LogoImage = () => <img src={Logo} alt="Logo" />;
+
+// As a background-image (inline style)
+import Logo from './logo.svg';
+const LogoImage = () => <div style={{ backgroundImage: `url(${Logo})` }}></div>;
+
+// As a background-image (external style)
+// Just reference it in CSS, the loader will find it
+// .logo { background-image: url('logo.svg'); }
+const LogoImage = () => <div class="logo"></div>;
```
LZoog, line 33: Oh, one more comment here, we should probably have `role="img"` and `aria-label="logo"` on line 33 just for best practice. We could have a note on the first option being the preferred method too, the second option starts to mix CSS-in-JS but we could do that if we needed to for some reason.

jodyheavener:
@LZoog per our discussion about data-uri-inlining SVGs, I spent a bit more time on it and posted on Stack Overflow, both with no luck. Considering the implementation would be the same, and all that we'd need to do is set up the loader, I think we should try to take a stab at this down the line as a wishlist item.

LZoog:
@jodyheavener that's fine, I'll create it as a follow up issue on Settings v2, I think this PR is good to merge👍

PR #5130, opened 2020-04-27
fix(chrome): fixed webcrypto check on Chrome

dannycoates:
Chrome was loading the webcrypto polyfill even though
it doesn't need it because it threw an error about the
input key length on HKDF and HMAC.


bbangert:
I'm not sure who would be qualified to verify this is correct, maybe one of our security people?

dannycoates:
The `checkWebCrypto` function determines if the browser supports the algorithms we use by trying to use them. If any of them fail we load a polyfill. Chrome is a bit more strict than other browsers about validating inputs. This changes the key lengths in the check to the length we actually use in the rest of the code.

PR #5139, opened 2020-04-28
fix(geodb): make geodb accuracy logging optional

jrgm:
Are we actually using these lines for analysis? If not, they are a large proportion of our logging costs for auth server. @davismtl @vbudhram might recall.

r? - @mozilla/fxa-devs

### DIFF
--- a/packages/fxa-auth-server/lib/geodb.js
+++ b/packages/fxa-auth-server/lib/geodb.js
@@ -42,10 +42,12 @@ module.exports = log => {
         confidence += 'no_accuracy_data';
       }
 
-      log.info('geodb.accuracy', { accuracy });
-      log.info('geodb.accuracy_confidence', {
-        accuracy_confidence: confidence,
-      });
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/geodb.js
```diff
@@ -42,10 +42,12 @@ module.exports = log => {
         confidence += 'no_accuracy_data';
       }
 
-      log.info('geodb.accuracy', { accuracy });
-      log.info('geodb.accuracy_confidence', {
-        accuracy_confidence: confidence,
-      });
+      if (config.logAccuracy) {
+        log.info('geodb.accuracy', { accuracy });
```
vbudhram, line 46: This was used to create a kibana dashboard which we don't have access anymore. Welcome to rip out the log statements too.

jrgm: Okay. I'll leave the lines as an option, but defaulted to off.

PR #5144, opened 2020-04-29
docs(adrs): CSS part 1 - Use CSS variables, prefer SCSS over css-in-js solutions

LZoog:


docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,181 @@
...
+
+Technical Story: [Github issue](https://github.com/mozilla/fxa/issues/4808)
+
+## Context and Problem Statement
+
+The [Settings Redesign](https://github.com/mozilla/fxa/issues/3740) project offers an ideal opportunity to review CSS conventions and architecture for FxA to use not only while building new components for this project but also for FxA moving forward.
+
+This [part 1](https://github.com/mozilla/fxa/issues/4808) of [2](https://github.com/mozilla/fxa/issues/5087) ADR concerns the use of CSS variables, how they come into play with SCSS (vs using SCSS variables), CSS-in-JS considerations, and CSS modules.
+
+Note that 1) "CSS variables" are technically called "CSS custom properties" but will be referred to as the common "variable" nomenclature throughout this ADR and 2) SASS is a CSS preprocessor while SCSS is our preferred syntax of SASS.
```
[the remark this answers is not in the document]
jodyheavener: (I can make this change and push if you agree)

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Styling placement
+
+- CSS-in-JS with a helper lib like [`styled-components`](https://styled-components.com/) or [`react-jss`](https://cssinjs.org/react-jss/)
+- CSS modules
+- SCSS
+
+## Decision Outcome
+
+Chosen options: "CSS variables" and "SCSS", because:
```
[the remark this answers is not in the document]
LZoog: Ooh look at that, I actually didn't know that we were using extends in FxA but we are in one or two spots. I'll add extends here, those should be no problem.

Maybe this is a better phrasing for line 39:
> While using CSS variables with SASS will prevent using mixins (or SASS functions like `darken`) that rely on variable values, they can be used without impeding core functionality FxA uses SASS for like nested selectors, extends, and functions and mixins that don't rely on variable values like generation and use of our media queries.

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,181 @@
...
+
+- CSS variables are native to the browser; they have good browser support and can be used and set without a preprocessor and after preprocessor compilation. They are targetable by JavaScript, easily themed, and can be scoped globally or on an element, providing options for usage.
+- CSS variables can be used in SCSS without impeding core functionality SASS offers like nested selectors plus functions and mixins that don't rely on variable values.
+- Scoped component solutions (CSS-in-JS, CSS modules) encourage self-contained components rather than building the UI as a whole. While FxA will have some component reuse, FxA will likely reap more benefits from a class based approach with shared styles as a much DRYer and consistent approach.
```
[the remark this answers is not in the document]
LZoog: Good call out. This also reminds me that by using classes, it may actually make some sharing cleaner because we won't have a bunch of if/elses in the styles based on props. Will update.

docs/adr/0015-use-css-variables-and-scss.md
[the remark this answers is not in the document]
jodyheavener: Devil's advocate: do we want to encourage `@extend` usage? If used appropriately it can be super beneficial, but I've sometimes found it can become a "junk-drawer" and the resulting selector ends up being this mammoth.

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,185 @@
...
+}
+
+.special h2 {
+  --h2-color: black;
+}
+```
+
+Neutral:
+
+- While CSS variables meet the minimum browser compatibility list for FxA and have [great support otherwise](https://caniuse.com/#feat=css-variables), they aren't supported in IE11.
```
meandavejustice, line 92: This should be in cons section

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,185 @@
...
+- CSS variables can be seen in developer tools. If there are many scoped to `:root` (globally) or the component in question, it can look like a lot of cruft, but this can also be convenient depending on what is being inspected.
```
meandavejustice, line 93: This is a good opportunity to discuss better options for displaying these with the devtools team. Also devtools is written as react components, so it's easier to experiment with

docs/adr/0015-use-css-variables-and-scss.md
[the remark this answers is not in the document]
LZoog: Added.

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,185 @@
...
+}
+
+.special h2 {
+  --h2-color: black;
+}
+```
+
+Neutral:
+
+- While CSS variables meet the minimum browser compatibility list for FxA and have [great support otherwise](https://caniuse.com/#feat=css-variables), they aren't supported in IE11.
```
chenba, line 92: We don't officially support IE11 so that's not a real concern.

docs/adr/0015-use-css-variables-and-scss.md
```diff
@@ -0,0 +1,185 @@
...
+- SASS variables are imperative and can only hold one value at a time so if the variable changes in value, earlier references will remain the same.
+- We'd need to add to our `rescripts` webpack custom config to expose global SASS variables to all of our files.
+
+Cons:
+
+- SASS variables offer a limited set of "pre-compilation" power because variables can't be set dynamically after the stylesheet has been loaded.
+
+#### CSS variables
+
+[CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties), or CSS variables, provide a powerful native browser solution to setting CSS properties to variables. They are declared as `--color: grey` and used as `color: var(--font-color)` and are present in outputted CSS after SASS compilation.
```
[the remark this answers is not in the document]
LZoog: Ha, total miss on my part, good catch.

docs/adr/0015-use-css-variables-and-scss.md
LZoog: Does this mean we can just nuke the need for a polyfill? I can move this to the pros section if so since they'll just work out of the box without anything else on our end.

chenba: Heh, I can only tell you what's in https://github.com/mozilla/fxa/blob/master/docs/supported-browsers.md.  Not sure what the implications are for CSS in new Settings.  But yeah, if we are going to do something _only_ for IE11, then we should not.

jodyheavener: I'm fine with it still being noted here, though -- no harm with it being thought about :)

docs/adr/0015-use-css-variables-and-scss.md
jodyheavener: I think we might be fine leaving it as neutral or, if the above discussion wants, eliminated entirely because we don't really need IE11 support. That cool?

docs/adr/0015-use-css-variables-and-scss.md
jodyheavener: That's a great point. Forgot how closely tied we are to them.

PR #5148, opened 2020-04-29
fix(experiments): Enabled variation B 100% for sms header experiment

vbudhram:
This PR enables variation B `Would you like to sync your phone?` for 100% of users. 

This should be uplifted as a point release on latest train, cc @chenba


vbudhram:
> will that be done later in a follow up?

Yea it will be, this is just easier than ripping out all the code.

PR #5157, opened 2020-04-29
feat(gql-api): add auth to graphql-api server

### LINKED ISSUE #5092: Graphql API authentication handler
The graphql-api server needs to have an authentication handler that looks up OAuth2 tokens to verify the user and extracts the token details.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1753)
┆Issue Number: FXA-1753

chenba:
This patch enables graphql-api server to accept an OAuth2 access token
on the graphql route.

It:
  1. validates the format of the authorization header value in middleware
  2. looks up the token in Redis
  3. ensures the client id of the token is in the allowed clients list
  4. includes the user's id and email in the 'authUser' property in
     Apollo's context
  5. includes the authorization header value as 'token' in Apollo's
     context

Fixes #5092 (FXA-1753)

### DIFF
--- a/packages/fxa-graphql-api/pm2.config.js
+++ b/packages/fxa-graphql-api/pm2.config.js
@@ -12,7 +12,7 @@ module.exports = {
       min_uptime: '2m',
       env: {
         NODE_ENV: 'development',
-        NODE_OPTIONS: '--inspect=9190',
+        NODE_OPTIONS: '--inspect=9200',
         TS_NODE_TRANSPILE_ONLY: 'true',
         TS_NODE_FILES: 'true',
         PORT: '8290', // TODO: this needs to get added to src/config.ts
--- a/packages/fxa-graphql-api/src/lib/server.ts
+++ b/packages/fxa-graphql-api/src/lib/server.ts
@@ -2,14 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { ApolloServer } from 'apollo-server-express';
+import { Container } from 'typedi';
+import { ApolloServer, AuthenticationError } from 'apollo-server-express';
 import { Logger } from 'mozlog';
 import * as TypeGraphQL from 'type-graphql';
-import { Container } from 'typedi';
 
 import { DatabaseConfig, setupDatabase } from './db';
 import { AccountResolver } from './resolvers/account-resolver';
 import { reportGraphQLError } from './sentry';
+import { userLookupFnContainerToken } from './constants';
 
 type ServerConfig = {
   authHeader: string;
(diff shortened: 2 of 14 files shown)


packages/fxa-graphql-api/pm2.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
       min_uptime: '2m',
       env: {
         NODE_ENV: 'development',
-        NODE_OPTIONS: '--inspect=9190',
+        NODE_OPTIONS: '--inspect=9290',
```
chenba, line 15: This was conflicting with fxa-admin-server.

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -31,20 +32,24 @@ export async function createServer(
   logger: Logger,
   context?: () => object
 ): Promise<ApolloServer> {
-  setupDatabase(config.database);
```
chenba, line 34: We are not going to access the database from this server, for now at least.

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -31,20 +32,24 @@ export async function createServer(
...
+    if (!authUser) throw new AuthenticationError('You must be logged in');
+
     return {
       authUser,
       logAction: (action: string, options?: object): void => {
-        logger.info(action, { authUser, ...options });
+        logger.info(action, { uid: authUser.userId, ...options });
       },
       logger,
+      token: bearerToken,
```
chenba, line 52: Copied over because we'll need to pass it onto the auth-server to access its API.

bbangert: Feels fine for now, Apollo has [some tooling ](https://www.apollographql.com/docs/apollo-server/data/data-sources/#rest-data-source)to make it easier to link portions of a graph object to remote API calls, so we'll likely need to plumb that through there later.

packages/fxa-graphql-api/src/test/lib/oauth/index.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Request, Response } from 'express';
+import { assert } from 'chai';
+import 'mocha';
+import sinon from 'sinon';
+
+import { loggerContainerKey, configContainerKey } from '../../../lib/constants';
+
+const sandbox = sinon.createSandbox();
+// tslint:disable-next-line: no-empty
+const mockLogger = { debug: () => {} };
```
bbangert, line 15: Would it be helpful to make this a sinon spy/stub so we can assert things on it?

packages/fxa-graphql-api/src/test/lib/oauth/index.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+Container.set(configContainerKey, mockConfig);
+
+import { getTokenId, oauthBearerTokenValidator } from '../../../lib/oauth';
+import { AppConfig } from '../../../config';
+
+describe('oauthBearerTokenValidator', () => {
+  const sendStub = sandbox.stub();
+  const statusStub = sandbox.stub().returns({ send: sendStub });
+  const res = ({ status: statusStub } as unknown) as Response;
```
bbangert, line 34: I know why this is here, and its one of the few things in TypeScript that still annoys me. 😬

packages/fxa-graphql-api/pm2.config.js
bbangert: Note, the README in the fxa/ root, not in this package.

packages/fxa-graphql-api/src/lib/server.ts
chenba: Ah, ok.  I will revert.

packages/fxa-graphql-api/src/lib/server.ts
chenba: Hmm the doc's examples are using `this.context.token`, which is this I believe.

PR #5160, opened 2020-04-30
fix(auth server): reference fxa-shared types in tsconfig

### LINKED ISSUE #5150: Error building docker image
From https://164938-36831081-gh.circle-artifacts.com/0/artifacts/fxa-builder.log

```
#10 573.0 fxa-auth-server: ../fxa-shared/l10n/localizeTimestamp.ts(5,33): error TS7016: Could not find a declaration file for module 'accept-language'. '/fxa/packages/fxa-shared/node_modules/accept-language/index.js' implicitly has an 'any' type.
#10 573.0 fxa-auth-server:   Try `npm install @types/accept-language` if it exists or add a new declaration (.d.ts) file containing `declare module 'accept-language';`
```

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1785)
┆Issue Number: FXA-1785

jodyheavener:
Closes #5150

This prevented CI from building properly.

```
Could not find a declaration file for module 'accept-language'. '/fxa/packages/fxa-shared/node_modules/accept-language/index.js' implicitly has an 'any' type.
Try `npm install @types/accept-language` if it exists or add a new declaration (.d.ts) file containing `declare module 'accept-language';`
```

@chenba I know your original comment was to add the type declaration file to fxa-auth-server, but I figured since the issue stemmed from a module referenced in a file from fxa-shared that we should have a single source of truth/keep it dry.

### DIFF
--- a/packages/fxa-auth-server/tsconfig.json
+++ b/packages/fxa-auth-server/tsconfig.json
@@ -14,11 +14,9 @@
     "allowSyntheticDefaultImports": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "types": [
-      "node"
-    ],
     "typeRoots": [
       "./types",
+      "../fxa-shared/types",
       "node_modules/@types"
     ]
   },


packages/fxa-auth-server/tsconfig.json
```diff
@@ -14,11 +14,9 @@
     "allowSyntheticDefaultImports": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "types": [
-      "node"
-    ],
     "typeRoots": [
       "./types",
+      "../fxa-shared/types",
```
bbangert, line 19: If at some point in the future, we want typeRoots to diverge based on building for deploy vs. running locally, we can make another tsconfig.prod.json that overrides just the bits to change when we build for it.

PR #5161, opened 2020-04-30
refactor(auth server): typescript lib/oauth/metrics

### LINKED ISSUE #4879: [auth-server] Add Typescript to files in /lib/oauth/metrics
This is part of a larger effort to [adopt Typescript in the Auth Server](https://jira.mozilla.com/browse/FXA-1500). It can be broken down into smaller issues if desired. This needs to be completed before [FXA-1589 (lib/oauth)](https://jira.mozilla.com/browse/FXA-1589) can be closed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1587)
┆Issue Number: FXA-1587

jodyheavener:
Closes #4879 

_Part of a larger effort to [adopt TypeScript in the Auth Server](https://jira.mozilla.com/browse/FXA-1500)._

---

I'm trying to take a slow, educational approach to TypeScript, so please feel free to tear this apart.

### DIFF
--- a/packages/fxa-auth-server/README.md
+++ b/packages/fxa-auth-server/README.md
@@ -129,7 +129,7 @@ Refer to https://github.com/mozilla/fxa-dev.git.
 Configuration of this project
 is managed by [convict](https://github.com/mozilla/node-convict),
 using the schema in
-[`config/index.js`](config/index.js).
+[`config/index.ts`](config/index.ts).
 
 Default values from this schema
 can be overridden in two ways:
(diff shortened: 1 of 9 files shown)


packages/fxa-auth-server/types/config/index.d.ts
```diff
@@ -0,0 +1,11 @@
+// This interface should be expanded on
+// as we slowly move over to TypeScript
```
chenba, line 2: It actually might be easier just change the config file to `.ts`, install `@types/convict`, and export `typeof conf` in the config ts.

jodyheavener: So many implications for changing the config file to typescript 😆

chenba:
If you add "[skip ci]" to your commit message of a wip commit, CircleCI will skip CI.

jodyheavener:
@chenba fair, yeah, this config change has just been so far reaching that I'm using CI to find all the spots 😅

jodyheavener:
@chenba Okay, so we're passing now, but you'll notice I had to up certain timeouts an quite a bit.

I'm guessing it has to do with mocha using TypeScript and needing to compile everything repeatedly. Or perhaps it's the use `npx` (because it will install ts-node if it is not present), but because we `npm i` before the test I wouldn't suspect it's that. At any rate, I don't feel great about the timeouts. If you have any thoughts I'm all ears.

PR #5174, opened 2020-04-30
fix(customs): Update the route timeout level for customs server

vbudhram:
This missed the train, but it would be nice to get on if we do a point release. It updates the connection timeout for customs server to reduce the socket hangups that occur from auth-server.

cc @jrgm @bbangert

### DIFF
--- a/packages/fxa-customs-server/lib/server.js
+++ b/packages/fxa-customs-server/lib/server.js
@@ -93,7 +93,16 @@ module.exports = async function createServer(config, log) {
   const api = Hapi.server({
     port: config.listen.port,
     host: config.listen.host,
+    // To reduce connection timeouts, increase the server and socket timeouts
+    routes: { timeout: { server: 4 * 60 * 1000, socket: 8 * 60 * 1000 } },
   });
+
+  // Allow Keep-Alive connections from the auth-server to be idle up to two
+  // minutes before closing the connection. If this is not set, the default
+  // idle-time is 5 seconds.  This can cause a lot of unneeded churn in server
+  // connections. Setting this to 120s makes node8 behave more like node6. -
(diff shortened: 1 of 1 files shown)


packages/fxa-customs-server/lib/server.js
```diff
@@ -93,6 +93,8 @@ module.exports = async function createServer(config, log) {
   const api = Hapi.server({
     port: config.listen.port,
     host: config.listen.host,
+    // Matches the original idle timeout of the customs-server
+    routes: { timeout: { socket: 120000 } },
```
jrgm, line 97: Hrm. https://hapi.dev/api/?v=19.1.1#-routeoptionstimeoutsocket sez the default is already 120000 (2min).

I'm thinking though that this is the wrong setting, not that I know where the right setting is.

packages/fxa-customs-server/lib/server.js
```diff
@@ -93,6 +93,8 @@ module.exports = async function createServer(config, log) {
   const api = Hapi.server({
     port: config.listen.port,
     host: config.listen.host,
+    // Matches the original idle timeout (120s) of the customs-server
+    routes: { timeout: { server: 120 * 60 * 1000 } },
```
jrgm, line 97: 120 minutes?

packages/fxa-customs-server/lib/server.js
vbudhram: > not that I know where the right setting is.

Heh, I'm not 100% sure either, but I think we can try this?

jrgm:
The other timeout changes are fine, but when we last fixed this, it was the HTTP keep-alive that fixed these socket hangups.

PR #5181, opened 2020-04-30
refactor(fxa-circleci): moved fxa-circleci image source

dannycoates:
Including fxa-circleci in the /packages folder was awkward
since it isn't part of the deployment and needed some special
handling in the test/build process.

This attempts to make that better.


_dev/docker/circleci/Dockerfile
```diff
@@ -2,30 +2,18 @@ FROM circleci/node:12-stretch-browsers
...
 RUN apt-get update && apt-get install -y \
-    bison \
-    build-essential \
-    cmake \
-    flex \
     graphicsmagick \
-    libboost-dev \
-    libboost-filesystem-dev \
-    libboost-regex-dev \
-    libboost-system-dev \
```
dannycoates, line 13: I remove the deps for `fxa-amplitude-send` since it isn't in the normal test flow and these are quite large deps

_dev/docker/circleci/Dockerfile
```diff
@@ -2,30 +2,18 @@ FROM circleci/node:12-stretch-browsers
...
 RUN pip install mozdownload mozinstall
 RUN mozdownload --version 68.0 --destination firefox.tar.bz2
 RUN wget https://s3-us-west-2.amazonaws.com/fxa-dev-bucket/fenix-pair/desktop/7f10c7614e9fa46-target.tar.bz2
 
 USER circleci
 
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
-ENV PATH=/home/circleci/.cargo/bin:$PATH
-RUN cargo install cargo-audit
-
```
dannycoates, line 28: I removed rust for the same reason, however i now notice `test-email-service` is using this image. I'll see if changing it to a ready-made rust image will work instead

dannycoates: `circleci/rust:latest-node` works for `test-email-service` 👍

PR #5182, opened 2020-04-30
fix(l10n): fix primary email update message l10n

### LINKED ISSUE #5136: "Primary email set to %(email)s" is not localized
It looks like the first string in #3258 did not get addressed in #3319, I’ve just seen it popping n English. Filing this issue to get it back in the backlog

Capture d’écran 2020-04-28 à 18 38 40

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1776)
┆Issue Number: FXA-1776

chenba:
Fixes #5136 (FXA-1776)

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/emails.js
+++ b/packages/fxa-content-server/app/scripts/views/settings/emails.js
@@ -11,7 +11,6 @@ import LastCheckedTimeMixin from '../mixins/last-checked-time-mixin';
 import preventDefaultThen from '../decorators/prevent_default_then';
 import SettingsPanelMixin from '../mixins/settings-panel-mixin';
 import UpgradeSessionMixin from '../mixins/upgrade-session-mixin';
-import Strings from '../../lib/strings';
 import showProgressIndicator from '../decorators/progress_indicator';
 import Template from 'templates/settings/emails.mustache';
 import Url from '../../lib/url';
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/views/settings/emails.js
```diff
@@ -187,7 +186,7 @@ var View = FormView.extend({
       account.unset('originalLoginEmail');
       this.setSearchString('email', email);
       this.displaySuccess(
-        Strings.interpolate(t('Primary email set to %(email)s'), { email }),
+        this.translate(t('Primary email set to %(email)s'), { email }),
```
lmorchard, line 189: Huh. Do we know why this stopped working? The code looks not recently changed here. Also... do we happen to have a test that could have caught this?

chenba: > 
> 
> Huh. Do we know why this stopped working? The code looks not recently changed here. 

It is possible that this never worked.  The user has to a) add a secondary email, b) set the secondary as the primary, c) pay attention the notification message, d) know to expect a localized message, and e) report a bug.  It was originally reported in #3258, but it was not fixed in the PR.

> Also... do we happen to have a test that could have caught this?

Off hand I don't remember how many tests we have for testing specific message l10n.  I'll dig around a bit.

chenba: The view translator actually has a test specific argument (https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/translator.js#L17) so it won't try to fetch translations.

I updated the view test to ensure that the correct function is called with the correct arguments.

PR #5184, opened 2020-05-01
fix: allow fixup script to remove bad settings

### LINKED ISSUE #5183: event-broker: "Resource not found" to topic, but topic exists and has not changed
https://sentry.prod.mozaws.net/operations/fxa-event-broker-prod/issues/6712914/events/latest/

It’s started happening when I ran the script. It’s complaining Resource not found (resource=rpQueue-E6EB0D1E856335FC but I’m looking at that topic in the console, and our terraform config says “nothing to do”. i.e., state of config has not changed in prod

One small difference is that the hex is uppercase in the error message, but the topic is lowercase, like all our usage of oauth client id in our code base. So that difference my not be relevant.

A historical sentry entry from also had uppercase hex for rpQueue-DEADBEEF

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1810)
┆Issue Number: FXA-1810

bbangert:
Because:

* The script was run unexpectedly with an uppercase
  client id, the wrong queue name is tried.

This commit:

* Allows ops to run the script to remove the entries
  made in error.

Closes #5183

### DIFF
--- a/packages/fxa-auth-server/bin/update_firestore_from_oauth.js
+++ b/packages/fxa-auth-server/bin/update_firestore_from_oauth.js
@@ -21,6 +21,7 @@ const log = require('../lib/log')(
 );
 
 const CLIENT_ID = process.env.CLIENT_ID;
+const REMOVE_CLIENT = process.env.REMOVE_CLIENT === 'true';
 
 async function main() {
   if (!CLIENT_ID) {
@@ -50,8 +51,20 @@ async function main() {
     if (doc.exists) {
       const data = doc.data();
       if (data && data.oauth_clients && data.oauth_clients[CLIENT_ID]) {
-        // Record is already in the database
-        continue;
+        if (REMOVE_CLIENT) {
+          // Remove this entry
+          await document.set(
+            {
+              oauth_clients: { [CLIENT_ID]: Firestore.FieldValue.delete() },
+            },
+            { merge: true }
+          );
+          log.info('Removed client id', { uid, clientId: CLIENT_ID });
+          continue;
+        } else {
+          // Record is already in the database
+          continue;
+        }
       }
     }
     log.info('Adding missing uid for client', { uid, clientId: CLIENT_ID });


packages/fxa-auth-server/bin/update_firestore_from_oauth.js
```diff
@@ -21,6 +21,7 @@ const log = require('../lib/log')(
 );
 
 const CLIENT_ID = process.env.CLIENT_ID;
```
jrgm, line 23: Can we add an assertion here that the client has the proper case. Abort on `/[A-Z]/.test(CLIENT_ID)`?

packages/fxa-auth-server/bin/update_firestore_from_oauth.js
```diff
@@ -50,8 +51,20 @@ async function main() {
     if (doc.exists) {
       const data = doc.data();
       if (data && data.oauth_clients && data.oauth_clients[CLIENT_ID]) {
-        // Record is already in the database
-        continue;
+        if (REMOVE_CLIENT) {
+          // Remove this entry
+          await document.set(
+            {
```
jrgm, line 57: So, did you want to test for case, and leave it alone if it is not uppercase?

packages/fxa-auth-server/bin/update_firestore_from_oauth.js
bbangert: I can't right now, or you won't be able to fix the ones with the error.

packages/fxa-auth-server/bin/update_firestore_from_oauth.js
bbangert: We know its already case sensitive, so this test is already also case sensitive. Building more fault detection in feels less useful as once this script is run and we get the appropriate results in prod, its going to be removed. This is a one-off script.

jrgm: Or is the intent to run with the uppercase CLIENT_ID value and REMOVE_CLIENT=true. If it finds a match using the uppercase, then nuke it.

bbangert: Yes, that was the intent. Use the upper-case version with REMOVE_CLIENT.

jrgm: Okay. Thanks.

PR #5189, opened 2020-05-01
chore(subscriptions): make fortress an upgradable sub

johngruen:
Slightly WIP b/c i have to run for the week


packages/fortress/package.json
```diff
@@ -20,19 +20,22 @@
   },
   "private": true,
   "dependencies": {
-    "bower": "*",
     "client-sessions": "0.6.x",
     "express": "4.16.4",
-    "morgan": "1.9.1"
+    "jquery": "^3.5.0",
```
johngruen, line 25: i can remove this

packages/fortress/static/main.css
```diff
@@ -0,0 +1,173 @@
...
+  box-shadow: 0 5px 18px rgba(0, 0, 0, 0.2);
+}
+
+.card.price-card {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.card.price-card > * {
```
LZoog, line 95: This really doesn't matter for this package, and is irrelevant in the grand scheme of performance things, but star selectors are bad CSS selectors performance wise since selectors are processed by the browser from right-to-left. Plus since this is all direct children, if your HTML changes this might do funny things, but again, not a big deal for this package, more of a tip.

packages/fortress/views/download.pug
```diff
@@ -0,0 +1,12 @@
+extends layout
+
+block content
+  h1 #{data.global.title} <span class="highlight">#{data.products[index].title}</span>
```
LZoog, line 4: If you want to be a pug extraordinaire, you can also [set a span this way](https://pugjs.org/language/interpolation.html#tag-interpolation).

packages/fortress/package.json
```diff
@@ -20,19 +20,22 @@
...
+    "morgan": "1.9.1",
+    "pug": "2.0.4"
```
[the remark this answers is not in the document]
johngruen: i tend to use pug a lot b/c it's terse, but I'm happy to stick to a convention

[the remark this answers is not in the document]
johngruen: converted to EJS, officially decided i like pug a lot more...it feels way easier to read at a glance to me.

packages/fortress/package.json
```diff
@@ -41,11 +44,12 @@
...
     "lint:eslint": "eslint .",
-    "postinstall": "bower install --config.interactive=false -s",
     "start": "pm2 start pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "restart": "pm2 restart pm2.config.js",
     "test": "npm run lint",
-    "format": "prettier '**' --write"
+    "format": "prettier '**' --write",
+    "dev": "nodemon server.js",
+    "ui": "browser-sync start --proxy='localhost:9292' --files='**/*.css, **/*.pug, **/*.js' --ignore=node_modules --port=6006 --no-open --serveStatic='static'"
```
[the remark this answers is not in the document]
johngruen: Browser-sync gives u livereloading, nodemon just restarts the server. i updated the readme to reflect this stuff

packages/fortress/views/footer.ejs
```diff
@@ -0,0 +1,3 @@
+<script src="/fortress.js"></script>
+</body>
+</html>
```
LZoog, line 3: I think the more typical approach for EJS projects is to organize files in `/pages` for pages and `/partials` for pieces of pages, but this OK for this app. `footer`/`header` makes me think of components and not layout-related pieces but Googling around shows me a lot of small projects don't tend to make this separation. 🤨

packages/fortress/data/index.js
```diff
@@ -0,0 +1,72 @@
+const data = {
+  global: {
+    title: 'Firefox Fortress',
+    tagline: 'Achieve your fortress dreams',
+    planPicker: 'Select your plan',
+    congrats: 'You are a hero of time and space',
+    paymentsUrl: '//localhost:3030/subscriptions/products',
```
LZoog, line 7: This is fine so long as this doesn't get deployed anywhere which, I don't believe it does?

packages/fortress/views/footer.ejs
johngruen: if only ejs had layouts :(

packages/fortress/data/index.js
johngruen: oh yeah, i was thinking about doing a follow up where i put the fortress images on our CDN...

packages/fortress/views/footer.ejs
lmorchard: Not important for this PR, but we could do something like layouts with an express wrapper like https://www.npmjs.com/package/express-ejs-layouts

(it's basically how we do layouts with handlebars in email senders)

LZoog:
@johngruen, I invested this a bit and there wasn't a `.stylelintrc` present.

It looks like we're ignoring all `.*` files... https://github.com/mozilla/fxa/blob/master/.gitignore#L54 So you probably created it locally and then it wasn't committed. The command works fine for me locally with this file present. I force added it to your branch, hoping this gets CI passing.

PR #5203, opened 2020-05-04
feat(settings): add sentry setup

### LINKED ISSUE #5017: [metrics] Setup Sentry

jodyheavener:
Closes #5017

This PR adds the files for Sentry (and by extension a logger), and sets it up in fxa-settings.

Notes:

- The Sentry and Logging files, as well as their accompanying tests, more or less came from fxa-payments-server, and were converted to TypeScript.
- The Sentry and Logging files are in fxa-shared because they are being used in AppErrorBoundary, which is itself a shared fxa-component. I am open to directory naming suggestions.

### DIFF
--- a/packages/fxa-components/jest.config.js
+++ b/packages/fxa-components/jest.config.js
@@ -1,3 +1,5 @@
+const { resolve } = require("path");
+
 module.exports = {
   roots: ["<rootDir>"],
   transform: {
@@ -5,5 +7,6 @@ module.exports = {
   },
   moduleNameMapper: {
     "\\.(css|scss)$": "identity-obj-proxy",
+    "^@fxa-shared/(.*)$": resolve(__dirname, "../fxa-shared", "$1")
   }
 };
--- a/packages/fxa-components/rescripts.js
+++ b/packages/fxa-components/rescripts.js
@@ -83,9 +83,14 @@ const componentsJestMapper = {
   jest: config => {
     return {
       ...config,
-      moduleNameMapper: {
-        "^@fxa-components/(.*)$": resolve(__dirname, "$1")
-      }
+      moduleNameMapper: Object.keys(additionalJSImports).reduce(
+        (previous, key) => {
+          return Object.assign(previous, {
+            [`^${key}/(.*)$`]: resolve(additionalJSImports[key], "$1")
+          });
+        },
+        {}
+      )
     };
   }
 };
--- a/packages/fxa-settings/src/index.tsx
+++ b/packages/fxa-settings/src/index.tsx
@@ -6,12 +6,11 @@ import React from 'react';
 import { render } from 'react-dom';
 import './index.scss';
 import App from './components/App';
+import sentryMetrics from '@fxa-shared/lib/sentry';
+import config from './lib/config';
 
 export async function init() {
-  // TODO: Configure Sentry logging
-  // if (config.sentry.dsn) {
-  //   sentryMetrics.configure(config.sentry.dsn, config.version);
-  // }
+  sentryMetrics.configure(config.sentry.dsn, config.version);
 
   render(
     <React.StrictMode>
--- a/packages/fxa-shared/lib/logger.ts
+++ b/packages/fxa-shared/lib/logger.ts
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Logging module that performs proper checks and string
+ * interpolation before logging message.
+ *
+ */
+
+interface Logger {
+  _window: Window;
+  info: (...args: any) => void;
+  trace: (...args: any) => void;
+  warn: (...args: any) => void;
+  error: (...args: any) => void;
+}
+
+interface ModuleError {
+  errorModule: {
+    toInterpolatedMessage: (arg0: ModuleError) => string;
+  };
+}
+
+/**
+ * Constructor of log module.
+ *
+ * @param {Object} logWindow Window object that contains console used for logging.
+ *
+ * @constructor
+ */
+const Logger = (function(this: Logger, logWindow?: Window) {
+  this._window = logWindow || window;
+} as any) as new (logWindow?: Window) => Logger;
+
+Logger.prototype = {
+  /**
+   * Wrapper for `console.info` function that checks for availability and
+   * then prints info.
+   *
+   */
+  info() {
+    this._window.console?.info?.apply(this._window.console, arguments);
+  },
+
+  /**
+   * Wrapper for `console.trace` function that checks for availability and then prints
+   * trace.
+   *
+   */
+  trace() {
+    this._window.console?.trace?.apply(this._window.console, arguments);
+  },
+
+  /**
+   * Wrapper for `console.warn` function that checks for availability and
+   * then prints warn.
+   *
+   */
+  warn() {
+    this._window.console?.warn?.apply(this._window.console, arguments);
+  },
+
+  /**
+   * Wrapper for `console.error` function that checks for availability and interpolates
+   * error messages if a translation exists.
+   *
+   * @param {Error} error Error object with optional errorModule.
+   */
+  error(error?: ModuleError) {
+    if (this._window.console?.error) {
+      let errorMessage = '';
+
+      // If error module is present, attempt to interpolate string, else use error object message
+      if (error?.errorModule) {
+        errorMessage = error.errorModule.toInterpolatedMessage(error);
+        this._window.console.error(errorMessage);
+      } else {
+        // Use regular console.error
+        this._window.console.error.apply(this._window.console, arguments);
+      }
+    }
+  },
+};
+
+export default Logger;
--- a/packages/fxa-shared/lib/sentry.ts
+++ b/packages/fxa-shared/lib/sentry.ts
@@ -1,0 +1,174 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as Sentry from '@sentry/browser';
+import Logger from './logger';
+
+const ALLOWED_QUERY_PARAMETERS = [
+  'automatedBrowser',
+  'client_id',
+  'context',
+  'entrypoint',
+  'keys',
+  'migration',
+  'redirect_uri',
+  'scope',
+  'service',
+  'setting',
+  'style',
+];
+
+/**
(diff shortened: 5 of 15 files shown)


packages/fxa-components/jest.config.js
```diff
@@ -1,9 +1,12 @@
+const { resolve } = require("path");
+
 module.exports = {
   roots: ["<rootDir>"],
   transform: {
     "^.+\\.tsx?$": "ts-jest"
   },
   moduleNameMapper: {
     "\\.(css|scss)$": "identity-obj-proxy",
+    "^@fxa-shared/(.*)$": resolve(__dirname, "../fxa-shared", "$1")
```
jodyheavener, line 10: Jest can now resolve the @fxa-shared module when being tested in fxa-components.

packages/fxa-components/rescripts.js
```diff
@@ -83,9 +83,14 @@ const componentsJestMapper = {
...
-        "^@fxa-components/(.*)$": resolve(__dirname, "$1")
-      }
+      moduleNameMapper: Object.keys(additionalJSImports).reduce(
+        (previous, key) => {
+          return Object.assign(previous, {
+            [`^${key}/(.*)$`]: resolve(additionalJSImports[key], "$1")
+          });
+        },
+        {}
+      )
```
jodyheavener, line 93: I refactored the fxa-components Jest module resolver to match anything we extend to the Webpack resolver

packages/fxa-settings/src/index.tsx
```diff
@@ -6,12 +6,13 @@ import React from 'react';
...
 import App from './components/App';
+import sentryMetrics from '@fxa-shared/lib/sentry';
+import config from './lib/config';
 
 export async function init() {
-  // TODO: Configure Sentry logging
-  // if (config.sentry.dsn) {
-  //   sentryMetrics.configure(config.sentry.dsn, config.version);
-  // }
+  if (config.sentry.dsn) {
```
jodyheavener, line 13: I'm not sure if there's a downside to this, but I noticed in content-server we get the error `No Sentry dsn provided` logged unnecessarily.

packages/fxa-shared/lib/logger.ts
```diff
@@ -0,0 +1,89 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Logging module that performs proper checks and string
+ * interpolation before logging message.
+ *
+ */
+
+// tslint:disable-next-line: no-var-requires
+require('jsdom-global')();
```
jodyheavener, line 12: Was getting complaints that `window` was undefined.

packages/fxa-shared/lib/logger.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Constructor of log module.
+ *
+ * @param {Object} logWindow Window object that contains console used for logging.
+ *
+ * @constructor
+ */
+const Logger = (function(this: Logger, logWindow?: Window) {
+  this._window = logWindow || window;
+} as any) as new (logWindow?: Window) => Logger;
```
jodyheavener, line 37: Open to suggestions here (and the one in sentry.ts). I got to this because VS Code was complaining:

> 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.

And I also had to bind `this` to its own type.

packages/fxa-settings/src/index.tsx
chenba: If there's no `config.sentry.dsn` then it's likely there's no `config.sentry`.  Check for that?

I actually like that message, so we know there's an issue.  We could add an 'enable' flag to explicitly disable Sentry.

packages/fxa-shared/lib/logger.ts
```diff
@@ -0,0 +1,89 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Logging module that performs proper checks and string
+ * interpolation before logging message.
+ *
+ */
+
+// tslint:disable-next-line: no-var-requires
+// require('jsdom-global')();
```
chenba, line 12: What's this?

packages/fxa-settings/src/index.tsx
jodyheavener: > I actually like that message, so we know there's an issue.

Is it ever active in local development?

packages/fxa-shared/lib/logger.ts
jodyheavener: Ah I'll remove. Leftover from testing something

packages/fxa-settings/src/index.tsx
jodyheavener: > If there's no config.sentry.dsn then it's likely there's no config.sentry.

I think `config.sentry` will always exist, but our default is an [empty string](https://github.com/mozilla/fxa/pull/5203/files#diff-6f6012abb4afb051e7e0b053b896fedaR15), which means if we keeps this I should check for string length instead I guess

packages/fxa-shared/lib/logger.ts
chenba: WFM

packages/fxa-settings/src/index.tsx
chenba: > 
> Is it ever active in local development?

No, unless someone is working on Sentry related stuff.  But we did see it on prod Payments and realized the dsn is missing there.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,175 @@
...
+      // Docs: https://docs.sentry.io/hosted/learn/rollups/#fallback-grouping
+      if (data.tags.errno) {
+        data.fingerprint = ['errno' + data.tags.errno];
+        // if it is a known error change the error level to info.
+        data.level = Sentry.Severity.Info;
+      }
+    }
+
+    if (data.exception?.values) {
+      data.exception.values.forEach(value => {
```
chenba, line 48: `value` needs a type.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,175 @@
...
+          value.stacktrace.frames.forEach((frame: { abs_path?: string }) => {
+            if (frame.abs_path) {
+              frame.abs_path = cleanUpQueryParam(frame.abs_path); // eslint-disable-line camelcase
+            }
+          });
+        }
+      });
+    }
+
+    if (data.request?.headers?.Referer) {
```
chenba, line 59: This is inside the `if (data.request)` block, so the first `?` isn't really necessary.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * function that gets called before data gets sent to error metrics
+ *
+ * @param {Object} data
+ *  Error object data
+ * @returns {Object} data
+ *  Modified error object data
+ * @private
+ */
+function beforeSend(data: Sentry.Event): Sentry.Event {
```
chenba, line 31: A function that mutates and returns :sadpanda:

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,175 @@
...
+ * This must be configured with the `configure` method before use.
+ *
+ * Read more at https://docs.sentry.io/platforms/javascript
+ *
+ * @constructor
+ */
+const SentryMetrics = (function(this: SentryMetrics) {
+  this._logger = new Logger();
+  // tslint:disable-next-line: callable-types
+} as any) as { new (): SentryMetrics };
```
chenba, line 119: Could update to `new () => SentryMetrics` for syntax consistency with the other function constructor above.

packages/fxa-shared/lib/sentry.ts
```diff
@@ -0,0 +1,175 @@
...
+
+SentryMetrics.prototype = {
+  /**
+   * Configure the SentryMetrics instance for this singleton.
+   *
+   * @param {String} dsn
+   * @param {String} [release] - content server release version
```
chenba, line 126: s/content server/[something else]/

packages/fxa-shared/test/lib/logger.ts
```diff
@@ -0,0 +1,116 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { assert } from 'chai';
+import Logger from '../../lib/logger';
+import sinon = require('sinon');
```
chenba, line 7: import and require?  Pick a lane, man!

packages/fxa-settings/src/index.tsx
jodyheavener: Fair point!

packages/fxa-shared/test/lib/logger.ts
jodyheavener: oh wow

packages/fxa-shared/test/lib/logger.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      const information = 'Hello Firefox!';
+
+      logger = new Logger();
+      logger.info(information);
+
+      assert.isTrue(
+        (window.console.info as sinon.SinonSpy).calledOnceWith(information)
+      );
+      (window.console.info as sinon.SinonSpy).restore();
```
chenba, line 24: My preference's been assigning the spy into a const and then I can restore on it without type assertions.

packages/fxa-shared/test/lib/logger.ts
```diff
@@ -0,0 +1,116 @@
...
+    it('should create logger', () => {
+      sinon.spy(window.console, 'info');
+
+      const information = 'Hello Firefox!';
+
+      logger = new Logger();
+      logger.info(information);
+
+      assert.isTrue(
+        (window.console.info as sinon.SinonSpy).calledOnceWith(information)
```
chenba, line 22: If we use `sinon.assert` then it'd show the diff between the actual and expected on failure.  That's more helpful than "found false when expecting true" or whatever the message is.

chenba: (Of course you could use equal or deepequal from chai to do the same thing.)

packages/fxa-shared/lib/sentry.ts
jodyheavener: Is that bad?

chenba: Well it's all convention, personal preference, etc.  But I prefer one not both, and I prefer return instead mutate.  Bad?  Probably not, and it's intended to be private anyway.

PR #5219, opened 2020-05-05
sync jwt access tokens

dannycoates:


packages/fxa-auth-server/config/dev.json
```diff
@@ -362,7 +363,8 @@
       "enabledClientIds": [
         "98e6508e88680e1a",
         "dcdb5ae7add825d2",
-        "325b4083e32fe8e7"
+        "325b4083e32fe8e7",
+        "5882386c6d801776"
```
rfk, line 367: Luckily I know what this one is, but is there any way we can put comments on these? grumble grumble JSON should have inline comments grumble....

packages/fxa-auth-server/lib/oauth/token.js
```diff
@@ -35,7 +36,20 @@ exports.getTokenId = async function getTokenId(accessToken) {
...
 exports.verify = async function verify(accessToken) {
-  // JWT access tokens are still database backed, continue
+  // This block is a special case for Sync atm. We don't store these in the db.
+  // Eventually we won't store any JWT tokens.
+  if (!validators.jwt.validate(accessToken).error) {
+    const t = await JWTAccessToken.verify(accessToken);
+    const scope = ScopeSet.fromString(t.scope);
+    if (scope.contains(OAUTH_SCOPE_OLD_SYNC)) {
+      return {
+        user: t.sub, // TODO: need to consider PPID?
```
rfk, line 46: You shouldn't need to consider PPID for sync, I don't think.

If you didn't want to special-case this for sync, you could special-case it for "JWTs expiring in less than 6 hours don't need to be looked up in the db", by a parallel logic to the one you use when deciding whether to write them into the db.  (That will avoid hitting the db for any JWT that is "expiring soon", but that seems OK to me).

packages/fxa-auth-server/config/dev.json
dannycoates: at least it's defined in the same file ;)

packages/fxa-auth-server/lib/oauth/jwt_access_token.js
```diff
@@ -18,7 +19,11 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
   // always include the client_id in the `aud` claim. A future iteration of this code
   // should instead infer an appropriate default `aud` based on the requested scopes.
   // Ref https://github.com/mozilla/fxa/issues/4962
-  const audience = grant.resource ? [clientId, grant.resource] : clientId;
+  const audience = grant.resource
+    ? [clientId, grant.resource]
+    : grant.scope.contains(OAUTH_SCOPE_OLD_SYNC)
+    ? 'https://token.services.mozilla.com'
```
rfk, line 25: nit: this could profitably be extracted into a constant.

Better yet, maybe we can get it from the same place that configures the value advertised in https://accounts.firefox.com/.well-known/fxa-client-configuration? I think it would be valuable to have the value here match the one advertised to clients there, for consistency.

packages/fxa-auth-server/lib/oauth/jwt_access_token.js
```diff
@@ -43,6 +48,14 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
     claims['fxa-subscriptions'] = grant['fxa-subscriptions'].join(' ');
   }
 
+  if (grant.generation) {
+    claims['fxa-generation'] = grant.generation;
+  }
+
+  if (grant.profileChangedAt) {
+    claims['fxa-profileChangedAt'] = grant.profileChangedAt;
+  }
```
rfk, line 57: I don't object to it, but I'm surprised anyone would want/need to get `profileChangedAt` our of here, is this in service of a specific consumer?

rfk: Oh, I see, I guess this is so you can return it from `/verify` endpoint? 👍

packages/fxa-auth-server/lib/oauth/token.js
```diff
@@ -35,7 +36,23 @@ exports.getTokenId = async function getTokenId(accessToken) {
  * Verify an accessToken.
  */
 exports.verify = async function verify(accessToken) {
-  // JWT access tokens are still database backed, continue
+  // JWT tokens with lifespan < SHORT_ACCESS_TOKEN_TTL_IN_MS are not
+  // stored in the db
+  if (!validators.jwt.validate(accessToken).error) {
+    const t = await JWTAccessToken.verify(accessToken);
+    const lifespan = (t.exp - t.iat) * 1000;
```
rfk, line 43: Oh, very nice use of `iat` to reconstruct the original rule, rather than just doing a "does this have `SHORT_ACCESS_TOKEN_TTL_IN_MS` left before it expires" 👍

packages/fxa-auth-server/lib/oauth/db/index.js
```diff
@@ -61,7 +69,13 @@ class OauthDB {
       vals.expiresAt,
       vals.ttl
     );
-    if (POCKET_IDS.includes(hex(vals.clientId))) {
+    if (
+      JWT_ACCESS_TOKENS_ENABLED &&
+      JWT_ACCESS_TOKENS_CLIENT_IDS.has(vals.clientId) &&
+      token.ttl <= SHORT_ACCESS_TOKEN_TTL_IN_MS
```
rfk, line 75: This will cause Firefox to make JWT access tokens for any sufficiently-short-lived token, regardless of scope. That seems fine and good to me (and could be helpful if we want to test whether other systems accept JWTs, e.g. the profile server) but I wanted to note it explicitly in case it wasn't intend. I don't recall whether we talked about limited by "oldsync" scope here or not.

dannycoates: I originally had it for oldsync only, but we create a lot of 5 minute tokens for the login screen to fetch the avatar (which are already JWT tokens) so I decided to try it for all types. It should noticeably decrease our redis dataset.

packages/fxa-auth-server/lib/oauth/jwt_access_token.js
dannycoates: yes, though I don't know how it's used downstream of /verify. I'm blindly adding it because it exists in the other case.

rfk: I expect it's used by the profile-server to help with cache invalidation in some way. Can't guarantee that other RS's aren't also using it, but I'd be surprised.

vbudhram: > I expect it's used by the profile-server to help with cache invalidation in some way

Yea that is how it is currently used by the profile-server. It never gets returned to [reliers](https://github.com/mozilla/fxa/blob/4ab4c64e3a1a847291d0dd8d130f9dc698283e80/packages/fxa-profile-server/lib/routes/profile.js#L54) from profile server.

PR #5222, opened 2020-05-06
docs(adrs): Use GraphQL and Apollo for Settings Redesign

### LINKED ISSUE #4810: [ADRs] GraphQL + Settings Redesign ADR
The `fxa-admin-panel` + `fxa-admin-server` project offered the team a chance to try out and vet GraphQL. We've decided it's a great use case for the Settings Redesign and should document why (pros and cons) with an ADR.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1530)
┆Issue Number: FXA-1530

LZoog:
Closes #4810 

This is more of a follow up / documentation of a decision than it is a true proposal.

I'll need to go through and proofread and possibly add a little more detail tomorrow morning but in general, this feels pretty complete from my end, but I fully expect comments and especially would like @bbangert to go through and point out what I'm missing. 😄

I added the folks involved in the Settings Redesign project estimation meeting under "deciders" since I asked during that meeting if there was any opposition or reason why we shouldn't do this. I will take your name off if you prefer.

### DIFF
--- a/docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
+++ b/docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
@@ -1,0 +1,140 @@
+# Use GraphQL and Apollo for Settings Redesign
+
+- Deciders: Lauren Zugai, Ben Bangert, Wil Clouser, Jody Heavener, Les Orchard, Danny Coates, Barry Chen, Vijay Budhram, Dave Justice, Alex Davis
+- Date: 2020-05-05
+
+## Context and Problem Statement
+
+The [Settings Redesign project](https://jira.mozilla.com/browse/FXA-840) will be created [as a new React application](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md) and in turn, has opened the door to assess certain pieces of our technology stack.
+
+[GraphQL](https://graphql.org/), or GQL, is not a database query language, but is instead a query language for APIs. It _describes_ data requirements, and is a powerful alternative to REST. Some benefits can be gained by using it on top of existing REST architecture.
+
+"Apollo" in this document refers to the "Apollo client" and "Apollo server" pieces of the Apollo platform¹, which can be described as a unified data layer that enables applications to interact with data from data stores and APIs. In other words, it allows us to write and handle GraphQL on the client and server. Apollo also gives us many tools out of the box like caching.
+
+This ADR serves to lay out pros and cons of using GraphQL and Apollo in the Settings Redesign project as an alternative to hitting our conventional REST endpoints.
+
+¹Apollo also offers Apollo Graph Manager and Apollo Federation which are paid services, [read more from their docs](https://www.apollographql.com/docs/intro/platform/). We do not need to use these to use GQL with Apollo server or Apollo client.
+
+## Decision Drivers
+
+- Performance implications; consideration around the number of network requests and data transferred
+- Ease of setup for client-side API calls
+- Clarity around expected data, React integration, and developer tooling
+- Development speed, both around initial setup and as new features roll out
+
+## Considered Options
+
+- Option A - Send standard CRUD requests to our REST endpoints
+- Option B - Layer GraphQL and Apollo on top of our REST architecture, making direct database calls for basic operations
+- Option C - Full GraphQL and Apollo integration with direct database calls
+
+## Decision Outcome
+
+Chosen option: "B - Layer GraphQL on top of our REST architecture", because:
+
+- GQL offers performance optimizations by allowing us to consolidate our network requests by sending data requirements up in a single request, asking for only what is needed on the client, shifting the burden of figuring out how and where to gather this data from the client onto the server. The server then compensates for overfetching by only sending back what is requested.
+- Allowing developers to query for and expect only exactly what is needed with end-to-end typing and a more "declarative" way of thinking towards data requirements, along with keeping data requirement schemas close to where they're consumed, makes it painfully clear what's being sent and received between client and server.
+- Does not preclude option C, as we can replace or supplement direct calls to the FxA auth-server down the line. This is faster for initial development, and this option over option C will also help mitigate risk around using a relatively novel piece of technology for FxA.
+- Nice side-effects include the GQL playground, managing only a single API endpoint, and the ability to store both local state and network data in the Apollo cache.
+
+## Pros and Cons of the Options
+
+### Option A - Send standard CRUD requests to our REST endpoints
+
+#### Description
+
+- Don't use GraphQL. Send all requests to their corresponding endpoints.
+
+##### Pros:
+
+- No novel tech.
+- No additional setup time.
+- No new technical debt.
+- Our REST API setup has been vetted by years of use in production.
+
+##### Cons:
+
+- While we would avoid new technical debt with this option, we would still want to work towards optimizing our Settings page network calls. We would likely need to spend additional time looking into how we can manually optimize our calls, as using our current REST pattern will result in sending many queries and overfetching data.
+- We may need to figure out a more ideal way to organize our API calls than using `fxa-payments-server` as a blueprint.
+- We'll be turning down all of the benefits laid out in the GraphQL section, like clarity on the client-side around what data is being sent and received.
+- If we later choose to use GraphQL, even with the option B approach, we'll face a lot of refactoring at a later time that could be mitigated now.
+
+### Options B & C - use GraphQL and Apollo
+
+#### Description
+
+This section outlines the pros and cons for using GraphQL and Apollo in general and applies to both options B and C.
+
+While GraphQL will be novel tech to the FxA stack, it's not entirely novel to the FxA ecosystem. That is to say, the Admin Panel project (`fxa-admin-panel`/`fxa-admin-server` packages) served as a prototype for experimenting with GQL and Apollo.
+
+##### Pros
+
+- We can use GQL for all Settings Redesign needs and won't need to mix GQL and REST.
+- GQL can consolidate network requests. Even if some GQL queries need to be fed into the auth-server (option B), the client will be relieved of a heavy network request burden and instead, a single request from the client asks the server to perform all of the requests. This should offer a performance boost especially on the main Settings page, where many sequential requests must be made.
+- Since only required data will be requested, the data returned won't be overfetched like it is in traditional REST architecture, resulting in a smaller payload.
+- GQL offers a flexible and declarative approach to data - both FE and BE developers can focus on describing the data rather than implementing and optimizing numerous REST endpoints. While REST patterns tend to define resources on the server, data requirement schemas for GQL are closer to the client, where the data is consumed. This makes it painfully clear what's being sent and received between client and server which can improve visibility into what's happening in components.
+- GraphQL and Apollo have gained a lot of popularity over the years and now have substantial community support and company backing (Airbnb, Facebook, SurveyGizmo, Github, and many others).
+- This stack promotes strong end-to-end typing.
+- Apollo offers great features out of the box like caching, error handling, optimistic rendering, and nice integration tools for React (and even provides React hooks).
+- The React Settings application will only have to manage interacting with a single endpoint and Apollo Client makes managing the data in our React app straightforward. Apollo Client creates its own internal Redux store under the hood and allows us to store local data in the Apollo cache, the same place that data from network requests will be stored.
+- Apollo also provides a "GraphQL playground" which is a graphical, interactive, in-browser GQL IDE. It not only provides easy to access schemas (and is "self-documenting" in a way), but it allows developers to write queries or mutations to send to the server so the developer can verify it behaves as expected. This can then be copied directly over to where it needs to be used in the application.
+- GraphQL makes it easier to evolve APIs over time. It provides a way to deprecate schema members and discourage usage of certain parts of APIs with custom messaging.
+- GraphQL is database agnostic.
+- GraphQL queries are easy to write and understand.
+
+#### Neutral
+
+- Queries always return a 200 status code, even if the query errored. If the query is unsuccessful, the response will instead have an `errors` key with associated error messages and stacktrace. However, the error messaging can be quite detailed and include all the resolvers and refer to the exact part of the query at fault.
+
+##### Cons
+
+- The learning curve that comes with introducing new tech.
+- Initial setup time of another service (`graphql-api-server`).
+- While GQL caching can be done at the database or client level though Apollo, GQL doesn't rely on HTTP caching methods.
+- While GQL has been around long enough to gain excellent community support, REST has been around with established patterns for even longer.
+
+### Option B - Layer GraphQL and Apollo on top of our REST architecture, making direct database calls for basic operations
+
+#### Description
+
+- This option allows us to begin working with GraphQL and Apollo on top of our REST architecture, proxying existing auth-server calls, except where it makes sense to interact with the database directly for simple operations.
+
+##### Pros:
+
+- Less risk and less upfront work than option C.
+- We'll still heavily reduce the number of initial network requests performed on the Settings frontend with an incremental approach. The Apollo Server can orchestrate the REST requests sufficient to fulfil just that query and return exactly the data requested.
+- Apollo server can proxy to existing auth-server calls. We can incrementally swap out auth-server API calls with direct implementation.
+
+##### Cons:
+
+- Creates tech debt. We won't be fully integrated until we finish swapping out the auth-server API calls.
+- Our stack will be less universal across FxA. We'll use REST for non-Settings FxA pages and for the payments server until refactoring can be done.
+- While GraphQL is database agnostic, we will be unable to use [GraphQL Subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/) because our database does not support real-time data.
+
+### Option C - Full GraphQL and Apollo integration with direct database calls
+
+#### Description
+
+- This option would extend option B. The more complex operations routed through the auth-server would be directly implemented through the Apollo server.
+
+##### Pros
+
+- Less tech debt later.
+- We would see a performance boost over option B.
+- An all-out dedication to this approach could open the door to evaluating our database options (and being able to work with [GraphQL subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/)) or more quickly pave a path to move our backend off of MySQL to something else for a more regional conscious architecture.
+
+##### Cons
+
+- Increased risk.
+  - While we'll be pushing the new Settings app to staging and production while we're building it out, the Settings Redesign project already has a lot of moving parts. An incremental approach where we can take it is more ideal.
+  - Existing REST endpoints in auth-server aren't just queries, they embed business logic. We'd have to duplicate that logic in GraphQL and test in tandem with REST to make sure we don't break anything on either side.
+- Much more time needed for setup, and access control, auth, and security already implemented by auth-server might be expensive to reinvent.
+
+## Additional Links
+
+- [Settings Redesign project](https://jira.mozilla.com/browse/FXA-840) (Jira)
+- [ADR for Settings Redesign as a new React App](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md)
+- [GraphQL docs](https://graphql.org/learn/)
+- [Apollo docs](https://www.apollographql.com/docs/intro/platform/)
+- [GraphQL subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/)
+- [GraphQL API Technical Specification Doc](https://docs.google.com/document/d/1ZoUsCCGwgObiG3_f99tNANODr7FZfvck2R4sOij0Lj4/edit#heading=h.2gazcsgmxkub) for Settings Redesign


docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+- GraphQL is database agnostic.
+- GraphQL queries are easy to write and understand.
+
+##### Cons
+
+- The learning curve that comes with introducing new tech.
+- Initial setup time of another service (`graphql-api-server`).
+- While GQL caching can be done at the database or client level though Apollo, GQL doesn't rely on HTTP caching methods.
+- While GQL has been around long enough to gain excellent community support, REST has been around with established patterns for even longer.
+- Queries always return a 200 status code, even if the query errored. If the query is unsuccessful, the response will instead have an `errors` key with associated error messages and stacktrace. However, the error messaging can be quite detailed and include all the resolvers and refer to the exact part of the query at fault.
```
jodyheavener, line 92: Hmmm, I might even consider this a neutral. It's not that a consistent 200 status is bad, per se, it's just _different_ than what's expected from a traditional REST setup. After all, the network request is successful, it's just the database operation that might not be, right?

Relatedly, what happens if the GQL server can't be reached at all?

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+- Deciders: Lauren Zugai, Ben Bangert, Wil Clouser, Jody Heavener, Les Orchard, Danny Coates, Barry Chen, Vijay Budhram, Dave Justice, Alex Davis
+- Date: 2020-05-05
+
+## Context and Problem Statement
+
+The [Settings Redesign project](https://jira.mozilla.com/browse/FXA-840) will be created [as a new React application](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md) and in turn, has opened the door to assess certain pieces of our technology stack.
+
+[GraphQL](https://graphql.org/) is not a database query language, but is instead a query language for APIs. It _describes_ data requirements, and is a powerful alternative to REST. Some benefits can be gained by using it on top of existing REST architecture.
+
+"Apollo" in this document refers to the "Apollo client" and "Apollo server" pieces of the Apollo platform¹, which can be described as a unified data layer that enables applications to interact with data from data stores and APIs. In other words, it allows us to use write and handle GraphQL on the client and server. Apollo also gives us many tools out of the box like caching.
```
jodyheavener, line 12: Should "us to use write and handle" be "us to write and handle", or possibly "us to use, write, and handle"

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+
+This ADR serves to lay out pros and cons of using GraphQL and Apollo in the Settings Redesign project as an alternative to hitting our conventional REST endpoints.
+
+¹Apollo also offers Apollo Graph Manger and Apollo Federation which are paid services, [read more from their docs](https://www.apollographql.com/docs/intro/platform/). We do not need to use these to use GQL and Apollo server and client.
```
jodyheavener, line 16: Could be "to use _the_ GQL and Apollo server and client", or possibly "to use the GQL server and Apollo client"?

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
+# Use GraphQL and Apollo for Settings Redesign
+
+- Deciders: Lauren Zugai, Ben Bangert, Wil Clouser, Jody Heavener, Les Orchard, Danny Coates, Barry Chen, Vijay Budhram, Dave Justice, Alex Davis
+- Date: 2020-05-05
+
+## Context and Problem Statement
+
+The [Settings Redesign project](https://jira.mozilla.com/browse/FXA-840) will be created [as a new React application](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md) and in turn, has opened the door to assess certain pieces of our technology stack.
+
+[GraphQL](https://graphql.org/) is not a database query language, but is instead a query language for APIs. It _describes_ data requirements, and is a powerful alternative to REST. Some benefits can be gained by using it on top of existing REST architecture.
```
jodyheavener, line 10: Maybe somewhere up here we can note that it's referred to as "GQL" in some places?

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+- Increased risk. While we'll be pushing the new Settings app to staging and production while we're building it out, the Settings Redesign project already has a lot of moving parts and an incremental approach definitely feels more ideal.
+- Much more time needed for setup, and access control, auth, and security already implemented by auth-server might be expensive to reinvent.
+
+## Additional Links
+
+- [Settings Redesign project](https://jira.mozilla.com/browse/FXA-840) (Jira)
+- [ADR for Settings Redesign as a new React App](https://github.com/mozilla/fxa/blob/master/docs/adr/0011-create-new-react-app-for-settings-redesign.md)
+- [GraphQL docs](https://graphql.org/learn/)
+- [Apollo docs](https://www.apollographql.com/docs/intro/platform/)
+- [GraphQL subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/)
```
jodyheavener, line 135: Do we want to add a link to Ben's "GraphQL API Technical Specification"?

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+- This stack promotes strong end-to-end typing.
+- Apollo offers great features out of the box like caching, error handling, optimistic rendering, and nice integration tools for React (and even provides React hooks).
+- The React Settings application will only have to manage interacting with a single endpoint and Apollo Client makes managing the data in our React app straightforward. Apollo Client creates its own internal Redux store under the hood and allows us to store local data in the Apollo cache, the same place that data from network requests will be stored.
+- Apollo also provides a "GraphQL playground" which is a graphical, interactive, in-browser GQL IDE. It not only provides easy to access schemas (and is "self-documenting" in a way), but it allows developers to write queries or mutations to send to the server so the developer can verify it behaves as expected. This can then be copied directly over to where it needs to be used in the application.
+- CRUD operations have been simplified to simply queries and mutations. CRUD in GraphQL becomes "MQMM" and developers don't have to think about a C/U/D differentiation outside of writing the mutation for them.
```
jodyheavener, line 81: Might be beneficial to link up or describe MQMM (and to a certain extent, CRUD, but less important).

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+- No additional setup time.
+- No tech debt.
+- Our REST API setup has been vetted by years of use in production.
+
+##### Cons:
+
+- A lack of needed performance gains -- we won't solve our heavy initial network request problem and we'll continue to overfetch data as REST does.
+- We may need to figure out a more ideal way to organize our API calls than using `fxa-payments-server` as a blueprint.
+- We'll be turning down all of the benefits laid out in the GraphQL section, like clarity on the client-side around what data is being sent and received.
+- If we later choose to use GraphQL, even with the option B approach, we'll face a lot of refactoring at a later time that could be mitigated now.
```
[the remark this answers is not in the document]
LZoog: I suppose I wanted to reiterate in the "cons" of "let's not do this new thing" are mostly around turning down all of the benefits we'd gain by using it, and using it in Settings now. This isn't meant to be a direct "GraphQL or use REST" technology comparison, but a comparison with FxA and the Settings Redesign project in mind. 🤔

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,135 @@
...
+
+##### Pros
+
+- We can use GQL for all Settings Redesign needs and won't need to mix GQL and REST.
+- GQL can consolidate network requests. Even if the GQL queries need to go through the auth server, the client will be relieved of a network request burden and instead, a single request from the client lets the server perform all of the requests. This will offer a performant boost especially on the main Settings page, where currently around 17 different requests must be made.
+- Since only required data will be requested, the data returned won't be overfetched like it is in traditional REST architecture, resulting in a smaller payload.
+- GQL offers a flexible and delcarative approach to data - both FE and BE developers can focus on describing the data rather than implementing and optimizing numerous REST endpoints. This makes it painfully clear what's being sent and received between client and server which can improve visibility into what's happening in components.
+- GraphQL and Apollo have gained a lot of popularity over the years and now have substantial company and community support.
```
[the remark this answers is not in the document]
LZoog: I'll be more clear here. It's general corporate interest, some pretty major companies use GQL now like Airbnb, Amplitude, SurveyGizmo, Facebook, Github, Wayfair, Shopify...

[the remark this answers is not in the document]
jodyheavener: Makes sense!

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: Hmm well, GQL isn't Apollo, but I'll rephrase to, "to use GQL with Apollo server or Apollo client" for clarity.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: Ah yes, good call, I should add that doc to the Redesign epic too.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: I'm actually not even sure this is a "pro"... Maybe it should be neutral. It's kind of a weird thing to get used to but I like the simplification.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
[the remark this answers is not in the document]
jodyheavener: Fair point - no, I think this is fine to leave. I guess my original thought was around the idea of "the presence of a new technology does not create a con for an old technology"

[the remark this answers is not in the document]
LZoog: Yeah, and that makes sense as well. I think the 4th point (if we later choose / refactoring) is fine/valid, but if anyone thinks I should take the third point out (turning down all benefits) I can remove👍

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,136 @@
...
+## Decision Drivers
+
+- Performance
+- Easy API setup and React integration
+- Clarity around "what's happening?" and data received from the server
+- Development speed, both around initial setup and as new features roll out
+
+## Considered Options
+
+- Option A - Send standard CRUD requests to our REST endpoints
```
chenba, line 27: This is a bit unfair to REST, but unfortunately we all know what you mean by that.

LZoog: I may add a couple more pros to option A for just REST in general. I definitely spent more time on the GQL pros section.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,136 @@
...
+### Option B - Layer GraphQL and Apollo on top of our REST architecture, making direct database calls for basic operations
+
+#### Description
+
+- This option allows us to begin working with GraphQL and Apollo on top of our REST architecture, proxying existing auth-server calls, except where it makes sense to interact with the database directly for simple operations.
+
+##### Pros:
+
+- Less risk and less upfront work than option C.
+- We'll still heavily reduce the number of initial network requests performed on the Settings frontend with an incremental approach. The Apollo Server can orchestrate the REST requests sufficient to fulfil just that query and return exactly the data requested.
```
chenba, line 103: "fulfill"

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
chenba: Oh, I don't know if it's intentional, but that seems to equate REST with CRUD.  While REST is not CRUD, as suggested my in previous comment, unfortunately that's what people think anyways.  I was not suggesting any changes. 😃

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,136 @@
...
+##### Pros:
+
+- No novel tech.
+- No additional setup time.
+- No tech debt.
+- Our REST API setup has been vetted by years of use in production.
+
+##### Cons:
+
+- A lack of needed performance gains -- we won't solve our heavy initial network request problem and we'll continue to overfetch data as REST does.
```
vbudhram, line 57: > A lack of needed performance gains

I am curious how much of a performance gain we get from reducing the network requests. The request are pretty light and return quickly.

chenba: Yeah, and I'm also curious how much gain we could get from HTTP/2.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
jodyheavener: TIL fulfil and fulfill are both words

chenba: WTF, English!

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
bbangert: Our requests in many cases are already using HTTP/2, as the load-balancers now support that we use. HTTP/2 only helps with the request setup, the problem here is that the requests are done sequentially. The requests could definitely be optimized manually in the current setup, perhaps even transparently to the app (by using a network intercept hook to cache the call if it's done more than once per 2-seconds for example). We could also fire all the requests at once, rather than sequentially, but none of this is efficient or great for our server load.

I believe graphql offers a winning proposition here, because it allows components in the page to indicate the data they need, which can be automatically collapsed by the apollo tooling and expressed as a single graphql request. Doing a similar thing for RESTful calls requires manual optimization every time the settings page changes, to ensure we're doing as little calls as needed for the data fetched. The server then benefits as well since it can do the minimum work required to get the data needed.

bbangert: Oops, to be more specific to this point, perhaps a more accurate phrasing would be that:
- REST results in us either overfetching data, doing more queries than needed, or spending additional time manually optimizing calls as we update the settings page.

chenba: Heh, I wasn't suggesting HTTP/2 would _fix_ how the requests are done, but curious how much of a gain it would be.

> Our requests in many cases are already using HTTP/2

This is not true for content server, auth, or profile.  Payments is using HTTP/2 though.

chenba: > the problem here is that the requests are done sequentially

This is a good point, and perhaps should be made explicit in the ADR.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: Ah I missed this comment from this morning.

I actually had this in "Neutral" at first because of the level of detail I've read you can add to errors, but moved it into "Cons" because I feel yeah, it's not what you'd expect to happen. I'll move it back to neutral.

I'm reading, "Non-200 response codes indicate that some issue occurred at the HTTP transport layer, not the GraphQL layer." If it can't be reached, we'll just see a `404 not found`, I just killed the `fxa-admin-server` while trying to query in `fxa-admin-panel` to double check. I do wonder if we hit some `500` error in the auth-server and Apollo sends us back a `200` how easily we can find that error code or what exactly we'll see under `errors`, maybe @bbangert knows?

jodyheavener: This explanation makes sense to me. Thanks for checking!

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+- Option B - Layer GraphQL and Apollo on top of our REST architecture, making direct database calls for basic operations
+- Option C - Full GraphQL and Apollo integration with direct database calls
+
+## Decision Outcome
+
+Chosen option: "B - Layer GraphQL on top of our REST architecture", because:
+
+- GQL seems to be the ideal solution to the heavy number of network requests on the main Settings page where around 17 different requests are sent to our REST API, and a lot of data received is superfluous. GQL moves us towards a performant solution that consolidates these initial requests and reduces overfetching in other requests.
+- Allowing developers to query for and expect only exactly what is needed with end-to-end typing and a more "declarative" way of thinking towards data requirements makes it painfully clear what's being sent and received between client and server.
+- The React Settings application will only have to manage interacting with a single endpoint and Apollo Client makes managing the data in our React app straightforward.
```
dannycoates, line 37: This is not a very strong _argument_ in favor, more of a nice side-effect.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+
+"Apollo" in this document refers to the "Apollo client" and "Apollo server" pieces of the Apollo platform¹, which can be described as a unified data layer that enables applications to interact with data from data stores and APIs. In other words, it allows us to write and handle GraphQL on the client and server. Apollo also gives us many tools out of the box like caching.
+
+This ADR serves to lay out pros and cons of using GraphQL and Apollo in the Settings Redesign project as an alternative to hitting our conventional REST endpoints.
+
+¹Apollo also offers Apollo Graph Manager and Apollo Federation which are paid services, [read more from their docs](https://www.apollographql.com/docs/intro/platform/). We do not need to use these to use GQL with Apollo server or Apollo client.
+
+## Decision Drivers
+
+- Performance
```
dannycoates, line 20: what kind? vague drivers are hard to evaluate

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+- Easy API setup and React integration
+- Clarity around "what's happening?" and data received from the server
```
dannycoates, line 22: I'd simplify this as "developer tooling" or something similar

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
lmorchard: Edit: I think this stuff is mentioned later in the "use GraphQL and Apollo" pros. Could be worth tl;dr'ing here in a few more words' detail.

We'd talked about the ability to reduce requests and "wasted" data queries, back in Berlin. Not sure we have that documented or recorded anywhere, but it would be the sort of thing worth mentioning here if only briefly.

Level of detail in ADRs is fuzzy: We don't necessarily want to sketch out complete tech specs, but it's worth including information that drives decisions. (e.g. what, if anything, about GraphQL makes this easier / more effective than more classic REST patterns)

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+- Option A - Send standard CRUD requests to our REST endpoints
+- Option B - Layer GraphQL and Apollo on top of our REST architecture, making direct database calls for basic operations
+- Option C - Full GraphQL and Apollo integration with direct database calls
+
+## Decision Outcome
+
+Chosen option: "B - Layer GraphQL on top of our REST architecture", because:
+
+- GQL seems to be the ideal solution to the heavy number of network requests on the main Settings page where around 17 different requests are sent to our REST API, and a lot of data received is superfluous. GQL moves us towards a performant solution that consolidates these initial requests and reduces overfetching in other requests.
+- Allowing developers to query for and expect only exactly what is needed with end-to-end typing and a more "declarative" way of thinking towards data requirements makes it painfully clear what's being sent and received between client and server.
```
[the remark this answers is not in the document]
lmorchard: Yeah, I think this is a kernel of the performance argument. 

That is, we tell the computer what data we want and it figures out how to gather exactly that. Versus, say, making many requests for multiple resources that pre-define what they provide - which may or may not directly match what we want (e.g. "wasted" queries).

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
lmorchard: I think the GraphQL argument here is that it moves the data schemas & requirements more into the client, closer to where the data's used. The classic server-side API pattern is that the endpoints define what they return and the clients request and rework the data as necessary.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+##### Pros:
+
+- No novel tech.
+- No additional setup time.
+- No new technical debt.
+- Our REST API setup has been vetted by years of use in production.
+
+##### Cons:
+
+- We'd still want to work towards optimizing the way we do our Settings page network calls, which has around 17 executed sequentially. REST results in either overfetching data, doing more queries than is needed, or spending additional time manually optimizing calls as we update the Settings page.
```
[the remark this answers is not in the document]
lmorchard: I think GraphQL's selling point is that it implicitly optimizes queries with frontend declaration of data needs. But yeah, I don't know how much thinking we've thrown at manually optimizing what we have. I remember Phil had started doing some of that thinking back around the start of all-things-subscriptions with [the `/account` endpoint in auth-server](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/account.js#L1431), but we didn't go much further with it.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+
+##### Pros
+
+- We can use GQL for all Settings Redesign needs and won't need to mix GQL and REST.
+- GQL can consolidate network requests. Even if some GQL queries need to be fed into the auth-server (option B), the client will be relieved of a heavy network request burden and instead, a single request from the client asks the server to perform all of the requests. This will offer a performance boost especially on the main Settings page, where currently around 17 different requests must be made.
+- Since only required data will be requested, the data returned won't be overfetched like it is in traditional REST architecture, resulting in a smaller payload.
+- GQL offers a flexible and declarative approach to data - both FE and BE developers can focus on describing the data rather than implementing and optimizing numerous REST endpoints. This makes it painfully clear what's being sent and received between client and server which can improve visibility into what's happening in components.
```
lmorchard, line 75: Ah, I think this is the point I was looking for in earlier comments. One of the key GraphQL vs REST points, IMO, is moving data requirement schemas from server closer to client where the data is consumed. REST patterns tend to define resources on the server, GraphQL accepts data requirements from the client.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
```diff
@@ -0,0 +1,138 @@
...
+
+##### Pros
+
+- Less tech debt later.
+- We would see a performance boost over option B.
+- An all-out dedication to this approach could open the door to evaluating our database options (and being able to work with [GraphQL subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/)) or more quickly pave a path to move our backend off of MySQL to something else for a more regional conscious architecture.
+
+##### Cons
+
+- Increased risk. While we'll be pushing the new Settings app to staging and production while we're building it out, the Settings Redesign project already has a lot of moving parts and an incremental approach definitely feels more ideal.
```
lmorchard, line 128: I'd say it's worth further noting the risk here of having two different systems hitting the same database: Existing REST endpoints in auth-server aren't just queries, they embed business logic. We'd have to duplicate that logic in GraphQL and test in tandem with REST to make sure we don't break anything on either side.

My concern here is that the tangle between existing settings-with-REST and new settings-with-GraphQL could stomp over a lot of the productivity gains of adopting GraphQL in the first place. 

I might be over-worried about that, and maybe it's a quick rewrite. But surprises and the second-system effect is why folks so often [advise against greenfield rewrites](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/).

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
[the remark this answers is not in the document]
LZoog: Yeah, I have under "pros" that there's no "new" technical debt, but my point with this "con" was that we would not be without tech debt if we went with this option. We'd want to look into optimizing this regardless.

I can rephrase this to be:

> While we would avoid new technical debt with this option, we would still want to work towards optimizing our Settings page network calls which currently sends around 17 sequential requests from the client. Using our current REST pattern will result in sending more queries than is needed and overfetching data. We would likely need to spend additional time looking into how we can manually optimize our calls.

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: I expounded upon this a bit from your comment (pushing changes up soon).

docs/adr/0016-use-graphql-and-apollo-for-settings-redesign.md
LZoog: Thanks for comments from both of you in this section, slightly rephrasing a couple of points.

PR #5233, opened 2020-05-07
fix(payments-server): remove border in update pay

### LINKED ISSUE #5054: Extra border in update billing form
!image-2020-04-22-00-57-36-739.png|thumbnail!

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1723)
┆Issue Number: FXA-1723
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/13100/image-2020-04-22-00-57-36-739.png">image-2020-04-22-00-57-36-739.png</a>

meandavejustice:
- fixes #5054

### DIFF
--- a/packages/fxa-payments-server/src/routes/Subscriptions/index.scss
+++ b/packages/fxa-payments-server/src/routes/Subscriptions/index.scss
@@ -23,6 +23,10 @@
     margin: 0 -16px;
   }
 
+  hr + hr {
+    border: 0;
+  }
+
   .subscription {
     margin: 25px 32px;
 


packages/fxa-payments-server/src/routes/Subscriptions/index.scss
```diff
@@ -23,6 +23,10 @@
     margin: 0 -16px;
   }
 
+  hr + hr {
+    border: 0;
```
jodyheavener, line 27: Do our horizontal rules take up any vertical space (like margin or padding)? If so we may be better off setting `display: none`, otherwise it'll be visually hidden while still taking up space. If not, nvm!

meandavejustice: In this case we want it to take up space

jodyheavener: I dont want to block you on this train, but I’d probably recommend reworking some of the styles to add the additional space that’s desired here, instead of doubling up on HRs. Can be a follow up or just a mental note. It’s also possible I’m missing some context here. Carry on!

meandavejustice: HRs for spacing is not the most semantic, I agree. Just wanted to get this quick fix out and it was already doubled up HRs, I will file a follow up

PR #5234, opened 2020-05-07
chore(images): add fortress icons to cdn

johngruen:
@LZoog quick merge? just adding the fortress images to the CDN


lmorchard:
Probably need to add a note to this train's deploy doc - I don't think these images are automatically uploaded to the CDN as a part of the usual push

jrgm:
Just to followup, these were uploaded to cdn with _scripts/upload-assets-to-cdn.sh on Tuesday.

PR #5251, opened 2020-05-08
feat (auth-server): add plan changed metadata to subscription

### LINKED ISSUE #5243: Add Plan Change Metadata to Subscription
When a user changes the Plan associated with their existing subscription, the following metadata should be added to the Subscription:

previous_plan_id

plan_change_date

Adding this information to the Subscription will allow for us to add it to our Zendesk Support Panel, so that our agents will be able to see when changes were made to a user's subscription.

https://mana.mozilla.org/wiki/display/COPS/SP+Tiered+Product+Support#SPTieredProductSupport-ZendeskSupport

https://mana.mozilla.org/wiki/display/COPS/SP+Tiered+Product+Support#SPTieredProductSupport-Subscription

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1847)
┆Issue Number: FXA-1847

jackiemunroe:
- add previous_plan_id and plan_change_date to subscription on update

Because:
- we want to be able to provide support agents with information concerning when a plan was changed and what the previous plan was to allow for better support for customers

fixes #5243

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const Sentry = require('@sentry/node');
+const moment = require('moment');
 const error = require('../error');
 
 const stripe = require('stripe').Stripe;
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -604,23 +605,28 @@ class StripeHelper {
...
       items: [
         {
           id: subscription.items.data[0].id,
-          plan: planId,
+          plan: newPlanId,
         },
       ],
+      metadata: {
+        previous_plan_id: currentPlanId,
+        plan_change_date: moment().unix(),
```
lmorchard, line 628: Do we need moment here? Would `Math.floor(Date.now() / 1000)` would give the same result?

jackiemunroe: I thought that it would read more clearly to use `moment().unix()`. A new contributor may not realize that `Math.floor(Date.now() / 1000)` results in a Unix timestamp in seconds.

jackiemunroe: You are also correct that it made testing easier, since I could easily stub the result 😅

lmorchard: Cool, those sound like good reasons to me! :+1:

Issue #5244, opened 2020-05-07
[setup] Setup Tailwind config and any additional CSS foundation


data-sync-user:
Whatever decision comes out of FXA-1748 will need to be setup. At the time of writing, it seems we're leaning towards the Tailwind option which is going to need a config file, but this shouldn't be worked on until the ADR has been approved.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1849)
┆Issue Number: FXA-1849

PR #5252, opened 2020-05-08
docs(adr): Use Utility-First CSS (Tailwind) with Custom SCSS

### LINKED ISSUE #5087: [ADRs] CSS part 2 (CSS architecture, names, helper classes or tiny lib, VD consistency)
This issue is part 2 of 2 of our CSS architecture.

We need to determine the way in which we want to handle CSS in the settings redesign that will propagate out to other areas of the FxA ecosystem.

1) FXA-1528; CSS variables vs SCSS variables, separate vs CSS-in-JS (`styled-components` or `Tailwind`), CSS modules, SCSS organization & use (this issue)
2) This issue; CSS architecture, standards, & naming conventions; flexbox vs CSS grid, possible creation of (tiny) CSS framework, helper classes, variable values working with visual design

Careful consideration of options and working with Phil Walmsley to come up with a design system for FxA will set us up for great consistency and ease-of-use in the future.

See some discussion here for part 1: https://docs.google.com/spreadsheets/d/1WNRY_hvOrn4ypIv-WHY-DGnxNpje4IlgCaK11F6eus8/edit?disco=AAAAJYHxp78

While creating a design guide page for UX & devs will be a separate task, it would be ideal to have at least several examples of usage alongside these ADRs for reference while building out Settings (for part 2).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1748)
┆Issue Number: FXA-1748

jodyheavener:
Closes #5087

> This ADR serves to determine how we'll approach our CSS architecture in the Settings Redesign project, evaluating libraries and frameworks to determine which if any will be the best option for the FxA ecosystem.

📖 [Pretty version](https://github.com/mozilla/fxa/blob/jh/5087/css-adr-p2/docs/adr/0018-use-tailwind-with-custom-scss.md)

### DIFF
--- a/docs/adr/0018-use-tailwind-with-custom-scss.md
+++ b/docs/adr/0018-use-tailwind-with-custom-scss.md
@@ -1,0 +1,145 @@
+# Use Utility-First CSS (Tailwind) with Custom SCSS
+
+- Status: proposed
+- Deciders: Jody Heavener, Lauren Zugai, Danny Coates
+- Date: 2020-05-08
+
+## Context and Problem Statement
+
+The [Settings Redesign project](https://github.com/mozilla/fxa/issues/3740) provides us with an opportunity to review how FxA approaches and employs CSS, both while building out new components for this project and for FxA going forward.
+
+Historically, the Firefox Accounts codebase has not adhered to a formal CSS structure. This ADR serves to determine how we'll approach our CSS architecture in the Settings Redesign project, evaluating libraries and frameworks to determine which if any will be the best option for the FxA ecosystem. It is part 2 of two [Settings Redesign CSS ADRs](https://github.com/mozilla/fxa/issues/5087); part 1, detailing how we'll approach build conventions and variables, [can be found here](https://github.com/mozilla/fxa/blob/master/docs/adr/0015-use-css-variables-and-scss.md).
+
+Considerations around class naming conventions, color and measurement standards, interoperability across shared components, and custom configuration options offered by each library to meet Settings Redesign design standards are taken into account. Notably, the new design uses space measurements in increments of 8px and [colors](https://protocol.mozilla.org/fundamentals/color.html) are based in Mozilla Protocol's design system, where a hue's brightness scales in increments of 10.
+
+## Decision Drivers
+
+- **Reusability** - does the approach yield DRY, lean code that can be reused and repurposed?
+- **Longevity** - will the approach be supported in upcoming years and will it provide a stable platform for years of revolving HTML through added features and bug fixes?
+- **Developer experience** - are some team members already familiar with the approach, making the transition easier than an unfamiliar one?
+- **Ease of use** - will the choice result in a large learning curve or be easy for engineers to pick up?
+
+## Considered Options
+
+- Option A - Use an existing UI library
+- Option B - Use an existing utility library
+- Option C - Develop our own hybrid component and utility library
+- Option D - Custom SCSS with loose structure
+
+## Decision Outcome
+
+Chosen options: "Option B" with Tailwind CSS for majority styling, and implementation details from "Option D" when utility classes don't meet the entire need, because:
+
+- Of the options set forth, a utility library provides us with the most flexible yet durable set of tools.
+  - Single-purpose classes are performant and reduce the possibility of overly-complex or convoluted stylesheets.
+  - A utility library is leaner and less opinionated compared to a set of UI components and other options, allowing greater flexibility and reusability across various projects.
+- Our team has prior experience with Tailwind in particular and newcomers should ramp up quickly with a utility pattern.
+- Tailwind is highly configurable without being cumbersome, allowing us to modify type and spacing scales, define color ranges, and set up media queries to meet our exact needs.
+- For cases when we do need to write custom SCSS we will structure our React components to initially rely on utility classes, but allow additional custom styles to be written in an adjacent SCSS file when needed. This is also applicable to components in `fxa-components` where the component can accept a `classes` prop with a list of needed utility classes, and any additional styling can be done in an external SCSS file located where the component was composed as needed (e.g., outside of `fxa-components`). CSS variables can be shared across the Tailwind configuration and in custom SCSS.
+  - Note: class name conventions for the custom SCSS will be declared when the library configuration is setup, as we'd like them to make sense together. Examples will be provided in the `fxa-settings` README at this time as well.
+
+## Pros and Cons of the Options
+
+### Option A - Use an existing UI library
+
+Use an existing third-party UI library that provides prebuilt components such as cards, form elements, loaders, various grid and section layouts, and modals, which we would then adjust to suit the needs of our design.
+
+This could include generally "full purpose" UI libraries such as [Semantic UI](https://semantic-ui.com/), [Tailwind UI](https://tailwindui.com/), [Bootstrap](https://getbootstrap.com/), [Bulma](https://bulma.io/), or [Pure.css](https://purecss.io/) designed to cover a wide variety of needs. These libraries follow various CSS architecture patterns or their own custom approach.
+
+Pros:
+
+- A lot of the work is done for us; less time spent creating custom styles.
+- Certain libraries provide us with the option to configure base styles and integrate with our build system.
+- Most have drop in React subsets that would allow us to utilize UI styling and corresponding markup as proper React components.
+- Most come with SASS/SCSS files to allow for configuration.
+
+Cons:
+
+- Won’t cover 100% of our design needs and possibly even less so than other options, which would lead to us needing to create custom styles or classes from scratch or _also_ use a utility library.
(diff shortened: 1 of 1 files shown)


docs/adr/0018-use-tailwind-with-custom-scss.md
```diff
@@ -0,0 +1,136 @@
...
+
+- While most libraries offer some level of configuration, many require working in SASS to override variables and maps, and the resulting output may not always be predictable. Additionally, SASS-level configuration forces us to use SASS variables, and yet in [CSS ADR part 1](https://github.com/mozilla/fxa/blob/master/docs/adr/0015-use-css-variables-and-scss.md), we determined we would like to avoid SASS variables in favor of CSS variables. Tailwind mitigates these inconveniences by offering a root-level JS configuration file which is a better option for organization, clarity, and ease-of-use. Furthermore, Tailwind's configuration options are vast and allow us to set our own incremental spacing, custom font-sizes, determine classname conventions, set custom breakpoints, etc. as we see fit to align with the Settings Redesign visual designs.
+- We can use CSS variables to share stored values with any additional custom SCSS when applicable and for future theming options.
+- Several engineers on our team have prior experience with Tailwind and like it.
+- Tailwind is [significantly more popular](https://www.npmtrends.com/tachyons-vs-tailwindcss-vs-basscss) than other options.
```
jackiemunroe, line 92: I would add the specific numbers here, since this link goes to a dynamic webpage. It is conceivable that in the future this may not be a true statement. I might consider something along the lines of:
``` At the time of this decision, Tailwind is significantly more popular (description of data + screenshot). This data was gathered using the following npm package download comparison tool (link to npm trends page)```

Comment (jodyheavener) [reply docs/adr/0018-use-tailwind-with-custom-scss.md]: updated!

Comment (jackiemunroe) [reply docs/adr/0018-use-tailwind-with-custom-scss.md]: Thanks!

**Comment by dannycoates** [comment]
I'd like to see some analysis regarding the expected maintenance burden for each option. For external packages in keeping up with releases and internally with inevitable design changes and the fight against bit-rot.

File: docs/adr/0018-use-tailwind-with-custom-scss.md [hunk 2]
```diff
@@ -0,0 +1,136 @@
...
+## Considered Options
+
+- Option A - Use an existing UI library or framework
+- Option B - Use an existing utility library
+- Option C - Develop our own hybrid component and utility framework
+- Option D - Custom SCSS with loose structure
+
+## Decision Outcome
+
+Chosen options: "Option B" with Tailwind CSS for majority styling, and implementation details from "Option D" when utility classes don't meet the entire need, because:
```
vbudhram, line 31: Would we actively discourage the use of `Option D`? I've found it useful for `!important` that the developer documents exactly why they had to introduce this and maybe things they tried. We could do something similar for this.

docs/adr/0018-use-tailwind-with-custom-scss.md
```diff
@@ -0,0 +1,136 @@
...
+- Likely has the highest learning curve out of all considered options. This would require developers to know what already exists for use, likely causing constant documentation checks or searching for what’s available. An IDE extension could possibly ease this burden slightly, depending on the popularity of the chosen library.
+- Frameworks like Bootstrap are much heavier than other options. They'll almost certainly come with features/components that we will not make use of, though this could possibly be alleviated with a tool like PurgeCSS.
+- Introduces at least one more dependency to our stack. In fact, because some components would be shared and some would be project-specific it's likely that this dependency would exist in multiple areas of the repo. Even more dependencies would be required if we opt to use custom React components provided by the library or framework due to needing to install corresponding TypeScript definitions.
+- As some of these libraries include JavaScript to make components fully functional (e.g. accordions, tabs) we would likely need to have the security team vet it.
+
+### Option B - Use an existing utility framework
```
6a68, line 64: This option seems like a library, not a framework.

jodyheavener: Haha @LZoog and I spent some time discussing this; I think her position was that it sounds like a library too. IMO these two terms have been muddied and are used so interchangeably these days it's tough to say exactly what's what. Your vote makes two, though, so I'm changing it!

docs/adr/0018-use-tailwind-with-custom-scss.md
jodyheavener: I think Option D (needing to use custom styles) will inevitably bleed into any of the chosen solutions at some point. While not true 100% of the time, !important often hints at a larger issue, so I definitely like the idea of explaining why you absolutely need to use it. Would it help if we outlined that and some additional "free-styling" guidance in the decision outcome?

LZoog: I think it should definitely be discouraged _if we have a Tailwind class that would apply the same style_. Some things are inevitable and don't need an explanation, like background-images set on devices & apps. I think we'll have a lot of elements that can be properly styled with only utility classes so I like the idea of a comment at the top of a set of custom styles. 🙂 If anything, it'll serve as a reminder for someone else going in to edit that file that there's Tailwind classes available and you shouldn't just go hacking away at the CSS. Jody, we could add this example when one of us grabs #5244 next sprint.

(I also _really_ dislike `!important`. 😅)

docs/adr/0018-use-tailwind-with-custom-scss.md
6a68: Heh, I did not mean to publish that, thought it was part of a draft review 🤦

LZoog:
> I’d be interested to see how you would style one or two example components using the different approaches, if there’s time. No worries if not.

Yeah, this would be neat to see. I feel to do this with an accurate representation of various approaches, we'd want to kind of map out examples of not just specific libraries would look like for us, but what various CSS architectures would end up as. It's ideal IMO to select the architecture that makes sense for the project, and then evaluate the available libraries that use that methodology. This ADR looked to kind of simplify that, and took a "utility library (Atomic CSS) vs non-utility (everything else)" approach, because the non-utility options were, in Jody's and my opinion, not the right fit for our team, where we like to move quickly with our CSS. Having a bunch of single-purpose classes feels more maintainable for our team regarding a learning curve, speed in development, and time spent during PR reviews.

> I’m particularly curious what you have in mind about mixing Tailwind with our own custom styles.

Ideally, we'll only use Tailwind classes. I think this will work for almost all of our components, but we can't guarantee they'll cover _every_ use case so we want some type of fallback if we need it. One place in particular I'm thinking of is when we set `background-image`s for the icons in our "Devices & apps" section, we base this off of the `data-name` or `data-os`. It may be possible to create custom Tailwind classes for these... But we still may need custom styles for RTL languages or other one-offs. Essentially the goal will be to keep any custom SCSS at an absolute minimum, though.

> I personally don’t have issues with the way the FxA styles work now, but I don’t have experience maintaining or refactoring the styling specifically.

Yeah, it's gotten a little gnarly. Our class names vary greatly, we have a lot of competing selectors (especially around our buttons), we have a lot of inconsistency with our line-heights, font-sizes, and padding, and there's a few places we could really DRY it up. I think Jody and I were able to capture it with this comment: "A lack of convention can lead to unwieldy selector depth, competing rule sets/use of `!important`, a lack of DRY code, inconsistent naming and styling, hard-to-manage SASS `@extends`, and poor readability."

> Inlining styles increases the cost of making small, widespread changes, like mass-updating button or link colors, or adjusting layouts across all components

They're not technically inlined styles since they're not scoped to the component, but I totally get what you mean referring to them as that. I'm not sure what you mean by increasing the cost of mass-updating something though - actually, this makes it incredibly easy! 🙂 We would just need to change the value in one spot, the Tailwind config file.

> I’m not sure how much code is really in Tailwind, but FxA has already managed to outlast a couple generations of front-end libraries/frameworks...

Yeah, this is definitely a factor. Compared to a lot of other UI libraries (especially other Atomic CSS considerations), Tailwind seems to be popular, so it is the hope of course that it continues to stick around. Since it does a lot of code generation for us based on a config file, I can see it being a good long-term solution for us. CSS-in-JS definitely seems to be "the hotness" right now, but we outlined in the first CSS ADR why we've chosen not to go in that direction. Worst case scenario, though... It would be WAY less effort to refactor FxA's CSS at a later time due to whatever reason than say, moving away from React.

Thanks for sharing your thoughts! 👍

6a68:
> > Inlining styles increases the cost of making small, widespread changes, like mass-updating button or link colors, or adjusting layouts across all components
> 
> They're not technically inlined styles since they're not scoped to the component, but I totally get what you mean referring to them as that. I'm not sure what you mean by increasing the cost of mass-updating something though - actually, this makes it incredibly easy! 🙂 We would just need to change the value in one spot, the Tailwind config file.

TIL! I had totally missed that when I was skimming articles about Tailwind :-)

dannycoates:
@6a68 @vbudhram Send uses tailwind so if you'd like an example of what the "D" part of it looks like for things the utility classes don't cover have a look here: https://github.com/mozilla/send/blob/master/app/main.css

I've definitely learned some things since then and would do a few things differently, mainly around how colors are applied, but we'll be able to apply that learning in fxa. Since we're using css variables I think this will be a good approach: https://github.com/adamwathan/theming-tailwind-demo

PR #5253, opened 2020-05-08
docs(adr): adr 16 for auth settings

bbangert:
Because:

* During implementation of auth for graphql the effort required to use
  OAuth2 was substantially greater than anticipated.

This commit:

* Add's ADR 16 with a new proposed auth scheme for the settings
  redesign.

### DIFF
--- a/docs/adr/0016-auth-for-settings-redesign.md
+++ b/docs/adr/0016-auth-for-settings-redesign.md
@@ -1,14 +1,14 @@
 # Authentication in Settings Redesign
 
-- Status: superseded by [ADR-0016](0016-auth-for-settings-redesign.md)
+- Status: proposed
 - Deciders: Les Orchard, Lauren Zugai, Ben Bangert, Ryan Kelly
-- Date: 2020-05-08
+- Date: 2020-04-15
 
 Issue: https://jira.mozilla.com/browse/FXA-1559
 
 ## Context and Problem Statement
 
-The Settings Redesign app needs to query & mutate the same protected user data as the existing settings app hosted on content-server. This will require some form of authentication & authorization to manage that data.
+The Settings Redesign app needs to query & mutate the same protected user data as the existing settings app hosted on content-server. This will require some form of authentication & authorization to manage that data. [ADR-0014] included several assumptions that turned out to be inaccurate during implementation efforts. Using OAuth2 tokens in the auth-server for the API calls needed is a larger effort than anticipated and will not meet time/work constraints for the Settings Redesign. This ADR includes a new option that is proposed to meet these decision drivers.
 
 ## Decision Drivers
 
@@ -19,6 +19,7 @@ The Settings Redesign app needs to query & mutate the same protected user data a
 
 ## Considered Options
 
+- Authenticate with sessionToken, then use a unique Bearer token for auth-server APIs.
 - Authenticate via OAuth2 to use auth-server APIs
 - Authenticate via OAuth2 to use DB directly
 - Reuse existing sessionToken on content-server with auth-server APIs
@@ -26,22 +27,35 @@ The Settings Redesign app needs to query & mutate the same protected user data a
 
 ## Decision Outcome
 
-Chosen option: "Authenticate via OAuth2 to use auth-server APIs", because:
+Chosen option: "Authenticate with sessionToken, then use a unique Bearer token for auth-server APIs", because:
 
 - It requires minimal changes to existing auth-server implementation and infrastructure.
 - It relies on an authentication mechanism with relatively well-known security properties.
 
 ### Positive Consequences
 
-- The OAuth2 access token mechanism is better suited to this purpose than a novel scheme to share session token credentials.
-- The new Settings Redesign app can use existing auth-server APIs with minimal modifications to accept scoped OAuth2 access tokens. This can constrain most of the novelty in the project to the redesigned & reimplemented settings UX.
+- We do not have to re-implement substantial logic in the auth-server for using OAuth2 tokens.
 
 ### Negative Consequences
 
-- We will need to modify auth strategies in auth-server APIs, taking care not to affect existing usage in the production settings app.
+- We will need to modify auth strategies in auth-server APIs.
 - We don't have an entirely greenfield project, which could be a bit of a drag. But, rediscovering lessons learned in existing code can also be a drag.
+- We will be using a unique Bearer token between the graphql API and the auth-server that purports to represent the original Hawk request.
 
 ## Pros and Cons of the Options
+
+### Authenticate with sessionToken, then use a unique Bearer token for auth-server APIs.
+
+While we could like to someday deprecate the session token mechanism, in this case a substantial portion of the API's used in auth-server assume a sessionToken and the data returned when authenticating it will be used. The session token includes additional relevant authentication method information that is currently not present on the OAuth2 tokens which further complicates adapting the auth-server methods to use an OAuth 2 token.
+
+- Pros:
+  - Can be done with seamless UX - user visits new settings without confirmation
+  - Minimal work involved as the new auth-server auth strategy can lookup originating sessionTokenID.
+- Neutral:
+  - GraphQL server needs to proxy auth-server APIs
+- Cons:
+  - More systems that continue to use sessionTokens which we would like to deprecate.
+  - We use Hawk request signing with the session token credentials for authentication. Passing these credentials from the frontend to a GQL backend essentially defeats the purpose of Hawk, since the scheme is in part meant to minimize the transmission of credentials over the network.
 
 ### Authenticate via OAuth2 to use auth-server APIs
 
@@ -49,12 +63,13 @@ We would like to someday deprecate the session token mechanism altogether. This 
 
 - Pros:
   - Security properties of OAuth2 authn / authz are known and not entirely novel
-  - Auth-server APIs can be reused with small changes
 - Neutral:
   - GraphQL server needs to proxy auth-server APIs
 - Cons:
   - APIs on auth-server need auth strategy changes to accept OAuth tokens
   - May initially require a hack similar to payments-server where a pre-generated access token is passed along
+  - Many APIs built against assumption of data loaded with the sessionToken which will incur substantial work in the auth-server to accommodate.
+  - OAuth2 token is missing information relevant to the users active session that is present in the sessionToken.
 
 ### Authenticate via OAuth2 to use DB directly
 
@@ -68,7 +83,8 @@ We would like to someday deprecate the session token mechanism altogether. We wo
 - Cons:
   - Auth UX flow will require user to confirm permissions
   - Skipping auth-server requires reinventing many things
-  - New implementation will require security reviews of auth-server reimplementations
+  - New implementation will require security reviews of auth-server re-implementations
+  - OAuth2 token is missing information relevant to the users active session that is present in the sessionToken.
 
 ### Reuse existing sessionToken on content-server with auth-server APIs
 
@@ -94,3 +110,5 @@ We could make a complete break from everything but the database and start fresh.
   - Novel security properties of new login session system
   - Need to maintain storage for persistent sessions - e.g. a database or redistribution
   - Skipping auth-server requires reinventing many auth-server things
+
+[adr-0014]: 0014-auth-for-settings-redesign.md


docs/adr/0016-auth-for-settings-redesign.md
```diff
@@ -0,0 +1,114 @@
...
+While we could like to someday deprecate the session token mechanism, in this case a substantial portion of the API's used in auth-server assume a sessionToken and the data returned when authenticating it will be used. The session token includes additional relevant authentication method information that is currently not present on the OAuth2 tokens which further complicates adapting the auth-server methods to use an OAuth 2 token.
+
+- Pros:
+  - Can be done with seamless UX - user visits new settings without confirmation
+  - Minimal work involved as the new auth-server auth strategy can lookup originating sessionTokenID.
+- Neutral:
+  - GraphQL server needs to proxy auth-server APIs
+- Cons:
+  - More systems that continue to use sessionTokens which we would like to deprecate.
+  - We use Hawk request signing with the session token credentials for authentication. Passing these credentials from the frontend to a GQL backend essentially defeats the purpose of Hawk, since the scheme is in part meant to minimize the transmission of credentials over the network.
```
lmorchard, line 58: This seems like a copy from "Reuse existing sessionToken". Should be different here, if I understand the new proposed mechanism:

We wouldn't send the Hawk sessionToken credentials over the network. Instead, we'd generate and send a unique Bearer token that's one step removed from those original credentials. 

That should mean we're using a different mechanism, not defeating Hawk in general. Maybe a subtle difference, but worth highlighting?

docs/adr/0016-auth-for-settings-redesign.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Status: proposed
+- Deciders: Les Orchard, Lauren Zugai, Ben Bangert, Ryan Kelly
+- Date: 2020-04-15
+
+Issue: https://jira.mozilla.com/browse/FXA-1559
+
+## Context and Problem Statement
+
+The Settings Redesign app needs to query & mutate the same protected user data as the existing settings app hosted on content-server. This will require some form of authentication & authorization to manage that data. [ADR-0014] included several assumptions that turned out to be inaccurate during implementation efforts. Using OAuth2 tokens in the auth-server for the API calls needed is a larger effort than anticipated and will not meet time/work constraints for the Settings Redesign. This ADR includes a new option that is proposed to meet these decision drivers.
```
lmorchard, line 11: > [ADR-0014] included several assumptions that turned out to be inaccurate during implementation efforts. 

Which assumptions? [Is it mainly this stuff about the levels of auth?](https://github.com/mozilla/fxa/pull/4931#discussion_r411828476) If so, would be helpful to link to that for context in how this decision was made for later.

docs/adr/0016-auth-for-settings-redesign.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Decision Drivers
+
+- Smooth UX
+- Security
+- Development velocity
+- Ease of integration
+
+## Considered Options
```
lmorchard, line 20: Looks like these are a copy of the superseded ADR. Might help to just discuss the options with updated discoveries. There aren't new details surfaced about the previously rejected ones, so they don't seem to factor into the decision here.

docs/adr/0016-auth-for-settings-redesign.md
```diff
@@ -0,0 +1,114 @@
+# Authentication in Settings Redesign
```
lmorchard, line 1: Since we made a decision, and now we're making a new decision, we probably want a new title here - e.g. Switch from OAuth2 to unique Bearer token in settings redesign auth.

dannycoates:
I feel like I'm missing something obvious but I'm just going to roll with it.

It seems to me that if the answer to the following question is _yes_ then we can avoid any new auth scheme on the fxa-auth-server:

> Should we allow the browser to send the sessionToken to the GQL server?

#### Assuming the answer is yes, here's my logic:

- The new settings app has the same access to localstorage as the existing settings
- GQL server is trusted and accessed over https
- New settings can authenticate to gql-server however it wants and transmit the sessionToken as a request parameter for queries that need to make auth-server requests
- With the sessionToken the GQL server can use HAWK to make requests to auth-server the same as any browser client
    - security side-note: The GQL server won't make any request to auth-server that would return a keyFetchToken or key data

There would be no changes to auth-server if this is the case.

#### If the answer is no, I have more questions:

- what is wrong with you and why are you making this so hard? (j/k)
- what specifically is wrong with sharing the sessionToken with the GQL server?
  - not a question: I do not accept "defeating the purpose of HAWK" as a valid reason 😝
  - can we address these concerns without introducing a new auth-server auth scheme?

lmorchard:
> It seems to me that if the answer to the following question is yes then we can avoid any new auth scheme on the fxa-auth-server

Yeah, this was option ["Reuse existing sessionToken on content-server with auth-server APIs"](https://github.com/mozilla/fxa/blob/master/docs/adr/0014-auth-for-settings-redesign.md#reuse-existing-sessiontoken-on-content-server-with-auth-server-apis) in the previous ADR. It would be a lot simpler, but we essentially [decided that defeating Hawk was a security showstopper for that option](https://github.com/mozilla/fxa/pull/4931#discussion_r409218596).
 
> not a question: I do not accept "defeating the purpose of HAWK" (alone) as a valid reason 

Why do you not accept it? (i.e. what did we miss?)

I'm honestly asking, because defeating Hawk seems like a showstopper. Otherwise, why are we using Hawk at all?

If we're okay with that (maybe we are?) then, yeah, a lot of things are simple. Are there any other showstopper security concerns with doing things this way?

docs/adr/0016-auth-for-settings-redesign.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Status: proposed
+- Deciders: Les Orchard, Lauren Zugai, Ben Bangert, Ryan Kelly
+- Date: 2020-04-15
```
LZoog, line 5: I think @dannycoates should be added here. Also, the date should be `05` not `04`).

(Also not entirely sure I should be given any credit here, I don't know that I added anything meaningful to this conversation. 😛)

docs/adr/0016-auth-for-settings-redesign.md
LZoog: Hmm yeah I agree, this ADR could just be the old approach from ADR 14 vs the new proposed one with pros/cons of each direction, and the decision drivers can be the reasons for the new approach.

LZoog:
I think there's a little point of confusion on my end with the conversation above, trying to wrap my head around all of this...

> Should we allow the browser to send the sessionToken to the GQL server?

> It would be a lot simpler, but we essentially decided that defeating Hawk was a security showstopper for that option.

So I thought we could send send hawk-authenticated requests up to the GQL server from the client, and the GQL server would send this over to the auth server for that unique bearer token. Isn't this using Hawk while passing around the sessionToken or, is this only possible if we pass along the `sessionTokenId` and not the entire sessionToken, or is this (taken from one of Les's comments) "using a different mechanism, not defeating Hawk in general"?

lmorchard:
> So I thought we could send send hawk-authenticated requests up to the GQL server from the client, and the GQL server would send this over to the auth server for that unique bearer token. 

In the Bearer token scenario, the key thing is that it's the *client* which asks for the Bearer token from auth-server. GQL server never receives the sessionToken in this scheme, only the Bearer token. I think the flow is this: 

* Client asks auth-server for a Bearer token - using sessionToken & Hawk like normal. 
* The Bearer token is a new, different credential than sessionToken.
* Client gives Bearer token to GQL server in a request.
* GQL server uses the Bearer token (not sessionToken) for requests to auth-server.

> Isn't this using Hawk while passing around the sessionToken or, is this only possible if we pass along the sessionTokenId and not the entire sessionToken, or is this (taken from one of Les's comments) "using a different mechanism, not defeating Hawk in general"?

The Bearer token is a different mechanism than Hawk.

As I understand Hawk, it's designed so that the complete sessionToken only gets sent from server to client on login. The client builds *request signatures* based on the sessionToken for each request, but the sessionToken itself never gets transmitted from client to server.

To allow the GQL server to make Hawk-authenticated requests using sessionToken directly, we'd have to send the complete sessionToken down to the GQL server from the client. The GQL server would then do its own signed requests to auth-server.

But, it's the sending of the sessionToken credentials from a client to a server that Hawk was designed to prevent - i.e. so that's "breaking Hawk"

That said, if we're okay with "breaking Hawk", it's super simple to just share the complete sessionToken over the wire with GQL server and let it make the normal Hawk request signatures that auth-server already expects.

dannycoates:
> Why do you not accept it? (i.e. what did we miss that makes it okay?)

_disclaimer: this is my personal recollection of events from nearly 7 years ago_

Long ago, the precursor to [onepw](https://github.com/mozilla/fxa-auth-server/wiki/onepw-protocol) had 2-phase signup/login that would establish a secret key that would be used to encrypt/decrypt the sessionToken so that it was never in the clear on the wire. It was secure enough that you could login over http and still be safe. We'd still use https, but defense in depth, yadda yadda. Once we had the sessionToken we used Hawk for api requests to maintain that level of security. I implemented this protocol and it worked as designed. _(pats self on the back)_ The problem at the time was we wanted to support low-end phones on slow networks and the extra round trips meant we were too slow; bad UX. Instead of scrapping the whole thing we modified the protocol to the one we use today. The sessionToken is sent "in the clear" (though not really because it's over https) in the signup/login response, and the "secure even over http" dream was gone. We kept Hawk _because we already had it_ and it wasn't hurting anything, except maybe to annoy us with timestamp errors... _another story_.

The rest here is my personal opinion. The auth-server api would be equally secure today (with one possible caveat that I'll address later) if the sessionToken were used as a simple bearer token instead of as the credentials for Hawk signatures. Hawk is designed for authentication over http. Quoting from the [design goals](https://hapi.dev/module/hawk/) of Hawk:

* simplify and improve HTTP authentication for services that are unwilling or unable to deploy TLS for all resources,
* secure credentials against leakage (e.g., when the client uses some form of dynamic configuration to determine where to send an authenticated request), and
* avoid the exposure of credentials sent to a malicious server over an unauthenticated secure channel due to client failure to validate the server's identity as part of its TLS handshake.

Literally none of those apply to fxa. What about other aspects of Hawk? Replay protection, [we ignore](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/server.js#L87-L93). Same with timestamps. The one thing we get is request authenticity, i.e. we know it wasn't modified between client and server. #hashtagsarcasticwowemoji

The caveat. Our beloved sessionToken isn't used directly as the credentials input for Hawk. Instead we use it to [derive](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/auth/hawk.ts#L9-L37) an `id`, `key` and `bundleKey`. `id` and `key` are used for the Hawk signature, but `bundleKey` is used as an additional layer of encryption when retrieving the oh so precious **keys** from the server. We probably don't want the GQL server to have that ability. But guess what? I lied. There's a separate token for that called `keyFetchToken` so the `bundleKey` is useless and ignored when dealing with sessionTokens. I'm not mentioning this to troll anyone, it's just not obvious when looking at the code what's going on with `bundleKey` and when/why it's important. The lie just set up the narrative. No caveat, just a fun diversion.

The real concern in my opinion, and it's a small one in this context, is that sessionTokens are long lived. Too long if your ask me. So if one leaked somehow it might be a problem. I think it's more likely for a sessionToken to leak off of the client device than on the wire or from one of our servers. And I'm not worried about either tbh.

So, "defeats the purpose of Hawk".... friends, it's time to move on. ❤️

bbangert:
I'm closing this and passing the mantle on it to @dannycoates, @rfk, and @lmorchard . I have insufficient knowledge of any of the schemes involved to make meaningful contributions to these discussions.

lmorchard:
> I'm closing this and passing the mantle on it

@bbangert Is there an issue open for this, so we don't lose the work? There's still a decision to be made & documented, right? Might be better to keep this PR open and revise

rfk:
> So, "defeats the purpose of Hawk".... friends, it's time to move on. ❤️

Danny and I met to discuss this over zoom. I remain of the opinion that "defeating hawk" is the wrong way to approach solving the problem posed in this ADR. But if we have independently decided that Hawk is providing no value and should be removed, then I think I'm OK with sending the sessionToken and defeating hawk *as an intermediate state on the way to a Hawk-free world*.

Danny is going to draft a separate "we should get rid of Hawk" ADR for consideration.

PR #5262, opened 2020-05-09
feat(fx-content-server): add autocomplete annotations to change password

ChristianMurphy:
adds https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#Allowing_autocomplete

Which gives hints to password managers, like Firefox Lockwise, on where to enter the existing password, and where to suggest a new password.

---

also runs `npm run format` on the file, it appears to have been a while since the format task was last applied to the file.
https://github.com/mozilla/fxa/pull/5262/files?w=1 shows a clearer diff.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
+++ b/packages/fxa-content-server/app/scripts/views/settings/change_password.jsx
@@ -20,223 +20,226 @@ const translator = new Translator();
 const t = msg => translator.get(msg);
 
 function ChangePassword(props) {
- return (
-  <div id="change-password" className="settings-unit">
-   <ChangePasswordHeader />
-   <ChangePasswordForm
-      account={props.account}
-      submit={props.submit}
-      displayError={props.displayError}
-      isPanelOpen={props.isPanelOpen}
-      showValidationError={props.showValidationError}
-   />
-  </div>
- );
+  return (
+    <div id="change-password" className="settings-unit">
+      <ChangePasswordHeader />
+      <ChangePasswordForm
+        account={props.account}
+        submit={props.submit}
+        displayError={props.displayError}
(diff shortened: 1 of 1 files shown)


LZoog:
Hey @ChristianMurphy, thank you so much for putting in a PR!

I ran this by the FxA design lead and we like this idea but unfortunately we need the user to be able to see the password strength balloon at all breakpoints and this hides just that in mobile:

auto-complete

... This is something we could toggle on/off by device size, but it'd be a lot of trouble since this element is one of the few lone React elements inside of our `fxa-content-server`. We're actually working on redesigning some FxA content with this change password flow included, so I don't think it's worth the effort at the moment since it will be replaced later.

However, the auto complete on the current PW element seems fine since it doesn't hide any messaging. I'm not sure why CI is failing, will you remove `autocomplete="new-password"` on L205 and L221, leaving `autocomplete="current-password"` on L186, and try squashing the commit/rebasing against master?

Thanks!

ChristianMurphy:
Thanks for the feedback @LZoog!
That makes sense.

Looking forward to seeing the redesign!
In the mean time, removed the `new-password` attributes, and rebased.

LZoog:
@ChristianMurphy I apologize for not realizing your commit isn't [GPG signed](https://help.github.com/en/github/authenticating-to-github/managing-commit-signature-verification) when reviewing your PR the first time around, but it's a requirement before we can merge it. 😕 Your PR is approved though, I can merge it whenever you've got that `verified` text next to your commit!

Glad you are looking forward to the redesign, I am as well! :D & Thanks for circling back.

ChristianMurphy:
Whoops, missed that on the rebase, updated with the GPG signature.

PR #5275, opened 2020-05-11
fix(bug): small fixes for id_token_hint

6a68:
Grab bag of small fixes. Biggest change is allowing ID tokens to be used as id_token_hint for up to 7 days after being issued. Also tweaked response schema for id-token-verify auth server route, added a wrapper function to the right (new) fxa-js-client, and updated related tests.

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/jwt_id_token.js
+++ b/packages/fxa-auth-server/lib/oauth/jwt_id_token.js
@@ -7,6 +7,20 @@ const jwt = require('./jwt');
 const logger = require('./logging')('jwt_id_token');
 
 /**
+ * Verify the Expiration Time ('exp') claim value from an ID Token.
+ *
+ * This function was extracted into a helper to ease unit testing,
+ * because jsonwebtoken won't sign a token with an invalid data type
+ * for the 'exp' claim.
+ *
+ * @param {Any} exp
+ * @returns {Boolean} whether the value is a valid number
+ */
+exports._isValidExp = exp => {
+  return typeof exp === 'number' && !isNaN(exp);
+};
+
+/**
  * Verify an OIDC ID Token, following the flow in 3.1.3.7 of the spec:
  * https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
  *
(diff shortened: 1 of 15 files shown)


packages/fxa-auth-server/lib/oauth/jwt_id_token.js
```diff
@@ -97,10 +110,17 @@ exports.verify = async function verify(idToken, clientId) {
...
   //    Claim.
   //
+  // Note that we optionally allow for expired tokens via the
+  // `expiryGracePeriod` argument.
+  //
   // Note that this time is specified to be in seconds, not milliseconds.
   const currentTime = Math.round(Date.now() / 1000);
-  if (claims.exp < currentTime) {
-    logger.debug('verify.error.exp', { exp: claims.exp, currentTime });
+  if (claims.exp + expiryGracePeriod < currentTime) {
```
rfk, line 118: Be careful here, the `ignoreExpiration` flag causes jsonwebtoken to skip *all* validation of the `exp` field, including checking whether it's a number. That's probably OK since we've already checked the signature and we'd never issue a malformed JWT, but it's worth being aware of.

packages/fxa-content-server/app/scripts/lib/constants.js
```diff
@@ -162,4 +162,8 @@ module.exports = {
   // This is compared against all secondary email
   // records, both verified and unverified
   MAX_SECONDARY_EMAILS: 3,
+
+  // Allow ID Tokens used as the id_token_hint argument in a prompt=none
+  // request to be this many seconds past their expiration.
+  ID_TOKEN_HINT_GRACE_PERIOD: 60 * 60 * 24 * 7,
```
vbudhram, line 168: Was this default specified in the spec? A weeks seems really long to me.

6a68: I didn't see any discussion of specific times in the spec. The alternative here is for RPs to pass the user's email as a `login_hint`, and we do allow that for a controlled set of RPs, so it seemed reasonable to be liberal with the staleness of the token used as the `id_token_hint`. @rfk, any thoughts on a good upper limit?

packages/fxa-auth-server/lib/oauth/jwt_id_token.js
6a68: ooh, good point, I had missed that. I'll add a check that the 'exp' claim value is present, and is a number no more than 5 minutes in the future.

packages/fxa-content-server/app/scripts/lib/constants.js
rfk: I don't think there's any harm in accepting an arbitrary lifetime here. If I signed in to AMO, went away for two months, came back to AMO and had to re-authenticate, it seems entirely reasonable for AMO to send through the two-month-old id_token and for us to accept it as a hint for which user to sign in.

I don't think we're making any security decisions based on the data in this token (yet), are we? We're using it for the single bit of information of "has the user signed in to this RP in the past?".

diox: FWIW AMO doesn't store these tokens. We ask the `/v1/token` endpoint to give us one from `code` + `client_id` + `client_secret` when the user comes back from FxA at login (thereby verifying that the `code` is correct), and for the case we want to force 2FA on that user, immediately redirect the user back to FxA with the token we just got as `id_token_hint` so that they don't have to re-authenticate. So if `/v1/token` gives us fresh tokens, we have no reason to ever submit an old one.

6a68: @rfk That was my read of the spec as well, thanks for weighing in. Probably will leave it as-is for now, I guess 🤷‍♂️ 

> I don't think we're making any security decisions based on the data in this token (yet), are we? We're using it for the single bit of information of "has the user signed in to this RP in the past?".

Yup. We don't do anything with this token beyond verify it's valid and check that, if there's currently a logged-in user, it matches the account in the token.

PR #5282, opened 2020-05-11
feat(settings | content): store flowId post-auth in localStorage and retrieve it in settings app

### LINKED ISSUE #5019: [metrics] Get flowId from content server
Get flowId from the content server so that flows can be logged in the new settings app.

Flow params from content server: https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/src/lib/flow-event.ts#L4:L8

Check out how we're doing this in the content server for payments.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1694)
┆Issue Number: FXA-1694

jodyheavener:
Closes #5019

The new Settings app needs to be able to retrieve the `flowId` for the current user.

For fxa-payments this is achieved by creating a redirect from the Content Server that appends the flowId as a query parameter, which Payments later picks up.

In the case of fxa-settings once you log in or sign up on the Content Server you might not immediately be redirected to the new Settings app, so instead of a query parameter I'm storing the value in `localStorage`, and because both fxa-content-server and fxa-settings are served from the same domain the Settings app is able to retrieve it.

Additionally, since both packages set and get values in localStorage I took this as an opportunity to move our `storage` lib and its dependents, `null-storage` and `url`, to fxa-shared. They are mostly verbatim copies from the content server, with the exception if null-storage and url becoming TS friendly.

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
@@ -10,6 +10,7 @@ import NavigateBehavior from '../behaviors/navigate';
 import ResumeTokenMixin from './resume-token-mixin';
 import VerificationMethods from '../../lib/verification-methods';
 import VerificationReasons from '../../lib/verification-reasons';
+import Storage from '@fxa-shared/dist/lib/storage';
 
 export default {
   dependsOn: [ResumeTokenMixin],
@@ -251,6 +252,11 @@ export default {
     // This event ties the signin success to a screen.
     // Currently, can be oauth, signin, signup, signin-unblock
     this.logViewEvent('signin.success');
+
+    // Now that the user is logged in store thier flowId in
+    // localStorage so it can be retrieved by fxa-settings
+    const { flowId } = this.metrics.getFlowEventMetadata();
+    Storage.factory('localStorage', this.window).set('flowId', flowId);
 
     const brokerMethod = this.afterSignInBrokerMethod || 'afterSignIn';
     const navigateData = this.afterSignInNavigateData || {};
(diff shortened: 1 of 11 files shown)


packages/fxa-content-server/app/scripts/views/mixins/signin-mixin.js
```diff
@@ -10,6 +10,7 @@ import NavigateBehavior from '../behaviors/navigate';
 import ResumeTokenMixin from './resume-token-mixin';
 import VerificationMethods from '../../lib/verification-methods';
 import VerificationReasons from '../../lib/verification-reasons';
+import Storage from '@fxa-shared/lib/storage';
```
jodyheavener, line 13: Note I'm using the fxa-shared version of this library and not the one available in the content server, for get/set consistency between the content and settings apps. One day we should point all storage lib references to fxa-shared.

packages/fxa-content-server/webpack.config.js
```diff
@@ -52,6 +52,8 @@ const webpackConfig = {
       'node_modules',
     ],
     alias: {
+      '@fxa-components': path.resolve(__dirname, '..', 'fxa-components'),
+      '@fxa-shared': path.resolve(__dirname, '..', 'fxa-shared'),
```
jodyheavener, line 56: These two are unrelated to the PR, but opens up the ability to import from these external packages. TSConfig is also reflected.

jodyheavener:
Pulling back to draft for now because this doesn't build.

chenba:
- We also need the flowBeginTime and deviceId.
- I just double checked: every time the app on content server loads, it starts a new flow.  So unfortunately this breaks what's considered a flow.

Sorry, but I don't think this approach is going to work.  ☹️

PR #5288, opened 2020-05-12
feat(qr): Add first version of device connected polling

### LINKED ISSUE #5239: Add long poll for detecting an added device after CAD QR
We should in theory be able to use the `/attached_clients` endpoint to detect when a user has logged into a new device. Once they log in, the QR code page will be redirected to a "Device connected" page.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1845)
┆Issue Number: FXA-1845

vbudhram:
Fixes #5239 

This PR adds the polling functionality to the CAD via QR code screen. 

* utilizes the [/account/devices](https://github.com/mozilla/fxa/blob/03c438651e0d2525b5e5dad0187ad08811d40c37/packages/fxa-auth-server/lib/routes/devices-and-sessions.js#L488) endpoint on the auth-server to do the polling
* Add more basic tests to views
* Fix the broken `Use sms` links
* Polling disabled when user is not signed into an account

To test
1. Run `npm start firefox` and sign into browser A with an account X
2. Goto [http://localhost:3030/post_verify/cad_qr/scan_code](http://localhost:3030/post_verify/cad_qr/scan_code)
3. Create a new profile via `npm start firefox` and sign into browser B with account X
4. Upon connecting to browser B, browser A will navigate to http://localhost:3030/post_verify/cad_qr/connected


packages/fxa-content-server/app/scripts/models/polls/device-connected.js
```diff
@@ -0,0 +1,100 @@
...
+  constructor(data, options = {}) {
+    super(options, data);
+
+    if (!options.account) {
+      throw new Error('options.account required');
+    }
+
+    this._account = options.account;
+    this._pollIntervalInMS =
+      options.pollIntervalInMS || VERIFICATION_POLL_IN_MS;
```
vladikoff, line 18: I wonder if it makes sense to add a new constant instead of reusing the `VERIFICATION_POLL_IN_MS`. The device call could end up expensive (it used to be in the past) and we might need to change that value

vbudhram: Updated!

PR #5297, opened 2020-05-13
feat(content-server): console warn if you're not using localhost

jodyheavener:
Open to a no on this. Since our [switch](https://github.com/mozilla/fxa/pull/4938) from `127.0.0.1` to `localhost` we've had a few instances where someone has been confused about why their dev environment isn't working as expected, and the culprit was related to the host name (e.g. creating a CSP violation).

This PR warns you in development if you're accessing the content server from a host name other than `localhost`.


dannycoates:
im a little bit 👎 on this but not strongly. I think most, if not all of us who were used to the loopback have learned about the change by now. Since the docs use localhost no new folks should have this problem, I hope.

PR #5303, opened 2020-05-13
feat(content + settings): ensure visiting beta/settings/ always has flow data

### LINKED ISSUE #5019: [metrics] Get flowId from content server

jodyheavener:
Closes #5019

Alternative approach to #5282.

Here's how:
- Set up minimal library in fxa-settings to parse flow data params, and redirect you to fxa-content-server to fetch new flow data if they're not present
  - The majority of the flow-event lib comes from the existing one in the payments server, with the exception that it can optionally redirect you to the content server for flow data is it's not present
- Set up new route in fxa-content-server (`/get_flow`) to retrieve a new set of flow data, and send you back to your original address

Here's how I'm testing this:

```
// Should redirect you to /beta/settings with Flow data params
http://localhost:3030/get_flow?redirect_to=/beta/settings

// Should do nothing (console error about no redirect_to)
http://localhost:3030/get_flow

// Should do nothing (console error about using an absolute address)
http://localhost:3030/get_flow?redirect_to=http://localhost:3000/beta/settings

// Should send you to /get_flow and then redirect you to /beta/settings with Flow data params
http://localhost:3030/beta/settings

// Should send you to /get_flow and redirect you to /beta/settings with Flow data params and the custom params
http://localhost:3030/beta/settings?extra=123

// Should not send you to /get_flow
http://localhost:3030/beta/settings?device_id=devide-id&flow_begin_time=1589378014871&flow_id=flow-id&extra=123
```

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/get_flow.js
+++ b/packages/fxa-content-server/app/scripts/views/get_flow.js
@@ -1,0 +1,77 @@
+/* eslint-disable camelcase */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Cocktail from '../lib/cocktail';
+import FlowEventsMixin from './mixins/flow-events-mixin';
+import Url from '../lib/url';
+import BaseView from './base';
+import Template from 'templates/redirect_loading.mustache';
+
+class GetFlowView extends BaseView {
+  mustAuth = true;
+  template = Template;
+
+  initialize() {
+    // Flow events need to be initialized before the navigation
+    // so the flow_id and flow_begin_time are propagated
+    this.initializeFlowEvents();
+  }
+
+  afterRender() {
+    const { redirect_to: redirectTo } = Url.searchParams(
+      this.window.location.search
+    );
+
+    if (!redirectTo) {
+      this.window.console.error(
+        'This page requires a `redirect_to` path parameter'
+      );
+      return;
+    }
+
+    // Do not allow redirects to external addresses by
+    // trying to parse it and expect an exception
+    try {
+      // eslint-disable-next-line no-new
+      new URL(redirectTo);
+      return this.window.console.error(
+        '`redirect_to` must not be an absolute address'
+      );
+    } catch (error) {
+      // noop
+    }
+
+    const {
+      deviceId,
+      flowBeginTime,
+      flowId,
+    } = this.metrics.getFlowEventMetadata();
+
+    let redirectPath = redirectTo;
+    let redirectParams = {};
+    if (redirectTo.includes('?')) {
+      redirectPath = redirectTo.slice(0, redirectTo.indexOf('?'));
+      redirectParams = Url.searchParams(redirectTo);
+    }
+
+    const queryString = Url.objToSearchString({
+      device_id: deviceId,
+      flow_begin_time: flowBeginTime,
+      flow_id: flowId,
+      ...redirectParams,
+    });
+
+    const url = `${redirectPath}${queryString}`;
+
+    // We need to `navigateAway` and not `navigate` here
+    // because we are leaving the main content server app
+    this.navigateAway(url);
+  }
+}
+
+Cocktail.mixin(GetFlowView, FlowEventsMixin);
+
+export default GetFlowView;
--- a/packages/fxa-content-server/app/tests/spec/views/get_flow.js
+++ b/packages/fxa-content-server/app/tests/spec/views/get_flow.js
@@ -1,0 +1,143 @@
+/* eslint-disable camelcase */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import $ from 'jquery';
+import Account from 'models/account';
+import { assert } from 'chai';
+import sinon from 'sinon';
+import User from 'models/user';
+import View from 'views/get_flow';
+import Notifier from 'lib/channels/notifier';
+import WindowMock from '../../mocks/window';
+import Url from '../../../scripts/lib/url';
+
+describe('views/get_flow', function() {
+  let account;
+  let user;
+  let view;
+
+  const flowData = {
+    deviceId: 'foo',
+    flowBeginTime: 987654,
+    flowId: 'h2k3v5',
+  };
+
+  function render() {
+    return view.render().then(() => view.afterVisible());
+  }
+
+  beforeEach(function() {
+    account = new Account();
+    user = new User();
+
+    view = new View({
+      user,
+      notifier: new Notifier(),
+      window: new WindowMock(),
+    });
+
+    sinon.stub(user, 'sessionStatus').resolves(account);
+    sinon.stub(view, 'getSignedInAccount').returns(account);
+    sinon.stub(view, 'initializeFlowEvents');
+    sinon.stub(view, 'navigateAway');
+    sinon.stub(view.metrics, 'getFlowEventMetadata').returns(flowData);
+
+    sinon.stub(view.window.console, 'error');
+  });
+
+  afterEach(function() {
+    $(view.el).remove();
+    view.window.console.error.restore();
+    view.metrics.getFlowEventMetadata.restore();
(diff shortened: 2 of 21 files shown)


packages/fxa-content-server/app/scripts/templates/get_flow.mustache
```diff
@@ -0,0 +1,9 @@
+<div id="main-content" class="card">
+  <header>
+    <h1>{{#t}}One moment{{/t}}</h1>
+  </header>
+
+  <section>
+    {{#t}}You should be redirected shortly...{{/t}}
+  </section>
+</div>
```
chenba, line 9: Perhaps rename https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/scripts/templates/subscriptions_redirect.mustache and share between the two views for consistency.

packages/fxa-content-server/app/scripts/views/get_flow.js
```diff
@@ -0,0 +1,73 @@
...
+      this.window.location.search
+    );
+
+    if (!redirectTo) {
+      console.error('This page requires a `redirect_to` path parameter');
+      return;
+    }
+
+    // Do not allow redirects to external addresses by
+    // trying to parse it and expect an exception
```
chenba, line 34: heh "expect an exception"

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -5,9 +5,18 @@
...
 import './index.scss';
 
-export const App = () => {
+type AppProps = {
+  queryParams: QueryParams;
+};
+
+export const App = ({ queryParams }: AppProps) => {
+  const flow = FlowEvents.init(queryParams, true);
+  console.log('Flow data', flow);
```
chenba, line 18: It's here because this is still a draft?

jodyheavener: Gone now. I had originally not set up anything beyond parsing the incoming params, but I decided to add the full amplitude lib setup.

packages/fxa-settings/src/lib/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface QueryParams {
+  device_id?: string;
+  flow_id?: string;
+  flow_begin_time?: number;
+}
+
+export interface GenericObject {
+  [propName: string]: any;
+}
```
chenba, line 9: I feel like "object" is already super overloaded.  Let's name it "hash"?

We could also add a type param so that it's possible narrow the value type, which in this case is string I think.

`Hash<T> {
  [key: string]: T;
}`

Something like that?

packages/fxa-content-server/app/scripts/views/get_flow.js
```diff
@@ -0,0 +1,73 @@
...
+import FlowEventsMixin from './mixins/flow-events-mixin';
+import Url from '../lib/url';
+import BaseView from './base';
+import Template from 'templates/redirect_loading.mustache';
+
+class GetFlowView extends BaseView {
+  mustAuth = true;
+  template = Template;
+
+  initialize(_options) {
```
chenba, line 17: Could leave out `_options` here.

packages/fxa-content-server/app/tests/spec/views/get_flow.js
```diff
@@ -0,0 +1,81 @@
...
+    user = new User();
+
+    view = new View({
+      user,
+      notifier: new Notifier(),
+      window: new WindowMock(),
+    });
+
+    sinon.stub(user, 'sessionStatus').callsFake(() => Promise.resolve(account));
+    sinon.stub(view, 'getSignedInAccount').callsFake(() => account);
```
chenba, line 43: This is probably copied from existing tests.  Just FYI, `sinon.stub().resolves()` and `sinon.stub().returns()` should work here.

packages/fxa-content-server/app/tests/spec/views/get_flow.js
```diff
@@ -0,0 +1,81 @@
...
+      assert.lengthOf(view.$('.redirect-loading'), 1);
+      sinon.assert.calledOnce(view.initializeFlowEvents);
+
+      sinon.assert.calledWithExactly(
+        view.navigateAway,
+        `${betaSettingsPath}${flowParamsString}`
+      );
+    });
+  });
+});
```
chenba, line 81: I only see a happy path test here.  The error scenarios and the extra query params handling are missing.

packages/fxa-settings/src/components/App/index.test.tsx
```diff
@@ -7,6 +7,40 @@ import { render } from '@testing-library/react';
...
+        queryParams: {
+          device_id: 'moz9000',
+          flow_begin_time: 9001,
+          flow_id: 'ipsoandfacto',
+        },
+      })}
+    />
+  );
+
+  expect(window.location.replace).not.toHaveBeenCalled();
```
chenba, line 45: Add a test to see if `FlowEvent.init` was called with the correct args?

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -5,9 +5,17 @@
...
+import FlowEvents from '../../lib/flow-event';
 import './index.scss';
 
-export const App = () => {
+type AppProps = {
+  queryParams: QueryParams;
+};
+
+export const App = ({ queryParams }: AppProps) => {
+  FlowEvents.init(queryParams, true);
```
chenba, line 17: If that second arg is always true then we could remove the param from `FlowEvent.init`.

packages/fxa-settings/src/lib/flow-event.ts
```diff
@@ -0,0 +1,86 @@
...
+      return window.location.replace(
+        `${window.location.origin}/get_flow?redirect_to=${encodeURIComponent(
+          redirectPath
+        )}`
+      );
+    }
+  }
+}
+
+export function logAmplitudeEvent(
```
chenba, line 51: I need some time tomorrow to look at this thoroughly.  This stuff looks copied from payments, _however_, it is now sending to content server.  The metrics endpoints on content and payments are not identical.

packages/fxa-settings/src/components/App/index.tsx
jodyheavener: In an ideal world this would be shared between Payments and Settings, which is why I had it as an option. But I don't think that's going to happen this PR, so I'll remove it.

packages/fxa-content-server/app/tests/spec/views/get_flow.js
jodyheavener: Updated.

packages/fxa-settings/src/lib/flow-event.ts
jodyheavener: Good point. I did some checking, and the resulting `eventData` object in `flow-events.ts` is the bare minimum structure needed to successfully deliver a metrics payload to the content server. I've only set up the values to get it to send; they still need to be properly set and validated, but if it's cool with you I don't think that should block this PR as we have a [few other issues](https://jira.mozilla.com/browse/FXA-1692) that should cover it.

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -5,9 +5,19 @@
...
 
-export const App = () => {
+type AppProps = {
+  queryParams: QueryParams;
+};
+
+export const App = ({ queryParams }: AppProps) => {
+  FlowEvents.init(queryParams);
+
+  FlowEvents.logAmplitudeEvent('bleep', 'bloop', { sup: 'hey' });
```
chenba, line 19: Is this testing code?

jodyheavener: Ahhh sorry, removing.

packages/fxa-settings/src/components/App/index.tsx
chenba: Can you file an issue (maybe outside of the Settings epic) for that?  I want that as well.

jodyheavener: Yes!

PR #5304, opened 2020-05-13
feat(content-server): Integrate interruptive surveys

### LINKED ISSUE #4971: Integrate survey component to content server
Content server is higher priority then payment server

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1661)
┆Issue Number: FXA-1661

meandavejustice:
- fixes #4971 
You will need to uncomment the survey object in `packages/fxa-content-server/server/lib/configuration.js`


packages/fxa-content-server/app/scripts/views/base.js
```diff
@@ -173,6 +173,7 @@ var BaseView = Backbone.View.extend({
     this.user = options.user;
     this.lang = options.lang;
     this.window = options.window || window;
+    this.surveys = options.config ? options.config.surveys : [];
```
chenba, line 176: I don't think we need to attach the surveys to all the views. IIUC the only place that need to examine all the survey is the module that decides if the user get a survey.

packages/fxa-components/Survey/index.tsx
```diff
@@ -5,7 +5,7 @@
 import React, { useState } from 'react';
 import { CSSTransition } from 'react-transition-group';
 
-import './index.scss';
+// import './index.scss';
```
chenba, line 8: Delete it if we don't need it.

packages/fxa-components/postcss.config.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
chenba, line 1: What's the purpose of this?

packages/fxa-content-server/app/scripts/views/survey.js
```diff
@@ -0,0 +1,115 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*eslint-disable no-unused-vars*/
+import React, { useState, useEffect } from 'react';
+/*eslint-enable no-unused-vars*/
+import ReactDOM from 'react-dom';
+import Cocktail from 'cocktail';
+import BaseView from './base';
+import SettingsPanelMixin from './mixins/settings-panel-mixin';
```
chenba, line 11: Are we only showing the surveys within settings?  I didn't get that impression from reading some of the docs.

I also can't tell what from the settings panel mixin is used in here.

packages/fxa-content-server/app/scripts/views/survey.js
```diff
@@ -0,0 +1,115 @@
...
+  const [showSurvey, setShowSurvey] = useState(true);
+  const [surveyComplete, setSurveyComplete] = useState(false);
+
+  useEffect(() => {
+    const handleIframeTask = CreateHandleIframeTask(() => {
+      setSurveyComplete(true);
+      setTimeout(() => {
+        setShowSurvey(false);
+      }, 300);
+    });
```
chenba, line 50: Are these always updated together?  That is, can `showSurvey` and `surveyComplete` both be true or both false?  If not, we can use a single state here I think.

packages/fxa-content-server/app/scripts/views/survey.js
```diff
@@ -0,0 +1,115 @@
...
+
+  openPanel() {
+    // force a re-render so that the input element is focused.
+    return this.render();
+  },
+
+  submit() {},
+
+  render() {
+    return new Promise(checkValidSurveys.bind(this)).then((surveys) => {
```
chenba, line 103: I don't understand why this has to be a promise.  Nothing in `checkValidSurveys` appears to be async.

packages/fxa-content-server/app/scripts/views/survey.js
```diff
@@ -0,0 +1,115 @@
...
+      const surveyURL = surveys[Math.floor((Math.random() * surveys.length))].url;
```
chenba, line 104: Can this logic be captured elsewhere?  Preferably the code to determine whether a user gets a survey and which survey is separated from the rendering code.

packages/fxa-content-server/app/scripts/views/survey.js
```diff
@@ -0,0 +1,115 @@
...
+      ReactDOM.render(<SurveyView {...{ surveyURL }} />, this.$el.get(0));
```
chenba, line 105: I believe `this.el` is the equivalent of `this.$el.get(0)`

packages/fxa-content-server/server/lib/csp/blocking.js
```diff
@@ -37,6 +37,7 @@ module.exports = function(config) {
   const PAIRING_SERVER_HTTP = PAIRING_SERVER_WEBSOCKET.replace(/^ws/, 'http');
   const SENTRY_SERVER = 'https://sentry.prod.mozaws.net';
   const SURVEY_GIZMO = 'https://qsurvey.mozilla.com';
+  const SURVEYS = config.get('surveys').map((s) => s.url).map(getOrigin);
```
chenba, line 40: I think this would produce an array of duplicate origins when the survey urls have the same origin.  I don't know if this will be an issue later when the array is used.

packages/fxa-components/postcss.config.js
meandavejustice: Silences an error

packages/fxa-content-server/app/scripts/views/survey.js
meandavejustice: This was just to get the survey rendering quickly. The method you have presented in https://github.com/mozilla/fxa/compare/master...chenba:1708-survey-targeting works better

packages/fxa-content-server/app/scripts/views/survey.js
meandavejustice: This only has a promise so that it follows the pattern of the other subpanel views, It should be removed once we move to the other rendering method

packages/fxa-content-server/app/scripts/views/survey.js
meandavejustice: Yes, it should be pulled out into a separate utility function. Along with all of the code for checking the cases to show the survey (lang, region, referrer, etc). There is an issue https://jira.mozilla.com/browse/FXA-1838 that should be address this coming sprint

packages/fxa-content-server/server/lib/csp/blocking.js
meandavejustice: Good point, should check for that

packages/fxa-components/Survey/index.tsx
meandavejustice: We want to have styles declared next to the components. However the css ADR isn't complete, I've spoken with @jodyheavener about this, and the plan is to have it ironed out during the settings work. Then we can loop back around and remove the survey.scss file in content server and use the single stylesheet here

packages/fxa-content-server/app/scripts/views/survey.js
meandavejustice: They are separate so that we can wait about 3 seconds before hiding the survey to give the user a chance to see the surveyComplete message

meandavejustice:
Started merging @chenba's changes from https://github.com/mozilla/fxa/compare/master...chenba:1708-survey-targeting into this. Probably won't be finished until this evening; I'm release owner for this week and have meetings from 1:30-4.

PR #5310, opened 2020-05-13
fix(bug): fix typescript error

6a68:
🤓 whoops, not sure how the tests missed that

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/auth/client.ts
+++ b/packages/fxa-content-server/app/scripts/lib/auth/client.ts
@@ -1116,7 +1116,7 @@ export default class AuthClient {
     clientId: string,
     expiryGracePeriod?: number
   ) {
-    const payload = {
+    const payload: any = {
       id_token: idToken,
       client_id: clientId,
     };


packages/fxa-content-server/app/scripts/lib/auth/client.ts
```diff
@@ -1116,7 +1116,7 @@ export default class AuthClient {
     clientId: string,
     expiryGracePeriod?: number
   ) {
-    const payload = {
+    const payload: any = {
```
chenba, line 1119: Err why do we need this?  TS should easily infer the type of a const.

6a68: You would think so, but apparently the shape of a const object is fixed in TS unless it's given the magic `any` shape.

See: `TS2339: Property 'expiry_grace_period' does not exist on type '{ id_token: string; client_id: string; }'.` from https://app.circleci.com/pipelines/github/mozilla/fxa/10427/workflows/a9e54f1b-6e7c-4bf6-80a4-edc66e36c009/jobs/167767/artifacts

6a68: See also https://stackoverflow.com/questions/12710905/how-do-i-dynamically-assign-properties-to-an-object-in-typescript

chenba: Oh heh, I didn't look at the source of the error.  I'm on TS's side here.

I suppose an `any` is ok here since we are not returning it out of the function.

bbangert: TypeScript will infer this correctly, however it will be with the most narrow type as it happens at declaration type when using `const`. Manually widening the type is needed in these cases, or using `any` to allow it to be anything (which is unfortunate). If we know what additional attributes, or shapes will be added, this could be widened to allow them with:
``
const payload: {[key: string]: string} = {....}
```

Or whatever shape it is. If it's a shape we use frequently, it'd be handy to define it as a type so we could just declare that's the shape this should be assumed to be.

See: https://github.com/microsoft/TypeScript-New-Handbook/blob/master/reference/Widening-and-Narrowing.md

The Programming TypeScript book has a section on this as well.

PR #5318, opened 2020-05-14
feat(settings): Add Modal and UnitRow components

### LINKED ISSUE #4902: [main] Create reusable unit row components
Settings > Profile and Settings > Security contain "rows" that are visually similar. Create a component (such as `<UnitRow>`) that sets up shared HTML and associated styles/classes. It should be set up to conditionally display a button to cover the rows with a header and text but no button, like the "Primary email" row.

We'll probably want a second, separate component for "Secondary email", "Recovery key," and "Two-step authentication" as shown [in this image](https://user-images.githubusercontent.com/13018240/79361710-4b205780-7f0b-11ea-8879-ccdf1d1b4e55.png) from the [UI chart issue](https://jira.mozilla.com/browse/FXA-1544?focusedCommentId=54587&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-54587) since their layout is slightly different. The refresh buttons will be handled in FXA-1798.

Styles set should be responsive.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1607)
┆Issue Number: FXA-1607

### LINKED ISSUE #4918: [global] Add or create a modal component
Ideally, we should move `DialogMessage` from `fxa-payments-server` into the shared React directory and use this component for our modals (I would also suggest we change the name to `Modal`). However, payments has l10n implemented and we may not be able to share it until l10n is setup in the new Settings app.

At the very least, `DialogMessage` from payments can be used as a reference to create a new `Modal` component since it's already setup to exit on outside click, or "X" escape close. If a new component needs to be made, and assuming these largely look and act the same functionally, we should create a follow-up issue to use a shared component between payments and settings after l10n has been implemented.

It would be ideal to do https://jira.mozilla.com/browse/FXA-1691 in conjunction with this task if we have to create a new component because there are several a11y issues that would be worth addressing now, especially because its order in the DOM needs to be changed. However, at the very least, we can use payments modal functionality, and do FXA-1691 as a follow up.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1621)
┆Issue Number: FXA-1621

LZoog:
* Adds Modal component in fxa-settings,* example usage in rough draft Settings component
* Creates rough UnitRow and UnitRowWithImage components
* Moves two shared hooks, CloseIcon, and Portal from fxa-payments into fxa-components
* Changes payments to build with rescripts instead of react-scripts to use the shared fxa-react package
* Adds license to some files

The modal at this point is basically a copy of payments `DialogMessage` with L10N ripped out. I'm working on a follow up with a11y updates... they may or may not get shared. Might create a follow up issue for that, we'll see.

fixes #4918
fixes #4902

The "to do" comments added have issues already filed in the settings redesign epic. Styles will be added in a follow up once #5341 is merged.

Only thing lacking here is CI symlink problems. :P


packages/fxa-components/Portal/index.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.portal#top-bar { // scss-lint:disable IdSelector
+  left: 0;
+  position: fixed;
+  right: 0;
+  top: 0;
+  z-index: 9999;
+}
+
+.portal#dialogs { // scss-lint:disable IdSelector
```
jodyheavener, line 13: Don't need a class if we're using an ID

jodyheavener: Update - since this getting ripped out soon don't let this block you

packages/fxa-components/lib/hooks.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  useClickOutsideEffect,
+  useBooleanState,
+} from './hooks';
+
+describe('useBooleanStateResult', () => {
+  const Subject = ({ initialState = false }: { initialState?: boolean }) => {
+    const [state, setTrue, setFalse] = useBooleanState(initialState);
+    return (
+      <div>
+        <div data-testid="result">{state ? 'true' : 'false'}</div>
```
jodyheavener, line 19: Might could just `state.toString()`

packages/fxa-payments-server/src/components/AlertBar/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
+import Portal from '@fxa-components/Portal';;
```
jodyheavener, line 3: extra `;`

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -32,7 +32,7 @@ import AlertBar from '../../components/AlertBar';
 import DialogMessage from '../../components/DialogMessage';
 import FetchErrorDialogMessage from '../../components/FetchErrorDialogMessage';
 import { LoadingOverlay } from '../../components/LoadingOverlay';
-import CloseIcon from '../../components/CloseIcon';
+import { ReactComponent as CloseIcon } from '@fxa-components/images/close.svg';
```
jodyheavener, line 35: `ReactComponent` SVGs FTW

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import classNames from 'classnames';
+import Portal from '@fxa-components/Portal';
+import { ReactComponent as CloseIcon } from '@fxa-components/images/close.svg';
+
+type ModalProps = {
+  className?: string;
+  onDismiss?: Function;
+  onConfirm?: () => void;
+  children: ReactNode;
+  'data-testid'?: string;
```
jodyheavener, line 16: Does this need to be optional if it's assigned a default value in `Modal` instantiation?

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+ import React, { useCallback } from 'react';
+ import { useBooleanState } from '@fxa-components/lib/hooks';
+ import Modal from '../Modal';
+
+ export const UnitRow = () => {
+  const [modalRevealed, revealModal, hideModal] = useBooleanState();
+
+  const onConfirmClick = useCallback(() => {
+    console.log('confirmed - do the thing');
```
jodyheavener, line 13: Hey!

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+          onDismiss={hideModal}
+          onConfirm={onConfirmClick}
+        >
+          <h2>Are you sure?</h2>
+          <p>You can't undo the thing.</p>
+        </Modal>
+    )}
+  </div>
+  );
+};
```
jodyheavener, line 34: I think the indentation here got a little wonky

packages/fxa-components/Portal/index.scss
LZoog: Yeah, I'm going to address this in the follow up, it's leftover from payments so I just left it as-is.

packages/fxa-settings/src/components/Modal/index.tsx
LZoog: Yeah, I just removed it to double check and I get complaints.

packages/fxa-settings/src/components/UnitRow/index.tsx
jodyheavener: Do you want to keep this console log in place?

LZoog: When we actually have this hooked up, we'll have a function execute here, but right now we're not actually doing anything... I could take it out, but if someone pokes at this in the meantime, they wouldn't be able to tell the difference between dismissing and confirming.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -32,7 +32,7 @@ import AlertBar from '../../components/AlertBar';
 import DialogMessage from '../../components/DialogMessage';
 import FetchErrorDialogMessage from '../../components/FetchErrorDialogMessage';
 import { LoadingOverlay } from '../../components/LoadingOverlay';
-import CloseIcon from '../../components/CloseIcon';
+import { ReactComponent as CloseIcon } from '@fxa-react/images/close.svg';
```
jodyheavener, line 35: Hm, I see this is being re-used and is being converted to a React Component, but the SVG itself is not really React-related. I have concerns about storing them in a dedicated directory like this for fear that @fxa-react becomes another junk drawer for tangentially related files. Could this SVG live adjacent to the components' files?

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            Note: a secondary email won't restore your information—you'll need a{' '}
+            <a href="#recovery-key">recovery key</a> for that.
+          </p>
+
+          {modalRevealed && (
+            <Modal onDismiss={hideModal} onConfirm={onSecondaryEmailConfirm}>
+              <h2>Verify primary email first</h2>
+              <p>
+                Before you can add a secondary email, you must verify your
+                primary email. To do this, you'll need access to{' '}
```
jodyheavener, line 85: Is this forcing a space? I'm surprised the newline isn't creating one on its own

LZoog: Ah yeah, our linter inserts this sometimes when the line needs to fall to the next but we need a space, it's sprinkled all over payments too. The line break doesn't create a space.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
LZoog: If we had images located inside the component where they're being used, we wouldn't be able to cleanly reuse them, but I see your point about it not being "React". We were [exporting it as a React component](https://github.com/mozilla/fxa/pull/5318/files#diff-7efa0c29a1218ac2d1c78c1527dfce15L1-L2) (you may need to click "reveal" for `CloseIcon.tsx`) but if we did it that way, where would the SVG ultimately live if we need it in payments and settings? I guess we could also copy the contents of the SVG into an exported React component, but I personally prefer to keep the raw SVGs around. Let's rename to `fxa-shared-frontend`? :D (kidding, lol)

jodyheavener: Alright, approving on the back of your Slack chat about this being the best option. Let it be known that I'm only like... 90% on board, I still think we should find a better way to organize files that might be frontend-related but are not exclusive to React.

packages/fxa-settings/src/components/Settings/index.tsx
jodyheavener: Huh, that doesn't seem ideal. I don't know if we should be making funky accomodations to suit linting, esp if it's common. I wonder if we can set eslint or prettier or whatever to ignore text node lines inside JSX.

packages/fxa-settings/src/images/avatar-default.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="58" height="58" viewBox="0 0 58 58"><defs><linearGradient id="linear-gradient" x1="-3.82" y1="63.66" x2="56.35" y2="-0.64" gradientUnits="userSpaceOnUse"><stop offset="0.26" stop-color="#9059ff"/><stop offset="0.96" stop-color="#00b3f4"/></linearGradient></defs><title>default-profile</title><g id="Layer_2" data-name="Layer 2"><g id="Layer_1-2" data-name="Layer 1"><circle cx="28.8" cy="28.8" r="24.54" style="fill:#fbfbfe"/><path d="M28.8,0A28.8,28.8,0,1,0,57.6,28.8,28.79,28.79,0,0,0,28.8,0Zm0,8.23A12.34,12.34,0,1,1,16.46,20.57,12.33,12.33,0,0,1,28.8,8.23ZM45.73,40.34a20.39,20.39,0,0,1-33.86,0,4.15,4.15,0,0,1,.28-5.05C16.5,30.27,22.53,37,28.8,37s12.3-6.76,16.65-1.74A4.15,4.15,0,0,1,45.73,40.34Z" style="fill:url(#linear-gradient)"/></g></g></svg>
```
jodyheavener, line 1: Will this avatar eventually become a component of its own?

PR #5328, opened 2020-05-14
fix(settings): add missing emailVerified property to GQL query

### LINKED ISSUE #5284: [admin-panel] verified user in stage shows as `not verified`
I see the query goes in as 
```
select `accounts`.* from `accounts` 
inner join `emails` 
on `emails`.`uid` = `accounts`.`uid` 
where `emails`.`normalizedEmail` = 'jmorrison@mozilla.com'
```
and the db result set comes back with `emailVerified = 1` but the front end shows that account as `not verified`. 

The json response for https://fxa-admin-panel.stage.mozaws.net/graphql so no `emailVerified`.
```
{"data":{"accountByEmail":{"uid":"e87ad9e08db742d78a012366e56d951a",
"email":"jmorrison@mozilla.com",
"createdAt":1587063730813,"emailBounces":[],"__typename":"Account"}}}
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1869)
┆Issue Number: FXA-1869

jodyheavener:
Closes #5284 

Real simple, we were just forgetting to include the `emailVerified` response property when looking up an account.

I also added some basic UI tests.

### DIFF
--- a/packages/fxa-admin-panel/src/components/EmailBlocks/index.tsx
+++ b/packages/fxa-admin-panel/src/components/EmailBlocks/index.tsx
@@ -29,6 +29,7 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
       uid
       email
       createdAt
+      emailVerified
       emailBounces {
         email
         createdAt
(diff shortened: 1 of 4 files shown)


packages/fxa-admin-panel/src/components/EmailBlocks/index.tsx
```diff
@@ -29,6 +29,7 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
       uid
       email
       createdAt
+      emailVerified
```
jodyheavener, line 32: this is all we were missing

LZoog: 🤦‍♀️

PR #5331, opened 2020-05-14
refactor(shared components): Rename 'fxa-components' to 'fxa-react'

LZoog:
In #5318, we begin sharing React hooks and at least one SVG. Expanding beyond just `components`, it makes more sense to call this package `fxa-react` and move components into a `components` directory.

`fxa-shared-react` could have been slightly more descriptive, but importing from `@fxa-shared-react/components` feels slightly less sweet than from `@fxa-react/components`. 🤔

Opening this as a draft to see if CI complains. I'll move the components into a `components` directory in a separate commit.


dannycoates:
it may be better for me to do this rename as part of my yarn changes. lmk if you're blocked on this

LZoog:
@dannycoates I went ahead and pushed up moving the components in a `components/` dir as well so I was going to switch this from draft to PR once CI turns green, it would be ideal for me to have these changes in for my other draft PR since I'm pulling in hooks/images there but not sure I'd call it a blocker. Would this be fine to merge perhaps later today or would you rather we wait?

PR #5332, opened 2020-05-15
fix(metrics): Don't log `sendPush.success` message unless it actually succeeded.

rfk:
Because:

* The push logic was indiscriminately logging `sendPush.success` messages
  even in cases where the send had actually failed.

This commit:

* Restricts these messages to only be logged when the send succeeds.


rfk:
> Merging #5332 into master will increase coverage by 30.40%.

Cool story codecov...

PR #5336, opened 2020-05-15
Log push expired callbacks separately

eoger:
Looking at the [send tab metrics document](https://docs.google.com/document/d/1XxNHWuotwNpaOSzrdN3KPu0bFPe28SSxZZp4HNWezhM), I think it could be useful to figure out what's the split between push subscription missing and push subscription expired. I'm thinking of the "verified device" case where the push notification is "sent" before the new device had a chance to register the push callback. 
One thing we could also do is log the "reason" of the push (verified, command, etc) in the logs, wdyt?

### DIFF
--- a/packages/fxa-auth-server/lib/push.js
+++ b/packages/fxa-auth-server/lib/push.js
@@ -9,6 +9,7 @@ const base64url = require('base64url');
 const webpush = require('web-push');
 
 const ERR_NO_PUSH_CALLBACK = 'No Push Callback';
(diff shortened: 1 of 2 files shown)


rfk:
>  One thing we could also do is log the "reason" of the push (verified, command, etc) in the logs, wdyt?

We do kind of get this already, e.g. the `push.account_verify.push_callback_expired` in the new test case, would instead be a `push.command_received.push_callback_expired` if it was encountered during a device-commands call.

PR #5341, opened 2020-05-15
feat(settings): add foundational tailwind setup

### LINKED ISSUE #5244: [setup] Setup Tailwind config and any additional CSS foundation

jodyheavener:
Closes #5244

Here's what this PR does:

- Add Tailwind and PostCSS dependencies.
- Configure Tailwind with our [Protocol colors](https://protocol.mozilla.org/fundamentals/color.html) and our standard [breakpoint values](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/app/styles/_breakpoints.scss#L25-L30).
- Set up and configures PostCSS to compile Tailwind prior to building and running, and PurgeCSS to eliminate unused styles from `src/tailwind.css` in production builds.
- Import `tailwind.out.css` into our main `src/index.scss` file so it can be used freely across all TSX files under `src/`.
- Add some basic documentation in the README about how to do things, including some ideas about how to write custom styles.
- Add our fonts


packages/fxa-settings/tailwind.config.js
```diff
@@ -0,0 +1,181 @@
...
+        50: '#ffffcc',
+        100: '#ffff98',
+        200: '#ffea80',
+        300: '#ffd567',
+        400: '#ffbd4f',
+        500: '#ffa436',
+        600: '#e27f2e',
+        700: '#c45a27',
+        800: '#a7341f',
+        900: '#960e18',
```
LZoog, line 127: Do we want to go ahead and include all of these colors off the bat when we likely won't be using them, or should we start with what we need (blues, greys, whatever else)? I can see both sides - it doesn't hurt anything since we've got purgeCSS and this aligns us to colors from Protocol which Phil said we'd be using, but it's also some extra clutter and length in this file... what do you think? I like the `color-number` approach though.

packages/fxa-settings/README.md
```diff
@@ -14,7 +14,55 @@ You can import React components into this project. This is currently restricted
...
+</p>
+```
+
+It's fairly straightforward once you understand how size, scales and other sequential patterns work. The [Tailwind docs](https://tailwindcss.com/docs) are very helpful in understanding these concepts (and they have an excellent search), and there are plugins for various IDEs ([VS Code](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss), [IntelliJ IDEs](https://plugins.jetbrains.com/plugin/12074-tailwindcss), [Vim](https://github.com/iamcco/coc-tailwindcss)).
+
+Here are a few additional things to keep in mind when developing with Tailwind:
+
+- We overwrite some of the default values provided by Tailwind, such as color and breakpoints. Refer to [tailwind.config.js](./tailwind.config.js) to see how values vary.
+- You do not need to import any custom or built-in styles/dependencies to use Tailwind. Just write your class names.
+- When it comes to time to create a production build we use PostCSS and PurgeCSS to strip out unused styles. PurgeCSS will look through all of the TSX files and identify which class names to keep styles for. In order for this to work properly it's important to avoid dynamic class names:
```
LZoog, line 37: Probably good to make a note here that conditional class names work just fine, it's the interpolation/concat that can cause issues right? I did a quick search and found [this comment](https://www.viget.com/articles/a-better-approach-for-using-purgecss-with-tailwind/):

> For dynamic classes, try to avoid concatenating the name; if this is unwieldy, inline all the possible values beneath your logic in a comment, which keeps these whitelisted classes associated with the relevant code.


Could be another good note to have. 🤔

packages/fxa-settings/README.md
```diff
@@ -14,7 +14,55 @@ You can import React components into this project. This is currently restricted
...
+  <Button textColor="green-500">Hello, world</Button>;
+
+  // Good!
+  const Button = ({ textColor, children }) => {
+    return <button className={textColor}>{children}</button>;
+  };
+  <Button textColor="text-green-500">Hello, world</Button>;
+  ```
+
+- If you find a class name is getting stripped out erroneously in production builds, it can be [whitelisted](https://purgecss.com/whitelisting.html) in [tailwind.config.js](./tailwind.config.js) or in the stylesheet itself by placing `/* purgecss ignore */` directly above the selector. Please use sparingly.
```
LZoog, line 53: This should be pretty rare if we follow the conventions set. I suppose if someone has a weird case and wants to check it builds correctly they can just build the project and view the source to check that class name is present in the output.

packages/fxa-settings/tailwind.config.js
```diff
@@ -0,0 +1,181 @@
+module.exports = {
+  purge: {
+    content: ['./src/**/*.tsx', './public/index.html'],
+    whitelist: [],
+  },
+  theme: {
+    screens: {
+      sm: '480px',
```
LZoog, line 8: I think we may actually want to call these `mobileLandscape`, `tablet`, `desktop`, and `desktopXl` to make it more clear what we're targeting. (That's my preference anyway)

LZoog:
So @jodyheavener I had another thought... I still feel like what we discussed here and in Slack is the right approach, to allow classes to be set on shared components only through a prop and likely allowing non-shared components to have a default class list, but I'm wonder how this is going to affect Storybook in `fxa-react`. Can we import that Tailwind out file into Storybook there? 🤔 One other hiccup is eventually getting all of our Storybook builds up on GH pages.

packages/fxa-settings/tailwind.config.js
jodyheavener: I take your point, my personal preference is to keep them in, simply because they're not adding any extra bulk to the output and if we do end up needing to use a color it'll immediately be available / no out of scope diff in the commit. If it's a file length thing I would be cool moving it to a shared JSON file that we import? I can be swayed to remove them though if there are strong opinions!

jodyheavener:
> Can we import that Tailwind out file into Storybook there?

Oof yeah good point. I'm sure there's a way to do it, but IMO sharing should only flow from projects intended to be shared. If we start importing Settings styles into fxa-react we would also need to think about how to do this with Payments, or Admin, or Content Server. It's not impossible but it feels like that could get messy fast.

We could potentially have another Tailwind setup (which could share a common config with Settings) and have that build before Storybook starts up. But that begs the question: is everything under fxa-react going to be Tailwind? Like, I know we have Survey components in there as well.

packages/fxa-settings/README.md
jodyheavener: Added this and an example

packages/fxa-settings/tailwind.config.js
LZoog: I think one other thing that could come out of listing _all_ of the values (of blues for example) could be someone accidentally using `blue-300` when they should have used `blue-400`, when `blue-50` through `blue-300` shouldn't be used anywhere in the project. It could open up the possibility of using one assumed to be the right one because just because it's there, but that could also happen with someone using `blue-500` when they should have used `blue-400`. I guess ultimately it's up to dev/UX to catch it so I'm kind of torn.

I guess we can't sync up these values to protocol's package because we don't want that as a dependency, but what if their colors change in the future? We could have a bunch of colors we're not even using that would be inaccurate. Maybe that's not a concern though because that would only happen with rebranding?

Maybe a happy medium is pulling out values we _really_ don't think will be used, but even then I feel like there are quite a few. I don't see us ever using `violet`, `pink`, `purple`, `ink`, possibly `marketing-gray`... If we want our styles to be really consistent, maybe we should only list the ones we're actually going to use in the design, add others as needed, and keep the others somewhere else easily accessible when we do need them since you've already done the work for it. They could be listed when we get around to a design or branding doc. I guess ultimately I'm OK with these living in another file we pull in, but I'm not convinced we're _ever_ going to use some of them in FxA so I'm not sure we need them now because they could change.

LZoog:
> We could potentially have another Tailwind setup (which could share a common config with Settings) and have that build before Storybook starts up. But that begs the question: is everything under fxa-react going to be Tailwind? Like, I know we have Survey components in there as well.

I think we have one single Survey component but I believe implementation is WIP. Worst case scenario it can have that adjacent SCSS file until we get around to using it like the others, we do want to move towards this as our CSS solution for FxA. We probably do want to share the config between `fxa-settings` and `fxa-react` then like you mentioned just for Storybook purposes there, but we don't want PurgeCSS set up there since class names won't be present in the TSX files. We can extend this to payments when we start Tailwinding it up there.

jodyheavener:
@LZoog I'll make those changes you requested, but I won't be able to until I can get Inspect access. Should happen soon-ish.

packages/fxa-settings/tailwind.config.js
dannycoates: I'm fine with having them all here from the start. I'd rather have the snapshot of all the Protocol colors than bring them in as needed because it's one less thing to change during regular dev and color updates will prob be easier to diff when done all at once.

LZoog: @dannycoates Jody and I had a meeting with Phil and Jess earlier, they have a different set of colors for us to use that aren't exactly Protocol colors. Not all of them will be used with the current design but we were told those colors will be the bucket we pull from in the future when we do need to use those shades so we're all on the same page. 👍

packages/fxa-settings/package.json
```diff
@@ -60,6 +62,8 @@
     "eslint-plugin-react": "^7.19.0",
     "node-sass": "^4.14.0",
     "pm2": "^4.2.3",
-    "sass-lint": "^1.13.1"
+    "postcss-cli": "^7.1.1",
+    "sass-lint": "^1.13.1",
```
dannycoates, line 66: _\*cough\*_ stylelint

jodyheavener: That's coming out here! https://github.com/mozilla/fxa/pull/5343

packages/fxa-settings/README.md
```diff
@@ -14,7 +14,54 @@ You can import React components into this project. This is currently restricted
...
+
+#### Custom styles
+
+In the event you need to write styles that Tailwind is not able to affectively cover you should try to keep the following guidelines in mind:
+
+- SCSS is available.
+- PurgeCSS does not currently strip styles written outside of `src/styles/tailwind.css`, so it's up to you to keep styles limited to what you're using in your component.
+- Custom styles should be restricted to the component they are being used in. Typically, this is just an `index.scss` file adjacent to your component's `index.tsx` and then imported.
+- Because custom styles, once imported, are included in the global styles you should still be careful about how you're naming classes to avoid collisions. Generally you should match the class name to the name of your component. If you're using a shared component you could name according to its usage (e.g. using a shared UnitRow to display secondary emails could be `secondary-emails-row`).
```
LZoog, line 62: I would recommend this being `unit-row-secondary-emails`, [component]-[descriptor].

packages/fxa-settings/src/styles/fonts.scss
```diff
@@ -0,0 +1,44 @@
+@font-face {
+  font-family: 'Inter';
+  font-style:  normal;
+  font-weight: 400;
+  font-display: swap;
```
LZoog, line 5: I think `swap` is a good choice for now, if we notice any delayed loading and don't like seeing a fallback switch to the actual font we can try `block` or just resort to `fallback`.

packages/fxa-settings/src/styles/fonts.scss
```diff
@@ -0,0 +1,44 @@
...
+       url("fonts/Metropolis-Regular.woff?v=10.0.0") format("woff");
+}
+
+@font-face {
+  font-family: 'Metropolis';
+  font-style:  normal;
+  font-weight: 600;
+  font-display: swap;
+  src: url("fonts/Metropolis-SemiBold.woff2?v=10.0.0") format("woff2"),
+       url("fonts/Metropolis-SemiBold.woff?v=10.0.0") format("woff");
```
LZoog, line 34: I think we convinced UX not to use `semi-bold` to limit the custom fonts?

packages/fxa-settings/tailwind.config.js
jodyheavener: Ah, I thought I had addressed this. I will make this change.

packages/fxa-settings/tailwind.config.js
```diff
@@ -0,0 +1,174 @@
...
+      md: '768px',
+      lg: '1024px',
+      xl: '1441px',
+    },
+    fontSize: {
+      xs: '12px',
+      sm: '14px',
+      base: '16px',
```
LZoog, line 16: Should we call this `md` to match the other size conventions?

packages/fxa-settings/src/index.scss
```diff
@@ -2,6 +2,13 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-@import 'modern-normalize';
+@import 'styles/fonts';
+@import 'styles/tailwind.out';
 
-// Add app-wide styles here
+body {
+  font-size: 16px;
```
LZoog, line 9: Can we reference the Tailwind value here as well as the font-family?

packages/fxa-settings/tailwind.config.js
jodyheavener: Per our slack convo, gonna keep it for now as this is a [Tailwind convention](https://tailwindcss.com/docs/font-size/#app).

packages/fxa-settings/tailwind.config.js
```diff
@@ -0,0 +1,174 @@
+module.exports = {
+  purge: {
+    content: ['./src/**/*.tsx', './public/index.html'],
+    whitelist: [],
+  },
+  theme: {
+    screens: {
+      mobileLandscape: '480px',
+      tablet: { raw: '(min-width: 768px) and (min-height: 481px)' },
```
jodyheavener, line 9: Just bringing this to your attention.

PR #5352, opened 2020-05-18
Yarn

dannycoates:
The main point of this is to allow us to use `"workspace:*"` references to share code between internal packages. The `"file:../whatever"` method we were using has several bugs and unexpected behavior that makes them fragile and troublesome to work with. If you haven't noticed any problems with them yet consider yourself lucky. Workspaces are designed to do what we expect. They are currently in [yarn](https://yarnpkg.com/features/workspaces), but [npm 7 will also support them](https://github.com/npm/rfcs/blob/latest/accepted/0026-workspaces.md).

This PR is large but the changes are relatively straightforward. I had to go further down the rabbit hole than I expected to get everything working because yarn handles deps differently so some things that happened to work with npm don't with yarn. Fortunately the changes should now work with both.

I'll add review comments for specific changes, but in general here's what's going on:

1. update package.json dependencies to make yarn happy
2. update file paths in code to use `require.resolve` instead of assuming local `node_modules`
3. removed fxa-amplitude-send to avoid updating that biz. Thanks, Jared for the ok.
4. removed tslint because something was broken, it's deprecated, and we should just use eslint anyway
5. updated the docker build to work with yarn
6. removed lerna because yarn does what we needed it for
7. updated pm2 configs to work with yarn
8. switched forked packages to use mozilla-fxa


lmorchard:
I don't personally care either way for npm vs yarn - but what does this change get us?

dannycoates:
> but what does this change get us?

I'm writing an ADR for this. There's a few reasons, but better code sharing is the main driver. Our current setup is extremely fragile.

packages/123done/static/index.html
```diff
@@ -9,12 +9,12 @@
     <meta name="firefox-accounts" content="supported" />
     <link
       rel="stylesheet"
-      href="/bower_components/normalize-css/normalize.css"
+      href="/components/normalize.css"
```
dannycoates, line 12: I remove bower because we _really_ shouldn't be using it anymore. Instead of adding a build step I just added the files to git. This is 123done people, don't judge me.

packages/fortress/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "firefox-fortress",
+  "name": "fortress",
```
dannycoates, line 2: name needs to match dirname for workspaces to work. idk why

packages/fxa-admin-panel/pm2.config.js
```diff
@@ -2,35 +2,43 @@
...
       min_uptime: '2m',
       env: {
+        SKIP_PREFLIGHT_CHECK: 'true',
         NODE_ENV: 'development',
         PUBLIC_URL: 'http://localhost:8091',
         BROWSER: 'NONE',
         PORT: '8092',
+        PATH: process.env.PATH.split(':')
+          .filter(p => !p.includes(process.env.TMPDIR))
+          .join(':'),
```
dannycoates, line 39: yarn sets up a sandbox dir to run tasks and sets PATH and some other env vars to use it, but by the time pm2 starts the script the dir is gone, so this cleans up the path to remove the dead dir.

packages/fxa-content-server/app/styles/_modules.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../node_modules/jquery-modal/jquery.modal';
+@import '~jquery-modal/jquery.modal';
```
dannycoates, line 1: `~` is a special sass-loader flag to let it know to resolve the import externally

packages/fxa-content-server/server/lib/version.js
```diff
@@ -89,26 +89,9 @@ function getL10nVersion() {
 
 function getTosPpVersion() {
   try {
-    const pkgPath = '../../node_modules/legal-docs/package.json';
-    const pkgInfo = require(pkgPath);
-    let gitHead = undefined;
-    // npm < 5.8 has the git hash in `gitHead`
-    if (pkgInfo && pkgInfo.gitHead) {
```
dannycoates, line 96: yarn doesn't add these properties to the package.json of dependencies

packages/fxa-content-server/webpack.config.js
```diff
@@ -52,62 +52,32 @@ const webpackConfig = {
       'node_modules',
     ],
     alias: {
-      'asmcrypto.js': path.resolve(
-        __dirname,
-        'node_modules/asmcrypto.js/asmcrypto.min.js'
+      'asmcrypto.js': require.resolve('asmcrypto.js/asmcrypto.min.js'),
```
dannycoates, line 55: yarn lifts as much as it can to the root level `node_modules` to dedupe. We need to `resolve` the location instead of assuming where it is.

packages/fxa-payments-server/scripts/test-ci.sh
```diff
@@ -1,25 +1,10 @@
...
-npm ci
-
-# TODO rm the CI=false
-PUBLIC_URL=/ INLINE_RUNTIME_CHUNK=false CI=false npm run build
-CI=yes npm test
+# yarn workspaces focus fxa-payments-server
+
+# payments-server has some missing dependency that
+# causes tests to fail when only it's deps are installed
+# but passes when all fxa deps are installed
```
dannycoates, line 7: i have no idea what's causing this, but here's what it looks like: https://app.circleci.com/pipelines/github/mozilla/fxa/10575/workflows/8d14f766-9196-4c58-9715-5c11b90e40de/jobs/168829

packages/fxa-payments-server/src/lib/sentry.test.js
```diff
@@ -183,13 +182,4 @@ describe('lib/sentry', function() {
...
   });
-
-  describe('captureException', () => {
-    it('calls Sentry.captureException', () => {
-      jest.spyOn(Sentry, 'captureException');
-      sentryMetrics.captureException(new Error('testo'));
-      expect(Sentry.captureException).toHaveBeenCalled();
-      Sentry.captureException.mockRestore();
-    });
-  });
```
dannycoates, line 194: idk, why this started triggering an error now but this was the error:

https://stackoverflow.com/questions/53162001/typeerror-during-jests-spyon-cannot-set-property-getrequest-of-object-which

packages/fxa-support-panel/lib/api.ts
```diff
@@ -258,7 +257,8 @@ export function init(
       options: {
         handler: supportController.displayUser,
         validate: {
-          query: (queryValidator as unknown) as hapiJoi.ObjectSchema,
```
dannycoates, line 261: the newer version of typescript didn't like this an I couldn't figure out how to fix it

.circleci/base-install.sh
```diff
@@ -5,12 +5,14 @@ DIR=$(dirname "$0")
 cd "$DIR/.."
 
 # npm install just enough to run these scripts
-npm ci --ignore-scripts --no-optional --only=prod
+npm i --ignore-scripts --no-optional --only=prod --no-package-lock
```
jodyheavener, line 8: Do we care that this doesn't first remove node_modules like `ci` does?

packages/fxa-admin-server/tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "forceConsistentCasingInFileNames": true,
     "noEmitHelpers": true,
     "importHelpers": true,
-    "typeRoots": ["./types", "node_modules/@types"]
+    "typeRoots": ["./types", "node_modules/@types", "../../node_modules/@types"]
```
jodyheavener, line 18: I wonder if this could exist in a common tsconfig, given how widely used it is. Maybe a future task.

packages/fxa-auth-server/pm2.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   apps: [
     {
       name: 'auth',
-      script: 'ts-node bin/key_server.js',
+      script: 'node -r ts-node/register bin/key_server.js',
```
jodyheavener, line 9: What's the difference between `ts-node` and `node -r ts-node/register`?

packages/fxa-profile-server/scripts/test-ci.sh
```diff
@@ -2,12 +2,7 @@
...
-cd $DIR/../../fxa-shared
-npm ci
-
-cd ../fxa-profile-server
-cp ../version.json config
-npm ci
-npm test
+cp "$DIR/../../version.json" "$DIR/../config"
+yarn workspaces focus fxa-profile-server
+yarn workspace fxa-shared run build
```
jodyheavener, line 7: Some scripts use npx; do we want that here?

packages/fxa-auth-server/scripts/test-local.sh
```diff
@@ -16,8 +16,7 @@ npx ts-node ./scripts/gen_keys.js
 npx ts-node ./scripts/gen_vapid_keys.js
 npx ts-node ./scripts/oauth_gen_keys.js
 node ../fxa-auth-db-mysql/bin/db_patcher > /dev/null
-npm run lint:eslint
-grunt copyright
+npm run lint
```
jodyheavener, line 19: Is there a reason why some instances of `npm` get swapped out for `[npx ]yarn <command>` and others are left alone?

packages/fxa-auth-server/pm2.config.js
jodyheavener: Ah [this](https://stackoverflow.com/a/53802296/717633) looks like a decent reason.

.circleci/base-install.sh
dannycoates: this runs in circleci so there's no existing node_modules

packages/fxa-admin-server/tsconfig.json
dannycoates: Yes, standardizing tsconfig is the next follow-up task. I just had to do a few small changes to get the existing ones working with yarn.

packages/fxa-auth-server/pm2.config.js
dannycoates: it's a quirk of the shell env differences with yarn. With yarn at the time this line executes `ts-node` isn't in the `PATH`

packages/fxa-profile-server/scripts/test-ci.sh
dannycoates: I used npx when I can't be sure yarn is installed globally. On circleci it is, our docker images and teamcity it isn't.

lmorchard:
> I'm writing an ADR for this.

Not worth blocking this PR over it, but: FWIW, we should really get the ADR to record the decision before executing on the decision. Then, we know why we're doing the thing, what we're not doing & why. We haven't been consistent about that, since some ADRs have been retroactive to fill in context.

packages/fxa-payments-server/scripts/test-ci.sh
bbangert: Wondering if this is related to the issues pointed out here:
https://medium.com/@rossbulat/typescript-working-with-paths-packages-and-yarn-workspaces-6fbc7087b325

Also, I assume that when we want to use fxa-shared in a typescript package, we will need the path alias and tsconfig changes the above post cites?

packages/fxa-support-panel/lib/api.ts
bbangert: typescript 3.9 also has some issues with other libs. we won't be able to use it for awhile.

packages/fxa-auth-server/scripts/test-local.sh
dannycoates: I'm trying not to go all-in on yarn (for now at least) so I tried to only changed the spots that _needed_ to change. I may not have ended up at the minimum but that was the plan.

packages/fxa-support-panel/lib/api.ts
dannycoates: I was able to get everything to 3.8.3 but 3.9 broke a couple of our 3rd party deps.

packages/fxa-payments-server/scripts/test-ci.sh
dannycoates: There's still work to do after this PR to improve our typescript situation (thanks for the link) but I think this is a missing dependency, I just don't know which.

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ commands:
 jobs:
   test-package:
     docker:
-      - image: mozilla/fxa-circleci
```
lmorchard, line 57: Why the image switch? (I might figure this out further through the PR)

README.md
```diff
@@ -94,6 +96,18 @@ When you signup for an account using the form on `localhost:3030/signup` the "in
 
 If you get an `error` status for any of the servers please verify that you installed all required dependencies. Otherwise file an issue on this repository or [connect with the team on Firefox Accounts Riot](https://chat.mozilla.org/#/room/#fxa:mozilla.org).
 
+### Managing dependencies
+
+Use `yarn` to add dependencies. To add a dependency to a single package, which is what you'll usually want, run something like:
+
+```sh
+yarn workspace fxa-shared add --dev eslint
```
lmorchard, line 104: Ah, okay, seems like this workspace thing is part of what lerna unnecessary.

_scripts/base-docker.sh
```diff
@@ -7,7 +7,30 @@ for d in ./packages/*/ ; do
...
-npm i lerna
-npx lerna bootstrap --hoist pm2
-npx lerna run --stream build
-npx lerna exec --stream --concurrency 2 --no-bail -- npm prune --production
+# `npx yarn` because `npm i -g yarn` needs sudo
+npx yarn install
+SKIP_PREFLIGHT_CHECK=true npx yarn workspaces foreach --topological run build
+rm -rf node_modules
+rm -rf packages/*/node_modules
+npx yarn workspaces focus --production \
```
lmorchard, line 15: Does this kind of serve a similar purpose as `npm ci` for minimal install?

_scripts/gh-pages.sh
```diff
@@ -12,13 +12,8 @@ echo "Building docs."
...
 cargo doc --no-deps
 
-# fxa-payments-server relies on fxa-content-server .scss styles, which in turn
-# rely on some modules in package.json
-cd ../../packages/fxa-content-server
-npm ci
-
 cd ../../packages/fxa-payments-server
-npm ci
+yarn workspaces focus fxa-payments-server
```
lmorchard, line 16: Hmm, does this break anything in the storybook build? IIRC, part of the `npm ci` over in content-server for this particular thing is the post-install stuff over there that payments-server needs for its own build

Will likely replace this with PR #5357 - but I'm doing this same `npm ci` there too

_scripts/test-package.sh
```diff
@@ -10,17 +10,17 @@ npm test fxa-auth-db-mysql fxa-auth-server
...
-  scopes="$scopes --scope $scope"
+  workspaces="$workspaces --include $workspace"
 done
 
 if [[ -z "$PACKAGE" ]]; then
   >&2 echo "$help"
   exit 1
 elif [[ "$PACKAGE" == "all" ]]; then
-  lerna run test --stream --no-prefix --loglevel success --concurrency 1 --ignore fxa-amplitude-send
+  yarn workspaces foreach --topological-dev run test
```
lmorchard, line 23: I love a good weird CLI option (lol @ `topological-dev`)

packages/fxa-payments-server/scripts/test-ci.sh
lmorchard: I think payments-server relies on some assets built by content-server, not sure if that's the cause here.

.circleci/config.yml
dannycoates: This is a case of me noticing that we don't need a special image here while I was making the other change. The circleci images tend to load faster than custom ones.

_scripts/base-docker.sh
dannycoates: yes, `focus --production` is the way to do the equivalent of `npm ci --production`. Plain `yarn install` doesn't have that option.

_scripts/gh-pages.sh
dannycoates: "postinstall" will still run on content-server since payments-server has it as a dependency

packages/fxa-settings/package.json
```diff
@@ -25,7 +25,7 @@
     "build": "INLINE_RUNTIME_CHUNK=false rescripts build",
     "test": "rescripts test",
     "eject": "react-scripts eject",
-    "lint": "npm-run-all --parallel lint:*",
+    "lint": "npm-run-all --parallel lint:eslint lint:sass",
```
lmorchard, line 28: Not a huge deal, but why make all these an explicit list? For clarity's sake?

packages/fxa-shared/package.json
```diff
@@ -4,13 +4,11 @@
   "description": "Shared module for FxA repositories",
   "main": "dist/index.js",
   "scripts": {
-    "postinstall": "npm run build",
```
lmorchard, line 7: Will this end up being different than `npm run build`?

packages/fxa-content-server/grunttasks/yeoman.js
```diff
@@ -2,10 +2,12 @@
...
 
+// const { dirname } = require('path');
+
 module.exports = function(grunt) {
-  var TEMP_DIR = '.tmp';
-  var TEMPLATE_ROOT = 'server/templates';
-  var TOS_PP_REPO_ROOT = 'node_modules/legal-docs';
+  const TEMP_DIR = '.tmp';
+  const TEMPLATE_ROOT = 'server/templates';
+  const TOS_PP_REPO_ROOT = '../../node_modules/legal-docs'; //dirname(require.resolve('legal-docs'));
```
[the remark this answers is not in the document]
dannycoates: waiting on https://github.com/mozilla/legal-docs/pull/1464

packages/fxa-settings/package.json
dannycoates: something about the internal implementation of npm-run-all and yarn not liking the `*` 🤷

packages/fxa-shared/package.json
dannycoates: I'd like for this `postinstall` to go away with a little tsconfig magic in the future

packages/fxa-payments-server/package.json
```diff
@@ -3,17 +3,17 @@
...
     "start": "pm2 start pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "restart": "pm2 restart pm2.config.js",
-    "build": "PUBLIC_URL=/ INLINE_RUNTIME_CHUNK=false react-scripts build",
+    "build": "SKIP_PREFLIGHT_CHECK=true PUBLIC_URL=/ INLINE_RUNTIME_CHUNK=false CI=false react-scripts build",
     "eject": "react-scripts eject",
-    "test": "npm-run-all test:*",
-    "test:frontend": "rescripts test --coverage --verbose",
+    "test": "npm-run-all test:frontend test:server",
+    "test:frontend": "SKIP_PREFLIGHT_CHECK=true PUBLIC_URL=/ INLINE_RUNTIME_CHUNK=false CI=yes rescripts test --coverage --verbose",
```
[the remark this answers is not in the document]
dannycoates: something like this https://github.com/yarnpkg/berry/issues/1342#issuecomment-627163187

packages/fxa-graphql-api/pm2.config.js
```diff
@@ -15,9 +15,10 @@ module.exports = {
         NODE_OPTIONS: '--inspect=9200',
         TS_NODE_TRANSPILE_ONLY: 'true',
         TS_NODE_FILES: 'true',
-        PORT: '8290', // TODO: this needs to get added to src/config.ts
```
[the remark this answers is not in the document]
dannycoates: im working on consolidating our prettier configs. TIL we have 21 .prettierrc files

packages/fxa-payments-server/scripts/test-ci.sh
dannycoates: which assets? I didn't see any references to content-server outside of scss files. FWIW this also fails in the same way; builds fine, tests fail:
```
yarn workspaces focus fxa-payments-server
yarn workspace fxa-content-server run build
yarn build
yarn test
```

vbudhram:
Oh dang...didn't release I caused merge conflicts for you, sorry about :(

packages/fxa-payments-server/package.json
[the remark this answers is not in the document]
lmorchard: Hmm, I wonder if CRA's been fixed for this and we just need to upgrade? That or maybe one more reason to eject from the magical world of CRA someday

PR #5362, opened 2020-05-19
fix(auth server): use ts-node-script shebang when executing ts-dependent scripts

### LINKED ISSUE #5261: scripts/delete-account.js in fxa-auth-server is broken again
I haven't been following the Typescript conversions on fxa-auth-server, but I think we might have broken `scripts/delete-account.js` again. ([Previously](https://github.com/mozilla/fxa/issues/4032)) A few other scripts affected too, like `clean-up-old-ci-stripe-customers.js`

Looks like `config/index.js` is now `config/index.ts`, which leads to this:
```
➜  fxa-auth-server git:(master) ✗ node scripts/delete-account.js lorchard@mozilla.com
internal/modules/cjs/loader.js:796
    throw err;
    ^

Error: Cannot find module '../config'
Require stack:
- /home/lmorchard/devel/fxa/packages/fxa-auth-server/scripts/delete-account.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:793:17)
    at Function.Module._load (internal/modules/cjs/loader.js:686:27)
    at Module.require (internal/modules/cjs/loader.js:848:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/lmorchard/devel/fxa/packages/fxa-auth-server/scripts/delete-account.js:25:16)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10)
    at internal/main/run_main_module.js:17:11 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/lmorchard/devel/fxa/packages/fxa-auth-server/scripts/delete-account.js'
  ]
}
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1857)
┆Issue Number: FXA-1857

jodyheavener:
Closes #5261

Independently executed Auth Server scripts (like `delete-account.js, the issue) that import the config, which was recently converted to TypeScript, no longer work because they're not using TS themselves.

So, I've gone ahead and updated the shebangs in these files to compile with ts-node (specifically [ts-node-script](https://www.npmjs.com/package/ts-node#shebang)). I did not modify any scripts that do not use the config.

Also, per the initial issue, I ran `delete-account.js` and it additionally fails because it can't find `../lib/subhub/client`. Sure enough, `lib/subhub` does not exist in the repo. Any ideas where this comes from?

### DIFF
--- a/packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
+++ b/packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
@@ -1,15 +1,15 @@
-#!/usr/bin/env node
+#!/usr/bin/env ts-node-script
 
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const program = require('commander');
-const package = require('../package.json');
+const pckg = require('../package.json');
 
 async function init() {
   program
-    .version(package.version)
+    .version(pckg.version)
     .option('-c, --config [config]', 'Configuration to use. Ex. dev')
     .option(
       '-k, --stripe-key [key]',
--- a/packages/fxa-auth-server/scripts/activate-new-signing-key.js
+++ b/packages/fxa-auth-server/scripts/activate-new-signing-key.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env ts-node-script
 
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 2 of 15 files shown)


packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
```diff
@@ -1,15 +1,15 @@
-#!/usr/bin/env node
+#!/usr/bin/env ts-node-script
 
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const program = require('commander');
-const package = require('../package.json');
+const pckg = require('../package.json');
```
jodyheavener, line 8: `package` is a reserved word, hence the change

Issue #4325, opened 2020-02-28
Add ecosystem_anon_id field to core profile data returned by auth server


data-sync-user:
Also, pass the ecosystem_anon_id through the profile data returned by the profile server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1233)
┆Issue Number: FXA-1233

data-sync-user:
➤ Jared Hirsch commented:

This is partially implemented in https://github.com/mozilla/fxa/pull/5364, ( https://github.com/mozilla/fxa/pull/5364, ) but that PR adds API surface without DB integration.

Note for working on this: add ETag and If-Match support to ensure we don't have lost updates (two clients POSTing new anon_id values at the same time).

data-sync-user:
➤ Jared Hirsch commented:

PR: https://github.com/mozilla/fxa/pull/5481

PR #5364, opened 2020-05-19
feat(profile): Add API endpoints for ecosystem_anon_id

6a68:
Starting off the AET work with an API that will need to agree with changes that ride the trains into desktop. This ecosystem_anon_id field isn't yet persisted in the core profile fields in the auth DB, that's a different bug. Not 100% sure I'm logging correctly. Still need to add tests to api.js.

### DIFF
--- a/packages/fxa-profile-server/docs/API.md
+++ b/packages/fxa-profile-server/docs/API.md
@@ -63,6 +63,8 @@ The currently-defined error responses are:
 - [DELETE /v1/avatar/:id][delete]
 - [GET /v1/display_name][display_name]
 - [POST /v1/display_name][display_name-post]
+- [GET /v1/ecosystem_anon_id][ecosystem_anon_id]
+- [POST /v1/ecosystem_anon_id][ecosystem_anon_id-post]
 - [DELETE /v1/cache/:uid][cache_delete]
 
 ### GET /v1/profile
@@ -84,6 +86,7 @@ curl -v \
 ```js
 {
   "uid": "6d940dd41e636cc156074109b8092f96",
+  "ecosystemAnonId": "eyJhbGciOiJFQ0RILUVTIiwia2lkIjoiMFZFRTdmT0txbFdHVGZrY0taRUJ2WWl3dkpMYTRUUGlJVGxXMGJOcDdqVSIsImVwayI6eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2IiwieCI6InY3Q1FlRWtVQjMwUGwxV0tPMUZUZ25OQlNQdlFyNlh0UnZxT2kzSWdzNHciLCJ5IjoiNDBKVEpaQlMwOXpWNHpxb0hHZDI5NGFDeHRqcGU5a09reGhELVctUEZsSSJ9LCJlbmMiOiJBMjU2R0NNIn0.A_wzJya943vlHKFH.yq0JhkGZiZd6UiZK6goTcEf6i4gbbBeXxvq8QV5_nC4.Knl_sYSBrrP-aa54z6B6gA",
   "email": "user@example.domain",
   "locale": "en-US",
   "amrValues": ["pwd", "otp"],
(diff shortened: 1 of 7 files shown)


rfk:
I'll take a more detailed look, but this seems broadly as I expected 👍 

Could we please also include this field in the response from `GET /profile` so that it can be fetched along with other data in a single operation? (It may be that we get this for free by virtue of having it in `_core_profile`, but the docs don't indicate this to be the case).

packages/fxa-profile-server/lib/routes/ecosystem_anon_id/post.js
```diff
@@ -0,0 +1,31 @@
...
+  auth: {
+    strategy: 'oauth',
+    scope: ['profile:ecosystem_anon_id:write'],
+  },
+  validate: {
+    payload: {
+      ecosystemAnonId: Joi.string().required(),
+    },
+  },
+  handler: async function ecosystemAnonIdPost(req) {
```
rfk, line 20: Ideally, clients would be able to `If-Match` or `If-Unmodified-Since` on this post, for the case when we're filling in a placeholder anon id and we don't want two clients to race in setting it.

rfk:
To clarify, is the idea that we'll land and ship just the API surface in this bug, and plumb in the actual storing in a followup?

6a68:
> To clarify, is the idea that we'll land and ship just the API surface in this bug, and plumb in the actual storing in a followup?

Yes, exactly. I'm hoping to unblock desktop client work by starting with the API, then I'll work downwards.

6a68:
Thanks for taking a look! I'll fixup the bits you mentioned tomorrow morning and ping you for a followup :+1:

packages/fxa-profile-server/lib/routes/ecosystem_anon_id/post.js
6a68: Sounds good. I've added a note to https://github.com/mozilla/fxa/issues/4325, since I can't test the ETag is checked against the If-Match value until the DB returns an actual response.

6a68:
> Could we please also include this field in the response from GET /profile so that it can be fetched along with other data in a single operation? (It may be that we get this for free by virtue of having it in _core_profile, but the docs don't indicate this to be the case).

I think this should work now, as I've added `ecosystemAnonId` as an optional field in the response schema for GET /profile. It appears that GET /profile pulls data from lib/profileCache, which in turn aggregates data from several routes: _core_profile, plus profile-specific routes like avatar and displayName. I'll double-check that turns out to be true when I get eco_anon_id fully plumbed through to the auth DB, but I think we're good.

packages/fxa-profile-server/docs/API.md
```diff
@@ -310,6 +312,57 @@ curl -v \
...
+
+#### Response
+
+```json
+{
+  "ecosystemAnonId": "eyJhbGciOiJFQ0RILUVTIiwia2lkIjoiMFZFRTdmT0txbFdHVGZrY0taRUJ2WWl3dkpMYTRUUGlJVGxXMGJOcDdqVSIsImVwayI6eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2IiwieCI6InY3Q1FlRWtVQjMwUGwxV0tPMUZUZ25OQlNQdlFyNlh0UnZxT2kzSWdzNHciLCJ5IjoiNDBKVEpaQlMwOXpWNHpxb0hHZDI5NGFDeHRqcGU5a09reGhELVctUEZsSSJ9LCJlbmMiOiJBMjU2R0NNIn0..A_wzJya943vlHKFH.yq0JhkGZiZd6UiZK6goTcEf6i4gbbBeXxvq8QV5_nC4.Knl_sYSBrrP-aa54z6B6gA"
+}
+```
+
+Returns a `204 No Content` if there is no ecosystem_anon_id for the user.
```
vbudhram, line 337: I'm cool with this, I've seen other endpoints return 200 and an empty response.

packages/fxa-profile-server/lib/routes/profile.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
   },
   response: {
     schema: {
+      ecosystemAnonId: Joi.string().optional(),
```
vbudhram, line 26: nit: Update docs

packages/fxa-profile-server/test/api.js
```diff
@@ -1461,4 +1485,123 @@ describe('api', function() {
       });
     });
   });
+
+  describe('/ecosystem_anon_id', function() {
+    var tok = token();
+
+    describe('GET', function() {
+      xit('should return an ecosystem_anon_id', function() {});
```
vbudhram, line 1493: TIL `xit`

packages/fxa-profile-server/test/api.js
```diff
@@ -765,6 +767,28 @@ describe('api', function() {
           assertSecurityHeaders(res);
         });
     });
+
+    it('should omit `ecosystem_anon_id` if not returned by auth-server', () => {
```
vbudhram, line 771: In the auth-server I've been slowly updating the test code to use `async/await` since it makes things a little easier to read.

packages/fxa-profile-server/lib/routes/ecosystem_anon_id/post.js
```diff
@@ -0,0 +1,31 @@
...
+    const uid = req.auth.credentials.user;
+    logger.info('activityEvent', { event: 'ecosystemAnonId.post', uid: uid });
+
+    return req.server.methods.profileCache.drop(uid).then(() => {
```
vladikoff, line 24: could this just use `await` ? We've been slowly converting the routes

packages/fxa-profile-server/test/api.js
6a68: oh, nice! yeah, I was just trying to follow the existing style in the files, but I'll start converting to async/await :+1:

packages/fxa-profile-server/lib/routes/ecosystem_anon_id/post.js
6a68: yeah, definitely

PR #5384, opened 2020-05-20
feat(graphql-api): add sessionToken auth scheme

### LINKED ISSUE #5359: sessionToken auth for graphql-api
Per the latest auth ADR, we will be using sessionToken's for auth in the graphql-api. The existing oauth2 should be replaced with sessionToken use.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1914)
┆Issue Number: FXA-1914

bbangert:
Because:

* prettier 2.0 has different defaults than the git commit hook
  which uses prettier 1.
* We decided to use sessionToken auth for calling into auth-server
  instead of OAuth2.

This commit:

* Updates the prettier config to use prettier 1 compatible defaults.
* Restores prettier 1 styling.
* Replaces OAuth2 with sessionToken lookup using the auth-server.

Closes #5359


packages/fxa-graphql-api/package.json
```diff
@@ -34,6 +34,7 @@
     "apollo-server": "^2.13.1",
     "apollo-server-express": "^2.13.1",
     "convict": "^6.0.0",
+    "fxa-js-client": "^1.0.25",
```
vladikoff, line 37: 👀 should this the new client or at least reference the client in the local repo? I Think the npm version is gonna be deprecated

bbangert: Oh, I don't know. @dannycoates?

bbangert: There's a change that' s upcoming that makes some large changes to how things reference other bits in the local repo. So it'd probably be better to wait for that to land then switch to it.

dannycoates: yeah, we should use "workspace:*" for this after yarn lands

dannycoates: in the meantime, this is fine

bbangert: My next PR will need to immediately switch to the version in this monorepo as the NPM release does not have `attachedClients` which is needed.

6a68, line 37: FYI there's a typescript fxa-js-client WIP that's in content-server for now, but should be available soon as a subpackage

packages/fxa-graphql-api/.prettierrc
```diff
@@ -1,4 +1,7 @@
 {
   "printWidth": 100,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none",
```
lmorchard, line 4: Saw this in #5386 but looks like it came from over here, so copying the comment: We have `"trailingComma": "es5"` [in other `.prettierrc` configs](https://github.com/mozilla/fxa/search?q=%22trailingComma%22%3A+%22es5%22&unscoped_q=%22trailingComma%22%3A+%22es5%22) - probably want that here too?

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -31,25 +32,26 @@ export async function createServer(
...
     container: Container,
-    resolvers: [AccountResolver],
+    resolvers: [AccountResolver]
   });
+  const authUser = Container.get(SessionTokenAuth);
   const debugMode = config.env !== 'production';
   const defaultContext = async ({ req }: any) => {
     const bearerToken = req.headers[config.authHeader.toLowerCase()];
-    const userId = bearerToken.split('Bearer ')[1];
+    const userId = await authUser.lookupUserId(bearerToken);
```
6a68, line 41: What happens if `lookupUserId` throws? Should this line be wrapped in a try/catch, and rethrown as a more generic error?

packages/fxa-graphql-api/.prettierrc
bbangert: Do we? I've been having the commit hook go and drop them.

packages/fxa-graphql-api/.prettierrc
```diff
@@ -1,4 +1,7 @@
...
+  "arrowParens": "avoid",
+  "endOfLine": "auto"
```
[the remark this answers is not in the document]
bbangert: Will change.

packages/fxa-graphql-api/package.json
bbangert: Yep, waiting for yarn change to land before trying to bring that in.

packages/fxa-graphql-api/src/lib/server.ts
bbangert: If it throws, then we'll get a sentry error reported. There's another graphql hook we can use to rewrite the top level error that would be a good spot to do that. But I'm not yet what those should be, which will probably be hard to determine until we start running it.

PR #5386, opened 2020-05-21
feat(graphql-api): implement security/devices resolvers

### LINKED ISSUE #5230: Implement query resolver for security/devices on settings page
The two sections (Security, Attached Devices) of the new Settings page should be implemented by calling into auth-server. These resolvers should then return that data.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1837)
┆Issue Number: FXA-1837

bbangert:
Because:

* We want to show a users attached clients and security settings.

This commit:

* Retrieves the requested graph fields from the auth-server and
  resolves them.

Closes #5230

### DIFF
--- a/packages/fxa-graphql-api/src/lib/datasources/authServer.ts
+++ b/packages/fxa-graphql-api/src/lib/datasources/authServer.ts
@@ -1,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { DataSource, DataSourceConfig } from 'apollo-datasource';
+import { Client } from 'fxa-js-client';
+import { Container } from 'typedi';
+
+import { Context } from '../server';
+import { fxAccountClientToken } from '../constants';
+
+function snakeToCamel(str: string) {
+  return str.replace(/(_\w)/g, (m: string) => m[1].toUpperCase());
+}
+
+/**
+ * Converts an object with keys that are possibly snake_cased to an
+ * object with keys that are camelCased.
+ *
+ * @param obj Object with string keys
+ */
+export function snakeToCamelObject(obj: { [key: string]: any }) {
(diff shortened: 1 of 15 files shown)


lmorchard:
Starting to take a look

packages/fxa-graphql-api/.prettierrc
```diff
@@ -1,4 +1,7 @@
 {
   "printWidth": 100,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "none",
```
lmorchard, line 4: We have `"trailingComma": "es5"` [in other `.prettierrc configs](https://github.com/mozilla/fxa/search?q=%22trailingComma%22%3A+%22es5%22&unscoped_q=%22trailingComma%22%3A+%22es5%22) - probably want that here too?

packages/fxa-graphql-api/src/lib/datasources/authServer.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Converts an object with keys that are possibly snake_cased to an
+ * object with keys that are camelCased.
+ *
+ * @param obj Object with string keys
+ */
+export function snakeToCamelObject(obj: { [key: string]: any }) {
+  return Object.fromEntries(Object.entries(obj).map(([k, v]) => [snakeToCamel(k), v]));
```
lmorchard, line 23: Not a real issue, probably not even a nit so much as a preference but: I kind of hate doing this sort of normalization. I've been bitten in the past by mistakes / confusion around properties from upstream data sources. I might be missing it - but is there a test or a test with a fixture that looks like auth-server API response to exercise this property renaming?

bbangert: Threw a test in to verify the camelCase conversion.

bbangert: The auth-server has a substantial amount of response validation to ensure it returns this correctly. I'm not sure adding more validation to make sure our response validation works helps a lot.

PR #5391, opened 2020-05-21
feat(a11y settings): Improve modal a11y

### LINKED ISSUE #5016: [a11y] Modal pulled from payments needs improved a11y
The `DialogMessage` from `fxa-payments-server` will be pulled in or the functionality will referenced for FXA-1621. However, going through this with a screenreader, there are a few issues that would be a11y wins if implemented:

* The modal should close on escape key press
* On modal close, shift focus back to the element that opened the modal
* When the modal is open, content behind it should not be read. Likely, the modal will need to be pulled out where it currently lies in the DOM, and placed adjacent to main page content. When the modal is active, the div for the main page content would then receive an `aria-hidden="true"` along with any other adjacent divs. When the modal is not active, it should have `aria-hidden="true"` and the main page content should _not_ have this property.
* To prevent scrolling in the background, we could add a class to the `body` (or container component of modal/page content) when the modal is open to set `overflow:hidden`. A dark background with opacity set (up to design) could be set here.
* Needs an `aria-labelledby="Modal_Header_ID"` and `aria-describedby="Modal_Content_ID`.
* Should have `aria-modal="true"` (this is new, do we want to use this?) or `role="dialog"` present.
* May need a tab fence once pulled out of the DOM (an empty div with a `tabindex="0"`) to ensure items inside are tabbable. Check for this last - go through the page with a screenreader and trigger a modal... can you tab around as expected, or does tab focus seem to get lost somewhere? If it does, try adding a tab fence.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1691)
┆Issue Number: FXA-1691

LZoog:
This commit:

* Allows the Modal component to close on escape key press
* Shifts focus back to the element that opened the modal when closed
* Disallows interaction with root element when modal is opened via aria-hidden, pointer events, and scroll prevention
* Adds `aria-labelledby`, `aria-describedby`, and `role='dialog'` properties
* Adds a tab fence so screenreaders are directed to the modal after opening

Because:
The current modal component needed some accessibility improvements.

Fixes #5016

This also makes `onDismiss` for `Modal` required because all of the modals in settings have a cancel/X button. I'll be opening an issue for us to backlog to better consolidate `DialogMessage` from payments with some of these updates, in the meantime I went with a check on `id` in the shared Portal component.

### DIFF
--- a/packages/fxa-react/components/Portal/index.test.tsx
+++ b/packages/fxa-react/components/Portal/index.test.tsx
@@ -7,7 +7,7 @@ import Portal from './index';
 afterEach(cleanup);
 
 it('renders children in a new element outside original DOM parent', () => {
-  const { debug, container, getByTestId } = render(
+  const { container, getByTestId } = render(
     <div data-testid="portal-parent">
       <Portal id="foo">
         <div data-testid="children">Hi mom</div>
@@ -18,10 +18,14 @@ it('renders children in a new element outside original DOM parent', () => {
   expect(getByTestId('portal-parent')).not.toContainElement(childrenEl);
   const containerParent = container.parentNode as ParentNode;
   expect(containerParent.querySelector('#foo')).toContainElement(childrenEl);
+  expect(containerParent.querySelector('#foo')).toHaveAttribute(
+    'role',
+    'dialog'
+  );
 });
 
 it('renders multiple instances with the same ID to the same DOM parent', () => {
-  const { debug, container, getByTestId } = render(
+  const { container, getByTestId } = render(
     <div data-testid="portal-parent">
       <Portal id="foo">
         <div data-testid="p1">Hi mom</div>
@@ -43,4 +47,37 @@ it('renders multiple instances with the same ID to the same DOM parent', () => {
   expect(fooPortal).toContainElement(getByTestId('p1'));
   expect(fooPortal).toContainElement(getByTestId('p2'));
   expect(fooPortal).not.toContainElement(getByTestId('p3'));
+});
+
+it('applies a11y improvements when id is set to "modal"', () => {
+  document.body.innerHTML = `
+    <div data-testid="root"></div>
+    <div data-testid="adjacent-to-root"></div>
+  `;
+
+  const headerId = 'some-header-id';
+  const descId = 'some-desc-id';
+
+  const { getByTestId } = render(
+    <Portal id="modal" {...{ headerId, descId }}>
+      <div>Hi mom</div>
+    </Portal>
+  );
+
+  const body = document.body;
+  const root = getByTestId('root');
+  const adjacentToRoot = getByTestId('adjacent-to-root');
+  const modal = body.querySelector('#modal');
+
+  expect(body.classList).toContain('overflow-hidden');
+
+  expect(root).toHaveAttribute('aria-hidden', 'true');
+  expect(adjacentToRoot).toHaveAttribute('aria-hidden', 'true');
+
+  expect(root.classList).toContain('pointer-events-none');
+  expect(adjacentToRoot.classList).toContain('pointer-events-none');
+
+  expect(modal).not.toHaveAttribute('aria-hidden', 'true');
+  expect(modal).toHaveAttribute('aria-labelledby', headerId);
+  expect(modal).toHaveAttribute('aria-describedby', descId);
 });
--- a/packages/fxa-react/components/Portal/index.tsx
+++ b/packages/fxa-react/components/Portal/index.tsx
@@ -5,16 +5,50 @@ import './index.scss';
 
 type PortalProps = {
   id: string;
+  headerId?: string;
+  descId?: string;
   children: React.ReactNode;
 };
 
-const Portal = ({ id, children }: PortalProps): React.ReactPortal | null => {
+const TOP_LEVEL_NONMODAL_DIVS_SELECTOR = 'body > div:not(#modal)';
+
+const setA11yOnAdjacentElementsAndBody = (els: NodeListOf<HTMLElement>) => {
+  document.body.classList.add('overflow-hidden');
+  els.forEach((el) => {
+    el.setAttribute('aria-hidden', 'true');
+    el.classList.add('pointer-events-none');
+  });
+};
+
+const resetA11yOnAdjacentElementsAndBody = (els: NodeListOf<HTMLElement>) => {
+  document.body.classList.remove('overflow-hidden');
+  els.forEach((el) => {
+    el.removeAttribute('aria-hidden');
+    el.classList.remove('pointer-events-none');
+  });
+};
+
+const Portal = ({
+  id,
+  headerId = '',
+  descId = '',
+  children,
+}: PortalProps): React.ReactPortal | null => {
   let el = document.getElementById(id);
   if (!el) {
     el = document.createElement('div');
     el.setAttribute('class', 'portal');
     el.setAttribute('id', id);
+    el.setAttribute('role', 'dialog');
     document.body.appendChild(el);
+
+    if (id === 'modal') {
+      el.setAttribute('aria-labelledby', headerId);
+      el.setAttribute('aria-describedby', descId);
+      setA11yOnAdjacentElementsAndBody(
+        document.querySelectorAll(TOP_LEVEL_NONMODAL_DIVS_SELECTOR)
+      );
+    }
   }
 
   useEffect(() => {
@@ -23,6 +57,12 @@ const Portal = ({ id, children }: PortalProps): React.ReactPortal | null => {
       let el = document.getElementById(id);
       if (el && el.children.length === 1) {
         el.remove();
+
+        if (id === 'modal') {
+          resetA11yOnAdjacentElementsAndBody(
+            document.querySelectorAll(TOP_LEVEL_NONMODAL_DIVS_SELECTOR)
+          );
+        }
       }
     };
   }, [id]);
--- a/packages/fxa-react/lib/hooks.test.tsx
+++ b/packages/fxa-react/lib/hooks.test.tsx
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-import { render, cleanup, fireEvent } from '@testing-library/react';
+import { render, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 
 import { useClickOutsideEffect, useBooleanState } from './hooks';
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -23,7 +23,7 @@
     "restart": "npm run build-postcss && pm2 restart pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "build": "tsc --build ../fxa-react && NODE_ENV=production npm run build-postcss && SKIP_PREFLIGHT_CHECK=true INLINE_RUNTIME_CHUNK=false rescripts build",
-    "test": "rescripts test",
+    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test",
     "eject": "react-scripts eject",
     "lint": "npm-run-all --parallel lint:eslint lint:sass",
     "lint:eslint": "eslint .",
--- a/packages/fxa-settings/src/components/Modal/index.tsx
+++ b/packages/fxa-settings/src/components/Modal/index.tsx
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useEffect, useRef } from 'react';
 import { useClickOutsideEffect } from 'fxa-react/lib/hooks';
 import classNames from 'classnames';
 import Portal from 'fxa-react/components/Portal';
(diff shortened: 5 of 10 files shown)


LZoog:
Going to change the base to `master` but since this was bumped to the next train, I'm going to take some time today to clean this up where I think I can make some improvements.

packages/fxa-content-server/app/images/default-profile.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="57.6" height="57.6" viewBox="0 0 57.6 57.6"><defs><linearGradient id="linear-gradient" x1="-3.82" y1="63.66" x2="56.35" y2="-0.64" gradientUnits="userSpaceOnUse"><stop offset="0.26" stop-color="#9059ff"/><stop offset="0.96" stop-color="#00b3f4"/></linearGradient></defs><g id="Layer_2" data-name="Layer 2"><g id="Layer_1-2" data-name="Layer 1"><circle cx="28.8" cy="28.8" r="24.54" style="fill:#fbfbfe"/><path d="M28.8,0A28.8,28.8,0,1,0,57.6,28.8,28.79,28.79,0,0,0,28.8,0Zm0,8.23A12.34,12.34,0,1,1,16.46,20.57,12.33,12.33,0,0,1,28.8,8.23ZM45.73,40.34a20.39,20.39,0,0,1-33.86,0,4.15,4.15,0,0,1,.28-5.05C16.5,30.27,22.53,37,28.8,37s12.3-6.76,16.65-1.74A4.15,4.15,0,0,1,45.73,40.34Z" style="fill:url(#linear-gradient)"/></g></g></svg>
```
LZoog, line 1: This might could be a separate commit but all I did was remove the `<title>` tag here, it wasn't hurting anything but I saw it in the DOM and it's not doing anything so I removed it. I like using `aria-label` instead since it's external to the SVG (also not sure if all screenreaders hit `title`).

jodyheavener: Noted, cool with me. Now that I'm looking at it, I would also recommend we consider running  SVGs through a scrubber. There's no need to have all the `id="Layer_1-2"` crud, and I think it could help with anomalies in the output (e.g. `style="fill:#fbfbfe"` could just be `fill="#fbfbfb"` I think?). I personally use [svgcleaner](https://github.com/RazrFalcon/svgcleaner-gui).

packages/fxa-react/components/Portal/index.test.tsx
```diff
@@ -44,3 +48,35 @@ it('renders multiple instances with the same ID to the same DOM parent', () => {
...
+  document.body.innerHTML = `
+    <div data-testid="root"></div>
+    <div data-testid="adjacent-to-root"></div>
+  `;
+
+  const headerId = 'some-header-id';
+  const descId = 'some-desc-id';
+
+  const { getByTestId } = render(
+    <Portal id="modal" {...{ headerId }} {...{ descId }}>
```
jodyheavener, line 62: I think this can be `{...{ headerId, descId }}`?

packages/fxa-react/components/Portal/index.test.tsx
```diff
@@ -44,3 +48,35 @@ it('renders multiple instances with the same ID to the same DOM parent', () => {
...
+      <div>Hi mom</div>
+    </Portal>
+  );
+
+  const body = document.body;
+  const root = getByTestId('root');
+  const adjacentToRoot = getByTestId('adjacent-to-root');
+  const modal = body.querySelector('#modal');
+
+  expect(body.classList.contains('overflow-hidden')).toBe(true);
```
jodyheavener, line 72: This might be simplified with `expect(body.classList).toContain('overflow-hidden')`

packages/fxa-react/components/Portal/index.test.tsx
```diff
@@ -44,3 +48,35 @@ it('renders multiple instances with the same ID to the same DOM parent', () => {
...
+
+  expect(root).toHaveAttribute('aria-hidden', 'true');
+  expect(adjacentToRoot).toHaveAttribute('aria-hidden', 'true');
+  expect(root.classList.contains('pointer-events-none')).toBe(true);
+  expect(adjacentToRoot.classList.contains('pointer-events-none')).toBe(true);
```
jodyheavener, line 77: Same idea with these two

packages/fxa-react/components/Portal/index.tsx
```diff
@@ -5,16 +5,48 @@ import './index.scss';
 
 type PortalProps = {
   id: string;
+  headerId?: string;
+  descId?: string;
```
jodyheavener, line 9: Do these need to be optional if they get defaulted to `''` when `Portal` is instantiated?

packages/fxa-react/lib/hooks.test.tsx
```diff
@@ -2,40 +2,40 @@
...
-    expect(getByTestId("result")).toHaveTextContent("false");
-    fireEvent.click(getByText("setTrue"));
-    expect(getByTestId("result")).toHaveTextContent("true");
-    fireEvent.click(getByText("setFalse"));
-    expect(getByTestId("result")).toHaveTextContent("false");
+    expect(getByTestId('result')).toHaveTextContent('false');
+    fireEvent.click(getByText('setTrue'));
+    expect(getByTestId('result')).toHaveTextContent('true');
+    fireEvent.click(getByText('setFalse'));
+    expect(getByTestId('result')).toHaveTextContent('false');
```
jodyheavener, line 29: Rebasing might get rid of these prettier diff changes

packages/fxa-settings/package.json
```diff
@@ -23,7 +23,7 @@
     "restart": "npm run build-postcss && pm2 restart pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "build": "NODE_ENV=production npm run build-postcss && INLINE_RUNTIME_CHUNK=false rescripts build",
-    "test": "rescripts test",
+    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test",
```
jodyheavener, line 26: Ahaha you did this too! Was it because of a babel-loader (or similar) mismatch?

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -2,39 +2,61 @@
...
+  useEffect(() => {
+    if (tabFenceRef.current) {
+      tabFenceRef.current.focus();
+    }
+  }, []);
+
+  // close on esc keydown
+  useEffect(() => {
+    const handler = ({ key }: KeyboardEvent) => {
+      if (key === 'Escape') {
```
jodyheavener, line 43: Is `Escape` reliable here? Or do we need to check a KeyCode?

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -2,39 +2,61 @@
...
+        onDismiss();
+      }
+    };
+    window.addEventListener('keydown', handler);
+    return () => window.removeEventListener('keydown', handler);
+  }, [onDismiss]);
+
   return (
-    <Portal id="modal">
+    <Portal id="modal" {...{ headerId }} {...{ descId }}>
```
jodyheavener, line 52: Can these be combined?

packages/fxa-content-server/app/images/default-profile.svg
LZoog: Actually I'll run this through SVGO, I'm not sure if it was before.

The problem with automatically running SVGs through a scrubber is that you can't attach image metadata that persists in git, or at least you couldn't a couple of years ago when I tried to do just this, so we wouldn't know what's been ran through and what hasn't without some flag, and then very occasionally, throwing an SVG through one automatically can remove something it shouldn't have and it looks funky and you won't know until after it's merged and noticed in staging. 😓 I'm all ears for a solution on both fronts though because I'd love to automate that.

packages/fxa-react/components/Portal/index.test.tsx
LZoog: Ah it totally can, you know I'm all about that syntactic sugar 🎉

packages/fxa-react/components/Portal/index.tsx
LZoog: Yes, it doesn't type check on defaults, but what's passed in.

packages/fxa-settings/package.json
LZoog: I had to add this because it wouldn't run tests without it and suggested adding this🤔

packages/fxa-content-server/app/images/default-profile.svg
jodyheavener: Oh that makes perfect sense. Yeah. Without having looked into it, I would suspect there should be a way, a la linting ignore statements, to ignore the next line / a specific tag? If not, it sounds like a great PR for one of the tools :D

packages/fxa-settings/src/components/Modal/index.tsx
LZoog: Yeah good call out, when I actually made this I wondered why 'Escape' worked so effortlessly, but I did look up examples doing this and they don't use a key code. I'll double check this will be reliable.

packages/fxa-settings/package.json
jodyheavener: Same!

packages/fxa-react/components/Portal/index.test.tsx
jodyheavener: ✨

packages/fxa-content-server/app/images/default-profile.svg
LZoog: > Without having looked into it, I would suspect there should be a way, a la linting ignore statements, to ignore the next line / a specific tag?

Maybe. We'd have to actually open/read the file and parse through the text for a particular flag (maybe a specific class name or something) and then add that if it successfully ran through, and skip any SVGs with that flag. The other problem isn't anything to scoff at though, I have been burned _many_ times by sending something through SVGO and pushing it up without checking it after processing and for some reason the problems always occured in not just IE but FF as well. If we actually build some type of CLI tool that asked you to confirm the SVG looked correct before saving it I think that could work and would be really neat, but that's a lot of work for as few SVGs as we have.

packages/fxa-settings/src/components/Modal/index.tsx
LZoog: So it turns out `key` [has wide support now](https://caniuse.com/#search=KeyboardEvent.key). We don't support IE11 so this should be fine. (`keyCode` is [actually deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode))

FxaStorybooks:
# Storybook Deployment
Pull Request: https://mozilla-fxa.github.io/storybooks/pulls/5391/
Commit: https://mozilla-fxa.github.io/storybooks/commits/22caaa96e48d145ca1dcac03d7ea886c43a3ed61/

FxaStorybooks:
# Storybook Deployment
Pull Request: https://mozilla-fxa.github.io/storybooks/pulls/5391/
Commit: https://mozilla-fxa.github.io/storybooks/commits/03e2c52e8f702ca15b5c40aeb5793224ca3f7ec1/

FxaStorybooks:
# Storybook Deployment
Pull Request: https://mozilla-fxa.github.io/storybooks/pulls/5391/
Commit: https://mozilla-fxa.github.io/storybooks/commits/471b62adc69ba8820edc24045b41362694760652/

packages/fxa-settings/src/components/Modal/index.tsx
jodyheavener: Hooray!

packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const prevRevealedRef = useRef(revealed);
+  const prevRevealed = prevRevealedRef.current;
+
+  useEffect(() => {
+    if (revealed !== undefined) {
+      prevRevealedRef.current = revealed;
+    }
+    if (triggerElement.current && prevRevealed === true && revealed === false) {
+      triggerElement.current.focus();
+    }
```
jodyheavener, line 22: Clever

packages/fxa-react/components/Portal/index.tsx
```diff
@@ -23,6 +55,12 @@ const Portal = ({ id, children }: PortalProps): React.ReactPortal | null => {
       let el = document.getElementById(id);
       if (el && el.children.length === 1) {
         el.remove();
+
+        if (id === 'modal') {
+          resetA11yOnAdjacentElementsAndBody(
+            document.querySelectorAll('body > div:not(#modal)')
```
jodyheavener, line 61: Given the specificity and re-use, `body > div:not(#modal)` might be better suited to a constant

packages/fxa-react/components/Portal/index.tsx
```diff
@@ -5,16 +5,48 @@ import './index.scss';
...
     el.setAttribute('class', 'portal');
     el.setAttribute('id', id);
+    el.setAttribute('role', 'dialog');
     document.body.appendChild(el);
+
+    if (id === 'modal') {
+      el.setAttribute('aria-labelledby', headerId);
+      el.setAttribute('aria-describedby', descId);
+      setA11yOnAdjacentElementsAndBody(
+        document.querySelectorAll('body > div:not(#modal)')
```
jodyheavener, line 47: So, if I'm understanding this correctly, this will only really target `#root`, right?

LZoog: Right now, yes, and it'll likely stay that way, but I wanted to future-proof this in case something else is ever pulled outside of and set adjacent to `#root`.

FxaStorybooks:
# Storybook Deployment
Pull Request: https://mozilla-fxa.github.io/storybooks/pulls/5391/
Commit: https://mozilla-fxa.github.io/storybooks/commits/0366ffa8466411aa40aa9e2bce1c88fe77225894/

PR #5392, opened 2020-05-21
fix(test): grep teamcity intern.js for pattern

jrgm:
r? - @vbudhram


packages/fxa-content-server/tests/teamcity/run.sh
```diff
@@ -67,6 +67,12 @@ else
 fi
 echo "GIT_COMMIT          $GIT_COMMIT"
 
+# optionally run a pattern of tests
+if [ -z "$GREP" ]; then
+  GREP=""
+fi
```
jrgm, line 73: Well, that's what you get from doing it live. I was thinking of doing something more there, but instead I kinda left some no-op logic o_O.

PR #5399, opened 2020-05-21
fix(tests): support optional --suites

jrgm:
r? - @vbudhram 

Seemed possibly useful to expose this, so I did.


jrgm:
https://github.com/mozilla/fxa/pull/5399/files?w=1 is more readable

PR #5401, opened 2020-05-21
refactor(logging): reduce more logging

### LINKED ISSUE #5306: [auth-server] lower logging of metricsEvents.emitFlowEvent "missingFlowId":true
Thanks for the fix in https://github.com/mozilla/fxa/pull/5223 

I was looking again at our log usage for auth server, and there remains this kind of missingFlowId message:
```
... "Logger":"fxa-auth-server","Type":"metricsEvents.emitFlowEvent","Severity":2,
"Pid":1,"EnvVersion":"2.0","Fields":{"event":"route./verify.200","missingFlowId":true}}
```
and usually its for the verify.200 route. It counts for ~12% of the log volume in bytes now (after other reductions https://github.com/mozilla/fxa/pull/5223 and for geodb in https://github.com/mozilla/fxa/pull/5139).

That's from https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/metrics/events.js#L212-L216. If that's not actionable, perhaps that error can be lowered to log.trace level (?).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1880)
┆Issue Number: FXA-1880

clouserw:
Because:

* At the higher logging level it fills up our production logs.  At a

This commit:

* Changes missingFlowId logging from `info` to `trace`

Closes #5306

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/events.js
+++ b/packages/fxa-auth-server/lib/metrics/events.js
@@ -210,7 +210,7 @@ module.exports = (log, config) => {
 
           log.flowEvent(data);
         } else if (!OPTIONAL_FLOW_EVENTS[event]) {
-          log.error('metricsEvents.emitFlowEvent', {
+          log.trace('metricsEvents.emitFlowEvent', {
             event,
             missingFlowId: true,
           });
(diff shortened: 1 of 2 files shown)


clouserw:
Let's see how the tests look on this.  My tests fail with no output locally (....)

PR #5402, opened 2020-05-21
refactor(logging): reduce some unnecessary logging

### LINKED ISSUE #5311: [profile] possibly switch batch.cached logline to statsd
About 40% of the log _lines_ for profile server are like this line:
```
{"Timestamp":1589406163905000000,"Logger":"fxa-profile-server",
"Type":"routes.profile.batch.cached","Severity":6,"Pid":1,"EnvVersion":"2.0",
"Fields":{"storedAt":1589371622438,"ttl":8658533}}
```
Not to over-optimize logging, but perhaps the `ttl` could be better used as a sampled statsd value, with the logging being reduced to trace level. - https://github.com/mozilla/fxa/blob/master/packages/fxa-profile-server/lib/routes/profile.js#L75-L81

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1883)
┆Issue Number: FXA-1883

clouserw:
Because:

* At the higher logging level it fills up our production logs.  At a

This commit:

* Changes cache hit logging from `info` to `trace`

Closes #5311

### DIFF
--- a/packages/fxa-profile-server/lib/routes/profile.js
+++ b/packages/fxa-profile-server/lib/routes/profile.js
@@ -71,13 +71,13 @@ module.exports = {
       }
       const lastModified = cached ? new Date(cached.stored) : new Date();
       if (cached) {
-        logger.info('batch.cached', {
+        logger.trace('batch.cached', {
           storedAt: cached.stored,
           error: report && report.error,
           ttl: cached.ttl,
         });
       } else {
-        logger.info('batch.db');
+        logger.trace('batch.db');
       }
 
       return rep.header('last-modified', lastModified.toUTCString());


clouserw:
I rebased thinking that might help the tests.  :crossed_fingers:

clouserw:
rebased again...

packages/fxa-profile-server/lib/routes/profile.js
```diff
@@ -71,13 +71,13 @@ module.exports = {
       }
       const lastModified = cached ? new Date(cached.stored) : new Date();
       if (cached) {
-        logger.info('batch.cached', {
+        logger.trace('batch.cached', {
```
vladikoff, line 74: this logging is important to make sure that profile server caching is working. If we make this `trace`, we just want to make sure that we will be able to tell if something breaks in production w.r.t to the cache. It can also be that we haven't monitored this logging since Kibana days.

clouserw: I don't think we have.  jrgm mentioned in the issue that statsd would be a better alternative to logger if getting exact cache numbers was important.

Does watching general cache metrics answer the questions you'd have if we made this change?

PR #5404, opened 2020-05-21
fix(support): don't exit the form when you hit enter

### LINKED ISSUE #5377: Hitting "enter" in the support form exits
Noticed that if you're in a text field or dropdown on the `/support` form it will exit the form. If anything it should submit the form, or at least confirm that you want to exit.

I believe it's because the Cancel button in that form needs a `type="button"` on it.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1925)
┆Issue Number: FXA-1925

jodyheavener:
Closes #5377

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/support.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/support.mustache
@@ -76,7 +76,7 @@
                 </div>
               </div>
               <div class="button-row">
-                <button class="settings-button secondary-button cancel">{{#t}}Cancel{{/t}}</button>
+                <button class="settings-button secondary-button cancel" type="button">{{#t}}Cancel{{/t}}</button>
                 <button type="submit" class="settings-button primary-button" disabled>
                   <span class="submit-content">{{#t}}Send{{/t}}</span>
                   <div class="spinner hidden"></div>
--- a/packages/fxa-content-server/tests/functional/lib/helpers.js
+++ b/packages/fxa-content-server/tests/functional/lib/helpers.js
@@ -347,6 +347,17 @@ const type = thenify(function (selector, text, options = {}) {
     .end();
 });
 
+/**
+ * Send a keycode to the selector
+ *
+ * @param {string} selector
+ * @param {string} key
+ * @returns {promise}
+ */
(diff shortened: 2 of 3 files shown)


jodyheavener:
I'd like to test this with a native Enter key press, but I'm having trouble setting that up with our current test setup. Something like...

```
const typeNative = thenify(function(selector, key) {
  return this.parent
    .findByCssSelector(selector)
    .sendKeys(Key[key])
    .end();
});

// usage
.then(typeNative('#topic_chosen a.chosen-single', 'ENTER'))
```

But not sure where to look up Selenium keys.

jackiemunroe:
@jodyheavener could you rebase this onto master so that it using `yarn` instead of `npm` for package management?

jackiemunroe:
> I'd like to test this with a native Enter key press, but I'm having trouble setting that up with our current test setup. Something like...
> 
> ```
> const typeNative = thenify(function(selector, key) {
>   return this.parent
>     .findByCssSelector(selector)
>     .sendKeys(Key[key])
>     .end();
> });
> 
> // usage
> .then(typeNative('#topic_chosen a.chosen-single', 'ENTER'))
> ```
> 
> But not sure where to look up Selenium keys.
https://www.selenium.dev/documentation/en/webdriver/keyboard/

jodyheavener:
Hmm, yeah, I got it working by installing the `selenium-webdriver` package but I thought I'd be able to import something from one of our existing testing deps. Just not sure how it's all set up with Intern 🤔

6a68:
> Hmm, yeah, I got it working by installing the `selenium-webdriver` package but I thought I'd be able to import something from one of our existing testing deps. Just not sure how it's all set up with Intern 🤔

@jodyheavener Intern uses a browser driver called leadfoot, which has a `pressKeys` method you can use: https://theintern.io/leadfoot/module-leadfoot_Session.html#pressKeys

There's a ton of reusable code in fxa-content-server/tests/functional/lib/helpers.js, you can skim the list of helpers at the bottom of the file. Surprisingly, though, I don't see one that handles non-printable characters.

jodyheavener:
awesome thanks @6a68!

PR #5405, opened 2020-05-21
1838 survey rules config

meandavejustice:


chenba:
Note: even though @meandavejustice created this PR, he's also the reviewer because the commit's actually from me.  (But I cannot set @meandavejustice as the reviewer though.)

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -0,0 +1,240 @@
...
+// All but the default export here is to make testing easier.
+
+const falseFn = () => false;
+export const NONE = Symbol();
+
+export const participatedRecently = (
+  previousParticipationTime,
+  doNotBotherSpan
+) => {
+  if (!previousParticipationTime) return true;
```
meandavejustice, line 16: I think that this line should return false

chenba: I think it needs a stricter check, if the reason for that suggestion is that I passed zeroes in some tests.  That was my bad; I'll fix those tests.  And I'll update this to check for `null` and `undefined`, and to return true on those.

chenba: Oh actually, those zeroes were for a different param.  Could you explain why you think this should return `false`?  If `surveyTime` didn't exist, it would mean that the user has never taken a survey before.

chenba: Oh I see the problem now. Cool, thanks for fixing this.  I'll keep the null and undefined check around.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -0,0 +1,243 @@
...
+export const participatedRecently = (
+  previousParticipationTime,
+  doNotBotherSpan
+) => {
+  if (
+    previousParticipationTime === null ||
+    previousParticipationTime === undefined
+  ) {
+    return false;
+  }
```
chenba, line 21: @meandavejustice Updated this and added a corresponding test at https://github.com/mozilla/fxa/pull/5405/files#diff-913ef57336afdd82b4c000a1aedc0056R801-R812

PR #5406, opened 2020-05-21
bug (content server): fix support form success message

### LINKED ISSUE #5298: Contact support confirmation popup message is incomplete
#### Affected versions:
- All Firefox

#### Affected Platforms:
- Windows 10 (64-bit)

#### Preconditions
- existent FxA account with active subscription

#### Steps to reproduce
1. Open /subscriptions page: https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/subscriptions
2. Click the “Contact Support” button
3. Fill out all fields and click “Send”
4. Observe confirmation popup message 

#### Expected result:
- There are no issues in the confirmation message..

#### Actual result:
The value of the “What are you contacting us about?” field is missing.

#### Notes:
- Reproducing for both (Firefox Guardian/Other) options.

for

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1875)
┆Issue Number: FXA-1875

jackiemunroe:
Because:
* the support form success message is missing the product value

This commit:
* fix the message creation function to reference the correct form value
* remove the 'FxA - ' prefix that is added for Zendesk Agents, to match mockups and to show users the value they selected accurately

fixes #5298

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/support.js
+++ b/packages/fxa-content-server/app/scripts/views/support.js
@@ -25,6 +25,8 @@ import SupportFormErrorTemplate from 'templates/partial/support-form-error.musta
 import SupportFormSuccessTemplate from 'templates/partial/support-form-success.mustache';
 import Template from 'templates/support.mustache';
 
+const productPrefix = 'FxA - ';
+
 const t = (msg) => msg;
 
 const proto = BaseView.prototype;
@@ -165,8 +167,11 @@ const SupportView = BaseView.extend({
   },
 
   formatProductName: function (name) {
-    const productNamePrefix = 'FxA - ';
-    return `${productNamePrefix}${name}`;
+    return `${productPrefix}${name}`;
+  },
+
+  unformatProductName: function (name) {
+    return name.slice(productPrefix.length);
   },
 
   submitButtonUIToggle: function () {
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -162,24 +159,28 @@ const SupportView = BaseView.extend({
...
 
-  formatProductName: function(name) {
+  formatProductName: function (name) {
     const productNamePrefix = 'FxA - ';
     return `${productNamePrefix}${name}`;
   },
 
-  submitButtonUIToggle: function() {
+  unformatProductName: function (name) {
+    const productNamePrefix = /^FxA - /;
```
chenba, line 173: With this, the `FxA - ` prefix is defined in two places; we should move the one on line 168 out to a wider scope so this function can use it.  It also needn't to be regex.  The prepending always happens so we could just strip the prefix without searching.

packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -162,24 +159,28 @@ const SupportView = BaseView.extend({
...
+    return name.replace(productNamePrefix, '');
+  },
```
chenba, line 175: nit^2: newline here

packages/fxa-content-server/app/scripts/views/support.js
jackiemunroe: sounds good!

PR #5409, opened 2020-05-21
feat(dev): added post-checkout git hook to keep deps in sync

dannycoates:
Submitting this as a proposal... It adds a tiny bit of lag when doing a `git checkout` but I've found it useful on other projects. Now that there's only one file to diff it seems worth adding.

It's easy to not notice when to run a `yarn install` either
when switching branches or doing a git pull. This adds a
hook to diff your previous yarn.lock with the one you're changing
to, and if there's changes running 'yarn install' for you.

### DIFF
--- a/_scripts/git-checkout-hook.sh
+++ b/_scripts/git-checkout-hook.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+IFS=' '
+read -ra G_PARAMS <<< "$HUSKY_GIT_PARAMS"
+PREV=${G_PARAMS[0]}
+NEXT=${G_PARAMS[1]}
+if [ "$PREV" != "$NEXT" ]; then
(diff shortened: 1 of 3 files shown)


6a68:
I'm fine either way on this proposal, as long as there's a one-liner to disable it if I'm just peeking at different branches while the server's running. (It seems like the git checkout > yarn install would then trigger a server restart as well, is that right?)

dannycoates:
I'd be cool with an opt-in env variable we can add to our shell rc files.

rfk:
As a nowadays-occasional contributor who rarely pulls from master, this would be very helpful to me, especially if it also worked on a `git pull` that updates the lockfile.

dannycoates:
@6a68 I made it opt-in with a console message of how to do so. r?

lmorchard:
If this defaults to on, we should add a note about it in README. I kind of hate adding friction to common git commands, but I haven't lived with it yet so 🤷

dannycoates:
Let's try it out. We can revert if we hate it

PR #5410, opened 2020-05-22
chore(tests): clean up old Stripe customers after functional tests

### LINKED ISSUE #5326: Add Script to Remove Test Users from Dev Stripe
Our functional tests create subscriptions with test users on SUB_PLAT_DEV.  We should have a script that we can run periodically to remove those users.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1893)
┆Issue Number: FXA-1893

lmorchard:
fixes #5326

So I *thought* there weren't going to be any code changes for this. Looks like there probably should be a few.

Also, rather than run this on a schedule, it seemed maybe easier to just tack it on after content-server functional tests. The script deletes customers with an email pattern of 'signin.*@restmail.net' that are over a day old. That should only cover testing customers and leave them around for a little while in case we have to peek at them after failed tests. Running this after each batch of functional tests should be pretty quick.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,11 @@ commands:
             CIRCLE_NODE_INDEX: << parameters.index >>
             CIRCLE_NODE_TOTAL: << parameters.total >>
           command: ./.circleci/test-package.sh fxa-content-server
+      - run:
+          name: Clean up old Stripe customers from testing
+          command: |
+            cd packages/fxa-auth-server && \
+            yarn run clean-up-old-ci-stripe-customers
       - store_artifacts:
           path: ~/.pm2/logs
           destination: logs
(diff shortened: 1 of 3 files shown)


lmorchard:
Oops - also meant to only open this as a draft after I saw at least one successful CI test run, but I guess YOLO

lmorchard:
Alright, looks like I finally got tests to pass after a few retries for unrelated transient failures. Got this output from one of the trials, which tells me the cleanup works:

```
#!/bin/bash -eo pipefail
cd packages/fxa-auth-server && \
yarn run clean-up-old-ci-stripe-customers

cus_HJwmK2SeEQeTVx - email: signin0.758865834641852@restmail.net uid: 6b59354b587e4255a6be9b86753f0771 age: 86446.55399990082
Deleted cus_HJwmK2SeEQeTVx
Deleted 1 customers.
```

.circleci/config.yml
```diff
@@ -44,6 +44,11 @@ commands:
             CIRCLE_NODE_INDEX: << parameters.index >>
             CIRCLE_NODE_TOTAL: << parameters.total >>
           command: ./.circleci/test-package.sh fxa-content-server
+      - run:
+          name: Clean up old Stripe customers from testing
+          command: |
+            cd packages/fxa-auth-server && \
+            yarn run clean-up-old-ci-stripe-customers
```
dannycoates, line 51: should this have a `when: always`? https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute

lmorchard: Eh, probably not. It only deletes customers over a day old, so as long as *some* CI run executes this at some point, it'll catch things

PR #5431, opened 2020-05-24
Mass code reformat

dannycoates:
consolidate to one prettier config and reformat.

see https://medium.com/moxio/ignoring-bulk-change-commits-with-git-blame-5d50835e45c2 for how to ignore this change in git blame.


jodyheavener:
So I know most packages can be prettified with their respective `format` scripts, or through `lint-staged`, but I noticed VS Code is no longer auto-formatting as I type/paste. Is there a way to fix this?

jodyheavener:
Nevermind, found the setting.

rfk:
> Nevermind, found the setting.

So uh, what's the setting @jodyheavener? Asking for a friend...

jodyheavener:
@rfk in VS Code settings search for `prettier.configPath` and set it to the full system path of `_dev/.prettierrc`

PR #5436, opened 2020-05-25
refactor tsconfigs

dannycoates:
Extends #5430 to refactor all tsconfig files and update `require` paths and moves all custom ts type declarations to `/types`


dannycoates:
Adding @jodyheavener for review since you seem stoked about it :)

packages/fxa-auth-server/lib/geodb.js
```diff
@@ -5,7 +5,7 @@
 'use strict';
 
 const config = require('../config').get('geodb');
-const geodb = require('../../fxa-geodb')(config);
+const geodb = require('fxa-geodb')(config);
```
jodyheavener, line 8: How is this working? Does node know to resolve this somewhere (like fxa-react), and if so can this change be extended to the other relative require statements?

types/fxa-jwtool/index.d.ts
```diff
@@ -177,7 +177,7 @@ declare namespace fxa_jwtool {
 
   type PublicKeyType = {
     e: string;
-    'fxa-createdAt'?: number;
+    "fxa-createdAt"?: number;
```
jodyheavener, line 180: Is prettier skipping this file?

packages/fxa-auth-server/lib/geodb.js
dannycoates: it's resolve through package.json using `"fxa-geodb": "workspace:*"` and yes we should be using this style everywhere. idk why i didn't catch them all with my grep

types/fxa-jwtool/index.d.ts
dannycoates: weird, idk

PR #5460, opened 2020-05-27
feat(settings): Add header component

### LINKED ISSUE #4865: [header/footer] Initial stubout of sticky header
We can likely move the `Header` component from the `fxa-admin-panel` package into the shared location to use here as well. While elements can be conditionally rendered to account for differences, changes needed for Settings should take priority over the admin panel since that was designed by engineering with Settings in mind. In other words, it's fine for the header in `fxa-admin-panel` to change stylistically and reflect Settings styles. If it can't be used for whatever reason, it can be used as a reference for semantics, styling, placement, and prop passing references. Make sure the component is a `<header>` component with `role="banner"`.

The header should have dummy Bento/Avatar placements (functionality will be implemented in follow up issues).

The header should also reflect a `position: sticky` style.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1573)
┆Issue Number: FXA-1573

LZoog:
Desktop:
image

Tablet:
image

Mobile:
image

Functionally done as the ticket is for a dummy component, but still need to:
* ~Go look at what I broke in `fxa-admin-panel`~ Temporary styles are added until we add Tailwind in a follow up.
* ~Add things to Storybook~
* ~Figure out what we're doing about shared classes in `fxa-react`, might need to setup Tailwind there for Storybook and we need to determine what our convention is between setting "default" classes vs requiring all classes to be passed in vs allowing style overrides~ Jody and I discussed this, we'll need a follow up issue.
* ~Add some tests~
* ~Run SVGs through SVGO~

fixes #4865


packages/fxa-admin-panel/src/components/AppLayout/index.scss
```diff
@@ -30,3 +30,24 @@
...
+    color: var(--color-white);
+    font-size: var(--header-font-size-lg);
+    font-weight: var(--font-weight-semi-bold);
+    margin: 0 0 0 20px;
+    text-shadow: 1px 1px 3px var(--color-black);
+  }
+
+  .logo {
+    width: 50px;
+  }
```
dannycoates, line 52: If I'm reading this right, most of these can go in the tsx?

packages/fxa-admin-panel/src/components/AppLayout/index.scss
```diff
@@ -30,3 +30,24 @@
 main {
   flex: 4;
 }
+
+// custom styles from when Header lived in this package; it has since been
+// pulled into `fxa-shared` and the leftover styles were placed here.
```
jodyheavener, line 35: 👍

packages/fxa-settings/src/styles/tailwind.css
```diff
@@ -4,4 +4,5 @@
 
 body {
   @apply text-base font-body;
+  @apply bg-grey-50;
```
jodyheavener, line 7: Is this the right grey?

packages/fxa-react/components/LogoLockup/index.tsx
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { ReactElement } from 'react';
+import { ReactComponent as Logo } from '../../images/ff-logo.svg';
```
jodyheavener, line 6: Can this ff-logo not live adjacent to the component files? Ideally we are keeping component parts close to each other.

packages/fxa-settings/src/styles/tailwind.css
LZoog: It's not, I know you added `-20` in your PR, I'm going to add that here and update it👍

packages/fxa-react/components/LogoLockup/index.tsx
LZoog: We talked about doing that sometime ago in payments I believe and determined we wanted a separate `images` folder in case something needed to be reused. It could be argued that the logo could live in the `LogoLockup` directory, but we'd need to duplicate it if we needed it somewhere else, especially if for whatever reason we used a different component for the login flow or in payments down the line, and we want to reference just one SVG to make sure they're consistent... The `default-avatar` in particular is going to see a lot of reuse across the first `UnitRow` in the `Header`, and in the doorhanger drop down.

jodyheavener: Right, that makes sense, I just assumed we would default to standalone _until_ it needs to be re-used, instead of defaulting to shared. Because the opposite is also possible, where we think we might re-use something, leave it in shared images, and then never re-use it. Admittedly that's less likely to happen with the FF logo, but it gives me pause. Also, sorry if this sounds naggy 🙈

LZoog: It definitely doesn't sound naggy, this is definitely the time to bring up our conventions and file placements. :D I was mostly following our convention we set in payments - I personally think I'd prefer our images to be in one central spot than some in a central spot and some not, but if you prefer the latter, maybe make an issue so we can discuss / make that convention standard in payments as well?

jodyheavener: image

packages/fxa-admin-panel/src/components/AppLayout/index.scss
LZoog: Just to circle back, I added a more descriptive comment above. We need to get Tailwind setup in the admin panel and styles showing up in `fxa-react` Storybook, likely Jody or I will work on that next sprint.

packages/fxa-admin-panel/src/components/AppLayout/index.scss
```diff
@@ -30,3 +30,40 @@
 main {
   flex: 4;
 }
+
+// TODO: Setup Tailwind in admin-panel and remove these custom styles.
```
jodyheavener, line 34: I can't remember, is this in a new ticket?

packages/fxa-react/components/Header/index.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Header } from './index';
+import { LogoLockup } from '../LogoLockup';
+
+storiesOf('components/Header', module)
+  .add('basic', () => (
+    <Header left={<div>left content</div>} right={<div>right content</div>} />
+  ))
```
jodyheavener, line 9: Do we need basic anymore? I believe the header is always going to have the logo, no?

jodyheavener: Mmm you know what, cancel that. I see this is in fxa-react and could very well be used without a logo as the `left` prop.

jodyheavener: Though I would be curious to know the use case.. 🤔

packages/fxa-react/components/types.d.ts
```diff
@@ -3,11 +3,9 @@
 // e.g. `import { ReactComponent as Logo } from './logo.svg';`
 declare module '*.svg' {
   import * as React from 'react';
-
   export const ReactComponent: React.FunctionComponent<
     React.SVGProps<SVGSVGElement> & { title?: string }
   >;
-
```
jodyheavener, line 10: Oh I see how it is.

packages/fxa-react/storybooks.js
```diff
@@ -29,18 +29,27 @@ const customizeWebpackConfig = ({ config, mode }) => {
...
-    config.module.rules.unshift({
-      test: /\.svg$/,
-      use: ['@svgr/webpack', assetLoader],
-    });
-    config.module.rules = [{ oneOf: config.module.rules }];
+  } else {
+    assetLoader = {
+      loader: resolve(
+        __dirname,
+        '../../node_modules/@storybook/core/node_modules/file-loader/dist/cjs.js'
```
jodyheavener, line 43: It might be worth throwing a little explainer above this. Something like "fxa-react provides its own file-loader, but fxa-settings doesn't, YMMV with additional storybook setups". I can see this being a somewhat fragile setup.

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Help
+        aria-label="Help"
+        title="Help"
+        role="img"
+        className="w-6"
+        data-testid="header-help"
+      />
+    </LinkExternal>
+    <Bento
+      aria-label="Firefox Account Menu"
```
jodyheavener, line 49: Nit: should this be sentence case? "Default avatar" below is

packages/fxa-react/components/Header/index.stories.tsx
LZoog: Yeah, it's purposefully agnostic so you can pass whatever React elements you need to in the prop, I'm not sure when we wouldn't use it with the logo but this way you can pass whatever you want in there.

packages/fxa-settings/src/components/HeaderLockup/index.tsx
LZoog: Eh, it could be just for consistency sake, but screenreaders don't care about casing thankfully. :D

packages/fxa-react/storybooks.js
LZoog: Yeah, good call. I will do this in a follow up PR shortly after this merges, I don't want this to go through CI again for a comment.

packages/fxa-admin-panel/src/components/AppLayout/index.scss
LZoog: Yes, this is part of that ticket I'm making👍

PR #5461, opened 2020-05-27
refactor(logging): reduce profile cache logging from info to verbose

dannycoates:
At the higher logging level it fills up our production logs.  At a

supersedes #5402


clouserw:
image

dannycoates:
@clouserw I think yours fails because `trace` gets a stacktrace. we'll see 🎰

clouserw:
I'll take it!  Thanks @dannycoates

PR #5462, opened 2020-05-27
feat(metrics): Emit statsd metrics around push message sending.

### LINKED ISSUE #5367: Add statsd metrics around push notifications
The Sync team has recently been investigating the end-to-end quality of Send Tab, and one of the pieces we needed to dig in to was the success rate of sending push notifications to connected devices. See [here](https://docs.google.com/document/d/1XxNHWuotwNpaOSzrdN3KPu0bFPe28SSxZZp4HNWezhM/edit#heading=h.lsh1ooublr1l) for the ad-hoc analysis I was able to do on top of the existing logs in bigquery.

I'd like to make it easier to keep an eye on these metrics in future, and maybe even allow for some alerting if the success-rate of send-tab push delivery falls below a certain level. I believe that statsd is the preferred mechanism for this in FxA these days, so my strawman proposal is to modify the [`incrementPushAction`](https://github.com/mozilla/fxa/blob/b49b01a260ea25164a44341d5c3014f925352e48/packages/fxa-auth-server/lib/push.js#L105) to emit stats metrics such as:

* `statsd.increment("push.send", tags)` when we attempt to send a push
* `statsd.increment("push.send.success", tags)` when that succeeds
* `statsd.increment("push.send.error", tags)` when that fails

We can use tags to provide more fine-grained info, such as:

* `tags["reason"] = "accountVerify"|"commandReceived"|...` for the type of push being sent
* `tags["error"] = "no_callback"|"no_keys"|"reset"|...` for the cause of a failure to send
* `tags["ua_os"] = "ios"|"android"|"windows"|...` for the platform of the target device
* Maybe uaBrowserVersion etc here as well? I'm not sure how thrifty we need to be with possible tags.

This these metrics going to statsd, we could set up a dashboard for easy viewing and easy exploration by the different tags.

@bbangert what do you think, does this sound like a sensible approach?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1917)
┆Issue Number: FXA-1917

rfk:
Because:

- Robust delivery of push messages to connected devices is an
  important part of "send tab" and other account-related experiences.
- It's hard to monitor or alert on changes to push delivery success
  rate.

This commit:

- Emits "attempt", "success" and "failure" counters to statsd,
  where they can be graphed and monitored more easily.
- Lightly refactors and adds additional metadata to push-related
  logs that go into bigquery, based on experience from recent
  ad-hoc log exploration.

Fixes #5367.

(To review this, please [ignore identation changes](https://github.com/mozilla/fxa/pull/5462/files?w=1))

### DIFF
--- a/packages/fxa-auth-server/lib/push.js
+++ b/packages/fxa-auth-server/lib/push.js
@@ -8,13 +8,6 @@ const crypto = require('crypto');
 const base64url = require('base64url');
 const webpush = require('web-push');
 
-const ERR_NO_PUSH_CALLBACK = 'No Push Callback';
-const ERR_PUSH_CALLBACK_EXPIRED = 'Push Callback expired';
-const ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)';
-const ERR_TOO_MANY_DEVICES = 'Too many devices connected to account';
-
-const LOG_OP_PUSH_TO_DEVICES = 'push.sendPush';
-
 const PUSH_PAYLOAD_SCHEMA_VERSION = 1;
 const PUSH_COMMANDS = {
   DEVICE_CONNECTED: 'fxaccounts:device_connected',
@@ -26,6 +19,49 @@ const PUSH_COMMANDS = {
   COMMAND_RECEIVED: 'fxaccounts:command_received',
 };
 
+const PUSH_REASONS = new Set([
+ 'accountVerify',
+ 'accountConfirm',
+ 'passwordReset',
+ 'passwordChange',
+ 'deviceConnected',
+ 'deviceDisconnected',
+ 'profileUpdated',
+ 'devicesNotify',
+ 'accountDestroyed',
+ 'commandReceived',
+]);
+
+const PUSH_ERRORS = new Set([
+  "noCallback",
+  "noKeys",
+  "expiredCallback",
+  "resetCallback",
+  "serverError",
+  "unknown",
+]);
+
+const PUSH_ERROR_MESSAGES = {
+  "noCallback": "No push callback",
+  "noKeys": "Data payload preset but missing keys",
+  "expiredCallback": "Push callback expired",
+  "resetCallback": "Push callback indicated it should be reset",
+  "unknown": "Unknown push error",
+};
+
+const ERR_NO_PUSH_CALLBACK = "noCallback";
+const ERR_DATA_BUT_NO_KEYS = "noKeys";
+const ERR_PUSH_CALLBACK_EXPIRED = "expiredCallback";
+const ERR_PUSH_CALLBACK_RESET = "resetCallback";
+const ERR_PUSH_UNKNOWN = "unknown";
+
+const LOG_OP_PUSH_SEND_ATTEMPT = "push.send.attempt";
+const LOG_OP_PUSH_SEND_FAILURE = "push.send.failure";
+const LOG_OP_PUSH_SEND_SUCCESS = "push.send.success";
+const LOG_OP_PUSH_UNEXPECTED_ERROR = "push.sendPush.unexpectedError";
+const LOG_OP_TOO_MANY_DEVICES = "push.sendPush.tooManyDevices";
+const LOG_OP_DEVICE_UPDATE_FAILED = "push.sendPush.deviceUpdateFailed";
+
 const TTL_DEVICE_DISCONNECTED = 5 * 3600; // 5 hours
 const TTL_PASSWORD_CHANGED = 6 * 3600; // 6 hours
 const TTL_PASSWORD_RESET = TTL_PASSWORD_CHANGED;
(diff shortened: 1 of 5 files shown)


rfk:
Pushing for early visibility - the push module tests do not lend themselves well to the refactor that I've done here, and they seem overdue for a cleanup and asyncification anyway, so I'm going to try to push through and clean them all up in the style of the ones I've already done in this WIP PR.

packages/fxa-auth-server/test/local/push.js
```diff
@@ -70,110 +78,69 @@ describe('push', () => {
...
+    assert.equal(mockLog.info.getCall(1).args[1].deviceId, mockDevices[0].id)
+    assert.equal(mockLog.info.getCall(2).args[0], "push.send.success")
+    assert.equal(mockLog.info.getCall(2).args[1].reason, "accountVerify")
+    assert.equal(mockLog.info.getCall(2).args[1].deviceId, mockDevices[0].id)
+    assert.equal(mockLog.info.getCall(3).args[0], "push.send.attempt")
+    assert.equal(mockLog.info.getCall(3).args[1].reason, "accountVerify")
+    assert.equal(mockLog.info.getCall(3).args[1].deviceId, mockDevices[1].id)
+    assert.equal(mockLog.info.getCall(4).args[0], "push.send.success")
+    assert.equal(mockLog.info.getCall(4).args[1].reason, "accountVerify")
+    assert.equal(mockLog.info.getCall(4).args[1].deviceId, mockDevices[1].id)
```
rfk, line 112: @mozilla/fxa-devs before I get too far down the path of asyncifying and cleaning up these tests, do y'all have a current preferred style for doing this sort of "assert that calls were made as expected" stuff using sinon? I find the above pretty wordy, but clearer and more robust than what it was doing before.

chenba: You could try using sinon.assert.  For example, `sinon.assert.calledWithExactly(mockLock, ...)`.

packages/fxa-auth-server/lib/push.js
```diff
@@ -36,42 +72,13 @@ const TTL_COMMAND_RECEIVED = TTL_PASSWORD_CHANGED;
...
-    'profileUpdated',
-    'devicesNotify',
-    'accountDestroyed',
-    'commandReceived',
-  ];
-  const events = {};
-  for (const reason of reasons) {
-    const id = reason.replace(/[A-Z]/, (c) => `_${c.toLowerCase()}`); // snake-cased.
-    events[reason] = {
-      send: `push.${id}.send`,
```
rfk, line 56: These "push.${reason}.${outcome}` names remind me of datadog, where we had to cram things into the event name because of restrictions on tags. I think all of our current tooling can handle a simpler operation name with the reason etc as tags, and they were mildly annoying to work with in bigquery. So, I've changed them. I don't think we had anything depending on the precise details of naming here.

packages/fxa-auth-server/lib/push.js
```diff
@@ -82,34 +89,6 @@ module.exports = function (log, db, config) {
...
-   *
-   * @param {String} name
-   */
-  function incrementPushAction(name) {
-    if (name) {
-      log.info(LOG_OP_PUSH_TO_DEVICES, {
-        name: name,
-      });
-    }
-  }
```
rfk, line 111: I've refactored these a little, and also moved them down to the bottom of the file so they're closer to their callers.

packages/fxa-auth-server/lib/push.js
```diff
@@ -371,93 +350,131 @@ module.exports = function (log, db, config) {
...
-      const events = pushReasonsToEvents[reason];
-      if (!events) {
+      if (!PUSH_REASONS.has(reason)) {
         throw `Unknown push reason: ${reason}`;
       }
       // There's no spec-compliant way to error out as a result of having
       // too many devices to notify.  For now, just log metrics about it.
       if (devices.length > MAX_ACTIVE_DEVICES) {
-        reportPushError(new Error(ERR_TOO_MANY_DEVICES), uid, null);
+        log.warn(LOG_OP_TOO_MANY_DEVICES, { uid });
```
rfk, line 359: Here and elsewhere, I've moved from logging these at `error` level and dropped it to `warn` instead. They don't represent anything that we've done wrong so ISTM they shouldn't be logged as errors. But, I'm very open to being convinced otherwise.

packages/fxa-auth-server/lib/push.js
```diff
@@ -371,93 +350,131 @@ module.exports = function (log, db, config) {
...
+        const metricsTags = {
+          reason,
+          uid,
+          deviceId,
+          uaOS: device.uaOS,
+          uaOSVersion: device.uaOSVersion,
+          uaBrowser: device.uaBrowser,
+          uaBrowserVersion: device.uaBrowserVersion,
+        };
+        this.reportPushAttempt(device.pushCallback, metricsTags);
```
rfk, line 374: This now reports an `attempt` for every device, even ones that we know are definitely going to fail due to e.g. a missing or expired callback. I think that will make analysing these events easier overall, as they form a more deliberate funnel.

packages/fxa-auth-server/lib/push.js
```diff
@@ -371,93 +350,131 @@ module.exports = function (log, db, config) {
...
+            // set the push endpoint expired flag
+            // Warning: this method is called without any session tokens or auth validation.
+            device.pushEndpointExpired = true;
+            try {
+              await db.updateDevice(uid, device);
+            } catch (err) {
+              log.warn(LOG_OP_DEVICE_UPDATE_FAILED, { uid, deviceId, err });
             }
+          } else {
+            log.error(LOG_OP_PUSH_UNEXPECTED_ERROR, { err} );
```
rfk, line 436: This one is unexpected, so it gets logged at error level.

packages/fxa-auth-server/test/local/push.js
```diff
@@ -2,1025 +2,802 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-'use strict';
+"use strict";
```
rfk, line 5: There is a lot of churn in this file, so much that github is hiding it by default. Summary:

* I added two new tests for the success/failure log info and statsd calls.
* I asyncified the existing tests, which were using explicit promise chains.
* I migrated the exists tests to use higher-level `sinon.assert` functions for (IMHO) improved clarity.
* I updated the assertions in existing tests where appropriate, to match the logging changes.

PR #5472, opened 2020-05-27
fix(oauth): quiet logging on >24 hour tokens

jrgm:
This log message isn't really useful these days, so lower it to trace logging.


packages/fxa-auth-server/lib/oauth/token.js
```diff
@@ -83,7 +83,7 @@ exports.verify = async function verify(accessToken) {
     // Log a warning if reliers are using access tokens that are more
     // than 24 hours old.  Eventually we will shorten the expiry time
     // on access tokens and such old tokens won't be allowed.
-    logger.warn('token.verify.expiring_soon', {
+    logger.trace('token.verify.expiring_soon', {
```
dannycoates, line 86: let's go with `verbose`. `trace` generates a stacktrace

PR #5473, opened 2020-05-27
docs(adrs): Added workspaces ADR

dannycoates:
Finally wrote some words

### DIFF
--- a/docs/adr/0019-use-workspace-dependencies.md
+++ b/docs/adr/0019-use-workspace-dependencies.md
@@ -1,0 +1,51 @@
+# Use workspace dependencies in package.json
+
+- Deciders: Danny Coates
+- Date: 2020-05-27
+
+## Context and Problem Statement
+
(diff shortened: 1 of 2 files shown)


docs/adr/0019-use-workspace-dependencies.md
```diff
@@ -0,0 +1,47 @@
+# Use workspace dependencies in package.json
+
+- Deciders: Danny Coates
+- Date: 2020-05-27
+
+## Context and Problem Statement
+
+Since becoming a monorepo we haven't found a way to share code across packages that works well in local development, CI, and production without fragile, ad-hoc workarounds. We've used a mix of the following strategies so far:
+
+#### Using "long" relative paths
```
lmorchard, line 10: FWIW, I'd throw these at the end of the doc. IMO, ADRs are best when a quick summary of the decision comes at the top and further context comes later. Inverted pyramid structure and all that.

PR #5481, opened 2020-05-27
feat(aet): add ecosystem_anon_id field to auth server core profile data

6a68:
The ecosystem_anon_id work for AET (Jira epic FXA-880) started from the
API used by desktop code, in order to unblock that work, which can be
done in parallel. This new identifier is eventually going to be part
of the core profile fields stored in the auth server DB, which means
there will be changes in mysql, auth-db, auth-server, and profile-server.

The initial no-op profile-server API implementation landed as
https://github.com/mozilla/fxa/pull/5364.

This change moves one step further down the stack, updating the
profile-server API handler from a no-op to an actual call into the
_core_profile, and adding a deceptively simple change to the auth server
that allows an ecosystemAnonId field to be passed through from auth-db.

Future commits will integrate this new field into the auth-db and
eventually into mysql.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -905,6 +905,7 @@ module.exports = (
         },
         response: {
           schema: {
+            ecosystemAnonId: isA.string().optional(),
             email: isA.string().optional(),
             locale: isA.string().optional().allow(null),
             authenticationMethods: isA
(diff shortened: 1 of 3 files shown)


6a68:
> LGTM. I'm curious if you think we'll end up trying to write the `ecosystemAnonId` value through a `POST` variant of this same API, or whether there's another possible approach.

Great question. There is a [no-op `POST` endpoint](https://github.com/mozilla/fxa/blob/master/packages/fxa-profile-server/lib/routes/ecosystem_anon_id/post.js) that landed last week, and which could be plumbed through to a no-op in the auth server; I missed that in this commit. I'll add that change in a followup commit and open a new PR.

PR #5482, opened 2020-05-28
fix(payments): hack to get 1.172.2 to build

jrgm:
Let's see if this passes on circleci ...


jrgm:
Obviously not a real fix, but payments kept failing with 
`Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'`

The internets said to use this version of preset-env, so forces it here. For 1.172.2, we only want an image for content server; this unblocks that build.

r? - @vbudhram

Issue #5396, opened 2020-05-21
Mutations for password/email


data-sync-user:
Implement the mutations needed for the the password and email changes in the profile section.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1947)
┆Issue Number: FXA-1947

bbangert:
Closed by #5487

PR #5487, opened 2020-05-28
feat(graphql-api): add profile/email mutations

### LINKED ISSUE #5397: Mutations for profile manipulation
Implement the mutations needed for the profile of the new settings redesign:

* avatar modification
* display name change

This will be somewhat more complex perhaps because of the file upload for the avatar modification.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1946)
┆Issue Number: FXA-1946

bbangert:
Because:

* We want to alter/update the profile avatar/displayname and modify
  email addresses in use.

This commit:

* Add's mutations to upload an avatar, change the display name, and
  manipulate email addresses in use.

Closes #5397, #5396


packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -39,25 +58,42 @@ export async function createServer(
   const schema = await TypeGraphQL.buildSchema({
     container: Container,
     resolvers: [AccountResolver],
+    validate: false,
```
bbangert, line 61: This is only needed when we're using the `class-validator` package to validate input beyond the basic GraphQL type declarations.

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -39,25 +58,42 @@ export async function createServer(
...
   });
+  const authHeader = config.authHeader.toLowerCase();
   const authUser = Container.get(SessionTokenAuth);
   const debugMode = config.env !== 'production';
-  const defaultContext = async ({ req }: any) => {
-    const bearerToken = req.headers[config.authHeader.toLowerCase()];
+  const defaultContext = async ({ req }: { req: Request }) => {
```
bbangert, line 66: Adding the correct type to the `req` object highlighted that a check needs to be done before assuming its a string.

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -39,25 +58,42 @@ export async function createServer(
...
   return new ApolloServer({
     context: context ?? defaultContext,
     dataSources: () => ({
       authAPI: new AuthServerSource(),
+      profileAPI: new ProfileServerSource(),
     }),
-    formatError: err => reportGraphQLError(debugMode, logger, err),
+    formatError: (err) => formatError(debugMode, logger, err),
     schema,
+    uploads: false,
```
bbangert, line 90: Disables apollo's upload property which isn't needed when using the `graphql-upload` middleware package with express.

packages/fxa-graphql-api/src/lib/datasources/profileServer.ts
```diff
@@ -0,0 +1,73 @@
...
+import { configContainerToken, fxAccountClientToken } from '../constants';
+import { Context } from '../server';
+
+export class ProfileServerSource extends DataSource {
+  private authClient: Client;
+  private oauthToken: String | undefined;
+  private oauthClientId: string;
+  private profileServerUrl: string;
+
+  constructor(private token: string = '') {
```
lmorchard, line 18: Maybe more a note to myself, since I was confused that this `token` parameter didn't seem used anywhere: TIL [TypeScript supports parameter properties](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties), which take care of declaring & initializing an object property all at once.

packages/fxa-graphql-api/src/lib/resolvers/account-resolver.ts
```diff
@@ -1,31 +1,142 @@
...
+  @Mutation((returns) => UpdateAvatarPayload, {
+    description: 'Update the avatar in use.',
+  })
+  @CatchGatewayError
+  public async updateAvatar(
+    @Ctx() context: Context,
+    @Arg('input', (type) => UpdateAvatarInput) input: UpdateAvatarInput
+  ): Promise<UpdateAvatarPayload> {
+    const file = await input.file;
+    const fileData = await getStream.buffer(file.createReadStream());
```
lmorchard, line 66: Looks like this loads the incoming image into memory before sending via superagent in `avatarUpload`. 

[Could we pipe the `input.file` stream to the profile-server POST request instead?](https://visionmedia.github.io/superagent/#piping-data) Maybe not worth thinking too hard about since we're hopefully talking about a couple megabytes here.

lmorchard:
Oh yeah, I should assign myself since I've been looking at this for the last 30 min or so and am about to post all my review comments 😅

packages/fxa-graphql-api/src/test/lib/datasources/authServer.spec.ts
```diff
@@ -1,21 +1,22 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
```
lmorchard, line 1: Only other vague grumble I might have, is that I wish these `*.spec.ts` modules lived in the same directories as the `*.ts` modules they're testing. But, that's fully preference and a late bit of feedback since we're down the road a bit with this file structure

packages/fxa-graphql-api/src/test/lib/datasources/profileServer.spec.ts
```diff
@@ -0,0 +1,100 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import 'mocha';
+import 'reflect-metadata';
```
lmorchard, line 5: Oh and one other thing I'm scratching my head at: What does reflect-metadata do for us? Seems like a polyfill, saw google results that suggested tests on decorator-using code needed it. But I was curious and commented it out to see what happens, and the tests still pass.

packages/fxa-graphql-api/src/lib/resolvers/types/payload/update-display-name.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Field, ObjectType } from 'type-graphql';
+
+@ObjectType()
+export class UpdateDisplayNamePayload {
+  @Field({
+    description: 'A unique identifier for the client performing the mutation.',
+    nullable: true,
+  })
+  public clientMutationId?: string;
```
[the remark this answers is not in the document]
bbangert: I looked at that, and while GraphQL does support interfaces, they're only for data types like a query would return. Not for mutation inputs or payloads. There was some complex reason I found for this, as I actually was trying to set this up like that originally, before I ran into problems because that's not allowed.

packages/fxa-graphql-api/src/lib/resolvers/account-resolver.ts
bbangert: Yes, we could pipe it into the request, the actual restriction in profile server on uploads is 1MB. nginx will typically buffer a large portion of that both in send and receive buffers which are in front of our apps, so I wasn't too worried about it here. It's worth remembering if we see any funky behavior though.

packages/fxa-graphql-api/src/test/lib/datasources/profileServer.spec.ts
bbangert: The Dependency Injection libraries like inversify and typedi (used here) require reflect-metadata to be present, before any of their code loads. Technically it only needs to happen once, so a root test runner could do it and none of these files would then need it. However, then we wouldn't be able to run the tests just for a single test file, like with the VS Code 'Mocha (Run current file)' launch task. Thus, I end up adding these to each spec file, cause I really like running just the file I'm working on.

packages/fxa-graphql-api/src/test/lib/datasources/authServer.spec.ts
bbangert: Ah, I'd be good with that, having to navigate to the test mirror directory with all the spec files is annoying, and harder to see if you've missed tests for something than having them right there like you mention.

packages/fxa-graphql-api/src/test/lib/datasources/profileServer.spec.ts
lmorchard: Ah yeah, that kind of thing belatedly occurred to me this morning... If we need more of this sort of thing, we could add a global [`test-setup.ts`](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/test-setup.ts) that runs for all tests (even single runs)

PR #5493, opened 2020-05-28
fix(docker): Fix docker deploy build errors and remove hack

vbudhram:


_dev/docker/builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12-slim
+FROM node:12.16.3-slim
```
jrgm, line 1: Hrm. Was this change intentional for some reason?

dannycoates: though moot at this point, since we aren't releasing this, the build error in payments server was triggered by node 12.17

PR #5495, opened 2020-05-28
fix(tests): teamcity run.sh npx yarn workspaces focus fxa-content-server fxa-js-client

jrgm:
This seems to get this going.


jrgm:
Damn. Fast.

PR #5507, opened 2020-05-29
Storybook + Tailwind improvements in CRA projects

### LINKED ISSUE #5490: [setup] Allow access to Tailwind classes in fxa-settings & fxa-react Storybooks, admin panel, & possibly Jest
Currently, `fxa-settings` and `fxa-react` Storybook builds don't have access to the styles outputted by Tailwind. We need to expose these classes to Storybook so the components render with the proper styles.

We also need to import Tailwind classes in `fxa-admin-panel` since some components have been moved into `fxa-react` and are using Tailwind classes there. Most of the styles under [this TODO comment](https://github.com/mozilla/fxa/blob/master/packages/fxa-admin-panel/src/components/AppLayout/index.scss#L34) should be deleted after this is configured.
^Note: perhaps there is a way to 'extend' the Tailwind config file from settings, but to look for classes in `fxa-react` and `fxa-admin-panel` instead of `fxa-settings` for purging?

Additionally, it could be ideal to look into exposing our CSS to our testing environments. [This todo](https://github.com/mozilla/fxa/blob/master/packages/fxa-settings/src/components/HeaderLockup/index.test.tsx#L12) in particular pointed out that we can't use Jest's [toBeVisible](https://github.com/testing-library/jest-dom#tobevisible) method which would be very handy for testing things that are hidden or shown at various breakpoints. (Yes, we can test for the existence of a class on an element, but it feels much more fail-proof to test for the functionality of that class and would move some E2E test burden onto unit tests)

[This todo](https://github.com/mozilla/fxa/blob/master/packages/fxa-settings/tailwind.config.js#L6) is related but doesn't necessarily need to be done here.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2000)
┆Issue Number: FXA-2000

jodyheavener:
Closes #5490

Because:

- We need Storybook instances to be able to render components that make use of Tailwind classes 
- Currently we are production-building Tailwind classes to include _all_ shared component styles
- We have a variety of shared configurations

This PR:

- Sets up Storybook and Tailwind for fxa-admin-panel
- Shares Tailwind config across all projects, moves fxa-react shared configs to `./_dev` to match other shared configs
- Compile tailwind for use in fxa-react, fxa-settings, and fxa-admin-panel Storybooks
- Adds a script to Tailwind configs to only run PostCSS (specifically PurgeCSS) class identification to fxa-react components actually being used in a given project, instead of all of them

I am eager to hear your thoughts on [this extraction script](https://github.com/mozilla/fxa/pull/5507/files#diff-c7caca7975efe69940e2118b5a415004). Can it be improved? Is it too bulky / slow?


_dev/rescripts.js
```diff
@@ -5,8 +5,8 @@
 const { resolve } = require('path');
 
 const additionalJSImports = {
-  'fxa-react': __dirname,
-  'fxa-shared': resolve(__dirname, '../fxa-shared'),
+  'fxa-react': resolve(__dirname, '../packages/fxa-react'),
```
dannycoates, line 8: im _pretty sure_ these could be like `require.resolve('fxa-react')`

_dev/storybooks.js
```diff
@@ -40,7 +49,7 @@ const customizeWebpackConfig = ({ config, mode }) => {
     assetLoader = {
       loader: resolve(
         __dirname,
-        '../../node_modules/@storybook/core/node_modules/file-loader/dist/cjs.js'
+        '../node_modules/@storybook/core/node_modules/file-loader/dist/cjs.js'
```
dannycoates, line 52: We should avoid relying on exact paths under `node_modules` since hoisting plus version changes can move them unexpectedly. Since this depends on `file-loader` we should add it as a dependency and use `require.resolve` here.

idk what I think about putting this in `_dev`. I think I understand why you did, but im leaning toward fxa-react being the better place since storybook and react go hand in hand.

lmorchard: Yeah, same with `_dev/rescripts.js` - that's very specifically a Create React App thing. I'd keep that stuff in fxa-react

jodyheavener: Truuue, but it could be argued that Tailwind and Storybook are not necessarily exclusive to React (though they are in our case, currently). I put them here for the same reason the TypeScript configs live here, as common build configs (none of them are bundled). But I do take your point, they'll live in `fxa-react/configs` for now 😄

_dev/rescripts.js
jodyheavener: Looks like this resolves to `path/fxa-shared/dist/index.js`

packages/fxa-react/extract-imported-components.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function getExtractableFiles(dir) {
+  const subdirs = readdirSync(dir);
+  return subdirs
+    .map((subdir) => {
+      const res = resolve(dir, subdir);
+      return statSync(res).isDirectory() ? getExtractableFiles(res) : res;
+    })
+    .reduce((arr, files) => arr.concat(files), [])
```
dannycoates, line 20: for some code golf try [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

packages/fxa-settings/package.json
```diff
@@ -69,6 +70,7 @@
     "eslint-plugin-react": "^7.19.0",
     "fxa-shared": "workspace:*",
     "node-sass": "^4.14.0",
+    "nodemon": "^2.0.4",
```
dannycoates, line 73: where are we using this?

packages/fxa-admin-panel/package.json
```diff
@@ -73,11 +83,13 @@
     "@types/react-router-dom": "^5.1.3",
     "@types/serve-static": "^1.13.3",
     "@types/supertest": "^2.0.8",
+    "@types/webpack": "4.41.16",
     "@typescript-eslint/eslint-plugin": "^2.21.0",
     "@typescript-eslint/parser": "^2.21.0",
     "apollo-cache-inmemory": "^1.6.2",
     "apollo-link": "^1.2.12",
     "apollo-utilities": "^1.3.2",
+    "babel-loader": "^8.1.0",
```
dannycoates, line 92: where's this used?

packages/fxa-admin-panel/package.json
```diff
@@ -86,10 +98,13 @@
...
     "pm2": "^4.4.0",
+    "postcss-cli": "^7.1.1",
     "prettier": "^2.0.5",
     "supertest": "^4.0.2",
+    "tailwindcss": "^1.4.6",
     "ts-jest": "^25.2.1",
     "ts-node": "^8.10.1",
-    "typescript": "3.8.3"
+    "typescript": "3.8.3",
+    "webpack": "^4.43.0"
```
dannycoates, line 108: I assume this is a peerDep of babel-loader

packages/fxa-settings/package.json
jodyheavener: whoops, this was from an experiment

packages/fxa-admin-panel/package.json
jodyheavener: Storybook makes use of this and webpack (as well as @babel/core), yarn complains when they're not present. These are present in fxa-settings, fxa-payments-server as well.

packages/fxa-react/components/Header/index.tsx
```diff
@@ -7,11 +7,15 @@ import React, { ReactElement } from 'react';
...
   right: ReactElement;
+  className?: string;
 };
 
 export const Header = (props: HeaderProps) => {
   return (
-    <header role="banner" className="header-page sticky top-0 bg-grey-10">
+    <header
+      role="banner"
+      className={props.className || 'sticky top-0 bg-grey-10'}
```
LZoog, line 17: Noice👌 You could also destructure `className` to avoid `props.className` but this works.

packages/fxa-settings/tailwind.config.js
```diff
@@ -1,183 +1,19 @@
...
-    },
-  },
-  variants: {},
-  plugins: [],
-};
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { resolve } = require('path');
```
LZoog, line 5: It might be good to leave a comment at the top of this file and any other Tailwind config files that all of our Tailwind configurations across packages point to `fxa-react/configs/tailwind` - or maybe that's obvious by L7?

LZoog, line 5: It might be good to leave a comment at the top of this file and any other Tailwind config files that all of our Tailwind configurations across packages point to `fxa-react/configs/tailwind` - or maybe that's obvious by L7?

packages/fxa-settings/package.json
LZoog: Was this to run `postcss` when the tailwind config or tailwind.css changes? I feel like you could sneak that in here too, or it could be done in a follow up (it's been pretty annoying for me to manually run, didn't set that up in my branch in case you were doing that her)

packages/fxa-react/jest.config.js
```diff
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
LZoog, line 3: It feels a little funny having some config files root level and others inside a `config/` dir. Should they all just be top-level? 🤔

packages/fxa-settings/package.json
LZoog: Actually, that task would need to watch and recompile at any scss changes as well to a accommodate what I've got in my open PR with those custom files/classes. We can make a follow up issue at an est of a 1 or 2 if that can't make it in here for the sprint.

packages/fxa-react/jest.config.js
jodyheavener: This config and all others at the root are for fxa-react exclusively. The configs inside `/configs` are meant to be shared.

packages/fxa-settings/package.json
jodyheavener: Yeah, so I've got these set up in the package.json:

```
    "poststart": "watch-postcss",
    "postrestart": "watch-postcss",
    "watch-postcss": "nodemon --watch ./src/tailwind.css --exec npm run build-postcss"
```

Though it's not currently working. I also suspect this would have issues as it doesn't know when the PM2 instance gets killed. And as you mentioned, it doesn't work with _any_ SCSS files, only the main Tailwind file. How this would work with SCSS files (especially external component SCSS files), without also conflicting with Hot Module Reloading that watches these files already, I'm not sure,

jodyheavener: I'll make a followup issue for this.

jodyheavener:
Strange. I can't get this to fail locally with regular test command. If I run the circleci `test-many` job on my machine I _can_ get it to fail with the same type of error, but if I remove `fxa-admin-panel` from test-many, which runs before it, it seems to not fail. @dannycoates @LZoog any ideas?

jodyheavener:
I actually just tried to run this (test-many via circleci) locally on master and had the same result - fails if fxa-admin-panel is built before fxa-react's tests run fxa-react will fail. @dannycoates is this possibly what #5531 is addressing?

dannycoates:
> is this possibly what #5531 is addressing?

@jodyheavener, yes, the "clean" task in that PR is the fix

packages/fxa-react/extract-imported-components.js
```diff
@@ -0,0 +1,42 @@
...
+  });
+}
+
+function extractImportedComponents(dirToScan, prefix, suffix) {
+  console.log('Extracting externally imported component paths...');
+  const filePaths = getExtractableFiles(dirToScan);
+  return parseForImports(filePaths, prefix, suffix);
+}
+
+module.exports = extractImportedComponents;
```
LZoog, line 42: This helper is so cool 😀

packages/fxa-settings/tailwind.config.js
```diff
@@ -1,198 +1,21 @@
...
-    },
-  },
-  variants: {},
-  plugins: [],
-};
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Please refer to fxa-react/configs/tailwind for the main config file
```
LZoog, line 5: So polite.

PR #5513, opened 2020-06-01
style(settings): UnitRow, UnitRowWithAvatar, & Modal styles + Storybook, allow postcss imports

### LINKED ISSUE #5394: [global] Add Tailwind styles to modal & unit rows + show in Storybook
With the Tailwind config merged, we can begin adding styles to Settings. Let's start with the components already in use and get them looking good in Storybook.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1942)
┆Issue Number: FXA-1942

LZoog:
Because:
* Now that Tailwind is configured, we can style existing Settings components and the current layout.

This commit:
* Changes UnitRowWithImage to UnitRowWithAvatar because we likely will not reuse it
* Styles Modal, UnitRow, and UnitRowWithAvatar across mobile, tablet, & desktop
* Adds Modal, UnitRow, and UnitRowWithAvatar to Storybook
* Adds postcss-import and sets a convention for shared class styles with Tailwind with ctas.scss, UnitRow.scss, and .link-blue
* Modifies the shared close.svg

fixes #5394


_dev/.stylelintrc
```diff
@@ -20,9 +20,10 @@
           "if",
           "each",
           "include",
-          "mixin"
+          "mixin",
+          "screen"
         ]
       }
-    ]
+    ],
```
LZoog, line 27: I'll remove this unnecessary comma, snuck its way in.

packages/fxa-settings/src/styles/custom/UnitRow.scss
```diff
@@ -0,0 +1,59 @@
...
+.unit-row {
+  @apply flex flex-wrap px-4 pt-7 pb-8;
+
+  &-header {
+    @apply w-full mb-1;
+  }
+
+  &-header-value {
+    @apply w-full;
+  }
```
LZoog, line 14: Using `&-` here to one layer felt right to help the styles here be a bit more concise, two seemed to make it less readable. We could choose not to ever use `&-` because this does make it less searchable.

packages/fxa-settings/src/styles/custom/UnitRow.scss
```diff
@@ -0,0 +1,59 @@
...
+  &-cta-container {
+    @apply w-full;
+  }
+
+  &-hr {
+    @apply mx-5 border-grey-100;
+  }
+}
+
+.cta {
```
LZoog, line 25: For consistency this could be called `unit-row-cta` for now, I don't really want to call it a `button` because sometimes this is a link - the design guide calls it `button-neutral`.  We'll also be using this in the `Connected Services` section and when we do that, `cta` could be in its own file and makes sense. We could call the "Disable" buttons / "Delete account" link as something like `cta-warning`.

packages/fxa-settings/src/styles/custom/UnitRow.scss
jodyheavener: I'm torn about these re-used classes in general. I can see why they were created, and how they are handy, but at a certain point it starts to look like we're writing custom CSS (if only to use `@apply`), and an added drawback IMO is needing a `custom/` dir. Could these classes simply exist in component classNames? There will be duplication of class names (as is the spirit of Tailwind), but it's more flexible in the long run and alleviates the prior concerns.

What do you think?

packages/fxa-settings/src/styles/custom/animations.scss
```diff
@@ -0,0 +1,29 @@
...
+  animation-duration: 800ms;
+}
+
+.animate-repeat-infinite {
+  animation-iteration-count: infinite;
+}
+
+.animate-rotate {
+  animation-name: rotate;
+}
```
jodyheavener, line 29: In theory these are meant to be globally available, and not custom/specific to a given component -- should they exist in `custom/`? What is `styles/` if not a directory of custom CSS files? 😉

packages/fxa-settings/tailwind.config.js
```diff
@@ -12,6 +12,17 @@ module.exports = {
...
+      padding: {
+        7: '1.75rem',
+      },
+      margin: {
+        11: '2.75rem',
+      },
+      borderRadius: {
+        xl: '.75rem',
+      },
+    },
```
jodyheavener, line 25: Just checking -- all of these are in addition to and not overriding existing classes, right?

packages/fxa-settings/tailwind.config.js
```diff
@@ -177,6 +188,11 @@ module.exports = {
         900: '#321C64',
       },
     },
+    flex: {
+      1: '1',
+      2: '2',
+      4: '4',
+    },
```
jodyheavener, line 195: Should this live inside the `extend` object?

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -38,69 +38,85 @@ export const Settings = () => {
...
+              revealModal,
+              modalRevealed,
+            }}
+          >
+            <p className="text-sm mt-3">
+              Access your account if you can't log in to your primary email.
+            </p>
+            <p className="text-grey-400 text-xs mt-2">
+              Note: a secondary email won't restore your information—you'll need
+              a <a href="#recovery-key">recovery key</a> for that.
```
jodyheavener, line 98: Is a11y linter going to yell about this href?

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -38,69 +38,85 @@ export const Settings = () => {
...
+
+          <hr className="unit-row-hr" />
+
+          <UnitRow
+            header="Password"
+            headerValueClassName="tracking-wider"
+            headerValue="••••••••••••••••••"
+            route="#"
+          >
+            <p className="text-grey-400 text-xs mobileLandscape:mt-3">
```
jodyheavener, line 69: I feel like I broke a convention making the breakpoints camelCase. Should we change them to dashes?

packages/fxa-settings/src/styles/custom/UnitRow.scss
LZoog: As I've continued working on this and looking at the wires, I see the need for a `cta-default` class (what `.cta` is now), `cta-default-lg` for modals, `cta-primary`, and `cta-warning` or whatever we want to call the red one. If we _don't_ do it this way, there's going to be _a lot_ of places where we're going to need to add/remove classes when something needs to be updated. I feel like there's greater risk of missing something when we have a repeated set of classes. I did this originally with `UnitRow` because if we needed to change the padding on something there, well, it would be _very_ easy to update it in `UnitRow` and forget to update it in `UnitRowWithImage`.

Check out [this video from the docs](https://tailwindcss.com/course/composing-utilities-with-apply/#app). They actually do exactly this so I still think it's very Tailwindy. 😎

packages/fxa-settings/src/styles/custom/animations.scss
LZoog: I could take them out of this dir if you think that's preferable, we might end up with a decent amount of files under `styles/` though. 🤔

packages/fxa-settings/tailwind.config.js
LZoog: Correct, anything under `extend` extends the theme 👍

packages/fxa-settings/src/styles/custom/animations.scss
jodyheavener: Is that bad?

packages/fxa-settings/tailwind.config.js
LZoog: There's only one `flex-#` (`flex-1`) class and I didn't realize that class was even there, so I'll move this under `extend` with only 2/4 needed 👍

packages/fxa-settings/src/components/Settings/index.tsx
LZoog: No, it only complains about `#`. `#anything` is fine.

packages/fxa-settings/src/components/Settings/index.tsx
LZoog: Hmm well the only one we have is `mobileLandscape`, I think I actually prefer it that way because `mobile-landscape` might look less obvious as its own entity than `mobileLandscape` but if you strongly prefer `mobile-landscape` then I think that's fine (maybe we'd change it after this PR is merged) 👍

packages/fxa-settings/src/styles/custom/UnitRow.scss
jodyheavener: Right, I do understand why you've set it up this way, and I agree it is handy. I guess my thinking goes back to concerns that when we start applying global styles to re-used components we could be painting ourselves into a corner if an instance of the component needs to be overridden in such a way that competes with existing styles. Is this approach not the same as applying base or concatenated Tailwind classes to shared components instead of passing all of them in as props for each instance?

packages/fxa-settings/src/components/Settings/index.tsx
jodyheavener: desktopXl as well

packages/fxa-settings/src/styles/custom/animations.scss
LZoog: I think that's OK until we have a lot of files. Maybe we won't get there, I'll pull them out of `custom/`.

packages/fxa-settings/src/styles/tailwind.css
```diff
@@ -1,34 +1,22 @@
...
-  @apply bg-grey-10;
+  @apply text-base font-body bg-grey-10;
 }
 
-@keyframes rotate {
-  0% {
-    transform: rotate(0);
-  }
+/* a {
+  @apply underline text-blue-500;
```
LZoog, line 17: Ack, I'll fix this in the morning

packages/fxa-settings/src/styles/custom/UnitRow.scss
LZoog: If it competes with existing styles, only the styles the component needs should be set on that component.

> Is this approach not the same as applying base or concatenated Tailwind classes to shared components instead of passing all of them in as props for each instance?

Well, check out `cta-default`, I'm referencing that class in 3 places in this PR (and made a `cta-default-lg` since the modal one had similarities but not the exact styles, and noticed we'd need the same style for the "separate page" flows). I don't think every single instance of what I've got for `cta-default` should be pulled into its own component, that would feel complicated just for the sake of not doing it this way, and I think that'd really complicate `UnitRow` and `UnitRowWithImage` if we want to keep those the same. If you want to poke around the branch I'm all ears on other solutions, this one just feels like the cleanest, and Tailwind docs do recommend it when you've got a long set of classes used in more than one spot.

LZoog: BTW, after we reach consensus here, not sure if I'll have time this sprint but I want to write up a quick "follow these conventions" note in the README; the one you wrote on SVGs and what we've already got in there is handy. :)

jodyheavener: > If it competes with existing styles, only the styles the component needs should be set on that component.

I think that's where potential problems lie, though? Let's say Foo component is styled with a `foo` class that applies Tailwind classes to set properties A through F. And this works for everything, until we need an instance of Foo with a different C value. Does that mean we stop using `foo` (is this possible if the class is baked into? Or apply the additional className and have classes that introduce competing values?

All things considered, though, I recognize this set of classes is for a really far-reaching set of elements, so I'm less concerned about this in general, but I agree that we should put up some examples and establish if this is considered standard practice, or if it's an exception. Thank you for the debating here :)

LZoog: Yeah, you definitely have a point here - depending on the situation, maybe that one differing style could be pulled out and applied, so right now we have:

```
.unit-row {
  @apply flex flex-wrap px-4 pt-7 pb-8;
}
```
Say we need to use this class, but we need a `pb-4` instead. Our options are:
1. Apply a `unit-row pb-4` combination class, overwriting `pb-8`
2. Applying `flex flex-wrap px-4 pt-7 pb-4` directly on the one-off
3. Remove `pb-8` from `unit-row`, finding all instances of where it's used, and apply `unit-row pb-8` there, and `unit-row pb-4` on the one-off instance

I think it could depend on the circumstance. I could potentially see cases where all 3 might be OK, but if you'd like to avoid 1 entirely, we can make note of that.

packages/fxa-settings/src/components/Settings/index.tsx
LZoog: Ah, well, I'm fine with either, it doesn't bother me. :)

packages/fxa-settings/src/styles/ctas.scss
```diff
@@ -0,0 +1,45 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.cta-neutral {
```
LZoog, line 5: @jodyheavener I had this as `cta-default` because I saw it named that in Figma but the actual page showing all the different button types calls it "neutral" and that seems more fitting.

packages/fxa-settings/src/styles/UnitRow.scss
```diff
@@ -0,0 +1,47 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.unit-row {
+  @apply flex flex-wrap px-4 pt-7 pb-8;
+
+  &-header {
+    @apply w-full mb-1;
+  }
+
+  &-header-value {
```
jodyheavener, line 12: I'm not sure this needs to indicate that it's the header's value, since it is _the_ value (i.e. there's no main non-header value in this section, is there?). It could even be reduced to `&-content` IMO.

packages/fxa-settings/src/styles/UnitRow.scss
```diff
@@ -0,0 +1,47 @@
...
+    @apply w-full;
+  }
+
+  &-cta-container {
```
jodyheavener, line 16: Because we'll be putting more than standard CTA buttons, such as icon buttons, and in theory we could put other content here down the road I wonder if this could be more generic, such as `&-actions`, or `&-extra` (less enthused about this).

packages/fxa-settings/src/styles/UnitRow.scss
```diff
@@ -0,0 +1,47 @@
...
+      @apply flex-1 flex justify-end;
+    }
+  }
+}
+
+@screen desktop {
+  .unit-row-header-value {
+    @apply flex-4;
+  }
+}
```
jodyheavener, line 47: I think I know why you named it this way, but how would you feel about calling this file `unit-row.scss`, so we're consistent? It could go in a `components` dir if you want to make the distinction?

LZoog: That's fine by me👍

packages/fxa-settings/src/styles/UnitRow.scss
LZoog: Hmm yeah, we can go with `&-content`. I wasn't sure what to call it and since folks are familiar with key/value pairs I went with the "value" of the "header".

packages/fxa-settings/src/components/UnitRow/index.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { storiesOf } from '@storybook/react';
+import { useBooleanState } from 'fxa-react/lib/hooks';
+import { UnitRow } from '.';
+import { Modal } from '../Modal';
+
+storiesOf('components/UnitRow', module)
+  .add('basic, with falsey headerValue', () => (
+    <UnitRow header="Some header" headerValue={null} />
+  ))
+  .add('with CTA and falsey headerValue', () => (
```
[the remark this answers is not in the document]
LZoog: Yeah actually, it does make sense to do this in `UnitRow`. `UnitRowWithAvatar` needs to have `null` explicitly passed in since `avatarUrl` could be `null`.

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -38,69 +38,89 @@ export const Settings = () => {
...
+              revealModal,
+              modalRevealed,
+            }}
+          >
+            <p className="text-sm mt-3">
+              Access your account if you can't log in to your primary email.
+            </p>
+            <p className="text-grey-400 text-xs mt-2">
+              Note: a secondary email won't restore your information—you'll need
+              a{' '}
```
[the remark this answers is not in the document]
LZoog: This was autoformatted by prettier🤷‍♀️

packages/fxa-settings/src/components/AppLayout/index.test.tsx
```diff
@@ -8,11 +8,12 @@ import '@testing-library/jest-dom/extend-expect';
 import AppLayout from '.';
 
 it('renders the children', () => {
-  const rendered = render(
+  const { getByTestId } = render(
```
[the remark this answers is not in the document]
LZoog: I do agree, but I feel that's kind of a separate PR and for "new tests from now on," this was just a quick change for me to get tests passing😅

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -216,7 +216,7 @@ export const Subscriptions = ({
               aria-label="Close modal"
               onClick={clearSuccessAlert}
             >
-              <CloseIcon className="close" />
+              <CloseIcon role="img" className="close" />
```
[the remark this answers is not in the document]
LZoog: Hmm yeah, I'll change `.close` to `.close-alert-bar` and remove the duplicate code (`width`/`height`) from it and add the `close-icon` class to `<CloseIcon />` in Payments.

packages/fxa-settings/src/styles/UnitRow.scss
LZoog: Good point, I'm fine with `&-actions` due do that "delete" button we'll add later. (I guess that _could_ be looked at as a call to action, but eh)

packages/fxa-settings/src/components/Settings/index.tsx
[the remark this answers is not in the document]
jodyheavener: 😠

packages/fxa-settings/src/components/UnitRow/index.stories.tsx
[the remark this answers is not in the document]
LZoog: @jodyheavener actually, we can't do this because `displayName` can also be `null`.

packages/fxa-settings/src/styles/UnitRow.scss
LZoog: Ah, actually if we're _just_ talking about the class name it makes sense here - I was going to replace the `headerValue` prop with `content` but that doesn't really work since `{children}` are part of the "content," but I'll make the class name update as it's on the container of `headerValue` and `{children}`.

PR #5516, opened 2020-06-01
fix(customs): Fix undefined errors when blocking a request from ip service

vbudhram:
Putting this up for feedback, I still need to target current train.

### DIFF
--- a/packages/fxa-customs-server/lib/dataflow.js
+++ b/packages/fxa-customs-server/lib/dataflow.js
@@ -15,6 +15,11 @@ const TYPES = new Map([
   ['sourceaddress', 'ip'],
 ]);
 
+const RECORD_TYPES = new Map([
+  ['email', 'emailRecord'],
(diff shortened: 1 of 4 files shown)


packages/fxa-customs-server/lib/dataflow.js
```diff
@@ -75,12 +81,16 @@ module.exports = (config, log, fetchRecords, setRecords) => {
           [type]: action.indicator,
         });
 
-        records[type][action.suggested_action]();
+        if (recordType) {
+          records[recordType][action.suggested_action]();
```
vbudhram, line 85: this was undefined before because it was doing `records['email'].block()`, it should be `records['emailRecord'].block()`

jrgm:
You'll also want a build-ci.sh noop for payments. (it's test and build fail on 172)

vbudhram:
Closing in favor of https://github.com/mozilla/fxa/pull/5516

PR #5517, opened 2020-06-01
fix(customs): Fix undefined errors when blocking a request from ip service

vbudhram:
When attempting to block a request that the IP reputation says the custom server should block, the server throws defined at https://github.com/mozilla/fxa/blob/6fe29d2bb7fa08900de2fad3efdfabbf95ae5555/packages/fxa-customs-server/lib/dataflow.js#L78. The code was checking record['email'] instead of record['emailRecord'], this fixes that.


packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/qr-code-cad.js
```diff
@@ -17,7 +17,7 @@ const GROUPS = [
   'treatment-b', // Screen displayed in non-sms markets
 ];
 
-const ROLLOUT_RATE = 1.0;
+const ROLLOUT_RATE = 0.0;
```
vbudhram, line 20: This was enabled in 1.172.2 but never deployed, lets keep it disabled.

packages/fxa-customs-server/lib/dataflow.js
```diff
@@ -72,12 +81,16 @@ module.exports = (config, log, fetchRecords, setRecords) => {
           [type]: action.indicator,
         });
 
-        records[type][action.suggested_action]();
-
-        await setRecords(records);
+        if (recordType) {
+          records[recordType][action.suggested_action]();
```
vbudhram, line 85: This was undefined before because it was doing records['email'].block(), it should be records['emailRecord'].block()

jrgm:
Add build-ci.sh for payments.

PR #5526, opened 2020-06-02
refactor: clean up the fluent file

clouserw:
Because:

* Groups include all group comments for all the strings which makes
these spammy

This commit:

* Splits groups out


clouserw:
fixes #5527

PR #5530, opened 2020-06-02
refactor(fxa-react): replace imported fxa-content-server scss variables, remove dependency

### LINKED ISSUE #5439: Replace Survey imported content-server variables values with literal values
Maybe with the new style guidelines fxa-react will naturally lose this reference to content-server?
https://github.com/mozilla/fxa/blob/master/packages/fxa-react/components/Survey/index.scss#L1

If anything we should invert the fxa-react/fxa-content-server dependency.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1973)
┆Issue Number: FXA-1973

jodyheavener:
Closes #5439 

In case ya wanna reference: https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.css

### DIFF
--- a/packages/fxa-react/components/Survey/index.scss
+++ b/packages/fxa-react/components/Survey/index.scss
@@ -1,8 +1,8 @@
-@import '../../../fxa-content-server/app/styles/variables';
-
 $shadow: 0 12px 18px 2px rgba(34, 0, 51, 0.04),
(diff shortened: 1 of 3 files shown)


jodyheavener:
> You could stick these as vars at the top

Done

PR #5544, opened 2020-06-03
feat(settings): Pull account query into HOC, add Avatar component

### LINKED ISSUE #4898: [main] Create shell `<Settings>` component with GQL query
Create the `<Settings>` component which can be a shell at this stage. Craft the GQL query needed for all of the data shown on the wire and setup the `useQuery` hook to use the query. You can likely reference the `fxa-admin-panel`, but keep in mind that uses `useLazyQuery`.

We'll probably also want a HOC (higher order component) to execute the query in (is this `AppLayout` or do we need a new one?). The initial query should be able to feed into the nav to conditionally render the subscriptions link and the avatar drop-down to display the avatar and user's email, but the actual functionality will be done in follow up tickets.

This may overlap with an issue from "Server GQL setup." When the GQL server is up, this query may have already been written.

Things to note:

* The asterisks for "Password" are hardcoded.
* On the currents settings page we don't currently display "Created on" in the password section but we should have this information (not location) available.
* We need the user's primary email in the primary header avatar drop-down. Let's make sure this is only called once. We may need to perform this query inside a higher order component to populate the data there.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1605)
┆Issue Number: FXA-1605

LZoog:
Because:
* The Header and Nav components inside AppLayout need access to portions of the initial account query.
* We need to display the avatar in multiple components.

This commit:
* Adds a HOC component of which AppLayout is a child of and moves the account query into this component and passes the data into AppLayout and Settings.
* Creates the Avatar component and displays it correctly in Header.

fixes #4898


packages/fxa-settings/src/components/AppLayout/index.tsx
```diff
@@ -7,25 +7,36 @@ import HeaderLockup from '../HeaderLockup';
...
+      </main>
+    </div>
 
-      {/*TO DO: pull `Footer` in from `fxa-admin-panel` into
-      `fxa-react` and replace this*/}
-      <footer></footer>
-    </>
-  );
-};
+    {/*TO DO: pull `Footer` in from `fxa-admin-panel` into
```
dannycoates, line 36: `TO DO`?! 😳

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -51,17 +72,9 @@ const rightHeaderContent = (
       className="w-6 mx-6 desktop:mx-8"
       data-testid="header-bento"
     />
-    <DefaultAvatar
-      aria-label="Default avatar"
-      role="img"
-      className="w-10"
-      data-testid="header-avatar-default"
-    />
+    {console.log(`TODO display ${primaryEmail} in the header in FXA-1575`)}
```
dannycoates, line 75: `TODO` 😊

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -23,19 +19,7 @@ export const Settings = () => {
...
-  }
-
-  if (error) {
-    return <AppErrorDialog data-testid="error-dialog" {...{ error }} />;
-  }
-
-  const account: accountData = data.account;
-  const primaryEmail = account.emails.find((email) => email.isPrimary)!;
-  // const secondaryEmails = account.emails.filter((email) => !email.isPrimary);
+  const primaryEmail = account.emails.find((email: any) => email.isPrimary)!;
```
jodyheavener, line 22: Because `account` is typed isn't the interface of `email` here implicitly `{ email: string; isPrimary: boolean; verified: boolean; }`? If not I'd recommend we type it since it's likely we'll be using this type of object a lot throughout this project.

packages/fxa-settings/src/components/UnitRowWithAvatar/index.test.tsx
```diff
@@ -3,48 +3,29 @@
...
+    ).toContain('Add');
+    expect(screen.getByTestId('avatar-default')).toBeInTheDocument;
+    expect(screen.queryByTestId('avatar-nondefault')).toBeNull();
   });
 
   it('renders as expected with the user avatar', () => {
-    const { getByTestId, queryByTestId } = render(
-      <UnitRowWithAvatar avatarUrl="some-fake-image.png" />
-    );
+    render(<UnitRowWithAvatar avatarUrl="some-fake-image.png" />);
```
jodyheavener, line 22: worth asserting the URL? idk

packages/fxa-settings/src/components/Avatar/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const Avatar = ({ avatarUrl, className }: AvatarProps) => {
+  if (avatarUrl) {
+    return (
+      <img
+        data-testid="avatar-nondefault"
+        src={avatarUrl}
+        alt="Your avatar"
+        className={classNames('rounded-full', className)}
+      />
```
jodyheavener, line 22: Should we add a background color to this in the event the image fails to load?

packages/fxa-settings/src/components/Avatar/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    );
+  }
+
+  return (
+    <DefaultAvatar
+      data-testid="avatar-default"
+      role="img"
+      aria-label="Default avatar"
```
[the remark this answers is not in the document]
LZoog: I think "default" gives a screen reader more context since there _is_ a difference between the two, but it's probably up to @MeridelW.

packages/fxa-settings/src/components/Settings/index.tsx
LZoog: Yes, this should be taken out. I added this when I was trying to get fancy separating primary email and secondary emails and the type was no longer inferred so good catch.

packages/fxa-settings/src/components/UnitRowWithAvatar/index.test.tsx
LZoog: Hmm since it's just `#` now maybe not, but when this gets changed to `<Link` later, yes.

packages/fxa-settings/src/components/Avatar/index.tsx
LZoog: 👍 That'll be nicer than the flash of alt text, although I think we eventually want this to be a small grey spinner similar to what we've got now. I didn't lift that over because I think it can be simplified a bit (and is a bunch of custom SCSS).

jodyheavener: Works for me. If you want to leave it for now that's fine.

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -6,11 +6,32 @@ import React from 'react';
...
+  const right = (
+    <HeaderContentRight
+      {...{
+        avatarUrl,
+        primaryEmail,
+      }}
+    />
+  );
+
+  return <Header left={leftHeaderContent} {...{ right }} />;
```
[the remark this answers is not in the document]
LZoog: I actually meant to change `leftHeaderContent` to `HeaderContentLeft` anyway ;)

packages/fxa-settings/src/components/Avatar/index.tsx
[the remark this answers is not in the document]
MeridelW: This is the alt text when user has not set an avatar image, yes? If so, yes, the best alt text would be "Default avatar"

PR #5545, opened 2020-06-03
feat(surveys): evaluate survey conditions for the rendering view only

### LINKED ISSUE #5387: [surveys] Build view path to surveys map lazily
At the time of writing, the code (not yet in master) for constructing the view path to surveys map executes regardless which view the user is on.  This is can be pretty wasteful.  We should only build the list of surveys for the requested view.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1938)
┆Issue Number: FXA-1938

chenba:
Because:
 - eagerly evaluating all the survey conditions can be costly and does
   not work

This commit:
 - evaluate the configured surveys for a view, and only that view, when
   the view is being rendered
 - cache model/state data at the filter function level so that they are
   up to date per navigation but cached for evaluating multiple surveys
   configured for the same view
 - DRY up the filter code

Fixes #5387 (FXA-1938)

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/survey-filter.js
+++ b/packages/fxa-content-server/app/scripts/lib/survey-filter.js
@@ -6,7 +6,11 @@ import UserAgent from './user-agent';
 
 // All but the default export here is to make testing easier.
 
+// This subsitutes a (curried) comparator when we know the result is false
+// without doing the comparison.
 const falseFn = () => false;
+
+// When a condition is not found in a survey's configurations.
 export const NONE = Symbol();
 
 export const participatedRecently = (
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -31,56 +35,189 @@ export const getConditionWithKey = (conditions, key) => {
...
+
+      if (m === undefined) {
+        return undefined;
+      }
+
+      prop = prop || (await m[getPropFnName]());
+
+      return prop;
+    } catch {
+      return undefined;
```
jodyheavener, line 128: Think you can just return without the `undefined` in these couple spots, no?

chenba: heh, true.  I changed them from nulls to undefineds and didn't slow down to think about that.  I'll update.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -31,56 +35,189 @@ export const getConditionWithKey = (conditions, key) => {
...
+ */
+export const createFetchUaFn = (window) => {
+  let ua;
+
+  return () => {
+    if (!window || !window.navigator || !window.navigator.userAgent) {
+      return undefined;
+    }
+
+    ua = ua || new UserAgent(window.navigator.userAgent);
```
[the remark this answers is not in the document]
chenba: It would be defined after the first call (assuming happy path) to the function returned by `createFetchUaFn`, since it's in the closure of that function.

PR #5555, opened 2020-06-04
chore(release): use dashes instead of asterisks in release script

jodyheavener:
Because:

* All our changelogs use dashes for bullets, and our release script generates using asterisks

This commit:

* Change dashes in release script changelog bullets to asterisks for consistency


chenba:
🤘

PR #5556, opened 2020-06-04
feat(css): watch and rebuild css on changes

### LINKED ISSUE #5533: Watch + recompile PostCSS when changes occur
Packages fxa-settings, fxa-react, and fxa-admin-panel (and eventually others) use PostCSS to compile Tailwind. Currently it only builds prior to re/starting each service and compiling Storybook, which means you need to manually recompile when changes to anything Tailwind are made. It can be cumbersome.

To close this, each package that uses PostCSS should support watching its primary Tailwind CSS file, component SCSS files (including from external components?), and the Tailwind config file (including external base config?) and recompile PostCSS when changes are detected.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2025)
┆Issue Number: FXA-2025

dannycoates:
fixes #5533

### DIFF
--- a/packages/fxa-admin-panel/pm2.config.js
+++ b/packages/fxa-admin-panel/pm2.config.js
@@ -42,5 +42,23 @@ module.exports = {
       },
       filter_env: ['npm_', 'BERRY_BIN_FOLDER'],
     },
+    {
+      name: 'admin-css',
+      script: 'yarn build-postcss',
+      cwd: __dirname,
+      env: {
+        PATH,
+      },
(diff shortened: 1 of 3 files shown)


dannycoates:
it takes a few seconds but it does seem to work with HMR

LZoog:
It would be nice for "stopped" in `./pm2 status` not to reflect red, but🤷‍♀️ I think this also could have also been done with nodemon in each package along with the `start` command but maybe this is cleaner.

PR #5563, opened 2020-06-05
feat(settings): start injecting server config into beta settings

### LINKED ISSUE #5550: [setup] Parse config from content server
FXA-1546 set up allowing us to parse the config, but now we need to actual send the server config from the content server.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2034)
┆Issue Number: FXA-2034

jodyheavener:
Closes #5550


jodyheavener:
Fair! I've updated to test the actual HTML string replacement. I was also looking at testing that the request triggers the string replacement for HTML content-type. Looks like something like [node-mocks-http](https://www.npmjs.com/package/node-mocks-http) would work, but I wanted to ask if you knew a better approach we've been using in the content-server (e.g. maybe we already have a dep that I don't know about).

chenba:
Fair question!  I figured we could just mock/stub the arguments for `modifyResponse`.  It does require a little bit of work.

I think `proxyRes` can be a real instance of whatever type it is that the test trigger events and spy on.

I have not used the modules you linked.  But I wasn't thinking of triggering/simulating a request; I was thinking of calling the function directly in the tests.

packages/fxa-content-server/server/lib/beta-settings.js
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+      res.send(body);
+    }
+
+    res.end();
+  });
+}
+
+const settingsMiddleware = createProxyMiddleware({
+  target: 'http://localhost:3000',
```
LZoog, line 58: Not sure if this will work on latest.dev.lcip. We've got the beta page behind an `enableBeta` flag, I _think_ we have that set to false on staging and beyond at the moment so this may need to be an env var instead? (cc @dannycoates)

jodyheavener:
@chenba 

> I think proxyRes can be a real instance of whatever type it is that the test trigger events and spy on.

Good call. I am trying to set this up to use a real response instance but there seems to be some subtle differences with what I'm using versus the response object actually passing through. I'm pulling in `require('express/lib/response')`, and maybe I'm just using it incorrectly, but this doesn't have a `headers` property. Spinning my wheels with this specific test 😛 

```javascript
const proxyingResponse = response;
const proxiedResponse = response;

modifyResponse(proxyingResponse, null, proxiedResponse);

proxyingResponse.send(dummyHtml);
```

chenba:
Apparently proxyRes is a nodejs http.IncomingMessage.  🤷‍♂️

jodyheavener:
Ooh thanks, I'll run with that 👍

jodyheavener:
Oof, I think I'm even more confused now, because the first argument of [onProxyRes](https://github.com/chimurai/http-proxy-middleware#http-proxy-events) (what `modifyResponse` is being passed to) looks like it's supposed to receive an instance of `http.ServerResponse` (an Express [response](https://github.com/expressjs/express/blob/master/lib/response.js)), but it's Express [request](https://github.com/expressjs/express/blob/master/lib/request.js) that utilizes `http.IncomingMessage`.

chenba:
Sorry to have mislead you. :(  

I didn't have the branch checked out so I was just browsing around.  Sounds like you've finally found the correct type though?

jodyheavener:
@chenba Oh no don't worry about that, I was just talking out loud. I did find a solution, but it still feels a little gross. Thoughts?

PR #5565, opened 2020-06-05
fix(customs): Call setRecord instead of setRecords in dataflow

vbudhram:
Since dataflow can only set one record at a time, update method to call `setRecord` instead of `setRecords`

This has a few breaking tests, fixing them up.


packages/fxa-customs-server/lib/dataflow.js
```diff
@@ -83,7 +83,7 @@ module.exports = (config, log, fetchRecords, setRecords) => {
 
         if (recordType) {
           records[recordType][action.suggested_action]();
-          await setRecords(records);
+          await setRecord(records[recordType]);
```
vbudhram, line 86: This is really the only important line change, others are formatting.

Issue #5454, opened 2020-05-26
Update Payment Success Amplitude Data


data-sync-user:
Add sources.data.country from Stripe Customer data to the Amplitude event properties.

/packages/fxa-payments-server/src/lib/amplitude.ts

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1986)
┆Issue Number: FXA-1986

PR #5566, opened 2020-06-05
chore(subscriptions): get the country code from payment details

chenba:
Because:
 - we want some metrics around countries when subscriptions are
   available internationally

This commit:
 - get the ISO country code from the payment details within the latest
   invoice for the subscription and include it in the response

Part of #5454 (closes FXA-2022)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -195,6 +195,22 @@ class DirectStripeRoutes {
...
 
+    // Get the country from the payment details.
+    // There's only one charge (the latest), per Stripe's docs.
+    let sourceCountry;
+    const paymentMethodDetails =
+      subscription.latest_invoice.payment_intent.charges.data[0]
+        .payment_method_details;
+
+    if (
+      paymentMethodDetails &&
```
lmorchard, line 206: Hmm, someday this might be a nice area to clean up when we're further into TypeScript and can use optional chaining

PR #5567, opened 2020-06-05
fix(l10n): on train-172, temporarily disable two failing tests (ar & ru)

jrgm:
I don't really like doing this, but l10n may have diverged between 172 and current translations, making these tests fail.


jrgm:
Aside for all the reformatting, the real change is just `/* ... */` here - https://github.com/mozilla/fxa/pull/5567/files#diff-58a0ccaa2b6a7d583d3d636be92f60ceL1676-L1725

PR #5569, opened 2020-06-05
feat(settings): Add Footer component

### LINKED ISSUE #4868: [header/footer] Create the primary footer
We should share the `<Footer>` component from `fxa-admin-panel` in the shared location. It's OK for all previous styles etc. to be removed in favor of reflecting the footer for Settings, we want this to be consistent

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1576)
┆Issue Number: FXA-1576

LZoog:
Because:
* The new settings page needs the footer with moz logo and links.

This commit:
* Moves Footer from fxa-admin-panel into fxa-react for use in admin-panel and
  new Settings, plus styles it per design in desktop, tablet, and mobile.

fixes #4868

image
image


packages/fxa-settings/src/styles/tailwind.css
```diff
@@ -24,3 +24,15 @@ body {
     @apply text-blue-800;
   }
 }
+
+.smooth-transition {
```
jodyheavener, line 28: Because this class is only used in a single file with elements that already have a wide variety of classes applied to them, is this helper class really necessary?

LZoog: I expect we're going to use this in every place where hovering changes some style, which is going to be quite a lot. These same styles are being applied to the ctas (I didn't use `@apply` there since I'd need to extract this into its own file and import it there, plus it doesn't seem best practice to `@apply` your own custom classes). This actually should also be applied to `link-blue`.

Do you think this makes a case for `@apply smooth-transition`? I just don't like the idea of, if we decide the transition should be longer or shorter, finding and replacing every instance of `transition-150` especially if we could have that set for some other animation, when we could do it in one place. 😩

LZoog:
>  Can we use flexbox to push it to the bottom of the window?

Bah, fine. I don't think anyone will see this in Settings unless they have it open in a vertical monitor but I will comply. 😂

jodyheavener:
> I don't think anyone will see this in Settings unless they have it open in a vertical monitor

I was more thinking of somethin like this:

Screen Shot 2020-06-08 at 12 28 11 PM

LZoog:
> I was more thinking of somethin like this:

... Yeah, I stand corrected ;)

packages/fxa-settings/src/styles/tailwind.css
jodyheavener: If this is really going to be used in multiple areas and not just the footer then leave it in. I'm just trying my best to keep this custom CSS as lean as possible. Yes, _if_ we change a transition or other property it would be more effort than necessary, and I will say I'm less concerned about this class because it's covering a specific area (transitions, and not a broad collection of properties), but I still hesitate to let a single class implement multiple properties when one of the main points of Tailwind is to be using the individual classes to style individual properties.

As an aside, but not too concerned (and more of a trolling), I don't know that `smooth` is necessary since we're likely not going to have a `bumpy` or `jagged` transition. It could simply be `standard-transition`.

LZoog: I do like `standard-transition` better.

I feel like having some custom component classes is to be expected at the beginning of a Tailwind project, there's even a [section of their docs](https://tailwindcss.com/docs/extracting-components/#extracting-css-components-with-apply) that recommends doing this. I understand wanting to keep things lean, but I also don't want to have a design request of updating something, and then it not be reflected across the board because we have duplication that we didn't catch. I understand your concern, but I don't think what we have so far is over the top extraction. 

And oh, I can certainly take the transition styles off of the ctas and make sure the ctas use `cta-neutral standard-transition` (or whatever cta class applies). That'll help a bit.

LZoog:
@jodyheavener should be g2g, added a couple more tests in `AppLayout` as well now that it feels a bit more solid there. -waits to see if you hate `transition-standard`- 😉

packages/fxa-react/components/Footer/index.stories.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
```
jodyheavener, line 1: Wanna stick an MPL notice up in this?

packages/fxa-react/components/Footer/index.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+  it('renders as expected', () => {
+    render(<Footer />);
+
+    const linkMozilla = screen.getByTestId('link-mozilla');
+
+    expect(linkMozilla).toHaveAttribute(
+      'href',
+      'https://www.mozilla.org/about/?utm_source=firefox-accounts&utm_medium=Referral'
+    );
+    expect(linkMozilla.firstElementChild).toHaveAttribute('role', 'img');
```
jodyheavener, line 20: I'm 100% cool with this approach, but I'm curious if this was intentional over using test ID on the child elements? Is it better in some way, or just different?

packages/fxa-react/components/Footer/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+      data-testid="link-privacy"
+      href="https://www.mozilla.org/en-US/privacy/websites/"
+      className="smooth-transition w-full text-xs my-3 hover:text-grey-500 hover:underline mobileLandscape:my-0 mobileLandscape:w-auto mobileLandscape:mx-10 mobileLandscape:self-end"
+    >
+      Website Privacy Notice
+    </LinkExternal>
+    <LinkExternal
+      data-testid="link-terms"
+      href="https://www.mozilla.org/en-US/about/legal/terms/services/"
+      className="smooth-transition w-full text-xs mobileLandscape:self-end hover:text-grey-500 hover:underline mobileLandscape:w-auto"
```
jodyheavener, line 34: Are these `smooth-transition`s leftover?

LZoog: 🤦‍♀️ I _know_ I removed these but had a weird rebasing issue and wonder if my changes here are somehow lost, not sure what happened but yeah I'll remove these, good catch.

packages/fxa-react/components/Footer/index.test.tsx
LZoog: Eh, I could have used a testID here, I just wanted to make sure that the logo was inside of the link.

jodyheavener: Works for me

PR #5573, opened 2020-06-06
fix(customs): flag emails in customs server blocking

vbudhram:


packages/fxa-customs-server/lib/email_record.js
```diff
@@ -197,6 +199,36 @@ module.exports = function(limits, now) {
...
+    this.trimBadLogins(now());
+    this.lf.push(now());
+  };
+
+  EmailRecord.prototype.trimBadLogins = function(now) {
+    if (this.lf.length === 0) {
+      return;
+    }
+    // lf is naturally ordered from oldest to newest
+    // and we only need to keep up to limits.maxBadLoginsPerEmail + 1
```
dannycoates, line 217: since our memcache ops aren't atomic I'd rather do a sort here and not have to think about order even if it's impossible for it to be out of order. this list is small

packages/fxa-customs-server/lib/config/config.js
```diff
@@ -89,6 +89,13 @@ module.exports = function(fs, path, url, convict) {
         format: 'nat',
         env: 'MAX_BAD_LOGINS_PER_IP',
       },
+      maxBadLoginsPerEmail: {
+        doc:
+          'Number failed login attempts within rateLimitIntervalSeconds on a single email before throttling',
+        default: 3,
```
clouserw, line 95: I think 3 is a little low for normal life, but works for now

PR #5576, opened 2020-06-08
feat(storybook): allow Tailwind SCSS compilation support

### LINKED ISSUE #5551: Issue with PostCSS parsing SCSS in Storybook
The compilation process before storybook starts in appears to not be properly parsing SCSS files.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2035)
┆Issue Number: FXA-2035

jodyheavener:
Closes #5551 

Because:

* Currently fxa-settings imports SCSS into a Tailwind file, but outside of Create React App (which handles SCSS compilation in the React app itself) SCSS is not compiled.

This PR:

* Updates both Tailwind and Storybook's setup so that you can now compile down to CSS or SCSS depending on your needs, and ensure Storybook's SCSS loader picks it up


packages/fxa-react/.storybook/preview.js
```diff
@@ -2,4 +2,4 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import '../styles/tailwind.out.css';
+import '../styles/tailwind.out';
```
jodyheavener, line 5: Stylesheet-type-agnostic importing

packages/fxa-settings/package.json
```diff
@@ -4,7 +4,7 @@
   "homepage": "https://accounts.firefox.com/beta/settings",
   "private": true,
   "scripts": {
-    "build-postcss": "postcss src/styles/tailwind.css -o src/styles/tailwind.out.css",
+    "build-postcss": "postcss src/styles/tailwind.scss -o src/styles/tailwind.out.scss",
```
jodyheavener, line 7: "Why is this outputting SCSS here but our other Tailwind setups output CSS?" Because we don't import SCSS in those, and CRA/Storybook/Tailwind setups don't care what they receive now. I am only changing this because it itself because it imports SCSS files.

jodyheavener: I should note that at any point if we _do_ start importing SCSS into those other files, it's as simple as changing the extension.

PR #5621, opened 2020-06-09
feat: add must-change-password script

### LINKED ISSUE #5592: Create must-change-password script
Similar to the [must-reset]() script, we can create a new script must-change-password, which sets the `passwordChangeRequired` on the account table. 

* Support setting on uid or email
* Input is json array of account objects

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2068)
┆Issue Number: FXA-2068

bbangert:
Because:

* We want to require batches of users to change their password.
* The graphql-api package has easy access to the auth database for
  update statements.

This commit:

* Add's a must-change-password script that will batch larger update
  commands to quickly set the lockedAt flag used to determine if the
  user needs to change their password.

Closes #5592


packages/fxa-graphql-api/src/scripts/must-change-password.ts
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env ts-node-script
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+/*
+ Usage:
+
+ dist/fxa-graphql-api/src/scripts/must-change-password.js -i ./reset.json
```
clouserw, line 8: Can you add some more here about the format of `reset.json`?  Thanks

PR #5623, opened 2020-06-09
feat(surveys): add "rate" configuration

### LINKED ISSUE #5420: [surveys] Implement "rate"
In the configs for a survey, there's a "rate".  It currently doesn't affect anything.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1960)
┆Issue Number: FXA-1960

chenba:
Because:
 - we want to limit a survey to a certain percentage of visitors

This commit:
 - use the rate value from a survey's configuration in the process that
   determines whether the visitor should receive a survey

Fixes #5420 (FXA-1960)


chenba:
@LZoog Thanks for volunteering to review this.  In terms of just pure logic, this change is super simple: it compares a random number against a configured number.

packages/fxa-content-server/server/config/surveys.json
```diff
@@ -4,6 +4,7 @@
     "conditions": {
       "browser": "firefox"
     },
+    "rate": 0.05,
```
chenba, line 7: This value is from https://jira.mozilla.com/browse/FXA-1998.  If you wish to see the survey locally, then change it to 0.99 or 1 or something high enough.

PR #5638, opened 2020-06-10
refactor(settings): build settings with content-server config interpolation

jodyheavener:
Hot off the heels of #5563 Danny pointed out that we don't use the Settings CRA server in production, we only serve the pre-built static assets. So here's my proposal for an alternative solution.

The gist is, instead of modifying the body of each response as it comes through, just build the Settings index file with the content-server config baked in.

---

Because:

- Currently we are modifying the response of /beta/settings to include server config, which only applies to local development when we also need it for actual builds

This commit:

- Stops modifying any fxa-settings/fxa-content-server response to inject server config
- Utilizes the already available interpolate-html-plugin provided by CRA to inject server config during builds, both in development and production

---

Note that because I pulled out the proxied-response modification for local development you will now be required to restart your Settings CRA server if you make a change to the content-server config so it rebuilds the index.html file. I could be persuaded to leave it in, but I'm not a fan of the idea of having two completely separate spots where server config can be injected depending on NODE_ENV.

### DIFF
--- a/packages/fxa-settings/.rescriptsrc.js
+++ b/packages/fxa-settings/.rescriptsrc.js
@@ -8,8 +8,28 @@ const {
   componentsJestMapper,
 } = require('fxa-react/configs/rescripts');
 
+function loadServerConfig(config) {
+  const serverConfig = require('../fxa-content-server/server/lib/configuration');
+
+  // CRA already uses the interpolate-html-plugin, so let's
+  // add our server config replacement string to it.
+  config.plugins.forEach((plugin) => {
+    if (
+      plugin.constructor &&
+      plugin.constructor.name === 'InterpolateHtmlPlugin'
+    ) {
+      plugin.replacements['SERVER_CONFIG'] = encodeURIComponent(
+        JSON.stringify(serverConfig)
+      );
+    }
+  });
+
+  return config;
+}
+
 module.exports = [
   permitAdditionalJSImports,
   setupAliasedPaths,
   componentsJestMapper,
+  loadServerConfig,
 ];
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -34,6 +34,7 @@
   "dependencies": {
     "@apollo/client": "^3.0.0-beta.50",
     "classnames": "^2.2.6",
+    "fxa-content-server": "workspace:*",
     "fxa-react": "workspace:*",
     "graphql": "14.5.8",
     "graphql-tag": "^2.10.3",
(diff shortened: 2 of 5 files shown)


jodyheavener:
@dannycoates it looks like this change requires content-server dependencies to be installed, at least for tests. Whats our way for doing this now with yarn workspaces? Looks like we don't use Lerna bootstrap anymore

packages/fxa-settings/.rescriptsrc.js
```diff
@@ -8,8 +8,28 @@ const {
   componentsJestMapper,
 } = require('fxa-react/configs/rescripts');
 
+function loadServerConfig(config) {
+  const serverConfig = require('../fxa-content-server/server/lib/configuration');
```
chenba, line 12: We need to be selective on what to send.  (For example, here's where the config for the content server frontend is defined: https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/routes/get-index.js#L54.)

jodyheavener: Interesting. I think we were sending everything in the previous implementation?

jodyheavener: Would what's in your example above also be what we'd want in the settings frontend? Or some other subset?

chenba: > 
> 
> Interesting. I think we were sending everything in the previous implementation?

Yikes

chenba: > 
> 
> Would what's in your example above also be what we'd want in the settings frontend? Or some other subset?

I haven't been following the settings work that closely.  At the very least, we'd need something for the gql-api and sentry?

chenba:
If this PR drags on for a bit, I think we should revert the previous PRs before tagging the next release.

packages/fxa-settings/.rescriptsrc.js
jodyheavener: Yeah, so we start with the content server config and merge it with settings-specific (including gql stuff and sentry) config [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-settings/src/lib/config.ts).

jodyheavener:
> I think we should revert the previous PRs

Is this a concern about the content-server config? I'll note that currently it's only sent in development.

chenba:
> Is this a concern about the content-server config being erroneously exposed? I'll note that currently it's only sent in development. (which is what this PR is trying to address)

I understand that we won't actually expose the configs.  But we also won't be doing another tag for a couple weeks (I think, not too sure), and reverts aren't difficult, so I'd prefer to have that cleaned up.  Not a must though.

Will reverting break local dev?

jodyheavener:
Fair point, #5648 is up and I'll rebase here once merged.

dannycoates:
@jodyheavener what are the config variables settings is interested in, and are they all known at build time?

dannycoates:
I suspect we do want something like your original PR, except that in prod we want to have a custom route for `/beta/settings` that reads the static `index.html` and subs in the config instead of using the proxy or serving the static file directly. Makes sense?

jodyheavener:
Fair point.

PR #5645, opened 2020-06-11
feat(settings): Add AlertBar

### LINKED ISSUE #4929: [global] Create global message alert component, allow dismissing
Take in a message via props and attach an alert style bar visually fixed underneath the sticky primary header. Allow dismissing by clicking on the "X" similar to the `DialogMessage` in the `fxa-payments-server`. It should have `role="alert"` for screenreaders.

This element will be used to display success messages on the main Settings page and also to display any server errors within the separate page flows. In both cases, it should be the first DOM element hit when tabbing or using a screenreader after getting past the header/navigation in the main content. This PR should cover the creation and functionality of this component, but adding into the Settings component and separate page flows will be done in follow ups.

This should scroll with the user as they make their way down the page. Ensure this looks good across breakpoints.

One example wire: https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/407712010

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1629)
┆Issue Number: FXA-1629

LZoog:
Because:
* The new Settings page needs an alert-style bar to display messages to users.

This commit:
* Creates an AlertBar component in fxa-settings and places it in the correct DOM location
* Moves shared hooks with Modal into hooks.tsx
* Adds titles to AlertBar/Modal close buttons per content strategy request
* Removes 'transition:desktop' class in AppLayout because it canceled out position:fixed Nav functionality

fixes #4929

--

Known issues which I will create a follow up for, these will be fine to do after new Settings is in beta:
* If you reveal the alert bar then activate a modal, `esc` keypress will hide the modal _and_ alert bar when it should just hide the modal.
* If you reveal the alert bar the same time you close the modal (this is what I've done with this PR), focus is changed back to the element that triggered the modal and instead, focus should direct to the alert bar. We'll do this at least once according to the wires (when a user tries to add a secondary email when their primary email hasn't been verified).

### DIFF
--- a/packages/fxa-settings/src/components/Settings/index.tsx
+++ b/packages/fxa-settings/src/components/Settings/index.tsx
@@ -7,14 +7,18 @@ import { useBooleanState } from 'fxa-react/lib/hooks';
 import UnitRow from '../UnitRow';
 import UnitRowWithAvatar from '../UnitRowWithAvatar';
 import Modal from '../Modal';
+import AlertBar from '../AlertBar';
 import { AccountData } from '../AccountDataHOC/gql';
 
 export const Settings = ({ account }: { account: AccountData }) => {
   const [modalRevealed, revealModal, hideModal] = useBooleanState();
+  const [alertBarRevealed, revealAlertBar, hideAlertBar] = useBooleanState();
+
   const onSecondaryEmailConfirm = useCallback(() => {
     console.log('confirmed - resend verification code');
     hideModal();
-  }, [hideModal]);
+    revealAlertBar();
+  }, [hideModal, revealAlertBar]);
 
   const modalHeaderId = 'modal-header-verify-email';
   const modalDescId = 'modal-desc-verify-email';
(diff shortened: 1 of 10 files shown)


packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -7,22 +7,38 @@ import { useBooleanState } from 'fxa-react/lib/hooks';
...
+import AlertBar from '../AlertBar';
 import { AccountData } from '../AccountDataHOC/gql';
 
 export const Settings = ({ account }: { account: AccountData }) => {
   const [modalRevealed, revealModal, hideModal] = useBooleanState();
   const onSecondaryEmailConfirm = useCallback(() => {
     console.log('confirmed - resend verification code');
     hideModal();
+    revealAlertBar();
   }, [hideModal]);
```
jodyheavener, line 19: ESLint is complaining about this on build -- I believe `revealAlertBar` should be added to this array.

LZoog:
> Should we add Stories using the other styles of alert bar (blue / red)?

Oh I didn't see this comment, we can do that when we actually use a red/blue alert bar, could have done that here but don't want to fiddle with that now and get CI passing again 🤔

PR #5594, opened 2020-06-08
chore(docs): Update fxa-settings README with Tailwind + shared component details

LZoog:
Because:
* The settings redesign project has introduced Tailwind into FxA and
  some standards have been set. We need a 'getting started' guide.

This commit:
* Adds to the fxa-settings README.

fixes #5552

PR #5654, opened 2020-06-11
chore(docs): Fix settings README typos

LZoog:
Fixed some typos I noticed while reading over my additions in #5594. In the paragraph edit, `2.75rem` should have been `2.25rem` and I added a clarifying note ('which is equivalent to').


packages/fxa-settings/README.md
```diff
@@ -77,7 +77,7 @@ This makes more sense as a new utility class instead of a component because some
 }
 ```
 
-Assuming this package uses `postcss-import`, all we need to do now is `@import` ctas.scss into our `tailwind.css` file and apply our new utility class:
+Assuming this package uses `postcss-import`, all we need to do now is `@import ./ctas.scss` into our `tailwind.css` file and apply our new utility class:
```
jodyheavener, line 80: hah

PR #5655, opened 2020-06-11
feat(dev): handle skip ci label to skip tests

dannycoates:
Set this label if you want to skip ci tests on a PR. Add `fxa-` labels on top to skip specific packages.

### DIFF
--- a/.circleci/modules-to-test.js
+++ b/.circleci/modules-to-test.js
@@ -4,12 +4,17 @@ const { moduleDependencies } = require('../package.json').fxa;
 const fs = require('fs');
 const path = require('path');
 
-process.on('unhandledRejection', e => {
+process.on('unhandledRejection', (e) => {
   console.error(e);
   process.exit(1);
 });
 
-const testable = fs.readdirSync(path.resolve(__dirname, '..', 'packages'), { withFileTypes: true }).filter(de => de.isDirectory()).map(de => de.name);
+const testable = fs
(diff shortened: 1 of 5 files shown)


.circleci/modules-to-test.js
```diff
@@ -39,6 +46,31 @@ async function getModules(org, repo, prNumber) {
...
+async function modulesToSkip(org, repo, prNumber, branch) {
+  try {
+    const labelRes = await fetch(
+      `https://api.github.com/repos/${org}/${repo}/issues/${prNumber}/labels`
+    );
+    const labels = await labelRes.json();
+    const skipModules = branch.startsWith('email-service')
+      ? []
+      : ['fxa-email-service'];
+    if (Array.isArray(labels) && labels.find((l) => l.name === '🙈 skip ci')) {
```
6a68, line 58: 🙈 😆

jodyheavener:
Can we add some documentation for this? Even a little note in the Contributing doc would be helpful

.circleci/modules-to-test.js
dannycoates: I thought the emoji might sort to the top of the label list, but no. still worth keeping though :)

clouserw:
I added what I think the label would be to Jira, but afterwards I realize we don't sync PRs to jira anyway, so, we probably won't see it there.

jodyheavener:
Thanks @dannycoates :D

PR #5652, opened 2020-06-11
fix(devices): improve lastAccessTime accuracy for OAuth devices

eoger:
Continuation of https://github.com/mozilla/fxa/pull/5618.

PR #5656, opened 2020-06-12
fix(devices): List refresh tokens by direct db access, not internal request routing

rfk:
@eoger I spent a big of time trying to get #5652 working using the `oauthdb` absatraction, but it ended up getting out of hand. Then I noticed that some newer routes were just directly importing `./oauth/db` and using that, and doing similarly here makes things much easier.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/devices-and-sessions.js
+++ b/packages/fxa-auth-server/lib/routes/devices-and-sessions.js
@@ -7,6 +7,7 @@
 const { URL } = require('url');
 const Ajv = require('ajv');
 const ajv = new Ajv();
+const hex = require('buf').to.hex;
 const error = require('../error');
 const fs = require('fs');
 const isA = require('@hapi/joi');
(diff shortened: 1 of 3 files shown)


rfk:
(Note that this targets merge to #5652 rather than to master, and we should squash them together into a single commit for subsequent merge to master)

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -28,7 +29,7 @@ const DEFAULT_COMMAND_TTL = new Map([
 module.exports = (
   log,
   db,
-  oauthdb,
+  oauth,
```
eoger, line 32: Is `oauth` the Best Practice (r) name for the raw db?

rfk: I just copied this from other routes that were using it; `oauthdb` seems better but risks confusion with the other `oauthdb`.

PR #5668, opened 2020-06-12
feat(settings | content server): inject server config into settings requests

### LINKED ISSUE #5550: [setup] Parse config from content server

jodyheavener:
Third time's the charm?

Because:

- We need to make a subset of the Content Server config available to the new Settings React app, in both development and production

This PR:

- Sets up response modification for `/beta/settings` to inject server config.
  - In development this is the proxied request to the CRA server
  - In production this is the static index file that we build when deploying

Closes #5550 hopefully?

---

@dannycoates can I get your take on this? You've been helpful in the other PRs/Slack in guiding this setup.


packages/fxa-content-server/server/lib/beta-settings.js
```diff
@@ -0,0 +1,103 @@
...
+
+const { readFileSync } = require('fs');
+const { join } = require('path');
+const { createProxyMiddleware } = require('http-proxy-middleware');
+const config = require('./configuration');
+
+const env = config.get('env');
+const settingsIndexFile =
+  env !== 'development'
+    ? readFileSync(join('../../../', 'beta/settings', 'index.html'), {
```
jodyheavener, line 13: I got this location from where it looks like the build will exists [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/webpack.config.js#L204). Is this location correct?

dannycoates: try something like:
```
join(
  __dirname,
  '..',
  '..',
  config.get('static_directory'),
  'beta',
  'settings',
  'index.html'
)
```

packages/fxa-content-server/server/lib/beta-settings.js
```diff
@@ -0,0 +1,113 @@
...
+  sentry: {
+    dsn: config.get('sentry.client_errors_dsn'),
+  },
+  servers: {
+    gql: {
+      url: config.get('settings_gql_url'),
+    },
+    auth: {
+      url: config.get('fxaccount_url'),
+    },
```
chenba, line 39: These two server configs are not in `fxa-content-server/server/lib/configuration.js` AFAICT.

chenba: hahaha never mind 🤦

packages/fxa-content-server/server/lib/beta-settings.js
```diff
@@ -0,0 +1,113 @@
...
+let settingsIndexFile;
+const env = config.get('env');
+const settingsIndexPath = join(
+  __dirname,
+  '..',
+  '..',
+  config.get('static_directory'),
+  'beta',
+  'settings',
+  'index.html'
```
chenba, line 19: I can't find this file.

packages/fxa-content-server/server/bin/fxa-content-server.js
```diff
@@ -89,11 +93,9 @@ function makeApp() {
...
       })
     );
-    const { createProxyMiddleware } = require('http-proxy-middleware');
-    app.use(
-      '/beta/settings',
-      createProxyMiddleware({ target: 'http://localhost:3000', ws: true })
-    );
+    app.use('/beta/settings', useSettingsProxy);
+  } else {
+    app.get('/beta/settings', modifySettingsStatic);
```
chenba, line 98: nit^2: could define a const for '/beta/settings' so there's just one copy.

packages/fxa-content-server/server/lib/beta-settings.js
jodyheavener: This is the location of the build file, so if you run `yarn build` it will production build everything and you'll find it at `dist/beta/settings/index.html`

jodyheavener:
> What's the reason for having both ways of including the configs in the response?

If I'm not mistaken the two distinct setups are needed for different things:

- In development we are running the Content Server as well as the Settings CRA server (whose requests are proxied through the Content Server). In this case we are checking each request and if the response looks like the Settings HTML file we inject the server config.
- In production the Settings files are static files, so we are listening for any GET request to /beta/settings/ and responding with the index file modified to inject the server config.

That being said, if there is a more streamlined way to combine these two approaches I am definitely open to suggestions.

packages/fxa-content-server/server/lib/beta-settings.js
chenba: Thanks! TIL

chenba:
I didn't know that the index template is copied during build time when I made that comment.

Since there's just one copy of that file: 👍

PR #5687, opened 2020-06-17
refactor(devices): make the devices notify route a no-op

eoger:
Connects to https://github.com/mozilla/application-services/issues/3230, we might want to wait a little bit before merging till Firefox for Android gets migrated since Fennec is still stuck using the legacy send tab mechanism that uses `/devices/notify`.


clouserw:
@eoger is this still on hold?  Should we land it?

eoger:
I'll re-open in a month

PR #5688, opened 2020-06-17
fix(experiment): Fix attempt for qr code cad enrollment

### LINKED ISSUE #5476: CAD via QR code enrollment not even with other groups
After looking at initial amplitude graphs, the enrollment data are not even for each group. 

Ref: https://analytics.amplitude.com/mozilla-corp/dashboard/nw8gr6m

Investigating why enrollment could be off.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1997)
┆Issue Number: FXA-1997

vbudhram:
Fixes #5476 maybe?

This removes the `newsletterCadChooser` check that should have in theory made the qr code and newsletter experiment exclusive. Since we are not running the newsletter experiment yet, it is probably safe to remove the check to see if it was interfering with the qr code experiment.

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/qr-code-cad.js
+++ b/packages/fxa-content-server/app/scripts/lib/experiments/grouping-rules/qr-code-cad.js
@@ -17,7 +17,7 @@ const GROUPS = [
   'treatment-b', // Screen displayed in non-sms markets
 ];
 
-const ROLLOUT_RATE = 0.0;
+const ROLLOUT_RATE = 1.0;
 
 module.exports = class QrCodeCad extends BaseGroupingRule {
   constructor() {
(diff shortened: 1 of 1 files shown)


vbudhram:
> Is there a way we could verify the behavior in automated tests?

The mutually exclusive experiment tests are [here](https://github.com/mozilla/fxa/blob/cc595fc2b44bc1e15857c016539622a6a750bd5d/packages/fxa-content-server/app/tests/spec/lib/experiments/grouping-rules/base.js#L207) and should have caught the enrollment issues. I looked at those tests and it appears to be working as expect. Not really sure what is going on. 

If this does fix the issue then there is most likely an issue with mutually exclusive experiment logic. Luckily we don't have to many of those 😅

Issue #4322, opened 2020-02-28
Server endpoint & DB storage for ecosystem_anon_id


data-sync-user:
Add an endpoint to POST ecosystem_anon_id to the server for storage & store to DB.

Need to store historical anon_ids for CCPA account deletion compliance.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1229)
┆Issue Number: FXA-1229

data-sync-user:
➤ Jared Hirsch commented:

I've already started on this work while working on FXA-1230.

data-sync-user:
➤ Vijay Budhram commented:

Jody your PR will close this, assigning to you!

Issue #4321, opened 2020-02-28
Add ecosystem_anon_id to core accounts table


data-sync-user:
Add a new column to the core accounts table to hold the current ecosystem_anon_id

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1230)
┆Issue Number: FXA-1230

data-sync-user:
➤ Jared Hirsch commented:

I wound up working on a number of related issues at once :-\

I'll have a PR out for this today, but we won't want to land it until after the virtual all hands, because ops would like to delay making database changes with the current release.

data-sync-user:
➤ Chelsea Lewis commented:

[~jhirsch@mozilla.com] iirc this is 'done' but needs some Ops help to make those database changes? Should this be moved to the current sprint accordingly or?

data-sync-user:
➤ Jared Hirsch commented:

[~clewis@mozilla.com] Yeah, this just landed with help from [~vbudhram@mozilla.com]. I'll mark this as done and add a note to the deploy doc about the change needed to the accounts table. I've previously discussed this DB change with [~jmorrison@mozilla.com] and he thought it should work out fine in production.

data-sync-user:
➤ Jared Hirsch commented:

Fixed by https://github.com/mozilla/fxa/pull/5689

Issue #512, opened 2018-03-14
Support verifier versioning and opportunistic verifier upgrades


rfk:
(Fair warning: this is a long one!  But I've been thinking about this for a while and wanted to get it all organized and out of my head.)
 
The [onepw auth protocol](https://github.com/mozilla/fxa-auth-server/wiki/onepw-protocol) hard-codes several details about how to take the user's password and turn it into a verifier hash, key-wrapping key, etc.  We're now at the point where it would be useful to change at least two of these details:

* The derivation of `authPW` uses a fixed number of iterations of PBKDF2, which it's [likely worth increasing](https://bugzilla.mozilla.org/show_bug.cgi?id=1320222) in response to evolving device capabilities.  Since the server never handles the user's password, the only way to achieve this is for the client to opportunistically calculate and upload a new verifier the next time the user logs in.
* The derivation of `authPW` uses the account email address as a salt.  We've shipped the ability for users to change their primary email address, but users who do that are now stuck performing two round trips to the server for every login - they first call `/account/login` with an `authPW` derived using the updated primary email, we return an "incorrect email" error, and they try again with an `authPW` derived using the original primary email.  This is also why [changing your primary email address doesn't "free up" the original email](https://github.com/mozilla/fxa-auth-server/issues/2102).  As above, the server can't fix this, and the only way to avoid two round trips is for the client to calculate and upload a new verifier as part of the login flow.

Can we evolve our login protocol to improve both of these situations?  There are some [brief notes](https://github.com/mozilla/fxa-auth-server/wiki/onepw-protocol#changing-client-side-protection) in the original spec, but here are my thoughts on the details of how we might implement it in practice. 

Broadly, we'd need two things:

* An explicit versioning scheme for the `authPW` calculation, so that client and server can ensure they agree on the details.
* A way for the client to opportunistically update the details on the server when it finds that an obsolete version or an obsolete salt has been used.

Here's how it might look...

----

Suppose we introduce a new "version 2" quickStretch algorithm that uses more iterations of PBKDF2.  When a user logs to their account, the client would calculate `authPW` using the version 2 quickStretch params and their email address as the salt, and would attempt the login like so:

```
    > POST /v1/account/login
    > {
    >    "email": "test@example.com",
    >    "authPW": "AAAABBBBCCCCDDDD",
    >    "quickStretchVersion": 2
    >  }
```

[...]

PR #5689, opened 2020-06-17
feat(aet): Add ecosystemAnonId to auth-db and mysql

6a68:
* Add ecosystemAnonId column to accounts table
* Update stored procedures that create or return an account to include
ecosystemAnonId
* Add stored procedure to update ecosystemAnonId for an account, given
the uid
* Update the authdb API to include ecosystemAnonId in endpoints that
create or return an account
* Add endpoint to authdb API to update ecosystemAnonId
* Add db API tests and update existing tests
* Remove references to deleted memory DB from README

I'm looking forward to ensuring the profile, auth, and auth-db API changes all line up correctly, but that seems like a separate issue from what's in issue #4321, so opening this separately.

This PR also closes part of #4322 (except for the historical anonid tracking)

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -126,6 +126,10 @@ function createServer(db) {
   );
   api.post('/account/:id/locale', withIdAndBody(db.updateLocale));
   api.get('/account/:id/sessions', withIdAndBody(db.sessions));
+  api.put(
+    '/account/:id/ecosystemAnonId',
+    withIdAndBody(db.updateEcosystemAnonId)
+  );
 
   api.get('/account/:id/emails', withIdAndBody(db.accountEmails));
   api.post('/account/:id/emails', withIdAndBody(db.createEmail));
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-110-111.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-110-111.sql
@@ -1,0 +1,205 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('110');
+
+-- Add a column to the accounts table to track the user's current
+-- ecosystemAnonId. For simplicity, we store this JWE as a base64 URL-safe
+-- string, and don't encode or decode it here.
+ALTER TABLE accounts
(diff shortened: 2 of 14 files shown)


6a68:
Ah, the joys of debugging `Unhandled rejection blee` 😠

6a68:
This may not be ready to land just yet. I think I need to include some auth-authdb integration to get things working

chenba:
The profile endpoint is borked when the scope wants the ecosystem anon id but the id is null in the account.

packages/fxa-auth-db-mysql/lib/db/schema/patch-110-111.sql
```diff
@@ -0,0 +1,205 @@
...
+        a.uid = (SELECT uid FROM emails WHERE normalizedEmail = LOWER(inEmail))
+    AND
+        a.uid = e.uid
+    AND
+        e.isPrimary = true;
+END;
+
+-- Update the email record stored procedure to return `ecosystemAnonId` along
+-- with other `account` fields.
+CREATE PROCEDURE `emailRecord_5` (
```
vbudhram, line 79: FWIW, I hope to eventually depreciate the `emailRecord` procedure. It is limited because it only checks the account table for the email and verses `accountRecord` which pulls from the emails table. Long read but good description about reasons [here](https://github.com/mozilla/fxa/issues/512).

packages/fxa-auth-db-mysql/db-server/index.js
```diff
@@ -126,6 +126,10 @@ function createServer(db) {
   );
   api.post('/account/:id/locale', withIdAndBody(db.updateLocale));
   api.get('/account/:id/sessions', withIdAndBody(db.sessions));
+  api.put(
```
vbudhram, line 129: I think we could add some remote tests here that go through the entire flow, create account, update anon id, etc.

https://github.com/mozilla/fxa/blob/cc595fc2b44bc1e15857c016539622a6a750bd5d/packages/fxa-auth-db-mysql/db-server/test/backend/remote.js#L144

6a68:
Thanks @chenba! I'll dig into it. I noticed that sign up fails on this branch, too. Not sure I can land it without finishing the cross-service integration.

clouserw:
@6a68 - what's the status of this PR?

6a68:
> @6a68 - what's the status of this PR?

Trying to get tests to pass so it can land. Vijay added a helpful commit, but there are some sms tests that pass locally but fail on CI.

6a68:
Wow, great work @vbudhram! 🍻

Issue #5683, opened 2020-06-16
Revisit removing Redux from the payments-server


data-sync-user:
Toward the end of last year, I realized that we don't actually need Redux in payments-server. We mainly use it to orchestrate HTTP API requests, but in a way that's inefficient and rife with paperwork.

We could strip Redux out of the project entirely. I have a months' old feature branch where I did that, but wanted to get better test coverage and split it into more incremental changes. Wouldn't mind reviving that and seeing if I could get some PRs out of it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2105)
┆Issue Number: FXA-2105

data-sync-user:
➤ Les Orchard commented:

Some previous issues for context:

  * Refactor payments-server API calls to use axios-hooks rather than Redux? ( https://github.com/mozilla/fxa/issues/3526 )
  * Refactor to use Redux hooks in payments-server components ( https://github.com/mozilla/fxa/issues/3020 )

PR #5690, opened 2020-06-17
feat(react): introduce useAwait hook

lmorchard:
Because:

* I'd like to refactor payments-server to no longer use Redux to orchestrate API calls

This PR:

* Introduces a new useAwait hook for managing promise-based API calls closer to the component using the APIs.

issue #5683

### DIFF
--- a/packages/fxa-react/lib/hooks.stories.tsx
+++ b/packages/fxa-react/lib/hooks.stories.tsx
@@ -1,0 +1,110 @@
+import React, { useCallback } from 'react';
+import { storiesOf } from '@storybook/react';
+import { useAwait, PromiseState } from './hooks';
+import LoadingSpinner from '../components/LoadingSpinner';
+
+const API_URL = '//worldtimeapi.org/api/ip';
+
+storiesOf('hooks|useAwait', module)
+  .add('basic', () => (
+    <>
+      <UseAwaitExample />
+      <UseAwaitExample apiUrl="//worldtimeapi.org/api/timezone/Europe/London" />
+      <UseAwaitExample apiUrl="//worldtimeapi.org/api/timezone/America/Detroit" />
+    </>
+  ))
+  .add('with initial state', () => (
+    <UseAwaitExample
+      initialState={{
+        result: initialResult(),
+        pending: false,
+        error: undefined,
+      }}
+    />
+  ))
+  .add('with immediate execution', () => (
+    <UseAwaitExample
+      executeImmediately={true}
+      fetchApiFn={() => fetchApi(API_URL)}
+    />
+  ))
+  .add('with error', () => (
+    <UseAwaitExample apiUrl="//worldtimeapi.org/badPage" />
+  ));
+
+const UseAwaitExample = ({
+  fetchApiFn = fetchApi,
+  apiUrl = API_URL,
+  style = {},
+  executeImmediately = undefined as boolean | undefined,
+  initialState = undefined as PromiseState<ApiResult, any> | undefined,
+}) => {
+  const [apiState, apiExecute, apiReset] = useAwait(fetchApiFn, {
+    executeImmediately,
+    initialState,
+  });
+  const executeWithApiUrl = useCallback(() => apiExecute(apiUrl), [
+    apiUrl,
+    apiExecute,
+  ]);
+
+  const buttonClassNames = 'mr-1 py-1 px-2 bg-grey-100 border-grey-600';
+  return (
+    <div className="m-4 p-4">
+      <button className={buttonClassNames} onClick={apiReset}>
+        Reset
+      </button>
+      <button className={buttonClassNames} onClick={executeWithApiUrl}>
+        Request
+      </button>
+      {apiUrl}
+      <ApiStateDisplay {...apiState} />
+    </div>
+  );
+};
+
+const ApiStateDisplay = ({
+  pending,
+  error,
+  result,
+}: PromiseState<ApiResult, any>) => {
+  if (!pending && !error && !result) {
+    return <p>Reset state</p>;
+  } else if (pending) {
+    return <LoadingSpinner />;
+  } else if (error) {
+    return <p>Error: {error.message}</p>;
+  } else if (result) {
+    const { datetime, timezone, client_ip, abbreviation } = result;
+    return (
+      <div>
+        <p>Hello, client from {client_ip}</p>
+        <p>
+          It is {datetime} in {timezone} ({abbreviation})
+        </p>
+      </div>
+    );
+  }
+  return <p>Unknown state</p>;
+};
+
+// Partial type of worldtimeapi result
+type ApiResult = {
+  abbreviation: string;
+  client_ip: string;
+  datetime: string;
+  timezone: string;
+};
+
+const fetchApi = async (urlToFetch: string): Promise<ApiResult> => {
+  const resp = await fetch(urlToFetch);
+  const data = await resp.json();
+  return data;
+};
+
+const initialResult = (): ApiResult => ({
+  abbreviation: 'BLA',
+  client_ip: '1.2.3.4',
+  datetime: 'blah blah blah',
+  timezone: 'blah/blah',
+});
--- a/packages/fxa-react/lib/hooks.test.tsx
+++ b/packages/fxa-react/lib/hooks.test.tsx
@@ -2,11 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React from 'react';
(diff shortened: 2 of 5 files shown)


lmorchard:
Trying to revive [my branch from last year that removes Redux from payments-server](https://github.com/lmorchard/fxa/commit/8ccea50b63c0e588b0ac85a41d0017c26cd9daaa), but in smaller chunks. So, here's a first chunk that introduces the key hook that enables the change.

I'm sticking it in fxa-react, because it might be generally useful wherever we want to make API calls or use promise-based APIs in components.

Seems a decent chunk of code worth some looking at in isolation from the rest of the changes to replace Redux with it.

I tried finding a pre-existing equivalent to this hook. But, it seemed like everything I found was missing one of these features:

* TypeScript support
* Ability to supply an initial state (e.g. for storybook stories)
* Ability to reset state (e.g. to dismiss an error dialog)
* Retry a request with new parameters

packages/fxa-react/package.json
```diff
@@ -27,7 +27,7 @@
     "start": "pm2 start pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "storybook": "npm run build-postcss && start-storybook -p 6007",
-    "test": "yarn clean && jest"
+    "test": "yarn clean && jest --env=jsdom-fourteen"
```
lmorchard, line 30: Also, still trying to completely understand why this option is needed. These tests work under fxa-payments-server, where they originally lived in my branch. They fail without this option in fxa-react with an error about `MutationObserver`

More context from where I found the option mentioned: https://github.com/facebook/create-react-app/issues/7491#issuecomment-520157683

lmorchard: (Aaand I just noticed that this very thing failed in CI when it worked locally. Ugh.)

lmorchard:
Passed CI, so I'm just going to open it up for review.

packages/fxa-react/lib/hooks.test.tsx
```diff
@@ -68,3 +74,202 @@ describe('useClickOutsideEffect', () => {
...
+    expect(parseState(getByTestId)).toEqual(initialState);
+  });
+
+  it('executes immediately when executeImmediately = true', async () => {
+    const expected = 'asdasdasd';
+    const { findByTestId, getByTestId } = render(
+      <Subject {...{ fn: async () => expected, executeImmediately: true }} />
+    );
+    await findByTestId('pending');
+    await findByTestId('not-pending');
```
LZoog, line 142: What do you think about using `screen` instead of destructuring these?

```
render(
  <Subject {...{ fn: async () => expected, executeImmediately: true }} />
);
await screen.findByTestId('pending');
await screen.findByTestId('not-pending');
```

I know Jody and I have preferred this syntax and started going this route with new tests but we didn't exactly bring it up as a potential new standard.

packages/fxa-react/lib/hooks.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+  abbreviation: 'BLA',
+  client_ip: '1.2.3.4',
+  datetime: 'blah blah blah',
+  timezone: 'blah/blah',
+});
+
+const containerStyle = {
+  margin: '1em',
+  padding: '1em',
+};
```
LZoog, line 114: Jody and I have been trying to use Tailwind classes even in Storybook just to keep things consistent. Should be `className="m-4 p-4"` here, you can checkout that `tailwind.out.css` file for class references on `buttonStyle`, will be easier once the design guide is merged.

packages/fxa-react/lib/hooks.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <p>
+          It is {datetime} in {timezone} ({abbreviation})
+        </p>
+      </div>
+    );
+  }
+  return <p>Unknown state</p>;
+};
+
+// Partial type of worldtimeapi resul
```
LZoog, line 90: ~resul~ result

packages/fxa-react/lib/hooks.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const ApiStateDisplay = ({
+  pending,
+  error,
+  result,
+}: PromiseState<ApiResult, any>) => {
+  if (!pending && !error && !result) {
+    return <p>Reset state</p>;
+  } else if (pending) {
+    return <LoadingSpinner />;
```
LZoog, line 73: Hm, the loader isn't looking right, but I'm seeing that on the `LoadingSpinner` Storybook as well so nothing to do with implementation here.

packages/fxa-react/lib/hooks.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+      initialState={{
+        result: initialResult(),
+        pending: false,
+        error: undefined,
+      }}
+    />
+  ))
+  .add('with immediate execution', () => (
+    <UseAwaitExample
+      executeImmediately={true}
```
LZoog, line 27: When actually being used I'd pass `executeImmediately` and not explicitly `={true}`, but that's just a syntax preference - I thought about suggesting not supplying a default so you have to pass `executeImmediately` or `executeImmediately={false}` to keep it to a required boolean and not optional/possibly `undefined` for a better visual of what's going on and slightly more restrictive typing, but maybe that's more of a chore since we probably don't want to typically execute immediately.

packages/fxa-react/lib/hooks.stories.tsx
jodyheavener: Huh. That it looks correct when being used in fxa-settings, but not in fxa-react, tells me something's up with Tailwind. [I made an issue.](https://jira.mozilla.com/browse/FXA-2152)

packages/fxa-react/lib/hooks.test.tsx
lmorchard: Hmm I kind of don't like that syntax, if only for the redundant `screen` everywhere. If it made things more clear, maybe? But I think the function name is pretty verbose about it's purpose in the scope already.

I think it's also technically a `RenderResult` type, since the result of render can be an individual component and not a full screen of content. We use `renderResult` as a name [in a few spots](https://github.com/mozilla/fxa/search?q=renderResult&unscoped_q=renderResult) where that needs to be passed around as an intact structure

packages/fxa-react/lib/hooks.stories.tsx
lmorchard: Yeah, most of the uses I had in mind for this thing were in response to actions. So, the most common use would not be immediately executed. A few examples that *would* be immediate might be initial loading of plans & subscriptions on app start. The rest would be initiated by form submissions and such

packages/fxa-react/lib/hooks.stories.tsx
lmorchard: I was kind of just trying a quick and dirty style here since it looks like CSS reset nukes all the defaults that I would have otherwise done with. I can tweak a few things, but really it's the behavior that's more important than the style

packages/fxa-react/lib/hooks.test.tsx
LZoog: That suggestion came from reading over this article: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

I've found it to be nice not having to update the destructure and to see a list of available methods on `screen`.  Though, hmm, I suppose you can see that while destructuring as well if you type one letter.

lmorchard:
Rebased against master, tweaked styles and fixed a typo

packages/fxa-react/lib/hooks.test.tsx
lmorchard: Oh, that actually squicks out me a little bit - looks like he's using a `screen` module-global variable rather than `const screen = render(...)`. Missed that on first reading. Maybe a preference, but I *really* don't like that

LZoog: 🤣 Maybe so. I know @jodyheavener likes it, I've liked it as well but as long as we're consistent I'm fine with either approach

jodyheavener: @lmorchard I'd be curious to know more about your dislike of it. If I understand correctly `screen` best reflects what a user actually sees, versus an isolated rendering. Is there a downside I'm not aware of?

LZoog: I just poked through the Jest documentation and all of their examples under "queries" use it. As it seems to be the way to query `document.body`, maybe it makes sense in some cases and less in others. Though, still might come down to personal preferences as in most test cases we render components in isolation. https://testing-library.com/docs/dom-testing-library/api-queries#screen

PR #5694, opened 2020-06-18
fix(password-input): Avoid 'password too short' check on current password

### LINKED ISSUE #5433: The change-password page should not enforce constraints on the old password
Via reddit here:

  https://www.reddit.com/r/firefox/comments/gpavdi/change_password_of_firefox_sync_account_not/

This user appears to have created an account with a short password, perhaps years ago when we didn't have as thorough enforcement of password strength rules. Now they are unable to change their password because it gets rejected for being too short!

It doesn't seem worthwhile to enforce password strength rules on the current account password after it has been set.

LZoog:
Because:
* A user reported a problem with changing their password because their current password is
less than 8 characters - our 'password too short' error message prevents them from submitting.
They likely created their account before our password strength requirements enforced this.

This commit:
* Skips the "password too short" check on the old password input field.
* Changes 'autocomplete' to 'autoComplete' due to a console warning.

fixes #5433

--

Man, I really went down a horrible functional test rabbit hole for this because the React/Backbone unit tests feel awkward. This seems to cover us though - the test fails without the `password-input.js ` change, fails without setting the password fields with `.val(something)`, and does indeed hit `handleSubmit`.


packages/fxa-content-server/app/scripts/views/elements/password-input.js
```diff
@@ -16,7 +16,12 @@ export default {
 
     if (!value) {
       throw AuthErrors.toError('PASSWORD_REQUIRED');
-    } else if (Vat.password().validate(value).error) {
+      // Don't check password length on an old password in case the user
+      // created their password before the 8 character minimum requirement.
```
chenba, line 20: This comment should be in the `else if` block?

packages/fxa-content-server/app/tests/spec/views/settings/change_password.jsx
```diff
@@ -125,6 +125,30 @@ describe('views/settings/change_password', function () {
...
     });
 
+    describe('handleSubmit', () => {
+      const oldPass = 'pass';
+      const newPass = 'NewPass123';
+
+      beforeEach(() => {
+        sinon.stub(user, 'changeAccountPassword').callsFake(() => Promise.resolve({}));
+        sinon.stub(view, 'displaySuccess').callsFake(() => {});
+        return view.submit(oldPass, newPass);
```
chenba, line 135: Test is passing because of this `submit` here.

packages/fxa-content-server/app/tests/spec/views/settings/change_password.jsx
```diff
@@ -125,6 +125,12 @@ describe('views/settings/change_password', function () {
       });
     });
 
+    describe('handleSubmit', () => {
```
chenba, line 128: heh 'handleSubmit' might be a little misleading since that's not called in the test.  (But feel free to keep it; not a deal breaker.)

LZoog: Hrm yeah, the validation happens in that method, but would make more sense with more coverage.

PR #5695, opened 2020-06-18
[HOLD] chore(docs): Replace 'master' with 'main' throughout

6a68:
@dannycoates brought this up on Slack last week, and it seemed like a good smallish "innovation sprint" change to make this afternoon.

The actual work to change FxA’s default remote branch is really just a matter of clicking around the github website UI, but I thought I’d help move things along by updating existing references to an fxa ‘master’ branch throughout our docs and code docs. I also updated some stale links from the pre-monorepo days, since I was in there anyhow.

I’ve also pushed a `main` branch to fxa-private, since `release.sh` refers to it as well. I figure we can move both repos at the same time.

(Unrelated: I think the content-server docs that claim the templates for emails live in content-server are out of date; is that right? In `fxa-content-server/docs/architecture`, the 'Additional Templates' section.)

As of this commit, remaining `master` references are:

1. #5698: mysql or database config-related (I don't know how to proceed here, but seems like something we could handle in a followup)
2. #5699: references to “master key material” and/or “sync master key” in various contexts. I think we should be OK to change those to “primary” or “main” key, but that’ll also require updating sync docs and finally updating our onepw protocol doc, seems like a self-contained bit of work to split off.
3. references to the ‘master’ branch of other repos we don’t control 🤷 
4. #5700: I don’t know if we can change the fxa-l10n repo from ‘master’ to ‘main’ without breaking pontoon integration, so I’ve retained references to ‘master’ in the l10n section of CONTRIBUTING.md.

@fxa-devs R?

### DIFF
--- a/packages/browserid-verifier/README.md
+++ b/packages/browserid-verifier/README.md
@@ -1,5 +1,3 @@
-[![Docker Build Status](https://circleci.com/gh/mozilla/browserid-verifier/tree/master.svg?style=shield&circle-token=da8b9c7c4c778b238eeadb6d8bc695f2be4e22df)](https://circleci.com/gh/mozilla/browserid-verifier/tree/master)
-
 ## A BrowserID verification server
 
 This repository contains a flexible BrowserID verification server authored in
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,7 +225,7 @@ workflows:
       - test-many:
           filters:
             branches:
-              ignore: master
+              ignore: main
             tags:
               ignore: /.*/
       - test-package:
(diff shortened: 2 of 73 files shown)


6a68:
Thanks, Danny! Enjoy your PTO :-)

I've finished the last few easy replacements and filed followup bugs for the rest. The PR description is now updated with those issue links.

6a68:
I can't figure out why all the tests are failing. Tried rebasing against latest, we'll see if that fixes it.

6a68:
I think this might not pass tests until we switch the primary branch to 'main'.

6a68:
@lmorchard pointed out that I missed some files under .circle-ci, trying again.

packages/browserid-verifier/README.md
```diff
@@ -1,5 +1,3 @@
-[![Docker Build Status](https://circleci.com/gh/mozilla/browserid-verifier/tree/master.svg?style=shield&circle-token=da8b9c7c4c778b238eeadb6d8bc695f2be4e22df)](https://circleci.com/gh/mozilla/browserid-verifier/tree/master)
```
clouserw, line 1: Is this removal on purpose?

6a68: Good question. That link pointed to the old pre-monorepo location of the browserid-verifier, and it seemed like the "build passing" badge might not reflect current reality. Should I keep that line, though?

6a68:
Updated the title to ensure this doesn't accidentally get merged before we make the corresponding default branch change in the github admin UI.

Issue #5662, opened 2020-06-12
Payments should support the same locales content-server does


data-sync-user:
From looking at the code, it looks like fxa-payments-server is pulling what locales should be enabled from [https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/public/locales/locales.json](https://github.com/mozilla/fxa/blob/master/packages/fxa-payments-server/public/locales/locales.json)

We should be using the shared list from the rest of FxA at [https://github.com/mozilla/fxa/blob/master/packages/fxa-shared/l10n/supportedLanguages.json](https://github.com/mozilla/fxa/blob/master/packages/fxa-shared/l10n/supportedLanguages.json)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2113)
┆Issue Number: FXA-2113

data-sync-user:
➤ Dave Justice commented:

pr ready for review: https://github.com/mozilla/fxa/pull/5713

PR #5713, opened 2020-06-22
task(payments-server): use shared lang list

meandavejustice:
Pull in shared list of supported languages to payments server

Closes: #5662

## Checklist


packages/fxa-payments-server/src/lib/AppLocalizationProvider.test.js
```diff
@@ -115,17 +113,6 @@ describe('<AppLocalizationProvider/>', () => {
     expect(getByTestId('result')).toHaveTextContent('untranslated');
   });
 
-  it('throws when locales.json is not found', () => {
```
meandavejustice, line 118: No longer using locales.json so this test is no longer needed 👍

PR #5721, opened 2020-06-22
chore(l10n): copy .ftl files from fxa-content-server-l10n into payments

### LINKED ISSUE #5663: Payments server should use .ftl files from fxa-content-server-l10n
It looks like the fxa-payments-server is expecting .ftl files to be in [https://github.com/mozilla/fxa/tree/master/packages/fxa-payments-server/public/locales](https://github.com/mozilla/fxa/tree/master/packages/fxa-payments-server/public/locales) but our ftl files are localized in a separate repository at [https://github.com/mozilla/fxa-content-server-l10n/](https://github.com/mozilla/fxa-content-server-l10n/).  The payments server should be using those ftl files. 

When browsing the site we can see 404 requests hitting the server looking for those files, eg.  [https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/locales/fr-FR/main.ftl](https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/locales/fr-FR/main.ftl)

I'll file a separate issue to make them publicly available.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2114)
┆Issue Number: FXA-2114

### LINKED ISSUE #5664: Make .ftl files publicly available
Browsers expect to be able to load .ftl files dynamically in the JavaScript.  Each locale at [https://github.com/mozilla/fxa-content-server-l10n/](https://github.com/mozilla/fxa-content-server-l10n/) has ftl files in it now that should live in a public place. 

* These should deploy with all the other code, separately, for dev/stage/prod
* The other files in that repository can be public too.  Ie. it doesn't matter if the .po files are out there as well

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2115)
┆Issue Number: FXA-2115

chenba:
Because:
 - Payments need to serve the translated strings in .ftl files

This commit:
 - add a postinstall script to copy the .ftl files from fxa-content-server-l10n
 - update Payments' \_\_version__ to show the git hash of the l10n checkout
 - git ignore the copied files

closes #5663 (FXA-2114)
closes #5664 (FXA-2115)

### DIFF
--- a/packages/fxa-payments-server/.gitignore
+++ b/packages/fxa-payments-server/.gitignore
@@ -1,0 +1,9 @@
+# copied l10n files
+/public/fxa-content-server-l10n
+/public/locales/git-head.txt
+/public/locales/**/*.ftl
+
+# must exclude the original that's copied _into_ fxa-content-server-l10n
+!/public/locales/en-US/*.ftl
+
+/server/lib/fxa-content-server-l10n
--- a/packages/fxa-payments-server/package.json
+++ b/packages/fxa-payments-server/package.json
@@ -3,6 +3,7 @@
   "version": "1.176.0",
   "description": "Firefox Accounts Payments Service",
   "scripts": {
+    "postinstall": "scripts/download_l10n.sh",
     "lint": "npm-run-all --parallel lint:eslint lint:sass",
     "lint:eslint": "eslint .",
     "audit": "npm audit --json | audit-filter --nsp-config=.nsprc --audit=-",
--- a/packages/fxa-payments-server/scripts/download_l10n.sh
+++ b/packages/fxa-payments-server/scripts/download_l10n.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env sh
+
+set -e
+
+if [ -n "$FXA_L10N_SKIP" ]; then
+    echo "Skipping fxa-content-server-l10n update..."
+    exit 0
+fi
+
+if [ -z "$FXA_L10N_SHA" ]; then
+    FXA_L10N_SHA="master"
+fi
+
+DOWNLOAD_PATH="https://github.com/mozilla/fxa-content-server-l10n.git"
+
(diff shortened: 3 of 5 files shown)


dannycoates:
Ok, now we have three things that ad-hoc pull l10n. We ought to come up with a better way soon

chenba:
> 
> 
> Ok, now we have three things that ad-hoc pull l10n. We ought to come up with a better way soon

Created https://jira.mozilla.com/browse/FXA-2156 to follow up.

packages/fxa-payments-server/.gitignore
```diff
@@ -0,0 +1,6 @@
+# copied l10n files
+/public/fxa-content-server-l10n
+/public/locales/git-head.txt
+/public/locales/**/*.ftl
```
lmorchard, line 4: Did we or are we going to *move* the FTL files out of payments-server?

There's no en-US FTL here: https://github.com/mozilla/fxa-content-server-l10n/tree/master/locale/en_US/

lmorchard: Also, we probably *don't* want to move en-US FTL out of payment-server since we'll want to make edits to that along with other components. There's no extraction of strings with FTL, so we manage those like any other code changes

lmorchard: Yeah, we should add an exception for `/public/locales/en-US/*.ftl` here. The "string extraction" step just copies the en-US locale from payments-server over to fxa-content-server-l10n

https://github.com/mozilla/fxa-content-server-l10n/commit/213c9803bae870bcfb1587b2df0b9190f5a1ebc7#diff-b5fabfd5020a3903580439f0be3efa04R87

chenba: Added.  Thanks for catching that!

PR #5728, opened 2020-06-23
Added test for signing in as an existing user

rbillings:
## Because

\* This adds coverage for testing the ability to log in twice

## This commit

\* adds a test for signing in an existing user

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/sign_in.js
+++ b/packages/fxa-content-server/tests/functional/sign_in.js
@@ -152,6 +152,34 @@ registerSuite('signin', {
         .then(testElementExists(selectors.SETTINGS.HEADER));
     },
 
+    'signin as an existing user': function () {
+      return (
+        this.remote
+          .then(createUser(email, PASSWORD, { preVerified: true }))
+          .then(openPage(ENTER_EMAIL_URL, selectors.ENTER_EMAIL.HEADER))
+          .then(fillOutEmailFirstSignIn(email, PASSWORD))
+
+          // success is seeing the error message.
+          .then(visibleByQSA(selectors.SETTINGS.HEADER))
+          .then(testElementExists(selectors.SETTINGS.HEADER))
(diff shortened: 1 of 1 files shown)


rbillings:
@vbudhram r?

packages/fxa-content-server/tests/functional/sign_in.js
```diff
@@ -152,6 +152,34 @@ registerSuite('signin', {
...
     },
 
+    'signin as an existing user': function () {
+      return (
+        this.remote
+          .then(createUser(email, PASSWORD, { preVerified: true }))
+          .then(openPage(ENTER_EMAIL_URL, selectors.ENTER_EMAIL.HEADER))
+          .then(fillOutEmailFirstSignIn(email, PASSWORD))
+
+          // success is seeing the error message.
```
vbudhram, line 162: I think this should say seeing the header element is success?

rbillings: Agreed!

rbillings:
@vbudhram - if you approve of the change, would you please merge?

lmorchard:
Looks like this needs a GPG signature before we can merge... I can't remember, is there an easy way for one of us core devs to do that or does the original PR branch need a signed commit?

meandavejustice:
Here is the post I always reference when I need to setup a new machine, https://micropipes.com/blog/2016/08/31/signing-your-commits-on-github-with-a-gpg-key/ (it's from wil's blog)

lmorchard:
Yeah, but I was kind of hoping we didn't have to force rbillings to set up GPG if not already setup

rbillings:
My last key expired, so I created a new GPG key and added it.

rbillings:
Since half of this was done unsigned it seems to be stuck in a weird state so I'm closing and re-doing it,

Issue #5704, opened 2020-06-18
Update content-server to call `remind me later` CAD endpoints


data-sync-user:
This will require the following

* Update our js-client to call the new auth-server endpoints
* Update the existing `remind me later` buttons to use js-client

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2145)
┆Issue Number: FXA-2145

Issue #5703, opened 2020-06-18
Create auth-server API endpoints to trigger `remind me later` CAD


data-sync-user:
The new endpoint should be along these lines

* POST `/emails/reminders/cad`
* This should utilize https://jira.mozilla.com/browse/FXA-2143
* Should not be able to create more than 2 reminders
* Create and expose configuration interval
* Something similar to https://github.com/mozilla/fxa/blob/14b0c2960f517f45786ab2859cbfdc66f1799019/packages/fxa-auth-server/config/index.ts#L1588

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2144)
┆Issue Number: FXA-2144

data-sync-user:
➤ Jody Heavener commented:

[~vbudhram@mozilla.com] is there a need pass any flow data along to the CAD reminder mechanism? Only asking because the verification reminder one seems to receive it.

PR #5729, opened 2020-06-23
feat(auth + content): add endpoint and hook up CAD QR flow to trigger reminder

jodyheavener:
## Because:

* The CAD QR flow has two spots to "reminder me later" that need to actually do that
* We need an auth server endpoint to facilitate creating new CAD reminders for a user

## This PR:

* Sets up the auth server endpoint to talk to the CAD Reminders mechanism for creating new reminders
* Updates the content server JS client to talk to the auth server and create reminders
* Updates the template and view for the CAD QR flow to invoke the JS client that talks to the auth server

Closes: #5703 
Closes: #5704 

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -84,6 +84,7 @@ module.exports = (
   customs,
   push,
   verificationReminders,
+  cadReminders,
   signupUtils,
   zendeskClient,
   /** @type import('../payments/stripe').StripeHelper */
@@ -1035,6 +1036,39 @@ module.exports = (
           secondaryEmail: matchedEmail.email,
           uid,
         });
+
+        return {};
+      },
+    },
+    {
+      method: 'POST',
+      path: '/emails/reminders/cad',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function (request) {
+        log.begin('Account.CadReminderEmail', request);
+
+        const sessionToken = request.auth.credentials;
+        const uid = sessionToken.uid;
+
+        const reminders = await cadReminders.get(uid);
+        const exists = cadReminders.keys.some((key) => {
+          return reminders[key] !== null;
+        });
+
+        if (!exists) {
+          await cadReminders.create(uid);
+          log.info('cad.reminder.created', {
+            uid,
+          });
+        } else {
+          log.info('cad.reminder.exists', {
+            uid,
+          });
+        }
 
         return {};
       },
--- a/packages/fxa-content-server/app/scripts/views/post_verify/cad_qr/get_started.js
+++ b/packages/fxa-content-server/app/scripts/views/post_verify/cad_qr/get_started.js
@@ -8,13 +8,15 @@ import FlowEventsMixin from './../../mixins/flow-events-mixin';
 import FormView from '../../form';
 import Template from 'templates/post_verify/cad_qr/get_started.mustache';
 import preventDefaultThen from '../../decorators/prevent_default_then';
+import { MOZ_ORG_SYNC_GET_STARTED_LINK } from '../../../lib/constants';
 
 class GetStarted extends FormView {
   template = Template;
   viewName = 'get-started';
+  settingReminder = false;
 
   events = assign(this.events, {
-    'click #maybe-later-btn': preventDefaultThen('clickMaybeLater'),
+    'click #maybe-later-link': preventDefaultThen('clickMaybeLater'),
   });
 
   submit() {
@@ -22,7 +24,20 @@ class GetStarted extends FormView {
   }
 
   clickMaybeLater() {
-    return this.navigate('/sms');
+    if (this.settingReminder) {
+      return;
+    }
+
+    this.settingReminder = true;
+    const account = this.getSignedInAccount();
+    return account
+      .createCadReminder()
+      .then(() => {
+        this.navigateAway(MOZ_ORG_SYNC_GET_STARTED_LINK);
+      })
+      .catch(() => {
+        this.settingReminder = false;
+      });
   }
 }
 
--- a/packages/fxa-content-server/app/tests/spec/views/post_verify/cad_qr/get_started.js
+++ b/packages/fxa-content-server/app/tests/spec/views/post_verify/cad_qr/get_started.js
@@ -15,6 +15,7 @@ import User from 'models/user';
 import View from 'views/post_verify/cad_qr/get_started';
 import WindowMock from '../../../../mocks/window';
 import $ from 'jquery';
+import { MOZ_ORG_SYNC_GET_STARTED_LINK } from '../../../../../../app/scripts/lib/constants';
 
 describe('views/post_verify/cad_qr/get_started', () => {
   let account;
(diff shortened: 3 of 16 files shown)


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -1036,6 +1037,30 @@ module.exports = (
...
+      path: '/emails/reminders/cad',
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+      },
+      handler: async function (request) {
+        log.begin('Account.CADReminderEmail', request);
+
+        const sessionToken = request.auth.credentials;
```
vbudhram, line 1054: Since this sends an email, we will need to rate limit it. We would need to add an entry [here](https://github.com/mozilla/fxa/blob/c5eb726cb8c90fcffc136381ab56231a34339a99/packages/fxa-customs-server/lib/actions.js#L42), maybe `sendCadReminder`.

Then call it with `await customs.check(request, sessionToken.email, 'sendCadReminder);`

packages/fxa-auth-server/config/index.ts
```diff
@@ -1631,6 +1631,46 @@ const conf = convict({
       },
     },
   },
+  cadReminders: {
```
vbudhram, line 1634: Lets pull this out, I think it will be easier for me to add in my PR.

packages/fxa-content-server/app/scripts/views/post_verify/cad_qr/get_started.js
```diff
@@ -8,21 +8,32 @@ import FlowEventsMixin from './../../mixins/flow-events-mixin';
...
   }
 
-  clickMaybeLater() {
-    return this.navigate('/sms');
+  async clickMaybeLater() {
+    if (this.settingReminder) {
+      return;
+    }
+
+    this.settingReminder = true;
```
vbudhram, line 31: Is the intention that `this.settingReminder` only allow the user to click the button once? I don't think this buys anything because the page does a navigate away and if the user clicks back, these values would be reloaded. It might be easier for us to rate limit on auth-server side of things.

jodyheavener: I agree about rate limiting, but depending on network speed you could rapidly click the link and create multiple requests. Probably super edge case but couldn't hurt?

packages/fxa-auth-server/lib/routes/emails.js
jodyheavener: Might need clarification here - my understanding is this handler only calls the method to create the reminder request in Redis, and a little further down we limit this to 2 max, so wouldn't this just exit early if multiple requests from the same sessionToken come in?

vbudhram: True, since this doesn't actually send the email it is ok not to put and explicit rate limit check here. 

I've updated the logic a little to create a reminder if one doesn't exist in any interval. Since we are using a sorted set in redis, a user can only have on entry in each time interval.

vbudhram:
I pushed some updates here so worthwhile to get another set of eyes to review.

#### Testing
1. Goto http://localhost:3030/post_verify/cad_qr/get_started. 
2. Click `send me a reminder` button. 
3. Observe redis has an entry under `cadReminders` with users uid and timestamp.

I've used https://www.npmjs.com/package/redis-commander to view entries in redis.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -1036,6 +1037,39 @@ module.exports = (
...
+        const uid = sessionToken.uid;
+
+        const reminders = await cadReminders.get(uid);
+        const exists = cadReminders.keys.some((key) => {
+          return reminders[key] !== null;
+        });
+
+        if (!exists) {
+          await cadReminders.create(uid);
```
jodyheavener, line 1063: I assume this create method handles the type of interval it's creating internally.

packages/fxa-content-server/app/scripts/views/post_verify/cad_qr/get_started.js
```diff
@@ -8,21 +8,36 @@ import FlowEventsMixin from './../../mixins/flow-events-mixin';
...
+    const account = this.getSignedInAccount();
+    return account
+      .createCadReminder()
+      .then(() => {
+        this.navigateAway(MOZ_ORG_SYNC_GET_STARTED_LINK);
+      })
+      .catch(() => {
+        this.settingReminder = false;
```
jodyheavener, line 39: The likelihood of this catching is pretty low, but I wonder if this should `alert` or something minimal to inform of the error. I'm leaning towards no, but could be convinced otherwise.

packages/fxa-content-server/app/scripts/views/post_verify/cad_qr/get_started.js
```diff
@@ -8,21 +8,36 @@ import FlowEventsMixin from './../../mixins/flow-events-mixin';
...
+    if (this.settingReminder) {
+      return;
+    }
+
+    this.settingReminder = true;
+    const account = this.getSignedInAccount();
+    return account
+      .createCadReminder()
+      .then(() => {
+        this.navigateAway(MOZ_ORG_SYNC_GET_STARTED_LINK);
```
jodyheavener, line 36: Oh yeah, @vbudhram I'm assuming this URL is correct for both pages. I picked it up from `read_to_scan`

packages/fxa-content-server/app/tests/spec/views/post_verify/cad_qr/get_started.js
```diff
@@ -57,6 +58,8 @@ describe('views/post_verify/cad_qr/get_started', () => {
       user,
     });
 
+    sinon.stub(account, 'createCadReminder').callsFake(() => Promise.resolve());
```
jodyheavener, line 61: IIRC this could just be `sinon.stub(account, 'createCadReminder').resolves();`

packages/fxa-content-server/app/tests/spec/views/post_verify/cad_qr/ready_to_scan.js
```diff
@@ -57,6 +58,8 @@ describe('views/post_verify/cad_qr/ready_to_scan', () => {
       user,
     });
 
+    sinon.stub(account, 'createCadReminder').callsFake(() => Promise.resolve());
```
jodyheavener, line 61: same here (i think)

packages/fxa-content-server/tests/functional/post_verify/cad_qr.js
```diff
@@ -72,8 +73,20 @@ registerSuite('cad_qr_signin', {
           .then(
             testElementExists(selectors.POST_VERIFY_CAD_QR_GET_STARTED.HEADER)
           )
+          .then(click(selectors.POST_VERIFY_CAD_QR_GET_STARTED.LATER))
+          .getCurrentUrl()
+          .then(function (url) {
+            assert.isTrue(url.includes('mozilla.org'));
```
jodyheavener, line 79: Worth noting that I'm using a partial match check instead of `testUrlEquals(MOZ_ORG_SYNC_GET_STARTED_LINK)` because this address resolves to a locale-specific address (in my case `en-CA`) and thus always fails.

vasilicatamas:
I accessed the https://accounts.stage.mozaws.net/post_verify/cad_qr/get_started, clicked the “I’ll do this later, send me a reminder” link and after the script was run I received the following error: ReferenceError: require is not defineddebugger eval code:21:16.

And I’m not sure I understand step 5: Check received me.

@vbudhram are there other steps that can be manually performed in order to verify this implementation?

PR #5735, opened 2020-06-23
fix(shared): ensure *.ts tests run

lmorchard:
- also fix a Sentry test that failed once it started running

### DIFF
--- a/packages/fxa-shared/lib/sentry.ts
+++ b/packages/fxa-shared/lib/sentry.ts
@@ -5,6 +5,12 @@
 import * as Sentry from '@sentry/browser';
 import Logger from './logger';
 
+// HACK: allow tests to stub this function from Sentry
+// https://stackoverflow.com/questions/35240469/how-to-mock-the-imports-of-an-es6-module
+export const _Sentry = {
+  captureException: Sentry.captureException,
+};
+
 const ALLOWED_QUERY_PARAMETERS = [
   'automatedBrowser',
   'client_id',
@@ -160,7 +166,7 @@ SentryMetrics.prototype = {
           );
         }
       });
-      Sentry.captureException(err);
+      _Sentry.captureException(err);
     });
   },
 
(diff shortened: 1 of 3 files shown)


packages/fxa-shared/lib/sentry.ts
```diff
@@ -5,6 +5,12 @@
 import * as Sentry from '@sentry/browser';
 import Logger from './logger';
 
+// HACK: allow tests to stub this function from Sentry
```
lmorchard, line 8: This is a weird but probably least-ugly hack. 

The `Sentry` from `import * as Sentry from '@sentry/browser';` is a convenience object but not actually an export from `@sentry/browser`. 

That means tests can't stub properties of `Sentry` because it's not shared. Both the test module and the subject module have their own `Sentry` instances with references to read-only ES6 exports.

packages/fxa-shared/package.json
```diff
@@ -8,7 +8,7 @@
     "start": "pm2 start pm2.config.js",
     "stop": "pm2 stop pm2.config.js",
     "restart": "pm2 restart pm2.config.js",
-    "test": "tsc --build && node ./scripts/mocha-coverage.js -r ts-node/register --recursive test",
+    "test": "tsc --build && node ./scripts/mocha-coverage.js -r ts-node/register --recursive test/**/*.{js,ts}",
```
lmorchard, line 11: This appears to do the trick for making fxa-shared pick up `*.ts` test modules.

packages/fxa-shared/test/lib/sentry.ts
```diff
@@ -191,10 +191,10 @@ describe('lib/sentry', () => {
 
   describe('captureException', () => {
     it('calls Sentry.captureException', () => {
-      const captureException = sinon.spy(Sentry, 'captureException');
-      captureException(new Error('testo'));
-      sinon.assert.calledOnce(captureException);
-      captureException.restore();
+      const sentryCaptureException = sinon.stub(_Sentry, 'captureException');
```
lmorchard, line 194: This test failed once it started getting run. Needed a tweak to properly mock `captureException` and call the *wrapper* rather than the *mock* as the test subject.

PR #5747, opened 2020-06-23
fix(oauth): Allow token requests to specify arbitrarily large TTL values

### LINKED ISSUE #5660: Return a valid JWT even if the requested TTL is over the maximum limit
When a relying party asks for a JWT with a TTL that is larger than the server allows, we return

`{{Unauthorized` ( 401: {"code":401,"errno":100,"error":"Bad Request","message":"Unauthorized"})}}

We should (and the documentation says) that we will return a valid JWT with the max server TTL in it.

This was originally inspired by https://jira.mozilla.com/browse/FXA-1993

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2110)
┆Issue Number: FXA-2110

6a68:
We still return a token with the TTL no greater than the configured
maximum, but should not throw if the requested TTL is larger.

Fixes #5660.


packages/fxa-auth-server/test/oauth/routes/authorization.js
```diff
@@ -92,6 +92,22 @@ describe('/authorization POST', function () {
...
+          assertion: BASE64URL_STRING,
+          scope: 'bar',
+          state: 'foo',
+          response_type: 'token',
+          ttl: ONE_YEAR_IN_SECONDS,
+        },
+        validation
+      );
+    });
+
```
rfk, line 110: Testing nit: I think it would also be worth testing that the route actually returns the desired smaller TTL in this case.

packages/fxa-auth-server/lib/oauth/routes/authorization.js
```diff
@@ -59,15 +59,11 @@ module.exports = {
...
-      ttl: Joi.number()
-        .positive()
-        .max(MAX_TTL_S)
-        .default(MAX_TTL_S)
-        .when('response_type', {
-          is: RESPONSE_TYPE_TOKEN,
-          then: Joi.optional(),
-          otherwise: Joi.forbidden(),
-        }),
+      ttl: Joi.number().positive().default(MAX_TTL_S).when('response_type', {
```
vladikoff, line 62: @6a68 does this also need to be fixed for the `token` endpoint? 

see https://github.com/mozilla/fxa/blob/6a7106a3f218d029899cceaec9e7558e0f7442eb/packages/fxa-auth-server/lib/oauth/routes/token.js#L100

6a68: Yes, thank you! Discovered that when updating the tests as rfk suggested :-P

6a68:
Thanks for the reviews! Squashed, merging once tests pass

PR #5748, opened 2020-06-24
added CAD reminders to scripts/verification-reminders

dannycoates:
extends #5716

### DIFF
--- a/packages/fxa-auth-server/scripts/verification-reminders.js
+++ b/packages/fxa-auth-server/scripts/verification-reminders.js
@@ -29,6 +29,8 @@ const verificationReminders = require(`${LIB_DIR}/verification-reminders`)(
   config
 );
 
+const cadReminders = require(`${LIB_DIR}/cad-reminders`)(config, log);
+
 const Mailer = require(`${LIB_DIR}/senders/email`)(log, config, oauthdb);
 
 run()
@@ -42,8 +44,15 @@ run()
   });
 
 async function run() {
-  const [allReminders, db, templates, translator] = await Promise.all([
+  const [
+    vReminders,
+    cReminders,
+    db,
+    templates,
+    translator,
+  ] = await Promise.all([
     verificationReminders.process(),
+    cadReminders.process(),
     require(`${LIB_DIR}/db`)(config, log, {}, {}).connect(
       config[config.db.backend]
     ),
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/scripts/verification-reminders.js
```diff
@@ -126,7 +135,77 @@ async function run() {
...
+      },
+      Promise.resolve([])
+    );
+
+    if (failedReminders.length > 0) {
+      log.info('cadReminders.reinstating', {
+        count: reminders.length,
+        key,
+      });
+      return cadReminders.reinstate(key, failedReminders);
```
vbudhram, line 203: The cad reminders library doesn't have the reinstate call. I'm not sure how valuable this is in practice, do we need it?

dannycoates: i'll remove it

packages/fxa-auth-server/scripts/verification-reminders.js
```diff
@@ -126,7 +135,77 @@ async function run() {
...
+    const method = `cadReminder${key[0].toUpperCase()}${key.substr(1)}Email`;
+    const reminders = cReminders[key];
+
+    log.info('cadReminders.processing', {
+      count: reminders.length,
+      key,
+    });
+
+    const failedReminders = await reminders.reduce(
+      async (promise, { timestamp, uid, flowId, flowBeginTime }) => {
```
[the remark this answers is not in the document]
dannycoates: i don't care unless it's hurting something. they'll just be undefined, right?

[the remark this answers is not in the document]
vbudhram: I don't believe they will hurt anything, the mailer should strip undefined header values.

packages/fxa-auth-server/scripts/verification-reminders.js
```diff
@@ -29,6 +29,8 @@ const verificationReminders = require(`${LIB_DIR}/verification-reminders`)(
   config
 );
 
+const cadReminders = require(`${LIB_DIR}/cad-reminders`)(config, log);
```
LZoog, line 32: Maybe this file should be renamed to `reminders.js` or `email-reminders.js` now that it covers CAD and verification reminders?

dannycoates: yes, if this experiment succeeds we have several tasks to refactor and improve the architecture. the goal at this point is to do the least possible so that it's easy to revert.

Issue #5740, opened 2020-06-23
WError: An internal server error occurred


clouserw:
https://sentry.prod.mozaws.net/operations/auth-prod/issues/8781502/?referrer=github_plugin

```
WError: An internal server error occurred
  File "/fxa/packages/fxa-auth-server/lib/routes/subscriptions.js", line 152, in DirectStripeRoutes.createSubscription
    const paymentMethodDetails = subscription.latest_invoice.payment_intent.charges.data[0]
  ?, in runMicrotasks
...
(3 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2169)
┆Issue Number: FXA-2169

data-sync-user:
➤ Barry Chen commented:

This bug is happening in the code we added to get the country of the payment and add that as an event property to Amplitude.

A quick check against Grafana and Amplitude tells me that this is happening with only a subset of the subscriptions.

PR #5750, opened 2020-06-24
fix(payments): handle "charges" missing from the subscription object graph

chenba:
We get the payment source country code from "charges" in the payment
intent of the latest invoice of a created subscription to include in an
Amplitude event.  However, it works most but not all of the time.
Occasionally "charges" is not in the object graph, according an error
captured by Sentry.

Because:
 - "charges" is missing from the subscription object graph in the Stripe
   resposne on some rare instances

This commit:
 - check for the existence of "charges"
 - log the subscription id of the subscription with the missing
   "charges" to help with debugging in Stripe

Closes: #5740 (FXA-2169)

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -195,20 +195,38 @@ class DirectStripeRoutes {
       );
     }
 
-    // Get the country from the payment details.
-    // There's only one charge (the latest), per Stripe's docs.
-    let sourceCountry;
-    const paymentMethodDetails =
-      subscription.latest_invoice.payment_intent.charges.data[0]
-        .payment_method_details;
+    let sourceCountry = null;
 
+    // Get the country from the payment details.
+    // However, historically there were (rare) instances where `charges` was
+    // not found in the object graph, hence the defensive code.
     if (
-      paymentMethodDetails &&
-      paymentMethodDetails.type &&
-      paymentMethodDetails[paymentMethodDetails.type] &&
-      paymentMethodDetails[paymentMethodDetails.type].country
+      subscription.latest_invoice.payment_intent.charges &&
+      subscription.latest_invoice.payment_intent.charges.data
     ) {
-      sourceCountry = paymentMethodDetails[paymentMethodDetails.type].country;
+      // There's only one charge (the latest), per Stripe's docs.
+      const paymentMethodDetails =
+        subscription.latest_invoice.payment_intent.charges.data[0]
+          .payment_method_details;
+
+      if (
+        paymentMethodDetails &&
+        paymentMethodDetails.type &&
+        paymentMethodDetails[paymentMethodDetails.type] &&
+        paymentMethodDetails[paymentMethodDetails.type].country
+      ) {
+        sourceCountry = paymentMethodDetails[paymentMethodDetails.type].country;
+      }
+    } else {
+      Sentry.withScope((scope) => {
+        scope.setContext('stripeSubscription', {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -195,20 +195,38 @@ class DirectStripeRoutes {
...
 
-    // Get the country from the payment details.
-    // There's only one charge (the latest), per Stripe's docs.
     let sourceCountry;
-    const paymentMethodDetails =
-      subscription.latest_invoice.payment_intent.charges.data[0]
-        .payment_method_details;
 
+    // Get the country from the payment details.
+    // However, historically there were (rare) instances where `charges` was
```
lmorchard, line 201: A few questions here:

* Do we have any idea why/when `charges` can be missing from the object? 
* Will this sentry result maybe clue us into that? 
* What happens after sourceCountry is returned as undefined?
** Guessing we'll respond with a 500 ISE when we hit [this validator](https://github.com/mozilla/fxa/blob/217e8e146664582bbc2dc33aa2d936e4871437e2/packages/fxa-auth-server/lib/routes/subscriptions.js#L1151)
** Is there a different / better error we could surface to the user?

chenba: > 
> 
> A few questions here:
> 
>     * Do we have any idea why/when `charges` can be missing from the object?

No.  Once we log the subscription ids on those subscriptions then whoever with prod Stripe access can dig up some clues hopefully.

These errors do not pop up every day (it skipped a day since it started, and none in the last three days), and it's about 2-3% on the days when it does happen (when compared to Amplitude events).

Since we only offer one product in a single currency and only accept credit card payments, I'm having a hard time coming up with variables that might cause this.

I'll peruse the Stripe docs a bit to see if I can find anything.

> 
>     * Will this sentry result maybe clue us into that?

Hopefully.  The Stripe dashboard has logs for pretty much everything, including the original response to the subscription creation request.

> 
>     * What happens after sourceCountry is returned as undefined?
>       
>       * Guessing we'll respond with a 500 ISE when we hit [this validator](https://github.com/mozilla/fxa/blob/217e8e146664582bbc2dc33aa2d936e4871437e2/packages/fxa-auth-server/lib/routes/subscriptions.js#L1151)

Thanks for catching this!  I've added allow(null) to the validation.

>       * Is there a different / better error we could surface to the user?

We don't want the users to get any errors.  (Unless we discover later that "charges" being missing is indicative of some actual issue with the payment.)

bbangert: > Do we have any idea why/when charges can be missing from the object?

From the PaymentIntent API docs for the charges attribute:
> Charges that were created by this PaymentIntent, if any.

A Payment Intent is created automatically to fulfill an invoice on a subscription. The payment intent is purely the attempt to charge the payment method. If the payment attempt does not succeed or has not completed, then there will be no corresponding charge. The docs don't clearly indicate under what conditions there may be no charge, as that should exist even for failed charges.

Perhaps as a safer alternative we should grab the `payment_method` off the [PaymentIntent](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-payment_method), and [retrieve it ](https://stripe.com/docs/api/payment_intents/retrieve) separately.

chenba:
> 
> 
> It would probably be better to fetch the payment_method, since that appears to be guaranteed to exist for a given PaymentIntent.

Thanks for the suggestion.  Since we are including the country in the response only for the purpose of having it as an event property for Amplitude, I don't think it warrants another request to Stripe.

Issue #5732, opened 2020-06-23
Make sure Zip/Postcode field handles different valid lengths correctly


data-sync-user:
Right now we reject anything but 5 character, as we move to accept payments in other markets, we need to make sure that we allow the submission of valid codes (4 char, and 7 char in particular for Canada and UK respectively)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2161)
┆Issue Number: FXA-2161

jodyheavener:
#4717 and possibly #4012 might be helpful here.

data-sync-user:
➤ Wil Clouser commented:

Let's remove our validation of this field and let Stripe's Radar to do the validation.  We also have this input as a numeric keyboard.  It should be changed back to standard.

[~jgruen@mozilla.com] will see if we should restrict on min/max characters.

/cc [~vtamas@mozilla.com]

data-sync-user:
➤ Wil Clouser commented:

[~jgruen@mozilla.com] are 4 and 7 good restrictions on min/max?

data-sync-user:
➤ Wil Clouser commented:

For now, let's ignore any validation here.  We'll let Stripe do it.

data-sync-user:
➤ Barry Chen commented:

I'll leave the empty string validation in place.  Also, it currently does not show the numeric keyboard; nothing to change there.

data-sync-user:
➤ Wil Clouser commented:

👍

PR #5763, opened 2020-06-25
chore(payments): remove zip code validation

chenba:
Because:
 - we want to accept non-US payments
 - Stripe should perform the validation

This commit:
 - remove zip code validation that enforce length and numbers only

Closes: #5732 (FXA-2161)

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -209,8 +209,6 @@ export const PaymentForm = ({
             type="text"
             name="zip"
             label="ZIP code"
-            maxLength={5}
-            minLength={5}
             required
             data-testid="zip"
             placeholder="12345"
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -353,21 +351,14 @@ const validateName: OnValidateFunction = (
 };
 
 const validateZip: OnValidateFunction = (value, focused, _props, getString) => {
-  let valid = undefined;
+  let valid = true;
   let error = null;
-  value = ('' + value).replace(/[^\d]+/g, '').substr(0, 5);
+  value = ('' + value).trim();
```
jodyheavener, line 356: I assume this is a string coercion. Looks funky.

chenba: :nod:

PR #5766, opened 2020-06-25
Release 1.177.0

vbudhram:


AUTHORS
```diff
@@ -156,5 +156,5 @@ vladikoff
 Wil Clouser
 Xavier RENE-CORAIL
 YFdyh000
-Zachary Carter
 Zach Carter
+Zachary Carter
```
jodyheavener, line 160: Lots of doubling up here - is there an easy way to de-dupe these?

Issue #5614, opened 2020-06-09
The links from post verify email received for a stage account point to production


vasilicatamas:
**Note**
This is a follow-up bug for #5400

**Build**
Stage [Train 1.174.0]

**Affected platforms**
Windows 10 x64

**Steps to reproduce**
1. Launch Firefox.
2. Create a Firefox account.
3. Confirm it.
4. Go to the “Account verified. Next, sync another device to finish setup” email.
5. Click on the “Setup next device” button and “another device” link.

**Expected results**
The stage user is redirected to stage links:
- [Setup next device](https://accounts.stage.mozaws.net/connect_another_device/?utm_medium=email&utm_campaign=fx-account-verified&utm_content=fx-connect-device)
- [Another device](https://accounts.stage.mozaws.net/sms?utm_medium=email&utm_campaign=fx-account-verified&utm_content=fx-connect-device) 

**Actual Results**
The stage user is redirected to production links:
- [Setup next device](https://accounts.firefox.com/connect_another_device/?utm_medium=email&utm_campaign=fx-account-verified&utm_content=fx-connect-device)
- [Another device](https://accounts.firefox.com/connect_another_device/?utm_medium=email&utm_campaign=fx-account-verified&utm_content=fx-connect-device)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2080)
┆Issue Number: FXA-2080

PR #5779, opened 2020-06-26
fix(auth): use stage-specific sync url in post verify emails, when in stage

jodyheavener:
## Because

- The post-verify Sync URL points to production in emails sent on stage

## This PR

- Sets the Sync URL in post-verify email to stage address when in stage env

Closes: #5614

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -307,11 +307,6 @@ const conf = convict({
         env: 'PREPEND_VERIFICATION_SUBDOMAIN_SUBDOMAIN',
       },
     },
-    syncUrl: {
-      doc: 'url to Sync product page',
-      format: String,
-      default: 'https://accounts.firefox.com/connect_another_device/',
-    },
     androidUrl: {
       doc: 'url to Android product page',
       format: String,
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -1804,6 +1804,13 @@ conf.set('smtp.subscriptionSupportUrl', `${baseUri}/support`);
 
 conf.set('isProduction', conf.get('env') === 'prod');
 
+if (conf.get('env') === 'stage') {
```
vbudhram, line 1807: I think we can do this by adding something like [here](https://github.com/mozilla/fxa/blob/14b0c2960f517f45786ab2859cbfdc66f1799019/packages/fxa-auth-server/config/index.ts#L1730)

Use content-server config to get the `accounts.stage.firefox.com` part.

jodyheavener: Aha, just noticed [this whole setup](https://github.com/mozilla/fxa/blob/14b0c2960f517f45786ab2859cbfdc66f1799019/packages/fxa-auth-server/config/index.ts#L1734-L1760).

Issue #5762, opened 2020-06-25
[payments][l10n] The wrong language is used when multiple languages are configured


data-sync-user:
I have multiple languages configured in Firefox, with 'en-US' ahead of 'de'.  But I see German on Payments.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2181)
┆Issue Number: FXA-2181

data-sync-user:
➤ Barry Chen commented:

I encountered this bug due to a confluence of factors:

  * by default, Fluent will try to get translation bundles of all available bundles for a language, e.g. 'en' and 'en-GB' when 'en-US' is one of the requested locales
  * Fluent will set a requested but not available locale to be the last fallback locale (in my case, this is 'en-US')
  * the request for 'en-GB' failed because of FXA-2182
  * the request for 'en' failed because it doesn't exist in Payments
  * the request for 'de' succeed

If FXA-2182 did not exist, Fluent would've used 'en-GB' and I would not have noticed this bug.  (At the time of writing this bug is still on Stage.)

PR #5789, opened 2020-06-26
chore(l10n): add en-US to shared list of supported locales

chenba:
Because:
 - `locale/en_US/LC_MESSAGES` exists in the fxa-content-server-l10n repo
   for content-server
 - `public/locales/en-US/main.ftl` exists in fxa-payments-server
 - Fluent will set a requested but not available locale as the last
   fallback locale

This commit:
 - add en-US to the shared list of supported locales

Closes: #5762 (FXA-2181)

## Checklist

### DIFF
--- a/packages/fxa-shared/l10n/supportedLanguages.json
+++ b/packages/fxa-shared/l10n/supportedLanguages.json
@@ -14,6 +14,7 @@
   "dsb",
   "en",
   "en-GB",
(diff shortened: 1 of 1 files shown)


chenba:
See [issue comment](https://github.com/mozilla/fxa/issues/5762#issuecomment-650365516) for a little more detail on the bug.

PR #5792, opened 2020-06-27
refactor(auth-server): restore TS check on stripe

bbangert:
Because:

* A variety of edge cases in our stripe handling were not handled
  correctly. Some of these have manifested in production as errors.

This commit:

* Add's explicit errors to property lookups flagged by type-checking
  that were not previously handled.


packages/fxa-auth-server/package.json
```diff
@@ -73,6 +73,7 @@
     "jed": "0.5.4",
     "jsonwebtoken": "^8.4.0",
     "keyv": "3.0.0",
+    "lodash": "^4.17.15",
```
dannycoates, line 76: I don't see this used anywhere

bbangert: One second, forgot to add that! :)

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -854,11 +938,14 @@ class StripeHelper {
       }
 
       const product = await this.expandResource(
+        // @ts-ignore
```
bbangert, line 941: These are being ignored here, because this entire method is going to be eliminated and we're going to return raw objects to the client.

packages/fxa-auth-server/lib/payments/stripe.js
```diff
@@ -1385,6 +1584,7 @@ class StripeHelper {
       throw error;
     }
 
+    // @ts-ignore
```
bbangert, line 1587: Stripe doesn't like key access like this, and without full TypeScript capabilities its not reasonable to express this in a type-safe manner, thus the ignore.

Issue #932, opened 2019-04-25
Deleting an account causes Firefox "Save password" dropdown to display, without an email address


shane-tomlinson:
<img width="730" alt="Screenshot 2019-04-25 at 12 04 05" src="https://user-images.githubusercontent.com/848085/56731409-50145600-6752-11e9-99fa-85c61884fbd5.png">

I imagine we need to cancel a form submit event.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-826)
┆Issue Number: FXA-826

mirefly:
I tried to cancel form submit event by adding the following to  delete_account.js, but "Save password" dropdown still shows up in Firefox...
```javascript
events: _.extend({}, FormView.prototype.events, {
  'submit form': cancelEventThen('onFormSubmit')
})
````

jodyheavener:
I suspect this may be related to #876 - the issue being a forward navigation with a changed password-type input.

victorphoenix3:
Can I work on this?

PR #5795, opened 2020-06-28
fix(content-server): clean the password input value.

### LINKED ISSUE #606: Account recovery: doorhanger displayed to update the password when tapping on "Cancel"
Environment:
Win10
Stage server - Train 119

Steps to reproduce:
1. Go to Manage account->Account recovery->Enable;
2. Tap on "Generate";
3. On re-enter password modal tap on "Cancel".

Actual results: a door-hanger is displayed to update the password after the modal is dismissed. 

Expected results: no doorhanger is displayed since the password is the same. 

2018_08_24_10_40_37_firefox_accounts_account_recovery_firefox_nightly

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-742)
┆Issue Number: FXA-742

byrw:
## Because

* The doorhanger to save the password appeared on "cancel".

* The password input value was not cleaned before navigating.
## This commit

* Clean the input before navigating to account recovery

fixes #606 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/base.js
+++ b/packages/fxa-content-server/app/scripts/views/base.js
@@ -1137,6 +1137,25 @@ var BaseView = Backbone.View.extend({
         return result;
       });
   },
+
+  /**
+   * Clear validation tooltips and all visible input values.
+   */
+  clearInput() {
+    const $inputEls = this.$('input');
+
+    $inputEls.each((i, inputEl) => {
+      // Called to clear validation tooltips. issues/5680
+      $(inputEl).change();
+    });
+
+    const formEl = this.$('form input:not(".hidden")');
+    if (formEl) {
+      for (let i = 0; i < formEl.length; i++) {
+        formEl[i].value = '';
+      }
+    }
+  },
 });
 
 Cocktail.mixin(
--- a/packages/fxa-content-server/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/settings-panel-mixin.js
@@ -57,22 +57,6 @@ export default {
     this.closePanel();
   },
 
-  clearInput() {
-    const $inputEls = this.$('input');
-
-    $inputEls.each((i, inputEl) => {
-      // Called to clear validation tooltips. issues/5680
-      $(inputEl).change();
-    });
-
-    const formEl = this.$('form  input:not(".hidden")');
-    if (formEl) {
-      for (let i = 0; i < formEl.length; i++) {
-        formEl[i].value = '';
-      }
-    }
-  },
-
   closePanel() {
     this.$el.closest('#fxa-settings-content').removeClass('animate-shadow');
     this.$('.settings-unit').removeClass('open');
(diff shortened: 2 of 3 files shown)


LZoog:
@byrw - thanks for working on this. Because this issue was created 2 years ago, I went to double check that this remains an issue and I'm not seeing an update/save password prompt when clicking "cancel" under Recovery Key. I checked this locally and also in production and cannot reproduce, are you definitely seeing the issue described #606?

I do still see this issue in under "Delete Account," #932. Are you seeing what's happening under "Delete Account" for "Recovery Key" cancel? If not, I think we can close that issue without a PR and if you'd like to take a stab at #932 that would be great. :) Otherwise, I can try investigating further why I'm not seeing the issue described in #606.

recovery-key

byrw:
Hey @LZoog, you have to type a password.

LZoog:
@byrw - Ha, just after posting that comment I deleted it because I realized my mistake, sorry about that. I'm looking at your solution now👍

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -17,7 +17,7 @@
 
             <div class="button-row">
                 <button class="primary-button generate-key-link">{{#t}}Continue{{/t}}</button>
-                <button class="button-link cancel-link">{{#t}}Cancel{{/t}}</button>
+                <button type="button" class="button-link cancel-link">{{#t}}Cancel{{/t}}</button>
```
LZoog, line 20: Buttons receives `type="button"` by default so we shouldn't need to add this, I checked into your branch and took this out and your solution still works without it. :)

packages/fxa-content-server/app/scripts/views/base.js
```diff
@@ -1137,6 +1137,25 @@ var BaseView = Backbone.View.extend({
...
+   */
+  clearInput() {
+    const $inputEls = this.$('input');
+
+    $inputEls.each((i, inputEl) => {
+      // Called to clear validation tooltips. issues/5680
+      $(inputEl).change();
+    });
+
+    const formEl = this.$('form  input:not(".hidden")');
```
LZoog, line 1152: _Tiny_ nit, double space between `form` and `input`.

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
byrw: You're right!

packages/fxa-content-server/app/scripts/views/base.js
byrw: I can't do it. After doing the commit message, the extra space disappear.

LZoog: Do you mean when you try committing the selector as `form input:not(".hidden")` it seems to reformat with a double space between `form` and `input`? That seems strange, we shouldn't have a prettier rule doing that.

byrw: Oh sorry, my mistake

LZoog: My comment may have been ambiguous, I see could have been more precise with what I was pointing out. Thanks for sticking with this, just saw your latest push and it LGTM, once CI gives the green light I'll r+ and merge.

Issue #5675, opened 2020-06-16
Email input on fenix has higher z-index than survey


data-sync-user:
screenshot and context included in this issue (linked to comment for context) https://jira.mozilla.com/browse/FXA-2072?focusedCommentId=86071&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-86071

Should be a quick fix codewise, the timedrain of this issue will be testing in android studio. I'm currently waiting to receive a laptop from HUB as my windows machine failed last week. Then I can knock this one out.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2124)
┆Issue Number: FXA-2124

Issue #5529, opened 2020-06-02
[surveys] Add an 'always' condition


data-sync-user:
We need a way in the surveys config to express "no filtering conditions just show it to everybody".  (Rate and recent survey recipient rules still apply.)  I propose

{always: true}

This issue tracks allowing surveys to always be shown (ie in a development env) regardless of rate or localstorage conditions

.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2024)
┆Issue Number: FXA-2024

data-sync-user:
➤ Dave Justice commented:

When testing, I would use this config.

```
  {
    "id": "showall",
    "conditions": {
      "referrer": null
    },
    "rate": 0.05,
    "view": "settings",
    "url": "https://www.surveygizmo.com/s3/5622367/Password-Reset-1"
  }
```

Which seemed to work for me. I think that the way this should work is if there are no conditions, we should just show the survey. This should be a quick fix. [~bchen@mozilla.com] Do you agree with this method? I feel that if we have always:true AND conditions it may be a little confusing for survey runners. Lemme know 😀

data-sync-user:
➤ Barry Chen commented:

Whatever is intuitive for those who define these configs.  We need a way to say [rate] of all audience.  [~jgruen@mozilla.com] any suggestions?

The example config in the above comment works because it has a non-existent condition type.  If "referrer" is implemented then it might stop working.  I suppose we could simply ask people to use `{"gobbledygook": true}` if they are okay with that.  No code required.  😮

data-sync-user:
➤ John Gruen commented:

maybe we could just test for env=dev/local and if true, set the rate to 1 and ignore the localstorage flag? Also probably good to log the config object in these cases just to provide a developer check?

data-sync-user:
➤ Barry Chen commented:

(I didn't see [~jgruen@mozilla.com] 's comment until [~jheavener@mozilla.com] pointed it out in a PR...)

When I filed this I was not thinking of a special case for testing in dev/local.  I thought we might want a way to simply show surveys to a certain % of users without any other filtering conditions.

A env based rate override for testing is a good idea though.

data-sync-user:
➤ Jody Heavener commented:

[...]

PR #5801, opened 2020-06-29
feat(surveys): add always condition to always show the survey

jodyheavener:
## Because

We wanted an option to always show the survey, regardless of configured rate and the last known time it was shown*.

\* At least, this is what I understood from @johngruen's [comment](https://jira.mozilla.com/browse/FXA-2024?focusedCommentId=86316&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-86316). If I'm off here I can update.

## This change

Allows you to specify `always: true` in survey configuration conditions to have the survey always show, regardless of configured rate and the last known time it was shown

Note this would still yield false if another condition, such as a valid subscription ID or locale, was not satisfied.

Closes: #5529
Closes: #5675 (this is untested, but I have a good feeling about it)

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/survey-filter.js
+++ b/packages/fxa-content-server/app/scripts/lib/survey-filter.js
@@ -350,11 +350,13 @@ export const createSurveyFilter = (
   return async (surveyConfig) =>
     !!(
       surveyConfig &&
-      surveyConfig.rate &&
-      withinRate(surveyConfig.rate) &&
+      parseInt(surveyConfig.rate) >= 0 &&
       surveyConfig.conditions &&
+      (surveyConfig.conditions.always === true ||
+        withinRate(surveyConfig.rate)) &&
       Object.keys(surveyConfig.conditions).length > 0 &&
-      !participatedRecently(previousParticipationTime, doNotBotherSpan) &&
+      (surveyConfig.conditions.always === true ||
+        !participatedRecently(previousParticipationTime, doNotBotherSpan)) &&
       languagesCheck(surveyConfig.conditions, fetchLangs) &&
       // User agent related checks
       userAgentChecks(surveyConfig.conditions, fetchUa) &&
--- a/packages/fxa-content-server/app/styles/modules/_survey.scss
+++ b/packages/fxa-content-server/app/styles/modules/_survey.scss
@@ -2,6 +2,11 @@ $shadow: 0 12px 18px 2px rgba(34, 0, 51, 0.04),
   0 6px 22px 4px rgba(7, 48, 114, 0.12), 0 6px 10px -4px rgba(14, 13, 26, 0.12);
 $survey-height: 360px;
 
+.survey-wrapped {
+  position: relative;
+  z-index: 2;
+}
+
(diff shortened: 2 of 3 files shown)


packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -350,11 +350,18 @@ export const createSurveyFilter = (
   return async (surveyConfig) =>
     !!(
       surveyConfig &&
-      surveyConfig.rate &&
-      withinRate(surveyConfig.rate) &&
+      parseInt(surveyConfig.rate) >= 0 &&
```
jodyheavener, line 353: I'm parsing this value because 0 is falsy and was giving false negatives when the rate was set to 0. I believe this is still safe because `withinRate` handles the actual rate check directly below.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -350,11 +350,18 @@ export const createSurveyFilter = (
...
       surveyConfig.conditions &&
+      withinRate(
+        surveyConfig.conditions.always === true ? 1 : surveyConfig.rate
```
jodyheavener, line 356: Explicitly check `true` so we don't allow other truthy values to accidentally trigger it.

chenba: I prefer `(surveyConfig.conditions.always === true || withinRate(surveyConfig.rate))` over a ternary.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -350,11 +350,18 @@ export const createSurveyFilter = (
...
+      ) &&
       Object.keys(surveyConfig.conditions).length > 0 &&
-      !participatedRecently(previousParticipationTime, doNotBotherSpan) &&
+      !participatedRecently(
+        surveyConfig.conditions.always === true
+          ? null
+          : previousParticipationTime,
+        doNotBotherSpan
+      ) &&
```
chenba, line 364: Same here, prefer `(cond1 || cond2)` over ternary.

chenba:
Let's wait until we figure out what we need.  We can discuss that in the issue comments.

meandavejustice:
> Let's wait until we figure out what we need. We can discuss that in the issue comments.

@chenba Is this about whether we want to add the logic around which env is set before calling this done?

chenba:
> 
> 
> > Let's wait until we figure out what we need. We can discuss that in the issue comments.
> 
> @chenba Is this about whether we want to add the logic around which env is set before calling this done?

No.  There's a fundamental question on what the issue is actually about.  Discussion is in the issue comments.

jodyheavener:
This PR doesn't really suit [the requirements](https://github.com/mozilla/fxa/issues/5529#issuecomment-653156870) (my bad!), so I'm going to close and open a new one.

chenba:
> 
> 
> (my bad!)

Naaaaaaah

PR #5803, opened 2020-06-29
feat(oauth): Store refresh_token last-used-at timestamp in redis.

dannycoates:
Because:

* The last-used-at timestamp on refresh tokens changes frequently,
  and it's good UX to be able to present an accurate value for this
  field.
* We're expecting a significant increase in the volume of clients
  using refresh tokens.
* Writes to MySQL are expensive at scale.

This commit:

* Stores the last-used-at timestamp in redis, where writes are cheaper.

### DIFF
--- a/packages/fxa-auth-server/lib/luaScripts/schema.md
+++ b/packages/fxa-auth-server/lib/luaScripts/schema.md
@@ -1,0 +1,131 @@
+# Schema for data stored in Redis
+
+We currently store three different types of data in redis:
+
+- high-churn sessionToken metadata
+- short-lived OAuth access tokens
+- high-churn refreshToken metadata
+
+Each type of data is stored under a different (and configurable) key prefix
+to avoid collisions.
+
+## Session Token Metadata
+
+Most sessionToken data is stored in MySQL and is updated rarely, but a small
+number of fields may change much more often. Writes to these fields are sent
+to redis because it's better able to cope with the write load, and there are
+no significant consequences to losing the data.
+
+The data is stored as a single record per user, using their `uid` as the key.
+The value is a JSON object whose keys are the ids of session tokens belonging
+to that user, and whose corresponding values are the metadata for that token.
+The metadata is a JSON **_list_** where each item corresponds to a particular
+field of the session token.
+
+For example, if a user has a single session token with the following metadata:
+
+- `id`: "000111222333444555666777888999"
+- `lastAccessTime`: 1583392038878
+- `location`:
+  - city: undefined,
+  - state: undefined
+  - stateCode: undefined,
+  - country: "Australia"
+  - countryCode: "AU"
+- uaBrowser: undefined,
+- uaBrowserVersion: undefined,
+- uaOS: "Windows",
+- uaOSVersion: 10,
+- uaDeviceType: undefined,
+- uaFormFactor: undefined
+
+Then this would be stored in redis as a string value as follows:
+
+- Key: `000111222333444555666777888999`
+- Value: `[1583392038878,[,,,"Australia","AU"],,,"Windows",10,,]`
+
+This scheme is designed to minimize storage space by not encoding field names,
+but leads to some complexity when encoding and decoding.
+
+There is also a legacy format where the value is a straightforward JSON object
+with named fields.
+
+## OAuth Access Tokens
+
+OAuth access tokens are created frequently and are short-lived, making them
+suitable for storage in redis.
+
+The data for each access token is stored as an individual record with a TTL,
+using the `tokenId` as the key and a JSON serialization of the token data
+as the value.
+
+In order to allow easy enumeration of the access tokens belonging to a given
+user, we also maintain a separate record for each user. This record is a redis
+set type whose key is the `uid` and whose members as the `tokenId`s of all
+acesss tokens belonging to that user.
+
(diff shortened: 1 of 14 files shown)


packages/fxa-auth-server/lib/luaScripts/schema.md
```diff
@@ -0,0 +1,126 @@
...
+- uaBrowserVersion: undefined,
+- uaOS: "Windows",
+- uaOSVersion: 10,
+- uaDeviceType: undefined,
+- uaFormFactor: undefined
+
+Then this would be stored in redis as a string value as follows:
+
+- Key: `000111222333444555666777888999`
+- Value: `[1583392038878,[,,,"Australia","AU"],,,"Windows",10,,]`
```
rfk, line 45: Every time I see this, I regret this serialization scheme O_O

packages/fxa-auth-server/lib/luaScripts/schema.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Refresh Token Metadata
+
+Most refreshToken data is stored in MySQL and is updated rarely, but the `lastUsedAt`
+field may change much more frequently. Writes to this field are send to redis
+because it's better able to cope with the write load, and there are no significant
+consequences to losing the data.
+
+The data is stored as a single record per user, using their `uid` as the key.
+The record is a redis hash type whose fields are the `tokenId`s or refresh tokens
```
rfk, line 111: typo (which I expect was mine): "or" -> "of"

packages/fxa-auth-server/lib/luaScripts/setRefreshToken_1.lua
```diff
@@ -0,0 +1,40 @@
...
+      k = v
+    else
+      result[k] = v
+    end
+  end
+  return result
+end
+
+local len = redis.call('hlen', uid)
+if len > 1 then
```
rfk, line 20: So IIUC, if the user has a single refresh token, then we won't prune no matter how old it is. Is the intentional, and if so, why 1 as opposed to some other low number? (I don't thing there's anything wrong with this, just trying to understand the whole mental model here).

packages/fxa-auth-server/lib/oauth/db/index.js
```diff
@@ -113,16 +114,69 @@ class OauthDB {
     return tokens.concat(otherTokens);
   }
 
+  async getRefreshToken(id) {
+    const db = await this.mysql;
+    const t = await db._getRefreshToken(id);
+    if (t) {
+      const extraMetadata = new RefreshTokenMetadata();
```
rfk, line 121: FWIW, it took me a little while to work out that this was internally setting `lastUsedAt` to the current time. When I initially read this code I thought "hrm, I wonder why he's resetting `extraMetadata` to a blank object..?".

You could consider passing in `new Date()` as an explicit argument here for clarity, but it's not a big deal, just wanted to share my first reaction.

packages/fxa-auth-server/lib/luaScripts/setRefreshToken_1.lua
dannycoates: Oh, you remind me that we should set / reset expiry on this whole key.

I wasn't really thinking very hard about the pruning criteria, just that we had some. We learned with access tokens not to rely on pruning with reads. I imagine we could make these numbers tunable with config.

rfk: Makes sense 👍; it would also address those couple of comments in the code about relying on getters for lazy cleanup.

packages/fxa-auth-server/lib/luaScripts/schema.md
dannycoates: Indeed. We could phase it out easily enough. At least it's encapsulated in redis now.

packages/fxa-auth-server/test/oauth/api.js
```diff
@@ -3532,6 +3532,7 @@ describe('/v1', function () {
             refresh_token_id: tokenId,
           }),
         });
+        console.log(res.result);
```
6a68, line 3535: Did you mean to leave this in here?

packages/fxa-auth-server/lib/luaScripts/schema.md
```diff
@@ -0,0 +1,131 @@
+# Schema for data stored in Redis
```
vbudhram, line 1: Thanks for documenting this!

packages/fxa-auth-server/test/oauth/api.js
dannycoates: nope, thanks :)

rfk: IIRC we used to have linting for this, did it get removed?

dannycoates:
From redis `MEMORY USAGE` the size of a refresh token metadata record is 187 bytes. The default expiration is 24h so we can expect our total size of (_tokens used in 24h * 187 bytes_). We can approximate the number used with `select count(token) from refreshTokens where lastUsedAt > DATE_SUB(NOW(), INTERVAL 1 DAY)` though the actual number will likely be higher because the new method of marking a token as used is more accurate.

PR #5805, opened 2020-06-29
feat(ci): cache .yarn in circleci

dannycoates:
caching node_modules didn't really speed us up with npm but caching .yarn might improve our build time. let's see


dannycoates:
## Before
Screen Shot 2020-06-29 at 3 09 37 PM

## After
Screen Shot 2020-06-29 at 3 18 15 PM

Seems worth it!

PR #5809, opened 2020-06-29
feat(emails): Add email template for post registration on non-mobile device

### LINKED ISSUE #5553: Follow up changes for post verify emali
This missed train 174, but we should update the email template

* Use cdn images
* Fix broken tests
* Add mobile specific language to templates

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2037)
┆Issue Number: FXA-2037

LZoog:
Because:
* We want to be able to include a "connect on another desktop device" link for users that register with a tablet (which largely send desktop UA) or desktop, differing slightly from the postVerify email.

This commit:
* Adds a postVerifyNonMobile template and has the mailer send this version of postVerify if during signup the UA is not mobile
* Replaces postVerify image and app badges with CDN version

fixes #5553 

---

@vbudhram, I'm opening this as a draft to make sure you like the approach and because I've got a couple of clarifying questions for you.


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1263,52 +1265,56 @@ module.exports = function (log, config, oauthdb) {
...
-    });
-  };
+        template: templateName,
+        templateValues: {
+          action,
+          androidLinkAttributes: linkAttributes(links.androidLink),
+          androidUrl: links.androidLink,
+          isDesktopTemplate,
+          cadLinkAttributes: isDesktopTemplate
+            ? linkAttributes('https://www.firefox.com')
```
LZoog, line 1304: @vbudhram should this use `config.smtp.redirectDomain`? It seems to use `localhost` when running `write-emails` locally, though I mostly didn't use it because we're only using it to actually _redirect_ the user, yet it is a link here. Would it be proper to use in this instance?

packages/fxa-auth-server/lib/senders/templates/partials/postVerifyPrimary.html
```diff
@@ -0,0 +1,24 @@
...
+  <td valign="top">
+    <h1 style="font-family: sans-serif; font-size: 14px; line-height: 29px; font-weight: normal; margin: 0 0 11px 0; text-align: center;">{{t "Firefox Account verified. You're almost there." }}</h1>
+
+    <h1 class="primary" style="font-family: sans-serif; font-size: 21px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">{{t "Next sync between your devices!"}}</h1>
+
+    <div style="text-align: center; margin-bottom: 20px;">
+      <img src="https://image.e.mozilla.org/lib/fe9915707361037e75/m/5/f9463f08-8831-49fb-bbc4-7b5072cb63be.png"
+           height="175"
+           width="238"
+           alt="Devices"
```
LZoog, line 11: @vbudhram I'm down to rename any of the files here, but since we have a `postVerifySecondary`, it seemed like this as a partial `postVerifyPrimary` used in `postVerify` and `postVerifyNonMobile` made sense, though honestly I think it'd make more sense to call the partial `postVerify` and the two templates `postVerifyPrimaryOnMobile` or something of the sort. I went with the `NonMobile` verbiage since Alex confirmed we can treat tablets like desktops here.

Additionally, I wondered if I should go ahead and remove "Devices" here as the alt text. It's better a11y, but we don't have alt text for any images I'm guessing so you don't see a flash of text?

packages/fxa-auth-server/test/local/routes/utils/signup.js
```diff
@@ -121,13 +121,43 @@ describe('verifyAccount', () => {
...
       assert.equal(
         mailer.sendPostVerifyEmail.args[0][2].service,
         options.service
       );
       assert.equal(mailer.sendPostVerifyEmail.args[0][2].uid, TEST_UID);
     });
+
+    it('should send the non-mobile post account verification email with tablet deviceType', () => {
+      sinon
+        .stub(request, 'headers["user-agent"]')
```
LZoog, line 139: Oh FWIW this is not complete / WIP lol.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -830,6 +830,39 @@ const TESTS = new Map([
...
+      ['X-SES-MESSAGE-TAGS', { test: 'equal', expected: sesMessageTagsHeaderValue('postVerifyNonMobile') }],
+      ['X-Template-Name', { test: 'equal', expected: 'postVerifyNonMobile' }],
+      ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.postVerifyNonMobile }],
+    ])],
+    ['html', [
+      { test: 'include', expected: 'Firefox Account verified. You&#x27;re almost there.' },
+      { test: 'include', expected: 'Next sync between your devices!' },
+      { test: 'include', expected: 'Sync privately keeps your bookmarks, passwords and other Firefox data the same across all your devices.' },
+      { test: 'include', expected: configHref('syncUrl', 'account-verified', 'registration-confirmation', 'utm_source=fxa') },
+      { test: 'include', expected: config.smtp.androidUrl },
```
LZoog, line 846: @vbudhram I also have a question here about the `params` const I set, as my tests here are failing since the `androidUrl` and `iosUrl` both have `utm_source=fxa` appended. I'm sure I can fix that, but, is it fine that `utm_source=fxa` is appended to all URLs or should I figure out how to target _only_ the `connect another desktop device` link?

packages/fxa-auth-server/lib/metrics/amplitude.js
```diff
@@ -40,6 +40,7 @@ const EMAIL_TYPES = {
   postChangePrimary: 'change_email',
   postRemoveSecondary: 'secondary_email',
   postVerify: 'registration',
+  postVerifyNonMobile: 'registration',
```
vbudhram, line 43: I think we can actually get by without introducing a new template here.

packages/fxa-auth-server/lib/routes/utils/signup.js
```diff
@@ -64,6 +65,8 @@ module.exports = (log, db, mailer, push, verificationReminders) => {
       // Our post-verification email is very specific to sync,
       // so only send it if we're sure this is for sync.
       if (service === 'sync') {
+        const onMobileDevice =
+          userAgent(request.headers['user-agent']).deviceType === 'mobile';
         const mailOptions = {
```
vbudhram, line 70: I would pass the `onMobileDevice` value into the `mailerOptions` and pass that into `sendPostVerifyEmail`.

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1263,52 +1265,56 @@ module.exports = function (log, config, oauthdb) {
...
+        ...message,
+        headers,
         subject,
-        supportLinkAttributes: links.supportLinkAttributes,
-        supportUrl: links.supportUrl,
-      },
-    });
-  };
+        template: templateName,
+        templateValues: {
```
vbudhram, line 1298: You could then pass in the `onMobileDevice` value here and check in the template.

packages/fxa-auth-server/lib/senders/email.js
vbudhram: We should probably expose new a config url to download Firefox browser, maybe something like `firefoxDesktopUrl` [here](https://github.com/mozilla/fxa/blob/2d566e40d74caadefaaf883fd3b28334eecde670/packages/fxa-auth-server/config/index.ts#L310) and corresponding `desktopLinkAttributes`. I think this should avoid issue above.

packages/fxa-auth-server/lib/senders/templates/partials/postVerifyPrimary.html
vbudhram: I agree, there is probably some partials optimization that could be done in here. If we used the `postVerify` template in this PR, I think we could spin the optimization stuff into its own issue.

> I wondered if I should go ahead and remove "Devices" here as the alt text

SGTM

LZoog:
Closing in favor of #5817

Issue #5719, opened 2020-06-22
[l10n][payments] Payment form Error Messages Incorrectly Translated


data-sync-user:
Three error messages on the payment form are not correctly translated for German.  They include the English words "is required".  See image below.

(Not sure if this is the correct project file this bug.  Please move to the proper l10n project as needed.)

!image-2020-06-22-16-01-47-245.png!

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2155)
┆Issue Number: FXA-2155
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/13471/image-2020-06-22-16-01-47-245.png">image-2020-06-22-16-01-47-245.png</a>

data-sync-user:
➤ Wil Clouser commented:

A quick look here makes me think these strings are being concatenated together.  The .ftl file has just the field names @ https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/de/main.ftl#L93 ( https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/de/main.ftl#L93 ) but the "is required" isn't in there.  We shouldn't be concatenating strings.

PR #5812, opened 2020-06-30
task(payments-server): l10n msg 4 required fields

meandavejustice:
## Because
Fix an issue with the way we were concatenating error message labels in our payment form. Decided to keep the english fallback in the case that there is no translation in the user's language

Closes: #5719

## Checklist

## Screenshots (Optional)

TRANSLATED:
is-required-correct

FALLBACK TO ENGLISH TRANSLATION:
is-required-falback

### DIFF
--- a/packages/fxa-payments-server/src/components/fields/index.tsx
+++ b/packages/fxa-payments-server/src/components/fields/index.tsx
@@ -214,19 +214,25 @@ export const Input = withLocalization(UnwrappedInput);
 
 type StripeElementProps = { onValidate?: OnValidateFunction } & FieldProps & {
     component: any;
+    getString: Function | undefined;
   } & ReactStripeElements.ElementProps;
 
 export const defaultStripeElementValidator: OnValidateFunction = (
   value,
   focused,
-  props
+  props,
+  getString
 ) => {
   if (!value || value.empty) {
     if (props.required) {
       return {
         value,
         valid: false,
-        error: focused ? null : `${props.label} is required`,
+        error: focused
+          ? null
+          : getString
+          ? getString('input-error-is-required', { label: props.label })
+          : props.label + ' is required',
       };
     }
   } else if (value.error && value.error.message) {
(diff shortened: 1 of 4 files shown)


packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -214,19 +214,25 @@ export const Input = withLocalization(UnwrappedInput);
...
 ) => {
   if (!value || value.empty) {
     if (props.required) {
       return {
         value,
         valid: false,
-        error: focused ? null : `${props.label} is required`,
+        error: focused
+          ? null
+          : `${props.label} ${
```
lmorchard, line 233: We can't handle translated strings with concatenation like this. The "is required" string must have a placeholder like "${fieldLabel} is required` and allow the localizer to swap word order as necessary.

meandavejustice: Gotcha, I'll update

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -214,19 +214,27 @@ export const Input = withLocalization(UnwrappedInput);
...
       return {
         value,
         valid: false,
-        error: focused ? null : `${props.label} is required`,
+        error: focused
+          ? null
+          : `${
+              getString
+                ? getString('input-error-is-required', { label: props.label })
+                : props.label + ' is required'
```
meandavejustice, line 236: @lmorchard updated, I still have the concatenated fallback, it seems it still gets the message across if the translation doesn't exist.

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -214,19 +214,27 @@ export const Input = withLocalization(UnwrappedInput);
...
 ) => {
   if (!value || value.empty) {
     if (props.required) {
       return {
         value,
         valid: false,
-        error: focused ? null : `${props.label} is required`,
+        error: focused
+          ? null
+          : `${
```
lmorchard, line 233: Don't really need the backtick-string at this point - but 🤷‍♂️

meandavejustice: removed the backticks, glad you caught that. I'll merge once we are green again

Issue #5640, opened 2020-06-10
[surveys] Pass matched condition values into the survey


data-sync-user:
Currently we check the configured conditions during the filtering phase to determine eligibility, but we don't surface the matched values.  We could pass these values into the survey as query params.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2094)
┆Issue Number: FXA-2094

data-sync-user:
➤ Dave Justice commented:

This would be great, [~bchen@mozilla.com] you are more familiar with the actual filter code, would you mind sizing this one?

data-sync-user:
➤ John Gruen commented:

This, to me is the most important outstanding remaining work on surveys

data-sync-user:
➤ Dave Justice commented:

I was thinking through this and it seems that we should complete 2095 and 2096 first,

The current implementation states that all of the conditions need to be met in order to show the survey, so passed results would be the same as config. Once we update  https://jira.mozilla.com/browse/FXA-2095 and https://jira.mozilla.com/browse/FXA-2096, we would actually have some variability in the results and it would make sense to return the filtered results instead of just pass the config conditions along to the url.

Does this sound correct [~bchen@mozilla.com]?

data-sync-user:
➤ Barry Chen commented:

Mostly true.  The only exception is the languages condition, where, for example, "en" will match "en" or any en-* languages.

data-sync-user:
➤ Barry Chen commented:

[~jgruen@mozilla.com] for subset matches like subscriptions, do you want the complete set of values or the subset from the survey's conditions?  For example, if the condition is subscription plans [b, c] but the user is subscribed to [a, b, c, p, q], do you want [b, c] passed to the survey, or do you want to know what the user's subscribed to?

data-sync-user:
➤ John Gruen commented:

a,b,c,p,q

PR #5816, opened 2020-06-30
feat(surveys): send matching conditions over to survey gizmo

jodyheavener:
## Because

We currently check the configured conditions during the filtering phase to determine eligibility, but we don't surface the matched values.

## This change

Extract user value based on satisfied conditions and passes them along as query params to Survey Gizmo.

- Originally we only passed along the values that satisfied the conditions, but per [this request](https://github.com/mozilla/fxa/issues/5640#issuecomment-653161604) this has changed to now pass along _all_ items of a given condition category belonging to the user. Some examples of this could be:
	- If the browser locale is `en-CA,fr-CA,zh-TW` and the `language` condition is `['en', 'fr-CA', 'es']`, this would be passing and the returned value would be `en-CA,fr-CA,zh-TW` because either `en-CA` _or_ `fr-CA` could have satisfied the condition.
	- If you have a default avatar and the `hasNonDefaultAvatar` condition is `false`, to indicate that you only want to target users with a default avatar, the returned value would be `false` because you do not have a non-default avatar. This should not be confused as a failing result.
	- This was a funky one, but precedent dictates that `null` is also a valid value, at least when it comes to the `relier` condition. Therefor a `null` value can be used to satisfying a condition, but because of how our URL query params are generated this condition will **not** end up in the URL.
- It is maintained that the absence of a condition during its corresponding check still yields a pass, but in this case a value would not be returned.
- Generated URLs will return the conditions as query params, converting arrays into comma-delimited strings per [this request](https://github.com/mozilla/fxa/pull/5816#discussion_r449193495).

Closes: #5640

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/survey-filter.js
+++ b/packages/fxa-content-server/app/scripts/lib/survey-filter.js
@@ -8,7 +8,10 @@ import UserAgent from './user-agent';
 
 // This subsitutes a (curried) comparator when we know the result is false
 // without doing the comparison.
-const falseFn = () => false;
+const undefinedFn = () => undefined;
+
+// Response for failing conditions
+const failureRes = { passing: false, conditions: {} };
 
 // When a condition is not found in a survey's configurations.
 export const NONE = Symbol();
@@ -53,19 +56,25 @@ export const getConditionWithKey = (conditions, key) => {
  * @param {function} comparator a function that compares two values and returns a boolean
  * @param {string} key the configuration key where the condition value can be found
  * @returns {function} a function that takes conditions and (f) (see description above) and
- *                     returns a boolean
+ *                     returns an object containing the pass/fail state and matching condition
+ *                     value if the condition is satisfied
  */
 export const createConditionCheckFn = (valSource) => (comparator) => (key) => (
   conds,
   fetchFn
 ) => {
-  // This turned out a little janky.  But we want to return true early if the
+  // This turned out a little janky.  But we want to return early with a pass if the
   // condition is not in the configs.
   const condVal = getConditionWithKey(conds, key);
   if (condVal === NONE) {
-    return true;
+    return { passing: true };
   }
-  return valSource(fetchFn)(comparator)(condVal);
+  const result = valSource(fetchFn)(comparator)(condVal);
+  return {
+    passing: result !== undefined,
+    value: result,
+    key: key,
+  };
 };
 
 // Like the function above but creates an async function because (f) can be async.
@@ -74,9 +83,14 @@ export const createAsyncConditionCheckFn = (valSource) => (comparator) => (
 ) => async (conds, fetchFn) => {
   const condVal = getConditionWithKey(conds, key);
   if (condVal === NONE) {
-    return true;
+    return { passing: true };
   }
-  return (await valSource(fetchFn)(comparator))(condVal);
+  const result = (await valSource(fetchFn)(comparator))(condVal);
+  return {
+    passing: result !== undefined,
+    value: result,
+    key: key,
+  };
 };
 
 export const createFetchLanguagesFn = (window) => {
@@ -192,7 +206,7 @@ export const fetchAndApplySourceVal = (fetchFn) => (checkFn) => {
   const x = fetchFn();
 
   if (x === undefined) {
-    return falseFn;
+    return undefinedFn;
   }
 
   return checkFn(x);
@@ -203,7 +217,7 @@ export const asyncFetchAndApplySourceVal = (fetchFn) => async (checkFn) => {
   const x = await fetchFn();
 
   if (x === undefined) {
-    return falseFn;
+    return undefinedFn;
   }
 
   return checkFn(x);
@@ -211,7 +225,7 @@ export const asyncFetchAndApplySourceVal = (fetchFn) => async (checkFn) => {
 
 // Comparator
 export const checkLanguages = (browserLanguages) => (val) => {
-  // True if _any_ of the languages from the condition matches a configured language in the browser.
+  // Returns browserLanguages if _any_ of the languages from the condition matches a configured language in the browser.
   // The language tags in the conditions can be in two forms:
   //  - just the language, e.g. "en", or
   //  - language followed by extlang, script, region, etc. e.g. ("en-CA")
@@ -220,7 +234,7 @@ export const checkLanguages = (browserLanguages) => (val) => {
 
   const separator = '-';
 
-  return val.some((lang) => {
+  const result = val.filter((lang) => {
     lang = lang.toLowerCase();
 
     return browserLanguages.some((browserLang) =>
@@ -229,6 +243,10 @@ export const checkLanguages = (browserLanguages) => (val) => {
         : browserLang.toLowerCase().startsWith(lang)
     );
   });
+
+  if (result.length) {
+    return browserLanguages;
+  }
 };
 
 export const languagesCheck = createConditionCheckFn(fetchAndApplySourceVal)(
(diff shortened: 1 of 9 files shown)


packages/fxa-content-server/app/tests/spec/lib/survey-filter.js
```diff
@@ -577,33 +577,6 @@ describe('lib/survey-filter', () => {
     });
   });
 
-  describe('userAgentChecks', () => {
```
meandavejustice, line 580: Assuming this was just part of getting started on the tests, but why was this one removed?

jodyheavener:
**Reviewer:** I apologize, this ended up becoming a Very Large PR. So large, in fact, that GitHub appears to be collapsing the [survey filter test file](https://github.com/mozilla/fxa/pull/5816/files#diff-913ef57336afdd82b4c000a1aedc0056). Take note when reviewing, and god speed 😅

chenba:
> This was a funky one, but precedent dictates that null is also a valid value, at least when it comes to the relier condition.

IIRC null is sync.  We can double check and maybe have a special case for that one 😢 (perhaps in a follow-up).

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -63,9 +63,13 @@ export const createConditionCheckFn = (valSource) => (comparator) => (key) => (
...
   if (condVal === NONE) {
-    return true;
+    return { passing: true, value: undefined };
   }
-  return valSource(fetchFn)(comparator)(condVal);
+  const result = valSource(fetchFn)(comparator)(condVal);
+  return {
+    passing: result !== undefined,
+    value: result,
+  };
```
chenba, line 72: Comments need to be updated since it mentions the return type.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -63,9 +63,13 @@ export const createConditionCheckFn = (valSource) => (comparator) => (key) => (
   // condition is not in the configs.
   const condVal = getConditionWithKey(conds, key);
   if (condVal === NONE) {
-    return true;
+    return { passing: true, value: undefined };
   }
-  return valSource(fetchFn)(comparator)(condVal);
+  const result = valSource(fetchFn)(comparator)(condVal);
+  return {
+    passing: result !== undefined,
```
chenba, line 70: Comments need to be updated to mention this change.

chenba, line 70: You need to replace `falseFn` with a function that returns `undefined` because of this change.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -347,30 +374,63 @@ export const createSurveyFilter = (
...
+        signedInReliers: await signedInReliersCheck(
+          surveyConfig.conditions,
+          fetchDeviceList
+        ),
+        // Non-default profile image check
+        nonDefaultAvatar: await nonDefaultAvatarCheck(
+          surveyConfig.conditions,
+          fetchProfileImage
+        ),
+      };
```
chenba, line 409: This removes the short-circuiting of the original.  There's no need to run `nonDefaultAvatarCheck` if `languagesCheck` already failed to satisfy the language condition.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -347,30 +374,63 @@ export const createSurveyFilter = (
...
+      const satisfiedConditions = Object.keys(conditionChecks).reduce(
+        (output, key) => {
+          const condition = conditionChecks[key];
+          if (condition.value) {
+            output[key] = condition.value;
+          }
+          return output;
+        },
+        {}
+      );
```
chenba, line 427: Could skip this if `passing` is false.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
jodyheavener: Hm, good point. I feel like this is going to be tricky to reconfigure given our desire to retain the output and possibility of async functions 🤔

packages/fxa-content-server/app/scripts/lib/url.js
```diff
@@ -112,18 +131,16 @@ export default {
...
-        params.push(paramName + '=' + encodeURIComponent(paramValue));
-      }
-    }
+    const params = this._getObjPairs(obj)
+      .map(([[key0, ...keysRest], value]) => {
+        value = value.toString();
+        if (value.length) {
+          return `${key0}${keysRest
+            .map((a) => `[${a}]`)
+            .join('')}=${encodeURIComponent(value)}`;
```
chenba, line 140: I can't find any good authoritative SurveyGizmo documentation on how to pass a list of values, but from the comment section of an SurveyGizmo, I think it wants a comma separate values string.  If that's the case, this won't work.

Maybe @johngruen can help on the format question here.

jodyheavener: @chenba [this article](https://help.surveygizmo.com/help/url-variables) appears to indicate that regular query params are accepted. Or are you referring to something else?

jodyheavener: Oh are you talking about an array value? Yeah, I'm not sure either. I would prefer to keep it like this, but if SG doesn't like it I suppose I could modify the return values to join them before it reaches this point.

chenba: (whoa why did this comment thread just give me deja vu...)

If you go down to the very bottom of that page to check the first comment and the response following it, you see some discussion passing a list of values throuh the _same_ parameter.

I think @johngruen or @davismtl  wouldn't want a list of `something` showing up on SurveyGizmo as `something[0]` = `a`, `something[1]` = `b`, and `something[2]` = `a`, but rather `something` = `a,b,c`?

jodyheavener: Thanks.

jodyheavener: We also have at least one condition that returns an object 😢

jodyheavener: There is also a possibility of null being included in an array value (`reliersList: [ null, '9001' ]`), and because null is the desired value I assume this needs to be converted into an identifiable string; whether that's `"null"` or something else, I need direction.

jodyheavener:
This PR is on hold until @johngruen clears up some unknowns above.

packages/fxa-content-server/app/scripts/lib/url.js
johngruen: > I can't find any good authoritative SurveyGizmo documentation on how to pass a list of values, but from the comment section of an SurveyGizmo, I think it wants a comma separate values string. If that's the case, this won't work.
> 
> Maybe @johngruen can help on the format question here.

So, SG can do conditional filters on query strings, so i think a thing=a,b,c,d is fine here
image

jodyheavener: Sounds good. What about the null values?

johngruen: for null, could we just drop that from the list?

jodyheavener: Perfect!

chenba:
> 
> 
> > This was a funky one, but precedent dictates that null is also a valid value, at least when it comes to the relier condition.
> 
> IIRC null is sync. We can double check and maybe have a special case for that one 😢 (perhaps in a follow-up).

> 
> 
> for null, could we just drop that from the list?

For RPs, I think this won't be an issue once Sync switches from BrowserID to OAuth.

jodyheavener:
@meandavejustice Woo, glad to hear it’s good and working for you!

> I'd like to see the 5 or so PRs ... merged into this branch and make sure there aren't any conflicts 

Oh interesting - I was thinking the opposite, where we would merge this first and then the others one at a time into main so we’re not merging a single even-larger change into main. Which do you prefer?

jodyheavener:
Eh, now that I think about it I suppose the overall diff would be the same size since it’s mostly those two survey lib files.

So long as reviewing is complete on this PR I’ll merge everything together - don’t wanna make it even harder for reviews on an already large PR 😉

meandavejustice:
> Oh interesting - I was thinking the opposite, where we would merge this first and then the others one at a time into main so we’re not merging a single even-larger change into main. Which do you prefer?

I'm fine with that as well, I had figured that merging here was your plan since this was the target branch on all of those prs. I think either way is fine

PR #5817, opened 2020-06-30
feat(emails): Add onMobileDevice mailOption for post registration email

### LINKED ISSUE #5553: Follow up changes for post verify emali

LZoog:
Because:
* We want to include "connect on another desktop device" text with a link to download Firefox when users register with a tablet (which largely send desktop UA) or desktop.

This commit:
* Allows `onMobileDevice` to be passed into `message` as an option for the mailer and displays different copy for the link following the app store links.
* Replaces postVerify and appBadges images with CDN versions.

fixes #5553

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -1268,7 +1268,7 @@ module.exports = function (log, config, oauthdb) {
       email: message.email,
       uid: message.uid,
     });
-
+    const onDesktopOrTabletDevice = !message.onMobileDevice;
     const templateName = 'postVerify';
     const subject = gettext(
       'Account verified. Next, sync another device to finish setup'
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1268,7 +1268,7 @@ module.exports = function (log, config, oauthdb) {
       email: message.email,
       uid: message.uid,
     });
-
+    const onDesktopOrTabletDevice = !message.onMobileDevice;
```
LZoog, line 1271: I needed to reverse this here or else the `appBadges.html` `if`/`else` got messy to keep `cadReminders` rendering “connect another device” instead of “connect another desktop device.”

I didn’t pass this in as `onDesktopOrTabletDevice` because reversing the boolean in `postVerifyEmail` instead makes testing for the two different versions of copy much easier.

PR #5819, opened 2020-06-30
fix(l10n): polyfill Intl.NumberFormat for emails

chenba:
Because:
 - Node's Intl.NumberFormat doesn't format currencies correctly prior to v13

This commit:
 - polyfill Intl.NumberFormat if Node is < 13

Part of #FXA-2202

## Checklist

### DIFF
--- a/packages/fxa-auth-server/package.json
+++ b/packages/fxa-auth-server/package.json
@@ -69,6 +69,7 @@
     "hkdf": "0.0.2",
     "hot-shots": "^7.6.0",
     "i18n-abide": "0.0.26",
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/package.json
```diff
@@ -69,6 +69,7 @@
     "hkdf": "0.0.2",
     "hot-shots": "^7.6.0",
     "i18n-abide": "0.0.26",
+    "intl": "1.2.5",
```
lmorchard, line 72: Weird... this module was last released 4 years ago? Guess that just means it's stable?

PR #5824, opened 2020-07-01
fix(legal): change upload script to handle pdf

jrgm:
ref: https://jira.mozilla.com/browse/FXA-2184


jrgm:
r? - @bbangert pending this works for him

PR #5828, opened 2020-07-02
fix(content-server): create fake hidden input

### LINKED ISSUE #606: Account recovery: doorhanger displayed to update the password when tapping on "Cancel"

### LINKED ISSUE #932: Deleting an account causes Firefox "Save password" dropdown to display, without an email address

byrw:
## Because

* When deleting an account, after all required inputs is completed, the "Save Password" is displayed.

## This commit

* Create a fake input with "required" atribute to trick the browser into not showing the "Save Password" doorhanger.

* Remove autcomplete="off", this was added 5 years ago and now it does not hide the doorhanger to save the password.

fixes #932

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
@@ -77,9 +77,13 @@
       </ul>
 
       <div class="input-row password-row">
-        <!-- add the autocomplete=off attribute - the user is deleting the account, they most likely don't want the password saved. -->
-        <input type="password" class="password tooltip-above" id="password" placeholder="{{#t}}Password{{/t}}" pattern=".{8,}" required autofocus autocomplete="off" />
+        <input type="password" class="password tooltip-above" id="password" placeholder="{{#t}}Password{{/t}}" pattern=".{8,}" required autofocus />
       </div>
+
+      <!-- The browser wait until all the required inputs is fulfilled to
+         - show the "Save Password" doorhange. This non-fulfilled input tricks
+         - the browser into not showing the doorhanger. -->
+      <input class="hidden" required />
 
       <div class="button-row">
         <button type="submit" class="settings-button warning-button delete-account-button" disabled>{{#t}}Delete account{{/t}}</button>


packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
```diff
@@ -77,10 +77,12 @@
       </ul>
 
       <div class="input-row password-row">
-        <!-- add the autocomplete=off attribute - the user is deleting the account, they most likely don't want the password saved. -->
-        <input type="password" class="password tooltip-above" id="password" placeholder="{{#t}}Password{{/t}}" pattern=".{8,}" required autofocus autocomplete="off" />
+        <input type="password" class="password tooltip-above" id="password" placeholder="{{#t}}Password{{/t}}" pattern=".{8,}" required autofocus />
       </div>
 
+      <!-- fake input to hide "Save Password" doorhanger -->
+      <input class="hidden" required/>
```
jodyheavener, line 84: This appears to work, but I'm not sure why. Are you able to expand on why this works in the comment above it?

Also, small nit: could add an additional space before the `/` to make it consistent 😉

byrw: Hey @jodyheavener , I'm not sure if my comment match the actual behavior of the browser. But I think it's a good guess.

jodyheavener: What you wrote makes sense to me. At the very least we know that it's intentional, so future devs can work with it as needed.

jodyheavener:
Thanks @byrw!

PR #5847, opened 2020-07-02
Enable dependabot

clouserw:
## Because

* We want to keep our dependencies up to date

## This commit

* Enables Github's Dependabot support

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

### DIFF
--- a/dependabot.yml
+++ b/dependabot.yml
@@ -1,0 +1,26 @@
+version: 2
+updates:
+  # Configuration for npm
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2
+    reviewers:
+      - "@mozilla/fxa-devs"
+  - package-ecosystem: "npm"
+    directory: "/packages/fxa-content-server"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


clouserw:
I _think_ this will enable dependabot, but the way I'm reading [the docs](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#configuration-options-for-dependabotyml) is that we'll have to add a separate section for each package.  I added a separate one for `fxa-content-server` just to see if my interpretation is right.

I didn't see a global `open-pull-requests-limit` which means that even if we made it 1 for each section we might still get overwhelmed with PRs when an update comes out.

I also added docker in here since they mentioned it, but I'm not sure it would be super useful to be honest.  I suppose it wouldn't hurt.

Anyway.... r?

dannycoates:
dependabot doesn't yet support yarn 2. we can revive this once it does

https://github.com/dependabot/dependabot-core/issues/1297
https://github.com/dependabot/dependabot-core/issues/2030

clouserw:
We're going to try landing this.  We think it will still open the PRs and we'll have to manually update the lock file.  If it explodes, we'll turn it back off.

PR #5851, opened 2020-07-03
fix(attached-clients): Ignore a device's refreshTokenId if it's not a live token

rfk:
## Because

* OAuth refresh tokens live in a separate database from device records,
  for hysterial raisins.
* We've observed at least one instance where the device record appears to
  have a "dangling pointer" to a refresh token that no longer exists.
* We can't really do anything useful with a `refreshTokenId` that doesn't
  correspond to an actual refresh token, apart from "ignore it" or
  "delete it".

## This commit

* Silently ignores dangling refreshTokenId references when listing
  the clients attached to a user's account.
* Opts not to try to repair the device record, mostly because this
  is easier and it seems harmless to ignore.

Closes: https://bugzilla.mozilla.org/show_bug.cgi?id=1647126

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/attached-clients.js
+++ b/packages/fxa-auth-server/lib/routes/attached-clients.js
@@ -104,7 +104,9 @@ module.exports = (log, db, oauthdb, devices, clientUtils) => {
           const client = {
             ...defaultFields,
             sessionTokenId: device.sessionTokenId || null,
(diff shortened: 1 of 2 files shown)


rfk:
> Merging #5851 into main will decrease coverage by 13.75%.

Listen, I don't want to be rude here codecov, but are you sure about the math on this one..?

PR #5853, opened 2020-07-03
feat(surveys): always show surveys in dev, regardless of rate or last shown timestamp

jodyheavener:
## Because

We want to be able to see surveys in development

## This change

Ensures a survey will always show up in development, regardless of when you last saw it or the rate for which it should be shown (including a rate of 0). Note that all other conditions must still be met. To aid in development the matched conditions will also be console table-logged.

Closes: #5529, though the requirements changed to [this](https://github.com/mozilla/fxa/issues/5529#issuecomment-653156870).

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/survey-filter.js
+++ b/packages/fxa-content-server/app/scripts/lib/survey-filter.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import UserAgent from './user-agent';
+import { ENV_DEVELOPMENT } from './constants';
 
 // All but the default export here is to make testing easier.
 
@@ -30,7 +31,7 @@ export const participatedRecently = (
 };
 
 export const withinRate = (rate) => {
-  if (rate === 0) {
+  if (isNaN(rate) || rate === 0) {
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -402,11 +403,12 @@ export const createSurveyFilter = (
     if (
       !(
         surveyConfig &&
-        surveyConfig.rate &&
+        parseInt(surveyConfig.rate) > -1 &&
```
vbudhram, line 406: Did you intend to use `parseInt` instead of `withinRate`? Seems like you could reuse it.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
```diff
@@ -402,11 +403,12 @@ export const createSurveyFilter = (
...
         surveyConfig.conditions &&
-        withinRate(surveyConfig.rate) &&
+        (env === 'development' || withinRate(surveyConfig.rate)) &&
         Object.keys(surveyConfig.conditions).length > 0 &&
-        !participatedRecently(previousParticipationTime, doNotBotherSpan)
+        (env === 'development' ||
```
vbudhram, line 410: You can probably pull `development` into a constants file.

packages/fxa-content-server/app/scripts/lib/survey-filter.js
jodyheavener: I did - I wanted to ensure that the property existed, was numeric, and was 0 or more. _But_ I think this check could just be moved into `withinRate`. Good catch.

Issue #5686, opened 2020-06-17
[surveys] Pass env and service into survey through query params


data-sync-user:
For all surveys we should include the env (development, prod, etc.) the service name (content, payments, etc.) on the survey url's query parameters.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2130)
┆Issue Number: FXA-2130

data-sync-user:
➤ Barry Chen commented:

[~jgruen@mozilla.com] you mentioned passing env but I didn't see an issue for that, so here it is.

PR #5856, opened 2020-07-03
feat(surveys): send env and service name to survey gizmo as query params

jodyheavener:
## Because

There is a desire to know which environment and which server is calling the survey.

## This change

Modifies the initial condition query params to also include env and service. Now all surveys created in the Content Server will pass this data along to Survey Gizmo.

Notes:

- _"What about Payments?"_ I don't actually see anything set up beyond survey config in the payments server, so with how this is being implemented today it would be up to a future dev to ensure these params get added down the road.
- I could be convinced that, instead of appending these two params to the URL in the content server, both env and service should be [props](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/components/Survey/index.tsx#L11) on the Survey component. However my current belief is that there is a single `surveyURL` prop on that component, and that is the extent to which the component should be concerned about the URL, leaving it up to the individual service to supply accurately.

Closes: #5686

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/survey-targeter.js
+++ b/packages/fxa-content-server/app/scripts/lib/survey-targeter.js
@@ -2,6 +2,7 @@ import SurveyWrapperView from '../views/survey';
 import Storage from './storage';
 import createSurveyFilter from './survey-filter';
 import Url from './url';
+import { ENV_DEVELOPMENT, ENV_PRODUCTION } from './constants';
 
 const lastSurveyKey = 'lastSurvey';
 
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/lib/survey-targeter.js
```diff
@@ -69,16 +69,20 @@ export default class SurveyTargeter {
       }
 
       const selectedSurvey = this._selectSurvey(qualifiedSurveys);
+      const queryParamData = Object.assign(selectedSurvey.conditions, {
+        service: 'content',
```
vbudhram, line 73: I might be a little confused here, I took the bug as passing along the `service` from content-server url, ex. `http://localhost:3030/reset_password?service=sync` to SurveyGizmo.

vbudhram: Or by service you mean server?

jodyheavener: I did mean server, but used service to make it sound better since it's technically a publicly visible address -- happy to change to server to make everything look the same :)

jodyheavener:
Closed with https://github.com/mozilla/fxa/pull/5816

Issue #5843, opened 2020-07-02
Add prefers-reduced-motion media query to survey


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2227)
┆Issue Number: FXA-2227

PR #5857, opened 2020-07-03
feat(surveys): don't animate survey when prefers-reduced-motion is present

jodyheavener:
## Because

We want to respect the accessibility system preference to reduce on-screen motion.

## This change

Disables the survey's vertical transition when the media query `prefers-reduced-motion` is present.

**Reviewer**: if you'd like to test this on macOS you can enable the preference under `System Preferences > Accessibility > Display > Reduce motion`; on iOS this is under `Settings > Accessibility > Motion > Reduce Motion`. I don't know about other operating systems. 🤷

Closes: #5843

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_survey.scss
+++ b/packages/fxa-content-server/app/styles/modules/_survey.scss
@@ -119,3 +119,18 @@ $survey-height: 360px;
 .button-inner-enter-active {
   transform: rotate(0deg);
 }
+
+@media (prefers-reduced-motion) {
+  .survey-inner-exit-active,
+  .survey-inner-enter-active {
+    transition-duration: 0s;
+  }
+
+  .survey-component {
+    visibility: hidden;
+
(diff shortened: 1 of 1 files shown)


LZoog:
I copied and pasted your changes here into my branch which is why you're seeing an X instead of the chevron in this screen capture:

survey-animation


... I feel this could be improved slightly if we hide the survey until it's loaded so there's less appearance of jankiness, as with what you've got here it just goes from minimized to maximized. What about this instead, only show when maximized?

```
@media (prefers-reduced-motion) {
  .survey-inner-exit-active,
  .survey-inner-enter-active {
    transition-duration: 0s;
  }

  .survey-component {
    display: none;

    &.survey-inner-enter-done {
      display: block;
    }
  }
}
```

Looks like this:
survey-animation-2

jodyheavener:
@LZoog I can dig it! Though, in an effort to not much with layout I am using the `visibility` property over `display`.

PR #5865, opened 2020-07-06
feat(auth-server): add setup intent methods

### LINKED ISSUE #5744: Create auth-server API endpoints for updating a card
To allow cards to be updated using SCA, SetupIntents must be used. Endpoints will be added to the auth-server that will:

1. Create a SetupIntent for the given user and return the object for use by the client.
1. Update the invoice_settings for a new default payment method given the supplied payment method id

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2171)
┆Issue Number: FXA-2171

bbangert:
Because:

* We want to use the SetupIntent API to update the default payment
  method used for a customer.

This commit:

* Allows the client to create a SetupIntent API for a given customer
  and set the payment method id (after confirming it) as the customer's
  default payment method.

Closes #5744

## Checklist

### DIFF
--- a/packages/fxa-auth-server/docs/api.md
+++ b/packages/fxa-auth-server/docs/api.md
@@ -3560,6 +3560,31 @@ Update the user's default payment method using a payment token.
 :lock: authenticated with OAuth bearer token
 Returns customer details, including limited payment information.
 
+#### POST /oauth/subscriptions/customer
+
+:lock: authenticated with OAuth bearer token
+Create a new customer object for use with subscription payments.
+
+#### POST /oauth/subscriptions/active/new
+
+:lock: authenticated with OAuth bearer token
+Subscribe the user to a price using a payment method id.
+
+#### POST /oauth/subscriptions/invoice/retry
+
+:lock: authenticated with OAuth bearer token
+Retry an incomplete subscription invoice with a new payment method id.
+
+#### POST /oauth/subscriptions/setupintent/create
+
+:lock: authenticated with OAuth bearer token
+Create a new setup intent for attaching a new payment method to the user.
+
+#### POST /oauth/subscriptions/paymentmethod/default
+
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/docs/api.md
```diff
@@ -3560,6 +3560,31 @@ Update the user's default payment method using a payment token.
...
+
+#### POST /oauth/subscriptions/setupintent/create
+
+:lock: authenticated with OAuth bearer token
+Create a new setup intent for attaching a new payment method to the user.
+
+#### POST /oauth/subscriptions/paymentmethod/default
+
+:lock: authenticated with OAuth bearer token
+Update a users default payment method for invoices to the attached payment method id.
```
chenba, line 3586: "user's"

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -283,6 +284,62 @@ describe('StripeHelper', () => {
...
+          assert.equal(err, apiError);
+        }
+      );
+    });
+  });
+
+  describe('updateDefaultPaymentMethod', () => {
+    it('updates the default payment method', async () => {
+      const expected = deepCopy(newCustomerPM);
+      sandbox.stub(stripeHelper.stripe.customers, 'update').returns(expected);
```
chenba, line 318: s/returns/resolves/ here and various other spots.  I can file a PR for only that change throughout this file and subscriptions route tests.

bbangert: Indeed, this does work, but only out of an oddly happy accident. A substantial amount of other tests also are using returns here, which is where I originally copied it from. They should all be updated for resolves. Separate PR for that seems good as it will update more tests than just these.

PR #5868, opened 2020-07-06
feat(payments): New PaymentFormV2 component with Stripe upgrades

### LINKED ISSUE #4294: Upgrade payments-server to react-stripe-js from react-stripe-elements
It looks like [the @stripe/react-stripe v7.0.0 package has moved repositories and been renamed to @stripe/react-stripe-js v1.0.0](https://github.com/stripe/react-stripe-js/blob/master/CHANGELOG.md#v100-beta1-2019-12-20)

There are also some changes to usage patterns that now involve hooks and function-as-child. Seems cleaner but will be a bit of revision to make work with our payments-server app

https://github.com/stripe/react-stripe-js/blob/master/docs/migrating.md

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1211)
┆Issue Number: FXA-1211

lmorchard:
- new PaymentFormV2 component using new Stripe components, based on
  PaymentForm with tweaks for all-in-one card widget and shifts
  responsibility for card processing onto parent component

- plumbing in AppContext for new method for loading Stripe API

- add @stripe/react-stripe-js and @stripe/stripe-js dependencies

- switch to explicit tooltip parent wrapper for Stripe elements

fixes #4294


lmorchard:
Opening a draft until at least one green CI run.

Also, this looks huge, but PaymentFormV2 is a copy of PaymentForm with modifications. Did it this way to avoid a feature branch and so we could try deploying both new & old payment routes together. Eventually, V2 will get moved over on top of the old.

A decent way to see the changes between PaymentForm and PaymentFormV2 is with a directory diff:

* `diff src/components/PaymentForm src/components/PaymentFormV2`
* [Current diff as of this comment](https://gist.github.com/lmorchard/982c57787f318a2e31f2bb83ae5d3d82)

lmorchard:
Fell down a very deep rabbit hole with Intern & Leadfoot for content-server functional tests. Turns out that the specific number for the iframe representing each Stripe element becomes unstable when we load the Stripe API and elements in a different order. Took a long while to figure out how to try to make that more robust by selecting specific iframes within named fields.

packages/fxa-payments-server/src/components/PaymentFormV2/index.test.tsx
```diff
@@ -0,0 +1,516 @@
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import { render, cleanup, act, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import waitForExpect from 'wait-for-expect';
+
+import { Omit } from '../../lib/types';
```
chenba, line 7: Huh!  Could use TS's built in Omit instead?

lmorchard: This is copied over from the existing PaymentForm tests. I think at the time this was written, TS didn't have Omit yet. Can try dropping our definition of it though

packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
```diff
@@ -0,0 +1,372 @@
...
+} from '../../lib/validator';
+import { useCallbackOnce } from '../../lib/hooks';
+import {
+  getLocalizedCurrency,
+  getDefaultPaymentConfirmText,
+} from '../../lib/formats';
+import { AppContext } from '../../lib/AppContext';
+import { Plan } from '../../store/types';
+import {
+  DEFAULT_PRODUCT_DETAILS,
```
chenba, line 37: I don't think this is used anywhere.

lmorchard:
Rebased against main, small tweaks from review, will probably merge if CI goes green

Issue #5842, opened 2020-07-02
Change minimize to close


data-sync-user:
We should remove minimize functionality and replace it with closing the survey

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2226)
┆Issue Number: FXA-2226

Issue #5870, opened 2020-07-06
Allow Survey to import scss file (better content-server/fxa-react sharing)


data-sync-user:
Right now, styles for the Survey component live in `styles/modules/_survey.scss` in the `fxa-content-server`. While we eventually want all of FxA to use Tailwind, at the very least it would be ideal to have the file live in `fxa-react/components/Survey/index.scss` so we can `import './index.scss'` in the component. `fxa-react` Storybook needs a style file inside the `fxa-react` package unless we setup a content-server dependency. The styles do not import correctly in the content server and we can't use images from `fxa-react/images/`.

If we don't fix this, we have to maintain a separate yet identical style file for Storybook (besides content server variables) and we have `close.svg` in `Survey/images/close.svg`, used in the content-server styles, _and_ `fxa-react/images/close.svg`, used in the Storybook styles (and other components.

Once this is fixed we should be able to delete `styles/modules/_survey.scss` (and the import statement in `_modules`) and `Survey/images` after updating the "close" background image URL.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2237)
┆Issue Number: FXA-2237

PR #5871, opened 2020-07-06
feat(surveys): Replace minimize/maximize functionality with close

LZoog:
## Because

* Legal requested we change the functionality to close.

## This commit

* Changes the minimize/maximize 'survey-control' button with a 'survey-close' button, including a swap from chevron to close icon.
* Fixes Storybook styles.

Closes: #5842

## Checklist

## Other information

#5870 is a spin-off ticket of an importing issue I ran into while working on resolving this. I added some comments referencing that ticket and removed a related TODO comment.


packages/fxa-content-server/app/scripts/views/survey.jsx
```diff
@@ -34,7 +38,7 @@ const SurveyView = (props) => {
     return () => window.removeEventListener('message', handleIframeTask);
   }, [setShowSurvey, setSurveyComplete, setTimeout]);
 
-  return <>{showSurvey && <Survey {...{ surveyComplete, surveyURL }} />}</>;
+  return <>{showSurvey && <Survey {...{ surveyComplete, surveyURL, onSurveyClose }} />}</>;
```
jodyheavener, line 41: I see this isn't part of your changes, but... is the fragment necessary?

packages/fxa-content-server/app/styles/modules/_survey.scss
```diff
@@ -1,3 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// If you make changes here, be sure to update
+// `fxa-react/components/Survey/index.scss` for Storybook. See FXA-2237
```
jodyheavener, line 6: thank you for adding this

packages/fxa-content-server/app/scripts/views/survey.jsx
jodyheavener: I see this a few times below as well, I think you can get away with just `return (condition && <Component />)`, no?

LZoog: Looks like that's the equivalent of `return false` when `showSurvey` is `false` - no complains in the JSX file, but Storybook has to have a React element returned. I can make that change here, will wait on other comments. 👍

jodyheavener: Ah okay as long as it's intentional! Carry on.

PR #5872, opened 2020-07-06
feat(auth-server): add stripe object filtering

### LINKED ISSUE #5849: Create/apply Stripe object filter for server-side data
Stripe objects contain a variety of data not needed by the client-side, we should run the return objects through a filter function that picks out only the needed properties.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2230)
┆Issue Number: FXA-2230

bbangert:
Because:

* We don't want to return sensitive data from Stripe objects.

This commit:

* Filters all attributes except for the ones we know we need for
  the client.

Closes #5849

## Checklist

Note that #5865 should be merged first.


packages/fxa-shared/subscriptions/stripe.ts
```diff
@@ -0,0 +1,136 @@
+import Stripe from 'stripe';
```
lmorchard, line 1: Could use some tests for these methods. Looks like they're *used* in tests, but aren't themselves tested.

PR #5882, opened 2020-07-07
AET auth db updates

vbudhram:
@6a68 This is just a draft to get tests etc passing, lmk what you think.

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -126,6 +126,10 @@ function createServer(db) {
   );
   api.post('/account/:id/locale', withIdAndBody(db.updateLocale));
   api.get('/account/:id/sessions', withIdAndBody(db.sessions));
+  api.put(
+    '/account/:id/ecosystemAnonId',
+    withIdAndBody(db.updateEcosystemAnonId)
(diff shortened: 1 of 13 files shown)


6a68:
LGTM so far, thanks @vbudhram!

6a68:
I've merged this commit into my branch for #5689, I'll see if I can get CI tests to pass over there.

packages/fxa-auth-db-mysql/db-server/index.js
```diff
@@ -126,6 +126,10 @@ function createServer(db) {
   );
   api.post('/account/:id/locale', withIdAndBody(db.updateLocale));
   api.get('/account/:id/sessions', withIdAndBody(db.sessions));
+  api.put(
```
jodyheavener, line 129: I think this should be a PATCH instead of PUT given we're only updating a single row, no?

vbudhram:
Closing this in favor of https://github.com/mozilla/fxa/pull/5689

packages/fxa-auth-db-mysql/db-server/index.js
vbudhram: I have to double check if restify supports this, I think technically either is fine.

Issue #5886, opened 2020-07-08
Response validation errors in /invoke_command with train-178


rfk:
Train-178 is causing `/invoke_command` to fail with validation errors like `ValidationError: "enqueued" is not allowed. "notified" is not allowed. "notifyError" is not allowed`.  This is an error introduced in https://github.com/mozilla/fxa/pull/5546. PR incoming.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2248)
┆Issue Number: FXA-2248

PR #5887, opened 2020-07-08
fix(devices): Fix response validation for /invoke_command

rfk:
## Because

- When I added metrics to the /invoke_command route, I modified it
  to return the metrics info to the client without really thinking
  too hard about the implications.
- When we tried to deploy this in train-178, the result was response
  validation errors in production.

## This commit

- Adjusts the response schema for /invoke_command to accommodate the
  new fields.
- Updates the device route tests to validate the response, in order
  to help avoid similar oversights in future.
- Fixes a bunch of unrelated mock values in the device route tests
  that do not pass validation.

Closes: #5886.

## Checklist

This PR targets the `train-178` release branch, for a point-release.

### DIFF
--- a/packages/fxa-auth-server/lib/push.js
+++ b/packages/fxa-auth-server/lib/push.js
@@ -473,6 +473,7 @@ module.exports = function (log, db, config, statsd) {
       } else {
         errCode = ERR_PUSH_UNKNOWN;
       }
+      err.errCode = errCode;
       metricsTags = { errCode, err, ...metricsTags };
       this.incrementPushMetric(LOG_OP_PUSH_SEND_FAILURE, metricsTags);
       log.warn(LOG_OP_PUSH_SEND_FAILURE, metricsTags);
--- a/packages/fxa-auth-server/lib/routes/devices-and-sessions.js
+++ b/packages/fxa-auth-server/lib/routes/devices-and-sessions.js
@@ -270,7 +270,11 @@ module.exports = (
           },
         },
         response: {
-          schema: {},
+          schema: {
+            enqueued: isA.boolean().optional(),
+            notified: isA.boolean().optional(),
+            notifyError: isA.string().optional(),
+          },
         },
       },
       handler: async function (request) {
@@ -351,8 +355,10 @@ module.exports = (
 
         return {
           enqueued: true,
-          notified: !!notifyError,
-          notifyError: notifyError,
+          notified: !notifyError,
+          notifyError: notifyError
+            ? notifyError.errCode || 'unknown'
+            : undefined,
         };
       },
     },
--- a/packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
+++ b/packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
@@ -7,6 +7,7 @@
 const sinon = require('sinon');
 const assert = { ...sinon.assert, ...require('chai').assert };
 const crypto = require('crypto');
+const Joi = require('@hapi/joi');
 const error = require('../../../lib/error');
 const getRoute = require('../../routes_helpers').getRoute;
 const isA = require('@hapi/joi');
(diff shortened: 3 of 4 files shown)


packages/fxa-auth-server/lib/push.js
```diff
@@ -473,6 +473,7 @@ module.exports = function (log, db, config, statsd) {
       } else {
         errCode = ERR_PUSH_UNKNOWN;
       }
+      err.errCode = errCode;
```
rfk, line 476: This is to help ensure that we return one of a fixed set of strings to the client, rather than an arbitrary `Error` instance.

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -351,8 +355,10 @@ module.exports = (
 
         return {
           enqueued: true,
-          notified: !!notifyError,
-          notifyError: notifyError,
+          notified: !notifyError,
```
rfk, line 358: This was just straight-up wrong with the `!!`; thank you, additional test coverage!

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -270,7 +270,11 @@ module.exports = (
           },
         },
         response: {
-          schema: {},
+          schema: {
+            enqueued: isA.boolean().optional(),
```
vladikoff, line 274: 👍

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -72,8 +73,23 @@ function makeRoutes(options = {}, requireMocks) {
...
 }
 
-function runTest(route, request, onSuccess, onError) {
-  return route.handler(request).then(onSuccess, onError).catch(onError);
+async function runTest(route, request, onSuccess, onError) {
+  try {
+    const response = await route.handler(request);
+    if (route.options.response.schema) {
+      await Joi.validate(response, route.options.response.schema);
+    }
```
rfk, line 81: This ensures that we run response validation on each test, which is something we do in a handful of other route test suites. There's a nice refactor in here somewhere to extract this into a shared helper, but not for this PR.

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -858,8 +877,22 @@ describe('/account/device/commands', () => {
...
+          },
+        },
+        {
+          index: 4,
+          data: {
+            sender: '88888888888888888888888888888888',
+            command: 'four',
+            payload: {},
+          },
+        },
```
rfk, line 895: Most of these changes are for mock data that didn't fit the response schema of its route.

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
```diff
@@ -270,7 +270,11 @@ module.exports = (
...
+            notified: isA.boolean().optional(),
+            notifyError: isA.string().optional(),
+          },
```
rfk, line 277: Here's the actual fix; I made them all optional since they're only an FYI for the client to include in e.g. detailed log files.

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
```diff
@@ -1230,7 +1263,10 @@ describe('/account/devices/invoke_command', () => {
       '/account/devices/invoke_command'
     );
 
-    return runTest(route, mockRequest).then(() => {
+    return runTest(route, mockRequest).then((response) => {
+      assert.ok(response.enqueued);
+      assert.ok(response.notified);
+      assert.isUndefined(response.notifyError);
```
rfk, line 1269: If we're going to return this data, let's test that we actually return the data we want...

packages/fxa-auth-server/test/mocks.js
```diff
@@ -639,7 +639,7 @@ function mockDevices(data, errors) {
       return data;
     }),
     synthesizeName: sinon.spy(() => {
-      return data.deviceName || null;
+      return data.deviceName || '';
```
rfk, line 642: The real `synthesizeName` never returns `null`, this makes the mock match.

packages/fxa-auth-server/lib/routes/devices-and-sessions.js
vladikoff: Hah I remember seeing this but for some reason thought you wanted `!!notifyError`🤦 Sorry about not commenting on that!

packages/fxa-auth-server/test/local/routes/devices-and-sessions.js
vbudhram: I imagine we will probably catch a few more of these response validation errors if we pull into a shared helper.

PR #5888, opened 2020-07-08
fix(docs): Clarify that JWT `typ` header is case-insensitive.

rfk:
## Because

* The docs for JWT OAuth tokens tell RPs check for for "at+jwt".
* The actual code for JWT OAuth tokens uses a value of "at+JWT".
* The relevant RFC says this value should be compared case-insensitively,
  and should be treated as if it has a magic "application/" prefix
  prepended to it (ref https://tools.ietf.org/html/rfc7515#section-4.1.9).

## This commit

* Rewords the docs on JWT OAuth tokens to clarify these requirements.
* Updates our own JWT-verification logic to implement them.

~~Closes: # (issue number)~~

~~It would have taken me longer to file the issue than to make this one-line PR...~~

OK sorry, turns out this was more than a one-line PR. But still, here we are...

## Checklist

- ~~[ ] If applicable, I have modified or added tests which pass locally.~~

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/jwt.js
+++ b/packages/fxa-auth-server/lib/oauth/jwt.js
@@ -36,8 +36,10 @@ exports.sign = function sign(claims, options) {
 
 exports.verify = async function verify(jwt, options = {}) {
   const getKey = (header, callback) => {
-    if (options.typ && header.typ !== options.typ) {
-      return callback(new Error('Invalid typ'));
+    if (options.typ) {
+      if (normalizeTyp(options.typ) !== normalizeTyp(header.typ)) {
+        return callback(new Error('Invalid typ'));
+      }
     }
 
     let signingKey;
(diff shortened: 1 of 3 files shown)


rfk:
(Update: turns out our own internal logic for checking the `typ` header was not following the recommended rules, so this is now more than a docs-only change)

packages/fxa-auth-server/lib/oauth/jwt.js
```diff
@@ -58,3 +60,14 @@ exports.verify = async function verify(jwt, options = {}) {
     ignoreExpiration: options.ignoreExpiration,
   });
 };
+
+function normalizeTyp(typ) {
+  // Ref https://tools.ietf.org/html/rfc7515#section-4.1.9 for the rules.
+  if (typ) {
+    typ = typ.toLowerCase();
+    if (typ.indexOf('/') === -1) {
```
chenba, line 68: I suppose `includes` reads a little better here.

rfk: Oh yes, thanks! My javascript is often stuck in the late 90s 😆

Issue #5897, opened 2020-07-08
[auth server] Update stubs in StripeHelper and subscriptions route tests to return promises where appropriate


data-sync-user:
In the auth-server's tests for StripeHelper and the subscriptions routes, some (most?) of the stubs for `async` function simply returns synchronously.  They should return a promise.  At the very least, that'd help to catch call sites that don't `await`.

(Not part of any particular EPIC)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2253)
┆Issue Number: FXA-2253

PR #5900, opened 2020-07-08
fix(tests): replace returns with resolves on stubs

chenba:
Because:
 - stubs in tests for async functions should return a promise

This commit:
 - replace `returns` with `resolves` on stubs in StripeHelper and
   subscriptions route tests
 - fix a bug found during the test updates

Closes: #5897 (FXA-2253)

## Checklist


.circleci/config.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - image: memcached
       - image: pafortin/goaws
       - image: circleci/mysql:5.7.27
+    resource_class: medium+
```
chenba, line 74: Not really part of the PR.  I will remove.

Issue #5906, opened 2020-07-09
Correctly handle Stripe Element Styles


data-sync-user:
I discovered this while trying to tweak some form styles for the new payments v2 page, but it looks like we are not actually reading any of the stye configuration passed into the CardElement object. Styles are passed in as a field of an `options` prop and not as an independent `style` prop. Additionally, given the v2 implementation we should adjust a few details like font size and form element padding.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2260)
┆Issue Number: FXA-2260

PR #5885, opened 2020-07-08
In-progress reworking of product subscription page for new payment flows

lmorchard:
Starting a draft of this PR for issue #5745. Lots of work left to do, but worth starting to take a look at.

Following along with [Stripe's fixed-price subscription example](https://stripe.com/docs/billing/subscriptions/fixed-price), we need to handle a small flowchart of process here.

It's growing into a significant re-write of the product subscription page. And, some of what's here might end up useful to reuse for using the endpoints from #5744 in updating payment method.

PR #5907, opened 2020-07-09
bug(styles): correctly pass styles into stripe card element

johngruen:
## Because

We do not correctly handle adding styles to Stripe's card object

## This commit

Resolves that issue by adding an options property to our StripeElement component. I've also made small adjustments to the layout of the input for the paymentsV2 form.

Closes: #5906

## Checklist

## Screenshots (Optional)

image

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
@@ -122,10 +122,21 @@ export const PaymentForm = ({
     [validator, onSubmitForParent, stripe, submitNonce, shouldAllowSubmit]
   );
 
-  const { matchMedia, navigatorLanguages } = useContext(AppContext);
-  const stripeElementStyles = mkStripeElementStyles(
-    matchMedia(SMALL_DEVICE_RULE)
-  );
+  const { navigatorLanguages } = useContext(AppContext);
+
+  const STRIPE_ELEMENT_STYLES = {
+    style: {
+      base: {
+        fontFamily:
+          'system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif',
+        fontSize: '16px',
+        fontWeight: '500',
+      },
+      invalid: {
+        color: '#0c0c0d',
+      },
+    },
+  };
 
   let termsOfServiceURL, privacyNoticeURL;
   if (confirm && plan) {
@@ -164,8 +175,8 @@ export const PaymentForm = ({
           component={CardElement}
           name="creditCard"
           label="Your card"
-          style={stripeElementStyles}
           className="input-row input-row--xl"
+          options={STRIPE_ELEMENT_STYLES}
           getString={getString}
           required
         />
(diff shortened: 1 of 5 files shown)


packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
```diff
@@ -122,10 +122,21 @@ export const PaymentForm = ({
     [validator, onSubmitForParent, stripe, submitNonce, shouldAllowSubmit]
   );
 
-  const { matchMedia, navigatorLanguages } = useContext(AppContext);
-  const stripeElementStyles = mkStripeElementStyles(
-    matchMedia(SMALL_DEVICE_RULE)
-  );
+  const { navigatorLanguages } = useContext(AppContext);
+
+  const STRIPE_ELEMENT_STYLES = {
```
johngruen, line 127: We don't actually need any kind of media type logic here (we can just uniformly set these rules with the new form.

packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
```diff
@@ -164,8 +175,8 @@ export const PaymentForm = ({
           component={CardElement}
           name="creditCard"
           label="Your card"
-          style={stripeElementStyles}
           className="input-row input-row--xl"
+          options={STRIPE_ELEMENT_STYLES}
```
johngruen, line 179: card elements don't take a style prop, but an options prop that has styles as a field in it...
https://github.com/stripe/react-stripe-js/blob/master/examples/class-components/0-Card-Minimal.js#L44

lmorchard: Oh, I think there's another problem here: While the new react-stripe-js elements take options.style, the old react-stripe-elements currently in use *do* take a style prop (and *not* options):

https://github.com/stripe/react-stripe-elements#using-individual-element-components

So, we might need to tweak this somehow to use the old style for the old components currently in use on the site - until/unless we launch with the new library.

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -47,6 +47,7 @@ type FieldProps = {
   maxLength?: number;
   minLength?: number;
   autoFocus?: boolean;
+  options?: object;
```
lmorchard, line 50: I don't think this needs to be added here - options isn't used within the Field component

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -262,11 +263,11 @@ export const StripeElement = (props: StripeElementProps) => {
     className,
     autoFocus,
     getString,
+    options = {},
```
lmorchard, line 266: Don't really need to add this here, either. If options is just passed through to the StripeElementComponent and not used within the wrapper here, it'll just end up in childProps

packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
lmorchard: Actually, disregard that... these are only changes to PaymentFormV2 and don't affect the old PaymentForm, so they're already separated between old & new Stripe elemenets

lmorchard:
Think I'll just merge this and get it folded into PR #5885

packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
johngruen: yeah i figured it was not worth the effort to sweat the current UI

Issue #5883, opened 2020-07-07
Update Pairing Page to Support iOS and Link to Android Beta (instead of nightly)


data-sync-user:
This is debatably a bug but certainly needs to be improved.

Steps to reproduce: 

* Visit https://accounts.firefox.com/pair/
* Click "Get App"

Expected Results:

* Options to download iOS or Android App
* Not nightly (at least Beta Android would be better)

Actual results:

* There's just a link to Android Nightly despite the iOS supporting pairing.

Proposal:

* Use app store badges
* Link android to: https://play.google.com/store/apps/details?id=org.mozilla.firefox_beta

CC [~vfilippov@mozilla.com] who reported this in Slack

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2246)
┆Issue Number: FXA-2246

vladikoff:
> Use app store badges

Yea!

> Link android to: https://play.google.com/store/apps/details?id=org.mozilla.firefox_beta

and when the main app updates, we should also not forget and update that !

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.180.0] under Windows 10 64-bit. App store badges for android and ios are successfully displayed and they point to: 
- [Firefox for Android Beta](https://play.google.com/store/apps/details?id=org.mozilla.firefox_beta&referrer=adjust_reftag%3DctSXSK5JPHYSz%26utm_source%3DProduct%2BMarketing%2B%2528Owned%2Bmedia%2529%26utm_campaign%3Dfxa-conf-page%26utm_content%3Dandroid%26utm_term%3Dbutton-autumn-2016-connect-another-device)
- [Firefox for iOS](https://apps.apple.com/app/apple-store/id989804926?pt=373246)

![Screenshot_13](https://user-images.githubusercontent.com/18114210/87790748-c31f5600-c849-11ea-8d7e-16f407f4b280.png)

PR #5916, opened 2020-07-12
fix(pair): Add marketings links (app badges) to pairing page

vbudhram:
## Because

* The pairing page needs to have more than links to install Android

## This commit

* Adds our marketing mixin which gives us App store badges for android and ios

Closes: #5883 

## Checklist

- [ ] ~I have added necessary documentation (if appropriate).~

## Screenshots

Screen Shot 2020-07-12 at 12 25 55 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/constants.js
+++ b/packages/fxa-content-server/app/scripts/lib/constants.js
@@ -114,11 +114,11 @@ module.exports = {
   MARKETING_ID_AUTUMN_2016: 'autumn-2016-connect-another-device',
 
   DOWNLOAD_LINK_TEMPLATE_ANDROID:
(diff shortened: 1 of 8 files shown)


jodyheavener:
🤔 I feel like we recently closed (but now I can't find) an issue about incorrect app store redirecting, because it had to do with the platform you were using to access the Adjust link. Maybe this fixes that.

clouserw:
It's not closed: https://bugzilla.mozilla.org/show_bug.cgi?id=1650634

jodyheavener:
Ah, I was looking in the wrong spot!

vbudhram:
@vladikoff made updates, this will link to the correct app store when going to pairing page.

Issue #5874, opened 2020-07-07
[fxa-profile-server] Uploading images fails with Istio/Envoy


enote-kane:
## Description

When setting up the FxA stack inside a Kubernetes cluster and using Istio as a service mesh with Envoy, avatar upload requests that are passed through to the profile server to the actual internal worker service for processing will fail with a HTTP `500`.

### Steps to reproduce

Set up the FxA stack in Kubernetes where Istio and Envoy are properly configured.

### Expected result

Avatar upload succeeds.

### Actual result

Avatar upload fails with HTTP `500` and errno `103`.

### Environment

_not applicable_

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2240)
┆Issue Number: FXA-2240

enote-kane:
Although I know that Kubernetes is not a major target for running FxA, I could provide a pull-request, which I've already verified at our side.

data-sync-user:
➤ Wil Clouser commented:

Hi Kane.  We don't support Kubernetes so it depends on what the PR is for and what it would do.  If it means more maintenance or testing on our side we probably wouldn't land it.

enote-kane:
OK, it is basically about the fact that the HTTP request to the worker gets all HTTP headers from the original request, including the `Host` header, which is against the spec ([RFC 7230, 5.4. Host](https://tools.ietf.org/html/rfc7230#section-5.4)):
> [...] When a proxy receives a request with an absolute-form of request-target, the proxy MUST ignore the received Host header field (if any) and instead replace it with the host information of the request-target. [...]

The PR I would create adds a configurable array of HTTP headers to ignore/exclude to the profile server. defaulting to `['host']`, but other headers may be needed in other scenarios.

clouserw:
If an empty list makes it do nothing (meaning, the functionality is the same as today) I think it sounds like a good patch.

enote-kane:
Yes, explicitly unconfiguring it with `null` or an empty array (or even some value that doesn't match any header, practically) will result in the current behavior.

However, again, the current behavior is - strictly speaking - violating the spec.

enote-kane:
@clouserw 
I've just added PR #5917 to address this issue in a flexible and non-breaking way. It would be great if someone could havea look at it.

PR #5917, opened 2020-07-13
[fxa-profile-server] Add block-/allow-lists for passed-through HTTP headers

enote-kane:
## Because

HTTP headers are re-used as is from the original incoming request to the profile server when uploading an avatar to the worker service, it also includes headers that might confuse the environment the services are running in.

Most notably, the `Host` header is re-used and the `request` module does not rewrite that for the worker request.

## This commit

Adds the following configuration keys:
* array `worker.headers_exclude` (env variable `WORKER_HEADERS_EXCLUDE`) for blocking headers from passing through to the worker request
* array `worker.headers_include` (env variable `WORKER_HEADERS_INCLUDE`) for passing only the specified headers to the worker request

Closes: #5874
Refs: #FXA-2240

## Checklist

The new default configuration removes the `Host` header by default, which is functionally not required but confuses some software routing environments.

### DIFF
--- a/packages/fxa-profile-server/lib/img-workers.js
+++ b/packages/fxa-profile-server/lib/img-workers.js
@@ -15,22 +15,30 @@ const WORKER_URL = config.get('worker.url');
 const ACCEPT_ENCODING_HEADER = 'identity';
 
 exports.upload = function upload(id, payload, headers) {
-  return new P(function (resolve, reject) {
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
+
+    // filter incoming headers
+    const header_excludes = config.get('worker.headers_exclude');
+    const worker_headers = {...headers};
+    for (const exclude of header_excludes) {
+      delete worker_headers[exclude.toLowerCase()];
+    }
+    logger.debug('upload.headers', worker_headers);
 
     // do not compress the response to the local worker
     // https://www.exratione.com/2014/07/nodejs-handling-uncertain-http-response-compression/
-    headers['Accept-Encoding'] = ACCEPT_ENCODING_HEADER;
+    worker_headers['accept-encoding'] = ACCEPT_ENCODING_HEADER;
 
     var opts = {
-      headers: headers,
+      headers: worker_headers,
       // disable gzip in request: https://github.com/request/request#requestoptions-callback
       gzip: false,
       json: true,
     };
     logger.verbose('upload', url);
     payload.pipe(
-      request.post(url, opts, function (err, res, body) {
+      request.post(url, opts, function(err, res, body) {
         var nestedError = null;
         if (body && body[0] && body[0].error) {
           nestedError = body[0].error;
@@ -51,7 +59,7 @@ exports.upload = function upload(id, payload, headers) {
         resolve(body);
       })
     );
-    payload.on('error', function (err) {
+    payload.on('error', function(err) {
       logger.error('upload.payload.error', err);
       reject(err);
     });
@@ -59,11 +67,11 @@ exports.upload = function upload(id, payload, headers) {
 };
 
 exports.delete = function deleteAvatar(id) {
-  return new P(function (resolve, reject) {
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
     var opts = { method: 'delete', json: true };
     logger.verbose('delete', url);
-    request(url, opts, function (err, res, body) {
+    request(url, opts, function(err, res, body) {
       if (err) {
         logger.error('delete.network.error', err);
         return reject(AppError.processingError(err));
--- a/packages/fxa-profile-server/lib/config.js
+++ b/packages/fxa-profile-server/lib/config.js
@@ -240,6 +240,12 @@ const conf = convict({
       default: 'http://localhost:1113',
       env: 'WORKER_URL',
     },
+    headers_exclude: {
+      doc: 'HTTP headers to not pass through to the worker requests (in lower case)',
+      format: Array,
+      env: 'WORKER_HEADERS_EXCLUDE',
+      default: ['host'],
+    },
   },
   authServerMessaging: {
     region: {
--- a/packages/fxa-profile-server/test/api.js
+++ b/packages/fxa-profile-server/test/api.js
@@ -1044,6 +1044,125 @@ describe('api', function () {
       });
     });
 
+    describe('upload-header-excludes', function() {
+      let origHeaderExcludes = null;
+      before(function() {
+        origHeaderExcludes = config.get('worker.headers_exclude');
+        config.set('worker.headers_exclude', ['host', 'x-my-header']);
+      });
+      after(function() {
+        config.set('worker.headers_exclude', origHeaderExcludes);
+      });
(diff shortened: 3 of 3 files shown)


packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,47 @@ const WORKER_URL = config.get('worker.url');
 const ACCEPT_ENCODING_HEADER = 'identity';
 
 exports.upload = function upload(id, payload, headers) {
-  return new P(function (resolve, reject) {
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
 
+    var header_excludes = config.get('worker.headers_exclude');
+    var header_includes = config.get('worker.headers_include');
```
chenba, line 22: Let's use `const` instead of `var` here.

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,47 @@ const WORKER_URL = config.get('worker.url');
...
+
+    worker_headers = {};
```
chenba, line 24: `worker_headers` isn't declared before this assignment.  `let worker_headers = {};`?

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,47 @@ const WORKER_URL = config.get('worker.url');
...
+    if (header_includes.length) {
+      // do not pass-though any HTTP headers except these
+      for (header in headers) {
```
chenba, line 27: `for (const header of headers)` here.

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,47 @@ const WORKER_URL = config.get('worker.url');
...
+        if (header_includes.includes(header.toLowerCase())) {
+          worker_headers[header] = headers[header];
+        }
+      }
+    } else if (header_excludes.length) {
+      // pass-through all headers except these
+      for (header in headers) {
```
chenba, line 34: `for (const .... of ...)`

enote-kane: Hm:
```
TypeError: headers is not iterable at [...]
```

chenba: Wow, sorry, I thought those were the configured array lists.  My fault for not reading carefully. :(  `for...in` is correct here.

enote-kane: ...but `for (const ... in ...)` works :) I am pretty new to node.js

packages/fxa-profile-server/lib/img-workers.js
enote-kane: done

packages/fxa-profile-server/lib/img-workers.js
enote-kane: dropped

packages/fxa-profile-server/lib/img-workers.js
enote-kane: done

enote-kane:
I agree. Initially I thought there might be situations where people want to allow only the necessary headers, but then that might break things (e.g. request tracing, ...).

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,39 @@ const WORKER_URL = config.get('worker.url');
...
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
 
+    const header_excludes = config.get('worker.headers_exclude');
+
+    let worker_headers = {};
+    if (header_excludes.length) {
+      // pass-through all headers except these
+      for (const header in headers) {
+        if (header_excludes.includes(header.toLowerCase())) {
```
chenba, line 27: style nit: I suppose we could save a line here if we check for `if (!header_excludes.includes(header.toLowerCase()))` and then copy the header inside the if block.

enote-kane: OK, did some thinking and actually was able to removed some more lines with the new version.

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,32 @@ const WORKER_URL = config.get('worker.url');
 const ACCEPT_ENCODING_HEADER = 'identity';
 
 exports.upload = function upload(id, payload, headers) {
-  return new P(function (resolve, reject) {
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
 
+    const header_excludes = config.get('worker.headers_exclude');
+    const worker_headers = Object.assign({}, headers);
```
chenba, line 22: The newfangled way of copying is `{...headers}` 😄

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,32 @@ const WORKER_URL = config.get('worker.url');
...
+    if (header_excludes.length) {
+      // filter incoming headers
+      for (const exclude of header_excludes) {
+        delete worker_headers[exclude];
```
chenba, line 26: I do like this approach of removing the unwanted headers.  However, we lose the case insensitivity comparisons from earlier; I think we should keep the case insensitivity.

packages/fxa-profile-server/lib/img-workers.js
enote-kane: ...still learning here... :D

packages/fxa-profile-server/lib/img-workers.js
enote-kane: Dammit...

packages/fxa-profile-server/lib/img-workers.js
enote-kane: done.

packages/fxa-profile-server/lib/img-workers.js
enote-kane: done.

packages/fxa-profile-server/lib/img-workers.js
```diff
@@ -15,22 +15,32 @@ const WORKER_URL = config.get('worker.url');
 const ACCEPT_ENCODING_HEADER = 'identity';
 
 exports.upload = function upload(id, payload, headers) {
-  return new P(function (resolve, reject) {
+  return new P(function(resolve, reject) {
     var url = WORKER_URL + '/a/' + hex(id);
 
+    const header_excludes = config.get('worker.headers_exclude');
+    const worker_headers = {...headers};
+    if (header_excludes.length) {
```
chenba, line 23: This all looks good, and I can merge as it is.  But in the name of even fewer lines of code, we don't need this `if` since `for...of` can handle an empty array gracefully.

Let me know if I should merge as it is, or if you push a change, I'll merge after that.  Thanks!

enote-kane: OK, removed the `if` here.

enote-kane:
I think that's it now, finally.

chenba:
Merged.  Thanks again for the patch.

PR #5922, opened 2020-07-13
fix(tests): update redis timeout

vbudhram:
## Because

* I've seen `[ioredis] Unhandled error event: Error: connect ETIMEDOUT` more frequently in circle and locally. I'm not really sure why it is happening more (or maybe I didn't notice it before?)

## This commit

* Updates default timeout from 10s to 1minute

Closes: NA

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -402,6 +402,12 @@ const conf = convict({
       format: 'port',
       doc: 'Port for Redis server',
     },
+    connectTimeout: {
+      default: '1 minute',
+      env: 'REDIS_CONNECT_TIMEOUT',
+      format: 'duration',
+      doc: 'Connect timeout for Redis server',
+    },
     accessTokens: {
       host: {
         default: 'localhost',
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -402,6 +402,12 @@ const conf = convict({
       format: 'port',
       doc: 'Port for Redis server',
     },
+    connectTimeout: {
```
vbudhram, line 405: @dannycoates Is this something worthwhile? The CI timeout errors looks like this https://app.circleci.com/pipelines/github/mozilla/fxa/12046/workflows/81796a6d-794e-447f-8595-a490df59a73e/jobs/181342/steps

dannycoates: that's a mysql timeout

vbudhram: Hrm, that one is indeed mysql. Let me see if I can reproduce the error I got locally with the error stack.

Do the oauth server tests pass for you locally?

vbudhram:
Closing here, my problem was that I was running redis 4.0 which needed the `replicate_commands` in lau scripts. New versions do this automatically.

Issue #5912, opened 2020-07-09
Extract TypeScript types shared between payments-server and auth-server


data-sync-user:
While walking through payment-server code, we noticed some pretty general types used that are potentially sharable with other parts of the system:

https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/store/types.tsx#L15

Many of these types are also represented on auth-server, if only as validators or imports from Stripe:

* https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/payments/stripe.js#L11
* https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/validators.js#L341

Could be a nice improvement to extract some of these into shared types under fxa-shared.

Could also be nice to just use types directly from the Stripe definitions. A wrinkle there is that we have some legacy subhub-shaped data schemas, but we could have an opportunity to reshape those too.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2266)
┆Issue Number: FXA-2266

PR #5926, opened 2020-07-13
Fxa 2215 use setup intent api

### LINKED ISSUE #5832: Use auth-server SetupIntent API when updating card information
To properly handle 3D Secure and SCA, we need to use the SetupIntent API with off_session set to true. This will properly trigger authentication and give us the required permissions to charge a card later. The client should be setup to request a new SetupIntent from the server, then use confirmCard which will confirm and add it to the customer on success.

See:

https://stripe.com/docs/billing/migration/strong-customer-authentication#scenario-2

https://stripe.com/docs/payments/save-and-reuse#web-create-setup-intent

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2215)
┆Issue Number: FXA-2215

jackiemunroe:
## Because:
* To support SCA we must migrate to using Stripe's SetupIntent API
    
## This commit:
    
* Add call to auth server to create SetupIntent
* Use secret from SetupIntent result to confirm card prior to creating a subscription
    
## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-payments-server/src/lib/apiClient.ts
+++ b/packages/fxa-payments-server/src/lib/apiClient.ts
@@ -1,5 +1,12 @@
 import { Config, defaultConfig } from './config';
-import { Plan, Profile, Customer, Subscription, Token } from '../store/types';
+import {
+  Plan,
+  Profile,
+  Customer,
+  Subscription,
+  Token,
+  FilteredSetupIntent,
+} from '../store/types';
 import * as Amplitude from './amplitude';
 
 // TODO: Use a better type here
@@ -307,5 +314,12 @@ export async function apiRetryInvoice(params: {
     'POST',
     `${config.servers.auth.url}/v1/oauth/subscriptions/invoice/retry`,
     { body: JSON.stringify(params) }
+  );
+}
+
+export async function apiCreateSetupIntent(): Promise<FilteredSetupIntent> {
+  return apiFetch(
+    'POST',
+    `${config.servers.auth.url}/v1/oauth/subscriptions/setupintent/create`
   );
 }
--- a/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.test.tsx
@@ -11,7 +11,7 @@ import {
   wait,
 } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
-import { PaymentMethod, PaymentIntent } from '@stripe/stripe-js';
+import { PaymentMethod, PaymentIntent, SetupIntent } from '@stripe/stripe-js';
 import { SignInLayout } from '../../../components/AppLayout';
 import { CUSTOMER, PROFILE, PLAN } from '../../../lib/mock-data';
 import { PickPartial } from '../../../lib/types';
@@ -108,9 +108,15 @@ const CONFIRM_CARD_RESULT = {
   error: undefined,
 };
 
+const CONFIRM_CARD_SETUP_RESULT = {
+  setupIntent: { status: 'succeeded' } as SetupIntent,
+  error: undefined,
+};
+
 const defaultStripeOverride = () => ({
   createPaymentMethod: jest.fn().mockResolvedValue(PAYMENT_METHOD_RESULT),
   confirmCardPayment: jest.fn().mockResolvedValue(CONFIRM_CARD_RESULT),
+  confirmCardSetup: jest.fn().mockResolvedValue(CONFIRM_CARD_SETUP_RESULT),
 });
 
 describe('routes/ProductV2/SubscriptionCreate', () => {
--- a/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx
+++ b/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx
@@ -25,12 +25,13 @@ type RetryStatus = undefined | { invoiceId: string };
 
 export type SubscriptionCreateStripeAPIs = Pick<
   Stripe,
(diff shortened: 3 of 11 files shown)


packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -1,5 +1,12 @@
 import { Config, defaultConfig } from './config';
-import { Plan, Profile, Customer, Subscription, Token } from '../store/types';
+import {
+  Plan,
+  Profile,
+  Customer,
+  Subscription,
+  Token,
+  FilteredSetupIntent,
```
lmorchard, line 8: Might want to house this type here with the apiClient stuff. Those other types are for the Redux store, which should someday go away / [move out into fxa-shared](https://github.com/mozilla/fxa/issues/5912)

packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.test.tsx
```diff
@@ -106,4 +111,11 @@ const defaultStripeOverride: Pick<
       error: undefined,
     };
   },
+  confirmCardSetup: async () => {
+    await wait(500);
```
lmorchard, line 115: I used a delay in stories to give a change to simulate a loading spinner, but we don't need it in tests.

packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx
```diff
@@ -222,7 +230,19 @@ async function handleSubscriptionPayment({
   onRetry: (status: RetryStatus) => void;
   onSuccess: () => void;
 } & SubscriptionCreateAPIs) {
-  // 1. Create the payment method.
```
lmorchard: I think this is in the wrong component - this is for creating a new subscription, where I think we want to use this SetupIntent stuff over in payment method updates in [the SubscriptionItem and PaymentUpdateForm neighborhood of components](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/Subscriptions/SubscriptionItem.tsx#L75)

jackiemunroe: Thanks for the clarification. Reading the docs that were linked, it seemed like we needed to use the SetupIntents for the initial payment method creation during checkout. I'll move this work 😅

lmorchard:
Also probably want to set up [a parallel `/v2` route](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/App.tsx#L106) like I did for the subscription creation page. 

In that case, I needed enough component changes to create a new directory structure. 

But, it occurs to me that SubscriptionItem and PaymentUpdateForm components are pretty deep in the hierarchy of the Subscription route. Might be easier to add a `/v2` route that just passes in a new boolean prop to the existing Subscription route, then flip between PaymentUpdateForm (old) and a new PaymentUpdateFormV2 with the new API flow.

Maybe something like:

```
                      <Route
                        path="/v2/subscriptions"
                        exact
                        render={(props) => (
                          <SettingsLayout>
                            <Subscriptions {{ ...props, newPaymentFlow: true }} />
                          </SettingsLayout>
                        )}
                      />
```

And then plumb that into SubscriptionItem so you can render PaymentUpdateForm or PaymentUpdateFormV2 conditionally

lmorchard:
Oh, and in case it's useful and since I'm thinking & brain dumping:

The existing PaymentUpdateForm component [uses Redux actions to manage & track API calls](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx#L34-L36).

We can & should stop doing that. Over in the SubscriptionCreate component, I switched to [just making API calls directly](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx#L270) and manage things [like an `inProgress` state](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx#L270) locally rather than in Redux state. Way simpler.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -91,6 +125,10 @@ export const PaymentUpdateForm = ({
   const onPayment = useCallback(
     (tokenResponse: stripe.TokenResponse) => {
       if (tokenResponse && tokenResponse.token) {
+        if (useSCAPaymentFlow) {
+          // Use SetupIntents to Confirm the Card
+          handleSCA(tokenResponse.token.id);
```
lmorchard, line 130: FWIW, I know this is in progress, but wanted to mention: 

We're also going to want to swap in [`PaymentFormV2`](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx) over here when `useSCAPaymentFlow` is true. 

That uses the new Stripe elements library. It also has [a different `onSubmit` callback](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx#L87) that replaces `onPayment` / `onPaymentError` callbacks no longer generates a payment token.

chenba:
There are issues with the response validators on the API side for the setupintent/create.  I see `null` for both `next_action` and `payment_method`, but the validator does not explicitly `allow(null)` on them.  So if you are getting 500s, that could be the reason.

Setting a default Payment Method in invoice settings on v1 customers with charges does appear to work.  However, the old card lingers as a payment method and the default charge for the customer.  I think we should remove the old payment methods and Payment Methods as part of the Payment Method update.

lmorchard:
Going to close this - superseded by PR #6030

PR #5927, opened 2020-07-13
fix(test): use ts-node in transpile only mode for script tests

dannycoates:
The tests were exceeding the timeout. Type checking here isn't necessary.

### DIFF
--- a/packages/fxa-auth-server/test/oauth/key-management-scripts.js
+++ b/packages/fxa-auth-server/test/oauth/key-management-scripts.js
@@ -28,9 +28,8 @@ describe('the signing-key management scripts', function () {
       return execFileSync(
         process.execPath,
         [
-          require.resolve('ts-node').replace('index.js', 'bin.js'),
-          '-P',
-          path.join(base, '../tsconfig.json'),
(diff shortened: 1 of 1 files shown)


meandavejustice:
😎

packages/fxa-auth-server/test/oauth/key-management-scripts.js
```diff
@@ -39,6 +38,8 @@ describe('the signing-key management scripts', function () {
             FXA_OPENID_NEWKEYFILE: newKeyFile,
             FXA_OPENID_OLDKEYFILE: oldKeyFile,
             NODE_ENV: 'dev',
+            TS_NODE_TRANSPILE_ONLY: true,
```
vbudhram, line 41: Should we do this for all the tests? https://github.com/mozilla/fxa/blob/99385e89a555afa85aa8b4b9ec66b9c5f0251c73/packages/fxa-auth-server/scripts/test-local.sh#L13

dannycoates: maybe?

Issue #5520, opened 2020-06-02
Advertize AET public key in /.well-known/fxa-client-configuration


data-sync-user:
(I was sure we already had an issue on file for this, but I can't find such - please feel free to close this as a dupe [~jhirsch@mozilla.com])

For FxA clients to be able to generate placeholder AET anon_id values, they need to be able to discover the appropriate public key with which to encrypt them. I propose that FxA advertize this value as part of its `/.well-known/fxa-client-configuration` bundle, as a list of keys (which to begin with would only contain a single item.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2014)
┆Issue Number: FXA-2014

data-sync-user:
➤ Ryan Kelly commented:

Strawman: add a new key to `fxa-client-configuration` named "ecosystem_anon_id_keys". Its value would be a list of JWK objects, to support eventual rotation of these keys. For the first version it would contain only a single key, corresponding to the one owned by the AET data pipeline.

data-sync-user:
➤ Chelsea Lewis commented:

[~wclouser@mozilla.com]/[~jhirsch@mozilla.com] Can we get an estimate on this one, assumine it's not a dupe (note rfk's comment within Description)

data-sync-user:
➤ Jared Hirsch commented:

Sure: the route for fxa-client-configuration already exists, and we just need to coordinate with ops and data eng teams to make sure the key is dropped into our configs as a list with a single item. Calling it a 3 because of the coordination pieces.

data-sync-user:
➤ Jared Hirsch commented:

See this example for exactly where to put the config value: https://github.com/mozilla/fxa/pull/5859/files#diff-9f17be0dc58fa7cd854fc8266811a301 ( https://github.com/mozilla/fxa/pull/5859/files#diff-9f17be0dc58fa7cd854fc8266811a301 )

Still need to coordinate with pipeline and fxa ops to understand how to expose different params in different fxa environments (local, staging, prod)

PR #5930, opened 2020-07-13
feat(AET): Add 'ecosystem_anon_id' field to content server config and expose keys in fxa-client-configuration

LZoog:
## Because

* We need the public key from the profile server exposed to the content-server so that desktop clients can encrypt the ecosystem user ID to the ecosystem anon ID.

## This pull request

* Adds 'ecosystem_anon_id.keys_file' and 'ecosystem_anon_id.keys' to the content server config, checking for a path set in a new env var ECOSYSTEM_ANON_ID_KEYS_FILE and setting it to the .keys value, then exposes it in the .well-known/fxa-client-configuration bundle.

Closes: #5520 (oopsy on my branch name here)

## Checklist

### Other information
* Defaults to an empty array when `ECOSYSTEM_ANON_ID_KEYS_FILE` env var isn't set
* Expects `ECOSYSTEM_ANON_ID_KEYS_FILE=file` to equal `config/[file]`. I had the thrown error in a check for `conf.get('env') !== 'development'` as well but removed it in case we want to set the env var/file and pull it in locally.

### DIFF
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -748,6 +748,18 @@ const conf = (module.exports = convict({
     env: 'SYNC_TOKENSERVER_URL',
     format: 'url',
   },
+  ecosystem_anon_id: {
+    keys_file: {
+      default: '',
+      doc: 'Path to a file containing an array of Account Ecosystem Telemetry pipeline JWK public key objects for encrypting the ecosystem user ID to the ecosystem anon ID',
+      env: 'ECOSYSTEM_ANON_ID_KEYS_FILE',
+      format: String,
+    },
+    keys: {
+      doc: 'Array of Account Ecosystem Telemetry pipeline JWK public key objects for encrypting the ecosystem user ID to the ecosystem anon ID',
+      default: [],
+    }
+  },
   template_path: {
     default: path.resolve(__dirname, '..', 'templates'),
     doc: 'The location of server-rendered templates',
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -748,6 +748,12 @@ const conf = (module.exports = convict({
     env: 'SYNC_TOKENSERVER_URL',
     format: 'url',
   },
+  ecosystem_anon_id_keys: {
+    default: [],
+    doc: 'Account Ecosystem Telemetry pipeline keys',
```
6a68, line 753: Let's additionally specify that they are public keys specified in JWK format

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -867,6 +879,24 @@ if (missingLangs.length) {
...
 conf.set('are_dist_resources', areDistResources);
 
+if (conf.get('ecosystem_anon_id.keys_file')) {
+  const keysFile = path.resolve(
+    __dirname,
+    '..',
+    'config',
+    conf.get('ecosystem_anon_id.keys_file')
+  );
+  // Alter `ecosystem_anon_id.key` if the keys file exists.
```
LZoog, line 889: Should be `ecosystem_anon_id.keys`, will update with any other needed edits after second review.

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -867,6 +879,24 @@ if (missingLangs.length) {
...
+  if (fs.existsSync(keysFile)) {
+    const keys = JSON.parse(fs.readFileSync(keysFile));
```
6a68, line 891: Might want to try/catch this line, since reading or parsing might fail.

LZoog: If I add that, our linter complains that it's an unnecessary try/catch. I just tried adding a file here with some jargon in an array and get an "unexpected token" error thrown at me from `JSON.parse` with a callstack so I think we're covered here. 👍

Issue #5659, opened 2020-06-12
“Cancel” button highlighter overlaps the “Continue” button from “Step 1 of 2: Re-enter password” modal


vasilicatamas:
**Build**
Stage [Train 1.175.0]

**Affected platforms**
Windows 10

**Steps to reproduce**
1. Launch Firefox and login using a Firefox account.
2. Navigate to Account Settings page.
3. Click on "Enable" button from the Account recovery section.
4. Click on “Generate” button.
5. Click on “Cancel” button.

**Expected results**
”Cancel” button is correctly displayed.

**Actual Results**
“Cancel” button highlighter is displayed over the “Continue” button.

![Screenshot_17](https://user-images.githubusercontent.com/18114210/84516756-e6993300-acd6-11ea-97b8-80087fc87f0e.png)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2111)
┆Issue Number: FXA-2111

byrw:
can I work on it?

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.180.0] under Windows 10 64-bit. "Cancel" button is correctly displayed.

![Screenshot_18](https://user-images.githubusercontent.com/18114210/87936964-77b0b600-ca9c-11ea-9f06-2b54ae9fd20e.png)

PR #5939, opened 2020-07-14
fix(style): add margin to the 'account recovery' button

byrw:
Because:

* The buttons were overlapping.

This commit:

* Add a margin to the button.

Closes: #5659

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings-account-recovery.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings-account-recovery.scss
@@ -147,3 +147,7 @@ button.button-link {
   background: transparent;
   color: $color-blue;
 }
+
+button.generate-key-link{
+  margin-bottom: 16px;
+}


LZoog:
Thanks for taking on this issue @byrw! I'm running this by our Design team real quick to see if they have a preference on the amount of space here or if this is good to go, I'll get back to you shortly. I think regardless we might want a more specific commit message like "fix(style): add bottom margin to the 'continue' recovery key button" rather than generic "button," though hold off on changing that in case our team prefers a different amount of spacing so you can update both at once. 🙂

packages/fxa-content-server/app/styles/modules/_settings-account-recovery.scss
```diff
@@ -147,3 +147,7 @@ button.button-link {
   background: transparent;
   color: $color-blue;
 }
+
+button.generate-key-link{
+  margin-bottom: 5px;
```
LZoog, line 152: Our design team got back to me and they'd actually like this to be `16px`. Will you update this here and edit your commit message to include `'account recovery' button` while we're at it?

Will r+ after, thank you!

byrw: done! :+1:

Issue #5915, opened 2020-07-10
[admin panel] Clarify primary/secondary email


data-sync-user:
If I search for an email on the admin panel, sometimes it returns a different email.  This is a primary/secondary relationship but it can be jarring if the emails aren't similar enough to realize how they are related.  We should clarify that in the UI.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2269)
┆Issue Number: FXA-2269

PR #5940, opened 2020-07-14
feat(admin): support secondary emails and query highlighting

jodyheavener:
## Because

Right now in the Admin Panel you can perform an Email Blocks query using an account's secondary email, and it will return the correct account, but only display the primary email. So you have little context as to why this account is being returned.

## This pull request

- Updates the Admin Server GraphQL query to resolve and return `emails`, an array of the account's emails, with `isPrimary` and `isVerified` properties (but keeps `email` and `emailVerified` top-level properties "just in case"). 
- Update the Admin Panel UI to reflect both primary and secondary emails, and highlights the email that matched your query.

**Notes for reviewer:**

- It looks like Prettify got its hands on some of these files, so the diff looks a little larger than it should 😢 
- Some of the markup/style changes are a little... bleh... however, I plan on doing another pass through the Admin Panel to start styling with Tailwind and do a light UI update in the near future.

Closes: #5915

## Checklist

## Screenshots

Screen Shot 2020-07-14 at 11 20 01 AM

### DIFF
--- a/packages/fxa-admin-panel/src/components/EmailBlocks/Account/index.tsx
+++ b/packages/fxa-admin-panel/src/components/EmailBlocks/Account/index.tsx
@@ -10,11 +10,11 @@ import './index.scss';
 
 type AccountProps = {
   uid: string;
-  email: string;
   createdAt: number;
-  emailVerified: boolean;
-  emailBounces: Array<EmailBounceProps>;
+  emails: EmailProps[];
+  emailBounces: EmailBounceProps[];
   onCleared: Function;
+  query: string;
 };
 
 type EmailBounceProps = {
@@ -22,6 +22,13 @@ type EmailBounceProps = {
   createdAt: number;
   bounceType: string;
   bounceSubType: string;
+};
+
+type EmailProps = {
+  email: string;
+  isVerified: boolean;
+  isPrimary: boolean;
+  createdAt: number;
 };
 
 const DATE_FORMAT = 'yyyy-mm-dd @ HH:MM:ss Z';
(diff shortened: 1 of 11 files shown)


packages/fxa-admin-panel/src/components/EmailBlocks/Account/index.tsx
```diff
@@ -46,7 +53,10 @@ export const ClearButton = ({
       return;
     }
 
-    clearBounces({ variables: { email } });
+    // This could be improved to clear bounces for individual email
+    // addresses, but for now it seems satisfactory to clear all bounces
+    // for all emails, since they own all of the addresses
+    emails.forEach((email) => clearBounces({ variables: { email } }));
```
jodyheavener, line 59: Note, currently clicking "Clear all bounces" doesn't clear bounces for secondary addresses. This should hopefully do the trick.

jodyheavener: @bbangert this is what I was referring to about executing on each address.

LZoog: PR LGTM besides a possible improvement here, we could pull this out and only clear the primary e-mail for now. Would say maybe hold until @bbangert chimes in here but I'll r+ since this isn't a blocker for this PR from my PoV, nice work @jodyheavener, the highlight is sweet!

bbangert: I see no reason not to clear them all. 👍

Issue #5348, opened 2020-05-18
[content server] Show a user friendly error message when uploaded avatar image is too large


data-sync-user:
This is a follow-up to FXA-1858.  When a user upload an avatar image that's greater than the configured size limit, we should display an error message to inform the user, instead of the current generic and confusing "System unavailable, try again soon."

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-1908)
┆Issue Number: FXA-1908

data-sync-user:
➤ Wil Clouser commented:

[~pwalmsley@mozilla.com] Could you give us some text to use for this?

data-sync-user:
➤ Philip Walmsley commented:

"The image is too large to uploaded and needs to be resized."

First pass without Content looking at it, good for now.

data-sync-user:
➤ Wil Clouser commented:

To be clear, we're talking about the file size of the image being uploaded, not the pixel dimensions of the image.  Should we change the wording to specify that?

data-sync-user:
➤ Philip Walmsley commented:

Yes, thank you for clarifying that.

"The image file size is too large to be uploaded."

vasilicatamas:
“The image file size is too large to be uploaded.” error message is successfully displayed on Stage [Train 1.180.0] under Windows 10 64-bit, Android 9 and iOS 13.4.1. But its position doesn't seem to be the right one and a new issues was filed for tracking this bug separately #5975 . Marking this issues as Verified Fixed.

![Screenshot_14](https://user-images.githubusercontent.com/18114210/87806261-4008fa80-c85f-11ea-8509-15fb89c6eb99.png)

Issue #5123, opened 2020-04-27
Unhelpful error when trying to upload an account avatar image that is too large [bz1632246]


bug-mirror:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1632246

## Steps to reproduce

When trying to upload a new account avatar image for my account, I kept getting an error message:  "Service Unavailable, try again later".

After some time and guesswork, I figured out that the image was just too large and after resizing it, it worked just fine.

## Expected behavior

I expected the site to indicate the issue was with the file size and not look like a 5xx error or something.

Bonus:  It would be _really_ nice to know that there are restrictions to the image size on the upload page.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1767)
┆Issue Number: FXA-1767

data-sync-user:
➤ Wil Clouser commented:

Thanks for filing this.  It would be good to know what the restrictions are.  Some quick testing showed unreliable results (eg. a solid black image resulted).  We should also enforce/announce limits on the client side to save people uploading time.

Vijay says the default config has this limit at 1MB.

I'm going to add the needs:ux flag for how to display the error and what the wording is.  /cc [~pwalmsley@mozilla.com]

rbillings:
**What type is it? How frequently will users hit this issue?**
Bug due to an unhelpful error
**Frequency**: low
**How was this found?**
User [dev] filed in Bugzilla.
**What caused it?**
**Why didn’t we catch it sooner?**
This happened for so long it wasn't noticed anymore
**How do we ensure it doesn’t repeat?**
It's a one time issue
**Is this a symptom of a broader issue**
No

vasilicatamas:
This issue seems to be similar to https://jira.mozilla.com/browse/FXA-1908

byrw:
can I work on it?

PR #5941, opened 2020-07-14
fix(content-server): create storage size limit to uploaded avatar image.

byrw:
Because:

* When the user uploaded a big image and an error message appeared, it was not possible to know the cause of the problem.

This commit:

* Create a limit size to uploaded images.

Closes: #5348, #5123

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/constants.js
+++ b/packages/fxa-content-server/app/scripts/lib/constants.js
@@ -86,6 +86,9 @@ module.exports = {
   PROFILE_IMAGE_MIN_WIDTH: 100,
   DEFAULT_PROFILE_IMAGE_MIME_TYPE: 'image/jpeg',
 
+  // Limit to 1.35 megabytes
+  PROFILE_FILE_IMAGE_MAX_UPLOAD_SIZE: 1.35 * 1024 * 1024,
+
   ONERROR_MESSAGE_LIMIT: 100,
 
   ACCOUNT_UPDATES_WEBCHANNEL_ID: 'account_updates',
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/app/scripts/lib/constants.js
```diff
@@ -86,6 +86,9 @@ module.exports = {
   PROFILE_IMAGE_MIN_WIDTH: 100,
   DEFAULT_PROFILE_IMAGE_MIME_TYPE: 'image/jpeg',
 
+  // Limit to 20 megabytes
+  PROFILE_FILE_IMAGE_MAX_UPLOAD_SIZE: 20 * 1000 * 1000,
```
jodyheavener, line 90: I thought the value for this was going to be [this](https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/lib/config.js#L100-L104), but it looks like in production it's actually `1024 * 1024 * 1.35`

byrw: I just picked an arbitrary value to be the limit. What number do you think it's better? Should I choose `1.35*1024*1024`?

jodyheavener: Yes I believe so, given that’s what we have for the server-side validation.

byrw: If there is a server-side validation, then why it is possible to upload a big image?

jodyheavener: I believe that's where folks are seeing the "System unavailable, try again soon" error message. But it's possible something else is going on that I am unaware of. 

Thank you for updating! I've verified locally that a 3MB file fails, but a 1.2MB file is accepted, as well as some other sizes to be sure.

Issue #5944, opened 2020-07-14
[admin-panel] mozilla observatory - two failing checks for consideration


data-sync-user:
|----|----|
|[Referrer Policy](https://infosec.mozilla.org/guidelines/web_security#referrer-policy)| |

|----|
|~~5\|Referrer~~Policy header set unsafely to `"origin"`, `"origin-when-cross-origin"`, or `"unsafe-url"`|

|----|----|
|Restricts where `<form>` contents may be submitted by using `form-action 'none'`, `form-action 'self'`, or specific URIs| |

|----|
| |

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2265)
┆Issue Number: FXA-2265

PR #5946, opened 2020-07-14
feat(admin): set referrer-policy header, form-action csp directive

jodyheavener:
## Because

[Mozilla Observatory](https://observatory.mozilla.org/) identified two areas where security could be improved with the Admin Panel:

1. We don't set the [`form-action` CSP directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/form-action), which dictates which addresses can be used in `<form>` submissions.
2. We don't set the [`Referrer-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) header, which means it defaults to `no-referrer-when-downgrade`. 

## This pull request

1. Sets the `form-action` CSP directive to `'none'`. This could also be `'self'`, but because currently our only form's submission is handled by React there is no need for an `action`. In the future we may want to relax this. I left a note in the code.
2. Sets the `Referrer-Policy` header to `same-origin`. I could be persuaded to use `no-referrer`, too.

Also, updated some tests to remove unnecessary use of Promises and assert all expected headers set by middleware.

Closes: #5944

## Checklist

### DIFF
--- a/packages/fxa-admin-panel/server/lib/server.test.ts
+++ b/packages/fxa-admin-panel/server/lib/server.test.ts
@@ -12,39 +12,43 @@ function expectValueNotToBeUnknown(value: string) {
 }
 
 describe('Test simple server routes', () => {
-  test('__version__ should have correct structure', () => {
-    return new Promise(async (done) => {
-      const response = await request(app).get('/__version__');
-      expect(response.status).toStrictEqual(200);
-      expect(response.header['content-type']).toStrictEqual(
-        'application/json; charset=utf-8'
-      );
-      const body = response.body;
-      expect(Object.keys(body).sort()).toStrictEqual([
-        'commit',
-        'source',
-        'version',
-      ]);
-      expectValueNotToBeUnknown(body.commit);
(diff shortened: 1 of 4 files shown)


packages/fxa-admin-panel/server/lib/server.test.ts
```diff
@@ -12,39 +12,43 @@ function expectValueNotToBeUnknown(value: string) {
...
-        'version',
-      ]);
-      expectValueNotToBeUnknown(body.commit);
-      expectValueNotToBeUnknown(body.source);
-      expectValueNotToBeUnknown(body.version);
-      expect(body.version).toStrictEqual(pkg.version);
-      // check that the commit value looks like a git SHA
-      expect(body.commit).toMatch(/^[0-9a-f]{40}$/);
-      done();
+  test('routes should have correct headers', async () => {
```
LZoog, line 15: Huh, TIL `it` in jest is an alias of `test`.

PR #5948, opened 2020-07-14
fix(l10n): fixed currency localization for "other than US" locales

dannycoates:
For en locales that don't have their own ftl file this change
sets their FluentResource to be that of en-GB, which does have
an ftl file in order for the currency to be rendered as 'US$x.xx'
instead of the US fallback of '$x.xx' which is ambiguous in which
currency is meant.

These locales are currently hard coded in AppLocalizationProvider
because other services fail when they can't load l10n files that
are present in supportedLanguages.json but not in the l10n repo.

### DIFF
--- a/packages/fxa-payments-server/src/lib/AppLocalizationProvider.tsx
+++ b/packages/fxa-payments-server/src/lib/AppLocalizationProvider.tsx
@@ -10,6 +10,8 @@ import React, { Component } from 'react';
 
 import availableLocales from 'fxa-shared/l10n/supportedLanguages.json';
 
+const OTHER_EN_LOCALES = ['en-NZ', 'en-CA', 'en-SG', 'en-MY']
+
 async function fetchMessages(baseDir: string, locale: string, bundle: string) {
   try {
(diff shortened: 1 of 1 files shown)


dannycoates:
I tried adding a test to `AppLocalizationProvider.test.js` but it ends up with no US specifier, i.e. `$1.23`. It works in a real browser so I wonder if this is a quirk of the test harness? @lmorchard @chenba ideas?

chenba:
> 
> 
> I tried adding a test to `AppLocalizationProvider.test.js` but it ends up with no US specifier, i.e. `$1.23`. It works in a real browser so I wonder if this is a quirk of the test harness? @lmorchard @chenba ideas?

Is there a gist or branch I can look at?

dannycoates:
@chenba https://gist.github.com/dannycoates/b642f757f7f6c52959b66b6e887026bf

chenba:
Yeah, `getLocalizedCurrency` still has `en-US` hard coded for its locale.  😭

chenba:
In other words, your test is too good.  I think we just need to know a string from 'en-GB' is used when the locale is 'en-NZ'?

PR #5951, opened 2020-07-14
feat(auth-server): use payment method on file for card

### LINKED ISSUE #5902: Return invoice default payment method details for customer
The auth-server customer endpoint returns source information. For customers that use payment methods with Payment Intent and Setup Intent's, the default payment method for invoices should be returned here instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2256)
┆Issue Number: FXA-2256

bbangert:
Because:

* We only looked at sources to return payment method before and
  now we also handle payent methods as a default invoice option.

This commit:

* Looks to see if a payment method was attached as a default invoice
  option and uses it for returned card data.

Closes #5902

## Checklist

I've also added expansion of this object to the cached customer object, even though its not needed at the moment so that they will match up as desired later.

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.js
+++ b/packages/fxa-auth-server/lib/payments/stripe.js
@@ -10,6 +10,7 @@ const stripe = require('stripe').Stripe;
 
 /** @typedef {import('stripe').Stripe.Customer} Customer */
 /** @typedef {import('stripe').Stripe.Event} StripeEvent */
+/** @typedef {import('stripe').Stripe.PaymentMethod} PaymentMethod */
 /** @typedef {import('stripe').Stripe.Plan} Plan */
 /** @typedef {import('stripe').Stripe.Product} Product */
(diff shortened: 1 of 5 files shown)


lmorchard:
Gave this a try with the `/v2/product` subscription flow - http://localhost:3031/v2/products/prod_GqM9ToKK62qjkK

But the getCustomer route throws a 500 error on page load. No real detail in logs. Can dig in further after dinner or early tomorrow morning.

bbangert:
Have I mentioned lately how annoying the Stripe docs can be sometimes? They make it out like that *should work*, oh well.

bbangert:
Phew, ok, what a pain. I tested the /v2/ path locally, thanks for the URL! It works, the card info shows up properly on the Subscriptions management page. I was unable to get the /v2/ working initially and dug in to find out the `next_action` validator didn't agree with whatever object was there, so I removed the validator entirely since there's not really any point validating it since Stripe says you should return it as is. That got the whole flow working.

PR #5955, opened 2020-07-15
fix(l10n): use accept language for currency l10n in emails

dannycoates:
same reason as 5f8495594d4edd8c3f959aad70bd70b2f5e1c253

### DIFF
--- a/packages/fxa-auth-server/test/local/senders/email.js
+++ b/packages/fxa-auth-server/test/local/senders/email.js
@@ -1244,6 +1244,15 @@ describe('lib/senders/email:', () => {
     });
   }
 
+  it('formats currency strings sanely', () => {
+    const result = mailer._getLocalizedCurrencyString(
+      123,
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -1244,6 +1244,15 @@ describe('lib/senders/email:', () => {
     });
   }
 
+  it('formats currency strings sanely', () => {
+    const result = mailer._getLocalizedCurrencyString(
+      123,
+      'USD',
+      'en-US;q=0.5,en;q=0.3,en-NZ'
```
lmorchard, line 1251: Hmm, trying to remember how these priorities work. Would have thought that the en-US q value would take priority. Seems to work whether I understand it or not, though. :)

dannycoates:
Maybe we want to do this for date & time too? (later)

PR #5958, opened 2020-07-15
chore(cad): disable qr code experiment

vbudhram:
## Because

- We have gathered enough experiment data and don't need to keep running while we do the analysis

## This pull request

- Stops the qr code experiment


vbudhram:
@davismtl this experiment will be disabled in next train (181).

Issue #5884, opened 2020-07-07
FPN listed many times in 'delete account' warning screen


6a68:
I signed up for FPN last week when testing an FPN bug. I've canceled my subscription and uninstalled the addon.

When I looked at the 'delete account' warning screen today on the settings page, I noticed a ton of duplicate FPN entries (see screenshot below).

<img width="617" alt="Screen Shot 2020-07-07 at 4 56 00 PM" src="https://user-images.githubusercontent.com/96396/86857278-0bcd5700-c073-11ea-8452-43237fef145a.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2247)
┆Issue Number: FXA-2247

PR #5964, opened 2020-07-16
fix(delete account): Filter active subscriptions for unique product names

LZoog:
## Because
* If users have multiple subscriptions of the same name they show multiple times.

## This pull request
* Creates a new Set of active subscription product names and displays this list rather than display all active subscriptions.

Closes: #5884

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
@@ -23,11 +23,9 @@
 
       <ul class="delete-account-product-list {{#hasTwoColumnProductList}}two-col{{/hasTwoColumnProductList}}">
         {{#subscriptions}}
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/templates/settings/delete_account.mustache
```diff
@@ -23,11 +23,9 @@
 
       <ul class="delete-account-product-list {{#hasTwoColumnProductList}}two-col{{/hasTwoColumnProductList}}">
         {{#subscriptions}}
-          {{#product_name}}
-            <li class="delete-account-product-subscription" title="{{product_name}}">
-              {{product_name}}
-            </li>
-          {{/product_name}}
+          <li class="delete-account-product-subscription" title="{{.}}">
+            {{.}}
```
jodyheavener, line 27: TIL

PR #5968, opened 2020-07-16
chore(build): build fxa-email-service less often

dannycoates:
We're building email-service on train tags more often
than we need to and doing so adds a good chunk of time to the
build process.

We've already got some logic in hash-source.sh to determine
when a build should happen. This change adjusts the criteria
to exclude the files that change when a tag is done even when
there's no meaningful changes.

### DIFF
--- a/packages/fxa-email-service/scripts/hash-source.sh
+++ b/packages/fxa-email-service/scripts/hash-source.sh
@@ -4,10 +4,13 @@ DIR=$(dirname "$0")
 cd $DIR/..
 
 git ls-files | \
-# exclude version.json because it changes on every build.
-# Cargo.toml serves as a signal for new releases
-grep -v -e 'version\.json$' | \
+# exclude these files because they change more often than we want to build.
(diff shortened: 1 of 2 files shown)


packages/fxa-email-service/scripts/hash-source.sh
```diff
@@ -4,10 +4,13 @@ DIR=$(dirname "$0")
...
-# exclude version.json because it changes on every build.
-# Cargo.toml serves as a signal for new releases
-grep -v -e 'version\.json$' | \
+# exclude these files because they change more often than we want to build.
+grep -v -e '^version\.json$' -e '^CHANGELOG\.md$' -e '^Cargo\.toml$' -e '^Cargo\.lock$' | \
 sort | \
 xargs cat | \
+# This is kind of a silly way to let dependency updates trigger a new hash without
+# including the package version line, but a better way involves more parsing than
+# I'd like to do a the moment. It skips the first 4 lines of Cargo.toml
```
chenba, line 13: Tiny typo here: at is missing its 't'.

dannycoates: nice catch

PR #5972, opened 2020-07-17
fix(payments): subscription emails should work without payment methods

### LINKED ISSUE #5903: Subscription emails should not require a payment method
Free trials with no card on file, or reviewer style accounts with no card on file will trigger errors when we try to send invoices. We should update our email/webhook handling to allow customers to not have a payment method on file.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2258)
┆Issue Number: FXA-2258

lmorchard:
- rather than throw a validation error, simply hide sections of emails
  that mention credit cards when the payment method info is unavailable

fixes #5903


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1501,25 +1486,15 @@ class StripeHelper {
...
-          `Expected card payment method details for invoice: ${invoice.id}`
-        )
-      );
+    let lastFour: string | null = null;
+    let cardType: string | null = null;
+    if (charge?.payment_method_details?.card) {
+      ({
+        brand: cardType,
+        last4: lastFour,
+      } = charge.payment_method_details.card);
```
chenba, line 1495: nit: these few lines are identical to those above, so perhaps wrap them into a function.

lmorchard:
Tweaked to DRY up card detail extraction, will merge after CI goes green

Issue #5952, opened 2020-07-15
Mozilla VPN icon is missing in the “Thanks!” resubscribe popup


ValentinaPC:
#### Affected versions:
- All Firefox;

#### Affected FxA versions:
- 1.179.3;

#### Affected Platforms:
- Windows 10 (64-bit);

#### Prerequisites:
- have an FxA account with active subscription;

#### Steps to reproduce:
1. Go to manage subscription page -https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/subscriptions using the account from prerequisites;
2. Cancel Mozilla VPN subscription;
3. Click the “Resubscribe” button and finalize resubscription;
4. Observe “Thanks!” for resubscribing popup;

#### Expected result:
- Mozilla VPN product icon is displayed above the thank-you message.

#### Actual result:
- Mozilla VPN product icon is NOT displayed above the thank-you message.
![iconmissing](https://user-images.githubusercontent.com/15685960/87529423-a6dbb780-c697-11ea-9660-ec5b9806d9b7.png)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2291)
┆Issue Number: FXA-2291

data-sync-user:
➤ Wil Clouser commented:

[~jgruen@mozilla.com] said this was fixed on Slack

ValentinaPC:
> 
> 
> ➤ Wil Clouser commented:
> 
> [~[jgruen@mozilla.com](mailto:jgruen@mozilla.com)] said this was fixed on Slack

Verified this on Stage 179.4, Fx 78.02, Win 10, and the issue is still reproducing. Also observed, that the icon is there, but it is white.
Please see below video.
![hideands](https://user-images.githubusercontent.com/15685960/87665981-7e73bc00-c770-11ea-9703-9ba1c634d31f.gif)

data-sync-user:
➤ John Gruen commented:

So the Icon is there, but it's all white...i think this gets fixed by adding a color background or something 🙁

ValentinaPC:
Verified as fixed in stage, using Fx78.0.2 and Windows 10 (64-bit).
Postfix screenshots below.
![image](https://user-images.githubusercontent.com/15685960/88532250-ed70d080-d00c-11ea-9e0f-12970ae5445a.png)
![image](https://user-images.githubusercontent.com/15685960/88532225-e21da500-d00c-11ea-871f-766903be8450.png)

PR #5973, opened 2020-07-17
chore(styles): make subscription icon background configurable as metadata

johngruen:
## Because

Icons for mozilla services will be white and will need different bg colors

## This pull request

makes bg color configurable as stripe metadata. Any valid CSS color value can be included in metadata (gradients, color names etc)

Closes: #5952

## Checklist

Notes:
* I may need to add a tests here, but could use a bit of help/direction if so
* I will add a PR for ecosystem docs to highlight the new metadata key once this lands

## Screenshots (Optional)

image
image

### DIFF
--- a/packages/fxa-payments-server/src/components/PlanDetails/index.tsx
+++ b/packages/fxa-payments-server/src/components/PlanDetails/index.tsx
@@ -37,13 +37,17 @@ export const PlanDetails = ({
     interval,
     interval_count,
   } = selectedPlan;
-  const { webIconURL } = metadataFromPlan(selectedPlan);
+  const { webIconURL, webIconBackground } = metadataFromPlan(selectedPlan);
   const productDetails = productDetailsFromPlan(
     selectedPlan,
     navigatorLanguages
   );
 
   const role = isMobile ? undefined : 'complementary';
+
+  const setWebIconBackground = webIconBackground
+    ? { background: webIconBackground }
+    : '';
 
   const planPrice = formatPlanPricing(
     amount,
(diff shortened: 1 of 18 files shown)


packages/fxa-payments-server/src/lib/mock-data.tsx
```diff
@@ -46,6 +46,7 @@ export const SELECTED_PLAN: Plan = {
   interval_count: 1,
   product_metadata: {
     webIconURL: 'http://placekitten.com/49/49?image=2',
+    webIconBackground: 'purple',
```
johngruen, line 49: So i pretty much added a color as mock data anywhere an icon was supplied

johngruen:
There are tests for not including an icon in metadata ([for example here](https://github.com/mozilla/fxa/blob/make-product-icon-bg-metadata/packages/fxa-payments-server/src/routes/Product/index.test.tsx#L113))...i may need to add these, but could use a little help if so...

lmorchard:
Also, good job figuring out how to add a product metadata prop 😅

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -64,7 +68,10 @@ export const PlanDetails = ({
         >
           <div className="plan-details-header">
             <div className="plan-details-header-wrap">
-              <div className="plan-details-logo-wrap">
+              <div
+                className="plan-details-logo-wrap"
+                style={{ ...setWebIconBackground }}
```
lmorchard, line 73: I guess if the markup around the icon ever gets more complex, we should consider extracting it out to a component that does stuff like the `setWebIconBackground` logic. But not worth holding this PR up for

johngruen:
@lmorchard WRT setting the background as color, I did this based on the hunch that lead page design will change far more rapidly than will logo design, so having a color element that is easier to swap than uploading a static asset felt right. It also gives us a little more flexibility about placing an svg on an arbitrarily large background if our page layout changes.

Issue #5954, opened 2020-07-15
Update fxa-js-client to update the anon_id


data-sync-user:
These values need to be passed to the auth-server `/account/login` and `/account/create` endpoints.

They should be sent over as an optional param.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2293)
┆Issue Number: FXA-2293

data-sync-user:
➤ Ryan Kelly commented:

>These values need to be passed to the auth-server `/account/login` and `/account/create` endpoints.

I don't have much context on the conversation that lead to this bug, but wanted to flag some potential complexity here. In order to calculate the eco_uid and turn it into an eco_anon_id, you have to have the user's kB, which means that you need to already be authenticated. So, I'm concerned that you won't have access to eco_anon_id at the time you're calling `/account/login` or `/account/create`.

data-sync-user:
➤ Vijay Budhram commented:

[~rkelly@mozilla.com] Thanks for flagging, we I've updated the bug to mention adding js-client functions to only update the id. We should be able to call this function after the user successfully logins and has access to all the required info to generate the id.

vbudhram:
> call this function after the user successfully logins and has access to all the required info to generate the id

I think the plan now is to not deal with rotating on login and only when user changes their password, which simplifies things a good bit.

PR #5979, opened 2020-07-17
feat(auth-js + content): add updateEcosystemAnonId to auth clients

jodyheavener:
## Because

- We need to be able to send an updated EAI to the auth server

## This pull request

- Updates the fxa-js-client to also support this
- Updates the content server auth client to support this
  - It delegates to the fxa-js-client
- Allows the EAI endpoint on the auth server to authenticate with a sessionToken

## Issue that this pull request references

Reference: #5954 - which turned into this PR after discussion.

I'm not sure if we can close that issue from this. It sounded like we might not want to implement it just yet.

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/lib/fxa-client.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/fxa-client.js
@@ -1884,4 +1884,61 @@ describe('lib/fxa-client', function () {
       });
     });
   });
+
+  describe('updateEcosystemAnonId', () => {
+    const sessionToken = 'the silence';
+    const ecosystemAnonId = 'the gold';
+
+    it('calls the client method', () => {
+      const clientMethod = sinon
+        .stub(realClient, 'updateEcosystemAnonId')
+        .resolves();
+
+      return client
+        .updateEcosystemAnonId(sessionToken, ecosystemAnonId)
+        .then((res) => {
+          sinon.assert.calledWith(clientMethod, sessionToken, ecosystemAnonId);
+        });
+    });
+
+    it('calls the request method', () => {
+      sinon.spy(realClient, 'updateEcosystemAnonId');
+      const requestMethod = sinon.stub(realClient, 'sessionPut').resolves();
+
+      return client
+        .updateEcosystemAnonId(sessionToken, ecosystemAnonId)
+        .then((res) => {
+          requestMethod.calledWith(
+            '/account/ecosystemAnonId',
+            sessionToken,
+            {
+              ecosystemAnonId,
+            },
+            new Headers({})
+          );
+        });
+    });
+
+    it('supports ifNoneMatch option', () => {
+      sinon.spy(realClient, 'updateEcosystemAnonId');
+      const requestMethod = sinon.stub(realClient, 'sessionPut').resolves();
+
+      return client
+        .updateEcosystemAnonId(sessionToken, ecosystemAnonId, {
+          ifNoneMatch: '*',
+        })
+        .then((res) => {
+          requestMethod.calledWith(
+            '/account/ecosystemAnonId',
+            sessionToken,
+            {
+              ecosystemAnonId,
+            },
+            new Headers({
+              'If-None-Match': '*',
+            })
+          );
+        });
+    });
+  });
 });
--- a/packages/fxa-auth-client/lib/client.ts
+++ b/packages/fxa-auth-client/lib/client.ts
@@ -178,6 +178,23 @@ export default class AuthClient {
     );
   }
 
+  private async sessionPut(
+    path: string,
+    sessionToken: string,
+    payload: object,
+    headers?: Headers
+  ) {
+    console.log('hey');
+    return this.hawkRequest(
+      'PUT',
+      path,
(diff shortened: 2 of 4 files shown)


packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -2838,6 +2838,36 @@ FxAccountClient.prototype.createSupportTicket = function (
...
+  return Promise.resolve()
+    .then(function () {
+      required(sessionToken, 'sessionToken');
+
+      return hawkCredentials(sessionToken, 'sessionToken', HKDF_SIZE);
+    })
+    .then(function (creds) {
+      // DISCUSS: the profile server endpoint utilizes the `If-None-Match: *` header/val
+      // to avoid overwriting an existing anon ID. Should this also use this approach?
+      // Guessing not if we're going to have a historical values table...
```
jodyheavener, line 2864: I intend on removing this comment, but wanted to discuss.

jodyheavener:
Why do we have an auth client inside the content server, and fxa-js-client? Is one meant to take over the other at some point?

dannycoates:
> Why do we have an auth client inside the content server, and fxa-js-client? Is one meant to take over the other at some point?

The one in content-server is new and meant to replace fxa-js-client, but there's some server tests that still use it because node doesn't have native WebCrypto. My hope is to write a small wrapper around it for node and replace fxa-js-client with the typescript version

jodyheavener:
> The one in content-server is new and meant to replace fxa-js-client

Ah okay. Worth adding new endpoints to fxa-js-client, or should I yank 'em?

packages/fxa-js-client/client/FxAccountClient.js
vbudhram: This part of AET is a bit fuzzy to me..curious what others think.

jodyheavener: Yeah I _think_ this was meant to server as a guard during the PoC stage, but now that we've bumped up the remaining work it seems to me that we'd want to make some changes to allow multiple IDs to be stored. It's not clear to me how we'll differentiate between racing clients versus intentionally updating; e.g. maybe we can use something like `If-Unmodified-Since`?

@6a68 whatchu think?

6a68: Yup, I think `if-unmodified-since` is the right approach: the goal here is just to make sure all clients use the same single anon_id value. In the case of conflicting updates, we can ask the client to upload what they think the current value is, and if that doesn't match the current value, reject the request with a 412 error.

jodyheavener: Since this will require a change in the auth and profile servers (to support if-unmodified-since), mind if I bump this specific change into a new issue? I imagine I can get it done this sprint.

jodyheavener: Got a little deep into this before realizing: we don't store the timestamp of the current anon ID, do we? How can we use if-unmodified-since?

jodyheavener: https://mozilla.slack.com/archives/G0177C5FTCZ/p1595349982175900

packages/fxa-js-client/client/FxAccountClient.js
```diff
@@ -2838,6 +2838,67 @@ FxAccountClient.prototype.createSupportTicket = function (
...
+ *   @param {String} [options.ifNoneMatch]
+ *   Set the If-None-Match header to the specified value.
+ *   Use '*' to only update ID if one is not already set.
+ *   @param {String} [options.ifMatch]
+ *   Set the If-Match header to the specified value.
+ *   Use the first part of an existing ID to only update
+ *   the ID if the values do not fuzzy match.
+ * @return {Promise} A promise that will be fulfilled with an empty response object
+ */
+FxAccountClient.prototype.updateEcosystemAnonId = function (
```
vbudhram, line 2857: I take it this part will get removed in Danny's PR?

jodyheavener: Mmm ya, I guess I can remove all this.

packages/fxa-content-server/app/tests/spec/lib/fxa-client.js
```diff
@@ -1884,4 +1884,23 @@ describe('lib/fxa-client', function () {
...
+    it('updates the ecosystem id', () => {
+      const sessionToken = 'the silence';
+      const ecosystemAnonId = 'the gold';
+      sinon.stub(realClient, 'updateEcosystemAnonId').resolves();
+
+      return client
+        .updateEcosystemAnonId(sessionToken, ecosystemAnonId)
+        .then((res) => {
+          assert.isTrue(
+            realClient.updateEcosystemAnonId.calledOnceWith(
```
vbudhram, line 1898: If you stub out the `realClient.sessionPut` would you be able to assert the correct path, method and headers?

jodyheavener: yeah, I'll have that updated shortly

PR #5981, opened 2020-07-17
feat(auth-server): enhance must-reset script to accept a plain text list

### LINKED ISSUE #5894: enhance must-reset.js to accept a plain text list of email addresses
When using this script, I typically have a list of email addresses which I have to resolve to associated uid with a database lookup and format into the JSON that this script accepts. It would be better if https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/scripts/must-reset.js could also accept the original list of email addresses, do the lookups, including secondary/primary, and do the reset.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2250)
┆Issue Number: FXA-2250

byrw:
## Because

* With a plain list of emails/uids it's easier to use the script.

## This pull request

* Add the option of running the script must-reset from a plain text
 list of uids or emails.
* Remove the option to read from a json.

fixes #5894

## Checklist

## Other information

This script from a secondary email can reset an account, even if the secondary email is not yet verified.

I've had to change `setTimeOut` value to make the tests pass locally, maybe this only happens with me.

When I run 'npm run test', why is the "script part" being skipped?

### DIFF
--- a/packages/fxa-auth-server/scripts/must-reset.js
+++ b/packages/fxa-auth-server/scripts/must-reset.js
@@ -8,7 +8,7 @@
 
  Usage:
 
- scripts/must-reset.js -i ./reset.json
+ scripts/must-reset.js -e -i ./reset.txt
 
  This script is used to put user accounts into a "must reset" state. It uses the
  same config file as key_server.js so should be run from a production instance.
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/scripts/must-reset.js
```diff
@@ -20,68 +20,64 @@
...
+  console.error('No `emails` or `uids` found inside the specified file');
+  process.exit(1);
+}
+
+function getItems(type) {
+  let input = '';
+  try {
+    input = fs
+      .readFileSync(path.resolve(commandLineOptions.input))
+      .toString('utf8');
```
bbangert, line 64: If the return statement from below was added after this, the input could be a const.

packages/fxa-auth-server/scripts/must-reset.js
```diff
@@ -20,68 +20,64 @@
...
-        console.error('failed', uid, err);
-        process.exit(1);
-      }
-    })
-  );
-
-  console.info('%s accounts reset', uids.length);
-  await db.close();
-  process.exit();
+function adjustText(input = '') {
```
bbangert, line 72: The function name is pretty ambiguous here, perhaps `splitCommaString`?

bbangert: `input` should also be more descriptive, perhaps `commaString` to demote its expected format?

packages/fxa-auth-server/scripts/must-reset.js
```diff
@@ -20,68 +20,64 @@
...
+  if (!input.length) {
+    return [];
+  }
+
+  const items = input.split(/[,\s]+/);
+  return items.filter((item) => !!item.trim().length);
```
bbangert, line 78: We'd probably want these trimmed, right?
```javascript
return input.split(/[,\s]+/).map(s => s.trim()).filter(s => !!s.length);
```

packages/fxa-auth-server/scripts/must-reset/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+module.exports = async function main(keys, dbFunction) {
+  const butil = require('../../lib/crypto/butil');
+  const config = require('../../config').getProperties();
+  const crypto = require('crypto');
+  const log = require('../../lib/log')({});
+  const P = require('../../lib/promise');
+  const Token = require('../../lib/tokens')(log, config);
+  const AuthDB = require('../../lib/db')(config, log, Token);
+  const oauth = require('../../lib/oauth/db');
```
bbangert, line 15: We typically put these above the export. Is there some other reason for putting the imports in the function?

bbangert:
Also, a rebase should hopefully fix the CI failures.

packages/fxa-auth-server/scripts/must-reset.js
byrw: This function also splits white space, maybe another name?

packages/fxa-auth-server/scripts/must-reset/index.js
byrw: I just implemented it the same way as dump-user.js

packages/fxa-auth-server/scripts/must-reset.js
byrw: changed

packages/fxa-auth-server/scripts/must-reset.js
bbangert: Ah, yes. How about a doc string on the function that indicates what kind of adjustment it does to the text?

packages/fxa-auth-server/scripts/must-reset/index.js
bbangert: Sigh, yup.

PR #5984, opened 2020-07-17
feat(auth-server): create a subs w/out a payment method in req

chenba:
Because:
 - the API should allow subscription creation with a payment method
   already on record for the customer

This commit:
 - make the payment method id in the create subscription route optional

Closes: FXA-2288

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -261,36 +261,41 @@ class StripeHelper {
   /**
    * Create a subscription for the provided customer.
    */
-  async createSubscriptionWithPMI(
-    customerId: string,
-    priceId: string,
-    paymentMethodId: string,
-    idempotencyKey: string
-  ) {
-    try {
-      await this.stripe.paymentMethods.attach(
-        paymentMethodId,
-        {
-          customer: customerId,
-        },
-        { idempotencyKey: `pma-${idempotencyKey}` }
-      );
-    } catch (err) {
-      if (err.type === 'StripeCardError') {
-        throw error.rejectedSubscriptionPaymentToken(err.message, err);
+  async createSubscriptionWithPMI(opts: {
+    customerId: string;
+    priceId: string;
+    paymentMethodId?: string;
+    subIdempotencyKey: string;
+  }) {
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -668,11 +669,17 @@ class DirectStripeRoutes {
...
+    if (!PMI) {
+      throw error.paymentFailed();
+    }
+
     const subIdempotencyKey = `${idempotencyKey}-createSub`;
     const subscription = await this.stripeHelper.createSubscriptionWithPMI(
       customer.id,
       priceId,
-      paymentMethodId,
+      PMI,
```
bbangert, line 682: This works? It looks like it can end up attempting to attach a source as a payment method. Should we just skip attaching the payment method in the `createSubscriptionWithPMI` call if we already have a source or payment method on file?

chenba: It works.  But you are probably right about the attaching the source as a payment method step being unnecessary.  I'll verify and make changes.

bbangert: I was thinking the payment method could be an optional arg. The easiest way to accommodate that would be turning the args into an object and making the payment method optional in TypeScript. We should really consider having a policy that as soon as a function passes 2 args, we should switch to an object. All the libraries typically do this, and it'd make it easier to avoid messing up positional placement of args.

chenba:
Simplified it to just making the Payment Method id optional.  If one is passed in, then it gets attached to the customer and saved into the default invoice settings.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -261,36 +261,41 @@ class StripeHelper {
...
-        },
-        { idempotencyKey: `pma-${idempotencyKey}` }
-      );
-    } catch (err) {
-      if (err.type === 'StripeCardError') {
-        throw error.rejectedSubscriptionPaymentToken(err.message, err);
+  async createSubscriptionWithPMI(opts: {
+    customerId: string;
+    priceId: string;
+    paymentMethodId: string;
```
bbangert, line 267: This should be marked optional with a `?`.

PR #5988, opened 2020-07-20
fix(style): add margin between buttons from account recovery.

### LINKED ISSUE #5986: Several buttons highlighters are incorrectly displayed
**Build**
Stage [Train 1.180.0]

**Affected platforms**
Windows 10 x64

**Steps to reproduce**
1. Launch Firefox and choose a Firefox user that has already a recovery key enabled.
2. Go to Sign In Page and tap on "Forget password?".
3. Insert the email and click on "Begin reset".
4. Go to email and tap on "Create new password".
5. Click on "Don't have a recovery key?" button.

**Expected results**
”Don't have a recovery key?” button is correctly displayed.

**Actual Results**
“Don't have a recovery key?” button highlighter is displayed over the “Confirm recovery key” button. 

**Additional notes**
Same issue is encountered for: 
 - “Remember password? Sign in” button from “Create new password” page. 
 - “Continue to my account” button from “Your password has been reset” page.
Screenshot_15
Screenshot_16
Screenshot_17

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2309)
┆Issue Number: FXA-2309

byrw:
Because:

* When highlighted the buttons were overlapping.

This commit:

* Add margin between the buttons from account recovery.

fixes #5986

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_settings-account-recovery.scss
+++ b/packages/fxa-content-server/app/styles/modules/_settings-account-recovery.scss
@@ -141,6 +141,10 @@
     font-size: 14px;
     line-height: 1.5;
   }
+
+  button.lost-recovery-key, button.remember-password{
+    margin-top: 16px;
+  }
 }
 
 button.button-link {
@@ -148,6 +152,6 @@ button.button-link {
   color: $color-blue;
 }
 
-button.generate-key-link{
+button.generate-key-link, button.btn-create-recovery-key{
   margin-bottom: 16px;
 }


byrw:
> Thanks for working on this! I just checked to see if we could maybe combine this style on the `button.button-link` class but there's a couple of cases where we wouldn't want this, which would make putting the style on that class a little messy. :(
> 
> I think this could perhaps be improved with a shared class, but at the very least, let's combine L145 and L149 to:
> 
> ```
> button.lost-recovery-key, button.remember-password {
>   margin-top: 16px;
> }
> ```
> 
> ... and then combine the styles for `.generate-key-link` and `btn-create-recovery-key` on L149 and L163 since they'd share `margin-bottom`. I'll approve this with that change. If you're interested in looking into #5659 which is related, it could maybe be an opportunity to clean this up a little, as "Don't have a recovery key?", "Remember password? Sign in", and "Continue to my account" should _probably_ be links (`<a href=...`), though I'd need to dive deeper into that issue to make sure we're not doing anything weird behind-the-scenes to warrant it being a button. I can check with our design team on the correct color/hover state there if you're interested, but no sweat on that issue otherwise.

Yes, I'm interested in working on this issue. Maybe a good solution for this would be simply reuse the existing implementation from the `.button-link` "Don't sync this device" link.

LZoog:
> Yes, I'm interested in working on this issue. Maybe a good solution for this would be simply reuse the existing implementation from the .button-link "Don't sync this device" link.

Ack, I meant to link #3951. If you'd still be interested feel free to grab it, and I don't actually think there are any outstanding questions for the design team.

Hmm yeah, since "Don't sync this device" is a link it shouldn't get `button.button-link` styles but I'd have to go pull that view up to see it - regardless I approved this PR and #3951 could be a good issue to explore how we can make shared styles around those buttons a bit cleaner so I'll merge this one in, thanks!

Issue #4320, opened 2020-02-28
Add eco_anon_id creation/storage code to Account model


data-sync-user:
This might be a small model or part of the account model that manages creating the anon_id from kB and the pipeline key, storing it locally, and exposing a function to check if the anon_id is present.

We can either persist the eco_anon_id on the account model or on a new one, which would be a new entry in localstorage. We do this for experiments, so we do have different areas in localstorage already.

Hydrating the model at load time may be a separate issue.

Edge case: if the password was reset, but the eco_anon_id wasn't uploaded after being generated. Less of a problem if we just always reset the anon_id at password time.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1225)
┆Issue Number: FXA-1225

PR #5989, opened 2020-07-20
feat(aet): Add account model methods to generate and store anon id

vbudhram:
Paired on this with @LZoog 

## Because

- We need to have methods that can update and store the ecosystem anon id for a user

## This pull request

- On loading the settings page, fetches and stores the anon id from the auth-server
- Adds account model methods to generate and store ecosystemAnonId to localstorage and auth-server
- Can generate `ecosystemAnonId` either from kB, using authentication token that can retrieve keys, or using password and sessionReauth

Closes: https://github.com/mozilla/fxa/issues/4320

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -18,12 +18,15 @@ import SignInReasons from '../lib/sign-in-reasons';
 import VerificationMethods from '../lib/verification-methods';
 import vat from '../lib/vat';
 import { emailsMatch } from 'fxa-shared/email/helpers';
+import VerificationReasons from '../lib/verification-reasons';
+import aet from '../lib/crypto/account-ecosystem-telemetry';
 
 // Account attributes that can be persisted
 const PERSISTENT = {
   accountResetToken: undefined,
   displayName: undefined,
   email: undefined,
+  ecosystemAnonId: undefined,
   grantedPermissions: undefined,
   hadProfileImageSetBefore: undefined,
   lastLogin: undefined,
@@ -58,6 +61,7 @@ const DEFAULTS = _.extend(
   {
     accessToken: undefined,
     declinedSyncEngines: undefined,
+    ecosystemAnonId: undefined,
     hasBounced: undefined,
     keyFetchToken: undefined,
     newsletters: undefined,
@@ -373,6 +377,91 @@ const Account = Backbone.Model.extend(
     },
 
     /**
+     * Attempt to generate and set the ecosystem anon. At a high level
+     * this function attempts to generate the id in the following manner:
+     *
+     * 1. Use kB if it is specified
+     * 2. Attempt to fetch keys if valid authentication token exists to do so
+     * 3. Attempt to use user's password to fetch keys
+     *
+     * @param {Object} [options={}]
+     *   @param {String} [options.kB] Optional The user's kB, if specified takes priority.
+     *   @param {String} [options.password] Optional The user's password
+     * @returns {Promise}
+     */
+    async generateEcosystemAnonId(options = {}) {
+      try {
+        let ecosystemAnonId;
+
+        if (!this._ecosystemAnonIdPublicKeys.length) {
+          return;
+        }
+
+        const randomIndex = Math.floor(
+          Math.random() * this._ecosystemAnonIdPublicKeys.length
+        );
+        const randomKey = this._ecosystemAnonIdPublicKeys[randomIndex];
+
+        if (options.kB) {
+          ecosystemAnonId = await aet.generateEcosystemAnonID(
+            this.get('uid'),
+            options.kB,
+            randomKey
+          );
+        } else if (this.canFetchKeys()) {
+          const keys = await this.accountKeys();
+          ecosystemAnonId = await aet.generateEcosystemAnonID(
+            this.get('uid'),
+            keys.kB,
+            randomKey
+          );
+        } else if (options.password) {
+          // TODO: Using sessionReauth to get keys could potentially pose some issues if the user
+          // is attempting to come from an unverified session. FxA only allows fetching keys from
+          // sessions that are verified (ex. performed some email verification loop).
+          const res = await this._fxaClient.sessionReauth(
+            this.get('sessionToken'),
+            this.get('email'),
+            options.password,
+            {
+              wantsKeys: () => true,
+              has: () => false,
+              isSync: () => false,
+            },
+            {
+              keys: true,
+              reason: VerificationReasons.ECOSYSTEM_ANON_ID,
+            }
+          );
+
+          if (res.keyFetchToken && res.unwrapBKey) {
+            const keys = await this._fxaClient.accountKeys(
+              res.keyFetchToken,
+              res.unwrapBKey
+            );
+            ecosystemAnonId = await aet.generateEcosystemAnonID(
+              this.get('uid'),
+              keys.kB,
+              randomKey
+            );
+          }
+        }
+
+        if (ecosystemAnonId) {
+          this.set('ecosystemAnonId', ecosystemAnonId);
+          await this._fxaClient.updateEcosystemAnonId(
+            this.get('sessionToken'),
+            ecosystemAnonId
+          );
+        }
+      } catch (err) {
+        // In the event of errors, we don't want to block any user actions, just log it
+        // TODO: Follow up with some better error handling
+        console.log('generateEcosystemAnonId failed...', err);
+      }
+    },
+
+    /**
      * Fetches account details from GET /account, for use by the settings views.
      * Caches its own result, because it's not intended for that endpoint to be
      * called repeatedly like some of the polling methods are. If you're really,
@@ -535,6 +624,7 @@ const Account = Backbone.Model.extend(
 
         this.setProfileImage(profileImage);
         this.set('displayName', result.displayName);
+        this.set('ecosystemAnonId', result.ecosystemAnonId);
 
         this.on('change', this._boundOnChange);
       });
--- a/packages/fxa-content-server/app/tests/spec/models/account.js
+++ b/packages/fxa-content-server/app/tests/spec/models/account.js
@@ -19,6 +19,7 @@ import SignInReasons from 'lib/sign-in-reasons';
 import sinon from 'sinon';
 import VerificationMethods from 'lib/verification-methods';
 import VerificationReasons from 'lib/verification-reasons';
+import aet from 'lib/crypto/account-ecosystem-telemetry';
 
 describe('models/account', function () {
   var account;
@@ -72,12 +73,12 @@ describe('models/account', function () {
     };
     ecosystemAnonIdPublicKeys = [
       {
-        'kty': 'EC',
+        kty: 'EC',
       },
       {
-        'kty': 'RSA',
+        kty: 'RSA',
       },
-    ]
+    ];
 
     account = new Account(
       {
@@ -93,7 +94,7 @@ describe('models/account', function () {
         profileClient: profileClient,
         sentryMetrics: sentryMetrics,
         subscriptionsConfig,
-        ecosystemAnonIdPublicKeys
+        ecosystemAnonIdPublicKeys,
       }
     );
   });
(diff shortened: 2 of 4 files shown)


packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+     * @returns {Promise}
+     */
+    async generateEcosystemAnonIdIfRequired(options = {}) {
+      try {
+        let ecosystemAnonId = this.get('ecosystemAnonId');
+        if (ecosystemAnonId) {
+          return;
+        }
+
+        // If no data pipeline keys are available, we can't generate the ecosystemAnonId, abort
```
6a68, line 433: This comment doesn't seem necessary to me, since it just describes what the code directly below it does, without adding context. I'd suggest removing it.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+        if (this._ecosystemAnonIdPublicKeys.length <= 0) {
+          return;
+        }
+
+        // Pick a random public key
```
6a68, line 438: Same feedback on this comment - when comments just restate what the code below does, the code and comment tend to diverge over time. Probably best to remove.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+        const randomEcosystemAnonIdPublicKey = this._ecosystemAnonIdPublicKeys[
+          Math.floor(Math.random() * this._ecosystemAnonIdPublicKeys.length)
+        ];
+
+        //
```
6a68, line 443: Stray empty comment here, please remove

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+     */
+    async generateEcosystemAnonIdIfRequired(options = {}) {
+      try {
+        let ecosystemAnonId = this.get('ecosystemAnonId');
+        if (ecosystemAnonId) {
+          return;
+        }
+
+        // If no data pipeline keys are available, we can't generate the ecosystemAnonId, abort
+        if (this._ecosystemAnonIdPublicKeys.length <= 0) {
```
6a68, line 434: Can we just do `if (!this._ecosystemAnonIdPublicKeys.length)` here? We use a non-strict falsy check on line 429 above. Let's be consistent either way. Also, I don't think JS arrays can have negative length.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+          return;
+        }
+
+        // Pick a random public key
+        const randomEcosystemAnonIdPublicKey = this._ecosystemAnonIdPublicKeys[
```
6a68, line 439: Might be more readable if the random index selection is pulled out into a variable:

```js
const randomIndex = Math.floor(Math.random() * this._ecosystemAnonIdPublicKeys.length);
const randomKey = this._ecosystemAnonIdPublicKeys[randomIndex]

File: packages/fxa-content-server/app/scripts/models/account.js [hunk 6]
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+     * 2. Attempt to fetch keys if valid authentication token exists to do so
+     * 3. Use user's password to fetch keys
+     *
+     *
+     * @param {Object} [options={}]
+     *   @param {String} [options.kB] Optional The user's kB, ff specified takes priority.
+     *   @param {String} [options.password] Optional The user's password
+     * @returns {Promise}
+     */
+    async generateEcosystemAnonIdIfRequired(options = {}) {
```
6a68, line 426: Hmm. Should I favor the `options` object pattern in the `generateEcosystemAnonId` method in my PR, rather than using two positional arguments?

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+        } else if (this.canFetchKeys()) {
+          const keys = await this.accountKeys();
+          ecosystemAnonId = await ecosystemAnonKeys.generateEcosystemAnonID(
+            keys.kB,
+            randomEcosystemAnonIdPublicKey
+          );
+        } else if (options.password) {
+          // TODO: Using sessionReauth to get keys could potentially pose some issues if the user
+          // is attempting to come from an unverified session. FxA only allows fetching keys from
+          // sessions that are verified (ex. performed some email verification loop).
```
6a68, line 458: This seems like an important TODO. How should we handle it? Maybe file a followup issue to discuss?

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+            options.password,
+            {
+              wantsKeys: () => true, // Ensure keys are requested
+              has: () => false,
+              isSync: () => false,
+            },
+            {
+              keys: true,
+
+              // TODO: The verification reason needs to be updated to be specific for ecosystemAnonId
```
6a68, line 471: Good catch, I think we need a followup bug for this one, too?

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+
+        // Update localstorage and report new ecosystemAnonId to auth-server
+        if (ecosystemAnonId) {
+          this.set('ecosystemAnonId', ecosystemAnonId);
+
+          await this._fxaClient.updateEcosystemAnonId(
+            this.get('sessionToken'),
+
+            // TODO: We need to ensure that the ecosystemAnonId is stored and updated in correct format.
+            // ie. hex or base64
```
6a68, line 496: Hex format is only needed for submission to the data pipeline. Everywhere else we can stick with the serialization suggested in the ['Anonymizing User Identifiers' AET doc](https://docs.google.com/document/d/1zH3eVVI_28Afg1JXe_McDrW4MTYuWhiJMQR6AQbli8I/edit), which is the [JWE compact serialization](https://tools.ietf.org/html/rfc7516#section-7.1) in url-safe base64. I'll add a note to the new bug around submitting to the data pipeline, and I think we're good to delete this TODO :+1:

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,128 @@ const Account = Backbone.Model.extend(
...
+            ecosystemAnonId
+          );
+        }
+      } catch (err) {
+        // In the event of errors, we don't want to block any user actions, just log it
```
6a68, line 501: If we get a 412 error, then we tried to update the anon_id, but another client beat us to it. We should handle that error by clearing the newly-created value out of localstorage and refetching the new value from the server. OK to handle this in a followup bug.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,127 @@ const Account = Backbone.Model.extend(
...
+              res.unwrapBKey
+            );
+            ecosystemAnonId = await ecosystemAnonKeys.generateEcosystemAnonID(
+              keys.kB,
+              randomEcosystemAnonIdPublicKey
+            );
+          }
+        }
+
+        // Update localstorage and report new ecosystemAnonId to auth-server
```
6a68, line 489: Sorry to be a bit nitpicky here, but this is another comment that doesn't add to the meaning of the code below it. I think it'd be good to remove this one, too.

packages/fxa-content-server/app/scripts/models/account.js
6a68: In general, I think if we get a 4xx or 5xx error here, we probably need to discard the new ecosystemAnonId value until we can phone home successfully.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,127 @@ const Account = Backbone.Model.extend(
...
+    _fetchAccountProfileResult: null,
+
+    /**
+     * Fetches the account profile from the GET /account/profile on the auth server and caches the
+     * result locally in _accountProfileFetchPromise. This also updates the account model
+     * with the new properties (minus any verificationMethod, verificationReason values)
+     *
+     * @returns {Promise}
+     */
+    async fetchAccountProfile() {
```
6a68, line 394: I'm confused about the need for this function. It seems like the existing `fetchProfile` method should pull back the ecosystem_anon_id value along with other profile data from the profile server. Could we just set the value there?

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -372,6 +382,127 @@ const Account = Backbone.Model.extend(
...
+     *   @param {String} [options.kB] Optional The user's kB, ff specified takes priority.
+     *   @param {String} [options.password] Optional The user's password
+     * @returns {Promise}
+     */
+    async generateEcosystemAnonIdIfRequired(options = {}) {
+      try {
+        let ecosystemAnonId = this.get('ecosystemAnonId');
+        if (ecosystemAnonId) {
+          return;
+        }
```
LZoog, line 431: If I'm understanding that Dave will be using this method in various views, do we want to allow for updating the ID when the user changes their password, in which case I think this could be `if (ecosystemAnonId && !options.kB)`?

packages/fxa-content-server/app/scripts/models/account.js
vbudhram: Ah I didn't noticed that this value was returned there. I can remove this bit then.

packages/fxa-content-server/app/scripts/models/account.js
vbudhram: Yea these should be removed. They were placeholders before I wrote code to make sure I understood each step of the process.

packages/fxa-content-server/app/scripts/models/account.js
vbudhram: I think its fine to keep yours as is, both your fields are required.

packages/fxa-content-server/app/scripts/models/account.js
vbudhram: Hrm, I wonder what value the `IfRequired` part gives us now. I'm going to drop that part and remove this check since we can't guarantee saving it.

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -72,12 +72,12 @@ describe('models/account', function () {
     };
     ecosystemAnonIdPublicKeys = [
       {
-        'kty': 'EC',
+        kty: 'EC',
       },
       {
-        'kty': 'RSA',
+        kty: 'RSA',
```
LZoog, line 78: I made this key `RSA` so there'd be difference between the two objects in case there was anything worth testing about picking one of the two keys but eh, don't think we need to and it looks like we'll only ever use `EC` so we _could_ update it to more accurately mock what these will be but it also doesn't really matter. 🤷

packages/fxa-content-server/app/scripts/models/account.js
LZoog: Filed new issue here: https://github.com/mozilla/fxa/issues/6017

packages/fxa-content-server/app/scripts/models/account.js
LZoog: Filed an issue here: https://github.com/mozilla/fxa/issues/6018

packages/fxa-content-server/app/tests/spec/models/account.js
```diff
@@ -3110,4 +3103,111 @@ describe('models/account', function () {
...
     });
   });
+
+  describe('generateEcosystemAnonId', () => {
+    const sandbox = sinon.createSandbox();
+
+    beforeEach(() => {
+      sandbox.stub(fxaClient, 'updateEcosystemAnonId').resolves();
+
+      // TODO: Update once #5994 lands
```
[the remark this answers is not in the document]
vbudhram: Looks like there are still some tweaks need on that PR, I'll probably merge this after ci passes.

PR #5993, opened 2020-07-20
Unified fxa-auth-client

dannycoates:
## Because

- We have two clients for the auth-server API, one in `content-server/app/scripts/lib/auth` and `fxa-js-client`. The content-server one was meant to replace fxa-js-client but I didn't finish the work to get it running on nodejs.

## This pull request

- Extracts the content-server implementation into `fxa-auth-client`
- Wraps the implementation with shims to work with nodejs in `server.ts`
- Updates references to fxa-js-client to use fxa-auth-client
- Removes fxa-js-client


packages/fxa-graphql-api/src/bin/main.ts
```diff
@@ -4,7 +4,7 @@
 import 'reflect-metadata';
 
 import express from 'express';
-import FxAccountClient from 'fxa-js-client';
+import AuthClient from 'fxa-auth-client';
```
dannycoates, line 7: @bbangert -- just a ping for this change

6a68:
@dannycoates Saw this go by on Slack, but double reminder: need to pull Jody's changes into this before landing, or vice versa

PR #5994, opened 2020-07-20
feat(AET): Generate the ecosystem_anon_id on the client

6a68:
First attempt at an account model function to generate and return the eco_anon_id.

Comments welcome :+1:

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/crypto/account-ecosystem-telemetry.js
+++ b/packages/fxa-content-server/app/scripts/lib/crypto/account-ecosystem-telemetry.js
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Crypto methods used for Account Ecosystem Telemetry.
+ */
+
+import importFxaCryptoDeriver from './deriver';
+import required from '../required';
+import ScopedKeys from './scoped-keys';
+import util from './util';
+
+/**
+ * Given a user's `kB` and a pipeline public key, derive the user's
+ * ecosystem_user_id and encrypt it with the pipeline key, yielding the
+ * ecosystem_anon_id. Return a promise that resolves to the ecosystem_anon_id.
+ *
+ * @param {String} uid - user's UID
+ * @param {String} kB - user's kB
+ * @param {String} pipelineKey - a pipeline public key in JWK format
+ * @returns {Promise} A promise that will be fulfilled with the user's
+ * ecosystem_anon_id in JWK format
+ */
+const generateEcosystemAnonID = async (uid, kB, pipelineKey) => {
+  required(uid, 'uid');
+  required(kB, 'kB');
+  required(pipelineKey, 'pipelineKey');
+
+  const fxaCryptoDeriver = await importFxaCryptoDeriver();
+  const fxaDeriverUtils = new fxaCryptoDeriver.DeriverUtils();
+
+  const keyData = {
+    identifier: 'https://identity.mozilla.com/ids/ecosystem_telemetry',
+    // Following fxa-auth-server/lib/oauth/routes/key_data.js#L14-L20, FxA
+    // doesn't yet support per-user scoped key rotation; use a block of
+    // zeros instead.
+    keyRotationSecret:
+      '0000000000000000000000000000000000000000000000000000000000000000',
+    keyRotationTimestamp: 1595287165358,
+  };
+
+  const ecosystemUserJWK = await ScopedKeys.deriveScopedKeys(kB, uid, keyData);
+
+  // Pull out just the 32 bytes used as eco_uid from the scoped key, then
+  // transform from urlsafe base64 to hex encoding, because the data pipeline
+  // expects the identifier in hex.
+  const ecosystemUserIDRaw = util.base64urlToUint8Array(ecosystemUserJWK.k);
+  const ecosystemUserID = util.uint8ToHex(ecosystemUserIDRaw);
+
+  // The node-jose code called by the deriver expects a base64-encoded key.
+  const encodedKey = fxaCryptoDeriver.jose.util.base64url.encode(
+    JSON.stringify(pipelineKey),
+    'utf8'
+  );
+
+  const ecosystemAnonID = await fxaDeriverUtils.encryptBundle(
+    encodedKey,
+    ecosystemUserID
+  );
+
+  return ecosystemAnonID;
+};
+
+export default {
+  generateEcosystemAnonID,
+};
--- a/packages/fxa-content-server/app/scripts/lib/crypto/scoped-keys.js
+++ b/packages/fxa-content-server/app/scripts/lib/crypto/scoped-keys.js
@@ -18,7 +18,7 @@ import required from '../required';
  * @returns {Promise} A promise that will resolve with an object having a scoped key
  *   The key is represented as a JWK object.
  */
-function _deriveScopedKeys(inputKey, uid, keyData) {
+function deriveScopedKeys(inputKey, uid, keyData) {
   return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
     required(inputKey, 'input key');
     required(uid, 'uid');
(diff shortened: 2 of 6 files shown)


packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -293,6 +295,37 @@ const Account = Backbone.Model.extend(
...
+     * in JWK format
+     * @returns {Promise} A promise that will be fulfilled with the user's
+     * ecosystem_anon_id in JWK format
+     */
+    async generateEcosystemAnonID(kB, pipelineKey) {
+      const uid = this.get('uid');
+      const keyData = {
+        // This identifier comes from the 'Account Ecosysten Telemetry
+        // Identifiers' google doc.
+        identifier: 'https://identity.mozilla.com/ids/ecosystem_telemetry',
```
rfk, line 315: BTW, I made this up more-or-less arbitrarily, bikeshedding most welcome on the identifier here; /cc @vladikoff

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -293,6 +295,37 @@ const Account = Backbone.Model.extend(
...
+        // Following fxa-auth-server/lib/oauth/routes/key_data.js#L14-L20, FxA
+        // doesn't yet support per-user scoped key rotation; use a block of
+        // zeros instead.
+        keyRotationSecret:
+          '0000000000000000000000000000000000000000000000000000000000000000',
+        keyRotationTimestamp: 1595287165358,
+      };
```
rfk, line 322: Ideally you would obtain these values from the server, but as a starting point that doesn't seem worth an extra request. I also expect that you don't actually use the `keyRotationTimestamp` value in this specific case, and it may be worth setting it to an obviously-fake value like `1000000000000`.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -293,6 +295,37 @@ const Account = Backbone.Model.extend(
...
+
+      const ecosystemUserId = await _deriveScopedKeys(kB, uid, keyData);
```
rfk, line 324: I expect that `derivedScopedKeys` here will be returning a full JWK object with `kid`, `crv` etc properties. You want *just* the `k` property which is the underlying key bytes. (Also I think the `k` property is a base64 string in this case, but the data pipeline wants it as a hex string).

vbudhram: > data pipeline wants it as a hex string

Oh, I thought it was base64 everywhere. @6a68 do we need to do any coordination with profile-server/auth-server to ensure we are saving in correct format?

packages/fxa-content-server/app/scripts/models/account.js
6a68: Interestingly it does get passed into the HMAC call through [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/lib/crypto/scoped-keys.js#L32-L33), we just don't do anything to vary it between users

6a68: The other occurrence of this `keyRotationTimestamp` has a millisecond timestamp dating to when the library was first written back in 2017, which is why I went with yesterday's date for the timestamp 🤷

packages/fxa-content-server/app/scripts/models/account.js
6a68: Good catch, @rfk! And good question @vbudhram, I believe we treat it as an opaque string without enforcing any particular encoding, so I think it should be encoding agnostic--but we should double check.

6a68: Following up on this base64/hex issue, we use the JWK compact serialization everywhere, which is urlsafe base64. The server endpoints just treat it as an opaque string. I've called out the need for conversion to hex in the data pipeline submission issue (#5998).

packages/fxa-content-server/app/scripts/models/account.js
6a68: Thoughts after staring at this for a bit:

I don't know if we'll put anything else under the `/ids/` namespace, but AET probably should stand alone in this sense.

I did notice that we have a dash-delimited URI in `https://identity.mozilla.com/cmd/open-uri`, so maybe for consistency, this scope should be `https://identity.mozilla.com/ids/ecosystem-telemetry`?

packages/fxa-content-server/app/scripts/models/account.js
rfk: To be clear, I'm talking specifically about the value *inside* the JWE here (that is, the `eco_uid`) which is not affected by the choice of encoding or formatting or the JWE wrapper. The data pipeline will decrypt the JWE and validate the contained value, rejecting any values that are not [a 64-character hex string](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/490cfe00c401e079277c00715d8dda55b7cd81cd/schemas/firefox-accounts/account-ecosystem/account-ecosystem.1.schema.json#L21).

rfk: To be even more explicit, I expect this code will need to look something like:

```
ecosystemUserJWK = await _deriveScopedKeys(kB, uid, keyData);
ecosystemUserId = encode_to_hex(decode_from_base64(ecosystemUserJWK.k));
ecosystemAnonId = await encrypt(ecosystemUserId, pipelineKey);
```

(Alternately we could decide that we want `ecosystemUserId` to be a base64 string everywhere and update the pipeline schema to match, but hex identifiers seem more common in other parts of telemetry)

6a68: Oh, of course, thanks for pointing that out @rfk. The value has to be encoded correctly _before_ it's encrypted using the pipeline key, since the pipeline will decrypt and then need to validate what it's decrypted.

I'll update my patch accordingly...and drop the incorrect comment from the metrics bug

6a68:
@rfk Think I've now gotten the eco_uid encoding correctly adjusted, thanks again for catching that. Not sure what to do about the 'unsupported algorithm' error I'm seeing in the tests when trying to encrypt with the test public pipeline key--any thoughts on the right way to go about fixing that?

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -293,6 +295,37 @@ const Account = Backbone.Model.extend(
...
+     * pipeline key, yielding the ecosystem_anon_id. Return the
+     * ecosystem_anon_id.
+     *
+     * @param {String} kB - user's kB
+     * @param {String} pipelineKey - a randomly-chosen pipeline public key
+     * in JWK format
+     * @returns {Promise} A promise that will be fulfilled with the user's
+     * ecosystem_anon_id in JWK format
+     */
+    async generateEcosystemAnonID(kB, pipelineKey) {
```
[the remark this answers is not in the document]
6a68: didn't quite get to typescript with time constraints this week, but did extract it out into a lib/crypto file :+1:

rfk:
>  Not sure what to do about the 'unsupported algorithm' error I'm seeing in the tests when trying
> to encrypt with the test public pipeline key--any thoughts on the right way to go about fixing that?

Oh, boo 😢. So the code in `a256gcm.js` only supports generating JWEs using the "direct" encryption algorithm with a symmetric key, not the public-key crypto we need here.

I believe you want the functionality used by `crypto/scoped-keys.js:createEncryptedBundle`, which produces JWEs using the "ecdh-es" algorithm (specifically `fxaDeriverUtils.encryptBundle`, which it might make sense to pull up into a top-level helper function).

6a68:
I think we're pretty much done here, although I do need to figure out exactly how to verify the eco_anon_id when it's fully baked. Glad node-jose is exposed by the `fxaCryptoDeriver`, because the underlying encryption code was not happy about a key getting passed in as an object, rather than as serialized base64 :+1:

6a68:
Looks like a spurious failure in test-content-server-5: 

```
× firefox on linux 4.15.0-1077-aws - signup with code - invalid code (3.311s)
 
    AssertionError: expected '' to include 'invalid or expired verification code'
```

Squashed and force-pushed; hopefully we'll get a clean test run.

6a68:
OK, going to land this 🤞

PR #5996, opened 2020-07-21
feat(AET): Pass ecosystem anon ID public keys through account model and expose to client

LZoog:
## Because
* In the account model, we need to generate the ecosystem_anon_id with one of these keys plus the user's kB.

## This pull request

* Adds 'ecosystemAnonIdPublicKeys' to 'configForFrontEnd'
* Passes this value into the user model and then account model

Partial fix for #4320 but does not close.

## Checklist

Looks like we test for this value in the FE config in `get-index` but not in `app-start`, and also only testing values in `user.js` we're actually using there so I replicated those tests. The one in `models/account` doesn't show much but the assertion can happen when we work on tests for `generateEcosystemAnonIdIfRequired` in #5989.

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/app-start.js
+++ b/packages/fxa-content-server/app/scripts/lib/app-start.js
@@ -371,6 +371,7 @@ Start.prototype = {
         profileClient: this._profileClient,
         sentryMetrics: this._sentryMetrics,
         subscriptionsConfig: this._config.subscriptions,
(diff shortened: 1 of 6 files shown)


LZoog:
> LGTM, good to merge. We can add the extra validation here or in another PR, your call 🤷

Hmm yeah I'll do this in a follow up since this passed CI, Vijay needs these changes, and all these changes are in the content-server.

PR #6007, opened 2020-07-22
feat(payments): feature flag support to select SCA or legacy payment UI as default

### LINKED ISSUE #5970: Implement feature flag support for the new SCA payment features
We can deploy the new SCA payment features in parallel with the pre-SCA payment features.

Accessing the new stuff involves using a new `/v2` route in the payment-server front-end.

We would like a way to easily flip between the pre-SCA scenario over to the new post-SCA scenario. This could help us have a roll-back option in case the new approach surfaces issues after being open to the public.

Might also be nice to have a QA-accessible way to flip back and forth in a single testing session - i.e. maybe a cookie or a localStorage flag rather than a site config setting that takes a re-deploy to change.

This is in advance of FXA-2275 and FXA-2229 - presumably we'll have tested & QA'ed & proven out the new routes via feature flipping before finally pulling out the old stuff.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2301)
┆Issue Number: FXA-2301

lmorchard:
- support for FEATURE_USE_SCA_PAYMENT_UI_BY_DEFAULT=1 env var to select
  SCA as default

- also adds support for /v1 and /v2 URL paths to select legacy or SCA
  payment UI, respectively

- tests to exercise App router against paths with mock routes

fixes #5970

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -12,6 +12,15 @@ convict.addFormats(require('convict-format-with-moment'));
 convict.addFormats(require('convict-format-with-validator'));
 
 const conf = convict({
+  featureFlags: {
+    useSCAPaymentUIByDefault: {
+      default: false,
+      doc:
+        'Whether to use newer SCA payment UI components by default rather than at the /v2 sub-path',
+      env: 'FEATURE_USE_SCA_PAYMENT_UI_BY_DEFAULT',
+      format: Boolean,
+    },
+  },
   amplitude: {
     enabled: {
       default: true,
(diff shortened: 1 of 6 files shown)


lmorchard:
I *think* this is all set for review? Ended up going down a testing rabbit hole since we had no coverage at all for the App component. Let's see what CI does and I'll sleep on it in the meantime.

lmorchard:
Futzed around with a few last bits of test coverage - technically 100% on App component now! Adding some review requests.

lmorchard:
@jackiemunroe FYI, this overlaps with FXA-2215 a bit, since the v1 / v2 routing setup is reorganized here to support the feature flag

Issue #6006, opened 2020-07-22
New emailRecord_5 stored procedure is much slower than emailRecord_4


jbuck:
When @jrgm deployed Train 180 to production we encountered 100% CPU on the RDS database immediately. After rolling back to Train 179 he had to manually kill some queries that were still running after the instances were terminated. After restoring service, we looked at Performance Insights and found significant waits and 30M rows/second being queried, which smelled like a query doing table scans instead of using an index. When @jrgm and I manually ran `CALL emailRecord('jbuckley@mozilla.com');` we found that it took >5 seconds to return results. When running the query inside the stored procedure it ran fine. So, something about the stored procedure was de-optimizing the query. @bbangert suggested duplicating `emailRecord_5` and adding an `EXPLAIN` to the `SELECT` statement to verify this behaviour and we found that it was indeed doing a table scan instead of using the index:

```
MySQL [fxa]> CALL emailRecord_5_explain('jbuckley@mozilla.com');
+----+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
| id | select_type | table | type | possible_keys | key  | key_len | ref  | rows      | Extra       |
+----+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
|  1 | SIMPLE      | a     | ALL  | NULL          | NULL | NULL    | NULL | 116557655 | Using where |
+----+-------------+-------+------+---------------+------+---------+------+-----------+-------------+
1 row in set (0.01 sec)
```

After some googling, I found [a blog post from 2013](https://lowleveldesign.org/2013/07/19/diagnosing-collation-issue-mysql-stored-procedure/) saying that procedure character sets and table index character sets need to match otherwise it'll be a table scan.

My suggested fix is that we should drop stored procedures like they're hot, and use queries like we do everywhere else.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2329)
┆Issue Number: FXA-2329

rfk:
For completeness, we do have a linting tool that complains about this issue:

```
$> npm run migration-lint

> fxa-auth-db-mysql@1.180.0 migration-lint /Users/rfk/repos/mozilla/fxa/packages/fxa-auth-db-mysql
> node scripts/migration-lint

Warning: expected "utf8 COLLATE utf8_bin" for inNormalizedEmail in createAccount_8!
IN `inNormalizedEmail` VARCHAR(255),

[...]

PR #6012, opened 2020-07-22
fix(db): Set collation on stored procedure email fields

6a68:
## Because

- MySQL doesn't use indexes on stored procedure queries if the collation of the field doesn't explicitly match that of the table.

## This pull request

- explicitly sets collation on email fields used to search the email table

- also removes the ecosystem_anon_id from the create account stored procedure, because in general we will never have that data at account creation time

Closes: #6006

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-111-112.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-111-112.sql
@@ -1,0 +1,121 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('111');
+
+-- Explicitly set the collation on the email field, so that mysql will use
+-- the email table index and not do a full table scan.
+-- See https://github.com/mozilla/fxa/issues/6006.
+CREATE PROCEDURE `emailRecord_6` (
+    IN `inEmail` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin
+)
+BEGIN
+    SELECT
+        a.uid,
+        a.email,
+        a.normalizedEmail,
+        a.emailVerified,
+        a.emailCode,
+        a.kA,
+        a.wrapWrapKb,
+        a.verifierVersion,
+        a.authSalt,
+        a.verifierSetAt,
+        a.lockedAt,
+        a.createdAt
+    FROM
+        accounts a
+    WHERE
+        a.normalizedEmail = LOWER(inEmail)
+    ;
+END;
+
+-- Remove the eco_anon_id field from createAccount, which we really don't ever
+-- plan to use in practice. Also explicitly set collation on the normalized
+-- email field, so that mysql will use the index and not do a full table read.
+-- See https://github.com/mozilla/fxa/issues/6006.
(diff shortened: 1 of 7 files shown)


dannycoates:
Since these don't do anything with ecosystemAnonId can't we just use the previous versions?

6a68:
> Since these don't do anything with ecosystemAnonId can't we just use the previous versions?

Yeah, we could. I wasn't sure if we might introduce confusion by doing so. I also didn't see the explicit collation in the previous versions of these stored procedures 🤷‍♂

packages/fxa-auth-db-mysql/lib/db/schema/patch-111-112.sql
```diff
@@ -0,0 +1,121 @@
...
+    IN `inEmailCode` BINARY(16),
+    IN `inEmailVerified` TINYINT(1),
+    IN `inKA` BINARY(32),
+    IN `inWrapWrapKb` BINARY(32),
+    IN `inAuthSalt` BINARY(32),
+    IN `inVerifierVersion` TINYINT UNSIGNED,
+    IN `inVerifyHash` BINARY(32),
+    IN `inVerifierSetAt` BIGINT UNSIGNED,
+    IN `inCreatedAt` BIGINT UNSIGNED,
+    IN `inLocale` VARCHAR(255),
```
jrgm, line 49: trying this on stage db and it whines about the stuff below, but I think it really means it doesn't like the trailing comma above.

dannycoates:
> Yeah, we could. I wasn't sure if we might introduce confusion by doing so

Maybe we add a migration to drop the bad procs as documentation?

packages/fxa-auth-db-mysql/lib/db/schema/patch-111-112.sql
```diff
@@ -0,0 +1,121 @@
...
+END;
+
+-- Remove the eco_anon_id field from createAccount, which we really don't ever
+-- plan to use in practice. Also explicitly set collation on the normalized
+-- email field, so that mysql will use the index and not do a full table read.
+-- See https://github.com/mozilla/fxa/issues/6006.
+CREATE PROCEDURE `createAccount_9`(
+    IN `inUid` BINARY(16) ,
+    IN `inNormalizedEmail` VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin,
+    IN `inEmail` VARCHAR(255),
```
jrgm, line 39: Noting Ryan's comment in https://github.com/mozilla/fxa/issues/6006#issuecomment-662201326
```
Warning: expected "utf8 COLLATE utf8_bin" for inEmail in createAccount_8!
IN `inEmail` VARCHAR(255),
```

6a68:
Looks like a couple of spurious content-server failures; restarted those.

6a68:
ugh, switching branches brought over a ton of other commits. let me do a little git surgery to clean this up...

6a68:
Looks like the failures may not be meaningful:

auth-server:

Screen Shot 2020-07-22 at 5 55 05 PM

content-server:

Screen Shot 2020-07-22 at 5 56 23 PM

jrgm:
all green now

PR #6014, opened 2020-07-22
feat(aet): support for if-match and if-none-match across the board

jodyheavener:
## Because

- Services internal and external to FxA need to be able to set an the Ecosystem Anon ID under various circumstances, such as only if one isn't set, or if it matches a specific existing value

## This pull request

- Improves [`If-None-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) and adds [`If-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match) header support across the Profile server, Auth Server, and Auth Client

## Checklist

For both implementations I tried to follow as close to spec as possible, which includes support for one or more specific values (comma-delimited), and `*` (If-None-Match only). Comparisons are made using an MD5 hashed version of the existing anon ID.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -15,6 +15,7 @@ const uuid = require('uuid');
 const validators = require('./validators');
 const authMethods = require('../authMethods');
 const ScopeSet = require('fxa-shared').oauth.scopes;
+const crypto = require('crypto');
 
 const { determineSubscriptionCapabilities } = require('./utils/subscriptions');
 
@@ -1494,7 +1495,7 @@ module.exports = (
       method: 'PUT',
       path: '/account/ecosystemAnonId',
       apidoc: {
-        errors: [error.invalidScopes, error.anonIdExists],
+        errors: [error.invalidScopes, error.anonIdUpdateConflict],
       },
       options: {
         auth: {
@@ -1512,8 +1513,15 @@ module.exports = (
 
         const { uid, scope } = request.auth.credentials;
         const { ecosystemAnonId } = request.payload;
-        const noneMatchHeader = request.headers['If-None-Match'];
         const scopeSet = ScopeSet.fromArray(scope);
+
+        function hashAnonId(anonId) {
+          const hash = crypto.createHash('sha256');
+          return hash.update(anonId).digest('hex');
+        }
+
+        const ifNoneMatch = request.headers['If-None-Match'];
+        const ifMatch = request.headers['If-Match'];
 
         if (!scopeSet.contains('profile:ecosystem_anon_id:write')) {
           throw error.invalidScopes(scopeSet);
@@ -1521,10 +1529,21 @@ module.exports = (
 
         await customs.check(request, uid, 'updateEcosystemAnonId');
 
-        if (noneMatchHeader === '*') {
-          const account = await db.account();
-          if (account.ecosystemAnonId != null) {
-            throw error.anonIdExists();
+        if (ifNoneMatch || ifMatch) {
+          const { ecosystemAnonId: existingAnonId } = await db.account();
+
+          if (existingAnonId) {
+            const hashedAnonId = hashAnonId(existingAnonId);
+
+            if (ifMatch && ifMatch !== hashedAnonId) {
+              throw error.anonIdUpdateConflict('If-Match');
+            }
+
+            if (ifNoneMatch === '*') {
+              throw error.anonIdUpdateConflict('If-None-Match');
+            } else if (ifNoneMatch === hashedAnonId) {
+              throw error.anonIdUpdateConflict('If-None-Match');
+            }
           }
         }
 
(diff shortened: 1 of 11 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1512,48 @@ module.exports = (
...
+          // Note the checks for both headers will accept a partial match
+          // So you can send as much or as little of the existing anon ID
+          // to compare against as you want.
+
+          if (existingAnonId != null && existingAnonId.length) {
+            if (ifNoneMatch.includes('*')) {
+              throw error.anonIdUpdateConflict('If-None-Match');
+            } else if (
+              ifNoneMatch.length &&
+              ifNoneMatch.some((value) => existingAnonId.includes(value))
```
dannycoates, line 1546: I thought we were going to do a proper content hash as the key/etag/whatever?

jodyheavener: 😑 Yup. Slack is hard.

jodyheavener:
Not sure if I did the hashing properly, but here we go!

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1513,51 @@ module.exports = (
...
-          const account = await db.account();
-          if (account.ecosystemAnonId != null) {
-            throw error.anonIdExists();
+        if (ifNoneMatch.length || ifMatch.length) {
+          const { ecosystemAnonId: existingAnonId } = await db.account();
+
+          if (existingAnonId != null && existingAnonId.length) {
+            const hashedAnonId = hashAnonId(existingAnonId);
+
+            if (ifNoneMatch.includes('*')) {
```
6a68, line 1546: I'd evaluate if-match headers first, following the precedence rules described in the [HTTP spec](https://tools.ietf.org/html/rfc7232#section-6).

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1513,51 @@ module.exports = (
 
         const { uid, scope } = request.auth.credentials;
         const { ecosystemAnonId } = request.payload;
-        const noneMatchHeader = request.headers['If-None-Match'];
         const scopeSet = ScopeSet.fromArray(scope);
 
+        function parseMultiValueHeader(named) {
```
6a68, line 1518: Not in keeping with the HTTP spec, by contrast, is there any reason to let people submit more than one value for this header? I can't think of a scenario where a client would have more than one value for the old anon_id, and just having '*' or a single hash to evaluate would simplify the logic here.

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1513,51 @@ module.exports = (
...
+          return value.toString().split(/,\s?/);
+        }
+
+        function hashAnonId(anonId) {
+          const hash = crypto.createHash('md5');
+          return hash.update(anonId).digest('hex');
+        }
+
+        const ifNoneMatch = parseMultiValueHeader('If-None-Match');
+        const ifMatch = parseMultiValueHeader('If-Match');
```
6a68, line 1532: I could see a human error case where new clients unintentionally submit the whole anon_id as the header value. What would you think about throwing if the header is longer than the 32 hex chars needed for an md5 hash?

packages/fxa-auth-server/lib/routes/account.js
6a68: After talking this over, I think this was an unnecessarily paranoid suggestion :-)

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1513,51 @@ module.exports = (
...
+        function parseMultiValueHeader(named) {
+          const value = request.headers[named];
+          if (!value || !value.length) {
+            return [];
+          }
+          return value.toString().split(/,\s?/);
+        }
+
+        function hashAnonId(anonId) {
+          const hash = crypto.createHash('md5');
```
dannycoates, line 1527: I'd like us to standardize on `sha-256` for hashes. For a few reasons, but the one on the top of my head is that we can do native sha in the browser

jodyheavener:
@6a68 @dannycoates pending tests, how does this look?

packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1512,19 +1513,40 @@ module.exports = (
 
         const { uid, scope } = request.auth.credentials;
         const { ecosystemAnonId } = request.payload;
-        const noneMatchHeader = request.headers['If-None-Match'];
         const scopeSet = ScopeSet.fromArray(scope);
 
+        function hashAnonId(anonId) {
+          const hash = crypto.createHmac(
```
dannycoates, line 1519: No need for mac, just hash is preferable https://nodejs.org/dist/latest-v12.x/docs/api/crypto.html#crypto_class_hash

PR #6020, opened 2020-07-23
feat(fxa-content-server): AET adding calls to check and generate

### LINKED ISSUE #4318: Check if anon_id exists on password entry and update
This is a small change in the view layer (cached credentials mixin), to check if the anon_id exists on the account model. If it doesn't exist, ask the model to generate it.

Note: we may opportunistically want to rotate the anon_id every time the user enters password, not just at password reset.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1227)
┆Issue Number: FXA-1227

meandavejustice:
#6004 without the capability check on reliers 

fixes #4318 

cc/ @vbudhram 

Should be merged after #5989 (for `account.generateEcosystemId` method)

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -1079,7 +1079,10 @@ const Account = Backbone.Model.extend(
             relier
           );
         })
-        .then(this.set.bind(this));
+        .then(this.set.bind(this))
+        .then(async () => {
+          await this.generateEcosystemAnonId({ password: newPassword });
+        });
     },
 
     /**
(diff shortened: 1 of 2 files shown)


meandavejustice:
If we prefer to go this route, I will squash this into one commit, close out #6004, and file an issue for the adding/refactoring of capability check on reliers

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -387,7 +387,6 @@ const Account = Backbone.Model.extend(
      * @param {Object} [options={}]
      *   @param {String} [options.kB] Optional The user's kB, if specified takes priority.
      *   @param {String} [options.password] Optional The user's password
-     * @returns {Promise}
```
vbudhram, line 390: Whoops!

PR #6021, opened 2020-07-23
fix(content-server): change the buttons from account-recovery to links.

### LINKED ISSUE #3951: '.button-link' color contrast issue, lack of hover state, should probably be a link in some cases

byrw:
## Because

* All the `.button-link` should be links and not buttons.

## This pull request

* Change the buttons with `.button-link` to links, and place it outside of `.button-row`.

* Remove the class `.button-link`.

fixes #3951

## Other information
I think the class `.centered` should have another name.

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/complete_reset_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/complete_reset_password.mustache
@@ -69,7 +69,9 @@
 
       <div class="button-row ">
           <button type="submit">{{#t}}Reset password{{/t}}</button>
-          <button class="button-link remember-password">{{#t}}Remember password? Sign in{{/t}}</button>
+      </div>
+      <div class="links">
+        <a href="#" class="remember-password">{{#t}}Remember password? Sign in{{/t}}</a>
       </div>
     </form>
   </section>
--- a/packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
@@ -17,7 +17,7 @@
 
             <div class="button-row">
                 <button class="primary-button generate-key-link">{{#t}}Continue{{/t}}</button>
-                <button class="button-link cancel-link">{{#t}}Cancel{{/t}}</button>
+                <button class="tertiary-button cancel-link">{{#t}}Cancel{{/t}}</button>
             </div>
         </form>
     </section>
--- a/packages/fxa-content-server/app/scripts/templates/account_recovery_confirm_key.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/account_recovery_confirm_key.mustache
@@ -42,7 +42,9 @@
 
                 <div class="button-row">
                     <button id="submit-btn" type="submit" class="primary-button">{{#t}}Confirm recovery key{{/t}}</button>
-                    <button class="button-link lost-recovery-key">{{#t}}Don't have a recovery key?{{/t}}</button>
+                </div>
+                <div class="links">
+                  <a href="#" class="lost-recovery-key">{{#t}}Don't have a recovery key?{{/t}}</a>
                 </div>
             </form>
         </section>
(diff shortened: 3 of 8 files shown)


packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -17,7 +17,9 @@
 
             <div class="button-row">
                 <button class="primary-button generate-key-link">{{#t}}Continue{{/t}}</button>
-                <button class="button-link cancel-link">{{#t}}Cancel{{/t}}</button>
+            </div>
+            <div class="links centered">
+              <a href="#" class="cancel-link">{{#t}}Cancel{{/t}}</a>
```
LZoog, line 22: I believe all of what you changed should indeed be links except for this one! I double checked our recently created modals in the `fxa-payments-server` package and also `fxa-settings` and we're considering a "cancel" an action (and should be a button).

It looks like we've already got a `primary-button` class and a `secondary-button` class. In case we need to use this later, we could create:

```
.tertiary-button {
  background: transparent;
  color: $blue-60;

  &:hover {
    color: $blue-70;
  }
  &:active {
    color: $blue-80;
  }
}
```

I based these colors on the primary button. Then you can change this back to `button` and add the `button-tertiary` class, but also be sure that margin you added between the primary button and this button is still present.

packages/fxa-content-server/app/scripts/templates/complete_reset_password.mustache
```diff
@@ -69,7 +69,9 @@
 
       <div class="button-row ">
           <button type="submit">{{#t}}Reset password{{/t}}</button>
-          <button class="button-link remember-password">{{#t}}Remember password? Sign in{{/t}}</button>
+      </div>
+      <div class="links centered">
+        <a href="#" class="remember-password">{{#t}}Remember password? Sign in{{/t}}</a>
```
LZoog, line 74: Hmm, at least this one has clickable whitespace on the left/right side. I think the `flex-direction: column` on `.links.centered` is causing that here. "Lost recovery key?" as well.

image

image

LZoog:
Looking good, when `cancel` is changed back to a button and we ensure some of those links don't have clickable space on the left/right side this will be good to go. As always, thanks for helping us out!

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
byrw: The "cancel" action from events seems to be working with the link, maybe it is not necessary to change to button.

packages/fxa-content-server/app/scripts/templates/complete_reset_password.mustache
byrw: It seems to have two solutions for this issue:
1 - Remove the `.centered`.
2 - On _custom-rows.css line 28, add
```
a {
  margin: auto;
}
```
What solution should I choose?

LZoog: I removed `centered` from a couple of these and that does seem to do the trick. Removing that class when we don't need it seems preferable to adding additional styles.👍

packages/fxa-content-server/app/scripts/templates/settings/account_recovery/confirm_password.mustache
byrw: Thanks for the review, today I learned something new. :)

I can not find where the `.secondary-button` is implemented. Do you know where it is located?

PR #6022, opened 2020-07-23
fix(aet): Add support for updating ecosystemAnonId using sessionToken

vbudhram:
## Because

- We need to support updating ecosystemAnonId with sessionToken
- sessionTokens are considered to contain all scopes
- Tested this against https://github.com/mozilla/fxa/pull/6020 and the ecosystemAnonId gets correctly set when user changes password

## This pull request

- Adds this support
- Some minor cleanup

Closes: #

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/db.js
+++ b/packages/fxa-auth-server/lib/db.js
@@ -1436,7 +1436,7 @@ module.exports = (config, log, Token, UnblockCode = null) => {
   };
 
   SAFE_URLS.updateEcosystemAnonId = new SafeUrl(
-    '/account/:id/ecosystemAnonId',
+    '/account/:uid/ecosystemAnonId',
     'db.updateEcosystemAnonId'
   );
   DB.prototype.updateEcosystemAnonId = async function (uid, ecosystemAnonId) {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-client/lib/client.ts
```diff
@@ -184,7 +184,6 @@ export default class AuthClient {
     payload: object,
     headers?: Headers
   ) {
-    console.log('hey');
```
jodyheavener, line 187: 🤦

packages/fxa-auth-server/lib/db.js
```diff
@@ -1436,7 +1436,7 @@ module.exports = (config, log, Token, UnblockCode = null) => {
   };
 
   SAFE_URLS.updateEcosystemAnonId = new SafeUrl(
-    '/account/:id/ecosystemAnonId',
+    '/account/:uid/ecosystemAnonId',
```
6a68, line 1439: doh, great catch

6a68:
@vbudhram Happy to help debug these tests once you've rebased

PR #6026, opened 2020-07-23
refactor(auth-server): use type-cacheable for caching

### LINKED ISSUE #6000: Replace custom cache with type-cacheable
Rather than an ad-hoc caching scheme, we can reduce our code maintenance by using a library for caching with Redis. type-cacheable utilizes the existing ioredis instance to provide caching.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2323)
┆Issue Number: FXA-2323

bbangert:
Because:

* Using a library for caching provides a nicer UX with additional
  capabilities to rolling our own.
* require() imports always resolve to any.

This commit:

* Uses type-cacheable for decorator based caching and clearing instead
  of our DIY function.
* Switches config.ts to use imports for Config type-safety and updates
  use of config for proper typing.

Closes #6000

## Checklist

This removes the logging assertions as the default strategy in type-cacheable does not log Redis failures. Our logging was never consulted, so it doesn't seem useful that we were logging before. If we do want to know when a Redis command fails, my recommendation for a future PR/issue would be to supply our own [default CacheStrategy](https://github.com/joshuaslate/type-cacheable/blob/master/packages/core/lib/strategies/DefaultStrategy.ts) that emits metrics of failures so we can graph it on a dashboard or even better, log a Sentry error.

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -1,13 +1,11 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import convict from 'convict';
+import fs from 'fs';
+import path from 'path';
+import url from 'url';
 
-'use strict';
-
-const fs = require('fs');
-const path = require('path');
-const url = require('url');
-const convict = require('convict');
 const DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages');
 
 convict.addFormats(require('convict-format-with-moment'));
@@ -1866,7 +1864,7 @@ if (conf.get('oauthServer.openid.keyFile')) {
   // If the file doesnt exist, or contains an empty object, then there's no active key.
   conf.set('oauthServer.openid.key', null);
   if (fs.existsSync(keyFile)) {
-    const key = JSON.parse(fs.readFileSync(keyFile));
+    const key = JSON.parse(fs.readFileSync(keyFile, 'utf-8'));
     if (key && Object.keys(key).length > 0) {
       conf.set('oauthServer.openid.key', key);
     }
@@ -1885,7 +1883,7 @@ if (conf.get('oauthServer.openid.newKeyFile')) {
   // If the file doesnt exist, or contains an empty object, then there's no new key.
   conf.set('oauthServer.openid.newKey', null);
   if (fs.existsSync(newKeyFile)) {
-    const newKey = JSON.parse(fs.readFileSync(newKeyFile));
+    const newKey = JSON.parse(fs.readFileSync(newKeyFile, 'utf-8'));
     if (newKey && Object.keys(newKey).length > 0) {
       conf.set('oauthServer.openid.newKey', newKey);
     }
@@ -1904,7 +1902,7 @@ if (conf.get('oauthServer.openid.oldKeyFile')) {
   // If the file doesnt exist, or contains an empty object, then there's no old key.
   conf.set('oauthServer.openid.oldKey', null);
   if (fs.existsSync(oldKeyFile)) {
-    const oldKey = JSON.parse(fs.readFileSync(oldKeyFile));
+    const oldKey = JSON.parse(fs.readFileSync(oldKeyFile, 'utf-8'));
     if (oldKey && Object.keys(oldKey).length > 0) {
       conf.set('oauthServer.openid.oldKey', oldKey);
     }
@@ -1931,5 +1929,6 @@ if (conf.get('isProduction')) {
 }
 
 export type conf = typeof conf;
+export type ConfigType = ReturnType<conf['getProperties']>;
 
 module.exports = conf;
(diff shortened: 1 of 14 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -1866,7 +1864,7 @@ if (conf.get('oauthServer.openid.keyFile')) {
   // If the file doesnt exist, or contains an empty object, then there's no active key.
   conf.set('oauthServer.openid.key', null);
   if (fs.existsSync(keyFile)) {
-    const key = JSON.parse(fs.readFileSync(keyFile));
+    const key = JSON.parse(fs.readFileSync(keyFile) as any);
```
bbangert, line 1867: Technically readFileSync can return non-string data, though it shouldn't in this case, thus the any type.

packages/fxa-auth-server/lib/oauth/logging/index.ts
```diff
@@ -1,13 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { Logger } from 'mozlog';
```
bbangert, line 4: Import here can properly bring in the type, while the require came in as a value.

packages/fxa-auth-server/lib/oauth/metrics/amplitude.ts
```diff
@@ -30,7 +30,7 @@ const EVENTS = {
 
 const FUZZY_EVENTS = new Map([]);
 
-module.exports = (log: Logger, config: conf) => {
+module.exports = (log: Logger, config: ReturnType<conf['getProperties']>) => {
```
bbangert, line 33: We might want to type alias this in `config/index.ts` now that I think about it, to avoid having to do this everywhere.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -84,39 +90,45 @@ function validateProductUpdate(
...
+  constructor(
+    log: Logger,
+    config: ReturnType<conf['getProperties']>,
+    statsd?: StatsD
+  ) {
     this.log = log;
-    this.cacheTtlSeconds =
-      config.subhub.plansCacheTtlSeconds ||
-      config.subscriptions.cacheTtlSeconds;
+    this.cacheTtlSeconds = config.subhub.plansCacheTtlSeconds;
```
bbangert, line 111: Type-checker flagged this upon having config type come through, seems `config.subscriptions.cacheTtlSeconds` does not exist, not sure if/when it last existed.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -446,47 +423,47 @@ class StripeHelper {
...
+    uid: string;
+    email: string;
+    forceRefresh?: boolean;
+    cacheOnly?: boolean;
+  }): Promise<Stripe.Customer | undefined> {
+    if (cacheOnly) {
+      return (
+        (await cacheManager.client?.get(
+          StripeHelper.customerCacheKey([uid, email])
+        )) || undefined
```
bbangert, line 460: Technically Redis returns a null here, which doesn't match the type signature, thus having to await and switch like this.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -558,37 +535,20 @@ class StripeHelper {
...
-   */
-  customerCacheKey(uid: string, email: string): string {
-    return `customer-${uid}|${email}`;
-  }
-
   /**
    * Remove the cache entry for a customer account
    * This is to be used on account deletion
    */
+  @CacheClear({ cacheKey: StripeHelper.customerCacheKey })
```
bbangert, line 549: Runs the function inside, and if it doesn't throw, will clear the cache key here per https://github.com/joshuaslate/type-cacheable#cacheclear

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -446,47 +423,47 @@ class StripeHelper {
...
-        }
-      } catch (err) {
-        this.log.error(`stripeHelper.customer.redis.get.failed`, { err });
-      }
+    cacheOnly = false,
+  }: {
+    uid: string;
+    email: string;
+    forceRefresh?: boolean;
+    cacheOnly?: boolean;
```
bbangert, line 454: I'll admit its not the most beautiful thing in TypeScript to specify the type and deconstruct it like this in one go. I switched to an object here because it was becoming increasingly confusing in `subscriptions.js` and elsewhere to see 2 boolean args at the end, and not be entirely sure what it was going to do. Switching to an object here made the caller's much more readable and clear.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -558,37 +535,20 @@ class StripeHelper {
...
-    const customerKey = this.customerCacheKey(uid, email);
-    try {
-      await this.redis.del(customerKey);
-    } catch (err) {
-      this.log.error(
-        `stripeHelper.removeCustomerFromCache failed to remove cache key: ${customerKey}`,
-        { err }
-      );
-    }
+    return;
```
chenba, line 551: Style nit: we don't need this line.

packages/fxa-auth-server/config/index.ts
bbangert: Because it could also come out as a Buffer, which doesn't like doing a direct cast to string, so then it would have to be something like:
`(fs.readFileSync(keyFile) as unknown) as string`

lmorchard: I think if we impose an encoding, this always results in a string - e.g. `fs.readFileSync(keyFile, 'utf-8')`

PR #6030, opened 2020-07-24
feat(payments): use SetupIntent API for SCA payment update

### LINKED ISSUE #5832: Use auth-server SetupIntent API when updating card information

lmorchard:
Because:

* To support SCA we must migrate to using Stripe's SetupIntent API

This commit:

* Add call to auth server to create SetupIntent
* Use secret from SetupIntent result to confirm card prior to creating a subscription

Closes #5832

Co-authored-by: Jackie Munroe <jmunroe@mozilla.com>
Co-authored-by: Les Orchard <me@lmorchard.com>

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.js
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.js
@@ -716,15 +716,6 @@ class DirectStripeRoutes {
     }
 
     const { paymentMethodId } = request.payload;
-
-    // Verify the payment method is already attached to the customer
-    const alreadyAttached = !!customer.sources.data.find(
-      (pm) => pm.id === paymentMethodId
-    );
-    if (!alreadyAttached) {
-      throw error.rejectedCustomerUpdate('Invalid payment method id');
-    }
-
     await this.stripeHelper.updateDefaultPaymentMethod(
       customer.id,
       paymentMethodId
@@ -1407,9 +1398,6 @@ const directRoutes = (
         },
         response: {
           schema: validators.subscriptionsStripeIntentValidator,
-        },
-        validate: {
-          payload: {},
         },
       },
       handler: (request) => directStripeRoutes.createSetupIntent(request),
--- a/packages/fxa-payments-server/src/lib/apiClient.test.ts
+++ b/packages/fxa-payments-server/src/lib/apiClient.test.ts
@@ -43,6 +43,12 @@ import {
   apiCancelSubscription,
   apiReactivateSubscription,
   apiUpdatePayment,
+  apiCreateCustomer,
+  apiCreateSubscriptionWithPaymentMethod,
+  FilteredSetupIntent,
+  apiCreateSetupIntent,
+  apiUpdateDefaultPaymentMethod,
+  apiRetryInvoice,
 } from './apiClient';
 
 describe('APIError', () => {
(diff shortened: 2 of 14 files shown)


lmorchard:
So, I *think* this PR is just about review-ready for FXA-2215. But, it's getting late and I need dinner. So, I'm opening this draft PR and will see what CI does with it by morning.

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -716,15 +716,6 @@ class DirectStripeRoutes {
     }
 
     const { paymentMethodId } = request.payload;
-
-    // Verify the payment method is already attached to the customer
-    const alreadyAttached = !!customer.sources.data.find(
```
lmorchard, line 721: Had to disable this check - the new Payment Methods aren't listed in customer sources, and the updateDefaultPaymentMethod call will throw if it's an invalid Payment Method anyway

packages/fxa-auth-server/lib/routes/subscriptions.js
```diff
@@ -1408,9 +1399,6 @@ const directRoutes = (
         response: {
           schema: validators.subscriptionsStripeIntentValidator,
         },
-        validate: {
-          payload: {},
```
lmorchard, line 1412: Removed this payload validator, since this call doesn't actually take any params. Otherwise, had to send an empty `{}` in the request.

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -369,4 +375,96 @@ describe('API requests', () => {
       requestMock.done();
     });
   });
+
+  describe('apiCreateCustomer', () => {
```
lmorchard, line 379: D'oh. Turns out I left off some apiClient test coverage from the earlier PR for SCA subscription create PR.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -120,6 +120,14 @@ const MockStripeElement = ({ testid }: { testid: string }) =>
       );
     }
   };
+  // Ensure we have a test ID to identify which kind of Stripe
+  // element this is, later on. Using Object.defineProperty because
+  // a simple mockElement.testid = testid makes TypeScript angry
+  return Object.defineProperty(mockElement, 'testid', {
```
lmorchard, line 126: This is kind of a weird hack, but it lets us verify that a `CardElement` was indeed passed through all the plumbing down to the `onSubmit` callback over in the payment update form.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateFormV2.test.tsx
```diff
@@ -0,0 +1,679 @@
...
+    );
+    expect(
+      apiClientOverrides.apiUpdateDefaultPaymentMethod
+    ).toHaveBeenCalledWith({
+      paymentMethodId: CONFIRM_CARD_SETUP_RESULT.setupIntent.payment_method,
+    });
+    expect(refreshSubscriptions).not.toHaveBeenCalled();
+  });
+
+  describe('Localized Plan Billing Description Component', () => {
```
lmorchard, line 407: All of these l10n tests in the second half of this new `PaymentUpdateFormV2.test.tsx` are copied verbatim from `PaymentUpdateForm.test.tsx`

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateFormV2.test.tsx
```diff
@@ -0,0 +1,679 @@
...
+    });
+
+    await waitForExpect(() =>
+      expect(
+        screen.queryByTestId('error-message-container')
+      ).not.toBeInTheDocument()
+    );
+  });
+
+  it('displays an error if apiCreateSetupIntent throws', async () => {
```
lmorchard, line 249: Was trying to find a way to DRY up and make these tests less verbose. They basically each just test one failure condition and the expected sequence of API calls & UI changes. Gave up and just did copypasta with tweaks.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateFormV2.tsx
```diff
@@ -0,0 +1,308 @@
...
+  typeof apiClient,
+  'apiCreateSetupIntent' | 'apiUpdateDefaultPaymentMethod'
+>;
+
+export type PaymentUpdateFormProps = {
+  customer: Customer;
+  customerSubscription: CustomerSubscription;
+  plan: Plan;
+  refreshSubscriptions: () => void;
+  setUpdatePaymentIsSuccess: () => void;
```
lmorchard, line 46: Needing to plumb in these `{set,reset}UpdatePaymentIsSuccess` functions is annoying. But, it seems to be the only way to support the "Your billing information has been updated successfully" banner on the subscription page. Rendering the AlertBar locally here ends up never showing it, because this component gets totally replaced on a customer refresh.

packages/fxa-payments-server/src/routes/Subscriptions/SubscriptionItem.tsx
```diff
@@ -72,16 +88,31 @@ export const SubscriptionItem = ({
...
-              {...{
-                plan,
-                customerSubscription,
-                customer,
-                updatePayment,
-                resetUpdatePayment,
-                updatePaymentStatus,
-              }}
-            />
+            {useSCAPaymentFlow ? (
```
lmorchard, line 91: The changes here in SubscriptionItem are plumbing and a component switch on `useSCAPaymentFlow`

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
```diff
@@ -1,18 +1,37 @@
...
 import { QueryParams } from '../../lib/types';
 import { APIError } from '../../lib/apiClient';
 import { FetchState } from '../../store/types';
 import { linkTo } from '@storybook/addon-links';
+import { CUSTOMER, FILTERED_SETUP_INTENT } from '../../lib/mock-data';
 
 function init() {
-  storiesOf('routes/Subscriptions', module)
+  setupVariantStories('routes/Subscriptions');
+  setupVariantStories('routes/SubscriptionsV2', {
```
lmorchard, line 21: Managed to re-purpose all the existing Subscription stories for the V2 route with a bit of component wrapping that flips the `useSCAPaymentFlow` bit and plumbs in the API mocks. Couldn't really think of any unique stories here.

packages/fxa-payments-server/src/routes/Subscriptions/index.test.tsx
```diff
@@ -53,6 +54,13 @@ import {
 import FlowEvent from '../../lib/flow-event';
 jest.mock('../../lib/flow-event');
 
+jest.mock('./PaymentUpdateFormV2', () => ({
```
lmorchard, line 57: Since `PaymentUpdateFormV2.test.tsx` hosts all the unit tests for the new form, we can just skip them here and mock the component for verifying the feature flip logic.

packages/fxa-payments-server/src/routes/Subscriptions/index.tsx
```diff
@@ -62,14 +71,35 @@ export const Subscriptions = ({
...
+  // Combined reset function to clear both previous success status and user's
+  // action to dismiss the alert.
+  const resetUpdatePaymentIsSuccessAndAlert = useCallback(() => {
+    resetUpdatePaymentIsSuccess();
+    resetSuccessAlert();
+  }, [resetUpdatePaymentIsSuccess]);
+
+  const showPaymentSuccessAlert =
+    !updatePaymentSuccessAlertIsHidden &&
+    // TODO: When we move to V2 SCA flow, we can drop updatePaymentStatus
```
lmorchard, line 101: These bits got a little more wrinkled, but the tl;dr: is that this manages the "Your billing information has been updated successfully" alert bar for both V1 and V2 payment update paths. We can simplify this when post-SCA cleanup happens

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateFormV2.tsx
```diff
@@ -0,0 +1,308 @@
...
+    amount,
+    currency,
+    interval,
+    intervalCount
+  );
+
+  return `You are billed ${planPricing} for ${name}. Your next payment occurs on ${formattedDateString}.`;
+}
+
+async function handlePaymentUpdate({
```
lmorchard, line 261: Oh yeah, and this `handlePaymentUpdate` function is the core of the new billing update flow.

PR #6040, opened 2020-07-24
refactor(auth-server): convert subscriptions to TS

### LINKED ISSUE #6002: [auth-server]: Refactor subscriptions.js to subscriptions.ts
Multiple edge cases were found moving stripe.js to stripe.ts. As a good portion of subscription logic is also in subscriptions.js, this file should also be moved to TypeScript.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2325)
┆Issue Number: FXA-2325

bbangert:
Because:

* We find edge cases and want to move our code to TypeScript.

This commit:

* Converts subscription routes from JS to TS and updates other types
  where needed.

Closes #6002

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.ts
@@ -1,26 +1,30 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-'use strict';
-
-const Sentry = require('@sentry/node');
-const error = require('../error');
-const isA = require('@hapi/joi');
-const { omitBy } = require('lodash');
-// @ts-ignore
-const ScopeSet = require('fxa-shared').oauth.scopes;
-const validators = require('./validators');
-const { splitCapabilities } = require('./utils/subscriptions');
-const { SUBSCRIPTION_UPDATE_TYPES } = require('../payments/stripe');
-// @ts-ignore
-const {
+import isA from '@hapi/joi';
+import * as Sentry from '@sentry/node';
+import ScopeSet from 'fxa-shared/oauth/scopes';
+import { metadataFromPlan } from 'fxa-shared/subscriptions/metadata';
+import {
+  DeepPartial,
   filterCustomer,
-  filterSubscription,
-  filterInvoice,
   filterIntent,
-} = require('fxa-shared').subscriptions.stripe;
-const { metadataFromPlan } = require('fxa-shared').subscriptions.metadata;
+  filterInvoice,
+  filterSubscription,
+} from 'fxa-shared/subscriptions/stripe';
+import omitBy from 'lodash/omitBy';
+import { Stripe } from 'stripe';
+
+import { ConfigType } from '../../config';
+import error from '../error';
+import {
+  AbbrevPlan,
+  StripeHelper,
+  SUBSCRIPTION_UPDATE_TYPES,
+} from '../payments/stripe';
+import { AuthLogger, AuthRequest } from '../types';
+import { splitCapabilities } from './utils/subscriptions';
+import validators from './validators';
 
 const SUBSCRIPTIONS_MANAGEMENT_SCOPE =
   'https://identity.mozilla.com/account/subscriptions';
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.ts
```diff
@@ -153,7 +118,7 @@ class DirectStripeRoutes {
       );
       for (const key of capabilityKeys) {
         const clientId = key.split(':')[1];
-        const capabilities = splitCapabilities(metadata[key]);
+        const capabilities = splitCapabilities((metadata as any)[key]);
```
lmorchard, line 121: Not sure it's much better, but might be able to use [this RawMetadata type](https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/subscriptions/types.ts#L3)? I think I spent a chunk of time months ago trying to see if I could cobble together a type around capability strings, but seems like it's more a runtime validation thing

packages/fxa-auth-server/lib/routes/subscriptions.ts
```diff
@@ -184,7 +149,7 @@ class DirectStripeRoutes {
       paymentToken,
       displayName,
       idempotencyKey,
-    } = request.payload;
+    } = request.payload as any;
```
lmorchard, line 152: Maybe a (future) opportunity here to share types with [payments-server's apiClient](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/lib/apiClient.ts#L125)?

packages/fxa-auth-server/lib/routes/subscriptions.ts
```diff
@@ -1190,6 +1202,15 @@ class DirectStripeRoutes {
       return [];
     }
 
+    if (!customer.subscriptions) {
+      throw error.internalValidationError(
+        'getSubscriptions',
+        {
+          customerId: customer.id,
+        },
+        'Customer did not have any subscriptions.'
```
lmorchard, line 1211: Miniscule nit, but: At first I thought, isn't it expected for a customer to not have any subscription if they were all deleted? But this is actually the the customer object is missing a subscriptions property altogether - not even an empty list. Not sure if that error would confuse us later.

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -23,13 +26,16 @@ export interface PlanMetadata {
 
 // https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=COPS&title=SP+Tiered+Product+Support#SPTieredProductSupport-MetadataAgreements
```
lmorchard, line 27: Might be worth updating this URL to point at [the ecosystem docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#product-metadata) while you're in here?

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -1,19 +1,22 @@
...
-  product_name: string;
   product_metadata?: RawMetadata;
-  currency: string;
-  amount: number;
-  interval: PlanInterval;
-  interval_count: number;
+  product_name: string;
 }
 
 // https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=COPS&title=SP+Tiered+Product+Support#SPTieredProductSupport-MetadataAgreements
```
lmorchard, line 22: This URL changed to [ecosystem docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#product-metadata) too, maybe?

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -23,13 +26,16 @@ export interface PlanMetadata {
...
 export interface ProductMetadata {
-  productSet?: string | null;
-  productOrder?: number | null;
+  appStoreLink?: string;
```
lmorchard, line 29: Hmm, this `appStoreLink` and `playStoreLink` don't appear in [the ecosystem docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#product-metadata). Probably need to be added there

packages/fxa-auth-server/lib/routes/subscriptions.ts
bbangert: Yea, I tried at least 4 different ways to type this, and all of them failed in various sad ways.

packages/fxa-auth-server/lib/routes/subscriptions.ts
bbangert: A future way to do this would be to re-type `validators.js` to TypeScript and swap `joi` for `typesafe-joi` which would allow us to get proper types from the existing joi schema. I used an approach like that in fxa-event-broker. Once that is done, then we could just indicate this should be typed as the validator for the payload, and woosh, type-safe access to what has been validated. Someday....

packages/fxa-shared/subscriptions/types.ts
bbangert: It should definitely be updated, I'm not sure when these options were added and how they are applied though.

packages/fxa-shared/subscriptions/types.ts
bbangert: It should definitely be updated, I'm not sure when these options were added and how they are applied though. I'll file an issue to update the ecosystem docs for them.

bbangert: Filed FXA-2349 to update the docs.

Issue #6017, opened 2020-07-22
Add better error handling in account method generateEcosystemAnonId


data-sync-user:
See https://github.com/mozilla/fxa/pull/5989#discussion_r458339010, comment below, and TODO listed in the code.

"If we get a 412 error, then we tried to update the anon_id, but another client beat us to it. We should handle that error by clearing the newly-created value out of localstorage and refetching the new value from the server. OK to handle this in a followup bug. In general, I think if we get a 4xx or 5xx error here, we probably need to discard the new ecosystemAnonId value until we can phone home successfully."

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2337)
┆Issue Number: FXA-2337

PR #6043, opened 2020-07-26
feat(aet): handle generateEcosystemAnonId update failure

jodyheavener:
## Because

- When generateEcosystemAnonId calls updateEcosystemAnonId it can fail with a 412 error, which means we tried to update the anon id but another client beat us to it. We need to handle this scenario.

## This pull request

- When updateEcosystemAnonId fails in this method, check if it's a 412 error, and retrieve the existing anon id from the auth server to store on the account instead
- If it’s not a 412 error, or if the profile query doesn’t yield an existing anon ID, just unset the value to avoid any confusion

Closes: #6017

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -363,6 +363,7 @@ const Account = Backbone.Model.extend(
      *   authenticationMethods,
      *   authenticatorAssuranceLevel,
      *   profileChangedAt,
+     *   ecosystemAnonId,
      * }
      */
     accountProfile() {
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -455,9 +456,23 @@ const Account = Backbone.Model.extend(
...
       } catch (err) {
-        // In the event of errors, we don't want to block any user actions, just log it
-        // TODO: Follow up with some better error handling
-        console.log('generateEcosystemAnonId failed...', err);
+        // If we get 412 Precondition Failed it means we
+        // tried to update, but another client beat us to
+        // it. Fetch the new anon ID and use it instead.
+        if (!AuthErrors.is(err, 'ECOSYSTEM_ANON_ID_UPDATE_CONFLICT')) {
+          this.unset('ecosystemAnonId');
+          console.log('generateEcosystemAnonId failed...', err);
```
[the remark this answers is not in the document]
vbudhram: I kinda of wonder if we should switch this around to match the comment.

```javascript
if (AuthErrors.is(err, 'ECOSYSTEM_ANON_ID_UPDATE_CONFLICT')) { ... fetch and set new anon id }
```

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -455,9 +456,23 @@ const Account = Backbone.Model.extend(
...
+        }
+
+        this.accountProfile().then((profile) => {
```
vbudhram, line 467: We will need to put `this.accountProfile()` into a try catch so that if it throws an error it doesn't bubble up to user.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -455,9 +456,23 @@ const Account = Backbone.Model.extend(
...
         }
       } catch (err) {
-        // In the event of errors, we don't want to block any user actions, just log it
-        // TODO: Follow up with some better error handling
-        console.log('generateEcosystemAnonId failed...', err);
+        // If we get 412 Precondition Failed it means we
+        // tried to update, but another client beat us to
+        // it. Fetch the new anon ID and use it instead.
+        if (!AuthErrors.is(err, 'ECOSYSTEM_ANON_ID_UPDATE_CONFLICT')) {
+          this.unset('ecosystemAnonId');
```
vbudhram, line 463: Feels like we should unset the id on any error and then only set it back if the server has a new value.

Issue #5920, opened 2020-07-13
Post verify email is not received for accounts created on Fenix or iOS


vasilicatamas:
**Build**
Prod [Train 1.178.0]
Stage [Train 1.179.0]

**Affected platforms**
iOS 13.4.1
Android 9 

**Steps to reproduce**
1. Create and verify a sync account using Fenix or Firefox for iOS
2. Observe the email received after the account confirmation.

**Expected results**
“Account verified. Next, sync another device to finish setup” email is successfully received.

**Actual Results**
“New sign-in to Firefox Preview” and “New Sign-in to Firefox iOS” emails are received.

**Notes**
This issue is not reproducible while using Fennec.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2272)
┆Issue Number: FXA-2272

data-sync-user:
➤ Chelsea Lewis commented:

[~adavis@mozilla.com] to get clarity per Leads discussion

data-sync-user:
➤ Alex Davis commented:

Just to be sure I understand, users are getting login notification emails when they register on iOS or Fenix (rather than the new template entitled “Account verified. Next, sync another device to finish setup”)?

If I am not mistaken, “New sign-in to Firefox Preview” and “New Sign-in to Firefox iOS” emails are not what we previously sent either. They would have received an account verified email.

vasilicatamas:
> Just to be sure I understand, users are getting login notification emails when they register on iOS or Fenix (rather than the new template entitled “Account verified. Next, sync another device to finish setup”)?

yes, that's correct!

![Screenshot_11](https://user-images.githubusercontent.com/18114210/87426612-227e2b80-c5e8-11ea-9ada-620e0f5319e5.png)

vasilicatamas:
This issue is verified as fixed on FxA Stage [1.182.0] under Android 9 and iOS 13.6. "Account verified. Next, sync another device to finish setup" email is successfully received by Android and iOS users after confirming the account.

![Screenshot_1](https://user-images.githubusercontent.com/18114210/89031064-68eabe80-d33a-11ea-9082-7b187053ad4f.png)

PR #6049, opened 2020-07-27
fix(emails): Send post account verify email when created from sync scope client

vbudhram:
## Because

- Account created using a sync scoped token did not send the post verify account email, instead they sent the "new device login" email

## This pull request

- Checks to see if the client doing the verification is using a clientId that is sync based and send the verification email if so
- Updates oauth token route to see if the account is newish in order to send the "new device login" email

Closes: #5920

## Checklist

@vladikoff @rfk Marking this as a draft to get your input on the approach here.

### DIFF
--- a/packages/fxa-auth-server/lib/constants.js
+++ b/packages/fxa-auth-server/lib/constants.js
@@ -6,4 +6,5 @@ module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
   SHORT_ACCESS_TOKEN_TTL_IN_MS: 1000 * 60 * 60 * 6,
+  MAX_NEW_ACCOUNT_AGE: 1000 * 60 * 60 * 24,
 };
--- a/packages/fxa-auth-server/lib/routes/utils/oauth.js
+++ b/packages/fxa-auth-server/lib/routes/utils/oauth.js
@@ -5,7 +5,10 @@
 'use strict';
 
 const encrypt = require('../../../lib/oauth/encrypt');
-const { OAUTH_SCOPE_OLD_SYNC } = require('../../constants');
+const {
+  OAUTH_SCOPE_OLD_SYNC,
+  MAX_NEW_ACCOUNT_AGE,
+} = require('../../constants');
 const ScopeSet = require('fxa-shared').oauth.scopes;
 
 // right now we only care about notifications for the following scopes
@@ -65,8 +68,10 @@ module.exports = {
     });
 
     // Email the user about their new device connection
-    // (but don't send anything if it was an existing device)
+    // (but don't send anything if it was an existing device or new account)
     if (!credentials.deviceId) {
+      const account = await db.account(credentials.uid);
+      const isNewAccount = Date.now() - account.createdAt < MAX_NEW_ACCOUNT_AGE;
       const geoData = request.app.geo;
       const ip = request.app.clientAddress;
       const emailOptions = {
@@ -78,12 +83,13 @@ module.exports = {
         uid: credentials.uid,
       };
 
-      const account = await db.account(credentials.uid);
-      await mailer.sendNewDeviceLoginEmail(
-        account.emails,
-        account,
-        emailOptions
-      );
+      if (!isNewAccount) {
+        await mailer.sendNewDeviceLoginEmail(
+          account.emails,
+          account,
+          emailOptions
+        );
+      }
     }
   },
 };
--- a/packages/fxa-auth-server/lib/routes/utils/signup.js
+++ b/packages/fxa-auth-server/lib/routes/utils/signup.js
@@ -5,6 +5,15 @@
 'use strict';
 const userAgent = require('../../userAgent');
 
+function isSyncBased(service) {
+  return (
+    service === 'sync' ||
+    service === '1b1a3e44c54fbb58' || // Firefox iOS browser
+    service === 'a2270f727f45f648' || // Fenix
+    service === '3c49430b43dfba77'
+  ); // Android Reference browser
+}
+
 module.exports = (log, db, mailer, push, verificationReminders) => {
   return {
     /**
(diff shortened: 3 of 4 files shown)


packages/fxa-auth-server/lib/routes/utils/signup.js
```diff
@@ -5,6 +5,15 @@
 'use strict';
 const userAgent = require('../../userAgent');
 
+function isSyncBased(service) {
+  return (
+    service === 'sync' ||
```
vbudhram, line 10: From my understanding, all these clients interact with sync or use sync. Is there a better way to identify that we should send the post verify sync email?

packages/fxa-auth-server/lib/routes/utils/oauth.js
```diff
@@ -65,8 +68,10 @@ module.exports = {
     });
 
     // Email the user about their new device connection
-    // (but don't send anything if it was an existing device)
+    // (but don't send anything if it was an existing device or new account)
     if (!credentials.deviceId) {
+      const account = await db.account(credentials.uid);
+      const isNewAccount = Date.now() - account.createdAt < MAX_NEW_ACCOUNT_AGE;
```
vbudhram, line 74: When creating a new account on a device, the account is already considered verified at this point. Is there a different way to detect that this is a new account and not to send the email?

packages/fxa-auth-server/lib/routes/utils/signup.js
rfk: I see what you're going for here, but I really hope we can find something better, because otherwise we will be constantly forgetting to add new clients to this list.

rfk: A more invasive, but perhaps more sustainable option would be to:

* Have `/session/verify_code` accept a `scope` argument in the payload alongside `service`.
* Teach fxa-content-server to pass the requested scopes when verifying the session, the same way it passes `service` right now.

That's more work up-front but would avoid having to maintain this list in future.

packages/fxa-auth-server/lib/constants.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
   SHORT_ACCESS_TOKEN_TTL_IN_MS: 1000 * 60 * 60 * 6,
+  MAX_NEW_ACCOUNT_AGE: 1000 * 60 * 60 * 24,
```
rfk, line 9: nit: this definitely needs a comment to explain what it's for :-)

packages/fxa-auth-server/lib/routes/utils/oauth.js
rfk: I can't think of a more clever way to do this that doesn't involve a major refactor.

The trouble we're having here is that we first create the account, and then in a second operation we create the requested refresh token. For non-oauth devices we avoid this problem by creating both the account and the sessionToken as a single operation, which means we can ensure we send exactly one email.

I can imagine a future world where there is an atomic "create a new account and provision these oauth tokens" operation that would give us a similar property for oauth clients, but that's a serious refactor!

packages/fxa-auth-server/lib/routes/utils/signup.js
```diff
@@ -63,13 +66,14 @@ module.exports = (log, db, mailer, push, verificationReminders) => {
...
-      if (service === 'sync') {
+      // so only send it if we're sure this is for sync or sync scoped client.
+      const scopeSet = ScopeSet.fromArray(scopes);
+      if (service === 'sync' || scopeSet.intersects(NOTIFICATION_SCOPES)) {
         const onMobileDevice =
           userAgent(request.headers['user-agent']).deviceType === 'mobile';
         const mailOptions = {
           acceptLanguage: request.app.acceptLanguage,
-          service,
+          service: 'sync',
```
vbudhram, line 76: The email template doesn't really need to know what scopes are involved, only that its sync based.

packages/fxa-content-server/app/scripts/views/confirm_signup_code.js
```diff
@@ -72,6 +72,7 @@ class ConfirmSignupCodeView extends FormView {
     const newsletters = account.get('newsletters');
     const options = {
       service: this.relier.get('service') || null,
+      scopes: this.relier.get('permissions') || null,
```
vbudhram, line 75: Client stores the scopes in an array called `permissions` and single string `scope` that stores a concatenation of all scopes. Using the array was a little easier to handle on the auth-server side of things.

packages/fxa-auth-server/lib/routes/utils/oauth.js
vbudhram: FWIW, I've went with the passing scopes into the session_verify function and then checking that one of the scopes was the "old sync scope"

packages/fxa-auth-client/lib/client.ts
```diff
@@ -540,6 +540,7 @@ export default class AuthClient {
       marketingOptIn?: boolean;
       newsletters?: string[];
       style?: string;
+      scopes?: string[];
```
rfk, line 543: micro-nit: if you put this field next to the existing `service` field it might make it slightly more obvious that they're kind of similar (in that they both indicate what the requested session is going to be used for).

packages/fxa-auth-server/lib/routes/utils/oauth.js
```diff
@@ -78,12 +83,13 @@ module.exports = {
         uid: credentials.uid,
       };
 
-      const account = await db.account(credentials.uid);
-      await mailer.sendNewDeviceLoginEmail(
-        account.emails,
-        account,
-        emailOptions
-      );
+      if (!isNewAccount) {
```
rfk, line 86: micro-nit: putting `const geoData` and friends inside the `if (!isNewAccount)` block might increase the readability of this slightly, since the use of the `isNewAccount` is closer to the point where you calculate it so it's more obvious what it's going to be for.

packages/fxa-content-server/app/scripts/views/confirm_signup_code.js
rfk: I think `permissions` may have some amount of post-processing done to it, e.g. removing scopes that aren't allowed or something like that? Anyway, it seems fine for this purpose.

PR #6050, opened 2020-07-27
chore(emails): add CC expiration email for Multiple Products

chenba:
Because:
 - a soon to expire credit card can be used to pay for multiple
   subscriptions

This commit:
 - send a slightly different reminder email to customers with multiple
   subscriptions

Closes: FXA-2319

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -1216,19 +1216,6 @@ export class StripeHelper {
       throw error.unknownCustomer(source.customer);
     }
 
-    // Follow subhub's lead and just use the customer's first active
-    // subscription as source of plan & product data.
-    //
-    // May need to refine this to search for the specific subscription paid
-    // for with a card if/when we allow multiple payment sources on multiple
-    // subscriptions.
-    //
-    // https://github.com/mozilla/subhub/blob/e224feddcdcbafaf0f3cd7d52691d29d94157de5/src/hub/vendor/customer.py#L204
-
-    /** @type {AbbrevProduct | undefined} */
-    let abbrevProduct;
-    /** @type {Plan | undefined | null} */
-    let plan;
     if (!customer.subscriptions) {
       throw error.internalValidationError(
         'extractSourceDetailsForEmail',
@@ -1236,6 +1223,9 @@ export class StripeHelper {
         new Error(`No subscriptions found for customer: ${customer.id}`)
       );
     }
+
+    let subscriptions = [];
+
     for (const subscription of customer.subscriptions.data) {
       if (['active', 'trialing'].includes(subscription.status)) {
         if (!subscription.plan) {
@@ -1247,13 +1237,37 @@ export class StripeHelper {
             )
           );
         }
-        plan = await this.expandResource(subscription.plan, PLAN_RESOURCE);
-        abbrevProduct = await this.expandAbbrevProductForPlan(plan);
-        break;
+
+        const plan = await this.expandResource(
+          subscription.plan,
+          PLAN_RESOURCE
+        );
+        const abbrevProduct = await this.expandAbbrevProductForPlan(plan);
+
+        const {
+          product_id: productId,
+          product_name: productName,
+        } = abbrevProduct;
+        const { id: planId, nickname: planName } = plan;
+        const productMetadata = this.mergeMetadata(plan, abbrevProduct);
+        const {
+          emailIconURL: planEmailIconURL = '',
+          downloadURL: planDownloadURL = '',
+        } = productMetadata;
+
+        subscriptions.push({
+          productId,
+          productName,
+          planId,
+          planName,
+          planEmailIconURL,
+          planDownloadURL,
+          productMetadata,
+        });
       }
     }
 
-    if (!plan || !abbrevProduct) {
+    if (subscriptions.length === 0) {
       throw error.missingSubscriptionForSourceError(
         'extractSourceDetailsForEmail',
         source
@@ -1271,24 +1285,11 @@ export class StripeHelper {
       email,
       metadata: { userid: uid },
     } = customer;
-    const { product_id: productId, product_name: productName } = abbrevProduct;
-    const { id: planId, nickname: planName } = plan;
-    const productMetadata = this.mergeMetadata(plan, abbrevProduct);
-    const {
-      emailIconURL: planEmailIconURL = '',
-      downloadURL: planDownloadURL = '',
-    } = productMetadata;
 
     return {
       uid,
       email,
-      productId,
-      productName,
-      planId,
-      planName,
-      planEmailIconURL,
-      planDownloadURL,
-      productMetadata,
+      subscriptions,
     };
   }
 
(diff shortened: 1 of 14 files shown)


chenba:
Note that customer with a single subscription will continue to receive the same CC expiration reminder prior to this patch.

chenba:
@johngruen These emails won't have any product specific content (aside from the list of product names).  Do we need the product specific TOS and Privacy Notice links in the emails for legal reasons?  Also, let me know what verbiage changes you'd like on the templates.

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -59,7 +59,6 @@ const stripePlan = require('../payments/fixtures/plan1.json');
 
 let config,
   log,
-  directStripeRoutes,
```
chenba, line 62: I don't want this to be in the same scope as `directStripeRotuesInstance`.  It's now a const where it is used.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1230,13 +1220,39 @@ export class StripeHelper {
             )
           );
         }
-        plan = await this.expandResource(subscription.plan, PLAN_RESOURCE);
-        abbrevProduct = await this.expandAbbrevProductForPlan(plan);
-        break;
+
+        /** @type {Plan | undefined | null} */
```
bbangert, line 1224: These are JS-doc style type annotations, we should be using proper TypeScript now, though I'm confused by why expandResource would not already have the right type here.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1230,13 +1220,39 @@ export class StripeHelper {
...
+        const plan = await this.expandResource(
+          subscription.plan,
+          PLAN_RESOURCE
+        );
+        /** @type {AbbrevProduct | undefined} */
```
bbangert, line 1229: Shouldn't be needed, this is TypeScript.

chenba: True.  These comments were moved along with their code.  I'll remove them.

chenba:
> 
> I'm kind of bummed that we introduced more code paths and emails...

There were really anything in the issue dictating one way or another.  Unifying into one email would mean we lose subscription specific content (the logo, the TOS and Privacy notice links) even when the customer has just one subscription.  (I already ping'd jgruen on verbiage confirmation above: https://github.com/mozilla/fxa/pull/6050#issuecomment-664586405, and do plan to wait for his 👍 before merging.)

chenba:
We could unify into a layout that's like 

```
[heading]

[message about expiring CC]

[per sub section with logo image, TOS and privacy notice links]

[footer]
```

@johngruen how 'about it?

packages/fxa-auth-server/lib/senders/templates/layouts/multiSubscriptions.html
```diff
@@ -0,0 +1,345 @@
+<!DOCTYPE html>
```
lmorchard, line 1: IMO, rather than a fully new layout, it could be better to use some [flags](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L2084) and [conditionals](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/templates/layouts/subscription.html#L274) to switch content as needed in the footer. We do this for [the subscription cancellation email](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L2038) to flip from "cancel" to "reactivate"

packages/fxa-auth-server/lib/senders/templates/multiSubscriptionsPaymentExpired.txt
```diff
@@ -0,0 +1,17 @@
+{{{subject}}}
+
+{{{t "Your credit card is about to expire" }}}
+
+{{{t "The credit card you're using to make payments for the following subscriptions is about to expire." }}}
+
+{{#subscriptionProductNames}}
+  - {{{t . }}}
```
lmorchard, line 8: Don't need `{{{t }}}` here, since `.` is not a localizable string.

packages/fxa-auth-server/lib/senders/templates/multiSubscriptionsPaymentExpired.html
```diff
@@ -0,0 +1,19 @@
+{{#*inline "title"}}
+  {{t "Your credit card is about to expire" }}
+{{/inline}}
+
+{{#*inline "content"}}
+  {{t "The credit card you're using to make payments for the following subscriptions is about to expire." }}
+  <br><br>
+  <ul>
+    {{#subscriptionProductNames}}
+    <li>{{t . }}</li>
```
lmorchard, line 10: Same thing here - can just use `{{ }}` since there's no content to extract for l10n

packages/fxa-auth-server/lib/senders/templates/multiSubscriptionsPaymentExpired.html
```diff
@@ -0,0 +1,19 @@
+{{#*inline "title"}}
```
lmorchard, line 1: Super-tiny nit that's more of a preference than anything: I was trying to prefix all these filenames and metric event things with "subscription..." (i.e. rather than something like "multi") so they naturally grouped together in alpha-sorted listings.

packages/fxa-auth-server/lib/senders/templates/multiSubscriptionsPaymentExpired.txt
chenba: 🤦

chenba:
Here's what the email looks like so far:
image

johngruen:
R+ for formatting

chenba:
> 
> 
> R+ for formatting

@johngruen hey sorry, I should've provided a summary.  We are debating whether to have a single unified layout that'll work for single and multiple subscriptions (maybe something like https://github.com/mozilla/fxa/pull/6050#issuecomment-665249006); or, leave the email for single subscription as it is and use the layout as shown above in the screenshot, which then leads to this question: https://github.com/mozilla/fxa/pull/6050#issuecomment-664586405.

johngruen:
So i like the layout in the screenshot and i don't think we need to link to individual TOS/PNs at all for this type of email.

chenba:
> 
> 
> So i like the layout in the screenshot and i don't think we need to link to individual TOS/PNs at all for this type of email.

Ok.  In that case, I'll just use that layout for all the expiring CC emails.

chenba:
Made some updates.  The layout is shared now, but the emails use different templates, because:
 - a list of one subscription is a bit weird
 - just the conditionals for pluralization make combined templates look silly

#### Screenshots

Screenshot_2020-08-07 MailDev
Screenshot_2020-08-07 MailDev(1)

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -1240,7 +1261,11 @@ describe('lib/senders/email:', () => {
         });
       });
 
-      await mailer[type](MESSAGE);
+      const { updateTemplateValues } = opts;
+      const tmplVals = updateTemplateValues
+        ? updateTemplateValues(MESSAGE)
+        : MESSAGE;
```
chenba, line 1267: I added an optional way to re-shape the message/template values since a) the templates can contain logic based on these values, and b) we don't _actually_ pass the same set of values for all emails.

Issue #6047, opened 2020-07-27
The user is not automatically connected after changing the password


vasilicatamas:
**Build**
Stage [Train 1.181.0]

**Affected platforms**
Windows 10

**Steps to reproduce**
1. Go to Sign In Page and click on "Forget password?".
2. Insert the email and click on "Begin reset".
3. Go to email and click on "Create new password".
4. Create a new password.

**Expected results**
The user is automatically connected.

**Actual Results**
“Reconnect to Firefox Account” yellow message is displayed.

![issue6](https://user-images.githubusercontent.com/18114210/88564408-db5a5680-d03b-11ea-9e10-be2fa209330c.gif)

**Additional notes**
This issue is not reproducible in Prod [Train 1.180.1]

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2352)
┆Issue Number: FXA-2352

vbudhram:
@vasilicatamas Thanks for finding this. It is related to the ecosystemAnonId parts we are adding.

<img width="370" alt="Screen Shot 2020-07-27 at 3 38 08 PM" src="https://user-images.githubusercontent.com/1295288/88584244-5bbe8e80-d01f-11ea-8cd6-8ab14faa0e86.png">

The browser is attempting to fetch keys, however the token to fetch them has already been used and the browser is complaining.

clouserw:
@vbudhram I think we can close this?

vasilicatamas:
This issue was verified as fixed on Prod [Train 1.181.2] under Windows 10 64-bit. But the user seems to not be automatically connected on mobile devices after resetting the password, issue tracked in a separate bug #6072.  Based on this I am setting the Verified label.

![issue8](https://user-images.githubusercontent.com/18114210/88816041-a75e6d00-d1c4-11ea-8d93-1b11bc9be642.gif)

PR #6052, opened 2020-07-27
fix(aet): On password reset, reauth to get new keyfetch token to set ecosystem id

vbudhram:
## Because

- We don't want to use the keyfetch token the browser will use, it will throw a 401 and not start syncing

## This pull request

- If a password is specified in `generateEcosystemAnonID` use the password to reauth and attempt to get a new keyfetch token
- Updated change password to not pass a password, this will make it use the `this.canFetchKeys()` if branch if possible

Closes: https://github.com/mozilla/fxa/issues/6047

## Checklist

Targets train-181

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/verification-reasons.ts
+++ b/packages/fxa-content-server/app/scripts/lib/verification-reasons.ts
@@ -8,7 +8,7 @@
 
 enum VerificationReasons {
   CHANGE_PASSWORD = 'change_password',
-  ECOSYSTEM_ANON_ID = 'ecosystem_anon_id',
+  ECOSYSTEM_ANON_ID = 'anon_id',
   FORCE_AUTH = 'force_auth',
   PASSWORD_RESET = 'reset_password',
   PASSWORD_RESET_WITH_RECOVERY_KEY = 'reset_password_with_recovery_key',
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/scripts/lib/verification-reasons.ts
```diff
@@ -8,7 +8,7 @@
 
 enum VerificationReasons {
   CHANGE_PASSWORD = 'change_password',
-  ECOSYSTEM_ANON_ID = 'ecosystem_anon_id',
+  ECOSYSTEM_ANON_ID = 'anon_id',
```
vbudhram, line 11: Had to make this change to fit the 16 character limit of auth-server.

packages/fxa-content-server/app/scripts/models/account.js
```diff
@@ -381,8 +381,11 @@ const Account = Backbone.Model.extend(
      * this function attempts to generate the id in the following manner:
      *
      * 1. Use kB if it is specified
-     * 2. Attempt to fetch keys if valid authentication token exists to do so
-     * 3. Attempt to use user's password to fetch keys
+     * 2. Attempt to use user's password to fetch keys
+     * 3. Attempt to fetch keys if valid authentication token exists to do so
+     *
+     * Note that on a password reset we will need to do a reauth since
+     * the browser will attempt to use the keyFetchToken.
```
LZoog, line 388: 👍

PR #6057, opened 2020-07-28
fix(style): add active and hover effect to the warning buttons from settings

### LINKED ISSUE #5658: The red buttons from Manage Account page do not have hover effect
**Build**
Stage [Train 1.175.0]

**Affected platforms**
Windows 10

**Steps to reproduce**
1. Launch Firefox and login using a Firefox account.
2. Navigate to https://accounts.stage.mozaws.net/settings.
3. Add a secondary email and check the “Remove” button.
4. Generate a recovery key and check the “Revoke” button.
5. Enable 2FA and check the “Disable” button.
6. Connect to some devices/services and check the “Disconnect...” button.

**Expected results**
All the red buttons from the Manage account page display a hover feedback effect.

** Actual Results**
Manage account red buttons do not have a hover effect.

issue9

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2109)
┆Issue Number: FXA-2109

byrw:
Closes #5658

## Because

* The warning buttons did not have an effect.

## This pull request

* Change the background color from the warning buttons, when the button receives hover or active.

fixes #5658

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_button-row.scss
+++ b/packages/fxa-content-server/app/styles/modules/_button-row.scss
@@ -67,15 +67,25 @@ button {
       background: $error-background-color;
     }
 
-    // 'type' is needed for selector specificity
-    &[type='submit'] {
-      &:active:enabled,
-      &:active:enabled:hover {
-        background: darken($error-background-color, 10);
+    &:enabled {
+      &:hover {
+        background: $error-background-color-hover;
       }
 
-      &:hover:enabled {
-        background: darken($error-background-color, 5);
+      &:active {
+        background: $error-background-color-active;
+      }
+
+      // 'type' is needed for selector specificity
+      &[type='submit'] {
+        &:active,
+        &:active:hover {
+          background: darken($error-background-color, 10);
+        }
+
+        &:hover {
+          background: darken($error-background-color, 5);
+        }
       }
     }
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/styles/modules/_button-row.scss
```diff
@@ -67,15 +67,25 @@ button {
...
+      // 'type' is needed for selector specificity
+      &[type='submit'] {
+        &:active,
+        &:active:hover {
+          background: darken($error-background-color, 10);
+        }
+
+        &:hover {
+          background: darken($error-background-color, 5);
+        }
```
LZoog, line 88: What is this being used for? I took it out and the issue seems resolved without it!

red-buttons

byrw: This just affects the buttons with "type=submit". If you try with the "Delete account" button, the button will display the background in blue.

LZoog: Ah I see, it's for delete account, good catch. I wonder if the delete account button is purposefully slightly lighter than the colors we were given for the other red actions, but that's how the delete account button is in production now and the settings is in the process of a redesign now so this'll work. :)

PR #6060, opened 2020-07-28
chore(subscriptions): use just "Subscriptions"

chenba:
Because:
 - these mocks https://zpl.io/bAWMO9B

This commit:
 - change "Subscriptions & Payments" to just "Subscriptions"

Closes: FXA-2318

## Checklist

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -32,7 +32,7 @@ card-error = Your transaction could not be processed. Please verify your credit 
 
 ## settings
 settings-home = Account Home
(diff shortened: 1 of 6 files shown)


packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -32,7 +32,7 @@ card-error = Your transaction could not be processed. Please verify your credit
 
 ## settings
 settings-home = Account Home
-settings-subscriptions = Subscriptions & Payments
+settings-subscriptions-title = Subscriptions
```
lmorchard, line 35: :+1: for remembering to tweak the message ID on string change

Issue #6031, opened 2020-07-24
500 Server Error when updating ecosystemAnonId in stage


rfk:
STR (using a stage account and some manually-generated tokens):

```
>>> token
'eyJhbGciOiJSUzI1NiIsInR5cCI6ImF0K0pXVCIsImtpZCI6IjIwMTkwNzMwLTE1ZTQ3M2ZkIn0.eyJpc3MiOiJodHRwczovL2FjY291bnRzLnN0YWdlLm1vemF3cy5uZXQiLCJhdWQiOiI1ODgyMzg2YzZkODAxNzc2IiwiY2xpZW50X2lkIjoiNTg4MjM4NmM2ZDgwMTc3NiIsImV4cCI6MTU5NTc0Mjg3MiwiaWF0IjoxNTk1NTcwMDcyLCJqdGkiOiI4YzA0ZTcxODYzNjZjMjcyYmNiZmIxNmU0MTA3ZDI3MjAzMzlkYmZjYTdhNjQ3OThkYzQ0ZTUzMzM1Nzc3NDkwIiwic2NvcGUiOiJwcm9maWxlOndyaXRlIiwic3ViIjoiNThiMDlkNWE0ZDVmNDA1ZmJiZGViYzk2Mjg0ODgzZjUiLCJmeGEtZ2VuZXJhdGlvbiI6MTU3MzcwMDg5NDA0MSwiZnhhLXByb2ZpbGVDaGFuZ2VkQXQiOjE1NzM3MDA4OTQwNDF9.fCz64Gt_klo3329nLTqGF7WMWd5lEuYhUZjfhv4PbE7x2DTwKWYCf1r-flcUJkyh0Qs3Xxgb3K306gwTxyB1QkUUSZkyjmz0247phRsNumvFGwbmcOJpgNUzY1wmEZE0sM36bnENIVEMWZyQVbxqADBiszpA_Q0IMN3NQfUMu8Hd6wWq87kbnTXEjftskIB9EeW4RF8j7Gbiz5siOMEqLjtRelSM58C7o5JSXeiel9pKjt_g5wyXE69wqqnoun0hkgPOwYVtLO7YgXAeEACPs3bMNyW7pKfh9YWpAwK4HkL1x5ypQOl8nfnBKDjA6dSX-EqYw_vWfCIjJtWJQ-3MCA'
>>>
>>> anonId
[...]

PR #6062, opened 2020-07-28
fix(profile): aet call to auth server should be put, not get

jodyheavener:
## Because

- We were GETing to the auth server '/account/ecosystemAnonId' endpoint, when it should be [PUTing](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/account.js#L1494)

## This pull request

- Fixes that, adds test to assert call happens

Closes: #6031

## Checklist

### DIFF
--- a/packages/fxa-profile-server/test/api.js
+++ b/packages/fxa-profile-server/test/api.js
@@ -1685,7 +1685,29 @@ describe('api', function () {
         return hash.update(anonId).digest('hex');
       }
 
-      xit('should post a new ecosystem_anon_id', function () {});
+      it('should put a new ecosystem_anon_id', async function () {
+        const payload = {
+          ecosystemAnonId: 'beans on pizza',
+        };
+        const headers = {
+          authorization: `Bearer ${tok}`,
+        };
+
+        mock.anonIdUpdated(payload, headers);
+        mock.token({
+          user: USERID,
+          scope: ['profile:ecosystem_anon_id:write'],
+        });
+
+        const res = await Server.api.post({
+          url: '/ecosystem_anon_id',
+          payload,
+          headers,
+        });
+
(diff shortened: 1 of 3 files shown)


packages/fxa-profile-server/test/api.js
```diff
@@ -1685,7 +1685,27 @@ describe('api', function () {
...
+          url: '/ecosystem_anon_id',
+          payload: {
+            ecosystemAnonId: '12345',
+          },
+          headers: {
+            authorization: 'Bearer ' + tok,
+          },
+        });
+
+        assert.equal(res.statusCode, 200);
```
jodyheavener, line 1706: This is good and all, and means we made it to the end of the request without any errors, but I'm not sure how to assert request body or headers to the auth server, with how we're [stubbing requests](https://github.com/mozilla/fxa/pull/6062/files#diff-7af5e8103c0a67bff380affec012e4c1R149). Any ideas?

vbudhram: > how to assert request body or headers to the auth server

Would you be able to spy on the nock object?

6a68: It looks like nock can return a specific response only if headers match ([`.matchHeader()`](https://www.npmjs.com/package/nock#request-headers-matching)) or only if the [body matches](https://www.npmjs.com/package/nock#specifying-request-body). @jodyheavener is that what you were looking for?

jodyheavener: Ahh, I get it now. I was looking for a way to reactively inspect the request, but it looks like in Nock's case the route is simply not called if _all_ the setup requirements are not met. So a successful call == the expected values have been met. Thanks!

PR #6063, opened 2020-07-28
fix(payments): add fxa-shared and fxa-react as CI dependencies for fxa-payments-server

lmorchard:
Looks like we're not running payments-server tests when fxa-shared or fxa-react change, but payments-server uses those. This hopefully fixes that

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,14 @@
       "fxa-profile-server": [
         "fxa-auth-server",
         "fxa-shared"
+      ],
+      "fxa-payments-server": [
+        "fxa-shared",
+        "fxa-react"
+      ],
+      "fxa-settings": [
+        "fxa-shared",
+        "fxa-react"
       ]
     }
   },


lmorchard:
@LZoog / @jodyheavener - I also noticed that fxa-settings isn't in the list for moduleDependencies. We probably want to add an entry for that and all the other packages it might rely on

lmorchard:
For more context, though the docs here are a bit out of date, I think this still accurately describes how we choose what to run in CI:

https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-tests-circleci#circleci-modules-to-testjs

meandavejustice:
Added jody and lauren as reviewers since there is a question for them.

jodyheavener:
Good catch @lmorchard, I agree. It looks like fxa-settings also only depends on fxa-react and fxa-shared (for now at least), so I'm happy to make a PR for that unless you want to add it to this one.

lmorchard:
Sure, I can just add fxa-settings to this PR. Wanted to be sure I wasn't missing any!

Issue #6074, opened 2020-07-29
Log old and new anon_ids at update time


data-sync-user:
In order for the data pipeline to (1) correlate AET data per user across anon_id changes and (2) reconstruct a user's historical list of anon_ids as part of a data deletion request, we need to log an event when the anon_id is updated. Ops will then relay this into the data pipeline on the backend.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2363)
┆Issue Number: FXA-2363

PR #6081, opened 2020-07-30
fix(aet): When the eco_anon_id is updated, log the old and new values

6a68:
## Because

- In order for the data pipeline to (1) correlate AET data per user across anon_id changes and (2) reconstruct a user's historical list of anon_ids as part of a data deletion request, we need to log an event when the anon_id is updated. Ops will then relay this into the data pipeline on the backend.

## This pull request

- Adds a log.info call in the auth-server route that updates the eco_anon_id, just after we successfully update the value in the auth-db database.

Closes: #6074.

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1536,25 +1536,33 @@ module.exports = (
 
         await customs.check(request, uid, 'updateEcosystemAnonId');
 
-        if (ifNoneMatch || ifMatch) {
-          const { ecosystemAnonId: existingAnonId } = await db.account();
+        const { ecosystemAnonId: existingAnonId } = await db.account();
 
-          if (existingAnonId) {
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1536,26 +1536,34 @@ module.exports = (
...
 
+        // Order matters here: if the awaited `db.updateEcosystemAnonId` call
+        // doesn't throw, *then* log the (old, new) pair, so that the data
+        // pipeline can maintain an accurate historical list of anon_ids.
+        if (existingAnonId) {
+          log.info('account.updateEcosystemAnonId.complete', {
+            previous: existingAnonId,
+            next: ecosystemAnonId,
+          });
+        }
```
jodyheavener, line 1565: Would there be analytical value in also recording whether an oauth or session token is responsible for the change (not the token itself, though)?

PR #6083, opened 2020-07-30
feat(aet): make either precondition header mandatory when updating anon id

jodyheavener:
## Because

- There's no valid reason to be performing an eco id change without either asserting one does not exist, or that an existing one matches

## This pull request

- Makes either If-Match or If-None-Match mandatory when performing the ID update

## Checklist

## Other information

This is going to conflict with #6081 - I'm happy to wait until it's merged and then resolve conflicts on my end.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1494,7 +1494,11 @@ module.exports = (
       method: 'PUT',
       path: '/account/ecosystemAnonId',
       apidoc: {
-        errors: [error.invalidScopes, error.anonIdUpdateConflict],
+        errors: [
+          error.invalidScopes,
+          error.anonIdUpdateConflict,
+          error.anonIdNoCondition,
+        ],
       },
       options: {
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1536,9 +1540,13 @@ module.exports = (
...
         await customs.check(request, uid, 'updateEcosystemAnonId');
 
+        if (!ifNoneMatch && !ifMatch) {
+          throw error.anonIdNoCondition()
+        }
+
         const { ecosystemAnonId: existingAnonId } = await db.account();
 
-        if ((ifNoneMatch || ifMatch) && existingAnonId) {
+        if (existingAnonId) {
```
6a68, line 1549: If there's no existing anon_id, but the client specifies "if-match: x", should we do the update?

jodyheavener: Per Slack convo, noted and may be good as a followup.

PR #6084, opened 2020-07-30
fix(auth): added exports to fxa-shared package.json

dannycoates:
Because auth-server doesn't use ts-node/register in production and fxa-shared compiles typescript, requiring a subpath like 'fxa-shared/subscriptions/metadata' finds no js to load. The "exports" field of package.json allows us to map required paths to other locations, in this case ./dist.

see https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_package_entry_points

### DIFF
--- a/packages/fxa-shared/package.json
+++ b/packages/fxa-shared/package.json
@@ -3,6 +3,10 @@
   "version": "1.181.2",
   "description": "Shared module for FxA repositories",
   "main": "dist/index.js",
+  "exports": {
+    ".": "./dist/index.js",
+    "./": "./dist/"
+  },
   "scripts": {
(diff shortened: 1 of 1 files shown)


bbangert:
This makes sense, but I'm curious how/why it worked previously, especially since it looks like payments-server does similar style imports. Why is it needed now, but it wasn't before?

dannycoates:
> Why is it needed now, but it wasn't before?

afaict the subpaths that are used in payments-server are either js like 'fxa-shared/express' or transpiled by webpack(?) like 'fxa-shared/subscriptions/metadata' which knows how to load ts files.

Issue #6092, opened 2020-07-31
Spurious "must provide If-Match or If-None-Match" error when updating ecosystemAnonId


rfk:
Trying out latest `main` with a local build of Firefox that manages ecosystemAnonId, I get a `400 Bad Request` error when trying post a new anonId value to the server. It tells me that I must provide an If-Match or If-None-Match header, despite the fact that I am already providing one.

PR incoming.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2368)
┆Issue Number: FXA-2368

PR #6093, opened 2020-07-31
fix(aet): Correctly check preconditions when updating ecosystemAnonId

rfk:
## Because

* Hapi normalizes all headers to lowercase, but the code for if-match checking was looking for headers with mixed case.
* You have to provide a uid when looking up an account in the db, but the route for updating ecosystemAnonId was not 😅.

## This pull request

* Fixes the precondition checks so they execute correctly.

Closes: #6092

## Checklist

- [ ] ~~I have added necessary documentation (if appropriate).~~

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.js
+++ b/packages/fxa-auth-server/lib/routes/account.js
@@ -1531,8 +1531,8 @@ module.exports = (
           return hash.update(anonId).digest('hex');
         }
 
-        const ifNoneMatch = request.headers['If-None-Match'];
-        const ifMatch = request.headers['If-Match'];
+        const ifNoneMatch = request.headers['if-none-match'];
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/account.js
```diff
@@ -1541,10 +1541,10 @@ module.exports = (
         await customs.check(request, uid, 'updateEcosystemAnonId');
 
         if (!ifNoneMatch && !ifMatch) {
-          throw error.anonIdNoCondition()
+          throw error.anonIdNoCondition();
         }
 
-        const { ecosystemAnonId: existingAnonId } = await db.account();
+        const { ecosystemAnonId: existingAnonId } = await db.account(uid);
```
jodyheavener, line 1547: Oof 🤦

PR #6096, opened 2020-07-31
chore(payments): upgrade to @fluent/react 0.12.0

stasm:
## Because

- `@fluent/react` has been rewritten to TypeScript. The rewrite required a few breaking changes in the API.

## This pull request

- Changes the `<LocalizationProvider>` to accept an instance of `ReactLocalization` rathen than an iterable of `FluentBundles`.
- Stores the instance of `ReactLocalization` in `<AppLocalizationProvider>`'s state.
- Updates the callsites of the `<Localized>` component to the new API. The new API is more explicit:
  - The `vars` prop is a dict of variables passed into the translation for interpolation.
  - The `elems` prop is a dict of React elements which will be inserted into the translation.

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).

PR #6098, opened 2020-07-31
chore(payments): upgrade @fluent/bundle et al.

stasm:
## Because

- Fluent packages used to ship two build artifacts: `index.js` written in the most current ES version, and `compat.js` transpiled to ES2017. In https://github.com/projectfluent/fluent.js/issues/472 we removed the `compat.js` builds and set `tsc` to target ES2018.
- I didn't think it was appropriate for upstream Fluent to make the decision about the browsers supported by the `compat.js` builds. It's something that app developers should have more control over.
- The expectation is that apps using Fluent will be able to transpile `node_modules` dependencies to match their browser compatibility requirements. 

## This pull request

- Is based on #6096. Only the last commit is new compared to #6096 
- Upgrades three `@fluent` packages used by `payments-server`: `@fluent/bundle`, `@fluent/react`, `@fluent/langneg` to their newest versions which ship ES2018 code.
- As far as I was able to verify it, the files built with `yarn build` in `build/static/js` are transpiled by FxA's build system correctly to match your `browerlist` config.

## Checklist


stasm:
Other than transpilation, it's worth to mention that Fluent uses fairly new methods like `Object.entries()`, `Array.prototype.includes()`, and `String.prototype.includes()`. Looking at https://github.com/mozilla/fxa/blob/main/docs/supported-browsers.md I think all target browsers should support them. Do you run any cross-browser integration tests to ensure compatibility?

chenba:
Let's rebase `main` once https://github.com/mozilla/fxa/pull/6152 is merged to see if these conflicts will go away.

chenba:
Closing in favor of #6291.

Issue #6075, opened 2020-07-29
Increase avatar file size limit to 2MB


data-sync-user:
Ops is going to increase the server limit from 1.35mb to 2mb. We will need to increase the validation [here](https://github.com/byrw/fxa/blob/master/packages/fxa-content-server/app/scripts/lib/constants.js#L90) and [here](https://github.com/byrw/fxa/blob/master/packages/fxa-graphql-api/src/config.ts#L137). It might not be a bad idea to change the default [server value](https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/lib/config.js#L102) to 2mb as well.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2364)
┆Issue Number: FXA-2364

PR #6099, opened 2020-07-31
refactor(avatars): increase max file size to 2mb

jodyheavener:
## Because

- The current max file size for avatar upload is 1mb locally, 1.35mb in prod.

## This pull request

- Bumps frontend and [default] server validation to 2mb across the board. Ops will still need to set the server env var to 2mb.

Closes: #6075

## Checklist

## Other information

**Hold merge**. We need to wait until this change is set in prod first (cc @jrgm).


jodyheavener:
Per ops meeting, the server value with increase in a 182 point release, so this can go into train 183.

Issue #6102, opened 2020-07-31
404 page message should be in a `p` tag and has incorrect line-height


LZoog:
"The page you requested is not found..." text should be in a `p` tag instead of a `section` and should have a line-height of `1.5em` to match our other styles.

<img width="1079" alt="image" src="https://user-images.githubusercontent.com/13018240/89052725-c02d6680-d31b-11ea-8659-c3b4ffa17ada.png">

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2371)
┆Issue Number: FXA-2371

data-sync-user:
➤ Vijay Budhram commented:

From mtg: Contributor filed a PR for this, Lauren working on review.

PR #6104, opened 2020-07-31
Changed the 404 page Text to be Match with Other Styles

SrilalS:
Put the 404 page message in a "p" tag and Corrected the Line-heigth to 1.5em

## Because

-Current one doesn't go with other Styles according to the ISSUE this is fixing.

## This pull request

- Fixed the issue by simply changing the tag and adding the necessary styles.

Closes: #6102 

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/styles/_layout.scss
+++ b/packages/fxa-content-server/app/styles/_layout.scss
@@ -89,6 +89,11 @@
   margin-top: 24px;
 }
 
+.fxa-404-p {
+  line-height: 1.5em;
+  margin: 0;
+}
+
 #static-footer {
   align-items: center;
   display: flex;
--- a/packages/fxa-content-server/server/templates/pages/src/404.html
+++ b/packages/fxa-content-server/server/templates/pages/src/404.html
@@ -23,10 +23,10 @@
           <h1 id="fxa-404-header">{{#t}}Page not found{{/t}}</h1>
         </header>
 
-        <section>
+        <p class="fxa-404-p" >
           {{#t}}The page you requested is not found. We've been notified and
           will fix any links that may be broken.{{/t}}
-        </section>
+        </p>
 
         <div class="button-row">
           <a href="/signup" class="button primary-button" id="fxa-404-home"
--- a/packages/fxa-payments-server/public/404.html
+++ b/packages/fxa-payments-server/public/404.html
@@ -19,10 +19,10 @@
     <div id="stage">
       <div id="main-content" class="card">
         <header><h1 id="fxa-404-header">Page not found</h1></header>
-        <section>
+        <p class="fxa-404-p">
           The page you requested is not found. We've been notified and will fix
           any links that may be broken.
-        </section>
+        </p>
         <div class="button-row">
           <a href="/signup" class="button primary-button" id="fxa-404-home"
             >Home</a


packages/fxa-content-server/server/templates/pages/src/404.html
```diff
@@ -23,10 +23,10 @@
           <h1 id="fxa-404-header">{{#t}}Page not found{{/t}}</h1>
         </header>
 
-        <section>
+        <p style="line-height: 1.5em;">
```
LZoog, line 26: Hey @SrilalS, thanks for the PR! Sorry for the delay in getting back to you, I noticed something funky with the 404 file you edited in the `fxa-payments-server` and wanted to dig around a bit, filed an issue for it here if you're curious: #6168 

We don't typically use inline styles so I think we should use a class here instead, especially to reuse this in the `fxa-payments-server`. It's not super apparent where this should live, but let's add it in the same place we're adding styles for `#fxa-404-home`, inside of `_layout.scss`. Our naming conventions here are also not very strict... We could follow the same convention as that ID, such as:

```
.fxa-404-p {
  line-height: 1.5em;
  margin: 0;
}
```

I noticed the `p` tag adds a larger top/margin bottom than we had with the `section` tag, so I've addressed that here. You can set the class on the `p` tag in both files. 🙂

packages/fxa-content-server/app/styles/_layout.scss
```diff
@@ -89,6 +89,11 @@
   margin-top: 24px;
 }
 
+#fxa-404-p {
```
LZoog, line 92: Much better! Although, we've actually got another issue setup around not using IDs to style, we should use a class instead. Would you mind changing this to a class (`.fxa-404-p`) instead of an ID?

Also, there's a couple of things around your commit before this can be merged. Check out [our contributing guidelines](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md#sending-pull-requests) but the TLDR to get this PR merged is to please:
* Make sure your commit is [GPG signed](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification)
* Squash your commits into one, you can use `git rebase -i main` to interactively "fix up" the second commit and "reword" the first commit
* Use the format suggested for your commit, such as:
```
fix(style): Update 404 pages paragraph styles

Because:
* The line-height on paragraphs on 404 pages did not match the line-height on other FxA pages.

This commit:
* Adds a class on the 404 page `p` tag and adds styles.
```

Doesn't have to match my wording but we've found using this format consistently helps provide a lot of context when using `git blame` or `git log`. 👍 Let me know if you've got any questions.

SrilalS: Thank you. I've made the change and have signed the Commit.

LZoog: @SrilalS thanks for making those changes! PR looks good to me. Hmm, although it looks like your commit signature is still showing as `unverified`. Will you double check that's set correctly? Additionally, we should only see one commit message here. If you run `git rebase -i main`, you should be able to mark the last 3 commits as "f" or "fix up". When rebased it'll merge into one new commit, and then you can force-push to this branch.

SrilalS: @LZoog Thank you for your help. I have rebased and fixed the unverified issue and force pushed it.

LZoog: @SrilalS no problem, thanks for sticking with this PR! Sorry to have even more back and forth here. 😓

Looks like you were able to successfully squash/fixup your commits, but I'm still not seeing that the commit signature is GPG signed - typically if you go to the "commits" tab you'll see a "verified" note next to the commit, here's a screenshot of one of mine: 
image
If you run `git config --list` do you see `commit.gpgsign=true`? If you run `git verify-commit 432e80cc` do you see a "Signature made using RSA key [key] with a "Good signature from [you]" listed?

We also have a [commit message format](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md#git-commit-guidelines) to follow, I'd pasted an example above under "Use the format suggested for your commit, such as:". While the "because" and "this commit" sections can help give more context, the required part is that your commit start with `<type>(<scope>):`, such as `fix(style):`. When we run our release script, it runs through all new commits from our previous release to create our changelog files and this format helps with sorting. After the GPG signature is sorted, reword your commit to start with this and force-push if needed, and we'll be good to go. 👌

SrilalS: @LZoog no problem. as a student I'm learning a lot here. I have verified the commit and changed the commit message according to the commit message format.

LZoog: @SrilalS So glad you were able to figure it out, thanks again for helping out here!

SrilalS: @LZoog Thank you for your Support.

PR #6105, opened 2020-07-31
refactor(rescripts): include all of fxa as valid paths in rescripts

dannycoates:
Because I was about to add fxa-auth-client to the additionalJSIpmorts but ran into problems. Instead of trying to include the minimum set of paths to add to the webpack rule this just lets all of fxa in.

Also, since we started using `workspace:*` in package.json we no longer need the aliases or jest module mappings!

### DIFF
--- a/packages/fxa-auth-client/lib/crypto.ts
+++ b/packages/fxa-auth-client/lib/crypto.ts
@@ -175,7 +175,7 @@ export async function checkWebCrypto() {
       window.asmCrypto = await import(/* webpackChunkName: "asmcrypto.js" */ 'asmcrypto.js');
       // prettier-ignore
       // @ts-ignore
-      await import(/* webpackChunkName: "webcrypto-liner" */ 'webcrypto-liner');
+      await import(/* webpackChunkName: "webcrypto-liner" */ 'webcrypto-liner/build/shim');
       return true;
     } catch (e) {
       return false;
--- a/packages/fxa-content-server/webpack.config.js
+++ b/packages/fxa-content-server/webpack.config.js
@@ -75,7 +75,6 @@ const webpackConfig = {
       'ua-parser-js': require.resolve('ua-parser-js/src/ua-parser'),
       uuid: require.resolve('node-uuid/uuid'),
       vat: require.resolve('node-vat/vat'),
-      'webcrypto-liner': require.resolve('webcrypto-liner/build/shim'),
       // Webpack 4 doesn't support the "exports" property of package.json
       // so unfortunately we need to remap it here as well.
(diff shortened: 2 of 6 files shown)


packages/fxa-auth-client/lib/crypto.ts
```diff
@@ -175,7 +175,7 @@ export async function checkWebCrypto() {
       window.asmCrypto = await import(/* webpackChunkName: "asmcrypto.js" */ 'asmcrypto.js');
       // prettier-ignore
       // @ts-ignore
-      await import(/* webpackChunkName: "webcrypto-liner" */ 'webcrypto-liner');
```
dannycoates, line 178: I'm preemptively removing the alias here for future work on fxa-setting

packages/fxa-content-server/webpack.config.js
```diff
@@ -75,7 +75,6 @@ const webpackConfig = {
       'ua-parser-js': require.resolve('ua-parser-js/src/ua-parser'),
       uuid: require.resolve('node-uuid/uuid'),
       vat: require.resolve('node-vat/vat'),
-      'webcrypto-liner': require.resolve('webcrypto-liner/build/shim'),
```
dannycoates, line 78: `#no-aliases`

packages/fxa-react/configs/rescripts.js
```diff
@@ -4,14 +4,10 @@
...
-  'fxa-react': resolve(__dirname, '../'),
-  'fxa-shared': resolve(__dirname, '../../fxa-shared'),
-};
-
 const permitAdditionalJSImports = (config) => {
-  const importPaths = Object.values(additionalJSImports);
-
+  // We're just gonna call all of fxa fair game ;)
+  const allFxa = resolve(__dirname, '../../')
+  const importPaths = [allFxa, resolve(__dirname, '../../../node_modules')]
```
jodyheavener, line 10: May need some additional the [tsconfigs](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/tsconfig.json#L17-L24) that previously resolved the specific paths. Or, maybe it can be moved to `tsconfig.browser.json` since it's fxa-wide now?

dannycoates: I'm not sure I follow. Each of the packages that use rescripts has it's own tsconfig and should add it's own references, which btw don't matter to react-scripts, only tsc.

dannycoates:
> Should all 3 React apps have their [moduleDependencies](https://github.com/mozilla/fxa/blob/main/package.json#L45) updated to include all fxa packages since they can technically import from anywhere now?

no, and I'd like to get rid of `moduleDependencies` in the future anyway and use yarn to resolve the internal dependency tree when picking which packages to test.

Issue #6065, opened 2020-07-28
Deploy App Store Badge Variation to 100% of Registrations


data-sync-user:
Results from our experiment show that App Store badges beat SMS and QR codes at registration. Let's roll this out to 100% of registration flows.

I'll file another task to add reminder functionality to the badge screen.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2358)
┆Issue Number: FXA-2358

data-sync-user:
➤ Alex Davis commented:

I updated this task to mention to only roll out badges for registrations.

PR #6106, opened 2020-07-31
chore(experiment): remove the qr cad experiment and apply winning treatments

vbudhram:
## Because

- The app badges treatment won for registration and QR CAD for login

## This pull request

- Removes the QR Cad experiment logic
- Shows App Badges for Signup flows, QR Code for Signin flows, Pairing for other flows (Fx toolbar menu)
- Removes the newsletter-sync-qr-code because its not need
- Clean up tests

Closes: #5835 #6147 #6146 #6065 

## Checklist


vbudhram:
@6a68 re-requested review because I made some changes to what treatment is show to a user

Summary:

- Signup flow
  - On Fx mobile - App Badges
  - On Fx Desktop - App Badges

- Login flow
  - On Fx mobile - App Badges
  - On Fx Desktop - QR CAD

- Toolbar menu
  - Signed in - Pairing
  - others - App Badges

I'm going to flag this as needs qa so they can go through each of these and verify.

vasilicatamas:
> Summary:
> 
> * Signup flow
>   
>   * On Fx mobile - App Badges 

The custom tab is automatically closed after confirming the account using the verification code and the user is redirected to Settings, therefore the app badges page is not displayed. Reproducible on both Fenix and iOS. Which should be the expected behavior here?
>   * On Fx Desktop - App Badges

Confirm that App badges page is successfully displayed.
> * Login flow
>   
>   * On Fx mobile - App Badges

The user is automatically redirected to Settings after typing their password (for accounts that have been used recently) or after confirming the accounts using the sign-in code (for accounts that have not been used for a while) Reproducible on both Fenix and iOS. Which should be the expected behavior here?

>   * On Fx Desktop - QR CAD 

This appears only for accounts that have not been used for a while and need to confirm their sign-in using the code
> * Toolbar menu
>   
>   * Signed in - Pairing

Confirm that Connect another device option from Toolbar points to pairing page
>   * others - App Badges

- What do you mean by others? Other accounts states like waiting for setup confirmation or sign-in authorization? Because in this cases Connect another device option from Toolbar is disabled
- Or are you referring to “Connect another device” option accessed from hamburger (redirects to pairing page) menu or about:sync#preferences (redirects to https://accounts.stage.mozaws.net/connect_another_device?context=fx_desktop_v3&entrypoint=preferences&service=sync&uid=9164e8e0b883430e8624bf4d845b0f07&email=vtamas%2B120%40mozilla.com)


@vbudhram Please let me know which of the above described scenarios are issues and should be tracked separately.

vbudhram:
@vasilicatamas Thanks for testing the different flows, here are updated expectations:

- Signup flow
  - On Fx mobile - No changes
  - On Fx Desktop - App Badges

- Login flow
  - On Fx mobile - No changes
  - On Fx Desktop - QR CAD

- Toolbar menu
  - Signed in - Pairing
  - ~others - App Badges~ This was just a catch all, the scenarios you described here are what I expect.

For the Fx Desktop login flow we should be defaulting to the QR CAD for all the different states, you could spin a bug off for that.

vasilicatamas:
> For the Fx Desktop login flow we should be defaulting to the QR CAD for all the different states, you could spin a bug off for that.

Filed #6165

packages/fxa-content-server/app/scripts/views/connect_another_device.js
```diff
@@ -33,54 +32,18 @@ class ConnectAnotherDeviceView extends FormView {
...
-              // New updated CAD via QR code experience
-              return this.replaceCurrentPageWithQrCadScreen(account);
-            }
-            case 'treatment-b': {
-              // Current non-SMS experience
-              return;
-            }
-          }
-        }
+    if (this.isSignIn() && this._areSmsRequirementsMet(account)) {
```
l-hedgehog, line 40: Hi @vbudhram,

We recently updated China FxA from train-181 to train-192, and some users are shown this CAD via QR screen, with the #use-sms-link in it, even without `CN` defined in [country-telephone-info.js](https://github.com/mozilla/fxa/blob/v1.192.0/packages/fxa-content-server/app/scripts/lib/country-telephone-info.js).

It looks like checking `_areSmsRequirementsMet` instead of `_isEligibleForSms` (part of original `getEligibleQrCodeCadGroup`) here might be the culprit?

PR #6108, opened 2020-07-31
bug(aet): Don't translate anon_id update errors

6a68:
## Because

- We don't intend for AET update errors to ever be displayed to users (who won't understand or be able to fix those errors)

## This pull request

- Removes the AET update precondition failed error from the list of translated error strings

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/auth-errors.js
+++ b/packages/fxa-content-server/app/scripts/lib/auth-errors.js
@@ -616,9 +616,7 @@ var ERRORS = {
   },
   ECOSYSTEM_ANON_ID_UPDATE_CONFLICT: {
     errno: 190,
-    message: t(
-      'Could not update Ecosystem Anon ID because precondition failed.'
-    ),
+    message: 'Could not update Ecosystem Anon ID because precondition failed.', // only used for logging, no need for translation
   },
 };
 


6a68:
I'm not sure if I also need to manually delete this error message from all the l10n files, or if it will be automatically removed.

packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -616,9 +616,7 @@ var ERRORS = {
   },
   ECOSYSTEM_ANON_ID_UPDATE_CONFLICT: {
     errno: 190,
-    message: t(
-      'Could not update Ecosystem Anon ID because precondition failed.'
-    ),
+    message: 'Could not update Ecosystem Anon ID because precondition failed.', // only used for logging, no need for translation
```
vbudhram, line 619: Thanks for the comment, it will help future us :)

jodyheavener: This is what I get for copy-pasting!

6a68: lol, don't feel bad - I missed it in the review :beers:

PR #6109, opened 2020-07-31
change password sketch

dannycoates:
just a sketch - nothing serious

### DIFF
--- a/packages/fxa-settings/src/components/Auth/index.tsx
+++ b/packages/fxa-settings/src/components/Auth/index.tsx
@@ -1,0 +1,51 @@
+import React, {useContext} from 'react';
+import AuthClient from 'fxa-auth-client/browser';
+import { useQuery, gql } from '@apollo/client';
+import { useAwait } from 'fxa-react/lib/hooks';
+import { AccountData } from '../AccountDataHOC/gql';
+import { sessionToken } from '../../lib/cache'
+
+export const GET_PASSWORD_INFO = gql`
+  query GetPasswordInfo {
+    account {
+      emails {
+        email
+        isPrimary
+      }
+    }
+    sessionToken @client
+    oldPassword @client
+    newPassword @client
+  }
+`
+
+export interface AuthContextValue {
+  client?: AuthClient
+};
+
+export const AuthContext = React.createContext<AuthContextValue>({})
+export {AuthClient}
+
+export function useAuth() {
+  const {client} = useContext(AuthContext)
+  if (!client) {
+    throw new Error('Are you forgetting an AuthContext.Provider?')
+  }
+  return client!;
+}
+
+export function usePasswordChanger() {
+  const auth = useAuth()
+  const { data } = useQuery(GET_PASSWORD_INFO)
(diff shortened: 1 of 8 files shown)


packages/fxa-settings/src/components/Auth/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export function useAuth() {
+  const {client} = useContext(AuthContext)
+  if (!client) {
+    throw new Error('Are you forgetting an AuthContext.Provider?')
+  }
+  return client!;
+}
+
+export function usePasswordChanger() {
```
dannycoates, line 37: This isn't a component so it will move somewhere else. I'm still trying to decide if hooks are the right style for these operations.

packages/fxa-settings/src/components/Auth/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const auth = useAuth()
+  const { data } = useQuery(GET_PASSWORD_INFO)
+  const primaryEmail = (data.account as AccountData).emails.find((email) => email.isPrimary)!;
+  const [passwordChangeState, execute, reset] = useAwait(async () => {
+    const response = await auth.passwordChange(primaryEmail.email, data.oldPassword, data.newPassword, { sessionToken: data.sessionToken })
+    sessionToken(response.sessionToken)
```
dannycoates, line 43: sets the local state with the new sessionToken

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -22,8 +24,11 @@ export const Settings = ({ account }: { account: AccountData }) => {
 
   const modalHeaderId = 'modal-header-verify-email';
   const modalDescId = 'modal-desc-verify-email';
+  oldPassword('asdzxcasd')
```
dannycoates, line 27: all the additions in this file are for testing out the hook. it'll all move into whatever change password component we make

packages/fxa-settings/src/index.tsx
```diff
@@ -30,25 +23,43 @@ try {
...
+    link: authLink.concat(httpLink),
+    typeDefs
   });
   const queryParams = searchParams(window.location.search);
 
   render(
     <React.StrictMode>
       <ApolloProvider client={apolloClient}>
-        <App {...{ queryParams }} />
+        <AuthContext.Provider value={{client: authClient}}>
```
dannycoates, line 60: here we set the authClient into a react context. idk if context is the best pattern for it yet but i started by mimicking the apollo client

packages/fxa-settings/src/lib/cache.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const typeDefs = gql`
+  extend type Query {
+    sessionToken: String!
+    oldPassword: String
+    newPassword: String
+  }
+`;
+
+export const cache = new InMemoryCache({
```
dannycoates, line 32: most examples i see break out the cache into it's own file. here you'll see i've made a couple reactive variables to hold old/new password. I imagine we'll do so with other local/non-persistent state.

dannycoates:
done with this

Issue #3673, opened 2019-12-13
Gracefully fail on invalid query parameters


clouserw:
In #3645 we're showing an error to an end user about an invalid utm parameter.  We should gracefully fail if utm parameters are wrong.  (Although logging would be great)

Maybe we should gracefully fail on non-utm query parameters also?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-20)
┆Issue Number: FXA-20

6a68:
Gracefully failing here probably means allowing the user to complete their sign in flow, as long as any required query params are present and valid.

davismtl:
@6a68 mentioned in a meeting that we should prob have RP put the utm params at the end of the URL so that any mistakes don't break the URL and so that the form submit can still be successful.

clouserw:
https://github.com/mozilla/fxa/issues/5991 has an example and a screenshot

rlr:
Until this is fixed, we're going to have to stop passing along the utm_* parameters we get to FxA. We have pretty much no control over what utm_* parameters we get from the wild so it's pretty bad UX to show the error to the user and not let them proceed.

chrismore:
Any idea how often this happens? I'm concerned now that we have a paid product using fxa and lots of inbound traffic to subscription pages (with utms) that this error could be leading to lost subscriptions or cancels.

What is the regex or definition of a valid or invalid parameter value?

data-sync-user:
➤ Wil Clouser commented:

This is a regular expression which we could loosen, and also we should log this silently – not show anything to the user.

PR #6117, opened 2020-08-03
fix(params): Gracefully handle invalid utm params

vbudhram:
## Because

- We shouldn't stop a user from logging in if we detect invalid utm params

## This pull request

- Replaces the invalid utm param with `invalid` and logs a Sentry error. Hopefully we can see how often this happens and what RPs are sending
- By replacing the utm param, it allows us to still submit the metrics

Closes: #3673 

## Checklist

## Testing

Try to login with an invalid utm param set. Ex [http://localhost:3030/signin_token_code?context=fx_desktop_v3&entrypoint=fxa_discoverability_native&action=email&service=sync&utm_term=(not%20ready)](http://localhost:3030/signin_token_code?context=fx_desktop_v3&entrypoint=fxa_discoverability_native&action=email&service=sync&utm_term=(not%20ready))

### Before
Screen Shot 2020-08-03 at 1 23 18 PM

### After
Screen Shot 2020-08-03 at 1 23 59 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/metrics.js
+++ b/packages/fxa-content-server/app/scripts/lib/metrics.js
@@ -16,6 +16,7 @@
 
 import $ from 'jquery';
 import _ from 'underscore';
+import AuthErrors from '../lib/auth-errors';
 import Cocktail from 'cocktail';
 import Constants from './constants';
 import Backbone from 'backbone';
@@ -27,6 +28,7 @@ import speedTrap from 'speed-trap';
 import Strings from './strings';
 import SubscriptionModel from 'models/subscription';
 import xhr from './xhr';
+import Validate from '../lib/validate';
 
 // Speed trap is a singleton, convert it
 // to an instantiable function.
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/app/scripts/lib/validate.js
```diff
@@ -273,6 +275,16 @@ var Validate = {
...
   },
+
+  /**
+   * Check if the utm param is valid.
+   *
+   * @param {String} value
+   * @returns {Boolean}
+   */
+  isUtmValid(value) {
+    return utmRegex.test(value) && value.length <= 128;
```
6a68, line 286: Instead of specifying length separately, could you replace the `+` with `{0,128}` in the regex?

packages/fxa-content-server/app/scripts/lib/metrics.js
```diff
@@ -115,6 +118,19 @@ function marshallProperty(property) {
   }
 }
 
+function marshallUtmProperty(property) {
+  if (
+    property &&
+    property !== NOT_REPORTED_VALUE &&
+    Validate.isUtmValid(property)
+  ) {
+    if (Validate.isUtmValid(property)) {
```
6a68, line 127: This syntax seems a little funny to me. Why are you checking `isUtmValid` twice?

vbudhram: Whoops, removing

PR #6123, opened 2020-08-03
fix(payments): ensure Customer data is not cached with an expiration TTL

### LINKED ISSUE #6097: “Subscriptions” option from Manage Account page disappears in aprox 10 minutes after a subscription
#### Affected versions:
- All Firefox;

#### Affected FxA versions:
- 1.182.0;

#### Affected Platforms:
- Windows 10 (64-bit);

#### Prerequisites:
- Have an FxA account;

 #### Steps to reproduce:
1. Load https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net/vpn page;
2. Click on the “Try Mozilla VPN” button;
3. Finalize subscription and click “Continue to download” button;
4. Click on the “Manage account” option from page header user menu;
5. Wait around 10 minutes;
6. Access the “Manage Account” page;
7. Observe the “Subscriptions” option;

#### Expected result:
- “Subscriptions” option is listed. User can click it and be redirected to the “Subscriptions” page.

#### Actual result:
- “Subscriptions” option disappears. 

#### Notes:
- Sign in out/in does not fix the problem.
- Making a “cancel subscription” action in the Subscription page (accessed from Mozilla VPN page - user menu - Manage subscriptions) will make the option reappear in manage account page.
disap

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2369)
┆Issue Number: FXA-2369

lmorchard:
- removes global TTL for StripeHelper cacheManager

- sets per-method ttlSeconds based on config setting

- tests to assert the TTL changes

- some TODOs for follow up issues

fixes #6097

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -93,7 +93,7 @@ export class StripeHelper {
   // Note that this isn't quite accurate, as the auth-server logger has some extras
   // attached to it in Hapi.
   private log: Logger;
-  private cacheTtlSeconds: number;
+  private plansAndProductsCacheTtlSeconds: number;
   private webhookSecret: string;
   private stripe: Stripe;
   private redis: ioredis.Redis | undefined;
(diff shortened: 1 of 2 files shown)


lmorchard:
FWIW, I was able to verify this did the expected thing locally by peeking into redis directly.

Before:
```
ttl subhub:customer-117bd7bfe4bd4184b34ed55c41021387|foo@example.com
:594
```
After: 
```
ttl subhub:customer-117bd7bfe4bd4184b34ed55c41021387|foo@example.com
:-1
```

lmorchard:
Also a fun fact I discovered in type-cacheable is that if `ttlSeconds: 0` is specified in the `@Cacheable` decorator, [it falls back to the global cacheManager setting](https://github.com/joshuaslate/type-cacheable/blob/a2223dbf09964bf066b6219fab79cf2d4ee936ca/packages/core/lib/decorators/Cacheable.ts#L53). 

This can be worked around by using a TTLBuilder function instead - because at least they're not false-y values. But by the point I figured out how to use TTLBuilder functions, I just went ahead and gave every method its own TTLBuilder and dumped the global setting.

Issue #4911, opened 2020-04-13
[main] Build out Settings > Security


data-sync-user:
Using the reusable UnitRow or a similar component, create and style the Security section.

Add the routes for `account_recovery` and  `two_step_authentication` on the corresponding buttons, setting them up for "inner view" functionality work.
"Why does resetting my password reset my data?" link should link-out to: https://support.mozilla.org/en-US/kb/reset-your-firefox-account-password-recovery-keys

Recovery key: hard code "Not set" and "Create" for now. Real data will be displayed in FXA-1615.
Two-step auth: hard code "Not set" and "Add" for now. Real data will be displayed in FXA-1616.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1614)
┆Issue Number: FXA-1614

data-sync-user:
➤ Dave Justice commented:

Filed https://jira.mozilla.com/browse/FXA-2384 for the updates to UnitRow component that will need to be made in order to have the differing buttons/CTA in this panel as well as connected services. Seemed to make the most sense to separate it out as a separate task so it can be made to work for both panels.

data-sync-user:
➤ Dave Justice commented:

[~lzugai@mozilla.com] got an A11y question for ya, Should the panel's outermost container element have a `role`?

data-sync-user:
➤ Dave Justice commented:

In progress pr at https://github.com/mozilla/fxa/pull/6128, ( https://github.com/mozilla/fxa/pull/6128, ) working through a couple questions in the airplane team channel, should have this wrapped up near end of today/early wednesday morning

Issue #4913, opened 2020-04-13
[main] Build out Settings > Security > Two-step auth + disable functionality


data-sync-user:
If the user doesn't have two-step auth set, it should say "Not set" and the button should read as "Create". Otherwise, it should read as "Enabled" and the button should read "Change" with a button labeled "Disable".

* Upon clicking "Disable," a modal should show, allowing the user to confirm or cancel disabling.
* "Cancel" should close the modal, and "Disable" should disable two-factor auth with a GQL mutation and display the success/error message to the user in the AlertBar.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1616)
┆Issue Number: FXA-1616

data-sync-user:
➤ Dave Justice commented:

This may expand to include https://jira.mozilla.com/browse/FXA-1618 in the same commit

data-sync-user:
➤ Dave Justice commented:

bumping to future sprint, as https://jira.mozilla.com/browse/FXA-1648 still has some questions we are figuring out.

Issue #4910, opened 2020-04-13
[main] Build out Settings > Security > Recovery key revoke functionality


data-sync-user:
The modal should already be built prior to starting on this issue.

If the user doesn't have the recovery key set, it should say "Not set" and the button should read as "Create". Otherwise, it should read as "Enabled" and the button should read "Remove".

Upon clicking on "Remove," a modal should show, allowing the user to confirm or cancel revoking. Allow for this to work functionally.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1615)
┆Issue Number: FXA-1615

data-sync-user:
➤ Danny Coates commented:

Bumped to 5 because this involves getting the ModalVerifySession component going and adding a mutation to graphql api

PR #6128, opened 2020-08-04
task(fxa-settings): Add security panel to settings

meandavejustice:
- Adds the security panel to settings

### _notes:_
- Buttons will be updated once (newly filed)  is complete.
- [figma link](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A1)
- Enabled/fresh load is hard coded (as suggested in issue), this will be updated in the future once we get the data flowing

Closes: #4911

## Checklist

## Screenshots (Optional)
image

image

### DIFF
--- a/packages/fxa-settings/src/components/Security/index.test.tsx
+++ b/packages/fxa-settings/src/components/Security/index.test.tsx
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import Security from '.';
+
+describe('Security', () => {
+  it('renders "fresh load" <Security/> with correct content', async () => {
+    const { findByText, findAllByText } = render(
+      <Security
+        accountRecoveryKeyEnabled={false}
+        twoFactorAuthEnabled={false}
+      />
+    );
+
+    expect(await findByText('Recovery key')).toBeTruthy;
+    expect(await findByText('Two-step authentication')).toBeTruthy;
+
+    const result = await findAllByText('Not Set');
+    expect(result).toHaveLength(2);
+  });
+
+  it('renders "enabled two factor" and "recovery key present" <Security/> with correct content', async () => {
+    const { findAllByText } = render(
+      <Security accountRecoveryKeyEnabled={true} twoFactorAuthEnabled={true} />
+    );
+
+    const result = await findAllByText('Enabled');
+    expect(result).toHaveLength(2);
+  });
+});
--- a/packages/fxa-settings/src/components/Security/index.stories.tsx
+++ b/packages/fxa-settings/src/components/Security/index.stories.tsx
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Security } from '.';
+
+storiesOf('Components|Security', module)
+  .add('default', () => (
+    <Security accountRecoveryKeyEnabled={false} twoFactorAuthEnabled={false} />
+  ))
+  .add('account recovery key set and two factor enabled', () => (
+    <Security accountRecoveryKeyEnabled={true} twoFactorAuthEnabled={true} />
+  ));
--- a/packages/fxa-settings/src/components/Security/index.tsx
+++ b/packages/fxa-settings/src/components/Security/index.tsx
@@ -1,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { UnitRow } from '../UnitRow';
+import LinkExternal from 'fxa-react/components/LinkExternal';
+
+type SecurityProps = {
+  twoFactorAuthEnabled: boolean;
+  accountRecoveryKeyEnabled: boolean;
+  className?: string;
+};
+
+export const Security = ({
+  twoFactorAuthEnabled,
+  accountRecoveryKeyEnabled,
+  className,
+}: SecurityProps) => {
+  const getValue = (settingOption: boolean) =>
+    settingOption ? 'Enabled' : 'Not Set';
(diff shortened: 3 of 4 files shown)


packages/fxa-settings/src/components/SecurityPanel/index.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+import SecurityPanel from '.';
+import { assert } from 'console';
+
+describe('SecurityPanel', () => {
+  it('renders SecurityPanel with expected attributes', () => {
+    render(<SecurityPanel />);
+
+    // auto fail this one so I remember to come back and fix it.
+    // - Do these panels need a 'role'?
+    expect(screen.getByTestId('settings-security')).toBeCalled;
```
meandavejustice, line 17: @LZoog Mentioned this in the jira ticket, but wondering if you think it makes sense to set a `role` on these panels?

packages/fxa-settings/src/components/SecurityPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  accountRecoveryKey?: object;
+  className?: string;
+};
+
+export const SecurityPanel = ({
+  twoFactorAuth,
+  accountRecoveryKey,
+  className,
+}: SecurityPanelProps) => {
+  const getValue = (settingOption?: object) =>
```
LZoog, line 19: `getValue` and `getClassName` gets us a head start in #4910 and #4913. You could remove these and hardcode the values to close this issue, otherwise we're probably going to want to require props like:

```
type SecurityPanelProps = {
  twoFactorAuthEnabled: boolean;
  accountRecoveryKeyEnabled: boolean;
  className?: string;
};
```

Then on L128 in `Settings/index.tsx`, you can pass in the real data there from `accountData` to get us further ahead on those two issues, or just pass in fake data (probably `false` since the ticket asks for "Not set" to be hardcoded).

packages/fxa-settings/src/components/SecurityPanel/index.test.tsx
LZoog: As mentioned in Slack I don't believe we need that here. We have `role="img" for inlined SVGs but this should be read properly/by our HTML semantics (a new section/divs).

packages/fxa-settings/src/components/SecurityPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+          </a>
+        </UnitRow>
+
+        <hr className="unit-row-hr" />
+
+        <UnitRow
+          header="Two-step authentication"
+          headerValueClassName={getClassName(twoFactorAuth)}
+          headerValue={getValue(twoFactorAuth)}
+          route="/settings/two_step_authentication"
```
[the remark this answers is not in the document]
LZoog: @meandavejustice sorry actually, in your branch, this is good for now, we might change this when the React Router PR is merged

packages/fxa-settings/src/components/SecurityPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <section className="mt-11" id="security" data-testid="settings-security">
+      <h2 className="font-header font-bold ml-4 mb-4">Security</h2>
+      <div className="bg-white tablet:rounded-xl shadow">
+        <UnitRow
+          header="Recovery key"
+          headerValueClassName={getClassName(accountRecoveryKey)}
+          headerValue={getValue(accountRecoveryKey)}
+          route="/settings/account_recovery"
```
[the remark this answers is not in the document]
LZoog: Same comment as before, this is actually right for now / we may need to update it when React Router is in

packages/fxa-settings/src/components/SecurityPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const SecurityPanel = ({
+  twoFactorAuth,
+  accountRecoveryKey,
+  className,
+}: SecurityPanelProps) => {
+  const getValue = (settingOption?: object) =>
+    settingOption ? 'Enabled' : 'Not Set';
+  const getClassName = (settingOption?: object) =>
+    settingOption ? 'text-green-800 tracking-wider' : 'tracking-wider';
```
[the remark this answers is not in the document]
meandavejustice: Oh I see what I did here, copied the format from the profile section and `tracking-wider` was copied from the password section, which makes sense for password but not the other sections. Nice catch! removing.

packages/fxa-settings/src/components/Security/index.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import Security from '.';
+
+const { findByText, findAllByText } = screen;
+
+describe('SecurityPanel', async () => {
+  it('renders "fresh load" SecurityPanel with correct content', () => {
```
LZoog, line 13: We should probably update this from `SecurityPanel` to `Security` now.

Also when running these tests, I'm seeing this: "Warning: You seem to have overlapping act() calls, this is not supported. Be sure to await previous act() calls before making a new one." 🤔

packages/fxa-settings/src/components/Security/index.test.tsx
```diff
@@ -0,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import Security from '.';
+
+const { findByText, findAllByText } = screen;
```
LZoog, line 10: I'm not sure these should be global for the entire file, or at least we haven't destructured these out of `screen` before - we've typically just used `screen.whatevermethod` because that allows you to see the available methods in VSCode (after typing `screen.`), or we use `const { findByText } = render(<SomeComponent`. We discussed this a bit [here](https://github.com/mozilla/fxa/pull/5690#discussion_r443627367) if you want to read more.

packages/fxa-settings/src/components/Security/index.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Security
+        accountRecoveryKeyEnabled={false}
+        twoFactorAuthEnabled={false}
+      />
+    );
+
+    expect(findByText('Recovery key')).toBeTruthy;
+    expect(findByText('Two-step authentication')).toBeTruthy;
+
+    findAllByText('Not Set').then((result) => {
```
LZoog, line 24: Is it possible to use `await` here instead of `.then`?

meandavejustice: I tried that but couldn't get it working like I wanted, searched around for examples with await in fxa-settings and didn't see any, so went with this

Issue #4899, opened 2020-04-13
[main] Setup React router + separate page functionality & shared component


data-sync-user:
Setup the router with lazy loading and allow for a separate page to be hit, hiding the nav on the left. This can be a dummy view that will be removed with the first actual separate page. Add the routes for `avatar/change`, `display_name`, and `change_password` on the corresponding buttons in the Profile section, setting them up for "inner view" functionality work.

Also create a simple wrapper component (such as `<FlowContainer>`, or maybe you have a better name in mind) with classes/styles to be shared across all separate pages. This can be pretty rough at this point but should include the "back" arrow shown in the visual designs, which should be a link.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1606)
┆Issue Number: FXA-1606

Issue #5028, opened 2020-04-20
[change pw] Create `/change_password` page with dummy form


data-sync-user:
The `/change_password` route should be hooked up from the settings page in FXA-1608.

Add the text and the form for this page and style it accordingly. Add the "Forgot password?" link. The inputs can be completely "dummy" at this point.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/397196811

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1705)
┆Issue Number: FXA-1705

PR #6131, opened 2020-08-04
feat(router): Add reach router support to settings

vbudhram:
## Because

- The new settings page needs navigation

## This pull request

- Adds initial react reach routing and lazy loading with Suspense
- Adds a basic `FlowContainer` component that acts as container for standalone views
- There are still some rough edges around the styling that I hope can be addressed in #5028 (ie mobile view doesn't show the back arrow in correct spot)

Closes: #4899 

## Checklist

## Screenshots

Storybook:
Screen Shot 2020-08-05 at 9 38 24 PM


packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -2,22 +2,23 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React from 'react';
+import React, { Suspense } from 'react';
```
dannycoates, line 5: dun-dun-DUH! 🕵️‍♀️

jodyheavener, line 5: Is this used tho? 👀

packages/fxa-settings/src/components/FlowContainer/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const FlowContainer = (_: RouteComponentProps) => {
+  const location = useLocation();
+
+  return (
+    // TODO: Style this per the style guide
+    <p>Href: {location.pathname}</p>
+  );
+};
+
+export default FlowContainer;
```
jodyheavener, line 17: Wanna add this to Storybook?

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -9,8 +9,11 @@ import UnitRowWithAvatar from '../UnitRowWithAvatar';
 import Modal from '../Modal';
 import AlertBar from '../AlertBar';
 import { AccountData } from '../AccountDataHOC/gql';
+import { RouteComponentProps } from '@reach/router';
```
jodyheavener, line 12: I was complaining about something similar to @LZoog earlier today, so maybe this could be in a future refactor, but I think if we add `@reach/router` to the tsconfig's `typeRoots` we should be able to avoid needing to import type defs. I think 🤔

packages/fxa-settings/src/components/App/index.test.tsx
```diff
@@ -43,7 +43,7 @@ it('redirects to /get_flow when flow data is not present', async () => {
   );
 });
 
-it('redirects to /get_flow when flow data is not present', async () => {
+it("doesn't redirect to /get_flow when flow data is present", async () => {
```
jodyheavener, line 46: I'm surprised this didn't pre-commit format these double quotes to singles 😮

packages/fxa-settings/src/components/Settings/index.tsx
vbudhram: There might be a better way to do this but it was mentioned in the react reach docs. Maybe its ok for now?

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -29,7 +29,17 @@ export const App = ({ queryParams }: AppProps) => {
             }
             hasSubscription={Boolean(account.subscriptions.length)}
           >
-            <Settings {...{ account }} />
+            <Suspense fallback="">
```
vbudhram, line 32: Not really sure what the fallback here should be?

packages/fxa-settings/src/components/App/index.tsx
jodyheavener: Ignore this, I see you've added it

packages/fxa-settings/src/components/Settings/index.tsx
jodyheavener: Oh yeah it's totally fine. :)

jodyheavener:
Perhaps this is also intended to be addressed in future work, but I noticed the clickable area of the back arrow expands to the height of the content box. Maybe a byproduct of the flex layout?

Screen Shot 2020-08-06 at 10 08 30 AM

We should also consider a more visible/styled focus state, but I'm not seeing a specific style in the mocks for it. Again, this might be suitable for future work though.

package.json
```diff
@@ -100,6 +100,7 @@
     ]
   },
   "devDependencies": {
+    "@types/reach__router": "^1.3.5",
```
jodyheavener, line 103: Did this get auto-added? Seems like it shouldn't be necessary as it's added [here](https://github.com/mozilla/fxa/pull/6131/files#diff-6f49eaae21bfa6a37a17d0c80d18d5bdR64) and only used within fxa-settings.

packages/fxa-settings/src/components/App/index.tsx
dannycoates: what are we suspending on here?

vbudhram: The payments server uses it for loading the router routes, https://github.com/mozilla/fxa/blob/9592b42337a75fdddefebbcf579408b7254be26b/packages/fxa-payments-server/src/App.tsx#L101. In my testing there seems to be a loading indicator on the page, but can't track down where/how that gets into the page. Maybe Suspense is not needed?

dannycoates: I don't _think_ it's needed. our spinner is in AccountDataHOC

packages/fxa-settings/src/components/FlowContainer/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+        to={backRoutePath}
+        className="pt-7 pr-6"
+        data-testid="flow-container-back-btn"
+      >
+        <BackArrow className="inline h-8 w-8" />
+      </Link>
+      <div className="bg-white shadow container pl-6 pr-10 pb-10">
+        <h1 className="font-header pt-8 pb-4">{title}</h1>
+        <div>
+          <p>Href: {location.pathname}</p>
```
dannycoates, line 33: Is this going to be where `{children}` goes?

vbudhram: @dannycoates ack yea it will be, updating 👍

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -56,13 +58,13 @@ export const UnitRow = ({
       <div className="unit-row-actions">
         <div>
           {route && (
-            <a
+            <Link
               className="cta-neutral transition-standard"
               data-testid="unit-row-route"
-              href={route}
+              to={route + queryParams}
```
LZoog, line 64: Do we need to pass `queryParams` down for metrics? We'll have them stored in that `queryParams` const when the application loads.

If they don't change and we need them in our components we could also use `Context` or store this in apollo cache to read where we need them.

packages/fxa-settings/src/components/UnitRowWithAvatar/index.tsx
```diff
@@ -25,13 +30,13 @@ export const UnitRowWithAvatar = ({ avatarUrl }: UnitRowWithAvatarProps) => {
       </div>
       <div className="unit-row-actions">
         <div>
-          <a
+          <Link
             className="cta-neutral transition-standard"
             data-testid="unit-row-with-avatar-route"
-            href="#"
+            to={'/beta/settings/avatar/change' + queryParams}
```
LZoog, line 36: Same question around needing to pass `queryParams` down into the URL. Also, I guess there's not a way to just make this `/avatar/change` instead of spelling out `/beta/settings/avatar/change`? I guess a find and replace for `/beta/settings` to `/settings` when we move this out of beta won't be so bad.

FWIW this can also be:

```
to={`/beta/settings/avatar/change${queryParams}`}
```

packages/fxa-settings/src/components/App/index.tsx
LZoog: If the content inside `FlowContainer` hangs for some reason I think we could use the spinner here, but I don't know that would happen in Settings because we're not querying anything in the flows, and even if we were, GQL provides that `loading` check that yeah in AccountDataHOC we use to display a loading spinner so I don't _think_ a fallback here would ever be shown.

packages/fxa-settings/src/components/FlowContainer/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const backRoutePath =
+    (backRoute ? backRoute : '/beta/settings') + location.search;
+  return (
+    <div className="flex" data-testid="flow-container">
+      <Link
+        to={backRoutePath}
+        className="pt-7 pr-6"
+        data-testid="flow-container-back-btn"
+      >
+        <BackArrow className="inline h-8 w-8" />
```
LZoog, line 28: This could use a `role="img"`. Also an `aria-label="alt text"`  _or_ the link can have a `title="alt text"` if it should have hover text. Might check with Meridel W. on what copy should be read and if this should have a title.

packages/fxa-settings/src/components/UnitRow/index.tsx
vbudhram: > Do we need to pass queryParams down for metrics?

No, I don't believe so. I can use the `Context` value here.

packages/fxa-settings/src/components/UnitRowWithAvatar/index.tsx
vbudhram: > instead of spelling out /beta/settings/avatar/change?

I couldn't really find a good way to do this that didn't seem excessively complicated. It is probably ok for now.

packages/fxa-settings/src/components/AppLayout/index.test.tsx
```diff
@@ -7,15 +7,30 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import { MockedCache } from '../../models/_mocks';
 import AppLayout from '.';
+import {createHistory, createMemorySource, LocationProvider} from "@reach/router";
 
-it('renders the app with children', () => {
-  const { getByTestId } = render(
+function renderWithRouter(
```
vbudhram, line 12: @jodyheavener @LZoog @dannycoates is there a better way to mock providers?

dannycoates: idk 🤷

LZoog: `history={history}` can become {...{history}}`, but without looking/poking further I don't know of a better way besides perhaps sticking this function in another file to import so we don't have it defined x3 times?

LZoog: @vbudhram if you pull latest, we do have `models/_mocks` with a `MockedCache` provider there where a helper for this could live. There's also a [mocked AlertBarContext provider](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/AlertBarContext.tsx#L24-L34) in another file that could be moved here to get all our mocked providers in one spot.

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -34,7 +37,7 @@ export const Settings = ({ account }: { account: Account }) => {
             header="Password"
             headerValueClassName="tracking-wider"
             headerValue="••••••••••••••••••"
-            route="#"
+            route="/beta/settings/change_password"
```
jodyheavener, line 40: I'm [currently discussing](https://github.com/mozilla/fxa/pull/6187#discussion_r468886110) this with @LZoog, but I'd like us to consider constantizing paths. Would love your input!

vbudhram: I was going to take a look at doing that in https://jira.mozilla.com/browse/FXA-1705

jodyheavener: Works for me!

vbudhram:
>  but am happy to punt that to a different issue as long as it's tracked

Yea my plan was to tackle this in https://jira.mozilla.com/browse/FXA-1705

Issue #5018, opened 2020-04-20
[metrics] Take an event and send it to the server


data-sync-user:
Confirm the metrics setup user story can be closed by sending an event to the server. See [this list](https://docs.google.com/spreadsheets/d/1nWDBUhn7HqPaW5C4VughcMfceX7LrJofoC_XICnpIf8/edit#gid=610481725) of events.

Would it be more ideal to batch these (send beacon)? If so, there will be an increased scope.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1695)
┆Issue Number: FXA-1695

PR #6140, opened 2020-08-05
feat(settings): metrics reporting

jodyheavener:
## Because

- The new Settings app needs to be able to report metrics events

## This pull request

- Starts implementing full metrics reporting to the content server metrics endpoint

Closes: #5018 

## Checklist

## Other information

- I did my best to translate what I found in the content server metrics library, but it's certainly possible I've misunderstood certain concepts.
- I suspect a few things in metrics reporting will evolve as the Settings redesign progresses. Please observe the notes and TODOs I've left.
- I've renamed `flow-event.ts` to `metrics.ts` because it does much more now.


jodyheavener:
I still need to expand on tests, but wanted to get some early eyes on it. @vbudhram if you get some time this week would you be able to take a peak?

jodyheavener:
**Reviewer heads up!** GitHub is collapsing `metrics.test.ts`

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -0,0 +1,377 @@
...
+    console.error('AppError', e);
+    sentryMetrics.captureException(e);
+  }
+}
+
+/**
+ * Log an event with the view name as a prefix
+ */
+export function logViewEvent(
+  /** Name of the view */
```
dannycoates, line 267: please don't comment self-evident parameters

packages/fxa-settings/package.json
```diff
@@ -42,7 +42,8 @@
     "react-dom": "^16.13.1",
     "react-scripts": "^3.4.1",
     "subscriptions-transport-ws": "^0.9.11",
-    "typescript": "3.8.3"
+    "typescript": "3.8.3",
+    "uuid": "^8.3.0"
```
dannycoates, line 46: it'd be cool if we could either use `uuid@^7.0.3` here or update the other packages to 8. run `yarn why uuid`

packages/fxa-settings/src/lib/metrics.ts
jodyheavener: I can see why it's redundant and I'll remove, but I generally lean toward more documentation than less; if anything the comment could be updated to be more descriptive.

packages/fxa-settings/package.json
jodyheavener: Oh neat. Thanks for catching this. Will use the existing version.

packages/fxa-settings/src/lib/metrics.ts
dannycoates: We should strive for consistent comment style across the project. Since most of our packages were separate and maintained by different people we've got a long way to go, but if you'd like to propose a target style, please do it in a form we can discuss as a team. I don't want to have a mix of everyone's individual styles. In the meantime try to roughly match the rest of the project. We've got more function header comments with param descriptions than inline in the declaration, for example.

jodyheavener: I totally agree - consistent comment style is important, and I can definitely update these code comments to live at function headers where appropriate. However I would still argue that a more descriptive argument comment is better than none at all. I do concede that my comment in its original form was redundant, though.

packages/fxa-content-server/app/scripts/views/get_flow.js
```diff
@@ -58,9 +63,14 @@ class GetFlowView extends BaseView {
...
-      flow_begin_time: flowBeginTime,
-      flow_id: flowId,
+      broker,
+      context,
+      deviceId,
+      flowBeginTime,
+      flowId,
+      isSampledUser,
+      service,
+      uniqueUserId,
```
jodyheavener, line 73: Some things that the metrics endpoint requires are just not set up in fxa-settings, so for now we can steal it during the get_flow process.

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -69,6 +69,11 @@ export const App = ({ queryParams }: AppProps) => {
   }
   const account = data!.account;
 
+  Metrics.setProperties({
+    lang: document.querySelector('html')?.getAttribute('lang'),
+    uid: account.uid,
+  });
```
jodyheavener, line 75: These two value could arguably also come from the get_flow process, but ideally we're not relying on it where we don't have to.

packages/fxa-settings/README.md
```diff
@@ -203,6 +203,30 @@ const LogoImage = () => (
...
+#### Event logging
+
+Log events to record when a user completes a measurable action. All previously mentioned payload data is included each time one of the following logging functions are called:
+
+- `logViewEvent(viewName, eventName, eventProperties)` can be used to record that a particular view (a "page") was visited.
+- `logExperiment(choice, group, eventProperties)` can be used to log the outcome of an experiment the user is participating in. This also calls `addExperiment` with the same choice and group.
+
+All logging methods have the argument `eventProperties`, which can be used to supply event-specific information to the payload.
+
+**Note:** take care when calling these methods as they attempt to log the event immediately. When logging view events inside React Components you'll want to place the call inside a `useEffect` hook to only execute on component render.
```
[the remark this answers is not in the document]
jodyheavener: That would probably be helpful. React has a way of detecting if a hook was not called at the top level of a function component body, and will actually come to your home with a lawyer to tell you you've broken then [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). But I wouldn't know where to begin with implementing something similar.

jodyheavener:
> If we aren’t going to keep this around in the future, do we have an issue where this is being tracked?

Good point. Just made one: https://jira.mozilla.com/browse/FXA-2419

packages/fxa-settings/README.md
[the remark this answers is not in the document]
dannycoates: I haven't read the whole diff yet, but if these are meant to be used like that we could make wrapper hook functions for them

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -0,0 +1,331 @@
...
+ * endpoint via navigator.sendBeacon
+ *
+ * @param eventData
+ */
+function postMetrics(eventData: EventData) {
+  if (!initialized || !window.navigator.sendBeacon) {
+    return;
+  }
+
+  window.navigator.sendBeacon('/metrics', JSON.stringify(eventData));
```
dannycoates, line 115: this is fine but fyi `fetch()` with `keepalive` is supposed to replace `sendBeacon` though [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) and caniuse aren't clear about support level

jodyheavener: Oh interesting. I think content server falls back to standard XHR if it's not present. I can prob do the same here.

packages/fxa-settings/README.md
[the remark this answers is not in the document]
jodyheavener: The one for view events could probably become a hook that wraps useEffect, but the others should be able to be called as needed 👍

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/**
+ * Reset Metrics setup; used for testing
+ */
+export function reset() {
+  initialized = false;
+  viewNamePrefix = null;
```
vbudhram, line 123: There is a ton of hidden complexity in the `viewNamePrefix`, and it was mostly used for flows involving signin/signup. I think its ok to leave leave this since it would be nice to refactor away from this.

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+}
+
+/**
+ * Log when a marketing snippet is shown to the user
+ *
+ * @param url - URL of marketing link
+ * @param campaignId - Marketing campaign id
+ */
+export function addMarketingImpression(url: string, campaignId?: string) {
```
vbudhram, line 304: Ah this might be another thing that is used in the signup/signin flows.

packages/fxa-settings/src/lib/metrics.ts
jodyheavener: Yeah, there's definitely a lot going on in that existing metrics lib 😄 Just to confirm, you think we can leave it like this? Or maybe I should remove it since we don't deal with signup/in flows in Settings?

packages/fxa-settings/README.md
```diff
@@ -203,6 +203,30 @@ const LogoImage = () => (
...
+Metrics reports are currently sent to the fxa-content-server metrics endpoint. Use the [metrics library](./src/lib/metrics.ts) to log events and other information.
+
+#### Payload data
+
+Relevant environment and account data, such as window measurements, user locale, and flow data are automatically included in metrics payloads, however additional data can be set as needed:
+
+- `setProperties({ key: value })` can be used to configure additional information about the user's environment and session. Refer to `ConfigurableProperties` in the metrics library for the properties that can be configured.
+- `setUserPreference` can be used to log when a user preference is updated.
+- `addExperiment(choice, group)` can be used to add details about an experiment the user is participating in.
+- `addMarketingImpression(url, campaignId)` and `setMarketingClick(url, campaignId)` can be used to add details about a marketing flow the user is a part of, and whether or not a marketing link was interacted with.
```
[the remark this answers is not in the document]
jodyheavener: I can add to my [followup issue](https://jira.mozilla.com/browse/FXA-2419) to verify how the marketing stuff should work.

PR #6141, opened 2020-08-05
feat(settings): add session verified state to graphql-api

dannycoates:
adding session verified state to gql and the frontend

### DIFF
--- a/packages/fxa-graphql-api/src/lib/auth.ts
+++ b/packages/fxa-graphql-api/src/lib/auth.ts
@@ -13,10 +13,9 @@ export class SessionTokenAuth {
   @Inject(fxAccountClientToken)
   private authClient!: AuthClient;
 
-  public async lookupUserId(sessionToken: string): Promise<string> {
+  public async getSessionStatus(sessionToken: string) {
     try {
-      const result = await this.authClient.sessionStatus(sessionToken);
-      return result.uid;
+      return await this.authClient.sessionStatus(sessionToken);
     } catch (err) {
       throw new AuthenticationError('Invalid session token');
     }
--- a/packages/fxa-graphql-api/src/lib/resolvers/session-resolver.ts
+++ b/packages/fxa-graphql-api/src/lib/resolvers/session-resolver.ts
@@ -1,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { GraphQLResolveInfo } from 'graphql';
+import {
+  Ctx,
+  Info,
+  Query,
+  Resolver,
+} from 'type-graphql';
+import { Context } from '../server';
+import { Session as SessionType } from './types/session'
+
+@Resolver((of) => SessionType)
+export class SessionResolver {
+
+  @Query((returns) => SessionType)
+  session(@Ctx() context: Context, @Info() info: GraphQLResolveInfo) {
+    context.logger.info('session', { uid: context.session.uid });
+    return {
+      verified: context.session.state === 'verified'
+    } as SessionType
+  }
+}
--- a/packages/fxa-graphql-api/src/lib/server.ts
+++ b/packages/fxa-graphql-api/src/lib/server.ts
@@ -12,6 +12,7 @@ import { SessionTokenAuth } from './auth';
 import { AuthServerSource } from './datasources/authServer';
 import { ProfileServerSource } from './datasources/profileServer';
 import { AccountResolver } from './resolvers/account-resolver';
+import { SessionResolver } from './resolvers/session-resolver';
 import { reportGraphQLError } from './sentry';
 
 type ServerConfig = {
@@ -39,7 +40,7 @@ export function formatError(debug: boolean, logger: Logger, err: GraphQLError) {
  * Context available to resolvers
  */
 export type Context = {
-  authUser: string;
+  session: { uid: string; state: 'verified' | 'unverified' };
   token: string;
   dataSources: DataSources;
   logger: Logger;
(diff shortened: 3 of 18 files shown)


packages/fxa-graphql-api/src/lib/auth.ts
```diff
@@ -13,10 +13,9 @@ export class SessionTokenAuth {
   @Inject(fxAccountClientToken)
   private authClient!: AuthClient;
 
-  public async lookupUserId(sessionToken: string): Promise<string> {
+  public async lookupUserId(sessionToken: string) {
     try {
-      const result = await this.authClient.sessionStatus(sessionToken);
-      return result.uid;
```
dannycoates, line 19: instead of keeping just the `uid` by keeping the `state` we can use it later. much of the downstream changes are just to handle the object instead of string

packages/fxa-graphql-api/src/lib/resolvers/session-resolver.ts
```diff
@@ -0,0 +1,25 @@
...
+  Resolver,
+} from 'type-graphql';
+import { Context } from '../server';
+import { Session as SessionType } from './types/session'
+
+@Resolver((of) => SessionType)
+export class SessionResolver {
+
+  @Query((returns) => SessionType)
+  session(@Ctx() context: Context, @Info() info: GraphQLResolveInfo) {
```
dannycoates, line 19: add `session` as a top-level query type. it's only field is `verified`

packages/fxa-settings/src/components/AccountDataHOC/index.tsx
```diff
@@ -15,9 +15,6 @@ type AccountDataHOCProps = {
   children: (props: AccountDataHOCChildrenProps) => React.ReactNode;
 };
 
-// A data extraction layer for initial load.
-// TODO: If an account is unverified we'll need to
-// requery with less requested properties.
```
dannycoates, line 20: as of #6137 this is no longer a problem

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -22,15 +23,17 @@ export const App = ({ queryParams }: AppProps) => {
...
+            <AppLayout
+              avatarUrl={account.avatarUrl}
+              primaryEmail={
+                account.emails.find((email) => email.isPrimary)!.email
+              }
+              hasSubscription={Boolean(account.subscriptions.length)}
+            >
+              <Settings {...{ account }} />
+            </AppLayout>
+          </VerifiedSessionGuard>
```
dannycoates, line 36: Just an example

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const GET_SESSION = gql`
+  query GetSession {
+    session {
+      verified
+    }
+  }
+`;
+
+export const VerifiedSessionGuard = ({
```
dannycoates, line 18: This component needs more work but the basic idea is to wrap things that require a verified session to view or act on with a "guard" for what to show if they're not verified

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  guard,
+  children,
+}: {
+  guard: React.ReactNode;
+  children: React.ReactNode;
+}) => {
+  const { loading, error, data } = useQuery(GET_SESSION, {});
```
dannycoates, line 25: this should just be `readQuery`

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (error) {
+    // not sure if this is necessary
+    throw error;
+  }
+
+  if (!data.session.verified) {
+    return <>{guard}</>;
+  } else {
+    return <>{children}</>;
+  }
```
jodyheavener, line 34: Could simplify this a bit with

`return data.session.verified ? children : guard`

But I'm Eh-Okay with this.

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  guard,
+  children,
+}: {
+  guard: React.ReactNode;
+  children: React.ReactNode;
+}) => {
+  const { error, data } = useQuery(GET_SESSION, { fetchPolicy: 'cache-only' });
+
+  if (error) {
+    // not sure if this is necessary
```
jodyheavener, line 26: I assume throwing here would yield the AppErrorBoundary, which is fine with me. If for some reason it doesn't I might log the exception to Sentry and return the guard so we're not disrupting the UX.

packages/fxa-graphql-api/src/lib/server.ts
```diff
@@ -39,7 +40,7 @@ export function formatError(debug: boolean, logger: Logger, err: GraphQLError) {
  * Context available to resolvers
  */
 export type Context = {
-  authUser: string;
+  session: { uid: string; state: 'verified' | 'unverified' };
```
jodyheavener, line 43: could enum `'verified'`, `'unverified'` since we use them in a number of places? might be overkill though.

dannycoates: if we add more states maybe

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
dannycoates: that doesn't work with `ReactNode` but does with `ReactElement` which ought to be ok for our use

packages/fxa-graphql-api/src/lib/auth.ts
```diff
@@ -13,10 +13,9 @@ export class SessionTokenAuth {
   @Inject(fxAccountClientToken)
   private authClient!: AuthClient;
 
-  public async lookupUserId(sessionToken: string): Promise<string> {
+  public async lookupUserId(sessionToken: string) {
```
6a68, line 16: It is a little funky to keep the name `lookupUserId` since the return value is no longer a `uid`. Why not just `getSessionStatus`?

packages/fxa-settings/src/lib/auth.ts
```diff
@@ -15,3 +15,6 @@ export function useAuth() {
   }
   return auth!;
 }
+
+// NOTE for future self
+// In cases where the sessionToken changes we should also flush the session { verified } value from apollo cache
```
6a68, line 20: should this be a followup bug?

packages/fxa-graphql-api/src/lib/auth.ts
dannycoates: true, i totally glossed over the fn name

packages/fxa-settings/src/lib/auth.ts
dannycoates: just something to remember when we implement change password

PR #6145, opened 2020-08-05
refactor settings state

dannycoates:
Because our assumptions about the data from gql have changed, we no longer need to worry about parts of the Account type needing a verified session, we can simplify the structure a bit.

In the future we can use `useAccount` in components to get the data they need instead of pushing down the `account` object.

This also introduces the `models` folder for organizing our types.

For testing components that `useAccount()` there's the `MockedCache` component.


packages/fxa-settings/src/components/AppLayout/index.tsx
```diff
@@ -8,22 +8,18 @@ import ContentSkip from '../ContentSkip';
...
-export const AppLayout = ({
-  avatarUrl,
-  primaryEmail,
-  hasSubscription,
-  children,
-}: AppLayoutProps) => {
+export const AppLayout = ({ children }: AppLayoutProps) => {
   const alertBarRootRef = useRef<HTMLDivElement>(null);
-
+  const account = useAccount();
```
dannycoates, line 19: this would replace pushing the `account` or it's properties down

packages/fxa-settings/src/lib/types.ts
```diff
@@ -9,3 +9,5 @@ export interface QueryParams {
   flow_id?: string;
   flow_begin_time?: number;
 }
+
+export type hexstring = string;
```
dannycoates, line 13: rm me

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -3,36 +3,78 @@
...
+  if (loading) {
+    return (
+      <LoadingSpinner className="bg-grey-20 flex items-center flex-col justify-center h-screen select-none" />
+    );
+  }
+
+  if (error) {
+    return <AppErrorDialog data-testid="error-dialog" {...{ error }} />;
+  }
+  const account = data!.account;
```
dannycoates, line 71: we won't even care about `data` in this component once `Settings` uses `useAccount`

packages/fxa-settings/src/models/Account.ts
```diff
@@ -1,4 +1,33 @@
...
+  }[];
+  totp: {
+    exists: boolean;
+    verified: boolean;
+  };
+  subscriptions: {
+    created: number;
+    productName: string;
+  }[];
+}
```
jodyheavener, line 30: This seems reasonable.

packages/fxa-settings/src/models/Account.ts
```diff
@@ -33,31 +62,9 @@ export const GET_ACCOUNT = gql`
...
-  };
-  subscriptions: {
-    created: number;
-    productName: string;
-  }[];
+export function useAccount() {
+  const { data } = useQuery<{ account: Account }>(GET_ACCOUNT, {
+    fetchPolicy: 'cache-only',
+  });
+  return data!.account;
```
jodyheavener, line 69: If I'm understanding the docs correctly, `loading` would never be used because `cache-only`, but it looks like this might still need to handle (or maybe just return) `error`. Something like:

```typescript
const { data, error } = useQuery<{ account: Account }>(GET_ACCOUNT, {
  fetchPolicy: 'cache-only',
});

return { account: data!.account, error };
```

jodyheavener: Err, unless `error` is only ever possible when performing network requests. I'm unsure if an error can occur after the initial fetch.

jodyheavener:
Yeah I'm happy with this approach, and it solves the importing type defs thing.

packages/fxa-settings/src/models/Account.ts
dannycoates: Yeah, I want to look at the source to figure out if `error` can occur with cache lookups

packages/fxa-settings/src/components/Nav/index.stories.tsx
```diff
@@ -4,12 +4,22 @@
...
+import { MockedProvider } from '@apollo/client/testing';
+import { createCache } from '../../models/_mocks';
 import { Nav } from '.';
 
 storiesOf('Components|Nav', module)
   .add('basic', () => (
-    <Nav hasSubscription={false} primaryEmail="user@xample.com" />
+    <MockedProvider cache={createCache()}>
+      <Nav />
+    </MockedProvider>
```
dannycoates, line 15: one downside about `useAccount` over props is the need to have a `MockedProvider` in tests and storybook. i wonder if/how we could reduce the boilerplate.

LZoog: At first glance I liked the idea of extending `MockedProvider` and passing in props to override the defaults we set, but I kind of feel like `cache={createCache({ // ... overrides` is more explicit and clear on what's going on. Not a great answer, I know... I'll put more thought into this. 🤔

packages/fxa-settings/src/components/HeaderLockup/index.test.tsx
```diff
@@ -5,14 +5,20 @@
...
 import HeaderLockup from '.';
 
 // TO DO: functional test for `data-testid="header-menu"` to be visible in
 // mobile & tablet but hidden at desktop
 
 describe('HeaderLockup', () => {
   it('renders as expected', () => {
-    render(<HeaderLockup avatarUrl={null} primaryEmail="user@example.com" />);
+    render(
+      <MockedProvider cache={createCache({ account: { avatarUrl: '' } })}>
```
LZoog, line 18: If the `avatarUrl` isn't set, we get back `null` and not an empty string

dannycoates: oh yeah, that's me mid debugging something else ;)

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -10,71 +10,62 @@ import Avatar from '../Avatar';
...
-}: HeaderLockupProps) => {
-  const right = (
-    <HeaderContentRight
-      {...{
-        avatarUrl,
-        primaryEmail,
-      }}
-    />
+export const HeaderLockup = () => {
+  const { avatarUrl, primaryEmail } = useAccount();
```
LZoog, line 16: I wonder if this should be called something like `getAccountCache` instead? `use` makes me think it's a hook.

dannycoates: it is a hook

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -3,36 +3,78 @@
...
+
   return (
     <AppErrorBoundary>
```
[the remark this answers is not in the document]
dannycoates: I moved the boundary up to `index.tsx` in order to wrap this

packages/fxa-settings/README.md
```diff
@@ -218,12 +218,25 @@ yarn test -t="renders as expected"
...
 
 You can also see the test coverage with details by running the following command:
+
 ```
 CI=yes yarn test --coverage --verbose
 ```
 
 Refer to Jest's [CLI documentation](https://jestjs.io/docs/en/cli) for more advanced test configuration.
 
+### Components that use `useAccount`
```
LZoog, line 228: Thanks for writing this, I like that it'll come up if someone searches `fxa-settings` for `MockedProvider`. It could go above the "Testing" section I think but it doesn't really matter.

packages/fxa-settings/src/components/HeaderLockup/index.tsx
LZoog: I guess it returns the result of another hook 😛

packages/fxa-settings/src/models/_mocks.tsx
```diff
@@ -0,0 +1,139 @@
...
+      session: {
+        verified,
+      },
+    },
+  });
+  return cache;
+}
+
+/**
+ * MockedCache is a sugary sweet version of MockedProvider.
```
LZoog, line 83: I'm fine with leaving this here but since it's so lengthy it could also go in the README with a link to the section in the docs here.

packages/fxa-settings/src/index.tsx
```diff
@@ -62,7 +63,9 @@ try {
     <React.StrictMode>
       <ApolloProvider client={apolloClient}>
         <AuthContext.Provider value={{ auth: authClient }}>
-          <App {...{ queryParams }} />
+          <AppErrorBoundary>
```
LZoog, line 66: I'm not entirely sure where this should live FWIW, in payments we wrap `StripeProvider` and `ReduxProvider` but not `AppContext.Provider` or `AppLocalizationProvider`...

dannycoates: we'll figure it out eventually ;)

packages/fxa-settings/src/models/Account.ts
dannycoates: From https://github.com/apollographql/apollo-client/blob/master/src/core/ObservableQuery.ts#L128 it'll return the same result as the original `GET_INITIAL_STATE` query. If there was an `error` then, there will be an `error` here, but since we already handle that in the `App` component it's safe to assume if we get this far there is no `error`.

jodyheavener:
I neglected to look at this again in its reviewable state, but just wanted to pop by to say hell yea.

PR #6148, opened 2020-08-05
fix(oidc): Allow prompt=none with id_token_hint for all RPs

### LINKED ISSUE #4963: Consider loosening `prompt=none` restrictions
We originally restricted `prompt=none` usage, because we only supported passing an email as the hint. In this version of the flow, a well-intentioned RP might confuse a user by silently logging them in.

Now that we've added support for the `id_token_hint`, in which the RP passes in an ID Token specific to that RP and created during the regular, user interaction required oauth flow, we could relax our restriction on which RPs can use `prompt=none`, provided they use an `id_token_hint`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1652)
┆Issue Number: FXA-1652

### LINKED ISSUE #6028: Help AMO with error=unauthorized_client problem
┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2341)
┆Issue Number: FXA-2341

6a68:
* If an RP uses prompt=none with an id_token_hint, don't check against the allowed client list. This is because the user must have signed in to the RP before, otherwise the RP could not provide us a valid id_token.

* We still require an RP to be on the allowlist if they want to use prompt=none with the user email (via either the login_hint or email parameters).

Closes #4963, closes #6028.

## Because

- AMO has run into an error using `id_token_hint` due to the `prompt=none` allowlist

- And because we have long planned (see #4963) to allow any RP with an ID Token to use `prompt=none` without being on the allowlist (see also discussion in the docs updated in this PR)

## This pull request

- modifies the `prompt=none` validation flow to only check against the RP allowlist if the RP uses the `login_hint` aka `email` query parameter, but not if the RP uses `id_token_hint`

- adds front-end tests for this new case

- updates the `prompt=none` docs accordingly 

Closes #4963, closes #6028.

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/docs/oauth/prompt-none.md
+++ b/packages/fxa-auth-server/docs/oauth/prompt-none.md
@@ -8,8 +8,13 @@ More formally, `prompt=none` is a flow described by the [OpenID Connect spec][#o
 
 ## prompt=none usage is controlled
 
-Usage of `prompt=none` is controlled to an authorized list of RPs since
+Usage of `prompt=none` with `login_hint` / `email` is controlled to an authorized list of RPs, since
 it bypasses the normal authorization flow and use could easily fall afoul of user expectations.
+
+However, _all_ RPs can check the user's FxA login state using `id_token_hint`: since RPs not on the
+[authorized list](#prompt=none-usage-is-controlled) can only obtain an [id_token][#oidc-id-token] by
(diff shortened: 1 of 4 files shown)


6a68:
Thanks for twiddling the CI knobs, @dannycoates! :beers:

Issue #4926, opened 2020-04-14
[global] Decide on global state management (Context/apollo client cache?) for success message functionality


data-sync-user:
See https://docs.google.com/spreadsheets/d/1WNRY_hvOrn4ypIv-WHY-DGnxNpje4IlgCaK11F6eus8/edit?disco=AAAAJYHxp8U

We need a way to manage state across our application. Can we use manage all local/remote data with and treat [Apollo Client cache](https://www.apollographql.com/docs/react/caching/cache-interaction/) as our single source of truth? Is the Context API sufficient for our needs across this project? If it is, this issue may be closed with a comment with an example of how we will convey a success up to the parent Settings component where success messages and users will be sent to read it*, or by setting up Apollo cache if it hasn't been already.

Otherwise, we need an explanation as to why these options won't work for us, and we will need to setup Redux and determine how it should play nicely with Apollo.

--

Edit: Context won't be sufficient here. At the time of writing, we have the AlertBar created but it's not setup to convey a message to a parent component. See `Settings/index.tsx` component which currently has this comment: While this is where the AlertBar needs to be in the DOM, it won't be composed here like this. We likely need some sort of alert bar root element and then AlertBar can return a React.Portal hooking into this element via a ref so that we can freely use <AlertBar> with content where needed while its placement in the DOM remains here.

Closing this issue 1) requires a little research to see if we can manage all of our state, including remote data from the GQL server and [needed REST API|FXA-2001] calls, plus our [local state](https://www.apollographql.com/docs/tutorial/local-state/) in the Apollo Link State/Apollo Client cache, and 2) setting up our state management if it's not already. Additionally, 3) this issue should set us up to use `<AlertBar>message</AlertBar>` in any component while displaying the bar in the DOM where noted in `Settings/index.tsx`.

More readings, this is deprecated but what was used prior to local state management being baked into Apollo: https://www.apollographql.com/docs/link/links/state/

This is now a "5" due to the needed research/experimenting and AlertBar ask.

If this grows into larger than a 3 pointer, feel free to break this up and file follow up on point #2 or #3.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1628)
┆Issue Number: FXA-1628

data-sync-user:
➤ Danny Coates commented:

My initial impression is that yes, Apollo Client can handle all of our state management needs. https://www.apollographql.com/docs/react/local-state/local-state-management/

Issue #4903, opened 2020-04-13
[main] Build Settings > Profile > Secondary email: real data + resend verification email


data-sync-user:
See the "Secondary Email states" section of [visual designs](https://www.figma.com/file/cuREPfNFQllVq4CXdwtLUz/FxA---Manage-Settings?node-id=0%3A1), found in the middle.

This section is currently hard-coded to display the “no secondary email” state. Using the data from accountData:

* Display the secondary email if the user has one set. Note users may have multiple secondary emails added.
* If the secondary email is not verified, display the "Verification needed. Resend verification email if it's not in your email or spam" text. "Resend verification email" should be a link.
* Clicking on "Resend verification email" should resend the verification email - there is a GQL mutation available called `resendSecondaryEmailCode` that should be hit. On success/fail, display the message to the user in the `AlertBar`.

Currently, part of FXA-1612 is fixed because of the modal displayed in `UnitRowSecondaryEmail`. Try to preserve this if possible (commenting it out is fine with a note about FXA-1612).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1610)
┆Issue Number: FXA-1610

PR #6149, opened 2020-08-05
feat(settings): Add AlertBar root el with Context, UnitRowSecondaryEmail

LZoog:
## Because

* We need to use the AlertBar component throughout the project while keeping it as the first DOM element after the main tag.

## This pull request

* Adds 'alert-bar-root' element and uses createPortal in AlertBar to mount
* Adds AlertBarContext for referencing the 'alert-bar-root' element
* Modifies the hook that sets focus on the element opening a modal, allowing for a 'triggerException' to allow AlertBar to take focus
* Pulls the secondary email section into its own component

Closes: #4926 

## Checklist

## Other information

This issue had a couple of different parts to it, but part of it was to look into Apollo client cache to ensure that would handle our state needs across our GQL and REST API calls plus anything that would require a global state. Through reading I've done plus Danny's work on #6125, we've determined Apollo client cache is what we want, as we can store any local data there and use it as our single source of truth. We have this initialized already.

The other part of that ticket was allowing us to use the `<AlertBar>` component wherever while rendering it just below `<main>` which is what this PR covers, since it was likely we'd need to use Context or something global to pass around.

I was going to write some docs for this, but I'm afraid the implementation may change later and leave this out of date. I'd gotten this far, I could add this to the README if it seems valuable, but stopped once I got to the example.

```
### React Portals and Using the AlertBar

The `Modal` component uses a [React Portal](https://reactjs.org/docs/portals.html) to render the component in a dynamically created root element adjacent to the `<div id="root"></div>` element, used to mount the React application on load. Similarly, the `AlertBar` uses a portal to render the component inside an `alert-bar-root` element which is the first element inside the `<main>` tag.

[was going to insert AlertBar example use here but you could just reference UnitRowSecondaryEmail ¯\_(ツ)_/¯ ]
```

### DIFF
--- a/packages/fxa-react/components/Portal/index.tsx
+++ b/packages/fxa-react/components/Portal/index.tsx
@@ -39,8 +39,11 @@ const Portal = ({
     el = document.createElement('div');
     el.setAttribute('class', 'portal');
     el.setAttribute('id', id);
-    el.setAttribute('role', 'dialog');
     document.body.appendChild(el);
+
+    if (id !== 'alert-bar-portal') {
+      el.setAttribute('role', 'dialog');
+    }
 
     if (id === 'modal') {
       el.setAttribute('aria-labelledby', headerId);
--- a/packages/fxa-settings/src/components/AlertBar/index.tsx
+++ b/packages/fxa-settings/src/components/AlertBar/index.tsx
@@ -2,47 +2,70 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useContext } from 'react';
+import { createPortal } from 'react-dom';
+import AlertBarContext from '../../lib/AlertBarContext';
 import { useEscKeydownEffect, useChangeFocusEffect } from '../../lib/hooks';
 import { ReactComponent as CloseIcon } from 'fxa-react/images/close.svg';
+import Portal from 'fxa-react/components/Portal';
 
 type AlertBarProps = {
-  children: ReactNode | string;
+  children: ReactNode;
   onDismiss: Function;
 };
 
 export const AlertBar = ({ children, onDismiss }: AlertBarProps) => {
+  const { alertBarRootRef } = useContext(AlertBarContext);
+
+  // Although `role="alert" is usually sufficient to trigger a screenreader
+  // without having to reset focus, if this component is rerendered before
+  // it's removed from the DOM, the message won't be read. Setting focus
+  // ensures screenreaders receive feedback and makes the "close" button
+  // the next tabbable element after reveal.
   const tabFenceRef = useChangeFocusEffect();
   useEscKeydownEffect(onDismiss);
 
-  return (
-    <div
-      className="flex fixed justify-center mt-2 mx-2 right-0 left-0"
-      role="alert"
-      data-testid="alert-bar"
-    >
-      <div className="max-w-2xl w-full desktop:min-w-sm flex shadow-md bg-green-500 rounded font-bold text-sm">
-        <div
-          tabIndex={0}
-          ref={tabFenceRef}
-          data-testid="alert-bar-tab-fence"
-          className="outline-none"
-        />
-        <div className="flex-1 py-2 px-8 text-center">{children}</div>
-
-        <div className="flex pr-1">
-          <button
-            data-testid="alert-bar-dismiss"
-            className="self-center"
-            onClick={onDismiss as () => void}
-            title="Close message"
-          >
-            <CloseIcon className="w-3 h-3 m-2" role="img" />
-          </button>
-        </div>
+  const alertBar = <div
+    className="flex fixed justify-center mt-2 mx-2 right-0 left-0"
+    role="alert"
+    data-testid="alert-bar"
+  >
+    <div className="max-w-2xl w-full desktop:min-w-sm flex shadow-md bg-green-500 rounded font-bold text-sm">
+      <div
+        tabIndex={0}
+        ref={tabFenceRef}
+        data-testid="alert-bar-tab-fence"
+        className="outline-none"
+      />
+      <div className="flex-1 py-2 px-8 text-center">
+        {children}
+      </div>
+      <div className="flex pr-1">
+        <button
+          data-testid="alert-bar-dismiss"
+          className="self-center"
+          onClick={onDismiss as () => void}
+          title="Close message"
+        >
+          <CloseIcon className="w-3 h-3 m-2" role="img" />
+        </button>
       </div>
     </div>
-  );
+  </div>;
+
+  if (alertBarRootRef.current) {
+    return createPortal(alertBar, alertBarRootRef.current);
+  } else {
+    // this should never happen, but log an error and use Portal as a backup
+    console.error('Attempted to mount AlertBar on element ID "alert-bar-root" but it was not found in the document.');
+    return(
+      <Portal id="alert-bar-portal">
+        <div className="fixed mt-16 top-0" data-testid="alert-bar-portal">
+          {alertBar}
+        </div>
+      </Portal>
+    );
+  }
 };
 
 export default AlertBar;
--- a/packages/fxa-settings/src/components/AppLayout/index.tsx
+++ b/packages/fxa-settings/src/components/AppLayout/index.tsx
@@ -2,11 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React from 'react';
+import React, { useRef } from 'react';
 import HeaderLockup from '../HeaderLockup';
 import ContentSkip from '../ContentSkip';
 import Nav from '../Nav';
 import Footer from 'fxa-react/components/Footer';
+import AlertBarContext from '../../lib/AlertBarContext';
 
 type AppLayoutProps = {
   avatarUrl: string | null;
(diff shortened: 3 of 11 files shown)


packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { useBooleanState } from 'fxa-react/lib/hooks';
+import UnitRow from '../UnitRow';
+import Modal from '../Modal';
+import AlertBar from '../AlertBar';
+
+type UnitRowSecondaryEmailProps = {
+  primaryEmail: string;
+};
+
+export const UnitRowSecondaryEmail = ({ primaryEmail }: UnitRowSecondaryEmailProps) => {
```
LZoog, line 15: This component doesn't have any tests yet because it doesn't do anything we aren't testing in other files. When it actually does stuff we'll add tests.

packages/fxa-react/components/Portal/index.tsx
```diff
@@ -39,9 +39,12 @@ const Portal = ({
     el = document.createElement('div');
     el.setAttribute('class', 'portal');
     el.setAttribute('id', id);
-    el.setAttribute('role', 'dialog');
     document.body.appendChild(el);
 
+    if (id !== 'alert-bar-portal') {
+      el.setAttribute('role', 'dialog');
+    }
```
LZoog, line 46: I know this file is getting a little gross. When payments uses Tailwind we can better combine `DialogMessage` and `Modal` to make this a bit cleaner...

packages/fxa-settings/src/components/AlertBar/index.tsx
```diff
@@ -3,46 +3,68 @@
...
+import Portal from 'fxa-react/components/Portal';
 
 type AlertBarProps = {
-  children: ReactNode | string;
+  children: ReactNode;
   onDismiss: Function;
 };
 
 export const AlertBar = ({ children, onDismiss }: AlertBarProps) => {
+  const alertBarRoot = document.getElementById("alert-bar-root");
```
LZoog, line 17: Eh so, I converted this back to a draft while I check on something here - while `document.getElementById` isn't ideal in React it seemed preferable to creating an element ref in `AppLayout` and passing it into every `AlertBar` instance and ref forwarding wasn't doing what I wanted, but going to look into one more idea...

packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -5,10 +5,12 @@
...
 
 // Focus on the element that triggered some action after the first
-// argument changes from `false` to `true`.
+// argument changes from `false` to `true` unless a `triggerException`
+// is also provided.
 export function useFocusOnTriggeringElementOnClose(
   revealed: boolean | undefined,
-  triggerElement: React.RefObject<HTMLButtonElement>
+  triggerElement: React.RefObject<HTMLButtonElement>,
+  triggerException?: boolean | undefined
```
jodyheavener, line 13: I think the presence of `?` infers `| undefined`, no?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [alertBarRevealed, revealAlertBar, hideAlertBar] = useBooleanState();
+  const [modalRevealed, revealModal, hideModal] = useBooleanState();
+
+  const modalHeaderId = 'modal-header-verify-email';
+  const modalDescId = 'modal-desc-verify-email';
+
+  const onSecondaryEmailConfirm = useCallback(() => {
+    console.log('confirmed - resend verification code');
```
jodyheavener, line 23: from debugging?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { UnitRowSecondaryEmail } from '.';
+
+storiesOf('Components|UnitRowSecondaryEmail', module)
+  .add('basic', () => (
+    <>
+      <div id="alert-bar-root" />
```
jodyheavener, line 12: this needs a closing `</div>` :)

packages/fxa-settings/src/components/AppLayout/index.tsx
```diff
@@ -20,25 +21,31 @@ export const AppLayout = ({
...
+          avatarUrl,
+          primaryEmail,
+        }}
+      />
+      <div className="max-w-screen-desktopXl w-full mx-auto flex flex-1 tablet:px-20 desktop:px-12">
+        <div className="hidden desktop:block desktop:flex-2">
+          <Nav {...{ hasSubscription, primaryEmail }} />
+        </div>
+        <main id="main" data-testid="main" className="desktop:flex-7">
+          <div id="alert-bar-root" data-testid="alert-bar-root" ref={alertBarRootRef} />
```
jodyheavener, line 41: Needs a closing `</div>` 😜

packages/fxa-settings/src/components/AlertBar/index.tsx
```diff
@@ -2,47 +2,70 @@
...
-  );
+  </div>;
+
+  if (alertBarRootRef.current) {
+    return createPortal(alertBar, alertBarRootRef.current);
+  } else {
+    // this should never happen, but log an error and use Portal as a backup
+    console.error('Attempted to mount AlertBar on element ID "alert-bar-root" but it was not found in the document.');
+    return(
+      <Portal id="alert-bar-portal">
```
jodyheavener, line 62: Will `AlertBar` ever be rendered more than once in the app? If so we will want to avoid using IDs, or find a way for Portal to internally increment this value, since there can only be one instance of a unique ID on a given page.

packages/fxa-settings/src/components/AppLayout/index.tsx
jodyheavener: Per our Slack chat, JSX is cool with this syntax. Kinda feels gross to be writing invalid HTML to me, but it's fully supported and I'm just a crank, so you can ignore this and the last comment!

packages/fxa-settings/src/components/AlertBar/index.tsx
LZoog: Hm, the AlertBar shouldn't ever _actually_ use Portal so we should be fine here, but it [looks like we test for this in Portal tests](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/components/Portal/index.test.tsx#L27) to ensure it _can_ happen. 😅 I've been meaning to make a ticket for cleaning up Portal/combine DialogMessage/Modal after payments gets Tailwind, I can add this to that ticket.

packages/fxa-settings/src/components/AppLayout/index.tsx
LZoog: Lol but JSX is not HTML, `className` isn't a thing ;)

jodyheavener: _Throws gloves on the ground_

**Well actually, I—**

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Yeah, we've had this in there. This will actually be done/taken out in #4903.

packages/fxa-react/components/Portal/index.tsx
jodyheavener: Possibly for a future iteration, but could we just use add a `role` prop to the component and set it directly instead of inferring it?

LZoog: Probably, but yeah it makes sense to address in another PR since this is used in payments, I can add that to the ticket I'll make for cleaning this file up.

packages/fxa-settings/src/components/AlertBar/index.tsx
```diff
@@ -2,47 +2,70 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useContext } from 'react';
+import { createPortal } from 'react-dom';
+import AlertBarContext from '../../lib/AlertBarContext';
```
jodyheavener, line 7: Nitpick - sometimes the module dependency is used (`fxa-settings/src/lib/AlertBarContext`), sometimes the relative path is used. I'm fine with either, but ideally we could be consistent.

packages/fxa-settings/src/components/AlertBar/index.tsx
```diff
@@ -2,47 +2,70 @@
...
+        className="outline-none"
+      />
+      <div className="flex-1 py-2 px-8 text-center">
+        {children}
+      </div>
+      <div className="flex pr-1">
+        <button
+          data-testid="alert-bar-dismiss"
+          className="self-center"
+          onClick={onDismiss as () => void}
```
jodyheavener, line 47: Instead of casting this here could we just define the function signature in the props type (instead of `Function`)? Or does that create type mismatches elsewhere?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+              primary email. To do this, you'll need access to{' '}
+              {primaryEmail}
+            </p>
+          </Modal>
+        )}
+
+        {alertBarRevealed && (
+          <AlertBar onDismiss={hideAlertBar}>
+            <p>Check the inbox for {primaryEmail} to verify your primary
+          email.</p>
```
jodyheavener, line 74: Did Prettier push this to a new line? Feel like it usually does that `{' '}` thing

packages/fxa-settings/src/components/AlertBar/index.tsx
LZoog: Ah good catch, this was probably from some auto import update thing I didn't notice since I'd originally named this something else.

packages/fxa-settings/src/components/AlertBar/index.tsx
LZoog: Hmm, I get `Type 'Function' provides no match for the signature '(): void'`. We're doing this in `Modal` and `DialogMessage` as well, might look at this in a follow up.

PR #6150, opened 2020-08-06
refactor(settings): move type defs to a declaration file so they're always available

jodyheavener:
## Because

- We currently export types from `src/lib/types.ts`, which is perfectly fine but requires an import any time you want to reference one

## This pull request

- Moves definitions to the already-existing `react-app-env.d.ts` file so they're globally available. Note that this should not preclude you from defining single-use types in the same file as needed, and any global `enum`s will need to be in a standalone `.ts` file so they can be bundled with code.

## Checklist

## Other information

We can't remove or rename `react-app-env.d.ts` (it gets regenerated if so), which is why I chose to use it since it's there anyway. I could be talked into creating a separate declaration file, or we could use more granular ones (such as ones specific to account or query data).


packages/fxa-settings/src/react-app-env.d.ts
```diff
@@ -3,3 +3,40 @@
...
+
+type Hash<T> = { [key: string]: T };
+
+interface QueryParams {
+  device_id?: string;
+  flow_id?: string;
+  flow_begin_time?: number;
+}
+
+interface AccountData {
```
dannycoates, line 15: idk how i feel about this one being here, but we can always move it later if we need to

jodyheavener: yeah I tend to agree. What if I moved this to its own accounts.d.ts in the AccountHOC component?

dannycoates: i've got an idea... lemme sketch it up

jodyheavener:
Closing in favor of an approach laid out in #6145. Thanks @dannycoates!

PR #6155, opened 2020-08-06
fix(tests): add job to run tests against remote target

jrgm:
## Because

- having a functional test of stage run before further testiing of stage is useful

## This pull request

- adds a job that can be triggered by the api to run tests in stage. (For now; later we can work out a way to automagically trigger test runs)

## Checklist

This will run a smaller suite of tests. In the future I may reintroduce parallelism for that suite.

### DIFF
--- a/packages/fxa-content-server/scripts/test-ci-remote.sh
+++ b/packages/fxa-content-server/scripts/test-ci-remote.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -ex
+
+DIR=$(dirname "$0")
+cd "$DIR/.."
+echo -e "\n###################################"
+echo "# testing fxa-content-server"
+echo -e "###################################\n"
+
+env | sort
+
+function test_suite() {
+  local suite=$1
+  node tests/intern.js \
+    --suites="${suite}" \
+    --fxaAuthRoot=https://api-accounts.stage.mozaws.net/v1 \
+    --fxaContentRoot=https://accounts.stage.mozaws.net/ \
+    --fxaOAuthApp=https://123done-stage.dev.lcip.org/ \
+    --fxaUntrustedOauthApp=https://321done-stage.dev.lcip.org/ \
+    --fxaEmailRoot=http://restmail.net \
+    --fxaProduction=true \
+    --output="../../artifacts/tests/${suite}-${numGroups}-${i}-results.xml" \
+    --firefoxBinary=./firefox/firefox \
+    || \
+  node tests/intern.js \
+    --suites="${suite}" \
+    --fxaAuthRoot=https://api-accounts.stage.mozaws.net/v1 \
+    --fxaContentRoot=https://accounts.stage.mozaws.net/ \
+    --fxaOAuthApp=https://123done-stage.dev.lcip.org/ \
+    --fxaUntrustedOauthApp=https://321done-stage.dev.lcip.org/ \
+    --fxaEmailRoot=http://restmail.net \
+    --fxaProduction=true \
+    --output="../../artifacts/tests/${suite}-${numGroups}-${i}-results.xml" \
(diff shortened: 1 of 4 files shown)


clouserw:
Thanks @jrgm.  I didn't realize you were working on this, but I'm excited for this

  /cc @ashrivastava-qa @rbillings

jrgm:
So @ashrivastava-qa and I were working on this together. In squashing this all into one commit, it reads as only one of us, which it isn't.

packages/fxa-content-server/scripts/test-ci-remote.sh
```diff
@@ -0,0 +1,48 @@
...
+echo -e "\n###################################"
+echo "# testing fxa-content-server"
+echo -e "###################################\n"
+
+function test_suite() {
+  local suite=$1
+  node tests/intern.js \
+    --suites="${suite}" \
+    --fxaAuthRoot=https://api-accounts.stage.mozaws.net/v1 \
+    --fxaContentRoot=https://accounts.stage.mozaws.net/ \
```
jrgm, line 14: These URL should be parameterized somewhat similar to the old scripts so that this same test script can be reused against other endpoints. But get stage testing working first.

packages/fxa-content-server/scripts/test-ci-remote.sh
```diff
@@ -0,0 +1,48 @@
...
+mkdir -p artifacts/tests
+
+cd packages/fxa-content-server
+mozinstall /firefox.tar.bz2
+test_suite functional_smoke
+# TODO - restore these
+#test_suite server
+#mozinstall /7f10c7614e9fa46-target.tar.bz2
+#test_suite pairing
+
```
jrgm, line 47: and yeah, also TODO

packages/fxa-content-server/tests/functional_smoke.js
```diff
@@ -0,0 +1,30 @@
...
+  'tests/functional/connect_another_device.js',
+  'tests/functional/delete_account.js',
+  //  'tests/functional/fx_fennec_v1_sign_in.js',
+  //  'tests/functional/fx_fennec_v1_sign_up.js',
+  //  'tests/functional/fx_firstrun_v2.js',
+  //  'tests/functional/fx_ios_v1_sign_in.js',
+  //  'tests/functional/fx_ios_v1_sign_up.js',
+  //  'tests/functional/oauth_reset_password.js',
+  //  'tests/functional/oauth_sign_in_token_code.js',
+  'tests/functional/oauth_sign_up.js',
```
jrgm, line 17: When this PR merges, we can add back more of these tests and work towards green on all of them.

packages/fxa-content-server/scripts/test-ci-remote.sh
```diff
@@ -0,0 +1,48 @@
...
+    --fxaEmailRoot=http://restmail.net \
+    --fxaProduction=true \
+    --output="../../artifacts/tests/${suite}-${numGroups}-${i}-results.xml" \
+    --firefoxBinary=./firefox/firefox \
+    --grep="$(<rerun.txt)"
+}
+
+yarn lint
+
+cd ../../
```
jrgm, line 37: Also TODO. Curl `__version__` to get the correct version of the tests to be run (i.e., running the `main` branch for tests when stage is on v1.nnn.m will result in false negatives)

jrgm:
> LGTM. Does this mean we're moving the smoke tests out of TeamCity, or is that an unrelated change?

Yeah, that can happen separately.

clouserw:
We're triaging PRs but we're not sure if/when this should land

jrgm:
> We're triaging PRs but we're not sure if/when this should land

Anytime. I'd merge it, but y'know...

PR #6169, opened 2020-08-09
fix(content-server): validate the sign up input before showing validation error

### LINKED ISSUE #3575: Improve edge cases for inline password validation

byrw:
## Because

* The "passwords do not match" tooltip is still displayed, even if there is a validation error.

## This pull request

* Check if the input is valid before showing the tooltip.

Closes #3575

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/form.js
+++ b/packages/fxa-content-server/app/scripts/views/form.js
@@ -573,6 +573,27 @@ var FormView = BaseView.extend({
     }
     return newValues;
   },
+
+  /**
+   * Checks whether the selected form is valid.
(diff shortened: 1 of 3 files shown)


jodyheavener (approved):
Nice, appears to do what it's meant to for me. Thanks for this @byrw!

PR #6174, opened 2020-08-10
refactor(settings + react): update tests to be more consistent and DRY

jodyheavener:
## Because

I went down a rabbit hole of making our new React app testing environments a little more consistent and DRY.

## This pull request

- Adds `setupTests.js` file in fxa-settings and fxa-react. Because fxa-settings is a create-react-app, react-scripts detects this file [automatically](https://create-react-app.dev/docs/running-tests/#initializing-test-environment) if it's present. fxa-react doesn't, so I updated the Jest config to do this. You should put any boilerplate code in these files.
- Stubs console log/error/warning. These just add noise to the test output. You can still assert on them if you want to ensure they were invoked (though, would you?)
- Removes all instances of `cleanup`. [Don't need it.](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup)
- Prefers usage of `screen` where the result of testing rendering was using `getBy*` multiple times per file. [Here's why.](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) I'm definitely open to feedback on this. 

## Checklist

### DIFF
--- a/packages/fxa-react/components/AppErrorBoundary/index.test.tsx
+++ b/packages/fxa-react/components/AppErrorBoundary/index.test.tsx
@@ -3,32 +3,20 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-import { render } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
+import { render, screen } from '@testing-library/react';
 import AppErrorBoundary from '.';
 
 describe('AppErrorBoundary', () => {
-  beforeEach(() => {
-    // HACK: Swallow the exception thrown by BadComponent
-    // it bubbles up unnecesarily to jest and makes noise
-    jest.spyOn(console, 'error');
-    (global.console.error as jest.Mock).mockImplementation(() => {});
-  });
-
-  afterEach(() => {
-    (global.console.error as jest.Mock).mockRestore();
-  });
-
   it('renders children that do not cause exceptions', () => {
     const GoodComponent = () => <p data-testid="good-component">Hi</p>;
 
-    const { queryByTestId } = render(
+    render(
       <AppErrorBoundary>
         <GoodComponent />
       </AppErrorBoundary>
     );
 
-    expect(queryByTestId('error-loading-app')).not.toBeInTheDocument();
+    expect(screen.queryByTestId('error-loading-app')).not.toBeInTheDocument();
   });
 
   it('renders a general error dialog on exception in child component', () => {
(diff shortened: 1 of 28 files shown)


packages/fxa-react/components/Portal/index.test.tsx
```diff
@@ -1,13 +1,11 @@
 import React from 'react';
-import { render, cleanup } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
-
+import { render, screen } from '@testing-library/react';
 import Portal from './index';
 
-afterEach(cleanup);
+const { getByTestId } = screen;
```
jodyheavener, line 5: Shoot @LZoog I just remembered we had a preference discussion about destructuring screen at the top. I'm not adamant on this format, happy to adjust to `screen.getBy*` an not destructure if you'd prefer!

packages/fxa-react/components/AppErrorBoundary/index.test.tsx
```diff
@@ -3,26 +3,16 @@
...
-    // HACK: Swallow the exception thrown by BadComponent
-    // it bubbles up unnecesarily to jest and makes noise
-    jest.spyOn(console, 'error');
-    (global.console.error as jest.Mock).mockImplementation(() => {});
-  });
-
-  afterEach(() => {
-    (global.console.error as jest.Mock).mockRestore();
-  });
+const { queryByTestId } = screen;
```
LZoog, line 9: If we need to `rerender`, could this change? I'm not sure this should be globally set here. Maybe just update `queryByTestId` references to `screen.queryByTestId`?

packages/fxa-react/components/Portal/index.test.tsx
LZoog: Ah oops I left a comment on the first instance of this in the PR without seeing this comment lol. Yeah I mean personally, I like using `screen`, but I don't like it being destructured at the top and it being truly global😅

jodyheavener:
Yeah FWIW I'd like to standardize using global `screen` over the resulting screen from a render. Seems like a good consistent thing to do.

> I don't like it being destructured at the top and it being truly global

Worth noting it's only global to the current file, not _all_ test files. It just saves you from having to type `screen.` every single time, and in a large test file this could dry things up/save time, and the likelihood of namespace conflict is pretty low.

> I just personally like being able to see the available methods when typing screen.

This is still [totally possible](https://user-images.githubusercontent.com/6392049/90019278-2ef3b380-dc7c-11ea-9e5f-b27ea6766841.png)

--- 

All this being said, it's all just preference. I'll update and we can discuss further if for some reason it comes up again. All I'm looking for is something we can standardize on 😸

PR #6176, opened 2020-08-10
feat(settings): redirect to login when session token is invalid

### LINKED ISSUE #6118: [global] Handle invalid sessionToken
Right now, if the user doesn't have a sessionToken in localStorage, we get the following error in the console:

Error initializing FXA Settings TypeError: can't access property "sessionToken", storedAccounts[currentAccountUid] is undefined

We'll need to redirect back to `get_flow` or login when this occurs; see [this TODO comment](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/index.tsx#L20) and reference payments for what we do there.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2379)
┆Issue Number: FXA-2379

dannycoates:
When the graphql server fails authentication it means we need to reauthenticate the user. We do this in a slightly roundabout way by redirecting to /get_flow and having it navigate to the login page so that we can redirect_to back to /beta/settings. Once new settings is no longer in beta we ought to be able to redirect straight to /signin.

fixes #6118

### DIFF
--- a/packages/fxa-settings/src/lib/gql.ts
+++ b/packages/fxa-settings/src/lib/gql.ts
@@ -1,0 +1,52 @@
+import { ApolloClient, createHttpLink, from } from '@apollo/client';
+import { setContext } from '@apollo/client/link/context';
+import { onError } from '@apollo/client/link/error';
+import { cache, sessionToken, typeDefs } from './cache';
+
+export function createApolloClient(gqlServerUri: string) {
+  // httpLink makes the actual requests to the server
+  const httpLink = createHttpLink({
+    uri: `${gqlServerUri}/graphql`,
+  });
+
+  // authLink sets the Authentication header on outgoing requests
+  const authLink = setContext((_, { headers }) => {
+    return {
+      headers: {
+        ...headers,
+        Authorization: sessionToken(),
+      },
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/lib/gql.ts
```diff
@@ -0,0 +1,52 @@
...
+  const errorLink = onError(({ graphQLErrors, networkError }) => {
+    let reauth = false;
+    if (graphQLErrors) {
+      for (const error of graphQLErrors) {
+        if (error.extensions?.code === 'UNAUTHENTICATED') {
+          reauth = true;
+        }
+      }
+    }
+    if (networkError && 'statusCode' in networkError) {
```
LZoog, line 32: Is `networkError && 'statusCode' in networkError` equivalent to `networkError?.statusCode`?

dannycoates: in this case unfortunately not. The type definition for `networkError` is a union of a bunch of different types so we need the `&&` and `in` as a type guard since there's not shorter syntax for "not undefined and this (more) specific type"

PR #6177, opened 2020-08-10
docs(adr): add application architecture

bbangert:
Because:

* FxA has no chosen application architecture, with several ad-hoc ones
  in place.

This commit:

* Proposes that an application architecture for our back-end services
  should be formally defined, chosen, and have documentation/tooling
  available.

## Checklist

### DIFF
--- a/docs/adr/0020-application-architecture.md
+++ b/docs/adr/0020-application-architecture.md
@@ -1,0 +1,123 @@
+# Use a documented Application Architecture for new back-end services
+
+- Deciders: Ben Bangert, Ryan Kelly, Vijay Budhram, Jody Heavener, Danny Coates, Dave Justice
+- Date: 2020-08-12
+
+## Context and Problem Statement
+
+The FxA back-end service stack contains an application architecture that is ad-hoc, not documented, and missing modern features (such as Dependency Injection) which results in the following problems:
+
+- New developers struggle to get up to speed as they must learn the architecture by reading the code as we have no documentation on the application structure, why they're structured the way they are, or how new components should be added to fit in. Each back-end service may vary in its ad-hoc architecture as well.
+- Adding new objects needed in a route handler can be time-consuming as the object must be plumbed through the entire initialization chain vs. more elegant methods like Dependency Injection (DI).
+- Not clear where/how to add new components and takes time to study/understand how things are currently setup in an attempt to mimic the structure for the new component.
+- Time consuming to setup boiler-plate for components, as we have no tooling to work with the current ad-hoc application architectures.
+- Our ad-hoc architecture frequently mixes concerns such as having business logic mixed in with request handling logic, and has other warts from its evolution over time vs. being planned up front.
+- New back-end services evolve differently resulting in more ad-hoc application architectures to learn.
+- Shared components in `fxa-shared` can't rely on basic object lifecycles or setup approaches as they may be used in multiple different ad-hoc application architectures.
+
+Not choosing an application framework means that we have choosen to make ad-hoc application architectures which will continue to exhibit the problems above.
+
+It is assumed that the four newest FxA back-end services (admin-server, support-panel, event-broker, gql-api) will be switched to the chosen approach for consistency.
+
+## Decision Drivers
+
+- Documented application architecture.
+- Tooling that reduces boiler-plate and creates consistent code architecture.
+- Modern paradigms to ease creation of global objects and their testability, such as DI.
+- Training materials so that new developers can understand and work on application features easily.
+- Ability to migrate legacy applications to similar conventions/setup.
+- Handles situations we need such as:
+  - Exception handling
+  - Validation
+  - Service instantiation (DI, etc)
+  - Authentication
+  - RESTful API's (Swagger/OpenAPI a bonus)
+  - GraphQL
+  - Use of components/services easily in a script
+
+## Considered Options
+
+- Evolve/improve our ad-hoc application architecture with docs, tools, etc.
+- Use an existing well-documented and popular application framework such as:
+  - NestJS
+  - Adonis
+  - LoopBack
+
+## Decision Outcome
+
+Chosen Option: Use an existing framework: NestJS
+
+NestJS and LoopBack are the two most compelling options, however NestJS has substantially better GraphQL support and a much larger user-base. LoopBack has its user-base divided between two versions with substantial changes between them. We will use NestJS for new projects and update the newest services (admin-server, support-panel, event-broker, gql-api) to NestJS.
+
+To reduce documentation needs and address that auth-server will not be reasonable to migrate, we will backport code organization schemes and directory/filename conventions from NestJS and use `typedi` as a DI system to ease configuration/testing in auth-server and possibly other legacy packages. This will allow developers familiar with NestJS projects to still locate and easily work on FxA legacy packages.
+
+### Positive Consequences
+
+- Less documentation to create, as we only document differences for legacy services.
+- NestJS GraphQL support matches existing typegraphql paradigm closely for minimal effort switch.
+
+### Negative Consequences
+
+- There will be some effort involved to migrate the 4 packages to NestJS.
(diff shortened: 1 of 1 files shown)


bbangert:
My personal preference is NestJS, as it has a robust set of well documented patterns that align well with what we're working on, has an Udemy course we can take (its in our business account), and is easy to migrate the 4 latest backend services to rather quickly.

docs/adr/0020-application-architecture.md
```diff
@@ -0,0 +1,110 @@
...
+### Evolve/improve our ad-hoc application architecture with docs, tools, etc.
+
+Our existing ad-hoc structure typically involves a few chosen libraries and a basic directory structure. Error handling differs in several of the back-end services, while validation is always provided by the `joi` library. To document our ad-hoc application architecture we would first need to make clear choices about what libraries we use and how their use should be structured in the application. We could then write tooling and documentation around these choices.
+
+- Pros:
+  - Could be easier to convert our existing backend services as they will likely be closer in libs/structure to the choosen structure if desired.
+  - Our choice of exactly what our application architectures look like and what libraries are used.
+- Cons:
+  - Creating even a portion of the documentation, tooling, and training material as already exists for any of the other three options is an incredible amount of effort/work.
+  - We will likely end up copying many of the features/layout/tools that already exist with communities for the other options.
```
rfk, line 62: I broadly endorse the list of cons here, for whatever that's worth; I think the process of trying to go through and clean it up to turn into a documented application architecture would look pretty similar to slowly evolving it towards some existing tool, so might as well head towards something with an existing ecosystem around it.

docs/adr/0020-application-architecture.md
```diff
@@ -0,0 +1,110 @@
...
+- Not clear where/how to add new components and takes time to study/understand how things are currently setup in an attempt to mimic the structure for the new component.
+- Time consuming to setup boiler-plate for components, as we have no tooling to work with the current ad-hoc application architectures.
+- Our ad-hoc architecture frequently mixes concerns such as having business logic mixed in with request handling logic, and has other warts from its evolution over time vs. being planned up front.
+- New back-end services evolve differently resulting in more ad-hoc application architectures to learn.
+- Shared components in `fxa-shared` can't rely on basic object lifecycles or setup approaches as they may be used in multiple different ad-hoc application architectures.
+
+Not choosing an application framework means that we have choosen to make ad-hoc application architectures which will continue to exhibit the problems above.
+
+It is assumed that the four newest FxA back-end services (admin-server, support-panel, event-broker, gql-api) will be switched to the chosen approach for consistency.
```
jodyheavener, line 20: I wonder how we could best approach transitioning our mature backend services. I don't think there's an overwhelming reason to transition them immediately, and in some cases it might make sense to not switch them at all, instead allowing them to exist until FxA architecture changes and they are retired. But waiting too long risks the new approach just adding to our list of varying tech stacks we have to learn and look after, especially if it's de-prioritized.

bbangert: I believe we can transition some of the existing services partly by moving the business logic into shared components that something like gql-api can use. This can at least reduce the amount of code in the legacy systems which should make it more manageable.

docs/adr/0020-application-architecture.md
```diff
@@ -0,0 +1,110 @@
...
+### Negative Consequences
+
+## Pros and Cons of the Options
+
+### Evolve/improve our ad-hoc application architecture with docs, tools, etc.
+
+Our existing ad-hoc structure typically involves a few chosen libraries and a basic directory structure. Error handling differs in several of the back-end services, while validation is always provided by the `joi` library. To document our ad-hoc application architecture we would first need to make clear choices about what libraries we use and how their use should be structured in the application. We could then write tooling and documentation around these choices.
+
+- Pros:
+  - Could be easier to convert our existing backend services as they will likely be closer in libs/structure to the choosen structure if desired.
```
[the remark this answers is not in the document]
bbangert: Agreed, that is definitely a notable pro. Though I think its feasible to back-port patterns/conventions from something like NestJS into our existing apps (like auth-server) so they can take advantage of the conventions even if they're not using the library.

An example:
In Nest, there's the convention that data transfer objects (DTO's) are in files, named as such along with the code they go to. While the routes are in a controller file. Business logic typically goes in a service. These filenames end like so:
- `*.dto.ts`
- `*.controller.ts`
- `*.service.ts`

So I could imagine copying the convention into auth-server for Stripe/subscription routing as being something like:
- Move URL handlers from `routes/subscriptions.ts` to `payments/payments.controller.ts`
- Consolidate business logic from `routes/subscriptions.ts` and `payments/stripe.ts` to `payments/stripe.service.ts`
- Move DTO's from `routes/validators.js` to `routes/dto/*.dto.ts`
- Move tests into `*.spec.ts` files that sit alongside what they test (as we use in front-end TS projects, and as Nest also uses)

Doing something like this we'd gain the advantage that even though some of the tooling doesn't work for our legacy projects, we can at least align them on other conventions that would make their style more inline with a new NestJS project.

Unfortunately we can't use any of these 3 full frameworks DI scheme without using the rest of the framework, so we might just have to use a standalone DI like `typedi` to benefit from that. 

This might be a good hybrid approach that lets us align the existing code-bases with NestJS naming/organization conventions while going completely to NestJS for the new services.

docs/adr/0020-application-architecture.md
bbangert: @jodyheavener see response I wrote to vijay, as it occurred to me how to backport some of the conventions into the mature services.

bbangert:
I have updated this to include a choice of NestJS in the absence of any alternatives or disagreement, as well as include a plan to apply conventions and code architecture organization schemes from NestJS to legacy packages for more consistency while minimizing needed documentation to only cover differences from NestJS.

I have moved a project of similar size to the admin-panel from a similar typegraphql setup to NestJS and it was a fairly minimal amount of work. I don't believe it will be more than a 3 pointer per each of the 4 latest back-end services to move to NestJS.

meandavejustice:
First of all @bbangert, thank you for putting this document together, it's been a long time coming. 

NestJS seems to be well maintained and I feel good/confident around it's popularity. I'm still a little taken aback by the fallout of HapiJS, right after we were all leaning towards having that as our chosen framework.

Keeping in mind that never worked with any of these frameworks directly (except poking at loopback a while ago), Nest seems to be the clear choice. I really like the idea of structuring auth-server with some of the same patterns, so long as that doesn't become an unwieldy task (it seems straight forward though).

I don't have anymore to add at the moment, I'm going to put some videos about each of these frameworks in the background over the next day or so and I'll speak up if anything jumps out at me as a risk 👍

docs/adr/0020-application-architecture.md
```diff
@@ -0,0 +1,123 @@
...
+- Adding new objects needed in a route handler can be time-consuming as the object must be plumbed through the entire initialization chain vs. more elegant methods like Dependency Injection (DI).
+- Not clear where/how to add new components and takes time to study/understand how things are currently setup in an attempt to mimic the structure for the new component.
+- Time consuming to setup boiler-plate for components, as we have no tooling to work with the current ad-hoc application architectures.
+- Our ad-hoc architecture frequently mixes concerns such as having business logic mixed in with request handling logic, and has other warts from its evolution over time vs. being planned up front.
+- New back-end services evolve differently resulting in more ad-hoc application architectures to learn.
+- Shared components in `fxa-shared` can't rely on basic object lifecycles or setup approaches as they may be used in multiple different ad-hoc application architectures.
```
jodyheavener, line 16: Side note, if there's an opportunity here to re-purpose `fxa-shared` so its not backend-specific, or rename it so it's less generic sounding, I'd like to do this.

docs/adr/0020-application-architecture.md
jodyheavener: Thanks for the followup! I would be happy to be involved in one of the small switches (even if just observing or reviewing PRs).

bbangert:
Updating deciders list and merging.

Issue #6154, opened 2020-08-06
[global] Read data from apollo client cache instead of pushing down component tree


data-sync-user:
At the moment we receive `accountData` at the highest level and drill it down into the child components.

We should be able to use `useQuery` with `{ fetchPolicy: 'cache-only' }` to read the data from the apollo client cache in the components where the data is needed instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2398)
┆Issue Number: FXA-2398

PR #6178, opened 2020-08-10
task(fxa-settings): convert settings component to use cache

meandavejustice:
## This pull request

- opts for `useAccount` over passing `account` object into components. builds off of @dannycoates pr, https://github.com/mozilla/fxa/pull/6145/files

Closes: #6154 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-settings/src/components/Settings/index.tsx
+++ b/packages/fxa-settings/src/components/Settings/index.tsx
@@ -7,10 +7,17 @@ import UnitRow from '../UnitRow';
 import UnitRowWithAvatar from '../UnitRowWithAvatar';
 import Security from '../Security';
 import UnitRowSecondaryEmail from '../UnitRowSecondaryEmail';
-import { Account } from '../../models';
 
-export const Settings = ({ account }: { account: Account }) => {
-  const primaryEmail = account.emails.find((email) => email.isPrimary)!;
+import { useAccount } from '../../models';
+
+export const Settings = () => {
+  const {
+    primaryEmail,
+    displayName,
+    avatarUrl,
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -7,26 +7,31 @@ import UnitRow from '../UnitRow';
...
-  const primaryEmail = account.emails.find((email) => email.isPrimary)!;
+import { useAccount } from '../../models';
+
+const {
+  emails,
+  displayName,
+  avatarUrl,
+  passwordCreated,
+  recoveryKey,
+} = useAccount();
```
dannycoates, line 19: This should be in the `Settings` component function because `useAccount()` needs to be called after the app has initialized instead of when this file is first loaded.

packages/fxa-settings/src/components/Settings/index.tsx
```diff
@@ -7,26 +7,31 @@ import UnitRow from '../UnitRow';
...
+
+export const Settings = () => {
+  const primaryEmail = emails.find((email) => email.isPrimary)!;
```
dannycoates, line 22: You can use `account.primaryEmail` which comes from here: https://github.com/mozilla/fxa/blob/issues%2F6118/packages/fxa-settings/src/lib/cache.ts#L35

packages/fxa-settings/src/components/Settings/index.tsx
meandavejustice: Just pushed a fix for this

PR #6179, opened 2020-08-10
fix(email): Lookup account by `originalLoginEmail` if present

vbudhram:
## Because

- When users change their primary email address would could potentially do an account lookup on the wrong email value
- The `originalLoginEmail` is a value that the users password is hashed with

## This pull request

- Uses the `originalLoginEmail` if present

Closes: https://bugzilla.mozilla.org/show_bug.cgi?id=1635027

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/utils/signin.js
+++ b/packages/fxa-auth-server/lib/routes/utils/signin.js
@@ -178,6 +178,18 @@ module.exports = (log, config, customs, db, mailer, cadReminders) => {
           // If we didn't load it above while checking unblock codes,
           // it's now safe to load the account record from the db.
           if (!accountRecord) {
+            // If `originalLoginEmail` is specified, we need to fetch the account record tied
+            // to that email. In the case where a user has changed their primary email, the `email`
+            // value here is really the value used to hash the password and has no guarantee to
+            // belong to the user.
+            if (request.payload.originalLoginEmail) {
+              return db
+                .accountRecord(request.payload.originalLoginEmail)
+                .then((result) => {
+                  accountRecord = result;
+                });
+            }
+
             return db.accountRecord(email).then((result) => {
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     // Permit, but do not report errors for JS files
     // TODO: Remove after transition to TS is complete
     "checkJs": false,
-    "types": ["accept-language", "mocha", "mozlog", "node"]
+    "types": ["accept-language", "mocha", "mozlog", "node"],
+    "esModuleInterop": true
```
dannycoates, line 10: `esModuleInterop` is already true in `tsconfig.base.json`. why here?

packages/fxa-auth-server/lib/routes/utils/signin.js
```diff
@@ -178,6 +178,18 @@ module.exports = (log, config, customs, db, mailer, cadReminders) => {
           // If we didn't load it above while checking unblock codes,
           // it's now safe to load the account record from the db.
           if (!accountRecord) {
+            // If `originalLoginEmail` is specified, we need to fetch the account record tied
+            // to that email. In the case where a user has changed their primary email, the `email`
+            // value here is really the value used to hash the password and has no guarantte to
```
meandavejustice, line 183: guarantte >> guarantee

meandavejustice:
What a scenario 😮

packages/fxa-auth-server/tsconfig.json
vbudhram: I was getting some typescript errors when running tests through webstorm...which now I can't seem to reproduce. From what I recall it was trying to use fs to load config and failed because it couldn't find the import. I'll revert this it seems to be working now 🤷🏽‍♂️

PR #6181, opened 2020-08-10
fix(payments): send subscription download email along with initial invoice

### LINKED ISSUE #6175: Only send download email in SCA flow when the subscription is successful
In FXA-2391, I found that we weren't sending the product download email at all by the end of the SCA flow.

[The patch I wrote](https://github.com/mozilla/fxa/pull/6126) now sends the download email with every attempt to process payment during the subscription creation process - that includes failed 3D Secure auth flows and invoice retries.

This should be refined to send the email only on successful payment status - and also at the end of successful invoice retry, not just initial subscription creation.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2415)
┆Issue Number: FXA-2415

lmorchard:
In the new SCA payment flow, there is at least one case where we'll only
know that the subscription has been successfully paid for when the
webhook event comes in from Stripe. (i.e. client-side 3D Secure
authentication)

But, as far as I can tell, we can't tell the difference between that and any other
case. So, this PR switches all cases over to sending the download email
along with initial invoice in response to the invoice webhook event.

fixes #6175

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -47,7 +47,7 @@ const VALID_RESOURCE_TYPES = [
   PLAN_RESOURCE,
   CHARGES_RESOURCE,
   INVOICES_RESOURCE,
-];
+] as const;
 
 export const SUBSCRIPTION_UPDATE_TYPES = {
   UPGRADE: 'upgrade',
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1587,7 +1587,7 @@ export class StripeHelper {
    */
   async expandResource<T>(
     resource: string | T,
-    resourceType: string
+    resourceType: typeof VALID_RESOURCE_TYPES[number]
```
lmorchard, line 1590: This type tweak isn't strictly necessary after I went through several iterations with this PR. But, it seems a decent improvement? Any module using `expandResource` in Typescript should now get hints as to the proper choices for `resourceType`

PR #6182, opened 2020-08-10
fix(payments): do not send payment failure emails during subscription creation flow

### LINKED ISSUE #6129: “Mozilla VPN payment failed” email received while the 3D Secure popup is triggered
#### Affected versions:
- All Firefox;

#### Affected FxA versions:
- 1.182.0;

#### Affected Platforms:
- Windows 10 (64-bit);

#### Prerequisites:
- Have an FxA account;

 #### Steps to reproduce:
1. Load https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net/vpn page;
2. Click on the “Try Mozilla VPN” button;
3. Sign in with the account from prerequisites;
4. Change the URL by adding a “/v2” value after .net/
5. Fill all mandatory fields and click the “Pay Now” button;
6. Open you email client and check inbox;

#### Expected result:
- No email is received while the 3D secure popup is triggered.

#### Actual result:
- A “payment failed” email is received when the popup is triggered.

#### Notes:
- Also, 2 “payment failed” mails are received for each of the following actions: cancel popup and fail authentication. (maybe intended?)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2385)
┆Issue Number: FXA-2385

lmorchard:
fixes #6129

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions.ts
@@ -1027,6 +1027,10 @@ class DirectStripeRoutes {
     event: Stripe.Event
   ) {
     const invoice = event.data.object as Stripe.Invoice;
+    if (invoice.billing_reason !== 'subscription_cycle') {
+      // Send payment failure emails only when processing a subscription renewal.
+      return;
+    }
     const { uid, email } = await this.sendSubscriptionPaymentFailedEmail(
       invoice
     );
--- a/packages/fxa-auth-server/test/local/routes/subscriptions.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions.js
@@ -2265,21 +2265,27 @@ describe('DirectStripeRoutes', () => {
     });
 
     describe('handleInvoicePaymentFailedEvent', () => {
-      it('sends email and emits a notification when an invoice payment fails', async () => {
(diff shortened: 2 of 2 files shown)


packages/fxa-auth-server/lib/routes/subscriptions.ts
```diff
@@ -1027,6 +1027,10 @@ class DirectStripeRoutes {
     event: Stripe.Event
   ) {
     const invoice = event.data.object as Stripe.Invoice;
+    if (invoice.billing_reason === 'subscription_create') {
```
bbangert, line 1030: Since the invoice was created because the subscription was being created, won't this be set the first time it fails? (Assuming I'm reading the docs correctly here: https://stripe.com/docs/api/invoices/object#invoice_object-billing_reason)

If I'm reading this right, repeat failures will then have billing_reason set to `subscription_update`.

bbangert: One more note, I *think* the only time we want to send a payment failed here is if the `billing_reason` is `subscription_cycle`, because it implies that it happened as a result of Stripe attempting to make the payment, vs. the initial or subsequent card updates.

lmorchard: Giving this another read and yeah I think `subscription_cycle` is what we want here.

Issue #5206, opened 2020-05-04
'Unverified session' error on force-2FA flow


6a68:
Looks like signing in to 123done with the 2fa required flow can sometimes run into an error where, after clicking 'sign in', the inline 2FA flow isn't opened. Instead, an "Unverified session" 500 error is shown.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1822)
┆Issue Number: FXA-1822

6a68:
Still exploring steps to reproduce this, but it appears this happens when

* account A is logged in to the browser
* account B logs in to 123done with the '2fa required' option
* neither account has 2fa already enabled

6a68:
I'm struggling to find steps to reproduce this bug.

The only time I've definitely seen this is when I was signed into the browser (production FxA environment), then tried to sign in to staging or stable dev 123done using the "require 2FA" button.

diox:
I can reproduce with my account on https://stable.dev.lcip.org without being logged in to the browser (fresh profile).

6a68:
Huh. Can you reproduce with a new account, or just that one?

diox:
I tried with a second existing account and got the 500 again. Tried creating a new account and it worked.

bsurd:
Tried the following:
Sign In (Require 2FA) - 2FA not setup - no issues
Sign In - 2FA not setup - no issues
Sign In (Require 2FA) - 2FA setup - no issues
Sign In - 2FA setup - no issues

But all of these were done using a new account, from what @diox mentioned it could be that the issue is reproduced with pre-existing accounts only. So there might be something there. Sadly, can't confirm if this is the case because I don't really have any pre-existing accounts as I tend to delete them after I'm done testing with them.

diox:
FWIW AMO QA team probably has a lot of old FxA accounts laying around on https://stable.dev.lcip.org given that's the server we use for our dev instance where they test patches after they land on AMO. But I'm not sure this would help give us steps to reproduce beyond "use an existing account".

vasilicatamas:
I constantly reproduce this issue running the following steps:
1. Launch Firefox with a clean profile.
2. Navigate to https://stable.dev.lcip.org/ 
3. Create and confirm a new account.
4. Launch Firefox with a new clean profile
5. Navigate to https://123done-stable.dev.lcip.org/
6. Click on “Sign In (Require 2FA)” button.
7. Type the account credentials and click the “Sign in” button.

[...]

PR #6183, opened 2020-08-11
fix(totp): Ensure the user has a verified session before starting inline TOTP setup

6a68:
Still need to update tests, but want to verify that I'm on the right track here.

## Because

- When a user with an unverified session tries to log in to an RP that wants to force TOTP setup, the user encounters a fatal error

## This pull request

- Explicitly checks the user's session verification status before rendering the inline TOTP setup flow. If the session is unverified, the view now redirects to the verification flow (via `/signin_token_code`), then redirects back.

Note that I had originally thought I could just rely on checking the `verified` status in the `hasTotpToken` auth-db response, but after looking at the authdb code, it seems that just tells us if the TOTP token is verified, not if the user's session is verified. It also seems nicer to rely on regular control flow, rather than exception flow, to check for this edge case.

Closes: #5206

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/inline_totp_setup.js
+++ b/packages/fxa-content-server/app/scripts/views/inline_totp_setup.js
@@ -72,15 +72,25 @@ var View = FormView.extend({
 
   beforeRender() {
     const account = this.getAccount();
+
     if (!account.get('sessionToken')) {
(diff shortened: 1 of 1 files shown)


6a68:
Hmm, not quite working yet. I'll reopen when it's farther along.

PR #6184, opened 2020-08-11
New settings: display alerts from content-server

### LINKED ISSUE #4933: [global] Account verified/pw reset messages, navigate/redirects to settings paths
Check out `fxa-content-server/app/scripts/views/behaviors/settings.js` + `/behaviors/navigate.js`, `fxa-content-server/app/scripts/models/auth_brokers/web.js`, and related files. This is where we place "Account verified successfully," "Password reset successfully," etc., and redirects users to where they should be in settings.

We need to ensure users will be able to verify their account etc. and see success messages on the new settings page, and that they'll be redirected to the correct flow based on the `redirectTo` param sent from the server. All of the affected routes may or may not be setup when the work for this issue starts.

Can we use these files with tweaks and allow this portion of the content server to handle this behavior for us (likely not for success messaging since we'd need to relay that to the React app), or do we need to set up a new version (or partial version just for messaging) of these files?

The PR for this should allow for a success message to show in the AlertBar component on the new Settings page when an account is verified and a user resets their password, but because we don't need this functionality while settings lives on `/beta/settings`, we may need to have this change live behind the `enableBeta` config option or a comment to reference on how others can test the functionality locally. File a follow up issue if needed to ensure this functionality will be "on" when we switch from `beta/settings` to `/settings`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1633)
┆Issue Number: FXA-1633

dannycoates:
Some actions in fxa-content-server end up displaying an alert message on the settings page. We want to be able to display those messages across the "app" boundary between the content-server front-end and new settings.

To do this we store the message in localStorage and whichever app gets to it first will display it and remove it from storage.

closes #4933

### DIFF
--- a/packages/fxa-settings/src/lib/cache.ts
+++ b/packages/fxa-settings/src/lib/cache.ts
@@ -1,27 +1,67 @@
-import { InMemoryCache, gql } from '@apollo/client';
+import { InMemoryCache, gql, makeVar } from '@apollo/client';
 import Storage from './storage';
 import { Email } from '../models';
 
 const storage = Storage.factory('localStorage');
 
-export function sessionToken(newToken?: string) {
+export interface OldSettingsData {
+  uid: hexstring;
+  sessionToken: hexstring;
+  alertText?: string;
+}
+
+type LocalAccount = OldSettingsData | undefined;
+type LocalAccounts = Record<hexstring, LocalAccount> | undefined;
+
+function accounts(accounts?: LocalAccounts) {
+  if (accounts) {
+    storage.set('accounts', accounts);
+    return accounts;
+  }
+  return storage.get('accounts') as LocalAccounts;
+}
+
+function currentAccount(account?: OldSettingsData) {
+  const all = accounts() || {};
+  const uid = storage.get('currentAccountUid') as hexstring;
+  if (account) {
+    all[account.uid] = account;
+    accounts(all);
+    return account;
+  }
+  return all[uid];
+}
+
+export function sessionToken(newToken?: hexstring) {
   try {
-    const storedAccounts = storage.get('accounts');
-    const currentAccountUid = storage.get('currentAccountUid');
+    const account = currentAccount();
     if (newToken) {
-      storedAccounts[currentAccountUid].sessionToken = newToken;
-      storage.set('accounts', storedAccounts);
+      account!.sessionToken = newToken;
+      currentAccount(account);
     }
-    return storedAccounts[currentAccountUid].sessionToken as string;
+    return account!.sessionToken;
   } catch (e) {
     return null;
   }
 }
 
+function consumeAlertTextExternal() {
+  const account = currentAccount();
+  const text = account?.alertText || null;
+  if (text) {
+    account!.alertText = undefined;
+    currentAccount(account);
+  }
+  return text;
+}
+
+export const alertTextExternal = makeVar(consumeAlertTextExternal());
+
 // sessionToken is added as a local field as an example.
 export const typeDefs = gql`
   extend type Account {
     primaryEmail: Email!
+    alertTextExternal: String
   }
   extend type Session {
     token: String!
@@ -36,6 +76,11 @@ export const cache = new InMemoryCache({
           read(_, o) {
             const emails = o.readField<Email[]>('emails');
             return emails?.find((email) => email.isPrimary);
+          },
+        },
+        alertTextExternal: {
+          read() {
+            return alertTextExternal();
           },
         },
       },
--- a/packages/fxa-content-server/app/scripts/models/account.js
+++ b/packages/fxa-content-server/app/scripts/models/account.js
@@ -55,6 +55,7 @@ const PERSISTENT = {
   sessionTokenContext: undefined,
   uid: undefined,
   verified: undefined,
+  alertText: undefined,
 };
 
 const DEFAULTS = _.extend(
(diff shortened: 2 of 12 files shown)


packages/fxa-settings/src/components/ExternalAlert/index.test.tsx
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { useRef, ReactNode } from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
jodyheavener, line 7: You should be able to boot this out now

packages/fxa-settings/src/components/ExternalAlert/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const ExternalAlert = () => {
+  const account = useAccount();
+
+  return account.externalAlert ? (
+    <AlertBar
+      onDismiss={() => {
+        externalAlert(null);
+      }}
+    >
+      <p data-testid="alert-text">{account.externalAlert}</p>
```
LZoog, line 15: Can we call this `external-alert-text` (or `alert-external-text` if we change the name of the component)?

packages/fxa-settings/src/components/ExternalAlert/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { AlertBar } from '../AlertBar';
+import { useAccount } from '../../models';
+import { externalAlert } from '../../lib/cache';
+
+export const ExternalAlert = () => {
```
LZoog, line 6: We _could_ call this`AlertExternal` instead of `ExternalAlert` - that way in the component tree list, it's next to `AlertBar`. We've been trying to stick with `[noun][descriptor]` rather than `[descriptor][noun]` for that reason but not sure that'll be easy to keep up with conventionally so I'm fine with either. If we leave this `ExternalAlert` it temps me to change `LinkExternal` to `ExternalLink`.

packages/fxa-settings/src/components/ExternalAlert/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <MockedCache account={{ externalAlert: 'ok' }}>
+        <AlertBarRootAndContextProvider>
+          <ExternalAlert />
+        </AlertBarRootAndContextProvider>
+      </MockedCache>
+    );
+    expect(screen.getByTestId('alert-bar-root')).toContainElement(
+      screen.getByTestId('alert-bar')
+    );
+    expect(screen.queryByTestId('alert-text')).toBeInTheDocument();
```
LZoog, line 29: I think `expect(screen.queryByTestId('alert-bar')).not.toBeInTheDocument()` might be more explicit - ensure the AlertBar isn't rendered at all.

packages/fxa-settings/src/lib/cache.ts
```diff
@@ -1,27 +1,67 @@
-import { InMemoryCache, gql } from '@apollo/client';
+import { InMemoryCache, gql, makeVar } from '@apollo/client';
 import Storage from './storage';
 import { Email } from '../models';
 
 const storage = Storage.factory('localStorage');
 
-export function sessionToken(newToken?: string) {
+export interface OldSettingsData {
+  uid: hexstring;
```
LZoog, line 8: This might be a "my machine" problem but I'm getting complaints on using `hexstring` (`Cannot find name 'hexstring'`), as well as where we're using `Hash`.

packages/fxa-settings/src/components/ExternalAlert/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    rerender(
+      <MockedCache>
+        <AlertBarRootAndContextProvider>
+          <ExternalAlert />
+        </AlertBarRootAndContextProvider>
+      </MockedCache>
+    );
+    expect(screen.queryByTestId('alert-text')).toBeNull();
+  });
+});
```
LZoog, line 47: Oh, one additional nice-to-have test here could be that the alert bar shows and then dismisses properly on close since `onDismiss` is handled slightly differently.

packages/fxa-settings/src/components/ExternalAlert/index.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      </MockedCache>
+    );
+    rerender(
+      <MockedCache>
+        <AlertBarRootAndContextProvider>
+          <ExternalAlert />
+        </AlertBarRootAndContextProvider>
+      </MockedCache>
+    );
+    expect(screen.queryByTestId('alert-text')).toBeNull();
```
LZoog, line 45: I think `expect(screen.queryByTestId('alert-bar')).not.toBeInTheDocument()` might be more explicit - ensure the AlertBar isn't rendered at all.

(sorry, I had this comment on the wrong line previously and deleted it)

packages/fxa-settings/src/lib/cache.ts
dannycoates: it ought to be in your `react-app-env.d.ts` file. in vscode you may need to `Cmd+Shift+P` - `Restart TS server`

packages/fxa-settings/src/components/ExternalAlert/index.test.tsx
dannycoates: _squints_ .... meh tbh

PR #6187, opened 2020-08-11
feat(settings): Build out secondary emails with real data, add resend verification GQL mutation

LZoog:
## Because

* Secondary email states were hardcoded and we needed to reflect real account data.
* Users need to be able to resend the verification code for unverified secondary emails.

## This pull request

* Displays various secondary email states with real account data - none set with verified/unverified primary email, single/multi secondary emails set unverfied/verified.
* Adds a functioning mutation for resendSecondaryEmailCode.

Closes: #4903

## Checklist

## Other information

I'm going to make a follow up issue to immediately address the single test I commented out plus DRYing up some of the test approaches here plus add some docs about our Providers (Context, Router).


packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,226 @@
...
+              {!secondaryEmailIsVerified && (
+                <span className="uppercase block text-orange-600 font-bold text-xs">
+                  unverified
+                </span>
+              )}
+            </p>
+            {secondaryEmailIsVerified &&
+              lastVerifiedSecondaryEmailIndex === index && (
+                <VerifiedEmailContent />
+              )}
```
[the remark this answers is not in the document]
LZoog: Yes, I actually have a comment here I haven't pushed up:
```
{/* Only include VerifiedEmailContent on last verified secondary email */}
```

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,226 @@
...
+            </div>
+          </div>
+        </div>
+      </>
+    );
+  };
+
+  if (!hasAtLeastOneSecondaryEmail) {
+    const route = primaryEmailIsVerified
+      ? '/beta/settings/secondary_email'
```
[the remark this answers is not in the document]
LZoog: `constantantized` is my new favorite word 😛 They both are set to `const` so I'm not sure I know what you mean - this was a bit funky though.

I also prefer dashes, but using the same URLs as current settings takes a lot of redirects out of the question. There's probably a lot of SUMO links, posts all around the 'net, even bookmarked URLs that we'd want to account for otherwise.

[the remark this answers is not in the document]
jodyheavener: > I'm not sure I know what you mean

Sorry, I could have been more clear. With how frequently specific routes will be references (from setting up the router, to `<Link />` usage throughout the app, to testing) I think it would be beneficial to use a global paths object, similar to how content-server uses a [constants file](https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/lib/constants.js). This would allow IDEs to auto-complete/auto-import paths, and eliminate accidental incorrect paths (e.g. someone uses a dash instead of an underscore).

> I also prefer dashes, but using the same URLs as current settings takes a lot of redirects out of the question. There's probably a lot of SUMO links, posts all around the 'net, even bookmarked URLs that we'd want to account for otherwise.

I think a relatively simple solution would be to 304 Redirect the old URLs to the new (looks like React Router might be able to do something similar with [`push`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md#push-bool)). IMO this feels like one of those things we should try to do while we have the opportunity, before things are set in stone? Then again, it's only underscores 😛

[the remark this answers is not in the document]
LZoog: Ah I see, that sounds like a good idea to me and should probably be done in #6131 or a follow up!

It might be worth bringing up to Alex/Meridel. I know we briefly talked about it but just agreed to stick with parity stuff for less complications.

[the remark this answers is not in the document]
jodyheavener: Sweet!

And Sweet! This is also something that could be done at a later date before launch.

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -13,8 +13,8 @@ type UnitRowProps = {
   noHeaderValueCtaText?: string;
   children?: React.ReactNode;
   headerValueClassName?: string;
-  route?: string;
-  revealModal?: () => void;
+  route?: string | null;
+  revealModal?: (() => void) | null;
```
jodyheavener, line 17: Code clarity nitpick: I see this is potentially `null` because of the how the props are [composed](https://github.com/mozilla/fxa/pull/6187/files#diff-a313ff76f12b71cf55da0fc953b5dabbR76). In the spirit of keeping the intended type declaration clear (as in, `null` serves no purpose other than placeholder), could we maybe update either how the prop object is constructed (e.g. if the value is not present, don't set the property instead of setting it to `null`), **or** instead of `null` set the [alternative](https://github.com/mozilla/fxa/pull/6187/files#diff-a313ff76f12b71cf55da0fc953b5dabbR65) to `undefined` (that way the types can remain unchanged)?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,56 +1,90 @@
...
 
-  const onSecondaryEmailConfirm = useCallback(() => {
-    console.log('confirmed - resend verification code');
-    hideModal();
-    revealAlertBar();
-  }, [hideModal, revealAlertBar]);
+  const UnitRowSecondaryEmailNotSet = () => {
+    const [modalRevealed, revealModal, hideModal] = useBooleanState();
+    const modalHeaderId = 'modal-header-verify-email';
+    const modalDescId = 'modal-desc-verify-email';
```
jodyheavener, line 46: From looking at the flow I think I know the answer, but just wanted to verify: this component can only ever appear one time per page, right?

packages/fxa-settings/src/components/UnitRow/index.tsx
LZoog: > or instead of null set the placeholder to undefined

Heh, I actually did this at first, but changed it to `null` since it usually feels weird to intentionally set something to `undefined`, although I know we do that in other places. I can change it back to `undefined` instead of `null` so the type looks a bit better.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Yep, just once. It's in its own component so it can manage its own state since we don't need the modal if an e-mail is verified.

packages/fxa-settings/src/components/Nav/index.tsx
```diff
@@ -6,7 +6,7 @@ import React from 'react';
 import classNames from 'classnames';
 import LinkExternal from 'fxa-react/components/LinkExternal';
 import { ReactComponent as OpenExternal } from './open-external.svg';
-import { useAccount } from 'fxa-settings/src/models';
+import { useAccount } from '../../models';
```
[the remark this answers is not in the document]
meandavejustice: I've been mostly seeing absolute paths for things within the package you are working it, I think it makes sense for those, as it shows you that the file is within the same package at a glance, but I would be happy either way, so long as it's consistent.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
 
-import React, { useCallback } from 'react';
+import React, { useCallback, useRef } from 'react';
 import { useBooleanState } from 'fxa-react/lib/hooks';
+import { gql, useMutation } from '@apollo/client';
+import { useFocusOnTriggeringElementOnClose } from '../../lib/hooks';
 import UnitRow from '../UnitRow';
 import Modal from '../Modal';
 import AlertBar from '../AlertBar';
+import { useAccount, Email } from '../../models';
```
meandavejustice, line 12: Do we always want to be using `useAccount` in the deepest possible component?

For instance, currently we are using `useAccount` in `Settings/index.tsx` and passing in some options to `<Security/>` (and `<Profile/>` in my coming pr, which I will rebase with this after it lands).

Is the philosophy to localize the usage of `useAccount` as much as possible and attempt to _never_ pass props?

meandavejustice:
This is looking good so far, I have a branch with some things for https://jira.mozilla.com/browse/FXA-1648, but It's really dependent on this work, so I'll set aside until this is complete. Feel free to ping me when you are ready for another review and I'll pick it up immediately :)

packages/fxa-settings/src/components/Nav/index.tsx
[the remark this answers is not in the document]
dannycoates: tbh i think the fxa-settings one was an auto-import (mistake) from vscode. we should use relative paths "../" since it plays nicer with both ts and compiled js

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+        {
+          email: 'johndope3@example.com',
+          isPrimary: false,
+          verified: true,
+        },
+        {
+          email: 'johndope4@example.com',
+          isPrimary: false,
+          verified: false,
+        },
```
jodyheavener, line 207: Heh, there are about 30 instances of this type of object. Maybe we can dry this up with a factory in `_mocks` or another shared test file? Something like:

```javascript
function mockAccountEmail(email, isPrimary, verified) {
  const email = email || autoGeneratedEmail
  return {
    email,
    isPrimary,
    verified,
  },
}
```

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+  // describe('resendSecondaryEmailCode', () => {
+  //   it('displays a success message in the AlertBar', () => {
+
+  //   });
+
+  //   it('displays an error message in the AlertBar', () => {
+
+  //   });
+  // })
```
jodyheavener, line 236: You likely already know about this, but dropping a comment to be sure :)

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
jodyheavener: That's my understanding, though I'm sure it wouldn't be the worst if you passed an individual prop down to a simple UI component. Maybe others have differing opinions on this.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
+  index: number;
 };
 
-export const UnitRowSecondaryEmail = ({ primaryEmail }: UnitRowSecondaryEmailProps) => {
+export const UnitRowSecondaryEmail = () => {
+  const account = useAccount();
+  const primaryEmail = account.primaryEmail.email;
+  const primaryEmailIsVerified = account.primaryEmail.verified;
+  const secondaryEmails = account.emails.filter((email) => !email.isPrimary);
+  const hasAtLeastOneSecondaryEmail = secondaryEmails.length;
```
jodyheavener, line 32: Nit: while a positive value would evaluate to true, it might be more clear to `!!` or `> 0`

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
+  const lastVerifiedSecondaryEmailIndex = secondaryEmails
+    .map((email) => email.verified === true)
```
jodyheavener, line 34: Nit: if `verified` is always bool you should be able to forgo `=== true` maybe

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
+              Before you can add a secondary email, you must verify your primary
+              email. To do this, you'll need access to {primaryEmail}
             </p>
           </Modal>
         )}
+      </UnitRow>
+    );
+  };
+
+  const UnitRowSecondaryEmailContentAndActions = ({
```
jodyheavener, line 112: Thought: since we are already in the "secondary email" context because of the component name, prefixing everything with "secondary" could be redundant. You could call the main record object `secondaryEmail` and destructure the two variables below down to `const { email, verified } = secondaryEmail;` without fear of namespace conflict. It also allows the variable names to be much shorter, less code overall, and you could use that sweet, sweet shorthand property syntax in the query on L121.

This idea could actually be implemented in all the areas that prefix the word secondary onto variables that exist inside an already-secondary context.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
+        <div className="mobileLandscape:flex unit-row-multi-row">
+          <div className="unit-row-content" data-testid="unit-row-content">
+            <p className="font-bold" data-testid="unit-row-header-value">
+              {secondaryEmail}
+              {!secondaryEmailIsVerified && (
+                <span
+                  data-testid="unverified-text"
+                  className="uppercase block text-orange-600 font-bold text-xs"
+                >
+                  unverified
```
jodyheavener, line 148: Nit: this might be visually uppercase but should prob be capitalized

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -1,80 +1,235 @@
...
+      <div className="unit-row-header">
+        <h3 data-testid="unit-row-header">Secondary email</h3>
+      </div>
+      <div className="mobileLandscape:flex-3 desktop:flex-5">
+        {secondaryEmails.map((secondaryEmailObj, index) => (
+          <UnitRowSecondaryEmailContentAndActions
+            key={secondaryEmailObj.email}
+            {...{
+              secondaryEmailObj,
+              index,
```
jodyheavener, line 207: Instead of the component trying to figure out if it's the last one of the bunch, could we just tell it? Then it doesn't have to do any index logic bidness

packages/fxa-settings/src/styles/unit-row.scss
```diff
@@ -9,6 +9,10 @@
     @apply font-header w-full mb-1;
   }
 
+  &-multi-row + &-multi-row {
+    @apply mt-4;
+  }
```
jodyheavener, line 14: Looks like the resulting selector here is `.unit-row-multi-row + .unit-row-multi-row { ... }`. I know this is far more valuable in the CSS than conditionally applying classes in JS, though. I'm just... nervous 😜

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Yeah, since `useAccount` reads from the cache it makes sense to me to use it where we need the data. It's kind of like Context in the sense that we're pulling in those app-global values. It could make sense to pass a piece along in the props but I'd say usually it'd make more sense to just pull it in from `useAccount` where needed.

 where we can instead of drilling props down. We could pass a single prop down to a component if it made sense to, but once that component has another component...

packages/fxa-settings/src/styles/unit-row.scss
LZoog: Yeah, doing it in JS is possible but I'd need to conditionally check the index in `UnitRowSecondaryEmailContentAndActions` and apply `mt-4` if it's not the first index. I chose to go this route because if we ever need multi-rows again, it's more readily available and doesn't have to be applied with JS. I get the concern but I think this is OK personally, it's only two selectors deep and our custom CSS is pretty sparse.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: You mean `isLastSecondaryEmail={index === lastVerifiedSecondaryEmailIndex}`? I can do that.

packages/fxa-settings/src/styles/unit-row.scss
jodyheavener: Yeah I don't think it needs to be changed. Just gives me the willies ;)

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Meh, why?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Ugh yeah, my problem with that is that this component also deals with the primary email in some instances, like checking `primaryEmailIsVerified`, showing the primary email in the alert bar, and resending the primary email verification code (done in a later task). I honestly feel like being explicit here is going to be much nicer when in a couple of years someone who's never touched this before can look at the file and tell what's going on without ambiguity. I'm a fan of shorter names, I just have this gut feeling that it's better to be overly specific here to help with any potential confusion later on if something is referring to primary or secondary. Whatchu think?

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
jodyheavener: Proper grammar? 😛

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.test.tsx
meandavejustice: This is fxa-1648

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.test.tsx
LZoog: Per Slack convo, I'm going to do this in a follow up and probably create a test utils or helpers file, it'll be paired with some doc updates.

Issue #1445, opened 2017-03-01
[Meta] Password Manager Saves FxA Password [bz1343269]


bug-mirror:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1343269

This is a meta bug to track all the various issues and concerns being filed about the FxA password being saved in the password manager.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-146)
┆Issue Number: FXA-146

PR #6197, opened 2020-08-13
fix(content-server): create fake hidden input on sign in

byrw:
Because:

* When trying to sign in with wrong password, the "Save Password" is displayed.

This commit:

* Create a fake input with "required" atribute to trick the browser to not showing the "Save Password" doorhanger.


jodyheavener:
Hey @byrw is there an issue associated with this PR?

byrw:
> Hey @byrw is there an issue associated with this PR?

 From the open issues #1445 is the only related, but if you want I can open a new issue.

jodyheavener:
No that's okay, it looks like there are other things that need to be done before it can be closed. I just like to have a "paper trail" for issues :)

PR #6198, opened 2020-08-13
fix(storybook): delete commits and pulls over 3 weeks old on every build

lmorchard:
Because I assumed that git retained file creation timestamps, I had a simple find command to delete old commits and pulls after 90 days. Two problems here:

1. 90 days is too long
2. git doesn't retain file creation timestamps anyway - so all files look as if they were created at the beginning of a CircleCI run.

Thus, aging storybook builds have been piling up in [mozilla-fxa/storybooks gh-pages branch](https://github.com/mozilla-fxa/storybooks/tree/gh-pages) and we hit a 10GB limit. 

Still thinking about a cleaner solution, maybe in node.js. But for now, this change should do [what I did manually](https://github.com/mozilla-fxa/storybooks/commit/2f692f3f8cdfa5ba156817db9809bcca4d502aaf).

It's an ugly hack that munges git log output to find `commit/*` and `pull/*` directories to delete old builds.

Not sure if this will run on Mac OS X with the specific Linux shell pipeline.

Trying to think if there might be a cleaner way to do this.

As a follow-up, we could consider moving all of this to an AWS S3 bucket and use object expiration policies there to auto-clean.

### DIFF
--- a/_scripts/build-storybooks.sh
+++ b/_scripts/build-storybooks.sh
@@ -13,7 +13,7 @@ STORYBOOKS_REPO="${STORYBOOKS_REPO:-mozilla-fxa/storybooks}"
 STORYBOOKS_URL="${STORYBOOKS_URL:-https://mozilla-fxa.github.io/storybooks}"
 STORYBOOKS_BRANCH="${STORYBOOKS_BRANCH:-gh-pages}"
 PUBLISH_ROOT="${PUBLISH_ROOT:-storybooks-publish}"
-MAX_STORYBOOK_AGE="${MAX_STORYBOOK_AGE:-90}"
+MAX_STORYBOOK_AGE="${MAX_STORYBOOK_AGE:-3 weeks ago}"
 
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/.storybook/config.js
```diff
@@ -8,11 +8,12 @@ const reqFromComponents = require.context(
   true,
   /\.stories.tsx?$/
 );
+
 const reqFromRoutes = require.context('../src/routes', true, /\.stories.tsx?$/);
 
 function loadStories() {
-  [reqFromComponents, reqFromRoutes].forEach(req =>
-    req.keys().forEach(filename => req(filename))
+  [reqFromComponents, reqFromRoutes].forEach((req) =>
```
lmorchard, line 15: Wanted to make a small change to trigger a payments-server storybook build and I guess prettier wanted in on the fun

_scripts/build-storybooks.sh
```diff
@@ -78,8 +78,13 @@ if [[ ! -z $CI_PULL_REQUEST ]]; then
 fi;
 
 # Delete storybooks older than 90 days
```
vbudhram, line 80: could probably just say `Delete older storybooks`

PR #6200, opened 2020-08-14
chore(event-broker): dependency update

bbangert:
Because:

* Staying up to date is helpful.

This commit:

* Updates event-broker dependencies.

## Checklist


dannycoates:
I prefer dep updates across all packages vs individual ones. It isn't always practical, but I think its good to keep dependency versions in sync across all packages (and it keeps our project footprint smaller). Looking at the lock diff, this change adds a bunch of new versions without removing old ones, meaning at least one of our other services is still using it. The main tool I've used to check dep versions across packages is `yarn why [dep]` which can be time consuming on a mass update like this, maybe there's a better tool idk. There's an easy way to see which deps are using unique versions per package by looking at their local node_modules dir since everything that's shared is up in the project root node_modules.

bbangert:
Okey, I can look at that for future updates. I was aiming for one package at a time because a mass update seems more likely to cause mass breakage that would then be hard to pinpoint on where it actually came from. This is magnified by things like content-server which in-advertantly test many parts of the whole.

PR #6203, opened 2020-08-14
chore: dependency updates

bbangert:
Because:

* We like to keep our dependencies up to date.
* Using shared versions is more size efficient.

This commit:

* Updates dependencies for graphql-api where applicable.
* Consolidates shared dependency updates to reduce the size.

## Checklist


bbangert:
Updates that didn't make it cause tests broke:
auth-server: nock 13.*
content-server: sinon

PR #6205, opened 2020-08-14
chore: update typescript

bbangert:
Because:

* Using the latest everywhere helps with consistency and smaller build
  size.

This commit:

* Updates for TypeScript 3.9.7 everywhere in FxA.

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/lib/test-utils.tsx
+++ b/packages/fxa-payments-server/src/lib/test-utils.tsx
@@ -166,7 +166,16 @@ function injectStripe<P extends Object>(
 ) {
   return (props: P) => {
     const { mockStripe } = useContext(MockStripeContext);
-    return <WrappedComponent {...{ ...props, stripe: mockStripe }} />;
+    return (
+      <WrappedComponent
+        {...{
+          ...props,
+          stripe: mockStripe as ReactStripeElements.InjectedStripeProps['stripe'],
+        }}
+        as
+        any
+      />
+    );
   };
 }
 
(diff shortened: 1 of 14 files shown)


packages/fxa-admin-server/package.json
```diff
@@ -70,7 +70,7 @@
     "supertest": "^4.0.2",
     "ts-node": "^8.10.2",
     "ts-sinon": "^1.2.0",
-    "typescript": "3.9.7",
+    "typescript": "^3.9.7",
```
bbangert, line 73: `yarn up` set these all to use `^`, not sure how I feel about it.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -166,7 +166,16 @@ function injectStripe<P extends Object>(
 ) {
   return (props: P) => {
     const { mockStripe } = useContext(MockStripeContext);
-    return <WrappedComponent {...{ ...props, stripe: mockStripe }} />;
+    return (
+      <WrappedComponent
+        {...{
+          ...props,
+          stripe: mockStripe as ReactStripeElements.InjectedStripeProps['stripe'],
```
bbangert, line 173: Newer typescript determined this was an issue.

packages/fxa-payments-server/src/lib/validator.ts
```diff
@@ -127,7 +127,7 @@ export type FieldType = 'input' | 'stripe';
 type FieldStateKeys = 'fieldType' | 'value' | 'required' | 'valid' | 'error';
 type FieldState = {
   fieldType: FieldType;
-  value: any;
+  value?: any;
```
bbangert, line 130: Code was relying on this being optional, newer TypeScript flagged this.

Issue #4975, opened 2020-04-16
[2FA] Download, copy, print button functionality


data-sync-user:
Transfer over the download/copy/print functionality for these buttons.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1666)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/14355/2fa-download-copy-print.png">2fa-download-copy-print.png</a>
┆Issue Number: FXA-1666

data-sync-user:
➤ Dave Justice commented:

It looks like this was completed by [~jheavener@mozilla.com] 2 months ago according to git blame.

!2fa-download-copy-print.png!

data-sync-user:
➤ Jody Heavener commented:

[~djustice@mozilla.com]I just added the buttons but I don't think the download and print buttons work?

data-sync-user:
➤ Barry Chen commented:

All the buttons do something, but the formatting is inconsistent, e.g. on Windows the downloaded file is just one long string with no (visible) separators.

data-sync-user:
➤ Dave Justice commented:

I've verified that copy, download, and print all work both in storybook and in local dev +1

data-sync-user:
➤ Danny Coates commented:

This isn't done. The downloaded codes aren't formatted correctly

data-sync-user:
➤ Barry Chen commented:

The copied codes also use commas as part of the separators, which could be confusing to people who do not know that the comma itself isn't part of the code.

Issue #4973, opened 2020-04-16
[2FA] Validate code on "Continue", show step 2 of 3, generate keys


data-sync-user:
If the code is valid, show the "step 2 of 3" component on click of "Continue."

Generate the codes and display them on this page.

Download, Copy, Print (`GetDataTrio`) and "Continue" buttons should be present but not yet functional.

"Cancel" should take the user back to the settings page.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/409237382

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1665)
┆Issue Number: FXA-1665

Issue #4966, opened 2020-04-16
[recovery] Download, copy, print button functionality


data-sync-user:
Transfer over the download/copy/print functionality for these buttons.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1656)
┆Issue Number: FXA-1656

Issue #4964, opened 2020-04-16
[recovery] PW validation on "Continue," create/show step 2 of 2


data-sync-user:
Ensure the PW validation is in place on press of "Continue" from step 1. Create the "step 2 of 2" component to display on continuation.

The generation of the account recovery key will be done in a follow-up and can be a placeholder.

The Download, Copy, Print, and "Done" buttons (`GetDataTrio`) should be present but not yet functional.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/408917105

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1655)
┆Issue Number: FXA-1655

PR #6206, opened 2020-08-16
feat(settings): various ui components

### LINKED ISSUE #5030: [change pw] Password "eye" functionality + PasswordCurrent component
Create a reusable `<PasswordCurrent>` component for sharing with account recovery and delete account which should contain a password input with password eye functionality. Transfer over any focus state and any other relevant changes from the `fxa-content-server`.

Note that the password eye image must be in an `<img src` (or an inlined SVG preferably) for Android a11y.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1707)
┆Issue Number: FXA-1707

jodyheavener:
## Because

- We need various UI components as Settings development continues

## This pull request

Adds the following `fxa-settings` reusable components:

`TextInput` - A dressed up `<input type="text" />` that displays an overlaid label. Also supports other text-based types: `email`, `tel`, `number`, `url`, and `password`.

Screen Shot 2020-08-16 at 6 11 12 PM

`PasswordInput` - A wrapped `TextInput` that supports toggle-able password visibility.

Screen Shot 2020-08-16 at 6 10 29 PM

~`TextArea` - Visually similar to `TextInput`, just a dressed up `<textarea>`. <small>We don't actually appear to use this anywhere in our current Settings redesign, but it was an easy win and good ot have in our catalogue.</small>~

`Checkbox` - A fancy `<input type="checkbox" />` that supports inline labels.

Screen Shot 2020-08-16 at 6 09 31 PM

`DataBlock` - A `<button>` that can be used to displays blocks of mono-space text. This will typically be also used to copy the text value(s) to the clipboard, so it provides this functionality.

Screen Shot 2020-08-16 at 6 10 02 PM

`GetDataTrio` - (I'm open to renaming this) The three buttons we use in a couple places to download, copy, or print codes. Note these are pretty basic right now; the copy functionality works, the print functionality relies on an additional page to trigger the print, and the download functionality uses the `download` attribute with a URL.

Screen Shot 2020-08-17 at 8 20 43 AM

Closes #5030 #4924 
Partially address #4975 #4966 #4964 #4973

## Checklist

- [x] I have modified or added tests which pass locally.

## Other information

These have been built/styled according to our Settings redesign [Figma doc](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA-Manage-Settings-Copy?node-id=0%3A1).

I have attempted to provide enhanced a11y support across all of these components. Would love feedback on it.

📖 These components can be previewed in a [deploy of Storybook](https://mozilla-fxa.github.io/storybooks/commits/4bf882993b32fedf1ab1f3efd47cc31915855fad/fxa-settings/index.html). I encourage you to try out all the variations!


jodyheavener:
Now that I know about our [UI Library icons](https://www.figma.com/file/LNDDv2Ah55QQlbJEWNyhvq/Protocol-UI-Library?node-id=21%3A1562) I am tempted to suggest using those for the three buttons because the stroke widths are a little more consistent 🤔

packages/fxa-react/components/DataBlock/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const [inputId, setInputId] = useState<string | undefined>(id);
+  const readerId = `${inputId}-sr`;
+
+  async function copyToClipboard() {
+    const copyValue = valueIsArray
+      ? (value as string[]).join(', ')
+      : (value as string);
+
+    await copy(copyValue);
+    onCopied && onCopied(copyValue);
```
LZoog, line 47: Is this an optional callback for use later? I'm not sure why we need `dataBlockFocus` and `dataBlockBlur`

packages/fxa-react/components/PasswordInput/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+            hasContent ? '-ml-3' : 'hidden'
+          }`}
+          onClick={() => {
+            setVisible(!visible);
+          }}
+          title={visible ? 'Hide password' : 'Show password'}
+          aria-label={
+            visible
+              ? 'Hide password from screen.'
+              : 'Show password as plain text. Your password will be visible on screen.'
```
LZoog, line 78: We should confirm this copy with Meridel W (can be done in a follow up).

packages/fxa-react/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+              ? 'border-grey-300 text-grey-300'
+              : `cursor-pointer ${
+                  checked
+                    ? `border-blue-500 text-blue-500 ${
+                        focussed ? 'shadow-input-blue-focus' : ''
+                      }`
+                    : `border-grey-400 text-grey-400 ${
+                        focussed ? 'shadow-input-grey-focus' : ''
+                      }`
+                }`
```
LZoog, line 112: Are `focussed` and `hovered` states used just for classes, specifically in the checkbox component? Can we just use Tailwind's `hover:whatever` etc. classes instead? Even `:hover:not(disabled)` as custom CSS seems a little easier to follow than rerendering the component on hover/focus change for a class update.

packages/fxa-react/components/DataBlock/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <>
+      <button
+        className={`flex rounded-xl px-7 font-mono text-sm text-green-900 bg-green-800 bg-opacity-10 flex-wrap ${
+          valueIsArray ? 'py-4' : 'py-5'
+        } ${
+          copyable
+            ? 'hover:bg-opacity-20 focus:bg-opacity-30 active:bg-opacity-30'
+            : 'cursor-default select-text'
+        }`}
+        aria-describedby={readerText ? readerId : undefined}
```
LZoog, line 88: I'm pretty sure this can't be `readerText`, and I'm not sure if we need this. `aria-describedby` should be the ID associated. An `aria-label` will read text if we need to do that instead, but you can have `aria-label`, `aria-describedby`, _and_ `aria-labelledby` if we need it. I think `aria-describedby` and `aria-labelledby` give enough context here because then the label will be read in addition. What use case do you think we may have?

packages/fxa-react/components/TextArea/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export type TextAreaProps = {
+  defaultValue?: string | number;
+  disabled?: boolean;
+  id?: string;
+  label: string;
+  onBlur?: (event: FocusEvent<HTMLTextAreaElement>) => void;
+  onChange?: (event: ChangeEvent<HTMLTextAreaElement>) => void;
+  onFocus?: (event: FocusEvent<HTMLTextAreaElement>) => void;
+  onKeyPress?: (event: KeyboardEvent<HTMLTextAreaElement>) => void;
```
LZoog, line 24: Why do we need `onKeyPress` here?

packages/fxa-react/components/Checkbox/checkmark.svg
```diff
@@ -0,0 +1 @@
+<svg height="12" viewBox="0 0 12 12" width="12" xmlns="http://www.w3.org/2000/svg"><path d="m3.98771 11.9923c-.26519-.0001-.51951-.1055-.707-.293l-2.999998-3.00003c-.3789722-.39238-.3735524-1.01608.012181-1.40182.385734-.38573 1.009437-.39115 1.401817-.01218l2.157 2.157 6.31599-9.022997c.3186-.4465571.9375-.553074 1.3871-.23873s.562.932187.2519 1.384727l-6.99999 10.00003c-.16865.2438-.43677.4002-.732.427-.02898.0014-.05802.0014-.087 0z" /></svg>
```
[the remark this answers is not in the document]
jodyheavener: I don't have a problem with this, but also, why?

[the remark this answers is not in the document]
LZoog: It's probably not very maintainable since SVGs are usually are given to us with a width/height. It's so that if they're to be reused we can pass in those values (and be explicit about it in the component to know what size it's going to render at) but I think this could also serve as a default if passing them later just overwrites them.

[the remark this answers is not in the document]
jodyheavener: AFAICT it just overwrites it. But I'm fine either way. Will update.

packages/fxa-react/components/PasswordInput/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      onChange={inputChange}
+      {...{
+        defaultValue,
+        disabled,
+        id,
+        label,
+        onKeyPress,
+        placeholder,
+      }}
+      extras={
```
LZoog, line 65: It could be appropriate to pass this in as `children` instead of an `extras` prop if it's always going to be React elements.

packages/fxa-react/components/Checkbox/index.stories.tsx
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jodyheavener: Per our Slack chat, I disagree and think this is generic enough that it could belong in fxa-react for future re-use. But I also do not mind it in fxa-settings. Let's stick it there for now.

packages/fxa-react/components/DataBlock/index.tsx
jodyheavener: > Is this an optional callback for use later?

Yes. We will likely be adding a tooltip or some other functionality so I want to expose this for future use.

packages/fxa-react/components/TextArea/index.tsx
jodyheavener: Per our Slack chat, I wanted to future-proof some of the more common uses of something like a text input field. Something like listening to key strokes is pretty common. No, we don't need it right now. My preference for this is because I don't like having to introduce foundational, component-wide changes before a specific feature that utilizes the component can be implemented as it could add more development/testing time and throws off the scope of a PR. However, I'm not willing to die on this hill so I'll remove all unnecessary event listeners across all these new components. We also don't need this TextArea altogether right now (it's not in the Figma doc), so I'll remove it.

packages/fxa-react/components/DataBlock/index.tsx
LZoog: Edit: oops I looked at this incorrectly, it's only ever `readerId` or `undefined`, so ignore the first sentence.

packages/fxa-react/components/Checkbox/index.tsx
jodyheavener: So, in this specific case, the "focussed" state would only be applied if the preceding "checked" evaluated to true, and that only happens when the real input is checked (since this is a faux checkbox and cannot be checked nor focussed natively). But I think there are some other areas where this can be improved, so good catch. I will eliminate non-CSS state changes where possible.

packages/fxa-react/components/DataBlock/index.tsx
jodyheavener: Slight confusion per our Slack chat, BUT, what if I remove all the describedby stuff and just allowed all the components to receive it as a prop, leaving it up to the dev in each specific use case if they want to add additional input descriptors?

Issue #4849, opened 2020-04-10
[E2E] Decide on E2E testing library


data-sync-user:
Unit tests will, of course, be written in tandem with the creation of new components for this project. However, we're going to want some sort of E2E testing setup with Selenium like we do now in the `fxa-content-server`.

This might be a good opportunity to reflect on pain points using our current functional testing stack. Does theintern/leadfoot offer enough support or should we also explore conveniences NightmareJS, Playwright, Puppeteer, WebDriverIO, or another library offers? (FWIW, theintern/leadfoot has been deprecated in favor of just theintern/intern.)

At the very least, let's ensure we're using async/await syntax with these tests.

This `needs:discussion`. If we decide to use another framework, we'll want an ADR written to document why this decision was made. If we go with what we currently have, this issue can be closed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1557)
┆Issue Number: FXA-1557

LZoog:
@rbillings @ashrivastava-qa Tagging you here for insight, discussion, and/or comments.

data-sync-user:
➤ Lauren Zugai commented:

Notes from a 4/16 meeting involving [~vbudhram@mozilla.com], [~jheavener@mozilla.com], [~rbillings@mozilla.com], [~ashrivastava@mozilla.com], and myself:

At the very least, we'll want to copy current tests, adjust them as needed, remove outdated tests, and use async/await. However, it does sound like it's worth identifying the painpoints of the current framework and outlining pros & cons of using another framework.

Pros of keeping theintern:

  * Current framework is straightforward by having helper functions and examples readily available
  * Might be more copy-and-pastable in new Settings E2E testing than starting from scratch (but would still need refactoring)

Cons:

  * Still requires a learning curve to understand how things work and are structured.
  * It's kind of clunky, would be a lot of effort to go refactor all tests to async/await and could be better to just rewrite them with another library
  * It's an older framework, a newer framework may offer greater support and features
  * Our tests are a little flaky, can another framework help mitigate false failures from CI?

Things to look into:

  * What are the learning curves on newer frameworks?
  * What features will we gain? I (Lauren) have had one issue with getting theintern to "tab" to something and verify focus changed.

Other notes

  * Playwright’s big claim to fame is testing in a variety of browsers. Is this a priority of FxA's?
[...]

PR #6210, opened 2020-08-17
chore(docs): Add E2E testing ADR for settings V2

vbudhram:
## Because

- We should document these decisions for future us

## This pull request

- Documents why we choose to use InternJS in the new settings V2 app

Closes: #4849

## Checklist

### DIFF
--- a/docs/adr/0021-use-internjs-testing.md
+++ b/docs/adr/0021-use-internjs-testing.md
@@ -1,0 +1,66 @@
+# Use existing InternJS for functional testing in settings v2
+
+- Deciders: Vijay Budhram, Jody Heavener, Lauren Zugai
+- Date: 2020-07-17
+
+## Context and Problem Statement
+
+For over the past 5 years (at least), we have been using [InternJS](https://github.com/theintern/intern) for our functional tests in the content-server. While this has worked well and caught several bugs, we should reevaluate it since we are migrating to react and have an opportunityto use newer/faster/better supported testing frameworks.
+
+## Decision Drivers
+
+- Ease of writing tests
+- Good community support
+- Learning curve of settings V2 development
+
+## Decision Outcome
+
+Use InternJS for functional testing in the settings v2 app, but also update the tests to support async/await. The biggest factor was not having to introduce another testing framework for developers to learn.
+
+## Considered Options
+
+- Option A - Use InternJS and update to async/await
+- Option B - Use [NightmareJS]()
+- Option C - Use [WebDriverIO](https://webdriver.io/)
+
+## Pros and Cons of the Options
+
+### Option A - Use InternJS and update to async/await
+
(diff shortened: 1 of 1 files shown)


docs/adr/0020-use-internjs-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Decision Drivers
+
+- Ease of writing tests
+- Good community support
+- Speed of tests
+
+## Decision Outcome
+
+Use InternJS for functional testing in the settings v2 app, but also update the tests to support async/await. The biggest
```
vbudhram, line 20: I still think our decision to stay with InternJS is the right call at this stage since it does reduce our overall risks...but dang WebDriverIO seems cool

docs/adr/0020-use-internjs-testing.md
```diff
@@ -0,0 +1,69 @@
...
+factor was not having to introduce another testing framework for developers to learn.
+
+## Considered Options
+
+- Option A - Use InternJS and update to async/await
+- Option B - Use [NightmareJS]()
```
LZoog, line 26: I'd be fine with option B simply being "use a different framework like WebDriverIO," I don't think we need to dig into the nitty gritty since we're going with option A.

docs/adr/0020-use-internjs-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+For over the past 5 years (at least), we have been using [InternJS](https://github.com/theintern/intern) for our functional tests in the content-server. While
+this has worked well and caught several bugs, we should reevaluate it since we are migrating to react and have an opportunity
+to use newer/faster/better supported testing frameworks.
+
+## Decision Drivers
+
+- Ease of writing tests
+- Good community support
+- Speed of tests
```
LZoog, line 16: I think we could add something here around the Settings learning curve - since we're already introducing new tech and conventions there, it makes sense to keep some things the same to mitigate that ramp up getting even steeper, as you note as a pro of keeping it. Maybe this just generally falls under "Ease of writing tests" 🤷

docs/adr/0020-use-internjs-testing.md
```diff
@@ -0,0 +1,69 @@
+# Use existing InternJS for functional testing in settings v2
+
+- Deciders: Vijay Budhram, Jody Heavener, Lauren Zugai
+- Date: 2020-07-17
+
+## Context and Problem Statement
+
+For over the past 5 years (at least), we have been using [InternJS](https://github.com/theintern/intern) for our functional tests in the content-server. While
+this has worked well and caught several bugs, we should reevaluate it since we are migrating to react and have an opportunity
```
LZoog, line 9: Do we have Prettier edit our MD files? There's a few weird "go to next line" spacing

PR #6211, opened 2020-08-17
feat(graphql): add new settings mutations

### LINKED ISSUE #6204: Add 2 session-related GQL mutation resolvers
We need a couple of GQL mutation resolvers (bolded) for Settings:

* **`sendSessionVerificationCode`** should send the user a verification code to their primary e-mail that will be sent up to the
* **`verifySession`** mutation which will accept the code and verify the session

If `fxa-client` already supports these, this task should be quick. If these aren't normally done with a sessionToken auth header, there might be some auth-server modifications needed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2425)
┆Issue Number: FXA-2425

dannycoates:
This adds the `verifySecondaryEmail` `sendSessionVerificationCode` and `verifySession` mutations to support new settings page.

closes #6204

### DIFF
--- a/packages/fxa-graphql-api/src/lib/resolvers/account-resolver.ts
+++ b/packages/fxa-graphql-api/src/lib/resolvers/account-resolver.ts
@@ -29,9 +29,12 @@ import {
   CreateTotpInput,
   DeleteTotpInput,
   EmailInput,
+  SendSessionVerificationInput,
   UpdateAvatarInput,
   UpdateDisplayNameInput,
   VerifyTotpInput,
+  VerifyEmailInput,
+  VerifySessionInput,
 } from './types/input';
 import {
   BasicPayload,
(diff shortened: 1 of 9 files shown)


packages/fxa-graphql-api/src/lib/resolvers/account-resolver.ts
```diff
@@ -162,41 +165,56 @@ export class AccountResolver {
   }
 
   @Mutation((returns) => BasicPayload, {
-    description: 'Remove the secondary email for the signed in account.',
+    description: 'Reset the verification code to a secondary email.',
   })
   @CatchGatewayError
-  public async deleteSecondaryEmail(
+  public async resendSecondaryEmailCode(
```
dannycoates, line 171: git really mangled the diff here. I just moved `deleteSecondaryEmail` down

PR #6212, opened 2020-08-17
refactor(settings): a few adjustments to FlowContainer

dannycoates:
Make the back arrow navigate back in history.
A couple style adjustments.

### DIFF
--- a/packages/fxa-settings/src/components/FlowContainer/index.tsx
+++ b/packages/fxa-settings/src/components/FlowContainer/index.tsx
@@ -3,35 +3,35 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-import { RouteComponentProps, useLocation, Link } from '@reach/router';
+import { RouteComponentProps } from '@reach/router';
 import { ReactComponent as BackArrow } from './back-arrow.svg';
 
 type FlowContainerProps = {
   title?: string;
-  backRoute?: string;
-  children?: React.ReactElement;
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/components/FlowContainer/index.tsx
```diff
@@ -3,35 +3,35 @@
...
+      className="mx-2 tablet:mt-12 tablet:mx-24 desktop:mx-56"
+      data-testid="flow-container"
+    >
+      <div className="flex items-start bg-white shadow py-8 px-6 rounded-xl">
+        <button
+          onClick={() => window.history.back()}
+          className="pr-6 desktop:pl-6 desktop:-ml-24"
+          data-testid="flow-container-back-btn"
+        >
+          <BackArrow role="img" className="inline h-8 w-8" />
```
LZoog, line 29: I know this isn't part of your PR but just because the line was touched - a pattern we've been following per Meridel is if there's an SVG inside of a button, we're adding a `title` to the button (like `title="Back"`), then we don't need `role="img"` or an `aria-label` on the button because "Back" is read. With just a `role="img"` here it'll be read as "image" and no other context.

dannycoates:
Thanks @LZoog, I added another commit on that changes where some flexboxes are because Nav and <main> in AppLayout were having some undesirable effects on the FlowContainer contents.

vbudhram:
@dannycoates Merging this it looks ready and I can get the latest change.. I can add any missing bits in mine.

PR #6213, opened 2020-08-17
refactor(settings): ui component followups

jodyheavener:
## Because

- There were still a few outstanding items to address as part of #6206

## This pull request

- Eliminates all component `readerText` props. We determined they didn't add a substantial amount of value, and any accompanying a11y content can be implemented on a case-by-case basis.
- As a result of removing readerText I also refactored the the usage of component IDs and was able to eliminate the UUID dependency.
- Makes `DataBlock` only copyable. We currently don't have a need for a read-only version.
- Further cleans up some leftover event handlers and weird syntax in the new components.

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/DataBlock/index.tsx
+++ b/packages/fxa-settings/src/components/DataBlock/index.tsx
@@ -2,37 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import React, {
-  useCallback,
-  useState,
-  useEffect,
-  MouseEvent,
-  FocusEvent,
-} from 'react';
-import { v4 as uuidv4 } from 'uuid';
+import React from 'react';
 import { copy } from '../../lib/clipboard';
 
 export type DataBlockProps = {
   value: string | string[];
-  copyable?: boolean;
-  id?: string;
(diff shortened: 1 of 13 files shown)


packages/fxa-settings/src/components/DataBlock/index.tsx
```diff
@@ -43,46 +22,24 @@ export const DataBlock = ({
...
-          <span>{value}</span>
-        )}
-      </button>
-      {readerText && (
-        <span data-testid="datablock-srtext" className="sr-only" id={readerId}>
-          {readerText}
-        </span>
+    <button
+      className={`flex rounded-xl px-7 font-mono text-sm text-green-900 bg-green-800 bg-opacity-10 flex-wrap hover:bg-opacity-20 focus:bg-opacity-30 active:bg-opacity-30 ${
+        valueIsArray ? 'py-4' : 'py-5'
```
LZoog, line 28: I think some of what what was done in this PR/the other could look a bit cleaner using `classNames` lib, but this works👍

packages/fxa-settings/src/components/TextInput/index.tsx
```diff
@@ -68,15 +43,14 @@ export const TextInput = ({
   );
 
   return (
-    <div
+    <label
```
LZoog, line 46: Oh nice catch :P

Issue #6041, opened 2020-07-24
[auth-server][stripe] Remove old Source/Payment Method when setting new default payment method


data-sync-user:
We allow customers to update their payment information, but we should also remove the previous Source (pre-SCA customers) or Payment Method.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2348)
┆Issue Number: FXA-2348

6a68:
Closed by #6232.

PR #6232, opened 2020-08-19
feat(auth-server): remove sources on payment update

### LINKED ISSUE #2348: Extract 2FA QR code screen into a mixin
One step along the way to enabling 2FA as part of the login flow. Refs #2024

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-308)
┆Issue Number: FXA-308

bbangert:
Because:

* When updating a customer with a payment method, we don't want any
  sources from the Sources API around.

This commit:

* Removes all sources from a customer after updating the default
  payment method with a payment method id.

Closes #2348

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -299,6 +299,34 @@ export class StripeHelper {
     });
   }
 
+  /**
+   * Remove all sources from a customer.
+   *
+   * For users that are using payment methods, we no longer wish to store
+   * sources so we remove them all.
+   *
+   * Returns the deleted cards.
+   *
+   * @param customerId
+   */
+  async removeSources(customerId: string): Promise<Stripe.Card[]> {
+    const sources = await this.stripe.customers.listSources(customerId, {
+      object: 'card',
+    });
+    if (sources.data.length === 0) {
+      return [];
+    }
+    return Promise.all(
+      sources.data.map(
+        (s) =>
+          (this.stripe.customers.deleteSource(
+            customerId,
+            s.id
+          ) as unknown) as Promise<Stripe.Card>
+      )
+    );
+  }
+
   /** END: NEW FLOW HELPERS FOR PAYMENT METHODS **/
 
   /**
--- a/packages/fxa-auth-server/test/local/payments/stripe.js
+++ b/packages/fxa-auth-server/test/local/payments/stripe.js
@@ -243,6 +243,49 @@ describe('StripeHelper', () => {
     });
   });
 
+  describe('removeSources', () => {
+    it('removes all the sources', async () => {
+      const ids = {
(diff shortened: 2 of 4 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -299,6 +299,34 @@ export class StripeHelper {
...
+    if (sources.data.length === 0) {
+      return [];
+    }
+    return Promise.all(
+      sources.data.map(
+        (s) =>
+          (this.stripe.customers.deleteSource(
+            customerId,
+            s.id
+          ) as unknown) as Stripe.Card
```
chenba, line 325: I think this should be `as Promise<Stripe.Card>` as the type of `this.stripe.customers.deleteSource` is `Promise<T>` (where `T` is a sum type of types in `Stripe`, one of which is `Stripe.Card`).

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -243,6 +243,32 @@ describe('StripeHelper', () => {
...
+    it('removes all the sources', async () => {
+      const ids = {
+        data: [{ id: uuidv4() }, { id: uuidv4() }, { id: uuidv4() }],
+      };
+      sandbox.stub(stripeHelper.stripe.customers, 'listSources').resolves(ids);
+      sandbox.stub(stripeHelper.stripe.customers, 'deleteSource').resolves({});
+      const result = await stripeHelper.removeSources('cust_new');
+      assert.deepEqual(result, [{}, {}, {}]);
+      sinon.assert.calledThrice(stripeHelper.stripe.customers.deleteSource);
+    });
```
chenba, line 256: nit-ish:
 - the test above doesn't check to see if the ids passed into `deleteSource` are the ones from the test.
 - the empty sources list branch of the code in `removeSources` isn't tested.

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -936,10 +936,16 @@ describe('DirectStripeRoutes', () => {
       directStripeRoutesInstance.stripeHelper.customer
         .onCall(1)
         .resolves(expected);
-
+      directStripeRoutesInstance.stripeHelper.customer.resolves({ data: [] });
```
chenba, line 939: Why is this necessary?  I don't see an additional call to `stripeHelper.customer` above.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Yup, updated.

6a68:
I'm thinking this was intended to close Jira bug FXA-2348, which means this PR closes #6041 on github :+1:

PR #6242, opened 2020-08-19
refactor(settings): Cleanup secondary email tests, catch mutation errors

### LINKED ISSUE #6192: Get secondary email mutation failure test passing & DRY up tests
https://github.com/mozilla/fxa/pull/6187#discussion_r469941324 pointed out the tests for secondary email could be DRY'd up a bit. I think we could move some other test helpers in another file as well (or maybe they'll just live in `_mocks`).

Additionally, we need to establish a way to test error states in mutation tests. Fix this TODO: https://github.com/mozilla/fxa/pull/6187/files#diff-ef3ddcb1f93137206feed27e81bccc67R337-R341

EDIT: this task originally also included adding docs around Context/Providers and tests, the alert bar, external alert, and maybe an expansion on the client cache, but that was split to  FXA-2445.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2420)
┆Issue Number: FXA-2420

LZoog:
Because:
* We setup many 'email' objects and need a helper for DRYer code
* We must have an onError option in mutations to allow tests to pass

This commit:
* Gets tests passing and DRYs them up

fixes #6192

## Checklist

### Additional information

I was going to wait to put this PR up because the secondary email section needs a refactor to change the `primaryEmail` verified state check to guard against a _session_ being unverified instead but I'll do that in a follow up since these changes are needed in another PR.

### DIFF
--- a/packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.stories.tsx
+++ b/packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.stories.tsx
@@ -8,12 +8,12 @@ import { InMemoryCache } from '@apollo/client';
 import { MockedProvider } from '@apollo/client/testing';
 import { UnitRowSecondaryEmail, RESEND_SECONDARY_EMAIL_CODE_MUTATION } from '.';
 import { AlertBarRootAndContextProvider } from '../../lib/AlertBarContext';
-import { MockedCache, MOCK_ACCOUNT } from '../../models/_mocks';
+import { MockedCache, MOCK_ACCOUNT, mockEmail } from '../../models/_mocks';
 import { GET_INITIAL_STATE } from '../App';
 
 // every unverified email with a functioning "Resend verification"
 // button must have a mock object created per mutation attempt.
-const createMock = (email: string) => ({
+const mockGqlSuccess = (email: string) => ({
   request: {
     query: RESEND_SECONDARY_EMAIL_CODE_MUTATION,
     variables: { input: { email } },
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.stories.tsx
```diff
@@ -173,26 +121,10 @@ storiesOf('Components|UnitRowSecondaryEmail', module)
...
-        isPrimary: false,
-        verified: false,
-      },
-      {
-        email: 'johndope4@example.com',
-        isPrimary: false,
-        verified: false,
-      },
+      mockEmail('johndope@example.com'),
+      mockEmail('johndope2@example.com', false, false),
```
meandavejustice, line 125: It looks like this one should be `mockEmail('johndope2@example.com', false),` based on the code it is replacing

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
```diff
@@ -36,8 +37,18 @@ export const UnitRowSecondaryEmail = () => {
 
   const [alertBarRevealed, revealAlertBar, hideAlertBar] = useBooleanState();
 
+  // TODO: DRY this up and don't import `sentryMetrics` into every component
+  // needing a mutation - we must have an `onError` option in mutations
+  // to allow tests to pass but providing one prevents an error from actually
+  // throwing so we need to manually report it to Sentry.
+  // See https://github.com/apollographql/react-apollo/issues/2614
```
meandavejustice, line 44: Is there an issue filed for this follow up?

LZoog: I'll do this now, there's one other thing I need to file as well (noted in the "other information").

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.stories.tsx
LZoog: Ah thanks for catching, I'll double check the others in Storybook are correct.

packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx
LZoog: Follow up for this is: FXA-2450
Follow up for what I've got in "other information" is: FXA-2449

PR #6246, opened 2020-08-19
fix(totp): Ensure the user has a verified session before starting inline TOTP setup

6a68:
## Because

- A verified session is required to enable 2FA.
- Existing flows do this from within the settings page, where we have the session upgrade mixin (this mixin does DOM twiddling to put an upgrade template in place of the regular template, and doesn't easily generalize to an inline flow)
- We do have a screen to do something similar inline, where users are asked to verify a code sent in an email

## This pull request

- Updates the inline totp setup view to redirect users to the signin_token_code screen if their session is unverified. Users will be redirected back to inline totp setup when that step is completed.

Closes: #5206

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-content-server/tests/functional/oauth_sign_in_token_code.js
+++ b/packages/fxa-content-server/tests/functional/oauth_sign_in_token_code.js
@@ -117,27 +117,5 @@ registerSuite('OAuth signin token code', {
           )
       );
     },
-
-    'verified - valid code then click back': function () {
-      return (
-        this.remote
-          .then(openFxaFromRp('enter-email', experimentParams))
-          .then(fillOutEmailFirstSignIn(email, PASSWORD))
-
-          // Correctly submits the token code and navigates to oauth page
-          .then(testElementExists(selectors.SIGNIN_TOKEN_CODE.HEADER))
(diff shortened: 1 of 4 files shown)


lmorchard:
I haven't fully dug into this PR, but the content-server failures don't look entirely unrelated?

* firefox on linux 4.15.0-1077-aws - OAuth signin token code - verified - valid code then click back
* firefox on linux 4.15.0-1077-aws - signup with code - invalid code

packages/fxa-content-server/tests/functional/oauth_sign_in_token_code.js
```diff
@@ -117,27 +117,5 @@ registerSuite('OAuth signin token code', {
           )
       );
     },
-
-    'verified - valid code then click back': function () {
```
6a68, line 121: talked through this test with @vbudhram and neither one of us quite understood its value. ripped it out 👷‍♂️

lmorchard: Cool, looks like that resolved one of the failures and the other didn't happen again

6a68:
Thanks for the reviews! Merging.

Issue #6227, opened 2020-08-18
WError: An internal validation check failed.


clouserw:
https://sentry.prod.mozaws.net/operations/auth-stage/issues/9391817/?referrer=github_plugin

```
WError: An internal validation check failed.
  File "/fxa/packages/fxa-auth-server/lib/error.js", line 1134, in Function.AppError.internalValidationError
    return new AppError({
  File "/fxa/packages/fxa-auth-server/lib/payments/stripe.js", line 814, in StripeHelper.extractInvoiceDetailsForEmail
    throw error_1.default.internalValidationError('extractInvoiceDetailsForEmail', { customerId: customer.id }, 'Customer missing email.');
  ?, in runMicrotasks
  File "/fxa/packages/fxa-auth-server/lib/routes/subscriptions.js", line 715, in async DirectStripeRoutes.sendSubscriptionInvoiceEmail
    const invoiceDetails = await this.stripeHelper.extractInvoiceDetailsForEmail(invoice);
  File "/fxa/packages/fxa-auth-server/lib/routes/subscriptions.js", line 661, in async DirectStripeRoutes.handleInvoicePaymentSucceededEvent
    const { uid, email } = await this.sendSubscriptionInvoiceEmail(invoice);
...
(7 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2438)
┆Issue Number: FXA-2438

data-sync-user:
➤ Benjamin Bangert commented:

Due to a lack of context, we're unable to proceed on this at the moment. I've moved this to the next sprint, as a PR did go in that will provide the context needed.

PR #6247, opened 2020-08-19
fix(auth-server): attach payload data to sentry

bbangert:
Because:

* We can't see the data that is invalid for internal validation errors
  in Sentry, as the error reports are missing the extra attached data.

This commit:

* Add's payload and payload extra data to the Sentry context.

Issue #6227

## Checklist


clouserw:
**[everyone liked this]**

PR #6251, opened 2020-08-19
chore(configs): remove basket proxy configs

chenba:
Because:
 - the need for basket proxy was removed in f2d2a2d

This commit:
 - delete some configs that are no longer used

## Checklist

### DIFF
--- a/packages/fxa-content-server/server/lib/csp/blocking.js
+++ b/packages/fxa-content-server/server/lib/csp/blocking.js
@@ -19,15 +19,12 @@ function getOrigin(link) {
  * blockingCspMiddleware is where to declare rules that will cause a resource
  * to be blocked if it runs afowl of a rule.
  */
-module.exports = function(config) {
+module.exports = function (config) {
   const AUTH_SERVER = getOrigin(config.get('fxaccount_url'));
   const BLOB = 'blob:';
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/server/lib/csp/blocking.js
```diff
@@ -38,10 +35,9 @@ module.exports = function(config) {
   const PAIRING_SERVER_HTTP = PAIRING_SERVER_WEBSOCKET.replace(/^ws/, 'http');
   const SENTRY_SERVER = 'https://sentry.prod.mozaws.net';
   // create a unique array of origins from survey urls
-  const SURVEYS = [
-    ...new Set(surveyList.map(s => getOrigin(s.url))),
-  ];
-  const surveysEnabledAndSet = (config.get('surveyFeature.enabled') && SURVEYS.length);
+  const SURVEYS = [...new Set(surveyList.map((s) => getOrigin(s.url)))];
+  const surveysEnabledAndSet =
+    config.get('surveyFeature.enabled') && SURVEYS.length;
```
chenba, line 40: Argh how do these formatting changes even sneak in?!  I `git add -p` and checked the staged changes and everything.  😖

Issue #6225, opened 2020-08-18
Content on mobile should take up full width of container


data-sync-user:
The mobile layout for Settings has the back arrow taking up a column. This should be fixed so that content after the header takes up the full width of the container. See screenshot.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2437)
┆Issue Number: FXA-2437
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/13796/Group+1.png">Group 1.png</a>

data-sync-user:
➤ Jody Heavener commented:

Also, that arrow is too big. Refer to the Figma doc.

PR #6252, opened 2020-08-20
A few lil settings tweaks here n there

jodyheavener:
## Because

There were a few code quality things and visual tweaks I wanted to address.

## This pull request

- Removes the CLI and visual prompt to upgrade Storybook
- Removes all `console.log`s (replacing them with TODOs). I would like to make an effort not to commit console calls unless they are intentional to aid with testing or provide broadly useful information when an error occurs. Developers should still consider logging errors to Sentry as well.
- Removes global console method stubbing. I added this not long ago and have already been bitten by it, and others have asked about it. If you're testing console methods you can stub/spy on a per-test basis. It's just way too easy to miss console errors or get confused why your temporary console.log isn't showing up.
- Update the new page layout in a few ways:
  - Rearranges content to take up full width of page container on mobile
  - Updates the back button to be a little smaller, on par with the Figma design
  - Adds a max-width to the page container; it's a little larger than the Figma design, but it matches our current Modal max-width
- Fixes a few issues where incorrect dependencies were being supplied to React Hooks
- Disables some ES Lint warnings that we either can't or won't deal with right now (feel free to debate me on these)
- Adds a shadow to the main header when you scroll down (feel free to reject this, just thought it looks nice)

Closes: #6225

## Checklist


packages/fxa-settings/src/components/ChangePassword/index.tsx
```diff
@@ -8,6 +8,7 @@ import FlowContainer from '../FlowContainer';
 import PasswordInput from '../PasswordInput';
 import LinkExternal from 'fxa-react/components/LinkExternal';
 
+// eslint-disable-next-line no-empty-pattern
 export const ChangePassword = ({}: RouteComponentProps) => {
```
jodyheavener, line 12: The object of type `RouteComponentProps` is apparently needed in order to match the function signature expected when calling this, but we don't use any of the properties from it so it violates the eslint rule 😒

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -12,8 +13,27 @@ import { ReactComponent as Bento } from './bento.svg';
 import { ReactComponent as Menu } from './menu.svg';
 import { useAccount } from '../../models';
 
+const SHADOW_SCROLL_THRESHOLD = 5;
```
jodyheavener, line 16: 5px is arbitrary

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -12,8 +13,27 @@ import { ReactComponent as Bento } from './bento.svg';
...
-  const { avatarUrl, primaryEmail } = useAccount();
+  const { avatarUrl } = useAccount();
+  const [showShadow, setShowShadow] = useState<boolean>(false);
+
+  useEffect(() => {
+    function handleScroll() {
+      const scrollTop =
+        window.pageYOffset || document.documentElement.scrollTop;
+      const shouldShow = scrollTop > SHADOW_SCROLL_THRESHOLD;
+      if (shouldShow !== showShadow) {
```
jodyheavener, line 27: don't bother updating state (and triggering rerender) if it's the same value

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -269,6 +269,7 @@ export function useViewEvent(
 ) {
   useEffect(() => {
     logViewEvent(viewName, eventName, eventProperties);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, dependencies);
```
jodyheavener, line 273: The warning that prompted me to add this:

> React Hook useEffect was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies

I actually don't understand this warning, because if you look a little further up the default value for this _is_ an array literal...

packages/fxa-settings/src/setupTests.js
```diff
@@ -3,7 +3,3 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import '@testing-library/jest-dom/extend-expect';
-
-window.console.error = jest.fn();
-window.console.log = jest.fn();
-window.console.warn = jest.fn();
```
jodyheavener, line 9: pobody's nerfect

packages/fxa-settings/src/components/ChangePassword/index.tsx
vbudhram: They used this in the react reach documentation, but maybe there is a better way to do this? My typescript foo still WIP :/

packages/fxa-settings/src/lib/metrics.ts
dannycoates: it's an array variable but not a literal. the warning is that it can't statically analyze the values of `dependencies`

PR #6259, opened 2020-08-20
refactor(settings): small updates to settings header shadow

jodyheavener:
Followup to #6224 

- Throttles scroll event listener (Note, because of this you now can easily scroll to the top and not have the shadow disappear, which is a little disappointing
- Uses `shadow-sm` instead of `shadow-md` - IMO this is barely noticeable, but we can always come back in design review to update if it's needed


jodyheavener:
Ooh @LZoog if we also replaced `bg-grey-10` with `bg-white` when scrolling this looks much more effective than just shadow:

Screen Shot 2020-08-20 at 11 59 37 AM

Or, it doesnt need to be a shadow at all:

Screen Shot 2020-08-20 at 12 00 19 PM

jodyheavener:
I feel like I like the border over the shadow 🤔

LZoog:
> Ooh @LZoog if we also replaced bg-grey-10 with bg-white when scrolling this looks much more effective than just shadow:

@jodyheavener can we send a screenshot to Phil / Jess to get their opinion? That'll solve the shadow/bg thing. :P

packages/fxa-settings/src/components/HeaderLockup/index.tsx
```diff
@@ -29,9 +30,13 @@ export const HeaderLockup = () => {
       }
     }
 
-    window.addEventListener('scroll', handleScroll);
+    const throttledHandler = throttle(() => {
+      handleScroll();
+    }, 100);
```
LZoog, line 35: > Note, because of this you now can easily scroll to the top and not have the shadow disappear, which is a little disappointing

Heh, we actually need a debounce to fix this. I've worked on an app where we had to throttle _and_ debounce, might need something similar here.

jodyheavener:
Updated this PR to remove it! :)

PR #6271, opened 2020-08-21
fix(content-server): fix padding in link from legal footer

### LINKED ISSUE #6235: Different focus borders for “Terms of Service” and “Privacy Notice” links from Settings page
****Build**
Prod [Train 1.184.0]

**Affected platforms**
Windows 10 x64

**Steps to reproduce**
1. Login using a Firefox account and navigate to https://accounts.firefox.com/settings
2. Open the “Terms of Service” and “Privacy Notice” links in new tabs. 
3. Observe the “Terms of Service” and “Privacy Notice” focus borders. 

**Expected results**
Both links have the same focus border shape.

**Actual Results**
 “Terms of Service” and “Privacy Notice” have different focus borders shapes.

Screenshot_44
Screenshot_45

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2443)
┆Issue Number: FXA-2443

byrw:
Because:

* Links from `#legal-footer` have different focus borders shapes.

This commit:

* FIx the padding in the last link from `#legal-footer`.

Closes #6235

### DIFF
--- a/packages/fxa-content-server/app/styles/modules/_legal.scss
+++ b/packages/fxa-content-server/app/styles/modules/_legal.scss
@@ -22,8 +22,8 @@
     a {
       padding: 0 16px;
 
-      &:last-child {
-        padding: 16px 0 16px 16px;
+      &.privacy {
+        margin-right: -16px;
       }
     }
   }


packages/fxa-content-server/app/styles/modules/_legal.scss
```diff
@@ -21,10 +21,6 @@
 
     a {
       padding: 0 16px;
-
-      &:last-child {
-        padding: 16px 0 16px 16px;
```
jodyheavener, line 26: It's unclear to me why this was styled like this to begin with, as it looks somewhat specific. That being said, I think it's fine to remove this (the vertical padding), but the horizontal padding pushes the item in 16px on the right. Just so we can maintain it flush against the edge can we also set `margin-right: -16px;` using this selector?

byrw: Good catch, I did not notice the horizontal padding.

jodyheavener: Hi @byrw I think we can actually forego the left margin on the privacy link, this seems to add excess space between the two items; I was referring to using a negative margin on the privacy link/last-child to offset it's 16 padding. Doing this will ensure it's inline with the "Sign out" button at the top, and the white container on smaller screens.

byrw: Sorry, I thought you mistyped.

Issue #6249, opened 2020-08-19
Set a convention for mutation error handling


data-sync-user:
We need to establish a pattern for error handling and implement it where we're already using the `useMutation` hook in Settings.

See [the UnitRowSecondaryEmail](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/UnitRowSecondaryEmail/index.tsx) component and the "TODO" listed above the `useMutation` reference:

```java
  // TODO: DRY this up and don't import `sentryMetrics` into every component
  // needing a mutation - we must have an `onError` option in mutations
  // to allow tests to pass but providing one prevents an error from actually
  // throwing so we need to manually report it to Sentry.
  // See https://github.com/apollographql/react-apollo/issues/2614
```

Without this `onError` catch, or wrapping the mutation call in a `try`/`catch`, tests fail, and we want to avoid the need to manually set this on every mutation and avoid pulling in `sentryMetrics` into every component using a mutation.

Also see [the UnitRowRecoveryKey](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx) and a short discussion in [https://github.com/mozilla/fxa/pull/6224/files#r474080213](https://github.com/mozilla/fxa/pull/6224/files#r474080213). Do we want to show all GraphQL errors (messaging from the server like "invalid code"), then show a generic message like, "There was a problem with your request. Please try again" for network etc. errors?

We _probably_ don't want to send GQL errors to Sentry, but send other errors. Part of this is confirming this, and working with @MeridelW on the "general" error messaging. If she would like a more specific message per component (such as, "There was a problem removing your recovery key. Please try again."), it's probably best to do that in a follow up towards the end of this project.

- - -
TLDR;

* Confirm we want to show GQL error messages to the user and not send them to Sentry, and for other errors show a general error message and send the error to Sentry
* Make our `useMutation` error handling consistent according to the above (should we have a helper to avoid importing `sentryMetrics` everywhere?), and document the preferred approach in the `fxa-settings` README
* Grab the "general error" text from Meridel and display it where we should. If Meridel would like more specific errors, file a follow up issue to do and review at a later time.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2450)
┆Issue Number: FXA-2450

Issue #6107, opened 2020-07-31
[main] Settings > Profile > Secondary email: remove + make primary buttons


data-sync-user:
If the secondary email is verified:

* The remove button/icon (trash can) should remove the secondary email - there’s a GQL mutation available called `deleteSecondaryEmail`. On success/fail, display the message to the user. The button should have a title of “remove” and make sure there’s adequate clickable space in mobile especially (check with design if the delete and “make primary” buttons feel too close).
* The “make primary” button should be present and should change the user’s primary email - there’s a GQL mutation available called `updatePrimaryEmail`.  On success/fail, display the message to the user.

The refresh button will be added in FXA-1798.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2372)
┆Issue Number: FXA-2372

data-sync-user:
➤ Dave Justice commented:

This should be able to be started once https://github.com/mozilla/fxa/pull/6187 ( https://github.com/mozilla/fxa/pull/6187 ) is merged

PR #6276, opened 2020-08-21
New settings: implement verify secondary email

dannycoates:
This adds the ui to add a secondary email and verify it with a code.

It adds the `/emails` and `emails/verify` routes. I used `/emails` in order to match the existing settings route.

PR #6283, opened 2020-08-24
feat(settings): introduce hook to handle errors in mutations

jodyheavener:
## Because

- We want to be able to automatically handle mutation errors without importing error handling libs each time `useMutation` is used.
- Documentation needed some syntax highlighting ;)

## This pull request

- Adds `useHandledMutation`. This wraps `useMutation` with the option to specify an onError handler, which can itself throw and send the error to Sentry. In all other cases onError is just a noop. Per #6249, the idea is that we'd like to do this by default instead of importaing the Sentry lib each time. Note this does not close that issue though, as it does not address conventions for a general error message and GQL vs. other error types.
- Adds doc syntax highlighting in a couple spots

This sets the stage for a few things I'm working on in #6107, and also touches on #6249.

## Checklist

### DIFF
--- a/packages/fxa-settings/src/lib/hooks.tsx
+++ b/packages/fxa-settings/src/lib/hooks.tsx
@@ -3,6 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { useEffect, useRef } from 'react';
+import sentryMetrics from 'fxa-shared/lib/sentry';
+import {
+  useMutation,
+  DocumentNode,
+  MutationHookOptions,
+  ApolloError,
+} from '@apollo/client';
 
 // Focus on the element that triggered some action after the first
 // argument changes from `false` to `true` unless a `triggerException`
@@ -19,7 +26,12 @@ export function useFocusOnTriggeringElementOnClose(
     if (revealed !== undefined) {
       prevRevealedRef.current = revealed;
     }
-    if (triggerElement.current && prevRevealed === true && revealed === false && !triggerException) {
+    if (
+      triggerElement.current &&
+      prevRevealed === true &&
+      revealed === false &&
+      !triggerException
+    ) {
       triggerElement.current.focus();
     }
   }, [revealed, triggerElement, prevRevealed, triggerException]);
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -50,3 +62,22 @@ export function useChangeFocusEffect() {
...
+export function useHandledMutation(
+  mutation: DocumentNode,
+  options: MutationHookOptions<any, Record<string, any>> | undefined = {}
+) {
+  return useMutation(
+    mutation,
+    Object.assign(options, {
+      onError: (error: ApolloError) => {
+        sentryMetrics.captureException(error);
+      },
```
dannycoates, line 80: We want to be able to handle the error at the call location as well in order to handle mutation specific errors. If there's an `onError` in `options` we should wrap it - probably in a try/catch and let the passed in handler either handle it or rethrow to here. Some errors don't belong in sentry so we shouldn't report unless it's unexpected.

packages/fxa-settings/src/models/Account.ts
```diff
@@ -74,3 +74,10 @@ export function useAccount() {
   });
   return data!.account;
 }
+
+export function useRefetchableAccount() {
+  const { data, refetch } = useQuery<{ account: Account }>(GET_ACCOUNT, {
+    fetchPolicy: 'cache-first',
+  });
+  return { account: data!.account, refetchAccount: refetch };
+}
```
dannycoates, line 83: I'm not sure this is necessary. Mutations can specify an `update` function or `refetchQueries` to update the cache. https://www.apollographql.com/docs/react/api/react/hooks/#options-2

I also discovered in #6276 that `useQuery` doesn't play nice with the router.

packages/fxa-settings/src/lib/hooks.tsx
jodyheavener: Makes sense.

packages/fxa-settings/src/models/Account.ts
jodyheavener: Oh nice. Will update. And I'll temporarily add [this](https://github.com/mozilla/fxa/pull/6276/files#diff-5dfa9e3125fdc91e2fae5233ce23fbbeR72) here as well, in case this is merged first.

packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -50,3 +62,31 @@ export function useChangeFocusEffect() {
...
+  options: MutationHookOptions<any, Record<string, any>> | undefined = {}
+) {
+  return useMutation(
+    mutation,
+    Object.assign(options, {
+      // Pass in options.onError to handle an error and
+      // optionally re-throw so it gets logged to Sentry
+      // By default errors are ignored
+      onError: (error: ApolloError) => {
+        if (options.onError) {
```
dannycoates, line 82: idk that ignore by default is what we want, maybe my prev comment was misleading... If no `onError` handler is passed I think we _do_ want to send to sentry & maybe console because in that case I think not passing in a handler is like, "I don't know of any errors that can happen here, lmk if there are".

jodyheavener: i can dig it

PR #6284, opened 2020-08-24
refactor(settings): make cta buttons more modular, uniform in size

jodyheavener:
Breakout from #6283

## Because

- We should have consistent button styles
  - Note the only visual change here is that I've update the type face to Metropolis, per the Figma doc. Everything else is just how classes are approached. My goal is to provide a consistent base class that can be layered on top of with sizing classes, whereas currently we have a different sized primary/neutral buttons

## This pull request

- Updates our CTA classes to use modular classes that supply the style and sizing.

## Checklist

### DIFF
--- a/packages/fxa-settings/src/styles/ctas.scss
+++ b/packages/fxa-settings/src/styles/ctas.scss
@@ -2,8 +2,13 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+.cta-neutral,
+.cta-primary {
+  @apply py-2 px-5 rounded w-full inline-block text-center border font-header box-border;
+}
+
 .cta-neutral {
-  @apply block w-full py-2 text-center mt-6 bg-grey-10 border border-grey-200 rounded;
+  @apply bg-grey-10 border-grey-200;
 
   &:hover {
     @apply border-grey-200 bg-grey-100 text-grey-400;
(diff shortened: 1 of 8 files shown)


packages/fxa-settings/src/styles/ctas.scss
```diff
@@ -14,32 +19,33 @@
...
-    @apply border-grey-400 bg-grey-100 text-grey-400;
+    @apply bg-blue-800 border-blue-800;
   }
 }
 
-.cta-primary {
-  @apply bg-blue-500 py-2 text-white border border-blue-600 rounded;
+.cta-lg {
+  @apply text-base;
+}
```
LZoog, line 36: Since `text-base` is default, do we even need this class?

I don't mind these changes, but I did notice we're missing a `:focus` state here. Would you like to add that perhaps? 😬 (not needed though)

jodyheavener: > we're missing a :focus state here

That's what the primary/neutral classes would be used for (which have focus states) -- one class for sizing, one for style. Ideally we could bake the "regular" size into both the style classes and only use cta-lg when we wanted them to be bigger, but because the regular size uses breakpoints and cta-lg doesn't it means we'd need to introduce breakpoints for it as well just to take selector priority, and it becomes a slippery slope from there. I hope that makes sense. I'm happy to demo if you'd like 😅

jodyheavener: Good catch for both of these :)

Issue #4869, opened 2020-04-10
[header/footer] Integrate the avatar drop-down


data-sync-user:
Show the user's primary email, avatar, and allow them to logout.

Create a `<DropDown>` (or <DoorHanger>?) component that we'll share with the Bento menu.

Make sure the button opening/closing has `aria-expanded="true"` or false to signal to screenreaders there is more content if they interact with the button. The `title` for the button should read "Firefox Account Menu".

See https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A792

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1575)
┆Issue Number: FXA-1575

Issue #5169, opened 2020-04-30
[l10n] Adjust styles to support rtl languages


data-sync-user:
Go through the new Settings page and flows and add rtl style overrides where necessary. Can we use https://github.com/RonMelkhior/tailwindcss-dir?

This should probably be done with multiple PRs, scoped to different flows or components.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1797)
┆Issue Number: FXA-1797

yehudab:
Mind if I give it a try?
Already found one adjustment needed for the photo upload:
![photo upload](https://user-images.githubusercontent.com/633345/95076493-9842f080-071a-11eb-8c17-85ae20473026.png)

My fix for the above is ready in this commit: https://github.com/yehudab/fxa/commit/1e907bd18f9b892b82b024fea89c7d3cd6334d8c

yehudab:
I see now that I was looking at the wrong path. Was checking `/settings` instead of `/beta/settings`
Anyway, this fix will probably be usefull for the new settings as well.
Will look now at `tailwindcss-dir`

yehudab:
Started the tailwindcss-dir integeration. Looks pretty straight forward, please more work is needed.
Initial commit available here: https://github.com/yehudab/fxa/commit/d2990ee3f879942b04c324c12e33354529f78394

Some samples:
Before:
![new-settings-rtl-before](https://user-images.githubusercontent.com/633345/95121459-bd098900-0757-11eb-8599-39af2e21eab9.png)

After:
![new-settings-rtl-after](https://user-images.githubusercontent.com/633345/95121256-76b42a00-0757-11eb-994d-127c9dcba1d4.png)

yehudab:
I also updated the project's README file. Should be easier to check the above diffs, after reading the new [RTL section](https://github.com/mozilla/fxa/blob/a623f0852a0e23d1c6de6e0bf691b402d1f8b607/packages/fxa-settings/README.md#rtl-support) in my branch.

yehudab:
After more twicks and fixes for the new settings page, I feel pretty confident with my work. I'll open the 2 PRs now.

PR #111, opened 2016-02-01
Add feature card for AMO migration.

rfk:
Thanks to @vladikoff's prompting over in https://github.com/mozilla/fxa-content-server/issues/2477#issuecomment-174552533, I've tried to capture our requirements for the AMO migration flow as a feature card.  There are three largely independent sub-features here, but since we want them all to ship together in support of the migration, I think it's probably OK to bundle them all into a single feature number.

@shane-tomlinson, this includes a detailed approach to resolving the permission-prompt issues discussed in https://github.com/mozilla/fxa-content-server/issues/2477, could you please r? based on the work you already did towards that.

@seanmonstar, this describes an approach to choosing between signup and signin based on your comments in https://github.com/mozilla/fxa-oauth-server/issues/379, could you please r? that part.

PR #6285, opened 2020-08-24
feat(settings): Add avatar drop down component

LZoog:
## Because
* Settings needs an avatar drop-down showing the user's avatar, display name or primary email, and sign out button.

## This pull request
* Adds DropDownAvatarMenu component and .drop-down-menu class for Bento menu reuse
* Adds client-side destroySession mutation
* Uses the useAccount hook in Avatar component instead of drilling the prop down and updates the default to use placekitten URL so the avatar actually loads in an img tag
* Adds truncate class to UnitRow showing the displayName

Closes: #4869

## Checklist

Tailwindcss needed a package update to work with gradients.

### DIFF
--- a/packages/fxa-settings/src/components/Avatar/index.stories.tsx
+++ b/packages/fxa-settings/src/components/Avatar/index.stories.tsx
@@ -4,12 +4,17 @@
 
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { Avatar } from '.';
+import { MockedCache } from '../../models/_mocks';
+import Avatar from '.';
 
 storiesOf('Components|Avatar', module)
   .add('default avatar', () => (
-    <Avatar avatarUrl={null} className="w-32 h-32" />
+    <MockedCache account={{ avatarUrl: null }}>
+      <Avatar className="w-32 h-32" />
+    </MockedCache>
   ))
   .add('non-default avatar', () => (
-    <Avatar className="w-32 h-32" avatarUrl="http://placekitten.com/256/256" />
+    <MockedCache>
+      <Avatar className="w-32 h-32" />
+    </MockedCache>
   ));
--- a/packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
+++ b/packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
@@ -1,0 +1,148 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { render, screen, fireEvent, act } from '@testing-library/react';
+import { MockedCache } from '../../models/_mocks';
+import { AlertBarRootAndContextProvider } from '../../lib/AlertBarContext';
+import DropDownAvatarMenu, { DESTROY_SESSION_MUTATION } from '.';
+
+const mockGqlSuccess = () => ({
+  request: {
+    query: DESTROY_SESSION_MUTATION,
+    variables: { input: {} },
+  },
+  result: {
+    data: {
+      resendSecondaryEmailCode: {
+        clientMutationId: null,
+      },
+    },
+  },
+});
+
+const mockGqlError = () => ({
+  request: {
+    query: DESTROY_SESSION_MUTATION,
+    variables: { input: {} },
+  },
+  error: new Error('Aw shucks'),
+});
+
+describe('DropDownAvatarMenu', () => {
+  it('renders and toggles as expected with default values', () => {
+    render(
+      <MockedCache account={{ avatarUrl: null, displayName: null }}>
+        <DropDownAvatarMenu />
+      </MockedCache>
+    );
+
+    const toggleButton = screen.getByTestId('drop-down-avatar-menu-toggle');
+    const dropDownId = 'drop-down-avatar-menu';
+    const dropDown = screen.queryByTestId(dropDownId);
+
+    expect(toggleButton).toHaveAttribute('title', 'Firefox Account Menu');
+    expect(toggleButton).toHaveAttribute('aria-controls', dropDownId);
+    expect(toggleButton).toHaveAttribute('aria-expanded', 'false');
+    expect(dropDown).not.toBeInTheDocument;
+
+    fireEvent.click(toggleButton);
+    expect(toggleButton).toHaveAttribute('aria-expanded', 'true');
+    expect(dropDown).toBeInTheDocument;
+    expect(screen.getByTestId('drop-down-name-or-email').textContent).toContain(
+      'johndope@example.com'
+    );
+
+    fireEvent.click(toggleButton);
+    expect(toggleButton).toHaveAttribute('aria-expanded', 'false');
+    expect(dropDown).not.toBeInTheDocument;
+  });
+
+  it('renders as expected with avatarUrl and displayName set', () => {
+    render(
+      <MockedCache>
+        <DropDownAvatarMenu />
+      </MockedCache>
+    );
+    fireEvent.click(screen.getByTestId('drop-down-avatar-menu-toggle'));
+    expect(screen.getByTestId('drop-down-name-or-email').textContent).toContain(
+      'John Dope'
+    );
+  });
+
+  it('closes on esc keypress', () => {
+    render(
+      <MockedCache>
+        <DropDownAvatarMenu />
+      </MockedCache>
+    );
+    const dropDown = screen.queryByTestId('drop-down-avatar-menu');
+
+    fireEvent.click(screen.getByTestId('drop-down-avatar-menu-toggle'));
+    expect(dropDown).toBeInTheDocument;
+    fireEvent.keyDown(window, { key: 'Escape' });
+    expect(dropDown).not.toBeInTheDocument;
+  });
+
+  it('closes on click outside', () => {
+    const { container } = render(
+      <MockedCache>
+        <div className="w-full flex justify-end">
+          <div className="flex pr-10 pt-4">
+            <DropDownAvatarMenu />
+          </div>
+        </div>
+      </MockedCache>
+    );
+    const dropDown = screen.queryByTestId('drop-down-avatar-menu');
+
+    fireEvent.click(screen.getByTestId('drop-down-avatar-menu-toggle'));
+    expect(dropDown).toBeInTheDocument;
+    fireEvent.click(container);
+    expect(dropDown).not.toBeInTheDocument;
+  });
+
+  describe('destroySession', () => {
+    it('redirects the user on success', async () => {
+      window.location.assign = jest.fn();
+      const mocks = [mockGqlSuccess()];
+
+      render(
+        <MockedCache {...{ mocks }}>
+          <DropDownAvatarMenu />
+        </MockedCache>
+      );
+
+      fireEvent.click(screen.getByTestId('drop-down-avatar-menu-toggle'));
+      await act(async () => {
+        fireEvent.click(screen.getByTestId('sign-out'));
+      });
+      expect(window.location.assign).toHaveBeenCalledWith(
+        `${window.location.origin}/signin`
+      );
+    });
+
+    it('displays an error in the AlertBar', async () => {
+      const mocks = [mockGqlError()];
+
+      const { rerender } = render(<AlertBarRootAndContextProvider />);
+      rerender(
+        <MockedCache {...{ mocks }}>
+          <AlertBarRootAndContextProvider>
+            <DropDownAvatarMenu />
+          </AlertBarRootAndContextProvider>
+        </MockedCache>
+      );
+
+      fireEvent.click(screen.getByTestId('drop-down-avatar-menu-toggle'));
+      await act(async () => {
+        fireEvent.click(screen.getByTestId('sign-out'));
+      });
+
+      expect(screen.getByTestId('sign-out-error').textContent).toContain(
+        'Aw shucks'
+      );
+    });
+  });
+});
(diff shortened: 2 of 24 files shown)


packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { displayName, primaryEmail } = useAccount();
+  const [isRevealed, setRevealed] = useState(false);
+  const toggleRevealed = () => setRevealed(!isRevealed);
+  const avatarMenuInsideRef = useClickOutsideEffect<HTMLDivElement>(
+    setRevealed
+  );
+  useEscKeydownEffect(setRevealed);
+  const dropDownId = 'drop-down-avatar-menu';
+
+  //TODO: mutation for Sign out
```
6a68, line 22: might be nice to include the bug number in the TODO

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+          className="drop-down-menu -left-52"
+        >
+          <div className="flex flex-wrap">
+            <div className="flex w-full p-4">
+              <div className="mr-3 flex flex-1">
+                <Avatar className="w-10" />
+              </div>
+              <div className="flex flex-4 truncate">
+                <p className="leading-5 max-w-full">
+                  <span className="text-grey-400 text-xs">Signed in as</span>
```
6a68, line 49: just curious, do we have a localization / RTL plan for this component / globally? IIRC `fluent-react` requires some additional work for inserting variables inside strings, but it's been a while.

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -51,7 +51,13 @@ export const UnitRow = ({
       <div className="unit-row-header">
         <h3 data-testid="unit-row-header">{header}</h3>
       </div>
-      <div className="unit-row-content">
+      {/* apply 'truncate' to the parent if 'headerValueClassName' contains it */}
```
6a68, line 54: truncation tends to lead to interesting edge cases in German locales. just curious, do we have storybook plans for specific localizations?

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    className="font-bold block max-w-full truncate"
+                    data-testid="drop-down-name-or-email"
+                  >
+                    {displayName || primaryEmail.email}
+                  </span>
+                </p>
+              </div>
+            </div>
+            <div className="w-full">
+              <div className="bg-gradient-to-r from-blue-500 via-pink-700 to-yellow-500 h-px" />
```
6a68, line 60: these classnames are 💯

packages/fxa-settings/src/components/Avatar/index.stories.tsx
```diff
@@ -4,12 +4,17 @@
 
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { Avatar } from '.';
+import { MockedCache } from '../../models/_mocks';
```
6a68, line 7: TIL! `MockedCache` is really clever

packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const toggleButton = screen.getByTestId('drop-down-avatar-menu-toggle');
+    const dropDownId = 'drop-down-avatar-menu';
+    const dropDown = screen.queryByTestId(dropDownId);
+
+    expect(toggleButton).toHaveAttribute('title', 'Firefox Account Menu');
+    expect(toggleButton).toHaveAttribute('aria-controls', dropDownId);
+    expect(toggleButton).toHaveAttribute('aria-expanded', 'false');
+    expect(dropDown).not.toBeInTheDocument;
+
+    fireEvent.click(toggleButton);
```
6a68, line 27: huh, so we don't need to wait a turn for the event to propagate and the DOM to update? it looks like the [testing-library/react `fireEvent`](https://github.com/testing-library/react-testing-library/blob/master/src/fire-event.js) doesn't do anything special (setTimeout / rAF / wait to resolve a promise)

packages/fxa-settings/src/components/Avatar/index.stories.tsx
LZoog: Yeah, it's really nice! There's a note [in the docs about it here](https://github.com/mozilla/fxa/tree/main/packages/fxa-settings#components-that-use-useaccount). :)

packages/fxa-settings/src/components/UnitRow/index.tsx
LZoog: We don't, but that's a _really_ great suggestion - we could have one for German and another for rtl languages. I've added a note for myself, I'll get a ticket filed in settings v2 for these two cases (let me know if there should be more), 'cause actually, we truncate it preeettty short right now:

Content server: 
image

With this PR:
image

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
LZoog: I have #5169 sitting in an l10n user story in which I'm hoping we can use that tailwindcss package linked in there for, but it might end up being a much larger task if we can't.

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
LZoog: This was why my PR was still a draft. Pushing updates in a sec...

packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
LZoog: It's magic!

Yeah, I'm not sure why we don't have to wait for a rerender in the test (it's nice we don't need to, guess it's automatic somehow), but we do have to wrap `fireEvent` in an `await act(async () => {` when we need to wait for a GQL mutation to run.

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    destroySession({
+      variables: { input: {} },
+    });
+  };
+
+  if (data) {
+    clearSignedInAccountUid();
+    window.location.assign(`${window.location.origin}/signin`);
+    return null;
+  }
```
LZoog, line 56: @dannycoates it feels like I'm missing something here, but I'm not entirely sure what.

In the content server, we `clearSignedInAccountUid` (removing account uid from `localStorage`), clear everything set on the `Session` object created in the content server, then navigate back to `signin` with `queryString` params. Here I'm removing the account uid from `localStorage`, but I don't _think_ we have anything else to clear, and, I don't think we need to transfer params since when the user goes to hit `beta/settings` they're redirected to `get_flow`?

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [isRevealed, setRevealed] = useState(false);
+  const toggleRevealed = () => setRevealed(!isRevealed);
+  const avatarMenuInsideRef = useClickOutsideEffect<HTMLDivElement>(
+    setRevealed
+  );
+  useEscKeydownEffect(setRevealed);
+  const [alertBarRevealed, revealAlertBar, hideAlertBar] = useBooleanState();
+  const dropDownId = 'drop-down-avatar-menu';
+
+  // TODO: FXA-6249
```
jodyheavener, line 35: FXA-6249 doesn't seem to exist, but I assume it's that one about DRYing up the useMutation. If #6292 is merged before this, you'll be able to use `useHandledMutation` properly here.

LZoog: Oh good catch, yeah I meant #6249 which is FXA-2450.

I think we'll still have a TODO here since that ticket also deals with figuring out what should report to Sentry, what error text copy to grab from Meridel, and cleaning up/standardizing all of the mutations and properly styling the alertbar accordingly.

packages/fxa-settings/src/lib/cache.ts
```diff
@@ -45,6 +45,10 @@ export function sessionToken(newToken?: hexstring) {
   }
 }
 
+export function clearSignedInAccountUid() {
+  storage.remove('currentAccountUid');
```
dannycoates, line 49: we should also clear that uid from 'accounts'

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
dannycoates: this redirect url seems fine

packages/fxa-settings/src/components/DropDownAvatarMenu/sign-out.svg
```diff
@@ -0,0 +1 @@
+<svg fill="none" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg"><path d="M6.333 9.26H17M12.733 4.993L17 9.26l-4.267 4.267M9.533 16.727H2.067A1.067 1.067 0 011 15.66V2.86c0-.589.478-1.067 1.067-1.067h7.466" stroke="#111" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>
```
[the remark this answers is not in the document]
LZoog: Well, the _stroke_ is set to #111, the fill is set to `none`. If I set the `fill` here to `currentColor`, it looks like this:

image

I'm not sure this one needs a hover state anyway since the text doesn't change color. It could be re-exported later if we want to add a hover color change.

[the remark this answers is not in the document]
jodyheavener: Oh right, maybe [`stroke-current`](https://tailwindcss.com/docs/stroke#app) then?

packages/fxa-settings/src/styles/drop-down-menu.scss
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.drop-down-menu {
+  @apply w-64 bg-white absolute shadow-sm rounded;
```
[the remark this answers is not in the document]
jodyheavener: Noted pwalm's comment, sweet!

packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
6a68: huh, cool. maybe the test stays in virtual DOM-land 🤷

packages/fxa-settings/src/components/DropDownAvatarMenu/sign-out.svg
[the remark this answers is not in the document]
LZoog: Huh, TIL. I've known about `currentColor` on `fill` for years, but I didn't know there was a stroke equivalent 😂 Made the change.

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  if (data) {
+    clearSignedInAccountUid();
+    window.location.assign(`${window.location.origin}/signin`);
+    return null;
+  }
+
+  return (
+    <>
+      {alertBarRevealed && error && (
+        <AlertBar onDismiss={hideAlertBar}>
```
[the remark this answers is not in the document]
jodyheavener: Now that #6292 is merged, once rebased, you should be able to give this a type of `error`

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+                      data-testid="drop-down-name-or-email"
+                    >
+                      {displayName || primaryEmail.email}
+                    </span>
+                  </p>
+                </div>
+              </div>
+              <div className="w-full">
+                <div className="bg-gradient-to-r from-blue-500 via-pink-700 to-yellow-500 h-px" />
+                <div className="p-6">
```
jodyheavener, line 103: Nitpick: something like `px-4 py-5` might be a little closer to the padding in Figma

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    className="pl-3 group"
+                    onClick={signOut}
+                    data-testid="sign-out"
+                  >
+                    <SignOut
+                      height="18"
+                      width="18"
+                      className="mr-3 inline-block stroke-current"
+                    />
+                    <span className="group-hover:underline">Sign out</span>
```
jodyheavener, line 114: Nitpick: this is ever so slightly improved with `vertical-align: middle;`

LZoog: _ever so slightly_ is like, less than 1px, this kind of cracked me up. But you're right, that less than 1px move does look slightly better 😂

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
LZoog: Hmm, I'm not sure, `p-6` is 24px all the way around. If anything, I could bump the left padding _up_ to a `7`.

image
image

LZoog: I'll leave this one as-is, we can address it in a design review if Phil/Jess prefer less padding here.

jodyheavener: for posterity

Artboard

Issue #6199, opened 2020-08-14
“Unverified session” error received when deleting an account logged in via pairing method


vasilicatamas:
**Build**
Stage [Train 1.184.0]

**Affected platforms**
Android 9
iOS 13.6

**Steps to reproduce**
1. Login on a desktop using a user which has 2FA enabled and navigate to https://accounts.stage.mozaws.net/pair (firefox.com/pair).
2. Login on a mobile device (Android/iOS) via pairing method.
3. On the mobile device, navigate to Settings page.
4. Delete the account.

**Expected results**
Firefox account is successfully deleted. 

**Actual Results**
“Unverified Session” error is displayed and the account is not deleted. 

**Additional notes**
This issue might be related to #4257

![Screenshot_23](https://user-images.githubusercontent.com/18114210/90255380-e5a68e00-de4c-11ea-9bc6-a25664d4292e.png)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2423)
┆Issue Number: FXA-2423

vbudhram:
@vasilicatamas I don't think this is a major issue/blocker for the current train. Not a lot of users go through this flow (2FA/pair/delete account) at the moment, so we could prioritize it for next train.

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.186.0] under iOS 13.6 and Android 9. The account logged in via pairing method is successfully deleted.

![issue9](https://user-images.githubusercontent.com/18114210/91577386-48ba1980-e951-11ea-8812-8fc731830269.gif)

Issue #4257, opened 2020-02-24
”Unverified session” error when deleting an account after password reset on account with 2FA enabled


vasilicatamas:
**Build:**
Stage [Train 1.159.0]

**Affected platforms:**
Windows 10

**Steps to reproduce:**
1. Launch Firefox and login using a Firefox account.
2. Go to Manage account page and enabled the two-step authentication.
3. Change the account password.
4. Try to delete the account, inserting the new password. 

**Expected results:**
Firefox account is successfully deleted.

**Actual Results:**
“Unverified Session” error is displayed and the account is not deleted. 

**Note:**
- A potential solution would be to show a dialog asking for the security code as a  reset password confirmation for accounts which have 2FA enabled.

![deleteaccount2](https://user-images.githubusercontent.com/18114210/75162144-b7f64d00-5725-11ea-8a81-4b9f7edd0856.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1193)
┆Issue Number: FXA-1193

rfk:
Thanks for finding this! A couple of notes from a quick look since I saw this in my inbox:

> Change the account password.

This step destroys all the user's sessionTokens, and then creates a new sessionToken so that they can appear to be seamlessly still signed in on the calling page. The code for creating this new sessionToken is [here](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/routes/password.js#L337), but it looks like it does not correctly copy the 2FA verification status. So when this happens:

> Try to delete the account, inserting the new password.

The FxA serve quite rightly refuses to process the request, thinking that the user has not done their 2FA verification loop.

@vbudhram I know we talked in the past about a way to clean up some of these flows by refactoring the way sessionTokens and the verification thereof are handled, perhaps it's time for us to revisit that topic.

vasilicatamas:
The error still appears on Stage [Train 1.192.0] but only for a few seconds, because the “Enter security code” screen is displayed. After typing the security code, the Manage Account page is refreshed and the user needs to insert again the password in order to finish the deletion process.

![issue19](https://user-images.githubusercontent.com/18114210/97182111-c8733180-17a4-11eb-904a-c88a9fff37af.gif)

vasilicatamas:
This issue was verified as fixed in FxA Stage [Train 1.199.0] under Windows 10 64-bit and Android 9. “Unverified Session” error is no longer displayed and the account is successfully deleted.

[...]

PR #6295, opened 2020-08-25
fix(totp): Add mixin to verify session if a request is `unverified session` based

vbudhram:
## Because

- When an account is paired, the paired session would not have the assurance level to delete the account (ie `unverified session`)

## This pull request

- Adds a mixin to redirect the user to verify the session with TOTP if they have it setup on the account

Closes: #6199 #4257

## Checklist

### Testing instructions

I wasn't actually able to test the pairing flow locally with 2FA enabled. You would need custom build of FxiOS (pointing to local FxA) and then be able to run it on a physical phone since the iOS simulator does not support a camera. I was able to simulate this by the following though

1. Disable session being verified for new accounts via [skipForNewAccount](https://github.com/mozilla/fxa/blob/db20d69957612ea411c1b89d992d52b20763d927/packages/fxa-auth-server/config/index.ts#L1349)
2. Create account A and sign into non Firefox browser A
  a. The account at this state is considered verified but the session is unverified
3. Sign into non Firefox browser B
4. Enable 2FA on non Firefox browser A
5. Attempt to delete account on non Firefox browser B

You will be prompted to enter the 2FA code before deleting the account.

### Screenshot
2fa

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/auth-errors.js
+++ b/packages/fxa-content-server/app/scripts/lib/auth-errors.js
@@ -149,6 +149,10 @@ var ERRORS = {
     errno: 136,
     message: t('This email was already verified by another user'),
   },
+  UNVERIFIED_SESSION: {
+    errno: 138,
+    message: t('Unverified session'),
+  },
   EMAIL_PRIMARY_EXISTS: {
     errno: 139,
     message: t('Secondary email must be different than your account email'),
--- a/packages/fxa-content-server/app/scripts/views/mixins/upgrade-session-redirect-mixin.js
+++ b/packages/fxa-content-server/app/scripts/views/mixins/upgrade-session-redirect-mixin.js
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import AuthErrors from '../../lib/auth-errors';
+
+/**
+ * View mixin to support a user upgrading their session to
+ * be verified. If a form is submitted and returns with an
+ * `unverified session` error, then the user will be redirected
+ * to either verify the session with either a sign-in code
+ * or a TOTP code.
+ *
+ */
+export default {
+  initialize() {
+    this.skipBaseOnFormSubmit = true;
+  },
+
+  async onFormSubmit() {
+    try {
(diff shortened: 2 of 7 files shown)


packages/fxa-content-server/app/scripts/lib/auth-errors.js
```diff
@@ -149,6 +149,10 @@ var ERRORS = {
     errno: 136,
     message: t('This email was already verified by another user'),
   },
+  UNVERIFIED_SESSION: {
+    errno: 138,
```
jodyheavener, line 153: I'm impressed this errno was available

packages/fxa-content-server/app/scripts/views/mixins/upgrade-session-redirect-mixin.js
```diff
@@ -0,0 +1,46 @@
...
+          } else {
+            // otw try to do a verification via an email code
+            return account.verifySessionResendCode().then(() => {
+              return this.replaceCurrentPage('/signin_token_code', {
+                redirectTo: this.window.location.href,
+              });
+            });
+          }
+        });
+      }
```
jodyheavener, line 43: Anything preventing us from async/awaiting all this goodness?

packages/fxa-content-server/app/scripts/views/sign_in_totp_code.js
```diff
@@ -37,6 +37,11 @@ const View = FormView.extend({
         if (result.success) {
           this.logFlowEvent('success', this.viewName);
 
+          const redirectTo = this.model.get('redirectTo');
+          if (redirectTo) {
+            return (this.window.location.href = redirectTo);
```
jodyheavener, line 42: Can we use `navigate` or `navigateAway` here?

packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
```diff
@@ -63,9 +63,9 @@ const View = FormView.extend({
       .then(() => {
         this.logViewEvent('success');
 
-        const redirectTo = this.model.get('redirectTo');
-        if (redirectTo) {
-          return (this.window.location.href = redirectTo);
+        const redirectPathname = this.model.get('redirectPathname');
+        if (redirectPathname) {
+          return this.navigate(redirectPathname);
```
vbudhram, line 68: This was might be the cause of https://github.com/mozilla/fxa/issues/6342

PR #6302, opened 2020-08-26
task(fxa-settings): rework Two-step auth unit row

meandavejustice:
## Because

- detect Two-step auth and add ability to disable

## This pull request

- created `UnitRowTwoStepAuth` component
- added `SecondaryButtonClassName`,` SecondaryCta`, and `SecondaryModalReveal`
options to UnitRow
- updated Security component, and all tests and stories to reflect
changes
- updated each story touched with `<LocationProvider/>` fix

Closes: #4913 #6327 

## Checklist

## Screenshots (Optional)

unset-totp

set-totp

### DIFF
--- a/packages/fxa-settings/src/components/Modal/index.tsx
+++ b/packages/fxa-settings/src/components/Modal/index.tsx
@@ -17,6 +17,8 @@ type ModalProps = {
   hasButtons?: boolean;
   headerId: string;
   descId: string;
+  confirmText?: string;
+  confirmBtnClassName?: string;
   'data-testid'?: string;
 };
 
@@ -28,6 +30,8 @@ export const Modal = ({
   hasButtons = true,
   headerId,
   descId,
+  confirmText = 'Confirm',
+  confirmBtnClassName = 'cta-primary',
   'data-testid': testid = 'modal',
 }: ModalProps) => {
   const modalInsideRef = useClickOutsideEffect<HTMLDivElement>(onDismiss);
@@ -43,7 +47,7 @@ export const Modal = ({
         <div
           data-testid="modal-content-container"
           className={classNames(
-            'max-w-lg bg-white mx-auto rounded-xl',
+            'max-w-md bg-white mx-auto rounded-xl',
             className
           )}
           ref={modalInsideRef}
@@ -64,7 +68,7 @@ export const Modal = ({
             </button>
           </div>
 
-          <div className="px-4 tablet:px-12 pb-10">
+          <div className="px-4 tablet:px-10 pb-10">
             <div>{children}</div>
             {hasButtons && (
               <div className="flex justify-center mx-auto mt-6 max-w-64">
@@ -78,11 +82,11 @@ export const Modal = ({
 
                 {onConfirm && (
                   <button
-                    className="mx-2 flex-1 cta-primary"
+                    className={classNames('mx-2 flex-1', confirmBtnClassName)}
                     data-testid="modal-confirm"
                     onClick={onConfirm}
                   >
-                    Confirm
+                    {confirmText}
                   </button>
                 )}
               </div>
--- a/packages/fxa-settings/src/components/UnitRow/index.stories.tsx
+++ b/packages/fxa-settings/src/components/UnitRow/index.stories.tsx
@@ -44,7 +44,6 @@ storiesOf('Components|UnitRow', module)
           headerValue={null}
           {...{
             revealModal,
-            modalRevealed,
           }}
         >
           <p className="text-sm mt-3">Content goes here.</p>
@@ -61,4 +60,35 @@ storiesOf('Components|UnitRow', module)
         </UnitRow>
       );
     }
-  );
+  )
+  .add('with modal-triggering secondary CTA', () => {
+    const [
+      secondaryModalRevealed,
+      revealSecondaryModal,
+      hideModal,
+    ] = useBooleanState();
+    return (
+      <UnitRow
+        header="Display name"
+        headerValue={null}
+        {...{
+          secondaryButtonClassName: 'bg-red-500 text-white border-red-600',
+          revealSecondaryModal,
+          ctaText: 'Change',
+          route: '#',
+        }}
+      >
+        <p className="text-sm mt-3">Content goes here.</p>
+        <p className="text-grey-400 text-xs mt-2">More content.</p>
+
+        {secondaryModalRevealed && (
+          <Modal onDismiss={hideModal} headerId="some-id" descId="some-desc">
+            <h2 id="some-id" className="font-bold text-xl text-center mb-2">
+              Modal header
+            </h2>
+            <p id="some-desc">Modal description here.</p>
+          </Modal>
+        )}
+      </UnitRow>
+    );
+  });
--- a/packages/fxa-settings/src/components/UnitRow/index.test.tsx
+++ b/packages/fxa-settings/src/components/UnitRow/index.test.tsx
@@ -83,4 +83,22 @@ describe('UnitRow', () => {
       'Create'
     );
   });
+
+  it('renders `secondaryCtaText`', () => {
+    renderWithRouter(
+      <UnitRow
+        header="Display name"
+        headerValue="Fred Flinstone"
+        route="/display_name"
+        revealSecondaryModal={() => {}}
+      />
+    );
+
+    expect(screen.getByTestId('unit-row-route').textContent).toContain(
+      'Change'
+    );
+    expect(screen.getByTestId('unit-row-modal').textContent).toContain(
+      'Disable'
+    );
+  });
 });
--- a/packages/fxa-settings/src/components/UnitRow/index.tsx
+++ b/packages/fxa-settings/src/components/UnitRow/index.tsx
@@ -7,16 +7,62 @@ import classNames from 'classnames';
 import { useFocusOnTriggeringElementOnClose } from '../../lib/hooks';
 import { Link, RouteComponentProps, useLocation } from '@reach/router';
 
+type ModalButtonProps = {
+  ctaText: string;
+  className?: string;
+  revealModal: () => void;
+  modalRevealed?: boolean;
+  alertBarRevealed?: boolean;
+  leftSpaced?: boolean;
+};
+
+export const ModalButton = ({
+  ctaText,
+  className,
+  revealModal,
+  modalRevealed,
+  alertBarRevealed,
+  leftSpaced,
+}: ModalButtonProps) => {
+  const modalTriggerElement = useRef<HTMLButtonElement>(null);
+  // If the UnitRow children contains an AlertBar that is revealed,
+  // don't redirect focus back to the element that opened the modal
+  // because focus will be set in the AlertBar.
+  useFocusOnTriggeringElementOnClose(
+    modalRevealed,
+    modalTriggerElement,
+    alertBarRevealed
+  );
+
+  return (
+    <button
+      className={classNames(
+        'cta-base transition-standard',
+        leftSpaced && 'ml-2',
+        className || 'cta-neutral'
+      )}
+      data-testid="unit-row-modal"
+      ref={modalTriggerElement}
+      onClick={revealModal}
+    >
+      {ctaText}
+    </button>
+  );
+};
+
 type UnitRowProps = {
   header: string;
   headerValue: string | null;
   noHeaderValueText?: string;
   ctaText?: string;
+  secondaryCtaText?: string;
+  secondaryCtaRoute?: string;
+  secondaryButtonClassName?: string;
   children?: React.ReactNode;
   headerValueClassName?: string;
   route?: string;
   revealModal?: () => void;
-  modalRevealed?: boolean;
+  revealSecondaryModal?: () => void;
   alertBarRevealed?: boolean;
 };
 
@@ -28,23 +74,17 @@ export const UnitRow = ({
   headerValueClassName,
   noHeaderValueText = 'None',
   ctaText,
+  secondaryCtaText = 'Disable',
+  secondaryCtaRoute,
+  secondaryButtonClassName,
   revealModal,
-  modalRevealed,
+  revealSecondaryModal,
   alertBarRevealed,
 }: UnitRowProps & RouteComponentProps) => {
   ctaText = ctaText || (headerValue ? 'Change' : 'Add');
 
-  const modalTriggerElement = useRef<HTMLButtonElement>(null);
-  // If the UnitRow children contains an AlertBar that is revealed,
-  // don't redirect focus back to the element that opened the modal
-  // because focus will be set in the AlertBar.
-  useFocusOnTriggeringElementOnClose(
-    modalRevealed,
-    modalTriggerElement,
-    alertBarRevealed
-  );
-
   const location = useLocation();
+  const multiButton = !!(route || secondaryCtaRoute);
 
   return (
     <div className="unit-row">
@@ -62,10 +102,10 @@ export const UnitRow = ({
       </div>
 
       <div className="unit-row-actions">
-        <div>
+        <div className="flex items-center">
           {route && (
             <Link
-              className="cta-neutral cta-base"
+              className="cta-neutral cta-base transition-standard mr-2"
               data-testid="unit-row-route"
               to={`${route}${location.search}`}
             >
@@ -74,14 +114,30 @@ export const UnitRow = ({
           )}
 
           {revealModal && (
-            <button
-              className="cta-neutral cta-base"
-              data-testid="unit-row-modal"
-              ref={modalTriggerElement}
-              onClick={revealModal}
+            <ModalButton
+              leftSpaced={multiButton}
+              {...{ revealModal, ctaText, alertBarRevealed }}
+            />
+          )}
+
+          {secondaryCtaRoute && (
+            <Link
+              className="cta-neutral cta-base transition-standard mr-2"
+              data-testid="unit-row-route"
+              to={`${secondaryCtaRoute}${location.search}`}
             >
-              {ctaText}
-            </button>
+              {secondaryCtaText}
+            </Link>
+          )}
+
+          {revealSecondaryModal && (
+            <ModalButton
+              leftSpaced={multiButton}
+              revealModal={revealSecondaryModal}
+              ctaText={secondaryCtaText}
+              className={secondaryButtonClassName}
+              alertBarRevealed={alertBarRevealed}
+            />
           )}
         </div>
       </div>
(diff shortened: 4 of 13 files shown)


packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -7,16 +7,59 @@ import classNames from 'classnames';
...
+  useFocusOnTriggeringElementOnClose(
+    modalRevealed,
+    modalTriggerElement,
+    alertBarRevealed
+  );
+
+  return (
+    <button
+      className={classNames(
+        'cta-neutral cta-base transition-standard ml-3',
```
LZoog, line 38: This looks good in Storybook and all the way up to tablet:
image


But at desktop, they do this: 
image

I think we can fix this by wrapping each button in a `div` so they take up equal space in `unit-row-actions` instead of wrapping both buttons in a `div`.

Also, I think we need a new class that's not `cta-neutral`, that one's for grey buttons and the hover state is a little off. We need a new one for "caution" (that's what Figma calls it anyway), you could add them to `ctas.scss`:
image

image

packages/fxa-settings/src/components/UnitRow/index.stories.tsx
```diff
@@ -59,4 +60,32 @@ storiesOf('Components|UnitRow', module)
         </UnitRow>
       );
     }
-  );
+  )
+  .add('with modal-triggering secondary CTA', () => {
+    const [modalRevealed, revealSecondaryModal, hideModal] = useBooleanState();
```
LZoog, line 65: I think we should probably call these `[secondaryModalRevealed, revealSecondaryModal, hideSecondaryModal]` just so it's clear these are all for the modal tied to the secondary CTA.

packages/fxa-settings/src/components/UnitRow/index.stories.tsx
```diff
@@ -59,4 +60,32 @@ storiesOf('Components|UnitRow', module)
...
+    return (
+      <UnitRow
+        header="Display name"
+        headerValue={null}
+        {...{
+          secondaryCtaText: 'Disable',
```
LZoog, line 71: It looks like `secondaryCtaText` is set to "Disable" by default, so I think you can take this out.

packages/fxa-settings/src/components/UnitRow/index.test.tsx
```diff
@@ -83,4 +83,28 @@ describe('UnitRow', () => {
...
   });
+
+  it('renders `secondaryCtaText`', () => {
+    renderWithRouter(
+      <UnitRow
+        header="Display name"
+        headerValue={null}
+        noHeaderValueText="Not set"
+        ctaText="Create"
+        secondaryCtaText="Disable"
```
LZoog, line 94: I think you can take `secondaryCtaText="Disable"` out since it gets "Disable" by default.

I also think `ctaText` can be removed, it should get "Create" when `headerValue` is falsey.

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -7,16 +7,59 @@ import classNames from 'classnames';
...
+  secondaryCtaText?: string;
+  secondaryCtaRoute?: string;
+  secondaryButtonClassName?: string;
   children?: React.ReactNode;
   headerValueClassName?: string;
   route?: string;
   revealModal?: () => void;
-  modalRevealed?: boolean;
+  revealSecondaryModal?: () => void;
   alertBarRevealed?: boolean;
```
LZoog, line 63: Looks like `alertBarRevealed` is no longer used here.

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        },
+      });
+    },
+  });
+
+  return (
+    <UnitRow
+      header="Two-step authentication"
+      headerValueClassName={exists ? 'text-green-800' : ''}
+      headerValue={exists ? 'Enabled' : 'Not Set'}
```
LZoog, line 57: I think we can simplify this a tad since `noHeaderValueText` is also a prop here and there's a lot of conditionals based on `exists`. What do you think about something like:

```
  const conditionalUnitRowProps = exists ? {
    headerValueClassName: 'text-green-800',
    headerValue: 'Enabled',
    secondaryButtonClassName: 'bg-red-500 text-white border-red-600',
    revealSecondaryModal: revealModal,
  } : {
    headerValue: null,
    noHeaderValueText: 'Not Set',
    secondaryCtaText: undefined,
    revealSecondaryModal: undefined
  };

  return (
    <UnitRow
      header="Two-step authentication"
      route="/beta/settings/two_step_authentication"
      {...conditionalUnitRowProps}
    >
```

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -7,16 +7,59 @@ import classNames from 'classnames';
 import { useFocusOnTriggeringElementOnClose } from '../../lib/hooks';
 import { Link, RouteComponentProps, useLocation } from '@reach/router';
 
+type ModalButtonProps = {
+  ctaText: string;
+  classname?: string;
```
LZoog, line 12: We probably want to call this `className` just for consistency.

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      {modalRevealed && (
+        <VerifiedSessionGuard onDismiss={hideModal} onError={onError}>
+          <Modal
+            onDismiss={hideModal}
+            onConfirm={disableTwoStepAuth}
+            headerId="some-id"
+            descId="some-desc"
+          >
+            <h2 id="some-id" className="font-bold text-xl text-center mb-2">
+              Disable Two-stop Authentication?
```
dannycoates, line 79: "two-step" ... also lower case "authentication"

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      hideModal();
+      revealAlertBar();
+    },
+    onError,
+    ignoreResults: true,
+    update: (cache) => {
+      cache.modify({
+        fields: {
+          account: (existing) => {
+            return { ...existing, totp: false };
```
dannycoates, line 46: should be `{...existing, totp: { exists: false, verified: false }}`

packages/fxa-settings/src/components/UnitRow/index.tsx
jodyheavener: Adding to this, the mobile layout needs some margin work.

Screen Shot 2020-08-26 at 2 21 31 PM

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
jodyheavener, line 79: The spelling/capitalization needs to be updated, but also this modal should more closely resemble the Figma doc. Looks like we need to use the Red disable button that says "Disable", instead of a blue button that says "Confirm", and the "replacing your recovery codes" link looks off. The button horizontal spacing is also off, but I am noticing this on main too, so it could be addressed later if not here.

Figma:

Screen Shot 2020-08-26 at 2 27 13 PM

Actual:

Screen Shot 2020-08-26 at 2 26 43 PM

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <p className="text-sm mt-3">
+        Prevent someone else from logging in by requiring a unique code only you
+        have access to.
+      </p>
+      {modalRevealed && (
+        <VerifiedSessionGuard onDismiss={hideModal} onError={onError}>
+          <Modal
+            onDismiss={hideModal}
+            onConfirm={disableTwoStepAuth}
+            headerId="some-id"
```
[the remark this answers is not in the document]
meandavejustice: lol, copy and paste

packages/fxa-settings/src/components/UnitRow/index.tsx
meandavejustice: It needs to stay, unfortunately since it is passed through to `ModalButton`

jodyheavener: Maybe it should be, but currently it's not being used 🤔 
Screen Shot 2020-08-27 at 1 34 01 PM

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
meandavejustice: I updated the text of and color of the disable button, I think we should file a new bug for fixing the default styles of the modal since it affects other parts of the app, and seems out of scope for this issue. Filed https://jira.mozilla.com/browse/FXA-2486

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          >
+            <h2 className="font-bold text-xl text-center mb-2">
+              Disable Two-step authentication?
+            </h2>
+            {/* "replacing recovery codes" link below will actually drop you into
+            recovery codes flow in the future. */}
+            <p>
+              You won't be able to undo this action. You also have the option of
+              <LinkExternal
+                className="link-blue text-xs mt-2"
```
jodyheavener, line 98: This class makes the link awkward. You can remove `text-sm mt-2`. Additionally I think you'll need to add `{' '}` right before the opening LinkExternal tag to force a space.
Screen Shot 2020-08-27 at 2 40 22 PM

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <Modal
+            onDismiss={hideModal}
+            onConfirm={disableTwoStepAuth}
+            headerId="two-step-auth-disable-header"
+            descId="two-step-auth-disable-description"
+            confirmText="Disable"
+            confirmBtnClassName="cta-caution"
+          >
+            <h2 className="font-bold text-xl text-center mb-2">
+              Disable Two-step authentication?
```
jodyheavener, line 91: Lowercase `two`

packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+                replacing your recovery codes
+              </LinkExternal>
+              .
+            </p>
+          </Modal>
+        </VerifiedSessionGuard>
+      )}
+      {/* TODO: style AlertBar in the error case */}
+      {alertBarRevealed && (
+        <AlertBar onDismiss={hideAlertBar}>
```
jodyheavener, line 110: Oh! `AlertBar` now supports a `type` -- this could be `success`, `error`, or `info`. Could we update this to use the type for when the update was successful or failed?

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -78,11 +82,14 @@ export const Modal = ({
 
                 {onConfirm && (
                   <button
-                    className="cta-primary transition-standard flex-1"
+                    className={classNames(
+                      'cta-primary transition-standard flex-1',
+                      confirmBtnClassName
```
jodyheavener, line 87: I just noticed this uses `confirmBtnClassName` _and_ `cta-primary`, which would create competing styles. Can `confirmBtnClassName` just default to `cta-primary`?

meandavejustice: I like that much better :)

jodyheavener:
@meandavejustice now that #6350 is merged you should be able to using the layout I set up for modal buttons. Additionally, since this your PR adds the cta-caution class and the new props to set modal class/text, could you uncomment this? https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx#L83-L84

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -28,6 +31,9 @@ export const Modal = ({
   hasButtons = true,
   headerId,
   descId,
+  widthClassName = 'max-w-lg',
```
jodyheavener, line 34: It looks like all modals are the same width in the Figma doc, is there are specific reason for needing varying modal max-widths? It might be best to make them consistent. Looks like `max-w-md` is closest, but if we do that we should reduce the horizontal padding so "Disable two-step authentication?" doesn't break to two lines

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -7,16 +7,56 @@ import classNames from 'classnames';
...
+  // because focus will be set in the AlertBar.
+  useFocusOnTriggeringElementOnClose(
+    modalRevealed,
+    modalTriggerElement,
+    alertBarRevealed
+  );
+
+  return (
+    <button
+      className={classNames('cta-base transition-standard ml-2', className)}
```
jodyheavener, line 37: I think this is missing a `cta-neutral` class:

Screen Shot 2020-09-08 at 3 06 07 PM

jodyheavener (commented):
🤠 Yeehaw this is looking and feeling great. Mutation works on my end. Just a couple cosmetic comments and then I think we should be good to ship.

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -61,29 +94,39 @@ export const UnitRow = ({
...
-          {route && (
-            <Link
-              className="cta-neutral cta-base"
-              data-testid="unit-row-route"
-              to={`${route}${location.search}`}
-            >
-              {ctaText}
-            </Link>
-          )}
+      <div className="flex items-center">
```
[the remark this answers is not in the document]
meandavejustice: This is what we decided because of the button breakage on mobile and certain wide desktop widths, I'm happy to move it back so long as it doesn't break again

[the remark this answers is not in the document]
meandavejustice: It appears to be working well now, had to remove some `ml-2` classes on the buttons we added when trying to get it working before.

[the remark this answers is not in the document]
jodyheavener: I'm not fully up to date on the specific issue, but if it persists with that class feel free to ping me with a description of what's going on and I'll see what I can do -- I love that kind of stuff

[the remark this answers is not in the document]
jodyheavener: It looks like this change resulted in a wonky layout for desktop:

Screen Shot 2020-09-09 at 1 35 03 PM

And while there's nothing visually wrong with this mobile layout, it doesn't match the Figma doc:

Screen Shot 2020-09-09 at 1 34 57 PM

However, this looks like a relatively easy fix. Check out this patch, it should do the trick:

```diff
diff --git a/packages/fxa-settings/src/components/UnitRow/index.tsx b/packages/fxa-settings/src/components/UnitRow/index.tsx
index f0687213b..0b43a30ed 100644
--- a/packages/fxa-settings/src/components/UnitRow/index.tsx
+++ b/packages/fxa-settings/src/components/UnitRow/index.tsx
@@ -13,6 +13,7 @@ type ModalButtonProps = {
   revealModal: () => void;
   modalRevealed?: boolean;
   alertBarRevealed?: boolean;
+  leftSpaced?: boolean;
 };
 
 export const ModalButton = ({
@@ -21,6 +22,7 @@ export const ModalButton = ({
   revealModal,
   modalRevealed,
   alertBarRevealed,
+  leftSpaced,
 }: ModalButtonProps) => {
   const modalTriggerElement = useRef<HTMLButtonElement>(null);
   // If the UnitRow children contains an AlertBar that is revealed,
@@ -35,8 +37,9 @@ export const ModalButton = ({
   return (
     <button
       className={classNames(
-        'cta-base cta-neutral transition-standard mobileLandscape:ml-2',
-        className
+        'cta-base transition-standard',
+        leftSpaced && 'ml-2',
+        className || 'cta-neutral'
       )}
       data-testid="unit-row-modal"
       ref={modalTriggerElement}
@@ -81,6 +84,7 @@ export const UnitRow = ({
   ctaText = ctaText || (headerValue ? 'Change' : 'Add');
 
   const location = useLocation();
+  const multiButton = !!(route || secondaryCtaRoute)
 
   return (
     <div className="unit-row">
@@ -98,10 +102,10 @@ export const UnitRow = ({
       </div>
 
       <div className="unit-row-actions">
-        <div>
+        <div className="flex items-center">
           {route && (
             <Link
-              className="cta-neutral cta-base transition-standard mobileLandscape:mr-2"
+              className="cta-neutral cta-base transition-standard mr-2"
               data-testid="unit-row-route"
               to={`${route}${location.search}`}
             >
@@ -110,12 +114,13 @@ export const UnitRow = ({
           )}
 
           {revealModal && (
-            <ModalButton {...{ revealModal, ctaText, alertBarRevealed }} />
+            <ModalButton
+              leftSpaced={multiButton} {...{ revealModal, ctaText, alertBarRevealed }} />
           )}
 
           {secondaryCtaRoute && (
             <Link
-              className="cta-neutral transition-standard mobileLandscape:mr-2"
+              className="cta-neutral cta-base transition-standard mr-2"
               data-testid="unit-row-route"
               to={`${secondaryCtaRoute}${location.search}`}
             >
@@ -125,6 +130,7 @@ export const UnitRow = ({
 
           {revealSecondaryModal && (
             <ModalButton
+              leftSpaced={multiButton}
               revealModal={revealSecondaryModal}
               ctaText={secondaryCtaText}
               className={secondaryButtonClassName}
```

jodyheavener:
@meandavejustice I left you one more followup layout issue comment, but after that I'm ready to hit green 👌

packages/fxa-settings/src/components/UnitRow/index.tsx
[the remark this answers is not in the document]
meandavejustice: So weird this wasn't showing up for me in storybook, I wonder if I need to be testing on a second monitor, or adjusting my resolution for retina screens. Anyways, I made the changes from the patch and it's looking good in all the states on my machine.

jodyheavener (approved):
Looking good. Thanks for all the effort you put in to this!

Now that the new `useAlertBar` hook is available ([docs](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/README.md#alertbar-and-alertexternal)) you have the option to implement it here as well, but I don't want to hold this up any longer than I already have, so it's you're call if you'd like to update before merging, or leave it for later.

Issue #4851, opened 2020-04-10
[E2E] Setup E2E testing


data-sync-user:
Once a decision from FXA-1557 has been made, we can hook up the library (or install a new one) and set us up to begin creating functional tests for the new Settings page. The PR for this should make it easy to begin adding (or transferring) new tests, and this issue may pair well with FXA-1559 for setup and completion of the first test.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1558)
┆Issue Number: FXA-1558

PR #6313, opened 2020-08-26
chore(tests): Add settings v2 initial functional testing

### LINKED ISSUE #6320: content-server flaky test
content-server functional test:

post_verify_account_recovery - create account recovery

https://app.circleci.com/pipelines/github/mozilla/fxa/13129/workflows/e58e6efe-f51a-4bfd-b18c-b2f4cd0fdfb7/jobs/189972

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2484)
┆Issue Number: FXA-2484

vbudhram:
## Because

- We need some basic functional tests to ensure the new settings page loads

## This pull request

- Adds a new test functional test suite to test settings v2 views
- I unfortunately couldn't move helpers to fxa-shared because some of them access libs from content-server. Moving helpers would require a bigger refactor of content-server
- I ended up moving settings functional tests to content-server functional test directory. I think this makes sense because fxa-settings already requires content-server to be running to access it. If we wanted to expand our smoke testing, its really easy to add them to that suite of tests.

Closes: #4851 Fixes #6320

## Checklist

- [x] I have added necessary documentation (if appropriate)

### DIFF
--- a/packages/fxa-settings/scripts/test-ci.sh
+++ b/packages/fxa-settings/scripts/test-ci.sh
@@ -1,5 +1,31 @@
 #!/bin/bash -ex
 
-yarn workspaces focus fxa-settings
-PUBLIC_URL=/ INLINE_RUNTIME_CHUNK=false CI=false SKIP_PREFLIGHT_CHECK=true yarn build
 CI=yes SKIP_PREFLIGHT_CHECK=true yarn test
+
+mkdir -p config
+
+cd ../../
+mkdir -p ~/.pm2/logs
+mkdir -p artifacts/tests
+yarn workspaces foreach \
+    --verbose \
+    --topological-dev \
+    --include browserid-verifier \
+    --include fxa-auth-db-mysql \
+    --include fxa-auth-server \
+    --include fxa-content-server \
+    --include fxa-profile-server \
+    --include fxa-react \
+    --include fxa-settings \
+    --include fxa-shared \
+    --include fxa-graphql-api \
+    run start > ~/.pm2/logs/startup.log
+npx pm2 ls
+# ensure email-service is ready
+_scripts/check-url.sh localhost:8001/__heartbeat__
+# ensure content-server is ready
+_scripts/check-url.sh localhost:3030/bundle/app.bundle.js
+
+cd packages/fxa-content-server
+mozinstall /firefox.tar.bz2
+node tests/intern.js --suites="settings_v2" --output="../../artifacts/tests/results.xml" --firefoxBinary=./firefox/firefox
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,20 @@ commands:
           path: ~/screenshots
           destination: screenshots
 
+  test-settings-server:
+    steps:
+      - base-install:
+          package: fxa-settings
+      - run:
+          name: Running test...
+          command: ./.circleci/test-package.sh fxa-settings
+      - store_artifacts:
+          path: ~/.pm2/logs
+          destination: logs
+      - store_artifacts:
+          path: ~/screenshots
+          destination: screenshots
+
 jobs:
   test-package:
     resource_class: medium+
@@ -106,7 +120,6 @@ jobs:
             'fxa-react' \
             'fxa-admin-server' \
             'fxa-graphql-api' \
-            'fxa-settings' \
             'fxa-admin-panel' \
             'fxa-payments-server' \
             'fxa-support-panel' \
@@ -128,6 +141,11 @@ jobs:
       - run:
           name: Reporting code coverage...
           command: bash <(curl -s https://codecov.io/bash) -X gcov
+
+  test-settings-server:
+    executor: content-server-executor
+    steps:
+      - test-settings-server
 
   test-content-server-0:
     executor: content-server-executor
@@ -249,6 +267,12 @@ workflows:
   test_pull_request:
     jobs:
       - test-many:
+          filters:
+            branches:
+              ignore: main
+            tags:
+              ignore: /.*/
+      - test-settings-server:
           filters:
             branches:
               ignore: main
--- a/packages/fxa-content-server/tests/functional/recovery_key.js
+++ b/packages/fxa-content-server/tests/functional/recovery_key.js
@@ -120,6 +120,11 @@ registerSuite('Recovery key', {
           // enter old key
           .then(fillOutRecoveryKey(recoveryKey))
           .then(
+            testElementExists(
+              selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP
+            )
+          )
+          .then(
             testElementTextInclude(
               selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP,
               'invalid'
@@ -159,6 +164,11 @@ registerSuite('Recovery key', {
 
           // enter invalid recovery key
           .then(fillOutRecoveryKey('N8TVALID'))
+          .then(
+            testElementExists(
+              selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP
+            )
+          )
           .then(
             testElementTextInclude(
               selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP,
--- a/packages/fxa-content-server/tests/functional/settings_v2/settings.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/settings.js
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const intern = require('intern').default;
+const { describe, it, beforeEach } = intern.getPlugin('interface.bdd');
+const selectors = require('../lib/selectors');
+const FunctionalHelpers = require('../lib/helpers');
+
+const config = intern._config;
+const EMAIL_FIRST = config.fxaContentRoot;
+const SETTINGS_V2_URL = `${config.fxaContentRoot}beta/settings`;
+const password = 'passwordzxcv';
+
+const {
+  createEmail,
+  createUser,
+  fillOutEmailFirstSignIn,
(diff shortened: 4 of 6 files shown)


packages/fxa-settings/e2e/functional/settings.js
```diff
@@ -0,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const intern = require('intern').default;
```
LZoog, line 7: Is it possible to convert this/any others into es6 module files so we don't have to use `require`?

packages/fxa-settings/e2e/functional/settings.js
```diff
@@ -0,0 +1,41 @@
...
+const { describe, it, beforeEach } = intern.getPlugin('interface.bdd');
+const FunctionalHelpers = require('../../../fxa-content-server/tests/functional/lib/helpers');
```
LZoog, line 9: I think we'd need to add `fxa-content-server` as a dependency to settings for this to work in CI. Maybe we want to create our own helper file for settings and only pull in helpers when we need to use them instead of all of them? Could give us an opportunity to clean them up to use async/await too and only use what we need, if we took that approach (since we'll _eventually_ get around to the signin/registration flow) we'll have a nice cleaned up helper file by the end of it.

packages/fxa-settings/e2e/intern.js
```diff
@@ -0,0 +1,111 @@
...
+
+intern.on('afterRun', () => {
+  if (failed.length) {
+    fs.writeFileSync('rerun.txt', failed.map((f) => f.name).join('|'));
+  }
+});
+
+intern.configure(config);
+intern.run().catch((e) => {
+  // This might not throw, BUG filed: https://github.com/theintern/intern/issues/868
```
LZoog, line 108: Looks like this bug was fixed!

packages/fxa-settings/e2e/intern.js
```diff
@@ -0,0 +1,111 @@
...
+// args.bailAfterFirstFailure comes in as a string.
+const bailAfterFirstFailure = args.bailAfterFirstFailure === 'true';
+
+const testProductId = args.testProductId || 'prod_GqM9ToKK62qjkK';
+
+// Intern specific options are here: https://theintern.io/docs.html#Intern/4/docs/docs%2Fconfiguration.md/properties
+const config = {
+  asyncTimeout: asyncTimeout,
+  bail: bailAfterFirstFailure,
+  defaultTimeout: 45000, // 30 seconds just isn't long enough for some tests.
```
LZoog, line 38: I kind of wonder if we should lower this, and when we have to increase it, we'll know which tests are causing the uppage?

packages/fxa-settings/e2e/tools/firefox_profile_creator.js
```diff
@@ -0,0 +1,127 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var UA_OVERRIDE =
```
LZoog, line 5: This is probably a good opportunity to change these all to `const`/`let`

packages/fxa-settings/e2e/tools/firefox_profile_creator.js
```diff
@@ -0,0 +1,127 @@
...
+if (process.argv.length > 1) {
+  try {
+    profile = JSON.parse(process.argv[2]);
+  } catch (e) {
+    throw new Error('Failed to parse args');
+  }
+}
+
+if (profile) {
+  if (profile.isTestingPairing) {
```
LZoog, line 21: Should we typescript-ify all/any of these files? Even small wins like `if (profile?.isTestingPairing) {` are nice (albeit small) wins.

packages/fxa-settings/e2e/functional/settings.js
```diff
@@ -0,0 +1,41 @@
...
+    email = createEmail();
+    await remote.then(createUser(email, password, { preVerified: true }));
+  });
+
+  it('can navigate to settings', async ({ remote }) => {
+    await remote
+      .then(openPage(EMAIL_FIRST, selectors.ENTER_EMAIL.HEADER))
+      .then(fillOutEmailFirstSignIn(email, password))
+      .then(testElementExists(selectors.SETTINGS.HEADER))
+      .then(openPage(SETTINGS_V2_URL, '#profile'));
```
LZoog, line 39: Can these `.then`s be converted to `await`s?

6a68 (commented):
LGTM so far, I think it makes sense to move things into fxa-shared that are shared across packages, but also OK to wait on that until the tests are further along.

packages/fxa-settings/e2e/functional/settings.js
jodyheavener: I'm for creating our own helper file and copying over what we need when we need it; there are a _lot_ of helpers in there and I am certain that there are some where cleaning them up or rewriting them altogether could be beneficial. It also allows us to make Intern upgrades to keep fxa-settings modern without breaking anything in content-server.

packages/fxa-settings/e2e/tools/firefox_profile.js
```diff
@@ -0,0 +1,31 @@
...
+
+  let encodedProfile = '';
+  const profileArgs = JSON.stringify(JSON.stringify(config));
+  const profileTool = path.join('e2e', 'tools', 'firefox_profile_creator.js');
+  try {
+    encodedProfile = childProcess.execSync(
+      ['node', profileTool, profileArgs].join(' ')
+    );
+  } catch (e) {
+    console.log('Note: execSync failed to run:', e);
```
jodyheavener, line 19: Assuming setting this is critical to test run, shouldn't we let this exception happen (instead of just console.logging)?

packages/fxa-settings/e2e/tools/firefox_profile_creator.js
```diff
@@ -0,0 +1,127 @@
...
+    // pairing UA override, these preferences can be moved to the rest
+    // once the main tests catch up with the user agent version
+    UA_OVERRIDE =
+      'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:67.0) Gecko/20100101 Firefox/67.0 FxATester/1.0';
```
jodyheavener, line 25: Time for a UA update? 😇 Here's mine: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:81.0) Gecko/20100101 Firefox/81.0`

packages/fxa-settings/scripts/test-ci.sh
```diff
@@ -3,3 +3,36 @@
...
+    --topological-dev \
+    --include fxa-auth-db-mysql \
+    --include fxa-auth-server \
+    --include fxa-content-server \
+    --include fxa-payments-server \
+    --include fxa-profile-server \
+    --include fxa-react \
+    --include fxa-settings \
+    --include fxa-shared \
+    --include fxa-support-panel \
```
jodyheavener, line 26: Do all of these need to be running for these tests to run?

jodyheavener (commented):
So happy this is coming along. I second Lauren's thoughts about modernizing some of this; using async/await, let/const, and import/export across the board, as well as taking this opportunity to start fresh with TypeScript, would be so killer.

vbudhram:
@LZoog @jodyheavener @6a68 Thanks for the feedback! I've updated this based on my findings, mostly just having the tests live in content-server since the internjs test plumbing couldn't be extract easily from content-server without major rework.

The async/await update will come in the follow up PR for testing.

jodyheavener (commented):
> The async/await update will come in the follow up PR for testing.

Is there a ticket for this? I don't want this to accidentally fall to the backburner 🙈

vbudhram:
@jodyheavener 

> Is there a ticket for this? I don't want this to accidentally fall to the backburner 🙈

I gotcha in https://github.com/mozilla/fxa/issues/6530

packages/fxa-settings/scripts/test-ci.sh
```diff
@@ -1,5 +1,31 @@
...
+    --include fxa-auth-db-mysql \
+    --include fxa-auth-server \
+    --include fxa-content-server \
+    --include fxa-profile-server \
+    --include fxa-react \
+    --include fxa-settings \
+    --include fxa-shared \
+    --include fxa-graphql-api \
+    run start > ~/.pm2/logs/startup.log
+npx pm2 ls
```
jodyheavener, line 23: This just outputs the processes, is this just for logging purposes? Do we need this?

packages/fxa-settings/scripts/test-ci.sh
```diff
@@ -1,5 +1,31 @@
...
+# ensure email-service is ready
+_scripts/check-url.sh localhost:8001/__heartbeat__
```
jodyheavener, line 25: I ran `circleci local execute -c .circleci/local.yml --job test-settings-server` locally and it hung here. I'm assuming this is a local-only thing that I'm not doing. Just wanted to note.

jodyheavener (approved):
Looks good! Couple comments. R+ but this is probably worth a note in the `fxa-settings` README about the existence of these tests and how one can (just) run them locally. Want to make sure we keep up with documentation 😄

Also, for the future, this could use more discussion and documentation, but I'd like to propose that we use `data-testid` over `id` selectors when writing functional tests where possible.

packages/fxa-settings/scripts/test-ci.sh
vbudhram: I've never been able to run `circleci local execute` so probably not the best person to answer 😬

packages/fxa-settings/scripts/test-ci.sh
vbudhram: I found it useful to verify that things are started as expected.

jodyheavener: No harm done! 👍

vbudhram:
@jodyheavener Documentation coming up once I have more of the tests flushed out in blocking issues for https://jira.mozilla.com/browse/FXA-1558

PR #6314, opened 2020-08-26
feat(payments): use existing card for new subscriptions

chenba:
Because:
 - a current subscriber should be able to use the credit card on record
   to subscribe to additional products

This commit:
 - display existing card brand and last four digit of card number
   instead of card form fields
 - update endpoint payload validation to make some new card related
   properties optional
 - update storybook to demo the UI with an existing card

Closes: #FXA-2290

## Checklist


packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -146,14 +172,23 @@ export const PaymentForm = ({
...
-      onSubmit={onSubmit}
-      className="payment"
-      {...{ onChange }}
-    >
+  const paymentSource = hasExistingCard ? (
+    <div className="card-details" data-testid="card-details">
+      <Localized
+        id="sub-update-card-ending"
+        vars={{
+          last: (customer as Customer).last4!,
```
chenba, line 180: `hasExistingCard` contains a check for `customer` and `last4` so this assertion should be ok.

Also, I don't know why it's `last` and not `last4` where, but I did check the ftl file and it is `last` in there.

packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.tsx
```diff
@@ -149,7 +149,7 @@ export const SubscriptionUpgrade = ({
               <Localized
                 id="payment-confirmation-cc-preview"
                 vars={{
-                  last4: cardLast4
+                  last4: cardLast4 as string,
```
chenba, line 152: This is an upgrade, so the card should already exist.

packages/fxa-payments-server/src/routes/ProductV2/SubscriptionCreate/index.tsx
```diff
@@ -232,13 +233,38 @@ async function handleSubscriptionPayment({
...
+      }
+    );
+    return handlePaymentIntent({
+      invoiceId: createSubscriptionResult.latest_invoice.id,
+      paymentIntentStatus:
+        createSubscriptionResult.latest_invoice.payment_intent.status,
+      paymentIntentClientSecret:
+        createSubscriptionResult.latest_invoice.payment_intent.client_secret,
+      paymentMethodId:
+        createSubscriptionResult.latest_invoice.payment_intent.payment_method,
```
chenba, line 253: Thankfully this already exists...

packages/fxa-payments-server/src/store/types.tsx
```diff
@@ -67,15 +67,15 @@ export interface CustomerSubscription {
...
+export type Customer = {
+  billing_name?: string | null;
+  brand?: string;
+  exp_month?: string;
+  exp_year?: string;
+  last4?: string;
+  payment_type?: string;
   subscriptions: Array<CustomerSubscription>;
-}
+};
```
chenba, line 78: This matches the endpoint response validator schema.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
lmorchard: It's fun to be inconsistent I guess?

packages/fxa-payments-server/src/components/PaymentFormV2/index.tsx
```diff
@@ -144,15 +158,23 @@ export const PaymentForm = ({
...
+  const paymentSource = hasExistingCard ? (
+    <div className="card-details" data-testid="card-details">
+      <Localized
+        id="sub-update-card-ending"
+        vars={{
+          last: (customer as Customer).last4!,
+        }}
+      >
+        <div
+          className={`new-sub c-card unbranded ${customer?.brand?.toLowerCase()}`}
```
lmorchard, line 170: Should this be something more like this? `${customer?.brand?.toLowerCase() || 'unbranded'}` 

Probably answering my own question: I think the more specific / later-defined CSS selectors win here, so it doesn't matter too much.

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -207,6 +212,7 @@ export const Product = ({
         <SubscriptionCreate
           {...{
             profile: profile.result,
+            customer: customer.result ?? undefined,
```
lmorchard, line 215: Oh fancy, I forgot `??` was a thing now!

PR #6315, opened 2020-08-26
feat(event-broker): migrate to NestJS

### LINKED ISSUE #6238: Convert event-broker to use Nest.js
Use NestJS for event-broker and update appropriately.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2444)
┆Issue Number: FXA-2444

bbangert:
Because:

* We are using NestJS for application architecture on new projects.

This commit:

* Moves the event-broker to a NestJS architecture.

Closes #6238

## Checklist


packages/fxa-admin-server/pm2.config.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
         NODE_OPTIONS: '--inspect=9150',
         TS_NODE_TRANSPILE_ONLY: 'true',
         TS_NODE_FILES: 'true',
-        PORT: '8090', // TODO: this needs to get added to src/config.ts
+        PORT: '8095', // TODO: this needs to get added to src/config.ts
```
bbangert, line 23: event-broker had 8090 first, so I'm moving admin-server to 8095 here.

packages/fxa-event-broker/package.json
```diff
@@ -3,13 +3,18 @@
...
-    "watch": "tsc -w",
-    "test": "mocha -r ts-node/register src/test/**/*.spec.ts src/test/**/**/*.spec.ts",
-    "start-prod": "npm run build && node ./dist/bin/worker.js",
+    "prebuild": "rimraf dist",
+    "build": "nest build",
+    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
     "start": "../../_scripts/check-url.sh localhost:9000/__heartbeat__ && pm2 start pm2.config.js",
+    "start:prod": "node dist/main",
+    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+    "test": "jest",
```
lmorchard, line 13: We're moving to Jest for testing here? [I thought we were only using Jest for front-end React stuff?](https://github.com/mozilla/fxa/blob/main/docs/adr/0009-testing-stacks.md#identify-a-consistent-stack-of-testing-tools-per-aspect-of-the-project) 

Assuming that's changing along with NestJS, we might want to draft an ADR to obsolete 0009 and explain why

bbangert: Yea, this came up in the NestJS ADR, I agree it should be formalized as a new ADR. Danny mentioned it as, "We should stick with NestJS defaults" on the architecture ADR, which is the brunt of it. I think that also implies a formalization of using Express, as that's the default web server used by NestJS as well.

lmorchard: Could even edit / expand the existing NestJS ADR to explicitly say "stick with NestJS defaults" means Jest & Express.

Normally, I'd say a new ADR is better. But, if these things were already understood & implicit in the discussion of 0020, it's worth just elaborating in the existing ADR. Then maybe link from 0009 as it being obsoleted by 0020.

I thought I remembered us drafting an ADR around Hapi vs Express, but doesn't seem like we did that. Otherwise, I'd say that could use a link too

bbangert: I'll update the ADR with notes to use the NestJS default of Express+Jest.

packages/fxa-event-broker/src/queueworker/queueworker.service.ts
```diff
@@ -0,0 +1,284 @@
...
+    private firestore: FirestoreService,
+    private clientCapability: ClientCapabilityService
+  ) {
+    const env = this.configService.get<string>('env');
+    this.disableQueueWorker =
+      this.configService.get<boolean>('disableQueueWorker') ?? false;
+    if (env === 'development') {
+      AWS.config.update({
+        accessKeyId: 'fake',
+        ['endpoint' as any]: 'localhost:4100',
```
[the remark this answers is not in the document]
bbangert: I'm trying to remember, but the underlying thing here is that `endpoint` is not listed in the typings for this, but it *is* a valid option to pass in. This forces TypeScript to allow it by effectively removing the type-checking of this arg.

PR #6321, opened 2020-08-27
feat(settings): strip flow params from url after they've been stored

jodyheavener:
⚠️ Please don't merge this yet. I would like discussion around it first.

## Because

- We don't really need flow params sticking around after the app stores them

## This pull request

- Use `replaceState` to strip them from the URL, leaving non-flow params intact

## Checklist

## Other information

This does what it says on the tin, and it's nice to not have those noisy params hanging around. But I have noticed it becomes incredibly annoying. Any time you refresh the page it loses the stored params and needs to perform the round-trip again. That means in development, when you hit save in a file, it reloads the page automatically, which triggers the round-trip. I would maybe consider disabling this stripping in development?

### DIFF
--- a/packages/fxa-settings/src/lib/metrics.test.ts
+++ b/packages/fxa-settings/src/lib/metrics.test.ts
@@ -101,6 +101,7 @@ afterEach(() => {
 describe('init', () => {
   beforeEach(() => {
     window.location.replace = jest.fn();
+    window.history.replaceState = jest.fn();
     window.console.error = jest.fn();
   });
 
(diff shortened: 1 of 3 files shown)


LZoog:
Danny and I briefly discussed this here: https://mozilla.slack.com/archives/G0177C5FTCZ/p1597878001087900

If the user is on a page and refreshes, I'm pretty sure they'd need to be redirected to `get_flow` then redirected back to that page. Maybe this is OK, but the user would then (I think?) be given a different flow ID - also not sure if this is alright or not.

jodyheavener:
> If the user is on a page and refreshes, I'm pretty sure they'd need to be redirected to get_flow then redirected back to that page.

Correct. "Any time you refresh the page it loses the stored params and needs to perform the round-trip again"

Additionally, this introduces a development consideration:

> That means in development, when you hit save in a file, it reloads the page automatically, which triggers the round-trip. I would maybe consider disabling this stripping in development?

---

It's unclear to me if this is the best approach, or if we can find another way. Right now, at the very least, it's not harming anything to keep them in place so I'll close this and we can always re-open if we'd like to go ahead with it or something similar.

Issue #6316, opened 2020-08-26
Make React hook for setting and revealing alert bar


data-sync-user:
We should have a hook that encapsulates our current usage of `useBooleanState` to hold and set the visibility state of an AlertBar, and also sets the contents (a string or ReactNode) and type (success, info, error) of the AlertBar.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2481)
┆Issue Number: FXA-2481

data-sync-user:
➤ Jody Heavener commented:

Related convo: https://github.com/mozilla/fxa/pull/6292#discussion_r477404650 ( https://github.com/mozilla/fxa/pull/6292#discussion_r477404650 )

PR #6322, opened 2020-08-27
feat(settings): add useAlertBar hook

jodyheavener:
## Because

We have a handful of hooks and functions that make up AlertBar usage

## This pull request

Consolidates the various hooks and functions we use with AlertBar into a single hook.

Now you can use the `useAlertBar` hook in combination with the `AlertBar` component to...

- Hide, show, and check the visibility of the alert bar
- Set the type and contents of the alert bar
- Use the `success`, `info`, and `error` convenience methods to do the above much quicker

Closes: #6316

## Checklist

### DIFF
--- a/packages/fxa-settings/src/lib/hooks.tsx
+++ b/packages/fxa-settings/src/lib/hooks.tsx
@@ -2,9 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, useState, ReactNode, useCallback } from 'react';
 import sentryMetrics from 'fxa-shared/lib/sentry';
 import { useMutation, DocumentNode, MutationHookOptions } from '@apollo/client';
+import { useBooleanState } from 'fxa-react/lib/hooks';
+import { AlertBarType } from '../components/AlertBar';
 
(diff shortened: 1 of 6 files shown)


packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -83,3 +85,59 @@ export function useHandledMutation(
...
+  defaultType = 'success',
+  defaultContent,
+}: {
+  defaultVisible?: boolean;
+  defaultType?: AlertBarType;
+  defaultContent?: string | ReactNode;
+} = {}) {
+  const [visible, show, hide] = useBooleanState(defaultVisible);
+  const [type, setType] = useState<AlertBarType>(defaultType);
+  const [content, setContent] = useState<string | ReactNode | null>(
```
dannycoates, line 100: ReactNode includes both string and null in it's typedef

jodyheavener:
I'll wait until #6351 is merged so I can update the docs to mention this hook and document its usage.

meandavejustice (approved):
Such a nice cleanup. I'm approving but sounds like you wanted to update the docs on this before merging.

LZoog:
Heh, apparently my review was left "pending" from last week, but my comments are now out of date.

I did want to say - YAAAS this is a great improvement. Get those doc edits in since what we've got under AlertBar will now be out of date and then we'll be g2g. (If you merge before then... 
image)

Issue #6317, opened 2020-08-26
CircleCI Error: browserName


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2483)
┆Issue Number: FXA-2483

data-sync-user:
➤ Jody Heavener commented:

Possibly helpful: https://github.com/theintern/intern/issues/874#issuecomment-493637873

PR #6341, opened 2020-08-28
chore(payments): clean up routes to just use SCA payment flow

### LINKED ISSUE #5928: Replace previous routes with "/v2" routes on payments-server front-end
In order to avoid having to work off in a feature branch, we created some new routes in the payments-server front-end app so that old & new implementation could be deployed in parallel temporarily. (e.g. `/v2/products/:productId` to reach `src/routes/ProductV2/index.tsx`)

A final clean-up task will be to take that "V2" code, copy it over on top of the old code, and merge those changes to replace the old routes entirely.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2275)
┆Issue Number: FXA-2275

lmorchard:
- also update content-server test helper to use the new all-in-one card
  Stripe element

fixes #5928


lmorchard:
Starting a draft to see how CI handles this. Looks like a lot of changes, but most of it is just literally copying the V2 stuff on top of the V1 and a few tweaks to remove V1 tests.

chenba:
💪 Functional test updated and passing.

I noticed that the `PaymentForm*` files aren't involved.

The large diffs seem to be mostly from v2 code in v1 files, so I wonder if it might be easier to review if we have a commit that's only v1 file deletions (and path name updates?).  And then another that's mainly renaming v2 files.  But, you already did all this work, so it's probably not worth the effort.

Anyways, LGTM so far!

lmorchard:
>I noticed that the PaymentForm* files aren't involved.

@chenba They *should* be involved - looks like they are to me? https://github.com/mozilla/fxa/pull/6341/files#diff-6d917e9408864fef370fc7b4205d29fb

chenba:
I meant `src/components/PaymentForm/index.tsx`, `src/components/PaymentFormV2/index.tsx`, etc.

lmorchard:
> I meant src/components/PaymentForm/index.tsx, src/components/PaymentFormV2/index.tsx, etc.

Oh - doh! Yep, I missed those!

lmorchard:
> The large diffs seem to be mostly from v2 code in v1 files, so I wonder if it might be easier to review if we have a commit that's only v1 file deletions (and path name updates?). And then another that's mainly renaming v2 files. But, you already did all this work, so it's probably not worth the effort.

@chenba Gave this a rebase and tried untangling the deletes and the renames. Still looks ugly in the combined diff, but the separate commits might be easier to review

lmorchard:
Whoops, lost the content-server functional test fix. Re-added that in a 3rd commit

chenba (approved):
Thanks for the multiple commits.  It was easier.  👍

Squash and win! 🎉

lmorchard:
Squashed and rebased against main, let's see if it actually passes CI. I might merge without green CI though since I think the failure is an unrelated known bug https://github.com/mozilla/fxa/issues/6317

Issue #6338, opened 2020-08-28
Extract direct db access from gql-api


data-sync-user:
To simplify db access in FxA projects, we are going to stop using stored procedures and use knex db methods directly. Some basic models exist in fxa-admin-server and fxa-graphql-api.

The models ([https://github.com/mozilla/fxa/tree/main/packages/fxa-graphql-api/src/lib/db/](https://github.com/mozilla/fxa/tree/main/packages/fxa-graphql-api/src/lib/db)) from fxa-graphql-api are more full featured and should be extracted to fxa-shared.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2496)
┆Issue Number: FXA-2496

PR #6344, opened 2020-08-31
refactor(db access) - Extract direct db access from fxa-graphql-api to fxa-shared

jackiemunroe:
## Because

- We want to get away from more stored procedures

## This pull request

- Extracts direct db access out of fxa-graphql-api and moved it to fxa-shared

Closes: #6338

## Checklist


meandavejustice:
oof-marone

This is a big pr! happy to see this happening, ping me when ur ready for review so we can get a couple eyes on it

packages/fxa-shared/config.ts
```diff
@@ -0,0 +1,211 @@
...
+const configDir = path.dirname(path.dirname(__dirname));
+let envConfig = path.join(configDir, 'config', `${conf.get('env')}.json`);
+envConfig = `${envConfig},${process.env.CONFIG_FILES || ''}`;
+const files = envConfig.split(',').filter(fs.existsSync);
+conf.loadFile(files);
+conf.validate({ allowed: 'strict' });
+const Config = conf;
+
+export type AppConfig = typeof Config;
+export default Config;
```
bbangert, line 211: I think you moved too much in this file. I was thinking about this earlier today as well, and I think for shared components like this, what makes sense would be to have the shared component include either the config or config generator that the package using it should import and use in its local config.

For example, in this case, the docs for the `fxa-shared-db/lib/db` (a README in its dir), could say something like:

> To use the auth db in your package, import `makeMySQLConfig` from `fxa-shared-db/lib/db/config`, and put it in your packages convict setup. Then pass the config into `setupAuthDatabase`.

So only the bits needed to ease proper use would be included.

packages/fxa-shared/lib/constants.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Token } from 'typedi';
+import { AppConfig } from '../config';
+import { Logger } from 'mozlog';
+import { RedisConfig } from '../config';
+import AuthClient from 'fxa-auth-client';
+
+export const configContainerToken = new Token<AppConfig>();
+export const loggerContainerToken = new Token<Logger>();
+
+export const fxAccountClientToken = new Token<AuthClient>();
```
bbangert, line 14: These aren't relevant to the database.

packages/fxa-shared/lib/constants.ts
```diff
@@ -0,0 +1,15 @@
...
+export const authRedisConfig = new Token<RedisConfig>();
```
bbangert, line 15: Oh, neither is this. The database actually doesn't need any container setup.

jackiemunroe:
replaced by https://github.com/mozilla/fxa/pull/6495

Issue #6327, opened 2020-08-27
Fix default styles for the Modal component


data-sync-user:
https://github.com/mozilla/fxa/pull/6302#discussion_r477502257

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2486)
┆Issue Number: FXA-2486

data-sync-user:
➤ Dave Justice commented:

Jody addressed much of this in https://github.com/mozilla/fxa/pull/6350/files, ( https://github.com/mozilla/fxa/pull/6350/files, ) but will be finished up in pr/6302

PR #6350, opened 2020-09-01
refactor(settings): various cosmetic changes to better reflect designs

jodyheavener:
## Because

- There are a handful of areas in new Settings that could be improved to be more consistent or to adhere more closely to the Figma design.

## This pull request

- Makes various stylistic changes that add a little more polish to the Settings app. These changes are all pretty subtle and may not be noticeable at first. This includes things like:
  - Consistent form button presentation
  - Small spacing improvements to match Figma
  - Updates to make spacing consistent across the various flows that implemented the same elements (think modals, forms)
  - Add transition to base cta class as it's clearly being used manually everywhere lol
  - Adds z-index to header so relatively positioned elements will appear under it
  - There's a TODO to uncomment two props once Dave's work in #6302 is merged

## Checklist

### DIFF
--- a/packages/fxa-react/configs/tailwind.js
+++ b/packages/fxa-react/configs/tailwind.js
@@ -51,6 +51,7 @@ module.exports = {
         desktop: '1024px',
         desktopXl: '1441px',
         48: '12rem',
+        64: '16rem',
       },
       inset: {
         '1/2': '50%',
--- a/packages/fxa-settings/src/components/Modal/index.tsx
+++ b/packages/fxa-settings/src/components/Modal/index.tsx
@@ -38,7 +38,7 @@ export const Modal = ({
     <Portal id="modal" {...{ headerId, descId }}>
       <div
         data-testid={testid}
-        className="flex flex-col justify-center fixed inset-0 w-full px-2 h-full bg-black bg-opacity-75"
+        className="flex flex-col justify-center fixed inset-0 z-50 w-full px-2 h-full bg-black bg-opacity-75"
       >
         <div
           data-testid="modal-content-container"
@@ -67,9 +67,9 @@ export const Modal = ({
           <div className="px-4 tablet:px-12 pb-10">
             <div>{children}</div>
             {hasButtons && (
-              <div className="flex mt-6">
+              <div className="flex justify-center mx-auto mt-6 max-w-64">
                 <button
-                  className="cta-neutral transition-standard flex-1"
+                  className="cta-neutral mx-2 flex-1"
                   data-testid="modal-cancel"
                   onClick={onDismiss}
                 >
@@ -78,7 +78,7 @@ export const Modal = ({
 
                 {onConfirm && (
                   <button
-                    className="cta-primary transition-standard flex-1"
+                    className="mx-2 flex-1 cta-primary"
                     data-testid="modal-confirm"
                     onClick={onConfirm}
                   >
--- a/packages/fxa-settings/src/components/ModalVerifySession/index.tsx
+++ b/packages/fxa-settings/src/components/ModalVerifySession/index.tsx
@@ -87,14 +87,14 @@ export const ModalVerifySession = ({
     >
       <form
         onSubmit={(event) => {
-          event.preventDefault();
+          event.preventDefault()
           verifySession({
             variables: {
               input: {
                 code,
               },
             },
-          });
+          })
         }}
       >
         <h2
@@ -106,23 +106,27 @@ export const ModalVerifySession = ({
         <p
           id="modal-verify-session-desc"
           data-testid="modal-desc"
-          className="my-6"
+          className="my-6 text-center"
         >
           Please enter the verification code that was sent to{' '}
           <span className="font-bold">{primaryEmail.email}</span> within 5
           minutes.
         </p>
-        <InputText
-          label="Enter your verification code"
-          onChange={(event) => {
-            setCode(event.target.value);
-          }}
-          {...{ errorText }}
-        ></InputText>
-        <div className="flex mt-6">
+
+        <div className="mt-4 mb-6">
+          <InputText
+            label="Enter your verification code"
+            onChange={(event) => {
+              setCode(event.target.value)
+            }}
+            {...{ errorText }}
+          ></InputText>
+        </div>
+
+        <div className="flex justify-center mx-auto max-w-64">
           <button
             type="button"
-            className="cta-neutral transition-standard flex-1"
+            className="cta-neutral mx-2 flex-1"
             data-testid="modal-verify-session-cancel"
             onClick={onDismiss}
           >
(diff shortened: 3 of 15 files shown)


packages/fxa-react/configs/tailwind.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
         24: '6rem',
       },
       maxWidth: {
+        '1xs': '16rem',
```
LZoog, line 49: I think this should be `64: '16rem'`.

packages/fxa-settings/src/components/AlertExternal/index.tsx
```diff
@@ -8,6 +8,7 @@ export const AlertExternal = () => {
 
   return account.alertTextExternal ? (
     <AlertBar
+      type="info"
```
LZoog, line 11: Shouldn't this actually be success?

packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -67,9 +67,9 @@ export const Modal = ({
           <div className="px-4 tablet:px-12 pb-10">
             <div>{children}</div>
             {hasButtons && (
-              <div className="flex mt-6">
+              <div className="flex justify-center mx-auto max-w-1xs">
```
LZoog, line 70: We might need that `mt-6` class: 
image

Also I think this PR is a partial fix for #6327, I think when @meandavejustice's PR gets merged with the caution buttons we can close it?

packages/fxa-settings/src/components/TextInput/index.tsx
```diff
@@ -56,12 +58,12 @@ export const TextInput = ({
 
   return (
     <label
-      className={`flex items-center rounded transition-all duration-100 ease-in-out border mt-3 mb-3 tooltip
+      className={`flex items-center rounded transition-all duration-100 ease-in-out border tooltip
```
LZoog, line 61: I think this could be cleaned up a bit with `classNames` perhaps?

packages/fxa-settings/src/styles/ctas.scss
```diff
@@ -4,7 +4,7 @@
 
 .cta-neutral,
 .cta-primary {
-  @apply py-2 px-5 rounded w-full inline-block text-center border font-header box-border;
+  @apply py-2 px-5 rounded w-full inline-block text-center border font-header box-border truncate transition-standard;
```
LZoog, line 7: I think moving `transition-standard` here is a good call, but do we want to add `truncate`? I'm not sure we want an ellipsis in this scenario, might be better to break the CTA text onto two lines even if it's not pretty so the user can read what they're doing. I dunno though, I'd say just leave that out for now unless we have a scenario where we need it already?

packages/fxa-react/configs/tailwind.js
jodyheavener: Should it? I was going with the scale we already use as the Tailwind [default](https://tailwindcss.com/docs/max-width). I'm okay changing to 64, just want to make sure it's consistent

packages/fxa-settings/src/components/AlertExternal/index.tsx
jodyheavener: It could be anything, really, since all messages are able to be sent from content server. I was just going the neutral route.

packages/fxa-settings/src/styles/ctas.scss
jodyheavener: Oh I can remove truncate yes!

packages/fxa-react/configs/tailwind.js
LZoog: Ah I see, I'm fine with that too. I was going by what we already have on [this line](https://github.com/mozilla/fxa/blob/41952909aa3fdd02861607d7f6e5be05a774739d/packages/fxa-react/configs/tailwind.js#L54), but that's much smaller than the defaults.

jodyheavener: Oh right, I see we've already started using the other named sizes (mobileLandscape, desktop) - I'll go with 64 here. Good catch

packages/fxa-settings/src/components/AlertExternal/index.tsx
LZoog: Right now, the only messages we convey are the "verified email success" messages, which should be green. I just hit `beta/settings` without an account, created one, and see this: (also, eep, I'm getting a `Attempted to mount AlertBar on element ID "alert-bar-root" but it was not found in the document` error here which is also why it looks a little cut off, if it can't be fixed easily here let's file a follow up?)
image

jodyheavener: I can dig it, will change to success (actually, I'll just remove it to default). Will look at that other thing in the morning.

jodyheavener: Ehh, I've been looking at this for a bit, but don't quite understand how it might be happening. AlertExternal is only ever invoked within the initial AlertBarContent provider set, so I'm not sure why the ref isn't being passed down in this specific context. This is filed here: https://jira.mozilla.com/browse/FXA-2505.

packages/fxa-settings/src/components/ModalVerifySession/index.tsx
```diff
@@ -87,14 +87,14 @@ export const ModalVerifySession = ({
     >
       <form
         onSubmit={(event) => {
-          event.preventDefault();
+          event.preventDefault()
```
LZoog, line 90: Did Prettier remove this semicolon? O.o

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
```diff
@@ -75,17 +75,20 @@ export const UnitRowRecoveryKey = () => {
       </LinkExternal>
       {modalRevealed && (
         <VerifiedSessionGuard onDismiss={hideModal} onError={onError}>
-          {/* TODO customize confirm button text and style */}
           <Modal
             onDismiss={hideModal}
             onConfirm={deleteRecoveryKey}
+            // TODO - Uncomment these two
+            // lines once #6302 is merged
```
LZoog, line 82: Let's comment in that PR about this line so that can be done there as I think this will be merged first

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
```diff
@@ -75,17 +75,20 @@ export const UnitRowRecoveryKey = () => {
...
+            // confirmBtnClassName="cta-caution"
+            // confirmText="Remove"
             headerId="some-id"
             descId="some-desc"
```
LZoog, line 86: headerId/descId here should be something like `recovery-key-header` and `recovery-key-desc` but I know that's not part of your PR 🤷‍♀️

packages/fxa-settings/src/components/ModalVerifySession/index.tsx
jodyheavener: strange 🤔

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
jodyheavener: I can do that.

Issue #6336, opened 2020-08-28
Wrong page is displayed if an error is received at subscription and we refresh the page


ValentinaPC:
#### Affected versions:
- All Firefox;

#### Affected FxA versions:
- 1.186.0;

#### Affected Platforms:
- Windows 10 (64-bit);

#### Prerequisites:
- Have a FxA account with NO subscription;

#### Steps to reproduce:
1. Load the Mozilla VPN main page: https://stage-vpn.guardian.nonprod.cloudops.mozgcp.net/
2. Click on the “Try Mozilla VPN” button;
3. Complete all mandatory fields and use a card that returns an error at subscription i.e. “3566002020360505”
4. Click the “Pay Now” button;
5. After receiving the subscription error, refresh the page;

#### Expected result:
- All fields are empty,the user can re enter valid data.

#### Actual result:
- User is redirected to the “multiple product” subscription page;

#### Notes:
- Any product user wants to subscribe after, returns the same page;
- “Pay now” from multiple product subscription page is also not working and returning the same error as the initial classic subscription form.
![wrong page](https://user-images.githubusercontent.com/15685960/91550422-30d29d80-e931-11ea-9313-99d27ff347f2.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2494)
┆Issue Number: FXA-2494

chenba:
@ValentinaPC If you don't refresh the page after receiving the error, but enter a working card, are you able to subscribe successfully then?

ValentinaPC:
@chenba Tried your suggestion, and did not refreshed the page, and entered a working card. 
User was not redirected to the “multiple product” subscription page, but the error message was still displayed and user was not able to subscribe successfully.
I've used the card number mentioned in the issue STRs and "4242 4242 4242 4242" for the working card number.
![norefresh](https://user-images.githubusercontent.com/15685960/91819187-713a6000-ec3e-11ea-9434-5001dd3800c9.gif)

chenba:
The team agreed that we'd have the frontend send a request to remove the failed payment method when this occurs.

data-sync-user:
➤ Barry Chen commented:

There are really two problems: 1) the retry, and 2) the refresh.

The retry issue can be fixed on the frontend by ignoring an "existing" card on a customer that does not have any subscriptions.  I'll fix this first.

The refresh solution is the one in the previous comment.

data-sync-user:
➤ Barry Chen commented:

[...]

PR #6366, opened 2020-09-02
chore(payments): detach payment method on failed first sub payment

chenba:
Because:
 - a customer is required to enter a different, working, credit card
   after a failed payment during the subscription to their first product,
   so there's no reason to keep the payment method created by the failed
   payment

This commit:
 - have the Payments frontend call the auth-server to detach the payment
   method

Closes: #6336 (FXA-2494)

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
@@ -31,6 +31,7 @@ export type SubscriptionCreateAuthServerAPIs = Pick<
   | 'apiCreateCustomer'
   | 'apiCreateSubscriptionWithPaymentMethod'
   | 'apiRetryInvoice'
+  | 'apiDetachFailedPaymentMethod'
 >;
 
 export type SubscriptionCreateProps = {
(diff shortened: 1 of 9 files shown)


chenba:
Actually, I am not sure how _necessary_ this is.  It's a cleanup for an edge case.  (Functionality was fixed with PR #6353, so the customer is no longer dead-ended.)

But it's still good to keep the bad payment methods to a minimum. 🤷

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -349,6 +361,16 @@ async function handlePaymentIntent({
...
     case 'requires_payment_method': {
+      // If this happens when a customer is trying to subscribe to their first
+      // product, detach the payment method.  The user must enter a working
+      // card to move on; we'll get a working payment method from that.
+      // This does not affect functionality, thus not blocking.
+      if (
+        (!customer || (customer && customer.subscriptions.length === 0)) &&
+        paymentMethodId
+      ) {
+        apiDetachFailedPaymentMethod({ paymentMethodId });
```
lmorchard, line 372: I think there might be a *slim* chance here that we could end up leaving a failed payment method in place - i.e. since we don't have retries, a network error at this point would do it. 

But it's so slim a chance that it's probably just a thing to keep in mind if a customer service issue comes up later.

lmorchard (approved):
LGTM, worst I could come up with is an edge condition that I *think* we already talked through in meetings

Issue #6367, opened 2020-09-02
If the oauth /verify route is passed a JWT with a PPID `sub`, it must not return the real `uid` to the caller


6a68:
Background:

If a JWT access token has a PPID, then the `sub` value in that token will not be a user's real `uid`, but a pseudonymous value. This is intended to allow two different untrusted RPs access to a user's data without enabling them to work together to link the user's activity across their two services. See [our PPID docs]() for more details.

While helping debug #5826, @dannycoates observed that, if the TTL for an access token is greater than 6 hours, the `token.verify` function doesn't return the PPID value as the `user` value (that is, the `sub` value from the token), but instead returns the real user ID value that was stored in the database: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/oauth/token.js#L95.

In contrast, if the TTL is less than or equal to 6 hours, then the PPID is correctly set as the `user` value: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/oauth/token.js#L44

For context, this code path would be called by any auth-server route that uses the `oauthToken` auth strategy, as that strategy [maps to](https://github.com/mozilla/fxa/blob/e440dae101bbd5d90e26dfcf60bea3edde8f362a/packages/fxa-auth-server/lib/server.js#L357-L358) the `auth-oauth` auth scheme, and the `token.verify` function is called by the `auth-oauth` auth scheme: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/auth-schemes/auth-oauth.js#L24-L28 .

The question for this bug is: does any auth/oauth route use this auth strategy and unwittingly return the `info.user` property back to the caller? And if so, then we'll need to modify the code to avoid leaking the real user ID for longer-lived JWTs that use PPIDs.

rfk:
> The question here is whether any auth/oauth route uses this auth strategy

IIUC this function is also exposed directly to the web via the [oauth-server `/verify` route](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/oauth/routes/verify.js), which I think makes it an outright bug regardless of whether any other routes use that auth strategy.

(In practice, I don't think we ever send any long-lived tokens with a PPID to third-party services, meaning there's no risk of a privacy leak *right now*, but we should plug it for future)

IMHO the best way to fix this is to never write tokens with a PPID into the database, so there's no opportunity to get confused about what details should be returned during verification.

rfk:
[...]

PR #6368, opened 2020-09-02
fix(oauth): do not issue JWT access tokens that both use PPIDs and grant access to profile scopes

### LINKED ISSUE #5826: Profile token endpoint errors with PPIDs and shorter TTL
Update: this issue does not block lowering the JWT default TTL down to 6 hours. The problems FPN encountered turned out to be specific to FPN's use case and changes over time (see rfk's comments in #6368 for details). I'll close this when #6368 lands and file followup bugs once I figure out whether there are already bugs for the quirks in FxA's PPID and JWT code.

-----

In the earlier issue https://jira.mozilla.com/browse/FXA-1780, I had planned to fix our daily usage numbers, starting by fixing a validation bug in auth-server that was preventing us from shortening the oauth token durations (also known as lifetime, or time-to-live, abbreviated 'TTL') down to the same six hour duration used in browserid assertions. Sync is migrating from browserid assertions to oauth tokens, so lowering the oauth token lifetimes to a matching six hours would get our daily device numbers back to normal.

Note that clients request these oauth tokens via the profile server, which relays the call through to the auth server. We discovered in production with FPN that, while the bug is fixed in auth server, a bug still exists in profile server: lowering the requested TTL down to 6 hours in profile-server requests throws an error.

So, in order to lower oauth token TTLs down to six hours, we now need to fix this profile-server bug, to restore accurate daily sync device numbers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2212)
┆Issue Number: FXA-2212

6a68:
Refs #5826. [Note: please see the review comments below, as this description is not quite accurate. In particular, enforcing ppid_seed not be passed for a profile grant request is a good edge case to handle, but doesn't fix the underlying issues FPN is encountering.]

## Because

- JWT access tokens are intended for use by any RP, not necessarily a trusted RP. PPIDs provide an extra layer of user protection by concealing the user's FxA uid while still enabling untrusted RPs to access resources owned by that user. In particular, PPIDs prevent two evil RPs from comparing user identifiers and correlating a user's behavior across their two services.

- As [mentioned](https://github.com/mozilla/fxa/blob/0ed71f677637ee5f817fa17c265191e952f5500e/packages/fxa-auth-server/fxa-oauth-server/docs/pairwise-pseudonymous-identifiers.md) in our PPID docs, it's nonsensical to issue a JWT access token that both uses a PPID to conceal the user's information, yet also grants access to the user's profile data, since that may reveal the user's FxA `uid` as well as their email and display name (depending on the precise scope of the grant).

- FPN unintentionally mixed PPIDs and the profile grant request. As a result of an unrelated bug (#6367), the user's real uid was returned when the TTL was greater than 6 hours, but when FPN set the TTL to exactly 6 hours, we attempted to look up the PPID in the accounts database (treating it as a uid, which it is not). This caused the profile lookup to fail, which led to the FPN outage around July 1st.

## This pull request

- Enforces that we not mix PPIDs and profile access tokens by throwing a 400 if the token request both includes a `ppid_seed` (signaling the RP would like to have a PPID in the returned token) and would grant access to any profile-related scope.

Closes: #5826 (but not by fixing the bug, only by discovering that this bug is a PPID bug, not a TTL bug that should affect other RPs)

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/docs/oauth/pairwise-pseudonymous-identifiers.md
+++ b/packages/fxa-auth-server/docs/oauth/pairwise-pseudonymous-identifiers.md
@@ -32,7 +32,7 @@ By default, Firefox accounts does not enforce sub rotation, but for sensitive RP
 
 ## Ensuring 3rd parties are unable to access user information
 
-An OAuth access token is a bearer token that can be presented by anyone to an OAuth service provider to access a protected resource. If a JWT access token contains a PPID `sub` claim that is meant to protect a user's privacy, but is granted the `profile` scope allows the service provider to present the same access token to the FxA profile server and learn the user's true identity. As such, tokens _MUST NOT_ be requested with the `profile` scope or any of its implicants such as `profile:uid` or `profile:email`.
+An OAuth access token is a bearer token that can be presented by anyone to an OAuth service provider to access a protected resource. If a JWT access token contains a PPID `sub` claim that is meant to protect a user's privacy, but is granted the `profile` scope, it will allow the service provider to present the same access token to the FxA profile server, and learn the user's true identity. As such, tokens _MUST NOT_ be requested with the `profile` scope or any of its sub-scopes, such as `profile:uid` or `profile:email`. If an RP requests a JWT access token including both the `ppid_seed` parameter and also a profile-related grant, then that request will be rejected with a 400 error with `errno` 123, and error message "PPID tokens cannot grant profile access".
 
 ## PPIDs and logging
 
--- a/packages/fxa-auth-server/lib/oauth/jwt_access_token.js
+++ b/packages/fxa-auth-server/lib/oauth/jwt_access_token.js
@@ -8,6 +8,8 @@ const jwt = require('./jwt');
 const sub = require('./jwt_sub');
 const { OAUTH_SCOPE_OLD_SYNC } = require('../constants');
 const config = require('../../config');
+const ScopeSet = require('fxa-shared').oauth.scopes;
+const SCOPE_PROFILE_WRITE = ScopeSet.fromString('profile:write');
 const TOKEN_SERVER_URL = config.get('syncTokenserverUrl');
 
 const HEADER_TYP = 'at+JWT';
@@ -26,6 +28,14 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
     : grant.scope.contains(OAUTH_SCOPE_OLD_SYNC)
     ? TOKEN_SERVER_URL
     : clientId;
+
+  if (
+    grant.ppidSeed &&
+    grant.scope &&
+    SCOPE_PROFILE_WRITE.intersects(grant.scope)
+  ) {
+    throw AppError.ppidConflict();
+  }
 
   // Claims list from:
   // https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt#section-2.2
--- a/packages/fxa-auth-server/test/oauth/jwt_access_token.js
+++ b/packages/fxa-auth-server/test/oauth/jwt_access_token.js
@@ -118,6 +118,63 @@ describe('lib/jwt_access_token', () => {
 
       assert.equal(signedClaims.aud, TOKEN_SERVER_URL);
     });
+
+    describe('PPID / uid scope tests', () => {
+      beforeEach(() => {
+        requestedGrant.ppidSeed = 1000;
+      });
+
+      const scopeTest = async (scopes) => {
+        requestedGrant.scope = ScopeSet.fromArray(scopes);
+        try {
+          await JWTAccessToken.create(mockAccessToken, requestedGrant);
+          assert.fail();
+        } catch (err) {
+          assert.instanceOf(err, AppError);
+          assert.equal(err.errno, 123);
+        }
+      };
+
+      it('should not allow a PPID token to grant access to a profile:uid scope', async () => {
+        await scopeTest(['profile:uid']);
+      });
+
+      it('should not allow a PPID token to grant access to a profile scope', async () => {
+        await scopeTest(['profile']);
+      });
+
+      it('should not allow a PPID token to grant access to a profile:write scope', async () => {
+        await scopeTest(['profile:write']);
+      });
+
+      it('should not allow a PPID token to grant access to a profile:write sub-scope', async () => {
+        await scopeTest(['profile:avatar:write']);
+      });
+
+      it('should not allow a PPID token to grant access to a compound scope including a profile sub-scope', async () => {
+        await scopeTest([
+          'https://identity.mozilla.com/apps/sync',
+          'profile:avatar',
+        ]);
+      });
+
+      it('should allow a PPID token to grant access to a random non-profile scope', async () => {
+        requestedGrant.scope = ScopeSet.fromArray([
+          'foo',
+          'bar',
+          'https://identity.mozilla.com/apps/sync',
+        ]);
+        try {
+          const token = await JWTAccessToken.create(
+            mockAccessToken,
+            requestedGrant
+          );
+          assert.ok(token);
+        } catch (err) {
+          assert.fail(err);
+        }
+      });
+    });
   });
 
   describe('tokenId', () => {
--- a/packages/fxa-auth-server/docs/oauth/api.md
+++ b/packages/fxa-auth-server/docs/oauth/api.md
@@ -54,6 +54,8 @@ The currently-defined error responses are:
 |     400     |  119  | stale authentication timestamp                |
 |     400     |  120  | mismatch acr value                            |
 |     400     |  121  | invalid grant_type                            |
+|     400     |  122  | unknown token                                 |
+|     400     |  123  | PPID tokens cannot grant profile access       |
 |     500     |  999  | internal server error                         |
 
 ## API Endpoints
--- a/packages/fxa-auth-server/lib/oauth/error.js
+++ b/packages/fxa-auth-server/lib/oauth/error.js
@@ -360,6 +360,15 @@ AppError.unknownToken = function unknownToken() {
   });
 };
 
+AppError.ppidConflict = function ppidConflict() {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: 123,
+    message: 'PPID tokens cannot grant profile access',
+  });
+};
+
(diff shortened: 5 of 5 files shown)


rfk (commented):
Hrm. I like this as a defensive measure and I think we should merge it - if you're supplying a `ppidSeed` and granting `profile` access to the resulting token then you're definitely doing something wrong.

However, I'm concerned that this may not actually fix the underlying problem with FPN (or at least, if it does fix the problem then I don't understand why it does so).

Observe that the decision of whether to generate a ppid is not based on the presence of `ppidSeed`, but rather is [based on the OAuth client id](https://github.com/mozilla/fxa/blob/2f4c5419f6af34af252f1ef8a7137170569c9f27/packages/fxa-auth-server/lib/oauth/jwt_sub.js#L39). The `ppidSeed` gets *mixed in* to the generated id, but will default to zero if not provided by the client. IIUC this means that *every* JWT generated by the FPN OAuth client will use a ppid for the `sub`.

Now, FPN is kinda weird in how it uses OAuth tokens, due to some concurrent evolution of our requirements and theirs. 

One one hand, FPN wants to make [JWT tokens with the "secure-proxy" scope](https://github.com/mozilla-services/guardian-website/blob/a9a68a542fd36c4b308720bb50bcb9106ac5be17/server/browser/utils.js#L236) and pass those to the third-party VPN provider as access tokens. It's very important that these be JWTs (so the provider can verify them locally) and that they include a ppid (for the privacy benefits described in the ppid documentation).

On the other hand, FPN also wants to make [access tokens for reading user profile data](https://github.com/mozilla-services/guardian-website/blob/a9a68a542fd36c4b308720bb50bcb9106ac5be17/server/browser/utils.js#L273) and use those for its own internal user-management purposes. It's very important that this profile data contains the *real* FxA uid, email address, etc in order to support proper user management.

So IIUC what's happening here, FPN is currently relying on a fragile combination of things that happen to work:

* When requesting an access token with `secure-proxy` scope, it uses a [short TTL](https://github.com/mozilla-services/guardian-website/blob/a9a68a542fd36c4b308720bb50bcb9106ac5be17/server/constants.ts#L211). This triggers FxA to generate a JWT access token with a ppid, and to not store it into redis.
  * The third-party VPN provider only ever sees these tokens, and the ppid embedded therein. Even if they pass this JWT to the FxA `/verify` endpoint, we won't leak the real uid because it's not stored in redis.
  * This works as intended, but it relies on the ttl being under a magic undocumented threshold to ensure that the resulting token is a JWT.
* When requesting an access token with `profile` scope for its own use, it uses a [longer TTL](https://github.com/mozilla-services/guardian-website/blob/a9a68a542fd36c4b308720bb50bcb9106ac5be17/server/constants.ts#L189). This triggers FxA to generate a database-backed token, wherein the concept of a ppid does not exist.
  * When FPN uses this token to look up the user's profile data, it gets the real FxA uid and profile data, as intended.
  * This works, but it relies on two magic undocumented behaviors:
      * the ttl being *over* the threshold at which we generated a JWT, meaning the token is created in redis.
      * the fact that ppids only exist in JWT tokens, not in redis, so it can learn the raw uid.

So what happens when we try to make the profile-scoped tokens have a short TTL? In my head it's something like this:

* FPN requests a `profile`-scoped access token with a TTL of 6 hours.
  * Because it's a short-lived token, we generate a JWT rather than writing the token into redis.
  * Because the FPN client_id has PPIDs enabled, we generate a JWT with a ppid rather than the raw FxA uid.
* FPN tries to use the generated token to fetch profile info.
  * The profile-server verifies the token. Because it's a short-lived JWT, verification directly returns the ppid as the uid.
  * The profile-server tries to look up profile info using the ppid as though it were a real uid. This does not go well.

If something like that is what's causing #5826, then unfortunately I don't think this PR will fix it.

(IMHO, this issue highlights a design problem with our current implementation of ppids. The thing we're actually trying to control is not the information available to the RP (which in this case is the first-party FPN service) but rather the information available to the *protected resource* (which in this case is the third-party VPN provider). OAuth doesn't currently have good models for separating those two concerns.)

rfk:
> FPN is kinda weird in how it uses OAuth tokens, due to some concurrent evolution of our requirements and theirs.

I'll add for completeness: this was less weird when FPN was a purely client-side app. FPN is now a hybrid client-server app, which is why it needs the raw profile information to do user management.

A better design for the current state of FPN, would completely eliminate the use of `secure-proxy`-scoped OAuth tokens that get passed to the VPN provider. Instead, FPN would just do an ordinary `profile`-scoped login request with FxA, and would then generate its own JWTs to use for authenticating calls to the VPN provider. There was some work towards this design in https://github.com/mozilla-services/guardian-website/issues/729 and https://github.com/mozilla-services/guardian-website/pull/582 but it looks like it's on hold for now.

Where possible, I would encourage us to support the revival of that work, and get out of this business of sending different user data to the RP than to the protected resource :-/

packages/fxa-auth-server/lib/oauth/jwt_access_token.js
```diff
@@ -27,6 +29,16 @@ exports.create = async function generateJWTAccessToken(accessToken, grant) {
     ? TOKEN_SERVER_URL
     : clientId;
 
+  // Ensure we don't mix PPIDs and 'profile' scope or sub-scopes. See #5826 and
+  // also the PPID docs: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/oauth/pairwise-pseudonymous-identifiers.md#ensuring-3rd-parties-are-unable-to-access-user-information
```
rfk, line 33: For the comment here to be accurate about "don't mix PPIDs and `profile` scope", we would need to forbid the FPN client from ever creating tokens with `profile` scope.

6a68:
Thanks for the review and additional context, @rfk. Your understanding here indeed matches what I've reproduced locally:

> So what happens when we try to make the profile-scoped tokens have a short TTL? In my head it's something like this:
> 
> * FPN requests a `profile`-scoped access token with a TTL of 6 hours.
>       
>   * Because it's a short-lived token, we generate a JWT rather than writing the token into redis.
>   * Because the FPN client_id has PPIDs enabled, we generate a JWT with a ppid rather than the raw FxA uid.
> 
> * FPN tries to use the generated token to fetch profile info.
>      
>   * The profile-server verifies the token. Because it's a short-lived JWT, verification directly returns the ppid as the uid.
>   * The profile-server tries to look up profile info using the ppid as though it were a real uid. This does not go well.

Also, you're correct here:

> If something like that is what's causing #5826, then unfortunately I don't think this PR will fix it.

Testing locally today, I see the `sub` value is indeed still different each time the `profile` access token is issued for FPN with ppid_seed removed, causing the same authentication errors as before.

I guess it's good to know that the TTL issue observed with FPN was just a side effect of our PPID implementation and the way FPN's token usage has evolved. The underlying goal of this work is to enable us to lower the TTL value to 6 hours, so that we can get back to the same resolution cross-device usage metrics that we had when sync used browserid assertions (#5143). I suppose the answer might be to make the default oauth token TTL just long enough to land in the DB: 6 hours plus one second, maybe?

rfk:
> I suppose the answer might be to make the default oauth token TTL just long enough to land in the DB: 6 hours plus one second, maybe?

This sounds like a fine short-term hack, assuming that we're fine with the increased verification traffic because of redis.

packages/fxa-auth-server/docs/oauth/pairwise-pseudonymous-identifiers.md
```diff
@@ -32,7 +32,7 @@ By default, Firefox accounts does not enforce sub rotation, but for sensitive RP
...
+An OAuth access token is a bearer token that can be presented by anyone to an OAuth service provider to access a protected resource. If a JWT access token contains a PPID `sub` claim that is meant to protect a user's privacy, but is granted the `profile` scope, it will allow the service provider to present the same access token to the FxA profile server, and learn the user's true identity. As such, tokens _MUST NOT_ be requested with the `profile` scope or any of its implicants such as `profile:uid` or `profile:email`. If an RP requests a JWT access token including both the `ppid_seed` parameter and also a profile-related grant, then that request will be rejected with a 400 error with `errno` 123, and error message "PPID tokens cannot grant profile access".
```
rfk, line 35: nit: s/implicants/sub-scopes/

(An aside: as the person responsible for introducing it [here](https://github.com/mozilla/fxa/blob/70cd23179af2e6888685f0898b90abc720aa28e9/packages/fxa-shared/oauth/scopes.js#L51), I feel like "implicant" isn't really working out as intended as a piece of terminology, maybe we'd be better off renaming it to "super-scope" or similar, by analogy to "subset" and "superset")

packages/fxa-auth-server/test/oauth/jwt_access_token.js
```diff
@@ -118,6 +118,63 @@ describe('lib/jwt_access_token', () => {
...
+      it('should not allow a PPID token to grant access to a profile scope', async () => {
+        await scopeTest(['profile']);
+      });
+
+      it('should not allow a PPID token to grant access to a profile:write scope', async () => {
+        await scopeTest(['profile:write']);
+      });
+
+      it('should not allow a PPID token to grant access to a profile:write sub-scope', async () => {
+        await scopeTest(['profile:write:avatar']);
```
rfk, line 151: You want `profile:avatar:write` here; my word I dislike how we did the `write` suffix on these initial set of scopes...

rfk (approved):

packages/fxa-auth-server/test/oauth/jwt_access_token.js
6a68: doh, thanks!

packages/fxa-auth-server/docs/oauth/pairwise-pseudonymous-identifiers.md
6a68: heh, I actually considered removing 'implicants' in favor of 'sub-scopes'. Glad to know there's not a more specific shade of meaning implied there. Happy to make the change

6a68: I'd be in favor of replacing implicants with super-scopes. Symmetry of language ftw

6a68:
Reverted this merge until we have a chance to work with the FPN team about their profile token requests that include ppid_seed.

Issue #6231, opened 2020-08-19
Silent failure to update ecosystemAnonId during password reset


rfk:
STR:

* Sign in to Sync on latest Firefox Nightly, which sets up your account to have an `ecosystemAnonId` in its profile data.
* Sign out, go to sign back in, but use the "forgot password" flow.
* When opening the password-reset link, open the network console so you can observe the outgoing requests
* Proceed with the password reset

Expected result:

* The password reset is successful
* The `ecosystemAnonId` field in my profile data is updated with a new value derived from my new `kB`.

Actual result:

* The password reset is indeed successful
* The attempt to update `ecosystemAnonId` fails with a "400 Bad Request", and this failure is silently ignored.
  * The error response says "Header If-None-Match or If-Match is required to perform the request".
  * Inspecting the request headers shows that neither of these headers was sent.

I think the "silently ignored" part is correct behaviour in the case of failure, but we shouldn't be seeing a failure here in the first place.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2440)
┆Issue Number: FXA-2440

jodyheavener:
cc @6a68. If I'm understanding correctly I believe the auth-client call here should be updated to set the `ifMatch` option with the old anon ID?

6a68:
@jodyheavener yup, that sounds right: we want to update the old value to the new value if the old value matches. I guess we should pull this into the current sprint.

data-sync-user:
➤ Chelsea Lewis commented:

[~wclouser@mozilla.com] I'm moving this into the next train but we can bump if necessary. This is required to be fixed prior to 81 hitting release per [~jhirsch@mozilla.com] + [~rkelly@mozilla.com] - I think it would be good to remove the risk by completing within the next 2 sprints.

PR #6379, opened 2020-09-08
feat(aet): set ifMatch option when generating new eco anon id on account model

jodyheavener:
## Because

- During password reset the eco anon id tries to update but is silently failing because the endpoint to update requires the old (current) anon ID be passed along via the `If-Match` header. This ensures that you're only updating the anon ID if it's what you currently know it to be.

## This pull request

- Updates the account model `generateEcosystemAnonId` method to pass in the old anon ID via the `ifMatch` option on `fxaClient.updateEcosystemAnonId`

Closes: #6231

## Checklist


vbudhram (approved):
@jodyheavener LGTM 👍

Issue #6343, opened 2020-08-31
The password is no longer changed after typing an incorrect recovery key


vasilicatamas:
**Build**
Stage [Train 1.186.0]

**Affected platforms**
Windows 10 x64 

**Prerequisites**
The account recovery key is generated.

**Steps to reproduce**
1. Go to Sign In Page and tap on "Forget password?".
2. Insert the email and click on "Begin reset".
3. Go to email and click on "Reset password".
4. Insert an invalid recovery key and click on "Confirm recovery key".
5. Click on "Don't have a recovery key?".
6. Type the new password and click “Reset password” button.

**Expected results**
The password is successfully changed and the “Your password has been reset” screen is displayed.

**Actual Results**
- “Reset password with recovery key” screen is repeatedly displayed.
- The password is not changed.
- The user is not logged in.

![issue8](https://user-images.githubusercontent.com/18114210/91719599-4bf61500-eb9e-11ea-89c0-7a10510b4dea.gif)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2499)
┆Issue Number: FXA-2499

vasilicatamas:
This issue is verified as fixed on FxA Stage [Train 1.187.0] under Windows 10 64-bit. The password is successfully changed and the user logged in.

![issue9](https://user-images.githubusercontent.com/18114210/92928783-4626ec00-f448-11ea-83c0-c347b7e2e7d1.gif)

PR #6380, opened 2020-09-08
fix(recovery): Use the existing accountResetToken if account model has it set

vbudhram:
## Because

- We can only create one `accountResetToken` per password reset. The recovery key flow and password reset flow attempts to create one

## This pull request

- If you start the recovery key flow and generate a `accountResetToken`, store the value in the account model so that it can be reused in the regular password reset flow

Closes: #6343 

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/fxa-client.js
+++ b/packages/fxa-content-server/app/scripts/lib/fxa-client.js
@@ -609,8 +609,13 @@ FxaClientWrapper.prototype = {
         sessionToken: true,
       };
 
-      return client
-        .passwordForgotVerifyCode(code, token, {})
+      return Promise.resolve()
+        .then(() => {
+          if (options.accountResetToken) {
+            return { accountResetToken: options.accountResetToken };
+          }
+          return client.passwordForgotVerifyCode(code, token, {});
+        })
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/app/scripts/lib/fxa-client.js
```diff
@@ -609,8 +609,13 @@ FxaClientWrapper.prototype = {
...
       };
 
-      return client
-        .passwordForgotVerifyCode(code, token, {})
+      return Promise.resolve()
+        .then(() => {
+          if (options.accountResetToken) {
+            return { accountResetToken: options.accountResetToken };
+          }
+          return client.passwordForgotVerifyCode(code, token, {});
```
jodyheavener, line 617: Looks like one of these paths returns a Promise and the other returns an Object. This ok?

jodyheavener (approved):
LGTM with one question

packages/fxa-content-server/app/scripts/lib/fxa-client.js
vbudhram: @jodyheavener Yea this should be ok since its all in the promise chain.

Issue #6219, opened 2020-08-18
Add /__version__ endpoint to 123done


data-sync-user:
We should have an `__version__` endpoint on all our servers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2433)
┆Issue Number: FXA-2433

PR #6381, opened 2020-09-08
feat(123done): add version route

jodyheavener:
## Because

- We should have version endpoints on the 123done server

## This pull request

- Adds basic versioning library to retrieve package.json version, source, and commit hash to 123done (also updates 321done)

Closes: #6219

## Checklist

Screen Shot 2020-09-08 at 12 01 37 PM


vbudhram (approved):
👍

Issue #6347, opened 2020-08-31
auth-server does not emit flow-complete events for Firefox 80


rfk:
In fxa-auth-server, we have a special mechanism for emitting a server-side `flow.complete` event to signal the successful completion of a login flow. I don't think it's well-documented, but it works something like this:

* When we see the user is signing in, we [look at the details of the signin request](https://github.com/mozilla/fxa/blob/a55097f4004a06a9ffae963e6d23c3c795a28b8c/packages/fxa-auth-server/lib/routes/utils/signin.js#L249) and decide what event will signal its successful completion.  This is called the "flow complete signal".
* This gets [stashed in memcache](https://github.com/mozilla/fxa/blob/a55097f4004a06a9ffae963e6d23c3c795a28b8c/packages/fxa-auth-server/lib/routes/utils/signin.js#L260) for later reference alongside other metrics-related data about the login attempt.
* Later, when we're [emitting a metrics event](https://github.com/mozilla/fxa/blob/6f0a4bd9bc9fb319f448e3be66aab789c268e4a3/packages/fxa-auth-server/lib/metrics/events.js#L91), we check whether that event matches the stashed `flowCompleteSignal` and if it does, we emit an additional [`flow.complete` event](https://github.com/mozilla/fxa/blob/6f0a4bd9bc9fb319f448e3be66aab789c268e4a3/packages/fxa-auth-server/lib/metrics/events.js#L122).

Why this elaborate stashing dance, you ask? We want `flow.complete` to signal that the user successfully did the thing they wanted to do. In particular, if the user is signing in to the browser, we want `flow.complete` to signal that the browser successfully obtained the FxA credentials and is attempting use them to sync. That happens in a separate request to `/account/login`, so we need to remember the metrics context across requests.

That's why, when we see a login flow that we think is coming from a signin to Firefox (such as [here](https://github.com/mozilla/fxa/blob/a55097f4004a06a9ffae963e6d23c3c795a28b8c/packages/fxa-auth-server/lib/routes/utils/signin.js#L250) and [here](https://github.com/mozilla/fxa/blob/e440dae101bbd5d90e26dfcf60bea3edde8f362a/packages/fxa-auth-server/lib/routes/password.js#L437)) we set `flowCompleteSignal = "account.signed"`, the even that indicates that the browser is using its FxA credentials to sign a BrowserID certificate in order to start syncing.

As of Firefox 80, this no longer works as intended.

[...]

PR #6383, opened 2020-09-08
fix(metrics: Emit the the `flow.complete` event when a user creates an oauth token for sync

vbudhram:
## Because

- We stopped emitting the `flow.complete` metrics for FF80 which we use in amplitue to signal that the user completed the specific action
- For FF 80 we moved to a sync oauth flow and don't call the `/certificate/signed` anypoint anymore, which would have emitted the `account.signed` event.

## This pull request

- Emits the `account.signed` event when an oauth token is created. If the flow has a `flowCompleteSignal=account.signed`, it will emit the `flow.complete` event.

Closes: #6347 

## Checklist

I will be uplifting this to the current train (186) since its metrics related.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth/index.js
+++ b/packages/fxa-auth-server/lib/routes/oauth/index.js
@@ -299,6 +299,16 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
             uid,
             ecosystemAnonId,
           });
+
+          // This is a bit of a hack, but we emit the `account.signed`
+          // event to signal to the flow it has been completed (see flowCompleteSignal).
+          // Previously, this event would get emitted on the `/certificate/sign`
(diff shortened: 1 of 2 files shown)


rfk (commented):
I think this'll work, but I worry that it blurs the meager abstraction boundaries provided by `./lib/metrics/events.js`. Would it be terrible to just unconditionally emit the `account.signed` event here, rather than inspecting the metrics context?

Please also add a test that the `flow.complete` event was emitted, along the lines of the ones in `test/local/routes/utils/signin.js`.

vbudhram:
> it blurs the meager abstraction boundaries provided by ./lib/metrics/events.js

Ah yea your right, I didn't realize that check is in there. I"ll update this to log it unconditionally.

6a68 (approved):
LGTM

PR #6395, opened 2020-09-09
fix(oauth): lower access token lifetimes to six hours plus five minutes

### LINKED ISSUE #5143: Amplitude user properties `sync_active_devices_*` doesn't count Fx iOS anymore and Fenix, and will break when sync moves to oauth

6a68:
Fixes #5143.

Our goal is to get sync access token lifetimes down to six hours, so
that the cross-device metrics resolution is the same as it previously
was, when sync used browserid assertions (with a six hour lifetime). By
keeping the TTL just above six hours, we ensure the tokens are backed by
redis, preserving existing validation behavior.

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -1062,7 +1062,17 @@ const conf = convict({
       accessToken: {
         doc: 'Access Tokens maximum expiration (can live shorter)',
         format: 'duration',
-        default: '2 weeks',
+        // Warning: here be dragons. 365 minutes is 6 hours plus 5 minutes.
+        // This value is intended to be just slightly larger than the constant
+        // that determines if a JWT is stored in the database or not,
+        // SHORT_ACCESS_TTL_TOKEN_IN_MS (found in lib/constants), currently
+        // 6 hours. We want to keep this FXA_EXPIRATION_ACCESS_TOKEN default
+        // value greater than 6 hours, because tokens not backed by the
+        // database are validated slightly differently (see lib/oauth/token).
+        // Setting this FXA_EXPIRATION_ACCESS_TOKEN config value at or below
+        // SHORT_ACCESS_TTL_TOKEN_IN_MS should be done with caution. See #5143
+        // and the discussion in #6368.
+        default: '365 minutes',
         env: 'FXA_EXPIRATION_ACCESS_TOKEN',
       },
       accessTokenExpiryEpoch: {


6a68:
Hmm. Now that I look at this I'm not sure if we'll need to change some sync configuration in-tree to actually accomplish this goal. @rfk thoughts?

R? @jbuck to make sure we aren't going to knock over redis by increasing traffic. I don't know which of our RPs currently use the default JWT expiration when they request access tokens.

rfk:
> Now that I look at this I'm not sure if we'll need to change some sync configuration in-tree to actually accomplish this goal.

I don't think so, any clients that request a non-default ttl are almost certainly doing so in order to get JWTs, and hence requesting something less than 6 hours already.

packages/fxa-auth-server/config/index.ts
```diff
@@ -1062,7 +1062,7 @@ const conf = convict({
       accessToken: {
         doc: 'Access Tokens maximum expiration (can live shorter)',
         format: 'duration',
-        default: '2 weeks',
+        default: '365 minutes',
```
rfk, line 1065: nit: it may be worth a brief inline comment to explain this weird number, lest our future selves mistake it for a typo :-)

rfk (approved):

packages/fxa-auth-server/config/index.ts
6a68: @rfk added perhaps excessively verbose comment here 🤷 feedback welcome

rfk: Haha, you did not muck around with this one, thanks!

6a68:
I'm going to go ahead and land this. I'll make a note of the change in the deploy doc for the next release.

PR #6396, opened 2020-09-09
fix(totp): Restore unverified session handling in the force-2FA flow

### LINKED ISSUE #6342: Impossible to log in to AMO when 2FA is enforced because of " Authorize this sign-in" page requiring 2FA, which is not set up yet
## Description

When logging in to AMO dev with a FxA account that doesn't have 2FA set up yet, and has 2FA enforcement set on the AMO side, the user is asked to authorize this sign-in before proceeding to the inline 2FA setup (previously it would fail before that with a 500, that was fixed in https://github.com/mozilla/fxa/issues/5206).

However, there is a chicken and egg problem: At this point the user doesn't have 2FA set yet, but when pasting the code received by email in the "Authorize this sign-in" page, the following error is shown: "This request requires two step authentication enabled on your account.".

fxa_2fa_fail

### Steps to reproduce

- Try to log in to AMO dev (https://addons-dev.allizom.org) with a FxA stage account that 1) doesn't have 2FA set up yet and 2) has 2FA enforcement set on AMO (ask an AMO engineer to set it on your account)

### Expected result

- You should log in to FxA as normal, then be presented with the inline flow to set up your second factor before going back to AMO, fully authenticated.

### Actual result

- You see the "Authorize this sign-in" page - which already is confusing, since the user just successfully entered their email and password, but then when pasting the code received by email, the error "This request requires two step authentication enabled on your account." appears and the user is blocked there.

Note: the code entered does appear to be validated properly - entering the wrong code shows "Invalid or expired verification code".

### Environment

What browser and operating system are you using?

Reproduced on Firefox 81.0a1 on Linux.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2498)
┆Issue Number: FXA-2498

6a68:
Fixes #6342.

## Because

- changes in #6295 might have regressed unverified session handling for the force-2FA flow

## This pull request

- restores the redirect behavior by adding the `redirectTo` parameter check back to the `sign_in_token_code` screen

Closes: #6342

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
+++ b/packages/fxa-content-server/app/scripts/views/sign_in_token_code.js
@@ -63,6 +63,14 @@ const View = FormView.extend({
       .then(() => {
         this.logViewEvent('success');
 
+        // Note: `redirectTo` looks similar to `redirectPathname`, but they
+        // work slightly differently: `redirectTo` redirects automatically,
+        // while `redirectPathname` redirects after a form submission.
+        const redirectTo = this.model.get('redirectTo');
+        if (redirectTo) {
+          return (this.window.location.href = redirectTo);
+        }
+
         const redirectPathname = this.model.get('redirectPathname');
         if (redirectPathname) {
           return this.navigate(redirectPathname);


vbudhram (approved):
@6a68 sorry about that! This LGTM 👍

Issue #6312, opened 2020-08-26
[Subscription Emails] 'en_US' is not a structurally valid language tag


chenba:
https://sentry.prod.mozaws.net/operations/auth-prod/issues/9386003/?referrer=github_plugin

```
WError: An internal server error occurred
  File "/fxa/packages/fxa-auth-server/lib/senders/email.js", line 174, in getLocalizedCurrencyString
    return new Intl.NumberFormat(locale, options).format(decimal);
  File "/fxa/packages/fxa-auth-server/lib/senders/email.js", line 314, in Mailer._getLocalizedCurrencyString
    return getLocalizedCurrencyString(amountInCents, currency,
  File "/fxa/packages/fxa-auth-server/lib/senders/email.js", line 1755, in Mailer.subscriptionSubsequentInvoiceEmail
    invoiceTotal: this._getLocalizedCurrencyString(invoiceTotalInCents, invoiceTotalCurrency, message.acceptLanguage),
  File "/fxa/packages/fxa-auth-server/lib/senders/index.js", line 41, in Object.wrappedMailer.<computed> [as sendSubscriptionSubsequentInvoiceEmail]
    return mailer[name]({
  File "/fxa/packages/fxa-auth-server/lib/routes/subscriptions.js", line 747, in DirectStripeRoutes.sendSubscriptionInvoiceEmail
    await this.mailer.sendSubscriptionSubsequentInvoiceEmail(...mailParams);
...
(10 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2480)
┆Issue Number: FXA-2480

data-sync-user:
➤ Barry Chen commented:

Some Sentry events also have 
RangeError: 'chrome://global/locale/intl.properties' is not a structurally valid language tag"
as the exception.

PR #6398, opened 2020-09-09
fix(emails): use default lang tag for sub emails when given invalid tag

chenba:
Because:
 - the Intl module throws an exception when given a language tag that does
   not conform to BCP 47

This commit:
 - handles the exception and use the default 'en-US' instead

Closes: #6312 (FXA-2480)

## Checklist


meandavejustice (approved):
LGTM! 🥳 I like the new code coverage comments also

PR #6414, opened 2020-09-11
fix(event-broker): include lbheartbeat and dev check

bbangert:
Because:

* Development mode is used on a deployed stack that doesn't use a
  local SQS testing instance.
* lbheartbeat was missing from the health routes.

This commit:

* Checks that the SQS URL includes localhost before turning on the
  local SQS dev mode flags.
* Includes a lbheartbeat route.

## Checklist


vbudhram (approved):
@bbangert Makes sense to me 👍

Issue #6411, opened 2020-09-10
Update header ID for post-download page on mozilla.org


6a68:
Looks like the mozilla.org post-download page header is no longer `#download-button-wrapper-desktop`, see this change: https://github.com/mozilla/bedrock/commit/16322320b#diff-0c6a1756770d0c1b653386e2c255d5bdL47 from https://github.com/mozilla/bedrock/pull/9274

To get content-server functional tests working again, we'll need to update the post-download header reference here: https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/tests/functional/lib/selectors.js#L165

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2527)
┆Issue Number: FXA-2527

PR #6421, opened 2020-09-11
fix(event-broker): dont use env for local checks

bbangert:
Because:

* The env var isn't sufficient to indicate whether local emulators
  should be used. Each service has its own config that more clearly
  reflects whether it is running in a local-dev env vs. deployed
  where remote services are available.

This commit:

* Updates the services to inspect the config such that they can
  determine whether to use mock/local emulators or attempt connecting
  to actual remote services.

## Checklist


6a68 (approved):
LGTM for now, seems like we should do something to unify dev environment config globally across the monorepo eventually

6a68:
Skipping content-server CI because of #6411

Issue #5170, opened 2020-04-30
[main] Add refresh buttons to secondary email, recovery key, 2-step auth


data-sync-user:
Conditionally display the refresh buttons for secondary email, recovery key, and 2-step authentication.

On click, send a GQL query request and update the data for that component.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1798)
┆Issue Number: FXA-1798

PR #6423, opened 2020-09-11
feat(settings): add ability to refresh account data via buttons in 2fa, secondary email, and recovery key sections

jodyheavener:
## Because

- We need to be able to refresh the statuses of an account's 2fa, secondary emails, and recovery key
- We have a handful of spots where a tiny button with an icon in it is used

## This pull request

- Add refresh buttons to each unit row to accomplish this.
- Move tiny icon buttons to a component to be re-used.

**Important**: This doesn't actually perform an update query on the specific section, but simply performs an account retrieval and updates the cache accordingly. I could be persuaded to set up individual queries that narrow the columns we look up, but it's all bundled in to one request anyway so 🤷 

Closes: #5170

## Checklist


packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
```diff
@@ -3,9 +3,27 @@
...
 import UnitRowRecoveryKey from '.';
-import { renderWithRouter, MockedCache } from '../../models/_mocks';
+import {
+  renderWithRouter,
+  MockedCache,
+  MOCK_ACCOUNT,
+} from '../../models/_mocks';
+import { GET_ACCOUNT } from 'fxa-settings/src/models';
+
+const mockAccountQuery = (modifications: { [key: string]: any } = {}) => {
```
LZoog, line 15: This might be worth moving into `_mocks` since we do this frequently. It would also be nice if `any` could somehow reflect the account interface 🤔

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
```diff
@@ -41,4 +59,24 @@ describe('UnitRowRecoveryKey', () => {
...
+        mocks={[mockAccountQuery({ recoveryKey: true })]}
+      >
+        <UnitRowRecoveryKey />
+      </MockedCache>
+    );
+    expect(screen.getByTestId('unit-row-header-value')).toHaveTextContent(
+      'Not Set'
+    );
+    await act(async () => {
+      fireEvent.click(screen.getByTestId('recovery-key-refresh'));
```
LZoog, line 76: I think we need account mocks in all 3 stories now to accommodate this. When you click on the refresh button on any of them it errors with "no more mocks available" and stays disabled.

meandavejustice:
Refresh icon is being cut off in storybook

refresh-2fa

meandavejustice (changes requested):
I left a comment about the icon being cutoff, that's the main thing I noticed. I'll let you and @LZoog discuss her suggestions.

I also do not have an opinion on this
>  I could be persuaded to set up individual queries that narrow the columns we look up

jodyheavener:
@meandavejustice I couldn't reproduce the icon being cut off, but I just pushed a change that maybe hopefully fixes it? Can you let me know? It might be a Windows-specific issue

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
jodyheavener: > It would also be nice if any could somehow reflect the account interface

I tried something like `{ [key: keyof Account]: Account[typeof key] }` but I got "An index signature parameter type cannot be a union type. Consider using a mapped object type instead." and my TypeScript bottoms out around there.

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
jodyheavener: I added mocks for account recovery and 2fa, but couldn't for secondary email because it already expects mocks for verification mutations.

LZoog (approved):
LGTM! Works as expected. That `useLazyAccount` is pretty sweet, I was going to suggest mentioning it in the docs but eh, it could be worth doing but I feel what we've got on `useAccount` is sufficient and the name makes it obvious it's the lazy version of `useAccount`.

meandavejustice (approved):
Looks like your change fixed up the svg icon on my side. 👍 

refresh-secondary-email-correct

Issue #5990, opened 2020-07-20
Track completed logins via device pairing


data-sync-user:
**Context**
We are not tracking successful logins done via device pairing. As a result, when device pairing landed on iOS via the FxA Rust component, we started to measure less logins (though we would expect more).

Fenix also has device pairing capabilities and we have the same inability to measure successful logins.

There are already a few issues open on this topic but they all seem much broader which is why none of them likely moved forward. See:

* FXA-232
* FXA-333
* FXA-525

**Request**
Trigger a `fxa_login - complete` event when someone successfully authenticates via device pairing.

Based on the proposed "Change 2" of FXA-333, we should add the event property `login_type` where the value is `qr`.

Fenix is going to production in 2 weeks and so, as of then, we will not be efficiently tracking all of our mobile logins across both platforms.

**Other resources**
Here is a Fenix dashboard to look at pairing (using Fenix client-side telemetry):
https://analytics.amplitude.com/mozilla-corp/dashboard/doctj72

Here is a Firefox for iOS dashboard to look at pairing (using iOS client-side telemetry):
https://analytics.amplitude.com/mozilla-corp/dashboard/3xlucrd

We have some events leading up to a successful pairing like a view of the call to action to pair:
https://analytics.amplitude.com/mozilla-corp/chart/new/174s8yd

**Warning**
Don't get confused with all of the `fxa_qr_connect_device - ` events already in fxa telemetry. Those are mostly all related to the connect another device experiment.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2311)
┆Issue Number: FXA-2311

vbudhram:
We are correctly tracking the login metric, however it not logging the associated service with it.

Ref: https://analytics.amplitude.com/mozilla-corp/chart/zxg4h0g

davismtl:
@vbudhram things are looking better but there is some uncertainty because we're not 100% sure if the logins we're observing are from QR because per request above, this part was not done:
> add the event property login_type where the value is qr.

Can we add the `login_type`?

davismtl:
@vbudhram did you ever add the login_type?

vbudhram:
> @vbudhram did you ever add the login_type?

No that wasn't added. Lets reopen this issue verses the other one since it has more info.

data-sync-user:
➤ Vijay Budhram commented:

Closing this because we won't be using amplitude as much.

PR #6424, opened 2020-09-13
fix(metrics): Emit the service name with pairing login complete

vbudhram:
## Because

- We were not emitting the service name with the `fxa_login - complete` amplitude event from pairing

## This pull request

- Sets the service name for oauth-redirect and oauth-webchannel brokers. In local testing the `oauth-redirect` broker is used, but in production its oauth-webchannel.
- Updates the channel server config to point to a valid server

Closes: #5990

## Checklist

## Notes

I wasn't able to verify this 100% with a real device since building to test apps are a bit difficult at the moment. However, using the functional pairing test, I was able to see an amplitude log with the service set to the oauth clientId, which I think is probably good enough.

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
+++ b/packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
@@ -103,6 +103,12 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
    */
   sendOAuthResultToRelier(result, account) {
     if (this.hasCapability('supportsPairing') || result.action === 'pairing') {
+      if (!this.relier.has('service')) {
+        // the service in the query parameter currently overrides the status message
+        // this is due to backwards compatibility
+        this.relier.set('service', this.relier.get('clientId'));
+        this._metrics.setService(this.relier.get('clientId'));
+      }
       this._metrics.logEvent('pairing.signin.success');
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
```diff
@@ -103,6 +103,12 @@ const OAuthWebChannelBroker = OAuthRedirectAuthenticationBroker.extend({
    */
   sendOAuthResultToRelier(result, account) {
     if (this.hasCapability('supportsPairing') || result.action === 'pairing') {
+      if (!this.relier.has('service')) {
+        // the service in the query parameter currently overrides the status message
+        // this is due to backwards compatibility
+        this.relier.set('service', this.relier.get('clientId'));
+        this._metrics.setService(this.relier.get('clientId'));
+      }
```
chenba, line 111: nit: these few lines are identical to the ones above.  It would be nice to have them in one place.

chenba (approved):
👍

packages/fxa-content-server/app/scripts/models/auth_brokers/oauth-webchannel-v1.js
vbudhram: Since I am not 100% this will fix the above issue, I'll probably wait to clean this up.

PR #6431, opened 2020-09-14
feat(payments): display product name in subscription payment signup / signin flow

### LINKED ISSUE #6121: Display subscription product name as service during signup / signin for payment flows
Currently:

If i enter a subscription flow by going straight to a subscription link i see an fxa flow that tells me to continue to settings....instead we should:

* not suggest the user is looking for sync
* indicate that they are going to the specific product

Before:
!image-2020-08-03-22-19-18-931.png|width=377,height=502!

After:

!image-2020-08-03-22-20-09-318.png|width=358,height=466!

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2381)
┆Issue Number: FXA-2381
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/13763/image-2020-08-03-22-19-18-931.png">image-2020-08-03-22-19-18-931.png</a> | <a href="https://jira.mozilla.com/secure/attachment/13762/image-2020-08-03-22-20-09-318.png">image-2020-08-03-22-20-09-318.png</a>

lmorchard:
- also hides "Looking for Firefox Sync?" when headed to payment
    
fixes #6121

This issue spans a weird slice of FxA auth views and subscription APIs. Opening a draft early before I get tests together, will maybe request some extra eyes on this to see whether this is a terrible hack.

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/reliers/relier.js
+++ b/packages/fxa-content-server/app/scripts/models/reliers/relier.js
@@ -18,8 +18,13 @@ import Constants from '../../lib/constants';
 import ResumeTokenMixin from '../mixins/resume-token';
 import UrlMixin from '../mixins/url';
 import Vat from '../../lib/vat';
+import xhr from '../../lib/xhr';
 
 const t = (msg) => msg;
+
+const SUBSCRIPTION_PRODUCT_ROUTE_RE = new RegExp(
+  '/subscriptions/products/(.*)'
+);
 
 const RELIER_FIELDS_IN_RESUME_TOKEN = [
   'entrypoint',
@@ -91,6 +96,8 @@ var Relier = BaseRelier.extend({
     resetPasswordConfirm: true,
     setting: null,
     serviceName: t(Constants.RELIER_DEFAULT_SERVICE_NAME),
+    subscriptionProductId: null,
+    subscriptionProductName: null,
     style: null,
     uid: null,
     utmCampaign: null,
@@ -101,6 +108,7 @@ var Relier = BaseRelier.extend({
   },
 
   initialize(attributes, options = {}) {
+    this._config = options.config;
     this.isVerification = options.isVerification;
     this.sentryMetrics = options.sentryMetrics;
     this.window = options.window || window;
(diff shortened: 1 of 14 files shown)


lmorchard:
I think this is finally code-complete and tests pass locally, removing draft status. Hopefully CircleCI cooperates.

packages/fxa-content-server/app/scripts/models/reliers/relier.js
```diff
@@ -161,6 +164,28 @@ var Relier = BaseRelier.extend({
...
+    if (!subscriptionProductId) {
+      return Promise.resolve();
+    }
+    const subscriptionProductName = this.get('subscriptionProductName');
+    if (subscriptionProductName) {
+      return Promise.resolve();
+    }
+    const productNameUrl = `${this._config.authServerUrl}/v1/oauth/subscriptions/productname?productId=${subscriptionProductId}`;
+    return this.window
+      .fetch(productNameUrl)
```
chenba, line 178: nit: this probably should use https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/lib/xhr.js for consistency.

packages/fxa-content-server/app/tests/spec/models/reliers/relier.js
```diff
@@ -89,6 +90,77 @@ describe('models/reliers/relier', function () {
...
+    const productNameUrl = `${config.authServerUrl}/v1/oauth/subscriptions/productname?productId=${subscriptionProductId}`;
+
+    let respJson;
+    beforeEach(() => {
+      respJson = sinon.stub().callsFake(() =>
+        Promise.resolve({
+          /* eslint-disable camelcase */
+          product_name: subscriptionProductName,
+          /* eslint-enable camelcase */
+        })
```
chenba, line 108: nit: could just be `sinon.stub().resolves({product_name: subscriptionProductName})`

chenba (approved):
WFM!

I'll see if I get the same test failures locally...

packages/fxa-content-server/app/scripts/models/reliers/relier.js
lmorchard: Oh hey, I hadn't even seen that library!

lmorchard:
Rebased against main and addressed nits. Still not having great luck with content-server tests though 😞

chenba:
The failed CI functional tests are among the failed local ones for me.  So maybe they are legit failure...?  Maybe? Doesn't make sense, but maybe?  🤷

lmorchard:
Can finally reproduce the functional test failures. Going to brain dump what I found because it was all very annoying:

First thing is that I've been using [Maildev](https://github.com/maildev/maildev) to capture mails from my local dev. 

But, functional tests expect to capture mails via the inbox process. Disabling Maildev got the failures properly reproducing for me.

After that, I finally worked out what's causing the problem. I added a `beforeRender` method to `service-mixin` to try to fetch the subscription product name before rendering the service name. Turns out all the views where functional tests are failing *also* define `beforeRender` methods.

That lead me to the [onsi/cocktail docs](https://github.com/onsi/cocktail#what-are-the-return-values-from-colliding-functions):

> What are the return values from colliding functions?

> The return value of the composite function is the last non-undefined return value from the chain of colliding functions.

> To be clear: let's say X mixes in A and B. Say X implements a method foo that returns bar, A implements foo but returns nothing (i.e. undefined is implicitly returned) and B returns baz. Then instances of X will return baz -- the last non-undefined return value from foo's X → A → B collision chain. 

Since the new `beforeRender` method of `service-mixin` was returning a Promise (and not `undefined`), that superseded whatever the View `beforeRender` might return.

So, those View's Promises produced from *their* `beforeRender` methods never got resolved and thus broke UI behavior in apparently unrelated code. Apparently unrelated to me because the magic of Cocktail is how they became related.

lmorchard:
And now even though all tests have passed, I'm doing one more sweep through before merging to decide whether or not I trust the fix. The code still uses `beforeRender` - it just returns `undefined` in cases that don't start from a subscription payment flow. I don't believe many of those cases are represented in functional tests, so I might have just swept the bug under the rug and not fixed it.

Issue #1871, opened 2019-07-22
[content server] Write functional test for expired card


philbooth:
Spun out of #1846 because I don't believe we have the ability to write this test yet.

When a user visits the settings view and their payment card has expired, an message is displayed. There's no functional test coverage for it yet.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-225)
┆Issue Number: FXA-225

PR #6432, opened 2020-09-14
chore(tests): add expired CC subscription functional test

chenba:
Because:
 - let's test to verify expired CC behavior

This commit:
 - add a functional test to verify that an expired CC triggers an error
   message to be shown

Closes: #1871 (FXA-225)

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/subscriptions.js
+++ b/packages/fxa-content-server/tests/functional/subscriptions.js
@@ -21,7 +21,9 @@ const {
   openPage,
   openRP,
   subscribeToTestProduct,
+  subscribeToTestProductWithCardNumber,
   testElementExists,
+  testElementTextInclude,
   visibleByQSA,
 } = FunctionalHelpers;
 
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/tests/functional/subscriptions.js
```diff
@@ -66,5 +68,39 @@ registerSuite('subscriptions', {
         .then(testElementExists(selectors['123DONE'].AUTHENTICATED))
         .then(visibleByQSA(selectors['123DONE'].SUBSCRIBED));
     },
+    'sign up, failed to subscribe due to expired CC': function () {
+      if (
+        process.env.CIRCLECI === 'true' &&
+        !process.env.SUBHUB_STRIPE_APIKEY
+      ) {
+        this.skip('missing Stripe API key in CircleCI run');
```
vbudhram, line 76: Should we throw here instead to avoid any false positives in circle?

vbudhram (approved):

packages/fxa-content-server/tests/functional/subscriptions.js
lmorchard: This would mean that the tests fail for any PR not on the main repo (i.e. from a contributor or one of our personal forks)

Issue #6434, opened 2020-09-15
Update web-push lib, to get newer `aes128gcm` encryption method


rfk:
While filing [a Fenix bug](https://github.com/mozilla-mobile/fenix/issues/15100) I happened to notice that the push messages being sent by FxA are using the legacy pre-standardization `aesgcm` encryption method. Per [conversation with @jrconlin](https://github.com/mozilla-mobile/fenix/issues/14582#issuecomment-686542269) this is not the correct standard to use, and we should be encouraging everyone to update to the newer `aes128gcm` standard.

The web-push library that we use [recently switched to `aes128gcm` as the default](https://github.com/web-push-libs/web-push/pull/404) so updating the version should suffice here.

However, there is a non-trivial risk that this switch will uncover downstream bugs in our push clients, so we should deliberately QA send-tab as part of shipping this update to production.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2532)
┆Issue Number: FXA-2532

PR #6435, opened 2020-09-15
Update web-push lib to latest version.

rfk:
## Because

* We're using a version of the `web-push` module that defaults to
  a legacy encryption scheme.
* The latest release of `web-push` has switched to using a more
  modern default encryption scheme.
* It's good to keep dependencies up to date.

## This pull request

* Updates the version of our `web-push` dependency to the latest
  release.

Closes: #6434

## Checklist

- ~[ ] If applicable, I have modified or added tests which pass locally.~
- ~[ ] I have added necessary documentation (if appropriate).~

As noted in #6434, there's a small but non-zero chance this will uncover bugs in the `aes128gcm` handling in our push client libraries or in their integration with FxA client code. We'll need to keep an eye out for bustage in send-tab push notifications as part of this change, and it seems worth explicitly QA-ing send-tab while this is on stage.

I'm happy to help with that but am a bit out of the loop on the FxA release process these days, so let me know whether/where I should flag ^ for QA attention.

### DIFF
--- a/packages/fxa-auth-server/package.json
+++ b/packages/fxa-auth-server/package.json
@@ -103,7 +103,7 @@
     "urijs": "1.19.1",
     "uuid": "1.4.1",
     "verror": "^1.10.0",
-    "web-push": "3.3.0"
+    "web-push": "3.4.4"
   },
   "devDependencies": {
     "@types/cbor": "5.0.0",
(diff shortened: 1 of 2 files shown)


dannycoates (approved):
👍

PR #6440, opened 2020-09-15
fix(tests): disable codecov beta check annotations

vbudhram:
## Because

- The "Unchanged files with check annotations" is still in beta and the comments clutter reviews

## This pull request

- Disables the check

Closes: NA

## Checklist


6a68 (approved):
Yes! thank you for this

PR #6442, opened 2020-09-15
feat(settings): standardize error/success messages and reporting

jodyheavener:
## Because

- We perform many mutations and actions across Settings that require error/success messages. Some of the error instances should be reported to Sentry.

## This pull request

- Tries to standardize how we show messages and report errors.

Closes: #6249

## Checklist

## Other information

Okay, so the goal of this issue was to set a convention for how we handle mutation errors. Here's exactly what I did. I encourage pushback:

- Moved all mutations to our custom `useMutation` hook.
  - This hook can be supplied an `onError` function that supplies an `ApolloError` object for you to check for `graphQLErrors`, which should then be rendered in the UI or AlertBar. You can also throw inside this function to send an error to report the error object or another object to Sentry.
  - Internally, `onError` will always report non-GQL errors to Sentry. As Settings evolves we can potentially start ignoring errors on a case-by-case basis.
  - All errors now report out to `console.error`, whether you supply a custom function or not. This is just the help debugging. I could be convinced to only enable this in development. Also, make sure you stub this in tests.
- One thing you'll notice is that when errors occur I am showing a pre-written error message if not rendering GQL errors. I'm not sure it's helpful to be showing things like network error messages in the alert bar. These error/success messages are not set in stone. I have a chat with @MeridelW coming up where we'll iron them out and make changes where needed (followup PR): https://jira.mozilla.com/browse/FXA-2540
- Consolidated how we use AlertBar by making sure the `useAlertBar` hook is present everywhere. This includes error and success messages. 
- Updated the documentation on how we should handle errors.

The TLDR:

- Use our custom `useMutation` hook.
- Handle errors with a custom `onError` function, and display GQL errors in the UI or an AlertBar, optionally throwing if you think it should be sent to Sentry.
- Use dynamic `AlertBar` rendering along with the `useAlertBar` hook.
- Display pre-written error messages instead of relying on the error message from Apollo.

### DIFF
--- a/packages/fxa-settings/README.md
+++ b/packages/fxa-settings/README.md
@@ -1,6 +1,6 @@
 # Firefox Accounts Settings
 
-This documentation is up to date as of 2020-09-01.
+This documentation is up to date as of 2020-09-23.
 
 ## Table of Contents
 
@@ -8,6 +8,7 @@ This documentation is up to date as of 2020-09-01.
 [Development](#development)\
 [GQL and REST API Calls](#gql-and-rest-api-calls)\
 — [Global Application Data](#global-application-data)\
+— [GQL error handling](#gql-error-handling)\
 [Components to Know](#components-to-know)\
 — [`LinkExternal`](#linkexternal)\
 — [`AlertBar`](#alertbar-and-alertexternal)\
@@ -26,6 +27,7 @@ This documentation is up to date as of 2020-09-01.
 [Testing and Mocks for Tests/Storybook](#testing-and-mocks-for-testsstorybook)\
 — [`AlertBar` or `AlertExternal`](#components-that-use-alertbar-or-alertexternal)\
 — [`useAccount`, `useSession`, or GQL Mocks](#components-that-use-useaccount-usesession-or-need-a-gql-mock)\
+— [Mocking mutation errors](#mocking-mutation-errors)\
 [Storybook](#storybook)\
 [License](#license)
 
@@ -56,6 +58,45 @@ While most API calls can be performed with GQL, there are a few calls that must 
 This application uses [Apollo client cache](https://www.apollographql.com/docs/react/caching/cache-configuration/) to store app-global state, holding both data received from the GQL server and [local data](https://www.apollographql.com/docs/tutorial/local-state/) that needs to be globally accessible. You can see the shape of this data in the schema found within the `GetInitialState` query, noting that a `@client` directive denotes local data.
 
 Access the client cache data in top-level objects via custom `use` hooks. At the time of writing, `useAccount` and `useSession` (see the "Sessions" section) will allow you to access `data.account` and `data.session` respectively inside components where that data is needed. See the "Testing" section of this doc for how to mock calls.
+
+### GQL error handling
+
+When an error occurs from a query or mutation, the Apollo Client GQL response distinguishes between two kinds of errors on the `error` object - `graphQLErrors`, an array of errors provided by a GQL server-side resolver, and `networkError`, an error that is thrown outside of a resolver. Apollo Server [can throw](https://www.apollographql.com/docs/apollo-server/data/errors/) `AuthenticationError`, `ForbiddenError`, `UserInputError`, or a generic `ApolloError` which Apollo Client stores in `graphQLErrors`. We don’t need to send these to Sentry on the client-side because all errors aside from `UserInputError` are logged by the `fxa-gql-api package`, but a `networkError` should be logged because it means the query was rejected and no data was returned, such as if Apollo Client failed to connect to the GQL endpoint. Learn more about [Apollo error handling](https://www.apollographql.com/docs/react/data/error-handling/).
+
+All errors should be handled. As a general rule of thumb, validation errors from `graphQLErrors` should be displayed the user, while `networkErrors` should report to Sentry.
+
+When performing a GraphQL mutation, use our custom [`useMutation`](./lib/hooks.tsx) hook and set the `onError` property to handle GraphQL Errors (such as validations). The hook will automatically send any network-level errors to Sentry and return `ApolloError` for you to access GQL errors and render them in the UI or an AlertBar.
+
+Example of a mutation with errors appearing in an `AlertBar`:
+
+```tsx
+const [destroySession, { data }] = useMutation(DESTROY_SESSION_MUTATION, {
+  onError: (error) => {
+    // The useAlertBar hook's `error` method can take an ApolloError object
+    // and conditionally render GQL errors if they're present, or fall back
+    // to the error string you provided.
+    alertBar.error('Sorry, there was a problem signing you out.', error);
+  },
+});
+```
+
+Example of a mutation with validation errors appearing in a tooltip, and other errors in an `AlertBar`:
+
+```tsx
+const [verifySecondaryEmail] = useMutation(VERIFY_SECONDARY_EMAIL_MUTATION, {
+  onError: (error) => {
+    // If we receive an GQL error it is related to invalid input, so we'll
+    // set it as a state string that is used in a tooltip component.
+    if (error.graphQLErrors?.length) {
+      setErrorText(error.message);
+    } else {
+      alertBar.error('There was a problem sending the verification code.');
+    }
+  },
+});
+```
+
+See the "Testing" section for mocking mutation errors.
 
 ## Components to Know
 
@@ -438,6 +479,37 @@ const mocks = [];
 <MockedCache account={{ avatarUrl: null }} verified={false} {...{ mocks }}>
   <HeaderLockup />
 </MockedCache>;
+```
+
+### Mocking mutation errors
+
+Testing for GQL and network errors is also pretty straightforward. In your tests, wrap your component in a [MockedCache](#components-that-use-useaccount-usesession-or-need-a-gql-mock) and provide it with mock mutations that produce either an array of `GraphQLError`s, or a standard `Error`. Example with both:
+
+```tsx
+const mocks = [
+  {
+    request: {
+      query: VERIFY_SESSION_MUTATION,
+      variables: { input: { code: '12345678' } },
+    },
+    result: {
+      errors: [new GraphQLError('invalid code')],
+    },
+  },
+  {
+    request: {
+      query: VERIFY_SESSION_MUTATION,
+      variables: { input: { code: '87654321' } },
+    },
+    error: new Error('network error'),
+  },
+];
+
+renderWithRouter(
+  <MockedCache verified={false} {...{ mocks }}>
+    <ModalVerifySession {...{ onDismiss, onError }} />
+  </MockedCache>
+);
 ```
 
 ### Common Errors
--- a/packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
+++ b/packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
@@ -30,6 +30,12 @@ const mockGqlError = () => ({
   error: new Error('Aw shucks'),
 });
 
+window.console.error = jest.fn();
+
+afterAll(() => {
+  (window.console.error as jest.Mock).mockReset();
+});
+
 describe('DropDownAvatarMenu', () => {
   it('renders and toggles as expected with default values', () => {
     render(
@@ -141,7 +147,7 @@ describe('DropDownAvatarMenu', () => {
       });
 
       expect(screen.getByTestId('sign-out-error').textContent).toContain(
-        'Aw shucks'
+        'Sorry, there was a problem signing you out.'
       );
     });
   });
--- a/packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
+++ b/packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
@@ -3,14 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React, { useState } from 'react';
-import { gql, useMutation } from '@apollo/client';
+import { gql } from '@apollo/client';
 import Avatar from '../Avatar';
 import AlertBar from '../AlertBar';
 import { useAccount } from '../../models';
-import sentryMetrics from 'fxa-shared/lib/sentry';
 import { clearSignedInAccountUid } from '../../lib/cache';
-import { useClickOutsideEffect, useBooleanState } from 'fxa-react/lib/hooks';
-import { useEscKeydownEffect } from '../../lib/hooks';
+import { useClickOutsideEffect } from 'fxa-react/lib/hooks';
+import { useEscKeydownEffect, useMutation, useAlertBar } from '../../lib/hooks';
 import { ReactComponent as SignOut } from './sign-out.svg';
 
 export const DESTROY_SESSION_MUTATION = gql`
@@ -29,19 +28,14 @@ export const DropDownAvatarMenu = () => {
     setRevealed
   );
   useEscKeydownEffect(setRevealed);
-  const [alertBarRevealed, revealAlertBar, hideAlertBar] = useBooleanState();
+  const alertBar = useAlertBar();
   const dropDownId = 'drop-down-avatar-menu';
 
-  // TODO: FXA-2450
-  const [destroySession, { data, error }] = useMutation(
-    DESTROY_SESSION_MUTATION,
-    {
-      onError: (error) => {
-        revealAlertBar();
-        sentryMetrics.captureException(error);
-      },
-    }
-  );
+  const [destroySession, { data }] = useMutation(DESTROY_SESSION_MUTATION, {
+    onError: (error) => {
+      alertBar.error('Sorry, there was a problem signing you out.', error);
+    },
+  });
 
   const signOut = () => {
     destroySession({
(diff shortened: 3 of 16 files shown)


LZoog (commented):
Looking good so far. I think my main question is around when/where we're checking for `error.graphQLErrors?.length`, which should be truthy if the error is coming from one of our resolvers - should we check for that every time an error is thrown and report it if it's not a GQLError?

packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx
```diff
@@ -3,14 +3,13 @@
...
+import { gql } from '@apollo/client';
 import Avatar from '../Avatar';
 import AlertBar from '../AlertBar';
 import { useAccount } from '../../models';
-import sentryMetrics from 'fxa-shared/lib/sentry';
 import { clearSignedInAccountUid } from '../../lib/cache';
-import { useClickOutsideEffect, useBooleanState } from 'fxa-react/lib/hooks';
-import { useEscKeydownEffect } from '../../lib/hooks';
+import { useClickOutsideEffect } from 'fxa-react/lib/hooks';
+import { useEscKeydownEffect, useMutation, useAlertBar } from '../../lib/hooks';
```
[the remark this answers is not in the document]
jodyheavener: Fair point, I'll add some documentation :)

packages/fxa-settings/src/components/PageSecondaryEmailAdd/index.tsx
```diff
@@ -22,15 +23,15 @@ export const PageSecondaryEmailAdd = (_: RouteComponentProps) => {
...
   const goBack = useCallback(() => window.history.back(), []);
 
   const [createSecondaryEmail] = useMutation(CREATE_SECONDARY_EMAIL_MUTATION, {
     onError: (error) => {
       if (error.graphQLErrors?.length) {
         setErrorText(error.message);
       } else {
-        sentryMetrics.captureException(error);
-        // TODO
+        alertBar.error('There was a problem creating this email.');
```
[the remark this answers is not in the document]
jodyheavener: Good point. I'm going to switch it up and change all error instances that we would not typically expect a GQL error from to also throw to Sentry.

packages/fxa-settings/src/components/ModalVerifySession/index.tsx
```diff
@@ -39,12 +40,15 @@ export const ModalVerifySession = ({
...
   const [sendCode] = useMutation(SEND_SESSION_VERIFICATION_CODE_MUTATION, {
     variables: { input: {} },
     ignoreResults: true,
     onError,
   });
+
   const [verifySession] = useMutation(VERIFY_SESSION_MUTATION, {
+    ignoreResults: true,
     onError: (error) => {
       if (error.graphQLErrors?.length) {
```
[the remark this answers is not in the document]
jodyheavener: Good point. Will update to prevent this.

jodyheavener:
Gonna take this back to draft while I make a few more changes. 🙂

meandavejustice:
Looks like there was a conflict with my changes in `UnitRowTwoStepAuth` so this will need a rebase

meandavejustice:
> By default, if an onError function is not supplied when using this hook, we'll capture any exceptions and log them to Sentry.

This is probably fine, but we should add a note to docs somewhere that if you are dealing with any PII to try and avoid letting this fall into Sentry

LZoog:
> This is probably fine, but we should add a note to docs somewhere that if you are dealing with any PII to try and avoid letting this fall into Sentry

Apollo server should be logging GQL errors so I believe we're just going to log [NetworkError error types from Apollo client](https://www.apollographql.com/docs/react/data/error-handling/#network-errors) so I don't _think_ we'll have to worry about this, but good call out. Jody, it might be worth seeing if we can trigger a network error with like, a bad query that tries to send up an email (like switching a primary email) to see what message is returned. 🤔

jodyheavener:
Let's try this again :)

packages/fxa-settings/README.md
```diff
@@ -51,6 +51,23 @@ FxA Settings communicates primarily with the FxA GraphQL API through use of [Apo
 
 While most API calls can be performed with GQL, there are a few calls that must directly communicate with the auth server (for now) for security reasons. To make REST API calls to the auth server, use the `useAuth` and `useAwait` hooks.
 
+### Error handling
+
+All errors that can occur as a result of a network action should be handled. As a general rule of thumb validation errors should be displayed the user while all others should be reported to Sentry unless they are know to occur and intentionally ignored.
```
LZoog, line 56: Add to table of contents? 😀 Also we may want to title this "GQL error handling" since REST errors (whenever we get there) won't be applicable.

For this paragraph, I think we could expound upon what's meant for "network action," something like this maybe? (double check me for accuracy 😉)

"When an error occurs from a query or mutation, the Apollo Client GQL response distinguishes between two kinds of errors on the `error` object - `graphQLErrors`, an array of errors provided by a GQL server-side resolver, and `networkError`, an error that is thrown outside of a resolver. We don't need to send `graphQLErrors` to Sentry because they're logged by the `fxa-gql-api` package, but a `networkError` should be logged because it means the query was rejected and no data was returned, such as if Apollo Client failed to connect to the GQL endpoint.

All errors should be handled. As a general rule of thumb, validation errors (which are in the `graphQLErrors` array) should be displayed the user, while all others should be reported to Sentry unless they are know to occur and should be intentionally ignored."

packages/fxa-settings/README.md
```diff
@@ -51,6 +51,23 @@ FxA Settings communicates primarily with the FxA GraphQL API through use of [Apo
...
+
+When performing a GraphQL mutation, use our custom [`useMutation`](./lib/hooks.tsx) hook and set the `onError` property to handle GraphQL Errors (such as validations). The hook will automatically send any network-level errors to Sentry and return an `ApolloError` for you to access GQL errors and render them in the UI or an AlertBar. Example:
+
+```tsx
+const [destroySession, { data }] = useMutation(DESTROY_SESSION_MUTATION, {
+  onError: (error) => {
+    // The useAlertBar hook's `error` method can take an ApolloError object
+    // and conditionally render GQL errors if they're present, or fall back
+    // to the error string you provided.
```
LZoog, line 65: Oh sweet! I was going to ask if we wanted to display the error to the user but the alert bar is smart.

packages/fxa-settings/src/components/PageSecondaryEmailVerify/index.tsx
```diff
@@ -21,15 +21,15 @@ export const PageSecondaryEmailVerify = ({ location }: RouteComponentProps) => {
   const [errorText, setErrorText] = useState<string>();
   const goBack = useCallback(() => window.history.back(), []);
   const navigate = useNavigate();
+  const alertBar = useAlertBar();
   const email = (location?.state as any)?.email as string | undefined;
 
   const [verifySecondaryEmail] = useMutation(VERIFY_SECONDARY_EMAIL_MUTATION, {
     onError: (error) => {
       if (error.graphQLErrors?.length) {
         setErrorText(error.message);
```
LZoog, line 30: Hm, it may also be good in the docs to call out that when we send up user input with a mutation, that's when we need this kind of check, IE verifying a secondary email needs `if (error.graphQLErrors?.length) {` because it could return "invalid code," but signing a user out doesn't take any user input so we don't need the check for it there.

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
```diff
@@ -26,25 +26,22 @@ export const UnitRowRecoveryKey = () => {
...
-    hideModal();
-    alertBar.show();
-  };
 
-  const [getAccount, { accountLoading }] = useLazyAccount((error) => {
-    setErrorText('Sorry, there was a problem refreshing the recovery key.');
-    alertBar.show();
+  const [getAccount, { accountLoading }] = useLazyAccount(() => {
+    hideModal();
+    alertBar.error('Sorry, there was a problem refreshing the recovery key.');
```
LZoog, line 32: Do we need to send in `error` to the alert bar here?

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
```diff
@@ -98,7 +95,16 @@ export const UnitRowRecoveryKey = () => {
...
       </LinkExternal>
       {modalRevealed && (
-        <VerifiedSessionGuard onDismiss={hideModal} onError={onError}>
+        <VerifiedSessionGuard
+          onDismiss={hideModal}
+          onError={(error) => {
+            hideModal();
+            alertBar.error(
+              'Sorry, there was a problem verifying your session',
+              error as ApolloError
```
LZoog, line 104: Why'd we need `error as ApolloError` here and not elsewhere?

packages/fxa-settings/README.md
LZoog: Oh also, Ben had mentioned validation errors aren't reported to Sentry but other errors on `graphQLErrors` are so maybe we add something about that. I'm not sure if we normally log "invalid code" and such in Sentry, it seems like that wouldn't be useful...

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
6a68: I would guess this is casting an `AuthenticationError` to an `ApolloError`? maybe we should just have gql-api always send down an `ApolloError`, like Ben suggested [here](https://github.com/mozilla/fxa/pull/6504#discussion_r492215731)

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
```diff
@@ -12,7 +13,7 @@ export const VerifiedSessionGuard = ({
   children,
 }: {
   onDismiss: () => void;
-  onError: (error: Error) => void;
+  onError: (error: Error | ApolloError) => void;
```
6a68, line 16: why not `onError: (error: Error | ApolloError | AuthenticationError)` ?

packages/fxa-settings/src/lib/hooks.tsx
```diff
@@ -109,8 +113,8 @@ export function useAlertBar({
   );
 
   const error = useCallback(
-    (message: ReactNode) => {
-      setContent(message);
+    (message: ReactNode, error?: ApolloError) => {
```
6a68, line 116: I think you could also allow the `AuthenticationError` here, because it does have a `message` on it. Then you wouldn't have to remember to cast the type in those few spots above.

6a68 (commented):
Nice work! I don't feel like I have enough context yet to 👍 or 👎, so I just left some comments 🍻

packages/fxa-settings/README.md
jodyheavener: Thanks for the updated copy. I think that works, but I'll keep the section about using the useMutation hook if thats cool. I also don't think we need to log non-GQL validation errors.

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
jodyheavener: Yes

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
jodyheavener: > Why'd we need error as ApolloError here and not elsewhere?

Huh, I did this because `ModalVerifySession`, by way of `VerifiedSessionGuard`, indicated it could return a regular `Error` or an `ApolloError`. But on closer inspection I don't see where `Error` is possible, so I've modified the function definition types.

packages/fxa-settings/src/components/DropDownAvatarMenu/index.test.tsx
```diff
@@ -30,6 +30,8 @@ const mockGqlError = () => ({
   error: new Error('Aw shucks'),
 });
 
+window.console.error = jest.fn();
```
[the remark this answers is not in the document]
jodyheavener: Good question. I assumed they were auto-restored and haven't had issues, but I don't see it documented anywhere, so I'll restore to be safe.

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
jodyheavener: Where is `AuthenticationError` thrown?

packages/fxa-settings/README.md
LZoog: Oh yeah absolutely, I just meant for the first paragraph. I think "When performing a GraphQL mutation" etc. is great.

packages/fxa-settings/src/components/PageSecondaryEmailVerify/index.tsx
jodyheavener: Huh, Jared you raise a good point. From quickly looking through our test files I don't think we actually explicitly test for graphQLErrors. We just test for _an_ error. I'll document this process.

packages/fxa-settings/src/components/VerifiedSessionGuard/index.tsx
LZoog: Ohh, `AuthenticationError` is an error typed in Apollo Server, not Apollo Client. This is a nice article: https://www.apollographql.com/blog/full-stack-error-handling-with-graphql-apollo-5c12da407210/

jodyheavener: Ewww, are all GQL errors prepended with `GraphQL error: `? If so, since this is presented to the user, I am tempted to chop that beginning off.

LZoog: I am not sure. FWIW, this PR sorta relates to what Jared is doing in #6504

packages/fxa-settings/README.md
```diff
@@ -57,6 +58,54 @@ This application uses [Apollo client cache](https://www.apollographql.com/docs/r
...
 
+### GQL error handling
+
+When an error occurs from a query or mutation, the Apollo Client GQL response distinguishes between two kinds of errors on the `error` object - `graphQLErrors`, an array of errors provided by a GQL server-side resolver, and `networkError`, an error that is thrown outside of a resolver. We don't need to send `graphQLErrors` to Sentry because they're logged by the `fxa-gql-api` package, but a `networkError` should be logged because it means the query was rejected and no data was returned, such as if Apollo Client failed to connect to the GQL endpoint.
+
+All errors should be handled. As a general rule of thumb, validation errors (which are in the `graphQLErrors` array) should be displayed the user, while all others should be reported to Sentry unless they are know to occur and should be intentionally ignored. Because GQL errors are essentially validation errors, broadly speaking only mutations that take an input need to be checked for GQL errors.
```
LZoog, line 65: I apologize in advance for being so picky here (and for not suggesting this originally, I did some reading this morning), but I think we can be even more explicit:

"When an error occurs from a query or mutation, the Apollo Client GQL response distinguishes between two kinds of errors on the `error` object - `graphQLErrors`, an array of errors provided by a GQL server-side resolver, and `networkError`, an error that is thrown outside of a resolver. Within `graphQLErrors`, the Apollo Server can return `AuthenticationError`, `ForbiddenError`, and `UserInputError`s. We don’t need to send these to Sentry on the client-side because all errors aside from `UserInputError` are logged by the `fxa-gql-api` package, but a `networkError` should be logged because it means the query was rejected and no data was returned, such as if Apollo Client failed to connect to the GQL endpoint.

All errors should be handled. As a general rule of thumb, validation errors from `graphQLErrors` should be displayed the user, while `networkError`s should report to Sentry.

[your “when performing a GQL mutation section" with example, but change “Example” to “Example of `useMutation` without user input”]

Example of `useMutation` with user input, checking for `graphQLErrors`:

[example checking for `if (error.graphQLErrors?.length) {`]"


I think we can move the “Testing” part into the “Testing” section too, in other sections we say "See the "Testing" section of this doc for how to X."

jodyheavener:
Got another update for ya @LZoog

packages/fxa-settings/README.md
```diff
@@ -57,6 +59,45 @@ This application uses [Apollo client cache](https://www.apollographql.com/docs/r
 
 Access the client cache data in top-level objects via custom `use` hooks. At the time of writing, `useAccount` and `useSession` (see the "Sessions" section) will allow you to access `data.account` and `data.session` respectively inside components where that data is needed. See the "Testing" section of this doc for how to mock calls.
 
+### GQL error handling
+
```
LZoog, line 63: > Within `graphQLErrors`, the Apollo Server can return `AuthenticationError`, `ForbiddenError`, and `UserInputErrors`.

`UserInputError` should be singular and it looks like the docs do mention a generic `ApolloError`, maybe this:

"Apollo Server can throw `AuthenticationError`, `ForbiddenError`, `UserInputError`, or a generic `ApolloError` which Apollo Client contains in `graphQLErrors`."

Could also be helpful to link to https://www.apollographql.com/docs/apollo-server/data/errors/ and https://www.apollographql.com/docs/react/data/error-handling/

LZoog (approved):
I only had one other comment, thanks for all of your work here!!

packages/fxa-settings/README.md
jodyheavener: Heh, I completely forgot to add what you had Slacked me. Apologies.

Issue #6100, opened 2020-07-31
[auth-server][tests] 33 intermittent oauth "Error: connect ETIMEDOUT" test failures


data-sync-user:
I think they are caused by something DB migration related.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2370)
┆Issue Number: FXA-2370

data-sync-user:
➤ Vijay Budhram commented:

From mtg: This has been in auth-server for a few weeks, Danny suggested rolling back to earlier mysql package.json

data-sync-user:
➤ Barry Chen commented:

Some findingz so far

  * the test suites pass when ran individually, e.g. `./scripts/test-local.sh test/oauth/whatever.js`
  * the order matters when running multiple suites

data-sync-user:
➤ Barry Chen commented:

I _think_ the culprits are test/oauth/api.js and test/oauth/db/index.js, as they both require lib/oauth/db/index.js

data-sync-user:
➤ Barry Chen commented:

The timeout happens in test/oauth/api.js's before() since it has a db.ping() that does not resolve.  There are multiple timeout errors from test/oauth/db/index.js because it does not have a before() with a db.ping().

data-sync-user:
➤ Barry Chen commented:

Wholly unscientific... when I run the test suites in their "natural" order, i.e. `./scripts/test-local.sh test/oauth`, I'd get the timeout errors 100% of the time locally.  But if I run the test suites with the timeout errors first, i.e. `./scripts/test-local.sh test/oauth/api.js test/oauth/db/index.js test/oauth/grant.js test/oauth/token.js ...`, then they are more intermittent, with failures more frequent the successes.  And if I simply exclude `test/oauth/util.js` from the end, I'd get successes frequently. 🤷‍♂️

data-sync-user:
➤ Barry Chen commented:

FWIW nothing useful was observed from the MySQL side; just fewer connections (1 or 2 vs 11 iirc) while the timeout errors occur.

data-sync-user:
➤ Barry Chen commented:

Danny is gonna take over as I'm fresh out of ideas at this point.

PR #6443, opened 2020-09-15
fix(oauthdb): split creating oauthdb from main process startup

dannycoates:
possibly a fix for #6100


dannycoates:
@chenba can you try this? I can't repro the bug locally very reliably

chenba:
Ran it a few times, 100% success rate. 💯

packages/fxa-auth-server/scripts/test-local.sh
```diff
@@ -16,6 +16,7 @@ node -r ts-node/register ./scripts/gen_keys.js
 node -r ts-node/register ./scripts/gen_vapid_keys.js
 node -r ts-node/register ./scripts/oauth_gen_keys.js
 node -r ts-node/register ../fxa-auth-db-mysql/bin/db_patcher > /dev/null
+node -r ts-node/register ./scripts/update-oauth-db.js
```
vbudhram, line 19: nit: A name like `oauth-db-patcher` reads a little cleaner to me what this script would do.

vbudhram (commented):
Also ran this a few times with 100% pass rate 🙌🏽

PR #6456, opened 2020-09-16
refactor(fxa-shared): cleanup sentry and add exc route

### LINKED ISSUE #6454: Cleanup sentry error logic and sentry error test
Sentry error reporting was a quick copy over from event-broker, it should be cleaned up into separate reporting helpers and the basic intercepter/service. To verify sentry is working we should also add a health controller route that triggers the error.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2537)
┆Issue Number: FXA-2537

bbangert:
Because:

* Sentry relocation didn't clean up reporting functionality from the
  service using it.
* We had no way to verify exceptions are thrown and captured by Sentry.

This commit:

* Refactors the sentry reporting functionality so it can be used
  directly by other code.
* Adds an __exception__ route with basic trigger protection to avoid
  excessive error reporting.

Closes #6454

## Checklist

### DIFF
--- a/packages/fxa-shared/nestjs/health/health.controller.ts
+++ b/packages/fxa-shared/nestjs/health/health.controller.ts
@@ -7,6 +7,8 @@ import { Version } from '../version';
 import { HEALTH_CONFIG } from './health.constants';
 import { HealthControllerConfigParams } from './health.module';
 
+let lastTrigger = Date.now() - 60_000;
+
 @Controller()
 export class HealthController {
   constructor(
@@ -29,5 +31,16 @@ export class HealthController {
   @Get('__version__')
   versionData(): Version {
     return this.config.version;
+  }
+
+  @Get('__exception__')
+  exc() {
+    // Limit how frequently this could be abused.
+    const now = Date.now();
+    if (now - lastTrigger >= 60_000) {
+      lastTrigger = now;
+      throw new Error('Test Exception');
+    }
+    return 'Too Frequent';
   }
 }
--- a/packages/fxa-shared/nestjs/sentry/reporting.ts
+++ b/packages/fxa-shared/nestjs/sentry/reporting.ts
@@ -1,0 +1,164 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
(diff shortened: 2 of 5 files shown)


packages/fxa-shared/nestjs/health/health.controller.ts
```diff
@@ -30,4 +32,15 @@ export class HealthController {
   versionData(): Version {
     return this.config.version;
   }
+
+  @Get('__exception__')
```
vbudhram, line 36: First time I've seen `__exception__`, didn't find anything on sentry or nextjs about this endpoint from some quick googling. Do you have a link for more info?

bbangert: That's the URL, so that we can hit `/__exception__` to intentionally cause an exception and verify it gets to Sentry.

vbudhram: @bbangert Makes sense, much easier than just adding throw in a random endpoint and hitting it 👍🏽

vbudhram (approved):

packages/fxa-shared/nestjs/sentry/reporting.ts
```diff
@@ -0,0 +1,164 @@
...
+import { GqlContextType, GqlExecutionContext } from '@nestjs/graphql';
+import * as Sentry from '@sentry/node';
+import { SQS } from 'aws-sdk';
+import { Request } from 'express';
+
+// Matches uid, session, oauth and other common tokens which we would
+// prefer not to include in Sentry reports.
+const TOKENREGEX = /[a-fA-F0-9]{32,}/gi;
+// RFC 5322 generalized email regex, ~ 99.99% accurate.
+const EMAILREGEX = /(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))/gi;
```
jodyheavener, line 14: I always get a little nervous about big regexs like this. If it's just for filtering logs maybe something like `/\S+@\S+\.\S+/` would be suitable.

jodyheavener (commented):
oh looks like I missed the boat on this 🤷

Issue #6458, opened 2020-09-16
Determine why event-broker is throwing 500 errors


data-sync-user:
During the conversion of event-broker to NestJS, the event-broker started throwing 500's. This could likely be caused by authentication errors not being reported as Sentry has nothing. The event-broker should be stood up locally and verified to be handling google auth correctly. If that's working fine, then additional logging should be added throughout the pubsub proxy controller to determine the issue of the 500's.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2538)
┆Issue Number: FXA-2538

PR #6459, opened 2020-09-16
fix(event-broker): possible guard fix + logging

bbangert:
Because:

* It's possible I named the strategy in the incorrect location when
  moving this to NestJS.
* We would like more debug logging to detect details about why a proxy
  request may have failed from an RP.

This commit:

* Moves the strategy name so the guard could pick it up.
* Adds more logging to the pubsub controller for additional debugging.

Issue #6458

## Checklist

## Additional Notes

I spent a bit of time with Lets Encrypt and a Google console to setup the event-broker locally and verify that the auth is working. It is! That means the error is possibly somewhere else in the delivery chain, which we need to verify in our dev environment. So additional logging is included here.


bbangert:
Ok, additional note, I did a verification locally. A 500 is thrown without the movement of the `googlejwt` definition. I had defined it incorrectly before, thus the error. So this will indeed fix the error.

lmorchard (approved):

PR #6462, opened 2020-09-16
refactor(profile): replaced graphicsmagick with sharp

dannycoates:
because I'm tired of the extra hoops to maintain gm


vbudhram (approved):
@dannycoates Never really thought about removing this, but sharp looks legit, thank you! 💯

PR #6467, opened 2020-09-17
task(fxa-settings): Tooltip component

### LINKED ISSUE #6253: Settings tooltip component
* clean up styles to match figma mockups
* pull this out into a component
* handle long text (probably ellipses with a title)
* get rid of .tooltip class and file, replace with direct tailwind classes in component

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2451)
┆Issue Number: FXA-2451

meandavejustice:
- fixes #6253

## Because

- cleaning up and consolidating the tooltip component

## Checklist

## Screenshots (Optional)


packages/fxa-settings/src/styles/tailwind.scss
```diff
@@ -38,3 +37,27 @@ body {
     @apply transition duration-150;
   }
 }
+
```
meandavejustice, line 40: I tried avoiding any non tailwind classes, but the only way I could find to use `:before` or other pseudo elements was to [use it in a plugin](https://github.com/yutahaga/tailwindcss-pseudo-elements), I tried it out but it seemed to be a much worse solution as you have to add each rule as a `utility`. https://github.com/yutahaga/tailwindcss-pseudo-elements#configuration

jodyheavener: I also noticed that some tooltips anchor from the center of the element:

Screen Shot 2020-09-18 at 11 59 15 AM

While others anchor from the left:

Screen Shot 2020-09-18 at 11 59 05 AM

I could see this being a prop on the component that uses different anchoring classes.

jodyheavener (changes requested):
Thanks for putting this together! I've left a few comments. In addition, I think we should try to get a little closer to the Figma designs; specifically in the typeface, box shadow, and background color (just for the neutral one, the error is correct I think?). Ping me if you want to chat about it some more. Happy to lend a hand if desired!

meandavejustice:
Just pushed an update that cleans up the style inconsistencies, and adds a prop to position the tooltip on top or bottom. Still working on the anchoring left option.

meandavejustice:
grey-500-discussion

We decided on `grey-500` for the default tooltip background color

packages/fxa-settings/src/components/Tooltip/index.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <p>Error tooltip pointing to this text</p>
+          <Tooltip type="error" message="My tooltip message here" />
+        </div>
+      </div>
+    </div>
+  ))
+  .add('bottom', () => (
+    <div className="p-20 max-w-md">
+      <div className="mb-3">
+        <div className="relative table">
```
jodyheavener, line 33: I don't think `table` is doing anything in these Storybooks, is it?

packages/fxa-react/configs/tailwind.js
```diff
@@ -59,6 +59,10 @@ module.exports = {
         '-3': '-0.75rem',
         '-52': '-13rem',
         '55': '13.75rem',
+        ten: '10%',
+        full: '100%',
+        center: '44%',
```
jodyheavener, line 64: I have hesitation about such a specific number. It looks like this is used to horizontally center the caret/triangle on the tooltip, and while it looks visibly center it's actually slightly off in practice:

Group

Instead of this additional `left-center` class we could used `left-1/2 transform -translate-x-1/2` which should take care of perfectly centering the caret.

packages/fxa-react/configs/tailwind.js
```diff
@@ -59,6 +59,10 @@ module.exports = {
...
+        beyond: '140%',
```
jodyheavener, line 65: This also seems to cause issues with the tooltip positioning because it's a specific relative number. Check out where this bottom-anchored tooltip ends up when the content grows:

Screen Shot 2020-09-22 at 11 37 20 AM

packages/fxa-settings/src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    bottom: 'top',
+  }[position];
+
+  return (
+    <div
+      data-testid="tooltip"
+      title={message}
+      className={classNames(
+        `z-50 absolute py-2 px-6 text-center text-white
+         p-3 rounded text-xs left-0 font-Metropolis font-bold
```
jodyheavener, line 40: The class `font-Metropolis` does not exist, I think you are looking for `font-header` here.

packages/fxa-settings/src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const bgColor = type === 'error' ? 'red-600' : 'grey-500';
+  // caret position and the positioning rule for tooltip are actually the opposite
+  // of the passed in position;
+  const oppositePositionRule = {
+    top: 'bottom',
+    bottom: 'top',
+  }[position];
+
+  return (
+    <div
```
jodyheavener, line 35: I think we're going to have trouble with this when it comes to narrow anchors. For example, we have this button to take a photo that will need to display a tooltip:

Screen Shot 2020-09-22 at 11 50 52 AM

But when I try to apply the tooltip to a narrow element it becomes off-center:

Screen Shot 2020-09-22 at 11 51 58 AM

Granted, some of this may be up to the specific use-case to supply custom classes to the component to, say, force the width of tooltip, but generally I think it should maintain its center positioning.

packages/fxa-settings/src/components/Tooltip/index.tsx
jodyheavener: It looks like in the Figma doc both tooltips have a box shadow. Can we apply that here?

packages/fxa-settings/src/styles/tailwind.scss
```diff
@@ -5,7 +5,6 @@
 @import './animations.css';
 @import './ctas.scss';
 @import './unit-row.scss';
-@import './tooltip.scss';
```
jodyheavener, line 8: 🎉

packages/fxa-settings/src/components/Tooltip/index.tsx
meandavejustice: I don't really know what to do to address this, would the `left-1/2 transform -translate-x-1/2` rule solve this?

packages/fxa-react/configs/tailwind.js
meandavejustice: I guess we can just use 100% and let the developer add margin classes passed into the Tooltip component instead

jodyheavener (commented):
Apologies, forgot to hit send on this yesterday. Just a couple small thoughts and then I think we're ready to go.

jodyheavener (approved):

PR #6496, opened 2020-09-19
refactor(event-broker): cleanup proxy logic

bbangert:
Because:

* It was not clear what the proxy logic did and difficult to fully
  test.

This commit:

* Splits up the proxy logic into logical components and add's a test
  for additional coverage.

## Checklist

### DIFF
--- a/packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.spec.ts
+++ b/packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.spec.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-import { Provider } from '@nestjs/common';
+import { Provider, HttpException } from '@nestjs/common';
 import { ConfigModule } from '@nestjs/config';
 import { Test, TestingModule } from '@nestjs/testing';
 import { MozLoggerService } from 'fxa-shared/nestjs/logger/logger.service';
@@ -66,14 +66,6 @@ const createValidPasswordMessage = (): string => {
   ).toString('base64');
 };
 
-const mockResponse = () => {
-  const res: any = {};
-  res.status = jest.fn().mockReturnValue(res);
-  res.set = jest.fn().mockReturnValue(res);
-  res.send = jest.fn().mockReturnValue(res);
-  return res;
-};
-
 describe('PubsubProxy Controller', () => {
   let controller: PubsubProxyController;
   let jwtset: any;
@@ -82,11 +74,9 @@ describe('PubsubProxy Controller', () => {
   const mockWebhook = () => {
     nock('http://accounts.firefox.com')
       .post('/webhook')
-      .reply(200, '', {
-        'X-AUTH': (req, res, body) => {
-          const auth = req.getHeader('authorization');
-          return typeof auth === 'string' ? auth : 'unknown';
-        },
+      .reply(function (uri, requestBody) {
+        const auth = this.req.getHeader('authorization');
+        return [200, { token: typeof auth === 'string' ? auth : 'unknown' }];
       });
   };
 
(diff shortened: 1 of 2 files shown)


packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.ts
```diff
@@ -32,54 +34,93 @@ export class PubsubProxyController {
   @UseGuards(GoogleJwtAuthGuard)
   @Post(':clientId')
   async proxy(
-    @Res() res: Response,
```
bbangert, line 35: When the framework specific Response is passed in, Nest disengages its error handling. To split this I needed the flow control of throwing HTTPExceptions. See more info at https://docs.nestjs.com/controllers#request-object

packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.ts
```diff
@@ -32,54 +34,93 @@ export class PubsubProxyController {
...
+    this.log.debug('proxyDeliverSuccess', {
+      statusCode: status,
+    });
+
+    // Timing for how long the message was in the queue, reported here
+    // only if the message was sent successfully to avoid repeat metrics
+    // for failures.
+    this.metrics.timing(`proxy.queueDelay`, queueDelay);
+
+    throw new HttpException(data, status);
```
bbangert, line 60: Without using a Response object, a status code can only be set with a decorator, or on a `HttpException`.

6a68: Maybe I'm just not that familiar with NestJS yet, but this seems like R- material to me.

If you need to use a decorator to set a status code, why not do that?

Using exception handling for ordinary flow control is an antipattern and throws off JIT execution speed

packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.spec.ts
```diff
@@ -171,53 +177,61 @@ describe('PubsubProxy Controller', () => {
...
+              message: { data: message, messageId: 'test-message' },
+              subscription: 'test-sub',
+            },
+            'abc1234'
+          );
+        } catch (err) {
+          const status = err.getStatus();
+          const body = err.getResponse();
+          expect(status).toBe(200);
+          expect(body).toStrictEqual({ token: 'Bearer ' + TEST_TOKEN });
```
6a68, line 193: Yeah, this seems very strange to me

6a68 (changes requested):
I don't have a lot of context on event-broker, but I can't understand why we would want to use exception handling for regular flow control. Could you go into the forces involved here a little bit more? Why can't we use the status code decorators you mentioned in one of your comments?

packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.spec.ts
bbangert: It's a unit test, so the exception handling that transforms `HTTPException` classes into responses isn't firing. Thus these statements to verify that the code-path executed properly and resulted in the expected `HTTPException`.

packages/fxa-event-broker/src/pubsub-proxy/pubsub-proxy.controller.ts
bbangert: Using a StatusCode decorator means the status code is defined in advance, you cannot dynamically change the statuscode then without using an `HttpException` as done here. In this case, the goal of this route is to proxy the status/response from the Relying Party to Google PubSub so that if its a 503/etc it is automatically retried by Google.

bbangert:
All of our code uses exceptions for flow control. The alternative is to have every single function call return at least one more parameter indicating if it worked or not. This is how its done in Go (poorly), and in Rust (using Result objects and special macro's for easier bailing out without writing more code).

To refactor a large block into smaller blocks, and not have to pass additional data back indicating if the request should stop (and all the additional code testing of the returned value), one must use exceptions. I do think its a bummer that there isn't a better way to tell Nest that I'd like to return both the status code and the body dynamically short of throwing an `HttpException`. I will put in a request with the project for that feature.

6a68 (approved):
OK, we talked about this a bit on Slack. The basic issue here is that Nest is an abstraction layer on top of a request-handling library (like express or fastify). So, if you want to set status codes on a response, you have to drop down into the library underneath NestJS, and then the validators, etc. don't run. This is kind of terrible! While it is possible to set a private field on the Request and use an Interceptor to read that and transform the response on the way out, unfortunately in NestJS [the exception handlers run after the interceptors](https://github.com/nestjs/nest/issues/1342#issuecomment-444666214), so this relatively cleaner approach wouldn't work for proxying errors without discarding the status code.

IMO this is a pretty unfortunate design choice for NestJS, but it does mean that the code is probably doing the right thing as written :-\

PR #6504, opened 2020-09-21
fix(settings): improve GQL error handling to avoid redirect loop

### LINKED ISSUE #6400: Don’t redirect endlessly when the GQL server can’t be reached
We direct to `get_flow` and back to settings in an endless loop when the "general application error" is shown. The page should instead stay in place at the general application error.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2524)
┆Issue Number: FXA-2524

6a68:
I'm annoyed by the branch name, but whatever. Final changes, from the commit message:

fix(settings): improve GQL error handling to avoid redirect loop

* Update the GQL API server to only return an AuthenticationError if it
    received an auth-related 400 or 401 error from auth-server. Otherwise,
    just return a generic ApolloError.

* Add tests for the front-end GQL library to be sure this works
    correctly.

* Add tests for the front-end GQL lib.

Fixes #6400.

-----

Original description from WIP draft:

The bug is that the gql-api was always returning an auth error, even if the problem was that it couldn't hit auth server. Setting up the (missing) tests for this lib has been a little complex; just looking for some feedback on test harness setup. Mocking out globals vs. wrapping in accessor functions might be a tradeoff that's been worked out elsewhere in new-settings?

### DIFF
--- a/packages/fxa-graphql-api/src/lib/auth.ts
+++ b/packages/fxa-graphql-api/src/lib/auth.ts
@@ -4,7 +4,7 @@
 
 import AuthClient from 'fxa-auth-client';
 import { Service, Inject } from 'typedi';
-import { AuthenticationError } from 'apollo-server';
+import { AuthenticationError, ApolloError } from 'apollo-server';
 
 import { fxAccountClientToken } from './constants';
 
@@ -17,7 +17,16 @@ export class SessionTokenAuth {
     try {
       return await this.authClient.sessionStatus(sessionToken);
     } catch (err) {
-      throw new AuthenticationError('Invalid session token');
+      if (err.code === 400 || err.code === 401) {
+        // AuthenticationError doesn't allow us to pass along the error code,
+        // but the error message is unique for most auth-server errors.
+        throw new AuthenticationError(err.message);
+      } else {
+        throw new ApolloError(err.message, '' + err.errno, {
+          errno: err.errno,
+          code: err.code,
+        });
+      }
     }
   }
 }
--- a/packages/fxa-settings/src/lib/gql.test.ts
+++ b/packages/fxa-settings/src/lib/gql.test.ts
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { errorHandler } from './gql';
+import { ErrorResponse } from '@apollo/client/link/error';
+import { Operation, NextLink, ServerError } from '@apollo/client/core';
+import { GraphQLError } from 'graphql';
+
+let errorResponse: ErrorResponse;
+
+describe('errorHandler', () => {
+  let realLocation: Location;
+
+  beforeAll(() => {
+    // Note: it's surprisingly difficult to mock out window.location cleanly.
+    // Messing with globals is gross, but works, and is really simple.
+    realLocation = window.location;
+    delete window.location;
+    window.location = {
+      replace: jest.fn(),
+      search: '',
+      pathname: 'foo',
+    };
+
+    // We don't verify that console.error gets called, but mocking it out
+    // avoids console noise filling up the command line as the tests run.
+    console.error = jest.fn();
+  });
+
+  afterAll(() => {
+    window.location = realLocation;
+    console.error.mockRestore();
+  });
+
+  beforeEach(() => {
+    window.location.replace.mockClear();
+  });
+
+  it('redirects to /get_flow if called with a GQL authentication error', () => {
+    errorResponse = {
+      graphQLErrors: [
+        new GraphQLError(
+          'Incorrect password',
+          undefined,
+          undefined,
+          undefined,
+          undefined,
+          undefined,
+          { code: 'UNAUTHENTICATED' }
+        ),
+      ],
+      operation: (null as any) as Operation,
+      forward: (null as any) as NextLink,
+    };
+
+    errorHandler(errorResponse);
+
+    expect(window.location.replace).toHaveBeenCalledWith(
+      '/get_flow?redirect_to=foo'
+    );
+  });
+
+  it('redirects to get_flow if called with a 401 NetworkError', () => {
+    let networkError: any;
+    networkError = new Error('Network error') as ServerError;
+    networkError.statusCode = 401;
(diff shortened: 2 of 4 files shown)


packages/fxa-graphql-api/src/lib/auth.ts
```diff
@@ -17,7 +17,16 @@ export class SessionTokenAuth {
     try {
       return await this.authClient.sessionStatus(sessionToken);
     } catch (err) {
-      throw new AuthenticationError('Invalid session token');
+      if (err.code === 400 || err.code === 401) {
```
6a68, line 20: auth client error codes are defined here: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/api.md#defined-errors

bbangert: 400 is a bad request, maybe we should always throw ApolloError if we can customize it better?

6a68: The slight abuse of the 400 error code is pretty clear from the auth server docs that I linked above. Seems weird that "unverified account" throws a 400, not a 401, but here we are.

We could instead always throw ApolloError, but then we'll lose the "redirect on auth error" logic that we have on the front-end, or have to hyper-specialize it to various auth server `errno` values :-\

6a68:
I'm annoyed by the branch name, but whatever. Final changes, from the commit message:

-----

fix(settings): improve GQL error handling to avoid redirect loop

* Update the GQL API server to only return an AuthenticationError if it
received an auth-related 400 or 401 error from auth-server. Otherwise,
just return a generic ApolloError.

* The settings code handles auth errors by redirecting to /get_flow,
which then redirects to /signin if there's no session. But if the user
has already been redirected to /get_flow once, then there will be flow
params on the URL; if these params are present, don't redirect, just
show an error screen.

* Add tests for the front-end GQL lib.

Fixes #6400.

packages/fxa-settings/src/lib/gql.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    beforeEach(() => {
+      errorResponse = {
+        graphQLErrors: [
+          new GraphQLError(
+            'Incorrect password',
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
```
6a68, line 49: I'm not a fan of this explicit passing of `undefined` thing. suggestions welcome on a nicer approach :-)

packages/fxa-settings/src/lib/gql.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+  });
+
+  describe('if called with a 401 NetworkError', () => {
+    let networkError: any;
+
+    beforeEach(() => {
+      // Clumsily imitating the ServerError type :-\
+      networkError = new Error('Network error');
+      networkError.statusCode = 401;
```
6a68, line 90: guess I should cast this to the ServerError type?

packages/fxa-settings/src/lib/gql.ts
```diff
@@ -1,8 +1,32 @@
 import { ApolloClient, createHttpLink, from } from '@apollo/client';
 import { setContext } from '@apollo/client/link/context';
-import { onError } from '@apollo/client/link/error';
+import { ErrorHandler, onError } from '@apollo/client/link/error';
 import { cache, sessionToken, typeDefs } from './cache';
 
+export const errorHandler: ErrorHandler = ({ graphQLErrors, networkError }) => {
```
6a68, line 6: extracted this out so I could unit test it

packages/fxa-settings/src/lib/gql.ts
```diff
@@ -1,8 +1,32 @@
...
+        reauth = true;
+      }
+    }
+  }
+  if (networkError && 'statusCode' in networkError) {
+    if (networkError.statusCode === 401) {
+      reauth = true;
+    }
+  }
+  if (reauth && !hasFlowParams) {
```
6a68, line 21: this `hasFlowParams` bit is the part that changed

packages/fxa-settings/package.json
```diff
@@ -71,6 +71,7 @@
     "eslint-plugin-jest": "^23.13.2",
     "eslint-plugin-react": "7.19.0",
     "fxa-shared": "workspace:*",
+    "jest-mock-extended": "^1.0.10",
```
6a68, line 74: I'm definitely not sure this is actually needed, it just made mocking out the `ErrorResponse` arguments easier. Happy to use something else to keep our dependencies down 🤷

dannycoates: in this case since the mocks aren't being used/checked we should be able to just do this in the test file to satisfy the compiler:
```ts
const operation = null as any as Operation;
const forward = null as any as NextLink;
```

packages/fxa-graphql-api/src/lib/auth.ts
dannycoates: This seems fine. The answer to any 4xx error on `/session/status` is reauth anyway

packages/fxa-settings/src/lib/gql.ts
dannycoates: I'm not sure `hasFlowParams` is the right thing to check, or if any addition here is necessary... iiuc the infinite loop was caused by the gql-api server not being able to access the auth-server, and it only sending auth errors. By returning an ApolloError for non-auth errors I think we've fixed that problem. I can imagine a situation where we get a legit auth error and do have flowParams where we still want to reauth instead of showing an error... like a page refresh after you deleted that session elsewhere

tldr: i think the gql server change is all we need/want

6a68: Good point. We definitely might need to reauth when there are flow params. And I'm not sure the flow params should be in the query string anyway, which means this would need to be tweaked whenever that other change happens. I'll remove this bit from the PR.

6a68: I'm still getting some weird reload loop if I nav to localhost:3000/beta/settings locally. Is that just me?

6a68: lol never mind, I should be going to localhost:**3030**/beta/settings 🤦

dannycoates (approved):
a few red squiggles in gql.test.ts but r+ 👍

Issue #6501, opened 2020-09-21
[support form] Add empty select option to the product specific app/service field


data-sync-user:
The field is suppose to be optional, but there's no empty option to make that so.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2552)
┆Issue Number: FXA-2552

ValentinaPC:
Verified as fixed in Stage, using Fx81 and Windows 10 (64-bit).
Postfix screenshot below.
![image](https://user-images.githubusercontent.com/15685960/94408128-0b79bf00-017d-11eb-8edf-f6d2e265425b.png)

PR #6506, opened 2020-09-21
fix(support form): add empty option for app/service field

chenba:
Because:
 - the app/service Support Form field is optional

This commit:
 - add an empty option to the app/service field

Closes: #6501 (FXA-2552)

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/support.js
+++ b/packages/fxa-content-server/app/scripts/views/support.js
@@ -181,6 +181,7 @@ const SupportView = BaseView.extend({
       if (apps.length) {
         this.appEl
           .empty()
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/views/support.js
```diff
@@ -181,6 +181,7 @@ const SupportView = BaseView.extend({
       if (apps.length) {
         this.appEl
           .empty()
+          .append('<option value="">&nbsp;</option>')
```
chenba, line 184: Not pretty, but it's what works with chosen.js for this.

lmorchard (approved):

PR #6514, opened 2020-09-21
feat(db): Create table to store user-customer relationship

jackiemunroe:
## Because:

* We want to be able to tell if an FxA user has an associate Stripe Customer account without having to query Stripe

## This Commit:

* Creates the userCustomers table that will be used to associate FxA user ids with Stripe Customer ids

## Checklist

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-112-113.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-112-113.sql
@@ -1,0 +1,17 @@
+-- -- create accountCustomers table to store the relationship between fxa users and external payment providers
+-- -- stripeCustomerId is nullable by default with the idea that this table can be added upon in the future
+-- -- By creating additional columns for new payment providers. This table will be able to act as a way
+-- -- to quickly determine: is a user also a customer and which payment provider(s) is/are they using?
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('112');
+
+CREATE TABLE IF NOT EXISTS accountCustomers (
+  uid BINARY(16) PRIMARY KEY,
+  stripeCustomerId VARCHAR(32),
+  createdAt BIGINT UNSIGNED NOT NULL,
+  updatedAt BIGINT UNSIGNED NOT NULL
+) ENGINE=InnoDB;
+
+UPDATE dbMetadata SET value = '113' WHERE name = 'schema-patch-level';
--- a/packages/fxa-auth-db-mysql/lib/db/patch.js
+++ b/packages/fxa-auth-db-mysql/lib/db/patch.js
@@ -5,4 +5,4 @@
 // The expected patch level of the database. Update if you add a new
 // patch in the ./schema/ directory.
 
-module.exports.level = 112;
+module.exports.level = 113;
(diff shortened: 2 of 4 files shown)


jackiemunroe:
@lmorchard when you have a sec, can you take a look at this and see if there is anything I overlooked? I was thinking about anything additional we might want to be stored here. I was on the fence about adding the createdAt, but decided to put it in there, since it _could_ be useful.

packages/fxa-auth-db-mysql/lib/db/schema/patch-112-113.sql
```diff
@@ -0,0 +1,10 @@
+-- -- create userCustomers table for storing fxa user to stripe customer relationship
+
+
+CREATE TABLE IF NOT EXISTS userCustomers (
+  fxaUserId VARCHAR(255) PRIMARY KEY,
```
lmorchard, line 5: I think `createdAt` is a good idea. Though, `fxaUserId` should probably instead be `uid binary(16) NOT NULL PRIMARY KEY` for consistency & maybe joins with the `accounts` table. I don't remember why we use a `binary(16)` column here, though. Storage size maybe?

```
✗ mysql --host 127.0.0.1 -u root fxa -e 'show create table accounts'
...
CREATE TABLE `accounts` (
  `uid` binary(16) NOT NULL,
  `normalizedEmail` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
  `email` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
  `emailCode` binary(16) NOT NULL,
  `emailVerified` tinyint(1) NOT NULL DEFAULT '0',
  `kA` binary(32) NOT NULL,
  `wrapWrapKb` binary(32) NOT NULL,
  `authSalt` binary(32) NOT NULL,
  `verifyHash` binary(32) NOT NULL,
  `verifierVersion` tinyint(3) unsigned NOT NULL,
  `verifierSetAt` bigint(20) unsigned NOT NULL,
  `createdAt` bigint(20) unsigned NOT NULL,
  `locale` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `lockedAt` bigint(20) unsigned DEFAULT NULL,
  `profileChangedAt` bigint(20) unsigned DEFAULT NULL,
  `keysChangedAt` bigint(20) unsigned DEFAULT NULL,
  `ecosystemAnonId` text CHARACTER SET ascii COLLATE ascii_bin,
  PRIMARY KEY (`uid`),
  UNIQUE KEY `normalizedEmail` (`normalizedEmail`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci |`
```

lmorchard (changes requested):
Looks good, except for one tweak thanks to FxA weirdness - `uid` is a binary column throughout the DB, so we should do that here too

packages/fxa-auth-db-mysql/lib/db/schema/patch-112-113.sql
```diff
@@ -0,0 +1,10 @@
+-- -- create userCustomers table for storing fxa user to stripe customer relationship
```
dannycoates, line 1: please add
```
SET NAMES utf8mb4 COLLATE utf8mb4_bin;

CALL assertPatchLevel('112');
```

to the top

packages/fxa-auth-db-mysql/lib/db/schema/patch-112-113.sql
```diff
@@ -0,0 +1,10 @@
...
+
+
+CREATE TABLE IF NOT EXISTS userCustomers (
```
lmorchard, line 4: Last nit I could think of here is maybe `userCustomers` should be `accountCustomers` for consistency, but wouldn't block on that

lmorchard (approved):

dannycoates (approved):
💯

lmorchard (approved):
:+1:

Issue #5094, opened 2020-04-23
[services] Create section with header, copy, and "missing" link


data-sync-user:
This should probably be its own component for organizational purposes (`<Services>` perhaps?). Add "Connected Services" header, "Everything you are using and signed in to." text, and "Missing or duplicate items?" link, which should use `LinkExternal` to link out to the SUMO link with UTM params shown in `fxa-content-server/app/scripts/views/settings/client.js`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1755)
┆Issue Number: FXA-1755

data-sync-user:
➤ Dave Justice commented:

got this started here, https://github.com/mozilla/fxa/tree/5094-connected-services-skeleton, ( https://github.com/mozilla/fxa/tree/5094-connected-services-skeleton, ) gunna add some more basic styles then will open pr

PR #6520, opened 2020-09-22
task(fxa-settings): Connected Services skeleton

meandavejustice:
## This pull request

- puts in the basic structure and placeholders for the Connected Services section in settings

Closes: #5094

## Checklist

## Screenshots (Optional)

image
image

### DIFF
--- a/packages/fxa-settings/src/components/ConnectedServices/index.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/index.tsx
@@ -1,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { useState } from 'react';
+import { AlertBar } from '../AlertBar';
+import { ButtonIconReload } from '../ButtonIcon';
+import { useAlertBar } from 'fxa-settings/src/lib/hooks';
+import { useAccount, useLazyAccount } from '../../models';
+import { LinkExternal } from 'fxa-react/components/LinkExternal';
+
+const UTM_PARAMS =
+  '?utm_source=accounts.firefox.com&utm_medium=referral&utm_campaign=fxa-devices';
+const DEVICES_SUPPORT_URL =
+  'https://support.mozilla.org/kb/fxa-managing-devices' + UTM_PARAMS;
+
+export const Service = () => {
+  return (
+    <div className="my-1" id="service" data-testid="settings-connected-service">
+      <div className="border-2 border-solid border-grey-100 rounded flex justify-around items-center">
+        <span>icon</span>
+
+        <div className="flex flex-col">
+          <h2>Service Name</h2>
+          <p className="text-sm">location</p>
+        </div>
+
+        <p>Active Now</p>
+
+        <button
+          className="cta-neutral cta-base disabled:cursor-wait whitespace-no-wrap"
+          data-testid="connected-service-sign-out"
+        >
+          Sign out
+        </button>
+      </div>
+    </div>
+  );
+};
+
+export const ConnectedServices = () => {
+  const alertBar = useAlertBar();
+  const [errorText, setErrorText] = useState<string>();
+  const onError = (e: Error) => {
+    setErrorText(e.message);
+    alertBar.show();
+  };
+  const [getAccount, { accountLoading }] = useLazyAccount((error) => {
+    setErrorText('Sorry, there was a problem refreshing the recovery key.');
+    alertBar.show();
+  });
+
+  return (
+    <section
+      className="mt-11"
+      id="connected-services"
+      data-testid="settings-connected-services"
+    >
+      <h2 className="font-header font-bold ml-4 mb-4">Connected Services</h2>
+      <div className="bg-white tablet:rounded-xl shadow px-6 pt-7 pb-9">
+        <div className="flex justify-between mb-4">
+          <p>Everything you are using and signed into.</p>
+          <ButtonIconReload
+            title="Refresh connected services"
+            classNames="hidden mobileLandscape:inline-block"
+            testId="connected-services-refresh"
+            disabled={accountLoading}
+            onClick={getAccount}
+          />
+        </div>
+
+        <Service />
+        <Service />
+        <Service />
+        <Service />
+        <Service />
+
+        <div className="mt-5">
+          <LinkExternal
+            href={DEVICES_SUPPORT_URL}
+            className="link-blue text-sm"
+          >
+            Missing or duplicate items?
+          </LinkExternal>
+        </div>
+
+        {alertBar.visible && (
+          <AlertBar
+            onDismiss={alertBar.hide}
+            type={errorText ? 'error' : 'success'}
+          >
+            {errorText ? (
+              <p data-testid="delete-recovery-key-error">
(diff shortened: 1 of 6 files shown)


packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+};
+
+export const ConnectedServices = () => {
+  const alertBar = useAlertBar();
+  const [errorText, setErrorText] = useState<string>();
+  const onError = (e: Error) => {
+    setErrorText(e.message);
+    alertBar.show();
+  };
+  const [getAccount, { accountLoading }] = useLazyAccount((error) => {
```
LZoog, line 52: This is part of a separate ticket so we don't need to fix it here but I thought our refresh buttons would only ask for the piece being refreshed, but maybe we want to just requery the entire account?

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <section
+      className="my-1"
+      id="service"
+      data-testid="settings-connected-service"
+    >
+      <div className="border-2 border-solid border-grey-100 rounded flex justify-around items-center">
+        <span>icon</span>
+
+        <div className="flex flex-col">
+          <h2>Service Name</h2>
```
LZoog, line 28: I think this is actually `12px` according to the comps, as well as the location.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <p className="text-sm">location</p>
+        </div>
+
+        <p>Active Now</p>
```
LZoog, line 32: Looks like this should be `14px`

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  'https://support.mozilla.org/kb/fxa-managing-devices' + UTM_PARAMS;
+
+export const Service = () => {
+  return (
+    <section
+      className="my-1"
+      id="service"
+      data-testid="settings-connected-service"
+    >
+      <div className="border-2 border-solid border-grey-100 rounded flex justify-around items-center">
```
LZoog, line 24: Should this be `justify-content: space-between` so the "sign out" button sticks to the right-hand side? This could also use `16px` padding according to the comps, and `rounded-lg` looks a bit closer to the comps than `rounded` to me personally.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <Service />
+        <Service />
+        <Service />
+        <Service />
+        <Service />
+
+        <div className="mt-3 mb-4">
+          <LinkExternal
+            href={DEVICES_SUPPORT_URL}
+            className="link-blue text-sm mt-2 opacity-75 hover:opacity-100 focus:opacity-100"
```
LZoog, line 85: We should probably remove `opacity-75 hover:opacity-100 focus:opacity-100` so `link-blue` stays consistent with all of our links, unless the `link-blue` class should be modified 🤔 Also, `mt-2` doesn't appear to do anything - I think the parent instead of `mt-3` should be `mt-5`, the comps are showing `22px` space which we can round up to 24px.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { LinkExternal } from 'fxa-react/components/LinkExternal';
+
+const UTM_PARAMS =
+  '?utm_source=accounts.firefox.com&utm_medium=referral&utm_campaign=fxa-devices';
+const DEVICES_SUPPORT_URL =
+  'https://support.mozilla.org/kb/fxa-managing-devices' + UTM_PARAMS;
+
+export const Service = () => {
+  return (
+    <section
```
LZoog, line 19: I think these should just be regular 'ol `div`s since the entire "connected services" bit is our "section".

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  return (
+    <section
+      className="mt-3"
+      id="connected-services"
+      data-testid="settings-connected-services"
+    >
+      <h2 className="font-header font-bold ml-4 mb-4">Connected Services</h2>
+      <div className="bg-white tablet:rounded-xl shadow py-4 px-6">
```
LZoog, line 64: I'm seeing `27px` of space between "Everything you are using and signed into" and the top of the section and `37px` at the bottom, so `py-4` should be `pt-7 pb-9` to get pretty close, and then we can take off `mb-4` on the container div of "Missing or duplicate items?"

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    alertBar.show();
+  };
+  const [getAccount, { accountLoading }] = useLazyAccount((error) => {
+    setErrorText('Sorry, there was a problem refreshing the recovery key.');
+    alertBar.show();
+  });
+
+  return (
+    <section
+      className="mt-3"
```
LZoog, line 59: To match the spacing on "Security," this should be `mt-11`

LZoog (commented):
Overall looking good, there's a few comments here but I think everything should be quick to address, though I also noticed mobile styles don't seem to be in place:

image

meandavejustice:
> I also noticed mobile styles don't seem to be in place

Styles for the unit rows aren't included in this issue, in fact the unit rows are just placeholder text, I was getting a head start on https://jira.mozilla.com/browse/FXA-1757

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: Probably just the piece being refreshed, I will address in the refresh ticket

jodyheavener: Ah, I'm assuming Dave is following the precedent I set in #6423, where I noted

> This doesn't actually perform an update query on the specific section, but simply performs an account retrieval and updates the cache accordingly. I could be persuaded to set up individual queries that narrow the columns we look up, but it's all bundled in to one request anyway so 🤷

I'm still on the fence about it tbh, but open for discussion.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: I'm going to ignore these comments for the unit row for individual services since it's going to be addressed in https://jira.mozilla.com/browse/FXA-1757

packages/fxa-settings/src/components/ConnectedServices/index.tsx
LZoog: Ah gotcha, it feels like the server is doing extra requests instead of asking for and receiving the requested information, I think all we just need another query asking only for `attachedClients`. Outside the scope of this issue though, I'll create a follow up for us in JIRA where we can discuss too.

LZoog (approved):
LGTM since the individual unit row styles will be addressed in the same ticket where we display real data, this gives that ticket a good head start 👍

PR #6523, opened 2020-09-22
feat(settings): added passwordChange call to new settings

dannycoates:


packages/fxa-settings/src/components/PageChangePassword/index.tsx
```diff
@@ -2,21 +2,94 @@
...
+            account.passwordCreated = response.authAt * 1000;
+            return account;
+          },
+          session: (existing: Session) => {
+            const session = cloneDeep(existing);
+            session.verified = response.verified;
+            return session;
+          },
+        },
+      });
```
dannycoates, line 47: I debated whether to do this in `auth.ts` but felt like keeping `usePasswordChanger` as simple as possible was the better approach. Having it here mirrors how we handle `useMutation` in other components.

packages/fxa-settings/src/lib/auth.ts
```diff
@@ -19,5 +20,38 @@ export function useAuth() {
...
+      const response = await auth.passwordChange(
+        email,
+        oldPassword,
+        newPassword,
+        {
+          sessionToken,
+        }
+      );
+      return response;
+    },
```
dannycoates, line 51: The meat of the PR

vbudhram (commented):
@dannycoates Code LGTM, however in local testing I get

```
TypeScript error in /Users/vijaybudhram/Desktop/working/fxa/packages/fxa-settings/src/lib/auth.ts(3,22):
Module '"../../../fxa-auth-client/browser"' has no exported member 'AuthServerError'. Did you mean to use 'import AuthServerError from "../../../fxa-auth-client/browser"' instead?  TS2614

    1 | import React, { useContext } from 'react';
    2 | import { useAsyncCallback } from 'react-async-hook';
  > 3 | import AuthClient, { AuthServerError } from 'fxa-auth-client/browser';
      |                      ^
    4 | 
    5 | export interface AuthContextValue {
    6 |   auth?: AuthClient;
```

I suspect its my local setup but still got it after `yarn install`. I have to step out today but can debug it further tomorrow.

vbudhram (approved):
@dannycoates Got it working locally (just needed to restart servers..). Overall the flow works minus all the things you mentioned in description. I did get myself stuck by changing to a 4 char password and then not being able to login since our login requires 8 chars.

Issue #6505, opened 2020-09-21
[Settings][metrics] Emit Amplitude Event


data-sync-user:
Acceptance criteria: an Amplitude event from new Settings is logged correctly in content-server

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2555)
┆Issue Number: FXA-2555

data-sync-user:
➤ Barry Chen commented:

Bumping this up to 5 because adding the service name doesn't seem to be straight forward (i.e. we can't just use the value in teh query param from the redirect).  With that and the new "settings_version" event properties, we'd need to make changes on the content-server.

PR #6534, opened 2020-09-23
feat(settings): emit settings view Amplitude event

chenba:
Because:
 - new settings should emit Amplitude events like old settings

This commit:
 - build a metrics payload for an Amplitude event that will result in a
   `amplitudeEvent` log entry on content-server
 - redirect to content server with the given query params intact so that
   content server could init with those values
 - add the "settings_version" Amplitude event properties
 - emit the "fxa_pref - view" event

Closes: #6505 (FXA-2555)

## Checklist


vbudhram (commented):
@chenba This looks good, however I found a few differences between the current settings view and beta settings.

Current:
```json
{"op":"amplitudeEvent","event_type":"fxa_pref - view","time":1600960759830,"user_id":"7a12251dc7ba4ffc8efab9872c591f07","device_id":"2ebc428839e2418d9c637c36a888af4e","session_id":1600960758902,"app_version":"188.1","language":"en-US","os_name":"Mac OS X","os_version":"10.15","event_properties":{"service":"undefined_oauth","oauth_client_id":"5882386c6d801776"},"user_properties":{"entrypoint":"fxa_discoverability_native","flow_id":"a895663550507099b4ddfde04b90433f0685135122826d0b04ce70b5c14d53a0","ua_browser":"Firefox","ua_version":"80.0","$append":{"fxa_services_used":"undefined_oauth","two_step_authentication":false,"emails":false,"account_recovery":false}}}
```

beta
```
{"op":"amplitudeEvent","event_type":"fxa_pref - view","time":1600960749981,"user_id":"7a12251dc7ba4ffc8efab9872c591f07","device_id":"8467500e1f0948cdbaf7fa44d7fa97ff","session_id":1600960741297,"app_version":"188.1","language":"en","os_name":"Mac OS X","os_version":"10.15","event_properties":{},"user_properties":{"entrypoint":"fxa_discoverability_native","flow_id":"dad9ea4964cfcfb04672794a466ab3b07de80b37d4c5020f03761e33f222eef8","ua_browser":"Firefox","ua_version":"80.0","$append":{"account_recovery":false,"emails":false,"two_step_authentication":false}}}
```

The new beta pages don't set a service name in the event properties. In the current settings page, I believe the query param `context=fx_desktop_v3` or `service=sync` gives a hint to what the service could be.

In the content-server the [relier](https://github.com/mozilla/fxa/tree/bc60082808540ef778a2cbf7d252e00fe8b8af1e/packages/fxa-content-server/app/scripts/models/reliers) models store the service name. I think we could probably do something similar in beta settings, where we have a new model to store this or put into the account model.

Finally, I wonder if we should add a new event/user property to tell which version of the page is being viewed (ie old vs beta settings).

chenba:
> The new beta pages don't set a service name in the event properties.

Ok, I'll take a look at this.  Thanks for catching it! 

> 
> Finally, I wonder if we should add a new event/user property to tell which version of the page is being viewed (ie old vs beta settings).

Yeah, I considered this as well when we initially went over metrics during the planning/estimating phase.  I'll check with Alex and Leif.

packages/fxa-content-server/app/scripts/views/get_flow.js
```diff
@@ -49,11 +49,14 @@ class GetFlowView extends BaseView {
...
-    const service = this.metrics.getFilteredValue('service');
-    const uniqueUserId = this.metrics.getFilteredValue('uniqueUserId');
+
+    const {
+      broker,
+      context,
+      isSampledUser,
+      service,
+      uniqueUserId,
+    } = this.metrics.getFilteredData();
```
chenba, line 59: No reason to have `getFilteredValue` to repeatedly call `getFilteredData`.

packages/fxa-settings/src/lib/metrics.ts
```diff
@@ -161,10 +180,14 @@ export function init(flowQueryParams: FlowQueryParams) {
         redirectPath += window.location.search;
       }
 
+      const search = window.location.search
+        ? `${window.location.search}&redirect_to=${encodeURIComponent(
+            redirectPath
+          )}`
+        : `?redirect_to=${encodeURIComponent(redirectPath)}`;
```
chenba, line 187: The redirect need to pass the query params to content server since content server's app-start rely on some of those values.  This change allows us to get the correct `service` value at the very least.

packages/fxa-shared/metrics/amplitude.js
```diff
@@ -111,8 +111,8 @@ function mapSettingsEventProperties(...args) {
 }
 
 function mapSettingsVersion(eventType, eventCategory, eventTarget, data) {
-  if (data && data.settings_version) {
-    return { settings_version: data.settings_version };
+  if (data && data.settingsVersion) {
+    return { settings_version: data.settingsVersion };
```
chenba, line 115: Updated to be consistent with camelCase in request payload but snake_case in Amplitude event properties.

Also, this already exists!  But the rest of the code required to get here was not.

packages/fxa-content-server/server/lib/validation.js
```diff
@@ -49,6 +49,7 @@ const TYPES = {
     .uri({ scheme: ['android-app', 'http', 'https'] })
     .allow('none'),
   RESUME: joi.string().regex(PATTERNS.BASE64),
+  SETTINGS_VERSION: joi.string().valid('old', 'new'),
```
vbudhram, line 52: I'm not 100% sold on the values here but don't have any suggestions either 😅

vbudhram (approved):
Thanks @chenba, this LGTM 👍🏽

```
old
{"op":"amplitudeEvent","event_type":"fxa_pref - view","time":1601392523150,"user_id":"b4718a4092c646d1abd5b5e8c0d042e9","device_id":"764f589ec8d940ffb90569d88a3de18e","session_id":1601392522134,"app_version":"188.1","language":"en-US","os_name":"Mac OS X","os_version":"10.15","event_properties":{"service":"undefined_oauth","oauth_client_id":"5882386c6d801776","settings_version":"old"},"user_properties":{"entrypoint":"fxa_discoverability_native","flow_id":"b72a6f591c9d7bd5b9ee442e323c172dd5352fad52dba68bcfd06ff5802fbb51","ua_browser":"Firefox","ua_version":"80.0","$append":{"fxa_services_used":"undefined_oauth","two_step_authentication":false,"emails":false,"account_recovery":false}}}

new
{"op":"amplitudeEvent","event_type":"fxa_pref - view","time":1601392553041,"user_id":"b4718a4092c646d1abd5b5e8c0d042e9","device_id":"91a89c88e35a4fa4a3c9a72ef93c0c90","session_id":1601392552003,"app_version":"188.1","language":"en","os_name":"Mac OS X","os_version":"10.15","event_properties":{"service":"undefined_oauth","oauth_client_id":"5882386c6d801776","settings_version":"new"},"user_properties":{"entrypoint":"fxa_discoverability_native","flow_id":"b6b6726fed1a111f89df33334ddcf5f6db044557965a67dce516eb840a0ddf6a","ua_browser":"Firefox","ua_version":"80.0","$append":{"fxa_services_used":"undefined_oauth","account_recovery":false,"emails":false,"two_step_authentication":false}}}
```

Only difference between the two metrics (besides version) is the language property (en-US vs en), this is probably ok to land as is and fix in follow up.

packages/fxa-content-server/server/lib/validation.js
chenba: These were suggested by @irrationalagent on Slack.

chenba:
> 
> Only difference between the two metrics (besides version) is the language property (en-US vs en), this is probably ok to land as is and fix in follow up.

That 'en' is from the lang attr in `fxa-settings/public/index.html`.

Issue #5096, opened 2020-04-23
[services] Create necessary client-side filters/mods for displaying data


data-sync-user:
Create a copy of the data received from the GQL query and make additions/adjustments as needed (or whatever works - we could store it in the apollo client cache as `filteredClients`). Clients should be organized as "active now" first followed by the most recent to the most distant last login. Ensure "last active" is formatted correctly.

OAuth clients need to be reduced into a single, most recent instance, along with a link out. "Revoke" is no longer present.

Some relevant files from the content server:

* models/attached-client
* models/attached-clients.js
* views/settings/clients.js

See https://mozilla.invisionapp.com/share/FTWHZ3LPR2M#/screens/414612859.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1756)
┆Issue Number: FXA-1756

Issue #5095, opened 2020-04-23
[services] Map through filtered data w/row component, update styles, add dummy button


data-sync-user:
We should be able to `.map` through the filtered data provided by FXA-1756 with a reusable row component (`<ServiceRow>`?). Style accordingly.

Add a dummy "Sign out" button for every device.

Edit: See comments. We need to make sure the link-outs (per the comps) from OAuth clients work, that tests are added for the filtered data, and it's styled to match Figma.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1757)
┆Issue Number: FXA-1757
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/14101/attachedclients.txt">attachedclients.txt</a>

data-sync-user:
➤ Dave Justice commented:

Styles started in https://github.com/mozilla/fxa/pull/6520, ( https://github.com/mozilla/fxa/pull/6520, ) need to address comments here, check in mobile, and test closely against figma

data-sync-user:
➤ Dave Justice commented:

> Since the OAuth clients will need to link-out to Monitor, Pocket etc., are you planning on say, maybe making a map of 'name': 'www.url-to-wherever.com' and pulling that in when working on #5095 ( https://github.com/mozilla/fxa/issues/5095 )? (We may want to bump that issue up to a 2)

https://github.com/mozilla/fxa/pull/6537#discussion_r493930091 ( https://github.com/mozilla/fxa/pull/6537#discussion_r493930091 )

data-sync-user:
➤ Dave Justice commented:

We'll also need to fill in test coverage for the filtering logic. I didn't include it in https://github.com/mozilla/fxa/pull/6537 ( https://github.com/mozilla/fxa/pull/6537 ) because I figured I may change it a bit and was planning to finish out the presentation tests. Added a story point to account for this

data-sync-user:
➤ Lauren Zugai commented:

Also: when pasting dummy data in, for some reason the location seems to have trouble rendering in a couple spots and I'm seeing commas. I'm going to attach the dummy data here (thanks Vlad!)

The `{name}` also should not be an `h2`, since the icon comes first I'm tempted to say it should just be a regular `p` (or `span`) tag rather than an `h3`. Take a look at it as a whole and go with what makes sense there.

PR #6537, opened 2020-09-23
task(fxa-settings): filtering connected devices

meandavejustice:
## This pull request

- adds filtering of the attachedClients and passes them, rudimentarily, into the `<Service/>` component.

Closes: #5096

## Checklist

We are probably going to want to add this to the cache, but I'm planning to look into that once we get this working and styled all in the settings page. The reason for this is that we may not need to store all of these fields. I'm going to get rid of any access data at that stage 👍. The reason I don't go through all of them in this pr, is because I'm not sure all of what I will need to determine which icon will be displayed, so keeping them around until I get that fixed up, I assume I'll be working on that in the next sprint.


packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -69,11 +117,18 @@ export const ConnectedServices = () => {
...
-        <Service />
-        <Service />
-        <Service />
-        <Service />
-        <Service />
+        {!!sortedAndUniqueClients.length &&
+          sortedAndUniqueClients.map((client, i) => (
+            <Service
+              {...{
+                key: i,
```
LZoog, line 124: Is there another `key` we can use like `clientId`? We shouldn't use the index of an array.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -6,40 +6,88 @@ import React, { useState } from 'react';
...
+
+  // move currently active client to the top
+  sortedAndUniqueClients.forEach((client, i) => {
+    if (client.isCurrentSession) {
+      sortedAndUniqueClients.splice(i, 1);
+      sortedAndUniqueClients.unshift(client);
+    }
+  });
+
+  return sortedAndUniqueClients;
```
LZoog, line 40: Since the OAuth clients will need to link-out to Monitor, Pocket etc., are you planning on say, maybe making a map of `'name': 'www.url-to-wherever.com'` and pulling that in when working on #5095? (We may want to bump that issue up to a 2)

LZoog (commented):
One suggestion/one question, and what about tests to make sure the sorting is as expected?

(I can take a more thorough look tomorrow AM, this was a quick pass)

packages/fxa-settings/src/models/Account.ts
```diff
@@ -6,12 +6,34 @@ import {
...
+  deviceId: string;
+  name: string;
+  lastAccessTime: number;
+  lastAccessTimeFormatted: string;
+  approximateLastAccessTime: number;
+  approximateLastAccessTimeFormatted: string;
+  location: Location;
+  os: string;
+}
+
```
bbangert, line 36: Oh bummer, we're duplicating the types from the GraphQL schema? Hmmmmm,
https://www.leighhalliday.com/generating-types-apollo

jodyheavener: Whaaaat I swear I inquired about this back at the beginning of Settings work. Very good to know!

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: That sounds like a good option. bumped to a 2 and added your comment to the jira task

packages/fxa-settings/src/models/Account.ts
meandavejustice: Nice, do we have an issue filed for this? Good to keep in mind for the next maintenance sprint.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: Just switched to clientId

meandavejustice:
> what about tests to make sure the sorting is as expected?

We should _definitely_ have tests over the sorting. I knew that some of this dom was going to be changing once I finish up the styles and add the links to connected services, so I was putting off tests until then. I'd like to test the filtering separate from the presentation code also. So I'll add a note to that issue and bump it up a story point so this pr can make it into this sprint.

packages/fxa-settings/src/models/Account.ts
meandavejustice: Filed https://jira.mozilla.com/browse/FXA-2573

LZoog (approved):
There's a couple of UI tweaks that we could do but this issue was around the sorting which seems to work as expected so r+'ing, presentation stuff will be addressed in #5095 👍

Issue #5320, opened 2020-05-14
Enable failed tests insights in CircleCI


clouserw:
There is a page of insights for our CircleCI runs here:  https://circleci.com/build-insights/gh/mozilla/fxa/master

It has sections at the bottom for most failed tests and slowest failed tests that aren't populated because we're not sending the data to CircleCI.  In our meeting last week it sounded like we were already generating these xml files, we just weren't copying them to the right place for CircleCI to pick up.  Let's start sending this data to CircleCI.

CircleCI docs are at https://circleci.com/docs/2.0/collect-test-data/

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1887)
┆Issue Number: FXA-1887

data-sync-user:
➤ Wil Clouser commented:

Do we need to connect to the CircleCI API to access these numbers now?

PR #6546, opened 2020-09-24
fix(tests): Store test results in insights

vbudhram:
Looks like this broke a little while back and we were logging the results for content-server.

Fixes https://github.com/mozilla/fxa/issues/5320

### DIFF
--- a/packages/fxa-content-server/tests/intern.js
+++ b/packages/fxa-content-server/tests/intern.js
@@ -76,7 +76,15 @@ const config = {
   fxaPaymentsRoot,
 
   pageLoadTimeout: 20000,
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/tests/intern.js
```diff
@@ -76,7 +76,15 @@ const config = {
   fxaPaymentsRoot,
 
   pageLoadTimeout: 20000,
-  reporters: 'runner',
+  reporters: [
```
vbudhram, line 79: Seemed more useful to make this the default since it would be used in other test suites.

dannycoates (approved):
💯 👍

PR #6547, opened 2020-09-24
feat(shared): add CRUD methods for maintaining fxa user and customer relationship

### LINKED ISSUE #6159: Implement DB API for managing Stripe customer / FxA user associations
Once we have a place in the DB to store the association between FxA user and Stripe customer, we need to implement support for auth-server code to manage and query the associations.

Thinking this should just be basic Create / Read / Update / Delete operations per user to manage association with Stripe customer ID.

Two paths for this:

1. follow the FxA tradition of adding stored procedures and fxa-auth-db-mysql HTTP APIs
1. or, add direct SQL queries to a library in fxa-shared that both auth-server and the new gql-server can use

The latter option may require some ops cooperation to ensure auth-server can directly access the DB. But, it's probably the better option looking forward, since this is a new query - and considering FXA-2335 and that we've been considering decommissioning fxa-auth-db-mysql by moving to direct queries in fxa-shared.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2400)
┆Issue Number: FXA-2400

jackiemunroe:
## Because
* We want to be able to maintain the relationships between customers and users

## This pull request

* Establishes CRUD methods to maintain this relationship

closes #6159

## Checklist

### DIFF
--- a/packages/fxa-shared/test/db/models/auth/helpers.ts
+++ b/packages/fxa-shared/test/db/models/auth/helpers.ts
@@ -27,6 +27,10 @@ export const emailsTable = fs.readFileSync(
   path.join(thisDir, './emails.sql'),
   'utf8'
 );
+export const accountCustomersTable = fs.readFileSync(
+  path.join(thisDir, './account-customers.sql'),
+  'utf8'
+);
 
 export function randomAccount() {
   const email = chance.email();
@@ -85,6 +89,7 @@ export async function testDatabaseSetup(): Promise<Knex> {
 
   await knex.raw(accountTable);
   await knex.raw(emailsTable);
+  await knex.raw(accountCustomersTable);
 
   /* Debugging Assistance
   knex.on('query', (data) => {
--- a/packages/fxa-shared/.vscode/launch.json
+++ b/packages/fxa-shared/.vscode/launch.json
@@ -36,6 +36,25 @@
       ],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
-    }
+    },
+    {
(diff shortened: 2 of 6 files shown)


packages/fxa-shared/test/db/models/auth/helpers.ts
```diff
@@ -10,6 +10,7 @@ import Knex from 'knex';
 
 import { setupAuthDatabase } from '../../../../db';
 import { Account } from '../../../../db/models/auth/account';
+import { AccountCustomers } from '../../../../db/models/auth/account-customers';
```
chenba, line 13: This is not used.

yarn.lock
```diff
@@ -32978,6 +32978,7 @@ resolve@~1.11.1:
   dependencies:
     atob: ^2.1.2
     decode-uri-component: ^0.2.0
+  checksum: 9bbd2656937ce422f1526026638a91dacba0822f1ab954ac34728c9b4f2960dbfc22a8463af9fbf74f28f3722263444df79b2b5ffa07de07dbcda2ec5a8c7513
```
chenba, line 32981: Let's back out these yarn.lock changes.

chenba (approved):
👌

chenba:
Was there a previous discussion on why we'd want the models in shared?

jackiemunroe:
> Was there a previous discussion on why we'd want the models in shared?

Yes. The conversation was mostly in Slack and, I think during one of our stand-ups, but the gist is related in this issue: https://github.com/mozilla/fxa/issues/6338

Issue #6086, opened 2020-07-30
Enabled code cov carry flags to get better reporting


vbudhram:
Code cov carry flags are meant to be used in monorepos and iterative testing projects. We need to implement this to get a better overall picture of our coverage.

https://docs.codecov.io/docs/carryforward-flags

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2366)
┆Issue Number: FXA-2366

PR #6552, opened 2020-09-25
fix(tests): Add carryflags to codecov

vbudhram:
## Because

- Our code coverage metrics are not always accurate

## This pull request

- Adds support for the [carryflags](https://docs.codecov.io/docs/carryforward-flags) which in theory should help. Not 100% sure I set it up correctly since it needs to be merged to master and get a base commit for coverage reports. Doesn't hurt to try

Closes: https://github.com/mozilla/fxa/issues/6086 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
           path: artifacts/tests
       - run:
           name: Reporting code coverage...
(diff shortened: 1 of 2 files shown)


chenba (approved):
Let's try it!

Issue #6160, opened 2020-08-07
Write a script to populate the DB with FxA user / Stripe customer associations


data-sync-user:
Once we have a DB table associating FxA users with Stripe customers (FXA-2399), we'll need to pre-load it with the existing data.

This is **probably** also blocked by FXA-2400 so the script can access the DB, but maybe we can make direct SQL queries?

We have an existing script that does this for Redis cache by walking through all Stripe customers - it could be adapted to write to the DB instead.

https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/scripts/populate-stripe-customer-cache.js

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2401)
┆Issue Number: FXA-2401

PR #6554, opened 2020-09-25
feat(auth): create script to populate accountCustomers table

jackiemunroe:
## Because

- To use the accountCustomers table effectively, we need to prepopulate it with Stripe data

## This pull request

- Create script to populate accountCustomers table from existing customers in Stripe

Closes: #6160

## Checklist

### DIFF
--- a/packages/fxa-auth-server/test/scripts/fixtures/account-customers.sql
+++ b/packages/fxa-auth-server/test/scripts/fixtures/account-customers.sql
@@ -1,0 +1,6 @@
+CREATE TABLE `accountCustomers` (
+  `uid` BINARY(16) PRIMARY KEY,
+  `stripeCustomerId` VARCHAR(32),
+  `createdAt` BIGINT UNSIGNED NOT NULL,
+  `updatedAt` BIGINT UNSIGNED NOT NULL
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
--- a/packages/fxa-auth-server/.vscode/launch.json
+++ b/packages/fxa-auth-server/.vscode/launch.json
@@ -148,6 +148,28 @@
       },
       "preLaunchTask": "Stop PM2 Auth Server",
       "postDebugTask": "Start PM2 Auth Server"
-    },
+    }, {
+      "type": "node",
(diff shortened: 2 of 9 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -103,6 +103,7 @@ export class StripeHelper {
       : undefined;
 
     this.stripe = new Stripe(config.subscriptions.stripeApiKey, {
+      // @ts-ignore
```
bbangert, line 106: What is being ignored?

packages/fxa-auth-server/scripts/populate-stripe-customer-db.js
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env ts-node-script
```
bbangert, line 1: If this is using ts-node, shouldn't it be a TypeScript file?

packages/fxa-auth-server/lib/payments/stripe.ts
jackiemunroe: Stripe has updated their version, according to their docs, this is the work around until we update to the latest API version.

packages/fxa-auth-server/scripts/populate-stripe-customer-db.js
jackiemunroe: I was following the convention of the other scripts and was uncertain if it should be .ts or .js. I can update that.

packages/fxa-auth-server/test/scripts/fixtures/account-customers.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE `accountCustomers` (
```
bbangert, line 1: Should these fixtures be closer to the code using them (in fxa-shared)?

bbangert (approved):
It would probably be better to have the fixtures closer to the models, otherwise this looks good!

jackiemunroe:
> It would probably be better to have the fixtures closer to the models, otherwise this looks good!

@bbangert I can clean that up in the upcoming related PRs 👍

Issue #6564, opened 2020-09-28
[payments][metrics] Amplitude event session_id can be non-numeric


data-sync-user:
The payload endpoint's validation for `flowBeginTime` is in the wrong place.  It needs to be part of the `data` validation.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2582)
┆Issue Number: FXA-2582

PR #6565, opened 2020-09-28
fix(metrics): fix payload validation for payments metrics

chenba:
Because:
 - the payload validation on Payments metrics endpoint was not
   validating the flow begin time and the flow id

This commit:
 - move the Joi validators into the correct spot

Closes: #6564

## Checklist


vbudhram (approved):
@chenba Thanks, are we putting this on a point release?

chenba:
> 
> 
> @chenba Thanks, are we putting this on a point release?

Considering that the bug's been around as long as the endpoint, I don't think that'd be necessary.

PR #6574, opened 2020-09-29
feat(settings): added validation to change password

dannycoates:


packages/fxa-settings/src/components/InputText/index.tsx
```diff
@@ -21,9 +21,10 @@ export type InputTextProps = {
   errorText?: string;
   errorTooltipClass?: string;
   className?: string;
-  inputRef?: RefObject<HTMLInputElement>;
+  inputRef?: Ref<HTMLInputElement>;
   onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
   type?: 'text' | 'email' | 'tel' | 'number' | 'url' | 'password';
+  name?: string;
```
dannycoates, line 27: `react-hook-form` requires a `name`

packages/fxa-settings/src/components/InputPassword/index.tsx
```diff
@@ -42,13 +43,16 @@ export const InputPassword = ({
...
+        name,
       }}
     >
       <button
+        type="button"
         data-testid="visibility-toggle"
         className={`w-5 px-3 py-2 text-grey-600 focus:text-blue-500 box-content ${
           hasContent ? '-ml-3' : 'hidden'
         }`}
+        tabIndex={-1}
```
dannycoates, line 55: otherwise the tab flow is weird

vbudhram (commented):
Better late than never 😬, this LGTM 👍

Issue #6530, opened 2020-09-23
Update functional test helpers to support async await


vbudhram:
Currently, the functional test helpers use a [thenify](https://github.com/mozilla/fxa/blob/3b0acdb34b3c768f9210cd4edf0aa0a397d7efee/packages/fxa-content-server/tests/functional/lib/helpers.js#L42) function to make it useable in a `.then` chain.

We should update these functions so that they support async/await.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2568)
┆Issue Number: FXA-2568

PR #6577, opened 2020-09-30
fix(tests): update functional testing docs and settings tests

### LINKED ISSUE #6528: Document functional testing in settings V2
We should document how to write/run and test this locally specifically for content-server. Most of the information in https://github.com/mozilla/ecosystem-platform/blob/master/docs/fxa-engineering/functional-testing.md is still relevant but could have a new section just for settings V2.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2566)
┆Issue Number: FXA-2566

vbudhram:
## Because

- We should probably document how to create functional tests
- While Internjs does support async/await, our test helpers are not in test most convenient format for them

## This pull request

- Adds docs
- Wrap our functional test helpers so that they can be used in async/await 

Closes: #6530 Closes #6528 

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/lib/helpers.js
+++ b/packages/fxa-content-server/tests/functional/lib/helpers.js
@@ -2890,3 +2890,17 @@ module.exports = {
   visibleByQSA,
   waitForUrl,
 };
+
+// Export helpers methods in a form that can be easily used in
+// async/await. The usage of the helpers are the same except
+// they expect the last argument to be the intern remote object.
+const helpersRemoteWrapped = {};
+const helpers = { ...module.exports };
+Object.keys(helpers).forEach((key) => {
+  helpersRemoteWrapped[key] = async function () {
+    const args = [...arguments];
+    const remote = args.pop();
+    return remote.then(helpers[key](...args));
+  };
+});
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/tests/functional/lib/helpers.js
```diff
@@ -2890,3 +2890,17 @@ module.exports = {
...
+// they expect the last argument to be the intern remote object.
+const helpersRemoteWrapped = {};
+const helpers = { ...module.exports };
+Object.keys(helpers).forEach((key) => {
+  helpersRemoteWrapped[key] = async function () {
+    const args = [...arguments];
+    const remote = args.pop();
+    return remote.then(helpers[key](...args));
+  };
+});
```
chenba, line 2905: nit: copying `module.exports` to `helpers` doesn't seem necessary.

chenba (approved):
👍

packages/fxa-content-server/tests/functional/lib/helpers.js
chenba: oops sorry, I wasn't very clear with my comment.  Couldn't we use `module.exports` instead of making a copy with `{...module.exports}`?

Issue #6161, opened 2020-08-07
Update StripeHelper methods to manage DB association between FxA user and Stripe customer


data-sync-user:
Once we have HTTP APIs to manage FxA user / Stripe customer associations (FXA-2400), we need to use those APIs in our StripeHelper. 

Main points of interest are probably these methods, for creating and removing the association respectively:

* createCustomer: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/payments/stripe.ts#L307
* removeCustomer: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/payments/stripe.ts#L483

Also, filing FXA-2403 as a separate issue to actually start _using_ this association. 

I think we could get all the data and its management merged and deployed as a separate sprint than actually starting to use it, which could be good to pair with having ops run the pre-population script from FXA-2401. Essentially, get all the groundwork in place before relying on it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2402)
┆Issue Number: FXA-2402

Issue #5850, opened 2020-07-02
Remove dead subscription API routes and code from the server


data-sync-user:
After the client-side switches to using the new API methods for Payment and Setup Intents, the unused routes and code should be removed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2229)
┆Issue Number: FXA-2229

data-sync-user:
➤ Barry Chen commented:

Instead of removal, we should have the endpoints behind a feature flag (the same flag as FXA-2301?).  That would allow QA to create pre-SCA customers in a test env (see FXA-2303).

data-sync-user:
➤ Les Orchard commented:

We do want to remove the code eventually as cleanup. Also talking about allowing QA to flip between new & old UI, which would mean that we don't disable the new server endpoints behind a feature flag and allow both old & new APIs to be used at the same time

data-sync-user:
➤ Barry Chen commented:

I'm moving this to the $ubPlat Polish epic, because:

  * we want the old code around for a while for QA purposes
  * this shouldn't block the SCA release

PR #6580, opened 2020-09-30
feat(auth): Add Create/Delete for Account-Customer Relationship

jackiemunroe:
## Because

- We want to have a persistent record connection account ids to stripe customer ids

## This pull request

- Adds relationship creation and deletion to the Stripe Customer Creation and Deletion flows

closes: #6161

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -5,46 +5,12 @@ import convict from 'convict';
 import fs from 'fs';
 import path from 'path';
 import url from 'url';
+import { makeMySQLConfig } from 'fxa-shared/db/config';
 
 const DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages');
 
 convict.addFormats(require('convict-format-with-moment'));
 convict.addFormats(require('convict-format-with-validator'));
-
-function makeMySQLConfig(envPrefix: string, database: string) {
-  return {
-    database: {
-      default: database,
-      doc: 'MySQL database',
-      env: envPrefix + '_MYSQL_DATABASE',
-      format: String,
-    },
-    host: {
-      default: 'localhost',
-      doc: 'MySQL host',
-      env: envPrefix + '_MYSQL_HOST',
-      format: String,
-    },
-    password: {
-      default: '',
-      doc: 'MySQL password',
-      env: envPrefix + '_MYSQL_PASSWORD',
-      format: String,
-    },
-    port: {
-      default: 3306,
-      doc: 'MySQL port',
-      env: envPrefix + '_MYSQL_PORT',
-      format: Number,
-    },
-    user: {
-      default: 'root',
-      doc: 'MySQL username',
-      env: envPrefix + '_MYSQL_USERNAME',
-      format: String,
-    },
-  };
-}
 
 const conf = convict({
   env: {
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -4,6 +4,11 @@
 import * as Sentry from '@sentry/node';
 import cacheManager, { Cacheable, CacheClear } from '@type-cacheable/core';
 import { useAdapter } from '@type-cacheable/ioredis-adapter';
+import {
+  createAccountCustomer,
+  deleteAccountCustomer,
+  getAccountCustomerByUid,
+} from 'fxa-shared/db/models/auth';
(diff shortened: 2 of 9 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -11,6 +11,41 @@ const DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages');
 convict.addFormats(require('convict-format-with-moment'));
 convict.addFormats(require('convict-format-with-validator'));
 
+function makeMySQLConfig(envPrefix: string, database: string) {
```
bbangert, line 14: I thought this function was exported somewhere in `fxa-shared`?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -120,6 +126,9 @@ export class StripeHelper {
         statsd.timing('stripe_request', response.elapsed);
       });
     }
+
+    // Establish database connection and bind instance to Model using Knex
+    setupAuthDatabase(config.database.mysql.auth);
```
bbangert, line 131: Setup tasks like this should be in `bin/key_server.ts` where other root objects are instantiated. The Redis instance should've also been setup there, and then fetched using `typedi`, but we didn't have that in place back then. We actually have the `statsd` object registered with a container now, so passing it all the way down the chain isn't needed either, this could prolly use some cleanup in general.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -327,7 +338,7 @@ export class StripeHelper {
     idempotencyKey: string
   ): Promise<Stripe.Customer> {
     try {
-      return await this.stripe.customers.create(
+      const stripeCustomer = await this.stripe.customers.create(
```
bbangert, line 341: Do we use these flows anymore?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -502,9 +515,16 @@ export class StripeHelper {
    * - remove the cache entry
    */
   async removeCustomer(uid: string, email: string) {
-    const customer = await this.fetchCustomer(uid, email);
-    if (customer) {
-      await this.stripe.customers.del(customer.id);
+    const customer = await getAccountCustomerByUid(uid);
+    if (customer && customer.stripeCustomerId) {
+      await this.stripe.customers.del(customer.stripeCustomerId);
+      const recordsDeleted = await deleteAccountCustomer(uid);
```
bbangert, line 521: If this fails, will we have enough information to locate the user manually to delete the record?

packages/fxa-auth-server/config/index.ts
jackiemunroe: No, `makeMySQLConfig` wasn't exported to `fxa-shared` just the interface `MySQLConfig` was extracted. There was some issue with exporting/importing `makeMySQLConfig`, that would result in an error when trying to use it to generate the configs: `makeMySQLConfig is not a function`

packages/fxa-auth-server/lib/payments/stripe.ts
jackiemunroe: Yes. If you look at the check below: if there are no records affected (the delete calls returns an integer indicating the number of rows that were removed), the user's account id is logged along with a message indicating that the item wasn't removed.

packages/fxa-auth-server/lib/payments/stripe.ts
lmorchard: Hmm, I don't think so - must be a spot I missed in #5850. I think we just use `createPlainCustomer` now

packages/fxa-auth-server/config/index.ts
chenba: Was `makeMySQLConfig` undefined or something else, since it wasn't a function?

jackiemunroe: It was defined and exported, but I was getting a typescript compilation error

jackiemunroe: I'll give it another go and see if I get the same issue

packages/fxa-auth-server/lib/payments/stripe.ts
jackiemunroe: I can remove the method if it is safe to @lmorchard

bbangert (approved):

Issue #6503, opened 2020-09-21
No Longer Tracking Activity for Sync on Desktop


data-sync-user:
On desktop, we normally track 2 types of activity metrics:

1) Are they signed into an account or not? `service=firefox-desktop`
2) Are they signed into Sync specifically? `service=sync`

We are no longer tracking if they are signed into Sync on desktop.

See (ignore chart title): 
https://analytics.amplitude.com/mozilla-corp/chart/alh9e7d/edit/46ajhvp

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2554)
┆Issue Number: FXA-2554

rfk:
Quick drive-by to note, looks like this drop corresponds to the release of Firefox 80 and its switch to using OAuth for Sync access.

data-sync-user:
➤ Barry Chen commented:

Some optional related reading material:

  * https://bugzilla.mozilla.org/show_bug.cgi?id=1632635 ( https://bugzilla.mozilla.org/show_bug.cgi?id=1632635 ) The origin of "fxa_activity - active"
  * https://github.com/mozilla/fxa/pull/6383 ( https://github.com/mozilla/fxa/pull/6383 ) PR for emitting a "fake" 'account.signed' ('fxa_activity - cert_signed') event
  * https://github.com/mozilla/fxa/pull/5820 ( https://github.com/mozilla/fxa/pull/5820 ) The origin of "fxa_activity - oauth_access_token_create"
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1591312 ( https://bugzilla.mozilla.org/show_bug.cgi?id=1591312 ) Restoration 'cert_signed' with 'service=sync'

The event in question here is "cert_signed", which no longer includes the service in its event properties after Sync's switch to OAuth.  Prior to the switch, the event is emitted when Sync sends a request to `/certificate/sign` with the query param 'service=sync'.  After the switch, this event is "faked" within the `/oauth/token` handler.  However, it does not include the service name in the event properties since there isn't one in the request.

What we could do (should do?) on the FxA side is to include `{service: sync}` in the new events emitted from the `/oauth/token` when "oldsync" is in the scopes.

I was made aware of https://github.com/mozilla/fxa/issues/6578 ( https://github.com/mozilla/fxa/issues/6578 ) literally in the middle of writing this comment, so perhaps we should also _only_ emit those events when "oldsync" is in the scope?

[...]

Issue #6578, opened 2020-09-30
Sync device counts should be consistent during a login flow


6a68:
While signing in to sync, the desktop browser makes a number of auth-server requests:

```
/account/login
/oauth/token <-- issues 'profile' token
/verify
/account/keys
/account/device <-- device count is incremented here
/oauth/token <-- issues 'oldsync' token
/account/scoped-key-data
```

(If interested, you can test this yourself by going to about:config, creating a boolean pref `identity.fxaccounts.allowHttp` and setting it to true, then filling in `identity.fxaccounts.autoconfig` with `http://localhost:3030`, then restarting. You'll now be able to sign in to local FxA from the browser toolbar FxA button, and you can watch the auth-server logs or network tab in devtools.)

Some events are logged in the course of this flow that include device info about the account, a little JS object of the form

```
{ sync_device_count, sync_active_devices_day, sync_active_devices_week, sync_active_devices_month }
```

Based on when in this flow the events are logged, they will carry the wrong device count.

Events logged before the device is incremented include:
* `fxa_email - sent`
* `fxa_activity - oauth_access_token_created` for the `profile` scope (this is why sync device counts are off by one)
* `fxa_activity - cert_signed` (not because we sign a cert, but to trigger `fxa_login - complete`, see #6347)
* `fxa_login - complete`

Events logged after the device count is incremented:
* `fxa_activity - oauth_access_token_created` for the `oldsync` scope (but missing the `event_properties: {service: sync}` on the event, which seems to exclude it from the sync device counts rollup)
* `fxa_activity - cert_signed` (not yet sure why this event is fired twice, but probably related to #6347)

We should rethink how we are recording these numbers, since the current approach isn't working. 

-----

Interestingly, one of the first things that happens in a sync login flow is [logging a "notify attached services" event](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/utils/signin.js#L306-L314), which includes a `deviceCount` field based on the session count, which seems to be correct in local testing.

Options:

* We could get rid of the `oauth_access_token_created` event, and instead modify `fxa_login - complete` to fire _after_ the device count is incremented, and base our sync counts on this event. (I think this is probably the right fix, but altering the order of events might cause some other bug to emerge.)

* Or, we could just base our device counts on `db.sessions.count` to track our sync metrics? (Not sure this will track sync logins per day/week/month without some more work, and not sure if this number will always be correct)

[...]

PR #6581, opened 2020-09-30
fix(metrics): add service name to sync oauth Amplitude events

chenba:
Because:
 - the "faked" 'cert_signed' event should have a service name of "sync"
   in its event properties when the oauth client is one of the old sync
   clients
 - the 'account.signed' event should be faked only when sync requests a
   token with the "oldsync" scope

This commit:
 - add `service: sync` to the event properties for the events in the
   /oauth/token handler when "oldsync" is in the scopes and the client
   is one of the old sync clients
 - emit the 'account.signed' event only when "oldsync" is in the scopes
 - set the service to the oauth client in the events except for the
   conditions mention in the first bullet

Closes: #6503 (and maybe #6578)

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth/index.js
+++ b/packages/fxa-auth-server/lib/routes/oauth/index.js
@@ -24,7 +24,10 @@ const {
 
 const error = require('../../error');
 const oauthRouteUtils = require('../utils/oauth');
-const { OAUTH_SCOPE_SESSION_TOKEN } = require('../../constants');
+const {
+  OAUTH_SCOPE_OLD_SYNC,
+  OAUTH_SCOPE_SESSION_TOKEN,
+} = require('../../constants');
 const ScopeSet = require('fxa-shared').oauth.scopes;
 
 const JWTIdToken = require('../../oauth/jwt_id_token');
@@ -299,6 +302,7 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
             uid,
             ecosystemAnonId,
             clientId: request.payload.client_id,
+            service: request.payload.client_id,
           });
 
           // This is a bit of a hack, but we emit the `account.signed`
@@ -306,10 +310,27 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
           // Previously, this event would get emitted on the `/certificate/sign`
           // endpoint however, with the move to sync oauth, this does not happen anymore
           // and we need to "fake" it.
-          await request.emitMetricsEvent('account.signed', {
-            uid: uid,
-            device_id: sessionToken.deviceId,
-          });
+          //
+          // A "fxa_activity - cert_signed" event will be emitted since
+          // "account.signed" is mapped to it.  And cert_signed is used in a
+          // rollup to generate the "fxa_activity - active" event in Amplitude
+          // (ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1632635), where
+          // we need the 'service' event property to distinguish between sync
+          // and browser.
+          //
+          // The service for this event gets special-cased to 'sync' so that it
+          // matches its pre-oauth `/certificate/sign` event.  ref:
+          // https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
+          if (
+            scopeSet.contains(OAUTH_SCOPE_OLD_SYNC) &&
+            config.oauth.oldSyncClientIds.includes(request.payload.client_id)
+          ) {
+            await request.emitMetricsEvent('account.signed', {
+              uid: uid,
+              device_id: sessionToken.deviceId,
+              service: 'sync',
+            });
+          }
         } catch (ex) {}
 
         return grant;
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -180,6 +180,7 @@
     },
     "clientIdToServiceNames": {
       "dcdb5ae7add825d2": "123done",
+      "5882386c6d801776": "firefox-desktop",
       "98e6508e88680e1a": "fxa-settings",
       "7377719276ad44ee": "pocket-mobile",
       "749818d3f2e7857f": "pocket-web"
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -838,6 +838,12 @@ const conf = convict({
       default: {},
       env: 'OAUTH_CLIENT_IDS',
     },
+    oldSyncClientIds: {
+      doc: 'Client IDs of sync clients that migrated to OAuth.',
+      format: Array,
+      default: ['5882386c6d801776', '1b1a3e44c54fbb58'],
+      env: 'OAUTH_OLD_SYNC_CLIENT_IDS',
+    },
     // A safety switch for disabling new signins/signups from particular clients,
     // as a hedge against unexpected client behaviour.
     disableNewConnectionsForClients: {
--- a/packages/fxa-auth-server/test/local/routes/oauth.js
+++ b/packages/fxa-auth-server/test/local/routes/oauth.js
@@ -14,10 +14,11 @@ const mocks = require('../../mocks');
 const error = require('../../../lib/error');
 const JWTIdToken = require(`${ROOT_DIR}/lib/oauth/jwt_id_token`);
 
+const { OAUTH_SCOPE_OLD_SYNC } = require(`${ROOT_DIR}/lib/constants`);
 const MOCK_CLIENT_ID = '0123456789abcdef';
 const MOCK_USER_ID = '0123456789abcdef0123456789abcdef';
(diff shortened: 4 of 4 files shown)


packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -278,39 +281,43 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+            const account = await db.account(uid);
+            const ecosystemAnonId = account.ecosystemAnonId;
 
-          await request.emitMetricsEvent('oauth.token.created', {
-            grantType: request.payload.grant_type,
-            uid,
-            ecosystemAnonId,
-            clientId: request.payload.client_id,
-          });
+            await request.emitMetricsEvent('oauth.token.created', {
```
rfk, line 301: I think it's important that we continue to emit this event even when not using the `oldsync` scope; would it work here if we reduce the `if contains SCOPE_OLD_SYNC` condition to *just* cover the emission of the `account.signed` event? (This would mean that the `oauth.token.created` event doesn't get `service: 'sync'`, but I don't think we've ever had that data for that event in the past).

vbudhram (commented):
@chenba There is some code in https://github.com/mozilla/fxa/blob/c34a98bd814ddef6ece4bf87451aa49f24bf2b5b/packages/fxa-shared/metrics/amplitude.js#L295 that should in theory look up the service name from the clientId and emit it. Maybe that is not working as intended?

@rfk do other services use the `oldsync` scope? Or is it only FF desktop?

rfk:
> do other services use the oldsync scope?

Yes, all our browsers use it.

chenba:
> that should in theory look up the service name from the clientId and emit it. Maybe that is not working as intended?

@vbudhram Thanks for the reminder on that!  The `account.signed` event did not have a `service` property in the event data.  The `oauth.token.created` event _does_ have the client id in its event data, _but_ with the key 'clientId' instead of 'service'.  (That explains the empty event properties in the BigQuery results for that event.)

I could add the client id as the 'service' property to those events, but the id is mapped to 'firefox-desktop'.  So it's still not what we want for the "cert_signed" event.

chenba:
> 
> 
> > do other services use the oldsync scope?
> 
> Yes, all our browsers use it.

Hmm, I think @davismtl is interested in desktop for #6503.  I think we need to check for oldsync _and_ desktop oauth client id.

rfk:
>  I think we need to check for oldsync and desktop oauth client id.

FWIW I don't think any harm would come from emitting these events for non-desktop clients; if you're using the `oldsync` scope here then you're definitely trying to access sync, and emitting an event with `service: "sync"` seems appropriate.

chenba:
> 
> 
> > I think we need to check for oldsync and desktop oauth client id.
> 
> FWIW I don't think any harm would come from emitting these events for non-desktop clients; if you're using the `oldsync` scope here then you're definitely trying to access sync, and emitting an event with `service: "sync"` seems appropriate.

Yep yep, that's what I have in progress.  Well, kind of.  I have the 'service' set to the client id, unless the client id is the desktop Firefox client id, then I have it set to "sync".  But let's confirm...

@davismtl (and maybe @irrationalagent) does the 'service' Amplitude event property of "sync" on "fxa_activity - active" mean only desktop sync, or all sync?

davismtl:
All **old** Sync clients can pass `sync` as the service. Historically that's what they've done, so it's preferable to maintain that for the older clients. It's only as of recently that newer clients are passing different `service` values like `firefox-ios` and `fenix` as we roll-out oauth.

> 'service' set to the client id

I'd avoid this in general. The service is supposed to be a human-readable name since we already have a property with the client id. Most of the people making charts in Amplitude are using the property `service` and don't know how to look up a client id.

image

I really hope I understood the problem and addressed the open questions. Let me know if I've misunderstood something.

chenba:
> 'service' set to the client id

Sorry I've caused confusion here.  Setting the 'service' property to the client id here mean having it mapped to a readable name (e.g. "fenix") in the actual Amplitude event.

> All old Sync clients can pass sync as the service.

Aside from Firefox desktop, what does "old" here mean?

davismtl:
>  Setting the 'service' property to the client id here mean having it mapped to a readable name (e.g. "fenix") in the actual Amplitude event.

Perfect!

> Aside from Firefox desktop, what does "old" here mean?

At the start of 2020, Firefox desktop, Firefox iOS and Firefox Android (Fennec) were all called `service=sync`. None of them were using OAuth.

- In May, Firefox for iOS migrated to OAuth. It is now `service=firefox-ios`. Consider prior to this being "old".
- With Firefox for Android migrating from Fennec to Fenix, we're observing a similar change. Fenix is using OAuth and the service name is `fenix`.
- Desktop moved to OAuth in August, also making a pretty big transition. In this case, we actually had 2 choices here when we really think about it.
1) Continue to call it Sync as we have historically done for many years.
2) Rename it to something like `sync-desktop` to distinguish the old/new as we've similarly done on iOS and Android.

I believe it's preferable to go with 1) since it would break fewer charts and because we've already been passing other events for that property value for the last month.

rfk:
This is a symptom of an ongoing ontological problem with our metrics, where we used "sync" to refer to two distinct categories of thing. Do we mean that the user has used sync the RP, or that they have used a particular type of client that happens to include syncing functionality, or that they have used a particular type of client specifically to access sync?

IMHO it would be better to emit events tagged with "service=sync" for *any* client that requests access to the sync service. But if we've all internalized the notion that "service=sync" in our metrics specifically means a syncing desktop client, setting it only for desktop makes sense.

(I think it would be even better to make these two different categories of thing show up as two different axes in our metrics, but that's a much more invasive change)

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -298,18 +301,34 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
             grantType: request.payload.grant_type,
             uid,
             ecosystemAnonId,
-            clientId: request.payload.client_id,
+            service: request.payload.client_id,
```
rfk, line 304: I think we should leave the field name as `clientId` here unless we have a specific reason to change it, if only to minimize any unexpected churn in other graphs that consume this event.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -298,18 +301,34 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
-          await request.emitMetricsEvent('account.signed', {
-            uid: uid,
-            device_id: sessionToken.deviceId,
-          });
+          if (scopeSet.contains(OAUTH_SCOPE_OLD_SYNC)) {
+            // The service for this event gets special-cased to 'sync' when the
+            // client is one of the "old" sync clients so that it matches its
+            // pre-oauth `/certificate/sign` event.
+            // ref: https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
+            const oldSyncClients = ['firefox-desktop', 'firefox-ios'];
```
rfk, line 317: nit: you may find the code here easier overall to maintain if you put the known client_ids in `oldSyncClients` with comments about what they are, rather than the three-line dynamic lookup by readable name that is used below. I don't feel strongly about it either way, it just look me a few cycles to parse out how this comparison works.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -298,18 +301,34 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+            const service =
+              config.oauth.clientIds &&
+              config.oauth.clientIds[request.payload.client_id] &&
+              oldSyncClients.includes(
+                config.oauth.clientIds[request.payload.client_id]
+              )
+                ? 'sync'
+                : request.payload.client_id;
```
rfk, line 325: We could simplify this by not emitting the event at all if the `service` would not be "sync"; we don't currently have any graphs that use this event for non-sync purposes and we haven't emitted it with any value other than `sync` for a long time. For clarity, I'm imagining something like:

```
if (scopeSet.includes(OLDSYNC) && oldSyncClients.includes(req.payload.client_id)) {
      await request.emitMetricsEvent('account.signed', {
        uid: uid,
        device_id: sessionToken.deviceId,
        service: "sync",
    });
}
```

Again, I don't feel strongly about it, but it might help future folks who find themselves in here if we keep it as minimal as possible.

rfk (approved):
This looks good to me, although I'm not 100% if the change to the field name on the `oauth.token.created` event would have any unexpected negative impact.

I left some comments on potential simplifications, but I don't feel strongly about them so please, take or leave as you like.

packages/fxa-auth-server/lib/routes/oauth/index.js
chenba: I added a configuration value for the old sync client ids and used that. 🤜 🤛

chenba:
CI is failing due to storybook only, so I'm merging.

PR #6588, opened 2020-10-01
chore(marketing): update marketing for vpn

johngruen:
## Because

We need to deprecate Send marketing and replace it with the same for the VPN

## This pull request

Does so

https://jira.mozilla.com/browse/FXA-2565

Closes: # (issue number)

## Checklist

- [ x ] My commit is GPG signed.


chenba (approved):
LGTM

PR #6591, opened 2020-10-01
Increase test timeout to make time for new migration

vbudhram:
## Because

- Our remote tests start the auth-server, which will now do some additional database connection/migration from https://github.com/mozilla/fxa/pull/6580
- Remote tests are run serially and migrations wont run if they have ran already, therefore probably don't need in other remote tests
- This targets train-190 to unblock it.

## This pull request

- Updates the test timeout to the first functional test to give it a little extra time for these connections

Closes: 

## Checklist


chenba (approved):
WFM

Issue #6579, opened 2020-09-30
Fix admin panel in dev, remove unneeded apollo dependencies


data-sync-user:
I'm already almost done with this but thought it could use a ticket:

* The default port in dev needs an update
* The auth header needs a string present for `'oidc-claim-id-token-email'` in dev
* We use various packages like `react-apollo`, `apollo-boost`, `apollo-client` etc. and should just use `@apollo/client`
* Better docs are part of what's included on the intern todo list for admin panel, but could use a look-over and getting started section

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2587)
┆Issue Number: FXA-2587

PR #6595, opened 2020-10-02
fix(admin-panel): Fix connection to admin server in dev, remove unneeded Apollo dependencies, update README

LZoog:
## Because
* The admin panel can't connect to the admin server in development
* We use '@apollo/client' in fxa-settings and should be consistent, preferring one package over a lot of separated dependencies

## This pull request
* Changes the gql server port to 8095
* Adds oidc-claim-id-token-email auth header in development
* Removes react-apollo, apollo-boost, apollo-client etc. in favor of @apollo/client
* Includes doc updates for admin-panel and admin-server
* Upgrades @apollo/client version in fxa-settings to be consistent

Closes: #6579 

## Checklist

## Other information

I realize duplicating `exampleAccountResponse` in `exampleNoResultsAccountResponse` isn't as "clean" as `Object.assign`, but I couldn't get the test passing without changing it to this. I tried a couple other things and kept getting test failures too and I didn't intend on spending this much time fixing up a test so it's fine to leave as-is IMO.

### DIFF
--- a/packages/fxa-admin-panel/package.json
+++ b/packages/fxa-admin-panel/package.json
@@ -35,8 +35,7 @@
     ]
   },
   "dependencies": {
-    "apollo-boost": "^0.4.9",
-    "apollo-client": "^2.6.4",
+    "@apollo/client": "^3.2.2",
     "body-parser": "^1.19.0",
     "convict": "^6.0.0",
     "convict-format-with-moment": "^6.0.0",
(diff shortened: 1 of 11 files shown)


packages/fxa-admin-panel/package.json
```diff
@@ -35,8 +35,7 @@
     ]
   },
   "dependencies": {
-    "apollo-boost": "^0.4.9",
-    "apollo-client": "^2.6.4",
+    "@apollo/client": "^3.2.1",
```
bbangert, line 38: Is this the same exact version used everywhere else?

bbangert (approved):
This looks good to me, we could also disable the guard in admin-server if its in development mode. But you've already done the work here, so 🤷

packages/fxa-admin-panel/package.json
LZoog: Oh looks like Settings uses `3.1.3`, I could upgrade that while I'm at it.

Issue #6372, opened 2020-09-03
[FxA-Settings] “Recovery key” link from Secondary email section does nothing


vasilicatamas:
**Build**
Stage [Train 1.186.0]

**Affected platforms**
Windows 10 x64
iOS 13.6 
Android

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/
2.  Login using an existing user.
3. Click on “Recovery key” link displayed in the “Secondary email” section.

**Expected results**
The “recovery key” link successfully opens the expected page or redirects to the intended section.

**Actual Results**
Nothing happens when clicking on the “Recovery key” link.

![issue4](https://user-images.githubusercontent.com/18114210/92129303-00e13980-ee0c-11ea-95af-8f71153c9b70.gif)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2513)
┆Issue Number: FXA-2513

jodyheavener:
Huh, once upon a time this just took you further down the page to the "Recovery Key" section. Looks like that section needs the ID "recovery-key".

vasilicatamas:
This issue is verified as fixed on Stage [1.192.0] under Windows 10 64-bit, Android 9 and iOS 13.6. Recovery key section is successfully focused when clicking on "recovery key" link from Secondary email section, but it is not entirely visible - issue tracked separately in #6805 

![issue27](https://user-images.githubusercontent.com/18114210/97717906-fa023a80-1acd-11eb-85ce-f01676234318.gif)

PR #6597, opened 2020-10-02
fix(fxa-settings): fix recovery key link

meandavejustice:
## This pull request

- fixes a broken link in secondary email section

Closes: #6372

## Checklist


chenba:
Couldn't we add an id to the Recovery Key heading and link to that instead?

jodyheavener:
> Couldn't we add an id to the Recovery Key heading and link to that instead?

Yes I would do something like that. When I left my [comment](https://github.com/mozilla/fxa/issues/6372#issuecomment-688467525) I didn't take into account that the sidebar does link to `#security`, so that ID should stay in place.

meandavejustice:
@chenba @jodyheavener Fixed it up

chenba (approved):
👍

Issue #5097, opened 2020-04-23
[services] Display icons


data-sync-user:
These should (hopefully) be pretty copy-and-pastable from the current CSS with any clean up modifications. However, it depends on FXA-58.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1758)
┆Issue Number: FXA-1758

PR #6613, opened 2020-10-06
Task(fxa-settings): connected services; icons, tests, styles

### LINKED ISSUE #5095: [services] Map through filtered data w/row component, update styles, add dummy button

meandavejustice:
Opening this to get early eyes on it, still need to add screenshots in the AM.

## This pull request

- finishes up connected services section and fills in tests for the services list filtering and sorting.

Closes: #5095 #5097 

## Checklist

## Screenshots (Optional)

connected-services-desktop
connected-services-mobile-1
connected-services-mobile-2
connected-services-mobileLandscape

I need to location the sync and vpn icons, they aren't in the protocol asset repo yet, (https://github.com/mozilla/protocol-assets/issues/49) cc/ @jescottFx

Left to complete:
  - [x] Clean up test data set and put into it's own file in DIR, (maybe pull this out into fxa-shared or mocks so it can be used througout the app tests)
  - [x] add filtering tests
    - filter out duplicates
    - use newest seen service when filtering duplicates
    - sorted correctly
    - only uniques
  - [x] add icon tests
    - test that we show the correct icon for each service and platform
  - [x] Add Screenshots
  - [x] Open up the pr and add notes about sync icon and vpn icon


packages/fxa-settings/src/models/Account.ts
```diff
@@ -7,10 +7,10 @@ import {
...
 
 export interface Location {
-  city: string;
-  country: string;
-  state: string;
-  stateCode: string;
+  city: string | undefined;
+  country: string | undefined;
+  state: string | undefined;
+  stateCode: string | undefined;
```
jodyheavener, line 13: I believe these could be:

```
  city?: string;
  country?: string;
  state?: string;
  stateCode?: string;
```

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -3,20 +3,41 @@
...
+    - only uniques
+  - consolidate icons, do they already live in fxa-react?
+  - add icon tests
+    - test that we show the correct icon for each service and platform
+  - Go through mockup again, any style tweaks
+  - Take screenshots at every width
+  - Are there more storybook states we need to include?
+  - Open up the pr and add notes about sync icon and vpn icon
+  - message channel about review
+*/
```
jodyheavener: I'm assuming this is going away at some point?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,134 @@
...
+            <div className="flex flex-start mobileLandscape:justify-center mobileLandscape:flex-1">
+              <p className="text-sm" data-testid="service-last-access">
+                {lastAccessTimeFormatted}
+              </p>
+            </div>
+          </div>
+        </div>
+
+        <div className="flex flex-grow w-full mobileLandscape:justify-end mobileLandscape:flex-1">
+          {true && (
```
jodyheavener, line 120: What's this for?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if (deviceType === 'desktop') {
+      iconSvg = <DesktopIcon data-testid="desktop-icon" />;
+    }
+  }
+
+  /*
+      - filtering tests
+      - icon tests
+    */
```
jodyheavener, line 79: Removing?

meandavejustice:
@jodyheavener II just removed that line, it was left over from testing

yarn.lock
```diff
@@ -23205,6 +23205,7 @@ fsevents@^1.2.7:
 "jquery@npm:3.5.1":
   version: 3.5.1
   resolution: "jquery@npm:3.5.1"
+  checksum: c50818c84837902a60938c85786feb612f346a9918f7285409d605449460790b3815ad8aaf5806f19e89f1a79b27d327802b421ab354fe213f3fd1801e3562e6
```
jodyheavener, line 23208: Please don't commit yarn.lock changes that aren't a result of adding or modifying dependencies in your branch

meandavejustice: I spoke to @dannycoates about this a week or so ago and I was told to check in checksum changes whenever they came up. We should make a teamwide decision for consistency

jodyheavener: If there was a mismatch somewhere and we need to get back in sync it should be done one time in a standalone branch. I've seen checksum changes get added to a couple PRs, and right now on main when I `yarn install` it generates a diff.

packages/fxa-settings/src/models/Account.ts
meandavejustice: I've tried `city?`, using `Partial<Location>` and this ^. None seem to fix my type error when running tests. Any suggestions?

```
Summary of all failing tests
 FAIL  src/components/ConnectedServices/index.test.tsx
  ● Connected Services › renders "fresh load" <ConnectedServices/> with correct content

    Failed: MissingFieldError {
      "clientOnly": false,
      "message": "Can't find field 'city' on object {}",
      "path": Array [
        "account",
        "attachedClients",
        0,
        "location",
        "city",
      ],
      "query": Object {
        "definitions": Array [
          [Object],
        ],
        "kind": "Document",
        "loc": [Object],
      },
      "variables": Object {},
    }

      10 | 
      11 | describe('Connected Services', () => {
    > 12 |   it('renders "fresh load" <ConnectedServices/> with correct content', async () => {
         |   ^
      13 |     renderWithRouter(
      14 |       <MockedCache account={{ attachedClients: MOCK_SERVICES }}>
      15 |         <ConnectedServices />

      at Env.it (../../node_modules/jest-jasmine2/build/jasmineAsyncInstall.js:91:24)
      at Suite.<anonymous> (src/components/ConnectedServices/index.test.tsx:12:3)
      at Object.<anonymous> (src/components/ConnectedServices/index.test.tsx:11:1)
```

yarn.lock
meandavejustice: I'm fine with that, but I hope we can resolve this soon. It's annoying to have to switch branches and open up a new pr so frequently on installs

jescottFx:
@meandavejustice Hi Dave I'm attaching the VPN and Sync icons here:
Sync
FPN Horizontal
FPN Vertical


I've also dropped these into the Assets page in the [FxA Settings Figma](https://www.figma.com/file/dfGrUPwV0RQ0jmzGkRKIF7/FxA-Manage-Settings-Copy-Copy?node-id=1209%3A5322) -- and screenshotted how to export a different file type just in case (png, svc, etc.)

meandavejustice:
@jescottFx Thanks for the icons, do we have a vpn icon without the text?

packages/fxa-settings/src/models/Account.ts
meandavejustice: Thanks for helping me fix this @dannycoates !

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  return (
+    <div className="my-1" id="service" data-testid="settings-connected-service">
+      <div className="p-4 border-2 border-solid border-grey-100 rounded flex mobileLandscape:justify-around items-center flex-col mobileLandscape:flex-row">
+        <div className="flex flex-grow w-full mobileLandscape:flex-2">
+          <span className="flex justify-center items-center flex-1">
+            {iconSvg}
+          </span>
+          <div className="flex flex-col flex-5 mobileLandscape:flex-row">
```
jodyheavener, line 85: Looks like we have some misalignment here:

Screen Shot 2020-10-07 at 2 59 46 PM

If you apply `items-center` here I _think_ that should address it.

jodyheavener: Oh I also just noticed how much space is to the left of the icon. Can we adjust to tighten it up per the design?

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -123,9 +83,10 @@ export const ConnectedServices = () => {
               {...{
                 key: client.clientId,
                 name: client.name,
+                deviceType: client.deviceType,
                 location: client.location,
                 lastAccessTimeFormatted: client.lastAccessTimeFormatted,
-                canSignOut: false,
+                canSignOut: true,
```
jodyheavener, line 89: Can this ever be false? If it's always true does it need to be a prop?

Also, should I be able to sign out of the session I am currently active on from here?

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
```diff
@@ -4,17 +4,128 @@
...
+  it('should show the sync icon and link', async () => {
+    const servicesList = MOCK_SERVICES.filter(
+      (item) => item.name === 'Firefox Sync'
+    );
+    renderWithRouter(
+      <MockedCache account={{ attachedClients: servicesList }}>
+        <ConnectedServices />
+      </MockedCache>
+    );
+    expect(await screen.findByTestId('sync-icon')).toBeTruthy;
```
jodyheavener, line 129: Hmmm, not a blocker, but something to noodle on: I suspect we could turn the meat of each of these tests into a single function that takes a service name, and icon test ID, and instead use it for each test case. Could help dry things up. Again, not a blocker!

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -37,51 +39,9 @@ function sortAndFilterConnectedClients(attachedClients: Array<AttachedClient>) {
...
-}) {
-  const { city, stateCode, country } = location;
-  const locationProvided =
-    city !== null || stateCode !== null || country !== null;
-
-  return (
-    <div className="my-1" id="service" data-testid="settings-connected-service">
-      <div className="border-2 border-solid border-grey-100 rounded flex justify-around items-center">
-        <span>icon</span>
+  debugger;
```
6a68, line 42: remove this

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.tsx
```diff
@@ -0,0 +1,203 @@
...
+      stateCode: null,
+    },
+    userAgent: '',
+    os: null,
+    createdTimeFormatted: '6 months ago',
+    lastAccessTimeFormatted: '6 months ago',
+    approximateLastAccessTime: null,
+    approximateLastAccessTimeFormatted: null,
+  },
+];
```
jodyheavener, line 203: Not a blocker, but it looks like a handful of these object keys are unchanging across each mock, and others are values, like timestamps and unique IDs, could be programmatically created (hooray for variability in tests!). You might consider turning these into a function that returns a mock.

Also, this doesn't need to be a TSX file 🙂

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68, line 89: Hmm, it looks like this is defaulting `canSignOut` to true for all clients. Is that true, though? Can we log out any RP?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,132 @@
...
+      iconSvg = <LockwiseIcon data-testid="lockwise-icon" />;
+    }
+    if (isSync) {
+      serviceLink =
+        'https://support.mozilla.org/en-US/kb/how-do-i-set-sync-my-computer';
+      iconSvg = <SyncIcon data-testid="sync-icon" />;
+    }
+    if (isFPN) {
+      serviceLink = 'https://vpn.mozilla.com/';
+      iconSvg = <FPNIcon data-testid="fpn-icon" />;
```
6a68, line 63: What happens if none of the `if` clauses are truthy? could that happen?

Lines 32-63 are one of those rare situations where I think it'd be cleaner to have a bunch of little FooService subclasses that implement a basic Service interface, rather than sprinkling conditional logic all over the place. That's a purely optional refactoring though.

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <div className="my-1" id="service" data-testid="settings-connected-service">
+      <div className="p-4 border-2 border-solid border-grey-100 rounded flex mobileLandscape:justify-around items-center flex-col mobileLandscape:flex-row">
+        <div className="flex flex-grow w-full mobileLandscape:flex-2">
+          <span className="flex justify-center items-center flex-1">
+            {iconSvg}
+          </span>
+          <div className="flex flex-col flex-5 mobileLandscape:flex-row">
+            <div className="flex flex-col mobileLandscape:flex-2">
+              {renderLink ? (
```
6a68, line 87: seems simpler to get rid of `renderLink` and just do `if (!!name)` here

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (deviceType === 'mobile') {
+      iconSvg = <MobileIcon data-testid="mobile-icon" />;
+    }
+    if (lowerName.includes('ipad')) {
+      iconSvg = <TabletIcon data-testid="tablet-icon" />;
+    }
+    if (deviceType === 'desktop') {
+      iconSvg = <DesktopIcon data-testid="desktop-icon" />;
+    }
+  }
```
jodyheavener, line 76: I don't know the likelihood of this happening, but each `if` statement from L44-64 and L67-75 is evaluated independently because `else`s are not used, which means more than one could evaluate to `true` and result in undesired output. As well, with such a large number of conditionals you might be better off using a switch case and matching against `true` instead of if/else here.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -105,7 +65,7 @@ export const ConnectedServices = () => {
       data-testid="settings-connected-services"
     >
       <h2 className="font-header font-bold ml-4 mb-4">Connected Services</h2>
-      <div className="bg-white tablet:rounded-xl shadow px-6 pt-7 pb-9">
+      <div className="bg-white tablet:rounded-xl shadow px-6 pt-7 pb-8">
```
jodyheavener, line 68: Looks like the padding here does not jive with the rest of the mobile padding

Screen Shot 2020-10-07 at 3 34 57 PM

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const lowerName = name.toLowerCase();
+    if (deviceType === 'mobile') {
+      iconSvg = <MobileIcon data-testid="mobile-icon" />;
+    }
+    if (lowerName.includes('ipad')) {
+      iconSvg = <TabletIcon data-testid="tablet-icon" />;
+    }
+    if (deviceType === 'desktop') {
+      iconSvg = <DesktopIcon data-testid="desktop-icon" />;
+    }
```
6a68, line 75: should these be if/else, making desktop the final else catch-all? could a Service ever not have a deviceType?

6a68 (approved):
LGTM, a stray `debugger` to remove and some questions about the code, fine to land once that's addressed

jodyheavener (changes requested):
This looks awesome. Couple cosmetic things and some code quality suggestions. Two other things:

1. What's the status of mutations? Should they be in this PR?
2. I can't demonstrate every possible Service type - can we add each of them to Storybook?

packages/fxa-settings/src/models/Account.ts
```diff
@@ -30,7 +30,7 @@ export interface AttachedClient {
   lastAccessTimeFormatted: string;
   approximateLastAccessTime: number;
   approximateLastAccessTimeFormatted: string;
-  location: Location;
+  location: Partial<DeviceLocation>;
```
chenba, line 33: Since you've already made the DeviceLocation properties optional in the interface change above, you don't need this.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: Suprised our linters didn't catch this

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: This was the ask in the jira issue, https://jira.mozilla.com/browse/FXA-1757. We have https://jira.mozilla.com/browse/FXA-1760 for the signout flow.

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: If none are true then we get the default web icon. Also in this case the "service" isn't actually a service and just a device, so we don't provide a link

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: There will always be a name, we handle service's with links differently here. In the case that it's just a regular device we only render the name (no link)

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: Final catch is the web icon

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: This is addressed in a separate issue(already linked in reply to jared's comment), the jira task that this pr is taking care of asks for a dummy button

meandavejustice:
@jodyheavener 

> What's the status of mutations? Should they be in this PR?

Mutations are to be handled in this issue https://jira.mozilla.com/browse/FXA-1949

> I can't demonstrate every possible Service type - can we add each of them to Storybook?

Yes I will add them now

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: https://github.com/mozilla/fxa/pull/6613#discussion_r501314042

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: The way these are ordered is on purpose, if mobile and ipad, we want to show the ipad logo. There should never be device type desktop with ipad as the device name. If we see that then that would expose an issue with either the way we are normalizing data in graphql or how we are collecting the device data.

meandavejustice: As for the difference between 44-67 (services) and 67-75(devices). the else statement means that 67-75 should never be hit by something that requires a service link. We never want them to be overidden by a device type icon, as the service Icon is more specific.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
meandavejustice: nice catch, fixed this up

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: Got this fixed up 👍

packages/fxa-settings/src/models/Account.ts
```diff
@@ -6,11 +6,11 @@ import {
...
-export interface Location {
-  city: string;
-  country: string;
-  state: string;
-  stateCode: string;
+export interface DeviceLocation {
+  city?: string;
+  country?: string;
+  state?: string;
+  stateCode?: string;
```
dannycoates, line 13: these should technically be like `city: string | null;` we discovered `undefined` doesn't work

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.tsx
jodyheavener: @meandavejustice this doesn't need to be a TSX file

jodyheavener:
> Yes I will add them now

These get added? I might be going crazy 🤔

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
jodyheavener: Just for code quality sake I feel like this should probably still be an if/else

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
jodyheavener: You could still use a switch case with a fall through, but this is fine.

jodyheavener: > There should never be device type desktop with ipad as the device name. If we see that then that would expose an issue with either the way we are normalizing data in graphql or how we are collecting the device data.

How are we monitoring for this? Maybe it should raise an error that sends it to Sentry?

meandavejustice:
>

    > Yes I will add them now

> These get added? I might be going crazy 🤔

@jodyheavener lol yes, I added the device that shows up with the web icon, it's listed as "David laptop". That was the only device that I wasn't seeing represented. Is there another that I am missing?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
meandavejustice: I'm not sure, I haven't seen any devices or stub data that is shaped in this way. Worst case scenario is sounds like is that it would just have the web icon. I think I'm not understanding the ask here. I'm happy to add error handling, just wanna make sure I get what case we are handling.

jodyheavener:
> I added the device that shows up with the web icon, it's listed as "David laptop". That was the only device that I wasn't seeing represented. Is there another that I am missing?

@meandavejustice 🤨 now I really feel like I'm losing it. We still talking about Storybook? I don't see any listed services [here](https://storage.googleapis.com/mozilla-storybooks-fxa/commits/e1323ae404b871a615ab6bb0ddd3674f49ba05de/fxa-settings/index.html?path=/story/components-connectedservices--default). Am I looking in the wrong spot?

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.tsx
meandavejustice: fixed it up :)

meandavejustice:
@jodyheavener 

> 🤨 now I really feel like I'm losing it. We still talking about Storybook? I don't see any listed services here. Am I looking in the wrong spot?

😭 😭 😭 😭 Okay I see what was going on, I had added the new device to MOCK_SERVICES but I NEVER ADDED MOCK_SERVICES TO STORYBOOK! I was manually linking it in the index file when doing my testing so it was automatically getting included in storybook. It's fixed now. Sorry for the confusion 😭

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
meandavejustice: Fixed this up

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
jodyheavener: Awesome! My suggestion is mostly based in the idea that we're displaying these services based on the string comparisons (does the name equal this, does the device type equal that), and while we may know all of the current values today, they could change down the road. I agree that the web icon should be the fallback if nothing matches, I was mostly suggesting we raise an exception under the hood if none of the currently known services are matched, so we could be proactive about updating instead of waiting for someone to report to us that something doesn't look right.

That being said, it's not critical to do today. What I would like to see down the road is maybe using constants here instead of strings, and one central frontend spot where we could review all possible services: https://jira.mozilla.com/browse/FXA-2607

jodyheavener (commented):
I left you one question about mapping the services out per a comment Lauren left; I'm around to r+ once that's looked at! Thank you for making those changes 🙏

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const renderLink = isPocket || isMonitor || isLockwise || isSync || isFPN;
+  let serviceLink = '';
+  let iconSvg = <WebIcon data-testid="web-icon" />;
+
+  if (renderLink) {
+    if (isPocket) {
+      serviceLink = 'https://www.mozilla.org/en-US/firefox/pocket/';
+      iconSvg = <PocketIcon data-testid="pocket-icon" />;
+    } else if (isMonitor) {
+      serviceLink = 'https://monitor.firefox.com/';
```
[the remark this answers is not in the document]
meandavejustice: I didn't opt not to do it, but I found an issue that covers making a map for the icons. So I'm planning to handle this next sprint, https://jira.mozilla.com/browse/FXA-58

jodyheavener (approved):
Lookin' good!

PR #6619, opened 2020-10-06
chore(logs): lower the log level of some lines

dannycoates:
Because we don't need this level of verbosity for them in production and which are costing us extra storage space and query time.

This lowers the log level of some lines identified by this query:

```
SELECT
  jsonPayload.type,
  COUNT(*) AS count,
  SUM(COUNT(*)) OVER () AS total,
  ROUND(COUNT(*) / SUM(COUNT(*)) OVER (),4) AS percentage
FROM
  `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_20200928`
GROUP BY
  1
ORDER BY
  4 DESC;
```

### DIFF
--- a/packages/fxa-auth-server/lib/cad-reminders.js
+++ b/packages/fxa-auth-server/lib/cad-reminders.js
@@ -100,7 +100,7 @@ class CadReminders {
           return result;
         }, {})
       );
-      this.log.info('cadReminders.delete', {
+      this.log.debug('cadReminders.delete', {
         uid,
       });
       return result;
--- a/packages/fxa-auth-server/lib/db.js
+++ b/packages/fxa-auth-server/lib/db.js
@@ -305,7 +305,7 @@ module.exports = (config, log, Token, UnblockCode = null) => {
       }
       return mergedToken;
     });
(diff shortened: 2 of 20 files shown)


dannycoates:
With the test changes ... I decided to remove asserts for log lines in affected files. We should only be asserting for important side-effects like metric events that downstream things depend on. Asserting for incidental logs makes it difficult to know if they can be safely changed/removed or not.

ckolos:
Dumb question perhaps, but in the logs being switched to `debug` should I be assuming that we'll get no output unless `debug` is explicitly set?

dannycoates:
> Dumb question perhaps, but in the logs being switched to `debug` should I be assuming that we'll get no output unless `debug` is explicitly set?

Yeah, it's based on the `LOG_LEVEL` env which defaults to `info` so `debug` won't normally be logged. Levels are https://github.com/seanmonstar/intel#setting-the-log-level

vbudhram (approved):
@dannycoates SGTM 👍

ckolos (approved):

Issue #4858, opened 2020-04-10
[E2E] Change password


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1568)
┆Issue Number: FXA-1568

PR #6627, opened 2020-10-06
fix(tests): Add change password functional test for beta settings

vbudhram:
## Because

- We don't have functional tests in the new beta settings page

## This pull request

- Adds functional tests for change password in beta settings
- Adds a helper method `formatDataTestId` that prepends a value to the `data-testid` field so that it is unique

Closes: #4858 

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_v2/change_password.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/change_password.js
@@ -1,0 +1,113 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const intern = require('intern').default;
+const { describe, it, beforeEach } = intern.getPlugin('interface.bdd');
+const selectors = require('../lib/selectors');
+const FunctionalHelpers = require('../lib/helpers');
+
+const config = intern._config;
+const EMAIL_FIRST = config.fxaContentRoot;
+const SETTINGS_V2_URL = `${config.fxaContentRoot}beta/settings`;
+const password = 'passwordzxcv';
+const newPassword = 'passwordzxcvb';
+
+const { createEmail } = FunctionalHelpers;
+
+const {
+  clearBrowserState,
+  click,
+  createUser,
+  openPage,
+  fillOutEmailFirstSignIn,
+  testElementExists,
+  type,
+} = FunctionalHelpers.helpersRemoteWrapped;
+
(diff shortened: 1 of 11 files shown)


packages/fxa-content-server/tests/functional/settings_v2/change_password.js
```diff
@@ -0,0 +1,113 @@
...
+    await testElementExists(selectors.SETTINGS.HEADER, remote);
+    await openPage(SETTINGS_V2_URL, selectors.SETTINGS_V2.HEADER, remote);
+    await click(
+      selectors.SETTINGS_V2.CHANGE_PASSWORD.OPEN_BUTTON,
+      selectors.SETTINGS_V2.CHANGE_PASSWORD.CURRENT_PASSWORD_LABEL,
+      remote
+    );
+
+    await click(
+      selectors.SETTINGS_V2.CHANGE_PASSWORD.CURRENT_PASSWORD_LABEL,
```
vbudhram, line 50: It looks like the input area is blocked by a span so we would need to click that first to gain focus.

packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -86,24 +87,29 @@ export const UnitRow = ({
...
+  function formatDataTestId(id: string) {
+    return prefixDataTestId ? `${prefixDataTestId}-${id}` : id;
+  }
+
   return (
     <div className="unit-row">
       <div className="unit-row-header">
         <span className="flex justify-between items-center">
-          <h3 data-testid="unit-row-header">{header}</h3>
+          <h3 data-testid={formatDataTestId('unit-row-header')}>{header}</h3>
```
vbudhram, line 105: @jodyheavener You mentioned wanting to have intern use the `data-testid` to identify elements for testing. This mostly works, however in the `unit-row` and `input` components, if you have multiple of that component, it will complain that selector is not specific enough.

Using `prefixDataTestId` to prefix the component seems to solve this issue. Is there a better react way of doing this?

jodyheavener: Apologies, I actually don't recall my suggestion of that 🤔 But it was probably in reference to the second point of [this comment](https://github.com/testing-library/react-testing-library/issues/479#issuecomment-529548194). Mostly a convenience thing; if we have the test-ids there it would be handy to utilize them. We could certainly prefix them to de-dupe, I don't know of a more standardized method, though.

vbudhram: @jodyheavener Sounds good, I'll stick with de-duping them.

dannycoates (approved):

Issue #6162, opened 2020-08-07
Update StripeHelper methods to use DB association between FxA user and Stripe customer


data-sync-user:
Once we have associations between FxA users and Stripe customers populated in the DB (FXA-2401) and StripeHelper methods manage the association going forward (FXA-2402), we should start using that association to decide whether a given user is a customer - and thus whether or not we should hit the Stripe API for further information.

Main point of interest is probably this method:

* customer: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/payments/stripe.ts#L454

Here, we support a `cacheOnly` parameter that other parts of the system use to indicate that Stripe APIs should not be called and only cached data should be used.

The current assumption here is that every user's Stripe customer data will always be present in Redis - otherwise the user is treated as a non-customer. (Thus, the reason why customer data expiring from the cache in FXA-2369 was a problem.)

We should drop support for this. Just use the DB association and Redis expiration TTL to decide whether to fetch and cache data. Then, users of StripeHelper.customer() will get reasonably fresh data for customers, and undefined for users without the DB association.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2403)
┆Issue Number: FXA-2403

PR #6631, opened 2020-10-06
feat(payments): Use AccountCustomer info for Fetching Customer info

jackiemunroe:
## Because

- We want to use our new stored account-customer relationship for determining Stripe calls for customer data

## This pull request

- Implements read-thru and write-thru caching based off of defined relationship in AccountCustomers table

Closes: #6162

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -388,25 +388,34 @@ export class StripeHelper {
   }
 
   /**
-   * Fetch a customer for the record from Stripe based on email.
+   * Fetch a customer for the record from Stripe based on user id.
    */
   async fetchCustomer(
     uid: string,
-    email: string,
     expand?: string[]
   ): Promise<Stripe.Customer | void> {
-    const customerResponse = await this.stripe.customers
-      .list({ email, expand })
-      .autoPagingToArray({ limit: 20 });
-    if (customerResponse.length === 0) {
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -388,25 +388,34 @@ export class StripeHelper {
...
     expand?: string[]
   ): Promise<Stripe.Customer | void> {
-    const customerResponse = await this.stripe.customers
-      .list({ email, expand })
-      .autoPagingToArray({ limit: 20 });
-    if (customerResponse.length === 0) {
+    const accountCustomer = await getAccountCustomerByUid(uid);
+    if (
+      accountCustomer === undefined ||
+      accountCustomer.stripeCustomerId === undefined
```
chenba, line 400: Is this possible?  How does it get to be undefined?

chenba (approved):
LGTM but got a question

packages/fxa-auth-server/lib/payments/stripe.ts
jackiemunroe: @chenba So, the customer account table has stripeCustomerId as a nullable field. This is to "future-proof" the table in the event we want to add additional columns linking a Firefox user to an external id for a different payment provider.

jackiemunroe: So, yes, there should not currently be any cases where the stripeCustomerId is undefined currently, this is just to reflect the possibility that it _might be_ in the future

Issue #6370, opened 2020-09-03
[FxA-Settings] “Make primary” button incorrectly displayed on mobile


vasilicatamas:
**Build**
Stage [Train 1.186.0]

**Affected platforms**
iOS 13.6 
Android

**Affected devices**
iPhone 11
Moto G6

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ 
2. Login using an user that has already added multiple secondary emails (at least 2)
3. Check the last “Make primary” button.

**Expected results**
The last “Make primary” button is displayed above the following text: “Access your account if you can't log in to your primary email.
Note: a secondary email won't restore your information—you'll need a recovery key for that.”

**Actual Results**
“Make primary” button that corresponds to the last secondary email is displayed under the secondary email text.

![Screenshot_35](https://user-images.githubusercontent.com/18114210/92123075-e22b7480-ee04-11ea-9301-6855347282aa.png)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2511)
┆Issue Number: FXA-2511

vasilicatamas:
This issue is verified as fixed on Stage [1.192.0] under Android 9 and iOS 13.6. The email text is successfully displayed under the last "Make primary" button.

![Screenshot_24](https://user-images.githubusercontent.com/18114210/97600947-47b96d00-1a12-11eb-9d1f-d801d5912573.png)

PR #6639, opened 2020-10-07
bug(fxa-settings): fix secondary email msg placement

meandavejustice:
## This pull request

- fixes the placement of secondary email msg in UnitRow

Closes: #6370

## Checklist

## Screenshots (Optional)

mobile-make-primary-button-fix


jodyheavener (commented):
This looks good, but I'm curious about the ordering. @LZoog do you think the "Access you account..." messaging needs to appear after unverified emails as well? I'm tempted to say no (it's also an edge case because multiple emails in the new UI will not be likely).

If the answer is no, we just need to add a little space between the "Note:..." and the following email section.

Screen Shot 2020-10-07 at 4 36 56 PM

jodyheavener (approved):
🤠

PR #6642, opened 2020-10-07
bug(payments): bad stripe expand fields

jackiemunroe:
The keys for the fields to expand on a customer call are different from a customer list call. The 'data' prefix for the key needs to be dropped.

## Checklist


dannycoates (approved):
Thanks @jackiemunroe!

Issue #4954, opened 2020-04-15
[display name] Cancel link + enable/disable save button


data-sync-user:
Allow "Cancel" to take the user back to the previous page and disable/enable "Save" depending on display name input validity. It should act as the current "display name" form does now - it should be enabled if the display name in the input value differs than the user's current display name. Double check other client-side validation from the current Settings `display_name` form is transferred over.

Handling submission will be handled in another issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1644)
┆Issue Number: FXA-1644

PR #6649, opened 2020-10-08
feat(settings): update save btn state on display name input

chenba:
Because:
 - the disabled state of the save button should be based on the input
   value of the display name

This commit:
 - update the disabled attr of the save button when the display name
   input value changes

Closes: #4954 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
+++ b/packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
@@ -2,11 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import 'mutationobserver-shim';
 import React from 'react';
 import '@testing-library/jest-dom/extend-expect';
-import { screen } from '@testing-library/react';
+import { act, fireEvent, screen } from '@testing-library/react';
 import PageDisplayName from '.';
-import { MockedCache, renderWithRouter } from 'fxa-settings/src/models/_mocks';
+import {
+  MockedCache,
+  MOCK_ACCOUNT,
+  renderWithRouter,
+} from 'fxa-settings/src/models/_mocks';
 import { AuthContext, createAuthClient } from 'fxa-settings/src/lib/auth';
 
 const client = createAuthClient('none');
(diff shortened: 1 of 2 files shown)


packages/fxa-settings/src/components/PageDisplayName/index.tsx
```diff
@@ -3,37 +3,60 @@
...
+            label="Enter display name"
+            className="mb-2"
+            data-testid="display-name-input"
+            inputRef={displayNameRef}
+            onChange={checkForm}
+          />
+        </div>
+        <div className="flex justify-center mb-4 mx-auto max-w-64">
+          <button
+            className="cta-neutral mx-2 flex-1"
```
dannycoates, line 46: prob want `type="button"` here

packages/fxa-settings/src/components/PageDisplayName/index.tsx
```diff
@@ -3,37 +3,60 @@
...
+            onClick={() => window.history.back()}
+          >
+            Cancel
+          </button>
+          <button
+            data-testid="submit-display-name"
```
dannycoates, line 52: may as well set `type="submit"` just to be clear

packages/fxa-settings/src/components/PageDisplayName/index.tsx
```diff
@@ -3,37 +3,60 @@
...
+  const user = useAccount();
+  const isValidDisplayName = validateDisplayName(user.displayName || '');
+  const [saveBtnDisabled, setSaveBtnDisabled] = useState(true);
+
+  const checkForm = useCallback(
+    (ev: ChangeEvent<HTMLInputElement>) => {
+      setSaveBtnDisabled(!isValidDisplayName(ev.target.value));
+    },
+    [setSaveBtnDisabled, isValidDisplayName]
+  );
```
dannycoates, line 26: This is fine. Just fyi I added [react-hook-form](https://react-hook-form.com/) while doing the more complicated validation for change password. It might make sense to use it throughout settings, but it doesn't need to be now.

packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
```diff
@@ -24,3 +28,40 @@ it('renders', async () => {
...
+      </MockedCache>
+    </AuthContext.Provider>
+  );
+
+  // initial value
+  expect(screen.getByTestId('submit-display-name')).toBeDisabled();
+
+  // empty value
+  await act(async () => {
+    fireEvent.change(screen.getAllByTestId('input-field')[0], {
```
dannycoates, line 46: I needed to change `fireEvent.change` to `fireEvent.input` on the change password form. _If_ that works here I think we should prefer `input` for form fields in general.

dannycoates (approved):
LGTM. button thing is the only one that's not optional imo

packages/fxa-settings/src/components/PageDisplayName/index.tsx
chenba: I can't seem to find where react-hook-form is currently used in the repo.  😕

dannycoates: https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/PageChangePassword/index.tsx#L55

chenba:
Apparently I started my branch on a pretty old rev of `main`?!  Anyways, made all the suggested changes.  Thanks, Danny!

PR #6658, opened 2020-10-09
fix(log): Update metrics error log to trace

### LINKED ISSUE #3580: Excessive “missingFlowId” errors in fxa-auth-server log
`route./.well-known/browserid.200` exists since forever.

Now that we've deployed our rebase on top of train-150, those from fxa-oauth-server like `route./verify.200` & `route./authorization.200` are taking even more spaces.

I understand there're multiple ways to suppress such an error in [lib/metrics/events.js](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/lib/metrics/events.js), but I'm not sure which one fits here the best.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-51)
┆Issue Number: FXA-51

### LINKED ISSUE #5306: [auth-server] lower logging of metricsEvents.emitFlowEvent "missingFlowId":true

vbudhram:
Noticed this while creating signin/signup papercut milestone, we really don't need this to be `error` since it really isnt.

Fixes #5306
Fixes #3580

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/events.js
+++ b/packages/fxa-auth-server/lib/metrics/events.js
@@ -189,7 +189,7 @@ module.exports = (log, config) => {
 
   function emitFlowEvent(event, request, optionalData) {
     if (!request || !request.headers) {
-      log.error('metricsEvents.emitFlowEvent', { event, badRequest: true });
+      log.trace('metricsEvents.emitFlowEvent', { event, badRequest: true });
       return P.resolve();
     }
 
--- a/packages/fxa-auth-server/test/local/metrics/events.js
+++ b/packages/fxa-auth-server/test/local/metrics/events.js
@@ -12,6 +12,7 @@ const log = {
   error: sinon.spy(),
   flowEvent: sinon.spy(),
   info: sinon.spy(),
+  trace: sinon.spy(),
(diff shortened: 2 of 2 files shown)


chenba (approved):
📜

PR #6659, opened 2020-10-09
chore(test): Remove custom firefox build for pairing tests

vbudhram:
Another small fix I found while making papercut milestone.


dannycoates (approved):
nice! also here: https://github.com/mozilla/fxa/blob/main/_dev/docker/circleci/Dockerfile#L12

PR #6664, opened 2020-10-09
Create codeql-analysis.yml

clouserw:
## Because

- @jodyheavener suggested we should check out the new auto-scanning feature

## This pull request

- enables the new auto-scanning feature

## Checklist


clouserw:
We tried this on a private fork and it found a bunch of false positives, but also a couple things we could adjust, and having alerts in the future could be cool.  Let's see if it works on the _mozilla_ org.  If it uses actions, I guess it just won't work?

chenba (approved):
My internal yaml parser says ok

Issue #4956, opened 2020-04-15
[display name] Process form response (GQL mutation)


data-sync-user:
You can reference how a mutation query functions in the `fxa-admin-panel`.

On form submission, capture the user's input response and send the GQL mutation request to the GQL endpoint.

A successful response should 1) take the user back to the main settings page 2) with a success message triggering the global alert message while 3) updating the display name on the main Settings page.

Any server errors should be displayed on the `/display_name` separate page in the global alert message.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1645)
┆Issue Number: FXA-1645

PR #6669, opened 2020-10-12
feat(settings): update display name

chenba:
Because:
 - new settings should save the account's new display name

This commit:
 - save the new display name with gql-api
 - allow empty display name (delete)

Closes: #4956 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
+++ b/packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
@@ -6,15 +6,62 @@ import 'mutationobserver-shim';
 import React from 'react';
 import '@testing-library/jest-dom/extend-expect';
 import { act, fireEvent, screen } from '@testing-library/react';
-import PageDisplayName from '.';
+import PageDisplayName, { UPDATE_DISPLAY_NAME_MUTATION } from '.';
 import {
+  createCache,
   MockedCache,
   MOCK_ACCOUNT,
   renderWithRouter,
-} from 'fxa-settings/src/models/_mocks';
-import { AuthContext, createAuthClient } from 'fxa-settings/src/lib/auth';
+} from '../../models/_mocks';
+import { AuthContext, createAuthClient } from '../../lib/auth';
+import { HomePath } from '../../constants';
+import { GraphQLError } from 'graphql';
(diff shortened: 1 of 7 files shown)


packages/fxa-settings/src/constants/index.tsx
```diff
@@ -0,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export const HomePath = '/beta/settings';
```
jodyheavener, line 5: yay

packages/fxa-settings/src/components/PageDisplayName/index.test.tsx
```diff
@@ -6,15 +6,64 @@ import 'mutationobserver-shim';
...
   renderWithRouter,
 } from 'fxa-settings/src/models/_mocks';
 import { AuthContext, createAuthClient } from 'fxa-settings/src/lib/auth';
+import { HomePath } from 'fxa-settings/src/constants';
+import { GraphQLError } from 'graphql';
+import { MockedProvider } from '@apollo/client/testing';
+import { Account, GET_ACCOUNT } from 'fxa-settings/src/models';
+import AlertBarContext, {
+  AlertBarRootAndContextProvider,
+} from 'fxa-settings/src/lib/AlertBarContext';
```
jodyheavener, line 23: Although I can see at least one example to the contrary, I think we're trying to be consistent in using relative paths for imports from within fxa-settings. Could we change to that here?

dannycoates: yeah, it vscode likes to auto-import this way. i can't quite remember why we decided to prefer relative... something about the ts build step i think

chenba: Yeah, a lot of my imports are just results of ctrl+.  I'll change it.

dannycoates (approved):
super! 🎉

chenba:
✔️ Updated a bunch of `import`s to use relative paths.

Issue #5098, opened 2020-04-23
[services] Create "FF for mobile and tablet" row, display conditionally


data-sync-user:
If a user does not have a mobile or tablet device attached to their account, show this row that links out to mobile apps.

The "send a download link to your device" link should go to `accounts.firefox.com/connect_another_device`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1759)
┆Issue Number: FXA-1759

data-sync-user:
➤ Dave Justice commented:

https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A4

PR #6670, opened 2020-10-12
feat(fxa-settings): add ConnectAnotherDevicePromo

meandavejustice:
## This pull request

- adds the `<ConnectAnotherDevicePromo />` component

Closes: #5098

## Checklist

## Screenshots (Optional)

mobile-promo-on-desktop
mobile-promo-on-mobile

> 


packages/fxa-settings/src/components/PageSettings/index.stories.tsx
```diff
@@ -8,6 +8,13 @@ import { PageSettings } from '.';
 import { LocationProvider } from '@reach/router';
 import AppLayout from '../AppLayout';
 import { MockedCache, mockEmail } from 'fxa-settings/src/models/_mocks';
+import { MOCK_SERVICES } from '../ConnectedServices/MOCK_SERVICES';
+
+const SERVICES_NON_MOBILE = MOCK_SERVICES.filter(el => {
+  if (el.deviceType === 'mobile') return false;
+  if (el.name.toLowerCase().includes('ipad')) return false;
+  return true;
```
chenba, line 16: I vaguely remember seeing this logic somewhere else.  It would be nice to capture it in one place.

chenba: haha I'm scrolling _up_... sorry I didn't mean somewhere else _in this PR_.  I meant before this PR.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -58,6 +59,12 @@ export const ConnectedServices = () => {
     alertBar.show();
   });
 
+  const showMobilePromo = !!~sortedAndUniqueClients.filter(el => {
+    if (el.deviceType === 'mobile') return true;
+    if (el.name.toLowerCase().includes('ipad')) return true;
+    return false;
+  }).length;
```
chenba, line 66: I don't think that `~` is necessary.

chenba (changes requested):
Looks fine in ltr.  But doesn't look right in rtl though.

meandavejustice:
@chenba switched `!!~` to `!` and fixed rtl.

rtl-handling

jodyheavener:
This is not aligned correctly Group

jodyheavener:
> But doesn't look right in rtl though.

I'm all for getting rtl nailed down out of the gate, but it sounds like we will [also do a pass later](https://mozilla.slack.com/archives/CLV3KMZ8B/p1602252762309800?thread_ts=1602247625.308600&cid=CLV3KMZ8B) to make sure it's all good.

packages/fxa-settings/src/components/PageSettings/index.stories.tsx
chenba: > 
> 
> I don't know of any other place we filter like this, there is a place where we make [similar checks to show 
the icons](https://github.com/mozilla/fxa/pull/6677),

Hmm maybe it was in another package where I saw that logic.  So for this, looks like we need something that tells us if a device is a "mobile" device or not.  i.e.

```JavaScript
const isMobileDevice = (d) => d.deviceType === 'mobile' || d.name.toLowerCase().includes('ipad');
```

For the first filter we can just use `.filter(isMobileDevice)`.  For the second one, we do need to do `.filter(x => !isMobileDevice(x)`.

In addition having the "mobile device" logic in one place, I think it also reads better when we give that logic a name.

> but that would be weird to pull out into this story IMO

I don't understand what's weird about it.  I saw it in this PR so I suggested here.  If there's a more appropriate story, then let me know and I'll leave a note there.

jodyheavener:
I don't see a specific screen for it at mobile, but I'd probably center the "Missing or duplicate items?" text on mobile now that items above and below it are both centered 
Screen Shot 2020-10-13 at 4 39 11 PM

meandavejustice:
@chenba Added `isMobileDevice` util. much cleaner thanks for the suggestion.

@jodyheavener Fixed centering of missing items copy on mobile, and centered buttons, the playstore svg has a built in transparent padding so we just needed to match that for the app store icon.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/app-store.svg
```diff
@@ -0,0 +1,126 @@
+<svg version="1.1" id="US_UK_Download_on_the" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
```
chenba, line 1: LOL @ cliffhanger of an `id`

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
```diff
@@ -8,6 +8,12 @@ import ConnectedServices, { sortAndFilterConnectedClients } from '.';
 import { renderWithRouter, MockedCache } from '../../models/_mocks';
 import { MOCK_SERVICES } from './MOCK_SERVICES';
 
+const SERVICES_NON_MOBILE = MOCK_SERVICES.filter(el => {
+  if (el.deviceType === 'mobile') return false;
+  if (el.name.toLowerCase().includes('ipad')) return false;
+  return true;
```
chenba, line 14: Could use `isMobileDevice` here.

chenba:
I was expecting `rtl` to look like 
image

Jody said that we can fix rtl stuff later though.

chenba (approved):
📱

PR #6671, opened 2020-10-13
feat(settings): implemented add recovery key

dannycoates:


packages/fxa-auth-client/lib/recoveryKey.ts
```diff
@@ -0,0 +1,42 @@
+import { jweEncrypt, hkdf } from './crypto';
+import { hexToUint8, uint8ToHex } from './utils';
+
+export async function generateRecoveryKey(
+  uid: string,
+  keys: { kA: string; kB: string }
```
chenba, line 6: Same type as https://github.com/mozilla/fxa/pull/6671/files#diff-4b170feaa918643ed9b30065a372669dc87993d5e83113cf851077783c46f6beR454-R457 ?  If so, maybe define a type for it.

chenba:
Looks good so far! 👍

packages/fxa-auth-client/lib/recoveryKey.ts
dannycoates: yeah im half-way through changing the types. I kind of wonder if labeling `hexstring` is worth it or not since it's just an alias. 🤔 ... thoughts?

chenba: I like it because then the readers know that they are hex strings.

packages/fxa-settings/scripts/test-ci.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash -ex
 
+CI=no yarn build
```
dannycoates, line 3: after https://github.com/mozilla/fxa/issues/6684 remove the `CI=no`

chenba:
@dannycoates I'm getting the following:

```
/home/chenba/code/fxa/packages/fxa-settings/src/lib/auth.ts
TypeScript error in /home/chenba/code/fxa/packages/fxa-settings/src/lib/auth.ts(5,3):
Module '"../../../fxa-auth-client/browser"' has no exported member 'generateRecoveryKey'. Did you mean to use 'import generateRecoveryKey from "../../../fxa-auth-client/browser"' instead?  TS2614
```

chenba (approved):
💯  And thanks for all the `hexstring`s.

Issue #3566, opened 2019-12-03
[services] Style client icons with `data-client-id` instead of `data-name`s?


LZoog:
Edit: See first comment.

---
I was able to close #1782 with a CSS fix, but didn't want the following suggestion to be lost.

- - -
_Originally posted by @rfk in https://github.com/mozilla/fxa/issues/1782#issuecomment-511614223_

So I had a quick poke around at [how the icons are determined](https://github.com/mozilla/fxa/blob/91c94c7464190cab341dfed27f00b93018571788/packages/fxa-content-server/app/styles/modules/_settings.scss#L503), and I want to propose a bit of a refactor to how that's done to work better with oauth clients that are also devices.

Currently we determine the icon by client name, using rules like:

```
.client-oAuthApp[data-name='Android Components Reference Browser'] {
    background-image: url('/images/icon-device-phone-android.svg');
}
```

This seems fragile, because we might change the name on the backend and forget to update the rules on the frontend. It also won't play well with oauth device records, which can customize their display name.

What if instead, we selected the icon by client id, and applied it to all attached clients rather specifically to those of type `oAuthApp`? We could set a new data property `data-client-id` on the HTML element, and use a rule like this:

```
.client[data-client-id='3c49430b43dfba77'] {
    background-image: url('/images/icon-device-phone-android.svg');
}
```

This would require us to be a little more explicit about the list of clients (we "cheat" in a few places by pattern-matching on names like "Firefox Send") but TBH I feel like explicit is probably good here.

@shane-tomlinson what do you think, is it worth exploring something like this?

- - -
_Originally posted by @shane~~tomlinson in https://github.com/mozilla/fxa/issues/1782#issuecomment~~511757385_

I've always wanted to use the client_id. The only downside is that client ids can change from dev servers to stable to stage/prod, so we might have a proliferation of client ids in the CSS. That, or we keep the same client id for all environments.

- - -
Personally, I prefer a selector like `&[data-name^='Firefox Send']` because it's more human readable and we don't have to either make IDs consistent across environments or have a bunch of extra CSS, and I don't foresee us tampering with product names in the back-end, but I may not understand the full implications of what this means: 

> It also won't play well with oauth device records, which can customize their display name.

Also, in response to this:

> What if instead, we selected the icon by client id, and applied it to all attached clients rather specifically to those of type `oAuthApp`?

[...]

PR #6677, opened 2020-10-13
task(fxa-setttings): refactor Service component

meandavejustice:
## This pull request

- dynamically loads svg icons
- refactors the icon choosing code into a switch statement
- adds the Icon component for dynamically loading svgs

Closes: #3566

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/ConnectedServices/Icon.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/Icon.tsx
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// https://codesandbox.io/s/react-dynamic-svg-import-ts-kzx7n?fontsize=14&hidenavigation=1&theme=dark
+import React, { useEffect } from 'react';
+
+interface IconProps extends React.SVGProps<SVGSVGElement> {
+  name: string;
+}
+
+export const Icon: React.FC<IconProps> = ({ name, ...rest }): JSX.Element | null => {
+  const ImportedIconRef = React.useRef<
+    React.FC<React.SVGProps<SVGSVGElement>>
+  >();
+  const [loading, setLoading] = React.useState(false);
+
+  useEffect((): void => {
+    setLoading(true);
+    const importIcon = async (): Promise<void> => {
+      try {
+        ImportedIconRef.current = (
+          await import(`./${name}.svg`)
+        ).ReactComponent;
+      } catch (err) {
+        throw err;
+      } finally {
+        setLoading(false);
+      }
+    };
+    importIcon();
+  }, [name]);
+
+  if (!loading && ImportedIconRef.current) {
+    const { current: ImportedIcon } = ImportedIconRef;
+    return <ImportedIcon {...rest} data-testid={`${name}-icon`} />;
+  }
+
+  return null;
+};
+
+export default Icon;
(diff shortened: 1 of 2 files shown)


packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -30,56 +22,51 @@ export function Service({
...
+      break;
+    default:
+      iconName = "web";
+      if (deviceType === "mobile") {
+        iconName = "mobile";
+      }
+      if (name.toLowerCase().includes("ipad")) {
+        iconName = "tablet";
+      } else if (deviceType === "desktop") {
+        iconName = "desktop";
```
jodyheavener, line 56: Did this file get reformatted? Looks like this and the above are using double quotes now

chenba: Yeah looks like everything in this commit is using double quotes.

meandavejustice: Fixed that, should be much more legible diff now

packages/fxa-settings/src/components/ConnectedServices/Icon.tsx
```diff
@@ -0,0 +1,42 @@
...
+// https://codesandbox.io/s/react-dynamic-svg-import-ts-kzx7n?fontsize=14&hidenavigation=1&theme=dark
+import React, { useEffect, FC } from 'react';
+
+interface IconProps extends React.SVGProps<SVGSVGElement> {
+  name: string;
+}
+
+export const Icon: FC<IconProps> = ({ name, ...rest }): JSX.Element | null => {
+  const ImportedIconRef = React.useRef<
+    React.FC<React.SVGProps<SVGSVGElement>>
```
chenba, line 14: nit: I suppose this could just be `FC` since it's imported and used on line 7.  Although I prefer the qualified name since `FC` is not a great name

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -30,56 +22,51 @@ export function Service({
...
+    default:
+      iconName = 'web';
+      if (deviceType === 'mobile') {
+        iconName = 'mobile';
+      }
+      if (name.toLowerCase().includes('ipad')) {
+        iconName = 'tablet';
+      } else if (deviceType === 'desktop') {
+        iconName = 'desktop';
+      }
```
chenba, line 57: If I understand this correctly... we could switch positions of the 'ipad' and 'mobile' conditions, then use `else if` consistently.  I.e.

```JavaScript
      iconName = 'web';
      if (name.toLowerCase().includes('ipad')) {
        iconName = 'tablet';
      } else if (deviceType === 'mobile') {
        iconName = 'mobile';
      } else if (deviceType === 'desktop') {
        iconName = 'desktop';
      }
```

packages/fxa-settings/src/components/ConnectedServices/Icon.tsx
meandavejustice: Switched so both instances use `React.FC`

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -30,56 +22,51 @@ export function Service({
...
-      iconSvg = <TabletIcon data-testid="tablet-icon" />;
-    } else if (deviceType === 'desktop') {
-      iconSvg = <DesktopIcon data-testid="desktop-icon" />;
-    }
+      iconName = 'fpn';
+      break;
+    case 'Firefox Sync':
+      serviceLink = 'https://support.mozilla.org/en-US/kb/how-do-i-set-sync-my-computer';
+      iconName = 'sync';
+      break;
```
chenba, line 47: We could use a dictionary here instead.  The names for the keys and the values would be of a type `{serviceLink: someUrlTypeThatIsAnAliasOfString, iconName: string}`.  Then we'd map into it for the link and icon.  Just a code style thing; the switch here works.

chenba (approved):
LGTM

Issue #4968, opened 2020-04-16
[2FA] Create `/two_step_authentication` page with dummy step 1 of 3 form


data-sync-user:
The `/two_factor_authentication` route should be hooked up from the settings page in FXA-1614.

Add the text, links, and the form for the step 1 of 3 page and style it accordingly. The QR code will be added in a follow up issue.

"Cancel" should bring the user back to the main Settings page and "Continue" should be enabled with any character in the input.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1659)
┆Issue Number: FXA-1659

PR #6682, opened 2020-10-14
feat(settings): add two step authentication page

chenba:
Because:
 - we need to scaffold the two step auth page

This commit:
 - add the two step auth page

Closes: #4968 

## Checklist


packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -65,25 +65,26 @@ export const UnitRowTwoStepAuth = () => {
...
-      }
-    : {
-        headerValue: null,
-        noHeaderValueText: 'Not Set',
-        ctaText: 'Add',
-        secondaryCtaText: undefined,
-        revealSecondaryModal: undefined,
-      };
+  const conditionalUnitRowProps =
+    exists && verified
```
chenba, line 69: The feature is not enabled until it is verified.

dannycoates (approved):
WFM! 👍

PR #6685, opened 2020-10-14
Settings v2 basic navigation test

ashrivastava-qa:
## Because

- `added basic navigation tests for settings v2`

## This pull request

- `This is for task FXA-2596`

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

@chenba @dannycoates @LZoog please review. Thanks!

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_v2/navigation.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/navigation.js
@@ -68,6 +68,47 @@ describe('navigation', () => {
     );
   });
 
+  it('can click secondary email menu add button and back', async ({
+    remote,
+  }) => {
+    await click(
+      selectors.SETTINGS_V2.SECONDARY_EMAIL.ADD_BUTTON,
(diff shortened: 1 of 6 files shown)


packages/fxa-content-server/tests/functional/settings_v2/navigation.js
```diff
@@ -0,0 +1,125 @@
...
+  click,
+  createUser,
+  openPage,
+  fillOutEmailFirstSignIn,
+  testElementExists,
+} = FunctionalHelpers.helpersRemoteWrapped;
+
+describe('navigation', () => {
+  let email;
+  beforeEach(async ({ remote }) => {
```
dannycoates, line 30: I think for simpler test suites like nav we should only do the signup/in part once in a `before` and let the `beforeEach` do a fresh navigation to the settings page. It should be noticeably faster that way.

dannycoates (approved):
LGTM. should we be asserting anything after each click?

Issue #4972, opened 2020-04-16
[2FA] QR code functionality


data-sync-user:
Reference the `createToken` from `two_step_authentication.js`. Receive the QR code and display it in "step 1 of 3".

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1660)
┆Issue Number: FXA-1660

PR #6695, opened 2020-10-16
feat(settings): show QR code for two step auth

chenba:
Because:
 - we need to display the QR code for users to set up their
   authenticators

This commit:
 - get and show the QR code

Closes: #4972 

## Checklist


chenba:
Setting it as a draft because I realized that I missed some styles around the QR code.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -17,12 +31,38 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
+      setSecret(x.createTotp.secret);
+      setRecoveryCodes(x.createTotp.recoveryCodes);
+    },
+    onError: () => {
+      alertBar.error('There was a problem retrieving your code.');
+    },
+  });
+
+  useEffect(() => {
+    createTotp({ variables: { input: {} } });
```
dannycoates, line 55: with an unverified session I get this screen...
Screen Shot 2020-10-19 at 12 37 32 PM

I think we need to put this call behind the guard somehow.

dannycoates (commented):
UI looks good. Other than the unverified session issue this is r+

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -17,12 +32,40 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
+    },
+    onError: () => {
+      alertBar.error('There was a problem retrieving your code.');
+    },
+  });
+
+  const session = useSession();
+
+  useEffect(() => {
+    session.verified && createTotp({ variables: { input: {} } });
```
chenba, line 58: @dannycoates made this change.  Thanks!

dannycoates (approved):
looks great!

Issue #6689, opened 2020-10-15
[payments] Add new customer when handling new subscription webhook event


data-sync-user:
In the past when we handle a new subscription webhook event, we query Stripe for the customer first.  After the changes to store customers in the DB, we no longer do that.  We query the db first.  This breaks the scenario where subscription is created directly on Stripe.

Relevant changes: https://github.com/mozilla/fxa/commit/466db2dddba1e3266fc4abcf8ded7d54b06fc115#diff-bcd34199bf3365c46cda6e45302a0b3b1ebd45899f1602d99ad3b2616efedf35L393

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2620)
┆Issue Number: FXA-2620

PR #6705, opened 2020-10-19
feat(subscriptions): handle 'customer.created' Stripe event

chenba:
Because:
 - we should add a Stripe customer to the local DB when a customer is
   created on Stripe

This commit:
 - handle the 'customer.created' webhook event from Stripe by inserting
   into the local DB if that account does not already have a record

Closes: #6689 

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -202,6 +202,13 @@ export class StripeHelper {
   }
 
   /**
+   * Insert a local db record for a customer that already exist on Stripe.
+   */
+  async createLocalCustomer(uid: string, stripeCustomer: Stripe.Customer) {
+    await createAccountCustomer(uid, stripeCustomer.id);
+  }
+
+  /**
    * Update an existing customer to use a new payment method id.
    */
   async retryInvoiceWithPaymentId(
@@ -414,7 +421,19 @@ export class StripeHelper {
       return;
     }
 
-    if (customer.metadata.userid !== uid) {
+    // Since the uid is just metadata and it isn't required when creating a new
+    // customer _on Stripe dashboard_, we have an edge case where the customer
+    // is created on Stripe and locally via the `customer.created` event, but
+    // the uid metadata is still missing.  Throwing an error here causes a
+    // profile fetch to fail, thus would block the user completely.
+    //
+    // Customers created through our regular flow will always have their uid in
+    // the metadata.
+    //
+    // So, we'll only throw an error if the uid metadata is found and it does
+    // not match.
+
+    if (customer.metadata.userid && customer.metadata.userid !== uid) {
       // Duplicate email with non-match uid
       const err = new Error(
         `Stripe Customer: ${customer.id} has mismatched uid in metadata.`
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -201,6 +201,21 @@ export class StripeHelper {
     return stripeCustomer;
   }
 
+  /**
+   * Insert a local db record for a customer that already exist on Stripe.
+   */
+  async createLocalCustomer(uid: string, stripeCustomer: Stripe.Customer) {
+    // There isn't a nice way to write INSERT...WHERE NOT EXIST in
```
bbangert, line 208: This solution doesn't work?
https://github.com/knex/knex/issues/1322#issuecomment-639618321

chenba: I'm sure dropping down into `knex.raw` would work.

bbangert: Nothing wrong with using `knex.raw`, the intention with knex/objection was to get typing and a dynamic query builder when needed. If the query is static, I'm all in favor of using `knex.raw` so that its easier to maintain the queries without having to learn knex/objection query syntax in addition to SQL.

chenba: Updated `createAccountCustomer` to use INSERT ... WHERE NOT EXISTS.

packages/fxa-shared/db/models/auth/index.ts
```diff
@@ -45,10 +45,25 @@ export async function createAccountCustomer(
...
+      'createdAt',
+      'updatedAt',
+    ])
+  ).insert(
+    knex.raw(
+      'SELECT ?, ?, ?, ? FROM DUAL WHERE NOT EXISTS (SELECT 1 FROM `accountCustomers` WHERE `uid` = ?)',
+      [uidBuffer, stripeCustomerId, now, now, uidBuffer]
+    )
+  );
+  return await getAccountCustomerByUid(uid);
```
bbangert, line 66: Don't need to await here, do we?

packages/fxa-auth-server/test/local/routes/subscriptions.js
```diff
@@ -1736,6 +1744,28 @@ describe('DirectStripeRoutes', () => {
         isActive
       );
 
+    describe('handleCustomerCreatedEvent', () => {
```
bbangert, line 1747: We should have a unit test for the condition that a customer is created without a userid in the metadata.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -414,7 +421,19 @@ export class StripeHelper {
       return;
     }
 
-    if (customer.metadata.userid !== uid) {
+    // Since the uid is just metadata and it isn't required when creating a new
+    // customer _on Stripe dashboard_, we have an edge case where the customer
```
bbangert, line 425: Nice catch on this case! While our instructions indicate the new record should have the userid set in metadata, I'm sure someone will forget to add it at some point before going back and updating the customer record for it. I suppose we could handle it one of two ways, this way, where we just ignore that its missing the userid and link based on email (which I think is fine), or we could also listen for `customer.updated` and if `userid` is being added in metadata, then add the local record. This approach here seems fine to me.

packages/fxa-auth-server/test/local/routes/subscriptions.js
chenba: Since we don't use the userid from the metadata, that unit test wouldn't be any different than the one below.

bbangert (approved):

Issue #6651, opened 2020-10-08
[FxA-Settings] 2FA “Refresh” button focus overlaps the “Disable” button


vasilicatamas:
**Build:**
Stage [Train 1.190.1]

**Affected platforms:**
Windows 10

**Steps to reproduce:**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/
2. Login using an existing user that has already enabled the two-step authentication.
3. Go to Two-step authentication section.
4. Click on the “Refresh” button.

**Expected results:**
”Refresh” button focus is correctly displayed.

**Actual Results:**
There is no space between “Refresh” button focus and “Disable” button.

**Additional notes:**
This issue is also reproducible for "Recovery key" Refresh button.
![Screenshot_16](https://user-images.githubusercontent.com/18114210/95479155-20cbc600-0993-11eb-869d-b28d0b998926.png)
![Screenshot_19](https://user-images.githubusercontent.com/18114210/95482246-92f1da00-0996-11eb-8c58-be8fc902e9d0.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2608)
┆Issue Number: FXA-2608

vasilicatamas:
This issue is verified as fixed on Stage [1.192.0] under Windows 10 64-bit. "Refresh" is no longer sticked on the other buttons.

![Screenshot_25](https://user-images.githubusercontent.com/18114210/97602303-d11d6f00-1a13-11eb-859f-e8b4e4f221a1.png)
![Screenshot_26](https://user-images.githubusercontent.com/18114210/97602310-d2e73280-1a13-11eb-8eff-128a5ecc58a7.png)

PR #6709, opened 2020-10-20
fix(fxa-settings): add refresh button margin

meandavejustice:
## This pull request

- Adds a bit of margin-left to the refresh button(same amount applied between normal buttons `ml-2`) so it doesn't look odd when focused

Closes: #6651 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
+++ b/packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
@@ -95,7 +95,7 @@ export const UnitRowTwoStepAuth = () => {
       headerContent={
         <ButtonIconReload
           title="Refresh two-step authentication"
-          classNames="mobileLandscape:hidden"
+          classNames="ltr:ml-1 rtl:mr-1 mobileLandscape:hidden"
           disabled={accountLoading}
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/UnitRowTwoStepAuth/index.tsx
```diff
@@ -94,15 +94,15 @@ export const UnitRowTwoStepAuth = () => {
       headerContent={
         <ButtonIconReload
           title="Refresh two-step authentication"
-          classNames="mobileLandscape:hidden"
+          classNames="ml-2 mobileLandscape:hidden"
```
dannycoates, line 97: does this look correct in rtl?

meandavejustice: Oh good point, gotta get in the habit of checking that. I'm going to add a checklist item to the pr template, make it part of our process.

meandavejustice:
Found a couple other spots that needed the cleanup, also bumped it down to `ml-1` from `ml-2` so it looks closer to mockups. Addressed rtl.

dannycoates (approved):
`ml-1` should be `ltr:ml-1` but r+

PR #6716, opened 2020-10-21
Show modal on connected services 'sign out' click

### LINKED ISSUE #5099: [services] Display modal on "sign out" click, add dummy Sync disconnect survey
On click of a "Sign out" button from sync device (the only button), trigger the modal.

Add the disconnect from Sync survey - reference the relevant content-server files. There should not be an item selected by default. Style accordingly.

"Cancel" should close the modal and "Sign out" can do nothing at this point, and will be done in a follow up issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1760)
┆Issue Number: FXA-1760

6a68:
Fixes #5099.

## Because

- We want to show a confirmation modal when the user clicks the 'sign out' button on a sync device in the new-settings connected services list

## This pull request

- Displays a modal when 'sign out' is clicked on a connected service item. 

## Questions for reviewers:

- The issue description mentions pulling the sync disconnect survey from "the relevant content-server files". Where exactly are these files?
- Should I modify the `Checkbox` component to allow me to pass in styling for the checkbox labels, or should I just use extra specificity to override the default label size? What's the right Tailwind-y React-y approach here?
- The specs show device specific info in the modal text. What's the nicest way to pass the specific device info back to the parent component, so that it passes this into the modal? Can I pass device state back using `revealModal`?
- Do we have an issue to ensure we only show the 'Sign Out' button if the connected service is sync? It's currently always set to show the sign out button.

Closes: #5099

## Checklist

## Screenshots (Optional)

Here's a screenshot of the modal from the spec, see also the [new settings figma](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)) for the real thing:

Screen Shot 2020-10-20 at 5 45 42 PM

Here's a screenshot of my WIP modal:

Screen Shot 2020-10-20 at 5 28 09 PM

### DIFF
--- a/packages/fxa-settings/src/components/Checkbox/index.tsx
+++ b/packages/fxa-settings/src/components/Checkbox/index.tsx
@@ -97,7 +97,7 @@ export const Checkbox = ({
       {label && (
         <span
           data-testid={formatDataTestId('checkbox-label')}
-          className="text-sm ml-3 font-body"
+          className="leading-tight ltr:ml-2 rtl:mr-2 font-body"
         >
           {label}
         </span>
--- a/packages/fxa-settings/src/components/ConnectedServices/Service.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/Service.tsx
@@ -3,9 +3,16 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
+
 import { LinkExternal } from 'fxa-react/components/LinkExternal';
+import { useBooleanState } from 'fxa-react/lib/hooks';
+
+import { Checkbox } from '../Checkbox';
 import { DeviceLocation } from '../../models/Account';
 import { Icon } from './Icon';
+import { Modal } from '../Modal';
+import { useAlertBar } from '../../lib/hooks';
+import { VerifiedSessionGuard } from '../VerifiedSessionGuard';
 
 export function Service({
   name,
@@ -20,9 +27,11 @@ export function Service({
   lastAccessTimeFormatted: string;
   canSignOut: boolean;
 }) {
+  const alertBar = useAlertBar();
   const { city, stateCode, country } = location;
   const locationProvided = Boolean(city && stateCode && country);
   let serviceLink, iconName;
+  const [modalRevealed, revealModal, hideModal] = useBooleanState();
 
   switch (name) {
     case 'Pocket':
@@ -42,7 +51,8 @@ export function Service({
       iconName = 'fpn';
       break;
     case 'Firefox Sync':
-      serviceLink = 'https://support.mozilla.org/en-US/kb/how-do-i-set-sync-my-computer';
+      serviceLink =
+        'https://support.mozilla.org/en-US/kb/how-do-i-set-sync-my-computer';
       iconName = 'sync';
       break;
     default:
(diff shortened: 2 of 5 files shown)


dannycoates:
> The issue description mentions pulling the sync disconnect survey from "the relevant content-server files". Where exactly are these files?

https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/templates/settings/client_disconnect.mustache

>  Should I modify the `Checkbox` component to allow me to pass in styling for the checkbox labels, or should I just use extra specificity to override the default label size? What's the right Tailwind-y React-y approach here?

Since this is the first (only?) place to use `<Checkbox>` I recommend styling the `<label>` directly to your needs. It's current style was just a first pass.

dannycoates:
> The specs show device specific info in the modal text. What's the nicest way to pass the specific device info back to the parent component, so that it passes this into the modal? Can I pass device state back using `revealModal`?

What if you put the modal inside the `Service` component?

dannycoates:
> Do we have an issue to ensure we only show the 'Sign Out' button if the connected service is sync? It's currently always set to show the sign out button.

I think "Sign out" is analogous to "Disconnect" on old settings, so I think the button will always show but the action/modal depends on the type. IDK if there's a specific issue for it. The closest I could find is FXA-1761

6a68:
> > The specs show device specific info in the modal text. What's the nicest way to pass the specific device info back to the parent component, so that it passes this into the modal? Can I pass device state back using `revealModal`?
> 
> What if you put the modal inside the `Service` component?

Because the modal is page-level UI that appears outside the boundaries of the row item, it seemed like the parent page-level component should manage it. But that would solve this particular issue, so I'll give it a try 👍

dannycoates:
> Because the modal is page-level UI that appears outside the boundaries of the row item

I don't think that distinction matters for our use of it. There's several components that embed modal dialogs on the page at any given time, but since they're modal there is no conflict.

6a68:
Updated, thanks for the comments @dannycoates :beers:

The old sync disconnect survey used radio buttons to ensure users only picked one option, but I went with checkboxes because they look like checkboxes in the spec...this could be adjusted later.

Here's a screenshot of how it looks now, a bit of a departure from the strings in the spec, but works better with the data that we have:

Screen Shot 2020-10-21 at 2 08 49 PM

packages/fxa-settings/src/components/Checkbox/index.tsx
```diff
@@ -97,7 +97,7 @@ export const Checkbox = ({
       {label && (
         <span
           data-testid={formatDataTestId('checkbox-label')}
-          className="text-sm ml-3 font-body"
+          className="leading-tight ml-2 font-body"
```
dannycoates, line 100: `ml-2` should be `ltr:ml-2 rtl:mr-2`

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -101,11 +111,69 @@ export function Service({
...
+              <p id="sign-out-desc" className="my-4 text-center">
+                Your browsing data will remain on your device ({name}), but it
+                will no longer sync with your account.
+              </p>
+
+              <p className="my-4 text-center">
+                What's the main reason for disconnecting this device?
+              </p>
+
+              <ul className="my-4 text-left">
```
dannycoates, line 156: is the `text-left` here doing something? If it is we should also handle rtl

dannycoates (approved):
LGTM 👍  ... just rtl nits

PR #6722, opened 2020-10-21
feat(settings): add step 2 of two step auth

chenba:
Because:
 - the user need to enter a valid auth code
 - the user should save recovery codes

This commit:
 - verify the code from the user's authenticator app
 - display the recovery codes

Closes: #4973 

## Checklist

- [x ] If applicable, I have modified or added tests which pass locally.

### DIFF
--- a/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
+++ b/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { gql } from '@apollo/client';
-import { RouteComponentProps } from '@reach/router';
+import { RouteComponentProps, useNavigate } from '@reach/router';
 import { useForm } from 'react-hook-form';
 import { useAlertBar, useMutation } from '../../lib/hooks';
 import FlowContainer from '../FlowContainer';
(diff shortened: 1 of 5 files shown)


packages/fxa-settings/src/lib/totp.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export async function checkCode(
+  secret: string,
+  code: string,
+  timestamp: number = Date.now(),
+  tries = 2
+): Promise<boolean> {
+  while (tries--) {
+    const x = await getCode(secret, 6, timestamp);
```
dannycoates, line 48: shouldn't `timestamp` change for each try?

chenba: 🤦 should be one of the default GH comment reaction emojis, because 🤦.  Thanks for catching that.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -56,73 +75,104 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
+        </form>
+      )}
+      {totpVerified && (
+        <form>
+          <div className="my-2" data-testid="2fa-recovery-codes">
+            Save these one-time use codes in a safe place for when you don’t
+            have your mobile device.
+            <div className="mt-6 flex flex-col items-center h-40 justify-between">
+              <DataBlock value={recoveryCodes}></DataBlock>
+              <GetDataTrio value={recoveryCodes}></GetDataTrio>
```
dannycoates, line 159: I think we need to adjust the string formatting of `GetDataTrio` actions for the string array case. _Download_ for example gives me one long string

dannycoates (approved):
LGTM. just the nit with the trio actions

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
chenba: We've found that https://github.com/mozilla/fxa/issues/4975 should handle that issue.

Issue #4976, opened 2020-04-16
[2FA] Create/display step 3 of 3 on "Continue"


data-sync-user:
Create the "step 3 of 3" component to display on "Continue" press from step 2 of 3. "Back" should take the user back to step 2 and should not regenerate the codes.

"Finish" should be enabled when at least one character has been typed in the recovery key input but functionality will be done in a follow-up issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1667)
┆Issue Number: FXA-1667

PR #6738, opened 2020-10-23
feat(settings): verify recovery code and enable two step auth

chenba:
Because:
 - we need to verify a recovery code and enable two step auth as the
   last step in the enabling two step auth

This commit:
 - check the recovery code and then verify the totp with gql-api

Closes: #4976 

## Checklist


dannycoates (approved):
LGTM 👍

Issue #4977, opened 2020-04-16
[2FA] Process form response from "Finish" (GQL)


data-sync-user:
On "Finish" button form submission, capture the input, send it up to the server in a GQL mutation, and receive the response from the server.

A successful response on should 1) take the user back to the main settings page 2) with a success message.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1669)
┆Issue Number: FXA-1669

PR #6744, opened 2020-10-26
feat(settings): show 2fa enabled message

chenba:
Because:
 - we should let the user know that 2fa was successfully enabled

This commit:
 - display a message on the main settings page after enabling 2fa

Closes: #4977

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
+++ b/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
@@ -19,6 +19,7 @@ import { useSession } from '../../models';
 import { checkCode, getCode } from '../../lib/totp';
 import { HomePath } from '../../constants';
 import { cloneDeep } from '@apollo/client/utilities';
+import { alertTextExternal } from '../../lib/cache';
 
 export const CREATE_TOTP_MUTATION = gql`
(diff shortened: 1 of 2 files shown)


chenba:
Note that the form submission handling requirement in #4977 is already done (as a part of PR #6738).

chenba:
Test failure not related to changes to this PR.  Back to draft status while I investigate.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -125,6 +126,7 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
 
   const [verifyTotp] = useMutation(VERIFY_TOTP_MUTATION, {
     onCompleted: () => {
+      alertTextExternal('Two-step authentication enabled');
```
dannycoates, line 129: This reminded me that I wanted to refactor `AlertBar` to store the state in the cache so that we wouldn't need to use external for this type of thing

dannycoates (approved):
👍

chenba:
Noticed some inconsistent results (between runs and envs) from tests written for PR #6738.  Not sure why it popped now.  I've wrapped some expectations in `wait()`; hopefully that'll fix what I suspect to be timing issues with callbacks.

Issue #5177, opened 2020-04-30
[2FA] Log parity two step events


data-sync-user:
Log two factor step related events that are currently being logged in current settings.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1806)
┆Issue Number: FXA-1806

PR #6781, opened 2020-10-28
feat(settings): add metrics for two step auth

chenba:
Because:
 - we try to achieve parity with content server on metrics

This commit:
 - emit metric events for two step auth
 - add a couple new helper functions to metrics lib

Closes: #5177

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
+++ b/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
@@ -20,6 +20,9 @@ import { checkCode, getCode } from '../../lib/totp';
 import { HomePath } from '../../constants';
 import { cloneDeep } from '@apollo/client/utilities';
 import { alertTextExternal } from '../../lib/cache';
+import { logViewEvent, useMetrics } from '../../lib/metrics';
+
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -20,6 +20,9 @@ import { checkCode, getCode } from '../../lib/totp';
 import { HomePath } from '../../constants';
 import { cloneDeep } from '@apollo/client/utilities';
 import { alertTextExternal } from '../../lib/cache';
+import { logViewEvent, useMetrics } from '../../lib/metrics';
+
+export const metricsPreInPostFix = 'settings.two-step-authentication';
```
dannycoates, line 25: this var name is cryptic to me but 🤷

dannycoates (approved):
LGTM 🍻

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
chenba: That string can be a pre-, in-, or post-fix, so yeah 🤷

PR #6782, opened 2020-10-28
chore(build): stop building docker images for each service by default

dannycoates:
## Because

- We're using `fxa-mono` in production and don't need images for each service anymore.

## This pull request

- Eliminates the build step that generates the default individual images. Custom dockerfiles and build scripts are still run.


6a68 (approved):
🎉 Great to see us simplifying our server setup

PR #6783, opened 2020-10-28
First pass at enabling docker login

ckolos:
Login using *any* configured docker credentials to avoid rate limiting from dockerhub

## Because

- Dockerhub will be enforcing rate-limiting for non-auth'd logins

## This pull request

- This PR will force a `docker login` to auth prior to any docker operations requiring login (pull/push/etc)

## Checklist

- [x ] My commit is GPG signed.

This will only affect the build process for docker containers. No application code should be affected.

### DIFF
--- a/.circleci/build-all.sh
+++ b/.circleci/build-all.sh
@@ -5,6 +5,11 @@ cd "$DIR"
 
 if [[ -n "${CIRCLECI}" ]]; then
   echo "Docker logs are located in the CircleCI build artifacts"
+  if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
+      echo "Skipping Login to Dockerhub, credentials not available."
+  else
+      echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
+  fi
 fi
 
 ../_scripts/build-builder.sh
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -13,6 +13,12 @@ if grep -e "$MODULE" -e 'all' "$DIR/../packages/test.list" > /dev/null; then
 
   mkdir -p ../../artifacts
 
+  if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
+      echo "Skipping Login to Dockerhub, credentials not available."
+  else
+      echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
+  fi
+
   if [[ -x scripts/build-ci.sh ]]; then
     time ./scripts/build-ci.sh
   elif [[ -r Dockerfile ]]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,9 @@ executors:
           FXA_EMAIL_ENV: dev
           FXA_EMAIL_LOG_LEVEL: debug
           RUST_BACKTRACE: 1
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
     environment:
       FXA_MX_RECORD_EXCLUSIONS: restmail.dev.lcip.org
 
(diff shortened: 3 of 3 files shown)


.circleci/config.yml
```diff
@@ -90,6 +93,9 @@ jobs:
       - image: memcached
       - image: pafortin/goaws
       - image: circleci/mysql:5.7.27
+        auth:
```
dannycoates, line 96: It isn't clear to me from their docs if `auth` is needed for each `image` or not. As far as yml structure goes I think it's only attached to one.

.circleci/build-all.sh
```diff
@@ -5,6 +5,11 @@ cd "$DIR"
 
 if [[ -n "${CIRCLECI}" ]]; then
   echo "Docker logs are located in the CircleCI build artifacts"
+  if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
+      echo "Skipping Login to Dockerhub, credentials not available."
+  else
+      echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
+  fi
```
dannycoates, line 12: I don't think we need to login for the build scripts, just deploy

.circleci/config.yml
```diff
@@ -15,6 +15,9 @@ executors:
           FXA_EMAIL_ENV: dev
           FXA_EMAIL_LOG_LEVEL: debug
           RUST_BACKTRACE: 1
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
```
dannycoates, line 20: Where are these define? I guess they'll be here? https://app.circleci.com/settings/project/github/mozilla/fxa/environment-variables?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fmozilla%2Ffxa

dannycoates (commented):
As noted near the top of the [new circleci docs](https://circleci.com/docs/2.0/private-images/) this isn't as urgent as we thought

.circleci/build.sh
```diff
@@ -13,6 +13,12 @@ if grep -e "$MODULE" -e 'all' "$DIR/../packages/test.list" > /dev/null; then
 
   mkdir -p ../../artifacts
 
+  if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
```
Cronus1007, line 16: in my opinion there are no such requirements to login for the build.sh scripts.

.circleci/config.yml
```diff
@@ -15,6 +15,9 @@ executors:
           FXA_EMAIL_ENV: dev
           FXA_EMAIL_LOG_LEVEL: debug
           RUST_BACKTRACE: 1
+        auth:
```
Cronus1007, line 18: Plzzz avoid changing the config files.

Cronus1007 (changes requested):

.circleci/build-all.sh
ckolos: It's my understanding that *any* docker actions (pull/push) will be subject to the rate-limiting, so any images used during build would count against that as well.

.circleci/config.yml
ckolos: Nah, they're actually def'd in the circleci env

.circleci/config.yml
ckolos: As above, I'm thinking that any docker hub actions will require auth; I'll ask around a bit on the team to see if I'm misunderstanding

.circleci/build-all.sh
dannycoates: build doesn't do any docker actions

.circleci/config.yml
dannycoates: isn't that what i said?

dannycoates:
im tired of seeing this in the open pr list. 😜 we shouldn't need it for a while

PR #6798, opened 2020-10-30
fix(fxa-settings): format 2fa recovery codes for export

meandavejustice:
## This pull request

- finishes up #4975, by correcting the formatting for COPY, PRINT, and DOWNLOAD functionality of the recovery codes.

Closes: #4975

## Checklist

The main change here is in that we are now creating a template for the print file in function `recoveryCodesPrintTemplate`.
also we are joining the arrays with ` \r\n`, two things we were doing in the content server. I'm a little unsure of a reliable way to test the template as it is just a string. Also not sure if we need to be testing a template like this, it may be redundant. Happy to take input of the reviewer on whatever we think is correct.

### DIFF
--- a/packages/fxa-settings/src/components/GetDataTrio/index.tsx
+++ b/packages/fxa-settings/src/components/GetDataTrio/index.tsx
@@ -13,21 +13,34 @@ export type GetDataTrioProps = {
   onAction?: (type: string) => void;
 };
 
+const recoveryCodesPrintTemplate = (recoveryCodes: string | string[]) => {
+  if (typeof recoveryCodes === 'string') recoveryCodes = [recoveryCodes];
+  return (`
+    <html>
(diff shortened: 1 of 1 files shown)


packages/fxa-settings/src/components/GetDataTrio/index.tsx
```diff
@@ -47,7 +60,7 @@ export const GetDataTrio = ({ value, onAction }: GetDataTrioProps) => {
         title="Copy"
         type="button"
         onClick={async () => {
-          const copyValue = Array.isArray(value) ? value.join(', ') : value;
+          const copyValue = Array.isArray(value) ? value.join(' ') : value;
```
dannycoates, line 63: lets use `'\r\n'` for both of these

dannycoates (approved):

Issue #6778, opened 2020-10-28
[FxA-Settings] Change password separate page incorrectly opened on Android


vasilicatamas:
**Build**
Stage [Train 1.192.0]

**Affected platforms**
Android 9

**Affected devices**
Motorola G6
Samsung S8

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Tap on the "Change" button displayed on the "Password" section.
3. Observe how the “Change password” separate page is displayed.

**Expected results**
The entire “Change password” form is in focus. 

**Actual Results**
The header of the “Change password” form is not visible.

![Screenshot_16](https://user-images.githubusercontent.com/18114210/97431045-6809fe80-1922-11eb-86d4-32405366546b.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2651)
┆Issue Number: FXA-2651

data-sync-user:
➤ bchen@mozilla.com commented:

This does not affect only Android, or only mobile, or on only the change password page.  The user just need to navigate to a page that's taller the the browser window.  (And the change password page just happens to be rather tall relatively to the other pages.)

data-sync-user:
➤ bchen@mozilla.com commented:

Ah ok, it has to be with teh scroll position on the main settings page before navigating into a feature page.  Since the change password button is farther down the main page (when in mobile view), it has the reported effect.

data-sync-user:
➤ bchen@mozilla.com commented:

Thanks for adding the point estimate, [~wclouser@mozilla.com].

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.193.0] under Android 9. Change password separate page is correctly displayed.

![issue36](https://user-images.githubusercontent.com/18114210/98662079-f08b9480-234f-11eb-8662-29487306dfcf.gif)

PR #6812, opened 2020-11-01
fix(settings): scroll to top on navigation

chenba:
Because:
 - a new page should start at the top

This commit:
 - scroll to the top when a user navigates to a new page (but not when
   back/forward with browser history)

Closes: #6778

## Checklist


dannycoates (approved):
👍

Issue #6795, opened 2020-10-29
[FxA-Settings] 2FA section content is incorrectly aligned


vasilicatamas:
**Build**
Stage [Train 1.192.0]

**Affected platforms**
Windows 10 64-bit

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/
2. Login using an existing user that has already enabled the two-step authentication.
3. Go to Two-step authentication section and check how “Enabled
Prevent someone else from logging in by requiring a unique code only you have access to.” text is displayed.

**Expected results**
2FA section content is correctly aligned with the content of the other sections.

**Actual Results**
2FA section content is displayed a bit to the left compared with the content of the other sections

![Screenshot_27](https://user-images.githubusercontent.com/18114210/97605580-7c7bf300-1a17-11eb-826d-bf450b7edf4f.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2659)
┆Issue Number: FXA-2659

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.193.0] under Windows 10 64-bit. 2FA section content is correctly aligned.
![Screenshot_27](https://user-images.githubusercontent.com/18114210/98560233-385adf00-22b0-11eb-8e92-4dc471c39617.png)

PR #6836, opened 2020-11-04
fix(fxa-settings): fix 2fa unit row alignment

meandavejustice:
## This pull request

- Sets a min-width to the unit-row-header so that we don't have misalignment in the 2fa section

Closes: #6795

## Checklist

### DIFF
--- a/packages/fxa-settings/src/styles/unit-row.scss
+++ b/packages/fxa-settings/src/styles/unit-row.scss
@@ -34,7 +34,8 @@
     @apply flex-no-wrap px-6;
 
     &-header {
-      @apply flex-1 mb-0 mr-2;
+      @apply flex-none mb-0 mr-2;
+      width: 10rem
     }
 
     &-content {


packages/fxa-settings/src/styles/unit-row.scss
```diff
@@ -7,6 +7,8 @@
 
   &-header {
     @apply font-header w-full mb-1;
+    min-width: 165px;
```
dannycoates, line 10: there's some screen widths where this doesn't work
Screen Shot 2020-11-04 at 9 41 47 AM

here's a diff that works better
```diff
diff --git a/packages/fxa-settings/src/styles/unit-row.scss b/packages/fxa-settings/src/styles/unit-row.scss
index 980a82d29..f03b824ff 100644
--- a/packages/fxa-settings/src/styles/unit-row.scss
+++ b/packages/fxa-settings/src/styles/unit-row.scss
@@ -27,21 +27,22 @@
   &-hr {
     @apply mx-5 border-grey-100;
   }
 }
 
 @screen mobileLandscape {
   .unit-row {
     @apply flex-no-wrap px-6;
 
     &-header {
-      @apply flex-1 mb-0 mr-2;
+      @apply flex-none mb-0 mr-2;
+      width: 10rem;
     }
 
     &-content {
       @apply flex-2;
     }
 
     &-actions {
       @apply flex-1 flex justify-end;
     }
   }
```

packages/fxa-settings/src/styles/unit-row.scss
```diff
@@ -7,6 +7,8 @@
...
+
```
chenba, line 11: The New Settings Style Guide: "... it's important to keep in mind that our CSS system for FxA styles increase in units of 4px. If a design reflects 17px, it's generally safe to go with the closest value divisible by 4, 16px (which is equivalent to 1rem) ..."

Any reason we cannot use a `rem` here?

On my computer, the row header for the other rows were less than a pixel wider than the one for the enabled Two Step Auth row.

meandavejustice:
@dannycoates Applied the changes you suggested and tested in storybook.

dannycoates (approved):
👍

Issue #5101, opened 2020-04-23
[services] Process Sync survey form submission (GQL)


data-sync-user:
On form submission, capture the user's input response and send the GQL mutation request to the GQL endpoint to destroy the attached client. Log the survey response.

If the survey response is "lost or stolen" or "suspicious" but the submission was successful, 1) display the corresponding text shown in https://docs.google.com/document/d/1LG29ep89juuriFtt9N83K6DkIZt_TTVpM4gq0AFegCk/edit# and on click of the "Okay, got it" button, 2) close the modal and 3) update the row in the Connected Services list to display "Logged out." Do not trigger the global success message.

A successful response should 1) close the modal 2) with a success message triggering the global alert message while 3) updating the row in the Connected Services list to display "Logged out."

Any server errors should be displayed similar to the avatar wire server errors.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1761)
┆Issue Number: FXA-1761

data-sync-user:
➤ Jared Hirsch commented:

Bumping points from 5 up to 8; this was a big one.

PR #6841, opened 2020-11-04
Connected Services modal and survey submission and secondary modal

6a68:
This PR is all but done; updating the description to make it easier to understand its current state.

## Because

- We need to replicate the connected services disconnect flow in new-settings for Web sessions and for Sync clients
- Sync clients should be shown a survey asking their primary reason for disconnecting the listed device; if the reason is a lost or suspicious device, we then show a second info modal with further instructions.

## This pull request

Fully wires up the connected services feature:
- disconnect handling on the GQL server (including exposing some new keys needed for the mutation to work)
- implementation of both modals for the Sync case and logic to manage the flow
- logic to avoid showing a "sign out" button for OAuth clients we can't sign out

Closes: #5101

## Checklist

## Screenshots (Optional)

See screenshots in [comment below](https://github.com/mozilla/fxa/pull/6841#issuecomment-725014683)

I'd prefer to land this as-is and fix remaining minor issues in followup bugs. That said, there are two areas where I'd welcome advice / suggestions:

- I've run into some complex and seemingly non-deterministic timing issues with toggling react component state, and wound up moving the disconnect-related logic into a plain JS data object. To fetch the disconnect reason from the DOM without relying on react state, I switched to unmanaged radio buttons and obtained the user's selection by checking the DOM directly, sidestepping react.

- [fixed] I also ran into some snags trying to add the type-related information suggested in an earlier review, and wound up disabling TS checks because our EventTarget type doesn't, for some odd reason, have a `closest()` method in its API. Is there an easy way to add missing methods to our base set of types?

- [done] There are also a few TODOs left in the code, comments welcome there


6a68:
Moving the click handlers into the parent, since the child rows are getting removed before the mutation / secondary modal can be managed.

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -20,18 +23,41 @@ export function Service({
...
   name: string;
   deviceType: string | null;
+  deviceId: string | null;
   location: DeviceLocation;
   lastAccessTimeFormatted: string;
   canSignOut: boolean;
+  isCurrentSession: boolean;
+  clientId: string;
+  sessionTokenId: string | null;
+  refreshTokenId: string | null;
```
chenba, line 41: There's a `hexstring` type alias that we could use for some of these.

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -20,18 +23,41 @@ export function Service({
...
+  const [
+    secondaryModalRevealed,
+    revealSecondaryModal,
+    hideSecondaryModal,
+  ] = useBooleanState();
+  const [disconnectReason, setDisconnectReason] = useState<string>();
+  const [disconnected, setDisconnected] = useBooleanState(false);
+  // Note: setting a default start time to avoid TS complaints about submitStartTime
+  // possibly being undefined. It's not a meaningful number.
+  const [submitStartTime, setSubmitStartTime] = useState<number>(Date.now());
```
chenba, line 57: We could add a `!` to where the error is (`submitStartTime!`) to let the type system that we know what we are doing.  Or, if we need to initialize a value to satisfying the type system, then let's use an obviously incorrect value like `0`?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -67,6 +93,88 @@ export function Service({
...
+          viewName,
+          `timing.clients.disconnect.${Date.now() - submitStartTime}`
+        );
+
+        if (isCurrentSession) {
+          clearSignedInAccountUid();
+          window.location.assign(`${window.location.origin}/signin`);
+        } else if (
+          disconnectReason === 'suspicious' ||
+          disconnectReason === 'lost'
```
chenba, line 125: Use an enum or a sum type for these disconnect reasons?

chenba:
Left a few type related comments; not a full review.  I'll take a look at the metric events and the secondary modal tomorrow.

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
```diff
@@ -2,36 +2,30 @@
...
   lastAccessTimeFormatted,
   canSignOut,
+  handleSignOut,
 }: {
   name: string;
   deviceType: string | null;
   location: DeviceLocation;
   lastAccessTimeFormatted: string;
   canSignOut: boolean;
+  handleSignOut: any;
```
chenba, line 24: `() => void` instead of `any`?

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +61,136 @@ export const ConnectedServices = () => {
     setErrorText('Sorry, there was a problem refreshing the recovery key.');
     alertBar.show();
   });
-
   const showMobilePromo = !sortedAndUniqueClients.filter(isMobileDevice).length;
 
+  // TODO: should this view name be the exact same as in old-settings? "settings.clients.disconnect"?
```
chenba, line 66: We have an event property in Amplitude events to distinguish between old and new settings metrics, so we have been using the same event names.  Not sure if that answers your question here.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +61,136 @@ export const ConnectedServices = () => {
...
+  // an informative modal with some advice on next steps to take.
+  const [
+    adviceModalRevealed,
+    revealAdviceModal,
+    hideAdviceModal,
+  ] = useBooleanState();
+
+  const clearDisconnectingState = (
+    errorMessage?: string,
+    error?: ApolloError | undefined
```
chenba, line 94: The `| undefined` isn't necessary since the `?` already made `error` optional.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: It does, thanks!

6a68:
Thanks for the helpful review @chenba!

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +61,136 @@ export const ConnectedServices = () => {
...
+  ) => {
+    hideConfirmDisconnectModal();
+    // TODO: if I uncomment these lines, stuff breaks. What's the right way to coordinate
+    // hanging on to enough local state until we're done? Maybe not pass around the AttachedClient,
+    // but instead pass around copies of the data? Like the name, clientId, lastAccessTime, etc?
+    //setDisconnectingClient(undefined);
+    //setDisconnectReason(undefined);
+    //setSubmitStartTime(0);
```
chenba, line 102: What was breaking?  I tried uncommenting those lines and didn't see any obvious failures.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +61,136 @@ export const ConnectedServices = () => {
...
+  };
+
+  const onConfirmDisconnect = () => {
+    setSubmitStartTime(Date.now());
+
+    if (!disconnectingClient) {
+      return clearDisconnectingState('Client not found, unable to disconnect');
+    }
+
+    // TODO: does this need to be a _flow_ event?
```
chenba, line 120: I don't think so.  The Amplitude event looks correct to me.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +61,136 @@ export const ConnectedServices = () => {
...
+    }
+  `;
+
+  const [deleteConnectedService] = useMutation(
+    ATTACHED_CLIENT_DISCONNECT_MUTATION,
+    {
+      onCompleted: () => {
+        logViewEvent(
+          viewName,
+          `timing.clients.disconnect.${Date.now() - submitStartTime!}`
```
chenba, line 149: This logs a `settings.clients.disconnect.timing.clients.disconnect.nnn` event, but we want a `flow.timing.clients.disconnect.nnn` event.  I don't think we've started logging any flow timing events in new settings.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: Oh, interesting! I wasn't able to trigger the secondary modal when I eagerly cleared the state. I'll give it another shot

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
6a68: Thanks, dodging that for now since the underlying types (src/models/Account.ts, also fxa-graphql-api/src/lib/resolvers/types/attachedClient.ts) use `string | null` and `string`, respectively

chenba:
Looks like the modal is coming up when disconnecting an OAuth RP, which I think should not be the case.  (You can try this locally with 123Done.)

6a68:
Thanks, yeah. I got a bit tangled up in the suggested TS type sharpening suggestions, might back those out for now and try to just get the main blob landed. Feels like there'll also be some followup work around adding an AttachedService model, too.

6a68:
Lot of small updates:



- only show the disconnect modal for sync clients specifically,
  otherwise proceed to disconnect the client

- insert correct secondary modal text

- minimally style the radio buttons, pending UX feedback

- hide cancel button in secondary modal

- only show Sign Out button in service list for web and sync sessions

- adjust attached client mutation comparator function to match old-settings

- hacky fix to flow timing event, add TODO to correctly use flow metrics

I need to manage disconnectingClient state differently, perhaps doing the `onCompleted` handler work sooner / somewhere else (`update` method? that doesn't seem right though the timing seems right). Disconnecting a sync client works ok, but disconnecting a web session doesn't work properly because of this disconnectingClient state issue. Also the 'lost' vs 'suspicious' modal state check is happening after disconnectingClient is unset. Suggestions welcome!

Also need to pass the correct stuff into AlertBar.

Here are some screenshots:

Screen Shot 2020-11-10 at 10 07 15 AM
Screen Shot 2020-11-10 at 11 36 55 AM
Screen Shot 2020-11-10 at 10 06 41 AM
Screen Shot 2020-11-10 at 9 57 10 AM

6a68:
I moved the disconnecting client / disconnect reason state into a local JS object, which fixed all the timing issues. But now I need to figure out the best way to sync the radio button state with React state, or just have react leave the radios unmanaged and check the DOM for the selection manually on submit. Suggestions, again, welcome :-)

6a68:
I think all the required functionality is now here and working.

chenba:
Just gave this a quick try. I think there's a bug with updating the local state after disconnecting a device.  With two browsers / profiles, A and B:

 1. Sign into settings with A
 1. Sign into Sync with B
 1. Refresh settings on A; both A and B are listed
 1. Disconnect B via settings on A

Expected: A's session listed in connected services

Actual: No connected services listed

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
chenba: Is there a reason for the `any` here?

6a68:
@chenba Hmm, I can't seem to repro that. I still see the web session after doing the disconnect. Do you see this bug consistently or intermittently? Can you get the web session back in the list if you hit the refresh button? I'm using Nightly for the web session and Developer Edition for the sync/browser session, what are you using?

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -2,22 +2,40 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// @ts-nocheck
```
chenba, line 5: This affects the entire file right?  Kinda overdoing it!

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +73,156 @@ export const ConnectedServices = () => {
...
+    hideConfirmDisconnectModal();
+    DS.client = null;
+    DS.reason = null;
+    setSubmitStartTime(0);
+    if (errorMessage) {
+      alertBar.error(errorMessage, error);
+    }
+  };
+
+  const onConfirmDisconnect = (evt?: MouseEvent) => {
```
chenba, line 113: If we make the type here `React.MouseEvent<HTMLButtonElement, MouseEvent>` then `target` won't be optionally null.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +73,156 @@ export const ConnectedServices = () => {
...
+    setSubmitStartTime(Date.now());
+
+    if (evt) {
+      // TODO: disabled ts-checking because the EventTarget type is missing
+      // 'closest'. Figure out how to most nicely tidy this up, even if that
+      // entails trying again with managed components / react state.
+      const modalEl = evt.target.closest('#modal');
```
chenba, line 120: `(evt.target as Element)`?

chenba:
> 
> 
> @chenba Hmm, I can't seem to repro that. I still see the web session after doing the disconnect. Do you see this bug consistently or intermittently? Can you get the web session back in the list if you hit the refresh button? I'm using Nightly for the web session and Developer Edition for the sync/browser session, what are you using?

Yes, refreshing brings it back.  I only tried it once though.  I'll double check.  I'm using Release on Windows and Ubuntu.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +73,156 @@ export const ConnectedServices = () => {
...
+      if (selected) {
+        DS.reason = selected.value;
+      }
+    }
+
+    if (!DS.client) {
+      return clearDisconnectingState('Client not found, unable to disconnect');
+    }
+
+    logViewEvent('settings.clients.disconnect', `submit.${DS.reason}`);
```
chenba, line 131: hmm `DS.reason` could be null here.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -55,9 +73,156 @@ export const ConnectedServices = () => {
...
+
+  const [deleteConnectedService] = useMutation(
+    ATTACHED_CLIENT_DISCONNECT_MUTATION,
+    {
+      onCompleted: () => {
+        // TODO: followup issue for ensuring flow events are logged correctly,
+        // including utilizing the MetricsContext.
+        logViewEvent(
+          viewName,
+          `flow.timing.clients.disconnect.${Date.now() - submitStartTime!}`
```
chenba, line 177: I don't think the current set of metrics function is really set up to log `flow` events. ☹️  This call logs a `settings.clients.disconnect.flow.timing.clients.disconnect.676`, but we don't want that view name prefix.

chenba: @irrationalagent or @davismtl  I cannot recall how and where the flow timing metrics are used.  Can you link us to an example?

packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -128,6 +304,163 @@ export const ConnectedServices = () => {
...
+                  className="my-4 text-center"
+                >
+                  Since your device was lost or stolen, to keep your information
+                  safe, you should change your Firefox account password in your
+                  account settings. You should also look for information from
+                  your device manufacturer about erasing your data remotely.
+                </p>
+              </>
+            ) : (
+              /* DS.reason === 'suspicious' */ <>
```
chenba, line 441: Commented out for debugging purposes?

6a68:
Thanks for the review @chenba! I pushed an update to switch the post-mutation cache filtering to use `lastAccessTime` and `name`, which is how the services are deduped and sorted for initial layout in new-settings, instead of filtering on various token IDs used by old-settings. Hopefully this resolves any lingering weirdness with post-disconnect rerendering.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: That was just trying to leave a little information for future selves, but I can remove it

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: Yikes, good catch. I'll remove this metrics event and file a followup issue

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: I don't know of any way to disable the check for just one part of the file :-\

chenba: I made a couple suggestions below that I think will obviate the need for this.

chenba:
> I think there's a bug with updating the local state after disconnecting a device.

I just tried it again on the latest commit (c7f15b4c1) and can no longer reproduce.  👍

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: ah, I'm calling `onConfirmDisconnect` from another spot with no argument passed in, if the user deactivated a web session.

packages/fxa-settings/src/components/ConnectedServices/index.tsx
6a68: Whoa, brilliant! That fixed the type error 🍻

6a68:
@chenba That `(evt.target as Element)` trick was fantastic! Do you think this looks nearly ready to land?

packages/fxa-settings/src/components/ConnectedServices/Service.tsx
6a68: Now that I've got TS-check re-enabled, your suggestion works :+1:

packages/fxa-settings/src/components/ConnectedServices/index.tsx
chenba: Sure, that makes `evt` undefined.  I was referring to `target`.  But doesn't matter though, now that we assert that `target` is an Element.

chenba:
> 
> 
> Do you think this looks nearly ready to land?

I think we need to add tests to `ConnectedServices/index.test.tsx` for the changes made to `ConnectedServices/index.tsx`.

6a68:
@chenba OK, added some tests and made a few final tweaks. Mind taking one more pass?

6a68:
Rebased against current `main` branch. hopefully tests will pass 🤞

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.js
```diff
@@ -226,3 +178,62 @@ export const MOCK_SERVICES = [
...
+    approximateLastAccessTime: null,
+    approximateLastAccessTimeFormatted: null,
+  },
+];
+
+export const MOCK_SERVICES = [].concat(
+  DESKTOP_SYNC_MOCKS,
+  OAUTH_SERVICE_MOCKS,
+  MOBILE_SYNC_SERVICE_MOCKS
+);
```
chenba, line 239: hmm don't think I understand this `concat` hipness.  Couldn't a plain [array literal] also work?

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
```diff
@@ -25,6 +33,58 @@ const getIconAndServiceLink = async (name: string, testId: string) => {
...
+];
+
+const clickFirstSignOutButton = async () => {
+  await act(async () => {
+    const signOutButtons = await screen.findAllByTestId(
+      'connected-service-sign-out'
+    );
+    fireEvent.click(signOutButtons[0]);
+  });
+  await wait();
```
chenba, line 63: What does this do when you don't wait for anything in particular?

chenba (approved):
🍻

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.js
6a68: hmm, `concat` is ancient and, now that you mention it, spread syntax is more modern and should do the same thing: `[...mocks, ...otherMocks]`

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
6a68: Not sure? I assume a `setTimeout` to flush the UI loop, but I don't know. I was following the approach used in the other tests in this file, though this method is marked deprecated in favor of `waitFor`, which includes an explicit wait condition.

packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.js
chenba: ah, oops, sorry, those are arrays already.  Gotcha.

PR #6926, opened 2020-11-12
docs: add paypal adr

### LINKED ISSUE #6924: Create ADR for PayPal
We should have an ADR documenting the PayPal approach we're choosing.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2709)
┆Issue Number: FXA-2709

bbangert:
Because:

* We want to document our Paypal integration approach.

This commit:

* Add's an ADR around how we plan to implement PayPal in FxA.

Closes #6924

## Checklist

### DIFF
--- a/docs/adr/0023-paypal-support.md
+++ b/docs/adr/0023-paypal-support.md
@@ -1,0 +1,105 @@
+# Add PayPal as a supported payment option
+
+- Deciders: Ben Bangert, Barry Chen, Tyler Duzan, Bianca Danforth, Se Yeon Kim, Wennie Leung
+- Date: 2020-11-11
+
+## Context and Problem Statement
+
+The Subscription Platform was implemented with Stripe as the payment processor and driver of subscription logic. Stripe is used as the source of truth for whether an account has subscriptions associated with it. We would like to add PayPal as a payment option for subscriptions, and integrate it appropriately with the existing system in a least-effort manner that is ideally flexible enough to support future expansions of payment processors, such as IAP.
+
+## Decision Drivers
+
+- Engineering resources
+- Future effort needed to integrate IAP
+- Code complexity as it relates to existing Subscription Platform infrastructure
+
+## Considered Options
+
+- Stripe-driven out-of-band invoice processing
+- PayPal subscriptions linked in FxA
+- Recurly
+
+## Decision Outcome
+
+Stripe-driven out-of-band invoice processing. This was originally considered to be the slower of the approaches, but using PayPal subscriptions ended up being approximately the same engineering effort. Given additional user experience benefits of the Stripe-driven approach, it is now the recommend implementation path.
+
+## Pros and Cons of the Options
+
+### Stripe-driven out-of-band invoice processing
+
+Stripe defaults subscriptions to billing automatically using a payment source on file with Stripe. Stripe also provides the option to have subscriptions that have a `collection_method` of `send_invoice`. In this model, Stripe can track subscriptions, and generate invoices that can be paid `out-of-band`. This allows Stripe to have an accurate record of charges, invoices, and subscriptions but how the invoice is considered 'paid' is up to FxA.
+
+PayPal has a feature called reference transactions that allow FxA to get a customers authorization to charge their PayPal account. This authorization allows for later payments to be charged against the PayPal account. We will track this reference ID in Stripe metadata for the customer and mark the subscription as `send_invoice`.
+
+FxA will implement additional logic to listen for invoice creation webhooks and switch their `auto_advance` off if the subscription has metadata indicating it should be paid via a PayPal reference ID. We will then check the invoice for the amount to charge and run the PayPal charge via PayPal API. The invoice will then be updated to reflect whether it was paid successfully or not.
+
+Pros:
+
+- Stripe continues to serve as an accurate source of truth regarding a customers subscription.
+- All the logic that determines active subscriptions and entitlements using Stripe can be used as-is.
+- Multiple product support works as-is.
+- Support representatives can see accurate payment history in Stripe for customers paying via PayPal.
+- Support can credit a PayPal customer, just as they can for a credit card customer.
+- Can extend in a similar manner to track a correlating IAP for a user.
+- Can handle switching payment method from credit card to PayPal reference ID, and vice versa.
+- PayPal authorization is retained so that additional products can be purchased/credited without requiring the user to return to the PayPal flow.
+- Cancel/resubscribe before the subscription expires works as-is.
+
+Cons:
+
+- Additional metadata in Stripe can start pushing us closer to metadata limits on objects.
+- Recreate Stripe's [automatic advancement and collection logic](https://stripe.com/docs/billing/invoices/overview#auto_advance) that runs PayPal transactions which we need to ensure work correctly (checks to prevent double-charging, retries, and transaction success verification).
+- We still could end up storing PayPal reference ID's in a separate db table to ensure we track
+  prior PayPal billing.
+- While we could more easily support multiple payment methods on the back end, this would add complexity to the subscription management flow on the front end.
+
+### PayPal Subscriptions linked/tracked by FxA
+
+In this approach, we would create products/plans in PayPal, and link them back to the Stripe equivalent by updating Stripe metadata on the products/plans. FxA would also gain an additional database table that maps PayPal subscription ids to the Fx account. All the logic in FxA that looks up a user's subscriptions would be updated to check the PayPal linking table to determine active subscriptions.
+
+PayPal would handle running the charges on a recurring basis per the PayPal plan that is setup. FxA would periodically check the subscriptions via PayPal API to ensure they're still actively being paid (PayPal has fairly lenient handling on whether payments occurred) on the schedule we want.
+
+Pros:
+
+- PayPal is responsible for charging the customer and managing the subscription lifecycle.
+- Can extend in a similar manner to track a correlating IAP for a user.
+
+Cons:
+
+- All the logic that determines active subscriptions and entitlements needs additional
+  logic/abstractions to query the database table and map active PayPal subscriptions to the
+  Stripe subscription.
+- Ability to easily cancel/re-subscribe on management page is lost, as canceled PayPal billing agreements
+  cannot be restored.
+- To reduce complexity, anyone paying via PayPal would be locked into using PayPal for later
+  subscriptions and users paying via Stripe would not be presented with any PayPal payment options.
+  This could frustrate users that want to switch, as they'd have to cancel, wait till the
+  subscription expires, and then subscribe again with PayPal.
+- Stripe subscription features can't be used, such as:
+  - Crediting a user for a month
+  - Coupons
+  - Consolidating multiple plans into a single subscription for aligned billing
+- Every additional subscription requires a full PayPal authorization flow.
+- PayPal occasionally misses delivery of webhooks, which requires us to build our own script
+  to periodically verify that PayPal subscriptions are still active and have had their latest
+  transaction processed successfully.
+- PayPal subscriptions can be suspended and resumed. This doesn't map to any equivalent in Stripe
+  subscriptions and would require custom handling in our code logic, entitlements, and UX.
+
+### Recurly
+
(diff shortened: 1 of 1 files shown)


docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Additional metadata in Stripe can start pushing us closer to metadata limits on objects.
+- Sensitive processing logic that runs PayPal transactions which we need to ensure work correctly.
+- We still could end up storing PayPal reference ID's in a separate db table to ensure we track
+  prior PayPal billing.
+
+### PayPal Subscriptions linked/tracked by FxA
+
+In this approach, we would create products/plans in PayPal, and link them back to the Stripe equivilant by updating Stripe metadata on the products/plans. FxA would also gain an additional database table that maps PayPal subscription id's to the Fx account. All the logic in FxA that looks up a users subscriptions would be updated to check the PayPal linking table to determine what subscriptions the user has active.
+
```
chenba, line 56: Small typo on "equivalent".

chenba: "ids" instead of "id's" but "user's" instead of "users".

And to nitpick, I find "what active subscriptions the user has" reads a little better.

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+PayPal would handle running the charges on a recurring basis per the PayPal plan that is setup. FxA would periodically check the subscription's via PayPal API to ensure they're still actively being paid (PayPal has fairly lenient handling on whether payments occur) on the schedule we want.
```
chenba, line 57: "subscriptions" and "occurred"?

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+
+- PayPal is responsible for charging the customer and managing the subscription lifecycle.
+- Can extend in a similar manner to track a correlating IAP for a user.
+
+Cons:
+
+- All the logic that determines active subscriptions and entitlements needs additional
+  logic/abstractions to query the database table and map active PayPal subscriptions to the
+  Stripe subscription.
+- Ability to easily cancel/re-subscribe on management page is lost, as canceled PayPal subscriptions
```
chenba, line 69: This seems to imply that we can easily resubscribe with the reference id.  Related question then: how and when does a reference id become invalid?

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+- To reduce complexity, anyone paying via PayPal would be locked into using PayPal for later
+  subscriptions and users paying via Stripe would not be presented with any PayPal payment options.
+  This could frustrate users that want to switch, as they'd have to cancel, wait till the
+  subscription expires, and then subscribe again with PayPal.
+- Stripe subscription features can't be used, such as:
+  - Crediting a user for a month
+  - Coupons
+  - Consolidating multiple plans into a single subscription for aligned billing
+- Every additional subscription requires a full PayPal authorization flow.
+- PayPal occasionally misses delivery of webhooks, which requires us to build our own script
```
chenba, line 80: Just curious: how do we know that PayPal isn't reliable with their webhooks?

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+  to periodically verify that PayPal subscriptions are still active and have had their latest
+  transaction processed successfully.
+- PayPal subscriptions can be suspended and resumed. This doesn't map to any equivilant in Stripe
```
chenba, line 83: Small "equivalent" typo here.

chenba (approved):
The plan sounds good to me!

docs/adr/0023-paypal-support.md
bbangert: It's more that since the subscription is invoice'd and tracked by Stripe, just as we do now, we can just mark the subscription as cancel_at_period_end. If they hit 're-subscribe', we just undo that mark. Cancelled PayPal subscriptions cannot be uncancelled in their system, a new subscription object entirely is needed, which muddies up when/how that needs to be done. In this case I wasn't referring to cancelling the reference ID.

To the related question, when a reference transaction is used, we get a flexible billing agreement that lets us run transactions without requiring further user approval. This billing agreement can be canceled by us, or by the user. The end-effect would be that our system would be in a similar state as when a user with a subscription has no valid cards on file. We would then have our subscription management page prompt the user to either put a new PayPal agreement on file, or a credit card for the next invoice cycle.

docs/adr/0023-paypal-support.md
bbangert: I unfortunately have personal experience running a separate system where this has occurred multiple times. The developer dashboard for PayPal is pretty lacking in this regard, as they don't seem to retry delivery either. Typical advice is to use PayPal API's to verify actions have occurred just to be sure.

biancadanforth (commented):
Thank you Ben for this document. I had some follow-up questions and clarifications mostly around the Stripe-driven approach. It sounds like the best option to me as well.

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+- Multiple product support works as-is.
+- Support representatives can see accurate payment history in Stripe for customers paying via PayPal.
+- Support can credit a PayPal customer, just as they can for a credit card customer.
+- Can extend in a similar manner to track a correlating IAP for a user.
+- Can handle switching payment method from credit card to PayPal reference ID.
+
+Cons:
+
+- Additional metadata in Stripe can start pushing us closer to metadata limits on objects.
+- Sensitive processing logic that runs PayPal transactions which we need to ensure work correctly.
```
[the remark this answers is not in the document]
bbangert: Mainly the fact that we want to ensure we charge the user, follow-up that the charge was processed successfully, and ensure our code doesn't double-charge or fail in some other way. So we need to come up with a scheme that is resilient to failure for ensuring a single charge occurs when it should.

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,102 @@
...
+- We still could end up storing PayPal reference ID's in a separate db table to ensure we track
+  prior PayPal billing.
```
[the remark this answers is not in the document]
bbangert: A reference transaction ID is not per subscription, its per customer as its a blanket permission to charge the PayPal account. If a user cancels a subscription, we will not cancel the reference transaction ID until the subscription periods have expired and there are no other subscriptions active. The table here is mainly for history, so that we know the user previously paid with PayPal, and can pull the cancelled billing agreement up with the id.

docs/adr/0023-paypal-support.md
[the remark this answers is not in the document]
say-yawn: I believe this was considered both pros and cons? Especially since this would allow us to have better understanding/definition of the invoice being paid and not relying on PayPal to retry on processing payments and handle late or denied payments.

But I suppose this is already presented on Pros:
```
- Support representatives can see accurate payment history in Stripe for customers paying via PayPal.
```

How about elaborating this line to be:
```suggestion
- Recreate Stripe's [automatic advancement and collection logic](https://stripe.com/docs/billing/invoices/overview#auto_advance) that runs PayPal transactions
```

docs/adr/0023-paypal-support.md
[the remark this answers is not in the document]
say-yawn: My understanding is that as long as the user is not cancelling the billing agreement as mentioned [here](https://developer.paypal.com/docs/archive/express-checkout/ec-set-up-reference-transactions/#cancel-a-billing-agreement) I think we can continue to use the reference ID to resubscribe or even create new subscriptions. I found this document helpful in understanding the relationship between the reference ID and the billing agreement. 

I am thinking in the DB we will only need the following columns: a PayPal reference ID, Stripe customer ID, and FXA user ID. Since having these information alone will allow us to get everything we need to validate, cancel, or create new subscriptions.

Also new to ADRs, but I think the specifics on columns for the table can be further narrowed after a decision is made?
@bbangert, can you share your thoughts here?

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Considered Options
+
+- Stripe-driven out-of-band invoice processing
+- PayPal subscriptions linked in FxA
+- Recurly
+
+## Decision Outcome
+
+Originally considered to be the 'fast-path', using PayPal subscriptions ended up being approximately the same engineering effort to implement as Stripe-driven invoice processing. Given additional user experience benefits of the Stripe-driven approach, it is now the recommend implementation path.
```
lmorchard, line 24: General ADR nit: Maybe lead with literally "Stripe-driven out-of-band invoice processing" to match up with the chosen option at a glance. Fine to elaborate after that, but best to very clearly highlight the decided option up front.

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,104 @@
...
+- Support representatives can see accurate payment history in Stripe for customers paying via PayPal.
+- Support can credit a PayPal customer, just as they can for a credit card customer.
+- Can extend in a similar manner to track a correlating IAP for a user.
+- Can handle switching payment method from credit card to PayPal reference ID, and vice versa.
+- PayPal authorization is retained so that additional products can be purchased/credited without requiring the user to return to the PayPal flow.
+- Cancel/resubscribe before the subscription expires works as-is.
+
+Cons:
+
+- Additional metadata in Stripe can start pushing us closer to metadata limits on objects.
```
lmorchard, line 50: FWIW, I think our [top-most Stripe metadata limit concerns are centered around products & plans](https://jira.mozilla.com/browse/FXA-2328). IIRC, we're not using a ton of per-customer metadata yet.

docs/adr/0023-paypal-support.md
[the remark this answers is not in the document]
lmorchard: ADRs are for overall abstract architectural concerns. If an exact DB table schema doesn't swing the decision one way or another, it doesn't really need to be nailed down in the ADR. 

But, unless I'm mistaken, I think the concern here is whether this weakens the notion that Stripe is a single source of subscription truth - because we'd store this payment-related data in more than one place (i.e. Stripe customer metadata *and/or* FxA DB table)? Not sure whether that's a huge concern, but it's worth noting.

lmorchard (approved):
Finally got around to looking this over, didn't really have much to add besides a nit or two

docs/adr/0023-paypal-support.md
```diff
@@ -0,0 +1,104 @@
...
+
+Originally considered to be the 'fast-path', using PayPal subscriptions ended up being approximately the same engineering effort to implement as Stripe-driven invoice processing. Given additional user experience benefits of the Stripe-driven approach, it is now the recommend implementation path.
+
+## Pros and Cons of the Options
+
+### Stripe-driven out-of-band invoice processing
+
+Stripe defaults subscriptions to billing automatically using a payment source on file with Stripe. Stripe also provides the option to have subscriptions that have a `collection_method` of `send_invoice`. In this model, Stripe can track subscriptions, and generate invoices that can be paid `out-of-band`. This allows Stripe to have an accurate record of charges, invoices, and subscriptions but how the invoice is considered 'paid' is up to FxA.
+
+PayPal has a feature called reference transactions that allow FxA to get a customers authorization to charge their PayPal account. This authorization allows for later payments to be charged against the PayPal account. We will track this reference ID in Stripe metadata for the customer and mark the subscription as `send_invoice`.
```
LZoog, line 32: Excuse the quick drive-by, but I read through this ADR out of curiosity and noticed [PayPal's docs](https://developer.paypal.com/docs/archive/express-checkout/ec-set-up-reference-transactions/) on reference transactions state that the feature is deprecated as of January 2017. Am I missing something here?

bbangert: They are apparently not really deprecated, if you ask nicely. [Recurly uses this method to run PayPal transactions ](https://docs.recurly.com/docs/paypal-payments#section-get-approved-to-use-reference-transactions)and notes here: 

> Note: The message in PayPal’s documentation saying that Reference Transactions are deprecated is inaccurate. In fact, Reference Transactions are still fully supported for new and existing Recurly merchants who want to add PayPal.

LZoog: 🤔 Well... that's interesting. If we're using this implementation in say, 3 years, that means we'll be using something marked as deprecated 7 years ago. Feels a bit funny, but perhaps they'll need to support it indefinitely.

clouserw (approved):

bbangert:
Ok, think I addressed the feedback and updated accordingly.

PR #6935, opened 2020-11-13
chore(tests): remove string helpers from helpersRemoteWrapped and applyRemote

chenba:
Because:
 - the error might be less confusing when someone tries to use one of those helpers from those returned by `helpersRemoteWrapped` or `applyRemote`

### DIFF
--- a/packages/fxa-content-server/tests/functional/lib/helpers.js
+++ b/packages/fxa-content-server/tests/functional/lib/helpers.js
@@ -2893,7 +2893,10 @@ module.exports = {
 // Export helpers methods in a form that can be easily used in
 // async/await. The usage of the helpers are the same except
 // they expect the last argument to be the intern remote object.
-const fnNames = Object.keys(module.exports);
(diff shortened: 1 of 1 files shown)


chenba:
> 
> 
> @chenba I haven't run into these specific error messages before. How can I test this change? Or do you happen to have a before and after paste of how this change updates the error output?

Dave run into a problem when he tried to use `createEmail` that was returned from `applyRemote`.  It's not something he committed I think.

With this PR, `createEmail` (and the other string helpers) simply won't be available from `applyRemote`.

Here's the error he received:
```
TypeError: callback.call is not a function
  at runMicrotasks @ anonymous
  at acc.<computed> @ tests/functional/lib/helpers.js:2910:21
  at Test.test @ tests/functional/settings_v2/external_links.js:78:19
  @ ../../src/lib/Test.ts:268:51
  at runMicrotasks @ anonymous
```

6a68 (approved):

Issue #6909, opened 2020-11-11
Ensure Beta settings button doesn't show up for non en-US locales


clouserw:
Maybe this is already done, but we should make sure.  /cc @meandavejustice @dannycoates

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2701)
┆Issue Number: FXA-2701

PR #6948, opened 2020-11-17
task(fxa-content-server): render beta link for en-US

meandavejustice:
## This pull request

- Checks for `en-US` before rendering the beta settings link in fxa-content-server

Closes: #6909

## Checklist


dannycoates (approved):

6a68 (changes requested):
Looks good, just one thing to tweak: all localized editions of firefox have navigator.languages that ends in `['en-US', 'en']`:

https://searchfox.org/l10n/search?q=%5Eintl.accept_languages%3D&path=intl.properties&case=false&regexp=true

so I think what you want here is just to check against `navigator.language`, which is the first value in this list.

6a68 (approved):

Issue #6936, opened 2020-11-13
[settings] Missing sign out buttons for OAuth RPs on Connected Services


data-sync-user:
STR (locally):

1. sign in
1. sign into 123Done
1. go to new settings

Expected: 123Done is listed and has a button for signing out

Actual: the sign out button is missing

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2713)
┆Issue Number: FXA-2713

data-sync-user:
➤ Dave Justice commented:

I don't think we ever had an issue for hooking up this button, we had one for creating a dummy button https://jira.mozilla.com/browse/FXA-1760, maybe we should throw this into settingsv2 version

vasilicatamas:
This issue is verified as fixed on Stage [Train 1.194.0] under Windows 10 64-bit, Android 9 and iOS 13.6. "Sign out" button is successfully displayed for 123Done.

![Screenshot_25](https://user-images.githubusercontent.com/18114210/100123183-e951a400-2e82-11eb-8fc9-ba4e4fdc1ddb.png)

vasilicatamas:
Checked again the [Figma](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A792) mock-ups and I'm not sure if the "Sign Out" button should be displayed. Filed #7002 for further discussions.

PR #6953, opened 2020-11-17
fix(settings): restore the sign out button for OAuth RPs

chenba:
Because:
 - connected OAuth RPs should be able to sign out

This commit:
 - restore the sign out for OAuth RPs

Closes: #6936 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.js
+++ b/packages/fxa-settings/src/components/ConnectedServices/MOCK_SERVICES.js
@@ -82,7 +82,7 @@ export const DESKTOP_SYNC_MOCKS = [
 const OAUTH_SERVICE_MOCKS = [
   {
     clientId: 'a8c528140153d1c6',
-    deviceId: 'dsalfjsajflkds',
+    deviceId: null,
     sessionTokenId: null,
     refreshTokenId:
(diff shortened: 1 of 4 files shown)


6a68 (approved):
LGTM, although looking again at [the UX specs in Figma](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A1), it appears the intention was to provide a link to oauth RPs instead of provide a revoke button:

Screen Shot 2020-11-17 at 9 49 04 AM

A related question for settings v2: if `canSignOut` is true for all types of RPs, maybe the `canSignOut` property isn't useful at all? Do we have any RPs that we can't sign out?

chenba:
> 
> 
> LGTM, although looking again at [the UX specs in Figma](https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A1), it appears the intention was to provide a link to oauth RPs instead of provide a revoke button:

Hmm is there another way for a user to revoke a RP's access without this?

chenba:
All right, unless a user can revoke a RP's access in some other way, we're keeping the sign out button on the connected services list.  We must offer a way for the users to revoke access.

> 
> A related question for settings v2: if `canSignOut` is true for all types of RPs, maybe the `canSignOut` property isn't useful at all? Do we have any RPs that we can't sign out?

I removed it.

Issue #6876, opened 2020-11-09
/graphql allows queries with high read and write amplification


data-sync-user:
The GraphQL supports a number of features that allow [medium-risk authenticated denial of service requests](https://r2c.dev/blog/2020/understanding-and-preventing-dos-in-web-apps/) that consume a large amount of server resources for relatively small queries fanning out requests to multiple downstream services or data stores.

The proof of concept in [https://gist.github.com/g-k/5c281b7bac31aa53eab6fd8cc4a21d11](https://gist.github.com/g-k/5c281b7bac31aa53eab6fd8cc4a21d11) uses [aliases](https://graphql.org/learn/queries/#aliases) and [fragments](https://graphql.org/learn/queries/#fragments) to achieve a read amplification factor of 33x and consumed 57s of server time. Adding more devices, subscriptions could further increase the read amplification factor.

Mutations with multiple writes to consume more CPU, network, or database resources are similarly possible e.g.

curl 'https://fxa-graphql-api.stage.mozaws.net/graphql' ~~H 'Content~~Type: application/json' ~~H 'authorization:  $STAGE_FXA_SESSION'~~ ~~data~~binary '{"query":"mutation addTotps($input: CreateTotpInput!) {\n  f1:createTotp(input: $input)

{\n    clientMutationId\n    qrCodeUrl\n    secret\n\t  recoveryCodes\n  }\n  f2: createTotp(input: $input) {n    clientMutationIdn    qrCodeUrln    secretnt  recoveryCodesn  }

\n}\n","variables":{"input":{}}}' --compressed

**Recommendation(s):**

* [Enforce a hard timeout on queries](https://www.howtographql.com/advanced/4-security/)

* Where possible, [add caching](https://graphql.org/learn/caching/) for reads
* Consider adding [query request size limits or query allowlists](https://www.apollographql.com/blog/securing-your-graphql-api-from-malicious-queries-16130a324a6b/)
* If cyclical relations are added to the GQL API, limit query depth

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2683)
┆Issue Number: FXA-2683

data-sync-user:
➤ Jared Hirsch commented:

It may be useful to break this down into sub-bugs based on the recommended tasks in the description, or based on the recommendations per individual endpoint exposed by GQL.

clouserw:
+cc @bbangert

data-sync-user:
➤ Danny Coates commented:

[...]

PR #6962, opened 2020-11-18
feat(gql-api): limit query complexity

dannycoates:
This is a stop-gap implementation of query complexity handling for #6876 with a hard coded limit of 100. The most complex query for new settings is currently 38. This should prevent the huge amplification mentioned in the issue, but we should revisit and have a more thorough design before leaving beta.

I've tested this locally and it indeed blocks queries over the complexity limit. Time permitting I'll add a test case.


chenba (approved):
🎉

PR #6970, opened 2020-11-19
feat(graphql-api): convert to NestJS

### LINKED ISSUE #6263: Convert gql-api to use NestJS
Convert the graphql-api to use NestJS and its patterns.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2458)
┆Issue Number: FXA-2458

bbangert:
Because:

* We want to use NestJS for new apps, and graphql-api is still small
  enough to feasibly convert to NestJS.

This commit:

* Converts graphql-api to a NestJS app.

Closes #6263

## Checklist


bbangert:
Opening this for review now, still need to fix a settings test to account for the change in the authorization header but otherwise this is ready.

jodyheavener (commented):
Still going through, but an initial thought: we've added `class-validator` as a dep, it would be nice to use them to decorate DTOs. I can see some spots where `@IsEmail`, `@IsUrl`, and `@Length` might be beneficial. But since this is more of a parity PR this could also be a nice follow-up.

packages/fxa-graphql-api/src/backend/profile-client.service.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+import { Test, TestingModule } from '@nestjs/testing';
+import nock from 'nock';
+
+import { AuthClientService } from './auth-client.service';
+import { ProfileClientService } from './profile-client.service';
+import { ConfigService } from '@nestjs/config';
+
+describe('ProfileClientService', () => {
+  let service: ProfileClientService;
+  let authClient: any;
```
jodyheavener, line 14: It looks like you could eliminate the need to mock `createOAuthToken` each test by just doing

```tsx
  const authClient = {
    createOAuthToken: jest.fn().mockResolvedValue({ access_token: 'test' }),
  };
```

packages/fxa-graphql-api/src/backend/profile-client.service.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
+
+  it('updates display name', async () => {
+    authClient.createOAuthToken = jest
+      .fn()
+      .mockResolvedValue({ access_token: 'test' });
+    (service as any).request = {
```
jodyheavener, line 51: Rather than creating a request alias for testing and using `any` here, would it be cleaner to just mock superagent? Something like:

```ts
(jest.spyOn(superagent, "post") as jest.Mock).mockReturnValueOnce({
  send: jest.fn().mockReturnValue({
    set: jest.fn().mockResolvedValue({ text: "{}" }),
  }),
});
```

packages/fxa-graphql-api/src/database/database.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  public authKnex: Knex;
+  public profileKnex: Knex;
+
+  constructor(configService: ConfigService<AppConfig>) {
+    const dbConfig = configService.get('database') as AppConfig['database'];
+    this.authKnex = setupAuthDatabase(dbConfig.mysql.auth);
+    this.profileKnex = setupProfileDatabase(dbConfig.mysql.profile);
+  }
+
+  async dbHealthCheck(): Promise<Record<string, any>> {
```
jodyheavener, line 23: Nit: could get a little more specific here with `Promise<{ db: { status: 'ok' | 'error' } }>`

packages/fxa-graphql-api/src/database/database.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    let status = 'ok';
+    try {
+      await Account.query().limit(1);
+    } catch (err) {
```
jodyheavener, line 27: Should this error get kicked over to Sentry?

packages/fxa-graphql-api/src/decorators.ts
```diff
@@ -0,0 +1,36 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { GqlExecutionContext } from '@nestjs/graphql';
+import { SessionTokenResult } from './auth/session-token.strategy';
+
+/**
+ * Extracts the token from an authenticated user for a GraphQL request.
+ */
+export const GqlSessionToken = createParamDecorator(
+  (data: unknown, context: ExecutionContext): string => {
```
jodyheavener, line 12: Nit: I don't know if it's convention or something I just picked up along the way, but `data: unknown` could be `_` if we don't intend on using it.

packages/fxa-graphql-api/src/database/database.service.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [DatabaseService, MockConfig],
+    }).compile();
+
+    service = module.get<DatabaseService>(DatabaseService);
+  });
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
```
jodyheavener, line 40: could use a `dbHealthCheck` test in here

packages/fxa-graphql-api/src/auth/session-token.strategy.ts
```diff
@@ -0,0 +1,43 @@
...
+  session: ThenArg<ReturnType<AuthClient['sessionStatus']>>;
+}
+
+@Injectable()
+export class SessionTokenStrategy extends PassportStrategy(Strategy) {
+  constructor(@Inject(AuthClientService) private authAPI: AuthClient) {
+    super({});
+  }
+
+  async validate(token: string): Promise<SessionTokenResult> {
```
jodyheavener, line 26: could use a test for this fella

jodyheavener (commented):
Few comments here and there, mostly just nitpicking. I don't think my comments on the tests are all that important. And it works as expected with Settings! 🎉

packages/fxa-graphql-api/src/auth/session-token.strategy.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Inject, Injectable, UnauthorizedException } from '@nestjs/common';
+import { PassportStrategy } from '@nestjs/passport';
+import AuthClient from 'fxa-auth-client';
+import { ExtendedError } from 'fxa-shared/nestjs/error';
+import { Strategy } from 'passport-http-bearer';
+
+import { AuthClientService } from '../backend/auth-client.service';
+
+// Unwrap the type from a promise
+type ThenArg<T> = T extends PromiseLike<infer U> ? U : T;
```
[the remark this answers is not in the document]
bbangert: Indeed, I was going to take a look at whether we could bump up to TS 4.1, or whether everything will break. 😀

jodyheavener:
Oh also, when I run `yarn test` the file `src/schema.gql` gets generated. If this is something we want to add to the repo let's do it! Otherwise we might want to ignore since it'll almost def get accidentally committed.

packages/fxa-graphql-api/src/database/database.service.ts
bbangert: I figured we might be adding more in the future, and having it narrowed to all possible values will get tedious vs. the overall shape of the object.

packages/fxa-graphql-api/src/database/database.service.ts
bbangert: It didn't before, this part hasn't changed in the NestJS transition. We catch them in other places to always return a value so this was following existing code. We could file a follow-up issue if we'd like to catch these as well.

bbangert:
@jodyheavener okey, think I got everything you mentioned.

jodyheavener (approved):
🤠

PR #6987, opened 2020-11-20
Restore mobile sync device counts by firing a flow complete event for mobile browsers

### LINKED ISSUE #5143: Amplitude user properties `sync_active_devices_*` doesn't count Fx iOS anymore and Fenix, and will break when sync moves to oauth

6a68:
When an oauth-based mobile sync device signs in, we're not seeing the corresponding `fxa_login - complete` metrics event that should mark the login, an event which includes the updated count of active sync devices per day for that user. Building on the changes in #6383 and #6581, I think we need to fire the 'account.signed' event when mobile browsers request the oldsync token, which will trigger the flow complete signal that fires the `fxa_login - complete` event (see #6347).

There's also a change here to avoid firing the 'account.signed' event if a 'profile' scoped token is requested, in a first step towards reducing the number of inaccurate device counts recorded in our metrics (see #6578), but I'm thinking we might want to remove that commit, since it's not strictly needed to address #5143 and introduces an unnecessary additional variable.

Refs #6578, fixes #5143.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth/index.js
+++ b/packages/fxa-auth-server/lib/routes/oauth/index.js
@@ -317,10 +317,6 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
           // (ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1632635), where
           // we need the 'service' event property to distinguish between sync
           // and browser.
-          //
-          // The service for this event gets special-cased to 'sync' so that it
-          // matches its pre-oauth `/certificate/sign` event.  ref:
-          // https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
           if (
             scopeSet.contains(OAUTH_SCOPE_OLD_SYNC) &&
             // Desktop requests a profile scope token before adding the device
@@ -328,13 +324,24 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
             // emit this event if the request is for an oldsync scope, not a
             // profile scope. See #6578 for details on the order of API calls
             // made by both desktop and fenix.
-            !scopeSet.contains('profile') &&
-            config.oauth.oldSyncClientIds.includes(request.payload.client_id)
+            !scopeSet.contains('profile')
           ) {
+            // For desktop, the 'service' parameter for this event gets
+            // special-cased to 'sync' so that it matches its pre-oauth
+            // `/certificate/sign` event.
+            // ref: https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
+            // Otherwise, for mobile browsers, just use the existing client ID
+            // to service name mapping used in the metrics code (see the
+            // OAUTH_CLIENT_IDS config value). #5143
+            const service = config.oauth.oldSyncClientIds.includes(
+              request.payload.client_id
+            )
+              ? 'sync'
+              : request.payload.client_id;
             await request.emitMetricsEvent('account.signed', {
               uid: uid,
               device_id: sessionToken.deviceId,
-              service: 'sync',
+              service,
             });
           }
         } catch (ex) {}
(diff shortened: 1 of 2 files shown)


rfk:
> I think we need to fire the 'account.signed' event when mobile browsers request the oldsync token, which will trigger
> the flow complete signal that fires the fxa_login - complete event

This change the the reasoning make sense to me FWIW. It would probably be better to re-work the "flow complete" logic to trigger directly on creation of a token rather than having to fake this even, but that's a much bigger change and this smaller change seems worthwhile.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -317,18 +317,28 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
+            // oldsync scope, but not a profile scope. #6578
+            !scopeSet.contains('profile')
           ) {
+            // For desktop, the 'service' parameter for this event gets
+            // special-cased to 'sync' so that it matches its pre-oauth
+            // `/certificate/sign` event.
+            // ref: https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
+            // Otherwise, for mobile browsers, just use the existing client ID
+            // to service name mapping used in the metrics code (see the
+            // OAUTH_CLIENT_IDS config value). #5143
```
chenba, line 332: 👍

6a68:
Thanks for the feedback! Tests updated, ready for final review.

packages/fxa-auth-server/lib/routes/oauth/index.js
```diff
@@ -317,18 +317,28 @@ module.exports = (log, config, oauthdb, db, mailer, devices) => {
...
           // and browser.
-          //
-          // The service for this event gets special-cased to 'sync' so that it
-          // matches its pre-oauth `/certificate/sign` event.  ref:
-          // https://github.com/mozilla/fxa/pull/6581#issuecomment-702248031
           if (
             scopeSet.contains(OAUTH_SCOPE_OLD_SYNC) &&
-            config.oauth.oldSyncClientIds.includes(request.payload.client_id)
+            // To hack around timing issues, only emit this event if it's an
+            // oldsync scope, but not a profile scope. #6578
```
rfk, line 323: I infer from this, that you've observed clients requesting  token with both `oldsync` and `profile` scope. Which clients, and do you know under what circumstances they do so?

In general clients should not be requesting tokens with both of those scopes, for security reasons (e.g. we don't want the profile-server to accidentally be able to read your encrypted sync data, because you used an access_token with additional scopes on it). The only circumstance I can think of where this would be OK, is when initially completing the OAuth login flow by redeeming an authorization code, which returns an access token with all the requested scopes.

rfk (approved):
👍

packages/fxa-auth-server/lib/routes/oauth/index.js
6a68: Great question! Right now, the only spot I've observed both scopes requested together is the auth code exchange step.

Maybe this comment could be reworded. The intent is to document the fact that both fenix and desktop update the account's list of devices before they request a token with the oldsync scope. However, desktop requests a profile token before adding the device to the account, so the right time to increment device usage for all sync clients is when an oldsync token is requested. (This timing assumption seems brittle to me, but also unlikely to change significantly.)

6a68: @rfk I've updated the code comment. Does it seem clearer?

6a68:
@rfk I've updated the code comment to try to clarify the context behind the oldsync vs profile token timing. Does it seem clearer?

rfk:
Thanks, that does help clarify 👍

PR #6990, opened 2020-11-20
fix(content-server): fix test-ci script and server tests

chenba:
Because:
 - server tests have been failing in CI without us knowing since they
   "passed"

This commit:
 - update the command for the server and pairing tests so that exit on
   error will work as intended
 - fix the failing tests, mainly from minor code changes or breaking
   dependency API changes

The main problem was that the test-ci script was calling a function that
was meant to run functional tests to run the server and pairing tests.
The command in that function included retry logic that was for the more
"flaky" functional tests.  When retry is attempted, the argument passed
to --grep errored out, but this DOES NOT EXIT WITH A NON-ZERO STATUS
APPARENTLY since no tests actually failed...?


packages/fxa-content-server/package.json
```diff
@@ -150,6 +149,7 @@
     "eslint": "7.7.0",
     "eslint-plugin-fxa": "^2.0.2",
     "firefox-profile": "4.0.0",
+    "got": "11.8.0",
```
chenba, line 152: Moved to dev dependencies since it's only used in tests.

dannycoates (approved):
Thanks @chenba 👍

Issue #5896, opened 2020-07-08
[content server] Apply the MX record validation feature flag to the validate-email-domain endpoint


data-sync-user:
There's a feature flag for the MX record validation feature, but currently it only affects the frontend.  Let's have the endpoint immediately return a 200 response if the feature flag is set to off.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2252)
┆Issue Number: FXA-2252

karthikshetty03:
I would like to work on this issue.
Can I know more about this issue as well as the filename where the code needs to be changed ?

chenba:
@karthikshetty03 The function that process validation request starts at https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/server/lib/routes/validate-email-domain.js#L69.  And this issue asking that if the configuration https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/server/lib/configuration.js#L367 is false, that function would then return a 200 response.

Thanks for working on this!

chenba:
@Cronus1007 Good question.  Let return a result of 'skip' along with the 200.  The list of available results are declared at https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/server/lib/routes/validate-email-domain.js#L16.  And when the frontend get that result, it just resolves the promise.

PR #7037, opened 2020-12-01
chore(fxa-content-server): appling the MX record validation feature flag to the validate-email-domain endpoint

Cronus1007:
## Because

-

## This pull request

-

Closes: #5896 

## Checklist

## Screenshots (Optional)

@chenba I think I dont have the required permission to add the reviewers section.

### DIFF
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -867,7 +867,6 @@ if (conf.get('ecosystem_anon_id.keys_file')) {
     );
   }
 }
-
 // TODO: convict 6+ doesn't like the schema definition we've got
 // for `scopedKeys.validation`. It doesn't cause runtime problems
 // but fails validation. We should change it to pass validation.
@@ -878,3 +877,4 @@ if (conf.get('ecosystem_anon_id.keys_file')) {
 
 // validate the configuration based on the above specification
 // conf.validate(options);
+module.exports.conf = conf;
--- a/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
+++ b/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
@@ -17,7 +17,6 @@ const results = ['MX', 'A', 'none'].reduce((accumulator, val) => {
   accumulator[val] = val;
   return accumulator;
 }, {});
-
 const NotFoundErrorCodes = [dns.NODATA, dns.NOTFOUND];
 const WrappedErrorCodes = [
   dns.FORMERR,
@@ -57,7 +56,7 @@ const tryResolveWith = (resolveFunc) => async (domain) => {
   }
 };
 
-module.exports = function () {
+module.exports = function (config) {
   return {
     method: 'get',
     path: '/validate-email-domain',
@@ -68,13 +67,17 @@ module.exports = function () {
     },
     process: async function (req, res, next) {
       const { domain } = req.query;
-
+      const value = config.get('mxRecordValidation');
+      console.log(value);
       try {
         if (await tryResolveWith(resolver.resolveMx.bind(resolver))(domain)) {
           return res.json({ result: results.MX });
         }
         if (await tryResolveWith(resolver.resolve4.bind(resolver))(domain)) {
           return res.json({ result: results.A });
+        }
+        if (!value) {
+          return res.sendStatus(200);
         }
         return res.json({ result: results.none });
       } catch (err) {


packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -878,3 +877,4 @@ if (conf.get('ecosystem_anon_id.keys_file')) {
 
 // validate the configuration based on the above specification
 // conf.validate(options);
+module.exports.conf = conf;
```
chenba, line 880: No need for this.

You can find examples of how the configs are passed into routes in `server/lib/routes.js`.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -68,14 +67,18 @@ module.exports = function () {
     },
     process: async function (req, res, next) {
       const { domain } = req.query;
-
+      const value = config.get('mxRecordValidation');
+      console.log(value);
```
chenba, line 71: Please remove the debug logging.  Also, let's name the variable with something less generic.  Maybe `shouldValidate` or something.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -68,14 +67,18 @@ module.exports = function () {
...
       try {
         if (await tryResolveWith(resolver.resolveMx.bind(resolver))(domain)) {
           return res.json({ result: results.MX });
         }
         if (await tryResolveWith(resolver.resolve4.bind(resolver))(domain)) {
           return res.json({ result: results.A });
         }
+        if (!value) {
+          return res.sendStatus(200);
+        }
```
chenba, line 81: We need to do this first.  And it can be outside of this `try`.

chenba (changes requested):
Please see inline comments and my comment on the issue regarding the response payload.  Thanks!

AUTHORS
```diff
@@ -167,5 +167,5 @@ Xavier RENE-CORAIL
 Yehuda B
 Yehuda Bar-Nir
 YFdyh000
-Zach Carter
 Zachary Carter
+Zach Carter
```
chenba, line 171: Why are there changes in this file?

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -13,11 +13,10 @@ const joi = require('joi');
 const dns = require('dns');
 const VError = require('verror');
 const resolver = new dns.promises.Resolver();
-const results = ['MX', 'A', 'none'].reduce((accumulator, val) => {
+const results = ['MX', 'A', 'none', 'skip'].reduce((accumulator, val) => {
   accumulator[val] = val;
   return accumulator;
 }, {});
-
```
chenba, line 20: @Cronus1007 just not a big fan of whitespace? 😀

chenba (changes requested):
The changes in the route handler look good.  But the route needs to be updated to pass the configs; see my comment here: https://github.com/mozilla/fxa/pull/7037#discussion_r533593653.

The change also need a test added to https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/tests/server/routes/validate-email-domain.js

And we need update the front to handle the new 'skip' response: https://github.com/mozilla/fxa/issues/5896#issuecomment-736697187

AUTHORS
Cronus1007: Actually I ran the command `yarn run authors` So this happened.

Cronus1007:
@chenba  Okk I will take a look :)

Cronus1007:
@chenba Can you please refer the codebase corresponding to the `client` for the `fxa-content-server` package to make the desired changes in the front.

chenba:
> 
> 
> @chenba Can you please refer the codebase corresponding to the `client` for the `fxa-content-server` package to make the desired changes in the front.

The frontend code is here: https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/lib/email-domain-validator.js.  And the tests are at https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js

Cronus1007:
@chenba Added the required `tests` as well as the made the required changes in the `front`

AUTHORS
```diff
@@ -21,18 +21,18 @@ bini11
 Brandon Ebersohl
 Brian Warner
 byrw
+César Carruitero
```
chenba, line 24: This file is updated when we tag a new release.  There's no need to include it here.

Cronus1007: @chenba  Plzzz tell the script to undo this file.

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -119,6 +119,11 @@ const checkEmailDomain = ($element, view) => {
           view.logFlowEvent('email-domain-validation.warn');
           return reject(REJECTION_CODE);
         }
+        if (result === 'skip') {
```
chenba, line 122: nit: for consistency sake, let's add an empty line above this one.

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -119,6 +119,11 @@ const checkEmailDomain = ($element, view) => {
...
+          view.logEvent('emailDomainValidation.skip');
+          view.logFlowEvent('email-domain-validation.skip');
```
chenba, line 124: Let's remove these event calls.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -68,7 +67,10 @@ module.exports = function () {
     },
     process: async function (req, res, next) {
       const { domain } = req.query;
-
+      const shouldValidate = config.get('mxRecordValidation');
```
chenba, line 70: The argument should be `mxRecordValidation.enabled` here.

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -134,7 +134,21 @@ describe('lib/email-domain-validator', () => {
...
+      assert.isTrue(view.logEvent.calledWith('emailDomainValidation.skip'));
+      assert.isTrue(
+        view.logFlowEvent.calledWith('email-domain-validation.skipped')
+      );
+      assert.isTrue(
+        view.logFlowEvent.calledWith('email-domain-validation.skip')
+      );
+      xhr.ajax.restore();
+    });
+  });
```
chenba, line 151: Since we do nothing but `resolve()`, let's add a `catch` and call `assert.fail` in it, since it should not have rejected.

chenba (changes requested):
Thanks for the frontend changes.  Would like some minor edits there.  Also I still don't see a test for the backend changes; see second line of https://github.com/mozilla/fxa/pull/7037#pullrequestreview-542255382.

Cronus1007:
> Thanks for the frontend changes. Would like some minor edits there. Also I still don't see a test for the backend changes; see second line of [#7037 (review)](https://github.com/mozilla/fxa/pull/7037#pullrequestreview-542255382).

@chenba  Is there a need to catch the error if it occurs in the routes.like 404 and other error codes.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,24 @@ registerSuite('routes/validate-email-domain', {
...
+        const req = { query: { domain: 'skip.abc.xyz' } };
+        const res = { json: sinon.stub() };
+        const route = validateEmailDomainRoute();
+        return route
+          .process(req, res)
+          .then(() => {
+            assert.isTrue(res.json.calledOnceWith({ result: 'skip' }));
+          })
+          .catch((err) => {
+            assert.fail(res.json.calledOnceWith(new Error(err.message)));
```
Cronus1007, line 75: @chenba  I was having a thought that we can return error message if the `route` fails due to internal server error so we ca return err.message

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,24 @@ registerSuite('routes/validate-email-domain', {
           assert.isTrue(res.json.calledOnceWith({ result: 'MX' }));
         });
       },
-
+      'responds with {result: "skip"} when there is a skip record': () => {
+        const dns = {
+          ...ERROR_CODES,
```
Cronus1007, line 62: @chenba  I am little bit confused with line 62 and 75

chenba:
> 
> 
> @chenba Is there a need to catch the error if it occurs in the routes.like 404 and other error codes.

Can you explain a little more?  What is wrong with the way the request is currently handled?

Cronus1007:
I mean that if some sort of `internal server error` occured so we should display an error message

Cronus1007:
Just a thought for every `route` since it needs to be consistent and this has a very `low` possibility but can occur.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
chenba: `dns` here is used to mock the Nodejs dns module in validate-email-domain.js.  You don't really need it for this test since validate-email-domain.js shouldn't be using `dns` to resolve a domain.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
chenba: You don't need a `catch` since a rejection will be a test failure.  But if you want to keep it, pass a message about the failure into `assert.fail`.  The code you added did not throw any errors, so there isn't a particular error to check here.

What sort of "internal server error" do you think we should handle?

chenba:
> 
> 
> I mean that if some sort of `internal server error` occured so we should display an error message

No, we don not want to display errors here.  There isn't anything for the user to do to "fix the error", and we won't block the user anyways.

chenba:
(sorry accidentally closed the PR 😄)

Cronus1007:
no issues :)

Cronus1007:
Okk so is the pr ready for merge .

Cronus1007:
@chenba If everything is ok so shall I force push all the commits and convert into one so that the PR would be merged and #5896  could be closed. :)

Cronus1007:
@chenba I am soon sending a new PR since i pushed some wrong commits

PR #7038, opened 2020-12-01
fix: update nvmrc to match needed node version

bbangert:
Because:

* It's useful for scripts that check nvmrc on directory change to see
  when the projects needed node version has changed.

This commit:

* Updates nvmrc to precisely match the needed node that is also checked
  when starting up all the services.

## Checklist


.nvmrc
```diff
@@ -1 +1 @@
-14
+14.15.1
```
Cronus1007, line 1: @bbangert  Just a thought that whenever we run `nvm install --lts` it returns the same version. So is there a need to use this file. We can just update the readme.

bbangert: I mentioned this in a different PR, but the reason for a `.nvmrc` file is to provide an additional sanity check that automated tooling can use. For example, I use a shell hook that when I change into a directory, if there's a `.nvmrc` file, it uses `nvm` to switch my active Node. I work on multiple node projects, and using the right version is important, its easy to use the wrong version without this check when switching between projects.

Using just `14` did not catch the difference, so I was confused when I ran `yarn start` today, and it did not start up, because I was on `14.15.0`.

bbangert: https://stackoverflow.com/questions/23556330/run-nvm-use-automatically-every-time-theres-a-nvmrc-file-on-the-directory has some information on how to setup a shell to do this automatically.

Cronus1007: @bbangert  Actually I totally agree with this point of yours but as per my observations while setting up the project for a newbie the main `README`  file lacks a lot and while doing `npm start` it only gives the errors.There is a great need to enhance the documentation for the complete repo.I think we should merge this PR since the change is necessary.

Cronus1007 (approved):

dannycoates (approved):

PR #7040, opened 2020-12-01
feat(gql-api): add cors

### LINKED ISSUE #6875: /graphql allows CORS requests from any origin
/graphql allows CORS requests from any origin:

curl ~~IX OPTIONS https://fxa~~graphql-api.stage.mozaws.net/graphql
...
Access-Control-Allow-Origin: *

Requests fail as unauthenticated, but it’d be best to not allow them in the first place.

Return the Access-Control-Allow-Origin header with the environment appropriate domain (e.g. https://accounts.firefox.com or https://accounts.stage.mozaws.net) e.g. “Access-Control-Allow-Origin: https://accounts.firefox.com”

NB: Use an application config value and not a user-provided request header like “Host”

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2682)
┆Issue Number: FXA-2682

bbangert:
Because:

* We want to enforce CORS origin on requests.

This commit:

* Adds a corsOrigin configuration value used the same as in the auth
  server.

Closes #6875

## Checklist

### DIFF
--- a/packages/fxa-graphql-api/src/gql/gql.module.ts
+++ b/packages/fxa-graphql-api/src/gql/gql.module.ts
@@ -32,6 +32,8 @@ export const GraphQLConfigFactory = async (
   playground: configService.get<string>('env') !== 'production',
   autoSchemaFile: join(path.dirname(__dirname), './schema.gql'),
   context: ({ req }) => ({ req }),
+  // Disabling cors here allows the cors middleware from NestJS to be applied
+  cors: false,
   uploads: false,
   validationRules: [
     queryComplexity({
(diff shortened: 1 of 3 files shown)


packages/fxa-graphql-api/src/gql/gql.module.ts
```diff
@@ -32,6 +32,7 @@ export const GraphQLConfigFactory = async (
   playground: configService.get<string>('env') !== 'production',
   autoSchemaFile: join(path.dirname(__dirname), './schema.gql'),
   context: ({ req }) => ({ req }),
+  cors: false,
```
dannycoates, line 35: maybe a comment about cors being enabled at the app level would be helpful

dannycoates (approved):
lgtm 👍

PR #7042, opened 2020-12-01
tests(Settings_V2): add tests to enable and disable 2FA

ashrivastava-qa:
`Added tests for Settings redesign to`
`- Enable the 2FA`
`- Disable the 2FA`

## This pull request
`- Closes task FXA-1565: Add/Revoke 2FA`

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

While writing this test I found an issue FXA-2765.
At line 103 and 104 I have made a comment about it as well. Once the bug fixes the tests can be fixed as well.
@mozilla/fxa-devs please review! Thanks!


chenba:
@ashrivastava-qa please rebase on latest `main`.  That should fix the failing test.

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
```diff
@@ -2,19 +2,24 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const intern = require('intern').default;
```
chenba, line 5: This isn't necessary.  Why was this added?  I just want to make sure that I didn't miss or misunderstand something.

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
```diff
@@ -2,19 +2,24 @@
...
 const FunctionalHelpers = require('../../lib/helpers');
 
-const { createEmail } = FunctionalHelpers;
+const { createEmail, generateTotpCode } = FunctionalHelpers;
 const config = intern._config;
 const EMAIL_FIRST = config.fxaContentRoot;
 const SETTINGS_V2_URL = config.fxaSettingsV2Root;
 const password = 'passwordzxcv';
+let secret;
+let securityCode;
```
chenba, line 15: These variable should not be scoped to the module.  They can be in the functions where they are used.

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
```diff
@@ -33,6 +38,62 @@ async function navigateToSettingsV2(remote) {
...
+    remote
+  );
+  await type(
+    selectors.SETTINGS_V2.SECURITY.TFA.RECOVERY_KEY_INPUT,
+    recoveryKey,
+    remote
+  );
+  await click(selectors.SETTINGS_V2.SECURITY.TFA.SUBMIT_RECOVERY_KEY, remote);
+}
+
```
chenba, line 94: I think these two new functions shouldn't be in the helpers as they are very 2FA specific.

packages/fxa-content-server/tests/functional/settings_v2/two_step_auth.js
```diff
@@ -0,0 +1,144 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const intern = require('intern').default;
```
chenba, line 7: Same as above: this is not necessary.

packages/fxa-content-server/tests/functional/settings_v2/two_step_auth.js
```diff
@@ -0,0 +1,144 @@
...
+  click,
+  testElementExists,
+  type,
+  visibleByQSA,
+  testElementTextEquals,
+} = FunctionalHelpers.helpersRemoteWrapped;
+
+describe('two step auth', () => {
+  beforeEach(async ({ remote }) => {
+    await navigateToSettingsV2(remote);
```
chenba, line 29: In addition to navigating, this also creates a new FxA user and sign in.  It's a bit wasteful in a `beforeEach` here, since you can use the same user to test enable and disable.  intern has a `before`, so you could create a user and nav once with that.

chenba (changes requested):
The tests work, but do have a few little things that I'd like to see fixed.  Thanks!

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
ashrivastava-qa: Yeah but then if I am writing separate tests for enable and disable, I will have to duplicate the same steps to first enable and then disable. Or if I can just write one whole test that first enables and then disables this could be removed. But I thought breaking functionalities into different tests is better. What do you suggest?

packages/fxa-content-server/tests/functional/settings_v2/two_step_auth.js
ashrivastava-qa: Yes I was trying with before but when running the disable test somehow the same user and email was giving me error while logging in.  I struggled a lot with it and hence I wrote it in beforeEach using the helper functions so that I don't have to repeat the steps. I wrote the helper functions because I saw something similiar(not totally) with the existing helper functions confirmRecoveryCode, confirmTotpCode.

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
chenba: Well, you can still keep them as functions, but in the test suite, not here.

chenba: Oh, yes, it's fine to  have just one test for all the steps as well.

packages/fxa-content-server/tests/functional/settings_v2/two_step_auth.js
chenba: > 
> 
> Yes I was trying with before but when running the disable test somehow the same user and email was giving me error while logging in. 

Why do you need to log in again?  The user's already logged in by the end of the enable test correct?  I think you just need to go back to the main settings page.

ashrivastava-qa: okay! Sure. Let me work on this to incorporate your comments and I'll make the changes. Thanks for the thorough review! :)

packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
```diff
@@ -33,6 +38,62 @@ async function navigateToSettingsV2(remote) {
...
+  );
+  await click(selectors.SETTINGS_V2.SECURITY.TFA.SHOW_CODE_LINK, remote);
+
+  // store the secret key to use later
+  await testElementExists(
+    selectors.SETTINGS_V2.SECURITY.TFA.SECRET_CODE_MODAL,
+    remote
+  );
+  secret = await remote
+    .findByCssSelector(selectors.SETTINGS_V2.SECURITY.TFA.SECRET_CODE_TEXT)
```
vbudhram, line 55: nit: I'm surprised we didn't have a helper method for "getElementText", welcome to add if you think its useful.

packages/fxa-content-server/tests/functional/settings_v2/two_step_auth.js
vbudhram: Having separate calls to create user and then navigate to new settings makes the test slightly more readable too.

ashrivastava-qa: @vbudhram Is it okay for now to keep just that one function? if I split it into two now, I'll have to make changes to a few different tests as well and I was thinking to do it later after I am done with the 2FA tests.

ashrivastava-qa:
Closing this PR as another PR has been created for theses tests and comments. https://github.com/mozilla/fxa/pull/7093

PR #7055, opened 2020-12-02
refactor(oauth): move /verify route

dannycoates:
## Because

- we want oauth routes better integrated into auth-server

## This pull request

- establishes the `lib/routes/oauth/index.js` file as the new "entrypoint" for oauth routes
- moves the `/verify` route
- renames and deprecates the "proxied" oauth route file


chenba (approved):
👍

Issue #7054, opened 2020-12-02
move /client/:id route


data-sync-user:
Following the pattern set here [https://github.com/mozilla/fxa/commit/67fde4359a0d44673f932c617e98db0482502ea3](https://github.com/mozilla/fxa/commit/67fde4359a0d44673f932c617e98db0482502ea3)

move the route and use auth-server logging and metrics and add it to lib/routes/oauth/index.js. leave the old implementation in place

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2766)
┆Issue Number: FXA-2766

PR #7060, opened 2020-12-03
refactor(oauth): move /client/{client_id} route

chenba:
Because:
 - we want less indirection with our OAuth routes

This commit:
 - move the /client/{client_id} route

Closes: #7054 

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth/client/get.js
+++ b/packages/fxa-auth-server/lib/routes/oauth/client/get.js
@@ -1,0 +1,50 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const hex = require('buf').to.hex;
+const Joi = require('@hapi/joi');
+
+const AppError = require('../../../oauth/error');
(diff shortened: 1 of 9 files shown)


packages/fxa-auth-server/lib/routes/oauth/client/get.js
```diff
@@ -0,0 +1,51 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const hex = require('buf').to.hex;
+const Joi = require('@hapi/joi');
+
+const AppError = require('../../../oauth/error');
+const db = require('../../../oauth/db');
+const validators = require('../../../oauth/validators');
+
+module.exports = (log) => ({
```
dannycoates, line 12: I wonder if we should move the `db` require into `index.js` and pass it in like `log`. I think eventually `oauth/db` will change to initialize after require, i.e. not export `new OauthDB()`

dannycoates (approved):

packages/fxa-auth-server/lib/routes/oauth/client/get.js
chenba: I moved the require into `index.js` and did some renaming.  Please take another quick look.  Thanks!

dannycoates (approved):
I like. 👍

PR #7061, opened 2020-12-03
fix: capture GraphQL errors in field resolvers

### LINKED ISSUE #7059: GraphQL errors should be reported to Sentry
Errors in graphql field resolvers are not being reported to Sentry. They should be.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2778)
┆Issue Number: FXA-2778

bbangert:
Because:

* Interceptors for NestJS do not capture field resolvers due to how
  apollo-server runs the query/mutation before then running separate
  queries to resolve fields.

This commit:

* Hooks into an apollo-server request lifecycle hook made for error
  reporting to accurately report errors to Sentry that occur when
  resolving fields.

Closes #7059

## Checklist

### DIFF
--- a/packages/fxa-shared/nestjs/sentry/reporting.ts
+++ b/packages/fxa-shared/nestjs/sentry/reporting.ts
@@ -118,10 +118,14 @@ export function captureSqsError(err: Error, message?: SQS.Message): void {
  * @param request A request object if available.
  */
 export function reportRequestException(
-  exception: Error,
+  exception: Error & { reported?: boolean },
   excContexts: ExtraContext[] = [],
   request?: Request
 ) {
+  // Don't report already reported exceptions
+  if (exception.reported) {
+    return;
+  }
   Sentry.withScope((scope: Sentry.Scope) => {
     scope.addEventProcessor((event: Sentry.Event) => {
       if (request) {
@@ -135,6 +139,7 @@ export function reportRequestException(
       scope.setContext(ctx.name, ctx.fieldData);
     }
     Sentry.captureException(exception);
+    exception.reported = true;
   });
 }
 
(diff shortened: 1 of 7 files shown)


packages/fxa-shared/nestjs/sentry/reporting.ts
```diff
@@ -122,6 +122,10 @@ export function reportRequestException(
   excContexts: ExtraContext[] = [],
   request?: Request
 ) {
+  // Don't report already reported exceptions
```
bbangert, line 125: This is needed because the Sentry interceptor *and* the lifecycle hook error reporter are triggered by the same mutation/query level error.

packages/fxa-shared/nestjs/sentry/reporting.ts
```diff
@@ -122,6 +122,10 @@ export function reportRequestException(
...
+  if ((exception as any).reported) {
```
jodyheavener, line 126: Maybe `exception`'s type could be changed to `exception: Error & { reported: boolean }` so we don't need to cast to `any`?

jodyheavener (approved):
R+ with a nit

packages/fxa-shared/nestjs/sentry/reporting.ts
bbangert: Oh, nice call! Updated the type to avoid the casts.

packages/fxa-shared/nestjs/sentry/sentry.plugin.ts
```diff
@@ -0,0 +1,53 @@
...
+          return;
+        }
+        for (const err of ctx.errors) {
+          // Only report internal server errors,
+          // all errors extending ApolloError should be user-facing
+          if (err instanceof ApolloError) {
+            continue;
+          }
+          // Skip errors with a status already set or already reported
+          if (err.originalError?.status) {
```
chenba, line 34: I see a `reported` property set in the file above, but I don't see it being checked here.  Unless I misunderstanding the comment, it should be checked here?

PR #7078, opened 2020-12-05
chore(fxa-content-server): apply the MX record validation feature flag to the validate-email-domain endpoint

Cronus1007:
## Because

-

## This pull request

-

Closes: #5896 
## Checklist

## Screenshots (Optional)

@chenba 
Fixxed all the requirements in the fxa-content-server for both the front and backend. As well added the required test cases in both the front and backend . Also one more question that I am unable to add the reviewers  since I am missing the reviewers section. Maybe I don't have the required permissions.
Screenshot from 2020-12-05 16-52-28

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
+++ b/packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
@@ -120,6 +120,9 @@ const checkEmailDomain = ($element, view) => {
           return reject(REJECTION_CODE);
         }
 
+        if (result === 'skip') {
+          return resolve();
+        }
         showInvalidDomainError(domain);
         view.logFlowEvent('email-domain-validation.block');
         reject(REJECTION_CODE);
--- a/packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
@@ -134,7 +134,17 @@ describe('lib/email-domain-validator', () => {
       xhr.ajax.restore();
     });
   });
-
+  it('should resolve on a sucessful skip record', () => {
+    sinon.stub(xhr, 'ajax').returns(Promise.resolve({ result: 'skip' }));
+    $el.val('quuz@great.skip.co.gd');
+    return checkEmailDomain($el, view)
+      .then(() => {
+        xhr.ajax.restore();
+      })
+      .catch(() => {
+        assert.fail('validation should have been skipped');
+      });
+  });
   it('should reject and show a tooltip on a successful A record lookup', () => {
     sinon.stub(xhr, 'ajax').returns(Promise.resolve({ result: 'A' }));
     const renderSpy = sinon.spy(Tooltip.prototype, 'render');
--- a/packages/fxa-content-server/server/lib/routes.js
+++ b/packages/fxa-content-server/server/lib/routes.js
@@ -31,7 +31,7 @@ module.exports = function (config, i18n) {
     require('./routes/redirect-download-firefox')(config),
     require('./routes/redirect-m-to-adjust')(config),
     require('./routes/get-500')(config),
-    require('./routes/validate-email-domain')(),
+    require('./routes/validate-email-domain')(config),
   ];
 
   if (config.get('csp.enabled')) {
--- a/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
+++ b/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
@@ -13,7 +13,7 @@ const joi = require('joi');
 const dns = require('dns');
 const VError = require('verror');
 const resolver = new dns.promises.Resolver();
-const results = ['MX', 'A', 'none'].reduce((accumulator, val) => {
+const results = ['MX', 'A', 'none', 'skip'].reduce((accumulator, val) => {
   accumulator[val] = val;
   return accumulator;
 }, {});
@@ -57,7 +57,7 @@ const tryResolveWith = (resolveFunc) => async (domain) => {
   }
 };
 
-module.exports = function () {
+module.exports = function (config) {
   return {
     method: 'get',
     path: '/validate-email-domain',
@@ -68,7 +68,10 @@ module.exports = function () {
     },
     process: async function (req, res, next) {
       const { domain } = req.query;
-
+      const shouldValidate = config.get('mxRecordValidation.enabled');
+      if (!shouldValidate) {
+        return res.json({ result: results.skip });
+      }
       try {
         if (await tryResolveWith(resolver.resolveMx.bind(resolver))(domain)) {
           return res.json({ result: results.MX });
--- a/packages/fxa-content-server/tests/server/routes/validate-email-domain.js
+++ b/packages/fxa-content-server/tests/server/routes/validate-email-domain.js
@@ -50,14 +50,26 @@ registerSuite('routes/validate-email-domain', {
         const validateEmailDomainRoute = proxyquireWithDns(dns);
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(true),
+        });
 
         return route.process(req, res).then(() => {
           assert.isTrue(resolveMxStub.calledOnceWith('abc.xyz'));
           assert.isTrue(res.json.calledOnceWith({ result: 'MX' }));
         });
       },
-
+      'responds with {result: skip} when there is a skip record': () => {
+        const validateEmailDomainRoute = require('../../../server/lib/routes/validate-email-domain');
+        const req = { query: { domain: 'skip.abc.xyz' } };
+        const res = { json: sinon.stub() };
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(false),
+        });
+        return route.process(req, res).then(() => {
+          assert.isTrue(res.json.calledOnceWith({ result: 'skip' }));
+        });
+      },
       'responds with {result: "A"} when there is an A record': () => {
         const resolveMxStub = sinon.stub().resolves([]);
         const resolve4Stub = sinon.stub().resolves(['abc.xyz']);
@@ -74,7 +86,9 @@ registerSuite('routes/validate-email-domain', {
         const validateEmailDomainRoute = proxyquireWithDns(dns);
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(true),
+        });
 
         return route.process(req, res).then(() => {
           assert.isTrue(resolveMxStub.calledOnceWith('abc.xyz'));
@@ -99,7 +113,9 @@ registerSuite('routes/validate-email-domain', {
         const validateEmailDomainRoute = proxyquireWithDns(dns);
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(true),
+        });
 
         return route.process(req, res).then(() => {
           assert.isTrue(resolveMxStub.calledOnceWith('abc.xyz'));
@@ -128,7 +144,9 @@ registerSuite('routes/validate-email-domain', {
         const validateEmailDomainRoute = proxyquireWithDns(dns);
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns('true'),
+        });
 
         return route.process(req, res).then(() => {
           assert.isTrue(resolveMxStub.calledOnceWith('abc.xyz'));
@@ -155,7 +173,9 @@ registerSuite('routes/validate-email-domain', {
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
         const next = sinon.stub();
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(true),
+        });
 
         return route.process(req, res, next).then(() => {
           assert.isTrue(resolveMxStub.calledOnceWith('abc.xyz'));
@@ -182,7 +202,9 @@ registerSuite('routes/validate-email-domain', {
         const req = { query: { domain: 'abc.xyz' } };
         const res = { json: sinon.stub() };
         const next = sinon.stub();
-        const route = validateEmailDomainRoute();
+        const route = validateEmailDomainRoute({
+          get: sinon.stub().returns(true),
+        });
 
         return route.process(req, res, next).then(() => {
           assert.isFalse(next.calledWith(error));


Cronus1007:
Extends over #7078

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
```diff
@@ -120,6 +120,9 @@ const checkEmailDomain = ($element, view) => {
           return reject(REJECTION_CODE);
         }
 
+        if(result==='skip') {
+          return resolve();
```
Cronus1007, line 124: resolved while making a skip record

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -134,7 +134,32 @@ describe('lib/email-domain-validator', () => {
       xhr.ajax.restore();
     });
   });
-
+  it('should resolve on a sucessful skip record', () => {
```
Cronus1007, line 137: frontend test for a successful skip record.

packages/fxa-content-server/server/lib/routes.js
```diff
@@ -31,7 +31,7 @@ module.exports = function (config, i18n) {
     require('./routes/redirect-download-firefox')(config),
     require('./routes/redirect-m-to-adjust')(config),
     require('./routes/get-500')(config),
-    require('./routes/validate-email-domain')(),
+    require('./routes/validate-email-domain')(config),
```
Cronus1007, line 34: passing the configs for the routes

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -13,7 +13,7 @@ const joi = require('joi');
 const dns = require('dns');
 const VError = require('verror');
 const resolver = new dns.promises.Resolver();
-const results = ['MX', 'A', 'none'].reduce((accumulator, val) => {
+const results = ['MX', 'A', 'none', 'skip'].reduce((accumulator, val) => {
```
Cronus1007, line 16: added a skip record

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -68,7 +68,10 @@ module.exports = function () {
     },
     process: async function (req, res, next) {
       const { domain } = req.query;
-
+      const shouldValidate = config.get('mxRecordValidation.enabled');
```
Cronus1007, line 71: passed configuration file to validate-email-domain.js

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,19 @@ registerSuite('routes/validate-email-domain', {
           assert.isTrue(res.json.calledOnceWith({ result: 'MX' }));
         });
       },
-
+      'responds with result {result: "skip" when there is a skip record}':()=>{
```
Cronus1007, line 60: test cases added for the updated backend

Cronus1007:
cc @chenba

chenba:
> I am unable to add the reviewers since I am missing the reviewers section. Maybe I don't have the required permissions.

Yeah, sorry.  People on our team do regularly check the open PRs, so it shouldn't be too much of  problem I think.

chenba:
There's a typo with "applying" in the commit message, can you please fix that?

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
chenba: Can you put spaces around `===`?  I'm surprised that prettier didn't auto-format it.

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -134,7 +134,32 @@ describe('lib/email-domain-validator', () => {
...
+    return checkEmailDomain($el, view)
+      .then(() => {
+        assert.isTrue(view.logEvent.calledWith('emailDomainValidation.skip'));
+        assert.isTrue(view.logEvent.calledWith('emailDomainValidation.skip'));
+        assert.isTrue(
+          view.logFlowEvent.calledWith('email-domain-validation.skipped')
+        );
+        assert.isTrue(
+          view.logFlowEvent.calledWith('email-domain-validation.skip')
+        );
```
chenba, line 149: The code doesn't log any events, so we don't need these assertions here.

packages/fxa-content-server/app/tests/spec/lib/email-domain-validator.js
```diff
@@ -134,7 +134,32 @@ describe('lib/email-domain-validator', () => {
...
+        xhr.ajax.restore();
+      })
+      .catch(() => {
+        assert.fail(view.logEvent.calledWith('emailDomainValidation.skip'));
+        assert.fail(view.logEvent.calledWith('emailDomainValidation.skip'));
+        assert.fail(
+          view.logFlowEvent.calledWith('email-domain-validation.skipped')
+        );
+        assert.fail(
+          view.logFlowEvent.calledWith('email-domain-validation.skipped')
```
chenba, line 159: We just need a `assert.fail('validation should have been skipped');` here.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,19 @@ registerSuite('routes/validate-email-domain', {
...
+        const dns = {
+          ...ERROR_CODES,
+          //Promises always resolved so no need to give a Resolver
+        };
+        const validateEmailDomainRoute = proxyquireWithDns(dns);
```
chenba, line 65: Since execution shouldn't get to the part of the code that uses `dns`, we don't need to `proxyquire`.  You can just `require` it.  However, you do need to pass a config object into it so that the `enabled` flag is set to false.

chenba (changes requested):
Thanks for the new commit!  The tests need some changes.

You can try running the tests, once you have the services running locally, by:

 - running `yarn test` from fxa/packages/fxa-content-server
 - going to http://localhost:3030/tests/index.html

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
chenba: Instead of me nitpicking about code formatting, you could run prettier with `yarn prettier --write --config ../../_dev/.prettierrc FILE_NAMES`, e.g. `yarn prettier --write --config ../../_dev/.prettierrc app/scripts/lib/email-domain-validator.js`, from the fxa-content-server directory.

If you use VS Code, there's a prettier extension.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +58,21 @@ registerSuite('routes/validate-email-domain', {
...
+        const dns = {
+          ...ERROR_CODES,
+          //Promises always resolved so no need to give a Resolver
+        };
+        const validateEmailDomainRoute = require(dns);
+        const enabled = config.get('mxRecordValidation');
+        enabled.enabled = false;
+        const req = { query: { domain: 'skip.abc.xyz' } };
+        const res = { json: sinon.stub() };
+        const route = validateEmailDomainRoute();
```
chenba, line 71: `validateEmailDomainRoute` should be the module you changed above; see line 32 for the `require` call.  Then when you call `validateEmailDomainRoute()`, pass in teh config object.  You can mock that config object with sinon: `{get: sinon.stub().returns(false)}`.

chenba (changes requested):
One of the tests need some changes.  Thanks!

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
Cronus1007: No i am using prettier ext

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
Cronus1007: @chenba rather we can use this 
const stubConfigGet= sinon.stub(config,get)
stubConfigGet.withArgs('mxRecordValidate.enabled').returns(false)
stubConfigGet.restore()

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
Cronus1007: i made prettier to write this file now. so no issue regarding the linitng errors

chenba: hmm are you using the prettierrc from the project?  I see different formatting when I run the command in my above comment?

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -57,7 +57,7 @@ const tryResolveWith = (resolveFunc) => async (domain) => {
   }
 };
 
-module.exports = function () {
+module.exports = function (config) {
```
chenba, line 60: We need to update the existing tests to pass a config object.  An example of an existing test failing:

```
× node - routes/validate-email-domain - route.process - responds with {result: "MX"} when there is a MX record
TypeError: Cannot read property 'get' of undefined
  at Object.process @ server/lib/routes/validate-email-domain.js:71:37
  at Test.responds with {result: "MX"} when there is a MX record [as test] @ tests/server/routes/validate-email-domain.js:55:22
  @ ../../src/lib/Test.ts:268:51
```

Cronus1007: @chenba Let me update the routes.js test file. After that everyting is going to be fine I guess :)

packages/fxa-content-server/app/scripts/lib/email-domain-validator.js
Cronus1007: yup.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
Cronus1007: @chenba here we have provided a config object to the complete function so each test is going to break since all the other test does not have the config object

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,17 @@ registerSuite('routes/validate-email-domain', {
           assert.isTrue(res.json.calledOnceWith({ result: 'MX' }));
         });
       },
-
+      'responds with result {result: "skip"} when there is a skip record}': () => {
+        const validateEmailDomainRoute = require('../../../server/lib/routes/validate-email-domain');
```
Cronus1007, line 61: @chenba here we have called the dunction from the validate-email-domain file rather in other tests we are doing the same via proxyrequiredns. Earlier this thing was ok since no config was passed but now it breaks. So we have to update  all the validateEmailDomainROutes to require from the file and while making the routes we have to pass config object there . Because in other tests if we see  `route.process` function is any and in the test file I wrote vscode intellisence say the `process ` function which I called.

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -57,7 +57,17 @@ registerSuite('routes/validate-email-domain', {
...
+        const req = { query: { domain: 'skip.abc.xyz' } };
+        const res = { json: sinon.stub() };
+        const route = validateEmailDomainRoute({
```
Cronus1007, line 64: @chenba  Here what I will do I will call `config` object from `../../../server/lib/configuration.js` and pass the config object to the route.

Cronus1007 (commented):
@chenba

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
chenba: Yes.  Please update the tests that are now failing due to the changes introduce in this PR.  Thanks!

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
chenba: The fix for the existing tests are almost identical to what's done here.  Instead of the stub for the mocked config's `get` returning false, it'd return true.

packages/fxa-content-server/server/lib/routes/validate-email-domain.js
Cronus1007: @chenba  Commiting the new changes.

Cronus1007:
@chenba pushed latest commits

Cronus1007:
@chenba Appears that everything is fine 
Screenshot from 2020-12-13 18-18-24
Screenshot from 2020-12-13 18-18-49

packages/fxa-content-server/tests/server/routes/validate-email-domain.js
```diff
@@ -38,43 +23,39 @@ registerSuite('routes/validate-email-domain', {
...
       'responds with {result: "MX"} when there is a MX record': () => {
         const resolveMxStub = sinon.stub().resolves(['mx.abc.xyz']);
-        const dns = {
-          ...ERROR_CODES,
-          promises: {
-            Resolver: function () {
-              this.resolveMx = resolveMxStub;
-            },
-          },
-        };
```
chenba, line 48: We still need to mock `dns` in the test cases that we uses `dns`.

If you run `yarn test-server` from the fxa-content-server directory, you can see which tests are failing.

chenba (changes requested):
Existing tests are failing since `dns` is not mocked and proxyquired.

chenba:
Here are the failing tests that I see:

```
× node - routes/validate-email-domain - route.process - responds with {result: "A"} when there is an A record
AssertionError: expected false to be true
  @ tests/server/routes/validate-email-domain.js:61:18
× node - routes/validate-email-domain - route.process - responds with {result: "none"} when there are no DNS records
AssertionError: expected false to be true
  @ tests/server/routes/validate-email-domain.js:79:18
× node - routes/validate-email-domain - route.process - responds with {result: "none"} when resolve functions throw NODATA or NOTFOUND errors
AssertionError: expected false to be true
  @ tests/server/routes/validate-email-domain.js:100:18
× node - routes/validate-email-domain - route.process - calls next() with an error when resolve functions throw a non- dns.* error
AssertionError: expected false to be true
  @ tests/server/routes/validate-email-domain.js:119:18
× node - routes/validate-email-domain - route.process - calls next() with a wrapped error on a dns.* error, excluding NODATA and NOTFOUND
AssertionError: expected false to be true
  @ tests/server/routes/validate-email-domain.js:137:18
```

Cronus1007:
@chenba  Only skip is passing with a little change .

Cronus1007:
@chenba  Actually the thing which is making all the conflicts is passing config in the form of a parameter.

Cronus1007:
@chenba  Have a look at the PR mentioned as a draft

chenba:
@Cronus1007 I think we are pretty close with this PR.  To fix the existing tests, you need to pass in a config object to the route handlers so that they won't skip the dns validation.  I.e.

```
        const route = validateEmailDomainRoute({
          get: sinon.stub().returns(false),
        });
```

Cronus1007:
@chenba  I think this looks fine now :) 
Screenshot from 2020-12-15 21-32-43

chenba:
@Cronus1007 Looks good.  Can you squash the commits into a single commit?  Thanks!

Cronus1007:
Done @chenba

Cronus1007:
@chenba Would be great if you recommend some more issues like this .

chenba:
> 
> 
> @chenba Would be great if you recommend some more issues like this .

I don't have anything in mind at the moment, but I will keep a lookout and @ you when I see one.

chenba:
@Cronus1007 I created https://github.com/mozilla/fxa/pull/7184 to run the CI checks and they are all passing.  But could you amend the commit to update the commit message so that it's just a single message?  Thanks!

Cronus1007:
> @Cronus1007 I created #7184 to run the CI checks and they are all passing. But could you amend the commit to update the commit message so that it's just a single message? Thanks!

@chenba I am unable to understand what you tried to quote.

Cronus1007:
@chenba  Done .

chenba (approved):
@Cronus1007 Thanks for the patch

PR #7081, opened 2020-12-07
refactor(oauth): move GET and POST /authorization routes

chenba:
Because:
 - we want oauth routes better integrated into auth-server

This commit:
 - move the GET and POST /authorization routes
 - copy some route config modifying logic from oauth/routes

Closes: #7052

### DIFF
--- a/packages/fxa-auth-server/lib/routes/oauth/authorization.js
+++ b/packages/fxa-auth-server/lib/routes/oauth/authorization.js
@@ -1,0 +1,248 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const hex = require('buf').to.hex;
+const Joi = require('@hapi/joi');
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/oauth/authorization.js
```diff
@@ -0,0 +1,244 @@
...
+          log.fatal('joi.response_type', {
+            response_type: req.payload.response_type,
+          });
+          throw AppError.invalidResponseType();
+      }
+    },
+  },
+});
+
+async function generateAuthorizationCode(client, payload, grant, log, oauthDB) {
```
dannycoates, line 159: I see you chose to pass `log` and `oauthDB` as arguments. This is different from how I converted `/token` where I used a closure. IDK which we should favor but we should stick to one style.

dannycoates (approved):
LGTM, but I have a question about style.

packages/fxa-auth-server/lib/routes/oauth/authorization.js
chenba: Closures work for me.  I'll update.  👍

PR #7084, opened 2020-12-08
feat(gql-api): add customs check on query/mutations

### LINKED ISSUE #6876: /graphql allows queries with high read and write amplification

bbangert:
Because:

* We want to limit query/mutations by uid/ip using customs.

This commit:

* Alters the session token verification to use a direct db query to load
  the session data from the auth database so the email is available for
  use with customs check.
* Adds a Gql customs check that queries customs for the given
  class/method.

Closes #6876

## Checklist

### DIFF
--- a/packages/fxa-graphql-api/src/auth/gql-customs.guard.ts
+++ b/packages/fxa-graphql-api/src/auth/gql-customs.guard.ts
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { GqlExecutionContext } from '@nestjs/graphql';
+import { Request } from 'express';
+import { CustomsService } from 'fxa-shared/nestjs/customs/customs.service';
+
+import { SessionTokenResult } from './session-token.strategy';
+
+@Injectable()
+export class GqlCustomsGuard implements CanActivate {
+  constructor(private customs: CustomsService) {
+    if (!customs) {
+      throw new Error('No customs service provided.');
+    }
+  }
+
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const ctx = GqlExecutionContext.create(context);
+    const req = ctx.getContext().req as Request;
+    const { email } = (req.user as SessionTokenResult).session;
+    const handlerName = context.getHandler().name;
+    const className = (context as any).constructorRef.name;
+    const customsName = [className, handlerName].join('.');
+    await this.customs.check({
+      action: customsName,
+      email,
+      ip: req.ip,
+      headers: req.headers,
+      query: req.query,
+    });
+    return true;
+  }
+}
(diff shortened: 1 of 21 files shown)


dannycoates:
first pass lgtm

g-k:
I'm not 100% sure on what's going on in this PR, but I like the approach and it lgtm so far

packages/fxa-graphql-api/src/auth/gql-customs.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Request } from 'express';
+import { CustomsService } from 'fxa-shared/nestjs/customs/customs.service';
+
+import { SessionTokenResult } from './session-token.strategy';
+
+@Injectable()
+export class GqlCustomsGuard implements CanActivate {
+  constructor(private customs: CustomsService) {
+    if (!customs) {
+      throw new Error('No customs service provided.');
```
jodyheavener, line 15: Huh, I thought Nest's dependency injection took care of this for us?

packages/fxa-shared/nestjs/customs/customs.service.ts
```diff
@@ -0,0 +1,80 @@
...
+  HttpException,
+  HttpStatus,
+  Injectable,
+} from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import superagent from 'superagent';
+
+type AnyObject = {
+  [key: string]: any;
+};
```
jodyheavener, line 15: `Record<string, any>`?

jodyheavener (commented):
Couple Qs but don't see anything glaring

packages/fxa-graphql-api/src/auth/gql-customs.guard.ts
bbangert: same here, but due to how guards work it seems that it did not locate it when loading in one case, such that I had to add this

dannycoates (approved):
This LGTM. So the behavior is that we'll limit each resolver individually, which seems ok.

PR #7087, opened 2020-12-08
chore(dep): update tailwindcss to version 2

vbudhram:
## Because

- I attempted to update to tailwindcss 2 as dependabot did in https://github.com/mozilla/fxa/pull/7079, and the supporting plugins (postcss-import and autoprefixer), however I still got the `Error: true is not a PostCSS plugin`. I am assuming somewhere in dependency tree a plugin hasn't been updated to support new tailwind but I couldn't find it
- For now we can use the capability version of [tailwindcss](https://tailwindcss.com/docs/installation#post-css-7-compatibility-build) 2 to get the latest features
- Once merged I can file a follow up issue to circle back and not use the capability version

## This pull request

- updates dependency

## Checklist


chenba (approved):
Pretty nice that they provide that option.

Issue #7015, opened 2020-11-26
[FxA-Settings] The selected section from Settings menu is focused only at a second click


vasilicatamas:
**Build**
Prod [Train 1.194.0]

**Affected platforms**
Windows 10 64-bit
Android 9
iOS 13.6

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on "Profile", "Security" or "Connected Services".
3. Click a second time on "Profile", "Security" or "Connected Services".

**Expected results**
Both menu button successfully display a tooltip

**Actual Results**
No tooltip is displayed for navigation menu and “X” button.

![issue54](https://user-images.githubusercontent.com/18114210/100371832-0b792c80-3011-11eb-862f-e5da96451a5b.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2744)
┆Issue Number: FXA-2744

data-sync-user:
➤ Barry Chen commented:

The fix for FXA-2651 is the cause of this bug.

vasilicatamas:
This issue is verified as fixed on FxA Stage [Train 1.197.0] under Windows 10 64-bit, Android 11 and iOS 13.6. The Settings sections are focused at the first click.

![issue73](https://user-images.githubusercontent.com/18114210/103897282-f6a5b380-50fb-11eb-81a5-2ec707460464.gif)

PR #7095, opened 2020-12-08
fix(settings): scroll to element id in url hash

chenba:
Because:
 - we have some logic for scrolling to the top of the page when the user
   navigates to a new route

This commit:
 - account for the hrefs with a hash and let the browser does its default
   thing

Closes: #7015


chenba:
Note to reviewer: the failing test has nothing to do with the change.

chenba:
Filed #7096 for that "failing" test.

dannycoates (approved):

PR #7105, opened 2020-12-09
fix(gql-api): disallow query in query string

bbangert:
Because:

* The complexity analysis doesn't seem to notice queries done in the
  query string.

This commit:

* Rejects requests that include a query param in the query string.

Issue #6876

## Checklist


dannycoates (approved):
cool

PR #7120, opened 2020-12-10
task(fxa-content-server): pass env variable to render beta link

meandavejustice:


packages/fxa-content-server/app/scripts/views/settings.js
```diff
@@ -80,6 +78,7 @@ const View = BaseView.extend({
     this._marketingEmailEnabled = options.marketingEmailEnabled !== false;
     this._subscriptionsManagementEnabled =
       options.subscriptionsManagementEnabled !== false;
+    this.enableBeta = options.config.enableBeta;
```
chenba, line 81: To be consistent I think this should be `this._enableBeta`.

chenba (approved):
Verified working locally.  Would liek the variable name changed for consistency.  Thanks

Issue #6975, opened 2020-11-19
[FxA-Settings] Connected Services missing icon


janiceahn:
## Description
No icon next my instance of Firefox.

![Screenshot_2020-11-19 Firefox Accounts(1)](https://user-images.githubusercontent.com/6219141/99707263-33ddb580-2a6a-11eb-89ad-d514ba045a27.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2843)
┆Issue Number: FXA-2843

jodyheavener:
No icons show up for me: 

<img width="500" alt="Screen Shot 2020-12-10 at 6 40 27 PM" src="https://user-images.githubusercontent.com/6392049/101843328-f70c5800-3b17-11eb-8740-5cefc2236597.png">

meandavejustice:
Found a fix for this

vasilicatamas:
This issue is verified as fixed on FxA Stage [Train 1.197.0] under Windows 10 64-bit, Android 11 and iOS 13.6. Connected services icons are successfully displayed in the new Settings page.

![Screenshot_21](https://user-images.githubusercontent.com/18114210/103885808-3dd67900-50e9-11eb-95d8-3ae1a5f404e6.png)

PR #7124, opened 2020-12-11
fix(fxa-settings): dynamic loading svg fix

meandavejustice:
Turned out to be some weirdness with create-react-app
found the solution here: https://github.com/facebook/create-react-app/issues/5276#issuecomment-665628393

Closes: #6975

## Checklist


meandavejustice:
This is now failing on all of the Connected Services tests, it appears that we import things differently within the tests 🤔

meandavejustice:
cc/ @chenba we'll probably want to do a point release with this, let's see if there's anything else we want to get in tomorrow though

chenba:
When I tried this locally I got:

`Unhandled Rejection (Error): Cannot find module '!!@svgr/webpack?-svgo,+titleProp,+ref!./web.svg'`

`yarn build` was successful though.

meandavejustice:
@chenba So apparently my workaround for the tests its what is causing the failure.

It works correctly when hard coded:
```
ImportedIconRef.current = (
          await import(`!!@svgr/webpack?-svgo,+titleProp,+ref!./${name}.svg`)
        ).default;
```

But fails with the `prefix` variable. Which I was using to detect when a test is being run. 🤔 

We may just statically load these icons if we can't find a reasonable workaround for the test failures, cus the dynamic loading isn't really giving us that many benefits anyways...

jodyheavener:
🚗 I'm all for manually loading each SVG 🚗

chenba:
image
image

Old vs. new.  Does that look right?

chenba (approved):

Issue #7140, opened 2020-12-11
[settings] Sentry dsn is empty


data-sync-user:
Sentry is init'd _before_ the config object is ready: https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/index.tsx#L18:L21

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2830)
┆Issue Number: FXA-2830

PR #7141, opened 2020-12-11
fix(settings): use Sentry dsn after config's loaded

chenba:
Because:
 - we are not using the Sentry dsn

This commit:
 - init Sentry after the config's ready

Closes: #7140


dannycoates (approved):
🤦😂

PR #7144, opened 2020-12-12
feat(db): Create table to store PayPal customer information

### LINKED ISSUE #7031: Add DB table or columns for PayPal fields
Referencing the [ADR|[https://github.com/mozilla/fxa/pull/6926](https://github.com/mozilla/fxa/pull/6926)]

It looks like the information we need to keep track of includes:

* PayPal Billing Agreement IDs (these are also issued by Reference Transactions)
  * A PayPal customer could have more than one Billing Agreement ID for example if they cancel an existing BA and then decide they want to uncancel, or BAs can get canceled for other reason like funding is invalid. If a BA is suspended, we should cancel it.
  * We need to keep track of all BA IDs to know the entire transaction history.
* Status information: Suspended, Canceled, Open
* When it was created
* When it was ended

We need to associate these IDs to a Stripe Customer ID. Since we already have Stripe Customer ID in the Account-Customers table, we don't need to store that. We can join this table with that table.

We would not hit this table to charge each month, since we store the active PayPal reference ID in Stripe metadata.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2760)
┆Issue Number: FXA-2760

biancadanforth:
Because:

* We want to be able to tell if an FxA user has an associated PayPal billing agreement without having to query PayPal

This Commit:

* Creates the paypalCustomers table that will be used to associate FxA user ids with PayPal billing agreements

Closes #7031

## Checklist

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/patch.js
+++ b/packages/fxa-auth-db-mysql/lib/db/patch.js
@@ -5,4 +5,4 @@
 // The expected patch level of the database. Update if you add a new
 // patch in the ./schema/ directory.
 
-module.exports.level = 113;
+module.exports.level = 114;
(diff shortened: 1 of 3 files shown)


chenba (approved):
LGTM

bbangert (approved):

dannycoates:
@biancadanforth we usually let the author do the merge after the r+ ;)

biancadanforth:
Hmm... looks like I am not authorized to merge into `main`. @clouserw , would you be able to authorize me?

PR #7155, opened 2020-12-15
fix(totp): downgrade session to email-based verification after TOTP is removed

### LINKED ISSUE #499: Should disabling TOTP remove TOTP verification state from existing sessions?
Here's a bit of oddness around enabling and then disabling TOTP:

* Create a new account, enable TOTP
* Sign in to an OAuth relier, they will see `amrValues: ["pwd", "otp"]` in the `id_token` as expected
* Disable TOTP on the account
* In your existing login session, sign in to the OAuth relier again.
* They will still see `amrValues: ["pwd", "otp"]` in the `id_token`, even though the account no longer has 2FA enabled.

This is acceptable behaviour, or should disabling TOTP remove the "this session was verified with TOTP" state from any existing session tokens? /cc @gdestuynder.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-673)
┆Issue Number: FXA-673

### LINKED ISSUE #5154: 2FA account state is not updated after 2FA is disabled
Noticed this while testing the 2FA-required flow: if a user enables 2FA, then disables it, the acr_values=AAL2 check will still allow the user to log in until the user logs out of their account.

I suspect it's a profile caching issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1788)
┆Issue Number: FXA-1788

### LINKED ISSUE #6762: Trying to login on AMO with disabled 2FA ends up endless redirect loop
## Description

Trying to login on AMO with a mozilla.com account that had 2FA enabled previously but has it disabled now makes me stuck in an endless FxA redirect loop.

### Steps to reproduce

1. Have an account with a mozilla.com email.
2. Enable 2FA.
3. Login on addons.mozilla.org and make sure login worked currently. Then logout.
4. Disable 2FA.
5. Try to login on addons.mozilla.org again.

### Expected result

A prompt to enable 2FA.

### Actual result

Endless redirect loop.

### Environment

Firefox 83.0b4 on macOS 10.15.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2644)
┆Issue Number: FXA-2644

6a68:
Retrying the fix for #7111 at the right level: when 2FA is removed from the account, immediately downgrade the session. 

## Because

- AMO requires developers to have 2FA enabled.
- To minimize friction, AMO first logs the user in, then checks their developer state against the user's Firefox Account 2FA state, and if 2FA is not enabled on a developer account, redirects back to the FxA using the [`prompt=none` flow](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/oauth/prompt-none.md) and the user is prompted to enable 2FA in an inline flow, redirecting back to AMO when done.
- In the edge case where a user sets up 2FA, then logs out of AMO, then disables 2FA, then logs back in to AMO, the user gets trapped in a redirect loop (#6762)
- The redirect occurs because, as outlined in an [amazingly thorough and helpful review comment](https://github.com/mozilla/fxa/pull/7111#issuecomment-742209730), a user's session verification state is upgraded to use 2FA when 2FA is enabled, but not downgraded back to using an email loop when 2FA is disabled

## This pull request

- Downgrades session verification when the 2FA token is destroyed, ensuring users won't fall into the redirect loop. Instead, they will be sent back into the inline 2FA setup flow.

Fixes #6762, fixes #5154, closes #499.

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/routes/totp.js
+++ b/packages/fxa-auth-server/lib/routes/totp.js
@@ -140,6 +140,12 @@ module.exports = (log, db, mailer, customs, config) => {
 
         await db.deleteTotpToken(uid);
 
+        // Downgrade the session to email-based verification when TOTP is
+        // removed. Because we know the session is already verified, there's
+        // no security risk in setting it as verified using a different method.
+        // See #5154.
+        await db.verifyTokensWithMethod(sessionToken.id, 'email-2fa');
+
         await log.notifyAttachedServices('profileDataChanged', request, {
           uid,
         });
--- a/packages/fxa-auth-server/test/local/routes/totp.js
+++ b/packages/fxa-auth-server/test/local/routes/totp.js
@@ -124,6 +124,19 @@ describe('totp', () => {
           'uid',
           'third argument was event data with a uid'
         );
+
+        assert.equal(
+          db.verifyTokensWithMethod.callCount,
+          1,
+          'called db.verifyTokensWithMethod'
+        );
+        const dbArgs = db.verifyTokensWithMethod.args[0];
+        assert.equal(dbArgs[0], sessionId, 'first argument was sessionId');
+        assert.equal(
+          dbArgs[1],
+          'email-2fa',
+          `second argument was reduced verification level`
+        );
       });
     });
 
@@ -424,6 +437,9 @@ function setup(results, errors, routePath, requestOptions) {
       qrCodeUrl: 'some base64 encoded png',
       sharedSecret: secret,
     });
+  });
+  db.verifyTokensWithMethod = sinon.spy(() => {
+    return P.resolve();
   });
   db.totpToken = sinon.spy(() => {
     return P.resolve({


dannycoates (approved):

packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -140,6 +140,10 @@ module.exports = (log, db, mailer, customs, config) => {
 
         await db.deleteTotpToken(uid);
 
+        // Downgrade the session to email-based verification when TOTP is
+        // removed. #5154
+        await db.verifyTokensWithMethod(sessionToken.id, 'email-2fa');
```
dannycoates, line 145: Do we need to updated other sessions besides the current one?

rfk: I think we should downgrade all sessions that have `totp-2fa`, yes. Although I FWIW I wouldn't be too upset to do that i a followup if we need to get this out urgently for AMO.

packages/fxa-auth-server/lib/routes/totp.js
```diff
@@ -140,6 +140,10 @@ module.exports = (log, db, mailer, customs, config) => {
 
         await db.deleteTotpToken(uid);
 
+        // Downgrade the session to email-based verification when TOTP is
+        // removed. #5154
```
rfk, line 144: nit: think it would be worth commenting here explicitly, that we know the session is already verified and so there's no security problem with re-marking it as verified with a different method.

rfk (approved):
👍

packages/fxa-auth-server/lib/routes/totp.js
6a68: Good catch! I'll do that in a followup and land this tonight

vbudhram (approved):
@6a68 LGTM 👍🏽

PR #7157, opened 2020-12-15
chore(fxa-content-server): apply the MX record validation feature fla…

Cronus1007:
…g to the validate-email-domain endpoint

## Because

-

## This pull request

-

Closes: #7078 

## Checklist

## Screenshots (Optional)

@chenba  Shall we proceed like this since through such a change can pass all the tests and the record whiich are `proxyquire` would have no harm as well the `skip` record can be added easily

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
+++ b/packages/fxa-content-server/server/lib/routes/validate-email-domain.js
@@ -13,7 +13,8 @@ const joi = require('joi');
 const dns = require('dns');
 const VError = require('verror');
 const resolver = new dns.promises.Resolver();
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/server/lib/routes/validate-email-domain.js
```diff
@@ -13,7 +13,8 @@ const joi = require('joi');
 const dns = require('dns');
 const VError = require('verror');
 const resolver = new dns.promises.Resolver();
-const results = ['MX', 'A', 'none'].reduce((accumulator, val) => {
+const config = require('../configuration');
```
chenba, line 16: We should pass in the config for this route like the others.  There's no need to break that pattern and potentially mislead/confuse future devs.

PR #7170, opened 2020-12-16
[fxa-auth-server] Upgrade uuid dependency

enote-kane:
## Because

The fxa-auth-server was depending on an ancient version of the `uuid` module.

## This pull request

Upgrades the `uuid` module dependency to benefit from the project-global cache.

Closes: _n/a_

## Checklist


dannycoates (approved):
Thank you @enote-kane 😄 this is great!

Issue #7091, opened 2020-12-08
remove lib/oauth/metrics


data-sync-user:
replace with `emitMetricsEvent` or equivalent

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2796)
┆Issue Number: FXA-2796

Issue #7090, opened 2020-12-08
remove lib/oauth/logging


data-sync-user:
replace all usage with the auth-server logger. this will require some rewiring

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2795)
┆Issue Number: FXA-2795

PR #7172, opened 2020-12-16
refactor(oauth): remove old oauth-server logger

dannycoates:
## Because

- we don't need multiple ways to log

## This pull request

- removes a bunch of "unused" log lines
- The one important change here is that the amplitude event for grant generated tokens moved from grant.js to the callers in order to clean up the import tree.

Closes: #7090
Closes: #7091


vbudhram (approved):
@dannycoates Digging all the code that is being removed 👍🏽.

It is probably worthwhile to update our deploy docs to mention the metrics thing just in case something does change downstream.

Issue #7096, opened 2020-12-08
[settings] Error in External Links Tests


data-sync-user:
`{{{{TypeError: this.skip is not a function`}}}}
{{ {{ {{ at Test.test @ tests/functional/settings_v2/external_links.js:59:12}}}}}}
{{ {{ {{ @ ../../src/lib/Test.ts:268:51}}}}}}{{ }}

See [https://app.circleci.com/pipelines/github/mozilla/fxa/15368/workflows/795ab9e3-e5bf-49ca-9f50-9f9d6c5b2313/jobs/208288](https://app.circleci.com/pipelines/github/mozilla/fxa/15368/workflows/795ab9e3-e5bf-49ca-9f50-9f9d6c5b2313/jobs/208288) for an example instance

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2798)
┆Issue Number: FXA-2798

data-sync-user:
➤ Barry Chen commented:

[~djustice@mozilla.com] assigned this to you because I remember you worked on those tests.  Feel free to unassign if you don't want to work on this.

PR #7182, opened 2020-12-17
fix(test): Fix settings v2 external link test

vbudhram:
## Because

- The external links functional tests were broken

## This pull request

- Updates the test to not use `async` because it wouldn't have [access to the scope](https://theintern.io/docs.html#Intern/4/api/lib%2FTest/skip) that contains the `skip` method.

Closes: #7096

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_v2/external_links.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/external_links.js
@@ -54,7 +54,7 @@ describe('external links', () => {
     );
   });
 
-  it('renders Subscriptions link in navigation when we are subscribed to a product', async () => {
+  it('renders Subscriptions link in navigation when we are subscribed to a product', async function () {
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/tests/functional/settings_v2/external_links.js
```diff
@@ -54,22 +54,24 @@ describe('external links', () => {
     );
   });
 
-  it('renders Subscriptions link in navigation when we are subscribed to a product', async () => {
+  it('renders Subscriptions link in navigation when we are subscribed to a product', function () {
```
dannycoates, line 57: I think it can be `async`, just not an arrow function

vbudhram: @dannycoates ah you're right good catch, updating!

dannycoates (approved):
💯

PR #7186, opened 2020-12-17
refactor(auth): use native promises more

dannycoates:
There's still a bunch of places that use bluebird specific apis but this eliminates all the easy ones.


packages/fxa-auth-server/lib/senders/select_email_services.js
```diff
@@ -4,7 +4,6 @@
 
 'use strict';
 
-const Promise = require('../promise');
```
bbangert, line 7: 😆

bbangert (approved):
LGTM! 👍

PR #7188, opened 2020-12-18
WIP downgrading multiple sessions on TOTP removal

### LINKED ISSUE #5154: 2FA account state is not updated after 2FA is disabled

6a68:
(Switching gears for the afternoon, but wanted to get this out now to be sure I'm going in the right direction.)

This is a followup from #7155 to downgrade _all_ sessions marked TOTP verified when the user disables TOTP. This should close #5154. It's basically a SQL one-liner that shouldn't cause any perf issues (the necessary index already exists on the sessionTokens table), wrapped in a ton of boilerplate changes to the auth and auth-db layers.

API thoughts: I'm adding a very specific `account/:id/downgradeSessions` API endpoint to auth-db to do this, which feels weird because it's such a specific edge case, but it also seems like the most direct path to a fix. The obvious (to me) alternative would be to modify the existing `POST /tokens/:tokenId/verifyWithMethod` endpoint to handle multiple tokens at once, but I didn't see a nice way to turn that single-token-oriented API into a batch-oriented API. I'm open to suggestions on nicer ways to handle this.

I still have to fill in some tests, but I verified locally that this does flip the verification method from 2 to 1 (i.e. from totp-2fa to email-2fa) on multiple sessions if present in the DB.

Unfortunately, looks like prettier did its thing with the remote/db_tests file. There's nothing to see there, feel free to ignore.

### DIFF
--- a/packages/fxa-auth-db-mysql/db-server/index.js
+++ b/packages/fxa-auth-db-mysql/db-server/index.js
@@ -119,6 +119,10 @@ function createServer(db) {
   api.post('/account/:id/reset', withIdAndBody(db.resetAccount));
   api.post('/account/:id/resetTokens', withIdAndBody(db.resetAccountTokens));
   api.post(
+    '/account/:id/downgradeSessions',
+    withIdAndBody(db.downgradeSessionVerificationMethod)
+  );
+  api.post(
     '/account/:id/verifyEmail/:emailCode',
     op(function (req) {
       return db.verifyEmail(req.params.id, req.params.emailCode);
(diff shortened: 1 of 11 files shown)


6a68:
Ran out of time; closing for now.

PR #7192, opened 2020-12-18
refactor(auth): use native promises in all production / non-test code

dannycoates:
there's still some bluebird in test and script code but that's a lower priority and needs more work.

## Because

- we should only need native promises and they ought to be more performant

### DIFF
--- a/packages/fxa-auth-server/lib/senders/sms.js
+++ b/packages/fxa-auth-server/lib/senders/sms.js
@@ -7,7 +7,6 @@
 const Cloudwatch = require('aws-sdk/clients/cloudwatch');
 const error = require('../error');
 const MockSns = require('../../test/mock-sns');
-const P = require('bluebird');
 const Sns = require('aws-sdk/clients/sns');
 const time = require('../time');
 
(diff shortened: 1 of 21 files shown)


packages/fxa-auth-server/lib/senders/sms.js
```diff
@@ -7,7 +7,6 @@
 const Cloudwatch = require('aws-sdk/clients/cloudwatch');
 const error = require('../error');
 const MockSns = require('../../test/mock-sns');
-const P = require('bluebird');
```
dannycoates, line 10: Converting this file / tests was interesting. The `setImmediate`s used in the tests, which were there to allow the `pollCurrentSpend` to run, aren't a sound way to ensure that it completed, and was leading to assertion races. I believe bluebird is more predictable in how it interacts with the event loop than native promises, but we shouldn't be relying on that. I refactored in order to manually trigger and wait for that function to complete. I then went on to convert to `await` and added a `close` function to stop the polling.

packages/fxa-auth-server/test/local/cad-reminders.js
```diff
@@ -199,7 +199,7 @@ describe('lib/cad-reminders', () => {
         } else {
           it('left the third reminders in redis', async () => {
             const reminders = await redis.zrange(reminder, 0, -1);
-            assert.deepEqual(reminders, ['wibble', 'blee']);
+            assert.deepEqual(new Set(reminders), new Set(['wibble', 'blee']));
```
dannycoates, line 202: the order of these has always been a source of flakiness

chenba (approved):
🥇

Issue #7016, opened 2020-11-26
[FxA-Settings] Settings sections are not correctly displayed when selected from the left menu


vasilicatamas:
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Windows 10 64-bit
Android 9
iOS 13.6

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on "Profile" ,"Security" or "Connected Services".
3. Observe how the selected section is displayed.

**Expected results**
The section is entirely displayed.

**Actual Results**
The top part of the section is not visible because it is covered by the header.

![Screenshot_46](https://user-images.githubusercontent.com/18114210/100373181-0cab5900-3013-11eb-8bf5-46ef6b7a5db0.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2745)
┆Issue Number: FXA-2745

data-sync-user:
➤ Alex Davis commented:

I noticed this too, it's pretty annoying and obvious.

vasilicatamas:
This issue is verified as fixed on FxA Stage [Train 1.197.0] under Windows 10 64-bit, Android 11 and iOS 13.6, Settings sections are correctly displayed when they are selected from the left menu.

![issue71](https://user-images.githubusercontent.com/18114210/103884904-e84d9c80-50e7-11eb-8003-62608c8e86d0.gif)

PR #7194, opened 2020-12-21
fix(settings): account for header height on anchor jump

chenba:
Because:
 - settings has a fixed heaer that covers up content

This commit:
 - account for the height of the header when navigating to a section
   with an anchor

Closes: #7016


dannycoates (approved):
WFM! it's awkward that on full-width the nav scrolls out of view but that's a separate issue

Issue #7128, opened 2020-12-11
Giant scary error when submitting an already used secondary email


data-sync-user:
From beta feedback:

Bad error message: when I tried to add a new secondary email address, the page told me that email address (my work email) was already associated with a different FxA (an old test account). Not a problem, but the error message the page gave me was a scary red JSON blob: 

!image-2020-12-11-08-29-23-374.png|thumbnail!

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2818)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/14700/image-2020-12-11-08-29-23-374.png">image-2020-12-11-08-29-23-374.png</a>
┆Issue Number: FXA-2818

data-sync-user:
➤ Barry Chen commented:

The message is from the graphql-api server as is.  This requires a server side fix.

data-sync-user:
➤ Barry Chen commented:

This appears to be caused by the NestJS conversion.

PR #7197, opened 2020-12-22
fix(gql-api): restore CatchGatewayError decorator

chenba:
Because:
 - new settings was receiving JSON as error messages

This commit:
 - restore the decorator that was used to catch and re-throw errors
   prior to the NestJS conversion

Closes: #7128

### DIFF
--- a/packages/fxa-graphql-api/src/gql/lib/error.ts
+++ b/packages/fxa-graphql-api/src/gql/lib/error.ts
@@ -1,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { ApolloError } from 'apollo-server';
+
+export const PROFILE_INFO_URL =
+  'https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/docs/API.md#errors';
+
+export function CatchGatewayError(
+  target: any,
+  propertyKey: string,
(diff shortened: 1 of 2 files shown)


chenba:
@bbangert I don't know why CatchGatewayError was removed.  Let me know if this is not the correct approach.

packages/fxa-graphql-api/src/gql/lib/error.ts
```diff
@@ -0,0 +1,38 @@
...
+  target: any,
+  propertyKey: string,
+  descriptor: PropertyDescriptor
+) {
+  const originalMethod = descriptor.value;
+  descriptor.value = async function (...args: any[]) {
+    try {
+      return await originalMethod.apply(this, args);
+    } catch (err) {
+      if (err.code && err.errno && err.message) {
```
vbudhram, line 20: Kind of stinks that auth-server and profile-server return different error objects and this is the only way to tell them apart.

Is there something in graphql that might make this better?

vbudhram (approved):
@chenba Should we add some basic tests for this? Otherwise LGTM 👍🏽 

Screen Shot 2021-01-04 at 10 28 35 AM

bbangert:
@chenba In NestJS, the correct way to do this would be to apply an Interceptor or ExceptionFilter on the class. There is a default interceptor which catches the errors and reports them (Sentry), but does not apparently handle reformatting the exceptions the same as the prior gateway exception catch did.

bbangert (changes requested):
See other comment on using a NestJS filter.

bbangert (approved):
This can merge for now, we can do this with an exception filter in a follow-up.

PR #7200, opened 2020-12-23
task(fxa-content-server): add log to track down entrypoint sentry error

meandavejustice:
## This pull request
Adding this to get more info about the request that is causing https://sentry.prod.mozaws.net/operations/fxa-content-server-prod/issues/6602579/events/latest/
- adds a log with query and url info for joi validation errors.

refs #3718 

## Checklist

Need to check with @jbuck to see how we will be able to access this once it is in production. Additionally we may be able to avoid this patch if I can get access to see `fxa-content-server.server.routes.validation.failed` in our logs. Waiting to hear back from @ckolos on that. It's the holidays so we may not get answers until after the break.

### DIFF
--- a/packages/fxa-content-server/server/bin/fxa-content-server.js
+++ b/packages/fxa-content-server/server/bin/fxa-content-server.js
@@ -217,6 +217,14 @@ function makeApp() {
 
   // log and capture any errors
   app.use((err, req, res, next) => {
+    // in order to track down the culprit of sentry issue, https://sentry.prod.mozaws.net/operations/fxa-content-server-prod/issues/6602579/events/latest/
+    if (err.joi && err.joi.isJoi)
+      console.error(
+        'GLOBAL SENTRY JOI ERROR: ',
+        req.path,
+        req.query,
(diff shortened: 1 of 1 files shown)


meandavejustice:
Here is the query for accessing the validation failure: `fxa-content-server.server.routes.validation.failed` in big query.

```SELECT * FROM `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_content_20201222` where jsonPayload.type = "server.routes.validation.failed"```


For many of these entries we have duplicate querys passed in:
```
 /metrics-flow/?utm_source=mozilla.org-whatsnew84&utm_medium=referral&utm_campaign=whatsnew84&entrypoint=mozilla.org-whatsnew84&utm_content=fpn-default&utm_term=fpn-default&entrypoint=fpn-site-visit&form_type=button
```

I'm going to test these strings locally and see if I can get an error. Both of the above `entrypoint` values pass our current regex

dannycoates (approved):
r+ if you still need this after looking at the existing logs

PR #7113, opened 2020-12-09
Rewire oauth

dannycoates:
This PR:
- remove `lib/oauth/routes`
- rewires tests to point to `lib/routes/oauth`
- rewires the server touchpoints to use `lib/routes/oauth`
- begins disassembling `lib/routes/oauth/proxied.js`
  - removes the `lib/oauthdb/*.js` route definitions and inlines joi validation in `proxied.js`
  - removes the `api` bits

PR #7204, opened 2021-01-04
refactor(oauth): add /config route back

dannycoates:
because token server still uses it

see https://github.com/mozilla/fxa/pull/7113#issuecomment-753505539


6a68 (approved):
LGTM. Huge thanks to @jackyzy823 for flagging this issue!

Issue #6532, opened 2020-09-23
Refresh buttons requery the entire account data instead of specific sections


data-sync-user:
See this thread: https://github.com/mozilla/fxa/pull/6520#discussion_r493668267

We’re using `useLazyQuery` for our refresh buttons which re-queries the entire account data. We should instead just query what was asked for and update the client cache with the new data, reducing the number of requests the `gql-api` must do per refresh.

At the time of writing we're only doing this in 3 files (Two Step Auth, Secondary Email, Recovery Key), so I'm marking it as a 3. If this is more work than it seems feel free to bump it up to a 5.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2570)
┆Issue Number: FXA-2570

PR #7217, opened 2021-01-06
refactor(settings): limit queried data for reloads

chenba:
Because:
 - querying everything for an account when reloading the state of
   recovery key, two step auth, or connected services incur unnecessary
   costs

This commit:
 - use reload queries that only query gql-api for what they need

The most impactful change is actually the 'keyFields' configuration for
Account in the InMemoryCache's 'typePolicies'.  It allows the query
results from the reloads to be automagically merged into the cache.
(The value of 'keyFields' is set to '[]', making the cached account a
singleton.)

However, setting 'keyFields' affected the `cache.modify` calls to update
the cached account.  The account passed into the field modifier function
is now a Reference, not a plain object, thus breaking the "clone object
then merged in updated properties" pattern.

Setting the proper `__typename` property is now also vital, at least for
Account.  Without it the cache will not able to merge updated propertie
into the cached account.  (gql-api already include `__typename` on its
results.)

Closes: #6532

### DIFF
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -32,7 +32,7 @@
     ]
   },
   "dependencies": {
-    "@apollo/client": "^3.2.2",
+    "@apollo/client": "^3.3.6",
     "@reach/router": "^1.3.4",
     "base32-decode": "^1.0.0",
     "base32-encode": "^1.1.1",
--- a/packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
+++ b/packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
@@ -3,12 +3,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-import { screen, fireEvent, act } from '@testing-library/react';
+import { screen, fireEvent, act, wait } from '@testing-library/react';
 import UnitRowRecoveryKey from '.';
 import {
   renderWithRouter,
   MockedCache,
-  mockAccountQuery,
+  mockRecoveryKeyExistsQuery,
 } from '../../models/_mocks';
 
 describe('UnitRowRecoveryKey', () => {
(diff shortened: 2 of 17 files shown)


packages/fxa-settings/package.json
```diff
@@ -32,7 +32,7 @@
     ]
   },
   "dependencies": {
-    "@apollo/client": "^3.2.2",
+    "@apollo/client": "^3.3.6",
```
chenba, line 35: Apollo docs:
> Prior to Apollo Client 3.3, readQuery would throw MissingFieldError exceptions to report missing fields. Beginning with Apollo Client 3.3, readQuery always returns null to indicate fields were missing.

At some point I was getting the MissingFiedError exception from some tests, but without a good way to find out why or where.  The code and tests have changed a bit since then, so I don't know if this upgrade is necessary, but I don't see any harm in keeping it.

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.test.tsx
```diff
@@ -47,20 +47,23 @@ describe('UnitRowRecoveryKey', () => {
...
+        </MockedCache>
+      );
+    });
     expect(
       screen.getByTestId('recovery-key-unit-row-header-value')
     ).toHaveTextContent('Not Set');
     await act(async () => {
       fireEvent.click(screen.getByTestId('recovery-key-refresh'));
     });
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
chenba, line 66: According to some doc, we need to wait a tick to get past the 'loading' state of the query.

jodyheavener: Alternatively, I think this could be turned into:

```tsx
waitFor(() => {
  expect(
    screen.getByTestId('recovery-key-unit-row-header-value')
  ).toHaveTextContent('Enabled');
});
```

chenba: I don't think we have `waitFor` since we are versions behind on whichever testing-library package has that func.

6a68: Huh. It looks like we do have `wait()`, the now-deprecated predecessor to `waitFor`

chenba: Heh, true.  I can make that change.

6a68 (approved):
Looks good to me.

Tested locally:
* add/remove recovery key in old settings, refresh new settings, works
* add/remove recovery key in new settings, auto-refreshes correctly
* add/remove 2FA in old settings, then refresh new settings, works
* add/remove 2FA in new settings, auto-refreshes correctly
* sign in / sign out from another browser, refresh new settings, works
* signing out another browser from new settings, auto-refreshes the list correctly

At a high level, I'm not convinced having a bunch of little different refresh buttons scattered around the settings page is good UX, and the cache-related changes here do introduce some complexity. That said, we don't have UX help right now to sort out an alternative refresh pattern, fetching less data should improve performance and reduce use of server resources, and reducing query surface area is something security folks would prefer.

:shipit: Let's ship it!

chenba:
> At a high level, I'm not convinced having a bunch of little different refresh buttons scattered around the settings page is good UX

Indeed, and there's already an issue filed for it.

chenba:
- Replaced the await a timeout in a promise line with `await wait()` on one of the tests, and added a comment explaining it.
- Fixed some typos in the commit message.

I'll merge once all checks pass.  Thank y'all for the comments/reviews/QA.

Issue #6562, opened 2020-09-28
[FxA-Settings] Newsletters button returns "Bad request"


vasilicatamas:
**Build**
Stage [Train 1.189.0]

**Affected platforms**
Windows 10 x64

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on “Newsletters” button.

**Expected results**
"Sign in to continue to Mozilla email preferences" page is displayed in a new tab.

**Actual Results**
- “Bad Request Invalid parameter: email” error is displayed.
- Url : https://accounts.firefox.com/authorization?client_id=c40f32fd2938f0b6&state=9d9226abd950f2fa95d26d15f51b63ad&redirect_uri=https%3A%2F%2Fbasket.mozilla.org%2Ffxa%2Fcallback%2F&scope=profile&email=vtamas%2089%40mozilla.com

![issue12](https://user-images.githubusercontent.com/18114210/94432291-3a089180-019f-11eb-98bb-28278d1d63d1.gif)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2578)
┆Issue Number: FXA-2578

data-sync-user:
➤ Chelsea Lewis commented:

I understand that Ops has resolved this due to the email-service config issue discussed in #fxa channel today. Please confirm fix or reopen if required.

vasilicatamas:
This issue is still reproducible and is not related to email-service config issue.

data-sync-user:
➤ Wil Clouser commented:

We think this is because the link in Settings Beta is pointing to production basket instead of staging basket.  This should be a quick fix.

data-sync-user:
➤ Danny Coates commented:

Not a quick fix because it requires making runtime changes to the external url. I'd prefer a wontfix and accept that it'll only work in prod

data-sync-user:
➤ Dave Justice commented:

Discussed this is last meeting, decided on a wont fix since this will work in prod

data-sync-user:
➤ Barry Chen commented:

This will be fixed for both stage and prod.

vasilicatamas:
This issue was verified as fixed on Stage [Train 1.198,1] and Prod [Train 1.197.1]  on Windows 10 64-bit, Android 11 and iOS 13.6. "Sign in to continue to Mozilla email preferences" page is successfully displayed in a new tab.

[...]

PR #7247, opened 2021-01-11
fix(settings): fix email communications link

chenba:
Because:
 - the Email Communications link on the nav was not working

This commit:
 - encode the email in the link
 - pass the configured email communication URL to the client
 - add a ConfigContext with a hook for accessing the config

Closes: #6562


dannycoates (approved):
LGTM! Thanks for adding the config context. the unencoded qs was 🤦

PR #7254, opened 2021-01-12
feat(auth-server): add paypal API base client

### LINKED ISSUE #7232: Create PayPal API base client
The PayPal API's we are using (NVP) do not have any mature clients around in the Node world, most likely due to their age and that few customers build with reference transactions. Before building a PayPalHelper, we must create a more basic PayPal NVP API Client.

This basic NVP API Client should do the following:

* Get instantiated with a logger, whether to use sandbox mode, and required API Signature credentials
* Have a private function to convert objects to NVP strings
* Have a private function to convert objects from NVP strings
* Create and make PayPal API requests

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2877)
┆Issue Number: FXA-2877

bbangert:
Because:

* We want to use the PayPal NVP API, and there isn't a suitable
  existing library.

This commit:

* Adds a base PayPal NVP API client for extending with additional
  methods.

Closes #7232

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-client.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-client.ts
@@ -1,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import querystring from 'querystring';
+import superagent from 'superagent';
+import pRetry from 'p-retry';
+
+export const PAYPAL_SANDBOX_API = 'https://api-3t.sandbox.paypal.com/nvp';
+export const PAYPAL_LIVE_API = 'https://api-3t.paypal.com/nvp';
+const PAYPAL_VERSION = '204';
+const L_LIST = /L_([A-Z]+)(\d+)$/;
+
+type PaypalOptions = {
+  sandbox: boolean;
+  user: string;
+  pwd: string;
+  signature: string;
+  retryOptions?: {
+    retries?: number;
+    minTimeout?: number;
+    factor?: number;
+  };
+};
+
+export type PAYPAL_METHODS =
+  | 'BillAgreementUpdate'
+  | 'CreateBillingAgreement'
+  | 'DoReferenceTransaction'
+  | 'GetTransactionDetails'
+  | 'SetExpressCheckout'
+  | 'TransactionSearch';
+
+export class PayPalClient {
+  private url: string;
+  private user: string;
+  private pwd: string;
+  private signature: string;
+  private retryOptions: {
+    retries: number;
+    minTimeout: number;
+    factor: number;
+  };
+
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -0,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import querystring from 'querystring';
+import superagent from 'superagent';
```
chenba, line 5: This jumped out at me on first glance because I know that auth-server already talk to other services.  But looks like those all buried behind `createBackendServiceAPI` (which in turn uses poolee), so this addition is 👍 with me.

(But if `createBackendServiceAPI` also works for this, then I'd prefer the consistency.)

packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -0,0 +1,109 @@
...
+import pRetry from 'p-retry';
+
+export const PAYPAL_SANDBOX_API = 'https://api-3t.sandbox.paypal.com/nvp';
+export const PAYPAL_LIVE_API = 'https://api-3t.paypal.com/nvp';
```
chenba, line 9: nice to have: these values as configs.  But also, I have no idea if they ever change.

packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -0,0 +1,109 @@
...
+type PaypalOptions = {
+  sandbox: boolean;
+  user: string;
+  pwd: string;
+  signature: string;
+  retryOptions?: {
+    retries?: number;
+    minTimeout?: number;
+    factor?: number;
+  };
```
chenba, line 22: This type is similar to the one for `retryOptions` on line 38.  Could define a type T for that and this can be `Partial<T>`.

chenba (approved):
LGTM

packages/fxa-auth-server/lib/payments/paypal-client.ts
bbangert: The API is deprecated, and hasn't changed in a decade or so.

packages/fxa-auth-server/lib/payments/paypal-client.ts
biancadanforth: Why would we not want to use `createBackendServiceAPI` here? Is it because this API is for PayPal servers rather than FxA servers (i.e. we don't need to declare routes here, but rather in the higher level PayPal Helper object in https://jira.mozilla.com/browse/FXA-2878)?

biancadanforth (commented):
I know this has already merged; just took a pass through it with some minor questions for you. It was great to pair as you wrote this.

Issue #5180, opened 2020-04-30
[delete] Log parity delete account events


data-sync-user:
Log delete account related events that are currently being logged in current settings.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1807)
┆Issue Number: FXA-1807

data-sync-user:
➤ Jared Hirsch commented:

The component already had logger calls added to it, so I don't think this issue was needed.

PR #7256, opened 2021-01-12
feat(fxa-settings): add delete account page

### LINKED ISSUE #4980: [delete] Conditionally show step 1/2 + checkbox validation
This page will need to conditionally display "step 1" of delete account or "step 2" depending on if the user has checked all of the checkboxes and clicked "continue" on the first page. Disable/enable the "Continue" button depending on if all checkboxes are checked, and on click of "Continue," render something else (dummy text or a shell step 2 component) in place. The step 2 component will be created in a follow up.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1674)
┆Issue Number: FXA-1674

### LINKED ISSUE #4981: [delete] Add the `delete_account` link/route, create page with step 1 of 2 dummy form
Place the link in the main settings component and add the route for `/delete_account`.

Transfer the functionality over around displaying products (and logic around hiding the container if 0 products, and 2 columns if over a certain number) from the current `delete_account` page.

Create the checkboxes. Allow "Cancel" to take the user back to Settings and leave "Continue" disabled, it'll be enabled in a follow up issue.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/408948620

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1673)
┆Issue Number: FXA-1673

### LINKED ISSUE #4983: [delete] Create step 2 with password input
A dummy step 2 component or text should be conditionally rendered for this flow if the user has checked the boxes and clicked "continue" in FXA-1674.

This issue builds out step 2. Display the copy, password input, and CTAs below.

The password input/eye functionality can be taken from the change password work.

The "Cancel" link should take the user back to Settings. Enable the "Delete account" button when the password input is not blank. Functionality will be done in a follow up.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/408948598

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1675)
┆Issue Number: FXA-1675

### LINKED ISSUE #4984: [delete] Process form response (REST)
This won't hit the GQL server. See FXA-2001 for details.

On form submission success, display the "Your account has been deleted." screen/text. Any errors should display on the "2 of 2" page in the global alert.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1676)
┆Issue Number: FXA-1676

meandavejustice:
## This pull request

- Add the delete account page to beta/settings
- small change to remove `errorTooltipClass` from `packages/fxa-settings/src/components/InputText/index.tsx` as it's not being used anywhere

- refs #5180
- fixes #4983
- fixes #4981
- fixes #4980
- fixes #4984

## Checklist

## Screenshots (Optional)

settings-page-delete-account-button
settings-page-delete-account-button-mobile
step-1
step-1-mobile
step-2-error
step-2-error-mobile
step-2-mobile
complete
complete-mobile

Opening this up before I finish tests and RTL adjustments so I can maybe get some feedback on spacing and style tweaks

### DIFF
--- a/packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
+++ b/packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
@@ -1,0 +1,216 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { useCallback, useState, ChangeEvent } from 'react';
+import { useForm } from 'react-hook-form';
+import { RouteComponentProps, useNavigate } from '@reach/router';
+import { useAccountDestroyer } from '../../lib/auth';
+import { sessionToken } from '../../lib/cache';
+import { useAlertBar } from '../../lib/hooks';
+import { useAccount } from '../../models';
+import InputPassword from '../InputPassword';
+import FlowContainer from '../FlowContainer';
+import VerifiedSessionGuard from '../VerifiedSessionGuard';
+import AlertBar from '../AlertBar';
+import {
+  HomePath,
+  LockwiseLink,
+  MonitorLink,
+  ROOTPATH,
+  VPNLink,
+} from '../../constants';
+import { logViewEvent, usePageViewEvent } from '../../lib/metrics';
+import { Checkbox } from '../Checkbox';
+
+type FormData = {
+  password: string;
+};
+
+const checkboxLabels = [
+  'Any paid subscriptions you have will be cancelled',
+  'Any may lose saved information and features within Mozilla products',
+  'Reactivating with this email may not restore your saved information',
+  'Any extensions and themes that you published to addons.mozilla.org will be deleted',
+];
+
+export const PageDeleteAccount = (_: RouteComponentProps) => {
+  usePageViewEvent('settings.delete-account');
+  const { handleSubmit, register, formState, setValue } = useForm<FormData>({
+    mode: 'all',
+    defaultValues: {
+      password: '',
+    },
+  });
+  // TODO: the `.errors.password` clause shouldn't be necessary, but `isValid` isn't updating
+  // properly. I think this is a bug in react-hook-form.
+  const disabled =
+    !formState.isDirty || !formState.isValid || !!formState.errors.password;
+  const [errorText, setErrorText] = useState<string>();
+  const [confirmed, setConfirmed] = useState<boolean>(false);
+  const [subtitleText, setSubtitleText] = useState<string>('Step 1 of 2');
+  const [checkedBoxes, setCheckedBoxes] = useState<string[]>([]);
+  const allBoxesChecked = checkboxLabels.every((element) =>
+    checkedBoxes.includes(element)
+  );
+  const navigate = useNavigate();
+  const alertBar = useAlertBar();
+  const goBack = useCallback(() => window.history.back(), []);
+
+  const { primaryEmail } = useAccount();
+
+  const advanceStep = () => {
+    setSubtitleText('Step 2 of 2');
+    setConfirmed(true);
+
+    logViewEvent('flow.settings.account-delete', 'terms-checked.success');
+  };
+
+  const deleteAccount = useAccountDestroyer({
+    onSuccess: () => {
+      // must use location.href over navigate() since this is an exernal link
+      window.location.href = `${ROOTPATH}?delete_account_success=true`;
+      logViewEvent('flow.settings.account-delete', 'confirm-password.success');
+    },
+    onError: (error) => {
+      if (error.errno === 103) {
+        setErrorText(error.message);
+        setValue('password', '');
+      } else {
+        alertBar.setType('error');
+        alertBar.setContent(error.message);
+        alertBar.show();
+        logViewEvent('flow.settings.account-delete', 'confirm-password.fail');
+      }
+    },
+  });
+
+  const handleConfirmChange = (labelText: string) => (
+    event: ChangeEvent<HTMLInputElement>
+  ) => {
+    event.persist();
+    setCheckedBoxes((existing) => {
+      if (event.target.checked) {
+        return [...existing, labelText];
+      } else {
+        return [...existing.filter((text) => text !== labelText)];
+      }
+    });
+  };
+
+  const onFormSubmit = ({ password }: FormData) => {
+    deleteAccount.execute(primaryEmail.email, password, sessionToken()!);
+  };
+
+  return (
+    <FlowContainer title="Delete Account" subtitle={subtitleText}>
+      {alertBar.visible && (
+        <AlertBar onDismiss={alertBar.hide} type={alertBar.type}>
+          <p data-testid="delete-account-error">{alertBar.content}</p>
+        </AlertBar>
+      )}
+      <VerifiedSessionGuard onDismiss={goBack} onError={goBack} />
+      {!confirmed && (
+        <div className="my-4 text-sm" data-testid="delete-account-confirm">
+          <p className="mb-4">
+            You've connected your Firefox account to Mozilla products that keep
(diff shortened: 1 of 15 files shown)


packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
```diff
@@ -0,0 +1,221 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React, { FormEvent, useCallback, useEffect, useState } from 'react';
```
jodyheavener, line 5: Use like `useEffect` is unused, as well as `cloneDeep` below:

Screen Shot 2021-01-13 at 10 19 30 AM

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+              </div>
+              <p className="mb-4">
+                Please acknowledge that by deleting your account:
+              </p>
+              <form
+                onChange={(evt: FormEvent<HTMLFormElement>) => {
+                  setHideContinueButton(
+                    evt.currentTarget.querySelectorAll('input:checked').length <
+                      4
+                  );
```
jodyheavener, line 131: This is an optional suggestion based on something I did over in Experimenter, but might yield an cleaner setup for these checkboxes:

- Set up [an array](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/FormRequestReview/index.tsx#L14-L17) of label strings.
- [Iteratively](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/FormRequestReview/index.tsx#L51-L64) render them (hooray DRY)
- When one is checked, [add/remove](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/FormRequestReview/index.tsx#L29-L40) it from an internal state
- On render, [compare](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/src/components/FormRequestReview/index.tsx#L25-L27) that internal state against the original array of labels to see if they match, and assign the result to `hideContinueButton`

I don't feel strongly about this, just a thought since it helps avoid repetition and needing to inspect the DOM.

packages/fxa-settings/src/components/PageSettings/index.tsx
```diff
@@ -32,6 +32,14 @@ export const PageSettings = (_: RouteComponentProps) => {
         <Profile />
         <Security />
         <ConnectedServices />
+        <div className="flex mx-4 tablet:mx-0">
+          <a
+            className="cta-caution cta-base text-sm transition-standard mt-12"
+            href="/beta/settings/delete_account"
```
jodyheavener, line 38: It looks like at some point we started using a [constants file](https://github.com/mozilla/fxa/blob/delete-account-page/packages/fxa-settings/src/constants/index.tsx#L5:L5) for paths, you might consider continuing with that (or using a base path constant) since at some point we're going to need to remove `beta/` from all these paths.

jodyheavener (commented):
For some reason I'm getting a GQL server CORS error so I can't view this new page locally, so I just left you a few minor thoughts on the code!

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+  };
+
+  return (
+    <>
+      {accountDeleted ? (
+        <FlowContainer title="Delete Account">
+          Your account has been deleted
+        </FlowContainer>
+      ) : (
+        <FlowContainer title="Delete Account" subtitle={subtitleText}>
```
vbudhram, line 85: Should distinct flows go into their own file or is it up to the developer to make that call? Having distinct files might make the code more readable, but probably introduces other problems.

packages/fxa-settings/src/components/PageSettings/index.tsx
vbudhram: Would you be able to use the `Link` component to do the navigation?

jodyheavener: Oh ya that's a good point 👍

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
meandavejustice: We don't have a decision on this, I followed `PageRecoveryKeyAdd` which has them all in the same document. Would be nice to have a short discussion on it

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+    setConfirmed(true);
+
+    logViewEvent('flow.settings.account-delete', 'terms-checked.success');
+  };
+
+  const deleteAccount = useAccountDestroyer({
+    onSuccess: () => {
+      setAccountDeleted(true);
+
+      logViewEvent('flow.settings.account-delete', 'confirm-password.success');
```
vbudhram, line 68: In the current settings page, when a user deletes their account, they are redirected to the email first login page. 

I think this is the correct behavior vs showing an "Account deleted" message, since they can't really do anything in the settings page. It is probably ok to fix in a follow up if it adds too much complexity to this PR.

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
vbudhram: Heh yea that would probably be good. 

Some quick thoughts of mine:

* Having all the views in one file make sense if its a wizard type page, except maybe if there are 10 sections?
* Maybe introduce a wizard page component
* Code is more readable in separate files but I don't believe our folder structure would scale well

vbudhram (approved):
@meandavejustice Just a couple questions about where to navigate after account delete, otw works as described 👍🏽

packages/fxa-settings/src/components/PageDeleteAccount/index.tsx
meandavejustice: I originally had a redirect, but I thought it may not be the best for accessibility. Seems like a a question for @LZoog

LZoog: If there's text that should be read like a confirmation message, yeah, a redirect isn't a11y friendly since a user might not see the message that the account was successfully deleted.

When you delete an account through old settings, it redirects you to the login page with the message "Account deleted successfully" there. That's totally acceptable since the user can see the message _after_ the redirect. Can we replicate similar behavior?

meandavejustice: Yeah that sounds good to me 👍

meandavejustice: Thanks @LZoog, After looking into how we show the "Account Deleted" message in content server after the redirect, I'm afraid we aren't going to be able to do it the same way. In the old settings we call `this.navigate` and pass the template through the backbone router in order to show the message on the signup page.

https://github.com/mozilla/fxa/blob/main/packages/fxa-content-server/app/scripts/views/settings/delete_account.js#L185

I'm thinking a workable solution would be to show the message "Your account has been deleted", as we currently are, but have a button to redirect to the `/` path. Though I'm not sure what the button copy should be 🤔

vbudhram: > I'm thinking a workable solution would be to show the message

You could also pass a flag via query params to the backbone page and let it display the message.

meandavejustice: That's a simple solution, I like that, I will add that so we can get this merged.

PR #7257, opened 2021-01-12
fix(oauth): fixed oauth -> auth error conversion after refactoring

dannycoates:
In anticipation of combining the auth/oauth error.js source files I prematurely removed the pre-existing logic to convert oauth errors into auth errors for certain routes. This restores that logic in the `appErrorFromOauthError` function and calls it as part of `AppError.translate` where appropriate.

I also renamed the AppError 'type' in oauth/error.js to OauthError to make them easier to differentiate.

This fixes https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/10391764

### DIFF
--- a/packages/fxa-auth-server/lib/error.js
+++ b/packages/fxa-auth-server/lib/error.js
@@ -213,6 +213,8 @@ AppError.translate = function (request, response) {
   }
   if (OauthError.isOauthRoute(request && request.route.path)) {
     return OauthError.translate(response);
+  } else if (response instanceof OauthError) {
+    return appErrorFromOauthError(response);
   }
   const payload = response.output.payload;
   const reason = response.reason;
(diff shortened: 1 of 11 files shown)


packages/fxa-auth-server/lib/error.js
```diff
@@ -1456,5 +1458,52 @@ function scrubHeaders(headers) {
   return scrubbed;
 }
 
+function appErrorFromOauthError(err) {
```
dannycoates, line 1461: This function previously live here: https://github.com/mozilla/fxa/blob/train-195/packages/fxa-auth-server/lib/oauthdb/utils.js#L15

6a68: I noticed the old one also logged an [`oauthdb.mapOAuthError` message](https://github.com/mozilla/fxa/blob/train-195/packages/fxa-auth-server/lib/oauthdb/utils.js#L68-L72) in the default case where the error's not found in the translation list. Did we ever look at those error logs in the past? Would it be useful to add a new logger call in the default case here?

6a68 (approved):
LGTM, I think you translated all the spots.

I did have one logging-related question, see comment.

Also, it is a little weird that oauth/util uses AppError, not OauthError, though that can wait for a followup.

packages/fxa-auth-server/lib/error.js
dannycoates: The switch cases are exhaustive as of today, but if we did add new oauth errnos in the future (which we shouldn't, OauthError should be deprecated and we should only be making new AppErrors) the default ought to be a pass-through, not an `unexpectedError`. So I'll change that default case.

PR #7258, opened 2021-01-13
feat(auth-server): add paypal helper

### LINKED ISSUE #7233: Create PayPalHelper base
The PayPalHelper is a helper class similar to the Stripe helper, that handles performing operations using the PayPal NVP API client created in FXA-2877. It should be instantiated in auth-server using TypeDI, with configuration setup similar to how the StripeHelper is instantiated.

This issue tracks the creation of the PayPalHelper module, the basic class skeleton that instantiates a PayPal NVP API Client, config variables, and its instantiation in auth-server's bin/key_server.js.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2878)
┆Issue Number: FXA-2878

bbangert:
Because:

* We want to have a helper for PayPal that takes a similar role as the
  StripeHelper.

This commit:

* Add's the PayPalHelper and wires in its configuration, initialization,
  and use in the subscription routes.

Closes #7233

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -794,6 +794,32 @@ const conf = convict({
       env: 'SUBSCRIPTIONS_ENABLED',
       default: false,
     },
+    paypalNvpSigCredentials: {
+      sandbox: {
+        doc: 'PayPal Sandbox mode',
+        format: Boolean,
+        env: 'PAYPAL_SANDBOX',
+        default: true,
+      },
+      user: {
+        doc: 'PayPal NVP API User name',
+        format: String,
+        default: 'user',
+        env: 'PAYPAL_NVP_USER',
+      },
+      pwd: {
+        doc: 'PayPal NVP API password',
+        format: String,
+        default: 'user',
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -794,6 +794,32 @@ const conf = convict({
...
+      user: {
+        doc: 'PayPal NVP API User name',
+        format: String,
+        default: 'user',
+        env: 'PAYPAL_NVP_USER',
+      },
+      pwd: {
+        doc: 'PayPal NVP API password',
+        format: String,
+        default: 'user',
```
chenba, line 813: heh not sure if you care about the actual value here, but this looks like an artifact of copy & paste.  Same with `signature` below.

chenba (approved):
👍

packages/fxa-auth-server/config/index.ts
bbangert: Oh, I did put that in on purpose just so there's some value so that it runs even if its not used.

Issue #6036, opened 2020-07-24
[l10n] Determine ftl file strategy


data-sync-user:
Best practices for fluent are to have separate ftl files for each section or area.  We should determine what that looks like.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2344)
┆Issue Number: FXA-2344

data-sync-user:
➤ Les Orchard commented:

Two things from payment-server that I think would eventually apply to any other React work:

  1. 👍 It might be nice to split the FTL up by route and/or major component. Could co-locate those files with other assets like styles & stories & tests so they're easy to find & associate with work in that neighborhood. Would need tweaks to scripting to collect all those FTLs and stick them in the right directory expected by Pontoon.
  2. 🤔 We currently only fetch one FTL resource on page load for the whole payments-server app. If we split that up, would we somehow need to have a smarter AppLocalizationProvider ( https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/lib/AppLocalizationProvider.tsx ) that somehow knows when to load appropriate FTL splits on demand? Could we somehow bundle them on build? Would we need to fetch all the split FTL resources?

Is there more context somewhere on these best practices that might address the above?

data-sync-user:
➤ Wil Clouser commented:

Thanks for the thoughts, Les.  [~axel@mozilla.com] and [~smalolepszy@mozilla.com] could give us tips on best practices

data-sync-user:
➤ Lauren Zugai commented:

I like the idea of splitting them up by routes or "major" components. If we get too granular with it we might run into issues with shared/reused components. For Settings, it could make sense to have a file for the main page (perhaps a separate file for header/ footer if it's pretty lengthy), and a new file for each flow/route (change password, display name, delete account, etc.).

It'd be nice if we could bundle them into one file on build... the alternatives seem like more of a pain.

data-sync-user:
➤ Les Orchard commented:

{quote}It'd be nice if we could bundle them into one file on build... the alternatives seem like more of a pain.{quote}

Yeah, I think what I'm kind of hinting around is - do these best practices come with tooling?

Or do we need to split up the FTL for Pontoon-side benefits and write a Webpack bundler plugin for FTL on build?

data-sync-user:
➤ Staś Małolepszy commented:

[...]

PR #7261, opened 2021-01-13
feat(l10n): Add initial fluent l10n for settings v2

vbudhram:
## Because

- Settings v2 needs to be localized

## This pull request

- Moves the [AppLocalizationProvider]() to fxa-react
- Updates payment server and settings server to use the provider
- Adds `settings.ftl` in locales/en-US
- Adds basic support for display name page localization
- Note that we will have to refactor this once we split the ftl files by route/components

Closes: #6036 

## Checklist

### DIFF
--- a/packages/fxa-react/lib/AppLocalizationProvider.tsx
+++ b/packages/fxa-react/lib/AppLocalizationProvider.tsx
@@ -3,14 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { FluentBundle, FluentResource } from '@fluent/bundle';
-import 'intl-pluralrules';
 import { negotiateLanguages } from '@fluent/langneg';
 import { LocalizationProvider, ReactLocalization } from '@fluent/react';
 import React, { Component } from 'react';
 
 import availableLocales from 'fxa-shared/l10n/supportedLanguages.json';
 
-const OTHER_EN_LOCALES = ['en-NZ', 'en-CA', 'en-SG', 'en-MY']
+const OTHER_EN_LOCALES = ['en-NZ', 'en-CA', 'en-SG', 'en-MY'];
 
 async function fetchMessages(baseDir: string, locale: string, bundle: string) {
   try {
@@ -41,9 +40,11 @@ async function createFluentBundleGenerator(
   bundles: Array<string>
 ) {
   const fetched = await Promise.all(
-    currentLocales.filter(l => !OTHER_EN_LOCALES.includes(l)).map(async (locale) => {
-      return { [locale]: await fetchAllMessages(baseDir, locale, bundles) };
-    })
+    currentLocales
+      .filter((l) => !OTHER_EN_LOCALES.includes(l))
+      .map(async (locale) => {
+        return { [locale]: await fetchAllMessages(baseDir, locale, bundles) };
+      })
   );
 
   const mergedBundle = fetched.reduce((obj, cur) => Object.assign(obj, cur));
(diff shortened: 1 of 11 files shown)


packages/fxa-settings/public/locales/en-US/settings.ftl
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+## branding
+project-brand = Firefox Accounts
+-brand-name-mozilla = Mozilla
+-brand-name-firefox = Firefox
+
+## Display name page
```
vbudhram, line 10: @chenba While building out parts of this I noticed that we may potentially have a bunch of smaller ftl files. If  flow reuses components (input, password, checkbox etc), would the ftl files live in those component folders.

chenba: The plan is to have ftl files next to the components where they are used.  This file (settings.ftl) will be a merged file of all those small ftl files, and not to be manually edited.  But it is the file that is served to the client (and copied to the l01n repo for the localizers), as you are doing in the PR.

packages/fxa-react/lib/AppLocalizationProvider.tsx
```diff
@@ -10,7 +10,7 @@ import React, { Component } from 'react';
 
 import availableLocales from 'fxa-shared/l10n/supportedLanguages.json';
 
-const OTHER_EN_LOCALES = ['en-NZ', 'en-CA', 'en-SG', 'en-MY']
```
chenba, line 13: LOL

packages/fxa-payments-server/src/lib/AppLocalizationProvider.test.js
```diff
@@ -1,142 +0,0 @@
...
-            <div>untranslated</div>
-          </Localized>
-        </main>
-      </AppLocalizationProvider>
-    );
-    await waitUntilTranslated();
-
-    expect(getByTestId('result')).toHaveTextContent('Hello ⁨US$1.23');
-  });
-});
```
chenba, line 142: This file is simply deleted and not moved?  Why is that?

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -61,22 +62,27 @@ export const App = ({ flowQueryParams }: AppProps) => {
...
-          <PageRecoveryKeyAdd path="/account_recovery" />
-          <PageSecondaryEmailAdd path="/emails" />
-          <PageSecondaryEmailVerify path="/emails/verify" />
-          <PageTwoStepAuthentication path="/two_step_authentication" />
-          <Page2faReplaceRecoveryCodes path="/two_step_authentication/replace_codes" />
-        </ScrollToTop>
-      </Router>
-    </AppLayout>
+    <AppLocalizationProvider
+      baseDir="/beta/settings/locales"
```
chenba, line 66: Best to use https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/constants/index.tsx#L5 instead of a literal here.

packages/fxa-react/lib/AppLocalizationProvider.test.js
```diff
@@ -7,12 +7,10 @@
 import IntlPolyfill from 'intl';
 Intl.NumberFormat = IntlPolyfill.NumberFormat;
```
chenba, line 8: Not really part of the issue for the PR, but we no longer need the polyfill as the node version we use now has that function.

chenba (approved):
👍  Looks good overall.  Got a nit about using the homepath constant.

PR #7280, opened 2021-01-15
Fix base-docker.sh print out of build error logs

enote-kane:
## Because

In case the `npx yarn install` command fails because of a module installation, the actual error output is stored in `/tmp/xfa-XYZ/build.log` files which are not easily accessible afterwards. This makes troubleshooting of the build process unnecessarily complicated and time consuming.

## This pull request

Fixes the already existing code to dump out these build logs whenever they exist and whenever the script `_scripts/base-docker.sh` exits (with whatever exit code).

During Docker image builds using the script `_scripts/build-builder.sh`, the resulting `artifacts/fxa-builder.log` will then also contain this output for further inspection.

Closes: n/a

## Checklist

### DIFF
--- a/_scripts/base-docker.sh
+++ b/_scripts/base-docker.sh
@@ -1,7 +1,5 @@
 #!/bin/bash -ex
 
-set -o pipefail
-
 DIR=$(dirname "$0")
 cd "$DIR/.."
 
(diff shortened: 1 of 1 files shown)


dannycoates (approved):
LGTM. thanks @enote-kane 😄

Issue #7253, opened 2021-01-12
[settings][l10n] Concatenate .ftl files on build


data-sync-user:
We'd like to keep .ftl files alongside the jsx/tsx in which they are used.  However, we don't wish to have localizers to work with many small ftl files, nor do we want to serve them to the client.

Please see FXA-2344 for more.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2893)
┆Issue Number: FXA-2893

PR #7283, opened 2021-01-15
chore(settings): use small .ftl files

chenba:
Because:
 - we want to use .ftl files that are located alongside the components
   in which they are used for better DX

This commit:
 - use Webpack to watch and merge small .ftl files into a combined file
   that's served to the client
 - add small .ftl files as examples

Closes: #7253

### DIFF
--- a/packages/fxa-settings/.license.header
+++ b/packages/fxa-settings/.license.header
@@ -1,0 +1,3 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/packages/fxa-settings/.rescriptsrc.js
+++ b/packages/fxa-settings/.rescriptsrc.js
@@ -2,10 +2,56 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const {
-  permitAdditionalJSImports,
-} = require('fxa-react/configs/rescripts');
+const path = require('path');
+const fs = require('fs');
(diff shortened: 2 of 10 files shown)


packages/fxa-settings/.license.header
```diff
@@ -0,0 +1,3 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
chenba, line 1: Not sure about the location and name of this file, but gotta put it somewhere. 🤷

packages/fxa-settings/.rescriptsrc.js
```diff
@@ -2,10 +2,56 @@
...
+        writeToDisk: (path) =>
+          /public\/locales\/\S+.ftl/.test(path) || oldWriteToDisk(path),
+      };
+
+      return newConfig;
+    },
+    webpack: (config) => {
+      let newConfig = { ...config };
+
+      if (!newConfig.output.path) {
```
chenba, line 38: When the output path is undefined, nothing is written to disk.

chenba:
Since this watches for file changes _and write out to a file that's tracked by git_, if you have `fxa-settings` running while executing git commands, this might interfere!  Unfortunately this is the only solution that gives us the file change _and_ auto reload; otherwise we need manually refresh the page to see `.ftl` changes (the compilation is still auto).  If we are willing to give up auto reload on `.ftl` saves, then we can try something else.

vbudhram (approved):
@chenba I think we can probably drop a little note in the readme about localization, otherwise LGTM 👍🏽

packages/fxa-settings/.license.header
chenba: I'll file a bug for potentially updating Payments to use the same approach, and we can centralize stuff there.

chenba:
Good point.  I'll file a bug in the ecosystem docs repo and assign it to myself.  Thanks!

Issue #7271, opened 2021-01-14
Checkout: Add config toggle to disable PayPal UI in payments server


data-sync-user:
Similar to [FXA-2900](https://jira.mozilla.com/browse/FXA-2900) which covers the back-end, there should be a feature flag for PayPal on the front-end.

This will likely go inside [packages/fxa-payments-server/server/config/index.js](https://github.com/mozilla/fxa/blob/b9588258d4c6a8048c9abdf2f2f924948c8e5807/packages/fxa-payments-server/server/config/index.js#L15).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2903)
┆Issue Number: FXA-2903

Issue #7097, opened 2020-12-08
Checkout: Add PayPal payment option on the Subscribe page if there is no saved payment information


data-sync-user:
Edit: This ticket covers creating a React component for the PayPal button. A separate ticket will handle the functionality for when the button is clicked.

This will be in the form of a "Pay with PayPal" button which will be provided by the [PayPal JavaScript SDK](https://developer.paypal.com/docs/business/javascript-sdk/). When there is no saved payment information, we should provide the option to pay either with Stripe or PayPal.

Open question(s):

* Are there any legal notice or other links we need to add to this page? E.g. for Stripe form, Stripe has a Privacy Policy link.

From Barry:

* Currently, the thing we check for is the [`last4` property on the `Customer` object](https://github.com/mozilla/fxa/blob/44bbb161d958f084c1bb39902554f69a2333de90/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L87). We need to update this to check the metadata for the PayPal Billing Agreement ID.
* We also need to modify the backend a bit to send that new metadata back.
* Importantly, anything we want to display to the user about re: their payment information would be added to the `Customer` object.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2800)
┆Issue Number: FXA-2800

data-sync-user:
➤ Barry Chen commented:

I think the important bit for this and FXA-2081 is the data model around the payment provider/methods, which is returned from https://github.com/mozilla/fxa/blob/main/packages%2Ffxa-auth-server%2Flib%2Froutes%2Fsubscriptions.ts#L373. ( https://github.com/mozilla/fxa/blob/main/packages%2Ffxa-auth-server%2Flib%2Froutes%2Fsubscriptions.ts#L373. )  Currently we simply attach some CC details directly on the returned object (e.g. customer.last4 is the last four CC digits), but I think we'd want another level of abstraction.

data-sync-user:
➤ Barry Chen commented:

We could break this issue into two, with the first updating the code to use the new customer object (see previous comment).  Once we have Stripe working with the new data model, then we'll add PayPal support as the second issue.

data-sync-user:
➤ Bianca Danforth commented:

For reference, see PayPal Checkout Flow Implementation ( https://docs.google.com/document/d/1j4qGzUK0IOBOh8VeB0yHt_ZEHAez3XNXKytrF6ZNAsw/edit# ) created by Ben.

[...]

PR #7284, opened 2021-01-16
feat(payments): Add feature flag for PayPal UI

biancadanforth:
## Because

- We want to implement Paypal support without deploying to production until it's ready, and it's a good idea in general to put new features behind a flag in case there are unforeseen problems.

## This pull request

- Adds a PayPal feature flag for the front-end that is disabled by default.
- Currently, an empty `<div>` is rendered in the `SubscriptionCreate` component if the flag is enabled.

Closes: #7271

## Checklist

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -20,6 +20,13 @@ const conf = convict({
       env: 'FEATURE_USE_SCA_PAYMENT_UI_BY_DEFAULT',
       format: Boolean,
     },
+    usePaypalUIByDefault: {
+      default: false,
+      doc:
+        'Whether to use PayPal payment UI by default rather than Stripe payment UI alone',
+      env: 'FEATURE_USE_PAYPAL_UI_BY_DEFAULT',
+      format: Boolean,
+    },
   },
   amplitude: {
     enabled: {
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
@@ -26,6 +26,8 @@ import waitForExpect from 'wait-for-expect';
 
 import SubscriptionCreate, { SubscriptionCreateProps } from './index';
 
+import { updateConfig } from '../../../lib/config';
+
 // TODO: Move to some shared lib?
 const deepCopy = (object: Object) => JSON.parse(JSON.stringify(object));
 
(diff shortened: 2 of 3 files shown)


packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -146,6 +148,21 @@ describe('routes/ProductV2/SubscriptionCreate', () => {
       queryAllByText('30-day money-back guarantee')[0]
     ).toBeInTheDocument();
     expect(queryByText('Billing Information')).toBeInTheDocument();
+    expect(
+      document.getElementById('paypal-button-container')
```
biancadanforth, line 152: I didn’t know how else to get the div#paypal-button-container element from the page, so I just used `document.getElementById`. Open to better approaches.

packages/fxa-payments-server/server/config/index.js
```diff
@@ -20,6 +20,13 @@ const conf = convict({
       env: 'FEATURE_USE_SCA_PAYMENT_UI_BY_DEFAULT',
       format: Boolean,
     },
```
biancadanforth, line 22: It looks like `useSCAPaymentUIByDefault` (the only other feature flag whose [patch](https://github.com/mozilla/fxa/commit/26266e48afbb20bc5cf6a35cca433666d0437aaa) I referenced for this patch) isn’t used anymore in `App.tsx`. Should it be removed? Happy to file a bug for that.

biancadanforth: In general, there's a [TODO comment for how this feature flag was added](https://github.com/mozilla/fxa/blob/c50d1175d522abd793cdb7daef0e43680c6d38d6/packages/fxa-payments-server/server/lib/server.js#L55)... -- how would this be improved, and when would be the appropriate time to add this improvement?

biancadanforth (commented):
Just a couple of minor questions mostly related to possible follow-ups.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
chenba: We use a data-testid attr.  See line 145 above.

packages/fxa-payments-server/server/config/index.js
chenba: Yeah, let's remove `useSCAPaymentUIByDefault` if it's not used.

As for feature flag improvements, it's not something we've discussed yet.  And we don't have any redis based feature flags afaik.

chenba (approved):
💵

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
biancadanforth: Thanks; I do ultimately need an `id` attribute to render the PayPal button using the PayPal page script (#7097 ), so I think I'll leave this here for now. I can always update the test with the final UI components.

PR #7292, opened 2021-01-20
chore(docs): Replace npm with yarn commands in payments README

biancadanforth:
## Because

- FxA root uses yarn workspaces which npm LTS doesn't currently support. Also test commands had extra '--' in them that caused them to fail.

## This pull request

- Replaces `npm run` with `yarn` and modifies testing commands, so that they work.

No issue on file

## Checklist


jodyheavener (approved):
LGTM. Guess we should do all the other packages soon too.

PR #7316, opened 2021-01-21
fix(metrics): fixed 'fxa - activity' metric

### LINKED ISSUE #7298: Significant change in MAU reporting with Train 197
We pushed Train 197 out on Jan 13th.  We immediately saw a spike in activity reported from the USA with a minor dip from all other countries.  Today it is still 3x higher than it was a week ago.  A graph can be seen here:

https://analytics.amplitude.com/mozilla-corp/chart/sxp5s3v/edit/en89d2j

Investigation is underway

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2917)
┆Issue Number: FXA-2917

dannycoates:
The token verify metric began including location data in 196 because it started using the auth-server metrics code path. Although seemingly a benign addition, due to how the location got used further down the ETL pipeline it got integrated into the rollup 'fxa - activity' event unexpectedly. Since a large portion of the verify events originate at the profile server this overwrote the user's actual location with our server's geoip.

This commit adds a `minimal` option to the amplitude metrics code that only logs the data originally submitted before 196 for the two oauth specific metrics.

fixes #7298

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/amplitude.js
+++ b/packages/fxa-auth-server/lib/metrics/amplitude.js
@@ -107,10 +107,12 @@ const EVENTS = {
   'token.created': {
     group: GROUPS.activity,
     event: 'access_token_created',
+    minimal: true,
   },
   'verify.success': {
     group: GROUPS.activity,
     event: 'access_token_checked',
+    minimal: true,
   },
 };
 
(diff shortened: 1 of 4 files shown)


chenba (approved):
📉

vbudhram (approved):
@dannycoates Nice find!

PR #7321, opened 2021-01-21
add / capture avatar pages

### LINKED ISSUE #5024: [avatar] Create "take photo" retake/cancel/save page + functionality
Once the photo has been taken, the user can "Retake photo," "Cancel," or "Save".

"Retake photo" takes the user back to the first "take photo" page. "Cancel" takes the user back to the first step (choosing upload or take photo) and "Save" will be handled in another issue.

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/407953640

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1701)
┆Issue Number: FXA-1701

### LINKED ISSUE #5025: [avatar] Create "Take photo" page
Request camera access and allow taking the photo with the button shown.

The "back" button should take the user back to step 1.

Request by UX: can we force trigger the "allow access to camera" message from FF if the user has this disabled?

See https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/407710955

The screen it takes the user to after a successful upload (https://mozilla.invisionapp.com/share/PTUIT2C4N8Z#/screens/407953640) will be done in a follow up issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1700)
┆Issue Number: FXA-1700

meandavejustice:
## Because

- We need to allow the user to add and capture an new avatar for parity with old settings

## This pull request

- Refactors the structure of the Avatar components
- Adds the Capture avatar page, route, and functionality

- fixes #5024
- fixes #5025

## Checklist

## Screenshots (Optional)

add-avatar-page
capture-page-1
capture-page-2

### DIFF
--- a/packages/fxa-settings/src/components/App/index.tsx
+++ b/packages/fxa-settings/src/components/App/index.tsx
@@ -26,7 +26,9 @@ import { HomePath } from '../../constants';
 import { useConfig } from 'fxa-settings/src/lib/config';
 import { observeNavigationTiming } from 'fxa-shared/metrics/navigation-timing';
 import AppLocalizationProvider from 'fxa-react/lib/AppLocalizationProvider';
-import PageAvatar from '../PageAvatar';
+import PageAddAvatar from '../PageAvatar/Add';
+import PageChangeAvatar from '../PageAvatar/Change';
+import PageCaptureAvatar from '../PageAvatar/Capture';
 
 export const GET_INITIAL_STATE = gql`
   query GetInitialState {
@@ -73,7 +75,9 @@ export const App = ({ flowQueryParams }: AppProps) => {
           <ScrollToTop default>
             <PageSettings path="/" />
             <PageDisplayName path="/display_name" />
-            <PageAvatar path="/avatar/change" />
+            <PageAddAvatar path="/avatar/add" />
+            <PageChangeAvatar path="/avatar/change" />
+            <PageCaptureAvatar path="/avatar/capture" />
             <PageChangePassword path="/change_password" />
             <PageRecoveryKeyAdd path="/account_recovery" />
             <PageSecondaryEmailAdd path="/emails" />
--- a/packages/fxa-settings/src/components/FlowContainer/index.tsx
+++ b/packages/fxa-settings/src/components/FlowContainer/index.tsx
@@ -24,7 +24,7 @@ export const FlowContainer = ({
 }: FlowContainerProps & RouteComponentProps) => {
   return (
     <div
-      className="max-w-lg mx-auto mt-6 p-6 pb-7 tablet:my-10 flex flex-col items-start bg-white shadow tablet:rounded-xl"
+      className={`max-w-lg mx-auto mt-6 p-6 pb-7 tablet:my-10 flex flex-col items-start bg-white shadow tablet:rounded-xl`}
       data-testid="flow-container"
     >
       <Head title={title} />
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -50,6 +50,7 @@
     "react-dom": "^16.13.1",
     "react-hook-form": "^6.11.0",
     "react-scripts": "^3.4.1",
+    "react-webcam": "^5.2.2",
     "subscriptions-transport-ws": "^0.9.18",
     "typescript": "3.9.7"
   },
--- a/packages/fxa-settings/src/components/AvatarCropper/index.tsx
+++ b/packages/fxa-settings/src/components/AvatarCropper/index.tsx
@@ -3,12 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-
-// import Cropper from 'react-cropper';
+//import Cropper from 'react-cropper';
 import 'cropperjs/dist/cropper.css';
-// import { ReactComponent as ZoomIn } from './zoom-in.svg';
-// import { ReactComponent as ZoomOut } from './zoom-out.svg';
-// import { ReactComponent as Rotate } from './rotate.svg';
+//import { ReactComponent as ZoomIn } from './zoom-in.svg';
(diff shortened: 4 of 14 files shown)


packages/fxa-settings/src/components/FlowContainer/index.tsx
```diff
@@ -17,14 +18,15 @@ type FlowContainerProps = {
 };
 
 export const FlowContainer = ({
+  width = 'lg',
```
6a68, line 21: I'm curious why you needed to do this here

6a68 (changes requested):
This looks great so far! I'm able to take an image on mac / nightly.

Some bigger picture issues to address:

1. The resulting image looks pretty small and blocky (see screenshot), would be good to tweak the react-webcam config values to get image quality parity with current settings.

Screen Shot 2021-01-21 at 11 16 04 AM


2. I noticed that this page doesn't show a spinner while the webcam browser prompt is displayed. We can get to parity by throwing a `LoadingSpinner` component in there while we're waiting. Here it is in old settings, for comparison:

Screen Shot 2021-01-21 at 11 05 00 AM

3. This page doesn't show an error message if the webcam can't be initialized. Here is the current settings page error (it could definitely be improved, but it's at least some visual indication that something's gone wrong):

Screen Shot 2021-01-21 at 10 54 28 AM

I had a couple other high-level questions, particularly about managing multiple screens from one URL versus several.

Happy to dig into details a bit more once the high-level stuff is addressed; ping me for a followup.

packages/fxa-settings/src/components/FlowContainer/index.tsx
meandavejustice: I replaced the default width class so that we can pass different size options into FlowContainer, setting `lg` as the default since other components likely depend on that

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -73,7 +75,9 @@ export const App = ({ flowQueryParams }: AppProps) => {
           <ScrollToTop default>
             <PageSettings path="/" />
             <PageDisplayName path="/display_name" />
-            <PageAvatar path="/avatar/change" />
+            <PageAddAvatar path="/avatar/add" />
+            <PageChangeAvatar path="/avatar/change" />
+            <PageCaptureAvatar path="/avatar/capture" />
```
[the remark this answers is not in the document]
meandavejustice: I agree, I think that is the better approach, this was easier for organizing the tasks in my head, but I think it makes sense to factor into the same flow pattern once we achieve feature parity https://jira.mozilla.com/browse/FXA-2940

packages/fxa-settings/src/components/FlowContainer/index.tsx
6a68: I'm a bit confused by this. Why would we want the width of this page to be different from the width of all the other pages?

6a68 (changes requested):
OK, this is getting closer!

The image size should be 200x200, based on the old avatar behavior. For some reason, the smoothing doesn't seem to be getting applied.

I don't quite understand why the avatar page is narrower than all the other pages. Is there a reason we couldn't just leave it at the default PageContainer size?

If the camera can't be initialized, I think we should hide the 'take photo' button and actually set the 'Save' button to be disabled, not just set the disabled classname on it.

Other than that, I think we're pretty much good to go. I'll do one last pass to squint at spacing and compare the mocks.

meandavejustice:
> for some reason, the smoothing doesn't seem to be getting applied.

I bumped up the image quality setting for react-cropper, but I also compared to the avatar capture in the old settings. In old settings we actually close the modal as soon as you capture the image, so you can't see the quality difference between the media stream and the capture taken. Though the quality of the capture appears the same to me between old and new settings. I think the only reason we are noticing now is because of the new behavior, I will keep poking around at the underlying cropper library and see if there are more things to tweak.

meandavejustice:
> The image size should be 200x200, based on the old avatar behavior.

I've been basing this off of the figma designs, the avatars in figma are 128 x 128 
image

> If the camera can't be initialized, I think we should hide the 'take photo' button and actually set the 'Save' button to be disabled, not just set the disabled classname on it.

👍 These changes are coming shortly.

> I don't quite understand why the avatar page is narrower than all the other pages. Is there a reason we couldn't just leave it at the default PageContainer size?

I noticed that we too wide for the design in our widths while doing the avatar page. Avatar page made it more apparent how off we are with how out of proportion it looked.

Figma width (416)
image

Medium width (avatar view)
image

Large width (what we have for most current views)
image

Not sure how we should handle this, probably a follow up issue to change the page widths to `md`, and make sure it still looks okay on those pages/doesn't break mobile views

6a68 (approved):
This is good to land once the final few things are handled.

* Save button behavior when disabled:

> I am setting the disabled attribute, what other behavior are you expecting here? A different cursor?

Yeah, we set the `cursor: not-allowed` style in the old view when the camera's not available and you mouse over the 'save' button. I'm just going by the old settings behavior, because I didn't see anything about button hover states in new-settings Figma.

* Panel width:

> I noticed that we too wide for the design in our widths while doing the avatar page. Avatar page made it more apparent how off we are with how out of proportion it looked.

> Not sure how we should handle this, probably a follow up issue to change the page widths to md, and make sure it still looks okay on those pages/doesn't break mobile views

Good point, I hadn't noticed that. For now, I would stick with one consistent layout for all the panels, so leave this at the same width as the others, and file a followup issue to get a final UX review at the end (including looking at the panel widths).

* Save button should initially be disabled

Another little thing: I just noticed that the 'Save' button should be disabled when the form initially loads, or if you haven't uploaded an image yet. This is also in keeping with the behavior of the other new-settings panels.

meandavejustice:
Filed https://jira.mozilla.com/browse/FXA-2946 for widths

meandavejustice:
Added requested changes, will merge when checks finish

Issue #5066, opened 2020-04-22
[l10n] Delete account flow translations


data-sync-user:
Add l10n support for the delete account separate page and flow.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1735)
┆Issue Number: FXA-1735

PR #7328, opened 2021-01-21
feat(l10n): setup localization for delete account page

vbudhram:
## Because

- Delete account page should be localized

## This pull request

- Adds localization..

Closes: #5066

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageDeleteAccount/en-US.ftl
+++ b/packages/fxa-settings/src/components/PageDeleteAccount/en-US.ftl
@@ -1,0 +1,32 @@
+## Delete account page
+delete-account-header =
+ .title = Delete Account
+
+delete-account-step-1-2 = Step 1 of 2
+delete-account-step-2-2 = Step 2 of 2
+
+delete-account-confirm-title = You've connected your Firefox account to Mozilla products that keep you secure and productive on the web:
+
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/PageDeleteAccount/en-US.ftl
```diff
@@ -0,0 +1,32 @@
...
+ .title = Delete Account
+
+delete-account-step-1-2 = Step 1 of 2
+delete-account-step-2-2 = Step 2 of 2
+
+delete-account-confirm-title = You've connected your Firefox account to Mozilla products that keep you secure and productive on the web:
+
+mozilla-vpn = Mozilla VPN
+firefox-lockwise = Firefox Lockwise
+firefox-monitor = Firefox Monitor
```
chenba, line 12: nit and maybe we don't need it right now, but having a prefix- for these more generic ids would be a good preventive measure against id clashing in the combined ftl.

chenba (approved):
:shipit:

packages/fxa-settings/src/components/PageDeleteAccount/en-US.ftl
vbudhram: Yea, I wasn't sure how to handle this because I can see some generic/branding text being used all over.

PR #7332, opened 2021-01-21
chore(docs): Add missing package README links

LZoog:
Was passing by and noticed some packages were missing links to their docs.

miss u fxa


6a68 (approved):
👋 Thanks!

Issue #5058, opened 2020-04-22
[l10n] Header, navigation, footer translations


data-sync-user:
Add l10n support for the primary header, navigation, and footer components.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1727)
┆Issue Number: FXA-1727

data-sync-user:
➤ Barry Chen commented:

The Nav l10n will be done as part of FXA-2893.

PR #7335, opened 2021-01-22
localize settings header and footer

6a68:
## Because

- we want to localize the header and footer

## This pull request

- localizes the header and footer :-)

- Note: I feel a little weird about putting the strings for the footer (which lives in fxa-react) over in the fxa-settings App component's ftl file. This is in keeping with the approach from #7327 and I can't think of a better approach, but other suggestions welcome!

- [Update: regular casing restored.] ~~Also note: I've used SpOnGeBoB cAsInG to convince myself that the l10n transforms are actually getting applied. But nobody else is doing that in the other l10n commits, so I can revert to normal casing if that's preferred.~~ 

Closes: #5058

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-settings/public/locales/en-US/settings.ftl
+++ b/packages/fxa-settings/public/locales/en-US/settings.ftl
@@ -7,6 +7,12 @@
 app-default-title = Firefox Accounts
 app-page-title = { $title } | Firefox Accounts
 
+# Application-wide footer
+
+app-footer-mozilla-logo-label = Mozilla logo
+app-footer-privacy-notice = Website Privacy Notice
+app-footer-terms-of-service = Terms of Service
+
 # HeaderLockup component
 
 header-menu-open = Close menu
--- a/packages/fxa-settings/public/locales/en-US/settings.ftl
+++ b/packages/fxa-settings/public/locales/en-US/settings.ftl
@@ -7,6 +7,17 @@
 app-default-title = Firefox Accounts
 app-page-title = { $title } | Firefox Accounts
 
+# HeaderLockup component
+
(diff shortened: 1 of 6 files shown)


LZoog:
> Also note: I've used SpOnGeBoB cAsInG to convince myself that the l10n transforms are actually getting applied. But nobody else is doing that in the other l10n commits, so I can revert to normal casing if that's preferred.

FWIW I vaguely remember @meandavejustice doing this in Payments (I think?) a while back for the same reason, and @clouserw pushing back on the off chance something fails and an end-user sees the weird casing.

6a68:
Cool, thanks @LZoog! I think Dave and I both picked up that habit from m&m when we were working on lockbox

packages/fxa-settings/public/locales/en-US/settings.ftl
```diff
@@ -7,6 +7,23 @@
...
+app-footer-privacy-notice = Website Privacy Notice
+app-footer-terms-of-service = Terms of Service
+
+# HeaderLockup component
+
+header-menu-open = Close menu
+header-menu-closed = Site navigation menu
+header-back-to-top-link =
+  .title = Back to top
+header-title = Firefox accounts
```
meandavejustice, line 22: This should be singular `Firefox account` according to figma

meandavejustice (changes requested):
* One comment on translation copy.
* need to remove `packages/fxa-settings/package-lock.json`

6a68:
Thanks for the review! Updated, re-pushed, ready for another look

chenba (approved):
WFM

PR #7337, opened 2021-01-22
feat(shared): add paypal BA model commands

### LINKED ISSUE #7032: Create database model for new PayPal table
Similar to the [Accounts-Customer model|https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/db/models/auth/account-customers.ts
], we need to create a database model for the new PayPal table.

This would live in fxa-shared. This model is intended to read/write to the new PayPal table; it should have a set of CRUD operations.

Some queries for the new PayPal table that this model should support:

* Ask for all PayPal Billing Agreements for a user.
* Change/update the status of a Billing Agreement for a user (comes from PayPal Webhooks)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2761)
┆Issue Number: FXA-2761

bbangert:
Because:

* We want to track billing agreement IDs used by customers using PayPal
  over time, and Stripe metadata is only used to track the currently
  active billing agreement ID to use.

This commit:

* Adds model commands and a model to use the existing paypalCustomers
  table that handle basic CRUD operations.

Closes #7032

## Checklist

### DIFF
--- a/packages/fxa-shared/db/models/auth/index.ts
+++ b/packages/fxa-shared/db/models/auth/index.ts
@@ -6,10 +6,18 @@ import { Account } from './account';
 import { AccountCustomers } from './account-customers';
 import { AuthBaseModel } from './auth-base';
 import { SessionToken } from './session-token';
+import { PayPalBillingAgreements } from './paypal-ba';
 
 export type AccountOptions = {
   include?: 'emails'[];
 };
+
+export type PayPalBillingAgreementStatusType =
+  | 'Pending'
+  | 'Active'
+  | 'Suspended'
+  | 'Cancelled'
+  | 'Expired';
 
 export async function sessionTokenData(
   tokenId: string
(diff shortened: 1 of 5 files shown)


packages/fxa-shared/db/models/auth/index.ts
```diff
@@ -128,6 +129,72 @@ export async function deleteAccountCustomer(uid: string) {
...
+ *
+ * @param uid
+ * @param billingAgreementId
+ * @param status
+ * @param endedAt
+ */
+export async function updatePayPalBA(
+  uid: string,
+  billingAgreementId: string,
+  status: string,
```
chenba, line 175: Would be nice to have a type for this so not to allow any string.

packages/fxa-shared/db/models/auth/paypal-ba.ts
```diff
@@ -0,0 +1,46 @@
...
+  public static idColumn = 'uid';
+
+  protected $uuidFields = ['uid'];
+
+  public uid!: string;
+  public billingAgreementId!: string;
+  public status!: string;
+
+  public createdAt!: number;
+  public endedAt!: number;
```
chenba, line 20: This can be null in the db.  If we make the type here `number | null`, would the ORM be smart enough to set it to null?  (This depends on how you feel about interpreting 0 as null...assuming the value here is 0 when it's null in the db record.)

chenba (approved):
💰

packages/fxa-shared/db/models/auth/paypal-ba.ts
bbangert: Ah, good poing!

Issue #5067, opened 2020-04-22
[l10n] Two step authentication flow translations


data-sync-user:
Add l10n support for the two step authentication (2FA) separate page and flow.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1734)
┆Issue Number: FXA-1734

data-sync-user:
➤ Barry Chen commented:

A few of the strings will be added as part of FXA-2893.

Issue #7322, opened 2021-01-21
Add localization for common password link


vbudhram:
From https://github.com/mozilla/fxa/pull/7290/files#r561961270, we should add localization to the common password link.

Fluent react does support having html elements in the ftl file but it wasn't obvious to me how this would look/work with a react component `LinkExternal`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2935)
┆Issue Number: FXA-2935

PR #7341, opened 2021-01-22
chore(settings): add l10n to two step auth

chenba:
Because:
 - we should set up l10n on new settings

This commit:
 - update two step auth to use localized strings

Closes: #5067

### DIFF
--- a/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
+++ b/packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
@@ -19,7 +19,7 @@ import { checkCode, getCode } from '../../lib/totp';
 import { HomePath } from '../../constants';
 import { alertTextExternal } from '../../lib/cache';
 import { logViewEvent, useMetrics } from '../../lib/metrics';
-import { Localized } from '@fluent/react';
+import { Localized, useLocalization } from '@fluent/react';
 
 export const metricsPreInPostFix = 'settings.two-step-authentication';
 
@@ -51,6 +51,8 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
   const goBack = () => window.history.back();
   const goHome = () => navigate(HomePath, { replace: true });
 
+  const { l10n } = useLocalization();
+
   const totpForm = useForm<TotpForm>({
     mode: 'onTouched',
   });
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -211,37 +240,55 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
-              </p>
+              <Localized
+                id="tfa-scan-this-code"
+                elems={{
+                  linkExternal: (
+                    <LinkExternal
+                      className="link-blue"
+                      href="https://support.mozilla.org/kb/secure-firefox-account-two-step-authentication"
+                    >
+                      {' '}
```
chenba, line 251: Just here to prevent a syntax error.  Actual content will be replaced.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -144,14 +165,22 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
     onCompleted: () => {
-      alertTextExternal('Two-step authentication enabled');
+      alertTextExternal(
+        l10n.getString('tfa-enabled', null, 'Two-step authentication enabled')
+      );
       goHome();
     },
     onError: (err) => {
       if (err.graphQLErrors?.length) {
         setRecoveryCodeError(err.message);
```
vbudhram, line 175: You should be able to set localized error

```
const localizedError = l10n.getString(`auth-error-${error.errno}`);
```

Although, now that I take another maybe we can have graphql set this.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -197,7 +226,7 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
 
   return (
     <FlowContainer
-      title="Two Step Authentication"
+      title={l10n.getString('tfa-title')}
```
vbudhram, line 229: Oh this makes more sense than what I was trying to do, passing in the title as a attr property.

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
vbudhram: Heh this is exactly what is needed for https://github.com/mozilla/fxa/issues/7322

vbudhram (approved):

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
chenba: Oops missed a spot.  Good catch.  Also, I don't know if gql-api gives us the errno from the original error; I'll double check.

Issue #5062, opened 2020-04-22
[l10n] Recovery key flow translations


data-sync-user:
Add l10n support for the recovery key separate page and flow.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1733)
┆Issue Number: FXA-1733

PR #7346, opened 2021-01-23
Localize the Recovery Key page

6a68:
## Because

- we want to localize the Recovery Key page

## This pull request

- localizes the Recovery Key page
- adds en-US strings

Closes: #5062 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-settings/public/locales/en-US/settings.ftl
+++ b/packages/fxa-settings/public/locales/en-US/settings.ftl
@@ -13,6 +13,16 @@ app-footer-mozilla-logo-label = Mozilla logo
 app-footer-privacy-notice = Website Privacy Notice
 app-footer-terms-of-service = Terms of Service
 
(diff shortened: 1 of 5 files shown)


meandavejustice (changes requested):
Looks good, I noticed two things that need to be translated here:
* GetDataTrio title attributes
* `add-recovery-key-error`

The error looks like it's coming from the server so doesn't need to be included in this patch, but I wanted to note it so that it doesn't go untranslated in the future. Do you know if there is a ticket for translating these?

6a68:
Great catch, thanks! For the error strings, it looks like they are [already translated](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/public/locales/en-US/settings.ftl#L161-L225), so I just updated the error handling to use those l10n ids. Ready for another review.

meandavejustice (approved):
👍

Issue #7350, opened 2021-01-25
Add a nonce to load PayPal inline script and style elements


data-sync-user:
While FXA-2800 adds PayPal to our CSP when there is an explicit origin to include, allowing inline script and style tags requires generating a secure nonce on the server side and passing it to the client.

This is the suggested route per [PayPal's docs](https://developer.paypal.com/docs/checkout/troubleshoot/support/#content-security-policy-rules).

[~bbangert@mozilla.com] Does FxA already do this? What would you recommend as a good pattern?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2944)
┆Issue Number: FXA-2944

data-sync-user:
➤ Bianca Danforth commented:

Per this Slack thread ( https://mozilla.slack.com/archives/CG0AJ6E77/p1611587196023300 ) with Ben and Barry:

  * [Ricky] I did have it on the VPN site at one point https://github.com/mozilla-services/guardian-website/commit/10bef75c87962e950e7a3dd5ad580aecc92890c9
  * [Ben] the server could just make a uuid4 as the nonce and put that on the script tag and header
  * [Barry] No, we don't have that for Stripe
  * [Barry] Inject the nonce into index.html, read it and add it to the app context
  * See fxa-payments-server/server/lib/server.js for where we inject config into index.html
  * [Barry] meta tag seems the most appropriate

Issue #7269, opened 2021-01-14
Checkout: Add PayPal button createOrder handler to the Subscribe page


data-sync-user:
Edit: This ticket covers the `createOrder` PayPal button lifecycle method handler. The `onApprove` handler will be covered by a separate ticket, [FXA-2976](https://jira.mozilla.com/browse/FXA-2976).

---

Referring to the [PayPal Checkout Flow Implementation](https://docs.google.com/document/d/1j4qGzUK0IOBOh8VeB0yHt_ZEHAez3XNXKytrF6ZNAsw/edit#heading=h.n5lzn0am6uzx) created by Ben, we should add the [PayPal Checkout page script as described in the docs](https://developer.paypal.com/docs/checkout/reference/upgrade-integration/#2-add-the-smart-payment-buttons) and add the function bodies for all the required lifecycle methods for `paypal.Buttons()` per the [Checkout ladder diagram](https://github.com/mozilla/ecosystem-platform/pull/115).

E.g. `createOrder`, `onAuthorize`, etc. We can stub out the fxa-auth-server routes temporarily if needed, as long as we file a follow up ticket to use the correct routes once implemented.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2902)
┆Issue Number: FXA-2902

Issue #7115, opened 2020-12-10
Modify existing Customer object to have additional PayPal properties


data-sync-user:
For PayPal customers, there will be some additional payment information we want to display in the front end. That is accessed via a `Customer` object, which currently only has a [list of subscriptions and a `last4` property](https://github.com/mozilla/fxa/blob/44bbb161d958f084c1bb39902554f69a2333de90/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L87) representing the last four digits of their credit card number from Stripe.

We will need some new properties to be able to pass some PayPal specific information. Particularly billing information. For example, we need to know if the user is an existing PayPal customer for the checkout flow. What additional billing information we may require depends on FXA-2799.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2812)
┆Issue Number: FXA-2812

data-sync-user:
➤ Bianca Danforth commented:

FXA-2938 handled part of this, as there is now a `paypalAgreementId` property on the Stripe Customer object.
Per Ben:
> we have a key for current paypal billing agreements, but if the billing agreement is cancelled, we remove that key. yet there's a subscription requiring paypal.
> the customer object we return to the front-end for the subscription status check should include a 'paypal' value instead of the last4/brand
> whether that value should be included is if there's an active/past_due subscription that has collection_method of send_invoice

Issue #7098, opened 2020-12-08
Checkout: Add PayPal payment option on Subscribe page if there is saved PayPal payment information


data-sync-user:
As a simplification, for the MVP, we would like to only allow one payment option at a time. Therefore, similar to paying with Stripe, if we have saved PayPal payment information, we should have a screen to enable the user to subscribe to a product using it.

Edit: Also once #7115 is complete, this patch should add the logic to check whether an FxA user is an existing PayPal customer with saved payment information.

Open question(s):

* If the payment information is invalid, expired or otherwise unusable, do we show an error message here as we do for subscription management?

Task:

* Use the new Customer object to get the PayPal properties we need for information.
* Update logic for what components to show based on new Customer object properties
* Create new front end components for PayPal UI
* Write tests for those components in Jest.

This should be much easier after FXA-2800; we just have to find out what billing information we can/should show.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2801)
┆Issue Number: FXA-2801

data-sync-user:
➤ Bianca Danforth commented:

> This should be much easier after FXA-2800; we just have to find out what billing information we can/should show.

Since there can be many different funding sources for PayPal ( https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#disable-funding ) (see also FXA-2783 ( https://jira.mozilla.com/browse/FXA-2783 )), in contrast to Stripe which is only credit card, the recommendation is to just show that their billing source is PayPal.

data-sync-user:
➤ Jessica Cook commented:

Hi [~bchen@mozilla.com] adding the link to the saved payment option spec here https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=696%3A4

data-sync-user:
➤ Barry Chen commented:

A few notes before I set this back to To Do.

The issue is currently blocked by FXA-3069.

There's a WIP branch at https://github.com/chenba/fxa/tree/2801-paypal-already.  That branch is based on the work Lesley did in FxA PR 7570, which is not yet in main.   ( https://github.com/chenba/fxa/tree/2801-paypal-already. )There a few failing existing tests in the branch; no new tests so far.  And form submission callback to call the API to create the additional subscription is not yet implemented.

data-sync-user:
➤ Bianca Danforth commented:

[...]

PR #7353, opened 2021-01-25
feat(payments): Add PayPal button to Subscribe page

biancadanforth:
## Because

- We want to provide the option to pay with PayPal on the Subscribe page.

## This pull request

- Adds a `PaypalButton` React component that renders if the PayPal feature is enabled (Issue #7271 ), and the customer is a new customer.

Closes: #7097 

## Checklist

## Screenshots (Optional)

Screen Shot 2021-01-25 at 3 08 16 PM

- To see the button, the PayPal feature must be enabled, see #7271 .
  - i.e. `FEATURE_USE_PAYPAL_UI_BY_DEFAULT=true pm2 restart payments --update-env`
- There is one CSP error that still needs to be handled by #7350 .
- The button needs to be shown conditionally based on whether there is any saved payment information. Currently, it is shown only if there is no saved Stripe payment information. After #7115 , #7098 will handle checking for saved PayPal payment information. 
- UI for this needs to be finalized with the TBD UX spec, see [FXA-2799](https://jira.mozilla.com/browse/FXA-2799), for example by adding the PayPal privacy policy.
- Button functionality will be handled by #7269 .

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
@@ -1,4 +1,10 @@
-import React, { useState, useCallback } from 'react';
+import React, {
+  useState,
+  useCallback,
+  useContext,
+  useEffect,
+  Suspense,
+} from 'react';
 import { Stripe, StripeCardElement, StripeError } from '@stripe/stripe-js';
 import { Plan, Profile, Customer } from '../../../store/types';
 import { State as ValidatorState } from '../../../lib/validator';
@@ -16,7 +22,7 @@ import * as Amplitude from '../../../lib/amplitude';
 import { Localized } from '@fluent/react';
 import * as apiClient from '../../../lib/apiClient';
 
-import { config } from '../../../lib/config';
+import { AppContext } from '../../../lib/AppContext';
 
 import '../../Product/SubscriptionCreate/index.scss';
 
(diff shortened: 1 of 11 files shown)


chenba:
@biancadanforth there's an error with the build

biancadanforth:
I am going to add a check for `hasExistingCard`, so this at least only shows if there is no saved Stripe payment information; #7098 covers adding the additional logic for PayPal information.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -61,6 +69,21 @@ export const SubscriptionCreate = ({
 }: SubscriptionCreateProps) => {
   const [submitNonce, refreshSubmitNonce] = useNonce();
 
+  useEffect(() => {
+    if (!config.featureFlags.usePaypalUIByDefault) {
+      return;
+    }
+    const script = document.createElement('script');
+    script.src = `https://www.paypal.com/sdk/js?client-id=${config.paypal.clientId}&vault=true&commit=false&intent=authorize&disable-funding=credit,card`;
```
rlr, line 77: nit: should we use `config.paypal.scriptUrl` since we have it instead of hard-coding `https://www.paypal.com`?

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -71,6 +94,10 @@ export const SubscriptionCreate = ({
     [selectedPlan]
   );
 
+  const { config } = useContext(AppContext);
+
+  const [paypalScriptLoaded, setPaypalScriptLoaded] = useState(false);
```
rlr, line 99: It's weird that this is here but it's used above. I would move the whole `useEffect(() => {...});` from above to somewhere below this.

chenba (approved):
LGTM.  Maybe move the `setState` call as suggested by rlr?

Issue #7236, opened 2021-01-09
Add DoRefernceTransaction command to PayPalHelper


data-sync-user:
DoRefernceTransaction should be added to the PayPalHelper, which should provide a typesafe API and then call a new underlying operation on the PayPal NVP API client.

PayPal docs for this API operation:

https://developer.paypal.com/docs/nvp-soap-api/do-reference-transaction-nvp/

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2881)
┆Issue Number: FXA-2881

data-sync-user:
➤ Sarah Bird commented:

GDoc https://docs.google.com/spreadsheets/d/1eLZZXvJ4vtTivplIOxGhUBkwDXcVFzJsrKlmLB6h4HQ/edit#gid=689788281 contains notes of which params to use

PR #7355, opened 2021-01-25
feat(auth-server): add chargeCustomer to Paypalhelper

birdsarah:
## Because

- We need to make a call to PayPal's DoReferenceTransaction to charge a customer

## This pull request

- Adds a doReferenceTransaction method to Paypal client
- Adds a chargeCustomer to PaypalHelper
- Adds fixtures for DoReferenceTransaction and CreateBillingAgreement
- Small clean-up, renames SetCheckout to SetExpressCheckout for consistency

Closes: #7236

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -4,6 +4,7 @@
 import pRetry from 'p-retry';
 import querystring from 'querystring';
 import superagent from 'superagent';
+import { ChargeCustomerOptions } from './paypal';
```
bbangert, line 7: It doesn't seem like the right separation of concerns to have a lower level client pull in types from the higher level helper library for its own use. Ideally the lower level client could be yanked out of FxA entirely on its own for use as a stand-alone library (though obviously we aren't going to do that at this time).

bbangert (changes requested):
Looks great to me, only change request is re: the type location.

packages/fxa-auth-server/lib/payments/paypal-client.ts
birdsarah: I agree. Will change.

bbangert (approved):
Nice! 💯

PR #7361, opened 2021-01-26
fix(tests): add script to run all the functional tests

ashrivastava-qa:
## Because
`Added script in the test-ci-remote.sh file to run the functional test suite after smoke tests`

-

## This pull request

-

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

@mozilla/fxa-devs please review. Thanks!

### DIFF
--- a/packages/fxa-content-server/scripts/test-ci-remote.sh
+++ b/packages/fxa-content-server/scripts/test-ci-remote.sh
@@ -41,10 +41,8 @@ mkdir -p artifacts/tests
 
 cd packages/fxa-content-server
 mozinstall /firefox.tar.bz2
-test_suite functional_smoke
+test_suite functional_smoke && yarn test-functional
 # TODO - restore these
 #test_suite server
 #mozinstall /7f10c7614e9fa46-target.tar.bz2
 #test_suite pairing
-
-


6a68 (approved):
LGTM, thanks for the PR

packages/fxa-content-server/scripts/test-ci-remote.sh
```diff
@@ -41,10 +41,8 @@ mkdir -p artifacts/tests
 
 cd packages/fxa-content-server
 mozinstall /firefox.tar.bz2
-test_suite functional_smoke
+test_suite functional_smoke && yarn test-functional
```
chenba, line 44: hmm actually I think we need `test_suite all` instead.

chenba: Also, "all" is a superset of "functional_smoke".  Maybe it would be even better if we define a new suite that's "all" minus "functional_smoke" so we don't repeat tests.

chenba (changes requested):
I think we need `test_suite all` instead of `yarn test-functional`.

chenba:
Also FYI @jbuck in case you aren't aware

packages/fxa-content-server/scripts/test-ci-remote.sh
ashrivastava-qa: Would `test_suite all` include all the suites defined (eg settings_v2 suite and others) ? if yes, why do we need `all` instead of the `functional`? @chenba

chenba: I think there's a bit of confusion here.  `functional` would not be a valid argument, so it's not `all` instead of `functional`.  You can see the suites options and their mapping at https://github.com/mozilla/fxa/blob/244765119c61079f33c47caef617779ce706d838/packages/fxa-content-server/tests/intern.js#L111.

I was suggesting that we need to use `test_suite {suites_option_value}` and not `yarn`.

`all` does not include `settings_v2` but it does include all the tests from `functional_smoke`.

ashrivastava-qa: okay in that case I think I'll keep it `test_suite functional_smoke && test_suite functional` and remove the tests in 'functional' that are in `functional_smoke`. That way there won't be any duplicates test running. Does that make sense?

chenba: `functional` is not a valid option.  If you need a set of test suites that is the difference between `all` and `functional_smoke`, then please create it and add that as a new case to the switch statement I linked above.

ashrivastava-qa: I do see `functional.js` file. It's just the a declaration and a switch case is not added with functional suite in the `intern.js` file here.
https://github.com/mozilla/fxa/blob/244765119c61079f33c47caef617779ce706d838/packages/fxa-content-server/tests/intern.js#L111

But if I add it, it should work. Currently also when I ran the command test_suite functional_smoke && test_suite functional on my local it worked.

ashrivastava-qa: Yes exactly that's what I was trying to convey but it wasn't clear I guess. I do see `functional.js` file already which I will use as a new case in the switch statement. But I will remove all the tests from this file that already exists in the `functional_smoke` suite.

chenba: There isn't a 1:1 mapping between the test suites options and filenames.

When the script in question calls `test_suite all`, it executes `node tests/intern.js --suites="all" ...`.  That then lead the switch statement and the "all" case.  That case then maps to the suites defined in `functional.js`.  You don't need to add a new case since it's already in place.

But please don't remove tests from "all"/`functional.js`; test-ci-remote isn't the only thing uses intern.js.  Please make a new set that's the difference between "all" and "functional_smoke".

ashrivastava-qa: Thanks @chenba for the review and suggestions. I'll make all the necessary changes in a new PR (since me and Rebecca needs to come up with a few ideas on smoke tests and regression test suite). Hence closing this.

Issue #7344, opened 2021-01-22
[settings][l10n] Pull reused strings into App ftl file


data-sync-user:
Instead of duplicating strings across new-settings, let's collect them in one spot. The top-level App component's ftl file seems like a fine place. This will minimize the workload on our volunteer localizers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2943)
┆Issue Number: FXA-2943

data-sync-user:
➤ Barry Chen commented:

Fluent messages can reference each other (https://www.projectfluent.org/fluent/guide/references.html) ( https://www.projectfluent.org/fluent/guide/references.html) ) so we could DRY up repeated strings as well (e.g. "Firefox Accounts").

data-sync-user:
➤ Barry Chen commented:

The error message for VerifiedSessionGuard is definitely repeated.  Maybe the component should provide a default.

PR #7363, opened 2021-01-26
feat(settings): extract reused strings into the App ftl file

6a68:
## Because

- we want to minimize the workload on our volunteer translators

## This pull request

- extracts reused strings into the App ftl file, so they only need to be
translated once, namespacing regular strings under `fxa-`, and converting
brand names into [Terms](https://www.projectfluent.org/fluent/guide/terms.html)

- replaces single quotes with true apostrophes, a subtle requirement
that seems to come up with every localization project.

Closes: #7344

## Checklist

## Screenshots (Optional)


meandavejustice (approved):
Nice 😎

Issue #5061, opened 2020-04-22
[l10n] Main Settings component translations


data-sync-user:
Add l10n support for the main Settings component. This includes all the data displayed from the FXA-1604 user story and the Connected Devices section.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1728)
┆Issue Number: FXA-1728

PR #7368, opened 2021-01-27
chore(settings): localize main settings page

chenba:
Because:
 - we should localize new settings

This commit:
 - localize the Settings component and the components therein
 - update the alert bar to make use of the auth-error-{errno} strings by
   default when available

Closes: #5061


dannycoates (approved):
so many strings. thanks @chenba this LGTM 👍

PR #7369, opened 2021-01-27
feat(auth-server): add paypal IPN handler

### LINKED ISSUE #7276: Create paypal webhook handler
To handle paypal IPN, we need a paypal webhook handler, URL route, and IPN validation per the [implementation guide](https://developer.paypal.com/docs/api-basics/notifications/ipn/IPNImplementation/). This paypal webhook class should then have code so that it can dispatch internally to additional methods depending on the [transaction type](https://developer.paypal.com/docs/api-basics/notifications/ipn/IPNandPDTVariables/#id08CTB0S055Z).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2906)
┆Issue Number: FXA-2906

bbangert:
Because:

* We want to support paypal IPN messages sent to us.

This commit:

* Adds a paypal IPN message handler with routes and types setup for
  incoming IPN messages.

Closes #7276

## Checklist


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -88,6 +93,39 @@ export type DoReferenceTransactionOptions = {
...
+  mp_notification?: any;
+  mp_pay_type: 'INSTANT' | 'ANY' | 'ECHECK';
+  // Agreement status, A for Active, I for inactive
+  // I is equivalent to cancelled.
+  mp_status: 'A' | 'I';
+  txn_id: string;
+  txn_type: 'merch_pmt' | 'mp_cancel';
+};
+
+export type IpnMessage = IpnBaseType;
```
chenba, line 115: Couldn't we rename IpnBaseType to IpnMessage and not have this alias?

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -4,11 +4,68 @@
...
+      }
+      this.log.debug('Unhandled Ipn message', { payload });
+    } catch (err) {
+      reportSentryError(err, request);
+      this.log.error('verifyAndDispatchEvent', {
+        payload: request.payload,
+        err,
+      });
+    }
+    return false;
```
chenba, line 53: Since the return value isn't used, I don't think we need to explicitly return false here.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -31,8 +88,21 @@ export const paypalRoutes = (
...
+        // We'll use the PayPals verification to authenticate.
+        auth: false,
+        // The raw payload is needed for authentication.
+        payload: {
+          output: 'data',
+          parse: false,
+        },
+      },
+      handler: (request: AuthRequest) =>
+        paypalNotificationHandler.handleIpnEvent(request),
```
chenba, line 105: Since `handleIpnEvent` returns a `{}`, should be it returned here as the response content?

bbangert: This will return it, right? An explicit isn't returned isn't needed since its not a block.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
bbangert: That's true, though it made the tests a little more elegant to verify the function made it to the desired end vs. just 'returning' an undefined implicitly.

chenba (approved):

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
chenba: oh doh, yes, I misread that.  Didn't notice it's a () => expression and not () => block. :-P

packages/fxa-auth-server/lib/payments/paypal-client.ts
bbangert: Ah, yes, that would probably be a bit better, I originally had this more complicated and forgot to remove this.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
chenba: To me, I interpret that as the function communicating some sort of status with the returned false.  (Obviously if I look at the call site of the function I would know that's not the case.)  YMMV with how others read the intention of that return.

PR #7387, opened 2021-01-29
feat(auth-server): add route for getCheckout Token

birdsarah:
## Because

- Route must be added to get paypal checkout token

## This pull request

- Adds the route

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
@@ -2,13 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import { ServerRoute } from '@hapi/hapi';
+import isA from '@hapi/joi';
 import Container from 'typedi';
 
 import { ConfigType } from '../../../config';
 import { PayPalHelper } from '../../payments/paypal';
 import { StripeHelper } from '../../payments/stripe';
-import { AuthLogger } from '../../types';
+import { AuthLogger, AuthRequest } from '../../types';
 import { StripeHandler } from './stripe';
+import { handleAuth } from './utils';
 
 export class PayPalHandler extends StripeHandler {
   protected paypalHelper: PayPalHelper;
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -26,6 +27,13 @@ export class PayPalHandler extends StripeHandler {
     super(log, db, config, customs, push, mailer, profile, stripeHelper);
     this.paypalHelper = Container.get(PayPalHelper);
   }
+
+  /**
+   * Get PayPal Checkout Token.
+   */
+  async getCheckoutToken() {
+    return await this.paypalHelper.getCheckoutToken();
```
bbangert, line 35: There's a few calls that are pretty standard on every http call:
https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts#L178

The minimum is generally the log statement, and the customs check (prevents abuse of our endpoints).

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -49,8 +57,20 @@ export const paypalRoutes = (
...
+    {
+      method: 'POST',
+      path: '/oauth/subscriptions/paypal-checkout',
+      options: {
+        auth: {
+          payload: false,
+          strategy: 'oauthToken',
+        },
+        response: {
+          schema: isA.string(),
```
bbangert, line 70: Our endpoints return JSON, so I'd expect the schema here to be an object with a token property.

bbangert (changes requested):
Two minor changes, other than that this looks great!

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -49,8 +64,23 @@ export const paypalRoutes = (
...
-  //        overall request deadline. Stripe retries must include a idempotency_key.
-  return [];
+  return [
+    {
+      method: 'POST',
+      path: '/oauth/subscriptions/paypal-checkout',
+      options: {
+        auth: {
+          payload: false,
+          strategy: 'oauthToken',
```
birdsarah, line 74: Is this is the correct strategy? I don't know which to pick and I'm not sure how to write a test for it.

bbangert: No test is needed for this bit.

bbangert:
The CI errors look legitimate for the TypeScript issues.

bbangert (approved):

PR #7393, opened 2021-02-01
fix(auth-server): Correct checks in createRoutes

birdsarah:
## Because

- Moving to new paradigm of Container means that paypalHelper is not passed into createRoutes.

## This pull request

- Fixes the createRoutes method to check the config, not expect a passed in PayPalHelper

Closes: # (issue number)

## Checklist

## Screenshots (Optional)


bbangert (approved):

birdsarah:
The test failure is a good one. Could this happen in the real world @bbangert? If yes, then we would need to update the other places this check happens.

bbangert:
I don't think it can happen in the real world because the config has defaults. It can only happen in test world where the entire config object is constructed manually.

PR #7396, opened 2021-02-01
feat(auth-server): add paypal subscription create route

### LINKED ISSUE #7326: Create paypal handler for customer authorization
A Paypal handler should be created for /oauth/subscription/active/new-paypal per the linked issue references. This will make a variety of calls per:

[https://github.com/mozilla/ecosystem-platform/blob/master/docs/fxa-engineering/fxa-sub-platform.md#paypal-checkout](https://github.com/mozilla/ecosystem-platform/blob/master/docs/fxa-engineering/fxa-sub-platform.md#paypal-checkout)

It will need to make multiple stripe helper and paypal helper calls to store references, create billing agreement, etc.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2938)
┆Issue Number: FXA-2938

bbangert:
Because:

* We want to handle creating a subscription and paying it via PayPal
  when a user chooses PayPal.

This commit:

* Adds a paypal-new subscription route that will create a billing
  agreement, subscription, and setup the Stripe data for recurring
  billing to PayPal.

Closes #7326

## Checklist


packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -82,5 +162,30 @@ export const paypalRoutes = (
...
+        auth: {
+          payload: false,
+          strategy: 'oauthToken',
+        },
+        response: {
+          schema: isA.object().keys({
+            subscription: validators.subscriptionsStripeSubscriptionValidator,
+            sourceCountry: validators.subscriptionPaymentCountryCode.required(),
+          }),
+        },
```
biancadanforth, line 178: The validation here still throws an error and returns a 500 response for me on the client. I only get a 200 when I add `failAction: 'log',` as we had to do for the demo. Can we fix this before merging else file a follow-up to fix it next?

biancadanforth (changes requested):
Thanks Ben for this PR.

I haven't looked at everything closely, but I just ran the end-to-end checkout flow on the `fxa-payments-server`, and I got back a `500` response again. It's the same validation issue on the `'/oauth/subscriptions/active/new-paypal'` as we ran into for the demo that doesn't show up in the auth server logs. See my comment in-line.

biancadanforth (approved):

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -557,6 +688,30 @@ export class StripeHelper {
...
+      throw error.internalValidationError(
+        'findCustomerSubscriptionByPlanId',
+        {
+          customerId: customer.id,
+        },
+        'Expected subscriptions to be loaded.'
+      );
+    }
+    return customer.subscriptions.data.find(
+      (sub) => sub.items.data.find((item) => item.plan.id === planId) != null
```
chenba, line 711: nit: `sub.items.data.some` would read better here.

bbangert: This method was just moved around, though wouldn't `some` get a list while this wants a single item?

chenba: `some` just help with removing the ` != null` comparison as it already returns a boolean.  Like I said, nit. :D

chenba (approved):
LGTM

Issue #7319, opened 2021-01-21
[l10n] copy files between settings and l10n repo


data-sync-user:
We need copy the en-US file into the l10n repo and copy the translated files into settings.

See https://jira.mozilla.com/browse/FXA-2344?focusedCommentId=216287&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-216287 and https://jira.mozilla.com/browse/FXA-2344?focusedCommentId=216294&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-216294 ([https://github.com/mozilla/fxa/issues/6036#issuecomment-760301898](https://github.com/mozilla/fxa/issues/6036#issuecomment-760301898) and https://github.com/mozilla/fxa/issues/6036#issuecomment-760306388).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2934)
┆Issue Number: FXA-2934

data-sync-user:
➤ Barry Chen commented:

https://github.com/mozilla/fxa-content-server-l10n/pull/464 ( https://github.com/mozilla/fxa-content-server-l10n/pull/464 ) is the patch for copying from settings into the l10n repo.

data-sync-user:
➤ Barry Chen commented:

I will do FXA-2156 as a part of this task.

data-sync-user:
➤ Barry Chen commented:

Put this back to To Do while we get the l10n side of things done.

data-sync-user:
➤ Barry Chen commented:

Bumped the points up because the number of changes needed for https://github.com/mozilla/fxa-content-server-l10n/pull/466

data-sync-user:
➤ Barry Chen commented:

I've been going through the comments on https://github.com/mozilla/fxa-content-server-l10n/pull/466/files ( https://github.com/mozilla/fxa-content-server-l10n/pull/466/files ) and making changes in FxA based on that feedback.  The WIP branch is at https://github.com/chenba/fxa/tree/2934-ftl-fixes-II. ( https://github.com/chenba/fxa/tree/2934-ftl-fixes-II. )  I mark each comment as resolved when I make the corresponding change in FxA, so the unresolved comments is a to-do list.

[...]

PR #7426, opened 2021-02-03
chore(l10n): combine the download l10n scripts

chenba:
Because:
 - we have multiple similar scripts that pull translations from the
   fxa-content-server-l10n repo
 - we need that functionality for new settings

This commit:
 - combine the scripts into a single one

Part of #7319


chenba:
An earlier version of this patch tried to check the l10n repo into a shared location.  That was a bit of a disaster since yarn execute things in parallel, so there were overlapping calls to the script.  This version doesn't change the location of the checkout for any module, thus very low risk.

dannycoates:
> That was a bit of a disaster since yarn execute things in parallel, so there were overlapping calls to the script.

Not that I think we should change this PR, but would putting the script in the top-level package.json postinstall work?

chenba:
> 
> would putting the script in the top-level package.json postinstall work?

I think so.  At least I cannot think of an obvious reason why it wouldn't.

_scripts/clone-l10n.sh
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+if [ -n "$FXA_L10N_SKIP" ]; then
+    echo "Skipping fxa-content-server-l10n update..."
+    exit 0
+fi
+
+DOWNLOAD_PATH="https://github.com/mozilla/fxa-content-server-l10n.git"
+SCRIPT_DIR=$(dirname "$0")
+ROOT_DIR=$(realpath "$SCRIPT_DIR/..")
```
dannycoates, line 19: `realpath` doesn't exist on mac. I think we should be ok using paths relative to `SCRIPT_DIR`

_scripts/clone-l10n.sh
```diff
@@ -0,0 +1,65 @@
...
+  esac
+
+  for src in locale/**/$FTL_FILENAME.ftl; do
+    [ -f "$src" ] || continue
+    dir=$(dirname "$src")
+    dir=$(echo "$dir" | sed "s/_/-/")
+    dir=$(basename "$dir")
+    base=$(basename "$src")
+    mkdir -p "$MODULE_PATH/public/locales/$dir"
+    cp "$src" "$MODULE_PATH/public/locales/$dir/$base"
```
dannycoates, line 61: If `$MODULE_PATH` is relative, this can be `"../public/locales/$dir"`

dannycoates (approved):
pre-approving but needs a fix for mac. it works on mac after i tweaked the paths in the script to be relative. i'd like to get to a single dir for fxa-content-server-l10n someday

chenba:
> 
> pre-approving but needs a fix for mac. 

Thanks.  Not writing shell script often enough to keep this compatibility stuff in mind.  :-/

> i'd like to get to a single dir for fxa-content-server-l10n someday

If you mean simply having a single checkout dir, then that should be pretty easy.  But, we'd still need to copy stuff around.  auth-server's mail translator requires the translations directly, so in order to keep the relative path from the translator to the translations constant, we'd need to copy fxa-content-server-l10n.  (Of course we could change translator itself...)  

refs: 
 - https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/translator.js#L26
 - https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/package.json#L12

PR #7430, opened 2021-02-03
feat(auth-server): handle invoice creation events

### LINKED ISSUE #7065: Implement Stripe Webhook handler for invoice creation
Add a new handler for this webhook: `invoice.created`:

When a Stripe Invoice is created:

* If the collection method is "send invoice" and the invoice reason is subscription_cycle, then we:
  * Add metadata for "payment attempts" and initialize it to 0.
  * Finalize the invoice with auto-advance set to false.
  * Note: Payment Retries field on Stripe invoice is only used when Stripe is processing charges and we cannot update this field.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2780)
┆Issue Number: FXA-2780

bbangert:
Because:

* We want to switch invoices that are charged using PayPal
  to not auto-advance and finalize them ourselves.

This commit:

* Adds a Stripe webhook handler for invoice.created events that
  will update the invoice to not auto-advance and finalize them.

Closes #7065

## Checklist


packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -149,6 +174,33 @@ export class StripeWebhookHandler extends StripeHandler {
     );
   }
 
+  /**
+   * Handle `invoice.created` events, if the subscription is for an invoice
+   * customer (PayPal), set auto_advance false and finalize.
+   */
+  async handleInvoiceCreatedEvent(request: AuthRequest, event: Stripe.Event) {
+    // Type-guard to require paypalHelper.
```
chenba, line 182: Since we are using TypeScript and type guards are a thing in TS, it might be a little confusing to call this a type guard.

chenba (approved):
LGTM 👍

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: A Type Guard in TS is pretty much anything that narrows the type, which this is doing.
https://basarat.gitbook.io/typescript/type-system/typeguard illustrates a variety of them, including one of this type.

PR #7439, opened 2021-02-04
fix(settings): fix startup error on some legacy browsers

dannycoates:
## Because

Apparently oldish ios safari won't let you extend EventTarget

Screen Shot 2021-02-04 at 1 43 42 PM

## This pull request

- uses a mock for those browsers


chenba (approved):
LGTM

Issue #7010, opened 2020-11-25
[FxA-Settings] FxA Bento menu content is different than Monitor Bento menu one


vasilicatamas:
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Windows 10 x64
Android 9
iOS 13.6

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on the Bento menu displayed in the header.
3. Navigate to https://monitor.firefox.com/ and open the Bento menu.
4. Compare the Firefox Services displayed in both menus.

**Expected results**
The same Firefox Services are displayed in both Bento menus 

**Actual Results**
FxA bento menu displays “Firefox Lockwise” instead of “Mozilla VPN” service.

![Screenshot_40](https://user-images.githubusercontent.com/18114210/100251110-0fd91300-2f47-11eb-89d0-7773c10334a5.png)
![Screenshot_41](https://user-images.githubusercontent.com/18114210/100251112-1071a980-2f47-11eb-91ee-f5e9f6b334fd.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2739)
┆Issue Number: FXA-2739

data-sync-user:
➤ Wil Clouser commented:

[~adavis@mozilla.com] should we be promoting the VPN instead of lockwise?  /cc [~pwalmsley@mozilla.com]

data-sync-user:
➤ Alex Davis commented:

We could. Lockwise is in a bit of a limbo state so the VPN would make sense.

data-sync-user:
➤ Philip Walmsley commented:

Yep, agreed.

data-sync-user:
➤ Barry Chen commented:

[~adavis@mozilla.com] What should the URL for the link be?  We want some proper marketing related params in it I'd think?

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [1.200.0] under Windows 10 64-bit, macOS 10,14,6, Android 9 and iOS 13.6. FxA bento menu successfully displays "Mozilla VPN” instead of “Firefox Lockwise” .

PR #7444, opened 2021-02-05
chore(settings): replace lockwise with vpn in bento menu

chenba:

Closes: #7010


dannycoates (approved):
🔐

PR #7451, opened 2021-02-08
fix(settings): when gql-api switched to nestjs the auth error changed

dannycoates:
## Because

- The new settings page started showing the blank "Application Error" page when the sessiontoken was no longer valid.

## This pull request

- fixes it to redirect to signin


chenba (approved):
🧅 Looks fine but I'm not that familiar with Nest

Issue #6747, opened 2020-10-26
Add an autocomplete to the email search


data-sync-user:
After 3 characters are typed lets add a short dropdown with autocomplete suggestions for the email address.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2633)
┆Issue Number: FXA-2633

PR #7456, opened 2021-02-08
Ap 2633 autocomplete

gsiriya:
## Because

the autocomplete feature for the admin panel is completed

## This pull request

allows an admin to search for an account by email while providing them a list of suggested emails based on entries in the database

Closes: #6747

## Checklist

- [x ] My commit is GPG signed.

## Screenshots (Optional)

image


packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+    getEmailsLike(search: $search) {
+      email
+    }
+  }
+`;
+
 export const AccountSearch = () => {
   const [inputValue, setInputValue] = useState<string>('');
   const [showResult, setShowResult] = useState<Boolean>(false);
+  const [suggestions, setSuggestion] = useState<Array<String>>([]);
```
dannycoates, line 63: this should be a lower case `string`. in fact `string[]` is more idiomatic than `Array<string>`

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+  const [text, setText] = useState<String>('');
```
dannycoates, line 64: should be lower case `string`

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+
 export const AccountSearch = () => {
   const [inputValue, setInputValue] = useState<string>('');
   const [showResult, setShowResult] = useState<Boolean>(false);
+  const [suggestions, setSuggestion] = useState<Array<string>>([]);
+  const [text, setText] = useState<string>('');
   const [getAccount, { loading, error, data, refetch }] = useLazyQuery(
     GET_ACCOUNT_BY_EMAIL
   );
+  const [getEmailLike, { data: data1, refetch: refetch1 }] = useLazyQuery(
```
dannycoates, line 68: Let's find better names than `data1` and `refetch1`, something with `email` in it maybe

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+      email
+    }
+  }
+`;
+
 export const AccountSearch = () => {
   const [inputValue, setInputValue] = useState<string>('');
   const [showResult, setShowResult] = useState<Boolean>(false);
+  const [suggestions, setSuggestion] = useState<Array<string>>([]);
+  const [text, setText] = useState<string>('');
```
dannycoates, line 64: What is `text`? Try to name things so they describe what they're used for. This is the value of the search input, right?

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+    getAccount({ variables: { email: text } });
     // Don't hide after the first result is shown
     setShowResult(true);
   };
 
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+    setSuggestion([]);
+    onTextChanged(event);
+    getEmailLike({ variables: { search: event.target.value } });
```
dannycoates, line 83: I see that `getEmailLike` is called in `onTextChanged`. Do we still need this one. We shouldn't call it more than necessary.

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+    //getEmails();
+  };
+
+  const onTextChanged = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value;
+
+    let filtered_list = [];
+    if (value.length > 0) {
```
dannycoates, line 91: For performance against the real database we should probably only start searching after 5 or so characters

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+      console.log('value:', value);
+      getEmailLike({ variables: { search: value } });
+
+      if (data1 != null) {
+        console.log(data1);
```
dannycoates, line 96: let's remove these `console.log`s before merging

packages/fxa-admin-panel/src/components/AccountSearch/index.tsx
```diff
@@ -49,20 +49,82 @@ export const GET_ACCOUNT_BY_EMAIL = gql`
...
+        for (let i = 0; i < data1.getEmailsLike.length; i++) {
+          filtered_list[i] = data1.getEmailsLike[i].email;
+        }
+      }
+    }
+    const regex = new RegExp(`^${value}`, 'i');
+    filtered_list = filtered_list.sort().filter((v) => regex.test(v));
```
dannycoates, line 103: If we change our query on the server side we won't need to filter these results and the sql query will be more efficient. See my comment in `account.resolver.ts`

packages/fxa-admin-server/src/gql/account/account.resolver.ts
```diff
@@ -61,6 +62,16 @@ export class AccountResolver {
       .first();
   }
 
+  //query to return all emails that are in the database
+  @Query((returns) => [EmailType], { nullable: true })
+  public getEmailsLike(@Args('search', { nullable: false }) search: string) {
+    return this.db.emails
+      .query()
+      .select(EMAIL_COLUMNS)
+      .where('email', 'like', `%${search}%`)
```
dannycoates, line 71: The `%${search}%` will search for that substring anywhere in the email address because it has `%` on both sides. We want our suggestions to be a prefix search so we only need the `%` on the end.

packages/fxa-admin-server/src/gql/account/account.resolver.ts
```diff
@@ -61,6 +62,16 @@ export class AccountResolver {
       .first();
   }
 
+  //query to return all emails that are in the database
```
dannycoates, line 65: This comment isn't correct anymore

dannycoates (changes requested):

dannycoates (approved):

PR #7457, opened 2021-02-08
fix(settings): only set hover css when supported

### LINKED ISSUE #6371: [FxA-Settings] “Make primary” button is displayed in hovered state after tapping on it on iOS
**Build**
Stage [Train 1.186.0]

**Affected platforms**
iOS 13.6 

**Affected devices**
iPhone 11

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/
2. Login using an user that has already added a secondary email
3. Tap on “Make primary” button.

**Expected results**
The “Make primary” button returns to the enabled state.

**Actual Results**
The “Make primary” button is displayed in the hover state.

issue5

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2512)
┆Issue Number: FXA-2512

dannycoates:
## Because

- iOS is weird about `hover` effects

## This pull request

- uses a `@media` query to set the hover styles

fixes #6371

### DIFF
--- a/packages/fxa-settings/src/styles/ctas.scss
+++ b/packages/fxa-settings/src/styles/ctas.scss
@@ -11,10 +11,6 @@
 .cta-neutral {
   @apply bg-grey-10 border-grey-200;
 
-  &:hover {
-    @apply border-grey-200 bg-grey-100 text-grey-400;
-  }
-
   &:active {
     @apply border-grey-400 bg-grey-100 text-grey-400;
   }
@@ -30,10 +26,6 @@
 
 .cta-primary {
   @apply bg-blue-500 border-blue-600 text-white;
-
-  &:hover {
-    @apply bg-blue-600 border-blue-600;
-  }
 
   &:active {
     @apply bg-blue-800 border-blue-800;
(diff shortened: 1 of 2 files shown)


jodyheavener:
Interesting. Doesn't the hover effect go away when you tap outside of the button?

chenba (approved):
TIL

dannycoates:
> Maybe we should wontfix

No, setting hover effects when hover isn't supported creates unexpected and inconsistent results across browsers.

jodyheavener:
Shouldn't we do it for all links then?

dannycoates:
> Shouldn't we do it for all links then?

What links? where else do we define `:hover` styles?

jodyheavener:
I don't think we write any other custom hover styles, but we do have a handful of hover styles coming from Tailwind classes, e.g. [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/PageAvatar/Capture.tsx#L49:L51) and [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/AlertBar/index.tsx#L61:L61).

Issue #7461, opened 2021-02-09
Checkout: Adjust the location of the paymentError message for Paypal


data-sync-user:
FXA-2902 added some basic error handling during Paypal checkout if something goes wrong after the user clicks the Paypal button. We use the default error message banner that has white text on a red background.

Once we have updated the Paypal Checkout UI to match Jess' UX spec, we should also update the location of this error message, and the error message(s) for FXA-2976.

[~jecook@mozilla.com], looking at the current spec's screen for Checkout when there is no saved payment information, does it make sense to put this error message directly above the "Pay with Card" and "Paypal" buttons on this screen?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3000)
┆Issue Number: FXA-3000

data-sync-user:
➤ Bianca Danforth commented:

Closing in favor of FXA-2803.

Issue #7450, opened 2021-02-08
Checkout: Add a processing screen before the successful/failed authorization page


data-sync-user:
Per [Jess' Checkout UX Spec](https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=421%3A0), there's a new processing screen with a loading gif while the user is waiting for SubPlat to process their PayPal authorization (Note: Per Jess, we can use the existing SubPlat/FxA loading gif instead of styling it differently as in the spec).

Part of this ticket should include tests that the PayPal button is disabled when the button is clicked and re-enabled when the PayPal window is closed; see also FXA-2902.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2994)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15341/firefox_XP4gPPJIrr-1.png">firefox_XP4gPPJIrr-1.png</a>
┆Issue Number: FXA-2994

data-sync-user:
➤ Dave Justice commented:

I'm having some issues with the paypal flow when investigating this(there is some issues with their sandbox api). I'd like to witness what the paypal flow feels like before implementing this. We currently handle the loading status and error messaging within the payment form. Before moving things over to this page I want to ensure that the paypal page leaves enough time to show the loading state, and we can get access to the errors in a reasonable way to display on this page. Will revisit once sandbox services are back up.

Included screenshot of current loading/error states in payment form.

!firefox_XP4gPPJIrr.png!

data-sync-user:
➤ Dave Justice commented:

After getting the paypal flow to work locally this morning, we definitely need a processing indicator for the paypal flow. I'm gunna start working on adding that in today. And we can discuss how we want to handle the credit card form processing. It may make sense to loop back and replace the payment form's processing/error handling with this processing page for flow parity as a follow up.

Issue #7449, opened 2021-02-08
Modify the Subscribe page with no saved payment information to match the UX spec


data-sync-user:
This page needs to be updated per [Jess' UX Spec for Checkout](https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=421%3A0), for example the PayPal icon needs to be added to the PayPal button.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2993)
┆Issue Number: FXA-2993

data-sync-user:
➤ Bianca Danforth commented:

Included in this ticket is to handle the situation where we cannot render the PayPal button at all (i.e. `ButtonBase` is `null` in the `PaypalButton` React component). I talked to [~jecook@mozilla.com], and we agreed that in this situation, we should not render the section with the button at all and only display the Stripe payment option.

See https://github.com/mozilla/fxa/pull/7462 for more context.

data-sync-user:
➤ Bianca Danforth commented:

Blocked by FXA-2803 assuming the error message is on the same screen as opposed to its own screen. Pending final UX.

PR #7462, opened 2021-02-09
feat(payments): Add createOrder handler to PayPal button

biancadanforth:
## Because

- We want to initialize the PayPal transaction when the user clicks the PayPal button.

## This pull request

- Adds a `createOrder` handler to the `PaypalButton` component, which creates a new customer if one doesn't already exist and posts to a new `fxa-auth-server` route to get a Paypal checkout token.
- Fixes some Paypal feature flag tests that were passing despite the Paypal button not rendering in the tests. For some reason, the below assertion was passing even when `queryByTestId('paypal-button')` was `null`:
```js
waitForExpect(() =>
  expect(queryByTestId('paypal-button')).toBeInTheDocument()
);
```

Closes: #7269 

## Checklist

## Screenshots (Optional)

Screenshot_2021-02-02 Firefox Accounts

- The `PaypalButton` is mocked in the tests, since it is ordinarily a third-party React component rendered in a cross-origin iframe.
- This PR adds two negative UI integration tests for when the post to get the checkout token fails, and when the Paypal `createOrder` framework method (or really any Paypal framework method) fails. A corresponding positive test will be covered in #7450 .
- After we update the UI to match the Checkout spec, we will need to move the location of the default error messages; I filed #7461 for that.

### DIFF
--- a/packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
@@ -1,5 +1,9 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import ReactDOM from 'react-dom';
+
+import * as apiClient from '../../../lib/apiClient';
+import { Customer } from '../../../store/types';
+import { SubscriptionCreateAuthServerAPIs } from '../SubscriptionCreate';
 
 declare var paypal: {
   Buttons: {
@@ -7,13 +11,77 @@ declare var paypal: {
   };
 };
 
-const PaypalButtonBase = paypal.Buttons.driver('react', {
-  React,
-  ReactDOM,
-});
+export type PaypalButtonProps = {
+  apiClientOverrides?: Partial<SubscriptionCreateAuthServerAPIs>;
+  customer: Customer | null;
+  setPaymentError: Function;
+  idempotencyKey: string;
+  ButtonBase?: React.ElementType;
+};
 
-export const PaypalButton = () => {
-  return <PaypalButtonBase data-testid="paypal-button" />;
+export type ButtonBaseProps = {
+  createOrder?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  onError?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+};
+
+export const PaypalButtonBase =
+  typeof paypal !== 'undefined'
+    ? paypal.Buttons.driver('react', {
+        React,
+        ReactDOM,
+      })
+    : null;
+
+export const PaypalButton = ({
+  apiClientOverrides,
+  customer,
+  setPaymentError,
+  idempotencyKey,
+  ButtonBase = PaypalButtonBase,
+}: PaypalButtonProps) => {
+  const createOrder = useCallback(async () => {
+    try {
+      const { apiCreateCustomer, apiGetPaypalCheckoutToken } = {
+        ...apiClient,
+        ...apiClientOverrides,
+      };
+      if (!customer) {
+        await apiCreateCustomer({
+          idempotencyKey,
+        });
+      }
+      const { token } = await apiGetPaypalCheckoutToken();
+      return token;
+    } catch (error) {
+      setPaymentError(error);
+    }
+    return null;
+  }, [
+    apiClient.apiCreateCustomer,
+    apiClient.apiGetPaypalCheckoutToken,
+    customer,
+    setPaymentError,
+    idempotencyKey,
+  ]);
+
+  const onError = useCallback(
+    (error) => {
+      setPaymentError(error);
+    },
+    [setPaymentError]
+  );
+
+  return (
+    <>
+      {ButtonBase && (
+        <ButtonBase
+          data-testid="paypal-button"
+          createOrder={createOrder}
+          onError={onError}
+        />
+      )}
+    </>
+  );
 };
 
 export default PaypalButton;
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
@@ -18,6 +18,7 @@ import { PickPartial } from '../../../lib/types';
 import {
   defaultAppContextValue,
   MockApp,
+  MOCK_CHECKOUT_TOKEN,
   mockStripeElementOnChangeFns,
   elementChangeResponse,
 } from '../../../lib/test-utils';
@@ -28,12 +29,19 @@ import SubscriptionCreate, { SubscriptionCreateProps } from './index';
 
 import { updateConfig } from '../../../lib/config';
 
+import { ButtonBaseProps } from '../PayPalButton';
+
 // TODO: Move to some shared lib?
 const deepCopy = (object: Object) => JSON.parse(JSON.stringify(object));
 
 type SubjectProps = PickPartial<
   SubscriptionCreateProps,
-  'isMobile' | 'profile' | 'customer' | 'selectedPlan' | 'refreshSubscriptions'
+  | 'isMobile'
+  | 'profile'
+  | 'customer'
+  | 'selectedPlan'
+  | 'refreshSubscriptions'
+  | 'paypalButtonBase'
 >;
 
 const Subject = ({
@@ -105,6 +113,7 @@ const defaultApiClientOverrides = () => ({
   apiDetachFailedPaymentMethod: jest
     .fn()
     .mockResolvedValue(DETACH_PAYMENT_METHOD_RESULT),
+  apiGetPaypalCheckoutToken: jest.fn().mockResolvedValue(MOCK_CHECKOUT_TOKEN),
 });
 
 const PAYMENT_METHOD_RESULT = {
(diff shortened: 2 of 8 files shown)


biancadanforth:
@chenba , locally, I get a server test failure, though I guess this test isn't run in CI. All tests (including this one) passed locally with this patch before I rebased. The regression range is therefore between [4b60dfbbb](https://github.com/mozilla/fxa/commit/4b60dfbbba04c51e417101b2b62a7cbad7b65c3a) and [7ed1df85f](https://github.com/mozilla/fxa/commit/7ed1df85f6cc42eb12e6b8d2bbbaa1ced932f7e4).

Actually, I just pinpointed the regression to [chore(l10n): combine the download l10n scripts](https://github.com/mozilla/fxa/commit/37887cf115bd2c5ba9cabed55a436ea840cd7e2a), specifically the addition of [these two lines](https://github.com/mozilla/fxa/blob/7ed1df85f6cc42eb12e6b8d2bbbaa1ced932f7e4/packages/fxa-payments-server/server/lib/version.js#L82-L83) to `packages/fxa-payments-server/server/lib/version.js`.

Why isn't this run in CI? What would you recommend for fixing this test failure? Should that be handled in a separate patch?

```
 FAIL  server/lib/server.test.js
  Test simple server routes
    ✕ __version__ should have correct structure (44 ms)
    ✓ __lbheartbeat__ should return as expected (6 ms)
  Test route dependencies
    ✓ server.js should pass the correct dependencies to routes (7 ms)

  ● Test simple server routes › __version__ should have correct structure

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

      Array [
        "commit",
    -   "l10n",
        "source",
        "version",
      ]

      21 | 
      22 |           const body = response.body;
    > 23 |           expect(Object.keys(body).sort()).toStrictEqual([
         |                                            ^
      24 |             'commit',
      25 |             'l10n',
      26 |             'source',

      at lib/server.test.js:23:44
```

chenba:
> @chenba , locally, I get a server test failure, though I guess this test isn't run in CI?
> 
> Why isn't this run in CI? 

Looks like that test passed in CI: https://app.circleci.com/pipelines/github/mozilla/fxa/16278/workflows/b454623e-00c9-4bba-9a0c-cd3377dec223/jobs/216056 (sorry there's no good way to link to a line)

> What would you recommend for fixing this test failure? Should that be handled in a separate patch?
>

I'll fix it.  Sorry about that.
 
Also, if you have the latest `main` rebased on your branch, could you try `yarn postinstall && yarn test` from the payments-server dir?

biancadanforth:
Thanks Barry. You're right it does run in CI. That's good, hah.
> Also, if you have the latest main rebased on your branch, could you try yarn postinstall && yarn test from the payments-server dir?

Yes the test passed locally if I call `yarn postinstall` first (rebased off latest `main` from late last night) -- I guess that didn't run when I did `yarn install` at the fxa root after rebasing?

chenba:
> 
> 
> Thanks Barry. You're right it does run in CI. That's good, hah.
> 
> > Also, if you have the latest main rebased on your branch, could you try yarn postinstall && yarn test from the payments-server dir?
> 
> Yes the test passed locally if I call `yarn postinstall` first (rebased off latest `main` from late last night) -- I guess that didn't run when I did `yarn install` at the fxa root after rebasing?

Good to hear that the test passed.  No, because you already ran `yarn install` prior to my changes, yarn thinks that the packages have been built so it didn't run the postinstall step on them. ☹️

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
```diff
@@ -1,19 +1,89 @@
...
+  idempotencyKey: string;
+  ButtonBase?: React.ElementType;
+};
+
+export type ButtonBaseProps = {
+  createOrder?: (
+    event: React.MouseEvent<HTMLButtonElement, globalThis.MouseEvent>
+  ) => void;
+  onError?: (
+    event: React.MouseEvent<HTMLButtonElement, globalThis.MouseEvent>
```
chenba, line 27: I don't think we need `globalThis.` for the MouseEvent type.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
```diff
@@ -1,19 +1,89 @@
...
+      };
+      if (!customer) {
+        await apiCreateCustomer({
+          idempotencyKey,
+        });
+      }
+      const { token } = await apiGetPaypalCheckoutToken();
+      return token;
+    } catch (error) {
+      console.error('onCreateOrder failed', error);
```
chenba, line 60: Remove?

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
```diff
@@ -1,19 +1,89 @@
...
+  ) => void;
+};
+
+export const PaypalButtonBase =
+  typeof paypal !== 'undefined'
+    ? paypal.Buttons.driver('react', {
+        React,
+        ReactDOM,
+      })
+    : null;
```
chenba, line 37: Hmm what happens if `PaypalButtonBase` is null and no `ButtonBase` is passed in?

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
```diff
@@ -1,19 +1,89 @@
...
+    apiClient.apiCreateCustomer,
+    apiClient.apiGetPaypalCheckoutToken,
+    customer,
+    setPaymentError,
+    idempotencyKey,
+  ]);
+
+  const onError = useCallback(
+    (error) => {
+      console.error('PaypalButton method failed', error);
```
chenba, line 74: Remove?

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -96,6 +103,10 @@ const DETACH_PAYMENT_METHOD_RESULT = {
   foo: 'quux',
 };
 
+const GET_PAYPAL_CHECKOUT_TOKEN_RESULT = {
+  token: 'EC-8NC18566WJ1581100',
+};
```
chenba, line 108: Could use the mock value I saw above.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -656,6 +688,70 @@ describe('routes/ProductV2/SubscriptionCreate', () => {
       expect(refreshSubscriptions).toHaveBeenCalledTimes(0);
       expect(stripeOverride.confirmCardPayment).not.toHaveBeenCalled();
     });
+
+    it('displays apiGetPaypalCheckoutToken failure', async () => {
+      const MockedButtonBase = ({
+        createOrder = () => {},
```
chenba, line 694: nit: I don't think we need a default value for `createOrder` since `PayPalButton` always pass in one.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -656,6 +688,70 @@ describe('routes/ProductV2/SubscriptionCreate', () => {
...
+      });
+      expect(apiClientOverrides.apiGetPaypalCheckoutToken).toHaveBeenCalled();
+      expect(
+        screen.queryByTestId('error-payment-submission')
+      ).toBeInTheDocument();
+    });
+  });
+
+  it('displays PaypalButton onError failure', async () => {
+    const MockedButtonBase = ({ onError = () => {} }: ButtonBaseProps) => {
```
chenba, line 731: nit: also don't need a default value for `onError`

chenba (approved):
👍  r+ with comments

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
biancadanforth: I think this autocompleted when we were pairing together -- why don't we need it?

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
biancadanforth: I copied this from the [catch block](https://github.com/mozilla/fxa/blob/deace7808044003eca11c5bd67c1cf17aa9ac262/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx#L136-L139) of `handleSubscriptionPayment` for Stripe form submission. Should that also be removed? I agree it's not all that helpful in production.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
chenba: It's okay to leave it.  Or remove `globalThis.MouseEvent` altogether even, because apparently `React.MouseEvent` have a default type for that...which is just an alias to `globalThis.MouseEvent` (dun dun dun).

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
biancadanforth: We get a General Application Error modal currently. I guess an example of this would be if we fail to load the PayPal page script that renders the button?

I could conditionally render the `<ButtonBase>` like so:
```jsx
  return (
    <>
    {ButtonBase && (
      <ButtonBase
        data-testid="paypal-button"
        createOrder={createOrder}
        onError={onError}
      />
    )}
    </>
  );
```
It looks like I'd have to wrap it in a React Fragment. Would we want to log this error e.g. in Sentry? I guess we don't post `fxa-payments-server` errors to our Sentry Slack integrations anymore.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
chenba: Is it helpful to have those errors reported to Sentry?  We can leave them if they help with monitor or debugging.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
chenba: > I guess we don't post fxa-payments-server errors to our Sentry Slack integration anymore.

Not to Slack, but they are still reported to Sentry.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
biancadanforth: Okay I removed it. Yeah I noticed that, heh.

packages/fxa-payments-server/src/routes/Product/PayPalButton/index.tsx
biancadanforth: I removed the new ones I added. I don't think they're that helpful.

biancadanforth:
@chenba  I rerequested your review just for the comment here:
> Hmm what happens if PaypalButtonBase is null and no ButtonBase is passed in?

Do you think what I did is an acceptable solution?

chenba:
> 
> 
> @chenba I rerequested your review just for the comment here:
> 
> > Hmm what happens if PaypalButtonBase is null and no ButtonBase is passed in?
> 
> Do you think what I did is an acceptable solution?

Yeah, that works for me.  I guess it would mean that the end user will not see a PayPal option?  Maybe UX has some input on what to do here.

chenba (approved):

biancadanforth:
> Yeah, that works for me. I guess it would mean that the end user will not see a PayPal option? Maybe UX has some input on what to do here.

I talked to our designer, and she agreed we shouldn't render the button. Since this isn't the final UX, I added this edge case to the ticket to modify this page to ensure we don't render the section with the PayPal button at all if we can't render the button (#7449 ).

Note: It may take a bit for the [Jira ticket](https://jira.mozilla.com/browse/FXA-2993) to sync with GitHub for it to show up in the GitHub issue.

PR #7492, opened 2021-02-10
Add Avatar upload mutation

### LINKED ISSUE #5026: [avatar] "Confirm" photo functionality (process response/GQL)
On form submission, capture the user's input response and use a mutation to send the image information to the GQL server.

A successful response should 1) take the user back to the main settings page 2) with a success message triggering the global alert message while 3) updating the image on the main Settings page, including the one in the avatar drop down.

Any server errors should be displayed on the separate page that "Confirm" was selected on, in the global alert message.

9/16: I'm going to increase the points from a 3 to a 5 here. It may very well be a 3 but we haven't sent up an image to the GQL server yet and I'm unsure of any unforeseen complications.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1702)
┆Issue Number: FXA-1702

### LINKED ISSUE #7334: Organize avatar pages to be in one flow, like two-step auth component
https://github.com/mozilla/fxa/pull/7321#discussion_r562124521

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2940)
┆Issue Number: FXA-2940

meandavejustice:
## Because

- Users need to be able to update their avatar from the new settings page

## This pull request

- Adds some tweaks to the graphql client to allow file uploads
- Adds the mutations and some handling around capture/file chooser to make sure we have the proper image formats for upload
- Moves everything into one page flow file, satisfying #7334 
- Adds tests for the new `buttons.tsx` file

- fixes #7334 
- fixes #5026 

## Checklist

## Screenshots (Optional)


dannycoates (commented):
first few nits...

packages/fxa-settings/package.json
```diff
@@ -32,11 +32,14 @@
     ]
   },
   "dependencies": {
-    "@apollo/client": "^3.3.7",
+    "@apollo/client": "^3.2.5",
```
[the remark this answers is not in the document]
meandavejustice: Me and @bbangert had to downgrade this in order to get file uploads working during our pairing session.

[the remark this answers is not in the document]
dannycoates: interesting 🤔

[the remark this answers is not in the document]
bbangert: I believe this was due to an object mismatch in an underlying library. The `apollo-client-upload` wanted a different version of `apollo-client` in this case.

packages/fxa-settings/src/lib/gql.ts
```diff
@@ -45,9 +41,12 @@ export function createApolloClient(gqlServerUri: string) {
   // errorLink handles error responses from the server
   const errorLink = onError(errorHandler);
 
+  // uploadLink makes the actual requests to the server(httpLink with uploads)
+  const uploadLink = createUploadLink({ uri: `${gqlServerUri}/graphql` });
+
   const apolloClient = new ApolloClient({
     cache,
-    link: from([errorLink, authLink, httpLink]),
+    link: from([errorLink, authLink, uploadLink as any]),
```
bbangert, line 49: @dannycoates This is the unfortunate place where I believe we had mismatches, though it looks like we ended up forcing it by casting to `any` anyways. So maybe we don't need to downgrade the `apollo-client`?

If there's some way to get the right object version here without the cast to any, and fixup the version deps involved, that'd definitely be preferable to having to use this kludge. Perhaps we should at least flag the object version mismatch here so its clear why the cast to any to bypass it is done.

meandavejustice: `as any` is not needed here, this was an artifact from our debugging session, I've removed it

meandavejustice: @bbangert I just tested with 
```
-    "@apollo/client": "^3.2.5",
+    "@apollo/client": "^3.3.9",
```
And with the new version we need `as any` here. This seems more desirable than the outdated dependency

bbangert: Possibly, it depends whether any of the other libraries checking on this object are going to do an `isinstance` check, if they do, the version mismatch will throw them as its not an instance of the same class.

meandavejustice: Updated with note explaining why we have `as any` here. And that it should be removed once, `apollo-upload-client` gets itself in sync with `apollo/client`

packages/fxa-settings/src/components/PageAvatar/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+          rotation
+        );
+        updateAvatar({
+          variables: { file: croppedImage },
+        });
+      } else throw new Error('no image data');
+    } catch (e) {
+      console.error(e);
+    }
+  }, [croppedAreaPixels, rotation, capturedImgSrc, updateAvatar]);
```
dannycoates, line 128: I think we could refactor this and `save` below to be a little simpler, but it's not a blocker.

dannycoates (approved):
Works great. Nice work @meandavejustice 👍

packages/fxa-settings/src/lib/gql.ts
dannycoates: if we do a `yarn up @apollo/client apollo-upload-client` then `yarn dedupe` everything will be on `3.3.9` ... assuming there's no unrelated breaking changes in other deps from the dedupe.

Issue #7026, opened 2020-11-27
[FxA-Settings] No notification bar displayed for 2FA download/copy/print processes


vasilicatamas:
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Windows 10
Android 9
iOS

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on the "Add" button from 2FA section.
3. Enter the code received after the QR code was scanned and tap on "Continue".
4. Download/copy/print the codes.

**Expected results**
Download/copy/print notification bars are successfully displayed.

**Actual Results**
No notification bar is displayed for download/copy/print processes.

![issue61](https://user-images.githubusercontent.com/18114210/100453735-2789d600-30c4-11eb-9bbc-04dcb6325b62.gif)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2755)
┆Issue Number: FXA-2755

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [1.200.0] under Windows 10 64-bit, Android 9 and iOS 13.6. Download/copy/print confirmation tooltips are successfully displayed, but there is a minor overlapping issue tracked in a separate bug FXA-3043

PR #7498, opened 2021-02-10
feat(settings): display success tooltip on datablock action

chenba:
Because:
  - there is no feedback to users when they download/copy/print codes

This commit:
  - display a tooltip when a user download/copy/print
  - remove the copy feature from DataBlock
    - it's confusing? why have that and the copy feature in
      GetDataTrio?
    - it's not obvious that it's a button and there's no feedback so
      users wouldn't know they copied something
    - it uses a different separator for arrays than GetDataTrio
  - update DataBlock to include GetDataTrio by default
    - they are always used together already
    - we need to display the tooltip in DataBlock on an action in
      GetDataTrio

Closes: #7026

## Screenshots (Optional)
image


chenba:
Back to draft while I update a couple functional tests...

dannycoates (approved):
💬👍🏻

PR #7510, opened 2021-02-11
feat(payments): Add onApprove handler to PayPal button

biancadanforth:
## Because

- We want to process a successful PayPal authorization during Checkout.

## This pull request

- Adds an `onApprove` handler to the `PaypalButton` component, which posts to a new `fxa-auth-server` route to perform a bunch of tasks (see [PayPal Checkout ladder diagram](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#paypal-checkout)), including creating a new subscription and attempting to capture the first payment.

Closes: #7420 

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/lib/test-utils.tsx
+++ b/packages/fxa-payments-server/src/lib/test-utils.tsx
@@ -311,6 +311,11 @@ export const MOCK_CHECKOUT_TOKEN = {
   token: 'EC-8NC18566WJ1581100',
 };
 
+export const MOCK_PAYPAL_SUBSCRIPTION_RESULT = {
+  sourceCountry: 'FR',
+  subscription: { what: 'ever' },
+};
+
 export const STRIPE_FIELDS = [
   'cardNumberElement',
   'cardCVCElement',
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
@@ -19,6 +19,7 @@ import {
   defaultAppContextValue,
   MockApp,
   MOCK_CHECKOUT_TOKEN,
+  MOCK_PAYPAL_SUBSCRIPTION_RESULT,
   mockStripeElementOnChangeFns,
   elementChangeResponse,
 } from '../../../lib/test-utils';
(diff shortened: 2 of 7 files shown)


packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -311,6 +311,11 @@ export const MOCK_CHECKOUT_TOKEN = {
   token: 'EC-8NC18566WJ1581100',
 };
 
+export const MOCK_PAYPAL_SUBSCRIPTION_RESULT = {
+  sourceCountry: 'Equatorial Kundu',
+  subscription: { what: 'ever' },
```
rlr, line 316: lol

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -489,6 +490,39 @@ describe('routes/ProductV2/SubscriptionCreate', () => {
...
+            paypalButtonBase: MockedButtonBase,
+          }}
+        />
+      );
+    });
+    await act(async () => {
+      fireEvent.click(screen.getByTestId('paypal-button'));
+    });
+    expect(apiClientOverrides.apiCapturePaypalPayment).toHaveBeenCalledTimes(1);
+    expect(refreshSubscriptions).toHaveBeenCalledTimes(1);
```
rlr, line 524: should we check that no errors were rendered?

rlr (approved):
LGTM!

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -311,6 +311,11 @@ export const MOCK_CHECKOUT_TOKEN = {
   token: 'EC-8NC18566WJ1581100',
 };
 
+export const MOCK_PAYPAL_SUBSCRIPTION_RESULT = {
+  sourceCountry: 'Equatorial Kundu',
```
chenba, line 315: Is this also made up or does PayPal return a full country name instead of a two character code country code?

chenba (approved):

packages/fxa-payments-server/src/lib/test-utils.tsx
biancadanforth: This is a [two-character country code per the fxa-auth-server validator](https://github.com/mozilla/fxa/blob/021fd5d4485aebf174fe668e3d6db15c6ae7cf1f/packages/fxa-auth-server/lib/routes/validators.js#L288-L291).

My client side TS only asserts it's a string. Looks like fxa-auth-server uses `hapi` to add more specificity here, but I don't see similar validation on the client in `apiClient.ts`. I can change this string in my test to be consistent however.

biancadanforth: Would that suffice here, or is there more I should do @chenba ?

chenba: Oh it wasn't a request for a change.  I was just curious.  I wasn't sure if you copied that from a (sandbox) response or something.

biancadanforth: I just put a made up country from [The West Wing](https://en.wikipedia.org/wiki/The_West_Wing). 😆

Issue #7471, opened 2021-02-09
Update paypal SetupExpressCheckout for currency option


data-sync-user:
The SetupExpressCheckout function should accept an argument to pass through for the payment currency. This should be populated in the create subscription handler and the update subscription handler to pass through the chosen currency from the plan.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3007)
┆Issue Number: FXA-3007

PR #7511, opened 2021-02-11
(feat)auth-server: Add define-able currencies to paypal checkout

birdsarah:
## Because

- We need to pass through the currency from the selected plan to paypal

## This pull request

- Makes the currency code configurable on the paypal-client SetExpressCheckout and follows back down the chain of calls adding the currencyCode as appropriate.

Closes: #7471

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
@@ -41,7 +41,9 @@ export class PayPalHandler extends StripeHandler {
     this.log.begin('subscriptions.getCheckoutToken', request);
     const { uid, email } = await handleAuth(this.db, request.auth, true);
     await this.customs.check(request, email, 'getCheckoutToken');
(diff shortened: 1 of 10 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -158,6 +160,11 @@ export const paypalRoutes = (
             token: isA.string().required(),
           }),
         },
+        validate: {
+          payload: {
+            currencyCode: isA.string().required(),
```
bbangert, line 165: I was originally wondering why the client is choosing the currency, but I see the only alternative at this point is to have the plan id sent. Since we're going to force a currency match later anyways, it doesn't seem to matter much. 😄

bbangert (approved):
LGTM! 👍

bbangert:
@birdsarah heh, my other PR caused a conflict here, sorry!

birdsarah:
> @birdsarah heh, my other PR caused a conflict here, sorry!

No sweat :D

PR #7512, opened 2021-02-11
fix(settings): account for sessionToken.mustVerify in new settings

### LINKED ISSUE #6240: Case where a verified session is required to get attached clients
Steps to reproduce:

* start firefox with a new profile
* sign in to sync via the account toolbar
* you should be redirected to `/signin_token_code`
* browse to `/beta/settings`

Note that signing in to a fresh profile via _browsing_ to the signin page does _not_ trigger this behavior

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2446)
┆Issue Number: FXA-2446

dannycoates:
There's a special token status of `mustVerify` that requires the session to be verified before it can be used at all. We need to reject any requests from unverified sessions when it is set.

The frontend should redirect to the `signin_token_code` page when this happens.

fixes #6240

ref https://github.com/mozilla/fxa/blob/67c3afa3ea38826d7b4b9f78fa3d706cfa07ecca/packages/fxa-auth-server/lib/routes/emails.js#L175-L181

### DIFF
--- a/packages/fxa-graphql-api/src/auth/session-token.strategy.spec.ts
+++ b/packages/fxa-graphql-api/src/auth/session-token.strategy.spec.ts
@@ -37,11 +37,23 @@ describe('SessionTokenStrategy', () => {
     });
   });
 
-  it('throws unauthorized', async () => {
+  it('throws unauthorized for an invalid token', async () => {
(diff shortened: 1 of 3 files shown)


chenba (approved):
LGTM

PR #7513, opened 2021-02-11
feat(auth-server): add transaction search to paypal

### LINKED ISSUE #7238: Add TransactionSearch command to PayPalHelper
TransactionSearch should be added to the PayPalHelper, which should provide a typesafe API and then call a new underlying operation on the PayPal NVP API client. The INVNUM specifically needs to be including as a API argument to ensure we can locate transactions by the INVNUM that we assigned when running them.

PayPal docs for this API operation:

https://developer.paypal.com/docs/nvp-soap-api/transaction-search-nvp/

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2883)
┆Issue Number: FXA-2883

bbangert:
Because:

* We want to search transactions for a given user/invoice.

This commit:

* Adds a paypal transaction search API command.

Closes #7238

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-client.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-client.ts
@@ -96,6 +96,30 @@ type BAUpdateData = {
   PAYERSTATUS: string;
 };
 
+export type TransactionStatus =
+  | 'Pending'
+  | 'Processing'
+  | 'Completed'
+  | 'Denied'
+  | 'Reversed';
+
+type TransactionSearchResult = {
+  AMT: string;
+  CURRENCYCODE: string;
+  EMAIL: string;
+  FEEAMT: string;
+  NAME: string;
+  NETAMT: string;
+  STATUS: TransactionStatus;
+  TIMESTAMP: string;
+  TRANSACTIONID: string;
+  TYPE: string;
+};
+
+type TransactionSearchData = {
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -358,4 +401,20 @@ export class PayPalClient {
...
+
+  public async transactionSearch(
+    options: TransactionSearchOptions
+  ): Promise<NVPTransactionSearchResponse> {
+    const data = {
+      STARTDATE: toIsoString(options.startDate),
+      ENDDATE: options.endDate ? toIsoString(options.endDate) : undefined,
+      EMAIL: options.email,
+      INVNUM: options.invoice,
+      TRANSACTIONID: options.transactionId,
```
chenba, line 413: huh I see in the type declaration above that only `startDate` is not optional, so I'm kinda surprised that the type checker didn't complain about lines 411-413.  (I don't have the branch open locally; I'm sure there's a good reason.)

chenba (approved):
✔️

packages/fxa-auth-server/lib/payments/paypal-client.ts
bbangert: They'll just be undefined in that case. In TS you can still use optional types like this.

Issue #7099, opened 2020-12-08
Checkout: Add success screen upon successful PayPal authorization and captured payment


data-sync-user:
This screen currently displays billing information for Stripe including an image for the card (e.g. Visa) and an "ending with XXXX" string for the last four digits of the card number.

The screen should be updated to show any PayPal billing information instead for a PayPal customer. What billing information to show is an open question.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2802)
┆Issue Number: FXA-2802

PR #7526, opened 2021-02-12
feat(payments): Modify PaymentConfirmation UI according to payment_provider

lesleyjanenorton:
## Because

- We want to vary the payment confirmation UI according to the `payment_provider`.

## This pull request

- Adds PayPal logo 
- Updates PaymentConfirmation component to:
  - When payment_provider is "paypal":
    - Show the PayPal logo instead of the credit card image and last 4 digits info
    - Hide the row (`className = 'billing-info'`) containing the user's displayName or email
- Updates storybook to reflect changes

- Nit fixes: Resizes the gradient check on the confirmation page to match the the new spec.

Closes: #7099 

## Checklist

## Screenshots (Optional)
After checking out with PayPal:
Screen Shot 2021-02-12 at 2 31 58 PM

After checking out with Stripe:
Screen Shot 2021-02-12 at 2 32 08 PM


packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -56,7 +59,11 @@ export const PaymentConfirmation = ({
       data-testid="payment-confirmation"
     >
       <header>
-        <img src={circledCheckbox} alt="circled checkbox" />
+        <img
+          className="circled-check"
+          src={circledCheckbox}
+          alt="circled checkbox"
```
rlr, line 65: Not your fault but this alt text isn't very useful and it isn't localized. What do you think? Maybe empty alt tag?

rlr (commented):
I glanced over this and LGTM 👍

chenba:
> Nit fixes: Resizes the gradient check on the confirmation page to match the the new spec.

zomg finally thank you 💯

packages/fxa-payments-server/src/components/PaymentConfirmation/index.test.tsx
```diff
@@ -78,6 +102,55 @@ describe('PaymentConfirmation', () => {
...
+            profile: userProfile,
+            selectedPlan,
+            customer: paypalCustomer,
+            productUrl,
+          }}
+        />
+      );
+    };
+
+    it('the paypal logo is visible', () => {
```
chenba, line 119: nit: I think the convention is to start with a verb so that it reads `[the thing in describe][does this thing]`.  Not that we've been very consistent with it.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -25,7 +25,10 @@ export const PaymentConfirmation = ({
 }: PaymentConfirmationProps) => {
   const { amount, currency, interval, interval_count } = selectedPlan;
   const { displayName, email } = profile;
-  const { brand, last4, subscriptions } = customer;
+
+  const { brand, last4, payment_provider, subscriptions } = customer;
+  const isPaypalPayment = payment_provider === 'paypal';
```
chenba, line 30: Would be nice if we have this in a lib somewhere.  Currently we have some logic checking to see if a customer has an existing CC that's duplicated in two places; preferably we don't do that with this too.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
chenba: Maybe leave it and file a follow-up issue to fix it?  I do agree that the alt text isn't very helpful.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -103,11 +112,19 @@ export const PaymentConfirmation = ({
...
             <Localized id="payment-confirmation-cc-preview" vars={{ last4 }}>
               <p className={`c-card ${brand.toLowerCase()}`}></p>
             </Localized>
           )}
+          {isPaypalPayment && (
+            <div
+              className={`payment-logo ${payment_provider}`}
+              data-testid="paypal-logo"
+            >
+              {payment_provider}
```
chenba, line 125: Should this be 'PayPal' not 'paypal'?

chenba (approved):
👍 r+ with comments

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
lesleyjanenorton: I wondered about this also! I used _paypal_ here for consistency since we're using _paypal_ and _stripe_ (instead of _PayPal_ and _Stripe_ ) elsewhere. I don't have a strong opinion about it though.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.test.tsx
```diff
@@ -61,6 +61,30 @@ const customer: Customer = {
   ],
 };
 
+const paypalCustomer: Customer = {
+  billing_name: 'Pay Pal Doe',
+  payment_provider: 'paypal',
+  payment_type: 'credit',
+  last4: '7777',
```
birdsarah, line 68: paypalCustomer shouldn't have these credit card options right?

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
lesleyjanenorton: Just filed https://jira.mozilla.com/browse/FXA-3047

PR #7542, opened 2021-02-16
chore(payments): Add a nonce to load PayPal inline script elements

rlr:
## Because

- Paypal's script injects an inline script which produces a CSP error because inline scripts aren't allowed.

## This pull request

- Creates a nonce that is sent in our `script-src` CSP headers and also passed to paypal so they can use it to inject the inline script.

Closes: #7350 

## Checklist

## Screenshots (Optional)


packages/fxa-payments-server/server/lib/server.js
```diff
@@ -153,6 +154,17 @@ module.exports = () => {
...
   if (config.get('csp.enabled')) {
+    app.use(function (req, res, next) {
+      // Generate nonce for CSP to allow paypal inline script.
+      res.paypalCspNonce = uuid();
+      next();
+    });
+
+    // Add nonce for paypal's inline script.
+    cspRulesBlocking.directives.scriptSrc.push((req, res) => {
+      return `'nonce-${res.paypalCspNonce}'`;
```
chenba, line 165: Not knowing express intimately and just from looking at this, I don't actually know if the function passed into `app.use` on line 157 has executed by the time when this line executes.  It does appear to work.  Is there doc where I can find this info?

rlr: `app.use()` basically adds middleware to the request handling. They will execute in the order they are declared. So, the middleware that sets the nonce above will run right before the csp middleware which is set up below on line 168.

From https://expressjs.com/en/guide/writing-middleware.html:

> The order of middleware loading is important: middleware functions that are loaded first are also executed first.

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -153,6 +154,17 @@ module.exports = () => {
   );
 
   if (config.get('csp.enabled')) {
+    app.use(function (req, res, next) {
+      // Generate nonce for CSP to allow paypal inline script.
+      res.paypalCspNonce = uuid();
+      next();
```
chenba, line 160: I think we need this on every page, but I'm not certain.  @biancadanforth We need to load the PayPal JS SDK on the subscription management page so we can have a link/button to take the customer to PayPal correct?

rlr: yeah, right now this is running on every single request. But we can add some paths so it only runs on certain paths.

packages/fxa-payments-server/server/lib/server.js
chenba: Right, that I do remember.  But `cspRulesBlocking.directives.scriptSrc.push` is not `app.use`.  Now that I re-read that bit about middleware order of execution, I see why it is working.  👍

chenba (approved):
WFM

biancadanforth (commented):
Thanks Ricky for this PR.

While the UUID does get regenerated on hard refresh, I am still seeing `Content Security Policy: The page’s settings blocked the loading of a resource at inline ("script-src").` for cached reloads of the `index.html` page.

I do see the `nonce-` CSP header get added to the response for index.html, but its value does not match the value added to the PayPal page script when the page is from the browser cache.

Is this worth doing something about? What could we do?

For one thing, I noticed we’re not passing in any values to recompute `useEffect` where we add the nonce to the script tag. Perhaps we’re using a stale UUID here sometimes? Is there a variable we can watch for to ensure the CSP header and nonce passed into the app always match?

packages/fxa-payments-server/server/lib/server.js
biancadanforth: I don't think we would use the PayPal JS SDK for Subscription Management per https://jira.mozilla.com/browse/FXA-2806, which suggests there's an external URL we would send people to for changing their payment method. I haven't looked into the Subscription Management part as closely yet...

biancadanforth: @bbangert is there any reason we'd need to use the Smart Button for Subscription Management? Maybe there's a scenario I'm forgetting...

chenba:
> While the UUID does get regenerated on hard refresh, I am still seeing...

Good catch.  We might need to disable etag.

rlr:
> > While the UUID does get regenerated on hard refresh, I am still seeing...
> 
> Good catch. We might need to disable etag.

ah, so the page can be cached? How do we disable that just for the payment page?

packages/fxa-payments-server/server/lib/server.js
```diff
@@ -250,9 +266,14 @@ module.exports = () => {
 
     INDEX_ROUTES.forEach((route) => {
       // FIXME: should set ETag, Not-Modified:
-      app.get(route, (req, res) => {
+      app.get(route, noCache(), (req, res) => {
```
rlr, line 269: @chenba We added this `nocache` middleware to the index routes. LMK us you are cool with that or not. Thanks!

biancadanforth: Is it possible to only nocache if we're on this `SubscriptionCreate` Product route? Do we care?

chenba: According to the `helmet` docs, we can turn off etags with `app.use(helmet.noCache({ noEtag: true }))`.  (`app.use(helmet.noCache())` is also a thing.)

> 
> 
> Is it possible to only nocache if we're on this `SubscriptionCreate` Product route? Do we care?

While I don't like the idea of no caching where it is helpful, I don't think we should worry too much here.  Payments doesn't get _that_ much traffic in the grand scheme of things.

rlr: yeah, those are my thoughts too. 1) This is probably the least trafficked public website at Mozilla. 2) It is just rendering a static HTML template with a few string replaces. 3) And, you only go through checkout once in a lifetime and will never need to go back. Hopefully :)

biancadanforth: > According to the helmet docs, we can turn off etags with app.use(helmet.noCache({ noEtag: true })). (app.use(helmet.noCache()) is also a thing.)

So Helmet moved out their `noCache` method to a separate Node package (we tried just using `helmet.noCache()` and it threw a descriptive error). Based on the [source code](https://github.com/helmetjs/nocache/blob/dba4c205867e77be4b55537b0844c7ea9ebc6c5a/index.ts#L3), there are no parameters for the exported function. Would the headers set by `noCache()` change the behavior substantially versus just removing the ETag header?

rlr: Express won't generate the `Etag` since we are setting the `CacheControl` header to `no-cache`. So it ends up having the same effect.

chenba: I think simply removing the etag header is fine.  As far as I can tell, we are not currently sending any other caching headers.

chenba (approved):

packages/fxa-payments-server/server/lib/server.js
biancadanforth: Per `noCache`'s [README](https://github.com/helmetjs/nocache/blob/dba4c205867e77be4b55537b0844c7ea9ebc6c5a/README.md), they set a bunch of cache-related headers:

> This sets four headers, disabling a lot of browser caching:
> - `Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate`
> - `Pragma: no-cache`
> - `Expires: 0`
> - `Surrogate-Control: no-store`

PR #7547, opened 2021-02-17
feat(auth-server): add metrics for paypal client

### LINKED ISSUE #7522: Add paypal client response emitter with helper metrics
The paypal client should have an emitter on it that works similarly to the stripe one, so that metrics can be augmented in from the paypal helper. The helper should log metrics for it similarly to how the stripe helper's does.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3037)
┆Issue Number: FXA-3037

bbangert:
Because:

* We want to gauge response times from PayPal and track client errors.

This commit:

* Adds an event emitter similar to Stripes so that the paypal helper
  can augment it with metric reporting.

Closes #7522

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-client.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-client.ts
@@ -4,6 +4,7 @@
 import pRetry from 'p-retry';
 import querystring from 'querystring';
 import superagent from 'superagent';
+import { EventEmitter } from 'events';
 
 export const PAYPAL_SANDBOX_BASE = 'https://api-3t.sandbox.paypal.com';
 export const PAYPAL_SANDBOX_IPN_BASE = 'https://ipnpb.sandbox.paypal.com';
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -201,6 +202,15 @@ export function isIpnMerchPmt(
   return ['merch_pmt', 'mp_cancel'].includes(ipnMessage.txn_type);
 }
 
+type ResponseEventType = {
+  error?: Error;
+  request_end_time: number;
+  version: string;
+  elapsed: number;
+  method: string;
+  request_start_time: number;
```
chenba, line 211: start time, end time, _and_ elapsed seem to be gilding the lily a bit.  Or at least I didn't see where we need the start and end time aside from calculating the elapsed time.

chenba (approved):
⏲️

bbangert:
I put those in mainly as the stripe response emitter had that too, maybe we'll want it later. 🤷

packages/fxa-auth-server/lib/payments/paypal-client.ts
bbangert: Heh, TIL, 'gilding the lily', didn't remember that one!

PR #7562, opened 2021-02-17
feat(auth-server): adds logic to handleMerchPayment method to PayPalNotificationHandler

### LINKED ISSUE #7277: Handle IPN transction type for payment
When we get a IPN transaction type of ``merch_pmt``, we should use the invoice number on it to update the appropriate invoice in Stripe if it's been paid.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2907)
┆Issue Number: FXA-2907

say-yawn:
## Because

- We need to update invoice based on the merchant payment IPN from PayPal

## This pull request

- Adds logic to update Stripe invoice based on the merchant payment IPN

Closes: #7277

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -11,7 +13,60 @@ import { AuthLogger, AuthRequest } from '../../types';
...
+      case 'Processed':
+        return this.stripeHelper.payInvoiceOutOfBand(invoice);
+      case 'Pending':
+      case 'In-Progress':
+        return;
+      case 'Denied':
+      case 'Failed':
+      case 'Voided':
+      case 'Expired':
+        return this.stripeHelper.updatePaymentAttempts(invoice);
```
bbangert, line 57: I think we're going to need a more idempotent function here that takes into account the details of the transaction causing this. Because we update the payment attempts in `processInvoice` and we *may* need to update them here.

bbangert (changes requested):
We need an idempotent payment attempt updater. Also looks like a syntax error in the tests.

packages/fxa-auth-server/test/local/routes/subscriptions/paypal-notifications.js
```diff
@@ -155,4 +156,134 @@ describe('PayPalNotificationHandler', () => {
       sinon.assert.calledOnce(log.error);
     });
   });
+
+  describe('handleMerchPayment', () => {
+    const message = {
+      txn_type: 'merch_pmt',
+      invoice: 'inv-000',
+    };
```
bbangert, line 164: It would be nice to have a complete merch_pmt sample message with all the fields they send.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -11,7 +13,77 @@ import { AuthLogger, AuthRequest } from '../../types';
...
+        message: 'Invoice not found',
+      });
+    }
+
+    if (invoice.status == null || !['draft', 'open'].includes(invoice.status)) {
+      this.log.error('handleMerchPayment', {
+        message: 'Invoice not in draft or open status',
+        ipnMessage: message,
+      });
+      throw error.internalValidationError('handleMerchPayment', {
```
bbangert, line 39: When we run the payment during checkout and it succeeds, we immediately pay the invoice and its in a paid status. Which means this event will throw an error. The easy way to know we can ignore this transaction would be to see if the invoice metadata has a transaction id that matches the transaction id on this message, if so it was paid already and we can stop.

bbangert (changes requested):
Minor tweak to avoid throwing an error and I think this will be perfect!

bbangert (approved):
👍

PR #7579, opened 2021-02-19
refactor(settings): refactor flow parameters to avoid /get_flow

dannycoates:
ref: https://jira.mozilla.com/browse/FXA-2542

Passing around flow parameters is ... tricky. We want to avoid redirecting navigation to /beta/settings through /get_flow, which we can do by making a fetch to /metrics-flow, except that doesn't get us quite all of the vars we need so there's a little more work to initialize those values in metrics.ts. This gets us to parity with /get_flow and we can eliminate that route.

Also, /get_flow didn't actually fulfill the hope of persisting flowId etc across the boundary from old-settings to new-settings since those values are initialized from the html on the page, i.e. get_flow was a new page load so it would always get a new flowId. The least intrusive way to pass the flowId across is via the query string of the link to beta settings, which this PR adds.


6a68 (approved):
LGTM, nice work! Side note: I look forward to rethinking the flow params at some point and making sure we even really need these to do flow analyses on the backend.

PR #7587, opened 2021-02-22
task(settings): do not show "take a photo" option on mobile

### LINKED ISSUE #5045: [avatar] Ensure mobile flow matches wires & functions as expected
The PR for this should at the very least make a tweak for displaying one "Upload" button if the user is on a mobile device rather than both "Upload" and "Take photo".

Ensure that on click, a native prompt asks the user if they'd like to take a new photo or use file upload. Double check this all works as expected, including removing the photo and rotating/zooming an uploaded photo, and make any needed tweaks for a smooth mobile experience.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1717)
┆Issue Number: FXA-1717

dannycoates:
## Because

- the figma designs exclude the "Take photo" option on mobile

closes #5045

### DIFF
--- a/packages/fxa-settings/src/components/PageAvatar/index.tsx
+++ b/packages/fxa-settings/src/components/PageAvatar/index.tsx
@@ -13,6 +13,7 @@ import LoadingSpinner from 'fxa-react/components/LoadingSpinner';
 import 'react-easy-crop/react-easy-crop.css';
 
 import { cache, sessionToken } from '../../lib/cache';
+import { isMobileDevice } from '../../lib/utilities';
 import { useAccount } from '../../models';
 import { HomePath } from '../../constants';
 import firefox from '../../lib/firefox';
@@ -297,7 +298,7 @@ export const PageAddAvatar = (_: RouteComponentProps) => {
                     }}
                   />
                 )}
-                {!mediaError && (
+                {!mediaError && !isMobileDevice() && (
                   <TakePhotoBtn
                     capturing={capturing}
                     onClick={() => {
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/components/PageAvatar/index.tsx
```diff
@@ -44,6 +44,7 @@ export const PageAddAvatar = (_: RouteComponentProps) => {
   const { avatar } = useAccount();
   const alertBar = useAlertBar();
   const [saveEnabled, setSaveEnabled] = useState(false);
+  const isMobile = /mobi/i.test(navigator.userAgent);
```
6a68, line 47: Hmm, what about `isMobileDevice` in `lib/utilities`? Let's merge this check into that check or just prefer that check.

6a68 (changes requested):
LGTM, one question about where the mobile check should live

packages/fxa-settings/src/components/PageAvatar/index.tsx
dannycoates: Thanks! I didn't think to check if we had one already

dannycoates: IDK how I feel about the combined approach but, meh 🤷

6a68: lol, useragent sniffing is an ugly hack regardless, this LGTM

6a68 (approved):

PR #7588, opened 2021-02-22
chore(settings): ignore settings.ftl

dannycoates:
## Because

- this file is a build artifact so we don't need it committed to this repo


6a68 (approved):
Thanks! 🍻

dannycoates:
We need to add a build command for fxa-settings here before merging: https://github.com/mozilla/fxa-content-server-l10n/blob/master/scripts/extract-and-pull-request.sh#L9

Issue #7030, opened 2020-11-27
[FxA-Settings] No notification bar displayed for several processes


vasilicatamas:
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Windows 10 64-bit
Android 9
iOS 13.6

**Steps to reproduce**
1.Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2.Add a Display name/ Update the Display name/ Change the password/ Add a secondary email/ Add a recovery key/ Change 2FA codes

**Expected results**
The notification bar is displayed for all completed processes.

**Actual Results**
Several processes do not display a green confirmation bar.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2759)
┆Issue Number: FXA-2759

data-sync-user:
➤ Danny Coates commented:

triggering the alertbar after navigating from a secondary page requires some refactoring of the alertbar component. prob more than a 3

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.201.0] under Windows 10 64-bit, Android 9 and iOS 14.4. I confirm that the green confirmation bar is displayed for all process, excepting the avatar setup and connected services sign-out process.

PR #7593, opened 2021-02-22
Show alert bar after completing tasks

6a68:
## Because

- we're supposed to show an alert bar in the settings page after completing tasks in individual sub-pages
- the sub-pages sometimes try to show an alert, but it isn't displayed before the user navigates away

## This pull request

- updates all sub-pages to use `alertTextExternal` to display a notification in the base settings page
- a couple of tests still need to be updated (this work has been deferred to #7596)

Closes: #7030

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-settings/src/components/PageRecoveryKeyAdd/index.tsx
+++ b/packages/fxa-settings/src/components/PageRecoveryKeyAdd/index.tsx
@@ -4,7 +4,7 @@ import base32encode from 'base32-encode';
 import { useForm } from 'react-hook-form';
 import { RouteComponentProps, useNavigate } from '@reach/router';
 import { useRecoveryKeyMaker } from '../../lib/auth';
-import { cache, sessionToken } from '../../lib/cache';
+import { alertTextExternal, cache, sessionToken } from '../../lib/cache';
 import { useAlertBar } from '../../lib/hooks';
 import { useAccount } from '../../models';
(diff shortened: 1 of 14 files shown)


packages/fxa-settings/public/locales/en-US/settings.ftl
```diff
@@ -247,7 +250,7 @@ delete-account-password-input =
  .label = Enter password
 
 delete-account-cancel-button = Cancel
-delete-account-delete-button = Delete Account
+delete-account-delete-button = Delete
```
6a68, line 253: I think this is a stray line from a different bug

dannycoates: I just merged the PR to gitignore this file

packages/fxa-settings/src/components/PageRecoveryKeyAdd/index.tsx
```diff
@@ -80,6 +85,7 @@ export const PageRecoveryKeyAdd = (_: RouteComponentProps) => {
     },
   });
   useEffect(() => {
+    // TODO what's this?
```
dannycoates, line 88: this is for if you arrive at this page with an invalid state like by doing a hard refresh on step 2

dannycoates (approved):

Issue #7591, opened 2021-02-22
[l10n] Partial translation not displayed in Support page


data-sync-user:
https://github.com/mozilla/fxa-content-server-l10n/issues/430

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3071)
┆Issue Number: FXA-3071

Issue #7592, opened 2021-02-22
[l10n] Subscription support form: Same string reused in different (cap or lower) case may break code


data-sync-user:
https://github.com/mozilla/fxa-content-server-l10n/issues/423

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3072)
┆Issue Number: FXA-3072

PR #7602, opened 2021-02-23
fix(l10n): fix support form l10n

chenba:
Because:
 - there are various issues with l10n on the subscription support form

This commit:
 - use translated strings directly in templates
 - stop lowercasing translated strings with JS

Closes: #7591, #7592 (l10n issues: https://github.com/mozilla/fxa-content-server-l10n/issues/430 and https://github.com/mozilla/fxa-content-server-l10n/issues/423)

### DIFF
--- a/packages/fxa-content-server/app/scripts/models/support-form.js
+++ b/packages/fxa-content-server/app/scripts/models/support-form.js
@@ -17,24 +17,19 @@ const TOPICS = [
   'Not listed',
 ];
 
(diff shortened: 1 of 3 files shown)


chenba:
> chenba requested a review from biancadanforth now

I guess technically this is part of SubPlat.  But it lives on the content-server.  🤷

6a68 (approved):
LGTM, thanks!

PR #7604, opened 2021-02-23
feat(settings): replace connected service display name for Mac OS X with macOS on retrieval

### LINKED ISSUE #7482: [FxA-Settings] Outdated nomenclature for macOS in Connected Services
## Description

Connected Services says “Mac OS X”, which I don't think has been a thing in quite a while (2016, per https://en.wikipedia.org/wiki/MacOS) 

-  Apple shortened the name to "OS X" in 2012 and then changed it to "macOS" in 2016, adopting the nomenclature that they were using for their other operating systems, iOS, watchOS, and tvOS. In 2020, Apple incremented the major version number to 11 in macOS Big Sur, although it retains the same base as the previous versions of macOS.
- 
Submitted by: pdehaan

### Steps to reproduce

- Manage account
- Connected services

### Expected result

macOS

### Actual result

macOS X

### Environment

macOS

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3019)
┆Issue Number: FXA-3019

dannycoates:
closes #7482

### DIFF
--- a/packages/fxa-auth-server/lib/routes/attached-clients.js
+++ b/packages/fxa-auth-server/lib/routes/attached-clients.js
@@ -206,6 +206,7 @@ module.exports = (log, db, devices, clientUtils) => {
           if (client.deviceId && !client.deviceType) {
             client.deviceType = 'desktop';
           }
+          client.name = client.name.replace('Mac OS X', 'macOS');
         }
 
         return attachedClients;


6a68 (approved):
Yup, I _think_ we're good here. The string "Mac OS X" sure shows up in a lot of places in auth and auth-db, but its occurrences seem to be in generic text fields in mysql and various models. I don't think changing the client.name will break anything, and it doesn't look like client.name is used in our metrics.

PR #7605, opened 2021-02-23
fix(settings): fixed casing for new settings strings

dannycoates:
## Because

- The designs use sentence case not title case for most strings


6a68 (approved):
Good catch. We should probably open a PR over in l10n repo with the updated strings. cc @peiying2

peiying2:
> Good catch. We should probably open a PR over in l10n repo with the updated strings. cc @peiying2

Looks good. Normally a revision in the string requires a new string ID as reuse of the string ID will not trigger the change in Pontoon. If the string is localized, the source string change will not be shown as a new string for review or revision. 
Having said that, these changes may likely affect English only, including en-CA, en-GB. They may miss the changes. I will send an email on this so localizers can be on t he lookout.

PR #7608, opened 2021-02-23
L10n fixes

6a68:
This PR fixes a few issues noticed by our localizers.

Fixes mozilla/fxa-content-server-l10n#470, fixes mozilla/fxa-content-server-l10n#473.

### DIFF
--- a/packages/fxa-settings/src/components/PageAvatar/en-US.ftl
+++ b/packages/fxa-settings/src/components/PageAvatar/en-US.ftl
@@ -20,8 +20,8 @@ avatar-page-rotate-button = Rotate
 avatar-page-camera-error = Could not initialize camera
 avatar-page-new-avatar =
   .alt = new profile picture
-avatar-page-file-upload-error = There was a problem uploading your profile picture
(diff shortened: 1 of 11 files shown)


packages/fxa-settings/src/components/PageAvatar/en-US.ftl
```diff
@@ -20,8 +20,8 @@ avatar-page-rotate-button = Rotate
 avatar-page-camera-error = Could not initialize camera
 avatar-page-new-avatar =
   .alt = new profile picture
-avatar-page-file-upload-error = There was a problem uploading your profile picture
-avatar-page-delete-error = There was a problem deleting your avatar
+avatar-page-file-upload-error-2 = There was a problem uploading your profile picture.
+avatar-page-delete-error-2 = There was a problem deleting your avatar.
```
dannycoates, line 24: Since we're making a new string we should also use consistent vocabulary. How about "profile picture"?

6a68: Nice! :+1:

dannycoates (approved):

PR #7613, opened 2021-02-23
feat(fxa-payments-server): add processing view to checkout

### LINKED ISSUE #7450: Checkout: Add a processing screen before the successful/failed authorization page

meandavejustice:
## Because

- We want to give the user feedback about the status of their paypal payments

## This pull request

- Adds a processing view to the paypal payment flow

- fixes #7450

## Checklist

## Screenshots (Optional)
processing-screen

image

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentProcessing/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentProcessing/index.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import { Localized } from '@fluent/react';
+
+import { LoadingSpinner } from '../LoadingSpinner';
+
+import './index.scss';
+
+export type PaymentProcessingProps = {
+  className?: string;
+};
+
+export const PaymentProcessing = ({
+  className = '',
+}: PaymentProcessingProps) => {
+  return (
+    <section
+      className={`container card payment-processing ${className}`}
+      data-testid="payment-processing"
+    >
+      <div className="wrapper">
+        <LoadingSpinner />
+        <Localized id="payment-processing-message">
+          <p>Please wait while we process your payment...</p>
+        </Localized>
+      </div>
+    </section>
+  );
+};
+
+export default PaymentProcessing;
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.stories.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.stories.tsx
@@ -7,7 +7,6 @@ import {
   PaymentIntent,
 } from '@stripe/stripe-js';
 import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
 import { linkTo } from '@storybook/addon-links';
 import MockApp, {
   defaultAppContextValue,
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
@@ -6,6 +6,7 @@ import React, {
   Suspense,
 } from 'react';
 import { Stripe, StripeCardElement, StripeError } from '@stripe/stripe-js';
+import classNames from 'classnames';
 import { Plan, Profile, Customer } from '../../../store/types';
 import { State as ValidatorState } from '../../../lib/validator';
 
@@ -19,6 +20,7 @@ import ErrorMessage from '../../../components/ErrorMessage';
 import AcceptedCards from '../../Product/AcceptedCards';
 import PaymentLegalBlurb from '../../../components/PaymentLegalBlurb';
 import { TermsAndPrivacy } from '../../../components/TermsAndPrivacy';
+import { PaymentProcessing } from '../../../components/PaymentProcessing';
 
 import * as Amplitude from '../../../lib/amplitude';
 import { Localized } from '@fluent/react';
@@ -73,6 +75,7 @@ export const SubscriptionCreate = ({
   paypalButtonBase,
 }: SubscriptionCreateProps) => {
   const [submitNonce, refreshSubmitNonce] = useNonce();
+  const [transactionInProgress, setTransactionInProgress] = useState(false);
 
   const onFormMounted = useCallback(
     () => Amplitude.createSubscriptionMounted(selectedPlan),
@@ -110,9 +113,11 @@ export const SubscriptionCreate = ({
     script.src = `${config.paypal.scriptUrl}/sdk/js?client-id=${config.paypal.clientId}&vault=true&commit=false&intent=capture&disable-funding=credit,card`;
     // Pass the csp nonce to paypal
     script.setAttribute('data-csp-nonce', cspNonce);
+    /* istanbul ignore next */
     script.onload = () => {
       setPaypalScriptLoaded(true);
     };
+    /* istanbul ignore next */
     script.onerror = () => {
       throw new Error('Paypal SDK could not be loaded.');
     };
@@ -179,7 +184,17 @@ export const SubscriptionCreate = ({
     <>
       <Header {...{ profile }} />
       <div className="main-content">
-        <div className="product-payment" data-testid="subscription-create">
+        <PaymentProcessing
+          className={classNames({
+            hidden: !transactionInProgress,
+          })}
+        />
+        <div
+          className={classNames('product-payment', {
+            hidden: transactionInProgress,
+          })}
+          data-testid="subscription-create"
+        >
           <div
             className="subscription-create-heading"
             data-testid="subscription-create-heading"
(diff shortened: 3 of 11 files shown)


meandavejustice:
@biancadanforth pinging you to make you aware of this patch so we don't run into conflicts, let me know if I need to add that sub confirmation header into this pr

biancadanforth:
> @biancadanforth pinging you to make you aware of this patch so we don't run into conflicts, let me know if I need to add that sub confirmation header into this pr

Thanks I will add the header to this screen as part of my current patch that makes it into its own component.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -53,6 +53,7 @@ export type SubscriptionCreateProps = {
   customer: Customer | null;
   selectedPlan: Plan;
   refreshSubscriptions: () => void;
+  setTransactionInProgress: (value: boolean) => void;
```
biancadanforth, line 56: Wouldn't we want to use a hook for this; i.e. `useState` similar to `setPaypalScriptLoaded`? Setting this to `true` in this way on clicking the `PayPalButton` wouldn't trigger a re-render of `SubscriptionCreate`.

biancadanforth (changes requested):
Hey Dave; I haven't looked through all of this, but do you see the processing screen after you successfully authorize with PayPal?

I cherry-picked your commit off the latest `main`, and I don't see it. I have a hypothesis that I comment on in-line here.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
biancadanforth: Edit: My bad you do use `useState`... hmmm... I don't see the screen though when I try it locally.

biancadanforth: Also why is `setTransactionInProgress` a prop here on `SubscriptionCreate`? I don't see anything other than a noop passed in for the tests.

biancadanforth: Answering my own questions here...

> Also why is setTransactionInProgress a prop here on SubscriptionCreate? I don't see anything other than a noop passed in for the tests.

Because you set it to `true` when the `PayPalButton` is clicked which is inside `SubscriptionCreate`.

Also I have a theory on why it may not be working, I think we'd want to pass the state to `SubscriptionCreate` and conditionally show `PaymentProcessing` there, since `Product` will only render `SubscriptionSuccess` after the processing is complete, so we'll never see it if it's conditionally shown there.

chenba:
I just tried subscribing with PayPal on the branch for this PR and didn't notice any loading screen like the one in the screenshots above.  What did I do wrong?

packages/fxa-payments-server/src/components/PaymentProcessing/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const PaymentProcessing = () => {
+  return (
+    <section
+      className={`container card payment-processing`}
+      data-testid="payment-processing"
+    >
+      <div className="wrapper">
+        <LoadingSpinner />
+        <Localized id="payment-processing-message">
+          <p></p>
```
chenba, line 17: Prefer to have some fallback content in case the .ftl request 404 or whatever.

meandavejustice:
> I just tried subscribing with PayPal on the branch for this PR and didn't notice any loading screen like the one in the screenshots above. What did I do wrong?

I'm fixing a conflict in SubscriptionCreate, I'll ping you on slack when I have it fixed

chenba:
Just tried again on the latest commit and I was stuck on the payment process screen.  The window that's opened upon clicking the PayPal button was immediately closed.  I was not able complete the subscription process.

I don't think it's an issue with PayPal since the process works on `main`.

chenba (changes requested):
I was stuck on the processing payment screen and not able to complete a subscription.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.stories.tsx
```diff
@@ -27,9 +26,22 @@ const wait = (delay: number) =>
 
 function init() {
   storiesOf('routes/ProductV2/SubscriptionCreate', module)
-    .add('default', () => <Subject />)
+    .add('default', () => (
+      <Subject
+        transactionInProgress={false}
+        setTransactionInProgress={() => {}}
```
chenba, line 32: Aren't these the defaults?  Why pass them so many times?

packages/fxa-payments-server/src/routes/Product/index.test.tsx
```diff
@@ -159,8 +159,9 @@ describe('routes/Product', () => {
         .get('/v1/oauth/subscriptions/customer')
         .reply(200, MOCK_CUSTOMER, { 'Access-Control-Allow-Origin': '*' }),
     ];
-    const { findByTestId } = render(<Subject />);
+    const { findByTestId, queryByTestId } = render(<Subject />);
     await findByTestId('error-loading-profile');
+    expect(queryByTestId('error-loading-profile')).toBeInTheDocument();
```
chenba, line 164: Do we need line 163 then?

meandavejustice: The linter was complaining about the lack of `expect` on these tests. In this case we do need line `163`, we need to wait until it is in the DOM before we can test for it.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.stories.tsx
meandavejustice: Removed excess arguements

packages/fxa-payments-server/src/routes/Product/index.tsx
```diff
@@ -59,6 +59,8 @@ export const Product = ({
     AppContext
   );
 
+  const [transactionInProgress, setTransactionInProgress] = useState(false);
```
chenba, line 62: Can this be in `SubscriptionCreate`?  I don't see why it should be in here.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
chenba: Or assign the element found on line 163 into a variable and use it.  There's no need to find/query multiple times.

packages/fxa-payments-server/src/routes/Product/index.tsx
meandavejustice: I had it here so that it can be unset by `customerIsSubscribedToProduct` [here](https://github.com/mozilla/fxa/pull/7613/files/cd99a322f208e000d4dc68211dd9748be93530cb#diff-9cb71f54629ac6b7b6ccf9d0d5e3d30ed33d99367527bb55519116778955e5d2R82). I guess we don't really need to unset, since it will be replaced by `SubscriptionSuccess`

chenba (changes requested):
The `transactionInProgress` state should be in `SubscriptionCreate`; it currently does not affect anything in `Product`.

packages/fxa-payments-server/src/routes/Product/index.tsx
chenba: `customerIsSubscribedToProduct` is just a lib function.  It doesn't use anything from the scope of `Product`.  We can move it somewhere so that both `Product` and `SubscriptionCreate` can use it, if that's the issue.

packages/fxa-payments-server/src/routes/Product/index.test.tsx
chenba: (Sorry I didn't see your response when I wrote my second comment...I thought GH UI is better than that)

So, a) await the rendering, or b) await the expect, or c) assign the element found by `findByTestId` and use it in the assertion.

chenba (approved):
WFM 👍

PR #7628, opened 2021-02-24
7020 scroll position

### LINKED ISSUE #7020: [FxA-Settings] “Cancel” and “Back” buttons take the user to the top of the Settings page
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Android 9
iOS 13.6

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Tap on the "Add" button from the Secondary email section.
3. Tap on “Cancel” or “Back“ buttons.

**Expected results**
The user is taken to the last visited section from the Settings Page.

**Actual Results**
The top of the Settings page is always displayed no matter from what page the user returns.

**Additional notes**
- This issue also reproduces for Change Password, Recovery key, 2FA pages.
- This issue does not reproduce on Windows.

issue56

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2749)
┆Issue Number: FXA-2749

6a68:
Ready for review. To ensure we scroll to the right spot in the page, we could do something fancy, complex, and brittle, like storing the scroll distance when navigating to a page, then restoring the scroll location if the user goes back. However, the intent of the bug is to not get thrown to the top of the screen when backing out of a sub-page, and the simplest approach is to add `id`s to each of the rows, then just navigate to that hash instead of using `history.back()`. 

This ran into a little bit of complexity, because reach router doesn't handle hashes, but hacking around that was easy enough and well-contained.

Fixes #7020.

### DIFF
--- a/packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.tsx
+++ b/packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.tsx
@@ -28,7 +28,8 @@ export const Page2faReplaceRecoveryCodes = (_: RouteComponentProps) => {
   const alertBar = useAlertBar();
   const navigate = useNavigate();
   const session = useSession();
-  const goBack = () => window.history.back();
+  const goBack = () =>
+    navigate(HomePath + '#two-step-authentication', { replace: true });
   const goHome = () => {
     alertTextExternal(
       l10n.getString(
@@ -37,7 +38,7 @@ export const Page2faReplaceRecoveryCodes = (_: RouteComponentProps) => {
         'Account recovery codes updated.'
       )
     );
-    navigate(HomePath, { replace: true });
+    navigate(HomePath + '#two-step-authentication', { replace: true });
   };
   const { l10n } = useLocalization();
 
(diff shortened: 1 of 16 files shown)


packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.tsx
```diff
@@ -37,7 +38,7 @@ export const Page2faReplaceRecoveryCodes = (_: RouteComponentProps) => {
         'Account recovery codes updated.'
       )
     );
-    navigate(HomePath, { replace: true });
```
dannycoates, line 40: nit? maybe just `goBack()` here? (doesn't matter that much)

dannycoates (approved):
This looks good.

This is only a suggestion but what if we renamed `goBack` to `goHome` and `goHome` to something like `alertSuccesAndGoHome`:
```
goHome = () => navigate(...)
alertSuccessAndGoHome = () => {
  alertTextExternal(...)
  goHome()
}
```

6a68:
I was thinking about that too. Let's catch it in the cleanup sprint

PR #7629, opened 2021-02-24
feat(auth-server): add paypal invoice processor

### LINKED ISSUE #7452: Create cron job to run PayPal job to process invoice
Create a script like fxa-auth-server/script/populate-stripe-customer-cache.js that is used by ops to run cron job to get/make PayPal transaction and advance Stripe Invoice accordingly.

Technical Implementation:

```java

var invoiceData = function(invoices) {
  Promise.all([updateStripeInvoiceAttemptCount(invoices),
               getPayPalTransaction(invoices)])
  .then((data) => {
    updateStripeInvoice(data)
  })
  .catch((error) => {
    // log error
  })
}

// get invoices with `auto_advance` set to `send_invoice` and `status` in `draft` or `open` using Stripe helper
// Stripe Invoice has a limit of 100 invoices returned at a time see https://stripe.com/docs/api/invoices/list#list_invoices-ending_before
invoices = StripeHelper.getInvoices();

// while Stripe have invoices
  invoiceData(invoices)
  invoices = StripeHelper.getInvoices(startingAfter=invoices.objectID)
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2853)
┆Issue Number: FXA-2853

bbangert:
Because:

* We want to ensure invoices paid via PayPal are processed when their
  initial attempt fails or payments are pending then completed later.

This commit:

* Adds a paypal-processor script intended to be run on an hourly basis
  that queries open invoices paid via paypal to process them as needed.

Closes #7452

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-client.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-client.ts
@@ -59,6 +59,12 @@ type NVPResponse = {
   CORRELATIONID: string;
   TIMESTAMP: string;
   VERSION: string;
+  L?: {
+    ERRORCODE: string;
+    SHORTMESSAGE: string;
+    LONGMESSAGE: string;
+    SEVERITYCODE: string;
+  }[];
 };
 
 type SetExpressCheckoutData = {
@@ -132,7 +138,7 @@ export type NVPDoReferenceTransactionResponse = NVPResponse &
 
 export type NVPBAUpdateTransactionResponse = NVPResponse & BAUpdateData;
 
-export type NVPTransactionSearchResponse = NVPResponse & TransactionSearchData;
+export type NVPTransactionSearchResponse = TransactionSearchData & NVPResponse;
 
 export type SetExpressCheckoutOptions = {
   currencyCode: string;
@@ -214,6 +220,7 @@ type ResponseEventType = {
 export class PayPalClientError extends Error {
   public raw: string;
   public data: NVPResponse;
+  public errorCode: number | undefined;
 
   constructor(raw: string, data: NVPResponse, ...params: any) {
     super(...params);
@@ -224,8 +231,10 @@ export class PayPalClientError extends Error {
     }
     this.raw = raw;
     this.data = data;
+    this.errorCode = data.L?.length ? parseInt(data.L[0].ERRORCODE) : undefined;
   }
 }
+
 export class PayPalClient {
   private url: string;
   private ipnUrl: string;
--- a/packages/fxa-auth-server/lib/payments/paypal-processor.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-processor.ts
@@ -1,0 +1,334 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { Logger } from 'mozlog';
+import Stripe from 'stripe';
+import { Container } from 'typedi';
+
+import { ConfigType } from '../../config';
+import error from '../../lib/error';
+import { reportSentryError } from '../sentry';
+import { PayPalHelper, TransactionSearchResult } from './paypal';
+import { PayPalClientError } from './paypal-client';
+import {
+  PAYPAL_BILLING_AGREEMENT_INVALID,
+  PAYPAL_SOURCE_ERRORS,
+} from './paypal-error-codes';
+import { StripeHelper } from './stripe';
+
+/**
+ * Generest a timestamp in seconds that is `hours` before the current
+ * time.
+ *
+ * @param hours
+ */
+function hoursBeforeInSeconds(hours: number): number {
+  const date = new Date();
+  date.setHours(date.getHours() - hours);
+  return date.getTime() / 1000;
+}
+
+/**
+ * Returns whether two dates are the same month/day/year.
+ *
+ * @param d1
+ * @param d2
+ */
+function sameDay(d1: Date, d2: Date) {
+  return (
+    d1.getUTCFullYear() === d2.getUTCFullYear() &&
+    d1.getUTCMonth() === d2.getUTCMonth() &&
+    d1.getUTCDate() === d2.getUTCDate()
+  );
+}
+
+export class PaypalProcessor {
+  private stripeHelper: StripeHelper;
+  private paypalHelper: PayPalHelper;
+
+  constructor(
+    private log: Logger,
+    config: ConfigType,
+    private graceDays: number,
+    private maxRetryAttempts: number,
+    private invoiceAge = 6
+  ) {
+    this.stripeHelper = Container.get(StripeHelper);
+    this.paypalHelper = Container.get(PayPalHelper);
+  }
+
+  /**
+   * Determine if the given invoice is within its grace period for
+   * additional payment retries or waiting on a pending transaction.
+   *
+   * Calculated by adding the `graceDays` in seconds to the invoice
+   * creation time. A `graceDays` value of 1 allows an invoice to be
+   * open for 24 hours hours (1 day) before not in grace period, etc.
+   *
+   * @param invoice
+   */
+  private inGracePeriod(invoice: Stripe.Invoice): boolean {
+    const graceDaysInSeconds = this.graceDays * 24 * 60 * 60;
+    return Date.now() / 1000 < invoice.created + graceDaysInSeconds;
+  }
+
+  private cancelInvoiceSubscription(invoice: Stripe.Invoice) {
+    // TODO: should send an email about subscription cancelled as overdue
+    return Promise.all([
+      this.stripeHelper.markUncollectible(invoice),
+      this.stripeHelper.cancelSubscription(invoice.subscription as string),
+    ]);
+  }
+
+  /**
+   * Ensures that an invoice matches the payment attempts returned from a PayPal
+   * Transaction Search.
+   *
+   * In general this should be the case, but in the event that an IPN notification
+   * is missed a mismatch could occur.
+   *
+   * @param invoice
+   * @param transactions
+   */
+  private async ensureAccurateAttemptCount(
+    invoice: Stripe.Invoice,
+    transactions: TransactionSearchResult[]
+  ) {
+    const paymentAttempts = this.stripeHelper.getPaymentAttempts(invoice);
+    if (paymentAttempts !== transactions.length) {
+      await this.stripeHelper.updatePaymentAttempts(
+        invoice,
+        transactions.length
+      );
+    }
+  }
+
+  /**
+   * Handles the condition where a transaction for an invoice has been paid.
+   *
+   * This occurrs when a previously pending transaction becomes paid, and the
+   * invoice should now be marked as paid.
+   *
+   * Returns whether the invoice has been paid.
+   *
+   * @param invoice
+   * @param transactions
+   */
+  private async handlePaidTransaction(
+    invoice: Stripe.Invoice,
+    transactions: TransactionSearchResult[]
+  ): Promise<boolean> {
+    const customer = invoice.customer;
+    const successTransactions = transactions.filter((t) =>
+      ['Completed', 'Processed'].includes(t.status)
+    );
+    if (successTransactions.length) {
+      if (successTransactions.length > 1) {
+        this.log.error('multipleCompletedTransactions', {
+          customer,
+          invoiceId: invoice.id,
+          transactionCount: successTransactions.length,
+          excessTransactions: successTransactions
+            .slice(1)
+            .map((t) => t.transactionId),
+        });
+        reportSentryError(
+          new Error('Multiple completed payments for invoice: ' + invoice.id)
+        );
+      }
+      const transaction = successTransactions[0];
+      await this.stripeHelper.updateInvoiceWithPaypalTransactionId(
+        invoice,
+        transaction.transactionId
+      );
+      await this.stripeHelper.payInvoiceOutOfBand(invoice);
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Handles the condition where a transaction for an invoice is pending.
+   *
+   * This occurs when a previous payment attempt has not completed. This
+   * method checks to see if there's a pending transaction for the invoice.
+   * Returns `true` if this invoice has valid pending transactions to wait
+   * for, `false` otherwise.
+   *
+   * @param invoice
+   * @param transactions
+   */
+  private async handlePendingTransaction(
+    invoice: Stripe.Invoice,
+    transactions: TransactionSearchResult[]
+  ): Promise<boolean> {
+    const customer = invoice.customer;
+    const inGracePeriod = this.inGracePeriod(invoice);
+    const outstandingTransactions = transactions.filter((t) =>
+      ['Pending', 'In-Progress', 'Under Review'].includes(t.status)
+    );
+    if (outstandingTransactions.length) {
+      if (outstandingTransactions.length > 1) {
+        this.log.error('multiplePendingTransactions', {
+          customer,
+          invoiceId: invoice.id,
+        });
+        reportSentryError(
+          new Error('Multiple pending payments for invoice: ' + invoice.id)
+        );
+      }
+      if (inGracePeriod) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Make a payment attempt on an invoice.
+   *
+   * @param invoice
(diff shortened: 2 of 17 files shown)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -132,7 +138,7 @@ export type NVPDoReferenceTransactionResponse = NVPResponse &
 
 export type NVPBAUpdateTransactionResponse = NVPResponse & BAUpdateData;
 
-export type NVPTransactionSearchResponse = NVPResponse & TransactionSearchData;
+export type NVPTransactionSearchResponse = TransactionSearchData & NVPResponse;
```
bbangert, line 141: These both now have a `L` attribute, so to ensure the caller gets the correct TS one out first we have to switch the order.

packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -224,8 +231,10 @@ export class PayPalClientError extends Error {
     }
     this.raw = raw;
     this.data = data;
+    this.errorCode = data.L?.length ? parseInt(data.L[0].ERRORCODE) : undefined;
```
bbangert, line 234: Currently we only pull the first error, we may need to change this later to error sets depending on how common it is that PayPal uses those.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,335 @@
...
+   * Ensures that an invoice matches the payment attempts returned from a PayPal
+   * Transaction Search.
+   *
+   * In general this should be the case, but in the event that an IPN notification
+   * is missed a mismatch could occur.
+   *
+   * @param invoice
+   * @param transactions
+   */
+  private async verifyAttemptcount(
```
chenba, line 93: The name is misleading since it actually changes the count.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,335 @@
...
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Make a payment attempt on an invoice.
+   *
+   * @param invoice
+   */
```
chenba, line 193: I think we can remove one of these comment blocks. :)

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,335 @@
...
+          new Error('Multiple completed payments for invoice: ' + invoice.id)
+        );
+      }
+      const transaction = successTransactions[0];
+      await Promise.all([
+        this.stripeHelper.updateInvoiceWithPaypalTransactionId(
+          invoice,
+          transaction.transactionId
+        ),
+        this.stripeHelper.payInvoiceOutOfBand(invoice),
```
chenba, line 145: How important is the order of these two calls?  In the (likely rare) scenario when one resolves but the other rejects, would that put an invoice in a state that we don't handle?

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,335 @@
...
+    } catch (err) {
+      if (err instanceof PayPalClientError) {
+        if (err.errorCode === PAYPAL_BILLING_AGREEMENT_INVALID) {
+          const uid = customer.metadata.userid;
+          await Promise.all([
+            this.stripeHelper.removeCustomerPaypalAgreement(customer),
+            updatePayPalBA(
+              uid,
+              this.stripeHelper.getCustomerPaypalAgreement(customer) as string,
+              'Cancelled',
```
chenba, line 220: So I learned from a l10n PR that apparently 'Cancelled' is the Canadian spelling of 'Canceled'?!

Anyways maybe we can do something like 
```
export const PayPalBillingAgreementStatuses = {
  PENDING: 'Pending',
  ACTIVE: 'Active',
  SUSPENDed: 'Suspended',
  CANCELLed: 'Cancelled',
  EXPIRED: 'Expired',
} as const;
type PayPalBillingAgreementStatuses = typeof PayPalBillingAgreementStatuses;
export type PayPalBillingAgreementStatusType = PayPalBillingAgreementStatuses[keyof PayPalBillingAgreementStatuses];
```

then we can use `PayPalBillingAgreementStatuses.CANCELLED` instead of a literal here.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,333 @@
...
+   *      be marked as uncollectible.
+   *   3. Search all transactions run for this invoice and:
+   *     a. Ensure the transaction count matches the invoice retry attempts.
+   *     b. If a transaction completed, mark invoice as paid, return.
+   *     c. If a transaction is pending/in-progress:
+   *       - If we're in grace period, skip invoice and wait for completion, return.
+   *   4. If we're past grace period, cancelInvoiceSubscription, return.
+   *   5. If we have no billing agreement, email user to inform them.
+   *   6. If retries for today remain:
+   *       a. Attempt paymanet on invoice.
```
chenba, line 259: Typo here.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,333 @@
...
+    // 6
+    if (this.attemptsToday(transactions) < this.maxRetryAttempts) {
+      await this.makePaymentAttempt(invoice);
+    }
+    return;
+  }
+
+  public async processInvoices() {
+    // Generate a time 6 hour prior.
+    const hourAgo = hoursBeforeInSeconds(6);
```
chenba, line 325: nit: `hoursAgo`?

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -31,6 +31,9 @@ export class PayPalNotificationHandler extends PayPalHandler {
       });
     }
 
+    // TODO: If the invoice is void/uncollectible and this is Completed/Processed, issue
+    // a refunded.
```
say-yawn, line 35: Ticket filed and assigned [here](https://jira.mozilla.com/browse/FXA-3086)

packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -108,13 +166,37 @@ export class PayPalHelper {
...
+   * API calls.
+   *
+   * @param invoiceId
+   * @param paymentAttempt
+   */
+  public generateIdempotencyKey(
+    invoiceId: string,
+    paymentAttempt: number
+  ): string {
+    return invoiceId + '-' + paymentAttempt;
```
chenba, line 180: nit^2: I believe the cool kids are writing stuff like `${invoiceId}-${paymentAttempt}`.

packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -108,13 +166,37 @@ export class PayPalHelper {
...
+   *
+   * @param idempotencyKey
+   */
+  public parseIdempotencyKey(
+    idempotencyKey: string
+  ): { invoiceId: string; paymentAttempt: number } {
+    const parsedValue = idempotencyKey.split('-');
+    return {
+      invoiceId: parsedValue[0],
+      paymentAttempt: parseInt(parsedValue[1] + 1),
```
chenba, line 198: Addition _after_ `parseInt` here?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -431,6 +442,25 @@ export class StripeHelper {
...
+   * Remove the PayPal Billing Agreement ID from a customer.
+   *
+   * @param customer
+   * @param agreementId
+   */
+  async removeCustomerPaypalAgreement(
+    uid: string,
+    customer: Stripe.Customer,
+    billingAgreementId: string
+  ): Promise<any[]> {
```
chenba, line 455: I prefer to either let TS infer the return type or use actual types (`Promise<Promise<Stripe.Customer | number>[]>`).

The type is a Promise of an array of Promises since the function is `async`.  But should the array be in a `Promise.all`?  An `any[]` resolves immediately.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -671,7 +736,7 @@ export class StripeHelper {
   /**
    * Fetch a subscription for a customer from Stripe.
    *
-   * Uses Redis caching if configured.
+   * Uses Redis cac hing if configured.
```
chenba, line 739: Accidentally spacebar here.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
bbangert: If adding the transaction id works, but it isn't marked paid, then the next run through when it checks for completed transactions it will 'fix it. If its vice versa, then we end up with an invoice properly marked as closed but without a transaction id. Nice catch, I'm going to make these sequential so we record the transaction id first.

packages/fxa-auth-server/lib/payments/paypal.ts
bbangert: I shall strive to be cooler.

packages/fxa-auth-server/lib/payments/paypal.ts
bbangert: Oh my, that was unintentional!

packages/fxa-auth-server/lib/payments/paypal-processor.ts
bbangert: I wonder if this means PayPal is originally Canadian. I should note that the `updatePayPalBA` function does have a strong type for the string, TS will fail to compile if you typo that string and VS Code gives proper completions there.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -56,15 +59,7 @@ export class PayPalNotificationHandler extends PayPalHandler {
             message: 'No idempotency key on PayPal transaction',
           });
         }
-        const { paymentAttempt } = this.paypalHelper.parseIdempotencyKey(
-          message.custom
```
chenba, line 60: What is `message.custom`?

packages/fxa-auth-server/package.json
```diff
@@ -25,7 +25,8 @@
     "format": "prettier --write --config ../../_dev/.prettierrc '**'",
     "gen-keys": "ts-node ./scripts/gen_keys.js; ts-node ./scripts/oauth_gen_keys.js; ts-node ./scripts/gen_vapid_keys.js",
     "write-emails": "node -r ts-node/register ./scripts/write-emails-to-disk.js",
-    "clean-up-old-ci-stripe-customers": "ts-node ./scripts/clean-up-old-ci-stripe-customers.js"
+    "clean-up-old-ci-stripe-customers": "ts-node ./scripts/clean-up-old-ci-stripe-customers.js",
+    "paypal-processor": "NODE_ENV=dev CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts"
```
chenba, line 29: Hard coding `NODE_ENV`? 🤨

bbangert: Oh, a convenience for myself, heh. Should remove that for sure.

packages/fxa-auth-server/scripts/paypal-processor.ts
```diff
@@ -0,0 +1,70 @@
...
+import { setupAuthDatabase } from 'fxa-shared/db';
+import { StatsD } from 'hot-shots';
+import Container from 'typedi';
+
+import { PaypalProcessor } from '../lib/payments/paypal-processor';
+import { StripeHelper } from '../lib/payments/stripe';
+import { configureSentry } from '../lib/sentry';
+
+const pckg = require('../package.json');
+const config = require('../config').getProperties();
```
chenba, line 15: nit: `import` or `require` for consistency?

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
bbangert: PayPal provides a `CUSTOM` field that can be used by the developer as desired and is retained (but not shown to the user).

packages/fxa-auth-server/test/local/payments/paypal-processor.js
```diff
@@ -0,0 +1,351 @@
...
+      const invoice = deepCopy(unpaidInvoice);
+      mockStripeHelper.getPaymentAttempts = sandbox.fake.returns(2);
+      mockStripeHelper.updatePaymentAttempts = sandbox.fake.resolves({});
+      await processor.ensureAccurateAttemptCount(invoice, [{}]);
+      sinon.assert.calledOnceWithExactly(
+        mockStripeHelper.updatePaymentAttempts,
+        invoice,
+        1
+      );
+      sandbox.reset();
```
chenba, line 122: I think a `mockStripeHelper.updatePaymentAttempts.resetHistory()` would also work here.

packages/fxa-auth-server/test/local/payments/paypal-processor.js
```diff
@@ -0,0 +1,351 @@
...
+        { customer: unpaidInvoice.customer, invoiceId: unpaidInvoice.id }
+      );
+    });
+
+    it('returns false if no pending exist', async () => {
+      processor.inGracePeriod = sandbox.fake.returns(true);
+      const result = await processor.handlePendingTransaction(unpaidInvoice, [
+        { status: 'Completed' },
+      ]);
+      assert.isFalse(result);
```
chenba, line 224: I think the `false` returned here is from the 'Completed' status, not the grace period check.

bbangert: What do you mean?

packages/fxa-auth-server/test/local/payments/paypal-processor.js
bbangert: It would, yes. As long as that one is reset it will pass.

packages/fxa-auth-server/test/local/payments/paypal-processor.js
chenba: I think I misread the title on this one last night.

packages/fxa-auth-server/test/local/payments/paypal-processor.js
```diff
@@ -0,0 +1,606 @@
...
+        unpaidInvoice
+      );
+    });
+  });
+
+  describe('makePaymentAttempt', () => {
+    it('processes zero invoice if its 0', async () => {
+      const invoice = deepCopy(unpaidInvoice);
+      invoice.amount_due = 0;
+      mockPaypalHelper.processZeroInvoice = sandbox.fake.resolves({});
```
chenba, line 250: I think an assert against this fake is missing.

packages/fxa-auth-server/test/local/payments/paypal-processor.js
```diff
@@ -0,0 +1,616 @@
...
+
+      for (const spy of [
+        processor.ensureAccurateAttemptCount,
+        processor.handlePaidTransaction,
+        processor.handlePendingTransaction,
+      ]) {
+        sinon.assert.calledOnceWithExactly(spy, invoice, []);
+      }
+      sinon.assert.calledOnceWithExactly(processor.inGracePeriod, invoice);
+      sinon.assert.notCalled(mockStripeHelper.getCustomerPaypalAgreement);
```
chenba, line 537: Appears to be missing an assert on `processor.cancelInvoiceSubscription` here.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -0,0 +1,331 @@
...
+    // 6
+    if (this.attemptsToday(transactions) < this.maxRetryAttempts) {
+      await this.makePaymentAttempt(invoice);
+    }
+    return;
+  }
+
+  public async processInvoices() {
+    // Generate a time 6 hour prior.
+    const sixHoursAgo = hoursBeforeInSeconds(6);
```
chenba, line 318: Any reason we don't want the #hour value passed in?

bbangert: Not really, this could be a default passed in.

chenba (approved):
LGTM 💰

Issue #7563, opened 2021-02-17
Checkout: PayPal Client to handle amoutnInCents


data-sync-user:
Checkout: Update Paypal Client to take in amoutnInCents and to process into paypal compatible amount string (should handle multiple currencies). See discussion here: [https://mozilla.slack.com/archives/CG0AJ6E77/p1613501264283800](https://mozilla.slack.com/archives/CG0AJ6E77/p1613501264283800)

After discussing with [~bbangert@mozilla.com] , we will restrict paypal to reject transations in anything other than USD and EUR by listing those as allowed currencies and when making the conversion to string, for paypal, checking.

Some docs:

* [https://developer.paypal.com/docs/nvp-soap-api/do-reference-transaction-nvp/](https://developer.paypal.com/docs/nvp-soap-api/do-reference-transaction-nvp/)
* [https://developer.paypal.com/docs/nvp-soap-api/currency-codes/](https://developer.paypal.com/docs/nvp-soap-api/currency-codes/)

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3059)
┆Issue Number: FXA-3059

PR #7635, opened 2021-02-25
fix(auth-server): Correctly format paypal AMT string

birdsarah:
## Because

- We were just converting the stripe AmountInCents to a string causing paypal customers to be charged 100x more than intended.
- Paypal has a lot of finicky rules around currency

## This pull request

- Implements a function to convert amountInCents to a string as per paypal specs
- Restricts currency support to a list of paypal supported currencies

Closes: #7563

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/payments/currencies.ts
+++ b/packages/fxa-auth-server/lib/payments/currencies.ts
@@ -5,14 +5,23 @@ import { ConfigType } from '../../config';
 
 export class CurrencyHelper {
   currencyToCountryMap: Map<string, string[]>;
+  payPalEnabled: boolean;
 
   constructor(config: ConfigType) {
-    // Vaildate currencyMap
+    // Is payPalEnabled?
+    this.payPalEnabled = config.subscriptions?.paypalNvpSigCredentials?.enabled;
+    // Validate currencyMap
     const currencyMap = new Map(Object.entries(config.currenciesToCountries));
     let allListedCountries: string[] = [];
     for (let [currency, countries] of currencyMap) {
       // Is currency acceptable
       if (!CurrencyHelper.validCurrencyCodes.includes(currency)) {
+        throw new Error(`Currency code ${currency} is invalid.`);
+      }
+      if (
+        this.payPalEnabled &&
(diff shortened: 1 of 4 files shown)


birdsarah:
@bbangert. I implemented this as you suggested. I think this is the right way for this paypal issue. 

But wanted to note that the codebase already has two copies of a function "getLocalizedCurrencyString" which uses `Intl.NumberFormat` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat).

packages/fxa-auth-server/lib/payments/currencies.ts
```diff
@@ -49,6 +58,39 @@ export class CurrencyHelper {
...
+   * The documentation for currency codes and the various restrictions is here: https://developer.paypal.com/docs/nvp-soap-api/currency-codes/
+   *
+   * Restrictions include
+   * - whether decimal point is or isn't allowed.
+   * - whether there's a transaction limit for the country
+   *
+   * As a result, we hard code the supported PayPal currencies and additional currencies
+   * can be added here as necessary, once the PayPal docs have been reviewed to ensure we've
+   * handled any restrictions.
+   *
```
bbangert, line 90: 💯 I like the call-out to why this is the way it is in the doc-string.

packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -154,9 +169,19 @@ export class PayPalHelper {
...
   ): Promise<ChargeResponse> {
-    const response = await this.client.doReferenceTransaction(options);
+    const doReferenceTransactionOptions = {
+      amount: this.currencyHelper.getPayPalAmountStringFromAmountInCents(
+        options.amountInCents
+      ),
+      billingAgreementId: options.billingAgreementId,
+      invoiceNumber: options.invoiceNumber,
+      idempotencyKey: options.idempotencyKey,
+    } as DoReferenceTransactionOptions;
```
bbangert, line 181: This needs a type-cast?

bbangert (approved):
Looks good, I'm curious if the type-cast is really needed though.

Issue #7264, opened 2021-01-13
Use PayPal billing information in Subscription Management emails


data-sync-user:
This ticket covers e-mails specific to Subscription Management. 

–

Invoice emails currently have a strings like "Charged to Visa" and "card ending XXXX". This is the same information shown under "Billing Information" on the Subscription Management page. We should make sure the PayPal billing information is displayed instead for PayPal customers.

Per Barry:

* There are 3 different types of e-mails right now that include the payment information if it’s there (it’d look for the `last4` property on the Customer object), so we’ll have to update those ones to check if it’s Stripe or PayPal… Similar to FXA-2801 logic.
  * First invoice e-mail
  * Subsequent invoice e-mails
  * If they cancel and then resubscribe e-mail
* Mini-spike:
  * These e-mails are driven by Stripe webhook events. Since we're manually invoicing with Stripe, we need to make sure whenever we update the invoice, we get the right webhook events. If we mark an invoice as paid, do we get the event that would trigger the payment succeeded e-mail? If not, we need to figure that out.

[~bchen@mozilla.com]: Does SubPlat own these emails? If so, how do we change them? If not, who do we need to contact?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2898)
┆Issue Number: FXA-2898

data-sync-user:
➤ Lesley Norton commented:

It was decided not to send receipt emails when a user's payment method is PayPal since PayPal already issues their own version of these emails.

data-sync-user:
➤ Bianca Danforth commented:

Per legal we do actually want to send these Subscription Management invoice e-mails; see FXA-2989.

Issue #7101, opened 2020-12-08
Use PayPal billing information in Checkout invoice email


data-sync-user:
This ticket covers e-mails specific to Checkout. 

--

Invoice emails currently have a strings like "Charged to Visa" and "card ending XXXX". This is the same information shown under "Billing Information" on the Subscription Management page. We should make sure the PayPal billing information is displayed instead for PayPal customers.

Per Barry:

* There are 3 different types of e-mails right now that include the payment information if it’s there (it’d look for the `last4` property on the Customer object), so we’ll have to update those ones to check if it’s Stripe or PayPal… Similar to FXA-2801 logic.
  * First invoice e-mail
  * Subsequent invoice e-mails
  * If they cancel and then resubscribe e-mail
* Mini-spike:
  * These e-mails are driven by Stripe webhook events. Since we're manually invoicing with Stripe, we need to make sure whenever we update the invoice, we get the right webhook events. If we mark an invoice as paid, do we get the event that would trigger the payment succeeded e-mail? If not, we need to figure that out.

[~bchen@mozilla.com]: Does SubPlat own these emails? If so, how do we change them? If not, who do we need to contact?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2804)
┆Issue Number: FXA-2804

data-sync-user:
➤ Barry Chen commented:

Yes, we own the emails.  They are in fxa-auth-server.  See fxa/packages/fxa-auth-server/lib/senders/email.js and the sendSubscription*Email functions in fxa/packages/fxa-auth-server/lib/routes/subscriptions.ts

data-sync-user:
➤ Barry Chen commented:

https://github.com/mozilla/fxa/#running-with-maildev ( https://github.com/mozilla/fxa/#running-with-maildev ) is very helpful when working with HTML emails.

data-sync-user:
➤ Lesley Norton commented:

We've decided not to send invoice receipts to customers checking out with PayPal. PayPal sends their own version of these emails and we are concerned that sending additional emails could lead to 1.) Users thinking they have been charged twice and 2.) Users feeling like we're cluttering up their inbox with superfluous emails.

data-sync-user:
➤ Bianca Danforth commented:

Legal says we need to keep sending these e-mails per Wil's comment in FXA-2989. Re-opening this ticket as TODO.

PR #7653, opened 2021-02-26
feat(auth-server): Modify transactional emails to accomodate PayPal payments

lesleyjanenorton:
## Because

- We need to modify some transactional emails when the user's method of paying is through PayPal

## This pull request

- Updates  `extractInvoiceDetailsForEmail()` and relevant email handlers so that `payment_provider` is available to the email templates. 
- Adds new email partial (paymentProvider.html) which handles what text we show based on the payment_provider
- Adds new handlebar helper {{#ifEquals something someOtherThing}} so that we can conditionally show bits of UI based on whether payment_provider === "stripe" or payment_provider === "paypal"
- Updates tests
- Corrects language in `subscriptionPaymentFailed`: Changes "3 days" to "24 hours". Maybe we should say "1 day" instead? I don't know. How do we ensure string changes bubble up to l10n?

Closes: #7264 #7101 

## Checklist

- [x] My commit is GPG signed

## Screenshots (Optional)

**Proposed language change to reactivation email**
Screen Shot 2021-02-25 at 7 53 59 PM
<br />
<br />
**Proposed language change in subscriptionPaymentFailed.**
Needs UX: Should we say "1 day" instead of "24 hours" ?
Screen Shot 2021-02-25 at 8 00 36 PM

**subscriptionFirstInvoice** when payment_method === "paypal"
Screen Shot 2021-02-25 at 8 11 34 PM

**subscriptionSubsequentInvoice** when payment_method === "paypal"
Screen Shot 2021-02-25 at 8 09 12 PM

How do we get updated strings to l10n? These emails are localized but the plumbing is an area of great mystery to me.


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -557,6 +557,18 @@ export class StripeHelper {
...
+    const subscription = customer.subscriptions?.data.find(
+      (sub: { status: string }) => ['active', 'past_due'].includes(sub.status)
+    );
+    if (subscription) {
+      return subscription.collection_method === 'send_invoice'
+        ? 'paypal'
+        : 'stripe';
+    }
+    return 'not_chosen';
+  }
```
chenba, line 570: nit: this function doesn't really need to be in a class.

packages/fxa-auth-server/lib/senders/templates/index.js
```diff
@@ -47,6 +49,13 @@ async function init(log) {
     return reduceOp(arguments, (a, b) => a || b);
   }
 
+  function isEqualHelper(a, b, options) {
+    if (a === b) {
+      return options.fn(this);
+    }
+    return options.inverse(this);
+  }
```
chenba, line 57: Consistency suggestion: using the pattern from `orHelper` above we can write
```
function isEqualHelper() {
  return reduceOp(arguments, (a, b) => a === b);
}
```

packages/fxa-auth-server/lib/senders/templates/partials/paymentProvider.html
```diff
@@ -0,0 +1,10 @@
+<br>
+{{#ifEquals payment_provider "paypal"}}
+  {{t "PayPal" }}
+{{/ifEquals}}
+
+{{#ifEquals payment_provider "stripe"}}
+  {{#if lastFour}}
+    {{t "%(cardType)s card ending in %(lastFour)s" }}
+  {{/if}}
+{{/ifEquals}}
```
chenba, line 10: Would be nice if we can keep in the logic in TS/JS and just pass a string to teh templates.

packages/fxa-auth-server/lib/senders/templates/subscriptionPaymentFailed.html
```diff
@@ -7,9 +7,9 @@
   <br><br>
   {{t "It may be that your credit card has expired, or your current payment method is out of date." }}
   <br><br>
-  {{{t "Help us fix it by <a href=\"%(updateBillingUrl)s\">updating your payment information</a> within three days." }}}
+  {{{t "Help us fix it by <a href=\"%(updateBillingUrl)s\">updating your payment information</a> within 24 hours." }}}
   <br><br>
-  {{t "Unfortunately, if we don’t hear from you within three days, you’ll lose access to %(productName)s." }}
+  {{t "Unfortunately, if we don’t hear from you within 24 hours, you’ll lose access to %(productName)s." }}
```
chenba, line 12: 👍

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -306,12 +311,10 @@ const TESTS = new Map([
       { test: 'include', expected: configHref('subscriptionSettingsUrl', 'subscription-reactivation', 'cancel-subscription', 'plan_id', 'product_id', 'uid', 'email') },
       { test: 'include', expected: configHref('subscriptionTermsUrl', 'subscription-reactivation', 'subscription-terms') },
       { test: 'include', expected: `reactivating your ${MESSAGE.productName} subscription` },
-      { test: 'include', expected: `be ${MESSAGE_FORMATTED.invoiceTotal} to the MasterCard card ending in ${MESSAGE.lastFour} on 04/19/2020.` },
       { test: 'notInclude', expected: 'utm_source=email' },
     ]],
     ['text', [
       { test: 'include', expected: `reactivating your ${MESSAGE.productName} subscription` },
-      { test: 'include', expected: `be ${MESSAGE_FORMATTED.invoiceTotal} to the MasterCard card ending in ${MESSAGE.lastFour} on 04/19/2020.` },
```
chenba, line 314: Don't we still need these tests?  I see in the partial above that we are still displaying the card type and last 4 digits.

chenba:
@lesleyjanenorton  I tried this locally and didn't get any emails when I paid with PayPal.  It did work when I paid with Stripe.  Is it working for you?

I don't think it's directly related to this PR; I'm confused as to why I'm not even getting the welcome email though.

packages/fxa-auth-server/lib/senders/templates/partials/paymentProvider.html
lesleyjanenorton: Yeah I agree - especially when there are more than 2 payment providers but I added the `{{#ifEquals}}...{{/ifEquals}}` checks in the templates since that is where the `{{#if lastFour}}..{{/if}}` was/is.

packages/fxa-auth-server/test/local/senders/email.js
lesleyjanenorton: Nope- for the reactivation email we are actually removing the language that contained the customer's card details and showing the same message to all customers, regardless of payment provider.

lesleyjanenorton:
@chenba Emails triggered by PayPal checkouts should be working now

packages/fxa-auth-server/lib/senders/templates/index.js
lesleyjanenorton: I can't find an instance of `orHelper` actually being used? I did try to follow that pattern initially but it results in this: 
Screen Shot 2021-03-01 at 7 51 02 PM

chenba: 🤦 Sorry about that.  In that case, please delete that `orHelper` dead code.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -76,8 +76,8 @@ export class StripeWebhookHandler extends StripeHandler {
             await this.handleInvoiceOpenEvent(request, event);
           }
           break;
-        case 'invoice.payment_succeeded':
-          await this.handleInvoicePaymentSucceededEvent(request, event);
+        case 'invoice.paid':
```
chenba, line 79: 👍

bbangert (approved):
Looks good to me!

chenba (approved):
WFM

packages/fxa-auth-server/lib/senders/templates/index.js
lesleyjanenorton: ok so - I was mistaken! This helper _is_ actually used [here](https://github.com/mozilla/fxa/blob/5c22cafd576ff7bcc28e98cb801297dde9579ca6/packages/fxa-auth-server/lib/senders/templates/downloadSubscription.html#L90).  I had been looking for `{{#or}}` but the usage was just a bit different. I've updated ifEqualHelper to match.

PR #7663, opened 2021-02-26
fix(auth-server): link invoice handler to webhook

bbangert:
Because:

* The paypal invoice open handler was not linked into the webhook
  routing to actually run.

This commit:

* Adds a case statement for `invoice.finalize` which triggers when
  the invoice becomes open to process the invoice with PayPal if
  needed.

## Checklist


chenba (approved):
👍 but do we need a test?

bbangert:
The other methods are run the same way, but I do agree that the main handler itself should have a test that verifies all the events are dispatched to. Maybe a generic test of some sort.

bbangert:
@chenba filed https://jira.mozilla.com/browse/FXA-3106 to address having a better stripe webhook test.

PR #7664, opened 2021-02-26
feat(auth-server): add paypal BA update handler

### LINKED ISSUE #7453: Create paypal update endpoint handler
An endpoint similar to the updateDefaultPaymentMethod handler needs to exist for the subscription management page. If there's an open invoice for a subscription that needs payment, it should have a payment attempt run.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2995)
┆Issue Number: FXA-2995

bbangert:
Because:

* We want to let customers who no longer have a valid paypal billing
  agreement on file to add a new one.

This commit:

* Adds a new handler for setting up a paypal billing agreement on a
  customer previously using paypal.

Closes #7453

## Checklist


packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +113,137 @@ export class PayPalHandler extends StripeHandler {
...
+        await this.paypalHelper.processInvoice({ customer, invoice });
+      }
+    } catch (err) {
+      reportSentryError(err, request);
+    }
+  }
+
+  /**
+   * Create and verify a billing agreement is funded from the appropriate
+   * country given the currency of the billing agreement.
```
birdsarah, line 209: given the currency of the customer - and maybe pass in the customer instead? Only a suggestion but this method feels a bit divorced from it's context. Maybe `createAndVerifyBillingAgreementForExistingPaypalCustomer` and then checks for existing subscriptions and currency could happen in here too?

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +113,137 @@ export class PayPalHandler extends StripeHandler {
...
+    if (this.stripeHelper.getCustomerPaypalAgreement(customer)) {
+      throw error.billingAgreementExists(customer.id);
+    }
+
+    const paypalSubscription = customer.subscriptions?.data.find(
+      (sub) =>
+        ['active', 'past_due'].includes(sub.status) &&
+        sub.collection_method === 'send_invoice'
+    );
+    if (!paypalSubscription || !customer.currency) {
```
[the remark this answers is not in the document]
birdsarah: Or actually. The customer.currency must be there if there's a paypalSubscription, and it's irrelevant if there's no paypalSubscription, so skip this check.

birdsarah (approved):
Looks fine. Made some optional structural comments.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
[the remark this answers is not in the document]
bbangert: It's in there to narrow the type, as `customer.currency` could be undefined until this line runs, which would prevent this from compiling due to line 162.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
bbangert: This is called from two contexts, in one of them, its being used for the first time where there has never been a subscription, so the customer object will not have a `currency` attribute. In that case it needs the plan currency, in the other context it has the customer currency. That's the main reason it doesn't have any additional logic folded into it, its just setting up the billing agreement and attaching it to the customer while verifying the billing agreement is in the right country for the currency chosen.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +113,137 @@ export class PayPalHandler extends StripeHandler {
...
+      token,
+      currency: customer.currency,
+    });
+
+    await this.stripeHelper.updateCustomerPaypalAgreement(
+      customer,
+      agreementId
+    );
+
+    const now = Date.now() / 1000;
```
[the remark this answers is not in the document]
bbangert: Created a small helper function for this, and updated the `fetchInvoices` doc string to be more useful about time units used by Stripe.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
birdsarah: got it, i missed the ref in createsubscription. no more late night code review.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +114,137 @@ export class PayPalHandler extends StripeHandler {
...
+   *
+   * Updates are not allowed if there's:
+   *  - No paypal sourced subscriptions active/past_due
+   *  - An existing billing agreement on file.
+   *
+   * If there's open invoices, attempts to pay them will be made in
+   * the background.
+   */
+  async updatePaypalBillingAgreement(request: AuthRequest) {
+    this.log.begin('subscriptions.createSubscriptionWithPaypal', request);
```
chenba, line 129: s/createSubscriptionWithPaypal/updatePaypalBillingAgreement/

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +114,137 @@ export class PayPalHandler extends StripeHandler {
...
+    const { uid, email } = await handleAuth(this.db, request.auth, true);
+    await this.customs.check(request, email, 'createSubscriptionWithPaypal');
```
chenba, line 131: s/createSubscriptionWithPaypal/updatePaypalBillingAgreement/ here as well

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +114,137 @@ export class PayPalHandler extends StripeHandler {
...
+
+    if (this.stripeHelper.getCustomerPaypalAgreement(customer)) {
+      throw error.billingAgreementExists(customer.id);
+    }
+
+    const paypalSubscription = customer.subscriptions?.data.find(
+      (sub) =>
+        ['active', 'past_due'].includes(sub.status) &&
+        sub.collection_method === 'send_invoice'
+    );
```
chenba, line 150: `getPaymentProvider` in `StripeHandler` pretty much does this.  Maybe move that function out into a lib and use it?

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -135,6 +114,137 @@ export class PayPalHandler extends StripeHandler {
...
+    if (!paypalSubscription || !customer.currency) {
+      throw error.internalValidationError('updatePaypalBillingAgreement', {
+        customerId: customer.id,
+        message: 'Cannot add PayPal with no paypal subscriptions.',
+        uid,
+      });
```
chenba, line 156: If the error is thrown when the customer doesn't have a currency, then this message is misleading.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -206,5 +316,26 @@ export const paypalRoutes = (
...
+          payload: false,
+          strategy: 'oauthToken',
+        },
+        response: {
+          schema: validators.subscriptionsStripeCustomerValidator,
+        },
+        validate: {
+          payload: {
+            token: validators.paypalPaymentToken.required(),
+            idempotencyKey: isA.string().required(),
```
chenba, line 333: I don't see this value used anywhere.

packages/fxa-auth-server/test/local/routes/subscriptions/paypal.js
```diff
@@ -237,4 +238,110 @@ describe('subscriptions payPalRoutes', () => {
...
+        '/oauth/subscriptions/paymentmethod/billing-agreement'
+      );
+      assert.deepEqual(actual, filterCustomer(customer));
+      sinon.assert.calledOnce(stripeHelper.customer);
+      sinon.assert.calledOnce(payPalHelper.createBillingAgreement);
+      sinon.assert.calledOnce(payPalHelper.agreementDetails);
+      sinon.assert.calledOnce(stripeHelper.updateCustomerPaypalAgreement);
+      sinon.assert.calledOnce(stripeHelper.fetchOpenInvoices);
+      sinon.assert.calledOnce(stripeHelper.getCustomerPaypalAgreement);
+      sinon.assert.calledOnce(payPalHelper.processZeroInvoice);
```
chenba, line 303: Maybe also assert that `processInvoice` was not called.

chenba (approved):
👍 r+ w/ comments

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
bbangert: It does, but that hasn't been merged yet afaik.

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
bbangert: Removed.

Issue #7516, opened 2021-02-12
Terms of Service is blank


data-sync-user:
This page is blank for me:  https://accounts.stage.mozaws.net/legal/terms

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3036)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15406/image-2021-02-17-11-27-57-750.png">image-2021-02-17-11-27-57-750.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15348/issue91.gif">issue91.gif</a> | <a href="https://jira.mozilla.com/secure/attachment/15403/privacy.png">privacy.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15304/Screenshot_2021-02-11+Terms+of+Service.png">Screenshot_2021-02-11 Terms of Service.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15347/Screenshot_5.png">Screenshot_5.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15404/terms-rendering.png">terms-rendering.png</a>
┆Issue Number: FXA-3036

data-sync-user:
➤ Dave Justice commented:

I'm not seeing this with the terms page, however I am seeing it with privacy. It appears to be an issue with loading the css, the content is in the dom.

!privacy.png!

data-sync-user:
➤ Wil Clouser commented:

QA said they are having problems with the CDN.  jbuck just flushed it - maybe that was making this bug too.

data-sync-user:
➤ Dave Justice commented:

Cannot reproduce currently in Firefox stable or Edge browser, only seeing this on privacy page in Nightly leads me to believe my browser is doing some hardcore caching here. I saw we wait till monday and see if we can reproduce

vasilicatamas:
"Term of Service" and "Privacy Notice" pages are still blank on Stage [Train 1.200.0] under Windows (Firefox 85.0.2), Android 9 and iOS 13.6.

![Screenshot_38](https://user-images.githubusercontent.com/18114210/107941005-be628200-6f91-11eb-85be-ac7da39a628e.png)

data-sync-user:
➤ Dave Justice commented:

!image-2021-02-17-11-27-57-750.png!

data-sync-user:
➤ Wil Clouser commented:

[...]

PR #7665, opened 2021-02-27
restore invisible terms / privacy pages

6a68:
## Because

- The terms and privacy pages aren't visible if they are visited via the link on the signin form (so that they are templated by the front-end code)

Pulling over my description of the bug from a comment on that issue:

> It turns out there used to be two versions of the terms and privacy notices: one rendered as a partial template, and one rendered as a full html page. The partials were lost in some recent changes, so when the page is served up, a page is loaded into the bigger page. The terms / privacy pages appear blank because there's a fade-in effect applied to the content wrapper, `<div id="stage">`, but now that an html page is getting loaded into another page, there are actually two nested content wrapper els with that ID, and the fade-in animation is only getting applied to the outer one:
> 
> Screen Shot 2021-02-25 at 3 53 55 PM
> 
> The fix is to update the l10n generator grunttasks to restore the missing partial templates.

## This pull request

- This commit builds on some recent l10n changes to fix this bug:

* restore the partial templates of the translated legal docs and, following the old approach, store them in /privacy and /terms directories under server/templates/pages/dist/.

* update the legal-templates lib to load the partials correctly

* Big open question: if I set defaultLegalLang to en, something breaks in the grunttasks. Are we okay with the default legal language being en_US, or does it need to be en?

For testing locally:
* the partial should render correctly if you navigate from the signin form to the terms or privacy pages (by clicking the link in the email input form).
* the full page should render correctly If you go to the /terms or /privacy pages directly, or if you follow the link from the signin form, then hard refresh the page (which triggers a server render).

Closes: #7516

## Checklist

## Screenshots (Optional)


6a68:
Hey @meandavejustice, I know you're busy with other stuff, but some of these changes build on the l10n changes you landed recently. Mind taking a quick look at this draft, just to call out anything important I might have missed (or broken :-P)?

packages/fxa-content-server/grunttasks/remarkable.js
```diff
@@ -12,15 +12,9 @@ module.exports = function (grunt) {
...
 
-    if (destFile.includes('privacy')) {
-      return path.join(destPath, lang, 'privacy.html');
-    } else if (destFile.includes('tos')) {
-      return path.join(destPath, lang, 'terms.html');
-    } else {
-      return path.join(destPath, i18n.localeFrom(destFile) + '.html');
-    }
+    const lang = i18n.localeFrom(destFile.split('/')[0]);
+    return path.join(destPath, lang + '.html');
```
6a68, line 17: for example, the path is something like `/server/templates/pages/dist/privacy/en_US.html`

packages/fxa-content-server/grunttasks/yeoman.js
```diff
@@ -16,12 +16,14 @@ module.exports = function (grunt) {
     es5: '.es5',
     page_template_dist: TEMPLATE_ROOT + '/pages/dist',
     page_template_src: TEMPLATE_ROOT + '/pages/src',
+    pp_html_dest: TEMPLATE_ROOT + '/pages/dist/privacy',
```
6a68, line 19: this and tos_html_dest are needed to put the partials in the correct locations

packages/fxa-content-server/server/lib/legal-templates.js
```diff
@@ -14,7 +14,7 @@ module.exports = function (i18n, root) {
 
     // Filenames are normalized to locale, not language.
     const locale = i18n.localeFrom(lang);
-    const templatePath = path.join(root, locale, type + '.html');
+    const templatePath = path.join(root, type, locale + '.html');
```
6a68, line 17: This matches the path in the grunttasks/remarkable diff above

dannycoates (approved):
lgtm if we fix the failing tests

6a68:
> lgtm if we fix the failing tests

s/en_US/en-US 🙄

Issue #7470, opened 2021-02-09
Update PayPal invoice processing to pass through plan currency


data-sync-user:
The invoice processing for paypal customers should include the plan currency in the reference transaction call.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3008)
┆Issue Number: FXA-3008

PR #7666, opened 2021-02-27
feat(auth-server): Paypal chargeCustomer must be called with currency…

birdsarah:
## Because

- We need to charge paypal customer in correct currency

## This pull request

- Adds currencyCode to doReferenceTransaction and upstream chargeCustomer method.

Closes: #7470

## Checklist

## Other information 

I verified that without this a FR customer would be charged in USD (aka Before):
Screenshot from 2021-02-26 19-14-26

And with this, they are charged in EUR (aka After): 
Screenshot from 2021-02-26 19-14-06

### DIFF
--- a/packages/fxa-auth-server/test/local/payments/paypal.js
+++ b/packages/fxa-auth-server/test/local/payments/paypal.js
@@ -211,6 +211,7 @@ describe('PayPalHelper', () => {
     const validOptions = {
       amountInCents: 1099,
       billingAgreementId: 'B-12345',
+      currencyCode: 'usd',
       invoiceNumber: 'in_asdf',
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/test/local/payments/paypal.js
```diff
@@ -489,7 +493,7 @@ describe('PayPalHelper', () => {
       const validInvoice = {
         ...mockInvoice,
         status: 'draft',
-        amount_due: 4.99,
+        amount_due: 499,
```
birdsarah, line 496: This is cleanup from a previous PR to remind future folks that amount_due is an amountInCents.

bbangert (approved):
Nice!

PR #7670, opened 2021-02-27
feat(auth-server): remove paypal BA and cancel at end

### LINKED ISSUE #7103: Update Stripe webhook handler code for cancelled subscription state
The stripe webhook handler for subscription cancelled should be updated to remove the customer billing agreement if its a PayPal user and there are no other active/past_due subscriptions.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2807)
┆Issue Number: FXA-2807

bbangert:
Because:

* We want to cancel the paypal billing agreement and remove it from
  the customer when they no longer have active/past_due subscriptions
  being paid for with paypal.

This commit:

* Hooks into the subscription cancellation event to determine if its the
  last one paid by paypal, and cancels/removes the billing agreement if
  that's the case.

Closes #7103

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal.ts
@@ -347,6 +347,36 @@ export class PayPalHelper {
   }
 
   /**
+   * Removes Paypal billing agreements on a customer if they paid with
+   * Paypal but no longer have an active/past_due subscription.
+   */
+  async conditionallyRemoveBillingAgreement(
+    customer: Stripe.Customer
+  ): Promise<boolean> {
+    const billingAgreementId = this.stripeHelper.getCustomerPaypalAgreement(
+      customer
+    );
+    if (!billingAgreementId) {
+      return false;
+    }
+    const paypalSubscription = customer.subscriptions?.data.find(
+      (sub) =>
+        ['active', 'past_due'].includes(sub.status) &&
+        sub.collection_method === 'send_invoice'
+    );
+    if (paypalSubscription) {
+      return false;
+    }
+    await this.cancelBillingAgreement(billingAgreementId);
+    await this.stripeHelper.removeCustomerPaypalAgreement(
+      customer.metadata.userid,
+      customer,
+      billingAgreementId
+    );
+    return true;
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -346,6 +346,36 @@ export class PayPalHelper {
...
+      customer
+    );
+    if (!billingAgreementId) {
+      return false;
+    }
+    const paypalSubscription = customer.subscriptions?.data.find(
+      (sub) =>
+        ['active', 'past_due'].includes(sub.status) &&
+        sub.collection_method === 'send_invoice'
+    );
```
chenba, line 366: deja vu all over again :-P

packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -346,6 +346,36 @@ export class PayPalHelper {
...
+      return false;
+    }
+    await this.cancelBillingAgreement(billingAgreementId);
+    await this.stripeHelper.removeCustomerPaypalAgreement(
+      customer.metadata.userid,
+      customer,
+      billingAgreementId
+    );
+    return true;
+  }
```
chenba, line 377: Why is this function returning a bool?  That return value is ultimately ignored in `handleWebhookEvent`.

chenba: I do see that you are using it in tests and it does make that easier.

packages/fxa-auth-server/lib/payments/paypal.ts
bbangert: Indeed, it'll be nice to have a single function that does this in shared that can be used everywhere.

packages/fxa-auth-server/lib/payments/paypal.ts
bbangert: While it doesn't need to return anything, it does help as another safe-guard on testing to see the branch went the right way.

chenba (approved):
👍

Issue #7677, opened 2021-03-01
Checkout: Display SubscriptionCreate screen while PayPal window is open during authorization


data-sync-user:
FXA-2994 added a new processing screen during PayPal checkout that displays after clicking the PayPal button to open a PayPal window.

Per [the design spec for this step in Checkout](https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=519%3A5608), we would like to show the SubscriptionCreate screen here, as if the user changes their mind and closes the PayPal window, they're back at the first step.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3108)
┆Issue Number: FXA-3108

data-sync-user:
➤ Bianca Danforth commented:

Just tacking on a small fix here in the scope of this bug as well: Don't auto-focus the input field on the Stripe widget on SubscriptionCreate.

PR #7680, opened 2021-03-01
feat(payments): Show processing screen onApprove for PayPalButton

biancadanforth:
## Because

- We want users to return to the first step of Checkout if they close the PayPal window (see [design spec](https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=519%3A5608)). This is a follow-up from #7450 .

## This pull request

- Replaces the `PayPalButton` `onClick` handler with calling `setTransactionInProgress` in its `onApprove` method. This means the `PaymentProcessing` screen is only shown after successul authorization while we are waiting for the `POST /oauth/subscription/active/new-paypal` request to return.

Closes: #7677 

## Checklist

## Screenshots
Before
Screen Shot 2021-03-01 at 2 59 14 PM

After
Screen Shot 2021-03-01 at 2 56 25 PM

## Notes
* Also remove autofocus from the Name field on the Stripe widget, since it's no longer the only interactive element on the screen.

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -182,7 +182,6 @@ export const PaymentForm = ({
           data-testid="name"
           placeholder="Full Name"
           required
-          autoFocus
           spellCheck={false}
           onValidate={(value, focused, props) =>
             validateName(value, focused, props, getString)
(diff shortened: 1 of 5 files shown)


biancadanforth:
`Jest: "global" coverage threshold for branches (94%) not met: 93.96%` -- working on it...haha.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -182,7 +182,6 @@ export const PaymentForm = ({
           data-testid="name"
           placeholder="Full Name"
           required
-          autoFocus
```
chenba, line 185: 💯 🏅

chenba (approved):
make sense to me

Issue #7615, opened 2021-02-24
Provide refund when we receive IPN for successful transaction with cancelled subscription


data-sync-user:
For customers with subscriptions that have pending transactions, we immediately cancel the subscription when the invoice is not paid within the grace period (24 hours). Since the transaction in PayPal can still go through (there is no way to cancel the pending transaction), we want to listen for the successful transaction IPN and refund the transaction.

Acceptance Criteria:

* while handling successful paypal transaction in "handleMerchPayment" function, check that invoice for the subscription is marked uncollectible then issue a refund
* we should make sure we have not already refunded this (since we use the msg_sub_id to make the refund, this should never happen)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3086)
┆Issue Number: FXA-3086

PR #7681, opened 2021-03-02
feat(auth-server): refund paypal transaction

say-yawn:
## Because

- There is no way to cancel the pending transaction, we want to listen for the successful transaction IPN and refund the transaction for invoice with uncollectible status.

## This pull request

- Adds refund transaction method in PayPal client and helper to refund successful merchant payment with uncollectible invoice found in IPN.

Closes: #7615

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
@@ -31,10 +31,15 @@ export class PayPalNotificationHandler extends PayPalHandler {
       });
     }
 
-    // TODO: If the invoice is void/uncollectible and this is Completed/Processed, issue
-    // a refunded.
-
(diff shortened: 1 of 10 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -31,10 +31,38 @@ export class PayPalNotificationHandler extends PayPalHandler {
...
+          this.log.error('handleMerchPayment', {
+            message: 'PayPal refund transaction unsuccessful',
+            ipnMessage: message,
+            refundResponse,
+          });
+          throw error.internalValidationError('handleMerchPayment', {
+            message: 'PayPal refund transaction unsuccessful',
+          });
+        }
+      }
```
bbangert, line 65: This whole block would be a bit easier to parse if it was its own function, `issueRefund` or something.

bbangert (approved):
Looks good!

Issue #7073, opened 2020-12-04
Add a new event property to Checkout related Amplitude events


data-sync-user:
This ticket covers events specific to Checkout.

--

We currently record metrics related to the funnel for subscriptions. We should record when a user signs up with PayPal.

A primer on metrics in FxA can be found at [https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-metrics.](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-metrics.)  The type of metrics in discussion for this issue is Amplitude.  There's a list of these events at [https://docs.google.com/spreadsheets/d/1ttgV60YeurrMhmdlCZ5WNBJXa-PcnRmrGgOEow71E1U/edit#gid=0.](https://docs.google.com/spreadsheets/d/1ttgV60YeurrMhmdlCZ5WNBJXa-PcnRmrGgOEow71E1U/edit#gid=0.)

Amplitude allows a dictionary of "event properties" to be attached to each event.  For events where the user has the option of using Stripe or PayPal, we should add an event property of "payment_provider" with the value "Stripe" or "PayPal".

An example of such an event is the "fxa_pay_setup - success" event.  That event is emitted at [https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/lib/apiClient.ts#L239:L242, where you can see the event properties are directly passed into the function call.](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/lib/apiClient.ts#L239:L242)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2789)
┆Issue Number: FXA-2789

chenba:
@irrationalagent @davismtl The current plan is to add an event property of "payment_provider" with the possible values of "Stripe" and "PayPal" to some of the SubPlat related Amplitude events.  Let us know if you think we should do something different.

PR #7701, opened 2021-03-03
feature: Add a new event property to Checkout related Amplitude events

rlr:
## Because

We should have metrics parity between Stripe and PayPal checkout events and we should be able to distinguish between them.

## This pull request

Adds pings to the `apiCapturePaypalPayment` call.

Closes: #7073

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-payments-server/src/lib/amplitude.test.ts
+++ b/packages/fxa-payments-server/src/lib/amplitude.test.ts
@@ -68,23 +68,31 @@ it('should capture error during logAmplitudeEvent', () => {
 });
 
 it('should call logAmplitudeEvent with subscription plan info', () => {
-  const plan = { plan_id: '123xyz_hourly', product_id: '123xyz' };
-  Amplitude.createSubscription_PENDING(plan);
+  const metricsData: Amplitude.EventProperties = {
+    plan_id: '123xyz_hourly',
+    product_id: '123xyz',
+    paymentProvider: 'Stripe',
+  };
+  Amplitude.createSubscription_PENDING(metricsData);
   const eventProps = (<jest.Mock>(
     FlowEvent.logAmplitudeEvent
   )).mock.calls[0].pop();
 
   expect(eventProps).toMatchObject({
-    planId: plan.plan_id,
-    productId: plan.product_id,
+    planId: metricsData.plan_id,
+    productId: metricsData.product_id,
+    paymentProvider: metricsData.paymentProvider,
   });
 });
 
 it('should call logAmplitudeEvent with reason for failure on fail event', () => {
-  const plan = { plan_id: '123xyz_hourly', product_id: '123xyz' };
+  const metricsData: Amplitude.EventProperties = {
+    plan_id: '123xyz_hourly',
+    product_id: '123xyz',
+  };
   const payload = { message: 'oopsie daisies' };
   Amplitude.createSubscription_REJECTED({
-    ...plan,
+    ...metricsData,
     error: payload,
   });
   const eventProps = (<jest.Mock>(
@@ -92,8 +100,8 @@ it('should call logAmplitudeEvent with reason for failure on fail event', () => 
   )).mock.calls[0].pop();
 
   expect(eventProps).toMatchObject({
-    planId: plan.plan_id,
-    productId: plan.product_id,
+    planId: metricsData.plan_id,
+    productId: metricsData.product_id,
     reason: payload.message,
   });
 });
--- a/packages/fxa-payments-server/src/lib/amplitude.ts
+++ b/packages/fxa-payments-server/src/lib/amplitude.ts
@@ -32,11 +32,12 @@ type GlobalEventProperties = {
   uid?: string;
 };
 
-type EventProperties = GlobalEventProperties & {
+export type EventProperties = GlobalEventProperties & {
   planId?: string;
   plan_id?: string;
   productId?: string;
   product_id?: string;
+  paymentProvider?: 'Stripe' | 'PayPal';
   error?: Error;
 };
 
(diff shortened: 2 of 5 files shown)


biancadanforth:
@chenba
* How would we/QA verify that a specific Amplitude event was sent? I don't see information on that in the [Ecosystem docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-metrics#amplitude).
* Are there SubPlat dashboards that would need to be updated? Who updates those?

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -316,6 +316,7 @@ describe('API requests', () => {
     const metricsOptions = {
       planId: params.priceId,
       productId: params.productId,
+      paymentMethod: 'Stripe',
```
chenba, line 319: is this meant to be `paymentProvider`?

rlr: uhm derp. Yep!

chenba:
Thanks for tackling Amplitude event work! 😅 

Currently this PR doesn't seem to work.  You can check for the events in the payment server's log.  Tail and (rip)grep for `amplitudeEvent`.

The final event object is produced in https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/metrics/amplitude.js#L274:L298, and if you follow that to `mapEventProperties`, you'll see where the product and plan ids are added.  `mapSubscriptionPaymentEventProperties` is an example of an event property that's added to only certain event type.

> 
> 
> @chenba, @bbangert
> 
>     * How would we/QA verify that a specific Amplitude event was sent? I don't see information on that in the [Ecosystem docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-metrics#amplitude).
>       
>       * We should probably add some QA test cases for verifying new PayPal checkout Amplitude events.
> 

I cannot remember whether QA has Amplitude access.  That's a @rbillings question.  But it'd be pretty easy to query for events in Amplitude after QA done testing, since the events in Stage do go into Amplitude.

>     * Are there SubPlat dashboards that would need to be updated? Who updates those?

It used to be jgruen afaik, but he's no longer at Moz.  @davismtl now maybe?  @davismtl for awareness anyways.

rlr:
@chenba I *think* this is good to go now?

packages/fxa-payments-server/src/lib/amplitude.ts
```diff
@@ -93,12 +95,15 @@ const normalizeEventProperties = (eventProperties: EventProperties) => {
     plan_id = undefined,
     productId = undefined,
     product_id = undefined,
+    paymentProvider = undefined,
+    payment_provider = undefined,
```
biancadanforth, line 99: Based on the comment above this function:
```
// Accepting both planId & plan_id allows us to directly use a Plan type
// as EventProperties
```
It doesn't seem like we'd need to have the snake case version of `paymentProvider` in various places?

biancadanforth (commented):
Thanks Ricky! I did verify using the `payments` server logs that we are getting these four amplitude events with the new event property `payment_provider`. I have one inline suggestion around whether we need both `payment_provider` and `paymentProvider`.

packages/fxa-payments-server/src/lib/amplitude.test.ts
```diff
@@ -68,7 +68,11 @@ it('should capture error during logAmplitudeEvent', () => {
 });
 
 it('should call logAmplitudeEvent with subscription plan info', () => {
-  const plan = { plan_id: '123xyz_hourly', product_id: '123xyz' };
+  const plan = {
```
chenba, line 71: Didn't occur to me to nit here last time: since `payment_provider` isn't part of a `plan`, this name is now misleading.

packages/fxa-payments-server/src/lib/amplitude.ts
chenba: Yeah, let's rely on the type system to complain when someone try to pass a `payment_provider` property.

chenba (approved):
👍 Since @biancadanforth already verified

packages/fxa-payments-server/src/lib/amplitude.test.ts
```diff
@@ -68,32 +68,41 @@ it('should capture error during logAmplitudeEvent', () => {
 });
 
 it('should call logAmplitudeEvent with subscription plan info', () => {
-  const plan = { plan_id: '123xyz_hourly', product_id: '123xyz' };
-  Amplitude.createSubscription_PENDING(plan);
+  const metricsData: {
+    plan_id: string;
+    product_id: string;
+    paymentProvider?: 'Stripe' | 'PayPal';
```
rlr, line 74: had to add this due to weird typescript issue. It was complaining that `paymentProvider` below was a string and didn't match the type of `'Stripe' | 'PayPal'...` 🤷‍♂️

rlr:
I removed passing around the snake case `payment_provider`. But now I'm confused if that was right.

packages/fxa-payments-server/src/lib/amplitude.test.ts
chenba: We could also export the `EventProperties` type where you added this sum type, then we can use `EventProperties['paymentProvider']` here.

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -253,9 +275,14 @@ export async function apiCreateSubscriptionWithPaymentMethod(params: {
   };
 }> {
   const { priceId, paymentMethodId, idempotencyKey } = params;
-  const metricsOptions = {
+  const metricsOptions: {
+    planId: string;
+    productId: string;
+    paymentProvider?: 'Stripe' | 'PayPal';
```
chenba, line 281: Same with `EventProperties['paymentProvider']` here.

chenba:
> 
> 
> I removed passing around the snake case `payment_provider`. But now I'm confused if that was right.

Yeah :(  It does feel weird to pass customer's `payment_provider` as `paymentProvider`, and then it's set back to `payment_provider` in the event properties.

chenba (approved):
Still WFM 👍

rlr:
yeeee. ^^ I just rebased and squashed into one commit.

PR #7708, opened 2021-03-03
fix: remove sandbox credentials

bbangert:
Because:

* Sandbox credentials for paypal were accidentally comitted.

This commit:

* Removes them.

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -158,14 +158,7 @@
   },
   "subscriptions": {
     "enabled": true,
-    "sharedSecret": "devsecret",
-    "paypalNvpSigCredentials": {
-      "enabled": true,
-      "sandbox": true,
-      "user": "sb-i11ty5077364_api1.business.example.com",
-      "pwd": "Q43EHXXYVFTF8KPH",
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/config/dev.json
```diff
@@ -158,14 +158,7 @@
...
   "subscriptions": {
     "enabled": true,
-    "sharedSecret": "devsecret",
-    "paypalNvpSigCredentials": {
-      "enabled": true,
-      "sandbox": true,
-      "user": "sb-i11ty5077364_api1.business.example.com",
-      "pwd": "Q43EHXXYVFTF8KPH",
-      "signature": "ATq0Lw3ury0wjGtXkF30nozpLja-ASxng61sj1bMjEhIJJj3HAF02QB2"
-    }
```
say-yawn, line 168: Do we want to leave the values empty rather than deleting the entire `paypalNvpSigCredentials` configuration keys?

say-yawn (approved):
One recommendation, not required. LGTM.

packages/fxa-auth-server/config/dev.json
bbangert: This restores the prior state of the file, so I just left this approach.

Issue #7448, opened 2021-02-08
Gate showing the Subscribe screen with multiple payment options based on the modified Customer object


data-sync-user:
FXA-2902 and FXA-2976 will add the Paypal button and the functionality required to process Paypal checkout. However, its logic should ultimately be updated to include a check for existing Paypal customers, as it currently only checks for existing Stripe customers.

Closing this issue also includes adding the PayPal button to the default SubscriptionCreate storybook or to a new variant of the SubscriptionCreate storybook that has the button.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2992)
┆Issue Number: FXA-2992

data-sync-user:
➤ Bianca Danforth commented:

Getting the `PayPalButton` to render in the new customer `SubscriptionCreate` story would require adding Storybook-only code which is discouraged, so I'm descoping that from this ticket. See this thread ( https://github.com/mozilla/fxa/pull/7717#discussion_r586970500 ) in GitHub for more details.

PR #7717, opened 2021-03-04
feat(payments): Gate SubscriptionCreate screen based on new/returning…

biancadanforth:
… PayPal customers.

Also simplify PaymentLegalBlurb component on Checkout screens.

## Because

- We want to show different versions of the `SubscriptionCreate` page based on whether a customer is new customer, a returning Stripe customer or a returning PayPal customer.
- We want to show different privacy notice messages depending on which payment method is being used and ensure they are consistent throughout Checkout.

## This pull request

- Gates different versions of the `SubscriptionCreate` page based on the `customer.payment_provider` property added in #7115.
- Move logic for which payment-related legal links to show inside the `PaymentLegalBlurb` component to make the Checkout screens consistent and DRYer.
- Update the `SubscriptionCreate` stories to show these three variations with the PayPal button.

Note: The final screen for returning PayPal customers is TBD and will be handled by [FXA-3119](https://jira.mozilla.com/browse/FXA-3119) once we get some design input.

Closes: #7448 

## Checklist

## Screenshots (Optional)

No visible changes to new customer before/after

**Before (Returning Stripe customer)**
Both Stripe and PayPal privacy policy
Screenshot_2021-03-03 Storybook(3)

**After (Returning Stripe customer)**
Stripe privacy policy only
Screenshot_2021-03-03 Storybook(1)

**Before (Returning PayPal customer)**
* New customer screen was incorrectly showing
* Both Stripe and PayPal privacy policy
Screenshot_2021-03-03 Firefox Accounts

**After (Returning PayPal customer)**
* Main content is blank (final screen TBD in [FXA-3119](https://jira.mozilla.com/browse/FXA-3119))
* PayPal privacy policy only
Screenshot_2021-03-03 Storybook(2)

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
@@ -6,10 +6,7 @@ import { Plan, Profile, Customer } from '../../store/types';
 import { PaymentProviderDetails } from '../PaymentProviderDetails';
 import SubscriptionTitle from '../SubscriptionTitle';
 import { TermsAndPrivacy } from '../TermsAndPrivacy';
-import {
-  PaypalPaymentLegalBlurb,
-  StripePaymentLegalBlurb,
-} from '../PaymentLegalBlurb';
+import PaymentLegalBlurb from '../PaymentLegalBlurb';
 
 import circledCheckbox from './images/circled-confirm.svg';
 
@@ -135,8 +132,7 @@ export const PaymentConfirmation = ({
               Continue to download
             </a>
           </Localized>
-          {Provider.isPaypal(payment_provider) && <PaypalPaymentLegalBlurb />}
-          {Provider.isStripe(payment_provider) && <StripePaymentLegalBlurb />}
+          <PaymentLegalBlurb provider={payment_provider} />
           <TermsAndPrivacy plan={selectedPlan} />
         </div>
       </section>
--- a/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentForm/index.tsx
@@ -39,6 +39,7 @@ import { Plan, Customer } from '../../store/types';
 import { productDetailsFromPlan } from 'fxa-shared/subscriptions/metadata';
 
 import './index.scss';
+import * as PaymentProvider from '../../lib/PaymentProvider';
 
 export type PaymentSubmitResult = {
   stripe: Stripe;
@@ -82,8 +83,10 @@ export const PaymentForm = ({
   onChange: onChangeProp,
   submitNonce,
 }: BasePaymentFormProps) => {
-  const hasExistingCard =
-    customer && customer.last4 && customer.subscriptions.length > 0;
+  const isExistingStripeCustomer =
+    customer &&
+    PaymentProvider.isStripe(customer?.payment_provider) &&
+    customer.subscriptions.length > 0;
 
   const stripe = useStripe();
   const elements = useElements();
@@ -121,7 +124,7 @@ export const PaymentForm = ({
       const { name } = validator.getValues();
       const card = elements.getElement(CardElement);
       /* istanbul ignore next - card should exist unless there was an external stripe loading error, handled above */
-      if (hasExistingCard || card) {
+      if (isExistingStripeCustomer || card) {
         onSubmitForParent({
           stripe,
           elements,
@@ -157,7 +160,7 @@ export const PaymentForm = ({
       navigatorLanguages
     ));
   }
-  const paymentSource = hasExistingCard ? (
+  const paymentSource = isExistingStripeCustomer ? (
     <div className="card-details" data-testid="card-details">
       <Localized
         id="sub-update-card-ending"
--- a/packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Localized } from '@fluent/react';
 
+import * as PaymentProvider from '../../lib/PaymentProvider';
+
 import './index.scss';
 
 function getPrivacyLinkText(): string {
@@ -15,7 +17,7 @@ function getStripePrivacyLinkText(): string {
   return 'View the <stripePrivacyLink>Stripe privacy policy</stripePrivacyLink>.';
 }
 
-export const PaypalPaymentLegalBlurb = () => (
+const PaypalPaymentLegalBlurb = () => (
   <div className="payment-legal-blurb">
     <Localized id="payment-legal-copy-paypal">
       <p>Mozilla uses Paypal for secure payment processing.</p>
@@ -38,7 +40,7 @@ export const PaypalPaymentLegalBlurb = () => (
   </div>
 );
 
-export const StripePaymentLegalBlurb = () => (
+const StripePaymentLegalBlurb = () => (
   <div className="payment-legal-blurb">
     <Localized id="payment-legal-copy-stripe">
       <p>Mozilla uses Stripe for secure payment processing.</p>
@@ -61,7 +63,7 @@ export const StripePaymentLegalBlurb = () => (
   </div>
 );
 
-export const PaymentLegalBlurb = () => (
+const DefaultPaymentLegalBlurb = () => (
   <div className="payment-legal-blurb">
     <Localized id="payment-legal-copy-stripe-paypal">
       <p>Mozilla uses Stripe and Paypal for secure payment processing.</p>
@@ -90,5 +92,18 @@ export const PaymentLegalBlurb = () => (
     </Localized>
   </div>
 );
+
+export type PaymentLegalBlurbProps = {
+  provider: PaymentProvider.ProviderType | undefined;
+};
+
+export const PaymentLegalBlurb = ({ provider }: PaymentLegalBlurbProps) => {
+  return (
+    (PaymentProvider.isPaypal(provider) && <PaypalPaymentLegalBlurb />) ||
+    (PaymentProvider.isStripe(provider) && <StripePaymentLegalBlurb />) || (
+      <DefaultPaymentLegalBlurb />
+    )
+  );
+};
 
 export default PaymentLegalBlurb;
--- a/packages/fxa-payments-server/src/lib/PaymentProvider.ts
+++ b/packages/fxa-payments-server/src/lib/PaymentProvider.ts
@@ -2,10 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export function isStripe(provider: string | undefined) {
+export type ProviderType = 'paypal' | 'stripe' | 'not_chosen';
+
+export function isStripe(provider: ProviderType | undefined): boolean {
   return provider === 'stripe';
 }
 
-export function isPaypal(provider: string | undefined) {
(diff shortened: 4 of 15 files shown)


packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -513,6 +513,7 @@ export const MOCK_ACTIVE_SUBSCRIPTIONS_AFTER_SUBSCRIPTION = [
 export const MOCK_CUSTOMER = {
   billing_name: 'Jane Doe',
   payment_type: 'card',
+  payment_provider: 'stripe',
```
biancadanforth, line 516: If I _don't_ modify `MOCK_CUSTOMER` to have the `payment_provider` key with a value of `"stripe"`, the `PaymentForm` test 'with existing card' > 'calls the submit handler' test case fails, but it passes if I _do_ modify it. I still don't understand why... Would you know? Both new and existing Stripe customer `PaymentForm`s have a submit button with a `data-testid="submit"`.

I do know that when `payment_provider` is `undefined`, we show the new customer version of `SubscriptionCreate`. When it's `"stripe"` we show the returning Stripe customer version of `SubscriptionCreate`.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -92,6 +93,19 @@ export const SubscriptionCreate = ({
...
+  const paymentProvider: ProviderType = customer?.payment_provider;
+
+  const isStorybookContext = Array.from(
+    document.head.getElementsByTagName('SCRIPT')
+  ).filter(
+    (ele) =>
+      // @ts-ignore: HTMLElement does not have property 'src'
+      ele.src ===
+      'https://www.paypal.com/sdk/js?client-id=sb&vault=true&commit=false&intent=authorize&disable-funding=credit,card'
+  ).length;
```
biancadanforth, line 107: I added a `ts-ignore` comment for a storybook-only code block in `SubscriptionCreate` -- how can I cast this result as an `HTMLScriptElement` array?

Also, in general, this is kind of a hacky way to get the `PaypalButton` to load just for Storybook. Open to alternatives if you have them!

biancadanforth (commented):
Just added some questions in-line for you @chenba .

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -82,8 +82,10 @@ export const PaymentForm = ({
   onChange: onChangeProp,
   submitNonce,
 }: BasePaymentFormProps) => {
-  const hasExistingCard =
-    customer && customer.last4 && customer.subscriptions.length > 0;
+  const isExistingStripeCustomer =
+    customer &&
+    customer.payment_provider === 'stripe' &&
```
chenba, line 87: There's a `PaymentProvider.isStripe` you could use here.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -135,8 +132,7 @@ export const PaymentConfirmation = ({
               Continue to download
             </a>
           </Localized>
-          {Provider.isPaypal(payment_provider) && <PaypalPaymentLegalBlurb />}
-          {Provider.isStripe(payment_provider) && <StripePaymentLegalBlurb />}
+          <PaymentLegalBlurb provider={payment_provider} />
```
chenba, line 135: 💯

packages/fxa-payments-server/src/components/PaymentLegalBlurb/index.tsx
```diff
@@ -91,4 +93,18 @@ export const PaymentLegalBlurb = () => (
...
+  provider: PaymentProvider.ProviderType;
+};
+
+export const PaymentLegalBlurb = ({ provider }: PaymentLegalBlurbProps) => {
+  return (
+    <>
+      {PaymentProvider.isPaypal(provider) && <PaypalPaymentLegalBlurb />}
+      {PaymentProvider.isStripe(provider) && <StripePaymentLegalBlurb />}
+      {PaymentProvider.isNotChosen(provider) && <DefaultPaymentLegalBlurb />}
+    </>
```
chenba, line 106: Tiny suggestion: if we group the `&&`s and put `||`s between them, then we can return early.  i.e.

```
  return (
    (PaymentProvider.isPaypal(provider) && <PaypalPaymentLegalBlurb />) ||
    (PaymentProvider.isStripe(provider) && <StripePaymentLegalBlurb />) ||
    (PaymentProvider.isNotChosen(provider) && <DefaultPaymentLegalBlurb />)
  );
```

packages/fxa-payments-server/src/lib/PaymentProvider.ts
```diff
@@ -2,10 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export function isStripe(provider: string | undefined) {
+export type ProviderType = 'paypal' | 'stripe' | 'not_chosen' | undefined;
```
chenba, line 5: If possible, leave the `undefined` out of this type.  I'd prefer to have the object properties to be optional (e.g. `paymentProvider?: ProviderType`) or function parameters to add undefined (e.g. `fn = (p: ProviderType | undefined | null | '') => {}`) if the function want to accept undefined.

packages/fxa-payments-server/src/lib/test-utils.tsx
chenba: Now that we are checking the 'payment_provider' value and not `last4`, so like you said, we show the new customer form, and you cannot submit that form unless you enter the valid values into the fields.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
chenba: `getElementsByTagName` takes a type arg so you can write: `    document.head.getElementsByTagName<'script'>('script')
` then everything in that returned collection is a `HTMLScriptElement`.

I don't have an alternative, but, IMHO as they used to say, I'd rather not have the story in storybook than adding code to accommodate it.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -502,7 +527,17 @@ async function handlePaymentIntent({
...
+  customer.payment_provider === 'stripe' &&
+  customer.subscriptions.length > 0;
+
+const isExistingPaypalCustomer = (customer: Customer | null) =>
+  customer &&
+  customer.payment_provider === 'paypal' &&
+  customer.subscriptions.length > 0;
+
+const isNewCustomer = (customer: Customer | null) =>
+  !isExistingStripeCustomer(customer) && !isExistingPaypalCustomer(customer);
```
chenba, line 541: The way this is written requires future devs to know to add to it when they add a third payment provider.  It'd be easier on them if it checks for the lack of any payment provider and not two specific payment providers.

chenba (approved):
👍 Looks good.  Got a few suggestions in the comments.

packages/fxa-payments-server/src/lib/test-utils.tsx
biancadanforth: Hm... true, but isn't that also true for returning Stripe customers? I.e. they must first check the `Checkbox` before they can submit the form?

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
biancadanforth: Hmm... that's a fair point. I thought of a much more minimal approach though where I just declare a global `isStorybookContext` in the Storybook head file and check for that. What do you think?

```diff
diff --git a/packages/fxa-payments-server/.storybook/preview-head.html b/packages/fxa-payments-server/.storybook/preview-head.html
index e0bb2403e..f041f156e 100644
--- a/packages/fxa-payments-server/.storybook/preview-head.html
+++ b/packages/fxa-payments-server/.storybook/preview-head.html
@@ -1,2 +1,5 @@
+<script>
+    const isStorybookContext = true;
+</script>
 <script src="https://js.stripe.com/v3/"></script>
 <script src = "https://www.paypal.com/sdk/js?client-id=sb&vault=true&commit=false&intent=authorize&disable-funding=credit,card"></script>
diff --git a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
index 111c6204d..1ddfc1487 100644
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
@@ -38,6 +38,8 @@ const PaypalButton = React.lazy(() => import('../../Product/PayPalButton'));
 type PaymentError = undefined | StripeError;
 type RetryStatus = undefined | { invoiceId: string };
 
+declare var isStorybookContext: boolean | undefined;
+
 export type SubscriptionCreateStripeAPIs = Pick<
   Stripe,
   'createPaymentMethod' | 'confirmCardPayment'
@@ -97,26 +99,12 @@ export const SubscriptionCreate = ({
   // customer can be null and customer.payment_provider can be undefined.
   const paymentProvider: ProviderType = customer?.payment_provider;
 
-  const isStorybookContext = Array.from(
-    document.head.getElementsByTagName('SCRIPT')
-  ).filter(
-    (ele) =>
-      // @ts-ignore: HTMLElement does not have property 'src'
-      ele.src ===
-      'https://www.paypal.com/sdk/js?client-id=sb&vault=true&commit=false&intent=authorize&disable-funding=credit,card'
-  ).length;
-
   useEffect(() => {
     if (!config.featureFlags.usePaypalUIByDefault) {
       return;
     }
 
-    if (paypalButtonBase) {
-      setPaypalScriptLoaded(true);
-      return;
-    }
-
-    if (isStorybookContext) {
+    if (paypalButtonBase || isStorybookContext) {
       setPaypalScriptLoaded(true);
       return;
     }
```

If you still think it's a bad idea, I can just revert the Storybook changes.

packages/fxa-payments-server/src/lib/PaymentProvider.ts
biancadanforth: Okay; I updated this using the former suggestion -- I am not clear on how I'd use the `fn = (p: ProviderType | undefined | null | '') => {}` part -- could you give an example for the future?

packages/fxa-payments-server/src/lib/test-utils.tsx
chenba: > 
> 
> Hm... true, but isn't that also true for returning Stripe customers? I.e. they must first check the `Checkbox` before they can submit the form?

Very true.  I don't have the answer off-hand.  I suspect it has something to do with the test setup.

packages/fxa-payments-server/src/lib/PaymentProvider.ts
chenba: That was just an example of a function that accepts `ProviderType` as well as some falsey/null-ish values.  The idea is that `ProviderType` is just `'paypal' | 'stripe' | 'not_chosen'` and we can widen the set of accepted value in the parameter declaration instead.

packages/fxa-payments-server/src/lib/test-utils.tsx
biancadanforth: Yeah I know there's a `confirm` prop that can be passed it that looks to bypasses disabling the submit button until all form elements have been filled in with valid data/completed; I just didn't see that it was being set to `true` anywhere.

Issue #7278, opened 2021-01-14
Handle IPN transction type for billing agreement cancel


data-sync-user:
When we get an IPN transaction type of ``mp_cancel``, we must:

* Mark the agreement as cancelled in our database.
* Remove the billing agreement from the Stripe customer metadata if the one in the metadata matches this one.
* If the current billing agreement matches the notification, then send an email to the customer notifying them that they should update their payment method.

https://developer.paypal.com/docs/api-basics/notifications/ipn/IPNandPDTVariables/#reference-transaction-and-billing-agreements-variables

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2908)
┆Issue Number: FXA-2908

data-sync-user:
➤ Se Yeon Kim commented:

[~bbangert@mozilla.com]  Should we consider storing the all billing agreement to look at historical invoice/transaction data? Or is the transaction ID available on the previous invoice enough?

PR #7725, opened 2021-03-04
feat(auth-server): handle IPN for billing agreement cancelled

say-yawn:
## Because

- When we receive a cancelled billing agreement notification we want to update the PayPalBillingAgreements and Stripe Customer

## This pull request

- Adds IPN message handler for billing agreement cancelled and respectively updates the PayPalBillingAgreements and Stripe Customer

Closes: #7278

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
@@ -11,6 +11,11 @@ import { StripeHelper } from '../../payments/stripe';
 import { reportSentryError } from '../../sentry';
 import { AuthLogger, AuthRequest } from '../../types';
 import { PayPalHandler } from './paypal';
+import {
+  getAccountCustomerByUid,
+  getPayPalBAByBAId,
+} from 'fxa-shared/db/models/auth';
+import { Stripe } from 'stripe';
 
 export class PayPalNotificationHandler extends PayPalHandler {
   /**
@@ -78,7 +83,19 @@ export class PayPalNotificationHandler extends PayPalHandler {
     }
   }
 
-  private async handleMpCancel(message: IpnMerchPmtType) {}
+  private async handleMpCancel(message: IpnMerchPmtType) {
+    const billingAgreement = await getPayPalBAByBAId(message.mp_id);
+    if (billingAgreement.status == 'Cancelled') {
+      return;
+    }
+    const accountCustomer = await getAccountCustomerByUid(billingAgreement.uid);
+    this.stripeHelper.removeCustomerPaypalAgreement(
+      accountCustomer.uid,
+      { id: accountCustomer.stripeCustomerId } as Stripe.Customer,
+      billingAgreement.billingAgreementId
+    );
+    // TODO: Send email to user that they must go to sub management and re-auth PayPal
+  }
 
   /**
    * Verify and dispatch IPN events from PayPal
--- a/packages/fxa-auth-server/test/local/routes/fixtures/mp_cancel_successful.json
+++ b/packages/fxa-auth-server/test/local/routes/fixtures/mp_cancel_successful.json
@@ -1,0 +1,20 @@
+{
+    "charset": "UTF-8",
+    "first_name": "John",
+    "ipn_track_id": "aa4335d8c3e27",
+    "last_name": "Doe",
+    "mp_currency": "USD",
+    "mp_custom": "",
+    "mp_desc": "Mozilla",
+    "mp_id": "B-75F71268HU787010S",
+    "mp_status": "1",
+    "notify_version": "3.9",
+    "payer_email": "syk3@personal.example.com",
+    "payer_id": "YRA75ZXLP83G8",
+    "payer_status": "verified",
+    "reason_code": "mp_2002",
+    "residence_country": "US",
+    "test_ipn": "1",
+    "txn_type": "mp_cancel",
+    "verify_sign": "AnneGm6iAtogaFpJq4cobwW29Mu4Ais1x8fLJdTdZDzGF8.oWAuk3SIY"
+}
--- a/packages/fxa-auth-server/test/local/routes/subscriptions/paypal-notifications.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions/paypal-notifications.js
@@ -14,10 +14,10 @@ const mocks = require('../../../mocks');
 const error = require('../../../../lib/error');
 const completedMerchantPaymentNotification = require('../fixtures/merch_pmt_completed.json');
 const pendingMerchantPaymentNotification = require('../fixtures/merch_pmt_pending.json');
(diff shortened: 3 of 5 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -78,7 +83,19 @@ export class PayPalNotificationHandler extends PayPalHandler {
...
-  private async handleMpCancel(message: IpnMerchPmtType) {}
+  private async handleMpCancel(message: IpnMerchPmtType) {
+    const billingAgreement = await getPayPalBAByBAId(message.mp_id);
+    if (billingAgreement.status == 'Cancelled') {
+      return;
+    }
+    const accountCustomer = await getAccountCustomerByUid(billingAgreement.uid);
+    this.stripeHelper.removeCustomerPaypalAgreement(
+      accountCustomer.uid,
+      { id: accountCustomer.stripeCustomerId } as Stripe.Customer,
```
bbangert, line 94: If `removeCustomerPaypalAgreement` only needs the customerId, we should change that to use it instead of having to type-cast structures like this to pass them in.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -78,7 +83,19 @@ export class PayPalNotificationHandler extends PayPalHandler {
...
+      billingAgreement.billingAgreementId
+    );
+    // TODO: Send email to user that they must go to sub management and re-auth PayPal
```
bbangert, line 97: We have the email handlers so we can do this now, right? We should also check here to see if they still have paypal in use before sending the email.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -78,7 +83,19 @@ export class PayPalNotificationHandler extends PayPalHandler {
     }
   }
 
-  private async handleMpCancel(message: IpnMerchPmtType) {}
+  private async handleMpCancel(message: IpnMerchPmtType) {
+    const billingAgreement = await getPayPalBAByBAId(message.mp_id);
+    if (billingAgreement.status == 'Cancelled') {
+      return;
+    }
+    const accountCustomer = await getAccountCustomerByUid(billingAgreement.uid);
```
bbangert, line 91: What happens if we don't find an agreement in our system?

say-yawn: That's a good point. I think we can just return and log that?

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
say-yawn: What do you mean by PayPal in use?

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
bbangert: Logging would be good, I could see this happening for example if we had already deleted a user before we get this event.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
bbangert: Lesley has a function that returns the payment provider, it does that by determining if the customer has any active/past_due subscriptions. If that indicates that the user has paypal on file, we need to send the email.

say-yawn: I am not understanding why we want to check if the customer's payment provider is PayPal again? `getPayPalBAByBAId` returns the billing agreement with the uid. With that uid we then find the account customer and the Stripe customer id. So does this not prove that the user Stripe customer with payment method set as PayPal? Can you provide an example on why we would want to check the payment provider before sending the email?

bbangert: Because if the user no longer has any subscriptions that need payment, it doesn't matter if they cancel their billing agreement. There's no reason to send them an email saying they'll lose their subscription if they don't put one on file.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -469,11 +470,11 @@ export class StripeHelper {
    */
   async removeCustomerPaypalAgreement(
     uid: string,
-    customer: Stripe.Customer,
+    customer_id: string,
```
bbangert, line 473: This should be `customerId`.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -11,6 +11,11 @@ import { StripeHelper } from '../../payments/stripe';
 import { reportSentryError } from '../../sentry';
 import { AuthLogger, AuthRequest } from '../../types';
 import { PayPalHandler } from './paypal';
+import {
+  getAccountCustomerByUid,
+  getPayPalBAByBAId,
+} from 'fxa-shared/db/models/auth';
```
bbangert, line 17: Using the TS Hero sort imports will put this in the right spot.

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -78,7 +83,54 @@ export class PayPalNotificationHandler extends PayPalHandler {
...
+      });
+      return;
+    }
+    this.stripeHelper.removeCustomerPaypalAgreement(
+      accountCustomer.uid,
+      customer.id,
+      billingAgreement.billingAgreementId
+    );
+
+    const nextPeriodValidSubscription = customer.subscriptions?.data.find(
```
bbangert, line 123: The subscription should be checked to see if its active as well.

bbangert (changes requested):

packages/fxa-auth-server/lib/routes/subscriptions/paypal-notifications.ts
```diff
@@ -78,7 +78,54 @@ export class PayPalNotificationHandler extends PayPalHandler {
...
+        message: 'Billing agreement not found',
+        ipnMessage: message,
+      });
+      return;
+    }
+    if (billingAgreement.status == 'Cancelled') {
+      return;
+    }
+    const accountCustomer = await getAccountCustomerByUid(billingAgreement.uid);
+    const customer = await this.stripeHelper.fetchCustomer(
```
bbangert, line 101: I see you couldn't use `customer` because it needs the email address. Unfortunately this function does not auto-expand anything, so the subscriptions are not guaranteed to be expanded. The additional expand argument should be provided here:
```
      'sources',
      'subscriptions',
      'invoice_settings.default_payment_method',
```

bbangert (approved):

Issue #7467, opened 2021-02-09
Error handling on front-end for payment/country mismatches


data-sync-user:
The payments frontend needs new error handling for errors regarding a users payment source funding not matching the chosen currency. The message should indicate that their payment source is not a valid country for the plan.

Error strings and different scenarios are documented here: [https://docs.google.com/document/d/1Eyl7vd87Y-nyUYryYgo8hxOkmvFDKpW3qP3n_YUbTVc/edit#heading=h.6jksqbfwxv40](https://docs.google.com/document/d/1Eyl7vd87Y-nyUYryYgo8hxOkmvFDKpW3qP3n_YUbTVc/edit#heading=h.6jksqbfwxv40)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3005)
┆Issue Number: FXA-3005

data-sync-user:
➤ Sarah Bird commented:

* Waiting for error strings from UX (should be available by Feb 26)
  * Implementer may need to wait for subscription management code to be implemented to implement all scenarios - not sure.

Issue #7726, opened 2021-03-04
Show error message for returning PayPal customers during Checkout


data-sync-user:
If we are descoping FXA-3078 for the initial PayPal launch, we should think about what to show/tell existing PayPal subscribers in place of FXA-2801 who would like to create another subscription (and also make sure it isn't possible for the user to get into some unsupported state).

Ideally this would involve as little work as possible, since the idea is to reduce scope for the initial launch.

[~jecook@mozilla.com], what should we show/say in the message? Do we want to link out anywhere? Should we repurpose an existing error screen?

Note: Even though this technically belongs to the Returning Checkout story (FXA-3078), I'm going to add this to the New Checkout story (FXA-3077) for now just to make sure we track it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3119)
┆Issue Number: FXA-3119

data-sync-user:
➤ Bianca Danforth commented:

[~jecook@mozilla.com], I am going to propose that we bootstrap off of the new error screen ( https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=523%3A4575 ) you designed for PayPal payment processing errors and say:

"Sorry. Currently, you can only sign up for one subscription at a time. Please check back soon."

This is what we will plan to implement unless you tell us otherwise in the next few days.

data-sync-user:
➤ Jessica Cook commented:

[~bdanforth@mozilla.com]
I agree with what you have proposed above that is essentially what we discussed in the check-in this week so I'm good with leveraging that screen.

The messaging you have makes sense. 👍

data-sync-user:
➤ Bianca Danforth commented:

Great thanks for your input. You're right we did talk about using that error screen in the last check-in; that's probably where I got that idea from, heh.

PR #7718, opened 2021-03-04
task(fxa-payments-server): add error page to checkout

meandavejustice:
## Because

- We need to be able to let the use know if there is an error with the paypal transaction

## This pull request

- Adds the error view to the checkout flow

## Issue that this pull request solves

- fixes #7617
- closes #7100

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [x] I have added necessary documentation (if appropriate).
- [x] I have verified that my changes render correctly in RTL (if appropriate).

## Screenshots (Optional)
error-desktop
error-mobile
error-tablet

## Other information (Optional)

Any other information that is important to this pull request.

cc/ @birdsarah per our error handling discussion

PR #7729, opened 2021-03-05
feat(payments-server): Display currency restrictions errors on front-end.

birdsarah:
## Because

- We need to display multi-currency errors to users

## This pull request

- Updates the getErrorMessage function to handle a variety of error message

Closes: #7467 

## Checklist

## Screenshots (Optional)

To test:

#### Pre req:
Make sure your server's running with:
a) paypal enabled and 
b) update auth-server config so that currenciesToCountries is:
```
    default: {
      USD: ['US', 'GB', 'NZ', 'MY', 'SG', 'CA', 'AS', 'GU', 'MP', 'PR', 'VI'],
      EUR: ['FR', 'DE'],
    },
```

#### Test 1 - When trying to subscribe a US PayPal account to a Euro plan.
Steps:
* Go to localhost:8080, subscribe to Euro plan
* Use a paypal account that's configured for the US as it's country

Screenshot from 2021-03-04 20-20-19

#### Test 2: When trying to subscribe a US credit card to a Euro Plan

Steps:
* Go to localhost:8080, subscribe to Euro plan
* Use normal stripe card 4242 4242 4242 4242 etc.

Alternatively, subscribe to a USD plan and use a, for example, french credit card (https://stripe.com/docs/testing#international-cards) 

Screenshot from 2021-03-04 20-19-28

#### Test 3: When trying to change a euro subscribed plan user who signed up with a french credit card to a USD credit card

Steps:
* Go to localhost:8080, subscribe to Euro plan
* Use a french stripe card to subscribe (https://stripe.com/docs/testing#international-cards) 
* Go to subscription management and try to change your card to standard US card 4242 ....

Screenshot from 2021-03-04 20-18-04

## Other information
I could not get the front-end into a state where I was returned the currency-currency-mismatch error that we throw in the back-end. Can you help me think through if I've actually got everything? cc @bbangert

### DIFF
--- a/packages/fxa-payments-server/src/lib/errors.test.tsx
+++ b/packages/fxa-payments-server/src/lib/errors.test.tsx
@@ -1,16 +1,28 @@
 import '@testing-library/jest-dom/extend-expect';
 import { getErrorMessage, BASIC_ERROR, PAYMENT_ERROR_1 } from './errors';
 
+it('returns the basic error text if undefined error type', () => {
+  expect(getErrorMessage(undefined)).toEqual(BASIC_ERROR);
+});
+
 it('returns the basic error text if not predefined error type', () => {
-  expect(getErrorMessage('NON_PREDEFINED_ERROR')).toEqual(BASIC_ERROR);
+  expect(getErrorMessage({ code: 'NON_PREDEFINED_ERROR' })).toEqual(
+    BASIC_ERROR
+  );
 });
 
 it('returns the payment error text for the correct error type', () => {
-  expect(getErrorMessage('approve_with_id')).toEqual(PAYMENT_ERROR_1);
+  expect(getErrorMessage({ code: 'approve_with_id' })).toEqual(PAYMENT_ERROR_1);
 });
 
 it('returns the payment error text for setup intent authentication failure', () => {
-  expect(getErrorMessage('setup_intent_authentication_failure')).toEqual(
+  expect(
+    getErrorMessage({ code: 'setup_intent_authentication_failure' })
+  ).toEqual(PAYMENT_ERROR_1);
+});
+
+it('returns payment error based on message if that is what is available in error map', () => {
+  expect(getErrorMessage({ code: 'test', message: 'approve_with_id' })).toEqual(
     PAYMENT_ERROR_1
   );
 });
--- a/packages/fxa-payments-server/src/lib/errors.ts
+++ b/packages/fxa-payments-server/src/lib/errors.ts
@@ -1,3 +1,11 @@
+import { StripeError } from '@stripe/stripe-js';
+
+export type GeneralError = {
+  code: string;
+  errno?: number;
+  message?: string;
+};
+
 // ref: fxa-auth-server/lib/error.js
 const AuthServerErrno = {
   UNKNOWN_SUBSCRIPTION_CUSTOMER: 176,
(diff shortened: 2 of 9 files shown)


birdsarah:
@bbangert I could not get the front-end into a state where I was returned the currency-currency-mismatch error that we throw in the back-end.  Can you help me think through if I've actually got everything?

packages/fxa-payments-server/src/lib/errors.test.tsx
```diff
@@ -1,16 +1,28 @@
...
 
+it('returns the basic error text if undefined error type', () => {
+  expect(getErrorMessage(undefined)).toEqual(BASIC_ERROR);
+});
+
 it('returns the basic error text if not predefined error type', () => {
-  expect(getErrorMessage('NON_PREDEFINED_ERROR')).toEqual(BASIC_ERROR);
+  expect(getErrorMessage({ code: 'NON_PREDEFINED_ERROR' })).toEqual(
+    BASIC_ERROR
+  );
```
biancadanforth, line 11: I know you didn't write this test, but I don't see this `'NON_PREDEFINED_ERROR'` used anywhere in the source code. How is it different from the `undefined` case above? Can we remove it?

packages/fxa-payments-server/src/lib/errors.ts
```diff
@@ -31,6 +44,10 @@ let errorMessageIndex: { [key: string]: string } = {
   coupon_expired: 'coupon-expired-error',
   card_error: 'card-error',
   // todo: handle "parameters_exclusive": "Your already subscribed to _product_"
+  // Currency errors
+  'Funding source country does not match plan currency.':
+    'country-currency-mismatch',
+  'Changing currencies is not permitted.': 'currency-currency-mismatch',
```
biancadanforth, line 50: Ah I finally figured out where these keys are coming from -- the auth server. It's a little odd having these new keys break the pattern, though I guess the other ones are almost all Stripe error codes.

packages/fxa-payments-server/src/lib/errors.ts
```diff
@@ -111,14 +128,24 @@ const paymentErrors2 = [
...
+  if (!error) {
+    return BASIC_ERROR;
+  }
+  let lookup = 'UNKNOWN';
+  // Handle case where we need to lookup by message (e.g. currency error)
+  if (error.message && errorToErrorMessageMap[error.message]) {
+    lookup = error.message;
+  } else if (error.code && errorToErrorMessageMap[error.code]) {
+    lookup = error.code;
+  }
```
biancadanforth, line 147: I guess we have to add this `message` property, since this check is on the server, and the `code` is an HTTP response code rather than a Stripe error code?

biancadanforth (commented):
Just poking around, since I am basing my patch off of this one for #7726 . I tried one of the cases randomly and it worked for me.

I understand now what you changed and why (relative to Dave's PR #7718 ), and it makes sense to me. I had one minor suggestion and a question or two in line.

packages/fxa-payments-server/src/lib/errors.test.tsx
birdsarah: It's different because `undefined` handles the case where nothing is passed to getErrorMessage. This handles the case where a mystery error with a code that the system isn't coded for is passed to getErrorMessage.

packages/fxa-payments-server/src/lib/errors.ts
birdsarah: Yes. And there may be paypal specific codes in the future. I agree this whole module feels janky, but I wasn't going to tackle refactoring it.

packages/fxa-payments-server/src/lib/errors.ts
birdsarah: I'm not sure what you mean bianca, can you elaborate? Or we can talk about it live.

meandavejustice (approved):
👍

birdsarah:
OK. Merging this. There is something wrong with upgrade flow in general. So we will worry about that in follow-on issue.

PR #7752, opened 2021-03-09
Fix use of MySQL cluster read-only nodes

enote-kane:
## Because

When running the fxa-auth-db-mysql service against a MySQL cluster with read-only secondaries, some functionality is broken.

The error thrown by the MySQL secondary is:
```
ERROR 1290 (HY000): The MySQL server is running with the --read-only option so it cannot execute this statement
```

...which apparently doesn't make it up into Sentry, but we can see the MySQL error:
```
ER_OPTION_PREVENTS_STATEMENT
```

One of the broken functionality is `verifyTokensWithMethod` and the corresponding MySQL procedure `verifyTokensWithMethod_3`, which internally wants to execute an `UPDATE`.

## This pull request

Makes use of the internal `write` function instead of `read` where needed.

Closes: #7753

## Checklist

* [MySQL server error reference](https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html#error_er_option_prevents_statement)

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/mysql.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mysql.js
@@ -1481,16 +1481,16 @@ module.exports = function (log, error) {
 
   const UPDATE_TOTP_TOKEN = 'CALL updateTotpToken_4(?, ?, ?)';
   MySql.prototype.updateTotpToken = function (uid, token) {
-    return this.read(UPDATE_TOTP_TOKEN, [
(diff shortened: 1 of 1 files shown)


dannycoates (approved):
Thank you @enote-kane :)

This is a bug that we never noticed because we don't run in a read-only secondary configuration. Although we don't explicitly support other configurations than our own this is a helpful fix. Thanks!

Issue #7689, opened 2021-03-02
Modify legal-md-to-pdf.sh and update CDN


data-sync-user:
Our "Payment received" emails link to the *product*'s ToS and Privacy Policy (meaning, if you bought the VPN, it will link to the VPN legal docs, not FxA's).

Those legal docs come from [the legal-docs repository](https://github.com/mozilla/legal-docs), are [generated one-at-a-time with a script](https://github.com/mozilla/fxa/blob/main/_scripts/legal-md-to-pdf.sh) and [saved in the repository](https://github.com/mozilla/fxa/tree/main/assets/legal) to eventually end up on our CDN.

We should optimize legal-md-to-pdf.sh script to traverse the whole directory,  convert VPN legal docs in each locale to PDF, and write them to assets/<locale>/<doc-name>.pdf.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3113)
┆Issue Number: FXA-3113

data-sync-user:
➤ Marco Mucci commented:

[~lnorton@mozilla.com] Should this issue be tagged as front-end or back-end?  Thanks.

data-sync-user:
➤ Lesley Norton commented:

[~mmucci@mozilla.com] that is a great question! I actually think the answer might be "neither" and that it falls more into something like an "ops" category.

data-sync-user:
➤ Marco Mucci commented:

[~lnorton@mozilla.com] Thanks for the clarification.

Issue #7690, opened 2021-03-02
Update legal links in transactional emails


data-sync-user:
Modify the emails to point to wherever you're doing the language detection and serving up the pdfs.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3114)
┆Issue Number: FXA-3114

data-sync-user:
➤ Marco Mucci commented:

[~lnorton@mozilla.com] Should this issue be tagged as front-end or back-end?  Thanks.

Issue #7688, opened 2021-03-02
Add new endpoint that redirects users to the correct legal PDF based on their locale


data-sync-user:
We need to link users to localized legal docs from the transactional emails. These legal docs are PDFs that are hosted in the CDN. We need to make a new endpoint that determines the user's locale and redirects them to the appropriate PDF.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3112)
┆Issue Number: FXA-3112

data-sync-user:
➤ Barry Chen commented:

"legal docs" here mean the "Terms and cancellation policy" (e.g. https://accounts-static.cdn.mozilla.net/legal/mozilla_vpn_privacy_notice/en-US.pdf) and "Privacy" (same pdf) in the subscription email footer, correct?  umm [~wclouser@mozilla.com]

data-sync-user:
➤ Lesley Norton commented:

That's correct. Right now they are only available in English but we will have French and German translations available as of Thursday and additional locales in the future. I think the idea is to have a single endpoint that determines the user's locale and redirects them to the appropriate PDF.

data-sync-user:
➤ Barry Chen commented:

Currently the URLs for those docs are driven by the Stripe product and plan metadata.  I think we can continue to do that.

I'll verify how well that works now and make some updates if necessary.

data-sync-user:
➤ Marco Mucci commented:

[~lnorton@mozilla.com] Should this issue be tagged as front-end or back-end?  Thanks.

data-sync-user:
➤ Barry Chen commented:

[~lnorton@mozilla.com] who originally suggested the new endpoint?  I want to be sure that I'm not missing something with this metadata approach.

data-sync-user:
➤ Barry Chen commented:

After discussing this issue with Wil, we've decided to take the endpoint approach, as it requires less manual (meta)data entry from people.

PR #7768, opened 2021-03-09
Update legal-md-to-pdf.sh script and add resulting localized PDFs.

lesleyjanenorton:
## Because

- We want to efficiently convert localized Mozilla VPN legal documents from markdown documents into PDFs.

## This pull request

- Updates `legal-md-to-pdf.sh` so that it:
  - takes the `path/to/where/legal-docs/live-locally` in as a command line argument, instead of prompting for it
  - traverses the legal-docs directory looking for `mozilla_vpn_privacy_notice.md` and `mozilla_vpn_tos.md` in each locale. When found, it converts these into PDFs and saves them according to the new legal-docs directory structure `(../assets/legal/<locale>/<doc-name>.pdf`.
 - Adds newly localized PDFs according to the new directory structure.
- Does not remove previously generated legal PDFs since the Mozilla VPN transactional emails still point to them. Those can be removed after #7688 and #7690 are closed.

Closes: #7689 

## Checklist

## Screenshots (Optional)
n/a

n/a


meandavejustice (approved):
Works great, really nice to have the script cleaned up too

Issue #7767, opened 2021-03-09
L10N failing for specific git commit


enote-kane:
## Description

When trying to specify a Git commit of the L10N repo to reproduce a build, the checked out workspace is still at its default branch while the script `_scripts/clone-l10n.sh` exits with error code `128`.

### Steps to reproduce

Execute the following at the top of the repo clone directory:

```bash
FXA_L10N_SHA=fa77dc93fba361f201ba16d4d7072f8e778b63c2 bash _scripts/clone-l10n.sh fxa-settings
```

### Expected result

The L10N repo workspace gets checked out at the wanted commit.

### Actual result

The following happens:
```
Downloading L10N files from https://github.com/mozilla/fxa-content-server-l10n.git...
Cloning into 'fxa-content-server-l10n'...
remote: Enumerating objects: 536, done.
remote: Counting objects: 100% (536/536), done.
remote: Compressing objects: 100% (357/357), done.
remote: Total 536 (delta 254), reused 315 (delta 141), pack-reused 0
Receiving objects: 100% (536/536), 2.52 MiB | 6.30 MiB/s, done.
Resolving deltas: 100% (254/254), done.
Updating L10N files
fatal: reference is not a tree: fa77dc93fba361f201ba16d4d7072f8e778b63c2
```

### Environment

_not applicable_

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3159)
┆Issue Number: FXA-3159

PR #7769, opened 2021-03-09
Enhance L10N handling

enote-kane:
## Because

The current script has quite some functional errors and shortcomings.

## This pull request

...fixes some functional issues, introduces new functionality useful for testing/debugging.

More specifically:
* fix checkout to a specific commit - with existing environment variable `FXA_L10N_SHA`
* fix script errors/warnings as reported by `shellcheck` and unify code style

New features:
* specify a different branch - env var `FXA_L10N_BRANCH`
* override Git repository URL (e.g. a fork for testing) - env var `FXA_L10N_REPO`
* support to just use a local directory as-is (existing clone) - env var `FXA_L10N_COPY_FROM`
* override directory into which the content is cloned/linked into - env var `FXA_L10N_DOWNLOAD_PATH`

Closes: #7767 

## Checklist

Output of `shellcheck -ax` of the existing script:
```

In _scripts/clone-l10n.sh line 9:
  echo "\nYou need specify a module.\n"
       ^-- SC2028: echo may not expand escape sequences. Use printf.


In _scripts/clone-l10n.sh line 23:
  echo "\nThe module/package does not appear to exist.\n"
       ^-- SC2028: echo may not expand escape sequences. Use printf.


In _scripts/clone-l10n.sh line 49:
    "$PAYMENT") FTL_FILENAME="main" ;;
     ^------^ SC2153: Possible misspelling: PAYMENT may not be assigned, but PAYMENTS is.


In _scripts/clone-l10n.sh line 54:
  for src in locale/**/$FTL_FILENAME.ftl; do
                       ^-----------^ SC2231: Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .

For more information:
  https://www.shellcheck.net/wiki/SC2153 -- Possible misspelling: PAYMENT may...
  https://www.shellcheck.net/wiki/SC2028 -- echo may not expand escape sequen...
  https://www.shellcheck.net/wiki/SC2231 -- Quote expansions in this for loop...
```

New output using a specific commit:

```bash
$ FXA_L10N_SHA=fa77dc93fba361f201ba16d4d7072f8e778b63c2 bash _scripts/clone-l10n.sh fxa-payments-server
[clone-l10n] Downloading L10N files from: https://github.com/mozilla/fxa-content-server-l10n.git
Cloning into 'fxa-content-server-l10n'...
remote: Enumerating objects: 16149, done.
remote: Counting objects: 100% (16149/16149), done.
remote: Compressing objects: 100% (6037/6037), done.
remote: Total 16149 (delta 9498), reused 15599 (delta 9004), pack-reused 0
Receiving objects: 100% (16149/16149), 47.59 MiB | 6.64 MiB/s, done.
Resolving deltas: 100% (9498/9498), done.
[clone-l10n] Updating L10N files in: /srv/dev/mozilla/fxa/packages/fxa-payments-server/fxa-content-server-l10n
```

...or when using an existing local clone instead (fastest, of course):
```bash
$ FXA_L10N_COPY_FROM=/srv/dev/mozilla/fxa-l10n/ bash _scripts/clone-l10n.sh fxa-auth-server
Using L10N files from: /srv/dev/mozilla/fxa-l10n/

$ ls -l packages/fxa-auth-server/fxa-content-server-l10n
lrwxrwxrwx 1 kane kane 28 Mar  9 22:30 packages/fxa-auth-server/fxa-content-server-l10n -> /srv/dev/mozilla/fxa-l10n/
```

### DIFF
--- a/_scripts/clone-l10n.sh
+++ b/_scripts/clone-l10n.sh
@@ -1,64 +1,142 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
+# ensure the script errors out when a command fails
 set -e
 
-MODULE=$1
+MODULE="$1"
 
-if [ -z "$MODULE" ]; then
-  echo "\nYou need specify a module.\n"
-  exit 0
-fi
+log() {
+    echo "[clone-l10n] $1"
+}
+
+error() {
+    log "$1" >&2
+    exit 1
+}
+
+[ -n "$MODULE" ] || error "You need to specify a module."
 
 if [ -n "$FXA_L10N_SKIP" ]; then
-    echo "Skipping fxa-content-server-l10n update..."
+    log "Skipping fxa-content-server-l10n update..."
     exit 0
 fi
 
-DOWNLOAD_PATH="https://github.com/mozilla/fxa-content-server-l10n.git"
-SCRIPT_DIR=$(dirname "$0")
-MODULE_PATH="$SCRIPT_DIR/../packages/$MODULE"
 
-if [ ! -d "$MODULE_PATH" ]; then
-  echo "\nThe module/package does not appear to exist.\n"
-fi
+# BEGIN: workaround for missing GNU realpath in BSD/MacOS-X
+backup_ifs="$IFS"
 
-if [ -z "$FXA_L10N_SHA" ]; then
-    FXA_L10N_SHA="master"
-fi
+get_realpath_missing() {
+    if hash -- realpath 2>/dev/null; then
+        realpath -m "$1"
+        return
+    fi
+
+    local final_path is_first
+    final_path="$(pwd -P)"
+
+    IFS=/
+    is_first=true
+    for path_part in $1; do
+        if [ -z "${path_part}" ]; then
+            if $is_first; then
+                # absolute path
+                final_path=""
+            fi # otherwise, just ignore (e.g. "//")
+            continue
+        fi
+        is_first=false
+
+        if [ "${path_part}" = ".." ]; then
+            # one up
+            final_path="$(dirname "${final_path}")"
+        elif [ "${path_part}" != "." ]; then
+            final_path="${final_path}/${path_part}"
+        fi # otherwise ignore - current directory
+    done
+    IFS="${backup_ifs}"
+
+    echo "${final_path}"
+}
+# END: workaround for missing GNU realpath in BSD/MacOS-X
+
+DOWNLOAD_PATH="${FXA_L10N_DOWNLOAD_PATH:-fxa-content-server-l10n}"
+SCRIPT_DIR="$(dirname "$0")"
+MODULE_PATH="$(get_realpath_missing "$SCRIPT_DIR/../packages/$MODULE")"
+
+[ -d "$MODULE_PATH" ] || error "The module/package does not exist at: $MODULE_PATH"
 
 # Clone and pull
 cd "$MODULE_PATH"
-if [ ! -d "fxa-content-server-l10n" ]; then
-	echo "Downloading L10N files from $DOWNLOAD_PATH..."
-	git clone --depth 1 $DOWNLOAD_PATH
+
+if [ -n "${FXA_L10N_COPY_FROM}" ]; then
+    echo "Using L10N files from: ${FXA_L10N_COPY_FROM}"
+    [ ! -d "${DOWNLOAD_PATH}" ] || rm -rf "${DOWNLOAD_PATH}"
+    ln -sf "${FXA_L10N_COPY_FROM}" "${DOWNLOAD_PATH}"
+else
+    # Git repository to clone
+    FXA_L10N_REPO="${FXA_L10N_REPO:-https://github.com/mozilla/fxa-content-server-l10n.git}"
+
+    # branch name
+    FXA_L10N_BRANCH="${FXA_L10N_BRANCH:-master}"
+
+    # number of commits to truncate the history for specific commit checkout
+    FXA_L10N_DEPTH="${FXA_L10N_DEPTH:-1000}"
+
+    clone_opts=(--depth 1)
+
+    # fetch some history when required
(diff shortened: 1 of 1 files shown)


_scripts/clone-l10n.sh
```diff
@@ -1,64 +1,104 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
clouserw, line 1: @jbuck is bash available on all our deploy boxes?

jbuck: Only need it when building the docker image, so yes

_scripts/clone-l10n.sh
```diff
@@ -1,64 +1,104 @@
...
-    dir=$(dirname "$src")
-    dir=$(echo "$dir" | sed "s/_/-/")
-    dir=$(basename "$dir")
-    base=$(basename "$src")
-    mkdir -p "../public/locales/$dir"
-    cp "$src" "../public/locales/$dir/$base"
-  done
-fi
-
-cd "$INIT_CWD"
```
clouserw, line 64: It feels sloppy to have this at the end of the script (especially with no check for an empty variable(!)) but I suspect it's there for a reason.  It looks like the `husky` package uses INIT_CWD just from grepping, but I don't know if that is related.

clouserw (approved):
Thanks for this patch.  I like these changes!  I'm going to mark it requesting review from @mozilla/fxa-devs and @jbuck but a first pass through by me is r+.  I want whoever is responsible for the train that sprint to look at this patch and be comfortable with it because if anything goes awry they'll need to be aware and prepared to make a patch.

clouserw:
@chenba might have some input as he made changes to that script last

_scripts/clone-l10n.sh
enote-kane: This is not how shells work.

Whenever you invoke a shell script it is executed in its own process, which means that you are not modifying the invoking shell, including the working directory.

A test:
```bash
kane@kane:~$ cat > cd-test.sh <<EOF
#!/usr/bin/env bash

[ -d test ] || mkdir test
cd test
echo "pwd = $(pwd)"
EOF
kane@kane:~$ chmod +x cd-test.sh 
kane@kane:~$ pwd
/home/kane
kane@kane:~$ ./cd-test.sh
pwd = /home/kane/test
kane@kane:~$ pwd
/home/kane
```
So no, that final `cd` is absolutely useless.

clouserw: There are many different ways to run bash scripts and this is part of a larger pipeline.  I don't know how this is called, so I'm pointing out that it might have been used in a larger context that would break if it's removed.  It's something for us to check for or watch for when the patch lands.

chenba:
> 
> 
> @chenba might have some input as he made changes to that script last

sorry, missed this ping.  :(

_scripts/clone-l10n.sh
```diff
@@ -1,64 +1,104 @@
...
+    log "Skipping fxa-content-server-l10n update..."
     exit 0
 fi
 
-DOWNLOAD_PATH="https://github.com/mozilla/fxa-content-server-l10n.git"
-SCRIPT_DIR=$(dirname "$0")
-MODULE_PATH="$SCRIPT_DIR/../packages/$MODULE"
+DOWNLOAD_PATH="${FXA_L10N_DOWNLOAD_PATH:-fxa-content-server-l10n}"
+SCRIPT_DIR="$(dirname "$0")"
+MODULE_PATH="$(realpath -m "$SCRIPT_DIR/../packages/$MODULE")"
```
chenba, line 26: I think this could be an issue for macOS users since I believe `realpath` isn't available there.

_scripts/clone-l10n.sh
chenba: I think it was just sloppiness on my part.

chenba:
It would be good to have a macOS user try running this script.

_scripts/clone-l10n.sh
enote-kane: AFAICS and as pointed out by @jbuck this script is only used in the context of Docker builds, which means the environment the script runs in is the same and depends on the Docker base image, not the executing host.

But if necessary, I could fall back to BSD `readlink` in case the GNU `realpath` is not available.

_scripts/clone-l10n.sh
enote-kane: @clouserw This behavior is guaranteed by [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_12):
> A subshell environment shall be created as a duplicate of the shell environment, except that signal traps that are not being ignored shall be set to the default action. Changes made to the subshell environment shall not affect the shell environment.

The only exceptions here are [`exec`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#exec) and [`source`/`.`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot), where `exec` will _replace_ the current execution environment and `source` executes all commands in a script _inside_ the current execution environment (and thus, will change the invokers environment).

I hope that clarifies things.

_scripts/clone-l10n.sh
enote-kane: Fixed, we are now checking for the availability of `realpath` and if not available, fall back to a simple path-building loop (without resolving symlinks - as that is not needed for functionality).

enote-kane:
@chenba @jbuck Updated the script for better interoperability, please have another look.

chenba (approved):
WFM!  Thank you.  (I'll tag you in next time I need to write shell scripts. 😁)

enote-kane:
@chenba @jbuck Anyone going to merge it?

chenba:
@mozilla/fxa-devs Can anyone with macOS please give this a try and give a 👍 ?

vbudhram (approved):
@chenba Works as expected on Mac 👍🏽

chenba:
@enote-kane Thanks for the patch!

PR #7771, opened 2021-03-10
feat(payments): Add Checkout error screen for returning PayPal customers

biancadanforth:
## Because

- We descoped the [Returning Checkout story](https://jira.mozilla.com/browse/FXA-3078) for PayPal customers for the initial launch, and we want a descriptive placeholder screen for the small number of users who may be affected in the meantime.

## This pull request

- Shows `PaymentErrorView` in Checkout for returning PayPal customers with a descriptive message.
- Updates `PaymentErrorView` to show a different button for this error with the label "Manage my subscription" instead of "Try again" that takes the user to Subscription Management instead of SubscriptionCreate on click (the latter would have effectively been a noop).
- Localizes `PaymentErrorView` default "Try again" button label.
- Adds `TermsAndPrivacy` to `PaymentErrorView` for consistency with the rest of our Checkout screens.

Closes: #7726 

## Checklist

## Screenshots (Optional)
Screenshot_2021-03-09 Firefox Accounts(2)

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentErrorView/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentErrorView/index.test.tsx
@@ -8,6 +8,13 @@ import {
 
 import { PaymentErrorView } from './index';
 
+const mockHistoryPush = jest.fn();
+jest.mock('react-router-dom', () => ({
+  useHistory: () => ({
+    push: mockHistoryPush,
+  }),
+}));
+
 afterEach(cleanup);
 describe('PaymentErrorView test with l10n', () => {
   const bundle = setupFluentLocalizationTest('en-US');
@@ -29,6 +36,12 @@ describe('PaymentErrorView test with l10n', () => {
       'An unexpected error has occured while processing your payment, please try again.';
     const actual = getLocalizedMessage(bundle, 'payment-error-3', {});
     expect(actual).toEqual(expected);
+
+    const retryButton = queryByTestId('retry-link');
+    expect(retryButton).toBeInTheDocument();
+
+    const termsAndPrivacy = queryByTestId('terms');
+    expect(termsAndPrivacy).toBeInTheDocument();
   });
 
   it('calls passed onRetry function when retry button clicked', async () => {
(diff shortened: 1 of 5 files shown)


packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
```diff
@@ -23,6 +21,34 @@ export const PaymentErrorView = ({
...
+
+  // We want the button label and onClick handler to be different depending
+  // on the type of error
+  const ActionButton = () => {
+    return error?.code === 'returning_paypal_customer_error' ? (
+      <Localized id="payment-error-manage-subscription-button">
+        <button
+          data-testid="manage-subscription-link"
+          className="button"
+          onClick={() => history.push('/subscriptions')}
```
biancadanforth, line 34: I used `history.push` here, though I don't see the `useHistory` hook used elsewhere. Is there a better/preferred way to route to the `'/subscriptions'` relative URL?

packages/fxa-payments-server/src/components/PaymentErrorView/index.test.tsx
```diff
@@ -46,4 +58,49 @@ describe('PaymentErrorView test with l10n', () => {
...
+        onRetry={onRetry}
+        error={{ code: 'returning_paypal_customer_error' }}
+      />
+    );
+
+    await act(async () => {
+      fireEvent.click(getByTestId('manage-subscription-link'));
+    });
+
+    expect(onRetry).not.toHaveBeenCalled();
```
biancadanforth, line 104: This assertion that `onRetry` isn't called here is suboptimal. What I'd really like to do is assert that a different `onClick` handler was called; i.e. that `history.push("/subscriptions")` happened. Is there a better assertion I can use here?

biancadanforth (commented):
@meandavejustice I had a couple of specific questions here that I ask in-line. It'd be great to get your thoughts on these.

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
meandavejustice: We could also use the [Link Component](https://reactrouter.com/web/api/Link)

packages/fxa-payments-server/src/components/PaymentErrorView/index.test.tsx
meandavejustice: I think that checking the history would be better also. I see that you are mocking useHistory above. According to [this stack overflow answer](https://stackoverflow.com/a/59451956) You should be able to check the mocked history in the test.

meandavejustice (approved):
Works for me 👍

I left some comments on the `useHistory` hook question, and the testing of `history.push`. The coverage does test that the correct behavior is happening. Even if it's not in the most ideal way. I would timebox mocking out the `history.push` test.

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
biancadanforth: Thanks for the suggestion. Hmm... I don't see that used anywhere else either. If I wrap the `<button>` in a `<Link>` to achieve this, the styles break, so if it's all the same to you, I'll just keep the `useHistory` hook.

packages/fxa-payments-server/src/components/PaymentErrorView/index.test.tsx
biancadanforth: Good suggestion thanks; I found this at the time, but for whatever reason, I couldn't get it to work. Now it works though. Yay a good night's sleep?

PR #7514, opened 2021-02-12
fix(settings): ensure modal callbacks adhere to the defined type

dannycoates:
The `Modal` and `ModalVerifySession` components were incidentally calling the `onDismiss` callback with the dom event argument from their originating element. This happened to break the `ConnectedServices` component which was using a fn that accepted optional parameters as the callback.

This commit ensures the callbacks are called with no arguments.

fixes #7509

PR #7777, opened 2021-03-10
task(fxa-payments-server):fix onretry typedef

meandavejustice:
- https://github.com/mozilla/fxa/pull/7718#discussion_r591622870

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
@@ -13,7 +13,7 @@ type GeneralError = {
 };
 
 export type PaymentErrorViewProps = {
-  onRetry: Function;
+  onRetry: () => void;
   error?: StripeError | GeneralError;
   className?: string;
 };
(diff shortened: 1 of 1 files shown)


chenba (approved):
Thanks!

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
```diff
@@ -46,7 +46,7 @@ export const PaymentErrorView = ({
             <button
               data-testid="retry-link"
               className="button retry-link"
-              onClick={() => onRetry()}
+              onClick={onRetry}
```
dannycoates, line 49: it's safer to leave this wrapped otherwise the event argument from onClick will get passed to onRetry. it _usually_ won't matter but can lead to subtle bugs like it did here https://github.com/mozilla/fxa/pull/7514

chenba: That's an interesting tip.  Thanks!

meandavejustice: @chenba Should I hold off on merging then?

chenba: Either way is fine with me.  Easier to just not merge and close?

PR #7780, opened 2021-03-10
Default beta

dannycoates:
Since quite a number of functional tests use old settings as part of setup and refactoring them will take some time this option uses the NEW_SETTINGS env to redirect /settings to /beta/settings. tests will use NEW_SETTINGS=false while prod will use NEW_SETTINGS=true.

A side-effect of this is that the url in the browser will show beta/settings instead of just /settings until we fully complete the changes to refactor the tests and remove the unused code.


packages/fxa-content-server/tests/functional/lib/helpers.js
```diff
@@ -2824,7 +2824,7 @@ const createUserAndLoadSettings = thenify(function (email) {
     .then(createUser(email, PASSWORD, { preVerified: true }))
     .then(openPage(ENTER_EMAIL_URL, selectors.ENTER_EMAIL.HEADER))
     .then(fillOutEmailFirstSignIn(email, PASSWORD))
-    .then(testElementExists(selectors.SETTINGS.HEADER));
+    .then(testElementExists(selectors.SETTINGS.APP));
```
dannycoates, line 2827: since this is used by both old and new settings tests we need a common selector

6a68 (approved):
great hack, LGTM.

One minor thing, I'm not sure how to get to this path in a 123done flow, but `createRecoveryKey` in `views/ready.js` I think needs to change the path from `settings/account_recovery/confirm_password` to just `settings/account_recovery`

PR #7786, opened 2021-03-11
feat(subscriptions): redirect to localized legal doc pdf

chenba:
Because:
 - we want customers to download legal documents in the language they
   want when the translation is available.

This commit:
 - add a new endpoint on Payments that will take a url in its query
   params and redirects appropriately.

Closes: #7688


packages/fxa-payments-server/package.json
```diff
@@ -101,11 +107,7 @@
     "@sentry/node": "^6.0.1",
     "@stripe/react-stripe-js": "^1.2.0",
     "@stripe/stripe-js": "^1.13.1",
-    "@types/nock": "^11.1.0",
-    "@types/react-document-title": "^2.0.3",
-    "@types/react-stripe-elements": "^1.3.5",
-    "@types/react-transition-group": "^4.2.3",
-    "@types/superagent": "^4.1.10",
```
chenba, line 108: @types are just dev dependencies

chenba:
The format of the doc URLs was discussed on Slack, but the script that generates them hasn't been update at this time.

packages/fxa-payments-server/package.json
chenba: err didn't meant to @ anyone!

chenba:
Not sure who should review...

@lesleyjanenorton since you have a follow-up issue to this
@bbangert because it's server side even though it's not API backend stuff?
@biancadanforth just because 😀

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,156 @@
...
+const cdnFqdn = config.get('legalDocLinks').cdnFqdn;
+const acceptLanuageParser = require('accept-language-parser');
+const supportedLanguages = require('fxa-shared/l10n/supportedLanguages.json');
+const DEFAULT_LOCALE = 'en';
+const CACHE_DURATION = 24 * 60 * 60; // a day in seconds
+const LANGUAGES_LIMIT = 3;
+
+// A simple dictionary for caching.  The key is a url and the value is an object with two properties:
+//  - checkedOn: timestamp (in ms) when the last check against the url was performed
+//  - result: boolean
```
chenba, line 20: I'll make a change to limit the size of this.

chenba:
I set it back to draft so I can make a change on the caching part.

chenba:
This is ready for review again.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+  async process(req, res) {
+    const docUrl = new URL(req.query.url);
+
+    // Limit the redirect to just the CDN that's hosting the legal docs
+    if (docUrl.hostname !== cdnFqdn) {
+      return res.sendStatus(400).end();
+    }
+
+    // For backwards compatibility and future single-locale products(?)
+    if (docUrl.pathname.endsWith('.pdf')) {
```
lesleyjanenorton, line 105: +1 - It's feasible that someone could open an old email and try the link

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+      scope.setContext('subscription.legalDocs', {
+        url: docUrl.href,
+        acceptLanguage: req.get('accept-language'),
+      });
+      Sentry.captureMessage(
+        `Legal doc redirect failed.`,
+        Sentry.Severity.Warning
+      );
+    });
+    return res.sendStatus(404).end();
```
lesleyjanenorton, line 179: Does it maybe make sense, if all else fails, to redirect to mozilla.org/privacy or mozilla.org/legal?

packages/fxa-payments-server/server/lib/routes/legal-docs.test.js
```diff
@@ -0,0 +1,201 @@
...
+
+describe('lib/routes/legal-docs', () => {
+  beforeEach(() => {
+    mockFetch.mockClear();
+    mockRes.sendStatus.mockClear();
+    mockRes.redirect.mockClear();
+  });
+
+  it('responds with a HTTP 400 when the domain is not allowed', async () => {
+    const req = { query: { url: 'https://nope.example.com/vpn' } };
```
lesleyjanenorton, line 38: `nope.example.com` 😂

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+  checkAndShrinkCache(cache, resultsCacheLimit);
+
+  try {
+    const res = await fetch(url, { method: 'HEAD' });
+    const entry = { checkedOn: Date.now(), result: res.status === 200 };
+    cache.set(url, entry);
+    return entry['result'];
+  } catch {
+    // NOOP
+    // fetch API exceptions are not 3xx-5xx responses
```
biancadanforth, line 84: Would we want to log these errors e.g. to Sentry?

packages/fxa-payments-server/server/lib/routes/legal-docs.js
biancadanforth, line 105: How does this compare to the new form of `docUrl`?

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+'use strict';
+
+const config = require('../../config');
+const joi = require('joi');
+const fetch = require('node-fetch');
+const Sentry = require('@sentry/node');
+const { cdnFqdn, resultsCacheLimit } = config.get('legalDocLinks');
+const acceptLanuageParser = require('accept-language-parser');
+const supportedLanguages = require('fxa-shared/l10n/supportedLanguages.json');
+const DEFAULT_LOCALE = 'en';
```
biancadanforth, line 14: The GitHub issue says the default locale should be `'en-US'`. Does it make a difference here?

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+    const defaultDocUrl = new URL(
+      getFullPath(docUrl.pathname, DEFAULT_LOCALE),
+      docUrl.origin
+    );
+    if (await isValidDocUrl(defaultDocUrl.href)) {
+      defaultDocUrl.search = docUrl.search;
+      return res.redirect(defaultDocUrl.href).end();
+    }
+
+    // :(
```
biancadanforth, line 168: :(

packages/fxa-payments-server/server/config/index.js
```diff
@@ -127,6 +127,19 @@ const conf = convict({
       env: 'PAYMENT_TERMS_OF_SERVCIE',
```
biancadanforth, line 127: I noticed there is a typo here -- can we "just" fix it as part of this PR?

biancadanforth (commented):
Hey Barry; thanks for this PR. I checkout your feature branch locally, but I am not actually sure what the endpoint is supposed to be to try this out -- specifically the value of the `url` query parameter. I've got `http://localhost:3031/legal-docs?url=` so far... Can you help clarify? I've only managed to get 4xx errors so far.

Other than that, it seems like you have thought of a lot of edge cases. I had some questions/suggestions around performance -- particularly for caching and the cache size.

lesleyjanenorton (approved):

packages/fxa-payments-server/server/config/index.js
chenba: I cannot say for certain without checking.  I guess that's a no.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
chenba: The difference is that the legal docs repo doesn't have an en-US.  I changed this from 'en-US' after noticing that in the legal docs repo.  See https://github.com/mozilla/legal-docs/

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * A map for caching.  The key is a url and the value is an object with two properties:
+ *  - checkedOn: timestamp (in ms) when the last check against the url was performed
+ *  - result: boolean
+ *
+ * The size of this map will also be limited.  There are 71 supported languages
+ * and two docs per product (not price/plan).  With some generous rounding, we
+ * have 200 per product.  The current default is set to 65536, which should
+ * give us a decent runway.
```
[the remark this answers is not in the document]
chenba: This is not pre-allocating; just setting a max.  There isn't a truly good way to arrive at a value.  I picked something that seem "big enough" for years to come.  The actual number is a combination of the number of products, the number of locales those products support, the customers' requested languages and their priority, and the number of customers actually clicking the links in the emails.  Or someone just send junk and fill it up.

> does it make a difference if it's a power of 2 or not?

Nah.  It was just for style points.  😀

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * @param {obj} lang A language object from accept-language-parser
+ */
+const getLanguageTag = (lang) =>
+  lang.region ? `${lang.code}-${lang.region}` : lang.code;
+
+const isSupportedLanguage = (tag) => supportedLanguages.includes(tag);
+
+const isCacheExpired = (lastChecked) =>
+  Math.round((Date.now() - lastChecked) / 1000) > CACHE_DURATION;
```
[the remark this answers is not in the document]
chenba: Yep. 👍

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+  const cachedVal = cache.get(url);
+  if (cachedVal !== undefined) {
+    if (isCacheExpired(cachedVal['checkedOn'])) {
+      cache.delete(url);
+    } else {
+      return cachedVal['result'];
+    }
+  }
+
+  checkAndShrinkCache(cache, resultsCacheLimit);
```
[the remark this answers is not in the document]
chenba: hmm, I suppose so.  Since I stated above that the number of entries in the cache is not an exact science.  This prevents the number of entries ever going above the max.  Once per request would temporarily allow the number of entries to go above the max.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
chenba: Not sure how helpful it is to log network errors here.  We don't check Sentry that often.  And if a network error persists for days, then I think we'd have a bigger problem than this endpoint.

I'm neutral on this.  If you feel  strongly about it, I can add it.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
chenba: The new pattern is from https://github.com/mozilla/fxa/blob/main/_scripts/legal-md-to-pdf.sh#L107

packages/fxa-payments-server/server/lib/routes/legal-docs.js
chenba: I don't know.  Feels like a product and/or legal question.  Good idea though.

@clouserw What do  you think?

packages/fxa-payments-server/server/lib/routes/legal-docs.js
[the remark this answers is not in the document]
clouserw: If this fills up do we just not have caching or does it fail in some other way?  No caching is fine, failing silently would be bad.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
clouserw: Our fallback should be English.  Beyond that, (eg. if English doesn't exist) I don't think we need more redirects, but it would be nice to alert Sentry so we could see the error.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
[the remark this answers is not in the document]
chenba: When it reaches the limit, the oldest half of the entries are flushed out.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
chenba: Yeah, the Sentry reporting code is just above the line in question.  It captures the requested url (from the metadata when it's legit) and the value of accept-language.

chenba:
@biancadanforth I've made the couple of cache relate changes you suggested.  Thanks!

packages/fxa-payments-server/server/lib/routes/legal-docs.js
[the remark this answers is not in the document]
biancadanforth: Ah if the Map is at max when we get a new request, then we'd have max + 1 briefly, I guess? Yeah; I don't think it's a big deal either way. Makes sense not to exceed a max though if one is set.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
biancadanforth: No I don't feel strongly.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
biancadanforth: That pattern is for a path relative to a local file system though, not a URL right? I guess I'm not understanding how the new pattern's pathname doesn't also end with `'.pdf'`. 🤔

chenba: The PDFs generated by that script is deployed to the CDN.  (Take a look at the `assets` dir under the FxA project root dir.)

The url in the query params will be from the product metadata, so they can be... anything really.  But when properly configured, it should be a direct url to a PDF, or, one without the locale and file extension, relying on this endpoint to find the best-match locale.

packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -0,0 +1,184 @@
...
+  if (cachedVal !== undefined) {
+    if (isCacheExpired(cachedVal['checkedOn'])) {
+      cache.delete(url);
+    } else {
+      return cachedVal['result'];
+    }
+  }
+
+  try {
+    const res = await fetch(url, { method: 'HEAD' });
```
biancadanforth, line 76: When I used the debugger, I found that the `url` passed into `isValidDocUrl` is: "http://localhost:5000/legal/mozilla_vpn_privacy_notice.en.pdf", and this `res` above's status code is 404, since indeed that URL is not valid (i.e. there is no `mozilla_vpn_privacy_notice.en.pdf` file in `./assets`).

Setup I used:
* `npx serve` from `./assets`
* update and restart payments with `PAYMENT_LEGAL_DOWNLOAD_FQDN=localhost`

I also tried "http://localhost:3031/legal-docs/?url=http://localhost:5000/legal/mozilla_vpn_privacy_notice/" with my browser's locale set to `"fr-FR"`, and got the same 404 error.

biancadanforth (changes requested):
Edit: Maybe it's just that these files don't yet exist in `./assets`? I see from Lesley's follow-up PR that they are empty (e.g. http://localhost:5000/legal/mozilla_vpn_privacy_notice.en.pdf), and when I open them locally, I also get a 404.

Hey Barry thanks for clarifying some details with me async!

IIUC, it doesn't look like the localized versions of these docs that are getting copied into `./assets` have the same URL as this PR expects? I'm getting 404 errors locally when I try e.g. `"http://localhost:3031/legal-docs/?url=http://localhost:5000/legal/mozilla_vpn_privacy_notice/"` for `"en-US"` and `"fr-FR"` locales.

More details inline...

biancadanforth:
Barry and I discussed this more async. It's not that the file format doesn't match between what's in `./assets` and what the code expects. Rebasing off `main`, the expected files are there, e.g. `./assets/legal/mozilla_vpn_privacy_notice.en.pdf` and `./assets/legal/mozilla_vpn_privacy_notice.fr.pdf`.

The fallback to `./assets/legal/mozilla_vpn_privacy_notice.en.pdf` works as expected when the desired local doesn't have a translated document in `./assets`, though it looks like we don't get redirected to `./assets/legal/mozilla_vpn_privacy_notice.fr.pdf` when the locale is set to `"fr-FR"`.

Per discussion with Barry, the fix would be to ensure we try looking for the locale without the region (i.e. just `"fr"` in this case in `supportedLanguages.json` if we don't find any matches in the first pass (i.e. `supported` is empty after the first pass).

chenba:
@biancadanforth I rebased on main and fixed the bug that you found.  Thanks!

biancadanforth (approved):
LGTM; thanks Barry!

PR #7809, opened 2021-03-12
fix(auth-server): allow payment error out

bbangert:
Because:

* We validate responses, and the optional property for paypal error
  messages wasn't added to the validator.

This commit:

* Adds the optional property to the validator so an internal error
  will not be thrown.

## Checklist


say-yawn (approved):
WOW 👍

PR #7821, opened 2021-03-15
docs(adrs): globalize service config

dannycoates:


docs/adr/0024-globalize-service-configuration.md
```diff
@@ -0,0 +1,68 @@
...
+
+Pros:
+
+- Faster config change rollout in production
+- Enables a broader set of features than a library
+
+Cons:
+
+- More complex design and implementation than a library
+- Unclear if we'll need or utilize the extra features
```
clouserw, line 68: I don't think we should build things if we don't know that we'll use them

docs/adr/0024-globalize-service-configuration.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Decision Outcome
+
+TBD
+
+## Pros and Cons of the Options
+
+### Shared library with unified schema
+
+With this approach we would create a shared library that defines the unified schema and then refactor each service to follow that schema. We'd continue to use `convict` as the validator and exported config API.
```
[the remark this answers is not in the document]
dannycoates: Plenty, but I don't think it's worth considering other modules. `convict` works fine and there'd be significant work to switch without much benefit.

docs/adr/0024-globalize-service-configuration.md
dannycoates: I fully agree

chenba:
👍 

I prefer option 1.  I'd rather invest the effort to achieve the consistency we want.  The adapters option feels worse than what we currently have.

docs/adr/0024-globalize-service-configuration.md
```diff
@@ -0,0 +1,68 @@
...
+
+Pros:
+
+- Simple implementation
+- Solves the duplication and inconsistency problem
+- Doesn't preclude a config service at a later time
+
+Cons:
+
+- Requires refactoring services to use the new schema
```
vbudhram, line 41: Not really a big con, but I guess depending on how its organized and structured, it might not be as easy to find the conf value you could be looking for.

vbudhram (commented):
I'm leaning towards option one since it seems more straightforward and less complicated.

docs/adr/0024-globalize-service-configuration.md
dannycoates: It will be a massive config. Ideas for good organization welcome! I imagine we'll have some form of service specific hierarchy as well as the shared bits. It'd be nice if we made it so it was easy to lookup all the config options each service depends on somehow.

docs/adr/0024-globalize-service-configuration.md
jbuck: it sorta-kinda exists for the customs server at the moment, but it gets used not often at all

jbuck:
Given that we're moving to k8s, which would let us change config without redeploying, I'm also in favour of option 1.

docs/adr/0024-globalize-service-configuration.md
bbangert: Would it make sense to structure it according to functional component and keep the component configuration closer to it? Alternatively, perhaps config doesn't have to be a huge single block in one file, but could be in multiple config files based on the feature. NestJS has a variety of approaches (and no clear opinion on the one ring to rule them all) that might be worth looking at: https://docs.nestjs.com/techniques/configuration#partial-registration

docs/adr/0024-globalize-service-configuration.md
bbangert: We sort of use some features like this in event-broker, so that redeployments aren't needed to update relying party webhook endpoints. The webhook endpoints are kept in FireBase, and FireBase handles pushing state changes out as they occur that the event-broker code watches and then uses as soon as it gets the state change.

docs/adr/0024-globalize-service-configuration.md
dannycoates: I want to POC the implementation before we finalize this ADR. I spent a few minutes sketching out a single config starting from auth-server and noticed that my intent with this proposal is two things: standardize/dedupe the env var names, and share "public" config across services. Public being any config determined by one service but used by another. There's also massive amounts of "private" config only used by the defining service that doesn't need to be shared, but also shouldn't conflict in the env vars. So I think we'll still want to end up with a config spec per service, but allow for a single definition json (or env var set)

dannycoates:
I'd like to bring this up again later but it doesn't seem like a high priority at the moment

PR #7832, opened 2021-03-16
docs(adrs): draft auth-db-mysql adr

dannycoates:


6a68:
This looks good so far. I guess the alternatives to consider would be doing nothing, and moving auth-db back to its own physical layer, to regain the defense-in-depth advantages of a separate database microservice.

chenba:
Sounds good to me.

> This separation was deemed unimportant at some point, though the details are probably lost to history ...

This feels like the kind of thing rfk would remember.

dannycoates:
> This feels like the kind of thing rfk would remember.

Technically, ***I*** should remember since it happened during my first tour 😜

enote-kane:
Just out of curiosity: how is the fxa-email-service going to deal with the removal?

dannycoates:
> how is the fxa-email-service going to deal with the removal?

good question. We'll update fxa-email-service to connect to the db directly.

Issue #7693, opened 2021-03-02
Copy billing agreement address/payer info to Stripe


data-sync-user:
We need to copy the billing/payer fields as completely as possible from the paypal billing agreement into the stripe customer record.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3117)
┆Issue Number: FXA-3117

data-sync-user:
➤ Se Yeon Kim commented:

Using the GetExpressCheckoutDetails (https://developer.paypal.com/docs/nvp-soap-api/get-express-checkout-details-nvp/ ( https://developer.paypal.com/docs/nvp-soap-api/get-express-checkout-details-nvp/ ))

Payer Info available:

EMAILEmail address of buyer.Character length and limitations: 127 single-byte characters.PAYERIDUnique PayPal Customer Account identification number.Character length and limitations: 13 single-byte alphanumeric characters.|PAYERSTATUS|Status of buyer. Value is:

  * verified
  * unverified|

Character length and limitations: 10 single-byte alphabetic characters.|

COUNTRYCODEBuyer's country of residence in the form of ISO standard 3166 twocharacter country codes.Character length and limitations: 2 singlebyte characters.BUSINESSBuyer's business name.Character length and limitations: 127 single-byte characters.Payment info available:
|PAYMENTINFO_n_CURRENCYCODE|The currency code ( https://developer.paypal.com/docs/nvp-soap-api/currency-codes/ ) of the financing amounts; default is USD.
 Character length and limitations: Three single-byte characters.|

[~bbangert@mozilla.com] Outside of the email, name, description, and metadata already added to the Stripe Customer during StripeHelper.createPlainCustomer() call, what are the other information that should be added to the Stripe Customer ( https://stripe.com/docs/api/customers/object ) record: currency, address.country?

PR #7838, opened 2021-03-16
feat(auth-server): copy billing address info to Stripe customer

say-yawn:
## Because

- when creating a new billing agreement we want to make sure the billing address information is copied to the Stripe Customer

## This pull request

- expands existing PayPal handler to retrieve billing address from Billing Agreement

Closes: #7693

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -66,6 +66,14 @@ export const SUBSCRIPTION_UPDATE_TYPES = {
   CANCELLATION: 'cancellation',
 };
 
+type BillingAddressOptions = {
+  city: string,
+  country: string,
+  line1: string,
+  line2: string,
+  postalCode: string,
+  state: string,
+}
 /**
  * Determine for two product metadata object's whether the new one
  * is a valid upgrade for the old one.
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -507,6 +507,37 @@ export class StripeHelper {
...
+   * @param customer_id 
+   * @param city 
+   * @param country 
+   * @param line1 
+   * @param line2 
+   * @param postal_code 
+   * @param state 
+   */
+  async updateCustomerBillingAddress(
+    customer_id: string,
```
bbangert, line 522: This is a lot of arguments, it would be much cleaner to pass this in as an object.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -507,6 +507,37 @@ export class StripeHelper {
...
+    city: string,
+    country: string,
+    line1: string,
+    line2: string,
+    postal_code: string,
```
bbangert, line 527: This should be camelCased.

bbangert (changes requested):
Looks good, except for the one function with a lot of positional args.

packages/fxa-auth-server/lib/payments/stripe.ts
say-yawn: Hmm, I purposely did `postal_code` because it used that in here https://stripe.com/docs/api/customers/object#customer_object-address

bbangert (approved):
👍🏻

Issue #7834, opened 2021-03-16
/.well-known/change-password 404s on CI


6a68:
This redirects correctly for me locally, but on CI seems to 404: https://app.circleci.com/pipelines/github/mozilla/fxa/17302/workflows/9fe7b0d0-68bd-4403-b9b1-206aa1203d2c/jobs/224495

Let's get it working

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3190)
┆Issue Number: FXA-3190

data-sync-user:
➤ Jared Hirsch commented:

Vijay fixed this in #7849.

PR #7849, opened 2021-03-17
fix(test): update well-known change password test

vbudhram:
## Because

- This is a failing test in the new settings
- I can't really pinpoint why this test is starting to fail with moving to new settings, possibly `got` request library might not be following the redirect, but...
- I've tested this functionality locally and it does follow the redirect, so it is very little risk

## This pull request

- Fixes test by removing it lol, nah I pulled the redirect url assert into a different part

Closes: #7834

## Checklist


6a68 (approved):
Looks great, thanks @vbudhram! Good to have you back 🎉

Issue #7703, opened 2021-03-03
[FxA-Settings] The picture taken with camera is offered for editing rotated at 90 degrees


data-sync-user:
**Build**
Stage [Train 1.201.1]

**Affected platforms**
iOS 14.4

**Steps to reproduce**

1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
1. Click on the "Add" button displayed on the right side of the Picture section.
1. Click on the "Add photo" button.
1. Click on the "Take photo" button.
1. Take the photo and tap on the "Use photo" option.
1. Observe how the picture is displayed.

**Expected results**
The picture appears in the edit area in the same position in which it was made.

**Actual Results**
The picture appears rotated at 90 degrees.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3121)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15396/gif17.gif">gif17.gif</a>
┆Issue Number: FXA-3121

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.203.2] under iOS 14.4. I confirm that the picture appears in the edit area in the same position in which it was made.

PR #7858, opened 2021-03-17
fix(avatar): Avatar upload tweaks for new settings

vbudhram:
## Because

- Avatar uploading from iOS doesn't always work as expected
- Taking a picture via camera on Firefox and Safari creates an image with different offsets (290px), so they will look different

## This pull request

- Removes auto rotation when uploading an image, `react-easy-cropper` allows the user the ability to rotate the image as needed
- When taking a photo from the camera, the user is taken to the cropper page to finalize the picture before saving

Closes: #7703

## Checklist


dannycoates:
This doesn't fix 7705 for me

dannycoates (approved):

PR #7859, opened 2021-03-17
fix(auth-server): skip currency check without pmi

bbangert:
Because:

* We can only do the currency check when a payment method id is
  provided. The Stripe createSubscriptionWithPMI intentionally
  allows no paymentMethodId to be provided to handle SCA which the
  check did not account for.

This commit:

* Only verifies currency matches when a payment method id is provided.

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
@@ -494,18 +494,21 @@ export class StripeHandler {
       idempotencyKey,
     } = request.payload as Record<string, string>;
 
-    const planCurrency = (await this.stripeHelper.findPlanById(priceId))
-      .currency;
-    const paymentMethodCountry = (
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -494,18 +494,21 @@ export class StripeHandler {
...
-      await this.stripeHelper.getPaymentMethod(paymentMethodId)
-    ).card?.country;
-    if (
-      !this.stripeHelper.currencyHelper.isCurrencyCompatibleWithCountry(
-        planCurrency,
-        paymentMethodCountry
-      )
-    ) {
-      throw error.currencyCountryMismatch(planCurrency, paymentMethodCountry);
+    // Skip the payment source check if there's no pyament method id.
```
dannycoates, line 497: this is ts not py ;)

dannycoates (approved):
LGTM

PR #7864, opened 2021-03-17
bug(l10n): Minor fixes to payments l10n

6a68:
## Because

- @peiying2 noticed some problems with the ftl file for payments

## This pull request

* Treats 'Stripe' as an untranslated brand, similar to how we handle 'Paypal'

* Corrects a typo (occured vs occurred)

* Replaces 3 periods with ellipses (thankfully this change doesn't require updating the l10n string IDs)

Closes: #mozilla/fxa-content-server-l10n#480.

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -7,6 +7,7 @@ project-brand = Firefox Accounts
 -brand-name-mozilla = Mozilla
 -brand-name-firefox = Firefox
 -brand-name-paypal = PayPal
+-brand-name-stripe = Stripe
 
 document =
   .title = Firefox Accounts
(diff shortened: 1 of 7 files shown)


packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -21,7 +22,7 @@ general-error-heading = General application error
 basic-error-message = Something went wrong. Please try again later.
 payment-error-1 = Hmm. There was a problem authorizing your payment. Try again or get in touch with your card issuer.
 payment-error-2 = Hmm. There was a problem authorizing your payment. Get in touch with your card issuer.
-payment-error-3 = An unexpected error has occured while processing your payment, please try again.
+payment-error-3b = An unexpected error has occurred while processing your payment, please try again.
```
chenba, line 25: oh boy, these unique ids...

chenba (approved):
@6a68 Thanks!

packages/fxa-payments-server/public/locales/en-US/main.ftl
6a68: yeah. `payment-error-3b` seemed better somehow than `payment-error-3-1` 🤷

clouserw: This is fine, since it's landed, but this didn't need a new ID since it was an english-only typo

PR #7877, opened 2021-03-18
fix(auth-server): handle failed checkout payments

### LINKED ISSUE #7852: [Bug] Subscription active with failed PayPal transaction in Checkout
STR:

1. Create a sandbox buyer account on PayPal with only $4 on the the PayPal credit and the bank account removed (credit card should stay as the only funding source)
1. Go the the product page and go through the checkout and click the PayPal button
1. Create a Billing agreement and before you click the green "Continue" button, go back to the buyer PayPal account and remove the credit card funding source to force transaction to fail
1. Click the green "Continue" and check back on your Subscription Platform page that you get "Something went wrong" error page.
1. Refresh the page and check that a valid subscription was created

Expected:

5. Refresh the page and check that you are redirected to start a new checkout.

This bug happened because in the checkout we do not end the subscription if the transaction has failed resulting to an open invoice with valid subscription. Fix is to remove the Subscription when we get a failed transaction in the checkout page.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3198)
┆Issue Number: FXA-3198

bbangert:
Because:

* We cannot mark send_invoice subscriptions as incomplete when an
  initial payment fails to require another payment attempt before
  activating the subscription.
* We should send subscription failed emails when processing an invoice
  via webhook.

This commit:

* Updates the checkout flow code so that the billing agreement and
  subscription are removed if the PayPal transaction fails. This allows
  the user to choose a new payment option and prevents temporary
  subscription entitlements.
* Adds error handling for invoice processing in the stripe webhook
  triggered invoice processing to ensure the customer is notified should
  the invoice fail to get paid.

Closes #7852

Co-authored-by: Se Yeon Kim <say-yawn@users.noreply.github.com>

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -4,6 +4,11 @@
 import { ServerRoute } from '@hapi/hapi';
 import isA from '@hapi/joi';
 import * as Sentry from '@sentry/node';
+import { PayPalClientError } from 'fxa-auth-server/lib/payments/paypal-client';
+import {
+  PAYPAL_BILLING_AGREEMENT_INVALID,
+  PAYPAL_SOURCE_ERRORS,
+} from 'fxa-auth-server/lib/payments/paypal-error-codes';
 import { Stripe } from 'stripe';
 import Container from 'typedi';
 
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -288,7 +293,36 @@ export class StripeWebhookHandler extends StripeHandler {
...
+          const uid = customer.metadata.userid;
+          const billingAgreementId = this.stripeHelper.getCustomerPaypalAgreement(
+            customer
+          ) as string;
+          await this.stripeHelper.removeCustomerPaypalAgreement(
+            uid,
+            customer.id,
+            billingAgreementId
+          );
+          await this.sendSubscriptionPaymentFailedEmail(invoice);
```
chenba, line 315: If we remove the PayPal BA and keep the subscription, what does the customer see in the subscription management page?  Maybe @biancadanforth or @meandavejustice know.

bbangert: They should see a prompt to put a new PayPal agreement on file. There's a specific response to the `GET .../customer` I added that the front-end will use to determine whether to show it.

chenba (approved):
The changes make sense to me.  But I'm not quite sure about the effect of removing the BA has on the subscription management page, especially since we also send an email suggesting the customer to go there to update their payment info.  @biancadanforth and/or @meandavejustice might know about the current state of the subscription management page.

Issue #7876, opened 2021-03-18
[iOS] Cannot access account settings


data-sync-user:
* Use iOS
* Be logged in to sync in Firefox

STR:

1) Use iOS, run Firefox
 2) Click Hamburger Menu
 3) Click your email address
 4) Click Manage Account & Devices

We see a flash (1-2 seconds) of a General Application Error page and then are redirected to a log in. This error should be resolved so that users can access Settings.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3208)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15477/Screenshot_2021-03-18+Slack+Threads+Mozilla.png">Screenshot_2021-03-18 Slack Threads Mozilla.png</a>
┆Issue Number: FXA-3208

data-sync-user:
➤ Danny Coates commented:

The error flash is fixed. (deploying soon)

"Since we are logged in to sync still, we shouldn't have to log in at this point."

This is more complicated and would require some browser work. The mobile browsers authenticate to setting separately from sync.

data-sync-user:
➤ Chelsea Lewis commented:

The error flash fix was deployed with FxA 202.3 on March 18th at 3:42pm PT. Reaching out to iOS and FxA PMs to confirm flow wrt ticket's A/C: "Since we are logged in to sync still, we shouldn't have to log in at this point."

data-sync-user:
➤ Wil Clouser commented:

"Since we are logged in to sync still, we shouldn't have to log in at this point." was my assumption and if not a regression is unrelated to this issue.

data-sync-user:
➤ Chelsea Lewis commented:

Removed the assumption from the ticket per conversation in Slack. As this is not a regression, and as the error is resolved, closing this bug.

PR #7881, opened 2021-03-18
fix(settings): fix session ttl in settings

dannycoates:
We weren't accounting for token ttl when fetching the session token from the gql-api server, which would cause the frontend to think it was ok, but the auth-server would 401/110 the requests for data, triggering a GAE error.

This is a temporary fix for #7876

### DIFF
--- a/packages/fxa-settings/src/components/App/index.tsx
+++ b/packages/fxa-settings/src/components/App/index.tsx
@@ -55,7 +55,10 @@ export const App = ({ flowQueryParams, navigatorLanguages }: AppProps) => {
     Metrics.init(flowQueryParams);
   }, [flowQueryParams]);
 
-  if (loading) {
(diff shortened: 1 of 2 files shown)


vbudhram (approved):
@dannycoates LGTM! Works locally when setting a short timeout.

packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -55,7 +55,7 @@ export const App = ({ flowQueryParams, navigatorLanguages }: AppProps) => {
     Metrics.init(flowQueryParams);
   }, [flowQueryParams]);
 
-  if (loading) {
+  if (loading || error?.message.includes('Invalid token')) {
```
vbudhram, line 58: :shipit:

Issue #7454, opened 2021-02-08
Subscription management: Add functionality to change PayPal billing information


data-sync-user:
Currently, clicking the "Change" button under Billing Information on the Subscription Management page displays a Stripe widget to enter a new credit card.

A new field called `payment_provider` is present on the customer object to determine whether to display the Stripe widget, or a button for PayPal.

Depending on the state of any active subscription(s), the Change button for PayPal may take one of two forms:

1. If there is a valid Billing Agreement ID in the Stripe Customer metadata, the Change button would link the user to PayPal where they can change their payment method any time.
  1. If the customer has a past_due invoice, a warning call-out indicating that the user must update their paypal agreement is shown.
  1. The URL can be generated for this as: [https://www.paypal.com/myaccount/autopay/connect/](https://www.paypal.com/myaccount/autopay/connect/) {BILLING_AGREEMENT_ID}
1. If the Billing Agreement ID is invalid, instead of a Change button, there would be a PayPal button that works similarly to the PayPal button used for new customer checkout (FXA-2800) and a warning call-out indicating that the user needs to create a new Billing Agreement.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2806)
┆Issue Number: FXA-2806

data-sync-user:
➤ Barry Chen commented:

(Sorry [~bdanforth@mozilla.com], when we were chatting about this I didn't realize there's a link to the PayPal docs in the description...)  If I understand correctly, updating the payment method does not actually change the billing agreement.  That means we don't need to update the billing agreement id on our end.  In that case, would providing a button to the payment method update form be enough?

data-sync-user:
➤ Bianca Danforth commented:

Thanks Barry; I think you're right that the billing agreement ID wouldn't change. It looks like there may be a way for PayPal customers to be redirected to PayPal to change their payment method ( https://developer.paypal.com/docs/archive/express-checkout/integration-guide/ECReferenceTxns/#change-payment-methods ) for the same billing agreement.

data-sync-user:
➤ Benjamin Bangert commented:

Agreed, changing your payment method is done on PayPal, not via our page. We could link a user to the PayPal account management page for our specific billing agreement though. Leaving this ticket open to track adding that link/information to the page.

data-sync-user:
➤ Bianca Danforth commented:

[...]

Issue #7903, opened 2021-03-19
The billing update notification overlaps the Firefox Accounts logo text


data-sync-user:
**Build:**

* Stage [Train 1.202.3]

**Affected platforms:**

* Windows 10 x64

**Prerequisites:**

* You have at least 1 active subscription on the FxA Subscriptions page.

**Steps to reproduce:**
1. Navigate to your FxA Subscriptions page.
2. Select one active subscription.
3. Change any data from the “Payment information” section.
4. Save the change and observe what happens next.

**Expected results:**

* The notification of the billing information update is displayed correctly, without overlapping any of the page’s elements.

**Actual Results:**

* The notification of the billing information update overlaps the Firefox Accounts logo text.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3225)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15577/logo+overlap.jpg">logo overlap.jpg</a>
┆Issue Number: FXA-3225

data-sync-user:
➤ Dave Justice commented:

We have an issue to update the location and style of notifications planned for the follow up to this epic, https://jira.mozilla.com/browse/FXA-3229

data-sync-user:
➤ Barry Chen commented:

Note that this was not a bug; the style has been that way.  I'll make a change to have the alert bar width match the content.  However, given the placement and z index of the alert bar, when the window is narrow enough, it'd still overlap the FxA heading.

Issue #7883, opened 2021-03-18
[subplat] Create a new Amplitude event for when the "Change" button is clicked


data-sync-user:
Quoting part of [this comment](https://jira.mozilla.com/browse/FXA-2895?focusedCommentId=243696&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-243696) in FXA-2895: 

_After a lengthy [Slack conversation with Barry](https://mozilla.slack.com/archives/CG0AJ6E77/p1616096159021900), we agreed that the best strategy is to recognize that Stripe and PayPal work differently, so we should handle this event differently. Specifically, we should create a new event that is recorded when the "Change" button is clicked. We pass `paymentProvider` this new event as well as to the existing `Amplitude.updateDefaultPaymentMethod_*()` events._

_What this means is that:_

* _The new event will see `paymentProvider` with values of both `'stripe'` and `'paypal'`, since clicking the "Change" button is what triggers the Stripe form to appear in the Stripe case._
* _The existing `Amplitude.updateDefaultPaymentMethod_*()` events will never have a value of `'paypal'`._

_This new event should be added to the [Subscription Platform metrics spreadsheet](https://docs.google.com/spreadsheets/d/1ttgV60YeurrMhmdlCZ5WNBJXa-PcnRmrGgOEow71E1U/edit#gid=0) and to our Amplitude dashboard (CC [~byee@mozilla.com] )._

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3213)
┆Issue Number: FXA-3213

data-sync-user:
➤ Bianca Danforth commented:

We should probably add a property to indicate if there is an error state here, so we can associate any downstream events like `updateDefaultPaymentMethod_*`. FXA-2806 will make use of this event, but we only know success/failure for the invalid billing agreement error state for PayPal customers.

PR #7909, opened 2021-03-19
feat(fxa-payments-server): add paypal to subscription management

meandavejustice:
## Because

- We need to be able to notify the user of issues with their payment provider, and allow them to fix it without our intervention.

## This pull request

- Adds notifications for 2 different paypal error cases, `billing_agreement_missing` and `funding_source`.
- Adds behavior to change paypal funding source or fix issues with missing billing agreements.
- This gets the styles pretty close to what they need to be, they should be made pixel perfect in the UI labeled issues in the `Support PayPal Phase 2` epic. 

Closes: #7454

## Checklist

## Screenshots (Optional)
funding-source-desktop
missing-agreement-desktop
missing-agreement-mobile
missing-agreement-modal-desktop
missing-agreement-modal-mobile

_a note on the notifications_

I made a decision here to match the notification style to what we already have within the payment-server. I created a follow up issue https://jira.mozilla.com/browse/FXA-3229 to address the notification refactor as a part of the `Support PayPal Phase 2` epic.


meandavejustice:
I know we don't like to pre-maturely open up Pull Requests, but the only thing left on this is to add 0.03% test coverage 👀.

```
Jest: "global" coverage threshold for branches (94%) not met: 93.97%
```
So I figured it would be reasonable to allow people to get some eyes on it, since there is a pretty big diff here.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -65,6 +93,68 @@ export const PaymentUpdateForm = ({
...
+  const errorAlertBarContent = (
+    <Localized
+      id="sub-route-invalid-payment-alert"
+      elems={{ div: actionButton }}
+    >
+      <span>
+        Invalid payment information, there is an error with your account.{' '}
+        {actionButton}
+      </span>
+    </Localized>
```
biancadanforth, line 118: Never seen this `elems` attribute. Is this why you have the `<div>` inside the `main.ftl` file? What is this doing exactly?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -65,6 +93,68 @@ export const PaymentUpdateForm = ({
...
+    /* istanbul ignore next */
+    script.onload = () => {
+      setPaypalScriptLoaded(true);
+    };
+    /* istanbul ignore next */
+    script.onerror = () => {
+      throw new Error('Paypal SDK could not be loaded.');
+    };
+    document.body.appendChild(script);
+  }, []);
```
biancadanforth, line 156: We should dedupe this hook; maybe export it from the `PaypalButton` component, so it can be used in `SubscriptionCreate` as well as `PaymentUpdateForm`?

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -420,6 +420,8 @@ module.exports.subscriptionsCustomerValidator = isA.object({
   payment_type: isA.string().optional(),
   paypal_payment_error: isA.string().optional(),
   brand: isA.string().optional(),
+  paypal_payment_error: isA.string().optional(),
+  billing_agreement_id: isA.string().optional(),
```
biancadanforth, line 424: Can this also be `null`, e.g. if the billing agreement is invalid? I see it on the `Customer` object in fxa-payments-server/src/store/types.tsx  as typed `string | null`.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -124,6 +216,47 @@ export const PaymentUpdateForm = ({
...
+                <div className="paypal-button">
+                  <PaypalButton
+                    currencyCode={currencyCode}
+                    customer={customer}
+                    idempotencyKey={submitNonce}
+                    priceId={priceId}
+                    refreshSubscriptions={refreshSubscriptions}
+                    setPaymentError={setPaymentError}
+                    ButtonBase={paypalButtonBase}
+                  />
```
biancadanforth, line 253: @bbangert Looking at the [Checkout ladder diagram](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#paypal-checkout), is it okay to use the same endpoint (`POST /oauth/subscription/active/new-paypal`) here for `onApprove` in our `PaypalButton`?

In this case, we don't want to create a new subscription; we want to update an existing subscription... Is there different accounting for this in the back-end? Do we want to send the same emails as for a new subscription (i.e. [Welcome and First Invoice e-mails](https://miro.com/app/board/o9J_lWaWaZA=/))?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </LinkExternal>
+    </Localized>
+  );
+
+  const paypalFundingSourceActionButton = (
+    <Localized id="pay-update-manage-btn">
+      <LinkExternal
+        data-testid="manage-payment-update-button"
+        className="settings-button error-button"
+        href={`https://www.paypal.com/myaccount/autopay/connect/${billing_agreement_id}`}
```
biancadanforth, line 60: We should note for QA that this link will look like an error when testing on Stage -- even if they change the domain manually after it opens in a new tab to `www.sandbox.paypal.com`. PayPal just doesn't support this webpage for sandbox accounts.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -65,6 +93,68 @@ export const PaymentUpdateForm = ({
...
+    paypal_payment_error,
+  } = customer;
+
+  const actionButton = (
+    <ActionButton
+      {...{ customer, onRevealUpdateClick, revealFixPaymentModal }}
+    />
+  );
+
+  const errorAlertBarContent = (
```
biancadanforth, line 109: This displays over the header e.g. for a `'missing_agreement'` error; but that's a bug captured in #7903 .

biancadanforth (changes requested):
Thanks Dave for this ambitious PR! You cover a lot of ground here.

I have a few requests around making things DRYer, and I'm wondering if we can "just" use the `new-paypal` endpoint from Checkout in the same way here (I pinged Ben inline).

I don't think we should add more scope to this PR, but can you also file a follow-up for what the expected UX should be after the customer successfully re-authorizes PayPal and creates a new Billing Agreement? CC @jess-cook03 I don't imagine we want to show the "Confirming subscription..." processing screen and then "Subscription confirmation" screen like we do in Checkout (if that even works -- I just mocked `customer` on the front-end so far to test this out...)?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
meandavejustice: @biancadanforth That's a great point. CC/ @bbangert Do you know if we have an answer for this currently?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Where would be a good spot to document this?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
meandavejustice: We also have https://jira.mozilla.com/browse/FXA-3229 logged for Phase 2 of Support Paypal

packages/fxa-auth-server/lib/routes/validators.js
meandavejustice: Great catch, I was under the impression that `optional` took care of the null case here, but I see above for `billing_name` we are doing:

```
billing_name: isA
    .alternatives(isA.string(), isA.any().allow(null))
    .optional(),
```
That should work here as well.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+      >
+        <span className="change-button" data-testid="change-button">
+          Change
+        </span>
+      </LinkExternal>
+    </Localized>
+  );
+
+  const paypalFundingSourceActionButton = (
+    <Localized id="pay-update-manage-btn">
```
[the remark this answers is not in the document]
meandavejustice: What is the issue you are seeing? I believe we want the label to be "Manage" here.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+    </Localized>
+  );
+
+  const paypalMissingAgreementActionButton = (
+    <button
+      data-testid="reveal-payment-update-button"
+      className="settings-button error-button"
+      onClick={revealFixPaymentModal}
+    >
+      <Localized id="pay-update-manage-btn">
```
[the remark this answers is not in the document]
meandavejustice: I don't think I understand what you are requesting with these ids, lets chat in slack about it

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
```diff
@@ -94,6 +97,57 @@ describe('routes/Subscriptions/PaymentUpdateFormV2', () => {
...
+      <Subject
+        customer={Object.assign({}, CUSTOMER, {
+          payment_provider: 'paypal',
+          paypal_payment_error: 'missing_agreement',
+        })}
+      />
+    );
+    expect(
+      screen.queryAllByTestId('reveal-payment-update-button').length
+    ).toEqual(2);
```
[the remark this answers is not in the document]
meandavejustice: One action button is rendered in the notification and one is in the settings view

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <span className="manage-button" data-testid="manage-button">
+          Manage
+        </span>
+      </LinkExternal>
+    </Localized>
+  );
+
+  const paypalMissingAgreementActionButton = (
+    <button
+      data-testid="reveal-payment-update-button"
```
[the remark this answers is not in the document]
meandavejustice: Updated

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
```diff
@@ -94,6 +97,57 @@ describe('routes/Subscriptions/PaymentUpdateFormV2', () => {
...
+    );
+  });
+
+  it('renders correctly for incorrect funding source for paypal', async () => {
+    render(
+      <Subject
+        customer={Object.assign({}, CUSTOMER, {
+          payment_provider: 'paypal',
+          paypal_payment_error: 'funding_source',
+        })}
```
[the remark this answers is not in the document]
meandavejustice: Ah yes, much better

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -28,27 +46,37 @@ export type PaymentUpdateAuthServerAPIs = Pick<
...
+  paypalButtonBase,
 }: PaymentUpdateFormProps) => {
   const [submitNonce, refreshSubmitNonce] = useNonce();
   const [updateRevealed, revealUpdate, hideUpdate] = useBooleanState();
   const [inProgress, setInProgress, resetInProgress] = useBooleanState(false);
+
+  const [
+    fixPaymentModalRevealed,
+    revealFixPaymentModal,
+    hideFixPaymentModal,
```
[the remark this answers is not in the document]
meandavejustice: I used `useBooleanState` in order to follow the established patterns. I personally prefer to just use `useState(false)` in these cases.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
meandavejustice: https://github.com/projectfluent/fluent.js/wiki/Localized#markup

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -304,6 +304,9 @@ sub-subscription-error =
 sub-customer-error =
   .title = Problem loading customer
 sub-billing-update-success = Your billing information has been updated successfully
+sub-route-payment-modal-heading = Invalid billing information
+sub-route-payment-modal-copy = There seems to be an error with your Paypal account, we need you to take the necessary steps to resolve this payment issue.
+sub-route-invalid-payment = Invalid payment information, there is an error with your account. <div>Manage</div>
```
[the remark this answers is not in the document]
meandavejustice: https://github.com/projectfluent/fluent.js/wiki/Localized#markup

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
bbangert: Yes, we do have a new endpoint for this condition that just updates the payment:
https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts#L356

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -28,27 +46,37 @@ export type PaymentUpdateAuthServerAPIs = Pick<
 >;
 
 export type PaymentUpdateFormProps = {
+  plan: Plan | null;
```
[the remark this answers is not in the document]
meandavejustice: This is because `planForId` can return null, https://github.com/mozilla/fxa/blob/44bbb161d958f084c1bb39902554f69a2333de90/packages/fxa-payments-server/src/routes/Subscriptions/index.tsx#L331

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -420,6 +420,9 @@ module.exports.subscriptionsCustomerValidator = isA.object({
   payment_type: isA.string().optional(),
   paypal_payment_error: isA.string().optional(),
   brand: isA.string().optional(),
+  billing_agreement_id: isA
+    .alternatives(isA.string(), isA.any().allow(null))
```
chenba, line 424: Isn't `isA.string()` a subset of `isA.any()`?  Do we need `isA.string()`?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+      className="settings-button"
+      onClick={onRevealUpdateClick}
+    >
+      <Localized id="pay-update-change-btn">
+        <span className="change-button" data-testid="change-button">
+          Change
+        </span>
+      </Localized>
+    </button>
+  );
```
chenba, line 39: nit: this causes React.createElement to be called immediately regardless we end up using the Stripe button or not.  We could have it as a function, `const stripeActionButton = () => (...` then call it where it's used.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const paypalActionButton = (
+    <Localized id="pay-update-change-btn">
```
chenba, line 42: Should this be around the `span`  and not all of `LinkExternal`?  Does the `span` in `LinkExternal` (https://github.com/mozilla/fxa/blob/7454-paypal-sub-management/packages/fxa-react/components/LinkExternal/index.tsx#L33) still get rendered with this?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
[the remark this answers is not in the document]
chenba: I cannot seem to find a `pay-update-manage-btn` in main.ftl

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+      actionButton = paypalMissingAgreementActionButton;
+      break;
+    }
+    case 'funding_source': {
+      actionButton = paypalFundingSourceActionButton;
+      break;
+    }
+    default: {
+      actionButton =
+        payment_provider === 'paypal' ? paypalActionButton : stripeActionButton;
```
chenba, line 95: There's a `PaymentProvider.isPaypal` checking the value.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
```
chenba, line 97: There's enough logic here that I think we should extract it to a function.  I also think checking the payment provider in a switch statement for the type of Paypal payment error is confusing.  If we have this in a function, then we could return early when the payment provider is Stripe.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return <div className="action">{actionButton}</div>;
+};
```
chenba, line 100: I don't see a `ActionButton.test.tsx` file.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
```diff
@@ -94,6 +97,55 @@ describe('routes/Subscriptions/PaymentUpdateFormV2', () => {
...
+        customer={{
+          ...CUSTOMER,
+          payment_provider: 'paypal',
+          paypal_payment_error: 'missing_agreement',
+        }}
+      />
+    );
+    expect(
+      screen.queryAllByTestId('reveal-payment-modal-button').length
+    ).toEqual(2);
```
chenba, line 119: I don't understand why `2` here.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
```diff
@@ -94,6 +97,55 @@ describe('routes/Subscriptions/PaymentUpdateFormV2', () => {
...
+        customer={{
+          ...CUSTOMER,
+          payment_provider: 'paypal',
+          paypal_payment_error: 'funding_source',
+        }}
+      />
+    );
+    expect(
+      screen
+        .queryAllByTestId('manage-payment-update-button')[0]
```
chenba, line 144: Does this imply that there are also multiple `manage-payment-update-button` buttons?  Why is that?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -65,6 +89,35 @@ export const PaymentUpdateForm = ({
...
+  const actionButton = (
+    <ActionButton
+      {...{ customer, onRevealUpdateClick, revealFixPaymentModal }}
+    />
+  );
+
+  const errorAlertBarContent = (
+    <Localized
+      id="sub-route-invalid-payment-alert"
+      elems={{ div: actionButton }}
```
chenba, line 108: Why `div` when it's not a `div`?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -102,17 +155,19 @@ export const PaymentUpdateForm = ({
...
-  const { last4, exp_month, exp_year, payment_provider } = customer;
-
   // https://github.com/iamkun/dayjs/issues/639
   const expirationDate = dayjs()
     .set('month', Number(exp_month) - 1)
     .set('year', Number(exp_year))
     .format('MMMM YYYY');
 
+  /* istanbul ignore next */
+  const currencyCode = plan && plan.currency ? plan.currency : 'USD';
```
chenba, line 167: Plan.currency has a string type so I don't think we need the truthy check.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -102,17 +155,19 @@ export const PaymentUpdateForm = ({
...
+  /* istanbul ignore next */
+  const priceId = plan && plan.plan_id ? plan.plan_id : 'UNKNOWN';
```
chenba, line 169: Same deal here, Plan.plan_id has a string type.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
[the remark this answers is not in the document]
meandavejustice: that'll do it!

chenba (changes requested):
Left some comments.  Also looks like the Paypal button won't do the right thing since it was written for subscribing to a product (per comment thread started by Bianca).

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
meandavejustice: Same as above, it is rendered twice: once in the notification and once in the subscription view

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
meandavejustice: notication and view

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
meandavejustice: the outer most element is a div

packages/fxa-auth-server/lib/routes/validators.js
meandavejustice: Here I was following the patterns throughout the file. I believe this is the way you write out `string | null`

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: All of the logic for `ActionButton.tsx` is covered in `PaymentUpdateForm.test.tsx`

biancadanforth:
> I don't think we should add more scope to this PR, but can you also file a follow-up for what the expected UX should be after the customer successfully re-authorizes PayPal and creates a new Billing Agreement?

Just wanted to say that I filed this ticket already (#7945 ), since Tony was asking me about estimation earlier today. I'll look more closely at your and Barry's comments tomorrow.

biancadanforth:
@meandavejustice Is this ready for another look? Whenever it's ready, can you re-request a review?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Added

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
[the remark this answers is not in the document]
biancadanforth: I think this was resolved with your latest update based on this [related comment](https://github.com/mozilla/fxa/pull/7909#discussion_r599161608).

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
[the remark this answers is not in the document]
biancadanforth: Hmm... not following here, but it doesn't look like this ended up in your latest patch. Just the Stripe action button has this test id now which WFM.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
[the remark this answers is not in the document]
biancadanforth: What I mean is that I don't understand the scenario in which `plan` would be `null`. Now that I'm thinking about it more, I suppose it could be possible that someone has multiple subscriptions, and in that case perhaps `plan` would be `null` for this form, which covers the payment method for all subscriptions?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
biancadanforth: I mentioned this in Slack as well, but I'd recommend we send QA an e-mail after this PR lands to let them know they can test this and list out any test cases that aren't yet captured in their existing test plan. We can work on that together if you'd like.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
biancadanforth: What was the approach before that, Dave/Barry? I don't remember, and it'd be good to keep it in mind.

meandavejustice: @biancadanforth Not sure what you are asking here, do you mean what was the code before I addressed your comment and turned the variables into functions instead of element declarations?

packages/fxa-auth-server/test/local/routes/subscriptions/stripe.js
```diff
@@ -1399,7 +1399,7 @@ describe('DirectStripeRoutes', () => {
               .hasSubscriptionRequiringPaymentMethod,
             customer
           );
-          sinon.assert.calledOnceWithExactly(
+          sinon.assert.calledWithExactly(
```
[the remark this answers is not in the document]
chenba: I still don't see test changes for https://github.com/mozilla/fxa/pull/7909/files#diff-bc4cc467438b7003118180f818c50d9a92f54b6f6e3e2959de0e1e8f5840f39bR378-R382

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </Localized>
+    </button>
+  );
+
+  const paypalActionButton = () => (
+    <Localized id="pay-update-change-btn">
+      <LinkExternal
+        data-testid="change-payment-update-button"
+        className="settings-button"
+        href={`https://www.paypal.com/myaccount/autopay/connect/${billing_agreement_id}`}
```
chenba, line 48: The server side changes (the optional property on the type and the response validator) are saying that `billing_agreement_id` could be undefined, but in this component we use it without handling that state.

packages/fxa-payments-server/src/components/PayPalButton/index.tsx
```diff
@@ -79,19 +84,26 @@ export const PaypalButton = ({
...
+        if (newPaypalAgreement && priceId) {
+          await apiCapturePaypalPayment({
+            idempotencyKey,
+            priceId,
+            token,
+          });
+        } else {
+          await apiUpdateBillingAgreement({
+            token,
+          });
```
chenba, line 105: I cannot seem to find the test changes related to this.

meandavejustice: The PayPalButton Test file doesn't have any tests except `Doesn't render the PayPal button if the PayPal script fails to load`. I'm assuming this was initially skipped because of the difficulty of loading the paypal button third party scripts in a test environment. I think if we want to tackle that it should not be within this pull request.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: My understanding is that if `billing_agreement_id` is missing, then `paypal_payment_error` should be `missing_agreement` in which we will return `paypalMissingAgreementActionButton`.

Is there a case where `billing_agreement_id` would come back undefined without `paypal_payment_error` being set?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Having trouble picturing what you are asking for here.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
biancadanforth: Exactly @meandavejustice -- how were the components declared before?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.test.tsx
chenba: Ah.  Sorry, I think the older comment was marked as resolved and it auto-collapsed by Github.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
biancadanforth: Something like this perhaps? Or do you find the switch statement problematic, Barry?

```js
  const setActionButton = () => {
    if (PaymentProvider.isStripe(payment_provider)) {
      return stripeActionButton;
    }
    switch (paypal_payment_error) {
      case 'missing_agreement': {
        return paypalMissingAgreementActionButton;
      }
      case 'funding_source': {
        return paypalFundingSourceActionButton;
      }
      default: {
        return paypalActionButton;
      }
    }
  }

  const actionButton = setActionButton();
```

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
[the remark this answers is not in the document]
chenba: hmmmm if I understand this correctly, we want a plan to determine the currency since we need that for PayPal.  But, when the plan is null, we default to USD.  Since we are dealing with a current customer, it's possible that the customer's subscriptions are not in USD.  And as far as I know, having a currency mismatch between the payment provider and the subscriptions is undefined behavior.

I think we shouldn't allow plan to be null here.

packages/fxa-payments-server/src/components/PayPalButton/index.tsx
biancadanforth: I think Barry means that the `apiUpdateBillingAgreement` function should have tests in `apiClient.test.ts`, just like `apiCapturePaypalPayment`.

chenba (changes requested):
Some tests appear to be missing; I think we should not default to USD when we don't know the currency of the customer's subscriptions; a suggestion on a potentially confusing switch statement where the result might not be based on the value of the expression in the switch.

packages/fxa-payments-server/src/components/PayPalButton/index.tsx
chenba: > 
> 
> The PayPalButton Test file doesn't have any tests except `Doesn't render the PayPal button if the PayPal script fails to load`. I'm assuming this was initially skipped because of the difficulty of loading the paypal button third party scripts in a test environment. 

Oh that's right.  hmmm now that we have some branching logic here, I was hope that get tested.  😢 

> I think Barry means that the apiUpdateBillingAgreement function should have tests in apiClient.test.ts, just like apiCapturePaypalPayment.

That's a good point.  👍

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
chenba: Ah, I see.  Thanks for the info!

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
chenba: > 
> 
> Something like this perhaps? Or do you find the switch statement problematic, Barry?
> 

Thanks, that's what I had in mind.  I don't have problem with switch statements in general.  😄

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
biancadanforth, line 48: Re: [this comment](https://github.com/mozilla/fxa/pull/7909#discussion_r599185034) for QA: I was wrong on the URL not working for sandbox PayPal. If you modify it before clicking the link, i.e. to "https://www.sandbox.paypal.com/myaccount/autopay/connect/${billingAgreementId}", it does work. Not sure what happened before...

biancadanforth (changes requested):
This is looking better; thanks!

Were you able to validate this works for you? I just tested this out manually, and the `PaypalButton` is not rendered when I try this out locally on `SubscriptionCreate`. It looks like something went awry when moving the `useEffect` hook to the `PaypalButton` component; I see the script in the page, though.

Screenshot_2021-03-26 Firefox Accounts

Also, if I sign up with a PayPal subscription and cancel my BA in PayPal and go back to the Subscriptions page, the page looks like there's no error, and the expected error properties/values on the `customer` object are incorrect. I.e. There's still a `billing_agreement_id` with the old value, and `paypal_payment_error` is `undefined`. I'm guessing this is a back-end issue (cc @bbangert ), though there may be more than one thing going on here. May be worth pairing with Ben next week to try and get the integration working?

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: @biancadanforth They were just about the same, but without being wrapped in a function.

```
const stripeActionButton = (
    <button
      data-testid="reveal-payment-update-button"
      className="settings-button"
      onClick={onRevealUpdateClick}
    >
      <Localized id="pay-update-change-btn">
        <span className="change-button" data-testid="change-button">
          Change
        </span>
      </Localized>
    </button>
  );
```

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Thanks for clarifying @chenba @biancadanforth :)

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  it('renders paypalMissingAgreementActionButton when paypal is payment provider and paypal_payment_error is set to "missing_agreement"', async () => {
+    const revealFixPaymentModal = jest.fn();
+    render(
+      <Subject
+        revealFixPaymentModal={revealFixPaymentModal}
+        customer={{
+          ...CUSTOMER,
+          payment_provider: 'paypal',
+          paypal_payment_error: 'missing_agreement',
+          billing_agreement_id: '',
```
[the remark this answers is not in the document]
meandavejustice: I think you are right, but let me check with @bbangert on the correct behavior on the backend with the ids

meandavejustice:
> I think we should not default to USD when we don't know the currency of the customer's subscriptions

Do you have a suggestion for what do to in place of this fallback? Should we not allow PayPal payments in this case?
@chenba

chenba:
> 
> 
> > I think we should not default to USD when we don't know the currency of the customer's subscriptions
> 
> Do you have a suggestion for what do to in place of this fallback? Should we not allow PayPal payments in this case?
> @chenba

I think we should show the generic error modal.  If we cannot get the plan for an existing subscription that seems severe enough for that.  There isn't anything the user can do to fix the problem.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
chenba: @biancadanforth  If that's the case, then we shouldn't hard coed the `href` here correct?  We'd like QA to use the sandbox when they are testing on stage.

biancadanforth: Yes if there's a good place to set the domain here based on whether we're in dev/stage (https://www.sandbox.paypal.com) or prod (https://www.paypal.com), that'd be a good change.

biancadanforth:
> Also, if I sign up with a PayPal subscription and cancel my BA in PayPal and go back to the Subscriptions page, the page looks like there's no error, and the expected error properties/values on the customer object are incorrect. I.e. There's still a billing_agreement_id with the old value, and paypal_payment_error is undefined. I'm guessing this is a back-end issue (cc @bbangert ), though there may be more than one thing going on here. May be worth pairing with Ben next week to try and get the integration working?

Talking with Ben, there is a bug in the back-end right now where we don't watch for the Stripe customer update event to update the cached copy of the customer that we use. That will be addressed among some other back-end fixes in PR #8008 .

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Looks like we could use `apiUrl` https://github.com/mozilla/fxa/blob/351fa2b439c323ef5199efcf78790b3aee8a4a2e/packages/fxa-payments-server/server/config/index.js#L355

packages/fxa-auth-server/test/local/routes/subscriptions/stripe.js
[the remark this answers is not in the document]
meandavejustice: Added tests for `apiUpdateBillingAgreement` 👍

packages/fxa-payments-server/src/components/PayPalButton/index.tsx
meandavejustice: Tests added for `apiUpdateBillingAgreement`

packages/fxa-payments-server/src/lib/amplitude.ts
```diff
@@ -182,24 +182,24 @@ export function updatePaymentEngaged() {
   );
 }
 
-export function updatePayment_PENDING() {
+export function updatePayment_PENDING(eventProperties: EventProperties) {
```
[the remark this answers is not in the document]
biancadanforth: Dave and I chatted about this synchronously, and we agreed to just use the existing `updateDefaultPaymentMethod_*` event with the new `paymentProvider` property. We will add more nuance here in a follow-up ticket, #7883 .

meandavejustice:
@biancadanforth @chenba 

I've requested your reviews prematurely, I'm still working on fixing the rendering of the paypal button in `SubscriptionCreate/index` and showing the generic error if plan or currency are unavailable.

biancadanforth:
> > > I think we should not default to USD when we don't know the currency of the customer's subscriptions
> > 
> > 
> > Do you have a suggestion for what do to in place of this fallback? Should we not allow PayPal payments in this case?
> > @chenba
> 
> I think we should show the generic error modal. If we cannot get the plan for an existing subscription that seems severe enough for that. There isn't anything the user can do to fix the problem.

@chenba I'm not following this entirely -- when might the plan or currency be unavailable? When would we expect this error modal -- after the customer clicks the "Change" or "Manage" button in the "Billing Information" section, when the Subscription Management page loads,...?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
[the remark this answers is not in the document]
chenba: @biancadanforth Responding to https://github.com/mozilla/fxa/pull/7909#issuecomment-810391433 here because it helps with context.

Yes, the plan wont' be null as far as I can tell.  I mainly worry about us defaulting USD.  Do you think asserting the Plan is not null is enough?

> When would we expect this error modal -- after the customer clicks the "Change" or "Manage" button in the "Billing Information" section, when the Subscription Management page loads,...?

It'd be after the customer click the button.  (But I don't know if the subscription management page would load without the plans.  I have not verify that.)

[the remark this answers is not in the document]
chenba: As long as we don't send a request with a (potentially) mismatching currency, I'm open to handling this in any way that others think is more appropriate.

biancadanforth (changes requested):
Just a mini-review: some of the subscription stories are broken on this branch relative to `main` (e.g. "subscribed" and "subscribed with upgrade offer"). Also a question inline about swapping between sandbox.paypal.com and paypal.com for the billing agreement URL.

meandavejustice:
@biancadanforth Paypal button is now rendering correctly in `CreateSubscription` view.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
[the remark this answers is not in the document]
biancadanforth: > It'd be after the customer click the button. (But I don't know if the subscription management page would load without the plans. I have not verify that.)

Thanks for clarifying. I tried this out, and we get the "Plan not found" modal when loading the Subscription Management page, so this is not a scenario we should get to.

I also found an existing bug where clicking the X button on that modal just reloads the same page and thus same error modal... I'll file it.

biancadanforth (approved):
Edit: There is one follow-up here which I mention in-line (to add this new environment variable for the PayPal API URL).

Thanks Dave; this is looking pretty good to me. It'd be nice to test the full integration with the back-end, but that's waiting on #8008 . AFAICT the front-end works as expected now.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+        customer={{
+          ...CUSTOMER,
+          payment_provider: 'paypal',
+        }}
+      />
+    );
+    const revealButton = screen.getByTestId('change-payment-update-button');
+    expect(revealButton).toBeInTheDocument();
+    expect(revealButton.getAttribute('href')).toEqual(
+      `${apiUrl}/myaccount/autopay/connect/${CUSTOMER.billing_agreement_id}`
```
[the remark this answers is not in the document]
chenba: > 
> In that case we should file a Bugzilla bug for setting this value in prod. 

We need a value for that regardless since the default is the sandbox...?  Maybe I understand why this patch is the reason for adding that env var for prod.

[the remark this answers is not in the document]
meandavejustice: I will add to the Deploy Doc

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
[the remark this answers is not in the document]
biancadanforth: > Yes, the plan wont' be null as far as I can tell.

I'm actually not going to file a bug for plan being `null`, now that I re-read this thread and think about it some more. It shouldn't be `null`. I'd rather not try and overthink a scenario that may not even be possible to hit in practice.

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const stripeActionButton = () => (
+    <button
+      data-testid="reveal-payment-update-button"
+      className="settings-button"
+      onClick={onRevealUpdateClick}
+    >
+      <Localized id="pay-update-change-btn">
+        <span className="change-button" data-testid="change-button">
+          Change
+        </span>
```
biancadanforth, line 43: Can you adjust the font style/size/hover effect of these Change/Modify buttons? E.g. I just noticed the font color turns blue on hover instead of the background turning a darker grey (and the font looks a little smaller) compared to the "Cancel" button below.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -102,17 +155,17 @@ export const PaymentUpdateForm = ({
...
-
   // https://github.com/iamkun/dayjs/issues/639
   const expirationDate = dayjs()
     .set('month', Number(exp_month) - 1)
     .set('year', Number(exp_year))
     .format('MMMM YYYY');
 
+  const currencyCode = plan && plan.currency ? plan.currency : 'UNKNOWN';
+  const planId = plan && plan.plan_id ? plan.plan_id : 'UNKNOWN';
+
```
chenba, line 168: @biancadanforth already verified that the customer won't see the subscriptions management page without the plans, so we are now discussing handling potential coding errors... if an error lead to a missing plan or `plan_id` or `currency`, we should show an error.  Setting these defaults will only delay an unrecoverable error for the user.  We already know there's an error at this point; let's handle it here.

chenba (approved):
👍

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.tsx
meandavejustice: Updated font colors, the size and other things are correct to the design, the other elements of the page will be updated in paypal phase 2

packages/fxa-payments-server/src/routes/Subscriptions/ActionButton.test.tsx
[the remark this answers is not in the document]
biancadanforth: I opened this PR to update the environment variable in prod: https://github.com/mozilla-services/cloudops-infra/pull/2993 .

PR #7921, opened 2021-03-22
fix(l10n): Fix profile-heading l10n label typo.

vbudhram:
Noticed this while working on other PRs.

### DIFF
--- a/packages/fxa-settings/src/components/Profile/en-US.ftl
+++ b/packages/fxa-settings/src/components/Profile/en-US.ftl
@@ -1,6 +1,6 @@
 ## Profile section
 
-porfile-heading = Profile
+profile-heading = Profile
 profile-display-name =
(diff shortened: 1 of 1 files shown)


dannycoates (approved):
should we also PR against fxa-content-server-l10n so we don't trigger a string change?

clouserw:
Yes, that would be appreciated.  A global grep/sed should be quick and will be picked up correctly by pontoon.  /cc @peiying2

vbudhram:
> A global grep/sed should be quick and will be picked up correctly by pontoon

Sounds good, I'll make that PR.

vbudhram:
> Sounds good, I'll make that PR.

PR up in l10n repo https://github.com/mozilla/fxa-content-server-l10n/pull/484

Issue #7705, opened 2021-03-03
[FxA-Settings] The picture avatar is black when setting an image taken with the camera on iOS


data-sync-user:
**Build**
 Stage [Train 1.201.1]

**Affected platforms**
 iOS 14.4

**Steps to reproduce**

1. Navigate to [https://accounts.stage.mozaws.net/beta/settings/](https://accounts.stage.mozaws.net/beta/settings/) and login using an existing account.
1. Tap on the "Add" button displayed on the right side of the Picture section.
1. Tap on the "Add photo" button.
1. Tap on the "Take photo" button.
1. Take the photo and tap on the "Use photo" option.
1. Tap on the "Save" button.

**Expected results**
 The picture captured with the camera is successfully displayed.

**Actual Results**
 The picture avatar is black.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3123)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15505/gif21.gif">gif21.gif</a>
┆Issue Number: FXA-3123

data-sync-user:
➤ Chelsea Lewis commented:

[~vtamas@mozilla.com] [~rbillings@mozilla.com] this has been added to next train (Fxa 203) as we do not see it as a blocker for release.

dannycoates:
This is because we're creating too big of a canvas here and ios nopes out: https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/lib/canvas-utils.tsx#L37

We should refactor this function to be more efficient

vbudhram:
> This is because we're creating too big of a canvas here and ios nopes out

Yea this is exactly what is going on, if you `maxSize/2` , then the picture gets uploaded properly. Seems like we could use something like https://github.com/jhildenbiddle/canvas-size, but maybe that is a bit overkill?

I'm not too familiar with this code, so a little uncertain how to make it more efficient.

vbudhram:
Looks like there are a few issues with Safair when using the camera and cropping in general. It doesn't crop correctly on desktop.

<img width="601" alt="Screen Shot 2021-03-16 at 1 37 24 PM" src="https://user-images.githubusercontent.com/1295288/111354715-fa563900-865c-11eb-961d-1df37660d6c0.png">

New settings uses `react-easy-crop` library but I haven't narrowed down the cause. @meandavejustice do you recall any issues when working on avatar upload or have any pointers to tackle this bug?

[...]

Issue #7555, opened 2021-02-17
[FxA-Settings] Uploaded picture is not centered in the avatar frame


data-sync-user:
**Build**
Stage [Train 1.200.0]

**Affected platforms**
Windows 10 64-bit
iOS 13.6
Android 9

**Steps to reproduce**

1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
1. Click on the "Add" button displayed on the right side of the Picture section.
1. Click on the "Add photo" button.
1. Select a large image.

**Expected results**
The uploaded picture fits in the avatar frame.

**Actual Results**
The uploaded picture is displayed outside the Profile picture form.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3052)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15479/gif28.gif">gif28.gif</a> | <a href="https://jira.mozilla.com/secure/attachment/15480/p03zx3tw.jpg">p03zx3tw.jpg</a> | <a href="https://jira.mozilla.com/secure/attachment/15468/p03zx3tw+%281%29.jpg">p03zx3tw (1).jpg</a> | <a href="https://jira.mozilla.com/secure/attachment/15469/Screenshot_11.png">Screenshot_11.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15503/Screenshot_29.png">Screenshot_29.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15328/Screenshot_53.png">Screenshot_53.png</a>
┆Issue Number: FXA-3052

data-sync-user:
➤ Danny Coates commented:

this should be fixed by https://github.com/mozilla/fxa/pull/7537

data-sync-user:
➤ Vasilica Tamas commented:

Tested this issue on Stage  [Train 1.201.1] under Windows 10 64-bit, Android 9, iOS 14.4 and noticed that the picture now fits in the editing area but it is still not centered. Reopening.

data-sync-user:
➤ Wil Clouser commented:

Can you give us the dimensions of the image?  Does this work with a normal picture instead of a solid color?  The image size limit is 2MB so this seems like it would be rare with non-solid color images.

data-sync-user:
➤ Vasilica Tamas commented:

It is not about the size limit, but rather about the shape of the picture. Attaching a new screenshot and the used picture.

data-sync-user:
[...]

PR #7924, opened 2021-03-22
fix(settings): Fix avatar black image on iOS

vbudhram:
## Because

- Avatar would not be set correctly on iOS devices and safari

## This pull request

- Cherry-picked @dannycoates [commit](https://github.com/mozilla/fxa/pull/7924/commits/1e73748e411d77ed8c644f2f8729610ee57f0cf5) which simpifes our `getCroppedImg` logic
- Set aspect ratio to 1 on cropper since it made testing a little easier

Closes: #7555
Closes: #7705

## Checklist

To test, open iOS simulator and goto settings page. Select and set image.

### DIFF
--- a/packages/fxa-settings/src/components/PageAvatar/index.tsx
+++ b/packages/fxa-settings/src/components/PageAvatar/index.tsx
@@ -242,8 +242,8 @@ export const PageAddAvatar = (_: RouteComponentProps) => {
         />
         <RotateBtn
           onClick={() => {
-            if (rotation < 315) {
-              return setRotation(rotation + 45);
+            if (rotation < 270) {
+              return setRotation(rotation + 90);
             } else {
               return setRotation(0);
             }
--- a/packages/fxa-settings/src/components/PageAvatar/index.tsx
+++ b/packages/fxa-settings/src/components/PageAvatar/index.tsx
@@ -206,13 +206,13 @@ export const PageAddAvatar = (_: RouteComponentProps) => {
           crop={crop}
           zoom={zoom}
           rotation={rotation}
+          aspect={1}
           showGrid={false}
           cropShape="round"
           onCropChange={setCrop}
           onCropComplete={onCropComplete}
           onZoomChange={setZoom}
           disableAutomaticStylesInjection={true}
-          cropSize={{ width: 160, height: 160 }}
           style={{ containerStyle: { borderRadius: '8px' } }}
         />
       </div>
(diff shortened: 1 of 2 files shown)


vbudhram:
@6a68 Tagging you for review since this includes a fix for https://github.com/mozilla/fxa/issues/7555 because it made testing easier.

packages/fxa-settings/src/lib/canvas-utils.tsx
```diff
@@ -29,11 +29,11 @@ export async function getCroppedImg(
...
 
+  // Calculate a safe area to store image so that it isn't clipped during rotation by
+  // canvas context. The 1.5 multiplier was found during manual testing of various
+  // photos on default iOS.
   const maxSize = Math.max(image.width, image.height);
-  const safeArea = 2 * ((maxSize / 2) * Math.sqrt(2));
-
-  // set each dimensions to double largest dimension to allow for a safe area for the
-  // image to rotate in without being clipped by canvas context
+  const safeArea = 1.5 * ((maxSize / 2) * Math.sqrt(2));
```
dannycoates, line 36: hmm, 1.5 probably buys us some more headroom than 2 but since (afaict) `maxSize` is based on the source image dimensions it doesn't seem like this fixes the issue completely. 

Since we have a fixed final size with `pixelCrop` we ought to be able to set the initial canvas size to that and use the resizing variant of [drawImage](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage). _(I haven't tried this)_

vbudhram: Sure, let me give that a try 👍🏽

dannycoates: I can still get this to black out on the simulator. I've got an ugly code sketch that fixes it... _i think_

packages/fxa-settings/src/components/PageAvatar/index.tsx
```diff
@@ -242,8 +242,8 @@ export const PageAddAvatar = (_: RouteComponentProps) => {
         />
         <RotateBtn
           onClick={() => {
-            if (rotation < 315) {
-              return setRotation(rotation + 45);
+            if (rotation < 270) {
```
vbudhram, line 245: Images can only rotate 90 degrees 👍🏽

dannycoates (approved):
AMAZING! 😜

Issue #7784, opened 2021-03-11
Create email for paypal agreement missing


data-sync-user:
If there's no invoice being billed, but we discover that the billing agreement for a user has been cancelled, we should send an email to the user saying they should update their billing with our subscription management page. This email doesn't exist right now.

The email should be created, and tied into the `paypal-notifications.ts` in `handleMpCancel` to send it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3165)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15582/image-2021-03-22-09-02-52-989.png">image-2021-03-22-09-02-52-989.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15583/image-2021-03-22-09-23-04-928.png">image-2021-03-22-09-23-04-928.png</a>
┆Issue Number: FXA-3165

data-sync-user:
➤ Jessica Cook commented:

Link to reference for transactional emails: https://miro.com/app/board/o9J_lWaWaZA=/ ( https://miro.com/app/board/o9J_lWaWaZA=/ )

data-sync-user:
➤ Barry Chen commented:

[~jecook@mozilla.com] Thanks for that link.  I didn't see anything PayPal billing agreement specific in there.  Are we reusing the "Sorry, we're having trouble with your payment" email?

data-sync-user:
➤ Barry Chen commented:

None of the existing emails seem to quite match this scenario.  I'll create a new one.

data-sync-user:
➤ Barry Chen commented:

[~jecook@mozilla.com] I've attached a couple of screenshots to demo the emails as they are in my development environment.  (Ignore the right and bottom black borders of the images; not sure why Jira added those.)  The copy is similar to the expiring credit card email.  Please let m know if you want any changes.

data-sync-user:
➤ Barry Chen commented:

[~byee@mozilla.com] Are you the person to ping for the content of these emails?

data-sync-user:
➤ Bianca Danforth commented:

Should we be more specific than "as soon as possible"? Is our grace period here also 24 hours? If we cancel the subscription in the same way here as for funding source issues, I would recommend being as explicit as possible. We are evaluating error-related cancellations in FXA-3239 (though IIUC, an invalid BA would be due to the user cancelling the payments in PayPal which is presumably intentional).

data-sync-user:
➤ Barry Chen commented:

[...]

PR #7936, opened 2021-03-23
feat(emails): add email for payment method cancellation

chenba:
Because:
 - we should let a customer when their payment method for active
   subscriptions has a problem

This commit:
 - send an email on a billing agreement cancelled event from PayPal

This commit also changes the declarative email tests from a Map to an
array.  A previous patch (mistakenly) treated the tests as an array, and
this patch changes that list of tests to match.  The change to an array
allows multiple tests of the same mailer method, which we need in this
patch.

Closes: #7784

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -1588,6 +1588,57 @@ export class StripeHelper {
     };
   }
 
+  async formatSubscriptionsForEmails(customer: Readonly<Stripe.Customer>) {
+    if (!customer.subscriptions) {
+      return [];
+    }
+
+    let formattedSubscptions = [];
+
+    for (const subscription of customer.subscriptions.data) {
+      if (ACTIVE_SUBSCRIPTION_STATUSES.includes(subscription.status)) {
+        if (!subscription.plan) {
+          throw error.internalValidationError(
+            'extractSourceDetailsForEmail',
+            customer,
+            new Error(
+              `Multiple plans for a subscription not supported: ${subscription.id}`
+            )
+          );
+        }
+
+        const plan = await this.expandResource(
+          subscription.plan,
+          PLAN_RESOURCE
+        );
+        const abbrevProduct = await this.expandAbbrevProductForPlan(plan);
+
+        const {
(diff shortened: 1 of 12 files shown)


chenba:
I won't merge this until I get confirmation on the copy.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1580,6 +1580,57 @@ export class StripeHelper {
     };
   }
 
+  async formatSubscriptionsForEmails(customer: Readonly<Stripe.Customer>) {
```
chenba, line 1583: Existing code moved into a function.  Not truly new.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -144,7 +144,7 @@ const COMMON_TESTS = new Map([
 ]);
 
 // prettier-ignore
-const TESTS = new Map([
+const TESTS = [
```
chenba, line 147: In a previous patch for the expiring CC reminder emails, TESTS was treated as an array in a `for...in`; this fixes that issue by changing TESTS to an array.  This allows us to have more than one test per mailer method.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -370,7 +370,7 @@ const TESTS = new Map([
     ])],
     ['html', [
       { test: 'include', expected: configHref('subscriptionPrivacyUrl', 'subscription-payment-expired', 'subscription-privacy') },
-      { test: 'include', expected: configHref('subscriptionSettingsUrl', 'subscription-payment-expired', 'cancel-subscription', 'plan_id', 'product_id', 'uid', 'email') },
+      { test: 'include', expected: configHref('subscriptionSettingsUrl', 'subscription-payment-expired', 'update-billing', 'plan_id', 'product_id', 'uid', 'email') },
```
chenba, line 373: This change and the two below are related to the Map to array change affecting existing tests.

bbangert (approved):
LGTM!

chenba:
I've made content changes based on input from Brian.  Merging now.

Issue #7810, opened 2021-03-12
[subplat] Payment Error Message Text Color is Too Light


data-sync-user:
It's very difficult to read.  Also according to [https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=523%3A4575](https://www.figma.com/file/ejDMYaKpm5bsYL5YvhItur/Payment-Integration?node-id=523%3A4575) it should match the color of the plan details.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3179)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15560/screenshot-1.png">screenshot-1.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15561/screenshot-2.png">screenshot-2.png</a>
┆Issue Number: FXA-3179

Yogita98:
The figma design link is not accessible. Can you look into it?

data-sync-user:
➤ Barry Chen commented:

It's not just the error message.  The spinner screen and some of the text on the subscription success page have the issue.  See attached img.

data-sync-user:
➤ Barry Chen commented:

Latest screenshot is to show the inconsistency in font colors.

PR #7944, opened 2021-03-23
style(subscriptions): use the same shade of grey in checkout

chenba:
Because:
 - some text in the checkout process use a different, lighter, shade of
   grey

This commit:
 - change the grey to the same shade we've been using for checkout

Closes: #7810

### DIFF
--- a/packages/fxa-payments-server/src/styles/_mixins.scss
+++ b/packages/fxa-payments-server/src/styles/_mixins.scss
@@ -14,7 +14,6 @@
   background: $color-white;
   box-shadow: 0px 1px 4px rgba(12, 12, 13, 0.1);
   border-radius: 8px;
-  // margin: 0 16px 32px; processing
   color: rgba(12, 12, 13, 0.8);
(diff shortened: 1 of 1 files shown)


packages/fxa-payments-server/src/styles/_mixins.scss
```diff
@@ -37,7 +36,7 @@
 
   p {
     margin: 0;
-    color: $grey-40;
+    color: $grey-50;
```
biancadanforth, line 39: `$grey-50` is #737373, though Jess' spec actually specifies a couple of other hex color codes that don't match any of our grey variables. I'm guessing this specific thing happened because the comments in the spec call out Grey-40, Grey-50 and so on, but their actual hex values don't match the greys in our Photon package. CC @jess-cook03

biancadanforth (approved):
Definitely an improvement from earlier; thanks Barry. We will hopefully make the greys more consistent and more closely follow the design spec in the fast-follow.

Issue #7931, opened 2021-03-23
[FxA-Settings] Eye icon format is not consistent in the application


data-sync-user:
**Build:**
 Stage [Train 1.202.3]

**Affected platforms:**
 Windows 10
 Android 8

**Steps to reproduce:**

1. Navigate to [https://accounts.stage.mozaws.net/](https://accounts.stage.mozaws.net/beta/settings/) and create a new account.
1. At the ‘Create a Firefox Account Continue to Account Settings’ page, type in a password.
1. Observe the eye icon and compare it with the one from the ‘Change password’ page.

**Actual result:** 
 The eye icon format is different. There is no consistency.

**Expected result:** 
 The eye icon format is similar.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3234)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15481/eye+icon+difference.PNG">eye icon difference.PNG</a>
┆Issue Number: FXA-3234

data-sync-user:
➤ Vijay Budhram commented:

Updated the points for this because it wasn't as easy as replacing the svg. Had to make the hover, focus and pressed states matched.

data-sync-user:
➤ Bianca Oltean commented:

I retested the issue on Stage [Train 1.203.2] under Windows 10 and Android 8 and I can confirm that the Eye icon format is similar in the application.

PR #7952, opened 2021-03-24
fix(password): Use the same show password icon everywhere

vbudhram:
## Because

- Our show password functionality is different in settings and during login/signup flows

## This pull request

- Updates our login/signup flows to function similarly to the new settings show password
- Removes the hover, focus and pressed styles for show password button

Closes: #7931 

## Checklist

## Screenshots (Optional)

Screen Shot 2021-03-24 at 3 01 42 PM
Screen Shot 2021-03-24 at 3 01 31 PM


packages/fxa-content-server/app/images/icon-show-password-closed.svg
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="55" height="37" viewBox="-10 -7 47 38" style="stroke:#0c0c0d">
```
vbudhram, line 1: Had to use new svgs here because the input field in login behaves differently than input field in settings. The height and viewBox were adjusted as well as adding the stroke color.

packages/fxa-content-server/app/scripts/views/mixins/password-mixin.js
```diff
@@ -133,7 +133,7 @@ export default {
         .attr('type', 'text')
         .attr('autocomplete', 'off')
         .attr('autocorrect', 'off')
-        .attr('autocapitalize', 'off');
+        .attr('autocapitalize', 'none');
```
vbudhram, line 136: Tests were failing locally for this, `none` functions the same as `off`.

jaredhirsch (approved):
LGTM, thanks!

PR #7954, opened 2021-03-24
fix(settings): useMetrics needs to useRef

### LINKED ISSUE #7793: Duplicate metrics request on add 2FA
## Description

Multiple metrics requests are sent for the same `screen.settings.two-step-authentication` event when there should be only one.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3170)
┆Issue Number: FXA-3170

dannycoates:
Without the useRef the returned functions are new each time and therefore will get called for each render inside a useEffect.

fixes #7793


jaredhirsch (approved):
Great catch. :+1:

PR #7972, opened 2021-03-25
Default monogram avatar

dannycoates:
When an account has no avatar set we attempt to create a default avatar using the first letter of either display name or email address. These avatars use theme colors within the Firefox native UI and a neutral grey in web content and are served by the profile server.

To test with theme colors in Firefox change `svg.context-properties.content.enabled` to `true` in `about:config`.

### DIFF
--- a/packages/fxa-profile-server/lib/routes/avatar/default.js
+++ b/packages/fxa-profile-server/lib/routes/avatar/default.js
@@ -1,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = {
+  handler: async function (request, h) {
+    const monogram = /^[a-zA-Z0-9]$/.test(request.params.id)
+      ? request.params.id
+      : '?';
+
+    // Only Firefox understands context-* values
+    const ua = request.headers['user-agent'] || '';
+    const useContext = ua.includes('Firefox/');
+    const fill = useContext ? 'context-fill #B2B2B4' : '#B2B2B4';
+    // contex-fill-opacity doesn't work on Firefox when a fallback value is present
+    const fillOpacity = useContext ? 'context-fill-opacity' : '0.7';
+
+    return h
+      .response(
+        `<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+        <defs>
+          <mask id="monogram">
+            <rect x="0" y="0" width="100%" height="100%" fill="white"/>
+            <text
+              x="50%" y="50%"
+              dominant-baseline="central"
+              text-anchor="middle"
+              fill="black"
+              font-family='-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif'
+              font-weight="700"
+              font-size="4rem">${monogram.toUpperCase()}</text>
+          </mask>
+        </defs>
+        <circle cx="50" cy="50" r="50" fill="${fill}" fill-opacity="${fillOpacity}" mask="url(#monogram)"/>
+      </svg>`
+      )
+      .type('image/svg+xml; charset=UTF-8')
+      .header('Cache-Control', 'public, max-age=604800, immutable')
+      .header('Vary', 'user-agent');
+  },
+};
--- a/packages/fxa-profile-server/lib/routes/avatar/get.js
+++ b/packages/fxa-profile-server/lib/routes/avatar/get.js
@@ -5,19 +5,13 @@
 const Joi = require('@hapi/joi');
 
 const db = require('../../db');
-const config = require('../../config');
 const hex = require('buf').to.hex;
 const validate = require('../../validate');
 const logger = require('../../logging')('routes.avatar.get');
 const avatarShared = require('./_shared');
 
-const DEFAULT_AVATAR = {
-  avatar: avatarShared.fxaUrl(config.get('img.defaultAvatarId')),
-  avatarDefault: true,
-  id: config.get('img.defaultAvatarId'),
-};
-
-function avatarOrDefault(avatar) {
+async function avatarOrDefault(uid) {
+  const avatar = await db.getSelectedAvatar(uid);
   if (avatar) {
     return {
       avatar: avatar.url,
@@ -25,7 +19,7 @@ function avatarOrDefault(avatar) {
       id: hex(avatar.id),
     };
   }
-  return DEFAULT_AVATAR;
+  return avatarShared.DEFAULT_AVATAR;
 }
 
 module.exports = {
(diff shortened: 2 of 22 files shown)


packages/fxa-profile-server/lib/routes/avatar/default.js
```diff
@@ -0,0 +1,41 @@
...
+        `<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+        <defs>
+          <mask id="monogram">
+            <rect x="0" y="0" width="100%" height="100%" fill="white"/>
+            <text
+              x="50%" y="50%"
+              dominant-baseline="central"
+              text-anchor="middle"
+              fill="black"
+              font-family='-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif'
```
clouserw, line 29: UX asked for [inter](https://rsms.me/inter/) font.  Is that one of these?

packages/fxa-profile-server/lib/routes/avatar/default.js
```diff
@@ -0,0 +1,41 @@
...
+              font-weight="700"
+              font-size="4rem">${monogram.toUpperCase()}</text>
+          </mask>
+        </defs>
+        <circle cx="50" cy="50" r="50" fill="${fill}" fill-opacity="${fillOpacity}" mask="url(#monogram)"/>
+      </svg>`
+      )
+      .type('image/svg+xml; charset=UTF-8')
+      .header('Cache-Control', 'public, max-age=604800, immutable')
+      .header('Vary', 'user-agent');
```
clouserw, line 39: Why vary on UA?

clouserw:
Will this mean a bunch more traffic hitting the server or were these requests already going there for the old avatar?

packages/fxa-profile-server/lib/routes/avatar/default.js
dannycoates: They originally asked for system fonts which these cover. Inter isn't available everywhere this svg is used unless we embed it.

packages/fxa-profile-server/lib/routes/avatar/default.js
dannycoates: because the svg varies by user agent. using `context-fill` in other browsers breaks the svg

packages/fxa-profile-server/lib/routes/avatar/default.js
mikeconley: Yeah, I think "Inter" was only suggested if we had to render a rasterized image. Since we're able to do this as an SVG, we can rely on the user's system fonts, which is preferable in this case.

clouserw: perfect, thanks

packages/fxa-profile-server/lib/routes/avatar/default.js
clouserw: ah, makes sense. thanks

dannycoates:
> Will this mean a bunch more traffic hitting the server or were these requests already going there for the old avatar?

@jbuck and I chatted a bit about this on slack. It will increase traffic and we need to watch it to see what kind of caching we should put in front of it. The good news is that the new route is highly cacheable.

jbuck:
ELB request counts: https://docs.google.com/spreadsheets/d/1w2outtqhdjv1TKRX6VPorIKfqqvWESndCO55eJA2V0s/edit#gid=0 (Mozilla-only)

We should see how often the new profile images are used once this change has been landed and as @dannycoates suggested, consider add a CDN in front of the ELB.

@dannycoates I wonder, should we look at adding a config option to let us change between the old default profile and the new monogram avatars? or I guess we could simply roll back the profile server to get the old default profile

packages/fxa-profile-server/lib/routes/avatar/get.js
```diff
@@ -42,20 +43,8 @@ module.exports = {
...
-          var info = {
-            event: 'avatar.get',
-            uid: uid,
-          };
-          logger.info('activityEvent', info);
-          rep = h.response(result).etag(result.id);
-        }
-        return rep;
-      });
+    const avatar = await avatarOrDefault(uid);
```
vbudhram, line 46: Does it make sense for this endpoint to get access to display or email to build out the new default avatar too? 

Screen Shot 2021-03-30 at 11 05 19 AM

vbudhram (commented):
@dannycoates Looks pretty good! A couple of things I noticed while doing some local testing.

Works as expected in a couple edge states

New profile and not clicking the menu:
Screen Shot 2021-03-29 at 5 39 16 PM

Logged in but need to verify:
Screen Shot 2021-03-30 at 11 08 43 AM

Did a quick test on mobile and it didn't load the new image. Didn't test on Fenix but I imagine the same behavior. Maybe this is ok to do in a follow up.
Screen Shot 2021-03-29 at 5 46 32 PM

packages/fxa-profile-server/lib/routes/avatar/get.js
dannycoates: I was hoping to avoid having to change `/avatar` since fetching the email needs a round-trip to the auth-server :(

But it seems both this and ios at least use it... so I'm looking for a way to make it better

packages/fxa-profile-server/lib/routes/profile.js
```diff
@@ -84,11 +89,26 @@ module.exports = {
...
+        if (displayName && ALPHANUMERIC.test(displayName)) {
+          avatarUrl = `${monogramUrl}/v1/avatar/${displayName[0]}`;
+          result.avatarDefault = false;
+        } else if (ALPHANUMERIC.test(result.email)) {
+          avatarUrl = `${monogramUrl}/v1/avatar/${result.email[0]}`;
+          result.avatarDefault = false;
+        }
+        if (avatarUrl !== result.avatar) {
+          result.avatar = avatarUrl;
+          db.addAvatar(
```
dannycoates, line 105: @vbudhram I think this is a reasonable compromise for making `/avatar` return the monogram. I was hoping to not need to use the db, but the alternative means we need to fetch email from auth-server inside `/avatar`. This seems less gross. It relies on calling `/profile` once before `/avatar` to set it, but in normal cases I think that'll always be the case.

packages/fxa-profile-server/lib/routes/avatar/get.js
dannycoates: It gets the monogram treatment now

vbudhram (approved):
@dannycoates Nice work 💯! Code looks good but I did find a few things locally that could be follow ups.

* You have an option to "Remove" the avatar from settings. Maybe we disable the button for default avatars?
* The iOS avatar doesn't load correctly still, but I believe this might be something on their side

packages/fxa-profile-server/lib/routes/profile.js
vbudhram: Seems fair, I think it might be worthwhile to leave a comment about this here too.

dannycoates:
Thanks @vbudhram,

>     * You have an option to "Remove" the avatar from settings. Maybe we disable the button for default avatars?

I fixed that one and one other for if you remove something which would put you back with the old default avatar.

>     * The iOS avatar doesn't load correctly still, but I believe this might be something on their side

I see ios making the request for the monogram avatar. I wonder if it doesn't support SVGs?

```
{"code":200,"method":"get","errno":0,"path":"/v1/avatar/y","agent":"Mozilla/5.0 (iPhone; CPU OS 14_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/0.0.1  Mobile/15E148 Safari/605.1.15","t":0,"remoteAddressChain":["127.0.0.1"]}
```

vbudhram:
> I wonder if it doesn't support SVGs?

Very possible, the image url is set in the code but image fails to load. They do have a fallback avatar so maybe this is ok.

Issue #7939, opened 2021-03-23
Localize DropDownAvatarMenu component


jaredhirsch:
Split off from #7847.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3241)
┆Issue Number: FXA-3241

Issue #7938, opened 2021-03-23
Localize BentoMenu component


jaredhirsch:
Split off from #7847.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3242)
┆Issue Number: FXA-3242

PR #8006, opened 2021-03-26
fix(l10n): Localize DropDownAvatarMenu and BentoMenu component

vbudhram:
## Because

- DropDownAvatarMenu and BentoMenu  were not localized

## This pull request

- Localizes them

Closes: #7939
Closes: #7938

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_v2/display_name.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/display_name.js
@@ -17,6 +17,7 @@ const {
   getWebChannelMessageData,
   storeWebChannelMessageData,
   testElementTextEquals,
+  testElementTextInclude,
   type,
 } = FunctionalHelpers.helpersRemoteWrapped;
 const CHANGE_PROFILE_COMMAND = 'profile:change';
@@ -53,7 +54,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.DISPLAY_NAME_LABEL,
       'Test User',
       remote
@@ -80,7 +81,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.DISPLAY_NAME_LABEL,
       email,
       remote
@@ -115,7 +116,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.DISPLAY_NAME_LABEL,
       'New User',
       remote
@@ -148,7 +149,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.DISPLAY_NAME_LABEL,
       'Test User',
       remote
@@ -181,7 +182,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.DISPLAY_NAME_LABEL,
       email,
       remote
--- a/packages/fxa-settings/src/components/BentoMenu/en-US.ftl
+++ b/packages/fxa-settings/src/components/BentoMenu/en-US.ftl
@@ -1,0 +1,12 @@
+# BentoMenu component
+
+bento-menu-title = { -brand-firefox } Bento Menu
+bento-menu-firefox-title = { -brand-firefox } is tech that fights for your online privacy.
+
+bento-menu-vpn = { product-mozilla-vpn }
+bento-menu-monitor = { product-firefox-monitor }
+bento-menu-pocket = { product-pocket }
+bento-menu-firefox-desktop = { -brand-firefox } Browser for Desktop
+bento-menu-firefox-mobile = { -brand-firefox } Browser for Mobile
+
+bento-menu-made-by-mozilla = Made by { -brand-mozilla }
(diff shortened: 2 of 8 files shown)


packages/fxa-settings/src/components/BentoMenu/en-US.ftl
```diff
@@ -0,0 +1,12 @@
+# BentoMenu component
+
+bento-menu-title = Firefox Bento Menu
+bento-menu-firefox-title = Firefox is tech that fights for your online privacy.
+
+bento-menu-vpn = { -product-mozilla-vpn }
+bento-menu-monitor = { -product-firefox-monitor }
+bento-menu-pocket = { -product-pocket }
+bento-menu-firefox-desktop = Firefox Browser for Desktop
+bento-menu-firefox-mobile = Firefox Browser for Mobile
```
jaredhirsch, line 10: Hmm. These strings on lines 9 and 10 look weird to me. I can't recall another spot where mozilla websites use the word "browser" in describing firefox or fenix, but idk. 

Aside from whether these are the right strings, I'm not sure if these should be localized like "Firefox Accounts" is, where "Firefox" is treated as a brand, but "Accounts" can be translated:

```
# “Accounts” can be localized, “Firefox” must be treated as a brand.
-product-firefox-accounts = Firefox Accounts
```

or if we just want to leave both as untranslated product/brand strings, and call it a day.

@flodolo or @peiying2, do you have any guidance on the right strings to use for desktop and mobile firefox in this "bento menu"? I'm not sure if mobile here means just android or android/ios.

packages/fxa-settings/src/components/BentoMenu/en-US.ftl
```diff
@@ -0,0 +1,12 @@
...
+
+bento-menu-made-by-mozilla = Made by Mozilla
```
jaredhirsch, line 12: `Made by { -brand-mozilla }`, I think

packages/fxa-settings/src/components/DropDownAvatarMenu/en-US.ftl
```diff
@@ -0,0 +1,7 @@
+# DropDownAvatarMenu component
+
+drop-down-menu-title = Firefox Account Menu
```
jaredhirsch, line 3: Hmm, another complex one. This looks like it could be ` { -product-firefox-account } Menu`, except that one has different casing, "Firefox account".

The right answer here might be to leave the string as-is, but add a localization note above this line similar to what's in the App ftl:

```
# “Account Menu” can be localized, “Firefox” must be treated as a brand.
# This is used to refer to a user's account, e.g. "update your Firefox account ..."
```

packages/fxa-settings/src/components/DropDownAvatarMenu/en-US.ftl
```diff
@@ -0,0 +1,7 @@
...
+drop-down-menu-signed-in-as = Signed in as
```
jaredhirsch, line 4: This should have a placeholder for the user's display name / email address, because not all languages follow the English pattern of subject-verb-object sentence construction, where it's correct to construct a phrase like `"Signed in as " + username`.

You can add a little comment to the localizers as well:
```
# This string is used to show the current user's name or email in the settings page menu.
# Variables:
#   $user (String) - the user's name (or email address, if they haven't added their name to their account)
drop-down-menu-signed-in-as = Signed in as { $user }`

File: packages/fxa-settings/src/components/DropDownAvatarMenu/index.tsx [hunk 5]
```diff
@@ -81,7 +83,9 @@ export const DropDownAvatarMenu = () => {
                   <Avatar className="w-10" />
                 </div>
                 <p className="leading-5 max-w-full truncate">
-                  <span className="text-grey-400 text-xs">Signed in as</span>
+                  <Localized id="drop-down-menu-signed-in-as">
```
jaredhirsch, line 86: IIRC the right way to update this section to handle the $user variable in the string is to insert the `<span>displayName</span>` using the `vars` attribute:

```
<Localized id="drop-down-menu-signed-in-as" 
  vars={{
    user: ( <span className="font-bold block truncate"
                    data-testid="drop-down-name-or-email"
                  >
                    {displayName || primaryEmail.email}
                </span>
     )
  }}
>
  <span className="text-grey-400 text-xs">Signed in as { $user }</span>
</Localized>
```

I might be mixing up `elems` and `vars` here, but I think this is roughly how it should work. See the Localized docs for details: https://github.com/projectfluent/fluent.js/wiki/Localized

jaredhirsch (changes requested):
Few questions about the strings and one sentence construction that needs to be rearranged

packages/fxa-settings/src/components/DropDownAvatarMenu/en-US.ftl
vbudhram: Does menu need to be capitalized here? The style guide doesn't show this view, https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A982.

I'm in favor of using lower case and going with `{ -product-firefox-account } menu`. Thoughts?

vbudhram: Guess this is based on the Monitor menu,
Screen Shot 2021-03-29 at 2 05 19 PM

Maybe the right thing here is to leave as is?

peiying2: On `Firefox Account` or `Firefox Accounts`,  this is how mozorg handles it (using French as an example): https://pontoon.mozilla.org/fr/mozillaorg/en/brands.ftl/?search=account&string=210503. Whether Account is translated or not, it's locale specific but it is one of the products that can be translated. Also, for mozorg and Firefox for desktop, the brands are put together in a separate file. If we expect more brands to be added, we can consider doing that.

jaredhirsch: > Does menu need to be capitalized here? The style guide doesn't show this view, https://www.figma.com/file/Nj2WjMOXoJBiPoWvyJ946X/FxA---Manage-Settings-(Copy)?node-id=0%3A982.
> 
> I'm in favor of using lower case and going with `{ -product-firefox-account } menu`. Thoughts?

@vbudhram Sure, that sounds good. We've used sentence case in a lot of places already :+1:

jaredhirsch:
> Oh, I think there might be a little disconnect on how FxA builds our locale files. From what I recall these are the current steps
> 
>     1. Create a locale file for the component you want localized (this is done to be developer friendly)
> 
>     2. A script runs in background to combine component locales into the one `settings.ftl` file
> 
>     3. Submit the `settings.ftl` to pontoon to be localized
> 
> 
> @jaredhirsch correct me if wrong here.
> 
> This shouldn't technically be a new file for translation and would have access to all values in `settings.ftl`.

Yup, that's correct - the export just contains one big ftl file for localizers. On the source code side, we've gathered the global strings, like brand info, in the .ftl file for the App component.

packages/fxa-content-server/tests/functional/settings_v2/display_name.js
```diff
@@ -53,7 +54,7 @@ describe('display name', () => {
       selectors.SETTINGS_V2.AVATAR_DROP_DOWN_MENU.MENU_BUTTON,
       remote
     );
-    await testElementTextEquals(
+    await testElementTextInclude(
```
vbudhram, line 57: After introducing https://github.com/projectfluent/fluent.js/wiki/Localized#markup component for the email in the drop down menu, it renders out an additional character before and after the email? I haven't been able to pinpoint what might be causing this but I suppose in meantime we can use `testElementTextInclude`.

jaredhirsch (approved):
LGTM, nice work!

Issue #8007, opened 2021-03-26
Localize Recovery key section in new settings


vbudhram:
Broken out from https://github.com/mozilla/fxa/issues/7847

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3276)
┆Issue Number: FXA-3276

PR #8009, opened 2021-03-26
fix(l10n): Fix recovery key l10n issues

vbudhram:
## Because

- There are some l10n issues in the recovery key row

## This pull request

- Fixes them

Closes: #8007 

## Checklist


packages/fxa-settings/src/components/UnitRowRecoveryKey/en-US.ftl
```diff
@@ -1,6 +1,7 @@
 ## Recovery key sub-section on main Settings page
 
-rk-enabld = Enabled
```
vbudhram, line 3: I'll have to file a PR in the l10n repo to pickup this change.

jaredhirsch (approved):
LGTM, good to land once tests are passing

Issue #7940, opened 2021-03-23
Localize ModalVerifySession component


jaredhirsch:
Split off from #7847.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3240)
┆Issue Number: FXA-3240

PR #8034, opened 2021-03-29
fix(l10n): Localize ModalVerifySession component

vbudhram:
## Because

- The ModalVerifySession component was not localized

## This pull request

- Localizes that component
- I also had to add some localization to the base `Modal` component to handle the tooltip

Closes: #7940 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/Modal/en-US.ftl
+++ b/packages/fxa-settings/src/components/Modal/en-US.ftl
@@ -1,0 +1,4 @@
+## Modal
+
+modal-close-title = Close
+modal-cancel-button = Cancel
--- a/packages/fxa-settings/src/components/Modal/index.tsx
+++ b/packages/fxa-settings/src/components/Modal/index.tsx
@@ -9,6 +9,7 @@ import classNames from 'classnames';
 import Portal from 'fxa-react/components/Portal';
 import { ReactComponent as CloseIcon } from 'fxa-react/images/close.svg';
 import { Link, useLocation } from '@reach/router';
+import { Localized, useLocalization } from '@fluent/react';
 
 type ModalProps = {
   className?: string;
@@ -42,6 +43,7 @@ export const Modal = ({
   const modalInsideRef = useClickOutsideEffect<HTMLDivElement>(onDismiss);
   const tabFenceRef = useChangeFocusEffect();
   const location = useLocation();
+  const { l10n } = useLocalization();
   useEscKeydownEffect(onDismiss);
 
   return (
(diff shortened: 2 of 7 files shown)


packages/fxa-settings/src/components/Modal/index.tsx
```diff
@@ -68,7 +70,7 @@ export const Modal = ({
             <button
               data-testid="modal-dismiss"
               onClick={(event) => onDismiss()}
-              title="Close modal"
+              title={l10n.getString('modal-close-tooltip', null, 'Close modal')}
```
dannycoates, line 73: `modal-close-title`?

packages/fxa-settings/src/components/Modal/en-US.ftl
```diff
@@ -0,0 +1,4 @@
+## Modal
+
+modal-close-title = Close modal
```
dannycoates, line 3: I wonder if we can do better than "Close modal"? Do most folks know what "modal" means?

packages/fxa-settings/src/components/Modal/index.tsx
vbudhram: >.< good catch

packages/fxa-settings/src/components/ModalVerifySession/index.tsx
```diff
@@ -109,26 +112,41 @@ export const ModalVerifySession = ({
...
-          Verify your email
-        </h2>
-        <p
-          id="modal-verify-session-desc"
-          data-testid="modal-desc"
-          className="my-6 text-center"
+        <Localized id="mvs-verify-your-email">
+          <h2
+            id="modal-verify-session-header"
+            className="font-bold text-xl text-center"
```
jaredhirsch, line 118: How about adding a testid to this header?

packages/fxa-settings/src/components/ModalVerifySession/index.tsx
```diff
@@ -109,26 +112,41 @@ export const ModalVerifySession = ({
...
+            email: <span className="font-bold"></span>,
+          }}
         >
-          Please enter the verification code that was sent to{' '}
-          <span className="font-bold">{primaryEmail.email}</span> within 5
-          minutes.
-        </p>
+          <p
+            id="modal-verify-session-desc"
+            data-testid="modal-desc"
```
jaredhirsch, line 133: How about `modal-verify-session-desc`, so the testid will be unique to this modal

jaredhirsch (approved):
LGTM, two testid-related suggestions then good to land

Issue #8049, opened 2021-03-30
Functional test failures - sign_in_recovery_code.js


jaredhirsch:
Failures are `can sign-in with recovery code - sync - recovery code` and `can regenerate recovery code when low - recovery code`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3296)
┆Issue Number: FXA-3296

Issue #8023, opened 2021-03-29
[FxA-Settings] 'Recovery code not found' error received for the second recovery code


data-sync-user:
**Build**
Stage [Train 1.203.2]

**Affected platforms**
Windows 10

**Prerequisites:**
the user has the 2FA enabled and the recovery codes saved

**Steps to reproduce**

1. Launch Firefox and follow the login steps (enter valid email and password credentials)
1. Click on "Trouble entering code?" link.
1. Copy and paste the 10-digit code from the saved list and click on the "Verify" button.
1. Log out.
1. Repeat Step1 and Step 2.
1. Copy and paste a different valid and unused 10-digit code from the saved list and click on the "Verify" button.
1. Click one more time on the "Verify" button.

**Expected results**
The user is successfully logged in at Step 6.

**Actual Results**

* Nothing happens on the first click on the "Verify" button.
* "Recovery code not found" error is received at the second click.

**Additional notes:**

* This issue is not reproducible in Prod [Train 1.202.3] which means that might be a recent regression or a stage issue.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3284)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15703/gif6.gif">gif6.gif</a> | <a href="https://jira.mozilla.com/secure/attachment/15702/Screenshot_28.png">Screenshot_28.png</a>
┆Issue Number: FXA-3284

vbudhram:
Dug into this a little more and found that the following test case is [failing](https://github.com/mozilla/fxa/blob/cc595fc2b44bc1e15857c016539622a6a750bd5d/packages/fxa-content-server/tests/functional/sign_in_recovery_code.js#L136). When the user is running low on recovery codes for 2FA, we redirect them to the settings page to generate more. They are not correctly getting redirected and the page is getting stuck.

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.203.4] under Windows 10 64-bit. The last 2 recovery codes are accepted and the user is successfully logged in.

PR #8071, opened 2021-03-31
fix(codes): Fix issue where user could not login if they have low codes

vbudhram:
## Because

- The user could not complete login when they were using a recovery code and only had 2 left (ie low codes)
- I looked at replacing recovery codes inline with the login flow and it is not a straightforward fix unfortunately (correct webchannel messages need to be sent)

## This pull request

- Removes the check for low recovery codes during the login process which allows the login to complete
- When a user has low recovery codes they are also sent an email with a link to generate new ones, which I think makes the inline replacing redundant

Closes: https://github.com/mozilla/fxa/issues/8023
Closes: https://github.com/mozilla/fxa/issues/8049

## Checklist

## Screenshots (Optional)

Email received when user running low:
Screen Shot 2021-03-31 at 3 45 52 PM


jaredhirsch (approved):
LGTM, glad to be reducing the number of fiddly edge cases around recovery codes 😅

Issue #7945, opened 2021-03-24
[subplat] Add loading screen and error/success messaging for PayPal billing agreement error


data-sync-user:
FXA-2806 added some functionality to handle a couple of PayPal-specific error states. While we can largely reuse the `PaypalButton` component from Checkout, after talking with Ben and Jess a bit, it sounds like we don't want to reuse the Checkout processing/success screens (see [notes from today's Design Check-in](https://docs.google.com/document/d/1r-tDrOrpdhNupHrue6JMtUvPs6xuP8fFKy8hBxPDTVo/edit#heading=h.3mojru3h55m2)).

[~jecook@mozilla.com] is working on a spec for what we should show instead, and plans to have something by the end of this week.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3246)
┆Issue Number: FXA-3246

data-sync-user:
➤ Bianca Danforth commented:

[~jecook@mozilla.com] while it's not ideal, in the interest of time, we may want to go with an option that requires as little change as possible to the way the front-end currently works right now. For example:

  1. Click the PayPal button
  2. Go through the reauthorization flow in the PayPal window
  3. Grey loading screen with the circular loading gif in the middle used elsewhere for page loads (to avoid a new variation/layout for the Checkout error screen due to the product card complication)
  4. Whether it succeeds or fails, we just refresh the subscriptions page. If it succeeded, this page looks normal now. Else it still shows the error.

If this sounds okay to you, then we can proceed without any further design spec needed.

data-sync-user:
➤ Jessica Cook commented:

[~bdanforth@mozilla.com] [~bchen@mozilla.com]  After reviewing this I think its this an ok UX/UI solution to move forward with the steps outline above.

PR #8082, opened 2021-04-01
feat(payments): Add processing screen for PayPal invalid billing agre…

biancadanforth:
…ement modal

## Because

- If a subscriber cancels their billing agreement in PayPal, they have to reauthorize recurring payments to keep their subscription active.
- This requires [creating a new billing agreement and some other processing on the back-end](https://github.com/mozilla/fxa/blob/38834e96445ce64a33afcdd8738a96e65fe792d7/packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts#L359-L378), so we should indicate to the user that this processing is taking place and whether it completed successfully or not.

## This pull request

- Adds a processing screen after the user successfully reauthorizes payment in PayPal for a subscription in the `PaymentUpdateForm` component of the Subscription Management page.
- Refreshes the page when processing is complete.
- Updates the `AlertBar` error message for funding source problems, since that error won't clear until the next time we run the [PayPal Processor script](https://github.com/mozilla/fxa/blob/38834e96445ce64a33afcdd8738a96e65fe792d7/packages/fxa-auth-server/scripts/paypal-processor.ts) (currently 1x/hour) and fixes a comma splice and incorrect localization ID for the original message.
- Makes the font size of the "Change" `ActionButton` the same size as the other buttons on the page.

Closes: #7945 

## Checklist

## Screenshots (Optional)

Screenshot_2021-03-31 Firefox Accounts
 
For the funding source error alert message, I added an extra sentence to inform the subscriber that the error may not clear immediately after they fix the problem in PayPal:

Before:
> Invalid payment information, there is an error with your account.

After:
> Invalid payment information; there is an error with your account. This alert may take some time to clear after you successfully update your information.

@jess-cook03: Does this message seem acceptable to you? Note: I _thought_ #7903 was going to handle the overlap onto the header (CC @chenba ), but that issue is closed, so it may need to be a separate follow-up here.

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -308,8 +308,9 @@ sub-customer-error =
   .title = Problem loading customer
 sub-billing-update-success = Your billing information has been updated successfully
 sub-route-payment-modal-heading = Invalid billing information
-sub-route-payment-modal-copy = There seems to be an error with your Paypal account, we need you to take the necessary steps to resolve this payment issue.
-sub-route-invalid-payment = Invalid payment information, there is an error with your account. <div>Manage</div>
+sub-route-payment-modal-message = There seems to be an error with your { -brand-name-paypal } account, we need you to take the necessary steps to resolve this payment issue.
+sub-route-missing-billing-agreement-payment-alert = Invalid payment information; there is an error with your account. <div>Manage</div>
+sub-route-funding-source-payment-alert = Invalid payment information; there is an error with your account. This alert may take some time to clear after you successfully update your information. <div>Manage</div>
 pay-update-manage-btn = Manage
 
 ## subscription create
--- a/packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
+++ b/packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
@@ -28,6 +28,11 @@ import { ButtonBaseProps } from '../../components/PayPalButton';
 import PaymentProviderDetails from '../../components/PaymentProviderDetails';
 
 import { ActionButton } from './ActionButton';
+import { LoadingOverlay } from '../../components/LoadingOverlay';
+import {
+  PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE,
+  PAYPAL_PAYMENT_ERROR_MISSING_AGREEMENT,
+} from 'fxa-shared/subscriptions/types';
 
 const PaypalButton = React.lazy(() => import('../../components/PayPalButton'));
 
@@ -65,13 +70,19 @@ export const PaymentUpdateForm = ({
 }: PaymentUpdateFormProps) => {
   const [submitNonce, refreshSubmitNonce] = useNonce();
   const [updateRevealed, revealUpdate, hideUpdate] = useBooleanState();
-  const [inProgress, setInProgress, resetInProgress] = useBooleanState(false);
+  const [
+    stripeSubmitInProgress,
+    stripeSubmitSetInProgress,
+    stripeSubmitResetInProgress,
+  ] = useBooleanState(false);
+  const [transactionInProgress, setTransactionInProgress] = useState(false);
 
   const [
     fixPaymentModalRevealed,
     revealFixPaymentModal,
     hideFixPaymentModal,
   ] = useBooleanState(false);
+
   const [paymentError, setPaymentError] = useState<PaymentUpdateError>(
     paymentErrorInitialState
   );
@@ -79,12 +90,12 @@ export const PaymentUpdateForm = ({
   const onRevealUpdateClick = useCallback(() => {
     refreshSubmitNonce();
     revealUpdate();
-    resetInProgress();
+    stripeSubmitResetInProgress();
     resetUpdatePaymentIsSuccess();
   }, [
     refreshSubmitNonce,
     revealUpdate,
-    resetInProgress,
+    stripeSubmitResetInProgress,
     resetUpdatePaymentIsSuccess,
   ]);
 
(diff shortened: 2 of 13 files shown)


packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -101,18 +104,47 @@ export const PaymentUpdateForm = ({
...
+        Invalid payment information; there is an error with your account. This
+        alert may take some time to clear after you successfully update your
+        information. {actionButton}
+      </span>
+    </Localized>
+  );
+
+  const getPaypalErrorAlertBarContent = () => {
+    switch (paypal_payment_error) {
+      case 'missing_agreement': {
```
chenba, line 134: Ok, a suggestion.  I don't think it should block the merge, and I know you didn't originally introduce the type for `paypal_payment_error` for the frontend.

1. There's actually a type for it: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts#L47  So we could move that to fxa-shared/subscriptions and use it as the type on the frontend as well.

2. Going further, we could additionally declare these values once somewhere and not use string literals for the cases here.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -177,13 +209,17 @@ export const PaymentUpdateForm = ({
           </AlertBar>
         )}
 
-        {paypal_payment_error && (
+        {!transactionInProgress && paypal_payment_error && (
           <AlertBar className="alert alertError">
             {errorAlertBarContent}
```
chenba, line 214: Inline `getPaypalErrorAlertBarContent()` here so that it is not called until the other two conditions are true?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
```diff
@@ -66,12 +67,14 @@ export const PaymentUpdateForm = ({
   const [submitNonce, refreshSubmitNonce] = useNonce();
   const [updateRevealed, revealUpdate, hideUpdate] = useBooleanState();
   const [inProgress, setInProgress, resetInProgress] = useBooleanState(false);
+  const [transactionInProgress, setTransactionInProgress] = useState(false);
```
chenba, line 70: Rename the `inProgress` vars on line 69 to reduce confusion?

chenba (approved):
👍

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
biancadanforth: I can do 1) easily. Do you have a suggestion for where to put these values for 2)? I could put it in `fxa-shared/subscriptions/metadata.ts`... Is the idea something like:
```ts
export const PAYPAL_PAYMENT_ERROR_MISSING_AGREEMENT: PaypalPaymentError = 'missing_agreement';
export const PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE: PaypalPaymentError = 'funding_source';
```
... and then import these consts where we use string literals?

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
biancadanforth: That function doesn't create the React components unless there's a `paypal_payment_error`, but I could also avoid calling it entirely instead of every render, so I'll change it.

packages/fxa-payments-server/src/routes/Subscriptions/PaymentUpdateForm.tsx
chenba: For 2) we could create a `fxa-shared/subscriptions/constants.ts`, or, stick them in the same `types.ts` since I was thinking of defining the type with them.


```
export const PAYPAL_PAYMENT_ERROR_MISSING_AGREEMENT = 'missing_agreement';
export const PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE = 'funding_source';

export type PaypalPaymentError = typeof PAYPAL_PAYMENT_ERROR_MISSING_AGREEMENT | typeof PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE;
```

Or, you can use an object:

```
export const PaypalPaymentErrors = {
  missingAgreement: 'missing_agreement',
  fundingSoruce: 'funding_source',
} as const;
export type PaypalPaymentError = typeof PaypalPaymentErrors[keyof typeof PaypalPaymentErrors];
```

I'll leave the style choice to you.

biancadanforth:
@chenba Here's the diff for the negative scenario as we discussed: https://github.com/mozilla/fxa/compare/337051f..48ced06 . The two test cases in `PaymentUpdateForm.test.tsx` are identical except for in one case, `apiUpdateBillingAgreement` succeeds and in the other, it fails.

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -309,7 +309,8 @@ sub-customer-error =
 sub-billing-update-success = Your billing information has been updated successfully
 sub-route-payment-modal-heading = Invalid billing information
 sub-route-payment-modal-copy = There seems to be an error with your Paypal account, we need you to take the necessary steps to resolve this payment issue.
```
chenba, line 311: Request!  Let's replace 'Paypal' with `{ -brand-name-paypal }` on line 311 so we don't need to do it in another PR. 😁

biancadanforth: Good call; I copied that from before, but you're right, we're not using the placeable, and also it's not "PayPal" (two capital Ps)

Issue #8025, opened 2021-03-29
[FxA-Settings] Focus behavior for ‘Show/ Hide password’ eye is not consistent


data-sync-user:
**Build:**
Stage (Train 1.203.2)

**Affected platforms:**
Windows 10

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/ and log in using an existing account.
1. Go to the ‘Password’ section and click the ‘Change’ button.
1. Type in a password in one of the fields.
1. Unfocus the password field and then click on the eye icon’s focus area.
1. Observe the eye icon’s focus behavior and compare it with the one from the ‘Sign in’ or  ‘Create a Firefox Account’ page.

**Actual result:**
The eye icon is colored in blue when it is focused. There is no consistency.

**Expected result:** 
The eye icon focus behavior is similar.

**Note:** Also the focus border is not fully visible.
The eye icon is also colored in blue in the 'The Recovery key' Step 1 of 2 page.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3286)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15706/Eye+icon+focus+issue.PNG">Eye icon focus issue.PNG</a> | <a href="https://jira.mozilla.com/secure/attachment/15705/Eye+icon+focus+issue2.gif">Eye icon focus issue2.gif</a>
┆Issue Number: FXA-3286

data-sync-user:
➤ Wil Clouser commented:

[~dcoates@mozilla.com] I think you just patched this?

data-sync-user:
➤ Danny Coates commented:

I haven't touched this

karinabk:
@jodyheavener may Outreachy applicants take these issues? I would like to fix it

data-sync-user:
➤ Bianca Oltean commented:

The issue was verified as Fixed on Stage [1.204.2] under Windows 10 64-bit. The focus border is not colored in blue anymore.

PR #8090, opened 2021-04-01
fix(fxa-settings): make eye behavior for password consistent

karinabk:
## Because

- To make the interface consistent

## This pull request

- Makes eye icon focus behavior consistent with Sign in page

Closes: https://github.com/mozilla/fxa/issues/8025

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/InputPassword/index.tsx
+++ b/packages/fxa-settings/src/components/InputPassword/index.tsx
@@ -55,8 +55,8 @@ export const InputPassword = ({
       <button
         type="button"
         data-testid={formatDataTestId('visibility-toggle')}
-        className={`w-5 px-3 py-2 text-grey-600 focus:text-blue-500 box-content ${
-          hasContent ? '-ml-3' : 'hidden'
+        className={`px-3 py-2 text-grey-600 box-content ${
+          !hasContent && 'hidden'
         }`}
         tabIndex={-1}
         onClick={() => {
--- a/packages/fxa-settings/src/components/InputText/index.tsx
+++ b/packages/fxa-settings/src/components/InputText/index.tsx
@@ -83,7 +83,7 @@ export const InputText = ({
       ${disabled ? 'border-grey-100 bg-grey-10' : 'bg-white'} ${className}`}
       data-testid={formatDataTestId('input-container')}
     >
-      <span className="block relative flex-auto">
+      <span className="block flex-auto">
         <span
           className={classNames(
             'px-3 w-full cursor-text absolute text-sm origin-top-left transition-all duration-100 ease-in-out truncate font-body',


packages/fxa-settings/src/components/InputPassword/index.tsx
```diff
@@ -55,7 +55,7 @@ export const InputPassword = ({
       <button
         type="button"
         data-testid={formatDataTestId('visibility-toggle')}
-        className={`w-5 px-3 py-2 text-grey-600 focus:text-blue-500 box-content ${
+        className={`w-5 px-3 py-2 text-grey-600 focus:outline-none box-content ${
```
LZoog, line 58: @karinabk This definitely does what the ticket asked ("make them consistent") since I'm not seeing a focus state on the sign up page password input, but it's actually slightly better for accessibility if we _do_ allow a focus state to show. I think if we let the default focus state show, the fix for the ticket will be satisfied since it will no longer be blue (which is very different inconsistent from the sign up page) but will also be better for a11y.

I notice if we simply remove the `focus:text-blue-500` class, the left side of the default focus outline is not shown:
image

It would be ideal if we could get it to look like this, I think there's one or two different ways to make this happen:
image

Let me know if you need help/clarification! 🙂

LZoog (changes requested):

packages/fxa-settings/src/components/InputPassword/index.tsx
karinabk: I see. I didn't remove it because on the 'Sign-in' page there is no outline, while on the 'Change password' page the default outline is too noticeable. However, I agree that with outline is better.
Screenshot from 2021-04-01 22-10-46

Should we leave the default outline anyway?
I also noticed that in the Recovery key step 1 (when the user needs to enter the password) eye icon is also not consistent. Do I need to fix it in this PR?

LZoog: Totally makes sense, you did exactly what the ticket asked!

Out of curiosity, what OS / browser are you using? I do think we should leave the default outline for now - we can likely address improving how it looks across `fxa-settings` in another issue because I imagine if it looks like that here, the focus state likely looks like that in other places on this page. Let's leave it for now.

It looks like the same component, `InputPassword`, is used across the board there (including in the Recovery key step 1), so if we fix it here, it should be fixed everywhere we use this component! Good thinking though👍

karinabk: I'm using Chrome Version 89.0.4389.90 (Official Build) on Ubuntu.
I forgot that it can be the same component :sweat_smile: In this case, why we cannot use the same component on the Sign-in page?
P.S I updated PR

LZoog: Ah that makes sense! Since a very large chunk of our users use Firefox, that's the primary one we look at. I checked out your latest push, and I'm still not seeing the outline on the left-hand side:

image

If we change the width back (I'm guessing it's at the size it is to create a larger clickable space for users using a mouse, but also users on a phone tapping), remove the negative margin on the button, and remove the adjacent span's `relative` class, that seems to work! Double check me though, I'm not sure why that `relative` class would need to be on there in the first place.

>  In this case, why we cannot use the same component on the Sign-in page?

That is a great question! We recently revamped the Settings page and it uses React. The sign-in/up page is still using Backbone which the Settings page also used until recently; I'm not sure when we'll get to it but we hope to eventually extend React over into the login flow, and at that point, they'll use the same component. This is totally not necessary to check out, but we have some related ADRs [here](https://github.com/mozilla/fxa/blob/main/docs/adr/0010-transition-fxa-from-backbone-to-react.md) and [here](https://github.com/mozilla/fxa/blob/main/docs/adr/0011-create-new-react-app-for-settings-redesign.md) about that conversion.

karinabk: Thank you for your hints! I changed PR and squashed commits

packages/fxa-settings/src/components/InputPassword/index.tsx
```diff
@@ -55,8 +55,8 @@ export const InputPassword = ({
       <button
         type="button"
         data-testid={formatDataTestId('visibility-toggle')}
-        className={`w-5 px-3 py-2 text-grey-600 focus:text-blue-500 box-content ${
-          hasContent ? '-ml-3' : 'hidden'
+        className={`w-5 px-3 py-2 text-grey-600 box-content ${
```
jodyheavener, line 58: It looks like the spacing is off on these buttons (compare the left/right whitespace)

Screen Shot 2021-04-05 at 10 19 10 AM

This appears to be due to the `w-5` constraining the width when then size of the icon wants to exceed it:

Screen Shot 2021-04-05 at 10 22 31 AM

Any reason we can't remove this explicit width and let it auto size? cc @LZoog

karinabk: I agree with your suggestion! thank you for pointing it out. What do you think? @LZoog

LZoog: That sounds good to me! I didn't realize that, so nice call out Jody.

karinabk: Good! I committed the changes

LZoog (approved):
Thanks for all the back and forth here! When I said "it looks like we can remove the `relative`class, but double check me" (since I only looked at the DOM), I actually didn't realize I was suggesting that we'd be taking off `relative` in a component that's used in several places (`InputText`), but I decided to check out those places where it's used, and I see no visual problems with the label or anything around it across our pages when I manually test. So, I'm going to call this good to go. 👍

Issue #8092, opened 2021-04-01
[FxA-Settings] "Replacing your recovery codes" 2FA link opens an incorrect page


data-sync-user:
**Build:**
Stage [Train 1.203.4]

**Affected platforms:**
Windows 10 64-bit
Android 9

**Prerequisites:**
The user has the 2FA enabled

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/settings/ and login using an existing account.
1. Click on the "Disable" button from the "Two-step authentication" section.
1. Click on the "replacing your recovery codes".

**Expected results:**
The user is redirected to a page where are described steps to change the recovery codes.

**Actual Results:**
The user is redirected to "Reset your Firefox Account password with Recovery Keys" [page](https://support.mozilla.org/en-US/kb/reset-your-firefox-account-password-recovery-keys), which is specific to Recovery key not 2FA codes.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3301)
┆Issue Number: FXA-3301

Yogita98:
I'm an Outreachy applicant and I'd like to fix this issue if that's okay @LZoog @jodyheavener 
Well, I have done a bit of homework before giving this a shot :p
I tried to reproduce the issue locally and I can see that the href link for "replacing your recovery codes" has explicitly been set as [this](https://support.mozilla.org/en-US/kb/reset-your-firefox-account-password-recovery-keys). So does this issue only requires the link to be changed? If yes, can I also know where are we supposed to redirect the user?

LZoog:
Hey @Yogita98, yes that is correct! Thanks for looking into this.

The link _should_ be: https://support.mozilla.org/en-US/kb/changing-your-two-step-authentication-device-firefox-account

I'm running this by someone to double check before you get started and will comment back here with confirmation, but I'll go ahead and assign you, thanks for finding this one!

Yogita98:
Sure, no problem @LZoog

LZoog:
@Yogita98 got confirmation that the link provided above is correct, thanks!

Yogita98:
Alright, I will get started with this!

Yogita98:
@LZoog I'm so sorry for the inconvenience. I accidentally closed the previous PR so I'm linking another PR with required changes. 
About the previous PR, I had to hard reset my last commit maybe which is why both the commits got included in this branch but now everything seems fine except for a little mess that I caused by making two PRs

[...]

PR #8104, opened 2021-04-01
fix(Settings): Add correct 'Replacing your recovery codes' 2FA link

Yogita98:
## Because

-

## This pull request

- Adds the correct redirection link for when the user clicks on 'Replacing your recovery codes' 2FA link in the Security section of Settings page

Closes: #8092 

## Checklist

## Screenshots (Optional)
image

### DIFF
--- a/packages/fxa-admin-panel/package.json
+++ b/packages/fxa-admin-panel/package.json
@@ -85,7 +85,7 @@
     "@types/serve-static": "1.13.9",
     "@types/supertest": "^2.0.8",
     "@types/webpack": "4.41.26",
-    "@typescript-eslint/eslint-plugin": "4.10.0",
+    "@typescript-eslint/eslint-plugin": "4.20.0",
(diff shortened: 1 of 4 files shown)


LZoog:
@Yogita98, looks like another commit made it into your branch! You can remove the commit, or reset your branch to `main` and `git cherry-pick` your commit in. Either way you'll need to force-push to your branch again.

The change itself looks good! I was surprised we don't have a test for it, and I was about to suggest you could write one for some bonus points, but I just noticed this in the code:

```
            {/* "replacing recovery codes" link below will actually drop you into
            recovery codes flow in the future. */}
```

So meh on the test, just the link update is great since it may be temporary anyway. 👍

Yogita98:
Hmmm, that's strange how the commit got landed here since I already pushed it to the other branch. Nevermind, I will follow the steps that you mentioned and push the appropriate commits required for closing the issue.

Issue #8013, opened 2021-03-26
Update email timestamps to place time below date


jodyheavener:
## Description

In many of our emails we include a block of information including a date and time:

![Screen Shot 2021-03-26 at 3 31 44 PM](https://user-images.githubusercontent.com/6392049/112685569-289cfb00-8e4b-11eb-8035-5a9bbd72e26e.png)

Let's move the time to it's own line, after the date line.

This will require modifying the `location` template [partial](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/templates/partials/location.html#L6) (as well as the [text version](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/templates/partials/location.txt#L5:L5)) to add a new line to render the `time`, and maybe rename `timestamp` to `date` or similar. It'll also involve updating the mailer's usage of [`_constructLocalTimeString`](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L374:L381), as well as updating any place where a mailer [passes](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L936:L940) in the `timestamp` variable to use our new date/time variables (whatever they're named, you decide).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3280)
┆Issue Number: FXA-3280

flozender:
Hi, I'd love to take this up. May I?

jodyheavener:
@flozender sure thing, it's all yours! You may want to check out how to set up [MailDev](https://github.com/mozilla/fxa#running-with-maildev) locally for testing, and the auth server's [`write-emails`](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/package.json#L27) command may also prove helpful for writing mail templates to disk.

flozender:
I'm trying to set up the repo locally but I get thrown this error: 
```
start fxa services...

###########################################################

# fxa couldn't start. see ./artifacts/start.log for details

###########################################################

The script called "start.default" which runs "pm2 start _dev/pm2/infrastructure.config.js && _scripts/pm2-all.sh start && echo "Use 'npm stop' to stop all the servers"" failed with exit code 1 https://github.com/sezna/nps/blob/master/other/ERRORS_AND_WARNINGS.md#failed-with-exit-code
npm ERR! code ELIFECYCLE
npm ERR! errno 1
[...]

PR #8114, opened 2021-04-02
feat(fxa-auth-server): move time to a new line

flozender:
## Because

- #8013 wanted the time to have its own line below the date.

## This pull request

- Accomplishes the goal

Closes: #8013 

## Checklist

## Screenshots (Optional)

image


flozender:
I'm trying to run `yarn test` to check if any tests need to be modified but it just says `keys file already exists` and the process exits. Would appreciate any pointers here.

jodyheavener:
Hey @flozender that's odd that it's exiting for you. I also get that same message, but the tests proceed. Can you make sure that when running auth-server tests you have the process stopped in PM2? (refer to [these docs](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/README.md#testing)). Also just to clarify, you're running `yarn test` in the `fxa-auth-server` package, correct?

vbudhram (changes requested):
@flozender Thanks for PR! I think this looks good but just needs our tests updated. I guess we never checked for a timestamp so it would be good to add this.

In [test/local/senders/email.js](https://github.com/mozilla/fxa/blob/e497d1f12c017296e6b1c3fcdac2d4a7f2fc8993/packages/fxa-auth-server/test/local/senders/email.js#L95), you can add a mock value for `time` and `date` and then update the corresponding template tests to verify that they render correctly.

You should be able to follow this pattern https://github.com/mozilla/fxa/blob/e497d1f12c017296e6b1c3fcdac2d4a7f2fc8993/packages/fxa-auth-server/test/local/senders/email.js#L1110.

flozender:
Thanks, @jodyheavener, I wasn't stopping the `fxa-auth-server` so the tests weren't working. @vbudhram, I was able to add tests to make sure the date and time were getting rendered properly.

vbudhram (approved):
@flozender Thank you! This LGTM 👍🏽

@jodyheavener I'll defer merging when you are ready.

jodyheavener (approved):
This is awesome. Way to go!

Issue #7970, opened 2021-03-25
Red buttons lose their border on focus


data-sync-user:
Refer to attached video. When you click or tab into a red button it loses its border causing the content to shift. It should either keep it's border or even just make the border transparent instead of removing it.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3252)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15597/Screen+Recording+2021-03-25+at+2.01.54+PM.mov">Screen Recording 2021-03-25 at 2.01.54 PM.mov</a>
┆Issue Number: FXA-3252

Jane-Kotovich:
Hey! I'm Outreachy applicant, could I please work on this issue! Thanks!

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.204.3] under Windows 10 64-bit and Android 9. There are no UI issues when clicking/taping on the red buttons.

PR #8122, opened 2021-04-04
style (fxa-settings): change border of red buttons on focus from none…

Jane-Kotovich:
## Because

- Two red buttons in profile dashboard: Disable two-step-authentication and Delete Account would lose their borders on focus.

## This pull request

- Add transparent borders to buttons on focus so whole design on the page doesn't move
- Add dotted outline to buttons on focus

- Closes: [7970](https://github.com/mozilla/fxa/issues/7970#issue-841181979)

## Checklist

## Screenshots (Optional)

https://user-images.githubusercontent.com/76935202/113506620-3825ed80-9589-11eb-84c7-df6329dd3937.mov

https://user-images.githubusercontent.com/76935202/113654847-18eba500-96dc-11eb-85b8-b6f1c1bbbe80.mov

### DIFF
--- a/packages/fxa-settings/src/styles/ctas.scss
+++ b/packages/fxa-settings/src/styles/ctas.scss
@@ -48,7 +48,7 @@
   }
 
   &:focus {
-    @apply border-0;
+    @apply border-transparent;
   }
 
   &:disabled {
--- a/packages/fxa-settings/src/styles/ctas.scss
+++ b/packages/fxa-settings/src/styles/ctas.scss
@@ -48,7 +48,7 @@
   }
 
   &:focus {
-    @apply border-transparent outline-black-dotted;;
+    @apply border-transparent outline-black-dotted;
   }
 
   &:disabled {
--- a/packages/fxa-settings/src/styles/ctas.scss
+++ b/packages/fxa-settings/src/styles/ctas.scss
@@ -48,7 +48,7 @@
   }
 
   &:focus {
-    @apply border-transparent;
+    @apply border-transparent outline-black-dotted;;
   }
 
   &:disabled {
--- a/packages/fxa-react/configs/tailwind.js
+++ b/packages/fxa-react/configs/tailwind.js
@@ -84,6 +84,9 @@ module.exports = {
         '30': '0.3',
         '40': '0.4',
       },
+      outline: {
+        'black-dotted': '1px dotted #000',
+      },
     },
     screens: {
       mobileLandscape: '480px',


packages/fxa-settings/src/styles/ctas.scss
```diff
@@ -48,7 +48,7 @@
   }
 
   &:focus {
-    @apply border-0;
+    @apply border-transparent;
```
jodyheavener, line 51: Awesome! While we're here, since this button has almost no visual indication when it's focussed, I think we should add a dotted outline on focus, similar to the rest of the buttons. The other buttons' focus outlines work right out of the box because it inherits the text color (black). In the red button's case the text is white, so the focus outline _is_ actually there, just not visible. Let's make this focus outline black:

In the [Tailwind config](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/configs/tailwind.js), under the `extend` property extend the `outline` classes with:

```js
outline: {
  'black-dotted': '1px dotted #000',
},
```

Then you can `@apply` this new class to the `:focus` ruleste of `cta-caution`:

```
@apply border-transparent outline-black-dotted;
```

You may need to `yarn build-postcss` in `fxa-settings` to make these styles take effect, but this should give you a focus outline on the red buttons.

Jane-Kotovich:
@jodyheavener Hey Jody, all done! However, the border outline goes for the whole btn which goes out of borders and gives this square shape. Gonna work on it now in case if you would like to change it.

Jane-Kotovich:
okay was trying all day..
the issue is that for outline I cannot set the border-radius, so it outlines whole element.
As an option, I was trying to apply box-shadow  box-shadow: 0px 0px 0px 1px fff inset or box-shadow as red glowing so whole button would be glowing. However when I implemented them in tailwind it would break the code and say that I am breaking some other rules by that..
So looks like for now thats what I have achieved :)

packages/fxa-settings/src/styles/ctas.scss
```diff
@@ -48,7 +48,7 @@
   }
 
   &:focus {
-    @apply border-0;
+    @apply border-transparent outline-black-dotted;;
```
jodyheavener, line 51: Want to remove that extra semi-colon?

jodyheavener (changes requested):
This is looking great! Tested on my end and it works perfectly.

https://user-images.githubusercontent.com/6392049/113719628-d023f400-96bb-11eb-8867-c04377b244ed.mov

Just one little nitpick about the semi-colon and we're good to merge.

Jane-Kotovich:
@jodyheavener yay! All done!

jodyheavener (approved):
🥳 Nice job!

Issue #7762, opened 2021-03-09
[FXA-Settings] increase modal horizontal padding


janiceahn:
## Description

Modal padding (left, right) needs to be increased

### Steps to reproduce
<img width="491" alt="Untitled 2" src="https://user-images.githubusercontent.com/6219141/110516128-2d398380-80d7-11eb-9649-7014637c671f.png">

### Expected result

<img width="372" alt="Untitled 2" src="https://user-images.githubusercontent.com/6219141/110515871-df248000-80d6-11eb-8c64-118a3ead687d.png">

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3151)
┆Issue Number: FXA-3151

data-sync-user:
➤ Wil Clouser commented:

This is a good first issue but we need to put the actual padding values we want to use here.  I assume it is in the spec.

LZoog:
Looks like this should be 48px of padding on either side:

<img width="605" alt="image" src="https://user-images.githubusercontent.com/13018240/112683283-4f553480-8e3f-11eb-97fb-932418286c31.png">


If a contributor would like to pick this up, they can find this modal on the Settings page. Be sure to use a Tailwind class.

rustproofed:
Hello, I'm an Outreachy applicant. I would also like to fix this.

Deek9399:
Hello! I'm an Outreachy applicant, Can I work on this issue? 
Thank You.

jodyheavener:
Sure thing @Deek9399

Deek9399:
Thank you @jodyheavener

data-sync-user:
➤ Vasilica Tamas commented:

Tested this issue on Stage [Train 1.204.2] and the result differs depending on the platforms:

  * iOS 14.4: the modal is correctly displayed as it appears in the "Expected results" screenshot
  * Android 9: the text is displayed on 4 lines 
  * Windows 10 and macOS: it seems that there is no 48px of padding as was suggested above

[~lzugai@mozilla.com] Any thoughts about these results?

PR #8130, opened 2021-04-05
fixed modal padding

Deek9399:
## Because

- To fix the padding on Remove Recovery Key Modal
 
## This pull request

- Deletes the previous padding from body of recovery model.
- Adds padding to the modal container.  px-6 tablet:px-10
- Proper padding appears on all devices when checked with dev tools.

Closes: #7762

## Checklist

## Screenshots (Optional)

Screenshot from 2021-04-05 18-41-19

### DIFF
--- a/packages/fxa-settings/src/components/Modal/index.tsx
+++ b/packages/fxa-settings/src/components/Modal/index.tsx
@@ -76,7 +76,7 @@ export const Modal = ({
             </button>
           </div>
 
-          <div className="px-4 tablet:px-10 pb-10">
+          <div className="px-6 tablet:px-10 pb-10">
             <div>{children}</div>
             {hasButtons && (
               <div className="flex justify-center mx-auto mt-6 max-w-64">


packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
```diff
@@ -162,7 +162,7 @@ export const UnitRowRecoveryKey = () => {
               </h2>
             </Localized>
             <Localized id="rk-remove-modal-content">
-              <p id="recovery-key-desc" className="my-6 text-center">
+              <p id="recovery-key-desc" className="my-6 text-center px-3">
```
jodyheavener, line 165: This is only going to target the body of the Recovery Key modal, when we should make this change consistent across all modals. The other modals may look fine for you (they do for me on desktop), but they may also appear to have too little padding across different locales or screen widths. Here's an example of the 2FA modal on mobile:

Screen Shot 2021-04-05 at 10 56 51 AM


What I would recommend is applying this padding on the modal container, so it applies to all modal bodies and titles. That file exists [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/components/Modal/index.tsx#L79:L79), and I would change it to something like `px-6 tablet:px-12`, or whatever matches the designs closest.

jodyheavener (changes requested):
One suggestion for where to apply this padding for you. 😀

jodyheavener:
@Deek9399 also, you'll want to modify your commit message to be GPG signed and match the format outlined in [this doc](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md).

Deek9399:
Okay 
I will make the required changes.

Deek9399:
@jodyheavener I tried px-6 tablet:px-12, setting my gpg key. I will update it in some time.

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
jodyheavener: @Deek9399 on closer inspection it looks like only the smaller breakpoint might need changing. If you looking at this layout using `tablet:px-12` it looks like a little too much padding and makes the title break to a new line:

Screen Shot 2021-04-06 at 9 45 27 AM

Compared to `tablet:px-10`:

Screen Shot 2021-04-06 at 9 46 33 AM

My recommendation would be to just change the padding at screen sizes less than `tablet`, so basically just changing `px-4` to `px-6`. What do you think?

jodyheavener:
@Deek9399 I left you one more followup comment, but that + squashing your commits and I think we should be good to merge! 😀

packages/fxa-settings/src/components/UnitRowRecoveryKey/index.tsx
Deek9399: Yeah, keeping `px-6 tablet:px-10` gives the required padding and keeps the title in a single line for the tablet. Thanks for pointing it out, I was more concerned about the padding around text/body. It is better to keep the title in a single line with the required padding wherever possible.
This also works for the recovery modal on various screen widths.
I will fix it asap.

jodyheavener:
@Deek9399 thanks for making those changes and squashing! The only thing left is to [format](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md#git-commit-guidelines) the commit message --  sorry, I know there's a lot of things to consider in our process!

Deek9399:
@jodyheavener It's a good opportunity for me to learn. Thanks for the suggestions and corrections.

jodyheavener (approved):
Looks great! Thank you! 🥳

Issue #7973, opened 2021-03-25
Registration email Pre-header is Useless


data-sync-user:
A pre-header in an email is a great way to provide additional context. We don't seem to be doing that with our Registration emails where we just see the end of a URL:

!image-2021-03-25-11-13-49-170.png!

I'd like us to display something in the pre-header like "Copy/paste this code into your registration form"

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3254)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15600/image-2021-03-25-11-13-49-170.png">image-2021-03-25-11-13-49-170.png</a>
┆Issue Number: FXA-3254

LZoog:
Here is the attached image:
![image](https://user-images.githubusercontent.com/13018240/112680610-96412b00-8e3b-11eb-998d-5373a66e7106.png)

The link in the pre-header is probably coming from the Firefox logo. We likely want to include the suggested text as the first paragraph of the email and hide it with CSS, but there may be some caveats around that. It's probably worth searching around for the best approach here if a contributor would like to pick this one up.

claubatista:
hi, I'm outreach applicant, I would like to work on this. :)

jodyheavener:
Hey @claubatista sounds good!

claubatista:
Hey @jodyheavener, @LZoog, this is ready for review. :D

data-sync-user:
➤ Vasilica Tamas commented:

Tested this fix on Stage [Train 1.204.3] under Windows 10 64-bit and noticed that the expected pre-header "Copy/paste this code into your registration form" is displayed only for the "Read" state. For the "Unread" state appears the same pre-header as for the Sign in code email "Is this you signing in?
Firefox on Windows 10 Baia Mare, MM, Romania (estimated) IP address:....." 

[~adavis@mozilla.com] Any thoughts about this?

PR #8132, opened 2021-04-05
feat(email): add a pre-header option to email template

claubatista:
## Because

- Registration email pre-header is not very informative.

## This pull request

- Add a pre-header option to email template.
- Use the pre-header in the verify short code email. 

Closes: #7973

## Checklist

## Screenshots (Optional)

Captura de Tela 2021-04-05 às 17 45 51

Hey folks, I have some questions about the changes. 

1. I didn't change the `.txt` file, because I didn't know if it was necessary. 
2. Should I add pre-headers in other emails? 
3. I was reading in this [blog post](https://www.activecampaign.com/blog/email-preheader) about pre-headers and it says it is good if the pre-header has 40-130 characters long. 
Should I add validation around this? 

Cheers!

### DIFF
--- a/packages/fxa-auth-server/lib/senders/templates/layouts/fxa.html
+++ b/packages/fxa-auth-server/lib/senders/templates/layouts/fxa.html
@@ -7,6 +7,17 @@
   </head>
 
   <body style="-ms-text-size-adjust: 100%; -webkit-text-size-adjust: 100%; margin: 0; padding: 0;">
+    {{#if preHeader}}
+      <div style="display: none; width: 0; height: 0; max-height: 0; line-height: 0; overflow: hidden;">
+        {{preHeader}}
+
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌‌‌
+      </div>
+    {{/if}}
     <table
      align="center"
      border="0"
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -647,6 +647,7 @@ module.exports = function (log, config) {
         subject,
         supportLinkAttributes: links.supportLinkAttributes,
         supportUrl: links.supportUrl,
+        preHeader: 'Copy/paste this code into your registration form.',
       },
     });
   };


packages/fxa-auth-server/lib/senders/templates/layouts/fxa.html
```diff
@@ -7,6 +7,17 @@
...
   <body style="-ms-text-size-adjust: 100%; -webkit-text-size-adjust: 100%; margin: 0; padding: 0;">
+    {{#if preHeader}}
+      <div style="display: none; width: 0; height: 0; max-height: 0; line-height: 0; overflow: hidden;">
+        {{preHeader}}
+
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+        ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌‌‌
```
claubatista, line 18: I know this is a little strange, but I based myself in the Netflix and Stripe promotional emails that I had in my inbox.
This is necessary to show only the pre-header and not the pre-header + the body email content. 

I did some tests around how much empty space is necessary in different screen widths and this is what I arrived at.

jodyheavener: Emails are so fun.

jodyheavener (commented):
> I didn't change the .txt file, because I didn't know if it was necessary.

I would say probably not, since the preheader in a text-based email would not be visibly hidden.

> Should I add pre-headers in other emails?

For now let's just do this email. Product can decide if we want to expand this. #8148

> I was reading in this blog post about pre-headers and it says it is good if the pre-header has 40-130 characters long. Should I add validation around this?

This is interesting. I wouldn't say it's not strictly necessary, as the copy will likely have been written by the time it gets to engineering, so it would need to be communicated further up. If you wanted a fun little exercise in validating emails by all means go for it, but I wouldn't block merging for it. It would almost be better suited to a linter or test warning than a straight up test failure. I'll let you decide!

claubatista:
> This is interesting. I wouldn't say it's not strictly necessary, as the copy will likely have been written by the time it gets to engineering, so it would need to be communicated further up. If you wanted a fun little exercise in validating emails by all means go for it, but I wouldn't block merging for it. It would almost be better suited to a linter or test warning than a straight up test failure. I'll let you decide!

I was looking how to [implement a rule](https://eslint.org/docs/developer-guide/working-with-rules) in eslint and I think it would be overkill in this case.

If I were to implements a warning in the tests, where exactly do you suggest I do that? I looked in the tests and I couldn't find any tests for the email templates. So, I was a little confused.

jodyheavener:
@claubatista check out `packages/fxa-auth-server/test/local/senders/email.js` for email tests. You can probably add one [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/test/local/senders/email.js#L1224:L1251), but you'll have to dig in to see if we already have a way to assert content length. This might be tricky with how our tests are set up. Refer to [these docs](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/README.md#testing) for how to run auth-server tests.

claubatista:
Hi @jodyheavener, I looked at the tests and I think the less disruptive way to do it is using a regex. 
The regex would check how many characters there are inside the pre-header div before the spaces (assuming we include a `pre-header` class to it). This regex will only match if there is between 1 - 130 characters in this space.

```js
{ test: 'notMatch', expected: /(?:class="pre-header">\s*).{1,130}(?:\s*&nbsp;)/i },
```

The problem with this approach is we will not have a warning, but a failure. So, we could add a third propriety in the tests cases `warning: true`, and when this is true we warn instead of assert. 

IMHO, I don't think this is worth it. I noticed in my cellphone the pre-header doesn't show completely, even if it is short. 
Any number of character we put as a limit would be ad hoc at this moment, I feel like. WDYT?

jodyheavener (approved):
@claubatista Thanks for investigating this so thoroughly! I would agree with your conclusion. This LGTM. 🥳

Issue #7589, opened 2021-02-22
Update paypal subscription handler for existing paypal customer


data-sync-user:
When a customer uses an existing stripe card on file for a new subscription, we have a route handling that case. We need a similar route for paypal to use when using the paypal billing a agreement on file. For stripe users, this is the same as the initial checkout handler, it just doesn't include a payment method id. We should handle it the same with the existing paypal new subscription handler.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3069)
┆Issue Number: FXA-3069

PR #8133, opened 2021-04-05
feat(subscriptions): create subscriptions for existing PayPal customer

chenba:
Because:
 - a customer using PayPal as the payment provider should be able to
   subscribe to more than one product

This commit:
 - update the API to allow the frontend to create a new subscription
   without creating a PayPal billing agreement first

This commit maintains the same API surface and branches on the now
optional PayPal payment token request payload value to
`/oauth/subscriptions/active/new-paypal`.

Closes: #7589

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -11,6 +11,7 @@ import {
   updatePayPalBA,
 } from 'fxa-shared/db/models/auth';
 import { AbbrevPlan, AbbrevProduct } from 'fxa-shared/dist/subscriptions/types';
+import { ACTIVE_SUBSCRIPTION_STATUSES } from 'fxa-shared/subscriptions/stripe';
 import { StatsD } from 'hot-shots';
 import ioredis from 'ioredis';
 import moment from 'moment';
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -75,11 +80,55 @@ export class PayPalHandler extends StripeWebhookHandler {
...
     }
 
+    const { token } = request.payload as Record<string, string>;
+    const currentBillingAgreement = this.stripeHelper.getCustomerPaypalAgreement(
+      customer
+    );
+
+    if (!token && !currentBillingAgreement) {
+      throw error.missingPaypalPaymentToken();
+    }
```
bbangert, line 90: This isn't quite the check needed, as it will miss someone with a paypal-funded subscription active that happened to have cancelled their billing agreement. That case will require a different error, informing them that they must fix their PayPal funding for their currently active subscription before adding more subscriptions.

bbangert (changes requested):

chenba:
@bbangert Made changes to the error condition guards.  Thanks!

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -11,6 +11,7 @@ import {
   updatePayPalBA,
 } from 'fxa-shared/db/models/auth';
 import { AbbrevPlan, AbbrevProduct } from 'fxa-shared/dist/subscriptions/types';
+import { ACTIVE_SUBSCRIPTION_STATUSES } from 'fxa-shared/subscriptions/stripe';
```
bbangert, line 14: 👍🏻

bbangert (approved):
LGTM!

Issue #7702, opened 2021-03-03
[FxA-Settings] "Save" button changes its dimensions while saving a profile picture


data-sync-user:
**Build**
 Stage [Train 1.201.1]

**Affected platforms**
 iOS 14.4
 Android 9
macOS 10.14.6

**Steps to reproduce**

1. Navigate to [https://accounts.stage.mozaws.net/beta/settings/](https://accounts.stage.mozaws.net/beta/settings/) and login using an existing account.
1. Click on the "Add" button displayed on the right side of the Picture section.
1. Click on the "Add photo" button.
1. Select an image and click the "Open" button.
1. Click on the "Save" button.

**Expected results**
 "Save" button maintains its default sizes.

**Actual Results**
 "Save" button modifies its height during the saving process.

**Additional notes:**
 This issue is not reproducible on Windows 10 64-bit.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3120)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15393/gif14.gif">gif14.gif</a>
┆Issue Number: FXA-3120

karinabk:
I implemented a simple fix here. If you have a more appropriate solution, please share! @jodyheavener @LZoog 
Also, I assumed that there is a mistake in the issue description since height doesn't change

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.204.2] under Windows 10 64-bit, macOS 10.14.6, Android 9 and iOS 14.4. I confirm that the "Save" button successfully maintains its default size during the saving process.

PR #8136, opened 2021-04-05
fix(fxa-settings): make "Save" button stable

karinabk:
## Because

-"Save" button changed its size during the saving process

## This pull request

-Removes strict padding
-Adds specific width

Closes:  https://github.com/mozilla/fxa/issues/7702

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/PageAvatar/buttons.tsx
+++ b/packages/fxa-settings/src/components/PageAvatar/buttons.tsx
@@ -171,7 +171,7 @@ export const ConfirmBtns = ({
     <div className="mt-4 flex items-center justify-center">
       <Localized id="avatar-page-close-button">
         <button
-          className="cta-neutral mx-2 px-10"
+          className="cta-neutral mx-2 px-10 w-full max-w-32"
           onClick={() => navigate(HomePath, { replace: true })}
           data-testid="close-button"
         >
@@ -180,7 +180,7 @@ export const ConfirmBtns = ({
       </Localized>
       <Localized id={saveStringId}>
         <button
-          className="cta-primary mx-2 px-10"
+          className="cta-primary mx-2 px-10 w-full max-w-32"
           onClick={onSave}
           disabled={!saveEnabled}
           data-testid="save-button"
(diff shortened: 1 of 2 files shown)


packages/fxa-settings/src/components/PageAvatar/buttons.tsx
```diff
@@ -171,19 +171,25 @@ export const ConfirmBtns = ({
     <div className="mt-4 flex items-center justify-center">
       <Localized id="avatar-page-close-button">
         <button
-          className="cta-neutral mx-2 px-10"
+          className="cta-neutral mx-2"
           onClick={() => navigate(HomePath, { replace: true })}
           data-testid="close-button"
+          style={{
+            width: '25%',
+          }}
```
jodyheavener, line 179: So while this works we do like to avoid inline styles wherever possible. It also gives us a funky layout at smaller screens:

Screen Shot 2021-04-06 at 10 21 42 AM

Since these buttons are flex children they can be expanded and squeezed without breaking, so what if we allow each button to grow as wide as it can to a maximum width, something like `w-full max-w-48`. I used `max-w-48` because it appears to be the smallest max-width [we have configured](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/configs/tailwind.js#L55:L55), but it's still a little too wide. We could add `max-w-8` (2rem) for a more reasonable max-width.

jodyheavener (changes requested):
This looks great, but let's use classes over inline styles if we can.

packages/fxa-settings/src/components/PageAvatar/buttons.tsx
karinabk: Yeah, your solution is much prettier than mine. I tried to make styling flexible using percents but I need to get used to the Tailwind classes and functionalities... Thank you!
I updated the PR

jodyheavener (approved):
Looks and works well for me. Nice job 😁

karinabk:
> Looks and works well for me. Nice job

I didn't know that it is allowed to merge with failed tests :sweat_smile: I reopened PR to rerun them but it didn't help

PR #8138, opened 2021-04-05
RFC: settings refactor

dannycoates:
I noticed while working on #8055 that we do a lot of our "heavy lifting" from within the react components, which so far has been _fine_, but doing so makes them less resilient to low level changes and testing them is heavy because of how much needs to be mocked.

I present a sketch for how we could push down that work to a lower level so that the component can be lighter and more resistant to changes like #8055 in the future.

The idea is that the `Account` class exposes an api that components can easily call / mock, while internally handling the server calls, cache updates, etc. So instead of a component using apollo or authClient and passing in data from Account, it just calls a function, like `account.changePassword(...)`. All we'll need in components is `useAccount`.

This will eliminate the need for `MockedCache` in tests, instead just mocking the parts of Account in use via an `AccountContext.Provider`

As for "why now?"... I broke a lot of the component tests with #8055 and I need to fix those anyway, so I'd rather do it like this than kick the can down the road. It'll be a big diff but not much more work than is required anyway.

### DIFF
--- a/packages/fxa-settings/src/components/PageChangePassword/index.test.tsx
+++ b/packages/fxa-settings/src/components/PageChangePassword/index.test.tsx
@@ -6,7 +6,6 @@ import 'mutationobserver-shim';
 import React from 'react';
 import '@testing-library/jest-dom/extend-expect';
 import { act, fireEvent, screen, wait } from '@testing-library/react';
-import { AuthContext, createAuthClient } from '../../lib/auth';
 import { HomePath } from '../../constants';
 import { MockedCache, renderWithRouter } from '../../models/_mocks';
 import PageChangePassword from '.';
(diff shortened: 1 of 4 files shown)


vbudhram (commented):
@dannycoates This approach seems much more straightforward IMO, I say we refactor to use this pattern 👍🏽

jaredhirsch:
Big +1 to the idea of factoring out presentation code from network accesses and data transformation.

It seems like an ADR is probably warranted: we're getting into the ill-defined part of React-based apps, and I imagine a lot of people on the team will have opinions. Also, anyone who rotates onto settings work will need to understand the decision and context.

This PR seems like a good step forward, and I think it's probably fine to land this change in parallel with figuring out the long-term vision through the ADR discussion process. Even if we wind up splitting the model here into functional slices, this step is a starting point that improves short-term maintainability.

I do wonder if this Account model might be too big and too much like an "everything" global object. Rather than one huge model that acts as a generalized facade for apollo and/or the auth server, might make sense to also consider smaller 'view models' that are page- or component-based. Here's an interesting recent article on front-end architecture for Apollo React apps, by someone on the Apollo developer relations team. I'm not sure I agree with all the suggestions here, but worth considering that they suggest keeping the data layer close to the presentational layer (see the "Concern — Presentation components" section): https://www.apollographql.com/blog/apollo-client-client-side-architecture-basics/

Finally, for historical reference, Shane leaned on the approach described in the now-vintage 'scalable front-end architecture' talk, which is still a pretty interesting read: https://www.slideshare.net/nzakas/scalable-javascript-application-architecture/. (2009 omg!)

dannycoates:
> I do wonder if this Account model might be too big and too much like an "everything" global object

I agree. I think it's small enough now for a flat structure but we shouldn't be afraid to break it up or design it in a way that doing so would be hard. Conversely I've seem models get weird when things are split up prematurely. I think a single Account model is a good first step, even if it doesn't last very long.

dannycoates:
closing as "question answered"

Issue #7706, opened 2021-03-03
[FxA-Settings] FxA Bento icon does not have a focus effect


data-sync-user:
**Build**
Stage [Train 1.201.1]

**Affected platforms**
Windows 10 x64
macOS 10.14.6

**Steps to reproduce**

1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
1. Focus the Bento icon using the Tab key.

**Expected results**
Bento icon successfully displays a focus effect.

**Actual Results**
There is no focus effect displayed when the Bento icon is selected using the Tab key.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3124)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15506/gif22.gif">gif22.gif</a>
┆Issue Number: FXA-3124

sunny775:
Hi! ✋ @LZoog . I am an Outreachy applicant and would love to work on this issue as my first contribution

LZoog:
Hey @sunny775, since I just assigned another good first issue to you I'll leave this one open, but feel free to circle back here when the PR is merged for the other! :)

Here's the attached image:
![gif22](https://user-images.githubusercontent.com/13018240/113456484-f9e6cd80-93d2-11eb-9cdc-8ee41e219a5a.gif)

sunny775:
Sure @LZoog . Thanks a lot :)

flozender:
I'd love to pick this up!

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage [Train 1.204.2] under Windows 10 64-bit, Android 9 and iOS 14.4.  I confirm that the Bento icon successfully displays a focus effect while using the Tab key.

PR #8147, opened 2021-04-06
fix(fxa-settings): add focus effect to Bento icon

flozender:
## Because

- Needed to add a focus effect on the Bento icon

## This pull request

- Adds the focus effect

Closes: #7706

## Checklist

## Screenshots (Optional)

Screenshot from 2021-04-06 13-50-51

### DIFF
--- a/packages/fxa-settings/src/components/BentoMenu/index.tsx
+++ b/packages/fxa-settings/src/components/BentoMenu/index.tsx
@@ -35,7 +35,7 @@ export const BentoMenu = () => {
         title={l10n.getString('bento-menu-title')}
         aria-expanded={isRevealed}
         aria-controls={dropDownId}
-        className="rounded p-1 w-7 mx-2 border-transparent hover:bg-grey-200 focus:outline-none transition-standard desktop:mx-8"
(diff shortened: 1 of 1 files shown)


packages/fxa-settings/src/components/BentoMenu/index.tsx
```diff
@@ -35,7 +35,7 @@ export const BentoMenu = () => {
         title={l10n.getString('bento-menu-title')}
         aria-expanded={isRevealed}
         aria-controls={dropDownId}
-        className="rounded p-1 w-7 mx-2 border-transparent hover:bg-grey-200 focus:outline-none transition-standard desktop:mx-8"
```
jodyheavener, line 38: @meandavejustice it looks like you authored these classes - any reason why we shouldn't use a focus outline? Testing on my end looks good but I might be missing some context. What you think?

https://user-images.githubusercontent.com/6392049/113728269-efbf1a80-96c3-11eb-9ab2-814c14ce40b9.mov

meandavejustice: Focus outline looks good 👍 👍, Probably just copy-pasta'd the `outline:none`

jodyheavener (approved):
LGTM!

Issue #8117, opened 2021-04-02
[FxA-Settings] No focus border for Download button


data-sync-user:
**Build:**
Production (Train 1.203.4)

**Affected platforms:**
Windows 10

**Steps to reproduce:**

1. Navigate to https://accounts.firefox.com/ and login using an existing account.
1. Go to 'Recovery key' section and create a new recovery key.
1. Enter the password and click 'Continue'.
1. Click the 'Download' button.

**Actual result:** 
No focus border is displayed for the button.

**Expected result:**
Focus border is displayed.

**Note:** Compare with Copy and Print buttons.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3308)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15914/No+focus+border+for+Download+.gif">No focus border for Download .gif</a>
┆Issue Number: FXA-3308

sunny775:
Hi @LZoog 👋. This could be a "good first issue" for me to work on, as an Outreacy applicant

LZoog:
Hi @sunny775, yes this is a great one to grab! I'll assign you.

Here's the screenshot attached:
![No focus border for Download ](https://user-images.githubusercontent.com/13018240/113456381-c60ba800-93d2-11eb-89c9-d7f89bac421c.gif)

sunny775:
Thanks @LZoog . I'm on it right away

data-sync-user:
➤ Bianca Oltean commented:

This issue was verified as Fixed on Stage [1.204.2] under Windows 10 64-bit and Android 11. Focus border is displayed for 'Download' button.

PR #8149, opened 2021-04-06
fix(fxa-settings): add focus border to download button

sunny775:
## Because

- clicking on download button should display focus border, just like copy and print buttons

## This pull request

- add focus border to the recovery key download button, just like the copy and the print buttons

Closes: #8117 

## Checklist


sunny775:
Hi @LZoog . I've updated my previous commits to use only css. I used a few tailwind focus utility classes to apply uniform focus border to the three buttons. I've used a gray focus border, to make it look similar to the initial focus border of the copy and print buttons, though I think a blue forcus border should look better since the active text color  for the buttons is blue. If you feel the same way, I can adjust the border colors to blue or so.

jodyheavener:
Two thoughts:

1. I can't actually get the focus to show up on these buttons by tabbing through them? I can use Devtools and manually toggle `focus-visible` and it shows up, but I can't get it with the keyboard. Can you? Also when I am able to manually toggle it, it's not a ring, but a dotted outline. Something feels funky there. [Demo video](https://user-images.githubusercontent.com/6392049/113871347-cca67080-9780-11eb-87b5-a3418340d24e.mov)
2. I think I would prefer to us to use a dotted outline over a ring anyway since those buttons do not have defined edges. We have this pattern with the logo, sidebar, and footer links versus proper buttons with backgrounds. You may need to rebase your branch but a couple days ago the Tailwind class `outline-black-dotted` was [added](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/configs/tailwind.js#L88-L90). Maybe you could try using that here?

sunny775:
@jodyheavener thanks for that info. I will apply the `outline-black-dotted` then. 
I'm actually able to see the focus outline when I tabbed through them, but just as you said it's a dotted outline. 
Hopefully apllying the custom `outline-black-dotted` fixes the incosistencies.

Issue #7626, opened 2021-02-24
UnitRowWithAvatar should use UnitRow


jaredhirsch:
Right now, UnitRowWithAvatar duplicates the basic layout of the UnitRow, but to match the other rows, it should be rendered by passing specific props to UnitRow.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3091)
┆Issue Number: FXA-3091

LZoog:
For context, I originally created a separate component because I didn't want _too_ many conditionals in `UnitRow` that would make it confusing and thought it was better to separate it at the time since it displayed an image which was different than other rows and wasn't sure how many additional conditionals we'd need by the time the Settings Redesign project finished. If it doesn't make `UnitRow` much hairier, SGTM.

flozender:
Can I please take this up? Thanks!

LZoog:
@flozender since both of your issues have a PR open now, I'll assign this one to you. Thanks!

PR #8155, opened 2021-04-06
refactor(fxa-settings): replace UnitRowWithAvatar with UnitRow

flozender:
## Because

- Wanted to refactor `UnitRowWithAvatar` to use `UnitRow` instead. 

## This pull request

- Adds `avatar` prop and replaces `UnitRowWithAvatar` with `UnitRow` 

Closes: #7626 

## Checklist

## Screenshots (Optional)

image

~~I am yet to figure out what tests I will need to add with this PR.~~


packages/fxa-settings/src/components/UnitRow/index.tsx
```diff
@@ -58,6 +60,7 @@ export const ModalButton = ({
 };
 
 type UnitRowProps = {
+  avatar?: Account['avatar'];
```
LZoog, line 63: Nice👍

packages/fxa-settings/src/components/Profile/index.tsx
```diff
@@ -22,7 +22,13 @@ export const Profile = () => {
       </h2>
 
       <div className="bg-white tablet:rounded-xl shadow">
-        <UnitRowWithAvatar />
+        <UnitRow
```
LZoog, line 25: It looks like we're going to want to wrap this in a `<Localized>` component, similar to how we're wrapping other `<UnitRows>` in `Profile/index.tsx`. We don't want to lose the l10n ID `avatar-heading` which will need to be copied from the `UnitRowWithAvatar` FTL file, and I'd double check that "Change" (`avatar-change-link`) and "Add" (`avatar-add-link`) look like they'll be covered with the `UnitRow` FTL file.

packages/fxa-settings/src/components/Profile/index.tsx
```diff
@@ -22,7 +22,13 @@ export const Profile = () => {
...
+          header="Picture"
+          headerValue={null}
+          avatar={avatar}
+          ctaText={avatar.url ? 'Change' : 'Add'}
+          route={`${HomePath}/avatar`}
```
LZoog, line 30: It looks like we currently have `${location.search}` tacked on to the end of this in `UnitRowWithAvatar`, I'd go ahead and include that here.

LZoog (changes requested):
This is a great start! I left you some comments, and also:
* We'll need to update `UnitRow` tests, you can probably copy the tests in `UnitRowWithAvatar` into `UnitRow`, and maybe include a line on an existing test to check that the avatar is _not_ rendered
* We can nuke the entire `components/UnitRowWithAvatar` directory

packages/fxa-settings/src/components/Profile/index.tsx
flozender: We have a `${location.search}` at the end of `to` here https://github.com/mozilla/fxa/blob/21e80ce3192ddbd2b9ce6acf8ada9b5ea34f6bbe/packages/fxa-settings/src/components/UnitRow/index.tsx#L152-L156
so I didn't include it here. Should that be ok?

packages/fxa-settings/src/components/Profile/index.tsx
flozender: I tried wrapping the `UnitRow` with a `<Localized id="avatar-heading" attrs={{ header: true }}>{...}</Localized>` but it adds a `Picture` text below the avatar.
image
Should I localize this instead, since that's where I assume we'd want it:
https://github.com/mozilla/fxa/blob/21e80ce3192ddbd2b9ce6acf8ada9b5ea34f6bbe/packages/fxa-settings/src/components/UnitRow/index.tsx#L133-L135

packages/fxa-settings/src/components/Profile/index.tsx
LZoog: Oh I see, great catch!

packages/fxa-settings/src/components/Profile/index.tsx
LZoog: I see what you mean. It looks like the localization header renders as as React `children` when done that way.

I took a look at the other unit rows and noticed we have the localization in `Profile/en-US.ftl` and specify `.header` in there. I tried [this out](https://gist.github.com/LZoog/86803fe44a4b451259a3d2a766e16dcc) and it looks like it's rendering correctly that way. Try that out and let me know if it's looking good to you too!

packages/fxa-settings/src/components/UnitRowWithAvatar/index.stories.tsx
```diff
@@ -1,22 +0,0 @@
...
-
-import React from 'react';
-import { storiesOf } from '@storybook/react';
-import { LocationProvider } from '@reach/router';
-import { MockedCache } from '../../models/_mocks';
-import { UnitRowWithAvatar } from '.';
-
-storiesOf('Components|UnitRowWithAvatar', module)
-  .addDecorator((getStory) => <LocationProvider>{getStory()}</LocationProvider>)
-  .add('with default avatar', () => (
```
LZoog, line 13: Thank you for the tests 👏 

One other thing here, sorry I forgot to mention this with my first review - we use Storybook to keep track of component states here, you can pull it up by running `yarn storybook` in `fxa-settings`. `UnitRowWithAvatar` had “with default avatar” and “with non-default avatar” states. Will you add these two stories to the existing `UnitRow` stories? 🙂

packages/fxa-settings/src/components/Profile/index.tsx
flozender: Yes, this works, thank you!

packages/fxa-settings/src/components/UnitRowWithAvatar/index.stories.tsx
flozender: Done!

packages/fxa-settings/src/components/UnitRow/en-US.ftl
```diff
@@ -1,5 +1,6 @@
 ## Sub-section row Defaults
 
+avatar-heading = Picture
```
LZoog, line 3: We can remove this now.

packages/fxa-settings/src/components/UnitRow/index.stories.tsx
```diff
@@ -91,4 +93,40 @@ storiesOf('Components|UnitRow', module)
...
+          route={`${HomePath}/avatar`}
+          {...{ avatar }}
+        />
+      </MockedCache>
+    );
+  })
+  .add('with non-default avatar', () => {
+    const avatar = {
+      id: null,
+      url: 'http://localhost:1111/v1/avatar/t',
```
LZoog, line 118: Since we push Storybook builds up to this URL, it'd be more ideal to change this to something that'll work everywhere. We tend to use silly placekitten images, if you change this to `http://placekitten.com/512/512?image=0` we'll be golden. 😛

LZoog (commented):
This LGTM, just a couple of minor tweaks. When you make those, will you squash your commits into one? 🙂

The CI failures seem to be false positives, when you force-push with your single commit they'll rerun.

LZoog (approved):
LGTM, thanks! :)

PR #8156, opened 2021-04-06
 Train 203.6

jaredhirsch:
Pulling in the monogram avatar default for stage-only testing


vbudhram (approved):

vbudhram:
Merging since it is failing on a known flaky test.

Issue #7025, opened 2020-11-27
[FxA-Settings] No 2FA download/copy/print icons states displayed


vasilicatamas:
**Build**
Stage [Train 1.194.0]

**Affected platforms**
Windows 10
Android 9
iOS

**Steps to reproduce**
1. Navigate to https://accounts.stage.mozaws.net/beta/settings/ and login using an existing account.
2. Click on the "Add" button from 2FA section.
3. Enter the code received after the QR code was scanned and tap on "Continue".
4. Check the download/copy/print icons.

**Expected results**
The icons successfully display all the states - as presented in figma.
![Screenshot_52](https://user-images.githubusercontent.com/18114210/100453066-007ed480-30c3-11eb-96ce-c5a6f0c674a6.png)

**Actual Results**
The icons do not have the hover/click/tap/focus states presented in figma.

karinabk:
@LZoog @jodyheavener While reproducing this issue on Chrome I noticed that rows overlapped 
![Screenshot from 2021-04-07 16-03-09](https://user-images.githubusercontent.com/35498147/113849398-42c7c900-97bb-11eb-8472-97101299d294.png)

There is no such issue on Firefox though. If you don't mind I will work on this issue

data-sync-user:
➤ Vasilica Tamas commented:

This issue is verified as fixed on Stage  [Train 1.204.3] under Windows 10 64-bit, Android 9 and iOS 14.4. The icons successfully display the hover, click/tap and focus states.

PR #8169, opened 2021-04-07
fix(fxa-settings): add 2FA icons states

karinabk:
## Because

- 2FA icons do not have states

## This pull request

- Adds hover to the download/copy/print icons

Related: https://github.com/mozilla/fxa/issues/7025

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/GetDataTrio/index.tsx
+++ b/packages/fxa-settings/src/components/GetDataTrio/index.tsx
@@ -55,7 +55,7 @@ export const GetDataTrio = ({ value, onAction }: GetDataTrioProps) => {
           )}
           download={`${primaryEmail.email} Firefox.txt`}
           data-testid="databutton-download"
-          className="w-12 h-12 relative inline-block text-grey-500 rounded active:text-blue-500 active:outline-black-dotted focus:outline-black-dotted"
+          className="w-12 h-12 relative inline-block text-grey-500 rounded active:text-blue-500 active:outline-black-dotted focus:outline-black-dotted hover:bg-grey-50"
           onClick={() => onAction?.('download')}
         >
           <DownloadIcon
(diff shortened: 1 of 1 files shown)


karinabk:
@LZoog @jodyheavener  I updated hover here. These icons are blue in the active state, since we removed the blue state from the eye icon (in the password input component) do we need to remove it here too? Also, I am not sure how to add click/tap status. Do you have any ideas?

LZoog:
@karinabk Hmm so this issue might be a little tricky to close out entirely with all of those states. I'm not sure if there's an ideal way to style that `title` attribute to match the image posted in the issue - I think it may need to be a tooltip, which increases the complexity here.

The click/tap state is that blue `focus` state that you're seeing. I think, until we get all of the states in place, the blue is fine for now - I'd like to double check this with our Design team before proceeding. I think to get us partially there, you could change the hover state to `bg-grey-50` since that more closely matches the color in the image, and that would be a small win there.

If you can find what's causing the [Chrome bug you found](https://github.com/mozilla/fxa/issues/7025#issuecomment-814787931), that would also be a win here, though it's not necessary to get this PR merged. I know it won't close out the issue, but that's OK. You can change "Work in progress: #7025" to "Related: #7025" so the issue remains linked, but not "closed" when the pull request is merged.

P.S. GitHub has a "draft" state pull request that you can use when you're putting a WIP PR up. The "WIP" text works in the title as well, but just wanted to let you know about that functionality, it's pretty new. 🙂

karinabk:
@LZoog Oh, interesting! I didn't know about the 'draft' state feature. 

I will try to solve the Chrome bug and add the tooltips ;) Let me know if there will be an update regarding icons styling.

LZoog:
(I responded in Matrix to [this comment](https://github.com/mozilla/fxa/pull/8169#issuecomment-815477971), I don't think the tooltips should be added as part of this PR)

@karinabk It looks like you've got a merge conflict, but this is otherwise g2g! If you pull latest `main`, you can resolve the conflict by keeping the incoming change and your change.

The Circle CI failure seems to be coming from a flaky test, when you repush they will rerun.

karinabk:
@LZoog I updated the forked project and changed PR. Thanks

jodyheavener:
Is this PR still WIP?

karinabk:
@jodyheavener No, we decided to split this issue into smaller ones. Thanks for the reminder ;)

jodyheavener (approved):
LGTM :)

PR #8173, opened 2021-04-07
fxa(chore): echo firefox version when running circle ci tests

vbudhram:
This would be helpful in tracking down circle ci errors


jaredhirsch (approved):
LGTM, thanks!

Issue #8047, opened 2021-03-30
[FxA Settings] Blank Firefox Accounts page is displayed when user selects Manage devices option from Synced Tabs


data-sync-user:
**Build:**
Stage (Train 1.203.2)

**Affected platforms:**
Windows 10

**Preconditions:**
The user must be signed in to Firefox Sync with the same account on desktop and a mobile device which has a few open tabs.

**Steps to reproduce:**

1. On desktop click on the ‘Show sidebars’ button in the toolbar and select the ‘Synced Tabs’ option.
1. Right-click a synced device to trigger the context menu.
1. Select the ‘Manage devices’ option.

**Actual result:** 
Firefox Accounts page is displayed with no content.

**Expected result:** 
Firefox Accounts page is displayed with focus on Connected services.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3294)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15720/Manage+devices+-+Blank+settings+page+displayed.gif">Manage devices - Blank settings page displayed.gif</a>
┆Issue Number: FXA-3294

flozender:
@LZoog is this a valid issue? Can I please pick this up?

vbudhram:
Hey @flozender, I have an upcoming patch for this bug. You are welcome to work on another issue!

data-sync-user:
➤ Bianca Oltean commented:

The issue was verified as Fixed on Stage [1.204.3] under Windows 10 64-bit. I confirm that Firefox Accounts page is displayed with focus on Connected services.

PR #8178, opened 2021-04-07
fix(bug): Redirect to client/services page from legacy link

vbudhram:
## Because

- Open manage devices link from Firefox is broken (From within the Sync Tabs submenu)

## This pull request

- Adds a redirect from the legacy url `/settings/clients` to the new url `/settings#connected-services`

Closes: #8047

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/ScrollToTop/index.tsx
+++ b/packages/fxa-settings/src/components/ScrollToTop/index.tsx
@@ -30,7 +30,8 @@ export const ScrollToTop = (
     // to it, because reach router can't handle hashes (see reach router
     // issue 32) 🙄
     if (hasHash) {
-      const el = document.querySelector(url.hash);
+      const hashTokens = url.hash.split('?');
+      const el = document.querySelector(hashTokens[0]);
       if (el) el.scrollIntoView();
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/ScrollToTop/index.tsx
```diff
@@ -30,7 +30,8 @@ export const ScrollToTop = (
     // to it, because reach router can't handle hashes (see reach router
     // issue 32) 🙄
     if (hasHash) {
-      const el = document.querySelector(url.hash);
+      const hashTokens = url.hash.split('?');
```
dannycoates, line 33: what's this for?

vbudhram: Kind of a hack on top of a hack 😕. Using the redirect from above creates a url like

`/settings#connected-services?context=fx_desktop_v3&entrypoint=sendtab&uid=554213....`

Which was causing an error here because `url.hash` included everything after connected-services and was failing to scroll into view. Maybe an alternative is to write the url?

dannycoates (approved):
lgtm

PR #8200, opened 2021-04-08
Train 203 merge main

vbudhram:
Per slack, pulling in everything from `MAIN` into train-203 for a point release. This is for QA to test subscription updates in stage only.


jaredhirsch (approved):
Thanks, @vbudhram!

For historical reference, the current train 203 is going a week longer than usual, to synchronize with Firefox release schedule, but we need to test some stuff on staging that's already landed in `main`. The simplest way to handle this is to create another point release on 203, and this PR pulls everything into train-203 as a preliminary step to the 203.7 point release paperwork.

PR #8205, opened 2021-04-08
fix(fxa-settings): add focus border to download button

sunny775:
## Because

- clicking on download button should display focus border, just like copy and print buttons

## This pull request

- add focus border to the recovery key download button, just like the copy and the print buttons

Closes: #8117 

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/GetDataTrio/index.tsx
+++ b/packages/fxa-settings/src/components/GetDataTrio/index.tsx
@@ -55,7 +55,7 @@ export const GetDataTrio = ({ value, onAction }: GetDataTrioProps) => {
           )}
           download={`${primaryEmail.email} Firefox.txt`}
           data-testid="databutton-download"
-          className="w-12 h-12 relative inline-block text-grey-500 rounded active:text-blue-500 focus:outline-black-dotted"
+          className="w-12 h-12 relative inline-block text-grey-500 rounded active:text-blue-500 active:outline-black-dotted focus:outline-black-dotted"
           onClick={() => onAction?.('download')}
         >
           <DownloadIcon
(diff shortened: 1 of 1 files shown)


sunny775:
@LZoog  @jodyheavener Kindly review this. I unintentionally got my previous PR untidy, and had to create a new one for the same issue. I've had serious struggles trying to get the fxa repo to work locally.

LZoog:
@sunny775 Rebasing against `main` can be tricky sometimes. Thanks for commenting on your other PR and linking to this one! 😀

So it looks like this was the original issue...
image

With your changes, I'm still seeing a hint of inconsistency:
getdatatrio

I'm guessing the "download" link is looking a bit different since it's actually a _link_ while the other two are _buttons_ (for what it's worth, I double checked and [it does seem like using a link here is correct](https://css-tricks.com/building-good-download-button/)).

I think all you need to do is change the `hover` class to `active`, that is, to `active:outline-black-dotted`. We need it on the "download" link because there is no active state at all, and we need it on the other two buttons to make them look consistent with the first one.

P.S. don't forget to record this PR [through this link](https://www.outreachy.org/outreachy-may-2021-internship-round/communities/firefox/revamp-firefox-accounts-emails/contributions/)! We mentioned this in the FxA Matrix channel very recently but we're closing the contribution period soon and you need to submit at least one before then in order to edit and add any others. Thanks!

sunny775:
@LZoog Thanks so much for the kind review.  I've made and committed the suggested changes

jodyheavener (approved):
This LGTM. Thank you! 🥳 I'll let @LZoog do the merge here in case she spots anything.

LZoog (approved):
LGTM, thanks!

@sunny775 Ah oops I jumped the gun a little here (but it's OK) - I should have double checked that you only had a single commit here. In the future, please squash or fix up your 2nd commit so that there's only one commit we merge in. :) It was my bad for missing it and not pointing it out. Thanks again!

sunny775:
Thanks a lot @LZoog  @jodyheavener . 
Noted, I'll squash my commits into one in in subsequent PRs

Issue #8003, opened 2021-03-26
Functional test failures - fx_fennec_v1_settings.js (3 tests failing)


data-sync-user:
┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3275)
┆Issue Number: FXA-3275

data-sync-user:
➤ Ankita Shrivastava commented:

The failing test file is: fxfennecv1_settings.js

and tests are: ** 

  1. sign in, change the password
  2. sign in, change the password by browsing directly to settings
  3. sign in, delete the account

Issue #7976, opened 2021-03-25
Functional test failures - sync_v3_settings.js (3 tests failing)


data-sync-user:
┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3272)
┆Issue Number: FXA-3272

PR #8221, opened 2021-04-09
fix(tests): Fix sync v3 and fennec functional setting tests

vbudhram:
## Because

- These tests were borked

## This pull request

- Unborks them

Closes: #8003
Closes: #7976

## Checklist


packages/fxa-content-server/tests/functional/fx_fennec_v1_settings.js
```diff
@@ -107,15 +107,15 @@ registerSuite('Fx Fennec Sync v1 settings', {
...
         this.remote
-          .then(click(selectors.SETTINGS_DELETE_ACCOUNT.MENU_BUTTON))
+          .then(click(selectors.SETTINGS_DELETE_ACCOUNT.DELETE_ACCOUNT_BUTTON))
           .then(visibleByQSA(selectors.SETTINGS_DELETE_ACCOUNT.DETAILS))
 
           .then(fillOutDeleteAccount(FIRST_PASSWORD))
           // Fx desktop requires fxaccounts:delete, Fennec requires
           // fxaccounts:delete_account
-          .then(testIsBrowserNotified('fxaccounts:delete_account'))
+          // .then(testIsBrowserNotified('fxaccounts:delete_account'))
```
vbudhram, line 116: I'm going to file a follow up bug here. Currently our settings page only sends `fxaccounts:delete`.

vbudhram: Actually, the corresponding bug was https://github.com/mozilla/fxa-content-server/issues/3432 which no longer applies because it was specific to Fennec.

jaredhirsch:
Looks like the avatar failures here are due #7626 getting closed without updating the avatar unit row test selector. Should be an easy fix, just need to set `prefixDataTestId` where the avatar UnitRow is instantiated.

packages/fxa-content-server/tests/functional/pages.js
```diff
@@ -47,9 +47,6 @@ var pages = [
   'reset_password_verified',
   'settings',
   'settings/avatar',
-  'settings/avatar/camera',
-  'settings/avatar/change',
-  'settings/avatar/crop',
```
vbudhram, line 52: These routes don't exist anymore. I don't think we need to add redirect here because I don't believe any external apps link directly to these pages.

jaredhirsch (approved):
LGTM, great work!

PR #8227, opened 2021-04-10
fix(auth-server): fix stripe webhook errors

### LINKED ISSUE #7203: Stripe delete webhook returns 404 instead of 200
Stripe sends a webhook on customer delete, this is used by FxA to ensure that if a Stripe customer record is deleted we have removed the Stripe customer id tracking row from the AccountSubscriptions table.

The Stripe webhook logs in production indicate that we are throwing 404's on these, as customers don't exist. My guess is that this is happening during normal customer delete operations, because we delete the FxA account, and then send a delete request to Stripe, so when Stripe then notifies us the customer was deleted, they're already gone.

We should instead respond to this with a 200, so Stripe stops trying to retry, which also flags our API as 'erroring'.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2867)
┆Issue Number: FXA-2867

bbangert:
Because:

* We were rejecting Stripe payloads that exceeded 16Kb, some were as
  large as 22Kb.
* Stripe webhooks that sent emails to customers who deleted their
  account were throwing errors that Stripe keeps retrying.

This commit:

* Updates the maxBytes for the Stripe webhook to accomodate the larger
  payload sizes observed.
* Adds error catching to the email calls in the webhooks to capture and
  report errors that we expect (account deleted), but returns a proper
  200 to Stripe.

Fixes #7203

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -1486,7 +1486,13 @@ export class StripeHelper {
   }
 
   /**
-   * Extract invoice details for billing emails
+   * Extract invoice details for billing emails.
+   *
+   * Note that this function throws an error in the following cases:
+   *   - Stripe customer is deleted.
+   *   - No plan in the invoice.
+   *   - No product attached to the plan.
+   *   - No email on the customer object.
    */
   async extractInvoiceDetailsForEmail(latestInvoice: Stripe.Invoice | string) {
     const invoice = await this.expandResource(latestInvoice, INVOICES_RESOURCE);
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -16,6 +16,7 @@ import { PayPalClientError } from '../../payments/paypal-client';
 import {
   PAYPAL_BILLING_AGREEMENT_INVALID,
   PAYPAL_SOURCE_ERRORS,
+  PAYPAL_BILLING_TRANSACTION_WRONG_ACCOUNT,
 } from '../../payments/paypal-error-codes';
 import { StripeHelper, SUBSCRIPTION_UPDATE_TYPES } from '../../payments/stripe';
 import { AuthLogger, AuthRequest } from '../../types';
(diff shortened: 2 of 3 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1486,7 +1486,13 @@ export class StripeHelper {
   }
 
   /**
-   * Extract invoice details for billing emails
+   * Extract invoice details for billing emails.
+   *
+   * Note that this function throws an error in the following cases:
+   *   - Stripe customer is deleted.
+   *   - No plan in the invoice.
+   *   - No product attached to the plan.
```
chenba, line 1494: ??? Does Stripe even allow this?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -240,7 +249,19 @@ export class StripeWebhookHandler extends StripeHandler {
...
   ) {
     const sub = event.data.object as Stripe.Subscription;
-    const { uid, email } = await this.sendSubscriptionDeletedEmail(sub);
+    let uid, email;
+    try {
+      ({ uid, email } = await this.sendSubscriptionDeletedEmail(sub));
+    } catch (err) {
+      // FIXME: If the customer was deleted, we don't send an email that their subscription
+      //        was cancelled. This is because the email requires a bunch of details that
+      //        only exist on non-deleted customer. A more robust solution is needed.
```
chenba, line 258: Is this already captured in a Jira?

chenba (approved):

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: Nope, filing an issue for it.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: I didn't think they do, but its in the code as a type guard check which means the Stripe API types believe it could occur.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: Filed FXA-3356 for this.

Issue #7978, opened 2021-03-25
Functional test failures - fx_desktop_handshake.js (4 tests failing)


data-sync-user:
┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3270)
┆Issue Number: FXA-3270

data-sync-user:
➤ Ankita Shrivastava commented:

the failing test file is: fxdesktophandshake.js

and tests are:

  1. Sync settings page - user signed into browser**
  2. NonSync settings page  user signed into browser
  3. Sync settings page - no user signed into browser
  4. NonSync settings page  no user signed into browser, no user signed in locally

Issue #8244, opened 2021-04-12
Support `fxa_status` webchannel message in new settings


vbudhram:
While working through https://jira.mozilla.com/browse/FXA-3270, the failing tests were associated with us not requesting and updating the signed in account based on the `fxa_status` message. We should add support for this and will eventually need to when we migrate on login pages to React.

* Update https://github.com/mozilla/fxa/blob/e220d692ad8abc0fc62647eaa524da2bc21513c2/packages/fxa-settings/src/lib/firefox.ts#L4, to support `fxaccounts:fxa_status` message
* Message is only supported in FF > 58
* Update/merge the local account with the response and store in local storage

Ref: https://github.com/mozilla/fxa/blob/14b0c2960f517f45786ab2859cbfdc66f1799019/packages/fxa-content-server/app/scripts/models/auth_brokers/base.js#L195

vbudhram:
Fixing this issue would also fix this [test](https://github.com/mozilla/fxa/issues/7981)

PR #8245, opened 2021-04-12
fix(tests): Fix firefox handshake functional tests

vbudhram:
## Because

- These tests were disabled after migrating to new settings

## This pull request

- Re-enables most of them however the others are broken because of https://github.com/mozilla/fxa/issues/8244

Closes: #7978 

## Checklist


jaredhirsch (approved):
Good idea. I'll probably take this same approach, so we can land the oauth_webchannel test fixes (2 of the 3 were easy, the third needs a webchannel message changed)

PR #8246, opened 2021-04-12
fix(tests): Make fxa-settings and fxa-content-server tests dependent on each other

vbudhram:
## Because

- A lot of the sign-in/sign-up functional tests go through various flows in new settings so we should ensure that both suites of tests run when a change occurs in either project

## This pull request

- Makes fxa-settings and fxa-content-server tests dependent on each other

Closes: NA

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
         "fxa-auth-client",
         "fxa-shared",
         "fxa-profile-server",
-        "fxa-payments-server"
+        "fxa-payments-server",
+        "fxa-settings"
       ],
       "fxa-auth-server": [
         "fxa-auth-db-mysql",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -71,7 +72,8 @@
       ],
       "fxa-settings": [
         "fxa-shared",
-        "fxa-react"
+        "fxa-react",
+        "fxa-content-server"
```
dannycoates, line 76: im not sure `modules-to-test.js` can handle cycles. it might

vbudhram: Looks like it since these are stored in a set, https://github.com/mozilla/fxa/blob/20a0acf8b0d864702afb6d7c052f20a7f9ced526/.circleci/modules-to-test.js#L32.

dannycoates (approved):

PR #8248, opened 2021-04-13
fix(auth-server): handle empty responses

### LINKED ISSUE #7067: NestJS Sentry error handler should ignore NestJS HttpExceptions
NestJS has HttpExceptions for flow control, these include a `status` attribute that should be checked, and if it exists, Sentry does not need to report it.

https://sentry.prod.mozaws.net/operations/fxa-event-broker-prod/issues/10187808/?referrer=github_plugin

```
Error: 
  File "/fxa/packages/fxa-event-broker/dist/pubsub-proxy/pubsub-proxy.controller.js", line 42, in PubsubProxyController.proxy
    throw new common_1.HttpException(data, status);
  ?, in runMicrotasks
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2784)
┆Issue Number: FXA-2784

bbangert:
Because:

* We relay the exact response back from the RP we deliver to, and
  the HttpException test only triggers if the data is present.

This commit:

* Changes the HttpException property testing to check if the properties
  exist rather than their value.

Fixes #7067

## Checklist

### DIFF
--- a/packages/fxa-shared/nestjs/sentry/reporting.ts
+++ b/packages/fxa-shared/nestjs/sentry/reporting.ts
@@ -122,12 +122,6 @@ export function reportRequestException(
   excContexts: ExtraContext[] = [],
   request?: Request
 ) {
-  // Don't report HttpExceptions, we test for its two attributes as its more reliable
-  // than instance checks of HttpException
-  if (exception.status && exception.response) {
-    return;
-  }
-
   // Don't report already reported exceptions
   if (exception.reported) {
     return;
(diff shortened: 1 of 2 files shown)


bbangert:
Just a note, this is the best guess I have based on what we're seeing in production, and the only way I could trigger it locally. Stage does not exhibit this issue. If this doesn't fix prod event-broker, I'll likely have to emit a lot more logging detail at this step before proceeding anymore.

dannycoates (approved):
👍

Issue #8222, opened 2021-04-09
[subplat] Create DB Table for Sent Subscription Emails


data-sync-user:
For certain types of emails, we need to know when the last email of a certain type was sent, in order to, e.g.

* prevent duplicate emails
* decide whether to send an email

The first feature to make use of this table will be FXA-3166.

The columns we need are:

* uid: the customer's FxA uid
* emailType: the email template name
* subscriptionId: the id of the Stripe subscription, nullable
* planId: the Stripe price/plan id, nullable
* sentAt: timestamp when the email was sent

We need both the subscription id and the plan/price id is because subscription upgrades and downgrades could change the plan/price.  They are nullable since we send emails that not subscription specific (e.g. expiring CC reminder).

We also need to update the stored procedure for account deletion to delete the records from this new table.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3350)
┆Issue Number: FXA-3350

data-sync-user:
➤ Barry Chen commented:

Bumped up a point for updating account deletion stored procedure.

PR #8267, opened 2021-04-13
 feat(emails): add tables and models for email history

chenba:
Because:
 - sometimes we need to know when an email of a certain type was last
   sent

This commit:
 - add the database tables and the ORM models to record when an
   email was sent
 - update the delete account stored procedure to delete the account's
   records from the new table

Closes: #8222


packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
```diff
@@ -0,0 +1,54 @@
...
+--
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('114');
+
+CREATE TABLE IF NOT EXISTS sentSubscriptionEmails (
+  id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  uid BINARY(16) NOT NULL,
+  emailType VARCHAR(60) NOT NULL,
```
vbudhram, line 13: Templates don't really change often but we do associate a version number with them if that info might be needed.

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
```diff
@@ -0,0 +1,54 @@
+--
+-- This migration introduces a table to track sent emails related to
+-- subscriptions.
+--
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('114');
+
+CREATE TABLE IF NOT EXISTS sentSubscriptionEmails (
```
vbudhram, line 10: Probably be helpful to leave a comment here on what the different column store.

packages/fxa-shared/test/db/models/auth/helpers.ts
```diff
@@ -95,11 +73,20 @@ export async function testDatabaseSetup(): Promise<Knex> {
     user: 'root',
   });
 
-  await knex.raw(accountTable);
-  await knex.raw(devicesTable);
-  await knex.raw(emailsTable);
-  await knex.raw(accountCustomersTable);
-  await knex.raw(paypalBATable);
+  const thisDir = path.dirname(__filename);
```
vbudhram, line 76: 👍🏽

vbudhram (approved):
@chenba Just to confirm, this will be called from a cron job to do the email sending?

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
chenba: Thanks for the reminder about template versions.  I don't think we'll need it here since we mainly want to know what type of email was sent not the content.

> Templates don't really change often

I struggled a tiny bit with whether to normalize this more with an email types table.  It seems a little fragile requiring devs to remember manually add to the email types table whenever they add more email templates, all the while we don't really gain much from it.

Do let me know if you feel strongly about this one way or the other.

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
bbangert: Why not make this more generic so we can track any/all emails sent, not just subscription?

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
```diff
@@ -0,0 +1,54 @@
...
+  id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  uid BINARY(16) NOT NULL,
+  emailType VARCHAR(60) NOT NULL,
+  subscriptionId VARCHAR(255) NULL,
+  planId VARCHAR(255) NULL,
```
bbangert, line 15: We might want to send emails not about a particular subscription. Also, planId was replaced by priceId. Not sure we want to hard-code these into table columns as Stripe changes them.

bbangert: Maybe we should use a JSON column type to add arbitrary email argument values of distinction for more flexibility.

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
```diff
@@ -0,0 +1,54 @@
...
+  sentAt BIGINT UNSIGNED NOT NULL,
+  INDEX sentSubscriptionEmailsIndex(uid, emailType, subscriptionId, planId)
```
bbangert, line 17: If subscriptionId/planId are replaced with a JSON column of say `params` or something, indexing it perhaps isn't needed. I'm not sure what the utility would be in finding emails across all users sent for different sub/plan id's.

bbangert (changes requested):

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
chenba: > 
> 
> We might want to send emails not about a particular subscription. 

Yes, which is why both of those columns are nullable.

> Also, planId was replaced by priceId. Not sure we want to hard-code these into table columns as Stripe changes them.

That was a decision to keep things consistent as we still use "plan" in most places.

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
chenba: I'm not against it.  Just didn't know people wanted/needed it.  I see below that you are considering a JSON column for arbitrary data.  Looks like the indexing can be a problem: https://dev.mysql.com/doc/refman/5.7/en/create-table-secondary-indexes.html#json-column-indirect-index

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
chenba: If we don't index anything in the JSON column, then MySQL would need to scan the results for uid+emailType.  That probably will be ok?  As a FxA user and VPN subscriber, the volume of emails from FxA is pretty low.

chenba:
> 
> 
> @chenba Just to confirm, this will be called from a cron job to do the email sending?

The first feature to use it, yes.  But it can be for, e.g., tracking emails sent in response to Stripe webhooks.

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
bbangert: Yea, I don't think we need to index the params.

packages/fxa-shared/test/db/models/auth/helpers.ts
dannycoates: `__dirname`?

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
vbudhram: > Do let me know if you feel strongly about this one way or the other.

I don't feel strongly about this. The table would be fairly small compared to other tables.

chenba: > 
> 
> > Do let me know if you feel strongly about this one way or the other.
> 
> I don't feel strongly about this. The table would be fairly small compared to other tables.

Because @bbangert suggested that the table should be made generic to be used for _all_ emails, and since the number of FxA users is orders of magnitude greater than the number of SubPlat subscribers, the potential size and growth of the table just got much bigger.  I'll add a table for the email types in my updated patch.  Thanks for the input!

bbangert: Since there hasn't been any strong interest in having this to track all emails sent, I'm good with this as-is.

bbangert (approved):

chenba:
@vbudhram @bbangert  I've made changes based on earlier discussions.  There is now a table for the email types and a `params` JSON column for the sent emails.  Let me know what you think.  Thanks!

bbangert (approved):
LGTM!

packages/fxa-auth-db-mysql/lib/db/schema/patch-114-115.sql
```diff
@@ -0,0 +1,105 @@
...
+CALL assertPatchLevel('114');
+
+CREATE TABLE IF NOT EXISTS emailTypes (
+  id SMALLINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  -- these are intended to be email template names
+  emailType VARCHAR(255) NOT NULL,
+  UNIQUE emailTypeIndex (emailType)
+) ENGINE=InnoDB;
+
+INSERT INTO emailTypes (emailType) VALUES
```
vbudhram, line 17: Would adding a new email template require a database migration? We don't add template often but I can see this as a big barrier.

vbudhram, line 17: Would adding a new email template require a database migration? We don't add template often but I can see this as a big barrier when we do.

chenba: > 
> 
> Would adding a new email template require a database migration? We don't add template often but I can see this as a big barrier when we do.

I definitely don't think it's required.  It's only necessary if we want to track when that type of emails are sent.  Aside from the foreign key constraint there's nothing reinforcing the relationships among the template names, the email types in the table, and what we want to track.

vbudhram: > Aside from the foreign key constraint there's nothing reinforcing the relationships among the template names

SGTM 👍🏽

vbudhram (approved):

PR #8269, opened 2021-04-14
fix(auth): Handle missing billing agreement id in handleInvoiceOpenEv…

### LINKED ISSUE #8265: [subplat] Handle internal server error in handleInvoiceOpenEvent
We are getting `Internal Server Error`s in Sentry (#FXA-AUTH-STAGE-2Z) for Stripe `invoice.finalized` webhook events when there is no billing agreement id on the `customer`, because we don't explicitly handle this type of error in our `catch` block in `handleInvoiceOpenEvent`.

After some discussion, we decided to handle this scenario further upstream, since in this case, we don't actually want to process the invoice – we can't process the invoice if there is no billing agreement.

Instead, we should `sendSubscriptionPaymentFailedEmail` to let the customer know that we cannot process a payment.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3372)
┆Issue Number: FXA-3372

biancadanforth:
…ent.

Because:

* We currently throw an unhandled exception, Internal Server Error when we try to process an invoice with no billing agreement id.

This commit:

* Modifies this behavior to send a payment failed email and early return instead.

Closes #8265, closes #FXA-AUTH-STAGE-2Z

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -337,6 +337,13 @@ export class StripeWebhookHandler extends StripeHandler {
     if (customer.deleted) {
       return;
     }
+    const billingAgreementId = this.stripeHelper.getCustomerPaypalAgreement(
+      customer
+    );
+    if (!billingAgreementId) {
+      await this.sendSubscriptionPaymentFailedEmail(invoice);
+      return;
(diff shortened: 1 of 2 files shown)


biancadanforth:
@bbangert It'd be great to get this into 204 if possible. I think they are tagging Thursday (4/13) morning.

bbangert (approved):
LGTM!

Issue #8268, opened 2021-04-13
[subplat] The Legal Docs Redirect Endpoint's URL Query Param is not Properly Encoded


data-sync-user:
Some query params are added to links in the subscription emails, including the legal doc download URLs.  The legal doc download URLs themselves are used as a query param to the redirect endpoint.  They are not encoded as such at the moment.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3374)
┆Issue Number: FXA-3374

data-sync-user:
➤ Barry Chen commented:

Removed the triage label because I'm going to fix this today.  I'll also update the description since the bug is actually with the emails not the endpoint.

PR #8276, opened 2021-04-14
fix(emails): encode the legal docs redirect url query param correctly

chenba:
Because:
 - the url query parameter for the legal docs redirect endpoint links in
   the subscription emails was not encoded correctly

This commit:
 - encode the url query param as a URI component

Closes: #8268


jodyheavener (approved):
🙂

chenba:
Failing test is not related to the changes.

Issue #8204, opened 2021-04-08
[l10n][settings][payments] Remove en-CA from the list locales that fall back to en-GB


data-sync-user:
When VPN expanded into other English speaking markets, some code was added so that the localized string fall back to the en-GB translation.  en-CA was one of those locale that's included in the hard coded fallback.  en-CA now has its own settings.ftl and main.ftl in the l10n repo, so we should remove en-CA from that list.

https://github.com/mozilla/fxa/blob/main/packages/fxa-react/lib/AppLocalizationProvider.tsx#L12

https://github.com/mozilla/fxa/blob/main/packages/fxa-react/lib/AppLocalizationProvider.tsx#L54

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3345)
┆Issue Number: FXA-3345

PR #8277, opened 2021-04-14
fix(l10n): use en-CA not en-GB for settings and payments

chenba:
Because:
 - we are using en-GB messages for en-CA even though we have
   translations for en-CA

This commit:
 - remove en-CA from the list of locales that falls back to en-GB
 - add en-CA to the list of supported languages

Closes: #8204


jodyheavener (approved):
bless u 🇨🇦

PR #8284, opened 2021-04-14
fix: stop reporting apollo errors

bbangert:
Because:

* Apollo Errors are already internally processed and used for control
  flow. They should not be reported to Sentry as the underlying error
  was already reported.

This commit:

* Properly ignores ApolloError's in the interceptor.

## Checklist

This is from Sentry: https://sentry.prod.mozaws.net/operations/fxa-graphql-prod/issues/10391832/
An issue was not filed.

### DIFF
--- a/packages/fxa-shared/nestjs/sentry/sentry.interceptor.ts
+++ b/packages/fxa-shared/nestjs/sentry/sentry.interceptor.ts
@@ -4,10 +4,11 @@
 import {
   CallHandler,
   ExecutionContext,
+  HttpException,
   Injectable,
   NestInterceptor,
-  HttpException,
 } from '@nestjs/common';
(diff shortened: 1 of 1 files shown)


packages/fxa-shared/nestjs/sentry/sentry.interceptor.ts
```diff
@@ -4,10 +4,11 @@
 import {
   CallHandler,
   ExecutionContext,
+  HttpException,
   Injectable,
   NestInterceptor,
-  HttpException,
 } from '@nestjs/common';
+import { ApolloError } from 'apollo-server-errors';
```
dannycoates, line 11: I _think_ this can just be 'apollo-server'

dannycoates (approved):
lgtm

PR #8288, opened 2021-04-15
fix(test): fix oauth functional test

### LINKED ISSUE #8281: Changing display name on new-settings should fire profile changed event
This is causing breakage in one of the oauth_permissions tests.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3377)
┆Issue Number: FXA-3377

dannycoates:
Ummmmmm. Local storage is not a good way to handle oauth permission prompts but fixing that is much different than fixing the test case.

fixes #8281

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
+++ b/packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
@@ -15,6 +15,7 @@ const {
   createUser,
   openPage,
   fillOutEmailFirstSignIn,
+  testElementExists,
 } = FunctionalHelpers.helpersRemoteWrapped;
 
 async function navigateToSettingsV2(remote) {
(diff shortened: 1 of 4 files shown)


packages/fxa-content-server/tests/functional/settings_v2/lib/helpers.js
```diff
@@ -24,7 +25,7 @@ async function navigateToSettingsV2(remote) {
 
   await openPage(EMAIL_FIRST, selectors.ENTER_EMAIL.HEADER, remote);
   await fillOutEmailFirstSignIn(email, password, remote);
-
+  await testElementExists(selectors.SETTINGS_V2.APP);
```
vbudhram, line 28: The `remote` object needs to be passed as last param to these helper methods.

dannycoates: how did this work for me locally?! 🤦

vbudhram (approved):

PR #8300, opened 2021-04-15
fix(tests): Fixes recovery key flaky test and updates functional regression

vbudhram:
Fixes https://app.circleci.com/pipelines/github/mozilla/fxa/18522/workflows/5f462bf6-309d-457f-b376-a470ace0674a/jobs/234255

Updates regression tests to use `settings_v2`.


packages/fxa-content-server/tests/functional/recovery_key.js
```diff
@@ -126,19 +126,13 @@ registerSuite('Recovery key', {
...
 
-          // enter old key
+          // enter old key and check for error tooltip
           .then(fillOutRecoveryKey(recoveryKey))
           .then(
             testElementExists(
               selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP
             )
           )
-          .then(
```
vbudhram, line 136: It appears that the tooltip takes longer than the maximum page load time to appear? Either way, it doesn't really provide any value to check the text in here if it is failing so much.

Ref https://234255-36831081-gh.circle-artifacts.com/0/screenshots/17d5b9fb.png

dannycoates (approved):

PR #8303, opened 2021-04-15
chore(ci): increase test-content-server-remote task to a large instance

dannycoates:
## Because

- we've got too many test failures due to slow rendering


vbudhram (approved):
@dannycoates Thank you!

PR #8305, opened 2021-04-15
fix(settings): ensure the object returned from useSession is stable

dannycoates:
Because the ModalVerifySession has a useEffect that depends on the session to send a verification email. Without this it could send multiple emails.


vbudhram (approved):
@dannycoates Works locally 👍🏽 

Unfortunately looks like we don't have functional tests for these flows anymore. If you get a chance could you file an issue in https://jira.mozilla.com/browse/FXA-1556.

PR #8322, opened 2021-04-16
fix(settings): clear local storage on account delete

### LINKED ISSUE #7855: Deleting account no longer clears storage on backbone side
I've noticed in updating the functional tests that deleting the account now leads to the "Enter password to sign in" form. I suspect this is because deleting the account clears data on the React side, but not on the Backbone side.

The simplest fix seems to be react clearing out the localstorage state on delete. I'm not sure if communication with the backbone app is necessary.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3200)
┆Issue Number: FXA-3200

dannycoates:
fixes #7855

### DIFF
--- a/packages/fxa-settings/src/models/Account.ts
+++ b/packages/fxa-settings/src/models/Account.ts
@@ -3,6 +3,7 @@ import config from '../lib/config';
 import AuthClient, { generateRecoveryKey } from 'fxa-auth-client/browser';
 import { currentAccount, sessionToken } from '../lib/cache';
 import firefox from '../lib/firefox';
+import Storage from '../lib/storage';
 
 export interface DeviceLocation {
   city: string | null;
@@ -721,5 +722,6 @@ export class Account implements AccountData {
       )
     );
     firefox.accountDeleted(this.uid);
+    Storage.factory('localStorage').clear();
   }
 }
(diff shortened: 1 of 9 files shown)


packages/fxa-settings/src/models/Account.ts
```diff
@@ -721,5 +722,6 @@ export class Account implements AccountData {
       )
     );
     firefox.accountDeleted(this.uid);
+    Storage.factory('localStorage').clear();
```
vbudhram, line 725: Would this remove the currently signed in user or all users in local storage?

vbudhram (approved):

packages/fxa-settings/src/models/Account.ts
dannycoates: It does remove all of them. IDK what the previous behavior was but it seems reasonable to me to clear the list at this point. If it isn't cleared the next visit defaults to the first account in the list, which seems surprising to me. I think its better to start from 'email first' after a 'delete account'. What do you think?

vbudhram: > If it isn't cleared the next visit defaults to the first account in the list, which seems surprising to me.

Yea that is surprising. Would be nice to have an account chooser option but that is out of scope.

Let's keep this as is and show the user email first 👍🏽

PR #8330, opened 2021-04-19
task(monorepo): Update legal docs

meandavejustice:
## Because

- We want our users to access the most up to date terms of service and privacy policy for the VPN product

## This pull request

- updates the legal-docs dependency
- ran script to generate new legal docs

## Checklist


chenba (changes requested):
The changes look fine, but these are two unrelated changes.  Can you separate them into two PRs?  Also, any reason we don't specify the commit for the legal-docs dependency in content-server?  I don't understand why we had it and why we are removing it now.

meandavejustice:
@chenba 
> The changes look fine, but these are two unrelated changes.

I disagree that the changes are unrelated, we were told that the terms/privacy policy had updated, so we need to update the package to get the latest and generate the new files. Am I missing something?

> Also, any reason we don't specify the commit for the legal-docs dependency in content-server

I had to look up why we were still including the commit hash, I added a [note](https://github.com/mozilla/fxa/pull/8330/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R463) this time.

chenba:
> 
> 
> @chenba
> 
> > The changes look fine, but these are two unrelated changes.
> 
> I disagree that the changes are unrelated, we were told that the terms/privacy policy had updated, so we need to update the package to get the latest and generate the new files. Am I missing something?
> 

I thought the request was to re-generate the legal doc MDs to PDFs (of course this is from memory and I could be completely wrong here), which are for the VPN.  The package.json update is for updating the ToS and privacy notice on the content-server.  And I'm glad that you remembered and updated that.  I just felt the PDFs for the CDN and the legal content for the content-server are unrelated.

But there isn't a jira or any sort of formal process for this, and there's no real harm to combining these changes, so I'm open to having just the one PR.

As for the note, thanks for adding it.  But that section of the doc is about the MD to PDF generation.  It doesn't really have anything to do with the content-server, so I don't think that's the right place for it.

meandavejustice:
@chenba I think I see where the confusion is coming from. content-server just happens to be where the `legal-docs` dependency is installed(for historical reasons?). I suppose it could be moved to the top level component, which would more accurately reflect the reality of our usage of the legal docs

chenba:
> 
> 
> @chenba I think I see where the confusion is coming from. content-server just happens to be where the `legal-docs` dependency is installed(for historical reasons?). I suppose it could be moved to the top level component, which would more accurately reflect the reality of our usage of the legal docs


No, I'm saying generating the legal PDFs for the CDN is independent of the content-server.  There's no direct relationship between that what's generated by that script and teh content-server's ToS and privacy policy pages.

README.md
```diff
@@ -458,3 +459,6 @@ The script traverses the legal-docs directory looking for localized copies of th
 Example:
 directory provided: `/Users/test/github/mozilla/legal-docs/`
 resulting file: `assets/legal/<document_name>.<locale>.pdf`
+
+_note on legal-docs_:
+We intentionally pin to a commit and return the hash at `/__version__` because the `legal-docs/package.json` is not kept up to date.
```
chenba, line 464: Generating these PDFs doesn't affect the hash in content-server `__version__` or vice-versa.

meandavejustice:
@chenba removed note in README

chenba (approved):

Issue #8260, opened 2021-04-13
Implement metadata validation on data update


data-sync-user:
Update validator, validate on initial fetch and updates, and throw Sentry error if invalid (hopefully this logs already e.g. to #fxa)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3369)
┆Issue Number: FXA-3369

data-sync-user:
➤ Dave Justice commented:

The metadata keys which are required and need to be validated:

https://mozilla.slack.com/archives/CG0AJ6E77/p1618266162310300?thread_ts=1618247342.299000&cid=CG0AJ6E77 ( https://mozilla.slack.com/archives/CG0AJ6E77/p1618266162310300?thread_ts=1618247342.299000&cid=CG0AJ6E77 )

from [~bchen@mozilla.com] :

to answer the original question:
 productSet: No
 productOrder: Yes but only if productSet has a value, otherwise No
 webIconURL: Yes, in the format of a URL
 capabilities: Yes but only if capabilities:{clientID}doesn't have a value
 capabilities:{clientID}: Yes but only if capabilitiesdoesn't have a value
 upgradeCTA: No
 downloadURL: Yes, in the format of a URL
 emailIconURL: No
 product:subtitle: No (but I don't know how the UI looks like without it)
 product:subtitle:{locale}: No
 product:details:{n}: Yes, at least one
 product:details:\{n}:{locale}: No
 product:playStoreLink: No
 product:appStoreLink: No
 product:termsOfServiceURL: Yes, in the format of a URL
 product:termsOfServiceURL:{locale}: No
 product:termsOfServiceDownloadURL: Yes, in the format of a URL, bonus if we ensure the domain is that of the FxA CDN, i.e. accounts-static.cdn.mozilla.net ( http://accounts-static.cdn.mozilla.net/ )
 product:privacyNoticeURL: Yes
 product:privacyNoticeURL:{locale}: No
 product:privacyNoticeDownloadURL:  Yes, in the format of a URL, bonus if we ensure the domain is that of the FxA CDN, i.e. accounts-static.cdn.mozilla.net ( http://accounts-static.cdn.mozilla.net/ )
 support:app:{x}: No

data-sync-user:
➤ Dave Justice commented:

[...]

Issue #7856, opened 2021-03-17
[subplat] Remove Default Product Metadata


data-sync-user:
There are default product metadata in the code base that act as the final fallback values when something is not found in the product/plan.  It made sense when VPN was the only product.  But with multiple products, it would produce false positives and hide incorrect product/plan configuration.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3201)
┆Issue Number: FXA-3201

data-sync-user:
➤ Bianca Danforth commented:

Probably small, but a lot of things assume a default is set.

PR #8332, opened 2021-04-19
fix(payments): Set plan as a required prop for the TermsAndPrivacy co…

### LINKED ISSUE #8296: [subplat] ToS and PN from “error subscription confirmation” page are for Browser extension and not VPN
**Affected versions:**

* All Firefox;

**Affected FxA versions:**

* 1.203.7;

**Affected Platforms:**

* Windows 10;

**Prerequisites:**

* Have a subscription made for VPN product using the PayPal payment method;

**Steps to reproduce:**

* Open VPN homepage https://www-dev.allizom.org/en-US/products/vpn/
* Try to subscribe to another product using the account from prerequisites;
* In the error subscription page, click the ToS/PN links

**Expected result:**

* ToS/PN for Mozilla VPN product are opened.

**Actual result:**

* ToS/PN for FPN browser extension are loaded.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3379)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15982/tospn.png">tospn.png</a>
┆Issue Number: FXA-3379

biancadanforth:
…mponent

Because:

* When the legal links metadata is missing from a Stripe product/plan, we use [fallback details](https://github.com/mozilla/fxa/blob/c950d28a5333f4718761aeed68cef258476ee520/packages/fxa-payments-server/src/components/TermsAndPrivacy/index.tsx#L23-L29), which can be incorrect.
* Since we only use the TermsAndPrivacy React component during Checkout, where there is only ever one Stripe product/plan selected, we should require the plan to be passed in instead of it being optional.

This commit:
* Makes the plan a required prop on the TermsAndPrivacy component, which ensures the same legal links are used throughout the Checkout process, and that they pull from Stripe product/plan metadata first.

In general, legal links to ToS and privacy notices should always be present on Stripe product/plan metadata (#8260), and we should not rely on a fallback at all (#7856).

Closes #8296

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/components/AppLayout/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/AppLayout/index.test.tsx
@@ -7,10 +7,16 @@ import { AppContext, defaultAppContext } from '../../lib/AppContext';
 import AppLayout, { SignInLayout, SettingsLayout } from './index';
 import TermsAndPrivacy from '../TermsAndPrivacy';
 import { DEFAULT_PRODUCT_DETAILS } from 'fxa-shared/subscriptions/metadata';
+import { SELECTED_PLAN } from '../../lib/mock-data';
 
 afterEach(cleanup);
 
-const { termsOfServiceURL, privacyNoticeURL } = DEFAULT_PRODUCT_DETAILS;
+const {
+  product_metadata: {
+    'product:termsOfServiceURL': termsOfServiceURL,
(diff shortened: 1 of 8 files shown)


chenba (approved):
👍

meandavejustice (approved):
👍

PR #8366, opened 2021-04-21
fix(config): use a regex for default config values instead of string

vbudhram:
## Because

- This should actually be a regex value instead of a string

## This pull request

- Updates the default value be what is needed in the functional tests

Closes: NA

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -1458,11 +1458,9 @@ const conf = convict({
   },
   forcePasswordChange: {
     forcedEmailAddresses: {
-      doc:
-        'Force sign-in confirmation for email addresses matching this regex.',
+      doc: 'Force password change for email addresses matching this regex.',
       format: RegExp,
-      default: '^$', // default is no one
-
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -1458,11 +1458,9 @@ const conf = convict({
   },
   forcePasswordChange: {
     forcedEmailAddresses: {
-      doc:
-        'Force sign-in confirmation for email addresses matching this regex.',
+      doc: 'Force password change for email addresses matching this regex.',
       format: RegExp,
-      default: '^$', // default is no one
-
+      default: /^forcepwdchange/,
```
dannycoates, line 1463: I think the default should still be no one

dannycoates (approved):

packages/fxa-auth-server/config/index.ts
vbudhram: Don't have a strong pref for this either way, only benefit is saving some space on pm2 config and specifying in cloud ops.

PR #8374, opened 2021-04-22
fix(profile): drop profileCache if the avatar changes

dannycoates:
We weren't dropping the profileCache when the avatar was changing so subsequent requests would also think the avatar needed to change and would write to the db until something else triggered the cache to be updated.

This change ensures the cache is dropped when the avatar changes.

### DIFF
--- a/packages/fxa-profile-server/lib/routes/profile.js
+++ b/packages/fxa-profile-server/lib/routes/profile.js
@@ -23,8 +23,7 @@ function computeEtag(profile) {
   return false;
 }
 
-function checkAvatar(response, uid) {
-  const result = response.value;
+function nextAvatar(result) {
   if (
     result.avatar &&
     (result.avatarDefault || result.avatar.startsWith(monogramUrl))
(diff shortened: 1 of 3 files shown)


packages/fxa-profile-server/lib/routes/profile.js
```diff
@@ -124,21 +121,22 @@ module.exports = {
...
+    const newAvatar = nextAvatar(result);
+    const avatarChanged = result.avatar !== newAvatar;
+    if (avatarChanged) {
+      await changeAvatar(newAvatar, creds.user);
+    }
+    // Check to see if the oauth-server is reporting a newer `profileChangedAt`
+    // timestamp from validating the token, if so, lets invalidate the cache
+    // and set new value.
+    if (avatarChanged || result.profileChangedAt < creds.profile_changed_at) {
+      await server.methods.profileCache.drop(creds.user);
```
dannycoates, line 135: we've got a cache per instance, right? so this prob isn't enough to ensure the `addAvatar` only hits once.

vbudhram (approved):
@dannycoates 💯 :shipit:

PR #8389, opened 2021-04-26
fix(tests): Fix regression functional tests

vbudhram:
## Because

- These tests were failing when they ran against stage environment

## This pull request

- Fixes many small issues with failing tests
  - Majority were clicking a button and not having a ready selector to let the test know to continue to next step
  - Couple were stage config related
- Adds a small delay (100ms) when opening Rp window, before making any clicks
  - I can't exactly pinpoint why this delay helps fixing these, but they were passing 5/5 times on stage 🤷🏾‍♂️

Closes: #8390

## Checklist

## Notes

Triggered the job directly and all tests were passing.

Screen Shot 2021-04-26 at 6 12 36 PM

### DIFF
--- a/packages/fxa-content-server/tests/functional_regression.js
+++ b/packages/fxa-content-server/tests/functional_regression.js
@@ -5,9 +5,13 @@
 const testsSettingsV2 = require('./functional_settings_v2');
 
 module.exports = testsSettingsV2.concat([
-  'tests/functional/oauth_webchannel.js',
-  'tests/functional/oauth_require_totp.js',
-  'tests/functional/sign_up_with_code.js',
+  // These oauth tests use untrusted clients which
+  // would fixed when https://github.com/mozilla/fxa/issues/5291 lands
+  // 'tests/functional/oauth_settings_clients.js',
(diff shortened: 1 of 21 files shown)


packages/fxa-content-server/tests/functional_regression.js
```diff
@@ -48,10 +50,10 @@ module.exports = testsSettingsV2.concat([
   'tests/functional/sign_in_blocked.js',
   'tests/functional/sign_in_cached.js',
   'tests/functional/sign_in_recovery_code.js',
+  'tests/functional/sign_up_with_code.js',
   'tests/functional/sign_up.js',
   'tests/functional/sync_v1.js',
   'tests/functional/sync_v2.js',
   'tests/functional/sync_v3_force_auth.js',
   'tests/functional/tos.js',
-  'tests/functional/verification_reminders.js',
```
vbudhram, line 56: This test gets rate limited against stage because it has too many failed login attempts from a single IP address. Disabled for now until we up the limit for this.

dannycoates (approved):

Issue #8031, opened 2021-03-29
[subplat] Privacy Policy link in HTML transactional emails is missing


data-sync-user:
While I was working on FXA-3114, I noticed that, while we include the product's terms of service link in both HTML and plain text emails, we don't include a link to the product's privacy policy in HTML emails. These links are to the localized PDF versions of the documents.

[~byee@mozilla.com] I wanted to surface this for your thoughts on how high a priority this bug is (e.g. from a legal perspective).

I should note that we do link to the ToS and Privacy Policy webpages during Checkout on the Subscription Platform website, though we only provide a downloadable PDF link for the ToS and not the Privacy Policy for some reason.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3290)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/15713/Checkout_footer.png">Checkout_footer.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15712/HTML_email_footer.png">HTML_email_footer.png</a> | <a href="https://jira.mozilla.com/secure/attachment/15916/privacy_tos.JPG">privacy_tos.JPG</a>
┆Issue Number: FXA-3290

data-sync-user:
➤ Barry Chen commented:

While working on this I noticed that we do have a privacy notice link!  It's at the bottom to the right of the "Legal" link.

data-sync-user:
➤ Barry Chen commented:

The privacy notice link is in the HTML emails already.

data-sync-user:
➤ Bianca Danforth commented:

Oh interesting; I would imagine that link should actually go to Mozilla's privacy statement, as opposed to a product-specific privacy policy? The Legal link next to it goes to https://www.mozilla.org/about/legal/terms/services/, so wouldn't we want that link to go to e.g. https://www.mozilla.org/en-US/privacy/?

data-sync-user:
➤ Barry Chen commented:

That'd make sense to me.  But let's see what Product/Legal think?

[~byee@mozilla.com]  The "Privacy" link in the attached HTML_email_footer.png (lower right, below the Mozilla logo) is a link to the product specific privacy policy doc.  Should that be a link to Mozilla's privacy policy page and the link for the product specific doc would go after "Terms and cancellation policy" (above the Mozilla logo)?

data-sync-user:
➤ Brian Yee commented:

So there is a:

[...]

PR #8397, opened 2021-04-26
fix(emails): use site privacy notice url not product privacy notice

chenba:
Because:
 - we should use Mozilla's privacy notice page in the footer
   subscription emails

This commit:
 - update the url for the privacy notice link in the footer of
   subscription emails

Closes: #8031

### DIFF
--- a/packages/fxa-auth-server/test/local/senders/email.js
+++ b/packages/fxa-auth-server/test/local/senders/email.js
@@ -193,7 +193,7 @@ const TESTS = [
       ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.downloadSubscription }],
     ])],
     ['html', [
-      { test: 'include', expected: configHref('subscriptionPrivacyUrl', 'new-subscription', 'subscription-privacy') },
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -224,7 +222,6 @@ const TESTS = [
       ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.subscriptionFirstInvoice }],
     ])],
     ['html', [
-      { test: 'include', expected: configHref('subscriptionPrivacyUrl', 'subscription-first-invoice', 'subscription-privacy') },
```
vbudhram, line 227: nit: Could check for new privacy url?

vbudhram (approved):

packages/fxa-auth-server/test/local/senders/email.js
chenba: heh yeah I skipped it because everything's hard coded; I'll add a check.  👍

PR #8437, opened 2021-04-29
task(fxa-auth-server): Improve validation for product metadata

meandavejustice:
## Because

- We want to have some validation of the metadata added in stripe

## This pull request

- Adds validation for the product metadata

Closes: #8260

## Checklist

## Screenshots

Example sentry error:

validation-update-example


bbangert (approved):
Looks good, is there going to be a follow-up to integrate this running somewhere?

chenba:
This change would break the GET plans endpoint if any plan does not pass validation.

chenba (changes requested):
We want to validate the response from Stripe, not our response to the clients.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1052,6 +1053,23 @@ export class StripeHelper {
         continue;
       }
 
+      const result = subscriptionProductMetadataValidator.validate(
+        item.product.metadata
+      );
+
+      if (result && result.error) {
+        this.log.error(
```
meandavejustice, line 1061: WIP, looking to see best way to pass validation error here.

meandavejustice:
This is just about finished up, I Just have to finish going through our dev product metadata in stripe and updating the newly required fields.

meandavejustice:
@chenba How do you propose we handle this?

chenba:
> @chenba How do you propose we handle this?
> 
> > When auth-server uses Stripe's sdk to get the list of plans, the plan's metadata do not have the product metadata already merged into them. We'd get a lot of false negatives here due to missing metadata.

First I should clarify that the validation need to be performed in two places: getPlans and the webhook handler.  The validation in getPlans is to prevent any plans with invalid metadata from being used.  The webhook handler is for quicker feedback on validation errors.

For getPlans, you can see that fetchAllPlans is returning a list of objects that have the properties of 'plan_metadata' and 'product_metadata'.  You can validate on a merged object of those two properties' properties.  Take a look at `productDetailsFromPlan` in fxa-shared/subscriptions/metadata.ts.  You cannot use that function though, since it'll fill in default values for the missing metadata.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -98,6 +99,12 @@ export class StripeWebhookHandler extends StripeHandler {
         case 'invoice.payment_failed':
           await this.handleInvoicePaymentFailedEvent(request, event);
           break;
+        case 'product.created':
```
meandavejustice, line 102: Still need to test that this works as expected on fresh product creation.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1055,6 +1056,24 @@ export class StripeHelper {
         continue;
       }
 
+      const result = subscriptionProductMetadataValidator.validate(
+        item.product.metadata
+      );
```
chenba, line 1061: This only validates the product's metadata.  See https://github.com/mozilla/fxa/pull/8437#pullrequestreview-657053420 about false negatives.  And see https://github.com/mozilla/fxa/pull/8437#issuecomment-840045090 on where to find an example of the metadata being merged.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -98,6 +99,9 @@ export class StripeWebhookHandler extends StripeHandler {
         case 'invoice.payment_failed':
           await this.handleInvoicePaymentFailedEvent(request, event);
           break;
+        case 'product.updated':
```
chenba, line 102: I don't think this event is triggered when a plan is updated.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +431,31 @@ export class StripeWebhookHandler extends StripeHandler {
...
   }
 
+  /**
+   * Validate product metadata on update
+   */
+  async handleProductUpdate(request: AuthRequest, event: Stripe.Event) {
+    const product = event.data.object as Stripe.Source;
+    const result = subscriptionProductMetadataValidator.validate(
+      product.metadata
+    );
```
chenba, line 441: The same issue with validating only the product metadata applies here.  On a product update, if the metadata have been updated (you can check the event's "previous_attributes" property for a "metadata" key), then we should validate all the plans for the product.

chenba (changes requested):
The frontend relies on metadata that's coalesced from the product and the plan, so we need to validate on that, not either the product metadata or the plan metadata.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
meandavejustice: I tested manually, you'll have to run `stripe listen --forward-to localhost:9000/v1/oauth/subscriptions/stripe/event` in order to have the events picked up locally.

bbangert: Yea, I'd expect the parent object to trigger. Same happens with most customer child objects triggering a customer.updated.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +431,35 @@ export class StripeWebhookHandler extends StripeHandler {
...
+  async handleProductUpdate(request: AuthRequest, event: Stripe.Event) {
+    const product = event.data.object as Stripe.Source;
+
+    const plans = this.stripeHelper.fetchPlansByProductId(product.id);
+    for await (const item of plans) {
+      const result = subscriptionProductMetadataValidator.validate(
+        Object.assign({}, product.metadata, item.metadata)
+      );
+
+      if (result && result.error) {
```
bbangert, line 446: Optional chaining.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1055,6 +1069,24 @@ export class StripeHelper {
         continue;
       }
 
+      const result = subscriptionProductMetadataValidator.validate(
+        Object.assign({}, item.product.metadata, item.metadata)
```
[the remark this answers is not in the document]
meandavejustice: I don't have an opinion on it, which would you prefer?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: At the time of my comment I verified it by simply editing a plan/price in the Stripe dashboard and the checked the events (since I didn't have Stripe CLI and everything else running) , and I didn't see a product.updated.  I just tried the same thing again and I still don't

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -98,6 +99,10 @@ export class StripeWebhookHandler extends StripeHandler {
         case 'invoice.payment_failed':
           await this.handleInvoicePaymentFailedEvent(request, event);
           break;
+        case 'product.updated':
+        case 'plan.updated':
+          await this.handleProductUpdate(request, event);
```
chenba, line 104: The same function is used to handle both a product.updated and plan.updated here but the shape of the event objects won't be the same.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,36 @@ export class StripeWebhookHandler extends StripeHandler {
     await this.updateCustomer(uid, email);
   }
 
+  /**
+   * Validate product metadata on update
+   */
+  async handleProductUpdate(request: AuthRequest, event: Stripe.Event) {
+    const product = event.data.object as Stripe.Source;
```
chenba, line 439: I don't understand why this is a Stripe.Source.  Shouldn't it be a Stripe.Product?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,36 @@ export class StripeWebhookHandler extends StripeHandler {
...
+
+    const plans = this.stripeHelper.fetchPlansByProductId(product.id);
```
chenba, line 441: I don't think `product.id` exists if this executes during a `plan.updated` event.  Also, we don't need to validate all the plans when only one plan has changed.  We only need to do that for `product.updated`.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -98,6 +99,12 @@ export class StripeWebhookHandler extends StripeHandler {
         case 'invoice.payment_failed':
           await this.handleInvoicePaymentFailedEvent(request, event);
           break;
+        case 'product.updated':
+          await this.handleProductUpdate(request, event);
+          break;
+        case 'plan.updated':
+          await this.handlePlanUpdate(request, event);
```
chenba, line 106: nit: the other event handling functions all have names in the format of handle*Event

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,70 @@ export class StripeWebhookHandler extends StripeHandler {
...
+   */
+  async handlePlanUpdate(request: AuthRequest, event: Stripe.Event) {
+    const plan = event.data.object as Stripe.Plan;
+    const products = await this.stripeHelper.allProducts();
+    let product = products.find((p) => p.product_id === plan.product);
+    if (!product) {
+      this.log.error(
+        `handlePlanUpdate - Could not locate Product ${plan.product}`,
+        { item: plan }
+      );
```
chenba, line 446: nit: can you return early here so we don't need the `else`?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,70 @@ export class StripeWebhookHandler extends StripeHandler {
...
+      if (result?.error) {
+        const msg = `handlePlanUpdate - Plan "${plan.id}"'s metadata failed validation; After Plan update.`;
+        this.log.error(msg, { item: plan, error: result.error });
+
+        Sentry.withScope((scope) => {
+          scope.setContext('validationError', {
+            error: result.error,
+          });
+          Sentry.captureMessage(
+            `Error validating metadata for plan ${plan.id} and product ${plan.product}`,
```
chenba, line 462: hmm why isn't `msg` used here?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,70 @@ export class StripeWebhookHandler extends StripeHandler {
...
+        { item: plan }
+      );
+    } else {
+      const result = subscriptionProductMetadataValidator.validate({
+        ...product.product_metadata,
+        ...plan.metadata,
+      });
+
+      if (result?.error) {
+        const msg = `handlePlanUpdate - Plan "${plan.id}"'s metadata failed validation; After Plan update.`;
```
chenba, line 454: nit: 'After Plan update' seems a little redundant since the message also has 'handlePlanUpdate'.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,70 @@ export class StripeWebhookHandler extends StripeHandler {
...
+          );
+        });
+      }
+    }
+  }
+
+  /**
+   * Validate product metadata on update
+   */
+  async handleProductUpdate(request: AuthRequest, event: Stripe.Event) {
```
chenba, line 473: I noticed that the plan update handler is relying on `allProducts`, which is cached.  We should invalidate that cache here.  Otherwise, the plan update handler could end up using stale products.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -427,6 +432,70 @@ export class StripeWebhookHandler extends StripeHandler {
...
+      if (result?.error) {
+        const msg = `handleProductUpdate - Plan "${item.id}"'s metadata failed validation; Product ${product.id} updated.`;
+        this.log.error(msg, { item: product, error: result.error });
+
+        Sentry.withScope((scope) => {
+          scope.setContext('validationError', {
+            error: result.error,
+          });
+          Sentry.captureMessage(
+            `Error validating metadata for product ${product.id} and plan ${item.id}`,
```
chenba, line 491: Could use `msg` here?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
meandavejustice: updated to use `fetchAllProducts`

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1089,6 +1102,25 @@ export class StripeHelper {
...
       }
 
+      const result = subscriptionProductMetadataValidator.validate({
+        ...item.product.metadata,
+        ...item.metadata,
+      });
+
+      if (result?.error) {
+        const msg = `fetchAllPlans - Plan "${item.id}"'s metadata failed validation`;
+        this.log.error(msg, { item: item, error: result.error });
```
chenba, line 1112: Use 'plan' instead of 'item' here for the key, since the object is a plan?  ('item' is just the name of the variable.)

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
...
+   * Validate plan metadata on update
+   */
+  async handlePlanUpdatedEvent(request: AuthRequest, event: Stripe.Event) {
+    const plan = event.data.object as Stripe.Plan;
+    const products = await this.stripeHelper.fetchAllProducts();
+    let product = products.find((p) => p.product_id === plan.product);
+    if (!product) {
+      this.log.error(
+        `handlePlanUpdatedEvent - Could not locate Product ${plan.product}`,
+        { item: plan }
```
chenba, line 445: s/item/plan/ ?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
     await this.updateCustomer(uid, email);
   }
 
+  /**
+   * Validate plan metadata on update
+   */
+  async handlePlanUpdatedEvent(request: AuthRequest, event: Stripe.Event) {
+    const plan = event.data.object as Stripe.Plan;
+    const products = await this.stripeHelper.fetchAllProducts();
```
chenba, line 440: Since this is a plan update, we can used the cached products from `allProducts`.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
...
+      return;
+    }
+    const result = subscriptionProductMetadataValidator.validate({
+      ...product.product_metadata,
+      ...plan.metadata,
+    });
+
+    if (result?.error) {
+      const msg = `handlePlanUpdatedEvent - Plan "${plan.id}"'s metadata failed validation`;
+      this.log.error(msg, { item: plan, error: result.error });
```
chenba, line 456: s/item/plan/ ?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
...
+    const plans = this.stripeHelper.fetchPlansByProductId(product.id);
+    for await (const item of plans) {
+      const result = subscriptionProductMetadataValidator.validate({
+        ...product.metadata,
+        ...item.metadata,
+      });
+
+      if (result?.error) {
+        const msg = `handleProductUpdatedEvent - Plan "${item.id}"'s metadata failed validation; Product ${product.id} updated.`;
+        this.log.error(msg, { item: product, error: result.error });
```
chenba, line 481: s/item/plan/ ?

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -32,6 +32,8 @@ const eventInvoicePaymentFailed = require('../../payments/fixtures/stripe/event_
 const eventCustomerUpdated = require('../../payments/fixtures/stripe/event_customer_updated.json');
 const eventCustomerSubscriptionUpdated = require('../../payments/fixtures/stripe/event_customer_subscription_updated.json');
 const eventCustomerSourceExpiring = require('../../payments/fixtures/stripe/event_customer_source_expiring.json');
```
chenba, line 34: I can't find the tests for the event handler functions.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: When I wrote the comment above I thought this patch was going to invalidate the products cache on the product updated event handler; I noticed that it is not.  In that case, we should fetch just one product here instead of all of them, since we have the product id.

chenba (changes requested):
A few nits, and 
 - I didn't see any tests for the webhook event handler functions
 - there's no need to fetch all the products from Stripe and then filter locally since we have the product id already

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -384,16 +384,48 @@ module.exports.subscriptionsSubscriptionListValidator = isA.object({
...
+    playStoreLink: isA.string().uri().optional(),
     productSet: isA.string().optional(),
     productOrder: isA.number().optional(),
-    iconURL: isA.string().optional(),
-    upgradeCTA: isA.string().optional(),
-    downloadURL: isA.string().optional(),
-    appStoreLink: isA.string().optional(),
-    playStoreLink: isA.string().optional(),
   })
+  .with('productOrder', 'productSet')
```
chenba, line 405: @meandavejustice Hey sorry but we were wrong about this.  We want allow a plan to inherit a productSet from the product but not have a productOrder set on it.

meandavejustice: updated

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
...
   }
 
+  /**
+   * Validate plan metadata on update
+   */
+  async handlePlanUpdatedEvent(request: AuthRequest, event: Stripe.Event) {
+    const plan = event.data.object as Stripe.Plan;
+    if (typeof plan.product !== 'string') return;
+    const product = await this.stripeHelper.fetchProductById(plan.product);
+    if (!product) {
```
chenba, line 442: I don't think this condition is possible.  If `stripe.products.retrieve` returns a product, then `abbrevProductFromStripeProduct` would return an object.  If `stripe.products.retrieve` fails/throws, then the `abbrevProductFromStripeProduct` return statement won't be reached anyway.

In order for this to make sense, we'd need to try-catch in `fetchProductById` and return a null in the catch.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -425,6 +432,64 @@ export class StripeWebhookHandler extends StripeHandler {
     await this.updateCustomer(uid, email);
   }
 
+  /**
+   * Validate plan metadata on update
+   */
+  async handlePlanUpdatedEvent(request: AuthRequest, event: Stripe.Event) {
+    const plan = event.data.object as Stripe.Plan;
+    if (typeof plan.product !== 'string') return;
```
chenba, line 440: Does this actually happen?  Stripe should be pretty clear about what they are passing in the webhook event.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +517,78 @@ describe('StripeWebhookHandler', () => {
...
+      beforeEach(() => {
+        scopeContextSpy = sinon.fake();
+        scopeSpy = {
+          setContext: scopeContextSpy,
+        };
+        sandbox.replace(Sentry, 'withScope', (fn) => fn(scopeSpy));
+      });
+
+      it('throws a sentry error if the update event data is invalid', async () => {
+        const updatedEvent = deepCopy(eventProductUpdated);
```
chenba, line 531: Don't know about other devs, but I usually expect mock data to be valid by default.  Either way though, since it is not valid, maybe update the name to say so?

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +517,78 @@ describe('StripeWebhookHandler', () => {
...
+        await StripeWebhookHandlerInstance.handleProductUpdatedEvent(
+          {},
+          updatedEvent
+        );
+
+        assert.isTrue(scopeContextSpy.called, 'Expected to call Sentry');
```
chenba, line 537: Would be nice if we verify that `fetchPlansByProductId` was called.  Also would be nice if we know that number of Sentry calls equals to the number of mocked plans returned by `fetchPlansByProductId` that will fail validation.  I don't remember seeing other tests verifying the `scope.setContext` and `Sentry.captureMessage` calls; this seems like a good place for that.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +517,78 @@ describe('StripeWebhookHandler', () => {
...
+        updatedEvent.data.object.metadata = {
+          'product:termsOfServiceDownloadURL:en-us':
+            'https://FAIL.net/legal/mozilla_vpn_tos',
+        };
+        await StripeWebhookHandlerInstance.handlePlanUpdatedEvent(
+          {},
+          updatedEvent
+        );
+
+        assert.isTrue(scopeContextSpy.called, 'Expected to call Sentry');
```
chenba, line 575: We should check that `fetchProductById` was called.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +517,78 @@ describe('StripeWebhookHandler', () => {
...
+        updatedEvent.data.object.metadata[
+          'product:termsOfServiceDownloadURL:en-us'
+        ] = 'https://accounts-static.cdn.mozilla.net/legal/mozilla_vpn_tos';
+        await StripeWebhookHandlerInstance.handlePlanUpdatedEvent(
+          {},
+          updatedEvent
+        );
+
+        assert.isTrue(scopeContextSpy.notCalled, 'Expected not to call Sentry');
+      });
```
chenba, line 589: There's an early return when the product is not found.  I don't see that tested in these tests.

chenba (changes requested):
Thanks for adding the tests and update the validation for 'productSet' and 'productOrder'.  Please see comments regarding fetchProductById and the tests.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
meandavejustice: image

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
meandavejustice: I will make the mock event valid and set it to invalid when necessary in tests

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1054,6 +1056,33 @@ export class StripeHelper {
...
+  fetchPlansByProductId(productId: string): Stripe.ApiListPromise<Stripe.Plan> {
+    return this.stripe.plans.list({
+      active: true,
+      product: productId,
+      expand: ['data.product'],
+    });
+  }
+
+  /**
+   * Fetches all a product by it's ID
```
chenba, line 1071: Missing word(s)?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: Sure, I understand what Stripe declared the type of Plan.product to be.  But in https://stripe.com/docs/api/events/types, it states that object for the event is a https://stripe.com/docs/api/plans/object, which has a expandable `product`.  afaik, the expandable properties are not automatically expanded.

If we were to handle the types accordingly here, then we should not return when `product` is a Product.  We should instead use the product, since that's the object we want on line 441.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -1322,8 +1302,94 @@ describe('StripeHelper', () => {
...
+        }),
+        {}
+      );
+      stripeHelper.stripe = mockStripe;
+    });
+
+    it('returns null if there is an issue fetching the product', async () => {
+      stripeHelper.stripe.products.retrieve.returns(
+        Promise.reject(new Error('product does not exist'))
+      );
```
chenba, line 1353: I think this can use `stripeHelper.stripe.products.retrieve.rejects`.  Just fyi; this works too.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -1322,8 +1302,94 @@ describe('StripeHelper', () => {
...
+      id: productId,
+      name: productName,
+      metadata: {
+        'product:termsOfServiceURL': termsOfServiceURL,
+        'product:privacyNoticeURL': privacyNoticeURL,
+      },
+    };
+    beforeEach(() => {
+      mockStripe = Object.entries({
+        plans: mockPlan,
```
chenba, line 1338: I don't think `plans.retrieve` is needed for the tests here.  There's no need to mock it.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +511,97 @@ describe('StripeWebhookHandler', () => {
...
+      let scopeContextSpy, scopeSpy;
+      beforeEach(() => {
+        scopeContextSpy = sinon.fake();
+        scopeSpy = {
+          setContext: scopeContextSpy,
+        };
+        sandbox.replace(Sentry, 'withScope', (fn) => fn(scopeSpy));
+      });
+
+      it('throws a sentry error if the update event data is invalid', async () => {
```
chenba, line 524: Since a product update triggers multiple validations, we should verify the correct number of errors are reported.

I also don't see that the tests verifying that Sentry.captureMessage has been called.

chenba (changes requested):
Please see inline comments.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: If we do expect Stripe to return any of the types in the sum type for `Stripe.Plan.product`, then we need to handle the case when it is a `Stripe.Product`.  That's a valid value and returning early would be a bug.

However, I do not think that value is ever a `Stripe.Product` or `Stripe.DeletedProduct`, since that property does not auto-expand.

If you are using that type check to get around the type checker, then use a type assertion instead.  This is one of the times where we know more than the type checker.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -486,6 +512,101 @@ describe('StripeWebhookHandler', () => {
...
+        captureMessageSpy = sinon.fake();
+        scopeContextSpy = sinon.fake();
+        scopeSpy = {
+          setContext: scopeContextSpy,
+        };
+        sandbox.replace(Sentry, 'withScope', (fn) => fn(scopeSpy));
+        sandbox.replace(Sentry, 'captureMessage', captureMessageSpy);
+      });
+
+      it('throws a sentry error if the update event data is invalid', async () => {
```
chenba, line 527: I think this test would still pass if `handleProductUpdateEvent` returns after validating the first plan.  Not a big deal I suppose, since an implementation like that is unlikely to get past review.  But this test doesn't verify if all the plans were validated.

chenba (changes requested):
Added a comment to conversation here: https://github.com/mozilla/fxa/pull/8437/files#diff-7864793ea6a48f6ad7ea612c6083bb230bab73cdcd742ec1311c77d7513bd1dcR440

Also, please let me know if you'd like to pair on this.

chenba (approved):
(Dave and I paired yesterday to address the two comments from the previous review.  I did not review the entire PR again.)

chenba:
@meandavejustice Let's not merge until we verified that the patch won't break stage and prod.

chenba:
I've verified the plans in the two products in production will pass the metadata validation from this patch.

chenba (changes requested):
I just noticed that FPN on prod does not have a privacy notice download URL and the validation is not catching that.

PR #8439, opened 2021-04-29
chore(settings): remove settings_version from metrics

### LINKED ISSUE #6474: [post-beta] Remove `settings_version` property from metrics events
In FXA-1737, we added an amplitude event property `settings_version` so we can compare amplitude events in old settings vs new settings.

Once "new settings" is just "settings," we can remove this property from the events.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2545)
┆Issue Number: FXA-2545

dannycoates:
closes #6474

### DIFF
--- a/packages/fxa-content-server/server/lib/routes/post-metrics.js
+++ b/packages/fxa-content-server/server/lib/routes/post-metrics.js
@@ -38,7 +38,6 @@ const {
   NEWSLETTERS: NEWSLETTERS,
   OFFSET: OFFSET_TYPE,
   REFERRER: REFERRER_TYPE,
-  SETTINGS_VERSION,
   STRING: STRING_TYPE,
   SYNC_ENGINES: SYNC_ENGINES_TYPE,
   TIME: TIME_TYPE,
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/server/lib/routes/post-metrics.js
```diff
@@ -142,7 +141,6 @@ const BODY_SCHEMA = {
     })
     .required(),
   service: STRING_TYPE.regex(SERVICE_PATTERN).required(),
-  settingsVersion: SETTINGS_VERSION.optional(),
```
vbudhram, line 145: This shouldn't reject requests with that have this property right? Which might happen if were are switching between versions in the deploy.

vbudhram (approved):
@dannycoates LGTM 👍🏽

packages/fxa-content-server/server/lib/routes/post-metrics.js
dannycoates: im not 100% sure we allow unspecified fields. we could spread this over two trains but i don't think it's that terrible to lose a few of these while the train is mid-deploy 🤷‍♂️

PR #8441, opened 2021-04-29
feat(auth-server): add stripe metric details

### LINKED ISSUE #8358: Add statsd monitoring for Stripe API calls
For example, failures to renew

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3404)
┆Issue Number: FXA-3404

bbangert:
Because:

* We want to know more details about Stripe calls, and whether they are
  throwing errors.

This commit:

* Adds a counter for the stripe call along with whether it resulted in
  an error.

Closes #8358

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -111,12 +111,13 @@ export class StripeHelper {
   private webhookSecret: string;
   private stripe: Stripe;
   private redis: ioredis.Redis | undefined;
+  private statsd: StatsD;
   public currencyHelper: CurrencyHelper;
 
   /**
    * Create a Stripe Helper with built-in caching.
    */
-  constructor(log: Logger, config: ConfigType, statsd?: StatsD) {
+  constructor(log: Logger, config: ConfigType, statsd: StatsD) {
     this.log = log;
     this.customerCacheTtlSeconds = config.subhub.customerCacheTtlSeconds;
     this.plansAndProductsCacheTtlSeconds = config.subhub.plansCacheTtlSeconds;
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -148,11 +149,16 @@ export class StripeHelper {
...
 
-    if (statsd) {
-      this.stripe.on('response', (response) => {
-        statsd.timing('stripe_request', response.elapsed);
+    this.statsd = statsd;
+    this.stripe.on('response', (response) => {
+      statsd.timing('stripe_request', response.elapsed);
+      // Note that we can't record the method/path as a tag
+      // because ids are in the path which results in too great
+      // of cardinality.
```
chenba, line 157: 👍

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -148,11 +149,16 @@ export class StripeHelper {
...
+      statsd.increment('stripe_call', {
+        error: (response.status >= 500).toString(),
```
chenba, line 159: Not interested in 4xx errors?  Might be misleading for anyone just looking at the data in InfluxDB when the label is 'error'; they wouldn't know it's >= 500.

chenba (approved):
r+ with comments

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Not really, mainly because it can happen easily and regularly, and just seeing a counter on a dashboard that it happened doesn't seem terribly useful. Ideally we want to know whether a large chunk of our requests are working correctly, or throwing errors. Debugging from a 4xx should be an error in Sentry.

PR #8392, opened 2021-04-26
chore(deps-dev): downgrade @storybook/addons from 6.2.9 to 5.3.19

lmorchard:
This reverts commit 72ac7648b28e16ebc79189f34d762426d7f547d1, reversing
changes made to ccd8007027a1db2657e507eec91dca52cdb84e94.

fixes #8391

PR #8449, opened 2021-04-30
Revert "chore(deps-dev): bump @storybook/addon-links from 6.2.8 to 6.2.9"

chenba:
Reverts mozilla/fxa#8447

### DIFF
--- a/packages/fxa-admin-panel/package.json
+++ b/packages/fxa-admin-panel/package.json
@@ -61,7 +61,7 @@
     "@babel/core": "^7.10.3",
     "@rescripts/cli": "0.0.16",
     "@storybook/addon-actions": "^5.3.19",
-    "@storybook/addon-links": "^6.2.9",
+    "@storybook/addon-links": "^5.3.19",
     "@storybook/addons": "^5.3.19",
     "@storybook/preset-create-react-app": "^3.1.7",
     "@storybook/react": "^5.3.19",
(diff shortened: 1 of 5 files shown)


vbudhram (approved):

packages/fxa-admin-panel/package.json
```diff
@@ -61,7 +61,7 @@
     "@babel/core": "^7.10.3",
     "@rescripts/cli": "0.0.16",
     "@storybook/addon-actions": "^5.3.19",
-    "@storybook/addon-links": "^6.2.9",
+    "@storybook/addon-links": "^6.2.8",
```
lmorchard, line 64: This needs to be reverted back to 5.3.19

lmorchard (changes requested):

packages/fxa-admin-panel/package.json
lmorchard: I'm confused how dependabot managed to go from 6.2.8 to 6.2.9 when [the my previous revert was down to 5.3.19](https://github.com/mozilla/fxa/pull/8392). I think 6.2.8 will still be broken

lmorchard: Ohhh, I see... my revert was for `@storybook/addons` and this is for `@storybook/addon-links` which somehow got upgraded earlier?

Issue #8341, opened 2021-04-20
[FxA-Settings] The date and time are not displayed in Verification Code email


data-sync-user:
**Build**
 Stage [Train 1.204.3]

**Affected platforms**
 Windows 10 64-bit

**Steps to reproduce**

1. Navigate to [https://accounts.stage.mozaws.net/settings/](https://accounts.stage.mozaws.net/settings/) and follow the account creation flow.
1. Go to the email address inbox and open the "Verification code: XXXXXX" email.
1. Check for the date and time.

**Expected results**
 The date and time is successfully displayed in the email under the IP address.

**Actual Results**
 The date and time is not displayed in the Verification code email.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3394)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/16112/Screenshot_60.png">Screenshot_60.png</a>
┆Issue Number: FXA-3394

data-sync-user:
➤ Wil Clouser commented:

We think this is correct and we don't think we've changed this email in a long time.  Is there something indicating we'd have a date/time there?

Yogita98:
Hey @mozilla/fxa-devs
I'd like to work on this issue. I just pulled the `verifyLoginCode` email template and saw that we have a date field just below the IP address as shown below:
![image](https://user-images.githubusercontent.com/37609519/115582606-c36fe480-a2e6-11eb-8a0f-f29ec398bb6a.png)

But the email that user gets doesn't have one as shown:
![image](https://user-images.githubusercontent.com/37609519/115583038-25c8e500-a2e7-11eb-905c-eb7a1709931d.png)

clouserw:
Thanks for your initiative.  That looks like an email for signing up a new account and an email for logging in to an existing one.

My original question for QA is still relevant though before we move forward on anything - is there a spec or document that says we should have a date code in this email?

Yogita98:
Ohhh I see! My bad and sorry for that. I guess the template that is being used is `verifyShortCode` 

No problem I'll wait for the reply. Thanks for the guidance!

data-sync-user:
➤ Vasilica Tamas commented:

[~wclouser@mozilla.com] This issue was logged based on the consistency compared with the other emails. As the “Sign-in code for Firefox” email which contains the verification code with limited availability displays the date and time, it would be helpful also to display these time details in the Registration email too just to emphasize the activation limited time.

[...]

PR #8452, opened 2021-05-02
fix(emails): Add date and time in verifyShortCodeEmail template

Yogita98:
## Because

- Date and time could be an important field to display so as to emphasize the activation limited time to the user

## This pull request

- Aims to display the date and time component in the email template

Closes: #8341 

## Checklist

## Screenshots (Optional)
image


Yogita98:
Hey @mozilla/fxa-devs
Can anyone review my PR? I have also modified the tests accordingly. One thing that I'd like to highlight is that the timeout set for the tests [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/scripts/test-local.sh#L13) is just 5s and maybe because of which timeout errors are being thrown, I had to make it 30s for the tests to pass locally.

LZoog (approved):
Before:
image


After:
image

LGTM, thanks @Yogita98! 😃

LZoog:
> One thing that I'd like to highlight is that the timeout set for the tests here is just 5s and maybe because of which timeout errors are being thrown, I had to make it 30s for the tests to pass locally.

Huh, that's strange. It seems to be passing fine in Circle CI so we'll leave it but that's good to know. 🤔

Yogita98:
Okay no worries! And thanks for reviewing my PR @LZoog

PR #8465, opened 2021-05-03
fix(auth): Pass statsd argument of correct type to StripeHelper in pa…

biancadanforth:
…ypal-processor

Because:

* The patch for #8358 made the `statsd` param required for the StripeHelper constructor, and we need pass the type check in the PayPal processor script.

This commit:

* Sets the type of `statsd` when it's disabled in the auth server config to be of type StatsD.

## Checklist


chenba (approved):

chenba:
I think the error was about the missing argument not the argument's type.  But what's stated in the PR description is also accurate.

PR #8467, opened 2021-05-03
chore(deps): upgrade to Storybook v6 across all packages

### LINKED ISSUE #7995: Upgrade to Storybook v6
A [straight major version upgrade of Storybook v5 to v6](https://github.com/mozilla/fxa/commit/37522c5a6c6a35a79c1da953c8995a43e8e665b9) breaks the storybook build - albeit in a quiet way that doesn't fail CI.

There is [a migration guide](https://storybook.js.org/blog/storybook-6-migration-guide/) for upgrading from v5 to v6. But, there is [a lot of custom tinkering going on in our shared Storybook config](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/configs/storybooks.js) that will take non-trivial effort to rework.

I took an hour or so to try troubleshooting it, but I think the built-in webpack config that our code modifies has just changed too much for it to be quick & easy.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3274)
┆Issue Number: FXA-3274

lmorchard:
Because:

* We want to upgrade to Storybook v6 to catch future upgrades and start
  using new features

This commit:

* Upgrades Storybook dependencies to v6.2.9

* Drops @storybook/preset-create-react-app addon from fxa-admin-panel
  and fxa-settings to keep webpack configs consistent for customization

* Reworks fxa-react/configs/storybooks.js to simplify and drop tweaks
  that seem redundant vs Storybook v6

fixes #7995


lmorchard:
Seems to work for me locally, but opening this as draft initially to see if this does actually result in a working Storybook build in CI.

packages/fxa-admin-panel/.storybook/main.js
```diff
@@ -4,9 +4,5 @@
 
 module.exports = {
   stories: ['../src/**/*.stories.tsx'],
```
lmorchard, line 6: Weird thing about fxa-admin-panel: It has a Storybook config, but apparently no stories?

packages/fxa-settings/.storybook/main.js
```diff
@@ -4,9 +4,5 @@
 
 module.exports = {
   stories: ['./design-guide/main.stories.tsx', '../src/**/*.stories.tsx'],
-  addons: [
-    '@storybook/addon-actions',
-    '@storybook/addon-links',
-    '@storybook/preset-create-react-app',
```
lmorchard, line 10: I'm not sure what this addon was doing for us, but a lot of things started working when I took it out.

Weirdly, fxa-react and fxa-payment-server weren't using it, but fxa-settings and fxa-admin-panel were. So, it resulted in a very different initial Webpack config when trying to apply the shared custom config patches.

We might want to revisit reintroducing it - but across the board rather than half-and-half. It promises to "just work" for CRA apps, so maybe it would replace our tweaks entirely at some point? 🤷‍♂️

lmorchard (commented):
Adding a couple of comments in self-review while I wait for the build to work

lmorchard:
Hmm, weird error building fxa-react that I don't see locally or when ssh'ing into CircleCI :(

Capture

jodyheavener:
@lmorchard does the fxa-react `dependencies` need `@svgr/webpack`?

lmorchard:
Rebased against main to pick up yarn.lock changes

lmorchard:
> but should @svgr/webpack be dependencies or devDependencies?

I think devDependencies - it's used by webpack and builds React components around SVGs on the fly, so it's not actually itself bundled into the compiled app

packages/fxa-react/configs/storybooks.js
```diff
@@ -3,80 +3,79 @@
...
-  config.module.rules = [{ oneOf: config.module.rules }];
-  config.resolve.alias = Object.assign(
-    config.resolve.alias,
-    additionalJSImports
-  );
-
-  return config;
-};
+const customizeWebpackConfig = ({ config }) => ({
+  ...config,
```
lmorchard, line 16: Oh, and I don't think I *had* to totally rewrite how this function customizes the webpack config. But, it's where I ended up after a long & twisty journey of trying to understand what was going on with config from different packages. Lots of console logging and re-adding tweaks one-by-one until things worked.

lmorchard:
Hmm, also looks like Storybook v6 doesn't support the pipe-delimiter in story names any more (e.g. "Design Guide|Settings"). Would kind of like to defer that to a follow up fix if it can get this PR over the merge hump faster and avoid another Dependabot round breaking things again

lmorchard:
Alright, I'm just going to open this up for review. Aiming to get this up to v6 without major breakage. Would prefer to spin off followup issues for refinements for minor issues. We can also consider starting to actually convert stories over to the cleaner v6 format along the way

dannycoates (approved):
Looks great!

PR #8476, opened 2021-05-04
Update 123Done and add support for heroku

vbudhram:
## Because

- 123done config is a bit outdated to support heroku
- If we go the route of hosting this service in heroku we will need to update AWS to point to new [trusted](stage-123done.herokuapp.com/) and [untrusted](http://stage-123done-untrusted.herokuapp.com/) versions.

## This pull request

- Exposes config the same way other services work
- Add docs for pushing 123done to heroku

Closes: #8491

## Checklist


packages/123done/config-stage-untrusted.json
```diff
@@ -0,0 +1,8 @@
+{
+  "cookieName": "321done",
```
dannycoates, line 2: `cookie_name`

dannycoates (approved):
LGTM. Would it be useful to add anything about deploying to heroku to the readme?

Issue #8497, opened 2021-05-06
delete-account.js is broken


jbuck:
```
$ node scripts/delete-account.js 
(node:27) UnhandledPromiseRejectionWarning: TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at new CurrencyHelper (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/currencies.js:10:44)
    at ContainerInstance.getServiceValue (/fxa/node_modules/typedi/ContainerInstance.js:207:21)
    at ContainerInstance.get (/fxa/node_modules/typedi/ContainerInstance.js:51:21)
    at Function.Container.get (/fxa/node_modules/typedi/Container.js:38:36)
    at new StripeHelper (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/stripe.js:117:50)
    at createStripeHelper (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/stripe.js:1548:12)
    at /fxa/packages/fxa-auth-server/dist/fxa-auth-server/scripts/delete-account.js:73:20
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:27) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:27) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3453)
┆Issue Number: FXA-3453

PR #8511, opened 2021-05-06
fix(scripts): fix dependencies for delete-account

chenba:
Because:
 - with PayPal enabled the delete account script now need dependencies
   that depends on the DI container to have such dependencies

This commit:
 - create and add dependencies into the DI container in the perfect
   order

Closes: #8497

### DIFF
--- a/packages/fxa-auth-server/scripts/delete-account.js
+++ b/packages/fxa-auth-server/scripts/delete-account.js
@@ -20,6 +20,8 @@
 
 'use strict';
 
+const { StatsD } = require('hot-shots');
+const { Container } = require('typedi');
 const readline = require('readline');
 const P = require('../lib/promise');
 const config = require('../config').getProperties();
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/scripts/delete-account.js
```diff
@@ -20,6 +20,8 @@
 
 'use strict';
 
+const { StatsD } = require('hot-shots');
+const { Container } = require('typedi');
```
vbudhram, line 24: Is it possible to add some tests for this in https://github.com/mozilla/fxa/tree/f4dc15a81d8bfa2df44b268367702cf7500c643f/packages/fxa-auth-server/test/scripts?

vbudhram (approved):
@chenba I tested this with a local account and it got deleted

`NODE_ENV=dev ts-node scripts/delete-account.js a@aa.com`

We should try to add some basic test coverage, otw LGTM 👍🏽

packages/fxa-auth-server/scripts/delete-account.js
chenba: Testing the delete-account script would require quite a bit of setup.  I'll file a jira for it; I don't want to block the fix because of it.

Issue #8527, opened 2021-05-10
Failing test in main


vbudhram:
Looks like this test is failing in multiple pull requests

`can change primary email and login - settings change email`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3471)
┆Issue Number: FXA-3471

PR #8528, opened 2021-05-10
fix(tests): rerun settings functional tests on failure and mark beforeEach as failure

vbudhram:
## Because

- Re-running failed test will better help confirm that the test is not flake
- Test failing in the `beforeEach` for functional tests were not marked as failures

## This pull request

- Re-runs those tests
- Marks `beforeEach` as failures

Closes: #8527

## Checklist

### DIFF
--- a/packages/fxa-content-server/tests/functional/settings/change_email.js
+++ b/packages/fxa-content-server/tests/functional/settings/change_email.js
@@ -38,7 +38,6 @@ const {
   testSuccessWasShown,
   type,
   visibleByQSA,
-  waitForUrl,
 } = FunctionalHelpers;
 
 registerSuite('settings change email', {
@@ -52,8 +51,8 @@ registerSuite('settings change email', {
         .then(fillOutEmailFirstSignUp(email, PASSWORD))
         .then(testElementExists(selectors.CONFIRM_SIGNUP_CODE.HEADER))
         .then(fillOutSignUpCode(email, 0))
-        .then(waitForUrl((url) => !url.includes('settings')))
-        .then(testElementExists(selectors.SETTINGS.AVATAR_ROW.HEADER))
+
+        .then(visibleByQSA(selectors.SETTINGS.PROFILE_HEADER))
         .then(click(selectors.EMAIL.MENU_BUTTON))
 
         .then(addAndVerifySecondaryEmail(secondaryEmail))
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/tests/intern.js
```diff
@@ -174,6 +174,12 @@ if (args.firefoxBinary) {
 
 const failed = [];
 
+intern.on('suiteEnd', (test) => {
```
dannycoates, line 177: why do we need both 'suiteEnd' and 'testEnd'?

packages/fxa-content-server/tests/functional/settings/change_email.js
```diff
@@ -52,8 +51,8 @@ registerSuite('settings change email', {
         .then(fillOutEmailFirstSignUp(email, PASSWORD))
         .then(testElementExists(selectors.CONFIRM_SIGNUP_CODE.HEADER))
         .then(fillOutSignUpCode(email, 0))
-        .then(waitForUrl((url) => !url.includes('settings')))
```
dannycoates, line 55: I added this because the following `testElementExists` was (sometimes) firing before the page navigated to `/settings` but apparently it wasn't enough. IDK that we need it but I'm not optimistic your change will help.

packages/fxa-settings/scripts/test-ci.sh
```diff
@@ -31,4 +31,13 @@ _scripts/check-url.sh localhost:3030/bundle/app.bundle.js
 
 cd packages/fxa-content-server
 mozinstall /firefox.tar.bz2
-node tests/intern.js --suites="settings" --output="../../artifacts/tests/results.xml" --firefoxBinary=./firefox/firefox
+node tests/intern.js \
+  --suites="settings" \
+  --output="../../artifacts/tests/results.xml" \
+  --firefoxBinary=./firefox/firefox \
+  || \
+node tests/intern.js \
```
dannycoates, line 39: I'd rather not but 🤷 :sigh:

dannycoates (approved):
worth a shot

packages/fxa-content-server/tests/functional/settings/change_email.js
vbudhram: Hrm let me try the `visibleByQSA` selector, maybe that would help.

packages/fxa-content-server/tests/intern.js
vbudhram: Yea, unfortunatley `testEnd` event doesn't get triggered if a failure in the `beforeEach`.

PR #8543, opened 2021-05-11
fix(auth-server): paypal customer fix script

### LINKED ISSUE #8544: Fix paypal customers affected by credit note misfire
Some users were affected by a credit note being issued to the account balance instead of refunding the paypal transaction id. A one-off script is needed to fix these customers, zero the account balance, and cancel any subscriptions not cancelled when they were refunded.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3479)
┆Issue Number: FXA-3479

bbangert:
Because:

* Credit notes issued as account balance aren't refunded in paypal
  correctly.

This commit:

* Locates invoices since launch that had a credit note applied that was
  credited to the balance instead of refunded and fixes the balance,
  issues a paypal refund, and cancels the subscription.

Closes #8544

## Checklist

### DIFF
--- a/packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
+++ b/packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
@@ -1,0 +1,189 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { AuthLogger } from 'fxa-auth-server/lib/types';
+import { ACTIVE_SUBSCRIPTION_STATUSES } from 'fxa-shared/subscriptions/stripe';
+import { StatsD } from 'hot-shots';
+import stripe from 'stripe';
+import Container from 'typedi';
+
+import { CurrencyHelper } from '../lib/payments/currencies';
+import { PayPalHelper } from '../lib/payments/paypal';
+import { PayPalClient } from '../lib/payments/paypal-client';
+import { STRIPE_INVOICE_METADATA, StripeHelper } from '../lib/payments/stripe';
+import { configureSentry } from '../lib/sentry';
+
+const config = require('../config').getProperties();
+
+class PayPalFixer {
+  private stripe: stripe;
+
+  constructor(
+    private log: AuthLogger,
+    private stripeHelper: StripeHelper,
+    private paypalHelper: PayPalHelper
+  ) {
+    this.stripe = (this.stripeHelper as any).stripe as stripe;
+  }
+
+  private activeSubscriptions(
+    subscriptions: stripe.Subscription[] | undefined
+  ) {
+    if (!subscriptions) {
+      return [];
+    }
+    return subscriptions.filter((sub) =>
+      ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status)
+    );
+  }
+
+  private async *nonRefundedPaypalInvoices(limit?: number) {
+    const startDate = new Date(2021, 3, 20).getTime() / 1000;
+    let count = 0;
+    for await (const invoice of this.stripe.invoices.list({
+      limit: 100,
+      status: 'paid',
+      collection_method: 'send_invoice',
+      created: { gt: startDate },
+    })) {
+      if (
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+      yield invoice;
+      count++;
+      if (limit && count >= limit) {
+        break;
+      }
+    }
+  }
+
+  async zeroAccountBalance(customer: stripe.Customer) {
+    if (customer.balance >= 0) {
```
bbangert, line 67: Positive amounts mean they owe us, negative amounts mean we owe them, which are the only ones we want to zero.

packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+        balance: customer.deleted ? 0 : customer.balance,
+        subs: subs.length,
+        deleted: !!customer.deleted,
+      });
+      i++;
+
+      if (!customer.deleted) {
+        await this.zeroAccountBalance(customer);
+      }
+      if (subs.length === 1) {
```
dannycoates, line 126: subs can't ever be > 1? ... just checking

bbangert: They can, I've run a read-only version of this and only one customer managed to get 2 subs. We'll need to handle that customer separately so we don't want to cancel the sub just yet.

packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+      return;
+    }
+    const zeroAmount = Math.abs(customer.balance);
+    await this.stripe.customers.createBalanceTransaction(customer.id, {
+      amount: zeroAmount,
+      currency: customer.currency!,
+    });
+  }
+
+  async issueRefund(invoice: stripe.Invoice) {
```
bbangert, line 77: This is almost identical to the issueRefund inside the paypalHelper, however since this runs as a script the stripeHelper updateInvoice here needs an await on it to ensure we don't skip one at the end.

packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+        invoiceId: invoice.id,
+        customerId: customer.id,
+        balance: customer.deleted ? 0 : customer.balance,
+        subs: subs.length,
+        deleted: !!customer.deleted,
+      });
+      i++;
+
+      if (!customer.deleted) {
+        await this.zeroAccountBalance(customer);
```
dannycoates, line 124: Should this come after `issueRefund`, in case that fails? tbh i don't know enough context to know which should go first

packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+        invoice.post_payment_credit_notes_amount === 0
+      ) {
+        continue;
+      }
+      yield invoice;
+      count++;
+      if (limit && count >= limit) {
+        break;
+      }
+    }
```
chenba, line 63: If I'm reading the limits correctly, this will only process a max of 100 customers per execution.  Are we going to run this script multiple times then?

bbangert: That's not actually used, I can remove it. We actually get everyone the first time, because we're using async iteration, and its 100 per batch.

chenba (approved):
Make sense to me

biancadanforth (approved):
Just added a few clarifying questions. Logic makes sense to me.

packages/fxa-auth-server/scripts/paypal-refund-fixer.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+  }
+
+  async issueRefund(invoice: stripe.Invoice) {
+    const transactionId = this.stripeHelper.getInvoicePaypalTransactionId(
+      invoice
+    );
+    if (!transactionId) {
+      return;
+    }
```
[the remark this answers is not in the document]
bbangert: No reason, there must be one here, but this makes TypeScript happy since the method signature says it could be undefined.

PR #8553, opened 2021-05-12
feat(sql): Use `JSON_OBJECTAGG` instead of aggregateNameValuePairs in session sql

vbudhram:
## Because

- MySQL 5.7 has some neat JSON functions
- We can use JSON_OBJECTAGG to do the same thing we are doing in [code](https://github.com/mozilla/fxa/blob/491ac1f553c928923ef0710771d66164c693acd2/packages/fxa-auth-db-mysql/lib/db/util.js#L125)

## This pull request

- Updates the `sessionTokenWithDevice` stored procedure to use `JSON_OBJECTAGG`

Closes: https://github.com/mozilla/fxa/issues/8573

## Checklist

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/mysql.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mysql.js
@@ -513,8 +513,6 @@ module.exports = function (log, error) {
     return this.readAllResults(SESSION_DEVICE, [id]).then((results) => {
       if (results.length === 0) {
         throw error.notFound();
-      } else if (!results[0].tokenData) {
-        throw error.notFound();
       }
       return results[0];
     });
--- a/packages/fxa-auth-db-mysql/lib/db/mysql.js
+++ b/packages/fxa-auth-db-mysql/lib/db/mysql.js
@@ -507,25 +507,17 @@ module.exports = function (log, error) {
   //          ut.tokenVerificationId, ut.mustVerify
   // Where  : t.tokenId = $1 AND t.uid = a.uid AND t.tokenId = d.sessionTokenId AND
   //          t.uid = d.uid AND t.tokenId = u.tokenId
-  var SESSION_DEVICE = 'CALL sessionWithDevice_18(?)';
+  var SESSION_DEVICE = 'CALL sessionWithDevice_19(?)';
 
   MySql.prototype.sessionToken = function (id) {
-    return this.readAllResults(SESSION_DEVICE, [id])
-      .then((rows) =>
-        dbUtil.aggregateNameValuePairs(
-          rows,
-          'deviceId',
-          'deviceCommandName',
-          'deviceCommandData',
-          'deviceAvailableCommands'
-        )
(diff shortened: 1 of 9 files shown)


packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
```diff
@@ -0,0 +1,75 @@
+--
+-- This migration uses the `JSON_OBJECTAGG` function to aggregate available device commands.
+-- Previously we were using a helper function `aggregateNameValuePairs` introduced in
+-- https://github.com/mozilla/fxa-auth-db-mysql/pull/354
+--
+-- Some things to note in this migration, `JSON_OBJECTAGG` errors if any key is NULL,
+-- which might happen when joining deviceCommands and deviceCommandIdentifiers tables.
```
vbudhram, line 7: Curious if someone knows a way to do this without having to use `JSON_REMOVE`, I poked around a lot and couldn't find a solution that wasn't ugly.

dannycoates: I can't think of one that avoids nulls since it's a left join. a subselect could do the objectagg at a lower level but idk that it'd be better. it'd also return null instead of {}.

```
...
  LEFT JOIN (
    (SELECT dc.uid, dc.deviceId, JSON_OBJECTAGG(ci.commandName, dc.commandData) as deviceAvailableCommands FROM 
    deviceCommands AS dc FORCE INDEX (PRIMARY)
    INNER JOIN deviceCommandIdentifiers AS ci FORCE INDEX (PRIMARY)
      ON ci.commandId = dc.commandId
    WHERE ci.commandName IS NOT NULL
    ) as dac
  ) ON (dac.uid = d.uid AND dac.deviceId = d.id)
...
```

packages/fxa-auth-db-mysql/lib/db/mysql.js
```diff
@@ -507,25 +507,17 @@ module.exports = function (log, error) {
...
-      .then((results) => {
-        if (results.length === 0) {
-          throw error.notFound();
-        }
-        return results[0];
-      });
+    return this.readAllResults(SESSION_DEVICE, [id]).then((results) => {
+      if (results.length === 0) {
+        throw error.notFound();
+      } else if (!results[0].tokenData) {
```
dannycoates, line 516: when do we get a result with null tokenData?

dannycoates (approved):
cool!

packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
vbudhram: Your example works well and allowed me to remove the extra check for tokenData 👍🏽

vbudhram:
Closing this PR and moving out of sprint for now. I won't be able to get to it.

Issue #8223, opened 2021-04-09
[subplat][emails] Add Email Templates and Sender Function for Subscription Renewal Reminder


data-sync-user:
We need to add email templates and a function to send the subscription renewal reminder email. See FXA-3166 for the copy.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3351)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/16410/Screen+Shot+2021-05-21+at+4.20.23+PM.png">Screen Shot 2021-05-21 at 4.20.23 PM.png</a>
┆Issue Number: FXA-3351

data-sync-user:
➤ Barry Chen commented:

[~byee@mozilla.com] we need copy for this email.  Are you the person for this?

data-sync-user:
➤ Bianca Danforth commented:

[~jecook@mozilla.com] , [~crooney@mozilla.com] : There is more context in the story ticket that this ticket fulfills,  FXA-3166, (we broke the work down into smaller tasks later on); let us know if you'd still like more information.

Actually, looking at that ticket, it looks like [~byee@mozilla.com] provided us with copy ( https://docs.google.com/document/d/1PbxlKSVt41kpU63axX6D8w0L6gnUyn1kmYzAkz0zf1U/edit?usp=sharing ) that has been approved by legal (see this comment in the story ticket).

I'll remove the `needs-content` label. Thanks Jess for following up.

data-sync-user:
➤ Bianca Danforth commented:

Now that I'm actually in the weeds creating this new email, a couple of product/content questions arose in relation to the email copy ( https://docs.google.com/document/d/1PbxlKSVt41kpU63axX6D8w0L6gnUyn1kmYzAkz0zf1U/edit?usp=sharing ) that Brian provided:

  1. [~byee@mozilla.com]: Currently, no other transactional emails include a link to a product-specific SUMO page, and we don't have a Stripe metadata property ( https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-sub-platform#stripe-productplans ) to define that URL (i.e. something like `product:supportURL`). Can I file a follow-up issue for this and remove the "help center" link in this new email in the interim?
[...]

PR #8572, opened 2021-05-13
feat(auth): create script to send subscription renewal reminder emails

### LINKED ISSUE #8224: [subplat][emails] Create Script to Send Subscription Renewal Reminder Emails
For subscriptions with a renewal period >=180 days, we need to send the customer a reminder email about the upcoming renewal.  We'll need a script to do the following:

For all auto recurring prices/plans with a renewal period >=180

* get all the active subscriptions for the price/plan
* for subscriptions where the end datetime of the current period is less than or equal 14 days away
* if an email has not been already sent in the (end_of_current_period - 14_days) window
  * send the email created in FXA-3351
  * record the sent datetime in the database table created in FXA-3350

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3352)
┆Issue Number: FXA-3352

biancadanforth:
Because:
* There is a legal requirement to send a reminder email two weeks before a subscription for a plan longer than 180 days is renewed.

This commit:
* Adds a new `subscription-reminder` script intended to be run on a (bi-)daily basis that queries active subscriptions with longer plan lengths and sends renewal reminder emails as needed.
* Adds `subscriptionRenewalReminder` emailType to `EmailTypes` table with forward and reverse migrations.
  * The template for this email will be added in #8223.

Closes #8224

Co-authored-by: Barry Chen <chenba@gmail.com>

## Checklist

### DIFF
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
@@ -1,0 +1,12 @@
+--
+-- This migration adds the subscriptionRenewalReminder emailType to
+-- the emailTypes table.
+--
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('115');
+
+INSERT INTO emailTypes (emailType) VALUE ('subscriptionRenewalReminder');
+
+UPDATE dbMetadata SET value = '116' WHERE name = 'schema-patch-level';
--- a/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
+++ b/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { setupAuthDatabase } from 'fxa-shared/db';
+import { StatsD } from 'hot-shots';
+import Container from 'typedi';
+import { Logger } from 'mozlog';
+
+import error from '../error';
+import { CurrencyHelper } from '../payments/currencies';
+import { StripeHelper } from './stripe';
+import { configureSentry } from '../sentry';
+
+const config = require('../../config').getProperties();
+
+export type AdditionalSetupFn = ({ log }: { log: Logger }) => void;
+
+export async function initShared(additionalSetupFn?: AdditionalSetupFn) {
+  configureSentry(undefined, config);
+  // Establish database connection and bind instance to Model using Knex
+  setupAuthDatabase(config.database.mysql.auth);
+
+  const statsd = config.statsd.enabled
+    ? new StatsD({
+        ...config.statsd,
+        errorHandler: (err) => {
+          // eslint-disable-next-line no-use-before-define
+          log.error('statsd.error', err);
+        },
+      })
+    : (({
+        increment: () => {},
+        timing: () => {},
+        close: () => {},
+      } as unknown) as StatsD);
+  Container.set(StatsD, statsd);
+
+  const log = require('../log')({ ...config.log, statsd });
+
+  const translator = await require('../senders/translator')(
+    config.i18n.supportedLanguages,
+    config.i18n.defaultLanguage
+  );
+  const senders = await require('../senders')(
+    log,
+    config,
+    error,
+    translator,
+    statsd
+  );
+  const redis = require('../redis')(
+    { ...config.redis, ...config.redis.sessionTokens },
+    log
+  );
+  const DB = require('../db')(
+    config,
+    log,
+    require('../tokens')(log, config),
+    require('../crypto/random').base32(config.signinUnblock.codeLength)
+  );
+  let database = null;
+  try {
+    database = await DB.connect(config, redis);
+  } catch (err) {
+    log.error('DB.connect', { err: { message: err.message } });
+    process.exit(1);
+  }
+
+  const currencyHelper = new CurrencyHelper(config);
+  Container.set(CurrencyHelper, currencyHelper);
+  const stripeHelper = new StripeHelper(log, config, statsd);
+  Container.set(StripeHelper, stripeHelper);
+
+  if (additionalSetupFn) {
+    additionalSetupFn({ log });
+  }
+
+  return {
+    log,
+    database,
+    senders,
+    stripeHelper,
+  };
+}
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -72,6 +72,11 @@ type BillingAddressOptions = {
   state: string;
 };
 
+export type TimeSpanInS = {
+  startTimeS: number;
+  endTimeS: number;
+};
+
 /**
  * Determine for two product metadata object's whether the new one
  * is a valid upgrade for the old one.
@@ -972,6 +977,33 @@ export class StripeHelper {
     }
 
     return subscriptions;
+  }
+
+  /**
+   * Find all active subscriptions for the given `planId`.
+   *
+   * It is expected that the `customer` is expanded.
+   */
+  async *findActiveSubscriptionsByPlanId(
+    planId: string,
+    timeSpan: TimeSpanInS,
+    limit: number = 50
+  ) {
+    const params: Stripe.SubscriptionListParams = {
+      price: planId,
+      current_period_end: {
+        gte: timeSpan.startTimeS,
+        lt: timeSpan.endTimeS,
+      },
+      limit,
+      expand: ['data.customer'],
+    };
+    for await (const subscription of this.stripe.subscriptions.list(params)) {
+      if (!ACTIVE_SUBSCRIPTION_STATUSES.includes(subscription.status)) {
+        continue;
+      }
+      yield subscription;
+    }
   }
 
   /**
--- a/packages/fxa-auth-server/lib/payments/subscription-reminders.ts
+++ b/packages/fxa-auth-server/lib/payments/subscription-reminders.ts
@@ -1,0 +1,211 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { Logger } from 'mozlog';
+import Stripe from 'stripe';
+
+import { ConfigType } from '../../config';
+import { reportSentryError } from '../sentry';
+import { SentEmailParams, Plan } from 'fxa-shared/subscriptions/types';
+import { StripeHelper, TimeSpanInS } from './stripe';
+import { SentEmail } from 'fxa-shared/db/models/auth';
+
+export const EMAIL_TYPE: string = 'subscriptionRenewalReminder';
+const DAYS_IN_A_WEEK: number = 7;
+const DAYS_IN_A_MONTH: number = 30;
+const DAYS_IN_A_YEAR: number = 365;
+export const MS_IN_A_DAY: number = 24 * 60 * 60 * 1000;
+
+interface EligiblePlansByInterval {
+  day(plan: Plan, planLength: number): Boolean;
+  week(plan: Plan, planLength: number): Boolean;
+  month(plan: Plan, planLength: number): Boolean;
+  year(plan: Plan, planLength: number): Boolean;
+}
+
+const eligiblePlansByInterval: EligiblePlansByInterval = {
+  day: (plan, planLength) => plan.interval_count >= planLength,
+  week: (plan, planLength) =>
+    plan.interval_count >= planLength / DAYS_IN_A_WEEK,
+  month: (plan, planLength) =>
+    plan.interval_count >= planLength / DAYS_IN_A_MONTH,
+  year: (plan, planLength) =>
+    plan.interval_count >= planLength / DAYS_IN_A_YEAR,
+};
+
+export class SubscriptionReminders {
+  private db: any;
+  private mailer: any;
+  private stripeHelper: StripeHelper;
+
+  constructor(
+    private log: Logger,
+    config: ConfigType,
+    private planLength: number,
+    private reminderLength: number,
+    db: any,
+    mailer: any,
+    stripeHelper: StripeHelper
+  ) {
+    this.db = db;
+    this.mailer = mailer;
+    this.planLength = planLength;
+    this.reminderLength = reminderLength;
+    this.stripeHelper = stripeHelper;
+  }
+
+  /**
+   * For all possible plan.intervals, determine if the plan is sufficiently
+   * long based on planLength.
+   */
+  private isEligiblePlan(plan: Plan): Boolean {
+    const { interval } = plan;
+    if (eligiblePlansByInterval[interval]) {
+      return eligiblePlansByInterval[interval](plan, this.planLength);
+    }
+    return false;
+  }
+
+  private async getEligiblePlans(): Promise<Plan[]> {
+    const allPlans = await this.stripeHelper.allPlans();
+    return allPlans.filter((plan) => this.isEligiblePlan(plan));
+  }
+
+  /**
+   * Returns a window of time in seconds { startTimeS, endTimeS }
+   * that is exactly one day, reminderLength days from now in UTC.
+   */
+  private getStartAndEndTimes(reminderLengthMs: number): TimeSpanInS {
+    const reminderDay = new Date(Date.now() + reminderLengthMs);
+    // Get hour 0, minute 0, second 0 for today's date
+    const startingTimestamp = new Date(
+      Date.UTC(
+        reminderDay.getUTCFullYear(),
+        reminderDay.getUTCMonth(),
+        reminderDay.getUTCDate(),
+        0,
+        0,
+        0
+      )
+    );
+    // Get hour 0, minute, 0, second 0 for one day from today's date
+    const endingTimestamp = new Date(startingTimestamp.getTime() + MS_IN_A_DAY);
+
+    const startTimeS = Math.floor(startingTimestamp.getTime() / 1000);
+    const endTimeS = Math.floor(endingTimestamp.getTime() / 1000);
+    return {
+      startTimeS,
+      endTimeS,
+    };
+  }
+
+  private async alreadySentEmail(
+    uid: string,
+    currentPeriodStartMs: number,
+    emailParams: SentEmailParams
+  ) {
+    const emailRecord = await SentEmail.findLatestSentEmailByType(
+      uid,
+      EMAIL_TYPE,
+      emailParams
+    );
+    // This could be the first email for a given subscription or a subsequent one.
+    return !!emailRecord && emailRecord.sentAt > currentPeriodStartMs;
+  }
+
+  private async updateSentEmail(uid: string, emailParams: SentEmailParams) {
+    await SentEmail.createSentEmail(uid, EMAIL_TYPE, emailParams);
+  }
+
+  /**
+   * Send out a renewal reminder email if we haven't already sent one.
+   */
+  async sendSubscriptionRenewalReminderEmail(
+    subscription: Stripe.Subscription
+  ): Promise<Boolean> {
+    const customer: Stripe.Customer | Stripe.DeletedCustomer | string =
+      subscription.customer;
+    if (typeof customer === 'string' || customer?.deleted) {
+      return false;
+    }
+    const uid = customer.metadata.userid;
+    if (!uid) {
+      return false;
+    }
+    const emailParams = { subscriptionId: subscription.id };
+    if (
+      await this.alreadySentEmail(
+        uid,
+        Math.floor(subscription.current_period_start * 1000),
+        emailParams
+      )
+    ) {
+      return false;
+    }
+    try {
+      const account = await this.db.account(uid);
+      this.log.info('sendSubscriptionRenewalReminderEmail', {
+        message: 'Sending a renewal reminder email.',
+        subscriptionId: subscription.id,
+        currentPeriodStart: subscription.current_period_start,
+        currentPeriodEnd: subscription.current_period_end,
+        currentDateMs: Date.now(),
+      });
+      await this.mailer.sendSubscriptionRenewalReminderEmail(
+        account.emails,
+        account,
+        {
+          acceptLanguage: account.locale,
+        }
+      );
+      await this.updateSentEmail(uid, emailParams);
+      return true;
+    } catch (err) {
+      this.log.error('sendSubscriptionRenewalReminderEmail', {
+        err,
+        subscriptionId: subscription.id,
+      });
+      reportSentryError(err);
+      return false;
+    }
+  }
+
+  /**
+   * Sends a reminder email for all active subscriptions for all plans
+   * as long or longer than `planLength`:
+   *   1. Get a list of all plans of sufficient `planLength`
+   *   2. For each plan get active subscriptions with `current_period_end`
+   *      dates `reminderLength` away from now.
+   *   3. Send a reminder email if one hasn't already been sent.
+   */
+  public async sendReminders() {
+    let success = true;
+
+    // 1
+    const plans = await this.getEligiblePlans();
+
+    // 2
+    const timePeriod = this.getStartAndEndTimes(
+      this.reminderLength * MS_IN_A_DAY
+    );
+    for (const { plan_id } of plans) {
+      // 3
+      for await (const subscription of this.stripeHelper.findActiveSubscriptionsByPlanId(
+        plan_id,
+        timePeriod
+      )) {
+        try {
+          await this.sendSubscriptionRenewalReminderEmail(subscription);
+        } catch (err) {
+          this.log.error('sendSubscriptionRenewalReminderEmail', {
+            err,
+            subscriptionId: subscription.id,
+          });
+          reportSentryError(err);
+          success = false;
+        }
+      }
+    }
+    return success;
+  }
+}
--- a/packages/fxa-auth-server/scripts/paypal-processor.ts
+++ b/packages/fxa-auth-server/scripts/paypal-processor.ts
@@ -2,20 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import program from 'commander';
-import { setupAuthDatabase } from 'fxa-shared/db';
-import { StatsD } from 'hot-shots';
-import Container from 'typedi';
 
-import error from '../lib/error';
-import { CurrencyHelper } from '../lib/payments/currencies';
+import Container from 'typedi';
 import { PayPalHelper } from '../lib/payments/paypal';
 import { PayPalClient } from '../lib/payments/paypal-client';
+
+import {
+  AdditionalSetupFn,
+  initShared,
+} from '../lib/payments/processing-tasks-shared';
 import { PaypalProcessor } from '../lib/payments/paypal-processor';
-import { StripeHelper } from '../lib/payments/stripe';
-import { configureSentry } from '../lib/sentry';
 
 const pckg = require('../package.json');
 const config = require('../config').getProperties();
+
+const additionalSetup: AdditionalSetupFn = ({ log }) => {
+  const paypalClient = new PayPalClient(
+    config.subscriptions.paypalNvpSigCredentials
+  );
+  Container.set(PayPalClient, paypalClient);
+  const paypalHelper = new PayPalHelper({ log });
+  Container.set(PayPalHelper, paypalHelper);
+};
 
 export async function init() {
   // Load program options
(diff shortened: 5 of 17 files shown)


packages/fxa-auth-server/test/local/payments/subscription-reminders.js
```diff
@@ -0,0 +1,389 @@
...
+        {
+          err: throwErr,
+          subscriptionId: subscription.id,
+        }
+      );
+      sinon.assert.notCalled(reminder.updateSentEmail);
+    });
+  });
+
+  describe('sendReminders', () => {
```
biancadanforth, line 326: Please give extra scrutiny to this `describe` block.

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
```diff
@@ -0,0 +1,389 @@
...
+        yield longSubscription2;
+      }
+      mockStripeHelper.findActiveSubscriptionsByPlanId = sandbox.fake(
+        (planId) => {
+          if (planId === longPlan1.plan_id) {
+            return genSubscriptionForPlan1();
+          }
+          return genSubscriptionForPlan2();
+        }
+      );
```
biancadanforth, line 346: There's probably a better way to stub out `mockStripeHelper.findActiveSubscriptionsByPlanId` that asserts which call maps to which arguments. I couldn’t get Sinon’s `stub` API to work.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -974,6 +979,33 @@ export class StripeHelper {
...
 
+  /**
+   * Find all active subscriptions for the given `planId`.
+   *
+   * It is expected that the `customer` is expanded.
+   */
+  async *findActiveSubscriptionsByPlanId(
+    planId: string,
+    timeSpan: TimeSpanInS,
+    limit: number = 50
```
biancadanforth, line 990: I set the `limit` for `findActiveSubscriptionsByPlanId` to `50` for these reasons. I'm not sure if I should be concerned about the size of the returned `data` with an expanded `customer`.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+    const emailParams = { subscriptionId: subscription.id };
+    if (
+      await this.alreadySentEmail(
+        uid,
+        Math.floor(subscription.current_period_start * 1000),
+        emailParams
+      )
+    ) {
+      return false;
```
biancadanforth, line 147: Should we throw here? I am envisioning we may run this script 1-2x/day, which means we could see some dupe subscriptions that have already been processed, and that wouldn’t be an error. So I’m just returning.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+      EMAIL_TYPE,
+      emailParams
+    );
+    // This could be the first email for a given subscription or a subsequent one.
+    return !!emailRecord && emailRecord.sentAt > currentPeriodStartMs;
+  }
+
+  private async updateSentEmail(uid: string, emailParams: SentEmailParams) {
+    await SentEmail.createSentEmail(uid, EMAIL_TYPE, emailParams);
+  }
```
biancadanforth, line 122: We may want to import so it can be used elsewhere. Where would be the best place for this? I don’t see an obvious place in the current fxa-auth-server file structure (e.g. in ./lib).

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { Logger } from 'mozlog';
+import Stripe from 'stripe';
+import { Container } from 'typedi';
```
biancadanforth, line 6: I’m not using `AppError` anywhere; in general unlike the PayPal processor script -- should I be? (Any thoughts on additional errors I should throw or log would be great.)

biancadanforth (commented):
I added some specific questions inline.

After this merges, I will add a note in the deploy doc for running this script. What should the params into `subscription-reminders` CLI be for Stage? Default values are values for Prod. 
Proposed:
`-p 7`
Reasoning: Daily subs are used for a lot of testing, and we don't want this email to detract from testing other things/emails. Though arguably, this email doesn't change the state of any subscriptions.
`-r 6`
Reasoning: This would mean QA or we would wait 1 day for this email to show up automatically without any additional action taken.

packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
```diff
@@ -0,0 +1,87 @@
...
+import error from '../error';
+import { CurrencyHelper } from '../payments/currencies';
+import { PayPalHelper } from './paypal';
+import { PayPalClient } from './paypal-client';
+import { StripeHelper } from './stripe';
+import { configureSentry } from '../sentry';
+
+const config = require('../../config').getProperties();
+
+export default async function initShared(isPaypalProcessor = true) {
```
chenba, line 17: I'm not a fan of this type of toggle params.  Since it's a boolean, when we have another script that needs additional setup, do we 1) add another boolean parameter, 2) change it to a sum type, or 3) change it to an enum (or something else)?

I understand the need for it though.  What do you think of accepting a function that allows further setup needed by the calling script?  That way this shared lib need not handle whatever's wanted by the calling scripts.

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: You shouldn't be concerned. 🙂

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+import { ConfigType } from '../../config';
+import { reportSentryError } from '../sentry';
+import { SentEmailParams, Plan } from 'fxa-shared/subscriptions/types';
+import { StripeHelper, TimeSpanInS } from './stripe';
+import { SentEmail } from 'fxa-shared/db/models/auth';
+
+import {
+  DEFAULT_PLAN_LENGTH,
+  DEFAULT_REMINDER_LENGTH,
+} from '../../scripts/subscription-reminders';
```
chenba, line 17: I think it's safe to leave out the default values and require the plan length and reminder length passed into the constructor.  TypeScript would complain if they are missing.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+    private log: Logger,
+    config: ConfigType,
+    private planLength: number = DEFAULT_PLAN_LENGTH,
+    private reminderLength: number = DEFAULT_REMINDER_LENGTH,
+    db: any,
+    mailer: any
+  ) {
+    this.db = db;
+    this.mailer = mailer;
+    this.stripeHelper = Container.get(StripeHelper);
```
chenba, line 57: Even when there are default values for the plan length and reminder length, I think you'd still need assign them to the object's properties.

chenba: Why not accept the StripeHelper through a parameter?  That why you won't have a mixture of argument values and DI container values.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  /**
+   * For all possible plan.intervals, determine if the plan is sufficiently
+   * long based on planLength.
+   */
+  private isEligiblePlan(plan: Plan): boolean {
+    for (let [interval, checkFn] of Object.entries(eligiblePlansByInterval)) {
+      if (plan.interval === interval && checkFn(plan, this.planLength)) {
+        return true;
+      }
```
chenba, line 68: I don't think we need to loop here.  If  `eligiblePlansByInterval[plan.interval]` then call `eligiblePlansByInterval[plan.interval]`?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+    return false;
+  }
+
+  private async getEligiblePlans(): Promise<Plan[]> {
+    const allPlans = await this.stripeHelper.allPlans();
+    return allPlans.filter((plan) => this.isEligiblePlan(plan));
+  }
+
+  /**
+   * Returns a window of time in seconds [startTimeS, endTimeS)
```
chenba, line 79: `[` not matching `)`, but did you mean `{}` to match the return type?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
chenba: What would be the interface for the exported `updateSentEmail`?  If it's the same, then it'd only work with the 'subscriptionRenewalReminder' email type.  In that case, I hope people are using this class to begin with.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+   * Send out a renewal reminder email if we haven't already sent one.
+   */
+  async sendSubscriptionRenewalReminderEmail(
+    subscription: Stripe.Subscription
+  ): Promise<Boolean> {
+    const { customer } = subscription;
+    // @ts-ignore
+    const { email } = customer;
+    // @ts-ignore
+    const uid = customer.metadata.userid;
```
chenba, line 134: Nit: assert `customer` to be `Stripe.Customer` instead of using ts-ignore?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
chenba: Sound reasoning. I don't think we need to throw if we already sent an email.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+      return false;
+    }
+    try {
+      const account = await this.db.account(uid);
+      this.log.info('sendSubscriptionRenewalReminderEmail', {
+        message: 'Sending a renewal reminder email.',
+        subscriptionId: subscription.id,
+        currentPeriodStart: subscription.current_period_start,
+        currentPeriodEnd: subscription.current_period_end,
+        currentDateS: Math.floor(Date.now() / 1000),
```
chenba, line 156: Can this be in ms since we are not dealing with Stripe's API here?  Higher resolution is better?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+    let success = true;
+
+    // 1
+    const plans = await this.getEligiblePlans();
+
+    // 2
+    const timePeriod = this.getStartAndEndTimes(
+      this.reminderLength * MS_IN_A_DAY
+    );
+    for (let { plan_id } of plans) {
```
chenba, line 195: `const` instead of `let`?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,216 @@
...
+        try {
+          await this.sendSubscriptionRenewalReminderEmail(subscription);
+        } catch (err) {
+          this.log.error('sendSubscriptionRenewalReminderEmail', {
+            err,
+            subscriptionId: subscription.id,
+          });
+          reportSentryError(err);
+          success = false;
+          continue;
```
chenba, line 210: The `continue` seems unnecessary.  (But also doesn't hurt anything.)

packages/fxa-auth-server/scripts/subscription-reminders.ts
```diff
@@ -0,0 +1,53 @@
...
+export const DEFAULT_PLAN_LENGTH = 180;
+export const DEFAULT_REMINDER_LENGTH = 14;
+
+async function init() {
+  program
+    .version(pckg.version)
+    .option(
+      '-p, --plan-length [days]',
+      'Plan length in days beyond which a reminder email before the next recurring charge should be sent. Defaults to 180.',
+      `${DEFAULT_PLAN_LENGTH}`
```
chenba, line 22: nit: doesn't need to be string.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
chenba: hmmm I don't see the PayPal processor script using `AppError` either?  Anyways, the PayPal processor makes a lot of calls to both the PayPal NVP API and Stripe API, so a lot more can go wrong.

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
```diff
@@ -0,0 +1,389 @@
...
+    it('sets log, planLength, reminderLength and stripe helper', () => {
+      assert.strictEqual(reminder.log, mockLog);
+      assert.equal(reminder.planLength, planLength);
+      assert.equal(reminder.reminderLength, reminderLength);
+      assert.strictEqual(reminder.stripeHelper, mockStripeHelper);
+    });
+    it('uses defaults for planLength and reminderLength if no values are passed for them', () => {
+      const reminder = new SubscriptionReminders(mockLog, mockConfig);
+      assert.equal(reminder.planLength, DEFAULT_PLAN_LENGTH);
+      assert.equal(reminder.reminderLength, DEFAULT_REMINDER_LENGTH);
```
chenba, line 85: huh ok, my comment about the default arguments not auto set into the object properties was totally wrong.  I've never used that syntax sugar before.

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
```diff
@@ -0,0 +1,389 @@
...
+      sinon.assert.calledOnceWithExactly(
+        mockLog.error,
+        'sendSubscriptionRenewalReminderEmail',
+        {
+          err: throwErr,
+          subscriptionId: longSubscription1.id,
+        }
+      );
+      stub.firstCall.calledWithExactly(longSubscription1);
+      stub.secondCall.calledWithExactly(longSubscription2);
```
chenba, line 386: extra nit: assert `sendSubscriptionRenewalReminderEmail` was called exactly twice.

packages/fxa-shared/db/models/auth/sent-email.ts
```diff
@@ -80,7 +80,7 @@ export class SentEmail extends AuthBaseModel {
     return SentEmail.query().findOne({ id: result[0] });
   }
 
-  static async findLatestSentEmaiByType(
+  static async findLatestSentEmailByType(
```
chenba, line 83: LOL

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
chenba: I don't think the mapping matters, or at least I couldn't spot where that would make a difference in the tests.  You could use `stub.onFirstCall` and `stub.onSecondCall`.

chenba (approved):
@biancadanforth Thanks for the big patch.  It looks good overall.  Got a few comments, but they are nits.

packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
```diff
@@ -0,0 +1,12 @@
+--
```
chenba, line 1: uh oh, PR #8553 is also adding a migration so there will be a conflict.

packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
bbangert: We should try and avoid default exports unless we need to interact with legacy code that assumes every module exports a function.

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
biancadanforth: Thanks; yeah I found a more readable, succinct way to achieve this just now:
```diff
diff --git a/packages/fxa-auth-server/test/local/payments/subscription-reminders.js b/packages/fxa-auth-server/test/local/payments/subscription-reminders.js
index e0d3533ba..4c0c79dc5 100644
--- a/packages/fxa-auth-server/test/local/payments/subscription-reminders.js
+++ b/packages/fxa-auth-server/test/local/payments/subscription-reminders.js
@@ -336,14 +336,12 @@ describe('SubscriptionReminders', () => {
       async function* genSubscriptionForPlan2() {
         yield longSubscription2;
       }
-      mockStripeHelper.findActiveSubscriptionsByPlanId = sandbox.fake(
-        (planId) => {
-          if (planId === longPlan1.plan_id) {
-            return genSubscriptionForPlan1();
-          }
-          return genSubscriptionForPlan2();
-        }
+      const stub = sandbox.stub(
+        mockStripeHelper,
+        'findActiveSubscriptionsByPlanId'
       );
+      stub.onFirstCall().callsFake(genSubscriptionForPlan1);
+      stub.onSecondCall().callsFake(genSubscriptionForPlan2);
     });
     it('returns true if it can process all eligible subscriptions', async () => {
       reminder.sendSubscriptionRenewalReminderEmail = sandbox.fake.resolves({});
(END)
```

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: True; I guess I was thinking I'd change the params to accept an `emailType`. Probably not worth worrying about now as we only use this in one place so far.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: Huh; you're right. Where was I seeing `AppError`? O_O Very true that the PayPal processor is making a lot more API calls to both Stripe and PayPal.

packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
biancadanforth: > What do you think of accepting a function that allows further setup needed by the calling script?

Sure that sounds good. I modified this with the diff below:
```diff
diff --git a/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts b/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
index 3fdd02624..e0935619f 100644
--- a/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
+++ b/packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
@@ -7,14 +7,12 @@ import Container from 'typedi';
 
 import error from '../error';
 import { CurrencyHelper } from '../payments/currencies';
-import { PayPalHelper } from './paypal';
-import { PayPalClient } from './paypal-client';
 import { StripeHelper } from './stripe';
 import { configureSentry } from '../sentry';
 
 const config = require('../../config').getProperties();
 
-export default async function initShared(isPaypalProcessor = true) {
+export async function initShared(additionalSetupFn?: Function) {
   configureSentry(undefined, config);
   // Establish database connection and bind instance to Model using Knex
   setupAuthDatabase(config.database.mysql.auth);
@@ -70,13 +68,8 @@ export default async function initShared(isPaypalProcessor = true) {
   const stripeHelper = new StripeHelper(log, config, statsd);
   Container.set(StripeHelper, stripeHelper);
 
-  if (isPaypalProcessor) {
-    const paypalClient = new PayPalClient(
-      config.subscriptions.paypalNvpSigCredentials
-    );
-    Container.set(PayPalClient, paypalClient);
-    const paypalHelper = new PayPalHelper({ log });
-    Container.set(PayPalHelper, paypalHelper);
+  if (additionalSetupFn) {
+    additionalSetupFn(log);
   }
 
   return {
diff --git a/packages/fxa-auth-server/scripts/paypal-processor.ts b/packages/fxa-auth-server/scripts/paypal-processor.ts
index 4b4ece454..d3ec7d111 100644
--- a/packages/fxa-auth-server/scripts/paypal-processor.ts
+++ b/packages/fxa-auth-server/scripts/paypal-processor.ts
@@ -3,12 +3,26 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import program from 'commander';
 
-import initShared from '../lib/payments/processing-tasks-shared';
+import Container from 'typedi';
+import { Logger } from 'mozlog';
+import { PayPalHelper } from '../lib/payments/paypal';
+import { PayPalClient } from '../lib/payments/paypal-client';
+
+import { initShared } from '../lib/payments/processing-tasks-shared';
 import { PaypalProcessor } from '../lib/payments/paypal-processor';
 
 const pckg = require('../package.json');
 const config = require('../config').getProperties();
 
+function additionalSetup(log: Logger) {
+  const paypalClient = new PayPalClient(
+    config.subscriptions.paypalNvpSigCredentials
+  );
+  Container.set(PayPalClient, paypalClient);
+  const paypalHelper = new PayPalHelper({ log });
+  Container.set(PayPalHelper, paypalHelper);
+}
+
 export async function init() {
   // Load program options
   program
@@ -26,7 +40,7 @@ export async function init() {
     )
     .parse(process.argv);
 
-  const { log, database, senders } = await initShared();
+  const { log, database, senders } = await initShared(additionalSetup);
 
   const processor = new PaypalProcessor(
     log,
diff --git a/packages/fxa-auth-server/scripts/subscription-reminders.ts b/packages/fxa-auth-server/scripts/subscription-reminders.ts
index df042edeb..7156d8b2e 100644
--- a/packages/fxa-auth-server/scripts/subscription-reminders.ts
+++ b/packages/fxa-auth-server/scripts/subscription-reminders.ts
@@ -4,7 +4,7 @@
 
 import program from 'commander';
 
-import initShared from '../lib/payments/processing-tasks-shared';
+import { initShared } from '../lib/payments/processing-tasks-shared';
 import { SubscriptionReminders } from '../lib/payments/subscription-reminders';
 
 const pckg = require('../package.json');
@@ -29,7 +29,7 @@ async function init() {
     .parse(process.argv);
 
   const isPaypalProcessor = false;
-  const { log, database, senders } = await initShared(isPaypalProcessor);
+  const { log, database, senders } = await initShared();
 
   const subscriptionReminders = new SubscriptionReminders(
     log,

```

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: I had added this for a test to assert that if no values are passed into `new SubscriptionReminders(...)`, we'd still fall back to the defaults, but it makes sense to just fall back in the CLI and require the values in the constructor.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: I meant that the window of time was inclusive of the lower bound and exclusive of the upper bound. It's a convention I picked up in some math classes in college. Looking at this method, that information is actually relevant to the `findActiveSubscriptionsByPlanId`, not this one, so I'll change these to `{` and `}`.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: If I change this to `const customer: Stripe.Customer = subscription.customer;`, I get these TS errors:
```
Type 'string | Customer | DeletedCustomer' is not assignable to type 'Customer'.
Type 'string' is not assignable to type 'Customer'.ts(2322)
```

This worked, so I guess I will go with that to avoid type errors at runtime?:
```ts
const customer: Stripe.Customer | Stripe.DeletedCustomer | string = subscription.customer;
if (typeof customer === "string") {
  return false;
}
if (customer?.deleted) {
  return false;
}
// ...
```

packages/fxa-auth-server/test/local/payments/subscription-reminders.js
biancadanforth: Yeah it seemed to work fine. I made it explicit anyway.

packages/fxa-auth-server/lib/payments/processing-tasks-shared.ts
biancadanforth: Barry and I discussed this sync and resolved it.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
biancadanforth: Barry and I resolved this sync.

packages/fxa-auth-db-mysql/lib/db/schema/patch-115-116.sql
biancadanforth: Thanks for pointing this out. If it's not merged by the time I'm ready to merge, I'll check with Vijay about it.

packages/fxa-auth-server/scripts/paypal-processor.ts
```diff
@@ -2,21 +2,29 @@
...
+  initShared,
+} from '../lib/payments/processing-tasks-shared';
 import { PaypalProcessor } from '../lib/payments/paypal-processor';
-import { StripeHelper } from '../lib/payments/stripe';
-import { configureSentry } from '../lib/sentry';
 
 const pckg = require('../package.json');
 const config = require('../config').getProperties();
 
+const additionalSetup: AdditionalSetupFn = ({ log }) => {
```
bbangert, line 19: It seems odd that we add code to end up with the same setup in the same locations, but pass around where its done. Why not just move this under the initialization?

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -104,3 +104,8 @@ export const PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE = 'funding_source';
 export type PaypalPaymentError =
   | typeof PAYPAL_PAYMENT_ERROR_MISSING_AGREEMENT
   | typeof PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE;
+
+export type SentEmailParams = {
+  subscriptionId: string;
+  planId?: string;
```
bbangert, line 110: Do we want an optional `planId`?

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,211 @@
...
+      uid,
+      EMAIL_TYPE,
+      emailParams
+    );
+    // This could be the first email for a given subscription or a subsequent one.
+    return !!emailRecord && emailRecord.sentAt > currentPeriodStartMs;
+  }
+
+  private async updateSentEmail(uid: string, emailParams: SentEmailParams) {
+    await SentEmail.createSentEmail(uid, EMAIL_TYPE, emailParams);
```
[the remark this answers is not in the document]
biancadanforth: Yeah, looking at the `SentEmail` ORM ([createSentEmail](https://github.com/mozilla/fxa/blob/b22f2e8b38f0ce0baf4ccecdd2ea3a96724f2013/packages/fxa-shared/db/models/auth/sent-email.ts#L52)), this method does an `INSERT` and then returns the newly added entry to the table.

Hmm... yeah I didn't think about the fact that we don't use the result of `updateSentEmail` for anything, so we don't need to await the Promise (except in tests). Thanks for pointing that out.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Send out a renewal reminder email if we haven't already sent one.
+   */
+  async sendSubscriptionRenewalReminderEmail(
+    subscription: Stripe.Subscription
+  ): Promise<Boolean> {
+    const customer: Stripe.Customer | Stripe.DeletedCustomer | string =
```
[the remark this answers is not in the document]
biancadanforth: I was getting a TS error:
```
Property 'metadata' does not exist on type 'string | Customer | DeletedCustomer'.
  Property 'metadata' does not exist on type 'string'.ts(2339)
```
Though on second glance, yeah TS infers the type; I just need to keep this:
```js
if (typeof customer === 'string' || customer?.deleted) {
  return false;
}
```
I removed the type on `customer`.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    const startTimeS = Math.floor(startingTimestamp.getTime() / 1000);
+    const endTimeS = Math.floor(endingTimestamp.getTime() / 1000);
+    return {
+      startTimeS,
+      endTimeS,
+    };
+  }
+
+  private async alreadySentEmail(
```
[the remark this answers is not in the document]
biancadanforth: Well this is the first thing to make use of the `sentEmails` table. Would you like me to make this a utility function available outside the `SendReminders` class? Where might be a good place to put it?

packages/fxa-shared/subscriptions/types.ts
biancadanforth: Hm... I don't remember why I added that. I'm not using `planId` anywhere for this type, so I'll remove it.

PR #8586, opened 2021-05-17
feat(debug): add new vscode extension recommendations and launch config

dannycoates:
## Because

- I finally published pm2-node-debugger to the marketplace


vbudhram (approved):
@dannycoates 💯

Issue #7039, opened 2020-12-01
Device Manager - Last seen date is broken


data-sync-user:
We no longer display accurately the date a device was last seen. This seems to align with the desktop migration to OAuth.

【 highlighted #172b4d 】Here are 2 devices. One is Nightly and the other is Beta. You can see when each device is detecting the other as having been active. Ultimately, it seems to detect that they were active prior to the migration to OAuth. Nightly was active 6 months ago. Beta was active 4 months ago. From these screenshots, you can see that each device is in fact actually still active.
 
 **I guess the main question is, are we migrating this bug to our new settings? If so, let's add it to our list of things to resolve with the new settings.**【 end highlighted 】

!image-2020-12-01-10-45-56-104.png!

!image-2020-12-01-10-51-02-171.png!

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2764)
┆Attachments: <a href="https://jira.mozilla.com/secure/attachment/14619/image-2020-12-01-10-45-56-104.png">image-2020-12-01-10-45-56-104.png</a> | <a href="https://jira.mozilla.com/secure/attachment/14623/image-2020-12-01-10-51-02-171.png">image-2020-12-01-10-51-02-171.png</a> | <a href="https://jira.mozilla.com/secure/attachment/14621/Screen+Shot+2020-12-01+at+10.36.13+AM.png">Screen Shot 2020-12-01 at 10.36.13 AM.png</a>
┆Issue Number: FXA-2764

rfk:
Signing a certificate [bumps the last-accessed time on the sessionToken](https://github.com/mozilla/fxa/blob/c39f8790513657110a304eff108512a7f80b8349/packages/fxa-auth-server/lib/routes/sign.js#L80) but it looks like [granting an access token does not](https://github.com/mozilla/fxa/blob/c39f8790513657110a304eff108512a7f80b8349/packages/fxa-auth-server/lib/routes/oauth/index.js#L219).

data-sync-user:
➤ Alex Davis commented:

[~dcoates@mozilla.com] and [~wclouser@mozilla.com] is this something that could/should be part of the OAuth maintenance sprint? It overlaps well with Settings too.

data-sync-user:
➤ Wil Clouser commented:

The OAuth maintenance sprint is already overfull.  It probably fits the category though.

data-sync-user:
➤ Jared Hirsch commented:

[...]

PR #8650, opened 2021-05-18
fix(session): Update sessionToken data in redis when used by route

vbudhram:
## Because

- Previously, the only time the "last access" value on a session token was updated was on the `/certification/sign` route which leads to confusing last accessed time

## This pull request

- Calls the `db.touchSessionToken` in the `/account/attach_devices` route to update the last used time
- This route is a lower volume route and should allow us to test out the overall performance

Closes: #7039

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/db/index.js
+++ b/packages/fxa-auth-server/lib/db/index.js
@@ -581,7 +581,7 @@ module.exports = (config, log, Token, UnblockCode = null) => {
    * frequently-called routes.
    *
    * To do a cheaper write of transient metadata that only hits
-   * redis, use touchSessionToken isntead.
+   * redis, use touchSessionToken instead.
    */
   SAFE_URLS.updateSessionToken = new SafeUrl(
     '/sessionToken/:id/update',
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/server.js
```diff
@@ -267,6 +267,33 @@ async function create(log, error, config, routes, db, translator, statsd) {
...
+      sessionToken.setUserAgentInfo({
+        uaBrowser,
+        uaBrowserVersion,
+        uaOS,
+        uaOSVersion,
+        uaDeviceType,
+        uaFormFactor,
+        lastAccessTime: Date.now(),
+      });
+      db.touchSessionToken(sessionToken, request.app.geo);
```
dannycoates, line 294: It's not a mysql hit, but it does run a lua fn in redis. I like the idea of updating on any use of the token, but I'd like to understand the effect on redis before we do. Is there a route that is lowish volume that'd get us the desired effect for sync?

dannycoates (commented):
If we do this in the response hook we should remove the other calls to `touchSessionToken`

packages/fxa-auth-server/lib/server.js
vbudhram: > Is there a route that is lowish volume that'd get us the desired effect for sync?

Fair point, I think in hindsight it is probably better to start small with this change. I can confirm in the sync slack what this route might be.

Outside of syncing I think we can get pretty close to "last used" value by touching the session token when we hit the `/account/attached_client` route. IIUC that route is only used when we load up settings and display users devices and apps.

dannycoates (approved):
lgtm

PR #8652, opened 2021-05-19
feat(auth-server): add locale to zendesk tickets

### LINKED ISSUE #8408: [subplat] Have support form pass the customer's language to Zendesk
The support form (content-server) send its submission to auth-server, which then creates a Request ([https://developer.zendesk.com/rest_api/docs/support/requests#json-format)](https://developer.zendesk.com/rest_api/docs/support/requests#json-format)) on Zendesk via their API.  An additional piece of information we would like to capture is the customer's language preferences, from the accept-language header.  It would need to be included in the Zendesk Request as a custom field.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3427)
┆Issue Number: FXA-3427

bbangert:
Because:

* We want to retain the users locale in zendesk.
* The support routes should be in the subscription route section.

This commit:

* Copies the users locale into the zendesk user.
* Refactors and updates the zendesk client library handling.

Closes #8408

## Checklist


packages/fxa-auth-server/config/index.ts
```diff
@@ -1802,42 +1802,42 @@ const conf = convict({
     },
     productNameFieldId: {
       doc: 'Zendesk support ticket custom field for the product name',
-      default: '360022027772',
+      default: 360022027772,
       env: 'ZENDESK_PRODUCT_NAME_FIELD_ID',
-      format: String,
+      format: Number,
```
chenba, line 1807: Huh!  I can't remember why they were strings.  Maybe no reason, I hope.

packages/fxa-auth-server/lib/routes/subscriptions/support.ts
```diff
@@ -0,0 +1,152 @@
...
+            | null
+            | undefined
+            | Record<string, any>;
+          if (!userFields?.user_id || !showRequest.locale) {
+            operation = 'updateUser';
+            await pRetry(async () => {
+              return await zendeskClient.users.update(zenUid, {
+                user: {
+                  user_fields: { user_id: uid },
+                  locale: request.app.locale,
```
chenba, line 135: Maybe you already know and decided that it's not a problem.  Just in case though: `request.app.locale` is accept-language filtered through our list of supported languages for the frontends and emails.  'de' and 'fr' are on that list, so it's not a problem right now.

The idea was to add the language tag as a custom field.  Maybe this is better for routing the support ticket?  I don't know enough abut Zendesk.  But we should check with JR.  Where does this locale end up in Zendesk?

bbangert: In a locale field to mark the users locale. I did talk to @dannycoates to verify that this value is correct.

chenba: Note: checked and resolved with JR in Slack

chenba (approved):
🇫🇷 🇩🇪

PR #7652, opened 2021-02-26
auth-server direct db reads

dannycoates:
This begins the process of eliminating fxa-auth-db-mysql. The first goal is to make all db reads directly from auth-server to the mysql server. We'll expand the use of `fxa-shared/db` to replace the http requests to fxa-auth-db-mysql in `auth-server/lib/db`.

## Benefits

- It will be significantly less code to maintain
- Less complex and easier to reason through and debug
- Future changes to the data model will be easier to implement
- Improved request latency from the more direct db access
- Expands our Typescript coverage

## Structure

- `fxa-shared/db/models/auth` contains the new db models
- `fxa-auth-server/lib/db/index` initializes the db connection and wraps the model api to the existing db api
  - eventually we can use the model api directly in other parts of auth-server

PR #8677, opened 2021-05-20
fix(auth): fixed email_notifications startup

dannycoates:
## Because

- The implementation of DB.connect changed in https://github.com/mozilla/fxa/pull/7652


jbuck (approved):
I manually patched this file in prod and it's processing bounce notifications correctly now

PR #8679, opened 2021-05-21
chore(auth): clean up subscription-reminders script

biancadanforth:
Because:

* There were some suboptimal aspects to the subscription-reminders script.

This commit:

* Uses a datetime library, luxon, to handle determining eligible plans and computing the time period for pulling subscriptions from Stripe.
* Renames processing-tasks-shared.ts to processing-tasks-setup.ts in lib/payments and its exported function from initShared to setupProcesingTaskObjects.
  * Removes the additionalSetup optional parameter from initShared.
* Reports a sentry error if there is no userid for a customer with an active subscription.
* Relies on existing library types instead of the custom TimeSpanInS type.
* Makes the call to updateSentEmail sync to avoid unnecessary spins of the event loop.

This is a follow-up to #8224.

## Checklist


packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -158,7 +145,7 @@ export class SubscriptionReminders {
           acceptLanguage: account.locale,
         }
       );
-      await this.updateSentEmail(uid, emailParams);
+      this.updateSentEmail(uid, emailParams);
```
bbangert, line 148: Still needs the await here, otherwise the promise won't be waited on and context lost if it fails.

packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -72,31 +61,21 @@ export class SubscriptionReminders {
...
+  private getStartAndEndTimes(reminderLengthMs: number): Interval {
+    const reminderDate = DateTime.fromMillis(Date.now() + reminderLengthMs, {
+      zone: 'utc',
+    });
+    const startDate = DateTime.fromObject({
+      year: reminderDate.year,
+      month: reminderDate.month,
+      day: reminderDate.day,
+      zone: 'utc',
+    });
```
bbangert, line 76: Could be replaced by:
```typescript
const reminderDate = DateTime.utc().plus(reminderLength).set({hour:0, minute: 0, second: 0 })
```
if `reminderLength` was passed in as a Duration.

bbangert (commented):
Looking good! Few little things and I think you nailed it!

bbangert (approved):
Awesome job on this, the cleanup looks great.

Issue #8571, opened 2021-05-13
[subplat][content-server] Allow Redirects to Payments Server's Product Route Without an Access Token


data-sync-user:
The content server handles redirects to the Payments server since, historically, the user needs to be authenticated in order to access the Payments frontend.

For the product route on Payments, we now want to allow anyone to access it.  The content server should be updated that access token is only appended when the user is already authenticated.  This _only_ applies to the payments product route.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3488)
┆Issue Number: FXA-3488

data-sync-user:
➤ Barry Chen commented:

Note that this already has an open PR (written during train 207), but we are de-prioritizing the epic for now, so I'll set it back to To Do.  cc [~wclouser@mozilla.com] in case that's not the right thing to do when an issue with an open PR get de-prioritized.

PR #8695, opened 2021-05-24
feat(subscriptions): allow redirects to Payments without token

chenba:
Because:
 - we want to enable people to sign up and subscribe on Payments in a
   new checkout flow

This commit:
 - allow unauthenticated redirects to the product route on Payments from
   content-server
 - add a feature flag to toggle the behavior (default is off)

Closes: #8571

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/payment-server.js
+++ b/packages/fxa-content-server/app/scripts/lib/payment-server.js
@@ -7,6 +7,10 @@
 
 import Url from './url';
 
+const allowedUnauthenticatedRoutes = ['product'];
+const isAllowedUnauthenticatedRoute = (redirectPath) =>
+  allowedUnauthenticatedRoutes.some((route) => redirectPath.startsWith(route));
+
 const PaymentServer = {
   /**
    * Navigate to the payments server after flushing metrics.
@@ -16,7 +20,7 @@ const PaymentServer = {
    * @param {String} redirectPath
    * @param {Object} [queryParams={}]
    */
-  navigateToPaymentServer(
+  async navigateToPaymentServer(
     view,
     subscriptionsConfig,
     redirectPath,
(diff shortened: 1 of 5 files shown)


packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -687,6 +687,13 @@ const conf = (module.exports = convict({
       env: 'SUBSCRIPTIONS_MANAGEMENT_URL',
       format: String,
     },
+    allowUnauthenticatedRedirects: {
+      default: false,
+      doc:
+        'Whether to allow any redirects to Payments for an unauthenticated user',
+      env: 'SUBSCRIPTIONS_UNAUTHED_REDIRECTS',
+      formlat: Boolean,
+    },
```
chenba, line 696: Added a feature flag because Payments currently requires authentication and a user who isn't signed in will be redirected to settings.  We don't want people end up in settings after sign up/in.

packages/fxa-content-server/app/scripts/lib/payment-server.js
```diff
@@ -35,31 +39,51 @@ const PaymentServer = {
...
+    const unauthenticatedRedirect = () => {
+      if (isAllowedUnauthenticatedRoute(redirectPath)) {
+        return view.navigateAway(url);
+      }
+
+      view.navigate('/');
+    };
+
+    if (account) {
+      return account
```
vbudhram, line 63: I think using async functions might allow us to read the code a little easier.

vbudhram (approved):
@chenba LGTM 👍🏽

packages/fxa-content-server/app/scripts/lib/payment-server.js
chenba: heh I fought my urge to use async/await here to be consistent with the rest of the content server frontend.  Yes, happy to make that change.  👍

Issue #8598, opened 2021-05-18
Allow access to auth-server's subscriptions/plans route without an access token


data-sync-user:
Since the new checkout flow allows unauthenticated users to subscribe to product plans, we need allow access to the GET plans route without an access token.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3501)
┆Issue Number: FXA-3501

data-sync-user:
➤ Barry Chen commented:

[~wclouser@mozilla.com] PR's already merged.  Should I add the Sprint back?

PR #8704, opened 2021-05-24
feat(subscriptions): allow plans list access without auth

chenba:
Because:
 - users who are not authenticated should be allow to access product
   plans

This commit:
 - stop requiring an access token on the plans endpoint

Closes: #8598


bbangert (approved):
👍

PR #8705, opened 2021-05-25
feat(emails): add email for subscription renewal reminder

### LINKED ISSUE #8223: [subplat][emails] Add Email Templates and Sender Function for Subscription Renewal Reminder

biancadanforth:
Because:
* There is a legal requirement to send a reminder email two weeks before a subscription for a plan longer than 180 days is renewed.

This commit:
* Adds a new subscriptionRenewalReminder email template
* Sends the new email for a qualifying subscription as determined by #8224

Closes #8223

## Checklist

## Screenshots (Optional)

Screen Shot 2021-05-24 at 10 03 57 PM

### DIFF
--- a/packages/fxa-auth-server/lib/payments/subscription-reminders.ts
+++ b/packages/fxa-auth-server/lib/payments/subscription-reminders.ts
@@ -94,7 +94,8 @@ export class SubscriptionReminders {
    * Send out a renewal reminder email if we haven't already sent one.
    */
   async sendSubscriptionRenewalReminderEmail(
-    subscription: Stripe.Subscription
+    subscription: Stripe.Subscription,
+    planId: string
   ): Promise<Boolean> {
     const { customer } = subscription;
     if (typeof customer === 'string' || customer?.deleted) {
@@ -132,11 +133,25 @@ export class SubscriptionReminders {
         currentPeriodEnd: subscription.current_period_end,
         currentDateMs: Date.now(),
       });
+      const { email } = account;
+      const formattedSubscription =
+        await this.stripeHelper.formatSubscriptionForEmail(subscription);
+      const { amount, currency, interval_count, interval } =
+        await this.stripeHelper.findPlanById(planId);
       await this.mailer.sendSubscriptionRenewalReminderEmail(
         account.emails,
         account,
         {
+          uid,
+          email,
           acceptLanguage: account.locale,
+          subscription: formattedSubscription,
+          reminderLength: this.reminderDuration.as('days'),
+          planIntervalCount: interval_count,
+          planInterval: interval,
+          // Using invoice prefix instead of plan to accomodate `yarn write-emails`.
+          invoiceTotalInCents: amount,
+          invoiceTotalCurrency: currency,
         }
       );
       await this.updateSentEmail(uid, emailParams);
@@ -177,7 +192,10 @@ export class SubscriptionReminders {
         }
       )) {
         try {
-          await this.sendSubscriptionRenewalReminderEmail(subscription);
+          await this.sendSubscriptionRenewalReminderEmail(
+            subscription,
+            plan_id
+          );
         } catch (err) {
           this.log.error('sendSubscriptionRenewalReminderEmail', {
             err,
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -2117,8 +2117,12 @@ module.exports = function (log, config) {
 
     if (subscriptions.length === 1) {
       productName = subscriptions[0].productName;
-      subject = translator.gettext(
-        'Payment information update required for %(productName)s'
+      const translatorParams = { productName };
+      subject = translator.format(
+        translator.gettext(
+          'Payment information update required for %(productName)s'
+        ),
+        translatorParams
       );
       links = this._generateLinks(
         null,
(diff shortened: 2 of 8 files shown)


packages/fxa-auth-server/lib/payments/subscription-reminders.ts
```diff
@@ -132,11 +132,32 @@ export class SubscriptionReminders {
...
+      const formattedSubscription = await this.stripeHelper.formatSubscriptionForEmail(
+        subscription
+      );
+      const {
+        amount,
+        currency,
+        interval_count,
+        interval,
+      } = await this.stripeHelper.findPlanById(
+        subscription.items.data[0].price.id
```
biancadanforth, line 145: This seems like a suboptimal way to get the plan ID. Given that I have the `subscription` object here with an expanded `customer`, is there a more direct way?

biancadanforth: Answering my own question here after a good night's sleep: In `sendReminders`, I have access to the `plan` for the `subscription` inside the `for` loop. So I could pass in `plan` or `plan_id` to `sendSubscriptionRenewalReminderEmail` as a second argument...

chenba:
Nothing to do with the PR and it is not a bug, but mixing € and a US formatted number ('10,000.00') actually looks kinda weird. 😛

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -2408,7 +2412,64 @@ module.exports = function (log, config) {
...
+    if (!enabled) {
+      return;
+    }
+
+    const headers = {};
+    const translator = this.translator(message.acceptLanguage);
+
+    const template = 'subscriptionRenewalReminder';
+
+    // We only have one subscription here, but `yarn write-emails` passes in a list
```
chenba, line 2432: I think it'd be better to update the write-emails-to-disks.js script to pass in a single `subscription` (as well as `subscriptions`) than it is to have the code accommodate the script.

chenba (changes requested):
LGTM overall.  There are a couple places where the code is expecting a list of subscriptions even though this email is for a single subscription.  I think that can be fixed by updating write-emails-to-disk.js script.

biancadanforth:
> Nothing to do with the PR and it is not a bug, but mixing € and a US formatted number ('10,000.00') actually looks kinda weird. 😛

Well isn't this a bug, since this is the output of a function?:
```js
_getLocalizedCurrencyString(
  message.invoiceTotalInCents,
  message.invoiceTotalCurrency,
  message.acceptLanguage
),
```

I.e. just changing `MESSAGE_FORMATTED` in the test to `'€10.000,00'` is insufficient.

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -319,6 +322,28 @@ const TESTS = [
...
+      { test: 'notInclude', expected: 'utm_source=email' },
+    ]],
+    ['text', [
+      { test: 'include', expected: `${MESSAGE.productName} automatic renewal notice` },
+      { test: 'include', expected: `Your current subscription is set to automatically renew in ${MESSAGE.reminderLength} days. At that time, Mozilla will renew your ${MESSAGE.planIntervalCount} ${MESSAGE.planInterval} subscription and a charge of ${MESSAGE_FORMATTED.invoiceTotal} will be applied to the payment method on your account.` },
+      { test: 'notInclude', expected: 'utm_source=email' },
+    ]]
+  ]),
+  {updateTemplateValues: x => (
+    {...x, subscription: {planId: MESSAGE.planId, productId: MESSAGE.productId, ...x.subscriptions[0]}})}],
```
chenba, line 346: I don't think this optional arg is necessary.

chenba (approved):
👍 but I don't think that optional arg to format the template values for the test is necessary.

packages/fxa-auth-server/test/local/senders/email.js
biancadanforth: It isn't if I add the `subscription` object to `MESSAGE` at the top of the file. I can do that instead.

chenba: Oh this is the first email that gets a `subscription` in the template values.  Sorry about that.

chenba: > 
> Edit: Assuming you are meaning I don't need to pass `updateTemplateValues` at all...

I did.

packages/fxa-auth-server/scripts/write-emails-to-disk.js
```diff
@@ -110,6 +110,8 @@ function sendMail(mailer, messageToSend) {
     productName: 'Firefox Fortress',
     planEmailIconURL: 'http://placekitten.com/512/512',
     planDownloadURL: 'http://getfirefox.com/',
+    planInterval: 'week',
```
flodolo, line 113: @peiying2 

Re: https://pontoon.mozilla.org/it/firefox-accounts/all-resources/?status=missing&string=224251

This looks hard-coded to me. Besides that, this structure ("$num $period subscription") won't work in any language but English.

PR #8719, opened 2021-05-25
fix(autofill): disable email autocomplete while we figure out why it broke

vbudhram:
Lets see if all these tests pass...


dannycoates (approved):

jonalmeida:
For reference: this [fixed a bug](https://github.com/mozilla-mobile/fenix/issues/19704) that I filed on behalf of others using Fenix.

PR #8771, opened 2021-06-01
Document some internal `fxa-*` claims inside JWT access tokens.

rfk:
Inherited from BrowserID, some internal Mozilla reliers need access
to special `fxa-*` prefixed claims. These were only documented in
the part of the API doc about BrowserID, so I've copied the text
over to the OAuth docs to make it easier to find, and added a few
notes about how we expect the claims to be used.

(This is the same change as https://github.com/mozilla/ecosystem-platform/pull/141
but in the docs in this repo; if we're not going to delete them from here then
it seems copying across for consistency...but maybe I should just delete this
and redirect to ecosystem-platform docs?)


vbudhram (commented):
@rfk If docs are the same, I say delete and redirect to eco system docs.

rfk:
> If docs are the same, I say delete and redirect to eco system docs.

done 👍🏻

dannycoates (approved):

PR #8781, opened 2021-06-02
refactor(db): move more auth-db functions into fxa-shared

dannycoates:
This is a continuation of #7652. It moves where the create and delete queries are executed from the auth-db-server to the auth-server via fxa-shared.

### DIFF
--- a/packages/fxa-auth-server/lib/db/index.js
+++ b/packages/fxa-auth-server/lib/db/index.js
@@ -36,10 +36,8 @@ module.exports = (config, log, Token, UnblockCode = null) => {
   } = Token;
   const MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE =
     config.tokenLifetimes.sessionTokenWithoutDevice;
-  const {
-    enabled: TOKEN_PRUNING_ENABLED,
-    maxAge: TOKEN_PRUNING_MAX_AGE,
-  } = config.tokenPruning;
+  const { enabled: TOKEN_PRUNING_ENABLED, maxAge: TOKEN_PRUNING_MAX_AGE } =
+    config.tokenPruning;
 
   const SAFE_URLS = {};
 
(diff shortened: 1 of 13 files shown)


packages/fxa-auth-server/lib/db/index.js
```diff
@@ -375,9 +312,8 @@ module.exports = (config, log, Token, UnblockCode = null) => {
 
     try {
       const [devices, redisSessionTokens = {}] = await Promise.all(promises);
-      const lastAccessTimeEnabled = features.isLastAccessTimeEnabledForUser(
-        uid
-      );
+      const lastAccessTimeEnabled =
```
vbudhram, line 315: As a side note, the `lastAccessTime` feature has been around for a sometime now it seems and I wonder if we can graduate it.

packages/fxa-auth-server/lib/db/index.js
```diff
@@ -1165,12 +1052,11 @@ module.exports = (config, log, Token, UnblockCode = null) => {
     }
   };
 
-  SAFE_URLS.deleteTotpToken = new SafeUrl('/totp/:uid', 'db.deleteTotpToken');
   DB.prototype.deleteTotpToken = async function (uid) {
     log.trace('DB.deleteTotpToken', { uid });
 
     try {
-      return await this.pool.del(SAFE_URLS.deleteTotpToken, { uid });
+      return await TotpToken.delete(uid);
```
vbudhram, line 1059: Don't need the await here.

packages/fxa-auth-server/lib/db/index.js
```diff
@@ -1295,17 +1169,9 @@ module.exports = (config, log, Token, UnblockCode = null) => {
...
-    'db.updateEcosystemAnonId'
-  );
   DB.prototype.updateEcosystemAnonId = async function (uid, ecosystemAnonId) {
     log.trace('DB.updateEcosystemAnonId', { uid, ecosystemAnonId });
-    return this.pool.put(
-      SAFE_URLS.updateEcosystemAnonId,
-      { uid },
-      { ecosystemAnonId }
-    );
+    return {};
```
vbudhram, line 1174: nit: Probably leave a comment here that we intend to remove this feature.

vbudhram (approved):
@dannycoates LGTM 🌮

packages/fxa-auth-server/lib/db/index.js
dannycoates: we do if we want to catch here

Issue #8570, opened 2021-05-13
Reduce the scope of the admin panel mysql user


data-sync-user:
Right now the admin panel mysql user has SELECT on more columns than it needs.  We should review what is needed and update those.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3487)
┆Issue Number: FXA-3487

jbuck:
Current permissions:

```
MySQL [fxa]> show grants for 'fxa_admin_202004';;
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Grants for fxa_admin_202004@%                                                                                                                                                                                                                        |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| GRANT USAGE ON *.* TO 'fxa_admin_202004'@'%'                                                                                                                                                                                                         |
| GRANT SELECT (sharedSecret, uid, epoch, createdAt, verified, enabled) ON `fxa`.`totp` TO 'fxa_admin_202004'@'%'                                                                                                                                      |
| GRANT SELECT (bounceSubType, bounceType, email, createdAt), DELETE ON `fxa`.`emailBounces` TO 'fxa_admin_202004'@'%'                                                                                                                                 |
| GRANT SELECT (email, createdAt, uid, emailVerified) ON `fxa`.`accounts` TO 'fxa_admin_202004'@'%'                                                                                                                                                    |
| GRANT SELECT (isPrimary, createdAt, uid, id, verifiedAt, normalizedEmail, email, isVerified) ON `fxa`.`emails` TO 'fxa_admin_202004'@'%'                                                                                                             |
| GRANT SELECT (verificationMethod, uaBrowser, uaOSVersion, uid, authAt, createdAt, mustVerify, uaOS, uaBrowserVersion, uaDeviceType, lastAccessTime, tokenData, verifiedAt, uaFormFactor, tokenId) ON `fxa`.`sessionTokens` TO 'fxa_admin_202004'@'%' |
| GRANT SELECT (recoveryData, uid, enabled, recoveryKeyIdHash, verifiedAt, createdAt) ON `fxa`.`recoveryKeys` TO 'fxa_admin_202004'@'%'                                                                                                                |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
7 rows in set (0.00 sec)
```

jbuck:
[...]

PR #8793, opened 2021-06-03
chore(admin): remove unnecessary columns from admin panel queries

dannycoates:
## Because

- We want to limit access to sensitive data

Closes: #8570

### DIFF
--- a/packages/fxa-admin-server/src/gql/account/account.resolver.ts
+++ b/packages/fxa-admin-server/src/gql/account/account.resolver.ts
@@ -31,33 +31,11 @@ const EMAIL_COLUMNS = [
   'uid',
 ];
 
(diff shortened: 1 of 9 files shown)


packages/fxa-admin-server/src/gql/account/account.resolver.ts
```diff
@@ -31,33 +31,11 @@ const EMAIL_COLUMNS = [
...
-  'uid',
-  'nameId',
-  'verified',
-  'ipAddrHmac',
-  'createdAt',
-  'tokenVerificationId',
-];
-const TOTP_COLUMNS = [
-  'uid',
-  'sharedSecret',
```
vbudhram, line 44: 👀

vbudhram (approved):
@dannycoates Good catch!

PR #8795, opened 2021-06-04
feat(payments): Use custom action button label if specified in Paymen…

### LINKED ISSUE #8713: After successful subscription flow, need the ability to customize the "continue to download" button message
Similar to the issue FxA 3435, the action after a successful purchase will return to a screen with "continue to download" button. It would be desirable for the application that initiate the purchase transaction to specify what message to be displayed with that button.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3589)
┆Issue Number: FXA-3589

biancadanforth:
…tConfirmation

Because:

* Unlike the VPN, some subscription products (e.g. web apps/services) may not have software for the subscriber to download.

This commit:

* Uses a custom action button label instead of 'Continue to download' or its localized equivalent if a new 'successActionButtonLabel' key/value is specified in Stripe product or plan metadata.

Closes #8713

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { Localized } from '@fluent/react';
+import React, { useContext } from 'react';
+import { Localized, useLocalization } from '@fluent/react';
 import * as Provider from '../../lib/PaymentProvider';
 import { getLocalizedCurrency, formatPlanPricing } from '../../lib/formats';
 import { Plan, Profile, Customer } from '../../store/types';
@@ -11,6 +11,8 @@ import PaymentLegalBlurb from '../PaymentLegalBlurb';
 import circledCheckbox from './images/circled-confirm.svg';
 
 import './index.scss';
+import { productDetailsFromPlan } from 'fxa-shared/subscriptions/metadata';
+import { AppContext } from '../../lib/AppContext';
 
 type PaymentConfirmationProps = {
   customer: Customer;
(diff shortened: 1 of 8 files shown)


packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -52,6 +55,8 @@ export const PaymentConfirmation = ({
     interval_count
   );
 
+  const { l10n } = useLocalization();
```
biancadanforth, line 58: I don't need to call this all the time; only if `buttonLabel` is falsy. Is there a way this can be done lazily?

packages/fxa-payments-server/src/components/PlanDetails/index.stories.tsx
```diff
@@ -56,12 +56,7 @@ storiesOf('components/PlanDetail', module)
     </MockApp>
   ))
   .add('localized to xx-pirate', () => (
-    <MockApp
-      appContextValue={{
-        ...defaultAppContext,
-        navigatorLanguages: ['xx-pirate'],
-      }}
-    >
+    <MockApp languages={['xx-pirate']}>
```
biancadanforth, line 59: I just noticed the localization story was broken, because `MockApp` (a Storybook construct) doesn't rely on `navigatorLanguages` prop in `AppContext`. I changed it in the obvious places; here and in the `TermsAndPrivacy` component story.

One alternative is of course to modify `MockApp` to rely on `AppContext` for this, but the `languages` prop approach was an easier change. If people feel strongly one way or the other, happy to take a different path here.

biancadanforth (commented):
Some questions inline. I'll open a PR tomorrow for the Ecosystem Docs update for Stripe metadata and link it as a comment in this PR when I do.

biancadanforth:
I opened mozilla/ecosystem-platform#142 as the doc component of this ticket for review as well.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
chenba: I like your line of thinking here, but I don't think we need to worry about it.  I suspect that that call isn't that expensive.

chenba (approved):
LGTM

PR #8811, opened 2021-06-07
fix(sessions): Add support for touchSessionToken to only update the lastAccess session property

vbudhram:
## Because

- Active sessions in device manager is showing "Boardman, OR, United States" for some devices, this is an AWS server that the gql request is coming from
- ~~Targeting as a point release on train 207 since it is a regression~~ Changin to main branch and will do a cherry pick to train-207

## This pull request

- Add support to only update the last access time property of a session token

Closes: TBD

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/db/index.js
+++ b/packages/fxa-auth-server/lib/db/index.js
@@ -36,10 +36,8 @@ module.exports = (config, log, Token, UnblockCode = null) => {
   } = Token;
   const MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE =
     config.tokenLifetimes.sessionTokenWithoutDevice;
-  const {
-    enabled: TOKEN_PRUNING_ENABLED,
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/db/index.js
```diff
@@ -548,28 +549,36 @@ module.exports = (config, log, Token, UnblockCode = null) => {
...
+        uaBrowserVersion: token.uaBrowserVersion,
+        uaDeviceType: token.uaDeviceType,
+        uaFormFactor: token.uaFormFactor,
+        uaOS: token.uaOS,
+        uaOSVersion: token.uaOSVersion,
+        id,
+      };
+    }
 
     return this.redis.touchSessionToken(uid, t);
```
dannycoates, line 583: looking at `touchSessionToken_1.lua` it seems like having `t` exclude the location / ua fields will erase them from redis, is that true? If so we prob want to change the lua to merge the update.

dannycoates (approved):

PR #8812, opened 2021-06-07
fix(fxa-payments-server): fix upgrade view layout

meandavejustice:
## Because

- we need to allow users to view the upgrade page in a non broken way :)

## This pull request

- updated Product/SubscriptionCreate/index.scss to account for layout
differences with the `grid` rule. We needed different ratios based on
whether `.subscription-title` section was present or not.

## Issue that this pull request solves

Closes: #5369 
Closes: #7805

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [x] I have added necessary documentation (if appropriate).
- [x] I have verified that my changes render correctly in RTL (if appropriate).

## Screenshots (Optional)

product page desktop
image

product page mobile
image

upgrade page desktop
image

upgrade page mobile
image

PR #8835, opened 2021-06-10
fix(payments): display card SVG for UnionPay, Diner's Club, JCB and u…

### LINKED ISSUE #5919: Missing items in “Review your upgrade” page
#### Affected versions:
- All Firefox;

#### Affected FxA versions:
- 1.179.0;

#### Affected Platforms:
- Windows 10 (64-bit);

#### Prerequisites:
- Have and active FxA account;

#### Steps to reproduce:
1. Go to https://123done-stage.dev.lcip.org/;
2. Click the “Sign up” button;
3. Sign up with the account from prerequisites;
4. Click on the “Subscribe for Pro” button from page header;
5. Enter payment details while using a Diners Club/JCB/UnionPay card and finalize subscription;
6. Load in the same browser session the URL: https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/products/prod_GVNS86uMUO3VRM 
7. Observe the page;

#### Expected result:
- “30-day money-back guarantee” text is displayed under the “Review your upgrade” title;
- An icon/img or name of the used card is displayed in the "Billing information" section;

#### Actual result:
- “30-day money-back guarantee” text is not displayed under the “Review your upgrade” title;
- Only the “ending in XXXX” text with a space in front is displayed, looking like something is missing;

missing info

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-2271)
┆Issue Number: FXA-2271

biancadanforth:
…nknown brands

Because:

* There is a blank space in the UI for certain brands that lack an SVG image to display in the PaymentProviderDetails component (which is used by the PaymentConfirmation, PaymentForm and PaymentUpdateForm components).

This commit:

* Adds card SVGs for the remaining supported brands by Stripe per https://stripe.com/docs/payments/cards/supported-card-brands.
* Ensures a generic card SVG is displayed for 'unknown' cards.

Closes #5919

## Checklist

## Screenshots (Optional)

Screen Shot 2021-06-09 at 10 28 49 PM

## Additional notes
* Card SVGs are taken from https://github.com/aaronfagan/svg-credit-card-payment-icons (provided by Peter Benvenuto by way of @jess-cook03 on the design team).
* I applied Dave's in-progress PR #8812 temporarily to take the screenshot above.


packages/fxa-payments-server/src/components/PaymentProviderDetails/index.test.tsx
```diff
@@ -17,7 +17,9 @@ describe('PaymentProviderDetails', () => {
 
     it('Renders credit card logo and last 4 digits', () => {
       const { queryByTestId } = subject();
-      expect(queryByTestId('card-logo-and-last-four')).toBeInTheDocument();
+      const element = queryByTestId('card-logo-and-last-four');
+      expect(element).toBeInTheDocument();
+      expect(element).toHaveClass(`${Customers.CUSTOMER.brand?.toLowerCase()}`);
```
biancadanforth, line 22: I added a single line to this test to assert that `customer.brand` is added as a class, but I’m not sure how else I may want to test this, if at all. This would also have passed before my patch, so I welcome any thoughts.

packages/fxa-payments-server/src/components/PaymentProviderDetails/index.scss
```diff
@@ -46,7 +46,7 @@
   background-image: url('../../images/paypal.svg');
 }
 
-.unbranded::before {
+.unknown::before {
```
biancadanforth, line 49: I removed the "unbranded" class in PaymentProviderDetails' stylesheet, because we were never applying that class. In the case that `brand` is `null`, we don’t even render the `PaymentProviderDetails` component.

Interestingly, while `customer.brand` can be `null` from the auth server, [`charge?.payment_method_details?.card?.brand`](https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-card-brand) can never be `null` from Stripe, so I don't think this is actually a scenario we will ever see. Stripe provides a value of "unknown" for a card brand that doesn't match any of the other known supported brands.

packages/fxa-payments-server/src/lib/test-utils.tsx
```diff
@@ -170,7 +170,8 @@ function injectStripe<P extends Object>(
       <WrappedComponent
         {...{
           ...props,
-          stripe: mockStripe as ReactStripeElements.InjectedStripeProps['stripe'],
+          stripe:
+            mockStripe as ReactStripeElements.InjectedStripeProps['stripe'],
```
biancadanforth, line 174: Prettier changes -- a previous version of this commit had modified this file (find-and-replace).

packages/fxa-payments-server/src/routes/Subscriptions/index.stories.tsx
```diff
@@ -165,19 +165,20 @@ type SubscriptionsRouteProps = {
...
+      <SubscriptionsRoute
+        {...{
+          routeProps: {
+            ...routeProps,
+            ...(baseRouteProps || {}),
+          },
+          ...otherProps,
+        }}
+      />
+    );
```
biancadanforth, line 181: Prettier changes -- a previous version of this commit had modified this file (find-and-replace).

biancadanforth (commented):
Some specific questions and comments in-line for the review.

meandavejustice (changes requested):
We should also update [<AcceptedCards/>](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/Product/AcceptedCards/index.tsx) in this patch

biancadanforth:
> We should also update [`<AcceptedCards/>`](https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/Product/AcceptedCards/index.tsx) in this patch

I don’t know that we actually want to do that, since we also have [currency/country restrictions](https://github.com/mozilla/fxa/blob/62f6acf9270ccddac047cf8209bbbb93524720a9/packages/fxa-auth-server/config/index.ts#L867-L875) that effectively prohibit some cards that Stripe more broadly accepts as payment.

For example, I cannot subscribe to a USD or Euro plan (the only plan currencies we support at the moment) with a China UnionPay card, even though Stripe accepts this card generally. Advertising that we support paying with a certain card and then denying the payment afterward seems like an undesirable pattern, so I would want some design/UX input on this first.

Given this, I suggest we limit this fix to only the issue reported by QA and file a separate issue to consider listing some or all of these cards in `AcceptedCards`.

meandavejustice (approved):

biancadanforth:
I filed [FXA-3631](https://jira.mozilla.com/browse/FXA-3631) for updating (or not) `AcceptedCards`.

PR #8840, opened 2021-06-10
feat(admin): disable account

### LINKED ISSUE #7651: Add feature to administratively disable account
We should have the ability to administratively disable accounts that would prevent users from making any changes to their account, as well as signing in to their account or any reliers

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3101)
┆Issue Number: FXA-3101

### LINKED ISSUE #9140: Add feature to administratively disable account
We should have the ability to administratively disable accounts that would prevent users from making any changes to their account, as well as signing in to their account or any reliers

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3101)

dannycoates:
closes #9140

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -528,6 +528,9 @@ export class AccountHandler {
         email
       );
       accountRecord = res.accountRecord;
+      if (accountRecord.disabledAt) {
+        throw error.cannotLoginWithEmail();
+      }
       // Remember whether they did a signin-unblock,
       // because we can use it to bypass token verification.
       didSigninUnblock = res.didSigninUnblock;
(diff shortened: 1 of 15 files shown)


packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -439,6 +439,9 @@ export class AccountHandler {
         email
       );
       accountRecord = res.accountRecord;
+      if (accountRecord.disabledAt) {
+        throw error.cannotLoginWithEmail();
+      }
```
dannycoates, line 444: This is the one and only auth-server touch point

dannycoates (commented):
The admin-panel ui isn't the greatest but it does the thing

Issue #8861, opened 2021-06-15
Increase stripe payload size allowed


data-sync-user:
We are currently rejecting stripe payloads because we restrict them to 52kb. We should increase this to 1 MB to avoid having this issue in the future as Stripe adds object on occasion for backwards compatibility in API revisions.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3635)
┆Issue Number: FXA-3635

data-sync-user:
➤ Barry Chen commented:

Current limit is hard code here: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts#L628:L630 ( https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts#L628:L630 )

We want to make it configurable.

PR #8873, opened 2021-06-15
fix(subscriptions): increase webhook payload limit

chenba:
Because:
 - some Stripe webhook requests were failing because their payload size
   larger than our endpoint's limit

This commit:
 - make the payload limit configurable
 - set the default to 1MB

Closes: #8861

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -111,8 +111,7 @@ const conf = convict({
   },
   memcached: {
     address: {
-      doc:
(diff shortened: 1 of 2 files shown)


chenba:
The config file has a lot of prettier formatting changes, but the only real change is at https://github.com/mozilla/fxa/pull/8873/files#diff-887024629b0aaf52c1338b1c4123e92fba1f360295a3e8bcef3bb7363fafdf86R834-R839.  Ignore the rest.

bbangert (approved):
👍🏻

PR #8874, opened 2021-06-15
Localize our rate limit on adding recovery key in new settings

vbudhram:
## Because

- We didn't show a localized rate limit in the new settings

## This pull request

- Add support to localize our rate limit when creating a recovery key

Closes: #9322

## Checklist

## Screenshots (Optional)

Screen Shot 2021-06-15 at 1 18 02 PM

### DIFF
--- a/packages/fxa-settings/src/lib/auth-errors/auth-errors.ts
+++ b/packages/fxa-settings/src/lib/auth-errors/auth-errors.ts
@@ -2,59 +2,60 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import {AuthServerError} from "fxa-auth-client/browser";
+import { AuthServerError } from 'fxa-auth-client/browser';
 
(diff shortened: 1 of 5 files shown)


packages/fxa-settings/src/lib/auth-errors/auth-errors.ts
```diff
@@ -604,48 +596,47 @@ const ERRORS = {
...
+).reduce((acc: { [key in ErrorKey]: AuthUiError }, [k, v]) => {
+  const e = new Error(v.message) as AuthUiError;
+  e.errno = v.errno;
+  acc[k] = e;
+  return acc;
+}, {} as Record<ErrorKey, AuthUiError>);
 
-export const AuthUiErrorNumbers: {
-    [key in ErrorKey]: number;
+export const AuthUiErrorNos: {
```
vbudhram, line 621: I must have added this with incorrect formatting :/. `AuthUiErrorNos` is the only changed function here.

dannycoates (approved):

PR #8875, opened 2021-06-15
moving db queries

### LINKED ISSUE #9723: Implement auth-server db api in fxa-shared
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3683)

dannycoates:
This eliminates the runtime dependency of auth-server on auth-db-mysql. It's no longer required to run the auth-db-mysql process in production for the auth-server. This change doesn't completely remove the need for auth-db-mysql, email-service still needs it, and in dev a portion of it is required to run migrations, but it should now be considered deprecated and we should continue working to remove it completely.

closes #9723


dannycoates:
Once this lands we can start using direct db calls in graphql-api-server instead of routing through auth-server 🎉

vbudhram (approved):
@dannycoates 🌮

Issue #8735, opened 2021-05-26
Create FxA account route for no password creation


data-sync-user:
To streamline the checkout flow, the checkout page will request an account to be created for a new email address. This route should work similarly to the existing create route, except instead of returning a sessionToken for a logged in user, it will return an auth token for use with subscription routes for subscription creation.

This route should have a placeholder TODO, marking where the 'Finish Setting up your Account' email will be triggered.

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/FXA-3599)
┆Issue Number: FXA-3599

PR #8892, opened 2021-06-17
refactor(auth): split account create functions out

bbangert:
Because:

* Creating an account with no password needs to re-use most of the parts
  of accountCreate with some slight alterations. The existing
  accountCreate needed its component parts split out for re-use.

This commit:

* Refactors the nested functions in accountCreate into private functions
  on the class for future re-use.

Issue #8735

## Checklist


vbudhram (approved):
@bbangert LGTM 👍🏽

Issue #9244, opened 2021-06-21
[emails M1] Set up state previewing solution


data-sync-user:
FXA-3503 / #8600 decided what state previewing solution we should use.

We need to add this on top of the templating system added in FXA-3504 / #8601. A dummy "hello world!" state would be great to show for now.

If we _didn't_ choose Storybook in #8600, we may need an additional ticket for deployment (which might be done outside of this epic). Additionally, it would be great to play with a "preview in Mail client" solution here, mentioned in #8600. If it gets a little hairy, we should create a different ticket to explore that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3505)

LZoog:
In #8765, we chose to use Storybook.

Storybook for EJS and MJML is a bit different than it is for React. Les created a fantastic [starter repo](https://github.com/lmorchard/mjml-in-storybook) for this kind of stack.

To close this ticket, we just need Storybook to run locally and show one template with state changes. Also, unless we need it for another reason, we can remove any related write-to-disk previewing stuff. We are likely going to need to create a "deploy EJS/MJML Storybook to GCP" follow up ticket and add it to a "nice to haves" emails epic - it's OK for us to do that later, after we implement it for development use.

PR #9704, opened 2021-06-22
feat(auth-server) : Setup EJS+MJML templating solution in fxa-auth-server

Yogita98:
## Because

- In regards to the decision taken in the [ADR](https://github.com/mozilla/fxa/blob/main/docs/adr/0024-upgrade-templating-toolset-of-auth-server-emails.md), we wanted to setup a new templating stack for our auth-server emails

## This pull request

- Includes packages and scripts for the new templating toolkit that could be used to develop emails

Closes: #9689 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
@@ -1,0 +1,45 @@
+export const appBadges = `
+  <mj-section>
+    <mj-group>
+      <mj-column>
+        <mj-image
+          css-class="app-badges"
+          href="<%= iosURL %>"
+          src="https://image.e.mozilla.org/lib/fe9915707361037e75/m/5/ebdfb903-4e62-4aeb-9eae-26e1ef495049.png"
+          alt="iOS badge"
+        ></mj-image>
+      </mj-column>
+      <mj-column>
+        <mj-image
+          css-class="app-badges"
+          href="<%= androidURL %>"
+          src="https://image.e.mozilla.org/lib/fe9915707361037e75/m/5/0dda16b4-21bb-4cc1-9042-2ebaaa7a9764.png"
+          alt="android badge"
+        ></mj-image>
+      </mj-column>
+    </mj-group>
+  </mj-section>
+  <mj-section>
+    <mj-column>
+      <mj-text css-class="secondary-text">
+        <% if(onDesktopOrTabletDevice){ %> Or, install on
+        <span
+          href="<%= anotherDeviceURL %>"
+          color="#0a84ff"
+          text-decoration="none"
+          font-family="sans-serif"
+          >another version</span
+        >
+        <% } else{ %> Or, install on
+        <span
+          href="<%= anotherDeviceURL %>"
+          color="#0a84ff"
+          text-decoration="none"
+          font-family="sans-serif"
+          >another device</span
+        >
+        <% } %>
+      </mj-text>
+    </mj-column>
+  </mj-section>
+`;
--- a/packages/fxa-auth-server/lib/senders/emails/partials/button.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/button.ts
@@ -1,0 +1,9 @@
+export const button = `
+  <mj-section>
+    <mj-column>
+      <mj-button>
+        <%= buttonText %>
+      </mj-button>
+    </mj-column>
+  </mj-section>
+`;
--- a/packages/fxa-auth-server/lib/senders/emails/renderer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/renderer.ts
@@ -1,0 +1,21 @@
+import ejs = require('ejs');
+import mjml2html = require('mjml');
+import * as templates from './templates';
+
+export const context = {
+  buttonText: 'Sync device',
+  onDesktopOrTabletDevice: true,
+  anotherDeviceURL:
+    'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  iosURL: '',
+  androidURL: '',
+};
+
+export function render(
+  templateName: keyof typeof templates,
+  context: Record<any, any>
+) {
+  const template = ejs.compile(templates[templateName].render());
+  const mjmlTemplate = template(context);
+  return mjml2html(mjmlTemplate).html;
+}
--- a/packages/fxa-auth-server/lib/senders/emails/styles.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/styles.ts
@@ -1,0 +1,53 @@
+export const style = `
+.header-text div {
+  font-family: sans-serif !important;
+  font-size: 21px !important;
+  line-height: 29px !important;
+  font-weight: normal !important;
+  text-align: center !important;
+}
+td {
+  padding: 5px !important;
+}
+span {
+  color: #0a84ff;
+}
+.primary-text div {
+  font-family: sans-serif !important;
+  font-size: 14px !important;
+  line-height: 21px !important;
+  font-weight: normal !important;
+}
+.secondary-text div {
+  font-family: sans-serif !important;
+  font-weight: normal !important;
+  text-align: center !important;
+  color: #737373 !important;
+  font-size: 11px !important;
+  line-height: 18px !important;
+}
+.app-badges img {
+    width: 152px !important;
+  height: 45px !important;
+}
+.fxa-logo td {
+  height: 60px !important;
+  width: 60px !important;
+}
+.sync-logo td {
+  height: 175px !important;
+  width: 238px !important;
+}
+.primary-button {
+  font-family: sans-serif !important;
+  color: #fff !important;
+  padding: 15px !important;
+  text-decoration: none !important;
+  width: 280px !important;
+  font-size: 18px !important;
+  line-height: 26px !important;
+  background-color: #0a84ff !important;
+  height: 50px !important;
+  width: 310px !important;
+}
+`;
--- a/packages/fxa-auth-server/package.json
+++ b/packages/fxa-auth-server/package.json
@@ -27,7 +27,9 @@
     "write-emails": "node -r ts-node/register ./scripts/write-emails-to-disk.js",
     "clean-up-old-ci-stripe-customers": "ts-node ./scripts/clean-up-old-ci-stripe-customers.js",
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
-    "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts"
+    "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
+    "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
   },
   "repository": {
     "type": "git",
@@ -47,6 +49,8 @@
     "@sentry/node": "^6.7.2",
     "@type-cacheable/core": "^5.2.0",
     "@type-cacheable/ioredis-adapter": "^5.2.0",
+    "@types/ejs": "^3.0.6",
+    "@types/mjml": "^4.7.0",
     "ajv": "^6.12.2",
     "aws-sdk": "^2.934.0",
     "base64url": "3.0.1",
@@ -58,6 +62,7 @@
     "convict": "^6.1.0",
     "convict-format-with-moment": "^6.0.1",
     "convict-format-with-validator": "^6.0.1",
+    "ejs": "^3.1.6",
     "email-addresses": "4.0.0",
     "fxa-geodb": "workspace:*",
     "fxa-jwtool": "^0.7.2",
@@ -75,6 +80,7 @@
     "lodash": "^4.17.21",
     "luxon": "^1.27.0",
     "memcached": "^2.2.2",
+    "mjml": "^4.10.1",
     "moment-timezone": "^0.5.33",
     "mozlog": "^3.0.2",
     "mysql": "^2.18.1",
@@ -115,6 +121,7 @@
     "@types/nock": "^11.1.0",
     "@types/node": "^15.12.2",
     "@types/node-zendesk": "^2.0.2",
+    "@types/nodemailer": "^6.4.2",
     "@types/request": "2.48.5",
     "@types/safe-regex": "1.1.2",
     "@types/uuid": "^8.3.0",
--- a/packages/fxa-auth-server/lib/senders/emails/preview-in-maidev.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/preview-in-maidev.ts
@@ -1,0 +1,33 @@
+import nodemailer = require('nodemailer');
+import * as templates from './templates';
+import { render, context } from './renderer';
+
+const transporter = nodemailer.createTransport({
+  port: 9999,
+  host: 'localhost',
+  tls: {
+    rejectUnauthorized: false,
+  },
+});
+
+const message = {
+  from: 'noreply@domain.com',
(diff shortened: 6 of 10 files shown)


Yogita98:
Hi @LZoog @jodyheavener 
I have opened this PR to setup the EJS+MJML templating solution for `fxa-auth-server` emails as discussed in the [ADR](https://github.com/mozilla/fxa/blob/main/docs/adr/0024-upgrade-templating-toolset-of-auth-server-emails.md). 
This PR includes `cadReminder` template (which may or may not be the final version) to demo how partials and styles could be implemented with this solution.

Yogita98:
Hi @LZoog @jodyheavener 
I have opened this PR to setup the EJS+MJML templating solution for `fxa-auth-server` emails as discussed in the [ADR](https://github.com/mozilla/fxa/blob/main/docs/adr/0024-upgrade-templating-toolset-of-auth-server-emails.md). 
This PR includes `cadReminder` template (which may or may not be the final version) to demo how partials and styles could be implemented with this solution.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
```diff
@@ -0,0 +1,21 @@
+export const appBadges = `<mj-section>
```
LZoog, line 1: I think the string literals are OK for the initial PR here, but we'll definitely want to create a follow up in this milestone to see what we can do here. This is not very dev-friendly 😝

packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
```diff
@@ -0,0 +1,21 @@
...
+    <mj-group>
+        <mj-column>
+            <mj-image css-class="app-badges" href="{{{iosUrl}}}" src="https://image.e.mozilla.org/lib/fe9915707361037e75/m/5/ebdfb903-4e62-4aeb-9eae-26e1ef495049.png" alt="ios-badge"></mj-image>
```
LZoog, line 4: Oh wow, I'm surprised we hardcode this URL in the old templates but we certainly do!

We may want to pass in "iOS badge" for the `alt` text as well as adjust the one for android, a screenreader may not read that differently than "ios-badge" but it _may_ and having a dash there seems odd.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
```diff
@@ -0,0 +1,21 @@
...
+        <mj-column>
+            <mj-image css-class="app-badges" href="{{{androidUrl}}}" src="https://image.e.mozilla.org/lib/fe9915707361037e75/m/5/0dda16b4-21bb-4cc1-9042-2ebaaa7a9764.png" alt="android-badge"></mj-image>
+        </mj-column>
+    </mj-group>
+</mj-section>
+<mj-section>
+    <mj-column>
+        <mj-text css-class="secondary-text">
+            <% if(onDesktopOrTabletDevice){ %>
+                Or, install on <span href="http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device" color="#0a84ff" text-decoration="none" font-family="sans-serif">another version</span>
```
LZoog, line 15: We definitely don't want any URLs containing `localhost`. This should be some sort of variable passed down.

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+let templateName: keyof typeof templates;
+for (templateName in templates) {
+  const template = ejs.compile(templates[templateName].render());
+  const mjmlTemplate = template(context);
+  const htmlTemplate = mjml2html(mjmlTemplate).html;
+  let prettyHtml = prettier.format(htmlTemplate, { parser: 'html' });
+  let outputFile = path.join(__dirname, 'dist', `${templateName}.html`);
+  fs.writeFileSync(outputFile, prettyHtml);
```
LZoog, line 21: You probably know this already but we probably won't need this `writeFileSync` when Storybook is integrated. Fine for now though. :)

packages/fxa-auth-server/lib/senders/emails/styles.ts
```diff
@@ -0,0 +1,61 @@
+export const style = `
+.header-text div{
```
LZoog, line 2: Oh man, the CSS being string literals feels almost worse than the MJML/EJS lol.

Hmm, I was going to ask if it was possible to adjust these selectors so that they don't often end in a generic element like `div` but looks like MJML is doing its MJML magic here.

packages/fxa-auth-server/lib/senders/emails/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+    font-family: sans-serif !important;
+    font-size: 21px !important;
+    line-height: 29px !important;
```
LZoog, line 5: I'm guessing all of the ones marked `!important` have to be done that way?

Ooh, let's definitely use this as an opportunity to make the styles more consistent. If you look at the design guide in Settings (`yarn storybook` in `fxa-settings`), you'll see we're going by units of `4px`. It's unfortunate [some email clients don't seem to support rem](https://www.caniemail.com/features/css-unit-rem/) which is what Tailwind/the design guide uses, so we'll have to use `px` or `em`. Let's try to set some patterns here by checking out the Settings design guide - I kind of feel like we could use all `em`s here to better translate from the design guide rather than using `px`, curious what @jodyheavener thinks.

packages/fxa-auth-server/lib/senders/emails/styles.ts
```diff
@@ -0,0 +1,61 @@
...
+.secondary-text div {
+    font-family: sans-serif !important;
+    font-weight: normal !important;
+    text-align: center !important;
+    color: #737373 !important;
+    font-size: 11px !important;
+    line-height: 18px !important;
+    }
+
+.app-badges img {
```
LZoog, line 34: I'm not sure these should go in a "global" styles file since not all of them will reference `app-badges`, this one is going to get inserted on every template right? Also OK for that to happen in a follow up once the CSS approach is better sussed out... it'd be nice if we could use SCSS in this kind of file.

LZoog (commented):
Looking good so far! Have a couple comments and a couple questions.

I don't think this closes #9244, though. It should aim to close #9689. (ugh, looks like Jira migration messed up some stuff with labels, I'll look into that tomorrow!)

packages/fxa-auth-server/lib/senders/emails/renderer.ts
jodyheavener: Since Storybook is the very next step I'm tempted to suggest that we try and convert this over in this PR. Could look something like this:

```ts
export function render(
  templateName: keyof typeof templates,
  context: Record<any, any>
) {
  const template = ejs.compile(templates[templateName].render());
  const mjmlTemplate = template(context);
  return mjml2html(mjmlTemplate).html;
}
```

Then again, this may look entirely different than what Les's Storybook implementation had in mind.

packages/fxa-auth-server/lib/senders/emails/styles.ts
jodyheavener: To any onlookers: we're using template literals here for now because we couldn't figure out how to import partials using EJS and it was blocking. We can switch later.

jodyheavener (commented):
This looks great and the rendered cadReminders email looks fantastic! I think you're going to want to get your VS Code set up to work with FXA's Prettier config. We generally use two-space indentation across JS/TS/CSS/JSON files. Let me know if you'd like a hand with that. I also agree with Lauren's comments. Nicely done!

packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
Yogita98: Yeah sure. I couldn't find another way of using `mjml` inside typescript, but sure we can do some digging around this down the line

packages/fxa-auth-server/lib/senders/emails/styles.ts
jodyheavener: I'm torn about this. I do agree that it would be nice to make these styles consistent, but I am resistant for two reasons:

1. They were written by someone, and given the specificity it seems like they were intentional
2. We aren't updating designs in this project, and while these are small changes they feel like a slippery slope, will also probably require a more thorough design review

Open to pushback though, I think if we're gonna do this we should make a formal plan for it!

LZoog: Hmm, but if we had chosen the React and Tailwind option, it would have forced us to style them for consistency. To me, one big win from "redoing" these was consistent styles - I feel like it would be a lost opportunity if we keep the random values.

> given the specificity it seems like they were intentional

As someone that has updated our emails and styles before... not really, lol. Most of the time, the devs just pick a px value they think looks good and they just go with it. I feel like it would be fine to round to the nearest unit of 4 for padding and margin, and then we can come up with font-sizes and corresponding line-heights very similar to what we're getting out of Tailwind.

Yogita98:
I have tried to address most of the comments in my recent commit as well as added 2 separate scripts for previewing the emails with `maildev` and saving the files on disk.
Thanks alot @jodyheavener for sharing the [snippet](https://github.com/mozilla/fxa/pull/9704#discussion_r658222260), I found it particularly useful as we can reuse it to implement multiple things including Storybook.
To preview the email with `maildev` though, we might have to follow the below instructions:
```
yarn global add maildev
maildev -s 9999
```

packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts
```diff
@@ -0,0 +1,21 @@
...
+    <mj-column>
+        <mj-text css-class="secondary-text">
+            <% if(onDesktopOrTabletDevice){ %>
+                Or, install on <span href="<%= anotherDeviceURL %>" color="#0a84ff" text-decoration="none" font-family="sans-serif">another version</span>
+                <% } else{ %>
+                    Or, install on <span href="<%= anotherDeviceURL %>" color="#0a84ff" text-decoration="none" font-family="sans-serif">another device</span>
+                    <% } %>
+        </mj-text>
+    </mj-column>
+</mj-section>`;
```
jodyheavener, line 21: Nitpick: can we make this indentation consistence with the rest of our HTML (2 spaces)?

packages/fxa-auth-server/lib/senders/emails/partials/button.ts
```diff
@@ -0,0 +1,7 @@
+export const button = `<mj-section>
+    <mj-column>
+        <mj-button>
+            <%= buttonText %>
+        </mj-button>
+    </mj-column>
+</mj-section>`;
```
jodyheavener, line 7: here too

packages/fxa-auth-server/package.json
```diff
@@ -141,6 +148,7 @@
     "load-grunt-tasks": "^5.1.0",
     "lodash.chunk": "4.2.0",
     "lodash.pick": "4.4.0",
+    "maildev": "^1.1.0",
```
jodyheavener, line 151: Can I get you to commit the changes to the `yarn.lock` file after a `yarn install`?

jodyheavener, line 151: If you install maildev globally (`yarn global add maildev`) do we need maildev as a project dependency?

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as templates from './templates';
+
+export const context = {
+  buttonText: 'Sync device',
+  onDesktopOrTabletDevice: true,
+  anotherDeviceURL:
+    'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  iosURL: '',
+  androidURL: '',
+};
```
jodyheavener, line 12: Looks like this context isn't being used any more and can be removed (context is coming from the function call now)

packages/fxa-auth-server/package.json
```diff
@@ -27,7 +27,9 @@
     "write-emails": "node -r ts-node/register ./scripts/write-emails-to-disk.js",
     "clean-up-old-ci-stripe-customers": "ts-node ./scripts/clean-up-old-ci-stripe-customers.js",
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
-    "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts"
+    "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
+    "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
+    "preview": "ts-node lib/senders/emails/preview-in-maildev.ts"
```
jodyheavener, line 32: Since this command exists in the fxa-auth-server space, calling it "preview" isn't super clear. Maybe `maildev-preview`?

jodyheavener (changes requested):
I think we're super close and the Maildev stuff works great! Few more changes I think are needed then we could get it in :)

packages/fxa-auth-server/lib/senders/emails/renderer.ts
Yogita98: I am actually using this context object in both the scripts [`preview-in-maidev`](https://github.com/Yogita98/fxa/blob/EJS%2BMJML-solution/packages/fxa-auth-server/lib/senders/emails/preview-in-maildev.ts#L3) and [`render-to-disk`](https://github.com/Yogita98/fxa/blob/EJS%2BMJML-solution/packages/fxa-auth-server/lib/senders/emails/render-to-disk.ts#L4) and couldn't really find a better place to import it from but I can definitely either create a separate file or declare the object individually in both the files. Let me know your thoughts

packages/fxa-auth-server/package.json
Yogita98: Yeah I'll definitely add it with my next commit

Yogita98: Yeah we don't need it. Thanks for pointing it out though I was trying to get a `maildev` server up with the help of a javascript file but since it's mentioned in the [docs](https://github.com/mozilla/fxa#running-with-maildev) to manually get it up, I stopped looking into it and probably forgot to remove that dependency.

packages/fxa-auth-server/lib/senders/emails/renderer.ts
jodyheavener: Oh my bad I didn't see that it was exported. Never mind!

jodyheavener:
@Yogita98 need to rebase to updte the yarn.lock file, but @LZoog I think I'm happy with this, what you think?

LZoog:
Yes this LGTM for closing out #9689!! 👏👏

@Yogita98 will you update your commit message to follow the contributor guidelines with the "this commit" and "because"? I'll check tomorrow morning, once you fix the lock file and fix the `yarn.lock` conflict, r+ and merge :shipit:

Yogita98:
Hey @LZoog 
I cannot see any merge conflicts in the PR so I think I may have been able to fix the `yarn.lock` file. Apologies for a little mess here though, I always have a hard time squashing the merge commits

LZoog (approved):
@Yogita98 No prob, thanks for your work here! Will you actually reword your commit message to follow the [commit guidelines](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md#git-commit-guidelines) so this shows up in the release script (which looks for that commit message format)? Will merge after that 😄 

(btw, left you some messages in Matrix!)

Yogita98:
Ahhh my bad! Also, thanks for sharing those pointers on Matrix as well as checking back on the PR on a weekend. You are just awesome :100:

LZoog:
@Yogita98 sorry to circle back so much! It looks like [the commit](https://github.com/mozilla/fxa/pull/9704/commits) is still worded the same. 🤪

Yogita98:
My bad! I hope it's ready to be merged now

PR #9707, opened 2021-06-22
fix(emails): send email for cancelation due to failed payments

chenba:
Because:
 - we should let a customer know that their subscription was canceled
   due to failed payments

This commit:
 - send an email on a subscription deleted event if the deletion is not
   from an account deletion or customer initiated cancelation

Closes: #9172 

## Screenshots (Optional)
image

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -540,6 +540,16 @@ export class StripeWebhookHandler extends StripeHandler {
   /**
    * Send out the appropriate email on subscription deletion, depending on
    * whether the user still has an account.
+   *
+   * We receive a subscription deleted event for the following:
+   *   1. A user canceled subscripiton.
+   *   2. Subscription canceled after multiple Stripe attempts to pay an invoice.
+   *   3. PayPal processor canceled the subscription after failed attempts.
+   *   4. A user deleted their account.
+   *   5. An admin or support agent cancelled the subscription.
+   *
+   * (1) and (5) are handled at the time of cancelation, so we do not send an additional
+   * email here.
    */
   async sendSubscriptionDeletedEmail(subscription: Stripe.Subscription) {
     if (typeof subscription.latest_invoice !== 'string') {
@@ -565,8 +575,18 @@ export class StripeWebhookHandler extends StripeHandler {
     let account;
     try {
       // If the user's account has not been deleted, we should have already
-      // sent email at subscription update when cancel_at_period_end = true
-      await this.db.account(uid);
+      // sent email at subscription update when cancel_at_period_end = true,
+      // _or_ the cancellation is from Stripe due to failed retries or the
+      // PayPal processor, which we'll handle here.
+      account = await this.db.account(uid);
+      await this.mailer.sendSubscriptionFailedPaymentsCancellationEmail(
+        account.emails,
+        account,
+        {
+          acceptLanguage: account.locale,
+          ...invoiceDetails,
+        }
+      );
     } catch (err) {
       // Has the user's account been deleted?
       if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
@@ -585,6 +605,7 @@ export class StripeWebhookHandler extends StripeHandler {
         );
       }
     }
+
     return invoiceDetails;
   }
 }
--- a/packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
+++ b/packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
@@ -1,0 +1,15 @@
+{{! Cancellation email has no icon }}
+{{#*inline "icon"}}
+{{/inline}}
+
+{{#*inline "title"}}
+  {{t "Your subscription has been cancelled" }}
+{{/inline}}
+
+{{#*inline "content"}}
+  {{t "We've cancelled your %(productName)s subscription because multiple payment attempts failed. To get access again, start a new subscription with an updated payment method." }}
+  <br><br>
+  {{{t "Please help us improve our services by taking this <a %(cancellationSurveyLinkAttributes)s>short survey</a>." }}}
+{{/inline}}
+
+{{> subscriptionEmail}}
--- a/packages/fxa-auth-db-mysql/lib/db/patch.js
+++ b/packages/fxa-auth-db-mysql/lib/db/patch.js
@@ -5,4 +5,4 @@
 // The expected patch level of the database. Update if you add a new
 // patch in the ./schema/ directory.
 
-module.exports.level = 117;
+module.exports.level = 118;
--- a/packages/fxa-auth-db-mysql/lib/db/schema/patch-117-118.sql
+++ b/packages/fxa-auth-db-mysql/lib/db/schema/patch-117-118.sql
@@ -1,0 +1,12 @@
+--
+-- This migration adds the subscriptionFailedPaymentsCancellation emailType to
+-- the emailTypes table.
+--
+
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('117');
+
+INSERT INTO emailTypes (emailType) VALUE ('subscriptionFailedPaymentsCancellation');
+
+UPDATE dbMetadata SET value = '118' WHERE name = 'schema-patch-level';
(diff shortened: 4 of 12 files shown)


packages/fxa-auth-db-mysql/lib/db/schema/patch-116-117.sql
```diff
@@ -0,0 +1,12 @@
+--
+-- This migration adds the subscriptionRenewalReminder emailType to
```
biancadanforth, line 2: Can you update the comment to say `subscriptionFailedPaymentsCancellation`?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -540,6 +540,15 @@ export class StripeWebhookHandler extends StripeHandler {
...
    * whether the user still has an account.
+   *
+   * We receive a subscription deleted event for the following:
+   *   1. A user canceled subscripiton.
+   *   2. Subscription canceled after multiple Stripe attempts to pay an invoice.
+   *   3. PayPal processor canceled the subscription after failed attempts.
+   *   4. A user deleted their account.
+   *
+   * (1) is handled at the time of cancelation, so we do not send an additional
+   * email here.
```
biancadanforth, line 551: This is awesome thank you for documenting this.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -565,8 +574,18 @@ export class StripeWebhookHandler extends StripeHandler {
...
+      // PayPal processor, which we'll handle here.
+      account = await this.db.account(uid);
+      await this.mailer.sendSubscriptionFailedPaymentsCancellationEmail(
+        account.emails,
+        account,
+        {
+          acceptLanguage: account.locale,
+          ...invoiceDetails,
+        }
+      );
```
biancadanforth, line 588: According to my understanding of [Stripe's docs on subscription cancellations](https://stripe.com/docs/billing/subscriptions/cancel#events), even users who voluntarily cancel a subscription (i.e. we mark it as `cancel_at_period_end` and get a `subscription.updated` webhook) will still trigger a `subscription.deleted` webhook once the subscription period ends. With the logic as written, we would send users who voluntarily cancel this involuntary cancellation email at that time as well.

I think we will want to update this logic to check for a voluntary cancellation prior. I'm not sure what the best way is to do that -- can we look at the webhook event history for this `subscription` and find the most recent `subscription.updated` event for the current billing period to see if the change was to set `cancel_at_period_end`?

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
```diff
@@ -0,0 +1,15 @@
+{{! Cancellation email has no icon }}
+{{#*inline "icon"}}
+{{/inline}}
+
+{{! This email has no title }}
+{{#*inline "title"}}
+{{/inline}}
```
biancadanforth, line 7: Nit: Maybe we should have a title here? The email looks a little odd without one, and the other cancellation email has a title ("Sorry to see you go").

Screen Shot 2021-07-07 at 2 27 00 PM
Screen Shot 2021-07-07 at 2 27 09 PM

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
```diff
@@ -0,0 +1,15 @@
...
+
+{{#*inline "content"}}
+  {{t "We've cancelled your %(productName)s subscription because multiple payment attempts failed. To get access again, start a new subscription with an updated payment method." }}
+  <br><br>
+  {{{t "Please help us improve our services by taking this <a %(cancellationSurveyLinkAttributes)s>short survey</a>." }}}
```
biancadanforth, line 12: I filed https://mozilla-hub.atlassian.net/browse/FXA-3717 to address this stale survey. I'd like us to remove it until we have a better alternative, but we'll see what others think.

biancadanforth (changes requested):
Thanks Barry for taking care of this longstanding bug.

IIUC, based on this logic, users who voluntarily cancel a subscription would also receive this involuntarily cancelled email (see more details in line), so that may be worth looking into more.

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
chenba: Sure.  Should that be a follow-up or should it block the issue?  I prefer it as a follow-up as we might end up waiting a while for the content.  (Unless you have a suggestion for the title?? 😀)

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
chenba: Leslie filed a PR to update the survey link and I reviewed and merge it.  I'll update the url here.  Thanks for catching it.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: This was my biggest concern while working on this patch.  The diff view is not great for this...if you look at line 568, you see that we return early if it was a voluntary cancellation.  I'll see about adding a test for it.

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
biancadanforth: Definitely not blocking. Up to you.

biancadanforth: Hm... "Your subscription has been cancelled"?

packages/fxa-auth-server/lib/senders/templates/subscriptionFailedPaymentsCancellation.html
biancadanforth: Oh thanks for letting me know; I updated that issue (which I may close or modify depending on what Wil/Tony/Brian think). I'm glad it's updated, though we'll still need to have a survey for all products eventually or support product-specific surveys.

chenba: err I meant rebase to update the survey url.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
biancadanforth: Oh okay I see that; my bad... So the logic looks right to me in that case for voluntary cancellations through Subscription Management. If Support cancels a subscription for a customer (using `cancel_at_period_end`), do they modify this metadata property? I know that's an edge case...

biancadanforth: I won't block on adding a test for this, since this isn't new logic that you added as part of this PR.

biancadanforth (approved):
The thing I was blocking on was a misunderstanding; we wouldn't send this involuntary cancellation email to a user who previously voluntarily cancelled.

None of my other comments are blocking (except please do update that one comment in the patch file), so I will leave it to your discretion to file follow-ups or add any additional changes.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
chenba: > If Support cancels a subscription for a customer (using cancel_at_period_end), do they modify this metadata property?

hmm good point.  I'm guessing that they don't.

chenba:
I've made some updates based on the initial review.  But @biancadanforth made a good point with https://github.com/mozilla/fxa/pull/9707#discussion_r665703630.

Since customers do receive emails regarding failed payments prior to their subscription being canceled, the cancellation is not a _complete_ surprise even if they don't get an email about it.  I want to spend some time to see if there's a way to distinguish between the failed payment cancellations and the admin/support agent cancellations; I prefer to not have any false positives for this email.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
biancadanforth: Is it worth pinging JR in the Jira ticket for this bug to ask them to update their instructions to include modifying this?

chenba: Perhaps as a last resort.  I prefer to not rely on a manual process if we can avoid it.

chenba:
The latest commit is only a comment update.  The metadata for an admin/support agent cancellation is the same as a voluntary (API) cancellation, so no code change is required.

chenba:
All right, I'm making an executive decision to merge despite the failed CI caused by something that's not related.

Issue #9706, opened 2021-06-22
[new user checkout] Create new user email form (baseline component, storybook, sanity test)


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3681)

data-sync-user:
➤ Barry Chen commented:

Dave Justice  can you add a description for what’s being done in this issue?  Also, doesn’t this block the email validation issue?

data-sync-user:
➤ Dave Justice commented:

Barry Chen yes it blocks email validation, the work is complete and waiting for review here https://github.com/mozilla/fxa/pull/9708 ( https://github.com/mozilla/fxa/pull/9708|smart-link )

data-sync-user:
➤ Bianca Danforth commented:

I think Barry’s asking if you would:

  * Add more context to this ticket, since the description is currently blank.
  * Link the email validation ticket to this ticket (and anything else that depends on it) as being blocked by it.

I’m still in the process of reviewing it, but I left some feedback already in the PR.

PR #9708, opened 2021-06-23
task(fxa-payments-server): create newUserEmailForm component

meandavejustice:
## Because

- we need to be able to get a new user's email address
- we need a method to check that the user doesn't have an existing account

## This pull request

- adds the `NewUserEmailForm` component

Closes: #9706

## Checklist

## Screenshots (Optional)

firefox_9FmAmFJqat

## Additional Information

https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=310%3A9638


packages/fxa-payments-server/src/components/NewUserEmailForm/index.scss
```diff
@@ -0,0 +1,34 @@
+@import '../../../../fxa-content-server/app/styles/variables';
+
+.new-user-email-form {
+    position: relative;
+
+    .sign-in-copy {
```
biancadanforth, line 6: It looks like the "Sign in" link should be black and underlined per the [final designs](https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=315%3A12945).

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+storiesOf('components/NewUserEmailForm', module).add('default', () => (
+  <div style={{ display: 'flex' }}>
+    <NewUserEmailForm
+      onBlurInitial={(evt) => {
+        console.log('onBlurInitial: ', evt);
+      }}
+      onBlurConfirm={(evt) => {
+        console.log('onBlurInitial: ', evt);
+      }}
```
biancadanforth, line 13: I don't see these logging when I blur these `Input` fields in the form. Do they log for you?

I do see a tooltip that says "Sign in with your Firefox Account", but I assume we'll change that once [email validation](https://mozilla-hub.atlassian.net/browse/FXA-3665) and/or [existing account checking](https://mozilla-hub.atlassian.net/browse/FXA-3663) is implemented.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+export type NewUserEmailFormProps = {
+  onBlurInitial: (event: React.FocusEvent<HTMLInputElement>) => void;
+  onBlurConfirm: (event: React.FocusEvent<HTMLInputElement>) => void;
+  getString?: Function;
+  validatorInitialState?: ValidatorState;
+  validatorMiddlewareReducer?: ValidatorMiddlewareReducer;
+};
+
+export const NewUserEmailForm = ({
```
biancadanforth, line 23: If we have a separate HTML `form` for the `NewUserEmailForm`, how do we:
* gate enabling/disabling the submit buttons and
* pass along the email address

...to the `PaymentForm` in the "Choose your payment method" section? This would make it so we have two separate HTML `form` elements on this page where conceptually, I would think we only want one.

I just want to make sure we have a plan for that going forward as we continue extending this page.

Curious what @chenba may think here as well.

biancadanforth (commented):
Hey Dave, thanks for this PR. I'm not done looking this over, but I think there's enough feedback here to share already. Hoping to finish looking things over later tonight.

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -362,3 +362,13 @@ payment-confirmation-amount-year = { $intervalCount ->
 
 payment-confirmation-download-button = Continue to download
 payment-confirmation-cc-card-ending-in = Card ending in { $last4 }
+
+## new user email form
+new-user-sign-in-link = Already have a Firefox account? <a>Sign in</a>
```
chenba, line 367: 'Firefox' is a brand name and there's a term for it.

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -362,3 +362,13 @@ payment-confirmation-amount-year = { $intervalCount ->
...
+new-user-email =
+  .placeholder = foxy@mozilla.com
+  .label = Enter your email
+new-user-confirm-email =
+  .label = Confirm your email
+new-user-subscribe-product-updates = I'd like to receive product updates from Firefox
```
chenba, line 373: Same with the brand name 'Firefox' here.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { NewUserEmailForm } from './index';
+
+storiesOf('components/NewUserEmailForm', module).add('default', () => (
+  <div style={{ display: 'flex' }}>
+    <NewUserEmailForm
+      onBlurInitial={(evt) => {
+        console.log('onBlurInitial: ', evt);
+      }}
+      onBlurConfirm={(evt) => {
+        console.log('onBlurInitial: ', evt);
```
chenba, line 12: Should this be `onBlurConfirm` instead?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
chenba: I think the general plan update the state of some parent component (which will in turn update `PaymentForm`) based on the state/events of those inputs.  I suppose they don't even need be in a Form, since we are not submitting those values through that form.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const validateEmail: OnValidateFunction = (
+  value,
+  focused,
+  _props,
+  getString
+) => {
+  // TODO: port error validation from content server https://mozilla-hub.atlassian.net/browse/FXA-3665
+  let valid = true;
+  if (value !== null && !value) {
```
chenba, line 112: Is this testing for an empty string?  Can we use `=== ''` in that case?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { NewUserEmailForm } from './index';
+
+describe('NewUserEmailForm test with l10n', () => {
+  it('renders as expected', () => {
+    const { queryByTestId } = render(
+      <NewUserEmailForm onBlurInitial={() => {}} onBlurConfirm={() => {}} />
+    );
+    const form = queryByTestId('newuseremailform');
+    expect(form).toBeInTheDocument();
+  });
```
biancadanforth, line 14: May be helpful to add a test that asserts that when the first/second fields are blurred in the form, that `onBlurInitial` (ideally renamed to avoid confusion) and `onBlurConfirm` are called.

biancadanforth (dismissed):
Okay here's the second half of my review. The form looks good in Storybook. There are a few things worth addressing though -- in particular:
* Ensure `onBlur` handlers are called as expected
* Use brand variables in localization as Barry noted
* Match the styles on the "Sign in" link to the design
* Some smaller items like maybe renaming some variables/data-testids/props for clarity.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+    <Form
+      data-testid="newuseremailform"
+      validator={validator}
+      className="new-user-email-form"
+    >
+      {/*
+        - sign in link needs to have link and metrics
+        - hook up product updates checkbox to basket(this will probably happen on the back end)
+        - if we detect that the email is already associated with an account(tested onBlur), we will show a tooltip "Sign in with your Firefox Account"
```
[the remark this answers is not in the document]
meandavejustice: There are already issues covering these, I just forgot to remove these comments

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
meandavejustice: This is a placeholder for the email validation calls

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: This is what was being used in `OnValidateFunction` in paymentform component. I assumed that it was checking for null here for a reason.

meandavejustice:
@biancadanforth I've addressed many of the comments here. I'm going to leave the onBlur methods alone for this patch as they are likely to change when implementing the account validation calls. With that in mind I don't want to write tests around it if it's going to change within the next week or so.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          Already have a Firefox account? <a>Sign in</a>
+        </p>
+      </Localized>
+      <Localized id="new-user-email" attrs={{ placeholder: true, label: true }}>
+        <Input
+          type="text"
+          name="new-user-email"
+          label="Enter your email"
+          data-testid="new-user-email"
+          placeholder="foxy@mozilla.com"
```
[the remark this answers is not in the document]
zachblk: Hahaha...I like the placeholder as well! Unfortunately, we don't have any official placeholder text. I pulled "johnsmith@gmail.com" from an older design Jessica gave me. I can check with Chris to see if it is any he recommends. But, foxy@mozilla.com does seem appropriate!

biancadanforth:
Hey Dave; unfortunately I didn't have time to check this out this afternoon, but Barry has kindly agreed to take another look while I am away for the next few days.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
chenba: > This is a placeholder for the email validation calls

I think Bianca meant that when she focus on the input and the blur, she did not see any console.log output.  Even if they are placeholders, they should execute?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
chenba: I see.  It just read so strangely.  It's checking for not null and not falsey, but null itself is falsey anyway, so why a separate check for not null.  Confusing.

chenba:
> 
> 
> ...I'm going to leave the onBlur methods alone for this patch as they are likely to change when implementing the account validation calls. With that in mind I don't want to write tests around it if it's going to change within the next week or so.

Regardless what teh callback functions are going to do, we still need to ensure that the functions are called on blur.  I don't think Bianca was asking for tests on the behavior of the actual callback functions.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import { NewUserEmailForm } from './index';
+
+describe('NewUserEmailForm test with l10n', () => {
```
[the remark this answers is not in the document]
chenba: @meandavejustice What about this comment from Bianca?  I don't think it's been addressed?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+  it('renders as expected', () => {
+    const { queryByTestId } = render(<NewUserEmailForm />);
+    const form = queryByTestId('new-user-email-form');
+    expect(form).toBeInTheDocument();
```
chenba, line 11: I think we should verify more than just the form element here.  We expect the form to have two email inputs, so we could check for that at least?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+import {
+  State as ValidatorState,
+  useValidatorState,
+  MiddlewareReducer as ValidatorMiddlewareReducer,
+} from '../../lib/validator';
+
+import './index.scss';
+
+export type NewUserEmailFormProps = {
+  getString?: Function;
```
chenba, line 16: Preferably we give this function a more specific type.  It sounds like it returns a string?  But just `Function` doesn't give the type system that.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+        - if we detect that the email is already associated with an account(tested onBlur), we will show a tooltip "Sign in with your Firefox Account"
+        */}
+      <Localized id="new-user-sign-in-link">
+        <p className="sign-in-copy">
+          Already have a Firefox account? <a>Sign in</a>
+        </p>
+      </Localized>
+      <Localized id="new-user-email" attrs={{ placeholder: true, label: true }}>
+        <Input
+          type="text"
```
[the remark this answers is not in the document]
chenba: @meandavejustice ☝️ ?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+          spellCheck={false}
+          onValidate={(value, focused, props) =>
+            validateEmail(value, focused, props, getString)
+          }
+        />
+      </Localized>
+
+      <Localized id="new-user-confirm-email" attrs={{ label: true }}>
+        <Input
+          type="text"
```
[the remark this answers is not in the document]
chenba: @meandavejustice 👆 ?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  );
+};
+
+const validateEmail: OnValidateFunction = (
+  value,
+  focused,
+  _props,
+  getString
+) => {
+  // TODO: port error validation from content server https://mozilla-hub.atlassian.net/browse/FXA-3665
```
chenba, line 99: Does this mean I should not review this function or where it is called?

chenba (changes requested):
Please see inline comments.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: This method was copied over from the `OnValidateFunction` in paymentform component. It can be reviewed and improved upon

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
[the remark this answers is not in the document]
meandavejustice: I looked around at our other tests, and we aren't doing l10n assertions in other components, so I've added more element tests here and adjusted the title of this test.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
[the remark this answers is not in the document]
chenba: @meandavejustice if there's reason you want to keep these as "text", please leave a comment.  Having them as "email" is consistent with the content-server's sign up form.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <Localized id="new-user-confirm-email" attrs={{ label: true }}>
+        <Input
+          type="text"
+          name="new-user-confirm-email"
+          label="Confirm your email"
+          data-testid="new-user-confirm-email"
+          required
+          spellCheck={false}
+          onValidate={(value, focused, props) =>
+            validateEmail(value, focused, props, getString)
```
chenba, line 66: Since we need to ensure the value of second email input is identical to the value of the first email input, we cannot have the same validation function for both inputs.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+) => {
+  // TODO: port error validation from content server https://mozilla-hub.atlassian.net/browse/FXA-3665
+  let valid = true;
+  if (value !== null && !value) {
+    valid = false;
+  }
+  const errorMsg = getString
+    ? /* istanbul ignore next - not testing l10n here */
+      getString('new-user-email-validate')
+    : 'Sign in with your Firefox Account';
```
chenba, line 107: This error message doesn't make sense to me given teh context how these inputs are used.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  _props,
+  getString
+) => {
+  // TODO: port error validation from content server https://mozilla-hub.atlassian.net/browse/FXA-3665
+  let valid = true;
+  if (value !== null && !value) {
+    valid = false;
+  }
+  const errorMsg = getString
+    ? /* istanbul ignore next - not testing l10n here */
```
chenba, line 105: It is testable in that we can check whether `getString` was called with the correct argument.

chenba (changes requested):
Please see inline comments.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: Thanks for pointing this out, it probably simplifies things to only match the two for the second input

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: This is the message that will show if an account exists, we will be showing a different error message if it fails other validation.

Correct messaging will be tweaked once email validation is actually added in https://mozilla-hub.atlassian.net/browse/FXA-3665

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: Copied from payment form, https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/components/PaymentForm/index.tsx#L356, Is this something you would like a test written for in order to have this approved or are you just making a note?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <Localized id="new-user-email" attrs={{ placeholder: true, label: true }}>
+        <Input
+          type="text"
+          name="new-user-email"
+          label="Enter your email"
+          data-testid="new-user-email"
+          placeholder="foxy@mozilla.com"
+          required
+          spellCheck={false}
+          onValidate={(value, focused, props) => {
```
chenba, line 53: Neither onValidate callbacks appear to be tested?

chenba (changes requested):
I can't find any tests for the validation callbacks.  https://github.com/mozilla/fxa/pull/9708/files#r668899299 and https://github.com/mozilla/fxa/pull/9708/files#r668899385 are still outstanding.

meandavejustice:
@chenba Fixed the inputs to use `type="email"`. As for the validation callbacks, validation is to be added in a follow up issue https://mozilla-hub.atlassian.net/browse/FXA-3665 and the account existence check is to be added in https://mozilla-hub.atlassian.net/browse/FXA-3663.

This pull request is for https://github.com/mozilla/fxa/issues/9706, which is the Base component, Storybook, and sanity tests.

chenba:
> 
> 
> > I can't find any tests for the validation callbacks. https://github.com/mozilla/fxa/pull/9708/files#r668899299 and https://github.com/mozilla/fxa/pull/9708/files#r668899385 are still outstanding.
> 
> @chenba Fixed the inputs to use `type="email"`. 

I don't know if I'm losing it or just looking at the wrong thing but I don't see that.  There's also no commit pushed between my comment and yours.

> As for the validation callbacks, validation is to be added in a follow up issue

Then why not just include two no-op callbacks?  Some form of validation is being added in this PR.  Is the next PR going to completely replaced the functions in this PR?

meandavejustice:
> I don't know if I'm losing it or just looking at the wrong thing but I don't see that. There's also no commit pushed between my comment and yours.

You are not losing it, I had forgotten to push that change. It is pushed now 🤦 

> Then why not just include two no-op callbacks? Some form of validation is being added in this PR. Is the next PR going to completely replace the functions in this PR?

They are essentially no-ops, I will be adding the rest of the functionality for email validation and checking if there is an existing account. The one line that verifies that the values match between the two fields will probably stay. But I may be changing the way the state is parsed in the form parsing issue follow up. These methods will have full test coverage after I can add the validation and ensure that the current way it is factored works in the way we want.

meandavejustice:
> Then why not just include two no-op callbacks? Some form of validation is being added in this PR. Is the next PR going to completely replace the functions in this PR?

I'm not sure I understand what the concern here is, is it the coverage results?

chenba:
> 
> 
> > Then why not just include two no-op callbacks? Some form of validation is being added in this PR. Is the next PR going to completely replace the functions in this PR?
> 
> I'm not sure I understand what the concern here is, is it the coverage results?

If they are essentially no-ops, then why not just use no-ops?  The concern is adding untested code.  I might not be the follow-up PR's reviewer, and that reviewer might not be aware of what went on in this PR and think that code is already tested.  (This is less of a problem since our team is currently very small.)  Also, it doesn't seem to make sense to split the code and the tests into two PRs; I don't see any advantages to that.  But, if the code is going to be replaced, then why expend the effort when we could just use no-ops?  (And why merge code into main that we don't intend to keep in the first place?)

meandavejustice:
@chenba The intention was never to split up tests and code into separate prs, it was just splitting up features. I'm happy to just change these functions into no-ops for the time being so we can get this merged.

chenba (approved):
✉️

PR #9741, opened 2021-06-24
fix(stripe): Enable CHF currency in paypal

vbudhram:
## This pull request

- Adds CHF to hardcoded paypal currency
- Targetted as point release for 208. Needs to be merged and tag before we tag 209 train

## Issue that this pull request solves

Closes: TBD

## Checklist

- [x] My commit is GPG signed.
- [ ] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [ ] I have verified that my changes render correctly in RTL (if appropriate).

PR #9743, opened 2021-06-24
fix(auth-server): allow swiss francs

bbangert:
Because:

* We want to launch in Switzerland.

This commit:

* Adds Swiss francs as a valid currency.

## Checklist


dannycoates (approved):

vbudhram:
Fixed in https://github.com/mozilla/fxa/pull/9741

PR #8920, opened 2021-06-21
task(fxa-auth-server): Improve validation for product metadata

meandavejustice:
- fixes #9192

TODO:
- [x] Ensure all DEV products and plans pass validation
- [x] Ensure all STAGE products and plans pass validation
- [ ] Ensure all PROD products and plans pass validation

PR #9744, opened 2021-06-24
chore(adr): add ADR for expanding subplat metadata

chenba:
Because:
 * we want to use more metadata than the limit Stripe placed on its
   metadata

This commit:
 * add an ADR to document our choices and decisions

### DIFF
--- a/docs/adr/0026-expand-subscription-platform-metadata.md
+++ b/docs/adr/0026-expand-subscription-platform-metadata.md
@@ -1,0 +1,124 @@
+# Expand Subscription Platform Metadata with Google Cloud Firestore
+
+- Status: accepted
+- Deciders: Ben Bangert, Barry Chen, Bianca Danforth, Danny Coates, Dave Justice, Les Orchard, Vijay Budhram
+- Date: 2021-07-23
+
+## Context and Problem Statement
+
+The Subscription Platform uses [Stripe Metadata][stripe-metadata] for numerous
+purposes, including product checkout content, legal document download URLs,
+and product support form options. The metadata can be updated by product
+managers through the Stripe dashboard, enabling product managers to manage
+certain aspects of their subscription products without involving Subscription
+Platform engineers. However, Stripe metadata has a limit on the number of
+values that we will reach in the near future. In order to continue to use
+metadata in the same fashion in the Subscription Platform, we need to overcome
+the limit placed by Stripe.
+
+[stripe-metadata]: https://stripe.com/docs/api/metadata
+
+## Decision Drivers
+
+- A higher storage limit than Stripe Metadata
+- A UI for the managers of products on the Subscription Platform to edit the
+  metadata
+- Access management for the UI
+- Updated data accessible in their respective environments without deployment
+
+## Considered Options
+
+- Google Cloud Firestore
+- Kinto
+- MySQL
+- Redis
+- Configuration Files
+
+## Decision Outcome
+
+Chosen option: Google Cloud Firestore, because
+
+- Data management UI and user access control provided through the Google Cloud
+  Platform (GCP) console.
+- It does not require deployments for data updates.
+- It is not novel tech; FxA already uses it elsewhere.
+
+## Pros and Cons of the Options
+
+### Google Cloud Firestore
+
+[Google Cloud Firestore][cloud-firestore] is a hosted NoSQL document based database. It has access management and data management UI through the GCP console. We use it for the Event Broker.
+
+- Pros
+  - Already in the FxA monorepo; not a novel tech or a new dependency.
+  - Includes access management and data editing UI in the form of the GCP
+    console.
+  - Does not require deployment for data edits.
+  - Built-in support for access control and validation with ["Security Rules"][firestore-security-rules].
+- Cons
+  - Product manager will need to learn to use the GCP console.
+  - Allowing updates on live data.
+  - Management and deployment of Security Rules.
+
+[cloud-firestore]: https://firebase.google.com/docs/firestore
+[firestore-security-rules]: https://firebase.google.com/docs/rules
+
+### Kinto
+
+[Kinto][kinto] is a JSON storage service that is used elsewhere in Mozilla.
+
+- Pros
+
+  - Already deployed and in use at Mozilla for Remote Settings in Firefox.
+  - Data can be updated without FxA deployments.
+  - Multiple collections per Kinto instance - e.g. dev / stage / prod product sets.
+  - Record creation UI and data validation via JSON schema - maybe friendlier than raw JSON.
+  - Implements a review process for pending changes with approve / reject flow.
+  - Maintains a changelog of updates and who was involved.
+  - Has ACLs and groups for read / write / review access.
+
+- Cons
+  - Backed by PostgreSQL, while FxA uses MySQL.
+  - Not as turn-key to set up as something like Firestore.
+  - Yet another service and a new dependency for FxA.
+
+[kinto]: https://docs.kinto-storage.org/
+
+### MySQL
+
+MySQL is the relational database system FxA uses.
+
+- Pros
+  - Existing tech that's immediately available.
+  - Engineers' familiarity with it.
+  - Does not require deployment for data updates.
+- Cons
+  - No data management UI for product managers.
+
+### Redis
+
+Redis is an in-memory data store that is part of the current FxA stack.
+
+- Pros
+
+  - Already in tech stack and familiarity.
+  - Can be updated without a deploy.
+
+- Cons
+  - No UI for management.
+  - Might need Ops to manually update entires.
+  - Would need to be populated/seeded with entries on startup.
+
+### Configuration Files
+
+We require product managers to add and edit configuration files and submit Github pull requests (PR) to get the updates into production.
+
+- Pros
+  - No dependency on an external service.
+  - Product managers are free to use any file editing tool.
+  - Edits are reviewed.
+  - Access control through PR approvals.
+- Cons
+  - Updates require the entire CI and deployment process.
+  - Product managers need to learn to use Github at a minimum.
+  - Product managers need to learn file format of the configuration files.


docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
+# Expand Subscription Platform Metadata with Google Cloud Firestore
+
+- Status: proposed
+- Deciders: Ben Bangert, Barry Chen, Bianca Danforth
```
chenba, line 4: @bbangert @biancadanforth included you since I've chatted with you about this already.

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Google Cloud Firestore
+
+[Google Cloud Firestore][cloud-firestore] is a hosted NoSQL document based database. It has access management and data management UI through the GCP console. We use it for the Event Broker.
+
+- Pros
+  - Already in the FxA monorepo; not a novel tech or a new dependency.
+  - Includes access management and data editing UI in the form of the GCP
+    console.
+  - Does not require deployment for data edits.
```
meandavejustice, line 50: We can also add validation which should appear in the console to product managers, https://firebase.google.com/docs/rules/data-validation

This is something we are lacking with our new metadata validation

meandavejustice (approved):

meandavejustice:
I wasn't a part of this discussion, but It sounds like the decision has been made. The only thing I would like in addition to the pros listed above, would be a platform that allows for data validation on input. Our current validation implementation #8920 is lacking in that it does not let the product managers know that the data is invalid directly and has us building out other scripts in order to notify them and the engineers of the failure. It looks like firestore may support something like this, I linked above..

chenba:
>It sounds like the decision has been made.

@meandavejustice I've been confused by the ADR process in the past as well, but the decision has not made.  An ADR PR is really a request for comments.  (The recent email templates ADR is actually a great example; the chosen option was not the one in the original draft.)

lmorchard:
> I've been confused by the ADR process in the past as well, but the decision has not made. 

Pardon the drive-by comment, but: Yeah, the ADR decision isn't really properly made until the PR gets merged. Folks can request changes & block merge in the meantime.

docs/adr/0026-expand-subscription-platform-metadata.md
bbangert: Wow, I hadn't seen that feature, that's slick!

bbangert (approved):
LGTM.

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: @meandavejustice Could you add an issue in the FXA-3655 epic for this?  Thanks!

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+- A UI for the managers of products on the Subscription Platform to edit the
+  metadata
+- Access management for the UI
+- Updated data accessible in their respective environments without deployment
+
+## Considered Options
+
+- Google Cloud Firestore
+- MySQL
+- Configuration Files
```
vbudhram, line 29: Another possible option is Redis (probably bad one at that)

Pros
- Already in tech stack and familiarity
- Can be updated without a deploy

Cons
- No UI for management
- Might need ops to manually update entires
- Would need to be populated/seeded with entries on startup

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+### MySQL
+
+MySQL is the relational database system FxA uses.
+
+- Pros
+  - Existing tech that's immediately available.
+  - Engineers' familiarity with it.
+  - Does not require deployment for data updates.
+- Cons
+  - No data management UI for product managers.
```
vbudhram, line 67: We might need to do a migration to add these entries?

vbudhram (approved):
@chenba Makes sense to me 👍🏽

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: Thanks, I'll include it in the final edit.

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: I was thinking that once we have a table, inserts and updates wouldn't require migrations.  Granted, we currently do not have a process of directly inserting and updating records on prod.

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+
+[Google Cloud Firestore][cloud-firestore] is a hosted NoSQL document based database. It has access management and data management UI through the GCP console. We use it for the Event Broker.
+
+- Pros
+  - Already in the FxA monorepo; not a novel tech or a new dependency.
+  - Includes access management and data editing UI in the form of the GCP
+    console.
+  - Does not require deployment for data edits.
+- Cons
+  - Access control might not be granular enough.
```
biancadanforth, line 52: What do you mean here? Can you give an example or how it may not be granular enough?

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+  - Product manager will need to learn to use the GCP console.
+  - Allowing updates on live data.
```
biancadanforth, line 54: This is a risk people often overlook; I'm glad you mentioned it. I guess we also have this risk with Stripe metadata currently, but at least as noted above, we could validate the data before it is saved.

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+We require product managers to add and edit configuration files and submit Github pull requests (PR) to get the updates into production.
+
+- Pros
+  - No dependency on an external service.
+  - Product managers are free to use any file editing tool.
+  - Edits are reviewed.
+  - Access control through PR approvals.
+- Cons
+  - Updates require the entire CI and deployment process.
+  - Product managers need to learn to use Github at a minimum.
```
biancadanforth, line 80: I have heard there is a lot of historical evidence to suggest it's not an easy sell.

biancadanforth (commented):
Thank you Barry for thinking about this. I added some non-blocking questions mostly along the line of seeking some more context from the perspective of someone who maybe isn't on our team.

Overall, I think this is the right technology choice -- especially concerning data validation before saving edits and not requiring deploys.

docs/adr/0026-expand-subscription-platform-metadata.md
bbangert: The ACL capability in Firestore actually looks more advanced and granular than MySQL: https://firebase.google.com/docs/rules

docs/adr/0026-expand-subscription-platform-metadata.md
bbangert: We can still request updates go through our team, so we don't have to allow live data changes for PMs.

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+- Deciders: Ben Bangert, Barry Chen, Bianca Danforth
+- Date: 2021-06-24
+
+## Context and Problem Statement
+
+The Subscription Platform uses [Stripe Metadata][stripe-metadata] for numerous
+purposes. However, Stripe metadata has a limit on the number of values that we
+will reach in the near future. In order to continue to use metadata in the
+same fashion in the Subscription Platform, we need to overcome the limit placed
+by Stripe.
```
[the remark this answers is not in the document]
chenba: > 
> 
> > The Subscription Platform uses [Stripe Metadata][stripe-metadata] for numerous
> > purposes.
> 
> Can you list a few purposes here to be more explicit? Who updates this metadata today?

Sure.

> 
> > However, Stripe metadata has a limit on the number of values that we will reach in the near future.
> 
> Maybe we can link to the Stripe docs here that discuss that limit?

Yep.


> How close are we to this limit today? What is the timeline in which we expect to exceed that limit? More broadly, are there other non-idealities to relying on Stripe metadata than this limit (e.g. metadata validation)?
> 

The number of metadata keys used is product and pricing plan dependent.  I could give some kind of napkin sketch math here if you think that'd be helpful.

As for the timeline, that's purely a Product thing; I don't truly know.  Would that affect the decision?

Validation can be a whole separate topic altogether, regardless of storage, transport, access control, etc.  We can discuss more in a future jira (the one I requested Dave to file).  It's not a basis for the decision here is what I think I'm trying to say.

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: I could remove that bullet point if the Firestore ACL rules meet our needs.

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: Even if someone from our team handles the editing, there still no review and approval process.

docs/adr/0026-expand-subscription-platform-metadata.md
lmorchard: Not that folks are necessarily looking for yet another option, but how about...

[Kinto](https://docs.kinto-storage.org/en/latest/)

Pros

* Already deployed and in use at Mozilla for Remote Settings in Firefox
* Data can be updated without FxA deployments
* Multiple collections per Kinto instance - e.g. dev / stage / prod product sets
* Record creation UI via JSON schema - maybe friendlier than raw JSON
* Implements a review process for changes with approve / reject flow
* Maintains a changelog of updates and who was involved
* Has ACLs and groups for read / write / review access

Cons

* Backed by postgres, while FxA uses mysql
* Not as turn-key to set up as something like Firestore
* Yet another service & a new dependency for FxA

Trying to think of other cons

docs/adr/0026-expand-subscription-platform-metadata.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Decision Drivers
+
+- A higher storage limit than Stripe Metadata
+- A UI for the managers of products on the Subscription Platform to edit the
+  metadata
+- Access management for the UI
+- Updated data accessible in their respective environments without deployment
+
+## Considered Options
```
lmorchard, line 25: Meant to hang this comment over here - not that folks are necessarily looking for yet another option, but how about...

[Kinto](https://docs.kinto-storage.org/en/latest/)

Pros

* Already deployed and in use at Mozilla for Remote Settings in Firefox
* Data can be updated without FxA deployments
* Multiple collections per Kinto instance - e.g. dev / stage / prod product sets
* Record creation UI and data validation via JSON schema - maybe friendlier than raw JSON
* Implements a review process for pending changes with approve / reject flow
* Maintains a changelog of updates and who was involved
* Has ACLs and groups for read / write / review access

Cons

* Backed by postgres, while FxA uses mysql
* Not as turn-key to set up as something like Firestore
* Yet another service & a new dependency for FxA

Trying to think of other cons

lmorchard (commented):
Just had a notion that [Kinto could be interesting](https://docs.kinto-storage.org/en/latest/overview.html#key-features) to consider here, after working with it for a bit over on Experimenter

docs/adr/0026-expand-subscription-platform-metadata.md
chenba: Thanks, Les!  With those cons, I can imagine jbuck being in favor of Firestore.  The pros sound great though.

I'll incorporate it into the final edit. 👍

docs/adr/0026-expand-subscription-platform-metadata.md
bbangert: If we wanted a full review/approval process, ops can manage Firestore documents. They are now doing this for the firestore documents backing event-broker. So there's a formalized process if we want one.

docs/adr/0026-expand-subscription-platform-metadata.md
bbangert: I think if anything, the ACL capability in Firestore should be a Pro. It's fully programmatic as well, allowing functions to be used to gauge level of access on a per-field and per document basis for each action.

docs/adr/0026-expand-subscription-platform-metadata.md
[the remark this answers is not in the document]
biancadanforth: > The number of metadata keys used is product and pricing plan dependent. I could give some kind of napkin sketch math here if you think that'd be helpful.
>
> As for the timeline, that's purely a Product thing; I don't truly know. Would that affect the decision?

Hm... no I don't think that'd be helpful here. I'm just curious about these things from an urgency perspective, but you're right that this sort of information is not the most relevant to include here.

> Validation can be a whole separate topic altogether, regardless of storage, transport, access control, etc. We can discuss more in a future jira (the one I requested Dave to file). It's not a basis for the decision here is what I think I'm trying to say.

Why wouldn't validation (before saving) be a basis for a decision here -- particularly if we are planning to go with an approach that does not provide a direct review/approval process? That seems like a "pro" to me, even if it isn't part of an initial implementation... Is it because pretty much all the approaches we have thus far considered have some mechanism to validate the data (unlike the Stripe metadata approach we are using currently)?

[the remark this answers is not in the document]
chenba: > Why wouldn't validation (before saving) be a basis for a decision here -- particularly if we are planning to go with an approach that does not provide a direct review/approval process? ...

The validation rules are from us, regardless the form they take, the way they are managed, or the context in which they are executed.  The feature that Firestore provide seems great (at least from some quick reading of their docs); I'll add it as a pro.  But generally, none of the options would prevent us from performing validation... or, what you said, and said it better.

dannycoates (approved):
SGTM

docs/adr/0026-expand-subscription-platform-metadata.md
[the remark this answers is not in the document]
chenba: > 
> 
> > Maybe we can link to the Stripe docs here that discuss that limit?
> 
> Yep.
> 
I just realized that the doc linked in the ADR discusses that limit.  There isn't another page to link.

chenba:
I pushed a final draft with everyone included in the deciders list and incorporated the review feedback.

Thank you all!

PR #9766, opened 2021-06-26
fix(auth): don't resend session otp codes when totp is enabled

dannycoates:
## Because

- As is the case with `/recovery_email/resend_code` this is unnecessary

### DIFF
--- a/packages/fxa-auth-server/lib/routes/session.js
+++ b/packages/fxa-auth-server/lib/routes/session.js
@@ -398,6 +398,20 @@ module.exports = function (
 
         request.emitMetricsEvent('session.resend_code');
 
+        // Check to see if this account has a verified TOTP token. If so, then it should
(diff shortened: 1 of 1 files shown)


rfk (approved):

packages/fxa-auth-server/lib/routes/session.js
```diff
@@ -398,6 +398,20 @@ module.exports = function (
 
         request.emitMetricsEvent('session.resend_code');
 
+        // Check to see if this account has a verified TOTP token. If so, then it should
+        // not be allowed to bypass TOTP requirement by sending a sign-in confirmation email.
+        try {
```
vbudhram, line 403: We could probably add similar check here https://github.com/mozilla/fxa/blob/56108382558d7d0c3ebc9e4b8341d83032d1bb3e/packages/fxa-auth-server/lib/routes/session.js#L445

PR #9775, opened 2021-06-28
feat(auth): don't delete an unverified account with an active subscri…

### LINKED ISSUE #8745: Prevent unverified deletes if the user has a stripe account.
Currently the /create account route will delete unverified accounts if the same email exists. We should prevent this behavior if there's an unverified account with a stripe account linked to ensure we don't delete paying customers that haven't finished setting up their account.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-3601)
┆Issue Number: FXA-3601

biancadanforth:
…ption

Because:

* We currently delete unverified accounts in a number of places, but with [FXA-3486](https://mozilla-hub.atlassian.net/browse/FXA-3486), it will be possible for an unverified account to have an active subscription.

This commit:

* Ensures we don't delete unverified accounts that have active subscriptions.

Closes #8745

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -816,6 +816,20 @@ export class StripeHelper {
   }
 
   /**
+   * Returns true if the FxA account with uid has an active subscription.
+   */
+  async hasActiveSubscription(uid: string): Promise<Boolean> {
+    const customer = await this.fetchCustomer(uid, ['subscriptions']);
+    if (!customer) {
+      return false;
+    }
+    const subscription = customer.subscriptions?.data.find((sub) =>
+      ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status)
+    );
+    return !!subscription;
+  }
+
+  /**
    * Returns whether or not the customer has any active subscriptions that
    * have an open invoice (payment has not been processed).
    *
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -81,10 +81,15 @@ export class AccountHandler {
   private async deleteAccountIfUnverified(request: AuthRequest, email: string) {
     try {
       const secondaryEmailRecord = await this.db.getSecondaryEmail(email);
-      // Currently, users can not create an account from a verified
+      // Currently, users cannot create an account from a verified
       // secondary email address
       if (secondaryEmailRecord.isPrimary) {
-        if (secondaryEmailRecord.isVerified) {
+        if (
+          secondaryEmailRecord.isVerified ||
+          (await this.stripeHelper.hasActiveSubscription(
+            secondaryEmailRecord.uid
+          ))
+        ) {
           throw error.accountExists(secondaryEmailRecord.email);
         }
         request.app.accountRecreated = true;
@@ -1771,4 +1776,7 @@ export const accountRoutes = (
   return routes;
 };
 
-module.exports = accountRoutes;
+module.exports = {
+  accountRoutes,
+  AccountHandler,
+};
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -152,7 +152,10 @@ module.exports = (
             // of accounts with invalid email addresses. These
             // can never be verified, so the best we can do is
             // to delete them so the browser will stop polling.
-            if (!validators.isValidEmailAddress(sessionToken.email)) {
+            if (
+              !validators.isValidEmailAddress(sessionToken.email) &&
+              !(await stripeHelper.hasActiveSubscription(sessionToken.uid))
+            ) {
               await db.deleteAccount(sessionToken);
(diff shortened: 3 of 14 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -815,6 +815,20 @@ export class StripeHelper {
...
+  async hasActiveSubscription(uid: string): Promise<Boolean> {
+    const customer = await this.fetchCustomer(uid, ['subscriptions']);
+    if (!customer) {
+      return false;
+    }
+    const subscription = customer.subscriptions?.data.find((sub) =>
+      ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status)
+    );
+    return !!subscription;
+  }
```
biancadanforth, line 830: This new helper looks suspiciously similar to `hasSubscriptionRequiringPaymentMethod`, except:
* It takes in a `uid`, since we don't know if there is a Stripe customer yet for this user.
* It doesn't care if the subscription is marked `cancel_at_period_end`.

If there's a way to make this DRYer, please let me know.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -152,7 +152,10 @@ module.exports = (
             // of accounts with invalid email addresses. These
             // can never be verified, so the best we can do is
             // to delete them so the browser will stop polling.
-            if (!validators.isValidEmailAddress(sessionToken.email)) {
+            if (
+              !validators.isValidEmailAddress(sessionToken.email) &&
+              !(await stripeHelper.hasActiveSubscription(sessionToken.uid))
+            ) {
```
biancadanforth, line 158: I wasn't sure if we wanted to add the check here, but I did for thoroughness.

IIUC from the comment above, we would not in practice expect someone with an invalid email (and therefore older account) to also have an active subscription, since  prior to [the epic for this patch](https://mozilla-hub.atlassian.net/browse/FXA-3486), only verified and authenticated users could create a subscription.

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -696,14 +699,23 @@ module.exports = (
...
                 await db.deleteAccount(secondaryEmailRecord);
                 log.info('accountDeleted.unverifiedSecondaryEmail', {
                   ...secondaryEmailRecord,
                 });
                 return;
-              } else {
+              } else if (!emailNewlyCreated) {
                 throw error.unverifiedPrimaryEmailNewlyCreated();
+              } else {
+                throw error.unverifiedPrimaryEmailHasActiveSubscription();
```
biancadanforth, line 718: There can be two reasons now why an unverified account is not deleted, so I created a new `AppError` here.

packages/fxa-auth-server/test/local/ip_profiling.js
```diff
@@ -67,8 +71,9 @@ function runTest(route, request, assertions) {
   }).then(assertions);
 }
 
-describe('IP Profiling', () => {
+describe('IP Profiling', function () {
   let route, accountRoutes, mockDB, mockMailer, mockRequest;
+  this.timeout(30000);
```
biancadanforth, line 76: When I update the `proxyquire`, I got a timeout error, where ` const { accountRoutes } = proxyquire('../../lib/routes/account', requireMocks || {});` takes more than 5 seconds to execute.

Why is it so slow? I added a timeout of 30s inside the test to allow it to pass.

biancadanforth (commented):
I have added questions and comments in line where possible, with a couple of additional questions:
1. I had thought `scripts/verification-reminders.js` was also a place where we deleted unverified accounts, but I didn't find where that happens. Can you point me to the right place?
2. I did not add a check to the `/account/destroy` handler even though it deletes an account, as this looks to be used when someone is voluntarily deleting their account. I just want to confirm that I should not modify this route?

packages/fxa-auth-server/lib/routes/emails.js
vbudhram: Can't hurt to add this here.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -81,10 +81,15 @@ export class AccountHandler {
   private async deleteAccountIfUnverified(request: AuthRequest, email: string) {
     try {
       const secondaryEmailRecord = await this.db.getSecondaryEmail(email);
-      // Currently, users can not create an account from a verified
+      // Currently, users cannot create an account from a verified
```
vbudhram, line 84: I wish we could pull this out into a shared function `deleteAccountOrEmailIfUnverified`, but it seems there are sutble differences in the routes :/

vbudhram (approved):
@biancadanforth LGTM 👍🏽

> I had thought scripts/verification-reminders.js was also a place where we deleted unverified accounts

I think this is correct, I might have led you wrong because the script itself is sending a reminder to help an unverified user get verified 😅

> I did not add a check to the /account/destroy handler even though it deletes an account, as this looks to be used when someone is voluntarily deleting their account

Yea I believe this is correct. On the settings page, the user has has a checkbox next to the service before they delete the account.

Issue #9255, opened 2021-06-21
[emails M1] Convert `cadReminderFirst` to new stack


data-sync-user:
The first template conversion is here, aww yeah!

Take `cadReminderFirst.html` and use the new stack/system set up thus far to convert the template which will also include adding a shared `cadReminder` component.  Convert tests and add states to the state previewing solution. Make sure it builds properly and that the new email is sent locally (by turning on the feature flag it sits behind).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3508)

data-sync-user:
➤ Lauren Zugai commented:

This ticket is basically going to be a catch-all for “convert anything remaining” to reach parity for cadReminderFirst.

  * Convert CadReminderFirst to MJML/EJS/CSS solution
  * Ensure significant states are represented in Storybook (it’s OK to move states from CadReminder into CadReminderFirst, I don’t know that we really need all states for partials)
  * Add doc property in Storybook, I’ll get back to you about what string to use here
  * Turn on feature flagging for yourself locally, and make sure you can receive the email locally in maildev (how to trigger this locally will be more clear once the string for doc is provided - may also need Vijay’s help here)
  * Copy tests over
  * Create plaintext files/solution, which I believe we mentioned in the ADR. This might be the most time consuming part of this ticket.

Yogita98:
There are a few really small things which are supposed to be done as part of this ticket:
1. Things pending in [this comment](https://github.com/mozilla/fxa/pull/9779#issuecomment-881476444)
2. Padding of the button is supposed to be 12px or 16px or it has to be kept as is. Further context over [here](https://github.com/mozilla/fxa/pull/9779#discussion_r671335101)

LZoog:
@Yogita98 asked our designer (Janice Cramer) and she said go with `16px`. :)

LZoog:
Let's also be sure to address:
* [This](https://github.com/mozilla/fxa/pull/10005#discussion_r682041571) now famously annoying CI problem (well, it's now famously annoying between you, @jodyheavener, and I 😀)
* A member of the l10n team [got back to me here](https://github.com/mozilla/fxa/pull/10005/files#r681843257) and provided the following screenshot which shows what RTL languages there are, and also gave [this link](https://pontoon.mozilla.org/projects/firefox-accounts/) which shows what languages FxA supports. Let's double check that array all/only the RTL languages listed there.
[...]

PR #8765, opened 2021-05-31
docs(adrs): proposal for "Upgradation of Auth server email templating toolset"

Yogita98:
## Because

- We want to revamp `fxa-auth-server` emails as the current templating stack doesn't allow us to include logic and the styling is done using inline styles.

## This pull request

- This PR contains an architectural document enlisting the alternatives that we could use for our templating stack to eliminate the cumbersome nature of emails.

## Issue that this pull request solves

Closes: #8600 

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [ ] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [ ] I have verified that my changes render correctly in RTL (if appropriate).

## Screenshots (Optional)

Please attach the screenshots of the changes made in case of change in user interface.

## Other information (Optional)

Any other information that is important to this pull request.

PR #9778, opened 2021-06-30
feat(fxa-auth-server): Setup storybook to preview auth-sever emails

Yogita98:
## Because

- We want a state previewing solution for our fxa-auth-sever emails

## This pull request

- Aims to include packages and files that could be used to preview the email templates with Storybook

Closes: #9244 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/.storybook/main.js
+++ b/packages/fxa-auth-server/.storybook/main.js
@@ -2,4 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export * as cadReminder from './cadReminder';
+module.exports = {
+  stories: [
+    '../lib/senders/emails/**/*.stories.tsx',
+    '../lib/senders/emails/**/*.stories.ts',
+  ],
+  addons: ['@storybook/addon-docs', '@storybook/addon-controls'],
+};
--- a/packages/fxa-auth-server/.storybook/preview.js
+++ b/packages/fxa-auth-server/.storybook/preview.js
@@ -2,4 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export * as cadReminder from './cadReminder';
+export const parameters = {
+  actions: { argTypesRegex: '^on[A-Z].*' },
+};
--- a/packages/fxa-auth-server/lib/senders/emails/renderer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/renderer.ts
@@ -1,14 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 import ejs = require('ejs');
 import mjml2html = require('mjml');
 import * as templates from './templates';
+const config = require('../../../config').getProperties();
 
 export const context = {
   buttonText: 'Sync device',
   onDesktopOrTabletDevice: true,
+  baseURL: config.contentServer.url,
   anotherDeviceURL:
-    'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
-  iosURL: '',
-  androidURL: '',
+    '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  iosURL: config.smtp.iosUrl,
+  androidURL: config.smtp.androidUrl,
 };
 
 export function render(
--- a/packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export interface StorybookEmailArgs {
+  template: string;
+  description?: string;
+  variables: Record<string, any>;
+}
+
+export const storybookEmail = ({
+  template,
+  description,
+  variables,
+}: StorybookEmailArgs): HTMLDivElement => {
+  const container = document.createElement('div');
+  container.innerHTML = 'Loading email...';
+  const emailDescriptionDiv = document.createElement('div');
+  if (description) {
+    emailDescriptionDiv.innerHTML = `
+        <p>${description}</p>
+        <hr />`;
+  }
+
+  renderUsingMJML({ template, variables })
+    .then((result) => {
+      container.innerHTML = emailDescriptionDiv.innerHTML + result;
+    })
+    .catch((error: Error) => {
+      container.innerHTML = `Error loading email: ${error.message}`;
+    });
+
+  return container;
+};
+
+async function renderUsingMJML({
+  template,
+  apiUrl = 'http://localhost:8192',
+  variables,
+}: {
+  template: string;
+  apiUrl?: string;
+  variables: Record<string, any>;
+}): Promise<string> {
+  const response = await fetch(apiUrl, {
+    method: 'POST',
+    body: JSON.stringify({
+      template,
+      ...variables,
+    }),
+  });
+  const result = await response.text();
+  if (response.status !== 200) {
+    throw new Error(result);
+  }
+  return result;
+}
+
+export default storybookEmail;
--- a/packages/fxa-auth-server/lib/senders/emails/styles.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/styles.ts
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 export const style = `
 .header-text div {
   font-family: sans-serif !important;
--- a/packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.ts
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Story, Meta } from '@storybook/html';
+import storybookEmail, { StorybookEmailArgs } from '../../storybook-email';
+
+export default {
+  title: 'Emails/cadReminder',
+} as Meta;
+
+const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
+
+const defaultVariables = {
+  buttonText: 'Sync device',
+  anotherDeviceURL:
+    '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  iosURL:
+    'https://accounts-static.cdn.mozilla.net/product-icons/apple-app-store.png',
+  androidURL:
+    'https://accounts-static.cdn.mozilla.net/product-icons/google-play.png',
+  onDesktopOrTabletDevice: true,
+};
+
+const commonPropsWithOverrides = (
+  overrides: Partial<typeof defaultVariables> = {}
+) =>
+  Object.assign({
+    template: 'cadReminder',
+    doc: 'The Connect Another Device Reminder is sent when [TODO: documentation].',
+    variables: {
+      ...defaultVariables,
+      ...overrides,
+    },
+  });
+
+export const CadReminderDesktopTablet = Template.bind({});
+CadReminderDesktopTablet.args = commonPropsWithOverrides({
+  onDesktopOrTabletDevice: true,
+});
+CadReminderDesktopTablet.storyName = 'User is on desktop or tablet device';
+
+export const CadReminderMobile = Template.bind({});
+CadReminderMobile.args = commonPropsWithOverrides({
+  onDesktopOrTabletDevice: false,
+});
+CadReminderMobile.storyName = 'User is on mobile device';
--- a/packages/fxa-auth-server/package.json
+++ b/packages/fxa-auth-server/package.json
@@ -28,8 +28,11 @@
     "clean-up-old-ci-stripe-customers": "ts-node ./scripts/clean-up-old-ci-stripe-customers.js",
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
     "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
-    "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
-    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
+    "write-templates": "NODE_ENV=dev ts-node lib/senders/emails/render-to-disk.ts",
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maildev.ts",
+    "start-mjml": "NODE_ENV=dev ts-node bin/mjml-server.ts",
+    "storybook": "start-storybook -p 6010 --no-version-updates",
+    "storybook-mjml": "npm-run-all -p start-mjml storybook"
   },
   "repository": {
     "type": "git",
@@ -107,6 +110,11 @@
     "web-push": "3.4.4"
   },
   "devDependencies": {
+    "@babel/core": "^7.14.6",
+    "@storybook/addon-controls": "^6.3.4",
+    "@storybook/addon-docs": "^6.3.4",
+    "@storybook/html": "^6.3.4",
+    "@types/babel__core": "7.1.14",
     "@types/chai": "^4.2.18",
     "@types/convict": "^5.2.2",
     "@types/hapi__hapi": "^19.0.3",
@@ -126,8 +134,10 @@
     "@types/safe-regex": "1.1.2",
     "@types/uuid": "^8.3.0",
     "@types/verror": "^1.10.4",
+    "@types/webpack": "5.28.0",
     "acorn": "^8.0.1",
     "audit-filter": "^0.5.0",
+    "babel-loader": "^8.2.2",
     "binary-split": "1.0.5",
     "chai": "^4.3.4",
     "eslint": "^7.29.0",
@@ -153,6 +163,7 @@
     "mocha-junit-reporter": "^2.0.0",
     "moment": "^2.29.1",
     "nock": "^13.1.1",
+    "npm-run-all": "^4.1.5",
     "nyc": "^15.1.0",
     "pm2": "^5.1.0",
     "prettier": "^2.3.1",
@@ -164,6 +175,8 @@
     "through": "2.3.8",
     "ts-node": "^10.0.0",
     "typescript": "^4.2.4",
+    "webpack": "^4.43.0",
+    "webpack-watch-files-plugin": "^1.1.0",
     "ws": "^7.5.0"
   }
 }
--- a/packages/fxa-auth-server/bin/mjml-server.ts
+++ b/packages/fxa-auth-server/bin/mjml-server.ts
@@ -1,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import http from 'http';
+import stream from 'stream';
+import net from 'net';
+import { render } from '../lib/senders/emails/renderer';
+
+const baseURL = require('../config').get('contentServer.url');
+
+const { PORT = 8192, HOST = '0.0.0.0' } = process.env;
+
+const MJML_OPTIONS = {};
+
+console.log(`Starting MJML API server at ${HOST}:${PORT}`);
+
+const server = http.createServer(handleRequest);
+server.on('clientError', handleError);
+server.listen({ port: PORT, host: HOST });
+
+async function handleRequest(
+  req: http.IncomingMessage,
+  res: http.ServerResponse
+) {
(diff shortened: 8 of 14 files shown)


packages/fxa-auth-server/.storybook/webpack.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require('fxa-react/configs/storybooks').customizeWebpackConfig;
```
Yogita98, line 1: I've not added any new config for webpack and just reusing the one from [fxa-settings](https://github.com/mozilla/fxa/blob/main/packages/fxa-react/configs/storybooks.js#L15) but we can definitely add more options if we need

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
+import http from 'http';
+import stream from 'stream';
+import net from 'net';
+import { render } from '../renderer';
+
+const { PORT = 8192, HOST = '0.0.0.0' } = process.env;
```
Yogita98, line 6: I took the host and port from Les's repo so let me know if 8192 is already occupied or we're good to use it for development purpose

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
+import http from 'http';
```
Yogita98, line 1: Initially I tried to setup a storybook build of the templates without making use of a mjml server, but since MJML doesn't render in browser context I used Les's [starter repo](https://github.com/lmorchard/mjml-in-storybook) as a base to deploy storybook in fxa-auth-server

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
jodyheavener: It doesn't look like 8192 is occupied, but we _do_ have a pattern of using the 6000 range for Storybook. So this could be 6010.

packages/fxa-auth-server/lib/senders/emails/EmailStory/index.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+);
+
+export const CadReminder = Template.bind({});
+CadReminder.args = {
+  template: 'cadReminder',
+  variables: {
+    buttonText: 'Sync device',
+    onDesktopOrTabletDevice: true,
+    anotherDeviceURL:
+      'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
```
jodyheavener, line 20: I think we'll need to conditionally attach the base URL so it's not hard-coded to localhost

packages/fxa-auth-server/package.json
```diff
@@ -95,18 +98,28 @@
     "poolee": "^1.0.1",
     "punycode.js": "2.1.0",
     "qrcode": "^1.4.4",
+    "react": "^16.13.1",
```
jodyheavener, line 101: I think we should consider not using React here. As far as I can tell it's only being used in fxa-auth-server to facilitate mounting the Storybook stories (a component that uses a hook to load the contents). I think we could accomplish something similar with a plain JS fetch call and insert the returned HTML into the plain HTML story ([@storybook/html](https://www.npmjs.com/package/@storybook/html)). This would allow us to remove/no longer add `react`, `react-dom`, `@storybook/react`, `@types/react`, and `@types/react-dom` as dependencies. @LZoog thoughts?

packages/fxa-auth-server/.storybook/preview.js
```diff
@@ -0,0 +1,9 @@
+export const parameters = {
+  actions: { argTypesRegex: '^on[A-Z].*' },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
```
jodyheavener, line 6: It doesn't look like we have the [Controls](https://storybook.js.org/docs/riot/essentials/controls) pane enabled, so these dynamic control matchers might not be neccesary (maybe they were default generated?)

jodyheavener (commented):
I'm excited how this looks! Left you a few thoughts.

LZoog (commented):
This is looking good, have a few comments I think you can address but I don't think it's far from merging.

I think we could restructure this slightly and go ahead and show the two significant states for `cadReminder` - I played a bit with this a bit here which may help you for future things: https://github.com/Yogita98/fxa/compare/state-previewing-solution...LZoog:state-previewing-solution

You can either make the changes on your machine or `git cherry-pick` my commit. This will also give you an idea of what I meant by "documentation above the component" with that `description` bit, though obviously it's unpolished and I already know @jodyheavener will point out how horrendous that `<hr>` looks. 😆

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
...
+async function handleRequest(
+  req: http.IncomingMessage,
+  res: http.ServerResponse
+) {
+  try {
+    const body = await readStream(req);
+    const data = JSON.parse(body.trim());
+    const { template: templateName, ...variables } = data;
+
+    console.log(`REQUEST ${templateName}`);
```
LZoog, line 25: I don't mind leaving this in if it helps for now but we may want to remove it when it becomes unhelpful (which is probably soon if not now, since this is hit for every story).

packages/fxa-auth-server/package.json
```diff
@@ -29,7 +29,10 @@
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
     "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
     "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
-    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts",
```
LZoog, line 32: Ohh, unrelated to your PR but could be done in a quick follow up - we're missing the "l" in `preview-in-maildev.ts` filename.

packages/fxa-auth-server/package.json
```diff
@@ -29,7 +29,10 @@
...
+    "start-mjml": "ts-node lib/senders/emails/bin/mjml-sever.ts",
+    "storybook": "start-storybook -p 6007 --no-version-updates",
+    "storybook-preview": "npm-run-all -p start-mjml storybook"
```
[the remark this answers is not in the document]
jodyheavener: Hmm, but it's also running the mjml server, whereas the other package.json `storybook` commands _only_ run storybook - I would think maybe we could rename this to `storybook-mjml` or something and keep the `storybook` command how it is since it's technically consistent with the other ones?

packages/fxa-auth-server/package.json
jodyheavener: Yeah I think it would be safe to move all the new react deps to devDependencies. I'm still a little unsure about it since it's a **lot** of additional dependencies to look after (keep updated) for a single Storybook component, where React is not being used for anything else in fxa-auth-server.

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
jodyheavener: Hence my confusion! 😛

packages/fxa-auth-server/package.json
LZoog: I think we can at least uninstall `@storybook/addon-links` and `@storybook/addon-actions`, I don't think we'll need either of those for this.

I definitely get your point though. @Yogita98 what would be the level of effort to see what this looks like with `@storybook/html`, would it take more than half a day? I don't think you should spend more than that here since this is working, but I'm curious what it would look like in comparison to the modified `.stories` file which [shows two different states for `cadReminder`](https://github.com/Yogita98/fxa/compare/state-previewing-solution...LZoog:state-previewing-solution#diff-721989f388164241c41b2b3cc4cd438e1aa0c53f2fd2866bbbd655ad9285739cR35-R45). 🤔

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function readStream(readable: stream): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const parts: any[] = [];
+    readable.on('data', (chunk) => parts.push(chunk));
+    readable.on('end', () => resolve(Buffer.concat(parts).toString('utf8')));
+    readable.on('error', (err) => reject(err));
+  });
+}
```
jodyheavener, line 56: Heads up that this file is called `mjml-sever`, missing an `r` in `server`.

packages/fxa-auth-server/package.json
jodyheavener: @Yogita98 I got a version working on my end using only @storybook/html - Gist is here: https://gist.github.com/jodyheavener/46dec26d85f1f567742f17d9730c12b0 (note that I renamed `RenderedEmail.tsx` to `storybook-email.ts`)

LZoog: _Sweet_, after seeing that, @Yogita98 I think we should move forward with the non-React version to eliminate the extra dependencies. You can take what Jody did there and combine it with my directory structure updates and showing two stories for `cadReminder` ([handy link](https://github.com/Yogita98/fxa/compare/state-previewing-solution...LZoog:state-previewing-solution)).

packages/fxa-auth-server/lib/senders/emails/styles.ts
```diff
@@ -1,53 +1,53 @@
 export const style = `
 .header-text div {
   font-family: sans-serif !important;
-  font-size: 21px !important;
-  line-height: 29px !important;
+  font-size: 20px !important;
```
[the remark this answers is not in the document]
Yogita98: Yeah right. Let's keep all the styles related work in the other branch

packages/fxa-auth-server/package.json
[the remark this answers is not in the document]
Yogita98: I'll go with `storybook-mjml` for now

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  title: 'Emails/Cad Reminder',
+} as Meta;
+
+const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
+
+export const DesktopDevice = Template.bind({});
+DesktopDevice.args = {
+  template: 'cadReminder',
+  description:
+    'The Connect Another Device Reminder is sent when [TODO: documentation].',
```
Yogita98, line 14: @LZoog Do have we have this `description` documented somewhere?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import { Story, Meta } from '@storybook/html';
+import storybookEmail, { StorybookEmailArgs } from '../../storybook-email';
+// const configURL = require('../../../../../config').get('contentServer.url')
```
Yogita98, line 3: @jody When I'm importing the config file in Storybook, I'm running into errors because that config file imports modules like `fs` which cannot be rendered in browser so keeping it like this for now. Will get back to this tomorrow!

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
LZoog: We don't, which is exactly why having some documentation here will be great. This isn't something I think you should or can fill out though since digging around for the context of each one would take time away from converting them over, so instead, we can file a follow up ticket for someone on the FxA team to help out with a bit later (unless there's some time left after template conversions).

Though, I'm kind of thinking to make the purpose of this more obvious, we could rename it from `description` to `doc`. `doc` is a property [we use to document config setting descriptions](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/config/index.ts#L27) with a package called convict so it'd lead to a tiny bit of consistency/clarity. Maybe when you work on finishing the `cadReminder` conversion in #9255 we can at least edit this one to show what it'll look like to have one in place as an example, leave the rest as blank to do later.

packages/fxa-auth-server/lib/senders/emails/styles.ts
```diff
@@ -27,7 +27,7 @@ span {
   line-height: 18px !important;
 }
 .app-badges img {
-    width: 152px !important;
+  width: 152px !important;
```
LZoog, line 30: While this fix looks correct, you can probably take out this file from the commit (same `git checkout` command as before) since you'll still conflict with yourself in the CSS PR.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export default {
+  title: 'Emails/Cad Reminder',
```
LZoog, line 6: @Yogita98 I think we should follow the name of the actual template here to make scanning for them easier (calling it `cadReminder`).

Also, I think we can DRY up the `args` needed with a helper [like commonPropsWithOverrides](https://github.com/Yogita98/fxa/commit/fa78b5afdf1788555d17e713ec69763442078dfb#diff-721989f388164241c41b2b3cc4cd438e1aa0c53f2fd2866bbbd655ad9285739cR21-R45). It's not too bad without it in `cadReminder` but if there's a template that has 3, 4, 5 stories, it'll be especially handy to change only the properties that differ from story to story and will clean things up.

Plus, adding types for the expected arguments in each template as they're worked on may set us up for some of [Les' wishful thinking](https://github.com/mozilla/fxa/pull/8765#issuecomment-862794694) that he added in the ADR. 🙂

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const DesktopOrTabletDevice = Template.bind({});
+DesktopOrTabletDevice.args = {
+  template: 'cadReminder',
+  description:
+    'The Connect Another Device Reminder is sent when [TODO: documentation].',
+  variables: {
+    buttonText: 'Sync device',
+    onDesktopOrTabletDevice: true,
+    anotherDeviceURL: `http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device`,
```
LZoog, line 19: @jodyheavener I know we called out not wanting to use `http://localhost:3030` here, but since this PR is dev-only Storybook, maybe we can add that in a follow up Storybook deployment ticket.

packages/fxa-auth-server/.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  stories: ['../lib/senders/emails/**/*.stories.tsx'],
+  addons: ['@storybook/addon-docs', '@storybook/addon-controls'],
```
Yogita98, line 3: Added these addons so as to implement things stated in [this comment](https://github.com/mozilla/fxa/pull/8765#issuecomment-862794694)

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -7,8 +7,10 @@ export const context = {
   onDesktopOrTabletDevice: true,
   anotherDeviceURL:
     'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
-  iosURL: '',
-  androidURL: '',
+  iosURL:
+    'https://accounts-static.cdn.mozilla.net/product-icons/apple-app-store.png',
```
Yogita98, line 11: Added these URLs in here since [this PR](https://github.com/mozilla/fxa/pull/9788) got merged

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+import storybookEmail, { StorybookEmailArgs } from '../../storybook-email';
+// const configURL = require('../../../../../config').get('contentServer.url')
+
+export default {
+  title: 'Emails/cadReminder',
+} as Meta;
+
+const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
+
+const commonPropsWithOverrides = (
```
Yogita98, line 11: Added these overriding props but I'm not sure how much we can benefit from this and storybook/controls in general as we might not be able to have a separate control for the objects as we have for other argtypes. Linking [this PR](https://github.com/storybookjs/storybook/pull/12685) for context.

LZoog: Huh, I didn't dive much into Les' comment / how those auto generated controls work in Storybook 6.0+ when I suggested this function. I haven't played much with Storybook controls, but they do seem pretty neat. If it's not too much reworking, maybe we should move forward with that approach. 🙂

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.tsx
LZoog: I think you'll want to require this in `mjml-server` and then send it down to the templates as a new variable.

BTW, I think this file can also be a `.ts` file rather than a `.tsx` file.

packages/fxa-auth-server/lib/senders/emails/renderer.ts
jodyheavener: Wait - are `iosURL` and `androidURL` meant to be URLs to the badge images, or URLs that a user can click on from the image badge anchor? Because I still see the old URL's being used in `packages/fxa-auth-server/lib/senders/emails/partials/appBadges.ts`, and on main I [can see](https://github.com/Yogita98/fxa/blob/main/packages/fxa-auth-server/lib/senders/templates/partials/appBadges.html#L3:L4) that those variables are being used as anchors (not image URLs).

packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
```diff
@@ -0,0 +1,51 @@
...
+  variables: Record<string, any>;
+}): Promise<string> {
+  const resp = await fetch(apiUrl, {
+    method: 'POST',
+    body: JSON.stringify({
+      template,
+      ...variables,
+    }),
+  });
+  return await resp.text();
```
jodyheavener, line 48: Currently, if an error occurs on the rendering/server side the fetch operation will still complete, but the returned body will be an error message. In order to use the try/catch for this scenario in the `storybookEmail` function you could replace this line with:

```js
const result = await resp.text();
if (resp.status !== 200) {
  throw new Error(result);
}
return result;
```

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+      buttonText: 'Sync device',
+      anotherDeviceURL:
+        '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+      iosURL:
+        'https://accounts-static.cdn.mozilla.net/product-icons/apple-app-store.png',
+      androidURL:
+        'https://accounts-static.cdn.mozilla.net/product-icons/google-play.png',
+      ...overrides,
+    },
+  });
```
jodyheavener, line 26: I think this function needs just a little more finessing.

To start, it looks like `storybookEmail` is looking for a `description` property, but `doc` is provided instead (are these the same thing?), and if you were to manually set `description` on the individual story, like:

```js
CadReminderMobile.args = commonPropsWithOverrides({
  description: 'This is a description',
});
```

This doesn't work because the argument object `overrides` is being merged into the child `variables` object, not the top-level object.

Additionally, `commonPropsWithOverrides` is typed to receive an object with an interface of `Partial<StorybookEmailArgs>`, so effectively:

```ts
Partial<{
  template: string;
  description?: string;
  variables: Record<string, any>;
}>
```

But since the `overrides` object is being applied directly to the child `variables` object it would seem we never need to override `template` or `description`. Therefor I think instead of typing it with `Partial<StorybookEmailArgs>` I would suggest you type it as `Record<string, string>`. Or if you want to get even more specific with the types you could do [this](https://gist.github.com/jodyheavener/fab1a8072c29b2cc45953c15d27330f7), which extracts types from the default values. Either of these solutions will also allow you to remove the `/*@ts-ignore */` bits.

jodyheavener (changes requested):
Just a few more suggestions and then I'll be happy to get this merged!

packages/fxa-auth-server/lib/senders/emails/renderer.ts
Yogita98: Ohhh my bad. Previously, I had left them empty so do you want me to keep as it was before or include the url from [config](https://github.com/mozilla/fxa/blob/60282403d3aecb61d95fe98e92ca8c959c0990d5/packages/fxa-auth-server/config/index.ts#L284)?

jodyheavener:
Oops! Before we forget, all files that support it need an MPL 2.0 header at the top of the file. You don't need to (or can't) add them to JSON, HTML, CSS files, but they should be added to TS, JS, and SCSS files. Refer to other JS or TS files for an example.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.stories.ts
Yogita98: I assumed that we do not want to override the template name and description of a story and was thinking to disable editable controls as well for them, something like [this](https://storybook.js.org/docs/riot/essentials/controls#disable-controls-for-specific-properties) maybe?

jodyheavener: Yup! We can still update the types for this

LZoog:
There's another conflict with `yarn.lock`, but after Jody gives the thumbs up on his requested changes, LGTM! :)

Yogita98:
Yeah everytime I was resolving the conflicts and pushing back, there was a change in dependency happening. The conflicts are resolved for now but I'm having a hard time rebasing the commits as there is a merge commit involved. Not sure but is it okay for a merge commit to be in there?

Issue #9247, opened 2021-06-21
[emails M1] Set up CSS solution


data-sync-user:
FXA-3503 / #8600 decided what CSS solution we should use.

We need to enable this CSS solution in the templating system added in FXA-3504 / #8601. Be sure that the styles load properly in whatever previewing solution was chosen and that when building the template, the styles are inline.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3506)

LZoog:
In #8765, we chose to use MJML and the kind of styling that it provides.

Some questions I have, after #9704 is merged:
* Can we use `.css` files instead of TS files that use stringified CSS?
* If we can use `.css`, can we actually use `.scss` files? SCSS would be awesome for SCSS variables.
* We may want to standardize some values for consistency (margin/padding, font-sizes) and maybe look at the FxA Settings design guide, [see discussion](https://github.com/mozilla/fxa/pull/9704#discussion_r657562561). @Yogita98 I'd say leave this 3rd point for when @jodyheavener and I return. If we can use SCSS, that'd make this easier to do.

PR #9779, opened 2021-07-01
feat(fxa-auth-server): Setup CSS solution for emails

Yogita98:
## Because

- We are using inline styles as part of our current email templating stack

## This pull request

- Aims to include a styling solution that could replace inline styles for auth-server emails

Closes: #9247 

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
Yogita98, line 1: There are a few files which I'm committing again in this PR though they were already pushed as part of the [storybook PR ](https://github.com/mozilla/fxa/pull/9778) since I wasn't sure how we'd test this branch without merging the other branch first.

packages/fxa-auth-server/lib/senders/emails/templates/welcome.ts
```diff
@@ -0,0 +1,16 @@
+export const render = () => {
+  return `
+    <mjml>
+      <mj-body>
+      <mj-include path="./lib/senders/emails/css/style.css" type="css" css-inline="inline" />
```
Yogita98, line 5: It was mentioned in the [issue](https://github.com/mozilla/fxa/issues/9247#issuecomment-868866887) if we can make use of a `.css` file instead of `typescript` file for styling so this change addresses that concern

packages/fxa-auth-server/package.json
```diff
@@ -29,7 +29,11 @@
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
     "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
     "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
-    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts",
+    "scss": "node-sass lib/senders/emails/scss -o lib/senders/emails/css",
```
Yogita98, line 33: We also wanted to verify if we can use `scss` for our templates so yes we probably can by setting up a build pipeline similar to this

packages/fxa-auth-server/lib/senders/emails/scss/style.scss
```diff
@@ -0,0 +1,30 @@
+$large-font-size: 20px;
+$small-font-size: 12px;
```
LZoog, line 2: -dances- I know I'm the one that suggested these names for our variables, but looking at the [Tailwind class names for font-size](https://tailwindcss.com/docs/font-size) and noting those classes also set `line-height`, I wonder if we could just copy the Tailwind naming conventions and their associated styles (using `px` instead of `rem` since it's safer for emails, meaning you'll multiply those rem values by `16`). For `text`, we could use [`@mixin`](https://sass-lang.com/documentation/at-rules/mixin)s.

It'd look something like this:

```
@mixin text-xs {
  font-size: 12px !important;
  line-height: 16px !important;
}
```

And then when you need to use it:
```
.primary-text div {
  @include text-small;
  font-family: sans-serif !important;
  font-weight: normal !important;
}
```

We could potentially use a Sass [placeholder](https://sass-lang.com/documentation/style-rules/placeholder-selectors) instead (`%text-sm` syntax), though mixins give us the option to add arguments later, like if we _really_ want a custom line-height for some reason so I feel it's more appropriate here. 

We can also add other patterns we identify to something like this - maybe `font-family: sans-serif` + `font-weight: normal` comes up enough that we want a custom mixin for that? Not sure, but we can probably set some `margin`, `padding`, and `color`s with regular SCSS variables.

Checking the [Tailwind padding docs](https://tailwindcss.com/docs/padding), `4px` would be `p-1`. So we could set:

```
$p-1: 4px !important;
```
Then:
```
td {
  padding: $p-1;
}
```

I don't think we need to go overboard, but I'll be super happy with how much we can DRY it up if we can force some consistency between `font-size`, `line-height`, `padding`, `margin`, and color properties. 👍 

I'm curious to hear @jodyheavener's thoughts on this.

packages/fxa-auth-server/lib/senders/emails/scss/style.scss
```diff
@@ -0,0 +1,30 @@
...
+  line-height: 20px !important;
+  font-weight: normal !important;
+}
+.fxa-logo td {
+  height: 60px !important;
+  width: 60px !important;
+}
+.sync-logo td {
+  height: 176px !important;
+  width: 240px !important;
```
LZoog, line 29: One more note on these... do we know if we _have_ to use `!important` on all of these lines? I'm sort of not sure when we need them because of MJML or because of email clients. Maybe it's safer to just slap an `!important` on whatever had one previously, but dang, that hurts my soul because I guarantee you we've just randomly stuck that on a lot of inline styles without knowing if they're needed.

packages/fxa-auth-server/lib/senders/emails/css/style.css
```diff
@@ -0,0 +1,32 @@
...
+  text-align: center !important;
+}
+
+td {
+  padding: 4px !important;
+}
+
+span {
+  color: #0a84ff;
+}
```
jodyheavener, line 15: We'll need to be careful with writing too-broad selectors in global files. For example, here, is every `span` across every template going to have this color? Or above, is every `td` going to have a 4px padding? We should consider using classes more generously so we're not painting ourselves into a corner with far-reaching selectors.

packages/fxa-auth-server/lib/senders/emails/partials/button.ts
```diff
@@ -1,9 +1,8 @@
...
-    <mj-column>
-      <mj-button>
-        <%= buttonText %>
-      </mj-button>
-    </mj-column>
-  </mj-section>
+<mj-include path="./lib/senders/emails/css/button.css" type="css" css-inline="inline" />
+    <mj-section>
+      <mj-column>
+        <mj-button css-class="p-button"><a href="www.google.com"><%= buttonText %></a></mj-button>
```
jodyheavener, line 5: Is this intentionally `www.google.com`? Just temporary for development?

packages/fxa-auth-server/lib/senders/emails/scss/appBadges.scss
```diff
@@ -0,0 +1,12 @@
+.app-badges img {
```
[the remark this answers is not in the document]
jodyheavener: > Kind of an aside note, but I think we can have a soft rule that partials don't necessarily need Storybook stories (anyone feel free to disagree with me) or tests if they're covered by templates.

I generally agree, but it would be nice to have stories for far-reaching partials, like buttons, and avatar, footer, etc - to allow for easier template building down the road. But I also don't want overkill here.

packages/fxa-auth-server/lib/senders/emails/scss/style.scss
jodyheavener: Yes I agree with this. Our emails have a fairly limited set of styles that often come grouped together, so this is a good approach IMO. My only hesitation would be if we eventually start more liberally grouping common styles together, and then eventually the mixins get just as bogged down as element-specific rulesets. But I'm less concerned with email styles.

packages/fxa-auth-server/package.json
jodyheavener: If we keep this command it should be renamed to something more specific, like `email-scss`

packages/fxa-auth-server/lib/senders/emails/scss/style.scss
jodyheavener: Yeah.. I'd love to avoid !important unless absolutely necessary. Might need to spend some time going through each one to test if it really is needed.

jodyheavener (commented):
Good stuff. Looking forward to seeing that other branch merged so the diff here is smaller. Nice job!

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
+import http from 'http';
```
vbudhram, line 1: Kinda of strange to have this bin file here.  Could this move to the fxa-auth-server bin directory?

packages/fxa-auth-server/lib/senders/emails/bin/mjml-sever.ts
```diff
@@ -0,0 +1,56 @@
...
+  res: http.ServerResponse
+) {
+  try {
+    const body = await readStream(req);
+    const data = JSON.parse(body.trim());
+    const { template: templateName, ...variables } = data;
+
+    console.log(`REQUEST ${templateName}`);
+
+    const result = render(templateName, variables);
```
vbudhram, line 27: We should be able to use this render function when building out the feature flagging parts 👍🏽

packages/fxa-auth-server/package.json
```diff
@@ -29,7 +29,11 @@
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
     "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
     "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
-    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts",
```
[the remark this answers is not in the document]
Yogita98: Ohh my bad. Thanks for pointing it out!

packages/fxa-auth-server/lib/senders/emails/partials/button.ts
Yogita98: Yeah just wanted to test if it's working in a email client, but I'll remove it in the next commit

packages/fxa-auth-server/lib/senders/emails/css/style.css
Yogita98: While converting `mjml` into `html`, MJML adds its own styles inlined with the html elements so if we're adding our custom styles on top of that and they are overlapping, they will get ignored unless we add `!important` and then our custom styles get prioritized. While, I do agree that `span` might not be used in all templates, `td` is something that needs to be taken care of everywhere. I observed that majority of mjml elements get narrowed down to `td` and each one of it has its own padding of around 10-25 px which looks weird in our emails. If adding `!important` to each and every style is not looking neat then we can also use mjml attributes and do something like [this](https://mjml.io/try-it-live/5JTenq09jXb).

packages/fxa-auth-server/lib/senders/emails/scss/appBadges.scss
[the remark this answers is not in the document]
Yogita98: Yeah the directory structure that you set out to me definitely looks way better than what I have currently. But I'm confused as to how we will be compiling those `.scss` files to `css` files. What I mean to say is currently `appBadges.scss` gets compiled to `appBadges.css` but with the structure that you've mentioned above, all the scss files have similar name i.e `index.scss` so are we planning on compiling all the scss files into one file say `main.css` or do you want me to create subdirectories inside the css folder. Something like this maybe:

```
templates/
  cadReminder/
    index.ts // hopefully we can change this to `.mjml` or `.ejs`?
    index.stories.tsx
    index.scss
    index.test.ts // can be done in that "convert cadReminder" issue

partials/
  cadReminder/
    index.ts 
    index.stories.tsx
    index.scss

css/
  cadReminder/
    index.css
  appBadges/
    index.css
```

packages/fxa-auth-server/package.json
```diff
@@ -85,6 +89,7 @@
     "mozlog": "^3.0.2",
     "mysql": "^2.18.1",
     "mysql-patcher": "0.7.0",
+    "node-sass": "^6.0.1",
```
Yogita98, line 92: Oops looks like node-sass is [deprecated](https://github.com/sass/node-sass#node-sass). Just a quick heads up that I'll be replacing it with [`sass`](https://www.npmjs.com/package/sass) package

_dev/.stylelintrc
```diff
@@ -22,6 +22,7 @@
           "each",
           "include",
           "mixin",
+          "use",
```
Yogita98, line 25: I had to add this rule in here as we are using `@use` in our scss files and the pre-commit hooks were failing

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -0,0 +1,11 @@
+$p-1: 4px !important;
```
Yogita98, line 1: Added a global stylesheet as well as separate stylesheets for templates and partials

packages/fxa-auth-server/lib/senders/emails/sass-compile-files.ts
```diff
@@ -0,0 +1,58 @@
+import sass from 'sass';
```
Yogita98, line 1: In order to setup the directory structure that you laid out to me, I am using this script so as compile and generate stylesheets from multiple directories

packages/fxa-auth-server/package.json
```diff
@@ -29,7 +29,11 @@
     "paypal-processor": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/paypal-processor.ts",
     "subscription-reminders": "CONFIG_FILES='config/secrets.json' ts-node ./scripts/subscription-reminders.ts",
     "write-templates": "ts-node lib/senders/emails/render-to-disk.ts",
-    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts"
+    "maildev-preview": "ts-node lib/senders/emails/preview-in-maidev.ts",
+    "email-scss": "node-sass lib/senders/emails/scss -o lib/senders/emails/css",
+    "start-mjml": "ts-node bin/mjml-sever.ts",
+    "storybook": "ts-node lib/senders/emails/sass-compile-files.ts && start-storybook -p 6007 --no-version-updates",
+    "storybook-preview": "npm-run-all -p start-mjml storybook"
```
Yogita98, line 36: Renaming the `mjml-server.ts` file and changing the storybook port is being taken care of in the other PR

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.ts
```diff
@@ -1,16 +1,14 @@
-import { style } from '../styles';
```
Yogita98, line 1: Removes the `styles.ts` file since it's no longer being used and also refactored the stylesheets wherever possible either by drying them up or removing `!important` syntax

jodyheavener (commented):
This generally is looking great to me. I'll want to take another final look once we merge the Storybook PR to see the true diff, but I think this is great.

packages/fxa-auth-server/lib/senders/emails/partials/button/index.scss
```diff
@@ -0,0 +1,28 @@
+@use '../../global.scss';
+
+@mixin text-lg {
+  font-size: 18px !important;
+  line-height: 28px !important;
+}
+$blue-background: #0a84ff !important;
```
LZoog, line 7: Let's call this `$blue-400` since it's a color and not necessarily a background color. `400` is the closest value this hex code is to what's in the `fxa-settings` design guide (500 looks much darker to me).

Let's also use the `blue-400` Tailwind color from our design guide which is `#0090ED`.

Our design team will be fine with small tweaks like this for consistency across FxA. 🙂

packages/fxa-auth-server/lib/senders/emails/partials/button/index.scss
```diff
@@ -0,0 +1,28 @@
...
+  @include text-lg;
+}
+
+.p-button td {
+  background: $blue-background;
+}
+
+.p-button a {
+  text-decoration: none;
+  color: #fff;
```
LZoog, line 27: Let's go ahead and set a variable to all colors. This one is fine as `$white` per the design guide.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.scss
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.header-text div {
+  @include text-xl;
+  font-family: global.$font-sans;
+  text-align: center !important;
+}
+
+span {
+  color: #0a84ff;
```
LZoog, line 19: Hmm, since we're using that color here too, let's pull that `$blue-400` into globals and then use it in both files.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.scss
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.sync-logo td {
+  width: 240px !important;
+}
+
+.primary-text div {
+  @include text-sm;
+  font-family: global.$font-sans;
+  text-align: center !important;
```
LZoog, line 29: If we end up pulling `.primary-text div` into globals to use later, we can probably leave the `text-align` center out of the globals file but leave it here.

LZoog (commented):
I'm really happy with where this is going! As Jody said I'd like to look it over once the Storybook PR is landed and I have a couple of small suggested tweaks but this is looking great.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.scss
Yogita98: Yeah we can continue drying up the stylesheets by extending the globals as and when new templates are added

packages/fxa-auth-server/lib/senders/emails/partials/index.ts
```diff
@@ -0,0 +1,2 @@
+export * as cadReminder from './appBadges';
+export * as welcome from './button';
```
jodyheavener, line 2: I'm confused by this export name `welcome`, could you elaborate?

packages/fxa-auth-server/package.json
```diff
@@ -106,14 +107,18 @@
     "stripe": "^8.157.0",
     "superagent": "^6.1.0",
     "typedi": "^0.8.0",
+    "typescript": "^4.2.4",
```
jodyheavener, line 110: Typescript is already a devDependency, I don't think we need it here

packages/fxa-auth-server/package.json
```diff
@@ -106,14 +107,18 @@
...
     "verror": "^1.10.0",
     "web-push": "3.4.4"
   },
   "devDependencies": {
     "@babel/core": "^7.14.6",
+    "@storybook/addon-actions": "^6.3.0",
     "@storybook/addon-controls": "^6.3.4",
     "@storybook/addon-docs": "^6.3.4",
+    "@storybook/addon-links": "^6.3.0",
+    "@storybook/addons": "^6.3.0",
```
jodyheavener, line 121: Do we still need these 3 storybook dependencies? I think we could remove them

packages/fxa-auth-server/lib/senders/emails/partials/index.ts
Yogita98: Ahh yes I messed up while rebasing the branch. I'll fix it and update you soon

Yogita98:
@jody @LZoog I think I have been able to rebase the branch properly. Let me know your thoughts on this

jody:
> @jody @LZoog I think I have been able to rebase the branch properly. Let me know your thoughts on this

Forwarding to @jodyheavener

packages/fxa-auth-server/lib/senders/emails/sass-compile-files.ts
```diff
@@ -0,0 +1,60 @@
...
+async function compileSass(dir, subdir) {
+  const styleResult = await sassRenderPromise({
+    file: dir,
+    outFile: subdir,
+  });
+
+  await writeFilePromise(subdir, styleResult.css, 'utf8');
+}
+
+/* @ts-ignore */
```
jodyheavener, line 35: How come ts-ignore? I don't see the warning

packages/fxa-auth-server/lib/senders/emails/sass-compile-files.ts
```diff
@@ -0,0 +1,60 @@
...
+import sass from 'sass';
+import { promisify } from 'util';
+import { writeFile, mkdir, existsSync, rmdir, readdirSync } from 'fs';
+import path from 'path';
+
+// promisify the functions in favor of async/await syntax
+const sassRenderPromise = promisify(sass.render);
+const writeFilePromise = promisify(writeFile);
+const makeDirPromise = promisify(mkdir);
+const removeDirPromise = promisify(rmdir);
```
jodyheavener, line 14: You shouldn't need to promisify any of these I don't think - they all have `*Sync` versions: https://gist.github.com/jodyheavener/bf60cafdf6b3dcdd49984c8a31f1bbab

jodyheavener (changes requested):
I think this is super close! I left one comment about the styles command, and I wanted to point out some small visual discrepancies between the [current template](https://user-images.githubusercontent.com/6392049/125793595-cf8883f7-0a3d-498f-af47-bc4451b6bf14.png) and your [new one](https://user-images.githubusercontent.com/6392049/125793621-56ece6eb-565e-49ca-95c9-2549042789a2.png):

- The max-width should be tighter; current is around 310px, new one is around 590px.
- This might be fixed when we tighten up the max-width, but the two download badges are not aligned well; they should either be aligned together (right/left), or apart (left/right).
- The vertical padding between the items (title-image, image-paragraph, paragraph-button, button-badges) is tight and could be increased a bit to match the original version.
- The button sticks out further than the badges/paragraph on left and right sides; it should be flush with them
- The footer font size in the current template is 11px, but the new one is 12px. This might have been intentional and maybe @LZoog will push back on this for accessibility, which is fine and I can get behind, but I wanted to point out that this is different.

These are small nitpicks (sorry!), but I think we should maintain the original designs as much as possible (with some exceptions).

packages/fxa-auth-server/lib/senders/emails/sass-compile-files.ts
```diff
@@ -0,0 +1,57 @@
...
+} from 'fs';
+import path from 'path';
+
+const getDirectories = (source: string) =>
+  readdirSync(source, { withFileTypes: true })
+    .filter((dirent: any) => dirent.isDirectory())
+    .map((dirent: any) => dirent.name);
+
+// construct arrays of partials and templates based on the directories
+const partials = getDirectories(path.join(__dirname, `partials`));
```
LZoog, line 21: This is a tiny nitpick, but we only really need the backtick if there's a variable in the string. We can change a lot of thes in this file beginning with backticks to `'`.

LZoog (commented):
Looking so good - Jody had one comment and I had one as well, both are really minor. I will r+ after those are addressed! :shipit:

Yogita98:
I'm not sure but are the tests failing related to my changes? My last push involved a bad lock file which got fixed in the current push but there are a lot of tests failing now.

LZoog:
@Yogita98 Doesn't look like you triggered those failures to me. When you push again they'll rerun and let's see then... 🤔

Sorry to throw this at ya when we thought we’d get this merged today, but after taking another peek this afternoon I think there’s a few improvements we can make to set up consistency for later. Hopefully going in and hacking away at most of what I was thinking makes that a little better… 😅

I pushed a few changes to my fork, you can see the comparison here: https://github.com/Yogita98/fxa/compare/styling-solution...LZoog:styling-solution

Note: There’s some margin weirdness happening in that footer text still, I probably missed adding an `!important`. I got you most of the way at least but there’s still just a few tweaks on this branch that are needed.

Feel free to simply cherry-pick the commit (or not). Here’s what I did:
* Changed `$p-#` values to `$s-#` to signify “space” so that way, we can use those values for padding and margins
* Added comments for organization in the globals file and added some utility classes - these match our Tailwind classes, similar to what we’re doing with variables. I still maintain that doing this will make development easier for devs since we’re using Tailwind in other packages. Maybe it’s a pipe dream, but I’d love to eventually have access to Tailwind styles here… Preemptively naming utility classes to match Tailwind classes would make that transition easier in the future.
* Nested styles where applicable
* Took the `!important`s out of the variables themselves, I think these probably belong in classes
* Took out the `span`s in the footer. I don’t want to fool us into thinking those are actually links when they aren’t yet. We’ll need to add those links in the `cadReminderFirst` ticket.

I’m kind of wondering if those mixins should just be classes that we `extend` from. I made them mixins originally because I wondered if we’d want to override the line-height/font-size later on but I’m not sure that we will. I’m fine with leaving them as mixins or just turning them into utility classes.

I think eventually, we’re probably going to want to separate files out of `global.scss`. We may want a `styles/` directory and then `_variables.scss` and `_utility-classes.scss` - I didn’t do that here since I didn’t want to also have to edit the compilation step, but just something to keep in mind for later as that file grows.

Here’s some things I think need to be addressed in that `cadReminder` ticket (we should probably drop a comment over there to make sure these are addressed) -

Two things I noticed about the button:
* The copy should be “Sync another device” and currently reads as “Sync device"
* Only the text is clickable in this new version. In the old version, the entire button is clickable.

Then in `appBadges/index.ts`, we have this:

```
<% if(onDesktopOrTabletDevice){ %> Or, install on
<span
href="<%= baseURL %><%= anotherDeviceURL %>"
color="#0a84ff"
text-decoration="none"
font-family="sans-serif"
>another version</span
>
<% } else{ %> Or, install on
<span
href="<%= baseURL %><%= anotherDeviceURL %>"
color="#0a84ff"
text-decoration="none"
font-family="sans-serif"
>another device</span
>
<% } %>
```

We should prefer classes instead of `color=“#a0a84ff”` etc. I changed these to anchors - we probably want some global styles for links. This can be addressed here or in that `cadReminder` ticket.

BTW, I _love_ that in Storybook, we can see that `doc` value without having to render it above the template.

I'll check back tomorrow morning and hopefully we can :shipit: 😏

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -0,0 +1,77 @@
...
+.text-blue-400 {
+  color: $blue-400;
+}
+
+.mt-6 {
+  margin-top: $s-6 !important;
+}
+
+.mb {
+  &-3 div {
```
Yogita98, line 52: I think we need to explicitly apply mirgin to the div of the class so I had to do it this way

Yogita98:
Just two things are remaining which can be completed as part of the cadReminder ticket.
1. Using anchor text for the footer text. The footer will eventually be moved out to partials which is why I didn't touch it for now
2. Creating separate individual stylesheets for utilty classes and variables if needed

packages/fxa-auth-server/lib/senders/emails/global.scss
LZoog: Ah interesting. If/when we need `mb-3` _without_ `div`, we should update this line to be `&-3, &-3 div`.

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -0,0 +1,77 @@
...
+  font-size: 14px !important;
+  line-height: 20px !important;
+}
+.text-lg {
+  font-size: 18px !important;
+  line-height: 28px !important;
+}
+.text-xl {
+  font-size: 20px !important;
+  line-height: 28px !important;
```
LZoog, line 35: We could nest these too:

```
.text {
  &-xs {
    font-size: 12px !important;
    line-height: 18px !important;
  }
  &-sm {
  ... // etc.
```

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -0,0 +1,77 @@
...
+  padding: $s-0 !important;
+}
+
+.fxa-logo td {
+  padding: $s-5 $s-0 !important;
+  height: 60px !important;
+  width: 60px !important;
+}
+
+.blue-anchor {
```
LZoog, line 73: Ohh, this isn't something you would know, but we actually have a `link-blue` class in `fxa-settings` similar to this. Let's call this `link-blue` too? :) 

BTW, we generally go with [thing]-[descriptor], so that way all "things" can be grouped together. We have components named like, `LinkExternal`, `LinkAbout` or `PageSettings`, `PageAvatar` etc. and the custom SCSS for settings follows the same pattern so that way they're all grouped together and it's easy to see variations.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.scss
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../global.scss';
+
+body > div:first-child {
```
LZoog, line 7: I wonder if this selector could be a bit nicer (like with a class instead) or if this'll end up being a global style. I think it's fine to leave for now, just thinking about future templates 🙂

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.scss
```diff
@@ -0,0 +1,30 @@
...
+  max-width: 310px;
+  margin: 0 auto;
+}
+
+.header-text div {
+  @extend .text-xl;
+  @extend .font-sans;
+  text-align: center !important;
+  margin-bottom: 20px !important;
```
LZoog, line 16: I think we should `@extend global.mb-5` here.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.ts
```diff
@@ -27,15 +24,14 @@ export const render = () => {
...
-        </mj-section>
         ${button}
         ${appBadges}
         <mj-section>
           <mj-column>
-            <mj-text css-class="secondary-text">This is an automated email; if you received it in error, no action is required. For more information, please visit Mozilla Support.</mj-text>
-            <mj-text css-class="secondary-text"><span>Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</span></mj-text>
-            <mj-text css-class="secondary-text"><span>Mozilla Privacy Policy</span></mj-text>
-            <mj-text css-class="secondary-text"><span>Firefox Cloud Terms of Service</span></mj-text>
+            <mj-text css-class="secondary-text mt-6 mb-3">This is an automated email; if you received it in error, no action is required.For more information, please visit <span>Mozilla Support<span></mj-text>
```
LZoog, line 31: I think `mt-6` isn't being applied, at least when I'm running it locally. Maybe it also needs to be `mt-6 div`?

LZoog (commented):
Just a few suggestions. Will r+ after those quick changes. Thanks for sticking this out Yogita, I'm super happy with where it's ended up and I think the patterns we're establishing here will make styling the new templates easier and more maintainable👌

packages/fxa-auth-server/lib/senders/emails/partials/button/index.scss
```diff
@@ -0,0 +1,20 @@
...
+  height: 56px !important;
+  width: 310px !important;
+
+  a {
+    @extend .font-sans;
+    @extend .text-lg;
+    background: global.$blue-400 !important;
+    width: 310px !important;
+    color: global.$white;
+    padding: 14px 0 !important;
```
LZoog, line 17: Ohh I missed this in my review I just submitted, this value should either be `12px` or `16px`.

I think our rule of thumb is becoming if you're setting `padding` or `margin`, use the variables provided. Though, I'll run this one by a designer and it can be fixed in the `cadReminder` ticket because I'm not sure which would be preferred.

Yogita98:
I have noted down the things pending in this issue in [here](https://github.com/mozilla/fxa/issues/9255#issuecomment-881562920) for handiness.

jodyheavener (approved):
Seems good to me, thanks for addressing my feedback, so I'll r+ but I'll let @LZoog make the final r+!

LZoog (approved):
LGTM too! 🎉

Issue #9345, opened 2021-06-21
[subplat] Idempotency Key is Not Updated Between Paying with PayPal Attempts


data-sync-user:
When a customer is paying with PayPal, if the first subscription attempt fails, the second attempt will use the idempotency key from the first attempt.  That causes Stripe to return the response from the first attempt.  From then on SubPlat is using the incorrect subscription and invoice ids.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3625)

data-sync-user:
➤ Barry Chen commented:

Valentina Virlics  Sorry that I didn’t provide any QA info.  Here it is.

  1. Create a new account.
  2. Try subscribe to https://accounts.stage.mozaws.net/subscriptions/products/prod_JqSgEW0LrHV90G ( https://accounts.stage.mozaws.net/subscriptions/products/prod_JqSgEW0LrHV90G ) with PayPal
  3. Observe the idempotencyKey in the payload to https://api-accounts.stage.mozaws.net/v1/oauth/subscriptions/active/new-paypal ( https://api-accounts.stage.mozaws.net/v1/oauth/subscriptions/active/new-paypal ) in the network tab of Firefox devtools.  (See screenshot.)
  4. Try again, and see that idempotencyKey changed.

Note that that plan will continue to fail when PayPal used.

!image-20210713-134752.png|width=1198,height=540!

data-sync-user:
➤ Valentina Virlics commented:

Verified this as fixed on FxA 2.210.1 uing Windows 10 and Firefox Release.

The idempotencyKey is changed each time a subscription is tried using PayPal payment method. Subscription cannot be made via PayPal.

Attaching screenshots;

!key.png|width=1152,height=560!

!image_2021_07_14T13_39_17_053Z.png|width=1920,height=923!

PR #9820, opened 2021-07-09
fix(subscriptions): update idempotency key on PayPal btn callback

chenba:
Because:
 - the idempotency key used for Stripe requests was not updated
   between PayPal subscription attempts

This commit:
 - refresh the idempotency key in the PayPal button callback

Closes: #9345

### DIFF
--- a/packages/fxa-payments-server/src/components/PayPalButton/index.tsx
+++ b/packages/fxa-payments-server/src/components/PayPalButton/index.tsx
@@ -16,6 +16,7 @@ export type PaypalButtonProps = {
   currencyCode: string;
   customer: Customer | null;
   idempotencyKey: string;
+  refreshSubmitNonce: () => void;
   refreshSubscriptions: () => void;
   setPaymentError: Function;
   priceId?: string;
@@ -47,6 +48,7 @@ export const PaypalButton = ({
   currencyCode,
   customer,
   idempotencyKey,
+  refreshSubmitNonce,
   refreshSubscriptions,
   setPaymentError,
   priceId,
@@ -119,6 +121,7 @@ export const PaypalButton = ({
           refreshSubscriptions();
         }
       }
+      refreshSubmitNonce();
       return null;
     },
     [
--- a/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
@@ -38,6 +38,15 @@ import { updateConfig } from '../../../lib/config';
 
 import { ButtonBaseProps } from '../../../components/PayPalButton';
 
+jest.mock('../../../lib/hooks', () => {
+  const refreshNonceMock = jest.fn().mockImplementation(Math.random);
+  return {
+    ...jest.requireActual('../../../lib/hooks'),
+    useNonce: () => [Math.random(), refreshNonceMock],
+  };
+});
+import { useNonce } from '../../../lib/hooks';
+
 // TODO: Move to some shared lib?
 const deepCopy = (object: Object) => JSON.parse(JSON.stringify(object));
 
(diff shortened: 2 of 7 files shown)


packages/fxa-payments-server/src/components/PayPalButton/index.tsx
```diff
@@ -119,6 +121,7 @@ export const PaypalButton = ({
           refreshSubscriptions();
         }
       }
+      refreshSubmitNonce();
```
chenba, line 124: This is the actual change.  Everything else is either updating types and props to pass this function down or prettier formatting changes.

biancadanforth: Hmm, when I try this locally, logging the `submitNonce` in `SubscriptionCreate`, I don't see that the nonce is changing regardless of the outcome of the `try..catch` block in `onApprove`. Do you?

Also, may be worth having a test here to assert that the submit nonce is a different value in `SubscriptionCreate` before and after `onApprove` is called?

biancadanforth (changes requested):
Hey Barry; I know you're still resolving some CI issues, but I tried out this PR, and I'm not seeing a new value for the submit nonce after `onApprove` executes. Are you getting a new value?

packages/fxa-payments-server/src/components/PayPalButton/index.tsx
chenba: > 
> 
> Hmm, when I try this locally, logging the `submitNonce` in `SubscriptionCreate`, I don't see that the nonce is changing regardless of the outcome of the `try..catch` block in `onApprove`. Do you?
> 
I tested it by using the "PayPal Invalid Billing Agreement Error" product in dev Stripe and looking at the idempotency key sent in the requests.  They were changing between requests.  I can check it your way and see what's happening there.

> Also, may be worth having a test here to assert that the submit nonce is a different value in `SubscriptionCreate` before and after `onApprove` is called? (or at least asserting `refreshSubmitNonce` is called, assuming it has its own tests to ensure it actually generates a new nonce...)

chenba: > 
> 
> Hmm, when I try this locally, logging the `submitNonce` in `SubscriptionCreate`, I don't see that the nonce is changing regardless of the outcome of the `try..catch` block in `onApprove`. Do you?
> 
Where are you logging the nonce?  I can see it changing if I log it right before the return.

chenba: Oh I see you were expecting to see the log message on success, that's fair.  SubscriptionCreate doesn't get re-rendered, so I believe that's why you are not seeing the message.

biancadanforth: Okay thanks I verified this now via the Network Monitor in DevTools as well as in the Stripe dashboard looking at the REST API logs for the fake customer upon retries.

biancadanforth (approved):
R+ assuming CI passes.

Verified the idempotencyKey changes with each attempt now to create a subscription. Thanks for the clarification Barry.

It'd be nice to get a test here to assert `refreshSubmitNonce` is called in `onApprove`, but I won't block on that if you want it to be a follow up, as I know this is a hotfix to deploy in a dot release.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -38,6 +38,15 @@ import { updateConfig } from '../../../lib/config';
 
 import { ButtonBaseProps } from '../../../components/PayPalButton';
 
+jest.mock('../../../lib/hooks', () => {
+  const refreshNonceMock = jest.fn().mockImplementation(Math.random);
+  return {
+    ...jest.requireActual('../../../lib/hooks'),
+    useNonce: () => [Math.random(), refreshNonceMock],
+  };
+});
```
chenba, line 47: @biancadanforth Here's the mock...

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -898,6 +914,7 @@ describe('routes/ProductV2/SubscriptionCreate', () => {
     expect(
       screen.queryByTestId('error-payment-submission')
     ).toBeInTheDocument();
+    expect(refreshSubmitNonce).toHaveBeenCalledTimes(1);
```
chenba, line 917: @biancadanforth ...and here is the test.

PR #9808, opened 2021-07-08
feat(emails): Add mjml feature flag

vbudhram:
## Because

- We need a feature flag to quickly and easily test the new mjml rendered email templates

## This pull request

- Allows you to specify which email template and which emails get the feature

## Issue that this pull request solves

Closes: https://github.com/mozilla/fxa/issues/9252

## Checklist

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [ ] I have verified that my changes render correctly in RTL (if appropriate).

PR #9821, opened 2021-07-12
feat(fxa-auth-server): Setup fluent localizer to localize our emails

Yogita98:
## Because

- We want to replace our current localizer with fluent

## This pull request

- Includes scripts and packages that can help us setup fluent in fxa-auth-server

Closes: #9257

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -416,7 +419,22 @@ module.exports = function (log, config) {
     );
   };
 
-  Mailer.prototype.localize = function (message) {
+  Mailer.prototype.localize = async function (message) {
+    if (newEmailPattern.test(message.template)) {
+      const { localizedHTML, localizedText, localizedSubject } =
+        await this.fluentLocalizer.localizeEmail(
+          message.template,
+          'hello-world'
```
Yogita98, line 427: @jodyheavener @LZoog Temporarily, I'm passing this string as a subject since I wasn't sure where exactly it's being created

packages/fxa-auth-server/lib/senders/emails/locales/en-US.ftl
```diff
@@ -0,0 +1,5 @@
+hello-world = Hello, world!
```
Yogita98, line 1: Let me know your thoughts on the directory structure

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder.ts
```diff
@@ -28,10 +28,11 @@ export const render = () => {
         ${appBadges}
         <mj-section>
           <mj-column>
-            <mj-text css-class="secondary-text">This is an automated email; if you received it in error, no action is required. For more information, please visit Mozilla Support.</mj-text>
+            <mj-text css-class="secondary-text" data-l10n-id="hello-world">This is an automated email; if you received it in error, no action is required. For more information, please visit Mozilla Support.</mj-text>
             <mj-text css-class="secondary-text"><span>Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</span></mj-text>
             <mj-text css-class="secondary-text"><span>Mozilla Privacy Policy</span></mj-text>
             <mj-text css-class="secondary-text"><span>Firefox Cloud Terms of Service</span></mj-text>
+            <mj-text><h3 data-l10n-id="hello-world">Is this localized?</h1></mj-text>
```
Yogita98, line 35: [fluent/dom](https://github.com/projectfluent/fluent.js/tree/master/fluent-dom) localizes the html elements that have `data-l10n-id` as an attribute. Here, it is able to localize the `h3` tag but nothing happens in case of `mj-text`. And while we can register [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) by extending the HTMLElement class, all mj-elements ultimately get converted over to html.

packages/fxa-auth-server/lib/senders/emails/templates/index.ts
```diff
@@ -1 +1 @@
-export * as cadReminder from './cadReminder';
+export * as cadReminderMJML from './cadReminder';
```
Yogita98, line 1: Exporting it as `cadReminderMJML` since @vbudhram has included a regex pattern in [this PR](https://github.com/mozilla/fxa/pull/9808/commits/3c77c9039b3c6ed22638be9ff0798e170ac08340) and I thought we could use `MJML` as a pattern to enable feature flagging for the new emails

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -416,7 +419,22 @@ module.exports = function (log, config) {
     );
   };
 
-  Mailer.prototype.localize = function (message) {
+  Mailer.prototype.localize = async function (message) {
+    if (newEmailPattern.test(message.template)) {
```
jodyheavener, line 423: I think Vijay's PR for feature flagging is already going to handle determining if an email should use the new or old template, so this shouldn't be necessary. Base on [this line](https://github.com/mozilla/fxa/pull/9808/files#diff-8f8d686a1e503181760ab061f610c01e8c357bd0da6620272cf9593d9a02e146R433) you should be able to (once that PR is merged) do something like:

```js
if (featureFlags.isMjmlEnabledForUser(message.email, message.template)) {
  // Needs to return `html` and `text` properties
  const localized = await this.fluentLocalizer.localizeEmail(
    message.template,
    'hello-world'
  );
  localizedEmailHtml = localized.html;
  localizedEmailText = localized.text;
}
```

And I _think_ this should allow you to just use the existing `cadReminderEmail` method and not need a new `cadReminderMJMLEmail`?

You could also define the subject Fluent key when calling `this.send` in the individual template mailer methods. Something like: `this.send({ mailSubject: 'hello-world' })`.

Finally, I see `localizeEmail` is importing and using `context` from `renderer.ts`, but I think you'll need to pass variables down to `this.fluentLocalizer.localizeEmail` instead. I think `message.templateValues` is what you'd pass down.

So... altogether, something like:

```js
if (featureFlags.isMjmlEnabledForUser(message.email, message.template)) {
  const { template, mailSubject, templateVariables } = message;
  const localized = await this.fluentLocalizer.localizeEmail(
    template,
    mailSubject,
    templateVariables,
  );
  localizedEmailHtml = localized.html;
  localizedEmailText = localized.text;
} else {
  // the old way
}
```

jodyheavener (commented):
Haven't dug too deeply in to this yet, I think we should wait until #9808 since it will greatly help us determine when to use the new or old localize operations.

packages/fxa-auth-server/config/dev.json
```diff
@@ -426,7 +426,7 @@
     "secondInterval": "2s"
   },
   "mjml": {
-    "enabledEmailAddress": "@mozilla.com$",
-    "templates":  []
+    "enabledEmailAddress": "@testuser.com$",
```
Yogita98, line 429: I figured that locally we're using this address to send emails so it's just pushed to test the changes

Yogita98:
Things done as part of this PR:
1. Added support for optional rendering with/without a layout
2. Added separate files for `fxa` layout and `partials` like `automatedEmailNoAction`
3. Created `.ftl` files for `en` and `en-US` locales, will add more once the directory structure is finalized
4. Implemented `fluent` localization to localize `cadReminderFirst` email

packages/fxa-auth-server/lib/senders/emails/locales/en-US.ftl
```diff
@@ -0,0 +1,7 @@
+sync-reminder = Here's your reminder to sync devices.
+sync-description = It takes two to sync. Syncing another device with Firefox privately keeps your bookmarks, passwords and other Firefox data the same everywhere you use Firefox.
+
+random1 = Or, install on <a %(desktopLinkAttributes)s>another desktop device</a>.
```
Yogita98, line 4: Still need to dig a little bit more into the localization part, so leaving it here as is for now

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
```diff
@@ -0,0 +1,54 @@
...
+          &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌‌‌
+        </div>
+        <% } %>
+      <% if(typeof(sync) == "undefined" || sync == null){ %>
+      <img
+             src="https://accounts-static.cdn.mozilla.net/product-icons/firefox-logo.png"
+             class="fxa-logo"
+            />
```
jodyheavener, line 36: Some funky indentation, maybe more like this?

```
<img
  src="https://accounts-static.cdn.mozilla.net/product-icons/firefox-logo.png"
  class="fxa-logo"
/>
```

That's how we write it for React components at least

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      <% } %>
+    </mj-raw>
+    ${body}
+    <mj-section>
+          <mj-column>
+            <mj-text css-class="secondary-text">Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</mj-text>
+            <mj-text css-class="secondary-text"><a class="link-blue" href="<% privacyUrl %>" class="blue-link">Mozilla Privacy Policy</a></mj-text>
+            <mj-text css-class="secondary-text"><a class="link-blue" href="https://www.mozilla.org/about/legal/terms/services/" class="blue-link">Firefox Cloud Terms of Service</a></mj-text>
+          </mj-column>
+        </mj-section>
```
jodyheavener, line 51: Nit: Indentation needs correcting

packages/fxa-auth-server/lib/senders/emails/partials/button/index.ts
```diff
@@ -6,7 +6,7 @@ export const button = `
   <mj-include path="./lib/senders/emails/css/button/index.css" type="css" css-inline="inline" />
   <mj-section>
     <mj-column>
-      <mj-button css-class="primary-button" href="<%= baseURL %><%= anotherDeviceURL %>"><%= buttonText %></mj-button>
+      <mj-button css-class="primary-button" href="<%= baseUrl %><%= anotherDeviceUrl %>"><%= buttonText %></mj-button>
```
jodyheavener, line 9: What if `anotherDeviceUrl` and other URLs incorporated `baseUrl` on the server side, so we don't need to pass down `baseUrl` as a variable everywhere?

packages/fxa-auth-server/lib/senders/emails/partials/metadata.ts
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export const metaData = `
+  <% if(oneClickLink){ %>
```
jodyheavener, line 6: Nit: we'll want to maintain JS syntax consistency, since prettier doesn't pick up on ERB (at least not yet): `<% if (oneClickLink) { %>`

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.ts
```diff
@@ -2,39 +2,38 @@
...
+    <mj-text css-class="header-text sync-reminder">Here's your reminder to sync devices.</mj-text>
+    <mj-image css-class="sync-logo sync-description" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
+  </mj-column>
+</mj-section>
+<mj-section>
+  <mj-column>
+    <mj-text css-class="primary-text">It takes two to sync. Syncing another device with Firefox privately keeps your bookmarks, passwords and other Firefox data the same everywhere you use Firefox.</mj-text>
+  </mj-column>
+</mj-section>
+${button}
```
jodyheavener, line 21: I just noticed this, up to you if you wanna address it in this PR or not: the button in the current template has a background of `#0a84ff`, and the new one uses `#0090ed`

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    await l10n.translateRoots();
+
+    const subject = await l10n.formatValue(mailSubject);
+    return {
+      html: document.documentElement.outerHTML,
+
+      text: document.documentElement
+        .querySelector('body')
+        ?.textContent?.replace(/(^[ \t]*\n)|(^[ \t]*)/gm, ''),
```
jodyheavener, line 92: What's this regex doing? Stripping HTML? At any rate text version is going to need some love -- not sure if we want to try and address this now or in a future PR:

Screen Shot 2021-07-19 at 2 31 56 PM

jodyheavener (commented):
I like how this has come along! Locally it works and renders correctly, just left a few comments so far

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
Yogita98: Yeah I was trying to replace all the white spaces and empty lines and it worked fine until those hidden headers were added. I'll see if I can gracefully remove those as well in this PR itself or we can create a follow-up PR on this!

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        <div class="hidden-div">
+        <%= preHeader %>
+          &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          &nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;
+          ‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌&nbsp;‌‌‌
+        </div>
+        <% } %>
+      <% if(typeof(sync) == "undefined" || sync == null){ %>
```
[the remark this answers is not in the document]
Yogita98: I tried to use the shortened condition, but it throws `undefined` error

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.ts
```diff
@@ -2,39 +2,57 @@
...
+};
+
+export const localizationAttributes = `
+  <mj-html-attributes>
+    ${addDataL10nId('.header-text div', 'sync-reminder')}
+    ${addDataL10nId('.primary-text div', 'sync-description')}
+    ${addDataL10nId('.automated-text div', 'automated-email-text')}
+    ${addDataL10nId('.install-device-link div', 'another-desktop-device')}
+    ${addDataL10nId('.install-desktop-device-link div', 'another-device')}
+    ${addDataL10nId('.privacy-url div', 'footer-privacy-url')}
```
Yogita98, line 51: I know the attribute names are looking weird, but I couldn't think about better ones. It'd be nice if we can set a convention for naming them

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
```diff
@@ -0,0 +1,97 @@
...
+  async localizeEmail(
+    templateName: string,
+    layoutName: string,
+    mailSubject: string,
+    variables: Record<any, any>,
+    acceptLanguage: string
+  ) {
+    const htmlDocument = renderWithOptionalLayout(
+      templateName,
+      { ...variables, ...context },
```
Yogita98, line 28: The templateValues object could be passed down in a better way. Maybe this can be done as part of the cadReminder ticket

packages/fxa-auth-server/lib/senders/emails/locales/ar.ftl
```diff
@@ -0,0 +1,8 @@
+# cadReminderFirst
```
LZoog, line 1: Oooh I hope you didn't manually copy each of these over by hand (or I hope it didn't take too long). 😅

I believe these get automatically converted for us when we build - we have an l10n team that helps with translations. I see we only have the English version committed in `fxa-settings`:
image

I am not as familiar with Fluent workings as some others may be, but it seems like we only want to commit an FTL file per template/partial? @jodyheavener does that seem right to you? 🤔 I tried to find documentation on Fluent and how everything hooks up, but only found docs for GetText.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/index.ts
LZoog: Hrm, I know you did this way since MJML doesn't allow custom attributes, but I think we had said before (maybe just once in our 3-way group chat) that we could potentially wrap text strings in `<span>` to apply the custom data-attribute to... I don't like the idea of extra DOM elements so I wondered if there was a better way, but doing this per template/per line to be extracted seems a little confusing/cumbersome (especially given the weirdness around MJML adding those `div`s or table cells). I don't necessarily have an alternative outside of suggesting `span`, but if we do keep this, we probably want to file a follow up ticket if even to just investigate how else we can do this.

packages/fxa-auth-server/package.json
```diff
@@ -24,7 +24,7 @@
     "test-remote": "MAILER_HOST=restmail.net MAILER_PORT=80 CORS_ORIGIN=http://baz mocha -r ts-node/register --timeout=300000 test/remote",
     "format": "prettier --write --config ../../_dev/.prettierrc '**'",
     "gen-keys": "ts-node ./scripts/gen_keys.js; ts-node ./scripts/oauth_gen_keys.js; ts-node ./scripts/gen_vapid_keys.js",
-    "write-emails": "node -r ts-node/register ./scripts/write-emails-to-disk.js",
+    "write-emails": "yarn emails-scss && node -r ts-node/register ./scripts/write-emails-to-disk.js",
```
LZoog, line 27: Related to this comment: https://github.com/mozilla/fxa/pull/9923/files#r677542177

I thought `yarn write-emails` was for the old templates? When I try to run this, I get:

image

Is there another step, or are we no longer using `write-templates`?

I think we also want to take use of `yarn` out and use `npm run` instead.

On a related note, CI is failing with this (not sure if you can see):
image

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.scss
```diff
@@ -0,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../global.scss';
+
+.hidden-div {
```
LZoog, line 7: Maybe we can just call this `hidden`, I know it's specifically on a `div` here but it feels more versatile/like our other classes with that naming convention.

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    l10n.connectRoot(document.documentElement);
+
+    await l10n.translateRoots();
+
+    const subject = await l10n.formatValue(mailSubject);
+    return {
+      html: document.documentElement.outerHTML,
+
+      text: document.documentElement
```
vbudhram, line 84: I think it might be better to do text rendering in a follow up PR. In meantime maybe pass the old text template to the email sender?

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const currentLocales = negotiateLanguages(
+      [...userLocales],
+      [...OTHER_EN_LOCALES, ...availableLocales],
+      {
+        defaultLocale: 'en-US',
+      }
+    );
+
+    async function fetchResource(locale: string) {
```
vbudhram, line 53: I believe this will attempt to fetch and merge ftl files on every attempt to localize an email. We could store this value in a Map and have an initalize function to do the work. Similar to https://github.com/mozilla/fxa/blob/5f63d848b9bbed1a7f478b671a018a6557d13b75/packages/fxa-auth-server/lib/senders/templates/index.js#L56

packages/fxa-auth-server/bin/mjml-server.ts
```diff
@@ -5,9 +5,9 @@
 import http from 'http';
 import stream from 'stream';
 import net from 'net';
-import { render } from '../lib/senders/emails/renderer';
+import { renderWithOptionalLayout } from '../lib/senders/emails/renderer';
```
vbudhram, line 8: I think you will also need to update https://github.com/mozilla/fxa/blob/8a5e52f0d05892b0bcdb1d44753e5d37bc856e46/packages/fxa-auth-server/lib/senders/emails/render-to-disk.ts#L18

vbudhram (commented):
@Yogita98 Just a few comments, but this looks really good. Storybook views are pretty neat!

Screen Shot 2021-07-27 at 2 22 35 PM

packages/fxa-auth-server/lib/senders/emails/fluentLocalizer.ts
Yogita98: Ohh yes, good point!

dannycoates:
I'm about to merge #9975 that'll probably make a merge conflict in `yarn.lock` here. To resolve it you should be able to rebase this branch off of `main` and run `yarn install`. Yarn _should_ fix the conflict in `yarn.lock` automatically. If `package.json` also has a conflict you should manually fix that after the rebase but before running `yarn install`. If things get too messed up I can help you fix it.

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -434,14 +436,21 @@ module.exports = function (log, config) {
...
+      let localized = await this.fluentLocalizer.localizeEmail(
+        template,
+        layout || 'fxa',
+        subject,
+        templateValues,
+        message.acceptLanguage
+      );
+      localizedEmailHtml = localized.html;
+      // pass old plaintext version of the email for now
+      localized = this.templates.render(
```
vbudhram, line 449: I think we can remove the else block below and always render original emails. Then overwrite with mjml templates as needed.

dannycoates:
@Yogita98 this patch should fix CircleCI

Download it and run `git apply yarnlock.patch.txt` (or just look at it and copy over the new hashes)

[yarnlock.patch.txt](https://github.com/mozilla/fxa/files/6901876/yarnlock.patch.txt)

Yogita98:
@dannycoates Thank you very much for helping us with this, really appreciate you taking out time :)

vbudhram (approved):
@Yogita98 PR looks good :)

Strangely some of our content-server tests failed to acquire a lock? I've restarted them 🤞🏽

adevade16:
I can see all the tests passing now, not sure what was causing the failures before 🤔

Yogita98:
I can see the tests passing all of a sudden, not sure what was causing the failures before 🤔

PR #9824, opened 2021-07-12
feat(payments): create new Checkout route for unauthenticated users

### LINKED ISSUE #9701: [new user checkout] create route and base component
Ticket to account for work setting up page layout, HTML skeleton for the page

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3676)

biancadanforth:
Because:

* We are creating a [passwordless flow](https://mozilla-hub.atlassian.net/browse/FXA-3486) for new users to sign up for a subscription.

This commit:

* Creates a new {payments-server}/checkout/:productid route; this is where the new flow ([designs](https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098)) will live.

Closes #9701

## Checklist

To try this out, you can go to http://localhost:3031/checkout/{productId}. It's just an empty page right now that will display the `selectedPlan` if available, else a plan error.

### DIFF
--- a/packages/fxa-payments-server/src/App.tsx
+++ b/packages/fxa-payments-server/src/App.tsx
@@ -21,6 +21,7 @@ import { LoadingOverlay } from './components/LoadingOverlay';
 import * as FlowEvents from './lib/flow-event';
 import { observeNavigationTiming } from 'fxa-shared/metrics/navigation-timing';
 
+const Checkout = React.lazy(() => import('./routes/Checkout'));
 const Product = React.lazy(() => import('./routes/Product'));
 const Subscriptions = React.lazy(() => import('./routes/Subscriptions'));
 
@@ -106,6 +107,14 @@ export const App = ({
                           render={(props) => (
                             <SignInLayout>
                               <Product {...props} />
+                            </SignInLayout>
+                          )}
+                        />
+                        <Route
+                          path="/checkout/:productId"
+                          render={(props) => (
+                            <SignInLayout>
+                              <Checkout {...props} />
                             </SignInLayout>
                           )}
                         />
--- a/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import PlanErrorDialog from './index';
+import { PLANS } from '../../lib/mock-data';
+import { FetchState, Plan } from '../../store/types';
+
+const locationReload = () => {};
+const plans: FetchState<Plan[], any> = {
+  error: null,
+  loading: false,
+  result: null,
+};
+
+afterEach(cleanup);
+
+describe('PlanErrorDialog', () => {
+  it('renders as expected for no plan for product', () => {
+    const subject = () => {
+      return render(
+        <PlanErrorDialog
+          {...{
+            locationReload,
+            plans,
+          }}
+        />
+      );
+    };
+    const { queryByTestId } = subject();
+    const errorMessage = queryByTestId('error-loading-plans');
+    expect(errorMessage).toBeInTheDocument();
+  });
+
+  it('renders as expected for no selectedPlan', () => {
+    const subject = () => {
+      return render(
+        <PlanErrorDialog
+          {...{
+            locationReload,
+            plans: { ...plans, result: PLANS },
+          }}
+        />
+      );
+    };
+    const { queryByTestId, queryByText } = subject();
+    const errorMessage = queryByTestId('no-such-plan-error');
+    expect(errorMessage).toBeInTheDocument();
+    expect(queryByText('Plan not found')).toBeInTheDocument();
+  });
+});
--- a/packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
+++ b/packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
@@ -1,0 +1,138 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import noc from 'nock';
+
+function nock(it: any) {
+  //@ts-ignore
+  return noc(...arguments).defaultReplyHeaders({
+    'Access-Control-Allow-Origin': '*',
+  });
+}
+
+import {
+  PRODUCT_ID,
+  PRODUCT_REDIRECT_URLS,
+  MOCK_PLANS,
+  expectNockScopesDone,
+  defaultAppContextValue,
+  MockApp,
+  setupMockConfig,
+  mockConfig,
+  mockServerUrl,
+  mockOptionsResponses,
+} from '../../lib/test-utils';
+
+jest.mock('../../lib/sentry');
+
+jest.mock('../../lib/flow-event');
+
+import { SignInLayout } from '../../components/AppLayout';
+import Checkout from './index';
+import { AppContextType } from '../../lib/AppContext';
+import { PLANS } from '../../lib/mock-data';
+
+describe('routes/Checkout', () => {
+  let authServer = '';
+  let profileServer = '';
+
+  beforeEach(() => {
+    setupMockConfig({
+      ...mockConfig,
+      productRedirectURLs: PRODUCT_REDIRECT_URLS,
+    });
+    authServer = mockServerUrl('auth');
+    mockOptionsResponses(authServer);
+    profileServer = mockServerUrl('profile');
+    mockOptionsResponses(profileServer);
+  });
+
+  afterEach(() => {
+    noc.cleanAll();
+    return cleanup();
+  });
+
+  const Subject = ({
+    productId = PRODUCT_ID,
+    planId,
+    matchMedia = jest.fn(() => false),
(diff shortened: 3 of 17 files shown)


packages/fxa-payments-server/src/App.tsx
```diff
@@ -109,6 +110,14 @@ export const App = ({
                             </SignInLayout>
                           )}
                         />
+                        <Route
+                          path="/checkout/:productId"
+                          render={(props) => (
+                            <SignInLayout>
```
biancadanforth, line 116: Is this still being used? If I remove it from the existing `Product` route, I don't see any noticeable style changes on reload. It was added in #1387.

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  afterEach(() => {
+    noc.cleanAll();
+    return cleanup();
+  });
+
+  const Subject = ({
+    productId = PRODUCT_ID,
+    planId,
+    accountActivated,
```
biancadanforth, line 61: Is this query param used for anything? I got this from the `Product` route. It's passed into `SubscriptionCreate` there, but AFAICT, nothing is done with it...

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const selectedPlan = useMemo(
+    () => getSelectedPlan(productId, planId, plansByProductId),
+    [plans]
+  );
+
+  if (plans.loading) {
+    return <LoadingOverlay isLoading={true} />;
+  }
+
+  if (!plans.result || plans.error !== null || !selectedPlan) {
```
biancadanforth, line 54: In Product and Checkout routes, I combined the `if` blocks for plan errors into one and conditionally load a new component, `PlanErrorDialog` to try making things DRYer, but I’m still repeating the `if` conditions in both routes (i.e. `if (!plans.result || plans.error !== null || !selectedPlan) {...}`). Is there a DRYer way to do this?

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.stories.tsx
```diff
@@ -25,7 +25,7 @@ const wait = (delay: number) =>
   new Promise((resolve) => setTimeout(resolve, delay));
 
 function init() {
-  storiesOf('routes/ProductV2/SubscriptionCreate', module)
+  storiesOf('routes/Product/SubscriptionCreate', module)
```
biancadanforth, line 28: I renamed this since I think we migrated to ProductV2 route as our default Product route a long time ago. I picked a couple stories, and they work under Product.

packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it('renders as expected for no selectedPlan', () => {
+    const subject = () => {
+      return render(
+        // @ts-ignore
+        <PlanErrorDialog
+          {...{
+            locationReload,
+            plans: { ...plans, result: PLANS },
```
biancadanforth, line 41: I get a mystifying TS error if I don't ignore, even though AFAICT this `plans` object matches the `FetchStateLoaded` type:
```
Type '{ result: Plan[]; error: null; loading: boolean; }' is not assignable to type 'FetchStateLoaded<Plan[]>'.
        Types of property 'loading' are incompatible.
        Type 'boolean' is not assignable to type 'false'.ts(2322)
```

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (window.onload) {
+      dispatchEvent(new Event('load'));
+    }
+
+    const checkoutEle = await findByTestId('checkout-route-container');
+    expect(checkoutEle).toBeInTheDocument();
+    expectNockScopesDone(apiMocks);
+  });
+
+  it('displays an error with invalid product ID', async () => {
```
biancadanforth, line 120: This test case  and one other is identical to that in the `Product` route. Is there a way to make this DRYer?

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const apiMocks = initApiMocks();
+    const { findByTestId, queryByTestId } = render(
+      <Subject productId="bad_product" />
+    );
+    await findByTestId('no-such-plan-error');
+    expect(queryByTestId('dialog-dismiss')).not.toBeInTheDocument();
+    expectNockScopesDone(apiMocks);
+  });
+
+  it('displays an error on failure to load plans', async () => {
```
biancadanforth, line 130: This test case  and one other is identical to that in the `Product` route. Is there a way to make this DRYer?

biancadanforth (commented):
To keep things DRY, I promoted the logic for getting `selectedPlan` to a new `lib/plan.ts` file that can be used by both `Product` and `Checkout` routes. To that end, I similarly created a new `PlanErrorDialog` component as well. It was less obvious how to dedupe some plan error test cases for the `Product` and `Checkout` routes, so I am open to suggestions!

packages/fxa-payments-server/src/App.tsx
meandavejustice: This sounds like a product question, it looks like all that `SignInLayout` does is determines if we want to hide the firefox logo or not. https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/components/AppLayout/index.tsx#L38

I'm not aware of the case where we need to hide the logo

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
meandavejustice: I'm not seeing a defined prop in `<SubscriptionCreate/>` for `accountActivated`. You are safe to remove.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
meandavejustice: This makes sense to me, I feel it is DRY enough for these cases, while still being clear to the next engineer in the file.

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
meandavejustice: I think it's okay for tests to not be DRY sometimes, it can help us have a piece of mind that the components used are still working in both places.

meandavejustice (changes requested):
Left a few comments

packages/fxa-payments-server/src/App.tsx
biancadanforth: Thanks. Yeah I don't think that `hideLogo` is ever `true` currently. I filed #9826 in our [Payments Technical Debt epic](https://mozilla-hub.atlassian.net/browse/FXA-3657).

packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
biancadanforth: I figured out a solution:
```diff
diff --git a/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx b/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
index 1d3c6bfe0..ed020d167 100644
--- a/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
@@ -4,9 +4,10 @@ import '@testing-library/jest-dom/extend-expect';
 
 import PlanErrorDialog from './index';
 import { PLANS } from '../../lib/mock-data';
+import { FetchState, Plan } from '../../store/types';
 
 const locationReload = () => {};
-const plans = {
+const plans: FetchState<Plan[], any> = {
   error: null,
   loading: false,
   result: null,
@@ -34,7 +35,6 @@ describe('PlanErrorDialog', () => {
   it('renders as expected for no selectedPlan', () => {
     const subject = () => {
       return render(
-        // @ts-ignore
         <PlanErrorDialog
           {...{
             locationReload,
```

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return <PlanErrorDialog locationReload={locationReload} plans={plans} />;
+  }
+
+  return (
+    <div data-testid="checkout-route-container">{selectedPlan?.plan_id}</div>
+  );
+};
+
+// TODO replace this with Redux hooks in component function body
+// https://github.com/mozilla/fxa/issues/3020
```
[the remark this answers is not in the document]
biancadanforth: Good catch; reading through that issue, it doesn't sound like that change would be beneficial (Les called it "yak shaving"). So I'll just remove every instance of this comment in each of the routes.

packages/fxa-payments-server/src/components/PlanErrorDialog/index.test.tsx
meandavejustice: @biancadanforth wonderful, thank you :)

meandavejustice (approved):

PR #9838, opened 2021-07-13
feat(auth): add tax ID to customer invoices

### LINKED ISSUE #9825: Add tax id as custom field to finalized invoice
For invoices in EU countries, we must add a tax id as a custom field to the invoice per:
https://stripe.com/docs/invoicing/customize#custom-fields

This custom field should be called 'Tax ID' and have the appropriate EU MOSS Tax ID for Mozilla (config option).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3734)

bbangert:
Because:

* We want to add a tax ID to customer invoices.

This commit:

* Adds the tax id as a custom field so it will show up on customers
  in VAT countries such using currency such as EUR and CHF.

Closes #9825

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -804,6 +804,15 @@ const conf = convict({
       format: String,
       doc: 'A shared secret to authenticate Stripe webhook requests',
     },
+    taxIds: {
+      doc: 'Mapping of currency to tax ID to show on invoices.',
+      format: Object,
+      default: {
+        EUR: 'EU1234',
+        CHF: 'CH1234',
+      },
+      env: 'TAXIDS',
+    },
     transactionalEmails: {
       // See also: https://jira.mozilla.com/browse/FXA-1148
       enabled: {
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -40,6 +40,8 @@ const CHARGES_RESOURCE = 'charges';
 const INVOICES_RESOURCE = 'invoices';
 const PAYMENT_METHOD_RESOURCE = 'paymentMethods';
 
+export const MOZILLA_TAX_ID = 'Tax ID';
+
 enum STRIPE_CUSTOMER_METADATA {
   PAYPAL_AGREEMENT = 'paypalAgreementId',
 }
@@ -97,6 +99,7 @@ export class StripeHelper {
   private stripe: Stripe;
   private redis: ioredis.Redis | undefined;
   private statsd: StatsD;
+  private taxIds: { [key: string]: string };
   public currencyHelper: CurrencyHelper;
 
   /**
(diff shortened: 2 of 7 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -336,8 +338,9 @@ export class StripeHelper {
     return this.stripe.subscriptions.create(
       {
         customer: customerId,
-        items: [{ price: priceId, tax_rates: taxRates }],
+        items: [{ price: priceId }],
         expand: ['latest_invoice.payment_intent'],
+        default_tax_rates: taxRates,
```
chenba, line 343: Unrelated fix?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -844,6 +848,31 @@ export class StripeHelper {
...
+      });
+    }
+  }
+
+  /**
+   * Returns the customers tax id if they have one.
+   **/
+  customerTaxId(customer: Stripe.Customer) {
+    return customer.invoice_settings.custom_fields?.find(
+      (field) => field.name === 'Tax ID'
```
chenba, line 872: One more time and I'm gonna need 'Tax ID' defined as a constant somewhere.  😀

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -844,6 +848,31 @@ export class StripeHelper {
...
+    );
+  }
+
```
chenba, line 875: The two functions added here don't seem to have tests?

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: It's a closely related fix, yes. The tax rates were in the wrong spot.

packages/fxa-auth-server/test/local/routes/subscriptions/paypal.js
```diff
@@ -181,6 +181,8 @@ describe('subscriptions payPalRoutes', () => {
       payPalHelper.agreementDetails = sinon.fake.resolves({
         countryCode: 'CA',
       });
+      stripeHelper.customerTaxId = sinon.fake.returns(false);
```
chenba, line 184: nit: This works for the purpose of the tests since `customerTaxId` is only used in `if` conditions so far, but it doesn't really match the type of the actual function.

packages/fxa-auth-server/config/index.ts
```diff
@@ -804,6 +804,15 @@ const conf = convict({
...
     },
+    taxIds: {
+      doc: 'Mapping of currency to tax ID to show on invoices.',
+      format: Object,
+      default: {
+        EUR: 'EU1234',
+        CHF: 'CH1234',
+      },
+      env: 'TAXIDS',
+    },
```
chenba, line 815: Update the deploy doc or whatever needed so we don't forget the actual tax id config values for prod?

chenba (changes requested):
I can't find tests for the new functions.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: They do now! 😁

packages/fxa-auth-server/test/local/routes/subscriptions/paypal.js
bbangert: Updated.

chenba (approved):
💶

PR #9851, opened 2021-07-14
feat(passwords): Add initial view for finishing setup of passwordless accounts

vbudhram:
## Because

- We need need a dedicated view for users to set the password on accounts created without a password

## This pull request

- Adds new screens
- Checks for valid query params
  - token (password forgot token)
  - code (account reset code)
  - email
  - redirectUrl
- If email are not valid redirects to standard account create page

Closes: https://github.com/mozilla/fxa/issues/9301

## Checklist

## Screenshots (Optional)

Based on UX at https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow

Screen Shot 2021-07-14 at 1 36 26 PM

## Testing

While all the parts to the passwordless account creation is not completed, we can still test the set password bits with these steps.

* Create an account but do not verify it
* Initiate a password reset flow and get the reset url
  * Copy the `token=sometoken&code=somecode` bits from the reset url
* Open url `http://localhost:3030/post_verify/finish_account_setup/set_password` and append params
  * email: user email
  * productName: some product name
  * token: password reset token
  * code: password reset code
  * redirectUrl: can currently only be `https://www.mozilla.org` or `https://accounts.firefox.com` domain
* Set password and get redirected to redirectUrl

### DIFF
--- a/packages/fxa-content-server/app/scripts/lib/router.js
+++ b/packages/fxa-content-server/app/scripts/lib/router.js
@@ -154,6 +154,9 @@ const Router = Backbone.Router.extend({
         type: VerificationReasons.RECOVERY_KEY,
       }
     ),
+    'post_verify/finish_account_setup/set_password': createViewHandler(
+      'post_verify/finish_account_setup/set_password'
+    ),
     'post_verify/cad_qr/get_started': createViewHandler(
       'post_verify/cad_qr/get_started'
     ),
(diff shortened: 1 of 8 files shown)


dannycoates (approved):
LGTM. Just the regex thing

Issue #9780, opened 2021-07-02
[FxA-Settings] Security code field from “Enter security code” page could be improved by limiting the number of digits allowed to 6 


data-sync-user:
**Build:**
Stage (Train 1.209.0)

**Affected platforms:**
Windows 10 64-bit
Android 11
iOS 14.6
macOS 10.14

**Preconditions:**
The user has the 2FA enabled.

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net and log in with an existing account.
1. At “Enter security code” page enter a code longer than 6 digits.

**Actual result:** 
A code longer than 6 digits can be entered.

**Expected result:** 
The code is limited to only 6-digits. (Compare with 2FA Step 1 of 3 page)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3706)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/15559">2FA security code field.gif</a>

data-sync-user:
➤ Wil Clouser commented:

Are you running into a situation where you needed to enter more than 6 characters?

data-sync-user:
➤ Wil Clouser commented:

I misread the bug - you’re saying we should be limiting the field to 6 to be consistent with the login experience.

BamBamboozled:
Should this be closed?

cknowles-admin:
FxA is no longer synchronizing all issues between Jira and Github.  We are closing open issues and will selectively synchronize in the future.

PR #9861, opened 2021-07-15
fix(FxA-Settings): only 6 digits can be entered as security code

JohannesBenedict:
Because help wanted: #9780 A code longer than 6 digits can be entered.

https://github.com/mozilla/fxa/issues/9780

- This PR solves issue


vbudhram (commented):
@JohannesBenedict Thanks for PR! When you get a chance could you rebase, we landed some test fixes so this should pass CI.

JohannesBenedict:
@vbudhram Thanks for reviewing. Sure no problem

dannycoates (approved):

Issue #9837, opened 2021-07-13
[subplat] Expiring CC Reminder Emails are No Long Being Sent


biancadanforth:
We upgraded to Stripe API 2020-08-27 in Train 206.  One of the changes in that version was: “The subscriptions property on Customers is no longer included by default.”  (ref: [https://stripe.com/docs/upgrades#2020-08-27](https://stripe.com/docs/upgrades#2020-08-27|smart-link) )  Our code was relying on that old behavior to detect whether a customer still has any subscriptions.  Since the subscriptions are no longer included by default, the code no longer send the expiring CC reminder email.

(Original bug description below.)

[https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/11514533/?referrer=github_plugin](https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/11514533/?referrer=github_plugin)

```java
Error: No subscriptions found for customer: cus_HhfEyh4bwvBXkb
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/stripe.js", line 1290, in StripeHelper.extractSourceDetailsForEmail
    '{snip} default.internalValidationError('extractSourceDetailsForEmail', customer, new Error(`No subscriptions found for customer: ${customer.id}`));
  ?, in runMicrotasks
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.js", line 355, in async StripeWebhookHandler.sendSubscriptionPaymentExpiredEmail
    const sourceDetails = await this.stripeHelper.extractSourceDetailsForEmail(paymentSource);
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.js", line 348, in async StripeWebhookHandler.handleCustomerSourceExpiringEvent
    const { uid, email } = await this.sendSubscriptionPaymentExpiredEmail(source);
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.js", line 75, in async StripeWebhookHandler.handleWebhookEvent
    await this.handleCustomerSourceExpiringEvent(request, event);
...
(6 additional frame(s) were not displayed)

WError: An internal validation check failed.
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/error.js", line 1464, in Function.AppError.internalValidationError
    return new AppError(
[...]

PR #9862, opened 2021-07-15
fix(subscriptions): include subscriptions when retrieving a customer

chenba:
Because:
 - a change in Stripe API removed subscriptions from a retrieved
   customer

This commit:
 - expand 'subscriptions' on a customer when `expandResource` is used to
   retrieve a Stripe customer

Closes: #9837

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -32,13 +32,13 @@ import error from '../error';
 import Redis from '../redis';
 import { CurrencyHelper } from './currencies';
 
-const CUSTOMER_RESOURCE = 'customers';
-const SUBSCRIPTIONS_RESOURCE = 'subscriptions';
-const PRODUCT_RESOURCE = 'products';
-const PLAN_RESOURCE = 'plans';
-const CHARGES_RESOURCE = 'charges';
-const INVOICES_RESOURCE = 'invoices';
-const PAYMENT_METHOD_RESOURCE = 'paymentMethods';
+export const CUSTOMER_RESOURCE = 'customers';
+export const SUBSCRIPTIONS_RESOURCE = 'subscriptions';
+export const PRODUCT_RESOURCE = 'products';
+export const PLAN_RESOURCE = 'plans';
+export const CHARGES_RESOURCE = 'charges';
+export const INVOICES_RESOURCE = 'invoices';
+export const PAYMENT_METHOD_RESOURCE = 'paymentMethods';
 
 export const MOZILLA_TAX_ID = 'Tax ID';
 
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2172,6 +2172,17 @@ export class StripeHelper {
...
     }
 
+    // We make an exception here for customers because we need to get the
+    // subscriptions for the customer.  The Stripe API stopped including
+    // subscriptions for a customer in version 2020-08-27; this ensures
+    // backwards compatibility for our code that's relying on that behavior.
+    if (resourceType === CUSTOMER_RESOURCE) {
+      // @ts-ignore
+      return this.stripe[CUSTOMER_RESOURCE].retrieve(resource, [
+        SUBSCRIPTIONS_RESOURCE,
```
bbangert, line 2182: This isn't the correct syntax for subscription expansion, but I think this should call the existing helper for loading a customer that's here because calling code here is quite likely to pass this object to our other functions which are faster if they preload the same things we frequently access on a customer object which our customer function loads.

bbangert (changes requested):

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: >This isn't the correct syntax for subscription expansion

D'oh!  Thanks for catching that.

> but I think this should call the existing helper for loading a customer 

By this you mean `StripeHelper.customer`?  I did look at that, and it does more than just getting the customer (in addition to a different interface requiring different arguments).  For example, code that calls `expandResource` to get a customer would explicitly check for `customer.deleted` on the return value, where as `fetchCustomer` (what `StripeHelper` ultimately calls to get a customer) handles `customer.deleted` and returns undefined.

For purpose of the expiring CC reminder bug, I could call `StripeHelper.customer` in `StripeHelper.extractSourceDetailsForEmail` and avoid this altogether.  What do you think?

bbangert: That's a good point, I think calling that in the other method instead of using expand resource on customers is a good approach. Might be useful to see if any other caller on `expandResource` uses customer, and if not, just remove it as a valid expand option in favor of always using the other method.

chenba: > Might be useful to see if any other caller on expandResource uses customer, and if not, just remove it as a valid expand option in favor of always using the other method.

I already did, and there are a few.  None of them rely on the 'subscriptions' property though, so we've got that going for us at least.

chenba: >  I think calling that in the other method instead of using expand resource on customers is a good approach. 

Looks like this isn't possible because of a chicken and egg problem. 😢 `Stripe.customer` requires a uid and an email addresses, but we won't have those until we get the customer.

chenba:
@bbangert I fixed the expand option syntax bug you pointed out in the latest commit.  I didn't use `StripeHelper.customer` because of https://github.com/mozilla/fxa/pull/9862#discussion_r670802531

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2172,6 +2172,17 @@ export class StripeHelper {
       throw error;
     }
 
+    // We make an exception here for customers because we need to get the
+    // subscriptions for the customer.  The Stripe API stopped including
+    // subscriptions for a customer in version 2020-08-27; this ensures
+    // backwards compatibility for our code that's relying on that behavior.
+    if (resourceType === CUSTOMER_RESOURCE) {
+      // @ts-ignore
```
bbangert, line 2180: Sorry, should've mentioned this before, but why the ts-ignore? Since its a customer here, might as well just do the customer retrieve the normal way so the type-checker can help out?

bbangert (approved):
LGTM!

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: The same reason for the @ts-ignore on line 2186, which is the generic type param `T` not matching the actual return type.

Issue #9863, opened 2021-07-15
Investigate why reminder emails are not being sent on CI


vbudhram:
Our verification reminders test seem to constantly fail in CI. There hasn't been any meaningful changes to this file that I believe might be causing this. Locally the tests pass, however on CI they keep polling for the verification email which never arrives.

I suspect something with running the command might have broke. In the meantime, lets disable the tests so we can debug them.

Ref: https://app.circleci.com/pipelines/github/mozilla/fxa/21126/workflows/431e04fc-f970-4c56-9cfb-ac662cb6d046/jobs/252227/tests

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3752)

cknowles-admin:
FxA is no longer synchronizing all issues between Jira and Github.  We are closing open issues and will selectively synchronize in the future.

PR #9864, opened 2021-07-15
fix(tests): disable verification reminders functional tests

vbudhram:
Disable these tests while we debug https://github.com/mozilla/fxa/issues/9863


chenba (approved):

vbudhram:
Also reverted change in https://github.com/mozilla/fxa/issues/9866 to get CI passing again.

PR #9865, opened 2021-07-15
fix(auth): update vat populate script to handle change

bbangert:
Because:

* The tax ID may change, requiring all customers to get the new default
  custom field.
* An incorrect tax rate might be deployed causing new customers to have
  the wrong tax added to their subscription.

This commit:

* Will check the prior tax id value to the current one so updates will
  be made if they don't match.
* Checks the current tax on the subscription and updates it if
  necessary.

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -869,6 +869,13 @@ export class StripeHelper {
   }
 
   /**
+   * Returns the correct tax id for a customer.
+   */
+  getTaxIdForCustomer(customer: Stripe.Customer) {
+    return this.taxIds[customer.currency?.toUpperCase() ?? ''];
+  }
+
+  /**
    * Returns the customers tax id if they have one.
    **/
   customerTaxId(customer: Stripe.Customer) {
--- a/packages/fxa-auth-server/scripts/populate-vat-taxes.ts
+++ b/packages/fxa-auth-server/scripts/populate-vat-taxes.ts
@@ -37,14 +37,12 @@ export async function init() {
           continue;
         }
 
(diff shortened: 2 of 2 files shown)


packages/fxa-auth-server/scripts/populate-vat-taxes.ts
```diff
@@ -82,6 +80,12 @@ export async function init() {
           continue;
         }
 
+        // Does the subscription already have the right tax rate attached?
+        const existingTaxRate = sub.default_tax_rates?.shift();
```
chenba, line 84: Not really an issue here given the scope of `sub`, but not a fan of using `shift` as it mutates the array.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -868,6 +868,13 @@ export class StripeHelper {
     }
   }
 
+  /**
+   * Returns the correct tax id for a customer.
+   */
+  getTaxIdForCustomer(customer: Stripe.Customer) {
+    return this.taxIds[customer.currency?.toUpperCase() ?? ''];
+  }
```
chenba, line 876: Could use some tests?

chenba (approved):
LGTM r+ with comments

packages/fxa-auth-server/scripts/populate-vat-taxes.ts
bbangert: Me neither, but the `?` syntax is fine with that, but it didn't like array access. 🤷🏻

vbudhram (commented):
If you rebase I think the failing tests should pass.

bbangert:
Woops, merged instead of rebased. 🤷🏻

Issue #9874, opened 2021-07-16
[FxA-Settings] Secondary email warning message issue


data-sync-user:
**Build:**
Production (Train 1.210.2)

**Affected platforms:**
Windows 10 64-bit
macOS 10.14.6
Android 11
iOS 14.4

**Steps to reproduce:**

1. Navigate to https://accounts.firefox.com and log in with an existing account.
1. Go to “Secondary email” section and add and verify a secondary email.
1. Then remove the secondary email.
1. Repeat step 2 and 3 other 4 times.
1. Add again a secondary email.
1. Observe the message displayed.

**Actual result:** 
There is an issue regarding the warning message displayed.

**Expected result:** 
The warning message is correctly displayed.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3756)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/15679/warning+issue.gif">warning issue.gif</a>

data-sync-user:
➤ Bianca Oltean commented:

Vijay Budhram I verified this issue on Production [Train 211.2] on Windows 10 and it is still reproducible. Do you know if this bug missed the Train?

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 1.212.0] under Windows 10 64 bit and Android 11. The warning message is correctly displayed. See new situation:

!new situation.PNG|width=472,height=271!

PR #9876, opened 2021-07-16
fix(errors): Fix issue where rate limit was not being localized

vbudhram:
## Because

- We were not localizing text in add secondary email page

## This pull request

- Passes the `retryAfter` value to fluent
- Also updates our comment to mention the momentjs automatically adds the `in` string when a timestamp is in the future. Ref https://momentjs.com/docs/#/displaying/fromnow/

Closes: https://github.com/mozilla/fxa/issues/9874

## Checklist

## Screenshots (Optional)
Screen Shot 2021-07-16 at 12 22 06 PM

### DIFF
--- a/packages/fxa-settings/src/components/PageSecondaryEmailAdd/index.tsx
+++ b/packages/fxa-settings/src/components/PageSecondaryEmailAdd/index.tsx
@@ -8,7 +8,7 @@ import FlowContainer from '../FlowContainer';
 import VerifiedSessionGuard from '../VerifiedSessionGuard';
 import { isEmailValid } from 'fxa-shared/email/helpers';
 import { useAccount, useAlertBar } from 'fxa-settings/src/models';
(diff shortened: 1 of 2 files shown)


dannycoates (approved):

peiying2:
I don't have the technical know-how to explain why it is best to put "in" outside the variable. Or maybe a different way to  ask is, why do we not put *in* outside the variable? I always thought variable is a number usually with corresponding unit. Is that the case?

vbudhram:
> I don't have the technical know-how to explain why it is best to put "in" outside the variable. Or maybe a different way to ask is, why do we not put _in_ outside the variable? I always thought variable is a number usually with corresponding unit. Is that the case?

I think normally that is the case, `in` would be outside. I haven't dug very keep into our time library [momentJS](https://momentjs.com/docs/#/displaying/fromnow/), but when the time is in the future it will append `in`. Similarly this library will also append `ago` if time was in the past ex. `7 seconds ago`. We could use another library but I think that might be a bit more work?

Issue #9792, opened 2021-07-07
[subplat] Make Valid Access Token Optional


data-sync-user:
Currently a valid access token is required to access any pages on Payments.  However, we need to allow potential new customers to access the new sign up and subscribe flow without signing in.

We should:

* Not immediately redirect when a valid access token was not found; leave this to the routes
* Create a way for the components to check if there was a valid access token on page load
* Set up redirect on no token for the current routes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3713)

data-sync-user:
➤ Barry Chen commented:

This affects both the products (new subscriptions) and subscriptions management routes.  However, it should not be noticeable when going through the current flows.  In order to test this issue, you’d need to use direct Payments urls (vs going through content-server first in a normal flow).

For example, on Stage: https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/products/prod_FiJ42WCzZNRSbS ( https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/products/prod_FiJ42WCzZNRSbS ) and https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/subscriptions ( https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/subscriptions )

If you are not already signed in, you should be asked to sign in.

data-sync-user:
➤ Barry Chen commented:

Added a qa+ label since we should verifiy that this does not cause any regressions.

PR #9881, opened 2021-07-16
refactor(payments): check for access token in routes

chenba:
Because:
 - we want some routes in Payments to be accessible without a valid
   access token

This commit:
 - stop immediately redirecting to Settings when there is no valid
   access token on a Payments page load
 - store the access token in the AppContext
 - check for a valid access token in the products and subscriptions
   routes and redirect to the correct content-server route (instead of
   Settings)

Closes: #9792

### DIFF
--- a/packages/fxa-payments-server/src/App.tsx
+++ b/packages/fxa-payments-server/src/App.tsx
@@ -36,6 +36,7 @@ type RouterOverride = () => (props: { children: ReactNode }) => JSX.Element;
 export type AppProps = {
   config: Config;
   store: Store;
+  accessToken: string | null;
   queryParams: QueryParams;
   matchMedia: (query: string) => boolean;
   matchMediaDefault: (query: string) => MediaQueryList;
@@ -50,6 +51,7 @@ export type AppProps = {
 export const App = ({
   config,
   store,
+  accessToken,
   queryParams,
   matchMedia,
   matchMediaDefault,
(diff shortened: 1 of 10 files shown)


meandavejustice:
@chenba Could you add some steps to reproduce/routes to try

chenba:
> 
> 
> @chenba Could you add some steps to reproduce/routes to try

Thanks for reminder.  I'll add a note in the issue; I'm sure QA would like to know as well.

chenba:
https://github.com/mozilla/fxa/issues/9792#issuecomment-882664162

meandavejustice:
@chenba 

https://user-images.githubusercontent.com/1844554/126207607-b2ab46f2-60cb-490b-921d-26073ec599fa.mp4

I was able to successfully input product page url, sign in, and be redirected to the correct page. But I'm getting an error when trying to visit `https://localhost:3031/subscriptions` which if I understand your comment (https://github.com/mozilla/fxa/issues/9792#issuecomment-882664162) I should be able to navigate to.

Let me know if I am missing something here

chenba:
@meandavejustice Thanks for including the video.  Here is what's happening: you are already signed in (since you can see the subscription payment form), but you don't have any active subscriptions, so the subscriptions management page redirects you back to setting, there you get a completely unrelated error (looks like CORS).

meandavejustice (approved):

PR #9902, opened 2021-07-20
feat(auth): add /account/stub route

### LINKED ISSUE #9327: Create FxA account route for no password creation
To streamline the checkout flow, the checkout page will request an account to be created for a new email address. This route should work similarly to the existing create route, except instead of returning a sessionToken for a logged in user, it will return an auth token for use with subscription routes for subscription creation.

This route should have a placeholder TODO, marking where the 'Finish Setting up your Account' email will be triggered.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3599)

dannycoates:
Enables subscription services to create an account without a password.

closes #9327

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -22,6 +22,8 @@ import emailUtils from './utils/email';
 import requestHelper from './utils/request_helper';
 import { determineSubscriptionCapabilities } from './utils/subscriptions';
 import validators from './validators';
+import { generateAccessToken } from '../oauth/grant';
+import { getClientById } from '../oauth/client';
 
 const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema;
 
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -502,6 +504,54 @@ export class AccountHandler {
...
+    const { email, clientId } = request.payload as any;
+    await this.customs.check(request, email, 'accountCreate');
+
+    if (this.OAUTH_DISABLE_NEW_CONNECTIONS_FOR_CLIENTS.has(clientId)) {
+      throw error.disabledClientId(clientId);
+    }
+
+    const client = await getClientById(clientId);
+
+    await this.deleteAccountIfUnverified(request, email);
```
dannycoates, line 518: This leaves a small window where the account could be deleted while the user is in the process of subscribing. If it were to happen the most common result would be adding the subscription would fail but there may be an **extremely** small window where the sub could be added and then account deleted. I don't think this will happen in real world use though, so I'm not very concerned about making it race-safe.

vbudhram: Seems reasonable to me, I don't think this will be a common case either.

vbudhram (approved):
@dannycoates LGTM, is this something that we will need to expose in the fxa-auth-client too?

dannycoates:
> is this something that we will need to expose in the fxa-auth-client too?

maybe. @chenba how do you plan on calling this?

chenba:
> 
> 
> > is this something that we will need to expose in the fxa-auth-client too?
> 
> maybe. @chenba how do you plan on calling this?
> 

Was not planning on using fxa-auth-client to call it.

Issue #9786, opened 2021-07-06
[subplat] Add Firefox Newsletter Opt-in Checkbox for New User Checkout Flow


data-sync-user:
'mozilla-and-you' is the newsletter.  That need to be included in the request when creating the new user.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3710)

data-sync-user:
➤ Barry Chen commented:

This will likely require multiple small patches, so, at the very least, we can work Payments frontend changes concurrently as a team.

data-sync-user:
➤ Bianca Danforth commented:

Okay I just spoke with Lauren Niolet and Paul McLanahan re: how to verify the newsletter signup (thank you):

  * Firstly, “mozilla-and-you” is indeed the correct newsletter slug.
  * Assuming we are consistently pointing to the same Basket environment in FxA Settings for the “Email Communications” link as when we post to sign people up to the newsletter, QA can verify that they are subscribed by looking in the “Email Communications” page and seeing that “Firefox News” is checked.
    * basket-stage: basket.allizom.org
      * “Email Communications” page will have domain http://www.mozilla.org ( http://www.mozilla.org|smart-link ) , as data will go to prod CTMS and Acoustic

    * basket-prod: basket.mozilla.org
    * basket-dev: basket-dev.allizom.org
      * “Email Communications” page will have domain http://www-dev.allizom.org ( http://www-dev.allizom.org|smart-link ) , as data will go to stage CTMS

  * Notably, while the first newsletter signup for an email address in other contexts does receive a confirm email, this is not the case for signup through FxA (this is in the Marketing team’s backlog).
    * I’m not 100% on this, but I think this is because FxA users are automatically signed up for the Firefox Accounts Tips newsletter when they create a Firefox Account.
    * Tony Cinotto , Marketing says this isn’t a small lift, but I am looping you in here in case this is something worth prioritizing in the future.

So what I need to do is verify that we are consistently pointing to the same Basket environment in FxA stage (either basket-stage or basket-dev) and similarly for FxA prod (basket-prod).

data-sync-user:
➤ Tony Cinotto commented:

Bianca Danforth just to confirm, a confirm email confirms that you want to receive these newsletter emails or you can progressively opt out? for our case, a user wouldn’t receive this confirm email and would start receiving the newsletter?

data-sync-user:
➤ Bianca Danforth commented:

[...]

Issue #2814, opened 2019-10-09
Eliminate the use of ResumeTokens


shane-tomlinson:
ResumeTokens are a way for us to pass information from a signup/signin flow to a verification flow without having to store data in the database. A resume token is a JSON bundle that's been btoa'd and contains a [whole bunch of fields](https://github.com/mozilla/fxa/blob/c759ba95c659d0bc527a2bff34a2ffcf9f83c1cb/packages/fxa-content-server/app/scripts/models/resume-token.js#L14) we need to propagate whenever a user clicks on a link from a verification email. They are a useful abstraction that adds a lot of complexity to the code base. I'd like to rip them out.

If we fully migrate to signup and signin codes, link based verification will only be used for verification reminders and password resets. Verification reminders don't use resume tokens, however password reset emails do. :sigh:

Now, if we could verify password resets with codes as well, then only verification reminders would contain a link, and again, those don't use ResumeTokens.

Needed to do this:

* [ ] Fully migrate signup to verification codes instead of links
* [ ] Fully migrate signin to verification codes instead of links
* [ ] Fully migrate password reset to verification codes instead of links
* [ ] Fully migrate add secondary email flow to verification codes instead of links

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-421)
┆Issue Number: FXA-421

data-sync-user:
➤ Wil Clouser commented:

Thank you for filing this issue. This is not currently on our roadmap, and in an effort to focus our work, we are closing old issues that we are unlikely to be closed in the future. Thanks again.

PR #9912, opened 2021-07-21
feat(subscriptions): add fn for creating resume tokens

chenba:
Because:
 - we need to, at a minimum, use a "resume token" to pass the
   newsletter the new user opted into on the new funnel flow

This commit:
 - move shareable resume token related code into fxa-shared
 - add a function in Payments for creating resume tokens

Part of #9786


chenba:
We'll be taking a different approach that won't require resume tokens.  Also, https://github.com/mozilla/fxa/issues/2814.

Issue #9260, opened 2021-06-21
[emails M1] FxA emails docs update


data-sync-user:
Assuming documentation isn't written alongside most PRs in this epic, the documentation around our emails could likely use an update. Some of this will be in the `fxa-auth-server` README, but if we end up creating a new directory for the templates, some of the other points can likely go in a README in that new directory.

We'd like to have:

* A quick explainer of the current state of the old system and new system behind feature flags (this should be really brief, as it's going to be removed later once all of the templates are in production and feature flags removed)
* A link to the ADR
* A small explainer of the stack (MJML + EJS + Fluent) with relevant links
* CSS guidelines with an explanation around the globals, the Tailwind-esq variables/mixins, using !important, etc.
* How to build Storybook and relevant docs there
* Anything else that may be relevant at this point

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3510)

Yogita98:
Right okay, I understood most of the requirements but would like a little more elaboration on this:

> * A quick explainer of the current state of the old system and new system behind feature flags (this should be really brief, as it's going to be removed later once all of the templates are in production and feature flags removed)

PR #9923, opened 2021-07-22
docs(fxa-auth-server): Update docs to include information about new e…

Yogita98:
…mail templates

## Because

- We revamped auth server email templates and want to document the changes

## This pull request

- Aims to update auth-server docs to include information about things done as part of revamping the emails

Closes: #9260 

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/README.md
```diff
@@ -147,6 +147,64 @@ This saves the HTML template into `/templates`. Then make changes to the `.txt`
...
+  <mj-include path="./lib/senders/emails/css/button/index.css" type="css" css-inline="inline" />
+  <mj-section>
+    <mj-column>
+      <mj-button css-class="primary-button"><%= buttonText %></mj-button>
+    </mj-column>
+  </mj-section>
+`;
+```
+
+- The emails for which the mjml feature flag is enabled can be rendered to disk using the `yarn write-emails` command
```
LZoog, line 185: Is this supposed to be `yarn write-templates`? I'm sort of confused by the similar naming between old and new templates. Maybe we can have another section to point out which commands are for the old/current templates along with where the old/current and new templates are located since if someone else goes to pick up any of this work it might be confusing for them too to get started.

Yogita98:
The docs could be previewed [here](https://github.com/Yogita98/fxa/blob/auth-server-docs-update/packages/fxa-auth-server/README.md)

packages/fxa-auth-server/README.md
```diff
@@ -147,6 +147,84 @@ This saves the HTML template into `/templates`. Then make changes to the `.txt`
...
+    </mj-column>
+  </mj-section>
+`;
+```
+
+The emails for which the mjml feature flag is enabled can be rendered to disk using the `yarn write-emails` command.
+
+#### Styles
+
+Another advantage of using MJML for emails is that it inlines styles with the compliled HTML elements making emails compatible with all the mail clients. Currently, we are using `scss` stylesheets which get compiled down to css and are included in the mjml templates. We are maintaining two kinds of stylesheets for our emails, a global stylesheet which contains variables and shared classes, and template-specific stylesheet for the styles scoped to the respective template/partial. Although there are lot of benefits of using stylesheets with MJML, one caveat that we'd like to highlight is the inevitible use of `!important` syntax with some of the styles. While compiling down the templates, mjml internally adds some default styles to the html elements, so if we were to add our custom styles on top of it we may have to use `!important` with them to override the default ones.
```
LZoog, line 187: My only nitpick here is maybe this:
> We are maintaining two kinds of stylesheets for our emails, a global stylesheet which contains variables and shared classes, and template-specific stylesheet for the styles scoped to the respective template/partial.

I think this may quickly become outdated (I think it may already be, with you extracting the rtl styles in the other PR?). Maybe say:

> We are maintaining shared stylesheets with common styles and variables, as well as template-specific stylesheet for the styles scoped to the respective template/partial.

LZoog (commented):
This is also looking good to me! I think it may be good to mention how the TXT files are generated, as well as mention those are displayed to users when they prefer not to see rich HTML emails.

I would go ahead and r+ this, but I think it might be good to wait until the other two issues in this milestone are closed to make sure we don't want to add or modify anything here 🙂

packages/fxa-auth-server/README.md
```diff
@@ -147,6 +147,84 @@ This saves the HTML template into `/templates`. Then make changes to the `.txt`
...
+```js
+<p data-l10n-id="update-application-info" data-l10n-args="{'version': '60.0'}">
+  You are using Firefox Version: 60.0.
+  <span class="bold">
+    Please, read the <a href="http://www.mozilla.org/privacy">privacy policy</a>
+  </span>
+</p>
+```
+
+By default emails render from left to right which could hamper the accessibility of emails for some locales so along with localizing the emails, we took care of rendering the emails from right to left for rtl localtes and vice versa.
```
jodyheavener, line 220: nit: "rtl _locales_ and"

packages/fxa-auth-server/README.md
```diff
@@ -147,6 +147,84 @@ This saves the HTML template into `/templates`. Then make changes to the `.txt`
...
+  <mj-include path="./lib/senders/emails/css/button/index.css" type="css" css-inline="inline" />
+  <mj-section>
+    <mj-column>
+      <mj-button css-class="primary-button"><%= buttonText %></mj-button>
+    </mj-column>
+  </mj-section>
+`;
+```
+
+The emails for which the mjml feature flag is enabled can be rendered to disk using the `yarn write-emails` command.
```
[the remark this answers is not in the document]
jodyheavener: Same with `storybook` and `Storybook`

jodyheavener (changes requested):
A handful of nitpicks and minor comments, but generally this looks great!

jodyheavener (approved):
🥳

packages/fxa-auth-server/README.md
```diff
@@ -147,6 +147,84 @@ This saves the HTML template into `/templates`. Then make changes to the `.txt`
...
+
+#### MJML Feature Flag for Testing
+
+The auth-server has an `mjml` configuration value that designates which email templates have been converted to the new stack ( `mjml.templates` array) and controls which user emails will receive those templates ( `mjml.enabledEmailAddress` regex). The auth-server exposes a feature flag method to check these values and send the correct rendered template accordingly (old or new) which allows for testing across environments.
+
+This flag and logic around it can be removed, as well as old templates and documentation, once all templates have been converted, verified by QA, and tested in production.
```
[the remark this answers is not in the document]
LZoog: Oh yeah, that would be great. Thanks @jodyheavener! We can stick that in the 4th milestone for now, though I imagine we'll scoot that out somewhere else later.

LZoog (approved):
This LGTM too! Lots of info packed in here. 🙂

Issue #9385, opened 2021-06-21
[new user checkout] check fxa account exists on email input


data-sync-user:
On user email input, check if existing fxa account.

I'm not seeing a ticket for the endpoint to actually check the existence on fxa account. cc/[~accountid:5e3400fb6bd5c40ca395d189] to add the endpoint we need to hit for this check.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3663)

data-sync-user:
➤ Dave Justice commented:

Benjamin Bangert for adding the route, thank you

data-sync-user:
➤ Benjamin Bangert commented:

Route for verifying account exists: https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/api.md#post-accountstatus ( https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/api.md#post-accountstatus )

PR #9913, opened 2021-07-21
task(fxa-payments-server) Add email validation

meandavejustice:
## Because

- We need to validate the new user email

## This pull request

- Adds validation checks to the email input on new user form
- Adds a check to ensure the confirm input matches the original email value

## Issue that this pull request solves

Closes: #9354

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [x] I have verified that my changes render correctly in RTL (if appropriate).

## Other information (Optional)

Luckily our email validation method was in `fxa-shared` which saved a lot of time.

PR #9924, opened 2021-07-22
9385 check account existance

meandavejustice:
## Because

- We want to know if a user in this flow has an account and just isn't signed in.

## This pull request

- Adds the check for the user input email to see if they have an existing account.
- (Sign up link in error tooltip will be added in https://mozilla-hub.atlassian.net/browse/FXA-3664)

Closes: #9385

## Checklist

Additional Information:

- This builds on https://github.com/mozilla/fxa/pull/9913, so is dependent on that being merged first
- The commit with relevant changes 👉 https://github.com/mozilla/fxa/pull/9924/commits/678c6f720874c250ed293fdad252bf2c05d97c17

### DIFF
--- a/packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
@@ -5,7 +5,7 @@ import { isEmailValid } from '../../../../fxa-shared/email/helpers';
 
 import shieldIcon from './images/shield.svg';
 
-import { Form, Input, Checkbox, OnValidateFunction } from '../fields';
+import { Form, Input, Checkbox } from '../fields';
 import {
   State as ValidatorState,
   useValidatorState,
@@ -13,15 +13,18 @@ import {
 } from '../../lib/validator';
 
 import './index.scss';
+import { config } from '../../lib/config';
 
 export type NewUserEmailFormProps = {
   getString: (id: string) => string;
+  checkAccountExists: Function;
   validatorInitialState?: ValidatorState;
   validatorMiddlewareReducer?: ValidatorMiddlewareReducer;
 };
 
 export const NewUserEmailForm = ({
   getString,
+  checkAccountExists,
   validatorInitialState,
   validatorMiddlewareReducer,
 }: NewUserEmailFormProps) => {
@@ -52,12 +55,13 @@ export const NewUserEmailForm = ({
           placeholder="foxy@mozilla.com"
           required
           spellCheck={false}
-          onValidate={(value, focused) => {
+          onValidatePromise={(value, focused) => {
             return emailInputValidationAndAccountCheck(
               value,
               focused,
               setEmailInputState,
-              getString
+              getString,
+              checkAccountExists
             );
           }}
         />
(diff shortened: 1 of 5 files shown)


packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -113,8 +113,16 @@ export type OnValidateFunction = (
...
+export type OnValidateFunctionPromise = (
+  value: any,
+  focused: boolean,
+  props: FieldProps,
+  getString?: Function
+) => Promise<{ value: any; valid: boolean | undefined; error: any }>;
+
 type InputProps = {
   onValidate?: OnValidateFunction;
+  onValidatePromise?: OnValidateFunctionPromise;
```
meandavejustice, line 125: Added this to avoid breaking functionality across the codebase. This works now, but I wonder if it would be better to have `onValidate` accept both it's current return signature and a promise wrapped version. I tried a couple of different methods but didn't figure it out fully. So took this direction to keep it moving since we have an august deadline. Would be happy to try suggestions here, or create a tech debt/follow up ticket to address it.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -104,28 +108,53 @@ export const NewUserEmailForm = ({
...
+  let hasAccount = false;
   setEmailInputState(value);
   if (isEmailValid(value)) {
-    // check if we have an existing account here
     valid = true;
+    await checkAccountExists(value).then((response: { exists: boolean }) => {
+      if (response.exists) {
+        hasAccount = true;
+      }
+    });
```
dannycoates, line 139: try not to mix `await` and `then`

packages/fxa-payments-server/src/components/fields/index.tsx
dannycoates: We ought to clean this up, but im ok with it for now. Can you create a maintenance issue for it?

dannycoates (approved):

packages/fxa-payments-server/src/components/fields/index.tsx
chenba: I believe this should work with the return value of the validation function being a sum type, i.e.:

```
export type ValidationResult = {
  value: any;
  valid: boolean | undefined;
  error: any;
};
export type OnValidateFunction = (
  value: any,
  focused: boolean,
  props: FieldProps,
  getString?: Function
) => ValidationResult | Promise<ValidationResult>;
```

And then await all the `onValidate` calls.

meandavejustice: https://mozilla-hub.atlassian.net/browse/FXA-3802

chenba: Did my suggestion not work?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: lol fixed, not sure what I was thinking there

packages/fxa-payments-server/src/components/fields/index.tsx
meandavejustice: @chenba I tried that initially before opening the pr and I was still having lots of test failures, so will address this in the follow up issue when we aren't working toward a deadline.

PR #9931, opened 2021-07-23
feat(content): redirect to new checkout route for unauthenticated users

### LINKED ISSUE #9793: [subplat][content-server] Redirect to the new checkout payments route
We plan to add a new payments route for the new user checkout flow in FXA-3486. The content-server should be updated to redirect to this new route if the user is not signed in (i.e. they don’t have an FxA account, or they are logged out).

Barry previously landed a patch that allows the content server to redirect to payments without requiring an accessToken (behind a flag) – FXA-3488. This is the area of the code that will need to be modified.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3714)

biancadanforth:
Because:

* There is a new Checkout route on the payments server for unauthned users that will create a FxA account (if none exists) at the same time as creating a new subscription.

This commit:

* Conditionally sets the redirectPath for redirecting to the payments server based on a user's authn state and the subscriptionsConfig.allowUnauthenticatedRedirects flag.
* The latter will ensure that we do not redirect unauthned users in prod to the new checkout route until we are ready to do so.

Closes #9793

## Checklist

I filed [FXA-3790](https://mozilla-hub.atlassian.net/browse/FXA-3790) to set `allowUnauthenticatedRedirects` to `true` in `prod` as it is [currently only enabled in `dev` and `stage`](https://github.com/mozilla-services/cloudops-deployment/pull/4335) for testing (thanks Barry).

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
+++ b/packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
@@ -15,7 +15,6 @@ class SubscriptionsProductRedirectView extends FormView {
   initialize(options) {
     this._currentPage = options.currentPage;
     this._productId = this._currentPage.split('/').pop();
-    this._queryParams = Url.searchParams(this.window.location.search);
     this.relier.set('subscriptionProductId', this._productId);
 
     this._subscriptionsConfig = {};
@@ -23,22 +22,38 @@ class SubscriptionsProductRedirectView extends FormView {
       this._subscriptionsConfig = options.config.subscriptions;
     }
 
-    this.mustAuth =
-      !!this._queryParams.signin ||
-      !this._subscriptionsConfig.allowUnauthenticatedRedirects;
-
     // Flow events need to be initialized before the navigation
     // so the flow_id and flow_begin_time are propagated
     this.initializeFlowEvents();
   }
 
+  render() {
+    const account = this.getSignedInAccount();
+    return Promise.resolve()
+      .then(() => {
+        const isSignedInPromise =
+          account?.isSignedIn() || Promise.resolve(false);
+        return isSignedInPromise.then((isSignedIn) => {
+          this._queryParams = Url.searchParams(this.window.location.search);
+          const forceSignin = !!this._queryParams.signin;
+          this._redirectPath =
+            !isSignedIn &&
+            this._subscriptionsConfig.allowUnauthenticatedRedirects &&
+            !forceSignin
+              ? `checkout/${this._productId}`
+              : `products/${this._productId}`;
+          this.mustAuth = this._redirectPath.startsWith('products');
+        });
+      })
+      .finally(() => FormView.prototype.render.bind(this)());
+  }
+
   afterRender() {
     delete this._queryParams.signin;
-    const redirectPath = `products/${this._productId}`;
     return PaymentServer.navigateToPaymentServer(
       this,
       this._subscriptionsConfig,
-      redirectPath,
+      this._redirectPath,
       this._queryParams
     );
   }
--- a/packages/fxa-content-server/app/scripts/lib/payment-server.js
+++ b/packages/fxa-content-server/app/scripts/lib/payment-server.js
@@ -7,7 +7,7 @@
 
 import Url from './url';
 
-const allowedUnauthenticatedRoutes = ['product'];
+const allowedUnauthenticatedRoutes = ['checkout'];
 const isAllowedUnauthenticatedRoute = (redirectPath) =>
   allowedUnauthenticatedRoutes.some((route) => redirectPath.startsWith(route));
 
--- a/packages/fxa-content-server/app/tests/spec/lib/payment-server.js
+++ b/packages/fxa-content-server/app/tests/spec/lib/payment-server.js
@@ -89,8 +89,8 @@ describe('lib/payment-server-redirect', () => {
     });
   });
 
-  it('redirects to the product route when there is no account', () => {
-    const REDIRECT_PATH = 'product/bleepbloop';
+  it('redirects to the checkout route when there is no account', () => {
+    const REDIRECT_PATH = 'checkout/bleepbloop';
     view.getSignedInAccount = sinon.stub().returns(null);
     PaymentServer.navigateToPaymentServer(
       view,
@@ -105,8 +105,8 @@ describe('lib/payment-server-redirect', () => {
     );
   });
 
-  it('redirects to the product route when account is not signed in', () => {
-    const REDIRECT_PATH = 'product/bleepbloop';
+  it('redirects to the checkout route when account is not signed in', () => {
+    const REDIRECT_PATH = 'checkout/bleepbloop';
     account._fxaClient = {
       isSignedIn: sinon.stub().resolves(false),
     };
--- a/packages/fxa-content-server/app/tests/spec/views/subscriptions_product_redirect.js
+++ b/packages/fxa-content-server/app/tests/spec/views/subscriptions_product_redirect.js
@@ -24,6 +24,7 @@ describe('views/subscriptions_product_redirect', function () {
   let windowMock;
   let config;
   let notifier;
+  let getSignedInAccountStub;
 
   function render() {
     return view.render().then(() => view.afterVisible());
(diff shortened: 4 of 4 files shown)


packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -15,31 +15,43 @@ class SubscriptionsProductRedirectView extends FormView {
...
+    return Promise.resolve()
+      .then(() => {
+        const isSignedInPromise =
+          account?.isSignedIn() || Promise.resolve(false);
+        isSignedInPromise.then((isSignedIn) => {
+          this._redirectPath =
+            !isSignedIn &&
+            this._subscriptionsConfig.allowUnauthenticatedRedirects
+              ? `checkout/${this._productId}`
+              : `products/${this._productId}`;
```
biancadanforth, line 41: This is still based on the feature flag, so that we don't automatically redirect unauthned users to the new checkout route on payments in prod until it's ready.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -15,31 +15,43 @@ class SubscriptionsProductRedirectView extends FormView {
...
-    this._queryParams = Url.searchParams(this.window.location.search);
     this.relier.set('subscriptionProductId', this._productId);
 
     this._subscriptionsConfig = {};
     if (options && options.config && options.config.subscriptions) {
       this._subscriptionsConfig = options.config.subscriptions;
     }
 
-    this.mustAuth =
-      !!this._queryParams.signin ||
```
biancadanforth, line 27: I removed the `signin` queryParam here and in the payments server (added largely in PR #9781 ), since IIUC that was introduced under the assumption that we would have the one "product" route on payments, while we later decided to have two payments routes, "product" for authned users and "checkout" for unauthned users.  @chenba , would you agree?

biancadanforth (commented):
Added a couple of questions/comments inline for the review.

biancadanforth:
Hm... just realized I ran all the unit tests locally, but not the functional tests. All of the failures are functional tets, and most are along the lines of:
```
Unable to locate element: div.product-payment
```
I'm wondering if these tests are getting redirected to the new "checkout" route, instead of the existing "product" route? I'll investigate when I return Monday. For now, I will mark this as DNM.

chenba:
> 
> 
> Hm... just realized I ran all the unit tests locally, but not the functional tests. 

There are a couple subscriptions related tests that only runs in CircleCI.  So there's that.  It's based on an env var ("CIRCLECI"), so you need to set that to "true" if you want to run them locally.

> I'm wondering if these tests are getting redirected to the new "checkout" route, instead of the existing "product" route?

Probably.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
chenba: It was added to force a sign in.  I think we still need that.  We are going to have a sign in link on the new flow, if we don't have a way of forcing sign in, the user would be taken back to "checkout" since they are not signed in.

biancadanforth: Hmm yes looking at the current logic, you are right. I'll add it back in... Thank you.

chenba: I think that will also fix the functional tests.

biancadanforth:
```diff
--- a/packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
+++ b/packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
@@ -33,7 +33,7 @@ class SubscriptionsProductRedirectView extends FormView {
       .then(() => {
         const isSignedInPromise =
           account?.isSignedIn() || Promise.resolve(false);
-        isSignedInPromise.then((isSignedIn) => {
+        return isSignedInPromise.then((isSignedIn) => {
           this._queryParams = Url.searchParams(this.window.location.search);
           const forceSignin = !!this._queryParams.signin;
           this._redirectPath =
```

The issue with the failed content-server tests was that I was not returning a nested promise from inside the `render` method, so the `afterRender` method of `SubscriptionsProductRedirectView` was executing and calling `navigateToPaymentServer` often with `this._redirectPath` and `this._queryParams` as `undefined`.

Consequently, when the functional tests were trying to load the payments Product route, it wasn't loading the subscribe page.

Thanks to @chenba for pairing with me on this to figure it out.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -15,30 +15,45 @@ class SubscriptionsProductRedirectView extends FormView {
...
-      !this._subscriptionsConfig.allowUnauthenticatedRedirects;
-
     // Flow events need to be initialized before the navigation
     // so the flow_id and flow_begin_time are propagated
     this.initializeFlowEvents();
   }
 
+  render() {
```
vbudhram, line 30: I think you might be able to do this in the `beforeRender`?

biancadanforth: Sure if it returns a promise that is awaited; that's a good point. I'll move it.

chenba: The Base view's `render` calls `beforeRender` _after_ it calls `checkAuthorization`, but `checkAuthorization` need the `mustAuth` value, so that's why `mustAuth` needs to be determined before all of that happens.

biancadanforth: So `render` [gets executed before `beforeRender`](https://github.com/mozilla/fxa/blob/32c09728bfc2f5ad8208744c76f406045dfbbd8c/packages/fxa-content-server/app/scripts/views/base.js#L264-L274)? How does that make any sense?

chenba: I think that given the name of these functions, we _naturally_ imagine that there's some lifecycle management code calling `beforeRender`, `render`, and then `afterRender`, in that order.  But that's not the case.  The only framework (Backbone.js) function is `render`; that's gets call because it's the rendering function.

biancadanforth: You are correct though, Barry.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
```diff
@@ -15,30 +15,45 @@ class SubscriptionsProductRedirectView extends FormView {
...
+    const account = this.getSignedInAccount();
+    return Promise.resolve()
+      .then(() => {
+        const isSignedInPromise =
+          account?.isSignedIn() || Promise.resolve(false);
```
chenba, line 35: I kinda forgot that we are in plain JS here, but I think it's ok to use optional chaining in JS at this point.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
vbudhram: I guess I'm thinking this code looks like it could be simplified a little. `navigateToPaymentServer` also calls `isSignedIn`. Maybe we could skip the promises here and just check if `this.getSignedInAccount();` returns an object, either way this is just a nit an nothing to block on :)

chenba (approved):
LGTM.  I suppose the total number of tests for the value of `mustAuth` should be eight since it's dependent on three booleans.  Not a big deal though; you can decide whether to add more tests for that.

packages/fxa-content-server/app/scripts/views/subscriptions_product_redirect.js
chenba: I don't think we can correctly set `mustAuth` without checking `isSignedIn`?

But yeah, maybe we can passed the signed in state so we don't need to call `isSignedIn` twice.  (And agreed that it's not blocking.)

biancadanforth: Thanks Vijay for your thoughts here.

I'm not sure I understand well enough the difference between the `getSignedInAccount` and `account.isSignedIn` methods. Is `getSignedInAccount` sufficient to say that an account is signed in if an object is returned? If so, what is the difference between that and `account.isSignedIn()` (other than their return values)?

I know this promise stuff is a little confusing (honestly more than a little). I'm inclined to move forward here with what we have, as I know some folks are anxious to get things merged, even if they aren't perfect. But I'm happy to chat with you more about this and potentially submit a follow-up later if that works for you.

Issue #9937, opened 2021-07-23
Update our set password view text to better support l10n


vbudhram:
Ref https://github.com/mozilla/fxa-content-server-l10n/pull/507#discussion_r675211483

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3791)

PR #9938, opened 2021-07-23
fix(l10n): Update `continue to ...` text to match other usages

vbudhram:
Matches how we use `Continue to ...` in our other views.

Fixes https://github.com/mozilla/fxa/issues/9937

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/post_verify/finish_account_setup/set_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/post_verify/finish_account_setup/set_password.mustache
@@ -16,7 +16,7 @@
         <h1 id="fxa-account-setup-set-password-header">
         {{#productName}}
             <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(productName)s" -->
-            {{#t}}Create a Firefox Account{{/t}} <span class="service">{{#t}}to continue to %(productName)s{{/t}}</span>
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/scripts/templates/post_verify/finish_account_setup/set_password.mustache
```diff
@@ -16,7 +16,7 @@
         <h1 id="fxa-account-setup-set-password-header">
         {{#productName}}
             <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(productName)s" -->
```
dannycoates, line 18: This comment doesn't match anymore

packages/fxa-content-server/app/scripts/templates/post_verify/finish_account_setup/set_password.mustache
```diff
@@ -16,7 +16,7 @@
...
-            {{#t}}Create a Firefox Account{{/t}} <span class="service">{{#t}}to continue to %(productName)s{{/t}}</span>
+            {{#t}}Create a Firefox Account{{/t}} <span class="service">{{#t}}Continue to %(productName)s{{/t}}</span>
```
dannycoates, line 19: Not an r- but this reads badly to me 🤷

dannycoates (approved):
🙈

PR #9939, opened 2021-07-23
feat(auth): add google iap service class

### LINKED ISSUE #9878: Create firestore class in auth-server
Auth-server needs to use a config similar to event-brokers firestore, and then have a new GoogleIAP service class in `lib/payments` to talk to Firestore with the configured collection for Google IAP use.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3740)

bbangert:
Because:

* We want to support Google IAP data using Firestore.

This commit:

* Sets up a Google IAP service class that uses Firestore.

Closes #9878

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/firestore-db.ts
+++ b/packages/fxa-auth-server/lib/firestore-db.ts
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { Firestore } from '@google-cloud/firestore';
+import * as grpc from '@grpc/grpc-js';
+
+import { ConfigType } from '../config';
+
+export function setupFirestore(config: ConfigType) {
+  const fsConfig = Object.assign({}, config.authFirestore);
+  // keyFilename takes precedence over credentials
+  if (fsConfig.keyFilename) {
+    /* istanbul ignore next */
+    // @ts-ignore
+    delete fsConfig.credentials;
+  }
+
+  // Utilize the local firestore emulator when the env indicates
+  if (process.env.AUTH_FIRESTORE_EMULATOR_HOST) {
+    return new Firestore({
+      customHeaders: {
+        Authorization: 'Bearer owner',
+      },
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/lib/firestore-db.ts
```diff
@@ -0,0 +1,32 @@
...
+  const fsConfig = Object.assign({}, config.authFirestore);
+  // keyFilename takes precedence over credentials
+  if (fsConfig.keyFilename) {
+    /* istanbul ignore next */
+    // @ts-ignore
+    delete fsConfig.credentials;
+  }
+
+  // Utilize the local firestore emulator when the env indicates
+  if (!!process.env.AUTH_FIRESTORE_EMULATOR_HOST) {
```
chenba, line 19: nit^2: `!!` isn't necessary since we are in an `if` here.

packages/fxa-auth-server/lib/firestore-db.ts
```diff
@@ -0,0 +1,32 @@
...
+    return new Firestore({
+      customHeaders: {
+        Authorization: 'Bearer owner',
+      },
+      port: 9090,
+      projectId: 'fx-event-broker',
```
chenba, line 25: This doesn't look right.

packages/fxa-auth-server/lib/firestore-db.ts
```diff
@@ -0,0 +1,32 @@
...
+      servicePath: 'localhost',
+      sslCreds: grpc.credentials.createInsecure(),
+    });
+  } else {
+    return new Firestore(fsConfig);
+  }
+}
```
chenba, line 32: I don't see tests for this module.  But since there are a couple `if`s in here, it would be nice to test the branching.

chenba (approved):
Make sense overall.  Tests for `setupFirestore` would be nice.

packages/fxa-auth-server/lib/firestore-db.ts
bbangert: Right, there aren't for the event-broker either. Later on there'll be integrated tests that verify the local Firestore emulator works, which effectively ensures this works correctly (as in the event-broker).

Issue #9787, opened 2021-07-06
[subplat][auth-server] Update the /newsletters endpoint to accept an access token


data-sync-user:
We’d like to allow the Payments frontend to subscribe to newsletters on a user’s behalf through the `/newsletters` endpoint.  Currently a client is allowed to access that endpoint only with a session token.  We need to update it so that it also accepts an access token.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3711)

data-sync-user:
➤ Barry Chen commented:

After chatting with Vijay and Danny, we’ve decided to have the Payments frontend to subscribe to the opted-in newsletter.  For that we need to update the /newsletters endpoint to accept the access token that Payments possesses.

I'm changing this issue to reflect that approach.

PR #9948, opened 2021-07-26
feat(auth): allow /newsletters access with access token

chenba:
Because:
 - we need to enable Payments to subscribe to newsletters on a user's
   behalf

This commit:
 - add a new scope value for newsletters access
 - add the oauthToken strategy to /newsletters

Closes: #9787


packages/fxa-shared/oauth/constants.js
```diff
@@ -5,6 +5,7 @@
 module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
+  OAUTH_SCOPE_NEWSLETTERS: 'https://identity.mozilla.com/account/newsletters',
```
chenba, line 8: @dannycoates How about it?  Include this in the scopes of the token returned after a passwordless user sign up?

dannycoates (approved):
works for me!

packages/fxa-shared/oauth/constants.js
dannycoates: sure

chenba: hmm I will change that 'account' to 'accounts' to be consistent with the other scopes.

Issue #9361, opened 2021-06-21
[new user checkout] consent checkbox that handles paypal 


data-sync-user:
_copied from https://docs.google.com/document/d/1802dPnvlwCeWlVg0l9OnL5PKl9ffkoMnxRFxmWuaIyU/edit#_

* Open question: [Tony] The checkbox and submit button on the Checkout page only applies to the Stripe form. Do we need a similar checkbox for PayPal customers? What does that look like? (3 points)
* Particularly because we are now creating an account for new users. They must agree to FxA policies as well as VPN policies.
* PSP team has an action item to resolve this with legal in the next week.
* Note: We may need to add some legal copy/links to one or more areas of the page to capture the outcome of this.
* Suggestion: move checkbox above, if not checked, unable to launch paypal (btn disabled) or submit disabled.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3667)

data-sync-user:
➤ Bianca Danforth commented:

This will need to move to before the PayPal button on the existing checkout screen as well as the new user checkout screen being created in this epic.

Also, we may need to disable the PayPal and Stripe submit buttons based on both the checkbox and the email forms being complete and valid email addresses unassociated with an existing FxA account.

data-sync-user:
➤ Dave Justice commented:

updated in figma https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098 ( https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098|smart-link ) 

This ticket can now be used to track the adding of this consent checkbox. Removing needsux label

PR #9965, opened 2021-07-28
task(fxa-payments-server): Create PaymentConsentCheckbox component

meandavejustice:
## Because

- We need to be able to add a consent checkbox into the new funnel flow view

## This pull request

- componentize the checkbox from the payment form
- consolidates checkbox tests
- use new component in `SubscriptionCreate` and `SubscriptionUpgrade`

refs: #9361 

## Checklist

## Screenshots (Optional)

image


packages/fxa-payments-server/.storybook/components/MockApp.tsx
```diff
@@ -74,7 +75,7 @@ export const MockApp = ({
     // should go away as soon as we switch to just the AppContext.stripePromise
     () =>
       applyStubsToStripe(
-        (window.Stripe(stripeApiKey) as unknown) as stripe.Stripe
```
meandavejustice, line 77: prettier change

packages/fxa-payments-server/.storybook/components/MockApp.tsx
```diff
@@ -105,4 +106,17 @@ export const MockApp = ({
   );
 };
 
+type MockPageProps = {
+  locale: string;
+  children: React.ReactNode;
+};
+
+export const MockPage = ({ locale, children }: MockPageProps) => {
```
biancadanforth, line 114: What is the purpose of `MockPage` here? It seems the `Localized` wrapper doesn't work properly without it for the `PaymentConsentCheckbox`, though it looks the same when I wrap with the existing `MockApp` instead?

Edit: I see this was moved from the `PaymentForm` component. Do you know why it was needed there?

packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { PaymentConsentCheckbox } from './index';
+import { Form } from '../fields';
+import useValidatorState from '../../lib/validator';
+import { MockPage } from '../../../.storybook/components/MockApp';
+import { SELECTED_PLAN } from '../../lib/mock-data';
+
+const WrapCheckbox = () => {
+  const validator = useValidatorState();
+  return (
+    <Form validator={validator}>
```
biancadanforth, line 12: TIL: We need to wrap the checkbox in a `Form`, because it gets the `validator` object from the form's context.

packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.tsx
```diff
@@ -0,0 +1,55 @@
+/*
+ * Note: test coverage for this component exists in
+ * components/PaymentForm/index.test.tsx under the
+ * title 'rendering the legal checkbox Localized component'
+ */
+
+import React, { useContext } from 'react';
```
biancadanforth, line 7: Nit: Doesn't look like we're using `React` explicitly here; so we don't need to import it.

biancadanforth (changes requested):
Thanks Dave for this PR.

I was looking this over -- one thing I don't see on reading the ticket this is supposed to close (#9361 ) -- I don't see anywhere where this checkbox is tying into the `PaypalButton` (i.e. the `PaypalButton` should be disabled when this checkbox is unchecked and vice versa)?

AFAICT, what this patch looks to be doing currently is to create a new component from existing JSX and replace it in the `PaymentForm`, which does not include the `PaypalButton`. It looks more like a refactor to me of the existing `PaymentForm`.

meandavejustice:
> I don't see anywhere where this checkbox is tying into the PaypalButton

@biancadanforth You are right, this ticket should not close #9361. I've updated the description to "refs #9361". I think the appropriate place to tie the checkbox component to the paypal button would be in https://mozilla-hub.atlassian.net/browse/FXA-3577 which I am working on today. I will make sure #9361 ticket is referenced and closed in that pr.

packages/fxa-payments-server/.storybook/components/MockApp.tsx
meandavejustice: I'm not sure I understand the question.

`MockPage` is just a wrapper around `MockApp` that provides the locale in order for the `<Localized>` component to work inside storybook.

Are you saying that you don't like the way the code is broken up here?

biancadanforth: I'm saying I tried this out with wrapping `MockApp` instead of `MockPage` and it seemed to work, so why do we need `MockPage` here specifically?

```diff
diff --git a/packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.stories.tsx b/packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.stories.tsx
index 3b6f906ef..90a8e2075 100644
--- a/packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.stories.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.stories.tsx
@@ -3,7 +3,7 @@ import { storiesOf } from '@storybook/react';
 import { PaymentConsentCheckbox } from './index';
 import { Form } from '../fields';
 import useValidatorState from '../../lib/validator';
-import { MockPage } from '../../../.storybook/components/MockApp';
+import { MockApp } from '../../../.storybook/components/MockApp';
 import { SELECTED_PLAN } from '../../lib/mock-data';
 
 const WrapCheckbox = () => {
@@ -17,10 +17,10 @@ const WrapCheckbox = () => {
 
 storiesOf('components/PaymentConsentCheckbox', module).add('default', () => (
   <div style={{ display: 'flex' }}>
-    <MockPage locale="auto">
+    <MockApp>
       <div className="product-payment">
         <WrapCheckbox />
       </div>
-    </MockPage>
+    </MockApp>
   </div>
 ));
```

Then I was asking a broader question about why `MockPage` is needed at all anywhere -- why is `MockApp` insufficient?

meandavejustice: It looks like `MockApp` is sufficient if it worked for you, I can remove it from both payment form and paymentconsentcheckbox

packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.tsx
meandavejustice: Declaring React like this is necessary whenever using JSX for the version we are using, if we ever update to React 17 we can leave these declarations out https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

meandavejustice:
@biancadanforth MockPage removed

packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.tsx
```diff
@@ -0,0 +1,55 @@
+/*
+ * Note: test coverage for this component exists in
+ * components/PaymentForm/index.test.tsx under the
+ * title 'rendering the legal checkbox Localized component'
```
biancadanforth, line 4: There are two places where we show this legal consent checkbox, one in `PaymentForm` and one in `SubscriptionUpgrade`. If we are going to make this consent checkbox a new component, we should probably dedupe the [consent checkbox in `SubscriptionUpgrade`](https://github.com/mozilla/fxa/blob/a63c7f05ef123bf964340f0a0b78cf41246421a2/packages/fxa-payments-server/src/routes/Product/SubscriptionUpgrade/index.tsx#L195-L223) as well and move the `"rendering the legal checkbox Localized component"` tests for `PaymentForm` and `SubscriptionUpgrade` to a single test for the new component.

biancadanforth (changes requested):
Edit: Nit: Also can you rename the commit subject to say `refactor(...)` instead of `task(...)`?

Thanks Dave. My first review was topical, as I was approaching it from the perspective of closing #9361 . Now that has been descoped, so this review is evaluating the patch as a refactor.

I added a comment inline, but the TL;DR is if we are going to create a new consent checkbox component, we should consolidate other uses of this consent checkbox to use this new component and the associated checkbox-specific rendering tests as well to one test for that new component.

Also, I haven't thought ahead too much in this regard -- how are you envisioning controlling the state of other components (like the `PaypalButton`) from this checkbox?

biancadanforth:
> > I don't see anywhere where this checkbox is tying into the PaypalButton
> 
> @biancadanforth You are right, this ticket should not close #9361. I've updated the description to "refs #9361". I think the appropriate place to tie the checkbox component to the paypal button would be in https://mozilla-hub.atlassian.net/browse/FXA-3577 which I am working on today. I will make sure #9361 ticket is referenced and closed in that pr.

Okay that’s fine. Though I'm also wary of scope creep on that ticket, which sounds like it could include a lot of things. Ideally individual PRs are small, so they can be reviewed faster and more effectively. Happy to help think through how to approach this for this epic more with you (and others) if desired.

packages/fxa-payments-server/src/components/PaymentConsentCheckbox/index.tsx
biancadanforth: Hm that's interesting thanks. We're using it now directly anyway in your latest patch for the onClick optional prop.

packages/fxa-payments-server/src/components/PaymentForm/index.test.tsx
```diff
@@ -175,215 +172,6 @@ it('includes the confirmation checkbox when confirm = true and plan supplied', (
...
-        const actual = getLocalizedMessage(bundle, msgId, {
-          ...args,
-          intervalCount: 6,
-        });
-        expect(actual).toEqual(expected);
-      });
-    });
-  });
-});
-
```
biancadanforth, line 386: Wow I didn't realize how many tests in the `PaymentForm` were for the checkbox.

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -343,7 +343,10 @@ export const StripeElement = (props: WrappedStripeElementProps) => {
   );
 };
 
-type CheckboxProps = { onValidate?: OnValidateFunction } & FieldProps &
+type CheckboxProps = {
+  onClick?: (event: React.MouseEvent<HTMLInputElement>) => void;
```
biancadanforth, line 347: You don't need to add this explicitly to the `CheckboxProps`, as it is covered by:
```
React.DetailedHTMLProps<
  React.InputHTMLAttributes<HTMLInputElement>,
  HTMLInputElement
>
```
You shouldn't need to modify this file at all actually. The `onClick` handler gets passed down implicitly if it's there as an optional prop. This is demonstrated already in the `onClick` prop for the consent checkbox in `SubscriptionUpgrade`.

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -362,6 +365,7 @@ export const Checkbox = (props: CheckboxProps) => {
     children,
     name,
     label,
+    onClick,
```
biancadanforth, line 368: Remove

packages/fxa-payments-server/src/components/fields/index.tsx
```diff
@@ -372,6 +376,7 @@ export const Checkbox = (props: CheckboxProps) => {
 
   const onChange = useCallback(
     (ev) => {
+      if (onClick) onClick(ev);
```
biancadanforth, line 379: Remove

biancadanforth (approved):
R+WC (with changes). Thanks Dave; this is looking good.

Before merging, please remove the references to the `onClick` prop in `fields/index.tsx`, as it is already covered and passed down.

Issue #9730, opened 2021-06-23
[subplat][content-server] Ensure Customer Cannot Change Their Email Address while Finishing Account Setup


data-sync-user:
In some account sign up use cases the user is allowed to change their email address while on the password setup form; ensure this is not presented to users signed up via SubPlat while subscribing to a product.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3677)

data-sync-user:
➤ Vijay Budhram commented:

Barry Chen I’m not 100% if https://github.com/mozilla/fxa/pull/9976 ( https://github.com/mozilla/fxa/pull/9976|smart-link )  is what you intended with this issue?

Or is this specially restricting the user on the UX?

data-sync-user:
➤ Barry Chen commented:

Vijay Budhram  Sorry that I was not clear on this issue.  ☹️  In one of the planning meeting someone brought up the fact that in some flows the user have the option of changing the email address with which they are trying to sign in/up.  This issue is to ensure that we do not offer that option on the content-server UI for the passwordless users when they go through the flow via the email link to set the password.  I just looked at your demo again and it doesn’t look like this is an actual issue.  Sorry about wasting your time!  I’m okay with marking this as cancelled or done.

data-sync-user:
➤ Vijay Budhram commented:

No worries, you do bring up an interesting point though. If a user creates a subscription and a passwordless account, then attempts to create an account via our regular flows, they will be prompted to verify the email address and not given an option to set their password.

We can have a dedicated screen when a user attempts this and resend the Finish Account Setup email. I’m not sure how often this might happen in practice though.

data-sync-user:
➤ Barry Chen commented:

{quote}We can have a dedicated screen when a user attempts this and resend the Finish Account Setup email.{quote}

I did bring this up in some of the design meetings.  But yeah, let’s wait to see how many confused users with passwordless accounts we gonna have.

PR #9976, opened 2021-07-28
feat(passwords): Ensure password set on account before changing primary

vbudhram:
## Because

- Users should not be able to change their primary email if the account has no password set. In practice I don't think this flow would be accessible, but wouldn't hurt to be explict in the endpoint.

## This pull request

- Checks that user has set password else throws errro
- Renames variables to make things clearer on what is happening

Closes: https://github.com/mozilla/fxa/issues/9730

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/emails.js
+++ b/packages/fxa-auth-server/lib/routes/emails.js
@@ -812,36 +812,40 @@ module.exports = (
       },
       handler: async function (request) {
         const sessionToken = request.auth.credentials;
-        const uid = sessionToken.uid;
-        const primaryEmail = sessionToken.email;
-        const email = request.payload.email;
+        const { uid, verifierSetAt } = sessionToken;
+        const currentEmail = sessionToken.email;
+        const newEmail = request.payload.email;
 
         log.begin('Account.RecoveryEmailSetPrimary', request);
 
-        await customs.check(request, primaryEmail, 'setPrimaryEmail');
+        await customs.check(request, currentEmail, 'setPrimaryEmail');
 
         if (sessionToken.tokenVerificationId) {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -813,35 +813,40 @@ module.exports = (
       handler: async function (request) {
         const sessionToken = request.auth.credentials;
         const uid = sessionToken.uid;
-        const primaryEmail = sessionToken.email;
-        const email = request.payload.email;
+        const currentEmail = sessionToken.email;
+        const newEmail = request.payload.email;
```
dannycoates, line 817: I like the name change 👍

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -813,35 +813,40 @@ module.exports = (
...
         if (sessionToken.tokenVerificationId) {
           throw error.unverifiedSession();
         }
 
-        const secondaryEmail = await db.getSecondaryEmail(email);
-        if (secondaryEmail.uid !== uid) {
+        const account = await db.account(uid);
+        if (account.verifierSetAt <= 0) {
+          throw error.unverifiedAccount();
+        }
```
dannycoates, line 830: LGTM

dannycoates (approved):

packages/fxa-auth-server/lib/routes/emails.js
```diff
@@ -890,8 +895,6 @@ module.exports = (
             }
           }
 
-          const account = await db.account(uid);
-
```
dannycoates, line 894: Here's the bug in the tests. `account.emails` is "old" now since it's fetched first

PR #9984, opened 2021-07-29
feat(auth): add iap play token endpoint

### LINKED ISSUE #9897: Add endpoint for initial subscription token from VPN client
After the Google IAP process completes, the client will send us the subscription token. This token then needs to be validated and the resulting subscription data stored in Firestore. They should then have their profile refreshed so that profile fetches return the subscription capability.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3762)

bbangert:
Because:

* We want to verify and record ownership of a Play purchase token.

This commit:

* Adapts the Google Play billing sample code to utilize FxA's logging,
  removes one time purchase handling, and streamlines some of the
  original types in the google sample.
* Optimizes API calls from the sample to follow Google Play
  Subscription guidelines.
* Adds a registerToken endpoint for Play purchase token registration
  and validation in Firestore.

Closes #9897

## Checklist


bbangert:
Currently in draft so review can begin while I start setting up the tests for it.

dannycoates:
This is looking good to me so far 👍

packages/fxa-auth-server/docs/api.md
```diff
@@ -3628,10 +3628,16 @@ Create a new setup intent for attaching a new payment method to the user.
 :lock: authenticated with OAuth bearer token
 Update a user's default payment method for invoices to the attached payment method id.
 
-#### GET /oauth/subscriptions/google/plans
+#### GET /oauth/subscriptions/iap/plans/{appName}
 
 Returns available plans for Android clients.
```
birdsarah, line 3633:
```suggestion
Returns available plans for in-app purchase clients.
```

bbangert:
Note that I'm leaving out the user-manager tests for now, they will be added when I add the auth-server function that queries google play purchases to determine active subscriptions.

bbangert:
Rebased against main due to a `yarn.lock` conflict, and now the package dependencies are all messed up. Whole bunch of package updates went through recently, I'm assuming one of them has broken main. 🤷🏻

bbangert:
@dannycoates I have no idea why the auth-server tests don't start. They run fine locally.

packages/fxa-auth-server/lib/payments/google-play/play-billing.ts
```diff
@@ -0,0 +1,79 @@
...
+  private log: AuthLogger;
+  private prefix: string;
+  private iapConfigDbRef: TypedCollectionReference<IapConfig>;
+
+  public purchaseManager: PurchaseManager;
+  public userManager: UserManager;
+
+  constructor() {
+    const config = Container.get(AppConfig);
+    this.prefix = `${config.authFirestore.prefix}googleIap`;
```
dannycoates, line 25: do we want any kind of character delimiter between parts? sometimes they're useful when grepping

packages/fxa-auth-server/lib/payments/google-play/purchase-manager.ts
```diff
@@ -0,0 +1,334 @@
...
+   */
+  public async querySubscriptionPurchase(
+    packageName: string,
+    sku: string,
+    purchaseToken: string,
+    triggerNotificationType?: NotificationType
+  ): Promise<SubscriptionPurchase> {
+    // STEP 1. Query Play Developer API to verify the purchase token
+    const apiResponse = await new Promise((resolve, reject) => {
+      this.playDeveloperApiClient.purchases.subscriptions.get(
```
dannycoates, line 70: `googleapis` has a promise api too afaict

dannycoates (approved):
This looks good to me. The new routes are good, I only skimmed most of the APL code

packages/fxa-auth-server/lib/payments/google-play/play-billing.ts
bbangert: The prefix stored ends in a `-`, I ended up keeping it this way to match the event-broker style.

packages/fxa-auth-server/lib/payments/google-play/purchase-manager.ts
bbangert: It does, I copied the original sample code here, and was tempted to change it to a try/except, but once one factors in the `let` statement, capturing the exception, applying the transform, etc. it seemed about the same LoC so I left it as is.

Issue #9729, opened 2021-06-23
[subplat][content-server] Emit Sentry Events on Errors during Finish Account Setup Step


data-sync-user:
Ensure that the content-server frontend emits a Sentry event whenever the customer encounters an error that would prevent them from finish setting up the account.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3678)

PR #9995, opened 2021-07-30
feat(passwords): Add sentry logging for set password errors

vbudhram:
## Because

- We need to be able to see why users fail to set passwords on their new accounts

## This pull request

- Adds `logError` to spots where we need them in set password view. This call sentry under the hood

Closes: https://github.com/mozilla/fxa/issues/9729

## Checklist

### DIFF
--- a/packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js
+++ b/packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js
@@ -45,6 +45,7 @@ class SetPassword extends FormView {
   showValidationErrorsEnd() {
     if (this._getPassword() !== this._getVPassword()) {
       const err = AuthErrors.toError('PASSWORDS_DO_NOT_MATCH');
+      this.logError(err);
       this.showValidationError(this.$(PASSWORD_INPUT_SELECTOR), err, true);
     }
   }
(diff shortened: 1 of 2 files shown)


packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js
```diff
@@ -45,6 +45,7 @@ class SetPassword extends FormView {
   showValidationErrorsEnd() {
     if (this._getPassword() !== this._getVPassword()) {
       const err = AuthErrors.toError('PASSWORDS_DO_NOT_MATCH');
+      this.logError(err);
```
dannycoates, line 48: we don't need this in sentry, do we?

dannycoates (approved):
I don't like the input validation one but the rest looks good

packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js
vbudhram: Maybe not?  My guess is that this will have similar rate like our signup flows, but I don't think it would hurt to have it in.

dannycoates: we should be getting rid of these kinds of things from sentry, that aren't bugs, not adding more

Issue #9210, opened 2021-06-21
Firefox accounts emails should respect dir=rtl


data-sync-user:
When Firefox Accounts required to send a message to the user, it send the message in the user language, but for RTL languages it doesn't set the page direction, which cause the message content to look unprofessional and not easily readable.

This was originally filed at https://bugzilla.mozilla.org/show_bug.cgi?id=1355355 .  **Please see that bug for discussion and improvements that can be made as it includes screenshots and test cases.**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3388)

data-sync-user:
➤ Lauren Zugai commented:

This is completed in the new emails but will continue to be an issue until the new emails hit production and the old emails are retired.

data-sync-user:
➤ Lauren Zugai commented:

Per Slack convo, since the work is finished here, we are marking this as “done.” We shouldn’t verify it until the new emails are in production, however.

PR #10005, opened 2021-08-02
fix(emails): Render email content right to left for rtl locales

Yogita98:
## Because

- Emails are rendered left to right by default

## This pull request

- Will render emails right to left for rtl locales and left to right for ltr locales

Closes: #10011 

## Checklist

## Screenshots (Optional)
Before:
Emails rendered ltr by default for all the locales
image

After:
Emails rendered rtl for ar (rtl) locale
image


packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -8,6 +8,8 @@ import * as templates from './templates';
 import * as layouts from './layouts';
 import path = require('path');
 const config = require('../../../config').getProperties();
+import { compileSassToCss } from './sass-compile-files';
```
Yogita98, line 11: When the tests for scripts are being run, the `emails-scss` step doesn't get executed as part of the `write-emails-to-disk` test and hence the validation errors were being thrown. Executing the compilation step will ensure sass files are compiled down to css irrespective of from where its being called.

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
```diff
@@ -17,7 +17,10 @@ export const render = (body: string) => {
       </mj-title>      
     </mj-head>
     <mj-body css-class="body-80">
-      <mj-raw>
+    <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
+    <mj-include path="./lib/senders/emails/css/fxa/index.css" type="css" css-inline="inline" />
+    <mj-include path="./lib/senders/emails/css/rtl.css" type="css" />
```
Yogita98, line 22: I had to create a separate file for rtl and ltr classes since we don't want these styles to be inlined. These styles are getting applied dynamically so either we do this way or render the styles from inside Js something like this:
```
if(isLocaleRenderedRtl){
      const head = document.getElementsByTagName('head')[0];
      const style = document.createElement("style");
      /* @ts-ignore */
      // style.type = 'text/css'
      style.innerHTML = `.rtl div {         
        direction: rtl !important;
       }`;
      head.appendChild(style);
    }
```

packages/fxa-auth-server/lib/senders/emails/templates/cadReminder/ar.ftl
```diff
@@ -0,0 +1,7 @@
+sync-reminder = هذا هو التذكير الخاص بك لمزامنة الأجهزة.
```
Yogita98, line 1: Even this file is added for testing purpose. I'll remove it when the Pr is mergeable

packages/fxa-auth-server/lib/senders/emails/renderer.ts
LZoog: Ah interesting, nice find. 🙂 I'm glad we know why CI was choking on this now.

I'm OK with leaving this here for now since this makes CI happy and this PR is about RTL styles if this isn't quick to change, but I think we can compile all of the SCSS to CSS at the auth-server build step. We don't need SCSS files except during development.

packages/fxa-auth-server/lib/senders/emails/rtl.scss
```diff
@@ -0,0 +1,17 @@
+.ltr {
```
LZoog, line 1: Since this also explicitly sets `ltr`, maybe we can call this `ltr-rtl.scss`, or `locale-dir.scss`?

LZoog (commented):
This is looking really good! I really just had a few minor comments. 😄

Also, will you update your "closes" to close #10011? I scooted out #9210 into the last milestone since technically that won't be fixed until all FxA emails support this.

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -12,6 +12,23 @@ import { loadFtlFiles } from './load-ftl-files';
...
+  'fa',
+  'ha',
+  'he',
+  'khw',
+  'ks',
+  'ps',
+  'ur',
+  'uz_AF',
+  'yi',
+];
```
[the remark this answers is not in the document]
Yogita98: I think I googled the [list](https://meta.wikimedia.org/wiki/Template:List_of_language_names_ordered_by_code) of locales that are rtl or ltr, but we can definitely run this by someone in the FxA team.

packages/fxa-auth-server/lib/senders/emails/renderer.ts
Yogita98: Ahhh okay so it's okay if the scss are not compiled down to css during testing

jodyheavener (approved):
So I've previewed this and it works as described, but because my RTL knowledge is limited I'm not entirely sure how to push it to it's limits/test edge cases. Unless others find things I'm happy with how this looks 👍

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -11,7 +11,7 @@ const config = require('../../../config').getProperties();
 
 const mjmlConfig: Record<any, any> = {
   validationLevel: 'strict',
-  // filePath: path.join(__dirname, 'css', 'global.css'),
+  ignoreIncludes: config.env === 'test' ? true : false,
```
LZoog, line 14: I think this can just be `config.env === 'test'` since that evaluates to `true`/`false`.

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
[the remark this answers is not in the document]
LZoog: 👍 I've asked someone from the l10n team to look at this, we can tweak it later if they don't get a chance before we merge this PR. 🙂

jodyheavener:
@Yogita98 I'm starting to think that failure in test-auth-server might be legit, let us know if you want another set of eyes on it!

Yogita98:
Right, mjml is throwing validation errors again for not being able to read the css files. I tried to do what you suggested me,[ ignoring include files](https://github.com/Yogita98/fxa/blob/fix-emails-rtl/packages/fxa-auth-server/lib/senders/emails/renderer.ts#L14) when the tests are being run but aren't we using dev env for our [tests](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/scripts/test-local.sh#L8) as well?

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -10,8 +10,8 @@ import path = require('path');
 const config = require('../../../config').getProperties();
 
 const mjmlConfig: Record<any, any> = {
-  validationLevel: 'strict',
-  // filePath: path.join(__dirname, 'css', 'global.css'),
+  validationLevel: 'soft',
```
Yogita98, line 13: Wasn't able to fix the CI test failures so turning this validation back to soft for now and we can circle back on the issue in a follow up PR

LZoog: That's fine by me, the `cadReminderFirst` ticket is that sort of catch-all for anything remaining/follow ups (plus after both briefly looking at it, we're not sure what the issue is 🤨), let's just be sure to note on that ticket that we need to address this. Jody gave this the r+ and it LGTM, so I'm going to merge it! 🎉

Issue #9358, opened 2021-06-21
[new user checkout] Create a no-password FxA account on form submit


data-sync-user:
* Create a customer based on email address and append a userid to Stripe metadata
* Post to a new endpoint to create a new FxA account with a temporary password.
* Back-end work: Create the new endpoint that creates the account and sends an email. [FXA-3599 Linked]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3666)

data-sync-user:
➤ Dave Justice commented:

what do we do when paypal is used here? can we use the callback snippet?

data-sync-user:
➤ Dave Justice commented:

Do we have legal concerns for creating the account with the paypal snippet?

data-sync-user:
➤ Bianca Danforth commented:

Dave Justice , what do you mean by paypal snippet and callback snippet here?

We may also need to create a Stripe customer here after creating the no-password FxA account, depending on when the customer is created downstream.

data-sync-user:
➤ Dave Justice commented:

Bianca Danforth This was about whether we need to add a consent checkbox for paypal users when creating this account.

data-sync-user:
➤ Dave Justice commented:

It looks like the design has since been updated to include both policies for stripe and paypal in the payment section. So I think that question is solved

PR #10007, opened 2021-08-02
feat(subscriptions): create acct and subscribe with Stripe

chenba:
Because:
 - we need to create an account and then subscribe to a plan for a new
   Payments route

This commit:
 - add the Payments OAuth client id as a config
 - add functions for creating the passwordless account and setting the
   access token for the API client
 - add a function that can be used in callbacks to create an account and
   then subscribe to a plan with Stripe payments

part of #9358

### DIFF
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -359,7 +359,7 @@
         "termsUri": "",
         "privacyUri": "",
         "trusted": true,
-        "allowedScopes": "https://identity.mozilla.com/account/subscriptions",
+        "allowedScopes": "https://identity.mozilla.com/account/subscriptions https://identity.mozilla.com/account/newsletters",
         "publicClient": true
       },
       {
@@ -428,6 +428,6 @@
   },
   "mjml": {
     "enabledEmailAddress": "@testuser.com$",
-    "templates":  ["cadReminderFirst"]
+    "templates": ["cadReminderFirst"]
   }
 }
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -10,8 +10,9 @@ const path = require('path');
 
 const versionInfo = require('./version');
 
-const DEFAULT_SUPPORTED_LANGUAGES = require('fxa-shared').l10n
-  .supportedLanguages;
+const { OAUTH_SCOPE_SUBSCRIPTIONS } = require('fxa-shared/oauth/constants');
+const DEFAULT_SUPPORTED_LANGUAGES =
+  require('fxa-shared').l10n.supportedLanguages;
 
 convict.addFormats(require('convict-format-with-moment'));
 convict.addFormats(require('convict-format-with-validator'));
@@ -19,8 +20,7 @@ convict.addFormats(require('convict-format-with-validator'));
 const conf = (module.exports = convict({
   allowed_iframe_contexts: {
     default: [],
-    doc:
-      'DEPRECATED - context query parameters allowed to embed FxA within an IFRAME',
+    doc: 'DEPRECATED - context query parameters allowed to embed FxA within an IFRAME',
     format: Array,
   },
   allowed_metrics_flow_cors_origins: {
@@ -116,8 +116,7 @@ const conf = (module.exports = convict({
     },
     reportOnly: {
       default: false,
-      doc:
-        'DEPRECATED - Only send the "Content-Security-Policy-Report-Only" header',
+      doc: 'DEPRECATED - Only send the "Content-Security-Policy-Report-Only" header',
       env: 'CSP_REPORT_ONLY',
     },
     reportOnlyEnabled: {
@@ -142,8 +141,7 @@ const conf = (module.exports = convict({
   },
   env: {
     default: 'production',
-    doc:
-      "What environment are we running in?  Note: all hosted environments are 'production'.",
+    doc: "What environment are we running in?  Note: all hosted environments are 'production'.",
     env: 'NODE_ENV',
     format: ['production', 'development'],
   },
@@ -169,8 +167,7 @@ const conf = (module.exports = convict({
       },
       initialBackoff: {
         default: '100 milliseconds',
-        doc:
-          'Initial backoff for feature-flagging Redis connection retries, increases exponentially with each attempt',
+        doc: 'Initial backoff for feature-flagging Redis connection retries, increases exponentially with each attempt',
         env: 'FEATURE_FLAGS_REDIS_TIMEOUT',
         format: 'duration',
       },
@@ -296,29 +293,25 @@ const conf = (module.exports = convict({
     fonts: {
       unsupportedLanguages: {
         default: [],
-        doc:
-          'DEPRECATED: These languages should use system fonts instead of Fira Sans',
+        doc: 'DEPRECATED: These languages should use system fonts instead of Fira Sans',
         format: Array,
       },
     },
     localeSubdirSuffix: {
       default: '',
-      doc:
-        'Enable alternative localized resources for Mozilla Online with private use subtag',
+      doc: 'Enable alternative localized resources for Mozilla Online with private use subtag',
       env: 'I18N_LOCALE_SUBDIR_SUFFIX',
       format: ['', '_x_mococn'],
     },
     supportedLanguages: {
       default: DEFAULT_SUPPORTED_LANGUAGES,
-      doc:
-        'List of languages this deployment should detect and display localized strings.',
+      doc: 'List of languages this deployment should detect and display localized strings.',
       env: 'I18N_SUPPORTED_LANGUAGES',
       format: Array,
     },
     translationDirectory: {
       default: path.resolve(__dirname, '../../app/i18n/'),
-      doc:
-        'The directory where per-locale .json files containing translations reside',
+      doc: 'The directory where per-locale .json files containing translations reside',
       env: 'I18N_TRANSLATION_DIR',
       format: String,
     },
@@ -388,8 +381,7 @@ const conf = (module.exports = convict({
       dcdb5ae7add825d2: '123done',
       '325b4083e32fe8e7': '321done',
     },
-    doc:
-      'Mappings from client id to service name: { "id1": "name-1", "id2": "name-2" }',
+    doc: 'Mappings from client id to service name: { "id1": "name-1", "id2": "name-2" }',
     env: 'OAUTH_CLIENT_IDS',
     format: Object,
   },
@@ -430,8 +422,7 @@ const conf = (module.exports = convict({
   },
   page_template_subdirectory: {
     default: 'src',
-    doc:
-      'Subdirectory of page_template_root for server-rendered page templates',
+    doc: 'Subdirectory of page_template_root for server-rendered page templates',
     env: 'PAGE_TEMPLATE_SUBDIRECTORY',
     format: ['src', 'dist'],
   },
@@ -442,8 +433,7 @@ const conf = (module.exports = convict({
         'a2270f727f45f648', // Fenix
         '1b1a3e44c54fbb58', // Firefox for iOS
       ],
-      doc:
-        'OAuth Client IDs that are allowed to pair. Remove all clients from this list to disable pairing.',
+      doc: 'OAuth Client IDs that are allowed to pair. Remove all clients from this list to disable pairing.',
       env: 'PAIRING_CLIENTS',
       format: Array,
     },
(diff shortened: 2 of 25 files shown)


packages/fxa-shared/oauth/constants.js
```diff
@@ -5,7 +5,9 @@
 module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
-  OAUTH_SCOPE_NEWSLETTERS: 'https://identity.mozilla.com/accounts/newsletters',
+  OAUTH_SCOPE_NEWSLETTERS: 'https://identity.mozilla.com/account/newsletters',
+  OAUTH_SCOPE_SUBSCRIPTIONS:
+    'https://identity.mozilla.com/account/subscriptions',
```
chenba, line 10: Oh I also moved this scope here so that identical string isn't repeated in multiple places.

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -668,16 +653,14 @@ const conf = (module.exports = convict({
       format: String,
     },
     managementScopes: {
-      default: 'profile https://identity.mozilla.com/account/subscriptions',
-      doc:
-        'OAuth scopes needed for the subscription management pages to access auth server APIs',
+      default: `profile ${OAUTH_SCOPE_SUBSCRIPTIONS}`,
```
chenba, line 656: This is the real change in this file; everything else's prettier's doing.

chenba:
> 
> 
> @chenba Hey Barry, could you add some steps for testing/verifying this? Thank You

In your checkout route (on the branch for this PR), create a `onSubmit` callback that's similar to the one for Stripe in `SubscriptionCreate`.  But instead of calling `handleSubscriptionPayment` in the callback, you'd call `handlePasswordlessSubscription`.  You can hard code the email address, and the client id is available in the config object.  Then when you go the checkout page without signing up/in, enter the cc info, and subscribe.

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -73,7 +73,7 @@ async function apiFetch(
     ...options,
     headers: {
       'Content-Type': 'application/json',
-      Authorization: `Bearer ${accessToken}`,
+      Authorization: accessToken ? `Bearer ${accessToken}` : undefined,
```
biancadanforth, line 76: What is the effect of an `Authorization` header of `undefined` here in the auth server? Is this better/different than omitting the header entirely? Looking at [`fxa-auth-server/lib/oauth/client.js`](https://github.com/mozilla/fxa/blob/521338b9f7cc102c348a55e70f02c2e2e640d69a/packages/fxa-auth-server/lib/oauth/client.js#L65-L67), it seems like either would have the same effect as long as we have e.g. a `clientId` in the request body instead?

packages/fxa-payments-server/src/lib/config.test.ts
```diff
@@ -155,28 +156,30 @@ const expectedMergedConfig = {
...
+            switch (selector) {
+              case headSelector(META_CONFIG):
+                return configValue;
+              case headSelector(META_FEATURE_FLAGS):
+                return flagsValue;
+              default:
+                return null;
+            }
+          },
+        };
```
biancadanforth, line 182: I'm guessing this is just git diff making a stylelint refactor very difficult to identify? 🤔

packages/fxa-payments-server/src/lib/errors.ts
```diff
@@ -128,12 +129,14 @@ const paymentErrors2 = [
...
 
 const paymentErrors3 = ['general-paypal-error'];
+const signupErrors = ['fxa_signup_error'];
 
 cardErrors.forEach((k) => (errorToErrorMessageMap[k] = CARD_ERROR));
 basicErrors.forEach((k) => (errorToErrorMessageMap[k] = BASIC_ERROR));
 paymentErrors1.forEach((k) => (errorToErrorMessageMap[k] = PAYMENT_ERROR_1));
 paymentErrors2.forEach((k) => (errorToErrorMessageMap[k] = PAYMENT_ERROR_2));
 paymentErrors3.forEach((k) => (errorToErrorMessageMap[k] = PAYMENT_ERROR_3));
+signupErrors.forEach((k) => (errorToErrorMessageMap[k] = FXA_SIGNUP_ERROR));
```
biancadanforth, line 139: I know you didn't create this in this PR, but why is this better than directly updating the `errorToErrorMessageMap` when there's a new localizable error message?

packages/fxa-payments-server/src/lib/stripe.test.ts
```diff
@@ -2,10 +2,101 @@
...
+  }),
+  apiDetachFailedPaymentMethod: jest.fn().mockResolvedValue({}),
+};
 
 // handleSubscriptionPayment and handlePaymentIntent are tested as part of SubscriptionCreate
 
+// At the time of writing, there is no UI component that uses
+// handlePasswordlessSubscription; once we have that implementation, we should
+// test handlePasswordlessSubscription in the same fashion that we test
+// handleSubscriptionPayment.
```
biancadanforth, line 41: Maybe add a "TODO #9358" in the comment?

In general, I know this PR is part of closing the ticket above -- what all else is left? Here's what I have:
* Write similar Paypal function to `handlePasswordlessSubscription`
* Create new onSubmit handlers for Stripe and Paypal in the Checkout route that wrap these functions (this is where my newsletter function will go from PR#10031) and pass them in as the `onSubmit` prop into `PaymentForm` and `PaypalButton`.
* Replace the `/account/stub` endpoint with the final route name and test the end-to-end integration?

packages/fxa-payments-server/src/lib/stripe.test.ts
```diff
@@ -2,10 +2,101 @@
...
+
+  it('calls handlePasswordlessSignUp then handleSubscriptionPayment on success', async () => {
+    const onFailure = jest.fn();
+
+    (handlePasswordlessSignUp as jest.Mock).mockResolvedValue({});
+    await handlePasswordlessSubscription({
+      email,
+      clientId,
+      stripe: stripeOverride,
+      name: 'BMO',
```
biancadanforth, line 59: Is this your cat's name or Bugzilla? :-D

packages/fxa-payments-server/src/lib/stripe.test.ts
```diff
@@ -2,10 +2,101 @@
...
+      selectedPlan: PLAN,
+      customer: null,
+      retryStatus: undefined,
+      ...apiClientOverrides,
+      onFailure,
+      onRetry: noop,
+      onSuccess: noop,
+    });
+
+    expect(handlePasswordlessSignUp).toBeCalledWith({ email, clientId });
```
biancadanforth, line 71: Wouldn't we also expect `handleSubscriptionPayment` to be called (with)?

packages/fxa-payments-server/src/lib/stripe.test.ts
```diff
@@ -2,10 +2,101 @@
...
+    expect(onFailure).toHaveBeenCalledWith(FXA_SIGNUP_ERROR);
```
biancadanforth, line 96: Wouldn't we also expect handleSubscriptionPayment to _not_ have been called?

packages/fxa-payments-server/src/lib/stripe.test.ts
```diff
@@ -2,10 +2,101 @@
...
+  });
+});
+
```
biancadanforth, line 99: Do we also want an error test case where `handlePasswordlessSignup` succeeds but `handleSubscriptionPayment` fails? Just thinking of permutations here.

biancadanforth (approved):
R+WC. Thanks Barry for this PR; we tested that this works together when adding the `onSubmit` callback in the Checkout route yesterday.

I mostly have a lot of questions, though I recommend a couple more test cases as well (that's the WC part).

One overall question: What all needs to be done to close #9358 after this merges? My thinking:
* Write similar Paypal function to `handlePasswordlessSubscription`
* Create new onSubmit handlers for Stripe and Paypal in the Checkout route that wrap these functions (this is where my newsletter function will go from PR#10031) and pass them in as the `onSubmit` prop into `PaymentForm` and `PaypalButton`.
* Replace the `/account/stub` endpoint to the final route name and test end-to-end?

packages/fxa-auth-server/config/dev.json
```diff
@@ -359,7 +359,7 @@
         "termsUri": "",
         "privacyUri": "",
         "trusted": true,
-        "allowedScopes": "https://identity.mozilla.com/account/subscriptions",
+        "allowedScopes": "https://identity.mozilla.com/account/subscriptions https://identity.mozilla.com/account/newsletters",
```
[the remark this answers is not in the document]
chenba: > 
> 
> Where do these `clients` objects live in non-dev contexts?

They are in the database.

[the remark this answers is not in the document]
chenba: > 
> 
> Seems like it might be in fxa-content-server/server/lib/configuration.js? I don't understand why they would live on two different servers though.

The ones in the content-server are used when creating an access token for Payments before it redirects users there.

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -10,9 +10,6 @@ function nock(it: any) {
 jest.mock('./sentry');
 
 import {
-  createSubscription_PENDING,
-  createSubscription_FULFILLED,
-  createSubscription_REJECTED,
```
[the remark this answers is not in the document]
chenba: I'll just remove them in this PR if you don't mind.

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -98,12 +101,42 @@ describe('APIError', () => {
...
-  const OAUTH_BASE_URL = 'http://oauth.example.com';
-  const AUTH_BASE_URL = 'http://auth.example.com';
-  const PROFILE_BASE_URL = 'http://profile.example.com';
+describe('Authorization header', () => {
+  let config: Config;
+
+  beforeEach(() => {
+    jest.spyOn(global, 'fetch');
+
+    config = defaultConfig();
```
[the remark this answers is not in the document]
chenba: It needs to be a valid Config object at least, and '' is a valid value of type string.  😄  It's populated from the stuff in the meta tag, which in turn, yes, get its values from `fxa-payments-server/server/config/index.js`

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -98,12 +101,42 @@ describe('APIError', () => {
...
+    mockOptionsResponses(PROFILE_BASE_URL);
+  });
 
+  afterEach(() => {
+    noc.cleanAll();
+    (global.fetch as jest.Mock).mockRestore();
+  });
+
+  it('should be undefine when there is no access token', async () => {
+    await apiFetchProfile();
```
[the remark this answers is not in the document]
chenba: Yes.  Just testing to see what's in the request, specifically the authorization header.

packages/fxa-payments-server/src/lib/apiClient.ts
chenba: It should be the same as omitting it.

packages/fxa-payments-server/src/lib/config.test.ts
chenba: LOL

packages/fxa-payments-server/src/lib/errors.ts
chenba: It's not really better when there's a one to one mapping.

packages/fxa-payments-server/src/lib/stripe.test.ts
chenba: Sure but we cannot spy on the `handleSubscriptonPayment` function that's enclosed in `handlePasswordlessSignUp`.

packages/fxa-payments-server/src/lib/stripe.test.ts
chenba: Same reason as above.

packages/fxa-payments-server/src/lib/stripe.test.ts
chenba: When `handleSubscriptionPayments` fails we'll have an uncaught exception in the success test above and that test will fail.  I cannot stub out `handleSubscriptionPayments` directly and have it fail; it'd need to be through the Stripe or API client overrides.

PR #10008, opened 2021-08-02
fix(content): fix inline totp edge case

dannycoates:
This sentry event https://sentry.prod.mozaws.net/operations/fxa-content-client-prod/issues/10992802/events/3323a6cddda74ce2b23852828b802b65/

is caused by a session being in an unverified state when the flow first starts. Here we were checking for the sessionToken and navigating, but also continuing to execute the sessionVerificationStatus which threw and error. It was benign but still sent it to sentry.


chenba (approved):
↩️

PR #10009, opened 2021-08-02
task(fxa-payments-server): add components to new checkout layout

### LINKED ISSUE #9361: [new user checkout] consent checkbox that handles paypal

### LINKED ISSUE #9750: Update layout for new user flow
We will need to create the layout for the new user page and place it into the checkout/{product_id} route

page components:

* header
* new user form
* consent checkbox
* payment selection
* terms

[https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098](https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098|smart-link)

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/FXA-3577)

meandavejustice:
## This pull request

-Adds components needed to fulfill the layout of the new checkout page.
- Adds the check to enable submit button once consent checkbox is
checked.

- fixes #9750
- refs #9361

## Checklist

## Screenshots (Optional)

image
image

### DIFF
--- a/packages/fxa-payments-server/src/components/Header/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/Header/index.test.tsx
@@ -33,6 +33,18 @@ describe('Header', () => {
     expect(avatar).toHaveAttribute('alt', userProfile.displayName);
   });
 
+  it('renders without profile', () => {
+    userProfile.displayName = null;
+    const subject = () => {
+      return render(<Header />);
+    };
+
+    const { queryByTestId } = subject();
+
+    const brand = queryByTestId('branding');
+    expect(brand).toBeVisible();
+  });
+
   it('alt falls back to email is displayName is null', () => {
     userProfile.displayName = null;
     const subject = () => {
--- a/packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
+++ b/packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
@@ -1,21 +1,35 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
 import { NewUserEmailForm } from './index';
 
+const WrapNewUserEmailForm = ({
+  accountExistsReturnValue,
+}: {
+  accountExistsReturnValue: boolean;
+}) => {
+  const [validEmail, setValidEmail] = useState<string>('');
+  const [accountExists, setAccountExists] = useState(false);
+  return (
+    <div style={{ display: 'flex' }}>
+      <NewUserEmailForm
+        signInURL={
+          'https://localhost:3031/subscriptions/products/productId?plan=planId&signin=yes'
+        }
+        setValidEmail={setValidEmail}
+        setAccountExists={setAccountExists}
+        getString={(id: string) => id}
+        checkAccountExists={() =>
+          Promise.resolve({ exists: accountExistsReturnValue })
+        }
+      />
+    </div>
+  );
+};
+
 storiesOf('components/NewUserEmailForm', module)
   .add('default', () => (
-    <div style={{ display: 'flex' }}>
-      <NewUserEmailForm
-        getString={(id: string) => id}
-        checkAccountExists={() => Promise.resolve({ exists: false })}
-      />
-    </div>
+    <WrapNewUserEmailForm accountExistsReturnValue={false} />
   ))
   .add('existing account', () => (
-    <div style={{ display: 'flex' }}>
-      <NewUserEmailForm
-        getString={(id: string) => id}
-        checkAccountExists={() => Promise.resolve({ exists: true })}
-      />
-    </div>
+    <WrapNewUserEmailForm accountExistsReturnValue={true} />
   ));
--- a/packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
   act,
   fireEvent,
@@ -13,15 +13,36 @@ import {
   checkAccountExists,
 } from './index';
 
+const WrapNewUserEmailForm = ({
+  accountExistsReturnValue,
+}: {
+  accountExistsReturnValue: boolean;
+}) => {
+  const [validEmail, setValidEmail] = useState<string>('');
+  const [accountExists, setAccountExists] = useState(false);
+  return (
+    <div style={{ display: 'flex' }}>
+      <NewUserEmailForm
+        signInURL={
+          'https://localhost:3031/subscriptions/products/productId?plan=planId&signin=yes'
+        }
+        setValidEmail={setValidEmail}
+        setAccountExists={setAccountExists}
+        getString={(id: string) => id}
+        checkAccountExists={() =>
+          Promise.resolve({ exists: accountExistsReturnValue })
+        }
+      />
+    </div>
+  );
+};
+
 describe('NewUserEmailForm test', () => {
   it('renders as expected', () => {
     let subject;
     act(() => {
       subject = render(
-        <NewUserEmailForm
-          getString={(id: string) => id}
-          checkAccountExists={() => Promise.resolve({ exists: false })}
-        />
+        <WrapNewUserEmailForm accountExistsReturnValue={false} />
       );
     });
     const form = subject.queryByTestId('new-user-email-form');
@@ -49,10 +70,7 @@ describe('NewUserEmailForm test', () => {
     let subject;
     await act(async () => {
       subject = render(
-        <NewUserEmailForm
-          getString={(id: string) => id}
-          checkAccountExists={() => Promise.resolve({ exists: false })}
-        />
+        <WrapNewUserEmailForm accountExistsReturnValue={false} />
       );
       const firstEmail = subject.getByTestId('new-user-email');
 
@@ -67,10 +85,7 @@ describe('NewUserEmailForm test', () => {
 
     await act(async () => {
       subject = render(
-        <NewUserEmailForm
-          getString={(id: string) => id}
-          checkAccountExists={() => Promise.resolve({ exists: false })}
-        />
+        <WrapNewUserEmailForm accountExistsReturnValue={false} />
       );
       const firstEmail = subject.getByTestId('new-user-email');
       fireEvent.change(firstEmail, { target: { value: 'valid@email.com' } });
@@ -84,10 +99,7 @@ describe('NewUserEmailForm test', () => {
     let subject;
     await act(async () => {
       subject = render(
-        <NewUserEmailForm
-          getString={(id: string) => id}
-          checkAccountExists={() => Promise.resolve({ exists: false })}
-        />
+        <WrapNewUserEmailForm accountExistsReturnValue={false} />
       );
       const firstEmail = subject.getByTestId('new-user-email');
       const secondEmail = subject.getByTestId('new-user-confirm-email');
@@ -106,10 +118,7 @@ describe('NewUserEmailForm test', () => {
     let subject;
     await act(async () => {
       subject = render(
-        <NewUserEmailForm
-          getString={(id: string) => id}
-          checkAccountExists={() => Promise.resolve({ exists: false })}
-        />
+        <WrapNewUserEmailForm accountExistsReturnValue={false} />
       );
       const firstEmail = subject.getByTestId('new-user-email');
       const secondEmail = subject.getByTestId('new-user-confirm-email');
@@ -136,6 +145,7 @@ describe('NewUserEmailForm test', () => {
       'foxy@mozilla.com',
       false,
       (state: string) => state,
+      (value: boolean) => value,
       (id: string) => id,
       checkAccountExists
     ).then((result) => {
--- a/packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
+++ b/packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { Dispatch, SetStateAction, useState } from 'react';
 import { Localized } from '@fluent/react';
 
 import { isEmailValid } from '../../../../fxa-shared/email/helpers';
@@ -16,14 +16,20 @@ import './index.scss';
 import { config } from '../../lib/config';
 
 export type NewUserEmailFormProps = {
-  getString: (id: string) => string;
-  checkAccountExists: Function;
+  getString?: (id: string) => string;
+  signInURL: string;
+  setValidEmail: (value: string) => void;
+  setAccountExists: (value: boolean) => void;
+  checkAccountExists: (email: string) => Promise<{ exists: boolean }>;
   validatorInitialState?: ValidatorState;
   validatorMiddlewareReducer?: ValidatorMiddlewareReducer;
 };
 
 export const NewUserEmailForm = ({
   getString,
+  signInURL,
+  setValidEmail,
+  setAccountExists,
   checkAccountExists,
   validatorInitialState,
   validatorMiddlewareReducer,
@@ -41,9 +47,12 @@ export const NewUserEmailForm = ({
       validator={validator}
       className="new-user-email-form"
     >
-      <Localized id="new-user-sign-in-link">
+      <Localized
+        id="new-user-sign-in-link"
+        elems={{ a: <a href={signInURL}></a> }}
+      >
         <p className="sign-in-copy" data-testid="sign-in-copy">
-          Already have a Firefox account? <a>Sign in</a>
+          Already have a Firefox account? <a href={signInURL}>Sign in</a>
         </p>
       </Localized>
       <Localized id="new-user-email" attrs={{ placeholder: true, label: true }}>
@@ -59,9 +68,11 @@ export const NewUserEmailForm = ({
             return emailInputValidationAndAccountCheck(
               value,
               focused,
+              setValidEmail,
+              setAccountExists,
               setEmailInputState,
-              getString,
-              checkAccountExists
+              checkAccountExists,
+              getString
             );
           }}
         />
@@ -97,7 +108,7 @@ export const NewUserEmailForm = ({
 
       <div className="assurance-copy" data-testid="assurance-copy">
         <img src={shieldIcon} alt="shield" />
-        <Localized id="new-user-subscribe-product-updates">
+        <Localized id="new-user-subscribe-product-assurance">
           <p>
             We only use your email to create your account. We will never sell it
             to a third party.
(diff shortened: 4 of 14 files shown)


packages/fxa-payments-server/src/components/Header/index.test.tsx
```diff
@@ -33,6 +33,18 @@ describe('Header', () => {
...
+  it('renders without profile', () => {
+    userProfile.displayName = null;
+    const subject = () => {
+      return render(<Header />);
+    };
+
+    const { queryByTestId } = subject();
+
+    const avatar = queryByTestId('branding');
+    expect(avatar).toBeVisible();
```
chenba, line 45: 'branding' is not 'avatar'?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -16,14 +16,18 @@ import './index.scss';
 import { config } from '../../lib/config';
 
 export type NewUserEmailFormProps = {
-  getString: (id: string) => string;
+  getString?: (id: string) => string;
+  setValidEmailSet: Function;
```
chenba, line 20: Why not set the actual email?  We do need to use it in a request.  Is there any way to get the email address instead?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -16,14 +16,18 @@ import './index.scss';
...
+  setAccountExists: Function;
```
chenba, line 21: If we know the types of these functions, would be good to declare them, and not just use `Function`.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -122,21 +128,26 @@ export async function checkAccountExists(userEmail: string) {
...
   if (isEmailValid(value)) {
     valid = true;
+    setValidEmailSet(true);
     const response = await checkAccountExists(value);
     if (response.exists) {
       hasAccount = true;
-    }
-  }
+      setAccountExists(true);
+    } else setAccountExists(false);
```
chenba, line 148: Could be
```
hasAccount = response.exists;
setAccountExists(response.exists);
```
so we won't have the nesting.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -226,6 +237,67 @@ export const PaymentForm = ({
...
+                  : payButtonL10nId(customer!)
+              }
+            >
+              <span className="lock">{submitButtonCopy}</span>
+            </Localized>
+          )}
+        </SubmitButton>
+      </Localized>
+    </div>
+  ) : null;
```
chenba, line 299: When do we want `buttons` to be `null`?

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -226,6 +237,67 @@ export const PaymentForm = ({
...
+          {showProgressSpinner ? (
+            <span data-testid="spinner-submit" className="spinner">
+              &nbsp;
+            </span>
+          ) : (
+            <Localized
+              id={
+                submitButtonL10nId
+                  ? submitButtonL10nId
+                  : payButtonL10nId(customer!)
```
chenba, line 290: We definitely don't have a `customer` in a couple of flows.  Instead of asserting `customer`, I think we should update `payButtonL10nId` to accept null or undefined.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
```diff
@@ -236,7 +308,16 @@ export const PaymentForm = ({
     >
       {paymentSource}
 
-      <hr />
+      {showLegal ? (
+        <>
+          <br />
+          <br />
+          <br />
```
chenba, line 315: heh classic 😉

packages/fxa-payments-server/src/routes/Checkout/index.test.tsx
```diff
@@ -104,13 +101,20 @@ describe('routes/Checkout', () => {
...
 
     const checkoutEle = await findByTestId('checkout-route-container');
     expect(checkoutEle).toBeInTheDocument();
+
+    const formEl = getByTestId('new-user-email-form');
+    expect(formEl).toBeInTheDocument();
+
+    const planDetailsEl = getByTestId('plan-details-component');
+    expect(planDetailsEl).toBeInTheDocument();
+
```
chenba, line 117: There's also a payment form from this patch.

chenba:
When I opened the checkout page after building the branch locally, the layout looks different than the screenshots above.
image

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -47,6 +99,31 @@ export const Checkout = ({
     [plans]
```
chenba, line 99: I know it's not part of this patch, but the selected plan should be based on the plan id and not the list of plans?

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -55,8 +132,138 @@ export const Checkout = ({
...
+        />
+        <div
+          className={classNames('product-payment', {
+            hidden: transactionInProgress || paymentError,
+          })}
+          data-testid="subscription-create"
+        >
+          <h2 className="step-header">1. Create a Firefox account</h2>
+          <hr />
+          {/* "Already have an account? <a>Sign In</a>" https://mozilla-hub.atlassian.net/browse/FXA-3664 */}
```
chenba, line 164: Why is this commented out?

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -55,8 +132,138 @@ export const Checkout = ({
...
+            hidden: transactionInProgress || paymentError,
+          })}
+        />
+        <div
+          className={classNames('product-payment', {
+            hidden: transactionInProgress || paymentError,
+          })}
+          data-testid="subscription-create"
+        >
+          <h2 className="step-header">1. Create a Firefox account</h2>
```
chenba, line 162: L10n?

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -55,8 +132,138 @@ export const Checkout = ({
...
+          {/* "Already have an account? <a>Sign In</a>" https://mozilla-hub.atlassian.net/browse/FXA-3664 */}
+          <NewUserEmailForm
+            setValidEmailSet={setValidEmailSet}
+            setAccountExists={setAccountExists}
+            checkAccountExists={checkAccountExists}
+          />
+
+          <hr />
+          <h2 className="step-header">2. Choose your payment method</h2>
+          <h4>required</h4>
```
chenba, line 173: Is this heading markup just for the bold appearance/visual emphasis?  I don't think we should do that.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -55,8 +132,138 @@ export const Checkout = ({
...
+          <hr />
+          {/* "Already have an account? <a>Sign In</a>" https://mozilla-hub.atlassian.net/browse/FXA-3664 */}
+          <NewUserEmailForm
+            setValidEmailSet={setValidEmailSet}
+            setAccountExists={setAccountExists}
+            checkAccountExists={checkAccountExists}
+          />
+
+          <hr />
+          <h2 className="step-header">2. Choose your payment method</h2>
```
chenba, line 172: L10n?

chenba (changes requested):
Thanks for the big patch, Dave.  Got some questions and suggestions in the comments.  Also, the page in my local env doesn't seem to match your screenshots (fwiw I have PayPal enabled).

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: I updated these types locally, I'm getting a typescript error in the test, I'd like to pass the correct thing here to silence the error, but I'm not sure how to fake this type, I haven't found anything terribly useful in my searches yet...

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
meandavejustice: This case is hit for when `submit` is false. In that case we would not want to show the submit button and buttons should be null.

This was left over from when I was going to provide my own submit button at the parent level. It's no longer valid or needed, I will remove.

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
meandavejustice: This is an existing method, I think it would be a good thing to clean up also, can you file an issue and tag it for a maintenance sprint?

packages/fxa-payments-server/src/routes/Checkout/index.tsx
meandavejustice: Not sure I follow, can you elaborate?

packages/fxa-payments-server/src/components/PaymentForm/index.tsx
chenba: I mean we are just removing a `!` here at the call site and then change the param type of `payButtonL10Id` to `(c?: Customer | null)`.  Filing an issue, triaging, creating another PR, having someone review it seems like a lot of unnecessary effort.  Would you accept a patch that you can apply? 

In this case, TypeScript is correct about the type of `customer` here.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
chenba: That's the array of dependencies for `useMemo`, and the dependencies used in that function are `productId`, `planId`, and `plansByProductId`, not `plans`.  I think VS Code should be complaining about it to you as it does to me.

Official docs here: https://reactjs.org/docs/hooks-reference.html#usememo

meandavejustice: Totally see what you mean now, I wonder why my linter wasn't picking up on that 🤔, just pushed that fix.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: I actually fixed this up by just including `setState` calls in the stories and tests, it works and it's more true to the operating state of the component 👍

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -16,14 +16,18 @@ import './index.scss';
 import { config } from '../../lib/config';
 
 export type NewUserEmailFormProps = {
-  getString: (id: string) => string;
-  checkAccountExists: Function;
+  getString?: (id: string) => string;
+  setValidEmail: Dispatch<SetStateAction<string>>;
+  setAccountExists: Dispatch<SetStateAction<boolean>>;
```
chenba, line 21: nit: Those are definitely the types for those functions.  But they are harder for other engineers to read.  `(value: string) => void;` and `(value: boolean) => void;` make it obvious what the param and return types are.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -41,9 +45,9 @@ export const NewUserEmailForm = ({
       validator={validator}
       className="new-user-email-form"
     >
-      <Localized id="new-user-sign-in-link">
+      <Localized id="new-user-sign-in-link" elems={{ a: <a></a> }}>
         <p className="sign-in-copy" data-testid="sign-in-copy">
-          Already have a Firefox account? <a>Sign in</a>
+          Already have a Firefox account? <a href="#">Sign in</a>
```
chenba, line 50: Any reason we don't have the actual url for the href here?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -122,21 +128,24 @@ export async function checkAccountExists(userEmail: string) {
 export async function emailInputValidationAndAccountCheck(
   value: string,
   focused: boolean,
+  setValidEmail: Function,
+  setAccountExists: Function,
   setEmailInputState: Function,
-  getString: Function,
-  checkAccountExists: Function
+  checkAccountExists: Function,
```
chenba, line 134: I see that you defined the type at https://github.com/mozilla/fxa/pull/10009/files#diff-e7d37441df07a444329b2215c48e3c74a7a1c5d26f586cea6958da8bb385a9a8R22 Why not use it here?  `setValidEmail` and `setAccountExists` too.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -122,21 +128,24 @@ export async function checkAccountExists(userEmail: string) {
...
     valid = true;
+    setValidEmail(value);
     const response = await checkAccountExists(value);
-    if (response.exists) {
-      hasAccount = true;
-    }
-  }
+    hasAccount = response.exists;
+    setAccountExists(response.exists);
+  } else setValidEmail('');
```
chenba, line 147: nit: even though there's one line for the `else`, have it in a `{}` block is still preferred.

chenba (changes requested):
@meandavejustice  thanks for the updated patch.  I got some nits, but more importantly:

- We should use the actual url for the href of the link that takes the user with an account to sign in; afaik there isn't another jira for doing that
- I actually cannot click on the "sign in" link above the text input; something might be blocking it?
- Does the PayPal button display for you?  It doesn't for me.

chenba:
Is the checkbox above the PayPal button suppose to affect the state of the PayPal button?  Currently I can click on the PayPal button without checking the checkbox.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
```diff
@@ -1,21 +1,35 @@
...
+}: {
+  accountExistsReturnValue: boolean;
+}) => {
+  const [validEmail, setValidEmail] = useState<string>('');
+  const [accountExists, setAccountExists] = useState(false);
+  return (
     <div style={{ display: 'flex' }}>
       <NewUserEmailForm
+        signInURL={
+          'https://localhost:3031/subscriptions/products/productId?plan=planId&signin=yes'
```
chenba, line 16: Hard coded to `localhost` because we are changing it in the other jira?

chenba:
> 
> 
> Is the checkbox above the PayPal button suppose to affect the state of the PayPal button? Currently I can click on the PayPal button without checking the checkbox.

Oh kind of answering my question... I noticed that #9361 is listed as one of the issues that this PR fixes.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -136,6 +142,7 @@ describe('NewUserEmailForm test', () => {
       'foxy@mozilla.com',
       false,
       (state: string) => state,
+      (id: boolean) => id,
```
chenba, line 145: `id`?

packages/fxa-payments-server/src/routes/Checkout/index.scss
```diff
@@ -1,3 +1,103 @@
...
+    }
+
+    @include max-width('tablet') {
+      margin-top: 16px;
+    }
+  }
+
+  .subscription-create-pay-with-other .paypal-button {
+    width: 240px;
+    height: 48px + 48;
```
chenba, line 98: Is this the same as having `96px`?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.stories.tsx
chenba: oh haha this is for storybook.  my bad 🤦

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -55,8 +129,137 @@ export const Checkout = ({
...
+                <div
+                  className="subscription-create-pay-with-other"
+                  data-testid="pay-with-other"
+                >
+                  <Suspense fallback={<div>Loading...</div>}>
+                    <div className="paypal-button">
+                      <PaypalButton
+                        currencyCode={selectedPlan.currency}
+                        customer={null}
+                        idempotencyKey={'abc'}
```
chenba, line 193: Is there a follow-up jira for this?

chenba (changes requested):
@meandavejustice it's looking pretty good.  That checkbox doesn't seem have any effect on the PayPal button though.

packages/fxa-payments-server/src/routes/Checkout/index.scss
meandavejustice: yeah not sure why it was written this way, copied from `subscriptionCreate/index.scss`. happy to change it

meandavejustice:
@chenba 
> Oh kind of answering my question... I noticed that #9361 is listed as one of the issues that this PR fixes.

I originally mistook this issue for only disabling the submit button. I'll look at disabling the paypal button also as a part of this

meandavejustice:
@chenba Addressed comments and updated description to read as "refs #9361" so I can complete the paypal disabling without holding this layout pr any longer.

The paypal disable needs to be called from within the `onInit` function passed into the button. The only issue is how we are to call this method again, once the `disabled` prop is updated. I'd like to think about this a tad more to make sure we doing it the React ™️ way.

chenba:
> 
> 
> @chenba Addressed comments and updated description to read as "refs #9361"...

Cool, can you also update the commit message?

chenba (approved):
💰

PR #10018, opened 2021-08-03
Convert cad reminder first email

Yogita98:
## Because

- We have setup new stack in fxa auth server

## This pull request

- Converts cadReminder email template to the new stack

Closes: #9255 

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -79,7 +79,7 @@ class FluentLocalizer {
 
     const subject = await l10n.formatValue(mailSubject);
     return {
-      html: document.documentElement.outerHTML,
+      html: document.documentElement.outerHTML.replace(/&amp;/gm, '&'),
```
Yogita98, line 82: jsdom encodes the html content but since our tests were failing, I had to replace back to `&` in the template

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
```diff
@@ -3,18 +3,19 @@
...
-          href="<%= iosUrl %>"
+          href="<%- iosUrl %>"
           src="https://accounts-static.cdn.mozilla.net/product-icons/apple-app-store.png" 
           alt="iOS badge">
         </mj-image>
       </mj-column>
       <mj-column>
         <mj-image css-class="app-badges"
-          href="<%= androidUrl %>"
+          href="<%- androidUrl %>"
```
Yogita98, line 18: In ejs, `<%=` escapes the characters present in the string, because of which all the special characters present in the url were getting escaped whereas `<%-` renders the string as is

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -1337,7 +1337,7 @@ const TESTS = [
       ['X-Template-Version', { test: 'equal', expected: TEMPLATE_VERSIONS.cadReminderFirst }],
     ])],
     ['html', [
-      { test: 'include', expected: "Here&#x27;s your reminder to sync devices." },
+      { test: 'include', expected: "Here's your reminder to sync devices." },
```
Yogita98, line 1340: Again, jsdom encodes and parses the html string because of which this test was failing so I had to tweak it a bit

Yogita98:
Added a new state in Storybook to preview the `cadReminderFirst` in `ar` locale.
Modified the list of rtl locales in reference to the comment [here](https://github.com/mozilla/fxa/issues/9255#issuecomment-892240130)

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
```diff
@@ -24,13 +24,13 @@ export const appBadges = `
   </mj-section>
   <mj-section>
     <mj-column>
-      <% if (onDesktopOrTabletDevice) { %>
+      <% if (typeof(onDesktopOrTabletDevice) == "undefined" || onDesktopOrTabletDevice) { %>
```
jodyheavener, line 27: This appears to be both checking that `onDesktopOrTabletDevice` is either undefined or not. What's it meant to be doing?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -25,6 +24,10 @@ const defaultVariables = {
   preHeader: 'Random headers',
   privacyUrl: '',
   supportUrl: '',
+  headerText: "Here's your reminder to sync devices.",
+  bodyText:
+    'It takes two to sync. Syncing another device with Firefox privately keeps your bookmarks, passwords and other Firefox data the same everywhere you use Firefox.',
```
jodyheavener, line 29: How come we are setting this header and body text as variables and not directly in the template?

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -101,13 +98,13 @@ class FluentLocalizer {
 
     const isLocaleRenderedRtl = RTL_LOCALES.includes(selectedLocale);
     if (isLocaleRenderedRtl) {
-      const body = document.getElementsByTagName('body')[0];
+      const body = document.getElementsByClassName('body-80')[0];
```
jodyheavener, line 101: This is looking pretty specific. How fragile is it? Maybe we can leave a comment in the code

packages/fxa-auth-server/scripts/test-local.sh
```diff
@@ -10,7 +10,7 @@ if [ -z "$CORS_ORIGIN" ]; then export CORS_ORIGIN="http://foo,http://bar"; fi;
 
 set -u
 
-DEFAULT_ARGS="--require ts-node/register --recursive --timeout 5000 --exit"
+DEFAULT_ARGS="--require ts-node/register --recursive --timeout 50000 --exit"
```
jodyheavener, line 13: I think we'll want to return this back to its original timeout amount -- going from 5 to 50 seconds is a big change. What was it for?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -25,6 +24,10 @@ const defaultVariables = {
   preHeader: 'Random headers',
```
jodyheavener, line 24: Should this still be "Random headers"?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -12,10 +12,9 @@ export default {
...
 
 const defaultVariables = {
-  buttonText: 'Sync another device',
+  action: 'Sync another device',
   oneClickLink: true,
-  anotherDeviceUrl:
-    '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  link: '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
   iosUrl:
     'https://accounts-static.cdn.mozilla.net/product-icons/apple-app-store.png',
```
jodyheavener, line 19: This link is to an image but shouldn't it be to the iOS download page?

packages/fxa-auth-server/lib/senders/emails/partials/metadata.ts
```diff
@@ -3,17 +3,17 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export const metaData = `
-  <% if (oneClickLink) { %>
+  <% if (typeof(oneClickLink) == "string" && oneClickLink) { %>
```
jodyheavener, line 6: What's this doing?

jodyheavener (commented):
Few questions for ya

packages/fxa-auth-server/config/dev.json
```diff
@@ -427,7 +427,7 @@
     "secondInterval": "2s"
   },
   "mjml": {
-    "enabledEmailAddress": "@testuser.com$",
+    "enabledEmailAddress": "@testuser.com$|@b.com",
```
LZoog, line 430: I'm guessing the `|@b.com` this is just for testing and will be removed? I don't mind leaving `@testuser.com`

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -15,18 +15,15 @@ const OTHER_EN_LOCALES = ['en-NZ', 'en-SG', 'en-MY'];
...
   'he',
-  'khw',
   'ks',
   'ps',
+  'fa',
+  'syr',
   'ur',
-  'uz_AF',
-  'yi',
+  'ug',
```
LZoog, line 26: Thanks for taking care of this list 😄

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
LZoog: Hmm yeah, is `byTagName` not working any longer?

Yogita98: I think storybook displays everything that's inside the body tag and it's own head and body elements which is why the `rtl` styles were not getting applied to the elements. I can definitely leave a note about it

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
LZoog: Ohh, where are we even using an ampersand in our emails?

I think we want to leave character encoding in-place. It's OK if the tests have to check for `&amp` instead of `&`.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
LZoog: This might be a good thing to add to the docs in this PR.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
LZoog: We should also always use the triple equals when doing a comparison like this

packages/fxa-auth-server/lib/senders/emails/partials/button/index.scss
```diff
@@ -14,7 +14,7 @@
     background: global.$blue-400 !important;
     width: 310px !important;
     color: global.$white;
-    padding: 14px 0 !important;
+    padding: 16px 0 !important;
```
LZoog, line 17: This `16px` should be a spacing variable.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -51,3 +54,9 @@ CadReminderMobile.args = commonPropsWithOverrides({
   onDesktopOrTabletDevice: false,
 });
 CadReminderMobile.storyName = 'User is on mobile device';
+
+export const CadReminderArLocale = Template.bind({});
+CadReminderArLocale.args = commonPropsWithOverrides({
+  acceptLanguage: 'ar',
+});
+CadReminderArLocale.storyName = 'User has ar locale';
```
LZoog, line 62: Sweeeet on this + the `doc` update😏

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
Yogita98: In the mustache templates [`headerText`](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/templates/partials/cadReminder.html#L4) and other text content is being passed down from the `message` object. I wasn't sure of the best way to do the same with our storybook setup so passed down those variables from our stories file.

packages/fxa-auth-server/scripts/test-local.sh
Yogita98: Ahhh yes I forgot to revert this change. All the tests were timing out for me so I had to change it to make em work

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.ts
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { cadReminder } from '../../partials';
+
+export const render = () => {
+  return `${cadReminder}`;
+};
```
LZoog, line 9: Will we need a file like this for every template? (this review is a first pass at this PR, haven't checked in and ran things). I _think_ you can maybe simplify this entire file to something like `export {default as render} from '../../partials'` if `cadReminder` is the default export, but I didn't test it myself. I can circle back to this later, seems like it could use something though👀

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -0,0 +1,994 @@
...
+  config.smtp.sesConfigurationSet = 'ses-config';
+}
+config.smtp.subscriptionTermsUrl = 'http://example.com/terms';
+
+// Force enable the subscription transactional emails
+config.subscriptions.transactionalEmails.enabled = true;
+
+const TEMPLATE_VERSIONS = require(`${ROOT_DIR}/lib/senders/templates/_versions.json`);
+
+const SUBSCRIPTION_TERMS_URL = 'https://example.com/subscription-product/terms';
```
LZoog, line 27: I'm not sure we should add all of this stuff until we get to the Subscriptions.

Can we maybe strip out anything that's not cadReminderFirst specific and add each piece in as needed with each template conversion? That'll help clean up this file along the way and make reviewing it a bit easier. Maybe we can have this file living in a gist so we can reference any Typescript wins as other templates are being converted.

packages/fxa-auth-server/test/local/senders/email.js
LZoog: Hmm, I don't know that we should need to touch these tests.  You created that `mjml-emails` file, which should run the tests for the MJML version of the templates. I think `email.js` should remain testing the old version of templates that are in production.

packages/fxa-auth-server/config/dev.json
Yogita98: Yeah I think we're using `a@b.com` address with our tests

LZoog:
I just left a review but also meant to comment to say, awesome work @Yogita98!! 👏 I'm excited for this one to land. 🙂

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -6,37 +6,18 @@ import ejs = require('ejs');
...
-  cssBaseDir: `${__dirname}/css`,
+  validationLevel: 'strict',
 };
 
 function compile(
   context: Record<any, any>,
   templateName: string,
   subTemplate?: string
 ) {
+  const ignoreIncludes = typeof global.it === 'function';
```
Yogita98, line 19: The CI issue seems to have been resolved with this addition. What I'm doing is detecting if mocha test is running the file and accordingly setting this config parameter to ignore `<mj-include>`s. Not sure if this is the best way but let me know your thoughts on this!

Yogita98:
CI is green finally yay 🥳

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
```diff
@@ -24,13 +24,13 @@ export const appBadges = `
   </mj-section>
   <mj-section>
     <mj-column>
-      <% if (onDesktopOrTabletDevice) { %>
+      <% if (typeof onDesktopOrTabletDevice !== "undefined" && onDesktopOrTabletDevice) { %>
```
jodyheavener, line 27: I'm still confused about why this check is so explicit - would `<% if (onDesktopOrTabletDevice != null) { %>` not work?

packages/fxa-auth-server/lib/senders/emails/partials/metadata.ts
```diff
@@ -3,17 +3,17 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export const metaData = `
-  <% if (oneClickLink) { %>
+  <% if (typeof(oneClickLink) !== "undefined") { %>
```
jodyheavener, line 6: Same here, this feels strange

packages/fxa-auth-server/lib/senders/emails/renderer.ts
jodyheavener: This is definitely clever, and for now I think it'll be fine, but I would recommend leaving a comment in the code to help others understand why it was neccesary (you could even put the PR # in there)

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
jodyheavener: @Yogita98 @LZoog Here's an idea: instead of trying to satisfy Storybook in fluent-localizer (should we be trying to satisfy a dev component in prod code?), and instead of needing to add a Story with an RTL locale just to see what RTL looks like, how about we extend Storybook!

I've created this quick little toolbar extension that allows you to set the text direction in any Story (it just adds the `rtl` class). It looks like this:

Screen Shot 2021-08-05 at 10 42 30 AM

It should persist across all Stories, too, so every email template can be previewed without needing to add additional RTL-locale stories.

Here's the code: https://gist.github.com/jodyheavener/18665a6109d9373d2f752d28cb5b4ff3

Let me know what you think.

jodyheavener (changes requested):
Handful more of thoughts and questions, and I'd love your thoughts on the Storybook tool extension.

Also, I am getting the following in the terminal output:

`[fluent] Missing translations in en-US: Your Friendly Reminder: How To Complete Your Sync Setup`

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
LZoog: @jodyheavener I like that idea, nice thinking!

I wonder if we should have a front-page in Storybook similar to the front-page of the design guide with something really brief about the `doc` label, a note about this, and anything else relevant. Not something for this PR, though.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
LZoog: I think it's explicitly `undefined` in this template, but I'm not sure why we need to check `typeof`. Does `if (onDesktopOrTabletDevice) {` output the same?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -12,19 +12,23 @@ export default {
 const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
 
 const defaultVariables = {
-  buttonText: 'Sync another device',
+  action: 'Sync another device',
   oneClickLink: true,
-  anotherDeviceUrl:
-    '/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
+  link: 'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
```
LZoog, line 17: Is there a reason this was renamed `link`? Since there's other links on the page, I don't know if it's clear what link this is for 🤔

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
jodyheavener: I'm guessing it's either going to be `undefined` or boolean, so `if (onDesktopOrTabletDevice) {` may return false negatives if the value is `false`. But if we're loosely checking `onDesktopOrTabletDevice != null` or even more strictly `onDesktopOrTabletDevice === undefined` I feel like that should work

packages/fxa-auth-server/lib/senders/emails/renderer.ts
```diff
@@ -48,7 +29,10 @@ function compile(
     );
   }
   const mjmlTemplate = template(context);
-  const htmlTemplate = mjml2html(mjmlTemplate, mjmlConfig).html;
+  const htmlTemplate = mjml2html(mjmlTemplate, {
+    ...mjmlConfig,
+    ignoreIncludes: ignoreIncludes,
```
LZoog, line 34: Syntactic sugar win: this line could be simply `ignoreIncludes,`

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
Yogita98: Checking for not null will work but we also need to check if the value is true or false

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
Yogita98: It's actually coming like that from the [`email.js`](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L2906) file

jodyheavener:
So it turns out I entirely missed looking at `mjml-emails.ts` earlier because GitHub collapsed it due to its size. Having a look through it seems like a lot of the same stuff from the original test runner, but I can't imagine we need all of that present to test this template's output. Could we prune it down to only what's necessary in order to run the tests for that `TESTS` const?

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
LZoog: @jodyheavener I think either way we just care if it's falsey. In our current templates, the conditional is `{{#if onDesktopOrTabletDevice}}`. So... does `if (onDesktopOrTabletDevice) {` not work? 🤔

Though... I've got latest on this branch, and in Storybook, I'm not seeing a difference in these two stories when I think the mobile version should not render the word "desktop":
image

image

packages/fxa-auth-server/lib/senders/emails/renderer.ts
LZoog: +1 on leaving a comment

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.ts
LZoog: After checking into the branch and poking around, I see what this is doing now and what I suggested won't work (even `cadReminder as render`) since this needs to be a function.

We can change L7-9 to `export const render () => cadReminder;` at least.

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
LZoog: Also, related, Jody mentioned this in his review comment but I want to make sure it's not missed, I'm also seeing a Fluent warning `Missing translations in en-US: Your Friendly Reminder: How To Complete Your Sync Setup` when running Storybook.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
Yogita98: `if (onDesktopOrTabletDevice) {` throws the following error so maybe we also need to check if the value is defined or not
image
It's mentioned in the mustache [docs](https://www.npmjs.com/package/mustache) that `{{#variable}}` checks for undefined, null and falsy values as well.

Yogita98: @LZoog thanks for bringing it to my notice, it's been fixed now!

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -117,4 +115,4 @@ class FluentLocalizer {
     };
   }
 }
-module.exports = FluentLocalizer;
+export default FluentLocalizer;
```
Yogita98, line 118: eslint check did not pass because of this change, so I tweaked the parser options in the config file. Refer [this](https://github.com/AtomLinter/linter-eslint/issues/462)

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
jodyheavener: > are you saying we should also remove the sample AR Fluent file/AR locale story?

That was part of my intent, yeah, as it's another thing we have to keep in sync if the template itself evolves. But I can see how having a non-English locale even for one story could be nice to have around. At any rate I would still hope to avoid modifying the fluent localizer to suit the storybook setup, though, which we no longer appear to be doing 👍 

Thanks for adding that Storybook toolbar extension @Yogita98!

Yogita98: Sure, thanks for sharing the gist @jodyheavener

jodyheavener:
Hey @Yogita98 just want to check on two things that we should get done in this PR.

1. It's unclear to me how we're generating plaintext versions of the template -- when I run `write-emails` it generates _something_ in plaintext but it looks like it's the old template. I know we were having issues getting HTML/MJML->Plaintext to work well, so maybe we can just manually write the plaintext versions for now.
2. I'm trying to get your `mjml-emails.ts` test file running but am not having any luck when running `yarn test`. Are you sure it's being included there? I'm not sure our mocha tests are set up to handle TS files, so if it's being finicky I would probably suggest converting it back to a JS file for now. At any rate we'll want to verify that these new tests are actually being run in CircleCI.

Let me know if you want a hand with either of these!

Yogita98:
Hey @LZoog and @jodyheavener 
I know both of you said it was okay if the tests were in `js` but I played around a little and was able to run the `mjml-emails.ts` test file with yarn test. I verified the same by saving the output of yarn test in a text file. I looked for `lib/senders/mjml-emails` and it was there with a tick ✔️ so I assumed the file is running perfectly fine locally and the tests for `cadReminderFirst` pass.
The other change that I've made is added support for plaintext files and they are also getting localized using fluent. Let me know your thoughts on both the changes. I'll look into the eslint config soon

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.txt
```diff
@@ -0,0 +1,9 @@
+<%- body %>
+
+Mozilla. 2 Harrison St, #175, San Francisco, CA 94105
+
+fxa-privacy-url = "Mozilla Privacy Policy"
+<%- privacyUrl %>
+
+fxa-support-url = "Firefox Cloud Terms of Service"
+<%- supportUrl %>
```
Yogita98, line 9: I have added these new text files to generate plaintext version with the new templates. A pattern is set: id = "value" in these text files for fluent to get the localized string. The idea is that if the `id` doesn't exit in the fluent bundle than it will fall back to using the default value present in these text files i.e `val`. Let me know if this approach needs to be further worked upon

packages/fxa-auth-server/config/dev.json
```diff
@@ -427,7 +427,7 @@
     "secondInterval": "2s"
   },
   "mjml": {
-    "enabledEmailAddress": "@testuser.com$",
+    "enabledEmailAddress": "@mozilla.com$",
```
Yogita98, line 430: Replaced this value back to what Vijay has set in this [PR](http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device)

packages/fxa-auth-server/config/dev.json
```diff
@@ -427,7 +427,7 @@
     "secondInterval": "2s"
   },
   "mjml": {
-    "enabledEmailAddress": "@testuser.com$",
-    "templates": ["cadReminderFirst"]
+    "enabledEmailAddress": "b@c.com",
```
Yogita98, line 430: When I set the email back to what Vijay had set in his [PR](https://github.com/mozilla/fxa/commit/3c77c9039b3c6ed22638be9ff0798e170ac08340), CI tests were failing so let's use a new email for now to test the mjml-emails

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -0,0 +1,316 @@
...
+  ['unblockCode', 'unblockCode'],
+]);
+
+interface Test {
+  test: 'equal' | 'include' | 'notInclude' | any;
+  expected: string;
+}
+
+// prettier-ignore
+const COMMON_TESTS = new Map<string, Test | any>([
```
jodyheavener, line 77: Nit: It looks like this block has 3 space indentations, can we make it 2 to be consistent?

jodyheavener (approved):
This is really great, @Yogita98. I'm curious to hear @LZoog's thoughts, especially with the plaintext localization, but I feel pretty good about this. Nicely done.

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.txt
LZoog: This seems clever to me! Maybe we can tweak it later if we run into any edge cases.

LZoog (approved):
This LGTM as well. Thank you for sticking with this, what you were able to get done in just one work day after Jody and I left you those comments is really impressive! 🙂

I'm going to go ahead and merge this! :shipit: 

Will you file a follow up ticket to address Jody's nit and my `write-emails` dev config/doc ask and maybe get that up and then move on to `cadReminderSecond`/whatever you'd like to tackle next in that second milestone? So excited to see this land, great job! 🙌

PR #10022, opened 2021-08-03
send the "setup account" email

### LINKED ISSUE #9304: [subplat][content-server] Finish Account Setup View - Complete Setup and Redirect
This is a Task in the FXA-3565 Story.  Please review that first.

Once the user entered valid passwords and submitted the form, we need to use a new auth-server endpoint to finish the sign up. We will need to update the FxA client to use the new endpoint. Aside from the new endpoint, this process should be like that of the sign up process.

Redirect with the given url as the last step.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3568)

dannycoates:
closes #9304

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -1920,8 +1920,6 @@ module.exports = function (log, config) {
       code,
     } = message;
 
-    const redirectUrl = message.redirectUrl || 'https://www.mozilla.org';
-
     const enabled = config.subscriptions.transactionalEmails.enabled;
 
     log.trace('mailer.subscriptionAccountFinishSetupEmail', {
@@ -1934,7 +1932,13 @@ module.exports = function (log, config) {
       return;
     }
 
-    const query = { email, productName, token, code, redirectUrl };
+    const query = {
+      email,
+      product_name: productName,
+      token,
+      code,
+      product_id: productId,
+    };
     const template = 'subscriptionAccountFinishSetup';
     const translator = this.translator(message.acceptLanguage);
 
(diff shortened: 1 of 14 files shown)


packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js
```diff
@@ -73,7 +73,14 @@ class SetPassword extends FormView {
       const token = this._verificationInfo.get('token');
       return account.isPasswordResetComplete(token).then((isComplete) => {
         if (isComplete) {
-          return this.navigateAway(this._verificationInfo.get('redirectUrl'));
+          return account.fetchSubscriptionPlans().then((plans) => {
+            const productId = this._verificationInfo.get('productId');
+            const plan = plans.find((p) => p.product_id === productId);
+            const url = plan
+              ? plan.product_metadata.downloadURL
+              : 'https://mozilla.org';
```
dannycoates, line 81: @vbudhram @chenba I think it's better to _pull_ the redirect url than to push it. We don't want an open redirect so we need to validate `redirectUrl`, but against what... a list of valid urls... which we get from the server... 

I think you see where I'm going.

vbudhram: @dannycoates Looks like you updated to fetch the url, which I think works better 👍🏽

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1934,7 +1932,7 @@ module.exports = function (log, config) {
       return;
     }
 
-    const query = { email, productName, token, code, redirectUrl };
+    const query = { email, productName, token, code, product_id: productId };
```
vbudhram, line 1935: nit: Looks like mix of camel and snake case. I think I introduced productName so wouldn't mind if converted to snake case.

vbudhram (approved):
@dannycoates LGTM 👍🏽

PR #10042, opened 2021-08-05
fix(payments): add checkout route to INDEX_ROUTES

### LINKED ISSUE #10041: Navigating to the new checkout route on stage returns a 404
See this [slack thread](https://mozilla.slack.com/archives/C4D36CAJW/p1628175054008600). Per Barry, this is because the new route isn’t added to `INDEX_ROUTES` in fxa-payments-server/server/lib/server.js.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3826)

biancadanforth:
Because:

* We are seeing a 404 on stage for the new checkout route

This commit:

* Ensures we correctly route to index.html for this new route.

Closes #10041

## Checklist

Regression introduced by #9793

### DIFF
--- a/packages/fxa-payments-server/server/lib/server.js
+++ b/packages/fxa-payments-server/server/lib/server.js
@@ -102,11 +102,8 @@ module.exports = () => {
   const INDEX_ROUTES = [
     '/',
     '/subscriptions',
+    '/checkout/:productId',
     '/products/:productId',
-    '/v1/subscriptions',
-    '/v1/products/:productId',
-    '/v2/subscriptions',
-    '/v2/products/:productId',
   ];
 
   app.disable('x-powered-by');


packages/fxa-payments-server/server/lib/server.js
```diff
@@ -107,6 +107,7 @@ module.exports = () => {
     '/v1/products/:productId',
     '/v2/subscriptions',
     '/v2/products/:productId',
+    '/v1/checkout/:productId',
```
chenba, line 110: I don't think we need the '/v1' part.

chenba (changes requested):
'/v1' needn't be part of the route.

packages/fxa-payments-server/server/lib/server.js
biancadanforth: Admittedly I was cargoculting -- but why is it there for the other routes?

biancadanforth: I guess I see without, with a v1 and a v2.

chenba (approved):

packages/fxa-payments-server/server/lib/server.js
chenba: Yeah, for a while we needed all those.  It should be safe to clean out the v1 and v2 stuff at this point.

biancadanforth: I'll do that here while I'm in here.

biancadanforth:
Latest one removes all the `/v1` and `/v2` routes.

PR #10050, opened 2021-08-06
chore(subscriptions): fix a scope rename that should've been in 3ced8935b

chenba:
Not sure how I missed that one spot.  But the subscription management scope was moved into a shared constant and renamed, and somehow I changed just one of two instances it in a particular file.


chenba:
Actually might've been a rebase that caused it.  😦

biancadanforth (approved):

PR #10051, opened 2021-08-06
feat(subscriptions): add an optional cb called in PayPal's createOrder

chenba:
Because:
 - we need a way to create a Firefox account before the PayPal button
   gets the PayPal checkout token

This commit:
 - add an optional callback to PayPalButton's props that's called at the
   start of its createOrder function

part of #9358; follows #10007


chenba:
See https://github.com/mozilla/fxa/blob/ac153ec7df4dc73faba894d97779d0a4b8eed605/packages/fxa-payments-server/src/components/PayPalButton/index.test.tsx#L51-L52 on why there are no accompanying tests.

Issue #5705, opened 2020-06-18
Create new email template for `remind me later` for CAD


data-sync-user:
Template for first reminder https://docs.google.com/document/d/1I7TSNa7mvUFipX48OhGSRbuVybIqSb5QdnCmivSjl-c/edit#bookmark=id.svmpy2gtxsjy

Template for second reminder https://docs.google.com/document/d/1I7TSNa7mvUFipX48OhGSRbuVybIqSb5QdnCmivSjl-c/edit#bookmark=id.svmpy2gtxsjy

We should be able to build one email template and pass in the contents via code. Docs for creating a new email [template](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-email)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2142)
┆Issue Number: FXA-2142

data-sync-user:
➤ Alex Davis commented:

There were some open questions about which subject line to use for the reminder email. We discussed and it should be:
Your Friendly Reminder: How-To Complete Your Sync Setup

vasilicatamas:
The "remind me later" emails are successfully received after 8 and 72 hours. 2 new issues (#5798 and #5799) were found and filed separately . Based on this I am marking this bug as verified.

PR #10084, opened 2021-08-11
feat(fxa-auth-server): Convert cadReminderSecond email to new stack

Yogita98:
## Because

- We want to convert cadReminderSecond template to new stack

## This pull request

- Converts the said template to new templating stack

Closes: #9265 

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/senders/emails/templates/cadReminderSecond/index.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+    'Syncing another device with Firefox privately keeps your bookmarks, passwords and other Firefox data the same everywhere you use Firefox.',
+  acceptLanguage: 'en-US',
+};
+
+const commonPropsWithOverrides = (
+  overrides: Partial<typeof defaultVariables> = {}
+) =>
+  Object.assign({
+    template: 'cadReminderSecond',
+    layout: 'fxa',
```
LZoog, line 39: Does it look like we're going to use `layout: 'fxa'` for all or nearly all of our emails, and any other common props (`acceptLanguage: 'en-US'`)? We can probably make something shared to pull those kinds of props in.

This is more of an aside and not something to do here but I wonder if there's a way to share some of the `defaultVariables` strings with the English Fluent files so there's less copy-and-paste (and then if some copy changes, we won't have to change it in two spots). 🤔

packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts
```diff
@@ -8,13 +8,13 @@ export const cadReminder = `
...
+      <mj-text css-class="header-text"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
       <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
     </mj-column>
   </mj-section>
   <mj-section>
     <mj-column>
-      <mj-text css-class="primary-text"><span data-l10n-id="sync-description"><%- bodyText %></span></mj-text>
+      <mj-text css-class="primary-text"><span data-l10n-id="<%- templateName %>-description"><%- bodyText %></span></mj-text>
     </mj-column>
   </mj-section>
```
LZoog, line 19: One super minor thing I noticed that we maybe didn't catch the first time, this is the old version of `cadReminder`:

image


This is the new version:
image

I think we need more space between the the buttons and the "or, install this on..." text.

LZoog (commented):
Things are looking great! Were you able to get past that test you mentioned via Matrix? The only failure I'm seeing in CI I'm also able to see locally, but it's a different one. I'm not sure why it's occurring just yet, when I remove your new tests I still see it so I don't think it's your new tests, but I know I've seen that "Please restart the server to begin using the new keys" message before and can't recall any details. I messaged Vijay about it in case he's seen it before, if he doesn't get a chance to get back to you/us by the morning and you address the PR comments, I think you're probably good to branch off of `main` and start on the next template while we investigate. 🙂

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderSecond/index.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+    doc: 'Connect Another Device (CAD) reminder emails are sent out 24 hours after a user clicks "send me a reminder" on the connect another device page.',
```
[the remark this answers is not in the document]
LZoog: @Yogita98 looks like QA commented here: https://github.com/mozilla/fxa/issues/5705#issuecomment-651146680

According to this, the first reminder is sent out after 8 hours, not 24, and then `cadReminderSecond` is sent after 72 hours.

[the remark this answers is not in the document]
vbudhram: Those times are accurate per https://github.com/mozilla/fxa/blob/fe315a8cfe79de8a6253cb5cc4d84bc71c303878/packages/fxa-auth-server/config/index.ts#L1760

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -61,6 +67,10 @@ $s-10: 40px;
   }
 }
 
+.ml-half img {
+  margin-left: 2px !important;
+}
+
```
Yogita98, line 73: I couldn't declare a spacing variable for 2px since it's supposed to be `0.5` per Tailwind docs and `sass` didn't allow naming a variable `$s-0.5`

Yogita98:
Rebased this branch against main after the other PR got merged. Also, added `commonArgs` for shared variables in storybook wherever possible!

packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
```diff
@@ -9,9 +9,22 @@ export interface StorybookEmailArgs {
...
+    'https://www.mozilla.org/privacy?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-privacy',
+  supportUrl:
+    'https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-support',
+};
+
 export const storybookEmail = ({
   template,
-  layout,
+  layout = 'fxa',
   doc,
```
LZoog, line 28: I think we should  add `acceptLanguage = 'en-US'` here like you gave `layout` the default value rather than include it in that `commonArgs` const since that's going to be the case across all of our templates in Storybook. `androidUrl`, `iosUrl`, etc. look cadReminder specific and probably shouldn't live in this file - we can rename that to something like `cadReminderCommonVars` and move that constant into `cadReminderFirst`, and then export it so you can import it in `cadReminderSecond`.

If we allow `layout` and `acceptLanguage` to have a default var, you can remove both of those in lowRecoveryCodes ([here](https://github.com/mozilla/fxa/pull/10092/files#diff-15a720ead2be22b837ac545709a374ad40147dd71ae3cacc3ed7ceec9449736cR21) and [here](https://github.com/mozilla/fxa/pull/10092/files#diff-15a720ead2be22b837ac545709a374ad40147dd71ae3cacc3ed7ceec9449736cR30)) and you won't have to reference those in other stories unless you need to override them. 👍

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -158,7 +186,7 @@ const TESTS: [string, any][] = [
       { test: 'include', expected: configUrl('supportUrl', 'low-recovery-codes', 'support') },
       { test: 'notInclude', expected: 'utm_source=email' },
     ]],
-  ])],
+  ])],
```
LZoog, line 189: Tiny nit: there's some extra spaces here.

LZoog (commented):
Just a couple more comments. I'll circle back to this tomorrow morning and likely r+ and merge. 😁

packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
```diff
@@ -5,20 +5,29 @@
 export interface StorybookEmailArgs {
   template: string;
   layout: string;
+  acceptLanguage: string;
   doc?: string;
   variables: Record<string, any>;
 }
 
+export const commonArgs = {
+  link: 'http://localhost:3030/connect_another_device?utm_medium=email&utm_campaign=fx-cad-reminder-first&utm_content=fx-connect-device',
```
LZoog, line 14: I see `fxa-cad-reminder-first` in `link`, `privacyUrl`, and `supportUrl` here. Anything template-specific should live in that template, so `link` should be moved to `cadReminderFirst` since it goes to `/connect_another_device`. We have a few options for `privacyUrl` and `supportUrl` which we appear to use across templates...

1) Change them to `#`. Since this is just Storybook, it's really just serving as a visual tool for us - if we ever change any of the parameters on the URL here, or the URL in general, the default vars we set in Storybook here is going to be out of date.
2) We can strip out the CAD-reminder params, in this case `utm_campaign=fx-cad-reminder-first`, and only use the link + `utm_medium=email` + `utm_content`. We might want to leave a comment in the code just to be explicitly clear that other parameters exist in production, so `privacyUrl` for example would become: ```https://www.mozilla.org/privacy?utm_medium=email&utm_content=fx-privacy` // in production, other `utm` parameters may exist```
3) turn `commonArgs` into a function that accepts the name of the template, which would turn `privacyUrl` for example to something like: `https://www.mozilla.org/privacy?utm_medium=email&utm_campaign=fx-${emailName}`
4) We can strip out the params entirely. `privacyUrl` would become `https://www.mozilla.org/privacy` with a similar comment in the code about UTM params existing in production. **I think this is what I'd do personally,** it gives the Storybook viewer a chance to see the actual URL but without us having to think about/manage the UTM params in multiple places. If someone wants to see what they will actually end up being, they can use `yarn write-emails`.

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderSecond/index.stories.ts
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Story, Meta } from '@storybook/html';
+import storybookEmail, { StorybookEmailArgs } from '../../storybook-email';
+import * as cadReminderFirst from '../cadReminderFirst/index.stories';
```
LZoog, line 7: We don't need to import everything (`*`) out of this template. If you change [this line](https://github.com/mozilla/fxa/blob/f6e682e795b9c3a537574fb8aff8f18921a98c54/packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts#L17) to `export defaultVariables = {`, you can use `import { defaultVariables as cadReminderDefaultVariables } from '../cadReminderFirst/index.stories'`. (or rename `defaultVariables` in that file to use `cadReminder` so you don't have to rename it here)

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderSecond/index.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export default {
+  title: 'Emails/cadReminderSecond',
+} as Meta;
+
+const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
+
+const defaultVariables = {
+  ...cadReminderFirst.CadReminderDesktopTablet.args.variables,
+  bodyText:
```
LZoog, line 17: I think you can take this line out. `bodyText` is identical to what you've got in `cadReminderFirst` vars so we shouldn't need to override it.

Yogita98: `cadReminderFirst` has this extra text: `It takes two to sync.` in body so is it okay if we crop it out of `cadReminderSecond`?

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderSecond/index.stories.ts
Yogita98: I think initially I tried to import the args this way but since they are getting exported from a stories file, storybook is treated it as a story and threw errors
image

LZoog: Oooh interesting! We could create an additional "shared" file for stuff like this but eh since it's just for storybook, we'll only have shared args like this between emails that have first version/second etc. and you're doing what the docs recommend I say let's leave it like how you have it for now👍

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -3,41 +3,41 @@
...
 export default {
-  title: 'Emails/cadReminder',
+  title: 'Emails/cadReminderFirst',
 } as Meta;
 
 const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
 
 const defaultVariables = {
+  ...commonArgs,
+  link: 'http://localhost:3030/connect_another_device',
```
Yogita98, line 19: I stripped out the `utm` params from here as well as from `lowRecoverCodes`

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
```diff
@@ -25,11 +25,11 @@ export const appBadges = `
   <mj-section>
     <mj-column>
       <% if (locals.onDesktopOrTabletDevice) { %>
-      <mj-text css-class="secondary-text mb-3"><span data-l10n-id="another-desktop-device"> Or, install on
+      <mj-text css-class="secondary-text mb-3 mt-2"><span data-l10n-id="another-desktop-device"> Or, install on
```
LZoog, line 28: This seems like such a small thing, but since the images themselves already have `mb-3`, can we just increase that to `mb-5` and remove `mt-2` here? If there's elements one after another, I try to keep required margins between them on one or the other element instead of a margin-bottom on one and a margin-top on the other. Since we reuse these app badges in other templates, it'll ensure that text after that is always `mb-5` across our templates, rather than having to add `mt-2` to text elements that follow (I think if you left this as-is, it'd require you adding `mt-2` to text elements in `postVerify` to match the spacing).

LZoog (approved):
Looking awesome! I left just one more comment about the margin after checking this out again locally. Going to r+, but will merge after that last comment is addressed 😄

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
Yogita98: Ohhh yeah I didn't think about other templates, but thanks for such a detailed explanation :)

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.ts
```diff
@@ -25,11 +25,11 @@ export const appBadges = `
   <mj-section>
     <mj-column>
       <% if (locals.onDesktopOrTabletDevice) { %>
-      <mj-text css-class="secondary-text mb-3"><span data-l10n-id="another-desktop-device"> Or, install on
+      <mj-text css-class="secondary-text mb-5"><span data-l10n-id="another-desktop-device"> Or, install on
```
LZoog, line 28: Ooh @Yogita98 I think the wrong class was changed. I think you can actually remove this `mb-5` from this text el, this is what I'm seeing right now:

image

The `margin` between those two pieces of text is coming from the text below it:
image


That spacing is fine (the class can be removed and there will be no visual change), but what I'm thinking we need to change is `mb-2` on app badges:

image

Which looks like needs changing in `emails/partials/appBadges/index.scss` from `global.$s-3` to `global.$s-5`.

Yogita98: Ahhh yes. Even I couldn't notice any visual changes when I kept the versions parallel to each other, thought something's wrong with my eyes lol. Btw are you using any tool while inspecting HTML, those dotted lines and colored spacing looks 🔥

LZoog: When there's true adjacent elements, [the top/bottom margin actually gets collapsed](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) to whatever value is larger. Padding doesn't do that for what it's worth. 🙂

I'm actually just using Firefox Nightly! 😁 Firefox Developer Edition has some cool looking tools too though, sometimes I'll use that.

PR #10120, opened 2021-08-16
feat(auth-server): Convert postVerify template to new stack

Yogita98:
## Because

## This pull request

- Converts postVerify email template to new stack

Closes: #9277 

## Checklist

## Screenshots (Optional)


packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -44,16 +38,8 @@ const commonPropsWithOverrides = (
     },
   });
 
-export const CadReminderDesktopTablet = Template.bind({});
-CadReminderDesktopTablet.args = commonPropsWithOverrides({
-  onDesktopOrTabletDevice: true,
-});
-CadReminderDesktopTablet.storyName = 'User is on desktop or tablet device';
-
```
Yogita98, line 52: @LZoog I removed the desktop state from both the cadReminders. I think we do not want to show the desktop version to the user and since we're also not passing the `onDesktopOrTabletDevice` from [`email.js`](https://github.com/mozilla/fxa/blob/949a3d73e56832bf855fdec18f9b55901fa65acd/packages/fxa-auth-server/lib/senders/email.js#L2902) file, it made sense to remove that state entirely from storybook as well

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { appBadges, button, automatedEmailNoAction } from '../../partials';
+
+export const render = () => `
+  <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
+   <mj-section>
+     <mj-column>
+       <mj-text css-class="primary-text"><span data-l10n-id="postVerify-subdescription">Firefox Account verified. You're almost there.</span></mj-text>
+       <mj-text css-class="header-text"><span data-l10n-id="postVerify-title">Next sync between your devices!</span></mj-text>
+       <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
```
jodyheavener, line 13: I think we need to cap the height of this image as it looks a little distorted, as well as add space below it -- left is original right is new template

Screen Shot 2021-08-17 at 3 08 17 PM

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.stories.ts
```diff
@@ -0,0 +1,47 @@
...
+import storybookEmail, {
+  StorybookEmailArgs,
+  commonArgs,
+} from '../../storybook-email';
+
+export default {
+  title: 'Emails/postVerify',
+} as Meta;
+
+const Template: Story<StorybookEmailArgs> = (args) => storybookEmail(args);
```
jodyheavener, line 15: Since this is duplicated in every story we may as well export it from `storybook-email.ts` and just import it into each story file, no?

packages/fxa-auth-server/package.json
```diff
@@ -33,7 +33,7 @@
     "maildev-preview": "ts-node ./lib/senders/emails/preview-in-maildev.ts",
     "start-mjml": "NODE_ENV=dev ts-node ./bin/mjml-server.ts",
     "storybook": "npm run emails-scss && start-storybook -p 6010 --no-version-updates",
-    "storybook-mjml": "npm-run-all -p start-mjml storybook"
+    "storybook-mjml": "nodemon -e scss,ts  --exec \"npm-run-all -p start-mjml storybook\""
```
jodyheavener, line 36: Still hoping to find a better solution for this.. will follow up shortly

jodyheavener (changes requested):
Couple thoughts, looks great though

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.ts
Yogita98: Ohhh yes, thanks for bringing it to my notice!

packages/fxa-auth-server/package.json
LZoog: FWIW if we don't find a better solution, I think this is fine for now if it covers the basis we need it for just to make developing these templates for at least a few more days a bit less painful, we can maybe even revisit in a few weeks.

jodyheavener (approved):

PR #10122, opened 2021-08-16
feat(auth-server): Convert postRemoveSecondary template to new stack

Yogita98:
## Because

## This pull request

- Converts postRemoveSecondary email template to new stack 

Closes: #9273 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
@@ -78,7 +78,7 @@ class FluentLocalizer {
     let selectedLocale: string = 'en-US';
 
     async function* generateBundles(currentLocales: Array<string>) {
-      let bundle = new FluentBundle(currentLocales);
+      let bundle = new FluentBundle(currentLocales, { useIsolating: false });
       for (const locale of currentLocales) {
         let source = await fetchResource(locale);
         if (source !== '' && locale !== 'en-US') {
--- a/packages/fxa-auth-server/lib/senders/emails/partials/manageAccount/index.txt
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/manageAccount/index.txt
@@ -1,0 +1,2 @@
+manage-account-plaintext = "Manage account:"
+<%- link %>
--- a/packages/fxa-auth-server/test/local/senders/mjml-emails.ts
+++ b/packages/fxa-auth-server/test/local/senders/mjml-emails.ts
@@ -163,9 +163,7 @@ const TESTS: [string, any][] = [
   
   ['lowRecoveryCodesEmail', new Map<string, Test | any>([
     ['subject', [
-      { test: 'include', expected: '2' },
-      { test: 'include', expected: 'recovery codes remaining' },
-      { test: 'notInclude', expected: '1' },
+      { test: 'include', expected: '2 recovery codes remaining' }
     ]],
     ['headers', new Map([
       ['X-Link', { test: 'equal', expected: configUrl('accountRecoveryCodesUrl', 'low-recovery-codes', 'recovery-codes', 'low_recovery_codes=true', 'email', 'uid') }],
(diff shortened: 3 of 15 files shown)


packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -78,7 +78,7 @@ class FluentLocalizer {
     let selectedLocale: string = 'en-US';
 
     async function* generateBundles(currentLocales: Array<string>) {
-      let bundle = new FluentBundle(currentLocales);
+      let bundle = new FluentBundle(currentLocales, { useIsolating: false });
```
Yogita98, line 81: I think by default, fluent inserts unicode characters while translating the variables so disabling the `useIsolating` option can help us avoid that. Reference: [fluent docs](https://projectfluent.org/fluent.js/bundle/classes/fluentbundle.html)

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -163,9 +163,7 @@ const TESTS: [string, any][] = [
   
   ['lowRecoveryCodesEmail', new Map<string, Test | any>([
     ['subject', [
-      { test: 'include', expected: '2' },
-      { test: 'include', expected: 'recovery codes remaining' },
-      { test: 'notInclude', expected: '1' },
+      { test: 'include', expected: '2 recovery codes remaining' }
```
Yogita98, line 166: Since we disabled the `useIsolating` option in fluent, we no longer need to check for specific values and can rather test for the entire phrase together just like the old tests

chenba:
A combination of https://github.com/mozilla/fxa/issues/10123 and that this PR is created from a branch of a fork, so the PR doesn't get the same env vars in CI, is causing at least some of CI failures.

chenba:
> 
> 
> A combination of #10123 and that this PR is created from a branch of a fork, so the PR doesn't get the same env vars in CI, is causing at least some of CI failures.

When a PR is created from a branch on mozilla/fxa, some additional env vars are applied in the CI tests, including Subscription Platform related ones.  #10123 is not causing a CI failure for those PRs because of that.  But when a PR is created from a branch of a fork, there's currently a 100% chance of a CI failure due to #10123.

jodyheavener (commented):
Couple thoughts, visually looks great though

packages/fxa-auth-server/lib/senders/emails/partials/manageAccount/index.txt
```diff
@@ -0,0 +1,2 @@
+manage-account = "Manage account:"
+<%- link %>
```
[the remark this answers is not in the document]
Yogita98: There is a `links` object as well, but I couldn't find any other `link` variables being passed down by the renderer

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
LZoog: What did this look like without vs with? From reading the docs it seems fine to set this to `false` since we are setting the text-direction with explicit markup which it seems to recommend, but just curious what it's outputting otherwise.

Yogita98: Remember I was talking about some test failures happening locally last week? We introduced usage of `selectors` with our fluent setup so when fluent was translating those selectors, my guess is that it was also adding some unicode characters around the text. 
If you still want to see how localization happens otherwise, turn useIsolating to true which is basically the default value and run the `mjml-emails` tests. You'll find the tests failing despite the expected text matching the actual one
Also, see this [issue](https://github.com/projectfluent/fluent.js/issues/484)

LZoog: Oh I see! Thanks for the explaination. 🙂

chenba:
> 
> When a PR is created from a branch on mozilla/fxa, some additional env vars are applied in the CI tests, including Subscription Platform related ones. #10123 is not causing a CI failure for those PRs because of that. But when a PR is created from a branch of a fork, there's currently a 100% chance of a CI failure due to #10123.

I think what we could do as a workaround, once this PR is approved, is to have someone on the team push the branch to mozilla/fxa and create a PR with that.  But keep this PR open.  Github will notice that the commit hashes are identical, and apply the CI results to this PR.  Pretty sure.

jodyheavener:
@chenba genius

Yogita98:
Thanks for all the help and support @chenba, we appreciate it :)

jodyheavener (approved):
yay

PR #10125, opened 2021-08-16
Use JWT for finish_setup

### LINKED ISSUE #10043: Handle expired tokens in finish_account_setup
The passwordForgotToken expires after 24h. If the user clicks the link from their setup email after that time we should have some UX to allow them to get a new token/code.

One way would be to add a created_at querystring parameter to the link, then if it's past expiration show a button to send a new code and have a code field before showing the set password form.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3828)

dannycoates:
Instead of using the existing password reset flow, where the `passwordForgotToken` expires, this method creates a jwt that can be used until a password is set on the account.

closes #10043

### DIFF
--- a/packages/fxa-auth-client/lib/client.ts
+++ b/packages/fxa-auth-client/lib/client.ts
@@ -448,6 +448,16 @@ export default class AuthClient {
     return accountData;
   }
 
+  async finishSetup(token: string, email: string, newPassword: string) {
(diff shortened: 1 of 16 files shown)


bbangert (approved):
LGTM!

vbudhram (commented):
Thanks @dannycoates!

PR #10074, opened 2021-08-10
task(fxa-payments-server): add legal links to checkout form

meandavejustice:
## This pull request

- Adds legal links for products and fxa to the payment form on checkout page.

## Issue that this pull request solves

Closes: #9794

## Checklist

_Put an `x` in the boxes that apply_

- [x] My commit is GPG signed.
- [x] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [x] I have verified that my changes render correctly in RTL (if appropriate).

## Screenshots

image

Issue #9365, opened 2021-06-21
[new user checkout] add form submitted success messaging


data-sync-user:
> Proposal (Tony/Barry): Instead of the “Continue to download” button, perhaps “Check your email to finish setting up…” (message TBD) (3 points)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3668)

data-sync-user:
➤ Bianca Danforth commented:

The MVP designs ( https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/%F0%9F%8E%A1MVP-Subscription-Optimization-Flow?node-id=400%3A2098 ) now have the final screen. Note: We will want to use the TermsAndPrivacy component updated in https://mozilla-hub.atlassian.net/browse/FXA-3715 ( https://mozilla-hub.atlassian.net/browse/FXA-3715|smart-link ) to also show the FxA account links on this page.

PR #10127, opened 2021-08-17
 feat(payments): handle new/previously unhandled errors in passwordless checkout

### LINKED ISSUE #9366: [new user checkout] add form submitted error messaging
* [Confirm with Ben] What are the classes of errors here? Do we need any new error messages/strings? 
* Due diligence to ensure we display all errors correctly, we have existing error page in place.(5 points)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3669)

biancadanforth:
**Important:** This PR is based on Dave's PR #10074 for FXA-3715; it should not be merged until that PR merges. Only the second commit is new.

Because:
* The new passwordless checkout flow adds two new types of errors (account creation and newsletter signup) that we need to handle, and there is a previously unhandled error related to loading the success page.
* These errors should all be caught, and we should display an appropriate and relevant error message to the user when they occur if possible.

This commit:
* Makes sure we handle the following errors in the new passwordless flow and display error messages based on the MVP designs.
  - FxA account creation
  - Subscription creation/payment
  - Newsletter signup
  - Success screen loading

Closes #9366

## Checklist

## Screenshots (Optional)

Note: The success page (including legal links) is not final; it will be updated in #9365 .

Screenshot 2021-08-16 at 23-01-03 Firefox Accounts

I also added the new legal links to the `PaymentErrorView` component.

Screenshot 2021-08-17 at 00-45-38 Firefox Accounts

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -40,8 +40,8 @@ duplicate-transaction = Hmm. Looks like an identical transaction was just sent. 
 coupon-expired = It looks like that promo code has expired.
 card-error = Your transaction could not be processed. Please verify your credit card information and try again.
 
-fxa-signup-error = There was a problem creating your account.  Please try again later.
-fxa-newsletter-signup-error = There was a problem subscribing you to the newsletter.
+fxa-account-signup-error = A system error caused your ${ productName } sign-up to fail. Your payment method has not been charged. Please try again.
+newsletter-signup-error = You're not signed up for product update emails. You can try again in your account settings.
 fxa-post-passwordless-sub-error = Subscription confirmed, but the confirmation page failed to load. Please check your email to set up your account.
 
 ## settings
--- a/packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { useHistory } from 'react-router-dom';
 import { Localized } from '@fluent/react';
 import { StripeError } from '@stripe/stripe-js';
@@ -10,6 +10,8 @@ import TermsAndPrivacy from '../TermsAndPrivacy';
 
 import './index.scss';
 import { Plan } from '../../store/types';
+import PaymentLegalBlurb from '../PaymentLegalBlurb';
+import AppContext from '../../lib/AppContext';
 
 export type PaymentErrorViewProps = {
   onRetry: Function;
@@ -17,6 +19,7 @@ export type PaymentErrorViewProps = {
   error?: StripeError | GeneralError;
   className?: string;
   subscriptionTitle?: React.ReactElement<SubscriptionTitle>;
+  isPasswordlessCheckout?: boolean;
 };
 
 const retryButtonFn = (onRetry: PaymentErrorViewProps['onRetry']) => (
@@ -51,8 +54,10 @@ export const PaymentErrorView = ({
   error,
   className = '',
   subscriptionTitle,
+  isPasswordlessCheckout = false,
 }: PaymentErrorViewProps) => {
   const history = useHistory();
+  const { config } = useContext(AppContext);
 
   // We want the button label and onClick handler to be different depending
   // on the type of error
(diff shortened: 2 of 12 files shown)


packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
```diff
@@ -88,8 +95,18 @@ export const PaymentErrorView = ({
...
+          case, we do not want a "Try again" or "Manage subscription" button. */}
+          {error.code !== 'fxa_fetch_profile_customer_error' ? (
+            <ActionButton data-testid={'error-view-action-button'} />
+          ) : null}
+          <PaymentLegalBlurb provider={undefined} />
+          <TermsAndPrivacy
+            showFXALinks={isPasswordlessCheckout}
+            plan={plan}
+            contentServerURL={config.servers.content.url}
+          />
```
biancadanforth, line 109: For this error case where the user subscribes successfully, but we fail to load the success page, what kind of button could we have here?

We probably don’t want the "Try again" label, since the sub was created successfully, and since they haven't set a password, they aren't logged in and couldn't go to "Manage subscriptions".

Give this, I propose no action button for this error case, but I am open to suggestions.

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -40,8 +40,8 @@ duplicate-transaction = Hmm. Looks like an identical transaction was just sent.
 coupon-expired = It looks like that promo code has expired.
 card-error = Your transaction could not be processed. Please verify your credit card information and try again.
 
-fxa-signup-error = There was a problem creating your account.  Please try again later.
-fxa-newsletter-signup-error = There was a problem subscribing you to the newsletter.
+fxa-account-signup-error = A system error caused your ${ productName } sign-up to fail. Your payment method has not been charged. Please try again.
```
chenba, line 43: Unsure if we need to leave comment to explain what $productName is to the localizers.  Couldn't hurt though.

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
chenba: Agreed.  The only thing we want the new customer to do is to check their email at this point.

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
```diff
@@ -88,8 +95,18 @@ export const PaymentErrorView = ({
...
         <div className="footer" data-testid="footer">
-          <ActionButton />
-          <TermsAndPrivacy plan={plan} />
+          {/* This error code means the subscription was created successfully, but
+          there was an error loading the information on the success screen. In this
+          case, we do not want a "Try again" or "Manage subscription" button. */}
+          {error.code !== 'fxa_fetch_profile_customer_error' ? (
+            <ActionButton data-testid={'error-view-action-button'} />
+          ) : null}
+          <PaymentLegalBlurb provider={undefined} />
```
chenba, line 104: Probably don't need to be part of this patch, but assuming `provider` here meaning payment provider, I think we can get that info to this scope.

chenba (approved):
LGTM!

packages/fxa-payments-server/src/components/PaymentErrorView/index.tsx
biancadanforth: We discussed this, and we don't have a `customer` to get the `customer.payment_provider` off of until the last function called, `fetchProfileAndCustomer` on form submit, so `customer` wouldn't be defined until most of the calls that could error have been made to this point.

There may be another way to get this information, but it isn't easily available, so I'm going to leave it for now. Also the [MVP designs](---https://www.figma.com/file/cOIAuHIIbrPbAlAIcpejNM/🎡MVP-Subscription-Optimization-Flow?node-id=1064%3A15109) show both Stripe and PayPal privacy policies on this screen.

packages/fxa-payments-server/public/locales/en-US/main.ftl
biancadanforth: Whoops I missed this; I [asked Vijay](https://github.com/mozilla/fxa/pull/10138#issuecomment-900707371) to add a comment if possible in his WIP PR for the success screen.

PR #10136, opened 2021-08-17
feat(auth-server): Convert verificationReminder templates to new stack

Yogita98:
## Because

## This pull request

- Converts verificationReminderFirst and verificationReminderSecond email templates to new stack

Closes: #9281 and #9285 

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/global.scss
+++ b/packages/fxa-auth-server/lib/senders/emails/global.scss
@@ -49,19 +49,18 @@ $s-10: 40px;
 }
 
 .mt {
-  &-2 div {
+  &-2 {
     margin-top: $s-2 !important;
   }
-  &-6 div {
+  &-6 {
     margin-top: $s-6 !important;
   }
 }
 
 .mb {
-  &-3 div {
+  &-3 {
     margin-bottom: $s-3 !important;
   }
-
   &-5 {
     margin-bottom: $s-5 !important;
   }
@@ -84,22 +83,46 @@ tr > td:first-child {
   font-family: $font-sans;
 }
 
-.header-text div {
+.text-header div {
   @extend .text-xl;
   @extend .font-sans;
+  @extend .mb-3;
   text-align: center !important;
+}
+
+%text-body-common {
+  @extend .text-sm;
+  @extend .font-sans;
+  text-align: center !important;
+}
+
+.text-body div {
+  @extend %text-body-common;
   @extend .mb-5;
 }
 
-.primary-text div {
-  @extend .text-sm;
-  @extend .mb-5;
+.text-sub-body div {
+  @extend %text-body-common;
+  @extend .mb-3;
+}
+
+.text-footer div {
+  @extend .text-xs;
   @extend .font-sans;
   text-align: center !important;
+  color: $grey-400 !important;
 }
 
 .sync-logo img {
   width: 240px !important;
   margin: 0 auto;
   @extend .mb-5;
+  @extend .mt-2;
+}
+
+.graphic-devices img {
+  width: 240px !important;
+  margin: 0 auto;
+  @extend .mb-5;
+  @extend .mt-2;
 }
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
@@ -40,9 +40,9 @@ export const render = (body: string) => {
       ${body}
       <mj-section>
         <mj-column>
-          <mj-text css-class="secondary-text">Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</mj-text>
-          <mj-text css-class="secondary-text"><a class="link-blue" data-l10n-id="fxa-privacy-url" href="<%- privacyUrl %>">Mozilla Privacy Policy</a></mj-text>
-          <mj-text css-class="secondary-text"><a class="link-blue" data-l10n-id="fxa-service-url" href="https://www.mozilla.org/about/legal/terms/services/">Firefox Cloud Terms of Service</a></mj-text>
+          <mj-text css-class="text-footer">Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</mj-text>
+          <mj-text css-class="text-footer"><a class="link-blue" data-l10n-id="fxa-privacy-url" href="<%- privacyUrl %>">Mozilla Privacy Policy</a></mj-text>
+          <mj-text css-class="text-footer"><a class="link-blue" data-l10n-id="fxa-service-url" href="https://www.mozilla.org/about/legal/terms/services/">Firefox Cloud Terms of Service</a></mj-text>
         </mj-column>
       </mj-section>
     </mj-body>
--- a/packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts
@@ -5,15 +5,16 @@
 import { appBadges, button, automatedEmailNoAction } from '../../partials';
 
 export const cadReminder = `
+  <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
   <mj-section>
     <mj-column>
-      <mj-text css-class="header-text"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
-      <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
+      <mj-text css-class="text-header"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
+      <mj-image css-class="graphic-devices" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
     </mj-column>
   </mj-section>
   <mj-section>
     <mj-column>
-      <mj-text css-class="primary-text"><span data-l10n-id="<%- templateName %>-description"><%- bodyText %></span></mj-text>
+      <mj-text css-class="text-body"><span data-l10n-id="<%- templateName %>-description"><%- bodyText %></span></mj-text>
     </mj-column>
   </mj-section>
   ${button}
--- a/packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { automatedEmailNoAction, button } from '../../partials';
+
+export const verificationReminder = ({
+  title,
+  description,
+  subDescription,
+}: Record<any, string>) => `
+   <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
+   <mj-section>
+     <mj-column>
+     <mj-text css-class="text-header"><span data-l10n-id="<%- templateName %>-title">${title}</span></mj-text>
+     </mj-column>
+   </mj-section>
+   <mj-section>
+     <mj-column>
+     <mj-text css-class="text-sub-body"><span data-l10n-id="<%- templateName %>-description">${description}</span></mj-text>
+     <mj-text css-class="text-body"><span data-l10n-id="<%- templateName %>-sub-description">${subDescription}</span></mj-text>
+     </mj-column>
+   </mj-section>
+   ${button}
+   ${automatedEmailNoAction}
+`;
--- a/packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
@@ -37,9 +37,9 @@ const commonPropsWithOverrides = (
     },
   });
 
-export const CadReminderMobile = Template.bind({});
-CadReminderMobile.args = commonPropsWithOverrides();
-CadReminderMobile.storyName = 'User is on mobile device';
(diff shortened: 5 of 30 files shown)


packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Yogita98, line 1: The base template for both `verificationReminderFirst` and `verificationReminderSecond` emails remains the same so I created a separate partial `verificationReminder` for it and just passing down the templatevalues to it while rendering the emails

packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const verificationReminder = ({
+  title,
+  description,
+  subDescription,
+}: Record<any, string>) => `
+   <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
+   <mj-section>
+     <mj-column>
+     <mj-text css-class="header-text"><span data-l10n-id="<%- templateName %>-title">${title}</span></mj-text>
```
LZoog, line 15: Heh, I was going to suggest reducing the margin-bottom on this to better match the original template, and then I realized it's using an already established class. Yay for consistency!

I think we could perhaps reduce the padding on the first paragraph though. I think it makes sense to have a larger spacing between the header and text as you have it, as well as space between the last paragraph and button as you have it. 

The current/old template is showing `11px` margin-bottom here on the "confirmed it" paragraph:
image

Then in the new template, it's `20px`, coming from `primary-text`:
image

I think this should be `12px`. I don't think we should just slap `primary-text mb-3` on there though since it overrides the `mb` from `text-primary` and we'll probably need to use that class again.

 I'm kind of torn on what we should call it, maybe because I think `primary-text` isn't necessarily the best class name (I know we were just pulling from the old templates though). 

We _could_ create a class called `body-text` for now which can go on the first paragraph...

```
.body-text div {
  @extend .text-sm;
  @extend .mb-3;
  @extend .font-sans;
  text-align: center !important;
}
```
... though, then the class name "primary-text" is almost weirder. What makes it "primary"? Maybe later in a follow up we can rename these classes.

I'm OK with that for now, but there may be a better solution. I don't think we can do something like this:

```
.primary-text:last-of-type div {
  @extend .mb-5;
}

.primary-text div {
  @extend .text-sm;
  @extend .mb-3;
  @extend .font-sans;
  text-align: center !important;
}
```
...because of the weird DOM output MJML does. I originally thought we could just add `@extend .mt-5;` to the `primary-button` class and then set `text-primary` to `mb-3` and let collapsing margins work in our favor, but I don't think that works either because of the DOM output.

LZoog (commented):
Looking good, all my comments are minor!

Is there any way we can also slap that `sans` class on here or something? Times New Roman hurts the eyes. 😆

image

packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
Yogita98: Yeah I was actually working on refactoring the styles a bit in the `verify` template PR (still WIP) so I'll just move those changes in here.

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -84,16 +89,21 @@ tr > td:first-child {
   font-family: $font-sans;
 }
 
-.header-text div {
+.text-large div {
   @extend .text-xl;
   @extend .font-sans;
   text-align: center !important;
-  @extend .mb-5;
```
Yogita98, line 91: We're actually not extending any of these margin classes anywhere except for `mb-5`. I think by keeping it as a separate class, we don't have to think about overriding styles anymore and can just add a separate class wherever we'd like

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.ts
```diff
@@ -40,9 +40,9 @@ export const render = (body: string) => {
       ${body}
       <mj-section>
         <mj-column>
-          <mj-text css-class="secondary-text">Mozilla. 2 Harrison St, #175, San Francisco, CA 94105</mj-text>
-          <mj-text css-class="secondary-text"><a class="link-blue" data-l10n-id="fxa-privacy-url" href="<%- privacyUrl %>">Mozilla Privacy Policy</a></mj-text>
-          <mj-text css-class="secondary-text"><a class="link-blue" data-l10n-id="fxa-service-url" href="https://www.mozilla.org/about/legal/terms/services/">Firefox Cloud Terms of Service</a></mj-text>
```
Yogita98, line 45: Even I agree these classnames do look a little weird so it makes sense to assign them better names. Assuming there'd be more sections in the bodytext in future template conversions, I gave them names according to what underlying font size is being used and since it is used for `<mj-text>` elements, it made sense to start the classname with `text`

packages/fxa-auth-server/lib/senders/emails/partials/verificationReminder/index.ts
```diff
@@ -0,0 +1,26 @@
...
+}: Record<any, string>) => `
+   <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
+   <mj-section>
+     <mj-column>
+     <mj-text css-class="text-xlarge mb-5"><span data-l10n-id="<%- templateName %>-title">${title}</span></mj-text>
+     </mj-column>
+   </mj-section>
+   <mj-section>
+     <mj-column>
+     <mj-text css-class="text-small mb-3"><span data-l10n-id="<%- templateName %>-description">${description}</span></mj-text>
```
Yogita98, line 20: This section now has the bottom margin of `12px` just like the old template

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -61,9 +61,13 @@ $s-10: 40px;
...
   }
-
   &-5 {
-    margin-bottom: $s-5 !important;
+    div {
+      margin-bottom: $s-5 !important;
+    }
+    img {
+      margin-bottom: $s-5 !important;
+    }
```
LZoog, line 70: We could do this instead here:
```suggestion
    div, img {
      margin-bottom: $s-5 !important;
    }
```
... but I think we should just leave the extended `mb-5` on that sync logo instead. 🤔

packages/fxa-auth-server/lib/senders/emails/global.scss
LZoog: Hmm it might be a tad confusing to call this `text-xlarge` when we've got the `text-xl` class too.

I think we may want to avoid manually adding `mb` classes along with another class because it makes it easier to use the wrong spacing class which leads to inconsistency.  I don't mind a `text-header` class (that might be better than `header-text`) with your `.text-xlarge` styles + `mb-5` extended and then, use that class on the header at the top of every email, but I think that's basically what I recommended with the placeholder suggestion yesterday, heh.

BTW looks like this will happen anyway, but I'd say wait until your other PRs are merged so you can update these classes in those templates as well.

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -84,16 +88,21 @@ tr > td:first-child {
...
 
-.primary-text div {
+.text-small div {
   @extend .text-sm;
-  @extend .mb-5;
   @extend .font-sans;
   text-align: center !important;
 }
+
+.text-xsmall div {
```
LZoog, line 103: We can probably call this one `text-footer`. Since the color is grey, I don't think we can reuse this elsewhere.

packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts
```diff
@@ -8,13 +8,13 @@ export const cadReminder = `
   <mj-include path="./lib/senders/emails/css/cadReminder/index.css" type="css" css-inline="inline" />
   <mj-section>
     <mj-column>
-      <mj-text css-class="header-text"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
-      <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
+      <mj-text css-class="text-xlarge mb-5"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
```
LZoog, line 11: Hmm yeah see my concern here is someone might just use `text-xlarge` and then like, add `mb-3` or `mb-4` here or on `sync-logo` or on anything else here and then lose that consistency we'd gain from it being in a single class.

packages/fxa-auth-server/lib/senders/emails/global.scss
LZoog: Maybe we can have the classes `text-header`, `text-body`, `text-body-mb-5`, and `text-footer`. I like all of these except the third one but am not sure what else to call it right now. 😆

Yogita98: Hmmm, right! But when I was working on converting `verify` email template, there was one more text section apart from header and body text and had styles essentially similar to the (new) `text-body` class but the font-color is darkish grey. See below

image

What do you think we should do in such situations? Not related to this PR ofc but we can potentially avoid refactoring the styles again in future PRs

LZoog: Huh, we could call it `text-body-grey` perhaps? Does it match the exact styles as `text-footer`?

Yogita98: I think the styles are similar to body text but yeah `text-body-grey` sounds good to me.

packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts
```diff
@@ -5,15 +5,16 @@
 import { appBadges, button, automatedEmailNoAction } from '../../partials';
 
 export const cadReminder = `
+  <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
   <mj-section>
     <mj-column>
-      <mj-text css-class="header-text"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
-      <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
+      <mj-text css-class="text-header"><span data-l10n-id="<%- templateName %>-title"><%- headerText %></span></mj-text>
+      <mj-image css-class="graphic-devices" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
```
Yogita98, line 12: Just realized this image is a little different than what we have in `fxa` layout so changed it's classname too

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.ts
```diff
@@ -8,14 +8,14 @@ export const render = () => `
   <mj-include path="./lib/senders/emails/css/global.css" type="css" css-inline="inline" />
    <mj-section>
      <mj-column>
-       <mj-text css-class="primary-text"><span data-l10n-id="postVerify-sub-title">Firefox Account verified. You're almost there.</span></mj-text>
-       <mj-text css-class="header-text"><span data-l10n-id="postVerify-title">Next sync between your devices!</span></mj-text>
+       <mj-text css-class="text-sub-body"><span data-l10n-id="postVerify-sub-title">Firefox Account verified. You're almost there.</span></mj-text>
```
Yogita98, line 11: While refactoring the styles, I observed that the new template (left side) has a little more margin at the bottom after the first text section
image
We kinda missed it in the `postVerify` branch, but the new styles have taken care of that too

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -84,22 +83,46 @@ tr > td:first-child {
...
+.text-header div {
   @extend .text-xl;
   @extend .font-sans;
+  @extend .mb-3;
   text-align: center !important;
-  @extend .mb-5;
 }
 
-.primary-text div {
+.text-body-common {
```
LZoog, line 93: If this isn't a class we actually use, we can turn it into a placeholder (`%text-body-common`). 😁

packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.stories.ts
```diff
@@ -37,9 +37,9 @@ const commonPropsWithOverrides = (
     },
   });
 
-export const CadReminderMobile = Template.bind({});
-CadReminderMobile.args = commonPropsWithOverrides();
-CadReminderMobile.storyName = 'User is on mobile device';
+export const CadReminderDefault = Template.bind({});
+CadReminderDefault.args = commonPropsWithOverrides();
+CadReminderDefault.storyName = 'default';
```
LZoog, line 42: Oh I'm sorry if I wasn't clear - I think we should use "default" if there's no other states, which is the case for both `verificationReminder` templates, but I don't think is the case for others. I think templates with two states, like this one being mobile and non-mobile, should have a description like how it was previously.

Yogita98: I think we moved that mobile state from `cadReminder` template to `postVerify`. I had left a [comment](https://github.com/mozilla/fxa/pull/10120#discussion_r689717790) as well in the PR

Yogita98: We actually removed the desktop state from `cadReminders` to `postVerify` template. I had also left a [comment](https://github.com/mozilla/fxa/pull/10120#discussion_r689717790) in that PR

LZoog: Gotcha! I didn't realize we weren't sending `onDesktopOrTabletDevice` down on CAD so there's only one state.

LZoog (approved):
:shipit:

Issue #10072, opened 2021-08-10
Email Input on Passwordless Subscription Check Deletes a Character of the TLD


data-sync-user:
STR:

1. Go to the /checkout page for a plan
1. When entering the email address, type the tld part of the email address very quickly

Expected:
The entire email address is entered into the input

Actual:
One of the characters of the TLD is deleted

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3832)

PR #10137, opened 2021-08-17
fix(fxa-payments-server): debounce account check

meandavejustice:
## This pull request

- debounces the "check existing account" function in input validation, so we don't have weird behavior for the user on quick input keystrokes on slow connections.

Closes: #10072

## Checklist


packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -163,23 +163,6 @@ describe('NewUserEmailForm test', () => {
     });
   });
 
-  it('shows error when an account already exists', async () => {
```
meandavejustice, line 166: Was having some race condition issues in this test because of the new debounce function. The below test actually does a better job of testing this functionality, so I removed this test.

chenba: @meandavejustice can you link to the test you are referring to?  I don't see a new test in the PR.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
```diff
@@ -170,9 +177,12 @@ export async function emailInputValidationAndAccountCheck(
...
     setValidEmail(value);
-    const response = await checkAccountExists(value);
-    hasAccount = response.exists;
-    setAccountExists(response.exists);
+    const response = await debouncedCheckAccountExists(value);
+    if (response?.exists) {
+      hasAccount = response.exists;
+      setAccountExists(hasAccount);
+    }
+    setAccountExists(hasAccount);
```
chenba, line 185: I don't understand what's going on here.  `setAccountExists` is called both in the `if` and out of the `if`.  In that case, could simply call it outside of the `if`?    But why do we need the if in the first place?

chenba (commented):
@meandavejustice Thanks for the patch.  Got a couple questions.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
meandavejustice: https://github.com/mozilla/fxa/pull/10137/files/11896dfbe5d68910d61605d1e84f5d3516179605#diff-e46a5e7e3b8430f12dcebce9d576a447060f43f87b6cdce9fdbc4007a26dbe0aR166

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: The `if` statement is here to avoid the error:
```
Uncaught (in promise) TypeError: can't access property "exists", response is undefined
```
I just pushed an update to remove the unnecessary call to `setAccountExists` within the `if` statement.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
chenba: That link didn't work.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
chenba: How do I get this error?

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
meandavejustice: https://github.com/mozilla/fxa/pull/10137/files#diff-e46a5e7e3b8430f12dcebce9d576a447060f43f87b6cdce9fdbc4007a26dbe0aR166

It is the test directly under this one in the file

packages/fxa-payments-server/src/components/NewUserEmailForm/index.tsx
meandavejustice: Start typing into the first input, when it gets to the time to input the tld the error will be in the console

meandavejustice: The error does not happen with my uodate

chenba: > 
> 
> The error does not happen with my uodate

:excellent:  Then let's remove that `if`.

meandavejustice: The `if` is necessary. That is what is stopping the error from happening

chenba: > 
> 
> > The error does not happen with my uodate
> 
> :excellent: Then let's remove that `if`.

I misunderstood what the update in the original message meant.

chenba (approved):

PR #10140, opened 2021-08-18
fix: allow Play config with private key

bbangert:
Because:

* We want to use consistent Google service account configuration.

This commit:

* Allows use of private key files for Google Play API access.

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -842,23 +842,37 @@ const conf = convict({
       },
     },
     playApiServiceAccount: {
+      credentials: {
+        client_email: {
+          default: 'test@localtest.com',
+          doc: 'The email of the service account to use for Play API calls',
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/config/index.ts
```diff
@@ -842,24 +842,38 @@ const conf = convict({
...
+        },
+        private_key: {
+          default: '',
+          doc: 'The private key of the service account to use for Play API calls',
+          env: 'SUBSCRIPTIONS_PLAY_PRIVATE_KEY_CREDENTIAL',
+          format: String,
+        },
+      },
+      keyFilename: {
+        default: path.resolve(__dirname, 'secret-key.json'),
```
dannycoates, line 860: is this the default we want? this is also the key we use to sign fxa assertions with

dannycoates (approved):

packages/fxa-auth-server/config/index.ts
bbangert: I dn't think it is, no. Replacing with empty string.

PR #10142, opened 2021-08-18
fix(auth): use working web-push version

### LINKED ISSUE #10141: Sent-tab no longer pushes tabs
A dependency update broke our push implementation as we use VAPID keys with an `=` sign in them, which is invalid and not allowed by the latest version of web-push (3.4.5). We must roll-back as the client is the one involved in VAPID key generation.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3855)

bbangert:
Because:

* web-push 3.4.5 doesn't allow = in our VAPID keys which we currently
  allow.

This commit:

* Reverses the web-push commit to a working version.

Closes #10141

## Checklist

### DIFF
--- a/packages/fxa-auth-server/package.json
+++ b/packages/fxa-auth-server/package.json
@@ -116,7 +116,7 @@
     "typedi": "^0.8.0",
     "uuid": "^8.3.2",
     "verror": "^1.10.0",
-    "web-push": "3.4.5"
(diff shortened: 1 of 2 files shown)


dannycoates:
i added this to train-212

PR #10165, opened 2021-08-19
fix(auth): rename subject of subscriptionAccountFinishSetup email

### LINKED ISSUE #10152: Change the subject of the Change your password email
Currently the '%(productName)s payment confirmed' is the subject of the email a user will receive that has the CTA to set their password. A user might be confused when they go to check their email based off the subject. I would like us to update the subject of the email to better reflect the content and flow for a user.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3858)

biancadanforth:
Because:

* The subscriptionAccountFinishSetup and subscriptionFirstInvoice emails previously had the same subject. This can cause confusion, since we send both after a subscription is successfully created.

This commit:

* Renames the subject of the subscriptionAccountFinishSetup email per the MVP spec to 'Welcome to %(productName)s: Please set your password.'

Closes #10152

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/senders/email.js
+++ b/packages/fxa-auth-server/lib/senders/email.js
@@ -1960,7 +1960,9 @@ module.exports = function (log, config) {
     const action = gettext('Create a password');
     const translatorParams = { productName };
     const subject = translator.format(
-      translator.gettext('%(productName)s payment confirmed'),
+      translator.gettext(
+        'Welcome to %(productName)s: Please set your password.'
+      ),
       translatorParams
     );
 
--- a/packages/fxa-auth-server/test/local/senders/email.js
+++ b/packages/fxa-auth-server/test/local/senders/email.js
@@ -262,7 +262,7 @@ const TESTS = [
     ]]
   ])],
   ['subscriptionAccountFinishSetupEmail', new Map([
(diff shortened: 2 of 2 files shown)


packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -1960,7 +1960,9 @@ module.exports = function (log, config) {
     const action = gettext('Create a password');
     const translatorParams = { productName };
     const subject = translator.format(
-      translator.gettext('%(productName)s payment confirmed'),
+      translator.gettext(
+        'Welcome to %(productName)s: Please set your password.'
```
biancadanforth, line 1964: Since we don't use Fluent for emails (yet), how do I ensure that this updated string will get localized? I see where we pull in localization files using Jed, but I don't see how that gets pushed to `fxa-content-server-l10n`.

biancadanforth: Ah I found part of it; the fxa-content-server-l10n repo pulls from various fxa packages, including the auth-server: https://github.com/mozilla/fxa-content-server-l10n/blob/master/scripts/extract_strings.sh and https://github.com/mozilla/fxa-content-server-l10n/blob/master/scripts/extract-and-pull-request.sh .

chenba (approved):
📧

PR #10173, opened 2021-08-19
Do not merge - feat(auth-server): Convert postVerify template to new stack

jodyheavener:
Clone of https://github.com/mozilla/fxa/pull/10120 but not on a fork to get CI passing


jodyheavener:
I've re-ran this a half dozen times and the same tests immediately fail. @chenba do you think this could be experiencing the same issue as with the forked branch?

Yogita98:
@jodyheavener Actually I saw the tests failing so thought pulling in new changes and rebasing the branch might work although the tests passed in the previous PR without any update, but if you think that might work then I have just force-pushed and updated the branch.

Issue #10171, opened 2021-08-19
[subplat] FxA ToS and PN available in “Subscription confirmation” page for old users


data-sync-user:
**Affected browser versions:**

* All Chrome;

**Affected FxA versions:**

* 1.213.1;

**Prerequisites:**

* Have a VPN set to Germany location;
* Be in the old "Subscription confirmation" page;

**Steps to reproduce:**
1. Observe footer;

**Expected result:**

* There are no FxA Tos and PN links available.

**Actual result:**

* The FxA Tos and PN links are displayed.

**Notes:**

* Both links redirect to a “Not Found” page ~~https://payments~~stage.fxa.nonprod.cloudops.mozgcp.net/products/undefined/legal/privacy;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3868)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16035/ToS+and+PN+not+found+for+FxA.gif">ToS and PN not found for FxA.gif</a>

biancadanforth:
Thanks for filing.

Developer note: This is because the `showFXALinks` flag in the `TermsAndPrivacy` component should be conditionally set based on `accountExists` in the `PaymentConfirmation` component.

data-sync-user:
➤ Bianca Oltean commented:

Bianca Danforth The FxA ToS and PN links in the new “Subscription confirmation“ page also redirect to a “Not found“ page.

!ToS and PN not found for FxA.gif|width=1905,height=1012!

data-sync-user:
➤ Bianca Danforth commented:

Thank you fellow Bianca (first time for everything).

Dave Justice is working on this fix. Dave, can you take a look? He is on PTO today and Monday, but he will be back Tuesday.

If I get some bandwidth before he’s back, I can look at it as well.

data-sync-user:
➤ Dave Justice commented:

fixed in https://github.com/mozilla/fxa/pull/10179 ( https://github.com/mozilla/fxa/pull/10179|smart-link )

data-sync-user:
➤ Bianca Danforth commented:

This should be verifiable in Train 213.4.

data-sync-user:
➤ Madalin Cotetiu commented:

Verified as fix on Train 1.213.5, on Chrome. There are no FxA Tos and PN links available for old users.

PR #10179, opened 2021-08-19
fix(fxa-payments-server): do not show fxa terms to existing users

meandavejustice:
## This pull request

- passes `accountExists` to TermsAndPrivacy showFXALinks prop, to that we don't show to existing users.

Closes: #10171

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
@@ -33,7 +33,7 @@ export const PaymentConfirmation = ({
   className = 'default',
   accountExists = true,
 }: PaymentConfirmationProps) => {
-  const { navigatorLanguages } = useContext(AppContext);
+  const { config, navigatorLanguages } = useContext(AppContext);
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -160,7 +160,7 @@ export const PaymentConfirmation = ({
 
         <div className="footer" data-testid="footer">
           <PaymentLegalBlurb provider={payment_provider} />
-          <TermsAndPrivacy plan={selectedPlan} showFXALinks={true} />
+          <TermsAndPrivacy plan={selectedPlan} />
```
biancadanforth, line 163: We want to pass `showFXALinks={!accountExists}` and `contentServerURL` just like for `Checkout`. The FxA links will never show on `PaymentConfirmation` otherwise.

biancadanforth (changes requested):
Thanks Dave. `PaymentConfirmation` still doesn't show the FxA links for the new flow. I added a comment inline.

chenba (approved):

biancadanforth (approved):

PR #10186, opened 2021-08-19
fix(auth): return a session token in /account/finish_setup

### LINKED ISSUE #10184: Sign Stub Account into FxA After the User Set Their Password
When a user with a stub account finish setting up their account by entering a password, they are not signed into FxA afterward.  Let's sign them in when they finish the account setup.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3875)

dannycoates:
closes #10184

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -569,7 +569,7 @@ export class AccountHandler {
       // Only proceed if the account is in the stub state.
       // This prevents a token from being used multiple times.
       if (account.verifierSetAt !== 0) {
-        return {};
+        throw error.unauthorized('token already used');
       }
       const { authSalt, wrapWrapKb } = account;
       const password = new this.Password(
@@ -590,12 +590,24 @@ export class AccountHandler {
         }
       );
       await this.db.resetAccountTokens(uid);
+      const sessionToken = await this.createSessionToken({
+        account,
+        request,
+        // this route counts as verification
+        tokenVerificationCode: undefined,
+        tokenVerificationId: undefined,
+      });
+      return {
+        uid,
+        sessionToken: sessionToken.data,
+        verified: sessionToken.emailVerified,
+      };
     } catch (err) {
       this.log.error('Account.finish_setup.error', {
         err,
       });
+      throw err;
     }
-    return {};
   }
 
   async login(request: AuthRequest) {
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -590,12 +590,24 @@ export class AccountHandler {
...
+        account,
+        request,
+        // this route counts as verification
+        tokenVerificationCode: undefined,
+        tokenVerificationId: undefined,
+      });
+      return {
+        uid,
+        sessionToken: sessionToken.data,
+        verified: sessionToken.emailVerified,
```
vbudhram, line 603: Would the email on the sessionToken ever be unverified? Could we return true here?

vbudhram (commented):
@dannycoates This looks good codewise, just gonna give it a test locally.

vbudhram (approved):
@dannycoates In my local testing this takes you the signin page and you just have to click sign in again (without password). Is this the correct behavior?

biancadanforth:
> @dannycoates In my local testing this takes you the signin page and you just have to click sign in again (without password). Is this the correct behavior?

No this is not the correct behavior IIUC. We are trying to avoid any indirection from the finish setup page to the download page.

dannycoates:
>  In my local testing this takes you the signin page and you just have to click sign in again (without password). Is this the correct behavior?

can you be more specific about what you're doing?

packages/fxa-auth-server/lib/routes/account.ts
dannycoates: it should always be true, but i think it's more helpful to reference the source of the value.

vbudhram:
> > In my local testing this takes you the signin page and you just have to click sign in again (without password). Is this the correct behavior?
> 
> can you be more specific about what you're doing?

1. Change https://github.com/mozilla/fxa/blob/87bb1808bcddf62d14f6a4ee10d9bc61f0de79f8/packages/fxa-content-server/app/scripts/views/post_verify/finish_account_setup/set_password.js#L74 to `http://localhost:3030` or `http://localhost:8080`
2. Goto http://localhost:8080 click `Subscribe to Pro (passwordless)`
3. Complete subscription
4. Get email and open link
5. Set password

Expected:

Redirected to settings page

Actual:

Redirect to signin page with signin button only

dannycoates:
We expect the url we redirect to to do oauth with `prompt=none`

Issue #10166, opened 2021-08-19
[subplat] Wrong payment information details in Create Password email


data-sync-user:
**Affected browser versions:**

* All Chrome;

**Affected FxA versions:**

* 1.213.1;

**Prerequisites:**

* Have a VPN set to Germany location;

**Steps to reproduce:**
1. Open VPN homepage [https://www-dev.allizom.org/de/products/vpn/;](https://www-dev.allizom.org/de/products/vpn/;)
2. Click “Get Mozilla VPN” and choose any plan;
3. Fill emails fields and select Paypal payment method;
4. Finalize subscription;
5. Go to your email and check the “Create password” email;

**Expected result:**

* The email contains correct payment details.

**Actual result:**

* The email contains wrong date in payment details section;

**Notes:**

* reproducible from both payment methods.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3863)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16015/image_2021_08_19T11_21_39_270Z.png">image_2021_08_19T11_21_39_270Z.png</a>

data-sync-user:
➤ Madalin Cotetiu commented:

Verified this as fixed on v. 1.213.3 - Staging. The email contains the correct payment details. See the attached screenshot. 

!image-20210823-134203.png|width=716,height=691!

PR #10194, opened 2021-08-20
fix(emails): handle Stripe timestamps that are in seconds

chenba:
Because:
 - Stripe uses seconds instead of milliseconds for timestamps

This commit:
 - update the finish stub account setup email to use a function that
   handle those timestamps

Closes: #10166


vbudhram (approved):
@chenba 👍🏽

PR #10196, opened 2021-08-20
fix(auth): use token handler for token endpoint

### LINKED ISSUE #10195: Token endpoint returns plans
The token endpoint was linked to the wrong handler function, it should be linked to the token handler.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3879)

bbangert:
Because:

* The token handler should be used for the token endpoint.

This commit:

* Uses the correct handler for the token endpoint.

Closes #10195

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/google.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/google.ts
@@ -109,7 +109,8 @@ export const googleIapRoutes = (): ServerRoute[] => {
             .required(),
         },
       },
-      handler: (request: AuthRequest) => googleIapHandler.plans(request),
+      handler: (request: AuthRequest) =>
+        googleIapHandler.registerToken(request),
     },
   ];
 };


chenba (approved):
👍   Is there a test that verifies which function in the GoogleIapHandler is called when the request handler is called?  Feel like we should have something to catch this.

packages/fxa-auth-server/lib/routes/subscriptions/google.ts
```diff
@@ -109,7 +109,8 @@ export const googleIapRoutes = (): ServerRoute[] => {
             .required(),
         },
       },
-      handler: (request: AuthRequest) => googleIapHandler.plans(request),
```
dannycoates, line 112: 😬 how did i miss that?

bbangert:
@chenba there isn't, nor for a bunch of other payment ones as well, runTest could probably do it. Though its a lot of test infra to rig up for a missed link like this.

PR #10199, opened 2021-08-20
chore(auth): add hawk payload validation to POSTs that require a payl…

### LINKED ISSUE #9850: Add hawk payload hash validation to POSTs that use tokens
We should be more consistent in how we apply the `payload: 'required'` property ([example](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/sign.js#L21)) on routes that accept a body.

For routes that always accept a body and use hawk authentication lets set `payload: 'required'` to enable payload hash validation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3747)

dannycoates:
closes #9850


chenba (approved):
🐦

PR #10203, opened 2021-08-21
fix(auth): use IAP scope for token register

### LINKED ISSUE #10202: Google token endpoint should use its own scope
The token endpoint incorrectly requires subscription scope access like we use for the checkout page, it should instead use a new subscription IAP permission.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3881)

bbangert:
Because:

* We should use a narrowly scoped permission for IAP token registration.

This commit:

* Adds a new scope for IAP access and applies it to the token
  registration endpoint.

Closes #10202

## Checklist

### DIFF
--- a/packages/fxa-shared/oauth/constants.js
+++ b/packages/fxa-shared/oauth/constants.js
@@ -2,12 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const OAUTH_SCOPE_SUBSCRIPTIONS =
+  'https://identity.mozilla.com/account/subscriptions';
+
 module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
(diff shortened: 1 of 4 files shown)


packages/fxa-shared/oauth/constants.js
```diff
@@ -2,12 +2,15 @@
...
+  'https://identity.mozilla.com/account/subscriptions';
+
 module.exports = {
   OAUTH_SCOPE_OLD_SYNC: 'https://identity.mozilla.com/apps/oldsync',
   OAUTH_SCOPE_SESSION_TOKEN: 'https://identity.mozilla.com/tokens/session',
   OAUTH_SCOPE_NEWSLETTERS: 'https://identity.mozilla.com/account/newsletters',
-  OAUTH_SCOPE_SUBSCRIPTIONS:
-    'https://identity.mozilla.com/account/subscriptions',
+  OAUTH_SCOPE_SUBSCRIPTIONS,
+  OAUTH_SCOPE_SUBSCRIPTIONS_IAP: `${OAUTH_SCOPE_SUBSCRIPTIONS}:iap`,
```
dannycoates, line 13: I was wrong originally about the `:` for url scopes, it should be `/` or `#`

https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/oauth/scopes.js#L518

dannycoates (approved):
pre-approving, but we need to change the scope string

Issue #10223, opened 2021-08-23
We don't pass UTM params when clicking sign-in link from passwordless view


vbudhram:
When a user clicks the sign-in link from the new passwordless view, we do not redirect with any exisiting query params.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3889)

PR #10228, opened 2021-08-23
fix(subscriptions): copy query params into sign-in link

chenba:
Because:
 - we need to keep the query params when an existing user use the
   sign-in link on the passwordless flow

This commit:
 - copy the query params from the checkout page into the sign-in link

Closes: #10223


vbudhram (approved):
@chenba :shipit:

Issue #10220, opened 2021-08-23
Use .mjml extension or export templates as strings


data-sync-user:
Right now, new email templates (MJML/EJS) are inside "`" tags to use them as a strings ([see here](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/emails/partials/cadReminder/index.ts#L7).

Find a way to get our templates out of this format (can we use an .mjml extension rather than `.ts`?) and then apply this method to the already converted templates/partials.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3888)

PR #10283, opened 2021-08-30
feat(fxa-auth-server): refactor renderer to parse MJML files

jodyheavener:
## Because

- We are currently writing email templates as JS template literals instead of actual MJML files. This works fine, but we lose syntax highlighting and generally create more clutter.

## This pull request

- Converts all template files from template literals to proper `.mjml` files
  - Also more generously hand-formatted the MJML (more line breaks/breathing room)
- Reworks the renderer to import and parse these files, cleaning up the flow a little
  - Updated some of the TS types
- Bonus: removes the `preview-in-maildev.ts` and `render-to-disk.ts` as they do not appear to be maintained or in use

Closes: #10220

## Checklist

- [x] I have added necessary documentation.
- [x] I have verified that my changes render correctly in RTL.

- Feel free to check for yourself, but I manually verified that the output for HTML and text is still the same and consistent with the current templates
- I initially was going to try and use `mj-include` to include partial template files, but since they have EJS within them we need to parse the EJS out first, flatten everything into one MJML string, and then parse the MJML


jodyheavener:
cc @Yogita98

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.mjml
```diff
@@ -0,0 +1,47 @@
+<mjml>
```
LZoog, line 1: Mentioned this in Slack, but we _could_ add a license to all of these files 😬

packages/fxa-auth-server/lib/senders/emails/render-to-disk.ts
```diff
@@ -1,21 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
LZoog, line 1: We could make it more prominant in the docs on using `yarn write-emails` to preview the old or new templates. We have this sentence:

> The emails for which the MJML feature flag is enabled can be rendered to disk using the `yarn write-emails` command.

but I think that could be a little clearer 🤔

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.mjml
```diff
@@ -0,0 +1,25 @@
...
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-sub-body">
+      <span data-l10n-id="postVerify-sub-title">Firefox Account verified. You're almost there.</span>
+    </mj-text>
+    <mj-text css-class="text-header">
+      <span data-l10n-id="postVerify-title">Next sync between your devices!</span>
+    </mj-text>
+    <mj-image css-class="sync-logo" alt="Devices" src="https://accounts-static.cdn.mozilla.net/other/graphic-laptop-mobile.png"></mj-image>
```
LZoog, line 11: Ah you know, we probably need a follow up issue for localizing alt text, I don't think we did that at all in current/previous emails (we just put `alt=""`) so we can probably do that after we've converted the remaining templates. We may have questions on what that alt text should be as we continue to build these out, so maybe the task can be an all-encompassing "make sure all images have alt text and that the alt text is localized" kind of ticket.

packages/fxa-auth-server/lib/senders/emails/templates/verificationReminderFirst/index.mjml
```diff
@@ -0,0 +1,5 @@
+<%- include('/partials/verificationReminder/index.mjml', {
+  title: 'Welcome to the Firefox family',
```
LZoog, line 2: I was going to insert an A7X "Welcome to the Family" gif but I can't find one so I'll settle for emojis:

🦇💀🦇
🤗 👨‍👩‍👧‍👦

LZoog (approved):
Niiiice, what a great improvement😍

Tested it out locally and don't see any issues :shipit:

packages/fxa-auth-server/lib/senders/emails/templates/postVerify/index.mjml
jodyheavener: 👍 https://mozilla-hub.atlassian.net/browse/FXA-3924

PR #10285, opened 2021-08-30
fix(auth): assume product change on play trigger

### LINKED ISSUE #10263: Google Play product comparison is not detecting changes to entitlements correctly
When RTDN messages come in that result in a subscription being paused, or cancelled, we are not broadcasting the change.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3913)

bbangert:
Because:

* We want to ensure that when a Play purchase change occurs, we send
  out a notification if the product capbilities could have changed.

This commit:

* Assumes that any change to a product purchased via Google Play is a
  binary change going from an active to inactive or vice versa change.
  This may end up broadcasting capabilities as going active/inactive
  multiple times even when no change has happened, but ensures relying
  parties have granted the entitlemenet appropriately.

Closes #10263

## Checklist


dannycoates (approved):
I don't see any obvious problems. confidence level: "let's see what happens"

chenba (approved):

PR #10286, opened 2021-08-30
chore(content): set SUBSCRIPTIONS_UNAUTHED_REDIRECTS to true by default

### LINKED ISSUE #10234: [subplat][content-server] Set the SUBSCRIPTIONS_UNAUTHED_REDIRECTS config's default value to true
MVP's done.  The checkout page is working on Payments.  It's safe to have true as the default value to allow redirects to the passwordless checkout flow.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3896)

biancadanforth:
Because:

* We have deployed the [passwordless flow feature](https://mozilla-hub.atlassian.net/browse/FXA-3486) to production, and it has been QA-verified.
* Therefore, we can safely enable the feature via this flag in all environments, which we have done everywhere except in the content server config file.

This commit:

* Sets SUBSCRIPTIONS_UNAUTHED_REDIRECTS to true by default to make this behavior the default in the content server config file; this saves the step of setting the flag and restarting the content server for local development.

Closes #10234

## Checklist

### DIFF
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -671,7 +671,7 @@ const conf = (module.exports = convict({
       format: String,
     },
     allowUnauthenticatedRedirects: {
-      default: false,
+      default: true,
       doc: 'Whether to allow any redirects to Payments for an unauthenticated user',
       env: 'SUBSCRIPTIONS_UNAUTHED_REDIRECTS',
       formlat: Boolean,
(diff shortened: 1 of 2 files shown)


biancadanforth:
Looks like I need to add a `signin` queryParam. I will fix this and resubmit shortly.

vbudhram (commented):
Related to `signup via product page and redirect after confirm` error in CI?

biancadanforth:
Yes; but that `PRODUCT_URL` in the test has the `signin` queryParam, so I don't know why it's loading the new checkout screen instead of the auth prompt.

biancadanforth:
Interestingly, `forceSignin` was incorrectly `false`, despite `signin` being appended to `PRODUCT_URL`. Apparently just the presence of the query param with no specified value means `!!this._queryParams.signin` returns `false`.

chenba:
> 
> 
> Interestingly, `forceSignin` was incorrectly `false`, despite `signin` being appended to `PRODUCT_URL`. Apparently just the presence of the query param with no specified value means `!!this._queryParams.signin` returns `false`.

Interesting indeed!  I guess at some point we switched from using the presence of a key to the value of that key.  (Or this is a different spot.)

chenba (approved):
💯

Issue #9957, opened 2021-07-27
[subplat] Layout issue for PayPal logo when upgrading the subscription 


data-sync-user:
**Affected versions:**

* All Safari;
* All Firefox;
* All Chrome; 

**Affected FxA versions:**

* 1.211.1;

**Prerequisites:**

* Have the VPN set to US;
* Have an FxA account without a subscription;

**Steps to reproduce:**
1. Subscribe via PayPal to the monthly subscription plan with the account from the prerequisites: 
https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=plan_FiJ55YK6UYftPa;
2. With the same account, upgrade your monthly subscription plan to the 12-months plan: https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=price_1J0Y1iKb9q6OnNsLXwdOFgDr; 
3. Observe the content of the “Payment information” section from the Subscription upgrade page.

**Expected result:**

* PayPal logo is displayed on the “Payment information” section;

**Actual result:**

* Credit card info + strange chars are displayed in the “Payment information” section.

**Notes:**

* See the attached screenshot.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3800)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/15774/Screenshot+2021-07-23+at+11.27.20.png">Screenshot 2021-07-23 at 11.27.20.png</a>

data-sync-user:
➤ Luminita Buruian commented:

Verified as fixed on Train 1.215.1 - Staging. 
The PayPal logo is now displayed on the “Payment information” section. See the attached screenshot.

!Screenshot 2021-09-17 at 10.40.26.png|width=1055,height=589!

Thank you.

PR #10289, opened 2021-08-30
fix(subscriptions): show correct payment provider info for upgrades

chenba:
Because:
 - the upgrade checkout page does not display the correct payment
   provider information when the customer uses PayPal

This commit:
 - use an existing component to show a customer's current payment
   provider info

Closes: #9957


biancadanforth (approved):
LGTM

biancadanforth:
Weirdly it's showing that one of the CI tasks is still running (test-content-server-3), yet when I click that task, it completed successfully in CircleCI.

PR #10322, opened 2021-09-01
feat(auth-server): convert verifyShortCode and verifySecondaryCode templates to new stack

jodyheavener:
## Because

- We need to convert the templates for verifyShortCode and verifySecondaryCode to our new MJML stack

## This pull request

- Converts them to the new stack
- Updates fluent localizer to strip excessive line breaks in plaintext emails

Closes: #9302
Closes: #9297

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -437,7 +437,8 @@
       "verificationReminderFirst",
       "verificationReminderSecond",
       "verify",
-      "verifyShortCode"
+      "verifyShortCode",
+      "verifySecondaryCode"
     ]
   }
 }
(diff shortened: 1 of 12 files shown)


LZoog (approved):
LGTM on both templates 👍

I was thinking since the "This is an automated email" message now expands to 3 lines due to the font-size increase we may consider extending the width of the email body, but to fit those on 2 lines (for English at least) we'd need to increase it from `310px` to around `340px`. We could do this since it's a `max-width` and seems to be responsive if someone was on like an iPhone 4 (320px width), but maybe we wouldn't like how that makes other text break. This is `verifySecondaryCode` with 340px, but now it breaks on "Firefox Account". So... meh. Maybe we can run the width increase by a designer at some point.

image

jodyheavener:
@LZoog I'm down to do this and think it would be worth while, though I would be hesitant to do it in this PR since it impacts all other templates. Maybe we can do a quick design review (like very cursory, no major changes) across all templates as part of a wrap-up task?

PR #10325, opened 2021-09-01
feat(auth-server): Add env vars for MJML addresses/templates

LZoog:
Because:
* We need to set these to values in staging

This commit:
* Creates two env vars: one for mjml enabled email addresses, and one for mjml templates


dannycoates (approved):
Thanks @LZoog :)

PR #10329, opened 2021-09-01
feat(auth-server): convert newDeviceLogin template to new stack

jodyheavener:
## Because

- We need to convert the template for newDeviceLogin to our new MJML stack

## This pull request

- Converts it to the new stack
- Moved `text-footer-automatedEmail` to global styles for re-use

Closes: #9264

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/dev.json
+++ b/packages/fxa-auth-server/config/dev.json
@@ -439,7 +439,8 @@
       "verify",
       "verifyShortCode",
       "verifySecondaryCode",
-      "passwordReset"
+      "passwordReset",
+      "newDeviceLogin"
     ]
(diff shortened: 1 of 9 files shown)


LZoog (approved):
LGTM!

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -121,6 +121,14 @@ tr > td:first-child {
   color: $grey-400 !important;
 }
 
+.text-footer-automatedEmail {
```
LZoog, line 124: Oh, this could be `&-automatedEmail` in `.text-footer`, but meh

Issue #9906, opened 2021-07-20
WError: Email account sent complaint


biancadanforth:
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/11127011/?referrer=github_plugin

```
WError: Email account sent complaint
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/error.js", line 760, in Function.AppError.emailComplaint
    return new AppError(
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/senders/email_service.js", line 71, in marshallBounceError
    return error.emailComplaint(bouncedAt);
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/senders/email_service.js", line 62, in marshallError
    return marshallBounceError(body.errno, body.bouncedAt || body.time);
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/senders/email_service.js", line 49, in Request._callback
    err = marshallError(res.statusCode, body);
...
(11 additional frame(s) were not displayed)
```

This is related to https://github.com/mozilla/fxa/issues/9367 in terms of policy on how to handle these, but we should make sure we [respond to a Stripe webhook event with a 200 and report the error instead of throwing](https://github.com/mozilla/fxa/blob/8e93ccba3f5a34e92e021e1caf52eda981ae9ceb/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts#L121-L126).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3777)

PR #10351, opened 2021-09-03
fix(subscriptions): ignore email complaint errors on webhooks

chenba:
Because:
 - we get errors in Sentry and Stripe when there's an email complaint
   error while handling a webhook event

This commit:
 - ignore the email complaint error so that Stripe stops retrying the
   webhook event

Closes: #9906

### DIFF
--- a/packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
+++ b/packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
@@ -346,6 +346,11 @@ describe('StripeWebhookHandler', () => {
         );
 
         it(
+          'ignores emailComplaint',
+          commonIgnorableErrorTest(error.emailComplaint(42))
+        );
+
+        it(
           'ignores missingSubscriptionForSourceError',
           commonIgnorableErrorTest(
             error.missingSubscriptionForSourceError(
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -26,6 +26,7 @@ import { StripeHandler } from './stripe';
 const IGNORABLE_STRIPE_WEBHOOK_ERRNOS = [
   error.ERRNO.UNKNOWN_SUBSCRIPTION_FOR_SOURCE,
   error.ERRNO.BOUNCE_HARD,
+  error.ERRNO.BOUNCE_COMPLAINT,
 ];
 
 export class StripeWebhookHandler extends StripeHandler {


packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -345,6 +345,11 @@ describe('StripeWebhookHandler', () => {
           commonIgnorableErrorTest(error.emailBouncedHard(42))
         );
 
+        it(
+          'ignores emailComplaint',
+          commonIgnorableErrorTest(error.emailComplaint(42))
```
biancadanforth, line 350: Not a blocker for this patch, but I think there may be an issue with our error filtering here.

I noticed some of our webhook handlers have their own `try..catch` block (e.g. [`handleInvoicePaidEvent`](https://github.com/mozilla/fxa/blob/7078624f185784f0cbfacfcdfe84660bd09a3121/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts#L399-L402)), and some, like the [`handleCustomerSourceExpiringEvent`](https://github.com/mozilla/fxa/blob/7078624f185784f0cbfacfcdfe84660bd09a3121/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts#L427-L436) handler used in this test, do not.

In the `handleInvoicePaidEvent` case, it does not rethrow after catching the error. Wouldn't we only pass errors through this `IGNORABLE_STRIPE_WEBHOOK_ERRNOS` filter in the cases that the handlers either catch and rethrow or have no `try...catch` at all? I'm concerned that legitimate errors that we should not ignore are not being rethrown in some cases, as in `handleInvoicePaidEvent`.

biancadanforth (approved):
Thanks Barry; LGTM but I had a non-blocking thought inline that I'd like your opinion on.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
chenba: Good catch, and generally I agree.  But without looking through the code I can't be sure what sort of error that catch in `handleInvoicePaidEvent` is catching; there's no guarantee that it's an error from lib/error.js of auth-server, where those `IGNORABLE_STRIPE_WEBHOOK_ERRNOS` originate.  Do you want to file a "maintenance" issue so we can looking into it at some point?

biancadanforth: Sure; I filed https://mozilla-hub.atlassian.net/browse/FXA-3963.

PR #10366, opened 2021-09-03
feat(auth-server): convert verifySecondary, passwordChanged, and unblockCode templates to new mjml stack

jodyheavener:
## Because

- We need to convert email templates to the new MJML stack

## This pull request

- Converts the `verifySecondary`, `passwordChanged`, and `unblockCode` templates to our new MJML stack

Closes: #9294
Closes: #9308
Closes: #9262

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/global.scss
+++ b/packages/fxa-auth-server/lib/senders/emails/global.scss
@@ -18,6 +18,7 @@ $s-3: 12px;
 $s-4: 16px;
 $s-5: 20px;
 $s-6: 24px;
+$s-8: 32px;
 $s-10: 40px;
 
 // Font-size and line-height
@@ -62,6 +63,9 @@ $s-10: 40px;
   }
   &-6 {
     margin-top: $s-6 !important;
+  }
+  &-8 {
+    margin-top: $s-8 !important;
   }
 }
 
@@ -112,10 +116,13 @@ tr > td:first-child {
   @extend .mb-5;
 }
 
-.text-body-top-padded div {
+.text-body-no-margin div {
+  @extend %text-body-common;
+}
+
+.text-body-top-margin div {
   @extend %text-body-common;
   @extend .mt-5;
-  @extend .mb-5;
 }
 
 .text-sub-body div {
@@ -134,7 +141,7 @@ tr > td:first-child {
   @extend .text-footer;
   div {
     @extend .mb-3;
-    @extend .mt-6;
+    @extend .mt-8;
   }
 }
 
--- a/packages/fxa-auth-server/lib/senders/emails/global.scss
+++ b/packages/fxa-auth-server/lib/senders/emails/global.scss
@@ -152,9 +152,20 @@ tr > td:first-child {
   @extend .mt-2;
 }
 
-.large-code div {
+%code-common {
   text-align: center !important;
   font-family: monospace;
-  @extend .text-3xl;
   @extend .mb-6;
+}
+
+.code {
+  &-medium div {
+    @extend %code-common;
+    @extend .text-lg;
+  }
+
+  &-large div {
+    @extend %code-common;
+    @extend .text-3xl;
+  }
 }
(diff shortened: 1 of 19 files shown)


packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.mjml
```diff
@@ -0,0 +1,30 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
```
LZoog, line 1: I just realized something weird with `verify`, though not directly related to your PR since this is not applicable to `verifySecondary`.

Is it supposed to show the connect another device image instead of the FF logo? 🤨 That's the way it is with the old templates, but it feels off to me.

image


Also, if it is supposed to be this, it looks sort of squished horizontally. Looks like the true dimension of this image is `395px` by `180px`, so if we want it to be `137px` high, the width should be `300px`, not `270px`.

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -143,9 +152,18 @@ tr > td:first-child {
...
 
-.large-code div {
+%code-common {
   text-align: center !important;
   font-family: monospace;
-  @extend .text-3xl;
   @extend .mb-6;
 }
+
+.large-code div {
```
LZoog, line 161: We could call this `code-large` and `code-medium` instead to follow our other conventions (and then you can nest these in a `.code { &-small ...` css set if you want)

packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.mjml
```diff
@@ -0,0 +1,30 @@
...
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-body-top-padded">
+      <span data-l10n-id="verifySecondary-post-verification">Once verified, this address will begin receiving security notifications and confirmations.</span>
+    </mj-text>
+  </mj-column>
+</mj-section>
+
+<%- include('/partials/automatedEmailNoAction/index.mjml') %>
```
LZoog, line 30: It looks like in the old template, there's much less space between the button and footer than what we've got in the new template. My original comment was "yay for consistency!" but as you click through the templates, we actually seem to be slightly inconsistent with the spacing here -  maybe we want a `margin-top` on `text-footer-email` - there's a lot of space in all of the `verify` templates, `cadReminder` templates, `passwordChanged`/`passwordReset`, and `postVerify`:

image


But then in some of our other templates, like `lowRecoveryKeys`, `newDeviceLogin`, `unblockCode`, and maybe one or two others that automated text doesn't have that spacing:

image

LZoog (commented):
Nothing major found, just a couple of comments, if we want to address those in a follow up that's cool with me, let me know and I'll come back to r+ :)

packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.mjml
LZoog: Circling back here, I see this is because `sync: true` is in the stories file. In the set of templates I'm working on, I added a story for sync as well as non-sync to show that the devices image shows instead of the FF logo, but I guess that affects all templates since that conditional is in the layout file?

jodyheavener: Makes sense!

packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.mjml
jodyheavener: @LZoog I did my best to make this a little more consistent, though I think this could use a task for a proper cleanup

packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.mjml
LZoog: For posterity: we're going to create a Story for "layouts" and show the difference there

LZoog (approved):
LGTM, though you probably want to remove the comments at the top of the fluent files for less changes in your other open PR

Issue #10177, opened 2021-08-19
Allow bearer token auth for support endpoint


data-sync-user:
The VPN client needs a way to submit support tickets for unauthenticated users. This can occur if there's an app issue, etc. The guardian website will broker this communication, the client does not use the FxA API directly.

To reduce the risk of support ticket abuse, guardian will use a basic Bearer token. Our endpoint needs to be updated to allow a bearer token for creation, and re-factored to allow new anonymous tickets. Guardian will use device heuristics/etc. to reduce abuse.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3871)

data-sync-user:
➤ Barry Chen commented:

Benjamin Bangert  What fields will Guardian send in the request?

data-sync-user:
➤ Benjamin Bangert commented:

I believe all of the ones the support form handles currently, as well as the version (which needs a new field from the other support issue).

data-sync-user:
➤ Barry Chen commented:

Benjamin Bangert with the Support Form the customer’s email address and uid implicitly come through the access token.  Will Guardian include either of those?

data-sync-user:
➤ Benjamin Bangert commented:

No, there will be no uid, just an email.

data-sync-user:
➤ Barry Chen commented:

Benjamin Bangert Does the person submitting the ticket need to be a FxA user?  Current patch in PR uses the email addr to look up the account.  But if the person doesn’t even need to be a FxA user, then I need to make some changes.

data-sync-user:
➤ Benjamin Bangert commented:

They do not need to be a FxA user. This is for VPN app users to report bugs that might prevent them from signing up, or if they have some other issue trying to subscribe or use the app.

data-sync-user:
➤ Barry Chen commented:

Sarah Bird (because I don’t know who to ping) in addition to the bearer token, the payload must also include an email address in an object property named email. 

The patch will be deployed as part of FxA Train 215.  It should be on Stage on Wed September 15 and Prod on Wed Sept 22 (prod deploy is really up to jbuck though).

jbuck can help with the value of the bearer token.

data-sync-user:
➤ Luminita Buruian commented:

Is this something the QA team can verify?

If so, could you please add some steps indicating how to check this? Benjamin Bangert  Barry Chen 

Thank you.

CC: Valentina Virlics

data-sync-user:
[...]

PR #10407, opened 2021-09-09
feat(subscriptions): accept Guardian submitted support tickets

chenba:
Because:
 - Guardian wants to submit support tickets on the behalf of users

This commit:
 - update the support ticket endpoint to
   - allow authn through a shared secret
   - look up a user by email from the payload

Closes: #10177

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -1923,6 +1923,14 @@ const conf = convict({
     env: 'SYNC_TOKENSERVER_URL',
     format: 'url',
   },
+  support: {
+    secretBearerToken: {
+      default: 'YOU MUST CHANGE ME',
+      doc: 'Shared secret to access the support endpoint.',
+      env: 'SUPPORT_AUTH_SECRET_BEARER_TOKEN',
+      format: 'String',
+    },
+  },
 });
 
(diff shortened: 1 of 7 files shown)


chenba:
@bbangert  Ok, FxA account no longer required.  This is ready for review again.

packages/fxa-auth-server/config/index.ts
```diff
@@ -1923,6 +1923,14 @@ const conf = convict({
     env: 'SYNC_TOKENSERVER_URL',
     format: 'url',
   },
+  guardian: {
```
bbangert, line 1926: I don't know if other RPs might need this as well for their app support. Maybe call it `support`?

packages/fxa-auth-server/lib/server.js
```diff
@@ -365,6 +365,14 @@ async function create(log, error, config, routes, db, translator, statsd) {
   );
   server.auth.strategy('supportPanelSecret', 'supportPanelSecret');
 
+  server.auth.scheme(
+    'guardianSecret',
```
bbangert, line 369: `supportSecret`?

bbangert (approved):
Looks good, I think it'd be useful to call it a support secret though, as not everyone is going to remember what guardian is for.

PR #10414, opened 2021-09-10
feat(auth-server): Convert verifyPrimary, verifyLogin, and verifyShortCode templates to new mjml stack

### LINKED ISSUE #9306: [emails M3] Convert `verifyLogin`
Take `verifyLogin.html` and any needed partials and convert the template(s) over to the new emails stack. Also convert tests, add states to the state previewing solution, and ensure the template is localized. Make sure it builds properly and that the new email is sent locally by turning on the feature flag it sits behind.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3536)

### LINKED ISSUE #9309: [emails M3] Convert `verifyLoginCode`
Take `verifyLoginCode.html` and any needed partials and convert the template(s) over to the new emails stack. Also convert tests, add states to the state previewing solution, and ensure the template is localized. Make sure it builds properly and that the new email is sent locally by turning on the feature flag it sits behind.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3537)

### LINKED ISSUE #9312: [emails M3] Convert `verifyPrimary`
Take `verifyPrimary.html` and any needed partials and convert the template(s) over to the new emails stack. Also convert tests, add states to the state previewing solution, and ensure the template is localized. Make sure it builds properly and that the new email is sent locally by turning on the feature flag it sits behind.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3538)

LZoog:
## Because

- We want to convert all email templates to the new MJML stack

## This pull request

 * Converts verifyPrimary, verifyLogin, and verifyShortCode templates
* Adds a conditional class for 'location' partial and tweaks footer margin-top for consistency, adds location mock to share across stories
* Adds stories, tests
* Fixes other stories/removes unused vars in other stories

closes #9312
closes #9309
closes #9306

## Checklist


packages/fxa-auth-server/lib/senders/emails/partials/automatedEmailResetPassword/index.mjml
```diff
@@ -1,15 +1,21 @@
...
   <mj-column>
     <mj-text css-class="text-footer-automatedEmail">
       <span data-l10n-id="automated-email-reset">
         This is an automated email; if you did not authorize this action, then
-        <a class="link-blue" href="<%- resetLink %>" data-l10n-name="resetLink">please reset your password</a>.
-        For more information, please visit
-        <a class="link-blue" href="<%- supportUrl %>" data-l10n-name="supportLink">Mozilla Support</a>.
+        <a class="link-blue" href="<%- resetLink %>" data-l10n-name="resetLink"
+          >please reset your password</a
+        >. For more information, please visit
```
LZoog, line 12: @jodyheavener We should probably get prettier figured out for these files sooner than later, maybe in a follow up we can mass format the MJML files we have so far. This seems to be how it default formats HTML files but I'm sure there's rules we can set because I also don't particularly like this 😅

packages/fxa-auth-server/lib/senders/emails/templates/lowRecoveryCodes/index.stories.ts
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { Meta } from '@storybook/html';
-import { Template, storyWithProps } from '../../storybook-email';
+import { storyWithProps } from '../../storybook-email';
```
LZoog, line 6: @jodyheavener We should probably also remove unused imports on save with eslint when we do that prettier follow up. I'm also not sure why `buildStory` wasn't erroring in the other story files I fixed 🤔

packages/fxa-auth-server/lib/senders/emails/templates/newDeviceLogin/index.stories.ts
```diff
@@ -13,15 +14,18 @@ const createStory = storyWithProps(
...
   }
 );
 
-export const NewDeviceLogin = createStory();
+export const NewDeviceLoginFirefox = createStory(
+  {},
+  'New device login through Firefox'
+);
+export const NewDeviceLoginOther = createStory(
+  { clientName: 'Some Other Relier' },
```
LZoog, line 29: I think we should show a different story for any variable that can differ like this so it's clear from a Storybook documentation perspective that it's dynamic.

packages/fxa-auth-server/lib/senders/emails/templates/verifyPrimary/index.stories.ts
```diff
@@ -0,0 +1,24 @@
...
+import { MOCK_LOCATION } from '../../partials/location/mocks';
+import { storyWithProps } from '../../storybook-email';
+
+export default {
+  title: 'Emails/verifyPrimary',
+} as Meta;
+
+const createStory = storyWithProps(
+  'verifyPrimary',
+  'Sent to users with an unverified primary email, meaning an unverified account, when they attempt an action requiring a verified account.',
```
LZoog, line 15: @vbudhram this seem right to you?

packages/fxa-auth-server/lib/senders/emails/partials/location/mocks.ts
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export const MOCK_LOCATION = {
```
LZoog, line 5: It'd probably benefit us to do this for some other variables too in future PRs, like `passwordChangeLink` / anything else we're having to copy and paste between stories.

packages/fxa-auth-server/lib/senders/emails/templates/unblockCode/index.stories.ts
```diff
@@ -3,19 +3,18 @@
...
+import { storyWithProps } from '../../storybook-email';
 
 export default {
   title: 'Emails/unblockCode',
 } as Meta;
 
-const createStory = buildStory(
+const createStory = storyWithProps(
   'unblockCode',
   'Sent to verify an account sign-in via code.',
```
LZoog, line 15: Just noticed, this documentation for `unblockCode` could probably be more descriptive, like, `'Sent to verify a blocked account sign-in via code.'` if that's correct. But even then, when is an account blocked or why would they get this instead of `verifyLoginCode`?

packages/fxa-auth-server/lib/senders/emails/partials/automatedEmailResetPassword/index.mjml
```diff
@@ -1,15 +1,21 @@
-<%# This Source Code Form is subject to the terms of the Mozilla Public
-  # License, v. 2.0. If a copy of the MPL was not distributed with this
-  # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+<%# This Source Code Form is subject to the terms of the Mozilla Public #
+License, v. 2.0. If a copy of the MPL was not distributed with this # file, You
+can obtain one at http://mozilla.org/MPL/2.0/. %>
```
jodyheavener, line 3: 👀 I'm guessing prettier did this? I think we should consider disabling prettier for MJML files since it's producing less-than-optimal results.

packages/fxa-auth-server/lib/senders/emails/partials/automatedEmailResetPassword/index.mjml
jodyheavener: Yeah agree, would you be okay with just not formatting MJML files for now?

packages/fxa-auth-server/lib/senders/emails/templates/verifyPrimary/index.txt
```diff
@@ -0,0 +1,13 @@
+verifyPrimary-title = "Verify primary email"
+
+verifyPrimary-description = "A request to perform an account change has been made from the following device:"
+<%- include('/partials/location/index.txt') %>
+
+verifyPrimary-action-plaintext = "Verify email:"
+<%- link %>
+
+verifyPrimary-post-verify = "Once verified, account changes like adding a secondary email will become possible from this device."
```
jodyheavener, line 9: This doesn't appear to be in the original plaintext version. I am okay with adding it in since it makes sense, but just want to acknowledge this.

packages/fxa-auth-server/lib/senders/emails/templates/verifyLogin/index.txt
```diff
@@ -0,0 +1,11 @@
...
+
+verifyLogin-description = "For added security, please confirm this sign-in from the following device:"
+
+<%- include('/partials/location/index.txt') %>
+
+verifyLogin-action = "Confirm sign-in:"
+<%- link %>
+
+<%- include('/partials/changePassword/index.txt') %>
+<%- include('/partials/support/index.txt') %>
```
jodyheavener, line 11: This doesn't appear to be in the original plaintext version. I'm also okay with keeping it here since it makes sense in this context, but should we consult with product folks for additions like this?

packages/fxa-auth-server/lib/senders/emails/partials/changePassword/en-US.ftl
```diff
@@ -0,0 +1,2 @@
+change-password-plaintext = If you suspect that someone is trying to gain access to your account, please change your password.
+{ $passwordChangeLink }
```
jodyheavener, line 2: I don't think FTL likes this. Could it be inline with the localization?

jodyheavener: Or actually, if the link is _supposed_ to be on a new line I would recommend moving the rendering of the link to EJS in the plaintext:

```
change-password-plaintext = "If you suspect that someone is trying to gain access to your account, please change your password."
<%- passwordChangeLink %>
```

In this case we wouldn't include it in the FTL at all.

packages/fxa-auth-server/lib/senders/emails/partials/location/index.mjml
```diff
@@ -1,30 +1,27 @@
...
+can obtain one at http://mozilla.org/MPL/2.0/. %>
 
 <mj-include path="./css/location/index.css" type="css" css-inline="inline" />
 
 <mj-section>
   <mj-column>
-    <mj-text css-class="text-body-grey">
-    <% if (locals.primaryEmail) { %>
+    <mj-text css-class="<%= locals.cssClass || 'text-body-grey' %>">
+      <% if (locals.primaryEmail) { %>
```
jodyheavener, line 10: Eek, I don't like that prettier is doing this.

packages/fxa-auth-server/lib/senders/emails/templates/lowRecoveryCodes/index.stories.ts
jodyheavener: Oh what the heck, I did a find and replace for all the instances of `buildStory`. I wonder how they got in...

jodyheavener: Ohhh I think I had a new template PR simultaneously open with the Storybook PR, so I made the storybook change but didn't update it in my template PR after it got merged.

packages/fxa-auth-server/lib/senders/emails/templates/passwordChanged/index.mjml
```diff
@@ -16,5 +16,5 @@
   </mj-column>
 </mj-section>
 
-<%- include('/partials/location/index.mjml') %>
+<%- include('/partials/location/index.mjml', { cssClass: 'text-body-grey-no-margin' }) %>
```
jodyheavener, line 19: Hmm, so this is a new pattern, and I don't mind it, but I kinda wonder if instead of modifying the margin here perhaps we could modify the margins around it? I think we already have classes for that?

packages/fxa-auth-server/lib/senders/emails/templates/unblockCode/index.stories.ts
jodyheavener: That's fair. I based these descriptions off of that [original spreadsheet](https://docs.google.com/spreadsheets/d/11xFyTRpYkoYcUc3Ac1x6z976AvMpgRaV6blGK9Q838g/edit#gid=1031776270), where they are pretty short and to the point. So that's what I kept doing here: short and to the point. I was actually going to maybe suggest that we shorten your description in the `verifyPrimary` story a tad to be consistent with the rest. Or we can go back in and update all the existing ones to be even more descriptive, but in some cases that might not be totally possible, and I don't think it should take significant time to hunt down and compose lengthy descriptions given the volume of templates we have left to convert.

packages/fxa-auth-server/lib/senders/emails/partials/location/mocks.ts
```diff
@@ -0,0 +1,11 @@
...
+  date: 'Thursday, Sep 2, 2021',
+  device: 'Firefox on Mac OSX 10.11',
+  ip: '10.246.67.38',
+  location: 'Madrid, Spain (estimated)',
+  time: '12:26:44 AM (CEST)',
+};
```
jodyheavener, line 11: Good thinking :)

packages/fxa-auth-server/lib/senders/emails/templates/verifyLogin/index.mjml
```diff
@@ -0,0 +1,33 @@
...
+</mj-section>
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-body"
+      ><span data-l10n-id="verifyLogin-description"
+        >For added security, please confirm this sign-in from the following
+        device:</span
+      ></mj-text
+    >
```
jodyheavener, line 27: 😣 dis formatting

packages/fxa-auth-server/lib/senders/emails/templates/verifyPrimary/index.mjml
```diff
@@ -0,0 +1,40 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public #
+License, v. 2.0. If a copy of the MPL was not distributed with this # file, You
+can obtain one at http://mozilla.org/MPL/2.0/. %>
+
+<mj-include path="./css/global.css" type="css" css-inline="inline" />
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-header">
+      <span data-l10n-id="verifyPrimary-title"> Verify primary email </span>
```
jodyheavener, line 10: Can we zap the spaces around this text? Or did Prettier do this?

packages/fxa-auth-server/lib/senders/emails/global.scss
```diff
@@ -67,6 +67,9 @@ $s-10: 40px;
   &-8 {
     margin-top: $s-8 !important;
   }
+  &-10 {
+    margin-top: $s-10 !important;
+  }
```
jodyheavener, line 72: Kinda wonder if it would be worth putting these `$s-X` values into a map and then iteratively generating all the `.mX-Y` classes. This wouldn't result in excessive classes since they're inlined anyway.

packages/fxa-auth-server/lib/senders/emails/partials/changePassword/index.txt
```diff
@@ -0,0 +1 @@
+change-password-plaintext = "If you suspect that someone is trying to gain access to your account, please change your password. { $passwordChangeLink }"
```
jodyheavener, line 1: Ahhh, cool - I also added this in #10418 (but I didn't do a partial like you did!)

This appears to be a variation on the `automatedEmailChangePassword` partial, which is itself a variation on `automatedEmailResetPassword` and `automatedEmailNoAction`. I think it would be worth a follow up to try to come up with a dynamic partial that could handle all these cases and allow us to remove all these partials.

What you think?

jodyheavener (changes requested):
Wicked! They all look great, and I just have a bunch of nitpicks and comments. I'd loooove to either revert the prettier formatting if we can't find an ideal solution for it in this PR.

packages/fxa-auth-server/lib/senders/emails/partials/automatedEmailResetPassword/index.mjml
LZoog: Sure, so long as you don't mind if I temporarily add `*.mjml` to the `.prettierignore` file. I'm not sure why yours was not formatting without sticking it in there 🤔

packages/fxa-auth-server/lib/senders/emails/partials/automatedEmailResetPassword/index.mjml
LZoog: Yeah, this was prettier's doing. I think if possible we might consider changing these MJML files to EJS files

packages/fxa-auth-server/lib/senders/emails/partials/changePassword/index.txt
LZoog: I did find it kind of weird that the plaintext version of this differs from what the HTML email reads, it makes me think the HTML was changed sometime for one of these template and the corresponding TXT was not updated. I kind of wondered if we had a meeting with... not sure, maybe Design? if we could just make this more consistent across the board rather than making a more dynamic partial. Either way I can file an issue for it.

packages/fxa-auth-server/lib/senders/emails/partials/location/index.mjml
LZoog: The extra tab? That one actually seems correct to me since it's in `mj-text` but yeah I'll revert some of these prettier changes.

packages/fxa-auth-server/lib/senders/emails/templates/passwordChanged/index.mjml
LZoog: You're supposed to say "gosh this is cool!" 😉

For this particular case, I didn't see a way around it. We either need to remove the margin-bottom of the location partial (which is what I did), or reduce the margin-top of the automated email footer text.

With change:
image

Without change:
image

I could see us using this pattern in other templates too, though I suppose I can't say for sure.

packages/fxa-auth-server/lib/senders/emails/partials/location/index.mjml
jodyheavener: Looks like the resulting formatting is flat:

Screen Shot 2021-09-10 at 11 40 27 AM

packages/fxa-auth-server/lib/senders/emails/templates/passwordChanged/index.mjml
jodyheavener: Sounds good! And is _IS_ cool!

packages/fxa-auth-server/lib/senders/emails/templates/unblockCode/index.stories.ts
LZoog: Hmm I feel the entire point of documenting these is to disambiguate when certain emails are sent/received though. Reading this, I don't know when a user would receive the `unblockCode` email over a more standard `verifyLoginCode` email.  I'm fine with simpler descriptions so long as we leave that follow up to go in and edit these further, otherwise I think the documentation description should be as clear as we can be while also not making it like multiple sentences or something. 🤔

packages/fxa-auth-server/lib/senders/emails/templates/verifyLogin/index.txt
LZoog: Heh yeah, good catch. I also noticed it was missing in the old version, but that seems like a mistake to me. We include this partial for any other time we have that support text in the HTML.

How about I leave it, and start a doc of some sort to keep note of anything like this that we can run by product or design or whomever to confirm?

packages/fxa-auth-server/lib/senders/emails/templates/verifyPrimary/index.txt
LZoog: That also seems like a mistake to me on the old templates. I'll add it to that doc I mentioned in another comment.

packages/fxa-auth-server/lib/senders/emails/templates/verifyLogin/index.txt
jodyheavener: That seems reasonable to me!

packages/fxa-auth-server/lib/senders/emails/global.scss
LZoog: I agree, though I just pushed an update containing everything except this - maybe one of us can take a stab at it with the next set of templates. It might also be worth separating some of these out into separate scss files and then importing them into this global file.

jodyheavener (approved):
These looks great! I have one more ask but I'll r+ so you can merge as soon as it's green again

packages/fxa-auth-server/lib/senders/emails/partials/changePassword/index.txt
```diff
@@ -0,0 +1,2 @@
...
```
jodyheavener, line 1: I think, since you’re rendering `passwordChangeLink` in EJS format, the inline `{ $passwordChangeLink }` bit can be removed

LZoog: Ah oops, yeah I missed that. Fixed.

PR #10443, opened 2021-09-15
feat(auth-server): Allow layout-only email stories

LZoog:
## Because

* Our emails have a different state at the layout level ('if sync') and we want to show those states. Also, a different layout exists for the SubPlat emails and we want to be able to show that.

## This pull request

* Adds a storybook-only directory with template and txt file, as the renderer expects to pull in a template and txt file from the templates/ dir. This was preferable over modifying when the renderer pulls in files because that same code is ran in prod and we didn't want to modify it just for layout-specific storybook files.
* Removes the global CSS file from templates, it's already included in the fxa layout file
* Fixes the width on sync image to correct ratio

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.stories.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.stories.ts
@@ -3,21 +3,24 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { Meta } from '@storybook/html';
-import { MOCK_LOCATION } from '../../partials/location/mocks';
 import { storyWithProps } from '../../storybook-email';
 
 export default {
-  title: 'Templates/verify',
+  title: 'Layouts/FxA',
 } as Meta;
 
-const createStory = storyWithProps(
-  'verify',
-  'Received by all who complete FxA registration form.',
-  {
-    ...MOCK_LOCATION,
-    link: 'http://localhost:3030/verify_email',
-    sync: true,
-  }
-);
+const createStory = storyWithProps('_storybook', 'The FxA email base layout.', {
+  sync: false,
+  subject: 'N/A',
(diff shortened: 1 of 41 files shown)


LZoog:
Here's a couple screenshots:

image
image

We might need a bit more tweaking once we create the subplat layout.

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.stories.ts
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Meta } from '@storybook/html';
+import { storyWithProps } from '../../storybook-email';
+
+export default {
+  title: 'Layouts/FxA',
```
jodyheavener, line 9: If we're gonna call this "Layouts" and not place it under the Emails section, can we rename the other ones from "Emails" to "Templates"?

jodyheavener (commented):
LGTM, I only have one small nitpick. And I guess this needs a rebase from when main was failing.

packages/fxa-auth-server/lib/senders/emails/layouts/fxa/index.stories.ts
LZoog: Huh I see your point. I guess it felt like those are the actual emails that get sent, so the "emails" title made sense and may be more clear from a product stand point... I was going to suggest changing this to "Email Layouts" and leaving it as "Emails," but maybe you're right that "Layouts and "Templates" makes sense. If we ever wanted to show partials in the future, it'd make sense to have the separation as you have it too.

jodyheavener (approved):
🙌

Issue #10462, opened 2021-09-16
123done passwordless stage url points to local host


vbudhram:
Add support for passwordless views in stage.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3998)

PR #10463, opened 2021-09-16
fix(tests): Update 123done stage url

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/10462


chenba:
I think if you rebase on main to undo a fluent/react update that'll fix CI.

chenba (approved):

PR #10476, opened 2021-09-20
fix(auth): include emails-scss in build command

jodyheavener:
Discovered by Danny, this just adds the `emails-scss` command to the build command in the auth server. This should resolve the sending issues in stage.

Confirmed that this works for me locally when sending emails using the dist auth server - and after understanding the solution was able to effectively repro the issue.

(The auth server build command copies everything from the lib folder, and I had been manually running that command so it was including the compiled CSS already, which is why I wasn't seeing it locally when trying to repro.)

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/renderer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/renderer.ts
@@ -10,7 +10,7 @@ import mjml2html = require('mjml');
 type TemplateComponent = 'layouts' | 'templates' | 'partials';
 export type TemplateContext = Record<string, any>;
 
-const TEMPLATES_DIR = './lib/senders/emails/';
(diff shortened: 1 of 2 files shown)


LZoog (approved):
Lol🤦‍♀️

LZoog:
@jodyheavener Actually I see Danny's comment from Slack:

> @Jody adding emails-scss to the build step fixed it for me, along with the path fix in renderer.ts

I think we need the [path fix here](https://github.com/mozilla/fxa/compare/main...lz/log-mjml-issue-with-path-fix#diff-a47098e5384a5e30dec4812a1d04b8e1480086cb0522bc4c24d6bb6ff3404731R14) as well?

dannycoates (approved):

PR #10477, opened 2021-09-20
feat(auth): unify sub update for stripe change

### LINKED ISSUE #10093: Update Stripe webhooks to account for IAP
Stripe webhooks trigger capability broadcasts, but fail to take into account capabilities granted by IAP. They should be updated to use the CapabilityService to track total capabilities and use the event broadcasts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3837)

bbangert:
Because:

* We want to ensure we don't tell relying parties to disable/enable
  products just because the Stripe subscription changes.

This commit:

* Updates stripe webhooks to use the capability service so they can take
  IAP purchases into account before possibly broadcasting changes.

Closes #10093

## Checklist


packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -72,6 +73,70 @@ export class CapabilityService {
...
   }
 
+  /**
+   * Handle a Stripe Webhook subscription change.
+   *
+   * This handles broadcasting and refreshing the subscription capabilities for
+   * the product ids that were possibly updated.
+   *
+   * Stripe supports aligned subscriptions such that a single subscription can
+   * include multiple items for multiple products. We iterate through this list
```
chenba, line 83: afaik we do not use the multiple product per subscription feature.  Do we plan to in the near future?

bbangert: We're not using it at the moment, but its easy enough to support now rather than making things more difficult in the future. We have had aligned subscriptions requested in the past.

chenba (approved):

Issue #555, opened 2019-02-21
Need to investigate if we can do scopeToArray using `fxa-shared`


vladikoff:
We have 2 places where we do stuff with scopes and use `scopeToArray`. 
We should migrate that to use the utils in `fxa-shared`

Follow up from: https://github.com/mozilla/fxa-content-server/pull/6876

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-440)
┆Issue Number: FXA-440

PR #10480, opened 2021-09-20
feat(auth): add subplat email layout

jodyheavener:
## Because

- We're going to start working on Subplat emails

## This pull request

- Adds a new layout for subplat emails
- Re-arranges some of the FxA layout styles and partials to work accomodate the new layout

Screen Shot 2021-09-20 at 4 28 03 PM

I expect I will have missed at least _something_ here, so another set of eyes will be good.


packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.scss
```diff
@@ -0,0 +1,54 @@
...
+  @extend .footer-text;
+
+  div {
+    @extend .mb-3;
+  }
+}
+
+.footer-link {
+  color: global.$white !important;
+}
```
LZoog, line 54: Do you think we should enforce consistency between the FxA and Subplat headings/body text? If we're going to match to parity here, I think FxA might need `text-body` and `text-header` scoped there and then different styles here since (especially the header text) they look fairly different. I'm also OK with poking a designer and asking them about making those things consistent now though.

packages/fxa-auth-server/lib/senders/emails/storybook-email.ts
```diff
@@ -116,3 +128,15 @@ export const storyWithProps = (
     return template;
   };
 };
+
+export const subplatStoryWithProps = (
```
LZoog, line 132: 👍

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.mjml
```diff
@@ -0,0 +1,97 @@
...
+        </div>
+      <% } %>
+    </mj-raw>
+
+    <mj-section full-width>
+      <mj-column css-class="header-container">
+        <mj-image css-class="fxa-logo"
+          href="https://www.mozilla.org/firefox/new"
+          src="https://accounts-static.cdn.mozilla.net/product-icons/firefox-logo-email.png"
+          alt="Firefox logo">
```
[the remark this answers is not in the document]
jodyheavener: > I was also going to suggest we reuse this image in a logo partial, but looks like this links out. 

Do you mean reuse it here and in the FxA layout? They're different images :D

> Do you think the fxa layout needs to use mj-image

I don't think it's the end of the world, but for consistency it'd probably be helpful.

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.scss
jodyheavener: I'm also leaning towards parity for now, I've added a comment to the confirmation doc

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.mjml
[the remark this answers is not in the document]
LZoog: > Do you mean reuse it here and in the FxA layout? They're different images :D

Ah oops, you rite. 

> I don't think it's the end of the world, but for consistency it'd probably be helpful.

Yeah, maybe next time we touch an FxA email, we can update that at the same time.

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.scss
LZoog: For parity, I think you need to move `text-body` and `text-header` into the FxA styles and then add the same classes with SubPlat styles to that subplat stylesheet. Still seeing FxA styles here:

image

LZoog: Actually/hopefully, I think the only thing you need to move to FxA styles is `@extend .font-sans` and the `text-align center` for those two classes and then set those in subplat styles. Maybe we can get away with sharing the font-size but if not we'll have to move that as well

jodyheavener: It ended up being a little more substantial just because the classes beneath `%text-body-common` are already constructed by the time the styles in the fxa-specific stylesheet are constructed, so overriding `%text-body-common` further down wouldn't have effect. Open to feedback for sure!

jodyheavener:
Updated:

Screen Shot 2021-09-21 at 2 10 17 PM

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.scss
```diff
@@ -0,0 +1,74 @@
...
+
+%text-body-common {
+  @extend %text-body-common;
+  @extend .text-md;
+}
+
+.text-body div {
+  @extend %text-body-common;
+  @extend .mb-6;
+  color: global.$grey-500 !important;
```
jodyheavener, line 28: This should technically be `#555` but I didn't want to introduce a fourth grey,

packages/fxa-auth-server/lib/senders/emails/layouts/subplat/index.scss
LZoog: I'll see if there's ways to improve this in future PRs but I don't want to block template conversions so seems fine to me

LZoog (approved):

PR #10488, opened 2021-09-21
chore(release): make sure you're not on main if you're doing a patch release

jodyheavener:
Got burned by this when doing my first patch release in a while :D


chenba (approved):
💯

PR #10498, opened 2021-09-21
fix(auth): stop reporting errors on legit credit notes

### LINKED ISSUE #9323: We should not report sentry errors for non-paypal credit notes.
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/11528137/?referrer=github_plugin

    Error: Credit note issued for account balance: cn_1J08XjJNcmPzuWtR4JdJljia
      File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.js", line 131, in StripeWebhookHandler.handleCreditNoteEvent
        sentry_1.reportSentryError(new Error(`Credit note issued for account balance: ${creditNote.id}`), request);
      ?, in runMicrotasks
      File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.js", line 59, in async StripeWebhookHandler.handleWebhookEvent
        await this.handleCreditNoteEvent(request, event);
    ...
    (4 additional frame(s) were not displayed)We should ignore invoices with charge automatically, and if its send_invoice and not out_of_band, then we report it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3619)

bbangert:
Because:

* We are incorrectly reporting Sentry errors when refunding users
  that paid via Stripe on credit notes.

This commit:

* Only reports via Sentry if a Stripe paying user was credited out of
  band or on their account balance.

Closes #9323

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -147,17 +147,23 @@ export class StripeWebhookHandler extends StripeHandler {
       'invoices'
     );
 
-    // This invoice must be for a paypal subscription for us to issue a
-    // paypal refund and be marked for out_of_band refund.
-    if (
-      invoice.collection_method !== 'send_invoice' ||
-      !creditNote.out_of_band_amount ||
-      creditNote.out_of_band_amount === 0
-    ) {
-      reportSentryError(
-        new Error(`Credit note issued for account balance: ${creditNote.id}`),
-        request
-      );
+    if (invoice.collection_method === 'charge_automatically') {
+      // This is a Stripe charge, report if needed and return as Stripe handles
+      // refunding the customer.
+      if (
+        creditNote.customer_balance_transaction ||
+        creditNote.out_of_band_amount
+      ) {
+        // We should be informed if it was applied to the account balance as they
+        // should be refunded to the card or if it was applied out of band.
+        reportSentryError(
+          new Error(
+            `Credit note issued for account balance or out of band: ${creditNote.id}`
+          ),
+          request
+        );
+      }
+
       return;
     }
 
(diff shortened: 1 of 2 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -147,17 +147,23 @@ export class StripeWebhookHandler extends StripeHandler {
...
-    ) {
-      reportSentryError(
-        new Error(`Credit note issued for account balance: ${creditNote.id}`),
-        request
-      );
+    if (invoice.collection_method === 'charge_automatically') {
+      // This is a Stripe charge, report if needed and return as Stripe handles
+      // refunding the customer.
+      if (
+        creditNote.customer_balance_transaction ||
```
biancadanforth, line 154: Wouldn't we still want to check for an account balance for PayPal customers and report an error? That would mean it was issued incorrectly, right?

Admittedly, I guess we weren't checking for that here originally for either Stripe or PayPal.

biancadanforth (approved):
Thanks Ben; LGTM. Just had a follow-up question re: account balances and PayPal customers.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: This only exists if the credit note adjusts the account balance, which only happens if the user is being credited to their balance which we currently don't want.

biancadanforth: Hmm yeah I understand that part, but we aren't checking if the credit note adjusts the account balance for PayPal customers with this logic -- only for Stripe customers. Shouldn't we also check this for PayPal customers?

bbangert: We will want that in a future PR to allow partial paypal refunds. There's a if block farther down here that ensures the out_of_band amount perfectly matches the invoice amount. If there was a account balance change for a PayPal user, then that check below will fail and it will error.

PR #10515, opened 2021-09-23
refactor(auth): default ftl locale to en

jodyheavener:
Because: `en`, instead of `en-US`, is now considered the source content for l10n.


vbudhram (approved):
Looks like github complaining about your email on commit.

LZoog:
For context here (because I didn't know this), cocomo mentioned this to Jody and I when looking at our new Fluent stuff:

> One thing on en-US.ftl, mozorg and Relay have decided to change it to en so if you have en-US only content, it can go there, but `en` is the source content for l10n.

Issue #10097, opened 2021-08-12
[subplat] Implement P2 metrics for passwordless flow


data-sync-user:
From https://docs.google.com/document/d/1x0lYak98rM8ickwDciTn99Sd9-r1j6PmvAa8SP7CxGg/edit#heading=h.yzcz14jk0h70

We should implment the P2 metrics

* How many users choose to sign into Firefox Account
* How many users click “Continue with PayPal” 
* How many users enter information into any field and don’t click subscribe. 
* Ex: A user put in their email address, and name on their credit card, but then didn’t end up subscribing. 
* How many times we show the message that a user already has a FxA when trying to create an account. 
* This is a guardrail metric to see if we need to modify this experience in the future.
* How many users set their password. How many users do not set their password.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3840)

data-sync-user:
➤ Luminita Buruian commented:

Vijay Budhram Is there something the QA team needs to/can do on this issue?

Thank you

PR #10516, opened 2021-09-23
fix(metrics): Add remaining metrics and `fxa_reg - complete` for pwordless accounts

vbudhram:
## Because

- We didn't emit the complete event when a user set their password
- By propagating the flowId, we can now correctly track all P2 metrics
- TBD on the tests

## This pull request

- Correctly stashes the flowId when stub account is created and unstashes when password is set

Closes: https://github.com/mozilla/fxa/issues/10097

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/context.js
+++ b/packages/fxa-auth-server/lib/metrics/context.js
@@ -196,6 +196,13 @@ module.exports = function (log, config) {
         id: request.payload.code,
       };
     }
+
+    if (request.payload && request.payload.uid && request.payload.token) {
+      return {
+        uid: request.payload.uid,
+        id: request.payload.uid,
+      };
+    }
   }
 
   /**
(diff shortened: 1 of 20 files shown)


packages/fxa-auth-server/lib/metrics/context.js
```diff
@@ -196,6 +196,13 @@ module.exports = function (log, config) {
         id: request.payload.code,
       };
     }
+
+    if (request.payload && request.payload.uid && request.payload.token) {
```
dannycoates, line 200: i get it but woof 🙈

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -577,8 +585,9 @@ export class AccountHandler {
         authSalt,
         this.config.verifierVersion
       );
+      const metricsContext = await request.gatherMetricsContext({});
+      await this.signupUtils.verifyAccount(request, account, {});
```
dannycoates, line 589: I _think_ this is fine. Are we sure we want to do everything in `verifyAccount`? (just double checking)

dannycoates (approved):

packages/fxa-auth-server/lib/routes/account.ts
vbudhram: I believe so. We send a post verify email for sync users in this function but that wouldn't really get triggered in flow.

Issue #9725, opened 2021-06-23
Send email directly from auth-server


data-sync-user:
Original title “Convert email-service to direct db access”

After further investigation it would be better to skip email-service altogether and send directly to SES.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3685)

data-sync-user:
➤ Danny Coates commented:

On track to be complete in train 216 (Sept 29)

Issue #10527, opened 2021-09-27
Record when a user is prevented from logging in due to an email block


data-sync-user:
Following our [FxA Ops weekly meeting on Sep 14](https://docs.google.com/document/d/1UV8SEeyk-8ZsZr7SZ2ZhTv58NiW10d5K6gHBQr2iAJE/edit#bookmark=id.g9s1d4faa0r), we would like to measure how many users are prevented from logging in, because FxA has stopped sending emails to them.

See also Wil’s doc: [Emails may be undeliverable to some users](https://docs.google.com/document/d/1lXa6ZwxtyXbzB-YQOr0HuRRwtKzqE2etiWUwEN9IRok/edit#).

Part of this issue is identifing different cases and when an email block does actually play a role during login. For example, a user who knows their password and whose email has already been verified would still be able to login even if FxA wasn't sending them emails anymore.

Vijay referenced the [Ecosystem Platform docs](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-email) as a good place to start.

If there are good places to add a StatsD metric, we should add that in the source code. A separate ticket should be filed to create a panel in Grafana to monitor these numbers over time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4001)

data-sync-user:
➤ Bianca Danforth commented:

Thanks Vijay; that would indeed be a great place to add some additional logging. I’m just not sure if we can pull from log.error in Grafana as a data source. I will confirm that next.

Note: I tried faking an email block for a fake FxA account locally, but I learned that the auth server isn’t configured to use the fxa-email-service in local dev, and that fxa-email-service is being deprecated soon by dcoates (along with fxa-auth-my-sql-db, which it relies on).

data-sync-user:
➤ Bianca Danforth commented:

[...]

PR #10532, opened 2021-09-27
feat(auth): enable nodemailer SES transport

dannycoates:
## Because

- I'd like to deprecate email-service

## This pull request

- Adds the nodemailer SES transport
- The AWS config is set by the environment and is used when no SMTP user:password are set
- We _could_ use liveConfig to run a subset of addresses through direct SES vs email-service
  - https://github.com/mozilla/fxa/blob/auth-ses/packages/fxa-auth-server/lib/senders/select_email_services.js#L18
- Gets used 100% when email-service has been configed off with `EMAIL_CONFIG_USE_REDIS=false`
- Resurrects some old code from https://github.com/mozilla/fxa-auth-server/pull/1934 🧟 🧙

## Notes

After this we'll have 3 ways to send via SES: API direct, SMTP, fxa-email-service. The goal is to use API direct and eliminate fxa-email-service, with an SMTP "backup" in case we need to support other providers.

Closes: #9725

### DIFF
--- a/_dev/pm2/infrastructure.config.js
+++ b/_dev/pm2/infrastructure.config.js
@@ -1,73 +1,66 @@
 module.exports = {
   apps: [
     {
-      name: "mysql",
-      script: "_scripts/mysql.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
+      name: 'mysql',
+      script: '_scripts/mysql.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
+      kill_timeout: 20000,
     },
     {
-      name: "redis",
-      script: "_scripts/redis.sh",
+      name: 'redis',
+      script: '_scripts/redis.sh',
       env: {
-        PORT: "6379"
+        PORT: '6379',
       },
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
+      max_restarts: '1',
+      min_uptime: '2m',
+      kill_timeout: 20000,
     },
     {
-      name: "memcache",
-      script: "_scripts/memcached.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
+      name: 'memcache',
+      script: '_scripts/memcached.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
+      kill_timeout: 20000,
     },
     {
-      name: "sns",
-      script: "_scripts/goaws.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
+      name: 'sns',
+      script: '_scripts/goaws.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
       autorestart: false,
-      kill_timeout: 20000
+      kill_timeout: 20000,
     },
     {
-      name: "pubsub",
-      script: "_scripts/pubsub.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
+      name: 'pubsub',
+      script: '_scripts/pubsub.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
+      kill_timeout: 20000,
     },
     {
-      name: "firestore",
-      script: "_scripts/firestore.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
+      name: 'firestore',
+      script: '_scripts/firestore.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
+      kill_timeout: 20000,
     },
     {
-      name: "sync",
-      script: "_scripts/syncserver.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
+      name: 'sync',
+      script: '_scripts/syncserver.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
       autorestart: false,
-      kill_timeout: 20000
+      kill_timeout: 20000,
     },
     {
-      name: "pushbox",
-      script: "_scripts/pushbox.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      args: "3306 root@mydb:3306",
-      kill_timeout: 20000
+      name: 'pushbox',
+      script: '_scripts/pushbox.sh',
+      max_restarts: '1',
+      min_uptime: '2m',
+      args: '3306 root@mydb:3306',
+      kill_timeout: 20000,
     },
-    {
-      name: "email",
-      script: "_scripts/fxa-email-service.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
-    }
-  ]
+  ],
 };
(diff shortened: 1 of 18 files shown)


_dev/pm2/infrastructure.config.js
```diff
@@ -1,73 +1,66 @@
...
+      args: '3306 root@mydb:3306',
+      kill_timeout: 20000,
     },
-    {
-      name: "email",
-      script: "_scripts/fxa-email-service.sh",
-      max_restarts: "1",
-      min_uptime: "2m",
-      kill_timeout: 20000
-    }
```
dannycoates, line 71: removing email-service from local dev

packages/fxa-auth-server/bin/key_server.js
```diff
@@ -110,10 +110,11 @@ async function run(config) {
   );
   const profile = require('../lib/profile/client')(log, config, statsd);
   Container.set(ProfileClient, profile);
+  const bounces = require('../lib/bounces')(config, database);
   const senders = await require('../lib/senders')(
     log,
     config,
-    error,
```
dannycoates, line 116: `error` was no longer used downstream so I reused it's slot for `bounces`

packages/fxa-auth-server/config/index.ts
```diff
@@ -387,6 +389,40 @@ const conf = convict({
       default: '',
       env: 'SES_CONFIGURATION_SET',
     },
+    bounces: {
```
dannycoates, line 392: resurrecting the pre-email-service config from https://github.com/mozilla/fxa-auth-server/pull/1934

packages/fxa-auth-server/config/index.ts
```diff
@@ -387,6 +389,40 @@ const conf = convict({
...
+        default: true,
+        env: 'BOUNCES_ENABLED',
+      },
+      complaint: {
+        doc: 'Tiers of max allowed complaints per amount of milliseconds',
+        default: {
+          // 0 are allowed in the past day.
+          // 1 is allowed in the past week.
+          0: ONE_DAY,
+          1: 7 * ONE_DAY,
```
dannycoates, line 404: changed this from 1 year to a more sane 7 days

packages/fxa-auth-server/config/index.ts
```diff
@@ -387,6 +389,40 @@ const conf = convict({
...
+        },
+        env: 'BOUNCES_COMPLAINT',
+      },
+      hard: {
+        doc: 'Tiers of max allowed hard bounces per amount of milliseconds',
+        default: {
+          // 0 are allowed in the past day.
+          // 1 is allowed in the past two days.
+          0: ONE_DAY,
+          1: 2 * ONE_DAY,
```
dannycoates, line 414: used to be 1 year. no wonder why people have problems with bounces 😳

packages/fxa-auth-server/lib/bounces.js
```diff
@@ -0,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
dannycoates, line 1: All from https://github.com/mozilla/fxa-auth-server/pull/1934

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -481,98 +490,113 @@ module.exports = function (log, config) {
...
-            text: localized.text,
-            html: localized.html,
-            xMailer: false,
-            headers,
-          };
+      const { mailer, emailAddresses, emailService, emailSender } = service;
+      const to = emailAddresses[0];
+
+      try {
+        await bounces.check(to);
```
dannycoates, line 512: we check bounces 🏀

packages/fxa-auth-server/test/local/bounces.js
```diff
@@ -0,0 +1,184 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
dannycoates, line 1: used unit tests - good condition, runs, new promises, $400 OBO

original owner: https://github.com/mozilla/fxa-auth-server/pull/1934

packages/fxa-auth-server/test/local/senders/email.js
```diff
@@ -1695,19 +1697,6 @@ describe('lib/senders/email:', () => {
...
-        email: 'test@restmail.net',
-        subject: 'subject',
-        template: 'verifyLogin',
-        uid: 'foo',
-      };
-
-      return mailer.send(message).then((status) => {
-        assert.deepEqual(status, [{ resp: 'ok' }]);
-      });
-    });
```
dannycoates, line 1709: nothing was using the result so i changed the `send` signature to return void

packages/fxa-auth-server/lib/senders/email.js
```diff
@@ -481,98 +490,113 @@ module.exports = function (log, config) {
...
+      } catch (err) {
+        log.error('mailer.send.error', {
+          err: err.message,
+          code: err.code,
+          errno: err.errno,
+          to,
+          emailSender,
+          emailService,
+          template,
+        });
```
dannycoates, line 522: This is copied from below in the `sendMail` callback, and is related to @biancadanforth's PR #10543 ... but I think it'd be better as it's own distinct log entry. I'm noting it here so that if we do change it we'll know about it for monitoring.

packages/fxa-auth-server/config/index.ts
biancadanforth: Yeah I [recommended changing these values](https://docs.google.com/document/d/1lXa6ZwxtyXbzB-YQOr0HuRRwtKzqE2etiWUwEN9IRok/edit?disco=AAAAPRMDFnc) as a short-term fix when I was poking around for #10527 .

packages/fxa-auth-server/lib/senders/email.js
biancadanforth: What do you mean by making it its own distinct log entry?

vbudhram (approved):
@dannycoates I gave this a test locally and found that this line needs to pass `row` https://github.com/mozilla/fxa/blob/a561ae1f32344a93c4d6ef68a1427fa472a89b08/packages/fxa-shared/db/models/auth/email-bounce.ts#L62. After that it worked like a charm with manually entered bounces.

Awesome work!

packages/fxa-auth-server/config/index.ts
biancadanforth: I didn't have a specific suggestion, as I don't understand SES' deliverability score too well and how lowering the threshold (especially substantially) might affect it. If you do lower the threshold, I would let Janet/Anna/Chelsea know, as it may substantively impact (hopefully improve?) our bounce rates or other email delivery metrics.

packages/fxa-auth-server/lib/senders/email.js
dannycoates: as in with it's own "name" that we can query like `email.bounce` instead of `mailer.send.error`

biancadanforth:
@dannycoates Are you going to merge this today before we tag? If so, I'll wait for it to merge, so I can rebase my PR off of it for the `template` property in the `log.error`.

dannycoates:
> @dannycoates Are you going to merge this today before we tag? If so, I'll wait for it to merge, so I can rebase my PR off of it for the `template` property in the `log.error`.

I'd like to but go ahead and merge yours first

PR #10543, opened 2021-09-28
chore(auth): add email template name to log.error if sending email fails

biancadanforth:
Because:

* We want to know when a user is blocked from logging in because they can't receive FxA emails.

This commit:

* Records the email template name when we fail to send an email, so we can filter for the emails which must be opened before the user can sign in (enumerated in #10527).

Closes: #10527

## Checklist

I am confirming with @jbuck that we can pull in log data from the auth server into Grafana, as ultimately, this data needs to go into a new panel. Otherwise, I will resubmit this PR using a StatsD metric.


vbudhram (approved):
@biancadanforth LGTM, fingers crossed no merge conflict with danny.

PR #10558, opened 2021-09-29
feat(services): remove fxa-auth-db-mysql, fxa-email-event-proxy, and fxa-email-service

dannycoates:
RIP


.circleci/README.md
```diff
@@ -2,13 +2,6 @@
 
 `config.yml` defines the jobs and workflows of our CircleCI deployment.
 
-## Special Cases
```
vbudhram, line 5: Couple more references to email service

https://github.com/mozilla/fxa/blob/ad80dab69ab1f5101afde4487b1ec8921d274a92/packages/fxa-auth-server/config/index.ts#L700

https://github.com/mozilla/fxa/blob/94ffcb4b3d8ccb6f853ca61716999834129ceda2/packages/fxa-auth-server/lib/routes/validators.js#L152

Issue #10575, opened 2021-09-30
"Password changed successfully" email text issue


data-sync-user:
**Build:**
Stage (Train 1.216.0)

**Affected platforms:**
Windows 10 64-bit

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net and login using an existing account.
1. Go to “Password” section and click “Change” button.
1. Enter valid passwords in all 3 fields and click "Save" button.
1. Go to inbox and verify the "Password updated" email.
1. Observe the text.

**Actual result:** 
Email title issue.

**Expected result:**  
Email title is correctly written.

**Note:**
Another issue is related to whitespace being put before the "." (dot)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4045)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16674/pwd+updated+email+issue.PNG">pwd updated email issue.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/16673/pwd+updated+email+-+whitespace+before+dot.PNG">pwd updated email - whitespace before dot.PNG</a>

PR #10576, opened 2021-09-30
fix(auth): wrap span in '<' in the reset password email 

annasob:
## Because

- we are seeing html tags in the reset password email header

## This pull request

- adds '<' to the html tag so it is no longer rendered as test

Closes: #10575

## Checklist

## Screenshots (Optional)


LZoog (approved):
Woo thanks! :)

LZoog:
@annasob ah you could change your commit message to begin with `fix(auth)` or `fix(auth-server)` since "settings" implies the fix is in `fxa-settings`.

PR #10578, opened 2021-10-01
feat(payments): add email address to download link

### LINKED ISSUE #10489: [subplat] Append User Email Address as Query Param to Product Download URL on Successful Subscription
It might be helpful to append the user's email address to the product download URL as a query param on the successful subscription confirmation page, especially for RPs with prompt=none enabled.  The post-finish-setup redirect for the passwordless subscription is already doing this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4005)

IvoJP:
Because:

* MDN requested to have the email address appeneded to the download link as a query param

This commit:

* URI encodes the email address of the user and appends to the download link on a successful subscription

Closes #10489

## Checklist

### DIFF
--- a/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
@@ -60,7 +60,10 @@ export const PaymentConfirmation = ({
   );
 
   const { l10n } = useLocalization();
-  const downloadLink = productUrl + `/?email=${encodeURIComponent(email)}`;
+  const downloadLink: string =
+    productUrl.indexOf('?') > -1
+      ? productUrl + `&email=${encodeURIComponent(email)}`
+      : productUrl + `?email=${encodeURIComponent(email)}`;
 
   return (
     <>
--- a/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
@@ -60,10 +60,8 @@ export const PaymentConfirmation = ({
   );
 
   const { l10n } = useLocalization();
-  const downloadLink: string =
-    productUrl.indexOf('?') > -1
-      ? productUrl + `&email=${encodeURIComponent(email)}`
-      : productUrl + `?email=${encodeURIComponent(email)}`;
+  const downloadUrl: URL = new URL(productUrl);
+  downloadUrl.searchParams.append('email', email);
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -146,7 +146,7 @@ export const PaymentConfirmation = ({
             <a
               data-testid="download-link"
               className="button download-link"
-              href={productUrl}
+              href={productUrl + `/?email=${encodeURIComponent(email)}`}
```
dannycoates, line 149: nit: I'd suggest not mixing concatenation and template strings /driveby

IvoJP: moved to its own variable outside of template

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -60,6 +60,7 @@ export const PaymentConfirmation = ({
   );
 
   const { l10n } = useLocalization();
+  const downloadLink = productUrl + `/?email=${encodeURIComponent(email)}`;
```
chenba, line 63: The productUrl is a url from the metadata, which could already have query params.  We need to accommodate that.

chenba (changes requested):
We need to make it work with any url.  See comment.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
IvoJP: updated to handle either case - query params already present or not.

packages/fxa-payments-server/src/components/PaymentConfirmation/index.tsx
```diff
@@ -60,6 +60,10 @@ export const PaymentConfirmation = ({
   );
 
   const { l10n } = useLocalization();
+  const downloadLink: string =
+    productUrl.indexOf('?') > -1
+      ? productUrl + `&email=${encodeURIComponent(email)}`
+      : productUrl + `?email=${encodeURIComponent(email)}`;
```
chenba, line 66: I think this probably would work for most cases.  But a safer way to do this is to use the URL API: https://developer.mozilla.org/en-US/docs/Web/API/URL.  Once the url's parsed you can add an entry into its search params (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

IvoJP: updated to use URL API

chenba (approved):
👍

PR #10588, opened 2021-10-01
feat(auth): store stripe docs in firestore

### LINKED ISSUE #10530: Update expandResource/webhooks to integrate Firestore fetch/store
Before we drop expandResource entirely in favor of passing all the stripe objects into the functions so they don't possibly trigger additional fetches on their own, we should update expandResource to utilize Firestore caching of the object, and have it populate Firestore if it wasn't already present there. This can reduce a large quantity of our in-advertant Stripe calls.

To ensure the firestore is kept up to date, we should insert/update Firestore on stripe webhooks as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4026)

bbangert:
Because:

* We want to keep firestore copies of Stripe docs up to date.
* We want to use our firestore copy of Stripe data instead of
  hitting Stripe frequently.

This commit:

* Uses the stripe webhook to populate Firestore with up-to-date
  objects from Stripe.
* Uses Firestore first when possible to fetch Stripe documents
  when expanding the resourcce.

Closes #10530

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe-firestore.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe-firestore.ts
@@ -12,7 +12,7 @@ export enum FirestoreStripeError {
   STRIPE_CUSTOMER_DELETED = 'StripeCustomerDeleted',
 }
 
-function newFirestoreStripeError(
+export function newFirestoreStripeError(
   message: string,
   code: FirestoreStripeError
 ): Error {
@@ -22,11 +22,53 @@ function newFirestoreStripeError(
 }
 
 export class StripeFirestore {
+  private subscriptionCollection: string;
+  private invoiceCollection: string;
+
   constructor(
     private firestore: Firestore,
     private customerCollectionDbRef: CollectionReference,
-    private stripe: Stripe
-  ) {}
+    private stripe: Stripe,
+    prefix: string
+  ) {
+    this.subscriptionCollection = `${prefix}subscriptions`;
+    this.invoiceCollection = `${prefix}invoices`;
+  }
+
+  /**
+   * Retrieve a customer from Stripe and insert it if it doesn't exist.
+   */
+  async retrieveAndFetchCustomer(customerId: string) {
+    try {
+      const customer = await this.retrieveCustomer({ customerId });
+      return customer;
+    } catch (err) {
+      if (err.name === FirestoreStripeError.FIRESTORE_CUSTOMER_NOT_FOUND) {
+        return this.fetchAndInsertCustomer(customerId);
+      }
+      throw err;
+    }
+  }
+
+  /**
+   * Retrieve a subscription from Stripe and populate the Firestore record for
+   * the customer and subscription if it doesn't exist.
+   */
+  async retrieveAndFetchSubscription(subscriptionId: string) {
+    try {
+      const subscription = await this.retrieveSubscription(subscriptionId);
+      return subscription;
+    } catch (err) {
+      if (err.name === FirestoreStripeError.FIRESTORE_SUBSCRIPTION_NOT_FOUND) {
+        const subscription = await this.stripe.subscriptions.retrieve(
+          subscriptionId
+        );
+        await this.fetchAndInsertCustomer(subscription.customer as string);
+        return subscription;
+      }
+      throw err;
+    }
+  }
 
   /**
    * Get a Stripe customer by id, and insert it into Firestore keyed to the fxa uid.
@@ -35,17 +77,18 @@ export class StripeFirestore {
    * loads the customers subscriptions into Firestore.
    */
   async fetchAndInsertCustomer(customerId: string) {
-    const customer = await this.stripe.customers.retrieve(customerId, {
-      expand: ['subscriptions'],
-    });
+    const [customer, subscriptions] = await Promise.all([
+      this.stripe.customers.retrieve(customerId),
+      this.stripe.subscriptions
+        .list({ customer: customerId })
+        .autoPagingToArray({ limit: 100 }),
+    ]);
     if (customer.deleted) {
       throw newFirestoreStripeError(
         `Customer ${customerId} was deleted`,
         FirestoreStripeError.STRIPE_CUSTOMER_DELETED
       );
     }
-    // Drop Stripe.Response attribute from API call.
-    delete (customer as any).lastResponse;
 
     const uid = customer.metadata.userid;
     if (!uid) {
@@ -56,12 +99,12 @@ export class StripeFirestore {
     }
 
     const inserts = [this.insertCustomerRecord(uid, customer)];
-    if (customer.subscriptions) {
-      for (const subscription of customer.subscriptions.data) {
+    if (subscriptions) {
+      for (const subscription of subscriptions) {
         inserts.push(
           this.customerCollectionDbRef
             .doc(uid)
-            .collection('subscriptions')
+            .collection(this.subscriptionCollection)
             .doc(subscription.id)
             .set(subscription, { merge: true })
         );
@@ -93,7 +136,7 @@ export class StripeFirestore {
     }
 
     return customerSnap.docs[0].ref
-      .collection('subscriptions')
+      .collection(this.subscriptionCollection)
       .doc(subscription.id)
       .set(subscription, { merge: true });
   }
@@ -117,9 +160,9 @@ export class StripeFirestore {
     }
 
     return customerSnap.docs[0].ref
-      .collection('subscriptions')
+      .collection(this.subscriptionCollection)
       .doc(invoice.subscription)
-      .collection('invoices')
+      .collection(this.invoiceCollection)
       .doc(invoice.id)
       .set(invoice, { merge: true });
   }
@@ -129,8 +172,8 @@ export class StripeFirestore {
    */
   async retrieveCustomer(
     options:
-      | { uid: string; customerId: undefined }
-      | { uid: undefined; customerId: string }
+      | { uid: string; customerId?: undefined }
+      | { uid?: undefined; customerId: string }
   ) {
     if (options.uid) {
       const customerSnap = await this.customerCollectionDbRef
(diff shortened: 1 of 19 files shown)


packages/fxa-auth-server/lib/payments/stripe-firestore.ts
```diff
@@ -117,9 +160,9 @@ export class StripeFirestore {
     }
 
     return customerSnap.docs[0].ref
-      .collection('subscriptions')
+      .collection(this.subscriptionCollection)
       .doc(invoice.subscription)
-      .collection('invoices')
+      .collection(this.invoiceCollection)
```
biancadanforth, line 165: So a customer document in Firestore will have multiple sub documents (called collections) here for the subscriptions and invoices for the customer?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -108,8 +109,8 @@ export class StripeHelper {
   private redis: ioredis.Redis | undefined;
   private statsd: StatsD;
   private taxIds: { [key: string]: string };
-  private firestore: Firestore;
-  private stripeFirestore: StripeFirestore;
+  private firestore?: Firestore;
+  private stripeFirestore?: StripeFirestore;
```
biancadanforth, line 113: Why are these optional? I guess we do have a `config.authFirestore.enabled` flag; is this temporary until we make sure it all works?

What's the difference between `config.firestore` and `config.authFirestore` in `fxa-auth-server/config/index.ts`?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2288,9 +2291,34 @@ export class StripeHelper {
...
+            await this.stripeFirestore.insertInvoiceRecord(invoice);
+            // @ts-ignore
+            return invoice;
+          }
+          throw err;
+        }
+      }
+    }
+
     if (resourceType === CUSTOMER_RESOURCE) {
```
biancadanforth, line 2319: I guess once we enable StripeFirestore permanently, we can get rid of the rest of the code from here down...

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2299,6 +2327,68 @@ export class StripeHelper {
     return this.stripe[resourceType].retrieve(resource);
   }
 
+  async processWebhookEventToFirestore(event: Stripe.Event) {
+    const { type, data } = event;
+
+    if (!this.stripeFirestore) {
+      return;
+    }
```
biancadanforth, line 2335: Nit: Why not put this before `const { type, data } = event;`?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2299,6 +2327,68 @@ export class StripeHelper {
...
+
+    // Note that we must insert before any event handled by the general
+    // webhook code to ensure the object is up to date in Firestore before
+    // our code handles the event.
```
biancadanforth, line 2339: 👍

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2299,6 +2327,68 @@ export class StripeHelper {
...
+        case 'customer.deleted':
+          const customer = data.object as Stripe.Customer;
+          // Ensure the customer and its subscriptions exist in Firestore.
+          // Note that we still insert the object here in case we've already
+          // fetched the customer previously.
+          await this.stripeFirestore.retrieveAndFetchCustomer(customer.id);
+          await this.stripeFirestore.insertCustomerRecord(
+            customer.metadata.userid,
+            customer
+          );
```
biancadanforth, line 2367: This is to handle `customer.updated` and `customer.deleted`, in which case we already have a customer record (that is now stale) that would be returned by `retrieveAndFetchCustomer` without updating the customer in Firestore?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2299,6 +2327,68 @@ export class StripeHelper {
...
+        case 'invoice.payment_failed':
+        case 'invoice.updated':
+        case 'invoice.deleted':
+          const invoice = data.object as Stripe.Invoice;
+          await this.stripeFirestore.retrieveAndFetchSubscription(
+            invoice.subscription as string
+          );
+          // Now that the subscription is fetched, we can safely insert the
+          // invoice record.
+          await this.stripeFirestore.insertInvoiceRecord(invoice);
```
biancadanforth, line 2354: Why are we doing this `retrieveAndFetchSubscription`?

This method will add the customer record and the subscription in Firestore if it doesn't exist, so I guess we're making sure there's a customer object with the associated subscription object loaded first, since the invoice collection is nested under the subscription collection (which is nested under customer)?

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -303,7 +309,7 @@ export class StripeWebhookHandler extends StripeHandler {
       );
       return;
     }
-    this.stripeHelper.refreshCachedCustomer(uid, account.email);
+    this.stripeHelper.removeCustomerFromCache(uid, account.email);
```
biancadanforth, line 312: What is this about? Are there other places where we are still pulling a customer record from Redis? Will this cause those places to break?

biancadanforth (commented):
Hi Ben; I know this is a draft, but I took a look and added a bunch of questions in line for my own understanding. In general I mostly see what you're up to, heh.

packages/fxa-auth-server/lib/payments/stripe-firestore.ts
bbangert: Their own sub collections, yes.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: I'm not entirely sure that the local firestore emulator provided by Google is capable of doing collectionGroup queries. I've had a tough time using some features with it, since it also doesn't run its own web interface like the production system. So having this optional allows us to turn it off for local development so that it isn't an additional hassle for every developer.

bbangert: Regarding `config.firestore` vs `config.authFirestore`, we previously had scripts in auth-server that needed to talk to the event-broker firestore config to load user login data up. The scripts read `config.firestore` which is the event-broker firestore config. I prefixed this to designate this as a different firestore config.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Depends on if the hassle of getting a local firestore working with this correctly outweighs the code penalty.

bbangert: A note, it looks like the local emulator does support the features, I'll add an issue to setup authFirestore to use the local emulator when in local dev mode.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: We want the parent path to exist first before inserting invoices. This is also key to ensuring we back-fill active customer data pushes.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: When processing these events, the data is not used again right here, and in fact, might not be used for hours/days afterwards. But refresh forces a new copy of the customer record into Redis, even though it might not be used, why? So instead this merely removes the record from Redis as we know there's a been a change invalidating it, and if its actually needed later, a new fetch will be done on-demand.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Given that we are not immediately back-filling Firestore, its quite likely we can get deleted events for people we have never inserted. This will cause many of these events to fail as they can't retrieve sufficient customer information if Stripe has already deleted it all. That's why this entire block is wrapped in a try/except to catch that scenario.

biancadanforth: Okay thanks; sorry my comment did not indicate the right set of lines here, as I was asking about calling `retrieveAndFetchCustomer` before calling `insertCustomerRecord`.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2288,9 +2291,34 @@ export class StripeHelper {
...
+        // @ts-ignore
+        return this.stripeFirestore.retrieveAndFetchSubscription(resource);
+      } else if (resourceType === INVOICES_RESOURCE) {
+        try {
+          const invoice = await this.stripeFirestore.retrieveInvoice(resource);
+          // @ts-ignore
+          return invoice;
+        } catch (err) {
+          if (err.name === FirestoreStripeError.FIRESTORE_INVOICE_NOT_FOUND) {
+            await this.stripeFirestore.retrieveAndFetchCustomer(resource);
```
chenba, line 2308: `resource` here is not a customer id.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2288,9 +2291,34 @@ export class StripeHelper {
     // subscriptions for the customer.  The Stripe API stopped including
     // subscriptions for a customer in version 2020-08-27; this ensures
     // backwards compatibility for our code that's relying on that behavior.
+    if (this.stripeFirestore) {
+      if (resourceType === CUSTOMER_RESOURCE) {
```
chenba, line 2295: nit: `switch`?

chenba (approved):

Issue #10535, opened 2021-09-27
Use fxa uid/email lookup instead of loading Stripe customer for customer id


data-sync-user:
In webhook handling we frequently have the customer id, but need the fxa uid/email. These places should be updated to use the new db function to locate the fxa uid/email instead of relying on fetching the customer record and getting the metadata out.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4030)

PR #10599, opened 2021-10-04
feat(db): update db and lookup method to locate FxA user by Stripe customer id

### LINKED ISSUE #10529: Update database and lookup method to locate FxA user by Stripe customer id
To reduce API calls to stripe, we should add an index to the customer id field for account-customers db table so that we can lookup the fxa uid/email based on stripe customer id. This would let us avoid Stripe or Firestore hits to get the fxa information based on a stripe customer id.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4025)

biancadanforth:
Because:

* We want to reduce API calls to Stripe. If we have the customer id, that's sufficient to get the uid and email from our accountCustomers and accounts tables, respectively, instead.

This commit:

* Adds an index to the stripeCustomerId field in the accountCustomers table
* Adds a new getUidAndEmailByStripeCustomerId helper

Closes #10529

## Checklist

## Other
#10535 handles using this new helper in various Stripe webhook handlers.


packages/fxa-shared/test/db/models/auth/index.spec.ts
```diff
@@ -57,6 +58,24 @@ describe('auth', () => {
     });
   });
 
+  describe('getUidAndEmailByStripeCustomerId', () => {
```
biancadanforth, line 61: Since the new helper is a join on accounts and accountCustomers, I made a new top level describe block in this test. Is this the best place for it?

packages/fxa-shared/db/models/auth/index.ts
```diff
@@ -78,6 +78,26 @@ export async function getAccountCustomerByUid(uid: string) {
...
+  const accounts = Account.tableName;
+  const accountCustomers = AccountCustomers.tableName;
+  return Account.query()
+    .select(`${accounts}.uid`, `${accounts}.email`)
+    .join(`${accountCustomers}`, `${accountCustomers}.uid`, `${accounts}.uid`)
+    .where({
+      [`${accountCustomers}.stripeCustomerId`]: customerId,
+    })
+    .limit(1)
+    .first();
```
chenba, line 98: Let's declare a return type for `getUidAndEmailByStripeCustomerId`.  The inferred type is Account, which isn't correct.

chenba (approved):
Looks good.  But let's explicitly declare a return type for the new function.  The returned value isn't fully an Account, which is the inferred type.

PR #10603, opened 2021-10-05
docs(db): update root README on how to connect to local FxA MySQL DBs

biancadanforth:
Because:

* We should document the basic process for new contributors.

This commit:

* Adds a new "Connecting to a local MySQL DB" section to the root project README.

Closes #No issue

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ The Firefox Accounts (fxa) monorepo
 [Node debugging](#node-debugging)\
 [Android debugging](#android-debugging)\
 [FxA Email Service](#fxa-email-service)\
+[Connecting to a local MySQL DB](#connecting-to-a-local-mysql-db)\
 [Firefox for iOS](#firefox-for-ios)\
 [Running with MailDev](#running-with-maildev)\
 [Documentation](#documentation)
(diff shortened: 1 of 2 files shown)


dannycoates:
Another method is if you have the mysql-client installed locally, (mine is in `/usr/local/opt/mysql-client/bin` via `brew install mysql-client`):

```
/usr/local/opt/mysql-client/bin/mysql -uroot --host=127.0.0.1 --port=3306
```

dannycoates:
as a one-liner we could put it in the root package.json scripts section

biancadanforth:
> Could combine those steps into `docker exec -it "$(docker container ls | grep mysql | cut -d' ' -f1)" mysql`

I thought about merging it all into one command, but I also wondered if that might be a little confusing for newer contributors. I'll merge it just to keep the section a little shorter.

biancadanforth:
> as a one-liner we could put it in the root package.json scripts section

Unfortunately with all the quotes in the command and the fact that the npm run script has to be wrapped in a set of quotes, I don't think it will work -- I'm getting:

```sh
bdanforth ~/Projects/fxa (update-readme-mysql-db) $ yarn run mysql
Error: No such container: $(docker container ls | grep mysql | cut -d
```

PR #10634, opened 2021-10-06
fix(payments): check if email exists on blur only

### LINKED ISSUE #10450: ”You already have an account. Sign in” error message does not disappear after editing the email address
Affected versions:

All;

Affected FxA versions:

1.213.5;

Prerequisites:

Have one FxA account without subscription;

Be in a country where VPN is available;

Be on new “Set up your subscription” page;

Steps to reproduce:
1. Start inserting an existing email address into “Enter your email” field and accept the autofill suggestion;
2. Once the ”You already have an account. Sign in” error message appear, edit the email address so that it does not match any existing FxA account email address ;

Expected result:

The error message disappears, as the email inserted is no longer associated with an FxA account;

Actual result:

The error message is stuck on the screen, although an unused email address is inserted;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3994)

IvoJP:
Because:

* The debounce function from lodash returns the value from the last invocation, it results in the incorrect behavior (showing/hiding the error message) after the field enters the error state.

This commit:

* Switches the validation behavior to only check if the email exists on blur and removes the lodash debounce function.

Closes #10450

## Checklist

lodash docs for debounce: https://lodash.com/docs/4.17.15#debounce


chenba (approved):
LGTM.  And if you haven't done so, maybe leave a comment for QA in the jira about this change so they won't file bugs about existing account check onchange is borked.

PR #10639, opened 2021-10-06
refactor(build): use esbuild instead of ts-node

dannycoates:
We got decent perf gains on auth-server so I'd like to standardize the rest of the packages on esbuild


packages/fxa-shared/test/metrics/navigation-timing.ts
```diff
@@ -3,25 +3,14 @@
...
 import sinon from 'sinon';
+import { observeNavigationTiming } from '../../metrics/navigation-timing';
 
 type MockEntryList = { getEntries: () => [object] };
 type ObsCallback = (_entries: MockEntryList, _obs: object) => undefined;
-type sinonStub = ReturnType<typeof sinon.stub>;
 let cb = (_entries: MockEntryList, _obs: object) => {};
 
 const sandbox = sinon.createSandbox();
-(global.navigator as unknown) = { sendBeacon: sandbox.stub() };
```
dannycoates, line 13: `navigator.sendBeacon` is the only hangup. it seems esbuild-register handles import caching differently enough from ts-node that this stub doesn't apply to the navigator in navigation-timing.ts. `sendFn` is simple enough that I don't think we need to assert the arguments passed to sendBeacon.

packages/fxa-shared/metrics/navigation-timing.ts
```diff
@@ -15,18 +15,6 @@ const NAV_ENTRY_TYPE = 'navigation';
 
 // export for testing
 export const sendFn = () => {
-  if ('keepalive' in new Request('')) {
```
dannycoates, line 18: esbuild got hungup on where `Request` was defined. Since firefox doesn't support `keepalive` I think we can just remove this block.

.circleci/config.yml
```diff
@@ -129,7 +129,6 @@ jobs:
             'fxa-event-broker' \
             'fxa-auth-db-mysql' \
             'fxa-profile-server' \
-            'fxa-metrics-processor' \
```
dannycoates, line 132: im removing this package in another PR. no need to test it

dannycoates (commented):
Maybe a little faster, but mostly just to standardize to follow auth-server

Issue #10492, opened 2021-09-21
Add UX changes for signup/signin


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4008)

PR #10646, opened 2021-10-07
feat(content-server): conditionally render enhanced auth ui for pocket clients

jodyheavener:
## Because

- We want to allow Pocket to tap into a special auth UI for migrating their users to FxA

## This pull request

- Conditionally renders special UI during sign up and sign in auth flows when a pocket client ID is present

Closes: #10492

## Checklist

## Screenshots (Optional)

Screen Shot 2021-10-07 at 12 10 57 PM
Screen Shot 2021-10-07 at 12 11 14 PM
Screen Shot 2021-10-07 at 12 11 40 PM


packages/fxa-content-server/app/scripts/templates/partial/firefox-family-services.mustache
```diff
@@ -5,10 +5,18 @@
...
       {{#t}}Sync your bookmarks, passwords, and more on every device.{{/t}}
     </li>
-    <li class="firefox-service service-lockwise">
-      <span class="logo logo-lockwise">{{#t}}Firefox Lockwise{{/t}}</span>
-      {{#unsafeTranslate}}Securely access passwords saved in Firefox anywhere &mdash; even outside of the browser.{{/unsafeTranslate}}
-    </li>
+    {{#isInPocketMigration}}
+      <li class="firefox-service service-pocket">
+        <span class="logo logo-pocket">{{#t}}Pocket{{/t}}</span>
+        {{#unsafeTranslate}}Save and discover the best stories from accross the web.{{/unsafeTranslate}}
```
vbudhram, line 11: I don't think this needs to be in an unsafeTranslate since it doesn't contain html.

packages/fxa-content-server/app/scripts/templates/index.mustache
```diff
@@ -1,15 +1,17 @@
 <div id="main-content" class="card">
   <header>
     <h1 id="fxa-enter-email-header">
-      <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(serviceName)s" -->
       {{#t}}Enter your email{{/t}}
+      <span class="service">
+        <!-- L10N: For languages structured like English, the phrase can read "to continue to %(serviceName)s" -->
         {{#isInPocketMigration}}
-            <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
+          {{#t}}Continue to{{/t}}
```
vbudhram, line 8: I think we could wrap this one in an unsafeTranslate since it might help give translators more context that an image is after the text.

packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
```diff
@@ -1,8 +1,17 @@
...
+        <span class="service">
+          <!-- L10N: For languages structured like English, the phrase can read "to continue to %(serviceName)s" -->
+          {{#isInPocketMigration}}
+            {{#t}}Continue to{{/t}}
+            <img id="graphic-client-pocket" src="/images/pocket_logo_rounded.svg" alt="Pocket" />
+          {{/isInPocketMigration}}
+          {{^isInPocketMigration}}
+            {{#t}}Continue to %(serviceName)s{{/t}}
+          {{/isInPocketMigration}}
+        </span>
```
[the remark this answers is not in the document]
vbudhram: I'm going to lean on not partializing this for now since these pages will eventually be moved to React and what we have covers the majority of users.

PR #10662, opened 2021-10-11
feat(auth): use getUidAndEmailByStripeCustomerId helper in Stripe webhook handlers

### LINKED ISSUE #10535: Use fxa uid/email lookup instead of loading Stripe customer for customer id

biancadanforth:
Because:

* getUidAndEmailByStripeCustomerId was added in #10529, but it isn't used anywhere yet.

This commit:

* Modifies the function signature of the helper, so that it can be used as a drop-in replacement for getCustomerUidEmailFromSubscription.
* Removes all references to getCustomerUidEmailFromSubscription.
* Uses this helper in various locations where we were previously expanding the Stripe customer object just to get the uid and email.

Closes #10535

## Checklist

### DIFF
--- a/packages/fxa-shared/db/models/auth/index.ts
+++ b/packages/fxa-shared/db/models/auth/index.ts
@@ -83,20 +83,15 @@ export async function getAccountCustomerByUid(uid: string) {
  * Fetch the FxA user id and primary email for a customer by Stripe customer id.
  */
 export async function getUidAndEmailByStripeCustomerId(
-  customer: string | Stripe.Customer | Stripe.DeletedCustomer | null
+  customerId: string
 ): Promise<{ uid: string | null; email: string | null }> {
-  if (typeof customer !== 'string') {
-    throw new Error(
-      `Argument is not a string for customer with id ${customer?.id}`
-    );
-  }
   const accounts = Account.tableName;
   const accountCustomers = AccountCustomers.tableName;
   const result = await Account.query()
     .select(`${accounts}.uid`, `${accounts}.email`)
     .join(`${accountCustomers}`, `${accountCustomers}.uid`, `${accounts}.uid`)
     .where({
-      [`${accountCustomers}.stripeCustomerId`]: customer,
+      [`${accountCustomers}.stripeCustomerId`]: customerId,
     })
     .limit(1)
     .first();
--- a/packages/fxa-auth-server/lib/payments/capability.ts
+++ b/packages/fxa-auth-server/lib/payments/capability.ts
@@ -13,6 +13,7 @@ import { AuthLogger, AuthRequest, ProfileClient } from '../types';
 import { PlayBilling } from './google-play/play-billing';
 import { SubscriptionPurchase } from './google-play/subscription-purchase';
 import { StripeHelper } from './stripe';
+import error from '../error';
 
 function hex(blob: Buffer | string): string {
   if (Buffer.isBuffer(blob)) {
@@ -94,6 +95,15 @@ export class CapabilityService {
     uid?: string | null;
     email?: string | null;
   }) {
+    if (typeof sub.customer !== 'string') {
+      throw error.internalValidationError(
+        'stripeUpdate',
+        {
+          subscriptionId: sub.id,
+        },
+        'Subscription customer was not a string.'
+      );
+    }
     if (!uid || !email) {
       ({ uid, email } = await getUidAndEmailByStripeCustomerId(sub.customer));
     }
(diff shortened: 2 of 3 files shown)


packages/fxa-shared/db/models/auth/index.ts
```diff
@@ -80,24 +82,29 @@ export async function getAccountCustomerByUid(uid: string) {
...
- * @param customerId
- * @returns {Object} result
- * @returns {string} result.uid - The FxA user id
- * @returns {string} result.email - The primary email for the FxA user
  */
 export async function getUidAndEmailByStripeCustomerId(
-  customerId: string
-): Promise<{ uid: string; email: string }> {
+  customerId: string | Stripe.Customer | Stripe.DeletedCustomer | null
+): Promise<{ uid: string | null; email: string | null }> {
```
biancadanforth, line 88: I modified the function signature of `getUidAndEmailByStripeCustomerId` to take a `customer` and do some type checking to make sure it’s a string and throw an error if not.

One possible issue is that I am importing an `error` object from the fxa-auth-server in my new check that `customer` is a string. Is this an issue?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1788,6 +1742,27 @@ export class StripeHelper {
...
+        'extractInvoiceDetailsForEmail',
+        { customerId: invoice.customer },
+        'Customer missing email or uid.'
+      );
+    }
+
+    const customer = await this.customer({
+      uid,
+      email,
+    });
```
biancadanforth, line 1761: Since I now have the uid and email from my new helper, I can call `this.customer` from the Stripe Helper which at least has a chance to pull from the Redis cache before fetching a customer from Stripe (unlike `expandResouce`, which was used before).

biancadanforth:
> The code looks fine and it make sense. But I know that @bbangert is also working on patch to completely remove the direct Stripe fetches. `getUidAndEmailByStripeCustomerId` is always followed by a customer get, which, with Ben's patch, would hit Firestore on a cache miss. And, I think, `getUidAndEmailByStripeCustomerId` failing to find the uid and email should be exceptions, so I don't think it prevents that many customer gets.
> 
> @bbangert I remember you wanting this but does it still make sense now with the Firestore caching?

Thanks Barry for taking a look; yeah I agree there's quite a bit of overlap with PR #10676 .

I talked with you and Ben about this earlier sync, but just to note it here, I will wait for that PR to close, and I will only add this new helper where we can avoid the additional call to fetch the customer entirely. That would just be the change to `lib/payments/capability.ts` in this PR currently.

packages/fxa-shared/db/models/auth/index.ts
```diff
@@ -80,24 +81,29 @@ export async function getAccountCustomerByUid(uid: string) {
...
- * @returns {Object} result
- * @returns {string} result.uid - The FxA user id
- * @returns {string} result.email - The primary email for the FxA user
  */
 export async function getUidAndEmailByStripeCustomerId(
-  customerId: string
-): Promise<{ uid: string; email: string }> {
+  customer: string | Stripe.Customer | Stripe.DeletedCustomer | null
+): Promise<{ uid: string | null; email: string | null }> {
+  if (typeof customer !== 'string') {
```
chenba, line 88: Can we not accept non-strings in the parameter?  Might be confusing to people using this function when they pass in one of the accepted types only to get an error.

chenba (approved):
LGTM but got a suggestion on restricting the function parameter type

packages/fxa-shared/db/models/auth/index.ts
biancadanforth: That's what I originally had, but the consumer (now just `stripeUpdate` in `lib/payments/capability.ts`) passes in a Stripe `subscription.customer` which can be a Stripe `Customer` or `DeletedCustomer` as well. I was trying to avoid checking the type of the value passed into the helper in every caller (though there's just one caller currently).

Looking at this one caller, `stripeUpdate`, since this only comes from Stripe webhooks where `event.data.object` is a Stripe `subscription` with a non-expanded customer, I guess I could do something like:
```
({ uid, email } = await getUidAndEmailByStripeCustomerId(sub.customer as string));
```
What do you think @chenba ?

PR #10663, opened 2021-10-11
fix(tests): fixed auth-server/scripts tests

dannycoates:


packages/fxa-auth-server/scripts/write-emails-to-disk.js
```diff
@@ -46,37 +46,38 @@ const mailSender = {
...
-  })
-  .then(() => {
-    console.info('done');
-    process.exit(0);
-  })
-  .catch((err) => {
-    console.error(err.stack);
-    process.exit(1);
-  });
+if (require.main === module) {
```
dannycoates, line 49: This is why the mocha process was exiting with a success exit code. It was executing this when the test file `require`ed it. Since it was the mocha process running it `process.exit(0)` was always terminating mocha with an ok exit code.

packages/fxa-auth-server/scripts/bulk-mailer/index.js
```diff
@@ -52,6 +52,7 @@ module.exports = async function (
         console.info(JSON.stringify(msg));
       }
     },
+    debug() {},
```
chenba, line 55: Not sure if this is intentional.

PR #10676, opened 2021-10-12
refactor(auth): use expandResource instead of retrieve

### LINKED ISSUE #10534: Avoid repopulating Stripe customer when unnecessary
We call refreshCachedCustomer in many places where we only need to purge the cache, not necesarilly trigger another cache load of the customer. In those places we should only delete the customer cache from Redis, not immediately re-populate it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4029)

bbangert:
Because:

* We want to avoid hitting Stripe when we can get the object from
  Firestore.

This commit:

* Switches use of direct Stripe API calls to instead use expandResource
  to fetch from Firestore when possible.

Closes #10534

## Checklist


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -911,13 +920,20 @@ export class StripeHelper {
...
-        (sub.latest_invoice as Stripe.Invoice).status === 'open'
+  async hasOpenInvoice(customer: Stripe.Customer) {
+    const activeInvoices = customer.subscriptions?.data
+      .filter((sub) => ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status))
+      .map((sub) => sub.latest_invoice)
+      .filter((invoice) => invoice !== null);
+    if (!activeInvoices) {
+      return false;
+    }
+    const invoices = await Promise.all(
```
chenba, line 931: Preferably we'd return true on the first invoice with an 'open' status instead of fetching all of them.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -911,13 +920,20 @@ export class StripeHelper {
...
-    const subscription = customer.subscriptions?.data.find(
-      (sub) =>
-        ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status) &&
-        (sub.latest_invoice as Stripe.Invoice).status === 'open'
+  async hasOpenInvoice(customer: Stripe.Customer) {
+    const activeInvoices = customer.subscriptions?.data
+      .filter((sub) => ACTIVE_SUBSCRIPTION_STATUSES.includes(sub.status))
+      .map((sub) => sub.latest_invoice)
+      .filter((invoice) => invoice !== null);
+    if (!activeInvoices) {
```
biancadanforth, line 928: I know this has already merged, but I had an in-progress comment here that I'll just ask anyway: if there are no active invoices, isn't `activeInvoices` an empty array? Isn't `[]` truthy?

PR #10694, opened 2021-10-13
fix(ux): Remove newsletters when in pocket migration from signup

vbudhram:
## Because

- Pocket does not want to show the newsletter form when signing up

## This pull request

- Removes it when in pocket migration

Connects to https://github.com/mozilla/fxa/issues/10492

## Checklist

- [] If applicable, I have modified or added tests which pass locally.

## Screenshots (Optional)

Screen Shot 2021-10-13 at 2 20 33 PM

### DIFF
--- a/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
+++ b/packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
@@ -43,17 +43,19 @@
 
       {{^isCWTSOnSignupPasswordEnabled}}
       <section class="get-involved">
-        {{#isAnyNewsletterEnabled}}
-        <header>
-          <h2>{{#t}}Practical knowledge is coming to your inbox. Sign up for more:{{/t}}</h2>
-        </header>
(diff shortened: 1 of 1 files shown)


packages/fxa-content-server/app/scripts/templates/sign_up_password.mustache
```diff
@@ -43,17 +43,20 @@
...
 
-        {{#newsletters}}
-          <div class="input-row marketing-email-optin-row">
-            <input id="{{slug}}" type="checkbox" class="marketing-email-optin" value="{{slug}}"><label for="{{slug}}">{{label}}</label>
-          </div>
-        {{/newsletters}}
+          {{#newsletters}}
+              <div class="input-row marketing-email-optin-row">
+                  <input id="{{slug}}" type="checkbox" class="marketing-email-optin" value="{{slug}}"><label
+                      for="{{slug}}">{{label}}</label>
```
jodyheavener, line 56: seems like these could be 2 space indents, and maybe this tag belongs on the line above?

PR #10713, opened 2021-10-15
feat(auth): require auth firestore usage

### LINKED ISSUE #10710: Remove the firestore optional flag
Update readmes to always enable firstore in local dev environments (remove the "optional" flag for firestore)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4097)

bbangert:
Because:

* Having a toggle to deactivate use of Firestore drastically increases
  code complexity.

This commit:

* Requires Firestore for running auth-server and uses the emulator for
  local running.

Closes #10710

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,11 +9,13 @@ executors:
       - image: memcached
       - image: pafortin/goaws
       - image: circleci/mysql:5.7.27
+      - image: jdlk7/firestore-emulator
         environment:
           NODE_ENV: dev
           FXA_EMAIL_ENV: dev
           FXA_EMAIL_LOG_LEVEL: debug
           RUST_BACKTRACE: 1
+          AUTH_FIRESTORE_EMULATOR_HOST: localhost:9090
     environment:
       FXA_MX_RECORD_EXCLUSIONS: restmail.dev.lcip.org
 
@@ -89,6 +91,11 @@ jobs:
       - image: memcached
       - image: pafortin/goaws
       - image: circleci/mysql:5.7.27
+      - image: jdlk7/firestore-emulator
+        environment:
+          NODE_ENV: dev
+          AUTH_FIRESTORE_EMULATOR_HOST: localhost:9090
+
     parameters:
       package:
         type: string
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -135,8 +135,8 @@ export class StripeHelper {
   private redis: ioredis.Redis | undefined;
   private statsd: StatsD;
   private taxIds: { [key: string]: string };
-  private firestore?: Firestore;
-  private stripeFirestore?: StripeFirestore;
+  private firestore: Firestore;
+  private stripeFirestore: StripeFirestore;
   public currencyHelper: CurrencyHelper;
 
   /**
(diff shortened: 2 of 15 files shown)


packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -4001,142 +4001,97 @@ describe('StripeHelper', () => {
...
+      stripeFirestore.retrieveCustomerSubscriptions = sandbox
+        .stub()
+        .resolves(deepCopy(customer.subscriptions.data));
+      const result = await stripeHelper.expandResource(
+        customer.id,
         CUSTOMER_RESOURCE
       );
-      assert.deepEqual(actual, customer1);
+      // Note that top level will mismatch because subscriptions is copied
+      // without the object type.
```
biancadanforth, line 4022: What do you mean here by object type? Can you clarify why a deepEqual for result and customer won't work?

biancadanforth (commented):
Hey Ben; I'm picking my car up from the shop, so I'll try this when I get back. Left a few small comments/questions.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2338,68 +2336,49 @@ export class StripeHelper {
...
-
-    // We make an exception here for customers because we need to get the
-    // subscriptions for the customer.  The Stripe API stopped including
-    // subscriptions for a customer in version 2020-08-27; this ensures
-    // backwards compatibility for our code that's relying on that behavior.
-    if (resourceType === CUSTOMER_RESOURCE) {
-      // @ts-ignore
-      return this.stripe.customers.retrieve(resource, {
-        expand: [SUBSCRIPTIONS_RESOURCE],
-      });
```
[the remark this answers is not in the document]
bbangert: Correct, I am removing the firestore being optional so the route goes through here now. This is existing code that is now the primary path.

packages/fxa-auth-server/test/local/payments/stripe.js
bbangert: These are existing tests that I haven't changed in this commit. They're just being moved. Object type is a field in the Stripe object being compared.

.circleci/config.yml
```diff
@@ -9,11 +9,13 @@ executors:
       - image: memcached
       - image: pafortin/goaws
       - image: circleci/mysql:5.7.27
+      - image: jdlk7/firestore-emulator
         environment:
           NODE_ENV: dev
           FXA_EMAIL_ENV: dev
           FXA_EMAIL_LOG_LEVEL: debug
           RUST_BACKTRACE: 1
+          AUTH_FIRESTORE_EMULATOR_HOST: localhost:9090
```
biancadanforth, line 18: Attaching this comment here for lack of a better place...

Where can I view the Firestore Emulator UI? I see that the emulator is running on localhost:9090. Per [these instructions](https://firebase.google.com/docs/emulator-suite/connect_and_prototype), the UI is by default at localhost:4000, but I get an “unable to connect” error in Firefox at that port.

Also when I execute an interactive shell in the docker container for the emulator, `which firebase` turns up nothing?

biancadanforth: Ah okay, so how can I verify the data in the Firestore emulator matches my expectations? I'm looking at the Docker image's khttps://github.com/JDLK7/firestore-emulator/blob/master/Dockerfile and at some [docs for the SDK](https://cloud.google.com/sdk/gcloud/reference/beta/emulators/firestore/) being used, and it doesn't look like there is any command than `gcloud beta emulators firestore start` which is the entry point of the docker image...

packages/fxa-auth-server/test/scripts/write-emails-to-disk.js
```diff
@@ -20,6 +20,7 @@ const execOptions = {
   env: {
     NODE_ENV: 'dev',
     LOG_LEVEL: 'error',
+    AUTH_FIRESTORE_EMULATOR_HOST: 'localhost:9090',
```
biancadanforth, line 23: I wonder why this is needed in all these test scripts -- I guess anywhere we're making use of Stripe objects needs to have authFirestore defined now in the auth sever?

PR #10719, opened 2021-10-15
feat(auth-server): convert subscriptionReactivation email template to stack

jodyheavener:
## Because

- We're actively converting old FxA emails to our new modernized stack.

## This pull request

- Converts the `subscriptionReactivation` subscription platform email to the new stack.

Closes: #9288

## Checklist

## Screenshots

Screen Shot 2021-10-15 at 10 38 05 AM

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionReactivation/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionReactivation/index.mjml
@@ -1,0 +1,25 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/senders/emails/templates/subscriptionReactivation/index.mjml
```diff
@@ -0,0 +1,25 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
+<mj-section css-class="mb-6">
+  <mj-column>
+    <mj-image src="<%- icon %>" alt="<%- productName %> icon" title="<%- productName %>" width="58px" height="58px"></mj-image>
```
LZoog, line 7: I think we may want width/height in a class in case we want to use the same in other templates🤔

PR #10724, opened 2021-10-15
feat(dev): add firebase emulator ui visit http://localhost:9299

dannycoates:


_scripts/firestore.sh
```diff
@@ -1,22 +1,8 @@
...
-
-# Create pushbox db on start (because pushbox doesn't create it)
-docker run --rm --name=firestore \
-  -p 9090:9090 \
-  jdlk7/firestore-emulator &
-
-while :; do read -r; done
+# the "demo-" prefix for --project is special
+# see https://firebase.google.com/docs/emulator-suite/connect_firestore#choose_a_firebase_project
+docker run --rm -p 4400:4400 -p 4500:4500 -p 8085:8085 -p 9090:9090 -p 9299:9299 -v "$(pwd)":/home/node --name firebase-tools andreysenov/firebase-tools:latest firebase emulators:start --project demo-fxa
```
dannycoates, line 8: @bbangert I could use a hand with identifying how `projectId` configs should be set in our service configs. Default to `demo-fxa`?

dannycoates: idk if it even matters

bbangert: It probably doesn't, the project id comes out of the account service file. I've played with it locally and it doesn't seem to matter once there's a keyfile that has all the data.

PR #10736, opened 2021-10-18
feat(auth): added metricsOptOut db column to fxa

### LINKED ISSUE #9075: Add opt-out field to account table in database
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-2918)

dannycoates:
closes #9075


packages/fxa-settings/src/components/DataCollection/index.test.tsx
```diff
@@ -5,6 +5,13 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { DataCollection } from '.';
+import { mockAppContext, renderWithRouter } from '../../models/mocks';
+import { Account, AppContext } from '../../models';
+
+const account = {
+  displayName: 'jrgm',
```
LZoog, line 12: Awww

packages/fxa-settings/src/components/DataCollection/index.test.tsx
```diff
@@ -23,5 +30,19 @@ describe('DataCollection', () => {
...
-  // TODO: submission tests with mutations, FXA-4106
+  it('toggles', async () => {
+    const x = renderWithRouter(
+      <AppContext.Provider value={mockAppContext({ account })}>
+        <DataCollection />
+      </AppContext.Provider>
+    );
+
+    const button = x.getByTestId('metrics-opt-out');
+    await button.click();
```
LZoog, line 41: You should be able to take out `await` from L41 and L45.

I think this could use a test for erroring but we can punt that to FXA-4106.

packages/fxa-shared/db/models/auth/account.ts
```diff
@@ -425,4 +427,16 @@ export class Account extends BaseAuthModel {
     }
     return account;
   }
+
+  static async setMetricsOpt(
+    uid: string,
+    state: 'in' | 'out',
```
LZoog, line 433: What's the advantage of using 'in' and 'out' strings rather than a boolean all the way through?

packages/fxa-settings/src/components/DataCollection/index.tsx
```diff
@@ -72,21 +61,19 @@ export const DataCollection = () => {
...
-                    <Localized id="dc-subheader">
-                      Analytics and Improvements
-                    </Localized>
-                  ),
-                }}
-              />
-            </form>
+            <Switch
+              {...{
+                isSubmitting: account.loading,
```
LZoog, line 66: I just realized I picked up this passing in the entire object pattern over setting the individual props from Experimenter and we don't do that as much in Settings 😅

Will r+ after this `loading` bug is addressed. I think I may also change this prop to `disabled` in my PR since that's what we're using for other buttons.

packages/fxa-shared/db/models/auth/account.ts
LZoog: Yeah, I like the descriptiveness, though right the type mixing is why I asked. I think it'd be more clear (for me at least) to make `number | null` a boolean but I won't block r+ on that.

I think having the timestamp in the DB is a good call too.

LZoog (approved):
LGTM 🎉

Issue #10756, opened 2021-10-19
[Pocket migration] Pocket logo missing from "Create a Firefox Account Continue to Pocket" page


data-sync-user:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox

**Steps to reproduce:**

1. Go to Pocket stage server: https://fxa-staging.web.readitlater.com/signup__ffbp=temp.backend.web.fxa_signup
1. Enter email and click "Sign up" button.
1. Enable feature flag by appending "&forceExperiment=pocketMigration&forceExperimentGroup=treatment" to url and press Enter.
1. Observe the "Create a Firefox Account Continue to Pocket" page.

**Actual result:**
Pocket logo is missing from the page.

**Expected result:**
Pocket logo is displayed on the page.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4122)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17003/logo+displayed.PNG">logo displayed.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/16988/pocket+logo+issue.PNG">pocket logo issue.PNG</a>

PR #10759, opened 2021-10-19
fix(content-server): move pocket logo img rendering to css

jodyheavener:
## Because

- We don't appear to swap img `src` values out with CDN URLs
 
## This pull request

- Changes the img for the Pocket logo to a div and moves the src image URL to the background-image in CSS.

Closes: #10756

## Checklist


vbudhram (approved):
@jodyheavener Thank you!

PR #10773, opened 2021-10-19
feat(auth): add reminder emails for accounts created through subscription flow

### LINKED ISSUE #9372: Update set password cron to handle setup required accounts
We have a cron job that sends reminder emails to users to verify their account. If a user is in a 'finish setup' state and has an active subscription, they should be sent another email reminding them to finish setting up their account.

If they do not have an active subscription, no email should be sent at all.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3656)

IvoJP:
Because:
    * We want to remind users to finish setting up their account password when they create an account by purchasing a subscription

This commit:
    * Adds logic to push keys to redis to track which users have not setup their account and adds scripts and templates to send out the appropriate emails

Closes #9372

## Checklist

This touches a few partial templates that are used in other emails, namely the reset password text appears above the footer.

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/account.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/account.ts
@@ -47,13 +47,13 @@ export async function sendFinishSetupEmailForStubAccount({
       token,
     });
 
-
-    console.log("########################");
     if (metricsContext && subscriptionAccountReminders) {
       await subscriptionAccountReminders.create(
         uid,
         metricsContext.flowId,
         metricsContext.flowBeginTime,
+        invoiceDetails.productId,
+        invoiceDetails.productName
       );
     }
   }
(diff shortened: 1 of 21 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/account.ts
```diff
@@ -43,5 +45,16 @@ export async function sendFinishSetupEmailForStubAccount({
       ...metricsContext,
       token,
     });
+
+    if (metricsContext && subscriptionAccountReminders) {
+      await subscriptionAccountReminders.create(
```
vbudhram, line 50: I think MetricsContext should be optional in this case. The email template shouldn't append those values if they are undefined. For some background, metrics context is what we use to follow a user through a signup/signin flow.

PR #10749, opened 2021-10-19
feat(settings): Add telemetry opt-out UI + Switch

LZoog:
## Because

* We want users to have the ability to opt-out of accounts-level telemetry if they wish to

## This pull request

* Adds the initial UI to Settings and Nav behind booleans only enabled in Storybook until the BE is finished
* Creates an a11y and l10n friendly on/off Switch component for this new section
* Updates storybook and tests

## Issue that this pull request solves

closes #9071 
closes #9088

---

This PR needs:
* Double confirmation this looks good from a copy standpoint, I was pinged towards the end of the day by someone legal conferred with and they'd like to review this
* I need check on RTL styles. This might need a _little_ more work on the switch in that regard

^Both of these can potentially be finished in #10732, but I'd ideally like to just get them done here

The new `Switch` is accessible and screenreader friendly. It's also l10n friendly, I tested some long words to make sure the on/off "label" (in quotes because it's not _actually_ the label) accounted for them.

Some screenshots:
image
image
image

German for "disabled":
image

telemetry-opt-out

PR #10786, opened 2021-10-20
feat(settings): DataCollection copy updates & rtl styles

LZoog:
follow ups to #10749

Because:
* Suggested copy was updated since the previous UI PR
* Styling tweaks were needed

This commit:
* Updates copy
* Adds RTL styles to DataCollection and Switch
* Better centers the Switch psuedo el when switch is submitting
* Tweaks Switch placement per UX suggestion

I also took the Switch button out of `<form>`. I previously added it out of habit but it makes the button "submit" by default onclick and we don't have other single-click interactive buttons in form tags. Doesn't affect anything from a screenreader POV either.

RTL support:
image

Opening this for review but ~I'm going to **wait until legal signs off on the new copy** before merging.~ They signed off on it.

### DIFF
--- a/packages/fxa-settings/src/components/DataCollection/index.tsx
+++ b/packages/fxa-settings/src/components/DataCollection/index.tsx
@@ -15,6 +15,10 @@ export const DataCollection = () => {
   const alertBar = useAlertBar();
   const { l10n } = useLocalization();
 
+  const localizedHeader = (
+    <Localized id="dc-heading">Data Collection and Use</Localized>
+  );
+
   const handleMetricsOptOutToggle = useCallback(async () => {
     try {
       setIsSubmitting(true);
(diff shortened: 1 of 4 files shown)


packages/fxa-settings/src/components/DataCollection/index.tsx
```diff
@@ -71,22 +77,16 @@ export const DataCollection = () => {
...
-              />
-            </form>
+          <div className="flex-1 flex justify-center tablet:justify-end tablet:pr-4 tablet:pt-1">
+            <Switch
+              {...{
+                isSubmitting,
+                isOn: metricsOptOut,
+                id: 'metrics-opt-out',
+                handler: handleMetricsOptOutToggle,
+                localizedLabel: localizedHeader,
```
LZoog, line 87: I'm a little torn here on what the best label read for screenreaders is with the new copy. "Analytics and Improvements" was the previous subheader copy which seemed appropriate to use here, but the new subheader is slightly more ambiguous ("Help improve Firefox Accounts"). I went with "Data Collection and Use"  since the user is basically toggling the entire section but it could read a little better IMO. "Allow Firefox Accounts to send technical and interaction data to Mozilla." is totally redundant since it's read just before the user hits the toggle.*

TLDR is this is fine a11y wise, but could possibly be slightly improved 🤷‍♀️ 

*It's possible in the future we should refactor this to use that content as the label, especially if this section gets multiple toggles, but the original copy here was pretty long and was awkward as a label. I don't see us adding more switches to this section any time in the near future and I'm happy with this for now.

dannycoates: if the subheader-content wasn't followed by a link we could have prob used aria-hidden on it and used it in the label, but yeah, it seems ok as it is here.

Issue #10866, opened 2021-10-29
[subplat] Subscription for Mozilla VPN can’t be finalized via PayPal


data-sync-user:
**Affected versions**:

* All;

**Affected FxA versions**:

* 1.218.2;

**Prerequisites**:

* Be in a country where Mozilla VPN is available;
* Be on the New set up your subscription page;

**Steps to reproduce**:
1. Complete all the mandatory fields;
2. Make the subscription for Mozilla VPN via PayPal;
3. Observe the behaviour;

**Expected result**:

* The user is able to successfully finalize the subscription via PayPal;

**Actual result**:

* The user is not able to successfully finalize the subscription via PayPal. “Error confirming subscription..Something went wrong. Please try again later.” error message is displayed on the screen.

**Notes**:

* 【 highlighted #d04437 】This a major issue because the user is not able to make the subscription via PayPal;【 end highlighted 】
* This is reproducible for users with FxA accounts too;
* The subscription can be finalized via Stripe method.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4157)

data-sync-user:
➤ Chelsea Lewis commented:

Marked as Priority: Blocker

data-sync-user:
➤ Madalin Cotetiu commented:

Verified this as fixed on Train 1.218.3 - Staging. The user is able to successfully finalize the subscription via PayPal.

PR #10871, opened 2021-11-01
fix(subscriptions): include empty subscriptions list for customer

chenba:
Because:
 - there's existing code that expect the "expanded" subscriptions list
   on a customer

This commit:
 - include the empty subscriptions list from Firestore on a customer, to
   match the previous behavior of expanding the subscriptions when
   retrieving a customer from Stripe

Closes: #10866


chenba:
cc @StaberindeZA since you ran into this

bbangert (approved):

PR #10872, opened 2021-11-01
feat(auth): add more logging to PayPalClientError

### LINKED ISSUE #8830: PayPalClientError: PayPal NVP returned a non-success ACK. See "this.raw" or "this.data" for more details. PayPal err...
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/11557150/?referrer=github_plugin

```
PayPalClientError: PayPal NVP returned a non-success ACK. See "this.raw" or "this.data" for more details. PayPal error code: 10414
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/paypal-client.js", line 146, in PayPalClient.doRequest
    throw new PayPalClientError(result.text, resultObj);
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/paypal.js", line 131, in async PayPalHelper.chargeCustomer
    const response = await this.client.doReferenceTransaction(doReferenceTransactionOptions);
  ?, in async Promise.all
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/paypal.js", line 285, in async PayPalHelper.processInvoice
    [transactionResponse] = (await Promise.all(promises));
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/paypal-processor.js", line 171, in async PaypalProcessor.makePaymentAttempt
    await this.paypalHelper.processInvoice({
...
(4 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4065)

IvoJP:
Because:

* We need more insight in the logs from the this.raw and this.data properties of the error

This commit:

* Logs the error object when it is thrown

Closes #8830

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-client.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-client.ts
@@ -5,6 +5,7 @@ import pRetry from 'p-retry';
 import querystring from 'querystring';
 import superagent from 'superagent';
 import { EventEmitter } from 'events';
+import { Logger } from 'mozlog';
 
 export const PAYPAL_SANDBOX_BASE = 'https://api-3t.sandbox.paypal.com';
 export const PAYPAL_SANDBOX_IPN_BASE = 'https://ipnpb.sandbox.paypal.com';
@@ -278,12 +279,13 @@ export class PayPalClient {
     factor: number;
   };
   private emitter: EventEmitter;
+  private log: Logger;
   public on: (
     event: 'response',
     listener: (response: ResponseEventType) => void
   ) => EventEmitter;
 
-  constructor(options: PaypalOptions) {
+  constructor(options: PaypalOptions, log: any) {
     this.url = options.sandbox ? PAYPAL_SANDBOX_API : PAYPAL_LIVE_API;
     this.ipnUrl = options.sandbox ? PAYPAL_SANDBOX_IPN : PAYPAL_LIVE_IPN;
     this.user = options.user;
@@ -297,6 +299,7 @@ export class PayPalClient {
     };
     this.emitter = new EventEmitter();
     this.on = this.emitter.on.bind(this.emitter);
+    this.log = log;
   }
 
   private objectToNVP(object: Record<string, any>): string {
@@ -376,7 +379,9 @@ export class PayPalClient {
     if (resultObj.ACK === 'Success' || resultObj.ACK === 'SuccessWithWarning') {
       return resultObj;
     } else {
-      throw new PayPalClientError(result.text, resultObj);
+      const err = new PayPalClientError(result.text, resultObj);
+      this.log.error('paypalClient.doRequest', err);
+      throw err;
     }
   }
 
--- a/packages/fxa-auth-server/bin/email_notifications.js
+++ b/packages/fxa-auth-server/bin/email_notifications.js
@@ -39,7 +39,8 @@ if (config.subscriptions && config.subscriptions.stripeApiKey) {
     const { PayPalClient } = require('../lib/payments/paypal-client');
     const { PayPalHelper } = require('../lib/payments/paypal');
     const paypalClient = new PayPalClient(
-      config.subscriptions.paypalNvpSigCredentials
+      config.subscriptions.paypalNvpSigCredentials,
+      log
     );
     Container.set(PayPalClient, paypalClient);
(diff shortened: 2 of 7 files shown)


packages/fxa-auth-server/lib/payments/paypal-client.ts
```diff
@@ -376,7 +379,9 @@ export class PayPalClient {
     if (resultObj.ACK === 'Success' || resultObj.ACK === 'SuccessWithWarning') {
       return resultObj;
     } else {
-      throw new PayPalClientError(result.text, resultObj);
+      const err = new PayPalClientError(result.text, resultObj);
+      this.log.error('paypalClient.doRequest', err);
+      throw err;
```
biancadanforth, line 384: We could add a log here, but IIUC those alone would only show up in BigQuery not Sentry, so we probably want to also modify the `PayPalClientError`, as its `this.message` is what we're seeing in Sentry right now.

If you look at that class' `constructor` method, you'll see that we parse the NVP response (initially `this.raw`, a string, for the `PayPalClientError` instance) into an object with type `NVPResponse` (`this.data`).

I think we want to make sure that one of those two values (`this.data` would be easier to read) is sent as part of the `PayPalClientError` to Sentry.

@bbangert or @chenba , is there a good example of adding this extra key-value data to an error captured by Sentry? Looking at [`AppError`](https://github.com/mozilla/fxa/blob/9838b2df66fcbaf2014b7b210e44d16a34439511/packages/fxa-auth-server/lib/error.js#L175), I see we have a `this.output.payload` which may be the right place?

chenba (dismissed):
@IvoJP sorry about the confusion here (especially since we labeled it as a good first issue!).  What we want is to add more context to the Sentry event, not log out to a file. You can grep `Sentry.withScope` to find how that's done elsewhere in the code base.

biancadanforth (changes requested):
Thanks Ivo. IIUC we would want to update the `PayPalClientError` class to report additional information, as the logs would show up in BigQuery, but not in Sentry.

I've asked Ben and Barry if there's a good example of adding this additional information to an error, so that Sentry picks it up. I think you'd wand to add the additional data to `this.output.payload` looking at our other error class, `AppError`.

packages/fxa-auth-server/lib/payments/paypal-client.ts
biancadanforth: Answer: `Sentry.withScope`

packages/fxa-auth-server/lib/error.js
```diff
@@ -283,6 +283,11 @@ AppError.translate = function (request, response) {
       info: payload.info,
       stack: response.stack,
     });
+
+    if (response.data) {
+      error.output.payload.data = JSON.stringify(response.data);
+    }
```
biancadanforth, line 289: Screen Shot 2021-11-08 at 2 19 56 PM

I confirmed locally that this logs a stringified version of `PayPalClientError.data` under Payload > data in Sentry when pointing `sentryDsn` to Ben's local test Sentry project in fxa-auth-server config.

packages/fxa-auth-server/lib/payments/paypal-client.ts
biancadanforth, line 384: I think we can simplify this patch by reverting the `this.log` stuff? Else update the `doRequest` error tests to assert that it is now called with what arguments.

packages/fxa-auth-server/lib/error.js
biancadanforth: I'm not going to block on it, but ideally there'd be a test in `fxa-auth-server/test/local/error.js` for this new behavior. You could use the `do_reference_transaction_failure.json` fixture for it.

biancadanforth (approved):
R+WC. Thanks Ivo. This looks to be working.

I think we could simplify this diff to just be the three lines in `error.js`, since we don't need the logs:
```js
    if (response.data) {
      error.output.payload.data = JSON.stringify(response.data);
    }
```

Also you could add a test for this new behavior (to `test/local/error.js), but I'm not going to block on it.

packages/fxa-auth-server/lib/error.js
IvoJP: test added

PR #10889, opened 2021-11-02
feat(auth): store payment methods in firestore

### LINKED ISSUE #10746: Store Payment Methods in Firestore
Invoice emails and other functions require the payment method. This can end up fetching it multiple times in the same request. We should store these and fetch them from Firestore.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4118)

bbangert:
Because:

* We query payment methods on emails which can hit Stripe APIs
  extensively for data that rarely changes.

This commit:

* Stores the payment method object in Firestore.

Closes #10746

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe-firestore.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe-firestore.ts
@@ -9,6 +9,7 @@ export enum FirestoreStripeError {
   FIRESTORE_CUSTOMER_NOT_FOUND = 'FirestoreCustomerNotFound',
   FIRESTORE_SUBSCRIPTION_NOT_FOUND = 'FirestoreSubscriptionNotFound',
   FIRESTORE_INVOICE_NOT_FOUND = 'FirestoreInvoiceNotFound',
+  FIRESTORE_PAYMENT_METHOD_NOT_FOUND = 'FirestorePaymentMethodNotFound',
   STRIPE_CUSTOMER_MISSING_UID = 'StripeCustomerMissingUid',
   STRIPE_CUSTOMER_DELETED = 'StripeCustomerDeleted',
 }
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/payments/stripe-firestore.ts
```diff
@@ -211,6 +232,61 @@ export class StripeFirestore {
...
+   * subscriptions.
+   */
+  async insertPaymentMethodRecordWithBackfill(
+    paymentMethod: Partial<Stripe.PaymentMethod>
+  ) {
+    try {
+      await this.insertPaymentMethodRecord(paymentMethod);
+    } catch (err) {
+      if (err.name === FirestoreStripeError.FIRESTORE_CUSTOMER_NOT_FOUND) {
+        await this.fetchAndInsertCustomer(paymentMethod.customer as string);
```
chenba, line 269: We'd still need to insert the payment method here correct?

bbangert: Yep, we sure do! Doh!

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2554,7 +2613,9 @@ export class StripeHelper {
         case 'invoice.payment_failed':
         case 'invoice.updated':
         case 'invoice.deleted':
-          await this.processInvoiceEventToFirestore(event);
+          if (!event.request?.id) {
+            await this.processInvoiceEventToFirestore(event);
+          }
```
chenba, line 2618: hmm I don't think we update the invoices directly in Firestore after mutations from us?

bbangert: Hmm, I'll have to check and review.

chenba (changes requested):
Please see inline comments.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: We do not update it each time, removing this.

chenba (approved):

Issue #10893, opened 2021-11-02
[subplat] Canceled Subscriptions are Displayed in the Subscriptions Management Page


data-sync-user:
We recently switch to a different end point for getting a customer's billing info and subscriptions (so that IAP subscriptions are included).  However, that introduced a bug where canceled Stripe based subscriptions are included.

STR:

1. Subscribe via the web
2. Find the subscription in Stripe Dashboard and cancel it with immediate effect
3. Go to the subscriptions management page

Expect:

The subscription is not shown (or if it was the last subscription, the user cannot view the subscription management page as the user is no longer a customer)

Actual:

The subscription is displayed on the subscription management page

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4168)

data-sync-user:
➤ Luminita Buruian commented:

Verified this on Train 1.219.3 - Staging environment.
After canceling the subscription from Stripe dashboard:

  * the subscription management page is no longer available in the user account, meaning that the subscription has been successfully canceled;
  * the user receives an email informing the user that the subscription has been canceled due to unsuccessful payments;
  * logging into the VPN client with the account with the canceled subscription, redirect the user to the Mozilla VPN main website, to subscribe.

Please see attached video for more information on the performed test:

!2021-11-15_14h59_39.mp4|width=1920,height=1080!

Barry Chen please let us know if there is anything else we need to check for this issue.

Thank you.

Valentina Virlics Madalin Cotetiu

PR #10896, opened 2021-11-02
fix(subscriptions): filter canceled subscriptions

chenba:
Because:
 - Stripe's default behavior of not returning canceled subscription no
   longer applies now that we get subscriptions from Firestore

This commit:
 - explicitly filtering for subscriptions with one of the active
   statuses for the endpoint that returns billing info and subscriptions

Closes: #10893


bbangert (approved):
Looks good, just needs a test fix.

Issue #10906, opened 2021-11-03
[subplat] [email] “$” currency displayed in the Welcome email for new users instead of “US$”


data-sync-user:
**Affected versions**:

* All;

**Affected FxA versions**:

* 1.218.4;

**Prerequisites**:

* Be in a Mozilla VPN supported country (US);
* Be on new Set up your subscription page;

**Steps to reproduce**:
1. Subscribe to Mozilla VPN via Stripe;
2. Open your email inbox for the account from step 1;
3. Open the “Welcome to Mozilla VPN: Please set your password.” email;
4. Observe the currency displayed in the email;

**Expected result**:

* The currency displayed is “US$”;

**Actual result**:

* The currency displayed is “$”;

**Notes**:

* Reproducible for en-SG, en-CA as well;
* “Mozilla VPN payment confirmed” email displays “US$” currency;
* See the attached screenshots.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4173)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17421/Bildschirmfoto+2021-11-03+um+12.34.09.png">Bildschirmfoto 2021-11-03 um 12.34.09.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/17420/Bildschirmfoto+2021-11-03+um+12.34.43.png">Bildschirmfoto 2021-11-03 um 12.34.43.png</a>

data-sync-user:
➤ Madalin Cotetiu commented:

Verified this as fixed on Train 1.218.5 - Staging. 

The currency displayed in the “Welcome to Mozilla VPN: Please set your password.” email is “US$”. 

See the attached screenshot. 

!Screenshot 2021-11-04 at 09.11.22.png|width=1196,height=794!

data-sync-user:
➤ Madalin Cotetiu commented:

Hello,

While testing on Train 1.218.6 - Staging we made some subscription for Mozilla VPN (locale: US) and we noticed that the currency displayed in the “Welcome to Mozilla VPN” email and “Mozilla VPN payment confirmed” email, is “$” instead of “US$”, regardless of the payment method (Stripe/PayPal).

Is this intended?

Thank you.

data-sync-user:
➤ Barry Chen commented:

Yes, that (only ‘$') is correct when the locale is en-US.

PR #10908, opened 2021-11-03
fix(emails): pass language for new sub/stub acct email

chenba:
Because:
 - we did not pass the account locale for the new account+subscription
   email

This commit:
 - pass in the account locale for the new account+subscription email

Closes: #10906


chenba:
Forgot to update some tests.  😢

vbudhram (approved):

PR #10912, opened 2021-11-03
feat(auth): remove redis customer caching

### LINKED ISSUE #10747: Remove redis caching of customer call
The customer call, once using expandResource and firestore, no longer should cache the customer object. This reduces the quantity of cache clears and async event loop spins with a minimal overhead of additional Firestore calls.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4119)

bbangert:
Because:

* We don't want to maintain multiple local caches now that
  we store copies in Firestore.
* It's more efficient to only manage one cache of Stripe data
  instead of maintaining multiple caches.
* Some methods were using customer functions that did not actually
  need the cached data, or all the data we fetched and cached.

This commit:

* Removes Redis caching of the customer object.
* Replaces customer calls with the appropriate underlying fetchCustomer
  call with the proper expansions in place as needed to reduce
  cache lookups.
* Replaces some customer calls with a db lookup to fetch the Stripe
  customer ID instead of a cache call.

Closes #10747

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -5,7 +5,6 @@ import { Firestore } from '@google-cloud/firestore';
 import * as Sentry from '@sentry/node';
 import cacheManager, {
   Cacheable,
-  CacheClear,
   CacheClearStrategy,
   CacheClearStrategyContext,
   CacheUpdate,
@@ -141,7 +140,6 @@ export class StripeHelper {
   // Note that this isn't quite accurate, as the auth-server logger has some extras
   // attached to it in Hapi.
   private log: Logger;
-  private customerCacheTtlSeconds: number;
   private plansAndProductsCacheTtlSeconds: number;
   private stripeTaxRatesCacheTtlSeconds: number;
   private webhookSecret: string;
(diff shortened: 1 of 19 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2550,17 +2473,12 @@ export class StripeHelper {
    * only change in their entirety.
    */
   async processPaymentMethodEventToFirestore(event: Stripe.Event) {
-    const { data } = event;
-
-    const paymentMethod = data.object as Stripe.PaymentMethod;
-
-    if (event.type === 'payment_method.detached') {
-      await this.stripeFirestore.removePaymentMethodRecord(paymentMethod.id!);
```
chenba, line 2558: Why don't we delete anymore?

chenba (approved):
`+255 −590` love that ratio

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Because if we want to check for card fingerprints used in the future, it can be useful to know who previously used the card.

Issue #10684, opened 2021-10-12
[subplat] Return 200 on Stripe Webhooks for Deleted Customers


data-sync-user:
Whenever we receive webhooks for a user whose customer record has been deleted, we should return a 200 to Stripe so that it'd stop retrying as the retries will never succeed.  If the missing customer record is an exception then report an error to Sentry.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4084)

chenba:
Sentry issue: [FXA-AUTH-PROD-E5](https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/13667940/?referrer=github_plugin)

data-sync-user:
➤ Barry Chen commented:

Reino Muhl I don’t recall seeing issues with deleted customers recently.  What do you say to we canceling this issue, and open new issue(s) to address any bugs as they appear in the future?

data-sync-user:
➤ Reino Muhl commented:

Yup, I agree with that! I’m happy to cancel this ticket.

PR #10921, opened 2021-11-04
feat(auth): 200 for webhook, deleted cust

StaberindeZA:
Because:

* If a customer has been deleted, the Stripe webhook will keep retrying.

This commit:

* Returns a 200 to Stripe for deleted customers.

Closes #
[Issue 10684](https://github.com/mozilla/fxa/issues/10684)

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -159,6 +159,12 @@ export class StripeWebhookHandler extends StripeHandler {
       }
       // Caught an ignorable error, so let's log but continue to 200 OK response
       this.log.error('subscriptions.handleWebhookEvent.failure', { error });
+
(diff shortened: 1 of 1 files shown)


chenba:
@StaberindeZA This is certainly wasn't what I had in mind.  However! It's totally my fault that I was not very detailed with the issue I filed.  Moreover, things have changed quite a bit with the addition of Firestore.  I think I need to go through the webhook event handlers to see what we need.

I guess one general comment that applies here is that we'd need tests for the changes.  (But don't worry about it yet in this instance since we don't know what changes we need.)

StaberindeZA:
@chenba I added some logic to send the error to sentry, but found a couple of other things while testing things. Would like to discuss this a bit more with you next week.

chenba:
> 
> 
> @chenba I added some logic to send the error to sentry, but found a couple of other things while testing things. Would like to discuss this a bit more with you next week.

Sorry about the late reply.  Yeah, we can talk about it.  In the meantime, can you set this PR as a draft.

PR #10923, opened 2021-11-04
fix(auth): always fetch latest stripe object to store

### LINKED ISSUE #10904: [subplat] Currency error message displayed when trying to update card information
**Affected versions**:

* All;

**Affected FxA versions**:

* 1.218.4;

**Prerequisites**:

* Be in a Mozilla VPN supported country (US);

**Steps to reproduce**:
1. Subscribe to Mozilla VPN monthly plan using Stripe and a US card (e.g. 4242424242424242);
2. Go to Subscription Management page for the same account;
3. Click on the “Change” button to update payment details (e.g. 4000056655665556);
4. Click on the “Update” button;
5. Observe the behaviour;

**Expected result**:

* The payment details are successfully updated. 

**Actual result**:

* “The currency of this subscription is not valid for the country associated with your payment” error message is displayed in the Payment section. The card number cannot be updated.

**Notes**:

* This is a major issue because the user is not able to update his payment information;
* Reproducible if using the same card number that was used for making the subscription;
* Reproducible for new and old users;
* Reproducible regardless of the currency;
* See the attached screenshot.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4171)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17413/Screenshot+2021-11-02+at+17.47.12.png">Screenshot 2021-11-02 at 17.47.12.png</a>

bbangert:
Because:

* Webhooks may come out of order, or be triggered by requests we perform
  without storing the latest object.

This commit:

* Always stores the latest object from stripe in firestore.

Closes #10904

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -15,6 +15,7 @@ import {
   createAccountCustomer,
   deleteAccountCustomer,
   getAccountCustomerByUid,
+  getUidAndEmailByStripeCustomerId,
   updatePayPalBA,
 } from 'fxa-shared/db/models/auth';
 import {
@@ -24,6 +25,7 @@ import {
 } from 'fxa-shared/subscriptions/stripe';
 import {
   AbbrevPlan,
+  AbbrevPlayPurchase,
   AbbrevProduct,
   MozillaSubscriptionTypes,
   PAYPAL_PAYMENT_ERROR_FUNDING_SOURCE,
@@ -34,7 +36,6 @@ import {
 } from 'fxa-shared/subscriptions/types';
 import { StatsD } from 'hot-shots';
 import ioredis from 'ioredis';
-import pick from 'lodash/pick';
 import moment from 'moment';
 import { Logger } from 'mozlog';
 import { Stripe } from 'stripe';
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2454,7 +2454,12 @@ export class StripeHelper {
    */
   async processCustomerEventToFirestore(event: Stripe.Event) {
     const { data } = event;
-    const customer = data.object as Stripe.Customer;
+    const customer = await this.stripe.customers.retrieve(
+      (data.object as Stripe.Customer).id
+    );
+    if (customer.deleted) {
```
chenba, line 2460: hmmm I noticed that the other set of handlers does not handle `customer.deleted`, so if we just return early here, we won't ever know that a customer has been deleted I think.

chenba (changes requested):
I think we need to save the customer into Firestore on delete.

packages/fxa-auth-server/lib/payments/stripe.ts
bbangert: Ah, good catch, because we will not have touched the record in Firestore.

chenba (approved):
💰

PR #10924, opened 2021-11-04
feat(settings): Display DataCollection in UI, increase test coverage

### LINKED ISSUE #10732: (Settings - do last) Enable UI for users, update tests
FXA-2931 sets up the front-end for the opt-out of account level telemetry.

However, we can't enable it for users until all other pieces are in place. This ticket just turns the feature "on" for the user, 1) removing conditionally displaying the components in the Nav and Settings components, 2) using `useAccount` to grab and use the real field name, and 3) replacing the settimeout from the PR for FXA-2931 with a mutation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4106)

LZoog:
Because:
* We want users to be able to use the feature now that we're more code complete

This commit:
* Removes conditionals hiding this feature
* Handles toggling errors
* Increases test coverage including metric events

fixes #10732

### DIFF
--- a/packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
@@ -10,8 +10,13 @@ import {
   renderWithRouter,
   mockAppContext,
 } from 'fxa-settings/src/models/mocks';
+import { logViewEvent } from '../../lib/metrics';
 import { isMobileDevice } from '../../lib/utilities';
 import { MOCK_SERVICES } from './mocks';
+
+jest.mock('../../lib/metrics', () => ({
+  logViewEvent: jest.fn(),
+}));
 
 const SERVICES_NON_MOBILE = MOCK_SERVICES.filter((d) => !isMobileDevice(d));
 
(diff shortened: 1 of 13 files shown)


LZoog:
Added a `do not merge` label - per a meeting with legal & others, we're holding on enabling this for users for now.

packages/fxa-settings/src/components/DataCollection/index.tsx
```diff
@@ -38,7 +39,16 @@ export const DataCollection = () => {
...
+    } catch (err) {
+      alertBar.error(
+        l10n.getString(
+          'dc-opt-in-out-error',
+          null,
+          'Sorry, there was a problem changing your data collection preference.'
+        )
+      );
+    }
+    setIsSubmitting(false);
```
dschom, line 51: Random nitpick. Use finally block for critical operations.

LZoog:
Circling back here, the earliest we should merge this is January 13.

I still need an r+ but know there's one comment above I can address too.

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
```diff
@@ -254,9 +268,14 @@ describe('Connected Services', () => {
...
     await clickFirstSignOutButton();
     await clickConfirmDisconnectButton();
+    expect(logViewEvent).toHaveBeenCalledWith(
+      'settings.clients.disconnect',
+      'submit.'
+    );
     const finalCount = (
       await screen.findAllByTestId('settings-connected-service')
     ).length;
+    // TODO: fix this test, it no workey
```
xlisachan, line 278: I think this is due to `clickFirstSignOutButton()`. Per our discussion, this does not affect your PR functionality. We can update this at a later time.

xlisachan (approved):

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
```diff
@@ -3,16 +3,21 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import React from 'react';
-import { act, fireEvent, screen } from '@testing-library/react';
+import { act, fireEvent, screen, waitFor } from '@testing-library/react';
```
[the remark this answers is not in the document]
LZoog: Oh nice catch. I think there's a prettier or eslint config we can set up to alphabetize/sort our imports for us that also removes these automatically, I already have a list of issues I need to file so I'll add that on there and see if it already exists somewhere.

packages/fxa-settings/src/components/DataCollection/index.tsx
```diff
@@ -38,7 +39,17 @@ export const DataCollection = () => {
...
+          null,
+          'Sorry, there was a problem changing your data collection preference.'
+        )
+      );
+    } finally {
+      setIsSubmitting(false);
+    }
   }, [account, alertBar, l10n, setIsSubmitting]);
 
   return (
```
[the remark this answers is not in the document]
LZoog: Thanks for catching this!

packages/fxa-settings/src/components/ConnectedServices/index.test.tsx
LZoog: Thanks for looking into it and calling it out, this PR is a couple months old so I'd forgotten about this. I created a follow up for it and will reference it on this TODO line: FXA-4453 / #11658

Issue #10542, opened 2021-09-28
[subplat][support-panel] List IAP subscriptions


data-sync-user:
The Support Panel lists subscriptions for a user.  However, that list currently only includes Stripe subscriptions.  Update the list to include IAP subscriptions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4033)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18059/image-20211119-224029.png">image-20211119-224029.png</a>

data-sync-user:
➤ Luminita Buruian commented:

Barry Chen could you please help us out with some details regarding what exactly we need to check and possibly some testing steps?

Thank you very much. Highly appreciated.

data-sync-user:
➤ Barry Chen commented:

Jon Buckley or Dustin Lactin  where is the Stage support panel?  and do we need corporate VPN in order to access it?

data-sync-user:
➤ Jon Buckley commented:

It’s located at https://fxa-support-panel.stage.mozaws.net ( https://fxa-support-panel.stage.mozaws.net ) and does require being on the corporate VPN

data-sync-user:
➤ Barry Chen commented:

Moved back to In Progress to do some minor refactoring.

data-sync-user:
➤ Barry Chen commented:

Luminita Buruian  Here are the prerequisites you’ll need in order to test this jira.

A Google Play IAP Subscription on StageI’ve never done this before.  Perhaps you can get the steps from Rebecca Billings or QA on the VPN team.

Corporate VPN AccessMaybe Rebecca Billings can also help here if you don’t have access?

The FxA User IDYou’ll need the “uid” of the FxA user with the Stage IAP subscription.  One way to find this is to open the browser developer tools while you are viewing the stage Settings page, navigate to the Storage tab, and expand Local Storage on the left navigation.  Click on ‘https://accounts.stage.mozaws.net ( https://accounts.stage.mozaws.net )' and copy the value with the key of ‘__fxa_storage.currentAccountUid’ (without the double quotes).  See 

!image-20211119-224029.png|width=453,height=206!

Once that’s all done, then go to https://fxa-support-panel.stage.mozaws.net/?uid=YOUR_UID_HERE

data-sync-user:
➤ Barry Chen commented:

[...]

PR #10948, opened 2021-11-09
feat(support-panel): add Play subscriptions to support-panel

chenba:
Because:
 - support agent might want to know whether a customer has IAP
   subscriptions

This commit:
 - replaced the endpoint that the Support Panel used to get a list of
   subscriptions with one that also includes Google Play subscriptions
 - update the Support Panel to use the new endpoint and display the
   Google Play subscriptions

Fixes: #10542

### DIFF
--- a/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
@@ -4,10 +4,10 @@
 
 import Container from 'typedi';
 import { internalValidationError } from '../../../lib/error';
-import { AppConfig } from '../../types';
 import { PlayBilling } from './play-billing';
 import { AbbrevPlayPurchase } from 'fxa-shared/subscriptions/types';
 import { SubscriptionPurchase } from './subscription-purchase';
+import { ConfigType } from '../../../config';
 
 // TODO move this when we add support for Apple IAP subscriptions
 export interface SubscriptionsService<T> {
@@ -34,8 +34,7 @@ export class PlaySubscriptions
 {
   private playBilling?: PlayBilling;
 
-  constructor() {
-    const config = Container.get(AppConfig);
+  constructor(config: ConfigType) {
     if (!config.subscriptions.enabled) {
       throw internalValidationError(
         'PlaySubscriptions',
--- a/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Container from 'typedi';
+import { internalValidationError } from '../../../lib/error';
+import { AppConfig } from '../../types';
+import { PlayBilling } from './play-billing';
+import { AbbrevPlayPurchase } from 'fxa-shared/subscriptions/types';
+import { SubscriptionPurchase } from './subscription-purchase';
+
+// TODO move this when we add support for Apple IAP subscriptions
+export interface SubscriptionsService<T> {
+  getSubscriptions: (uid: string) => Promise<T[]>;
+}
+
+/**
+ * Extract an AbbrevPlayPurchase from a SubscriptionPurchase
+ */
+export function abbrevPlayPurchaseFromSubscriptionPurchase(
+  purchase: SubscriptionPurchase
+): AbbrevPlayPurchase {
+  return {
+    auto_renewing: purchase.autoRenewing,
+    expiry_time_millis: purchase.expiryTimeMillis,
+    package_name: purchase.packageName,
+    sku: purchase.sku,
+    ...(purchase.cancelReason && { cancel_reason: purchase.cancelReason }),
+  };
+}
+
+export class PlaySubscriptions
+  implements SubscriptionsService<AbbrevPlayPurchase>
+{
+  private playBilling?: PlayBilling;
+
+  constructor() {
+    const config = Container.get(AppConfig);
+    if (!config.subscriptions.enabled) {
+      throw internalValidationError(
+        'PlaySubscriptions',
+        {},
+        new Error(
+          'Trying to new up PlaySubscriptions while subscriptions are disabled.  Check your dependency graph.'
+        )
+      );
+    }
+
+    if (Container.has(PlayBilling)) {
+      this.playBilling = Container.get(PlayBilling);
+    }
+  }
+
+  async getSubscriptions(uid: string) {
+    if (!this.playBilling) {
+      return [];
+    }
+
+    const allPurchases =
+      await this.playBilling.userManager.queryCurrentSubscriptions(uid);
+    const purchases = allPurchases.filter((purchase) =>
+      purchase.isEntitlementActive()
+    );
+
+    return purchases.map(abbrevPlayPurchaseFromSubscriptionPurchase);
+  }
+}
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -1054,9 +1054,11 @@ export class StripeHelper {
   /**
    * Append any matching product ids to their corresponding AbbrevPlayPurchase.
    */
-  async appendAbbrevPlayPurchasesWithProductIds(
+  async addProductInfoToAbbrevPlayPurchases(
     purchases: AbbrevPlayPurchase[]
-  ): Promise<(AbbrevPlayPurchase & { product_id: string })[]> {
+  ): Promise<
+    (AbbrevPlayPurchase & { product_id: string; product_name: string })[]
+  > {
     const products = await this.allAbbrevProducts();
     const appendedAbbrevPlayPurchases = [];
     for (const product of products) {
@@ -1068,6 +1070,7 @@ export class StripeHelper {
         appendedAbbrevPlayPurchases.push({
           ...matchingAbbrevPlayPurchase,
           product_id: product.product_id,
+          product_name: product.product_name,
         });
       }
     }
--- a/packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
@@ -3,10 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { ServerRoute } from '@hapi/hapi';
+import isA from '@hapi/joi';
 import { Container } from 'typedi';
 import error from '../../error';
 import { PaymentBillingDetails, StripeHelper } from '../../payments/stripe';
-import { CapabilityService } from '../../../lib/payments/capability';
+import { PlaySubscriptions } from '../../../lib/payments/google-play/subscriptions';
 import {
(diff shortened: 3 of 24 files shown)


packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
```diff
@@ -0,0 +1,67 @@
...
+        )
+      );
+    }
+
+    if (Container.has(PlayBilling)) {
+      this.playBilling = Container.get(PlayBilling);
+    }
+  }
+
+  async getSubscriptions(uid: string) {
```
chenba, line 54: The main change is that I moved the Play subscriptions fetching function here.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1107,6 +1107,7 @@ export class StripeHelper {
         appendedAbbrevPlayPurchases.push({
           ...matchingAbbrevPlayPurchase,
           product_id: product.product_id,
+          product_name: product.product_name,
```
chenba, line 1110: This is relevant to both of our issues.

chenba (commented):
@biancadanforth here's the WIP that might affect your current issue.  See inline comments.

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
biancadanforth, line 54: This method name is a little misleading IMO as it returns `AbbrevPlayPurchase[]`, not `GooglePlaySubscription[]`.

I renamed this method in #10964 to `getAbbrevPlayPurchases` and added a new method to the new `playSubscriptions` class created in this PR called `getActiveGooglePlaySubscriptions` whose return type is `GooglePlaySubscription[]`. What do you think about these changes? I can handle any name changes in my PR if preferred.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1107,6 +1109,7 @@ export class StripeHelper {
...
```
biancadanforth, line 1112: Yep; I had made this exact same change (with a slightly different method name).

packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
```diff
@@ -117,4 +132,28 @@ export class MozillaSubscriptionHandler {
       subscriptions: [...response.subscriptions, ...iapGooglePlaySubscriptions],
     };
   }
+
+  async getSubscriptionsForSupportPanel(request: AuthRequest) {
+    this.log.begin('mozillaSubscriptions.supportPanelSubscriptions', request);
+    const { uid } = request.query as Record<string, string>;
+    const iapPlaySubscriptions = await this.getPlaySubscriptions(uid);
+    const customer = await this.stripeHelper.fetchCustomer(uid);
```
biancadanforth, line 140: The `/account` route uses `this.stripeHelper.customer`. Does it make a difference? It looks like `customer` checks the Redis cache first, otherwise it also calls `fetchCustomer`.

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -377,6 +377,23 @@ module.exports.subscriptionsSubscriptionSupportValidator = isA.object({
   status: isA.string().required(),
   subscription_id: module.exports.subscriptionsSubscriptionId.required(),
 });
+module.exports.subscriptionsPlaySubscriptionSupportValidator = isA.object({
+  auto_renewing: isA.bool().required(),
+  cancel_reason: isA.number().optional(),
+  expiry_time_millis: isA.number().required(),
+  package_name: isA.string().optional(),
+  sku: isA.string().optional(),
+  product_id: isA.string().optional(),
```
biancadanforth, line 386: Are these optional, because they aren't currently used by the support panel?

packages/fxa-auth-server/test/local/routes/validators.js
```diff
@@ -755,4 +755,96 @@ describe('lib/routes/validators:', () => {
...
+        assert.ok(!res.error);
+      });
+
+      for (const x of required) {
+        it(`rejects when the required property ${x} is not present`, () => {
+          const s = { ...webSub, [x]: undefined };
+          const res =
+            validators.subscriptionsWebSubscriptionSupportValidator.validate(s);
+          assert.ok(res.error);
+        });
```
biancadanforth, line 803: Nice

packages/fxa-auth-server/test/local/routes/validators.js
```diff
@@ -755,4 +755,96 @@ describe('lib/routes/validators:', () => {
...
+        const res =
+          validators.subscriptionsSubscriptionSupportValidator.validate(subs);
+        assert.ok(!res.error);
+      });
+
+      it('accepts empty arrays', () => {
+        const subs = {
+          [MozillaSubscriptionTypes.WEB]: [],
+          [MozillaSubscriptionTypes.IAP_GOOGLE]: [],
+        };
```
biancadanforth, line 843: Is this a possible scenario for the Support Panel? Where there are no active subscriptions?

biancadanforth (commented):
Leaving some questions/comments in line. I haven't worked on the Support Panel before, so I will defer the official review to someone else.

I noticed the SubMan page on payments isn't loading due to a validation error. I ran into this in #10964 . The `/billing-and-subscriptions` endpoint's response is missing `product_name`.

packages/fxa-auth-server/lib/routes/validators.js
chenba: Yes.

packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
chenba: Ben has a patch that will remove all the customer Redis caching.  No difference, just some extra steps.

packages/fxa-auth-server/test/local/routes/validators.js
chenba: At the moment, Zendesk passes a FxA uid, so depending on how strictly we/Support prevent a non-customer from filing a Zendesk ticket, it's possible we'll receive a user who doesn't have subscriptions.

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
chenba: > 
> 
> This method name is a little misleading IMO ...

That's fair.  That name is from https://github.com/mozilla/fxa/pull/10948/files/dae26867f7073d09f3cb612059cf52c5aa165c91#diff-902713579f413cf56c1200a47bf8dfa9e80d3de15bb44ef04196904e032834d6R13-R15  That's because I want the same interface for any subscription type.  (I suppose we could argue whether that's helpful when the returned objects could be wildly different.)  If you change the name to something Play specific, then we cannot use that interface for web or Apple subscriptions.  And that's fine, but let's remove that interface in your PR.

chenba: (What the ... why is my reply here?!)

chenba: https://github.com/mozilla/fxa/pull/10948/files#r747067552 because I lost my ability to use Github?

chenba:
> 
> 
> Leaving some questions/comments in line. I haven't worked on the Support Panel before, so I will defer the official review to someone else.

I think that just leaves @bbangert ? 

> 
> I noticed the SubMan page on payments isn't loading due to a validation error. I ran into this in #10964 . The `/billing-and-subscriptions` endpoint's response is missing `product_name`.

hmm at some point one of my WIP commit was borked because of the response validator, so if that's when you tried, that _could_ have been it.  Let me try it locally and see if I can load the Subscriptions Management page...

biancadanforth:
>> Leaving some questions/comments in line. I haven't worked on the Support Panel before, so I will defer the official review to someone else.

>  I think that just leaves @bbangert ?

I can give it a go if there isn't a better person to do it. Let me know.

chenba:
> 
> 
> > > Leaving some questions/comments in line. I haven't worked on the Support Panel before, so I will defer the official review to someone else.
> 
> > I think that just leaves @bbangert ?
> 
> I can give it a go if there isn't a better person to do it. Let me know.

@biancadanforth  Let me know what you think you need in order to review.  The Support Panel is just a single page with basically no styles.  It displays some customer info with Handlebars.

biancadanforth:
> > > > Leaving some questions/comments in line. I haven't worked on the Support Panel before, so I will defer the official review to someone else.
> > 
> > 
> > > I think that just leaves @bbangert ?
> > 
> > 
> > I can give it a go if there isn't a better person to do it. Let me know.
> 
> @biancadanforth Let me know what you think you need in order to review. The Support Panel is just a single page with basically no styles. It displays some customer info with Handlebars.

Okay Barry. I'll just need to poke around a bit with the Support Panel locally if possible first. I recall someone mentioning it wasn't working, but I couldn't find it on first glance in the open issues.

chenba:
> 
> Okay Barry. I'll just need to poke around a bit with the Support Panel locally if possible first. I recall someone mentioning it wasn't working, but I couldn't find it on first glance in the open issues.
> 
> Edit: Perhaps it's not working in Zendesk, but I can load the iframe locally?

Sorry, just saw this.  Should we have a Zoom call to review this together, possibly with Ivo and Reino?

packages/fxa-support-panel/src/remote-lookup/remote-responses.dto.ts
```diff
@@ -17,7 +22,15 @@ interface Subscription {
   subscription_id: string;
 }
 
-export interface SubscriptionResponse extends Array<Subscription> {}
+type PlaySubscription = Omit<AbbrevPlayPurchase, 'expiry_time_millis'> & {
+  product_id: string;
+  expiry: string;
```
biancadanforth, line 27: Why is this named `expiry` instead? I don't see a similar key e.g. on `WebSubscription` above. I guess because it's not a ms value anymore but a formatted Date string?

biancadanforth (approved):
This LGTM; thanks for the help in getting the Support Panel loaded locally.

There are a couple of minor things that may be worth updating, but that's it assuming CI passes (are all those failures flaky tests?).

packages/fxa-auth-server/test/local/payments/google-play/subscriptions.js
```diff
@@ -0,0 +1,65 @@
...
+    };
+    Container.set(PlayBilling, mockPlayBilling);
+    playSubscriptions = new PlaySubscriptions();
+  });
+
+  afterEach(() => {
+    Container.reset();
+    sandbox.reset();
+  });
+
```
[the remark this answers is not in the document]
chenba: There wasn't a test for it in the old (current main) capability service tests.  The function was also private, so that could be why.

packages/fxa-support-panel/src/remote-lookup/remote-lookup.service.ts
```diff
@@ -46,22 +47,38 @@ export class RemoteLookupService {
...
-        current_period_start: String(
-          new Date(s.current_period_start * MS_IN_SEC)
-        ),
-        plan_changed: s.plan_changed
-          ? String(new Date(s.plan_changed * MS_IN_SEC))
-          : 'N/A',
-        previous_product: s.previous_product || 'N/A',
-      }));
+      return {
+        ...subscriptions,
```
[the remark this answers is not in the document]
chenba: Leftover from when I only map'd the web subscriptions.  I'll remove it.

Issue #10725, opened 2021-10-16
Show "Paid Suscriptions" link in FxA Settings when a user has an active IAP subscription


data-sync-user:
In testing [https://mozilla-hub.atlassian.net/browse/FXA-3838](https://mozilla-hub.atlassian.net/browse/FXA-3838|smart-link) locally, I noticed if my FxA user only has an IAP subscription, they can’t get to the Subscription Management page from FxA Settings like Web subscription subscribers can.

We should update the logic on displaying the “Paid Subscriptions” link to check for IAP subscriptions as well. This will likely mean using the new endpoint in [https://mozilla-hub.atlassian.net/browse/FXA-3838](https://mozilla-hub.atlassian.net/browse/FXA-3838|smart-link) or updating the GET `/account` route in the auth server.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4105)

data-sync-user:
➤ Bianca Danforth commented:

This feature will be available to test when Train 220 is in stage.

Steps to reproduce for QA verification

  1. Create a Google IAP subscription for the VPN in stage
  2. Navigate to accounts.firefox.com ( http://accounts.firefox.com ) and log in with the same FxA account used to create the Google IAP subscription.

Expected behavior:

The “Paid Subscriptions” link is present in the sidebar on the page.

data-sync-user:
➤ Luminita Buruian commented:

Verified as Fixed on Train 1.220.0 - Staging environment.

The Paid Subscriptions link is present in the Firefox account, both on Mobile and on desktop (see below images):

!image-20211124-111012.png|width=1071,height=624!

!image-20211124-111152.png|width=1370,height=340!

Clicking on the Manage button takes o the user to the Play store 

Verifying that e.g. the “Paid Subscriptions” link is not there when the IAP subscription is no longer active (and there are no other active subscriptions)will be done when Train 220 goes live in Prod.

Thank you.

CC: Madalin Cotetiu

PR #10668, opened 2021-10-11
feat(subscriptions): add endpoint for billing details and IAP & Stripe subscriptions

chenba:
Because:
 - some clients might need to know that a customer has IAP subscriptions

This commit:
 - add a new endpoint to include Google Play subscriptions as well as
   Stripe subscriptions
   - the response is a superset of the response from
     /oauth/subscriptions/customer

## Issue that this pull request solves

Closes: #10095

PR #10964, opened 2021-11-10
feat(auth): return IAP and web subscriptions in /account route

biancadanforth:
Because:
* We should show the 'Paid Subscriptions' link in FxA Settings for Google IAP subscribers.
* Currently, FxA Settings determines if a user has active subscriptions from the /account endpoint on the auth server.
* This endpoint previously only checked for subscriptions on the Stripe customer (what we call 'web subscriptions'), but a Google IAP subscriber with no web subscriptions won't have a Stripe customer.
    
This commit:
* Updates the /account route to return both IAP and web subscriptions.
    
Closes: #10725

## Checklist

This PR should not be merged before #10948 . Only the last commit is new.

### DIFF
--- a/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
@@ -56,8 +56,9 @@ export class PlaySubscriptions
       this.playBilling = Container.get(PlayBilling);
     }
 
-    // Since config.subscriptions.enabled is true here, StripeHelper will already exist.
-    this.stripeHelper = Container.get(StripeHelper);
+    if (Container.has(StripeHelper)) {
+      this.stripeHelper = Container.get(StripeHelper);
+    }
   }
 
   async getAbbrevPlayPurchases(uid: string) {
--- a/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
@@ -56,9 +56,8 @@ export class PlaySubscriptions
       this.playBilling = Container.get(PlayBilling);
     }
 
-    if (Container.has(StripeHelper)) {
-      this.stripeHelper = Container.get(StripeHelper);
-    }
+    // Since config.subscriptions.enabled is true here, StripeHelper will already exist.
+    this.stripeHelper = Container.get(StripeHelper);
   }
 
   async getAbbrevPlayPurchases(uid: string) {
--- a/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
+++ b/packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
@@ -1,13 +1,18 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
+import {
+  AbbrevPlayPurchase,
+  GooglePlaySubscription,
+  MozillaSubscriptionTypes,
+} from 'fxa-shared/subscriptions/types';
 import Container from 'typedi';
+
 import { internalValidationError } from '../../../lib/error';
+import { AppConfig } from '../../types';
+import { StripeHelper } from '../stripe';
 import { PlayBilling } from './play-billing';
-import { AbbrevPlayPurchase } from 'fxa-shared/subscriptions/types';
 import { SubscriptionPurchase } from './subscription-purchase';
-import { ConfigType } from '../../../config';
 
 // TODO move this when we add support for Apple IAP subscriptions
 export interface SubscriptionsService<T> {
@@ -30,11 +35,13 @@ export function abbrevPlayPurchaseFromSubscriptionPurchase(
 }
 
 export class PlaySubscriptions
-  implements SubscriptionsService<AbbrevPlayPurchase>
+  implements SubscriptionsService<GooglePlaySubscription>
 {
   private playBilling?: PlayBilling;
+  private stripeHelper?: StripeHelper;
 
-  constructor(config: ConfigType) {
+  constructor() {
+    const config = Container.get(AppConfig);
     if (!config.subscriptions.enabled) {
       throw internalValidationError(
         'PlaySubscriptions',
@@ -48,9 +55,13 @@ export class PlaySubscriptions
     if (Container.has(PlayBilling)) {
       this.playBilling = Container.get(PlayBilling);
     }
+
+    if (Container.has(StripeHelper)) {
+      this.stripeHelper = Container.get(StripeHelper);
+    }
   }
 
-  async getSubscriptions(uid: string) {
+  async getAbbrevPlayPurchases(uid: string) {
     if (!this.playBilling) {
       return [];
     }
@@ -62,5 +73,24 @@ export class PlaySubscriptions
     );
 
     return purchases.map(abbrevPlayPurchaseFromSubscriptionPurchase);
+  }
+
+  /**
+   * Gets all active Google Play subscriptions for the given user id
+   */
+  async getSubscriptions(uid: string): Promise<GooglePlaySubscription[]> {
+    if (!this.stripeHelper) {
+      return [];
+    }
+    const iapSubscribedGooglePlayAbbrevPlayPurchases =
+      await this.getAbbrevPlayPurchases(uid);
+    const iapAbbrevPlayPurchasesWithStripeProductData =
+      await this.stripeHelper.addProductInfoToAbbrevPlayPurchases(
+        iapSubscribedGooglePlayAbbrevPlayPurchases
+      );
+    return iapAbbrevPlayPurchasesWithStripeProductData.map((purchase) => ({
+      ...purchase,
+      _subscription_type: MozillaSubscriptionTypes.IAP_GOOGLE,
+    }));
   }
 }
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -1052,7 +1052,7 @@ export class StripeHelper {
   }
 
   /**
-   * Append any matching product ids to their corresponding AbbrevPlayPurchase.
+   * Append any matching product ids and names to their corresponding AbbrevPlayPurchase.
    */
   async addProductInfoToAbbrevPlayPurchases(
     purchases: AbbrevPlayPurchase[]
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -8,6 +8,10 @@ import {
   getAllPayPalBAByUid,
 } from 'fxa-shared/db/models/auth';
 import ScopeSet from 'fxa-shared/oauth/scopes';
+import {
+  GooglePlaySubscription,
+  WebSubscription,
+} from 'fxa-shared/subscriptions/types';
 import { Container } from 'typedi';
 import * as uuid from 'uuid';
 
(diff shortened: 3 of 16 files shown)


packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1470,39 +1472,47 @@ export class AccountHandler {
...
-          if (!customer) {
-            throw error.unknownCustomer(uid);
-          }
-          if (!customer.subscriptions) {
-            throw error.internalValidationError(
-              'Account.get',
-              { uid, customerId: customer.id },
-              'Expected customer to have subscriptions.'
-            );
-          }
```
biancadanforth, line 1493: I don't throw any of these errors anymore, because an IAP subscriber with no web subscriptions wouldn't have a Stripe `customer`. I now check if `customer && customer.subscriptions` below.

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -3396,95 +3420,205 @@ describe('/account', () => {
...
-    let failed = false;
-    try {
-      await runTest(buildRoute(), request, () => {});
-    } catch (err) {
-      failed = true;
-      assert.equal(err.errno, error.ERRNO.UNEXPECTED_ERROR);
-    }
+    it('should return formatted Google Play subscriptions when subscriptions are enabled', () => {
+      return runTest(buildRoute(), request, (result) => {
+        assert.deepEqual(mockStripeHelper.customer.args[0], [{ uid, email }]);
```
biancadanforth, line 3552: Why do we have to wrap the `args` assertion expected values in an array? I see that was done before my additional tests for e.g. `assert.deepEqual(mockStripeHelper.customer.args[0], [{ uid, email }]);`.

packages/fxa-auth-server/lib/routes/validators.js
```diff
@@ -350,24 +350,24 @@ module.exports.subscriptionsInvoicePIExpandedValidator = isA
 
 // This is subhub's perspective on an active subscription
 module.exports.subscriptionsSubscriptionValidator = isA.object({
```
biancadanforth, line 352: Rather than creating yet-another-validator-schema, I modified the existing subscriptionsSubscriptionValidator to make all fields but the subset for the IAP subscriptions optional.

In general Joi validation seems redundant to our type system, so I don’t think these validators are adding much in the first place, but I’m happy to create a new Joi schema if preferred and rename this one (or open to other suggestions).

biancadanforth (commented):
Added some specific notes/questions inline for the review.

biancadanforth:
Looking into the remote test failures...

biancadanforth:
> Looking into the remote test failures...

Two tests in test/remote/subscription_tests.js were failing:
* remote subscriptions: config.subscriptions.enabled = true and direct stripe access: with no subscriptions should return no active subscriptions
  - I have no idea what is invoking the /account handler for this test, but I noticed `playSubscriptions` was `{}` (and therefore truthy), so we were calling some Stripe Helper methods in that handler that had not been mocked.
* remote subscriptions: config.subscriptions.enabled = true and direct stripe access: with a subscription should return active subscriptions
  - Same story as above, and additionally, the fake web subscription in the test was failing my modified Joi validation.

Thanks to @chenba for pairing with me on this yesterday to get me onto the right track.

biancadanforth:
CI is passing and merge conflicts are resolved, so this is ready for review.

biancadanforth:
Working on resolving the merge conflicts again.

bbangert (changes requested):

packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
```diff
@@ -55,6 +55,27 @@ export const mozillaSubscriptionRoutes = ({
       handler: (request: AuthRequest) =>
         mozillaSubscriptionHandler.getBillingDetailsAndSubscriptions(request),
     },
+    {
+      // A Support Panel specific endpoint
```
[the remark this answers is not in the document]
chenba: This was me.  I can move it as part of FXA-4033.

biancadanforth:
I have 1 more failing remote test that I'm still working through, so I'm marking this as a draft...

packages/fxa-auth-server/lib/routes/account.ts
biancadanforth: This depends on the consumers of this endpoint and what they expect. Previously we threw an error, where now we return an empty array for `subscriptions`.

How does FxA Settings use this endpoint? Is it checking for if the `subscriptions` object is on the customer?

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
```diff
@@ -0,0 +1,98 @@
...
+import { internalValidationError } from '../../../lib/error';
+import { AppConfig } from '../../types';
+import { PlayBilling } from './play-billing';
+import {
+  AbbrevPlayPurchase,
+  GooglePlaySubscription,
+  MozillaSubscriptionTypes,
+} from 'fxa-shared/subscriptions/types';
+import { SubscriptionPurchase } from './subscription-purchase';
+import { StripeHelper } from '../stripe';
```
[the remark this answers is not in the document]
biancadanforth: Hmm I have that extension enabled, since it is recommended for FxA. I just enabled a setting for the extension called "TypeScript Hero > Imports: Organize on Save", and that seemed to do the trick. I'll do that for each modified file.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1499,7 +1511,7 @@ export class AccountHandler {
       }
     }
 
-    return { subscriptions };
+    return { subscriptions: [...abbrevIapSubscriptions, ...webSubscriptions] };
```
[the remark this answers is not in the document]
biancadanforth: Actually I got rid of the odd formatting here, so this is now just `iapGooglePlaySubscriptions` which is of type `GooglePlaySubscriptions[]`. Neither are abbreviated now.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1702,6 +1708,43 @@ export class StripeHelper {
...
+   * Formats IAP subscriptions for a customer into an appropriate response. This
+   * requires mapping e.g. Google Play SKUs to Stripe products via metadata.
+   *
+   * This method only checks for Google IAP subscriptions now; it will need to be
+   * updated when we add Apple IAP support.
+   */
+  async iapSubscriptionsToResponse(
+    subscriptions: IapSubscription[]
+  ): Promise<
+    Array<{ created: number; product_name: string; _subscription_type: string }>
```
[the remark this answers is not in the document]
biancadanforth: FWIW I got rid of this function entirely in my latest commit, but this is coming from an existing property added in PR #10668 (see [this discussion](https://github.com/mozilla/fxa/pull/10668/files#r730302068)).

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
```diff
@@ -0,0 +1,98 @@
...
+          'Trying to new up PlaySubscriptions while subscriptions are disabled.  Check your dependency graph.'
+        )
+      );
+    }
+
+    if (Container.has(PlayBilling)) {
+      this.playBilling = Container.get(PlayBilling);
+    }
+    if (Container.has(StripeHelper)) {
+      this.stripeHelper = Container.get(StripeHelper);
```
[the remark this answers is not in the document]
biancadanforth: Hmm I tried just calling `this.stripeHelper = Container.get(StripeHelper)`, but it broke an [unrelated route in CI](https://app.circleci.com/pipelines/github/mozilla/fxa/24778/workflows/80649443-cdbc-424b-91a4-db81d691fbe5/jobs/281524).

packages/fxa-graphql-api/src/gql/model/subscription.ts
```diff
@@ -5,7 +5,7 @@ import { Field, ObjectType } from '@nestjs/graphql';
 
 @ObjectType()
 export class Subscription {
-  @Field()
+  @Field({ nullable: true })
```
biancadanforth, line 8: For this field only, if I don't do this for an IAP subscription returned by the /account route (which does not have a `created` field) , I get a `"Error: Cannot return null for non-nullable field Subscription.created."` in the GraphQL API response in FxA Settings.

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
```diff
@@ -48,9 +55,13 @@ export class PlaySubscriptions
     if (Container.has(PlayBilling)) {
       this.playBilling = Container.get(PlayBilling);
     }
+
+    if (Container.has(StripeHelper)) {
+      this.stripeHelper = Container.get(StripeHelper);
```
chenba, line 60: Just echoing Ben here from another comment I saw: StripeHelper should exist here since subscriptions.enabled is true at this point.

packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
```diff
@@ -34,7 +32,7 @@ export const mozillaSubscriptionRoutes = ({
   playSubscriptions?: PlaySubscriptions;
 }): ServerRoute[] => {
   if (!playSubscriptions) {
-    playSubscriptions = new PlaySubscriptions(config);
+    playSubscriptions = new PlaySubscriptions();
```
chenba, line 35: Let's change to a `Container.get`.

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
biancadanforth: Yeah I pushed up a change and replied to Ben [here](https://github.com/mozilla/fxa/pull/10964#discussion_r751481217), but it broke some other tests due to `PlaySubscriptions` being instantiated as part of setting up the `mozillaSubscriptionRoutes`.

I can try and unset it or something before each of those tests.

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -3417,78 +3445,220 @@ describe('/account', () => {
...
+      package_name: 'org.mozilla.cooking.with.foxkeh',
+      sku: 'org.mozilla.foxkeh.yearly',
+      _subscription_type: MozillaSubscriptionTypes.IAP_GOOGLE,
+      product_id: 'iap_prod_lol',
+      product_name: 'LOL daily',
+    };
+
+    let subscriptionsEnabled, playSubscriptionsEnabled;
+
+    beforeEach(async () => {
```
chenba, line 3547: Looks like this fn doesn't need to be async.

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -3417,78 +3445,220 @@ describe('/account', () => {
...
-    const accountRoutes = makeRoutes({
-      config: {
-        subscriptions: {
-          enabled: subscriptionsEnabled,
-        },
-      },
-      log,
-      stripeHelper: mockStripeHelper,
+  describe('web subscriptions', () => {
+    beforeEach(async () => {
```
chenba, line 3454: No need to be async here I think.

chenba (approved):
r+ with comments

packages/fxa-auth-server/lib/routes/subscriptions/mozilla.ts
biancadanforth: Ah yes since `Container.get` will instantiate the module if it isn't set in the container yet.

packages/fxa-auth-server/lib/payments/google-play/subscriptions.ts
```diff
@@ -63,4 +74,23 @@ export class PlaySubscriptions
 
     return purchases.map(abbrevPlayPurchaseFromSubscriptionPurchase);
```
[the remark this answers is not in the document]
biancadanforth: Yes I agree with you it would be nice to refactor, but I would prefer to defer that to another patch, since those methods were not created in this PR, and we now have a few routes (not just /account) that leverage them. I can file a maintenance ticket for it.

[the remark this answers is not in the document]
biancadanforth: I filed https://mozilla-hub.atlassian.net/browse/FXA-4216 for the refactor.

biancadanforth:
I don't understand why these tests are all failing, as some of them start up the auth server with key_server.js, and they pass in config where `config.subscriptions.enabled` is `true` and `config.subscriptions.stripeApiKey` is defined (which would set StripeHelper on the container). Others are passing in our default config where Stripe is disabled.

Issue #10963, opened 2021-11-10
Oauth connection pool filled up and caused timeouts


data-sync-user:
On 2021-11-10 0600Z to 0635Z the FxA auth instances experienced increased latency and http 500's due to the connection pool filling up and hitting the connection pool timeout. After some debugging I decided to double the number of auth instances in service (from 15 to 30) and that resolved the latency and http 500's. [~accountid:5e3400fb6bd5c40ca395d189] looked into the code and noted that we're not configuring the minimum/maximum connection pool sizes (currently set to min: 2, max: 10) or the connection pool time outs. He also noted that we're running a new metrics opt out query for Amplitude receive event in this train that could've contributed to this resource exhaustion issue (approximately 3.5k calls/sec when we checked). We should double-check that the metricsOptOut caching on the session token is working in all cases: https://github.com/mozilla/fxa/pull/10817/files#diff-74ab6e1e0ed991af71049211d9627fb1c35ba66426a32585ec28cca3760e9e18R207 

I will set the FxA auth instances to run at double normal sizing by having them keep scaling at 30% CPU tonight and we will deploy a fix that allows us to configure the connection pool size/timeouts, and also see if we can collection some metrics on connection pools as well to catch resource exhaustion like this in the future.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4187)

data-sync-user:
➤ Jon Buckley commented:

Vijay Budhram wrote a PR https://github.com/mozilla/fxa/pull/10969 ( https://github.com/mozilla/fxa/pull/10969|smart-link )  and I wrote a PR https://github.com/mozilla-services/cloudops-deployment/pull/4448 ( https://github.com/mozilla-services/cloudops-deployment/pull/4448|smart-link ) and things are looking better so far

PR #10968, opened 2021-11-10
fix(metrics): Cache metricsOptOut property in request

vbudhram:
Fixes https://github.com/mozilla/fxa/issues/10963 I think...

### DIFF
--- a/packages/fxa-auth-server/lib/server.js
+++ b/packages/fxa-auth-server/lib/server.js
@@ -16,6 +16,7 @@ const sharedSecretAuth = require('./routes/auth-schemes/shared-secret');
 const pubsubAuth = require('./routes/auth-schemes/pubsub');
 const { HEX_STRING, IP_ADDRESS } = require('./routes/validators');
 const { configureSentry } = require('./sentry');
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/server.js
```diff
@@ -238,6 +239,40 @@ async function create(log, error, config, routes, db, translator, statsd) {
...
+      ) {
+        // sessionToken strategy comes with uid as uid
+        uid = request.auth.credentials.uid;
+      } else if (
+        request.auth &&
+        request.auth.credentials &&
+        request.auth.credentials.user
+      ) {
+        // oauthToken strategy comes with uid as user
+        uid = request.auth.credentials.user;
```
dannycoates, line 260: `metricsOptOutAt` will be set on `credentials` in the sessionToken case, and _i think_ in the oauth case too

dannycoates: actually it doesn't look like oauth has that. we might want to add it? 🤔

vbudhram:
Closing in favor of https://github.com/mozilla/fxa/pull/10990

PR #10969, opened 2021-11-10
fix(metrics): temporarily disable the metrics opt out check

vbudhram:
Disables the query to check for metrics opt out for now.

Fixes https://github.com/mozilla/fxa/issues/10963

### DIFF
--- a/packages/fxa-shared/db/index.ts
+++ b/packages/fxa-shared/db/index.ts
@@ -64,6 +64,9 @@ export function setupDatabase(opts: MySQLConfig): Knex {
     client: 'mysql',
     pool: {
       afterCreate: setupConnection,
+      min: 0,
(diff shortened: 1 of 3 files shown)


dannycoates (approved):

packages/fxa-shared/db/index.ts
```diff
@@ -64,6 +64,9 @@ export function setupDatabase(opts: MySQLConfig): Knex {
     client: 'mysql',
     pool: {
       afterCreate: setupConnection,
+      min: 0,
```
bbangert, line 67: Might be best to set this to a portion of the max (1/5th or so), to ensure there's idle connections around.

bbangert (approved):
The config change looks fine.

PR #10986, opened 2021-11-11
fix(auth): setup container objects for processor

### LINKED ISSUE #10985: paypal-processor errors on startup
`ServiceNotFoundError: Service "AUTH_FIRESTORE" was not found, looks like it was not registered in the container. Register it by calling Container.set before using service.at new ServiceNotFoundError (/fxa/node_modules/typedi/error/ServiceNotFoundError.js:20:28)at ContainerInstance.getServiceValue (/fxa/node_modules/typedi/ContainerInstance.js:159:19)at ContainerInstance.get (/fxa/node_modules/typedi/ContainerInstance.js:51:21)at Function.Container.get (/fxa/node_modules/typedi/Container.js:38:36)at new StripeHelper (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/stripe.js:129:45)at Object.setupProcesingTaskObjects (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/payments/processing-tasks-setup.js:50:26)at async init (/fxa/packages/fxa-auth-server/dist/fxa-auth-server/scripts/paypal-processor.js:26:40)`

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4191)

bbangert:
Because:

* Components started using containers that hadn't been setup for the
  paypal-processor resulting in it erroring on startup.

This commit:

* Sets up the needed objects and adds a basic statsd metric emit when
  startup completes.

Closes #10985

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/processing-tasks-setup.ts
+++ b/packages/fxa-auth-server/lib/payments/processing-tasks-setup.ts
@@ -5,9 +5,11 @@ import { setupAuthDatabase } from 'fxa-shared/db';
 import { StatsD } from 'hot-shots';
 import Container from 'typedi';
 
+import { setupFirestore } from '../firestore-db';
 import { CurrencyHelper } from '../payments/currencies';
-import { StripeHelper } from './stripe';
 import { configureSentry } from '../sentry';
+import { AuthFirestore, AuthLogger, ProfileClient } from '../types';
+import { StripeHelper } from './stripe';
 
 const config = require('../../config').getProperties();
 
@@ -31,12 +33,21 @@ export async function setupProcesingTaskObjects() {
       } as unknown as StatsD);
   Container.set(StatsD, statsd);
 
+  if (!Container.has(AuthFirestore)) {
+    const authFirestore = setupFirestore(config);
+    Container.set(AuthFirestore, authFirestore);
+  }
+
   const log = require('../log')({ ...config.log, statsd });
+  Container.set(AuthLogger, log);
 
   const translator = await require('../senders/translator')(
     config.i18n.supportedLanguages,
     config.i18n.defaultLanguage
   );
+  const profile = require('../profile/client')(log, config, statsd);
+  Container.set(ProfileClient, profile);
+
   const senders = await require('../senders')(
     log,
     config,
--- a/packages/fxa-auth-server/scripts/paypal-processor.ts
+++ b/packages/fxa-auth-server/scripts/paypal-processor.ts
@@ -2,13 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import program from 'commander';
-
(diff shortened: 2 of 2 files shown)


packages/fxa-auth-server/lib/payments/processing-tasks-setup.ts
```diff
@@ -31,12 +33,21 @@ export async function setupProcesingTaskObjects() {
...
+
   const log = require('../log')({ ...config.log, statsd });
+  Container.set(AuthLogger, log);
 
   const translator = await require('../senders/translator')(
     config.i18n.supportedLanguages,
     config.i18n.defaultLanguage
   );
+  const profile = require('../profile/client')(log, config, statsd);
+  Container.set(ProfileClient, profile);
```
biancadanforth, line 49: paypal-processor doesn't use this, but I guess some other cron jobs might?

packages/fxa-auth-server/scripts/paypal-processor.ts
```diff
@@ -48,7 +48,10 @@ export async function init() {
     database,
     senders.email
   );
+  const statsd = Container.get(StatsD);
```
biancadanforth, line 52: Do we know that `StatsD` is already set?

packages/fxa-auth-server/scripts/paypal-processor.ts
```diff
@@ -48,7 +49,11 @@ export async function init() {
...
+  statsd.increment('paypal-processor.startup');
   await processor.processInvoices();
+  statsd.increment('paypal-processor.shutdown');
+  await promisify(statsd.close).bind(statsd)();
```
biancadanforth, line 56: TIL: This is required to flush the `paypal-processor.shutdown` metric, otherwise it doesn't get logged.

biancadanforth (approved):
I have verified this locally compared to `main` and it does fix the issue and correctly report the two new StatsD metrics.

One thing I'm less certain of: this is doing a lot of `Container.set`s without checking if the object is already in the container. Is that a risk here? How are the containers for cronjobs shared, if at all, with the rest of the auth server?

packages/fxa-auth-server/lib/payments/processing-tasks-setup.ts
bbangert: Other objects created during the making of the stripe and paypal helper do want this around, yes.

packages/fxa-auth-server/scripts/paypal-processor.ts
bbangert: Yep, that's done above in the `setupProcesingTaskObjects` method.

packages/fxa-auth-server/scripts/paypal-processor.ts
bbangert: It's frequently not needed in longer running services where its not a big deal if a metric at the end doesn't make it out before the server shuts down. In this case we really want that metric.

PR #10990, opened 2021-11-11
fix(metrics): Re-enable toggling metrics for user

vbudhram:
Because:

* We disabled checking for the metrics opt out property on the user in train 218 since it caused unexpected amount of db load

This commit:

* Re-enables the metrics checking in auth-server
* Caches the metricsOptOut property in the request so it only is looked up once
* I am not 100% sure if this will fix the issue since the oauth metric events now need to do the opt out lookup.

Closes https://github.com/mozilla/fxa/issues/10963

## Checklist


packages/fxa-auth-server/lib/server.js
```diff
@@ -294,6 +344,11 @@ async function create(log, error, config, routes, db, translator, statsd) {
     'setMetricsFlowCompleteSignal',
     metricsContext.setFlowCompleteSignal
   );
+  server.decorate(
+    'request',
+    'getAndSetMetricsEnabled',
+    metricsContext.getAndSetMetricsEnabled
+  );
```
dannycoates, line 351: where is this used?

packages/fxa-auth-server/lib/server.js
```diff
@@ -238,6 +239,55 @@ async function create(log, error, config, routes, db, translator, statsd) {
...
+      // Note that unverified accounts can not be opted out of metrics.
+
+      let uid;
+      if (
+        request.auth &&
+        request.auth.credentials &&
+        request.auth.credentials.uid
+      ) {
+        // sessionToken strategy sets this property already
+        return !request.auth.credentials.metricsOptOutAt;
```
dannycoates, line 254: 👍

dannycoates (approved):

packages/fxa-auth-server/lib/server.js
vbudhram: This was left over from a different approach I was thinkign about, will remove.

packages/fxa-auth-server/lib/server.js
```diff
@@ -238,6 +239,55 @@ async function create(log, error, config, routes, db, translator, statsd) {
...
+        } catch (err) {
+          // Unknown accounts will have the default experience
+        }
+      }
+
+      if (!uid) {
+        return true;
+      }
+
+      return Account.metricsEnabled(uid);
```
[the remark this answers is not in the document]
vbudhram: Yea that is correct, it should only get called once on a request.

Issue #10652, opened 2021-10-08
[subplat] Two different behaviours when subscribing via PayPal depending on subscription flow 


data-sync-user:
**Affected versions**:

* All;

**Affected FxA versions**:

* 1.216.2;

**Prerequisites**:

* Be in a country where the Mozilla VPN is available;
* Have an FxA account without a subscription and be signed in;

**Steps to reproduce**:
1. Be on old Set up your subscription page;
2. Observe PayPal button;

**Expected result**:

* The PayPal button is inactive. 

**Actual result**:

* The PayPal button is active.

**Notes**:

* Following the new subscription flow, the PayPal button is inactive until the ““I authorise Mozilla, …” checkbox is checked;
* The user should not be able to make the payment without agreeing with the ToS and PN and both flows should have the same behavior.
* See the attached video;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4070)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16824/Screen+Recording+2021-10-08+at+14.16.12.mov">Screen Recording 2021-10-08 at 14.16.12.mov</a>

data-sync-user:
➤ Reino Muhl commented:

I found this previous issue with attached Figma designs for the “User already has FxA Account” flow. I’ll be making changes to match the Figma design, so that we have the same behaviour for both New and Existing FxA users.

data-sync-user:
➤ Madalin Cotetiu commented:

Verified this as fixed on Train 1.220.0 - Staging. The PayPal button is inactive on old Set up your subscription page.

After agreeing with the ToS and PN, the PayPal button is active.

See the attached screenshots. 

!Screen Shot 2021-11-24 at 11.40.33.png|width=1726,height=1015!

!Screen Shot 2021-11-24 at 12.00.21.png|width=1622,height=926!

PR #11006, opened 2021-11-15
fix(subscription): enable paypal on consent

StaberindeZA:
Because:

* The Paypal button should only be active once the consent checkbox is
  selected.

This commit:

* Moves the consent form in the existing user subscription flow to the
  top of the screen. Matches behaviour and look of new user flow.

Closes #
[10652](https://github.com/mozilla/fxa/issues/10652)

## Checklist

## Screenshots (Optional)

Without consent checked
image

With consent checked
image

### DIFF
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -397,3 +397,5 @@ new-user-submit = Subscribe Now
 
 manage-pocket-title = Looking for your { -brand-name-pocket } premium subscription?
 manage-pocket-body = To manage it, <a>click here</a>.
+
+payment-method-header = { $prefix }Choose your payment method
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -399,3 +399,4 @@ manage-pocket-title = Looking for your { -brand-name-pocket } premium subscripti
 manage-pocket-body = To manage it, <a>click here</a>.
 
 payment-method-header = { $prefix }Choose your payment method
+payment-method-required = Required
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -378,11 +378,9 @@ payment-confirmation-cc-card-ending-in = Card ending in { $last4 }
 ## new user email form
 new-user-sign-in-link = Already have a { -brand-name-firefox } account? <a>Sign in</a>
 new-user-step-1 = 1. Create a { -brand-name-firefox } account
-new-user-step-2 = 2. Choose your payment method
 # "Required" to indicate that the user must use the checkbox below this text to
 # agree to a payment method's terms of service and privacy notice in order to
 # continue.
-new-user-required-payment-consent = Required
 new-user-email =
   .label = Enter your email
 new-user-confirm-email =
@@ -398,5 +396,7 @@ new-user-submit = Subscribe Now
 manage-pocket-title = Looking for your { -brand-name-pocket } premium subscription?
 manage-pocket-body = To manage it, <a>click here</a>.
 
-payment-method-header = { $prefix }Choose your payment method
+payment-method-header = Choose your payment method
+## $prefix (string) - If header is part of a multi step process and needs a header. eg. '2.'
+payment-method-header-prefix = { $prefix } Choose your payment method
 payment-method-required = Required
--- a/packages/fxa-payments-server/public/locales/en-US/main.ftl
+++ b/packages/fxa-payments-server/public/locales/en-US/main.ftl
@@ -397,6 +397,6 @@ manage-pocket-title = Looking for your { -brand-name-pocket } premium subscripti
 manage-pocket-body = To manage it, <a>click here</a>.
 
 payment-method-header = Choose your payment method
-## $prefix (string) - If header is part of a multi step process and needs a header. eg. '2.'
+# $prefix (string) - If header is part of a multi step process and needs a header. eg. '2.'
 payment-method-header-prefix = { $prefix } Choose your payment method
 payment-method-required = Required
--- a/packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
@@ -1,0 +1,105 @@
+import React from 'react';
+import '@testing-library/jest-dom/extend-expect';
+import TestRenderer from 'react-test-renderer';
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import { PaymentMethodHeader } from '.';
+import {
+  getLocalizedMessage,
+  MOCK_PLANS,
+  setupFluentLocalizationTest,
+} from '../../lib/test-utils';
+import { act } from 'react-dom/test-utils';
+import { getLocalizedCurrency } from '../../lib/formats';
+
+jest.mock('../../lib/sentry');
+
+describe('components/PaymentMethodHeader', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+    return cleanup();
+  });
+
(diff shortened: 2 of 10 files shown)


packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -397,3 +397,5 @@ new-user-submit = Subscribe Now
 
 manage-pocket-title = Looking for your { -brand-name-pocket } premium subscription?
 manage-pocket-body = To manage it, <a>click here</a>.
+
+payment-method-header = { $prefix }Choose your payment method
```
chenba, line 401: Probably should leave a comment to explain what `{ $prefix }` is to the localizers.

chenba: And maybe have the space after the prefix here instead of on the prefix itself?

packages/fxa-payments-server/public/locales/en-US/main.ftl
```diff
@@ -397,3 +395,8 @@ new-user-submit = Subscribe Now
 
 manage-pocket-title = Looking for your { -brand-name-pocket } premium subscription?
 manage-pocket-body = To manage it, <a>click here</a>.
+
+payment-method-header = Choose your payment method
+## $prefix (string) - If header is part of a multi step process and needs a header. eg. '2.'
```
chenba, line 400: I think it should be just the one `#` here.  (See https://projectfluent.org/fluent/guide/comments.html)

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    it('Legal text displayed correctly for given plan', async () => {
+      const plan =
+        MOCK_PLANS.find((p) => p.plan_id === 'plan_daily') || MOCK_PLANS[0];
+      const props = { plan, onClick: () => {} };
+      const expectedMsg =
+        'I authorize Mozilla, maker of Firefox products, to charge my payment method <strong>$5.00 daily</strong>, according to <termsOfServiceLink>Terms of Service</termsOfServiceLink> and <privacyNoticeLink>Privacy Notice</privacyNoticeLink>, until I cancel my subscription.';
+
+      const testRenderer = TestRenderer.create(
```
chenba, line 104: Not sure why you switch to `react-test-renderer` here but let's stick to `Testing-Library/react`, which you are already using in the tests above.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  it('render header without prefix', async () => {
+    const plan = MOCK_PLANS[0];
+    const props = { plan, onClick: () => {} };
+    const { queryByTestId } = render(<PaymentMethodHeader {...props} />);
+
+    await waitForExpect(() => {
+      expect(queryByTestId('header')).toBeInTheDocument();
+      expect(queryByTestId('header-prefix')).not.toBeInTheDocument();
+    });
```
chenba, line 31: This is fine.  But FYI, if you wrap code that causes (re-)rendering in `act`, e.g.
```
await act(async () => render(...));
```
then you don't need to use `waitForExpect`.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  prefix?: string;
+};
+
+export const PaymentMethodHeader = ({
+  plan,
+  onClick,
+  prefix = '',
+}: PaymentMethodHeaderProps) => {
+  const checkboxValidator = useValidatorState();
+  const headerMarginTop = '40px';
```
chenba, line 20: Why the inline style?  I don't think that's something we've been doing in this frontend.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <>
+      {prefix ? (
+        <Localized id="payment-method-header-prefix" vars={{ prefix }}>
+          <h2
+            style={{ marginTop: headerMarginTop }}
+            data-testid="header-prefix"
+          >
+            Choose your payment method
```
chenba, line 29: `prefix` isn't used in this fallback string.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
```diff
@@ -164,9 +162,20 @@ describe('routes/Product/SubscriptionCreate', () => {
         />
       );
     });
-    waitForExpect(() =>
-      expect(queryByTestId('paypal-button')).toBeInTheDocument()
+    await waitForExpect(() => {
+      expect(queryByTestId('paypal-button')).toBeInTheDocument();
+    });
```
chenba, line 167: hmm `act` _should_ wait for all rendering to be done so the additional waiting doesn't seem necessary here.  (But I didn't actually debug here so maybe not?)

chenba (changes requested):
Looks good overall.  Got a few minor things that should be fixed.  Please see inline comments.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.tsx
StaberindeZA: Ah good spot, added the prefix to the fallback.

packages/fxa-payments-server/public/locales/en-US/main.ftl
StaberindeZA: I split this out into 1 translation with prefix and one without. The react code then decides which to render depending on the prefix value.

packages/fxa-payments-server/public/locales/en-US/main.ftl
StaberindeZA: Ah cool. Thank you for sharing the link. I had missed that.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
StaberindeZA: Cool updated the test and removed the `react-test-render` stuff.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.tsx
StaberindeZA: I created the `index.scss` file and added the class in there.

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.test.tsx
StaberindeZA: Strangely in this case it was necessary. Not sure why, but when I removed it, the test started failing.

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
StaberindeZA: Ah cool, yup that makes sense. I removed the waitForExpect, and works as...expected (excuse the pun)

packages/fxa-payments-server/src/components/PaymentMethodHeader/index.test.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import '@testing-library/jest-dom/extend-expect';
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import { PaymentMethodHeader } from '.';
+import {
+  getLocalizedMessage,
+  MOCK_PLANS,
+  setupFluentLocalizationTest,
+} from '../../lib/test-utils';
+import { act } from 'react-dom/test-utils';
```
chenba, line 10: I don't think this is the `act` we need; it should be from @testing-library/react

chenba (approved):
☑️  There's an import that I think needs fixing but otherwise looks good!

PR #11013, opened 2021-11-15
Automatic asset upload

dlactin:
## Because

- Uploading assets to the CDN is currently a manual process.

## This pull request

- Github action that will upload assets to stage/prod CDNs when changes to assets/product-icons assets/legal are merged to main. 

## Checklist

This has been tested in stage but the role has not been created in production yet.

### DIFF
--- a/.github/workflows/upload-assets-to-cdn.yml
+++ b/.github/workflows/upload-assets-to-cdn.yml
@@ -1,13 +1,10 @@
-# Upload files in assets/product-icons & assets/legal to S3 buckets for CDN distribution
+# Upload files in assets/product-icons & assets/legal to stage and prod S3 buckets for CDN distribution
 #
 # Requirements:
-# This action requires a set of AWS credentials with write access to the destination bucket directories (product-icons & legal)
-#
-#   Credentials must be passed in as the following secrets:
-#    - CDN_S3_AWS_ACCESS_KEY_ID
-#    - CDN_S3_AWS_SECRET_ACCESS_KEY
+# 1. Github OIDC Provider in destination AWS account - https://github.com/aws-actions/configure-aws-credentials
+# 2. IAM Role with write access to destination bucket paths, scoped to the repository running the action
 
-name: "Asset Upload"
+name: "CDN Asset Upload"
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-assets-to-cdn.yml
```diff
@@ -0,0 +1,45 @@
...
+  Upload-assets-to-CDN:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Configure Stage AWS credentials
+        uses: aws-actions/configure-aws-credentials@master
+        with:
+          aws-region: us-west-2
```
jbuck, line 26: Hmm, I thought stage would be in us-east-1... or does this refer to sts?

jbuck (approved):

jbuck:
@hwine tagging you for additional review, since this is the first time we've implemented this

.github/workflows/upload-assets-to-cdn.yml
dlactin: In this case the S3 bucket & IAM role but both are global, it complained without a region but changing it to us-east-1 for consistency might be good.

hwine:
Like the other one, I'm going to need more context here.

.github/workflows/upload-assets-to-cdn.yml
hwine: S3 is "global-ish" -- if you want to receive S3 events, the receiver has to be in the same region the bucket was configured in. (I got bit on a lambda with this.) So using the origin region might preserve the hair of an sre/dev down the road 😁

dlactin: Sounds good to me! 

Updated stage region to us-east-1.

dlactin:
I have created a slack app that is pending. Will need to create the slack secret when the app has been installed.

.github/workflows/upload-assets-to-cdn.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: "Configure Stage AWS credentials"
+        uses: aws-actions/configure-aws-credentials@master
+        with:
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::142069644989:role/fxa-content-cdn-stage-asset-upload
+          role-session-name: CDNAssetUpload
+
+      - name: "Asset upload to stage CDN S3 bucket"
+        run: |
+          aws s3 cp --cache-control 'public,max-age=86400' --recursive assets/product-icons --exclude "*" --include "*.svg" --include "*.png" s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/product-icons
```
jbuck, line 32: To @hwine's notes, maybe this should be `sync` instead of `cp` (though we should very this won't do anything unexpected the first time we use this)

jbuck:
@hwine also noted we should add CODEOWNERS for these files - we should check with the FxA folks who it'd make sense to put on that list for these files

jbuck:
Current `main` branch protection rules: https://github.com/mozilla/fxa/settings/branch_protection_rules/4907005

We should consider enabling the `Require review from Code Owners ` rule

.github/workflows/upload-assets-to-cdn.yml
dlactin: I verified this works as intended on the stage bucket with a dryrun.

```
aws s3 sync --cache-control 'public,max-age=86400' --exclude "*" --include "*.pdf" --content-disposition attachment  assets/legal s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/legal --dryrun
(dryrun) upload: assets/legal/legal4.pdf to s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/legal/legal4.pdf
(dryrun) upload: assets/legal/test1.pdf to s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/legal/test1.pdf
(dryrun) upload: assets/legal/test2.pdf to s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/legal/test2.pdf
```

```
aws s3 sync --cache-control 'public,max-age=86400' --exclude "*" --include "*.svg" --include "*.png" assets/product-icons  s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/product-icons --dryrun
(dryrun) upload: assets/product-icons/test-icon1.svg to s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/product-icons/test-icon1.svg
(dryrun) upload: assets/product-icons/test-icon2.png to s3://fxa-content-cdn-stage-distbucket-bpquvfnty86g/product-icons/test-icon2.png
```

.github/workflows/upload-assets-to-cdn.yml
```diff
@@ -41,8 +41,8 @@ jobs:
 
       - name: "Asset upload to production CDN S3 bucket"
         run: |
-          aws s3 cp --cache-control 'public,max-age=86400' --recursive assets/product-icons --exclude "*" --include "*.svg" --include "*.png" s3://fxa-content-cdn-prod-distbucket-gqg70i8xqycy/product-icons
-          aws s3 cp --cache-control 'public,max-age=86400' --recursive --content-disposition attachment --exclude "*" --include "*.pdf"  assets/legal s3://fxa-content-cdn-prod-distbucket-gqg70i8xqycy/legal
+          aws s3 sync --cache-control 'public,max-age=86400' --exclude "*" --include "*.svg" --include "*.png" assets/product-icons  s3://fxa-content-cdn-prod-distbucket-gqg70i8xqycy/product-icons
+          aws s3 sync --cache-control 'public,max-age=86400' --exclude "*" --include "*.pdf" --content-disposition attachment   assets/legal s3://fxa-content-cdn-prod-distbucket-gqg70i8xqycy/legal
```
hwine, line 45: Has the business signed off on this? If they want old docs to still link, `cp` was the right thing. I was just asking a question :)

hwine (approved):
lgtm - ideally other branch protections would include:
- include administrators
- Dismiss stale pull request approvals when new commits are pushed

PR #11025, opened 2021-11-16
fix(settings): load the user specified in url on settings

vbudhram:
## Because

- If you navigate to setting when you have multiple accounts logged in, the correct account might not load

## This pull request

- If `uid` is specified in url, load that user in settings if that have a valid local account

Closes: # (issue number)

## Checklist

### DIFF
--- a/packages/fxa-settings/src/lib/cache.ts
+++ b/packages/fxa-settings/src/lib/cache.ts
@@ -1,6 +1,7 @@
 import { InMemoryCache, gql } from '@apollo/client';
 import Storage from './storage';
 import { Email } from '../models';
+import { searchParam } from '../lib/utilities';
 import config from './config';
 
 const storage = Storage.factory('localStorage');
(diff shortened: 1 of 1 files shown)


packages/fxa-settings/src/lib/cache.ts
```diff
@@ -25,6 +27,14 @@ function accounts(accounts?: LocalAccounts) {
 
 export function currentAccount(account?: OldSettingsData) {
   const all = accounts() || {};
+
+  // Current user can be specified in url params (ex. when clicking
+  // `Manage account` from sync prefs.
+  const params = searchParams(window.location.search);
```
dannycoates, line 33: nit: singular `searchParam('uid', window.location.search)`
tangent: why aren't we using `new URL(window.location).searchParams` in utilities.ts?

packages/fxa-settings/src/lib/cache.ts
```diff
@@ -10,6 +11,7 @@ export interface OldSettingsData {
   sessionToken: hexstring;
   alertText?: string;
   displayName?: string;
+  email?: string;
```
dannycoates, line 14: i know it's there but we don't use it. this was meant to be the minimal used subset

dannycoates (approved):

packages/fxa-settings/src/lib/cache.ts
vbudhram: No reason, I'll update it.

Issue #11027, opened 2021-11-16
[subplat] Remove supported languages check from legal-docs redirect endpoint


data-sync-user:
The payments server's legal docs redirect endpoint uses FxA's supported languages to limit the accepted language tags to check for translated docs.  However, the languages for the legal docs no longer aligns with the supported locales for web content.  Let's remove that supported language check from the redirect endpoint.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4204)

PR #11028, opened 2021-11-16
feat(legal-docs): remove supported lang check on redirect

chenba:
Because:
 - the supported languages for translated FxA and Subscription Platform
   web content do not cleanly map to the translated legal docs for
   subscription products

This commit:
 - remove the supported language check from the legal docs redirect

Closes: #11027

### DIFF
--- a/packages/fxa-payments-server/server/lib/routes/legal-docs.js
+++ b/packages/fxa-payments-server/server/lib/routes/legal-docs.js
@@ -10,7 +10,6 @@ const fetch = require('node-fetch');
 const Sentry = require('@sentry/node');
 const { cdnFqdn, resultsCacheLimit } = config.get('legalDocLinks');
 const acceptLanuageParser = require('accept-language-parser');
-const supportedLanguages = require('fxa-shared/l10n/supportedLanguages.json');
 const DEFAULT_LOCALE = 'en';
 const CACHE_DURATION = 24 * 60 * 60 * 1000; // a day in milliseconds
 const LANGUAGES_LIMIT = 3;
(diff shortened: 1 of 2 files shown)


packages/fxa-payments-server/server/lib/routes/legal-docs.js
```diff
@@ -10,7 +10,6 @@ const fetch = require('node-fetch');
 const Sentry = require('@sentry/node');
 const { cdnFqdn, resultsCacheLimit } = config.get('legalDocLinks');
 const acceptLanuageParser = require('accept-language-parser');
-const supportedLanguages = require('fxa-shared/l10n/supportedLanguages.json');
```
biancadanforth, line 13: 1. What is the intended purpose of the `supportedLanguages` list?
From what I can ascertain, the idea is that we add a language to this list when the majority of strings have been localized in Pontoon.

2. Why remove this constraint altogether instead of adding `ms` to the list of `supportedLocales`?
The Malay locale in Pontoon is only 31% translated, so it would probably not be considered a supported language in this case.

3. I think this bug may highlight the larger issue of the fact that we don’t really have sufficient l10n translations for Malay?
I.e. Does this mean that it’s possible for a user in an "unsupported locale" to have other FxA content rendered in a different (supported) locale?E.g. A locale of `ms` would have no matches, and we would fall back to the default English for subscription sign-up, since `supportedLocales` is used by `AppLocalizationProvider`?

biancadanforth (commented):
@chenba I think this is okay short-term, but I have some longer-term questions that are raised by this issue and the proposed fix (inline so we can thread). May be worth syncing up on since this is for a dot release today ideally?

biancadanforth (approved):
R+, but I talked with Barry about this more, and we should discuss the fact that we don't really support this locale in our web content based on Pontoon.

Why are we concerned about localized legal docs, but not for web content, e.g. on FxA Settings or FxA Payments websites?

PR #11058, opened 2021-11-18
refactor(admin): use autogenerated ts types

jodyheavener:
## Because

- We currently manually define our GraphQL Typescript types, but we shouldn't need to 

## This pull request

- Generate them automatically, use them
- Bonus: adds the nest cli as a dev dependency so you don't need it installed on your system to call it
- Bonus: makes use of setupTests feature of Jest to import jest-dom globally


LZoog (approved):
yaaaaaas

dannycoates:
@jodyheavener can you add a workspace dependency to fxa-admin-panel on fxa-admin-server? The ci tests now use the dependency tree to determine which packages need to be tested

PR #11064, opened 2021-11-18
fix(settings): support contextual  translation of  'None'

### LINKED ISSUE #9126: Primary email and Secondary email need different "None" placeholders
Copying over https://github.com/mozilla/fxa-content-server-l10n/issues/472 so we don't miss it in triage:

> Problem: The string None is used in settings for both Primary email and Secondary email. In some languages, these two have two separate genders associated with them, thus None will be translated differently.
>
> [row-defaults-status = None](https://github.com/mozilla/fxa-content-server-l10n/blob/f9fd8c07866f9df13f67b431978241de8b1f3e51/locale/templates/settings.ftl#L379)
>
> Suggestion: Add another string for None, one for each of the emails, with detailed comments/or string ID name so localizers know where the word is used.
> 
> Fixes include new IDs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3081)

dschom:
This applies to primary/secondary email fields.

## Because

- In some languages, the translation of 'none' can differ depending on the context.
- In this case the translation of none as a default for primary email would not be the same as the translation of none as a default  for secondary email.

## This pull request

- Introduces a distinct translation for 'none' when used in the context of secondary email.

closes #9126, closes mozilla/fxa-content-server-l10n/issues/472

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/UnitRowSecondaryEmail/en-US.ftl
+++ b/packages/fxa-settings/src/components/UnitRowSecondaryEmail/en-US.ftl
@@ -30,5 +30,7 @@ se-make-primary = Make primary
 se-default-content = Access your account if you can’t log in to your primary email.
 se-content-note = Note: a secondary email won’t restore your information — you’ll
   need a <a>recovery key</a> for that.
+# Default value for the secondary email
+se-secondary-email-none = None
 
 ##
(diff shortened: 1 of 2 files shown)


LZoog (approved):
This LGTM, will you squash your commits into one and reword it using [these guidelines](https://github.com/mozilla/fxa/blob/main/CONTRIBUTING.md#git-commit-guidelines)? 😁 And also, your original comment can/should have "closes #9126, closes mozilla/fxa-content-server-l10n/issues/472" so when this is merged both of those will automatically close.

(per Slack convo, just reiterating here for others' context) We should probably comment back on the issue too and say that "None" can't be set for a primary email just to give some clarification there. Thanks for working through this!!

chenba:
> ... will you squash your commits into one ...

FWIW we have enabled the "Squash and merge" feature for PRs so you have the option of not doing that manually.

jodyheavener (commented):
LGTM

jodyheavener (approved):

Issue #10865, opened 2021-10-29
[subplat] “Contact support” button from Subscriptions page not working for IAP users


data-sync-user:
**Affected versions**:

* All;

**Affected FxA versions**:

* 1.218.2;

**Prerequisites**:

* Have an FxA account with an active subscription made via IAP Android;

**Steps to reproduce**:
1. Go to https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/subscriptions and sign in with the account from the prerequisites;
2. Click on the “Contact Support” button;
3. Observe the behaviour;

**Expected result**:

* The “Contact Support” form is displayed;

**Actual result**:

* The “Contact Support” form is not displayed. The page is refreshing. 

**Notes**:

* 【 highlighted #d04437 】This a major issue because the IAP users are not able to reach the Contact Support form from the Subscription page.【 end highlighted 】

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4156)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17419/Bildschirmaufnahme+2021-11-03+um+12.41.51.mov">Bildschirmaufnahme 2021-11-03 um 12.41.51.mov</a>

data-sync-user:
➤ Anna Sobiepanek commented:

Madalin Cotetiu can you include a screenshot or video plz

data-sync-user:
➤ Madalin Cotetiu commented:

Hello,

Sure. Attaching a video. 

!Bildschirmaufnahme 2021-11-03 um 12.41.51.mov|width=1920,height=1200!

data-sync-user:
➤ Madalin Cotetiu commented:

Verified this as fixed on Train 1.220.0 - Staging. The Contact support form is displayed when clicking on the “Contact Support” button.

PR #11073, opened 2021-11-18
feature(subscriptions): allow IAP subscribers to use Support Form

chenba:
Because:
 - IAP subscribers should be able to use the web support form

This commit:
 - handle Google Play IAP subscriptions for the support form by using
   the SKU as the plan id in metrics events
 - (the access to the form was granted as a side effect of adding the
   Google Play subscriptions to /account endpoint response)

Closes: #10865

### DIFF
--- a/packages/fxa-content-server/app/tests/spec/views/support.js
+++ b/packages/fxa-content-server/app/tests/spec/views/support.js
@@ -212,6 +212,36 @@ describe('views/support', function () {
         });
     });
 
+    it('uses the SKU in place for plan ID in metrics when given a Google Play Subscription', () => {
+      account.getSubscriptions.restore();
+      sinon.stub(account, 'getSubscriptions').resolves([
+        {
+          product_id: '123done_xyz',
+          product_name: '123Done Pro',
+          sku: 'com.moz.stuff.cool',
+        },
+      ]);
+      return view
+        .render()
+        .then(function () {
+          view.afterVisible();
+          $('#container').append(view.el);
+        })
+        .then(function () {
+          view
+            .$('#product option:eq(1)')
+            .prop('selected', true)
+            .trigger('change');
+          assert.equal(notifier.trigger.callCount, 5);
+          const args = notifier.trigger.args[4];
+          assert.lengthOf(args, 3);
+          assert.equal(args[0], 'subscription.initialize');
+          assert.instanceOf(args[1], SubscriptionModel);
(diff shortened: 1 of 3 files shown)


packages/fxa-content-server/app/tests/spec/views/support.js
```diff
@@ -212,6 +212,36 @@ describe('views/support', function () {
         });
     });
 
+    it('uses the SKU in metrics when given a Google Play Subscription', () => {
```
biancadanforth, line 215: Nit: Maybe note that the SKU is used for the `planId` in particular?

packages/fxa-content-server/app/tests/spec/views/support.js
```diff
@@ -212,6 +212,36 @@ describe('views/support', function () {
...
+          product_id: '123done_xyz',
+          product_name: '123Done Pro',
+          sku: 'com.moz.stuff.cool',
+        },
+      ]);
+      return view
+        .render()
+        .then(function () {
+          view.afterVisible();
+          $('#container').append(view.el);
```
biancadanforth, line 228: Wow, JQuery. Have not seen that in a long time.

biancadanforth (approved):
LGTM. Thanks also for notifying our DS/Data ETL teams for this metrics change.

Not related to this PR, but I couldn’t submit the support form locally -- can you? It looks like `request.app.geo` is `{}`, so it’s throwing an error at this [line](https://github.com/mozilla/fxa/blob/a6d791ed8fc00c48f9e9c226b5b04b7651f4ddfb/packages/fxa-auth-server/lib/routes/subscriptions/support.ts#L137). I'm guessing this is coming from geodb.

packages/fxa-content-server/app/tests/spec/views/support.js
chenba: In theory I agree but practically it's not worth the cost of running CI all over again.

chenba:
> Not related to this PR, but I couldn’t submit the support form locally -- can you? 

No.  That'd require actual Zendesk config to create a support ticket.  There's no such thing as dev Zendesk.

packages/fxa-content-server/app/tests/spec/views/support.js
chenba: Oh, I guess I'll be updating the commit message anyway, so sure, I can update this message along with that.

PR #11084, opened 2021-11-19
slow down pm2 startup

dannycoates:
I've noticed ci failures where the services don't even get started up completely happening more frequently. During the brief time that circleci showed resource graphs the pm2 startup pegged both cpu and memory. If we wait for the services that have a lot of building at startup to finish before starting the next service maybe we'll have better stability.

Screen Shot 2021-11-19 at 8 31 17 AM

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -1,21 +1,25 @@
 module.exports = {
   scripts: {
-    default: "nps help",
+    default: 'nps help',
     start: {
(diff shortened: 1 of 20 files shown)


dannycoates:
omg ci passed on the first run

dannycoates:
are there any other things we should wait on?

chenba (approved):
🥇

PR #11085, opened 2021-11-19
feat(auth): include validation details for product/plan

### LINKED ISSUE #10826: [subplat] Cached plans are not updated when their product's metadata state changes from invalid to valid
When handling a product.updated webhook we iterate through its plans to check whether the product update has made the plans valid or invalid.  _However_, we use the _cached_ plans for that; if the product was never valid while the plans/prices were added, that cached list of plans would be empty.

On a product update event, we must retrieve all of its plans from Stripe to update the plans cache.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4144)

bbangert:
Because:

* We want to update plans when the product information changes.
* We want more details about validation errors in product/plan
  information rather than just the first error.

This commit:

* Uses fetchAllPlans rather than the cached plans to ensure it can
  revalidate all of them when the product changes.
* Uses a new validation error Sentry reporter that extracts complete
  details of validation failures with abortEarly set to false so that
  all validation errors are included at once.

Closes #10826

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -10,16 +10,10 @@ import { Stripe } from 'stripe';
 import Container from 'typedi';
 
 import { ConfigType } from '../../../config';
-import { reportSentryError } from '../../../lib/sentry';
+import { reportSentryError, reportValidationError } from '../../../lib/sentry';
 import error from '../../error';
 import { CapabilityService } from '../../payments/capability';
 import { PayPalHelper } from '../../payments/paypal';
-import { PayPalClientError } from '../../payments/paypal-client';
-import {
-  PAYPAL_BILLING_AGREEMENT_INVALID,
-  PAYPAL_BILLING_TRANSACTION_WRONG_ACCOUNT,
-  PAYPAL_SOURCE_ERRORS,
-} from '../../payments/paypal-error-codes';
 import {
   INVOICES_RESOURCE,
   StripeHelper,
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -6,11 +6,12 @@ import isA from '@hapi/joi';
 import * as Sentry from '@sentry/node';
 import { Account } from 'fxa-shared/db/models/auth';
 import { ACTIVE_SUBSCRIPTION_STATUSES } from 'fxa-shared/subscriptions/stripe';
+import { isArray } from 'lodash';
```
chenba, line 9: VS Code trying to be helpful here?

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -190,4 +190,30 @@ async function configureSentry(server, config, processName = 'key_server') {
...
+ *
+ * @param {*} message
+ * @param {string | import('@hapi/joi').ValidationError} error
+ */
+function reportValidationError(message, error) {
+  const details = {};
+  if (typeof error === 'string') {
+    details.error = error;
+  } else {
+    console.log('error', error);
```
chenba, line 204: I don't think we need this?

chenba (approved):
👍   I think there are a couple lines that can be deleted.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: Looks like it, surprised that remained when it wasn't used.

packages/fxa-auth-server/lib/sentry.js
bbangert: Doh!

Issue #10518, opened 2021-09-23
Cache Stripe Plans Indefinitely


data-sync-user:
Stripe plans are cached for 10 minutes currently to reduce the number of requests to Stripe.  That's not working as well as we hoped.  There's no reason we cannot cache the plans forever and update the cache on Stripe webhook events.  We already have webhook handlers for `product.updated` and `plan.updated` to perform metadata validation; we just need to add to those.  (Listening for `product.created` and `plan.created` won't be helpful since a Stripe admin cannot add a new product or plan with all the required metadata in a single step.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4021)

data-sync-user:
➤ Barry Chen commented:

Benjamin Bangert Please add this to the appropriate epic once you created the epic.  Thanks!

data-sync-user:
➤ Bianca Danforth commented:

IIUC, this will have the nice side effect of making plan/product metadata updates immediate, assuming they pass validation (FXA-3369).

PR #11087, opened 2021-11-20
refactor(auth): cache tax rates in Redis (effectively) indefinitely

### LINKED ISSUE #10745: Cache Tax Rates in Redis
Tax rates are fetched on every check-out, they should be cached in Redis forever like our plans/products are and only updated when changed. Webhook events from Stripe can trigger a Redis refresh of the tax rates.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4117)

biancadanforth:
Because:

* We want to reduce API calls to Stripe to avoid being rate limited.
* Similar to Stripe Plans and Products (#10518), Tax Rates don't change very often.
* Tax rates are unnecessarily fetched from Stripe every time a subscription is created with a TTL of 10 minutes.

This commit:

* Caches tax rates in Redis (effectively) indefinitely.
* Refreshes the cache on tax_rate.created or tax_rate.updated webhooks.

Closes #10745

## Checklist

### DIFF
--- a/packages/fxa-auth-server/config/index.ts
+++ b/packages/fxa-auth-server/config/index.ts
@@ -3,9 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import convict from 'convict';
 import fs from 'fs';
+import { makeMySQLConfig } from 'fxa-shared/db/config';
 import path from 'path';
 import url from 'url';
-import { makeMySQLConfig } from 'fxa-shared/db/config';
 
 const DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages');
 const ONE_DAY = 1000 * 60 * 60 * 24;
@@ -830,7 +830,7 @@ const conf = convict({
     stripeTaxRatesCacheTtlSeconds: {
       doc: 'The number of seconds to cache tax rates from Stripe',
       format: 'int',
-      default: 600,
+      default: 64000000, // about a couple years
       env: 'SUBHUB_TAX_RATES_CACHE_TTL_SECONDS',
     },
   },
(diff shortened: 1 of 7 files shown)


biancadanforth:
Once this merges (presumably into Train 220), I will also need to add the `tax_rate.created` and `tax_rate.updated` events to the list of events we listen for in the Stripe dashboard for stage and prod (once we are closer to deploying to those envs).

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -117,6 +117,10 @@ export class StripeWebhookHandler extends StripeHandler {
         case 'plan.deleted':
           await this.handlePlanDeletedEvent(request, event);
           break;
+        case 'tax_rate.created':
+        case 'tax_rate.updated':
+          await this.handleTaxRateCreatedOrUpdatedEvent(request, event);
+          break;
```
biancadanforth, line 123: Based on the docs and my own examination of what you can do for a tax rate in the Stripe dashboard, there is no `tax_rate.deleted` event.

bbangert: There's also a type-check on this, so trying to put that in would fail to compile if it doesn't exist.

packages/fxa-auth-server/config/index.ts
```diff
@@ -830,7 +830,7 @@ const conf = convict({
     stripeTaxRatesCacheTtlSeconds: {
       doc: 'The number of seconds to cache tax rates from Stripe',
       format: 'int',
-      default: 600,
+      default: 64000000, // about a couple years
```
chenba, line 833: I don't know if we have a value set for this in stage or prod so it'd be good to check.

chenba (approved):
LGTM

packages/fxa-auth-server/config/index.ts
biancadanforth: I checked, and we do not set this env var upstream in mozilla-services/cloudops-deployment, so the default should be the value in all environments.

PR #11111, opened 2021-11-22
fix(settings): add confirm recovery code modal

### LINKED ISSUE #9530: [TOTP] "Confirm recovery code" modal is not displayed when generating new codes (replace)
Prerequisites:

* 2FA is enabled

### Steps to reproduce
1. From 2FA section tap on "Generate recovery codes". 
1. Save the codes and tap on "Continue".

### Expected result
"Confirm recovery code" modal is displayed.

### Actual result
The modal is dismissed. 2FA section is displayed and 2FA is enabled. 

### Environment
* Mobile, Desktop
* Train 154 - stage server

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-905)

dschom:
## Because

- When generating new recovery codes, there was no confirmation step, which might make it easy for a user to overlook the fact they should download / record their recovery codes.

## This pull request

- Introduces the same dialog, which existed in the initial 2FA flow, that prompted a user for at least one recovery code as a way to validate users had recorded them
- Introduces the same dialog, which existed in the initial 2FA flow, that prompted a user for at least one recovery code as a way to validate users had recorded them
- Introduces back end support for creating and updating recovery codes as a two step process
- Shifts recovery code generation to client, relocating crypto/random.ts to fxa-shared, so it's accessible
- Adds validation for recovery codes being updated
- Exposes recovery code length / count settings to client
- Adds section about running play write tests to read me

Closes: #9530

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/recovery-codes.js
+++ b/packages/fxa-auth-server/lib/routes/recovery-codes.js
@@ -6,12 +6,19 @@
 
 const errors = require('../error');
 const isA = require('@hapi/joi');
-const BASE_36 = require('./validators').BASE_36;
+const validators = require('./validators');
+
 const RECOVERY_CODE_SANE_MAX_LENGTH = 20;
 
 module.exports = (log, db, config, customs, mailer) => {
   const codeConfig = config.recoveryCodes;
   const RECOVERY_CODE_COUNT = (codeConfig && codeConfig.count) || 8;
+
+  // Validate recovery codes
+  const recoveryCodesSchema = validators.recoveryCodes(
+    RECOVERY_CODE_COUNT,
+    RECOVERY_CODE_SANE_MAX_LENGTH
+  );
 
   return [
     {
@@ -22,9 +29,7 @@ module.exports = (log, db, config, customs, mailer) => {
           strategy: 'sessionToken',
         },
         response: {
-          schema: {
-            recoveryCodes: isA.array().items(isA.string()),
-          },
+          schema: recoveryCodesSchema,
         },
       },
       async handler(request) {
@@ -66,6 +71,60 @@ module.exports = (log, db, config, customs, mailer) => {
       },
     },
     {
+      method: 'PUT',
+      path: '/recoveryCodes',
+
+      options: {
+        auth: {
+          strategy: 'sessionToken',
+        },
+        validate: {
+          payload: recoveryCodesSchema,
+        },
+        response: {
+          schema: {
+            success: isA.boolean(),
+          },
+        },
+      },
+      async handler(request) {
+        log.begin('updateRecoveryCodes', request);
+
+        const { authenticatorAssuranceLevel, uid } = request.auth.credentials;
+
+        // Since TOTP and recovery codes go hand in hand, you should only be
+        // able to replace recovery codes in a TOTP verified session.
+        if (!authenticatorAssuranceLevel || authenticatorAssuranceLevel <= 1) {
+          throw errors.unverifiedSession();
+        }
+
+        const { recoveryCodes } = request.payload;
+        await db.updateRecoveryCodes(uid, recoveryCodes);
+
+        const account = await db.account(uid);
+        const { acceptLanguage, clientAddress: ip, geo, ua } = request.app;
+
+        await mailer.sendPostNewRecoveryCodesEmail(account.emails, account, {
+          acceptLanguage,
+          ip,
+          location: geo.location,
+          timeZone: geo.timeZone,
+          uaBrowser: ua.browser,
+          uaBrowserVersion: ua.browserVersion,
+          uaOS: ua.os,
+          uaOSVersion: ua.osVersion,
+          uaDeviceType: ua.deviceType,
+          uid,
+        });
+
+        log.info('account.recoveryCode.replaced', { uid });
+
+        await request.emitMetricsEvent('recoveryCode.replaced', { uid });
+
+        return { success: true };
+      },
+    },
+    {
       method: 'POST',
       path: '/session/verify/recoveryCode',
       options: {
@@ -75,11 +134,12 @@ module.exports = (log, db, config, customs, mailer) => {
         },
         validate: {
           payload: {
-            code: isA
-              .string()
-              .max(RECOVERY_CODE_SANE_MAX_LENGTH)
-              .regex(BASE_36)
-              .required(),
+            // Validation here is done with BASE_36 superset to be backwards compatible...
+            // Ideally all recovery codes are Crockford Base32.
+            code: validators.recoveryCode(
+              RECOVERY_CODE_SANE_MAX_LENGTH,
+              validators.BASE_36
+            ),
           },
         },
         response: {
--- a/packages/fxa-content-server/server/lib/configuration.js
+++ b/packages/fxa-content-server/server/lib/configuration.js
@@ -467,6 +467,20 @@ const conf = (module.exports = convict({
     doc: 'The publically visible URL of the deployment',
     env: 'PUBLIC_URL',
   },
+  recovery_codes: {
+    count: {
+      default: 8,
+      doc: 'The default number of recovery codes to create',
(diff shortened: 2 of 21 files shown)


dschom:
@dannycoates In the related issue (#9530), you mentioned there may be back end changes needed. After working on the issue a bit, I do see your point. I am not sure, however, if the back end changes should be a separate request or not. This PR includes the recovery code confirmation dialog. This encourages users to record their recovery codes, so in my opinion, it is still a step forward. Let me know what you think. If you think the back end changes are necessary, I can build on top of this draft.

packages/fxa-auth-server/lib/routes/recovery-codes.js
```diff
@@ -65,6 +65,99 @@ module.exports = (log, db, config, customs, mailer) => {
...
+        validate: {
+          payload: {
+            recoveryCodes: isA.array().items(isA.string()),
+          },
+        },
+        response: {
+          schema: {
+            success: isA.boolean(),
+          }
+        }
```
dannycoates, line 119: I thought about this a bit more...

I think we're actually ok with accepting values from the client as long as we can validate them since we require a verified session, the codes are random and not connected to any other data, and we send a confirmation email. A round trip with a signed JWT is overkill for this.

Our validation should be more thorough than what's here so far. We should ensure they are the correct length and valid base32, no duplicates.

And since we're allowing client values we don't really need a server endpoint to generate them. We can use https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues on the client-side

packages/fxa-content-server/tests/functional/settings/two_step_auth.js
```diff
@@ -169,7 +170,7 @@ registerSuite('TOTP', {
         .then(testElementExists(selectors.ENTER_EMAIL.HEADER));
     },
 
-    'can navigate directly to recovery codes': function () {
+    'can change recovery codes': function () {
```
dannycoates, line 173: congratulations figuring out how to write one of these functional tests 😆

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
```diff
@@ -56,3 +75,42 @@ it('displays an error when fails to fetch new recovery codes', async () => {
   });
   expect(context.alertBarInfo?.error).toBeCalledTimes(1);
 });
+
+it('forces users to validate recovery code', async () => {
+  await renderPage2faReplaceRecoveryCodes();
+
+  await act(async () => {
```
dannycoates, line 82: @jodyheavener will have <strike>opinions</strike> suggestions about this 😉

dannycoates (changes requested):
This is looking great! I like how you implemented the server side really succinctly

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
```diff
@@ -4,42 +4,61 @@
...
+  generateRecoveryCodes: jest.fn().mockResolvedValue(recoveryCodes),
+  updateRecoveryCodes: jest.fn().mockResolvedValue({ success: true }),
 } as unknown as Account;
 
+const config = {
+  recoveryCodes: {
+    count: 1,
+    length: 10,
+  },
+} as unknown as Config;
```
jodyheavener, line 32: I don't think this needs to be `as unknown`

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
jodyheavener: yeah basically you shouldn't need to use `act` 99% of the time - [here's a presentation](https://drive.google.com/file/d/1-M_Dkh1gfJZNTJJYy_CEkDR9UkIp-Dak/view?usp=sharing) explaining why, as well as some other nifty tips

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
```diff
@@ -56,3 +75,42 @@ it('displays an error when fails to fetch new recovery codes', async () => {
   });
   expect(context.alertBarInfo?.error).toBeCalledTimes(1);
 });
+
+it('forces users to validate recovery code', async () => {
+  await renderPage2faReplaceRecoveryCodes();
+
+  await act(async () => {
+    await fireEvent.click(screen.getByTestId('ack-recovery-code'));
```
jodyheavener, line 83: also `fireEvent` isn't async, so you don't need `await` on them

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
```diff
@@ -56,3 +75,42 @@ it('displays an error when fails to fetch new recovery codes', async () => {
...
+  });
+  await act(async () => {
+    await fireEvent.click(screen.getByTestId('submit-recovery-code'));
+  });
+
+  expect(mockNavigate).toHaveBeenCalledWith(
+    HomePath + '#two-step-authentication',
+    { replace: true }
+  );
+});
```
jodyheavener, line 116: `act`-less version:

```tsx
it('allows user to finish', async () => {
  await renderPage2faReplaceRecoveryCodes();

  fireEvent.click(screen.getByTestId('ack-recovery-code'));
  await typeByTestIdFn('recovery-code-input-field')(recoveryCodes[0]);
  fireEvent.click(screen.getByTestId('submit-recovery-code'));

  await waitFor(() =>
    expect(mockNavigate).toHaveBeenCalledWith(
      HomePath + '#two-step-authentication',
      { replace: true }
    )
  );
});
```

packages/fxa-settings/src/lib/random.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        } else {
+          map[r]++;
+        }
+
+        assert.equal(r.length, 10);
+        assert.isTrue(regex.test(r));
+      }
+
+      assert.isFalse(Object.values(map).some((x) => x > 1));
+    }
```
dannycoates, line 120: I think we can trust the rng at this point. We'd need a lot more than 1000 iterations to increase our confidence if we were in doubt.

If you want to ensure `randomValue` is doing the right thing with those random bytes it'd be good to start by mocking `getRandomValues` with a known set of bytes (a test vector) and ensure you get the encoded string(s) you expect.

packages/fxa-auth-server/lib/routes/recovery-codes.js
```diff
@@ -6,13 +6,21 @@
 
 const errors = require('../error');
 const isA = require('@hapi/joi');
-const BASE_36 = require('./validators').BASE_36;
+const validators = require('./validators');
+
+const RECOVERY_CODE_SANE_MAX_COUNT = 20;
```
dannycoates, line 11: I think we can just use `RECOVERY_CODE_COUNT` from below

packages/fxa-content-server/server/lib/configuration.js
```diff
@@ -467,6 +467,20 @@ const conf = (module.exports = convict({
     doc: 'The publically visible URL of the deployment',
     env: 'PUBLIC_URL',
   },
+  recovery_codes: {
+    count: {
+      default: 3,
```
dannycoates, line 472: let's default this to the production value of 8 and set it in `server/config/local.json` as 3

dannycoates (approved):
This looks good @dschom 🎉 ... just a couple tiny things

dannycoates:
was this closed on purpose?

dschom:
It was not closed intentionally! I have corrected the problem. I _think_ this was the result of a botched push. I re-based off main and effectively undid my changes by accident :disappointed:. I think that with zero file changes, github then marked the PR as closed.

PR #11115, opened 2021-11-22
chore(deps): add dependencies on other workspaces for runtime dependency

dannycoates:
We use workspace dependencies to determine which test suites to run when code changes. Our services have runtime dependencies on each other, not necessarily code dependencies. Adding these runtime dependencies to package.json ensures we can test all dependent services, and maybe someday automate which services start when integration testing.


packages/fxa-payments-server/package.json
```diff
@@ -126,7 +126,7 @@
     "dayjs": "^1.9.4",
     "eslint-plugin-jest": "^24.5.2",
     "express": "^4.17.1",
-    "fxa-content-server": "workspace:*",
```
dannycoates, line 129: this would create a cyclic dependency. content-server functionally depends on payments-server, currently all the functional tests are there, while payments-server only depends on some css which it accesses directly, i.e. `../../.......`. Both the css dep and the functional tests all being under content-server are issues for another day

vbudhram (approved):

PR #11119, opened 2021-11-22
chore(dev): wait for containers on 'yarn start'

dannycoates:
On M1 we currently sometimes need to 'yarn restart services' after a 'yarn start' before things start working. Maybe if we wait for the containers to be responsive before starting services we can avoid this. The important one seems to be mysql.

### DIFF
--- a/_dev/pm2/infrastructure.config.js
+++ b/_dev/pm2/infrastructure.config.js
@@ -47,13 +47,15 @@ module.exports = {
       autorestart: false,
       kill_timeout: 20000,
     },
-    {
-      name: 'pushbox',
-      script: '_scripts/pushbox.sh',
-      max_restarts: '1',
-      min_uptime: '2m',
-      args: '3306 root@mydb:3306',
-      kill_timeout: 20000,
-    },
+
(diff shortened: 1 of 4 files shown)


_dev/pm2/infrastructure.config.js
```diff
@@ -47,13 +47,15 @@ module.exports = {
...
-    {
-      name: 'pushbox',
-      script: '_scripts/pushbox.sh',
-      max_restarts: '1',
-      min_uptime: '2m',
-      args: '3306 root@mydb:3306',
-      kill_timeout: 20000,
-    },
+
+    // TODO FIXME?? this has been broken for-ev-er, do we even want it?
```
chenba, line 51: 🤷   Fixing it mainly involves removing the `args` below correct?

chenba (approved):
💯

_dev/pm2/infrastructure.config.js
dannycoates: the pushbox image needs an overlay network between it and the mysql instance. i think we should just drop it

PR #11135, opened 2021-11-23
feat(admin-panel): replace majority of admin-panel with tailwind styles

jodyheavener:
## Because

- We want to start using tailwind styles in the admin-panel

## This pull request

- Moves most everything over to Tailwind. The one notable exception is the `Account` component and everything within it.

## Screenshots

Screen Shot 2021-11-23 at 3 42 23 PM

- This is not an exact 1:1 switch as Tailwind introduced different spacing values, colors, etc. But it's close, and doesn't look too bad IMO. Open to feedback of course.
- ~Our reusable `Header` component is not set up to receive CSS style props so I didn't include our header bg image. I could be persuaded to hack it up though.~ nvm I added it
- The change to Tailwind easily afforded us some mobile styles :)


dannycoates (approved):
LGTM 😍

LZoog (approved):
This looks so much better 👌 

I think we could maybe add a small text shadow on the heading text to make it slightly easier to read (I think the colors just scream "white doesn't work on that for ADA!"), but I checked the Tailwind docs and I don't think that comes out of the box. Will leave it up to you on if it's worth adding or not since it's an internal tool.

Thanks also for fixing that grey box below the blocked account search and icon! I know I filed an issue for that but can't seem to find it.

PR #11139, opened 2021-11-23
fix(auth): remove payment method from Firestore on payment_method.det…

### LINKED ISSUE #11126: [subplat] payment_method.detached webhook permafails
Since the prod deploy of Train 219, we are storing Stripe Payment Methods as a subcollection in Firestore under Stripe Customers ([FXA-4118]). That patch introduced a bug where, in the webhook handler, we pass `null` as the `customerId` into `fetchAndInsertCustomer` when backfilling the Customer object.

While the webhook `event.object` has a non-null `customer`, we are using the `customer` off the detached `paymentMethod` [fetched directly from Stripe](https://github.com/mozilla/fxa/blob/a7460d79d6935395b3b79e5a76d92762470fe331/packages/fxa-auth-server/lib/payments/stripe.ts#L2485), which is `null`.

Per [~accountid:5e3400fb6bd5c40ca395d189], this is not an urgent bug, but it is annoying due to our Slack alerts for webhook failures:
> ...the webhook should be updated to find the payment method by id (we have a group collection query that can find it without customer), then delete it, for those. otherwise there's no problem.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4253)

biancadanforth:
…ached webhook

Because:

* Currently, we handle the payment_method.detached webhook event the same as other payment_method* events in terms of Firestore caching.
  * Practically, this means we try to insert a detached Payment Method into Firestore with a null customerId, which is causing these webhook events to permafail.
* The detached payment method should be removed from Firestore, as it will no longer be used by the associated Stripe customer.

This commit:

* Removes the detached payment method from Firestore on payment_method.detached webhooks.

Closes #11126

## Checklist

I didn't rename the existing `processPaymentMethodEventToFirestore`, since it still covers several `payment_method.*` webhook events, and the name would be too long to list them all explicitly. Open to alternatives.

### DIFF
--- a/packages/fxa-auth-server/test/local/payments/stripe.js
+++ b/packages/fxa-auth-server/test/local/payments/stripe.js
@@ -69,6 +69,8 @@ const subscriptionCreatedInvoice = require('./fixtures/stripe/invoice_paid_subsc
 const eventInvoiceCreated = require('./fixtures/stripe/event_invoice_created.json');
 const eventSubscriptionUpdated = require('./fixtures/stripe/event_customer_subscription_updated.json');
 const eventCustomerUpdated = require('./fixtures/stripe/event_customer_updated.json');
+const eventPaymentMethodAttached = require('./fixtures/stripe/event_payment_method_attached.json');
+const eventPaymentMethodDetached = require('./fixtures/stripe/event_payment_method_detached.json');
 const closedPaymementIntent = require('./fixtures/stripe/paymentIntent_succeeded.json');
 const newSetupIntent = require('./fixtures/stripe/setup_intent_new.json');
 const {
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -3957,6 +3959,44 @@ describe('StripeHelper', () => {
       });
     }
 
+    for (const type of [
+      'payment_method.attached',
+      'payment_method.automatically_updated',
+      'payment_method.updated',
+    ]) {
```
biancadanforth, line 3966: I didn't see tests for the Firestore processing for `payment_method.*` webhooks in general, so I added some.

chenba (commented):
The code looks fine, but I think Ben want to save the old payment methods in Firestore.  We'll see what he says.

chenba (approved):
💳

biancadanforth:
> The code looks fine, but I think Ben want to save the old payment methods in Firestore. We'll see what he says.

Thanks Barry. As I mentioned in Slack, Ben had suggested we delete these, as I don't think the same payment method can be re-attached to the customer again.

Issue #11097, opened 2021-11-21
Panel: display template type with email bounces


data-sync-user:
Once FXA-4240 is done add this new field to the getAccountByEmail/getAccountByUid queries and display it in each email bounce record.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4241)

Issue #11096, opened 2021-11-21
Server: retrieve template type from emailBounces table


data-sync-user:
Once FXA-4214 is done we should be able to look up the template type and return it on the account query, which returns email bounces.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4240)

Issue #11056, opened 2021-11-18
Add email template to `emailBounces` table


data-sync-user:
We can add a new column to store the template that was bounced. This might be useful information to have for the admin panel.

1. Add a new migration https://github.com/mozilla/fxa/blob/1d1c630c1313197d9b587716d9cb5b5a56c7bd64/packages/fxa-auth-db-mysql/scripts/migration.sh
2. Pass email template name https://github.com/mozilla/fxa/blob/233de634204547fbe166f066faa46ebc1837ec13/packages/fxa-auth-server/lib/email/bounces.js#L159
3. Extend `EmailBounces` to create with template name, https://github.com/mozilla/fxa/blob/6ab822f7af055539377be145ef27085c2bcebde6/packages/fxa-shared/db/models/auth/email-bounce.ts#L44

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4214)

PR #11163, opened 2021-11-25
feat(db): add emailTypeId column to emailBounces, start recording templates, display in admin panel

jodyheavener:
## Because

- We would like to start recording the name of the template for the email that was sent when an email bounce occurs

## This pull request

- Add a new `emailTypeId` column to `emailBounces` in the DB
- Updates DB model to create records with template name, start recording template name when creating the bounce record
- Update random email bounce generator helper
- Add template to GQL type, display template in admin panel

Closes: #11056
Closes: #11096
Closes: #11097

## Checklist

## Screenshots

Screen Shot 2021-11-25 at 9 44 03 AM


dannycoates:
side-note: I hope we aren't considering [not-spam](https://www.rfc-editor.org/rfc/rfc6430.html) as blockable, but i don't see anything in the code indicating we handle it any differently

packages/fxa-admin-server/src/graphql.ts
```diff
@@ -34,6 +34,7 @@ export enum BounceSubType {
 
 export interface EmailBounce {
     email: string;
+    template: string;
```
vbudhram, line 37: I'm not really sure how to do this with gql, but we have an `emailTypes` table that has all our templates already. We might be able to use that instead of storing a string.

packages/fxa-shared/db/models/auth/email-bounce.ts
```diff
@@ -37,20 +38,28 @@ export class EmailBounce extends BaseAuthModel {
...
     email,
+    templateName,
     bounceType,
     bounceSubType,
   }: Pick<EmailBounce, 'email'> & {
+    templateName: string;
     bounceType: BounceType;
     bounceSubType: BounceSubType;
   }) {
+    const emailTypes = EmailType.bindKnex(this.knex);
```
jodyheavener, line 56: Feeling a little uncertain about this guy here 🤔

packages/db-migrations/databases/fxa/patches/patch-122-123.sql
```diff
@@ -0,0 +1,53 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('122');
+
+ALTER TABLE emailBounces
+ADD COLUMN emailTypeId BINARY(32) DEFAULT NULL,
+ALGORITHM = INPLACE, LOCK = NONE;
+
+CREATE PROCEDURE `createEmailBounce_2` (
```
vbudhram, line 9: Do we need `createEmailBounce_2`? I don't see it used in the code.

packages/fxa-shared/db/models/auth/email-bounce.ts
vbudhram: Me to 😅. Could you use the stored procedure to create the bounce instead? I'm not sure what the standard is since I noticed it is done both ways in other areas.

vbudhram (changes requested):
@jodyheavener This LGTM, minus the binary column and questions.

packages/db-migrations/databases/fxa/patches/patch-122-123.sql
jodyheavener: Whoops, I had missed updating it from `createEmailBounce_1`, it's used [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/email/bounces.js#L159) (and now in the fxa-shared model)

vbudhram (approved):
@jodyheavener :shipit:

Issue #11130, opened 2021-11-23
service parameter not passed from the fxa content server 


YeonjooSmith:
## Description
I was testing the acquisition funnel for the VPN and check if a bug(#10838) on the CTA was fixed.  
I went through the steps to subscribe VPN with a couple of different ways yesterday and check the flow ids, I see that the service parameter is not logged. It was passed before when I tested it in October but it is not anymore so something might have been changed. 

### Why is it important
the service parameter is the indicator to filter VPN customers from FxA data to track the user journey. The auth server and the payment server does not pass the service parameter so the service parameter from the content server is only one that let me weave all the event logs related to the VPN service from different servers. 

### Steps to reproduce
[Here](https://docs.google.com/document/d/1aBVBs_f-pt6XYQNdgUmI4E7-6et6nSMMfpajVHGhrWQ/edit?usp=sharing) is the steps that I went through with the url that includes flow id in each step to test the funnel in 11/21/2021. Also this document has links for the query and query results. 
 
### Expected result
From my testing funnel results in October([query here](https://sql.telemetry.mozilla.org/queries/83017/source), [query result in the spreadsheet](https://docs.google.com/spreadsheets/d/1HwB_zAEk8dBwTWNa9E3XzaFRLcXA_ndJI2X8PsWRl_8/edit?usp=sharing)), the content server should have the service parameter as `guardian-vpn`.  

### Actual result
The service parameter missing. Please refer to the doc link in **Steps to reproduce**. 

### Environment
included in the doc

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4255)

YeonjooSmith:
cc:@vbudhram

data-sync-user:
➤ Bianca Danforth commented:

Yeonjoo Yoo : I want to make sure I understand the issue here – so when you are an existing FxA user, logged out and going through the passwordless checkout flow, you do not see the service query param as expected in each of the URLs in the steps you describe here ( https://docs.google.com/document/d/1aBVBs_f-pt6XYQNdgUmI4E7-6et6nSMMfpajVHGhrWQ/edit# )?

Vijay Budhram Are you working on this already? Just want to make sure we don’t duplicate our efforts here.

YeonjooSmith:
[...]

PR #11180, opened 2021-11-29
fix(metrics): update content-server regex to match the auth-server regex

vbudhram:
## Because

- Our auth-server and content-server have different validation for the service param. Ref https://github.com/mozilla/fxa/blob/e81f5e077cee59ec0c0bbc5bcd78931358b7be18/packages/fxa-auth-server/lib/routes/validators.js#L96

## This pull request

- Updates content-server to use auth-server validaiton which is more generic

Closes: #https://github.com/mozilla/fxa/issues/11130

## Checklist

## Other information

We will probably want to put this on a point release if it is a blocker for metrics analysis.

### DIFF
--- a/packages/fxa-content-server/server/lib/validation.js
+++ b/packages/fxa-content-server/server/lib/validation.js
@@ -26,7 +26,7 @@ const PATTERNS = {
   EXPERIMENT: /^[\w.-]+$/,
   FORM_TYPE: /^(email|other|button|subscribe)$/,
   PRODUCT_ID: /^prod_[0-9A-Za-z]+$/,
-  SERVICE: /^(sync|content-server|none|[0-9a-f]{16})$/,
+  SERVICE: /^([a-zA-Z0-9\-]{1,16})$/,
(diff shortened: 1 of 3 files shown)


chenba (approved):

packages/fxa-content-server/server/lib/validation.js
```diff
@@ -26,7 +26,7 @@ const PATTERNS = {
   EXPERIMENT: /^[\w.-]+$/,
   FORM_TYPE: /^(email|other|button|subscribe)$/,
   PRODUCT_ID: /^prod_[0-9A-Za-z]+$/,
-  SERVICE: /^(sync|content-server|none|[0-9a-f]{16})$/,
+  SERVICE: /^(sync|content-server|none|[a-zA-Z0-9\-]{1,16})$/,
```
dschom, line 29: I noticed the auth server allows an empty service name. This seems more correct though. Perhaps the `sync|content-server|none` are redundant?

PR #11181, opened 2021-11-29
feat(auth): include product/plan validation details in Sentry message

### LINKED ISSUE #10827: handlePlanUpdatedEvent - Plan "price_1JmROfJNcmPzuWtR6od8OfDW"'s metadata failed validation
[https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/14057704/?referrer=github_plugin](https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/14057704/?referrer=github_plugin)

```java
handlePlanUpdatedEvent - Plan "price_1JmROfJNcmPzuWtR6od8OfDW"'s metadata failed validation
```

Until we have UI for [https://mozilla-hub.atlassian.net/browse/FXA-3655,](https://mozilla-hub.atlassian.net/browse/FXA-3655,) this will still be an issue, so we should update the error message in the meantime to say which key/value failed validation. This is already contained in the `error` object's `message` prop.

~~Ensure that Joi doesn’t abort early on error.  The validation error details are potentially in the ‘details’ prop.  If available, include all the error detail in the Sentry error name.~~ (handled by FXA-4144)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4145)

biancadanforth:
Because:

* RPs may enter invalid product/plan metadata into Stripe, and the only feedback are Sentry errors in #fxa until a better solution is made following FXA-3655.
* The Sentry error message is currently not descriptive regarding what the issue is, and RPs may not have access to Sentry to view more error details.

This commit:

* Includes as much actionable information in the Sentry error message about the validation error as possible, up to the ~100 character limit.

Closes #10827 

## Checklist

The error message limited to around 100 characters in Slack, and given that a product or plan id can have around 30 characters, this doesn't leave much room for more descriptive messages.

E.g. a plan update with multiple validation errors would have a Sentry message like: 'price_1JfcqJBVqmGyQTMadYoEH25q metadata invalid: "webIconURL" must be a valid …'
I’m not sure how I can trim this further without causing confusion, but I’m open to suggestions. I did try doubling the `maxValueLength` option in `Sentry.init()`.

Also, FWICT Sentry already truncates error messages, so I didn't include a test for really long messages.

### DIFF
--- a/packages/fxa-auth-server/lib/sentry.js
+++ b/packages/fxa-auth-server/lib/sentry.js
@@ -161,6 +161,8 @@ async function configureSentry(server, config, processName = 'key_server') {
         new Sentry.Integrations.LinkedErrors({ key: 'jse_cause' }),
         new ExtraErrorData({ depth: 5 }),
       ],
+      // https://docs.sentry.io/platforms/node/configuration/options/#max-value-length
+      maxValueLength: 500,
     });
     Sentry.configureScope((scope) => {
       scope.setTag('process', processName);
@@ -191,6 +193,26 @@ async function configureSentry(server, config, processName = 'key_server') {
 }
 
 /**
+ * Format a Stripe product/plan metadata validation error message for
+ * Sentry to include as much detail as possible about what metadata
+ * failed validation and in what way.
+ *
+ * @param {string} planId
+ * @param {string | import('@hapi/joi').ValidationError} error
+ */
+function formatMetadataValidationErrorMessage(planId, error) {
+  let msg = `${planId} metadata invalid:`;
+  if (typeof error === 'string') {
+    msg = msg + ' ' + error;
+  } else {
+    for (const errorItem of error.details) {
+      msg = msg + ' ' + errorItem.message + ';';
+    }
+  }
+  return msg;
+}
+
+/**
  * Report a validation error to Sentry with validation details.
  *
  * @param {*} message
@@ -215,4 +237,9 @@ function reportValidationError(message, error) {
   });
 }
 
-module.exports = { configureSentry, reportSentryError, reportValidationError };
+module.exports = {
+  configureSentry,
+  reportSentryError,
+  reportValidationError,
+  formatMetadataValidationErrorMessage,
+};
--- a/packages/fxa-auth-server/lib/sentry.js
+++ b/packages/fxa-auth-server/lib/sentry.js
@@ -203,11 +203,11 @@ async function configureSentry(server, config, processName = 'key_server') {
 function formatMetadataValidationErrorMessage(planId, error) {
   let msg = `${planId} metadata invalid:`;
   if (typeof error === 'string') {
-    msg = msg + ' ' + error;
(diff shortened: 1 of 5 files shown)


jodyheavener:
fwiw that failure is happening in another PR, so I think you can ignore it here

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -161,6 +161,8 @@ async function configureSentry(server, config, processName = 'key_server') {
         new Sentry.Integrations.LinkedErrors({ key: 'jse_cause' }),
         new ExtraErrorData({ depth: 5 }),
       ],
+      // https://docs.sentry.io/platforms/node/configuration/options/#max-value-length
```
chenba, line 164: Funny that their doc on the max value length does not mention a max.  I guess 500 is fine then?

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -190,6 +192,26 @@ async function configureSentry(server, config, processName = 'key_server') {
...
+ */
+function formatMetadataValidationErrorMessage(planId, error) {
+  let msg = `${planId} metadata invalid:`;
+  if (typeof error === 'string') {
+    msg = msg + ' ' + error;
+  } else {
+    for (const errorItem of error.details) {
+      msg = msg + ' ' + errorItem.message + ';';
+    }
+  }
```
chenba, line 211: Any reason for string concatenation with `+` here instead of template literals?

packages/fxa-auth-server/lib/sentry.js
```diff
@@ -190,6 +192,26 @@ async function configureSentry(server, config, processName = 'key_server') {
...
+ * @param {string | import('@hapi/joi').ValidationError} error
+ */
+function formatMetadataValidationErrorMessage(planId, error) {
+  let msg = `${planId} metadata invalid:`;
+  if (typeof error === 'string') {
+    msg = msg + ' ' + error;
+  } else {
+    for (const errorItem of error.details) {
+      msg = msg + ' ' + errorItem.message + ';';
+    }
```
chenba, line 210: code style comment: ``error.details.map(x => ` ${x};`).join('')`` instead of a for loop.

chenba (approved):

packages/fxa-auth-server/lib/sentry.js
biancadanforth: True. I only know 250 is the default. I imagine 500 which is the same order of magnitude should be all right... 🤔

packages/fxa-auth-server/lib/sentry.js
biancadanforth: I can do this; just need to add the message prefix and destructure the `error.details` object, which is itself an object literal.

chenba: oh haha, yeah... thanks for catch my bug in my comment 😓

PR #11184, opened 2021-11-30
fix(bulk-mailer): date normalization was wrong

dannycoates:
## Because

- `getUTCMonth` is 0 based
- `setUTCMonth` rolls over if the day of the month beforehand is larger than the target can handle


vbudhram (approved):
Oof 😂

PR #11189, opened 2021-11-30
feat(admin-panel): convert Account component to Tailwind classes

### LINKED ISSUE #11146: Panel: convert Account component to Tailwind classes
We need to finish converting the FxA Admin Panel to Tailwind styles, with the last major change being the [Account](https://github.com/mozilla/fxa/blob/main/packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx) component.

Converting this means replacing all the styles in its [style file](https://github.com/mozilla/fxa/blob/main/packages/fxa-admin-panel/src/components/AccountSearch/Account/index.scss) with Tailwind utility classes to create a similar look. It won't be an exact switch since Tailwind introduces colors, spacing, etc that are different from how the Panel looks now.

Once converted you should also be able to entirely remove the [base](https://github.com/mozilla/fxa/blob/main/packages/fxa-admin-panel/src/styles/_base.scss) and [variables](https://github.com/mozilla/fxa/blob/main/packages/fxa-admin-panel/src/styles/_variables.scss) global style files, as well as all the styles (excluding imports) in the [root](https://github.com/mozilla/fxa/blob/main/packages/fxa-admin-panel/src/index.scss) style file. After this change there should theoretically be no additional custom styles.

Since this is just style changes, unless you are restructuring HTML to suit the styles, there should be no test changes necessary for this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4261)

xlisachan:
Because:

* We want to utilize Tailwind styles in admin-panel

This commit:

* Converts Account component styles to Tailwind

Closes #11146

## Checklist

## Screenshots (Optional)
Before and After
Screen Shot 2021-11-30 at 9 35 22 AM

I don't think Tailwind has the border-image property, so the linear gradient is now grey, please let me know if another color should be used instead. In addition, as there is a separate issue for the Account History section, I can work on those updates and make a separate pull request.


packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx
```diff
@@ -195,64 +207,75 @@ export const Account = ({
...
       <ul>
-        <li className="flex justify-space-between">
-          <h3 data-testid="email-label">
+        <li className={`${styleClasses.li} flex justify-between`}>
+          <h3 data-testid="email-label" className="mt-0 my-0 mb-1 text-lg">
             <span
-              className={`${query === primaryEmail.email ? 'highlight' : ''}`}
+              className={`${
+                query === primaryEmail.email ? 'bg-yellow-100' : ''
+              }`}
```
LZoog, line 217: Another small nit: if we set an empty class name to an empty string like this in React, the "class" attribute is rendered on the DOM element with an empty string. If you set it to `undefined` instead, it'll just render `<span>` without the attribute.

Also, similarly to the other comment, I believe this can be simplified to:

`className={query === primaryEmail.email ? 'bg-yellow-100' : undefined}`

You could also consider using [classnames](https://www.npmjs.com/package/classnames) which we have in other packages but eh 🤷‍♀️

packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx
```diff
@@ -342,67 +371,61 @@ export const Account = ({
             ))}
           </>
         ) : (
-          <li data-testid="" className="gradient-info-display">
+          <li
+            data-testid=""
```
LZoog, line 375: If we're not testing these, we can just remove the `data-testid` attributes set to empty strings.

packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx
```diff
@@ -38,6 +37,14 @@ type DangerZoneProps = {
 
 const DATE_FORMAT = 'yyyy-mm-dd @ HH:MM:ss Z';
 
+const styleClasses = {
+  enabledOrVerified: 'font-semibold text-green-900',
+  notEnabledOrVerified: 'font-semibold text-red-600',
+  gradientInfoDisplay: 'border-l-2 border-grey-500 mb-8 pl-4',
```
LZoog, line 43: I think since this is no longer a gradient we could call `gradientInfoDisplay` something else 🤔

It also still sorta feels like we could possibly use `@apply` for `.account li`, but since we don't have any other custom SCSS now in the admin panel after this PR, it also feels fine to leave it how you have it.

packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx
```diff
@@ -195,64 +207,75 @@ export const Account = ({
...
+              primaryEmail.isVerified
+                ? styleClasses.enabledOrVerified
+                : styleClasses.notEnabledOrVerified
             }`}
           >
             {primaryEmail.isVerified ? 'verified' : 'not verified'}
           </span>
         </li>
-        <li className="flex justify-space-between">
+        <li className={`${styleClasses.li}`}>
```
LZoog, line 233: Tiny nit: instead of ```={`${styleClasses.li}`}``` (and similarly on other lines like L225 - L227, L235, and others), they can be `={styleClasses.li}`.

packages/fxa-admin-panel/src/components/AccountSearch/Account/index.tsx
```diff
@@ -342,67 +371,61 @@ export const Account = ({
...
+            <TableContainer component={Paper}>
+              <Table className="pt-1" aria-label="simple table">
+                <TableHead>
+                  <TableRow>
+                    <TableCell>Event</TableCell>
+                    <TableCell>Timestamp</TableCell>
+                  </TableRow>
+                </TableHead>
+                <TableBody>
+                  {securityEvents.map((securityEvents: SecurityEventsType) => (
```
LZoog, line 396: I know you didn't change this but a quick win, I think this is going to complain since we're not [setting a `key`](https://reactjs.org/docs/lists-and-keys.html) on the mapped `TableRow`s.

LZoog (approved):
This looks great!! I just had a few small nit picks but not blocking r+ 🙂

Issue #11047, opened 2021-11-17
Always send email for specific email templates


data-sync-user:
From [https://docs.google.com/document/u/1/d/1lsqv7-rctFyUiNn8RfaVuDQnu6_Xiv57CzhSF5MPhuU/edit](https://docs.google.com/document/u/1/d/1lsqv7-rctFyUiNn8RfaVuDQnu6_Xiv57CzhSF5MPhuU/edit), we decided that one thing we can do is always send emails that might prevent a user from getting access to their account regardless if it causes a bounce.

These emails template names are

* `verifyLoginCode`
  * Email recieved when user logs in and we send them a code to verify
* `unblockCode`
  * User has been rate limited and can reset the rate limit by entering the code we sent
* `subscriptionAccountFinishSetup`
  * Email sent after user creates an account via the passwordless flow. This email allows them to set a password
* `recovery`
  * Password reset email

It might be worthwhile to pull this into the bounces configuration for auth-server.

We can check for these templates at [https://github.com/mozilla/fxa/blob/41e8f9bb1be9e076ef2df335ebb7a62ae1fac868/packages/fxa-auth-server/lib/senders/email.js#L513](https://github.com/mozilla/fxa/blob/41e8f9bb1be9e076ef2df335ebb7a62ae1fac868/packages/fxa-auth-server/lib/senders/email.js#L513|smart-link) and then check the bounces as needed. We will also need to log how often these bounces and templates get skipped.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4212)

PR #11195, opened 2021-11-30
Don't bounce emails that might stop user access

vbudhram:
## Because

- We don't want to stop user from getting access to their account because of a bounced email

## This pull request

- Skips the bounce checks if it is a login verification emaill, password reset email, rate limiting unblock email or subscription service finish account setup email

Closes https://github.com/mozilla/fxa/issues/11047

## Checklist


LZoog (approved):
LGTM

Issue #11209, opened 2021-12-01
TypeError: Cannot read property 'toLowerCase' of undefined


vbudhram:
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/15590115/?referrer=github_plugin

```
TypeError: Cannot read property 'toLowerCase' of undefined
  File "/fxa/node_modules/@hapi/hawk/lib/crypto.js", line 69, in Object.exports.generateNormalizedString
    options.host.toLowerCase() + '\n' +
  File "/fxa/node_modules/@hapi/hawk/lib/crypto.js", line 46, in Object.exports.calculateMac
    const normalized = exports.generateNormalizedString(type, options);
  File "/fxa/node_modules/@hapi/hawk/lib/server.js", line 291, in Object.exports.header
    const mac = Crypto.calculateMac('response', credentials, artifacts);
  File "/fxa/node_modules/@hapi/hawk/lib/plugin.js", line 99, in Object.listener
    const header = Server.header(request.auth.credentials, request.auth.artifacts, { hash: Crypto.finalizePayloadHash(payloadHash) });
  File "/fxa/node_modules/@hapi/podium/lib/index.js", line 221, in module.exports.internals.Podium.emit
    const result = handler.context ? handler.listener.apply(handler.context, args) : handler.listener(...args);
...
(8 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4277)

PR #11210, opened 2021-12-01
fix(metrics): Don't override hapi request artifacts

vbudhram:
## Because

- We were overriding hapi request artifacts when setting metrics uid to be used later

## This pull request

- Adds the uid to object

Closes: https://github.com/mozilla/fxa/issues/11209

## Checklist

## Notes

Targeted as a point release on train-220.

### DIFF
--- a/packages/fxa-auth-server/lib/metrics/events.js
+++ b/packages/fxa-auth-server/lib/metrics/events.js
@@ -103,8 +103,8 @@ module.exports = (log, config) => {
 
       const request = this;
 
-      if (data && data.uid) {
-        request.auth.artifacts = { metricsUid: data.uid };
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/metrics/events.js
```diff
@@ -103,8 +103,8 @@ module.exports = (log, config) => {
 
       const request = this;
 
-      if (data && data.uid) {
-        request.auth.artifacts = { metricsUid: data.uid };
+      if (data && data.uid && request.auth && request.auth.artifacts) {
```
LZoog, line 106: typescript where you at

LZoog (approved):
👍

PR #11212, opened 2021-12-02
fix(settings): fix copy and paste of recovery codes

### LINKED ISSUE #11042: Select + copy of 2FA recovery codes does not include any white space between codes
[From Bugzilla 1741179](https://bugzilla.mozilla.org/show_bug.cgi?id=1741179)

When enabling 2FA from https://accounts.firefox.com/settings, the recovery codes are presented to the user. If the codes are selected and copied, the result is a concatenated string of all the codes, rather than the newline-separated string that clicking on the copy button that the site's UI provides.

Some explicit whitespace should be included in the HTML structure between the `<span>` elements wrapping each code to ensure that the results are at least legible.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4208)

dschom:
## Because:

* When a user would manually copy recovery codes, codes would be mashed together without any newline or white space to separate them.

## This pull request

- Introduces a trailing whitespace on recovery codes, which allows copy and paste to be more user friendly.
- Hijacks onCopy event, which allows white space to be converted to newlines.

Closes #11042

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-settings/src/components/DataBlock/index.tsx
+++ b/packages/fxa-settings/src/components/DataBlock/index.tsx
@@ -19,12 +19,16 @@ type actionFn = (action: actions) => void;
 export type DataBlockProps = {
   value: string | string[];
   prefixDataTestId?: string;
+  separator?: string;
+  onCopy?: (event: React.ClipboardEvent<HTMLDivElement>) => void;
   onAction?: actionFn;
 };
 
 export const DataBlock = ({
   value,
   prefixDataTestId,
+  separator,
+  onCopy,
   onAction = () => {},
 }: DataBlockProps) => {
   const valueIsArray = Array.isArray(value);
@@ -47,11 +51,13 @@ export const DataBlock = ({
           valueIsArray ? 'py-4' : 'py-5'
         }`}
         data-testid={dataTestId}
+        {...{onCopy}}
       >
         {valueIsArray ? (
           (value as string[]).map((item) => (
             <span key={item} className="flex-50% py-1">
               {item}
+              {separator || ''}
             </span>
           ))
         ) : (
--- a/packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
+++ b/packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
@@ -36,6 +36,10 @@ it('renders', async () => {
   expect(screen.getByTestId('2fa-recovery-codes')).toHaveTextContent(
     recoveryCodes[0]
   );
+
+  expect(screen.getByTestId('datablock').textContent?.trim()).toEqual(
+    recoveryCodes.join(' ')
+  );
 });
 
 it('displays an error when fails to fetch new recovery codes', async () => {
--- a/packages/fxa-settings/src/components/DataBlock/index.test.tsx
+++ b/packages/fxa-settings/src/components/DataBlock/index.test.tsx
@@ -21,11 +21,11 @@ const multiValue = [
   'D4J6KY8FL4',
 ];
 
-const account = ({
+const account = {
   primaryEmail: {
     email: 'pbooth@mozilla.com',
   },
-} as unknown) as Account;
+} as unknown as Account;
 
 Object.defineProperty(window.navigator, 'clipboard', {
   value: { writeText: jest.fn() },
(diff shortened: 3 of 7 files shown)


packages/fxa-settings/src/components/DataBlock/index.tsx
```diff
@@ -19,12 +19,16 @@ type actionFn = (action: actions) => void;
 export type DataBlockProps = {
   value: string | string[];
   prefixDataTestId?: string;
+  separator?:string,
+  onCopy?: (event:any) => void,
```
jodyheavener, line 23: I like to try and avoid `any` where possible; `React.ClipboardEvent<HTMLDivElement>` should work here. You'll also need to update the function to `copyRecoveryCodes(event: React.ClipboardEvent<HTMLDivElement>)`

packages/fxa-settings/src/components/DataBlock/index.tsx
```diff
@@ -47,11 +51,12 @@ export const DataBlock = ({
           valueIsArray ? 'py-4' : 'py-5'
         }`}
         data-testid={dataTestId}
+        onCopy={(e) => { onCopy && onCopy(e) }}
```
jodyheavener, line 54: Since `onCopy` is already optional on HTMLElements we should be able to just do `{...onCopy}` here

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
```diff
@@ -36,6 +36,8 @@ it('renders', async () => {
   expect(screen.getByTestId('2fa-recovery-codes')).toHaveTextContent(
     recoveryCodes[0]
   );
+
+  expect(screen.getByTestId('datablock').textContent?.trim()).toEqual(recoveryCodes.join(" "))
```
jodyheavener, line 40: This is more of a personal preference, but it's also shorter: instead of using a testid you could look up each code by its text using `multiValue.forEach((v) => screen.getByText(v))` - this more closely represents what a user sees.

Not a blocker, though!

jodyheavener (changes requested):
Couple nitpicks; would love to replace those `any`s. Otherwise LGTM!

Side note: do you have Prettier enabled in your IDE? We don't have post-commit auto-formatting enabled for fxa-settings, but we do have a prettier config that will auto-format on save or paste if you enable it. I see a few spots were it would format for ya.

packages/fxa-settings/src/components/DataBlock/index.tsx
dschom: Thanks, I'll change this. I might have some bad habits of doing non-short hand things. Keep this feedback coming :smile:

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
dschom: I guess the reason I did it this way is that I was explicitly checking that the 'textContent' had received spaces as per the separator prop. This is the crux of the issue.

packages/fxa-settings/src/components/DataBlock/index.tsx
dschom: Good call!

dschom:
>Side note: do you have Prettier enabled in your IDE?

Ah good point about prettier. I have it installed, but I don't think it was auto formatting on save. Are you good with the entire file getting formatted on save or would you rather just the changes?

packages/fxa-settings/src/components/Page2faReplaceRecoveryCodes/index.test.tsx
jodyheavener: 🤦 of course - my bad

jodyheavener:
> Are you good with the entire file getting formatted on save or would you rather just the changes?

Ideally others have been running formatting on this file up to now, so formatting the entire file shouldn't result in a huge diff (if it does you might double check that it's actually using our prettier config and not some default one), but yeah I'm okay with formatting the whole thing

packages/fxa-settings/src/components/DataBlock/index.tsx
```diff
@@ -47,11 +51,13 @@ export const DataBlock = ({
           valueIsArray ? 'py-4' : 'py-5'
         }`}
         data-testid={dataTestId}
+        onCopy={onCopy}
```
jodyheavener, line 54: Not a blocker, but I know @LZoog would point this out: you can de-dupe by spreading the props (`{...{ onCopy }}` instead of `onCopy={onCopy}`)

jodyheavener (approved):
Nicely done.

packages/fxa-settings/src/components/DataBlock/index.tsx
dschom: Definitely better style. Resolved.

PR #11213, opened 2021-12-02
feat(auth): add invoice preview route

### LINKED ISSUE #10854: Coupons - [API] Update Create Subscription API
Update the existing create subscription API. The new API should should include the coupon details. The api should return coupon specific error messages.

Add metadata to the create subscription metrics: 

* for all of the events add coupon code/id to the event properties.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4153)

### LINKED ISSUE #10856: Coupons - [API] Create preview invoice API
Add a new API that talks to Stripe. This API should check to ensure the coupon is valid (not maxed out, not expired, available for that product). If the coupon is invalid the api should return a coupon specific code. If the coupon is valid the api should return some metadata about the coupon (expiry date, name, amount of discount) to be used in the coupon component. As well as the subscription price details ( price, discount, tax, total due) 

Add metrics to track coupon usage:

coupon_submit: fire when user presses the apply button

coupon_error: fire when coupon validation failed. Capture failure type with coupon code used

coupon_success: fire when user presses apply and coupon is valid

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4155)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17297/Screen+Shot+2021-10-26+at+4.02.02+PM.png">Screen Shot 2021-10-26 at 4.02.02 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/17298/Screen+Shot+2021-10-26+at+4.04.17+PM.png">Screen Shot 2021-10-26 at 4.04.17 PM.png</a>

bbangert:
Because:

* We want to preview coupon code application to an invoice in the
  checkout flow.

This commit:

* Adds an invoice preview route that takes promotion codes and returns
  an invoice preview suitable for the checkout display.

Closes #10856

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe-formatter.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe-formatter.ts
@@ -1,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import * as invoiceDTO from 'fxa-shared/dto/auth/payments/invoice';
+import { Stripe } from 'stripe';
+
+/**
+ * Formats a Stripe Invoice to the InvoicePreview DTO format.
+ */
+export function stripeInvoiceToInvoicePreviewDTO(
+  invoice: Stripe.Invoice
+): invoiceDTO.InvoicePreview {
+  const invoicePreview: invoiceDTO.invoicePreviewSchema = {
+    subtotal: invoice.subtotal,
+    total: invoice.total,
+    line_items: invoice.lines.data.map((line) => ({
+      amount: line.amount,
+      currency: line.currency,
+      id: line.price!.id,
+      name: line.description || '',
+    })),
+  };
+
+  // Add tax if it exists
+  if (
+    invoice.total_tax_amounts.length > 0 &&
+    invoice.default_tax_rates.length > 0
+  ) {
+    const taxObject = invoice.default_tax_rates[0];
+    const tax = invoice.total_tax_amounts[0];
+    invoicePreview.tax = {
+      amount: tax.amount,
+      inclusive: tax.inclusive,
+      name: taxObject.display_name,
+      percentage: taxObject.percentage,
+    };
+  }
+
+  // Add discount if it exists
+  if (invoice.discount && invoice.total_discount_amounts) {
+    invoicePreview.discount = {
+      amount: invoice.total_discount_amounts[0].amount,
+      amount_off: invoice.discount.coupon.amount_off,
+      percent_off: invoice.discount.coupon.percent_off,
+    };
+  }
+  return invoicePreview;
+}
--- a/packages/fxa-shared/package.json
+++ b/packages/fxa-shared/package.json
@@ -112,7 +112,8 @@
     "reflect-metadata": "^0.1.13",
     "rxjs": "^7.2.0",
     "stripe": "^8.191.0",
-    "superagent": "^6.1.0"
+    "superagent": "^6.1.0",
+    "typesafe-joi": "^2.1.0"
   },
   "jest": {
     "moduleFileExtensions": [
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -71,7 +71,12 @@ enum STRIPE_CUSTOMER_METADATA {
   PAYPAL_AGREEMENT = 'paypalAgreementId',
 }
 
+export enum STRIPE_PRICE_METADATA {
+  PROMOTION_CODES = 'promotionCodes',
+}
+
 export enum STRIPE_PRODUCT_METADATA {
+  PROMOTION_CODES = 'promotionCodes',
   PLAY_SKU_IDS = 'playSkuIds',
 }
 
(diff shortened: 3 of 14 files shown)


bbangert:
Preview of using this locally:
```bash
 curl -X POST http://localhost:9000/v1/oauth/subscriptions/invoice/preview  \
      -H 'Content-Type: application/json' \
      -d '{"priceId":"plan_GqM9N6qyhvxaVk", "promotionCode": "FRIEND20"}'

{"subtotal":500,"total":1,"line_items":[{"amount":500,"currency":"usd","id":"plan_GqM9N6qyhvxaVk","name":"1 × 123Done Pro (at $5.00 / month)"}],"discount":{"amount":499,"amount_off":499,"percent_off":null}}
```

bbangert:
This PR sets up a slightly different arrangement for APIs and response objects (Data-Transfer Objects, aka DTOs) that are used by the payment-server front-end. API responses used by the payment server have their types in fxa-shared, and there's functions in auth-server that format responses to these objects. 

The way this was previously done  suffered from a variety of problems:
- It was ad-hoc and grew organically without structure
- Poorly named (many have 'Abbrev' prefixes yet they're not abbreviated really)
- Bled into internal auth-server objects that were not always being returned to the front-end
- Tied to Stripe API types, which hampers Stripe SDK upgrades when types change, but our response schema doesn't need to.
- Had response validation schemas in different sections of the code-base making it error-prone to update/alter them.
- Had no type-safe method to enforce that the validation schema matched the DTO type
- Had no type-safe match of the response object to the validation schema.

This method addresses all of the above concerns for this PR, future refactors will be needed for the other DTOs used between auth-server and payment-server. It does this by:

- Provides a file structure for DTOs returned by an FxA service, in this case only auth-server (auth) so far.
- Avoids Abbrev names and relies on file structure and namespace naming with DTO to designate DTO types/schemas
- Separates DTO formatting into a separate module.
- Uses no Stripe API types, instead replicating the underlying structure where needed.
- Uses typesafe-joi and keeps the validation schema next to the DTO it validates.
- Has a formatter function that returns the DTO type, while the handler has an explicit return type of the DTO validation schema type, this enforces type alignment between the DTO type the front-end will use, and the validation schema of the response from the auth-server.

Thoughts, questions?

vbudhram:
> Thoughts, questions?

I'm generally onboard with using this pattern. It kind of knudges us to use more typescript in the auth-server. In my experience with DTO, they can become cumbersome when you have a lot of different entiies. I don't believe that is the case with FxA, but maybe subscription services?

packages/fxa-shared/package.json
```diff
@@ -112,7 +112,8 @@
     "reflect-metadata": "^0.1.13",
     "rxjs": "^7.2.0",
     "stripe": "^8.191.0",
-    "superagent": "^6.1.0"
+    "superagent": "^6.1.0",
+    "typesafe-joi": "^2.1.0"
```
biancadanforth, line 116: [`typesafe-joi`](https://www.npmjs.com/package/typesafe-joi) has a warning label on it saying it’s a WIP without unit tests -- do we want to use this package in a production application? What are the risks?

I do see we are already using it in fxa-event-broker and fxa-support-panel...

biancadanforth (commented):
My review is limited to the layout change for how we format and validate our responses. I will defer to @IvoJP to review the PR for the coupons work, since he will likely be the one to make use of this new endpoint, and I'm not actively working on the Coupons epic at the moment. (Ivo, let me know if/how I could be of assistance though.)

I see the influence of Martin Fowler in this approach! How would we handle migrating the existing routes to use this approach? Can you document this change and/or file tickets accordingly as a follow-up to ensure this happens?

By the way, I think Jira borked the GitHub issue syncing here, as both FXA-4153 and FXA-4155 link to #10854 (I edited FXA-4155 just now), but actually the issue you should be referencing in your PR is #10856 . In addition to updating your commit body, you may want to check that FXA-4155 closes after this PR merges.

packages/fxa-auth-server/lib/payments/stripe-formatter.ts
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import * as invoiceDTO from 'fxa-shared/dto/auth/payments/invoice';
+import { Stripe } from 'stripe';
+
+export function stripeInvoiceToInvoicePreviewDTO(
+  invoice: Stripe.Invoice
+): invoiceDTO.InvoicePreview {
+  const invoicePreview: invoiceDTO.invoicePreviewSchema = {
```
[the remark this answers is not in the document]
bbangert: > Is it possible to specify the response format with TS types instead of Joi to get rid of the redundancy and risk altogether? E.g. IIUC to update other routes, devs would have to remember to use both the DTO type and DTO schema in the approach you have here.

I considered that, but the typesafe-joi object type is not quite a normal interface, which is what the payments server works with. This does unfortunately rely on ensuring the implementation of this pattern returns the interface in one context, while the handler function itself has a return type of the Joi validation schema type, which ensures they structurally match.

packages/fxa-shared/package.json
bbangert: Yea, its worked well enough with no issues so far, but its possible a more complex validation schema might not work with it.

IvoJP (approved):

PR #11214, opened 2021-12-02
chore(settings): add test coverage thresholds

### LINKED ISSUE #10774: Settings: Set line coverage test threshold, increase test coverage (+3-5%?) to meet
We don't currently have a jest test threshold set for fxa-settings. At the time of issue filing, we've got 65% of lines covered which could be worse but I think we should aim a little higher.

I propose we should set jest `coverageThreshold` lines to 68-70% and set CI to fail if we go under. We should have a follow up issue to increase the coverage and threshold further once this is done and to maybe call out important tests we're lacking, and have some discussion on where we ultimately want to end up if that doesn't happen on this issue (75% - 90%?). Also totally open to suggestions on if we want to hit other kinds of thresholds beyond just line coverage.

We could also consider not counting some `utils` files in our coverage.

Run `yarn test --coverage --watchAll` in `fxa-settings` to see this.

image

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4135)

dschom:
## Because

- We want to increase test coverage.

## This pull request

- Adds configuration for test coverage thresholds.

Closes #10774

## Checklist

### DIFF
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -16,7 +16,8 @@
     "stop": "pm2 stop pm2.config.js",
     "delete": "pm2 delete pm2.config.js",
     "storybook": "npm run build-postcss && start-storybook -p 6008 -s .storybook/design-guide --no-version-updates",
-    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test"
+    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test",
+    "test:coverage": "yarn test --coverage --watchAll=false"
   },
   "jest": {
     "resetMocks": false,
@@ -24,8 +25,32 @@
       "src/**/*",
       "!src/**/*.stories.*",
       "!src/**/mocks.*",
-      "!src/**/*.svg"
-    ]
+      "!src/**/*.svg",
+      "!src/index.*",
+      "!src/components/*(ScrollToTop|UnitRowRecoveryKey|PageAvatar|App)/*",
+      "!src/lib/*(cache.ts|canvas-utils.tsx|file-utils.tsx|firefox.ts|totp.ts|gql.ts)",
+      "!src/models/*"
+    ],
+    "coverageThreshold": {
+      "global": {
+        "branches": 0,
+        "functions": 0,
+        "lines": 0,
+        "statements": 0
+      },
+      "src/components/**/*": {
+        "branches": 50,
+        "functions": 25,
+        "lines": 65,
+        "statements": 50
+      },
+      "src/lib/**/*": {
+        "branches": 50,
+        "functions": 90,
+        "lines": 60,
+        "statements": 60
+      }
+    }
   },
   "eslintConfig": {
     "extends": [
--- a/packages/fxa-settings/package.json
+++ b/packages/fxa-settings/package.json
@@ -22,33 +22,19 @@
   "jest": {
     "resetMocks": false,
     "collectCoverageFrom": [
-      "src/**/*",
+      "src/**/*.{ts,tsx,js}",
       "!src/**/*.stories.*",
       "!src/**/mocks.*",
-      "!src/**/*.svg",
-      "!src/index.*",
-      "!src/components/*(ScrollToTop|UnitRowRecoveryKey|PageAvatar|App)/*",
-      "!src/lib/*(cache.ts|canvas-utils.tsx|file-utils.tsx|firefox.ts|totp.ts|gql.ts)",
-      "!src/models/*"
+      "!src/components/*(ScrollToTop|UnitRowRecoveryKey|PageAvatar|App)/*"
     ],
     "coverageThreshold": {
       "global": {
-        "branches": 0,
-        "functions": 0,
-        "lines": 0,
-        "statements": 0
+        "branches": 50,
+        "lines": 45
       },
       "src/components/**/*": {
         "branches": 50,
-        "functions": 25,
-        "lines": 65,
-        "statements": 50
-      },
-      "src/lib/**/*": {
-        "branches": 50,
-        "functions": 90,
-        "lines": 60,
-        "statements": 60
+        "lines": 65
       }
     }
   },
--- a/packages/fxa-settings/README.md
+++ b/packages/fxa-settings/README.md
@@ -499,7 +499,7 @@ All logging methods have the argument `eventProperties`, which can be used to su
 
 ## Testing and Mocks for Tests/Storybook
 
-This package uses [Jest](https://jestjs.io/) to test its code. By default `yarn test` will test all JS files under `src/`.
+This package uses [Jest](https://jestjs.io/) to test its code. By default `yarn test` will test all JS files under `src/`. Running `yarn test:coverage` will also provide a coverage report, which should be respected.
 
 Test specific tests with the following commands:
 
--- a/packages/fxa-settings/scripts/test-ci.sh
+++ b/packages/fxa-settings/scripts/test-ci.sh
@@ -5,7 +5,7 @@ DIR=$(dirname "$0")
 cd "$DIR/.."
 
 yarn build
-SKIP_PREFLIGHT_CHECK=true yarn test
+SKIP_PREFLIGHT_CHECK=true yarn test --coverage --watchAll=false
 
 mkdir -p config
 


packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
     "stop": "pm2 stop pm2.config.js",
     "delete": "pm2 delete pm2.config.js",
     "storybook": "npm run build-postcss && start-storybook -p 6008 -s .storybook/design-guide --no-version-updates",
-    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test"
+    "test": "SKIP_PREFLIGHT_CHECK=true rescripts test",
+    "test:coverage": "yarn test --coverage --watchAll=false"
```
LZoog, line 20: Ooo thanks for this and the doc update 😏

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
     "collectCoverageFrom": [
       "src/**/*",
       "!src/**/*.stories.*",
       "!src/**/mocks.*",
-      "!src/**/*.svg"
-    ]
+      "!src/**/*.svg",
+
+      "!src/index.*",
+      "!src/components/*(ScrollToTop|UnitRowRecoveryKey|PageAvatar|App)/*",
```
LZoog, line 31: Are we ignoring these because we need to increase coverage on them to reach the threshold set, and the idea is to file a follow up to increase coverage and remove this line later?

Same questions for `models/`, I think we would want coverage on at least the exported class `Account` right?

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
+      "!src/lib/*(cache.ts|canvas-utils.tsx|file-utils.tsx|firefox.ts|totp.ts|gql.ts)",
```
LZoog, line 32: This makes me think we should have a `utils` directory that we should move relevant files to and ignore instead 🤔

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
+      "!src/models/*"
+    ],
+    "coverageThreshold": {
+      "global": {
```
LZoog, line 36: I kind of wonder if we should just have global coverage here and not differing coverage for other parts, though I'm totally open to discussions if it makes sense to separate them. I'm only referencing this since I was recently rolled off this project but we did that with [Experimenter's nimbus-ui](https://github.com/mozilla/experimenter/blob/main/app/experimenter/nimbus-ui/package.json#L22-L34) and it worked just fine to set the global number (though I definitely don't think FxA needs 100% line coverage here).

Did you have opinions on what we should strive for as far as branches, functions, lines, and statements go in the future? I think we might be fine only setting lines and branches. Part of what closes that issue is kind of coordinating / gathering some folk's opinions on our best way to move forward with this for future issues and maintenance items to increase test coverage, though it's probably sufficient if you, me, and @jodyheavener (tagging since I'm sure he will have opinions on this PR) agree on expectations here.

packages/fxa-settings/package.json
dschom: Yeah, I was thinking these would need to improved one by one, because they had very low test coverage, and this would have really lowered the overall thresholds.

In my opinion, the models should not have logic in them, so there wouldn't be anything to test. Maybe that's a different maintenance item though.

packages/fxa-settings/package.json
dschom: I like that idea. To be fair though, I think some of these modules do have code that should be tested. Sort of like the ignored components, these had very low coverage and I didn't want to lower the overall bar.

packages/fxa-settings/package.json
dschom: There is one thing to note about the global config. It seems to combine all files together that aren't specified by subsequent glob paths and report on that. The issue I had with this is you no longer see the individual files that are in violation. For example, let's say file A had 100% coverage and file B has 10% coverage. Globally you would still hit a 50% threshold. This was a bit a surprising to me, but it's what I observed when tinkering with the config.

I don't have much of an opinion on different types of coverage. I'd tend to agree with you that we should focus on branches and lines. Functions are probably the biggest point of contention. For example, you might have a bunch of wrappers that are basically logic less... do they need to be tested, not really sure? Some people might still want a mock or spy as test coverage. In general, I really like the idea of standardizing folder structure and coming up with minimums based on that.

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
+        "branches": 0,
+        "functions": 0,
+        "lines": 0,
+        "statements": 0
+      },
+      "src/components/**/*":  {
```
jodyheavener, line 42: Nipick: running `yarn` will auto-format this file and remove any excess spaces for ya

packages/fxa-settings/package.json
jodyheavener: The models here perform operations such as making calls to the [auth client](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/Account.ts#L540:L540) and storing/retrieving data from [apollo](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/models/hooks.ts#L37), so I don't think it would hurt to have coverage count against them.

packages/fxa-settings/package.json
jodyheavener: Yeah, I think we should be testing most of these. But I understand we're excluding to bring the percentage up.

packages/fxa-settings/package.json
jodyheavener: I don't have strong opinions here. My only problem with the individual configs is that when I look at them I have no idea why those numbers were chosen.

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
   },
   "jest": {
     "resetMocks": false,
     "collectCoverageFrom": [
       "src/**/*",
       "!src/**/*.stories.*",
       "!src/**/mocks.*",
-      "!src/**/*.svg"
-    ]
+      "!src/**/*.svg",
```
jodyheavener, line 28: Was it including SVGs? Could whitelist instead of blacklisting, `"src/**/*.{js,jsx,ts,tsx}",` on the first line

packages/fxa-settings/package.json
```diff
@@ -16,16 +16,42 @@
...
+
+      "!src/index.*",
```
jodyheavener, line 30: Is this intended to be removed later once tests are written? I think we could test this file

jodyheavener (commented):
I don't have super strong opinions; while I agree we should try to test reasonably and thoroughly, I don't think I completely agree with mandatory coverage percentages. I almost wish, instead of CI failing, that we had a bot to yell at and shame you in the PR if you bring scores down, but still allow you to merge.

packages/fxa-settings/package.json
dschom: Can we pull those ops into a different space so that our models are pure?

packages/fxa-settings/package.json
dschom: I think it's mostly to get more targeted thresholds that can be improved over time. As @LZoog points out the original issue, we should have a maintenance task to improve threshold on a weekly basis until a minimum is hit. Eventually they should all get to 80% (or whatever we deem good enough), and no files should be excluded. I spent a decent amount of time tinkering with these values until I got them ‘close’ to what the average minimum could be at this point in time.
 
I’d also reiterate that I am not a fan of _just_ using ‘global’ coverage, since the coverage is averaged across all files, and the threshold hold error won’t call out specific files. IMO one global config is too simplistic, as not all files are equal in complexity, importance, or risk of breaking.

LZoog: > The issue I had with this is you no longer see the individual files that are in violation. For example, let's say file A had 100% coverage and file B has 10% coverage. Globally you would still hit a 50% threshold.

Hmmm, do you mean that you don't see a clear list of component names? I might be misunderstanding, but when I run it, I just have to vertically scan the "%" columns and see what's in red to see which components need increased coverage to meet thresholds.

I hear you on a global config possibly being too simplistic though.

> Eventually they should all get to 80% (or whatever we deem good enough)

Yeah I'm curious on what thresholds we'd want to aim for, though, it's not a simple answer. Maybe we should have a separate task (a spike?) to go through our files and determine which areas we deem the most important to have coverage on and decide from there what percentages we think are best and only set thresholds for line coverage and branches in the meantime if none of us feel strongly that we should have required coverage for functions/statements?

LZoog:
@jodyheavener 
>  I don't think I completely agree with mandatory coverage percentages. I almost wish, instead of CI failing, that we had a bot to yell at and shame you in the PR if you bring scores down

Lol, that would be fun. Yeah, that's one reason why I slapped that `needs:discussion` label on the other issue, it's good to welcome any and all opinions. What I hope to achieve with this is to help us ensure we're at an acceptable level of unit test coverage (duh, but) because it can be easy to rush to get something in, not test thoroughly at all, and push that coverage off until later and then maybe never get around to it. It's easy also to be unaware your test coverage is low when you write tests and think you're g2g, or, to not realize a test isn't hitting the lines you think it is. There's pros and cons for sure.

packages/fxa-settings/package.json
l0oDANo0l: That was already there... I can clean this up though.

packages/fxa-settings/package.json
l0oDANo0l: Yeap! Couldn't get us to reasonable coverage numbers without excluding some files.

packages/fxa-settings/package.json
dschom: >Hmmm, do you mean that you don't see a clear list of component names?

Yes, part of my gripe is not seeing a clear list of component names, and you are right it’s easy to just look at the report; however, the other thing I don’t care for is how 'global' averages the files together. This is useful in some situations, but not all. Here's an example where I think just using global results in an undesirable result:

Let’s say we decide 80% is an acceptable threshold, but we are currently sitting at 90% coverage. With only globals in place, I can see a new component being added with no tests. The threshold check would still pass because the other tests are basically compensating for this untested component. This is why I think at a minimum we should decide what the per file threshold is for certain folders, e.g. components and lib. This seems like a good way to ensure new code is continually put under test. I find it's much easier to write the tests upfront than after the fact, which is another reason why I have this opinion.

>...in the meantime if none of us feel strongly that we should have required coverage for functions/statements?

I am totally fine with only enforcing coverage on lines and branches. I’d also say that I am even okay with enforcing coverage on components, and doing global for everything else, but this is certainly a debatable stance. @LZoog  @jodyheavener if you are good with this, I can change the file accordingly.

packages/fxa-settings/package.json
dschom: @jodyheavener This was actually already in place, and isn't really part of the PR (the line change is just a trailing comma). I can clean this up though... To be honest, it seems unneeded, since checks should only be done on code, but I can double check this. 

I suspect whitelisting won't do what you want in conjunction with global coverage, but I guess this is what we are hashing out, at the moment..

packages/fxa-settings/package.json
dschom: Yep, I am not adding new tests just setting the thresholds, and this file had no test coverage if I recall.

dschom:
@jodyheavener We don't necessarily have to fail the CI job if coverage fails. I can think of ways around this. Furthermore a coverage report is generated, so we _could_ run a bot on these reports. This would all be more involved though.

Personally, I like mandatory test coverage, and it is the default behavior. In my mind, it's easier to write the tests up front than weeks later. Also, a test that may only take minutes to write, can save a lot of time when you consider all the ceremony surrounding a bug fix.

dschom:
@jodyheavener @LZoog, I've simplified the configuration, please check it out again and let me know your thoughts. I think this seems like a good middle ground given the comments made thus far. Changes are as follows:

- Global is used
- Components are tested individually
- Some components are still ignored due to low coverage. Improving this should be part of an ongoing maintenance process.
- We will work towards higher levels of coverage. Perhaps instead of deciding on a target number now, we should try and increase it by 5% every week until we think there are diminishing returns.

jodyheavener (approved):
I still don't have sure strong opinions here other than I don't think I believe in mandatory coverage %, but I understand I could be in the minority. I also appreciate that components are split out to their own coverage.

> We will work towards higher levels of coverage. Perhaps instead of deciding on a target number now, we should try and increase it by 5% every week until we think there are diminishing returns.

I don't mind this approach. I would be happy to volunteer a couple hours each week to plug away at a component and bring its tests up. Might be a nice break from other work. I'm not sure how we should measure this progress, if it needs individual tickets and clear definition, or we just kind of do it as we see fit and communicate it through team updates.

dschom:
@LZoog @jodyheavener I've added two follow up tasks: #11340 and #11339

Issue #10763, opened 2021-10-19
Settings a11y - Google play and App store images need alt text + darkened grey text


LZoog:
When using a screenreader, only "Link, opens in new window" is read.

In `ConnectAnotherDevicePromo`, let's add alt text matching what's on the images. Don't forget to localize it.

Also, the `text-grey` class used on the "Find Firefox in the Google Play and App Store or" doesn't pass ADA AA against the white background. It should be `text-grey-400`.

<img width="1065" alt="image" src="https://user-images.githubusercontent.com/13018240/137959937-24e2d304-473a-4e08-80dd-ab14fabf295f.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4127)

Issue #9363, opened 2021-06-21
[FxA-Settings] Badges text not translated in the corresponding language


data-sync-user:
**Build:**
Stage (Train 1.208)

**Preconditions:**
No mobile device is connected.

**Affected platforms:**
Windows 10 64-bit
macOS 10.14
Android 11
iOS 14.4

**Steps to reproduce:**

1. Launch Firefox using a fr, ge, ar, sq, cy, etc. build (or set the desired language from about:preferences -> Choose your preferred language for displaying pages).
1. Navigate to https://accounts.stage.mozaws.net and log in using an existing account.
1. Go to “Connected Services” section and observe the badges.

**Actual result:** 
Badges text is not translated in the corresponding language.

**Expected result:** 
Badges text is translated in the corresponding language.

**Note:**
Compare with “Connect another device” page.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3653)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/15176/DE+-+badges+not+translated.PNG">DE - badges not translated.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/15177/FR+-+badges+not+translated.PNG">FR - badges not translated.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

Reino Muhl Lauren Zugai I verified this issue on Stage (Train 222.0) under Windows 10 64-bit for all languages that are displayed as 100% translated in Pontoon ( https://pontoon.mozilla.org/projects/firefox-accounts/ ). The badges are not translated for all these languages. I added the results in this gdoc ( https://docs.google.com/document/d/14pYviP8m3A7Wzcf_1Di9dY5TFJL_mlfm5VHVU5Km5pc/edit ) (again I verified only the languages that appear as 100% translated).

PR #11229, opened 2021-12-02
fix(settings): add translated store images

StaberindeZA:
## Because:

* Currently only English language images are shown for the Google Play
  and Apple App store badges.
* The badges do not have translated alt texts.

## This pull request:

* Loads the image with the correct language, depending on the browsers
  selected language.
* Adds translated text to the image alt attribute.

Closes #
#9363
#10763 

## Checklist


packages/fxa-settings/src/assets/images/apple_app_store_button/da.svg
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
```
LZoog, line 1: I know I said this on Slack already but yeah be sure to send all these suckers through [svgo](https://github.com/svg/svgo) 😎

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
```diff
@@ -0,0 +1,145 @@
...
+import ptApple from '../../assets/images/apple_app_store_button/pt.svg';
+import ruApple from '../../assets/images/apple_app_store_button/ru.svg';
+import skApple from '../../assets/images/apple_app_store_button/sk.svg';
+import slApple from '../../assets/images/apple_app_store_button/sl.svg';
+import svSEApple from '../../assets/images/apple_app_store_button/sv-SE.svg';
+import trApple from '../../assets/images/apple_app_store_button/tr.svg';
+import zhCNApple from '../../assets/images/apple_app_store_button/zh-CN.svg';
+import zhTWApple from '../../assets/images/apple_app_store_button/zh-TW.svg';
+
+import caGoogle from '../../assets/images/google_play_store_button/ca.png';
```
LZoog, line 27: Gaaaah do we really not have these google play images as SVGs anywhere? 😭 It looks like the English version is currently an SVG. Honestly, I would ask around to see if someone knows of some folder somewhere where these exist as translated SVGs. We prefer to [inline our SVGs](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/README.md#working-with-svgs) where we can to reduce the number of network requests, but my bigger problem with this is that unfortunately changing this from an SVG to a `.png` makes this blurry on my retina display, especially when you see it next to the nice crisp apple store SVG:

image

You did also commit the `@2x` png versions, but to utilize those, we should be using the [`Picture` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) to render the correct version for us, meaning we'd need some slight refactoring _and_ you'd need to also import all of the `@2x` versions into this file. I'm really hoping we've got those SVGs somewhere...

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export function getStoreImageByLanguages(
+  store: StoreType,
+  userLanguages: readonly string[] = ['en']
+) {
+  // Iterate through available languages until logo is found, otherwise default to 'en'
+  for (let i = 0; i < userLanguages.length; i += 1) {
+    // If language string includes region, eg. zh-TW, check if an image is available.
+    const language =
+      userLanguages[i].length === 5 &&
```
LZoog, line 134: I think `.includes('-')` and then grabbing the substring before the hyphen will be a little less fragile, we have [tons of locales](https://github.com/mozilla/fxa-content-server-l10n/tree/master/locale) and I'm not sure if all of them will always fit this pattern. I think we may actually be missing some of these locales too which is another reason localizing text on images is not the best but this is currently all we're localizing in the content-server so 😅 

Thanks for the comments in this file btw.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/index.test.tsx
```diff
@@ -27,3 +28,50 @@ describe('Connect another device Promo', () => {
...
+
+  it('should return image with region code', () => {
+    const languages = ['pt-BR', 'pt'];
+    const expected = 'pt-BR.svg';
+    expect(getStoreImageByLanguages(StoreType.apple, languages)).toEqual(
+      expected
+    );
+  });
+
+  it('should return language image if region is not available', () => {
```
LZoog, line 70: TY for test coverage. 😄

LZoog:
@StaberindeZA one thing I just thought of... Maybe this would've been easier/nicer to set with a CSS background-image instead of importing all of these into that React component. You could potentially reference the `lang` attribute on the `html`, like `html[lang='whatever'] .play-store` with a background-image fallback to English. But... we'd still want some sort of map/function to iterate over all of the locales (which you can do with SCSS) and we won't get the benefits of an inlined SVG, plus you already did all the work for this approach, so, meh. ¯\_(ツ)_/¯

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
StaberindeZA: @LZoog I had a look around and we probably could find SVG versions somewhere, but I'm wondering if that's the right course of action.

The [Google badge generator](https://play.google.com/intl/en_us/badges/) only provides PNGs, and looking at the [Firefox homepage](https://www.mozilla.org/en-US/firefox/new/), it looks like they use the Google PNG there as well.

I did find these [unofficial SVGs](https://github.com/steverichey/google-play-badge-svg/tree/master/img), but I did notice the repo was last updated 4 years ago.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/en-US.ftl
```diff
@@ -4,4 +4,10 @@ connect-another-fx-mobile = Get { -brand-firefox } on mobile or tablet
 connect-another-find-fx-mobile = Find { -brand-firefox } in the { -google-play } and { -app-store } or
   <br /><linkExternal>send a download link to your device.</linkExternal>
 
+# Alt text for Google Play and Apple App stores that will be shown if the image can't be loaded.
+connect-another-play-store-image =
+    .alt = Get it on Google Play
+connect-another-app-store-image =
+    .alt = Download on the App Store
```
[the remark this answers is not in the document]
StaberindeZA: Thank you for all the detail here! :D

Regarding the translation, I actually noticed that `fxa-content-server` uses different alt texts, than what I initially included. I've updated the text to match what's in `fxa-content-server`. I'll also reach out to Peiying to see about reusing the existing translations.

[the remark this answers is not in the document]
LZoog: Regarding the new alt text, heh, I think we shouldn't use that period or exclamation unless the image itself does, so I don't think the content server is actually correct if that's how we're setting them over there. 😆

I think it's more so going to be a note, like hey, the community can use these images to reference what translations should be used in order to be consistent with what sighted users will read. She may actually want some sort of comment about this above the translations. Let me know if you'd like help with that, I could introduce you or answer questions that come up.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
StaberindeZA: Ah good catch. I've updated the logic on this to rather look for the `-`.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
LZoog: I personally think using the SVGs is the better approach so we don't have to bother with `@2x` PNG images to account for retina displays and for consistency with all of our other images. I wonder if you can find something to help you compare PNG versions with the SVG versions because yeah... I personally think it's fine to use the unofficial SVGs you found, but it'd be most ideal to compare the PNGs with them to make sure the text on them looks similar. We could also ask to see if we have these officially as SVGs somewhere.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/index.tsx
```diff
@@ -52,14 +54,36 @@ export function ConnectAnotherDevicePromo() {
...
+            id="connect-another-play-store-image"
+            attrs={{ alt: true }}
+          >
+            <img
+              src={getStoreImageByLanguages(
+                StoreType.google,
+                navigatorLanguages
+              )}
+              alt="Get it on Google Play"
+              style={{ maxHeight: '40px' }}
```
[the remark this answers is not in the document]
StaberindeZA: After adding creating the Google Play SVGs this was no longer necessary, so I've removed it.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/storeImageLoader.tsx
StaberindeZA: As suggested by the team, I was able to convert the Google badges `EPS` files to `SVG`. 🎉

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/en-US.ftl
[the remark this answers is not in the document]
StaberindeZA: Thank you for making time to quickly chat about this. As discussed, we decided on using the alt text used in `fxa-content-server`. The only difference is that the `.` and `!` were removed.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/index.tsx
```diff
@@ -52,14 +54,35 @@ export function ConnectAnotherDevicePromo() {
           data-testid="play-store-link"
           href="https://app.adjust.com/2uo1qc?redirect=https%3A%2F%2Fplay.google.com%2Fstore%2Fapps%2Fdetails%3Fid%3Dorg.mozilla.firefox"
         >
-          <img src={playStoreIcon} alt="" />
+          <Localized
+            id="connect-another-play-store-image"
+            attrs={{ alt: true }}
+          >
+            <img
+              src={getStoreImageByLanguages(
```
LZoog, line 62: Sorry for not catching this on my first pass, but this loads the SVG without inlining it, you can see the image loaded from the Network tab. I was going to say it's fine for us to do it in a maintenance follow up issue if you don't have time to do that before we tag midday tomorrow since it's not a big deal, but I decided to check into your branch and poke around at this real quick...

https://github.com/mozilla/fxa/compare/translate-appstore-logos...LZoog:translate-appstore-logos-inline?expand=1#diff-8c237250a12d1f150fe5bdc043cb71c45e70c5c03f0935e57aa9575baa3c3c04R144

This is how you should be able to inline these but you'd also need to update the `alt` text in the FTL file to an `aria-label` instead, we have examples of how to do this in a few places. I'm wondering if we're maybe missing something in the image map though, since I get Typescript complaints from line in that link above saying it's possibly undefined. 🤔

If we don't get these inlined here, let's at least make sure that can't come back as `undefined`. Also, I think to follow other patterns, we can just plop your directories of images in this component itself, as in have all of the new SVGs live in like `ConnectAnotherDevicePromo/google-play-store/[whatever].svg`, similar to how we import other SVGs. I think it might also be more consistent to use that kebab case based on our other icon names but I don't care too much about that.

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/en-US.ftl
```diff
@@ -4,4 +4,11 @@ connect-another-fx-mobile = Get { -brand-firefox } on mobile or tablet
 connect-another-find-fx-mobile = Find { -brand-firefox } in the { -google-play } and { -app-store } or
   <br /><linkExternal>send a download link to your device.</linkExternal>
 
+# Alt text for Google Play and Apple App store images that will be shown if the image can't be loaded.
+# These images are used to encourage users to download Firefox on their mobile devices.
+connect-another-play-store-image =
+    .title = Download { -brand-firefox } for Android
```
LZoog, line 10: I mentioned this in Slack but just bringing it up here too since I'm r+'ing:

We should add "Android" and "iOS" to `App/en-US.ftl` as a brand reference, see [the docs](https://projectfluent.org/fluent/guide/references.html) for more info.

LZoog (approved):
Thank you for sticking with this!! LGTM w/one comment 😄

packages/fxa-settings/src/components/ConnectAnotherDevicePromo/en-US.ftl
```diff
@@ -4,4 +4,11 @@ connect-another-fx-mobile = Get { -brand-firefox } on mobile or tablet
 connect-another-find-fx-mobile = Find { -brand-firefox } in the { -google-play } and { -app-store } or
   <br /><linkExternal>send a download link to your device.</linkExternal>
 
+# Alt text for Google Play and Apple App store images that will be shown if the image can't be loaded.
+# These images are used to encourage users to download Firefox on their mobile devices.
+connect-another-play-store-image =
+    .alt = Download { -brand-firefox } for Android
```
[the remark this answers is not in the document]
StaberindeZA: I ended up going with your recommendation here and updated the `.title` text as suggested.

PR #11239, opened 2021-12-03
chore(deps): Update fxa-dev-launcher to use esm

vbudhram:
## Because

- fxa-dev-launcher module uses `chalk` dep, which was updated to be an esm module, this caused fxa-dev-launcher to break because it couldn't use commonJS and esm syntax

## This pull request

- Updates fxa-dev-launcher to use new esm syntanx
- Removes audit filter checks sicne we don't that use that anymore

Closes: #

## Checklist


LZoog (approved):
Yaaas thank you for fixing this, I noticed it wasn't working just a few days ago and meant to ask around / file an issue for it!

PR #11240, opened 2021-12-03
fix(admin-server): move ts-morph to dependencies

jodyheavener:
Should fix broken admin server

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -281,6 +281,9 @@ jobs:
           name: Running playwright tests
           command: ./packages/functional-tests/scripts/test-ci.sh
       - store_artifacts:
+          path: ~/.pm2/logs
+          destination: logs
+      - store_artifacts:
           path: artifacts
       - store_test_results:
           path: artifacts/tests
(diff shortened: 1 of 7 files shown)


LZoog (approved):
TY!

dannycoates:
the playwright tests are failing because content-server is responding with `Error occured while trying to proxy: localhost:3030/settings`

I added the pm2 logs to the CI artifacts to see if that helps debug

dannycoates:
hmmm
```
2021-12-03T22:47:13: The build failed because the process exited too early. This probably means the system ran out of memory or someone called `kill -9` on the process.
```

dannycoates:
so settings-react is the memory hog of our pm2 processes. on my machine over 1G total for the rescripts and child processes, so it makes sense it's the first to be killed. idk why we're hitting this _now_, but it seems like circleci is being stingier with resources recently. anyway, it isn't until the browser process opens that settings is killed, so we can't just reduce concurrency of startup, which i did anyway while debugging, and it's prob good to keep those changes, so i folded and bumped the instance size to a large for now. unsatisfying but unblocks us

Issue #11250, opened 2021-12-06
Fix issue with pocket access tokens and profile server


data-sync-user:
For historical reasons we should not expire pocket access tokens, however in our profile server we currently do check for expiration. Until pocket moves to refresh tokens, we should be consistent with our policy. We should ignore the access token expiration when pocket uses it to fetch profile information.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4293)

PR #11254, opened 2021-12-06
fix(tokens): Don't check expiration dates for pocket access token on profile server

vbudhram:
## Because

- For legacy reasons, we still need to allow expired pocket tokens to fetch a users profile, regardless of expiration

## This pull request

- Skips expriation check for pocket client ids

Closes: #11250

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/oauth/token.js
+++ b/packages/fxa-auth-server/lib/oauth/token.js
@@ -53,23 +53,31 @@ exports.verify = async function verify(accessToken) {
   const token = await db.getAccessToken(await exports.getTokenId(accessToken));
   if (!token) {
     throw OauthError.invalidToken();
-  } else if (+token.expiresAt < Date.now()) {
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/oauth/token.js
```diff
@@ -65,7 +65,11 @@ exports.verify = async function verify(accessToken) {
     // by slowly purging older tokens from the db.
     if (
       +token.expiresAt >=
-      config.get('oauthServer.expiration.accessTokenExpiryEpoch')
+        config.get('oauthServer.expiration.accessTokenExpiryEpoch') ||
```
dannycoates, line 68: sort of a nit: as a casual reader the `||` makes it harder for me to follow. It'll probably be easier to understand, and to remove later, if we have a nested if.

```js
if (expired) {
  if (!pocket) {
   throw Error
```

dannycoates (approved):
one nit, otherwise LGTM 👍

Issue #10679, opened 2021-10-12
Stop logging bounces to sentry


dannycoates:
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/13622890/?referrer=github_plugin

```
WError: Email account hard bounced
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/error.js", line 778, in Object.AppError.emailBouncedHard
    return new AppError(
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/bounces.js", line 64, in null.<anonymous>
    throw ERRORS[type](tally.latest);
  ?, in Array.forEach
  File "/fxa/packages/fxa-auth-server/dist/fxa-auth-server/lib/bounces.js", line 54, in applyRules
    bounces.forEach((bounce) => {
  ?, in runMicrotasks
...
(10 additional frame(s) were not displayed)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4077)

data-sync-user:
➤ Dan Schomburg commented:

Up for review.

PR #11255, opened 2021-12-06
fix(auth): do not log hard email bounces to sentry

dschom:
## Because

- Logging hard email bounces to sentry is unnecessary.

## This pull request

- Adds a constant, IGNORED_ERROR_NUMBERS, that contains a set of error numbers to ignore that will not be reported to sentry.

Closes: #10679

## Checklist


dschom:
I had trouble figuring out how to manually test this. There are tests cases in place that validate it works, but any advice on how to reproduce this would be helpful.

Issue #9364, opened 2021-06-21
Unverified Account error in graphql-api


data-sync-user:
Pretty sure we shouldn't be seeing these...

https://sentry.prod.mozaws.net/operations/fxa-graphql-prod/issues/10467421

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-3638)

data-sync-user:
➤ Wil Clouser commented:

Benjamin Bangert is this something you can look at next sprint?

data-sync-user:
➤ Benjamin Bangert commented:

Sure.

data-sync-user:
➤ Chelsea Lewis commented:

Ben has added diagnostics in a related issue (see github link). Danny ran a functional test for this last week which worked, the graphql application error didn’t appear on main that time, so it would be interesting if this is fixed. Ben will continue to investigate.

data-sync-user:
➤ Wil Clouser commented:

Benjamin Bangert any change here?  Danny is thinking this may be fixed and he could verify/close this.

data-sync-user:
➤ Benjamin Bangert commented:

I think train 224 has a diagnostic addition. The error still seems to be happening in prod, so we should see what happens on the next release first.

data-sync-user:
➤ Wil Clouser commented:

Benjamin Bangert any updates here?

data-sync-user:
➤ Benjamin Bangert commented:

It’s still happening, a few times a day. This could definitely be in the current sprint.

cknowles-admin:
FxA is no longer synchronizing all issues between Jira and Github.  We are closing open issues and will selectively synchronize in the future.

PR #11285, opened 2021-12-09
feat(gql): add more detailed error on avatar

bbangert:
Because:

* The sentry error indicates an error with unverified accounts when
  querying the avatar on occasion.

This commit:

* Adds additional error details to determine what is happening.

Issue #9364

## Checklist


bbangert:
The sentry error is not terribly helpful, I'm capturing the error separately here to try and get more information about what's happening so this will not actually close the issue.

vbudhram (approved):
@bbangert LGTM 👍🏽

PR #11292, opened 2021-12-09
chore(auth-server): Update email storybook docs, update Fx 'Account' to Fx 'account'

LZoog:
This commit:
* Modifies a few email Storybook docs
* Changes 'Firefox Account' to 'Firefox account', uses the ftl placeholder

Because:
* Some descriptions could be more accurate
* We recently decided to use "account" rather than "Account", this follows our new pattern set in emails

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.stories.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.stories.ts
@@ -12,7 +12,7 @@ export default {
 
 const createStory = storyWithProps(
   'verifySecondary',
-  'Sent to verify the addition of a secondary email via link.',
+  '*NO LONGER USED* This used to be sent to verify the addition of a secondary email via link, but we only send a code now. This should be removed in the future.',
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/senders/emails/templates/verifySecondary/index.stories.ts
```diff
@@ -12,7 +12,7 @@ export default {
 
 const createStory = storyWithProps(
   'verifySecondary',
-  'Sent to verify the addition of a secondary email via link.',
+  '*NO LONGER USED* This used to be sent to verify the addition of a secondary email via link, but we only send a code now. This should be removed in the future.',
```
LZoog, line 15: We have a few of these now (I think only one so far that's already been converted), they're crossed out in the QA checklist sheet. I'll file an additional maintenance issue filed to remove associated files / remove references to these in the codebase. We might want one issue per task though, not sure, but I'll get it in a ticket either way.

xlisachan (approved):
r+

PR #11297, opened 2021-12-10
fix(auth-server): revise emails to update apostrophes

xlisachan:
Because:

* opening single quotes were used instead of closing quotes.

This commit:

* replaces opening quotes with closing ones.

## Checklist


LZoog (approved):
Thank you for catching this!

PR #11300, opened 2021-12-10
feat(fxa-shared): Tags sentry events on critical endpoints

dschom:
## Because

* We'd like to be extra aware of issues occurring on critical endpoints.

## This pull request

* Adds a new module to fxa-shared that allows critical events to be tagged.
* Introduces the this operation into Sentry's beforeSend callback which runs prior to sentry events being processed.

Closes: 

## Checklist

## Screenshots (Optional)

The fxa.endpoint.type: critical tag will now show up for errors that occur on critical endpoints.
image

This is a bit hard to test across all services as there isn't a standard before send operation. 

The list of known critical endpoints was generated from a traffic capture run during 'production-smoke-tests' job. If there is additional manual smoke testing, a follow up tasks is needed to capture the critical endpoints called during these test sessions.


packages/fxa-payments-server/src/lib/sentry.js
```diff
@@ -129,7 +130,9 @@ SentryMetrics.prototype = {
       Sentry.init({
         release,
         dsn,
-        beforeSend,
+        beforeSend: (event) => {
+          return tagCriticalEvent(event);
+        },
```
jodyheavener, line 135: Nitpick: I think some of these could be shortened to just `beforeSend: tagCriticalEvent` since the signatures match?

dschom: Cleaned those up with the short hand.

packages/fxa-shared/tags/sentry.js
```diff
@@ -0,0 +1,117 @@
+// Quick and dirty approach for defining critial endpoints. Future improvements:
+//  - Group by service
+//  - Use regex for precise matching
+//  - Make more config like
+//  - Endpoints are auto generated by traffic capture during smoke tests... so some
+//    of the end points that are only hit by manual testing may not have been captured
+//    and still need to be added.
+
+const CRITICAL_ENDPOINTS = [
```
vbudhram, line 9: Some of these I don't believe are critical endpoints? 

```
'/sockjs-node',
  '/sockjs-node',
  '/tests/bad%20image%20src',
  '/tests/bad%20url',
  '/tests/undefined/v1/account/profile',
  '/.well-known/openid-configuration',

  '/api/auth_status',
  '/api/email_first',
  '/api/oauth',
  '/api/todos/get',
  '/api/logout',
```

vbudhram (commented):
@dschom Code and approach LGTM, although I'm not sure the motivation for this? Maybe I missed it in our triage meeting 😅

packages/fxa-shared/tags/sentry.js
dschom: @vbudhram This was an effort to see if we could make sentry logs more targeted. See jira item FXA-4262 for more info.

vbudhram (approved):
@dschom Thanks Dan, that makes sense. I think we can ship this and evolve it to improve triage.

PR #11302, opened 2021-12-11
bug(auth): destroying attached client would fail

### LINKED ISSUE #11011: (Invalid parameter in request body) response to attachedClientDisconnect mutation
Found while writing a functional test...

"signin to sync and disconnect #1293471" in prod.smoke.spec.ts

https://github.com/mozilla/fxa/pull/10936/files#diff-4345795a3f5cc2229de09a61bd5dee5c2a3844526cd99e2e990002120485b231R4-R28

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4199)

dschom:
## Because

- A test started failing that was destroying attached devices

## This pull request

- Coerces device.sessionTokenId from a buffer into a hex string for comparison
- Coerces device.refreshTokenId from a buffer into a hex string for comparison

closes #11011

## Checklist

### DIFF
--- a/packages/functional-tests/tests/prod.smoke.spec.ts
+++ b/packages/functional-tests/tests/prod.smoke.spec.ts
@@ -15,13 +15,33 @@ test.describe('severity-1', () => {
     );
     await login.login(credentials.email, credentials.password);
     await settings.goto();
-    let services = await settings.connectedServices.services();
+    const services = await settings.connectedServices.services();
     const sync = services.find((s) => s.name !== 'playwright');
     await sync.signout();
     await page.click('text=Rather not say >> input[name="reason"]');
     await settings.clickModalConfirm();
-    await page.waitForNavigation({ waitUntil: 'networkidle' });
-    expect(page.locator('input[type=email]')).toBeVisible();
+    // FIXME
+    // Playwright isn't behaving like a vanilla browser when
+    // sync is disconnected. It's logging out of the web context
+    // but not the browser like it should. This could be due
+    // to a missing pref to handle the broadcast logout message???
+    // In the meantime we have this hack copied from
+    // settings/src/lib/firefox.ts to blast a direct command
+    await page.evaluate((uid) => {
+      window.dispatchEvent(
+        new CustomEvent('WebChannelMessageToChrome', {
+          detail: JSON.stringify({
+            id: 'account_updates',
+            message: {
+              command: 'fxaccounts:logout',
+              data: { uid },
+            },
+          }),
+        })
+      );
+    }, credentials.uid);
+    await login.setEmail(credentials.email);
+    expect(page.url()).toMatch(login.url);
   });
 
   // https://testrail.stage.mozaws.net/index.php?/cases/view/1293385
--- a/packages/functional-tests/tests/prod.smoke.spec.ts
+++ b/packages/functional-tests/tests/prod.smoke.spec.ts
@@ -20,8 +20,8 @@ test.describe('severity-1', () => {
     await sync.signout();
     await page.click('text=Rather not say >> input[name="reason"]');
     await settings.clickModalConfirm();
-    await page.waitForURL('**/signin', { waitUntil: 'networkidle' });
-    expect(page.locator('#password')).toBeVisible();
+    await page.waitForNavigation({ waitUntil: 'networkidle' });
+    expect(page.locator('input[type=email]')).toBeVisible();
   });
 
   // https://testrail.stage.mozaws.net/index.php?/cases/view/1293385
--- a/packages/functional-tests/tests/prod.smoke.spec.ts
+++ b/packages/functional-tests/tests/prod.smoke.spec.ts
@@ -19,7 +19,6 @@ test.describe('severity-1', () => {
     const sync = services.find((s) => s.name !== 'playwright');
     await sync.signout();
     await page.click('text=Rather not say >> input[name="reason"]');
-    await settings.clickModalConfirm();
     // FIXME
     // Playwright isn't behaving like a vanilla browser when
     // sync is disconnected. It's logging out of the web context
@@ -40,6 +39,10 @@ test.describe('severity-1', () => {
         })
       );
     }, credentials.uid);
+
+    // The clickModalConfirm needs to follow the above event. If
+    // it does not, a race condition can occur.
+    await settings.clickModalConfirm();
     await login.setEmail(credentials.email);
     expect(page.url()).toMatch(login.url);
   });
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/routes/attached-clients.js
```diff
@@ -227,19 +227,7 @@ module.exports = (log, db, devices, clientUtils) => {
...
-                .string()
-                .regex(HEX_STRING)
-                .allow(null)
-                .optional(),
-              refreshTokenId: validators.refreshToken.allow(null).optional(),
-              deviceId: DEVICES_SCHEMA.id.allow(null).optional(),
-            })
-            .or('clientId', 'sessionTokenId', 'refreshTokenId', 'deviceId')
-            .with('refreshTokenId', ['clientId']),
+          payload: validators.destroyDevice,
```
dannycoates, line 230: I like that you're thinking of how to test the validation. I also like having the validation defined inline because it's easier to reference when you're working on the route. We don't need to find a solution in this PR since the bug is unrelated but I'd prefer the test to do the flexing in order to keep the code more readable than the other way around. Let's leave it out of this PR though.

packages/fxa-auth-server/lib/routes/attached-clients.js
```diff
@@ -258,17 +246,20 @@ module.exports = (log, db, devices, clientUtils) => {
             request,
             payload.deviceId
           );
+
           if (
             payload.sessionTokenId &&
-            destroyedDevice.sessionTokenId !== payload.sessionTokenId
+            destroyedDevice.sessionTokenId?.toString('hex') !==
```
dannycoates, line 252: Instead of converting to hex here we ought to do it where we pull it out of the db, which is probably the source of this bug since the db layer changed most recently.

Here https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/db/models/auth/device.ts#L196

let's return something like this:
```js
return {
  sessionTokenId: uuidTransformer.from(result.sessionTokenId),
  refreshTokenId: uuidTransformer.from(result.refreshTokenId)
};
```

dannycoates (changes requested):
good find!

packages/functional-tests/tests/prod.smoke.spec.ts
```diff
@@ -9,10 +9,6 @@ test.describe('severity-1', () => {
     credentials,
     pages: { login, settings },
   }) => {
-    test.fixme(
-      true,
-      '(Invalid parameter in request body) response to attachedClientDisconnect mutation'
-    );
```
dannycoates, line 15: and now that it's working we should finish off this test by checking that the row was removed from the ui

packages/fxa-auth-server/lib/routes/attached-clients.js
dschom: Sounds good! I was tempted to put this in the DB layer, but was unsure if that was too aggressive for this fix. Thanks for the feedback.

packages/fxa-auth-server/lib/routes/attached-clients.js
dschom: No problem, I was also torn on this as well. I've gotten thrown off a couple times because the unit tests on the handlers bypass the validators. In this case, the validators for this endpoint seem to be untested, so I made this change in an effort to help pinpoint the bug. I've rolled back the change. I do agree it's nice to see the validation rules inline.

packages/functional-tests/tests/prod.smoke.spec.ts
dschom: A check that the row for the signed out device disappears is now in place. 

One minor thing to note here. After the initial login in this test, there are two devices. One row is for playwright and one row is for the browser being used to test. If the browser's device was logged out, then the browser would be directed back to the sign page. Because of this behavior, I now log out the playwright device instead of the browser. Doing so allows us to observe that the row for the signed out device disappears.

dannycoates: oh, right! yes, this test is specifically meant for sync so we should assert that you're logged out

dannycoates (approved):
LGTM 👍

packages/functional-tests/tests/prod.smoke.spec.ts
```diff
@@ -9,22 +9,19 @@ test.describe('severity-1', () => {
...
     await settings.goto();
-    const services = await settings.connectedServices.services();
+    let services = await settings.connectedServices.services();
     const sync = services.find((s) => s.name !== 'playwright');
     await sync.signout();
     await page.click('text=Rather not say >> input[name="reason"]');
     await settings.clickModalConfirm();
-    // The sync row should be removed but isn't
+    await page.waitForURL('**/signin', { waitUntil: 'networkidle' });
+    expect(page.locator('#password')).toBeVisible();
```
dannycoates, line 24: when used with a locator you want an `await expect`

dschom:
@dannycoates **tldr;** The change password test kept failing. At first it seemed random / flaky, but it was not. My solution to all this was to make sure that when a user ‘signs out’ of their Firefox sync account, a message will be sent to Firefox to also sign out of the current account. I am not 100% confident in this approach, but it seems to work well. Here’s a bit more context surrounding this change if you are interested in how I arrived at this.

When I addressed the feedback in https://github.com/mozilla/fxa/pull/11302#discussion_r768201963, I realized the change password test, which is the test following sign in to sync and disconnect, started to fail. Locally it would sometimes pass, but it would consistently fail in the CI. This led me to a few more realizations.

First, the flaky behavior on my side could be attributed to the fact that in the CI the tests always run with one worker, whereas locally the number of playwright workers is undefined and would run with 2 workers. This meant that occasionally the order of test execution would vary and if the ‘change password’ test were to execute first it would pass.

Second, I was really perplexed as to how this account data was being persisted. Even after the account data was hard set into local storage, and the settings page clearly showed the proper account in use, upon logout the email for the account used by ‘sign in to sync and disconnect’ would show up on the sign in page. The playwright documentation clearly states each test effectively gets a new browser context akin to a new incognito window. I tried to reproduce manually using incognito windows, but no dice… the behavior was not the same.

Ultimately I realized that the account data was getting retrieved from a web channel. The sign out action in the settings panel would effectively clear local storage, but a function invoked on the sign in page would then fallback to asking Firefox itself for these credentials, and would then restore them from the account that Firefox was signed into. In hindsight this makes a lot of sense, but being new to the system it sent me down a bit of a rabbit hole. 

I guess ultimately my question is if this sign out request should really be sent to Firefox in this scenario. If the ideal behavior is that the browser holds onto the synced account even after it has been signed out of in the settings panel, then I think we should address this in the playwright setup. In some ways the current behavior goes against the notion that each test should get a fresh context anyways. In my opinion the notion of a fresh context per test is important, because it guards against variance in state from test to the next, so perhaps we should address this in the tear down anyways.

dannycoates (approved):

PR #11343, opened 2021-12-14
feat(auth-server): convert `subscriptionsPaymentProviderCancelled`

xlisachan:
Because:

This commit:

* Converts the `subscriptionsPaymentProviderCancelled` subscription platform email to the new stack.

Closes: #9299 

## Checklist

## Screenshots (Optional)
Old template
Screen Shot 2021-12-15 at 9 32 34 AM

New template
Screen Shot 2021-12-15 at 9 48 24 AM

* The `Terms and cancellation policy` link has been added to the footer of emails with multiple subscriptions.
* The Storybook layout for multiple products has been updated.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
@@ -67,7 +67,7 @@
         </mj-text>
 
         <mj-text css-class="footer-text">
-          <% if (locals.productName) { %>
+          <% if (locals.productName || locals.subscriptions.length > 0) { %>
             <a href="<%- subscriptionTermsUrl %>" class="footer-link" data-l10n-id="subplat-terms-policy">Terms and cancellation policy</a>
             &nbsp;&nbsp;&bull;&nbsp;&nbsp;
           <% } %>
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
@@ -67,7 +67,7 @@
         </mj-text>
 
         <mj-text css-class="footer-text">
-          <% if (locals.productName || locals.subscriptions.length > 0) { %>
+          <% if (locals.productName || locals.subscriptions?.length > 0) { %>
             <a href="<%- subscriptionTermsUrl %>" class="footer-link" data-l10n-id="subplat-terms-policy">Terms and cancellation policy</a>
             &nbsp;&nbsp;&bull;&nbsp;&nbsp;
           <% } %>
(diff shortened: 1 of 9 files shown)


LZoog:
> I used the `subscriptionsPaymentExpired` email as a reference, should both of these emails have a `Terms and cancellation policy` link in the footer?

I think you can probably reference `subscriptionPaymentProviderCancelled`, the singular version of this one, which does have that link. Though to be sure and just to check this against the original template, if you go into that `write-emails` script and simply change the array of subscriptions to contain multiple and then you run the command, you'll see it output the multiple subscriptions version (this one).

xlisachan:
While the singular versions of both emails in the older templates have that link, the plural versions do not. I'll update the plural versions to include the link though as it should be in the footer.

LZoog:
> While the singular versions of both emails in the older templates have that link, the plural versions do not. I'll update the plural versions to include the link though as it should be in the footer.

SGTM, we can add to that confirmation needed doc. 🙂

jodyheavener (approved):
nice!

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
```diff
@@ -67,7 +67,7 @@
         </mj-text>
 
         <mj-text css-class="footer-text">
-          <% if (locals.productName) { %>
+          <% if (locals.productName || locals.subscriptions.length > 0) { %>
```
LZoog, line 70: Ooo this breaks at least the "multiple products" subscription storybook layout. If you change this + the `.txt` version to `locals.subscriptions?.length > 0` I think that fixes the problem.

I stared at this for a few minutes but I think it's g2g with that fix. I was going to say I kind of wonder why we need this conditional at all, like in what scenarios we _don't_ want this link to show, but I imagine if a user has already cancelled then `subscriptions.length` is `0` and we don't send in a specific `productName` that they're subscribed to then we don't want to display this... guess we'll see as more templates are converted.

LZoog (approved):
r+ but would say it'd be nice if you could address my comment before merge, otherwise get it with the next one 🙂

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
xlisachan: Good catch! I'll make that change.

Yep, there are a couple of emails that have yet to be converted to the new template that don't have `productName` - mainly reminder emails to prompt the user to finish creating their Firefox account.

LZoog: Looks good to me. 🙂 Maybe don't do this here since it's another change and CI run, but when we get to the kind of subscription layout that doesn't render the link, let's add that to the subscription layout stories list too.

PR #11355, opened 2021-12-16
feat(event-broker): add email to profile-change event 

### LINKED ISSUE #10691: Expose (new) email address in profile-change SET data
With the older SQS event notifications the payload for `primaryEmailChangedEvent` contained all the data needed to process the event (i.e. the new email address).  However with `profile-change` SET from the webhook it only contains the uuid, meaning the client has to make an additional call back to FxA to retrieve the profile data to get the new email address.

This generally seems quite wasteful, with added complexity needed to handle quite simple notifications (and in the case of AMO [https://github.com/mozilla/addons-server/issues/14730](https://github.com/mozilla/addons-server/issues/14730) is a problem because we don't store the refresh token for the user right now). Including the new email address would alleviate this.

Using the refresh token also means updates to the email address is only possible until the expiry of the token.  Email is the only way to contact users in the case of security problems so it really needs to stay up to date so they can be contacted, e.g. there is a security bug in their add-on that needs to be addressed.

[filed on the back of a Slack discussion with @bbangert )

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4085)

dannycoates:
closes #10691

### DIFF
--- a/packages/fxa-event-broker/src/queueworker/queueworker.service.ts
+++ b/packages/fxa-event-broker/src/queueworker/queueworker.service.ts
@@ -157,6 +157,8 @@ export class QueueworkerService
         event: message.event,
         timestamp: Date.now(),
         uid: message.uid,
+        //@ts-ignore
+        email: message.email,
       });
       this.log.debug('publishedMessage', { clientId, messageId });
     }
--- a/packages/fxa-event-broker/src/queueworker/queueworker.service.ts
+++ b/packages/fxa-event-broker/src/queueworker/queueworker.service.ts
@@ -31,7 +31,8 @@ function extractRegionFromUrl(url: string) {
 
 @Injectable()
 export class QueueworkerService
-  implements OnApplicationBootstrap, OnApplicationShutdown {
+  implements OnApplicationBootstrap, OnApplicationShutdown
+{
   private region: string;
   private topicPrefix: string;
   private readonly app: Consumer;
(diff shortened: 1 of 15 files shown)


dannycoates:
@bbangert mind having a peek a this? Before implementing RISC events I'm thinking this is what we want for the existing `profile-change` event. IDK that we need to add the old email like in the RISC equivalent since `sub` is the uid. Focus on the second commit to see the payload changes.

packages/fxa-event-broker/src/queueworker/queueworker.service.ts
```diff
@@ -132,14 +152,15 @@ export class QueueworkerService
...
     const clientIds = await this.firestore.fetchClientIds(message.uid);
     for (const clientId of clientIds) {
-      const topicName = this.topicPrefix + clientId;
-      const messageId = await this.pubsub.topic(topicName).publishJSON({
+      const messageId = await this.publishMessage(clientId, {
         changeTime: message.timestamp ? message.timestamp : message.ts * 1000,
         event: message.event,
         timestamp: Date.now(),
         uid: message.uid,
+        //@ts-ignore
```
vbudhram, line 160: What is this ignoring? Seems like it shouldn't be needed?

vbudhram (approved):
@dannycoates This looks good, my biggest question was should we be using a new event instead of exisiting one? I'm not sure which RPs are setup to use this and the impact on them.

packages/fxa-event-broker/src/queueworker/queueworker.service.ts
dannycoates: it's because the other dto types don't have `email` and I didn't feel like asserting the type just to put it in a plain json obj.

PR #11367, opened 2021-12-16
feat(auth-server): convert `subscriptionAccountFinishSetup`

xlisachan:
Because:

This commit:

* Converts the `subscriptionAccountFinishSetup` subscription platform email to the new stack.

Closes: #10225

## Checklist

## Screenshots (Optional)
Old template
Screen Shot 2021-12-16 at 3 38 59 PM

New template - Full details
Screen Shot 2021-12-16 at 3 39 17 PM

New template - Invoice Number
Screen Shot 2021-12-20 at 7 30 44 PM

New template - Invoice Date & Invoice Total
Screen Shot 2021-12-20 at 7 30 52 PM

New template - Next Invoice
Screen Shot 2021-12-20 at 7 30 59 PM

Per @LZoog 's request, update `1xl` to `2xl` in the following files:
* `emails/global.scss`
* `emails/layouts/subscription/index.scss`

Revisions to button partial - do not affect FxA email templates
* `emails/partials/button/index.mjml`: Added conditional `className` to button
* `emails/partials/button/index.scss`: Removed unnecessary import that affected the format of entire email

Added `icon` partial:
* `emails/partials/icon/index.mjml`

Updated emails to reflect icon partial
* `emails/templates/subscriptionDowngrade/index.mjml`
* `emails/templates/subscriptionFirstInvoice/index.mjml`
* `emails/templates/subscriptionPaymentFailed/index.mjml`
* `emails/templates/subscriptionReactivation/index.mjml`
* `emails/templates/subscriptionSubsequentInvoice/index.mjml`
* `emails/templates/subscriptionUpgrade/index.mjml`

Added `paymentPlanDetails` partial:
* `emails/partials/paymentPlanDetails/en.ftl`
* `emails/partials/paymentPlanDetails/index.mjml `
* `emails/partials/paymentPlanDetails/index.txt `

Possible fix:
* `templates/subscriptionAccountFinishSetup.txt` - I believe `productPlan` should be `productName`? The HTML and plaintext of the older template did not match as the product name was missing from the .txt file as `productPlan` was in place of where `productName` should be. I have added this to the "Confirmation needed" doc.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
@@ -5,7 +5,7 @@
 <mj-include path="./css/button/index.css" type="css" css-inline="inline" />
 
 <mj-section>
-  <mj-column>
+  <mj-column css-class="<%= locals.cssClass || undefined %>">
     <mj-button css-class="primary-button" href="<%- link %>">
       <span data-l10n-id="<%- template %>-action">Click here</span>
     </mj-button>
--- a/packages/fxa-auth-server/lib/senders/emails/partials/icon/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/icon/index.mjml
@@ -1,3 +1,7 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
 <mj-section css-class="mb-8">
   <% if (locals.productIconURLOld && locals.productIconURLNew) { %>
     <mj-group>
--- a/packages/fxa-auth-server/lib/senders/emails/partials/icon/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/icon/index.mjml
@@ -1,0 +1,18 @@
+<mj-section css-class="mb-8">
+  <% if (locals.productIconURLOld && locals.productIconURLNew) { %>
+    <mj-group>
+      <mj-column>
+        <mj-image src="<%- productIconURLOld %>" alt="<%- productNameOld %>" title="<%- productNameOld %>" css-class="product-icon">
+        </mj-image>
+      </mj-column>
+      <mj-column>
+        <mj-image src="<%- productIconURLNew %>" alt="<%- productNameNew %>" title="<%- productNameNew %>" css-class="product-icon">
+        </mj-image>
+      </mj-column>
+    </mj-group>
+  <% } else { %>
+    <mj-column>
+      <mj-image src="<%- icon %>" alt="<%- productName %>" title="<%- productName %>" css-class="product-icon"></mj-image>
+    </mj-column>
+  <% } %>
+</mj-section>
--- a/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
@@ -1,4 +1,3 @@
-payment-plan-product-name = { $productName }
 payment-plan-invoice-number = Invoice Number: { $invoiceNumber }
 payment-plan-charged = Charged: { $invoiceTotal } on { $invoiceDateOnly }
 payment-plan-next-invoice = Next Invoice: { $nextInvoiceDateOnly }
--- a/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
@@ -1,3 +1,10 @@
+# Variables:
+#  $invoiceNumber (String) - The invoice number of the subscription invoice, e.g. 8675309
 payment-plan-invoice-number = Invoice Number: { $invoiceNumber }
+# Variables:
+#  $invoiceDateOnly (String) - The date of the invoice, e.g. 01/20/2016
+#  $invoiceTotal (String) - The amount of the subscription invoice, including currency, e.g. $10.00
 payment-plan-charged = Charged: { $invoiceTotal } on { $invoiceDateOnly }
+# Variables
+#  $nextInvoiceDateOnly (String) - The date of the next invoice, e.g. 01/20/2016
 payment-plan-next-invoice = Next Invoice: { $nextInvoiceDateOnly }
--- a/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
@@ -1,0 +1,4 @@
+payment-plan-product-name = { $productName }
+payment-plan-invoice-number = Invoice Number: { $invoiceNumber }
+payment-plan-charged = Charged: { $invoiceTotal } on { $invoiceDateOnly }
+payment-plan-next-invoice = Next Invoice: { $nextInvoiceDateOnly }
--- a/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/index.mjml
@@ -3,27 +3,27 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
 
 <mj-text css-class="text-body">
-  <span><b data-l10n-id="payment-details">Payment details:</b></span>
+  <b data-l10n-id="payment-details">Payment details:</b>
   <ul>
-    <% if (productName) { %>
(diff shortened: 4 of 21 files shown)


packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
```diff
@@ -5,7 +5,7 @@
 <mj-include path="./css/button/index.css" type="css" css-inline="inline" />
 
 <mj-section>
-  <mj-column>
+  <mj-column css-class="<%= locals.cssClass || undefined %>">
```
LZoog, line 8: Only because this looks ever so slightly funky, my first inclination was to recommend...

```
<% if (locals.cssClass) { %> <%= locals.cssClass %> <% } %>
```

or (less confident about this one without testing), `<% locals.cssClass && locals.cssClass %>`?

Though honestly what you have is probably fine so long as `undefined` isn't actually output, and I don't think it is. In React we oftentimes set default classes to `undefined` so `class` isn't rendered on the DOM element so yeah, maybe if my second example works it could be slightly preferable but eh 🤷‍♀️

packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/en.ftl
```diff
@@ -0,0 +1,4 @@
+payment-plan-product-name = { $productName }
```
LZoog, line 1: We shouldn't need an ID here when it's just for something that won't be translated.

Also, this file is missing those variable comments.

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountFinishSetup/index.stories.ts
```diff
@@ -0,0 +1,27 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Meta } from '@storybook/html';
+import { subplatStoryWithProps } from '../../storybook-email';
+
+export default {
+  title: 'SubPlat Emails/Templates/subscriptionAccountFinishSetup',
+} as Meta;
+
+const createStory = subplatStoryWithProps(
```
LZoog, line 12: I don't think we need one for every single conditional, but we might call this one "Full details" and another story called "missing details" with all potential undefined vars not passed in.

packages/fxa-auth-server/lib/senders/templates/subscriptionAccountFinishSetup.txt
```diff
@@ -4,7 +4,7 @@
 
 {{{t "Your payment is processing and may take up to four business days to complete. Your subscription will renew automatically each billing period unless you choose to cancel." }}}
 
-{{#if productPlan }}{{ productPlan }}{{/if}}
+{{#if productName }}{{ productName }}{{/if}}
```
LZoog, line 7: Is this just a mistake in the old txt file, `productPlan` is never passed in?

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountFinishSetup/index.stories.ts
```diff
@@ -0,0 +1,27 @@
...
+  'subscriptionAccountFinishSetup',
+  'Sent to a user to inform them that their payment is processing and prompt them to create a password and download subscription.',
```
LZoog, line 14: I think we can be _slightly_ more explicit here and specify it's for users that purchased the product without an existing Firefox account. They're taken through the passwordless flow and then sent this email.

packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/index.mjml
```diff
@@ -0,0 +1,32 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
+<mj-text css-class="text-body">
+  <span><b data-l10n-id="payment-details">Payment details:</b></span>
```
LZoog, line 6: In the old template, this text looks larger than it is in the new template.

Also you can remove the `span` this is in.

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountFinishSetup/index.mjml
```diff
@@ -0,0 +1,34 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
+<mj-section css-class="mb-6">
+  <mj-column>
+    <mj-image src="<%- icon %>" alt="<%- productName %> icon" title="<%- productName %>" css-class="product-icon"></mj-image>
```
LZoog, line 7: This might be the case across other subplat templates too that I haven't noticed but at least in this one, I think there's more space below the image in the old templates than in the new. Maybe we can add 4-8px of more space here?

xlisachan: A partial for the icon will be created with updated spacing for margin-bottom.

packages/fxa-auth-server/lib/senders/templates/subscriptionAccountFinishSetup.txt
xlisachan: I think so since the HTML had the product name within the Payment Details section and the plain text did not as the variable used is `productPlan` (see below).

Screen Shot 2021-12-20 at 12 35 30 PM

Screen Shot 2021-12-20 at 12 34 18 PM

packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
xlisachan: I checked that `undefined` is not actually in the output. 🙂

packages/fxa-auth-server/lib/senders/templates/subscriptionAccountFinishSetup.txt
xlisachan: I took another look at this and it's odd that the HTML would call out if `productName` was undefined since `productName` is required for the email. 🤔 I'll add this to the Confirmation Needed doc.

packages/fxa-auth-server/lib/senders/emails/partials/paymentPlanDetails/index.mjml
xlisachan: I have the old template screenshot above the new one (lmk if I should have the new one first moving forward). 👀

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountFinishSetup/index.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+    subscriptionSupportUrl: 'http://localhost:3030/support',
+  }
+);
+
+export const SubscriptionFirstInvoiceMissingDetails = createStory(
+  {
+    invoiceDateOnly: '10/13/2021',
+    invoiceNumber: '8675309',
+    invoiceTotal: '$20.00',
+  },
```
LZoog, line 28: I think all of these can actually be missing, right? I like calling this "missing details," but I think it should show what it looks like when all possible missing details are missing. If the template should always have these variables, we shouldn't need the conditional for them.

Unless this is quick, I'm OK with exploring that in a follow up (in the bugs & tasks epic) as a 1-2 pointer to not block here because if those 3 can actually be null/undefined in practice we should update this story to show that, and if they can't, we should modify the template file and take out conditionally displaying them.

packages/fxa-auth-server/lib/senders/templates/subscriptionAccountFinishSetup.txt
LZoog: I just checked [this function in email.js](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L1995) and we never pull in `productPlan`. We probably forgot to update it at some point. It's probably good to note in that doc if even just for QA to see that it was supposed to be there before and wasn't but is present in the new template. Good catch 😎

packages/fxa-auth-server/lib/senders/emails/partials/icon/index.mjml
```diff
@@ -0,0 +1,18 @@
+<mj-section css-class="mb-8">
```
LZoog, line 1: Noiiice this is a solid improvement. 👍

We could use the license at the top of this file.

LZoog (approved):
I potentially have a request for a quick follow up but yeah, LGTM 🙂👍

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountFinishSetup/index.stories.ts
xlisachan: Good call, I'm aligned with this and will add these stories in!

Issue #11246, opened 2021-12-06
[new email templates] Wrong template displayed for HTML version of "Final reminder: Activate your account" email


data-sync-user:
**Build:**
Stage (Train 220.3)

**Preconditions:**
The user doesn’t have a Firefox account.

**Affected platforms:**
Windows 10 64-bit

**Steps to reproduce:**

1. Navigate to [https://accounts.stage.mozaws.net/](https://accounts.stage.mozaws.net/) and fill in a valid email address that doesn't have associated a FF account.
1. Click "Continue" and then enter valid credentials for password and age.
1. Click "Create account".
1. Stop the creation flow by leaving the verification code field empty.
1. Wait for about 15-20 min and then check your inbox for "Final reminder: Activate your account" email.

**Actual result:** 
The wrong template is displayed for "Final reminder: Activate your account" email. Template "verificationReminderFirst" is displayed instead of "verificationReminderSecond" for HTML version.

**Expected result:**
The correct template is displayed for "Final reminder: Activate your account" email for HTML version.

**Additional notes:**

The same issue occurs for verifyLoginCode HTML version. The verifyShortCode template is displayed instead.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4290)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18383/wrong+template1.PNG">wrong template1.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

I verified this issue on Stage [Train 222.0] under Windows 10 64-bit. I confirm that the correct template is displayed for HTML version of "Final reminder: Activate your account" email but for verifyLoginCode HTML version (mentioned in the Additional note) we still have the wrong template. CC: Dan Schomburg

data-sync-user:
➤ Bianca Oltean commented:

Created a new bug for the verifyLoginCode template issue here ( https://mozilla-hub.atlassian.net/browse/FXA-4467 )

Issue #10586, opened 2021-10-01
Blue button text is "Click here" in all emails instead of "Manage account"/ "Create new password"/ "Set up next device", etc.


data-sync-user:
**Build:**
Stage (Train 1.216.0)

**Affected platforms:**
Windows 10 64-bit
macOS 10.14
Android 11
iOS 14.8

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/ and log in using an existing account.
1. Go to “Secondary email” section and add a secondary email.
1. Enter an email address and click the “Save” button.
1. Verify your inbox for the verification code.
1. Copy the verification code and paste it in the field.
1. Click the “Verify” button.
1. Go to inbox and verify the "Secondary email added" email.

Actual result: 
The button text is "Click here" instead of "Manage account" in this case.

Expected result:
The button text is "Manage account" in this case.

**Note:**
This issue can be found in all emails that have a blue button.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4046)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16707/email+button+text+issue.PNG">email button text issue.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/16706/pwdResetAccountRecovery.PNG">pwdResetAccountRecovery.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/16705/Recovery+key+generated+email+issue.PNG">Recovery key generated email issue.PNG</a>

LZoog:
commenting to cross-link: https://github.com/mozilla/fxa/commit/be131108421ea523403a98652f594d34fd025842#r57369209

data-sync-user:
➤ Lauren Zugai commented:

NOTE: we thought we had this fixed, but it will actually be fixed when https://mozilla-hub.atlassian.net/browse/FXA-4372 ( https://mozilla-hub.atlassian.net/browse/FXA-4372|smart-link ) is closed this sprint. I would reopen this ticket since the bug isn’t fixed, but, then we’d have two tickets open for the same issue, so I’ll leave this one closed and just cross-link here instead. 🤷‍♀️

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on stage [Train 224.1] under Windows 10 64-bit.  The correct text is displayed on the button depending on the email sent.

[...]

PR #11380, opened 2021-12-18
bug(auth): Fix second verification reminder email

dschom:
## Because:
 
  - Second verification reminder produced the wrong email.

## This commit:

   - Reverts the use of the partials/verificationReminder/index.mjml template.
   - This is not a long term fix. It appears there maybe another underlying issue.

Closes: #11246

## Checklist

Note: The issue could not be reproduced locally. At best this is a quick fix and at worst it's a good sanity check. Both story book and manual testing generated the correct email body; however, testing on stage reproduces the issue. There will be another bug filed documenting this problem, as it seems like this is not an isolated incident and there are other templates which reference partials that have similar issues.


packages/fxa-auth-server/lib/senders/emails/templates/verificationReminderSecond/index.mjml
```diff
@@ -2,4 +2,24 @@
...
+    <mj-text css-class="text-header">
+      <span data-l10n-id="verificationReminderSecond-title">Still there?</span>
+    </mj-text>
+  </mj-column>
+</mj-section>
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-sub-body">
+      <span data-l10n-id="verificationReminderSecond-description">Almost a week ago you created a { -product-firefox-account } but never verified it. We’re worried about you.</span>
```
LZoog, line 16: Since this is the English fallback text, we shouldn't use placeholders here, this should be "created a Firefox account...'

LZoog (commented):
> This is not a long term fix. It appears there maybe another underlying issue.

I actually think this is fine. The problem here was the fallback text in `verificationReminder` is the fallback text for the first template, and Fluent wasn't replacing the text for us like we thought it might for partials in cases like this. I don't know if we should invest any time in "fixing" that on staging because of that scenario we mentioned where localization might not be in place for particular strings, and then I think folks would be shown the incorrect English fallback text. Doing it how you did here is more explicit which I think is nicer and less confusing, though yeah, it relates to at least #10586.

Also, can we remove `verificationReminder` partial if it's only used in `verificationReminderFirst`? We can probably copy all of that into the template and remove the partial.

packages/fxa-auth-server/lib/senders/emails/templates/verificationReminderSecond/index.mjml
dschom: Ah, make sense. Thanks for pointing this out.

LZoog (approved):
Going to go ahead and r+ since everything looks great but that one `template` comment. Thanks for the fix! 😀

Issue #10707, opened 2021-10-14
Coupons - [UI] coupon component


data-sync-user:
Make a coupon react component. The component should have an input box, an error message and a button.

* The button toggles from Apply to Remove
* The input toggles between an input and a text label that displays the coupon code.
* The error message only displays if there is an issue with the coupon code.
* API hookup will be done in [https://mozilla-hub.atlassian.net/browse/FXA-4155](https://mozilla-hub.atlassian.net/browse/FXA-4155|smart-link) 

!Screen Shot 2021-10-14 at 2.33.46 PM.png|width=375,height=556!

!Screen Shot 2021-10-14 at 2.33.34 PM.png|width=383,height=659!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4093)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/16935/Screen+Shot+2021-10-14+at+2.33.34+PM.png">Screen Shot 2021-10-14 at 2.33.34 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/16934/Screen+Shot+2021-10-14+at+2.33.46+PM.png">Screen Shot 2021-10-14 at 2.33.46 PM.png</a>

Issue #11150, opened 2021-11-24
Coupons - [UI] Update new & old flow(existing user) to allow coupons


data-sync-user:
Once the coupon component is created ensure it is available and working in the legacy non passwordless flow. 

!Screen Shot 2021-10-14 at 2.33.46 PM.png|width=375,height=556!

!Screen Shot 2021-10-14 at 2.33.34 PM.png|width=383,height=659!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4263)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18175/Screen+Shot+2021-10-14+at+2.33.34+PM.png">Screen Shot 2021-10-14 at 2.33.34 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/18176/Screen+Shot+2021-10-14+at+2.33.46+PM.png">Screen Shot 2021-10-14 at 2.33.46 PM.png</a>

Issue #10973, opened 2021-11-10
Coupons - [UI] add storybook entries for coupon component


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4190)

PR #11391, opened 2021-12-20
feat(payments): coupon Component UI

IvoJP:
## Because

- We want to have an input field for a user to enter a discount code

## This pull request

- adds the component to both old and new subscription flow

Closes: #10707, #11150, #10973
## Checklist

## Screenshots (Optional)

Screen Shot 2021-12-21 at 11 28 55 AM
Screen Shot 2021-12-21 at 11 28 26 AM
Screen Shot 2021-12-21 at 11 28 15 AM

Updated landscape in mobile to render components in the appropriate order (plan details > coupon > payment)


packages/fxa-auth-server/config/index.ts
```diff
@@ -854,7 +854,7 @@ const conf = convict({
     coupons: {
       enabled: {
         doc: 'feature flag for enabling coupons',
-        default: false,
+        default: true,
```
chenba, line 857: We probably should have this default to false until we get the feature working in prod.

packages/fxa-payments-server/server/config/index.js
```diff
@@ -14,7 +14,7 @@ convict.addFormats(require('convict-format-with-validator'));
 const conf = convict({
   featureFlags: {
     subscriptionCoupons: {
-      default: false,
+      default: true,
```
chenba, line 17: Oh, maybe this should default to false since it affects end use UI.  The auth-server flag is probably ok with a default of true.

packages/fxa-payments-server/src/components/Coupon/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Coupon from './index';
+
+storiesOf('components/Coupon', module)
+  .add('default', () => (
+    <MockApp>
+      <Coupon />
+    </MockApp>
+  ))
+  .add('localized to xx-pirate', () => (
+    <MockApp languages={['xx-pirate']}>
```
chenba, line 13: I don't think the coupon feature has translation from Stripe metadata?

packages/fxa-payments-server/src/components/Coupon/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+              <Localized id="coupon-remove">
+                <span>Remove</span>
+              </Localized>
+            </button>
+          </div>
+        ) : (
+          <Form validator={validator} onSubmit={onSubmit}>
+            <div className="input-row">
+              <Localized attrs={{ placeholder: true }} id="coupon-enter-code">
+                <input
```
chenba, line 82: Usually the `Input` component is used (not the `input` tag).  If the form is simple enough where that is not needed, that's fine, but we should then remove the `validator`.  As it is I don't think the validator is aware of that input.

IvoJP: the form is simple enough and theres no validation, i changed it to use the native form,input, and button elements.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -421,20 +429,22 @@ export const Checkout = ({
...
-            selectedPlan,
-            isMobile,
-            showExpandButton: isMobile,
-          }}
-        />
-        {config.featureFlags.subscriptionCoupons ? (
-          // To be updated in issue #7097
-          <div data-testid="coupon-container"></div>
-        ) : null}
+        <CouponContext.Provider value={{ coupon, setCoupon }}>
```
StaberindeZA, line 432: Could we change the Coupon component to use props instead of using Context? This seems to be the recommended approach and more closely matches what we do in the rest of the app.

IvoJP: The PlanDetails and Coupon components are sibling elements. I'm not super fluent in React, but I believe the proper way to pass data between siblings is either by using a data store (like redux or whatever) or with the context/hooks API. I think you probably could get something going with properties but usually properties are for passing data down from a parent into a child component. There are a few other uses of the context API in the app and if I recall correctly, context/hooks is the way Bianca mentioned the app moving towards.

packages/fxa-payments-server/src/components/Coupon/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  MouseEventHandler,
+  FormEventHandler,
+} from 'react';
+import { Localized } from '@fluent/react';
+import './index.scss';
+import { CouponContext } from '../../lib/CouponContext';
+
+export const Coupon = () => {
+  const [hasCoupon, setHasCoupon] = useState(false);
+  const [couponCode, setCouponCode] = useState('');
```
chenba, line 13: Having two states and manually set them feels fragile.  The hasCoupon state is dependent on the value of the couponCode, but the consts are unrelated.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
IvoJP: Reference: https://www.pluralsight.com/guides/how-to-pass-data-between-react-components

packages/fxa-payments-server/src/components/Coupon/index.tsx
IvoJP: One is holding the state of the coupon code as it is entered into the input field, the other is set/reset depending on if the coupon code was valid after hitting apply and manages the UI. Right now it is mainly used for handling the dummy data and I can probably rip it out afterwards, but I am not sure how the 2 might be consolidated.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
chenba: We've been passing functions through props.  For example, here's the checkout page passing functions to the new user form: https://github.com/mozilla/fxa/blob/main/packages/fxa-payments-server/src/routes/Checkout/index.tsx#L314:L325

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -109,6 +112,31 @@ export const PlanDetails = ({
...
+                            intervalCount: interval_count,
+                          }}
+                        ></Localized>
+                      </div>
+                    </div>
+                    <div className="plan-details-total-inner">
+                      <Localized id="coupon-discount">
+                        <div>Discount</div>
+                      </Localized>
+                      <div>-${coupon.amount}</div>
```
chenba, line 136: Negative sign in front is not universal unfortunately.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -117,15 +145,33 @@ export const PlanDetails = ({
                     id={`plan-price-${interval}`}
                     attrs={{ title: true }}
                     vars={{
-                      amount: getLocalizedCurrency(amount, currency),
+                      amount:
+                        coupon && config.featureFlags.subscriptionCoupons
+                          ? getLocalizedCurrency(
+                              amount ? amount - coupon.amount * 100 : amount,
```
chenba, line 151: The price and the discount are in different units?! 😢

packages/fxa-payments-server/src/routes/Checkout/index.tsx
StaberindeZA: The react docs are also a good reference for this.

https://reactjs.org/docs/context.html#before-you-use-context

packages/fxa-payments-server/src/lib/CouponContext.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+type CouponContextType = {
+  coupon: any;
+  setCoupon: any;
```
chenba, line 5: Since we know the types we shouldn't use `any` here.

packages/fxa-payments-server/src/components/Coupon/index.tsx
chenba: > One is holding the state of the coupon code as it is entered into the input field, the other is set/reset depending on if the coupon code was valid after hitting apply and manages the UI.

Ah, sorry, I misunderstood then.  Thanks for the clarification.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
IvoJP: context removed

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
IvoJP: yes, although once we hook this up to the back end and have an idea of how we are signaling whether or not the coupon is a fixed amount or percentage, we will need to update this

IvoJP: well i should also say that the amount of the plan is in cents essentially so $9.35 is 935, i am not sure how we are going to store the coupon amount. this will get removed once we hook this up to the back end

IvoJP: this is just for the dummy data, i'll just remove it and change the coupon to match the units. sorry i wasnt thinking about this too much but this will change anyway when we hook up the back end. we will also have to implement handling fixed amounts vs percentage discounts

StaberindeZA: I believe the backend takes care of that. It looks like regardless of whether the coupon is fixed amount or  percentage, it'll do the calculation for us and return a fixed amount.

packages/fxa-payments-server/src/lib/Coupon.ts
```diff
@@ -0,0 +1,3 @@
+export type Coupon = {
+  amount: number;
```
StaberindeZA, line 2: Could we add the `couponCode`, entered by the user, to this as well? We'll need it in upcoming tickets.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -429,11 +437,12 @@ export const Checkout = ({
             selectedPlan,
             isMobile,
             showExpandButton: isMobile,
+            coupon: coupon,
           }}
         />
         {config.featureFlags.subscriptionCoupons ? (
           // To be updated in issue #7097
```
StaberindeZA, line 444: nit, could you remove this comment?

packages/fxa-payments-server/src/routes/Product/SubscriptionCreate/index.tsx
```diff
@@ -321,12 +326,16 @@ export const SubscriptionCreate = ({
...
             }),
-            profile,
             selectedPlan,
             isMobile,
             showExpandButton: isMobile,
+            coupon: coupon,
           }}
         />
+        {config.featureFlags.subscriptionCoupons ? (
+          // To be updated in issue #7097
```
StaberindeZA, line 336: nit, could you remove this comment?

packages/fxa-payments-server/src/components/PlanDetails/index.test.tsx
```diff
@@ -162,6 +163,70 @@ describe('PlanDetails', () => {
...
+    //           }}
+    //         />
+    //     );
+    //   };
+
+    //   const { queryByTestId} = subject();
+    //   expect(queryByTestId('total-price')).toContain('7.35');
+    // });
+
+    it('displays a success message', () => {
```
StaberindeZA, line 210: Could we update this test to include checking that the correct totals and discount amounts are shown?

packages/fxa-payments-server/src/components/PlanDetails/index.test.tsx
```diff
@@ -162,6 +163,70 @@ describe('PlanDetails', () => {
...
+        />
+      );
+    };
+
+    const { queryByTestId } = subject();
+    expect(queryByTestId('coupon-success')).not.toBeInTheDocument();
+  });
+
+  describe('Valid Coupon Used', () => {
+    // it('updates price', () => {
```
StaberindeZA, line 185: nit, could we remove this if not needed?

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  coupon?: Coupon;
+  setCoupon: (coupon: Coupon | undefined) => void;
+};
+
+export const CouponForm = ({ coupon, setCoupon }: CouponFormProps) => {
+  const [hasCoupon, setHasCoupon] = useState(coupon ? true : false);
+  const [couponCode, setCouponCode] = useState(coupon ? 'test' : '');
+  const [error, setError] = useState(false);
+
+  const onSubmit: FormEventHandler = (event: any) => {
```
chenba, line 16: We shouldn't assert `event` to be `any`; we could just rely on `FormEventHandler`'s type, so you could just remove the type declaration for `event` and it'd have the correct type.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+      setHasCoupon(true);
+      setCoupon({
+        amount: 200,
+      });
+    } else {
+      setError(true);
+    }
+  };
+
+  const removeCoupon: MouseEventHandler<HTMLButtonElement> = (event: any) => {
```
chenba, line 29: No need for `any` here also.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setCoupon(undefined);
+  };
+
+  return (
+    <section
+      className={`coupon-component ${coupon ? 'has-coupon' : ''}`}
+      data-testid="coupon-component"
+    >
+      <div className="coupon-component-inner">
+        {hasCoupon ? (
```
chenba, line 42: Why two of these?  Seems possible to combine them?

packages/fxa-payments-server/src/components/PlanDetails/index.test.tsx
```diff
@@ -162,6 +163,70 @@ describe('PlanDetails', () => {
...
+    //             selectedPlan,
+    //             coupon: { amount: 200}
+    //           }}
+    //         />
+    //     );
+    //   };
+
+    //   const { queryByTestId} = subject();
+    //   expect(queryByTestId('total-price')).toContain('7.35');
+    // });
```
chenba, line 208: Delete the commented out code?

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -5,6 +5,7 @@ import React, {
   useState,
   useCallback,
   Suspense,
+  SetStateAction,
```
chenba, line 8: I can't find where this is used.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
```diff
@@ -429,11 +437,12 @@ export const Checkout = ({
             selectedPlan,
             isMobile,
             showExpandButton: isMobile,
+            coupon: coupon,
           }}
         />
         {config.featureFlags.subscriptionCoupons ? (
           // To be updated in issue #7097
-          <div data-testid="coupon-container"></div>
+          <CouponForm coupon={coupon} setCoupon={setCoupon} />
```
chenba, line 445: unimportant style comment: shorter/lazier as `{...{coupon, setCoupon}}`

chenba (approved):
r+ with comments.  A few minor things but LGTM overall.  Thanks for including the screenshots.

packages/fxa-payments-server/src/routes/Checkout/index.tsx
IvoJP: i think this is something i fat fingered and vs code added it. removing it...

packages/fxa-payments-server/src/components/CouponForm/index.tsx
IvoJP: i was following the pattern of the plan details component. probably possible to combine them but at the expense of refactoring the css a bit - this could possible be spun off into a separate ticket for a maintanence sprint where we do a bit of refactoring on templates/css

PR #11392, opened 2021-12-20
Train 221.1 conflict fix -> Train 221.0

LZoog:
I had a conflict in `paypal.js` when cherry-picking these in. Opening this as a PR into `train-221` to make sure all tests pass and such before tagging in the train branch.

After this is merged I'll tag in `train-221`.


vbudhram:
@LZoog Looks like some auth test failed? Maybe something missing in merge?

LZoog:
@vbudhram you're right, I must've messed up the conflict resolution somehow. Hopefully my latest push fixes it.

LZoog:
Closing this - the cherry-pick and conflict fix might have other dependencies and it's not essential to make this train tag since it's staying only in staging and we have another staging tag in two days. Will have another PR up with two cleanly cherry-picked commits + tag.

Issue #11067, opened 2021-11-18
Tax Reporting: For an existing stripe user Take stripe customer to pull out zip code from payment method


data-sync-user:
[stripe - existing] Take stripe customer to pull out zip code from payment method

* Need to identify a source to turn zip code into the state they’re in (open source & free csvs that will turn a zip code into a state)
* Update address (state/province, country) in stripe customer for all customers

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4221)

data-sync-user:
➤ Barry Chen commented:

Anna Sobiepanek  Where does this fit in?  Is this a one time script?  How should PayPal customers be handled?

data-sync-user:
➤ Anna Sobiepanek commented:

Barry Chen  the paypal customers will be handled in https://mozilla-hub.atlassian.net/browse/FXA-4220 ( https://mozilla-hub.atlassian.net/browse/FXA-4220|smart-link ). This is not a onetime script. The intent here is to loop through stripe paying customers get their postal/zip code from their CC and run that through the function developed in https://mozilla-hub.atlassian.net/browse/FXA-4225 ( https://mozilla-hub.atlassian.net/browse/FXA-4225|smart-link ) in order to get the location information and then update that user in stripe.

data-sync-user:
➤ Barry Chen commented:

Anna Sobiepanek Thanks.  I still don’t understand when we execute this.  What’s the use case that leads into the loop you mention?

data-sync-user:
➤ Barry Chen commented:

Some work in progress pushed to https://github.com/mozilla/fxa/tree/4221-stripe-city ( https://github.com/mozilla/fxa/tree/4221-stripe-city|smart-link ) 

Unassigning myself because I’ll be on PTO.

PR #11393, opened 2021-12-20
feat(subscriptions): set customer location w/ payment info

chenba:
Co-authored-by: Reino Muhl <reino.muhl@gmail.com>

Because:
 - we need to set the customers' state/province for tax purposes

This commit:
 - update a customer's address by looking up the state/province with the
   postal code and country from the payment method details

Closes: #11067


chenba:
@StaberindeZA since you are a co-author on the commit, please take a look to ensure that the code hasn't been edited in a way that misrepresents your ideas.

@biancadanforth you are probably the best person to review this at this time.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2371,27 +2436,36 @@ export class StripeHelper {
       const paymentMethod = customer.invoice_settings
         .default_payment_method as Stripe.PaymentMethod;
       if (paymentMethod.card) {
-        ({ last4: lastFour, brand: cardType } = paymentMethod.card);
+        ({ last4: lastFour, brand: cardType, country } = paymentMethod.card);
       }
+      if (paymentMethod.billing_details.address)
```
biancadanforth, line 2441: Do we know `billing_details` is going to be defined? I.e. should we use the optional chain operator here?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2354,15 +2409,25 @@ export class StripeHelper {
     };
   }
 
-  async extractCustomerDefaultPaymentDetails(uid: string) {
+  async extractCustomerDefaultPaymentDetails(
```
biancadanforth, line 2412: Hmm the split params depending on the caller for this helper is confusing. Since `setCustomerLocation` isn't used anywhere yet, perhaps it could take a `uid` instead of the `customer` as a param to avoid this?

I do like the pattern of giving back various pieces of info related to the payment method and having the caller destructure to get whatever it needs.

packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
```diff
@@ -65,6 +65,9 @@ async function init() {
   const config = require('../config').getProperties();
   const log = require('../lib/log')(config.log.level);
 
+  Container.set(AppConfig, config);
+  Container.set(AuthLogger, log);
```
biancadanforth, line 69: Glad you saw the comment in the Jira about how the GoogleMapsService assumes these two objects are already set on the container. Do we expect that service to be used in these various scripts?

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -3824,13 +3879,23 @@ describe('StripeHelper', () => {
...
       });
+
+      it('does not fetch the customer when a customer is given as the argument', async () => {
+        const customer = deepCopy(mockCustomer);
+        sandbox.stub(stripeHelper, 'fetchCustomer').rejects();
+        await stripeHelper.extractCustomerDefaultPaymentDetails({
+          customer,
+        });
+        sinon.assert.notCalled(stripeHelper.fetchCustomer);
+      });
```
biancadanforth, line 3898: I'm hoping we can avoid this test if we change the param for `setCustomerLocation` to take in a `uid`.

biancadanforth (approved):
R+. Thanks Barry. I see [FXA-4223](https://mozilla-hub.atlassian.net/browse/FXA-4223) will handle actually calling this new method when we create a Stripe customer for new Stripe subscriptions.

I do make a suggestion about changing the function signature to take in a `uid` instead of a Stripe `customer` to avoid the split params for `extractCustomerDefaultPaymentDetails`.

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: I anticipate the upcoming code will have a customer object, either from `stripe.customers.list` or a customer after a successful subscription.

We could simply have two functions.  The one that takes a uid would first fetch the customer then call the one that takes a customer.  What do you think?

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: I don't remember seeing any warnings from the type checker while working on it.

packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
chenba: It doesn't matter if the geocoding feature is used in the scripts or not.  GoogleMapsService is now a dependency in StripeHelper, and those are GoogleMapsService's dependencies.

chenba:
Split the payment details function into two and rebased on latest main.  Will merge once CI is green.

packages/fxa-auth-server/lib/payments/stripe.ts
biancadanforth: Sure that works too. Yeah I haven't thought about exactly where it will be called but I agree with you we will probably already have the Stripe customer.

packages/fxa-auth-server/scripts/clean-up-old-ci-stripe-customers.js
biancadanforth: Ah so anywhere that uses StripeHelper also needs to set those objects in the container. I got it thanks for clarifying.

PR #11394, opened 2021-12-20
Train 221.1/.2/.3 -> main

LZoog:
This is a dot release containing two issues. **This is staying on staging for early testing before we tag 223 in two days which will also stay in staging until the new year.**


LZoog:
I had a merge conflict with `fxa-shared/package.json` due to a dependabot merge this morning, which is why we've got a `.2` release now in this PR instead of `.1`. I pulled in that dependabot commit so we can merge this train branch in after it passes CI - I didn't open a separate train branch PR into the train since I already had this open and it shouldn't affect tests, fingers crossed.

vbudhram (approved):

PR #11403, opened 2021-12-21
feat(auth-server): convert `subscriptionRenewalReminder`

### LINKED ISSUE #10227: [emails M4] Convert `subscriptionRenewalReminder`
Take `subscriptionRenewalReminder.html` and any needed partials and convert the template(s) over to the new emails stack. Also convert tests, add states to the state previewing solution, and ensure the template is localized. Make sure it builds properly and that the new email is sent locally by turning on the feature flag it sits behind.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3893)

xlisachan:
Because:

This commit:

* Converts the `subscriptionRenewalReminder` subscription platform email to the new stack.

Closes #10227

## Checklist

## Screenshots (Optional)
Old template
Screen Shot 2021-12-21 at 10 00 12 AM

New template
Screen Shot 2021-12-21 at 9 59 58 AM

Added `subscriptionSupportContact` partial:
* `emails/partials/subscriptionSupportContact/en.ftl`
* `emails/partials/subscriptionSupportContact/index.mjml`
* `emails/partials/subscriptionSupportContact/index.txt`

Added `subscriptionUpdateBillingEnsure` partial:
* `emails/partials/subscriptionUpdateBillingEnsure/en.ftl`
* `emails/partials/subscriptionUpdateBillingEnsure/index.mjml`
* `emails/partials/subscriptionUpdateBillingEnsure/index.txt `

Revised `updateBilling` partial to `subscriptionUpdateBillingTry` to avoid confusion:
* `emails/partials/subscriptionUpdateBillingTry/en.ftl`
* `emails/partials/subscriptionUpdateBillingTry/index.mjml`
* `emails/partials/subscriptionUpdateBillingTry/index.txt`

Updated emails using `updateBilling` partial:
* `emails/templates/subscriptionPaymentFailed/index.mjml `
* `emails/templates/subscriptionPaymentFailed/index.mjml`

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/partials/subscriptionUpdateBillingTry/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/subscriptionUpdateBillingTry/index.mjml
@@ -5,7 +5,7 @@
 <mj-section>
   <mj-column>
     <mj-text css-class="text-body">
-      <span data-l10n-id="updateBilling">
+      <span data-l10n-id="subscriptionUpdateBillingTry">
         We’ll try your payment again over the next few days, but you may need to help us fix it by <a data-l10n-name="updateBillingUrl" href="<%- updateBillingUrl %>">updating your payment information</a>.
       </span>
     </mj-text>
--- a/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionRenewalReminder/index.stories.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionRenewalReminder/index.stories.ts
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Meta } from '@storybook/html';
+import { subplatStoryWithProps } from '../../storybook-email';
+
+export default {
+  title: 'SubPlat Emails/Templates/subscriptionRenewalReminder',
+} as Meta;
+
+const createStory = subplatStoryWithProps(
+  'subscriptionRenewalReminder',
+  'Sent when a user remind them to renew their subscription.',
+  {
+    productName: 'Firefox Fortress',
+    invoiceTotal: '$20.00',
+    planInterval: 'week',
+    planIntervalCount: '2',
+    reminderLength: '14',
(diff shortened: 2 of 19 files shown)


packages/fxa-auth-server/lib/senders/emails/partials/subscriptionUpdateBillingTry/index.mjml
```diff
@@ -5,7 +5,7 @@
 <mj-section>
   <mj-column>
     <mj-text css-class="text-body">
-      <span data-l10n-id="updateBilling">
+      <span data-l10n-id="subscriptionUpdateBillingTry">
```
LZoog, line 8: You may already know this, but changing the FTL string IDs can have some implications attached. I think it's fine here because Pontoon has strings in memory so if this is already been translated it shouldn't have to be retranslated, but if you ever need to change a string, we _have_ to update the ID with it.

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionRenewalReminder/index.stories.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Meta } from '@storybook/html';
+import { subplatStoryWithProps } from '../../storybook-email';
+
+export default {
+  title: 'SubPlat Emails/Templates/subscriptionRenewalReminder',
+} as Meta;
+
+const createStory = subplatStoryWithProps(
+  'subscriptionRenewalReminder',
+  'Sent when a user remind them to renew their subscription.',
```
LZoog, line 14: Reading the text, I think this is a reminder that it'll be automatically renewed. Maybe "Sent to remind a user of an upcoming automatic subscription renewal." Though, I'm curious when we do that, like if it's halfway through the billing cycle or what? Maybe the Stripe docs can tell us. If it's something we set through Stripe (probably need to ask the SubPlat team), we could say, "Sent to remind a user of an upcoming automatic subscription renewal X days out from charge, X being what is set in the Stripe dashboard." It's ok too if we want to specify what X is currently, but we could change that in the future 🤷‍♀️

LZoog (approved):
LGTM with one comment. 🙂

I think this is already noted, but if we don't already have "This is an automated email; if you received it in error, no action is required." in the confirmation needed doc for the subplat layout, can we add that? It's very strange that we have that message in the plaintext, but not in the HTML, and we don't state "You’re receiving this email because..." in the plaintext version. Maybe we figured they were sufficiently similar.

PR #11404, opened 2021-12-21
feat(auth-server): convert `subscriptionAccountReminderSecond`

xlisachan:
Because:

This commit:

* Converts the `subscriptionAccountReminderSecond` subscription platform email to the new stack.

Closes: #10911

## Checklist

## Screenshots (Optional)
Old template
Screen Shot 2021-12-21 at 10 27 23 AM

New template
Screen Shot 2021-12-21 at 10 27 12 AM

Note, `subscriptionAccountReminderSecond` is already included in `dev.json`

I have added the update of the button from "Create a password" to "Create Password" to the Confirmation Needed doc.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountReminderSecond/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountReminderSecond/index.mjml
@@ -1,0 +1,27 @@
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
+
+<mj-section>
+  <mj-column>
+    <mj-text css-class="text-header">
+      <span data-l10n-id="subscriptionAccountReminderSecond-title">Welcome to Firefox!</span>
+    </mj-text>
+
+    <mj-text css-class="text-body">
+      <span data-l10n-id="subscriptionAccountReminderSecond-content-info">
+        A few days ago you created a Firefox account but never verified it. We hope you’ll finish setting up your account, so you can use your new subscription.
+      </span>
+    </mj-text>
+
+    <mj-text css-class="text-body">
+      <span data-l10n-id="subscriptionAccountReminderSecond-content-select">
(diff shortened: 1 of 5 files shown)


packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountReminderSecond/index.mjml
```diff
@@ -0,0 +1,27 @@
...
+
+    <mj-text css-class="text-body">
+      <span data-l10n-id="subscriptionAccountReminderSecond-content-select">
+        Select “Create Password” to set up a new password and finish verifying your account.
+      </span>
+    </mj-text>
+  </mj-column>
+</mj-section>
+
+<%- include ('/partials/button/index.mjml', { cssClass: 'mb-8' }) %>
```
LZoog, line 25: Ohh, I think this is actually more space than we have in the original email, but this is much better in my opinion. We have "support link" right after the button, so, this creates more space between two clickable things, which is much nicer in mobile. 👍

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountReminderSecond/index.mjml
```diff
@@ -0,0 +1,27 @@
...
+<%- include ('/partials/subscriptionSupport/index.mjml') %>
+<%- include ('/partials/automatedEmailResetPassword/index.mjml') %>
```
LZoog, line 27: Piggybacking off my review on your other PR, this here + in some plaintext files _really_ makes me wonder if this is supposed to be in the HTML of every subplat email😅 Things to figure out later!

LZoog (approved):
L
G
T
M
🎉

> Note, subscriptionAccountReminderSecond is already included in dev.json

I think this was added on accident, this is a new template and I think it was added in there since it was thought it needed to be there.

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionAccountReminderSecond/index.mjml
xlisachan: I've added it to the Confirmation Needed doc. 🙂

Issue #11374, opened 2021-12-17
[subplat] [email] Misplaced text in the “You have upgraded to Mozilla VPN” email


data-sync-user:
**Affected versions:**

* All;

**Affected FxA versions:**

* 1.221.0;

**Prerequisites:**

* Have the VPN set to US;

**Steps to reproduce:**
1. Subscribe to the monthly subscription plan by using the following link:
[https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=plan_FvPMH5lVx1vhV0](https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=plan_FvPMH5lVx1vhV0)
2. With the same account, upgrade your monthly subscription plan to the 12-months plan: [https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=price_1J0Y1iKb9q6OnNsLXwdOFgDr](https://accounts.stage.mozaws.net/subscriptions/products/prod_FiJ42WCzZNRSbS?plan=price_1J0Y1iKb9q6OnNsLXwdOFgDr)
3. Open your client email;
4. Open “You have upgraded to Mozilla VPN” email;
5. Observe the subscription plans displayed in the email;

**Expected result:**

* “Starting with your next bill, your charge will change from $9.99 per month to $59.88 per year”.

**Actual result:**

* “Starting with your next bill, your charge will change from $9.99 per year to $59.88” text is displayed.

**Notes:**

* See the attached screenshot.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4341)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18689/image_2021_12_17T15_13_34_121Z.png">image_2021_12_17T15_13_34_121Z.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/18882/image-20211230-085109.png">image-20211230-085109.png</a>

data-sync-user:
➤ Lisa Chan commented:

Bianca Hidecuti , thanks for bringing this to our attention! After reviewing the text, please let us know if you are aligned with the following expected text instead:

“Starting with your next bill, your charge will change from $9.99 per year to $59.88 per year”

It would provide further clarification to the user of the payment amount per product payment cycle, as well as cover the cases in which a user updates a monthly subscription to an annual one.

[...]

PR #11405, opened 2021-12-21
fix(auth-server): misplaced text in emails

xlisachan:
Because:
* “Starting with your next bill, your charge will change from $9.99 per year to $59.88” should include the previous payment cycle and new payment cycle in cases where a user moves from a monthly subscription to an annual subscription. It should look similar to “Starting with your next bill, your charge will change from $9.99 per month to $59.88 per year”

This commit:
* Fixes the issue in the `subscriptionUpgrade` subscription platform email as well as `subscriptionDowngrade`

Closes: #11374

## Checklist

## Screenshots (Optional)
Upgrade - old template
Screen Shot 2021-12-21 at 10 37 46 AM

Upgrade - new template
Screen Shot 2021-12-22 at 1 49 35 PM

Downgrade - old template
Screen Shot 2021-12-21 at 10 37 55 AM

Downgrade - new template
Screen Shot 2021-12-22 at 1 49 24 PM

While the issue only addressed updates to the `subscriptionUpgrade` email, revisions have also been included for the `subscriptionDowngrade` email as well.


jodyheavener (dismissed):
🤠

xlisachan:
> 🤠

Lol @jodyheavener! There may actually be further changes to this.

packages/fxa-auth-server/lib/senders/emails/templates/subscriptionDowngrade/en.ftl
```diff
@@ -10,7 +10,7 @@ subscriptionDowngrade-content-switch = You have successfully switched from { $pr
 # $paymentAmountNew (String) - The amount of the new subscription payment, including currency, e.g. $10.00
 # $productPaymentCycle (String) - The interval of time from the end of one payment statement date to the next payment statement date, e.g. month
 # $paymentProrated (String) - The one time fee to reflect the higher charge for the remainder of the payment cycle, including currency, e.g. $10.00
-subscriptionDowngrade-content-charge = Starting with your next bill, your charge will change from { $paymentAmountOld } per { $productPaymentCycle } to { $paymentAmountNew }. At that time you will also be given a one-time credit of { $paymentProrated } to reflect the lower charge for the remainder of this { $productPaymentCycle }.
+subscriptionDowngrade-content-charge = Starting with your next bill, your charge will change from { $paymentAmountOld } to { $paymentAmountNew } per { $productPaymentCycle }. At that time you will also be given a one-time credit of { $paymentProrated } to reflect the lower charge for the remainder of this { $productPaymentCycle }.
```
LZoog, line 13: Darn, [my original comment in your other PR](https://github.com/mozilla/fxa/pull/11410#discussion_r773507992) applies here. We have to update the string IDs.

LZoog (changes requested):
@xlisachan this actually affects the old templates as well, since this was found in production.

We need to change the strings in the old template HTML files and tests.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2196,7 +2196,7 @@ export class StripeHelper {
 
     const abbrevProductNew = await this.expandAbbrevProductForPlan(planNew);
     const {
-      interval: productPaymentCycle,
+      interval: productPaymentCycleNew,
```
xlisachan, line 2199: I'm not sure if I needed to change this.

xlisachan: @biancadanforth @chenba, I am not sure if I need to update `productPaymentCycle` in `stripe.ts`, can you please review and let me know if this should be included in the fix? Thank you! 🙂

chenba: The renaming make sense.  But I cannot find where `productPaymentCycleOld` for the template is coming from.

xlisachan: Thanks for your feedback, @chenba! I have updated the PR to include `productPaymentCycleOld` for the template. When you have a moment, can you please take another look? Thanks again!

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -2422,6 +2422,7 @@ export class StripeHelper {
     const {
       amount: paymentAmountOldInCents,
       currency: paymentAmountOldCurrency,
+      interval: productPaymentCycleOld,
```
xlisachan, line 2425: @chenba , I believe `productPaymentCycleOld` should be included here. Please review when you have a moment. Thank you!

chenba:
@xlisachan @LZoog how does QA verify a PR that include changes to both old and new templates?  IIUC only old _or_ new templates can be enabled on stage?

LZoog:
@chenba right now on staging, the new templates are enabled for `@mozilla.com` emails. Any others will receive the old templates on staging.

chenba (approved):
📧

chenba:
> @chenba right now on staging, the new templates are enabled for `@mozilla.com` email addresses. Any others will receive the old templates on staging.

I don't like it...I love it.

LZoog:
@chenba I didn't dig in here but I'm curious where `productPaymentCycleOld` is pulled in from the Stripe API response? I haven't worked with the emails with Stripe data much so maybe it's all good, just not sure where [that value would come from](https://stripe.com/docs/api/plans/object) 🤔

chenba:
> @chenba I didn't dig in here but I'm curious where `productPaymentCycleOld` is pulled in from the Stripe API response? I haven't worked with the emails with Stripe data much so maybe it's all good, just not sure where [that value would come from](https://stripe.com/docs/api/plans/object) 🤔

The subscription update emails are triggered by webhooks.   The old plan can be found in the `previous_attributes` of the webhook event data.

LZoog (approved):
😎

PR #11408, opened 2021-12-21
chore(auth-server): Add 'merge-ftl' to 'storybook-mjml' command

LZoog:
Because:
* When you check into a branch with new or updated FTL strings and you run storybook-mjml, strings will be missing since the file hasn't been concatenated locally yet

This commit:
* Adds 'grunt merge-ftl' to the start of 'storybook-mjml' script


xlisachan (approved):
🎉🎉🎉🎉🎉🎉🎉

Issue #11350, opened 2021-12-15
[email templates] Text in plain version of postVerifySecondary template different than HTML version and old template


data-sync-user:
**Build:**
Stage (Train 221)

**Preconditions:**
N/A

**Affected platforms:**
Windows 10 64-bit
macOS 10.14
Android 12
iOS 14.8

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/ and log in using an existing account.
1. Go to "Secondary email" section and click "Add" button.
1. Enter a valid email address and click "Save".
1. Go to your inbox to email "Confirm secondary email" and copy the verification code.
1. Paste the code into the verification code field and click "Verify".
1. Go to your inbox again and check the email "Secondary email added" both HTML and plaintext versions.
1. Compare with the old template too.

**Actual result:**
The text in the plain version is a little bit different than the HTML version and the old template.

**Expected result:**
The text in the plain version is similar with the HTML version and the old template.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4329)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/18624">old template.PNG</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/18625">postVerifySecondary - text difference.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

Hi Lisa ChanThanks for making the changes. I verified this issue on Stage (Train 222.0) under Windows 10 64-bit and I see there is still a very small difference between the old plaintext and new plaintext and HTML.

Old template:

!old template (532f832e-8b5f-4b44-9b71-7e76ec181af9).PNG|width=992,height=262!

New template:

!new template plaintext.PNG|width=991,height=320!

New HTML:

!new HTML.PNG|width=360,height=299!

PR #11410, opened 2021-12-21
fix(auth-server): text in plain version of postVerifySecondary template

xlisachan:
Because:

* Text in plain version of postVerifySecondary template was different than HTML version and old template

This commit:

* Revises the plain text to reflect correct wording

Closes: #11350

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/templates/postVerifySecondary/index.txt
+++ b/packages/fxa-auth-server/lib/senders/emails/templates/postVerifySecondary/index.txt
@@ -1,6 +1,6 @@
 postVerifySecondary-title = "Secondary email added"
 
-postVerifySecondary-description = "You have successfully added <%- secondaryEmail %> as a secondary email from your Firefox account. Security notifications and sign-in confirmations will now be delivered to both email addresses."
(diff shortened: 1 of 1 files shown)


packages/fxa-auth-server/lib/senders/emails/templates/postVerifySecondary/index.txt
```diff
@@ -1,6 +1,6 @@
 postVerifySecondary-title = "Secondary email added"
 
-postVerifySecondary-description = "You have successfully added <%- secondaryEmail %> as a secondary email from your Firefox account. Security notifications and sign-in confirmations will now be delivered to both email addresses."
+postVerifySecondary-description = "You have successfully verified <%- secondaryEmail %> as a secondary email from your Firefox account. Security notifications and sign-in confirmations will now be delivered to both email addresses."
```
LZoog, line 3: Unfortunately we have to update the ftl ID any time we update a string because the ID is now forever tied to this string in Pontoon. You can make this `postVerifySecondary-description-2` here and in the template.

LZoog (changes requested):

LZoog (approved):

Issue #10850, opened 2021-10-28
Coupons - [UI] Hook Up Apply Button to BE validation


data-sync-user:
Enable coupon validation when the user presses the Apply Button in the coupon component. 

If there is an Error message:

* display the error in the coupon component.

If there is no error the API will return the new payment data and possibly some metadata about the coupon.

* Update the payment details Price, Discount, Tax, Total. 
* Update the coupon component read only text to show coupon metadata

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4149)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/17296/Screen+Shot+2021-10-26+at+4.02.02+PM.png">Screen Shot 2021-10-26 at 4.02.02 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/17293/Screen+Shot+2021-10-28+at+4.13.51+PM.png">Screen Shot 2021-10-28 at 4.13.51 PM.png</a>

PR #11411, opened 2021-12-21
 feat(Coupon): add backend call to coupon apply btn 

StaberindeZA:
## Because

- The apply button should validate the coupon entered by the user

## This pull request

- Calls the invoice preview endpoint with the priceId and coupon code
  when the Apply button is pressed.

Closes: #10850

## Checklist

### DIFF
--- a/packages/fxa-payments-server/server/config/index.js
+++ b/packages/fxa-payments-server/server/config/index.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-'use strict';
+
 
 const fs = require('fs');
 const path = require('path');
--- a/packages/fxa-payments-server/src/components/CouponForm/index.stories.tsx
+++ b/packages/fxa-payments-server/src/components/CouponForm/index.stories.tsx
@@ -1,14 +1,17 @@
-import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
+import React, { useState } from 'react';
+
+import CouponForm from '.';
 import MockApp from '../../../.storybook/components/MockApp';
-import CouponForm from './index';
 import { Coupon } from '../../lib/Coupon';
+import { SELECTED_PLAN } from '../../lib/mock-data';
 
 storiesOf('components/Coupon', module).add('default', () => {
   const [coupon, setCoupon] = useState<Coupon>();
+
   return (
     <MockApp>
-      <CouponForm coupon={coupon} setCoupon={setCoupon} />
+      <CouponForm planId={SELECTED_PLAN.plan_id} coupon={coupon} setCoupon={setCoupon} />
     </MockApp>
   );
 });
--- a/packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
+++ b/packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
@@ -1,15 +1,17 @@
-import React, { useState } from 'react';
-import { render, cleanup, fireEvent } from '@testing-library/react';
+import React from 'react';
+import { render, cleanup, fireEvent, act } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import CouponForm from './index';
 import { Coupon } from '../../lib/Coupon';
+import { SELECTED_PLAN } from '../../lib/mock-data';
+import waitForExpect from 'wait-for-expect';
 
 afterEach(cleanup);
 
 describe('CouponForm', () => {
   it('renders as expected', () => {
     const subject = () => {
-      return render(<CouponForm coupon={undefined} setCoupon={() => {}} />);
+      return render(<CouponForm planId={SELECTED_PLAN.plan_id} coupon={undefined} setCoupon={() => {}} />);
     };
 
     const { queryByTestId } = subject();
@@ -19,16 +21,16 @@ describe('CouponForm', () => {
     expect(couponButton).toBeInTheDocument();
   });
 
-  it('shows an error message when invalid coupon code is used', () => {
+  it('shows an error message when invalid coupon code is used', async () => {
     const subject = () => {
-      return render(<CouponForm coupon={undefined} setCoupon={() => {}} />);
+      return render(<CouponForm planId={SELECTED_PLAN.plan_id} coupon={undefined} setCoupon={() => {}} />);
     };
 
     const { queryByTestId, getByTestId } = subject();
     fireEvent.change(getByTestId('coupon-input'), { target: { value: 'a' } });
     fireEvent.click(getByTestId('coupon-button'));
 
-    expect(queryByTestId('coupon-error')).toBeInTheDocument();
+    await waitForExpect(() => expect(queryByTestId('coupon-error')).toBeInTheDocument());
   });
 
   it('shows the coupon code and hides the input when a coupon is used', () => {
@@ -37,7 +39,7 @@ describe('CouponForm', () => {
       couponCode: '',
     };
     const subject = () => {
-      return render(<CouponForm coupon={coupon} setCoupon={(coupon) => {}} />);
+      return render(<CouponForm planId={SELECTED_PLAN.plan_id} coupon={coupon} setCoupon={(coupon) => {}} />);
     };
 
     const { queryByTestId } = subject();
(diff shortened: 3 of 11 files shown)


StaberindeZA:
For testing you can use the following promotion code and plan.
- Code: ASJQLIUQ
- Plan: price_1JfcqJBVqmGyQTMadYoEH25q.

packages/fxa-payments-server/src/lib/apiClient.test.ts
```diff
@@ -440,6 +442,63 @@ describe('API requests', () => {
...
+      const requestMock = nock(AUTH_BASE_URL).post(path).reply(200, expected);
+
+      expect(
+        await apiInvoicePreview({
+          priceId: 'price_kkljasdk32lkjasd',
+          promotionCode: 'INVALID',
+        })
+      ).toEqual(expected);
+      requestMock.done();
+    });
```
chenba, line 499: This test seems unnecessary.  It's the same test as the one above.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -1,29 +1,64 @@
...
       setCoupon({
-        amount: 200,
-        couponCode: '',
+        couponCode: couponCode,
+        amount,
       });
-    } else {
+    } catch (err) {
+      setCoupon(undefined);
       setError(true);
```
chenba, line 59: If I recall from the other PR, setting error to true here would render the error message about the promotion code being invalid or expired.  But the `err` caught here could be an API error, and in that case, we'd be misinforming the user.

biancadanforth:
@StaberindeZA , can you rebase this off of `main` since Ivo's PR merged? It'd make it easier to review.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
biancadanforth: I just tested this, and Barry is correct. It's kind of the difference between a 400 and a 500 server response.

I don't think it needs to block this PR, but perhaps it's worth discussing in the next design check-in if we want to add at least one other generic error message here in the event there was a server error (i.e. nothing is wrong with the code the user entered, and nothing they can do, perhaps other than wait for us to fix it, will change the result). Can you add a topic to the agenda to this effect Reino?

At the very least for this PR, perhaps we can capture the error in Sentry if it's an API error?

biancadanforth (approved):
R + WC (with changes). Thanks for the quick PR here Reino. I think it's looking pretty good (and it works!). There are a number of minor issues that I request inline, and a few overall:
* Please rebase off the latest `main` and resolve the merge conflicts.
* Looks like a coupon-related test is still failing in CI.
* The issue referenced in the commit description is not the correct issue (should be #10850 ); looks like the GitHub/Jira sync broke somehow.
* Nit: Can you rename the commit message to "feat(payments): …", since it’s on the payments server? This commit message is parsed in `release.sh` for generating our changelogs.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  try {
+    const { discount } = await apiInvoicePreview({
+      priceId,
+      promotionCode: couponCode
+    });
+
+    if(discount)
+      return discount.amount;
+
+    throw new Error('No discount for coupon');
```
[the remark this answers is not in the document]
StaberindeZA: Yup that makes sense. I'll make the change.

packages/fxa-payments-server/src/components/CouponForm/index.stories.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import MockApp from '../../../.storybook/components/MockApp';
+import CouponForm from './index';
+import { Coupon } from '../../lib/Coupon';
+
+storiesOf('components/Coupon', module).add('default', () => {
+  const [coupon, setCoupon] = useState<Coupon>();
+
+  const selectedPlan = {
```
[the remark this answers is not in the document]
StaberindeZA: Ah sweet, didn't know we had this. Thank you!

packages/fxa-payments-server/src/components/CouponForm/index.tsx
StaberindeZA: I thought it best to send the error to Sentry in the `checkCouponCode` (`checkPromotionCode`) function.

I did update the logic a bit to send a different error message. I'll create a separate issue to properly implement this, but at least we'll have something for now.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+const checkCouponCode = async (selectedPlan: Plan, couponCode: string) => {
+  const { plan_id: priceId } = selectedPlan;
+  try {
+    const { discount } = await apiInvoicePreview({
+      priceId,
+      promotionCode: couponCode
+    });
+
+    if(discount)
+      return discount.amount;
```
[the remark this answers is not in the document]
StaberindeZA: I actually think the linter prefers without the brackets.

packages/fxa-payments-server/src/components/PlanDetails/index.tsx
```diff
@@ -52,6 +52,8 @@ export const PlanDetails = ({
     interval_count
   );
 
+  console.log(config.featureFlags.subscriptionCoupons)
```
IvoJP, line 55: small nit in case you get to bianca's before this is merged, can we remove the console.log statement to keep the console clean? otherwise, non-blocker.

packages/fxa-payments-server/server/config/index.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-'use strict';
+
```
chenba, line 5: Was this line breaking something?

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
```diff
@@ -19,16 +21,16 @@ describe('CouponForm', () => {
...
-      return render(<CouponForm coupon={undefined} setCoupon={() => {}} />);
+      return render(<CouponForm planId={SELECTED_PLAN.plan_id} coupon={undefined} setCoupon={() => {}} />);
     };
 
     const { queryByTestId, getByTestId } = subject();
     fireEvent.change(getByTestId('coupon-input'), { target: { value: 'a' } });
     fireEvent.click(getByTestId('coupon-button'));
 
-    expect(queryByTestId('coupon-error')).toBeInTheDocument();
+    await waitForExpect(() => expect(queryByTestId('coupon-error')).toBeInTheDocument());
```
chenba, line 33: Another way is to wrap the code that causing React rendering in `act()`.  E.g. `await act(async () => fireEvent.click(someButton));`.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    throw new Error('No discount for coupon');
+  } catch (err) {
+    if (err instanceof APIError)
+      sentry.captureException(err);
+    throw err;
```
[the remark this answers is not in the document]
chenba: > Style on this `if` block is also off (missing brackets).
> 
> Also, unlike the auth server, the payments server reports all errors to Sentry, so I don't think you need to differentiate here if it's only for the purposes of capturing the error in Sentry.

The calling code catches this error, so I think explicit Sentry reporting would be needed.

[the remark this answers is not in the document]
StaberindeZA: OK cool. I've left it in for now.

packages/fxa-payments-server/server/config/index.js
StaberindeZA: Not that I can think of. I think I touched this file while debugging, and the linter auto deleted this. Should I put it back?

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
StaberindeZA: So I tried doing that, but for some reason it wasn't working. It was late at night though, so I might have done something else wrong. 😅

packages/fxa-payments-server/server/config/index.js
chenba: Nah.  Just curious on the change.

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
chenba: Works > doesn't work. 😄

PR #11414, opened 2021-12-21
fix(subscriptions): use ip only customs check on invoicePreview

chenba:
Because:
 - there's no email in the request

This commit:
 - switch to use ip only check with customs

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
@@ -397,7 +397,7 @@ export class StripeHandler {
     request: AuthRequest
   ): Promise<invoiceDTO.invoicePreviewSchema> {
     this.log.begin('subscriptions.previewInvoice', request);
-    await this.customs.check(request, 'previewInvoice');
+    await this.customs.checkIpOnly(request, 'previewInvoice');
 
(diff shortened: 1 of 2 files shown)


dannycoates (approved):

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -397,7 +397,7 @@ export class StripeHandler {
     request: AuthRequest
   ): Promise<invoiceDTO.invoicePreviewSchema> {
     this.log.begin('subscriptions.previewInvoice', request);
-    await this.customs.check(request, 'previewInvoice');
+    await this.customs.checkIpOnly(request, 'previewInvoice');
```
biancadanforth, line 400: Maybe add a test to assert this `checkIpOnly` is called in this handler?

chenba: > Maybe add a test to assert this `checkIpOnly` is called in this handler? (Since it does make a difference regarding whether the handler works or not)

https://github.com/mozilla/fxa/pull/11414/files#diff-b4fdef13ad1896c20f283ef9e7a78d4b6ca576474c06cdbc2afb090136a28deeR573-R577

PR #11424, opened 2021-12-22
 fix(scripts): add deps and fix route init for delete acct script

chenba:
Because:
 - the script was erroring due to
   - missing dependencies
   - change in the account routes export

This commit:
 - add the ProfileClient Firestore deps to the Container
 - fix account routes init
 - add a test

### DIFF
--- a/packages/fxa-auth-server/scripts/delete-account.js
+++ b/packages/fxa-auth-server/scripts/delete-account.js
@@ -22,15 +22,31 @@
 
 const { StatsD } = require('hot-shots');
 const { Container } = require('typedi');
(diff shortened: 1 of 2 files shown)


vbudhram (dismissed):
@chenba Possible to add a small test for this, maybe based on https://github.com/mozilla/fxa/blob/6a42d1d7343a79e4e05200793e7a91a9b67ca1be/packages/fxa-auth-server/test/scripts/verification-reminders.js#L33-L32 which just test it can run. Otw LGTM!

dannycoates (approved):

PR #11430, opened 2021-12-22
chore(ci): use node 16 in circleci

dannycoates:
Last time node 16 seemed to cause higher test instability so we rolled it back but maybe we've fixed enough kinks now.


dannycoates:
If this works well we can update the docker image version to 16 too

chenba (approved):

Issue #11335, opened 2021-12-14
Unhandled Stripe event received.


bbangert:
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/13612736/?referrer=github_plugin

```
Unhandled Stripe event received.
```

In this case, our code is looking for payment_method.automatically_updated, but Stripe is sending payment_method.card_automatically_updated which we should account for. The Stripe types in the SDK appear to be incorrect for the naming.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4321)

PR #11431, opened 2021-12-23
fix(subscriptions): update a payment method webhook event type

chenba:
Because:
 - there's a mismatch between an event type in Stripe's SDK types and
   docs and what's actually sent

This commit:
 - update the type to the one that's actually in the payload

Closes: #11335


biancadanforth (approved):
LGTM. Unusual bug since it's also in their TS.

PR #11443, opened 2021-12-23
fix(scripts): add new deps to delete acct script

StaberindeZA:
## Because

* The accountRoutes function has some new args that need to be added to
  the delete account script.

## This pull request

* Adds the missing deps to the account script.

Closes: # (issue number)

## Checklist


dannycoates:
not urgent right now but we should add a test that creates an account and then deletes it with this script

chenba (approved):
Thanks! 🎄

StaberindeZA:
I'll have a follow up task to create the tests and will  get it done early next week!

PR #11445, opened 2021-12-23
bug(auth): Less brittle partials

dschom:
## Because

- Translated text in partials was difficult to control and would often fallback to the default value.

## This pull request

- Changes the button partial so that text and l10n id & args can be passed into template
- Adds tests validating this works.

Closes: #10586

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
@@ -7,7 +7,7 @@
 <mj-section>
   <mj-column css-class="<%= locals.cssClass || undefined %>">
     <mj-button css-class="primary-button" href="<%- link %>">
-      <span data-l10n-id="<%- template %>-action">Click here</span>
+      <span data-l10n-id="<%- locals.l10nTemplateId || "" %>", data-l10n-args="<%= JSON.stringify(locals.l10nArgs) || "" %>" ><%- locals.action || "Click here" %></span>
     </mj-button>
   </mj-column>
 </mj-section>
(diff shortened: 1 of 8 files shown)


packages/fxa-auth-server/lib/senders/emails/partials/button/index.mjml
```diff
@@ -7,7 +7,7 @@
 <mj-section>
   <mj-column css-class="<%= locals.cssClass || undefined %>">
     <mj-button css-class="primary-button" href="<%- link %>">
-      <span data-l10n-id="<%- template %>-action">Click here</span>
+      <span data-l10n-id="<%- locals.l10nTemplateId || "" %>", data-l10n-args="<%= JSON.stringify(locals.l10nArgs) || "" %>" ><%- locals.action || "Click here" %></span>
```
LZoog, line 10: What do you think about something like this instead?

```
// any template file, let's use lowRecoveryCodes as a random example

<%- include('/partials/cadReminder/index.mjml', { 
  actionText: "Generate codes",
  actionl10nId: "lowRecoveryCodes-action"
}) %>
```

```
// button/index.mjml

<span data-l10n-id="<%- locals.actionl10nId %>"><%- locals.actionText %></span>
```

Correct me if I'm wrong, but I don't think what you have necessarily fixes the problem on staging. If we used the solution above, we'd have to edit every template file using `button` unfortunately (though I think your solution would need that as well), but I think to avoid dynamically rendered l10 IDs which are hard to maintain, and for the "correct" English text to always be the fallback instead of "Click here," it might be preferable to just pass in the entire ID and fallback text this way and avoid reliance on the `templateName`.

LZoog: commenting for cross-linking: #11472 / FXA-4372

Issue #10218, opened 2021-08-23
Set up MJML server for deployed auth-server Storybook builds


data-sync-user:
FxA has a mechanism for building and deploying our Storybook instances, however auth-server’s Storybook relies on a separate server to render and return compiled MJML/ERB templates. Locally this is accomplished with `yarn storybook-mjml`, which starts the server and Storybook. However when deployed it’s just a static Storybook build, no server.

Set up the MJML server, or find an alternate way for the static Storybook builds to render and display MJML/ERB templates.

Additionally, remove references to `localhost` when building for deployment, if there's an env var we can check or some straightforward way. This could be done in a separate ticket if preferred, or we can just change all `localhost:3030` references to `accounts.firefox.com`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3886)

data-sync-user:
➤ Les Orchard commented:

{quote}find an alternate way for the static Storybook builds to render and display MJML/ERB templates.{quote}

FWIW, I think that’s what will need to be done here. The little server I prototyped is only intended for local dev machines and not for any kind of external deployment. For a static storybook build, we’d need to do a static rendering of all the templates. It could be possible to come up with a webpack plugin that builds a list of templates from stories and renders them out as static files

Yogita98:
@jodyheavener @LZoog I can try to look into this if it isn't already picked up

jodyheavener:
Thanks @Yogita98! This was stumping me yesterday, so if you've got an idea I'd love to see it :)

Yogita98:
Hey @jodyheavener 
I did a little digging into a few webpack plugins and how we can create one to solve this issue. Do we want to handle fluent localization and ejs compile step as well as part of that plugin? I came across a[ mjml plugin](https://github.com/tdeNL/tde-webpack-mjml-plugin) and it's loading the file and exporting final html contents. We could extend this and add ejs compilation and fluent localization in it or we can make use of [mjml-browser](https://www.npmjs.com/package/mjml-browser) package. The only downside of using this package is that we won't be able to use `<mj-include>` tags as mentioned in the docs. However, we could probably overcome that with the help of [juice](https://github.com/Automattic/juice) library. Basically, we can look for the `<mj-include>` components in the file, extract the name of the stylesheet from it, and pass it down to juice. Juice will load the file and inline the styles wherever required. Let me know which approach do you think would efficiently handle this problem.

[...]

PR #11462, opened 2021-12-28
bug(auth) - Button text missing in email templates

dschom:
## Because
- The language translation strings defining button text would not take hold.

## Because

- Sometimes localization strings would not display.

## This pull request

- Makes sure that data-l10n-args are always applied with the current context to elements tagged with data-l10n-id
- Adds tests to validate fluent-localizer
- Validates that dynamic generation of data-l10n-id works
- Validates that extraction of key value pairs from .ftl works
- Validates that place holders in language strings are updated by the time they rendered
- Addresses separate issue found in the regex that extracts key values from .ftl files
- Adds a root level ignore for .mjml files

Closes: #10586

## Checklist


.prettierignore
```diff
@@ -1 +1,2 @@
 .yarn
+*.mjml
```
LZoog, line 2: I know we talked in Slack about this and how the one in auth-server doesn't seem to be doing its job, does this actually remove the formatting on save? 👀 We can probably remove `*.mjml` in `fxa-auth-server` if so, it's very strange that it isn't working though and that we'd have to add it to the root ignore...

packages/fxa-auth-server/lib/senders/emails/templates/test1/index.mjml
```diff
@@ -0,0 +1 @@
+<%- include('/partials/test/index.mjml') %>
```
LZoog, line 1: Hmm I'm not sure about having `test1` and `test2` templates in here... we do have one that's not an actual email, `_storybook`, that has that underline to denote that it's different than some of the others, but I think it would feel better for this to be in a separate directory, maybe even in `senders/emails/tests/`, but I'm wondering if we need these at all.  I think we need to just remove all dynamic references to l10n IDs like `data-l10n-id="<%- template %>-title` altogether. If we ever need to change the string, we have to also change the ID, making this really tricky to maintain. 😖 I haven't filed an issue for this yet, but I saw in the payments-server where we're also setting something dynamically through the ID and unfortunately that's going to break if we ever need to change the copy and update that ID (I can't find where I saw this to reference but I'll find it sometime soon).

I don't think we want extra `.ftl` files if possible too, since the l10n team is going to put in work to localize all of these strings and are going to have questions around what `{$testTerm}` is and might push back on this. If you'd like to leave all of this in anyway, maybe we can at least use less IDs and add a comment to let localizers know it's for dev testing purposes and they don't need translations.

.prettierignore
l0oDANo0l: I think it has to be in both places. From what I’ve observed on my machine, when I run from root, it auto formats the templates until I add this line. If I run from packages/fxa-auth-server, then all is good, since the `*.mjml` is already in the .prettierignore file there.

dschom: I think it has to be in both places. From what I’ve observed on my machine, when I run from root, it auto formats the templates until I add this line. If I run from packages/fxa-auth-server, then all is good, since the `*.mjml` is already in the .prettierignore file there.

packages/fxa-auth-server/lib/senders/emails/templates/test1/index.mjml
dschom: I also had concerns about the test1, test2 folders. I actually tried setting them as _test1, _test2, but then they would not get picked up for some reason. Possibly I can just piggyback off an existing template that is similar in nature for this PR, and side step issues surrounding these test folders.

I agree that we should not use the `data-l10n-id="<%- template %>-title` convention, it’s a bit too implicit for my liking. However, it is used quite a bit, and a motivation for this branch was fixing the issue without editing all the templates. I see phasing out the dynamic id as a secondary maintenance issue. After further testing, I see no technical reason why this shouldn’t work as is.

On a side note, I think there is future value in setting up a `senders/emails/tests` folder. There are two issues with our current tests. First they are repetitive, and second they may not fully exercise the logic in a partial template. Creating specific tests scenarios would allow our tests to be more targeted, specifically when it comes partials. 

I am not sure how I feel about your comments on language strings. I see your point, but I also think that ‘testing’ strings could be ignored by the translation team. Perhaps we should ask them? I do very much like your idea of not exporting the the `*.ftl` files if they are in the `senders/emails/tests` folder. I'd happy to address these things in this PR, even if it is beyond the scope of the initial issue.

xlisachan: I think tests should go into the directory: `fxa-auth-server/test/local/senders/email`. I see where you are going with these email template tests and I am for it, but I don't think this is how it should be written - let's all regroup?

dschom:
@LZoog @xlisachan as discussed, some of the code in `fleunt-localizer.ts` is setup in a way that configuring the class for testing is a bit tricky at the moment. I had to make some changes for issue #10218 that addressed these underlying problems. Once those changes are in place, I think addressing some the feedback you all provided concerning the test folder/file structure and language translation files will be much easier.

packages/fxa-auth-server/test/local/senders/emails/fluent-localizer.ts
```diff
@@ -0,0 +1,169 @@
+import chai, { assert } from 'chai';
```
LZoog, line 1: This could use a license comment up top.

packages/fxa-auth-server/grunttasks/ftl.js
```diff
@@ -10,6 +10,13 @@ module.exports = function (grunt) {
       src: ['lib/senders/emails/auth.ftl', 'lib/**/senders/emails/**/en.ftl'],
       dest: 'public/locales/en/auth.ftl',
     },
+    'ftl-test': {
```
LZoog, line 13: Noice 👌

LZoog (approved):
Great work, again. Thanks for writing those tests too - I think we can probably remove [this todo comment](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/test/local/senders/mjml-emails.ts#L5) as well, but otherwise it can be removed with that cleanup emails ticket. 🙂

I kind of feel like this qualifies as a "small" 5 pointer considering the work put into this and the Zoom call, and your earlier PR was a "big" 5 pointer. Reading [the descriptions](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-dev-process#estimation-and-point-values), it could qualify, but I'll leave that up to you.

dschom:
> I kind of feel like this qualifies as a "small" 5 pointer considering the work put into this and the Zoom call, and your earlier PR was a "big" 5 pointer. Reading [the descriptions](https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-dev-process#estimation-and-point-values), it could qualify, but I'll leave that up to you.

Agreed, there were certainly a couple extra things going on here aside from simply fixing the issue. I'd be happy to bump the points.

Issue #11069, opened 2021-11-18
Tax Reporting: New Stripe customers store location information in stripe


data-sync-user:
When a user subscribes for a product using stripe. Use zipcode from the card metadata to get state and plug into stripe

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4223)

data-sync-user:
➤ Bianca Danforth commented:

Matt Becker or Xenia Tovchykh Hello! Can you confirm that country and region  (i.e. state/province for the U.S. and Canada) are the only two pieces of information we need to add to the Stripe customer? What format should this data be in? Here are some examples:

Using country and region names:

country: "United States"
region: "California"or using country and region codes (ISO 3166-1 alpha-2 for the country code):

country: "US"
region: "CA"CC Benjamin Bangert Reino Muhl

data-sync-user:
➤ Xenia Tovchykh commented:

Hi Anna Sobiepanek Bianca Danforth I can confirm these are the only two pieces missing right now. The output format (“United States” vs “US”) is up to you. Thank you!

PR #11490, opened 2022-01-03
feat(subscriptions): set customer location with payment info

chenba:
Because:
 - we need the Stripe customer's location for tax purposes

This commit:
 - on a new subscription or a payment method update, set the customer's
   location with the available country and postal code

Closes: #11069

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -806,35 +806,24 @@ export class StripeHelper {
    * Returns a boolean indicating success.  It does not throw any exceptions as
    * this operation should not block any functionality.
    *
-   * Currently this only works on customers who use Stripe as their payment
-   * provider.
-   *
    * This will _overwrite_ any existing customer address.
    */
-  async setCustomerLocation(customer: Stripe.Customer): Promise<boolean> {
+  async setCustomerLocation({
+    customerId,
+    postalCode,
+    country,
+  }: {
+    customerId: string;
+    postalCode: string;
+    country: string;
+  }): Promise<boolean> {
     try {
-      const { postalCode, country } =
-        await this.extractCustomerDefaultPaymentDetails(customer);
-
-      if (!postalCode || !country) {
-        Sentry.withScope((scope) => {
-          scope.setContext('setCustomerLocation', {
-            customer: { id: customer.id },
-          });
-          Sentry.captureMessage(
-            `Cannot find a postal code or country for customer ${customer.id}`,
-            Sentry.Severity.Error
-          );
-        });
-        return false;
-      }
-
       const state = await this.googleMapsService.getStateFromZip(
         postalCode,
         country
       );
 
-      await this.updateCustomerBillingAddress(customer.id, {
+      await this.updateCustomerBillingAddress(customerId, {
         line1: '',
         line2: '',
         city: '',
@@ -847,7 +836,9 @@ export class StripeHelper {
     } catch (err: unknown) {
       Sentry.withScope((scope) => {
         scope.setContext('setCustomerLocation', {
-          customer: { id: customer.id },
+          customer: { id: customerId },
+          postalCode,
+          country,
         });
         Sentry.captureException(err);
       });
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import { ServerRoute } from '@hapi/hapi';
 import isA from '@hapi/joi';
+import * as Sentry from '@sentry/node';
 import { getAccountCustomerByUid } from 'fxa-shared/db/models/auth';
 import { AbbrevPlan } from 'fxa-shared/dist/subscriptions/types';
 import * as invoiceDTO from 'fxa-shared/dto/auth/payments/invoice';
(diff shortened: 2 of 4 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -806,35 +806,24 @@ export class StripeHelper {
...
-  async setCustomerLocation(customer: Stripe.Customer): Promise<boolean> {
+  async setCustomerLocation({
+    customerId,
+    postalCode,
+    country,
+  }: {
+    customerId: string;
+    postalCode: string;
+    country: string;
+  }): Promise<boolean> {
```
chenba, line 819: I changed this because a customer does not have a default payment method prior to the first successful subscription.  So the options were either to fetch the customer after the subscription creation and then pass it in here, or use the payment method object that's already available.

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -484,6 +486,26 @@ export class StripeHandler {
...
+      });
+    } else {
+      Sentry.withScope((scope) => {
+        scope.setContext('createSubscriptionWithPMI', {
+          customerId: customer.id,
+          subscriptionId: subscription.id,
+          paymentMethodId: paymentMethod?.id,
+        });
+        Sentry.captureMessage(
+          `Cannot find a postal code or country for customer ${customer.id}`,
```
biancadanforth, line 503: Nit: and/or?

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -4626,22 +4542,21 @@ describe('StripeHelper', () => {
...
+      const actual = await stripeHelper.setCustomerLocation({
+        customerId: customer1.id,
+        postalCode: expectedAddressArg.postalCode,
+        country: expectedAddressArg.country,
+      });
       assert.isFalse(actual);
-      sinon.assert.calledOnceWithExactly(
-        stripeHelper.extractCustomerDefaultPaymentDetails,
-        customer1
-      );
```
biancadanforth, line 4644: Can we assert that Sentry.withScope and scope.setContext are called with the expected values here?

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -561,6 +584,29 @@ export class StripeHandler {
       customer.id,
       paymentMethodId
     );
+
+    if (
+      paymentMethod?.billing_details?.address?.postal_code &&
+      paymentMethodCountry
+    ) {
+      this.stripeHelper.setCustomerLocation({
```
biancadanforth, line 592: Same thing here; wondering if we should await.

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -561,6 +584,29 @@ export class StripeHandler {
...
+        country: paymentMethodCountry,
+      });
+    } else {
+      Sentry.withScope((scope) => {
+        scope.setContext('updateDefaultPaymentMethod', {
+          customerId: customer!.id,
+          paymentMethodId: paymentMethod?.id,
+        });
+        Sentry.captureMessage(
+          `Cannot find a postal code or country for customer ${customer!.id}`,
```
biancadanforth, line 604: Same nit for and/or

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -4524,96 +4521,15 @@ describe('StripeHelper', () => {
...
```
biancadanforth, line 4616: Can we assert that Sentry.withScope and scope.setContext are called with the expected values here?

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -484,6 +486,26 @@ export class StripeHandler {
     const sourceCountry =
```
biancadanforth, line 486: I know you didn't add this here, but we already have `paymentMethodCountry` above -- is `sourceCountry` different? I'm wondering if we need to execute this line at all. `paymentMethodCountry` is just block scoped, but that could be changed.

biancadanforth (approved):
Thanks Barry; this looks great. Works locally for me for new subscriptions and updating payment methods for Stripe customers. I made a few small, non-blocking suggestions in line.

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
chenba: Since the `paymentMethodId` arg can be empty, we need this.  I suppose an extra thing we could do is to check if `paymentMethodCountry` already exist and have a value before we call a function to get the country.  I'll keep that in mind in case I'm in this neighborhood again, but I'm not going to make that change in this PR.

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
```diff
@@ -484,6 +486,26 @@ export class StripeHandler {
...
       this.stripeHelper.extractSourceCountryFromSubscription(subscription);
 
+    if (paymentMethod?.billing_details?.address?.postal_code && sourceCountry) {
+      this.stripeHelper.setCustomerLocation({
```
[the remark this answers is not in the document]
chenba: No await, for the reason you stated.  But! I do need to chain the error handling to a .catch since we are not awaiting.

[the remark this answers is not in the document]
chenba: Oh actually never mind, `setCustomerLocation` itself is `catch`ing.

packages/fxa-auth-server/test/local/payments/stripe.js
chenba: The argument passed to Sentry.withScope is a callback function that uses scope.setContext.  The test is already verifying the arguments for scope.setContext.  But since we don't have a reference to the callback function we cannot do an exact match there.

chenba: https://github.com/mozilla/fxa/pull/11490/files#diff-b2ea299307ae708731a29f26e148068eab268e1c06e58723c636bdfcf956704eR4540 is where setContext's call is asserted.

packages/fxa-auth-server/lib/routes/subscriptions/stripe.ts
chenba: I was going to make this change along with the other changes.  But since I won't be making the other changes I thought I needed, I'm not going to make this small one.

Issue #11318, opened 2021-12-13
Add backend support for google auth


data-sync-user:
From [https://docs.google.com/document/d/1pFzugkDOIR6eoXrBCx9FWWJhfnxFgWnRtk2mWFOp8DQ/edit#heading=h.5cfc2365xg33](https://docs.google.com/document/d/1pFzugkDOIR6eoXrBCx9FWWJhfnxFgWnRtk2mWFOp8DQ/edit#heading=h.5cfc2365xg33|smart-link) 

    POST /account/verifyIDToken
    
    To verify the ID token received from Google, we will add our own verification endpoint. This endpoint will take the token and verify it based on the requirements here. 
    
    
    NOTE: Google has a NodeJS library that can be used to verify tokens [1]. We can also temporarily use Google's verification endpoint
    
    If FxA does not have the user specified in the token, we will create an account in our backend and set `verifierSetAt` to 0 to specify that a password has not been set.
    We then will create a new FxA session for the user and return those credentials
    
    We will also need to add a database migration to add columns googleAuthAt and googleRefreshToken to the account table.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4314)

data-sync-user:
➤ Vijay Budhram commented:

Here is a demo of this working, https://drive.google.com/file/d/1y4zyYXIhqAE5dRc45ecFrDjYyTRDLtsG/view?usp=sharing ( https://drive.google.com/file/d/1y4zyYXIhqAE5dRc45ecFrDjYyTRDLtsG/view?usp=sharing|smart-link )

Issue #11314, opened 2021-12-13
Create client id for Firefox account on google


data-sync-user:
Per [https://developers.google.com/identity/sign-in/web/sign-in](https://developers.google.com/identity/sign-in/web/sign-in|smart-link) , we need to create a clientId on the google credentials page. We should have credentials for prod, stage, and maybe local dev?

1. Go to the [Credentials page](https://console.developers.google.com/apis/credentials).
1. Click **Create credentials > OAuth client ID**.
1. Select the **Web application** application type.
1. Name your OAuth 2.0 client and click **Create**
  1. Name=Firefox account

After creating we will need to add authorized javascript origins

URIs: [http://localhost:3030](http://localhost:3030), http://localhost

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4309)

data-sync-user:
➤ Jon Buckley commented:

We’ve created a client id for local dev, and Vijay Budhram is testing it now. stage/prod will come next week

data-sync-user:
➤ Vijay Budhram commented:

The new client id works, I’m going to bump this to the next train so that prod account gets made. A couple things that would be needed for the prod client

  1. Support account email
  2. Logo

data-sync-user:
➤ Wil Clouser commented:

Jon Buckley we’d like to land this in prod in Train 225.  We plan for this to be behind a feature flag at first launch.

data-sync-user:
➤ Dustin Lactin commented:

Created stage credential, stored secrets in sops, and created PR to add env variables: https://github.com/mozilla-services/cloudops-deployment/pull/4516 ( https://github.com/mozilla-services/cloudops-deployment/pull/4516|smart-link )

data-sync-user:
➤ Dustin Lactin commented:

Created prod credential, stored in sops and merged PR to add env vars.

PR #11499, opened 2022-01-04
feat(google): Add backend support for google auth

vbudhram:
## Because

- We need to support to FxA backend 
- This is WIP and needs to be cleaned up and add tests

## This pull request

- Add new db columns `googleAuthAt` `googleId` on account table (used to associate google account to fxa account)
- Updates shared db models to set this property and unset it
- Adds new endpoint to verify a google id token and create fxa account as needed
- Adds auth-client function for new endpoint

Closes: https://github.com/mozilla/fxa/issues/11318

## Checklist


packages/db-migrations/databases/fxa/patches/patch-123-124.sql
```diff
@@ -0,0 +1,10 @@
+SET NAMES utf8mb4 COLLATE utf8mb4_bin;
+
+CALL assertPatchLevel('123');
+
+ALTER TABLE accounts
```
bbangert, line 5: We had discussed adding the googleId/etc to the email table along with the email/verification status the google Id verified, was it more compelling to leave it on the accounts table? If we store this on the account table, and then also store the Apple Auth on the main account table, does that mean we're intentionally prohibiting account merging with dual login (Google and Apple) and the merge must choose the valid auth scheme? I think its fine to make that a requirement of merging, though if that does happen, I can imagine people accidentally creating accounts again by clicking the wrong one.... needing yet more merging.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -30,6 +30,8 @@ import { AuthLogger, AuthRequest } from '../types';
 import emailUtils from './utils/email';
 import requestHelper from './utils/request_helper';
 import validators from './validators';
+import { Auth } from '@hapi/hapi';
+const { OAuth2Client } = require('google-auth-library');
```
bbangert, line 34: nit: Should be a typescript format import.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1569,6 +1571,86 @@ export class AccountHandler {
...
+        const email = payload.email;
+
+        try {
+          accountRecord = await this.db.accountRecord(payload.email);
+        } catch (err) {
+          if (err.errno === 102) {
+            const { hex16: emailCode, hex32: authSalt } =
+              await this.generateRandomValues();
+            const [kA, wrapWrapKb] = await random.hex(32, 32);
+            accountRecord = await this.db.createAccount({
```
bbangert, line 1608: Why not update this call to include the google info to avoid the other 2 async calls?

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1569,6 +1571,86 @@ export class AccountHandler {
 
     return {};
   }
+
+  async loginThirdParty(request: AuthRequest) {
```
bbangert, line 1575: `loginOrCreate...`? 😉

packages/db-migrations/databases/fxa/patches/patch-123-124.sql
```diff
@@ -0,0 +1,10 @@
...
+ADD COLUMN googleId VARCHAR(255) DEFAULT NULL,
```
bbangert, line 6: We need an index on this field for the query.

packages/db-migrations/databases/fxa/patches/patch-123-124.sql
vbudhram: > was it more compelling to leave it on the accounts table

I believe so. I think the biggest issue was the FxA allows users to change their email address and that would cause some extra complexity here.

Instead of adding to the accounts table, we could create a new table, `providers` and store `uid`, `id`, `provider`, `authAt` properties?

packages/fxa-auth-server/config/index.ts
```diff
@@ -137,6 +137,15 @@ const conf = convict({
       format: Boolean,
     },
   },
+  googleAuthConfig: {
+    clientId: {
+      default:
+        '218517873053-th4taguk9dvf03rrgk8sigon84oigf5l.apps.googleusercontent.com',
```
vbudhram, line 143: This is currently pointing to a clientId I generated, we should change this once https://github.com/mozilla/fxa/issues/11314 lands.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1718,6 +1801,19 @@ export const accountRoutes = (
       },
       handler: (request: AuthRequest) => accountHandler.login(request),
     },
+    {
+      method: 'POST',
+      path: '/account/login/third_party',
+      options: {
+        validate: {
+          payload: {
+            token: isA.string().required(),
```
dannycoates, line 1810: Are we planning on using this endpoint for multiple IDPs? I assume so from the name. It might be worth considering what the payload might look like once we add more. Adding a `provider` field (or something) now will make the logic easier later when we go from 1 to 2+.

dannycoates (approved):
I like!

packages/fxa-auth-server/lib/routes/account.ts
vbudhram: @dannycoates correct, I was hoping to have just one endpoint that we call and setup new users etc. I'll add the `provider` field 👍🏽

dannycoates: cool, i was thinking `provider` wouldn't be optional and we'd send 'google' from the start. Then we don't have to worry about the (small) transition time where no provider implicitly means google.

Issue #11378, opened 2021-12-17
Coupons - Add Promotion Code to Cancel Subscription Metrics Event Properties


data-sync-user:
We'd like to know whether a cancelled subscription was created with a coupon.  Once FXA-4342 is done, the frontend should have that info.  Add the promotion code to the subscription cancellation metrics event as an event property in the format of, e.g., `'promotionCode': 'FRIEND20'`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4343)

data-sync-user:
➤ Chelsea Lewis commented:

Reino Muhl Please let Edward Cho know when these are available for testing in staging so that he and his team can verify that they are emitting as expected.

data-sync-user:
➤ Yeonjoo Yoo commented:

Reino Muhl when I cancel the subscriptions in the staging site, the promotion code that I used to subscribe did not show in cancel subscription metric data in event_properties (promotionCode). For example, I used VPN10 coupon for this test run ( https://docs.google.com/spreadsheets/d/1mFzJHbQ7PmYulS4lGWEcHZeJNTTeDyBA4gnjh6PVioY/edit?usp=sharing ) which shows the promotion code used. When I cancel the subscription, I get this ( https://docs.google.com/spreadsheets/d/1uuomTJrkzmF_U3UUt_MGCQJ0R7p7xxoMC3tfYl1w-H0/edit?usp=sharing ) without the promotion code.  Would you let me know if I am missing something or I should look at different place to track if a canceled subscriptions with promotions? 

I used the same table you provided in https://mozilla-hub.atlassian.net/browse/FXA-4333 ( https://mozilla-hub.atlassian.net/browse/FXA-4333|smart-link )  and here is the doc ( https://docs.google.com/document/d/1Pam8P43J4c5y4Ea5G4HqgvupRvXIkP5IxAZWivUq3X0/edit?usp=sharing ) that I made to track my test run. The example I mentioned above is CASE3 in the doc. 

Daniel Thorn Winnie Chan Sean Rose Is there a way to test how the canceled subscriptions in staging site would show up in the all subscription table?

cc: Edward Cho

data-sync-user:
➤ Barry Chen commented:

[...]

PR #11500, opened 2022-01-04
feat(metrics): include promo code for cancel sub events

chenba:
Because:
 - we need to include a subscription's promotion code in the
   subscription cancellation metrics events

This commit:
 - add the subscription's promotion code as an event property to the
   cancellation events

Closes: #11378


chenba:
Note that this PR builds on https://github.com/mozilla/fxa/pull/11490.  Only the newer commit is for this PR.

IvoJP (approved):

Issue #11269, opened 2021-12-07
Subscription information not updated on profile immediately, starting v1.213.5


data-sync-user:
TL;DR - Guardian statsd events seem to indicate a regression in profile cache-ing starting v1.213.5

cc [~accountid:6113f9a38c15ca006fe13cee] who spotted this.

- - -
Guardian is an oauth RP customer of FxA that serves the Mozilla VPN clients. It is a private repo hosted here: [https://github.com/mozilla-services/guardian-website](https://github.com/mozilla-services/guardian-website). We have a function FXA.getUserProfile ([https://github.com/mozilla-services/guardian-website/blob/main/server/lib/fxa.ts#L94-L133)](https://github.com/mozilla-services/guardian-website/blob/main/server/lib/fxa.ts#L94-L133))

In semi-pseudocode it does the following:

```
GET oauthUrls (https://accounts.firefox.com/.well-known/openid-configuration)
GET oauthUrls.userinfo_endpoint - https://profile.accounts.firefox.com/v1/profile
fxaProfile = response.body

// TODO: FIXME: Temporary workaround to FxA caching issue.
// See https://github.com/mozilla/fxa/issues/2463 
if (!fxaProfile.subscriptions) {
  GET https://profile.accounts.firefox.com/v1/subscriptions
  if (this returns subscriptions) {
    log a statsd event
    update the fxaProfile with the subscriptions info
  }
}
return fxaProfile
```

The TODO was written 2 years ago and the FxA issue referenced was closed Oct 2019. However, due to tech debt accumulation we had not got round to doing this clean up.

[~accountid:6113f9a38c15ca006fe13cee] started looking into cleaning this up and noticed the statsd metric and thought to check it. Here is [the graph](https://earthangel-b40313e5.influxcloud.net/d/nc6RwIt7z/all-statsd-events-wip?orgId=1&from=now-1y&to=now&viewPanel=5).

!Screenshot from 2021-12-07 17-20-13.png|width=200,height=183!

We had no hits on this metric until August 25, which coincides with v1.213.5 deploy.

So, it appears that the original issue has been re-birthed and fxaProfile doesn't always get updated with subscriptions immediately.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4297)
[...]

PR #11501, opened 2022-01-04
chore(tests): enable subscribe test on dev

dannycoates:
Thanks @chenba for fixing the stripe metadata!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,6 +271,7 @@ jobs:
       - image: redis
       - image: memcached
       - image: circleci/mysql:5.7.31
+      - image: jdlk7/firestore-emulator
     environment:
       NODE_ENV: dev
     steps:
--- a/packages/functional-tests/tests/misc.spec.ts
+++ b/packages/functional-tests/tests/misc.spec.ts
@@ -5,7 +5,6 @@ test.describe('severity-1', () => {
     pages: { relier, login, subscribe },
   }, { project }) => {
     test.skip(project.name === 'production', 'prod needs a valid credit card');
-    test.fixme(project.name === 'local', 'needs correct product');
     test.slow();
     await relier.goto();
     await relier.clickSubscribe();


vbudhram (approved):

dannycoates:
I think we're hitting some kind of a race on the stripe customer here. From the auth-server logs:

```
{
  "status": 404,
  "errno": 176,
  "path": "/v1/oauth/mozilla-subscriptions/customer/billing-and-subscriptions",
  "lang": "en-US,en;q=0.5",
  "agent": "Mozilla/5.0 (X11; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0",
  "remoteAddressChain": ["127.0.0.1"],
  "t": 8,
  "uid": "b77f544c87b547c384c4f0717577e555",
  "keys": false,
  "method": "get"
}
```

where 176 is `UNKNOWN_SUBSCRIPTION_CUSTOMER`

according to the logs it looks like this is happening in prod too

chenba:
> I think we're hitting some kind of a race on the stripe customer here. From the auth-server logs:
> 
> ```
> ...
> ```
> 
> where 176 is `UNKNOWN_SUBSCRIPTION_CUSTOMER`
> 
> according to the logs it looks like this is happening in prod too

That's an expected error.  The payments frontend uses that error to determine whether a FxA user is a subscriber.

chenba:
When the test intermittently fails for me locally it's not the same failure as CI.  😩 

```
    Error: expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      15 |     await relier.clickEmailFirst();
      16 |     await login.submit();
    > 17 |     expect(await relier.isPro()).toBe(true);
         |                                  ^
      18 |   });
      19 |
      20 |   test('content-server mocha tests', async ({ target, page }, { project }) => {
```

dannycoates:
> When the test intermittently fails for me locally

I needed to restart my services to pick up the stripe capabilities change

chenba:
> > When the test intermittently fails for me locally
> 
> I needed to restart my services to pick up the stripe capabilities change

Yeah, they've been restarted.  And the test more often than not does pass.

biancadanforth:
FWIW I ran the test on this branch too, and while I did get a timeout error, it was on an earlier line, and I had to run the test 80 times to get one failure. Could be a difference in hardware as I am on an M1 Max MBP.

```
  1) [local] › misc.spec.ts:4:3 › severity-1 › subscribe and login to product ======================

    Timeout of 90000ms exceeded.

    Pending operations:
      - page.fill at pages/products/index.ts:5:22

    page.fill: Target closed
    =========================== logs ===========================
    waiting for selector "[data-testid="name"]"
    ============================================================

       9 |     await relier.goto();
      10 |     await relier.clickSubscribe();
    > 11 |     await subscribe.setFullName();
         |                     ^
      12 |     await subscribe.setCreditCardInfo();
      13 |     await subscribe.submit();
      14 |     await relier.goto();

        at SubscribePage.setFullName (/Users/bdanforth/Projects/fxa/packages/functional-tests/pages/products/index.ts:5:22)
        at /Users/bdanforth/Projects/fxa/packages/functional-tests/tests/misc.spec.ts:11:21

```

biancadanforth:
I found this [GitHub discussion](https://github.com/microsoft/playwright/issues/10832) where a user was having a similar issue in Firefox only, which ultimately [converged on the solution of not waiting for navigation explicitly](https://github.com/microsoft/playwright/issues/11167#issuecomment-1005848094) as that is handled automatically with page.click navigations.

I noticed there's a similar pattern of `Promise.all(..., waitForNavigation)` for all 3 failures, so I removed that call and reran the tests 240 times and it has passed each time. Worth a shot @dannycoates ?

biancadanforth:
Also @chenba had an interesting thought that this could perhaps be related to #11269 , which @bbangert plans to investigate during this sprint.

dannycoates:
sorry for the red herring, y'all. it was failing because there was no firestore emulator running. 🤦‍♂️

biancadanforth:
> sorry for the red herring, y'all. it was failing because there was no firestore emulator running. 🤦‍♂️

Thanks for the update -- though the firestore emulator runs by default in local dev, so it was running when I ran the test locally. I guess it wasn't running in your CI environment? I also don't understand how that would cause an intermittent failure rather than a permafail. 🤔

dannycoates:
it failed locally occasionally because stripe was rate limiting me

Issue #11406, opened 2021-12-21
Make /settings/avatar/change redirect to /settings/avatar/


data-sync-user:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1746357, older FF browsers use that link from Sync preferences to allow users to change avatars.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4346)

data-sync-user:
➤ Vijay Budhram commented:

fromt mtg:

Lets try to do this next sprint since this is a broken link in production.

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 223.2] under Windows 10 64-bit. When clicking the avatar icon on the about:preferences#sync page the user is redirected to Profile Picture section on the Settings page. See new situation:

!avatar from sync page - new situation.gif|width=1905,height=1012!

PR #11504, opened 2022-01-04
fix(settings): redirect /settings/avatar/change to /settings/avatar/

xlisachan:
Closes: #11406

## Checklist

### DIFF
--- a/packages/fxa-settings/src/components/App/index.tsx
+++ b/packages/fxa-settings/src/components/App/index.tsx
@@ -8,7 +8,7 @@ import LoadingSpinner from 'fxa-react/components/LoadingSpinner';
 import AppErrorDialog from 'fxa-react/components/AppErrorDialog';
 import * as Metrics from '../../lib/metrics';
 import { useAccount, useConfig, useInitialState } from '../../models';
-import { Router } from '@reach/router';
+import { Redirect, Router } from '@reach/router';
 import Head from 'fxa-react/components/Head';
 import PageSettings from '../PageSettings';
 import PageChangePassword from '../PageChangePassword';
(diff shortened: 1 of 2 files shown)


packages/fxa-settings/src/components/App/index.tsx
```diff
@@ -83,6 +82,7 @@ export const App = ({ flowQueryParams, navigatorLanguages }: AppProps) => {
               to="/settings#connected-services"
               noThrow
             />
+            <Redirect from="/avatar/change" to="/settings/avatar/" noThrow />
```
LZoog, line 85: So does `Redirect`'s `from` follow the [`HomePath`](https://github.com/mozilla/fxa/blob/main/packages/fxa-settings/src/constants/index.tsx#L6) and `to` does not? 🧐

packages/fxa-settings/src/components/App/index.test.tsx
```diff
@@ -22,42 +28,58 @@ const appProps = {
...
+        route: `/settings/avatar/change`,
+      }
+    );
 
-  expect(flowInit).toHaveBeenCalledWith(true, {
-    deviceId: DEVICE_ID,
-    flowId: FLOW_ID,
-    flowBeginTime: BEGIN_TIME,
+    screen.debug();
+    expect(screen.getByTestId('flow-container')).toBeInTheDocument();
```
LZoog, line 83: I think we have this on every page. I know the test isn't passing yet but maybe test for some avatar-related element here similar to the functional test.

I'm also thinking we probably don't need both a functional test and a unit test for this but you've already got most of the work done so keeping both is OK. I like unit tests for the line coverage reports and functional tests to be more like end to end testing for some flow that a user is going through but I know we already had another redirect test in the functional tests.

packages/fxa-settings/src/components/App/index.tsx
xlisachan: Yep, I believe because `Redirect` does not have to be a child of a Router, the `HomePath` does not get automatically prepended to `to`.

vbudhram, line 85: You could probably leave a comment here for future readers why this exists, Sync settings in Firefox link directly to the avatar page.

vbudhram (approved):
@xlisachan Tested this locally and works as described. If you wanted to do the additional App component testing you can file a follow up bug.

xlisachan:
Follow-up to include additional testing for the App component: [#11518](https://github.com/mozilla/fxa/issues/11518)

Issue #11481, opened 2021-12-30
[subplat][l10n] The ”Total amount” is cut by couponing section for some European countries while using Firefox browser


data-sync-user:
**Affected browsers:**

* Firefox;

**Affected FxA versions:**

* 1.222.0;

**Prerequisites:**

* Be in VPN available European country with non-English default language i.e “Spain”;
* Have the browser window in full screen mode;
* Have an FxA account without subscription;

**Steps to reproduce:**
1. Load “Set Up Your Subscription” page for that account;
2. Observe couponing section;

**Expected result:**

* There are no display or layout issues;

**Actual result:**

* The “Total” section is cut.

**Notes:**

* If the coupon is applied, the coupon confirmation green message is cut.
* Not reproducing for English and Italian;
* This is qa-major because the user cannot see the total amount he needs to pay for the product;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4375)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18889/cut.png">cut.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/18938/image-20220104-100446.png">image-20220104-100446.png</a>

data-sync-user:
➤ Chelsea Lewis commented:

Ivo Plamenac will take a look. To assist,

Valentina Virlics can you please try to reproduce this once more to let us know what the resolution is on your screen? If you can look for your exact viewport size in devtools that would be helpful!

data-sync-user:
➤ Valentina Virlics commented:

Chelsea Lewis Ivo Plamenac This issue is not window size dependent, only browser and language dependent, reproducing only on Firefox and for some non-en languages.

Display dimensions like: 1920x1080, 1920x1200, 1800x1120, were all affected.

Today, for some reasons, new flow is only affected on MacOS (no matter of the window/display dimensions), old flow is affected everywhere.

Screenshot from old flow:

!image-20220104-100446.png|width=1184,height=658!

data-sync-user:
➤ Chelsea Lewis commented:

Thank you so much Valentina, really appreciate your review here. Ivo is taking a look, we’re hoping to have this completed within the train 223.

data-sync-user:
➤ Valentina Virlics commented:

Verified as fixed on Train 223, while using Firefox browser and European countries, on different platforms.

Posfix screenshot below.

!image-20220106-093213.png|width=986,height=672!

Issue #11440, opened 2021-12-23
[subplat]  “Remove” discount button is cut on iPad landscape mode


data-sync-user:
**Device used:**

* iPad Air Pro 2 - iOS 14.4

**Affected browsers:**

* Safari;

**Affected FxA versions:**

* 1.222.0;

**Prerequisites:**

* Be in a country where Mozilla VPN is available;
* Be on the Set up your Subscription page; 

**Steps to reproduce:**
1. Rotate the iPad in landscape mode;
2. Apply a discount;
3. See “Remove” button;

**Expected result:**

* The button has no UI or layout issues.

**Actual result:**

* The button is cut.

**Notes:**

* Also, the user cannot scroll.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4361)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18827/1640269593506.JPEG">1640269593506.JPEG</a>

data-sync-user:
➤ Valentina Virlics commented:

Verified as fixed on the FxA Train 223 while using iPad in landscape mode. Tried different browsers.

The “Remove” discount button is no longer cut and a scroll is available for navigation.

Issue #11438, opened 2021-12-23
[subplat] Product details and couponing section overlaps the payment details at browser resize


data-sync-user:
**Affected versions:**

* all;

**Affected FxA versions:**

* 1.222.0;

**Prerequisites:**

* Be in a country where Mozilla VPN is available;
* Be on the Set up your Subscription page; 

**Steps to reproduce:**
1. Resize the browser to a width between 768px->844px;
2. Observe the Product details/Payment details sections;

**Expected result:**

* There are no UI or layout issues at browser resize.

**Actual result:**

* The product details and payment details overlap.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4359)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18825/1over2.gif">1over2.gif</a>

data-sync-user:
➤ Bianca Danforth commented:

Valentina Virlics : Thanks for filing this issue. Is this display issue only present upon window resize, or do the UI components overlap even if you just open the page at 768px?

data-sync-user:
➤ Bianca Danforth commented:

For the MVP, we could consider reverting some of Ivo’s changes to improve the landscape view in mobile; this would probably be the quickest fix. Ivo Plamenac : Can you attach a screenshot of what the landscape view in mobile looked like before versus now?

This is a regression that would affect checkout even without the new coupon component.

data-sync-user:
➤ Valentina Virlics commented:

Bianca Danforth Hey and thanks for looking over this.

I can confirm that the UI components overlap even if you just open the page directly at 768px.

Happy holidays!

data-sync-user:
➤ Valentina Virlics commented:

Verified as fixed on the FxA Train 223 using different browsers and locales.

Resizing the browser causes NO ui issues.

Postfix video below.

!resizebr.gif|width=1661,height=878!

Issue #11437, opened 2021-12-23
[subplat] “Show details'' button displayed in Product details section when the browser is in full screen mode


data-sync-user:
**Affected versions:**

* Safari;
* Chrome;

**Affected FxA versions:**

* 1.222.0;

**Prerequisites:**

* Be in a country where Mozilla VPN is available;
* Have the browser window in full screen mode
* Be on the Set up your Subscription page; 

**Steps to reproduce:**
1. Observe the Product details section;

**Expected result:**

* There is no “Show details” button in the Product details section when the browser window is in full screen mode;

**Actual result:**

* “Show details” button is displayed in the Product details section when the browser window is in full screen mode;

**Notes:**

* When clicking on the “Show details” button, the product details section expands and a cut “Hide details” button is displayed;
* Reproducible for both old and new Set up your subscription flows;
* See the attached screenshots.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4358)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18824/Screen+Shot+2021-12-23+at+09.47.37.png">Screen Shot 2021-12-23 at 09.47.37.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/18823/Screen+Shot+2021-12-23+at+09.48.11.png">Screen Shot 2021-12-23 at 09.48.11.png</a>

data-sync-user:
➤ Bianca Hidecuti commented:

Verified this as fixed on Train 1.223.0 - Staging, while on Chrome and Safari browsers. There is no “Show details” button in the Product details section when the browser window is in full screen mode.

See postfix screenshot. 

!Screen Shot 2022-01-06 at 09.26.32.png|width=1785,height=1039!

PR #11505, opened 2022-01-05
fix(payments): coupon Component Overlap

IvoJP:
## Because

- We want the components to not overlap depending on browser/device/language

## This pull request

- refactors templates and updates css to prevent components from floating over each other

Closes: #11481 , #11437 , #11438 , #11440 

## Checklist


packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -75,78 +75,76 @@ export const CouponForm = ({ planId, coupon, setCoupon }: CouponFormProps) => {
   };
 
   return (
-    <section
+    <div
```
chenba, line 78: huh does this make a difference wrt styles?

chenba:
image

small issue: there's a border or `<hr>` overlapping just below the total.

chenba:
> small issue: there's a border or `<hr>` overlapping just below the total.

non-blocking.  I was in desktop browser mode but with a narrow-ish window.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
IvoJP: not that i can tell? none of our css was selecting on element type, just classes. i imagine there are some inherent differences (obivously) but i think we are pretty explicit with our css and it appears to be rendering correctly on brave (chromium based) and firefox as well across different screen sizes on both.

chenba (approved):
Looked fine when I tried the various emulation options in Firefox's dev tools. 👍

packages/fxa-payments-server/src/components/CouponForm/index.tsx
chenba: I was curious why you made that change if it did not make a difference.

Issue #11513, opened 2022-01-05
[email templates] Text difference in postVerifySecondary template


data-sync-user:
There is still a very small difference between the HTML and plaintext for postVerifySecondary template.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4385)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18983/new+HTML.PNG">new HTML.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/18982/new+template+plaintext.PNG">new template plaintext.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 223.2] under Windows 10 64-bit. There is no difference between HTML and plaintext anymore. See new situation:

!postVerifySecond - new situation plain.PNG|width=959,height=304!

PR #11514, opened 2022-01-05
fix(auth-server): Text difference in postVerifySecondary template

xlisachan:
The old template reflects the word as 'for' instead of 'from'.

Note: this fix creates a new ftl ID, `postVerifySecondary-content`.

Closes: #11513

## Checklist


LZoog (approved):
Ohh nice catch. The bug describes a difference in the HTML and plaintext and this PR changes both the HTML and plaintext, but I looked into it and this LGTM.

PR #11522, opened 2022-01-05
bug(auth): Don't fail when l10n id is missing from en bundle

dschom:
##Because:
- Failing on a missing l10n id was probably too aggressive at the moment

##This commit:
- Comments out the error that requires all data-l10n-ids to be present in the en bundle

Closes: #11523

## Checklist

We are seeing some failing emails on stage despite all tests passing. This needs a bit more investigation.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
@@ -113,10 +113,10 @@ class FluentLocalizer {
       const args: any = {};
       const message = baseBundle.getMessage(attr.id);
 
-      // Require that all data-l10n-id values are present in en file.
-      if (!message) {
-        throw new Error('Missing data-1l0n-id detected. l10n-id=' + attr.id);
-      }
+      // TODO: Require that all data-l10n-id values are present in en file.
+      // if (!message) {
+      //   throw new Error('Missing data-l10n-id detected. l10n-id=' + attr.id);
+      // }
 
       if (message && message.value instanceof Array) {
         message.value
--- a/packages/fxa-auth-server/test/local/senders/emails/fluent-localizer.ts
+++ b/packages/fxa-auth-server/test/local/senders/emails/fluent-localizer.ts
@@ -137,7 +137,7 @@ describe('fluent localizer', () => {
     });
 
     describe('test template /test-missing-l10n-id', () => {
-      it('fails with missing l10n id', async () => {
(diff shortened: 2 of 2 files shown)


packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -113,10 +113,10 @@ class FluentLocalizer {
...
       const message = baseBundle.getMessage(attr.id);
 
-      // Require that all data-l10n-id values are present in en file.
-      if (!message) {
-        throw new Error('Missing data-1l0n-id detected. l10n-id=' + attr.id);
-      }
+      // TODO: Require that all data-l10n-id values are present in en file.
+      // if (!message) {
+      //   throw new Error('Missing data-l10n-id detected. l10n-id=' + attr.id);
+      // }
```
LZoog, line 119: I think I see why you're saying the `en/auth.ftl` file is required.

If you check out the [`en/`](https://github.com/mozilla/fxa-content-server-l10n/tree/master/locale/en/) and [`en_US/`](https://github.com/mozilla/fxa-content-server-l10n/tree/master/locale/en/) directories, they don't contain `main.ftl` or `settings.ftl` that you can see in the [`templates/`](https://github.com/mozilla/fxa-content-server-l10n/tree/master/locale/templates) directory. I'm pretty sure we just use the English fallback text in the UI components.

Maybe we can just run this check in development as a helper?

LZoog (approved):
Approving for the point release with a follow up maintenance task being filed. Thanks @dschom for jumping on this!

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
dschom: Sorry, I was under the incorrect assumption that en/*.ftl would always be present. The code sort of reads this way. If you look at this [line](https://github.com/mozilla/fxa/blob/54e59718efd9e4b78b8f38c927724704a9c3c0a9/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts#L49), you'll see for language negotiation 'en' is always the default language. So fluent would actually handle the fall backs for us, if we chose to deploy 'en'. However, it has become apparent that we do not deploy 'en' to  [fxa-content-server-l10n](https://github.com/mozilla/fxa-content-server-l10n).

Note that, this also means that the chunk of code following the error is no longer useful. The nice thing about that chunk of code was that it would have made the data-l10n-args attributes optional.

PR #11526, opened 2022-01-05
Revert "bug(auth): Don't fail when l10n id is missing from en bundle"

dschom:
Reverts mozilla/fxa#11522


LZoog (approved):
Context: we want to remove another piece too, so we're reverting this and will make the change in a single PR for easier dot releasing

Issue #11529, opened 2022-01-05
Emails: enable enforced l10n-ids for dev only, enable Fluent tests


data-sync-user:
We have a chunk of code enforcing that data-l10n-ids are present in the en FTL file and tries to automatically apply `data-l10n-args` that are used by the language translation string. This code was rolled back in the following PR, because the assumption that language strings for the default locale, en, was incorrect. This locale is not deployed in stage or production, and the placeholder text is used instead. 
[https://github.com/mozilla/fxa/pull/11530](https://github.com/mozilla/fxa/pull/11530|smart-link) 

Assuming that we want the place holder text to be displayed and not the english translation text the following could be done to further improve this PR.

When in ‘dev’ mode:
1. Validate that the text for the translation string exists in the public/locales/en/auth.ftl
2. Validate that the text for the language translation string matches the placeholder text. (This may be more difficult than it sounds.)
3. Validate that variables referenced by the l10n translation string are available in the current context.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4394)

data-sync-user:
➤ Dan Schomburg commented:

The following pull request reverted a bunch of tests and changes that were addressed to fix issue with improper text in email templates. This can be used as basis for future improvements to resolution of data-l10n-ids and data-l10n-args:

https://github.com/mozilla/fxa/pull/11530 ( https://github.com/mozilla/fxa/pull/11530|smart-link )

LZoog:
Note: I don't think data-l10n IDs should be "dev only" per se after seeing how we're handling l10n testing in payments.

This relates to FXA-4477 (a spike not in Github) which is, "Investigate making Fluent/extraction more consistent across packages & adding to fxa-react" which includes how we're testing some of this.

cknowles-admin:
FxA is no longer synchronizing all issues between Jira and Github.  We are closing open issues and will selectively synchronize in the future.

PR #11530, opened 2022-01-05
bug(auth): Don't fail when l10n id is missing from en bundle

dschom:
Because:
- Failing on a missing l10n id was probably too aggressive at the moment

This commit:
- Comments out the code that requires all data-l10n-ids to be present in the en bundle
- Comments out the code that auto generates data-l10n-args

Additional Info:
We are seeing some failing emails on stage despite all tests passing. This needs a bit more investigation.

## Because

-

## This pull request

-

Closes: # (issue number)

## Checklist

## Screenshots (Optional)


LZoog (approved):
LGTM. I'm sad we have to take this out right now as well as those tests but we'll get them back in with FXA-4394 / #11529 (a follow up, mentioning here for the cross-link). Thanks!

PR #11542, opened 2022-01-06
fix(subscriptions): do addr lookup for only US and CA

chenba:
Because:
 - accounting only need the province/state for CA and US customers

This commit:
 - skips the address lookup on new subscription or payment method update
   when the customer is not from CA or US

Co-authored-by: Bianca Danforth <bdanforth@mozilla.com>


biancadanforth (approved):
LGTM R+

PR #11544, opened 2022-01-06
fix(ts): fix auth-server TS errors

chenba:


biancadanforth (approved):
I filed #11545 to avoid the need for such PRs in the future. Thank you Barry for fixing my mistakes.

Issue #11498, opened 2022-01-04
[email templates] Legal and Privacy links are not displayed in the plaintext version of the templates


data-sync-user:
**Build:**
Stage (Train 222)

**Preconditions:**
The user has made a subscription for Mozilla VPN.

**Affected platforms:**
Windows 10 64-bit

**Steps to reproduce:**

1. Go to your inbox and open the “Mozilla VPN payment confirmed” email. 
1. Compare the plaintext version with the HTML version of the email.

**Actual result:**
The plaintext version doesn’t contain Legal and Privacy links that are in the HTML version but it contains a link that sends to a pdf Privacy doc.

**Expected result:**
Shouldn’t the plaintext version be the same with the HTML version?

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4381)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18943/different+links.PNG">different links.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 224.1] under Windows 10 64-bit. The “Legal“ and “Privacy“ links are displayed in the plaintext version and the “Privacy notice” link is displayed in the HTML version.

!privacy notice in HTML.PNG|width=517,height=379!

!legal and privacy new situation.PNG|width=923,height=328!

PR #11549, opened 2022-01-06
fix(auth-server): Legal and Privacy links are not displayed in the plaintext version of the templates

xlisachan:
Closes: #11498

Legal, Privacy, and Privacy notice links have been added into the plaintext version of the email templates, as reflected in the HTML version. 

Please note, as there is already an existing ftl ID for `subplat-privacy-plaintext` for the string `Privacy notice`, the ID created for the `Privacy` in the plaintext version is `subplat-privacy-website-plaintext` as the link is `https://www.mozilla.org/privacy/websites/`.

Additional minor updates have been included in this PR (missing colons that are included in previously created ftl IDs).

## Checklist

## Screenshots (Optional)
Screen Shot 2022-01-06 at 3 31 34 PM

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
@@ -1,4 +1,5 @@
 subplat-automated-email = This is an automated email; if you received it in error, no action is required.
+subplat-privacy-notice = Privacy notice
 subplat-privacy-plaintext = Privacy notice:
 subplat-update-billing-plaintext = { subplat-update-billing }:
 # Variables:
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
@@ -22,4 +22,6 @@ subplat-privacy-policy-plaintext = { subplat-privacy-policy }:
 subplat-cloud-terms = { -product-firefox-cloud } Terms of Service
 subplat-cloud-terms-plaintext = { subplat-cloud-terms }:
 subplat-legal = Legal
+subplat-legal-plaintext = { subplat-legal }:
 subplat-privacy = Privacy
+subplat-privacy-website-plaintext = { subplat-privacy }:
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
@@ -77,6 +77,11 @@
               &nbsp;&nbsp;&bull;&nbsp;&nbsp;
             <% } %>
 
+            <% if (locals.productName) { %>
+              <a href="<%- subscriptionPrivacyUrl %>" class="footer-link" data-l10n-id="subplat-privacy-notice">Privacy notice</a>
+              &nbsp;&nbsp;&bull;&nbsp;&nbsp;
+            <% } %>
+
             <% if (locals.isCancellationEmail) { %>
               <a href="<%- reactivateSubscriptionUrl %>" class="footer-link" data-l10n-id="subplat-reactivate">Reactivate subscription</a>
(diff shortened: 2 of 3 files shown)


packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
```diff
@@ -22,4 +23,6 @@ subplat-privacy-policy-plaintext = { subplat-privacy-policy }:
 subplat-cloud-terms = { -product-firefox-cloud } Terms of Service
 subplat-cloud-terms-plaintext = { subplat-cloud-terms }:
 subplat-legal = Legal
+subplat-legal-plaintext = { subplat-legal }:
```
vbudhram, line 26: Not really for this issue, but wonder if we could have stripped out the `:` on these strings and resuse the value in the html version?

xlisachan: FxA email template uses the same string values in both the HTML and plaintext versions for Legal and Privacy. I had added `:`s to strings to be consistent with the rest of the strings within the subplat email templates.

LZoog: I think we should leave them like we had them to reflect parity of the old templates, but, I think it's worth running by whomever we're going to run our [confirmation needed doc](https://docs.google.com/document/d/1dGwY1yOHk4X9yoXmg4sTidlzZ5c-wLUdTzMLiH7Pv4E/edit) too. 😬

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
```diff
@@ -77,6 +77,11 @@
               &nbsp;&nbsp;&bull;&nbsp;&nbsp;
             <% } %>
 
+            <% if (locals.productName) { %>
+              <a href="<%- subscriptionPrivacyUrl %>" class="footer-link" data-l10n-id="subplat-privacy-notice">Privacy notice</a>
+              &nbsp;&nbsp;&bull;&nbsp;&nbsp;
```
LZoog, line 82: Man I really don't like this `&nbsp;&nbsp;&bull;&nbsp;&nbsp;` pattern or using `&bull;` as a separator for links but I know that's the pattern we set here in the old templates and it doesn't matter much 😆

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
```diff
@@ -1,4 +1,5 @@
 subplat-automated-email = This is an automated email; if you received it in error, no action is required.
+subplat-privacy-notice = Privacy notice
 subplat-privacy-plaintext = Privacy notice:
```
LZoog, line 3: I wonder what the implications are around changing `subplat-privacy-plaintext` to equal `{ subplat-privacy-notice }:` (might be good to run by @peiying2 just so we know for sure), but leaving them separate as you did is probably safest since that ID was already created.

LZoog (approved):
LGTM. I'd definitely leave a comment on the original ticket for QA letting them know we fixed this only in the new templates since the old templates will be removed soon.

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
xlisachan: Same.

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
xlisachan: For consistency within subplat email templates, they will end with colons. It has been noted within our document.

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
xlisachan: To avoid any implications, I have left the separate.

Issue #11560, opened 2022-01-07
[subplat] l10n - Replace payment method header hardcoded prefix with separate message.


data-sync-user:
The l10n team requested changes to the Payment Method Header with prefix translation text.

The original discussion can be found [here on Github](https://github.com/mozilla/fxa-content-server-l10n/pull/540#discussion_r772116836).

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4411)

PR #11562, opened 2022-01-07
fix(subscriptions): l10n fix payment method header

StaberindeZA:
## Because

- Hard coded prefix can't be translated for different locales.

## This pull request

- Removes hard coded prefix from header, and instead provides a new
  message that can be translated by.

Closes: #11560

## Checklist


chenba (approved):
👍

PR #11566, opened 2022-01-08
docs(README): remove duplicate docs

clouserw:
Because:
* We have formal docs in another repository

This commit:
* Removes the duplicate docs

So, before Jody left, he converted a bunch of the sections of our README into mdx which I put in the ecosystem docs.  Example: https://mozilla.github.io/ecosystem-platform/tutorials/development-setup .  It looks really nice, not just the `tip` boxes and such, but note the tabs for macos, ubuntu, windows.  We can do more fancy stuff like that than we can in our default readme, not to mention it scales a lot better to not have everything in the same document.

That said, I don't want to move anyone's cheese, so, I'm looking for review before landing this.  Cheers!


vbudhram (approved):
@clouserw I don't have any issues with moving these docs. Was there a reason not to move the bit about scripts?

clouserw:
No, I just didn't have a good place for it.  I think it should be moved.

Issue #10853, opened 2021-10-28
Coupons - remove feature flag subscriptionCoupons


data-sync-user:
remove the feature flag subscriptionCoupons

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4152)

PR #11048, opened 2021-11-17
feat(payments): Add feature flag for Coupons

IvoJP:
## Because

- We want to have the ability to develop the coupons feature while retaining the ability to hide it in produciton

## This pull request

- Adds a feature flag and a dummy div that will render and appear in tests

## Issue that this pull request solves

Closes:  issue number is tied to the epic so I dont want this PR to close it

## Checklist

_Put an `x` in the boxes that apply_

- [X] My commit is GPG signed.
- [X] If applicable, I have modified or added tests which pass locally.
- [X] I have added necessary documentation (if appropriate).
- [X] I have verified that my changes render correctly in RTL (if appropriate).

PR #11576, opened 2022-01-10
feat(payments): remove Coupons Feature Flag

IvoJP:
## Because

- We no longer need the ability to hide coupons

## This pull request

- Removes the feature flag that hides the coupon feature

Closes: #10853 

## Checklist


biancadanforth:
@IvoJP : Thanks for the PR; it'd probably be a good idea to make sure QA can verify an active/valid coupon in prod before removing the flag. What do you think?

packages/fxa-auth-server/config/index.ts
```diff
@@ -857,14 +857,6 @@ const conf = convict({
...
       default: false,
     },
-    coupons: {
-      enabled: {
-        doc: 'feature flag for enabling coupons',
-        default: false,
-        env: 'COUPONS_ENABLED',
-        format: Boolean,
-      },
-    },
```
biancadanforth, line 867: Huh; I guess this was never used anywhere after it was added in #11048 . 😅 At least we separately blocked coupons in the front-end via the other flag.

biancadanforth (approved):
LGTM, but I don't think we should merge this until QA has verified coupons in prod. CCing @annasob in case she feels differently.

Also, as a follow-up, we may want to remove these flags from the ops repos (e.g. [payments flag](https://github.com/mozilla-services/cloudops-infra/blob/7b1e5fcaa459025eb1a3102596756264c936ec8f/projects/fxa/k8s/values/payments.yaml#L12-L13)) as well? We could remove the PayPal feature flag(s) while we're in there.

StaberindeZA:
@annasob should we wait for confirmation in Production before merging this, or could we merge this PR?

annasob:
Chelsea will talk to product about making a 1% off coupon for testing. Lets
wait on this.

bbangert:
@IvoJP Sounds like we're good for this to land, just needs a rebase for conflicts!

Issue #10299, opened 2021-08-31
Localize image alt text


data-sync-user:
We should localize the alt text for images in the new templates. This could be tricky with the new MJML/FTL format, but let's at least look into it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3924)

Issue #11489, opened 2022-01-03
Better organize email tests


data-sync-user:
The test for the new email templates are all included in `fxa-auth-server/test/local/senders/mjml-emails.ts`. While it was adapted from the test for the previous email template, there is room for improvement (e.g., separating tests to individual files and/or directories), as there is currently a lot of repetition.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4377)

cknowles-admin:
FxA is no longer synchronizing all issues between Jira and Github.  We are closing open issues and will selectively synchronize in the future.

Issue #10758, opened 2021-10-19
Update ecosystem-platform docs, emails + l10n


data-sync-user:
After the old emails have been retired, let's review a couple of pages on the ecosystem-platform docs and make sure they're up to date.

https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-email

https://mozilla.github.io/ecosystem-platform/docs/fxa-engineering/fxa-localization#extraction-process-deep-dive

We could also add a little more detail around l10n, like how to look at the Pontoon page, Fluentproject docs and things to know, Gettext to Fluent migration gotchas, and communication with the l10n team.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4124)

Issue #10617, opened 2021-10-05
Update auth-server email docs


data-sync-user:
After all emails have finished being converted and we’ve retired the old email templates, we should update the fxa-auth-server README email template section.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4062)

data-sync-user:
➤ Vijay Budhram commented:

from mtg:

Yes, please!

PR #11579, opened 2022-01-11
fix(auth-server): localize image alt text

xlisachan:
Closes: #10299

## Current images within email templates
Updates in this pull request localize the following image alt text:

FxA email template layout may include the following images:
* Firefox logo (note: this image is of the logo only)
* Sync Devices image

SubPlat email template layout includes the following images:
* Firefox logo (note: this image is of the logo and verbiage)
* Mozilla logo

The following partials and template include images:
* `appBadges` (partial) - badge images
* `cadReminder` (partial) - device image (same image in `postVerify`)
* `icon` (partial) - icon of subscription product
* `postVerify` (template) - device image (same image in `cadReminder`)

## Approach
Per [fluent documentation](https://projectfluent.org/dom-l10n-documentation/dom-overlays.html), child elements must be annotated with the `data-l10n-name` attribute. Image example provided by fluent documentation is as follows:
Screen Shot 2022-01-12 at 3 02 48 PM

Per [mjml documentation](https://documentation.mjml.io/#mj-html-attributes), we are able to set the `data-l10n-id` and `data-l10n-name` using `mj-html-attributes`. See example below:
Screen Shot 2022-01-12 at 4 39 42 PM

Therefore, when the MJML tags are translated into HTML, it should look as follows:
Screen Shot 2022-01-12 at 3 08 23 PM

Note (regarding `mjml-fluent`): As all of the email templates have now been converted/localized to the new stack and the image alt texts are the only elements left to localize, I don't believe there is a need to utilize [`mjml-fluent`](https://github.com/jodyheavener/mjml-fluent)/parts of it within this PR at this time. However, we may want to re-consider during the email revamp cleanup.

Note (regarding test): As we have a [ticket](https://github.com/mozilla/fxa/issues/11489) to better organize/separate our email tests, this pull request only spot checks whether or not emails have their respective alt texts (eg, FxA layouts only include either the Firefox logo or Sync devices image and not the Mozilla logo, SubPlat layouts include Firefox and Mozilla logos, etc).

~~To do:~~
~~- Localize image alt text for `appBadges`~~
~~- Check that image alt text appear when running `yarn write-emails` with "enabledEmailAddress" set to "@testuser.com" in `dev.json`~~
~~- Add respective tests~~
~~- Review `mjml-fluent` idea approach~~

## Checklist

Update localization of image alt text for `PageTwoStepAuthentication` component
* `fxa-settings/src/components/PageTwoStepAuthentication/en-US.ftl`
* `fxa-settings/src/components/PageTwoStepAuthentication/index.tsx`

Updates below were moved to PR #11630
~~`subscriptionFirstInvoiceDiscount`-related things:~~
~~* `fxa-auth-server/config/dev.json`~~
~~* `fxa-auth-server/lib/metrics/amplitude.js`~~
~~* `fxa-auth-server/lib/senders/templates/_versions.json`~~
~~* `fxa-auth-server/test/local/senders/mjml-emails.ts`~~


packages/fxa-settings/src/components/PageTwoStepAuthentication/en-US.ftl
```diff
@@ -22,8 +22,10 @@ tfa-scan-this-code = Scan this QR code using one of <linkExternal>these
 # This is the image alt text for a QR code.
 # Variables:
 #   $secret (String) - a long alphanumeric string that does not require translation
-tfa-qa-code-alt = Use the code { $secret } to set up two-step authentication in
-  supported applications.
+# DEV NOTE: Set image alt text per fluent/react documentation, do not use the below as an example
+tfa-qa-code-alt = Use the code { $secret } to set up two-step authentication in supported applications.
+tfa-qa-code =
+  .alt = { tfa-qa-code-alt }
 tfa-button-cant-scan-qr = Can’t scan code?
```
xlisachan, line 29: While the ftl ID `tfa-qa-code-alt` was translated, it was never included in the component and, therefore, was not utilized. We have revised the component within this PR to correctly localize the image alt text based on [fluent/react documentation](https://github.com/projectfluent/fluent.js/wiki/Localized). Can you please let us know if this is the best way to make this revision for the ftl ID? @peiying2
cc: @LZoog

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
```diff
@@ -1,3 +1,7 @@
+subplat-header-firefox-logo = <img data-l10n-name="fxa-logo-firefox" alt="Firefox logo">
```
xlisachan, line 1: Per [fluent documentation](https://projectfluent.org/dom-l10n-documentation/dom-overlays.html), this is the way to include an image, we just wanted to doublecheck the following:
(1) `Firefox logo` would be translated with just this line (line 1)?
(2) If the alt text is translated in line 1, we don't believe line 2 (`.alt = { -brand-firefox } logo`) is necessary and should be removed (as well as line 4). Can you please confirm? Thank you!
cc: @LZoog

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
```diff
@@ -1,33 +1,33 @@
-<%# This Source Code Form is subject to the terms of the Mozilla Public 
-  # License, v. 2.0. If a copy of the MPL was not distributed with this 
+<%# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
```
xlisachan, line 2: Auto-format updated this.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
```diff
@@ -1,33 +1,33 @@
...
   # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
 
 <mj-include path="./css/appBadges/index.css" type="css" css-inline="inline" />
 
-<% locals.iosUrl = locals.iosUrl || locals.appStoreLink %> 
-<% locals.androidUrl = locals.androidUrl ||  locals.playStoreLink %> 
+<% locals.iosUrl = locals.iosUrl || locals.appStoreLink %>
+<% locals.androidUrl = locals.androidUrl ||  locals.playStoreLink %>
```
xlisachan, line 8: Auto-format updated these lines as well.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
```diff
@@ -1,33 +1,33 @@
...
 
 <mj-section>
   <mj-group css-class="app-badges">
 
     <% if (locals.iosUrl) { %>
       <mj-column>
         <mj-image
-          css-class="app-badge"
+          css-class="app-badge-ios"
```
xlisachan, line 16: This `css-class`has been updated to utilize within `mj-html-attributes`.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
```diff
@@ -1,33 +1,33 @@
...
+          alt="Download Firefox on the App Store"
         >
         </mj-image>
       </mj-column>
     <% } %>
     <% if (locals.androidUrl) { %>
       <mj-column>
         <mj-image
-          css-class="app-badge"
+          css-class="app-badge-android"
```
xlisachan, line 27: This `css-class`has also been updated to utilize within `mj-html-attributes`.

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -41,7 +41,7 @@ const productMetadata = {
   'product:privacyNoticeDownloadURL': SUBSCRIPTION_PRIVACY_URL,
 };
 
-const MESSAGE = {
+const locals = {
```
xlisachan, line 44: This was added to check that the image alt texts for `productNameNew` and `productNameOld` were included in the `subscriptionUpgrade`/`subscriptionDowngrade` emails.

peiying2:
@eemeli could you help out with questions on localizing the image alt text and their respective ftl IDs?

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.mjml
```diff
@@ -10,6 +10,7 @@
       <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
     </mj-raw>
     <mj-title><%- subject %></mj-title>
+    <%- include('/partials/images.mjml') %>
```
LZoog, line 13: Commenting here to cross-link:

We should document how to add alt text to images and why we're doing it this way. We can do this in FXA-4124 / #10758 or FXA-4062 / #10617

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
LZoog: Thanks for the catch! Is there a reason we needed to change this from `MESSAGE` to `locals`, do you mean it wouldn't work without changing it or did this just seem like a more appropriate name? I thought `locals` was a keyword in EJS, the name change seems nice but I think we'd want to propagate it further by changing all `MESSAGE` references to `locals` instead of creating duplicate variable like `const MESSAGE = locals`, but it's possible I'm just missing something here 🙂

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -189,6 +196,10 @@ const TESTS: [string, any, Record<string, any>?][] = [
       { test: 'include', expected: 'another device' },
       { test: 'include', expected: decodeUrl(configHref('privacyUrl', 'cad-reminder-first', 'privacy')) },
       { test: 'include', expected: decodeUrl(configHref('supportUrl', 'cad-reminder-first', 'support')) },
+      { test: 'include', expected: 'alt="Firefox logo"' },
+      { test: 'include', expected: 'alt="Devices"' },
```
LZoog, line 200: Thanks for these tests!

packages/fxa-settings/src/components/PageTwoStepAuthentication/index.tsx
```diff
@@ -215,12 +215,18 @@ export const PageTwoStepAuthentication = (_: RouteComponentProps) => {
...
-                  <img
-                    className="mx-auto w-48 h-48 qr-code-border"
-                    data-testid="2fa-qr-code"
-                    src={totpInfo.result.qrCodeUrl}
-                    alt={`Use the code ${totpInfo.result.secret} to set up two-step authentication in supported applications.`}
-                  />
+                  <Localized
+                    id="tfa-qa-code"
+                    vars={{ secret: totpInfo.result.secret }}
+                    attrs={{ alt: true }}
```
LZoog, line 221: Aww yeah, get those bonus points. Thanks for this too.

packages/fxa-settings/src/components/PageTwoStepAuthentication/en-US.ftl
LZoog: @peiying2 let me know in Slack or otherwise if we need to explain this further. I asked @xlisachan to ask you about this one mostly for future reference on what would be the least amount of work for us and for translators - I was wondering if changing this:

```
tfa-qa-code-alt = Use the code { $secret } to set up two-step authentication in
  supported applications.
```

... to this:
```
tfa-qa-code-alt = 
  .alt = Use the code { $secret } to set up two-step authentication in supported applications.
```
was possible, or any other alternative. I'm guessing it's not since we're changing what's tied to the ID, and what Lisa changed it to works, we're just asking because I want to document FxA + Fluent best approaches.

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
LZoog: If this is because we need the selector to be unique, can we use IDs instead of unique classes with the same styles?

LZoog (commented):
I just had a couple of questions. I can circle back tomorrow morning to r+ so this makes the sprint, there's not blockers per se but my main hold up is the placeholder question.

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/en.ftl
xlisachan: I'm not sure if the alt text would literally appear as `{ -brand-firefox } logo` if we wrote it like that?

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
xlisachan: I don't believe so - per [mjml documentation](https://documentation.mjml.io/?html#mj-html-attributes), the `mj-html-attributes` tag allows us to add custom attributes on any html tag of the generated html, using css selectors.

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
xlisachan: It did not seem weird until you put it that way. An update is being made. Thanks!

packages/fxa-auth-server/lib/senders/emails/partials/appBadges/index.mjml
LZoog: Does it let you use the css selector `#app-badge-ios` and `#app-badge-android`? I think we should just use IDs like that then instead of creating extra classes if that's possible, it is less confusing that way IMO.

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -113,6 +118,8 @@ const MESSAGE = {
   ],
 };
 
+// const MESSAGE = locals;
```
LZoog, line 121: Missed comment

LZoog (approved):
LGTM, great work here!

Issue #11577, opened 2022-01-10
[subplat] Tweak customer location format in Stripe


data-sync-user:
After testing both FXA-4222 and FXA-4223 in stage (Train 223), I noticed a few discrepancies between the formats for Stripe versus PayPal customers that should be fixed ideally before Train 223 deploys to prod.

Specifically:

* We are recording the `city` for new PayPal customers, when we do not need this information.
* We are recording the long names for the `state` for new PayPal customers (e.g. "Pennsylvania") and the short names (e.g. "PA") for new Stripe customers. We should be consistent and use the short name.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4418)

data-sync-user:
➤ Bianca Danforth commented:

Reino Muhl Yes, https://mozilla-hub.atlassian.net/browse/FXA-4224 ( https://mozilla-hub.atlassian.net/browse/FXA-4224|smart-link ) is our last ticket, and it’s in progress. Thanks for checking.

Issue #11070, opened 2021-11-18
Tax Reporting: One-off script to update all current customers with location metadata


data-sync-user:
Write one-off script to run to populate all existing customers 

* iterate through all stripe customers
  * if they paid with stripe pull out their payment zip/postal code and convert that to state, country updating their customer object
  * if they paid with paypal (spike will determine how to get location data), add the state/county information into their stripe customer object
* Depends on all helpers for [stripe/paypal - existing] above.
* [Ops] run script

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4224)

data-sync-user:
➤ Barry Chen commented:

Assigning this to myself since I’ve started writing this script as part of https://mozilla-hub.atlassian.net/browse/FXA-4221 ( https://mozilla-hub.atlassian.net/browse/FXA-4221|smart-link ).

data-sync-user:
➤ Reino Muhl commented:

Just a note, for the 1 off script, once you integrate the getStateFromZip method created as part of FXA-4225, you'll need to add the Config and AuthLogger to the Container as done in fxa/packages/fxa-auth-server/bin/key_server.js, since these are required by the GoogleMapsService class.

data-sync-user:
➤ Bianca Danforth commented:

For updating existing PayPal users in Stripe:

  * I already have a +list of all PayPal FxA uids through 12/5+ ( https://mozilla-hub.atlassian.net/browse/FXA-4218?focusedCommentId=488265 ). The SQL query for the list of PayPal uids will need to be re-run for the intervening period (12/6 - present) and appended to the existing list, since my derived table does not update.
    * Also the query and/or the result should be modified to remove duplicate uid entries.

  * While we can write it in advance, we should not run this one-off script until we are certain that new PayPal customer locations from https://mozilla-hub.atlassian.net/browse/FXA-4222 ( https://mozilla-hub.atlassian.net/browse/FXA-4222|smart-link ) are correctly being recorded in Stripe in production.

data-sync-user:
➤ Bianca Danforth commented:

I heard back from Finance re: how to record the state/province for PayPal users when we don’t have  reliable or any information from our metrics logs. We should list their state in the customer address as "unknown" (see this comment ( https://mozilla-hub.atlassian.net/browse/FXA-4218?focusedCommentId=494552 )).

data-sync-user:
➤ Bianca Danforth commented:

[...]

PR #11584, opened 2022-01-11
FXA-4418 chore(auth): make customer address format consistent

biancadanforth:
Because:
* We want to update the Stripe customer address the same way for Stripe and PayPal customers.

This commit:
* Removes the city from a PayPal user's address, as it is not needed.
* Converts the state for a PayPal user to a state code.

Closes: #11577

## Checklist


biancadanforth:
The CI failure is unrelated; see the fix in #11585 .

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
```diff
@@ -17,7 +17,9 @@ import Container from 'typedi';
 
 import { ConfigType } from '../../../config';
 import error from '../../error';
+import { COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP } from '../../payments/customer-locations';
```
chenba, line 20: ...I guess customer-locations isn't used _only_ in the one time script now.  😢

chenba (approved):
LGTM.  We did talk about having customer-location being imported by the one time script only; this PR pulls it into the service.  Maybe consider moving those constants elsewhere?

packages/fxa-auth-server/lib/routes/subscriptions/paypal.ts
biancadanforth: IIUC, this will load the CustomerLocations module now and anything else it loads as part of the fxa-auth-server. Is the concern about just unnecessarily loading this module when I just need a static file? E.g. unnecessary loading. Or are there risks around dependencies (though I don't use `Container` here)?

biancadanforth: Either way, it's a good point (though I just noticed we also import this object from `customer-locations.ts` in lib/routes/subscriptions/stripe.ts). In the interest of time for today's dot release, since it's not blocking, I will handle moving this in FXA-4224 (#11070 ).

chenba: No real concerns.  Mainly just want keep code for one time scripts separate from services code so it's easier to clean up when deleting the one time code.

Issue #11553, opened 2022-01-07
[email templates] Small difference between plaintext and HTML versions of new subscriptionAccountFinishSetup


data-sync-user:
**Build:**
Stage (Train 223.1)

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest version

**Prerequisites:**
Have VPN set in a country where Mozilla VPN is available.
The user doesn't have a Firefox account.

**Steps to reproduce:**

1. Navigate to https://www-dev.allizom.org/en-US/products/vpn.
1. Click Get Mozilla VPN and then select a plan.
1. Fill in a valid email address that doesn't have associated a FF account.
1. Complete the subscribing process.
1. Go to inbox and check the "Welcome to Mozilla VPN: Please set your password" email.
1. Compare the HTML with the plaintext version.

**Actual result:**
"Payment details" text is not displayed in the plaintext version compared to HTML version.

**Expected result:**
"Payment details" text is displayed in the plaintext version to be consistent with HTML version.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4404)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/19077/HTML+version.PNG">HTML version.PNG</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/19076/Payment+details+text+missing+from+plaintext.PNG">Payment details text missing from plaintext.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 224.1] under Windows 10 64-bit. “Payment details“ text is added to plaintext version of the template.

!new situation.PNG|width=647,height=419!

PR #11594, opened 2022-01-11
fix(auth-server): small difference between plaintext and HTML

xlisachan:
Closes: #11553

`Payment details:` has been added to the plaintext version in the partial. The formatting of the plaintext has also been updated within this pull request.

## Checklist

## Screenshots (Optional)
Previous:
Screen Shot 2022-01-11 at 6 14 55 PM

Fixed within this PR:
Screen Shot 2022-01-11 at 3 40 44 PM


LZoog (approved):
LGTM! We could maybe use a test for "Payment details:" but I know we don't have every line explicitly tested anyway.

xlisachan:
When we update the tests, we could add a test for the partials that would incorporate this check for "Payments details:". 🙂

PR #11595, opened 2022-01-11
feat(subscriptions): save promo code to subscription metadata

chenba:
Because:
 - we need a sure way of finding the applied promotion code for a
   subscription
 - subscription.discount is null when a one time coupon is used

This commit:
 - save the applied promotion code to the subscription metadata

Closes: #11378

### DIFF
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -67,6 +67,8 @@ export const STRIPE_PRODUCTS_CACHE_KEY = 'listStripeProducts';
 export const STRIPE_PLANS_CACHE_KEY = 'listStripePlans';
 export const STRIPE_TAX_RATES_CACHE_KEY = 'listStripeTaxRates';
 
+export const SUBSCRIPTION_PROMOTION_CODE_METADATA_KEY = 'appliedPromotionCode';
+
 enum STRIPE_CUSTOMER_METADATA {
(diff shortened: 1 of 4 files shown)


packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -522,15 +520,45 @@ export class StripeHelper {
...
+    // subscription has been successfully created.
+    if (
+      promotionCode &&
+      (subscription.latest_invoice as Stripe.Invoice).discount
+    ) {
+      const subscriptionMetadata = {
+        ...subscription.metadata,
+        [SUBSCRIPTION_PROMOTION_CODE_METADATA_KEY]: promotionCode.code,
+      };
+      subscription.metadata = subscriptionMetadata;
```
biancadanforth, line 549: Nit: Can we do the assignment at the same time?

biancadanforth (approved):
LGTM. Works for me locally for one-time coupons (where the min charge is >= $0.50.

Very confusing about how `discount` shows up depending on the type of coupon and type of Stripe object...

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: I'm certainly not against it.  But now that CI is passing I don't want to run it again for just this change.  The assigned var is used below in the update call as well fwiw.

Issue #11468, opened 2021-12-29
[subplat] Hosted Storybook for payments is broken


data-sync-user:
The hosted Storybook for the payments server at http://mozilla.github.io/fxa/fxa-payments-server/?path=/story/ is mostly broken with the error `setting getter-only property "createToken"`.

This looks to be related to `defaultStripeStubs` in the `MockApp` component used for most of our stories.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4369)

data-sync-user:
➤ Reino Muhl commented:

I had to do a bit of digging, but it looks like we don’t use Github Pages anymore to host our Storybooks.

The latest builds of the various Storybooks are published to Google Cloud Platform, using this GCP Storybook Publisher ( https://github.com/mozilla-fxa/storybook-gcp-publisher ), and can be found in the following ways.

  1. Via direct URL
    1. Navigate to the Mozilla Storybooks FXA ( https://storage.googleapis.com/mozilla-storybooks-fxa/index.html ) URL
    2. Click the commit you’d like to view. (This includes both main and PR commits)
    3. Click fxa-payments-server ( https://storage.googleapis.com/mozilla-storybooks-fxa/commits/7018a5c488b3ac97c4fe52891b7e027dcf6ea4ca/fxa-payments-server/index.html )

  2. From Github ( https://github.com/mozilla/fxa ) (Courtesy of Laurens Slack message) 
    1. Click on the green check mark next to the latest commit off of main. (Near the top of the page)
    2. Scroll down and click on “Details” for entry storybooks:pull request
    3. Click fxa-payments-server ( https://storage.googleapis.com/mozilla-storybooks-fxa/commits/7018a5c488b3ac97c4fe52891b7e027dcf6ea4ca/fxa-payments-server/index.html )

data-sync-user:
➤ Bianca Danforth commented:

Thanks Reino Muhl for looking into this. Perhaps we switched to GCP at some point? The GitHub Pages URL I got is from the [payments server’s README|#storybook-in-circleci], so perhaps that’s all that needs updating?

PR #11609, opened 2022-01-12
chore(docs): update storybook publish links

StaberindeZA:
## Because

- Latest versions of Storybook builds are no longer published to Github
  pages.

## This pull request

- Update docs to indicate new location where Storybooks are published.

Closes: #11468

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-payments-server/README.md
+++ b/packages/fxa-payments-server/README.md
@@ -6,9 +6,20 @@ This is the server that handles payments.
 
 This project uses [Storybook](https://storybook.js.org/) to show each screen without requiring a full stack.
 
-You can view the Storybook built from the most recent main at http://mozilla.github.io/fxa/fxa-payments-server/
-
 In local development, `yarn storybook` should start a Storybook server at <http://localhost:6006> with hot module replacement to reflect live changes.
+
+### Published links
+
+The latest builds of the various FXA Storybooks are published to Google Cloud Platform, using this GCP Storybook Publisher, and can be found in the following ways.
+
+- Via direct URL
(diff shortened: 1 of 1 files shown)


packages/fxa-payments-server/README.md
```diff
@@ -6,10 +6,21 @@ This is the server that handles payments.
 
 This project uses [Storybook](https://storybook.js.org/) to show each screen without requiring a full stack.
 
-You can view the Storybook built from the most recent main at http://mozilla.github.io/fxa/fxa-payments-server/
-
```
biancadanforth, line 10: Is there a 'latest' `main` link we can give folks who aren't concerned about a particular revision? The previous link did this. It saved people a few clicks, and took them directly to the payments server Storybook.

packages/fxa-payments-server/README.md
```diff
@@ -6,10 +6,21 @@ This is the server that handles payments.
...
 In local development, `yarn storybook` should start a Storybook server at <http://localhost:6006> with hot module replacement to reflect live changes.
 
+### Published links
```
biancadanforth, line 11: Nit: Maybe we could call this section "Latest builds"?

biancadanforth (approved):
R+WC (with changes). Thanks Reino. This looks good.

Ideally there'd be a "latest" payments URL we could point folks to that takes them directly there, but maybe that doesn't exist for GCP hosted Storybooks?

Also, I suggest changing the commit subject from "fix(storybook)" to "chore(docs)", since this patch is not a storybook bug, but a doc update.

packages/fxa-payments-server/README.md
StaberindeZA: Unfortunately none that I could find. I agree that would be ideal.

Issue #7304, opened 2021-01-20
FxA server: decide if we can submit performance metrics for opted out users


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-2924)
┆Issue Number: FXA-2924

PR #11623, opened 2022-01-13
feat(payments): Disable metrics collection for opted out users

### LINKED ISSUE #10989: Do not submit metrics for opted out users in payments-server
We've done this in the content-server and fxa-settings. We need to do it in payments as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4193)

LZoog:
Because:
* Users will be able to opt out of metrics collection in Settings and if a logged in user has opted out, we don't want any metrics collection across our apps

This commit:
* Checks if a user has disabled metrics collection before redirecting them to payments, and if they have disabled it, don't include any flow params, which must be present for flow event logging to occur
* Updates the storage cache when a mutation request is made in Settings so the content server always reads the latest data
* Enables navigation timing in payments and settings only if a user has not opted out
* Adds license to touched payment server files

closes #10989 

---

Opening this as a draft while I finish tests.

I personally preferred removing the flow params altogether rather than subscribe to the redux store and not submitting the events that way. We already have tests in place for not submitting if missing flow params are present and it seemed nicer for the user to not see unneeded params if they've opted out of data collection.

To easily toggle `metricsEnabled` for any local manual testing, change `showDataCollection = false` to `true` in `PageSettings/index.tsx`.


LZoog:
Things _seem_ to be passing for me locally so I'm going to mark this as ready for review.

Also bumping up the points here to a 5; while the code footprint isn't super large, this ticket took me longer than a 2-3 should and I had to dive a bit into settings, content-server, and payments and handle different cases for each of them. We had a spike previously deciding opted out users shouldn't have performance metrics submitted, but we didn't create a separate ticket for it, so I checked for that and added tests here.

packages/fxa-payments-server/src/components/Header/index.stories.tsx
```diff
@@ -12,6 +17,7 @@ const userProfile = {
   locale: 'en-US',
   twoFactorAuthentication: false,
   uid: 'UIDSTRINGHERE',
+  metricsEnabled: true,
```
LZoog, line 20: All of these stories were failing in CI without this added. I can't help but think there should be a single Profile mock that's pulled in everywhere and modified as needed, I'm going to file an issue for that and a couple of others I came across while spelunking around in payments

packages/fxa-payments-server/src/App.tsx
```diff
@@ -48,6 +53,20 @@ export type AppProps = {
...
+    // explicitly check for `false` because `undefined` means
+    // the user doesn't have an account or isn't signed in
+    if (profile?.result && profile.result.metricsEnabled !== false) {
+      observeNavigationTiming('/navigation-timing');
+      unsubscribe?.();
+    }
+  };
+  unsubscribe = store.subscribe(metricsObs);
+}
+
```
chenba, line 69: Note the navigation timing metrics do not include any user specific data.  This still could be the right thing to do for an opt-out, but I want to make that clear.  😄

packages/fxa-settings/src/lib/firefox.ts
```diff
@@ -20,6 +20,15 @@ export interface FirefoxMessage {
   };
 }
 
+interface ProfileUid {
+  uid: hexstring;
+}
+
+interface ProfileMetricsEnabled {
+  metricsEnabled: boolean;
+}
```
chenba, line 29: hmm interesting that you use `interface`s here and then `type` for the sum type below.  I'd like learn more about the reason for that.

packages/fxa-settings/src/models/Account.ts
```diff
@@ -764,6 +764,10 @@ export class Account implements AccountData {
         variables: { input: { state } },
       })
     );
+    const legacyLocalStorageAccount = currentAccount()!;
+    legacyLocalStorageAccount.metricsEnabled = state === 'in';
+    currentAccount(legacyLocalStorageAccount);
+    firefox.profileChanged({ metricsEnabled: this.metricsEnabled });
```
chenba, line 770: heh, never mind my above comment about the Profile type being passed for `profileChanged`. 😛

chenba (approved):
👍

packages/fxa-payments-server/src/App.tsx
LZoog: 👍 Yeah, it just seemed like the right thing to do considering the decision made in #7304. I think we might've needed a separate ticket out of that but we didn't make one for it so I just did it here. 😬

packages/fxa-settings/src/lib/firefox.ts
chenba: "the doc" as in the official TS doc?

PR #11624, opened 2022-01-13
feat(auth): use latest stripe object with proper invoice close

### LINKED ISSUE #10934: [subplat] Fetch latest objects from Stripe on webhook events
In FXA-4171, we learned that:

* Webhooks can be delivered multiple times and out of order, and
* A Stripe API call can trigger multiple webhooks for multiple Stripe objects, and those webhook events will have the same `event.request.id`.
  * E.g. As we saw in FXA-4171, creating a subscription triggers a `customer.subscription.created` webhook as well as a `customer.updated` webhook, where the latter updates the `currency` property on the `customer`.
* Consequently, we should:
  * Process all webhooks regardless of `event.request.id`
  * Never insert objects from webhooks into Firestore or act on them as if that object reflects the latest state in Stripe.

The fix for FXA-4171 ignores `event.request.id` and has us fetching the latest objects from Stripe on some webhook events for those Stripe objects we currently cache in Firestore (`customer`, `subscription` and `invoice` objects), but we should do the same for other webhook events.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4182)

bbangert:
Because:

* Our code was using the invoice object rather than the latest version
  in some parts.
* We didn't check if the invoice was in draft state before finalizing.

This commit:

* Ensures the latest version of the object is always fetched before
  running the webhook handler.
* Only finalizes invoices if they're in draft state.

Closes #10934

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
+++ b/packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
@@ -21,6 +21,7 @@ import { PayPalHelper } from '../../payments/paypal';
 import {
   INVOICES_RESOURCE,
   StripeHelper,
+  STRIPE_OBJECT_TYPE_TO_RESOURCE,
   SUBSCRIPTION_UPDATE_TYPES,
 } from '../../payments/stripe';
 import { AuthLogger, AuthRequest } from '../../types';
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
```diff
@@ -357,7 +389,10 @@ export class StripeWebhookHandler extends StripeHandler {
       return;
     }
     const invoice = event.data.object as Stripe.Invoice;
-    if (!(await this.stripeHelper.invoicePayableWithPaypal(invoice))) {
+    if (
+      !(await this.stripeHelper.invoicePayableWithPaypal(invoice)) ||
+      invoice.status !== 'draft'
```
chenba, line 394: Maybe swap the conditions so that it can short-circuit on the easier one.

packages/fxa-auth-server/test/local/routes/subscriptions/stripe-webhooks.js
```diff
@@ -212,6 +213,9 @@ describe('StripeWebhookHandler', () => {
           );
           await StripeWebhookHandlerInstance.handleWebhookEvent(request);
           assertNamedHandlerCalled(expectedHandlerName);
+          assert.isTrue(
+            StripeWebhookHandlerInstance.stripeHelper.expandResource.calledOnce
+          );
```
chenba, line 218: Could check the arguments in addition to the number of calls.

chenba (approved):
LGTM, but the else for the unsupported resource types could use a test.

packages/fxa-auth-server/lib/routes/subscriptions/stripe-webhook.ts
bbangert: Yea, I considered that, but we don't accidentally finalize this very often, so it didn't seem very impactful.

PR #11625, opened 2022-01-13
chore(deps): update docker images to node 16

dannycoates:
local dev and CI have been stable on 16 for a while. Let's update prod.


chenba (approved):
let's try it

PR #11627, opened 2022-01-13
fix(emails): use static l10n-ids in email templates

### LINKED ISSUE #11472: Don't use dynamic/concatted l10n IDs (e.g. templateName-action) in .mjml files
We currently have several instances in our templates where we reference l10n IDs based on the `template` (the name of the template), and then append something like `-action` to create `lowRecoveryCodes-action` for example. This is fragile because when the string is updated in Fluent files, we have to update the l10n ID as well and a common practice is to just use a `2` at the end.

We do this in at least the `button` partial and the `cadReminder` partial, but we should double check we're not doing this in other places as well. We should remove all references to this in the templates and pass in the entire ID instead.

We essentially need what was started in [this PR/comment](https://github.com/mozilla/fxa/pull/11445/files#r776417851). This will touch a lot of files.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4372)

dannycoates:
fixes #11472


packages/fxa-auth-server/lib/senders/emails/templates/cadReminderFirst/index.mjml
```diff
@@ -2,4 +2,26 @@
   # License, v. 2.0. If a copy of the MPL was not distributed with this
   # file, You can obtain one at http://mozilla.org/MPL/2.0/. %>
 
-<%- include('/partials/cadReminder/index.mjml') %>
+<mj-section>
```
dannycoates, line 5: I decided to inline the cadReminder partial. The extra indirection and variables for one additional email doesn't seem worth it.

packages/fxa-auth-server/lib/senders/emails/templates/postAddTwoStepAuthentication/index.mjml
```diff
@@ -24,6 +24,9 @@
   </mj-column>
 </mj-section>
 
-<%- include('/partials/button/index.mjml') %>
+<%- include('/partials/button/index.mjml', {
+  buttonL10nId: "postAddAccountRecovery-action",
```
xlisachan, line 28: ^ postAddTwoStepAuthentication

xlisachan (approved):
r+, just needs a minor revision for the `postAddTwoStepAuthentication` plaintext file. Thanks for updating this! 😀

PR #11628, opened 2022-01-13
FXA-4224 feat(auth): add one-off script to update customer locations in Stripe

chenba:
Because:
* We need to report user state/country for tax reporting purposes.

This commit:
* Adds a one-off script to obtain the state for both Stripe and PayPal
  users and update the Stripe customer address when possible.

Co-authored-by: Bianca Danforth <bdanforth@mozilla.com>

Closes: #11070 

We'll need to work with jbuck on running this script on stage and prod.


packages/fxa-auth-server/lib/payments/customer-locations.ts
```diff
@@ -1,77 +1,75 @@
...
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { BigQuery } from '@google-cloud/bigquery';
+import { PayPalBillingAgreements } from 'fxa-shared/db/models/auth';
+import { uuidTransformer } from 'fxa-shared/db/transformers';
 import { Logger } from 'mozlog';
-
-// Region codes from: https://en.wikipedia.org/wiki/ISO_3166-2:US and https://en.wikipedia.org/wiki/ISO_3166-2:CA
+import Stripe from 'stripe';
+import { StripeHelper, COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP } from './stripe';
```
bbangert, line 9: Always nice to TS hero sort these.

packages/fxa-auth-server/lib/payments/customer-locations.ts
```diff
@@ -1,77 +1,75 @@
...
-} as { [key: string]: string };
+// GCP credentials are stored in the GOOGLE_APPLICATION_CREDENTIALS env var
+// See https://cloud.google.com/docs/authentication/getting-started
+const GCP_PROJECT_NAME =
+  process.env.FXA_TAX_REPORTING_GCP_PROJECT_NAME || 'bdanforth-fxa-dev';
+const GCP_DATASET_ID =
+  process.env.FXA_TAX_REPORTING_GCP_DATASET_ID || 'fxa_auth_stage';
+const GCP_TABLE_NAME =
+  process.env.FXA_TAX_REPORTING_GCP_TABLE_NAME ||
+  'fxa_4218_uid_state_country_count_01042022';
```
bbangert, line 32: This is really odd to have in application code. Configuration values should be looked up from the config object.

packages/fxa-auth-server/lib/payments/customer-locations.ts
```diff
@@ -1,77 +1,75 @@
...
+const DATASET_LOCATION = 'US';
 
 export class CustomerLocations {
+  private bigquery: any;
```
bbangert, line 36: I thought Google's SDK's were typically done in TypeScript?

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -861,7 +870,7 @@ export class StripeHelper {
     customerId: string;
     postalCode: string;
     country: string;
-  }): Promise<boolean> {
+  }): Promise<void> {
```
bbangert, line 873: It's not useful to know whether the location was actually set anymore?

packages/fxa-auth-server/scripts/set-stripe-customer-locations.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { log, stripeHelper } = await setupProcesingTaskObjects(
+    'set-stripe-customer-location'
+  );
+  const isDryRun = parseDryRun(program.dryRun);
+  const limit = parseMaxCustomers(program.maxCustomers);
+  const delay = parseInt(program.delay);
+
+  const statsd = Container.get(StatsD);
+  statsd.increment('set-stripe-customer-locations.startup');
+  const customerLocations = new CustomerLocations({ log, stripeHelper });
```
bbangert, line 52: Rather than have customer locations setup globals and yank them out of process.env, this script should take the project data as CLI arguments, then pass them through here during setup.

packages/fxa-auth-server/scripts/set-stripe-customer-locations.ts
```diff
@@ -0,0 +1,66 @@
...
+  await customerLocations.backfillCustomerLocation({ limit, isDryRun, delay });
+  statsd.increment('set-stripe-customer-locations.shutdown');
+  await promisify(statsd.close).bind(statsd)();
```
bbangert, line 55: As a one-off, this doesn't matter. 😉

bbangert:
For a one-off script, this mixes in a lot of deeper library code that has different code constraints on it than a one-off script. It's definitely cleaner to break down the code for the script into a module, but there's precedent in the `scripts/` directory of other scripts that needed modules that were very script-oriented so they're close by rather than in general library module code where it would be assumed they're used by the running auth-server.

packages/fxa-auth-server/lib/payments/customer-locations.ts
chenba: > This is really odd to have in application code.

Heh can't disagree.

I did confirm with Bianca that she's okay with leaving these values here.  I suppose this type of defaults is not that different from using default values in the convict configs.  We'll remove the defaults.

> Configuration values should be looked up from the config object.

Since this is a one time script we do not plan to add more config values.

packages/fxa-auth-server/lib/payments/stripe.ts
chenba: We changed the calling code around a bit and ended up not using the return value.  However, we should re-throw the exception so the calling code will log it.

packages/fxa-auth-server/lib/payments/customer-locations.ts
```diff
@@ -100,79 +98,229 @@ export class CustomerLocations {
...
+    })) {
+      if (count >= limit) {
+        break;
+      }
+      count++;
+
+      try {
+        const paymentMethod = customer.invoice_settings
+          ?.default_payment_method as Stripe.PaymentMethod;
+        const paymentMethodCountry = paymentMethod?.card?.country;
```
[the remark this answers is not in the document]
chenba: I think this could be an effect of the script started off as quick one-off script.  But yeah, that's a safe change we could do.  Thanks

chenba:
> For a one-off script, this mixes in a lot of deeper library code that has different code constraints on it than a one-off script. It's definitely cleaner to break down the code for the script into a module, but there's precedent in the `scripts/` directory of other scripts that needed modules that were very script-oriented so they're close by rather than in general library module code where it would be assumed they're used by the running auth-server.

Sorry @bbangert I'm having a hard time parsing this.  😦  `customer-locations` is only imported by the script.  If it's the file location that's the issue, that's easy enough to address.

packages/fxa-auth-server/lib/payments/customer-locations.ts
biancadanforth: These are now required options passed into the script command with no defaults. Unfortunately I couldn't get `commander` to enforce the "required" part, so I am throwing errors instead.

TIL: We use `commander` v 2.18 in the auth server, and it's at 9.0.0 at the moment. >.>

packages/fxa-auth-server/lib/payments/stripe.ts
biancadanforth: We added the return value back in, as it turns out we need it for the CustomerLocations call site.

bbangert:
> Sorry @bbangert I'm having a hard time parsing this. 😦 `customer-locations` is only imported by the script. If it's the file location that's the issue, that's easy enough to address.

That's correct, if the only consumer of this library is a one-off script, it should be in a module directory in the scripts dir, like other modules used only by scripts.

packages/fxa-auth-server/lib/payments/customer-locations.ts
```diff
@@ -100,79 +104,234 @@ export class CustomerLocations {
...
+    const pause = async () =>
+      new Promise((resolve) => {
+        setTimeout(resolve, delay);
+      });
+    const lookupCountries = Object.values(
+      COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP
+    );
+
+    for await (const customer of this.stripeHelper.stripe.customers.list({
+      expand: ['data.invoice_settings.default_payment_method'],
```
bbangert, line 226: It'd be quite a bit more efficient to fetch these in batches of 100 or something ab it higher than the default of 10.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -132,6 +132,15 @@ export type PaymentBillingDetails = ReturnType<
...
 
+// The countries we need region data for
+export const COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP = {
+  // The long name is used in the BigQuery metrics logs; the short name is used
+  // in the Stripe customer billing address.  The long names are also used to
+  // index into the country to states maps.
+  'United States': 'US',
+  Canada: 'CA',
+} as { [key: string]: string };
+
```
bbangert, line 143: Is this used at all in `stripe.ts`?

chenba: It's used in the route handlers (`routes/subscriptions/{paypal,stripe}.ts`

chenba:
New commit to move a couple files and set the limit to 100 on the `customer.list` call: https://github.com/mozilla/fxa/pull/11628/commits/5aceac363abb7c10d8eb13e6f706e0613d58f348

bbangert (approved):

Issue #11629, opened 2022-01-13
[auth-server] add mjml test for `subscriptionFirstInvoiceDiscount`


data-sync-user:
A new email has been created - `subscriptionFirstInvoiceDiscount` - add its test within `mjml-emails.ts`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4441)

Issue #11471, opened 2021-12-29
Move gettext actions & subjects from `email.js`, improve related `fluent-localizer.ts`


data-sync-user:
This is blocked by the email revamp cleanup ticket so we have more freedom modifying `email.js` values that get passed into the templates.

We should improve how we dynamically render actions, subjects, and possibly titles into the email templates in at least `email.js` and `fluent-localizer.ts`. At the moment, we’ll [still be using gettext for some of these](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L1494-L1495) in the auth-server and we should be able to remove these and move them into the template files. We may need to pass in the l10n ID into the template to render the subject, but hopefully there’s a more elegant solution. We also need to fix a [related bug in fluent-localizer](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts) (search that file for `TODO: #11471`).

At the time of writing we haven’t explored [this section](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/senders/email.js#L349) of `email.js`. It’s possible we need a follow up for addressing this section if that’s not straightforward, and we will likely need an additional follow up to [retire gettext in the auth server by moving certain remaining localizations](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/session.js#L509-510) to the client if that’s possible, or to use Fluent there, or make some documented decision to keep it around. When we do retire it, we should remove `fxa-content-server-l10n` from the `fxa-auth-server`'s build `dist`, since `public/locales` contains the Fluent file.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4371)

data-sync-user:
➤ Lauren Zugai commented:

I’m kicking this out of the Revamp Follow Up epic because this turned much more into a maintenance task to improve our renderer/l10n set up than it was new emails project related.

This turned into a large-ish refactor of our fluent-localizer class which is being split into a renderer class and a localizer class. This allows us to remove the old renderer used for emails and sms, and it’s refactored in such a way that we can use the new localizer class across the auth server to retire gettext, and to allow email subjects and actions which were previously set in emails.js into their own template directories. There will be a follow up issue to finish moving all of the gettext strings into the new infrastructure but the bulk of the work will be done here.

PR #11630, opened 2022-01-13
fix(auth-server): add mjml test for `subscriptionFirstInvoiceDiscount`

xlisachan:
Closes: #11629 (FXA-4441)

## Checklist


StaberindeZA:
Awesome, thank you for adding in the missing pieces for this new template!

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -1543,6 +1744,597 @@ describe('lib/senders/mjml-emails:', () => {
...
+  });
+
+  it('set properties on self from config correctly', () => {
+    Object.entries(mailerConfig).forEach(([key, expected]) => {
+      assert.equal(mailer[key], expected, `${key} was correct`);
+    });
+  });
+});
+
+describe('email translations', () => {
```
LZoog, line 2290: I think we should remove these for now and anything else referencing "localization" added here - we purposefully have kept those out since it's handled differently with the new emails. We'll have Fluent translations tested when we get #11529 in, and the `_getLocalizedCurrencyString` should be handled when #11471 is in.

packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -1543,6 +1744,597 @@ describe('lib/senders/mjml-emails:', () => {
...
+        assert.equal(args[0].to, 'foo@restmail.net');
+        assert.equal(args[0].subject, 'subject');
+        assert.equal(args[0].headers['X-Template-Name'], 'verifyLogin');
+        assert.equal(args[0].headers['X-Uid'], 'foo');
+        assert.equal(args[0].provider, undefined);
+        assert.equal(typeof mailer.mailer.sendMail.args[0][1], 'function');
+      });
+    });
+
+    it('sends request to fxa-email-service when selectEmailServices tells it to', () => {
```
LZoog, line 2048: Hmm I didn't dig into this PR at all but we removed `fxa-email-service` in #10558. 🤔

LZoog (approved):
Thanks for adding all of these! 😁

Issue #11642, opened 2022-01-17
[email templates] Wrong email title in plaintext version of verifyLogin template


data-sync-user:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release

**Prerequisites:**
N/A

**Steps to reproduce:**

1. Don't have actual steps for triggering this email. It was triggered automatically when I didn't even expect it.

**Actual result:**
Email title in plaintext version is wrong. It should be "New sign-in to Firefox" instead of "Activate the Firefox family of products".

**Expected result:**
Email title in plaintext version is correct "New sign-in to Firefox".

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4444)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/19410/wrong+email+title+plaintext.PNG">wrong email title plaintext.PNG</a>

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on Stage [Train 224.1] under Windows 10 64-bit. Email title in plaintext version is correct "New sign-in to Firefox". See new situation:

!new situation.PNG|width=789,height=362!

PR #11645, opened 2022-01-18
fix(auth-server): wrong email title in plaintext version of verifyLogin template

xlisachan:
Closes: #11642

The email title in the plaintext version should be "New sign-in to Firefox" instead of "Activate the Firefox family of products".

Note: While this error was reflected in emails using `yarn write-emails`, this error was not reflected in Storybook.

## Checklist

## Screenshots (Optional)
Old 
wrong email title plaintext

New
Screen Shot 2022-01-18 at 9 21 10 AM


LZoog (approved):
Oops🤪

Btw I think this should be a 1-pointer? Thanks for the fix 😁

Issue #11648, opened 2022-01-18
Don't show iOS banner when in Pocket sign-in screen


vbudhram:
### Steps to reproduce

On iOS, have Firefox iOS installed

1. Download and install Pocket iOS app
2. Click Sign-in button

### Expected result

Shown FxA sign-in page without iOS banner.

### Actual result

Shown FxA sign-in page with iOS banner. Clicking the banner will open Firefox iOS and not do anything

![IMG_6545](https://user-images.githubusercontent.com/1295288/149972968-19e2401b-f596-480e-a754-59ae236bb8db.PNG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4447)

PR #11660, opened 2022-01-19
fix(ios): Don't show Firefox iOS banner when loading signup/signin views

vbudhram:
## Because

- We shouldnt show this banner because if FxA is opened in a webview in another app, the user would be take to Firefox iOS to continue signup/signin which is confusing

## This pull request

- Removes the Firefox iOS banner

Closes: https://github.com/mozilla/fxa/issues/11648

## Checklist

- [ ] I have verified that my changes render correctly in RTL (if appropriate)

### DIFF
--- a/packages/fxa-content-server/server/templates/pages/src/index.html
+++ b/packages/fxa-content-server/server/templates/pages/src/index.html
@@ -14,12 +14,6 @@
             content="width=device-width,initial-scale=1,minimum-scale=1,maximum-scale=2,user-scalable=yes"
         />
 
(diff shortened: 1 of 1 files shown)


dannycoates (approved):
differentiating between webview / real browser doesn't seem worth it so 👍

LZoog (approved):
Are we removing this for _all_ cases in case the user is in the webview? I kind of wonder if there's some sort of check/conditional we can do instead, but I know we don't have many non-FF users anyway and we want to upgrade the content-server to React anyway.

Issue #11678, opened 2022-01-20
[email templates] Revise `images` partial


data-sync-user:
The `images` partial includes the App Store and Google Play badges, and it passes the `productName` variable for image alt text. As there are both FxA and SubPlat emails that either should have `Firefox` set to `productName` or do not have a `productName` variable (reminder emails to finish account setup, payment emails, etc), a conditional needs to be added so the templates do not break. While this affects all templates, the following emails are currently not loading within Storybook:

* subscriptionAccountReminderFirst
* subscriptionAccountReminderSecond
* subscriptionsPaymentExpired
* subscriptionsPaymentProviderCancelled

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4462)

data-sync-user:
➤ Bianca Oltean commented:

This issue was verified as fixed. The mentioned templates are loading in Storybook.

!templates in Storybook.gif|width=1905,height=1012!

PR #11679, opened 2022-01-20
fix(auth-server): revise `images` partial

xlisachan:
## Because

- The images partial includes the App Store and Google Play badges, and it passes the `productName` variable for image alt text. As there are both FxA and SubPlat emails that either should have Firefox set to `productName` or do not have a `productName` variable (reminder emails to finish account setup, payment emails, etc), a conditional needs to be added so the templates do not break. While this affects all templates, the following emails are currently not loading within Storybook:
- `subscriptionAccountReminderFirst`
- `subscriptionAccountReminderSecond`
- `subscriptionsPaymentExpired`
- `subscriptionsPaymentProviderCancelled`

## This pull request

- Fixes the issue by applying the conditional within the `images` partial.

Closes: #11678

## Checklist

### DIFF
--- a/packages/fxa-auth-server/test/local/senders/mjml-emails.ts
+++ b/packages/fxa-auth-server/test/local/senders/mjml-emails.ts
@@ -196,11 +196,8 @@ const TESTS: [string, any, Record<string, any>?][] = [
       { test: 'include', expected: decodeUrl(configHref('supportUrl', 'cad-reminder-first', 'support')) },
       { test: 'include', expected: 'alt="Firefox logo"' },
       { test: 'include', expected: 'alt="Devices"' },
-      // FIXME: these tests fail as all email tests utilize the same testing parameters
-      //    (e.g, while productName does not exist for this email template, it does within the test
-      //    returns Download Firefox Fortress on the App Store as it believes it has a productName)
-      // { test: 'include', expected: `alt="Download Firefox on the App Store"` },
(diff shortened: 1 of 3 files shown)


packages/fxa-auth-server/test/local/senders/mjml-emails.ts
```diff
@@ -196,8 +196,11 @@ const TESTS: [string, any, Record<string, any>?][] = [
       { test: 'include', expected: decodeUrl(configHref('supportUrl', 'cad-reminder-first', 'support')) },
       { test: 'include', expected: 'alt="Firefox logo"' },
       { test: 'include', expected: 'alt="Devices"' },
-      { test: 'include', expected: `alt="Download Firefox on the App Store"` },
-      { test: 'include', expected: `alt="Download Firefox on Google Play"` },
+      // FIXME: these tests fail as all email tests utilize the same testing parameters
```
LZoog, line 199: I think you can fix this by using `updateTemplateValues`

xlisachan: Updated, thanks!

LZoog (approved):
Noice, thanks for jumping on this immediately😎

We don't always use the `needs:uplift` label for PRs needed in a dot releases but since there's at least one other issue QA just filed I slapped it on there.

Issue #11680, opened 2022-01-20
[email templates] subscriptionFirstInvoiceDiscount and subscriptionFirstInvoice emails are not received


data-sync-user:
**Build:**

Stage

**Affected platforms:**

Windows 10 64-bit

**Browsers tested on:**

Firefox, latest release

**Prerequisites:**

Have the VPN set in a country where Mozilla VPN is available.

**Steps to reproduce:**

1. Navigate to [https://www-dev.allizom.org/en-US/products/vpn](https://www-dev.allizom.org/en-US/products/vpn|smart-link)  and click “Get Mozilla VPN“.
1. Select a monthly plan.
1. Create a new FF account from the subscription page and fill in all the other payment info.
1. Enter a discount coupon and click “Apply”.
1. Click “Subscribe now.“

**Actual result:**

“Mozilla VPN payment confirmed“ email is not received.

**Expected result:**

“Mozilla VPN payment confirmed“ email is received.

**Notes:**

The same issue occurs for subscriptionFirstInvoice.

The same issue occurs if you use the old subscription flow.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4463)

data-sync-user:
➤ Bianca Oltean commented:

This issue is verified as fixed on stage [Train 224.1] under Windows 10 64-bit. “Mozilla VPN payment confirmed“ email is received when you make a subscription with or without a discount using both the old and the new flow.

PR #11682, opened 2022-01-20
fix(subscriptions): use latest stripe object fix

StaberindeZA:
## Because

- Stripe webhook events handler fails when expanding resource.

## This pull request

- Ensures correct arguments are passed to expand resource.

Closes: #11680

## Checklist

## Screenshots (Optional)


bbangert (approved):
Ah, the joy of mocking everything and not using property names everywhere. Nice catch!

Issue #10303, opened 2021-08-31
Delete Stripe Customer when Deleting Stub Account


data-sync-user:
When we delete a stub account, that account potentially has a customer in Stripe.  We should delete that customer from Stripe.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-3927)

data-sync-user:
➤ Bianca Danforth commented:

We’ve had a high number of permafail invoice.payment_failed webhook failures with the error “Unknown account” (+example event+ ( https://dashboard.stripe.com/events/evt_1K9mF8JNcmPzuWtR6o0N5ALl )) for the last several days for the same two users where we have the same email address but multiple Stripe customer ids and uids (+customer with email 1+ ( https://dashboard.stripe.com/customers/cus_KaqjtqjDW5bCLd ) and customer with email 2 ( https://dashboard.stripe.com/customers/cus_KbmEBcyzYPpi67 )). Looking the various uids up in the fxa-admin-panel, we find there is indeed no account, so we are guessing these users may have hit this bug.

What we don’t know is how they got past the existing account check to create a new stub account with a new Stripe customer for the same email address. At least one of the customers ended up subscribing to a plan with a different currency (which requires a page refresh), so it’s not due to us caching the accountExists state in payments.

(Thanks to Barry Chen for helping me investigate here).

data-sync-user:
➤ Ivo Plamenac commented:

we believe this sentry record is related to stripe customers not being deleted: https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/15647824/?query=is%3Aunresolved ( https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/15647824/?query=is%3Aunresolved )

data-sync-user:
➤ Ivo Plamenac commented:

It seems users are still able to subscribe to VPN but since we are not able to reproduce yet we do not know what the impact is from a UI/UX perpsective. Setting severity to S3 for now but this is something we want to fix sooner than later since it is also causing a lot of noise in our monitoring.

data-sync-user:
➤ Ivo Plamenac commented:

Also look to see which stripe webhooks are triggered by deleting a customer and ensure that we are not creating more errors downstream from this.

data-sync-user:
➤ Barry Chen commented:

In order to verified this, you need access to the Stripe dashboard of the environment in which you are testing.

[...]

PR #11683, opened 2022-01-20
fix(subscriptions): delete customer when deleting stub acct

chenba:
Because:
 - we see duplicate customers in Stripe for some email addresses

This commit:
 - delete a customer without any subscriptions when we delete the
   associated "stub" account

Closes: #10303


chenba:
I'll rebase on main along any other changes I need to make from PR review feedback.  Hopefully that will get CI to ✔️

biancadanforth (approved):
LGTM assuming CI passes; I tried it out with the STR from FXA-4274 as suggested and it WFM locally. Really glad to see this get fixed.

Issue #11223, opened 2021-12-02
Fix/improve/remove flaky test-content-server-1 functional tests


data-sync-user:
We keep having the same flaky tests fail which causes a lot of CI retriggers. We need to fix these or maybe there's a better way to get the same coverage.

https://app.circleci.com/pipelines/github/mozilla/fxa/25260/workflows/d1515f62-836b-4942-bffa-275fa0b903f2/jobs/286251

!image-2021-12-02-11-01-28-380.png|thumbnail!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4281)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18315/image-2021-12-02-11-01-28-380.png">image-2021-12-02-11-01-28-380.png</a>

data-sync-user:
➤ Dan Schomburg commented:

Flakiness is correlated related to resources available. I did some experiments using Linux cgroups to constrain the FXA stack resources. At 4GB of memory the stack would generally fail to start. At 8GB tests would generally pass. At 6GB things became unstable again. I wrote up a document describing how to constrain resources with cgroups in case people are not familiar with this and would like to experiment with this in the future. https://docs.google.com/document/d/1wNDgN1Rj3raq7RxDk5MnvIW1d0Qs1it3BSVMznqBd3A/edit?usp=sharing ( https://docs.google.com/document/d/1wNDgN1Rj3raq7RxDk5MnvIW1d0Qs1it3BSVMznqBd3A/edit?usp=sharing|smart-link )

PR #11684, opened 2022-01-20
fix(content-server): Decrease flakiness of content server tests

dschom:
#Because:
- Tests in CI exhibited flakiness

## This pull request
- Addresses edge cases where the document could be caught in an unloaded state when running querySelector
- Increases allowed pageLoadTimeout. When flakiness does occur, it appears to be due to resource constraints in the CI runner. Allowing a little more time for things to settle reduces the chance something fails due to a random slow down.

Closes: #11223

## Checklist

Reproducing flakiness seen in the CI can be tricky to do locally. To accomplish this I used cgroups and did several experiments with different memory limits in place. Ultimately, it seems like flakiness is correlated to available system resources, which has more or less been the suspicion all along. When there are plenty of resources and things are running full throttle, flakiness is not observed.


dannycoates (approved):
nice. let's not do any more work on intern after this

PR #11699, opened 2022-01-21
chore(circleci): Update expiring images

clouserw:
Because:

* The convenience images are expiring on Feb 7

This commit:

* Updates to use the new images


clouserw:
I noticed here https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034 that our mysql images have been deprecated since Jan 1 of last year and are no longer supported starting on Feb 7.  This will update them to the new images

dannycoates (approved):

PR #11704, opened 2022-01-21
FXA-4454 fix(auth): check for unique state on multiple results from Geocoding API

### LINKED ISSUE #11650: Cannot find a postal code for customer cus_KvJ5Br8b6qjbMI
https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/17125290/?referrer=github_plugin

```
Cannot find a postal code for customer cus_KvJ5Br8b6qjbMI
```

Let's remove the customer ID from the message name, so Sentry groups these errors together. Right now they're all treated as separate events.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4449)

### LINKED ISSUE #11661: [subplat] Update getStateFromZip to return unique State, even if Google returns multiple results
**Background**

Currently GoogleMapsService.getStateFromZip will throw an error if a unique result could not be found for the given postal code (zip) and country.

E.g. [Error: Could not find unique results. (10000, United States of America)](https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/17036516/?referrer=slack)

**Requirements**

Update the getStateFromZip method, so that even if the results are not unique, as long as the State is unique, return the unique State value. This change requires the following changes.

* Filter results where the postal_code matches the input Postal Code (zip)
* Iterate over results and see if State is unique

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4454)

biancadanforth:
Because:

* The Geocode API could return multiple address results for a given zipcode.
* Those results may have a unique state, so we shouldn't error until we've checked.

This commit:

* Checks for a unique state when multiple results are returned.

Closes #11661

----

Because:

* Sentry was treating each error as unique, when they should be grouped.

This commit:

* Removes the customer id from some Sentry error messages, so they can be correctly grouped.

Closes #11650

## Checklist

I'm fixing a small one-pointer in this same PR; each ticket has its own commit.

### DIFF
--- a/packages/fxa-auth-server/lib/google-maps-services.ts
+++ b/packages/fxa-auth-server/lib/google-maps-services.ts
@@ -64,8 +64,23 @@ export class GoogleMapsService {
     if (status === Status.ZERO_RESULTS)
       throw new Error(`Could not find any results for address. (${address})`);
 
-    if (results.length > 1)
-      throw new Error(`Could not find unique results. (${address})`);
+    if (results.length > 1) {
+      const stateShortNames = results.map(
+        (result) =>
+          result.address_components.find((addressComponent) =>
+            addressComponent.types.includes(
+              PlaceType2.administrative_area_level_1
+            )
+          )?.short_name
+      );
+      // Check if the state short names match for each result
+      const states = new Set(stateShortNames);
+      if (states.size === 1) {
+        return results[0];
+      } else {
+        throw new Error(`Could not find unique results. (${address})`);
+      }
+    }
 
     return results[0];
   }
--- a/packages/fxa-auth-server/lib/google-maps-services.ts
+++ b/packages/fxa-auth-server/lib/google-maps-services.ts
@@ -58,14 +58,24 @@ export class GoogleMapsService {
   /**
    * Retrieve a unique GeocodeResult for a given address.
    */
-  private async getOneGeocodeResult(address: string): Promise<GeocodeResult> {
+  private async getOneGeocodeResult(
+    address: string,
+    zip: string
+  ): Promise<GeocodeResult> {
     const { results, status } = await this.getGeocodeData(address);
 
     if (status === Status.ZERO_RESULTS)
       throw new Error(`Could not find any results for address. (${address})`);
 
     if (results.length > 1) {
-      const stateShortNames = results.map(
+      const resultsWithMatchingPostalCodes = results.filter((result) =>
+        result.address_components.some(
+          (addressComponent) =>
+            addressComponent.types.includes(PlaceType2.postal_code) &&
+            addressComponent.short_name === zip
+        )
+      );
+      const stateShortNames = resultsWithMatchingPostalCodes.map(
         (result) =>
           result.address_components.find((addressComponent) =>
             addressComponent.types.includes(
@@ -103,7 +113,7 @@ export class GoogleMapsService {
         );
 
       const { address_components: addressComponents } =
-        await this.getOneGeocodeResult(address);
+        await this.getOneGeocodeResult(address, zip);
 
       const state = addressComponents.find((address) =>
         address.types.includes(PlaceType2.administrative_area_level_1)
--- a/packages/fxa-auth-server/test/local/google-maps-services.js
+++ b/packages/fxa-auth-server/test/local/google-maps-services.js
@@ -12,6 +12,10 @@ const { GoogleMapsService } = require('../../lib/google-maps-services');
 const { default: Container } = require('typedi');
 const { AuthLogger, AppConfig } = require('../../lib/types');
 
+function deepCopy(object) {
+  return JSON.parse(JSON.stringify(object));
+}
+
 const geocodeResultMany = {
   data: {
     results: [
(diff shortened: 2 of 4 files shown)


packages/fxa-auth-server/lib/google-maps-services.ts
```diff
@@ -64,8 +64,23 @@ export class GoogleMapsService {
...
-      throw new Error(`Could not find unique results. (${address})`);
+    if (results.length > 1) {
+      const stateShortNames = results.map(
+        (result) =>
+          result.address_components.find((addressComponent) =>
+            addressComponent.types.includes(
+              PlaceType2.administrative_area_level_1
+            )
+          )?.short_name
+      );
```
biancadanforth, line 75: `Array.map` will guarantee that `stateShortNames` has the same length as `results`, so I don't check `if (results.length === stateShortNames.length) {...}`.

Also, it's possible a `result` may not have a state in any of its `addressComponents`, in which case the corresponding `stateShortNames` element will have a value of `undefined`.

Overall, for the 3 zipcodes I know return multiple `results` (10000, 42242 and 55868), none have every result with a state and every state matching, so all still yield error we were throwing before this patch. This means I have yet to find a valid zipcode where this change would make a difference.

packages/fxa-auth-server/test/local/google-maps-services.js
```diff
@@ -132,12 +136,34 @@ describe('GoogleMapsServices', () => {
...
+      );
+    });
+
+    it('Throws error if more than 1 result is returned with mismatching states', async () => {
+      const expectedMessage = 'Could not find unique results. (11111, Germany)';
+      googleClient.geocode = sinon.stub().resolves(geocodeResultMany);
 
       try {
-        await googleMapsServices.getStateFromZip('11111', 'Germany');
+        await googleMapsServices.getStateFromZip('11111', 'DE');
```
biancadanforth, line 166: Some of the edits look weird here, because I moved the previous "multiple `results`" test case up to be adjacent to the new test case.

While I'm here, I can update the arguments to be consistent if desired, but I don't think it matters for the error test cases in question.

biancadanforth (commented):
Added some additional context/notes in line.

For some perspective, I found only 28 unique instances of this error in the prod logs from FXA-4446 (see [comment](https://mozilla-hub.atlassian.net/browse/FXA-4446?focusedCommentId=502829)). Doing the math, this error comprises << 1% of all errors that prevented us from updating an address for an existing Stripe customer, so I don’t think it's worth spending a lot more time on.

packages/fxa-auth-server/lib/google-maps-services.ts
pdehaan: TIL a ZIP code can belong to multiple states! Neat!
https://stackoverflow.com/questions/12450960/find-zip-codes-associated-with-more-than-one-state/12451921
https://gis.stackexchange.com/questions/53918/determining-which-us-zipcodes-map-to-more-than-one-state-or-more-than-one-city

chenba: > TIL a ZIP code can belong to multiple states! Neat!
> 

Yes, neat, in the same way that working with time zones is neat.  😛

biancadanforth:
> (Github added me to the list of "reviewers" because I responded to a comment. Since I paired with Bianca, I'm not sure if I should be reviewing this PR. FWIW LGTM 😉)

Sorry about that Barry; I will add the Co-Authored-By line before merging. I'll wait to see if there's any more feedback first.

chenba:
> Sorry about that Barry; I will add the Co-Authored-By line before merging.

nah, no worries.

packages/fxa-auth-server/lib/google-maps-services.ts
```diff
@@ -64,8 +64,23 @@ export class GoogleMapsService {
     if (status === Status.ZERO_RESULTS)
       throw new Error(`Could not find any results for address. (${address})`);
 
-    if (results.length > 1)
-      throw new Error(`Could not find unique results. (${address})`);
+    if (results.length > 1) {
+      const stateShortNames = results.map(
```
StaberindeZA, line 68: Before `.map` could we also `.filter` on results where `PlaceType2.postal_code` actually equals the input postal code?

For example with postal code 10000 for the United States, none of the results returned actually have a `PlaceType2.postal_code` that equals 10000.

This was actually a bit of an oversight during initial implementation.

biancadanforth: Sure I changed this, though for the specific example of 10000, the result is the same (i.e. we throw an error anyway, since one of the `results` did not have a state in its `addressComponents` list).

packages/fxa-auth-server/lib/google-maps-services.ts
biancadanforth: FWIW I did find an example zip (in Canada) that this patch would fix: J5L 0A3.

packages/fxa-auth-server/lib/google-maps-services.ts
biancadanforth: So weird to get a result with a non-matching postal code...

StaberindeZA (approved):
LGTM, thank you.

PR #11706, opened 2022-01-21
feat(auth-server): Don't include email UTM params for metrics opted out users

### LINKED ISSUE #10874: Don't emit email UTM params for telemetry opted out users
In FXA-2876, we allowed users to opt out of accounts-level telemetry/metrics. If a user has opted out, we shouldn't send UTM params on links in FxA emails.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4163)

LZoog:
Because:
* Users that have opted out of metrics collection shouldn't have UTM params attached to links in emails

This commit:
* Passes 'metricsEnabled' into the mailer via metricsOptOutAt, which then checks for this value before adding UTM params to links

fixes #10874

---

Thankfully I over-complicated this with my first pass (re: my comments in the bug), with some fresh eyes it wasn't so bad 😛

Running `COMMON_METRICS_OPT_OUT_TESTS` for _every_ template seemed unnecessary, but I can easily add more if preferred.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/index.js
+++ b/packages/fxa-auth-server/lib/senders/index.js
@@ -67,6 +67,7 @@ module.exports = async (
             ccEmails: cc,
             email: to || account.email,
             uid: account.uid,
+            metricsEnabled: !account.metricsOptOutAt,
           });
         };
 
@@ -91,6 +92,7 @@ module.exports = async (
       email: emails[0].email,
       primaryEmail: account.email,
       uid: account.uid,
+      metricsEnabled: !account.metricsOptOutAt,
     });
   };
 
@@ -101,6 +103,7 @@ module.exports = async (
       email: emails[0].email,
       primaryEmail: account.email,
       uid: account.uid,
+      metricsEnabled: !account.metricsOptOutAt,
     });
   };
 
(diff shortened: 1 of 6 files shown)


packages/fxa-auth-server/lib/senders/index.js
```diff
@@ -67,6 +67,7 @@ module.exports = async (
             ccEmails: cc,
             email: to || account.email,
             uid: account.uid,
+            metricsEnabled: !account.metricsOptOutAt,
```
vbudhram, line 70: I can't say for certain that we pass an account record to all mailer functions (might be an object with email and uid). I think this is fine to do and then fallback to default of enabling metrics.

vbudhram (approved):
@LZoog :shipit:

packages/fxa-auth-server/lib/senders/index.js
LZoog: Hmm, I wonder if that's why I ran into problems the first time I looked at this. I could've sworn I couldn't see the account data on at least one email, but on a second pass, it seemed to be present everywhere. I agree this seems fine either way, it's not worth the alternative😅

Issue #11573, opened 2022-01-10
[Subplat] Although an error message is displayed, the subscription for Relay is finished using a discount coupon


data-sync-user:
Solution:

For paypal do not allow a $0 total. For stripe honour the minimums per currency found: [https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts](https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts|smart-link) 

**Affected browsers:**

* All;

**Affected FxA versions:**

* 1.223.1;

**Prerequisites:**

* Be in a country where Relay is available;
* Be on the [Set up you subscription](https://payments-stage.fxa.nonprod.cloudops.mozgcp.net/checkout/prod_KEq0LXqs7vysQT?device_id=a06051cfdb6d4ebc864fe4ce32b340bb&flow_begin_time=1639771956279&flow_id=de83d3687dc8f95944a66bebe6016bcf007a61c6bd77ede84ee8b068d2e1b7e2&plan=price_1JmRSRJNcmPzuWtRN9MG5cBy) page;

**Steps to reproduce:**
1. Complete the email fields;
2. Apply a discount coupon that is with fixed amount i.e 50OFFRELAY;
3. Make the payment;

**Expected result:**

* Confirmation page is displayed and both Payment and Welcome emails are received by the user;

**Actual result:**

* Although an error message is displayed, the subscription is being created;
* The Payment confirmation and Welcome emails are received;

**Additional notes:**

* This is a major issue because, although the account and subscription are being created, the user may not know that they have been created;
* See the attached screenshot;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4414)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/19129">relay.png</a>

data-sync-user:
➤ Maria Berlinger commented:

Hi, Reino Muhl  could you please take a look at this issue?
Thank you.

data-sync-user:
➤ Chelsea Lewis commented:

Sounds like we will have a metrics issue filed by Reino Muhl related to this but it will not address the issue described here. 

This is a potential blocker to release - Reino Muhl  will continue investigating.

cc: Brian Yee for awareness

data-sync-user:
➤ Reino Muhl commented:

Quick update, after spending some time yesterday investigating we found the following, with more information provided below.

[...]

PR #11719, opened 2022-01-24
fix(subscriptions): allow zero dollar invoices

StaberindeZA:
## Because

- For invoices where  the amount due is less than Stripe minimums, the
  subscription is successfully created, however the user is shown an
  error message on completion. This applies to both Stripe and PayPal.

## This pull request

- Updates the PayPal zero amount invoice logic.
- Updates the Stripe logic not to expect a payment intent.

Closes: #11573 

## Checklist

## Screenshots (Optional)

Paypal 100% off ($0.00 invoice)
image

Paypal discount puts amount beneath minimum amount ($0.50)
image

Credit Card 100% off ($0.00 invoice)
image

Credit Card discount puts amount beneath minimum amount ($0.50)
image

### DIFF
--- a/packages/fxa-auth-server/lib/payments/paypal-processor.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-processor.ts
@@ -16,7 +16,7 @@ import {
   PAYPAL_BILLING_AGREEMENT_INVALID,
   PAYPAL_SOURCE_ERRORS,
 } from './paypal-error-codes';
-import { StripeHelper } from './stripe';
+import { StripeHelper, STRIPE_MINIMUM_CHARGE_AMOUNTS } from './stripe';
 
 /**
  * Generest a timestamp in seconds that is `hours` before the current
@@ -222,7 +222,8 @@ export class PaypalProcessor {
    */
   private async makePaymentAttempt(invoice: Stripe.Invoice) {
     const customer = invoice.customer as Stripe.Customer;
-    if (invoice.amount_due === 0) {
+    const minAmount = STRIPE_MINIMUM_CHARGE_AMOUNTS[invoice.currency] || 50;
+    if (invoice.amount_due < minAmount) {
       await this.paypalHelper.processZeroInvoice(invoice);
       return true;
     }
--- a/packages/fxa-auth-server/lib/payments/paypal-processor.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-processor.ts
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-import { STRIPE_MINIMUM_CHARGE_AMOUNTS } from 'fxa-shared/subscriptions/stripe';
+import { getMinimumAmount } from 'fxa-shared/subscriptions/stripe';
 import { Logger } from 'mozlog';
 import Stripe from 'stripe';
 import { Container } from 'typedi';
@@ -223,8 +223,8 @@ export class PaypalProcessor {
    */
   private async makePaymentAttempt(invoice: Stripe.Invoice) {
     const customer = invoice.customer as Stripe.Customer;
-    const minAmount = STRIPE_MINIMUM_CHARGE_AMOUNTS[invoice.currency] || 50;
-    if (invoice.amount_due < minAmount) {
+
+    if (invoice.amount_due < getMinimumAmount(invoice.currency)) {
       await this.paypalHelper.processZeroInvoice(invoice);
       return true;
     }
--- a/packages/fxa-auth-server/lib/payments/paypal-processor.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal-processor.ts
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { STRIPE_MINIMUM_CHARGE_AMOUNTS } from 'fxa-shared/subscriptions/stripe';
 import { Logger } from 'mozlog';
 import Stripe from 'stripe';
 import { Container } from 'typedi';
@@ -16,7 +17,7 @@ import {
   PAYPAL_BILLING_AGREEMENT_INVALID,
   PAYPAL_SOURCE_ERRORS,
 } from './paypal-error-codes';
-import { StripeHelper, STRIPE_MINIMUM_CHARGE_AMOUNTS } from './stripe';
+import { StripeHelper } from './stripe';
 
 /**
  * Generest a timestamp in seconds that is `hours` before the current
--- a/packages/fxa-auth-server/lib/payments/paypal.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal.ts
@@ -397,10 +397,10 @@ export class PayPalHelper {
    * @param invoice
    */
   public processZeroInvoice(invoice: Stripe.Invoice) {
-    return Promise.all([
-      this.stripeHelper.finalizeInvoice(invoice),
-      // this.stripeHelper.payInvoiceOutOfBand(invoice),
-    ]);
+    // It appears for subscriptions that do not require payment, the invoice
+    // transitions to paid automatially.
+    // https://stripe.com/docs/billing/invoices/subscription#sub-invoice-lifecycle
+    return this.stripeHelper.finalizeInvoice(invoice);
   }
 
   /**
--- a/packages/fxa-auth-server/lib/payments/paypal.ts
+++ b/packages/fxa-auth-server/lib/payments/paypal.ts
@@ -399,7 +399,7 @@ export class PayPalHelper {
   public processZeroInvoice(invoice: Stripe.Invoice) {
     return Promise.all([
       this.stripeHelper.finalizeInvoice(invoice),
-      this.stripeHelper.payInvoiceOutOfBand(invoice),
+      // this.stripeHelper.payInvoiceOutOfBand(invoice),
     ]);
   }
 
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -725,13 +725,13 @@ export class StripeHelper {
             promotionCode,
           });
 
-        if (discount && total_discount_amounts) {
-          couponDetails.discountAmount = total_discount_amounts[0].amount;
-        }
-
         const minAmount = getMinimumAmount(currency);
         if (total !== 0 && minAmount && total < minAmount) {
           throw error.invalidPromoCode(promotionCode);
+        }
+
+        if (discount && total_discount_amounts) {
+          couponDetails.discountAmount = total_discount_amounts[0].amount;
         }
       } catch (error) {
         if (
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -158,6 +158,34 @@ export const COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP = {
   Canada: 'CA',
 } as { [key: string]: string };
 
+// Stripe minimum charge amounts as set here
+// https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts
+export const STRIPE_MINIMUM_CHARGE_AMOUNTS = {
+  usd: 50,
+  aed: 200,
+  aud: 50,
+  bgn: 100,
+  brl: 50,
+  cad: 50,
+  chf: 50,
+  czk: 1500,
+  dkk: 250,
+  eur: 50,
+  gbp: 30,
+  hkd: 400,
+  huf: 17500,
+  inr: 50,
+  jpy: 50,
+  mxn: 1000,
+  myr: 200,
+  nok: 300,
+  nzd: 50,
+  pln: 200,
+  ron: 200,
+  sek: 300,
+  sgd: 50,
+} as { [key: string]: number };
+
 /**
  * The CacheUpdate decorator has an _optional_ property in its options
  * parameter named `cacheKeysToClear`.  However, if you do not pass in a value
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -158,34 +158,6 @@ export const COUNTRIES_LONG_NAME_TO_SHORT_NAME_MAP = {
   Canada: 'CA',
 } as { [key: string]: string };
 
-// Stripe minimum charge amounts as set here
-// https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts
-export const STRIPE_MINIMUM_CHARGE_AMOUNTS = {
-  usd: 50,
-  aed: 200,
-  aud: 50,
-  bgn: 100,
-  brl: 50,
-  cad: 50,
-  chf: 50,
-  czk: 1500,
-  dkk: 250,
-  eur: 50,
-  gbp: 30,
-  hkd: 400,
-  huf: 17500,
-  inr: 50,
-  jpy: 50,
-  mxn: 1000,
-  myr: 200,
-  nok: 300,
-  nzd: 50,
-  pln: 200,
-  ron: 200,
-  sek: 300,
-  sgd: 50,
-} as { [key: string]: number };
-
 /**
  * The CacheUpdate decorator has an _optional_ property in its options
  * parameter named `cacheKeysToClear`.  However, if you do not pass in a value
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -739,6 +739,14 @@ export class StripeHelper {
           error.errno === AppError.ERRNO.INVALID_PROMOTION_CODE
         ) {
           throw error;
+        } else {
+          Sentry.withScope((scope) => {
+            scope.setContext('retrieveCouponDetails', {
+              priceId,
+              promotionCode,
+            });
+            Sentry.captureException(error);
+          });
         }
       }
 
--- a/packages/fxa-auth-server/lib/payments/stripe.ts
+++ b/packages/fxa-auth-server/lib/payments/stripe.ts
@@ -53,6 +53,8 @@ import { CurrencyHelper } from './currencies';
 import { SubscriptionPurchase } from './google-play/subscription-purchase';
 import { FirestoreStripeError, StripeFirestore } from './stripe-firestore';
 import * as Coupon from 'fxa-shared/dto/auth/payments/coupon';
+import { getMinimumAmount } from 'fxa-shared/subscriptions/stripe';
+import AppError from '../error';
 
 export const CARD_RESOURCE = 'sources';
 export const CHARGES_RESOURCE = 'charges';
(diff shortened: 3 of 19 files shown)


StaberindeZA:
After spending some time investigating this, both Credit Card and Paypal payments had different reasons why $0.00 invoices were failing.

- Paypal payments
  - The processZeroInvoice logic was only called when `amount_due` was 0. However, zero invoices also occur if the `amount_due` is less than the Stripe minimum allowed amounts. (e.g $0.50)
  - processZeroInvoice logic called both `finalizeInvoice` and `payInvoiceOutOfBand`. This ended up throwing an error on attempting to pay the invoice, since it's already in status paid. Seems when an invoice doesn't require a payment, it automatically marks the status as paid. [Mentioned here](https://stripe.com/docs/billing/invoices/subscription#sub-invoice-lifecycle).

- Credit Card payments
  - The UI assumes that on a successful payment, the `latest_invoice` will have a `payment_intent`. However in the case of a zero invoice, there won't be a payment intent, which is why the UI was throwing an error.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -222,7 +223,8 @@ export class PaypalProcessor {
    */
   private async makePaymentAttempt(invoice: Stripe.Invoice) {
     const customer = invoice.customer as Stripe.Customer;
-    if (invoice.amount_due === 0) {
+    const minAmount = STRIPE_MINIMUM_CHARGE_AMOUNTS[invoice.currency] || 50;
```
chenba, line 226: This `50` seems arbitrary.  If we cannot find a min amount, then process the invoice.  If we get an error, then we'll know what needs to be fixed.

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
```diff
@@ -1,11 +1,33 @@
...
 import '@testing-library/jest-dom/extend-expect';
-import CouponForm from './index';
-import { Coupon } from '../../lib/Coupon';
-import { SELECTED_PLAN } from '../../lib/mock-data';
+
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import React from 'react';
 import waitForExpect from 'wait-for-expect';
 
+import CouponForm from '.';
```
chenba, line 7: TIL

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
```diff
@@ -28,6 +50,38 @@ describe('CouponForm', () => {
...
+      expect(queryByTestId('coupon-error')).toBeInTheDocument();
+      expect(mockSetCoupon).toBeCalledWith(undefined);
+    });
+
+    const couponError = getByTestId('coupon-error');
+
+    expect(couponError).toHaveTextContent(
+      'The code you entered is invalid or expired.'
+    );
+  });
```
chenba, line 78: Thanks for adding this test.  Doesn't seem to be related to zero charge invoices, but 👍 anyway.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
StaberindeZA: Oh yes that makes more sense.

packages/fxa-auth-server/lib/payments/paypal.ts
```diff
@@ -397,10 +397,10 @@ export class PayPalHelper {
...
    */
   public processZeroInvoice(invoice: Stripe.Invoice) {
-    return Promise.all([
-      this.stripeHelper.finalizeInvoice(invoice),
-      this.stripeHelper.payInvoiceOutOfBand(invoice),
-    ]);
+    // It appears for subscriptions that do not require payment, the invoice
+    // transitions to paid automatially.
+    // https://stripe.com/docs/billing/invoices/subscription#sub-invoice-lifecycle
+    return this.stripeHelper.finalizeInvoice(invoice);
```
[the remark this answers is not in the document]
StaberindeZA: Before removing `payInvoiceOutOfBand` it would actually error on that line, with a message along the lines of "This has already been paid".

That's why I thought it's safe to remove, but I'm not sure if it applies in all scenarios with zero invoices.

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
StaberindeZA: Hmm, TS Hero teaches us all. :D

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
StaberindeZA: It's coming up in my next PR, so I thought might as well add it in here as well. :)

StaberindeZA (commented):
I'll get started on the changes, but thought I'd leave some replies already.

packages/fxa-payments-server/src/components/CouponForm/index.test.tsx
```diff
@@ -28,6 +50,38 @@ describe('CouponForm', () => {
...
+      );
+    };
+
+    const { queryByTestId, getByTestId } = subject();
+    fireEvent.change(getByTestId('coupon-input'), { target: { value: 'a' } });
+    fireEvent.click(getByTestId('coupon-button'));
+
+    await waitForExpect(() => {
+      expect(queryByTestId('coupon-error')).toBeInTheDocument();
+      expect(mockSetCoupon).toBeCalledWith(undefined);
```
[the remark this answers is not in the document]
StaberindeZA: Oh actually in this case we do want to ensure it's not called with any arguments.

packages/fxa-auth-server/lib/payments/paypal.ts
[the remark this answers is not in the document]
chenba: Ah I see.  I think it would be nice to still mark it as paid out of band (not a blocker though).

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -15,23 +15,30 @@ import sentry from '../../lib/sentry';
...
       priceId: planId,
       promotionCode,
     });
 
     if (!discount?.amount) {
       throw new Error('No discount for coupon');
     }
+
+    if (total < getMinimumAmount(currency) && total !== 0) {
+      throw new Error('Total below minimum amount allowed.');
```
chenba, line 36: I know that this error is currently not shown the end users.  And I don't think we should.  It would be a strange one to actually show users since this is likely a misconfiguration on the price.

I'll file a jira on preventing that misconfiguration.

packages/fxa-auth-server/lib/payments/paypal-processor.ts
```diff
@@ -222,7 +223,8 @@ export class PaypalProcessor {
    */
   private async makePaymentAttempt(invoice: Stripe.Invoice) {
     const customer = invoice.customer as Stripe.Customer;
-    if (invoice.amount_due === 0) {
+
+    if (invoice.amount_due < getMinimumAmount(invoice.currency)) {
```
chenba, line 227: `<=` I think, if I understand the Stripe docs correctly.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
```diff
@@ -15,23 +15,30 @@ import sentry from '../../lib/sentry';
...
+    }
+
     return discount.amount;
   } catch (err) {
+    console.log(err);
```
chenba, line 41: Delete?

packages/fxa-shared/subscriptions/stripe.ts
```diff
@@ -231,3 +259,12 @@ export const getSubscriptionUpdateEligibility: (
 
   return SubscriptionUpdateEligibility.INVALID;
 };
+
+export const getMinimumAmount = (currency: string): number => {
+  if (STRIPE_MINIMUM_CHARGE_AMOUNTS.hasOwnProperty(currency))
+    return STRIPE_MINIMUM_CHARGE_AMOUNTS[currency];
```
chenba, line 265: nit: I think we usually wrap the if block in `{}` even when there's just one statement.  (Not a convention written down anywhere though. 🤷)

packages/fxa-shared/subscriptions/stripe.ts
```diff
@@ -231,3 +259,12 @@ export const getSubscriptionUpdateEligibility: (
 
   return SubscriptionUpdateEligibility.INVALID;
 };
+
+export const getMinimumAmount = (currency: string): number => {
+  if (STRIPE_MINIMUM_CHARGE_AMOUNTS.hasOwnProperty(currency))
```
chenba, line 264: Any reason you prefer `hasOwnProperty` over `STRIPE_MINIMUM_CHARGE_AMOUNTS[currency]`?

packages/fxa-auth-server/lib/payments/paypal-processor.ts
chenba: haha oops, never mind.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
pdehaan: Interesting. That should be the ESLint `curly` rule, IIRC.
https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/default.js#L21

The ESLint `plugin:fxa/default` config should set this to `'curly': [2, 'all'],` (where `2` is old-style for "error").
I don't see why this would not be enabled via the following configs:
| ESLint CONFIG | EXTENDS
|:----|:-----
| https://github.com/mozilla/fxa/blob/main/_dev/.eslintrc | `"extends": ["plugin:fxa/recommended"],`
| https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/.eslintrc | `"extends": ["plugin:fxa/recommended"],`
| https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/recommended.js | `extends: ['plugin:fxa/default', 'prettier']`
| https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/default.js | `extends: ['eslint:recommended'],`

And yet, looking at ESLint's `--parse-config` for packages/fxa-shared/subscriptions/stripe.ts, it looks like the `curly` rule is disabled:
```sh
npx eslint --config=_dev/.eslintrc.json --print-config packages/fxa-shared/subscriptions/stripe.ts | jq '.rules.curly' | npx prettier --parser=json
# [0, "all"]
```

So I think the only other candidate is probably the rule is being disabled by **eslint-config-prettier**:
https://github.com/prettier/eslint-config-prettier/blob/7ee41f151c7b17d2a6d5cc1211b3910712eadc1b/index.js#L10

```js
    "curly": 0,
```

packages/fxa-shared/subscriptions/stripe.ts
pdehaan: Interesting. That should be the ESLint `curly` rule, IIRC.
https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/default.js#L21

The ESLint `plugin:fxa/default` config should set this to `'curly': [2, 'all'],` (where `2` is old-style for "error").
I don't see why this would not be enabled via the following configs:
| ESLint CONFIG | EXTENDS
|:----|:-----
| https://github.com/mozilla/fxa/blob/main/_dev/.eslintrc | `"extends": ["plugin:fxa/recommended"],`
| https://github.com/mozilla/fxa/blob/main/packages/fxa-shared/.eslintrc | `"extends": ["plugin:fxa/recommended"],`
| https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/recommended.js | `extends: ['plugin:fxa/default', 'prettier']`
| https://github.com/mozilla/eslint-plugin-fxa/blob/fbcfb9db9881d644d147fa9461a4273884160553/lib/configs/default.js | `extends: ['eslint:recommended'],`

And yet, looking at ESLint's `--parse-config` for packages/fxa-shared/subscriptions/stripe.ts, it looks like the `curly` rule is disabled:
```sh
npx eslint --config=_dev/.eslintrc.json --print-config packages/fxa-shared/subscriptions/stripe.ts | jq '.rules.curly' | npx prettier --parser=json
# [0, "all"]
```

So I think the only other candidate is probably the rule is being disabled by **eslint-config-prettier**:
https://github.com/prettier/eslint-config-prettier/blob/7ee41f151c7b17d2a6d5cc1211b3910712eadc1b/index.js#L10

```js
    "curly": 0,
```

packages/fxa-shared/subscriptions/stripe.ts
pdehaan: I think ESLint has feelings on this as well; https://eslint.org/docs/rules/no-prototype-builtins
But interestingly, <kbd>eslint --print-config</kbd> reports that the "no-prototype-builtins" rule is set to `"error"` currently.

```sh
npx eslint --config=_dev/.eslintrc.json --print-config packages/fxa-shared/subscriptions/stripe.ts | jq --sort-keys '.rules["no-prototype-builtins"]' | npx prettier --parser=json
# ["error"]
```

**NOTE:** But I have a feeling that we're not running ESLint against .ts/.tsx files, so not sure this would have been caught by husky/lint-staged.
I need to write up a Google doc on my limited understanding of the current state of linting in this repo.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
StaberindeZA: One day I'll get a PR up without any console.logs...not today though.

packages/fxa-shared/subscriptions/stripe.ts
StaberindeZA: I was playing around trying to get JS to throw an error when it can't find a key in an object, and I think `hasOwnProperty` just stuck around from after that. (I was trying to avoid the `if` all together. Any ideas how to do that?)

I'll change it to your suggestion though. Does make it easier to read.

packages/fxa-auth-server/lib/payments/paypal.ts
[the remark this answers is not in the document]
StaberindeZA: Unfortunately even with `auto_advance` set to false on the finalize call, it looks like the invoice is still returned as `paid` after finalizes completes.

packages/fxa-shared/subscriptions/stripe.ts
StaberindeZA: I'll play around a bit with my `formatOnSave` settings in VSCode. That might be the cause of this.

For now I've manually fixed it.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
chenba: hmm does this need to be on the front end at all?  I believe the invoice preview endpoint can handle it.

packages/fxa-shared/subscriptions/stripe.ts
```diff
@@ -231,3 +259,13 @@ export const getSubscriptionUpdateEligibility: (
...
 };
+
+export const getMinimumAmount = (currency: string): number => {
+  if (STRIPE_MINIMUM_CHARGE_AMOUNTS[currency]) {
+    return STRIPE_MINIMUM_CHARGE_AMOUNTS[currency];
+  }
+
+  throw new Error(
+    `Currency "${currency}" does not have a minimum amount available.`
+  );
```
chenba, line 270: This is ok for now since I don't think anything is catching this error.  But generally speaking, determining the type of error by checking the error message text is not ideal.

chenba (changes requested):
Make sense overall.  I think we can keep that logic backend only by adding a status on the response of the invoice preview endpoint.

packages/fxa-payments-server/src/components/CouponForm/index.tsx
chenba: I wrote a semi-long winded response and then deleted it because I think I should move back a few steps: what's this validation for?  If it's just to show some message on the front end but still allow the subscription (is it still an error then?), then sure let's keep it on the front end.

StaberindeZA: I would say it is an error, but in the context of the Coupon.

It won't block the subscription from going through, but it will block the coupon from being applied.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -716,18 +718,28 @@ export class StripeHelper {
...
+        const minAmount = getMinimumAmount(currency);
+        if (total !== 0 && minAmount && total < minAmount) {
+          throw error.invalidPromoCode(promotionCode);
+        }
+      } catch (error) {
+        if (
+          error instanceof AppError &&
+          error.errno === AppError.ERRNO.INVALID_PROMOTION_CODE
+        ) {
+          throw error;
```
chenba, line 741: Since we are only re-throwing the invalid promotion code error here does that mean if `previewInvoice` or `getMinimumAmount` throws an error, we simply won't know about it?

packages/fxa-payments-server/src/lib/apiClient.ts
```diff
@@ -412,10 +414,6 @@ export async function apiRetrieveCouponDetails(params: {
       { body: JSON.stringify(params) }
     );
 
-    if (!result.discountAmount) {
-      throw new Error(CouponErrorMessageType.Generic);
-    }
-
     Amplitude.coupon_FULFILLED(metricsOptions);
```
chenba, line 417: Now that we don't throw in this function anymore, maybe we should clarify whether this metric mean successful request or valid coupon?

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -156,3 +156,7 @@ export const SubscriptionUpdateEligibility = {
 
 export type SubscriptionUpdateEligibility =
   typeof SubscriptionUpdateEligibility[keyof typeof SubscriptionUpdateEligibility];
+
+export enum SubscriptionStripeErrorType {
+  NO_MIN_AMOUNT_FOR_CURRENCY = 'Currency does not have a minimum amount available.',
```
chenba, line 161: minimum _charge_ amount maybe?

packages/fxa-shared/subscriptions/types.ts
```diff
@@ -156,3 +156,7 @@ export const SubscriptionUpdateEligibility = {
...
+}
```
chenba, line 162: Have you considered extending `Error` to give these error a different type?

chenba (commented):
Looks good overall.  There are a couple concerns on error throwing/handling changes that I'd see addressed.

packages/fxa-payments-server/src/lib/apiClient.ts
StaberindeZA: The VPN Coupons MVP doc says the following.

```
coupon_error: coupon validation failed. Capture failure type with coupon code used
coupon_success: fire when user presses apply and coupon is valid

```

It sounds like only fire the success event when the coupon is actually valid.

As part of these changes, I moved all the coupon metric events to the CouponForm component, since that's where the logic is at the moment that determines whether or not a coupon is valid.

packages/fxa-auth-server/lib/payments/stripe.ts
StaberindeZA: A good catch. Added an else, which will report the error to Sentry.

packages/fxa-auth-server/test/local/payments/stripe.js
```diff
@@ -1300,6 +1349,68 @@ describe('StripeHelper', () => {
       });
       assert.deepEqual(actual, expected);
     });
+
+    it('retrieves details when previewInvoice rejects', async () => {
```
chenba, line 1353: Very possible that it was just me but I found this test title confusing.  I was trying to understand _what_ is retrieving coupon details when previewInvoice rejects, when there's no retrieving call of any sort after previewInvoice in the function.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -716,18 +718,36 @@ export class StripeHelper {
...
+        } else {
+          Sentry.withScope((scope) => {
+            scope.setContext('retrieveCouponDetails', {
+              priceId,
+              promotionCode,
+            });
+            Sentry.captureException(error);
+          });
```
chenba, line 749: I cannot find a test for this new code.

packages/fxa-shared/subscriptions/types.ts
chenba: Not a blocker.  But I was thinking that we could use `instanceOf` to check for the type of error.  `SubscriptionStripeErrorType` seems too generic for that purpose.  E.g. `new StripeNoMinChargeCurrencyError('de')`.  Something like that.

chenba (approved):
Well, patches grow big so quickly these day!  😁  It looks good.  Got a few comments.  Would like to see that missing test added at least.

packages/fxa-auth-server/lib/payments/stripe.ts
StaberindeZA: Updated the one test to check that Sentry was called, and added another one for getMinAmount exceptions.

packages/fxa-auth-server/test/local/payments/stripe.js
StaberindeZA: I updated the text a little that should hopefully clarify it a bit. 

These are all `retrieveCouponDetails` tests. That test specifically should still return a result even though the method called by it, previewInvoice, rejects.

StaberindeZA:
> Well, patches grow big so quickly these day! 😁 It looks good. Got a few comments. Would like to see that missing test added at least.

Thank you so much for the multiple code reviews, especially considering things changed a few times throughout this PR.

PR #11722, opened 2022-01-24
refactor(scripts): removing old unused scripts

clouserw:
Because:

* `gh-pages.sh` and `upload-assets-to-cdn.sh` are no longer used

This commit:

* removes them.  Once this lands I will update the documentation as
well.


chenba:
huh?  How do we upload to the CDN now?

chenba (approved):

PR #11733, opened 2022-01-25
chore(scripts): Add 'other' assets directory to upload script

### LINKED ISSUE #4478: Firefox account requests two factor authentication or recovery  key even if they have been disabled or revoked [bz1621175]
From https://bugzilla.mozilla.org/show_bug.cgi?id=1621175

User Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.132 Safari/537.36

Steps to reproduce:

I disabled the two factors authentication and revoked the recovery key the last time I signed according to the following information:
https://support.mozilla.org/en-US/kb/changing-your-two-step-authentication-device-firefox-account
https://support.mozilla.org/en-US/kb/reset-your-firefox-account-password-recovery-keys

Actual results:

Firefox sync account requests two factor authentication or recovery key even if they have been removed.

I cannot access to my account because I deleted also the two factor authentication and recovery key. I only can change the password.

Expected results:

Firefox sync account let me sign without two step authentication and recovery key or let me delete the account and create a new with the same email that previously I used.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXA-1331)
┆Issue Number: FXA-1331

clouserw:
Because:

* the /other/ directory doesn't get uploaded for some reason

This commit:

* adds it to the upload script

Closes #4478

-------------------

Looking for a review from @dlactin or @jbuck .  I noticed the original script was specifically *excluding* the other directory, so, aside from a review for if this will work or not, I want to make sure this is what we want.  Lastly, I noticed it's supposed to be sending a notice to the slack channel, but I'm not sure I've ever seen that notice -- is that hooked up properly?

### DIFF
--- a/.github/workflows/upload-assets-to-cdn.yml
+++ b/.github/workflows/upload-assets-to-cdn.yml
@@ -11,7 +11,6 @@ on:
       - main
     paths:
       - 'assets/**/'
-      - '!assets/other/**'
 
 jobs:
   Upload-assets-to-CDN:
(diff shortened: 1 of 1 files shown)


jbuck (approved):

jbuck:
> Lastly, I noticed it's supposed to be sending a notice to the slack channel, but I'm not sure I've ever seen that notice -- is that hooked up properly?

Lets find out when we merge this PR

dlactin (approved):

chenba (approved):

Issue #11313, opened 2021-12-13
Add frontend screens login screens


data-sync-user:
From [https://docs.google.com/document/d/1pFzugkDOIR6eoXrBCx9FWWJhfnxFgWnRtk2mWFOp8DQ/edit](https://docs.google.com/document/d/1pFzugkDOIR6eoXrBCx9FWWJhfnxFgWnRtk2mWFOp8DQ/edit|smart-link) we will need to modify our sign-in page to show the `Continue with Google` button.

Old approach 
1. Load the Google auth script [https://apis.google.com/js/api:client.js](https://apis.google.com/js/api:client.js)
  1. I’m thinking dynamically loading the library in mixin when the user is part of the experiment or if pocket client is loaded
1. In our signin mixin we can initialize and complete signin if the user chooses this option
1. After authenticating we will get an ID token and verify it on our backend (TBD on endpoint name)
1. Upon successful validation, we will redirect the user to their intended service

New approach
- Use Google oauth api to get a code and then have FxA server exchange it for and access token
  - Adds a new page for the oauth redirect
  - Stores oauth query params in local storage so that we continue login
- Create a mixin to handle all oauth related calls
- Redirect the user back to the RP after succesful verification

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4307)

PR #11743, opened 2022-01-25
feat(google): Add Google auth front end screens

vbudhram:
## Because

- This allows the user to log into FxA with a Google account
- By default this is disabled on the frontend

## This pull request

- Adds frontend screens
- Uses Google oauth api directly instead of loading their library
- Shuffles around backend routes to use `/linked_accounts` path

Closes: https://github.com/mozilla/fxa/issues/11313

## Screenshots (Optional)

This UX is based on the login page from Pocket

Screen Shot 2022-01-25 at 10 58 04 AM


packages/fxa-content-server/app/scripts/templates/index.mustache
```diff
@@ -29,6 +29,8 @@
         <button id="submit-btn" type="submit">{{#t}}Continue{{/t}}</button>
       </div>
     </form>
+    {{{ unsafeThirdPartyAuthHTML }}}
```
vbudhram, line 32: Commenting this out would disable the `Continue with Google` button

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1816,20 +1734,6 @@ export const accountRoutes = (
       },
       handler: (request: AuthRequest) => accountHandler.login(request),
     },
-    {
-      method: 'POST',
-      path: '/account/login/third_party',
```
vbudhram, line 1821: This was moved to own route file `linked-accounts`

packages/fxa-content-server/app/scripts/views/post_verify/third_party_auth/callback.js
```diff
@@ -0,0 +1,20 @@
...
+
+import BaseView from '../../base';
+import Template from '../../../templates/post_verify/third_party_auth/callback.mustache';
+import GoogleAuthMixin from '../../mixins/google-auth-mixin';
+import Cocktail from 'cocktail';
+
+class ThirdPartyAuthCallback extends BaseView {
+  template = Template;
+
+  async afterVisible() {
```
vbudhram, line 13: While this file is pretty sparse, it will be instrumented with metrics in some follow up issues.

packages/fxa-content-server/app/scripts/lib/fxa-client.js
```diff
@@ -1431,6 +1431,14 @@ FxaClientWrapper.prototype = {
       return getUpdatedSessionData(email, relier, accountData);
     });
   }),
+
+  verifyAccountThirdParty: withClient((client, relier, token, provider) => {
```
dannycoates, line 1435: nit: We've always been comically inconsistent between "login" and "signin", but I think one of those would be better than "verify" in names throughout here. Maybe also in the route name `/linked_account/login`

dannycoates (approved):
this looks great @vbudhram 🎉

PR #11753, opened 2022-01-26
fix(auth): Copy 'public' to dist, don't load non-existent Fluent bundles, Fluent strategy tweak

### LINKED ISSUE #11745: Fluent localization in new emails not showing
QA did some l10n testing for new emails a month or two ago and some of the translations were present but not all of them. We figured we needed to wait for the l10n team to retranslate strings that weren’t leveraged from the old email strings, but it was recently reported that the translations still weren’t in place. We have a cleanup ticket for removing gettext translations (used for titles, location, probably other small things), but it’s possible we just saw the translations from gettext and assumed it was working.

I started looking into it this morning and what’s in 224.1 has German translations in place, for example, in the `auth.ftl` file but changing my locale to German, they aren’t pulled in. It looks to be a problem locally as well, currently investigating.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4488)

LZoog:
Because:
* For Fluent translations, we pull in ftl files from public/locales, which wasn't present in dist
* Fluent was attempting to load bundles from 'OTHER_EN_LOCALES' which was causing "missing translation" errors
* We were attempting to load English dialects that may not have been correct, e.g. showing 'en-CA' dialect to US users, due to the Fluent strategy

This commit:
* Adds 'public' dir to fxa-auth-server copy list on build
* Tells the bundler to load 'en-GB' if the user locale matches one in 'OTHER_EN_LOCALES', moved into 'fxa-shared' to share with Payments and renamed to EN_GB_LOCALES
* Changes the Fluent strategy from 'filtering' to 'matching'
* Adds tests

fixes #11745

---

I think unfortunately when we work on #11529 is the right time to add the tests for this.

To repro locally:
* Change your browser language to something else, like German (de)
* Change auth-server's `dev.json` `mjmlEnabledEmail` to whatever you want to use locally (currently `b@c.com`)
* Run `yarn build` in the auth server off of `main`
* `yarn start` FxA, then in the auth server, stop the auth process, go into `dist`, and run `yarn start`
* Use MailDev. Register for a new account locally, observe that the sign up email is only partially translated (some gettext translations still exist, will be taken care of in #11471)

To see the fix, follow the same steps as above except run the yarn commands in this branch.

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
+++ b/packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
@@ -7,9 +7,7 @@ import { FluentBundle, FluentResource } from '@fluent/bundle';
 import { negotiateLanguages } from '@fluent/langneg';
 import { LocalizerBindings, TemplateContext } from './localizer-bindings';
 import availableLocales from 'fxa-shared/l10n/supportedLanguages.json';
-import { PatternElement, VariableReference } from '@fluent/bundle/esm/ast';
-
-const OTHER_EN_LOCALES = ['en-NZ', 'en-SG', 'en-MY'];
+import { EN_GB_LOCALES } from 'fxa-shared/l10n/otherLanguages';
 
 const RTL_LOCALES = [
   'ar',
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -24,6 +21,11 @@ const RTL_LOCALES = [
   'ug',
 ];
 
+// these do not exist in the l10n repo, but we want users with these locales to
+// receive the 'en-GB' translations to see 'US $x.xx' instead of the en fallback '$x.xx'
+// which is ambiguous
```
LZoog, line 26: This comment came from #5948

LZoog: I moved the array into `fxa-shared` instead of redefining it.

packages/fxa-auth-server/lib/senders/emails/fluent-localizer.ts
```diff
@@ -39,21 +37,23 @@ class FluentLocalizer {
...
       {
         defaultLocale: 'en',
       }
     );
 
     const fetched: Record<string, string> = {};
-    for (const locale of currentLocales.filter(
-      (l) => !OTHER_EN_LOCALES.includes(l)
-    )) {
+    for (const locale of currentLocales) {
```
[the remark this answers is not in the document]
dschom: Alright, tests added, and I feel better about this class now.

packages/fxa-auth-server/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "test"
   },
   "scripts": {
-    "build": "yarn emails-scss && tsc --build && cp -R config fxa-content-server-l10n lib scripts dist/fxa-auth-server",
+    "build": "yarn emails-scss && tsc --build && cp -R config fxa-content-server-l10n public lib scripts dist/fxa-auth-server",
```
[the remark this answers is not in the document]
dschom: As we talked about in a side discussion, this seems okay as long as the fxa-content-server-l10n gets phased out, and it appears there's a maintenance task for this.

dschom (approved):
LGTM!

packages/fxa-auth-server/test/local/senders/emails/fluent-localizer.ts
```diff
@@ -2,13 +2,155 @@
...
+
+      assert.deepEqual(currentLocales, ['de', 'it', 'en-US', 'en']);
+    });
+
+    it('selects the proper locale', async () => {
+      const { selectedLocale } = await localizer.setupLocalizer(
+        'de-DE,en-US;q=0.7,en;q=0.3'
+      );
+
+      // This needs investigation.
```
LZoog: I'm taking out this comment since I don't think this is true any longer.

Also fixing a typo in a comment and adding another comment.

LZoog (commented):
@dschom Thanks again for the collab here and those tests. I made a couple of minor tweaks I've called out but I'm going to go ahead and squash & merge this once it passes CI again, but I'm going to change the base into the train `uplift-2` branch to make it easier to run a dot release.

Issue #11759, opened 2022-01-26
[intermittent] “Welcome to Mozilla VPN: Please set your password” email has some strings colored in purple


data-sync-user:
**Affected FxA versions:**

* 1.224.1;

**Prerequisites:**

* Be in a country where Mozilla VPN is available;
* Have no FxA account;
* Be on the Set up your Subscription page;

**Steps to reproduce:**

1. Subscribe to Mozilla VPN while using valid credentials for the country you are in;
1. Open your email client;
1. Open the “Welcome to Mozilla VPN: Please set your password” email;
1. Observe the content of the email;

**Expected result:**

* The content of the email has no display issues and is displayed in black;

**Actual result:**

* “Charged” and “Next Invoice” strings are purple; 

**Notes:**

* Reproducible regardless of the payment method and price plan;
* See the attached screenshots;

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4494)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/19829/Screen+Shot+2022-01-26+at+1.16.59+PM.png">Screen Shot 2022-01-26 at 1.16.59 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/19789/Screen+Shot+2022-01-26+at+12.54.06.png">Screen Shot 2022-01-26 at 12.54.06.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/19790/Screen+Shot+2022-01-26+at+12.55.11.png">Screen Shot 2022-01-26 at 12.55.11.png</a>

data-sync-user:
➤ Lisa Chan commented:

Bianca Hidecuti I tried to reproduce the results, but the strings are not purple (see screenshot). Can you please let us know what email client you are using? Thank you!

!Screen Shot 2022-01-26 at 1.16.59 PM.png|width=955,height=1126!

data-sync-user:
➤ Bianca Hidecuti commented:

Lisa Chan I am using Gmail email client.

data-sync-user:
➤ Lisa Chan commented:

Bianca Hidecuti I used Gmail as well and I am still unable to reproduce this,  but we are going to revise the stylesheet to have the font color of the list items match the rest of the text. 

We are planning to land this fix in 224.2, if this bug still appears when it is re-tested, can you please inspect the element for us and include the DOM/styles within the screenshot? Thank you! 🙂

[...]

PR #11769, opened 2022-01-27
fix(auth-server): `subscriptionAccountFinishSetup` email has strings in purple

xlisachan:
While I unable to reproduce the bug within the issue, the revised stylesheet sets the font color of `li` items to match the rest of the `div`.

Closes: #11759

## Checklist

## Screenshots (Optional)
Old
Screen Shot 2022-01-26 at 12 54 06

Current
Screen Shot 2022-01-27 at 9 58 36 AM

### DIFF
--- a/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.scss
+++ b/packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.scss
@@ -28,6 +28,11 @@
   @extend .mb-6;
 }
 
+// NOTE: some list items appeared purple in stage
+.text-body li {
+  color: global.$grey-500 !important;
+}
+
 .text-body-no-bottom-margin div {
   @extend %text-body-common;
   @extend .mb-0;


LZoog:
@xlisachan did you check this in Gmail since QA got back to you and stated the client was in Gmail, and you're still unable to reproduce?

I don't mind merging this much since it's a very small change, but it feels weird that we can't repro. 🤔

packages/fxa-auth-server/lib/senders/emails/layouts/subscription/index.scss
```diff
@@ -28,6 +28,10 @@
   @extend .mb-6;
 }
 
+.text-body li {
```
LZoog, line 31: We could add a comment above this with a short reason why we need it. Otherwise someone might remove this later because they don't see any visual difference.

LZoog (approved):

xlisachan:
Confirming that I checked this in Gmail and was unable to reproduce it. :/

LZoog:
> Confirming that I checked this in Gmail and was unable to reproduce it. :/

I would say definitely have QA retest this when we merge it to confirm it's fixed. If it's not, we should remove the code and possibly have them inspect the element for us and take a screenshot of the DOM/styles applied so we can see what's going on.

PR #11774, opened 2022-01-27
chore(repo): set fxa-devs to own all things

chenba:
Because:
 - we shouldn't depend on humans to remember to request a PR review from
   mozilla/fxa-devs

This commit:
 - update and move CODEOWNERS so that mozilla/fxa-devs is the owner of
   the entire repo


dannycoates (approved):
Let's try it 👍

PR #11787, opened 2022-01-28
fix(subscriptions): match expandable resource type to Stripe obj type list

chenba:
Because:
 - a mismatch between two Stripe type lists led to false negatives on
   resource expansion

This commit:
 - update the valid expand resource types to base on the Stripe object
   to types map


StaberindeZA (approved):
LGTM. TY

PR #11790, opened 2022-01-28
feat(config): Add options for connection pool limit & queue limit

jbuck:
## Because

- We want to increase the mysql connection limit

## This pull request

- Adds an option to configure the mysql connection limit (and the queue limit just in case)

Closes: # (issue number)

## Checklist

## Screenshots (Optional)

### DIFF
--- a/packages/fxa-profile-server/lib/config.js
+++ b/packages/fxa-profile-server/lib/config.js
@@ -186,6 +186,14 @@ const conf = convict({
       default: '3306',
       env: 'MYSQL_PORT',
     },
+    connectionLimit: {
+      default: 10,
+      env: 'MYSQL_CONNECTION_LIMIT',
+    },
+    queueLimit: {
+      default: 0,
+      env: 'MYSQL_QUEUE_LIMIT',
+    },
   },
   oauth: {
     url: {


packages/fxa-profile-server/lib/config.js
```diff
@@ -186,6 +186,14 @@ const conf = convict({
...
       env: 'MYSQL_PORT',
     },
+    connectionLimit: {
+      default: 10,
+      env: 'MYSQL_CONNECTION_LIMIT',
+    },
+    queueLimit: {
+      default: 0,
+      env: 'MYSQL_QUEUE_LIMIT',
+    },
```
chenba, line 196: I don't think these are used anywhere in the profile server?  Not sure how this patch is going affect the service when these values aren't used.  😕

jbuck: They should get passed all the way into https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/lib/db/mysql/index.js#L29 right?

chenba: Never mind.  The entire `mysql` config object is passed so searching for these individual configs won't turn up anything.

chenba (approved):

PR #11809, opened 2022-01-31
fix(settings): Disconnect all associated clients/sessions for given client name on sign out

### LINKED ISSUE #11559: Address users needing to "sign out" multiple times due to multiple sessions
We currently filter the connected services list and display unique entries by name, masking multiple sessions as one row in the UI, meaning users may have to click “Sign out” multiple times before the session appears to be cleared or worse yet, they "sign out" and appear to be signed out, then revisit later and are left wondering if the previous operation didn't work because it reappears.

We did it this way because signing into an RP generates both a “Web Session” and an entry for the access/refresh token created for the RP and the list became very long. In the long run, at least at the time of writing, we’d like to separate out a “devices” list and a “services” list, but we need to address this problem in the interim.

It looks to me like we can tweak the client-side to batch multiple mutations in single GQL request for multiple sessions, if multiple sessions exist. In our epic pow wow meeting on 1/6, Ben mentioned we could also do this on the server-side by creating a new mutation to remove them all, but this seems more involved (more code/overhead for a new mutation, resolver, then client-side edits) and we might need an additional account lookup. Since we already have all of the connected service data sent down to the client, let's try batching mutation requests together in one GQL request.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4410)

### LINKED ISSUE #11658: Fix broken ConnectedServices unit test
Somewhere along the way our `.toBeTruthy()` checks changed to `.toBeTruthy`, meaning they didn't really run. In [this PR](https://github.com/mozilla/fxa/pull/10924/files) (#10924), we fixed a lot of these, but for some reason, one in ConnectedServices has a test failure.

Search ConnectedServices tests for “FXA-4453” and you’ll find the test.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4453)

LZoog:
Because:
* Since we display only unique client/service names in the UI, users were having to click 'Sign out' multiple times before the session appeared to be cleared if they had multiple sessions of the same service name

This commit:
* Uses Promise.all to disconnect all clients associated with the selected client if multiple sessions exist

fixes #11559
fixes #11658

---

One way to reproduce this and the fix locally, assuming your Nightly version doesn't match FF:
1. On `main`, create/login to an account locally on Nightly
1. Open a private window in Nightly and login to that account - this creates two sessions under one name
1. Log in on FF (not Nightly). Click "sign out" for the Nightly session. Refresh the page, and see that the Nightly session is still present

To verify the fix:
* On this branch, when refreshing the page on step 3, you'll see the Nightly session is no longer present

Also, it's possible we could send all the mutation requests up in one network request instead of using a `Promise.all` for parallel requests, but it'd be more leg work and from what I've seen I don't think users are typically amassing more than 2-3 unique sessions where the performance gains would be minimal.

### DIFF
--- a/packages/fxa-settings/src/components/ConnectedServices/index.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/index.tsx
@@ -98,9 +98,11 @@ export const ConnectedServices = () => {
         logViewEvent('settings.clients.disconnect', `submit.${reason}`);
 
         // disconnect all clients/sessions with this name since only unique names
-        // are displayed to the user
+        // are displayed to the user. This is batched into one network request request
+        // via BatchHttpLink
         const clientsWithMatchingName = groupedByName[client.name];
-        if (clientsWithMatchingName.length > 1) {
+        const hasMultipleSessions = clientsWithMatchingName.length > 1;
+        if (hasMultipleSessions) {
           await Promise.all(
             clientsWithMatchingName.map(
               async (c) => await account.disconnectClient(c)
@@ -112,7 +114,11 @@ export const ConnectedServices = () => {
 
         // TODO: Add `timing.clients.disconnect` flow timing event as seen in
         // old-settings? #6903
-        if (client.isCurrentSession) {
+        if (
+          client.isCurrentSession ||
+          (hasMultipleSessions &&
+            clientsWithMatchingName.find((c) => c.isCurrentSession))
+        ) {
           clearSignedInAccountUid();
           window.location.assign(`${window.location.origin}/signin`);
         } else if (reason === 'suspicious' || reason === 'lost') {
--- a/packages/fxa-settings/src/components/ConnectedServices/index.tsx
+++ b/packages/fxa-settings/src/components/ConnectedServices/index.tsx
@@ -47,16 +47,15 @@ export function sortAndFilterConnectedClients(
     }
   });
 
-  return sortedAndUniqueClients;
+  return { groupedByName, sortedAndUniqueClients };
 }
 
 export const ConnectedServices = () => {
   const alertBar = useAlertBar();
   const account = useAccount();
   const attachedClients = account.attachedClients;
-  const sortedAndUniqueClients = sortAndFilterConnectedClients([
-    ...attachedClients,
-  ]);
+  const { groupedByName, sortedAndUniqueClients } =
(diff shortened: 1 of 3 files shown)


packages/fxa-settings/src/components/ConnectedServices/index.tsx
```diff
@@ -97,7 +96,20 @@ export const ConnectedServices = () => {
     async (client: AttachedClient) => {
       try {
         logViewEvent('settings.clients.disconnect', `submit.${reason}`);
-        await account.disconnectClient(client);
+
+        // disconnect all clients/sessions with this name since only unique names
+        // are displayed to the user
```
vbudhram, line 101: I would probably mention why we are doing multiple network requests here.

vbudhram (approved):
Thanks @LZoog, I don't have any issues with doing it this way for now but I do think we should have a follow up to disconnect all clients/sessions in one network request.

xlisachan:
r+wc, the fix does allow the user to disconnect multiple sessions of the same service name. Per our discussion, a follow-up could include redirecting the user to the signin page if it was the current session.

LZoog:
@vbudhram Heh so I mostly did it this way since it fixed the bug without much overhead and I didn't think I would have time this sprint to account for that after reading [this](https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/), and [this](https://stackoverflow.com/questions/50889548/how-can-i-run-one-mutation-multiple-times-with-different-arguments-in-one-reques), and some others; it seemed like it might not be worth doing right now, _but_ then I just tried replacing our `HttpLink` with `BatchHttpLink` (see [docs](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/)) and it seems to work automagically with the default options (which seem OK to me), if you check the network tab before and after this commit you'll see two requests vs one. Tests in settings pass, but I'm going to push that up to see CI run, and see if I can address what Lisa found (you're not always redirected to sign in upon signing out) in the meantime.

LZoog:
OK, also pushed a fix + tests to account for redirecting the user to the login page if they have multiple sessions and one of them is currently active.

xlisachan (approved):
r+! Ran it and works! 🎉

Issue #11197, opened 2021-12-01
[subplat] The checkout form should not consider an unverified stub account an "existing" Firefox account


data-sync-user:
When someone uses the checkout form to create an account and subscribe to a plan (the "new flow"), we check whether the email address belongs to an existing account.  And if so, we require the user to sign in before continuing to subscribe.

On the same form, we'd create a unverified, stub account on submit for the user as a FxA account is required for a Stripe customer.  When an subscription attempt fails, this stub account is considered an existing account on subsequent existing account checks.

STR:
1. Visit the checkout page for a plan
2. Use an email address that does not belong to any account in the testing env 
3. Use `4000000000000341` for the CC number
4. Fill out the rest of the form with valid values
5. Submit the form
6. Dismiss the error message by pressing "Try again"
7. Place a cursor into the first email address input and then tab out of it, or repeat steps 1 and 2 by reloading the page or using a new tab

Expected:
The user is allowed to the use the same email address that was previously used in the failed subscription attempt

Actual:
The user is told that they have an existing account, despite that the user never set a password or verified the email address.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4273)

data-sync-user:
➤ Ivo Plamenac commented:

We need to determine whether or not we can delete an account. If this occurs because of a pending authorization, deleting the account will cause them to lose access when the payment may eventually complete.

biancadanforth:
Sentry issue: [FXA-AUTH-PROD-GC](https://sentry.prod.mozaws.net/operations/fxa-auth-prod/issues/15647824/?referrer=github_plugin)

This was previously linked to #10303, but we're still seeing the error after the fix for that ticket got into prod, so I'm linking the second ticket in its place in Sentry to more easily track whether it is resolved once this issue hits prod in Train 226.

PR #11812, opened 2022-01-31
fix(subscriptions): stop considering unverified stub accts as existing

chenba:
Because:
 - the checkout form for the "passwordless" subscription flow treats an
   unverified account as an existing account, even though a user could
   easily get into that state by encountering an error while trying to
   subscribe (e.g. a failed payment)

This commit:
 - update the account status endpoint to return two properties: exists
   and verified
 - update the payments frontend to consider only verified accounts as
   existing accounts

Closes: #11197

### DIFF
--- a/packages/fxa-auth-server/lib/routes/account.ts
+++ b/packages/fxa-auth-server/lib/routes/account.ts
@@ -1065,9 +1065,10 @@ export class AccountHandler {
     await this.customs.check(request, email, 'accountStatusCheck');
 
     try {
-      const exist = await this.db.accountExists(email);
+      const account = await this.db.accountRecord(email);
       return {
-        exists: exist,
+        exists: !!account,
+        verifierSet: !!account && account.verifierSetAt > 0,
       };
     } catch (err) {
       if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
@@ -1762,6 +1763,7 @@ export const accountRoutes = (
         response: {
           schema: {
             exists: isA.boolean().required(),
+            verifierSet: isA.boolean().optional(),
           },
         },
       },
--- a/packages/fxa-auth-server/test/local/routes/account.js
+++ b/packages/fxa-auth-server/test/local/routes/account.js
@@ -3956,3 +3956,43 @@ describe('/account/ecosystemAnonId', () => {
     });
   });
 });
+
+describe('POST /account/status', () => {
+  let dbMock = {};
+  const email = 'testo@example.gg';
+  const mockRequest = mocks.mockRequest({ payload: { email } });
+
+  it('responds with false exists prop when account is not in db', async () => {
+    dbMock = {
+      accountRecord: sinon.stub().throws(new error.unknownAccount(email)),
+    };
+    const accountRoutes = makeRoutes({ db: dbMock });
+    const route = getRoute(accountRoutes, '/account/status', 'POST');
+    await runTest(route, mockRequest, (res) => {
+      assert.deepEqual(res, { exists: false });
+      sinon.assert.calledOnceWithExactly(dbMock.accountRecord, email);
+    });
+  });
+
+  it('responds with exists:true and verified:false properties', async () => {
+    dbMock = {
+      accountRecord: sinon.stub().resolves({ verifierSetAt: 0 }),
+    };
+    const accountRoutes = makeRoutes({ db: dbMock });
+    const route = getRoute(accountRoutes, '/account/status', 'POST');
+    await runTest(route, mockRequest, (res) => {
+      assert.deepEqual(res, { exists: true, verifierSet: false });
+    });
+  });
+
+  it('responds with exists:true and verified:true properties', async () => {
+    dbMock = {
+      accountRecord: sinon.stub().resolves({ verifierSetAt: 9001 }),
+    };
+    const accountRoutes = makeRoutes({ db: dbMock });
+    const route = getRoute(accountRoutes, '/account/status', 'POST');
+    await runTest(route, mockRequest, (res) => {
+      assert.deepEqual(res, { exists: true, verifierSet: true });
+    });
+  });
+});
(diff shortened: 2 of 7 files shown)


packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1065,9 +1065,10 @@ export class AccountHandler {
     await this.customs.check(request, email, 'accountStatusCheck');
 
     try {
-      const exist = await this.db.accountExists(email);
+      const account = await this.db.accountRecord(email);
       return {
-        exists: exist,
+        exists: !!account,
+        verified: !!account && account.verifierSetAt > 0,
```
chenba, line 1071: @vbudhram do you have an idea of how unsafe this additive change is?  If existing clients/RPs use the response without doing any sort of validation on its exact shape, it should be all right I'd think.  Also, this would expose the verified status of an email, which I'm not quite sure about.

vbudhram: > do you have an idea of how unsafe this additive change is?

I'm not actually sure how many RP/clients use this endpoint. It is currently used on the [FxA frontend](https://github.com/mozilla/fxa/blob/9fd866c8cf02c658d4cebd4e34a4c4dd9589e817/packages/fxa-content-server/app/scripts/views/index.js#L237) to help redirect users to signin/signup if email exists, but that is as far as I can tell.

In FxA, we have had a tendency to overload the meaning of `verified` (session, email, account) and it always get confusing. Using `verifierSet` as response might make things a little more clear.

> Also, this would expose the verified status of an email

We currently have `/account/profile` and `/recovery_email/status` endpoints that exposes similar data, but they require a session token. This endpoint is unauthenicated but does have rate limiting in place.

I'm inclined to say that this is probably ok, if an attacker knew that no password was set, best they could do is force a password reset.

chenba:
@vbudhram Thanks for the input.  Because of the endpoint I'm modifying in this PR, I'd like some buy in from FxA, so could you please review?

@IvoJP or @StaberindeZA could one of you review this PR?

StaberindeZA:
> @vbudhram Thanks for the input. Because of the endpoint I'm modifying in this PR, I'd like some buy in from FxA, so could you please review?
> 
> @IvoJP or @StaberindeZA could one of you review this PR?

For sure, I'll take a look.

StaberindeZA (approved):
LGTM. I tested a couple of scenarios and worked well. Very nice change!

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1065,9 +1065,10 @@ export class AccountHandler {
     await this.customs.check(request, email, 'accountStatusCheck');
 
     try {
-      const exist = await this.db.accountExists(email);
+      const account = await this.db.accountRecord(email);
       return {
-        exists: exist,
+        exists: !!account,
+        verifierSet: !!account && account.verifierSetAt > 0,
```
bbangert, line 1071: If the unverified account has a subscription on it already, wouldn't we consider that to exist?

chenba: Where do we make that call?  Not sure if I understand.  The only code making use of that new property is the front end code in this PR.

vbudhram: > If the unverified account has a subscription on it already, wouldn't we consider that to exist?

Yea we should consider that account to still exisit. The code above would return the account in the case, so I think it would still be good?

packages/fxa-auth-server/test/local/routes/account.js
```diff
@@ -3956,3 +3956,43 @@ describe('/account/ecosystemAnonId', () => {
...
+    };
+    const accountRoutes = makeRoutes({ db: dbMock });
+    const route = getRoute(accountRoutes, '/account/status', 'POST');
+    await runTest(route, mockRequest, (res) => {
+      assert.deepEqual(res, { exists: false });
+      sinon.assert.calledOnceWithExactly(dbMock.accountRecord, email);
+    });
+  });
+
+  it('responds with exists:true and verified:false properties', async () => {
```
vbudhram, line 3977: Should be `verifierSet:false` for here and test below.

packages/fxa-auth-server/lib/routes/account.ts
```diff
@@ -1065,9 +1065,10 @@ export class AccountHandler {
     await this.customs.check(request, email, 'accountStatusCheck');
 
     try {
-      const exist = await this.db.accountExists(email);
+      const account = await this.db.accountRecord(email);
       return {
-        exists: exist,
+        exists: !!account,
```
vbudhram, line 1070: I think this can be replaced with `exists: true`? If no account is found, the new method would throw.

packages/fxa-payments-server/src/components/NewUserEmailForm/index.test.tsx
```diff
@@ -192,4 +195,18 @@ describe('NewUserEmailForm test', () => {
       </Localized>
     );
   });
+
+  it('considers an unverified account to not be an existing account', async () => {
```
vbudhram, line 199: nit: Maybe `considers an stub account to not be an exisiting account`

vbudhram (approved):
@chenba Code wise I think this looks good. 

The only question I had was for what would happen if a user created a stub account, subscribes and then goes through the subscription flow again before setting a password. The subscription endpoint would complain that they already have a subscription?

chenba:
> @chenba Code wise I think this looks good, just a few comments/nits.
> 
> The only question I had was for what would happen if a user created a stub account, subscribes and then goes through the subscription flow again before setting a password. The subscription endpoint would complain that they already have a subscription?

Yes.  The subscription on the existing stub account will block the deletion of the stub account when the user tries to go through the flow again.  The error on the front end will be generic though, so it's not great UX either.  If I were to don my Product Manager hat, I'd argue that getting the user through the initial subscription is more important than incorrectly forcing the user to set their password before continue with their first subscription.  However, it's edge-case-y; the user will need to do something to trigger the existing user validation again on the front end.  And since this patch involves changing an auth API endpoint and exposing a new bit of user info, I'm going to close this without merging.  Maybe there's a front end only approach we can take.

Thank you everyone for your review and input!

Issue #11090, opened 2021-11-21
Server: refactor sessions into attached clients


data-sync-user:
We currently return a list of an account’s sessions, however we’d like to also incorporate things such as the devices and oauth clients. You can [see here](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/lib/routes/attached-clients.js#L62:L62) how the auth-server is looking up and merging this info into a consolidated list called “attached clients”, which is used in FxA Settings in the Connected Services section.

Refactor the admin-server’s SessionTokens model to do something similar, by adding device and connected oauth client lookups. Ultimately we should return at least the name of the device or oauth client, when it was last used, and the existing relevant session details.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4232)

PR #11817, opened 2022-02-01
feature(shared,auth,admin): Refactor attached-clients so routines can be shared with admin panel

dschom:
## Because

- The admin panel needs access to the actual list of attached clients, and the logic to do this already exists in the auth server.

## This pull request

- Introduces fxa-shared/connected-services/factories.ts. Classes therein contain the logic for building and merging attached clients from sessions, devices, and oauthClients.
- Introduces  fxa-shared/connected-services/accessors.ts. Classes therein contain the ‘accessors’ layer through which any database (Redis or MySql) interactions flow. Note that these are base classes and are designed to be extended from. Currently the OauthDB class and the OAuthRedisClass in fxa-auth-server are extensions. As more shared functionality is required, code can be hoisted to these base classes to provide shared functionality.
- Introduces fxa-shared/connected-services/formatters. Classes therein manipulate the state of attached clients so that various fields are formatted properly.
- Introduces fxa-shared/util.ts. This is a catch all for auxiliary methods that were generally useful and used in more than one place.
- Introduces fxa-shared/models. This is a catch all for types used by the connected services code. Hopefully this gives it more structure. Note that there are still some places where output format is defined. A follow up item has been added to supply types for all database structures. One question I had while doing this is if these types should actually belong in fxa-shared/db/models auth. I ultimately put them here because they are truly the subset of 'types' needed by connected services.
- Introduces fxa-shared/auth. This holds a hash implementation that lived in the auth server in place that it can now be used more generally. Note that, all references to this in fxa-auth-server should be updated now.
- Introduces fxa-shared/db/redis.ts which provides a base class for Redis interactions that can be extended and shared across packages. Similarly to fxa-shared/connected-services/accessors.ts this class is designed to be extended from.
- Updates fxa-shared/db/mysql.ts to be more typescript-like. This means that MySqlStoreShared is now a proper class.
- Adds tests in fxa-shared/test/connected-services and get all code coverage into a green status.
- Updates fxa-shared/db/config.ts to include the ability to produce build configs. 
- Fixes issue with ‘openHandles’ which was detected by jest in the admin panel. There were two occurrences where open handles, which are basically hanging async operations, existed. 
  - There was no way to close database connections. As a result, close methods have been added to MySqlStoreShared and RedisStoreShared which make sure underlying connections are terminated. During test tear down these close() calls are now invoked and awaited.
  - In the redis implementation, there were lots of Promise.race invocations, where a Redis call was being raced against a timeout. The issue here was that nothing cancelled the setTimeouts thereby leaving hangfire. A small abstraction was added so that these timeouts could be cancelled.
- Updates the admin panel account.resolver to incorporate the ‘attached clients’ that are now accessible via fxa-shared/connected-services.
- Updates the admin panel GraphQL model to include support for attache clients.
Standardize the config in the admin panel and relies on fxa-shared to produce the DB config thereby ensuring it’s consistent.
- Updates tests in the admin panel to ensure the database service is still accessible, and the ‘connected services db’ can be accessed. This required adding more tables to the testAdmin database that gets constructed during testing.
- Added new table declarations to fxa-shared/test/db/models/auth.
- Other small tasks included making sure the source code terms were present in code files, and running TypeScript hero to make sure import statements were organized.

Closes: #11090

## Checklist

I know this is a large PR due to it's refactor nature. The majority of changes here were simply moving code from fxa-auth-server to fxa-shared, adding some abstractions to make the code more reusable, and cleaning things up to support typescript . Ideally all the logic in the underlying routines has gone untouched.


packages/fxa-auth-server/lib/oauth/db/redis.js
```diff
@@ -98,31 +75,6 @@ class OauthRedis {
     return this.redisAccessTokens.removeAccessTokensForUser(uid);
   }
 
-  /**
-   *
-   * @param {Buffer | string} uid
-   * @param {Buffer | string} tokenId
-   * @return {Promise<RefreshTokenMetadata>}
-   */
-  async getRefreshToken(uid, tokenId) {
```
dschom, line 107: Hoisted to base class.

packages/fxa-auth-server/lib/oauth/db/redis.js
```diff
@@ -98,31 +75,6 @@ class OauthRedis {
...
-    }
-    return this.redisRefreshTokens.getRefreshToken(uid, tokenId);
-  }
-
-  /**
-   *
-   * @param {Buffer | string} uid
-   * @return {Promise<{[key: string]: RefreshTokenMetadata}>}
-   */
-  async getRefreshTokens(uid) {
```
dschom, line 119: Hoisted to base class.

packages/fxa-auth-server/lib/oauth/db/redis.js
```diff
@@ -158,20 +110,8 @@ class OauthRedis {
     }
     return this.redisRefreshTokens.removeRefreshTokensForUser(uid);
   }
-
-  /**
-   *
-   * @param {Buffer | string} uid
-   * @param {(Buffer | string)[]} tokenIdsToPrune
-   */
-  pruneRefreshTokens(uid, tokenIdsToPrune) {
```
dschom, line 167: Hoisted to base class.

packages/fxa-auth-server/lib/redis.js
```diff
@@ -4,112 +4,18 @@
...
-const scriptNames = readdirSync(resolve(__dirname, 'luaScripts'), {
-  withFileTypes: true,
-})
-  .filter((dirent) => dirent.isFile() && extname(dirent.name) === '.lua')
-  .map((dirent) => basename(dirent.name, '.lua'));
+class FxaRedis extends RedisShared {
+  constructor(config) {
+    super(config);
 
-function readScript(name) {
```
dschom, line 20: Hoisted to base class.

packages/fxa-shared/db/models/auth/refresh-token-meta-data.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class RefreshTokenMetadata {
+  public lastUsedAt: Date;
+
+  constructor(lastUsedAt?: Date | null) {
+    this.lastUsedAt = lastUsedAt || new Date();
+  }
+
+  /**
+   *
```
vbudhram, line 13: nit: In general, I've always found function comments like this to be a lil strange without a description. I would probably remove comment or add a blurb about it.

packages/fxa-shared/l10n/localizeTimestamp.ts
```diff
@@ -15,9 +14,9 @@ type localizeOptions = {
  */
 export function localizeTimestamp({
   defaultLanguage,
-  supportedLanguages = [],
-}: localizeOptions) {
-  if (supportedLanguages.length === 0) {
+  supportedLanguages,
```
vbudhram, line 17: Curious why the default value was removed?

packages/fxa-shared/lib/logger.ts
```diff
@@ -1,19 +1,15 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+import { ILogger } from '../log';
```
vbudhram, line 4: nit: I'm not sure if there is a lot of added value in having a separate file for `ILogger`, possibly if we had plans it use it somewhere else?

packages/fxa-shared/test/connected-services/formatters.ts
```diff
@@ -0,0 +1,91 @@
...
+    request.app.acceptLanguage = 'de-DE';
+    client.location = {
+      city: 'Berlin',
+      countryCode: 'DE',
+    };
+    clientFormatter.formatLocation(client, request);
+
+    assert.exists(client.location?.country);
+    // Not set, is this a bug?
+    // assert.exists(client.location?.city);
```
vbudhram, line 79: It may very well be? I would think if the city is set in the location, the formatter should return it as well.

packages/fxa-shared/test/db/models/auth/clients.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE clients (
```
vbudhram, line 1: Does it make sense to pull these into `/test/db/models/auth/oauth/*` directory?

packages/fxa-shared/test/db/models/auth/sp_accountDevices.sql
```diff
@@ -0,0 +1,41 @@
+CREATE PROCEDURE `accountDevices_16` (
```
vbudhram, line 1: I don't have a good solution for this, but updates to auth-server procedure would mean this needs to be updated as well.

vbudhram (commented):
This is great @dschom 🎉

I've done a first pass of this and only have a few questions and nits. Tomorrow I will take a fresh look at it and run it locally.

vbudhram (approved):
@dschom I was able to run through this locally and it appears to be working as expected. I ran into https://bugzilla.mozilla.org/show_bug.cgi?id=1753785 which I was hoping this would resolve but doesnt :/ 

Since the PR is touching the `/attached_clients` endpoint (which FF desktop uses) and is a fairly big change, we should do the following as well

1. Update deploy docs to call this out and let ops know to watch for  performance changes in `/attached_clietns`
2. Let QA know and try to verify different devices show up, send tab works and connected devices work as expected.

Other than that, awesome work!

dschom:
@vbudhram thanks for reviewing this! I've addressed as much feedback as I think is possible in this PR. The following things won't be addressed, since they are large in scope, but I'll file follow up issues for them: 

- Improve types in connected-services/accessors.ts.
- Look into potential bug in formatters.ts. (There is a comment on the test case.)
- Look into issue with duplicate attached devices showing up, and get this under test (bugzilla - 1753785)
- Add maintenance item for improving testDatabaseSetup. I think this could probably be automated using msyqldump and targeting local databases fxa, fxa_oath, and fxa_profiles.

Hopefully, the next PR I send your way will not be so big 😅 . And I'll be sure to let QA and ops know about these changes.

Issue #10839, opened 2021-10-27
the service parameter is not emitted to the Auth server under the DNT setting 


YeonjooSmith:
## Description
The Auth server respects the DNT setting from the users so they do not pass certain set of parameters that are related to the user's privacy such as utm parameters. Currently, the service parameter is not emitted to the Auth server under the DNT setting even though it shouldn't be one of the parameters that provide user information. 
### Steps to reproduce
1. I created a FxA account during the VPN subscription funnel with the Firefox private browser. 
2. I queried the flow_id that I got assigned in step 1. 
```
SELECT
  *
FROM
  mozdata.firefox_accounts.fxa_content_auth_events
WHERE
  DATE(timestamp) >= DATE(2021,10,15)
  AND DATE(timestamp) < DATE(2021,10,17)
  AND flow_id IN ('a60216185672b3afe5d211bfa8e353343c8af61742efaf0f9f1fb87ff98cee0f',
    'a820a97c9f3221d8f58408526e9a59f502fb59337bb1844400512fd660cbaab6')-- the flow ids that were assigned during my test
  --AND DATE(timestamp) < DATE(2021,10,10)
ORDER BY
  timestamp
```

### Expected result & Actual result
With the rows with `logger = 'fxa-auth-server` in the query result should have `guardian-vpn` but it is missing. 

### Environment

Firefox private browser

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4148)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/19668/image-20220124-173218.png">image-20220124-173218.png</a>

data-sync-user:
➤ Vijay Budhram commented:

Marking this bug closed since vpn needs to update the request /get-metrics to pass the service name. That should set the value in these events.

data-sync-user:
➤ Yeonjoo Yoo commented:

Vijay Budhram  Has the request /get-metrics to pass the service name been updated?

While I was investigating on another issue, I have two cases with/without DNT enabled and both cases, the service parameter is not passed (show up as undefined_oauth or NULL). I think undefined_oauth might be due to the service parameter added to the url as we discussed before?  

!image-20220124-173218.png|width=939,height=230!

PR #11818, opened 2022-02-01
fix(metrics): Populate relier `service` value if param exists

vbudhram:
## Because

- The service value was not being sent to the auth server

## This pull request

- Populates the `service` param correctly so that it can be sent to the auth-server

Closes: https://github.com/mozilla/fxa/issues/10839

## Checklist


LZoog (approved):
Read over the bug and I think I've got enough context to give this the r+ 👍

PR #11819, opened 2022-02-01
feat(auth): use capabilities of subscribed price, not all prices

### LINKED ISSUE #11804: Capabilities for all product/prices are returned for a subscriber of just one price
When a user has a subscription to price X, we should return capabilities for price X and the capabilities for the product its under. However, we are currently returning all the capabilities for all the prices under the product.

See linked issue for example from MDN.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXA-4516)

bbangert:
Because:

* Prior code grabbed capabilities from all the prices under a users
  subscribed product, rather than just the capabilities for the
  specific price the user subscribed to.

This commit:

* Revamps the capability service to use price id's and merge the
  capabilities of the subscribed price with only its product when
  determining a users capabilities.

Closes #11804

## Checklist

### DIFF
--- a/packages/fxa-auth-server/lib/payments/capability.ts
+++ b/packages/fxa-auth-server/lib/payments/capability.ts
@@ -2,19 +2,18 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import { getUidAndEmailByStripeCustomerId } from 'fxa-shared/db/models/auth';
-import { metadataFromPlan } from 'fxa-shared/subscriptions/metadata';
 import { ACTIVE_SUBSCRIPTION_STATUSES } from 'fxa-shared/subscriptions/stripe';
 import { ClientIdCapabilityMap } from 'fxa-shared/subscriptions/types';
 import Stripe from 'stripe';
 import Container from 'typedi';
 
+import error from '../error';
 import { authEvents } from '../events';
 import { AuthLogger, AuthRequest, ProfileClient } from '../types';
 import { PlayBilling } from './google-play/play-billing';
 import { SubscriptionPurchase } from './google-play/subscription-purchase';
-import { StripeHelper } from './stripe';
-import error from '../error';
 import { PurchaseQueryError } from './google-play/types';
+import { StripeHelper } from './stripe';
 
 function hex(blob: Buffer | string): string {
   if (Buffer.isBuffer(blob)) {
@@ -55,7 +54,7 @@ export class CapabilityService {
     this.stripeHelper = Container.has(StripeHelper)
       ? Container.get(StripeHelper)
       : ({
-          purchasesToProductIds: () => Promise.resolve([]),
+          purchasesToPriceIds: () => Promise.resolve([]),
           fetchCustomer: () => Promise.resolve(null),
           allAbbrevPlans: () => Promise.resolve([]),
         } as unknown as StripeHelper);
@@ -77,15 +76,43 @@ export class CapabilityService {
   }
 
   /**
+   * Create a price id changeset, by returning a priorPriceIds that is a disjoint
+   * set from the currentPriceIds given the affectedPriceIds.
+   *
+   * Due to the asynchronous nature of Stripe, Google, and Apple, we have no method
+   * that lets us know with certainty what price ids the user had active before the
+   * incoming event that triggered this change. To compensate for this, we assume
+   * that whatever the new current state of a users price ids are, that they were
+   * different than before this event. This does imply that we may broadcast that a
+   * user has had a capability removed or added multiple times even if they already
+   * had it, but relying parties can handle this gracefully.
+   */
+  private createPriceIdChangeset({
+    currentPriceIds,
+    affectedPriceIds,
+  }: {
+    currentPriceIds: string[];
+    affectedPriceIds: string[];
+  }) {
+    const priorPriceIds = new Set([...currentPriceIds, ...affectedPriceIds]);
+    for (const affectedPriceId of affectedPriceIds) {
+      if (currentPriceIds.includes(affectedPriceId)) {
+        // Remove the price id from the prior list for processing to assume that it
+        // was previously inactive and ensure we broadcast a change.
+        priorPriceIds.delete(affectedPriceId);
+      }
+    }
+    return [...priorPriceIds];
+  }
+
+  /**
    * Handle a Stripe Webhook subscription change.
    *
    * This handles broadcasting and refreshing the subscription capabilities for
-   * the product ids that were possibly updated.
+   * the price ids that were possibly updated.
    *
    * Stripe supports aligned subscriptions such that a single subscription can
-   * include multiple items for multiple products. We iterate through this list
-   * comparing the current and prior product ids to determine which products
-   * have been added or removed.
+   * include multiple items for multiple products.
    */
   public async stripeUpdate({
     sub,
@@ -112,32 +139,23 @@ export class CapabilityService {
       return;
     }
 
-    // Stripe subscriptions from events do not have product expanded, we filter
-    // by product being the non-expanded string for type checks.
-    const affectedProductIds = sub.items.data
-      .map((item) => item.price.product)
-      .filter((product) => typeof product === 'string') as string[];
-    if (affectedProductIds.length === 0) {
+    // Stripe subscriptions from events do not have price expanded, we filter
+    // by price being the non-expanded string for type checks.
+    const affectedPriceIds = sub.items.data.map((item) => item.price.id);
+    if (affectedPriceIds.length === 0) {
       return;
     }
-    const currentProductIds = await this.subscribedProductIds(uid);
-    let priorProductIds = new Set([
-      ...currentProductIds,
-      ...affectedProductIds,
-    ]);
-    for (const affectedProductId of affectedProductIds) {
-      if (currentProductIds.includes(affectedProductId)) {
-        // Remove the product id from the prior list for processing to assume that it
-        // was previously inactive and ensure we broadcast a change.
-        priorProductIds.delete(affectedProductId);
-      }
-    }
+    const currentPriceIds = await this.subscribedPriceIds(uid);
+    const priorPriceIds = this.createPriceIdChangeset({
+      currentPriceIds,
+      affectedPriceIds,
+    });
     return Promise.all([
       this.profileClient.deleteCache(uid),
-      this.processProductIdDiff({
+      this.processPriceIdDiff({
         uid,
-        priorProductIds: [...priorProductIds],
-        currentProductIds,
+        priorPriceIds,
+        currentPriceIds,
       }),
     ]);
   }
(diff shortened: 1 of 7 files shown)


packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -80,11 +82,11 @@ export class CapabilityService {
    * Handle a Stripe Webhook subscription change.
    *
    * This handles broadcasting and refreshing the subscription capabilities for
-   * the product ids that were possibly updated.
+   * the price ids that were possibly updated.
    *
    * Stripe supports aligned subscriptions such that a single subscription can
    * include multiple items for multiple products. We iterate through this list
```
biancadanforth, line 88: This may perhaps be more future-proofing (as with multiple prices per subscription), but if a subscription can have multiple items for multiple products, then should this code still also iterate through products? I.e. is it possible the affected prices could be from different products? Would that make a difference here?

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -112,32 +114,27 @@ export class CapabilityService {
...
     }
-    const currentProductIds = await this.subscribedProductIds(uid);
-    let priorProductIds = new Set([
-      ...currentProductIds,
-      ...affectedProductIds,
-    ]);
-    for (const affectedProductId of affectedProductIds) {
-      if (currentProductIds.includes(affectedProductId)) {
+    const currentPriceIds = await this.subscribedPriceIds(uid);
+    let priorPriceIds = new Set([...currentPriceIds, ...affectedPriceIds]);
```
biancadanforth, line 124: You had asked me about code comments before -- maybe it's worth mentioning here or in the JSDoc-style comment at the top of this method that we do not know what the change is, because our DB updates and such occur in parallel, which is why we have to do the Set operation? You have a similar disclaimer for the `playUpdates` method below.

Also perhaps we could explain a little more here what `priorPriceIds` is intended to represent? I get what affected are -- those price ids from the subscription that changed in some way TBD. I get what current are -- those price ids the user is currently subscribed to. What is priorPriceIds? Those price ids the user is no longer subscribed to? Maybe we can change the local variable names here to be `activePriceIds` (instead of `currentPriceIds`) and `inactivePriceIds` (instead of `priorPriceIds`)?

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -112,32 +114,27 @@ export class CapabilityService {
...
+    for (const affectedPriceId of affectedPriceIds) {
+      if (currentPriceIds.includes(affectedPriceId)) {
         // Remove the product id from the prior list for processing to assume that it
```
biancadanforth, line 127: Nit: s/product/price

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -157,32 +154,30 @@ export class CapabilityService {
...
    */
   public async playUpdate(uid: string, purchase: SubscriptionPurchase) {
-    const affectedProductId = (
-      await this.stripeHelper.purchasesToProductIds([purchase])
+    const affectedPriceId = (
+      await this.stripeHelper.purchasesToPriceIds([purchase])
     ).shift();
-    if (!affectedProductId) {
+    if (!affectedPriceId) {
       // Purchase is not mapped to a product id.
```
biancadanforth, line 161: Nit: s/product/price

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -157,32 +154,30 @@ export class CapabilityService {
...
-      priorProductIds = currentProductIds.filter(
-        (id) => id !== affectedProductId
-      );
+      priorPriceIds = currentPriceIds.filter((id) => id !== affectedPriceId);
     } else {
       // Its not a current product. Make sure we broadcast that the capabilities
       // associates with this product are gone by adding it to a prior product list.
-      priorProductIds = [...currentProductIds, affectedProductId];
+      priorPriceIds = [...currentPriceIds, affectedPriceId];
     }
```
biancadanforth, line 174: Why not use the same `Set` approach as in `stripeUpdate`; the logic looks the same -- in fact it may be worth promoting to dedupe and avoid divergent code?

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -378,38 +368,43 @@ export class CapabilityService {
...
     // Run through all plans and collect capabilities for subscribed products
-    const plans = await this.stripeHelper.allAbbrevPlans();
-    for (const plan of plans) {
-      if (!subscribedProducts.includes(plan.product_id)) {
+    const prices = await this.stripeHelper.allAbbrevPlans();
+    for (const price of prices) {
+      if (!subscribedPrices.includes(price.plan_id)) {
         continue;
       }
-      const metadata = metadataFromPlan(plan);
```
biancadanforth, line 404: This is the helper that was overwriting any product capabilities I imagine.

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -378,38 +368,43 @@ export class CapabilityService {
...
+        }
+        const capabilityKeys = Object.keys(metadata).filter((key) =>
+          key.startsWith('capabilities')
+        );
+        for (const key of capabilityKeys) {
+          const capabilities = splitCapabilities((metadata as any)[key]);
+          const clientId =
+            key === 'capabilities' ? '*' : key.split(':')[1].trim();
+          for (const capability of capabilities) {
+            (allCapabilities[clientId] ??= new Set()).add(capability);
```
biancadanforth, line 406: Whew; this is a little challenging to understand; TIL the "nullish coalescing" operator in TS. I see it as: add the capability to the existing Set for this client ID if it exists, else create a new set and add the capability to the new set.

biancadanforth (commented):
Leaving some initial thoughts here. Quite a lot of product -> price changes, heh.

Are you planning to file a follow-up ticket to update the Google IAP SKUs in the product/plan metadata in Stripe?

I still need to try it out locally, and I'm not sure about timing for tomorrow since I'll be OoO in the afternoon, so feel free to poke others if needed.

packages/fxa-auth-server/lib/payments/capability.ts
bbangert: Yup, this has been here since before this change. And its a slick little feature, but definitely something to add to ones repertoire. 😉

packages/fxa-auth-server/lib/payments/capability.ts
bbangert: Exactly, it merged them.

packages/fxa-auth-server/lib/payments/capability.ts
chenba: I don't understand the question here.  But afaict the code is not making any assumption on whether the prices in a subscription is from one product or different products.

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -378,38 +368,43 @@ export class CapabilityService {
...
   ): Promise<ClientIdCapabilityMap> {
     const allCapabilities: Record<string, Set<string>> = {};
-
     // Run through all plans and collect capabilities for subscribed products
-    const plans = await this.stripeHelper.allAbbrevPlans();
-    for (const plan of plans) {
-      if (!subscribedProducts.includes(plan.product_id)) {
+    const prices = await this.stripeHelper.allAbbrevPlans();
+    for (const price of prices) {
+      if (!subscribedPrices.includes(price.plan_id)) {
```
chenba, line 390: Just talking to myself here... but this reminds me of the "archived" (soft delete) plans we discussed recently.  We definitely cannot remove archived plans from our cached list.

packages/fxa-auth-server/lib/payments/capability.ts
chenba: Then should we put a fix in `metadataFromPlan` or some other central spot if we should never merge capabilities metadata in that fashion?  That is, after the fix to correctly combine the capabilities metadata, the for loop below will only need to iterate through the plan's metadata.  (Unfortunately we have at least a couple spots where we do the spread operator merge.)

packages/fxa-auth-server/lib/payments/capability.ts
bbangert: It does iterate through all the products relevant for the plans.

packages/fxa-auth-server/lib/payments/capability.ts
bbangert: I think for most metadata the plans should override the products. Such as translation strings, etc. Capabilities however are the only one I know where we need to extend them.

packages/fxa-auth-server/lib/payments/stripe.ts
```diff
@@ -1277,56 +1277,57 @@ export class StripeHelper {
...
+   * Return a list of skus for a given price.
    */
-  productToPlaySkus(product: AbbrevProduct) {
-    const productSkus =
-      product.product_metadata[STRIPE_PRODUCT_METADATA.PLAY_SKU_IDS] || '';
-    return productSkus
+  priceToPlaySkus(price: AbbrevPlan) {
+    const priceSkus =
+      price.plan_metadata?.[STRIPE_PRICE_METADATA.PLAY_SKU_IDS] || '';
+    return priceSkus
```
chenba, line 1285: Where we check for the SKU ids is probably the most confusing part of this PR to me.

IIRC
 - there's no process preventing product managers from adding the SKU ids to a product and/or the product's prices
 - we currently do have these ids on at least one product

Shouldn't we check the "merged" metadata of a plan and its product for the SKU ids?

bbangert: Oh, that's correct, this does need to be merged just like the others.

bbangert: Ah, actually, this works slightly differently. We do merge them, this code is to pull the SKU from the price. If we looked at the product, then *all the prices for this product* would be considered subscribed prices.

packages/fxa-auth-server/lib/payments/capability.ts
```diff
@@ -76,16 +75,44 @@ export class CapabilityService {
...
+    const priorPriceIds = new Set([...currentPriceIds, ...affectedPriceIds]);
+    for (const affectedPriceId of affectedPriceIds) {
+      if (currentPriceIds.includes(affectedPriceId)) {
+        // Remove the price id from the prior list for processing to assume that it
+        // was previously inactive and ensure we broadcast a change.
+        priorPriceIds.delete(affectedPriceId);
+      }
+    }
+    return [...priorPriceIds];
+  }
```
chenba, line 106: Now that this is in one function, we can add a test for this logic.  I think this is tested indirectly right now?

chenba (approved):
Looks good overall, but I haven't had to the chance to try anything out locally.

packages/fxa-auth-server/lib/payments/capability.ts
chenba: Yes I understand.  I think we leaving the door open to another subtle bug when another engineer doesn't know about these changes and think that `metadataFromPlan` does what its name promises.

packages/fxa-auth-server/lib/payments/capability.ts
bbangert: That's correct, its tested indirectly at the moment.

Issue #11312, opened 2021-12-13
Add feature flag


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXA-4308)

PR #11829, opened 2022-02-02
feat(experiment): Add feature flag for google auth

vbudhram:
## Because

- We need to be able to enable/disable the third party auth of Google

## This pull request

- Adds a feature flag so that we can specify based on `clientId`

Closes: https://github.com/mozilla/fxa/issues/11312

## Checklist


LZoog (approved):
LGTM! Normally when I think of a "feature flag" I usually think of an env var of some sort we can turn on or off and not an experiment/AB test, but I'm guessing the idea is we change the rollout rate to 1.0 to turn this on and then only those with a `clientId` will be enrolled.

